PR #6010, opened 2020-05-07
Fixed multipart upload on namespace bucket (Azure as resource)

### LINKED ISSUE #4203: (GAP) Namespace - Upload (multipart) to Azure blob not working
### Environment info
- Version: **2.0**
- Deployment: **N/A**
- Customer: **N/A**

### Actual behavior
1. Azure blob multipart upload is not working in our namespace implementation.

### Expected behavior
1. Should work

### Steps to reproduce
1. Create a namespace bucket on Azure blob resource
2. Use S3 multipart upload on the namespace bucket

romayalon:
### Explain the changes
1. Changed the ETag at the answer to the upload_multipart request to be the hex representation of the block_id instead of the content-md5 header.   
2. Added the { UncommittedBlocks: block_list, CommittedBlocks: [], LatestBlocks: [] } object instead of an array of the ETags.

1. Fixed #4203 

### Testing Instructions:
1.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "noobaa-core",
-  "version": "5.3.1",
+  "version": "5.5.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -2208,7 +2208,8 @@
     "acorn": {
       "version": "7.1.1",
       "resolved": "https://registry.npmjs.org/acorn/-/acorn-7.1.1.tgz",
-      "integrity": "sha512-add7dgA5ppRPxCFJoAGfMDi7PIBXq1RtGo7BhbLaxwrXPOmw8gq48Y9ozT01hUKy9byMjlR20EJhu5zlkErEkg=="
+      "integrity": "sha512-add7dgA5ppRPxCFJoAGfMDi7PIBXq1RtGo7BhbLaxwrXPOmw8gq48Y9ozT01hUKy9byMjlR20EJhu5zlkErEkg==",
+      "dev": true
     },
     "acorn-jsx": {
       "version": "5.1.0",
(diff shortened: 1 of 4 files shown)


nimrod-becker (approved):

package.json
```diff
@@ -116,6 +116,7 @@
     "setimmediate": "1.0.5",
     "sinon": "7.5.0",
     "ssh2": "0.8.9",
+    "stream-buffers": "3.0.2",
```
jeniawhite, line 119: Why did you add this if you don't require it?

package-lock.json
```diff
@@ -4240,12 +4241,30 @@
       "integrity": "sha1-4Ob+aijPUROIVeCG0Wkedx3iqLk="
     },
     "hash-base": {
-      "version": "3.0.4",
-      "resolved": "https://registry.npmjs.org/hash-base/-/hash-base-3.0.4.tgz",
-      "integrity": "sha1-X8hoaEfs1zSZQDMZprCj8/auSRg=",
+      "version": "3.1.0",
```
jeniawhite, line 4244: Just make sure that this version change did not break anything.
If there is a reason to do it.

package-lock.json
```diff
@@ -7342,6 +7361,11 @@
       "resolved": "https://registry.npmjs.org/statuses/-/statuses-1.5.0.tgz",
       "integrity": "sha1-Fhx9rBd2Wf2YEfQ3cfqZOBR4Yow="
     },
+    "stream-buffers": {
```
jeniawhite, line 7364: I don't see it getting used in the code.
Are you sure that we need this?

src/sdk/namespace_blob.js
```diff
@@ -459,15 +458,15 @@ class NamespaceBlob {
         const md5_hash = crypto.createHash('md5');
         const block_list = _.times(num_parts, num => {
             const part = part_by_num[num + 1];
-            md5_hash.update(part.etag);
-            return part.etag;
+            const block_id = Buffer.from(part.etag, 'hex').toString('ascii');
+            return block_id;
         });
-
+        const block_obj = { UncommittedBlocks: block_list, CommittedBlocks: [], LatestBlocks: [] };
```
jeniawhite, line 464: According to this: https://azure.github.io/azure-storage-node/BlobService.html
`The wrapper for block ID list contains block IDs that make up the blob.
Three kinds of list are provided, please choose one to use according to requirement.`
I would change to: { UncommittedBlocks: block_list }.
We aren't interested in sending other parameters and they are not required.
It also requests to send one list, so sending several lists (even if empty) might create a bug in the future.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "noobaa-core",
-  "version": "5.3.1",
+  "version": "5.5.0",
```
jeniawhite, line 3: You said that you've updated Azure-sdk version.
I don't see any update for `azure-storage`.
Am I'm missing something?

src/sdk/namespace_blob.js
```diff
@@ -459,15 +458,15 @@ class NamespaceBlob {
         const md5_hash = crypto.createHash('md5');
         const block_list = _.times(num_parts, num => {
             const part = part_by_num[num + 1];
-            md5_hash.update(part.etag);
```
jeniawhite, line 462: You've remoted the update of the md5_hash.
On complete_object_upload you return an empty md5_hash.
`
res.etag = md5_hash.digest('hex') + '-' + num_parts;
return res;
`

jeniawhite (changes requested):

package-lock.json
romayalon: deleted from description, there wasn't an update after all.

jeniawhite (approved):

PR #6013, opened 2020-05-11
Adding dbg for travis runs

liranmauda:
### Explain the changes
.trvis.yaml:
When the PR headline contains the string "dbg" it will run in dbg mode:

- Will not call to "SUPPRESS_LOGS=true"

This will eliminate the need to edit travis yaml every time we want to debug it.

deploy_minikube.sh:
- Update minikube to v1.8.2 and Kubernetes to 1.7.3
- Installing conntrack
- Update the checks

### Limitations
When it is fully working "Unit Tests" job might fail on:
```
The job exceeded the maximum log length, and has been terminated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,12 @@ env:
     - IMAGE_TAG="noobaa-${TRAVIS_COMMIT}"
     - TESTER_TAG="noobaa-tester-${TRAVIS_COMMIT}"
 
+    - NO_CACHE="NO_CACHE=true"
+    - SUPPRESS_LOGS="SUPPRESS_LOGS=true"
+    - DEPLOY_MINIKUBE_REDIRECT="/dev/null"
+
+    - PR_TITLE=$(curl https://github.com/${TRAVIS_REPO_SLUG}/pull/${TRAVIS_PULL_REQUEST} 2> /dev/null | grep "title" | head -1)
+
 matrix:
   include:
     - language: generic
@@ -20,9 +26,13 @@ matrix:
       services:
         - docker
 
+      before_install:
+        # When the PR headline contains the string "dbg" it will run in dbg mode:
+        # Will not call to "SUPPRESS_LOGS=true"
+        - if [[ "${PR_TITLE}" =~ "dbg" ]] ; then export export SUPPRESS_LOGS=" " ; fi
       script:
-        - make tester NO_CACHE=true SUPPRESS_LOGS=true || exit 1
-        - make test SUPPRESS_LOGS=true
+        - make tester ${NO_CACHE} ${SUPPRESS_LOGS} || exit 1
+        - make test ${SUPPRESS_LOGS}
 
     - language: node_js
       node_js:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by nimrod-becker** [comment]
cool idea

**Review by nimrod-becker — APPROVED** [review]

File: .travis.yml [hunk 1]
```diff
@@ -11,6 +11,11 @@ env:
     - IMAGE_TAG="noobaa-${TRAVIS_COMMIT}"
     - TESTER_TAG="noobaa-tester-${TRAVIS_COMMIT}"
 
+    - NO_CACHE="NO_CACHE=true"
+    - SUPPRESS_LOGS="SUPPRESS_LOGS=true"
+    - if [[ $TRAVIS_PULL_REQUEST_BRANCH =~ "dbg" ]]; then
```
guymguym, line 16: @liranmauda What does this check mean? How can it be used? Maybe worth a code comment?

liranmauda:
> cool idea, just see that we don't overflow the limit of the buffer :) (log)

in most cases, it fails before we overflow the log, there is a possibility that it will, but most of the logs in our case is on the tests themselves.

.travis.yml
guymguym: Just saw the answer in the PR title, better have it in the code too I think

liranmauda: @guymguym 
`[[ $TRAVIS_PULL_REQUEST_BRANCH =~ "dbg" ]]` is:
if the variable contain the string, probebly better to use:
`[[ "$TRAVIS_PULL_REQUEST_BRANCH" =~ "dbg" ]]`
I will update that

guymguym: But is this variable containing the PR headline? Does headline edits work if I want to add debug?

nimrod-becker:
@liranmauda isn't it the branch name?
```
TRAVIS_PULL_REQUEST_BRANCH:
if the current job is a pull request, the name of the branch from which the PR originated.
if the current job is a push build, this variable is empty ("").
```
maybe
```
TRAVIS_COMMIT_MESSAGE
```
but that required modifying the commit message and then back again after the issue is solved

liranmauda:
`TRAVIS_COMMIT_MESSAGE` is not working, its output looks like: 
```
$ echo "$TRAVIS_COMMIT_MESSAGE"
Merge 892915481bd74d30a1d9313efc62b35af9b5f93d into 27974b4b381ca2befc75ea7c066b8b5058bc75b9
```

.travis.yml
liranmauda: apparently no, this method doesn't work as the $TRAVIS_PULL_REQUEST_BRANCH doesn't contain the PR name.

liranmauda:
@nimrod-becker 
> cool idea, just see that we don't overflow the limit of the buffer :) (log)
When it is fully working "Unit Tests" job will fail on
```
The job exceeded the maximum log length, and has been terminated.
```

@guymguym 
> Just saw the answer in the PR description, better have it in the code too I think

Added the comment in .travis.yaml

> But is this variable containing the PR headline? 

This variable wasn't the way, we get the PR headline by doing curl on the PR 

> Does headline edits work if I want to add debug?

Edit the headline by itself will not add/remove the debug. we should edit the headline then restart the travis job (by going into travis, or by pushing a commit).

.travis.yml
```diff
@@ -11,6 +11,12 @@ env:
     - IMAGE_TAG="noobaa-${TRAVIS_COMMIT}"
     - TESTER_TAG="noobaa-tester-${TRAVIS_COMMIT}"
 
+    - NO_CACHE="NO_CACHE=true"
+    - SUPPRESS_LOGS="SUPPRESS_LOGS=true"
+    - DEPLOY_MINIKUBE_REDIRECT="/dev/null"
+
+    - PR_TITLE=$(curl https://github.com/${TRAVIS_REPO_SLUG}/pull/${TRAVIS_PULL_REQUEST} 2> /dev/null | grep "title" | head -1)
```
nimrod-becker, line 18: hehe, nice

nimrod-becker (approved):

PR #6016, opened 2020-05-13
Measurement utils with (almost) no memory pressure.

nb-ohad:
### Explain the changes
1. Add the ability to take time measurments point in the code without inducing any allocation, and by doing so does not introduce and new memory pressure that might affect the measurements. 
2. The measurement samples or a compiled report can be requested for analysis

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/util/object_pool.js
+++ b/src/util/object_pool.js
@@ -1,0 +1,93 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const { make_array } = require('./js_utils');
+
+const alloc_empty_obj = () => ({});
+const noop = () => { /* noop */ };
+
+const rz_policy = Object.freeze({
+    NO_RESIZE: 1,
+    GROW_WHEN_EMPTY: 2,
+    GROW_AND_SHRINK: 3
+});
+
+class ObjectPool {
+    static get resize_policy() {
+        return rz_policy;
+    }
+
+    get capacity() {
+        return this._items.length;
+    }
+
+    get available() {
+        return this.capacity - this._next;
+    }
+
+    constructor(capacity, options = {}) {
+        const {
+            allocator,
+            initializer,
+            empty_value = null,
+            resize_policy = rz_policy.GROW_WHEN_EMPTY
+        } = options;
+
+        if (!capacity && resize_policy === rz_policy.NO_RESIZE) {
+            console.warn('ObjectPool - Invalid operation, cannot create a pool without capacity');
+        }
+
+        this._initalize = initializer || noop;
+        this._empty_value = empty_value;
+        this._resize_policy = resize_policy;
+        this._next = 0;
+
+        this._allocate = allocator || alloc_empty_obj;
+        this._items = make_array(capacity, this._allocate);
+    }
+
+    alloc() {
+        if (this._next === this.capacity) {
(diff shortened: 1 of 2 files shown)


src/util/measurement_utils.js
```diff
@@ -0,0 +1,327 @@
...
+// (which mean only for the first measurement for a new tag). Tags are supposed
+// to identify a location in code or and operation which mean that all should be
+// allocated after enough time, after that the measurement process will be
+// allocation free.
+//
+// Calculating the metrics for the current measuerments and producing a report
+// is an allocaiton heavy operation in shouldbe done outside of paths we would
+//  like to measure.
+//
+// IMPORTANT: Any change introduced to this file should continue to uphole
```
guymguym, line 17: If that's such an important property of this module then you should use a different way to specify it - rename this file to `alloc_free_measurement_utils.js` and remove this comment (it's not the most inviting comment).

BTW for me the term instrumentation is more suitable than measurement. A measurement is a generic term that does not imply that you are measuring the code itself. See https://en.wikipedia.org/wiki/Instrumentation_(computer_programming).

nb-ohad: Ok, the naming thing makes sense I will change. Regarding the comment, I am not sure the name alone can convey the desired result, especially that the file contains also allocating function (massively allocating functions) to produce reports.

guymguym: I don't see how that comment helps more. I would still prefer to remove this comment because the tone of it is not sending the right message.

src/util/object_pool.js
```diff
@@ -0,0 +1,49 @@
...
+        const item = this._items[this._next];
+        this._items[this._next] = this._empty_value;
+        this._next += 1;
+
+        return item;
+    }
+
+    release(item) {
+        if (this._next === 0) {
+            console.warn('ObjectPool - Could not release, pool is at full capacity');
```
guymguym, line 39: Why not grow it to double the previous capacity?

src/util/measurement_utils.js
guymguym: And every instrumentation code will have a reporting function that is not part of the instrumentation code itself but only parses its info

src/util/object_pool.js
nb-ohad: In the general case, you would like to grow the pool in case you cannot allocate any more. Here we are trying to return something to the pool but everything is already in the pool which means something was returned to the pool that was not created from the pool.

Regarding the `alloc()` case, I wanted a stricter pool that enforces a fixed size no preserve the no allocation rule. 
If we will use this for other use cases we can add code to decide what behavior we want upon pool creation (Fixed, Resize, Shrink, etc.)

src/util/measurement_utils.js
nb-ohad: done

src/util/object_pool.js
nb-ohad: I have added a resize_policy as an option the `ObjectPool` constructor, currently, I support the following: `NO_RESIZE`, `GROW_WHEN_EMPTY` (default), `GORW_AND_SHRINK`
I believe this will suffice for most use cases.
I also externalize the capacity and availability of the pool

nimrod-becker (approved):

PR #5984, opened 2020-04-20
Adding accounts cache to S3 endpoints

jeniawhite:
### Explain the changes
1. Added accounts cache to S3 endpoints

### Issues: Fixed #xxx / Gap #xxx
1. 

### Testing Instructions:
1.

PR #6017, opened 2020-05-13
Reuse RPC clients for endpoint requests

nb-ohad:
### Explain the changes
1. creating a new RPC client for each endpoint request induces a measurable memory pressure when doing a lot of small requests in a row by shorting the time between GC cycles.
Reusing the same RPC client for non-overlapping requests prevents that

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -232,33 +232,34 @@ function create_endpoint_handler(rpc, internal_rpc_client, options) {
     async function endpoint_request_handler(req, res) {
         // Allocate an rpc client for the request.
         const rpc_client = rpc_client_pool.alloc();
+        try {
+            // generate request id, this is lighter than uuid
+            req.request_id = `${
+                Date.now().toString(36)
+            }-${
+                process.hrtime()[1].toString(36)
+            }-${
+                Math.trunc(Math.random() * 65536).toString(36)
+            }`;
+            http_utils.parse_url_query(req);
 
-        // generate request id, this is lighter than uuid
-        req.request_id = `${
-            Date.now().toString(36)
-        }-${
-            process.hrtime()[1].toString(36)
-        }-${
-            Math.trunc(Math.random() * 65536).toString(36)
-        }`;
-        http_utils.parse_url_query(req);
+            if (req.url.startsWith('/2015-03-31/functions')) {
+                req.func_sdk = new FuncSDK(rpc_client);
+                return lambda_rest_handler(req, res);
+            }
 
-        if (req.url.startsWith('/2015-03-31/functions')) {
-            req.func_sdk = new FuncSDK(rpc_client);
-            return lambda_rest_handler(req, res);
-        }
+            if (req.headers['x-ms-version']) {
+                req.object_sdk = new ObjectSDK(rpc_client, internal_rpc_client, object_io);
+                return blob_rest_handler(req, res);
+            }
 
-        if (req.headers['x-ms-version']) {
+            req.virtual_hosts = VIRTUAL_HOSTS;
             req.object_sdk = new ObjectSDK(rpc_client, internal_rpc_client, object_io);
-            return blob_rest_handler(req, res);
+            return s3_rest_handler(req, res);
+
+        } finally {
+            rpc_client_pool.release(rpc_client);
         }
-
-        req.virtual_hosts = VIRTUAL_HOSTS;
-        req.object_sdk = new ObjectSDK(rpc_client, internal_rpc_client, object_io);
-        const ret = await s3_rest_handler(req, res);
-
-        rpc_client_pool.release(rpc_client);
-        return ret;
     }
 }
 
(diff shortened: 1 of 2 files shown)


guymguym (changes requested):
@nb-ohad rpc client was meant to be a lightweight object that only holds the auth_token and such. And this is currently a problem with sharing it among requests as we set the auth token to the rpc_client options - 
https://github.com/noobaa/noobaa-core/blob/146eda2619df40793b07e567de31963977b9a1e9/src/sdk/object_sdk.js#L221

There are several ways we can solve:
1. Move the token elsewhere, or make sure we stop using a separate token per request which is what @jeniawhite started to work on in #5984.
2. Move the other GC heavy from rpc client to a shared object - this is mainly because of the APIClient structure that gets created per request here - 
https://github.com/noobaa/noobaa-core/blob/146eda2619df40793b07e567de31963977b9a1e9/src/api/api.js#L41-L84

I prefer to solve option 2.

What do you think?

nb-ohad:
@guy Any kind of object creation in a tight loop that is long-living creates memory pressure, the bigger (or deeper) the object the more pressure it creates. Also a long-lived object whole cycle through the 2 pages of the new generation and then the old generation and it's memory can be recollected only with a Mark an Sweep GC (which is not lightweight).

We should adore to object re-usage (recycling) or to lightweight objects that are created used and then thrown away, like many of the intermediate objects onside functions.

I understand the problem you presented in your comment but I would like to offer a third option:
let's create an RPC client pool to be used by the Endpoint, we can use the ObjectPool implementation I added in my other PR (#6016). Each request will "rent" an object from the pool and return it to the pool in the end.

This way each request will get its own RPC client and all RPC clients will live in the old generation forever (or until the process ends).

What do you think?

guymguym:
Not sure I'm following your logic - why shouldn't we make rpc_client a lightweight object that only has the state of auth_token? Clearly every http request allocates data that will live through the entire lifetime of that http request so that's just part of the same memory which might be scavengered or if the request lives too long in memory it will have to be Marked and sweeped anyhow.

About ObjectPool - sure you can do that, but I still want to separate rpc_client from the all the generic APIClient that is today in the same object.

nb-ohad:
* This memory will not be scavenged for sure. On my tests, each individual request has lived much more than 2 GC cycles (and this is one of the things I think we should improve upon).
Of course, if the memory allocation for such objects is small it will fill the old generation very slowly (if we compare to the way it is now) and defer Mark and Sweeps.

* There aren't many objects that live per request for the whole duration of a request (buffers are the exception). Many "heavy" objects are created used then thrown away. The RPC client is one of these objects that do live the entire request.

* I believe you have desire is to keep API interface as is (so we will not have to touch any call sites) and with this assumption, I must say that the separation you propose is doable but very tricky with to the way the code is currently written (ECMA proxy objects might help):
Each API call is an arrow function closure that preserves context state that is part of the actual RPC client. 

* Memory wise it will be preferable to just pass the auth_token throughout the endpoint lifecycle and create a new RPC on every call site that wants to call, then throw that client away (an API object will be part in a single scavenge at most).
I don't think this is the way to go because I believe that all the gain will be negated by the CPU time taken to create the "heavy" client for each call site.

This is how I came to the conclusion the pooling the clients will be memory efficient and easy to implement and will get us the desired result without changing any call sites. 
Of course, lowering the memory footprint of the client by extracting all the interfaces implementation out into an object that will be everliving should be done, if possible, anyway

src/endpoint/endpoint.js
```diff
@@ -231,18 +244,21 @@ function create_endpoint_handler(rpc, internal_rpc_client, options) {
...
             return blob_rest_handler(req, res);
         }
 
         req.virtual_hosts = VIRTUAL_HOSTS;
-        req.object_sdk = new ObjectSDK(rpc.new_client(), internal_rpc_client, object_io);
-        return s3_rest_handler(req, res);
+        req.object_sdk = new ObjectSDK(rpc_client, internal_rpc_client, object_io);
+        const ret = await s3_rest_handler(req, res);
+
+        rpc_client_pool.release(rpc_client);
```
guymguym, line 260: What about calling release in all the other cases?

nb-ohad: Good catch. I didn't notice that we have early returns

nb-ohad: Changed to use a finally clause to catch all exit conditions (even throws)

guymguym (changes requested):
I am worried that our handlers defer the use of this rpc_client, after the await with rpc_client in its scope.
I know we moved the usage_report monitor to work with internal_rpc_client so that's ok, but even GET object causes this problem because the read stream is piped into the http response stream, but the handler returns and the stream continues...

I think we can think together how to fix rpc client to only create an `options` object and get all the rest from a prototype. What do you think?

nb-ohad:
Maybe I am missing something but the defer is on the RPC and not the API which means that even if you create a new client each time you will get the same problem (if it is really a problem).

I tried to rewrite the client the way you suggested but didn't manage without call site changes, even tried to use ECMA proxy objects.

guymguym:
The defer is on the object_io with that object_sdk + rpc_client.

nb-ohad:
But the RPC client is allocated until the end of the request. Which means all object_io using this rpc_client id done.

guymguym:
OK so as we discussed, we are closing this PR in favor of #6022, right?

nb-ohad:
Closed in favor of #6022

PR #6019, opened 2020-05-14
Functions - add smaller memory sizes options

QuintinBecker:
### Explain the changes
1. In order to be able to choose smaller memory reservations for small functions I added some more selectable memorySizeOptions for function in the UI.

1.noobaa/noobaa-core/issues/6018

### Testing Instructions:
1.


jeniawhite:
@QuintinBecker 
I've attached the relevant people to review your contribution.
Thank you!

frontend/src/app/utils/func-utils.js
```diff
@@ -9,6 +9,22 @@ const handlerFuncNameRegExp = /^[a-zA-Z_$][a-zA-Z0-9_$]*$/;
...
+        label: '16 MB'
+    },
+    {
+        value: 32,
+        label: '32 MB'
+    },
+    {
+        value: 64,
+        label: '64 MB'
+    },
```
nb-ohad, line 27: Currently, the code just uses the first option as the default value. This change will make the default value (in the selection box) 8MB. I think we would like that because most users just use the default value

A good fix is to add a boolean field to the default value:
```
{
        value: 128,
        label: '128 MB'	        
        default: true
}
```

Then change line 144 in `create-func-modal.js` from:
```
        handlerFunc: '',
        memorySize: memorySizeOptions[0].value,
        timeoutMinutes: 0,
```
to: 
```
        handlerFunc: '',
        memorySize: memorySizeOptions.find(opt => opt.default).value,
        timeoutMinutes: 0,
```

nb-ohad (changes requested):

guymguym:
@QuintinBecker Thank you for the PR.
If I remember correctly we have a gap passing the memory limit to the actual function process.

Here is the code where we create the func process, and we should somehow set the new fork memory limit with the `--max-old-space-size` v8 flag. Not sure how to do that for a fork call though.

https://github.com/noobaa/noobaa-core/blob/e02e83e74125ebf31c00033439bcffa07450b854/src/agent/func_services/func_node.js#L34

nb-ohad (approved):

nb-ohad:
Hi #QuintinBecker
Please note that I approved the frontend changes, @guymguym point is still valid.

guymguym:
@QuintinBecker I see in [node.js fork docs](https://nodejs.org/dist/latest-v12.x/docs/api/child_process.html#child_process_child_process_fork_modulepath_args_options) that we can set the execArgv option like this (I didn't test this code):
```
const max_heap_size_mb = Math.min(8, Math.ceil(func.config.memory_size / 1024 / 1024));
const proc = child_process.fork(FUNC_PROC_PATH, [], {
    execArgv: [
        `--max-heap-size=${max_heap_size_mb}`,
        ...process.execArgv,
    ],
    ...
```

See from `node --v8-options`:
```
  --max-semi-space-size (max size of a semi-space (in MBytes), the new space consists of two semi-spaces)
  --max-old-space-size (max size of the old space (in Mbytes))
  --max-heap-size (max size of the heap (in Mbytes) both max_semi_space_size and max_old_space_size take precedence. All three flags cannot be specified at the same time.)
```

I think using the heap size is what we aim for, although not sure if perhaps it will work better if we only set this limit on the old space (which is the one that grows mostly). @nb-ohad any thoughts?

guymguym:
Hey @QuintinBecker 
In any case I think `--max-heap-size` is a good starting point, and if we find it is too restrictive we can change to set the limit only on old space.

I wanted to suggest a quick way to check this - you can manually patch the code on a running noobaa system , see below, BUT NOTE - this patch is only persistent in the pod filesystem and will be gone whenever the pod is deleted and recreated
```bash
$ kubectl exec -it noobaa-core-0 bash

    # opens pod shell

$$ vi /root/node_modules/noobaa-core/src/agent/func_services/func_node.js

    # add --max-heap-size to execArgs

$$ supervisorctl restart all
$$ exit
```

QuintinBecker:
Hey @guymguym, 
sorry for my late response on this, had some other issues in the pipeline. 
We just checked the adoption in cluster and localy and had no issues so far, I pushed all changes in my master branch. 
I will let the changes keep running for some more performance tests. 
Thank you for your support an this topic.

src/agent/func_services/func_node.js
```diff
@@ -31,8 +31,14 @@ class FuncNode {
     invoke_func(req) {
         return this._load_func_code(req)
             .then(func => new P((resolve, reject) => {
+                const max_heap_size_mb = Math.min(8, Math.ceil(func.config.memory_size / 1024 / 1024));
                 const proc = child_process.fork(FUNC_PROC_PATH, [], {
                         cwd: func.code_dir,
+                        execArgv: [
+                        process.execArgv,
```
guymguym, line 38: You want to use the spread operator to flatten the parent's args and append, like so:
```suggestion
                            ...process.execArgv,
```

guymguym (changes requested):
@QuintinBecker Thanks! see small comment above.
Also can you git rebase from master and squash to have a single commit?
Let me know if you want any assistance with git on that.

guymguym:
@QuintinBecker ignore the deepscan error, it's a false positive (or rather a false-new-positive because the reported issue was not affected by your PR).

guymguym:
@nimrod-becker Can you take a look on this deep scan false positive so we can merge? It doesn't seem related to @QuintinBecker's change.

nimrod-becker:
Updated deepscan

src/agent/func_services/func_node.js
```diff
@@ -31,8 +31,14 @@ class FuncNode {
     invoke_func(req) {
         return this._load_func_code(req)
             .then(func => new P((resolve, reject) => {
+                const max_heap_size_mb = Math.min(8, Math.ceil(func.config.memory_size / 1024 / 1024));
```
guymguym, line 34: You probably mean `Math.max` here to have 8 as the minimal number, right?

guymguym (changes requested):
@QuintinBecker See one comment above.
Also please rebase from master and squash to a single commit to prepare for merge.
Thanks

guymguym:
Hey @QuintinBecker are you able to rebase and fix that Math.max ^ 
It would be great to merge this PR

guymguym:
Hey @QuintinBecker - Would you prefer that I finish this PR?
It's pending for a while and I wouldn't want to let it stale here.
So I plan to merge it this week.

QuintinBecker:
@guymguym
Sorry for beeing unrespnsive for so long. Vacation coupled with some issues on OKD4 with the changes we made on the noobaa core. 
We ran into some problems with trigger not running if the option was set.
In addition I think **func.config.memory_size** already stores the requested size in mb and does not need the division. 
As far as I cann see the **memory_size** within the func-object is set to the selected value from the **memorySizeOptions** .

Currently we are running an image excluding the **max_heap_size_mb** modification and it seems to be running stable so far.

guymguym (approved):

guymguym:
Hey @QuintinBecker 
So I merged and then reverted.
Can you please squash you commits to a single commit + sign it off with git commit -s, and then reopen a PR from the branch?
Thanks

PR #6021, opened 2020-05-17
Skip allocations, serialization and deserialization of request data for in-memory requests

nb-ohad:
### Explain the changes
Add  in-memory request type that holds the request data in memory
The new request type is used by in-process requests. to skip the buffer allocations, serialization, and deserialization of the request data and buffers.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/rpc/in_memory_rpc_request.js
+++ b/src/rpc/in_memory_rpc_request.js
@@ -1,0 +1,26 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const RpcRequest = require('./rpc_request');
+
+// Am im memory mapping request key to request data (using en empty buffer
+// as key to comply to current API).
+const req_data_map = new WeakMap();
+
+class InMemoryRpcRequest extends RpcRequest {
+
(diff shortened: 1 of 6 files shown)


guymguym (changes requested):
@nb-ohad I am not sure about all these changes in rpc.js.
I don't want all those functions that used to take a request and operate on it to change to avoid it as it just makes the code complex.
If you just want to control the encoding, maybe we can simply move the encode/decode functions from  request to the connection? Would that work?

nb-ohad:
@guymguym 
I don't see the problem, the request is the same (the new request derive from the last one) the only thing that changed is the internal representation of the data in the request.

I don't really want only to control the encoding/decoding I want to eliminate the allocation of a new set of buffers + serialization and deserialization.

RPC works just the way it worked before with a slight change where the class for requests is selected in run time based on the connection type.

I already thought on a little patch to the current code where on decode I will reset the buffers on the request (copy the reference from the week map) which will solve the problem of the failing tests.

Thoughts?

PR #6022, opened 2020-05-17
Move most of APIClient interface to its promise chain

nb-ohad:
### Explain the changes

This change will decrease the size of an APiClient object to an object with a single field named `options`.
The interface objects (which contain the method declaration) will be generated on the fly on each access and will consist of a single field named `client` that point back to the client that created them.
The methods themselves are not implemented on the prototype of each API object.

A side effect of this change is that we cannot pass an API method directly as a callback because it loses its context. A workaround is to pass an arrow function that preserves the context as I did for node.n2n_signal

This PR is an alternative to the solution presented in PR #6017

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -198,7 +198,7 @@ class Agent {
         );
 
         // register rpc n2n
-        this.n2n_agent = this.rpc.register_n2n_agent(this.client.node.n2n_signal);
+        this.n2n_agent = this.rpc.register_n2n_agent((...args) => this.client.node.n2n_signal(...args));
 
         // TODO these sample geolocations are just for testing
         this.geolocation = _.sample([
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -1,7 +1,6 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
-const _ = require('lodash');
 const url = require('url');
 const { RPC, RPC_BUFFERS, RpcSchema } = require('../rpc');
 const { get_base_address, get_default_ports } = require('../util/addr_utils');
@@ -36,78 +35,84 @@ api_schema.register_api(require('./func_api'));
 api_schema.register_api(require('./func_node_api'));
 api_schema.compile();
 
-class APIClient {
+const client_factory = client_factory_from_schema(api_schema);
 
-    constructor(rpc, default_options) {
-        this.rpc = rpc;
-        this.options = _.create(default_options);
-        this.RPC_BUFFERS = RPC_BUFFERS;
+function client_factory_from_schema(schema) {
+    const client_proto = {
+        RPC_BUFFERS,
 
-        // define the client properties
-        this.auth = undefined;
-        this.account = undefined;
-        this.system = undefined;
-        this.tier = undefined;
-        this.node = undefined;
-        this.host = undefined;
-        this.bucket = undefined;
-        this.events = undefined;
-        this.object = undefined;
-        this.agent = undefined;
-        this.block_store = undefined;
-        this.stats = undefined;
-        this.scrubber = undefined;
-        this.debug = undefined;
-        this.redirector = undefined;
-        this.tiering_policy = undefined;
-        this.pool = undefined;
-        this.cluster_server = undefined;
-        this.cluster_internal = undefined;
-        this.server_inter_process = undefined;
-        this.hosted_agents = undefined;
-        this.frontend_notifications = undefined;
-        this.func = undefined;
-        this.func_node = undefined;
+        async create_auth_token(params) {
+            const res = await this.auth.create_auth(params);
+            this.options.auth_token = res.token;
+            return res;
+        },
 
-        _.each(rpc.schema, api => {
-            if (!api || !api.id || api.id[0] === '_') return;
-            const name = api.id.replace(/_api$/, '');
-            if (name === 'rpc' || name === 'options') throw new Error('ILLEGAL API ID');
-            this[name] = {};
-            _.each(api.methods, (method_api, method_name) => {
-                this[name][method_name] = (params, options) => {
-                    options = _.create(this.options, options);
-                    return rpc._request(api, method_api, params, options);
-                };
-            });
+        async create_access_key_auth(params) {
+            const res = await this.auth.create_access_key_auth(params);
+            this.options.auth_token = res.token;
+            return res;
+        },
+
+        async create_k8s_auth(params) {
+            const res = await this.auth.create_k8s_auth(params);
+            this.options.auth_token = res.token;
+            return res;
+        },
+
+        _invoke_api(api, method_api, params, options) {
+            options = Object.assign(
+                Object.create(this.options),
+                options
+            );
+            return this.rpc._request(api, method_api, params, options);
+        }
+    };
+
+    for (const api of Object.values(schema)) {
+        if (!api || !api.id || api.id[0] === '_') {
+             continue;
+        }
+
+        // Skip common api and other apis that do not define methods.
+        if (!api.methods) {
+            continue;
+        }
+
+        const name = api.id.replace(/_api$/, '');
+        if (name === 'rpc' || name === 'options') {
+            throw new Error('ILLEGAL API ID');
+        }
+
+        const api_proto = {};
+        for (const [method_name, method_api] of Object.entries(api.methods)) {
+            // The following getter is defined as a function and not as an arrow function
+            // to prevent the capture of "this" from the surrounding context.
+            // When invoked, "this" should be the client object. Using an arrow function
+            // will capture the "this" defined in the invocation of "new_client_factory"
+            // which is "undefined"
+            api_proto[method_name] = function(params, options) {
+                return this.client._invoke_api(api, method_api, params, options);
+            };
+        }
+
+        // The following getter is defined as a function and not as an arrow function
+        // on purpose. please see the last comment (above) for a detailed explanation.
+        Object.defineProperty(client_proto, name, {
+            enumerable: true,
+            get: function() {
+                const api_instance = Object.create(api_proto);
+                api_instance.client = this;
+                return Object.freeze(api_instance);
+            }
         });
     }
 
-    /**
-     * extend the rpc client prototype with convinient methods
-     */
-    create_auth_token(params) {
-        return this.auth.create_auth(params)
-            .then(res => {
-                this.options.auth_token = res.token;
-                return res;
-            });
-    }
-
-    create_access_key_auth(params) {
-        return this.auth.create_access_key_auth(params)
-            .then(res => {
-                this.options.auth_token = res.token;
-                return res;
-            });
-    }
-
-    async create_k8s_auth(params) {
-        const res = await this.auth.create_k8s_auth(params);
-        this.options.auth_token = res.token;
-        return res;
-    }
-
+    return (rpc, options) => {
+        const client = Object.create(client_proto);
+        client.rpc = rpc;
+        client.options = options ? Object.create(options) : {};
+        return client;
+    };
 }
 
 function new_router_from_base_address(base_address) {
@@ -156,7 +161,7 @@ function new_rpc_from_routing(routing_table) {
     }
 
     return new RPC({
-        APIClient,
+        client_factory,
         schema: api_schema,
         router: routing_table,
         api_routes: {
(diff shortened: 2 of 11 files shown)


src/api/api.js
```diff
@@ -36,78 +35,85 @@ api_schema.register_api(require('./func_api'));
...
+        if (name === 'rpc' || name === 'options') {
+            throw new Error('ILLEGAL API ID');
+        }
+
+        const api_proto = {};
+        for (const [method_name, method_api] of Object.entries(api.methods)) {
+            // The following getter is defined as a function and not as an arrow function
+            // to prevent the capture of the surrounding context when using this
+            api_proto[method_name] = function(params, options) {
+                return this.client._invoke_api(api, method_api, params, options);
```
guymguym, line 96: Can you add a comment that explains what is `this` in this context?
I think that our linter also requires it for non-class functions that use this by adding a jsdoc comment `/** @this {TYPE} */`

src/api/api.js
```diff
@@ -36,78 +35,85 @@ api_schema.register_api(require('./func_api'));
...
+        return this.rpc._request(api, method_api, params, options);
     }
+};
 
+function new_client_factory(rpc) {
+    // Copy the shared part instead of using it as a proto
+    // to skip one hoop when invoking these methods.
+    const client_proto = {
+        ...shared_client_proto,
+        rpc
```
guymguym, line 73: Perhaps we can leave this rpc property undefined and the rpc itself will set the rpc property to this after the creation of the client? This way we can compile this factory method just from the schema.

src/api/api.js
```diff
@@ -36,78 +35,85 @@ api_schema.register_api(require('./func_api'));
...
 
-    async create_k8s_auth(params) {
+    create_k8s_auth: async function(params) {
         const res = await this.auth.create_k8s_auth(params);
         this.options.auth_token = res.token;
         return res;
+    },
+
+    _invoke_api(api, method_api, params, options) {
+        options = Object.assign(
```
guymguym, line 60: Isn't `Object.create(this.options, options)` doing that?

src/api/api.js
```diff
@@ -36,78 +35,85 @@ api_schema.register_api(require('./func_api'));
...
-     * extend the rpc client prototype with convinient methods
-     */
-    create_auth_token(params) {
-        return this.auth.create_auth(params)
-            .then(res => {
-                this.options.auth_token = res.token;
-                return res;
-            });
-    }
+    create_auth_token: async function(params) {
```
guymguym, line 41: It's not clear why you have to make these properties that refer to functions instead of just member functions?
If this is really needed then maybe explain in a short comment what is the reason.

src/api/api.js
```diff
@@ -156,7 +162,7 @@ function new_rpc_from_routing(routing_table) {
     }
 
     return new RPC({
-        APIClient,
+        new_client_factory,
```
guymguym, line 165: As I commented above, if possible I would try to pass to RPC a compiled new_client function that already includes the schema.

src/agent/agent.js
```diff
@@ -198,7 +198,9 @@ class Agent {
         );
 
         // register rpc n2n
-        this.n2n_agent = this.rpc.register_n2n_agent(this.client.node.n2n_signal);
+        this.n2n_agent = this.rpc.register_n2n_agent((...args) =>
```
guymguym, line 201: (style) I would keep those one liners if not too long, because these are just a small technicality.

src/api/api.js
```diff
@@ -36,78 +35,85 @@ api_schema.register_api(require('./func_api'));
...
 api_schema.compile();
 
-class APIClient {
-
-    constructor(rpc, default_options) {
-        this.rpc = rpc;
-        this.options = _.create(default_options);
-        this.RPC_BUFFERS = RPC_BUFFERS;
-
-        // define the client properties
```
guymguym, line 46: The reason I added all these properties explicitly was to allow our typescript compiler be ab le to understand the structure of that client and not give errors in the editor on every access to these properties.

So instead of setting those explicitly, can we just apply a type using jsdoc? You can define the type itself (APIClient) in nb.d.ts and then just make sure the new_client function is defined to `@return {nb.APIClient}`. Let me know if you need more help to get this done.

src/api/api.js
nb-ohad: I will, no problem. 
FYI, the linter will only complain if it was an arrow function. Making this a traditional function fix all of these problem

src/api/api.js
nb-ohad: We can but then this will make every client proto twice as big (it will hold the options object and the RPC reference.

The fact is that the RPC is shared for all these instances which make sense to put it on the prototype

src/api/api.js
nb-ohad: No, `Object.create`second parameter is an object properties descriptor, not a plain object. What we want is to copy the options from `options` into the newly created object, this is exactly what Object.assign was designed to do.

src/api/api.js
nb-ohad: This is a template object, member function are not exactly the same as properties that hold functions. What we want here is a plain object holding reference to functions and properties that are then copied to the actual client prototype.

I will put a comment in the code.

src/agent/agent.js
nb-ohad: I think this makes it much less readable. but if you insist I will change

src/api/api.js
nb-ohad: Yes. this is a good idea.
But this will only give you a shallow structure, what really we want is a scheme to typescript definition converter (the same way I implemented for the FE schema). But this is for another task.

src/api/api.js
nb-ohad: Please see my comment above regarding the RPC reference.

src/api/api.js
nb-ohad: done

src/api/api.js
nb-ohad: done

src/agent/agent.js
nb-ohad: done

src/api/api.js
guymguym: This is the rule - https://eslint.org/docs/rules/no-invalid-this
Arrow functions are not relevant to it at all as this in them will resolve to outer context.
But for normal functions this rule makes sure `this` is used only in "class" context.

src/api/api.js
guymguym: The size different is negligible.
I don't think this goes against the change as we discussed it this change is mostly to allow scavenging, which a reference to rpc will not change.

src/api/api.js
guymguym: I probably mixed up with the api of https://lodash.com/docs/4.17.15#create which was used in this code before your changes to create the combined options `options = _.create(this.options, options);`

src/api/api.js
guymguym: Shallow is enough to start with

src/api/api.js
guymguym: Another thing on readability - I think the separation of shared_client_proto and the client_proto feels a bit weird because in any case this is called just once to create a client_proto, and shared_client_proto is very small, so why not just create it once inside new_client_factory?

guymguym: About the rpc reference, I still think the rpc reference should be added to the client_proto, so that client objects are not even effected in size, right?

nb-ohad: I agree that if we remove the RPC form the proto then we can generate the factory only once and not once per RPC instance. In this case, we can merge the shard and actual proto.
The separation was done because in the current case we have a new client_porto for each RPC instance.

Your last comment does not make sense to me. Changing the proto of an object after creation is not recommended performance-wise. It will cause decompilation of all code that uses instances of objects that use this proto.
Also, this will bring us back to a client_proto per RPC instance which you advocated against in your comments

guymguym: I don't suggest changing the proto after creating objects from it.
I meant before - when RPC ctor is running.

nb-ohad: You can't you have one proto for many RPCs

src/agent/agent.js
```diff
@@ -198,7 +198,7 @@ class Agent {
         );
 
         // register rpc n2n
-        this.n2n_agent = this.rpc.register_n2n_agent(this.client.node.n2n_signal);
+        this.n2n_agent = this.rpc.register_n2n_agent((...args) => this.client.node.n2n_signal(...args));
```
guymguym, line 201: We also have `n2n_proxy` being used in a similar way in server_rpc.
I think you need to fix it too, right?
Can we somehow make sure we didn't miss any other rpc functions passed to functions instead of called from their object?

src/api/api.js
```diff
@@ -36,78 +35,84 @@ api_schema.register_api(require('./func_api'));
...
+            return res;
+        },
+
+        async create_k8s_auth(params) {
+            const res = await this.auth.create_k8s_auth(params);
+            this.options.auth_token = res.token;
+            return res;
+        },
+
+        _invoke_api(api, method_api, params, options) {
```
guymguym, line 62: why not async?

src/api/api.js
guymguym: Can you add this APIClient type as part of this PR?
I hate to see our editors start marking all our rpc calls in red.

nb-ohad: I will, as a last thing after I fix the failing tests

src/api/api.js
nb-ohad: I can, as a marking.
But please note that adding the async keyword without a need will cause another nextTick schedule instead of returning immediately.

src/agent/agent.js
nb-ohad: Sadly I have no automatic check I can run to make sure. only the tests we run for each build.
I am reviewing the failed test in order to find other locations

src/api/api.js
guymguym: but rpc._request always returns a promise so it shouldn't reschedule twice, right?

src/api/api.js
guymguym: It errors with strict type checking, here is an example:
```bash
$ cat a.js
'use strict';
function new_client() { return {}; }
const client = new_client();
client.bucket.create_bucket();

$ tsc --allowJs --checkJs --noEmit a.js
<no errors>

$ tsc --allowJs --checkJs --noEmit --strict a.js
a.js:4:8 - error TS2339: Property 'bucket' does not exist on type '{}'.

4 client.bucket.create_bucket();
         ~~~~~~

Found 1 error.
```

src/api/api.js
nb-ohad: I decoupled the API from the RPC (adding the RPC reference to the client).

src/api/api.js
nb-ohad: It will reschedule twice. Even If V8 had an optimization to understand that the return value will always be a promise and not do a second reschedules it will not be able to do it in our case because we do not return native promises from RPC, we return bluebird promises (and lose a lot of the new optimization V8 introduce to the compilation of promise-based code).

We want to just pass through the promise so let just drop the async for now

src/agent/agent.js
nb-ohad: All tests have passed so I can only assume the in the coveraged code we do not have these kinds of problems.

src/api/api.js
nb-ohad: Your case is not accurate: 
```
'use strict';
const proto = {};
function new_client() { return Object.create(proto); }
const client = new_client();
client.bucket.create_bucket();

tsc --allowJs --checkJs --noEmit a.js
<no errors>

tsc --allowJs --checkJs --noEmit --strict a.js
<no errors>
```

nb-ohad: But I still see the benefit for auto-complete reasons

src/api/api.js
guymguym: Maybe I'm missing something, if I define an `async function` it still doesn't schedule anything, only `await` will do that, and since the function body is still just returning a promise from another function, then isn't `async` only a declarative annotation that makes this function always convert its return value to a promise?

guymguym: On a second note, we can change rpc _request to return native promise too, shouldn't be difficult at all.

nb-ohad: I will try to clarify. By spec, every function marked as async returns a promise so even:
```
function Async() { return 1; }
```
will return a promise.

Now the V8 engine does not know about alternate implementations of the Promise interface so in order to guarantee a return value of a promise it will wrap whatever you give him with a promise. If the return value is a "thanable" it will chain to it, which is the case with bluebird. The reason that not every "thanable" is treated as a promise is that V8 it cannot guarantee compliance with the spec. 

So even:
```
async function () { return P.resolve() }  
```
will be rescheduled twice, once internally inside bluebord using `nextTick` and once internally inside V8

nb-ohad: Regarding the rpc_request change.  It's a good direction head to but even with a native promise, there is no guarantee that the optimization will kick in. There are reasons that are far too complex to explain here. I will try to find the V8 blog article and link.

ESLint has a rule that complains about wrong uses, for these exact reasons, of the async keyword but we choose to turn it off.

guymguym: OK thanks. We should get rid of bluebird whenever we can.

src/api/api.js
nb-ohad: done

guymguym (approved):

guymguym:
@nb-ohad Can you squash your commits?

PR #6024, opened 2020-05-19
Remove deprecated phone home code

nimrod-becker:
### Explain the changes
1. Remove deprecated phone home code


src/server/bg_services/server_monitor.js
```diff
@@ -69,61 +66,11 @@ async function run_monitors() {
...
-                        `DNS server/s cannot resolve Phone home server name`,
-                        Dispatcher.rules.once_daily);
-                    break;
-                case "CANNOT_REACH_DNS_SERVER":
-                    monitoring_status.dns_status = "UNREACHABLE";
-                    break;
-                default:
-                    break;
-            }
-            if (_.isEmpty(server_conf.dns_servers)) {
```
jeniawhite, line 122: Notice that we delete the dns_servers in case we do not have them in our server_conf.
This change was removed in your new proposed code, this means that we will always send `dns_status = "OPERATIONAL"` even when we do not have `dns_servers` in our `server_conf`.
Currently, this value is exported in cluster_info which is used by the FE (but not this particular value from what I saw).

jeniawhite (approved):

src/server/bg_services/server_monitor.js
nimrod-becker: fixed

src/server/system_services/stats_aggregator.js
```diff
@@ -1157,7 +1093,7 @@ async function _perform_full_cycle() {
         })
     });
 
-    await _handle_payload(payload);
+    await _handle_payload();
```
jackyalbo, line 1096: maybe a better name if it doesn't handle any payload... send_object_usage or something like this

src/server/bg_services/server_monitor.js
```diff
@@ -69,61 +69,14 @@ async function run_monitors() {
...
-                delete monitoring_status.dns_status;
-            }
-        })
-        .catch(err => dbg.warn('Error when trying to check dns and phonehome status.', err.stack || err));
+    if (!_.isEmpty(server_conf.dns_servers)) {
```
jackyalbo, line 72: don't understand why we are still returning those... sound like it will never change. maybe we don't need check_dns_and_phonehome at all...

nimrod-becker: I think we don't I just wanted to remove the ping we do to the old ph server. But agree we need to clean up the schema and API especially around the server HB area

PR #6027, opened 2020-05-19
RPC: Decouple message creation from message encoding/decoding

nb-ohad:
### Explain the changes
1.  Create a new RpcMessage class to represent a generic non-encoded RPC message that can describe any op (in contrast to RpcReqeust that can be used solely for requests and responses).
2. Move encoding/decoding responsibilities (from and to transit representation) into rpc_connection where the base class just delegate to the old encode/decode code inside RpcRequest.
4. Decode on "message" and emit a "decoded_message" event with the decoded RpcMessage.
5. Override the encoding and decoding for `fcall` connections to passthrough the RpcMessage without encoding or decoding.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/rpc/rpc.js
+++ b/src/rpc/rpc.js
@@ -1,6 +1,8 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
+/** @typedef {import('./rpc_base_conn')} RpcBaseConnection **/
+
 const _ = require('lodash');
 const util = require('util');
 const assert = require('assert');
@@ -8,11 +10,13 @@ const assert = require('assert');
 const EventEmitter = require('events').EventEmitter;
 
 const P = require('../util/promise');
+const js_utils = require('../util/js_utils');
 const dbg = require('../util/debug_module')(__filename);
 const config = require('../../config');
 const RpcError = require('./rpc_error');
 const url_utils = require('../util/url_utils');
 const RpcRequest = require('./rpc_request');
+const RpcMessage = require('./rpc_message');
 const RpcWsServer = require('./rpc_ws_server');
 const RpcN2NAgent = require('./rpc_n2n_agent');
 const RpcTcpServer = require('./rpc_tcp_server');
@@ -213,7 +217,7 @@ class RPC extends EventEmitter {
                 }
 
                 // send request over the connection
-                return req.connection.send(req._encode_request(), 'req', req);
+                return req.connection.send(req.make_request_message(), req);
 
             })
             .then(() => {
@@ -320,7 +324,7 @@ class RPC extends EventEmitter {
                 'reqid', msg.body.reqid,
                 'connid', conn.connid);
             req.error = new RpcError('NO_SUCH_RPC_SERVICE', 'No such RPC Service ' + srv);
-            return conn.send(req._encode_response(), 'res', req);
+            return conn.send(req.make_response_message(), req);
         }
 
         return P.resolve()
@@ -378,8 +382,7 @@ class RPC extends EventEmitter {
                     'reqid', req.reqid,
                     'connid', conn.connid);
 
-                return conn.send(req._encode_response(), 'res', req);
-
+                return conn.send(req.make_response_message(), req);
             })
             .catch(err => {
 
@@ -397,7 +400,7 @@ class RPC extends EventEmitter {
                     req.error = new RpcError(err.rpc_code || 'INTERNAL', err.message, { retryable: true });
                 }
 
-                return conn.send(req._encode_response(), 'res', req);
+                return conn.send(req.make_response_message(), req);
             });
     }
 
@@ -584,14 +587,14 @@ class RPC extends EventEmitter {
         conn.once('connect', () => {
             if (this.routing_hint) {
                 dbg.log0('RPC ROUTING REQ SEND', this.routing_hint, conn.connid, conn.url.href);
-                conn.send(RpcRequest.encode_message({
+                conn.send(new RpcMessage({
                     op: 'routing_req',
                     reqid: conn._alloc_reqid(),
                     routing_hint: this.routing_hint,
                 }));
             }
         });
-        conn.on('message', msg => this._on_message(conn, msg));
+        conn.on('decoded_message', msg => this._on_message(conn, msg));
         conn.on('close', err => this._connection_closed(conn, err));
         // we let the connection handle it's own errors and decide if to close or not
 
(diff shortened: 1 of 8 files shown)


guymguym (changes requested):

guymguym (approved):
Looks good, did you check the CI failures?

nb-ohad:
@guymguym I restarted the build

src/rpc/rpc_message.js
```diff
@@ -0,0 +1,72 @@
...
+    constructor(body, buffers) {
+        this.body = body;
+        this.buffers = buffers;
+    }
+
+    clone() {
+        const { body, buffers } = this;
+        return new RpcMessage(
+            _.cloneDeep(body),
+            buffers && buffers.map(Buffer.from)
```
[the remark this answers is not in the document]
nb-ohad: I already added a comment on the rpc_fcall call site

guymguym:
@nb-ohad It failed. Not clear if related so I restarted the tests now.

nb-ohad:
It seems that not serializing the payload (body) into a buffer (and then deserializing) creates a difference with regard to encoding. 

I am working on trying to pinpoint the problem. 
I will update here

guymguym:
any news on that?

nb-ohad:
@guymguym ping on review for the second commit

src/rpc/rpc_base_conn.js
```diff
@@ -129,12 +144,37 @@ class RpcBaseConnection extends EventEmitter {
         return this._state === STATE_CLOSED;
     }
 
+    _encode_message(msg) {
```
guymguym, line 147: jsdoc:
```suggestion
    /**
     * @param {RpcMessage} msg
     * @returns {Buffer[]}
     */
    _encode_message(msg) {
```

src/rpc/rpc_base_conn.js
```diff
@@ -129,12 +144,37 @@ class RpcBaseConnection extends EventEmitter {
...
+        return msg.encode();
+    }
+
+    _decode_message(encoded_message) {
```
guymguym, line 151: jsdoc:
```suggestion
    /**
     * @param {Buffer[]} encoded_message
     * @returns {RpcMessage}
     */
    _decode_message(encoded_message) {
```

guymguym (changes requested):
@nb-ohad One last comment about making the omit conditional to avoid copying the params/reply objects, the copy is shallow but still this is very much unneeded.
Other than that it looks good to me.

src/rpc/rpc.js
```diff
@@ -732,8 +733,7 @@ class RPC extends EventEmitter {
     /**
      *
      */
-    _on_message(conn, msg_buffer) {
-        const msg = RpcRequest.decode_message(msg_buffer);
+    _on_message(conn, msg) {
```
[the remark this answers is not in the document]
nb-ohad: I do need to learn this JSDoc thingy (mainly syntax)

src/rpc/rpc_base_conn.js
nb-ohad: This will break as this does not return buffers in all cases, especially for the case of fcall

src/rpc/rpc_base_conn.js
nb-ohad: This will break as this does accept  buffers in all cases, especially for the case of fcall

src/rpc/rpc.js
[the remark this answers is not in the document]
nb-ohad: @guy  
I cannot use RpcBaseConnection here because it is not imported into the file.
If I import it then it is an unused variable.

Any suggestions?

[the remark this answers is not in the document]
guymguym: oops, I accidently wrote `@type` instead of `@param`. thanks for fixing, resolving.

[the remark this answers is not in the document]
guymguym: About RpcBaseConn you can use it like this example - 
https://github.com/noobaa/noobaa-core/blob/c3cce9ba10174b0ab41481fe1aed204b8fd02157/src/sdk/map_api_types.js#L5

src/rpc/rpc_base_conn.js
guymguym: OK so you can use `@returns {Buffer[] | object}`

guymguym: Or set a `/** @template T */` on the RpcBaseConn class and that will be the encoded message type, so that `RpcFcallConn extends RpcBaseConn<object>` but the others `extends RpcBaseConn<Buffer[]>`.

guymguym (commented):
@nb-ohad To summarize this PR - it tries to improve how rpc_fcall (which is an optimized rpc path for calls which are executed in the same process) by getting rid of the intermediate large memory allocations created by JSON.parse(JSON.stringify(..)) and instead use cloneDeep. It does introduce some subtleties of cloning in rpc_fcall, but the rest of the system is not affected.

However, how do you measure the effect of this change? Did it help? is it possible that it is somehow slower/heavier than previous solution? Of course - I don't really think so. But we need to work with some numbers here.

I would suggest to enhance rpc_benckmark.js for that. You can add a new `--fcall` flag that will do the benchmark inside one process (both client and server). And then our main metric is operations/second and the higher the better, or something more complicated is needed to see the effect? We can run it with `node --trace-gc` for GC info. And we can add anything else we need to that benchmark to simulate the effect on a real process...

nb-ohad:
@guymguym This is optimization It is not about getting rid of `JSON.parse`/`JSON.stringify` because `deepClone` will also create a new object tree. It is about removing the need in the allocation of buffers for interprocess calls (header and body buffers)

regarding measurements, This is a thing that is very hard to measure and the normal `rpc_banchmark` will not yield any meaningful results. We initiated this task as a result of the measurement that we did with a high volume, small object tests
The results indicated that a lot of memory pressure was produced in this kind of scenario coming from 3-4 prime locations in the code, one of them was the allocation of buffers and cloning the message body object Tree. 

We decided to tackle the points according to their impact and ease of fixing. So we choose to improve memory allocations in the instantiation of our API clients (https://github.com/noobaa/noobaa-core/pull/6022) and minimizing allocations for RPC calls via 'fcall' connections

src/rpc/rpc_base_conn.js
nb-ohad: done

src/rpc/rpc_base_conn.js
nb-ohad: done

nb-ohad:
@guymguym as of your request for enhancements to `rpc_benckmark`:
https://github.com/noobaa/noobaa-core/pull/6173

guymguym (approved):

PR #6028, opened 2020-05-20
Updating ownership of mongo PV will now run recursively

jackyalbo:
### Explain the changes
1. kube_pv_chown will now recursively run over mongo directories tree - changing the owner uid
2. will run on /data instead of /mongo_data 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/deploy/NVA_build/noobaa_init.sh
+++ b/src/deploy/NVA_build/noobaa_init.sh
@@ -110,14 +110,12 @@ prepare_server_pvs() {
 }
 
 prepare_mongo_pv() {
-  local dir="/mongo_data/mongo/cluster/shard1"
+  local dir="/data/mongo/cluster/shard1"
 
   # change ownership and permissions of mongo db path
   ${KUBE_PV_CHOWN} mongo
 
   mkdir -p ${dir}
-  chgrp 0 ${dir}
-  chmod g=u ${dir}
 }
 
 run_endpoint_container() {
--- a/src/native/tools/kube_pv_chown.cpp
+++ b/src/native/tools/kube_pv_chown.cpp
@@ -1,11 +1,13 @@
 #ifndef WIN32
 #include <cstring>
+#include <dirent.h>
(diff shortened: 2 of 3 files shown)


src/deploy/NVA_build/noobaa_init.sh
```diff
@@ -116,8 +116,6 @@ prepare_mongo_pv() {
   ${KUBE_PV_CHOWN} mongo
 
   mkdir -p ${dir}
```
dannyzaken, line 118: change dir above to `/data/mongo/cluster/shard1`
also, look for all other places that use `mongo_data`

src/native/tools/kube_pv_chown.cpp
```diff
@@ -35,6 +37,28 @@ change_path_permissions(const char* path, long uid)
...
+    if (!dir) {
+        if (errno != ENOTDIR) {
+            cerr << "Error: opendir(" << path << ") failed " << strerror(errno) << endl;
+            exit(1);
+        }
+        return;
+    }
+    struct dirent* entry;
+    while ((entry = readdir(dir)) != NULL) {
+        if (std::string(entry->d_name).compare(".") != 0 && std::string(entry->d_name).compare("..") != 0) {
```
guymguym, line 54: I think this can be more readable:
```suggestion
        std::string entry_name = entry->d_name;
        std::string entry_path = path + "/" + entry_name;
        if (entry_name != "." && entry_name != "..") {
            change_path_permissions_recursive(entry_path, uid);
        }
```

dannyzaken:
requires a more complete fix that includes setting SecurityContext and matching SCC. closing for now

guymguym:
@jackyalbo @dannyzaken How about reopening this PR.
Even if we decide to use another solution, I think this code is improved with this PR.
Do you have a concern that these changes will cause issues?
Perhaps we call the recursive code only when a `--recursive` flag is used so we can experiment with it manually?

PR #6037, opened 2020-05-27
md5_mb (multibuffer) using isa-l_crypto

guymguym:
### Explain the changes
All mains now import `src/util/fips` module on startup which auto detects if running in fips mode based on `/proc/crypto/fips`. This can also be forced using the FIPS=1 env.

In fips mode in order to allow md5 for non cryptographic use cases needed for the S3 protocol, we use md5_mb from isa-l_crypto. Since crypto.createHash is called from our code and also node modules, we monkey-patch the crypto library and replace the createHash('md5'), and also use it natively in the upload pipeline (ChunkSplitter).

NOTE: Using md5_mb with a single stream provides lower performance compared to openssl MD5. See this issue on single buffer performance - https://github.com/intel/isa-l_crypto/issues/45. We try to leverage the multibuffer capability by sharing a thread_local ctx mgr and deferring the flushing of contexes so that if multiple streams are processed concurrently it will be flushed together. The single CPU core performance gain can be very significant - see the performance report here - https://github.com/intel/isa-l/wiki/Documentation.

1. NA

### Testing Instructions:
1. Run on FIPS enabled host.

### DIFF
--- a/src/deploy/NVA_build/dev.Dockerfile
+++ b/src/deploy/NVA_build/dev.Dockerfile
@@ -1,0 +1,35 @@
+FROM centos:8 
+
+ENV container docker
+
+RUN dnf update -y -q && \
+    dnf install -y -q \
+        bash bash-completion \
+        wget curl nc unzip which less vim \
+        python2 python2-setuptools \
+        python3 python3-setuptools \
+        gdb strace lsof \
+        openssl && \
+    dnf --enablerepo=PowerTools install -y -q yasm && \
+    dnf group install -y -q "Development Tools" && \
+    dnf clean all
+
+RUN mkdir -p /usr/local/lib/python3.6/site-packages
+RUN alternatives --set python /usr/bin/python3
+
+WORKDIR /noobaa
+
+COPY ./.nvmrc ./.nvmrc
+COPY ./src/deploy/NVA_build/install_nodejs.sh ./
+RUN chmod +x ./install_nodejs.sh && \
+    ./install_nodejs.sh $(cat .nvmrc) && \
+    npm config set unsafe-perm true && \
+    echo '{ "allow_root": true }' > /root/.bowerrc
+
+COPY ./package*.json ./
+RUN npm install && \
+    npm cache clean --force
+
+COPY . ./
+USER 0:0
+CMD [ "bash" ]
--- a/src/tools/coding_speed.js
+++ b/src/tools/coding_speed.js
@@ -1,6 +1,8 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
+require('../util/fips');
+
 // const _ = require('lodash');
 const argv = require('minimist')(process.argv);
 const stream = require('stream');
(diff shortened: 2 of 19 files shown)


src/deploy/NVA_build/dev.Dockerfile
```diff
@@ -0,0 +1,34 @@
+FROM centos:8
```
liranmauda, line 1: 1. What is the use case for this Docker file?
2. Can we add an entry in the Makefile?
3. We can Update the https://github.com/noobaa/noobaa-core/wiki/Building-and-Testing

src/tools/coding_speed.js
```diff
@@ -29,6 +29,7 @@ argv.erase = Boolean(argv.decode && (argv.erase !== false)); // default is true
 argv.ec = Boolean(argv.ec); // default is false, use --ec
 argv.md5 = Boolean(argv.md5); // default is false, use --md5
 argv.sha256 = Boolean(argv.sha256); // default is false
+argv.md5_mb = Boolean(argv.md5_mb); // default is false, use --md5_mb
```
liranmauda, line 32: What happens if we are passing argv.md5 and argv.md5_mb? 
We can change that from 2 flags to one enum

liranmauda (approved):

src/tools/coding_speed.js
guymguym: This was changed so that the mode is detected in src/util/fips. Resolving.

src/deploy/NVA_build/dev.Dockerfile
guymguym: I am ignoring this for now. Will get back to it later.

src/util/fips.js
```diff
@@ -0,0 +1,85 @@
...
+function detect_fips_mode() {
+    if (process.env.FIPS) return true;
+    const fips_proc_file = '/proc/crypto/fips';
+    try {
+        const value = fs.readFileSync(fips_proc_file, 'utf8').trim();
+        console.log(`detect_fips_mode: found ${fips_proc_file} with value ${value}`);
+        return value === '1';
+    } catch (err) {
+        if (err.code !== 'ENOENT') {
+            console.warn(`detect_fips_mode: failed to read ${fips_proc_file}:`, err);
```
liranmauda, line 75: This will happen in most cases as centos:8 which the upstream based on have no /proc/crypto/fips 
We can check if the file even exist and console log it
What do you think?

nimrod-becker:
Don't forget
```
const ZERO_SIZE_ETAG = 'd41d8cd98f00b204e9800998ecf8427e';
```
in object_server

guymguym:
> Don't forget
> 
> ```
> const ZERO_SIZE_ETAG = 'd41d8cd98f00b204e9800998ecf8427e';
> ```
> 
> in object_server

no need anymore, monkey-patching works for that as well.

src/util/fips.js
guymguym: resolved by excluding the print on ENOTDIR.

nimrod-becker (approved):

PR #6038, opened 2020-05-30
Website redirection fix

jeniawhite:
### Explain the changes
1. Converted the protocol lowercase (in case we pull from schema enum)
2. Added ':' after the protocol.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -189,8 +189,8 @@ async function _get_redirection_bucket(req, bucket) {
     const redirect = bucket_website_info.website_configuration.redirect_all_requests_to;
     if (redirect) {
         const dest = redirect.host_name;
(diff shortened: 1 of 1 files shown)


src/endpoint/s3/s3_rest.js
```diff
@@ -190,7 +190,7 @@ async function _get_redirection_bucket(req, bucket) {
     if (redirect) {
         const dest = redirect.host_name;
         const protocol = redirect.protocol || req.secure ? 'https' : 'http';
```
guymguym, line 192: Better add parenthesis to separate between two possible operator precedence options:
```
        const protocol = (redirect.protocol || req.secure) ? 'https' : 'http';
```
vs.
```
        const protocol = redirect.protocol || (req.secure ? 'https' : 'http');
```

nimrod-becker (approved):

PR #6047, opened 2020-06-10
exposing storage_limit for cloud_pools

jackyalbo:
### Explain the changes
1. storage_limit which used to apply for different type of pools will now be exposed for cloud pools as well using the pool_api
2. it will limit the use of the cloud storage to a given storage_size - mainly for the use of tiering

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -301,6 +301,13 @@ class Agent {
         await this.block_store.init();
     }
 
+    update_storage_limit(storage_limit) {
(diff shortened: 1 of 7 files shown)


nimrod-becker:
@jackyalbo lets add an update RPC as well ? We can check that the size requested is not lower than the current set or current used and then pass it

dannyzaken (approved):

PR #6053, opened 2020-06-17
Do not allow new buckets on internal on downstream

jeniawhite:
### Explain the changes
1. We are not interested in using internal on downstream
2. Creation of new buckets will be blocked until there is an OPTIMAL pool that is not internal

1. 

### Testing Instructions:
1.

### DIFF
--- a/config.js
+++ b/config.js
@@ -490,3 +490,6 @@ config.INLINE_MAX_SIZE = 4096;
 ///////////////////////////////
 
 config.WORM_ENABLED = false;
+
+// Should we allow the creation of buckets on internal storage
+config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -43,6 +43,8 @@ const EXTERNAL_BUCKET_LIST_TO = 30 * 1000; //30s
 
 const trigger_properties = ['event_name', 'object_prefix', 'object_suffix'];
 
+let optimal_pool_existed = false;
+
 function new_bucket_defaults(name, system_id, tiering_policy_id, owner_account_id, tag, lock_enabled) {
     let now = Date.now();
     return {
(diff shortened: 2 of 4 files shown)


config.js
```diff
@@ -490,3 +490,10 @@ config.INLINE_MAX_SIZE = 4096;
 ///////////////////////////////
 
 config.WORM_ENABLED = false;
+
+/////////////////////////////////
+// DOWNSTREAM/UPSTREAM RELATED //
+/////////////////////////////////
+
+// We should not allow creation of buckets on internal storage in downstream 
+config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
```
nimrod-becker, line 499:
```suggestion
//Should we allow the creation of buckets on internal storage
config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
```

src/server/system_services/bucket_server.js
```diff
@@ -201,6 +202,13 @@ async function create_bucket(req) {
     return get_bucket_info({ bucket: created_bucket });
 }
 
+async function validate_pool_constraints() {
+    if (config.ALLOW_BUCKET_CREATE_ON_INTERNAL !== true) {
```
nimrod-becker, line 206: Sounds harsh for every create bucket, maybe calculate first time and as long as non optimal. 
Once optimal returned a "cached" reply of true ?

jeniawhite: It is possible, but the risk is that we will have invalid pools and allow the creation of buckets.

nimrod-becker: What's the scenario in which we will allow with an invalid pool ?

jeniawhite: In case we had an OPTIMAL pool and it got broken (never comes back to OPTIMAL).
We will create buckets that will never be attached to a valid pool (until someone creates one).

nimrod-becker: that's the same as we had before with the infinite loop once the conditions was satisfied once.

jeniawhite: Yeah, I'm not saying that I've added something bad. I'm just saying that in theory, we have the same consequences as previously.

nimrod-becker (approved):

PR #6055, opened 2020-06-19
arch != x64: rebase cm256, no -msse4.1, no isa-l/erasure_code

guymguym:
CC @liranmauda @yselkowitz @SaravanaStorageNetwork

### Explain the changes
1. Update g256 code from upstream cm256
2. Added condition to only use -msse4.1 when arch=x64
3. Skip isa-l/erasure_code compile on non x64

1. None

### Testing Instructions:
1. Build linux x64
2. Build s390x

### DIFF
--- a/src/deploy/NVA_build/builder.Dockerfile
+++ b/src/deploy/NVA_build/builder.Dockerfile
@@ -10,7 +10,7 @@ LABEL maintainer="Liran Mauda (lmauda@redhat.com)"
 ENV container docker
 RUN dnf update -y -q && \
     dnf clean all
-RUN dnf install -y -q wget unzip which vim python3 && \
+RUN dnf install -y -q wget unzip which vim python2 python3 && \
     dnf --enablerepo=PowerTools install -y -q yasm && \
     dnf group install -y -q "Development Tools" && \
     dnf clean all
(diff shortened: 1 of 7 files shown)


yselkowitz:
This isn't enough for Power or Z, a few additional changes are needed.   I would like to suggest breaking this up:

1. Conditionalize -msse4.1
2. Rebase cm256
3. Add additional code for P&Z

guymguym:
@yselkowitz So 1 & 2 are handled here. Just item 3 is missing?

guymguym:
Added a condition to skip isa-l/erasure_code compile on non x64.
Refer to this stale issue for actual isa-l support for PPC - https://github.com/intel/isa-l/issues/7

liranmauda (approved):

nimrod-becker (approved):

mykaul:
Shouldn't we compile with AVX2?

guymguym:
@mykaul ISA-L doesn't support s390x arch yet. See the linked issue above.

mykaul:
I thought about compiling with AVX2 for X64. Did we remove the per-platform optimizations altogether now?

guymguym:
@mykaul Ok so this PR is only for **non** x86_64.
On x64, ISA-L is compiled with all optimizations and it can check for supported cpu features on runtime (sse, avx, avx2, avx512). 

Specifically for s390x there is no optimized Instructions support in ISA-L, and for powerpc the support is there but we haven’t tried that yet.

Is that clearer?

mykaul:
Sure, thanks.

PR #6057, opened 2020-06-22
Add a "none" option to chunk code config fields

### LINKED ISSUE #5934: create_bucket api should allow disabling encrypt/compress/digest
### Environment info
- NooBaa Version: **5.3**
- Platform: **NA**

### Actual behavior
1. Currently the API does not allow to disable the chunk_coder_config values and passing undefined is falling back to the defaults defined in config.js

See:

https://github.com/noobaa/noobaa-core/blob/d284f630eccce910d2b7d86bedafff1937f7a8e0/src/server/utils/chunk_config_utils.js#L7-L31

https://github.com/noobaa/noobaa-core/blob/d284f630eccce910d2b7d86bedafff1937f7a8e0/src/api/common_api.js#L600-L613

### Expected behavior
1. We should consider either to refactor the semantics of undefined to mean "none" and not override it with a default, but this might require fixing code that did not supply all options assuming it will use the defaults.
2. We can use an empty string value to disable, that might not require as many code changes as most checks are like `!digest_type` which will work right also for empty string.
3. We can use a `'none'` string value to disable, but that will require identifying this new enum value in the code.

### Steps to reproduce
1. NA

jackyalbo:
### Explain the changes
1. digest_type, compress_type, and cipher_type will now have a 'none' option - meaning no digest/compression or encryption is asked for chunks in this tier 

1. Fixed #5934 

### Testing Instructions:
1. Set a 'none' on the above fields, for example: 
```
nb.api.tier.update_tier({name: "bvl#kbqlne8r", chunk_coder_config: { compress_type: "none", cipher_type: "none"}})
```
check in DB a new ccc is created with all the above fields not set, but default do: 
```
{ "_id" : ObjectId("5ef0c44bc226db0024923b93"), "system" : ObjectId("5eef0ddf65636b0029b918e2"), "chunk_coder_config" : { "digest_type" : "sha384", "frag_digest_type" : "sha1", "replicas" : 1, "data_frags" : 1, "parity_frags" : 0 }, "last_update" : 1592837195393 }
```

### DIFF
--- a/src/server/utils/chunk_config_utils.js
+++ b/src/server/utils/chunk_config_utils.js
@@ -12,7 +12,8 @@ function new_chunk_code_config_defaults(chunk_coder_config) {
             cipher_type: config.CHUNK_CODER_CIPHER_TYPE,
             ...chunk_coder_config
         },
-        _.isUndefined);
+        // omit entries with undefined or 'none' values to disable features
+        val => val === undefined || val === 'none');
 
     if (ccc.parity_frags) {
         // Erasure Codes
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -599,17 +599,17 @@ module.exports = {
 
         digest_type: {
             type: 'string',
-            enum: ['sha1', 'sha256', 'sha384', 'sha512']
+            enum: ['sha1', 'sha256', 'sha384', 'sha512', 'none']
         },
 
         compress_type: {
             type: 'string',
-            enum: ['snappy', 'zlib']
+            enum: ['snappy', 'zlib', 'none']
         },
 
         cipher_type: {
             type: 'string',
-            enum: ['aes-256-gcm']
+            enum: ['aes-256-gcm', 'none']
         },
 
         parity_type: {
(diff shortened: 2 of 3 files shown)


src/server/utils/chunk_config_utils.js
```diff
@@ -5,14 +5,16 @@ const _ = require('lodash');
...
+    const ccc = _.omitBy(
+        _.omitBy({
+                digest_type: config.CHUNK_CODER_DIGEST_TYPE,
+                frag_digest_type: config.CHUNK_CODER_FRAG_DIGEST_TYPE,
+                compress_type: config.CHUNK_CODER_COMPRESS_TYPE,
+                cipher_type: config.CHUNK_CODER_CIPHER_TYPE,
+                ...chunk_coder_config
+            },
+            _.isUndefined), // set default values when none provided
+        val => val === 'none'); // remove values set to none
```
guymguym, line 17: It looks weird to me to see these two omitBy.
How about merging them to one?
```
    const ccc = _.omitBy({

        // list of default values which will apply for undefined keys
        digest_type: config.CHUNK_CODER_DIGEST_TYPE,
        frag_digest_type: config.CHUNK_CODER_FRAG_DIGEST_TYPE,
        compress_type: config.CHUNK_CODER_COMPRESS_TYPE,
        cipher_type: config.CHUNK_CODER_CIPHER_TYPE,

        // apply the provided config ('none' values will overwrite default values)
        ...chunk_coder_config

        // omit entries with undefined or 'none' values to disable features
    }, v => v === undefined || v === 'none');
```

guymguym (approved):
See a the small comment above, but looks good. 
Let's rebase and merge.

PR #6058, opened 2020-06-23
Namespace Gaps - Azure Copy part implementation

romayalon:
### Explain the changes
1. Added implementation of copy part.

1. When copying without specifying range of bytes and copy source is larger than 100 MB 
Blob return error unlike AWS.

### Testing Instructions:
1.

### DIFF
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -377,35 +377,57 @@ class NamespaceBlob {
         // generating block ids in the form: '95342c3f-000005'
         // this is needed mostly since azure requires all the block_ids to have the same fixed length
         const block_id = this.blob.getBlockId(params.obj_id, params.num);
+        let res;
         dbg.log0('NamespaceBlob.upload_multipart:',
             this.container,
             inspect(_.omit(params, 'source_stream')),
             'block_id', block_id
         );
         if (params.copy_source) {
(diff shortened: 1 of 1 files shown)


src/sdk/namespace_blob.js
```diff
@@ -377,35 +377,67 @@ class NamespaceBlob {
...
-        );
+            const start = params.copy_source.ranges ? params.copy_source.ranges[0].start : 0;
+            const end = params.copy_source.ranges ? params.copy_source.ranges[0].end : params.size;
+
+            const maxCopySource = 100 * 1024 * 1024;
+            if (end - start > maxCopySource) {
+                throw new Error('The specified copy source is larger than the maximum allowable size for a copy source (100mb).');
+            }
+
+            let expiryDate = new Date();
```
jackyalbo, line 396: Maybe we can move this logic to cloud_utils, it seems we already used same logic in  _get_shared_access_signature in block_store_azure

jackyalbo: Also maybe take some time before current time like in that function - to be on the safe side.

jackyalbo (commented):
Just look most of this is for creating the token, I think it can be valuable to get it out for the future.

jackyalbo (approved):

PR #5977, opened 2020-04-07
NS-CACHE: WIP 1

guymguym:
### Explain the changes
1. Initial implementation of a cache bucket.
2. This PR is focused only on the namespace_cache module - which is how we combine the cache and hub to a single namespace.
3. Upcoming PRs will handle ranged-reads and eviction.

### Issues: Fixed #xxx / Gap #xxx
1. Fixes #5714

### Testing Instructions:
1. Currently quite manual - need to set the env `NAMESPACE_CACHE_DEV=true` for the endpoints, which makes every "namespace bucket" use the namespace_cache module. The write namespace is used as hub, and the bucket's tiering policy is used as cache storage.

PR #6060, opened 2020-06-29
fixing issue with agents' rpc

jackyalbo:
### Explain the changes
1. After the last changes in  PR #5977 we fixed a issue that was covering on a issue of getting server address without port
2. Now we added a new function that will check if we get a null port when the protocol is a well known one and returns the protocol official port:
ws/http - port 80
wss/http - port 443 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -70,7 +70,7 @@ function client_factory_from_schema(schema) {
 
     for (const api of Object.values(schema)) {
         if (!api || !api.id || api.id[0] === '_') {
-             continue;
+            continue;
         }
 
         // Skip common api and other apis that do not define methods.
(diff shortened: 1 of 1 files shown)


nimrod-becker (approved):

src/api/api.js
```diff
@@ -115,9 +115,31 @@ function client_factory_from_schema(schema) {
...
+        }
+        case 'https:':
+        case 'wss:': {
+            return '443';
+        }
+        default: {
+            return '';
+        }
+    }
+}
```
nb-ohad, line 132: I think a more clear solution is to return the actual port number and not a string representation and `undefined` for the default case. (just `reutrn;`)
so line 142 could be written as:
```
const ports = get_default_ports(Number(port || get_protocol_port(protocol));
```

jackyalbo: you mean ```const ports = get_default_ports(Number(port) || get_protocol_port(protocol));```
Agree it much cleaner - So as it already merged - will be part of my next PR

PR #6064, opened 2020-07-01
Fixing issue with removing tiering policies of wrong buckets

jackyalbo:
### Explain the changes
1. add to system_store check if id is of type object_id for both update and delete - otherwise it will throw exception
2. changed bucket reclaimer to use one make_changes for each run - instead of calling bucket_server multiple times

1.  Fixed BZ1839117

### Testing Instructions:
1. write multiple files in multiple buckets
2. use delete_bucket_and_objects from bucket_api to delete some of the buckets
3. see the buckets deleted successfully 
4. No log errors and No undeleted buckets with deleted tiering_policy

### DIFF
--- a/src/server/bg_services/buckets_reclaimer.js
+++ b/src/server/bg_services/buckets_reclaimer.js
@@ -26,6 +26,7 @@ class BucketsReclaimer {
         }
 
         let has_errors = false;
+        this._reset_delete_lists();
         dbg.log0('bucket_reclaimer: starting batch work on buckets: ', deleting_buckets.map(b => b.name).join(', '));
         await P.all(deleting_buckets.map(async bucket => {
             try {
@@ -43,13 +44,7 @@ class BucketsReclaimer {
                 });
                 if (is_empty) {
                     dbg.log0(`bucket ${bucket.name} is empty. calling delete_bucket`);
-                    await this.client.bucket.delete_bucket({ name: bucket.name, internal_call: true }, {
-                        auth_token: auth_server.make_auth_token({
-                            system_id: system._id,
-                            account_id: system.owner._id,
-                            role: 'admin'
-                        })
-                    });
+                    this._add_bucket_to_delete_lists(bucket);
                 } else {
                     dbg.log0(`bucket ${bucket.name} is not empty yet`);
                 }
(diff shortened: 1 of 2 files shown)


src/server/system_services/system_store.js
```diff
@@ -738,6 +740,7 @@ class SystemStore extends EventEmitter {
         _.each(changes.remove, (list, name) => {
             get_collection(name);
             _.each(list, id => {
+                if (!mongo_utils.is_object_id(id)) throw new Error(`SystemStore: make_changes id is not of type object_id: ${id}`);
```
jeniawhite, line 743: @jackyalbo 
What about other ops besides `changes.remove`?
We have the same thing in `insert`, `update`, `db_delete`.
Shouldn't we have the same handling there?

src/server/bg_services/buckets_reclaimer.js
```diff
@@ -58,6 +53,23 @@ class BucketsReclaimer {
...
+        if (this.buckets_to_delete.length) {
+            try {
+                await system_store.make_changes({
+                    remove: {
+                        buckets: this.buckets_to_delete,
+                        tieringpolicies: this.tiering_to_delete,
+                        tiers: this.tiers_to_delete,
+                    },
+                    update: {
+                        accounts: this.account_updates,
```
jeniawhite, line 65: @jackyalbo 
If there are no implicit permissions to the `buckets_to_delete`, you will get an empty array and send an empty array to the make_changes.

src/server/bg_services/buckets_reclaimer.js
```diff
@@ -83,6 +95,29 @@ class BucketsReclaimer {
...
     }
 
+    _add_bucket_to_delete_lists(bucket) {
+        this.buckets_to_delete.push(bucket._id);
+        const tiering_policy = bucket.tiering;
+        this.tiering_to_delete.push(tiering_policy._id);
+        this.tiers_to_delete.push(...tiering_policy.tiers.map(t => t.tier._id));
+        system_store.data.accounts.forEach(account => {
+            if (!account.allowed_buckets || (account.allowed_buckets && account.allowed_buckets.full_permission)) return;
+            this.account_updates.push({
```
jeniawhite, line 105: @jackyalbo 
Unnessasery and heavy update query on accounts collection.
In worst-case scenario, it will trigger an update to all accounts that do not have `full_permission` even if they do not have the bucket in their `permissions_list`.
There is no reason to do this heavy lifting in Mongo, you already have all the data in the cache.
Just code another _.filter that will filter only the accounts that have the implicit permissions to the deleted buckets and then trigger the updates on these accounts.

src/server/system_services/system_store.js
jackyalbo: @jeniawhite 
changes.update have it - you can take a look again.
insert will fail without id as it part of schema validation
db_delete I will add

src/server/bg_services/buckets_reclaimer.js
jackyalbo: removed the account without the bucket in their permission list from acconut_updates

src/server/bg_services/buckets_reclaimer.js
jackyalbo: now checking if account_updates is not empty before sending the update

jeniawhite (approved):

PR #6066, opened 2020-07-02
Fix issue with system store concurrent loads, reads, and writes

nb-ohad:
### Explain the changes
- Add a lock on the load semaphore during make_changes.
  - Fixing concurrent loads and reads in the same process
- Publish the last_update_time for each make_changes so other processes could update their last update time in order to prevent missing some of the changes during the incremental load process. 
  - Fixing concurrent reads and writes between diff processes

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -467,12 +467,20 @@ class SystemStore extends EventEmitter {
         }
     }
 
-    async load() {
+    async load(since) {
         // serializing load requests since we have to run a fresh load after the previous one will finish
         // because it might not see the latest changes if we don't reload right after make_changes.
         return this._load_serial.surround(async () => {
             try {
                 dbg.log3('SystemStore: loading ...');
+
+                // If we get a load request with an timestamp older then our last update time
+                // we ensure we load everyting from that timestamp by updating our last_update_time.
+                if (!_.isUndefined(since) && since < this.last_update_time) {
+                    dbg.log0('SystemStore.load: Got load request with a timestamp older then my last update time');
+                    this.last_update_time = since;
+                }
+
                 let new_data = new SystemStoreData();
                 let millistamp = time_utils.millistamp();
                 await this._register_for_changes();
@@ -658,6 +666,28 @@ class SystemStore extends EventEmitter {
      *
      */
     async make_changes(changes) {
+        const { any_news, last_update } = await this._load_serial.surround(
+            () => this._make_changes_internal(changes)
+        );
+
+        // Reloading must be done outside the semapore lock because the load is
+        // locking on the same semaphore.
+        if (any_news) {
+            if (this.is_standalone) {
+                await this.load(last_update);
+            } else {
+                // notify all the cluster (including myself) to reload
+                await server_rpc.client.redirector.publish_to_cluster({
+                    method_api: 'server_inter_process_api',
+                    method_name: 'load_system_store',
+                    target: '',
+                    request_params: { since: last_update }
+                });
+            }
+        }
+    }
+
+    async _make_changes_internal(changes) {
         const bulk_per_collection = {};
         const now = new Date();
         const last_update = now.getTime();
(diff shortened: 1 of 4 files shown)


src/server/system_services/system_store.js
```diff
@@ -658,125 +666,132 @@ class SystemStore extends EventEmitter {
      *
      */
     async make_changes(changes) {
```
guymguym, line 668: May I suggest that you make this change by adding a wrapping function like this:
```javascript
async make_changes(changes) {
    return this._load_serial.surround(async () => this._make_changes_internal(changes));
}

async _make_changes_internal(changes) {
// ... HERE COMES THE ORIGINAL CODE
```

nimrod-becker: also easier for readability and history

nb-ohad: Ok, but it will make it harder to export all the values we need for the publishing load requests to the cluster.

guymguym: why not?

guymguym: ah I see because you publish outside the semaphore

guymguym: OK, still worth it I think that the internal method will return what's needed

nb-ohad: done

src/server/system_services/system_store.js
```diff
@@ -658,125 +666,132 @@ class SystemStore extends EventEmitter {
...
-        };
-        const get_bulk = name => {
-            const bulk = bulk_per_collection[name] ||
-                mongo_client.instance().collection(name).initializeUnorderedBulkOp();
-            bulk_per_collection[name] = bulk;
-            return bulk;
-        };
+        let last_update;
+
+        await this._load_serial.surround(async () => {
```
jeniawhite: This means that we won’t have parallel make_changes and refreshes/loads will hold up make_changes.
This could mean a slow down in the performance of the system.
Just making sure that we truly need that surround, since loads/refreshes affect it.

guymguym: @jeniawhite That's right. 
Where do you think this has an impact? 
md_aggregator updates? 
Somewhere else?

nb-ohad: @jeniawhite Yes, this was the intention

src/server/system_services/system_store.js
guymguym: Perhaps you can move back make_changes_in_background back to where it was? 
I share your desire to organize the code for reading, but we should minimize it to larger refactors.

src/server/system_services/system_store.js
jeniawhite: @guymguym I would assume multiple part uploads and mainly object update flows.
Did not deep dive and check it though.

nb-ohad: 1. concurrent make_change might update the same documents which will lose information.
2. loading while writing is also problematic regarding the validity of the data that we try to read, especially when we do differential reads

nb-ohad: @jeniawhite we are talking about system store updates only

jeniawhite: @nb-ohad 
You are right regarding system_store updates. Object flow irrelevant.

src/server/system_services/system_store.js
nb-ohad: @guymguym this means to put this new method at the end of the file or else something will move

guymguym: no I meant just to move make_changes_in_background back to where it was to avoid a change on lines that are not required to change

nb-ohad: done

jeniawhite (approved):

nimrod-becker:
Given the comments and fixes already made, I'm approving, Ohad can you just put the original func (_background) to its place and and we can merge

nimrod-becker (approved):

jackyalbo (approved):
LGTM

PR #6068, opened 2020-07-02
Add and expose the system state via API  (currently only initialization states)

nb-ohad:
### Explain the changes
1. Add state field to system data
2. Trace the current state of the system and update the field
3. allow querying the state of the system via an API call

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/system_api.js
+++ b/src/api/system_api.js
@@ -14,6 +14,31 @@ module.exports = {
     id: 'system_api',
 
     methods: {
+        get_system_status: {
+            doc: 'Query for the status of a system',
+            method: 'GET',
+            reply: {
+                type: 'object',
+                required: ['state', 'last_state_change'],
+                properties: {
+                    state: {
+                        type: 'string',
+                        enum: [
+                            'DOES_NOT_EXIST',
+                            'INITIALIZING',
+                            'COULD_NOT_INITIALIZE',
+                            'READY',
+                        ]
+                    },
+                    last_state_change: { idate: true }
+                }
+            },
+            auth: {
+                account: false,
+                system: false,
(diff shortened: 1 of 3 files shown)


src/server/system_services/system_server.js
```diff
@@ -142,6 +142,10 @@ function new_system_defaults(name, owner_account_id) {
         _id: system_store.new_system_store_id(),
         name: name,
         owner: owner_account_id,
+        state: {
```
nimrod-becker, line 145: why do we need it ?

src/api/system_api.js
```diff
@@ -14,6 +14,31 @@ module.exports = {
...
+                    state: {
+                        type: 'string',
+                        enum: [
+                            'DOES_NOT_EXIST',
+                            'INITIALIZING',
+                            'COULD_NOT_INITIALIZE',
+                            'READY',
+                        ]
+                    },
+                    last_state_change: { idate: true }
```
nimrod-becker, line 33: why do we need it ?

src/server/system_services/system_server.js
nb-ohad: I am not sure what do you mean. This field holds the status of the system

src/api/system_api.js
nb-ohad: This will enable us to take action in case of long phases, for example we could find out that a system  is stuck on initializing state in the operator and we might decide to undo the creation

src/server/system_services/system_server.js
nimrod-becker: the timestamp

nimrod-becker (approved):

PR #6069, opened 2020-07-02
Adding bucket_claim_owner support for accounts created by operator

jackyalbo:
### Explain the changes
1. Introducing support to bucket_claim_owner for account created by the noobaa operator as part of the obc flow
2. adapting both account api and schema to support the bucket_claim_owner property
3. if an account has bucket_claim_owner in it - this account will get FC access to the bucket set in this property and will act as another bucket owner to the operator account itself
4. Only the operator account could set such a property on a new account - for other request such a property will be ignored

1. Partially Fixes BZ#1849771

### Testing Instructions:
1.

### DIFF
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -49,6 +49,9 @@ module.exports = {
                             enum: ['admin', 'user', 'viewer', 'operator']
                         }
                     },
+                    bucket_claim_owner: {
+                        $ref: 'common_api#/definitions/bucket_name'
+                    },
                     //Special handling for the first account created with create_system
                     new_system_parameters: {
                         type: 'object',
@@ -561,6 +564,9 @@ module.exports = {
                 },
                 default_pool: {
                     type: 'string',
+                },
+                bucket_claim_owner: {
+                    $ref: 'common_api#/definitions/bucket_name',
                 },
                 external_connections: {
                     type: 'object',
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -315,8 +315,9 @@ async function authorize_request_policy(req) {
         const method = _get_method_from_req(req);
         const account = await req.object_sdk.rpc_client.account.read_account({});
         // system owner by design can always change bucket policy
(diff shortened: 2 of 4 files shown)


guymguym (commented):
Wait, why not just update the owner from the operator to the new account after it gets created?

jackyalbo:
> Wait, why not just update the owner from the operator to the new account after it gets created?

It was a requirement for this task, as the obc user can be deleted/changed. I think @nimrod-becker can explain better

nimrod-becker:
Were not sure how it will behave in RETAIN cases when a new OBC will be created, also had a technical issue regarding the operator flow. So we said we'd give an obc owner effective permissions as owner

dannyzaken:
I read the BZ but stull I don't really understand the need for this change. what are we trying to get that is missing today? 
right now the operator creates a new account with S3 access only to the requested bucket (which is either new or existing). why should this account be treated as an owner? what about the case that the OBC is connecting to an existing bucket?

nimrod-becker:
@dannyzaken When using bucket policies, the bucket owner differs from other users. There has been a request that we would change the owner of an OBC created bucket to be the account that created it (makes sense, its how it behaves in AWS and in our S3 implementation as well...). The example given was to put a policy that blocks all other users access besides the owner.

Changing the owner itself proved to introduce more issues, so we decided to have the account created the obc as an owner without actually changing the db.

dannyzaken:
@nimrod-becker we just need to be aware of the scenario where the OBC reference an existing bucket. in that case, any OBC account that gets access to the bucket also gets an owner like permissions

nimrod-becker:
What are we missing here to merge ?

jackyalbo:
> What are we missing here to merge ?

We need a decision that this solution is acceptable, and if so someone to approve this PR. Then we can go a head and PR the operator side.

nimrod-becker:
@guymguym @dannyzaken 
I'm acking this, do you see a reason not to?

@jackyalbo lets rebase

nimrod-becker (approved):

PR #6073, opened 2020-07-05
deploy_minikube.sh: Adding support for Centos8

liranmauda:
### Explain the changes
deploy_minikube.sh:
- Adding support for Centos8
- Moving minikube and kubectl from /usr/local/bin to /usr/bin

### DIFF
--- a/.travis/deploy_minikube.sh
+++ b/.travis/deploy_minikube.sh
@@ -14,34 +14,67 @@ set -x
 # NOTE: This script was originally copied from the Cojaeger-operator build
 # https://github.com/jaegertracing/jaeger-operator/blob/master/.travis/setupMinikube.sh
 
-# socat is needed for port forwarding
-sudo apt-get update && sudo apt-get install socat && sudo apt-get install conntrack
-
 export MINIKUBE_VERSION=v1.8.2
 export KUBERNETES_VERSION=v1.17.3
 
-sudo mount --make-rshared /
-sudo mount --make-rshared /proc
-sudo mount --make-rshared /sys
+source /etc/os-release
+
+if [ "${ID}" == "ubuntu" ]
+then 
+    # socat is needed for port forwarding
+    sudo apt-get update && sudo apt-get install socat && sudo apt-get install conntrack
+    sudo mount --make-rshared /
+    sudo mount --make-rshared /proc
+    sudo mount --make-rshared /sys
+elif [ "${ID}" == "centos" ]
+then
+    dnf install -y sudo
+    sudo dnf -y update && sudo dnf -y install socat conntrack 
+    dnf install -y dnf-plugins-core
+    sudo dnf config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+    sudo dnf install -y docker-ce docker-ce-cli containerd.io --nobest
+    #disable SELinux
+    SELinux_status=$(sestatus | grep "SELinux status" | awk -F ":" '{print $2}' | xargs )
+    if [ "${SELinux_status}" == "enabled" ]
+    then
+        sudo setenforce 0
+    fi
+else
+    echo "${ID} is not supported for ${0}, Exiting."
+    exit 1
+fi
 
 curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/${KUBERNETES_VERSION}/bin/linux/amd64/kubectl && \
     chmod +x kubectl &&  \
-    sudo mv kubectl /usr/local/bin/
+    #moving kubectl to /usr/bin/ and not /usr/local/bin/ because on Centos it fails to sudo from /usr/local/bin/
+    # it fails to sudo from /usr/local/bin/ even if it is the ${PATH}
+    sudo mv kubectl /usr/bin/
 
 curl -Lo minikube https://storage.googleapis.com/minikube/releases/${MINIKUBE_VERSION}/minikube-linux-amd64 && \
     chmod +x minikube &&  \
-    sudo mv minikube /usr/local/bin/minikube
+    #moving minikube to /usr/bin/ and not /usr/local/bin/ because on Centos it fails to sudo from /usr/local/bin/
+    # it fails to sudo from /usr/local/bin/ even if it is the ${PATH}
+    sudo mv minikube /usr/bin/minikube
 
 mkdir "${HOME}"/.kube || true
 touch "${HOME}"/.kube/config
 
 # minikube config
+minikube config set WantUpdateNotification false
 minikube config set WantNoneDriverWarning false
 minikube config set vm-driver none
 
+cat ~/.minikube/config/config.json
+
 minikube version
-sudo minikube start --kubernetes-version=$KUBERNETES_VERSION #--insecure-registry="${LOCAL_IP}:5000" #TODO Remove insecure
-sudo chown -R travis: /home/travis/.minikube/
+#sudo minikube start --kubernetes-version=$KUBERNETES_VERSION #--insecure-registry="${LOCAL_IP}:5000" #TODO Remove insecure
+sudo minikube start --kubernetes-version=$KUBERNETES_VERSION
+
+current_user=$(whoami)
+if [ "${current_user}" != "root"  ]
+then
+    sudo chown -R ${current_user}: /home/${current_user}/.minikube/
+fi
 
 minikube update-context
 


.travis/deploy_minikube.sh
```diff
@@ -14,15 +14,35 @@ set -x
...
+    sudo mount --make-rshared /
+    sudo mount --make-rshared /proc
+    sudo mount --make-rshared /sys
+elif [ "${ID}" == "centos" ]
+then
+    dnf install -y sudo
+    sudo dnf -y update && sudo dnf -y install socat conntrack 
+    dnf install -y dnf-plugins-core
+    sudo dnf config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+    sudo dnf install -y docker-ce docker-ce-cli containerd.io --nobest
```
mykaul, line 35: why docker-ce and not podman and/or buildah?

.travis/deploy_minikube.sh
```diff
@@ -14,15 +14,35 @@ set -x
...
+    #disable SELinux
+    SELinux_status=$(sestatus | grep "SELinux status" | awk -F ":" '{print $2}' | xargs )
+    if [ "${SELinux_status}" == "enabled" ]
+    then
+        sudo setenforce 0
```
mykaul, line 40: oy :(

Why?

.travis/deploy_minikube.sh
liranmauda: originally it was on deploying Ubuntu (as it for travis runs).
We can change to podman when running on Centos

.travis/deploy_minikube.sh
liranmauda: Currently, kubeadm on CentOS needs SELinux to be disabled:

https://github.com/kubernetes/minikube/issues/6014#issuecomment-563367526

mykaul: Does that happen with podman as well?

mykaul: (I mean, somehow, magically it is working in RHEL, no?)

.travis/deploy_minikube.sh
liranmauda: By the look of it, minikube support for podman is still WIP.
We can see 2 things:
1. Using the podman driver is experimental
2. even though the driver is podman, the run-time is docker and changing it to cri-o still looks for docker (https://github.com/kubernetes/minikube/issues/6795).

We are using MINIKUBE_VERSION=v1.8.2 and KUBERNETES_VERSION=v1.17.3 due to https://github.com/kubernetes/minikube/issues/7828 

running with podman results in:
```
+ main@./1.sh:70 cat /root/.minikube/config/config.json
{
    "WantNoneDriverWarning": false,
    "WantUpdateNotification": false,
    "container-runtime": "cri-o",
    "driver": "podman",
    "vm-driver": "none"
}+ main@./1.sh:72 minikube version
minikube version: v1.8.2
commit: eb13446e786c9ef70cb0a9f85a633194e62396a1
+ main@./1.sh:74 minikube start --kubernetes-version=v1.17.3
😄  minikube v1.8.2 on Centos 8.1.1911
    ▪ MINIKUBE_VERSION=v1.8.2
✨  Using the podman (experimental) driver based on user configuration
E0705 15:30:53.797867   10377 cache.go:106] Error downloading kic artifacts:  error loading image: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```
once podman support in minikube will be better, and we will be able to upgrade the minikube version we can switch to podman on Centos8.

liranmauda: I have opened a new issue to track this: https://github.com/noobaa/noobaa-core/issues/6075

.travis/deploy_minikube.sh
liranmauda: We should look into it one we can switch into podman (see previous comment https://github.com/noobaa/noobaa-core/pull/6073#discussion_r449873237)

liranmauda:
Adding dbg to the subject to debug the failed deploy on travis

liranmauda:
@nixpanic @mykaul can you have a look?

.travis/deploy_minikube.sh
nixpanic: I have seen similar problems with minikube+podman, it does not really work well yet.

.travis/deploy_minikube.sh
nixpanic: There are some locations where minikube stores executables, and these try to access configuration files. SElinux prevents that. I am not sure if the location that minikube uses is dynamic, or predictable. In the 2nd case, creating the directories in advance and setting appropriate labels might work.

This obviously need some more research. Moving to Permissive mode is acceptable for the moment.

.travis/deploy_minikube.sh
```diff
@@ -14,34 +14,63 @@ set -x
...
 curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/${KUBERNETES_VERSION}/bin/linux/amd64/kubectl && \
     chmod +x kubectl &&  \
     sudo mv kubectl /usr/local/bin/
 
 curl -Lo minikube https://storage.googleapis.com/minikube/releases/${MINIKUBE_VERSION}/minikube-linux-amd64 && \
     chmod +x minikube &&  \
-    sudo mv minikube /usr/local/bin/minikube
+    #moving minikube to /usr/bin/ and not /usr/local/bin/ because on Centos it fails to sudo from /usr/local/bin/
+    # it fails to sudo from /usr/local/bin/ even if it is the ${PATH}
+    sudo mv minikube /usr/bin/minikube
```
nixpanic, line 55: `sudo` clears the environment and only allows certain environment variables to stay set. You could place them in `/etc/environment`, or indeed just prevent the need to use `/usr/local/bin` :+1:

.travis/deploy_minikube.sh
```diff
@@ -14,34 +14,63 @@ set -x
...
+cat ~/.minikube/config/config.json
+
 minikube version
-sudo minikube start --kubernetes-version=$KUBERNETES_VERSION #--insecure-registry="${LOCAL_IP}:5000" #TODO Remove insecure
-sudo chown -R travis: /home/travis/.minikube/
+#sudo minikube start --kubernetes-version=$KUBERNETES_VERSION #--insecure-registry="${LOCAL_IP}:5000" #TODO Remove insecure
+sudo minikube start --kubernetes-version=$KUBERNETES_VERSION
+if [ -d /home/travis/ ]
+then
+    sudo chown -R travis: /home/travis/.minikube/
```
nixpanic, line 72: This script may run as non-root and non-travis? In that case you will need to `chown` the contents as the current user too.

This is not an issue for the CentOS CI (as you would normally run everything as root by default), but you may want to add it for completeness

.travis/deploy_minikube.sh
```diff
@@ -14,34 +14,63 @@ set -x
...
+fi
 
 minikube update-context
```
nixpanic, line 75: This command seems to execute `kubectl`:
```
+ main@./deploy_minikube.sh:75 minikube update-context
🙄  minikube IP was already correctly configured for 192.168.121.2
+ main@./deploy_minikube.sh:78 kubectl cluster-info
./deploy_minikube.sh: line 78: kubectl: command not found
```
It is placed in `/usr/local/bin` on line 49.

nixpanic (changes requested):
You should probably put kubectl in `/usr/bin` like minikube.

When running in a Vagrant VM (as user 'vagrant'), `kubectl` does not work:
```
[vagrant@localhost vagrant]$ kubectl cluster-info

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
The connection to the server localhost:8080 was refused - did you specify the right host or port?
[vagrant@localhost vagrant]$ sudo /usr/local/bin/kubectl cluster-info
Kubernetes master is running at https://192.168.121.2:8443
KubeDNS is running at https://192.168.121.2:8443/api/v1/namespaces/kube-system/services/kube-dns:dns/proxy

To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
```
There is a `/root/.kube` directory, but not one in `/home/vagrant`.

This is does not need to be blocker in case everything is intended to be run as root.

.travis/deploy_minikube.sh
liranmauda: If we don't have root access and no sudo installed we will fail in:
```
+ main@./1.sh:31 dnf install -y sudo
Error: This command has to be run under the root user.
```
Even if we have sudo it will ask for root password:
```
+ main@./1.sh:32 sudo dnf -y update

We trust you have received the usual lecture from the local System
Administrator. It usually boils down to these three things:

    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.

[sudo] password for liranm:
```
for now, we will assume we are running as root

liranmauda: also chown -R to a user will require:
1. password 
2. that the user will be in the `/etc/sudoers` file
```
+ sudo chown -R liranm: /home/liranm/.minikube/
[sudo] password for liranm:
liranm is not in the sudoers file.  This incident will be reported.
```

nixpanic (commented):
> for now, we will assume we are running as root

with that being the case, things look good, but you'll have to place `kubectl` in a location where `sudo kubectl` works in CentOS (not `/usr/local/bin`)

.travis/deploy_minikube.sh
```diff
@@ -14,34 +14,67 @@ set -x
...
 minikube version
-sudo minikube start --kubernetes-version=$KUBERNETES_VERSION #--insecure-registry="${LOCAL_IP}:5000" #TODO Remove insecure
-sudo chown -R travis: /home/travis/.minikube/
+#sudo minikube start --kubernetes-version=$KUBERNETES_VERSION #--insecure-registry="${LOCAL_IP}:5000" #TODO Remove insecure
+sudo minikube start --kubernetes-version=$KUBERNETES_VERSION
+
+current_user=$( who | awk '{print $1}')
+if [ "${current_user}" != "root"  ]
+then
+    sudo chown -R ${current_user}: /home/${current_user}/.minikube/
```
nixpanic, line 76: Running the script as root (ssh into the VM as user 'vagrant', then `sudo -i`) causes it to fail here:
```
+ main@./deploy_minikube.sh:73 who
+ main@./deploy_minikube.sh:73 awk '{print $1}'
+ main@./deploy_minikube.sh:73 current_user=vagrant
+ main@./deploy_minikube.sh:74 '[' vagrant '!=' root ']'
+ main@./deploy_minikube.sh:76 sudo chown -R vagrant: /home/vagrant/.minikube/
chown: cannot access '/home/vagrant/.minikube/': No such file or directory
[root@localhost vagrant]# who
vagrant  pts/0        2020-07-06 07:15 (192.168.121.1)
[root@localhost vagrant]# whoami 
root
```

When you use `whoami` to get the `current_user`, things should work better :wink:

nixpanic (approved):
Thanks for the corrections, it works for me in a Vagrant VM :+1:

jackyalbo (approved):

PR #6074, opened 2020-07-05
Namespace gaps - part 2

romayalon:
### Explain the changes
1. Object upload - metadata is missing - 
Added metadata to object upload and threw an error for metadata keys that contains invalid characters (there is a bug in the SDK - it doesn't return the correct error for '-'). catched and translated metadata errors for other special chars.

2. list objects - continuation token is not synced - 
called directly to the API instead of calling the relevant SDK function and by that got a unified continuation token for blobs and directories.

3. Multipart upload - metadata is missing - 
  In initiate multipart upload - if there are xattr - saving the upload details including xattr in the db if it's valid - else we will 
 throw an error.
  In complete multipart upload - reading the upload xattr, setting it to the new obj and aborting the upload in our db.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/endpoint/blob/blob_utils.js
+++ b/src/endpoint/blob/blob_utils.js
@@ -6,6 +6,8 @@ const url = require('url');
 const _ = require('lodash');
 const time_utils = require('../../util/time_utils');
 const endpoint_utils = require('../endpoint_utils');
+const { make_http_request, parse_xml_to_js } = require('../../util/http_utils');
+const { read_stream_join } = require('../../util/buffer_utils');
 
 function set_response_object_md(res, object_md) {
     res.setHeader('ETag', '"' + object_md.etag + '"');
@@ -53,6 +55,60 @@ function parse_copy_source(req) {
 }
 
 
+
+
+//////// BLOB API UTILS ////////
+
+// https://docs.microsoft.com/en-us/rest/api/storageservices/list-blobs
+// This API called directly because of an inefficient matching function of the Azure blob SDK.
+async function list_objects(params, account_name, container, sasToken) {
+
+    const hostname = `${account_name}.blob.core.windows.net`;
+    const delimiter = params.delimiter || '/';
+    const maxResults = params.limit === 0 ? 1 : params.limit;
+
+    let path = `https://${account_name}.blob.core.windows.net/${container}?restype=container&comp=list&delimiter=${delimiter}` +
+        `&maxresults=${maxResults}&${sasToken}`;
+    if (params.key_marker) path += `&marker=${params.key_marker}`;
+
+    let response;
+    try {
+        response = await make_http_request({ hostname, port: 443, path, method: 'GET' });
+    } catch (err) {
+        throw new Error(`GET ${path} did not responed or returned with an error ${err}`);
+    }
+    const status_code = response.statusCode;
+    const buffer = await read_stream_join(response);
+    const body = buffer.toString('utf8');
+    if (status_code !== 200) {
+        throw new Error(`Could not get blobs and diresctories list, (status code: ${status_code}) got ${body}`);
+    }
+
+    let blobs;
+    let dirs;
+    let next_marker;
+    try {
+        const parsed = await parse_xml_to_js(body);
+        blobs = parsed.EnumerationResults.Blobs[0].Blob;
+        dirs = parsed.EnumerationResults.Blobs[0].BlobPrefix;
+        next_marker = parsed.EnumerationResults.NextMarker[0];
+        let parse_blobs = key => {
+            const props = key.Properties[0];
+            let obj = Object.keys(props).reduce((acc, p) => {
+                acc[(_.lowerFirst(_.camelCase(p)))] = props[p][0];
+                return acc;
+            }, { name: key.Name[0] });
+            return obj;
+        };
+        blobs = _.map(blobs, obj => parse_blobs(obj));
+    } catch (err) {
+        throw new Error(`could not parse body ${body} got error ${err}`);
+    }
+    return { blobs, dirs, next_marker };
+}
+
+
+exports.list_objects = list_objects;
 exports.set_response_object_md = set_response_object_md;
 exports.get_request_xattr = get_request_xattr;
 exports.set_response_xattr = set_response_xattr;
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -12,7 +12,7 @@ const blob_utils = require('../endpoint/blob/blob_utils');
 const azure_storage = require('../util/azure_storage_wrap');
 const stream_utils = require('../util/stream_utils');
 const stats_collector = require('./endpoint_stats_collector');
-
+const mongodb = require('mongodb');
 
 const MAP_BLOCK_LIST_TYPE = Object.freeze({
     uncommitted: 'UncommittedBlocks',
@@ -22,12 +22,13 @@ const MAP_BLOCK_LIST_TYPE = Object.freeze({
 
 class NamespaceBlob {
 
-    constructor({ namespace_resource_id, rpc_client, connection_string, container }) {
+    constructor({ namespace_resource_id, rpc_client, connection_string, container, account_name }) {
         this.namespace_resource_id = namespace_resource_id;
         this.connection_string = connection_string;
         this.container = container;
         this.blob = azure_storage.createBlobService(connection_string);
         this.rpc_client = rpc_client;
+        this.account_name = account_name;
     }
 
     is_same_namespace(other) {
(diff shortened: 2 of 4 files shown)


src/util/http_utils.js
```diff
@@ -201,12 +201,16 @@ function read_request_body(req, options) {
     });
 }
 
+function parse_xml_to_js(body, xml_options) {
+    return P.fromCallback(callback => xml2js.parseString(body, xml_options, callback));
+}
```
guymguym, line 206: If we are already changing this code then lets change from Blubird to native promises with util.promisify:
```suggestion
const parse_xml_to_js = util.promisify(xml2js.parseString);
```

src/server/object_services/object_server.js
```diff
@@ -1306,6 +1306,22 @@ function report_endpoint_problems(req) {
     }
 }
 
+/**
+ *
+ * READ_UPLOAD_MD
+ *
+ */
+async function read_upload_md(req) {
+    let obj;
```
guymguym, line 1315: Prefer to use `const obj = ...` as much as possible vs let.

src/server/object_services/object_server.js
```diff
@@ -1306,6 +1306,22 @@ function report_endpoint_problems(req) {
...
+    load_bucket(req);
+    let res = {};
+    obj = await MDStore.instance().find_object_latest(req.bucket._id, req.rpc_params.key, true);
+    if (obj) {
+        res = { obj_id: obj._id, key: obj.key, xattr: obj.xattr };
+    }
+    return res;
```
guymguym, line 1322: Why are we failing silently?
I wouldn't want this to be the normal API semantics, only if there's really a good reason for it.

src/server/object_services/md_store.js
```diff
@@ -176,19 +176,21 @@ class MDStore {
...
-        }, {
-            hint: 'latest_version_index',
-            sort: { bucket: 1, key: 1, version_past: 1 },
-        });
+        };
+        let proj = { sort: { bucket: 1, key: 1, version_past: 1 } };
+        if (!upload_in_progress) {
+            params.upload_started = null;
+            proj.hint = 'latest_version_index';
+        }
```
guymguym, line 192: I don't understand the cases.
I would prefer to separate into two functions, there seems to be more differences than similarities.

src/server/object_services/md_store.js
```diff
@@ -176,19 +176,21 @@ class MDStore {
         return mongo_utils.populate(docs, doc_path, this._objects.col(), fields);
     }
 
-    async find_object_latest(bucket_id, key) {
-        return this._objects.col().findOne({
+    async find_object_latest(bucket_id, key, upload_in_progress = false) {
+        let params = {
```
guymguym, line 180: `const`

src/sdk/object_sdk.js
```diff
@@ -257,7 +257,8 @@ class ObjectSDK {
                 namespace_resource_id: ns_info.id,
                 rpc_client: this.rpc_client,
                 container: ns_info.target_bucket,
-                connection_string: cloud_utils.get_azure_connection_string(ns_info)
+                connection_string: cloud_utils.get_azure_connection_string(ns_info),
+                account_name: ns_info.access_key.unwrap()
```
guymguym, line 261: account_name is access_key ? can we fix the terminology?

src/api/object_api.js
```diff
@@ -535,6 +535,33 @@ module.exports = {
...
+                type: 'object',
+                properties: {
+                    obj_id: { objectid: true },
+                    key: { type: 'string' },
+                    xattr: { $ref: '#/definitions/xattr' }
+                }
+            },
+            auth: {
+                system: ['admin', 'user'],
+                anonymous: true,
```
guymguym, line 561: why is that anonymous? really needed? CC @jeniawhite

src/api/object_api.js
```diff
@@ -535,6 +535,33 @@ module.exports = {
...
+                }
+            },
+            reply: {
+                type: 'object',
+                properties: {
+                    obj_id: { objectid: true },
+                    key: { type: 'string' },
+                    xattr: { $ref: '#/definitions/xattr' }
+                }
+            },
```
guymguym, line 558: Wait, what is this api about? returning an upload from key? but there could be multiple uploads per key.
Can you elaborate a little on the case that requires it and how to handle concurrent/leftover uploads to the same key..

src/sdk/namespace_blob.js
```diff
@@ -48,54 +51,61 @@ class NamespaceBlob {
...
-            P.fromCallback(callback =>
-                this.blob.listBlobDirectoriesSegmentedWithPrefix(
-                    this.container,
-                    params.prefix || null,
-                    request_token,
-                    options,
-                    callback)
-            )
-        );
+        let path = `https://${account_name}.blob.core.windows.net/${this.container}?restype=container&comp=list&delimiter=${delimiter}&maxresults=${maxResults}`;
```
guymguym, line 59: OK so you just have to add some comments to explain what we're doing here and why.
That the Azure storage SDK does not allow it, and provide a reference to the SDK code, etc.
But this is a pile of code which is un-parseable without some comments in free language.
Also I would highly recommend to write this code in a utility module rather than inline in namespace_blob.

guymguym (changes requested):

src/server/object_services/object_server.js
romayalon: ok so we call that function in complete multipart upload (please notice that in init multipart upload we will store the upload md only if params.xattr exist - I did it that way because I wanted to save time and place complexity)  
we are failing silently here because of the fact that we don't know (in complete multipart upload) if there is MD for the object or not.

another solution is to always store the upload md and throw here error when needed instead of {}. do you prefer it ?

guymguym: So just catch the error in complete_multipart_upload() and ignore it if that's what needed.

Also I'm not really following the optimization and flow around xattr - how is xattr related to locating upload md?

src/api/object_api.js
romayalon: deleted from api

src/api/object_api.js
romayalon: deleted from api

src/sdk/namespace_blob.js
romayalon: moved to blob_utils and added comment :)

src/endpoint/blob/blob_utils.js
```diff
@@ -53,6 +55,58 @@ function parse_copy_source(req) {
 }
 
 
+
+
+//////// BLOB API UTILS ////////
+
+async function list_objects(params, account_name, container, sasToken) {
```
guymguym, line 62: missing function doc - why does this exist, where is it taken from, etc.

romayalon: I wrote why it exist in the function call and that it's taken from the API. should I add/move it to this file ?

jeniawhite (approved):

src/endpoint/blob/blob_utils.js
```diff
@@ -53,6 +55,60 @@ function parse_copy_source(req) {
...
+    const body = buffer.toString('utf8');
+    if (status_code !== 200) {
+        throw new Error(`Could not get blobs and diresctories list, (status code: ${status_code}) got ${body}`);
+    }
+
+    let blobs;
+    let dirs;
+    let next_marker;
+    try {
+        const parsed = await parse_xml_to_js(body, undefined);
```
guymguym, line 91: that undefined `options` argument is optional anyway and it is less cluttered to read without it.

src/sdk/namespace_blob.js
```diff
@@ -363,16 +341,40 @@ class NamespaceBlob {
     async create_object_upload(params, object_sdk) {
         // azure blob does not have start upload operation
         // instead starting multipart upload is implicit
-        // TODO how to handle xattr that s3 sends in the create command
-        const obj_id = crypto.randomBytes(24).toString('base64');
+        let obj_id = crypto.randomBytes(30).toString('base64');
```
guymguym, line 344: the entire handling of obj_id & etag in this namespace seems to be a bit shaky.
maybe we should refactor that handling to a function or two that will handle the generation of new ids for upload, the parsing of upload ids for multipart, and translation between existing blobs to s3 etag.

Also a very important thing - we need to make sure that etag that we return is either the real md5 of the data, or otherwise we the etag muat include a dash (`-`) because this is how S3 identifies etags of multipart uploads that are also not md5 of the assembled object. So the point is that returning etag=md5 of the object is the most compatible etag to AWS. After that return an etag with a dash which AWS SDK will identify and exclude from md5 checking. The dash option is less compatible than real md5, but if we don't intend on reading and computing the object md5 then the dash will make the etag more compatible than without one.

guymguym (changes requested):

src/sdk/namespace_blob.js
```diff
@@ -12,7 +12,7 @@ const blob_utils = require('../endpoint/blob/blob_utils');
 const azure_storage = require('../util/azure_storage_wrap');
 const stream_utils = require('../util/stream_utils');
 const stats_collector = require('./endpoint_stats_collector');
-
+const { MDStore } = require('../server/object_services/md_store');
```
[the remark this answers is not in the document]
romayalon: agreed, changed to import mongodb

src/endpoint/blob/blob_utils.js
romayalon: deleted

guymguym (approved):
I would at least suggest to document in code comments the cases and how we handle obj_id/etag.
Looks good!

src/sdk/namespace_blob.js
romayalon: This specific obj_id describes a multipartUpload uploadId - as a fix to this comment I returned the obj_id to be randomBytes(24) as it was before or to be the upload obj_id in a case where there are xattr, this fix simplify the changes in multipartUpload uploadId area.
also I added comments :)

PR #6078, opened 2020-07-06
Fixing issues with default pool and with missing bucket for chunk

jackyalbo:
### Explain the changes
1. handling chunks with no buckets in scrubber - avoiding issues with building chunks that don't have a bucket set due to deleting bucket before deleting all the objects
2. fixing issue when we didn't update account default pool after the first optimal pool was created - due to breaking the loop if no system when start running pool_server

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/pool_server.js
+++ b/src/server/system_services/pool_server.js
@@ -61,11 +61,6 @@ async function _init() {
         system_store.is_finished_initial_load
     );
 
-    const system = system_store.data.systems[0];
-    if (!system) {
-        return;
-    }
-
     // Ensure that all account are not using mongo pool as default resource after
     // at least one pool is in optimal state.
     update_account_default_resource();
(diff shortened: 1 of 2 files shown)


nimrod-becker (approved):

src/server/system_services/pool_server.js
```diff
@@ -1131,37 +1126,39 @@ async function get_optimal_non_mongo_pool_id() {
...
-                        _is_mongo_pool(account.default_pool)
-                    )
-                    .map(account => ({
-                        _id: account._id,
-                        $set: {
-                            'default_pool': optimal_pool_id
-                        }
-                    }));
+        const system = system_store.data.systems[0];
+        if (system) {
```
nb-ohad, line 1130: I would move this code inside the try block.

nb-ohad (approved):

src/server/system_services/pool_server.js
jackyalbo: no need as we know system_store is already initiated - but in order to ease up you mind

PR #6081, opened 2020-07-08
[Phase 2] jenkins: build container images with podman, run with minikube/docker

nixpanic:
Building container images with Docker in the CentOS CI fails due to some
network conflicts. The containers running under Docker do not have full
access to the internet.

Instead of reconfiguring the whole deployment, build the container
images with podman. After building, the images will not be available in
the local Docker registry, so export and import them.


liranmauda (approved):
LGTM

guymguym:
@liranmauda Is this ready to merge?

PR #6086, opened 2020-07-13
NS-CACHE: Eviction from the cache

guptsanj:
### Explain the changes
Cached chunks are evicted when a namespace bucket with cache starts to fill up.
The tier ttf worker will periodically perform the eviction.
When bucket reaches a predefined limit, entries are no longer cached
Any in transit objects/chunks are also dropped.

1. 

### Testing Instructions:
1.


frontend/src/app/components/modals/create-namespace-bucket-modal/create-namespace-bucket-modal.js
```diff
@@ -169,7 +172,8 @@ class CreateNamespaceBucketModalViewModel extends ConnectableViewModel {
             isStepValid: isFormValid(form),
             readPolicy,
             writePolicy,
-            cacheTTL
+            cacheTTL,
+            cacheEvictionType
```
nb-ohad, line 176: There is no `cacheEvictionType` prop on the view model (the current class), so this assignment is meaningless.
Also, I am not sure if you need it on the view model at all, because in the view you could always read it from `$form.cacheEvictionType`

frontend/src/app/components/modals/create-namespace-bucket-modal/create-namespace-bucket-modal.js
```diff
@@ -118,6 +119,8 @@ class CreateNamespaceBucketModalViewModel extends ConnectableViewModel {
         const readPolicy = getFieldValue(form, 'readPolicy');
         const writePolicy = getFieldValue(form, 'writePolicy');
         const cacheTTL = getFieldValue(form, 'cacheTTL');
+        const cacheEvictionType = getFieldValue(form, 'cacheEvictionType');
```
nb-ohad, line 122: A better way to read more than one or two values from a form is to use `getFormValues` to fetch all the values at once instead of fetching them one by one.

```suggestion
        const { readPolicy, writePolicy, cacheTTL, cacheEvictionType } = getFormValues(form);
```

frontend/src/app/components/modals/create-namespace-bucket-modal/create-namespace-bucket-modal.js
```diff
@@ -97,7 +97,8 @@ class CreateNamespaceBucketModalViewModel extends ConnectableViewModel {
         readPolicy: [],
         writePolicy: undefined,
         cacheTTL: 60000,
-        enableCaching: false
+        enableCaching: false,
+        cacheEvictionType: 'LRU'
```
nb-ohad, line 101: I cannot find any edits on the corresponding view file (the HTML file) that is using this form field. If the view does not use it please remove it or add the change to the view file.

if you remove it all other changes in this are meaningless and should also be removed.
If you just need to send LRU in the dispatch for `createNamespaceBucket` below you don't need all these changes and can just hardcode the `"LRU"` value in the `createNamespaceBucket` call

nb-ohad (changes requested):

frontend/src/app/components/modals/create-namespace-bucket-modal/create-namespace-bucket-modal.js
guptsanj: I added a new type config.NAMESPACE_CACHING.DEFAULT_EVICTION_TYPE
This is called in used in `src/server/system_services/bucket_server.js` to set the eviction property when a bucket is created

frontend/src/app/components/modals/create-namespace-bucket-modal/create-namespace-bucket-modal.js
guptsanj: Removed my changes

frontend/src/app/components/modals/create-namespace-bucket-modal/create-namespace-bucket-modal.js
guptsanj: Removed

src/sdk/namespace_cache.js
```diff
@@ -48,7 +58,7 @@ class NamespaceCache {
 
     async list_objects(params, object_sdk) {
         // TODO listing from cache only for deevelopment
-        return this.namespace_nb.list_objects(params, object_sdk);
+        return this.namespace_hub.list_objects(params, object_sdk);
```
paulxuzhang, line 61: Can we support listing from cache option through query parameter get_from_cache? If get_from_cache is set, it lists objects from cache.

src/server/object_services/map_deleter.js
```diff
@@ -74,6 +75,19 @@ async function delete_blocks(blocks) {
...
 
+/**
+ * @param {nb.Chunk[]} chunks
+ */
+async function delete_chunk_parts(chunks) {
+    if (!chunks || !chunks.length) return;
+    const parts = get_all_chunk_parts(chunks).filter(part => !part.to_db().deleted);
+    const part_ids = parts.map(part => part._id);
+    dbg.log0('delete_chunk_parts: parts ', part_ids);
+    await Promise.all([
```
jackyalbo, line 86: No need for promise.all for only one promise,
```suggestion
    await MDStore.instance().delete_parts_by_ids(part_ids)
```

src/sdk/namespace_cache.js
```diff
@@ -47,8 +57,9 @@ class NamespaceCache {
     /////////////////
 
     async list_objects(params, object_sdk) {
-        // TODO listing from cache only for deevelopment
-        return this.namespace_nb.list_objects(params, object_sdk);
+
+        return params.get_from_cache ? this.namespace_nb.list_objects(params, object_sdk) :
```
jackyalbo, line 61: where are this param will come from? doesn't it suppose to be part of s3_get_bucket?

src/sdk/namespace_cache.js
```diff
@@ -77,6 +88,7 @@ class NamespaceCache {
             if (get_from_cache) {
```
jackyalbo, line 88: the same - looks like we are missing the s3_get_bucket_uploads code

src/server/object_services/map_server.js
```diff
@@ -417,16 +417,20 @@ async function make_room_in_tier(tier_id, bucket_id) {
 
         const tiering_status = node_allocator.get_tiering_status(tiering);
         const next_tier = mapper.select_tier_for_write(tiering, tiering_status, tier_and_order.order + 1);
-        if (!next_tier) {
+
+        if (!next_tier && !bucket.namespace.caching.eviction) {
```
jackyalbo, line 421: I don't really get this... seems like we must have eviction with caching - there is a default always (correct me if wrong) - so shouldn't we check just for namespace? or to add a no-eviction option to this PR?

jackyalbo (changes requested):

src/server/object_services/map_server.js
guptsanj: I added this here as at the moment, we have just one eviction policy, LRU. In future if we have any other policy, it will be very easy to update the code. If eviction policy is not defined, we let the default behavior go through.
I have updated all other issues

frontend/src/app/config.json
```diff
@@ -1,6 +1,7 @@
 {
     "patchFile": "/fe/config.patch.json",
     "inputThrottle": 300,
+    "eviction_type": "LRU",
```
nb-ohad, line 4: The console (frontend) config file is used for configuration related to visual settings. I don't think `eviction_type` should be added here.

Another small note,  the FE code conventions dictates the usage of camelcase and not snakecase

frontend/src/app/components/modals/create-namespace-bucket-modal/create-namespace-bucket-modal.js
```diff
@@ -250,7 +251,8 @@ class CreateNamespaceBucketModalViewModel extends ConnectableViewModel {
             closeModal(),
             createNamespaceBucket(bucketName, readPolicy, writePolicy,
                 this.enableCaching() ? {
-                    ttl_ms: cacheTTL
+                    ttl_ms: cacheTTL,
+                    eviction: eviction_type
```
nb-ohad, line 255: Because this is a constant, I think it is better to encode it directly in the epic that is sending the request 
at `frontend/src/app/epics/create-namespace-bucket.js.
There is no need to pass the information through the ViewModel.

nb-ohad (dismissed):

frontend/src/app/components/modals/create-namespace-bucket-modal/create-namespace-bucket-modal.js
guymguym: irrelevant in latest changes

frontend/src/app/config.json
guymguym: irrelevant in latest changes

src/server/object_services/map_server.js
guymguym: irrelevant in latest changes

src/sdk/namespace_cache.js
guymguym: Right, I see that @guptsanj added it to s3_get_bucket in #6130

src/api/scrubber_api.js
```diff
@@ -23,6 +23,9 @@ module.exports = {
                         items: { objectid: true }
                     },
                     tier: { objectid: true },
+                    purge_chunks: {
```
guymguym, line 26: Lets try to update our terminology to `evict` across this PR

src/sdk/map_api_types.js
```diff
@@ -493,8 +493,20 @@ function get_all_chunks_blocks(chunks) {
...
+/**
+ * 
+ * @param {nb.Chunk[]} chunks 
+ * @returns {nb.Part[]}
+ */
+function get_all_chunk_parts(chunks) {
+    return /** @type {nb.Part[]} */ (
+        /** @type {unknown} */
+        (_.flatMapDeep(chunks, chunk => chunk.parts))
+    );
```
guymguym, line 505: Since this is just one level then we don't need the deep flat map variant and can use the native array method, which shouldn't have all that hassle with types (I think so):
```suggestion
    return chunks.flatMap(chunk => chunk.parts);
```

src/server/object_services/map_deleter.js
```diff
@@ -43,6 +44,18 @@ async function delete_chunks_if_unreferenced(chunk_ids) {
     }
 }
 
+/**
+ * @param  {nb.ObjectMD} object
+*/
+async function delete_empty_object(object) {
+    if (!object) return;
+    dbg.log2('delete_empty_objects: object_ids', object);
+    const part_ids = await MDStore.instance().find_all_parts_of_object(object);
```
guymguym, line 53: This is doing a lot more work than we need - we only care if there exist any parts, and we don't care how many or to return their information. So we can do a findOne type of query (limit results to 1), but in any case I would set the MDStore API to only ask if `has_any_parts_for_object()` similar to all the other `has_any_*` functions there.

guymguym (commented):
Still reviewing, but sending out my comments so far.

src/server/object_services/map_builder.js
```diff
@@ -116,7 +119,9 @@ class MapBuilder {
 
         /** @type {nb.Chunk[] } */
         const chunks_to_build = [];
-
+        if (this.purge_chunks) {
+            Array.prototype.push.apply(chunks_to_delete, live_chunks);
```
guymguym, line 123: We have a new syntax for that :)
```suggestion
            chunks_to_delete.push(...live_chunks);
```

guymguym:
Another note was about the call to `usage_aggregator.get_bandwidth_report()` in the worker here:
https://github.com/noobaa/noobaa-core/blob/0f732250b83ce4581fe73639e70bbc51140db32b/src/server/bg_services/tier_ttf_worker.js#L74-L79

@nb-ohad suggested to use lower time_range and sum up the results.

src/server/object_services/map_deleter.js
```diff
@@ -43,6 +44,17 @@ async function delete_chunks_if_unreferenced(chunk_ids) {
     }
 }
 
+/**
+ * @param  {nb.ObjectMD} object
+*/
+async function delete_object_if_no_parts(object) {
+    if (!object) return;
+    dbg.log1('delete_object_if_no_parts: object_ids', object);
+    if (!MDStore.instance().has_any_parts_for_object(object)) {
```
guymguym, line 53: missing await, but I would not add it inside an `if` so I suggest:
```suggestion
    const has_parts = await MDStore.instance().has_any_parts_for_object(object);
    if (!has_parts) {
```

src/server/object_services/map_deleter.js
```diff
@@ -43,6 +44,17 @@ async function delete_chunks_if_unreferenced(chunk_ids) {
     }
 }
 
+/**
+ * @param  {nb.ObjectMD} object
```
guymguym, line 48: Please add a comment:
```suggestion
 * Delete the object_md of objects that have no parts.
 * This is needed for the cleanup of cache buckets once all parts are evicted.
 * @param  {nb.ObjectMD} object
```

guymguym (changes requested):

guymguym (changes requested):
@guptsanj See changes requested. Please let me know if I can help.

guymguym:
@jackyalbo Please review and change your review to approved if you have no additional comments - we are trying to complete this PR this week. Thanks

jackyalbo (approved):

src/server/object_services/map_builder.js
```diff
@@ -148,18 +153,23 @@ class MapBuilder {
...
         await Promise.all([
+            map_deleter.delete_parts_by_chunks(chunks_to_delete_uniq),
             map_deleter.delete_blocks(blocks_to_delete),
             map_deleter.delete_chunks(chunks_to_delete_uniq),
-        ]);
 
+        ]);
+        // Deleting objects with no parts here as the delete_parts_by_chunks need to finish before
+        // any attempt is made to delete the objects.
+        await Promise.all(objects_uniq.map(map_deleter.delete_object_if_no_parts));
```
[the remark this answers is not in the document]
guptsanj: done

src/server/bg_services/tier_ttf_worker.js
```diff
@@ -139,8 +166,8 @@ class TieringTTFWorker {
         return bucket.tiering.tiers.find(t => String(t.tier._id) === String(tier._id)).order;
     }
 
-    async _build_chunks(chunk_ids, next_tier) {
-        return this.client.scrubber.build_chunks({ chunk_ids, tier: next_tier }, {
+    async _build_chunks(chunk_ids, next_tier, cache_evict) {
+        return this.client.scrubber.build_chunks({ chunk_ids, tier: next_tier, cache_evict }, {
```
guymguym, line 170: You called this just `evict` in scrubber_api so that should fail you on schema validation.
```suggestion
        return this.client.scrubber.build_chunks({ chunk_ids, tier: next_tier, evict: cache_evict }, {
```

src/server/object_services/map_builder.js
```diff
@@ -35,13 +36,14 @@ const builder_lock = new KeysLock();
...
+     * @param {nb.ID[]} chunk_ids
      * @param {nb.Tier} [move_to_tier]
+     * @param {boolean} [evict]
      */
-    constructor(chunk_ids, move_to_tier) {
+    constructor(chunk_ids, move_to_tier, evict) {
         this.chunk_ids = chunk_ids;
         this.move_to_tier = move_to_tier;
-
+        this.evict = evict;
```
guymguym, line 46: I think perhaps it would be wise to assert here that evict and move_to_tier are mutually exclusive, so that we can reduce our checks later to be simply `if (this.evict)...` instead of repeating `if (this.evict && !this.move_to_tier) ...`
```javascript
if (evict) assert.strictEqual(move_to_tier, undefined);
```

src/server/object_services/map_builder.js
```diff
@@ -150,16 +156,33 @@ class MapBuilder {
             'chunks_to_delete_uniq', chunks_to_delete_uniq.length,
             'blocks_to_delete', blocks_to_delete.length);
 
+        // Delete parts only if this is eviction mode, and not move to tier
+        const delete_parts = (this.evict && !this.move_to_tier) ?
+                        map_deleter.delete_parts_by_chunks(chunks_to_delete_uniq) :
+                        Promise.resolve();
+
         await Promise.all([
+            delete_parts,
```
guymguym, line 165: usually it's easy enough to add such conditions inline and keep the "IO" flow simpler to read:

```suggestion
        await Promise.all([
            // Delete parts only if this is eviction mode, and not move to tier
            this.evict && map_deleter.delete_parts_by_chunks(chunks_to_delete_uniq),
```

guymguym (changes requested):
Hi @guptsanj Good job, this is almost done.
I found a couple more things - see below.
Also what about `usage_aggregator.get_bandwidth_report()` - didn't you need to change it?

guymguym (approved):
@guptsanj Great work. See my comment above - I think removing the age filter for objects_to_gc is better, but in any case this is not a real issue and we can go ahead with this PR and improve object_md eviction in next PR.

PR #6090, opened 2020-07-15
Fixes #5993 - fixes unzipping a function package with node_modules

### LINKED ISSUE #5993: Question: What is the planed strategy for using Noobaa-Function with additional modules needed.
Is there any documentation available on how to include additional modules used by functions?

We tried adding modules within a zip-file with a subfolder (./node_modules/…) which resulted in a function that can no longer be triggered (Errorcode: “new invocation failed”).

If we build core-images on our own we include modules but lack the possibility of automatic updates.

dannyzaken:
### Explain the changes
1. when a function package contained node_modules it was not recognized as a directory
2. the problem was that `path.resolve` omitted the `/` at the end of the `path_name`
3. testing the original `filename` for `/` instead of `path_name`

1. Fixes #5993 

### Testing Instructions:
1. From the UI - create a new function and upload the attached `wc.zip` code package
2. Try to invoke (**Set Event And Invoke** under Code tab in Functions section)

[wc.zip](https://github.com/noobaa/noobaa-core/files/4926692/wc.zip)

### DIFF
--- a/src/util/zip_utils.js
+++ b/src/util/zip_utils.js
@@ -91,9 +91,10 @@ function unzip_to_mem(zipfile) {
 
 function unzip_to_dir(zipfile, dir) {
     return unzip_to_callback(zipfile, (entry, stream) => {
-        const path_name = path.resolve(dir, '.' + path.sep + entry.fileName.toString());
+        const filename = entry.fileName.toString();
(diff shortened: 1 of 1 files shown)


dannyzaken:
@nimrod-becker can you review?

nimrod-becker (approved):

PR #6091, opened 2020-07-15
[WIP] Addition of bulks as transactions

jeniawhite:
### Explain the changes
1. Converted bulks into transactions and let system_store run on top of PostgreSQL

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -26,8 +26,8 @@ const time_utils = require('../../util/time_utils');
 const size_utils = require('../../util/size_utils');
 const os_utils = require('../../util/os_utils');
 const mongo_utils = require('../../util/mongo_utils');
-const mongo_client = require('../../util/mongo_client');
 const config = require('../../../config');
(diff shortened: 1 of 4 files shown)


nimrod-becker (approved):

src/server/system_services/system_store.js
```diff
@@ -540,17 +540,18 @@ class SystemStore extends EventEmitter {
...
-                .toArray()
-                .then(res => {
-                    for (const item of res) {
-                        this._check_schema(col, item, 'warn');
-                    }
-                    target[col.name] = res;
-                })
+                col => {
+                    let f = db_client.instance().collection(col.name).find(non_deleted_query);
+                    if (!config.USE_POSTGRESQL) f = f.toArray();
```
guymguym, line 547: @jeniawhite Can't you hide this inside the db_client? I wouldn't want the system_store to be filled with references to config.USE_POSTGRESQL. In general we will keep supporting both db modes, and possibly add another, so this should be clean from the caller side.

PR #6115, opened 2020-08-04
Centralize creation of HTTP/S agents with support for proxied environment

nb-ohad:
### Explain the changes
1.  Respect proxy setting acquired from the HTTP_PROXY, HTTPS_PROXY and NO_PROXY env variables
2. centeralize HTTP/S agent creation via `http_utils.get_http_agent`
3. All aws clients, direct http/s requests,  http rpc and ws rpc are now using `get_http_agent` to get an agent with proxy support

1. Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1862755

### Testing Instructions:
1. 


src/util/http_utils.js
```diff
@@ -5,15 +5,23 @@ const _ = require('lodash');
...
 const P = require('./promise');
 const dbg = require('./debug_module')(__filename);
 const xml_utils = require('./xml_utils');
 const cloud_utils = require('./cloud_utils');
 const util = require('util');
 
+
+const no_proxy_list = process.env.NO_PROXY ?
+    process.env.NO_PROXY.split(',') :
+    [];
```
guymguym, line 23: what do you think about mapping each item to a should_proxy() function based on it's type, something like:
```suggestion
/**
 * should_not_proxy_list is an array of functions that each check a given address for a no proxy match
 * @type Array<(address: string) => boolean>
 */
const should_not_proxy_list = (process.env.NO_PROXY || []).split(',').map(cond => {

    // Cover the case where cond is an IP
    if (ip.isV4Format(cond) || ip.isV6Format(cond)) return x => ip.isEqual(x, cond);

    // Cover the case where cond is a CIDR subnet (eg. 10.20.30.40/24)
    let subnet;
    try { subnet = ip.cidrSubnet(cond); } catch () {}
    if (subnet) return x => subnet.contains(x);

    // Cover the case where cond is domain name suffix
    if (cond.startsWith('.')) return x => x.endsWith(cond);

    // Cover the case where cond is an explicit FQDN
    return x => (x === addr);
});

function should_proxy(address) {
    // if all the "should not proxy" functions return false, we should proxy.
    // if any one of "should not proxy" functions returns true for address, we shouldn't proxy.
    return should_not_proxy_list.every(fn => !fn(address));
}

```

src/util/cloud_utils.js
```diff
@@ -22,7 +22,7 @@ function find_cloud_connection(account, conn_name) {
 
 
 function get_signed_url(params) {
-    let agent = http_utils.get_unsecured_http_agent(params.endpoint);
+    const agent = http_utils.get_http_agent(params.endpoint, { rejectUnauthorized: false });
```
guymguym, line 25: Why did you cancel the secure vs unsecure separation?
It was intentional to have a separate identifier for unsecure calls.

src/sdk/object_sdk.js
```diff
@@ -233,9 +233,10 @@ class ObjectSDK {
             ns_info.endpoint_type === 'S3_COMPATIBLE' ||
             ns_info.endpoint_type === 'FLASHBLADE' ||
             ns_info.endpoint_type === 'IBM_COS') {
-            const httpOptions = (ns_info.endpoint_type === 'AWS') ? undefined : {
-                agent: http_utils.get_unsecured_http_agent(ns_info.endpoint)
-            };
+
+            const agent = http_utils.get_http_agent(ns_info.endpoint, {
+                rejectUnauthorized: ns_info.endpoint_type !== 'AWS'
```
guymguym, line 238: you have a similar check inside get_http_agent to detect aws endpoints.
why do you have it in both?

guymguym (changes requested):

src/util/http_utils.js
```diff
@@ -287,19 +295,66 @@ function send_reply(req, res, reply, options) {
...
+    for (const addr of no_proxy_list) {
+        // Cover the case where the address is an IP
+        if ((ip.isV4Format(addr) || ip.isV6Format(addr)) && ip.isEqual(hostname, addr)) {
+            return false;
+        }
+        // Cover the case where the address is a CIDR subnet
+        try {
+            if (ip.cidrSubnet(addr).contains(hostname)) {
+                return false;
+            }
```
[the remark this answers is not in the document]
nb-ohad: First, The try catch block is here to protect against the case where  `addr` is not in the cidr subnet format `#.#.#.#/#` not because of the format of `hostname`

Second, if the hostname does not match the cidr then we still must no return (even if the `addr` is a cidr because we still need to check against the other addresses in the list

src/util/http_utils.js
```diff
@@ -287,19 +295,66 @@ function send_reply(req, res, reply, options) {
...
+        }
+
+    } else if (HTTP_PROXY && should_proxy(hostname)) {
+        return new HttpProxyAgent({
+            ...options,
+            ...url.parse(HTTP_PROXY)
+        });
+
+    } else {
+        return new http.Agent(options);
```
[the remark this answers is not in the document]
nb-ohad: Not sure if an agent is so expensive but I can map to global agents. The issue here will be that the agent options might change for each request so it is not so simple.

Also, the old code also created a new agent for each call, I believe for the same reason.

src/util/http_utils.js
nb-ohad: I was also thinking about it,
But it will just create more complexity in a non-crutial code path. I prefer to provide a simpler solution until we will prove that we are bottlenecked on the addresses type check

src/util/cloud_utils.js
nb-ohad: In the old code, the only case for a custom agent was an unsecured agent. Now we need an agent also because of the proxy stuff. So I just generalized the agent creation code and passed the un-secure flag when necessary  (as part of  an options object)

src/sdk/object_sdk.js
nb-ohad: It is not exactly the same. 
This condition checks the `endpoint_type` while the other checks the actual endpoint address. I didn't want to take the chance and covered the case where we have an AWS type is set but not an endpoint with `amazon.com` in it. 
You might call it defensive coding

Also, I took this approach because this check was there even before the change which made me assume that the above case might happen.

src/util/http_utils.js
nb-ohad: @guymguym Also, you have a bug in the proposed code. you cannot split on an empty array, and even if you choose to set the default value to an empty string it will create an array with one empty item and not an empty array.

src/util/http_utils.js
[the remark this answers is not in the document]
nb-ohad: Another Idea is to cache the agent on the call site. in cases where the call site can use the same agent every time.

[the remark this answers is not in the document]
guymguym: which agent options are we passing per request??

src/sdk/object_sdk.js
jeniawhite, line 238: This is incorrect. If we are talking to AWS we want `rejectUnauthorized = true` in this case we will have false.

src/util/http_utils.js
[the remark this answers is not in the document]
nb-ohad: The obvious one which is widely used is `rejectUnautorized`, for proxied agents we also set the proxy info.
But as a util I think it might be best to allow agent creation with any custom agent config and still make sure that on proxied evns we get a proxied agent. 

I can suggest the following: 
Return a cached agent in the following cases:
1. non-proxied HTTP agent without custom options
2. proxied HTTP agent without custom options
3. non-proxied non-secured HTTPS agent without custom options
4. proxied non-secured HTTPS agent without custom options
5. non-proxied secured HTTPS agent without custom options
6. proxied secured HTTPS agent without custom options

In any other case create a specific agent and return it.

@guymguym what do you think?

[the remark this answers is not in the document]
nb-ohad: Found another gap, we are setting `keepAlive` option  on the global agent, but now we do not use the global agent anymore

src/util/http_utils.js
nb-ohad: Changed the code a bit to pre-process the list immediately after the split. It is not exactly your suggestion but it allows us to select tests based on the hostname type (IP or FQDN) and does not invoke a multitude of functions.

src/util/http_utils.js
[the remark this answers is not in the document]
guymguym: This is nodejs http.Agent - https://github.com/nodejs/node/blob/master/lib/_http_agent.js
I really don't think this is meant to be created per request, and I don't see why it should.

From my point of view there should be 4 global static agents 1. proxy&secure 2. proxy&unsecure 3. no-proxy&secure 4. no-proxy&unsecure.

And I would like to have two separate exported functions - 1. get_secure_http_agent 2. get_unsecure_http_agent which will return the proper one of these 4.

src/util/http_utils.js
guymguym: ok, resolving.

src/util/cloud_utils.js
guymguym: I still prefer that it's a separate function.

src/sdk/object_sdk.js
guymguym: I would leave this in one place - in object_sdk in my opinion.
because finding such a condition in a generic http_utils is quite a surprising side effect.

guymguym: keep the util passive, and the caller can choose secure/unsecure according to specific needs.

src/util/cloud_utils.js
nb-ohad: This is not a problem I can add a specialized function just got that

src/sdk/object_sdk.js
nb-ohad: I see your point, and I was also surprised to see that inside the original utils function, but removing this from the utils code might have side effects when calling sites assumes that amazon will always be secured.
Analyzing every calling site for the possibility that an `amazon.com` address might be used there should be its own task that we should target for a later PR. This PR is mainly for resolving the issue attached in the description.

src/util/http_utils.js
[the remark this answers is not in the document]
nb-ohad: Please refer to my list above, you are missing the other 2 agents, the HTTP one and the PROXIED HTTP one. 
And on top of that, every call site might need it's one instance of an agent when the parameters to the agent differ.
For example `keepAlive=true` vs `keepAlive=false`.

This is why I think caching the agent at the call site is better. you will get a single agent for every invocation from a single call site or even multiple call sites in the same file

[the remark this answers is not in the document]
guymguym: ok so 6, but even if there are 100 the fact theey are static is important. 
Why do we needs control over keepAlive per request??
I don't think we want to control the agent properties per request, just to pick the agent and make our request with it.

src/util/cloud_utils.js
guymguym: great, 2 specialized functions

guymguym: make the internal function actually internal

nb-ohad: It is now internal

src/util/http_utils.js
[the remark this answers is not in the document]
nb-ohad: I now keep the 6 as static agents and allow to set global settings for all HTTP and HTTPS agents at once using new funciton in http_utils: `update_http_agents` and `update_https_agents`

src/sdk/object_sdk.js
nb-ohad: @guymguym can this be resolved?

src/util/cloud_utils.js
```diff
@@ -35,7 +34,7 @@ function get_signed_url(params) {
         s3DisableBodySigning: disable_s3_compatible_bodysigning(params.endpoint),
         region: 'eu-central-1',
         httpOptions: {
-            agent: agent
+            agent: http_utils.get_unsecured_agent(params.endpoint)
```
guymguym, line 37: I am pretty sure that this is unneeded because we only call getSignedUrl() on this s3 client and the actual http call will be made by the client that gets this url (endpoint or browser). No harm by passing it, but maybe just worth noting that in a comment so that it will be clearer that this is not really a must.

guymguym (approved):

src/util/cloud_utils.js
nb-ohad: It does not harm, would you like me to remove that agent?

guymguym: no. just a small comment would be nice because as a reader it's hard to understand if this was intentional or not.

nb-ohad: Comment added

PR #6117, opened 2020-08-09
Backport to 5.5

nimrod-becker:
- Respect proxy setting acquired from the HTTP_PROXY, HTTPS_PROXY and NO_PROXY env variables
- Remove constraints on max number of allowed socket (will now  default to infinity)

(cherry picked from commit a3dd1988e1f2a8f01aba30c81aece0358046e0e1)

### DIFF
--- a/src/agent/block_store_services/block_store_client.js
+++ b/src/agent/block_store_services/block_store_client.js
@@ -15,6 +15,8 @@ const LRUCache = require('../../util/lru_cache');
 const config = require('../../../config');
 const { RPC_BUFFERS, RpcError } = require('../../rpc');
 const { get_block_internal_dir } = require('../../agent/block_store_services/block_store_base');
+const util = require('util');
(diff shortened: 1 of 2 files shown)


src/endpoint/blob/blob_utils.js
```diff
@@ -53,6 +54,60 @@ function parse_copy_source(req) {
...
+                return acc;
+            }, { name: key.Name[0] });
+            return obj;
+        };
+        blobs = _.map(blobs, obj => parse_blobs(obj));
+    } catch (err) {
+        throw new Error(`could not parse body ${body} got error ${err}`);
+    }
+    return { blobs, dirs, next_marker };
+}
```
nb-ohad: This is not part of the fix and should not be introduced into 5.5

nb-ohad (changes requested):

Issue #6070, opened 2020-07-02
Replace Bluebird with native promise


guymguym:
See the list of occurrences below.
I think we can start from the top and resolve them one by one.
For example `P.resolve` can be replaced with `Promise.resolve`.
But `Promise.fromCallback` requires refactor or keeping a util function.
I suggest to keep this module in `require(./src/util/promise')` to have promise utilities and even merge into it the utils from `require('./src/util/promise_utils')`.

```
$ rg 'P\.\w+|new P\(' --only-matching --no-heading --no-line-number --no-filename | sort | uniq -c | sort -nr
 334 P.resolve
 101 P.fromCallback
  98 P.delay
  84 P.map
  75 P.fcall
  46 P.all
  41 P.ninvoke
  37 new P(
  34 P.join
  16 P.try
  15 P.defer
  11 P.nfcall
  10 P.reject
  10 P.each
   5 P.TimeoutError
   4 P.promisifyAll
   4 P.hijackQ
   3 P.config
   2 P.test
   2 P.prototype
   2 P.props
   2 P.promisify
   2 P.any
   1 P.then
   1 P.catch
```

liranmauda:
👍 
I think that in simple function like: https://github.com/noobaa/noobaa-core/blob/master/src/agent/agent.js#L222:L235
We can even eliminate the use of P.resolve like this:
```
    async start() {
        const dbg = this.dbg;
        this.is_started = true;
        this.rpc.set_disconnected_state(false);
      try { 
       await this._init_node();
       await this._do_heartbeat();
      } catch (err) {
                dbg.error('server failed to start', err.stack || err);
                this.stop();
                throw err;
      }
    }
```
don't code review it, Just my 2 cents... 🙂

guymguym:
Sure, but I would actually start with a simple search replace of `P.resolve(` to `Promise.resolve(` because that case is at least trivial to replace without refactoring to async-await.

liranmauda:
Changing `P.resolve()` to `Promise.resolve()` where there is `.return()` `.timeout()` `.delay()` and `.spread()` is not possible without refractoring.

guymguym:
@liranmauda can we close?

liranmauda:
@guymguym I guess so.
Closing.

PR #6120, opened 2020-08-09
Replace Bluebird with Promise - next stage

liranmauda:
### Explain the changes
- Using delay in a native way (new delay function at promise_utils)
- Removed all `.tap()`, `.spread()` and `.return()`
- Removed most of `P.join`
- Wrapped with bluebird some of the timeout and catch

### Out of scope
- Did not replaced `.timeout()`
- Did not replaced `.cancel()`
- Did not replaced `.isPending()`
- Did not replaced `.isRejected()`
- Did not replaced `.props()`
- Did not replaced 2 `.delay()`

Step 1 for #6070 

### DIFF
--- a/src/agent/agent.js
+++ b/src/agent/agent.js
@@ -12,6 +12,7 @@ const os = require('os');
 const util = require('util');
 
 const P = require('../util/promise');
+const Bluebird = require('bluebird');
 const api = require('../api');
 const pkg = require('../../package.json');
 const DebugLogger = require('../util/debug_module');
@@ -638,9 +639,11 @@ class Agent {
                 'old', params.old_base_address);
             // test this new address first by pinging it
             try {
-                await this.client.node.ping(null, {
-                        address: params.base_address
-                    })
+                await Bluebird.resolve(
+                        this.client.node.ping(null, {
+                            address: params.base_address
+                        })
+                    )
                     .timeout(MASTER_RESPONSE_TIMEOUT);
             } catch (err) {
                 dbg.error(`failed ping to new base_address ${params.base_address}. leave base_address at ${this.base_address}`);
@@ -660,9 +663,11 @@ class Agent {
             const { old_master_address = this.master_address } = params;
             if (params.master_address !== old_master_address) {
                 try {
-                    await this.client.node.ping(null, {
-                            address: params.master_address
-                        })
+                    await Bluebird.resolve(
+                            this.client.node.ping(null, {
+                                address: params.master_address
+                            })
+                        )
                         .timeout(MASTER_RESPONSE_TIMEOUT);
                 } catch (err) {
                     dbg.error(`failed ping to new master_address ${params.master_address}.`,
@@ -700,11 +705,11 @@ class Agent {
         clearTimeout(this._test_connection_timeout);
 
         this._test_connection_timeout = setTimeout(() => {
-            this.client.node.ping()
-                .timeout(MASTER_RESPONSE_TIMEOUT)
-                .then(() => {
-                    this._test_connection_timeout = null;
-                })
+            Bluebird.resolve(this.client.node.ping())
+                    .timeout(MASTER_RESPONSE_TIMEOUT)
+                    .then(() => {
+                        this._test_connection_timeout = null;
+                    })
                 .catch(P.TimeoutError, err => {
                     dbg.error('node_server did not respond to ping. closing connection', err);
                     this._server_connection.close();
@@ -857,7 +862,7 @@ class Agent {
                 //     }
                 // });
             })
-            .return(reply);
+            .then(() => reply);
     }
 
     async get_agent_storage_info(req) {
@@ -1001,7 +1006,7 @@ class Agent {
                 });
         };
 
-        return P.all(_.times(concur, next)).return(reply);
+        return P.all(_.times(concur, next)).then(() => reply);
     }
 
     collect_diagnostics(req) {
--- a/src/agent/agent_cli.js
+++ b/src/agent/agent_cli.js
@@ -329,8 +329,8 @@ class AgentCLI {
             current_path = current_path.replace('./', '');
             //hiding storage folder
             return child_process.execAsync('attrib +H ' + current_path)
-                .then(() => P.join(os_utils.is_folder_permissions_set(current_path), fs.readdirAsync(current_path)))
-                .spread(function(permissions_set, noobaa_storage_initialization) {
+                .then(() => Promise.all([os_utils.is_folder_permissions_set(current_path), fs.readdirAsync(current_path)]))
+                .then(([permissions_set, noobaa_storage_initialization]) => {
                     if (!permissions_set) {
                         if (_.isEmpty(noobaa_storage_initialization)) {
                             dbg.log0('First time icacls configuration');
--- a/src/agent/agent_diagnostics.js
+++ b/src/agent/agent_diagnostics.js
@@ -68,7 +68,7 @@ function collect_agent_diagnostics(storage_path) {
                 });
         })
         .finally(() => diag_log.end())
-        .return()
+        .then(() => undefined)
         .catch(err => {
             console.error('Error in collecting server diagnostics', err);
             throw new Error('Error in collecting server diagnostics ' + err);
--- a/src/agent/block_store_services/block_store_fs.js
+++ b/src/agent/block_store_services/block_store_fs.js
@@ -61,7 +61,7 @@ class BlockStoreFs extends BlockStoreBase {
     }
 
     get_storage_info() {
-        return P.join(
+        return Promise.all([
                 this._get_usage(),
                 os_utils.get_drive_of_path(this.root_path)
                 .catch(err => {
@@ -69,8 +69,8 @@ class BlockStoreFs extends BlockStoreBase {
                     this._test_root_path_exists();
                     throw err;
                 })
-            )
-            .spread((usage, drive) => {
+            ])
+            .then(([usage, drive]) => {
                 const storage = drive.storage;
                 storage.used = usage.size;
                 const total_unreserved = Math.max(storage.total - config.NODES_FREE_SPACE_RESERVE, 0);
@@ -89,10 +89,10 @@ class BlockStoreFs extends BlockStoreBase {
         const block_path = this._get_block_data_path(block_md.id);
         const meta_path = this._get_block_meta_path(block_md.id);
         dbg.log1('fs read block', block_path);
-        return P.join(
+        return Promise.all([
                 fs.readFileAsync(block_path),
-                fs.readFileAsync(meta_path))
-            .spread((data_file, meta_file) => ({
+                fs.readFileAsync(meta_path)])
+            .then(([data_file, meta_file]) => ({
                 block_md: block_md,
                 data: data_file,
             }))
@@ -119,9 +119,9 @@ class BlockStoreFs extends BlockStoreBase {
                 overwrite_stat = stat;
                 dbg.log1('_write_block', block_path, data.length, overwrite_stat);
                 // create/replace the block on fs
-                return P.join(
+                return Promise.all([
                     fs.writeFileAsync(block_path, data),
-                    fs.writeFileAsync(meta_path, block_md_data));
+                    fs.writeFileAsync(meta_path, block_md_data)]);
             })
             .catch(err => {
                 if (err.code === 'ENOENT') {
@@ -185,7 +185,7 @@ class BlockStoreFs extends BlockStoreBase {
         let del_stat;
         let md_del_stat;
         dbg.log1("delete block", block_id);
-        return P.join(
+        return Promise.all([
                 fs.statAsync(block_path).catch(ignore_not_found)
                 .then(stat => {
                     del_stat = stat;
@@ -195,7 +195,7 @@ class BlockStoreFs extends BlockStoreBase {
                 .then(stat => {
                     md_del_stat = stat;
                     return fs.unlinkAsync(meta_path).catch(ignore_not_found);
-                }))
+                })])
             .then(() => {
                 if (this._usage && del_stat) {
                     let usage = {
@@ -218,7 +218,8 @@ class BlockStoreFs extends BlockStoreBase {
                 this._usage = usage; // object with properties size and count
                 // update usage file
                 let usage_data = JSON.stringify(this._usage);
-                return fs.writeFileAsync(this.usage_path, usage_data).return(usage);
+                return fs.writeFileAsync(this.usage_path, usage_data)
+                    .then(() => usage);
             });
     }
 
--- a/src/endpoint/ftp/ftp_filesystem.js
+++ b/src/endpoint/ftp/ftp_filesystem.js
@@ -90,7 +90,6 @@ class FtpFileSystemNB {
                 delimiter: '/',
                 limit: 100
             }))
-            .tap(console.log)
             .then(res => {
                 console.log(`in get - got res = `, res);
                 if (res.objects.length) {
@@ -101,7 +100,10 @@ class FtpFileSystemNB {
                     return this._get_new_file_stats_object(name, true, 0, new Date());
                 }
             })
-            .tap(console.log)
+            .then(res => {
+                console.log(res);
+                return res;
+            })
             .catch(err => {
                 dbg.error('got error in get(). ', err);
                 err.code = 'ENOENT';
@@ -123,12 +125,18 @@ class FtpFileSystemNB {
         console.log(`calling list_objects with params =`, params);
 
         return this.authenticate.then(() => this.object_sdk.list_objects(params))
-            .tap(console.log)
+            .then(res => {
+                console.log(res);
+                return res;
+            })
             .then(res => [this._get_new_file_stats_object('.', true, 0, new Date())]
                 .concat(res.objects.filter(obj => !obj.key.endsWith('/')) // filter out dirs
                     .map(obj => this._get_new_file_stats_object(path.parse(obj.key).base, false, obj.size, obj.create_time)))
                 .concat(res.common_prefixes.map(prefix => this._get_new_file_stats_object(path.parse(prefix).base, true, 0, new Date()))))
-            .tap(list => console.log(`retruning list`, list));
+            .then(list => {
+                console.log(`returning list`, list);
+                return list;
+            });
     }
 
     chdir(dir_path = '.') {
--- a/src/rpc/ice.js
+++ b/src/rpc/ice.js
@@ -276,13 +276,13 @@ Ice.prototype.accept = function(remote_info) {
  */
 Ice.prototype._add_local_candidates = function() {
     var self = this;
-    return P.join(
+    return Promise.all([
             self._add_udp_candidates(),
             self._add_tcp_active_candidates(),
             self._add_tcp_permanent_passive_candidates(),
             self._add_tcp_transient_passive_candidates(),
             self._add_tcp_simultaneous_open_candidates()
-        )
+        ])
         .then(function() {
             return {
                 credentials: self.local_credentials,
@@ -1496,7 +1496,7 @@ function listen_on_port_range(port_range) {
         server.listen(port);
         // wait for listen even, while also watching for error/close.
         return promise_utils.wait_for_event(server, 'listening')
-            .return(server)
+            .then(() => server)
             .catch(function(err) {
                 dbg.log1('ICE listen_on_port_range: FAILED', port, err);
                 server.close();
--- a/src/rpc/rpc.js
+++ b/src/rpc/rpc.js
@@ -904,7 +904,8 @@ class RPC extends EventEmitter {
         dbg.log0('RPC register_tcp_transport');
         const tcp_server = new RpcTcpServer(tls_options);
         tcp_server.on('connection', conn => this._accept_new_connection(conn));
-        return P.resolve(tcp_server.listen(port)).return(tcp_server);
+        return Promise.resolve(tcp_server.listen(port))
+            .then(() => tcp_server);
     }
 
 
@@ -917,7 +918,8 @@ class RPC extends EventEmitter {
         dbg.log0('RPC register_ntcp_transport');
         const ntcp_server = new RpcNtcpServer(tls_options);
         ntcp_server.on('connection', conn => this._accept_new_connection(conn));
-        return P.resolve(ntcp_server.listen(port)).return(ntcp_server);
+        return Promise.resolve(ntcp_server.listen(port))
+            .then(() => ntcp_server);
     }
 
 
--- a/src/rpc/rpc_http_server.js
+++ b/src/rpc/rpc_http_server.js
@@ -53,7 +53,7 @@ class RpcHttpServer extends events.EventEmitter {
             http.createServer();
         this.install_on_server(server, options.default_handler);
         return P.fromCallback(callback => server.listen(port, callback))
-            .return(server);
+            .then(() => server);
     }
 
     /**
--- a/src/rpc/stun.js
+++ b/src/rpc/stun.js
@@ -5,6 +5,7 @@
 var _ = require('lodash');
 var util = require('util');
 var P = require('../util/promise');
+const promise_utils = require('../util/promise_utils');
 var url = require('url');
 // var util = require('util');
 var dgram = require('dgram');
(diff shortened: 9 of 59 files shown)


guymguym (changes requested):
@liranmauda That's a good progress indeed! 

But since you are replacing a bunch of functions that used to return Bluebird's to return native Promise instead, we risk that callers that depend on Bluebird instance methods will fail because a native Promise is returned somewhere in the middle of the promise chain.

So I think with this PR we also need to go over \*\*all\*\* the calls to Bluebird methods and either eliminate them or make sure they are scoped to a locally constructed Bluebird promise. Notice that we don't need to go over static functions like P.fromCallback, only methods that exist on a Bluebird instance) - 

```javascript
> native_methods = Object.getOwnPropertyNames(Promise.prototype)
[ 'constructor', 'then', 'catch', 'finally' ]

> Object.getOwnPropertyNames(require('bluebird').prototype).filter(n => !n.startsWith('_') && !native_methods.includes(n))
[
  'suppressUnhandledRejections',
  'lastly',
  'tap',
  'tapCatch',
  'toString',
  'caught',
  'reflect',
  'done',
  'spread',
  'toJSON',
  'all',
  'error',
  'bind',
  'cancel',
  'break',
  'isCancellable',
  'thenReturn',
  'return',
  'thenThrow',
  'throw',
  'catchThrow',
  'catchReturn',
  'isCancelled',
  'isPending',
  'isRejected',
  'isFulfilled',
  'isResolved',
  'value',
  'reason',
  'call',
  'get',
  'map',
  'nodeify',
  'asCallback',
  'props',
  'race',
  'reduce',
  'settle',
  'some',
  'delay',
  'timeout',
  'disposer',
  'any',
  'each',
  'mapSeries',
  'filter'
]
```

guymguym:
This gets the list 
```bash
$ RE=$(node -p '
  Object.keys(require("bluebird").prototype)
  .filter(n => !n.startsWith("_") && ![
    "then", "catch", "finally",                       // Promise
    "filter", "map", "reduce", "some", "any", "all",  // Array
    "each", "get",                                    // lodash
    "error",                                          // console/dbg
    "call", "bind",                                   // Function
    "toString", "toJSON",                             // Object
  ].includes(n))
  .map(n=>`\\.${n}\\(`)
  .join("|")
')
$ rg \'$RE\' src | egrep -v '^src/test/' | wc -l
     161
$ rg --only-matching \'$RE\' src | egrep -v '^src/test/' | cut -d':' -f2- | sort | uniq -c | sort -n
   1 .cancel(
   1 .isPending(
   1 .isRejected(
   2 .props(
  10 .tap(
  12 .spread(
  37 .timeout(
  47 .return(
  50 .delay(
```
So most of the bluebird methods we use are `delay, return, timeout, spread, tap...` etc.
If you eliminate those 160 callers it will become much easier to replace bluebird promises with native promises because we won't worry that callers may fail.

BTW 1 - notice that I filtered out `src/tests/` assuming that we should detect test errors by failures.
BTW 2 - I filtered quite a lot of known function names that collide with other objects such as arrays, so it might still be that we used one of those bluebird functions somewhere, but that's not very likely.

liranmauda:
> If you eliminate those 160 callers it will become much easier to replace bluebird promises with native promises because we won't worry that callers may fail.

by eliminate you mean  do not replace to native where we call those functions (or functions that return bluebird that are chained)?

liranmauda:
.cancel() - chained to ClientRequest which we didn’t touced. 
.isPending() - chained to RpcRequest which we didn’t touced. 
.isRejected() - doesn't look like we touched.
.props() - all is chained to P. 
.tap():

Being called here: 
https://github.com/noobaa/noobaa-core/blob/master/src/endpoint/ftp/ftp_filesystem.js#L93
And here:
https://github.com/noobaa/noobaa-core/blob/master/src/endpoint/ftp/ftp_filesystem.js#L126
Which are not bluebird (on master) 

.spread() - is called mostly from P.join and once from P.fromCallback
all the other was looked at the first commit :) 
Hope we didn't missed anything.

guymguym:
> by eliminate you mean do not replace to native where we call those functions (or functions that return bluebird that are chained)?

@liranmauda No I meant that we need to remove those references to Bluebird methods so that we are more certain that we can't crash on that.

I think we just need to replace those with some other simple one liners or utility, for example:
```
Promise.resolve()
    .then(() => 42)
    .delay(1000)
    .then(console.log)
```
can be replaced to 
```
Promise.resolve()
    .then(() => 42)
    .then(promise_utils.delay(1000))
    .then(console.log)
```

and get rid also from: .spread() .timeout() .return().

liranmauda:
- Removed all .tap() and .return() 
- Removed most of P.join 
- Removed most of .spread()

guymguym:
You should probably remove all of .spread calls because it will fail on native promises and is risky to leave it, and we can easily replace it with array param destructoring..

guymguym:
What about .timeout?

liranmauda:
> You should probably remove all of .spread calls because it will fail on native promises and is risky to leave it, and we can easily replace it with array param destructoring..

Yes, I removed most in some of the commits but I fail on unit tests (only on travis) so I am removing and adding commits.
In later commit i left only 4 .spread 

> What about .timeout? we can replace it with a promise_utils and get rid of those bluebird calls

Still didn't figure out how to replace .timeout

liranmauda:
removed the delay chang, Will add it once I figure out why tarvis fails on it :)

guymguym:
BTW this also means to change those catch timeout such as:
https://github.com/noobaa/noobaa-core/blob/9b6ec6bf3263d2cb531ec2237f73a0128c9bf184/src/rpc/rpc_base_conn.js#L128-L135

because this is again a non-standard catch variant of bluebird.

so with promise_utils.timeout we can write:

```javascript
try {
  const encoded_message = this._encode_message(msg);
  const send_promise = this._send(encoded_message, msg.body && msg.body.op, req);
  const res = await promise_utils.timeout(config.RPC_SEND_TIMEOUT, send_promise);
  return res;
} catch(err) {
  if (err instanceof promise_utils.TimeoutError) {
    this.emit_error(new RpcError('RPC_SEND_TIMEOUT', 'RPC SEND TIMEOUT'));
  } else {
    this.emit_error(err);
  }
}
```

src/agent/agent_cli.js
```diff
@@ -329,7 +329,7 @@ class AgentCLI {
             current_path = current_path.replace('./', '');
             //hiding storage folder
             return child_process.execAsync('attrib +H ' + current_path)
-                .then(() => P.join(os_utils.is_folder_permissions_set(current_path), fs.readdirAsync(current_path)))
+                .then(() => Promise.all([os_utils.is_folder_permissions_set(current_path), fs.readdirAsync(current_path)]))
                 .spread(function(permissions_set, noobaa_storage_initialization) {
```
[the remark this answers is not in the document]
liranmauda: Converted all the `.spreds()` into this: 
```
.then(function(res) {
                    let permissions_set = res[0];
                    let noobaa_storage_initialization = res[1];
...
```
should I refractor it to just `.then( ([ arg1, arg2 ]) => { ... })`?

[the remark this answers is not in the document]
guymguym: yes - .then( ([ arg1, arg2 ]) => { ... })

src/util/promise_utils.js
```diff
@@ -368,7 +381,9 @@ function map_values(obj, func) {
                 new_obj[key] = res;
             })
         ))
-        .return(new_obj);
+        .then(function() {
+            return new_obj;
+        });
```
guymguym, line 386:
```suggestion
        .then(() => new_obj);
```

liranmauda: sure, I will do it for all those entries.

src/agent/agent_diagnostics.js
```diff
@@ -68,7 +68,9 @@ function collect_agent_diagnostics(storage_path) {
                 });
         })
         .finally(() => diag_log.end())
-        .return()
+        .then(() => {
+            // do nothing.
```
guymguym, line 72: I think it's important to note that we "return undefined" and not that we do nothing.
It is even better be explicit like `.then(() => undefined)`

src/rpc/ice.js
```diff
@@ -1496,11 +1496,13 @@ function listen_on_port_range(port_range) {
         server.listen(port);
         // wait for listen even, while also watching for error/close.
         return promise_utils.wait_for_event(server, 'listening')
-            .return(server)
+            .then(function() {
+                return server;
+            })
```
guymguym, line 1501: you don't want to define functions in promise chains because any access to `this` will not work, so prefer arrow functions in chains:
```suggestion
            .then(() => server)
```

src/endpoint/ftp/ftp_filesystem.js
```diff
@@ -90,7 +90,10 @@ class FtpFileSystemNB {
                 delimiter: '/',
                 limit: 100
             }))
-            .tap(console.log)
+            .then(res => {
+                console.log(res);
+                return res;
+            })
             .then(res => {
                 console.log(`in get - got res = `, res);
```
guymguym, line 98: just delete that then because the next one prints anyhow

src/agent/block_store_services/block_store_fs.js
```diff
@@ -61,16 +61,18 @@ class BlockStoreFs extends BlockStoreBase {
...
                     dbg.error('got error from get_drive_of_path. checking if root_path exists', err.message);
                     this._test_root_path_exists();
                     throw err;
                 })
-            )
-            .spread((usage, drive) => {
+            ])
+            .then(res => {
+                let usage = res[0];
+                let drive = res[1];
```
guymguym, line 75:
```suggestion
            .then(([usage, drive]) => {
```

src/agent/block_store_services/block_store_fs.js
```diff
@@ -218,7 +220,10 @@ class BlockStoreFs extends BlockStoreBase {
                 this._usage = usage; // object with properties size and count
                 // update usage file
                 let usage_data = JSON.stringify(this._usage);
-                return fs.writeFileAsync(this.usage_path, usage_data).return(usage);
+                return fs.writeFileAsync(this.usage_path, usage_data)
+                    .then(function() {
+                        return usage;
+                    });
```
guymguym, line 226: use arrow
```suggestion
                    .then(() => usage);
```

src/rpc/rpc.js
```diff
@@ -904,7 +904,10 @@ class RPC extends EventEmitter {
         dbg.log0('RPC register_tcp_transport');
         const tcp_server = new RpcTcpServer(tls_options);
         tcp_server.on('connection', conn => this._accept_new_connection(conn));
-        return P.resolve(tcp_server.listen(port)).return(tcp_server);
+        return Promise.resolve(tcp_server.listen(port))
+            .then(function() {
+                return tcp_server;
+            });
```
guymguym, line 910: prefer arrow
```suggestion
            .then(() => tcp_server);
```

src/rpc/rpc.js
```diff
@@ -917,7 +920,10 @@ class RPC extends EventEmitter {
         dbg.log0('RPC register_ntcp_transport');
         const ntcp_server = new RpcNtcpServer(tls_options);
         ntcp_server.on('connection', conn => this._accept_new_connection(conn));
-        return P.resolve(ntcp_server.listen(port)).return(ntcp_server);
+        return Promise.resolve(ntcp_server.listen(port))
+            .then(function() {
+                return ntcp_server;
+            });
```
guymguym, line 926:
```suggestion
            .then(() => ntcp_server);
```

src/rpc/rpc_http_server.js
```diff
@@ -53,7 +53,9 @@ class RpcHttpServer extends events.EventEmitter {
             http.createServer();
         this.install_on_server(server, options.default_handler);
         return P.fromCallback(callback => server.listen(port, callback))
-            .return(server);
+            .then(function() {
+                return server;
+            });
```
guymguym, line 58: arrow...

src/rpc/stun.js
```diff
@@ -617,7 +618,7 @@ function test() {
                 return P.join(
                         P.ninvoke(socket, 'send', req, 0, req.length, stun_url.port, stun_url.hostname),
                         P.ninvoke(socket, 'send', ind, 0, ind.length, stun_url.port, stun_url.hostname))
-                    .delay(stun.INDICATION_INTERVAL * chance.floating(stun.INDICATION_JITTER))
+                    .then(promise_utils.delay(stun.INDICATION_INTERVAL * chance.floating(stun.INDICATION_JITTER)))
```
guymguym, line 621: OK this is a pitfall.
You must call the delay only when you actually need to start waiting, which is inside the promise chain, and not from the outside function. So please make sure to always use like this `.then(() => promise_utils.delay(555))`

```suggestion
                    .then(() => promise_utils.delay(stun.INDICATION_INTERVAL * chance.floating(stun.INDICATION_JITTER)))
```

guymguym: BTW notice that for all the cases you use it with await there is no need to use an arrow function because that's only needed for `.then(() => ...)`

src/server/analytic_services/activity_log_store.js
```diff
@@ -44,13 +44,15 @@ class ActivityLogStore {
             .then(() => this._activitylogs.validate(activity_log))
             .then(() => this._activitylogs.col().insertOne(activity_log))
             .catch(err => mongo_utils.check_duplicate_key_conflict(err, 'audit_log'))
-            .return(activity_log);
+            .then(function() {
+                return activity_log;
+            });
```
guymguym, line 49: arrow
```suggestion
            .then(() => activity_log);
```

src/server/bg_services/agent_blocks_verifier.js
```diff
@@ -108,7 +108,9 @@ class AgentBlocksVerifier {
                         });
                 });
             })
-            .return();
+            .then(() => {
+                // do nothing.
```
guymguym, line 112:
```suggestion
                // return nothing. 
```

src/server/bg_services/cluster_hb.js
```diff
@@ -60,8 +60,12 @@ function do_heartbeat({ skip_server_monitor } = {}) {
...
                 }),
                 os_utils.read_drives(),
-                os_utils.get_raw_storage()))
-            .spread((mongo_status, drives, raw_storage) => {
+                os_utils.get_raw_storage()
+            ]))
+            .then(res => {
+                let mongo_status = res[0];
+                let drives = res[1];
+                let raw_storage = res[2];
```
guymguym, line 68:
```suggestion
            .then(([mongo_status, drives, raw_storage]) => {
```

src/agent/agent_diagnostics.js
liranmauda: sure, I will do it for all those entries.

src/server/bg_services/cluster_hb.js
```diff
@@ -119,11 +123,13 @@ function do_heartbeat({ skip_server_monitor } = {}) {
                             `^Server .*${current_clustering.owner_secret} configuration is below minimum requirements.*`
                         ));
                 }
-                return P.resolve();
+                return Promise.resolve();
             })
-            .return();
+            .then(() => {
+                // do nothing.
```
guymguym, line 129:
```suggestion
                // return nothing. 
```

src/server/bg_services/md_aggregator.js
```diff
@@ -254,21 +254,21 @@ function range_md_aggregator({
...
-        )
-        .spread((
-            existing_chunks_aggregate,
-            deleted_chunks_aggregate,
-            existing_objects_aggregate,
-            deleted_objects_aggregate,
-            existing_blocks_aggregate,
-            deleted_blocks_aggregate) => {
+        ])
+        .then(res => {
```
guymguym, line 265: .then(([
 ...
]) => {

liranmauda:
now we left with:
```
$ rg --only-matching \'$RE\' src | egrep -v '^src/test/' | cut -d':' -f2- | sort | uniq -c | sort -n
   1 .cancel(
   1 .isPending(
   1 .isRejected(
   1 .spread(
   2 .props(
  37 .timeout(
  46 .delay(
```
where delay number is way less (guess 2), as it counts it wrongly :)

src/util/promise_utils.js
```diff
@@ -62,7 +64,8 @@ function iterate(array, func) {
         return P.try(() => func(array[i], i, array)).then(next);
     }
 
-    return P.try(next).return(results);
+    return P.try(next)
+        .then(() => results);
```
guymguym, line 68:
```suggestion
    return Promise.resolve().then(next).then(() => results);
```

guymguym: we don't want bluebird here..

liranmauda: sure, in that function we still have 2 blubirds:
https://github.com/noobaa/noobaa-core/blob/9b6ec6bf3263d2cb531ec2237f73a0128c9bf184/src/util/promise_utils.js#L39
and 
https://github.com/noobaa/noobaa-core/blob/9b6ec6bf3263d2cb531ec2237f73a0128c9bf184/src/util/promise_utils.js#L62

should I leave them?

src/server/system_services/cluster_server.js
```diff
@@ -186,13 +186,16 @@ function add_member_to_cluster_invoke(req, my_address) {
...
                 fs.readFileAsync(config.MONGO_DEFAULTS.SERVER_CERT_PATH, 'utf8'),
                 fs.readFileAsync(config.MONGO_DEFAULTS.CLIENT_CERT_PATH, 'utf8')
-            );
+            ]);
         })
-        .spread((root_ca, server_cert, client_cert) => {
+        .then(res => {
+            let root_ca = res[0];
+            let server_cert = res[1];
+            let client_cert = res[2];
```
guymguym, line 198:
```suggestion
        .then(([root_ca, server_cert, client_cert])=> {
```

src/server/system_services/stats_aggregator.js
```diff
@@ -569,7 +569,9 @@ function get_nodes_stats(req) {
                     account_id: support_account._id
                 })
             }))
-        .spread((nodes_results, hosts_results) => {
+        .then(res => {
+            let nodes_results = res[0];
+            let hosts_results = res[1];
```
guymguym, line 574:
```suggestion
        .then(([nodes_results, hosts_results])=> {
```

src/server/system_services/tier_server.js
```diff
@@ -101,11 +100,13 @@ function read_tier(req) {
...
-    return P.join(
+    return Promise.all([
             nodes_client.instance().aggregate_nodes_by_pool(_.compact(pool_names), req.system._id),
             nodes_client.instance().aggregate_data_free_by_tier([String(tier._id)], req.system._id)
-        )
-        .spread(function(nodes_aggregate_pool, available_to_upload) {
+        ])
+        .then(function(res) {
+            let nodes_aggregate_pool = res[0];
+            let available_to_upload = res[1];
```
guymguym, line 109:
```suggestion
        .then(([nodes_aggregate_pool, available_to_upload]) {
```

src/server/system_services/tier_server.js
```diff
@@ -538,11 +539,13 @@ function read_policy(req) {
...
-    return P.join(
+    return Promise.all([
             nodes_client.instance().aggregate_nodes_by_pool(pool_names, req.system._id),
             nodes_client.instance().aggregate_hosts_by_pool(pool_names, req.system._id),
-        )
-        .spread(function(nodes_aggregate_pool, hosts_aggregate_pool) {
+        ])
+        .then(function(res) {
+            let nodes_aggregate_pool = res[0];
+            let hosts_aggregate_pool = res[1];
```
guymguym, line 548:
```suggestion
        .then(([nodes_aggregate_pool, hosts_aggregate_pool]) {
```

src/server/system_services/stats_aggregator.js
liranmauda: when doing that we get a lint error:
`Unexpected block statement surrounding arrow body; move the returned value immediately after the =>.eslint(arrow-body-style)`

src/server/system_services/tier_server.js
liranmauda: when doing that we get a lint error:
`Unexpected block statement surrounding arrow body; move the returned value immediately after the =>.eslint(arrow-body-style)`

src/server/system_services/tier_server.js
liranmauda: when doing that we get a lint error:
`Unexpected block statement surrounding arrow body; move the returned value immediately after the =>.eslint(arrow-body-style)`

src/server/system_services/tier_server.js
guymguym: I was just missing the arrow `=>` Fixed the suggestion

src/server/system_services/tier_server.js
guymguym: I was just missing the arrow `=>` Fixed the suggestion

src/server/system_services/tier_server.js
liranmauda: @guymguym not it, I manually added the arrow and got this lint error
so the code is:
```
        .then(([nodes_aggregate_pool, available_to_upload]) => {
            return get_tier_info(tier, nodes_aggregate_pool, { mirror_storage: available_to_upload[String(tier._id)] });
        });
```
but lint doesn't like it

guymguym: because the linter is asking you to get rid of the return like this:
```javascript
        .then(([nodes_aggregate_pool, available_to_upload]) =>
            get_tier_info(tier, nodes_aggregate_pool, { mirror_storage: available_to_upload[String(tier._id)] })
        );
```

src/agent/agent.js
```diff
@@ -700,11 +704,12 @@ class Agent {
...
 
         this._test_connection_timeout = setTimeout(() => {
-            this.client.node.ping()
-                .timeout(MASTER_RESPONSE_TIMEOUT)
-                .then(() => {
-                    this._test_connection_timeout = null;
-                })
+            P.resolve(
+                    this.client.node.ping()
+                    .timeout(MASTER_RESPONSE_TIMEOUT)
```
guymguym, line 709:
```suggestion
                    Bluebird.resolve(this.client.node.ping())
                        .timeout(MASTER_RESPONSE_TIMEOUT)
```

guymguym:
@liranmauda Let me know when you rebase from master and push the updated code.

src/util/promise_utils.js
```diff
@@ -63,7 +65,7 @@ function iterate(array, func) {
         return P.try(() => func(array[i], i, array)).then(next);
```
guymguym, line 65: What about this P.try?

src/util/promise_utils.js
```diff
@@ -5,11 +5,13 @@ const _ = require('lodash');
 const child_process = require('child_process');
 
 const P = require('./promise');
+const util = require('util');
 const dbg = require('../util/debug_module')(__filename);
 const Semaphore = require('./semaphore');
 
 require('setimmediate');
 
+const setTimeoutAsync = util.promisify(setTimeout);
```
guymguym, line 14: so how about just defining
```
const delay = util.promisify(setTimeout);
```

src/util/promise_utils.js
```diff
@@ -169,18 +179,18 @@ function next_tick() {
 // for the sake of tests to be able to exit we schedule the worker with unblocking delay
 // so that it won't prevent the process from existing if it's the only timer left
 function run_background_worker(worker) {
-    var DEFUALT_DELAY = 10000;
+    var DEFAULT_DELAY = 10000;
```
guymguym, line 182: const

src/agent/block_store_services/block_store_fs.js
```diff
@@ -92,7 +92,7 @@ class BlockStoreFs extends BlockStoreBase {
         return P.join(
```
guymguym, line 92: What about this P.join?

guymguym (changes requested):
@liranmauda I'm not sure what this currently includes.
I see more P.try and P.join left behind.
Why is that?

guymguym:
@liranmauda What's the status? Is it ready?

liranmauda:
@guymguym 
It can be reviewed. 
so there was some more `P.join` that I replaced and found some more `P.delay` that I have missed.
where bluebird is needed We wrapped it with a bluebird call.

P.resolve is still bluebird, we will hijack it Next PR when we will handle the timeout and defer (what do you think?)

guymguym:
OK I think P.resolve should be completely replaced to either Promise.resolve or Bluebird.resolve.

guymguym (approved):

PR #6125, opened 2020-08-12
Remove top level code that uses util.promisify in FE imported files

nb-ohad:
### Explain the changes
`util.promisify` is not supported in the browser and cannot be used by top-level code in files that are imported by FE code.
Doing so will not allow the browser to execute the FE code bundle.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -15,7 +15,6 @@ const P = require('./promise');
 const dbg = require('./debug_module')(__filename);
 const xml_utils = require('./xml_utils');
 const cloud_utils = require('./cloud_utils');
-const util = require('util');
 
 const { HTTP_PROXY, HTTPS_PROXY, NO_PROXY } = process.env;
 const http_agent = new http.Agent();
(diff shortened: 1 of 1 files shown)


src/util/http_utils.js
```diff
@@ -248,7 +247,9 @@ function read_request_body(req, options) {
     });
 }
 
-const parse_xml_to_js = util.promisify(xml2js.parseString);
+function parse_xml_to_js(body, xml_options) {
+    return P.fromCallback(callback => xml2js.parseString(body, xml_options, callback));
```
guymguym, line 251: apparently there's promise support already - 
https://github.com/Leonidas-from-XIV/node-xml2js#promise-usage

```suggestion
    return xml2js.parseStringPromise(body, xml_options);
```

guymguym (changes requested):

src/util/http_utils.js
nb-ohad: Done, but a little bit differently. 
Just pulled the actual `parseStringPromise` into a const named `parse_xml_to_js`.

guymguym (approved):

PR #6128, opened 2020-08-17
NS-CACHE: support multi deletes operation

paulxuzhang:
### Explain the changes
1.  Support multi deletes procedure in namespace caching

1. Part of the namespace caching feature

### Testing Instructions:
1.  Delete multiple objects from namespace bucket enabled with caching, objects shall be deleted from both cache and hub.
2. If error happens in cache or hub delete operations, return error to client.

### DIFF
--- a/src/sdk/namespace_cache.js
+++ b/src/sdk/namespace_cache.js
@@ -583,29 +583,37 @@ class NamespaceCache {
     }
 
     async delete_multiple_objects(params, object_sdk) {
-        const deleted_res = await this.namespace_hub.delete_multiple_objects(params, object_sdk);
-
         const operation = 'ObjectRemoved';
         const load_for_trigger = object_sdk.should_run_triggers({ active_triggers: this.active_triggers, operation });
+        const head_res = load_for_trigger && await P.map(params.objects, async obj => {
+            const request = {
+                bucket: params.bucket,
+                key: obj.key,
+                version_id: obj.version_id
+            };
+            let obj_md;
+            try {
+                obj_md = _.defaults({ key: obj.key }, await this.namespace_hub.read_object_md(request, object_sdk));
+            } catch (err) {
+                if (err.rpc_code !== 'NO_SUCH_OBJECT') throw err;
+            }
+            return obj_md;
+        });
+
+        const [hub_res, cache_res] = await Promise.allSettled([
+            this.namespace_hub.delete_multiple_objects(params, object_sdk),
+            this.namespace_nb.delete_multiple_objects(params, object_sdk),
+        ]);
+        if (hub_res.status === 'rejected') {
+            throw hub_res.reason;
+        }
+        if (cache_res.status === 'rejected') {
+            throw cache_res.reason;
+        }
+
         if (load_for_trigger) {
-
-            const head_res = await P.map(params.objects, async obj => {
-                const request = {
-                    bucket: params.bucket,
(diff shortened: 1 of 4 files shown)


src/sdk/namespace_cache.js
```diff
@@ -583,7 +583,16 @@ class NamespaceCache {
     }
 
     async delete_multiple_objects(params, object_sdk) {
-        const deleted_res = await this.namespace_hub.delete_multiple_objects(params, object_sdk);
+        const [hub_res, cache_res] = await Promise.allSettled([
+            this.namespace_hub.delete_multiple_objects(params, object_sdk),
```
jeniawhite, line 587: @paulxuzhang 
This breaks the trigger logic.
If you have a trigger it will not work properly.
Notice that later on, you run a query:
```obj_md = _.defaults({ key: obj.key }, await this.namespace_hub.read_object_md(request, object_sdk));```
There is no way that this query will find the object_md since you've already deleted it in ```Promise.allSettled```.
Unless I'm missing something.

File: src/test/utils/namespace_context.js [hunk 2]
```diff
@@ -137,6 +137,12 @@ class NamespaceContext {
         }
     }
 
+    async delete_files_from_noobaa(type, file_names) {
```
jeniawhite, line 140: @paulxuzhang 
I would also consider adding support and tests for version_ids and delete_markers.
This is a general comment not necessarily for this function.

jeniawhite (changes requested):

src/sdk/namespace_cache.js
paulxuzhang: You are right. I will fix.

src/test/utils/namespace_context.js
paulxuzhang: version id support is done in different PR. I will see what I can do.

paulxuzhang: Actually per the design, we will reject any request with version Id for namespace bucket enabled with caching. I will not add the relevant tests in this PR. @guptsanj has been working on handling version id. He can add relevant tests on handling version id.

src/sdk/namespace_cache.js
paulxuzhang: @jeniawhite Fixed. Could you verify?

jeniawhite: @paulxuzhang 
Looks valid. 
Please run the code and make sure that it works, thank you.

Side question:
Is there a reason why you are reading the object_md from ```namespace_hub``` and not from ```namespace_nb``` or a merge between them?
Wouldn't a read from ```namespace_nb``` be quicker?

src/test/utils/namespace_context.js
jeniawhite: @paulxuzhang 
Alright, I got you.
Just making sure that we won't forget to cover that part.
Since delete_markers and version_ids are a significant portion of the multiple_delete flow logic.

jeniawhite (approved):

src/sdk/namespace_cache.js
paulxuzhang: Object gets cached only when it is read or uploaded through noobaa endpoint. Object can be uploaded or updated directly against hub without going through noobaa endpoint. So cache does not always have all the objects, hub is the ultimate truth.

guymguym (approved):

PR #6129, opened 2020-08-17
NS-CACHE: Get object support partNumber query parameter and preconditions

paulxuzhang:

### Explain the changes
1. If partNumber query parameter is set in get object request, bypass cache and proxy the request to hub. More info about partNumber can be found at
https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax

2. If the precondition headers (e.g. if-match) are set in get object request, evaluate the conditions on cache if cache ttl is expired, or proxy the request with the conditions to hub if object not cached or cache ttl has expired.

1.  Part of the namespace caching feature.

### Testing Instructions:
1. See the above description on the changes.

### DIFF
--- a/src/endpoint/s3/ops/s3_head_object.js
+++ b/src/endpoint/s3/ops/s3_head_object.js
@@ -4,6 +4,7 @@
 // const S3Error = require('../s3_errors').S3Error;
 const s3_utils = require('../s3_utils');
 const http_utils = require('../../../util/http_utils');
+const S3Error = require('../s3_errors').S3Error;
 
 /**
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectHEAD.html
@@ -17,6 +18,9 @@ async function head_object(req, res) {
         md_conditions: http_utils.get_md_conditions(req),
         encryption
     };
+    if (req.query.partNumber) {
+        params.part_number = s3_utils.parse_part_number(req.query.partNumber, S3Error.InvalidArgument);
+    }
     if (req.query.get_from_cache !== undefined) {
         params.get_from_cache = true;
     }
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -263,7 +263,7 @@ function set_response_object_md(res, object_md) {
     res.setHeader('ETag', '"' + object_md.etag + '"');
     res.setHeader('Last-Modified', time_utils.format_http_header_date(new Date(object_md.create_time)));
     res.setHeader('Content-Type', object_md.content_type);
-    res.setHeader('Content-Length', object_md.size);
+    res.setHeader('Content-Length', object_md.content_length === undefined ? object_md.size : object_md.content_length);
     res.setHeader('Accept-Ranges', 'bytes');
     if (config.WORM_ENABLED && object_md.lock_settings) {
         if (object_md.lock_settings.legal_hold) {
@@ -277,6 +277,8 @@ function set_response_object_md(res, object_md) {
     if (object_md.version_id) res.setHeader('x-amz-version-id', object_md.version_id);
     set_response_xattr(res, object_md.xattr);
     if (object_md.tag_count) res.setHeader('x-amz-tagging-count', object_md.tag_count);
+    if (object_md.multipart_count) res.setHeader('x-amz-mp-parts-count', object_md.multipart_count);
+    if (object_md.content_range) res.setHeader('Content-Range', object_md.content_range);
     return object_md;
 }
 
--- a/src/sdk/namespace_cache.js
+++ b/src/sdk/namespace_cache.js
@@ -179,31 +179,36 @@ class NamespaceCache {
         let object_info_cache = null;
         let cache_etag = '';
         const get_from_cache = params.get_from_cache;
-        try {
-            // Remove get_from_cache if exists for maching RPC schema
-            params = _.omit(params, 'get_from_cache');
-            object_info_cache = await this.namespace_nb.read_object_md(params, object_sdk);
-            if (get_from_cache) {
-                dbg.log0('NamespaceCache.read_object_md get_from_cache is enabled', object_info_cache);
-                object_info_cache.should_read_from_cache = true;
-                return object_info_cache;
+        // part_number is set to the query parameter partNumber in s3 request. If set,
+        // it should be a positive integer between 1 and 10,000. We will bypass cache
+        // and proxy request to hub.
+        if (!params.part_number) {
+            // partNumber is not set
+            try {
+                // Remove get_from_cache if exists for maching RPC schema
+                params = _.omit(params, 'get_from_cache');
+                object_info_cache = await this.namespace_nb.read_object_md(params, object_sdk);
+                if (get_from_cache) {
+                    dbg.log0('NamespaceCache.read_object_md get_from_cache is enabled', object_info_cache);
+                    object_info_cache.should_read_from_cache = true;
+                    return object_info_cache;
+                }
+
+                const cache_validation_time = object_info_cache.cache_last_valid_time;
+                const time_since_validation = Date.now() - cache_validation_time;
+
+                if ((this.caching.ttl_ms > 0 && time_since_validation <= this.caching.ttl_ms) || this.caching.ttl_ms < 0) {
+                    object_info_cache.should_read_from_cache = true; // mark it for read_object_stream
+                    dbg.log0('NamespaceCache.read_object_md use md from cache', object_info_cache);
+                    return object_info_cache;
+                }
+
+                cache_etag = object_info_cache.etag;
+            } catch (err) {
+                dbg.log0('NamespaceCache.read_object_md: error in cache', err);
+                if (get_from_cache) throw err;
             }
-
-            const cache_validation_time = object_info_cache.cache_last_valid_time;
-            const time_since_validation = Date.now() - cache_validation_time;
-
-            if ((this.caching.ttl_ms > 0 && time_since_validation <= this.caching.ttl_ms) || this.caching.ttl_ms < 0) {
-                object_info_cache.should_read_from_cache = true; // mark it for read_object_stream
-                dbg.log0('NamespaceCache.read_object_md use md from cache', object_info_cache);
-                return object_info_cache;
-            }
-
-            cache_etag = object_info_cache.etag;
-        } catch (err) {
-            dbg.log0('NamespaceCache.read_object_md: error in cache', err);
-            if (get_from_cache) throw err;
         }
-
         try {
             const object_info_hub = await this.namespace_hub.read_object_md(params, object_sdk);
             if (object_info_hub.etag === cache_etag) {
@@ -334,7 +339,9 @@ class NamespaceCache {
 
         let hub_read_stream;
         try {
-            hub_read_params.if_match_etag = params.object_md.etag;
+            if (!hub_read_params.md_conditions) {
+                hub_read_params.md_conditions = { if_match_etag: params.object_md.etag };
+            }
             hub_read_stream = await this.namespace_hub.read_object_stream(hub_read_params, object_sdk);
         } catch (err) {
             if (err.rpc_code === 'IF_MATCH_ETAG') {
@@ -357,7 +364,8 @@ class NamespaceCache {
             hub_read_stream.pipe(range_stream);
         }
 
-        // Object or part will only be uploaded to cache if size is not too big
+        // Object or part will only be uploaded to cache if size is not too big and
+        // the preconditions (if-match header etc.) are not set.
         if (hub_read_size <= params.bucket_free_space_bytes) {
             // We use pass through stream here because we have to start piping immediately
             // and the cache upload does not pipe immediately (only after creating the object_md).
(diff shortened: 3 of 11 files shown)


src/endpoint/s3/s3_utils.js
```diff
@@ -277,6 +277,7 @@ function set_response_object_md(res, object_md) {
     if (object_md.version_id) res.setHeader('x-amz-version-id', object_md.version_id);
     set_response_xattr(res, object_md.xattr);
     if (object_md.tag_count) res.setHeader('x-amz-tagging-count', object_md.tag_count);
+    if (object_md.part_count) res.setHeader('x-amz-mp-parts-count', object_md.part_count);
```
guymguym, line 280: I don't want to confuse our internal structure between our own object parts count and the multipart count so I suggest to rename this `object_md.multipart_count`

src/sdk/namespace_s3.js
```diff
@@ -119,16 +119,30 @@ class NamespaceS3 {
...
+                request.IfUnmodifiedSince = new Date(params.md_conditions.if_unmodified_since);
+            }
+        }
+    }
+
     async read_object_md(params, object_sdk) {
         try {
             dbg.log0('NamespaceS3.read_object_md:', this.bucket, inspect(params));
-            const request = { Key: params.key, Range: `bytes=0-${config.INLINE_MAX_SIZE - 1}` };
+            const request = { Key: params.key, Range: `bytes=0-${config.INLINE_MAX_SIZE - 1}`, PartNumber: params.part_number };
```
guymguym, line 138: Do you think that part_number makes sense here in the inline read?
Seems like this will either be ignored with Range or have some other unexpected behavior...

src/sdk/namespace_s3.js
```diff
@@ -523,13 +538,23 @@ class NamespaceS3 {
...
     }
 
-    _translate_error_code(err) {
+    _translate_error_code(params, err) {
         if (err.code === 'NoSuchKey') err.rpc_code = 'NO_SUCH_OBJECT';
-        if (err.code === 'NotFound') err.rpc_code = 'NO_SUCH_OBJECT';
-        if (err.code === 'PreconditionFailed') err.rpc_code = 'IF_MATCH_ETAG';
+        else if (err.code === 'NotFound') err.rpc_code = 'NO_SUCH_OBJECT';
+        else {
+            const md_conditions = params.md_conditions;
```
guymguym, line 549: Don't you need check that md_conditions is defined?
```suggestion
        else if (params.md_conditions) {
            const md_conditions = params.md_conditions;
```

src/sdk/namespace_cache.js
```diff
@@ -401,6 +407,13 @@ class NamespaceCache {
 
 
     async read_object_stream(params, object_sdk) {
+        if (params.part_number) {
```
guymguym, line 410: What about part_number===0? I think valid multipart part_number starts from 1 but even then its better to add an explicit comment that will not be confusing to read.

src/sdk/namespace_cache.js
```diff
@@ -357,8 +362,9 @@ class NamespaceCache {
             hub_read_stream.pipe(range_stream);
         }
 
-        // Object or part will only be uploaded to cache if size is not too big
-        if (hub_read_size <= params.bucket_free_space_bytes) {
+        // Object or part will only be uploaded to cache if size is not too big and
+        // the preconditions (if-match header etc.) are not set.
+        if (hub_read_size <= params.bucket_free_space_bytes && _.isUndefined(params.md_conditions)) {
```
guymguym, line 367: simplify check:
```suggestion
        if (hub_read_size <= params.bucket_free_space_bytes && params.md_conditions) {
```

guymguym: But actually why do preconditions should prevent us from putting to cache at all?

src/sdk/namespace_cache.js
```diff
@@ -334,7 +337,9 @@ class NamespaceCache {
 
         let hub_read_stream;
         try {
-            hub_read_params.if_match_etag = params.object_md.etag;
+            if (_.isUndefined(hub_read_params.md_conditions)) {
```
guymguym, line 340:
```suggestion
            if (hub_read_params.md_conditions) {
```

src/sdk/namespace_cache.js
```diff
@@ -251,7 +252,9 @@ class NamespaceCache {
         const range_hub_read = await this._range_read_hub_check(params, object_sdk);
         if (!params.object_md.should_read_from_cache) {
             // Object not in cache
-            if (range_hub_read && params.read_size <= params.bucket_free_space_bytes) {
+            if (range_hub_read &&
+                params.read_size <= params.bucket_free_space_bytes &&
+                _.isUndefined(params.md_conditions)) {
```
guymguym, line 257:
```suggestion
                params.md_conditions) {
```

src/sdk/namespace_cache.js
```diff
@@ -179,31 +179,32 @@ class NamespaceCache {
...
-        try {
-            // Remove get_from_cache if exists for maching RPC schema
-            params = _.omit(params, 'get_from_cache');
-            object_info_cache = await this.namespace_nb.read_object_md(params, object_sdk);
-            if (get_from_cache) {
-                dbg.log0('NamespaceCache.read_object_md get_from_cache is enabled', object_info_cache);
-                object_info_cache.should_read_from_cache = true;
-                return object_info_cache;
-            }
+        if (!params.part_number) {
```
guymguym, line 182: Here too need to be careful with `part_number===0` which will fall out this condition - but since multipart numbers start from 1 then it should be ok, I would just add a comment to explain it because it might not be easily visible here.

guymguym (changes requested):

src/endpoint/s3/ops/s3_head_object.js
```diff
@@ -17,6 +19,9 @@ async function head_object(req, res) {
         md_conditions: http_utils.get_md_conditions(req),
         encryption
     };
+    if (!_.isUndefined(req.query.partNumber)) {
+        params.part_number = s3_utils.parse_part_number(req.query.partNumber, S3Error.InvalidArgument);
```
guymguym, line 23: Is this really supported in headObject?

src/sdk/namespace_cache.js
```diff
@@ -334,7 +337,9 @@ class NamespaceCache {
 
         let hub_read_stream;
         try {
-            hub_read_params.if_match_etag = params.object_md.etag;
+            if (_.isUndefined(hub_read_params.md_conditions)) {
+                hub_read_params.md_conditions = { if_match_etag: params.object_md.etag };
```
guymguym, line 341: Can you explain this logic? Is this the revalidation of an object with expired TTL? If so I would expect to send if_not_match instead so that hub will return an "error" if the object is the same instead of the data.

guymguym: I would also use if_modified in conjunction with it, for safer validation.

guymguym (commented):
@paulxuzhang Few small comment. 
Also will open a tech debt to re-validate the md_conditions cases.

src/sdk/namespace_cache.js
```diff
@@ -334,7 +339,9 @@ class NamespaceCache {
 
         let hub_read_stream;
         try {
-            hub_read_params.if_match_etag = params.object_md.etag;
+            if (hub_read_params.md_conditions === undefined) {
```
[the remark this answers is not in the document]
paulxuzhang: ok

src/sdk/namespace_nb.js
```diff
@@ -64,6 +64,8 @@ class NamespaceNB {
 
     read_object_md(params, object_sdk) {
         if (this.target_bucket) params = _.defaults({ bucket: this.target_bucket }, params);
+        // Noobaa bucket does not currrently support partNumber query parameter. Ignore it for now.
+        _.unset(params, 'part_number');
```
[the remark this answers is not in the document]
paulxuzhang: ok

src/sdk/namespace_s3.js
```diff
@@ -119,16 +119,34 @@ class NamespaceS3 {
...
+            }
+        }
+    }
+
     async read_object_md(params, object_sdk) {
         try {
             dbg.log0('NamespaceS3.read_object_md:', this.bucket, inspect(params));
-            const request = { Key: params.key, Range: `bytes=0-${config.INLINE_MAX_SIZE - 1}` };
+            const request = { Key: params.key, PartNumber: params.part_number };
+            // If set, part_number is positive integer from 1 to 10000.
```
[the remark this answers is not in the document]
paulxuzhang: ok

src/endpoint/s3/s3_utils.js
```diff
@@ -277,6 +277,7 @@ function set_response_object_md(res, object_md) {
     if (object_md.version_id) res.setHeader('x-amz-version-id', object_md.version_id);
     set_response_xattr(res, object_md.xattr);
     if (object_md.tag_count) res.setHeader('x-amz-tagging-count', object_md.tag_count);
+    if (object_md.multipart_count) res.setHeader('x-amz-mp-parts-count', object_md.multipart_count);
```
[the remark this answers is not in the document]
paulxuzhang: You are right. I will add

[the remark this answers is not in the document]
guymguym: So AWS return it? COS too? I was just wondering because it seemed logical but didn't see it in AWS docs.

src/sdk/namespace_nb.js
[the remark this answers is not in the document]
paulxuzhang: Actually setting params.part_number = undefined will cause invalid schema RPC error since the current schema validation treats undefined as that parameter is set. I guess we would have to use _,unset() @guymguym, dince part_number is set to undefined in s3_get_object.js if no partNumber query parameter. If you are still against calling _.unset(), perhaps because it is always called in namespace_nb.js, we can change s3_get_object.js to only set part_number when partNumber query is set. Made the suggest changes.

src/endpoint/s3/s3_utils.js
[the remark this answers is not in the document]
paulxuzhang: COS returns it. I think AWS returns it too since it is in their spec.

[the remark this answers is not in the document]
paulxuzhang: COS returns it. I think AWS returns it too. https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_ResponseSyntax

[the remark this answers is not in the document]
guymguym: so you're seeing that by calling `HEAD /key?partNumber=17` I can get back the Content-Range header that tells me what is the object range of that part? Is there a way to list the ranges of all parts like I can do for an ongoing multipart upload with listParts() S3 call?

src/sdk/namespace_s3.js
```diff
@@ -513,7 +534,7 @@ class NamespaceS3 {
             obj_id: res.UploadId || etag,
             bucket: bucket,
             key: res.Key,
-            size: ranges || res.ContentLength || res.Size || 0,
+            size: res.ContentLength || ranges || res.Size || 0,
```
guymguym, line 537: Wait, is that safe??

paulxuzhang: Ran into issue with content-length in noobaa's response header to s3 client when partNumber presents. The changes is outdated. I added content_length in the return of the function.

src/endpoint/s3/s3_utils.js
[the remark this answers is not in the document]
paulxuzhang: I don't think there is. I could not find it in the spec.

src/endpoint/s3/s3_utils.js
```diff
@@ -263,7 +263,7 @@ function set_response_object_md(res, object_md) {
     res.setHeader('ETag', '"' + object_md.etag + '"');
     res.setHeader('Last-Modified', time_utils.format_http_header_date(new Date(object_md.create_time)));
     res.setHeader('Content-Type', object_md.content_type);
-    res.setHeader('Content-Length', object_md.size);
+    res.setHeader('Content-Length', object_md.content_length === undefined ? object_md.size : object_md.content_length);
```
guymguym, line 266: Ok so now this logic moved here, but I still don't understand the case when the size property is not set the way we need, can you explain?

src/sdk/namespace_nb.js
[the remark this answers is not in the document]
guymguym: ok

src/endpoint/s3/s3_utils.js
paulxuzhang: Content-Length is not always the size of object. In the case of range read or read with partNumber, it is the length of range to be read.

src/sdk/namespace_cache.js
guymguym: opened issue https://github.com/noobaa/noobaa-core/issues/6146. resolving.

src/endpoint/s3/s3_utils.js
guymguym: So you're saying that this calculation below is not working correct with partNumber? Can you explain? Maybe we need to adjust that code.
https://github.com/noobaa/noobaa-core/blob/abbafc463045736a4a2be973b1e15c282cb41d97/src/sdk/namespace_s3.js#L532-L537

paulxuzhang: The format of Content-Range is like 'content-range': 'bytes 100200-280000/3145735'. Given the example, the line at 532 returns 3145735 which is the size of entire object.

guymguym: will defer to new issue.

guymguym (approved):

PR #6130, opened 2020-08-17
NS-CACHE: Support S3 operations for cache enabled buckets

guptsanj:
### Explain the changes
1. Reject any versioning request to buckets with cache property
2. Server Side copy for a cache miss for a source bucket different than the target bucket
3. Multipart upload copy from a same/different sources
4. List objects from cache/hub
5. List Object versions defaults to List objects for cache enabled buckets
6. List multipart uploads from hub
7. Handle add/remove/get object tags from cache/hub buckets
8. Store last modified time from hub for cached objects. Return this value as the last modified time with cached objects.
Update Unit test

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/endpoint/s3/ops/s3_get_bucket.js
+++ b/src/endpoint/s3/ops/s3_get_bucket.js
@@ -30,7 +30,11 @@ async function get_bucket(req) {
             (cont_tok_to_key_marker(cont_tok) || start_after) : req.query.marker,
     };
 
+    if (req.query.get_from_cache) {
+        params.get_from_cache = req.query.get_from_cache;
+    }
     const reply = await req.object_sdk.list_objects(params);
+
     let res_params = {
         'Name': req.params.bucket,
         'Prefix': req.query.prefix,
@@ -41,7 +45,8 @@ async function get_bucket(req) {
     if (req.query.delimiter !== '') {
         res_params.Delimiter = req.query.delimiter;
     }
-
+    // Always have last_modified_time take precedence. This time is set only for cached objects.
+    // Non cached objects will always default to obj.create_time
     return {
         ListBucketResult: [res_params, list_type === '2' ? {
                 'ContinuationToken': cont_tok,
@@ -55,7 +60,8 @@ async function get_bucket(req) {
             _.map(reply.objects, obj => ({
                 Contents: {
                     Key: obj.key,
-                    LastModified: s3_utils.format_s3_xml_date(obj.create_time),
+                    // if the object specifies last_modified_time we use it, otherwise take create_time.
+                    LastModified: s3_utils.format_s3_xml_date(obj.last_modified_time || obj.create_time),
                     ETag: `"${obj.etag}"`,
                     Size: obj.size,
                     Owner: (!list_type || req.query['fetch-owner']) && s3_utils.DEFAULT_S3_USER,
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -152,6 +152,16 @@ async function handle_request(req, res) {
     usage_report.s3_usage_info.total_calls += 1;
     usage_report.s3_usage_info[op_name] = (usage_report.s3_usage_info[op_name] || 0) + 1;
 
+
+
+    if (req.query && req.query.versionId) {
+        const caching = await req.object_sdk.read_bucket_sdk_caching_info(req.params.bucket);
+        if (caching) {
+            dbg.error('S3 Version request not (NotImplemented) for buckets with caching', op_name, req.method, req.originalUrl);
+            throw new S3Error(S3Error.NotImplemented);
+        }
+    }
+
     const op = S3_OPS[op_name];
     if (!op || !op.handler) {
         dbg.error('S3 TODO (NotImplemented)', op_name, req.method, req.originalUrl);
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -261,7 +261,12 @@ function format_copy_source(copy_source) {
 
 function set_response_object_md(res, object_md) {
     res.setHeader('ETag', '"' + object_md.etag + '"');
-    res.setHeader('Last-Modified', time_utils.format_http_header_date(new Date(object_md.create_time)));
+
+    if (object_md.last_modified_time) {
+        res.setHeader('Last-Modified', time_utils.format_http_header_date(new Date(object_md.last_modified_time)));
+    } else {
+        res.setHeader('Last-Modified', time_utils.format_http_header_date(new Date(object_md.create_time)));
+    }
     res.setHeader('Content-Type', object_md.content_type);
     res.setHeader('Content-Length', object_md.content_length === undefined ? object_md.size : object_md.content_length);
     res.setHeader('Accept-Ranges', 'bytes');
@@ -549,6 +554,18 @@ function parse_website_to_body(website) {
     return reply;
 }
 
+function get_http_response_date(res) {
+    const r = get_http_response_from_resp(res);
+    if (!r.httpResponse.headers.date) throw new Error("date not found in response header");
+    return r.httpResponse.headers.date;
+}
+
+function get_http_response_from_resp(res) {
+    const r = res.$response;
+    if (!r) throw new Error("no $response in s3 returned object");
+    return r;
+}
+
 exports.STORAGE_CLASS_STANDARD = STORAGE_CLASS_STANDARD;
 exports.DEFAULT_S3_USER = DEFAULT_S3_USER;
 exports.OP_NAME_TO_ACTION = OP_NAME_TO_ACTION;
@@ -573,3 +590,6 @@ exports.parse_lock_header = parse_lock_header;
 exports.parse_body_object_lock_conf_xml = parse_body_object_lock_conf_xml;
 exports.parse_to_camel_case = parse_to_camel_case;
 exports._is_valid_retention = _is_valid_retention;
+exports.get_http_response_from_resp = get_http_response_from_resp;
+exports.get_http_response_date = get_http_response_date;
+
--- a/src/sdk/namespace_cache.js
+++ b/src/sdk/namespace_cache.js
@@ -11,6 +11,7 @@ const RangeStream = require('../util/range_stream');
 const P = require('../util/promise');
 const buffer_utils = require('../util/buffer_utils');
 const Semaphore = require('../util/semaphore');
+const S3Error = require('../endpoint/s3/s3_errors').S3Error;
 
 const _global_cache_uploader = new Semaphore(cache_config.UPLOAD_SEMAPHORE_CAP, {
     timeout: cache_config.UPLOAD_SEMAPHORE_TIMEOUT,
@@ -27,7 +28,17 @@ class NamespaceCache {
     }
 
     get_write_resource() {
-        return this.namespace_hub;
+        return this;
+    }
+
+    get_bucket() {
+        return this.namespace_hub.get_bucket();
+    }
+
+    is_same_namespace(other) {
+        return other instanceof NamespaceCache &&
+            this.namespace_hub === other.namespace_hub &&
+            this.namespace_nb === other.namespace_nb;
     }
 
     async _delete_object_from_cache(params, object_sdk) {
@@ -157,18 +168,30 @@ class NamespaceCache {
     /////////////////
 
     async list_objects(params, object_sdk) {
-        // TODO listing from cache only for deevelopment
-        return this.namespace_nb.list_objects(params, object_sdk);
+        const get_from_cache = params.get_from_cache;
+        params = _.omit(params, 'get_from_cache');
+        if (get_from_cache) {
+            return this.namespace_nb.list_objects(params, object_sdk);
+        }
+        return this.namespace_hub.list_objects(params, object_sdk);
     }
 
     async list_uploads(params, object_sdk) {
-        // TODO listing from cache only for deevelopment
-        return this.namespace_nb.list_uploads(params, object_sdk);
+        const get_from_cache = params.get_from_cache;
+        params = _.omit(params, 'get_from_cache');
+        if (get_from_cache) {
+            return this.namespace_nb.list_uploads(params, object_sdk);
+        }
+        return this.namespace_hub.list_uploads(params, object_sdk);
     }
 
     async list_object_versions(params, object_sdk) {
-        // TODO listing from cache only for deevelopment
-        return this.namespace_nb.list_object_versions(params, object_sdk);
+        const get_from_cache = params.get_from_cache;
+        params = _.omit(params, 'get_from_cache');
+        if (get_from_cache) {
+            return this.namespace_nb.list_objects(params, object_sdk);
+        }
+        return this.list_object_versions(params, object_sdk);
     }
 
     /////////////////
@@ -468,6 +491,9 @@ class NamespaceCache {
         const operation = 'ObjectCreated';
         const load_for_trigger = object_sdk.should_run_triggers({ active_triggers: this.active_triggers, operation });
 
+        function stream_final_from_promise(promise) {
+            return callback => promise.then(() => callback(), err => callback(err));
+        }
         const bucket_free_space_bytes = await this._get_bucket_free_space_bytes(params, object_sdk);
         let upload_response;
         let etag;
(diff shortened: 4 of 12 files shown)


src/sdk/object_sdk.js
```diff
@@ -95,6 +95,15 @@ class ObjectSDK {
         return bucket.namespace;
     }
 
+    async read_bucket_sdk_caching_info(name) {
+        try {
+            const { bucket } = await bucket_namespace_cache.get_with_cache({ sdk: this, name });
+            return bucket.namespace.caching;
```
jeniawhite, line 101: This crashes if we get here with a regular non-namespace bucket.
Since you are calling that on every request in s3_rest, you will just crash the S3 endpoint and all requests on non-namespaced buckets won't work.

jeniawhite (dismissed):

src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -29,7 +29,9 @@ async function get_bucket(req) {
         key_marker: list_type === '2' ?
             (cont_tok_to_key_marker(cont_tok) || start_after) : req.query.marker,
     };
-
+    if (req.query.get_from_cache !== undefined) {
```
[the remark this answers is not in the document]
guymguym:
```suggestion
    if (req.query.get_from_cache) {
```

src/sdk/object_sdk.js
guymguym: fixed

src/endpoint/s3/ops/s3_get_bucket.js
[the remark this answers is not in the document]
guymguym: get_from_cache was added for testing.
Open technical debt with options to implement:
- HTTP `Cache-Control` header
- Custom Header `X-NooBaa-Cache-Control`

src/sdk/object_sdk.js
```diff
@@ -443,9 +452,16 @@ class ObjectSDK {
...
+                params.copy_source.obj_id = source_md.obj_id;
+                params.size = source_params.end - source_params.start;
             }
+
+            source_params.object_md = source_md;
+            source_params.obj_id = source_md.obj_id;
+
             params.source_stream = await source_ns.read_object_stream(source_params, this);
-            if (params.size > (100 * size_utils.MEGABYTE)) {
+            if (params.size && params.size > (100 * size_utils.MEGABYTE)) {
```
guymguym, line 464: Can you explain what did you try to achieve with these additions? 
It doesn't look right to me - you use copy_source when not the same namespace?

src/endpoint/s3/s3_rest.js
```diff
@@ -152,6 +152,13 @@ async function handle_request(req, res) {
     usage_report.s3_usage_info.total_calls += 1;
     usage_report.s3_usage_info[op_name] = (usage_report.s3_usage_info[op_name] || 0) + 1;
 
+
+    const caching = await req.object_sdk.read_bucket_sdk_caching_info(req.params.bucket);
+    if (req.query && req.query.VersionId && caching) {
```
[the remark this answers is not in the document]
guymguym: @guptsanj Can you please check first `if (req.query && req.query.VersionId)` and only then await the caching info? That would remove an unnecessary await in the common handling of any S3 operation so it's good to be efficient here.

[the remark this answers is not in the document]
guptsanj: Done.
DeleteObjects version check is done in `namespace_cache.js: delete_multiple_objects`

src/sdk/namespace_s3.js
```diff
@@ -228,11 +245,27 @@ class NamespaceS3 {
...
+                if (response.httpResponse.statusCode > 200) return; // will be handled by error event
+                last_modify_date = response.httpResponse.headers.date;
+                statusCode = response.httpResponse.statusCode;
+                return response;
+            }).promise();
         }
         dbg.log0('NamespaceS3.upload_object:', this.bucket, inspect(params), 'res', inspect(res));
         const etag = s3_utils.parse_etag(res.ETag);
-        return { etag, version_id: res.VersionId };
+        return { etag, statusCode: statusCode, version_id: res.VersionId, date: last_modify_date };
```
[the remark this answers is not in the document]
guptsanj: I was using the statusCode to check if the upload operation succeeded. Instead of using `statusCode`, I just check if the date is set in the response in namespace_cache.js

src/endpoint/s3/ops/s3_get_bucket.js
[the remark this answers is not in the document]
guptsanj: This will be handled in a new issue. Leaving this unresolved until the issue is created.

src/sdk/object_sdk.js
guptsanj:
```
                params.copy_source.bucket = actual_source_ns.get_bucket(bucket);
                params.copy_source.obj_id = source_md.obj_id;
                params.size = source_params.end - source_params.start;
```
This was added for a case where a byte range is copied from an object bucket into a namespace bucket.
The test is at: https://github.com/noobaa/noobaa-core/blob/684f4e672eda381e9f63159c51d3c5a9cf61af79/src/test/unit_tests/test_s3_ops.js#L384

guptsanj: @guymguym 
```
            source_params.object_md = source_md;
            source_params.obj_id = source_md.obj_id;
```
Added because of exceptions thrown when 
https://github.com/noobaa/noobaa-core/blob/684f4e672eda381e9f63159c51d3c5a9cf61af79/src/test/unit_tests/test_s3_ops.js#L344
is executed.

guptsanj: The tests which give this exception are
https://github.com/noobaa/noobaa-core/blob/684f4e672eda381e9f63159c51d3c5a9cf61af79/src/test/unit_tests/test_s3_ops.js#L509
https://github.com/noobaa/noobaa-core/blob/684f4e672eda381e9f63159c51d3c5a9cf61af79/src/test/unit_tests/test_s3_ops.js#L513

src/endpoint/s3/s3_utils.js
```diff
@@ -549,6 +554,13 @@ function parse_website_to_body(website) {
     return reply;
 }
 
+function get_http_response_from_ret(res) {
+    const r = res.$response;
+    if (!r) throw new Error("no $response in s3 returned object");
+    if (!r.httpResponse.headers.date) throw new Error("...");
```
guymguym, line 560: Edit the error text

src/sdk/namespace_cache.js
```diff
@@ -157,18 +168,27 @@ class NamespaceCache {
...
+            });
+        }
+        return this.namespace_hub.list_objects(params, object_sdk);
     }
 
     async list_uploads(params, object_sdk) {
-        // TODO listing from cache only for deevelopment
-        return this.namespace_nb.list_uploads(params, object_sdk);
+
+        return this.namespace_hub.list_uploads(params, object_sdk);
```
guymguym, line 186: How about we use the same testing flag `req.query.get_from_cache` to allow both modes?
CC @paulxuzhang

src/sdk/namespace_cache.js
```diff
@@ -157,18 +168,27 @@ class NamespaceCache {
...
     }
 
     async list_object_versions(params, object_sdk) {
-        // TODO listing from cache only for deevelopment
-        return this.namespace_nb.list_object_versions(params, object_sdk);
+        // Always return latest version.
+        return this.list_objects(params, object_sdk);
```
guymguym, line 191: So this means that even if we have versions in the hub we do not list those.
I don't think this is what we should be doing.
I think we should list version from hub, and just avoid caching versions other than latest on GET.

src/test/unit_tests/test_namespace_cache.js
```diff
@@ -239,26 +239,27 @@ class MockNamespace {
 
                 this._buf = Buffer.concat(recv_buf);
                 const etag = this._md5.digest('hex');
-
+                const create_time = new Date().getTime();
```
guymguym, line 242:
```suggestion
                const create_time = Date.now();
```

src/sdk/namespace_s3.js
```diff
@@ -253,7 +253,8 @@ class NamespaceS3 {
         }
         dbg.log0('NamespaceS3.upload_object:', this.bucket, inspect(params), 'res', inspect(res));
         const etag = s3_utils.parse_etag(res.ETag);
-        return { etag, version_id: res.VersionId };
+        const last_modified_time = s3_utils.get_http_response_from_ret(res).httpResponse.headers.date;
+        return { etag, version_id: res.VersionId, date: last_modified_time };
```
guymguym, line 257: why not return it as `last_modified_time`?

src/sdk/namespace_s3.js
```diff
@@ -253,7 +253,8 @@ class NamespaceS3 {
         }
         dbg.log0('NamespaceS3.upload_object:', this.bucket, inspect(params), 'res', inspect(res));
         const etag = s3_utils.parse_etag(res.ETag);
-        return { etag, version_id: res.VersionId };
+        const last_modified_time = s3_utils.get_http_response_from_ret(res).httpResponse.headers.date;
```
guymguym, line 256: I suggest this will be it's own utility function `s3_utils.get_http_response_date(req)` that uses `s3_utils.get_http_response_from_ret()` inside, but it's best to hide these details from the namespace code if possible.

guymguym (changes requested):

src/sdk/namespace_cache.js
guptsanj: The `list_objects` function will return from the hub or from the cache based on the `get_from_cache` param value. So this should return from the hub if there are versions on the hub. I have updated the comment

src/endpoint/s3/s3_utils.js
```diff
@@ -573,3 +590,5 @@ exports.parse_lock_header = parse_lock_header;
 exports.parse_body_object_lock_conf_xml = parse_body_object_lock_conf_xml;
 exports.parse_to_camel_case = parse_to_camel_case;
 exports._is_valid_retention = _is_valid_retention;
+exports.get_http_response_from_ret = get_http_response_from_resp;
```
guymguym, line 593: please fix the exported name and function name to match

src/sdk/namespace_s3.js
```diff
@@ -253,7 +253,8 @@ class NamespaceS3 {
         }
         dbg.log0('NamespaceS3.upload_object:', this.bucket, inspect(params), 'res', inspect(res));
         const etag = s3_utils.parse_etag(res.ETag);
-        return { etag, version_id: res.VersionId };
+        const last_modified_time = s3_utils.get_http_response_date(res);
+        return { etag, version_id: res.VersionId, last_modified_time: last_modified_time };
```
guymguym, line 257:
```suggestion
        return { etag, version_id: res.VersionId, last_modified_time };
```

src/sdk/namespace_cache.js
```diff
@@ -157,18 +168,32 @@ class NamespaceCache {
     /////////////////
 
     async list_objects(params, object_sdk) {
-        // TODO listing from cache only for deevelopment
-        return this.namespace_nb.list_objects(params, object_sdk);
+        const get_from_cache = params.get_from_cache;
+        params = _.omit(params, 'get_from_cache');
+        if (get_from_cache) {
+            return this.namespace_nb.list_objects(params, object_sdk).then(res => {
```
guymguym, line 174: Avoid using `promise.then()`, use `await` instead -
```suggestion
            const res = await this.namespace_nb.list_objects(params, object_sdk);
```

src/sdk/namespace_cache.js
```diff
@@ -157,18 +168,32 @@ class NamespaceCache {
...
+                _.forEach(res.objects, obj => {
+                    obj.create_time = obj.last_modified_time ? obj.last_modified_time : obj.create_time;
```
guymguym, line 176: Why is this needed? Shouldn't that be in our s3 level like we do in `s3_utils.set_response_object_md()`?

src/endpoint/s3/ops/s3_get_bucket.js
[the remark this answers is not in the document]
guymguym: opened https://github.com/noobaa/noobaa-core/issues/6147. resolving.

src/server/object_services/schemas/object_md_schema.js
```diff
@@ -77,7 +77,7 @@ module.exports = {
         upload_started: { objectid: true },
         create_time: { date: true },
         cache_last_valid_time: { date: true },
-
+        last_modified_time: { date: true },
```
guymguym, line 80: Can you please add a comment that explains this new meta-data property.
Also I think a comment for previous line's cache_last_valid_time will be good.
```suggestion
        // last_modified_time is an optional property that is set on cache buckets
        // to separate the times of creation of the cache object (create_time)
        // vs the "real" LastModifiedTime from the hub.
        last_modified_time: { date: true },
```

src/sdk/namespace_cache.js
guptsanj: The object listing does not go through `s3_utils.set_response_object_md()`

guymguym: ok, but it seems that this handling is needed in s3_get_bucket instead of just here in namespace_cache.

src/sdk/namespace_cache.js
```diff
@@ -486,9 +511,19 @@ class NamespaceCache {
...
+                }
+                const update_params = _.pick(_.defaults(
+                    { bucket: this.namespace_nb.target_bucket }, params), 'bucket', 'key', 'last_modified_time');
+                update_params.last_modified_time = (new Date(res.last_modified_time)).getTime();
+                setImmediate(() => object_sdk.rpc_client.object.update_object_md(update_params));
+                return res;
+            };
             // defer the final callback of the cache stream until the hub ack
-            const cache_finalizer = callback => hub_promise.then(() => callback(), err => callback(err));
+            const cache_finalizer = hub_promise.then(data => hub_callback(data), err => hub_callback(err));
```
guymguym, line 526: The type of `cache_finalizer`  should be a function and you assign it a promise - that won't work.
What we wanted to achieve was to hold the cache stream open until the hub promise completes.
So the previous code did that by setting cache_finalizer to a function that puts the cache stream final callback as the callback of the hub_promise.

guymguym:
```
function stream_final_from_promise(promise) {
   return callback => promise.then(() => callback(), err => callback(err));
}

const cache_final_promise = hub_promise.then(async upload_res => {
   await object_sdk.rpc_client.object.update_object_md(update_params);
});
const cache_finalizer = stream_final_from_promise(cache_final_promise);

```

src/sdk/object_sdk.js
guymguym: so we want to remove these changes and see why the test is failing,
because it might be a new test case that is the cause

src/sdk/namespace_cache.js
guymguym: ok

src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -30,7 +30,18 @@ async function get_bucket(req) {
...
 
+    if (req.query.get_from_cache) {
+        params.get_from_cache = req.query.get_from_cache;
+    }
     const reply = await req.object_sdk.list_objects(params);
+
+    // Update the object creation time in case this is from cache
+    if (req.query.get_from_cache) {
+        _.forEach(reply.objects, obj => {
+            obj.create_time = obj.last_modified_time ? obj.last_modified_time : obj.create_time;
```
guymguym, line 41: You know, we don't really want to modify these objects properties meaning,
why not put this inside the response building in line 69 like so:
```
                    // if the object specifies last_modified_time we use it, otherwise take create_time.
                    LastModified: s3_utils.format_s3_xml_date(obj.last_modified_time || obj.create_time),
```

src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -30,7 +30,18 @@ async function get_bucket(req) {
             (cont_tok_to_key_marker(cont_tok) || start_after) : req.query.marker,
     };
 
+    if (req.query.get_from_cache) {
+        params.get_from_cache = req.query.get_from_cache;
+    }
     const reply = await req.object_sdk.list_objects(params);
+
+    // Update the object creation time in case this is from cache
+    if (req.query.get_from_cache) {
```
guymguym, line 39: I don't think this is the condition is needed, I think obj.last_modified_time should always have precedence over create_time if the former is set.

guymguym (approved):
Looks awesome!

PR #6131, opened 2020-08-17
chore: update ftp-srv to 4.3.4

matt-forster:
We've just fixed a critical security vulnerability in `ftp-srv`, and given that your dependencies are pinned, I thought I would bring this into your view directly.

https://github.com/autovance/ftp-srv/security/advisories/GHSA-jw37-5gqr-cf9j

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,7 +77,7 @@
     "dclassify": "1.1.2",
     "event-stream": "4.0.1",
     "express": "4.17.1",
-    "ftp-srv": "4.3.1",
+    "ftp-srv": "4.3.4",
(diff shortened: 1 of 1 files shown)


guymguym (approved):
@forstermatth Thank you very much for the PR! I really appreciate the heads up.

We use dependabot alerts configured to the repo and one hour after your notification, we also got an automated PR from github's dependabot (see https://github.com/noobaa/noobaa-core/pull/6132) - so at least in this case it worked, but it's hard to prioritize there because in many cases our use of a library is for some niche functionality or tests.

Issue #6137, opened 2020-08-21
NS-CACHE: Eviction of object_mds should be done in its own worker


guymguym:
### Environment info
- NooBaa Version: **5.6**
- Platform: **NA**
- CC: @guptsanj @jackyalbo @nimrod-becker 

### Actual behavior

This is a followup of https://github.com/noobaa/noobaa-core/pull/6086#discussion_r474922307

For chunks (with their parts and blocks) we trigger eviction based on capacity and evict chunks from the cache based on LRU. FIFO is even easier to implement just by not moving chunks to the front of the LRU on hit - and FIFO is age based. So if we will prefer simple age based vs access based caching, we can do that - actually we should just add this for dev/test.

For objects_md cleanup - keep in mind that even if we are limited on cache capacity and have to evict, it is still very much likely that cached object_mds have not been deleted or overwritten in the hub, so keeping them is not a bad thing at all. But we do need to evict eventually if the workload will read too many of these to the cache than we can store in the DB. Today we do not have a capacity limit for object_mds and we do not order them in LRU. However we can easily iterate object_mds in FIFO order by traversing the _id index in order or creation in the cache and then check if this object has no parts and then delete.

For now I thought we can piggy back the chunks eviction, BUT adding the age condition I suggested is presenting an issue - if we evict all chunks/parts of an object but it is too young and we filter it out from objects_to_gc, we may never come back to evict more chunks for this same object_md and it will never be evicted. So I would remove the filter for now, but in any case to make it work well we need to move this to its own worker.

### Expected behavior

Ultimately I think object_md cleanup calls for a separate worker - because the iteration subject and order are completely different than tier_ttf_worker. In its own worker we should check how close we are to the object_mds limit we can store in the DB (either config or calculate based on PV), and cleanup in LRU/FIFO order.

### Steps to reproduce
1. Read/write 100 small objects to cache bucket.
2. Read/write 100 huge objects to cache bucket that will cause eviction to occur for the small objects.
3. Check if the object_mds (not the chunks!) of the small objects got evicted or not.

guptsanj:
Created https://github.com/noobaa/noobaa-core/pull/6144 to remove the time based filter added for object_md deletion

PR #6144, opened 2020-08-24
NS-Cache: Do not use a time based filter for removing object_mds when performing eviction in cache buckets.

guptsanj:
Part of issue: https://github.com/noobaa/noobaa-core/issues/6137
When object md has no constituent parts, we need to remove the object md from the database. 

The current removal of the object mds is dependent on a bucket getting some chunks evicted. If there is no eviction, there can be object_mds in the database even though there is no activity on the objects.

The ideal fix will be to move the object_mds in a separate worker so it is not dependent on eviction. 
This will be handled as part of complete fix for issue 6137

### Explain the changes
1. 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/server/object_services/map_builder.js
+++ b/src/server/object_services/map_builder.js
@@ -6,7 +6,6 @@ const assert = require('assert');
 
 const P = require('../../util/promise');
 const dbg = require('../../util/debug_module')(__filename);
-const config = require('../../../config');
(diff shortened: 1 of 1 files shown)


guymguym (approved):

guymguym:
@guptsanj Please rebase

PR #6151, opened 2020-08-28
NS-Cache:  make test_s3_ops unit test run with an external endpoint or default noobaa endpoint

guptsanj:
-     Fix the s3 ops unit test so that if an external endpoint is provided
       via env variables, the endpoint is used for external buckets
       If non is provided, noobaa buckets are created for tests
- Check that GET and LIST objects works on cache bucket with the `get_from_cache` variable.

    Tests run on regular, namespace and namespace cache buckets

Following env variables need to be defined for external buckets to be used:

-  ENDPOINT 
-  ENDPOINT_TYPE 
-  AWS_ACCESS_KEY_ID  
-  AWS_SECRET_ACCESS_KEY  

In addition, following buckets need to already exist on the external endpoint:

-  remote-s3-ops-bucket
-  remote-s3-ops-bucket-caching
-  remote-s3-ops-source
-  remote-s3-ops-source-caching

### Explain the changes
1. 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/test/unit_tests/test_s3_ops.js
+++ b/src/test/unit_tests/test_s3_ops.js
@@ -5,31 +5,32 @@ const _ = require('lodash');
 // setup coretest first to prepare the env
 const coretest = require('./coretest');
 coretest.setup({ pools_to_create: coretest.POOL_LIST });
-
+const config = require('../../../config');
 const AWS = require('aws-sdk');
-const http = require('http');
+const http_utils = require('../../util/http_utils');
 const mocha = require('mocha');
 const assert = require('assert');
 const querystring = require('querystring');
-const SKIP_TEST = !process.env.AWS_ACCESS_KEY_ID || !process.env.AWS_SECRET_ACCESS_KEY;
 
-const AWS_TARGET_BUCKET = 's3-ops-test-bucket';
-const AWS_SOURCE_BUCKET = 's3-ops-source';
+// If any of these variables are not defined,
+// use the noobaa endpoint to create buckets
+// for namespace and namespace cache bucket testing.
+const USE_REMOTE_ENDPOINT = process.env.USE_REMOTE_ENDPOINT === 'true';
 const { rpc_client, EMAIL } = coretest;
 const BKT1 = 'test-s3-ops-bucket-ops';
 const BKT2 = 'test-s3-ops-object-ops';
 const BKT3 = 'test2-s3-ops-object-ops';
 const BKT4 = 'test3-s3-ops-object-ops';
 const BKT5 = 'test5-s3-ops-objects-ops';
-const CONNECTION_NAME = 'aws_connection1';
-const RESOURCE_NAME = 'namespace_target_bucket';
-const RESOURCE_NAME_SOURCE = 'namespace_source_bucket';
-
+const CONNECTION_NAME = 's3_connection';
+const NAMESPACE_RESOURCE_SOURCE = 'namespace_target_bucket';
+const NAMESPACE_RESOURCE_TARGET = 'namespace_source_bucket';
+const TARGET_BUCKET = 's3-ops-target';
+const SOURCE_BUCKET = 's3-ops-source';
 const file_body = "TEXT-FILE-YAY!!!!-SO_COOL";
 const file_body2 = "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx";
 const sliced_file_body = "TEXT-F";
 const sliced_file_body1 = "_COOL";
-const source_bucket = 's3-ops-source';
 const text_file1 = 'text-file1';
 const text_file2 = 'text-file2';
 const text_file3 = 'text-file3';
@@ -38,6 +39,11 @@ const text_file5 = 'text-file5';
 mocha.describe('s3_ops', function() {
 
     let s3;
+    // Bucket name for the source namespace resource
+    let source_namespace_bucket;
+    // Bucket name for the target namespace resource
+    let target_namespace_bucket;
+    let source_bucket;
 
     mocha.before(async function() {
         const self = this; // eslint-disable-line no-invalid-this
@@ -53,7 +59,7 @@ mocha.describe('s3_ops', function() {
             computeChecksums: true,
             s3DisableBodySigning: false,
             region: 'us-east-1',
-            httpOptions: { agent: new http.Agent({ keepAlive: false }) },
+            httpOptions: { agent: http_utils.get_unsecured_agent(coretest.get_http_address()) },
         });
         coretest.log('S3 CONFIG', s3.config);
     });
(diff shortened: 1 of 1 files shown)


src/test/unit_tests/test_s3_ops.js
```diff
@@ -307,97 +321,95 @@ mocha.describe('s3_ops', function() {
...
-            const res = await s3.getObject({ Bucket: bucket_name, Key: text_file1 }).promise();
+            const res = await s3_local.getObject({ Bucket: bucket_name, Key: text_file1 }).promise();
             assert.strictEqual(res.Body.toString(), file_body);
             assert.strictEqual(res.ContentType, 'text/plain');
             assert.strictEqual(res.ContentLength, file_body.length);
         });
 
-        mocha.it('should get text-file sliced 1', async function() {
-            const res = await s3.getObject({ Bucket: bucket_name, Key: text_file1, Range: 'bytes=0-5' }).promise();
+         mocha.it('should get text-file sliced 1', async function() {
```
liranmauda, line 330:
```suggestion
        mocha.it('should get text-file sliced 1', async function() {
```

src/test/unit_tests/test_s3_ops.js
```diff
@@ -438,69 +450,79 @@ mocha.describe('s3_ops', function() {
         });
         mocha.it('should list objects with text-file', async function() {
             // eslint-disable-next-line no-invalid-this
-            this.timeout(60000);
-            const res1 = await s3.listObjects({ Bucket: bucket_name }).promise();
+            this.timeout(6000000);
```
liranmauda, line 453: Why this timeout so big? 
Unit tests should be small and fast, can we reduce it under 5 min?

liranmauda (dismissed):
There is a deepscan issue, lets see if it is valid.

src/test/unit_tests/test_s3_ops.js
```diff
@@ -37,14 +37,15 @@ const text_file5 = 'text-file5';
...
+    let s3_local;
+    let REMOTE_SOURCE_BUCKET;
+    let REMOTE_TARGET_BUCKET;
     mocha.before(async function() {
         const self = this; // eslint-disable-line no-invalid-this
         self.timeout(60000);
 
         const account_info = await rpc_client.account.read_account({ email: EMAIL, });
-        s3 = new AWS.S3({
+        s3_local = new AWS.S3({
```
guymguym, line 48: What is the purpose of this rename from s3 to s3_local?
Can we revert it?

guptsanj: Reverted.

src/test/unit_tests/test_s3_ops.js
```diff
@@ -5,16 +5,15 @@ const _ = require('lodash');
...
 coretest.setup({ pools_to_create: coretest.POOL_LIST });
-
+const config = require('../../../config');
 const AWS = require('aws-sdk');
 const http = require('http');
 const mocha = require('mocha');
 const assert = require('assert');
 const querystring = require('querystring');
-const SKIP_TEST = !process.env.AWS_ACCESS_KEY_ID || !process.env.AWS_SECRET_ACCESS_KEY;
+const USE_LOCAL_ENDPOINT = !process.env.AWS_ACCESS_KEY_ID || !process.env.AWS_SECRET_ACCESS_KEY ||
```
guymguym, line 14:
```suggestion
const USE_LOCAL_ENDPOINT = process.env.USE_LOCAL_ENDPOINT === 'true';
```

guymguym (approved):

src/test/unit_tests/test_s3_ops.js
```diff
@@ -94,51 +95,63 @@ mocha.describe('s3_ops', function() {
...
                 const write_resource1 = RESOURCE_NAME_SOURCE;
-                const ENDPOINT = process.env.ENDPOINT ? process.env.ENDPOINT : 'https://s3.amazonaws.com';
-                const ENDPOINT_TYPE = process.env.ENDPOINT_TYPE ? process.env.ENDPOINT_TYPE : 'AWS';
+                const ENDPOINT = USE_LOCAL_ENDPOINT ? coretest.get_https_address() : process.env.ENDPOINT;
+                const ENDPOINT_TYPE = USE_LOCAL_ENDPOINT ? 'AWS' : process.env.ENDPOINT_TYPE;
+                const AWS_ACCESS_KEY_ID = USE_LOCAL_ENDPOINT ? s3.config.accessKeyId : process.env.AWS_ACCESS_KEY_ID;
+                const AWS_SECRET_ACCESS_KEY = USE_LOCAL_ENDPOINT ? s3.config.secretAccessKey : process.env.AWS_SECRET_ACCESS_KEY;
+
+                if (USE_LOCAL_ENDPOINT) {
+                    process.env.NODE_TLS_REJECT_UNAUTHORIZED = 0;
```
guymguym, line 123: Why is that needed in such a global way? 
For endpoints with self-signed certificates we usually set the option of a local s3 sdk client with `agent: http_utils.get_default_agent(endpoint)` or `agent: http_uttils.get_unsecured_agent(endpoint)`, for example see here:
https://github.com/noobaa/noobaa-core/blob/74b79637d8a9a1dca2baf51598860579711a5c6b/src/agent/block_store_services/block_store_s3.js#L32-L62

src/test/unit_tests/test_s3_ops.js
```diff
@@ -5,16 +5,15 @@ const _ = require('lodash');
...
 const http = require('http');
 const mocha = require('mocha');
 const assert = require('assert');
 const querystring = require('querystring');
-const SKIP_TEST = !process.env.AWS_ACCESS_KEY_ID || !process.env.AWS_SECRET_ACCESS_KEY;
 
-const AWS_TARGET_BUCKET = 's3-ops-test-bucket';
-const AWS_SOURCE_BUCKET = 's3-ops-source';
+// Uses the noobaa endpoint to create buckets for namespace and namespace cache bucket testing.
+const USE_LOCAL_ENDPOINT = process.env.USE_LOCAL_ENDPOINT === 'true';
```
guymguym, line 16: Hmm, didn't we mean the default to be the opposite?

guymguym (changes requested):

src/test/unit_tests/test_s3_ops.js
guptsanj: @guymguym I reverted the changes back to the original state and added some comments and logging.
The issue if we use`const USE_LOCAL_ENDPOINT = process.env.USE_LOCAL_ENDPOINT === 'true';` then I have to define default behavior in case neither the AWS endpoint nor USE_LOCAL_ENDPOINT is defined.

We will have to skip the `namespace bucket` add `namespace cache bucket` tests if neither variable is defined. This will make the test more complicated. We should just run all tests as part of travis/local testing. 

Please let me know if this is acceptable, or I will add a default behavior: no variable defined means skip the `namespace` related s3 tests.

src/test/unit_tests/test_s3_ops.js
guptsanj: I was running into errors such as `NetworkingError: self signed certificate`

At: https://github.com/noobaa/noobaa-core/blob/a075e9bdc8cced6e4b517e6161fe5aed27ff4052/src/test/unit_tests/test_s3_ops.js#L51, a http endpoint is created  (This is where I was adding get_unsecured_agent/get_default_agent)

When creating an external connection, I added an `https` connection (https://github.com/noobaa/noobaa-core/blob/a075e9bdc8cced6e4b517e6161fe5aed27ff4052/src/test/unit_tests/test_s3_ops.js#L120)

When a namespace bucket is created using this connection, I was getting (a PUT object to the namespace bucket) either a 
* `self signed certificate` error if I used `https` for external connection endpoint -or-
* `Error: Non-file stream objects are not supported with SigV4` if I used `http` for the external endpoint connection.

Setting `process.env.NODE_TLS_REJECT_UNAUTHORIZED` I was able to get the tests to run. 
I did try out the same sequence in the management console and did not encounter any issues.

src/test/unit_tests/test_s3_ops.js
guptsanj: Well, using noobaa bucket as default will make the tests on travis not dependent on an external resource. 

If we make the external bucket as default for the s3 unit tests, the s3 test needs to be updated so that a unique external bucket is created for each pull request (does not happen at the moment).

If external endpoint should be the default, I can do that in the next pull request.
I need to create an external bucket with the commit id to ensure uniqueness

guymguym: by default I meant that if we do not set USE_LOCAL_ENDPOINT then it should run the default mode.
So is it going to be looking for external endpoint if I do * * * not * * * set USE_LOCAL_ENDPOINT?

guymguym: I am trying to say that I think we want to reverse this env meaning to:
```
const USE_REMOTE_ENDPOINT = process.env.USE_REMOTE_ENDPOINT === 'true';
```

src/test/unit_tests/test_s3_ops.js
guymguym: try to update this line:
https://github.com/noobaa/noobaa-core/blob/7c8eedea337f7fbca975f720ace1927065e46692/src/test/unit_tests/test_s3_ops.js#L56

to:
```
agent: http_utils.get_unsecured_agent(endpoint) 
```

src/test/unit_tests/test_s3_ops.js
guptsanj: Ok, sorry, I misunderstood, I will do that.

src/test/unit_tests/test_s3_ops.js
guptsanj: yes, I tried it. I get: `NetworkingError: self signed certificate`

So the error is triggered at: https://github.com/noobaa/noobaa-core/blob/a075e9bdc8cced6e4b517e6161fe5aed27ff4052/src/test/unit_tests/test_s3_ops.js#L189

for a PUT object to the `namespace`/`caching namespace` bucket. 

The namespace bucket is created with the external connection at `https://github.com/noobaa/noobaa-core/blob/a075e9bdc8cced6e4b517e6161fe5aed27ff4052/src/test/unit_tests/test_s3_ops.js#L131`.

When the external endpoint is set to IBM COS, the write to the namespace bucket is successful. This only happens when the endpoint is the noobaa endpoint (`coretest.get_https_address()`)

guymguym: Oh ok, but I see you replaced it in latest update. Is that working now?

src/test/unit_tests/test_s3_ops.js
```diff
@@ -94,51 +97,61 @@ mocha.describe('s3_ops', function() {
...
+                /**
+                 * NOTE: Remote buckets are not created in the test. It is assumed that
+                 * a. The buckets exist
+                 * b. The buckets are empty
+                 * Following buckets need to exist on the remote endpoint:
+                 *
+                 * remote-s3-ops-bucket-caching
+                 * remote-s3-ops-source-caching
+                 */
+                if (!USE_REMOTE_ENDPOINT) {
```
guymguym, line 132: Shouldn't this be the opposite condition to decide if to create the remote buckets?
```suggestion
                if (USE_REMOTE_ENDPOINT) {
```

src/test/unit_tests/test_s3_ops.js
```diff
@@ -94,51 +97,61 @@ mocha.describe('s3_ops', function() {
...
+                REMOTE_TARGET_BUCKET = caching ? TARGET_BUCKET + '-caching' : TARGET_BUCKET;
+
                 const read_resources = [RESOURCE_NAME];
                 const write_resource = RESOURCE_NAME;
                 const read_resources1 = [RESOURCE_NAME_SOURCE];
                 const write_resource1 = RESOURCE_NAME_SOURCE;
-                const ENDPOINT = process.env.ENDPOINT ? process.env.ENDPOINT : 'https://s3.amazonaws.com';
-                const ENDPOINT_TYPE = process.env.ENDPOINT_TYPE ? process.env.ENDPOINT_TYPE : 'AWS';
+                const ENDPOINT = USE_REMOTE_ENDPOINT ? process.env.ENDPOINT : coretest.get_https_address();
+                const ENDPOINT_TYPE = USE_REMOTE_ENDPOINT ? process.env.ENDPOINT_TYPE : 'AWS';
```
guymguym, line 109: Hmm I think this should be s3-compatible for using the noobaa bucket.
This might have been the cause of the ssl certificates you had.

guptsanj: This fixed the issue. Just updated the PR. Thanks!!

guymguym:
@guptsanj What about the condition - was that intentionally creating the remote buckets only if NOT using remote endpoint?

guptsanj:
The names for bucket are misnamed. We need 2 buckets for 2 namespace resources. Those buckets are named `REMOTE_TARGET_BUCKET` and `REMOTE_SOURCE_BUCKET`
For non remote endpoints, these buckets are created.
The remote endpoint test assumes the buckets already exist at the remote endpoint (will fix that in the pull request)

src/test/unit_tests/test_s3_ops.js
guptsanj: We need 2 buckets for both remote and non remote endpoint.
For `USE_REMOTE_ENDPOINT` the buckets are assumed to exist on the remote endpoint.
For a non remote endpoint, the buckets are created in noobaa.

The remote endpoint's buckets are an issue as there can be multiple pull requests running this test, and all of the pull requests will use the same buckets for reading/writing objects with the same name. This will cause conflicts and fail the tests. The ideal solution will be for the remote buckets have a unique name per pull request test (say the commit id).

This is not an issue when a non remote endpoint is used as the noobaa core gets set up for the specific pull request test so the bucket name has isolation 

In the above code, 2 buckets are created only if `USE_REMOTE_ENDPOINT` is false, then the tests run on noobaa based buckets

guymguym (approved):

PR #6152, opened 2020-08-28
NS-CACHE: Added support for GET and PUT object ACL

mcrepeau:
### Explain the changes
1. Added support for PUT object ACL
2. Added support for GET object ACL
3. Added configuration param to define behavior: "reject", "pass-through", else default behavior will apply

1. Missing unit tests

### Testing Instructions:
1. Set the config.ACL_HANDLING parameter to either "reject", "pass-through" or anything else
2. Make PUT or GET object ACL request

### DIFF
--- a/config.js
+++ b/config.js
@@ -456,6 +456,8 @@ config.NAMESPACE_CACHING = {
     MIN_OBJECT_AGE_FOR_GC: 1000 * 60 * 60 * 24,
     UPLOAD_SEMAPHORE_CAP: Math.floor(
         Number(process.env.CONTAINER_MEM_REQUEST ? process.env.CONTAINER_MEM_REQUEST : os.totalmem()) / 8),
+    /** @type string 'reject' | 'pass-through' | '' */
+    ACL_HANDLING: "",
 };
 
 assert(config.NAMESPACE_CACHING.DEFAULT_BLOCK_SIZE <= config.NAMESPACE_CACHING.DEFAULT_MAX_CACHE_OBJECT_SIZE);
--- a/src/endpoint/s3/ops/s3_get_object_acl.js
+++ b/src/endpoint/s3/ops/s3_get_object_acl.js
@@ -1,34 +1,36 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
-const s3_utils = require('../s3_utils');
+const _ = require('lodash');
 
 /**
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGETacl.html
  */
-function get_object_acl(req) {
-    return req.object_sdk.read_object_md({
-            bucket: req.params.bucket,
-            key: req.params.key,
-            version_id: req.query.versionId
-        })
-        .then(object_md => ({
-            AccessControlPolicy: {
-                Owner: s3_utils.DEFAULT_S3_USER,
-                AccessControlList: [{
+async function get_object_acl(req) {
+    const reply = await req.object_sdk.get_object_acl({
+        bucket: req.params.bucket,
+        key: req.params.key,
+        version_id: req.query.versionId,
+    });
+
+    return {
+        AccessControlPolicy: {
+            Owner: reply.owner,
+            AccessControlList:
+                _.map(reply.access_control_list, acl => ({
                     Grant: {
                         Grantee: {
                             _attr: {
                                 'xmlns:xsi': 'http://www.w3.org/2001/XMLSchema-instance',
-                                'xsi:type': 'CanonicalUser',
+                                'xsi:type': acl.Grantee.Type,
                             },
-                            _content: s3_utils.DEFAULT_S3_USER,
+                            _content: _.omit(acl.Grantee, 'Type')
                         },
-                        Permission: 'FULL_CONTROL'
+                        Permission: acl.Permission
                     }
-                }]
-            }
-        }));
+                }))
+        }
+    };
 }
 
 module.exports = {
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -18,6 +18,14 @@ const DEFAULT_S3_USER = Object.freeze({
     DisplayName: 'NooBaa'
 });
 
+const DEFAULT_OBJECT_ACL = Object.freeze({
+    owner: DEFAULT_S3_USER,
+    access_control_list: [{
+        Grantee: { ...DEFAULT_S3_USER, Type: "CanonicalUser" },
+        Permission: "FULL_CONTROL"
+    }]
+});
+
 const OP_NAME_TO_ACTION = Object.freeze({
     delete_bucket_analytics: { regular: "s3:putanalyticsconfiguration" },
     delete_bucket_cors: { regular: "s3:putbucketcors" },
@@ -358,7 +366,6 @@ function parse_lock_header(req) {
     return lock_settings;
 }
 
-
 function parse_tagging_header(req) {
     const tagging_header = req.headers['x-amz-tagging'];
     if (!tagging_header) return;
@@ -568,6 +575,7 @@ function get_http_response_from_resp(res) {
 
 exports.STORAGE_CLASS_STANDARD = STORAGE_CLASS_STANDARD;
 exports.DEFAULT_S3_USER = DEFAULT_S3_USER;
+exports.DEFAULT_OBJECT_ACL = DEFAULT_OBJECT_ACL;
 exports.OP_NAME_TO_ACTION = OP_NAME_TO_ACTION;
 exports.format_s3_xml_date = format_s3_xml_date;
 exports.get_request_xattr = get_request_xattr;
--- a/src/sdk/namespace_cache.js
+++ b/src/sdk/namespace_cache.js
@@ -12,6 +12,8 @@ const P = require('../util/promise');
 const buffer_utils = require('../util/buffer_utils');
 const Semaphore = require('../util/semaphore');
 const S3Error = require('../endpoint/s3/s3_errors').S3Error;
+const s3_utils = require('../endpoint/s3/s3_utils');
+const config = require('../../config');
 
 const _global_cache_uploader = new Semaphore(cache_config.UPLOAD_SEMAPHORE_CAP, {
     timeout: cache_config.UPLOAD_SEMAPHORE_TIMEOUT,
@@ -739,6 +741,34 @@ class NamespaceCache {
         return this.namespace_hub.get_blob_block_lists(params, object_sdk);
     }
 
+    //////////
+    // ACLs //
+    //////////
+
+    async get_object_acl(params, object_sdk) {
+        if (config.NAMESPACE_CACHING.ACL_HANDLING === "reject") {
+            throw new S3Error(S3Error.AccessDenied);
+        }
+
+        if (config.NAMESPACE_CACHING.ACL_HANDLING === "pass-through") {
+            return this.namespace_hub.get_object_acl(params, object_sdk);
+        }
+
+        await this.read_object_md(params, object_sdk);
+        return s3_utils.DEFAULT_OBJECT_ACL;
+    }
+
+    async put_object_acl(params, object_sdk) {
+        if (config.NAMESPACE_CACHING.ACL_HANDLING === "reject") {
+            throw new S3Error(S3Error.AccessDenied);
+        }
+
+        if (config.NAMESPACE_CACHING.ACL_HANDLING === "pass-through") {
+            return this.namespace_hub.put_object_acl(params, object_sdk);
+        }
+
+        await this.read_object_md(params, object_sdk);
+    }
 }
 
 
--- a/src/sdk/namespace_nb.js
+++ b/src/sdk/namespace_nb.js
@@ -5,6 +5,7 @@ const _ = require('lodash');
 
 const P = require('../util/promise');
 const blob_translator = require('./blob_translator');
+const s3_utils = require('../endpoint/s3/s3_utils');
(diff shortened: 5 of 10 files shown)


src/endpoint/s3/ops/s3_get_object_acl.js
```diff
@@ -1,12 +1,46 @@
...
 /**
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGETacl.html
  */
-function get_object_acl(req) {
+async function get_object_acl(req) {
+    if (config.ACL_HANDLING === "reject") {
+        throw new S3Error(S3Error.AccessDenied);
+    } else if (config.ACL_HANDLING === "pass-through") {
+        //TODO: form a XML from the info in request
+        const reply = await req.object_sdk.get_object_acl({
```
guymguym, line 17: This new flow seem too similar to the existing code (the next case).
Can't we merge those cases?

src/sdk/namespace_nb.js
```diff
@@ -312,6 +312,18 @@ class NamespaceNB {
     put_object_retention(params, object_sdk) {
         return object_sdk.rpc_client.object.put_object_retention(params);
     }
+
+    ////////////
+    //  ACLs  //
+    ////////////
+
+    get_object_acl(params, object_sdk) {
+        return object_sdk.rpc_client.object.get_object_acl(params);
```
guymguym, line 321: You are referring to API's that do not exist on object_api...

guymguym (changes requested):
@mcrepeau See some comments/questions. Thanks!

src/sdk/namespace_nb.js
mcrepeau: Should they be added to object_api then?

config.js
```diff
@@ -521,3 +521,9 @@ config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
 
 // Temporary flag to turn PostgreSQL as DB
 config.USE_POSTGRESQL = false;
+
+///////////////////////////////
+//        ACL RELATED       //
+///////////////////////////////
+
+config.ACL_HANDLING = "";
```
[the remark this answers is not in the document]
guymguym: move into config.NAMESPACE_CACHING

src/endpoint/s3/ops/s3_get_object_acl.js
```diff
@@ -1,34 +1,53 @@
...
+const config = require('../../../../config');
 const s3_utils = require('../s3_utils');
 
 /**
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectGETacl.html
  */
-function get_object_acl(req) {
-    return req.object_sdk.read_object_md({
+async function get_object_acl(req) {
+    if (config.ACL_HANDLING === "reject") {
```
guymguym, line 13: We want to push the logic into namespace_cache.

src/sdk/namespace_nb.js
guymguym:
```
const res = await req.object_sdk.read_object_md({ .. });
return { 
  owner: s3_utils.DEFAULT_S3_USER, 
  access_control_list: [{ 
    Grantee: { ID: s3_utils.DEFAULT_S3_USER, Type: "CanonicalUser" }, 
    Permission: "FULL_CONTROL" 
  }]
};
```

guymguym (changes requested):

src/sdk/namespace_nb.js
```diff
@@ -312,6 +313,30 @@ class NamespaceNB {
...
+            owner: s3_utils.DEFAULT_S3_USER,
+            access_control_list: [{
+                Grantee: { ID: s3_utils.DEFAULT_S3_USER, Type: "CanonicalUser" },
+                Permission: "FULL_CONTROL"
+            }]
+        };
+    }
+
+    put_object_acl(params, object_sdk) {
+        return object_sdk.rpc_client.object.put_object_acl(params);
```
guymguym, line 338: Since this api does not exist in object_api we want this flow to do what s3_put_object_acl used to do which is to read_object_md to detect if bucket/object is missing or other errors, and then ignore the acl.

config.js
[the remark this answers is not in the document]
guymguym: Still need to move this...

guymguym (changes requested):

src/sdk/namespace_nb.js
```diff
@@ -312,6 +313,34 @@ class NamespaceNB {
...
+            owner: s3_utils.DEFAULT_S3_USER,
+            access_control_list: [{
+                Grantee: { ID: s3_utils.DEFAULT_S3_USER, Type: "CanonicalUser" },
+                Permission: "FULL_CONTROL"
+            }]
+        };
+    }
+
+    async put_object_acl(params, object_sdk) {
+        return await object_sdk.read_object_md({
```
guymguym: We should return just `{version_id}` right? So maybe best not to return the entire object_md if this is not how other namespaces work, and do it
like you did for get_object_acl.

Also can you put the todo comment that was in this code :
```
// TODO S3 ignoring put_object_acl for now
```

src/sdk/namespace_cache.js
```diff
@@ -739,6 +753,33 @@ class NamespaceCache {
...
     }
 
+    //////////
+    // ACLs //
+    //////////
+
+    async get_object_acl(params, object_sdk) {
+        if (config.NAMESPACE_CACHING.ACL_HANDLING === "reject") {
+            throw new S3Error(S3Error.AccessDenied);
+        } else if (config.NAMESPACE_CACHING.ACL_HANDLING === "pass-through") {
```
guymguym, line 763: No need for else after throw/return.

src/sdk/namespace_cache.js
```diff
@@ -163,6 +165,18 @@ class NamespaceCache {
         return false;
     }
 
+    async _update_object_acl(params, object_sdk) {
```
guymguym, line 168: Seems unused, is it?

src/sdk/namespace_nb.js
mcrepeau: Gotcha

src/endpoint/s3/ops/s3_get_object_acl.js
```diff
@@ -1,34 +1,37 @@
...
+    const reply = await req.object_sdk.get_object_acl({
+        bucket: req.params.bucket,
+        key: req.params.key,
+        version_id: req.query.versionId,
+    });
+
+    return {
+        AccessControlPolicy: {
+            Owner: reply.owner,
+            AccessControlList: [
```
guymguym, line 19: I don't think you need to wrap the map in an array

src/sdk/namespace_cache.js
```diff
@@ -739,6 +741,35 @@ class NamespaceCache {
...
+            return this.namespace_hub.get_object_acl(params, object_sdk);
+        }
+
+        return {
+            owner: s3_utils.DEFAULT_S3_USER,
+            access_control_list: [{
+                Grantee: { ID: s3_utils.DEFAULT_S3_USER, Type: "CanonicalUser" },
+                Permission: "FULL_CONTROL"
+            }]
+        };
```
guymguym, line 763: I think before we return the default ACL we need to check if the object exists, just like we do in namespace_nb, but in this case our call to this.read_object_md() will use the namespace_cache implementation:
```
            await this.read_object_md({ ... }, object_sdk);
            return s3_utils.DEFAULT_OBJECT_ACL; // see other comment on moving this to s3_utils...
```

src/sdk/namespace_cache.js
```diff
@@ -739,6 +741,35 @@ class NamespaceCache {
...
+    }
+
+    async put_object_acl(params, object_sdk) {
+        if (config.NAMESPACE_CACHING.ACL_HANDLING === "reject") {
+            throw new S3Error(S3Error.AccessDenied);
+        }
+
+        return this.namespace_hub.put_object_acl(params, object_sdk);
```
guymguym, line 771: Should we call the hub only when ACL_HANDLING is pass-through?
When ACL_HANDLING is empty string we can do the same as namespace_nb does and call read_object_md and then return without applying the ACL.

guymguym (changes requested):
Hey @mcrepeau I've completed the review. See comments above. Thanks

src/endpoint/s3/ops/s3_get_object_acl.js
```diff
@@ -1,34 +1,37 @@
...
+                _.map(reply.access_control_list, acl => ({
                     Grant: {
                         Grantee: {
                             _attr: {
                                 'xmlns:xsi': 'http://www.w3.org/2001/XMLSchema-instance',
-                                'xsi:type': 'CanonicalUser',
+                                'xsi:type': acl.Grantee.Type,
                             },
-                            _content: s3_utils.DEFAULT_S3_USER,
+                            _content: _.without(acl.Grantee, function(grantee) { return grantee.Type; })
```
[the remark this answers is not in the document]
mcrepeau: No because there are the ID and DisplayName fields, either of which can be displayed. So all I want to do here is to remove the Type field, which is stored just to be used in the `xsi:type` and display the rest. Without does exactly what I need

src/sdk/namespace_nb.js
```diff
@@ -312,6 +313,33 @@ class NamespaceNB {
...
+                Permission: "FULL_CONTROL"
+            }]
+        };
+    }
+
+    async put_object_acl(params, object_sdk) {
+        //TODO S3 ignoring put_object_acl for now
+        return object_sdk.read_object_md({
+            version_id: params.versionId
+        });
```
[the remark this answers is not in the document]
mcrepeau: Ohh, ok, I see. I misunderstood you when you previously said `We should return just {version_id} right?`. And what should we return then, still version_id?

src/sdk/namespace_nb.js
```diff
@@ -312,6 +313,33 @@ class NamespaceNB {
     put_object_retention(params, object_sdk) {
         return object_sdk.rpc_client.object.put_object_retention(params);
     }
+
+    ////////////
+    //  ACLs  //
+    ////////////
+
+    async get_object_acl(params, object_sdk) {
+        await object_sdk.read_object_md({
```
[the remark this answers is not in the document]
mcrepeau: Ah yes, I was wondering. This makes sense

src/sdk/namespace_nb.js
```diff
@@ -312,6 +313,33 @@ class NamespaceNB {
...
+            version_id: params.versionId
+        });
+
+        return {
+            owner: s3_utils.DEFAULT_S3_USER,
+            access_control_list: [{
+                Grantee: { ID: s3_utils.DEFAULT_S3_USER, Type: "CanonicalUser" },
+                Permission: "FULL_CONTROL"
+            }]
+        };
```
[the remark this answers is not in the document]
mcrepeau: 👍

src/sdk/namespace_cache.js
mcrepeau: Should we return nothing, then?

guymguym: Yes

src/sdk/namespace_nb.js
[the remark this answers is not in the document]
guymguym: I think put acl doesn't need anything to be returned, so just no return will make it return void (undefined)

src/endpoint/s3/s3_utils.js
```diff
@@ -18,6 +18,14 @@ const DEFAULT_S3_USER = Object.freeze({
     DisplayName: 'NooBaa'
 });
 
+const DEFAULT_OBJECT_ACL = Object.freeze({
+    owner: DEFAULT_S3_USER,
+    access_control_list: [{
+        Grantee: { ID: DEFAULT_S3_USER, Type: "CanonicalUser" },
```
guymguym, line 24:
```suggestion
        Grantee: { 
            ...DEFAULT_S3_USER,
            Type: "CanonicalUser"
        },
```

src/endpoint/s3/ops/s3_get_object_acl.js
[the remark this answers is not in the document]
guymguym:
```suggestion
                            _content: _.omit(acl.Grantee, 'Type')
```

src/sdk/namespace_cache.js
```diff
@@ -739,6 +741,32 @@ class NamespaceCache {
...
+
+    async put_object_acl(params, object_sdk) {
+        if (config.NAMESPACE_CACHING.ACL_HANDLING === "reject") {
+            throw new S3Error(S3Error.AccessDenied);
+        }
+
+        if (config.NAMESPACE_CACHING.ACL_HANDLING === "pass-through") {
+            return this.namespace_hub.put_object_acl(params, object_sdk);
+        }
+    }
```
guymguym, line 769: put_object_acl when ACL_HANDLING === '' then we also need to load the object for the errors:
```
await this.read_object_md(params, object_sdk);
```

guymguym (changes requested):

guymguym (commented):
Also for namespace_blob add:
```
    async get_object_acl() {
        await this.read_object_md(params, object_sdk);
        return s3_utils.DEFAULT_OBJECT_ACL;
    }

    async put_object_acl() {
        await this.read_object_md(params, object_sdk);
    }
```

guymguym (approved):
Looks good @mcrepeau

guymguym:
@mcrepeau But you need to rebase from master

guymguym:
@mcrepeau Please again on master and squash to a single commit, so we can merge

PR #6163, opened 2020-09-03
NS-CACHE: Cache object with size <= config.INLINE_MAX_SIZE

### LINKED ISSUE #6150: NS-CACHE: < 4k Objects not cached if SS copied or Written out of band (direct upload to hub)
### Environment info
Environment info

NooBaa Version: 5.6
Platform: NA
CC: @guymguym 

### Actual behavior
1. Create a namespace bucket with caching enabled. 
2. Upload a < 4k Object to the hub (not via the namespace bucket)
3. GET object via the namespace bucket (object_api.read_object_md exceptions in s3 log), 
4. GET object via the namespace bucket (same exception again)
A look in the `db.objectmds` collection shows the object was not added to the metadata.

Looks like the object does not get cached 
Same behavior is encountered if a < 4k object is SS Copied or Multi part uploaded to the namespace cache bucket.

### Expected behavior
At the 2nd GET operation, I will expect to see the object is retrieved from the cache 

### Steps to reproduce

See above

paulxuzhang:

### Explain the changes
1. Fix the issue that object with size <= config.INLINE_MAX_SIZE does not get cached.

2. Fix issue in updating last_modified_time
It appears that updating last_modified_time does not fully work. In the existing implementation, updating last_modified_time happens in the finalizer of cache stream and before complete_object_upload gets called. The update fails because the db operation looks for object with "upload_started" NOT set, while it can only happen after  complete_object_upload() gets called, so the update eventually fails with upload not found.

1.  Fix the issue #6150 

### Testing Instructions:
1.  upload object with size <= config.INLINE_MAX_SIZE and validate that it is cached.

### DIFF
--- a/src/sdk/namespace_cache.js
+++ b/src/sdk/namespace_cache.js
@@ -243,6 +243,23 @@ class NamespaceCache {
 
                 return object_info_cache;
 
+            } else if (object_info_hub.first_range_data && object_info_hub.size <= object_info_hub.first_range_data.length) {
+                // If the inline read covers entire object, we will submit it to cache.
+                _global_cache_uploader.submit_background(
+                    object_info_hub.size,
+                    async () => {
+                        const upload_params = {
+                            source_stream: buffer_utils.buffer_to_read_stream(object_info_hub.first_range_data),
+                            bucket: params.bucket,
+                            key: params.key,
+                            size: object_info_hub.size,
+                            content_type: params.content_type,
+                            xattr: object_info_hub.xattr,
+                            last_modified_time: (new Date(object_info_hub.create_time)).getTime(),
+                        };
+                        return this.namespace_nb.upload_object(upload_params, object_sdk);
+                    }
+                );
             } else if (cache_etag === '') {
                 object_info_hub.should_read_from_cache = false;
             } else {
@@ -286,14 +303,19 @@ class NamespaceCache {
                     'bucket',
                     'key',
                     'content_type',
+                    // The following fields are actually NOT set in the input "params".
+                    // Picking them is for the purpose of type validations in IDE such as VS Code,
+                    // so that IDE does not complain about undefined field.
                     'size',
                     'etag',
                     'xattr',
-                    'complete_upload'
+                    'complete_upload',
+                    'last_modified_time',
                 );
                 create_params.size = params.object_md.size;
                 create_params.etag = params.object_md.etag;
                 create_params.xattr = params.object_md.xattr;
+                create_params.last_modified_time = (new Date(params.object_md.create_time)).getTime();
                 create_params.complete_upload = true;
                 // Create partial object md
                 const create_reply = await object_sdk.rpc_client.object.create_object_upload(create_params);
@@ -428,6 +450,7 @@ class NamespaceCache {
                         { md_conditions: params.md_conditions, cache_object_md: params.object_md });
                 }
             } else {
+                upload_params.last_modified_time = (new Date(params.object_md.create_time)).getTime();
                 _global_cache_uploader.submit_background(
                     params.object_md.size,
                     async () => this.namespace_nb.upload_object(upload_params, object_sdk)
@@ -493,9 +516,6 @@ class NamespaceCache {
         const operation = 'ObjectCreated';
         const load_for_trigger = object_sdk.should_run_triggers({ active_triggers: this.active_triggers, operation });
 
-        function stream_final_from_promise(promise) {
-            return callback => promise.then(() => callback(), err => callback(err));
-        }
         const bucket_free_space_bytes = await this._get_bucket_free_space_bytes(params, object_sdk);
         let upload_response;
         let etag;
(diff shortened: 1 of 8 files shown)


src/sdk/namespace_cache.js
```diff
@@ -561,6 +568,16 @@ class NamespaceCache {
 
             if (cache_ok) {
                 assert.strictEqual(hub_res.value.etag, cache_res.value.etag);
+
+                const update_params = _.pick(_.defaults(
+                    { bucket: this.namespace_nb.target_bucket }, params), 'bucket', 'key', 'last_modified_time');
+                update_params.last_modified_time = (new Date(hub_res.value.last_modified_time)).getTime();
+                try {
+                    await object_sdk.rpc_client.object.update_object_md(update_params);
```
guymguym, line 576: We need to decide what order we prefer:
1. Complete the upload to an object, then update_object_md - opens up a small window where the last_modified_time can be returned from the cache incorrectly.
2. Update_*upload*_md, and then complete upload - need to add the update_upload_md() api for that.
Need to decide...

src/sdk/namespace_cache.js
```diff
@@ -561,6 +568,16 @@ class NamespaceCache {
 
             if (cache_ok) {
                 assert.strictEqual(hub_res.value.etag, cache_res.value.etag);
+
+                const update_params = _.pick(_.defaults(
+                    { bucket: this.namespace_nb.target_bucket }, params), 'bucket', 'key', 'last_modified_time');
```
guymguym, line 573: Need to match obj_id as well here, to avoid overriding another object with the same key (concurrent overwrites).

src/sdk/namespace_cache.js
paulxuzhang: @guymguym I pushed commit proposing solution addressing the issue in updating last_modified_time. Could you take a look at it? If good, we can apply the same to upload_object() calls in the read scenarios. 
We also need to handle the partial object creation case, propose to add last_modified_time in create_object_upload API for partial object (complete upload) scenario. So if partial object is created, last_modified_time needs to be provided just like etag. Thought?

src/server/object_services/object_server.js
```diff
@@ -78,7 +78,11 @@ async function create_object_upload(req) {
         if (!req.rpc_params.etag) {
             throw new RpcError('INVALID_REQUEST', 'etag must be provided when using complete_upload');
         }
+        if (!req.rpc_params.last_modified_time) {
+            throw new RpcError('INVALID_REQUEST', 'last_modified_time must be provided when using complete_upload');
+        }
```
guymguym, line 83: I think this check can be relaxed - it is common for our object_mds to have just create_time and then we use it instead of modify_time.

guymguym (changes requested):

src/sdk/namespace_cache.js
```diff
@@ -289,11 +306,13 @@ class NamespaceCache {
                     'size',
                     'etag',
                     'xattr',
-                    'complete_upload'
+                    'complete_upload',
+                    'last_modified_time',
```
guymguym, line 310: make a comment that we don't really expect it from params and will override anyway but is here for the type info.

src/sdk/object_io.js
```diff
@@ -211,6 +213,7 @@ class ObjectIO {
             'bucket',
             'key',
             'md_conditions',
+            'last_modified_time',
```
guymguym, line 216: Add also to create_params ... for symmetry.

src/server/object_services/object_server.js
paulxuzhang: As discussed, I will keep this changes. We can always relax it in the future if there is need.

src/sdk/namespace_cache.js
paulxuzhang: ok

src/sdk/object_io.js
paulxuzhang: ok

src/sdk/namespace_cache.js
```diff
@@ -243,6 +243,23 @@ class NamespaceCache {
...
+                        const upload_params = {
+                            source_stream: buffer_utils.buffer_to_read_stream(object_info_hub.first_range_data),
+                            bucket: params.bucket,
+                            key: params.key,
+                            size: object_info_hub.size,
+                            content_type: params.content_type,
+                            xattr: object_info_hub.xattr,
+                            last_modified_time: (new Date(object_info_hub.create_time)).getTime(),
+                        };
+                        this.namespace_nb.upload_object(upload_params, object_sdk);
```
guymguym, line 260: should we return this promise back from the submit handler?

guymguym (approved):

guymguym:
@paulxuzhang rebase and check the last comment above...

src/sdk/namespace_cache.js
paulxuzhang: Yes

paulxuzhang: @guymguym Resolved and rebased.

PR #6166, opened 2020-09-09
Prometheus reporting enhancements

nb-ohad:
### Explain the changes
- Support multiple separate reports each with different metrics, which includes all built-in metrics under one report and any number of custom metrics under multiple other reports
- For each process dedicate a port to serve its reports
- Use proxying to serve all core container reports (from all processes) using the webserver and under a structured URL path tree
- Deprecate old metrics path in web server but still serve them for backward compatibility

1. 

### Testing Instructions:
1. 

### DIFF
--- a/config.js
+++ b/config.js
@@ -395,7 +395,15 @@ config.MIN_MEMORY_FOR_UPGRADE = 1200 * 1024 * 1024;
 
 config.SEND_EVENTS_REMOTESYS = true;
 config.PROMETHEUS_ENABLED = true;
+config.PROMETHEUS_SERVER_RETRY_COUNT = Infinity;
+config.PROMETHEUS_SERVER_RETRY_DELAY = 5 * 60 * 1000; // 5 min
 config.PROMETHEUS_PREFIX = 'NooBaa_';
+
+// Ports where prometheus metrics are served
+config.WS_METRICS_SERVER_PORT = 7001;
+config.BG_METRICS_SERVER_PORT = 7002;
+config.HA_METRICS_SERVER_PORT = 7003;
+config.EP_METRICS_SERVER_PORT = 7004;
 
 //////////////////////////////
 // OAUTH RELATES            //
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -8,12 +8,14 @@ require('../util/dotenv').load();
 require('../util/panic');
 require('../util/fips');
 
+const dbg = require('../util/debug_module')(__filename);
+dbg.set_process_name('Endpoint');
+
 const http = require('http');
 const https = require('https');
 const FtpSrv = require('ftp-srv');
 const os = require('os');
 const P = require('../util/promise');
-const dbg = require('../util/debug_module')(__filename);
 const FuncSDK = require('../sdk/func_sdk');
 const ObjectIO = require('../sdk/object_io');
 const ObjectSDK = require('../sdk/object_sdk');
@@ -33,6 +35,8 @@ const md_server = require('../server/md_server');
 const auth_server = require('../server/common_services/auth_server');
 const server_rpc = require('../server/server_rpc');
 const { ENDPOINT_MONITOR_INTERVAL } = require('../../config');
+const prom_reporting = require('../server/analytic_services/prometheus_reporting');
+const config = require('../../config');
 
 const {
     ENDPOINT_BLOB_ENABLED,
@@ -86,17 +90,21 @@ function get_rpc_router(env) {
     };
 }
 
-function start_all() {
-    dbg.set_process_name('Endpoint');
+async function start_all() {
+    await Promise.all([
+        // Start the endpoint server
+        run_server({
+            s3: true,
+            lambda: true,
+            blob: ENDPOINT_BLOB_ENABLED,
+            ftp: ENDPOINT_FTP_ENABLED,
+            md_server: LOCAL_MD_SERVER,
+            n2n_agent: LOCAL_N2N_AGENT
+        }),
 
-    run_server({
-        s3: true,
-        lambda: true,
-        blob: ENDPOINT_BLOB_ENABLED,
-        ftp: ENDPOINT_FTP_ENABLED,
-        md_server: LOCAL_MD_SERVER,
-        n2n_agent: LOCAL_N2N_AGENT
-    });
+        // Try to start the endpoint metrics server
+        prom_reporting.start_server(config.EP_METRICS_SERVER_PORT)
+    ]);
 }
 
 async function run_server(options) {
--- a/src/hosted_agents/hosted_agents_starter.js
+++ b/src/hosted_agents/hosted_agents_starter.js
@@ -5,16 +5,14 @@ require('../util/dotenv').load();
 require('../util/panic');
 require('../util/fips');
 
-const url = require('url');
-
 const dbg = require('../util/debug_module')(__filename);
+dbg.set_process_name('HostedAgents');
+
+const url = require('url');
 const server_rpc = require('../server/server_rpc');
 const mongo_client = require('../util/mongo_client');
-
-dbg.set_process_name('HostedAgents');
-mongo_client.instance().connect();
-
-register_rpc();
+const prom_reporting = require('../server/analytic_services/prometheus_reporting');
+const config = require('../../config.js');
 
 function register_rpc() {
     server_rpc.register_hosted_agents_services();
@@ -26,3 +24,15 @@ function register_rpc() {
         logging: true,
     });
 }
+
+async function start_hosted_agents() {
+    await Promise.all([
+        mongo_client.instance().connect(),
+        register_rpc(),
+
+        // Try to start the hosted agents metrics server
+        prom_reporting.start_server(config.HA_METRICS_SERVER_PORT)
+   ]);
+}
+
+start_hosted_agents();
--- a/src/server/analytic_services/prometheus_reporting.js
+++ b/src/server/analytic_services/prometheus_reporting.js
@@ -1,485 +1,101 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
-const _ = require('lodash');
-const prom_client = require('prom-client');
-const config = require('../../../config.js');
+const dbg = require('../../util/debug_module')(__filename);
+const http = require('http');
+const config = require('../../../config');
+const promise_utils = require('../../util/promise_utils');
 
-function get_metric_name(name) {
-    return config.PROMETHEUS_PREFIX + name;
+// Import the reports.
+const { NodeJsReport } = require('./prometheus_reports/nodejs_report');
+const { NooBaaCoreReport } = require('./prometheus_reports/noobaa_core_report');
+const { NooBaaEndpointReport } = require('./prometheus_reports/noobaa_endpoint_report');
+
+// Currenty supported reprots
+const reports = {
+    nodejs: new NodeJsReport(),
+    core: null, // optional
+    endpoint: null // optional
+};
+
+function get_nodejs_report() {
+    return reports.nodejs;
 }
 
-const METRIC_RECORDS = Object.freeze([{
-    metric_type: 'Gauge',
-    metric_variable: 'cloud_types',
-    configuration: {
-        name: get_metric_name('cloud_types'),
-        help: 'Cloud Resource Types in the System',
-        labelNames: ['type']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'projects_capacity_usage',
-    configuration: {
-        name: get_metric_name('projects_capacity_usage'),
-        help: 'Projects Capacity Usage',
-        labelNames: ['project']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'accounts_usage_read_count',
-    configuration: {
-        name: get_metric_name('accounts_usage_read_count'),
-        help: 'Accounts Usage Read Count',
-        labelNames: ['account']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'accounts_usage_write_count',
-    configuration: {
-        name: get_metric_name('accounts_usage_write_count'),
-        help: 'Accounts Usage Write Count',
-        labelNames: ['account']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'accounts_usage_logical',
-    configuration: {
-        name: get_metric_name('accounts_usage_logical'),
-        help: 'Accounts Usage Logical',
-        labelNames: ['account']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'bucket_class_capacity_usage',
-    configuration: {
-        name: get_metric_name('bucket_class_capacity_usage'),
-        help: 'Bucket Class Capacity Usage',
-        labelNames: ['bucket_class']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'unhealthy_cloud_types',
-    configuration: {
-        name: get_metric_name('unhealthy_cloud_types'),
-        help: 'Unhealthy Cloud Resource Types in the System',
-        labelNames: ['type']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'object_histo',
-    configuration: {
-        name: get_metric_name('object_histo'),
-        help: 'Object Sizes Histogram Across the System',
-        labelNames: ['size', 'avg']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'providers_bandwidth_write_size',
-    configuration: {
-        name: get_metric_name('providers_bandwidth_write_size'),
-        help: 'Providers bandwidth write size',
-        labelNames: ['type']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'providers_bandwidth_read_size',
-    configuration: {
-        name: get_metric_name('providers_bandwidth_read_size'),
-        help: 'Providers bandwidth read size',
-        labelNames: ['type']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'providers_ops_read_num',
-    configuration: {
-        name: get_metric_name('providers_ops_read_num'),
-        help: 'Providers number of read operations',
-        labelNames: ['type']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'providers_ops_write_num',
-    configuration: {
-        name: get_metric_name('providers_ops_write_num'),
-        help: 'Providers number of write operations',
-        labelNames: ['type']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'providers_physical_size',
-    configuration: {
-        name: get_metric_name('providers_physical_size'),
-        help: 'Providers Physical Stats',
-        labelNames: ['type']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'providers_logical_size',
-    configuration: {
-        name: get_metric_name('providers_logical_size'),
-        help: 'Providers Logical Stats',
-        labelNames: ['type']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'system_capacity',
-    configuration: {
-        name: get_metric_name('system_capacity'),
-        help: 'System capacity',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'system_info',
-    configuration: {
-        name: get_metric_name('system_info'),
-        help: 'System info',
-        labelNames: ['system_name', 'system_address']
-    }
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'num_buckets',
-    configuration: {
-        name: get_metric_name('num_buckets'),
-        help: 'Object Buckets',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'total_usage',
-    configuration: {
-        name: get_metric_name('total_usage'),
-        help: 'Total Usage',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'accounts_num',
-    configuration: {
-        name: get_metric_name('accounts_num'),
-        help: 'Accounts Number',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'num_objects',
-    configuration: {
-        name: get_metric_name('num_objects'),
-        help: 'Objects',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'num_unhealthy_buckets',
-    configuration: {
-        name: get_metric_name('num_unhealthy_buckets'),
-        help: 'Unhealthy Buckets',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'num_unhealthy_pools',
-    configuration: {
-        name: get_metric_name('num_unhealthy_pools'),
-        help: 'Unhealthy Resource Pools',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'num_pools',
-    configuration: {
-        name: get_metric_name('num_pools'),
-        help: 'Resource Pools',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'num_unhealthy_bucket_claims',
-    configuration: {
-        name: get_metric_name('num_unhealthy_bucket_claims'),
-        help: 'Unhealthy Bucket Claims',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'num_buckets_claims',
-    configuration: {
-        name: get_metric_name('num_buckets_claims'),
-        help: 'Object Bucket Claims',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'num_objects_buckets_claims',
-    configuration: {
-        name: get_metric_name('num_objects_buckets_claims'),
-        help: 'Objects On Object Bucket Claims',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'reduction_ratio',
-    configuration: {
-        name: get_metric_name('reduction_ratio'),
-        help: 'Object Efficiency Ratio',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'object_savings_logical_size',
-    configuration: {
-        name: get_metric_name('object_savings_logical_size'),
-        help: 'Object Savings Logical',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'object_savings_physical_size',
-    configuration: {
-        name: get_metric_name('object_savings_physical_size'),
-        help: 'Object Savings Physical',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'rebuild_progress',
-    configuration: {
-        name: get_metric_name('rebuild_progress'),
-        help: 'Rebuild Progress',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'rebuild_time',
-    configuration: {
-        name: get_metric_name('rebuild_time'),
-        help: 'Rebuild Time',
-    },
-    generate_default_set: true,
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'bucket_healthy',
-    configuration: {
-        name: get_metric_name('bucket_status'),
-        help: 'Bucket Health',
-        labelNames: ['bucket_name'],
-    },
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'bucket_capacity_precent',
-    configuration: {
-        name: get_metric_name('bucket_capacity'),
-        help: 'Bucket Capacity Precent',
-        labelNames: ['bucket_name'],
-    },
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'bucket_quota_precent',
-    configuration: {
-        name: get_metric_name('bucket_quota'),
-        help: 'Bucket Quota Precent',
-        labelNames: ['bucket_name'],
-    },
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'resource_healthy',
-    configuration: {
-        name: get_metric_name('resource_status'),
-        help: 'Resource Health',
-        labelNames: ['resource_name'],
-    },
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'system_links',
-    configuration: {
-        name: get_metric_name('system_links'),
-        help: 'System Links',
-        labelNames: ['resources', 'buckets', 'dashboard'],
-    },
-}, {
-    metric_type: 'Gauge',
-    metric_variable: 'health_status',
-    configuration: {
-        name: get_metric_name('health_status'),
-        help: 'Health status',
-    },
-    generate_default_set: true,
-}]);
+function get_core_report(peek = false) {
+    if (!reports.core_report && !peek) reports.core_report = new NooBaaCoreReport();
+    return reports.core_report;
+}
 
+function get_endpoint_report(peek = false) {
+    if (!reports.endpoint && !peek) reports.endpoint = new NooBaaEndpointReport();
+    return reports.endpoint;
+}
 
-class PrometheusReporting {
-    static instance() {
-        PrometheusReporting._instance = PrometheusReporting._instance || new PrometheusReporting();
-        return PrometheusReporting._instance;
+function export_all_metrics() {
+    return Object.values(reports)
+        .filter(Boolean)
+        .map(report => report.export_metrics())
+        .join('\n\n');
+}
+
+async function start_server(
+    port,
+    retry_count = config.PROMETHEUS_SERVER_RETRY_COUNT,
+    delay = config.PROMETHEUS_SERVER_RETRY_DELAY
+) {
+    if (!config.PROMETHEUS_ENABLED) {
+        return;
     }
 
-    constructor() {
-        METRIC_RECORDS.filter(metric => metric.generate_default_set).forEach(metric => {
-            this[`set_${metric.metric_variable}`] = function(data) {
-                if (!this.enabled()) return;
-                this._metrics[metric.metric_variable].set(data);
-            };
-        });
-        this._prom_client = prom_client;
-        if (this.enabled()) {
-            this.define_metrics();
+    const server = http.createServer((req, res) => {
+        // Serve all metrics on the root path.
+        if (req.url === '' || req.url === '/') {
+            res.writeHead(200, { 'Content-Type': 'text/plain' });
+            res.end(export_all_metrics());
+            return;
         }
-    }
 
-    client() {
-        return this._prom_client;
-    }
-
-    enabled() {
-        return config.PROMETHEUS_ENABLED;
-    }
-
-    define_metrics() {
-        this._metrics = {};
-        prom_client.collectDefaultMetrics({ prefix: get_metric_name('defaults_') });
-
-        //Currently called from stats_aggregator, md_aggregator and other stats collectors are probebly as interesting to look at
-        METRIC_RECORDS.forEach(metric => {
-            this._metrics[metric.metric_variable] = new prom_client[metric.metric_type](metric.configuration);
-        });
-    }
-
-    export_metrics() {
-        const exported = _.map(METRIC_RECORDS, metric =>
-                this._prom_client.register.getSingleMetricAsString(metric.configuration.name))
-            .join('\n');
-
-        return exported;
-    }
-
-    set_cloud_types(types) {
-        if (!this.enabled()) return;
-        this._metrics.cloud_types.set({ type: 'AWS' }, types.pool_target.amazon);
-        this._metrics.cloud_types.set({ type: 'Azure' }, types.pool_target.azure);
-        this._metrics.cloud_types.set({ type: 'GCP' }, types.pool_target.gcp);
-        this._metrics.cloud_types.set({ type: 'Kubernetes' }, types.pool_target.kubernetes);
-        // TODO: Fill this up when we will know how to recognize
-        // this._metrics.cloud_types.set({ type: 'Ceph' }, types.pool_target.ceph);
-        this._metrics.cloud_types.set({ type: 'S3_Compatible' }, types.pool_target.s3_comp);
-    }
-
-    set_unhealthy_cloud_types(types) {
-        if (!this.enabled()) return;
-        this._metrics.unhealthy_cloud_types.set({ type: 'AWS' }, types.unhealthy_pool_target.amazon_unhealthy);
-        this._metrics.unhealthy_cloud_types.set({ type: 'Azure' }, types.unhealthy_pool_target.azure_unhealthy);
-        this._metrics.unhealthy_cloud_types.set({ type: 'GCP' }, types.unhealthy_pool_target.gcp_unhealthy);
-        this._metrics.unhealthy_cloud_types.set({ type: 'Kubernetes' }, types.unhealthy_pool_target.kubernetes_unhealthy);
-        // TODO: Fill this up when we will know how to recognize
-        // this._metrics.unhealthy_cloud_types.set({ type: 'Ceph' }, types.unhealthy_pool_target.ceph_unhealthy);
-        this._metrics.unhealthy_cloud_types.set({ type: 'S3_Compatible' }, types.unhealthy_pool_target.s3_comp_unhealthy);
-    }
-
-    set_bucket_class_capacity_usage(usage_info) {
-        if (!this.enabled()) return;
-        this._metrics.bucket_class_capacity_usage.reset();
-        for (let [bucket_class, value] of Object.entries(usage_info)) {
-            this._metrics.bucket_class_capacity_usage.set({ bucket_class }, value);
+        // Serve report's metrics on the report name path
+        const report_name = req.url.substr(1);
+        const report = reports[report_name];
+        if (report) {
+            res.writeHead(200, { 'Content-Type': 'text/plain' });
+            res.end(report.export_metrics(report_name));
+            return;
         }
-    }
 
-    set_projects_capacity_usage(usage_info) {
-        if (!this.enabled()) return;
-        this._metrics.projects_capacity_usage.reset();
-        for (let [project, value] of Object.entries(usage_info)) {
-            this._metrics.projects_capacity_usage.set({ project }, value);
-        }
-    }
+        res.writeHead(404, { 'Content-Type': 'text/plain' });
+        res.end();
+    });
 
-    set_accounts_io_usage(accounts_info) {
-        if (!this.enabled()) return;
-        this._metrics.accounts_usage_logical.reset();
-        this._metrics.accounts_usage_read_count.reset();
-        this._metrics.accounts_usage_write_count.reset();
-        accounts_info.accounts.forEach(account_info => {
-            const { account, read_count, write_count, read_write_bytes } = account_info;
-            this._metrics.accounts_usage_logical.set({ account }, read_write_bytes);
-            this._metrics.accounts_usage_read_count.set({ account }, read_count);
-            this._metrics.accounts_usage_write_count.set({ account }, write_count);
-        });
-    }
+    // Try to open the port for listening, will retry then exist.
+    while (retry_count) {
+        try {
+            await new Promise((resolve, reject) => {
+                server.listen(port, err => (err ? reject(err) : resolve()));
+            });
 
-    set_object_sizes(sizes) {
-        if (!this.enabled()) return;
-        for (const bin of sizes) {
-            if (!_.isEmpty(bin.bins)) {
-                for (const cur of bin.bins) {
-                    this._metrics.object_histo.set({ size: cur.label, avg: cur.avg }, cur.count, new Date());
-                }
+            // Set retry_count to 0 to exit the loop
+            retry_count = 0;
+
+        } catch (err) {
+            if (retry_count) {
+                dbg.error(`Metrics server failed to listen on ${port} (retries left: ${retry_count}), got`, err);
+                await promise_utils.delay_unblocking(delay);
+            } else {
+                dbg.error(`Metrics server failed to listen on ${port} too many times, existing the process`);
+                process.exit();
             }
         }
     }
-
-    set_providers_bandwidth(type, write_size, read_size) {
-        if (!this.enabled()) return;
-        this._metrics.providers_bandwidth_read_size.set({ type }, read_size);
-        this._metrics.providers_bandwidth_write_size.set({ type }, write_size);
-    }
-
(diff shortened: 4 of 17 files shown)


src/server/web_server.js
```diff
@@ -149,17 +155,21 @@ app.use(express_compress());
 ////////////
 // ROUTES //
 ////////////
-if (prom_reports.instance().enabled()) {
+if (prom_reporting.is_enabled()) {
+    // Enable proxying for all metrics servers
+    app.use('/metrics/web_server', express_proxy(`localhost:${config.WS_METRICS_SERVER_PORT}`));
```
nimrod-becker, line 160: 1. How does this appear in Prometheus ? Any diff between a the same metric exported here and via /metrics ?
2. Isn't /metric the default path Prometheus is scraping ? will it scape by default /metrics/web_server ?
3. What happens when both routes are scraped (old and new) duplication ? override ? (in prometheus)

src/server/web_server.js
```diff
@@ -149,17 +155,21 @@ app.use(express_compress());
...
+    app.use('/metrics/bg_workers', express_proxy(`localhost:${config.BG_METRICS_SERVER_PORT}`));
+    app.use('/metrics/hosted_agents', express_proxy(`localhost:${config.HA_METRICS_SERVER_PORT}`));
+
+    // TODO: this is here for backward competability, remove when all is working and tested
     app.get('/metrics', function(req, res) {
-        res.set('Content-Type', prom_reports.instance().client().register.contentType);
-        res.end(prom_reports.instance().client().register.metrics());
+        res.set('Content-Type', 'text/plain');
```
nimrod-becker, line 166: why ? what was the client (lib) default content-type ?

src/server/web_server.js
```diff
@@ -149,17 +155,21 @@ app.use(express_compress());
...
+        res.end(prom_reporting.export_all_metrics());
     });
-
     app.get('/metrics/counter', function(req, res) {
-        res.set('Content-Type', prom_reports.instance().client().register.contentType);
-        res.end(prom_reports.instance().export_metrics());
+        res.set('Content-Type', 'text/plain');
```
nimrod-becker, line 170: same as aboive

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -2,359 +2,423 @@
...
-        return config.PROMETHEUS_ENABLED;
-    }
+register_report(
+    config.NODEJS_METRICS_REPORT,
+    NodeJsReport
+);
 
-    define_metrics() {
-        this._metrics = {};
-        prom_client.collectDefaultMetrics({ prefix: get_metric_name('defaults_') });
```
nimrod-becker, line 340: isn't removing this means everything will be spilled directly on NooBaa_ ?
If so, I don't think its a good choice.

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -2,359 +2,423 @@
...
+    get metric_prefix() {
+        return `${super.metric_prefix}${dbg.get_process_name()}_`;
     }
+}
 
-    enabled() {
-        return config.PROMETHEUS_ENABLED;
-    }
+register_report(
+    config.NODEJS_METRICS_REPORT,
```
nimrod-becker, line 126: Why do we need to register the base ? its always there if I remember correctly

src/server/web_server.js
nb-ohad: currently, prom is only the old path this is why I still support it.
After I will push a PR to noobaa-operator I will get rid of the old paths

src/server/web_server.js
nb-ohad: text/plain

src/server/analytic_services/prometheus_reporting.js
nb-ohad: This is a bug. I missed it in the refactoring. 
Thank you

src/server/web_server.js
nimrod-becker: Have you verified we can change paths ? I mean besides the podmonitor which you have tested ...

src/server/analytic_services/prometheus_reporting.js
nb-ohad: we are not registering the base. We are registering the nodejs report which enables the default metrics.
They are not inheriting from each other

nimrod-becker: I will rephrase, isn't registring a customer collection (so core for example) bring with it the base ?

src/server/analytic_services/prometheus_reporting.js
nb-ohad: my mistake. this is not a bug each process gets it's own prefix using the process name. So it will be
`NooBaa_WebServer_*` for example

src/server/web_server.js
nb-ohad: yes. checked that it is possible. It is an edit on the service monitor CR. And it supports multiple scrape paths

src/server/analytic_services/prometheus_reporting.js
nb-ohad: No, not after the refactor. It is a diff report on a diff client registry

src/server/analytic_services/prometheus_reporting.js
nimrod-becker: How would endpoints look like if there are more than 1 ?
What would happen to customer metrics on webserver ? Path change ?

nb-ohad: each endpoint export its own metrics, same prefix, diff pod name.
The new path for  custom metrics is `/metrics/web_server/noobaa_core`.

In general, the path template for every metric  served from the `web_srever ` is `/metrics/<procsse_name>/<report_name>`
To consume all  metrics from all reports for a process the path will be `/metrics/<procsse_name>`

But in this PR I still support the old path `/metrics` and `/metrics/counter`

src/server/analytic_services/prometheus_reporting.js
nimrod-becker: from the prom client module, I might be mistaken, but I remember adding anything automatically adds the node.js counters

nb-ohad: Only if you add to the default registry and collect also the nodejs metrics into the same registry (which is the default behavior)

src/server/analytic_services/prometheus_reporting.js
nimrod-becker: I understand the route
I'm wandering about the prometheus side.

How would metrics, 
- lets say NooBaa_X previously would look like now
- metrics from the endpoint in case more than 1 exists

nb-ohad: Ok, let me explain:
- The only metrics that were consumed are the custom ones from the code which stayed exactly the same just served from a new route.
- the default metrics change their prefix, but nobody that I know of ever consumed them
- The endpoint metrics will be differentiated by pod name/id (all of them will have the `NooBaa_Endpoint_*` prefix

nimrod-becker: Ok, can we have the node.js metrics in the manner of
NooBaa_Endpoint_<ID>_NODE_

and core
NooBaa_NODE_

and bg
NooBaa_BG_NODE_

etc, ?

there's quite a lot of counters on the default node.js part, lets group them under a sub topic

nb-ohad: What do you mean? we do not have control over the individual default nodejs metric.
They are defined and managed by the prom client

nb-ohad: Also, what does NODE represent in your answer? I do not get it. The name of the default metric is defined by the client

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -2,359 +2,423 @@
...
+                ...m.configuration,
+                name: `${this.metric_prefix}${m.name}`,
+                registers: [this._registry]
+            });
+
+            // Create a default setter, are moving the name into a local var
+            // in order to prevent the closure to capture the entire m object.
+            const { name } = m;
+            if (m.generate_default_set) {
+                this[`set_${m.name}`] = data => this._metrics[name].set(data);
```
paulxuzhang, line 416: New function is added to class. Would we have performance related issue if there are many metrics? Would it be simple to just provide function in the base class e.g.? 
Also is set() only for Gauge type of metric? we should call corresponding function based upon metric type.

```
function observe(labels, data) {
   if (this._metrics[name].type === 'Counter') {
       this._metrics[name].inc(labels, data);
   }
   else if (this._metrics[name].type === 'Gauge') {
       this._metrics[name].set(labels, data);
   } else if (this._metrics[name].type === 'Histogram') {
       this._metrics[name].observe(labels, data);
   }
```

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -2,359 +2,423 @@
...
+        super();
 
-        return exported;
+        this._metrics = {};
+        for (const m of NOOBAA_CORE_METRICS) {
+            this._metrics[m.name] = new prom_client[m.type]({
+                ...m.configuration,
+                name: `${this.metric_prefix}${m.name}`,
+                registers: [this._registry]
+            });
```
paulxuzhang, line 410: Those are common functionality. Can we move those into the base class BasePrometheusReport so that we don't duplicate code in inheriting classes.

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -461,25 +514,33 @@ class PrometheusReporting {
...
     update_providers_ops(type, write_num, read_num) {
-        if (!this.enabled()) return;
         this._metrics.providers_ops_read_num.inc({ type }, read_num);
         this._metrics.providers_ops_write_num.inc({ type }, write_num);
     }
 }
 
-// EXPORTS
-exports.PrometheusReporting = PrometheusReporting;
+register_report(
```
paulxuzhang, line 534: What is the purpose of prometheus_reporting.js? It seems that we should refactor this file. If endpoint imported it, endpoint would register NooBaaCore metrics. We would not want that.  We can,
- Only keep the common functions (e.g. start_server) and class BasePrometheusReport in the file
- Move inheriting class (e.g.) NooBaaCoreReport to the space of its corresponding service
- Move calling register_report() to the space of corresponding service

src/server/analytic_services/prometheus_reporting.js
nimrod-becker: Can you paste, from prometheus the full one (metric name and value) for one of the core custom metrics and one of the core node.js metrics ?

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -2,359 +2,423 @@
...
-        labelNames: ['type']
+    return report.instance;
+}
+
+function export_report_metrics(report_name) {
+    const report = get_report(report_name, false);
+    return report ? report.export_metrics() : '';
+}
+
+function export_all_metrics(report_name) {
```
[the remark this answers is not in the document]
nb-ohad: You are correct. Thanks will remove the parameter

config.js
```diff
@@ -397,6 +397,16 @@ config.SEND_EVENTS_REMOTESYS = true;
 config.PROMETHEUS_ENABLED = true;
```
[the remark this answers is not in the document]
nb-ohad: Not sure, 
You can just configure Prometheus to scrape only the endpoints for some services/reports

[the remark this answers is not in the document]
nimrod-becker: Since this config is build time, I agree with Ohad and I Don't see the need to break into different ones.
If we would need it in the future, we'd need to move to env based anyway

src/server/analytic_services/prometheus_reporting.js
nb-ohad: @paulxuzhang I also don't like adding methods on the fly like that. 
It is an artifact of the old code which gathered metrics for core custom metrics. 
I didn't want to touch it in order to not break the existing metrics collection.

Regarding performance: each report is a singleton which means that only one instance of the class and each of the methods will be instantiated. And because of the way javascript works, a method on an object might be even faster than a method on the object's prototype. But in modern js engines (like V8) it does not really matter and the JIT might even inline in the call site in both cases

src/server/analytic_services/prometheus_reporting.js
nb-ohad: I don't think this should be a common functionality because I would like to implement the new reports differently.
I think that trying to be "metadata-driven" in this case is a bad choice. The code of a report can be greatly simplified by just creating the metric itself in the constructor. For example for new reports I will do the following in the report constructor:
```
// Define storage for the metrics:
this._metrics = {
    // This is a dummy metric that is used as an example, Gauge is use
    // here forbut it can be any of the metric types defined in the prom
    // client lib at https://github.com/siimon/prom-client with its
    // specific configuration.
    demo_metric: new client.Gauge({
        name: this.get_prefixed_name('demo_metric'),
        registers: [this._registry],
        help: 'Description line for demo_metric',
        // Other configurqation that should apply for this metric
    }),
};
```

Again, I didn't want to touch the existing metrics.

src/server/analytic_services/prometheus_reporting.js
nb-ohad: prometheus_reporting.js is the overall service which gives you the ability access, export, and exposes reports over the network for scraping and endpoint should import it.

Registring a report does not mean that we are collecting its metrics or even instantiating the report class.
If you look at:
```
function get_report(name, create_non_existing = true) {
    const report = reports.get(name);
    if (!report) {
        throw new Error(`get_report() - No report registered for name ${name}`);
    }

    if (!report.instance && create_non_existing) {
        report.instance = new report.Type();
    }
    return report.instance;
}
```

You can see that a report is instantiated only if you try to access it, so which report/metrics are collected and exposed depends on the code that runs in a give process.

src/server/analytic_services/prometheus_reporting.js
nimrod-becker: I think its easier to understand, maintain, and add custom metrics when they are written in a JSON format in a human-readable manner rather than code.

src/server/analytic_services/prometheus_reporting.js
nb-ohad: Ok, I am pasting one metric from each report that we expose:

NodeJS report for WebServer process:
`NooBaa_WebServer_nodejs_eventloop_lag_seconds 0.006712859`

NodeJS report for BgWorker process :
`NooBaa_BGWorkers_nodejs_eventloop_lag_seconds 0.014447736`

NodeJS report for HostedAgents process:
`NooBaa_HostedAgents_nodejs_eventloop_lag_seconds 7.513233968`

NodeJS report for Endpoint process:
`NooBaa_Endpoint_nodejs_eventloop_lag_seconds 2.703802685`

NooBaa Core report (custom metrics):
`NooBaa_providers_bandwidth_read_size{type="KUBERNETES"} 2347008`
(the prefix here is only "NooBaa" in order to not break current consumers)

NooBaa Endpoint report (custom metrics):
will start with  `NooBaa_` but the full matrix can be decided when we will go to implement it

nimrod-becker: ok, and we plan to keep NooBaa_ for all webserver's custom metrics right ? 
Then all is good if the answer is yes

src/server/analytic_services/prometheus_reporting.js
nb-ohad: I must disagree writing it as part of the constructor like that:
```
this._metrics = {
   demo_metric: new client.Gauge({
       name: this.get_prefixed_name('demo_metric'),
       registers: [this.registry],
       help: 'Description line for demo_metric',
       // Other configurqation that should apply for this metric
       }),
 };
```
has the exact same benefits and the exact same number of lines of codes and clarity. 

And trying to generate function automatically will only make things more confusing for the reader/consumer:
- You cannot find all the setter when reading the class and even worse when some of them are also defined in the class (the more complicated ones)
- The JS linter cannot know about these methods 
- You will always have exceptions that you will implement in the class itself which means that in order to check for a method you need to understand that some of them are generated and some not. 
These exceptions just complicate things even more because  now the reader needs to check a flag on the "metadata" in order to understand if a setter is generated or not

I believe that this kind of generation just add to the code complexity and I strongly suggest not to use it in future reports

config.js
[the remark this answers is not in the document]
nb-ohad: I just want to add that even if we override the config using an env variable the pod will restart. 
This is the way k8s works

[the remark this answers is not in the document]
nimrod-becker: That is fine, we might need to be able to be more selective on what we collect in the future (we might not as well). Let's handle it if and when the need arises

src/server/analytic_services/prometheus_reporting.js
nb-ohad: The answer is YES, 
I would have loved to address `core` to the custom metrics but I don't want to break existing consumers

src/server/analytic_services/prometheus_reporting.js
nb-ohad: @paulxuzhang 
I just like to add that I pushed a commit that split prometheus_reporting.js and moves reports codes into their own files.
This is not exactly what you suggested as we are still registering all reports but I think my last comment can explain the reasoning

src/hosted_agents/hosted_agents_starter.js
```diff
@@ -26,3 +24,13 @@ function register_rpc() {
         logging: true,
     });
 }
+
+async function start_hosted_agents() {
+    mongo_client.instance().connect();
+    register_rpc();
+
+    // Try to start the hosted agents metrics server
+    await prom_reporting.start_server(config.HA_METRICS_SERVER_PORT);
```
jeniawhite, line 33: What happens if you fail? You should handle that.

src/hosted_agents/hosted_agents_starter.js
```diff
@@ -26,3 +24,13 @@ function register_rpc() {
...
+}
+
+start_hosted_agents();
```
jeniawhite, line 36: In Endpoint, you want on it. Here not. 
Is there a reason?

src/server/bg_workers.js
```diff
@@ -209,19 +211,26 @@ function run_master_workers() {
...
+        run_immediate: true
+    }, cluster_hb.do_heartbeat);
 
-register_bg_worker({
-    name: 'cluster_heartbeat_writer',
-    delay: config.CLUSTER_HB_INTERVAL,
-    run_immediate: true
-}, cluster_hb.do_heartbeat);
+    // Try to start the bg workers metrics server
+    prom_reporting.start_server(config.BG_METRICS_SERVER_PORT);
```
jeniawhite, line 228: Same comment as above. 
What happens if we fail.

src/server/bg_workers.js
```diff
@@ -209,19 +211,26 @@ function run_master_workers() {
...
+
+    dbg.log('BG Workers Server started');
+}
 
-dbg.log('BG Workers Server started');
+start_bg_workers_server();
```
jeniawhite, line 233: Same comment as above, is there a reason for not waiting?

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -1,485 +1,115 @@
...
-        this._metrics.bucket_quota_precent.reset();
-        this._metrics.bucket_capacity_precent.reset();
-        buckets_info.forEach(bucket_info => {
-            this._metrics.bucket_healthy.set({ bucket_name: bucket_info.bucket_name }, Number(bucket_info.is_healthy));
-            this._metrics.bucket_quota_precent.set({ bucket_name: bucket_info.bucket_name }, bucket_info.quota_precent);
-            this._metrics.bucket_capacity_precent.set({ bucket_name: bucket_info.bucket_name }, bucket_info.capacity_precent);
-        });
-    }
+    return new Promise((resolve, reject) => {
+        server.listen(port, err => {
```
jeniawhite, line 74: Maybe we should handle the failure and attempt to bring it up later on? After a delay or something? 
Comes back to my other comments regarding waiting and handling failures in different processes.

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
```diff
@@ -0,0 +1,44 @@
...
+        // Cache the client for ease of use.
+        const client = this.prom_client;
+
+        // Define storage for the metrics:
+        this._metrics = {
+            // This is a dummy metric that is used as an example, Gauge is use
+            // here forbut it can be any of the metric types defined in the prom
+            // client lib at https://github.com/siimon/prom-client with its
+            // specific configuration.
+            demo_metric: new client.Gauge({
```
jeniawhite: I think that this is more complex and prone to errors, rather than using the JSON configuration as we do in the NooBaa Core metrics.
This is just writing inline everything including object creation, instead of configuring that in a JSON and letting a dynamic code to configure all of that.

src/server/node_services/nodes_monitor.js
```diff
@@ -228,8 +228,9 @@ class NodesMonitor extends EventEmitter {
             );
             nodes_stats.forEach(stats => {
                 dbg.log0(`resetting stats in prometheus:`, stats);
-                prom_report.instance().set_providers_ops(stats.service, stats.write_count, stats.read_count);
-                prom_report.instance().set_providers_bandwidth(stats.service, stats.write_bytes, stats.read_bytes);
+                const core_report = prom_reporting.get_report(config.NOOBAA_CORE_METRICS_REPORT);
```
jeniawhite, line 231: If you are using the same report twice, maybe it is better to require if so you won't get in multiple times from `prom_reporting.get_report `?

src/server/system_services/stats_aggregator.js
```diff
@@ -910,43 +910,45 @@ function partial_cycle_parse_prometheus_metrics(payload) {
...
-    prom_report.instance().set_object_savings_logical_size(logical_size);
-    prom_report.instance().set_bucket_class_capacity_usage(usage_by_bucket_class);
-    prom_report.instance().set_projects_capacity_usage(usage_by_project);
-    prom_report.instance().set_accounts_io_usage(accounts_stats);
-    prom_report.instance().set_accounts_num(accounts_num);
-    prom_report.instance().set_total_usage(total_usage);
-    // TODO: Currently mock data, update with the relevant values.
-    prom_report.instance().set_rebuild_progress(100);
-    prom_report.instance().set_rebuild_time(0);
+    const core_report = prom_reporting.get_report(config.NOOBAA_CORE_METRICS_REPORT);
```
jeniawhite, line 913: Same as the previous comment regarding multiple `prom_reporting.get_report`

src/server/bg_workers.js
nb-ohad: same answer

src/hosted_agents/hosted_agents_starter.js
nb-ohad: I assume you mean "wait on it" T
We never wait in the global scope because NodeJS does not allow it (in the current version)
But even if it will allow it I see no reason. Starting a service is "call and forget about it" thingy

src/server/analytic_services/prometheus_reporting.js
nb-ohad: I do not agree,
If the port is taken it will not suddenly be available. 
If it will we have an issue we need to deal with and not just hide it be "trying later"

jeniawhite: So deal with it instead of trying later.
I do not see you dealing with it.
The current state is that you do not deal with it and run without metrics.

src/server/node_services/nodes_monitor.js
nb-ohad: 1. It is a bad practice to add code to require as it will make it much harder to move native imports.
2. we can move this to global scope but I don't see the value. It is like advocating that in the old code we should not call `instance()` more than once.

src/server/system_services/stats_aggregator.js
nb-ohad: same answer

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -1,485 +1,111 @@
 /* Copyright (C) 2016 NooBaa */
```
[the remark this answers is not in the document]
nb-ohad: The reports are the singletons. prom reporting is just provided utils which makes it by design a singleton

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
nb-ohad: @jeniawhite It is almost exactly the same text, you don't gain anything by removing it outside you are just making it harder to mentally associate the metrics.

It also comes with a baggage of cons, please see the discussion above with paul regarding this issue

src/server/analytic_services/prometheus_reporting.js
nimrod-becker: I understand the argument about time constraint and not wanting to handle errors properly now, that is valid and I agree with it.
However, I don't want a one try here, which we know would fail at some point. I don't think we know enough and not it's not just about port taken and then being avail. 

Please add a retry mechanism, we can do 10s between each retry and we can limit on, say 30 such retries (which would account for an entire aggregation cycle). We should also log in high severity that we failed between retries. That should not consume much time to add.

nb-ohad: @nimrod-becker I highly disagree with that:

A retry in any case that should not fail (like opening a port) will only hide a real problem.
Think about a race condition between agents (not noobaa agent, conceptual ones) that try to open the same port.
The best caser scenario, with respect to support and debugging, is that the issue can be reproduced in a consistent manner. Adding a retry loop can and will hide such collisions if the other side is also doing the same. 

All of the above will lead to unreproducible bugs, on even to 2 different runs on the same machines that yield different results! this is very bad for us, for QE, and for the customer.

In my opinion, a retry should be added only if we know that an action would fail (not should) and that the situation is temporary (like connectivity to the cloud). 
I cannot say this is the case when trying to start a server and you can even see that we took the same decision with our  HTTP/HTTPS servers (see endpoint code).

We should handle edge cases (error handling)  when we will encounter them (not upfront). Then if we have a situation we know how to handle,  a retry will be a good solution after the error handling.

Saying all of the above,
Let leave it as is

config.js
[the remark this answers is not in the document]
paulxuzhang: Ok.

src/server/analytic_services/prometheus_reporting.js
paulxuzhang: ok

src/server/analytic_services/prometheus_reporting.js
paulxuzhang: ok

src/server/analytic_services/prometheus_reporting.js
jeniawhite: @nb-ohad 
```
It is a different case altogether, in the agent case you are trying to establish a connection here you are trying to listen on a port. Trying to establish a connection might be affected by network hiccups that might resolve with a retry.
A retry IS THE FIX in that case but not in this one.
```
So you are saying there couldn't be any hiccups that can be resolved by a retry when you attempt to listen to a port?

```
Other failures are persistent and will fail in every retry.
Retry or not it will not make a difference.
```
The same way as establishing connections can be persistent and cannot be fixed by retrying.
And I do not agree with that. Listening on ports can be non-persistent issues.

```
Trying to listen on a port by more than one actor is not persistent and prune to races.
Total network issues, like problematic network adaptors, will fail our actual server not only the metrics server.
```
```
Trying to listen on a port by more than one actor is not persistent and prune to races.
Total network issues, like problematic network adaptors, will fail our actual server not only the metrics server.
```
You keep on locking down to multiple processes that listen on the same port.
Obviously it won't fail on the metrics server, because you wouldn't deploy it since you give up immediately.
There could be problems with acquiring ports on certain machines that can be fixed and retried and are not persistent, already had customers with such scenarios.
You keep on stating that port listening is a persistent problem that can only happen if someone is acquired that port and if not then it is not our problem.

```
Under the assumption that there are bugs in the product, ask a customer if he/she prefers a product with a hidden bug that a retry might be or not be a workaround but with longer support time and a "half" ability to diagnose or a product that will fail consistently but with a better diagnose and response time from support.
```
First of all, unknown bugs are unknown... So there is not much said there. Let's assume that there are bugs.
Let's gather customers and ask them then (I assume not feasible, but we can try).
I do not see why adding a retry will cause "longer support time and a "half" ability to diagnose or a product that will fail consistently".
Please explain how adding retry will cause: a longer cycle of support, make the product not fail consistently.
If the customer has a consistent problem then it will fail consistently with the retry and without.
If the problem is not consistent then it will not fail consistently with and without the retry.
A longer cycle of support because of adding a retry? Why? You have the same debug logs and information in both cases (with and without retry).

```
If you think metrics are so valuable you should advocate that a failure to start the metrics server should be a service failure and the service should be stoped. I don't believe that you believe that, but please correct me if I am wrong.
```
I believe that metrics are valuable enough for us to try and redeploy them several times instead of throwing our hands on the first failure. I'm not in any way saying that the failure to deploy a metrics server should block the working of NooBaa, all I'm saying that we shouldn't give up on the first try.

```
Two options:

1) The retry always works - you will never know of the underlying bug.
2) The retry might work but not always, reproducing the problem and getting consistent results can be very problematic, and because this might happen a long time from now, the chance that we will immediately understand that we have retries that mask the problem is almost non-existent, we will waste a long time in just trying to reproduce and diagnose why it is happening only sometimes.
```
1) The retry always works - you have the same information as you did without the retry.
2) If the problem is consistent you have consistent results with or without retry. You will understand that we have retries because you can read the code (if you want you can add the retry number in the error print). 
Reproduction is the same for both cases with or without the retry. You will need to diagnose if it happens only sometimes both with and without the retry (hence the persistent or non-persistent error).

```
The underlaying this here is that this case does not fall under the category of "retry as a solution" (like establishing a connection to a server) but into the category "retry as a workaround" and a workaround should be employed after you have a specific failure case and in that specific case, you want to dim a retry as a viable workaround. not doing so and applying a retry loop to all cases will hide bigger problems that should be addressed.
```
I'm not quite sure why you consider retrying agent initial connection as a "solution", yet retry on initial listening on ports as a "workaround".
We don't even deal with the specific error on agent initial connect and then retry, it is literally the same.
By your logic of "workaround" and "solution", there shouldn't be any error handling in code unless you know all of the possible errors fix them, and then retry.
Did I understand you correctly?

nb-ohad: @jeniawhite 
This conversation does not go anywhere and I did not find any new information or insight in your last post that would convince me otherwise.
 
The bottom line and my general belief are that we should always strive (even if it is not always possible) for a system that behaves consistently. 

If something should fail for an unknown reason it should always fail for that reason and I don't care what it is.
if we know of a potential reason and decide to fortify the product against it then we should do it, even if we decide that for that specific error the solution is a retry.

But what we should not do is play with dices, it will create a system that behaves unexpectedly in all cases. These kinds of systems are hard to diagnose and debug. And the turn-around time for solution is always bigger in such systems.

jeniawhite: @nb-ohad 
We've already had another point of view from @nimrod-becker 
We should bring it up with other team members in addition to @guymguym , as well then.

guymguym: @jeniawhite @nb-ohad You've summoned me. 
What can I do for you? You have 3 wishes (per PR), use them wisely :)

nimrod-becker: 3 Wishes are enough :) We need 1 I think.

Jenia and I seem to think we better handle a potential error on the listen in the form of a few more retries (not forever, we can decide how many and how much backoff).

Ohad raises the point that we better fail immediately since its both very unexpected to fail to listen and it can mask the original problem (Ohad I hope I did not misrepresent your opinion)

nb-ohad: @guymguym a better description of my point of view can be summarized in my last comment above

guymguym: OK. I've taken a close look.

I think the theological discussion, while important, is hard on this platform, so I would encourage you to raise such topics in the next community meeting - just add it to the agenda and take 10 minutes to discuss about it.

I think if we enabled this feature and it fails to start up correctly, we should exit.
See how this is done in web server (regardles that we should deprecate the use of P.ninvoke)
https://github.com/noobaa/noobaa-core/blob/22de4be4bdf1b0cfac7d89dda371f67e6e178264/src/server/web_server.js#L91-L94

src/server/web_server.js
```diff
@@ -90,10 +91,14 @@ async function start_web_server() {
...
         server_rpc.rpc.register_ws_transport(https_server);
         await P.ninvoke(https_server, 'listen', https_port);
+
     } catch (err) {
         dbg.error('Web Server FAILED TO START', err.stack || err);
         process.exit(1);
     }
+
+    // Try to start the metrics server.
+    prom_reporting.start_server(config.WS_METRICS_SERVER_PORT);
```
guymguym, line 101: missing await

guymguym: actually why not move it under the same try catch as above?

src/server/bg_workers.js
guymguym, line 228: await

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
```diff
@@ -0,0 +1,44 @@
...
+    get registry() {
+        return this._registry;
+    }
+
+    get metric_prefix() {
+        return config.PROMETHEUS_PREFIX;
+    }
+
+    get enabled() {
+        return is_enabled();
```
guymguym, line 30: Seems like this is just propagating the same config value to another global function and here also to a report method although there is no "enabled" state per report, which got me a bit confused. Is this here for convenience? or maybe can be removed so that global configs will be accessed through the main singleton instance in prometheus_reporting.js

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
```diff
@@ -0,0 +1,44 @@
...
+const prom_client = require('prom-client');
+const config = require('../../../../config.js');
+
+function is_enabled() {
+    return config.PROMETHEUS_ENABLED;
+}
+
+class BasePrometheusReport {
+    constructor() {
+        const client = this.prom_client;
```
guymguym, line 13: What am I reading here. This line make any sense?

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
guymguym: Same for metric_prefix()

guymguym (changes requested):
Hey @nb-ohad 

I reviewed and added a variety of important vs. nit picking comments above :)

All in all I think the PR is great and the capability it awesome. Also thanks a lot for adding the example metric for the endpoint - this will speed up things for @paulxuzhang ...

But I think there is room to improve with regards to:
1. Some error handling seems to be missing, mostly around the start calls.
1. No tests - how will we trust this feature is still working overtime? I suggest to enable this code to run in coretest and create a test that scrapes once to verify all the reports are returned and have valid data.

What do you think about adding a test case as I suggested - @nb-ohad and @nimrod-becker ?

src/server/analytic_services/prometheus_reporting.js
nimrod-becker: To sum up yesterday's discussion

1) we would add 5 retries with 20s between each retry, print each failure to the log on high sev
2) we would not fail the server since metrics unavailability (even for dashboards and optimization) would be better then data unavailability if core can't come up

src/hosted_agents/hosted_agents_starter.js
```diff
@@ -26,3 +24,13 @@ function register_rpc() {
         logging: true,
     });
 }
+
+async function start_hosted_agents() {
```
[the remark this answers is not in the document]
nb-ohad: This is a function that is called from the global scope who cannot wait so waiting or not waiting does not make a difference. 

Before the change all of the calls where separately on the global scope (again without the ability to wait) and I did not change it, as I wrote above, it will be meaningless because we have nothing to do after we wait.

If you insist I can add the wait

config.js
```diff
@@ -397,6 +397,17 @@ config.SEND_EVENTS_REMOTESYS = true;
...
+// Report names
+config.NODEJS_METRICS_REPORT = 'nodejs';
+config.NOOBAA_CORE_METRICS_REPORT = 'noobaa_core';
+config.NOOBAA_ENDPOINT_METRICS_REPORT = 'noobaa_endpoint';
+
+// Ports where prometheus metrics are served
+config.WS_METRICS_SERVER_PORT = 7001;
+config.BG_METRICS_SERVER_PORT = 7002;
+config.HA_METRICS_SERVER_PORT = 7003;
+config.EP_METRICS_SERVER_PORT = 7004;
```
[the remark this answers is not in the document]
nb-ohad: One container may, and will in our case, run more than one server.
Our core container will run 3 servers, one for each process.

src/server/web_server.js
nb-ohad: Failing to report metrics, from a customer side, has different much lower severity then denial of service, so we agreed (RedHat's noobaa team) that it should not exist in that case. 
This is also why I didn't merge the code with the server code try and catch block. 

Now  `start_server` cannot fail because it will catch all rejects, log them, and then ignore. which means await is not needed as it is an "initiate and don't wait for" kind of operation.

src/server/bg_workers.js
nb-ohad: See reply from the last comment

src/hosted_agents/hosted_agents_starter.js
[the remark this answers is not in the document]
guymguym: If all I need to do is insist then I insist :)
My point it that async functions better be awaiting promises and not spread them around careleslly.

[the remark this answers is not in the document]
guymguym: And of course it wans't your PR that created this code, but I'm trying to align that code to be more predictable as you also desire.

config.js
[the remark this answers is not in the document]
guymguym: So I would make BG add 1 to the base port, and hosted agents add 2...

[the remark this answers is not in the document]
guymguym: I mean - we can set it with:
```
config.BG_METRICS_PORT_INC = 1;
config.HOSTES_AGENTS_METRICS_PORT_INC = 2;
```
but I think the base metrics port is best be controlled from env to make it easy to configure for different environments...

src/server/web_server.js
guymguym: But the current code just leaves it unhandled, and even with retries this will eventually fail, and then what?

guymguym: We need better error handling here.

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
nb-ohad: Regarding: `is_enabled/enabled`:
I want outside code to consume the enabled from `prometheus_reporting.js` as it is the interface for the whole system but I cannot implement it in `prometheus_reporting.js` because I want to consume it in the reports` implementations which will create a cyclic dependency that will fail to resolve (as some of this code run in the parse phase. 

I want a single truth and I wanted it to not be a config var but an abstraction on top of that because we have plans to change it in the near future and we might localize it per a report, on top of the global config.

regarding: `metrics_prefix`: 
It is not the same for all reports, and the idea was to let report override the base class prefix, you can see and example in  the `nodejs` report

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
nb-ohad: I am localizing the client into a local var for 2 reasons:
1. it is going to be accessed in a loop immediately after. 
2 it is going to be used with the `new` keyword which makes the code `new client[type]` much more readable than `new this.prom_client[type]` because it removes the ambiguity of what is being "newed"

guymguym: I don't see any loop after...

src/endpoint/endpoint.js
```diff
@@ -86,9 +90,7 @@ function get_rpc_router(env) {
     };
 }
 
-function start_all() {
-    dbg.set_process_name('Endpoint');
-
+async function start_all() {
     run_server({
```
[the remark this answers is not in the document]
nb-ohad: A fully agree that we need some mark that says that the intention here is not to wait. I will add a comment for now (although this was not part of my change at all)  but I must say I like your idea with a `noawait` call which should be mapped to a `noop`

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
guymguym: These are a lot of big words for a flag which is in config...
Why aren't you just using config.PROMETHEUS_ENABLED wherever you need to use it?
We are getting too many new lines of code for a very slight value.

src/endpoint/endpoint.js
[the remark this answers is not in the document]
guymguym: so why don't you add a noawait function? why would you settle for a comment?

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -1,485 +1,111 @@
...
-                if (!this.enabled()) return;
-                this._metrics[metric.metric_variable].set(data);
-            };
-        });
-        this._prom_client = prom_client;
-        if (this.enabled()) {
-            this.define_metrics();
-        }
-    }
+const reports = new Map();
```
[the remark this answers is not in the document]
nb-ohad: I can understand the pros and cons of having them as a global state and not stored in a map.
On the plus side, I like that the code will be simpler, with no need for name strings of config values for the name.
On the minus side, it will make adding new reports much more verbose and troublesome because you need to be aware of each report individually. 

I think in this case the pros are greater than the cons, so I will do the change, but In any case, I don't see any value in adding a `Reports` class. 
My change will be to have, for each report a global variable and a `get_XXXX_report` function exposed in `prometheus_reporting.js`. For 3 reasons:
1. calling code will be simpler, instead of `prom_reporting.Reports.instance.get_XXXX_report()` it will be just `prom_reporting.get_X_report()`
2. `prom_reporting` is acting more like a utils collection at this point.
3 . Adding a class that is a singleton is kind of meaningless in the nodejs world where you have module scope which acts like a singleton. I can understand the need in other languages like C or C++ where you do not have the semantics of just running code in the global scope.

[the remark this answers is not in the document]
guymguym: Well, not only, singletons and classes are better for creating tests.
I don't care for singletons that prevent creation of more objects, only such that have the default instance for the entire system to easily locate it.
Using classes allows you to do better whitebox testing by replacing some functions in it to mock some behaviors.
When you'll get to testing this, you'll see :)

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
nb-ohad: You are correct. I was thinking about `noobaa_endpoint_report.js`. I will remove

src/server/web_server.js
nb-ohad: @guymguym It is handled, `start_server` literally cannot fail (in the current implementation)

src/server/analytic_services/prometheus_reporting.js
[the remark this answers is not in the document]
nb-ohad: I agree about white boxing on the reports themselves not utils.
> Using classes allows you to do better whitebox testing by replacing some functions in it to mock some behaviors.
When you'll get to testing this, you'll see ;)

In js the simplest thing to mock is a module function, you just replace the function on the exports objects

src/server/web_server.js
guymguym: still, when you don't await an async function it just feels wrong.
both for the possible new error conditions we might introduce inside start_server as time goes by,
and also because we might need to add more setup code and making it ordered is simpler.

config.js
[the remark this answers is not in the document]
nb-ohad: @guymguym 
I am not sure if you know it or not, but we can change and of the values defined in `config.js` with an env variable.

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
nb-ohad: Ok, I will use `config.PROMETHEUS_ENABLED` from the outside and in the base. I still want every derived class to be able to override if different logic is needed.

src/server/analytic_services/prometheus_reporting.js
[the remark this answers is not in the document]
guymguym: Perhaps it is possible as you suggest, but why do you think it's easier to mock a module exports object which all the other code in the system refers to, rather than create a new unrelated instance that the other code cannot refer to. It's not critical, but I would encourage you to class a class to encapsulate state and functionality.

config.js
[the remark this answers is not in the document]
guymguym: OK but I'm not following the logic.
If this is an environmental setting, then we use env to set it.
If this is a developer configuration then we use config.js.
What is your case?

src/hosted_agents/hosted_agents_starter.js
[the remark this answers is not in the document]
nb-ohad: Done, waiting for all now

config.js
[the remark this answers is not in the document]
guymguym: OK so I think what you're suggesting is that this is a development config, and that for any case if we need to disable it we will use the CONFIG_JS_ env prefix to override from k8s.

So I think that can also stay as is. Resolving.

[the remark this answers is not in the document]
nb-ohad: Sorry, missed the `CONFIG_JS_` prefix part. I will change the env names

src/server/web_server.js
nb-ohad: Done, I am now waiting for all of them, just for the sake of consistency

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
nb-ohad: done

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
nb-ohad: done

src/hosted_agents/hosted_agents_starter.js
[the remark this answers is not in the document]
guymguym: La Insistance! :) Resolving.

src/server/analytic_services/prometheus_reporting.js
nb-ohad: After the discussion (slack) with I did the following changes. 
1. will retry a fixed number of times based on config/env, logging each failure, with a  delay based on config/env.
2. will exist the process after all retries 

Default values are:
- Infinity for retry_count.
- delay is 5 min

src/endpoint/endpoint.js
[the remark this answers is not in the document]
nb-ohad: I decided to wait for all in this case

config.js
```diff
@@ -394,9 +394,17 @@ config.MIN_MEMORY_FOR_UPGRADE = 1200 * 1024 * 1024;
 //////////////////////////////
 
 config.SEND_EVENTS_REMOTESYS = true;
-config.PROMETHEUS_ENABLED = true;
+config.CONFIG_JS_PROM_ENABLED = true;
```
guymguym, line 397: You don't need to set CONFIG_JS_ prefix, it will be stripped from env, right?

nb-ohad: done

guymguym (approved):
👍

nimrod-becker (approved):

Issue #6164, opened 2020-09-09
Get object in namespaced s3 compatible bucket only returns first 4kiB


lallinger-arbeit:
### Environment info
- NooBaa Version: master+74b79637d8a9a1dca2baf51598860579711a5c6b
- Platform: OKD 4.5

### Actual behavior
1. Performing a `GET` to any object in a namespaced bucket returns only the first 4kiB.

### Expected behavior
1. Performing a `GET` to any object in a namespaced bucket should return the full object

### Steps to reproduce
1. Create namespaced s3 compatible bucket
2. Upload an object
3. `GET` the object

### More information - Screenshots / Logs / Other output
Also the metadata of all objects is reported incorrectly, as `Content-Length` Noobaa always states `4096`.

As far as i can see [this commit](https://github.com/noobaa/noobaa-core/commit/abbafc463045736a4a2be973b1e15c282cb41d97) from @paulxuzhang  introduced the problem here:
https://github.com/noobaa/noobaa-core/blob/72a0fbc92ac4c29a98a8ab3eca3118a167860f5b/src/sdk/namespace_s3.js#L142-L144

Also i noticed that even when you send a headObject request to Noobaa this function (read_object_md) gets called and performs a getObject, instead of only calling headObject on the backend.

guymguym:
Might be related to this addition:
https://github.com/noobaa/noobaa-core/blob/bcc21725dd3ee45e73f05b87e7ac581d73f11b76/src/endpoint/s3/s3_utils.js#L279

We should try to add a unit test that fails on this issue.
@lallinger-arbeit Is it possible for you to take a stab?

guymguym:
@lallinger-arbeit A question - can you print the headers you get back on the GET response?

guymguym:
Actually wait, not sure the headers are needed.

paulxuzhang:
@guymguym Created PR https://github.com/noobaa/noobaa-core/pull/6167 for this issue. The root cause is that the changes supporting partNumber query parameter sets content-length to the one returned by hub, it ends up that content-length is set to Inline Range Length (4kb) for object with size > 4kb if partNumber query parameter is NOT set.

We should whether we have existing tests or create new for validating that.

guymguym:
Fixed by #6167

PR #6167, opened 2020-09-09
Fix content-length in get object response on namespace bucket

paulxuzhang:
### Explain the changes
1. Fix the issue that content-length is not set correctly in get object on namespace bucket

1. #6164 

### Testing Instructions:
1. create namespace bucket
2. upload object with size > 4k
3. get the object and check content-length is set to object size

### DIFF
--- a/src/test/unit_tests/test_s3_ops.js
+++ b/src/test/unit_tests/test_s3_ops.js
@@ -11,6 +11,7 @@ const http_utils = require('../../util/http_utils');
 const mocha = require('mocha');
 const assert = require('assert');
 const querystring = require('querystring');
+const P = require('../../util/promise');
 
 // If any of these variables are not defined,
 // use the noobaa endpoint to create buckets
@@ -328,6 +329,27 @@ mocha.describe('s3_ops', function() {
             assert.strictEqual(res.ContentLength, file_body.length);
         });
 
+        mocha.it('should get text-file with size > inline range', async function() {
+            const ORIG_INLINE_MAX_SIZE = config.INLINE_MAX_SIZE;
+            // Change the inline max size so the objects get cached.
+            config.INLINE_MAX_SIZE = 1;
+            const res = await s3.getObject({ Bucket: bucket_name, Key: text_file1 }).promise();
+            config.INLINE_MAX_SIZE = ORIG_INLINE_MAX_SIZE;
+            assert.strictEqual(res.Body.toString(), file_body);
+            assert.strictEqual(res.ContentType, 'text/plain');
+            assert.strictEqual(res.ContentLength, file_body.length);
+        });
+
+        mocha.it('should head text-file with size > inline range', async function() {
+            const ORIG_INLINE_MAX_SIZE = config.INLINE_MAX_SIZE;
+            // Change the inline max size so the objects get cached.
+            config.INLINE_MAX_SIZE = 1;
+            const res = await s3.headObject({ Bucket: bucket_name, Key: text_file1 }).promise();
+            config.INLINE_MAX_SIZE = ORIG_INLINE_MAX_SIZE;
+            assert.strictEqual(res.ContentType, 'text/plain');
+            assert.strictEqual(res.ContentLength, file_body.length);
+        });
+
         mocha.it('should get text-file sliced 1', async function() {
             const res = await s3.getObject({ Bucket: bucket_name, Key: text_file1, Range: 'bytes=0-5' }).promise();
             assert.strictEqual(res.Body.toString(), sliced_file_body);
@@ -464,7 +486,13 @@ mocha.describe('s3_ops', function() {
             await s3.getObject({ Bucket: bucket_name, Key: text_file2 }).promise();
             await s3.getObject({ Bucket: bucket_name, Key: text_file3 }).promise();
 
+            assert.strictEqual(res1.Contents[0].Key, text_file1);
+            assert.strictEqual(res1.Contents[0].Size, file_body.length);
+            assert.strictEqual(res1.Contents.length, 3);
+
+            if (!caching) return;
             const query_params = { get_from_cache: true };
+            await P.delay(300);
             const res2 = await s3.listObjects({ Bucket: bucket_name }).on('build', req => {
                 if (!caching) return;
                 const sep = req.httpRequest.search() ? '&' : '?';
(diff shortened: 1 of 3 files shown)


src/test/unit_tests/test_s3_ops.js
```diff
@@ -464,7 +476,13 @@ mocha.describe('s3_ops', function() {
             await s3.getObject({ Bucket: bucket_name, Key: text_file2 }).promise();
             await s3.getObject({ Bucket: bucket_name, Key: text_file3 }).promise();
 
+            assert.strictEqual(res1.Contents[0].Key, text_file1);
+            assert.strictEqual(res1.Contents[0].Size, file_body.length);
+            assert.strictEqual(res1.Contents.length, 3);
+
+            if (!caching) return;
             const query_params = { get_from_cache: true };
+            await P.delay(200);
```
liranmauda, line 485: as we are trying not to use bluebird, I think we can change that to:
```suggestion
            const delay = util.promisify(setTimeout);
            await delay(null, 200);
```
@guymguym what do you think?

src/test/unit_tests/test_s3_ops.js
```diff
@@ -11,6 +11,7 @@ const http_utils = require('../../util/http_utils');
 const mocha = require('mocha');
 const assert = require('assert');
 const querystring = require('querystring');
+const P = require('../../util/promise');
```
liranmauda, line 14: as we are trying not to use bluebird, I think we can change that to:
```suggestion
const util = require('util');
```
@guymguym what do you think?

romayalon (commented):
I tried your testing instruction but sadly it seems that the download is still failing. 
can you please verify it again?

paulxuzhang:
@romayalon I just tested again.  I have two objects, one called new.file is 71 bytes, the other called 32k.file is 32KB. 

I tried reading these two files from NB namespace bucket. The content-length headers are set correctly in the debug output. How did you tested it? Did you pick up my PR?

- s3 cp s3://cos-nb/32k.file --debug ../32k.file.download
```
2020-09-14 08:58:49,466 - ThreadPoolExecutor-0_0 - botocore.parsers - DEBUG - Response headers: {'x-amz-request-id': 'kf2lf604-5kbw9h-e70', 'x-amz-id-2': 'kf2lf604-5kbw9h-e70', 'access-control-allow-origin': '*', 'access-control-allow-credentials': 'true', 'access-control-allow-methods': 'GET,POST,PUT,DELETE,OPTIONS', 'access-control-allow-headers': 'Content-Type,Content-MD5,Authorization,X-Amz-User-Agent,X-Amz-Date,ETag,X-Amz-Content-Sha256', 'access-control-expose-headers': 'ETag,X-Amz-Version-Id', 'etag': '"bb7df04e1b0a2570657527a7e108ae23"', 'last-modified': 'Mon, 14 Sep 2020 13:58:33 GMT', 'content-type': 'binary/octet-stream', 'content-length': '32768', 'accept-ranges': 'bytes', 'x-amz-meta-noobaa-namespace-s3-bucket': 'test-nb', 'content-range': 'bytes 0-4095/32768', 'date': 'Mon, 14 Sep 2020 13:58:49 GMT', 'connection': 'keep-alive'}

```
- s3 cp s3://cos-nb/new.file --debug ../new.file.download
```
2020-09-14 09:03:27,099 - ThreadPoolExecutor-0_0 - botocore.parsers - DEBUG - Response headers: {'x-amz-request-id': 'kf2ll4d6-2m7ycx-tl5', 'x-amz-id-2': 'kf2ll4d6-2m7ycx-tl5', 'access-control-allow-origin': '*', 'access-control-allow-credentials': 'true', 'access-control-allow-methods': 'GET,POST,PUT,DELETE,OPTIONS', 'access-control-allow-headers': 'Content-Type,Content-MD5,Authorization,X-Amz-User-Agent,X-Amz-Date,ETag,X-Amz-Content-Sha256', 'access-control-expose-headers': 'ETag,X-Amz-Version-Id', 'etag': '"ff99f8c6eee15742a9c55c040c91b639"', 'last-modified': 'Mon, 14 Sep 2020 13:57:53 GMT', 'content-type': 'binary/octet-stream', 'content-length': '72', 'accept-ranges': 'bytes', 'x-amz-meta-noobaa-namespace-s3-bucket': 'test-nb', 'content-range': 'bytes 0-71/72', 'date': 'Mon, 14 Sep 2020 14:03:27 GMT', 'connection': 'keep-alive'}
```

romayalon:
@paulxuzhang 
yes, I picked up your code.
1. created namespace bucket 
2. uploaded a file size 31 KB using cyber duck
3. downloaded the file using cyber duck - and the download fails.

nimrod-becker:
@romayalon NS bucket was on top of S3 right ? single resource ?

romayalon:
@nimrod-becker 
yes, created namespace resource on S3 compatible and used it as a single resource.

romayalon:
@paulxuzhang 
found the problem,
cyber duck calls head before get and for some reason we are returning range 4096-4095 in our head reply, and because of that cyber duck calls get with this range.
when I ran script that does only getObject from the namespace bucket I got the full file.

paulxuzhang:
@romayalon Ok. I will fix head procedure as well. Checked head response in aws cli, the content-length is not right.

romayalon:
@paulxuzhang 
I re-checked the last commit using cyber duck and now it's working :)
thanks for your fix

src/test/unit_tests/test_s3_ops.js
guymguym: You're right @liranmauda, but I want to focus this review. Resolving and we'll handle P.delay across the code later

guymguym (changes requested):
@paulxuzhang Couple of questions:

1. Now namespace_s3 differs from namespace_nb (and other ns) in the fact that whenever partNumber or Range header are provided namespace_s3 will return object_md.content_length, and for namespace_nb ranged operations require further handling in s3_get_object and s3_head_object (they should actually return exactly the same headers for content-length/range...). Can we move this to the namespace code perhaps?

2. Notice that we when partNumber is set in namespace_s3.read_object_md, we do not set the Range header, but we also do not change the request to HEAD, so we end up calling GET for that part when the request from us was to HEAD... This should be fixed:
https://github.com/noobaa/noobaa-core/blob/bbd254e2b5d4d6ed6036a2aa6f5aefa551601a92/src/sdk/namespace_s3.js#L142-L144

guymguym (approved):

PR #6169, opened 2020-09-10
Fixing BZ 1875901 and removing unused npm deps

liranmauda:
### Explain the changes
Fixing BZ 1875901 :
- Bumped google-cloud/storage to 5.3.0
- Bumped googleapis to 59.0.0

Removing:

- [x]  dclassify
- [x] event-stream
- [x] ini
- [x] mongodb-uri
- [x] multer
- [x] why-is-node-running


package.json
```diff
@@ -74,32 +74,26 @@
...
     "http-proxy-agent": "4.0.1",
     "https-proxy-agent": "5.0.0",
-    "ini": "1.3.5",
     "ip": "1.1.5",
     "jsonwebtoken": "8.5.1",
     "linux-blockutils": "0.1.0",
     "lodash": "4.17.19",
     "mime": "2.4.4",
     "minimist": "1.2.5",
-    "mkdirp": "0.5.1",
```
nimrod-becker, line 95: still be required and used in create_path in fsutils

liranmauda: yes, I reverted the change I made and missed adding it back to the package.json
We will be able to remove it from fs_utils by replacing it with fs.mkdir(dir, { recursive: true }, callback)
We can then remove the mode as the default mode is 0o777

nimrod-becker (approved):

PR #6171, opened 2020-09-13
Adding secrets encryption capability to noobaa

romayalon:

### Explain the changes
1. Added master key Manager that will handle all the encryption/decryption ops.
2. Master key manager will also handle the load of root key secret that was created by the noobaa-operator.
3. Created a master key for system/bucket/account. The system master key is encrypted by root key, the bucket and account master keys are encrypted by the system master key.
4. every chunk that is related to bucket will hold the master key of the bucket and will encrypt the chunks cipher key.
5. every account's access key/cloud credentials which are encrypted by their account master key id.
6. Im every load of system store the accounts access keys/ cloud credentials are decrypted and saved to system store decrypted.
7. Added simple unit test.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/sdk/map_api_types.js
+++ b/src/sdk/map_api_types.js
@@ -51,6 +51,7 @@ class ChunkAPI {
     get frag_size() { return this.chunk_info.frag_size; }
     get digest_b64() { return this.chunk_info.digest_b64; }
     get cipher_key_b64() { return this.chunk_info.cipher_key_b64; }
+    get master_key_id() { return this.chunk_info.master_key_id; }
     get cipher_iv_b64() { return this.chunk_info.cipher_iv_b64; }
     get cipher_auth_tag_b64() { return this.chunk_info.cipher_auth_tag_b64; }
     get chunk_coder_config() { return this.chunk_info.chunk_coder_config; }
@@ -132,6 +133,7 @@ class ChunkAPI {
     to_api() {
         return {
             _id: this.chunk_info._id,
+            master_key_id: this.chunk_info.master_key_id,
             bucket_id: this.chunk_info.bucket_id,
             tier_id: this.chunk_info.tier_id,
             chunk_coder_config: this.chunk_info.chunk_coder_config,
@@ -157,6 +159,7 @@ class ChunkAPI {
     to_db() {
         return {
             _id: this._id,
+            master_key_id: this.master_key_id,
             bucket: this.bucket_id,
             tier: this.tier_id,
             size: this.size,
@@ -164,7 +167,7 @@ class ChunkAPI {
             frag_size: this.frag_size,
             dedup_key: from_b64(this.chunk_info.digest_b64),
             digest: from_b64(this.chunk_info.digest_b64),
-            cipher_key: from_b64(this.chunk_info.cipher_key_b64),
+            cipher_key: this._encrypt_cipher_key(this.chunk_info.cipher_key_b64, this.chunk_info.master_key_id),
             cipher_iv: from_b64(this.chunk_info.cipher_iv_b64),
             cipher_auth_tag: from_b64(this.chunk_info.cipher_auth_tag_b64),
             chunk_config: this.chunk_config._id,
@@ -174,6 +177,10 @@ class ChunkAPI {
         };
     }
 
+    _encrypt_cipher_key(cipher_key, master_key_id) {
+        if (!master_key_id) return from_b64(cipher_key);
+        return this.system_store.master_key_manager.encrypt_buffer_with_master_key_id(from_b64(cipher_key), master_key_id);
+    }
 }
 
 /**
--- a/src/sdk/nb.d.ts
+++ b/src/sdk/nb.d.ts
@@ -39,6 +39,7 @@ interface System extends Base {
     tiers_by_name: { [name: string]: Tier };
     pools_by_name: { [name: string]: Pool };
     chunk_configs_by_id: { [id: string]: ChunkConfig };
+    master_key_id: ID;
 }
 
 interface Account extends Base {
@@ -56,6 +57,7 @@ interface Account extends Base {
         access_key: SensitiveString;
         secret_key: SensitiveString;
     }[];
+    master_key_id: ID;
 }
 
 interface NodeAPI extends Base {
@@ -177,6 +179,7 @@ interface Bucket extends Base {
     };
     lifecycle_configuration_rules?: Object;
     lambda_triggers?: Object;
+    master_key_id: ID;
 }
 
 interface CacheConfig {
@@ -237,6 +240,7 @@ interface Chunk {
     readonly cipher_iv_b64: string;
     readonly cipher_auth_tag_b64: string;
     readonly chunk_coder_config: ChunkCoderConfig;
+    master_key_id?: ID;
 
     dup_chunk_id?: ID;
     had_errors?: boolean;
@@ -428,6 +432,7 @@ interface ChunkInfo {
     is_accessible?: boolean;
     is_building_blocks?: boolean;
     is_building_frags?: boolean;
+    master_key_id?: ID;
 
     // Properties not in the API but used in memory
     data?: Buffer;
@@ -523,6 +528,7 @@ interface ChunkSchemaDB {
     frags: FragSchemaDB[];
     parts?: PartSchemaDB[]; // see MDStore.load_parts_objects_for_chunks()
     objects?: any[]; // see MDStore.load_parts_objects_for_chunks()
+    master_key_id?: ID;
 }
 
 interface FragSchemaDB {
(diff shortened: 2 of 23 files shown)


src/server/system_services/bucket_server.js
```diff
@@ -129,13 +131,19 @@ async function create_bucket(req) {
         changes.insert.tiers = [tier];
     }
 
+    const bucket_m_key = master_key_manager.get_instance().new_master_key({
+        description: 'Bucket master key',
```
jeniawhite, line 135: `description` should somehow explain the key, we shouldn't just have the same description to everyone.
Please check in all descriptions (system, bucket, account etc...) and maybe add some uniqueness to them.

src/server/system_services/master_key_manager.js
```diff
@@ -0,0 +1,246 @@
...
+
+class MasterKeysManager {
+    constructor() {
+        this.is_initialized = false;
+        this.resolved_master_keys_by_id = {};
+        this.master_keys_by_id = {};
+        this.secret_keys_cache = new LRUCache({
+            name: 'SecretKeysCache',
+            expiry_ms: Infinity,
+            max_usage: 5000,
```
jeniawhite, line 23: In theory, I wouldn't want to limit, we can use Infinity.
Shouldn't have plenty of keys anyway.

src/server/system_services/master_key_manager.js
```diff
@@ -0,0 +1,246 @@
...
+     * @param {Object} options 
+     * @returns {Object} 
+     */
+    new_master_key(options) {
+        const { description, master_key_id, cipher_type } = options;
+        const _id = db_client.new_object_id();
+        const m_key = _.omitBy({
+            _id,
+            description,
+            master_key_id: master_key_id && master_key_id !== ROOT_KEY ? db_client.parse_object_id(master_key_id) : undefined,
```
jeniawhite, line 89: This causes the system to not have master_key_id which is weird since it's master key is the root key.

src/server/system_services/master_key_manager.js
```diff
@@ -0,0 +1,246 @@
...
+        return this.resolved_master_keys_by_id[_id.toString()] || this._resolve_master_key(mkey);
+    }
+
+    /**
+     * 
+     * @param {Object} m_key 
+     * @returns {Object}  
+     */
+    _resolve_master_key(m_key) {
+        const m_of_mkey_id = (m_key.master_key_id && m_key.master_key_id._id) || m_key.master_key_id;
```
jeniawhite, line 121: You should consider handling this in a more consistent way, you shouldn't guess if objects are populated or not.

romayalon: deleted the last part of the phrase, it was in a case where the master_key of the master key id is the root secret

src/server/system_services/bucket_server.js
romayalon: added ids for the uniqueness

src/server/system_services/system_server.js
```diff
@@ -234,7 +242,15 @@ function new_system_changes(name, owner_account_id) {
...
         }]
     );
-    const bucket = bucket_server.new_bucket_defaults(default_bucket_name, system._id, policy._id, owner_account_id);
+
+    const first_bucket_m_key = master_key_manager.get_instance().new_master_key({
+        description: 'First bucket master key',
+        master_key_id: m_key._id,
+        cipher_type: m_key.cipher_type
+    });
+    const bucket = bucket_server.new_bucket_defaults(default_bucket_name, system._id, policy._id, owner_account_id,
```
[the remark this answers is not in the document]
romayalon: I moved the set of the master_key_id outside of this function so it's irrelevant any more.

src/server/system_services/master_key_manager.js
```diff
@@ -0,0 +1,246 @@
...
+        if (!_id) return value;
+        const m_key = this.get_master_key_by_id(_id);
+        if (!m_key) throw new Error('NO_SUCH_KEY');
+
+        const cipher = crypto.createCipheriv(m_key.cipher_type, m_key.cipher_key, m_key.cipher_iv);
+        let updated_value = cipher.update(Buffer.from(value.unwrap()));
+        if (m_key.cipher_type !== 'aes-256-gcm') updated_value = Buffer.concat([updated_value, cipher.final()]);
+
+        const ciphered_value = new SensitiveString(updated_value.toString('base64'));
+        if (options.cache) this.secret_keys_cache.put_in_cache(ciphered_value, value);
```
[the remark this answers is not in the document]
romayalon: deleted

src/server/system_services/pool_server.js
```diff
@@ -290,6 +294,9 @@ function create_namespace_resource(req) {
 async function create_cloud_pool(req) {
     var name = req.rpc_params.name;
     var connection = cloud_utils.find_cloud_connection(req.account, req.rpc_params.connection);
+    connection.secret_key = master_key_manager.get_instance().encrypt_sensitive_string_with_master_key_id(
```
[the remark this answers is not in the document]
romayalon: deleted

src/server/system_services/pool_server.js
```diff
@@ -25,6 +25,7 @@ const IoStatsStore = require('../analytic_services/io_stats_store').IoStatsStore
 const pool_ctrls = require('./pool_controllers');
 const func_store = require('../func_services/func_store');
 const { KubeStore } = require('../kube-store.js');
+const master_key_manager = require('../system_services/master_key_manager');
```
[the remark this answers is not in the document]
romayalon: agreed, changed in all code.

src/sdk/nb.d.ts
```diff
@@ -237,6 +240,7 @@ interface Chunk {
     readonly cipher_iv_b64: string;
     readonly cipher_auth_tag_b64: string;
     readonly chunk_coder_config: ChunkCoderConfig;
+    master_key_id: ID;
```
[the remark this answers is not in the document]
romayalon: fixed.

src/server/system_services/schemas/master_key_schema.js
```diff
@@ -0,0 +1,34 @@
...
+    required: [
+        '_id',
+    ],
+    properties: {
+        _id: { objectid: true },
+        description: { type: 'string' },
+        // If missing - encrypted with root key
+        // 1. ENV - Mounted key from Kubernetes secret
+        // 2. HSM configuration TBD - Get key from Vault
+        master_key_id: {
```
jeniawhite, line 16: @guymguym 
Can you please take a look at the oneOf, not sure if we shall push it like that?
Maybe use a custom ObjectId for the master_key that will be constant instead of the string 'noobaa-root-master-key'

src/sdk/map_api_types.js
```diff
@@ -174,6 +177,10 @@ class ChunkAPI {
         };
     }
 
+    _encrypt_cipher_key(cipher_key, master_key_id) {
+        if (!master_key_id) return cipher_key;
+        return this.system_store.master_key_manager.encrypt_buffer_with_master_key_id(from_b64(cipher_key), master_key_id).toString('base64');
```
jeniawhite, line 182: Why do we need to do 2 conversions.
toString('base64') and then fromBase64().

src/server/system_services/schemas/master_key_schema.js
guymguym: The reason I'm not a fan of oneOf is that it defines the property with a confusing type, and unless we really feel like this is the only option we can do it (like in our bigint type for example) but if it's easy to separate to different properties that each has a concise type I think it should be preferred.

romayalon: it's object_id or the root_key name - so it's very easy to separate between those two, I agree that it's better than hard coded objectId

src/sdk/map_api_types.js
romayalon: fixed, it was a minor bug when master_key_id is undefined 👍

src/server/system_services/master_key_manager.js
romayalon: just updating, the part after the || operator is necessary also for times when m_key.master_key_id._id doesn't exist which happened before loading from system store at the beginning.

jeniawhite (approved):

src/test/unit_tests/test_encryption.js
```diff
@@ -0,0 +1,81 @@
+/* Copyright (C) 2016 NooBaa */
```
jeniawhite, line 1: I think that the addition of regenerating access keys for accounts is also a good test case.
Edit cloud connections and see the regeneration.
Check the hierarchy on system_store load (that all master_keys are resolved in the cache of master_keys_manager).

jeniawhite (approved):

src/test/unit_tests/test_encryption.js
romayalon: Addad many tests, please review it again 👍

nimrod-becker (approved):

PR #6173, opened 2020-09-17
Fix rpc_benchmark and add fcall mode

nb-ohad:
### Explain the changes
1. Fix rpc_banchmark (was broken since https://github.com/noobaa/noobaa-core/pull/6022 was merged)
2. Refactor start as an async/await function
3. Add a `--fcall` flag to test RPC over `fcalll` connection (process acts as both server and client)

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/rpc/rpc_benchmark.js
+++ b/src/rpc/rpc_benchmark.js
@@ -14,22 +14,28 @@ const dbg = require('../util/debug_module')(__filename);
 const RPC = require('./rpc');
 const RpcSchema = require('./rpc_schema');
 const ssl_utils = require('../util/ssl_utils');
-const { RPC_BUFFERS } = require('.');
+const { RPC_BUFFERS } = require('./rpc_request');
 
 const MB = 1024 * 1024;
 
 // test arguments
-// client/server mode
+// client/server/fcall mode
 argv.client = argv.client || false;
 argv.server = argv.server || false;
+argv.fcall = argv.fcall || false;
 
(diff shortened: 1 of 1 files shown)


src/rpc/rpc_benchmark.js
```diff
@@ -110,14 +117,15 @@ schema.compile();
 
 // create rpc
 const rpc = new RPC({
-    schema: schema,
+    client_factory: api.client_factory_from_schema(schema),
```
guymguym, line 120: Hmm, I want to make sure `rpc` module is usable even without `api` module.
Do you think this can be done?

guymguym (approved):
See one comment, but overall this is good. Any results from running it?

src/rpc/rpc_benchmark.js
nb-ohad: Ok we have it now in a sepetate PR: https://github.com/noobaa/noobaa-core/pull/6178
We can wait to merge this one untill the other one is merged

guymguym (changes requested):
Please rebase master

src/rpc/rpc_benchmark.js
nb-ohad: Rebased against `master` and removed dependency with the `api` module

guymguym (approved):
Looks great. Do you have the results of running this between the new fcall that clones messages and the previous which was using JSON.stringify/parse?

nb-ohad:
No, as this change was merged to master after the fcall changes

PR #6175, opened 2020-09-18
NS-CACHE: Support non-latency metrics

paulxuzhang:

### Explain the changes
1.  Support the following metrics for namespace caching
- cache_read_bytes
- cache_write_bytes
- cache_range_read_miss_count
- cache_object_read_miss_count
- cache_object_read_count
- cache_range_read_count
- hub_read_bytes
- hub_write_bytes

1.  Part of the namespace caching feature

### Testing Instructions:
1.  Create namespace caching bucket
2. Perform operations, upload object, read object (cached and non-cached), range read object (cached and non-cached)
3. Verify the metrics through the endpoint metric URL, e.g. curl http://127.0.0.1:7004

### DIFF
--- a/src/sdk/endpoint_stats_collector.js
+++ b/src/sdk/endpoint_stats_collector.js
@@ -6,7 +6,7 @@ const mime = require('mime');
 const promise_utils = require('../util/promise_utils');
 const _ = require('lodash');
 const dbg = require('../util/debug_module')(__filename);
-
+const prom_report = require('../server/analytic_services/prometheus_reporting');
 
 
 // 30 seconds delay between reports
@@ -18,6 +18,7 @@ class EndpointStatsCollector {
     constructor(rpc_client) {
         this.rpc_client = rpc_client;
         this.reset_all_stats();
+        this.prom_metrics_report = prom_report.get_endpoint_report();
     }
 
     static instance(rpc_client) {
@@ -59,7 +60,7 @@ class EndpointStatsCollector {
         };
     }
 
-    update_namespace_read_stats({ namespace_resource_id, size = 0, count = 0, is_err }) {
+    update_namespace_read_stats({ namespace_resource_id, bucket_name, size = 0, count = 0, is_err }) {
         this.namespace_stats[namespace_resource_id] = this.namespace_stats[namespace_resource_id] || this._new_namespace_stats();
         const io_stats = this.namespace_stats[namespace_resource_id];
         if (is_err) {
@@ -69,10 +70,13 @@ class EndpointStatsCollector {
             io_stats.read_count += count;
             io_stats.read_bytes += size;
         }
+        if (bucket_name) {
+            this.prom_metrics_report.inc('hub_read_bytes', { bucket_name }, size);
+        }
         this._trigger_send_stats();
     }
 
-    update_namespace_write_stats({ namespace_resource_id, size = 0, count = 0, is_err }) {
+    update_namespace_write_stats({ namespace_resource_id, bucket_name, size = 0, count = 0, is_err }) {
         this.namespace_stats[namespace_resource_id] = this.namespace_stats[namespace_resource_id] || this._new_namespace_stats();
         const io_stats = this.namespace_stats[namespace_resource_id];
         if (is_err) {
@@ -81,6 +85,9 @@ class EndpointStatsCollector {
         } else {
             io_stats.write_count += count;
             io_stats.write_bytes += size;
+        }
+        if (bucket_name) {
+            this.prom_metrics_report.inc('hub_write_bytes', { bucket_name }, size);
         }
         this._trigger_send_stats();
     }
@@ -106,6 +113,23 @@ class EndpointStatsCollector {
     update_bucket_write_counters({ bucket_name, key, content_type, }) {
         this._update_bucket_counter({ bucket_name, key, content_type, counter_key: 'write_count' });
         this._trigger_send_stats();
+    }
+
+    update_namespace_cache_stats({ bucket_name, read_bytes, write_bytes, read_count = 0, miss_count = 0, range_op = false }) {
+        if (read_bytes) {
+            this.prom_metrics_report.inc('cache_read_bytes', { bucket_name }, read_bytes);
+        }
+        if (read_count) {
+            this.prom_metrics_report.inc(range_op ? 'cache_range_read_count' : 'cache_object_read_count',
+                { bucket_name }, read_count);
+        }
+        if (miss_count) {
+            this.prom_metrics_report.inc(range_op ? 'cache_range_read_miss_count' : 'cache_object_read_miss_count',
+                { bucket_name }, miss_count);
+        }
+        if (write_bytes) {
+            this.prom_metrics_report.inc('cache_write_bytes', { bucket_name }, write_bytes);
+        }
     }
 
     get_all_stats() {
--- a/src/sdk/namespace_cache.js
+++ b/src/sdk/namespace_cache.js
@@ -14,6 +14,7 @@ const Semaphore = require('../util/semaphore');
 const S3Error = require('../endpoint/s3/s3_errors').S3Error;
 const s3_utils = require('../endpoint/s3/s3_utils');
 const config = require('../../config');
+const stats_collector = require('./endpoint_stats_collector');
 
 const _global_cache_uploader = new Semaphore(cache_config.UPLOAD_SEMAPHORE_CAP, {
     timeout: cache_config.UPLOAD_SEMAPHORE_TIMEOUT,
@@ -27,6 +28,7 @@ class NamespaceCache {
         this.namespace_nb = namespace_nb;
         this.active_triggers = active_triggers;
         this.caching = caching;
+        this.stats_collector = stats_collector.instance(namespace_hub.rpc_client);
     }
 
     get_write_resource() {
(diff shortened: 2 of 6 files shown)


src/server/analytic_services/prometheus_reports/base_prometheus_report.js
```diff
@@ -32,6 +32,15 @@ class BasePrometheusReport {
...
     }
+
+    get_metric(name) {
+        const metric_name = this.get_prefixed_name(name);
+        return this.registry.getSingleMetric(metric_name);
+    }
+
+    reset_metrics() {
+        return this.registry.resetMetrics();
+    }
```
nb-ohad, line 43: From the review in the last communitee meeting, I understand that these functions are used only accessed from tests.
In that case, I think it is preferable to define them as util function in the tests framework itself, so they are not part of the base report interface.

For exmple:
```
function reset_metrics(report) {
    report.registry.resetMetrics();   
}
```

In this case, the code base uses the reports as a blackbox while the test code uses the reports as a white box.

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
```diff
@@ -2,38 +2,114 @@
...
+            if (type === 'Counter') {
+                this[`set_${name}`] = (...args) => {
+                    if (!this._metrics) return;
+                    this._metrics[name].inc(...args);
+                };
+            } else if (type === 'Histogram') {
+                this[`set_${name}`] = (...args) => {
+                    if (!this._metrics) return;
+                    this._metrics[name].observe(...args);
+                };
```
nb-ohad, line 104: Creating a function dynamically on a class instance will introduce some problems:
1. The js linter will not be aware of these functions.
2. Reading the class does not expose the class API, the API is hidden inside the constructor implantation
3. We may find in the future that we have some "special cases" that does not fall into these two categories, for example, a new type of metric or even worse, we would like to treat the setting of two or more metrics together as one call. Which leaves us with 2 options, either we make the generating code much more complex or we start to add some methods to the class, but in that case, we just split the definition of the API of the class into two separate locations and the reader will just get even more confused.

This is exactly the case we currently have with the `noobaa-core` report, which is bad, and I think we should not repeat the same mistake twice.

My approach would be as follows (the example is for `cache_range_read_count` but should be applied to all metrics):
In the constructor:
```
this._metrics['cache_range_read_count'] = new client.Counter({
  name: this.get_prefixed_name('cache_range_read_count'),
  registers: [this.registry],
  help: `Counter on range reads in namespace cache bucket`,
  labelNames: ['namespace_resource_id'] 
})
```

And one or more class methods that exposes the metric:
```
inc_cache_range_read_count(value) {
    if (!this._metrics) return;
    this._metrics['inc_cache_range_read_count'].inc(value)
}
```

I know it is more verbos, but in my eye this is exactly what er need it to be without adding more unnececary complexity

src/sdk/namespace_cache.js
```diff
@@ -257,7 +263,11 @@ class NamespaceCache {
                             xattr: object_info_hub.xattr,
                             last_modified_time: (new Date(object_info_hub.create_time)).getTime(),
                         };
-                        return this.namespace_nb.upload_object(upload_params, object_sdk);
+                        const upload_res = await this.namespace_nb.upload_object(upload_params, object_sdk);
+                        this._update_metric('set_cache_write_bytes', object_info_hub.first_range_data.length);
+                        this._update_metric('set_cache_object_read_miss_count', 1);
+                        this._update_metric('set_cache_object_read_count', 1);
+                        return upload_res;
```
nb-ohad, line 270: Why do we need the update_metric indirection? why not just:
```
this. stats_collector. set_cache_write_bytes(`object_info_hub.first_range_data.length`);
this.stats_collector.set_cache_object_read_miss_count(1);
this.stats_collector.set_cache_object_read_count(1);
```
It is much more verbose, less prone to the string changing as the js linter should know what is the interface of the report class (see my comments on the `endpoint_report` changes)

Please note that this comment should also apply to all the other this._update_metric class in this file

nb-ohad (changes requested):

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
paulxuzhang: ok

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
paulxuzhang: I agree with 1 and 2. Regarding 3, I don't think supporting different types of metrics will cause generated functions to be complex. They all fall into similar prototype, set(labels, value), inc(labels, value), observe(labels, value) etc. 

Regarding your proposal manually adding individual class method for each metric, the class methods would have similar code. It would be error prone in terms of coding, if programmer tried to copy and paste to create class method for new metric, and forgot to change the name of the metric in (this._metrics['wrong_name']), it would be annoying issue to figure it out. If there are too many metrics, it would just too boring to write. It is why we tend to create generic function to make code concise and easy to maintain.

If you don't think creating class methods during run-time is good idea, would you be ok with different way of creating generic function?, such as 
```
function observe(metric_name, labels, value) {
    if (!this.metrics) return;
    const metric = this._metrics['inc_cache_range_read_count'];
    if (!metric) throw new Error(...'');
    if (metric.type === 'Counter') metric.inc(labels, value);
    else if if (metric.type === 'Histogram') metric.observe(labels, value);
    else throw new Error('...');
}
```

nb-ohad: To my eyes, this seems much easier to understand than the original version as it defines a clean interface to the class (althougth a generic one). 

But let's consider the points you have made:
1. Yes the code will be similar, but it will be small and readable. The linter and autocomplete features of our IDEs will help us when using the class as the interface is fully defined. So trying to update a non existant metric will be catched immediately. 

2. I agree that while coding you might make errors while writing new accessor methods, but consider that these methods are only written once, while errors in passing a wrong name for a metric in call sites has greater possiblity of occurence. 
These errors can be cached only in runtime and only when these code paths are active.
Let's say tomorrow you would like to rename a metric and it's accessor, in my proposal, as soon as you change the method name the linter will complain on all possible call sites. In your proposal we will have to find any potantial bug in runtime. Also, the number of metrics is finite which means that this effort should only happen once.

3. Sometimes code is boarding, this is the nature of the beast.

4. You skipped the biggest point on my list: It might happen that you will find that some code paths need to update one metric based on another or even conditionally. In that case, you would like to capture this logic inside the report class in order to avoid duplication in all call sites (you can see an example of this happening to us with the core's report). 
The cleanest way to do that is to create a non-naive setter that capture that logic, trying to describe this using a data model is hard and over complicating the entire process. In addition to that, you will not want to give access to the actual metrics behind the scenes so you will now have logic that skips the creation of auto setters for these metrics which again make it even more complex.

Saying all of the above I still think a verbose interface is the cleanest, with minimal complexity, more readable, easier to change,  and will do the job just fine.

If you still want to do something else, I will suggest not having one `observe` methods but one for each metric kind: 
` inc(name, labels, value)`
`set(name, value)`
`observe(labels, value)`

At the minimum, this solution will allow the reader of the call site to understand the updated pattern.

paulxuzhang: @nb-ohad 
Regading 2, if calling function passes wrong name, it would be easy to catch since exception on unknown metric would be thrown. Let's say developer creates new metric function by copying and pasting existing one, and he/she only changes the function name and forgets to change the metric name inside the function, you will still have problem in runtime, and it is annoying to find the cause, since you would have two metrics having wrong value. In short, I see both methods have pros and cons. Probably not all will agree on same one.

Regarding 4, in the case that logic of updating metric is complex, we will create customize function on top of the basic ones (e.g, inc, set, observe). I am not against creating such function. It is what we have to do anyway.

nb-ohad: @paulxuzhang I get your point, and I understand the pros and cons. 
And as I said before, it is just my opinion, coming from my past experience.

So let's agree to do your suggestion, the one from the first comment, with or without the separation to `inc/set/observe` to your choosing. sounds ok?

src/sdk/namespace_cache.js
nb-ohad: @paulxuzhang are you ok with this request?

paulxuzhang: Pushed changes.

nb-ohad (approved):

src/sdk/endpoint_stats_collector.js
```diff
@@ -108,6 +111,23 @@ class EndpointStatsCollector {
         this._trigger_send_stats();
     }
 
+    update_namespace_cache_stats({ namespace_resource_id, read_bytes, write_bytes, read_count = 0, miss_count = 0, range_op = false }) {
```
guymguym, line 114: Use bucket_name:
https://github.com/noobaa/noobaa-core/blob/a8b8f835b964d0c44e87615e7e7de5d56fedb869/src/sdk/object_sdk.js#L395-L396

paulxuzhang: @guymguym  Done.

guymguym (approved):

PR #6178, opened 2020-09-22
Move RPC client creation to the RPC schema class

nb-ohad:
### Explain the changes
1.  Move `client_factory_from_schema` logic from `api.js` to `rpc_schema.js` and produce a client factory upon schema compilation. 
2. Expose client creation from schema objects via the new `new_client(rpc, options)` method
4. Mark complied API schemas, so we could assert changes made to the schema after compilation.  

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/api.js
+++ b/src/api/api.js
@@ -2,7 +2,7 @@
 'use strict';
 
 const url = require('url');
-const { RPC, RPC_BUFFERS, RpcSchema } = require('../rpc');
+const { RPC, RpcSchema } = require('../rpc');
 const { get_base_address, get_default_ports } = require('../util/addr_utils');
 
 // registring all api's on the same RpcSchema object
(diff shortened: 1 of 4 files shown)


src/rpc/rpc_client_utils.js
```diff
@@ -0,0 +1,104 @@
...
+    return {};
+}
+
+function client_factory_from_schema(schema) {
+    if (!schema.compiled) {
+        throw new Error('Invalid schema: schema is not compiled');
+    }
+
+     // If a factory can be found in the cache we provide it from there.
+    const client_factory = schema_client_cache.get(schema);
```
guymguym, line 18: Is client_factory_from_schema getting called more than once per schema?
How about moving to RpcSchema to "compile a client"?

nb-ohad: @guymguym I started by having this code inside the RpcSchema but it seems like the wrong place to place it as It has nothing to do with the scehma itself so I added a seperate utils file.  
But as you observed there is nothing holding us from putting it in  RpcSchema.

Please note that in that implementation the schema will hold the factory, where here we just weakly key the factory with the schema as key

guymguym: Why do you say that it has nothing to do with the schema itself? 
This entire client_proto is being built from the schema definitions.
So yeah it makes a lot of sense to me as a functionality there.

nb-ohad: @guymguym I changed the code to provide clients directly from the schema object via a `new_client(rpc, options)` method. This makes the client factory creation an optimization/implementation detail of the RPCSchema class and simplify even more the code inside the RPC's new_client code.

I hope you will find this change to your liking

guymguym (approved):

PR #6179, opened 2020-09-22
Throw error for insufficient pv size in noobaa-core

romayalon:

### Explain the changes
1. Throw an Error from create_hosts_pool in noobaa core when insufficient PV size.
2. Follow up to https://github.com/noobaa/noobaa-operator/pull/420

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/pool_server.js
+++ b/src/server/system_services/pool_server.js
@@ -123,8 +123,14 @@ async function create_hosts_pool(req) {
     }
     const { system, rpc_params, account } = req;
     const pool = new_pool_defaults(rpc_params.name, system._id, 'HOSTS', 'BLOCK_STORE_FS');
+    const MIN_PV_SIZE_GB = 16;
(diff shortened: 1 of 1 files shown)


nimrod-becker:
@guymguym  ^^

romayalon:
@nimrod-becker @guymguym is BAD_REQUEST the right error code? any better suggestions?

guymguym (approved):
@romayalon I think BAD_REQUEST is good ... we can separate to finer code when needed, but for now it's good because it signals back to the operator that this is a persistent type of error.

guymguym:
Perhaps need to fix tests for this to work?

romayalon:
> Perhaps need to fix tests for this to work?

@guymguym no, I just had to add check for ```rpc_params.host_config``` is not undefined, now it passes the tests.

guymguym (approved):

PR #6180, opened 2020-09-23
Adding back es6-promise to package-lock.json

liranmauda:
### Explain the changes
- Adding back es6-promise to package-lock.json


guymguym (commented):
What is npm rebuild and why is it needed?

liranmauda:
Hi @guymguym 
npm rebuild recompile the modules. 
The reasons for adding that are:
1. Making sure we are getting the modules for the correct arch 
2. The downstream uses this (due to reason 1) and we don't want the diversion.

liranmauda:
Rethinking of it, it will probably not rebuild anything as we don't have any modules in that stage... 
removing this part.

nimrod-becker (approved):

PR #6181, opened 2020-09-23
Add hub (CSAFE-72420) and cache (CSAFE-72418) bucket read/write laten…

mcrepeau:
…cy metric

Signed-off-by: Matthieu Crépeau <matthieu.crepeau@ibm.com>

### Explain the changes
1. Support the following metrics for namespace caching:
- hub read latency
- hub write latency
- cache read latency
- cache write latency

1. Part of the namespace caching feature

### Testing Instructions:
1. Create namespace caching bucket
2. Perform operations, upload object, read object (cached and non-cached), range read object (cached and non-cached)
3. Verify the metrics through the endpoint metric URL, e.g. curl http://127.0.0.1:7004


src/sdk/namespace_cache.js
```diff
@@ -119,6 +120,10 @@ class NamespaceCache {
         return params.object_md.size <= cache_config.DEFAULT_BLOCK_SIZE;
     }
 
+    _update_metric(update_metric_fn_name, value) {
+        this.stats_collector[update_metric_fn_name]({ namespace_resource_id: this.namespace_hub.namespace_resource_id }, value);
+    }
```
nb-ohad, line 125: This will not work, as the `endpoint_report` / `stats_collector` does not expose individual metric setters.
You can see these changes here: https://github.com/noobaa/noobaa-core/pull/6175

src/sdk/namespace_cache.js
```diff
@@ -259,7 +264,9 @@ class NamespaceCache {
                             xattr: object_info_hub.xattr,
                             last_modified_time: (new Date(object_info_hub.create_time)).getTime(),
                         };
-                        const upload_res = await this.namespace_nb.upload_object(upload_params, object_sdk);
+
+                    var start = process.hrtime();    
+                    const upload_res = await this.namespace_nb.upload_object(upload_params, object_sdk).then(() => this._update_metric('set_hub_read_latency', (process.hrtime()[1] - start[1])/1000000));
```
nb-ohad, line 269: A couple of issues:
1. this code lives in an async function so there are no reasons the use `.then()`, using `then()` will also change the return value that is assigned to the `upload_res` variable
2. `process.hrtime` can take a param of prev invocation of `process.hrtime` and will provide the delta itself.

The code can be written as follows:
```
const start = process.hrtime();    
const upload_res = await this.namespace_nb.upload_object(upload_params, object_sdk);
this._update_metric('set_hub_read_latency', process.hrtime(start)/1000000));
```

Please note that this comment should also apply to all changes made to the file below

nb-ohad (changes requested):

src/sdk/namespace_cache.js
```diff
@@ -4,6 +4,7 @@
 const _ = require('lodash');
 const stream = require('stream');
 const assert = require('assert');
+const process = require('process');
```
liranmauda, line 7: If I recall correctly, we don't need to require process as it is a native global object

src/sdk/namespace_cache.js
mcrepeau: Ah my bad, I based my changed on @paulxuzhang 's first version

nb-ohad (approved):

nb-ohad:
@mcrepeau I approved the PR but I still have one observation:
Almost every call to `update_cache_latency_stats` is followed with a call to `update_cache_stats`, both functions are keying the data based on `bucket_name` while all other params are optional. 
Can we simplify the interface by merging both functions into a `update_cache_stats`?

src/sdk/namespace_s3.js
```diff
@@ -184,7 +184,7 @@ class NamespaceS3 {
                     req.removeListener('httpError', AWS.EventListeners.Core.HTTP_ERROR);
                     let count = 1;
                     const count_stream = stream_utils.get_tap_stream(data => {
-                        stats_collector.instance(this.rpc_client).update_namespace_read_stats({
+                        stats_collector.instance(this.rpc_client).update_hub_read_stats({
```
guymguym, line 187: I think the previous name made more sense. 

Also you seem to change only namespace_s3 but not namespace_blob...

But first I'm thinking - can we separate the hub metric update and call it from namespace_cache?

guymguym (changes requested):

src/sdk/namespace_s3.js
mcrepeau: I just wanted to differentiate more clearly between hub-related stats and cache-related stats. Is there a name that you think would make sense and would still have that differentiation?

Re the namespace_blob change, I forgot to change it

What do you mean about separating the hub metric update?

src/sdk/namespace_cache.js
```diff
@@ -353,8 +363,13 @@ class NamespaceCache {
         };
 
         try {
+            const start = process.hrtime();
             const cache_read_stream = await this.namespace_nb.read_object_stream(params, object_sdk);
-            this.stats_collector.update_namespace_cache_stats({
+            this.stats_collector.update_cache_latency_stats({
```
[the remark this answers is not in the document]
guymguym: Example with tap_stream:
```javascript
const start = process.hrtime.bigint();
const read_stream = await ...

// update latency stats on 'end'
read_stream.once('end', () => {
  this.stats_collector.update_cache_latency_stats({
    bucket_name: params.bucket,
    cache_read_latency: Number(process.hrtime.bigint() - start) / 1e6,
  });
});

// update bytes stats on 'data' events but with a tap stream
const tap_stream = stream_utils.tap_stream(data => {
  this.stats_collector.update_cache_stats({
    bucket_name: params.bucket,
    read_bytes: params.read_size,
  });
});
read_stream.pipe(tap_stream);

// when returning the read stream to be read from we actually need to return the last stream 
// in our pipeline which is tap_stream so the data will be pulled from it.
return tap_stream;
```

guymguym (changes requested):

src/sdk/namespace_cache.js
```diff
@@ -582,7 +665,15 @@ class NamespaceCache {
...
             const hub_params = { ...params, source_stream: hub_stream };
+            const start = process.hrtime.bigint();
             const hub_promise = this.namespace_hub.upload_object(hub_params, object_sdk);
+            // update latency stats on 'end'
+            hub_promise.once('end', () => {
+                this.stats_collector.update_hub_latency_stats({
+                bucket_name: params.bucket,
+                hub_write_latency: Number(process.hrtime.bigint() - start) / 1e6,
+                });
+            });
```
guymguym, line 676: for a promise we need to use .then like this:
```suggestion
            hub_promise.then(() => this.stats_collector.update_hub_latency_stats({
                bucket_name: params.bucket,
                hub_write_latency: Number(process.hrtime.bigint() - start) / 1e6,
            }));
```

src/sdk/namespace_cache.js
```diff
@@ -599,6 +690,13 @@ class NamespaceCache {
                 params.size,
                 async () => this.namespace_nb.upload_object(cache_params, object_sdk)
             );
+            // update latency stats on 'end'
+            cache_promise.once('end', () => {
+                this.stats_collector.update_hub_latency_stats({
+                bucket_name: params.bucket,
+                cache_write_latency: Number(process.hrtime.bigint() - start) / 1e6,
+                });
+            });
```
guymguym, line 699:
```suggestion
            // update latency stats on 'end'
            cache_promise.then(() => this.stats_collector.update_hub_latency_stats({
                bucket_name: params.bucket,
                cache_write_latency: Number(process.hrtime.bigint() - start) / 1e6,
            }));
```

src/sdk/namespace_cache.js
```diff
@@ -472,26 +527,39 @@ class NamespaceCache {
...
             } else {
                 upload_params.last_modified_time = (new Date(params.object_md.create_time)).getTime();
+                const start = process.hrtime.bigint();
                 _global_cache_uploader.submit_background(
                     params.object_md.size,
                     async () => {
                         const upload_res = await this.namespace_nb.upload_object(upload_params, object_sdk);
-                        this.stats_collector.update_namespace_cache_stats({
+                        // update latency stats on 'end'
+                        upload_res.once('end', () => {
```
guymguym, line 543: As noted on other cases, when using `await upload_object` we don't get back a stream and we do no need to wait more, so we can just update the stats directly after await.

src/sdk/namespace_cache.js
```diff
@@ -472,26 +527,39 @@ class NamespaceCache {
...
 
-                this.stats_collector.update_namespace_cache_stats({
+                this.stats_collector.update_cache_stats({
                     bucket_name: params.bucket,
                     miss_count: 1,
                     range_op: true,
                 });
             } else {
                 upload_params.last_modified_time = (new Date(params.object_md.create_time)).getTime();
+                const start = process.hrtime.bigint();
```
guymguym, line 537: Perhaps we should move this into the upload async function so that it will not count the time sitting in the queue?

I'm not sure what we want here actually because the issue with including the queue time in the latency is that it means we account for other cache writes work multiple times. My instinct is to not include it, or maybe include it as a separate latency stat. 

WDYT?

src/sdk/namespace_cache.js
```diff
@@ -259,15 +264,30 @@ class NamespaceCache {
                             xattr: object_info_hub.xattr,
                             last_modified_time: (new Date(object_info_hub.create_time)).getTime(),
                         };
+
+                        const start = process.hrtime.bigint();
                         const upload_res = await this.namespace_nb.upload_object(upload_params, object_sdk);
-                        this.stats_collector.update_namespace_cache_stats({
+
+                        // update latency stats on 'end'
+                        upload_res.once('end', () => {
```
guymguym, line 272: not relevant to upload calls, only read_object_stream.
so after `await upload_object` we can just update the stats.

src/sdk/namespace_cache.js
```diff
@@ -259,15 +264,30 @@ class NamespaceCache {
...
+                        const tap_stream = stream_utils.get_tap_stream(data => {
+                            this.stats_collector.update_cache_stats({
+                                bucket_name: params.bucket,
+                                write_bytes: object_info_hub.first_range_data.length,
+                                read_bytes: object_info_hub.first_range_data.length,
+                                read_count: 1,
+                                miss_count: 1,
+                            });
+                        });
+                        upload_res.pipe(tap_stream);
```
guymguym, line 289: upload_res is not a stream.

guymguym (changes requested):

src/sdk/namespace_cache.js
```diff
@@ -259,15 +264,30 @@ class NamespaceCache {
...
+                            });
                         });
-                        return upload_res;
+                        
+                        // update bytes stats on 'data' events but with a tap stream
+                        const tap_stream = stream_utils.get_tap_stream(data => {
+                            this.stats_collector.update_cache_stats({
+                                bucket_name: params.bucket,
+                                write_bytes: object_info_hub.first_range_data.length,
+                                read_bytes: object_info_hub.first_range_data.length,
```
guymguym, line 284:
```suggestion
                                write_bytes: data.length,
                                read_bytes: data.length,
```

src/sdk/namespace_cache.js
```diff
@@ -353,12 +373,26 @@ class NamespaceCache {
...
+                bucket_name: params.bucket,
+                cache_read_latency: Number(process.hrtime.bigint() - start) / 1e6,
+                });
+            });
+            
+            // update bytes stats on 'data' events but with a tap stream
+            const tap_stream = stream_utils.get_tap_stream(data => {
+                this.stats_collector.update_cache_stats({
                 bucket_name: params.bucket,
                 read_bytes: params.read_size,
```
guymguym, line 391:
```suggestion
                read_bytes: data.length,
```

guymguym:
To summarize - we are using multiple hooking methods to get the stats updated at the right time.

For latency stats:
1. `await upload_object()` - add the update of latency right after the await.
2. `promise = upload_object()` - add the update on promise.then()
3. `stream = await read_object_stream()` - then we need to update on 'end' event of the stream.

For bytes stats:
1. `await upload_object(upload_params)` - pass a hook method inside upload_params that will be called after some data was written and we can update stats inside object_io as we write parts.
1'. `await upload_part..()` - same as 1
2. `promise = upload_object()` - same as 1
3. `stream = await read_object_stream()` - for cache the tap_stream works as it is now in the PR. for hub we are currently accounting in namespace_s3 that calls update_hub_cache_stats - which is confusing, so need to figure out how to resolve.

src/sdk/namespace_cache.js
mcrepeau: Yes, I moved it under the async, it makes more sense to me

src/native/util/common.h
```diff
@@ -105,7 +105,7 @@ class Exception : public std::exception
 
     virtual const char* what() const throw()
     {
-        return (std::string("Exception: ") + _msg).c_str();
+        return "test";
```
guymguym, line 108: I tested and this works as the quickest fix:
```suggestion
        return _msg.c_str();
```

src/sdk/namespace_cache.js
```diff
@@ -258,15 +263,25 @@ class NamespaceCache {
...
                             last_modified_time: (new Date(object_info_hub.create_time)).getTime(),
+                            async_update_cache_stats: async () => {
+                                this.stats_collector.update_cache_stats({
+                                    bucket_name: params.bucket,
+                                    write_bytes: data.length,
+                                    read_bytes: data.length,
+                                    read_count: 1,
+                                    miss_count: 1,
+                                });
+                            }
```
guymguym, line 274:
```suggestion
                            update_cache_stats: data_length => this.stats_collector.update_cache_stats({
                                  bucket_name: params.bucket,
                                  write_bytes: data_length,
                             });
```

src/sdk/namespace_cache.js
```diff
@@ -449,7 +486,14 @@ class NamespaceCache {
                     upload_params.end = hub_read_range.end;
                     // Set object ID since partial object has been created before
                     upload_params.obj_id = params.object_md.obj_id;
+                    upload_params.async_update_cache_stats = async () => {
+                        this.stats_collector.update_cache_stats({
+                            bucket_name: params.bucket,
+                            write_bytes: upload_params.end - upload_params.start,
+                        });
+                    };
```
guymguym, line 494:
```suggestion
                            update_cache_stats: data_length => this.stats_collector.update_cache_stats({
                                  bucket_name: params.bucket,
                                  write_bytes: data_length,
                             });
```

src/sdk/namespace_cache.js
```diff
@@ -258,15 +263,25 @@ class NamespaceCache {
...
                         const upload_res = await this.namespace_nb.upload_object(upload_params, object_sdk);
-                        this.stats_collector.update_namespace_cache_stats({
+
+                        this.stats_collector.update_cache_latency_stats({
                             bucket_name: params.bucket,
-                            write_bytes: object_info_hub.first_range_data.length,
-                            read_bytes: object_info_hub.first_range_data.length,
-                            read_count: 1,
-                            miss_count: 1,
+                            hub_read_latency: Number(process.hrtime.bigint() - start) / 1e6,
```
guymguym, line 282: cache_write_latency

src/sdk/namespace_cache.js
```diff
@@ -449,7 +486,14 @@ class NamespaceCache {
...
                     // Set object ID since partial object has been created before
                     upload_params.obj_id = params.object_md.obj_id;
+                    upload_params.async_update_cache_stats = async () => {
+                        this.stats_collector.update_cache_stats({
+                            bucket_name: params.bucket,
+                            write_bytes: upload_params.end - upload_params.start,
+                        });
+                    };
 
+                    const start = process.hrtime.bigint();
```
guymguym, line 496: move into the async call just before the upload

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
```diff
@@ -74,6 +75,38 @@ const NOBBAA_ENDPOINT_METRICS = js_utils.deep_freeze([
             labelNames: ['bucket_name']
         }
     },
+    {
+        type: 'Histogram',
```
guymguym, line 79: https://github.com/siimon/prom-client#histogram
need to define the buckets.

src/sdk/endpoint_stats_collector.js
```diff
@@ -132,6 +142,24 @@ class EndpointStatsCollector {
         }
     }
 
+    update_cache_latency_stats({ bucket_name, cache_read_latency, cache_write_latency }) {
+        if (cache_read_latency) {
+            this.prom_metrics_report.inc('cache_read_latency', { bucket_name }, cache_read_latency);
```
guymguym, line 147: inc => observe for histogram

src/sdk/namespace_cache.js
mcrepeau: You mean `upload_params.update_cache_stats: data_length => this.stats_collector.update_cache_stats({` ?

src/sdk/object_io.js
```diff
@@ -237,6 +240,10 @@ class ObjectIO {
                 complete_params.last_modified_time = await params.async_get_last_modified_time();
             }
 
+            if (params.async_update_cache_stats) {
+                complete_params.cache_stats = await params.async_update_cache_stats();
+            }
```
guymguym, line 245: Move down the pipeline to _upload_chunks and call like this:
```
            await mc.run();
            if (mc.had_errors) throw new Error('Upload map errors');
            if (params.update_cache_stats) params.update_cache_stats(params.range.end - params.range.start);
            return callback();
```

src/sdk/namespace_cache.js
```diff
@@ -449,20 +482,28 @@ class NamespaceCache {
                     upload_params.end = hub_read_range.end;
                     // Set object ID since partial object has been created before
                     upload_params.obj_id = params.object_md.obj_id;
+                    upload_params.update_cache_stats = data_length => {
+                        this.stats_collector.update_cache_stats({
+                            bucket_name: params.bucket,
+                            write_bytes: upload_params.end - upload_params.start,
```
guymguym, line 488:
```suggestion
                            write_bytes: data_length,
```

src/sdk/namespace_cache.js
```diff
@@ -119,6 +120,10 @@ class NamespaceCache {
         return params.object_md.size <= cache_config.DEFAULT_BLOCK_SIZE;
     }
 
+    _update_metric(update_metric_fn_name, value) {
```
guymguym, line 123: remove

src/sdk/namespace_cache.js
```diff
@@ -258,15 +263,21 @@ class NamespaceCache {
                             content_type: params.content_type,
                             xattr: object_info_hub.xattr,
                             last_modified_time: (new Date(object_info_hub.create_time)).getTime(),
+                            update_cache_stats: data_length =>
```
guymguym, line 266: make a small utility function:
```javascript
function update_cache_stats_hook(bucket_name) {
  return write_bytes =>this.stats_collector.update_cache_stats({ bucket_name, write_bytes });
} 
```
and then call it here liket his:
```suggestion
                            update_cache_stats: update_cache_stats_hook(params.bucket),
```

src/sdk/namespace_cache.js
```diff
@@ -472,26 +513,35 @@ class NamespaceCache {
...
 
-                this.stats_collector.update_namespace_cache_stats({
+                this.stats_collector.update_cache_stats({
                     bucket_name: params.bucket,
                     miss_count: 1,
                     range_op: true,
                 });
             } else {
                 upload_params.last_modified_time = (new Date(params.object_md.create_time)).getTime();
+                upload_params.async_update_cache_stats = async () => {
```
guymguym, line 523:
```suggestion
                upload_params.update_cache_stats = update_cache_stats_hook(params.bucket);
```

src/sdk/namespace_cache.js
```diff
@@ -449,20 +482,28 @@ class NamespaceCache {
                     upload_params.end = hub_read_range.end;
                     // Set object ID since partial object has been created before
                     upload_params.obj_id = params.object_md.obj_id;
+                    upload_params.update_cache_stats = data_length => {
```
guymguym, line 485:
```suggestion
                    upload_params.upload_chunk_hook = update_cache_stats_hook(params.bucket),
```

guymguym (changes requested):
More questions on this PR:
- pass size to update_cache/hub_latency_stats (can ignore inside for now, just collect)
- beautify/indentation
- document on PR comment an example of the histograms

Improvements out of scope for this PR:
- Make each namespace account read/write byte counters
- Make namespace_nb account read/write byte counters in object_io
- Add size histogram to the side of the latency histogram
- Explore if prometheus can really store more complex information about latency events.

src/sdk/object_io.js
```diff
@@ -495,6 +498,7 @@ class ObjectIO {
             });
             await mc.run();
             if (mc.had_errors) throw new Error('Upload map errors');
+            if (params.async_update_cache_stats()) params.async_update_cache_stats(params.range.end - params.range.start);
```
guymguym, line 501: we don't need it to be called async, and the condition should not run the function - just check for it's existence.
```suggestion
            if (params.update_cache_stats) params.update_cache_stats(params.range.end - params.range.start);
```

src/sdk/object_io.js
```diff
@@ -215,6 +217,7 @@ class ObjectIO {
             'key',
             'md_conditions',
             'last_modified_time',
+            'cache_stats'
```
guymguym, line 220: remove

src/sdk/object_io.js
```diff
@@ -54,6 +54,8 @@ Object.isFrozen(RpcError); // otherwise unused
  * @property {boolean} [complete_upload]
  * @property {number} [last_modified_time]
  * @property {function} [async_get_last_modified_time]
+ * @property {number} [cache_stats]
```
guymguym, line 57: remove

src/sdk/namespace_cache.js
```diff
@@ -258,15 +259,17 @@ class NamespaceCache {
                             content_type: params.content_type,
                             xattr: object_info_hub.xattr,
                             last_modified_time: (new Date(object_info_hub.create_time)).getTime(),
+                            update_cache_stats: update_cache_stats_hook(params.bucket)
                         };
+
+                        const start = process.hrtime.bigint();
```
guymguym, line 265: Can you please rename it across the file to `start_time` or `timestamp`.
I'm a bit worried using "start" in this code where we often use it as a file offset, and we might shadow another variable (lint should error on that, but there are cases where this might miss).

src/sdk/namespace_cache.js
guymguym: right, but with an assignment `=`

src/sdk/object_io.js
guymguym: I suggested to call this property `upload_chunk_hook` - what do you think?

src/sdk/endpoint_stats_collector.js
```diff
@@ -70,9 +70,14 @@ class EndpointStatsCollector {
             io_stats.read_count += count;
             io_stats.read_bytes += size;
         }
-        if (bucket_name) {
-            this.prom_metrics_report.inc('hub_read_bytes', { bucket_name }, size);
-        }
+        this._update_hub_read_stats({
+            bucket_name: bucket_name,
+            size: size
+        })
```
guymguym, line 76:
```suggestion
        this._update_hub_read_stats({ bucket_name, size });
```

src/sdk/endpoint_stats_collector.js
```diff
@@ -86,9 +91,14 @@ class EndpointStatsCollector {
             io_stats.write_count += count;
             io_stats.write_bytes += size;
         }
-        if (bucket_name) {
-            this.prom_metrics_report.inc('hub_write_bytes', { bucket_name }, size);
-        }
+        this._update_hub_write_stats({
+            bucket_name: bucket_name,
+            size: size
+        })
```
guymguym, line 97:
```suggestion
        this._update_hub_write_stats({ bucket_name, size });
```

src/sdk/object_io.js
```diff
@@ -495,6 +496,7 @@ class ObjectIO {
             });
             await mc.run();
             if (mc.had_errors) throw new Error('Upload map errors');
+            if (params.update_cache_stats()) params.update_cache_stats(params.range.end - params.range.start);
```
guymguym, line 499: The condition should not call the function like so `params.update_cache_stats()`
instead it should just check if the property exists like this `params.update_cache_stats`.

Another point - we suggested the name upload_chunk_hook here right?

So see the code suggestion, but also need to change in namespace_cache to use the new property name in the upload_params (inline or assign).

```suggestion
            if (params.upload_chunk_hook) params.upload_chunk_hook(params.range.end - params.range.start);
```

src/sdk/namespace_cache.js
```diff
@@ -759,6 +818,10 @@ class NamespaceCache {
         return hub_res.value;
     }
 
+    function update_cache_stats_hook(bucket_name) {
+        return write_bytes => this.stats_collector.update_cache_stats({ bucket_name,write_bytes })
```
guymguym, line 822: you need to use semicolons to end statements.
You can run `npm run lint` locally to make sure before pushing..

src/sdk/namespace_cache.js
```diff
@@ -449,20 +474,24 @@ class NamespaceCache {
                     upload_params.end = hub_read_range.end;
                     // Set object ID since partial object has been created before
                     upload_params.obj_id = params.object_md.obj_id;
+                    upload_params.update_cache_stats = update_cache_stats_hook(params.bucket);
+                    }
```
guymguym, line 478: looks like this line should be deleted

src/sdk/namespace_cache.js
```diff
@@ -403,7 +420,15 @@ class NamespaceCache {
             if (!hub_read_params.md_conditions) {
                 hub_read_params.md_conditions = { if_match_etag: params.object_md.etag };
             }
+            const start_time = process.hrtime.bigint();
             hub_read_stream = await this.namespace_hub.read_object_stream(hub_read_params, object_sdk);
+            // update latency stats on 'end'
+            hub_read_stream.once('end', () => {
+                this.stats_collector.update_cache_latency_stats({
+                bucket_name: params.bucket,
+                hub_read_latency: Number(process.hrtime.bigint() - start_time) / 1e6,
```
guymguym, line 429: bad indentation - repeats in more cases below

src/sdk/namespace_cache.js
mcrepeau: Oh I didn't know about that command. That sounds good

src/sdk/object_io.js
mcrepeau: It was `update_cache_stats_hook` so I changed it to that. And the other thing was an oversight, I corrected it

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
```diff
@@ -74,6 +75,42 @@ const NOBBAA_ENDPOINT_METRICS = js_utils.deep_freeze([
...
+        name: 'hub_read_latency',
+        configuration: {
+            help: 'hub read latency in namespace cache bucket',
+            labelNames: ['namespace_resource_id'],
+            buckets: [0.1, 10, 100, 200, 500, 1000, 2000],
```
guymguym, line 84: Lets set the buckets of all latency histograms to something like this - added higher/lower values because this includes also small/large objects. 
```suggestion
            buckets: [
                0.001, 
                0.01, 
                0.1, 0.2, 0.5,
                1, 2, 5, 
                10, 20, 50, 
                100, 200, 500, 
                1000, 2000, 5000,
                10000, 20000, 50000,
            ],
```

src/sdk/object_io.js
```diff
@@ -495,6 +496,7 @@ class ObjectIO {
             });
             await mc.run();
             if (mc.had_errors) throw new Error('Upload map errors');
+            if (params.update_cache_stats_hook) params.update_cache_stats_hook(params.range.end - params.range.start);
```
guymguym, line 499: set the name to `upload_chunks_hook`:
```suggestion
            if (params.upload_chunks_hook) params.upload_chunks_hook(params.range.end - params.range.start);
```

src/sdk/object_io.js
```diff
@@ -54,6 +54,7 @@ Object.isFrozen(RpcError); // otherwise unused
  * @property {boolean} [complete_upload]
  * @property {number} [last_modified_time]
  * @property {function} [async_get_last_modified_time]
+ * @property {function} [update_cache_stats_hook]
```
guymguym, line 57:
```suggestion
 * @property {function} [upload_chunks_hook]
```

src/sdk/namespace_cache.js
```diff
@@ -258,23 +259,24 @@ class NamespaceCache {
                             content_type: params.content_type,
                             xattr: object_info_hub.xattr,
                             last_modified_time: (new Date(object_info_hub.create_time)).getTime(),
+                            update_cache_stats: update_cache_stats_hook(params.bucket)
```
guymguym, line 262:
```suggestion
                            upload_chunks_hook: update_cache_stats_hook(params.bucket)
```

src/sdk/namespace_cache.js
```diff
@@ -449,396 +471,445 @@ class NamespaceCache {
...
-        return ret_stream;
-    }
+            this.stats_collector.update_cache_stats({
+                bucket_name: params.bucket,
+                miss_count: 1,
+                range_op: true,
+            });
+        } else {
+            upload_params.last_modified_time = (new Date(params.object_md.create_time)).getTime();
+            upload_params.update_cache_stats = update_cache_stats_hook(params.bucket);
```
guymguym, line 506:
```suggestion
            upload_params.upload_chunks_hook = update_cache_stats_hook(params.bucket);
```

src/sdk/namespace_cache.js
```diff
@@ -449,396 +471,445 @@ class NamespaceCache {
...
-                                    write_bytes: upload_params.end - upload_params.start,
-                                });
-                            return upload_res;
-                        }
-                    );
-                    dbg.log0('NamespaceCache._read_hub_object_stream: started uploading part to cache', params.object_md);
-                } else {
-                    dbg.log0('NamespaceCache._read_hub_object_stream: etags are different or non if-match preconditions, skip uploading part to cache',
-                        { md_conditions: params.md_conditions, cache_object_md: params.object_md });
                 }
```
guymguym, line 474: this is wrong

mcrepeau: Dammit, I missed that

src/sdk/namespace_cache.js
```diff
@@ -531,15 +560,23 @@ class NamespaceCache {
             // For testing purpose: get_from_cache query parameter is on
             try {
                 dbg.log0('NamespaceCache.read_object_stream: get_from_cache is on: read object from cache', params);
+                const start_time = process.hrtime.bigint();
                 read_response = await this.namespace_nb.read_object_stream(params, object_sdk);
+                // update latency stats on 'end'
+                read_response.once('end', () => {
```
guymguym, line 566: missing tap stream to update cache read_bytes metrics

src/sdk/namespace_cache.js
```diff
@@ -599,20 +653,25 @@ class NamespaceCache {
                 params.size,
                 async () => this.namespace_nb.upload_object(cache_params, object_sdk)
             );
+            // update latency stats on 'end'
+            cache_promise.then(() => this.stats_collector.update_hub_latency_stats({
```
guymguym, line 657: wrong call -
```suggestion
            cache_promise.then(() => this.stats_collector.update_cache_latency_stats({
```

src/sdk/namespace_cache.js
```diff
@@ -582,12 +629,19 @@ class NamespaceCache {
...
+            hub_promise.then(() => this.stats_collector.update_hub_latency_stats({
+                bucket_name: params.bucket,
+                hub_write_latency: Number(process.hrtime.bigint() - start_time) / 1e6,
+            }));
 
             const cache_finalizer = callback => hub_promise.then(() => callback(), err => callback(err));
 
             const cache_stream = new stream.PassThrough({ final: cache_finalizer });
-            const cache_params = { ...params,
+            const cache_params = {
```
guymguym, line 643: what about setting `cache_params.upload_chunks_hook` to account cache write_bytes - seems missing or something else is accounting it?

guymguym (changes requested):
Two last things I'm not sure what to do about yet:
- missing size of the operation for the latency metrics.
- are we accounting the hub write_bytes or letting namespace_s3/blob do that?

src/sdk/endpoint_stats_collector.js
```diff
@@ -72,8 +72,8 @@ class EndpointStatsCollector {
         }
         if (bucket_name) {
             this.prom_metrics_report.inc('hub_read_bytes', { bucket_name }, size);
+            this._trigger_send_stats();
```
guymguym, line 75: move outside the condition.

src/sdk/endpoint_stats_collector.js
```diff
@@ -88,8 +88,8 @@ class EndpointStatsCollector {
         }
         if (bucket_name) {
             this.prom_metrics_report.inc('hub_write_bytes', { bucket_name }, size);
+            this._trigger_send_stats();
```
guymguym, line 91: move outside the condition.

guymguym (approved):
@mcrepeau Please just open an issue on the skipped tests in test_namespace_cache. Thanks for all the hard work!

PR #6082, opened 2020-07-08
Fix isa-l_crypto for multi-arch

yselkowitz:
### Explain the changes
These commits, in addition to those pending elsewhere, should fix the build on ppc64le/s390x.

### Testing Instructions:
1. Import https://github.com/catid/gf256/pull/6 into src/native/third_party/cm256
2. Build on x86_64 and ppc64le and/or s390x

PR #6189, opened 2020-09-29
Enable SHA-NI and AVX512 extensions in isa-l_crypto

yselkowitz:
YASM does not support these extensions, so NASM must be used instead,
and preprocessor defines are required to enable as well.

### Explain the changes
1. Use NASM instead of YASM to compile the x86 assembly.

1. Fixed #1878714

### Testing Instructions:
1. Install nasm instead of yasm, then npm run build:native.

### DIFF
--- a/src/deploy/NVA_build/builder.Dockerfile
+++ b/src/deploy/NVA_build/builder.Dockerfile
@@ -11,7 +11,7 @@ ENV container docker
 RUN dnf update -y -q && \
     dnf clean all
 RUN dnf install -y -q wget unzip which vim python2 python3 && \
-    dnf --enablerepo=PowerTools install -y -q yasm && \
+    dnf --enablerepo=PowerTools install -y -q nasm && \
     dnf group install -y -q "Development Tools" && \
     dnf clean all
 RUN alternatives --set python /usr/bin/python3
--- a/src/deploy/NVA_build/dev.Dockerfile
+++ b/src/deploy/NVA_build/dev.Dockerfile
@@ -11,7 +11,7 @@ RUN dnf update -y -q && \
         python3 python3-setuptools \
         gdb strace lsof \
         openssl && \
-    dnf --enablerepo=PowerTools install -y -q yasm && \
+    dnf --enablerepo=PowerTools install -y -q nasm && \
     dnf group install -y -q "Development Tools" && \
     dnf clean all
 
--- a/src/native/asm.gypi
+++ b/src/native/asm.gypi
@@ -2,6 +2,10 @@
 {
     'conditions': [
 
+        [ 'node_arch=="x64"', {
+            'defines': ['HAVE_AS_KNOWS_AVX512', 'HAVE_AS_KNOWS_SHANI'],
+        }],
+
         # LINUX
         [ 'OS=="linux"', {
             'rules': [{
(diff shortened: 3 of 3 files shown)


liranmauda:
Hi @yselkowitz 
thanks for the fix 👍 
We also need to change here:
https://github.com/noobaa/noobaa-core/blob/775f1c5e9fadd2d3a98cc03d4bba4138fe632f30/src/deploy/NVA_build/builder.Dockerfile#L14

then rebase and signed and let's see if travis pass :)

liranmauda (approved):
LGTM

jackyalbo:
trying this code on top of a FIPS cluster didn't do the trick sadly - after building noobaa, got the same result - md5 was wrong
```
> const fips = require('./src/util/fips');
detect_fips_mode: found /proc/sys/crypto/fips_enabled with value 1
OpenSSL 1.1.1g  21 Apr 2020 setting up
init_rand_seed: starting ...
read_rand_seed: opening /dev/random ...
undefined
> read_rand_seed: reading 32 bytes from /dev/random ...
read_rand_seed: got 32 bytes from /dev/random, total 32 ...
init_rand_seed: seeding with 32 bytes
rand_seed: OpenSSL 1.1.1g  21 Apr 2020 seeding randomness
init_rand_seed: done

> fips.get_fips_mode()
true
> console.log(crypto.createHash('md5').update('jacky').digest('hex'));
d41d8cd98f00b204e9800998ecf8427e
undefined
```
It seems this fix does fix the avx512 issue that we saw but not fixing getting the wrong md5 result without specifying a specific arch in ssl_napi.cpp

yselkowitz:
two unrelated issues then?

jackyalbo:
> two unrelated issues then?

yes @yselkowitz It seems that something in your previous PR #6082 is breaking our MD5 calculation, and it seems that this avx512 issue we saw is not really related to it. We need to figure out what is the breaking change in this previous PR.

yselkowitz:
It was pointed out to me that MD5, being a weak cypher by modern standards, might be disabled in FIPS mode.  The md5sum you are getting appears to correspond to an empty string.  Could this simply be a result of a deliberate disablement in FIPS?

nimrod-becker:
MD5 is being required in the S3 protocol....

jackyalbo:
Hey @yselkowitz last time I checked - reverting your last PR - with its 3 commits fixed the issue on top of FIPS. So something in that PR is breaking md5. What exactly was done in the previous PR? and why did we imported MD5 and SHA1 base implementations and those specific files?

nimrod-becker:
@yselkowitz I understand. However, a single noobaa is deployed on a single k8s, either FIPS or not. It's not like AWS where we can provide different endpoints.

We might want to consider removing MD5 at all, but that would mean removing auth v2 and even AWS::S3 keeps support for that for existing customers. In any case it's a decision we need to make on the project and not just a byproduct of enabling multiarch.

Jacky is working on bumping the entire isa-l lib to 2.22 to begin with (#6215) and we can take it from there. For now, we must have MD5 supported in whatever path/solution we choose.

yselkowitz:
Are you still interested in this patch?

liranmauda:
> Are you still interested in this patch?

@yselkowitz 
Yes, we are. Can you rebase and I will marge it?

PR #6192, opened 2020-10-04
Fix data optimization when compression is disabled

romayalon:

### Explain the changes
1. When compression is disabled pick the original size of the chunk instead of the compressed size.

1. Fixed BZ #1842254 (comment 4)

### Testing Instructions:
1.

### DIFF
--- a/src/util/mongo_functions.js
+++ b/src/util/mongo_functions.js
@@ -82,8 +82,9 @@ function map_aggregate_objects() {
  * @this mongodb doc being mapped
  */
 function map_aggregate_chunks() {
+    const compress_size = this.compress_size || this.size;
     emit(['', 'compress_size'], this.compress_size);
-    emit([this.bucket, 'compress_size'], this.compress_size);
+    emit([this.bucket, 'compress_size'], compress_size);
 }
 
 /**


guymguym (changes requested):
@romayalon Notice that `config.CHUNK_CODER_COMPRESS_TYPE` is only a default value, but each chunk actually refers to a chunk_coder (a DB collection) where we define the compress_type configuration which applies to that chunk. This is meant to provide flexibility to have chunks with different compression types in the same system or bucket (for example disable compression for already-compressed media files, or switch to another compression).

So I think we should fix it in map_aggregate_chunks() to use size if compress_size is unset - this makes sense because when a chunk used "no" compression then freely speaking the compressed size is the same as size, it's just that we didn't store it in the DB. This fix should keep everything else intact and also account compress_size correctly for when we disable compression altogether.

WDYT?

```javascript
function map_aggregate_chunks() {
    const compress_size = this.compress_size || this.size;
    emit(['', 'compress_size'], compress_size);
    emit([this.bucket, 'compress_size'], compress_size);
}
```

romayalon:
@guymguym agreed and fixed.

nimrod-becker (approved):

jeniawhite:
@guymguym @nimrod-becker @romayalon 
This needs to be updated in PostgreSQL PL/PGSQL as well.

src/util/mongo_functions.js
```diff
@@ -82,8 +82,9 @@ function map_aggregate_objects() {
  * @this mongodb doc being mapped
  */
 function map_aggregate_chunks() {
+    const compress_size = this.compress_size || this.size;
     emit(['', 'compress_size'], this.compress_size);
```
guymguym, line 86: @romayalon use `compress_size` here too

guymguym: and not `this.compress_size`...
FYI @nimrod-becker

nimrod-becker: ah!

guymguym:
@jeniawhite Right. 
Can you point out where? 
Best if you can add a comment with a permalink to the source.

jeniawhite:
@guymguym 
https://github.com/noobaa/noobaa-core/blob/master/src/util/sql_functions/map_aggregate_chunks.sql

PR #6197, opened 2020-10-05
Return heap allocated memory instead of stack allocated memory for Exception:what()

nb-ohad:
### Explain the changes
1. Allocate a c style string (char array) on the heap to hold the c style variant returned from the `what` function

This is NOT a full fix because the returned memory (the c style string memory) is tied to the lifetime of the exception object. 
If the call site will hold the memory after the deletion of the exception we will still have the same error. 
As of now, it seems that all call sites are not behaving in this manner. 

1. Prevent us from returning an invalid memory address, as the stack frame for the `what` function is invalid upon return
2. Fix a compile error when using clang 12 

### Testing Instructions:
1. 

### DIFF
--- a/src/native/util/common.h
+++ b/src/native/util/common.h
@@ -99,13 +99,19 @@ class Exception : public std::exception
 public:
     Exception(std::string msg)
         : _msg(msg)
-        , _bt(new Backtrace()) {}
+        , _bt(new Backtrace()) {
+            std::string tmp = std::string("Exception: ") + _msg;
+            _c_str_msg = new char[tmp.length() + 1];
+            strcpy(_c_str_msg, tmp.c_str());
+        }
 
-    virtual ~Exception() throw() {}
+    virtual ~Exception() throw() {
+        delete [] _c_str_msg;
+    }
 
     virtual const char* what() const throw()
     {
-        return (std::string("Exception: ") + _msg).c_str();
+        return _c_str_msg;
     }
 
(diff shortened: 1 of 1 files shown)


guymguym (changes requested):
@nb-ohad Lets just make `what()` return an `std::string`.

There are just 3 calls to this method which also could have just used the regular `cerr << ex` option (which includes the stack trace too).

nb-ohad:
@guymguym  I was under the impression this must be (const char *) because it is part of the interface exposed to nodejs or something like that.

guymguym:
@nb-ohad Nope.

guymguym:
@nb-ohad So you were right that we can't change the return type, though the reason is that it is inherited from the base class std::exception.

However since the only complaint is that the pointer returned is local to the function, but we can use a member string like `what() { return _msg.c_str() }` - the docs only require that string to be scoped to the exception instance.

So I suggest this:

```cpp
class Exception : public std::exception
{
public:
    Exception(std::string msg)
        : _msg(XSTR() << "Exception: " << msg)
        , _bt(new Backtrace()) {}

    virtual ~Exception() throw() {}

    virtual const char* what() const throw()
    {
        return _msg.c_str();
    }

    friend std::ostream& operator<<(std::ostream& os, const Exception& e)
    {
        return os << e._msg << " " << *e._bt;
    }

private:
    const std::string _msg;
    const std::shared_ptr<Backtrace> _bt;
};
```

guymguym:
Fixed in #6209. Thanks

PR #6201, opened 2020-10-07
Removing `make tester` from Jenkins unit tests

liranmauda:
### Explain the changes
For some reason, the noobaa-test image is not there when we run `make test` so removing `make tester` from Jenkins unit tests to shorten the cycle (reduce ~14 min).

### DIFF
--- a/.jenkins/unit.groovy
+++ b/.jenkins/unit.groovy
@@ -50,7 +50,7 @@ node('cico-workspace') {
 
 		// real test start here
 		stage('Unit Tests') {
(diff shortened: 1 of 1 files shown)


liranmauda:
@nixpanic can you have a look?

nimrod-becker (approved):

nixpanic (approved):
LGTM, but note that I am not really sure what the `make tester` does, or why it is (not) needed :smile:

liranmauda:
> LGTM, but note that I am not really sure what the `make tester` does, or why it is (not) needed 😄

`make tester` only builds the test image
`make test` or `make tests` (which is an alias for the first) will build and test.
`make test` require tester first: 
https://github.com/noobaa/noobaa-core/blob/ec765a5737e3a9e680ec9759c81990c517254a68/Makefile#L62
If the image is already there then building via test will nor rebuild as all the layers are cached. 
as we can't see the image previously built it is building it again

PR #6207, opened 2020-10-11
Addition of collections for coretest to init

jeniawhite:
### Explain the changes
1. Added require for collections to init them prior to running coretest unit tests 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/server/bg_services/usage_aggregator.js
+++ b/src/server/bg_services/usage_aggregator.js
@@ -96,8 +96,7 @@ async function get_accounts_bandwidth_usage(query) {
             }
 
             return reportList.reduce(
-                (acc, report) => _accumulate_bandwidth(acc, report),
-                { account: account.email, ...ZERO_STATS }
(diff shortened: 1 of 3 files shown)


src/server/analytic_services/endpoint_stats_store.js
```diff
@@ -11,7 +11,7 @@ const endpoint_group_report_schema = require('./endpoint_group_report_schema.js'
 const { ENDPOINT_MONITOR_INTERVAL } = require('../../../config');
 
 class EndpointStatsStore {
-    static get instance() {
+    static instance() {
```
nb-ohad, line 14: Why? I don't see the reasoning behind this change.
instance as a property makes more sense, and functionally they are the same.

nb-ohad (changes requested):

nb-ohad (approved):

src/server/analytic_services/endpoint_stats_store.js
jeniawhite: Rolledback

PR #6208, opened 2020-10-12
Jenkins: Killing the previous job from the same PR

liranmauda:
### Explain the changes
-  Killing the previous job from the same PR

### DIFF
--- a/.jenkins/build-sanity.groovy
+++ b/.jenkins/build-sanity.groovy
@@ -35,28 +35,54 @@ node('cico-workspace') {
 			}
 			firstAttempt = false
 			cico = sh(
-				script: "cico node get -f value -c hostname -c comment --release=8 --retry-count=${cico_retries} --retry-interval=${cico_retry_interval}",
+				script: "cico node get -f value -c hostname -c comment --release=8 \
+							--retry-count=${cico_retries} --retry-interval=${cico_retry_interval}",
 				returnStdout: true
 			).trim().tokenize(' ')
 			env.CICO_NODE = "${cico[0]}.ci.centos.org"
 			env.CICO_SSID = "${cico[1]}"
+			env.CICO_NODE_SSH = "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE}"
 		}
 	}
 
 	try {
 		stage('prepare bare-metal machine') {
 			sh 'scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ./.jenkins/prepare.sh root@${CICO_NODE}:'
-			sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./prepare.sh --workdir=/opt/build/noobaa-core --gitrepo=${ci_git_repo} --ref=${ci_git_ref}"
+			sh "${CICO_NODE_SSH} ./prepare.sh --workdir=/opt/build/noobaa-core --gitrepo=${ci_git_repo} --ref=${ci_git_ref}"
+		}
+
+		stage('stop jobs from the same PR') {
+			jobs = sh (
+				script: "${CICO_NODE_SSH} 'cd /opt/build/noobaa-core/.jenkins/ && ./get_job_numbers.sh --jobName ${JOB_NAME} \
+							--currentBuild ${currentBuild.number} --JENKINS_URL ${JENKINS_URL}'",
+				returnStdout: true
+			).trim().tokenize(' ')
+			if ( jobs.isEmpty() ) {
+				println "Nothing to abort"
+			} else { 
+				for (JobNumber in jobs) {
+					int JobNumberint = JobNumber as int
+					println "Aborting ${JOB_NAME} ${JobNumber}"
+					Jenkins.instance.getItemByFullName("${JOB_NAME}")
+									.getBuildByNumber(JobNumberint)
+									.finish(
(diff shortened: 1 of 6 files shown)


liranmauda:
/retest ci/centos/unit 
test

liranmauda:
@nixpanic Can you have a look?

.jenkins/build-sanity.groovy
```diff
@@ -35,28 +35,55 @@ node('cico-workspace') {
 			}
 			firstAttempt = false
 			cico = sh(
-				script: "cico node get -f value -c hostname -c comment --release=8 --retry-count=${cico_retries} --retry-interval=${cico_retry_interval}",
+				script: "cico node get -f value -c hostname -c comment --release=8 \
```
nixpanic, line 38: I never tried breaking lines in Groovy, not sure it will work. Would be nice to know, finally!

.jenkins/build-sanity.groovy
```diff
@@ -35,28 +35,55 @@ node('cico-workspace') {
...
 
 	try {
 		stage('prepare bare-metal machine') {
 			sh 'scp -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no ./.jenkins/prepare.sh root@${CICO_NODE}:'
-			sh "ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no root@${CICO_NODE} ./prepare.sh --workdir=/opt/build/noobaa-core --gitrepo=${ci_git_repo} --ref=${ci_git_ref}"
+			sh "${CICO_NODE_SSH} ./prepare.sh --workdir=/opt/build/noobaa-core --gitrepo=${ci_git_repo} --ref=${ci_git_ref}"
+		}
+
+		stage('stop jobs from the same PR') {
+			def JobName = "noobaa-core_build-sanity"
```
nixpanic, line 55: just like `currentBuild.number`, there might be a variable for the job name too

.jenkins/build-sanity.groovy
```diff
@@ -35,28 +35,55 @@ node('cico-workspace') {
...
+			jobs = sh (
+				script: "${CICO_NODE_SSH} 'cd /opt/build/noobaa-core/.jenkins/ && ./get_job_numbers.sh --jobName ${JobName} \
+							--currentBuild ${currentBuild.number} --JENKINS_URL ${JENKINS_URL}'",
+				returnStdout: true
+			).trim().tokenize(' ')
+			if ( jobs.isEmpty() == true ) {
```
nixpanic, line 61: do you need the `== true`?

.jenkins/build-sanity.groovy
```diff
@@ -35,28 +35,55 @@ node('cico-workspace') {
...
+				println "Nothing to abort"
+			} else { 
+				for (JobNumber in jobs) {
+					int JobNumberint = JobNumber as int
+					println "Aborting ${JobName} ${JobNumber}"
```
nixpanic, line 66: In addition to aborting, you need to call `cico node done <session>` to return the bare-metal system that was reserved. That probably means setting some additional metadata (the `CICO_SSID`) in the job, fetching it and running the command.

If a reserved machine is not returned, it will stay counted against the quota of the project, until the machine is force-returned by the cico scheduler (after 24h since reserving). That means, you may not be able to reserve machines if jobs get aborted regularly.

.jenkins/build-sanity.groovy
liranmauda: Yes, it was tested and it is working.

.jenkins/build-sanity.groovy
liranmauda: Not sure, I will test.

.jenkins/build-sanity.groovy
liranmauda: after the abort, it goes into:
```
finally {
		stage('return bare-metal machine') {
			sh 'cico node done ${CICO_SSID}'
		}
	}
```

nixpanic: oh, if that is the case, it's all fine!

.jenkins/build-sanity.groovy
liranmauda: We can probably do it with `def JobName = System.getenv('JOB_NAME')` but I don't want to allow the system calls.

nixpanic:
On Mon, Oct 26, 2020 at 05:35:40AM -0700, liranmauda wrote:
> We can probably do it with `def JobName = System.getenv('JOB_NAME')` but I don't want to allow the system calls. 

I think `env.JOB_NAME` should have the same result?

liranmauda:
> I think `env.JOB_NAME` should have the same result?

looks like `JOB_NAME` is already an environment variable.

nixpanic:
This looks pretty good to me. Just one question:

- How do you validate that the job that is selected for aborting, is a job for the same PR as the current job?

liranmauda:
> This looks pretty good to me. Just one question:
> 
> * How do you validate that the job that is selected for aborting, is a job for the same PR as the current job?

We are doing that in the `get_job_numbers.sh` script: 
1. First, we are getting the current job PR number into `current_job_pr_number` by calling the function `get_pr_number`
2. Then we are getting all the URLs into the variable `running_jobs_xml`
3. After that we are running all the URLs (in the `while`) and calling  the function `get_pr_number`
4. If the PR number of the running job we are checking is the same as the current job PR number we are adding it to the array.

nixpanic (approved):

nimrod-becker (approved):

PR #6215, opened 2020-10-14
update to isa-l crytpo 2.22.0

jackyalbo:
### Explain the changes
1. Updating isa-l crypto to 2.22.0
2. MD5 (And SHA1 - marked out due to #6223) unit-test

1. Fixed BZ#1878714

### Testing Instructions:
1. MD5 unit test


src/native/third_party/isa-l.gyp
```diff
@@ -129,7 +129,6 @@
                 'isa-l_crypto/include/types.h',
                 'isa-l_crypto/include/multi_buffer.h',
                 'isa-l_crypto/md5_mb/md5_ctx_base.c',
-                'isa-l_crypto/md5_mb/md5_ctx_base_aliases.c',
```
guymguym, line 132: Don't you need to remove aliases from sha1 too ?? or is it different ?

guymguym: Also can you add the bases of the rest of the hashes defined in this file?

jackyalbo: seems like sha1 have sha1_ctx_base_aliases.c in this version

src/native/third_party/isa-l_crypto/md5_mb/Makefile.am
```diff
@@ -27,32 +27,37 @@
...
+lsrc_x86_64 += 	md5_mb/md5_mb_mgr_submit_avx512.asm \
+		md5_mb/md5_mb_mgr_flush_avx512.asm \
+		md5_mb/md5_mb_x16x2_avx512.asm \
+		md5_mb/md5_ctx_avx512.c
 
-src_include +=  -I $(srcdir)/md5_mb
-extern_hdrs +=  include/md5_mb.h \
+lsrc_x86_32 += $(lsrc_x86_64)
+
+lsrc_aarch64 += md5_mb/md5_ctx_base.c
```
guymguym, line 57: @jackyalbo Can you verify that all of the sources listed here are in our gyp?
Notice the difference in arch they added (32, 64, ARM).
I would go over the other hashes we compile and verify each one has all the sources per arch in the gyp - md5, sha1, sha256, sha512.

guymguym (changes requested):
See couple of comments

src/native/third_party/isa-l_crypto/md5_mb/Makefile.am
jackyalbo: there is some files missing like include/memcpy_inline.h, and md5_mb/md5_ref.c. Do we need it, I think those files are not new

jackyalbo:
@guymguym please take a look if it's looks ok - added the bases for sha256 and sha512
as well as adding sha1_ni_x2.asm and sha256_ni_x2.asm.

didn't add those: include/memcpy_inline.h, and md5_mb/md5_ref.c
let me know if we need them

src/native/third_party/isa-l.gyp
jackyalbo: it is not - I guess I should remove?

src/native/third_party/isa-l.gyp
```diff
@@ -245,14 +234,15 @@
...
                 'isa-l_crypto/sha256_mb/sha256_opt_x1.asm',
                 'isa-l_crypto/sha256_mb/sha256_ni_x1.asm',
+                'isa-l_crypto/sha256_mb/sha256_ni_x2.asm',
                 'isa-l_crypto/sha256_mb/sha256_ctx_sse_ni.c',
                 'isa-l_crypto/sha256_mb/sha256_ctx_avx512_ni.c',
                 'isa-l_crypto/sha256_mb/sha256_mb_mgr_submit_sse_ni.asm',
                 'isa-l_crypto/sha256_mb/sha256_mb_mgr_flush_sse_ni.asm',
                 'isa-l_crypto/sha256_mb/sha256_mb_mgr_flush_avx512_ni.asm',
-            ],
+            ]}]],
```
guymguym, line 243: If I understand the sha1_mb/Makefile.am correctly for non x64 cases, I think that when they set `lsrc_base_aliases += sha1_mb/sha1_ctx_base_aliases.c` and then in the top Makefile.am it uses `if CPU_UNDEFINED libisal_crypto_la_SOURCES += ${lsrc_base_aliases}` that aliases is needed for the other cases. Maybe we need to add an "else" option to this condition (the third dict of the condition is "else").

```suggestion
            ]} , {
                'sources': [
                    'isa-l_crypto/sha1_mb/sha1_ctx_base_aliases.c',
                ]
            }]],
```

src/native/third_party/isa-l.gyp
```diff
@@ -125,47 +125,31 @@
             ],
             'sources': [
                 # include
-                'isa-l_crypto/include/md5_mb.h',
-                'isa-l_crypto/include/types.h',
-                'isa-l_crypto/include/multi_buffer.h',
                 'isa-l_crypto/md5_mb/md5_ctx_base.c',
```
guymguym, line 128: I'm not sure about if on x64 the base source should be *after* or *before* the other sources of sse/avx to allow them to take over when the CPU supports it, or the order doesn't matter. Can you check? Maybe use linux `nm` on the output library to see where the symbols come from...

Also I don't think it makes sense that we don't have any unit-test that will verify our build works for the target platform.

src/test/unit_tests/test_nb_native_hashes.js
```diff
@@ -0,0 +1,68 @@
...
+            md5(Buffer.from(s));
+        });
+        mocha.it(`SHA1 ${s}`, function() { // known issue #6223
+            sha1(Buffer.from(s));
+        });
+    }
+
+    for (let i = 0; i < 1000; ++i) {
+        const input = crypto.randomBytes(i);
+        mocha.it.only(`MD5 length${i}`, function() {
```
guymguym, line 49: remove `.only`, also in the random test below...

src/test/unit_tests/test_nb_native_hashes.js
```diff
@@ -0,0 +1,68 @@
...
+        ]) {
+        mocha.it.only(`MD5 ${s}`, function() {
+            md5(Buffer.from(s));
+        });
+        mocha.it(`SHA1 ${s}`, function() { // known issue #6223
+            sha1(Buffer.from(s));
+        });
+    }
+
+    for (let i = 0; i < 1000; ++i) {
```
guymguym, line 47: I think we can reduce to 10

src/test/unit_tests/test_nb_native_hashes.js
```diff
@@ -0,0 +1,68 @@
...
+        const input = crypto.randomBytes(i);
+        mocha.it.only(`MD5 length${i}`, function() {
+            md5(input);
+        });
+        mocha.it(`SHA1 length${i}`, function() { // known issue #6223
+            sha1(input);
+        });
+    }
+
+    for (let i = 0; i < 1000; ++i) {
```
guymguym, line 57: I think we can reduce to 10

guymguym (commented):
Need to fix the test, but otherwise looks good

guymguym (approved):

PR #6216, opened 2020-10-14
Pre-build lib files and include it in the repository

nb-ohad:
### Explain the changes
Refactor the lib build process into an offline task. Upon lib dependencies change the developer should run this task on his dev environment and include the results (the src/lib folder) in the commit that is pushed to the repository.

This removes any dependencies on bower in our upstream and downstream automated build processes. automated build processes.

1. 

### Testing Instructions:
1.


liranmauda (approved):
One comment:
Instead of running `npm run build-deps` we can add an entry to the Makefile

nb-ohad:
@liranmauda I am not sure a make file entry is needed as this command should run only when the dependencies change and on the dev computer itself and the result should be pushed into the repo

nimrod-becker (approved):

nimrod-becker:
@guymguym just FYI

nb-ohad:
We can always add a make target in a future commit if we find that it is necessary

PR #6225, opened 2020-10-19
config.js: organize mem limits for buffering + allow local-config to export a function

guymguym:

### Explain the changes
1. See title.
2. The reasoning is explained in the initial comment in config.js.
3. Specifically - we now use CONTAINER_MEM_LIMIT as the input for our semaphores that limit buffers, rather than CONTAINER_MEM_REQUEST. When these are configured to the same value there is clearly no difference, but when  configured with a limit above the request, it is actually the buffering activities that should be using extended memory for burst processing of requests.
4. This change will affect both object_io and namespace_cache - CC @paulxuzhang.

1. NA

### Testing Instructions:
1. To test that the fallbacks work set resources in spec without `requests`, which will make it "0" in the container env, for example:
```
kind: NooBaa
...
spec:
  coreResources: {}
  dbResources: {}
  endpoints:
    resources: {}
```


config.js
```diff
@@ -5,16 +5,53 @@
...
+config.CONTAINER_MEM_REQUEST = Number(process.env.CONTAINER_MEM_REQUEST || '');
+config.CONTAINER_CPU_REQUEST = Number(process.env.CONTAINER_CPU_REQUEST || '');
+
+// For buffer limits we use the mem LIMIT and not REQUEST because -
+// if we are in guaranteed mode then it will equal to requests anyway,
+// but if limit is higher then we want to allow buffering using burst memory,
+// since these buffers will get garbage collected after processing.
+//
+// We divide by 4 because we don't have a central buffer manager, and we need
+// some fixed top limit to how much these buffers can cosume from the total,
```
liranmauda, line 43:
```suggestion
// some fixed top limit to how much these buffers can consume from the total,
```

config.js
```diff
@@ -5,16 +5,53 @@
...
+//
+// When to use LIMIT:
+// - For limitting amounts of transient memory.
+// - Like we do with semaphores to backpressure incoming I/O requests.
+// - Suitable for memory that will be garbage collected soon enough after being processed.
+// -  The heap will be able to burst up and down.
+//
+// When to use REQUEST:
+// - Less common and is currently unused
+// - Might be needed when limitting pinned memory like DB caches.
```
liranmauda, line 30:
```suggestion
// When to use LIMIT:
// - For limiting amounts of transient memory.
// - As we do with semaphores to backpressure incoming I/O requests.
// - Suitable for memory that will be garbage collected soon enough after being processed.
// -  The heap will be able to burst up and down.
//
// When to use REQUEST:
// - Less common and is currently unused
// - Might be needed when limiting pinned memory like DB caches.
```

src/endpoint/endpoint.js
```diff
@@ -190,11 +189,13 @@ async function start_monitor(internal_rpc_client) {
                 group_name,
                 hostname,
                 cpu: {
-                    count: os_utils.get_cpus(),
+                    // using the upper limit here which will includes optional burst resources
```
liranmauda, line 192:
```suggestion
                    // using the upper limit here which will include optional burst resources
```

src/endpoint/endpoint.js
```diff
@@ -190,11 +189,13 @@ async function start_monitor(internal_rpc_client) {
...
+                    count: config.CONTAINER_CPU_LIMIT,
                     usage: cpu_usage
                 },
                 memory: {
-                    total: os_utils.get_memory(),
+                    // using the upper limit here which will includes optional burst resources
```
liranmauda, line 197:
```suggestion
                    // using the upper limit here which will include optional burst resources
```

jeniawhite (approved):

nimrod-becker (approved):

PR #6227, opened 2020-10-20
Add travis job to verify frontend library build

nb-ohad:

### Explain the changes
1. 

1. 

### Testing Instructions:
1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,13 @@ matrix:
         - make tester ${NO_CACHE} ${SUPPRESS_LOGS} || exit 1
         - make test ${SUPPRESS_LOGS}
 
+    - language: generic
+
+      name: "Frontend Library Build Test"
+
+      script:
+        - make verify-fe-lib
+
     - language: node_js
       node_js:
         - "12"
--- a/src/deploy/NVA_build/FrontendLib.Dockerfile
+++ b/src/deploy/NVA_build/FrontendLib.Dockerfile
@@ -1,0 +1,31 @@
+FROM noobaa-builder
+
+######################################################################
+# Layers:
+#   Title: npm install (using package.json)
+#   Size: ~ 805 MB
+#   Cache: Rebuild when there is new package.json or package-lock.json
+######################################################################
+COPY ./package*.json ./
+RUN npm install --production && \
+    npm cache clean --force
+
+######################################################################
+# Layers:
+#   Title: npm install the frontend
+#   Size: ~ 116 MB
+#   Cache: Rebuild when there is new package.json or package-lock.json 
+######################################################################
+COPY ./frontend/package*.json ./frontend/
+RUN cd frontend && \
+    npm install
+
+######################################################################
+# Layers:
+#   Title: Copying the code and Building the frontend library
+#   Size: ~ 245 MB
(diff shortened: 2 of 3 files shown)


src/deploy/NVA_build/FrontendLib.Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+#   Cache: Rebuild when there is new package.json or package-lock.json 
+######################################################################
+COPY ./frontend/package*.json ./frontend/
+RUN cd frontend && \
+    npm install
+
+##############################################################
+# Layers:
+#   Title: Copying the code and Building the frontend
+#   Size: ~ 131 MB
```
liranmauda, line 26: The Title should represent the building of the frontend dependencies.
The size is probably different, you can remove it (or update the value)

.travis.yml
```diff
@@ -34,6 +34,13 @@ matrix:
         - make tester ${NO_CACHE} ${SUPPRESS_LOGS} || exit 1
         - make test ${SUPPRESS_LOGS}
 
+    - language: generic
```
nimrod-becker, line 37: Don't think it should be another spawn on its own. It will take another machine and we are bound on resources as is.

Since systems tests are the shorter one (for now) I think we can add this as a step to that job (after the tests if you are afraid of noise created by the compilation)

src/deploy/NVA_build/FrontendLib.Dockerfile
```diff
@@ -0,0 +1,31 @@
+FROM noobaa-builder
```
nimrod-becker, line 1: this dockerfile is for the test only right ?

nimrod-becker (changes requested):

src/deploy/NVA_build/FrontendLib.Dockerfile
liranmauda: Yes, as building the frontend dependencies needs tools such as node, npm, and some dependencies from the core, the easiest way to get it is by running docker built.
Noobaa-builder containers all the build needs, and on this Dockerfile (FrontendLib)  we are also running `npm install --production` on the core so we will have the modules required by the frontend.

.travis.yml
liranmauda: We can, but currently, this job is fast (~4 min) so this has little effect. 
The upside of adding it as another spawn is that we can add to it all the frontend unit tests

nimrod-becker (approved):

.travis.yml
nb-ohad: No, We cannot add the frontend tests as this build is destructive for the frontend, this is also a reason to run it as a diff spawn. I think the FE tests should be incorporated into the "Unit Tests" job

liranmauda: It is not distractive as it is inside an image...

nb-ohad: I mean you cannot run the FE tests (on a container build with that image) after you run this code

src/deploy/NVA_build/FrontendLib.Dockerfile
nb-ohad: done

liranmauda (approved):

PR #6231, opened 2020-10-21
Deepscan fixes - Parameter of local function is not used.

liranmauda:
### Explain the changes
- Deepscan fixes - Parameter  of local function  is not used.


liranmauda:
closing as @nimrod-becker removed `UNUSED_PARAM` from the deepscan rules

PR #6233, opened 2020-10-22
Moving to isa-l-2.23 to support Z

### LINKED ISSUE #6223: native code sha1 calculation is wrong
### Environment info
- NooBaa Version: master
- Platform: local

### Actual behavior
1. running the following code to calculate sha1:
SHA1_MB)().update(Buffer.from("jacky")).digest().toString("hex")
2. got eea339da0d4b6b5eefbf5532901860950907d8af
3. expected result is: 64366715a186eea098dd2391f69c1afda4507574

### Expected behavior
1. expected result is: 64366715a186eea098dd2391f69c1afda4507574

### Steps to reproduce
1. complie native code 
2. run the above command  
```node -p 'new (require("./src/util/nb_native")().SHA1_MB)().update(Buffer.from("jacky")).digest().toString("hex")'```
3. verify it's as expected - 64366715a186eea098dd2391f69c1afda4507574

### PR DESCRIPTION
**Pull request by jackyalbo**

### Explain the changes
1. updated isa-i-crypto to 2.23 instead of 2.22 - in order to support IBM Z
2. added some missing adaptions to 2.23 in order to support both endians for md5 and sha1, contributed by Ulrich Weigand from IBM

1. Fixed #6223 

### Testing Instructions:
1.

### CODE REVIEW

File: src/native/tools/ssl_napi.cpp [hunk 1]
```diff
@@ -87,7 +89,10 @@ class Hasher : public Napi::ObjectWrap<HASHER_ARGS>
     {
         Napi::Env env = info.Env();
         submit_and_flush(0, 0, HASH_LAST);
-        return Napi::Buffer<uint8_t>::Copy(env, reinterpret_cast<const uint8_t*>(&hash_ctx_digest(&_ctx)[0]), LEN);
+        uint32_t digest[NWORDS];
+        for (unsigned long i = 0; i < NWORDS; i++)
```
guymguym, line 93: wrap the for body in { }

guymguym: also change from `unsigned long` to size_t

src/native/tools/ssl_napi.cpp
```diff
@@ -87,7 +89,10 @@ class Hasher : public Napi::ObjectWrap<HASHER_ARGS>
...
+            digest[i] = WORDS_BE ? be32toh(hash_ctx_digest(&_ctx)[i]) : le32toh(hash_ctx_digest(&_ctx)[i]);
+        return Napi::Buffer<uint8_t>::Copy(env, reinterpret_cast<const uint8_t*>(digest), NWORDS * 4);
```
guymguym, line 95: I think we should be able to just return like this (didn't test) -
```suggestion
        return Napi::Buffer<uint32_t>::Copy(env, digest, NWORDS);
```

src/native/tools/ssl_napi.cpp
```diff
@@ -127,7 +133,8 @@ typedef Hasher<
 
 typedef Hasher<
     SHA1_MB_ALG,
-    SHA1_DIGEST_NWORDS * 4,
+    SHA1_DIGEST_NWORDS,
+    true, // Digest words for SHA1 are big-endian
```
guymguym, line 137: Can you open an issue to ask on this for isa-l_crypto?
It is unclear why this is the chosen behavior right?

src/native/third_party/isa-l.gyp
```diff
@@ -198,6 +202,7 @@
             ]}, {
                 'sources': [
                     'isa-l_crypto/sha1_mb/sha1_ctx_base_aliases.c',
+		            'isa-l_crypto/sha1_mb/sha1_ref.c',
```
guymguym, line 205: fix indentation. I actually think they added it to lsrc_base_aliases by mistake but not sure so it's ok to add too.

src/native/tools/ssl_napi.cpp
jackyalbo: both compilation and unit test passed - so changed to this

src/native/tools/ssl_napi.cpp
jackyalbo: opened https://github.com/intel/isa-l_crypto/issues/57

src/native/third_party/isa-l.gyp
jackyalbo: fixed - also added support for ARM. so please take a quick re-look.

src/native/chunk/splitter.cpp
```diff
@@ -86,7 +87,9 @@ Splitter::finish(uint8_t* md5, uint8_t* sha256)
     if (md5) {
         if (_md5_mb_ctx) {
             md5_mb_submit_and_flush(0, 0, HASH_LAST);
-            memcpy(md5, hash_ctx_digest(_md5_mb_ctx), MD5_DIGEST_NWORDS * 4);
+            uint32_t *digest = reinterpret_cast<uint32_t*>(md5);
+            for (int i = 0; i < MD5_DIGEST_NWORDS; i++)
```
guymguym, line 91: warp the for body in curly braces

src/native/third_party/isa-l.gyp
```diff
@@ -150,7 +150,18 @@
                 'isa-l_crypto/md5_mb/md5_mb_mgr_flush_avx512.asm',
                 'isa-l_crypto/md5_mb/md5_mb_x16x2_avx512.asm',
                 'isa-l_crypto/md5_mb/md5_ctx_avx512.c',
-            ]}]],
+            ]}, 'node_arch=="arm64"', {'sources': [
+                'isa-l_crypto/md5_mb/aarch64/md5_ctx_aarch64_asimd.c',
+                'isa-l_crypto/md5_mb/aarch64/md5_mb_aarch64_dispatcher.c',
+                'isa-l_crypto/md5_mb/aarch64/md5_mb_mgr_aarch64_asimd.c',
+                'isa-l_crypto/md5_mb/aarch64/md5_mb_asimd_x4.S',
```
guymguym, line 157: I don't think if gyp has a builtin rule for .S or we need a custom rule for it like we do for .asm in:
https://github.com/noobaa/noobaa-core/blob/master/src/native/asm.gypi

See https://stackoverflow.com/a/51110745/2702164
I would look in the Makefile.am/make.inc if gcc handles it or yasm/nasm.

jackyalbo: $(O)/%.o: $(host_cpu)/%.S
	@echo "  ---> Building $< $(msg)"
	@$(AS) $(ASFLAGS) -o $@ $<

indeed yasm/nasm

gbtucker: But in this case AS is gcc and gcc passes to gas.

    ifeq ($(arch),aarch64)
        AS=$(CC) -D__ASSEMBLY__

guymguym: @gbtucker does ISA-L support both nasm/yasm/gas all the same?
Can we use gas to build on all archs?
Is there any assembler we should prefer to use for ISA-L?

gbtucker: For x86, nasm is the most up-to-date.  Nasm can build the x86 code for unix or windows.  The arm optimizations were written for gas so I think you must use gcc when building for arm unix.  You can mix other compilers for x86; gcc, icc, vc, etc.

guymguym: thank you very much @gbtucker!

liranmauda:
@jackyalbo 
I have rebased in my own branch from master and run the unit-tests
It passed on both Mac and Linux.
I don’t see any noticeable issue there.
Can you rebase, then run tests?
We can merge if we don’t see any new change/Issue.

liranmauda (approved):
LGTM

liranmauda:
closing in favor of #6398

PR #6234, opened 2020-10-22
system_store: fix to filter deleted properly on first load

guymguym:

### Explain the changes
1. Before this fix I've seen errors in the logs when the system store data was first loaded because it found deleted items.
2. This happened since `this.old_db_data` was undefined or null, so itskipped calling `this._update_data_from_new` which means that deleted items were not filtered from the new_data.
3. I fixed to always call it with an empty object for the first time.

1. NA

### Testing Instructions:
1. Watch the logs after core/endpoint restart


jackyalbo (approved):
LGTM

jeniawhite (approved):

PR #6239, opened 2020-10-27
bulk updates changes for postgres_client

dannyzaken:

### Explain the changes
1. changed `PgTransaction` to a more generic class that only handles transactions
2. change bulk operations to use PgTransaction` instead of extending it.
3. use `PgTransaction` in other places that use transactions (e.g. `findOneAndUpdate`)

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/test/unit_tests/test_system_store.js
+++ b/src/test/unit_tests/test_system_store.js
@@ -15,6 +15,7 @@ const assert = require('assert');
 const system_store = require('../../server/system_services/system_store').get_instance();
 
 mocha.describe('system_store', function() {
+    this.timeout(90000); // eslint-disable-line no-invalid-this
 
     // eslint-disable-next-line no-undef
     afterEach(function() {
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -48,146 +48,178 @@ function get_id(data) {
     return data._id;
 }
 
+function not_implemented() {
+    throw new Error('NOT IMPLEMENTED');
+}
+
+
+let trans_counter = 1;
+
 class PgTransaction {
 
-    constructor({ client, name, ordered }) {
-        this.name = name;
-        this.ordered = ordered;
-        this.length = 0;
+    constructor(client) {
+        this.transaction_id = trans_counter;
+        trans_counter += 1;
         this.client = client;
-        this.operations = [];
         this._init = this._begin();
-        this.nInserted = 0;
-        this.nMatched = 0;
-        this.nModified = 0;
     }
 
     async _begin() {
         this.pg_client = await this.client.pool.connect();
-        await _do_query(this.pg_client, { text: 'BEGIN TRANSACTION' });
+        await _do_query(this.pg_client, { text: 'BEGIN TRANSACTION' }, this.transaction_id);
         this._init = null;
     }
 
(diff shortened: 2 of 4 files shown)


src/util/postgres_client.js
```diff
@@ -49,146 +49,178 @@ function get_id(data) {
...
-    }
+class UnorderedBulkOp extends BulkOp {
 
     find(selector) {
         return {
             // TODO length?
-            length: this.length,
-            remove: () => { throw new Error('TODO'); },
+            length: this.queries.length,
+            remove: () => { throw new Error('NOT IMPLEMENTED'); },
```
jeniawhite, line 197: Call the `not_implemented` method that you've declared above?

src/util/postgres_client.js
```diff
@@ -49,146 +49,178 @@ function get_id(data) {
...
-    }
+class OrderedBulkOp extends BulkOp {
 
     find(selector) {
         return {
-            delete: () => { throw new Error('TODO'); },
-            deleteOne: () => { throw new Error('TODO'); },
-            replaceOne: doc => { throw new Error('TODO'); },
-            update: doc => { throw new Error('TODO'); },
+            delete: () => { throw new Error('NOT IMPLEMENTED'); },
```
jeniawhite, line 218: Call the `not_implemented` method that you've declared above?

src/util/postgres_client.js
```diff
@@ -268,6 +300,14 @@ class PostgresTable {
         return {};
     }
 
+    async insertOneTransactional(transaction, data) {
```
jeniawhite, line 303: I would maybe add `_` to mark it as internal, we don't want to expose this as far as I understand.

src/util/postgres_client.js
```diff
@@ -300,6 +340,22 @@ class PostgresTable {
             return res;
         } catch (err) {
             dbg.error(`updateOne failed`, selector, update, query, err);
+            throw err;
+        }
+    }
+
+    async updateOneTransactional(transaction, selector, update, options = {}) {
```
jeniawhite, line 347: I would maybe add `_` to mark it as internal, we don't want to expose this as far as I understand.

src/test/unit_tests/test_system_store.js
```diff
@@ -15,6 +15,7 @@ const assert = require('assert');
 const system_store = require('../../server/system_services/system_store').get_instance();
 
 mocha.describe('system_store', function() {
+    this.timeout(90000); // eslint-disable-line no-invalid-this
```
jeniawhite, line 18: Not sure if we shall increase the timeout now. Maybe do it later after optimizations.
Since this code is relevant to MongoDB as well.

jeniawhite (approved):

src/test/unit_tests/test_system_store.js
dannyzaken: in mongo it shouldn't affect since it already passes with the default TO of 2 seconds. I prefer it will also pass for postgres. it will be easier to test during development

jeniawhite: The thing is that we have other tests that are not passing due to timeout.
But okay, we can add timeouts to them lazily.

PR #6253, opened 2020-11-03
code maintenance - cleanup NET_STORAGE

guymguym:

### Explain the changes
1. Remove type from schemas/apis/code.
2. Make sure all apis and schemas use the same common definitions like so - we had multiple copies of those enums before -
```
                endpoint_type: { $ref: 'common_api#/definitions/endpoint_type' },
                auth_method: { $ref: 'common_api#/definitions/cloud_auth_method' },
```
3. Remove cp_code field

1. NA

### Testing Instructions:
1. @nimrod-becker we might to issue a warning in the release notes in case someone happened to use these options.


guymguym:
When reviewing this PR please notice:
1. 3 schemas changed - account_schema, pool_schema, namespace_resource_schema.
2. 5 apis changed - common_api, account_api, node_api, pool_api, hosted_agents_api.
3. If a system happened to use these options in the previous version it will break on upgrade - I think we should denote that somehow in the release notes, but I wouldn't do a major version because we do not expect use cases outside those specific vendors which we are unaware of.

nb-ohad (approved):

nb-ohad:
I have only one observation:
If we already have `netstorage` or `flashblade` "values" in the DB these changes will cause schema errors and FE will crush.
But I think that there are no active systems that we are aware of that have cloud connections/cloud resources set up this way.

nb-ohad:
Another minor observation:
The changes in this PR also affect the order of cloud types as they are presented in some dropdowns in the Management Console. The order is controlled by the order in the lists that are provided to the dropdown components.

For "Edit" forms/modals the order does not matter at all as the value will be the value of the entity loaded into the form, but for dropdowns in creation forms/modals, the default value is usually the first item in the list

This PR will change the default value from "AWS" to "S3 V2 Compatible". 
I guess the change from a specific vendor to s3 compatible is ok, but S3 V2 is on its way out so I think V4 should be the default.

guymguym:
I changed the PR to only cleanup NET_STORAGE.
@nb-ohad @jeniawhite - can you make a quick pass?

PR #6255, opened 2020-11-03
code maintenance - fix tsc/eslint errors

guymguym:
### Explain the changes
1. Fixed the problems complaint by tsc for not having type information on the s3 errors and ops.
2. Same for blob and lambda.
3. NOTE: now the error spec objects are frozen.
4. Fixed some more lint/tsc errors.

1. NA

### Testing Instructions:
1. NA

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -131,7 +131,13 @@ module.exports = {
 
         'function-call-argument-newline': 'off',
 
-        'space-unary-ops': ['error', { words: false, nonwords: false }],
+        'space-unary-ops': ['error', {
+            words: true,
+            nonwords: false,
+            overrides: {
+                typeof: false,
+            },
+        }],
 
         // max line length is 80 by default, allow some slack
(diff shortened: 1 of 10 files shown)


liranmauda (approved):
LGTM.
Just one thing, I don't see that we are using the details in all the class error extensions.
Where it should be used?

liranmauda:
There is a lint error: 
```
/noobaa-core/src/util/fips.js
  61:41  error  Unary word operator 'new' must be followed by whitespace  space-unary-ops
✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

guymguym:
@liranmauda We use it manually for the MalformedPolicy error:
https://github.com/noobaa/noobaa-core/blob/18a375934e8df29e30ea13c20de69e9730be0489/src/endpoint/s3/s3_rest.js#L445-L448

There could be cleaner way to do this, but since this is just one error then I think we can wait for the next use case before we generalize.

PR #6258, opened 2020-11-09
Adding yasm back to builder image

liranmauda:
### Explain the changes
- As we can't compile on Mac, we will use yasm. We will install yasm and nasm on the builder image.


guymguym:
But why is the linux docker image relevant to mac?

liranmauda:
You are correct, I missed the fact that it is under: `[ 'OS=="mac"', {`
Closing this PR.

PR #6263, opened 2020-11-12
Bumping NodeJS to v14

liranmauda:
### Explain the changes
.nvmrc:
- Bump NodeJS from 12.18.2 to 14.15.0

package.json:
- remove unused `12` node module
- Bumped all packages

napi.h:
- Removed pragma GCC diagnostic

commin.gypi:
- Moving the `'-std=c99'` flag into `cflags_c`
- Adding `-Wno-cast-function-type` to `cflags`

Other changes:
- Bumped `C++11` to `C++17`
- Replaced `mongo_utils.is_object_id` with `schema_utils.is_object_id` 
- console wrapper: Changing the way we hijack the global.console
- Tests.Dockerfile rearrange
- Enable ws mask
- Removing pipeline.js and replacing it with native pipeline

test_s3_encryption.js:
- Moving the skip into the describe so if there are no AWS credentials found in env. we will skip the entire test

### Testing Instructions:
1. Test full regression
2. special care for the native code path as we change the `C++11` to `C++17` and removed the `'-Werror'`


src/native/nb_native.gyp
```diff
@@ -7,7 +7,7 @@
             '-W',
             '-Wall',
             '-Wextra',
-            '-Werror',
+            # '-Werror',
```
guymguym, line 10: I would really prefer to keep failing on warnings for most of the code.
Wouldn't it be enough to disable warnings for third_party code?
For any code that we wrote we can easily fix warnings.
Check out this common_third_party.gypi and see maybe this can be done here:
https://github.com/noobaa/noobaa-core/blob/5d97fd2e7c97d7d778d69e11deb7730332dfa254/src/native/third_party/common_third_party.gypi#L4-L8

src/native/third_party/common_third_party.gypi
```diff
@@ -2,7 +2,7 @@
 {
     'includes': ['../common.gypi'],
     'variables': {
-        'disabled_warnings': ['-W', '-Wall'],
+        'disabled_warnings': ['-W', '-Wall', '-Werror'],
```
guymguym, line 5: Can you try to disable just `-Wextra` and `-Wpedantic` instead of disabling all warning-as-errors and see if that works for third_party? I would still rather the default `-Werror` to be applied for warnings that we choose to enable.

src/native/common.gypi
```diff
@@ -18,10 +18,13 @@
                     '-fno-exceptions',
                 ],
                 'cflags': [
+                    '-Wno-cast-function-type',
```
guymguym, line 21: How many of these do we have? Maybe it would be best not to ignore and add explicit cast if needed.

liranmauda: Removing this flag we have:
12 `-Wimplicit-fallthrough=` all in `gf256.cpp`

2 `-Wsign-compare in `snappy.cc`

1 `-Wsign-compare in `utp_internal.cpp`

1 `-Werror=cast-function-type` in `node.h`

**another note:** is when we have the `'-Wno-cast-function-type',` flag we don't need the  `'-Werror'` in the `disabled_warnings` on  `common_third_party.gypi`

src/rpc/rpc_ws.js
```diff
@@ -57,7 +58,7 @@ class RpcWsConnection extends RpcBaseConnection {
             // may send fake http messages inside the websocket
             // in order to poison intermediate proxy caches.
             // reduces rpc throughput to ~70 MB/s
-            mask: false,
+            // mask: false,
```
nimrod-becker, line 61: ?

nimrod-becker (approved):

src/rpc/rpc_ws.js
liranmauda: we are enabling the ws mask now as the websockets is closing the connection if the mask is set.
https://github.com/websockets/ws/commit/4f293a8726092c75539287dd07358afaf151a2e5

guymguym: @liranmauda can you test the performance of rpc_benchmark over websocket with masking?
See the comment above the mask saying that without it we experienced a very low throughput of 70MB/sec.
you need to run one server and one client like this:
```
node src/rpc/rpc_benchmark.js --server --proto wss
node src/rpc/rpc_benchmark.js --client --proto wss
```

Here are the results I got with node 12 and older libs when toggling masking:
```
{ nodejs: 'v12.8.2', ws: '7.2.1', proto: 'wss', mask: false } => 500 MB/sec
{ nodejs: 'v12.8.2', ws: '7.2.1', proto: 'wss', mask: true } => 400 MB/sec
```

with the ws protocol (no TLS):
```
{ nodejs: 'v12.8.2', ws: '7.2.1', proto: 'ws', mask: false } => 650 MB/sec
{ nodejs: 'v12.8.2', ws: '7.2.1', proto: 'ws', mask: true } => 550 MB/sec
```

can you test these 4 cases too on node 14 and new ws?
seems like the effect is not as drastic as it was before.

guymguym: Two more points:
1. We can explore is if we can make mask:false work on the new we library version. We just failed to run it but it might be possible.
2. We can quite easily change our RPC to use the `tls:` protocol instead, in which we are doing the simple message framing on the socket, this is already being used for n2n connections so this is very stable. See [here](https://github.com/noobaa/noobaa-core/blob/50afd6ee7e161954aad2590dc66c2888432f6efc/src/rpc/rpc.js#L526-L530)

PR #6264, opened 2020-11-12
metrics for namespace buckets

romayalon:

### Explain the changes
1. Added metrics: number of namespace buckets, number of unhealthy namespace buckets, for each namespace bucket printed its name and health status.
An unhealthy bucket is a namespace bucket that its read resources (includes the write resource too) had more than 3 bucket related errors in the last 5 minutes.
2. Added issues report for namespace resource - which is an array of size 10 max of the most recent issues on namespace resource that are related to bucket problems.
3. Added calls for updating the namespace resource issues reports on failures in namespace_s3 and namespace blob - on read_object_md, put object, and put part.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -123,6 +123,11 @@ class NamespaceBlob {
         } catch (err) {
             this._translate_error_code(err);
             dbg.warn('NamespaceBlob.read_object_md:', inspect(err));
+            object_sdk.rpc_client.pool.update_issues_report({
+                namespace_resource_id: this.namespace_resource_id,
+                error_code: err.code,
+                time: Date.now(),
+            });
             throw err;
         }
     }
@@ -256,6 +261,11 @@ class NamespaceBlob {
             } catch (err) {
                 this._translate_error_code(err);
                 dbg.warn('NamespaceBlob.upload_object:', inspect(err));
+                object_sdk.rpc_client.pool.update_issues_report({
+                    namespace_resource_id: this.namespace_resource_id,
+                    error_code: err.code,
+                    time: Date.now(),
+                });
                 throw err;
             }
         }
@@ -273,7 +283,7 @@ class NamespaceBlob {
     // BLOCK BLOB UPLOADS //
     ////////////////////////
 
-    async upload_blob_block(params) {
+    async upload_blob_block(params, object_sdk) {
         dbg.log0('NamespaceBlob.upload_blob_block:',
             this.container,
             inspect(_.omit(params, 'source_stream'))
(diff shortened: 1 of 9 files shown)


src/server/system_services/pool_server.js
```diff
@@ -1216,6 +1217,30 @@ async function update_account_default_resource() {
...
+
+    const invalid_bucket_errors = ['ContainerNotFound', 'NotSignedUp', 'NoSuchBucket', 'InvalidSecurity',
+    'InvalidBucketState', 'InternalError', 'AccessDenied', 'InvalidAccessKeyId', 'AccountProblem'];
+
+    if (!_.includes(invalid_bucket_errors, error_code)) {
+        return;
+    }
+    const updates = {
+            _id: parse_object_id(namespace_resource_id),
+            $push: {
```
jeniawhite, line 1231: Need to check if PostgreSQL supports this.

src/sdk/namespace_blob.js
```diff
@@ -422,15 +432,24 @@ class NamespaceBlob {
...
+                        this.container,
+                        params.key,
+                        params.source_stream.pipe(count_stream),
+                        params.size, // streamLength really required ???
+                        callback)
+                );
+            } catch (err) {
+                object_sdk.rpc_client.pool.update_issues_report({
+                    namespace_resource_id: this.namespace_resource_id,
+                    error_code: err.rpc_code || err.code,
```
jeniawhite, line 448: Take the error code instead of rpc_code, because you do not check rpc_code errors in `update_issues_report`

src/server/system_services/stats_aggregator.js
```diff
@@ -628,6 +651,21 @@ function get_bucket_sizes_stats(req) {
     return ret;
 }
 
+async function calc_ns_bucket_health(namespace_info, bucket, {start_date, end_date}) {
```
jeniawhite, line 654: Why is this async?

jeniawhite (changes requested):

src/server/system_services/pool_server.js
dannyzaken: @romayalon It's better to modify the array using js and not mongo operators. similar to what's done in `nodes_monitor`
https://github.com/noobaa/noobaa-core/blob/4e026d9e52822f03aab31acc9e35ed6b0ee68607/src/server/node_services/nodes_monitor.js#L3497-L3511

src/sdk/namespace_blob.js
romayalon: correct. deleted

src/server/system_services/stats_aggregator.js
romayalon: thanks

src/server/system_services/pool_server.js
romayalon: np, changed to modification on js and not using mongo

PR #6266, opened 2020-11-15
Addition of PostgreSQL indexes

jeniawhite:
### Explain the changes
1. Added indexes and re-wrote declarations

1. Gap https://github.com/noobaa/noobaa-core/issues/6294

### Testing Instructions:
1.

### DIFF
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -326,6 +326,7 @@ class PostgresTable {
 
         if (!process.env.CORETEST) {
             // Run once a day
+            // TODO: Configure from PostgreSQL
             setInterval(this.vacuumAndAnalyze, 86400000, this);
         }
     }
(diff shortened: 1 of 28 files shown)


dannyzaken:
@jeniawhite @guymguym @nimrod-becker 

I think we should move the tables and indexes creation out of the stores constructor. this process is only necessary once for each new version of noobaa core, so I feel it should be part of upgrade\system creation. what do you think?

src/util/postgres_client.js
```diff
@@ -270,7 +271,39 @@ class PostgresTable {
...
             throw err;
         }
-        // TODO: create indexes
+
+        if (this.db_indexes) {
+            try {
+                await Promise.all(this.db_indexes.map(async index => {
+                    const { fields, options = {} } = index;
+                    try {
+                        const index_name = options.name || Object.keys(fields).join('_');
```
dannyzaken, line 280: what happens if a field was added after an index was already created (in a previous version)? maybe we should make the index name required?

src/util/postgres_client.js
```diff
@@ -270,7 +271,39 @@ class PostgresTable {
...
+                        dbg.log0(`creating index ${index_name} in table ${this.name}`);
+                        const col_arr = [];
+                        _.forIn(fields, (value, key) => {
+                            col_arr.push(`(data->>'${key}') ${value > 0 ? 'ASC' : 'DESC'}`);
+                        });
+                        const col_idx = `(${col_arr.join(',')})`;
+                        const uniq = options.unique ? 'UNIQUE' : '';
+                        const partial = options.partialFilterExpression ? `WHERE ${mongo_to_pg('data', options.partialFilterExpression)}` : '';
+                        const idx_str = `CREATE ${uniq} INDEX idx_btree_${this.name}_${index_name} ON ${this.name} USING BTREE ${col_idx} ${partial}`;
```
dannyzaken, line 289: shouldn't we check for index existence before creating?
maybe this is helpful (didn't check it myself) https://stackoverflow.com/questions/45983169/checking-for-existence-of-index-in-postgresql/45983490

guymguym (commented):
@jeniawhite Only had a quick look and have a question -
Why do we need those new indexes for the system store collections?
We are not querying those collections on demand, because we load the entire system store at once.
The only entity that might require it soon is to support millions of buckets, but for the rest of the entities do you see a need to query based on the index?

Perhaps this is mainly to ensure uniqueness in the database level? We never had much issues with uniqueness of these entities in mongo, that we know of.
But note that this PR adds new indexes that apply to both DB clients, and I wanted to highlight this point - can you please add it as another point in this PR description?

dannyzaken (approved):

PR #6269, opened 2020-11-18
Fix list on Blob

romayalon:

### Explain the changes
1. deleted '/' as default delimiter.
2. Added Prefix option.

1. Fixes #1898830

### Testing Instructions:
1.

### DIFF
--- a/src/endpoint/blob/blob_utils.js
+++ b/src/endpoint/blob/blob_utils.js
@@ -64,12 +64,12 @@ function parse_copy_source(req) {
 async function list_objects(params, account_name, container, sasToken) {
 
     const hostname = `${account_name}.blob.core.windows.net`;
-    const delimiter = params.delimiter || '/';
-    const maxResults = params.limit === 0 ? 1 : params.limit;
 
-    let path = `https://${account_name}.blob.core.windows.net/${container}?restype=container&comp=list&delimiter=${delimiter}` +
-        `&maxresults=${maxResults}&${sasToken}`;
+    let path = `https://${account_name}.blob.core.windows.net/${container}?restype=container&comp=list` +
+        `&maxresults=${params.limit}&${sasToken}`;
     if (params.key_marker) path += `&marker=${params.key_marker}`;
+    if (params.delimiter) path += `&delimiter=${params.delimiter}`;
+    if (params.prefix) path += `&prefix=${params.prefix}`;
 
     let response;
     try {
@@ -80,15 +80,24 @@ async function list_objects(params, account_name, container, sasToken) {
     const status_code = response.statusCode;
     const buffer = await read_stream_join(response);
     const body = buffer.toString('utf8');
-    if (status_code !== 200) {
-        throw new Error(`Could not get blobs and diresctories list, (status code: ${status_code}) got ${body}`);
-    }
 
     let blobs;
     let dirs;
     let next_marker;
     try {
         const parsed = await parse_xml_to_js(body);
+        if (status_code !== 200) {
+            if (parsed.Error) {
+                const code = parsed.Error.Code && parsed.Error.Code[0];
+                const faulty_query_param = parsed.Error.QueryParameterName &&
+                    parsed.Error.QueryParameterName[0];
+                if (code === 'OutOfRangeQueryParameterValue' &&
+                    faulty_query_param === 'maxresults' && params.limit === 0) {
+                        return { blobs: [], dirs: [], next_marker: '' };
+                    }
+            }
+            throw new Error(`Could not get blobs and diresctories list, (status code: ${status_code}) got ${body}`);
+        }
         blobs = parsed.EnumerationResults.Blobs[0].Blob;
         dirs = parsed.EnumerationResults.Blobs[0].BlobPrefix;
         next_marker = parsed.EnumerationResults.NextMarker[0];


src/endpoint/blob/blob_utils.js
```diff
@@ -64,12 +64,13 @@ function parse_copy_source(req) {
 async function list_objects(params, account_name, container, sasToken) {
 
     const hostname = `${account_name}.blob.core.windows.net`;
-    const delimiter = params.delimiter || '/';
     const maxResults = params.limit === 0 ? 1 : params.limit;
 
-    let path = `https://${account_name}.blob.core.windows.net/${container}?restype=container&comp=list&delimiter=${delimiter}` +
+    let path = `https://${account_name}.blob.core.windows.net/${container}?restype=container&comp=list` +
         `&maxresults=${maxResults}&${sasToken}`;
```
jeniawhite, line 70: Maybe remove maxresults as well? 
It is optional and has a default of 5000.
I think that it is better to handle it like all others below.

src/endpoint/blob/blob_utils.js
```diff
@@ -64,12 +64,13 @@ function parse_copy_source(req) {
 async function list_objects(params, account_name, container, sasToken) {
 
     const hostname = `${account_name}.blob.core.windows.net`;
-    const delimiter = params.delimiter || '/';
     const maxResults = params.limit === 0 ? 1 : params.limit;
```
jeniawhite, line 67: This is incorrect.
First of all the default is 5000 and not 1.
Secondly, setting maxresults to a value less than or equal to zero results in error response code 400 (Bad Request).
It is a bug to check for zero and assign a 1.

jeniawhite (changes requested):

src/endpoint/blob/blob_utils.js
romayalon: it's not a part of my changes but ok, 
If i'm not mistaken it's not a check for default and the reason behind it it's for handling maxResults=0 is allowed on AWS and in Blob ```Setting maxresults to a value less than or equal to zero results in error response code 400 (Bad Request).```

src/endpoint/blob/blob_utils.js
romayalon: maxResults get a value of 1000 in s3_get_bucket.js for consistency with AWS, I think we should keep it for not getting 5000 records and to get 1000 records as default instead.

src/endpoint/blob/blob_utils.js
jeniawhite: In any case, this isn't compatible.
`(base) jeniawhite@MacBook-Pro noobaa-core-code % aws s3api list-objects --bucket jenia --max-keys 0
{
    "IsTruncated": false,
    "Marker": "",
    "Name": "jenia",
    "Prefix": "",
    "MaxKeys": 0,
    "EncodingType": "url"
}
(base) jeniawhite@MacBook-Pro noobaa-core-code % aws s3api list-objects --bucket jenia --max-keys 1
{
    "IsTruncated": true,
    "Marker": "",
    "Contents": [
        {
            "Key": "file_with_data",
            "LastModified": "2020-10-04T15:47:59+00:00",
            "ETag": "\"2e237c53edcc537f2e12ad1b0e613b0e\"",
            "Size": 4692,
            "StorageClass": "STANDARD",
            "Owner": {
                "DisplayName": "nbecker",
                "ID": "6ad7c630ff458fbcf3fc06c2625a64359f33904bf84c4e6531a3a59d94557d08"
            }
        }
    ],
    "Name": "jenia",
    "Prefix": "",
    "MaxKeys": 1,
    "EncodingType": "url"
}
`
While in the list of blob we would get a key.
I know that this is not a part of your change, but this seems like a bug.
We can open an issue and handle it later if you'd like.

romayalon: of course, it's not compatible to list 1 result instead of 0 but it's that way for not returning Bad Request error.
I can check for maxResults ==0 before like we do today and return empty values like AWS does instead of checking maxResults to 1. WDYT?

src/endpoint/blob/blob_utils.js
jeniawhite: It is also a confusing factor for someone that wants to use the Azure NS only and sends requests for 5000 blobs but gets up to 1000.

romayalon: so should we be compatible with AWS or Azure?
return by Default 1000 records or 5000?

src/endpoint/blob/blob_utils.js
jeniawhite: In my opinion, it is better not to return values on 0 (even though Azure doesn't allow such requests to be sent), rather than returning values that the user did not request.
The second option is just to throw an exception from Azure NS and our NS merge will handle that.
If we'd have a merge between AWS and Azure then we will get a partial response from AWS, but if we will have a single Azure NS then we will throw an exception.
There are other variants as well, but I think these are the main two.

romayalon: @guymguym 
what was the reason to change the limit to 1 in the first place instead of just returning empty values?
We decided to go with the first option instead if you don't have counterclaims

jeniawhite (approved):

PR #6274, opened 2020-11-22
Fail frontend test if FE should have build

liranmauda:
### Explain the changes
Fail frontend test if there is a change in npm-module that require FE built, and Frontend was not built or pushed.

### DIFF
--- a/src/deploy/NVA_build/FrontendLib.Dockerfile
+++ b/src/deploy/NVA_build/FrontendLib.Dockerfile
@@ -30,3 +30,7 @@ COPY ./frontend/ ./frontend/
 RUN echo '{ "allow_root": true }' > /root/.bowerrc
 RUN cd frontend && \
     npm run build-lib
+
+COPY ./frontend/src/lib/lib* /tmp/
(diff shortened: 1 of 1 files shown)


src/deploy/NVA_build/FrontendLib.Dockerfile
```diff
@@ -30,3 +30,7 @@ COPY ./frontend/ ./frontend/
 RUN echo '{ "allow_root": true }' > /root/.bowerrc
 RUN cd frontend && \
     npm run build-lib
+
+COPY ./frontend/src/lib/lib* /tmp/
+RUN diff ./frontend/src/lib/lib.js /tmp/
+RUN diff ./frontend/src/lib/lib.js.map /tmp/
```
nb-ohad, line 36: I don't think this is a good indication because bower modules are not locked.
As is it is possible that the current `build-lib` will produce a lib that is slightly diff than the one stored in GitHub as packages have their Z version updated frequently. 
I am not aware of a method to lock bower dependency to a specific version (including the Z part)

PR #6275, opened 2020-11-22
Mongo - Postgres migration - core side

jackyalbo:

### Explain the changes
1. 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/test/unit_tests/test_postgres_upgrade.js
+++ b/src/test/unit_tests/test_postgres_upgrade.js
@@ -1,0 +1,278 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+// setup coretest first to prepare the env
+// const coretest = require('./coretest');
+const mocha = require('mocha');
+const { Migrator } = require('../../upgrade/migrator');
+const assert = require('assert');
+const EventEmitter = require('events').EventEmitter;
+const _ = require('lodash');
+const postgres_client = require('../../util/postgres_client');
+const wtf = require('wtfnode');
+
+
+// coretest.setup({ pools_to_create: [], db_name: 'upgradetest' });
+process.env.CORETEST = 'upgrade-test';
+
+const default_schema = {
+    id: 'default_schema',
+    type: 'object',
+    required: [
+        '_id',
+        'key1',
+        'key2',
+        'key3',
+    ],
+    properties: {
+        _id: { objectid: true },
+        key1: { type: 'string' },
+        key2: { type: 'boolean' },
+        key3: {
+            type: 'array',
+            items: {
+                type: 'string',
+                enum: ['val1', 'val2', 'val3', 'val4']
+            }
+        },
+        key4: {
+            type: 'object',
+            properties: {
+                key1: {
+                    type: 'number',
+                },
+                key2: {
+                    type: 'string',
+                },
+                when: {
+                    date: true
+                }
+            }
+        },
+        epoch: {
+            idate: true
+        }
+    }
+};
+
+class TestCollection {
+
+    /**
+     * @param {TestCollection} col
+     * @returns {nb.DBCollection}
+     */
+    static implements_interface(col) { return col; }
+
+    constructor(col, size) {
+        TestCollection.implements_interface(this);
+        this.schema = col.schema;
+        this.name = col.name;
+        this.db_indexes = col.db_indexes;
+        this.size = size;
+        this.data = _.times(this.size, i => ({ _id: i + 1,
+            key1: 'value1-' + i,
+            key2: Boolean(i % 2),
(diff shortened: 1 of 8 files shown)


dannyzaken (commented):
1. maybe separate the general migration script to another file. it will be easier to test and reuse. you can then just pass any collections list to the generic migration script regardless of which DBs are involved.

2. for the source DB (mongo) don't pass the DB indexes - you don't want to cause it to start building indexes right before the migration.

4. we should test if it's worth it to migrate several collections in parallel. 
5. before starting to migrate a collection drop the collection in the `to_client` to get rid of failed attempts.
6. store in the DB the status for each collection, so you can skip them on retries.
7. for the large collections (chunks, blocks, etc.) we should consider storing retry markers so we won't need to restart the entire collection on failures.
8. in `migrate_collection` catch errors and retry a few times before failing the entire migration
9. check if the query need to be sorted

jackyalbo:
@dannyzaken I addressed some of your requests and added a basic test. Still WIP but want your feedback (and to check test is passing). Thanks.

src/test/unit_tests/test_postgres_upgrade.js
```diff
@@ -0,0 +1,250 @@
...
+    type: 'object',
+    required: [
+        '_id',
+        'key1',
+        'key2',
+        'key3',
+    ],
+    properties: {
+        _id: { objectid: true },
+        key1: { type: 'string' },
```
dannyzaken, line 30: maybe use more types in the test other than string

src/test/unit_tests/test_postgres_upgrade.js
```diff
@@ -0,0 +1,250 @@
...
+        this.name = col.name;
+        this.db_indexes = col.db_indexes;
+        this.size = size;
+        this.data = _.times(this.size, i => ({ _id: i + 1, key1: 'value1-' + i, key2: 'value2-' + i, key3: 'value3-' + i}));
+    }
+
+    async find(query, options = {}) {
+        const start = (query && query._id && query._id.$gt) || 0;
+        const end = options.limit ? start + options.limit : this.size;
+        console.log("data to be moved", this.data.slice(start, end), query);
```
dannyzaken, line 56: won't it make the test very chatty? I think it's better to print only on failures

src/upgrade/migrator.js
```diff
@@ -0,0 +1,88 @@
...
+        for (const collection of this.collection_list) {
+            await this.from_client.define_collection(_.omit(collection, 'db_indexes'));
+            await this.to_client.define_collection(collection);
+        }
+        await this.from_client.connect(true);
+        await this.to_client.connect();
+        this.upgrade_table = this.to_client.collection('migrate_status');
+        this.migrate_status = await this.upgrade_table.findOne({});
+        if (!this.migrate_status) {
+            this.migrate_status = {
```
dannyzaken, line 40: add a schema to the `migrate_status` collection. postgres_client uses the schema to encode\decode the json

src/upgrade/migrator.js
```diff
@@ -0,0 +1,88 @@
...
+                _id: this.to_client.new_object_id(),
+                collection_index: 0,
```
dannyzaken, line 42: add more info to help with debugging - collection_name, write_date, total_size, num_inserted, etc.

src/upgrade/migrator.js
```diff
@@ -0,0 +1,88 @@
...
+                }
+                total += docs.length;
+                console.log(`migrated ${total} documents to table ${col_name}`);
+                marker = docs[docs.length - 1]._id;
+                await this.upgrade_table.updateOne({}, { $set: { marker }});
+            } else {
+                done = true;
+            }
+        }
+        console.log(`completed migration of ${total} documents to table ${col_name}`);
```
dannyzaken, line 84: as a sanity check, maybe check that the sizes of the collections are the same

src/test/unit_tests/test_postgres_upgrade.js
jackyalbo: will do

src/test/unit_tests/test_postgres_upgrade.js
jackyalbo: was for debugging - removed

src/upgrade/migrator.js
jackyalbo: done

src/upgrade/migrator.js
jackyalbo: done

src/upgrade/migrator.js
jackyalbo: done

dannyzaken (approved):

PR #6280, opened 2020-11-24
Prepare to use a Docker Hub mirror in the CentOS CI

nixpanic:
### Explain the changes
1. Use registry.centos.org for pulling container images. This reduces the
  load on Docker Hub so that the rate limit for image pulls is not hit as
  quickly anymore.
1. Docker Hub has added rate limits for container image pulls. This can
  cause problems when pulling images within the CentOS CI. By deploying a
  local mirror, images can be pulled without hitting the limits that
  quickly.

1. this prepares to make the CentOS CI less dependent on docker.io

### DIFF
--- a/src/deploy/NVA_build/builder.Dockerfile
+++ b/src/deploy/NVA_build/builder.Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:8 
+FROM registry.centos.org/centos:8
 LABEL maintainer="Liran Mauda (lmauda@redhat.com)"
 
 ##############################################################
--- a/src/deploy/NVA_build/dev.Dockerfile
+++ b/src/deploy/NVA_build/dev.Dockerfile
@@ -1,5 +1,5 @@
 # dev.Dockerfile is meant to be used manually for developer testing
-FROM centos:8 
+FROM registry.centos.org/centos:8
 
 ENV container docker
 
--- a/.jenkins/deploy/Dockerfile
+++ b/.jenkins/deploy/Dockerfile
@@ -1,4 +1,4 @@
-FROM centos:latest
+FROM registry.centos.org/centos:latest
 
 RUN true \
  && yum -y install git make python3-pip \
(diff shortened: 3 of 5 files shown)


nixpanic:
@liranmauda this only prepares things, manual deployment in the CentOS OpenShift project is needed (after editing the `.yaml` file with usernames/passwords/tokens).

src/deploy/NVA_build/builder.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:8 
+FROM registry.centos.org/centos:8
```
liranmauda, line 1: Not a CI image, We need to keep it `centos:8`

src/deploy/NVA_build/dev.Dockerfile
```diff
@@ -1,5 +1,5 @@
 # dev.Dockerfile is meant to be used manually for developer testing
-FROM centos:8 
+FROM registry.centos.org/centos:8
```
liranmauda, line 2: Not a CI image, We need to keep it `centos:8`

liranmauda (changes requested):

liranmauda:
We should handle the Dockerfiles replacement here:
https://github.com/noobaa/noobaa-core/blob/6bcadfb0d8830a890973c1b5dd439c98833722d3/.jenkins/prepare.sh#L98-L99 

We will build the base and builder and push/pull them from the registry.centos.org registy

src/deploy/NVA_build/builder.Dockerfile
nixpanic: Well, this way it is clear where the image comes from, it is a recommended practise to include the registry. A default search for pulling the image may contain something else than docker.io and provide some (old/fake/..) centos image. So, IMHO it is better to pull the image from the official registry instead.

nixpanic:
> We will build the base and builder and push/pull them from the registry.centos.org registry

The registry.centos.org contains only images from the CentOS project, and is not the same as the mirror-registry that gets deployed with the `.jenkins/deploy/docker-mirror.yaml` file. The Docker Hub mirror in the CI will be reachable through mirror-noobaa.apps.ocp.ci.centos.org.

liranmauda (approved):
LGTM

liranmauda:
Currently, we are not using Centos Sig Jenkins anymore due to https://pagure.io/centos-infra/issue/159 
We will close this PR and revive it if neccacery

PR #6281, opened 2020-11-24
Namespace bucketclass and NamespaceStore CRDs - core side

romayalon:

### Explain the changes
1. Deleted the condition to pass deletion if it's namespace bucket. needed for Namespace bucketclass and Namespacestore crds.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -754,10 +754,6 @@ function check_for_lambda_permission_issue(req, bucket, removed_accounts) {
 async function delete_bucket_and_objects(req) {
     var bucket = find_bucket(req);
 
-    // fail on namespace bucket
-    if (bucket.namespace) {
-        throw new RpcError('BAD_REQUEST', 'cannot perform delete_bucket_and_objects on namespace bucket');
-    }
     const now = new Date();
     // mark the bucket as deleting. it will be excluded from system_store indexes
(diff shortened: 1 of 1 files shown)


src/server/system_services/bucket_server.js
```diff
@@ -754,10 +754,6 @@ function check_for_lambda_permission_issue(req, bucket, removed_accounts) {
 async function delete_bucket_and_objects(req) {
     var bucket = find_bucket(req);
 
-    // fail on namespace bucket
```
dannyzaken, line 757: `delete_bucket_and_objects ` is called from the OBC delete flow. I feel that in the case of namespace it is not well defined.
what will happen in different namespace types?

romayalon: correct me if I missed something - bucket reclaimer does:
1. find deleting buckets
2. search in MD_store for objects of the deleting bucket and deleted them.
3. search for the bucket's tiers and tiering policy 
4. remove the bucket from the allowed buckets permission lists of accounts  

if i'm not mistaken the only different step is 2 and I can skip it if the bucket is namespace bucket...
WDYT?

dannyzaken: @romayalon that sounds right. just make sure to test this flow

jeniawhite (approved):

dannyzaken (approved):

PR #6295, opened 2020-11-26
Refactoring insertMany

jeniawhite:
### Explain the changes
1. Rewriting the interface and PostgreSQL optimization for the query

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -451,21 +451,15 @@ class PostgresTable {
         return {};
     }
 
-    async insertMany(data, options) {
+    // This is done mainly to be quick
+    // Notice that the behaviour between MongoDB and PostgreSQL differs
+    // In PostgreSQL we either push everything at once or do not push anything at all
+    // In MongoDB we might push partially (succeed pushing several documents and fail on others), and it is done in parallel
+    async insertManyUnordered(data) {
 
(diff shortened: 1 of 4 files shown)


src/util/postgres_client.js
```diff
@@ -451,21 +451,22 @@ class PostgresTable {
...
-        let queries = [];
-        const ordered = _.isUndefined(options.ordered) ? true : options.ordered;
+    // This is done mainly to be quick
+    // Notice that the behaviour between MongoDB and PostgreSQL differs
+    // In PostgreSQL we either push everything at once or do not push anything at all
+    // In MongoDB we might push partially (succeed pushing several documents and fail on others), and it is done in parallel
+    async insertManyUnordered(data) {
+
+        const args = [];
+        let values_str = '';
```
dannyzaken, line 461: you can get `args` and `values_str` without the for loop:

```suggestion
        const args = _.flatten(data.map(doc => [doc._id, encode_json(doc)]));
        let values_str = _.times(data.length, i => `($${i*2+1} , $${i*2+2} )`).join(', ');
```

dannyzaken: your way seems to work too, but it was a bit hard to follow and read

dannyzaken (approved):

PR #6296, opened 2020-11-29
Upgrade master keys implementation

romayalon:

### Explain the changes
1. Added upgrade script for upgrading master keys.
2. In the upgrade function created master keys for each existing entity in the system.
system master key is inserted to the DB encrypted by the root master key.
accounts/buckets master keys inserted to the DB encrypted by the system master key.
3. specific actions of entities: 
a. if the entity is an account - encrypted its secrets in access keys and sync credentials by its master key, and encrypted the secrets of the associated pools and namespace resources of the account.
b. if the entity is a bucket - added master key for the bucket (the map builder BG worker will encrypt the chunks cipher keys in the bucket and will attach them the bucket's master key.
 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/master_key_manager.js
+++ b/src/server/system_services/master_key_manager.js
@@ -215,9 +215,9 @@ class MasterKeysManager {
         let updated_value = cipher.update(Buffer.from(value.unwrap()));
         if (m_key.cipher_type !== 'aes-256-gcm') updated_value = Buffer.concat([updated_value, cipher.final()]);
 
-        const ciphered_value = new SensitiveString(updated_value.toString('base64'));
+        const ciphered_value = updated_value.toString('base64');
         this.secret_keys_cache.put_in_cache(ciphered_value, value);
-        return ciphered_value;
+        return new SensitiveString(ciphered_value);
     }
 
     is_m_key_disabled(id) {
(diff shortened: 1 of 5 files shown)


src/server/system_services/system_server.js
```diff
@@ -1550,6 +1550,97 @@ function get_entity_info(entity, entity_type) {
...
+    }
+    // upgrade buckets master keys
+    _.map(system_store.data.buckets, bucket => {
+        if (bucket.master_key_id) return;
+        const bucket_m_key = system_store.master_key_manager.new_master_key({
+            description: `master key of ${bucket._id} bucket`,
+            master_key_id: system_master_key._id,
+            cipher_type: system_master_key.cipher_type
+        });
+        buckets_updates.push({
```
jeniawhite, line 1577: Do we want to encrypt all of the buckets and their chunks upon upgrade?
I remember that we've discussed not re-encrypting data (in that case it is the chunk's cipher), but the question still remains.

src/server/system_services/system_server.js
```diff
@@ -1550,6 +1550,97 @@ function get_entity_info(entity, entity_type) {
...
+     _.map(system_store.data.accounts, account => {
+        if (account.master_key_id || account.email.unwrap() === 'support@noobaa.com') return;
+        const account_m_key = system_store.master_key_manager.new_master_key({
+            description: `master key of ${account._id} account`,
+            master_key_id: system_master_key._id,
+            cipher_type: system_master_key.cipher_type
+        });
+
+        // encrypt access
+        const encryptes_access_keys = _.map(account.access_keys, acc => {
```
jeniawhite, line 1596: encrypted_access_keys I assume

src/server/system_services/system_server.js
romayalon: yes, after discussion we decided to keep this flow.

src/server/system_services/master_key_manager.js
```diff
@@ -245,10 +245,9 @@ class MasterKeysManager {
                 if (this.is_m_key_disabled(account.master_key_id._id)) {
                     continue;
                 }
-                const decipher = crypto.createDecipheriv(m_key.cipher_type, m_key.cipher_key, m_key.cipher_iv);
```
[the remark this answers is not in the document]
romayalon: keeping it outside doesn't work, since it's just an optimization, let's keep it that way for now, and i'll optimize it soon

PR #6319, opened 2020-12-10
SCC fixes

jackyalbo:

### Explain the changes
1. adding postgres init option to noobaa_init
2. change all permissions setting to recursive

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/deploy/NVA_build/noobaa_init.sh
+++ b/src/deploy/NVA_build/noobaa_init.sh
@@ -151,14 +151,38 @@ prepare_server_pvs() {
 }
 
 prepare_mongo_pv() {
-  local dir="/mongo_data/mongo/cluster/shard1"
+  local main_dir="/mongo_data"
+  local shard_dir="/mongo_data/mongo/cluster/shard1"
+  local recursive_file="/mongo_data/recursive_file"
+  local dir_id=$(stat -c '%u' ${main_dir})
+  local current_id=$(id -u)
 
   # change ownership and permissions of mongo db path
-  ${KUBE_PV_CHOWN} mongo
+  if [ "${dir_id}" != "${current_id}" ] || [ ! -f ${recursive_file} ]; then
+    echo "uid change has been identified - will change from uid: ${dir_id} to new uid: ${current_id}"
+    time ${KUBE_PV_CHOWN} mongo ${current_id}
+    touch ${recursive_file}
+  fi
 
-  mkdir -p ${dir}
-  chgrp 0 ${dir}
-  chmod g=u ${dir}
+  if [ ! -d ${shard_dir} ]; then
+    echo "creating shard directory: ${shard_dir}" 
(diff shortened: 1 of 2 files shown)


src/deploy/NVA_build/noobaa_init.sh
```diff
@@ -151,16 +151,37 @@ prepare_server_pvs() {
...
-  chmod g=u ${dir}
 }
 
+prepare_postgres_pv() {
+  local dir="/var/lib/pgsql"
+  local uid=$(stat -c '%g' ${dir})  
+  local gid=$(stat -c '%u' ${dir})
+
+  # change ownership and permissions of mongo db path
+  if [ "${uid}" != "10001" ] || [ "${gid}" != "0" ]
```
dannyzaken, line 177: we are still not resilient to uid change, we just assume that the uid:gid will be 10001:0.
I think we should maybe pass the EXPECTED_UID and EXPECTED_GID as envs. this will give us more flexibility if a uid is changed

src/native/tools/kube_pv_chown.cpp
```diff
@@ -49,13 +74,12 @@ main(int argc, char* argv[])
...
 
-    if (deployment_type == "server") {
-        change_path_permissions("/log", uid);
-    } else if (deployment_type == "mongo") {
-        change_path_permissions("/mongo_data", uid);
-        change_path_permissions("/log", uid);
+    if (deployment_type == "mongo") {
+        change_path_permissions_recursive("/mongo_data", uid);
+    } else if (deployment_type == "postgres") {
+        change_path_permissions_recursive("/var/lib/pgsql", uid);
```
dannyzaken, line 80: if we are using an ENV to pass the EXPECTED_UID and EXPECTED_GID then use it here as well

src/deploy/NVA_build/noobaa_init.sh
```diff
@@ -151,16 +151,40 @@ prepare_server_pvs() {
...
-  chmod g=u ${dir}
+prepare_postgres_pv() {
+  local dir="/var/lib/pgsql"
+  local dir_id=$(stat -c '%u' ${dir})    
+  local current_id=$(id -u)
+
+  # change ownership and permissions of mongo db path
+  if [ "${dir_id}" != "${current_id}" ]
+  then
+    ${KUBE_PV_CHOWN} postgres ${current_id}
```
dannyzaken, line 183: why didn't use `recursive_file` here too?

jackyalbo: just because we shouldn't have remains of none recursive runs in postgres. the recursive_file was used for cases of an old none-recursive run that updated the main directory but didn't update the rest of the files in the directory.

dannyzaken (approved):
Just the one comment regarding the `recursive_file` for postgres. add it if needed

PR #6324, opened 2020-12-14
Added api for get namespace_resource secrets

romayalon:

### Explain the changes
1. Added api for get namespace_resource secrets for upgrade namespace in operator.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/pool_api.js
+++ b/src/api/pool_api.js
@@ -98,7 +98,19 @@ module.exports = {
                     },
                     target_bucket: {
                         type: 'string',
-                    }
+                    },
+                    namespace_store: {
+                        type: 'object',
+                        required: ['name', 'namespace'],
+                        properties: {
+                            name: {
+                                type: 'string',
+                            },
+                            namespace: {
+                                type: 'string',
+                            },
+                        }
+                    },
                 }
             },
             auth: {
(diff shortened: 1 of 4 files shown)


src/server/system_services/pool_server.js
```diff
@@ -1010,6 +1013,49 @@ function get_namespace_resource_info(namespace_resource) {
...
+    const info = _.omitBy({
+        access_key: namespace_resource.connection.access_key,
+        secret_key: namespace_resource.connection.secret_key,
+        need_k8s_sync: namespace_resource.namespace_store.need_k8s_sync
+    }, _.isUndefined);
+
+    return info;
+}
+
+function unset_namespace_store_sync(req) {
```
dannyzaken, line 1033: shouldn't this be called from `create_namespace_resource`?

src/api/pool_api.js
```diff
@@ -427,7 +439,54 @@ module.exports = {
...
+                        type: 'boolean'
+                    }
+                }
+            },
+            auth: {
+                system: 'admin'
+            }
+        },
+
+        unset_namespace_store_sync: {
```
nimrod-becker, line 471: maybe better called sync_namespace_store ? WDYT ?

src/server/system_services/pool_server.js
```diff
@@ -1010,6 +1013,49 @@ function get_namespace_resource_info(namespace_resource) {
...
+    if (req.auth.role !== 'operator') {
+        throw new RpcError('UNAUTHORIZED', 'Must be operator');
+    }
+    const namespace_resource = find_namespace_resource_by_name(req);
+    if (!namespace_resource.namespace_store ||
+        !namespace_resource.namespace_store.need_k8s_sync) {
+        dbg.log0(`no need to create namespace resource.`);
```
nimrod-becker, line 1040:
```suggestion
        dbg.log0(`no need to sync namespace resource.`);
```
?

src/server/system_services/pool_server.js
romayalon: we are not calling ```create_namespace_resource``` from the operator side since we already know it exist,
I had to add this api because I saw we eventually are not calling it..

src/api/pool_api.js
romayalon: this function puts the namespacestore details on the namespace resource and turns off the sync flag, not sure sync_namespace_store name describes it the best..

nimrod-becker: right, something like set_namespace_store_info ?
Since it does more than just the unset ?

romayalon: sure, updated.

dannyzaken (approved):

Issue #6265, opened 2020-11-13
S3 and WebUI not responsive after heavy load by NoobaaFunctions


QuintinBecker:
### Environment info
- NooBaa Version: master+30ee3535b5 , built in house with additional dependencies  
- Platform: OKD 4.5

### Actual behavior
WebUI is not working after a heavy load was produced due to Noobaa-Function calls. 
The limit seems to be somewhere around 5k Events handled. Till this border is hit the WebUI is responsive and "Function Statistics" are displayed correctly. After hitting this "limit" WebUI does not display any information.

Durig this situation, no S3 connections to noobaa are possible.

After some time, like 10 min, the WebUI is reachable but the "Function Statistics" is not displaying anything.

### Expected behavior
There should be no limit for "Function Statistics" to be displayed.

### Steps to reproduce
1. Write a minimal handler function and call it for more than 5k times within the displayed time (60 min). 

### More information - Screenshots / Logs / Other output
We have a static Endpoint-Count of five and have limits set appropriate to the load caused by the function-calls (limit is 4 core 8Gi). 

Noobaa-Core is running with limits of 1 Core 4 Gi. 

Load Scenario during Upload of files:
![image](https://user-images.githubusercontent.com/64270807/99071485-80307d80-25b2-11eb-80dd-c90372538bae.png)

WebUI with last update of Statistics function just befor it crashes:
![image](https://user-images.githubusercontent.com/64270807/99071568-ae15c200-25b2-11eb-9a70-0c2fdb1ce27f.png)

WebUI is no longer updated dynamically and after reload of the page we get:
![image](https://user-images.githubusercontent.com/64270807/99072050-912dbe80-25b3-11eb-97ad-c9b4f584425a.png)

There is no heavy load on the pods during the this absence of the WebUI. This is the load while it is not working and no load is on the Endpoints:
![image](https://user-images.githubusercontent.com/64270807/99072314-16b16e80-25b4-11eb-9672-0d9bd30e0ac8.png)

This situation is present for at least 10 minutes before the WebUI is working properly, but the "Function Statistics" is not displayed.

It would be great if you could give me some target to look for if we reach this situation.

QuintinBecker:
We are hitting this issue more and more often and it is becoming a growing problem right now. It would be awsome to get some hints where to go on digging for the root cause of this issue. 

We just noticed some strange Logs in the Core-Node:

Some Error in ´´read_usage_gridfs´´:

`
[...]

PR #6367, opened 2021-02-02
Add Retry To Lambda Functions and Add the Error to the DB

### LINKED ISSUE #6349: Noobaa Functions - Logging of failed invocations and retry them
When a function call fails within the Nooba Endpoint, there is no mechanism available to retry the function-call.

It would be awsom if there would be an integration of a retry mechanism. 

We thought about a logging mechnism in the first place to get all failed invocations from noobaa with some info of the event causing the fail. 
If possible it would be great to retry the function call a second time or n-times if it can be configured by the user.

liranmauda:
### Explain the changes
Add retry to lambda functions:
- Added retries (default is 2)
- Added UI Field in the add/edit retries
- async/await some flows

### Issues: 
~1. Edit form is not loading the current number of attempts~
~2. There is no tooltip that explains the attempts field~ 

### out of scope
~1. Writing a failed invoke log into the DB~
2. converting the lambda function and save it in the DB

### Testing Instructions:
For the retries: 
1. Add/Edit a trigger With 1 or more retires
2. See that on "fail" it retries
3. See that on "success" It does not retry
4. Check that if 0 retries are set, then the trigger runs once.

For the failed error in the db:
1. run a failed function
2. see in the db that the `error_msg` is there:
    run `db.func_stats.find({'error':true}).pretty()`
```
{
	"_id" : ObjectId("601ab5ac0d4f4c000d395964"),
	"system" : ObjectId("60196a75933e03002de1f296"),
	"func" : ObjectId("60196c60933e03002de1f2b7"),
	"time" : ISODate("2021-02-03T14:39:18.870Z"),
	"took" : 21597,
	"error" : true,
	"error_msg" : "Func handler not a function main.no"
}
```

### DIFF
--- a/src/server/system_services/schemas/bucket_schema.js
+++ b/src/server/system_services/schemas/bucket_schema.js
@@ -319,6 +319,9 @@ module.exports = {
                     object_suffix: {
                         type: 'string'
                     },
+                    attempts: {
+                        type: 'integer'
+                    },
                 }
             }
         },
--- a/frontend/src/app/components/modals/add-bucket-trigger-modal/add-bucket-trigger-modal.html
+++ b/frontend/src/app/components/modals/add-bucket-trigger-modal/add-bucket-trigger-modal.html
@@ -69,6 +69,18 @@
             <p class="hint">optional</p>
         </editor>
 
+        <editor params=
+            "label: 'Retries', 
+            tooltip: retriesTooltip
+        ">
+            <input type="number"
+                min="0"
+                max="99"
+                ko.value="$form.attempts"
+                ko.validationCss="$form.attempts"
+            />
+        </editor>
+
         <editor params="label: 'Trigger Active'">
             <toggle-switch params="value: $form.active"></toggle-switch>
         </editor>
--- a/frontend/src/app/components/modals/add-bucket-trigger-modal/add-bucket-trigger-modal.js
+++ b/frontend/src/app/components/modals/add-bucket-trigger-modal/add-bucket-trigger-modal.js
@@ -11,6 +11,8 @@ import {
     closeModal
 } from 'action-creators';
 
+const retriesTooltip = 'Number of retries Noobaa will make if the trigger fails';
+
 function _mapDataBucketOption(bucket) {
     return {
         value: bucket.name,
@@ -26,6 +28,7 @@ function _mapNamespaceBucketOption(bucket) {
 }
 
 class AddBucketTriggerModalViewModel extends ConnectableViewModel {
+    retriesTooltip = retriesTooltip;
     formName = this.constructor.name;
     bucketName = '';
     existingTriggers = null;
@@ -81,6 +84,7 @@ class AddBucketTriggerModalViewModel extends ConnectableViewModel {
                 event: '',
                 prefix: '',
                 suffix: '',
+                attempts: 0,
                 active: true
             } : undefined
         });
@@ -88,7 +92,7 @@ class AddBucketTriggerModalViewModel extends ConnectableViewModel {
 
     onValidate(values) {
         const errors = {};
-        const { event, func } = values;
+        const { event, func, attempts } = values;
 
         if (!func) {
             errors.func = 'Please select a function';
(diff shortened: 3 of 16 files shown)


liranmauda:
@QuintinBecker @lallinger-arbeit

lallinger-arbeit:
@liranmauda Thank you very much for such a fast implementation! If it is in a state where we should test it just hit me up!

liranmauda:
> @liranmauda Thank you very much for such a fast implementation! If it is in a state where we should test it just hit me up!

@lallinger-arbeit It is in a state to test it.
I will later add the "out of scope"  parts (maybe in another PR).

lallinger-arbeit:
> 
> 
> > @liranmauda Thank you very much for such a fast implementation! If it is in a state where we should test it just hit me up!
> 
> @lallinger-arbeit It is in a state to test it.
> I will later add the "out of scope" parts (maybe in another PR).

ok, then we will try to test it this week!

liranmauda:
Added the failed invoke error message into the func_stats collection.

lallinger-arbeit:
@liranmauda looks very good so far. I just noticed that in the UI the attempts field is always displayed empty, but it should show the configured attempts.
Also as the information of failed functions is saved in the `func_stats` collection, it would be best the problem with the [slow mapreduce job of the MongoDB](https://github.com/noobaa/noobaa-core/issues/6265) on this collection would be fixed too. Because currently we disabled the code which writes to the `func_stats`, but would definitely need it for this retry feature

lallinger-arbeit:
@liranmauda it would also be nice to have a message in `func_stats` which states if a event exceeded its maximum retries

liranmauda:
> @liranmauda looks very good so far. I just noticed that in the UI the attempts field is always displayed empty, but it should show the configured attempts.

Yes, I am aware of this issue, waiting for @nb-ohad Input.

liranmauda:
Added tooltip.
Not sure it is very helpful, any suggestion for the text will be welcome.

liranmauda:
> > @liranmauda looks very good so far. I just noticed that in the UI the attempts field is always displayed empty, but it should show the configured attempts.
> 
> Yes, I am aware of this issue, waiting for @nb-ohad Input.

Fixed with @nb-ohad help.

lallinger-arbeit:
@liranmauda The attempts field now shows up as desired. Thinking about it i think it would be more straightforward if the field would be called `Retries` instead of `Attempts`. As attempts=0 would mean the trigger shouldn't be run...

If you rename it to retries i would suggest the following tooltip: "Number of retries Noobaa will make if the trigger fails"

liranmauda:
> @liranmauda The attempts field now shows up as desired. Thinking about it i think it would be more straightforward if the field would be called `Retries` instead of `Attempts`. As attempts=0 would mean the trigger shouldn't be run...
> 
> If you rename it to retries i would suggest the following tooltip: "Number of retries Noobaa will make if the trigger fails"

We can change it to retries. 
Currently, If you try to put 0 in the attempts field it will not let you.
If we change it to retries we should allow 0

lallinger-arbeit:
> We can change it to retries.
> Currently, If you try to put 0 in the attempts field it will not let you.
> If we change it to retries we should allow 0

I would prefere this, sounds better than attempts

liranmauda:
Changed to retries in the UI, changed the tooltip as suggested, and allowed 0 to be an input (as now it is "retry" and not "attempt")

lallinger-arbeit:
@liranmauda update me when you have added the retry field to the log message in the db, i will retest then

liranmauda:
> > Changed to retries in the UI, changed the tooltip as suggested, and allowed 0 to be an input (as now it is "retry" and not "attempt")
> 
> thanks!
> 
> @liranmauda update me when you have added the retry field to the log message in the db, i will retest then

@lallinger-arbeit I think we should open a gap and do this in a different PR, And merge this one.
The `P.retry` that we are using is not reflecting the current number of attempts left.

lallinger-arbeit:
> @lallinger-arbeit I think we should open a gap and do this in a different PR, And merge this one.
> The `P.retry` that we are using is not reflecting the current number of attempts left.

i'm fine with that

src/server/system_services/schemas/bucket_schema.js
```diff
@@ -319,6 +319,9 @@ module.exports = {
                     object_suffix: {
                         type: 'string'
                     },
+                    attempts: {
```
lallinger-arbeit, line 322: @liranmauda i think the field should be called `retries` in the API too

liranmauda: The API is with Attempts and not retries as it reflects the behavior better.
We are using `P.retry` and it uses attempts. 
Providing 1 will run it once without retries, providing more will do retries.
https://github.com/noobaa/noobaa-core/blob/de261f32fb59a973fb244e84512f572be0a892f4/src/util/promise.js#L191-L195
I have changed the UI to be more straightforward, The BE left with attempts.

lallinger-arbeit: ok

jeniawhite (approved):

PR #6373, opened 2021-02-09
Fixing NS Auth

### LINKED ISSUE #6371: Unauthorized download of objects is possible with namespaced buckets
### Environment info
- NooBaa Version: current master
- Platform: OKD 4.5

### Actual behavior
1. Create a namespaced bucket `somebucket`
2. Place `somefile.txt` with any data in it in the bucket
3. Run this python code it is possible to download any object without specifying a access key or secret key (set `endpoint_url` correctly):
```
import boto3
from botocore import UNSIGNED
from botocore.client import Config
s3 = boto3.client('s3',
                  endpoint_url="https://s3:443",
                  verify=False, use_ssl=True, config=Config(signature_version=UNSIGNED))
s3.download_file('somebucket', 'somefile.txt', 'out.txt')
```
4.  `somefile.txt` will be downloaded as `out.txt`

### Expected behavior
1. Download fails with "Unauthorized" or similiar error

### Steps to reproduce
1. see above

jeniawhite:
### Explain the changes
1. Passing arguments correctly

1. Fixed https://github.com/noobaa/noobaa-core/issues/6371

### Testing Instructions:
1.

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -299,7 +299,7 @@ function authenticate_request(req) {
 
 async function authorize_request(req) {
     await Promise.all([
-        req.object_sdk.authorize_request_account(req.params.bucket),
+        req.object_sdk.authorize_request_account(req),
         authorize_request_policy(req)
     ]);
 }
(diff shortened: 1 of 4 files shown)


guymguym (approved):
Approving, but please see how to add an anonymous test for a bucket that does not allow it. thanks

PR #6382, opened 2021-02-18
Add claim_id tag to Prometheus metrics - buckets labels

romayalon:

### Explain the changes
1.  bucket.tag is an unused field in bucket info. Exported it if exists in Prometheus metrics of buckets & namespace buckets.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/stats_api.js
+++ b/src/api/stats_api.js
@@ -728,7 +728,7 @@ module.exports = {
                     type: 'array',
                     items: {
                         type: 'object',
-                        required: ['bucket_name', 'quota_precent', 'capacity_precent', 'is_healthy'],
+                        required: ['bucket_name', 'quota_precent', 'capacity_precent', 'is_healthy', 'tagging'],
                         properties: {
                             bucket_name: {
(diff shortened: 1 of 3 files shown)


src/api/stats_api.js
```diff
@@ -742,6 +742,9 @@ module.exports = {
                             is_healthy: {
                                 type: 'boolean'
                             },
+                            tagging: {
```
jeniawhite, line 745: Make required if always exists

src/api/stats_api.js
```diff
@@ -770,6 +773,9 @@ module.exports = {
...
```
jeniawhite, line 776: Make required if always exists

jeniawhite (approved):

src/api/stats_api.js
romayalon: added, thanks.

src/api/stats_api.js
romayalon: added, thanks.

PR #6389, opened 2021-02-24
Added delete external connection to encryption unit tests

romayalon:

### Explain the changes
1. 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/test/unit_tests/test_encryption.js
+++ b/src/test/unit_tests/test_encryption.js
@@ -414,6 +414,10 @@ mocha.describe('Encryption tests', function() {
                 await rpc_client.account.delete_account({ email: cur_account.email});
             }));
         });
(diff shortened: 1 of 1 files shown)


src/test/unit_tests/test_encryption.js
```diff
@@ -414,6 +414,10 @@ mocha.describe('Encryption tests', function() {
                 await rpc_client.account.delete_account({ email: cur_account.email});
             }));
         });
+        mocha.it('delete external connections1 succefully', async function() {
```
liranmauda, line 417: Maybe a different description as it creates/deletes and not only deletes.

liranmauda (approved):

src/test/unit_tests/test_encryption.js
romayalon: thanks,  fixed.

PR #6390, opened 2021-02-24
Lambda Functions - Write the code into the DB

liranmauda:
### Explain the changes
- Writing the code into the DB
- Fix an issue where child process called generate_entropy on a container
- Async await some code

### Issues:
1. when using postgres db the Function Statistics in the UI is not working (We are missing some MapReduce functionality in postgres, regarding func_stats) 

### Limitation
1. As we are missing func_stats MapReduce for postgres, the func_stats will not be written in the db when using postgres

### Testing Instructions:
1. Test lambda function create/delete/update/trigger
2. Test with mongodb and with postgres

### DIFF
--- a/src/server/func_services/func_server.js
+++ b/src/server/func_services/func_server.js
@@ -2,10 +2,12 @@
 'use strict';
 
 const _ = require('lodash');
-const stream = require('stream');
 const request = require('request');
 
+const crypto = require('crypto');
 const P = require('../../util/promise');
+const stream = require('stream');
+const buffer_utils = require('../../util/buffer_utils');
 const { RpcError, RPC_BUFFERS } = require('../../rpc');
 const dbg = require('../../util/debug_module')(__filename);
 const FuncNode = require('../../agent/func_services/func_node');
@@ -63,14 +65,7 @@ async function create_func(req) {
         }
     }
     const resource_name = `arn:noobaa:lambda:region:${system}:function:${name}:${version}`;
-    const code_stream = await _get_func_code_stream(req, func_code);
-
-    const {
-        id: code_gridfs_id,
-        sha256: code_sha256,
-        size: code_size
-    } = await func_store.instance()
-        .create_code_gridfs({ system, name, version, code_stream });
+    const { code, code_sha256, code_size } = await _get_func_code_b64(req, func_code);
 
     const func_id = func_store.instance().make_func_id();
     await func_store.instance().create_func({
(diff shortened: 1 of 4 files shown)


src/server/func_services/func_server.js
```diff
@@ -162,15 +164,28 @@ async function update_func(req) {
...
 }
 
 async function read_func(req) {
     await _load_func(req);
-    const reply = await _get_func_info(req.func);
+    const reply = _get_func_info(req.func);
     if (req.params.read_code) {
-        const zipfile = await func_store.instance().read_code_gridfs(req.func.code_gridfs_id);
+        const system = req.system._id;
+        const name = req.params.name || _.get(req, 'params.config.name');
```
jeniawhite, line 177: This does not exist in the API
`params.config`

src/server/func_services/func_server.js
```diff
@@ -274,16 +289,20 @@ async function invoke_func(req) {
...
+                node.name, node.pool);
+            res = await server_rpc.client.func_node.invoke_func(params, {
+                address: node.rpc_address
+            });
+        } else {
+            dbg.log0('invoking on server', func.name, req.params.event);
+            res = await server_func_node.invoke_func({ params });
+        }
+    } catch (e) {
+        dbg.error('invoke returned with error:', e);
```
jeniawhite, line 305: Add a throw here to not eat the errors.

src/server/func_services/func_store.js
```diff
@@ -36,122 +27,77 @@ class FuncStore {
...
-            }, {
-                $set: set_updates
-            });
+    async update_func(func_id, set_updates) {
+        const res = await this._funcs.updateOne({
+            _id: func_id,
+        }, {
+            $set: set_updates
         });
+        return res;
```
jeniawhite, line 56: We shouldn't return the res of updateOne.

jeniawhite (commented):
Please re-check all of the RPC APIs regarding parameters so you won't have the same issue as you've had with `params.config.`

src/server/func_services/func_server.js
```diff
@@ -316,23 +321,19 @@ function check_event_permission(req) {
...
     if (func_code.zipfile_b64) {
         // zipfile is given as base64 string
-        return new stream.Readable({
-            read(size) {
-                this.push(Buffer.from(func_code.zipfile_b64, 'base64'));
-                this.push(null);
-            }
-        });
+        code = func_code.zipfile_b64;
+        code_size = Buffer.from(func_code.zipfile_b64).length;
```
jeniawhite, line 332: Is there a reason you did not do 'base64' or you just missed it?
Buffer.from(func_code.zipfile_b64, 'base64')

liranmauda: This flow is when the zipfile is given as base64 string

jeniawhite (approved):

PR #6391, opened 2021-03-01
removing double pipe causing issues on get_object

jackyalbo:
Fix for BZ1930075

### Explain the changes
1. removed one of the 2 pipes to count_stream. Checked without for both small (1M) and big files(100M) and seems to do the trick. Counting seems to work as well

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -3,7 +3,6 @@
 
 const _ = require('lodash');
 const util = require('util');
-const stream = require('stream');
 const crypto = require('crypto');
 
 const P = require('../util/promise');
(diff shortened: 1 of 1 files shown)


src/sdk/namespace_blob.js
```diff
@@ -215,7 +215,7 @@ class NamespaceBlob {
                         inspect(_.omit(params, 'object_md.ns')),
                         'callback res', inspect(res)
                     );
-                    return resolve(read_stream.pipe(count_stream));
+                    return resolve(read_stream);
```
dannyzaken, line 218: I think that this `resolve` might be redundant since the promise is resolved when the stream becomes readable. in any case, I don't think you should resolve in two different ways in different flows. you can maybe pipe once when creating the stream

dannyzaken, line 218: after thinking about it I think that the pass through `read_stream` is redundant. just use the `count_stream`

jackyalbo: removed the read_stream

src/sdk/namespace_blob.js
```diff
@@ -160,12 +160,12 @@ class NamespaceBlob {
                 // clear count for next updates
                 count = 0;
             });
-            const read_stream = new stream.PassThrough();
+            // const read_stream = new stream.PassThrough();
```
dannyzaken, line 162: remove this

dannyzaken (approved):

PR #6393, opened 2021-03-02
NSFS Fix mtime compare in dir cache

guymguym:
### Explain the changes
1. Comparison was incorrect because mtime is Date object.

1. NA

### Testing Instructions:
1. Test list-objects with large (many thousands) of items to verify that readdir is done once into the cache.
2. Notice that we might also need to tweak config.NSFS_DIR_CACHE_MAX_DIR_SIZE to higher value for that to work, and we will probably want to bump it anyhow.


guymguym:
will open a new PR - best use a fork for DCO to work. closing.

PR #6396, opened 2021-03-03
Initial changes for exporting dirs as buckets - NSFS

romayalon:

### Explain the changes
1. namespace resource definition, API, schema: 
     a. removed connection, target_bucket, endpoint, endpoint type from being required.  
     b. Added fs_config to namespace resource properties (includes fs_path and protocol)
2. object_sdk - returned new namespaceFS when write resource contains fs_config (instead of when env var)
3. pool_server - made some changes on namespace_resource according to the API/Schema changes.
4. namespace_fs - added fs_path and protocol to the constructor (instead of root_path) and removed the redundant bucket name addition to the path.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/pool_api.js
+++ b/src/api/pool_api.js
@@ -91,7 +91,7 @@ module.exports = {
             method: 'POST',
             params: {
                 type: 'object',
-                required: ['name', 'connection', 'target_bucket'],
+                required: ['name'],
                 properties: {
                     name: {
                         type: 'string',
@@ -101,6 +101,9 @@ module.exports = {
                     },
                     target_bucket: {
                         type: 'string',
+                    },
+                    nsfs_config: {
+                        $ref: 'common_api#/definitions/nsfs_config'
                     },
                     namespace_store: {
                         type: 'object',
@@ -529,7 +532,7 @@ module.exports = {
 
         namespace_resource_info: {
             type: 'object',
-            required: ['name', 'endpoint_type', 'endpoint', 'target_bucket'],
+            required: ['name'],
             properties: {
                 name: {
                     type: 'string'
@@ -552,6 +555,12 @@ module.exports = {
                 identity: {
                     $ref: 'common_api#/definitions/access_key'
                 },
+                fs_path: {
+                    $ref: 'common_api#/definitions/fs_path'
+                },
+                fs_backend: {
+                    $ref: 'common_api#/definitions/fs_backend'
+                },
                 mode: {
                     type: 'string',
                     enum: ['OPTIMAL', 'STORAGE_NOT_EXIST', 'AUTH_FAILED', 'IO_ERRORS']
@@ -565,7 +574,7 @@ module.exports = {
         // Currently the extended info has an addition of secret_key
         namespace_resource_extended_info: {
             type: 'object',
-            required: ['name', 'endpoint_type', 'endpoint', 'target_bucket'],
+            required: ['name'],
             properties: {
                 id: {
                     objectid: true
@@ -590,6 +599,12 @@ module.exports = {
                     type: 'string'
                 },
                 secret_key: { $ref: 'common_api#/definitions/secret_key' },
+                fs_path: {
+                    $ref: 'common_api#/definitions/fs_path'
+                },
+                fs_backend: {
+                    $ref: 'common_api#/definitions/fs_backend'
+                },
             }
         },
 
@@ -794,6 +809,6 @@ module.exports = {
                 'AUTH_FAILED',
                 'OPTIMAL'
             ]
-        }
+        },
     }
 };
--- a/src/core/nsfs.js
+++ b/src/core/nsfs.js
@@ -93,10 +93,14 @@ async function main(argv = minimist(process.argv.slice(2))) {
         // };
 
         const bs = new BucketSpaceFS({ fs_root });
-        const ns = new NamespaceFS({ fs_root });
         const object_sdk = new ObjectSDK(null, null, null);
+        const fs_namespaces = {};
         object_sdk._get_bucketspace = () => bs;
-        object_sdk._get_bucket_namespace = async name => ns;
+        object_sdk._get_bucket_namespace = async bucket_name => {
+            const ns = fs_namespaces[bucket_name] || new NamespaceFS({ fs_path: fs_root + '/' + bucket_name });
+            fs_namespaces[bucket_name] = ns;
+            return ns;
+        };
         object_sdk.get_auth_token = noop;
         object_sdk.set_auth_token = noop;
         object_sdk.authorize_request_account = noop;
--- a/src/sdk/object_sdk.js
+++ b/src/sdk/object_sdk.js
@@ -58,9 +58,6 @@ class ObjectSDK {
         this.object_io = object_io;
         this.namespace_nb = new NamespaceNB();
         this.bucketspace_nb = new BucketSpaceNB({ rpc_client });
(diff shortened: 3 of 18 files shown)


src/core/nsfs.js
```diff
@@ -29,13 +29,16 @@ Help:
 const USAGE = `
 Usage:
 
-    noobaa-core nsfs <root-path> [options...]
+    noobaa-core nsfs <root-path> <fs-path> <protocol> <bucket-name> [options...]
```
guymguym, line 32: @romayalon the nsfs command is meant to behave differently - not to export a single bucket, but instead to export a root dir so that every subdir is a bucket. Can you preserve this behavior?

romayalon: I removed these flags and kept nsfs with the minimal changes I can, but the behavior won't be exactly the same since there are more changes in namespace_fs for exporting a single bucket.

src/core/nsfs.js
```diff
@@ -93,7 +93,7 @@ async function main(argv = minimist(process.argv.slice(2))) {
         // };
 
         const bs = new BucketSpaceFS({ fs_root });
-        const ns = new NamespaceFS({ fs_root });
+        const ns = new NamespaceFS({ fs_path: fs_root, protocol: '' });
         const object_sdk = new ObjectSDK(null, null, null);
         object_sdk._get_bucketspace = () => bs;
         object_sdk._get_bucket_namespace = async name => ns;
```
guymguym, line 99: so the change we need here is a bit different for this to work like before - we need _get_bucket_namespace to be implemented now correctly so that it will create a new namespace based on the bucket name and use `fs_path: fs_root + '/' + bucket.name`. only nuance is that _get_bucket_namespace is called on every op and assumes it won't be heavy, so I would just use a map per bucket name and keep the created namespaces inside it.

guymguym (changes requested):
Added some more comments.

src/sdk/object_sdk.js
```diff
@@ -170,14 +167,6 @@ class ObjectSDK {
...
         try {
 
-            if (this.namespace_fs) {
-                return {
-                    ns: this.namespace_fs,
-                    bucket,
-                    valid_until: time + config.OBJECT_SDK_BUCKET_CACHE_EXPIRY_MS,
-                };
-            }
-
```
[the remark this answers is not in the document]
romayalon: sure, added it back.

src/server/system_services/schemas/namespace_resource_schema.js
```diff
@@ -52,6 +51,18 @@ module.exports = {
                 }
             }
         },
+        fs_config: {
```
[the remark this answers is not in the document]
romayalon: fixed

src/core/nsfs.js
romayalon: cool, fixed.

src/server/system_services/schemas/namespace_resource_schema.js
```diff
@@ -52,6 +51,18 @@ module.exports = {
...
+            type: 'object',
+            required: ['fs_path', 'protocol'],
+            properties: {
+                fs_path: {
+                    type: 'string'
+                },
+                protocol: {
+                    type: 'string'
+                }
```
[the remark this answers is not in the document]
romayalon: fixed

src/api/pool_api.js
```diff
@@ -552,6 +565,13 @@ module.exports = {
                 identity: {
                     $ref: 'common_api#/definitions/access_key'
                 },
+                fs_path: {
+                    type: 'string'
+                },
+                fs_backend: {
+                    type: 'string',
+                    enum: ['CEPH_FS', 'GPFS']
+                },
```
guymguym, line 574: why is this flat here? perhaps would be best to move nsfs_config to the definitions and $ref to it from all those places we use it?

src/server/system_services/schemas/namespace_resource_schema.js
```diff
@@ -52,6 +51,19 @@ module.exports = {
                 }
             }
         },
+        nsfs_config: {
+            type: 'object',
+            required: ['fs_path', 'fs_backend'],
```
guymguym, line 56: again fs_backend should not be required

src/api/pool_api.js
romayalon: it's flat because I wanted to keep the same behavior as for the connection properties (endpoint, endpoint_type etc are flat here too).

src/api/pool_api.js
```diff
@@ -102,6 +102,19 @@ module.exports = {
                     target_bucket: {
                         type: 'string',
                     },
+                    nsfs_config: {
+                        type: 'object',
+                        required: ['fs_path', 'fs_backend'],
```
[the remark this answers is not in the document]
romayalon: deleted

src/api/pool_api.js
```diff
@@ -552,6 +555,13 @@ module.exports = {
                 identity: {
                     $ref: 'common_api#/definitions/access_key'
                 },
+                fs_path: {
+                    type: 'string'
+                },
+                fs_backend: {
+                    type: 'string',
+                    enum: ['CEPH_FS', 'GPFS']
```
guymguym, line 563: Lets move this to definitions - we don't want to repeat enums multiple times

guymguym: Also @nimrod-becker - I think we want to add NFSv4 to the list as this will be a common use case - what do you say? I thought that using empty might be good for non-specific cases, but with NFSv4 I think it can get more specific than posix - WDYT?

guymguym: Oh @romayalon I see you already added it to definitions so just need to replace with $ref here and other occurrences of it

src/server/system_services/schemas/namespace_resource_schema.js
```diff
@@ -52,6 +51,19 @@ module.exports = {
...
+        nsfs_config: {
+            type: 'object',
+            required: ['fs_path'],
+            properties: {
+                fs_path: {
+                    type: 'string'
+                },
+                fs_backend: {
+                    type: 'string',
+                    enum: ['CEPH_FS', 'GPFS']
```
guymguym, line 63: why is this defined here again? this meanss you want the nsfs_config definition to move to common_api and used in both.

src/api/pool_api.js
romayalon: added NFSv4 and moved the definition to common api and replaced all occurrences to $ref

src/server/system_services/schemas/namespace_resource_schema.js
romayalon: moved it, thanks

src/api/pool_api.js
guymguym: @nimrod-becker Would still be good to get your take on NFSv4 here and if we need other granularity of NFSv4.1 or NFSv3 etc.

nimrod-becker: adding NFSv4 as the default is good, I don't think we'll start actually using the specific features of 4.1/4.2 or having to understand the differences between 3 and 4.
I think we can keep 1 for now.

guymguym: OK But I would still make this an optional field, and when not provided use the lowest common denominator possible of POSIX. I consider NFSv4 as a backend with more capabilities than POSIX.

nimrod-becker: Sure, we can do POSIX as default and NFSv4 as higher

guymguym (approved):

PR #6403, opened 2021-03-10
Disable Centos Jenkins jobs

liranmauda:
### Explain the changes
Disable Centos Jenkins jobs

Jenkins has an issue with pulling and GH Action is doing what we currently need, so we are disabling the Centos Jenkins jobs.
We might want to add to the Centos Jenkins jobs new tests (if GH Actions can’t handle the load) so we are keeping the code.


liranmauda:
@nixpanic Can you review

nixpanic (approved):
LGTM, but I always appreciate the reason for a change in the commit message too. This just disables automatic runs without explaining why that is done.

nixpanic (approved):
Thanks! :+1:

nimrod-becker (approved):

PR #6405, opened 2021-03-10
Introduction of NB FS Native module

jeniawhite:
### Explain the changes
1. Addition of FS native layer
2. Changing the namespace_fs to use nb fs native module
3. Added FileWrapper, DirWrapper
4. Ops added: stat, unlink, rename, rmkir, mkdir, writeFile, readFile, readdir

GAPS (#TODO):
1. Add UID capability
2. Rewrite createReadStream and createWriteStream
3. Fill up unit_tests for native
4. Passing more complex options to ops (bufferSize and etc...)

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -1,0 +1,721 @@
+/* Copyright (C) 2016 NooBaa */
+#include "../util/b64.h"
+#include "../util/common.h"
+#include "../util/napi.h"
+
+#include <uv.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <dirent.h>
+#include <vector>
+#include <math.h>
+// #include <errno.h>
+// #include <sys/fsuid.h>
+
+namespace noobaa
+{
+
+struct Entry {
+    std::string name;
+    ino_t ino;
+    uint8_t type;
+};
+
+template <typename T>
+static Napi::Value api(const Napi::CallbackInfo& info)
+{
+    auto w = new T(info);
+    Napi::Promise promise = w->_deferred.Promise();
+    w->Queue();
+    return promise;
+}
+
+/**
+ * FSWorker is a general async worker for our fs operations
+ */
+struct FSWorker : public Napi::AsyncWorker
+{
+    Napi::Promise::Deferred _deferred;
+    int _errno;
+    std::string _desc;
+
+    FSWorker(const Napi::CallbackInfo& info)
+        : AsyncWorker(info.Env())
+        , _deferred(Napi::Promise::Deferred::New(info.Env()))
+        , _errno(0)
+    {
+    }
+    void Begin(std::string desc)
+    {
+        _desc = desc;
+        LOG("FS::FSWorker::Begin: " << _desc);
+    }
+    void SetSyscallError()
+    {
+        if (_errno) {
+            int current_errno = errno;
+            LOG("FS::FSWorker::SetSyscallError: errno already exists " << _desc << DVAL(_errno) << DVAL(current_errno));
+        } else {
+            _errno = errno;
+            std::string errmsg = strerror(errno);
+            SetError(errmsg);
+        }
+    }
+    virtual void OnOK()
+    {
+        LOG("FS::FSWorker::OnOK: undefined " << _desc);
+        _deferred.Resolve(Env().Undefined());
+    }
+    virtual void OnError(Napi::Error const &error)
+    {
+        Napi::Env env = Env();
+        LOG("FS::FSWorker::OnError: " << _desc << " " << DVAL(error.Message()));
+        auto obj = error.Value();
+        if (_errno) {
+            obj.Set("code", Napi::String::New(env, uv_err_name(uv_translate_sys_error(_errno))));
+        }
+        _deferred.Reject(obj);
+    }
+};
+
+/**
+ * Stat is an fs op
+ */
+struct Stat : public FSWorker
+{
+    std::string _path;
+    struct stat _stat_res;
+
+    Stat(const Napi::CallbackInfo& info) : FSWorker(info)
+    {
+        _path = info[0].As<Napi::String>();
+        Begin(XSTR() << DVAL(_path));
+    }
+    virtual void Execute()
+    {
+        int r = stat(_path.c_str(), &_stat_res);
+        if (r) SetSyscallError();
+    }
+    virtual void OnOK()
+    {
+        LOG("FS::Stat::OnOK: " << DVAL(_path) << DVAL(_stat_res.st_ino) << DVAL(_stat_res.st_size));
+        Napi::Env env = Env();
+        auto res = Napi::Object::New(env);
+        res["dev"] = Napi::Number::New(env, _stat_res.st_dev);
+        res["ino"] = Napi::Number::New(env, _stat_res.st_ino);
+        res["mode"] = Napi::Number::New(env, _stat_res.st_mode);
+        res["nlink"] = Napi::Number::New(env, _stat_res.st_nlink);
+        res["uid"] = Napi::Number::New(env, _stat_res.st_uid);
+        res["gid"] = Napi::Number::New(env, _stat_res.st_gid);
+        res["rdev"] = Napi::Number::New(env, _stat_res.st_rdev);
+        res["size"] = Napi::Number::New(env, _stat_res.st_size);
+        res["blksize"] = Napi::Number::New(env, _stat_res.st_blksize);
+        res["blocks"] = Napi::Number::New(env, _stat_res.st_blocks);
+
+        // https://nodejs.org/dist/latest-v14.x/docs/api/fs.html#fs_stat_time_values
+        #if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
+            double atimeMs = (double(1e3) * _stat_res.st_atimespec.tv_sec) + (double(1e-6) * _stat_res.st_atimespec.tv_nsec);
+            double ctimeMs = (double(1e3) * _stat_res.st_ctimespec.tv_sec) + (double(1e-6) * _stat_res.st_ctimespec.tv_nsec);
+            double mtimeMs = (double(1e3) * _stat_res.st_mtimespec.tv_sec) + (double(1e-6) * _stat_res.st_mtimespec.tv_nsec);
+            double birthtimeMs = (double(1e3) * _stat_res.st_birthtimespec.tv_sec) + (double(1e-6) * _stat_res.st_birthtimespec.tv_nsec);
+        #else
+            double atimeMs = (double(1e3) * _stat_res.st_atim.tv_sec) + (double(1e-6) * _stat_res.st_atim.tv_nsec);
+            double ctimeMs = (double(1e3) * _stat_res.st_ctim.tv_sec) + (double(1e-6) * _stat_res.st_ctim.tv_nsec);
+            double mtimeMs = (double(1e3) * _stat_res.st_mtim.tv_sec) + (double(1e-6) * _stat_res.st_mtim.tv_nsec);
+            double birthtimeMs = ctimeMs; // Posix doesn't have birthtime
+        #endif
+
+        res["atimeMs"] = Napi::Number::New(env, atimeMs);
+        res["ctimeMs"] = Napi::Number::New(env, ctimeMs);
+        res["mtimeMs"] = Napi::Number::New(env, mtimeMs);
+        res["birthtimeMs"] = Napi::Number::New(env, birthtimeMs);
+        res["atime"] = Napi::Date::New(env, uint64_t(round(atimeMs)));
+        res["mtime"] = Napi::Date::New(env, uint64_t(round(mtimeMs)));
+        res["ctime"] = Napi::Date::New(env, uint64_t(round(ctimeMs)));
+        res["birthtime"] = Napi::Date::New(env, uint64_t(round(birthtimeMs)));
+
+        _deferred.Resolve(res);
+    }
+};
+
+/**
+ * Unlink is an fs op
+ */
+struct Unlink : public FSWorker
+{
+    std::string _path;
+    Unlink(const Napi::CallbackInfo& info) : FSWorker(info)
+    {
+        _path = info[0].As<Napi::String>();
+        Begin(XSTR() << DVAL(_path));
+    }
+    virtual void Execute()
+    {
+        int r = unlink(_path.c_str());
+        if (r == -1) SetSyscallError();
+    }
+};
+
+
+/**
+ * Mkdir is an fs op
+ */
+struct Mkdir : public FSWorker
+{
+    std::string _path;
+    Mkdir(const Napi::CallbackInfo& info) : FSWorker(info)
+    {
+        _path = info[0].As<Napi::String>();
+        Begin(XSTR() << DVAL(_path));
+    }
+    virtual void Execute()
+    {
+        int r = mkdir(_path.c_str(), S_IRWXU);
+        if (r == -1) SetSyscallError();
+    }
+};
+
+
+/**
+ * Rmdir is an fs op
+ */
+struct Rmdir : public FSWorker
+{
+    std::string _path;
+    Rmdir(const Napi::CallbackInfo& info) : FSWorker(info)
+    {
+        _path = info[0].As<Napi::String>();
+        Begin(XSTR() << DVAL(_path));
+    }
+    virtual void Execute()
+    {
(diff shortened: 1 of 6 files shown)


src/deploy/NVA_build/builder.Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.centos.org/centos:8 
+FROM centos:8
```
guymguym, line 1: @jeniawhite Is this change intended? CC @liranmauda

jeniawhite: My mistake, the repo did not work for me. Disregard.
Thank you.

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+    {
+        LOG("FS::FSWorker::OnOK: undefined " << _desc);
+        _deferred.Resolve(Env().Undefined());
+    }
+    virtual void OnError(Napi::Error const &error)
+    {
+        Napi::Env env = Env();
+        LOG("FS::FSWorker::OnError: " << _desc << " " << DVAL(error.Message()));
+        auto obj = error.Value();
+        obj.Set("code", Napi::String::New(env, uv_err_name(uv_translate_sys_error(_errno))));
```
guymguym, line 70: Maybe need to set this only `if (_errno)`

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+        int fd = open(_path.c_str(), O_WRONLY | O_CREAT);
+        if (fd < 0) {
+            SetSyscallError();
+            return;
+        }
+
+        ssize_t len = write(fd, _data, _len);
+        if (len < 0) {
+            SetSyscallError();
+            return;
```
guymguym, line 219: I think we want to avoid return here to try and close the fd on error

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+        }
+        if ((size_t)len != _len) {
+            SetError(XSTR() << "Writefile: partial write error " << DVAL(len) << DVAL(_len));
+            return;
```
guymguym, line 223: again - best continue after setting the error and try to close the fd

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+    {
+    }
+    void Begin(std::string desc)
+    {
+        _desc = desc;
+        LOG("FS::FSWorker::Begin: " << _desc);
+    }
+    void SetSyscallError()
+    {
+        _errno = errno;
```
guymguym, line 56: I think we want to only set it if _errno is not set (0) otherwise log and return.

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+            SetSyscallError();
+            return;
+        }
+
+        ssize_t len = write(fd, _data, _len);
+        if (len < 0) {
+            SetSyscallError();
+            return;
+        }
+        if ((size_t)len != _len) {
```
guymguym, line 221: else if iss needed if we remove the previous return

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+        _len = stat_res.st_size;
+        _data = new uint8_t[_len];
+
+        uint8_t *p = _data;
+        int remain = _len;
+        while (remain > 0) {
+            ssize_t len = read(fd, p, remain);
+            if (len < 0) {
+                SetSyscallError();
+                return;
```
guymguym, line 278: again, best try to break to close the fd

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+        if (fd < 0) {
+            SetSyscallError();
+            return;
+        }
+    
+        struct stat stat_res;
+        int r = fstat(fd, &stat_res);
+        if (r != 0) {
+            SetSyscallError();
+            return;
```
guymguym, line 266: need to close here

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+
+
+/**
+ * Readdir is an fs op
+ */
+struct Readdir : public FSWorker
+{
+    std::string _path;
+    bool _withFileTypes;
+    DIR *_dir;
```
guymguym, line 302: _dir should not be a member here, move to local in Execute

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+        Napi::Env env = Env();
+        Napi::Array res = Napi::Array::New(env);
+        int index = 0;
+        if (_withFileTypes) {
+            for (auto it = _entries.begin(); it != _entries.end(); ++it) {
+                // Ingnore parent and current directories 
+                if (it->name == "." || it->name == "..") {
+                    continue;
+                }
+                auto dir_rec = Napi::Array:: Object::New(env);
```
guymguym, line 356: something here looks wrong. need to be Napi::Object::New.

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+            }
+        }
+
+        int r = closedir(_dir);
+        if (r) SetSyscallError();
+    }
+    virtual void OnOK()
+    {
+        Napi::Env env = Env();
+        Napi::Array res = Napi::Array::New(env);
```
guymguym, line 348: Best init with size `_entries.size() - 2` (without `.` and `..`)

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+
+//         // TODO - read(fd, buf...)
+//     }
+// };
+
+
+// /**
+//  * Writefile is an fs op
+//  */
+// struct Writefile : public FSWorker
```
guymguym, line 527: still needed? I see we already have the Writefile op above

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+        }
+    }
+    virtual void OnOK()
+    {
+        Napi::Env env = Env();
+        _deferred.Resolve(Napi::Number::New(env, _br));
+    }
+};
+
+// struct FileWrite : public FSWorker
```
guymguym, line 498: Are you going to complete this op in this PR?

src/test/unit_tests/test_nb_native_fs.js
```diff
@@ -0,0 +1,173 @@
...
+
+// const _ = require('lodash');
+const fs = require('fs');
+const mocha = require('mocha');
+const assert = require('assert');
+const nb_native = require('../../util/nb_native');
+
+mocha.describe('nb_native fs', function() {
+
+    // mocha.describe('stat', function() {
```
guymguym, line 12: what's up with this test? Why is it all commented out?

src/sdk/namespace_fs.js
```diff
@@ -711,14 +724,13 @@ class NamespaceFS {
         // IMPORTANT NOTICE - we must return an etag that contains a dash!
         // because this is the criteria of S3 SDK to decide if etag represents md5
         // and perform md5 validation of the data.
-        const ident_str = 'inode-' + stat.ino.toString() + '-mtime-' + stat.mtimeMs.toString();
+        const ident_str = 'inode-' + stat.ino.toString() + '-mtime-' + stat.mtime.toString();
```
guymguym, line 727: what is the resolution of the mtime we return from stat?

src/sdk/namespace_fs.js
```diff
@@ -783,13 +793,60 @@ class NamespaceFS {
...
             throw err;
         }
     }
 
+    async _create_path(dir) {
+        const sep = path.sep;
+        const initDir = path.isAbsolute(dir) ? sep : '';
+        const baseDir = '.';
+
+        return dir.split(sep).reduce(async (parentDir, childDir) => {
```
guymguym, line 809: is this code copied from somewhere?
I would not use this without permission to copy.
In any case I don't like how this is written at all.
I would much prefer to write like this:
```js
let dir_path = path.isAbsolute(dir) ? path.sep : '';
for (const item of dir.split(path.sep)) {
    dir_path = path.join(dir_path, item);
    try {
        await nb_native().mkdir(dir_path);
    } catch (err) {
        if (err.code !== 'EEXIST') throw err;
    }
}
```

src/sdk/namespace_fs.js
```diff
@@ -783,13 +793,60 @@ class NamespaceFS {
...
+        let entries = await nb_native().readdir(dir, /* withFileTypes */ true);
+        let results = await Promise.all(entries.map(entry => {
+          let fullPath = path.join(dir, entry.name);
+        //   if (fullPath.startsWith('/tmp/')) {
+          let task = isDirectory(entry) ? this._folder_delete(fullPath) : nb_native().unlink(fullPath);
+          return task.catch(error => ({ error }));
+        //   }
+        }));
+        results.forEach(result => {
+          // TODO: Handle errors correctly
```
guymguym, line 841: seems like this code is still work in progress - is it ready for review?

guymguym (changes requested):

src/sdk/namespace_fs.js
jeniawhite: The timestamp is in seconds

jeniawhite: #if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
	struct  timespec st_atimespec;  /* time of last access */
	struct  timespec st_mtimespec;  /* time of last data modification */
	struct  timespec st_ctimespec;  /* time of last status change */
#else
	time_t          st_atime;       /* [XSI] Time of last access */
	long            st_atimensec;   /* nsec of last access */
	time_t          st_mtime;       /* [XSI] Last data modification time */
	long            st_mtimensec;   /* last data modification nsec */
	time_t          st_ctime;       /* [XSI] Time of last status change */
	long            st_ctimensec;   /* nsec of last status change */
#endif

As you can see I can take the value from `st_mtimensec` but not on darwin, so I will do an if there.

jeniawhite: `
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
	struct  timespec st_atimespec;  /* time of last access */
	struct  timespec st_mtimespec;  /* time of last data modification */
	struct  timespec st_ctimespec;  /* time of last status change */
#else
	time_t          st_atime;       /* [XSI] Time of last access */
	long            st_atimensec;   /* nsec of last access */
	time_t          st_mtime;       /* [XSI] Last data modification time */
	long            st_mtimensec;   /* last data modification nsec */
	time_t          st_ctime;       /* [XSI] Time of last status change */
	long            st_ctimensec;   /* nsec of last status change */
#endif
`
And this is the alias
`
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
#define st_atime st_atimespec.tv_sec
#define st_mtime st_mtimespec.tv_sec
#define st_ctime st_ctimespec.tv_sec
#define st_birthtime st_birthtimespec.tv_sec
#endif /* (!_POSIX_C_SOURCE || _DARWIN_C_SOURCE) */
`
As you can see on darwin we do not have the option for nanoseconds, but I will add a code that will pull from nanoseconds as well and on darwin, I will just multiply for our local tests to pass.

guymguym: Darwin uses `struct timespec` which provides nano seconds too:
```
$ tail /Library/Developer/CommandLineTools/SDKs/MacOSX10.15.sdk/usr/include/sys/_types/_timespec.h
#define _STRUCT_TIMESPEC        struct timespec

#include <machine/types.h> /* __darwin_time_t */

_STRUCT_TIMESPEC
{
	__darwin_time_t tv_sec;
	long            tv_nsec;
};
#endif /* _STRUCT_TIMESPEC */
```

guymguym: In any case - returning seconds is wrong - we need to convert to ms in order to be compatible with js dates.

src/native/fs/fs_napi.cpp
```diff
@@ -0,0 +1,747 @@
...
+        res["nlink"] = Napi::Number::New(env, _stat_res.st_nlink);
+        res["uid"] = Napi::Number::New(env, _stat_res.st_uid);
+        res["gid"] = Napi::Number::New(env, _stat_res.st_gid);
+        res["rdev"] = Napi::Number::New(env, _stat_res.st_rdev);
+        res["size"] = Napi::Number::New(env, _stat_res.st_size);
+        res["blksize"] = Napi::Number::New(env, _stat_res.st_blksize);
+        res["blocks"] = Napi::Number::New(env, _stat_res.st_blocks);
+        res["atime"] = Napi::Number::New(env, _stat_res.st_atime);
+        res["mtime"] = Napi::Number::New(env, _stat_res.st_mtime);
+        res["ctime"] = Napi::Number::New(env, _stat_res.st_ctime);
```
guymguym, line 110: Based on what we discussed below on times - we need to convert the times to milliseconds to be compatible with js dates. Also need to take into account the nano seconds precision which requires some difference between Mac and Linux. This example works on both linux and mac:

```
$ cat a.cpp
#include <stdio.h>
#include <sys/stat.h>
int main()
{
	struct stat st;
	int r = stat(".", &st);
#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
	double d = (double(1e3) * st.st_mtimespec.tv_sec) + (double(1e-6) * st.st_mtimespec.tv_nsec);
#else
	double d = (double(1e3) * st.st_mtim.tv_sec) + (double(1e-6) * st.st_mtim.tv_nsec);
#endif
	printf("mtime = %lf \n", d);
	return 0;
}
```

src/native/fs/fs_napi.cpp
jeniawhite: Next PR, I will also probably rewrite the logic for uploads and reads to use the FileWrapper

src/test/unit_tests/test_nb_native_fs.js
jeniawhite: I will fill this up in the next PR as well

src/sdk/namespace_fs.js
```diff
@@ -75,8 +80,8 @@ const dir_cache = new LRUCache({
         return { time, stat, sorted_entries, usage };
     },
     validate: async ({ stat }, dir_path) => {
-        const new_stat = await fs.promises.stat(dir_path);
-        return (new_stat.ino === stat.ino && new_stat.mtime.getTime() === stat.mtime.getTime());
+        const new_stat = await nb_native().stat(dir_path);
+        return (new_stat.ino === stat.ino && new_stat.mtime === stat.mtime);
```
guymguym, line 84: CAREFUL HERE - this change is wrong now that you changed stat.mtime to be a Date object !
This was a bug I fixed in https://github.com/noobaa/noobaa-core/pull/6394
It would be best to actually compare mtimeMs instead here.
Please fix.

jeniawhite: Somehow missed leaving the getTime().
I've changed to `mtimeMs`.
Thanks.

src/sdk/namespace_fs.js
```diff
@@ -62,11 +67,11 @@ const dir_cache = new LRUCache({
     name: 'nsfs-dir-cache',
     load: async dir_path => {
         const time = Date.now();
-        const stat = await fs.promises.stat(dir_path);
+        const stat = await nb_native().stat(dir_path);
         let sorted_entries;
         let usage = config.NSFS_DIR_CACHE_MIN_DIR_SIZE;
         if (stat.size <= config.NSFS_DIR_CACHE_MAX_DIR_SIZE) {
-            sorted_entries = await fs.promises.readdir(dir_path, { withFileTypes: true });
+            sorted_entries = await nb_native().readdir(dir_path, /* withFileTypes */ true);
```
guymguym, line 74: In next PR I recommend to keep the second arg an object like fs.promises does.

guymguym: Hmm on second thought - do we even need this option??
Isn't it always better to return the file types instead of having this option?

jeniawhite: I don't see a reason, since this just creates more checks in the native C++ layer and we aren't actually getting any value from it in the NodeJS code.
Do you want me to remove it?

guymguym: I think so. If we later decide we need it we can always introduce it.

jeniawhite: done

guymguym (approved):

PR #6410, opened 2021-03-15
Return NoSuchKey for directories head request.

eshelmarc:
### Explain the changes
1. lookup for keys (get head) of directory should return not found since they are not real objects

1. no issue 

### Testing Instructions:
1. get head on a directory

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -353,6 +353,11 @@ class NamespaceFS {
             const file_path = this._get_file_path(params);
             const stat = await nb_native().fs.stat(DEFAULT_FS_CONFIG, file_path);
             console.log(file_path, stat);
+            if (isDirectory(stat)) {
+                const err = new Error('NoSuchKey');
+                err.code = 'ENOENT';
+                throw this._translate_object_error_codes(err);
+            }
             return this._get_object_info(params.bucket, params.key, stat);
         } catch (err) {
             throw this._translate_object_error_codes(err);


src/sdk/namespace_fs.js
```diff
@@ -333,7 +333,14 @@ class NamespaceFS {
...
             console.log(file_path, stat);
-            return this._get_object_info(params.bucket, params.key, stat);
+            if (stat.isDirectory()) {
+                const err = new Error('NoSuchKey');
+                err.code = 'ENOENT';
+                throw this._translate_object_error_codes(err);
+            }
+	    else {
+	      return this._get_object_info(params.bucket, params.key, stat);
+            }
```
guymguym, line 343: something is off with indent - make sure your replace tabs with 4 spaces.
Also since we throw just before and now have made sure the rest of the flow is only for files, we can ditch the else clause:
```suggestion
            return this._get_object_info(params.bucket, params.key, stat);
```

guymguym (commented):
For the DCO check to pass you need to signoff your commits (see https://github.com/apps/dco)
One simple way to do that is with `git fetch --all && git rebase --signoff upstream/master && git push origin -f`

src/sdk/namespace_fs.js
eshelmarc: I thought I pushed the changes you recommended but I don't see them here, I must have made a mistake

eshelmarc: still failing some test

src/sdk/namespace_fs.js
```diff
@@ -333,7 +333,11 @@ class NamespaceFS {
             const file_path = this._get_file_path(params);
             const stat = await fs.promises.stat(file_path);
             console.log(file_path, stat);
-            return this._get_object_info(params.bucket, params.key, stat);
+            if (stat.isDirectory()) {
+                const err = new Error('NoSuchKey');
+                err.code = 'ENOENT';
+                throw this._translate_object_error_codes(err);
+            }
```
guymguym, line 340: @eshelmarc you need to add back this line:
```
return this._get_object_info(params.bucket, params.key, stat);
```

guymguym: basically you need to push your new if block before the existing return get object info line, but keep it after your new block.

eshelmarc:
how does it look? should I have used amend?

src/sdk/namespace_fs.js
guymguym: the "Unit Tests with Postgres" failures should be ignore for now.
right @liranmauda @nimrod-becker ?

guymguym:
Looks good now @eshelmarc.
About amend - yes you could either use amend commit or squash commits in the PR when the review is complete (the result is the same).
See how to squash now - https://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed/5668050#5668050

src/sdk/namespace_fs.js
liranmauda: @guymguym correct
@nimrod-becker @dannyzaken we might want to disable the tests that we know are failing, so we will not ignore the whole suit, and we will actually know if something was broken.

src/sdk/namespace_fs.js
```diff
@@ -333,6 +333,11 @@ class NamespaceFS {
             const file_path = this._get_file_path(params);
             const stat = await fs.promises.stat(file_path);
             console.log(file_path, stat);
+            if (stat.isDirectory()) {
```
guymguym, line 336: @eshelmarc A comment on this condition block will com in handy for future readers. Can you please add a short explanation? Thanks

guymguym:
@eshelmarc please rebase from upstream master.
See here - https://medium.com/@topspinj/how-to-git-rebase-into-a-forked-repo-c9f05e821c8a

eshelmarc:
When using a filesystem to implement object store we can use directories to implement buckets or put object in different directories but the directory name is just part of the object key and by itself the directory is not an object, so when we get a HEAD request with a key that is pointing to a directory we return NoSuchKey (object not found)

guymguym:
Hey sorry if I wasn't clear, but I meant to add a short comment on this new case inside the code itself, just to mention that dirs are not considered an object and therefore we need to error with not found. also need to rebase upstream/master again.

guymguym (approved):
Looks good to me.

However unit-tests are failing.

The postgres tests were already disabled on master so these would not be a problem once you rebase from master.

On the main unit-tests failure - I asked @jeniawhite to try and understand where the failure is, but since the debug prints of the new native fs code is overwhelming the test output, it's hard to understand, so I suggested that Jenia will reduce debug logging using DBG1 DBG2 ... macros. Once that is done on master, and you rebase and push your PR, the tests can re-run and we can see if the failure persists.

Thanks.

src/sdk/namespace_fs.js
guymguym: Resolved in #6418 so this failure will be gone once you rebase from master.

jeniawhite:
@guymguym @eshelmarc 
```2021-03-20T16:50:00.7929401Z   1) namespace_fs
2021-03-20T16:50:00.7933370Z        read_object_md:
2021-03-20T16:50:00.7934481Z      TypeError: stat.isDirectory is not a function
2021-03-20T16:50:00.7935889Z       at NamespaceFS.read_object_md (src/sdk/namespace_fs.js:79:34)
2021-03-20T16:50:00.7937354Z       at async Context.<anonymous> (src/test/unit_tests/test_namespace_fs.js:3:1085)
```
I can see that you've added `.isDirectory()` on the stat response.
The thing is that we do not have that capability in our native code.
As an initial implementation, we've used the response of `stat(2)` (https://man7.org/linux/man-pages/man2/lstat.2.html).

src/sdk/namespace_fs.js
```diff
@@ -346,6 +346,13 @@ class NamespaceFS {
             const file_path = this._get_file_path(params);
             const stat = await nb_native().stat(DEFAULT_FS_CONFIG, file_path);
             console.log(file_path, stat);
+            // directories are not considered objects and therefore we need
+            // to return with error not found.
+            if (stat.isDirectory()) {
```
jeniawhite, line 351: Not supported in the initial commit of the `naitve_fs` module.

jeniawhite (changes requested):

src/sdk/namespace_fs.js
eshelmarc: so for now this code will not work?

jeniawhite: With the current code base, it won't work.
I will now implement the option to query the directory on stat response and push it to the code base.
When it will be in the codebase I will just update you on the changes and you will push your PR as well on top of a new code base.

jeniawhite, line 351: @eshelmarc
```suggestion
            if (isDirectory(stat)) {
```
This should give out the answer if that is a directory or not.
Please make sure to rebase with the master branch, re-run the tests with this addition, and squash the commits prior to merging the PR.

jeniawhite (changes requested):

jeniawhite (approved):

PR #6415, opened 2021-03-17
Using Docker BuildKit as a default

liranmauda:
### Explain the changes
Using Docker BuildKit as a default

- In the Makefile we can pass `DOCKER_BUILDKIT="DOCKER_BUILDKIT=0"` to disable BuildKit
- When we use podman we will not use BuildKit as it is not supported
- The test image will now use the untar Directory of noobaa.


nimrod-becker (approved):

nimrod-becker:
doesn't break builds US/DS?

liranmauda:
> doesn't break builds US/DS?

No, DS build doesn't use any of those files.

PR #6416, opened 2021-03-17
Addition of setting FS config for native

jeniawhite:
### Explain the changes
1. Added the capability to set fs config to the native layer to transfer uid, gid, backend and trigger ops with these values

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -7,12 +7,14 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <fcntl.h>
-#include <unistd.h>
 #include <dirent.h>
 #include <vector>
 #include <math.h>
-// #include <errno.h>
-// #include <sys/fsuid.h>
+#include <unistd.h>
+#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
+#else
+    #include <sys/fsuid.h>
+#endif
 
 namespace noobaa
 {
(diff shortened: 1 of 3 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -38,20 +38,72 @@ static Napi::Value api(const Napi::CallbackInfo& info)
...
+    uid_t _req_gid;
+    std::string _backend;
     int _errno;
     std::string _desc;
 
     FSWorker(const Napi::CallbackInfo& info)
         : AsyncWorker(info.Env())
         , _deferred(Napi::Promise::Deferred::New(info.Env()))
+        , _orig_uid(getuid())
+        , _orig_gid(getgid())
```
guymguym, line 53: why do we need every op to record this? Why not do that globally once in our fs_napi_init function?

src/native/fs/fs_napi.cpp
```diff
@@ -38,20 +38,72 @@ static Napi::Value api(const Napi::CallbackInfo& info)
...
         , _errno(0)
     {
+        Napi::Object _config = info[0].As<Napi::Object>();
```
guymguym, line 56: rename to `config` it's a local var not a class member.

guymguym (changes requested):

src/native/fs/fs_napi.cpp
```diff
@@ -38,20 +38,72 @@ static Napi::Value api(const Napi::CallbackInfo& info)
...
+        _req_gid = _config.Has("gid") ? _config.Get("gid").ToNumber() : _orig_gid;
+        // TODO: Fill the relevant type
+        _backend = _config.Has("backend") ? _config.Get("backend").ToString() : Napi::String::New(info.Env(), "POSIX");
     }
     void Begin(std::string desc)
     {
         _desc = desc;
         LOG("FS::FSWorker::Begin: " << _desc);
     }
+    virtual void Work() {
```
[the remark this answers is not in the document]
jeniawhite: done

guymguym (approved):

PR #6420, opened 2021-03-21
Addition of directory for stats and refactor

jeniawhite:
### Explain the changes
1. Added directory recognition from native layer
2. Refactored the code to be under module FS of the native layer

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -19,6 +19,8 @@
 namespace noobaa
 {
 
+DBG_INIT(0);
+
 struct Entry {
     std::string name;
     ino_t ino;
@@ -123,11 +125,11 @@ struct FSWorker : public Napi::AsyncWorker
     void Begin(std::string desc)
     {
         _desc = desc;
-        LOG("FS::FSWorker::Begin: " << _desc);
+        DBG1("FS::FSWorker::Begin: " << _desc);
     }
     virtual void Work() = 0;
     void Execute() {
-        LOG("FS::FSWorker::Start Execute: " << _desc << 
+        DBG1("FS::FSWorker::Start Execute: " << _desc << 
             " req_uid:" << _req_uid << 
             " req_gid:" << _req_gid << 
             " backend:" << _backend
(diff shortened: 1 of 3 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -810,25 +810,38 @@ Napi::Value DirWrap::read(const Napi::CallbackInfo& info)
     return api<DirReadEntry>(info);
 }
 
+static Napi::Value
+ISDIR(const Napi::CallbackInfo& info)
```
guymguym, line 814: Instead of making a native call every time we want to check a mode,
can we export S_IFMT and S_IFDIR to js land?
Are these constants always defined (perhaps with different values) in Linux and Mac?

guymguym: Like node is doing:
https://github.com/nodejs/node/blob/baa9e652142d0fa02ef21f561ddcdfda6372305f/lib/internal/fs/utils.js#L393-L395
https://github.com/nodejs/node/blob/baa9e652142d0fa02ef21f561ddcdfda6372305f/lib/internal/fs/utils.js#L466-L472

src/native/fs/fs_napi.cpp
```diff
@@ -810,25 +810,38 @@ Napi::Value DirWrap::read(const Napi::CallbackInfo& info)
...
     LOG("FS::fs_napi:" << " orig_uid:" << orig_uid << " orig_gid:" << orig_gid);
-    exports["stat"] = Napi::Function::New(env, api<Stat>);
-    exports["unlink"] = Napi::Function::New(env, api<Unlink>);
-    exports["rename"] = Napi::Function::New(env, api<Rename>); 
-    exports["mkdir"] = Napi::Function::New(env, api<Mkdir>); 
-    exports["rmdir"] = Napi::Function::New(env, api<Rmdir>); 
-    exports["writeFile"] = Napi::Function::New(env, api<Writefile>);
-    exports["readFile"] = Napi::Function::New(env, api<Readfile>);
-    exports["readdir"] = Napi::Function::New(env, api<Readdir>);
+    auto exports_object = Napi::Object::New(env);
```
guymguym, line 824:
```suggestion
    auto fs_obj = Napi::Object::New(env);
```

guymguym (approved):

PR #6421, opened 2021-03-21
Export nsfs bucket and addition of uid and gid properties to account

romayalon:

### Explain the changes
1. addition of uid and gid properties to account api + schema.
2. Added set of uid and gid properties to create_account & update_account.
3. added export_dir_as_bucket to bucket_server + bucket_api.
4. list_buckets in bucket_server - added code that checks if it's an nsfs bucket and calls stat.
5. delete_bucket - added code that checks if it's an nsfs bucket and calls rmdir and delete namespace_resource.
6. added tests - currently will pass locally but not in a container. 
7. added check for fs_backend type in create_namespace_resource.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -70,6 +70,9 @@ module.exports = {
                             },
                         },
                     },
+                    nsfs_account_config: {
+                        $ref: 'common_api#/definitions/nsfs_account_config'
+                    },
                 },
             },
             reply: {
@@ -182,6 +185,9 @@ module.exports = {
                                 $ref: 'common_api#/definitions/ip_range'
                             }
                         }]
+                    },
+                    nsfs_account_config: {
+                        $ref: 'common_api#/definitions/nsfs_account_config'
                     },
                     preferences: {
                         type: 'object',
@@ -647,6 +653,9 @@ module.exports = {
                             enum: ['DARK', 'LIGHT']
                         }
                     }
+                },
+                nsfs_account_config: {
+                    $ref: 'common_api#/definitions/nsfs_account_config'
                 }
             },
         },
--- a/src/api/bucket_api.js
+++ b/src/api/bucket_api.js
@@ -27,24 +27,7 @@ module.exports = {
                     chunk_coder_config: { $ref: 'common_api#/definitions/chunk_coder_config' },
                     tag: { type: 'string' },
                     object_lock_configuration: { $ref: '#/definitions/object_lock_configuration' },
-                    namespace: {
-                        type: 'object',
-                        required: ['write_resource', 'read_resources'],
-                        properties: {
-                            write_resource: {
-                                type: 'string'
-                            },
-                            read_resources: {
-                                type: 'array',
-                                items: {
-                                    type: 'string'
-                                },
-                            },
-                            caching: {
-                                $ref: 'common_api#/definitions/bucket_cache_config'
-                            }
-                        }
-                    },
+                    namespace: { $ref: '#/definitions/namespace_bucket_config' },
                     lock_enabled: {
                         type: 'boolean'
                     },
@@ -919,7 +902,6 @@ module.exports = {
                 system: 'admin'
             },
         },
-
     },
 
     definitions: {
@@ -942,9 +924,7 @@ module.exports = {
                     }
                 },
                 versioning: { $ref: '#/definitions/versioning' },
-                namespace: {
-                    $ref: '#/definitions/bucket_namespace'
-                },
+                namespace: { $ref: '#/definitions/namespace_bucket_config' },
                 bucket_claim: { $ref: '#/definitions/bucket_claim' },
                 tiering: {
                     $ref: 'tiering_policy_api#/definitions/tiering_policy'
@@ -1130,28 +1110,6 @@ module.exports = {
             }
         },
 
-        bucket_namespace: {
-            type: 'object',
-            required: [
-                'read_resources',
-                'write_resource',
-            ],
-            properties: {
-                read_resources: {
-                    type: 'array',
-                    items: {
-                        type: 'string'
-                    }
-                },
-                write_resource: {
-                    type: 'string'
-                },
-                caching: {
-                    $ref: 'common_api#/definitions/bucket_cache_config'
-                }
-            }
-        },
-
         bucket_claim: {
             type: 'object',
             required: ['bucket_class', 'namespace'],
@@ -1194,11 +1152,25 @@ module.exports = {
                         read_resources: {
                             type: 'array',
                             items: {
-                                $ref: 'pool_api#/definitions/namespace_resource_extended_info'
+                                type: 'object',
+                                required: [
+                                    'resource',
+                                ],
+                                properties: {
+                                    resource: { $ref: 'pool_api#/definitions/namespace_resource_extended_info' },
+                                    path: { type: 'string' }
+                                }
                             }
                         },
                         write_resource: {
-                            $ref: 'pool_api#/definitions/namespace_resource_extended_info'
+                            type: 'object',
+                            required: [
+                                'resource',
+                            ],
+                            properties: {
+                                resource: { $ref: 'pool_api#/definitions/namespace_resource_extended_info' },
+                                path: { type: 'string' }
+                            }
                         },
                         caching: {
                             $ref: 'common_api#/definitions/bucket_cache_config'
@@ -1269,21 +1241,7 @@ module.exports = {
                 //         type: 'null'
                 //     }]
                 // },
-                namespace: {
-                    type: 'object',
-                    required: ['write_resource', 'read_resources'],
-                    properties: {
-                        write_resource: {
-                            type: 'string'
-                        },
-                        read_resources: {
-                            type: 'array',
-                            items: {
-                                type: 'string'
-                            },
-                        }
-                    }
-                },
+                namespace: { $ref: '#/definitions/namespace_bucket_config' },
                 versioning: { $ref: '#/definitions/versioning' },
             }
         },
@@ -1451,6 +1409,36 @@ module.exports = {
                             ]
                         }
                     }
+                }
+            }
+        },
+
+        // namespace bucket configuration
+
+        namespace_resource_config: {
+            type: 'object',
+            required: ['resource'],
+            properties: {
+                resource: { type: 'string' },
+                path: { type: 'string' }
+            }
+        },
+
+        namespace_bucket_config: {
+            type: 'object',
+            required: ['write_resource', 'read_resources'],
+            properties: {
+                write_resource: {
+                    $ref: '#/definitions/namespace_resource_config'
+                },
+                read_resources: {
+                    type: 'array',
+                    items: {
+                        $ref: '#/definitions/namespace_resource_config'
+                    },
+                },
+                caching: {
+                    $ref: 'common_api#/definitions/bucket_cache_config'
                 }
             }
         },
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -790,9 +790,7 @@ module.exports = {
             }
         },
 
-        fs_path: {
-            type: 'string'
-        },
+        // nsfs 
 
         fs_backend: {
             type: 'string',
@@ -801,15 +799,23 @@ module.exports = {
 
         nsfs_config: {
             type: 'object',
-            required: ['fs_path'],
+            required: ['fs_root_path'],
             properties: {
-                fs_path: {
-                    $ref: '#/definitions/fs_path'
+                fs_root_path: {
+                    type: 'string'
                 },
                 fs_backend: {
                     $ref: '#/definitions/fs_backend'
                 }
             }
-        }
+        },
+        nsfs_account_config: {
+            type: 'object',
+            properties: {
+                uid: { type: 'number' },
+                gid: { type: 'number' },
+                new_buckets_path: { type: 'string' },
+            }
+        },
     }
 };
--- a/src/api/pool_api.js
+++ b/src/api/pool_api.js
@@ -555,8 +555,8 @@ module.exports = {
                 identity: {
                     $ref: 'common_api#/definitions/access_key'
                 },
-                fs_path: {
-                    $ref: 'common_api#/definitions/fs_path'
+                fs_root_path: {
+                    type: 'string'
                 },
                 fs_backend: {
                     $ref: 'common_api#/definitions/fs_backend'
@@ -599,8 +599,8 @@ module.exports = {
                     type: 'string'
                 },
                 secret_key: { $ref: 'common_api#/definitions/secret_key' },
-                fs_path: {
-                    $ref: 'common_api#/definitions/fs_path'
+                fs_root_path: {
+                    type: 'string'
                 },
                 fs_backend: {
                     $ref: 'common_api#/definitions/fs_backend'
--- a/src/sdk/bucketspace_nb.js
+++ b/src/sdk/bucketspace_nb.js
@@ -1,6 +1,11 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
+const _ = require('lodash');
+const P = require('../util/promise');
+const nb_native = require('../util/nb_native');
+const dbg = require('../util/debug_module')(__filename);
+
 /**
  * @implements {nb.BucketSpace}
  */
@@ -14,8 +19,18 @@ class BucketSpaceNB {
     // BUCKET //
     ////////////
 
-    async list_buckets() {
-        return this.rpc_client.bucket.list_buckets();
+    async list_buckets(object_sdk) {
+        const { buckets } = (await this.rpc_client.bucket.list_buckets());
+        const has_access_buckets = (await P.all(_.map(
+            buckets,
+            async bucket => {
+                const namespace_bucket_config = await object_sdk.read_bucket_sdk_namespace_info(bucket.name.unwrap());
+                if (await this._has_access_to_nsfs_dir(namespace_bucket_config, object_sdk.requesting_account)) {
+                    return bucket;
+                }
+            }
+        ))).filter(bucket => bucket);
+        return { buckets: has_access_buckets };
     }
 
     async read_bucket(params) {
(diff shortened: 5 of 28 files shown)


src/server/system_services/bucket_server.js
```diff
@@ -1809,6 +1846,59 @@ async function put_object_lock_configuration(req) {
...
 }
 
+function _is_bucket_nsfs_exported(bucket) {
+    return bucket.namespace && bucket.namespace.write_resource &&
+        bucket.namespace.write_resource.nsfs_config;
+}
+
+// 1. create namespace resource 
+// 2. create namespace bucket
+async function export_dir_as_bucket(req) {
```
guymguym, line 1856: I would avoid adding this new api which just wraps the two calls specific to nsfs - the creation of a namespace resource + namespace bucket is a general problem that we created by having namespace resource represent a specific single bucket/dir. I think that what we should focus on doing is to allow namespace resources to be non-specific and only define the base fs_path (or base s3/blob account) and then namespace buckets will be able to reuse those resources just like we do for pools.

Adding more automation to create/delete the resource automatically with the bucket is just digging the hole deeper for these redundant configuration objects.

guymguym (changes requested):
I have described few doubts about how we want to model these new capabilities. I suggest we discuss further to design this. Sorry for the hassle but I want to make sure we don't add more and more code that will make it harder to adapt to our needs.

guymguym:
@romayalon ready for review?

romayalon:
> @romayalon ready for review?

yes, thank you.

src/test/utils/bucket_functions.js
```diff
@@ -175,8 +176,8 @@ class BucketFunctions {
             await this._client.bucket.update_bucket({
                 name,
                 namespace: {
-                    read_resources,
-                    write_resource
+                    read_resources: read_resources.map(function(rr) { return { resource: rr }; }),
```
guymguym, line 179:
```suggestion
                    read_resources: read_resources.map(rr => ({ resource: rr })),
```

src/api/common_api.js
```diff
@@ -810,6 +812,44 @@ module.exports = {
                     $ref: '#/definitions/fs_backend'
                 }
             }
-        }
+        },
+        nsfs_account_config: {
+            type: 'object',
+            properties: {
+                uid: { type: 'string' },
+                gid: { type: 'string' },
```
guymguym, line 820: @jeniawhite @romayalon I think this should be `type: 'number'`.
Any reason this should be a string?
Even with number we can check wether it's undefined vs. 0 if we need to support uid/gid of root.

src/api/common_api.js
```diff
@@ -790,6 +790,8 @@ module.exports = {
             }
         },
 
+        // nsfs 
+
         fs_path: {
```
guymguym, line 795: Seems like we can get rid of this schema type and just use type string everywhere we need.

src/api/common_api.js
```diff
@@ -810,6 +812,44 @@ module.exports = {
...
+        },
+
+        // namespace bucket configuration
+
+        namespace_resource_config: {
+            type: 'object',
+            required: ['resource'],
+            properties: {
+                resource: { type: 'string' },
+                fs_dir_path: { type: 'string' },
```
guymguym, line 832: rename to `path`

src/api/common_api.js
```diff
@@ -810,6 +812,44 @@ module.exports = {
...
+            }
+        },
+
+        namespace_bucket_config: {
```
guymguym, line 836: move to be local definition in bucket_api

src/api/bucket_api.js
```diff
@@ -1194,11 +1152,25 @@ module.exports = {
...
                             type: 'array',
                             items: {
-                                $ref: 'pool_api#/definitions/namespace_resource_extended_info'
+                                type: 'object',
+                                required: [
+                                    'resource',
+                                ],
+                                properties: {
+                                    resource: { $ref: 'pool_api#/definitions/namespace_resource_extended_info' },
+                                    fs_dir_path: { $ref: 'common_api#/definitions/fs_path' }
```
guymguym, line 1161: rename to `path`

src/api/bucket_api.js
```diff
@@ -1194,11 +1152,25 @@ module.exports = {
...
                         },
                         write_resource: {
-                            $ref: 'pool_api#/definitions/namespace_resource_extended_info'
+                            type: 'object',
+                            required: [
+                                'resource',
+                            ],
+                            properties: {
+                                resource: { $ref: 'pool_api#/definitions/namespace_resource_extended_info' },
+                                fs_dir_path: { $ref: 'common_api#/definitions/fs_path' }
```
guymguym, line 1172: `path`

src/api/bucket_api.js
```diff
@@ -1273,14 +1245,10 @@ module.exports = {
                     type: 'object',
                     required: ['write_resource', 'read_resources'],
                     properties: {
-                        write_resource: {
-                            type: 'string'
-                        },
+                        write_resource: { $ref: 'common_api#/definitions/namespace_resource_config' },
```
guymguym, line 1248: ref entire namespace field to namespace_bucket_config

src/server/system_services/schemas/bucket_schema.js
```diff
@@ -95,11 +95,25 @@ module.exports = {
...
                     type: 'array',
                     items: {
-                        objectid: true // namespace resource id
+                        type: 'object',
+                        required: ['resource'],
+                        properties: {
+                            resource: {
+                                objectid: true // namespace resource id
+                            },
+                            fs_dir_path: { type: 'string' },
```
guymguym, line 104: `path`

src/server/system_services/schemas/bucket_schema.js
```diff
@@ -95,11 +95,25 @@ module.exports = {
...
                 },
                 write_resource: {
-                    objectid: true // namespace resource id
+                    type: 'object',
+                    required: ['resource'],
+                    properties: {
+                        resource: {
+                            objectid: true // namespace resource id
+                        },
+                        fs_dir_path: { type: 'string' },
```
guymguym, line 115: `path`

src/sdk/nb.d.ts
```diff
@@ -197,6 +197,7 @@ interface NamespaceResource {
     system: System;
     account: Account;
     connection: object;
+    fs_dir_path: string;
```
guymguym, line 200: `path`

src/test/unit_tests/test_ns_list_objects.js
```diff
@@ -67,7 +67,8 @@ function test_ns_list_objects(ns, object_sdk, bucket) {
             const r = await ns.list_objects({
                 bucket,
                 delimiter: '/',
-                prefix: 'folder'
+                prefix: 'folder',
+                path: dir_path
```
guymguym, line 71: Why is `path` now in the params to all of the ns operations?
I didn't see it coming...

src/server/system_services/pool_server.js
```diff
@@ -257,6 +257,11 @@ async function create_namespace_resource(req) {
             dbg.error(`fs path ${already_used_by.nsfs_config.fs_path} already exported by ${already_used_by.name}`);
             throw new RpcError('IN_USE', 'Target already in use');
         }
+        const backend_values = ['CEPH_FS', 'NSFv4', 'GPFS'];
```
guymguym, line 260: I don't think that's necessary - json schema will do that for you anyway because you defined fs_backend as enum with these options...

src/sdk/namespace_fs.js
```diff
@@ -654,7 +655,7 @@ class NamespaceFS {
         try {
             await this._load_bucket(params);
             for (const { key } of params.objects) {
-                const file_path = this._get_file_path({ key });
+                const file_path = this._get_file_path({ path: params.path, key });
```
guymguym, line 658: shouldn't the path come from the bucket config?

src/api/account_api.js
```diff
@@ -70,6 +70,12 @@ module.exports = {
                             },
                         },
                     },
+                    UID: {
+                        $ref: 'common_api#/definitions/UID'
+                    },
+                    GID: {
+                        $ref: 'common_api#/definitions/GID'
+                    },
```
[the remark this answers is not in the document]
romayalon: fixed.

src/api/bucket_api.js
```diff
@@ -920,6 +920,21 @@ module.exports = {
...
 
+        export_dir_as_bucket: {
+            method: 'PUT',
+            params: {
+                type: 'object',
+                required: ['bucket_name', 'fs_path'],
+                properties: {
+                    bucket_name: { $ref: 'common_api#/definitions/bucket_name' },
+                    fs_path: { $ref: 'common_api#/definitions/fs_path' },
+                    fs_backend: { $ref: 'common_api#/definitions/fs_backend' },
```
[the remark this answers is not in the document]
romayalon: fixed.

src/server/system_services/bucket_server.js
```diff
@@ -850,6 +851,31 @@ async function delete_bucket(req) {
...
+                name: write_resource.name
+            }, {
+                auth_token: req.auth_token
+            });
+        } catch (err) {
+            dbg.log0('delete_bucket: failed deleting namespace resource', err);
+            throw err;
+        }
+        try {
+            await nb_native().rmdir({
```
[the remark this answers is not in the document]
romayalon: removed.

src/server/system_services/bucket_server.js
```diff
@@ -893,10 +919,21 @@ function delete_bucket_lifecycle(req) {
...
  */
-function list_buckets(req) {
+async function list_buckets(req) {
+
+    let buckets_by_name_responses = {};
+    await P.all(_.map(
+        req.system.buckets_by_name,
+        async bucket => {
+            buckets_by_name_responses[bucket.name] = !_is_bucket_nsfs_exported(bucket) ||
+                await _has_access_to_nsfs_dir(bucket, req.account);
```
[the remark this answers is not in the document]
romayalon: moved it.

src/server/system_services/bucket_server.js
romayalon: removed.

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -9,16 +10,21 @@ const fs_utils = require('../../util/fs_utils');
 const NamespaceFS = require('../../sdk/namespace_fs');
 const buffer_utils = require('../../util/buffer_utils');
 const test_ns_list_objects = require('./test_ns_list_objects');
+const coretest = require('./coretest');
+//const { rpc_client, EMAIL, PASSWORD, SYSTEM} = coretest;
+// const AWS = require('aws-sdk');
+// const http = require('http');
+coretest.setup({ pools_to_create: [coretest.POOL_LIST[0]] });
```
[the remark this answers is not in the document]
romayalon: seperated.

src/test/utils/bucket_functions.js
romayalon: fixed.

src/api/common_api.js
romayalon: removed.

src/test/unit_tests/test_ns_list_objects.js
romayalon: it's the directory path and it's not saved on the namespace resource, there are other solutions I thought about, we can do a discussion which is preferred.

src/api/pool_api.js
```diff
@@ -556,7 +556,7 @@ module.exports = {
                     $ref: 'common_api#/definitions/access_key'
                 },
                 fs_path: {
```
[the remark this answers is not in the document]
romayalon: I think it's too confusing, maybe we should rename it to root_path, WDYT?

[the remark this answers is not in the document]
guymguym: I think `path` is OK in this context, and I don't think that `root_path` would clarify it.
Lets start by thinking how would this be confusing - what might `path` represent in this context that might be wrong compared to what we mean...
We might claim that it's not clear to what this path applies, whether it applies to the bucket that uses it or the target bucket, and in that case we can name it `target_path`.

src/server/system_services/pool_server.js
romayalon: sure, but I wanted to throw an error that is better understood than INVALID_SCHEMA_DB

guymguym: But your code will not even get executed. I don't think it's needed. 
Bad schema is easy to troubleshoot as it is, and if not we should make it so.

src/api/pool_api.js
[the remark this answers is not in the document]
romayalon: this fs_path is applied to the file system root and not to the actual directory, so changing all fs_path occurrences to `path`, might be a bit confusing with the path to the directory attached in the namespace_bucket config.

src/api/common_api.js
jeniawhite: @guymguym Agree, it should be a Number.
I do not see a reason for it to be a String.

src/server/system_services/pool_server.js
romayalon: ok deleted

src/sdk/namespace_fs.js
romayalon: according to our meeting, can I resolve this comment?

src/test/unit_tests/test_ns_list_objects.js
romayalon: according to our meeting, can I resolve this comment?

src/server/system_services/bucket_server.js
```diff
@@ -575,21 +582,24 @@ function get_bucket_changes_namespace(req, bucket, update_request, single_bucket
...
     }
     _.set(single_bucket_update, 'namespace.read_resources', read_resources);
-    const wr_obj = req.system.namespace_resources_by_name[update_request.namespace.write_resource];
-    const write_resource = wr_obj && wr_obj._id;
+    const wr_obj = req.system.namespace_resources_by_name[update_request.namespace.write_resource.resource];
+    const write_resource = wr_obj && { resource: wr_obj._id, path: update_request.namespace.write_resource.path };
     if (!write_resource) throw new RpcError('INVALID_WRITE_RESOURCES');
     _.set(single_bucket_update, 'namespace.write_resource', write_resource);
-    if (!_.includes(update_request.namespace.read_resources, update_request.namespace.write_resource)) {
+    if (!_.find(update_request.namespace.read_resources, update_request.namespace.write_resource)) {
```
guymguym, line 597: it's hard to notice that this works ok because _.find really knows how to match objects correctly. Maybe add a comment to explain?

src/sdk/object_sdk.js
```diff
@@ -188,9 +208,9 @@ class ObjectSDK {
                         valid_until: time + config.OBJECT_SDK_BUCKET_CACHE_EXPIRY_MS,
                     };
                 }
-                if (bucket.namespace.write_resource && bucket.namespace.write_resource.fs_path) {
+                if (bucket.namespace.write_resource && bucket.namespace.write_resource.resource.fs_path) {
```
guymguym, line 211: why are we checking that fs_path is not empty here?

src/sdk/object_sdk.js
```diff
@@ -71,8 +72,13 @@ class ObjectSDK {
...
 
     async _get_bucket_namespace(name) {
-        const { ns } = await bucket_namespace_cache.get_with_cache({ sdk: this, name });
-        return ns;
+        return bucket_namespace_cache.get_with_cache({ sdk: this, name });
+    }
+
+    set_namespace_bucket_path(params, bucket) {
+        if (bucket && bucket.namespace && bucket.namespace.write_resource.path) {
+            params.path = bucket.namespace.write_resource.path;
```
guymguym, line 80: I don't understand - isn't this the same as the value we are keeping in namespace_fs this.fs_path? Why do you need to set it on the params too??

src/sdk/namespace_fs.js
guymguym: I still don't feel comfortable with this approach of passing the path on every param when this is a property of the namespace itself...

guymguym: And I am still not sure why this.fs_path is not exactly the same as the params.path you used

guymguym (changes requested):

src/sdk/object_sdk.js
guymguym: OK so I see now, the path from inside the resource vs the path that each bucket sets to use from the resource.
But both are still properties of the namespace, so we should initialize the namespace this.fs_path to already join these two together.

src/api/pool_api.js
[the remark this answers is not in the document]
guymguym: ok, but fs_path is confusing vs. the namespace_fs this.fs_path member which I think should have a different meaning - which is the joined fs_path of the resource, and the path of the bucket namespace policy. so perhaps we can rename that this.fs_path in namespace_fs to something else say this.ns_path or something ?

src/sdk/bucketspace_nb.js
```diff
@@ -14,8 +19,23 @@ class BucketSpaceNB {
...
-    async list_buckets() {
-        return this.rpc_client.bucket.list_buckets();
+    async list_buckets(object_sdk) {
+        const buckets = (await this.rpc_client.bucket.list_buckets()).buckets;
+        let buckets_by_name_responses = {};
+        await P.all(_.map(
+            buckets,
+            async bucket => {
+                // TODO: can be more efficient, should we change list_buckets to return namespace info?
+                const bucket_info = (await this.rpc_client.bucket.read_bucket({name: bucket.name.unwrap()}));
```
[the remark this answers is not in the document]
romayalon: fixed.

src/sdk/bucketspace_nb.js
```diff
@@ -14,8 +19,23 @@ class BucketSpaceNB {
     // BUCKET //
     ////////////
 
-    async list_buckets() {
-        return this.rpc_client.bucket.list_buckets();
+    async list_buckets(object_sdk) {
+        const buckets = (await this.rpc_client.bucket.list_buckets()).buckets;
+        let buckets_by_name_responses = {};
+        await P.all(_.map(
```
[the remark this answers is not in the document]
romayalon: fixed.

src/sdk/object_sdk.js
romayalon: fixed.

src/api/pool_api.js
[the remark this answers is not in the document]
romayalon: fixed.

src/sdk/object_sdk.js
```diff
@@ -117,33 +119,56 @@ class ObjectSDK {
...
+            const bucket_allowed = _.get(this.requesting_account, 'allowed_buckets.full_permission', false) ||
+                _.find(
+                    _.get(this.requesting_account, 'allowed_buckets.permission_list') || [],
+                    name => name.unwrap() === bucket
+                );
+            if (!bucket_allowed) throw new RpcError('UNAUTHORIZED', `No permission to access bucket`);
+        }
+    }
+
+    async set_requesting_account() {
```
guymguym, line 160: Is this still needed? I don't see where we call it now

guymguym (approved):

PR #6423, opened 2021-03-25
fixed delete_object_version to also find and delete uncompleted multipart uploads 

romayalon:

### Explain the changes
1. when calling delete_object_version, we used to findOne while using upload_started: null and 'null_version_index' index which filters entries contains upload_started property. this fact caused an infinite loop when deleting a bucket (and it's objects, for example when deleting OBC) that has uncompleted multipart upload attached to it. Added another function that doesn't use filter uncompleted multipart uploads. 

1. BZ #1940476

### Testing Instructions:
1.

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -185,6 +185,19 @@ class MDStore {
         });
     }
 
+    async find_object_or_upload_null_version(bucket_id, key) {
+        return this._objects.findOne({
+                // index fields:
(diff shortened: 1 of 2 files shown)


dannyzaken (commented):
LGTM, but I guess @jeniawhite can have more input on that.

see if it's not a big effort to add unit tests in test_md_store. this will help us validate it for both postgres and mongo

liranmauda (approved):
LGTM

PR #6438, opened 2021-04-06
Refactor uploads and reads and add thread logging

jeniawhite:
### Explain the changes
1. Added thread_id logging
2. Refactored read and write to use our native instead of streams

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -11,11 +11,39 @@
 #include <vector>
 #include <math.h>
 #include <unistd.h>
+#include <map>
+#include <thread>
 
 namespace noobaa
 {
 
 DBG_INIT(0);
+
+const static std::map<std::string,int> flags_to_case = {
+    {"r",O_RDONLY},
+    {"rs",O_RDONLY | O_SYNC},
+    {"sr",O_RDONLY | O_SYNC}, 
+    {"r+",O_RDWR},
+    {"rs+",O_RDWR | O_SYNC},
+    {"sr+",O_RDWR | O_SYNC},
+    {"w",O_TRUNC | O_CREAT | O_WRONLY},
+    {"wx",O_TRUNC | O_CREAT | O_WRONLY | O_EXCL},
+    {"xw",O_TRUNC | O_CREAT | O_WRONLY | O_EXCL},
+    {"w+",O_TRUNC | O_CREAT | O_RDWR},
+    {"wx+",O_TRUNC | O_CREAT | O_RDWR | O_EXCL},
+    {"xw+",O_TRUNC | O_CREAT | O_RDWR | O_EXCL},
+    {"a",O_APPEND | O_CREAT | O_WRONLY},
+    {"ax",O_APPEND | O_CREAT | O_WRONLY | O_EXCL},
+    {"xa",O_APPEND | O_CREAT | O_WRONLY | O_EXCL},
+    {"as",O_APPEND | O_CREAT | O_WRONLY | O_SYNC},
+    {"sa",O_APPEND | O_CREAT | O_WRONLY | O_SYNC},
+    {"a+",O_APPEND | O_CREAT | O_RDWR},
+    {"ax+",O_APPEND | O_CREAT | O_RDWR | O_EXCL},
+    {"xa+",O_APPEND | O_CREAT | O_RDWR | O_EXCL},
+    {"as+",O_APPEND | O_CREAT | O_RDWR | O_SYNC},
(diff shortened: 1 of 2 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -563,30 +639,45 @@ struct FileRead : public FSWorker
...
+        if (_bw < 0) {
+            SetSyscallError();
+        } else if ((size_t)_bw != _len) {
+            SetError(XSTR() << "FS::FileWrite::Execute: partial write error " << DVAL(_bw) << DVAL(_len));
+        }
+    }
+    virtual void OnOK()
+    {
+        Napi::Env env = Env();
+        _deferred.Resolve(Napi::Number::New(env, _bw));
```
guymguym, line 678: Why do we return the number of bytes if this call will always fail unless the number is exactly the same as the buffer length? Seems like it would be a simpler API to just remove this return value.

src/native/fs/fs_napi.cpp
```diff
@@ -480,20 +544,31 @@ struct FileOpen : public FSWorker
...
+        , _flags(0)
+        , _mode(0666)
     {
         _path = info[1].As<Napi::String>();
-        // TODO - info[1] { mode, readonly }
+        if (info.Length() > 2 && !info[2].IsUndefined()) {
+            _flags = _stringToFlags(info[2].As<Napi::String>());
+        }
+        if (info.Length() > 3 && !info[3].IsUndefined()) {
+            _mode = unsigned(info[3].As<Napi::Number>());
```
guymguym, line 560: I think you should use `info[3].As<Napi::Number>().Uint32Value()`

https://github.com/nodejs/node-addon-api/blob/main/doc/number.md#uint32value

src/native/fs/fs_napi.cpp
```diff
@@ -11,12 +11,69 @@
...
+    {"ax",13},
+    {"xa",14},
+    {"as",15},
+    {"sa",16},
+    {"a+",17},
+    {"ax+",18},
+    {"xa+",19},
+    {"as+",20},
+    {"sa+",21}
+};
```
guymguym, line 45: Why not map directly to the flags value? why do we need this intermediate index values:
```
{"r", O_RDONLY},
{"rs", O_RDONLY | O_SYNC},
{"sr", O_RDONLY | O_SYNC},
{"r+", O_RDWR},
...
```

src/sdk/namespace_fs.js
```diff
@@ -574,35 +580,51 @@ class NamespaceFS {
...
+        let read_file;
+        let write_file;
         try {
             const { multiparts = [] } = params;
             multiparts.sort((a, b) => a.num - b.num);
             await this._load_multipart(params);
             const file_path = this._get_file_path(params);
             const upload_path = path.join(params.mpu_path, 'final');
-            const upload_stream = fs.createWriteStream(upload_path);
+            write_file = await nb_native().fs.open(DEFAULT_FS_CONFIG, upload_path, 'a');
```
guymguym, line 591: Why "a" ??? we do not want to keep appending to existing file in any case.
I would open as "w" or "wx".

src/sdk/namespace_fs.js
```diff
@@ -496,15 +495,22 @@ class NamespaceFS {
...
-            .once('error', reject)
-            .pipe(
-                fs.createWriteStream(upload_path, write_options)
-                .once('error', reject)
-                .once('finish', resolve)
-            )
-        );
+        let target_file;
+        try {
+            target_file = await nb_native().fs.open(DEFAULT_FS_CONFIG, upload_path, 'a');
```
guymguym, line 500: Why "a" ??? we do not want to keep appending to existing file in any case.
I would open as "w" or "wx".

guymguym (changes requested):

src/sdk/namespace_fs.js
```diff
@@ -496,15 +495,22 @@ class NamespaceFS {
...
+            for await (const data of source_stream) {
+                await target_file.write(DEFAULT_FS_CONFIG, data);
+            }
```
guymguym, line 503: We need to think a bit on how to optimize this write flow.
The source_stream is typically the HTTP request (although it might be a copy object stream too).
This means that the data chunks we will get from that stream might be quite small.
You can print data.length and see for yourself.
In the case of small buffers we would want to collect up to `config.NSFS_BUF_SIZE` bytes and submit a single writev call.
https://linux.die.net/man/2/writev

jeniawhite: Should it be done in this PR?

guymguym (approved):

PR #6442, opened 2021-04-08
Optimization for uploads and fix for mkdir permissions

jeniawhite:
### Explain the changes
1. Added writev to native
2. Optimize uploads to use writev
3. Added permissions to mkdir

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -13,6 +13,7 @@
 #include <unistd.h>
 #include <map>
 #include <thread>
+#include <sys/uio.h>
 
 namespace noobaa
 {
@@ -251,14 +252,18 @@ struct Unlink : public FSWorker
 struct Mkdir : public FSWorker
 {
     std::string _path;
-    Mkdir(const Napi::CallbackInfo& info) : FSWorker(info)
+    int _mode;
+    Mkdir(const Napi::CallbackInfo& info) 
+        : FSWorker(info)
+        , _mode(0777)
     {
         _path = info[1].As<Napi::String>();
+        _mode = info[2].As<Napi::Number>();
         Begin(XSTR() << DVAL(_path));
     }
     virtual void Work()
     {
-        int r = mkdir(_path.c_str(), S_IRWXU);
+        int r = mkdir(_path.c_str(), _mode);
         if (r == -1) SetSyscallError();
     }
 };
@@ -490,6 +495,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
           InstanceMethod("close", &FileWrap::close),
           InstanceMethod("read", &FileWrap::read),
           InstanceMethod("write", &FileWrap::write),
+          InstanceMethod("writev", &FileWrap::writev),
         });
         constructor = Napi::Persistent(func);
         constructor.SuppressDestruct();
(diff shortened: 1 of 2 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -254,11 +255,12 @@ struct Mkdir : public FSWorker
     Mkdir(const Napi::CallbackInfo& info) : FSWorker(info)
     {
         _path = info[1].As<Napi::String>();
+        // TODO: Handle mode
         Begin(XSTR() << DVAL(_path));
     }
     virtual void Work()
     {
-        int r = mkdir(_path.c_str(), S_IRWXU);
+        int r = mkdir(_path.c_str(), S_IRWXU | S_IRWXG | S_IRWXO);
```
guymguym, line 263: Are these all the 3 sticky bits for User/Group/Other ?
Can you add a comment to explain why we are using this mode?

src/native/fs/fs_napi.cpp
```diff
@@ -646,6 +650,43 @@ struct FileWrite : public FSWorker
...
+    FileWrap *_wrap;
+    std::vector<struct iovec> iov_vec;
+    ssize_t _bw;
+    FileWritev(const Napi::CallbackInfo& info) 
+        : FSWorker(info)
+        , _bw(0)
+    {
+        _wrap = FileWrap::Unwrap(info.This().As<Napi::Object>());
+        auto buffers = info[1].As<Napi::Array>();
+        const int buffers_len = buffers.Length();
```
guymguym, line 664: Since the length is known, I would resize the vector to the length here, and then directly update the elements instead of push_back one by one potentially causing dynamic allocations and reallocations...

src/native/fs/fs_napi.cpp
```diff
@@ -646,6 +650,43 @@ struct FileWrite : public FSWorker
 };
 
 
+struct FileWritev : public FSWorker
+{
+    FileWrap *_wrap;
+    std::vector<struct iovec> iov_vec;
+    ssize_t _bw;
```
guymguym, line 657: seems like this is not needed to be a member and can move to be local in Work()

src/sdk/namespace_fs.js
```diff
@@ -494,13 +494,30 @@ class NamespaceFS {
...
         let target_file;
+        let q_buffers = [];
+        let q_size = 0;
+        const flush = async () => {
+            if (q_buffers.length) {
+                await target_file.writev(DEFAULT_FS_CONFIG, q_buffers);
+                q_buffers = [];
+                q_size = 0;
+            }
+        };
```
guymguym, line 507: I suggest to insert this code into the try block right below the `target_file = ...` line.

guymguym (commented):
looks good, few small comments

src/native/fs/fs_napi.cpp
```diff
@@ -646,6 +652,41 @@ struct FileWrite : public FSWorker
...
+    }
+    virtual void Work()
+    {
+        int fd = _wrap->_fd;
+        std::string path = _wrap->_path;
+        if (fd < 0) {
+            SetError(XSTR() << "FS::FileWritev::Execute: ERROR not opened " << path);
+            return;
+        }
+        ssize_t _bw = writev(fd, iov_vec.data(), iov_vec.size());
```
guymguym, line 682: rename `_bw` to `bw`

src/sdk/namespace_fs.js
```diff
@@ -494,13 +494,30 @@ class NamespaceFS {
         }
     }
 
-    async _upload_stream(source_stream, upload_path, write_options) {
+    async _upload_stream(source_stream, upload_path) {
         let target_file;
+        let q_buffers = [];
+        let q_size = 0;
```
guymguym, line 500: move into try-catch

guymguym (approved):
Two minor comments, otherwise looks good, approving

PR #6443, opened 2021-04-11
Bumped version to 5.8 and updated node modules versions

liranmauda:
### Explain the changes
- Bumped version to 5.8.0
- Bumped sinon and move it to be dev

- googleapis 66.0.0 -> 70.0.0

googleapis 66.0.0 -> 70.0.0 is updating:
1. google-auth-library 6.1.6 -> 7.0.4
2. googleapis-common 4.4.3 -> 5.0.2

both have breaking changes that seems not to affect us

### Gap:
1.  Not migrating ajv from 6 to 8. In order to migrate, we need to follow: [v6-to-v8-migration](https://github.com/ajv-validator/ajv/blob/master/docs/v6-to-v8-migration.md)


frontend/package.json
```diff
@@ -9,13 +9,13 @@
         "numeral": "2.0.6"
     },
     "devDependencies": {
-        "@babel/cli": "7.12.10",
-        "@babel/core": "7.12.10",
-        "@babel/plugin-proposal-class-properties": "7.12.1",
+        "@babel/cli": "7.13.14",
```
nimrod-becker, line 12: please run the UI just to verify sanity

frontend/package.json
```diff
@@ -24,10 +24,10 @@
         "babel-plugin-transform-runtime": "6.23.0",
         "babel-register": "6.26.0",
         "babelify": "10.0.0",
-        "bower": "1.8.8",
+        "bower": "1.8.12",
         "browserify": "17.0.0",
         "del": "6.0.0",
-        "eslint": "7.15.0",
+        "eslint": "7.23.0",
```
nimrod-becker, line 30: any new rules we should fix for?

package.json
```diff
@@ -60,16 +60,16 @@
     "./config-local": false
   },
   "dependencies": {
-    "@google-cloud/storage": "5.7.0",
```
nimrod-becker, line 63: please verify cloud resource AWS and Azure, we've seen API breaks on new versions

frontend/package.json
liranmauda: UI is running.

package.json
liranmauda: There are no noticeable breaking changes from 5.7.0 to 5.8.3

nimrod-becker (approved):

PR #6444, opened 2021-04-11
standalone core reuse start endpoint

guymguym:

### Explain the changes
1. Reuse start endpoint code (besides rpc setup)
2. Standalone core now starts both http and https ports.
3. Added optional env `DISABLE_INIT_RANDOM_SEED=true` to disable the call to init_random_seed in nb_native which currently fails in standalone mode with `TypeError: fh.read is not a function` or `TypeError: fh.close is not a function` due to incompatibility of the return type of `fs.promises.open()` in pkg.
4. Add the nb_native.node and other libraries as assets so that the binary is portable on its own.

1. NA

### Testing Instructions:
1. `npm run build:core`
2. `DISABLE_INIT_RANDOM_SEED=true ./build/noobaa-core-macos nsfs .`

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -12952,7 +12952,8 @@
         "y18n": {
             "version": "3.2.2",
             "resolved": "https://registry.npmjs.org/y18n/-/y18n-3.2.2.tgz",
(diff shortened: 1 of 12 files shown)


nimrod-becker (commented):
test

dannyzaken (commented):
test

jeniawhite (approved):

PR #6447, opened 2021-04-12
Mergify: configuration update

nimrod-becker:
This change has been made by @nimrod-becker from https://mergify.io config editor.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,8 @@
+pull_request_rules:
+  - name: Automatic merge on approval
+    conditions:
+      - "#approved-reviews-by>=1"
+      - label!="WIP PR"
+    actions:
+      merge:
+        method: merge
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -0,0 +1,8 @@
+pull_request_rules:
+  - name: Automatic merge on approval
+    conditions:
+      - "#approved-reviews-by>=1"
+      - label!="WIP PR"
```
dannyzaken, line 5: add a condition that the title is not containing [WIP]. this way also contributors. that cannot add labels are able to block the merge 

I think it's something like this
```suggestion
    conditions:
      - "#approved-reviews-by>=1"
      - label!="WIP PR"
      - -title~=\[WIP\]
```

.mergify.yml
```diff
@@ -0,0 +1,9 @@
+pull_request_rules:
+  - name: Automatic merge on approval
+    conditions:
+      - "#approved-reviews-by>=1"
```
nimrod-becker, line 4:
```suggestion
      - "#approved-reviews-by>=2"
```

dannyzaken (approved):

PR #6449, opened 2021-04-13
Add copy object to nsfs

### LINKED ISSUE #6424: NSFS support for copy-object
### Environment info
- NooBaa Version: **5.9**
- Platform: **NA**

### Actual behavior
1. Copy-object is not handled for NSFS

### Expected behavior
1. Should handle as optimal as possible - best to just create a new filesystem link to existing file.
2. If link is not possible, copy the data through the endpoint.

liranmauda:
### Explain the changes
- Add copy object to nsfs
- Add `link`, `linkat` and `unlinkat` ops
- Add unit test

1. Fixes: #6424

### Testing Instructions:
1. Run copyObject on nsfs bucket

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -142,9 +142,11 @@ class NamespaceFS {
         return bucket;
     }
 
-    is_server_side_copy(other) {
-        // in noobaa namespace case just check that other is also same fs
-        return other instanceof NamespaceFS && other.bucket_path === this.bucket_path;
+    is_server_side_copy(other, params) {
+        return other instanceof NamespaceFS &&
+            other.bucket_path === this.bucket_path &&
+            other.fs_backend === this.fs_backend && //Check that the same backend type
+            params.xattr_copy; // TODO, DO we need to hard link at TaggingDirective 'REPLACE'?
     }
 
     /////////////////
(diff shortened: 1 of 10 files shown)


liranmauda:
```
$ ll /nsfs
total 24
drwxrwxrwx  4 root  root  120 Apr 22 07:22 .
drwxr-xr-x 19 root  root  500 Apr 21 16:52 ..
drwxr-xr-x  3 10001 root   60 Apr 21 16:57 .noobaa-nsfs
-rw-r--r--  2 10001 root 8334 Apr 22 07:05 a_hard_ling_for_the_png_file
-rw-r--r--  2 10001 root 8334 Apr 22 07:05 a_png_file
drwxrwxrwx  2 root  root   40 Apr 21 16:52 nsfs
```
```
$ stat /nsfs/a_hard_ling_for_the_png_file
  File: /nsfs/a_hard_ling_for_the_png_file
  Size: 8334      	Blocks: 24         IO Block: 4096   regular file
Device: 10h/16d	Inode: 4040273     Links: 2
Access: (0644/-rw-r--r--)  Uid: (10001/ UNKNOWN)   Gid: (    0/    root)
Access: 2021-04-22 07:06:45.856864436 +0000
Modify: 2021-04-22 07:05:09.919849567 +0000
Change: 2021-04-22 07:22:18.258437908 +0000
 Birth: -
```
```
$ stat /nsfs/a_png_file
  File: /nsfs/a_png_file
  Size: 8334      	Blocks: 24         IO Block: 4096   regular file
Device: 10h/16d	Inode: 4040273     Links: 2
Access: (0644/-rw-r--r--)  Uid: (10001/ UNKNOWN)   Gid: (    0/    root)
Access: 2021-04-22 07:06:45.856864436 +0000
Modify: 2021-04-22 07:05:09.919849567 +0000
Change: 2021-04-22 07:22:18.258437908 +0000
 Birth: -
```

liranmauda:
Open questions:
1. In the event of `[Error: File exists] { code: 'EEXIST' }` the currant flow will fail on `link` and will do a copy, so it will override the linked file and create a copy file with the same name and contact, just not linked. Is this what we want?
2. We are currently using `link` and not `linkat`. for `linkat` we need `fd`. Do we want to get the `fd` inside the `fs_napi.cpp` or do we want it to be an argument?
@guymguym @jeniawhite your thoughts?

guymguym:
@liranmauda To handle the atomicity of put and the fact that link will not replace the target file if exists - I think we can first create the link as a temp name in the internal dir, and only then move over to the target name, and since move will replace the target name, that will handle the replace-object case. WDYT?

src/sdk/namespace_fs.js
```diff
@@ -492,24 +492,93 @@ class NamespaceFS {
...
+            } catch (err) {
+                console.warn('NamespaceFS: upload_object - copy_source file close error', err);
+            }
+        }
+    }
+
+    //Checking if we can creat a link 
+    async _can_create_link(other) { //LMLM what is other? 
+        //Add here more conditions
+        if (this.is_same_namespace && other.fs_backend === this.fs_backend) {
```
guymguym, line 578: did you mean to call is_same_namespace()?

guymguym: Also why isn't the compare of fs_backend added into is_same_namespace?

liranmauda: @guymguym you are correct, 
I wanted this function at first to add some checks like `File exists` or any other that we will think of.
I could think of a few:
- is same namespace types (which are and should be in `is_same_namespace`)
- are POSIX calls, such as `File exists` and any other restrictions that prevent a link (maybe "No such file or directory"), but calling the ops might be expensive so we should just try to do link and failback to normal copy if we are not succeeding.
- Copy requests such as  MetadataDirective 'COPY' or maybe TaggingDirective 'COPY'.

I think that the `_can_create_link` function is not needed as we fall back to copy on fail.

liranmauda:
> @liranmauda To handle the atomicity of put and the fact that link will not replace the target file if exists - I think we can first create the link as a temp name in the internal dir, and only then move over to the target name, and since move will replace the target name, that will handle the replace-object case. WDYT?

Sounds good.

liranmauda:
> > @liranmauda To handle the atomicity of put and the fact that link will not replace the target file if exists - I think we can first create the link as a temp name in the internal dir, and only then move over to the target name, and since move will replace the target name, that will handle the replace-object case. WDYT?
> 
> Sounds good.

Now I see the issue, rename of a hard link will make a copy of it so we will leave a trail.
for example, do copy using a link on the same file 3 times will result in 5 links (1 original, 1 the desired name, 3 temp names)

liranmauda:
> I don't see the problem.
> try this:
> 
> ```shell
> # target already exists
> touch target_file
> touch source_file
> 
> # simulate put-copy with link()+rename()
> ln source_file temp_file
> 
> # now source_file and temp_file point to same inode with nlinks=2
> stat target_file source_file temp_file 
> 
> # move and replace target
> mv temp_file target_file
> 
> # now the previous target_file got deleted, and instead it now points to the same inode as source_file, with nlinks=2
> stat target_file source_file
> ```

This is when the target file is not a link of the source.
when the target is a link you can see:

```shell
# target already exists and it is a link
$ touch  source_file
$ ln source_file target_file

# simulate put-copy with link()+rename()
$ ln source_file temp_file

# now source_file, target_file  and temp_file point to same inode with nlinks=3
$ stat target_file source_file temp_file

# move and replace target and see it fails: 
$ mv temp_file target_file
mv: 'temp_file' and 'target_file' are the same file
```
running strace we can see :
```shell
renameat2(AT_FDCWD, "temp_file", AT_FDCWD, "target_file", RENAME_NOREPLACE) = -1 EEXIST (File exists)
```

src/sdk/namespace_fs.js
```diff
@@ -492,23 +494,90 @@ class NamespaceFS {
...
-        } catch (err) {
-            throw this._translate_object_error_codes(err);
+    // Checking that 2 files have the same inode. if so, we will return the etag of the file_path
+    async _is_same_inode(fs_account_config, source_file_path, file_path) {
+        try {
+            const file_path_stat = await nb_native().fs.stat(fs_account_config, file_path);
+            const file_path_inode = file_path_stat.ino.toString();
+            const source_file_stat = await nb_native().fs.stat(fs_account_config, source_file_path);
+            const source_file_inode = source_file_stat.ino.toString();
+            if (file_path_inode === source_file_inode) {
```
guymguym, line 541: I think you also need to compare fsid

liranmauda: as discussed in Slack:
> Comparing both device and inode number (st_dev and st_ino returned by stat) will tell you whether two different file names refer to the same thing.

src/sdk/namespace_fs.js
```diff
@@ -492,23 +494,94 @@ class NamespaceFS {
...
+                const data = buffer.slice(0, bytesRead);
+                await target_file.write(fs_account_config, data);
+            }
+            //Closing the source_file and target files
+            await source_file.close(fs_account_config);
+            source_file = null;
+            await target_file.close(fs_account_config);
+            target_file = null;
+        } catch (e) {
+            console.error('Failed to copy object', e);
```
jeniawhite, line 577: Why you don't throw the error here? I would expect to have a throw when we did not succeed to copy.

src/sdk/namespace_fs.js
```diff
@@ -492,23 +494,94 @@ class NamespaceFS {
...
+            const file_path_inode = file_path_stat.ino.toString();
+            const file_path_device = file_path_stat.dev.toString();
+            const source_file_stat = await nb_native().fs.stat(fs_account_config, source_file_path);
+            const source_file_inode = source_file_stat.ino.toString();
+            const source_file_device = source_file_stat.dev.toString();
+            if (file_path_inode === source_file_inode && file_path_device === source_file_device) {
+                return { etag: this._get_etag(file_path_stat) };
+            }
+        } catch (e) {
+            return false;
```
jeniawhite, line 549: Why do we ignore failures and just assume that if the stat failed the inode is not the same?

src/sdk/namespace_fs.js
```diff
@@ -492,23 +494,94 @@ class NamespaceFS {
...
+        // Uploading to a temp file then we will rename it. 
+        const upload_id = uuidv4();
+        const upload_path = path.join(this.bucket_path, config.NSFS_TEMP_DIR_NAME, 'uploads', upload_id);
+        // console.log('NamespaceFS.upload_object:', upload_path, '->', file_path);
+        await Promise.all([this._make_path_dirs(file_path, fs_account_config), this._make_path_dirs(upload_path, fs_account_config)]);
+        if (params.copy_source) {
+            const source_file_path = path.join(this.bucket_path, params.copy_source.key);
+            try {
+                const same_inode = await this._is_same_inode(fs_account_config, source_file_path, file_path);
+                if (same_inode) return same_inode;
```
jeniawhite, line 510: A bit confusing to understand.
I would expect `_is_same_inode` to return a Boolean.
Instead of either returns an Object or `false`.
A suggestion is to make the function return Object or nothing and rename it in order to be more understandable.

src/sdk/namespace_fs.js
```diff
@@ -492,23 +494,94 @@ class NamespaceFS {
...
+                // Doing a hard link.
+                await nb_native().fs.link(fs_account_config, source_file_path, upload_path);
+            } catch (e) {
+                try {
+                    console.warn('NamespaceFS: COPY using link failed with:', e);
+                    await this._copy_stream(source_file_path, upload_path, fs_account_config);
+                } catch (err) {
+                    throw this._translate_object_error_codes(err);
```
jeniawhite, line 518: This catch is the last catch for both IF cases, you can handle both IF blocks with this single catch instead of multiplying it and creating more nested try/catch blocks.

src/sdk/namespace_fs.js
liranmauda: This function is to decide if we should try to link or not.
In the event of an already exist copy, which is with the same name and the same inode, linking will work but `renameat2` will fail and we will leave with more than 2 links (we will have an additional one in the temp dir).
So we are checking if we would link or not, and if we fail for any reason, we will try to link.

src/sdk/namespace_fs.js
liranmauda: We will need this logic somewhere, as when we are deciding not to do anything the copy should return the etag. 
and since we have the etag as we did `stat` inside `_is_same_inode` we can return it. 
so, if `same_inode` returns false we will try and link, and if it returns the etag then we are set and just return it.

jeniawhite (approved):

PR #6466, opened 2021-04-22
NSFS - s3 create and delete bucket flows + support namespace resource as account default resource

romayalon:

### Explain the changes
1. bucketspace_nb:
1.1.  in create_bucket() if default resource is namespace resource -> will mark `should_create_uls = true; ` in namespace bucket config -> create underlying storage (currently implememnted in namespace_fs only).
1.2. in delete_bucket()  if `should_create_uls = true; ` in namespace bucket config -> delete underlying storage (currently implememnted in namespace_fs only) 

2. renamed default_pool to default_resource.
3. Added s3 create and delete bucket on nsfs as default resource unit tests. 
4. removed tiering from being required in bucket schema + api since when having namespace resource as default resource, tiering is redundant and can not be configured.
5. Added new_buckets_path as a posssible argument for  create_account and update_account.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/bucket_api.js
+++ b/src/api/bucket_api.js
@@ -830,7 +830,7 @@ module.exports = {
 
         claim_bucket: {
             method: 'PUT',
-            required: ['name', 'tiering', 'email', 'create_bucket', 'namespace'],
+            required: ['name', 'email', 'create_bucket', 'namespace'],
             params: {
                 type: 'object',
                 properties: {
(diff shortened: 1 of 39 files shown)


jeniawhite (approved):

src/api/bucket_api.js
```diff
@@ -1439,7 +1442,8 @@ module.exports = {
                 },
                 caching: {
                     $ref: 'common_api#/definitions/bucket_cache_config'
-                }
+                },
+                should_create_uls: { type: 'boolean' } // should create underlying storage
```
guymguym, line 1446: I dont think this abbrev makes sense because it is impossible to know what it means unless you are the one who invented it... I would keep the it long instead

src/server/system_services/schemas/account_schema.js
```diff
@@ -29,7 +29,7 @@ module.exports = {
         next_password_change: { date: true },
 
         // default policy for new buckets
-        default_pool: { objectid: true },
+        default_resource: { objectid: true },
```
guymguym, line 32: This requires conversion of existing accounts... but i didn't see it included in the pr..

romayalon: correct thanks, added upgrade for accounts and namespace buckets.

src/api/bucket_api.js
romayalon: sure

PR #6471, opened 2021-04-26
NSFS | Fix copy when not the same namespace

liranmauda:
- Renamed is_same_namespace function to is_server_side_copy
- Fix copy when not the same namespace


romayalon (approved):
LGTM

PR #6480, opened 2021-04-28
fix NSFS permissions issues

### LINKED ISSUE #6469: All buckets are listed to all users instead of different gid and uid
### Environment info
NooBaa Version:
[root@curlicue-inf ~]# noobaa version
INFO[0000] CLI version: 5.8.0
INFO[0000] noobaa-image: noobaa/noobaa-core:5.8.0-20210418
INFO[0000] operator-image: noobaa/noobaa-operator:5.8.0
[root@curlicue-inf ~]# oc version
Client Version: 4.6.18
Server Version: 4.6.18
Kubernetes Version: v1.19.0+f173eb4
[root@curlicue-inf ~]#

### Actual behavior
1. Any user can perform any action on buckets instead of different gid and uid

### Expected behavior
1. Update, Create, Delete and listing operation should not be performed by the different gid uid users

### Steps to reproduce
1. Create user1 with some gid and uid
2. Create user2 with different gid and uid
3. Set alias of user2 and do s3 ls
4. User 2 can list , update, delete buckets of user1

romayalon:

### Explain the changes
1. added _stat_with_permissions_check  function that checks for uid and gid and others permission after running stat.
2. changed mkdir to create directories with mode 0o770.
3. removed ACL wrong test from native_fs tests.
4. added sudo index for running tests that require sudo permissions.

1. Fixed #6469 

### Testing Instructions:
1.

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -227,6 +227,32 @@ struct Stat : public FSWorker
     }
 };
 
+
+/**
+ * EffectiveAccess is an fs op
+ */
+struct EffectiveAccess : public FSWorker
+{
+    std::string _path;
+    struct stat _stat_res;
+
+    EffectiveAccess(const Napi::CallbackInfo& info) : FSWorker(info)
+    {
+        _path = info[1].As<Napi::String>();
+        Begin(XSTR() << DVAL(_path));
+    }
+    virtual void Work()
+    {
+        int fd = open(_path.c_str(), O_RDONLY);
+        if (fd < 0) {
+            SetSyscallError();
+            return;
+        }
+        int r = close(fd);
+        if (r) SetSyscallError();
+    }
+};
+
 /**
  * Unlink is an fs op
  */
@@ -258,7 +284,9 @@ struct Mkdir : public FSWorker
         , _mode(0777)
     {
         _path = info[1].As<Napi::String>();
-        _mode = info[2].As<Napi::Number>();
+        if (info.Length() > 2 && !info[2].IsUndefined()) {
+            _mode = info[2].As<Napi::Number>().Uint32Value();
+        }
         Begin(XSTR() << DVAL(_path));
     }
     virtual void Work()
@@ -866,6 +894,7 @@ fs_napi(Napi::Env env, Napi::Object exports)
     auto exports_fs = Napi::Object::New(env);
     
     exports_fs["stat"] = Napi::Function::New(env, api<Stat>);
+    exports_fs["effectiveAccess"] = Napi::Function::New(env, api<EffectiveAccess>);
     exports_fs["unlink"] = Napi::Function::New(env, api<Unlink>);
     exports_fs["rename"] = Napi::Function::New(env, api<Rename>); 
     exports_fs["mkdir"] = Napi::Function::New(env, api<Mkdir>); 
--- a/src/sdk/bucketspace_nb.js
+++ b/src/sdk/bucketspace_nb.js
@@ -5,6 +5,7 @@ const _ = require('lodash');
 const P = require('../util/promise');
 const nb_native = require('../util/nb_native');
 const dbg = require('../util/debug_module')(__filename);
+const path = require('path');
 
 /**
  * @implements {nb.BucketSpace}
@@ -189,16 +190,19 @@ class BucketSpaceNB {
             _.isUndefined(account.nsfs_account_config.gid)) return false;
         try {
             dbg.log0('_has_access_to_nsfs_dir', namespace_bucket_config.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
+            // always allow root access
+            if (account.nsfs_account_config.uid === 0 && account.nsfs_account_config.gid === 0) return true;
 
-            await nb_native().fs.stat({
+            await nb_native().fs.effectiveAccess({
                     uid: account.nsfs_account_config.uid,
                     gid: account.nsfs_account_config.gid,
                     backend: namespace_bucket_config.write_resource.resource.fs_backend
-                }, namespace_bucket_config.write_resource.resource.fs_root_path);
+                }, path.join(namespace_bucket_config.write_resource.resource.fs_root_path, namespace_bucket_config.write_resource.path || ''));
 
             return true;
         } catch (err) {
-            if (err.code === 'EPERM' && err.message === 'Operation not permitted') return false;
+            dbg.log0('_has_access_to_nsfs_dir4', err);
+            if (err.code === 'EACCES' || (err.code === 'EPERM' && err.message === 'Operation not permitted')) return false;
             throw err;
         }
     }
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -48,6 +48,10 @@ function isDirectory(ent) {
     }
 }
 
+function get_umasked_mode(mode) {
+     // eslint-disable-next-line no-bitwise
+    return mode & ~config.NSFS_UMASK;
+}
 /**
  * @param {fs.Dirent} e
  * @returns {string}
@@ -65,6 +69,7 @@ function make_named_dirent(name) {
     entry.name = name;
     return entry;
 }
+
 
 /**
  * @typedef {{
@@ -401,7 +406,7 @@ class NamespaceFS {
         try {
             await this._load_bucket(params, fs_account_config);
             const file_path = this._get_file_path(params);
-            file = await nb_native().fs.open(fs_account_config, file_path);
+            file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(0o666));
 
             const start = Number(params.start) || 0;
             const end = isNaN(Number(params.end)) ? Infinity : Number(params.end);
@@ -517,7 +522,7 @@ class NamespaceFS {
         try {
             let q_buffers = [];
             let q_size = 0;
-            target_file = await nb_native().fs.open(fs_account_config, upload_path, 'w');
+            target_file = await nb_native().fs.open(fs_account_config, upload_path, 'w', get_umasked_mode(0o666));
             const flush = async () => {
                 if (q_buffers.length) {
                     await target_file.writev(fs_account_config, q_buffers);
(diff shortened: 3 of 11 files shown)


src/test/unit_tests/index.js
```diff
@@ -52,9 +52,6 @@ require('./test_s3_list_objects');
 require('./test_nb_native_b64');
 require('./test_bucket_chunks_builder');
 require('./test_mirror_writer');
-require('./test_namespace_fs');
```
liranmauda, line 55: Do we really need to move also the test_namespace_fs into the sudo index? 
my concern here is that locally we will still use `npm run tests` and not `npm run test:local` which means that we will skip this.

romayalon: I agree that now we have 2 options of running tests and it's confusing,
but since we call namespace_fs functions and these functions need to set_uid() and set_gid() correctly, I believe that it's needed for future tests (currently, there is no permissions check in test_namespace_fs so it still passes without root permissions).

liranmauda: I think that for now, as the suit does not need permissions, we can leave it in the main index. 
We can move it later if we will add permissions to that suit 
WDYT?

romayalon: I think the tests are not running properly without root permissions so running them incorrectly can cause future issues that we want to avoid, @jeniawhite WDYT?

src/sdk/namespace_fs.js
```diff
@@ -66,6 +66,25 @@ function make_named_dirent(name) {
...
+// this function will call native stat and check for matching uid, gid and `others` mode of a given account config
+// octal_others_mode < 4 is test is for read permission.
+// this number might be changed after supporting more backends types
+async function _stat_with_permissions_check(fs_account_config, stat_path) {
+    const res = await nb_native().fs.stat(fs_account_config, stat_path);
+    const octal_others_mode = parseInt(res.mode.toString(8), 10) % 10;
+    console.log('_stat_with_permissions_check', fs_account_config, res, octal_others_mode);
+    // handle root access
+    if (fs_account_config.uid === 0 && fs_account_config.gid === 0) return res;
+    if (res.uid !== fs_account_config.uid && res.gid !== fs_account_config.gid &&
```
guymguym, line 79: I don't follow - why are we checking for exact matching of uid and gid?
Why isn't the stat operation itself sufficient to fail with permission error if the file ACL is not allowing access?

src/test/unit_tests/test_bucketspace.js
```diff
@@ -42,8 +42,14 @@ mocha.describe('bucket operations - namespace_fs', function() {
...
-
-    mocha.before(async () => fs_utils.create_fresh_path(tmp_fs_root));
+    mocha.before(function() {
+        if (process.getgid() !== 0 || process.getuid() !== 0) {
+            coretest.log('No Root permissions found in env. Skipping test');
+            this.skip(); // eslint-disable-line no-invalid-this
+        }
+    });
+    mocha.before(async () => fs_utils.create_fresh_path(tmp_fs_root, 0o777));
+    mocha.before(async () => fs_utils.create_fresh_path(tmp_fs_root + bucket_path, 0o770));
```
guymguym, line 52: can you explain - why 770?

src/sdk/bucketspace_nb.js
```diff
@@ -189,15 +190,24 @@ class BucketSpaceNB {
...
                     uid: account.nsfs_account_config.uid,
                     gid: account.nsfs_account_config.gid,
                     backend: namespace_bucket_config.write_resource.resource.fs_backend
-                }, namespace_bucket_config.write_resource.resource.fs_root_path);
+                }, path.join(namespace_bucket_config.write_resource.resource.fs_root_path, namespace_bucket_config.write_resource.path || ''));
 
-            return true;
+            if (!stat) return false;
+
+            const octal_others_mode = parseInt(stat.mode.toString(8), 10) % 10;
```
guymguym, line 204: isn't it simpler to use modulo 8?
```suggestion
            const octal_others_mode = stat.mode % 8;
```

src/sdk/namespace_fs.js
romayalon: according to: https://man7.org/linux/man-pages/man2/lstat.2.html
`No permissions are required on the file itself` 
so for every uid gid it didn't fail, that's why I had to add these checks.

src/test/unit_tests/index.js
romayalon: fixed.

src/test/unit_tests/test_bucketspace.js
romayalon: I want that `other` users won't be able to access the directory.

src/sdk/bucketspace_nb.js
romayalon: yes, fixed it thanks.

src/sdk/namespace_fs.js
guymguym: ok, but checking for exact uid+gid match is not enough. We need to get the filesystem to check the permissions, because it might be using ACLs which we don't want to reimplement. 

Maybe we should https://man7.org/linux/man-pages/man2/access.2.html? 
Or perhaps instead of calling stat directly we should call open('r') and fstat() and then open for reading would fail?

WDYT @eshelmarc @motorman-ibm @nimrod-becker @dannyzaken

src/test/unit_tests/test_bucketspace.js
guymguym: ok got it

src/sdk/namespace_fs.js
```diff
@@ -864,24 +884,25 @@ class NamespaceFS {
...
     }
 
-    async _create_path(dir, fs_account_config) {
+    async _create_path(dir, fs_account_config, mode) {
         let dir_path = path.isAbsolute(dir) ? path.sep : '';
         for (const item of dir.split(path.sep)) {
             dir_path = path.join(dir_path, item);
             try {
-                await nb_native().fs.mkdir(fs_account_config, dir_path, 0o777);
+                await nb_native().fs.mkdir(fs_account_config, dir_path, mode || 0o777);
```
guymguym, line 905: Why are we setting the mode? Why not let the default apply? In linux every user has a umask that sets the default mode for new files/dirs, so I would say that we want to use that as much as possible. If we can't then I would at least set the default in config like this:
```javascript
config.NSFS_DEFAULT_MODE_DIR=0o755;
config.NSFS_DEFAULT_MODE_FILE=0o644;
```

guymguym: see https://geek-university.com/linux/set-the-default-permissions-for-newly-created-files/

guymguym: if we need more granular control on the umask for each account we can add it to our nsfs_account_config...

src/sdk/namespace_fs.js
```diff
@@ -501,10 +520,11 @@ class NamespaceFS {
             const upload_path = path.join(this.bucket_path, config.NSFS_TEMP_DIR_NAME, 'uploads', upload_id);
             // console.log('NamespaceFS.upload_object:', upload_path, '->', file_path);
 
-            await Promise.all([this._make_path_dirs(file_path, fs_account_config), this._make_path_dirs(upload_path, fs_account_config)]);
+            await Promise.all([this._make_path_dirs(file_path, fs_account_config, 0o770),
```
guymguym, line 523: Why this mode specifically? I think the default mode should apply..

src/sdk/namespace_fs.js
```diff
@@ -842,14 +862,14 @@ class NamespaceFS {
...
-            await nb_native().fs.stat(fs_account_config, this.bucket_path);
+            await _stat_with_permissions_check(fs_account_config, this.bucket_path);
         } catch (err) {
             if (err.code === 'ENOENT') {
                 // err.rpc_code = 'NO_SUCH_BUCKET';
                 // Instead of returning NO_SUCH_BUCKET we create the bucket dir.
                 // TODO should verify bucket exists in the system?
                 // or was it verified already for the request?
-                await this._create_path(this.bucket_path, fs_account_config);
+                await this._create_path(this.bucket_path, fs_account_config, 0o770);
```
guymguym, line 872: again, I think using default mode.

src/sdk/namespace_fs.js
eshelmarc: The file system will check permission on each operation so as long as you set uid/gid the fs will do the rest

romayalon: fixed.

src/sdk/namespace_fs.js
romayalon: according to our discussion, umask (007) now applied in a global context. the only issue here is that files get -rw-rw-- permissions, I can change it by setting the mode to 667 but I find it too hardcoded. 
WDYT? @guymguym 
do we have to preserve the execute permissions on files?

romayalon:
@guymguym 
access is not suitable for our needs - it uses the real uid and gid and not on the effective. that's why I'm still using open.
https://man7.org/linux/man-pages/man2/access.2.html

src/native/fs/fs_napi.cpp
```diff
@@ -866,6 +894,7 @@ fs_napi(Napi::Env env, Napi::Object exports)
     auto exports_fs = Napi::Object::New(env);
     
     exports_fs["stat"] = Napi::Function::New(env, api<Stat>);
+    exports_fs["effectiveAccess"] = Napi::Function::New(env, api<EffectiveAccess>);
```
guymguym, line 897: On second thought I would rename this action to `checkAccess`

src/sdk/namespace_fs.js
```diff
@@ -401,7 +406,7 @@ class NamespaceFS {
         try {
             await this._load_bucket(params, fs_account_config);
             const file_path = this._get_file_path(params);
-            file = await nb_native().fs.open(fs_account_config, file_path);
+            file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(0o666));
```
guymguym, line 409: I would still define these 0o666 and 0o777 as constants in the file scope:
```
const BASE_MODE_FILE = 0o666;
const BASE_MODE_DIR = 0o777;
```

src/sdk/bucketspace_nb.js
```diff
@@ -189,16 +190,19 @@ class BucketSpaceNB {
             _.isUndefined(account.nsfs_account_config.gid)) return false;
         try {
             dbg.log0('_has_access_to_nsfs_dir', namespace_bucket_config.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
+            // always allow root access
+            if (account.nsfs_account_config.uid === 0 && account.nsfs_account_config.gid === 0) return true;
```
guymguym, line 194: I am not sure about this - it does sound reasonable that root user should see all nsfs buckets, but on the other hand, perhaps there's a more complex ACL on the bucket, and root user is intentionally squashed and not authorized by ACL.
See https://www.ibm.com/docs/ja/spectrum-scale/5.1.0?topic=system-remote-user-access-gpfs-file.
Perhaps we can rethink this check and decide if we really need it, or should simply continue to check access even for root accounts.

PR #6481, opened 2021-05-02
NSFS cleanup empty dirs on objects delete

### LINKED ISSUE #6425: NSFS cleanup empty dirs on objects delete
### Environment info
- NooBaa Version: **5.9**
- Platform: **NA**

### Actual behavior
1. Creating object with key `a/b/c/name` will do the equivalent of mkdir -p `a/b/c/`
2. When the object is deleted the folders `a/` and `a/b/` and `a/b/c/` are left behind.

### Expected behavior
1. Object deletions should delete empty parent folders.

### Steps to reproduce
1. See above.

liranmauda:
### Explain the changes

- Cleanup empty dirs on objects delete
- Add delete path to test_namespace_fs.js

1. Fixes: #6425 

### Testing Instructions:
1. Create a file with a path (`/a/b/c/<filename>`)
2. Delete the file 
3. See that all the path was deleted.
4. Create 2 files: (`/a/b/c/<filename>` and `/a/<filename_2>`)
5. Delete the file `/a/b/c/<filename>`
6. See that only the empty path was deleted  and the `/a/<filename_2>` remains

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -685,6 +685,7 @@ class NamespaceFS {
             const file_path = this._get_file_path(params);
             console.log('NamespaceFS: delete_object', file_path);
             await nb_native().fs.unlink(fs_account_config, file_path);
+            await this._delete_object_empty_path(file_path, fs_account_config);
             return {};
         } catch (err) {
             throw this._translate_object_error_codes(err);
@@ -699,6 +700,7 @@ class NamespaceFS {
                 const file_path = this._get_file_path({ key });
                 console.log('NamespaceFS: delete_multiple_objects', file_path);
                 await nb_native().fs.unlink(fs_account_config, file_path);
+                await this._delete_object_empty_path(file_path, fs_account_config);
             }
             // TODO return deletion reponse per key
             return params.objects.map(() => ({}));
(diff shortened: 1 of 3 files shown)


src/sdk/namespace_fs.js
```diff
@@ -889,6 +891,19 @@ class NamespaceFS {
         }
     }
 
+    async _delete_object_empty_path(obj, fs_account_config) {
+        let dir = path.dirname(obj);
```
romayalon, line 895: seems to me that you need to add error handling here since readdir and rmdir can fail,
also need to think if this should fails the whole function or not.

liranmauda: Sure, I will wrap it with try and catch. 
I Will prompt a warning but will not fail the entire delete.
In this situation, we will level the path there

romayalon (approved):

src/sdk/namespace_fs.js
```diff
@@ -889,6 +891,23 @@ class NamespaceFS {
         }
     }
 
+    async _delete_object_empty_path(obj, fs_account_config) {
+        try {
+            let dir = path.dirname(obj);
+            while (dir !== this.bucket_path) {
+                const entries = await nb_native().fs.readdir(fs_account_config, dir);
```
guymguym, line 898: Why are you calling readdir?? This is very wasteful. Just try rmdir instead and catch errors - specifically ENOTEMPTY is expected, but also other errors can be ignored (after logging).

PR #6497, opened 2021-05-11
print fix

jackyalbo:

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -104,12 +104,12 @@ struct FSWorker : public Napi::AsyncWorker
         // pid_t tid = syscall(__NR_gettid);
         auto tid = std::this_thread::get_id();
 #endif
-        DBG1("FS::FSWorker::Start Execute: " << _desc << 
-            " orig_uid:" << orig_uid << 
-            " orig_gid:" << orig_gid << 
-            " req_uid:" << _req_uid << 
-            " req_gid:" << _req_gid << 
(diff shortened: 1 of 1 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -379,11 +379,10 @@ struct Writefile : public FSWorker
         auto buf = info[2].As<Napi::Buffer<uint8_t>>();
         _data = buf.Data();
         _len = buf.Length();
-        Begin(XSTR() << DVAL(_path) << DVAL(_len));
+        Begin(XSTR() << "FS::Writefile" << DVAL(_path) << DVAL(_len));
```
guymguym, line 382: missing space after the method names
```suggestion
        Begin(XSTR() << "FS::Writefile " << DVAL(_path) << DVAL(_len));
```

guymguym: Also there's no need for the `FS::` prefix

guymguym (changes requested):
See comments on adding space after method name in the description and removing the FS:: prefix from it.

I still see a other Begin() calls that do not pass the method name - can you add. For example Rename, Rmdir, etc.

guymguym (commented):
Also I see that in FSWorker::Execute the print is weird:
```
DBG1("FS::FSWorker::Start Execute: ...
```
Should be 
```
DBG1("FS::FSWorker::Execute: ...
```

guymguym (approved):

PR #6508, opened 2021-05-13
fixing build procedure for macos

jackyalbo:

### Explain the changes
1.  Added syscall implementation for macos. source: https://github.com/streamich/libsys/blob/master/syscall/darwin.c
2. SYS_setresuid and SYS_setresgid doesn't have a define in macos. used SYS_setreuid and SYS_setregid instead
3. Not a Full solution as under using warp in macos we still get panic on setgid

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -20,6 +20,29 @@
 namespace noobaa
 {
 
+#if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
+    #define CARRY_FLAG_BIT 1
+    #define RETURN_SYSCALL_RESULT(result, flags) return (flags & CARRY_FLAG_BIT) ? -result : result;
+
+    // syscall for macos from here: https://github.com/streamich/libsys/blob/master/syscall/darwin.c
+    static inline int64_t
+    syscall_2(int64_t num, int64_t arg1, int64_t arg2)
+    {
+        int64_t result;
+        int64_t flags;
+
+        __asm__ __volatile__(
+            "syscall;\n"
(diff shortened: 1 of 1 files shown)


jackyalbo:
We can use SYS_setreuid and SYS_setregid instead of SYS_setresuid and SYS_setresgid and then we won't need the special define - I'm not sure if we need to set the saved uid.

src/native/fs/fs_napi.cpp
```diff
@@ -115,29 +138,29 @@ struct FSWorker : public Napi::AsyncWorker
         bool change_uid = orig_uid != _req_uid;
         bool change_gid = orig_gid != _req_gid;
         if (change_gid) {
-            int r = syscall(SYS_setresgid, -1, _req_gid, -1);
+            int r = syscall_3(SYS_SETRESGID, -1, _req_gid, -1);
```
guymguym, line 141: isn't it just for Mac? what about linux?

guymguym:
@jackyalbo What did you say happens with the current fix? 
Does it still change the uid/gid for the entire process instead of per thread?

jackyalbo:
> @jackyalbo What did you say happens with the current fix?
> Does it still change the uid/gid for the entire process instead of per thread?

@guymguym  This is my assumption as with current code I get this: ```PANIC: FS::FSWorker::Execute setgid before work failed GID:0|0 UID:30|0```

jackyalbo:
@guymguym I added the thread mutexes - please take another look. Thanks.
The warp run has passed - so it probably was the threads issue in continue to your previous question

guymguym:
I see that there is another way - see `man pthread_setugid_np`.
I opened an alternative PR - #6543

PR #6518, opened 2021-05-18
Delete account by uid/gid

romayalon:

### Explain the changes
1. 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -658,6 +658,50 @@ function delete_account(req) {
         );
 }
 
+
+/**
+ *
+ * DELETE_ACCOUNT BY PROPERTY
+ *
+ */
+ function delete_account_by_property(req) {
+    let roles_to_delete = [];
+    let accounts_to_delete = system_store.get_accounts_by_nsfs_account_config(req.rpc_params.nsfs_account_config)
+        .map(account_to_delete => {
+            if (!account_to_delete) {
+                throw new RpcError('NO_SUCH_ACCOUNT', 'No such account email: ' + req.rpc_params.email);
+            }
+            if (account_to_delete.is_support) {
+                throw new RpcError('BAD_REQUEST', 'Cannot delete support account');
+            }
+            if (String(account_to_delete._id) === String(req.system.owner._id)) {
+                throw new RpcError('BAD_REQUEST', 'Cannot delete system owner account');
+            }
+            if (!is_support_or_admin_or_me(req.system, req.account, account_to_delete)) {
+                throw new RpcError('UNAUTHORIZED', 'Cannot delete account');
+            }
+            if (system_store.data.buckets.find(b => String(b.owner_account) === String(account_to_delete._id))) {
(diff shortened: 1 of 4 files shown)


src/test/unit_tests/test_bucketspace.js
```diff
@@ -167,6 +167,64 @@ mocha.describe('bucket operations - namespace_fs', function() {
...
+        } catch (err) {
+            assert.ok(err.rpc_code === 'NO_SUCH_ACCOUNT');
+        }
+
+        try {
+            let deleted_account_exist = await rpc_client.account.read_account({ email: 'account_wrong_uid1@noobaa.com' });
+            assert.fail(`found account: ${deleted_account_exist} - account should be deleted`);
+        } catch (err) {
+            assert.ok(err.rpc_code === 'NO_SUCH_ACCOUNT');
+        }
```
liranmauda, line 207: should it be in a for?

src/test/unit_tests/test_bucketspace.js
```diff
@@ -167,6 +167,64 @@ mocha.describe('bucket operations - namespace_fs', function() {
...
+        let list_account_resp2 = await rpc_client.account.list_accounts();
+        console.log(inspect(list_account_resp2));
```
liranmauda, line 209: Do we want to do something with this list? 
If not, then why are we listing?

nimrod-becker: I think the idea was to verify we don't break anything after deletion, but Romy can correct me if I'm wrong

liranmauda: In that case, We probably need an assert to verify it.

src/server/system_services/account_server.js
```diff
@@ -658,6 +658,50 @@ function delete_account(req) {
...
+    let roles_to_delete = [];
+    let accounts_to_delete = system_store.get_accounts_by_nsfs_account_config(req.rpc_params.nsfs_account_config)
+        .map(account_to_delete => {
+            if (!account_to_delete) {
+                throw new RpcError('NO_SUCH_ACCOUNT', 'No such account email: ' + req.rpc_params.email);
+            }
+            if (account_to_delete.is_support) {
+                throw new RpcError('BAD_REQUEST', 'Cannot delete support account');
+            }
+            if (String(account_to_delete._id) === String(req.system.owner._id)) {
```
liranmauda, line 677: Can the _id be from a different type? 
Why do we cast?

romayalon: the case here is taken from the original delete_account function, it's done because its type is objectId and comparing between 2 objectIDs should be done by using the mongo equal function / casting to strings.

src/test/unit_tests/test_bucketspace.js
romayalon: so it was for my verification since the first try catch assertion is probably enough for this test but I added another assert for not deleting all accounts for example.

src/test/unit_tests/test_bucketspace.js
romayalon: ok thanks

liranmauda (approved):

PR #6525, opened 2021-05-20
Fixing error codes for NSFS

jeniawhite:
### Explain the changes
1. Added handling or error code for the head of a directory

1. https://issues.redhat.com/browse/MCGI-26 

### Testing Instructions:
1.

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -378,11 +378,7 @@ class NamespaceFS {
             const file_path = this._get_file_path(params);
             const stat = await nb_native().fs.stat(fs_account_config, file_path);
             console.log(file_path, stat);
-            if (isDirectory(stat)) {
-                const err = new Error('NoSuchKey');
-                err.code = 'ENOENT';
-                throw this._translate_object_error_codes(err);
-            }
+            if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
             return this._get_object_info(params.bucket, params.key, stat);
         } catch (err) {
             throw this._translate_object_error_codes(err);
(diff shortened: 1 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -378,11 +378,7 @@ class NamespaceFS {
             const file_path = this._get_file_path(params);
             const stat = await nb_native().fs.stat(fs_account_config, file_path);
             console.log(file_path, stat);
-            if (isDirectory(stat)) {
-                const err = new Error('NoSuchKey');
-                err.code = 'ENOENT';
-                throw this._translate_object_error_codes(err);
-            }
+            if (isDirectory(stat)) throw _.defaults(new Error('NoSuchKey'), { code: 'ENOENT' });
```
guymguym, line 381: I'm not following this change - was the previous code not working?
Why are you not calling _translate_object_error_codes?
In any case I would advise against _.defaults in such a case where code is not even expected to be there.

jeniawhite: Is throwing `_translate_object_error_codes` and then catching that with another `_translate_object_error_codes` like we have prior to this commit, is better than handling it once?
There is no error and we create the error so it is either assigning or creating it with defaults, they basically do the same thing.
Do you want me to revert?

guymguym: ah thanks for explaining, I didn't notice this was inside a catch+translate.
So right, no need to translate twice. But what about avoiding _.defaults? How about using standard js function Object.assign (we don't ever expect that new Error will provide the error code?) -
```suggestion
            if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
```

PR #6526, opened 2021-05-20
S3 GET/HEAD Object - support partNumber query #6138 

### LINKED ISSUE #6138: S3 GET/HEAD Object - support partNumber query
### Environment info
- NooBaa Version: **5.6**
- Platform: **NA**

### Actual behavior
1. No support for partNumber (multipart number) as explained in S3 docs:

> Part number of the object being read. This is a positive integer between 1 and 10,000. Effectively performs a 'ranged' GET request for the part specified. Useful for downloading just a part of an object.

https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_RequestSyntax

2. PR #6129 added support to namespace_cache/s3 but for namespace_nb we need to support it in object_api/server.

### Expected behavior
1. Support get/head of specific multipart number.

### Steps to reproduce
1. `s3.getObject({ Bucket: 'bucket', Key: 'key', PartNumber: 1 })`

baum:
Effectively performs a 'ranged' GET request for the part specified.

    1. Include the requested multipart range in metadata as optional fields in ObjectMD/ObjectInfo.
    2. In Object service layer, while finding the object's metadata, calculate the multipart range
       using MDStore if a partNumber was requested in the S3 getObject query.
    3. In S3 Op layer, initiate a 'ranged' GET either if  partNumber
       was specified in the S3 ObjectGET query or alternatively, if ranges
       were specified in HTTP header.
    4. Remove namespace_nb part_number logic.
    5. Add system test for head/get object part functionality.
    6. Add CI unit test for head/get object part functionality which verifies NooBaa,
       Namespace & Namespace-Cache buckets.

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -23,6 +23,14 @@ module.exports = {
             maximum: 10000
         },
 
+        object_range: {
+            type: 'object',
+            properties: {
+                start: { type: 'integer' },
+                end: { type: 'integer' },
+            }
+        },
+
         bucket_trigger_event: { // Based on AWS: https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html#supported-notification-event-types
             enum: ['ObjectCreated', 'ObjectRemoved', 'ObjectRead', /* 'ObjectCreated:Put', 'ObjectCreated:CompleteMultipartUpload', ... */ ],
             type: 'string',
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -16,6 +16,13 @@ module.exports = {
 
     definitions: {
 
+
+        multipart: {
+            type: 'integer',
+            minimum: 1,
+            maximum: 10000
+        },
+
         bucket_trigger_event: { // Based on AWS: https://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html#supported-notification-event-types
             enum: ['ObjectCreated', 'ObjectRemoved', 'ObjectRead', /* 'ObjectCreated:Put', 'ObjectCreated:CompleteMultipartUpload', ... */ ],
             type: 'string',
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -17,7 +17,7 @@ module.exports = {
     definitions: {
 
 
-        multipart: {
+        multipart_num: {
             type: 'integer',
             minimum: 1,
             maximum: 10000
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -1384,8 +1384,13 @@ module.exports = {
                 num_multiparts: {
                     $ref: 'common_api#/definitions/multipart'
                 },
-                multipart_start: { type: 'integer' },
-                multipart_end: { type: 'integer' },
+                multipart_range: {
+                    type: 'object',
+                    properties: {
+                        start: { type: 'integer' },
+                        end: { type: 'integer' },
+                    }
+                },
                 content_type: { type: 'string' },
                 content_encoding: { type: 'string' },
                 create_time: { idate: true },
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -152,7 +152,7 @@ module.exports = {
                             required: ['num', 'etag'],
                             properties: {
                                 num: {
-                                    $ref: 'common_api#/definitions/multipart'
+                                    $ref: 'common_api#/definitions/multipart_num'
                                 },
                                 etag: {
                                     type: 'string'
@@ -221,7 +221,7 @@ module.exports = {
                         type: 'string',
                     },
                     num: {
-                        $ref: 'common_api#/definitions/multipart'
+                        $ref: 'common_api#/definitions/multipart_num'
                     },
                     size: {
                         type: 'integer',
@@ -274,7 +274,7 @@ module.exports = {
                         type: 'string',
                     },
                     num: {
-                        $ref: 'common_api#/definitions/multipart'
+                        $ref: 'common_api#/definitions/multipart_num'
                     },
                     multipart_id: {
                         objectid: true
@@ -362,7 +362,7 @@ module.exports = {
                             ],
                             properties: {
                                 num: {
-                                    $ref: 'common_api#/definitions/multipart'
+                                    $ref: 'common_api#/definitions/multipart_num'
                                 },
                                 size: {
                                     type: 'integer'
@@ -570,7 +570,7 @@ module.exports = {
                     key: { type: 'string' },
                     md_conditions: { $ref: '#/definitions/md_conditions' },
                     multipart_number: {
-                        $ref: 'common_api#/definitions/multipart'
+                        $ref: 'common_api#/definitions/multipart_num'
                     },
                     encryption: { $ref: 'common_api#/definitions/object_encryption' },
                     adminfo: {
@@ -1382,7 +1382,7 @@ module.exports = {
                 delete_marker: { type: 'boolean' },
                 num_parts: { type: 'integer' },
                 num_multiparts: {
-                    $ref: 'common_api#/definitions/multipart'
+                    $ref: 'common_api#/definitions/multipart_num'
                 },
                 multipart_range: {
                     type: 'object',
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -152,9 +152,7 @@ module.exports = {
                             required: ['num', 'etag'],
                             properties: {
                                 num: {
-                                    type: 'integer',
-                                    minimum: 1,
-                                    maximum: 10000
+                                    $ref: 'common_api#/definitions/multipart'
                                 },
                                 etag: {
                                     type: 'string'
@@ -223,9 +221,7 @@ module.exports = {
                         type: 'string',
                     },
                     num: {
-                        type: 'integer',
-                        minimum: 1,
-                        maximum: 10000
+                        $ref: 'common_api#/definitions/multipart'
                     },
                     size: {
                         type: 'integer',
@@ -278,9 +274,7 @@ module.exports = {
                         type: 'string',
                     },
                     num: {
-                        type: 'integer',
-                        minimum: 1,
-                        maximum: 10000
+                        $ref: 'common_api#/definitions/multipart'
                     },
                     multipart_id: {
                         objectid: true
@@ -368,9 +362,7 @@ module.exports = {
                             ],
                             properties: {
                                 num: {
-                                    type: 'integer',
-                                    minimum: 1,
-                                    maximum: 10000
+                                    $ref: 'common_api#/definitions/multipart'
                                 },
                                 size: {
                                     type: 'integer'
@@ -578,9 +570,7 @@ module.exports = {
                     key: { type: 'string' },
                     md_conditions: { $ref: '#/definitions/md_conditions' },
                     multipart_number: {
-                        type: 'integer',
-                        minimum: 1,
-                        maximum: 10000
+                        $ref: 'common_api#/definitions/multipart'
                     },
                     encryption: { $ref: 'common_api#/definitions/object_encryption' },
                     adminfo: {
@@ -1392,9 +1382,7 @@ module.exports = {
                 delete_marker: { type: 'boolean' },
                 num_parts: { type: 'integer' },
                 num_multiparts: {
-                    type: 'integer',
-                    minimum: 1,
-                    maximum: 10000
+                    $ref: 'common_api#/definitions/multipart'
                 },
                 multipart_start: { type: 'integer' },
                 multipart_end: { type: 'integer' },
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -152,7 +152,9 @@ module.exports = {
                             required: ['num', 'etag'],
                             properties: {
                                 num: {
-                                    type: 'integer'
+                                    type: 'integer',
+                                    minimum: 1,
+                                    maximum: 10000
                                 },
                                 etag: {
                                     type: 'string'
@@ -222,6 +224,8 @@ module.exports = {
                     },
                     num: {
                         type: 'integer',
+                        minimum: 1,
+                        maximum: 10000
                     },
                     size: {
                         type: 'integer',
@@ -275,6 +279,8 @@ module.exports = {
                     },
                     num: {
                         type: 'integer',
+                        minimum: 1,
+                        maximum: 10000
                     },
                     multipart_id: {
                         objectid: true
@@ -362,7 +368,9 @@ module.exports = {
                             ],
                             properties: {
                                 num: {
-                                    type: 'integer'
+                                    type: 'integer',
+                                    minimum: 1,
+                                    maximum: 10000
                                 },
                                 size: {
                                     type: 'integer'
@@ -569,6 +577,11 @@ module.exports = {
                     bucket: { $ref: 'common_api#/definitions/bucket_name' },
                     key: { type: 'string' },
                     md_conditions: { $ref: '#/definitions/md_conditions' },
+                    multipart_number: {
+                        type: 'integer',
+                        minimum: 1,
+                        maximum: 10000
+                    },
                     encryption: { $ref: 'common_api#/definitions/object_encryption' },
                     adminfo: {
                         type: 'object',
@@ -1378,6 +1391,13 @@ module.exports = {
                 is_latest: { type: 'boolean' },
                 delete_marker: { type: 'boolean' },
                 num_parts: { type: 'integer' },
+                num_multiparts: {
+                    type: 'integer',
+                    minimum: 1,
+                    maximum: 10000
+                },
+                multipart_start: { type: 'integer' },
+                multipart_end: { type: 'integer' },
                 content_type: { type: 'string' },
                 content_encoding: { type: 'string' },
                 create_time: { idate: true },
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -1385,11 +1385,7 @@ module.exports = {
                     $ref: 'common_api#/definitions/multipart_num'
                 },
                 multipart_range: {
-                    type: 'object',
-                    properties: {
-                        start: { type: 'integer' },
-                        end: { type: 'integer' },
-                    }
+                    $ref: 'common_api#/definitions/object_range'
                 },
                 content_type: { type: 'string' },
                 content_encoding: { type: 'string' },
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -14,10 +14,10 @@ async function get_object(req, res) {
     const agent_header = req.headers['user-agent'];
     const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
     const encryption = s3_utils.parse_encryption(req);
-    let part_number;
-    // If set, part_number should be positive integer from 1 to 10000
+    let multipart_number;
+    // If set, S3 part number should be positive integer from 1 to 10000
     if (req.query.partNumber) {
-        part_number = s3_utils.parse_part_number(req.query.partNumber, S3Error.InvalidArgument);
+        multipart_number = s3_utils.parse_part_number(req.query.partNumber, S3Error.InvalidArgument);
     }
     const md_conditions = http_utils.get_md_conditions(req);
 
@@ -31,8 +31,8 @@ async function get_object(req, res) {
     if (req.query.get_from_cache !== undefined) {
         md_params.get_from_cache = true;
     }
-    if (part_number) {
-        md_params.part_number = part_number;
+    if (multipart_number) {
+        md_params.multipart_number = multipart_number;
     }
     const object_md = await req.object_sdk.read_object_md(md_params);
 
@@ -52,24 +52,20 @@ async function get_object(req, res) {
     if (md_params.get_from_cache) {
         params.get_from_cache = true;
     }
-    if (part_number) {
-        params.part_number = part_number;
+    if (multipart_number) {
+        params.multipart_number = multipart_number;
     }
     try {
-        const ranges = http_utils.normalize_http_ranges(
-            http_utils.parse_http_ranges(req.headers.range),
-            obj_size
-        );
-        if (ranges) {
-            // reply with HTTP 206 Partial Content
-            res.statusCode = 206;
-            params.start = ranges[0].start;
-            params.end = ranges[0].end;
-            const content_range = `bytes ${params.start}-${params.end - 1}/${obj_size}`;
-            dbg.log1('reading object range', req.path, content_range, ranges);
-            res.setHeader('Content-Range', content_range);
-            res.setHeader('Content-Length', params.end - params.start);
-            // res.header('Cache-Control', 'max-age=0' || 'no-cache');
+        // A 'ranged' GET is performed if a  "part number"
+        // is specified in the S3 ObjectGET query or alternatively,
+        // ranges could be specified in HTTP header
+        const range = s3_utils.get_object_range(req, object_md);
+
+        // If range is specified, set it in HTTP response headers
+        if (range) {
+            params.start = range.start;
+            params.end = range.end;
+            s3_utils.set_range_response_headers(req, res, range, obj_size);
         } else {
             dbg.log1('reading object', req.path, obj_size);
         }
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -55,35 +55,8 @@ async function get_object(req, res) {
     if (multipart_number) {
         params.multipart_number = multipart_number;
     }
-    try {
-        // A 'ranged' GET is performed if a  "part number"
-        // is specified in the S3 ObjectGET query or alternatively,
-        // ranges could be specified in HTTP header
-        const range = s3_utils.get_object_range(req, object_md);
+    s3_utils.s3_range(req, res, object_md, obj_size, params);
 
-        // If range is specified, set it in HTTP response headers
-        if (range) {
-            params.start = range.start;
-            params.end = range.end;
-            s3_utils.set_range_response_headers(req, res, range, obj_size);
-        } else {
-            dbg.log1('reading object', req.path, obj_size);
-        }
-    } catch (err) {
-        if (err.ranges_code === 400) {
-            // return http 400 Bad Request
-            dbg.log1('bad range request', req.headers.range, req.path, obj_size);
-            throw new S3Error(S3Error.InvalidArgument);
-        }
-        if (err.ranges_code === 416) {
-            // return http 416 Requested Range Not Satisfiable
-            dbg.warn('invalid range', req.headers.range, req.path, obj_size);
-            // let the client know of the relevant range
-            res.setHeader('Content-Range', 'bytes */' + obj_size);
-            throw new S3Error(S3Error.InvalidRange);
-        }
-        throw err;
-    }
     // first_range_data are the first 4K data of the object
     // if the object's size or the end of range is smaller than 4K return it, else get the whole object
     if (params.object_md.first_range_data) {
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -81,6 +81,7 @@ async function get_object(req, res) {
         if (read_stream && read_stream.close) read_stream.close();
     });
 
+    dbg.log0('BAUM get_object response:', res);
     read_stream = await req.object_sdk.read_object_stream(params, res);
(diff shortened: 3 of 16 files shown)


src/api/object_api.js
```diff
@@ -565,6 +565,7 @@ module.exports = {
                     bucket: { $ref: 'common_api#/definitions/bucket_name' },
                     key: { type: 'string' },
                     md_conditions: { $ref: '#/definitions/md_conditions' },
+                    part_number: { type: 'integer' },
```
rimar-blink, line 568: hi

src/server/object_services/object_server.js
```diff
@@ -1415,6 +1416,14 @@ async function find_object_md(req) {
         obj = await MDStore.instance().find_object_latest(req.bucket._id, req.rpc_params.key);
     }
     check_object_mode(req, obj, 'NO_SUCH_OBJECT');
+
+    // fetch the objects parts ranges if a partNumber was requested
+    // and the object was uploaded using multipart upload
+    if (req.rpc_params.part_number && obj.num_parts) {
+        const parts = await MDStore.instance().find_all_parts_of_object(obj);
```
guymguym, line 1423: Why loading all parts? Why not just the requested one?

baum: Hey Guy. Good point. Refactored the change a bit. Added new MDStore interface to fetch a single part range for a specific part number, filtering is performed by the DB now. On metadata level, over RPC boundary, pass single requested range star/end instead of the whole object parts array.

guymguym (changes requested):
@baum See comments above - in our case there are several parts per multipart - sounds weird, but the meaning is that we maintain both the multipart concept from S3 (ObjectMultipart), and we chunk these to sub-parts (ObjectPart) based on dedup chunking (parts contain a reference to their multipart id if were uploaded in multipart-upload).

src/server/object_services/object_server.js
```diff
@@ -1415,6 +1417,19 @@ async function find_object_md(req) {
         obj = await MDStore.instance().find_object_latest(req.bucket._id, req.rpc_params.key);
     }
     check_object_mode(req, obj, 'NO_SUCH_OBJECT');
+
+    // fetch the object part range if a partNumber was requested
+    // and the object was uploaded using multipart upload
+    if (req.rpc_params.part_number && obj.num_parts) {
+        const parts = await MDStore.instance().find_part_of_object_by_number(obj, req.rpc_params.part_number);
```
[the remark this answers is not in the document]
baum: Hello Guy.

Thank you for your explanation, So S3 parts map into Noobaa multiparts and Noobaa multiparts could consist of one or more Noobaa de-dup parts. 

So before going into the woods of DB schema changes, fixed the calculation of the S3 part range for a specific part number using the existing MD data:

https://github.com/baum/noobaa-core/blob/75f549bd077eb1b62c9b2f32883a098aa2e48125/src/server/object_services/object_server.js#L1425-L1447

As for MDStore, added the following method, for multiparts less than equal given part number:

https://github.com/baum/noobaa-core/blob/75f549bd077eb1b62c9b2f32883a098aa2e48125/src/server/object_services/md_store.js#L917-L924

As for adding a start field for the multipart collection ( size if already there),  this might be not the most natural solution though. The offset within the object of the multipart is not known during the part uploaded, i.e. when a multipart is inserted into MDStore, however each multipart start could be calculated during multipart upload completion request. Then it is possible to go over all multiparts of the object and update each start field, does not sound cheap. So there is a trade off here.

What do you think?

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -52,14 +52,19 @@ async function get_object(req, res) {
     if (md_params.get_from_cache) {
         params.get_from_cache = true;
     }
-    if (part_number) {
-        params.part_number = part_number;
-    }
```
guymguym, line 57: Hmm this means that we now also affect how all namespaces also process reads with part number.
I'm thinking on NS-S3 for example - so now instead of passing part number to the GET request, we only send it on HEAD and then send a GET with a range. Not sure what to think of it yet, but I wonder if this is what we desire. We could limit this only to NS-NB if we prefer. WDYT? 
FYI @nimrod-becker @jeniawhite

src/api/object_api.js
```diff
@@ -1373,6 +1374,8 @@ module.exports = {
                 is_latest: { type: 'boolean' },
                 delete_marker: { type: 'boolean' },
                 num_parts: { type: 'integer' },
+                start: { type: 'integer' },
+                end: { type: 'integer' },
```
guymguym, line 1378: Should we name these more explicitly as `multipart_start`, `multipart_end`?
This might still be confusing a bit, unless we also rename `part_number` to `multipart_number`?

src/server/object_services/object_server.js
```diff
@@ -1415,9 +1417,35 @@ async function find_object_md(req) {
...
+ * @param part_number integer in 1..10000 range
+ * @param {nb.ObjectMD} obj
+ */
+async function calculate_multipart_range(part_number, obj) {
+    if (!part_number) {
+        return; // part number was not specified in the S3 Query
+    }
+
+    // fetch the multiparts of this obj which are lte(<=) the requested part number
+    const multiparts = await MDStore.instance().find_lte_multiparts_of_object_by_number(obj._id, part_number);
```
guymguym, line 1437: Can you at least add a comment with todo:
```javascript
// TODO store the range inside each multipart in _complete_object_multiparts() so we can read just the requested multipart here
```

guymguym (changes requested):

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -52,14 +52,19 @@ async function get_object(req, res) {
     if (md_params.get_from_cache) {
         params.get_from_cache = true;
     }
-    if (part_number) {
```
dannyzaken, line 55: I don't think this code should be removed for all cases. 
notice that `part_number` is checked in `namespace_cache.js` both for `read_object_md` and `read_object_stream`. 
since `part_number` was sent in `md_params` above, you will not get the start and end in object_md, but you also don't send the `part_number` now so you will get the wrong result

dannyzaken, line 55: I don't think this code should be removed for all cases.   
notice that `part_number` is checked in `namespace_cache.js` both for `read_object_md` and `read_object_stream`. 
since `part_number` was sent in `md_params` above, you will not get the start and end in object_md, but you also don't send the `part_number` now so you will get the wrong result

dannyzaken, line 55: I don't think this code should be removed for all cases.   
notice that `part_number` is checked in `namespace_cache.js` both for `read_object_md` and `read_object_stream`. 
since `part_number` was sent in `md_params` above, you will not get the start and end in object_md, but you also don't send the `part_number` now so you will get the wrong result

src/endpoint/s3/ops/s3_get_object.js
dannyzaken: I think that since start\end are only returned by namespace_nb we can set `params.part_number` if `req.query.partNumber` exists and there is no start\end on the `object_md`

src/server/object_services/object_server.js
```diff
@@ -1415,9 +1417,35 @@ async function find_object_md(req) {
...
 }
 
+/**
+ * calculate (start, end) range of the specified part_number
+ *
+ * @param part_number integer in 1..10000 range
+ * @param {nb.ObjectMD} obj
+ */
+async function calculate_multipart_range(part_number, obj) {
+    if (!part_number) {
```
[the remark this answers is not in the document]
baum: Sounds good

src/endpoint/s3/ops/s3_get_object.js
baum: Restoring

src/server/object_services/object_server.js
baum: Sure

src/server/object_services/md_store.js
```diff
@@ -914,6 +914,15 @@ class MDStore {
         return this._multiparts.find({ obj: obj_id, deleted: null });
     }
 
+    /**
+     * @param {nb.ID} obj_id
+     * @param part_number integer in 1..10000 range (S3 part number)
+     * @returns {Promise<nb.ObjectMultipart[]>}
+     */
+    async find_lte_multiparts_of_object_by_number(obj_id, part_number) {
```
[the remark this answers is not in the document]
baum: Sounds good

src/sdk/namespace_nb.js
```diff
@@ -86,9 +83,6 @@ class NamespaceNB {
             client: object_sdk.rpc_client,
             bucket: this.target_bucket,
         }, params);
-        // Noobaa bucket does not currrently support partNumber query parameter. Ignore it for now.
-        // If set, part_number is positive integer from 1 to 10000
-        if (params.part_number) _.unset(params, 'part_number');
```
dannyzaken, line 91: as a follow-up to the comment above in `s3_get_object`, maybe you should always pass `params.part_number` when it's set on the S3 request, and just leave this `_.unset` here

dannyzaken:
@baum you need to sign your commits (see the DCO failure). you can fix it now by rebasing with signoff 
```
git rebase master --signoff
```

dannyzaken:
@baum also check what is needed to support `partNumber` for S3 headObject

src/api/object_api.js
baum: Ok

src/endpoint/s3/ops/s3_get_object.js
baum: OK, restoring

baum:
@dannyzaken added partNumber for S3 headObject logic in the latest push [c25c900](https://github.com/noobaa/noobaa-core/commit/c25c9000a175fdb3d19da70cf7da3d44a6b3613e)

extracted common get/head logic (set_range_response_headers(),  get_object_range() to [s3_utils](https://github.com/noobaa/noobaa-core/blob/c25c9000a175fdb3d19da70cf7da3d44a6b3613e/src/endpoint/s3/s3_utils.js).

https://github.com/noobaa/noobaa-core/blob/c25c9000a175fdb3d19da70cf7da3d44a6b3613e/src/endpoint/s3/s3_utils.js#L259-L291

src/sdk/namespace_nb.js
baum: So guess we agreed this looks good.

baum:
According to S3 spec, multipart get/head object should return should return PartsCount in  x-amz-mp-parts-count HTTP header.

Partial support existed in namespace_s3 get_object_info(), see https://github.com/noobaa/noobaa-core/blob/7a62863e7c3dfeee2c99fe4c057dd12b72af2efa/src/sdk/namespace_s3.js#L690 
and in endpoint/s3_utils set response header, see https://github.com/noobaa/noobaa-core/blob/7a62863e7c3dfeee2c99fe4c057dd12b72af2efa/src/endpoint/s3/s3_utils.js#L306

@guymguym  went into DB Schema change woods 😃
Added num_multiparts into MDStore object md schema:: https://github.com/baum/noobaa-core/blob/0a162c1cfb895b5618a639bc5dd905b72d8d0a8b/src/server/object_services/schemas/object_md_schema.js#L71-L73

The new object md field is updated  upon multipart upload completion: https://github.com/baum/noobaa-core/blob/0a162c1cfb895b5618a639bc5dd905b72d8d0a8b/src/server/object_services/object_server.js#L1986

And allowed to pass num_multipart md field through RPC boundary.

Compared system test output with native  Amazon S3, looks OK to me .

nimrod-becker:
@baum do we need to worry about upgrade? Can we even fix it during upgrade/lazy? pretty sure we can't...
What happens when issuing a partNumber query for an object that was written before the schema change?

baum:
@nimrod-becker 

Q. do we need to worry about upgrade? 
A. DB schema change usually requires some upgrade logic, not sure of the current case implications. 

Q.  Can we even fix it during upgrade/lazy? pretty sure we can't...
A.  Since the num_multiparts object_md field could be calculated using info present in multiparts DB collection, I am optimistic about the lazy update/upgrade.

Q. What happens when issuing a partNumber query for an object that was written before the schema change? 
A. The system would not return PartsCount in x-amz-mp-parts-count HTTP header. Basically, the current behavior is expected.

The bottom line is as mentioned above, the number of multiparts of an object could be calculated by counting documents in multiparts DB collection, so alternative implementation, which does not involve DB schema change is possible.

nimrod-becker:
@baum so we might want to lazily calculate from the parts in the DB and populate the new field if partNumber is requested for an object without this field in its MD, WDYT?

baum:
@nimrod-becker   lazily calculate sounds elegant, considering upgrade.

Added lazy calculation/population

https://github.com/baum/noobaa-core/blob/304b35793ece5df341614b4b7679a11145292c65/src/server/object_services/object_server.js#L1449-L1462

Invoked from here

https://github.com/baum/noobaa-core/blob/304b35793ece5df341614b4b7679a11145292c65/src/server/object_services/object_server.js#L1425

baum:
probably could update num_multiparts even lazier, do not await update:

https://github.com/baum/noobaa-core/blob/87aa9c439642bd62afab54ae540db323ad9c1fc9/src/server/object_services/object_server.js#L1461-L1463

dannyzaken (approved):
LGTM.

@guymguym have time to take a look again?

src/api/object_api.js
```diff
@@ -1373,6 +1374,9 @@ module.exports = {
                 is_latest: { type: 'boolean' },
                 delete_marker: { type: 'boolean' },
                 num_parts: { type: 'integer' },
+                num_multiparts: { type: 'integer' },
+                multipart_start: { type: 'integer' },
+                multipart_end: { type: 'integer' },
```
igorpick, line 1379: You can add minimum and maximum validation:
```
"minimum": 1,
"maximum": 1000
```
https://json-schema.org/understanding-json-schema/reference/numeric.html

baum:
@guymguym any chance to move forward with this PR?

src/server/object_services/object_server.js
```diff
@@ -1416,9 +1419,52 @@ async function find_object_md(req) {
...
+ */
+ async function calculate_num_multiparts(obj) {
+    // fetch the commited multiparts of this obj
+    const multiparts = await MDStore.instance().find_commited_multiparts_of_object(obj._id);
+    if (multiparts.length) {
+        obj.num_multiparts = multiparts.length;
+
+        // schedule it - forget it
+        // don't await the update, perform in "background"
+        MDStore.instance().update_object_by_id(obj._id, { num_multiparts: obj.num_multiparts });
```
jeniawhite, line 1464: This looks like a point that is prone to race conditions or failures
I would have been worried about triggering in the background, is there a reason besides not slowing down the GET? 
If I'm not mistaken since this works on a particular object_id of finished uploads then there isn't risk of two parallel overrides setting different values and one of them winning the race condition and updating the invalid value of parts
I assume the above-mentioned case is not something that would happen since two overrides would have different object_ids, but I would verify if I'm missing something

src/server/object_services/schemas/object_md_schema.js
```diff
@@ -68,6 +68,10 @@ module.exports = {
         // NOTE: only updated once upload ends
         num_parts: { type: 'integer' },
 
+        // number of objects S3 multiparts for the object created by multipart upload
+        // NOTE: only updated once the upload of all parts is completed
+        num_multiparts: { type: 'integer' },
```
jeniawhite, line 73: What about upgrading for pre-existing objects?
I understand that you have a lazy update in ```calculate_num_multiparts``` upon reads, but this is a weird state in the database (hence your NOTE comment is invalid for pre-existing objects until they would have been read once after upgrade)

src/server/object_services/md_store.js
```diff
@@ -906,6 +906,23 @@ class MDStore {
         });
     }
 
+    /**
+     * @param {nb.ID} obj_id
+     * @returns {Promise<nb.ObjectMultipart[]>}
+     */
+    async find_commited_multiparts_of_object(obj_id) {
+        return this._multiparts.find({ obj: obj_id, deleted: null, uncommitted: null });
```
jeniawhite, line 914: We should not forget that ```uncommitted``` is not indexed
Just commenting out since I can see that this query is added on this property in addition to pre-existing queries: ```find_parts_by_start_range``` and ```find_parts_sorted_by_start```

src/server/object_services/object_server.js
dannyzaken: I'm not too worried about race conditions, but maybe to be on the safe side we can add the `await` here and not do it in the background. 
since this is only called when the part number is requested, it shouldn't affect the more common operations.

src/server/object_services/schemas/object_md_schema.js
dannyzaken: in general, we don't want to upgrade object-mds, mainly for scale reasons. I think that the lazy update is reasonable

src/server/object_services/md_store.js
dannyzaken: I think we should think about adjusting the indexes, but not as part of this PR

src/server/object_services/object_server.js
baum: @jeniawhite @dannyzaken  since 
- it is allowed for `num_multiparts` to be not initialized
- there is a single value  `num_multiparts` could take once initialized

I am not sure there is a race here. WDYT?

dannyzaken: I agree that there shouldn't be a race here, but I also understand Jenia's point - doing things in the background (especially updating DB) is bug-prone, and I think that in this case, it does not have a genuine performance improvement

baum: 👍, going back before [that comment](https://github.com/noobaa/noobaa-core/pull/6526#issuecomment-852747220) then.

src/api/object_api.js
```diff
@@ -1373,6 +1386,13 @@ module.exports = {
                 is_latest: { type: 'boolean' },
                 delete_marker: { type: 'boolean' },
                 num_parts: { type: 'integer' },
+                num_multiparts: {
+                    type: 'integer',
+                    minimum: 1,
+                    maximum: 10000
+                },
+                multipart_start: { type: 'integer' },
+                multipart_end: { type: 'integer' },
```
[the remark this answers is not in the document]
guymguym: Actually, maybe best to return it inside a single object like `multipart_range: {start,end}`

src/endpoint/s3/s3_utils.js
```diff
@@ -274,6 +274,40 @@ function set_encryption_response_headers(req, res, encryption = {}) {
...
+ * A 'ranged' GET/HEAD is performed if a  "part number" is specified in the S3
+ * query or alternatively, ranges could be specified in HTTP header
+ *
+ * @param req HTTP request
+ * @param {nb.ObjectInfo} object_md
+ * @returns an range object { start: XXX, end: YYY }
+ */
+function get_object_range(req, object_md) {
+    // S3 part number
+    if (req.query.partNumber && object_md.multipart_start !== undefined && object_md.multipart_end !== undefined) {
```
guymguym, line 296: hmm, this condition is a bit too soft - if `req.query.partNumber` is provided but `object_md.multipart_start/end` is undefined, shouldn't we error??

src/server/object_services/md_store.js
guymguym: Why are we calling find and not count? Is there any need to read those multiparts to memory?

guymguym (commented):
Hey @baum this is looking good but I reviewed from the top and posted my comments with clarifications and suggestions. Thanks!

frontend/src/app/schema/common.js
```diff
@@ -65,6 +65,12 @@ export const port = {
     maximum: 65535
 };
 
+export const multipart = {
```
dannyzaken, line 68: this should be defined in `src/api/common_api.js` I would also change the name to something clearer like `multipart_count` or something similar

src/api/object_api.js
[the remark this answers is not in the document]
dannyzaken: @baum I agree. let's move it to a `multipart_range` object

src/endpoint/s3/s3_utils.js
baum: Thank you for chiming in @guymguym 
The soft condition is OK, which happens in NamespaceS3, where the multipart_range is initialized from HTTP headers.

src/sdk/nb.d.ts
```diff
@@ -395,6 +398,8 @@ interface ObjectInfo {
     delete_marker?: boolean;
     size: number;
     num_parts: number;
+    multipart_start?: number;
```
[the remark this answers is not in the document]
baum: @guymguym, what about it?

src/test/system_tests/test_get_object_part.js
```diff
@@ -0,0 +1,129 @@
...
+assert(process.env.AWS_ACCESS_KEY_ID, "Please set AWS_ACCESS_KEY_ID env. variable, example 'XXXXXXXXXXXXXXXXXXXX'");
+assert(process.env.AWS_SECRET_ACCESS_KEY, "Please set AWS_SECRET_ACCESS_KEY env. variable, example 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'");
+
+const s3 = new AWS.S3({
+    endpoint: process.env.AWS_ENDPOINT, // 'https://s3.amazonaws.com/',
+    accessKeyId: process.env.AWS_ACCESS_KEY_ID,
+    secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY
+});
+
+async function main() {
```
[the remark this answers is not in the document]
baum: Great idea 🤘 @guymguym. The advantage is it tests across regular, namespace, namespace-cache buckets.

src/sdk/nb.d.ts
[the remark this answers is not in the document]
guymguym: should we add it to ObjectInfo?

src/api/common_api.js
```diff
@@ -16,6 +16,13 @@ module.exports = {
 
     definitions: {
 
+
+        multipart: {
```
guymguym, line 20: This identifier is too short and hard to understand what I will find inside.
I would rename to `multipart_num`.

src/api/object_api.js
```diff
@@ -1373,6 +1376,16 @@ module.exports = {
...
+                num_multiparts: {
+                    $ref: 'common_api#/definitions/multipart'
+                },
+                multipart_range: {
+                    type: 'object',
+                    properties: {
+                        start: { type: 'integer' },
+                        end: { type: 'integer' },
+                    }
+                },
```
guymguym, line 1388: This is better off moved to `common_api#/definitions/object_range`

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -52,42 +52,11 @@ async function get_object(req, res) {
...
-            dbg.warn('invalid range', req.headers.range, req.path, obj_size);
-            // let the client know of the relevant range
-            res.setHeader('Content-Range', 'bytes */' + obj_size);
-            throw new S3Error(S3Error.InvalidRange);
-        }
-        throw err;
+    if (multipart_number) {
+        params.multipart_number = multipart_number;
     }
+    s3_utils.s3_range(req, res, object_md, obj_size, params);
```
guymguym, line 58: This function name is neither imperative nor declarative.
I would rename to something like `handle_range` or `set_response_range`.

src/sdk/nb.d.ts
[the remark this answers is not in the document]
baum: @guymguym num_multiparts in ObjectInfo is defined [here](https://github.com/baum/noobaa-core/blob/00623e258bf32f05b0129799241f8628e4aed756/src/sdk/nb.d.ts#L423)

src/test/system_tests/test_get_object_part.js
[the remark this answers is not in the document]
guymguym: I would make the unittests do all of those - there's a huge disadvantage in system tests - they don't run in CI.

src/server/object_services/object_server.js
```diff
@@ -1950,6 +1992,7 @@ async function _complete_object_multiparts(obj, multipart_req) {
     await Promise.all([
         context.parts_updates.length && MDStore.instance().update_parts_in_bulk(context.parts_updates),
         used_multiparts_ids.length && MDStore.instance().update_multiparts_by_ids(used_multiparts_ids, undefined, { uncommitted: 1 }),
+        used_multiparts_ids.length && MDStore.instance().update_object_by_id(obj._id, { num_multiparts: used_multiparts_ids.length }),
```
guymguym, line 1995: When used_multiparts_ids.length is 0, should we really leave this field unset?

src/sdk/nb.d.ts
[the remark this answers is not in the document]
guymguym: ok thanks

src/endpoint/s3/s3_utils.js
guymguym: What do you mean? Can you refer to that code?

src/sdk/namespace_s3.js
```diff
@@ -148,21 +148,24 @@ class NamespaceS3 {
...
             }
             this._set_md_conditions(params, request);
             this._assign_encryption_to_request(params, request);
             let res;
             try {
-                res = can_use_get_inline ?
+                // params.multipart_number was added, since head
+                // does not return Content-Range, which is needed
+                // for GetObject with part num.
+                res = (can_use_get_inline || params.multipart_number) ?
```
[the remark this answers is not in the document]
baum: Well, [HeadObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html) does not contain Content-Range, get [GetObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html) does. Does it still look fishy 🐟 ?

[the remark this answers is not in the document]
guymguym: Yes, this code is an optimization path, which uses GET object instead of HEAD, when there is a chance that this is a small object and we can get the entire data in one trip.
So why does having partNumber in the request causes us to use GET?
We actually avoid GET when partNumber is provided - [here](https://github.com/noobaa/noobaa-core/blob/e04e45f2ab6236ce060a1ea26b0c63bee396aa2e/src/sdk/namespace_s3.js#L157).
So I still don't follow the new logic.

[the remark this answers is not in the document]
guymguym: @romayalon - Can you take a look at this change too - it's a delicate optimization logic and I hope I'm not confusing.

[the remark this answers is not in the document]
baum: @guymguym this is for setting the `Content-Range` field in the response. You could call me if you have more questions...

src/server/object_services/object_server.js
baum: @guymguym, I see no reason why.
Am I missing something?

src/sdk/namespace_s3.js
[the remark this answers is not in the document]
guymguym: Why is it needed for HEAD? I don't see it in S3 API docs -
https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html

src/server/object_services/object_server.js
guymguym: because it will cause us to recalculate it every time

guymguym: isn't that what you do when it's missing?

src/sdk/namespace_s3.js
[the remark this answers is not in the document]
baum: @guymguym The read_object_md() function is called both by GET and HEAD flows.  WDYT?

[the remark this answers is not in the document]
guymguym: I still don't see the relation of that fact to the change you made

src/endpoint/s3/s3_utils.js
baum: @guymguym Sure, [here](https://github.com/baum/noobaa-core/blob/08b3a8cd51e4b1ab65ae3ed2bf33708f80fc8460/src/sdk/namespace_s3.js#L694).

src/server/object_services/object_server.js
baum: @guymguym I guess if used_multiparts_ids.length is 0, then there are no "multiparts". Could you please describe a scenario when used_multiparts_ids.length is 0 and it causes us to recalculate it every time?

src/sdk/namespace_s3.js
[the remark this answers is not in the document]
baum: Please call me, hope I could explain it.

PR #6529, opened 2021-05-23
Added bucket id to NSFS_TEMP_DIR_NAME per bucket

### LINKED ISSUE #6484: NSFS separate the bucket/.noobaa folder per bucket id
### Environment info
- NooBaa Version: **master-20210504**
- Platform: **NA**

### Actual behavior
1. Currently NSFS uses a fixed internal folder name `config.NSFS_TEMP_DIR_NAME='.noobaa'` to manage the bucket state such as uploads and multipart uploads, and this dir is created in each bucket folder.
2. The problem is that if two buckets somehow map to the same filesystem folder, it will mix up the bucket state between these two buckets, for example listing multipart uploads.
3. Even if we disallow mapping buckets to the same folder we still want to resolve this issue to avoid conversion of the dir structure later.

### Expected behavior
1. Buckets mapped to the same filesystem folder should have separate state.
2. We should include the bucket id in the internal dir path.
3. Lookup `config.NSFS_TEMP_DIR_NAME` in the code.

### Steps to reproduce
1. Map two buckets to the same target folder.
2. Start a multipart upload.
3. List multipart uploads from both buckets should show the same upload regardless of which bucket originated it.

romayalon:

### Explain the changes
1. Added _id to read_bucket_sdk_info reply.
2. Added bucket_id to namespaceFS constructor and passed it where new instance of namespaceFS created (including test_namespace_fs.js and nsfs.js standalone).
3. In namespaceFS, replaced config.NSFS_TEMP_DIR_NAME with a call to  get_bucket_tmpdir() which concatinates config.NSFS_TEMP_DIR_NAME and this.bucket_id.

1. Fixed #6484 

### Testing Instructions:
1.

### DIFF
--- a/src/test/unit_tests/test_namespace_fs.js
+++ b/src/test/unit_tests/test_namespace_fs.js
@@ -35,8 +35,8 @@ mocha.describe('namespace_fs', function() {
     const ns_src_bucket_path = `./${src_bkt}`;
     const ns_tmp_bucket_path = `${tmp_fs_path}/${src_bkt}`;
 
-    const ns_src = new NamespaceFS({ bucket_path: ns_src_bucket_path });
-    const ns_tmp = new NamespaceFS({ bucket_path: ns_tmp_bucket_path });
+    const ns_src = new NamespaceFS({ bucket_path: ns_src_bucket_path, bucket_id: '1'});
+    const ns_tmp = new NamespaceFS({ bucket_path: ns_tmp_bucket_path, bucket_id: '2'});
 
     mocha.before(async () => {
         await P.all(_.map([src_bkt, upload_bkt, mpu_bkt], async buck =>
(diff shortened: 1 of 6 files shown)


src/test/unit_tests/test_namespace_fs.js
```diff
@@ -35,8 +35,8 @@ mocha.describe('namespace_fs', function() {
     const ns_src_bucket_path = `./${src_bkt}`;
     const ns_tmp_bucket_path = `${tmp_fs_path}/${src_bkt}`;
 
-    const ns_src = new NamespaceFS({ bucket_path: ns_src_bucket_path });
-    const ns_tmp = new NamespaceFS({ bucket_path: ns_tmp_bucket_path });
+    const ns_src = new NamespaceFS({ bucket_path: ns_src_bucket_path, bucket_id: '000000000000000000000001'});
```
guymguym, line 38: why not just `bucket_id: '1'`?

romayalon: fixed :)

nimrod-becker (approved):

PR #6537, opened 2021-05-27
Update nsfs account config -  changed the uid/gid/new_buckets_path to be optional params 

### LINKED ISSUE #6535: NSFS - request to change the uid/gid as optional params for the account update api
The version of noobaa installed on the HPO cluster
```
# noobaa version
INFO[0000] CLI version: 5.9.0
INFO[0000] noobaa-image: noobaa/noobaa-core:master-20210520
INFO[0000] operator-image: noobaa/noobaa-operator:5.9.0
```

we see with the latest build that the user required to provide all the params of the nsfs_account_config i.e. uid, gid and new_buckets_path. 
```
# noobaa api account_api update_account '{"email":"hpouser","nsfs_account_config":{"uid":304,"gid":403,"new_buckets_path":"/nsfs/fs1/hpouser"}}'
INFO[0001] ✅ Exists: NooBaa "noobaa"
INFO[0001] ✅ Exists: Service "noobaa-mgmt"
INFO[0001] ✅ Exists: Secret "noobaa-operator"
INFO[0001] ✅ Exists: Secret "noobaa-admin"
INFO[0001] ✈️  RPC: account.update_account() Request: map[email:hpouser nsfs_account_config:map[gid:403 new_buckets_path:/nsfs/fs1/hpouser uid:304]]
WARN[0001] RPC: GetConnection creating connection to wss://localhost:41979/rpc/ 0xc00003b4f0
INFO[0001] RPC: Connecting websocket (0xc00003b4f0) &{RPC:0xc0000d1400 Address:wss://localhost:41979/rpc/ State:init WS:<nil> PendingRequests:map[] NextRequestID:0 Lock:{state:1 sema:0} ReconnectDelay:0s}
INFO[0001] RPC: Connected websocket (0xc00003b4f0) &{RPC:0xc0000d1400 Address:wss://localhost:41979/rpc/ State:init WS:<nil> PendingRequests:map[] NextRequestID:0 Lock:{state:1 sema:0} ReconnectDelay:0s}
INFO[0001] ✅ RPC: account.update_account() Response OK: took 87.2ms
null
```
From HPO CLI, we would like to only allow user to update the new_buckets_path for an existing account. Keeping in the uid , gid same as before. So, this is breaking current HPO CLI/REST implementation.

We will like make uid/gid as optional params, user doesn't need to update the uid/gid when updating `new_buckets_path`
```
# noobaa api account_api update_account '{"email":"hpouser","nsfs_account_config":{"new_buckets_path":"/nsfs/fs1/hpouser"}}'
```

romayalon:

### Explain the changes
1. removed required from API of update_account
2. Added check in update_account function for nsfs_account_config containing at least one of uid, gid, new_buckets_path.
3. added tests.

1. Fixed #6535 

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -475,14 +475,17 @@ function update_account(req) {
     if (params.ips && !_.every(params.ips, ip_range => (net.isIP(ip_range.start) && net.isIP(ip_range.end)))) {
         throw new RpcError('FORBIDDEN', 'Non valid IPs');
     }
-
+    if (req.rpc_params.nsfs_account_config && _.isUndefined(req.rpc_params.nsfs_account_config.uid) &&
+            _.isUndefined(req.rpc_params.nsfs_account_config.gid) && !req.rpc_params.nsfs_account_config.new_buckets_path) {
+                throw new RpcError('FORBIDDEN', 'Invalid nsfs_account_config');
+    }
     let updates = {
         name: params.name,
         email: params.new_email,
         next_password_change: params.must_change_password === true ? new Date() : undefined,
         allowed_ips: (!_.isUndefined(params.ips) && params.ips !== null) ? params.ips : undefined,
         preferences: _.isUndefined(params.preferences) ? undefined : params.preferences,
-        nsfs_account_config: req.rpc_params.nsfs_account_config
+        nsfs_account_config: req.rpc_params.nsfs_account_config && _.assign(account.nsfs_account_config, req.rpc_params.nsfs_account_config)
     };
 
     let removals = {
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -187,7 +187,12 @@ module.exports = {
                         }]
                     },
                     nsfs_account_config: {
-                        $ref: 'common_api#/definitions/nsfs_account_config'
+                        type: 'object',
+                        properties: {
+                            uid: { type: 'number' },
+                            gid: { type: 'number' },
+                            new_buckets_path: { type: 'string' },
+                        }
                     },
                     preferences: {
                         type: 'object',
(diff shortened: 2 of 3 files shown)


nimrod-becker (approved):

src/server/system_services/account_server.js
```diff
@@ -475,14 +475,17 @@ function update_account(req) {
...
+                throw new RpcError('FORBIDDEN', 'Invalid nsfs_account_config');
+    }
     let updates = {
         name: params.name,
         email: params.new_email,
         next_password_change: params.must_change_password === true ? new Date() : undefined,
         allowed_ips: (!_.isUndefined(params.ips) && params.ips !== null) ? params.ips : undefined,
         preferences: _.isUndefined(params.preferences) ? undefined : params.preferences,
-        nsfs_account_config: req.rpc_params.nsfs_account_config
+        nsfs_account_config: req.rpc_params.nsfs_account_config && _.assign(account.nsfs_account_config, req.rpc_params.nsfs_account_config)
```
guymguym, line 488: @romayalon You shouldn't be trying to update the actual account which is in system store.
Instead, create a new object like this -
```
        nsfs_account_config: req.rpc_params.nsfs_account_config && {
            // first take existing configs from account
            ...account.nsfs_account_config,
            // then override with request configs
            ...req.rpc_params.nsfs_account_config
        }
```

guymguym: the reason why updating the account directly is bad is that this update was not yet applied to the system store, and we don't want any other request during that time to see that change.
I think at some point we will want to make the system store objects frozen (`js_utils.deep_freeze`) which will cause such calls to fail (on runtime).

romayalon: thanks for your input, I fixed it in https://github.com/noobaa/noobaa-core/pull/6542

PR #6548, opened 2021-06-02
Addition of NSFS MD

### LINKED ISSUE #6426: NSFS use posix xattr to get/set custom object meta-data
### Environment info
- NooBaa Version: **5.9**
- Platform: **NA**

### Actual behavior
1. Currently custom meta-data (i.e. x-amz-meta-XXX headers) are not handled in PUT and not returned on HEAD/GET.

### Expected behavior
1. Should store and retrieve custom meta-data using posix xattr.
2. Should also support the copy-object directives for meta-data (replace/copy).

jeniawhite:
### Explain the changes
1. Adding the option to work with the user metadata for NSFS

1. Fixed https://issues.redhat.com/browse/MCGI-25
2. Fixed https://github.com/noobaa/noobaa-core/issues/6426

### Testing Instructions:
1.

### DIFF
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -28,6 +28,8 @@ const DEFAULT_OBJECT_ACL = Object.freeze({
     }]
 });
 
+const XATTR_SORT_SYMBOL = Symbol('XATTR_SORT_SYMBOL');
+
 const OP_NAME_TO_ACTION = Object.freeze({
     delete_bucket_analytics: { regular: "s3:putanalyticsconfiguration" },
     delete_bucket_cors: { regular: "s3:putbucketcors" },
@@ -131,9 +133,31 @@ function get_request_xattr(req) {
 }
 
 function set_response_xattr(res, xattr) {
-    _.each(xattr, (val, key) => {
-        res.setHeader(X_AMZ_META + key, val);
-    });
+    let size_for_md_left = config.S3_MD_SIZE_LIMIT;
+    const keys = Object.keys(xattr);
+    // Using a special symbol on the xattr object to specify when sorting of keys is needed
+    // which will make sure the returned headers are consistently being truncated.
+    // The symbol is set by namespace_fs and could be used in other namespaces where
+    // xattrs are more likely to exceed the size limit and therefore sorting is desired,
+    // but for most namespaces where the size is also imposed on PUT, we can avoid it.
+    if (xattr[XATTR_SORT_SYMBOL]) {
+        keys.sort();
+    }
+    let returned_keys = 0;
+    for (const key of keys) {
+        const md_header_size =
+            X_AMZ_META.length +
+            4 + // for ': ' and '\r\n'
+            Buffer.byteLength(key, 'utf8') +
+            Buffer.byteLength(xattr[key], 'utf8');
+        if (md_header_size > size_for_md_left) {
+            res.setHeader('x-amz-missing-meta', keys.length - returned_keys);
+            break;
+        }
+        returned_keys += 1;
+        size_for_md_left -= md_header_size;
+        res.setHeader(X_AMZ_META + key, xattr[key]);
+    }
 }
 
 function parse_etag(etag, err) {
@@ -657,3 +681,4 @@ exports._is_valid_retention = _is_valid_retention;
 exports.get_http_response_from_resp = get_http_response_from_resp;
 exports.get_http_response_date = get_http_response_date;
 exports.has_bucket_policy_permission = has_bucket_policy_permission;
+exports.XATTR_SORT_SYMBOL = XATTR_SORT_SYMBOL;
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -3,6 +3,7 @@
 #include "../util/common.h"
 #include "../util/napi.h"
 #include "../util/os.h"
+#include "../util/buf.h"
 
 #include <dirent.h>
 #include <fcntl.h>
@@ -16,6 +17,7 @@
 #include <typeinfo>
 #include <unistd.h>
 #include <uv.h>
+#include <sys/xattr.h>
 #include <vector>
 
 namespace noobaa
@@ -54,6 +56,7 @@ struct Entry
     ino_t ino;
     uint8_t type;
 };
+
 
 template <typename T>
 static Napi::Value
@@ -223,6 +226,7 @@ struct Stat : public FSWorker
         _deferred.Resolve(res);
     }
 };
+
 
 /**
  * CheckAccess is an fs op
@@ -417,18 +421,23 @@ struct Writefile : public FSWorker
     std::string _path;
     const uint8_t* _data;
     size_t _len;
+    mode_t _mode;
     Writefile(const Napi::CallbackInfo& info)
         : FSWorker(info)
+        , _mode(0666)
     {
         _path = info[1].As<Napi::String>();
         auto buf = info[2].As<Napi::Buffer<uint8_t>>();
         _data = buf.Data();
         _len = buf.Length();
-        Begin(XSTR() << "Writefile " << DVAL(_path) << DVAL(_len));
+        if (info.Length() > 3 && !info[3].IsUndefined()) {
+            _mode = info[3].As<Napi::Number>().Uint32Value();
+        }
+        Begin(XSTR() << "Writefile " << DVAL(_path) << DVAL(_len) << DVAL(_mode));
     }
     virtual void Work()
     {
-        int fd = open(_path.c_str(), O_WRONLY | O_CREAT);
+        int fd = open(_path.c_str(), O_TRUNC | O_CREAT | O_WRONLY, _mode);
         if (fd < 0) {
             SetSyscallError();
             return;
@@ -600,6 +609,8 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
                 InstanceMethod<&FileWrap::read>("read"),
                 InstanceMethod<&FileWrap::write>("write"),
                 InstanceMethod<&FileWrap::writev>("writev"),
+                InstanceMethod<&FileWrap::setxattr>("setxattr"),
+                InstanceMethod<&FileWrap::getxattr>("getxattr"),
             }));
         constructor.SuppressDestruct();
     }
@@ -618,6 +629,8 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
     Napi::Value read(const Napi::CallbackInfo& info);
     Napi::Value write(const Napi::CallbackInfo& info);
     Napi::Value writev(const Napi::CallbackInfo& info);
+    Napi::Value setxattr(const Napi::CallbackInfo& info);
+    Napi::Value getxattr(const Napi::CallbackInfo& info);
 };
 
 Napi::FunctionReference FileWrap::constructor;
@@ -788,6 +801,148 @@ struct FileWritev : public FSWrapWorker<FileWrap>
     }
 };
 
+/**
+ * TODO: Not atomic and might cause partial updates of MD
+ */
+struct FileSetxattr : public FSWrapWorker<FileWrap>
+{
+    std::map<std::string, std::string> _md;
+    FileSetxattr(const Napi::CallbackInfo& info)
+        : FSWrapWorker<FileWrap>(info)
+    {
+        auto md = info[1].As<Napi::Object>();
+        auto keys = md.GetPropertyNames();
+        for (uint32_t i = 0; i < keys.Length(); ++i) {
+            auto key = keys.Get(i).ToString().Utf8Value();
+            auto value = md.Get(key).ToString().Utf8Value();
+            _md[key] = value;
+        }
+        Begin(XSTR() << "FileSetxattr " << DVAL(_wrap->_path));
+    }
+    virtual void Work()
+    {
+        int fd = _wrap->_fd;
+        for (auto it = _md.begin(); it != _md.end(); ++it) {
+            #ifdef __APPLE__
+                auto r = fsetxattr(fd, it->first.c_str(), it->second.c_str(), it->second.length(), 0, 0);
+            #else
+                auto r = fsetxattr(fd, it->first.c_str(), it->second.c_str(), it->second.length(), 0);
+            #endif
+            if (r == -1) {
+                SetSyscallError();
+                return;
+            }
+        }
+    }
+};
+
+struct FileGetxattr : public FSWrapWorker<FileWrap>
+{
+    std::map<std::string, std::string> _md;
+    FileGetxattr(const Napi::CallbackInfo& info)
+        : FSWrapWorker<FileWrap>(info)
+    {
+        Begin(XSTR() << "FileGetxattr " << DVAL(_wrap->_path));
+    }
+    virtual void Work()
+    {
+        int fd = _wrap->_fd;
+        // Serves as a "lock" mechanism when reading the xattrs of a file
+        // Since this is non atomic operation, we abort if ctime of file changed
+        struct stat stat_res;
+        auto r = fstat(fd, &stat_res);
+        if (r) {
+            SetSyscallError();
+            return;
+        }
+        auto start_ctime = stat_res.st_ctime;
+
+        #ifdef __APPLE__
+            ssize_t buf_len = flistxattr(fd, NULL, 0, 0);
+        #else
+            ssize_t buf_len = flistxattr(fd, NULL, 0);
+        #endif
+
+        // No xattr, nothing to do  
+        if (buf_len == 0) return;
+
+        if (buf_len == -1) {
+            SetSyscallError();
+            return;
+        }
+
+        Buf buf(buf_len);
+
+        #ifdef __APPLE__
+            buf_len = flistxattr(fd, buf.cdata(), buf_len, 0);
+        #else
+            buf_len = flistxattr(fd, buf.cdata(), buf_len);
+        #endif
+
+        if (buf_len == -1) {
+            SetSyscallError();
+            return;
+        }
+
+        while (buf.length()) {
+            std::string key(buf.cdata());
+
+            #ifdef __APPLE__
+                ssize_t value_len = fgetxattr(fd, key.c_str(), NULL, 0, 0, 0);
+            #else
+                ssize_t value_len = fgetxattr(fd, key.c_str(), NULL, 0);
+            #endif
+
+            if (value_len == -1) {
+                SetSyscallError();
+                return;
+            }
+
+            if (value_len > 0) {
+                Buf val(value_len);
+
+                #ifdef __APPLE__
+                    value_len = fgetxattr(fd, key.c_str(), val.cdata(), value_len, 0, 0);
+                #else
+                    value_len = fgetxattr(fd, key.c_str(), val.cdata(), value_len);
+                #endif
+
+                if (value_len == -1) {
+                    SetSyscallError();
+                    return;
+                }
+                
+                _md[key] = std::string(val.cdata(), value_len);
+            } else if (value_len == 0) {
+                _md[key] = "";
+            }
+
+            buf.slice(key.size() + 1, buf.length());
+        }
+
+        r = fstat(fd, &stat_res);
+        if (r) {
+            SetSyscallError();
+            return;
+        }
+        auto end_ctime = stat_res.st_ctime;
+        if (start_ctime != end_ctime) {
+            // Cancel on ctime change due to non atomic nature of this call
+            SetError(XSTR() << "FileGetxattr: " << DVAL(_wrap->_path) << " cancelled due to ctime change");
+        }
+    }
+    virtual void OnOK()
+    {
+        DBG1("FS::FSWorker::OnOK: FileGetxattr " << DVAL(_wrap->_path));
+        Napi::Env env = Env();
+        Napi::Object res = Napi::Object::New(env);
+        for (auto it = _md.begin(); it != _md.end(); ++it) {
+            res.Set(it->first, it->second);
+        }
+        _deferred.Resolve(res);
+    }
+};
+
 Napi::Value
 FileWrap::close(const Napi::CallbackInfo& info)
 {
@@ -810,6 +965,18 @@ Napi::Value
 FileWrap::writev(const Napi::CallbackInfo& info)
 {
     return api<FileWritev>(info);
+}
+
+Napi::Value
+FileWrap::setxattr(const Napi::CallbackInfo& info)
+{
+    return api<FileSetxattr>(info);
+}
+
+Napi::Value
+FileWrap::getxattr(const Napi::CallbackInfo& info)
+{
+    return api<FileGetxattr>(info);
 }
 
 /**
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -13,6 +13,7 @@ const config = require('../../config');
 const s3_utils = require('../endpoint/s3/s3_utils');
 const stream_utils = require('../util/stream_utils');
 const buffer_utils = require('../util/buffer_utils');
+// const http_utils = require('../util/http_utils');
 const LRUCache = require('../util/lru_cache');
 const Semaphore = require('../util/semaphore');
 const nb_native = require('../util/nb_native');
@@ -20,6 +21,8 @@ const RpcError = require('../rpc/rpc_error');
 
 const buffers_pool_sem = new Semaphore(config.NSFS_BUF_POOL_MEM_LIMIT);
 const buffers_pool = new buffer_utils.BuffersPool(config.NSFS_BUF_SIZE, buffers_pool_sem);
+
+const XATTR_USER_PREFIX = 'user.';
 
 /**
  * @param {fs.Dirent} a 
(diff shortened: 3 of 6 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -55,6 +56,16 @@ struct Entry
...
 };
 
+
+struct KeyValue {
+    std::string key;
+    std::string value;
+};
+
+
+std::string USER_MD_PREFIX = "user.s3.";
```
guymguym, line 66: I think we better move it to config.js and send as arg to native calls.
We want this to be easily configurable. Certainly in the current stage.

src/native/fs/fs_napi.cpp
```diff
@@ -219,6 +230,258 @@ struct Stat : public FSWorker
...
+    {
+        _path = info[1].As<Napi::String>();
+        Begin(XSTR() << "Getxattr " << DVAL(_path));
+    }
+    virtual void Work()
+    {
+        ssize_t buf_len, key_len, value_len;
+        char *buf, *key, *val;
+
+        #if !defined(_POSIX_C_SOURCE) || defined(_DARWIN_C_SOURCE)
```
guymguym, line 246:
```suggestion
#ifdef __APPLE__
```

src/native/fs/fs_napi.cpp
```diff
@@ -219,6 +230,258 @@ struct Stat : public FSWorker
...
+                buf_len = listxattr(_path.c_str(), NULL, 0, 0);
+        #else
+                buf_len = listxattr(_path.c_str(), NULL, 0);
+        #endif
+
+        if (buf_len <= 0) {
+            if (buf_len == -1) {
```
guymguym, line 253: What is the else case? Why is it just ignored? Even if this code is correct, it is not trivial and requires a comment.

src/native/fs/fs_napi.cpp
```diff
@@ -219,6 +230,258 @@ struct Stat : public FSWorker
...
+            return;
+        }
+
+        struct stat stat_res;
+        auto r = stat(_path.c_str(), &stat_res);
+        if (r) {
+            SetSyscallError();
+            return;
+        }
+        auto start_ctime = stat_res.st_ctime;
```
guymguym, line 265: Shouldn't this be taken before the first listxattr? Why after?

src/native/fs/fs_napi.cpp
```diff
@@ -219,6 +230,258 @@ struct Stat : public FSWorker
...
+
+        buf = static_cast<char*>(malloc(buf_len));
```
guymguym, line 267: This is leak prone.
For safe memory allocation use `Buf buf(len)` - see [here](https://github.com/noobaa/noobaa-core/blob/master/src/native/util/buf.h)
For more complex cases you can use [StackCleaner](https://github.com/noobaa/noobaa-core/blob/f1b3a1b40262b88a01a647483d0d097d8813f777/src/native/util/common.h#L120), but this case is simple

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +224,258 @@ struct Stat : public FSWorker
...
+ */
+struct Getxattr : public FSWorker
+{
+    std::string _path;
+    std::vector<KeyValue> _md;
+    Getxattr(const Napi::CallbackInfo& info) 
+        : FSWorker(info)
+        , _md(0)
+    {
+        _path = info[1].As<Napi::String>();
```
guymguym, line 238: I would send the prefix as arg (optional) here, so this will filter only the xattrs matching the prefix.

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +224,258 @@ struct Stat : public FSWorker
...
+        Begin(XSTR() << "Getxattr " << DVAL(_path));
+    }
+    virtual void Work()
+    {
+        ssize_t buf_len, key_len, value_len;
+        char *buf, *key, *val;
```
guymguym, line 244: This is unsafe c style. 
Best move closed to scope that uses these to avoid dangling pointers.

src/native/fs/fs_napi.cpp
guymguym: I think you meant to identify the `buf_len == 0` case which means there are no xattrs at all?
Better be explicit about it.

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +224,258 @@ struct Stat : public FSWorker
...
+        r = stat(_path.c_str(), &stat_res);
+        if (r) {
+            SetSyscallError();
+            return;
+        }
+        auto end_ctime = stat_res.st_ctime;
+        if (start_ctime != end_ctime) {
+            // Cancel on ctime change due to non atomic nature of this call
+            errno = ECANCELED;
+            SetSyscallError();
```
guymguym, line 356: This is not a syscall error so why overloading this ECANCELED error?
Why not just use a SetError?

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +224,258 @@ struct Stat : public FSWorker
...
+            return;
+        }
+
+        struct stat stat_res;
+        auto r = stat(_path.c_str(), &stat_res);
+        if (r) {
+            SetSyscallError();
+            return;
+        }
+        auto start_ctime = stat_res.st_ctime;
```
guymguym, line 265: You have to include a comment on this. It cannot be bare code like this without any reasoning.

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +224,258 @@ struct Stat : public FSWorker
     }
 };
 
+/**
+ * Getxattr is an fs op
+ */
+struct Getxattr : public FSWorker
+{
+    std::string _path;
+    std::vector<KeyValue> _md;
```
guymguym, line 233: Why vector?? why not `std:map<string,string>`?

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +224,258 @@ struct Stat : public FSWorker
...
+
+        if (buf_len == -1) {
+            SetSyscallError();
+            free(buf);
+            return;
+        }
+
+        key = buf;
+        while (buf_len > 0) {
+            // strncmp shouldn't cause error if length of key < USER_MD_PREFIX
```
guymguym, line 287: You should construct an std::string for every key right at the start.
Since std::string can be constructed from null terminated c strings, this is as simple as:
```
std::string key(buf);
buf += key.size() + 1;
buf_len -= key.size() + 1;
```

guymguym (changes requested):
@jeniawhite The code needs more love and care.
Please try to make it more robust in terms of adding comments that explain the non trivial things, avoid unautomated memory allocations, and prefer c++ strings over c strings.

src/native/fs/fs_napi.cpp
jeniawhite: Did not want to trigger stat call if we have no xattrs.
But I will move it to avoid a case when we read that we have xattrs and take the time after reading their size.

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +216,182 @@ struct Stat : public FSWorker
...
+        #else
+            buf_len = listxattr(_path.c_str(), buf.cdata(), buf_len);
+        #endif
+
+        if (buf_len == -1) {
+            SetSyscallError();
+            return;
+        }
+
+        while (buf_len > 0) {
```
guymguym, line 278:
```suggestion
        while (buf.length()) {
```

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +216,182 @@ struct Stat : public FSWorker
...
+                    value_len = getxattr(_path.c_str(), key.c_str(), val.cdata(), value_len);
+                #endif
+
+                if (value_len == -1) {
+                    SetSyscallError();
+                    return;
+                }
+                
+                val[value_len] = 0;
+                _md[key] = std::string(val.cdata());
```
guymguym, line 318: send the length to string ctor and then you don't need to add null terminator and it won't need to find its length on its own:
```suggestion
                _md[key] = std::string(val.cdata(), value_len);
```

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +216,182 @@ struct Stat : public FSWorker
...
+            } else if (value_len == 0) {
+                _md[key] = "";
+            }
+
+            buf_len -= key.size() + 1;
+            buf.slice(key.size() + 1, buf_len);
```
guymguym, line 324: slice will already reduce the buf.length() appropriately, so no need to track the remaining characters also in the buf_len variable - see that I commented also on the while loop condition.

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +216,182 @@ struct Stat : public FSWorker
...
+        }
+    }
+    virtual void OnOK()
+    {
+        DBG1("FS::FSWorker::OnOK: Getxattr " << DVAL(_path));
+        Napi::Env env = Env();
+        Napi::Object res = Napi::Object::New(env);
+        for (auto it = _md.begin(); it != _md.end(); ++it) {
+            std::string key = it->first;
+            if (key.rfind(_md_prefix, 0) == 0) key.erase(0, _md_prefix.size());
```
guymguym, line 345: I would move all these manipulations to js. Easier to write and more dynamic.

guymguym (changes requested):

src/endpoint/s3/s3_utils.js
```diff
@@ -131,7 +131,22 @@ function get_request_xattr(req) {
 }
 
 function set_response_xattr(res, xattr) {
-    _.each(xattr, (val, key) => {
+    let size_for_md_left = config.S3_MD_SIZE_LIMIT;
+    const sliced_xattr = {};
+    const sorted_xattr = Object.keys(xattr).sort().reduce((obj, key) => {
```
guymguym, line 136: The sort order is lost once you add keys to object.
I suggest you just keep a sorted list of the keys, and when you need the value to just get it from xattr[key].
```js
const sorted_keys = Object.keys(xattr).sort();
```

src/endpoint/s3/s3_utils.js
```diff
@@ -131,7 +131,22 @@ function get_request_xattr(req) {
...
+    for (const [key, value] of Object.entries(sorted_xattr)) {
+        const property_size = Buffer.byteLength(key, 'utf8') + Buffer.byteLength(value, 'utf8');
+        if (property_size > size_for_md_left) {
+            res.setHeader('x-amz-missing-meta', Object.keys(sorted_xattr).length - Object.keys(sliced_xattr).length);
+            break;
+        }
+        size_for_md_left -= property_size;
+        sliced_xattr[key] = value;
+    }
+    _.each(sliced_xattr, (val, key) => {
```
guymguym, line 149: Why is this _.each still needed? You can simply set the headers in the loop.

src/endpoint/s3/s3_utils.js
```diff
@@ -131,7 +131,22 @@ function get_request_xattr(req) {
...
+    let size_for_md_left = config.S3_MD_SIZE_LIMIT;
+    const sliced_xattr = {};
+    const sorted_xattr = Object.keys(xattr).sort().reduce((obj, key) => {
+        obj[key] = xattr[key];
+        return obj;
+    }, {});
+    for (const [key, value] of Object.entries(sorted_xattr)) {
+        const property_size = Buffer.byteLength(key, 'utf8') + Buffer.byteLength(value, 'utf8');
+        if (property_size > size_for_md_left) {
+            res.setHeader('x-amz-missing-meta', Object.keys(sorted_xattr).length - Object.keys(sliced_xattr).length);
```
guymguym, line 143: Better maintain this count instead of regenerate whole lists of keys just to get the length.

src/sdk/namespace_fs.js
```diff
@@ -406,17 +475,23 @@ class NamespaceFS {
...
             await this._load_bucket(params, fs_account_config);
             const file_path = this._get_file_path(params);
             const stat = await nb_native().fs.stat(fs_account_config, file_path);
             dbg.log0(file_path, stat);
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
-            return this._get_object_info(params.bucket, params.key, stat);
+            check_md_conditions(params.md_conditions, stat);
+            target_file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
+            const xattr = get_transformed_xattr(await target_file.getxattr(fs_account_config));
+            return this._get_object_info(params.bucket, params.key, Object.assign(stat, { xattr }));
```
guymguym, line 489: I prefer not to hang xattr on the stat object. Just pass as another param to get_object_info.

src/sdk/namespace_fs.js
```diff
@@ -565,8 +647,20 @@ class NamespaceFS {
             }
             // TODO use file xattr to store md5_b64 xattr, etc.
             const stat = await nb_native().fs.stat(fs_account_config, upload_path);
+            if (params.xattr) {
```
guymguym, line 650: isn't it always defined, but usually an empty object?

src/sdk/namespace_fs.js
```diff
@@ -71,6 +72,74 @@ function make_named_dirent(name) {
...
+        }
+    }
+    if (conditions.if_unmodified_since) {
+        if (!matched && (!stat || conditions.if_unmodified_since < data.last_modified)) {
+            throw Object.assign(new Error('check_md_conditions failed'), { rpc_code: 'IF_UNMODIFIED_SINCE', rpc_data: data });
+        }
+    }
+}
+
+function get_transformed_xattr(xattr, add_user_prefix) {
```
guymguym, line 131: This should be split to two functions, instead of passing a flag that changes completely what the function does.

src/sdk/namespace_fs.js
```diff
@@ -71,6 +72,74 @@ function make_named_dirent(name) {
...
+    if (add_user_prefix) {
+        return _.transform(xattr, (result, value, key) => {
+            result[`user.${String(key)}`] = value;
```
guymguym, line 134: Why is 'user.' hard coded like this? Make a const/config.

src/sdk/namespace_fs.js
```diff
@@ -71,6 +72,74 @@ function make_named_dirent(name) {
...
+        }, {});
+    } else {
+        return _.transform(xattr, (result, value, key) => {
+            const key_str = String(key);
+            if (key_str.startsWith('user.')) result[key_str.replace('user.', '')] = value;
```
guymguym, line 139:
```suggestion
            if (key_str.startsWith(XATTR_USER_PREFIX)) result[key_str.slice(XATTR_USER_PREFIX.length)] = value;
```

src/sdk/namespace_fs.js
```diff
@@ -406,17 +475,23 @@ class NamespaceFS {
...
-            const fs_account_config = this.set_cur_fs_account_config(object_sdk);
             await this._load_bucket(params, fs_account_config);
             const file_path = this._get_file_path(params);
             const stat = await nb_native().fs.stat(fs_account_config, file_path);
             dbg.log0(file_path, stat);
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
-            return this._get_object_info(params.bucket, params.key, stat);
+            check_md_conditions(params.md_conditions, stat);
+            target_file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
+            const xattr = get_transformed_xattr(await target_file.getxattr(fs_account_config));
```
guymguym, line 488: I would push the get_transformed_xattr() call into _get_object_info().

src/sdk/namespace_fs.js
```diff
@@ -565,8 +647,20 @@ class NamespaceFS {
             }
             // TODO use file xattr to store md5_b64 xattr, etc.
             const stat = await nb_native().fs.stat(fs_account_config, upload_path);
+            if (params.xattr) {
+                let target_file;
+                try {
+                    target_file = await nb_native().fs.open(fs_account_config, upload_path, 'a', get_umasked_mode(config.BASE_MODE_FILE));
```
guymguym, line 653: I would push this logic to a function and call it from inside _upload_stream and _copy_stream since it already has the target_file open there.

guymguym (changes requested):
See comments.
I really think the inclusion of check_md_conditions() in this PR is not a good idea.
It is complex enough to code and review each of these features alone.

src/endpoint/s3/s3_utils.js
```diff
@@ -131,9 +131,19 @@ function get_request_xattr(req) {
...
 
 function set_response_xattr(res, xattr) {
-    _.each(xattr, (val, key) => {
-        res.setHeader(X_AMZ_META + key, val);
-    });
+    let size_for_md_left = config.S3_MD_SIZE_LIMIT;
+    const sorted_keys = Object.keys(xattr).sort();
+    let returned_keys = 0;
+    for (const key of sorted_keys) {
+        const property_size = Buffer.byteLength(key, 'utf8') + Buffer.byteLength(xattr[key], 'utf8');
```
guymguym, line 138: I think you should include the actual http header size, which is: `X_AMZ_META + key + ': ' + value + '\n'`
which means:
```suggestion
        const md_header_size = 
            X_AMZ_META.length +
            3 + // for ': ' and '\n'
            Buffer.byteLength(key, 'utf8') +
            Buffer.byteLength(xattr[key], 'utf8');
```

src/sdk/namespace_fs.js
```diff
@@ -618,6 +691,7 @@ class NamespaceFS {
                 buffer_pool_cleanup = null;
                 callback();
             }
+            if (xattr) await target_file.setxattr(fs_account_config, get_xattr_write(xattr));
```
guymguym, line 694:
```suggestion
            if (fs_xattr) {
                await target_file.setxattr(fs_account_config, fs_xattr);
            }
```

src/sdk/namespace_fs.js
```diff
@@ -543,8 +608,16 @@ class NamespaceFS {
...
+        const xattr = _.isEmpty(params.xattr) ? undefined : params.xattr;
         // dbg.log0('NamespaceFS.upload_object:', upload_path, '->', file_path);
         try {
+            // try {
+            //     const stat = await nb_native().fs.stat(fs_account_config, file_path);
+            //     check_md_conditions(params.md_conditions, stat);
+            // } catch (err) {
+            //     // Skip check for non existing objects
+            //     if (err.code !== 'ENOENT') throw err;
+            // }
```
guymguym, line 620: delete

src/sdk/namespace_fs.js
```diff
@@ -543,8 +608,16 @@ class NamespaceFS {
         // Uploading to a temp file then we will rename it. 
         const upload_id = uuidv4();
         const upload_path = path.join(this.bucket_path, this.get_bucket_tmpdir(), 'uploads', upload_id);
+        const xattr = _.isEmpty(params.xattr) ? undefined : params.xattr;
```
guymguym, line 611: If we are making fs_xattr then lets call it like that, and also move the empty check into the function to make it easier to prepare:
```suggestion
        const fs_xattr = get_fs_xattr_for_write(params.xattr);
```

src/sdk/namespace_fs.js
```diff
@@ -71,6 +74,63 @@ function make_named_dirent(name) {
...
+// }
+
+function get_xattr_read(xattr) {
+    return _.transform(xattr, (result, value, key) => {
+        const key_str = String(key);
+        if (key_str.startsWith(XATTR_USER_PREFIX)) result[key_str.slice(XATTR_USER_PREFIX.length)] = value;
+    }, {});
+}
+
+function get_xattr_write(xattr) {
```
guymguym, line 128: Rename and add the empty check inside:
```suggestion
function get_fs_xattr_for_write(xattr) {
    if (_.isEmpty(xattr)) return undefined;
```

src/sdk/namespace_fs.js
```diff
@@ -71,6 +74,63 @@ function make_named_dirent(name) {
...
+    return _.transform(xattr, (result, value, key) => {
+        result[`${XATTR_USER_PREFIX}${String(key)}`] = value;
+    }, {});
```
guymguym, line 131: Not sure why you need the explicit cast to string here, but you can use mapKeys which is easier:
```suggestion
    return _.mapKeys(xattr, (val, key) => XATTR_USER_PREFIX + key);
```

src/sdk/namespace_fs.js
```diff
@@ -71,6 +74,63 @@ function make_named_dirent(name) {
...
+//         }
+//     }
+// }
+
+function get_xattr_read(xattr) {
+    return _.transform(xattr, (result, value, key) => {
+        const key_str = String(key);
+        if (key_str.startsWith(XATTR_USER_PREFIX)) result[key_str.slice(XATTR_USER_PREFIX.length)] = value;
+    }, {});
+}
```
guymguym, line 126:
```suggestion
function get_xattr_from_fs(fs_xattr) {
    const xattr = _.mapKeys(fs_xattr, (val, key) => 
        key.startsWith(XATTR_USER_PREFIX) ?
        key.slice(XATTR_USER_PREFIX.length) :
        '';
    // keys which do not start with prefix will all map to the empty string key, so we remove it once
    delete xattr[''];
    return xattr;
}
```

src/sdk/namespace_fs.js
```diff
@@ -71,6 +74,63 @@ function make_named_dirent(name) {
     return entry;
 }
 
+// function check_md_conditions(conditions, stat) {
+//     if (!conditions) return;
+//     if (!conditions.if_match_etag &&
+//         !conditions.if_none_match_etag &&
+//         !conditions.if_modified_since &&
+//         !conditions.if_unmodified_since) return;
```
guymguym, line 82: delete. keep somewhere else.

src/native/fs/fs_napi.cpp
```diff
@@ -473,7 +487,7 @@ struct Readfile : public FSWorker
     }
     virtual void Work()
     {
-        int fd = open(_path.c_str(), O_RDONLY);
+        int fd = open(_path.c_str(), O_RDONLY, _mode);
```
guymguym, line 490: Maybe I'm missing something here, but why do you need to pass a mode to open on read? Isn't that only relevant when you use O_CREAT?

src/sdk/namespace_fs.js
```diff
@@ -406,17 +466,22 @@ class NamespaceFS {
...
-            const fs_account_config = this.set_cur_fs_account_config(object_sdk);
             await this._load_bucket(params, fs_account_config);
             const file_path = this._get_file_path(params);
             const stat = await nb_native().fs.stat(fs_account_config, file_path);
-            dbg.log0(file_path, stat);
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
-            return this._get_object_info(params.bucket, params.key, stat);
+            // check_md_conditions(params.md_conditions, stat);
+            target_file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
+            const xattr = await target_file.getxattr(fs_account_config);
```
guymguym, line 478: Name it fs_xattr so we can differentiate the two different types of xattr object we use (fs / s3)
```suggestion
            const fs_xattr = await target_file.getxattr(fs_account_config);
```

src/sdk/namespace_fs.js
```diff
@@ -406,17 +466,22 @@ class NamespaceFS {
     /////////////////
 
     async read_object_md(params, object_sdk) {
+        let target_file;
```
guymguym, line 469: rename to file, doesn't make sense to call this target_file in operations where there is no source file
```suggestion
        let file;
```

src/sdk/namespace_fs.js
```diff
@@ -980,7 +1069,7 @@ class NamespaceFS {
             is_latest: true,
             delete_marker: false,
             tag_count: 0,
-            xattr: {},
+            xattr: _.isEmpty(xattr) ? {} : get_xattr_read(xattr),
```
guymguym, line 1072: move this _.isEmpty check into the function
```suggestion
            xattr: get_xattr_from_fs(fs_xattr),
```

src/endpoint/s3/s3_utils.js
```diff
@@ -131,9 +131,19 @@ function get_request_xattr(req) {
 }
 
 function set_response_xattr(res, xattr) {
-    _.each(xattr, (val, key) => {
-        res.setHeader(X_AMZ_META + key, val);
-    });
+    let size_for_md_left = config.S3_MD_SIZE_LIMIT;
+    const sorted_keys = Object.keys(xattr).sort();
```
guymguym, line 135: I just noticed that you added this sort in our generic s3 utils.
Not sure this is a good thing to do for all of our namespaces.
I think the size limit is ok, but sorting sounds risky.
Not sure what to suggest, perhaps we can set a special Symbol('XATTR_SORT') on the xattr object in namespace_fs and here we will check for this symbol existence...

src/endpoint/s3/s3_utils.js
```diff
@@ -131,9 +133,27 @@ function get_request_xattr(req) {
 }
 
 function set_response_xattr(res, xattr) {
-    _.each(xattr, (val, key) => {
-        res.setHeader(X_AMZ_META + key, val);
-    });
+    let size_for_md_left = config.S3_MD_SIZE_LIMIT;
+    const keys = Object.keys(xattr);
+    // eslint-disable-next-line no-prototype-builtins
+    if (xattr.hasOwnProperty(XATTR_SORT_SYMBOL)) {
```
guymguym, line 139: I think you can just use brackets to access it, and then there's no need to disable eslint.
In any case we should definitely add a comment on this code block to explain *why* we introduced this symbol and condition.
```suggestion
    // Using a special symbol on the xattr object to specify when sorting of keys is needed
    // which will make sure the returned headers are consistently being truncated.
    // The symbol is set by namespace_fs and could be used in other namespaces where
    // xattrs are more likely to exceed the size limit and therefore sorting is desired,
    // but for most namespaces where the size is also imposed on PUT, we can avoid it.
    if (xattr[XATTR_SORT_SYMBOL]) {
```

guymguym (approved):
Looks good. Added few minor comments (like renaming the functions to simpler names).

PR #6549, opened 2021-06-03
Add NSFS monitoring & refactor NS monitoring

liranmauda:
### Explain the changes
1. Adding BG worker to the endpoint and running the `namespace_monitor` on it
2. Refractor `namespace_monitor`
3. Add the ability to decide if we should monitor a namespace or not (passing a `should_monitor` function to the `namespace_monitor` constructor.
4. Writing and deleting a small file into the `fs_root_path`
5. Reporting on a monitor fail and on an I/O fail
6. Adding namespace_resource_id to the namespaceFS constructor, and will not report if it is undefined.

### DIFF
--- a/src/core/nsfs.js
+++ b/src/core/nsfs.js
@@ -104,7 +104,11 @@ async function main(argv = minimist(process.argv.slice(2))) {
         object_sdk._get_bucket_namespace = async bucket_name => {
             const existing_ns = namespaces[bucket_name];
             if (existing_ns) return existing_ns;
-            const ns_fs = new NamespaceFS({ bucket_path: fs_root + '/' + bucket_name, bucket_id: '000000000000000000000000' });
+            const ns_fs = new NamespaceFS({
+                bucket_path: fs_root + '/' + bucket_name,
+                bucket_id: '000000000000000000000000',
+                namespace_resource_id: undefined
+            });
             namespaces[bucket_name] = ns_fs;
             return ns_fs;
         };
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -34,6 +34,8 @@ const server_rpc = require('../server/server_rpc');
 const auth_server = require('../server/common_services/auth_server');
 const system_store = require('../server/system_services/system_store');
 const prom_reporting = require('../server/analytic_services/prometheus_reporting');
+const background_scheduler = require('../util/background_scheduler').get_instance();
+const { NamespaceMonitor } = require('../server/bg_services/namespace_monitor');
 
 /**
  * @typedef {http.IncomingMessage & {
@@ -130,6 +132,14 @@ async function start_endpoint(options = {}) {
 
         await prom_reporting.start_server(config.EP_METRICS_SERVER_PORT);
 
+        if (internal_rpc_client) {
+            // Register a bg monitor on the endpoint
+            background_scheduler.register_bg_worker(new NamespaceMonitor({
+                name: 'namespace_fs_monitor',
+                client: internal_rpc_client,
+                should_monitor: nsr => Boolean(nsr.nsfs_config),
+            }));
+        }
         // Start a monitor to send periodic endpoint reports about endpoint usage.
         start_monitor(internal_rpc_client, endpoint_group_id);
 
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -123,13 +123,15 @@ class NamespaceFS {
      *  bucket_path: string;
      *  fs_backend?: string;
      *  bucket_id: string;
+     *  namespace_resource_id?: string;
      * }} params
      */
-    constructor({ bucket_path, fs_backend, bucket_id }) {
+    constructor({ bucket_path, fs_backend, bucket_id, namespace_resource_id }) {
         dbg.log0('NamespaceFS: buffers_pool', buffers_pool);
         this.bucket_path = path.resolve(bucket_path);
         this.fs_backend = fs_backend;
         this.bucket_id = bucket_id;
+        this.namespace_resource_id = namespace_resource_id;
     }
 
     set_cur_fs_account_config(object_sdk) {
@@ -168,6 +170,22 @@ class NamespaceFS {
             'other.fs_backend', other.fs_backend, 'this.fs_backend', this.fs_backend,
             'params.xattr_copy', params.xattr_copy);
         return is_server_side_copy;
+    }
+
+    run_update_issues_report(object_sdk, err) {
+        //We want to avoid the report when we have no error code.
+        if (!err.code) return;
+        //In standalone, we want to avoid the report.
+        if (!this.namespace_resource_id) return;
+        try {
+            object_sdk.rpc_client.pool.update_issues_report({
+                namespace_resource_id: this.namespace_resource_id,
+                error_code: err.code,
+                time: Date.now(),
+            });
+        } catch (e) {
+            console.log('update_issues_report on error:', e, 'ignoring.');
+        }
     }
 
     /////////////////
@@ -401,6 +419,7 @@ class NamespaceFS {
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
             return this._get_object_info(params.bucket, params.key, stat);
         } catch (err) {
+            this.run_update_issues_report(object_sdk, err);
             throw this._translate_object_error_codes(err);
         }
     }
@@ -530,7 +549,10 @@ class NamespaceFS {
         const upload_path = path.join(this.bucket_path, this.get_bucket_tmpdir(), 'uploads', upload_id);
         // dbg.log0('NamespaceFS.upload_object:', upload_path, '->', file_path);
         try {
-            await Promise.all([this._make_path_dirs(file_path, fs_account_config), this._make_path_dirs(upload_path, fs_account_config)]);
+            await Promise.all([
+                this._make_path_dirs(file_path, fs_account_config),
+                this._make_path_dirs(upload_path, fs_account_config)
+            ]);
             if (params.copy_source) {
                 const source_file_path = path.join(this.bucket_path, params.copy_source.key);
                 try {
@@ -550,6 +572,7 @@ class NamespaceFS {
             await nb_native().fs.rename(fs_account_config, upload_path, file_path);
             return { etag: this._get_etag(stat) };
         } catch (err) {
+            this.run_update_issues_report(object_sdk, err);
             throw this._translate_object_error_codes(err);
         }
     }
@@ -698,6 +721,7 @@ class NamespaceFS {
             const stat = await nb_native().fs.stat(fs_account_config, upload_path);
             return { etag: this._get_etag(stat) };
         } catch (err) {
+            this.run_update_issues_report(object_sdk, err);
             throw this._translate_object_error_codes(err);
         }
     }
--- a/src/sdk/object_sdk.js
+++ b/src/sdk/object_sdk.js
@@ -180,8 +180,13 @@ class ObjectSDK {
         try {
             // NAMESPACE_FS HACK
             if (process.env.NAMESPACE_FS) {
+                const namespace_resource_id = bucket.namespace.write_resource?.resource.id;
                 return {
-                    ns: new NamespaceFS({ bucket_path: process.env.NAMESPACE_FS + '/' + bucket.name, bucket_id: String(bucket._id) }),
+                    ns: new NamespaceFS({
+                        bucket_path: process.env.NAMESPACE_FS + '/' + bucket.name,
+                        bucket_id: String(bucket._id),
+                        namespace_resource_id,
+                    }),
                     bucket,
                     valid_until: time + (100 * 356 * 24 * 3600 * 1000), // 100 years
                 };
@@ -301,7 +306,8 @@ class ObjectSDK {
             return new NamespaceFS({
                 fs_backend: ns_info.fs_backend,
                 bucket_path: path.join(namespace_resource_config.resource.fs_root_path, namespace_resource_config.path || ''),
-                bucket_id: String(bucket_id)
+                bucket_id: String(bucket_id),
+                namespace_resource_id: ns_info.id,
             });
         }
         // TODO: Should convert to cp_code and target_bucket as folder inside
--- a/src/server/bg_services/namespace_monitor.js
+++ b/src/server/bg_services/namespace_monitor.js
@@ -7,17 +7,25 @@ const auth_server = require('../common_services/auth_server');
 const dbg = require('../../util/debug_module')(__filename);
 const system_utils = require('../utils/system_utils');
 const cloud_utils = require('../../util/cloud_utils');
+const nb_native = require('../../util/nb_native');
 const config = require('../../../config');
 const P = require('../../util/promise');
 const AWS = require('aws-sdk');
-const _ = require('lodash');
 
 class NamespaceMonitor {
 
-    constructor({ name, client}) {
+    /**
+     * @param {{
+     *   name: string;
+     *   client: nb.APIClient;
+     *   should_monitor: (nsr: nb.NamespaceResource) => boolean;
+     * }} params
+     */
+    constructor({ name, client, should_monitor }) {
         this.name = name;
         this.client = client;
         this.nsr_connections_obj = {};
+        this.should_monitor = should_monitor;
     }
 
     async run_batch() {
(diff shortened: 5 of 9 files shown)


src/sdk/namespace_fs.js
```diff
@@ -532,6 +547,17 @@ class NamespaceFS {
...
+                console.log('LMLM before upload_object report');
+                object_sdk.rpc_client.pool.update_issues_report({
+                    namespace_resource_id: this.namespace_resource_id,
+                    error_code: err.code,
+                    time: Date.now(),
+                });
+                console.log('LMLM after upload_object report');
+            } catch (e) {
+                console.log('update_issues_report on error:', e, 'ignoring.');
+            }
```
guymguym, line 560: 1. Refactor to a function - we don't want to repeat this boilerplate everytime.
2. Lets filter to report only on errors with `err.code`

src/server/bg_services/namespace_fs_monitor.js
```diff
@@ -0,0 +1,86 @@
...
+const path = require('path');
+const P = require('../../util/promise');
+const config = require('../../../config');
+const nb_native = require('../../util/nb_native');
+const system_utils = require('../utils/system_utils');
+const dbg = require('../../util/debug_module')(__filename);
+const auth_server = require('../common_services/auth_server');
+const system_store = require('../system_services/system_store').get_instance();
+
+class NamespaceFSMonitor {
```
guymguym, line 13: Why is this separated from NamespaceMonitor?

src/sdk/object_sdk.js
```diff
@@ -180,8 +180,13 @@ class ObjectSDK {
         try {
             // NAMESPACE_FS HACK
             if (process.env.NAMESPACE_FS) {
+                const namespace_resource_id = _.extend({}, bucket.namespace.write_resource).resource.id;
```
guymguym, line 183: I prefer more explicit and not to make an object just for checking existence.
We can start using the new `?.` operator for these cases:
```suggestion
                const namespace_resource_id = bucket.namespace.write_resource?.resource.id;
```

src/core/nsfs.js
```diff
@@ -104,7 +104,11 @@ async function main(argv = minimist(process.argv.slice(2))) {
         object_sdk._get_bucket_namespace = async bucket_name => {
             const existing_ns = namespaces[bucket_name];
             if (existing_ns) return existing_ns;
-            const ns_fs = new NamespaceFS({ bucket_path: fs_root + '/' + bucket_name, bucket_id: '000000000000000000000000' });
+            const ns_fs = new NamespaceFS({
+                bucket_path: fs_root + '/' + bucket_name,
+                bucket_id: '000000000000000000000000',
+                namespace_resource_id: '9999'
```
guymguym, line 110: can we pass undefined `namespace_resource_id` and then namespace_fs will avoid sending reports?
because in standalone more there is nowhere to send it.

guymguym (changes requested):

src/server/bg_services/namespace_fs_monitor.js
liranmauda: It felt more right, As this Monitor is only on the endpoint and not on the core (at list for now).
Another thing is that the `requires` are different and the test also.
The flow is different as we don't even need a connection.

guymguym: To me it makes more sense to have it as another case under the same namespace monitor class, rather than duplicating the entire class and add if(nsfs) and if(!nsfs) in each one respectfully. Sounds like you could extend the existing class with another function and that's it.

liranmauda: It is not just namespace vs nsfs, It is where it runs.
We are running it on the endpoint, and if we will run the Namespace_monitor on the endpoint it will run per namespace, even if it is not nsfs.
of course, we can also protect that, but I find it cleaner to divide into 2 files, one that runs in core and checks namespace and the other that's run in the endpoint and checks nsfs. 
WDYT?

src/server/bg_services/namespace_monitor.js
```diff
@@ -46,96 +49,140 @@ class NamespaceMonitor {
...
-        } else {
-            dbg.error('namespace_monitor: invalid endpoint type', endpoint_type);
+        } catch (err) {
+            dbg.log1('test_single_namespace_blob_resource_validity: got error:', err);
+            throw err;
+        }
+    }
+
+    async test_single_namespace_fs_resource_validity(nsr) {
+        const file = path.join(nsr.nsfs_config.fs_root_path, 'create_object_upload');
```
jeniawhite, line 178: Do we actually want to create an object with the name `create_object_upload` under the root path of the FS?
Doesn't seem like something reasonable to do.
Even if we want to create an object I would expect it to be with the `.` prefix so it won't be visible by the FS and definitely not with this name.
@guymguym

src/server/bg_services/namespace_monitor.js
```diff
@@ -46,96 +49,140 @@ class NamespaceMonitor {
...
-                    endpoint,
-                    access_key: access_key,
-                    secret_key: secret_key
-                });
-                conn = azure_storage.createBlobService(conn_string);
-                break;
-            }
-            default:
-                dbg.error('namespace_monitor: invalid endpoint type', nsr.endpoint_type);
+            return true;
```
jeniawhite, line 90: Maybe it is better to print the resource as well as we do below for other types?

jeniawhite (changes requested):

src/sdk/namespace_fs.js
```diff
@@ -162,6 +164,20 @@ class NamespaceFS {
             params.xattr_copy; // TODO, DO we need to hard link at TaggingDirective 'REPLACE'?
     }
 
+    run_update_issues_report(object_sdk, err) {
+        if (!err.code) return;
+        if (!this.namespace_resource_id) return;
```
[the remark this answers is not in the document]
guymguym: @jeniawhite - This is for standalone nsfs where we don't have a resource id, and do not even have the pool server to send the update_issues_report api call. So when namespace_resource_id is undefined, we skip reporting.

@liranmauda - You should definitely add comments to explain this in the constructor, and also here.

src/server/bg_services/namespace_monitor.js
```diff
@@ -46,96 +49,140 @@ class NamespaceMonitor {
...
+        }, {
+            auth_token: auth_server.make_auth_token({
+                system_id: system_store.data.systems[0]._id,
+                account_id: system_store.data.systems[0].owner._id,
+                role: 'admin'
+            })
+        });
+    }
+
+    init_nsr_s3_connection_to_target(nsr) {
```
guymguym, line 114: I would just embed this function content directly inside `test_s3_resource()`.

src/server/bg_services/namespace_monitor.js
```diff
@@ -46,96 +49,140 @@ class NamespaceMonitor {
...
+                secretAccessKey: secret_key.unwrap()
+            },
+            s3ForcePathStyle: true,
+            sslEnabled: false
+        });
+
+        this.nsr_connections_obj[nsr._id] = conn;
+    }
+
+    init_nsr_blob_connection_to_target(nsr) {
```
guymguym, line 129: I would just embed this function content directly inside `test_blob_resource()`.

src/server/bg_services/namespace_monitor.js
```diff
@@ -46,96 +49,140 @@ class NamespaceMonitor {
...
         }
     }
 
-    async test_single_namespace_resource_validity(nsr_info) {
-        if (nsr_info.nsfs_config) {
-            dbg.log1('namespace_monitor: namespace resource of type FS, skipping validity test...');
-            return;
-        }
-        const { endpoint_type, target_bucket} = nsr_info.connection;
+    async test_single_namespace_s3_resource_validity(nsr) {
```
guymguym, line 143: Rename - I would drop extra words that don't add meaning:
```suggestion
    async test_s3_resource(nsr) {
```

src/server/bg_services/namespace_monitor.js
```diff
@@ -46,96 +49,140 @@ class NamespaceMonitor {
...
             }).promise();
-        } else if (endpoint_type === 'AZURE') {
+        } catch (err) {
+            dbg.log1('test_single_namespace_s3_resource_validity: got error:', err);
+            throw err;
+        }
+
+    }
+
+    async test_single_namespace_blob_resource_validity(nsr) {
```
guymguym, line 160: Rename - I would drop extra words that don't add meaning:
```suggestion
    async test_blob_resource(nsr) {
```

src/server/bg_services/namespace_monitor.js
```diff
@@ -46,96 +49,140 @@ class NamespaceMonitor {
...
             );
-        } else {
-            dbg.error('namespace_monitor: invalid endpoint type', endpoint_type);
+        } catch (err) {
+            dbg.log1('test_single_namespace_blob_resource_validity: got error:', err);
+            throw err;
+        }
+    }
+
+    async test_single_namespace_fs_resource_validity(nsr) {
```
guymguym, line 177: Rename - I would drop extra words that don't add meaning:
```suggestion
    async test_nsfs_resource(nsr) {
```

src/server/bg_services/namespace_monitor.js
guymguym: Right, but I don't know why are we writing in the first place - why not just stat the fs_root_path?
What else is missing? detecting no-space? maybe we should just use statfs?
https://man7.org/linux/man-pages/man2/statfs.2.html

src/server/bg_workers.js
```diff
@@ -92,6 +92,13 @@ function remove_master_workers() {
 
 }
 
+function should_monitor(nsr) {
```
guymguym, line 95: unneeded as top level function, this is a small arrow func

src/endpoint/endpoint.js
```diff
@@ -138,6 +147,13 @@ async function start_endpoint(options = {}) {
     }
 }
 
+function should_monitor(nsr) {
```
guymguym, line 150: unneeded as top level function, this is a small arrow func

src/server/bg_services/namespace_monitor.js
```diff
@@ -7,17 +7,20 @@ const auth_server = require('../common_services/auth_server');
...
 const config = require('../../../config');
 const P = require('../../util/promise');
 const AWS = require('aws-sdk');
+const path = require('path');
 const _ = require('lodash');
 
 class NamespaceMonitor {
 
-    constructor({ name, client}) {
+    constructor({ name, client, should_monitor }) {
```
guymguym, line 19: Use jsdoc!
```suggestion
    /**
     * @param {{
     *   name: string;
     *   client: nb.APIClient;
     *   should_monitor: (nsr: nb.NamespaceResource) => boolean;
     * }} params
     */
    constructor({ name, client, should_monitor }) {
```

src/sdk/namespace_fs.js
```diff
@@ -121,13 +121,15 @@ class NamespaceFS {
      *  bucket_path: string;
      *  fs_backend?: string;
      *  bucket_id: string;
+     *  namespace_resource_id: string;
```
guymguym, line 124: Since namespace_resource_id is optional you want the jsdoc to represent that -
```suggestion
     *  namespace_resource_id?: string;
```

src/sdk/namespace_fs.js
```diff
@@ -516,8 +533,8 @@ class NamespaceFS {
         const upload_id = uuidv4();
         const upload_path = path.join(this.bucket_path, this.get_bucket_tmpdir(), 'uploads', upload_id);
         // console.log('NamespaceFS.upload_object:', upload_path, '->', file_path);
-        await Promise.all([this._make_path_dirs(file_path, fs_account_config), this._make_path_dirs(upload_path, fs_account_config)]);
         try {
+            await Promise.all([this._make_path_dirs(file_path, fs_account_config), this._make_path_dirs(upload_path, fs_account_config)]);
```
guymguym, line 537: Break this line
```suggestion
            await Promise.all([
                this._make_path_dirs(file_path, fs_account_config), 
                this._make_path_dirs(upload_path, fs_account_config)
            ]);
```

guymguym (changes requested):

src/sdk/namespace_fs.js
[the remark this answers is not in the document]
liranmauda: This is per @guymguym request https://github.com/noobaa/noobaa-core/pull/6549#discussion_r646101060
In `nsfs.js` which is standalone, we want to avoid the report.

src/server/bg_services/namespace_monitor.js
liranmauda: We thought of having a hidden directory for this, but as we create and delete the file we agreed that we will not create a directory but write the file into the `fs_root_path`.
With that said we can defiantly create a hidden file (and change the name).
@guymguym @nimrod-becker @jeniawhite WDYT?

src/endpoint/endpoint.js
```diff
@@ -354,6 +370,19 @@ function setup_http_server(server) {
     // });
 }
 
+function register_bg_endpoint_worker(worker, run_batch_function) {
```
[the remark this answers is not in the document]
liranmauda: True, but as loading it requires not to `register_common_services` twice, not `register_bg_services`, and maybe remove the `md_server.js` completely, We should not do it now.

src/server/bg_services/namespace_monitor.js
liranmauda: `stat` is not working properly.
for example, when you don't have the directory in your file system (it was deleted directly), you will still have this directory on the endpoint pod. in this case, `stat` will return an answer instead of an error that we expect.

src/sdk/namespace_fs.js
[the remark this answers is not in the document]
guymguym: Add a comment to explain that in the code!

src/server/bg_services/namespace_monitor.js
```diff
@@ -46,96 +55,120 @@ class NamespaceMonitor {
...
-                conn = azure_storage.createBlobService(conn_string);
-                break;
-            }
-            default:
-                dbg.error('namespace_monitor: invalid endpoint type', nsr.endpoint_type);
-        }
-        if (conn) {
+        if (this.nsr_connections_obj[nsr._id]) {
+            conn = this.nsr_connections_obj[nsr._id];
+        } else {
```
guymguym, line 104:
```suggestion
        let conn = this.nsr_connections_obj[nsr._id];
        if (!conn) {
```

src/server/bg_services/namespace_monitor.js
guymguym: But why do you think that writing/deleting/whatever will fail in case of an unmounted dir?
I don't think we are really trying to verify that the dir is mounted.
That's for kubernetes to do.

guymguym: I would still avoid writing and use stat/statfs instead.

guymguym (changes requested):
Some more comments. The biggest thing left from my point of view is that writing to the FS is not really needed and we better avoid it in the monitoring flow IMHO.

src/server/bg_services/namespace_monitor.js
liranmauda: > But why do you think that writing/deleting/whatever will fail in case of an unmounted dir?

Per my tests, I am deleting the fs_root_path in the minikube, and write fails with `ENOENT`
`stat` did not fail. 

So at list, in the case I checked `stat` is not monitoring and `write` does.

liranmauda: > I would still avoid writing and use stat/statfs instead.

What about `unkink` without `write`?  WDYT?

guymguym: OK can you give details on your test - what type of PVC did you use - hostpath? what did you delete in minikube exactly? are you saying that stat() succeeded on a missing dir??

Reminding of the problems we want to identify -

Operator:
- PVC not ready / mount failure - Update nsstore status to rejected. If possible, remove  from endpoints to keep serving other resources.

Filesystem:
- No/low space - ENOSPC / statfs
- No permissions - EPERM / EACCESS
- Not exists - ENOENT / EBADF? / ENOTFOUND?

So how are we doing in identifying all those?

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -35,8 +35,8 @@ mocha.describe('namespace_fs', function() {
     const ns_src_bucket_path = `./${src_bkt}`;
     const ns_tmp_bucket_path = `${tmp_fs_path}/${src_bkt}`;
 
-    const ns_src = new NamespaceFS({ bucket_path: ns_src_bucket_path, bucket_id: '1'});
-    const ns_tmp = new NamespaceFS({ bucket_path: ns_tmp_bucket_path, bucket_id: '2'});
+    const ns_src = new NamespaceFS({ bucket_path: ns_src_bucket_path, bucket_id: '1', namespace_resource_id: '1' });
```
guymguym, line 38: why are we setting namespace_resource_id here? 
does it work to report issues on this fake id in the test?

src/endpoint/endpoint.js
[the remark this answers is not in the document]
guymguym: I suggest we move the register_bg_worker function from -
https://github.com/noobaa/noobaa-core/blob/3f34c1a7389e68f1f20967e046128545a28995ca/src/server/bg_workers.js#L76-L86
into the utility module in -
https://github.com/noobaa/noobaa-core/blob/master/src/util/background_scheduler.js
And then share it from both.
Can work?

src/test/unit_tests/test_namespace_fs.js
liranmauda: true, I will change it explicitly to `namespace_resource_id: undefined`

src/server/bg_workers.js
```diff
@@ -72,18 +71,7 @@ function register_rpc() {
...
-    if (run_batch_function) {
-        worker.run_batch = run_batch_function;
-    }
-    if (!worker.name || !_.isFunction(worker.run_batch)) {
-        console.error('Name and run function must be supplied for registering bg worker', worker.name);
-        throw new Error('Name and run function must be supplied for registering bg worker ' + worker.name);
-    }
-    background_scheduler.run_background_worker(worker);
-}
+const register_bg_worker = background_scheduler.register_bg_worker;
```
guymguym, line 74: this will not work. this function is not bound to its instance so calling it will fail to resolve `this`.
You could replace with -
```suggestion
const register_bg_worker = 
    (worker, run_batch_function) => 
    background_scheduler.register_bg_worker(worker, run_batch_function)
```

guymguym (approved):

src/server/bg_workers.js
```diff
@@ -72,18 +71,9 @@ function register_rpc() {
...
-    if (run_batch_function) {
-        worker.run_batch = run_batch_function;
-    }
-    if (!worker.name || !_.isFunction(worker.run_batch)) {
-        console.error('Name and run function must be supplied for registering bg worker', worker.name);
-        throw new Error('Name and run function must be supplied for registering bg worker ' + worker.name);
-    }
-    background_scheduler.run_background_worker(worker);
-}
+const register_bg_worker =
```
jeniawhite, line 74: I don't understand the point of this transformation.
You basically create a function to point to a function without doing any manipulation.
Why not just call the function directly?

liranmauda: https://github.com/noobaa/noobaa-core/pull/6549#discussion_r649153223

liranmauda: + We want to avoid refractoring all the 21 places from `register_bg_worker` to `background_scheduler.register_bg_worker`

src/server/bg_services/namespace_monitor.js
```diff
@@ -46,96 +54,116 @@ class NamespaceMonitor {
...
+            if (!this.should_monitor(nsr)) return;
+
+            const endpoint_type = nsr.nsfs_config ? 'NSFS' : nsr.connection;
+
             try {
-                if (!this.nsr_connections_obj[nsr._id]) {
-                    this.init_nsr_connection_to_target(nsr);
+                if (endpoint_type === 'NSFS') {
+                    await this.test_nsfs_resource(nsr);
+                } else if (['AWS', 'S3_COMPATIBLE', 'IBM_COS'].includes(endpoint_type)) {
```
jeniawhite, line 64: In the case of non-NSFS, the connection holds the type inside the endpoint_type variable.
It is an object, not a type.
Just take the endpoint_type from the connection above when you assign.

jeniawhite (changes requested):

jeniawhite (approved):

PR #6556, opened 2021-06-07
Added nsfs_only property and the needed validation

### LINKED ISSUE #6495: Differentiate between different types of accounts - mainly regular and NF FS only
Description of what the problem is.
Currently "NS FS" accounts would also have access to regular buckets (data or NS). This can cause more overhead for managing allowed buckets per each account, if we want to create a flow where NS FS dedicated accounts only have access to NS FS-backed buckets. 

Possibly add an enum to indicate the type of account

Description of alternative solutions or features you've considered.

**Additional context**

romayalon:

### Explain the changes
1. Added allowed_non_fs_buckets property to common_api (account_schema and account_api).
2. Added validation in object_sdk.authorize_request (for ops in which bucket is provided/op is not create bucket) - only non nsfs account / nsfs account in which allowed_non_fs_buckets = true, are allowed for do the request.
3. Added validation in bucketspace_nb.list_buckets - only non nsfs account / nsfs account in which allowed_non_fs_buckets = true, are allowed to list non nsfs buckets (won't fail just skip the non nsfs bucket).
4. Added validation in bucket_server.create_bucket - only non nsfs account / nsfs account in which allowed_non_fs_buckets = true, are allowed for creation of non nsfs buckets.
5. Added unit tests.

1. Fixed #6495

### Testing Instructions:
1.

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -763,7 +763,7 @@ module.exports = {
 
         bucket_cache_config: {
             type: 'object',
-            required: [ ],
+            required: [],
             properties: {
                 ttl_ms: {
                     $ref: '#/definitions/bucket_cache_ttl'
@@ -811,11 +811,12 @@ module.exports = {
         },
         nsfs_account_config: {
             type: 'object',
-            required: ['uid', 'gid', 'new_buckets_path'],
+            required: ['uid', 'gid', 'new_buckets_path', 'nsfs_only'],
             properties: {
                 uid: { type: 'number' },
                 gid: { type: 'number' },
                 new_buckets_path: { type: 'string' },
+                nsfs_only: { type: 'boolean' }
             }
         },
     }
--- a/src/sdk/object_sdk.js
+++ b/src/sdk/object_sdk.js
@@ -139,8 +139,8 @@ class ObjectSDK {
         // check for a specific bucket
         if (bucket && req.op_name !== 'put_bucket') {
             // ANONYMOUS: cannot work without bucket, cannot work on namespace bucket (?)
+            const ns = await this.read_bucket_sdk_namespace_info(bucket);
             if (!token) {
-                const ns = await this.read_bucket_sdk_namespace_info(bucket);
                 // TODO: Anonymous access to namespace buckets not supported
                 if (ns) {
                     throw new RpcError('UNAUTHORIZED', `Anonymous access to namespace buckets not supported`);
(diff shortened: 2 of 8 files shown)


src/sdk/object_sdk.js
```diff
@@ -149,6 +149,13 @@ class ObjectSDK {
                     return;
                 }
             }
+
+            dbg.log0('authorize_request_account: allowed_non_fs_buckets: ', ns && ns.write_resource.resource);
+            const is_nsfs_bucket = ns && ns.write_resource.resource.fs_root_path;
```
guymguym, line 154: is this how we identify nsfs buckets?

guymguym: in any case I would move this into the validate_allowed_nsfs_account function to encapsulate better

src/sdk/object_sdk.js
```diff
@@ -158,6 +165,12 @@ class ObjectSDK {
         }
     }
 
+    // validates requests for non nsfs buckets from nsfs accounts which are allowed_non_fs_buckets 
+    validate_allowed_nsfs_account(account) {
+        if (!account) return true;
```
guymguym, line 170: when is account undefined? and why is that allowing access?

src/api/common_api.js
```diff
@@ -811,11 +811,12 @@ module.exports = {
         },
         nsfs_account_config: {
             type: 'object',
-            required: ['uid', 'gid', 'new_buckets_path'],
+            required: ['uid', 'gid', 'new_buckets_path', 'allowed_non_fs_buckets'],
             properties: {
                 uid: { type: 'number' },
                 gid: { type: 'number' },
                 new_buckets_path: { type: 'string' },
+                allowed_non_fs_buckets: { type: 'boolean' }
```
guymguym, line 819: How about we rename (and reverse meaning) to `nsfs_only`?

src/sdk/object_sdk.js
romayalon: yes, I moved it to a function so it'll be clearer.
I moved it inside the function.

src/sdk/object_sdk.js
romayalon: I added it since we are accessing properties of account, and I changes the logic, now when no account it throws an UNAUTHORIZED error.

src/api/common_api.js
romayalon: renamed, make more sense, thank you.

src/api/common_api.js
```diff
@@ -811,11 +811,12 @@ module.exports = {
         },
         nsfs_account_config: {
```
jeniawhite, line 812: Do we want to also let nsfs_only be configured via update_account_s3_access and delete_account_by_property?
If so then that RPC is not using this type.
Please make sure to go across all subtypes of that and verify that we do not want to add this capability to them as well.

jeniawhite (approved):

src/api/common_api.js
romayalon: added nsfs_only to update_account_s3_access api but not to delete_account_by_property

PR #6561, opened 2021-06-08
Fix update_bucket

### LINKED ISSUE #6185: NS-CACHE: Edit placement policy resets the bucket.namespace.caching property
### Environment info
- NooBaa Version: **master**
- Platform: **minikube**

### Actual behavior
1. Edit placement policy for cache bucket resets the bucket.namespace.caching property to undefined

### Expected behavior
1. Edit placement policy should not reset the caching policy.

### Steps to reproduce
1. Create namespace cache bucket and check namespace.caching:
```
> db.buckets.findOne({name:"hub-local-cache2"}).namespace.caching
{ "ttl_ms" : 10000 }
```
2. Edit placement policy from the UI to change from one namespace resource to another.
3. Check namespace.caching object is reset to undefined:
```
> db.buckets.findOne({name:"hub-local-cache2"}).namespace.caching === undefined
true
```

BarakSason:
Prior to this change, if a namespace caching policy was set on a bucket, performing a update_bucket operation would cause the caching value to be overridden to "undefined".
This change modifies the  update_bucket operation so:
1 - If the update operation contains a caching value, the bucket's caching will be set to this value
2 - If the  update operation dos not contain a caching value, but a caching policy was already set on the bucket, the caching policy will not be overridden to "undefined".

Fixed: #6185

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -506,7 +506,7 @@ async function read_bucket_sdk_info(req) {
                 path: bucket.namespace.write_resource.path,
             },
             read_resources: _.map(bucket.namespace.read_resources, rs =>
-                ({ resource: pool_server.get_namespace_resource_extended_info(rs.resource), path: rs.path})
+                ({ resource: pool_server.get_namespace_resource_extended_info(rs.resource), path: rs.path })
             ),
             caching: bucket.namespace.caching,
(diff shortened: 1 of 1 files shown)


BarakSason:
The fix consists of lines 628-636, other changes are due to "Beautify" formatting.

liranmauda (approved):

Issue #6427, opened 2021-03-25
NSFS add Etag-MD5 compatibility mode to calc MD5 on PUT-object


guymguym:
### Environment info
- NooBaa Version: **5.9**
- Platform: **NA**

### Actual behavior
1. Currently all etags in nsfs are generated from inode number and mtime.
2. By using a dash `-` we make sure that the etag is not considered as md5 by the client SDK.
3. Some applications have  relied on Etags being MD5 in the PUT response and might not be able to run if PUT-object will return a non MD5 Etag.

### Expected behavior
1. We should introduce a new nsfs_account_config flag that will set that account to use etag_mode=md5 to support those applications by spending the additional CPU to calculate MD5 while streaming the data and store that in the file xattr.

jeniawhite:
Seems relevant
https://github.com/noobaa/noobaa-core/issues/6228

nimrod-becker:
Fixed with #6651

PR #6567, opened 2021-06-10
Add Etag-MD5 compatibility mode to calc MD5 on PUT-object

liranmauda:
### Explain the changes
1. Add a `get_stream_md5` function in `stream_utils.js` that gets the md5 out of a stream
2. Calling `get_stream_md5` where needed on the put object (upload/multipart upload/copy)

***Still Draft***
Currently as a temp step, hanging it on the pool 
and also on the contractor setting this to true
We probably do not need to do it in the contractor and check the field wherever we decide to place it.

### Testing Instructions:
***Will be set after Draft***

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -6,6 +6,7 @@ const fs = require('fs');
 const path = require('path');
 const util = require('util');
 const mime = require('mime');
+const crypto = require('crypto');
 const { v4: uuidv4 } = require('uuid');
 
 const config = require('../../config');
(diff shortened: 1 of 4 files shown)


nimrod-becker:
https://github.com/noobaa/noobaa-core/issues/6427

guymguym (changes requested):
We don't want to calculate md5 in js.
We should move it to our FSWorker threads.
Lets discuss to design.

src/sdk/namespace_fs.js
```diff
@@ -162,6 +166,14 @@ class NamespaceFS {
             params.xattr_copy; // TODO, DO we need to hard link at TaggingDirective 'REPLACE'?
     }
 
+    // get a stream and returns it's md5_b64
+    get_stream_md5_b64(stream) {
+        const md5_stream = stream_utils.get_stream_md5();
+        stream.pipe(md5_stream);
+        const md5_b64 = md5_stream.md5.digest('base64');
```
guymguym, line 173: This cannot work - the digest is not yet known at the time of setting up the pipe.
Only once the entire stream pipes through, then it will have the digest ready.
In any case, as we discussed, we don't want to run the hash in main thread.

liranmauda:
Closing In favor of a different approach. 
We will make a new PR with md5 calc on a different thread.

PR #6576, opened 2021-06-14
Added check_access() call to list objects

romayalon:

### Explain the changes
1. Added a call to check_access in list_objects of namespace_fs
2. Added initial tests - still in progress.

Comment from Romy 25 Feb 25 -
We added the check access because the readdir in the directory cache load() will fail the whole list objects command, therefore we checked if the user has access to the directory.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -7,12 +7,12 @@ const path = require('path');
 const util = require('util');
 const mime = require('mime');
 const { v4: uuidv4 } = require('uuid');
+const dbg = require('../util/debug_module')(__filename);
 
 const config = require('../../config');
 const s3_utils = require('../endpoint/s3/s3_utils');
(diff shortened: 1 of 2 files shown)


jeniawhite (approved):

src/sdk/namespace_fs.js
```diff
@@ -1104,6 +1104,29 @@ class NamespaceFS {
             throw this._translate_object_error_codes(err);
         }
     }
+
+    async check_access(fs_account_config, dir_path) {
+        try {
+            dbg.log0('check_access: dir_path', dir_path, 'fs_account_config', fs_account_config);
+            await nb_native().fs.checkAccess(fs_account_config, dir_path);
```
jeniawhite, line 1111: A bit weird that we are calling open on directories and not opendir, but it seems to work

jeniawhite (approved):

PR #6587, opened 2021-06-18
quick fix for create triggers not working with namespaced bucket #6575 [draft]

### LINKED ISSUE #6575: Triggers not working anymore with namespaced buckets with s3 backend
### Environment info
- NooBaa Core Version: master+d9dd10ae7d33f4a21f4acdce183e4603630c3bb4
- Platform: OKD 4.6

### Actual behavior
1. Trigger not executed on namespaced bucket

### Expected behavior
1. Trigger gets executed

### Steps to reproduce
1. Create namespace resource with s3 compatible v2 backend
2. Create namespace bucket with the same backend bucket as read and write resource

I think this addition
https://github.com/noobaa/noobaa-core/blob/4a3c59daa150a208d655cb977222c59449d44278/src/sdk/object_sdk.js#L203-L209
is the problem, as now no longer a NamespaceMerge object is created but instead a NamespaceS3 object and the NamespaceS3 class does not implement trigger functionality

lallinger-tech:
### Explain the changes
1. Quick implementation for missing create triggers on namespaced buckets

1. #6575 
2. Only fixes ObjectCreated triggers on simple upload and multipart upload

### Testing Instructions:
1. Create function
2. Create bucket
3. Add ObjectCreated trigger with function to trigger
4. Upload object to bucket, which fires trigger
5. Verify trigger was called

### DIFF
--- a/src/sdk/namespace_s3.js
+++ b/src/sdk/namespace_s3.js
@@ -18,13 +18,14 @@ const config = require('../../config');
 class NamespaceS3 {
 
 
-    constructor({ namespace_resource_id, rpc_client, s3_params }) {
+    constructor({ namespace_resource_id, rpc_client, s3_params , active_triggers}) {
         this.namespace_resource_id = namespace_resource_id;
         this.access_key = s3_params.accessKeyId;
         this.endpoint = s3_params.endpoint;
         this.s3 = new AWS.S3(s3_params);
         this.bucket = String(this.s3.config.params.Bucket);
         this.rpc_client = rpc_client;
+        this.active_triggers = active_triggers;
     }
 
     get_write_resource() {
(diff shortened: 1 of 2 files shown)


src/sdk/object_sdk.js
```diff
@@ -258,7 +258,7 @@ class ObjectSDK {
         });
     }
 
-    _setup_single_namespace(namespace_resource_config, bucket_id) {
+    _setup_single_namespace(namespace_resource_config, bucket_id, bucket) {
```
liranmauda, line 261: We can pass this function only the bucket and get `bucket.namespace.write_resource` and `bucket._id` from it

src/sdk/namespace_s3.js
```diff
@@ -18,13 +18,14 @@ const config = require('../../config');
 class NamespaceS3 {
 
 
-    constructor({ namespace_resource_id, rpc_client, s3_params }) {
+    constructor({ namespace_resource_id, rpc_client, s3_params , active_triggers}) {
```
liranmauda, line 21: Do we need `active_triggers` in the contractor? 
Can we get it from the bucket? 
We need to change all the `new NamespaceS3` with this change if we leave it here.

src/sdk/namespace_s3.js
```diff
@@ -298,12 +299,37 @@ class NamespaceS3 {
...
+            object_sdk.dispatch_triggers({ active_triggers: this.active_triggers, operation, obj, bucket: params.bucket });
+        }
+
         dbg.log0('NamespaceS3.upload_object:', this.bucket, inspect(params), 'res', inspect(res));
         const etag = s3_utils.parse_etag(res.ETag);
         const last_modified_time = s3_utils.get_http_response_date(res);
         return { etag, version_id: res.VersionId, last_modified_time };
     }
 
+    test_should_run_triggers({ active_triggers, operation, obj }) {
```
liranmauda, line 321: Why do we need this function here? 
Can we trigger it from `object_sdk`?

src/sdk/namespace_s3.js
```diff
@@ -298,12 +299,37 @@ class NamespaceS3 {
                 throw err;
             }
         }
+        const operation = 'ObjectCreated';
+        const load_for_trigger = this.test_should_run_triggers({ active_triggers: this.active_triggers, operation });
```
liranmauda, line 303: I think we can call it from `object_sdk`

liranmauda (changes requested):

liranmauda:
Hi @lallinger-arbeit 
Thanks for this PR, I have written some comments.
Once they are done, you need to sign the commit in order to resolve the DCO.
regarding the fail tests, they are lint errors, but the changes I asked for should resolve them. If there will be another lint error we will resolve them too.

src/sdk/namespace_s3.js
lallinger-arbeit: I just copied this as it is in the NamespaceMerge class, there it is in the constructor too

lallinger-arbeit:
> Hi @lallinger-arbeit
> Thanks for this PR, I have written some comments.
> Once they are done, you need to sign the commit in order to resolve the DCO.

Hi @liranmauda 
as discussed in the community meeting this is just my quick and dirty implementation, and should only be seen as draft, which does not need to get merged. I think it is best you implement the neccessary changes for all the triggers, so also ObjectRead and ObjectRemoved, as you have way deeper insight in the code. And also it is not a problem for us if it takes a few weeks, as this implementation is currently working for our use case

liranmauda:
> Hi @liranmauda
> as discussed in the community meeting this is just my quick and dirty implementation, and should only be seen as draft, which does not need to get merged. I think it is best you implement the neccessary changes for all the triggers, so also ObjectRead and ObjectRemoved, as you have way deeper insight in the code. And also it is not a problem for us if it takes a few weeks, as this implementation is currently working for our use case

This is a great PR, And we can marge it after we address the changes above. 
We will probably do the same + adding the missing parts.

PR #6588, opened 2021-06-20
Adding `lstat`, `fstat` and `statfs` to fs_napi

liranmauda:
### Explain the changes
- Adding `lstat`, `fstat` and `statfs` to fs_napi

### Gap
- In `statfs` we are not returning fsid in the results.
- In `statfs` we are not returning MacOs special properties.

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -20,6 +20,13 @@
 #include <sys/xattr.h>
 #include <vector>
 
+#ifdef __APPLE__
+    #include <sys/param.h>
+    #include <sys/mount.h>
+#else
+    #include <sys/statfs.h>
+#endif
+
 namespace noobaa
 {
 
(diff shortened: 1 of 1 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +214,112 @@ struct Stat : public FSWorker
     }
 };
 
+/**
+ * LStat is an fs op
+ */
+struct LStat : public FSWorker
```
jeniawhite, line 220: LStat and Stat are identical.
Except for the symbolic link which is transparent to us since we just call the POSIX and the FS handles that.
Maybe instead of copy/pasting the code and maintaining similar copies to code, it is better to provide an option to the native layer and call the functions respectively?
Maybe even having the fstat as well passing an FD as an option.
These three functions basically have the same code and the only difference is what POSIX we are calling.
You can check the first parameter if number (fd) or string (path) and that way you between fstat and regular stat, and the second parameter will be provided when there is a requirement to do lstat.

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +214,112 @@ struct Stat : public FSWorker
...
+    {
+        int r = statfs(_path.c_str(), &_statfs_res);
+        if (r) SetSyscallError();
+    }
+    virtual void OnOK()
+    {
+        DBG1("FS::Statfs::OnOK: " << DVAL(_path) << DVAL(_statfs_res.f_type) << DVAL(_statfs_res.f_bsize));
+        Napi::Env env = Env();
+        auto res = Napi::Object::New(env);
+        res["__fsword"] = Napi::Number::New(env, _statfs_res.f_type);
```
jeniawhite, line 302: I do not understand why you assign different values to the same properties multiple times.
2 assignments for __fsword
3 assignments for fsblkcnt
2 assignments for fsfilcnt
3 assignments for __fsword

jeniawhite (changes requested):

src/native/fs/fs_napi.cpp
guymguym: Good point. I agree that the similarity is high.
But I would prefer to see the code reuse inside and still allow different function names outside to match the well known linux syscalls.

src/native/fs/fs_napi.cpp
```diff
@@ -213,6 +214,112 @@ struct Stat : public FSWorker
...
+    {
+        int r = lstat(_path.c_str(), &_stat_res);
+        if (r) SetSyscallError();
+    }
+    virtual void OnOK()
+    {
+        DBG1("FS::LStat::OnOK: " << DVAL(_path) << DVAL(_stat_res.st_ino) << DVAL(_stat_res.st_size));
+        Napi::Env env = Env();
+        auto res = Napi::Object::New(env);
+        res["dev"] = Napi::Number::New(env, _stat_res.st_dev);
```
guymguym, line 241: How about creating a common function to do all this work to read from stat_res and populate res?

src/native/fs/fs_napi.cpp
```diff
@@ -186,41 +233,92 @@ struct Stat : public FSWorker
...
+        #ifdef __APPLE__   
+            //  res["otype"] =  Napi::Number::New(env, _statfs_res.f_otype);
+            //  res["oflags"] =  Napi::Number::New(env, _statfs_res.f_oflags);
+            //  res["iosize"] =  Napi::Number::New(env, _statfs_res.iosize);
+            res["owner"] =  Napi::Number::New(env, _statfs_res.f_owner);
+            //  res["reserved1"] =  Napi::Number::New(env, _statfs_res.f_reserved1);
+        #else
+            res["namelen"] = Napi::Number::New(env, _statfs_res.f_namelen);
+            res["frsize"] = Napi::Number::New(env, _statfs_res.f_frsize);
+        #endif
```
guymguym, line 321: Are we going to use any of those OS specific properties??
I hope not, and I would remove those to avoid confusion and errors between developer env and runtime env.

src/native/fs/fs_napi.cpp
```diff
@@ -186,41 +233,92 @@ struct Stat : public FSWorker
     }
     virtual void OnOK()
     {
-        DBG1("FS::Stat::OnOK: " << DVAL(_path) << DVAL(_stat_res.st_ino) << DVAL(_stat_res.st_size));
+        DBG1("FS::Stat::OnOK: " << DVAL(_path) << DVAL(_stat_res.st_ino) << DVAL(_stat_res.st_size));    
+        Napi::Env env = Env();
+        auto res = set_stat_res(env, _stat_res);
```
guymguym, line 238: The function name is misleading. 
If you want to `set_stat_res()`, then you should send in the object on which to set.
Currently the function is creating and populating the object, so a more suitable name `is make_stat_res()`
Personally I would prefer that the callers will create the object and pass it in to set_stat_res() and then you don't need to return it (no radical auto in function response type...)

guymguym (changes requested):
See two comments, in addition I think we should add a stat() function on FileWrap too, which is now very simple after your refactor.

src/native/fs/fs_napi.cpp
```diff
@@ -65,6 +72,69 @@ api(const Napi::CallbackInfo& info)
     return promise;
 }
 
+void
+set_stat_res(Napi::Object res, Napi::Env env, struct stat _stat_res)
```
guymguym, line 76: send struct by reference, and rename without leading `_`
```suggestion
set_stat_res(Napi::Object res, Napi::Env env, struct stat &stat_res)
```

src/native/fs/fs_napi.cpp
```diff
@@ -65,6 +72,69 @@ api(const Napi::CallbackInfo& info)
...
+    res["birthtimeMs"] = Napi::Number::New(env, birthtimeMs);
+    res["atime"] = Napi::Date::New(env, uint64_t(round(atimeMs)));
+    res["mtime"] = Napi::Date::New(env, uint64_t(round(mtimeMs)));
+    res["ctime"] = Napi::Date::New(env, uint64_t(round(ctimeMs)));
+    res["birthtime"] = Napi::Date::New(env, uint64_t(round(birthtimeMs)));
+
+}
+
+void
+set_statfs_res(Napi::Object res, Napi::Env env, struct statfs _statfs_res)
```
guymguym, line 114:
```suggestion
set_statfs_res(Napi::Object res, Napi::Env env, struct statfs &statfs_res)
```

guymguym (approved):

guymguym:
FYI @jeniawhite this will allow to change read_object_md() to use `file = open(); file.stat(); file.getxattr();`

jeniawhite (approved):

PR #6597, opened 2021-06-22
Bucket replication configuration 

romayalon:

### Explain the changes
1. Added bucket replication configuration to bucket API and schema.
2. Added api validation put_bucket_replication, added getter get_bucket_replication and added delete_bucket_replication.
3. Note 1: implementation of validate_replication:
I decided on some rules regarding the validity of bucket replication configuration:
a. Destination bucket of a rule must exist in NooBaa system.
b. Destination bucket of a rule must not be the same bucket on which the user defines the bucket replication config.
c. rule_ids of 2 different rules must not be equal.
d. if the destination bucket of rule 1 equals to destination bucket of rule 2  - prefix of rule 1 should not be a prefix of the prefix of rule 2 (and vice versa) 
4. Note 2: max number of different destination buckets is 100 and the max number of rules is 1000.
5. Added unit tests 
1. 

### Testing Instructions:
1.


nimrod-becker:
Shouldn't we add the new definition (replication_policy) to the bucket_info so it would be returned as part of the read_bucket and create bucket. Or, if we don't want to return it with the bucket info, add a getter for the rules?

src/server/system_services/bucket_server.js
```diff
@@ -247,6 +251,49 @@ function validate_pool_constraints({ mongo_pool, default_pool }) {
     }
 }
 
+function validate_replication(req) {
+    const replication_policy = req.rpc_params.replication_policy;
+    // num of rules in configuration must be in defined limits
+    if (replication_policy.length > config.BUCKET_REPLICATION_MAX_RULES ||
+        replication_policy.length < config.BUCKET_REPLICATION_MIN_RULES) throw new RpcError('INVALID_REPLICATION_POLICY');
```
nimrod-becker, line 258: < config.BUCKET_REPLICATION_MIN_RULES fails, so how do I remove rules altogether (getting the bucket to 0 rules)

nimrod-becker:
I think we should also add a design doc, for this part you only did the API so we can do only the API in the design doc. With each added PR we can add the new section in the design doc, WDYT?

src/test/unit_tests/test_bucket_replication.js
```diff
@@ -0,0 +1,106 @@
+/* Copyright (C) 2020 NooBaa */
```
nimrod-becker, line 1: You might want to add this new test to 
src/test/unit_tests/index.js

Unless you don't want it to run automatically at this point

src/server/system_services/bucket_server.js
romayalon: I decided to add delete_bucket_replication function for deletion of all rules. WDYT?

src/test/unit_tests/test_bucket_replication.js
romayalon: added, thanks.

docs/design/BucketReplication.md
```diff
@@ -0,0 +1,140 @@
...
+    and optionally defines a filter based on the object key prefix
+
+Structure of replication policy which was added to bucket document in bucket collection:
+
+```
+    replication_policy: {
+        type: 'array',
+        items: {
+            type: 'object',
+            required: ['destination_bucket', 'rule_id'],
```
dannyzaken, line 35: I'm not sure I understand the purpose of the `rule_id`

romayalon: it's for compatibility with S3 bucket replication

config.js
```diff
@@ -514,6 +514,18 @@ config.DB_TYPE = process.env.DB_TYPE || 'mongodb';
...
 
+
+////////////////////////////////
+//      BUCKET REPLICATOR     //
+////////////////////////////////
+
+config.BUCKET_REPLICATOR_DELAY = 15 * 60 * 1000;
+config.BUCKET_REPLICATOR_LIST_LIMIT = 1000;
+config.BUCKET_REPLICATION_MAX_RULES = 1000;
+config.BUCKET_REPLICATION_MIN_RULES = 1;
```
dannyzaken, line 525: why do we need MIN for replication rules? will we want to enforce more than one rule

src/server/system_services/bucket_server.js
```diff
@@ -251,6 +251,49 @@ function validate_pool_constraints({ mongo_pool, default_pool }) {
     }
 }
 
+function validate_replication(req, configured_buckets, min_required) {
+    const replication_policy = req.rpc_params.replication_policy;
+    // num of rules in configuration must be in defined limits
+    if (replication_policy.length > config.BUCKET_REPLICATION_MAX_RULES ||
+        (min_required && replication_policy.length < config.BUCKET_REPLICATION_MIN_RULES)) throw new RpcError('INVALID_REPLICATION_POLICY');
```
dannyzaken, line 258: we can just check that `replication_policy.length > 0`

src/server/system_services/bucket_server.js
nimrod-becker: So update for deletion of a single rule and delete replication for all?
It's weird that I can delete one by one besides the last one , WDYT? Maybe add a way to remove all rules with the same API of update ?

config.js
romayalon: same, for compatibility with S3 bucket replication
'A replication configuration must include at least one rule, and can contain a maximum of 1,000. '
https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketReplication.html

src/server/system_services/bucket_server.js
romayalon: if we don't want to have the minimum in config, yes

src/server/system_services/bucket_server.js
romayalon: I can remove the min for now and update will work as you want but then delete_bucket_replication is redundant and we won't be compatible with s3 bucket replication

nimrod-becker: If its for compatibility, lets keep it

src/server/system_services/schemas/replication_configuration_schema.js
```diff
@@ -0,0 +1,36 @@
...
+    type: 'object',
+    required: [
+        '_id',
+        'replication_policy',
+    ],
+    properties: {
+
+        // identifiers
+        _id: { objectid: true },
+        replication_policy: {
```
dannyzaken, line 15: maybe rename it to `rules`, so in your code you will have `replication_policy.rules` instead of `replication_policy.replications_policy`

config.js
dannyzaken: I think that we will never want to change it to something that is not 1, so adding a config for that just makes it a bit confusing and harder to read when looking at the code

src/server/system_services/bucket_server.js
```diff
@@ -251,6 +252,49 @@ function validate_pool_constraints({ mongo_pool, default_pool }) {
     }
 }
 
+function validate_replication(req, configured_buckets, min_required) {
```
dannyzaken, line 255: I'm not a fan of validation functions that do other stuff. I think it's better to split it into 2 functions, `validate_replication` and `noramalize_replication_policy` or something

src/server/system_services/bucket_server.js
```diff
@@ -1900,6 +1948,79 @@ function validate_non_nsfs_bucket_creation(req) {
     }
 }
 
+// updates replication policy of all bucket class buckets
+async function update_bucket_class_replication(req) {
```
dannyzaken, line 1952: I think you can refactor such that both `update_bucket_class_replication` and `put_bucket_replication` can use the same internal function to update a replication policy

romayalon: merged the logic, thanks

src/server/system_services/bucket_server.js
romayalon: splitted

src/server/system_services/schemas/replication_configuration_schema.js
romayalon: renamed

config.js
romayalon: removed from config

src/server/system_services/bucket_server.js
```diff
@@ -1900,6 +1905,112 @@ function validate_non_nsfs_bucket_creation(req) {
...
+function validate_replication(req, bucket_name, min_required) {
+    const replication_rules = req.rpc_params.replication_policy;
+    // num of rules in configuration must be in defined limits
+    if (replication_rules.length > config.BUCKET_REPLICATION_MAX_RULES ||
+        (min_required && replication_rules.length < 1)) throw new RpcError('INVALID_REPLICATION_POLICY');
+
+    let rule_ids = [];
+    let pref_by_dst_bucket = {};
+
+    _.map(replication_rules, rule => {
```
dannyzaken, line 1970: you don't need to use `_.map`. just use for-of 
```suggestion
    for (const rule of replication_rules) {
```

src/server/system_services/bucket_server.js
```diff
@@ -1900,6 +1905,112 @@ function validate_non_nsfs_bucket_creation(req) {
...
+                }
+            }
+        }
+    }
+}
+
+function normalize_replication(req) {
+    const replication_rules = req.rpc_params.replication_policy;
+
+    const validated_replication = _.map(replication_rules, rule => {
```
dannyzaken, line 2006: since `replication_rules` is an array I think it's more readable to use `replication_rules.map`
```suggestion
    const validated_replication = replication_rules.map(rule => {
```

dannyzaken (approved):
just a few minor comments. other than that LGTM

dannyzaken (approved):
LGTM

PR #6601, opened 2021-06-23
postgres performance improvements - Write path

dannyzaken:
This PR improves most of the queries in the write path. Read path is not handled in this PR

#### in postgres_client
* Added option to enable query logging using env `PG_ENABLE_QUERY_LOG`. this will print for each query
  the time it took and if enabled (with `PG_EXPLAIN_QUERIES`) the EXPLAIN result of the query
* fixes in encode_json and handle_ops_encoding
  * encode mongo.objectid as string in postgres_client encode_json - when mongo object id is passed as object to mongo-to-postgres lib, it will convert the query to access data->_id as a jsonb struct. this causes queries to not hit the indexes. when passed as string the lib generates the query using `->>` that better matches the index
  * handle_ops_encoding - handle multiple ops in one query
  * handle comparison operators (`$eq`, `$ne`, etc.)
* add `{disableContainmentQuery: true}` when calling `mongo-query-to-postgres-jsonb`
* fix sorts to use indexes
  * access `_id` as text
  * remove NULLS FIRST\LAST and use default as we do in the indexes

#### other changes
* added datachunk index by _id descending for iterate_all_chunks
* in md_store - pass `$exists` in some queries to match indexes
* in system_server - cache result of list_undeletable_buckets
  to avoid calling it on each read_system
* bumped version of `mongo-query-to-postgres-jsonb` to `0.2.14` 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -23,10 +23,20 @@ const fs = require('fs');
 const mongo_functions = require('./mongo_functions');
 const { RpcError } = require('../rpc');
 const SensitiveString = require('./sensitive_string');
-
+const time_utils = require('./time_utils');
 mongodb.Binary.prototype[util.inspect.custom] = function custom_inspect_binary() {
     return `<mongodb.Binary ${this.buffer.toString('base64')} >`;
 };
+
+
+const COMPARISON_OPS = [
+    '$eq',
+    '$ne',
+    '$gt',
+    '$gte',
+    '$lt',
+    '$lte',
+];
 
 
 // temporary solution for encode\decode
@@ -81,6 +91,10 @@ function encode_json(schema, val) {
     const ops = handle_ops_encoding(schema, val);
     if (ops) return ops;
 
+    if (schema.objectid === true && val instanceof mongodb.ObjectID) {
+        return val.toString();
+    }
+
     if (schema.binary === true) {
         // assuming val is of type Buffer. convert to base64 string
         return val.toString('base64');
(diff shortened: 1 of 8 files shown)


src/util/postgres_client.js
```diff
@@ -411,7 +448,16 @@ class PostgresTable {
...
+        this.db_indexes = _.concat([{
+                fields: {
+                    _id: 1,
+                },
+                options: {
+                    name: '_id_index',
+                    unique: true,
+                }
+            }],
+            db_indexes || []);
```
guymguym, line 460: I would replace concat with `[item, ...arr]`
```suggestion
        const id_index = {
            fields: {
                _id: 1,
            },
            options: {
                name: '_id_index',
                unique: true,
            }
        };
        this.db_indexes = [id_index,  ...(db_indexes || [])];
```

src/util/postgres_client.js
```diff
@@ -149,13 +153,46 @@ function handle_ops_encoding(schema, val) {
...
+        // dbg.log0(`postgres_client: ${tag}: ${q.text}`, util.inspect(q.values, { depth: 6 }));
+        const millistart = time_utils.millistamp();
+        const pool_count = `total=${pg_client.totalCount} waiting=${pg_client.waitingCount} idle=${pg_client.idleCount}`;
         const res = await pg_client.query(q);
-        dbg.log1(`postgres_client: ${tag}: got result:`, util.inspect(res, { depth: 6 }));
+        const millitook = time_utils.millitook(millistart);
+        try {
+            await log_query(pg_client, q, tag, millitook, pool_count, transaction_counter === 0);
+        } catch (err) {
+            console.error('error on log query', err);
```
guymguym, line 194: can log_query fail?

src/util/postgres_client.js
```diff
@@ -149,13 +153,46 @@ function handle_ops_encoding(schema, val) {
...
     try {
-        dbg.log1(`postgres_client: ${tag}: ${q.text}`, util.inspect(q.values, { depth: 6 }));
+        // dbg.log0(`postgres_client: ${tag}: ${q.text}`, util.inspect(q.values, { depth: 6 }));
+        const millistart = time_utils.millistamp();
+        const pool_count = `total=${pg_client.totalCount} waiting=${pg_client.waitingCount} idle=${pg_client.idleCount}`;
         const res = await pg_client.query(q);
-        dbg.log1(`postgres_client: ${tag}: got result:`, util.inspect(res, { depth: 6 }));
+        const millitook = time_utils.millitook(millistart);
+        try {
+            await log_query(pg_client, q, tag, millitook, pool_count, transaction_counter === 0);
```
guymguym, line 192: Can you add a comment to explain why `should_explain = transaction_counter === 0` ?
It's not clear to me.

src/util/postgres_client.js
```diff
@@ -149,13 +153,46 @@ function handle_ops_encoding(schema, val) {
 
 let query_counter = 0;
 
+async function log_query(pg_client, q, tag, millitook, pool_count, should_explain) {
+    const explain_q = { text: 'explain ' + q.text, values: q.values };
+    let explain_rows = ['explain:\tno explain'];
+    try {
+        if (should_explain && process.env.EXPLAIN_QUERIES === 'true') {
```
guymguym, line 160: I would rename the env to PG_EXPLAIN_QUERIES or DB_EXPLAIN_QUERIES

src/util/postgres_client.js
```diff
@@ -149,13 +153,46 @@ function handle_ops_encoding(schema, val) {
...
+    let stack = (new Error()).stack.split('\n').slice(3).join(',');
+    const msgs = [];
+    msgs.push(`stack: ${stack}`);
+    msgs.push(`tag: ${tag}`);
+    msgs.push(`query: ${q.text}`);
+    msgs.push(`took: ${millitook}`);
+    msgs.push(`pool: ${pool_count}`);
+    msgs.push(`explain:`);
+    let log_message = msgs.concat(explain_rows).join('\nQUERY_LOG:\t');
+    console.log(`\nQUERY_LOG:\t${log_message}`);
```
guymguym, line 178: There seems to be q lot of objects being created here just for logging.
I would try to keep this more straightforward code that prints the string.
Also I think it would be best to keep the format as a json object and not a multiline log print.

guymguym (approved):
small comments, otherwise looks good

src/util/postgres_client.js
dannyzaken: it can fail when running explain

src/util/postgres_client.js
dannyzaken: sure. 
In general the log_query code is a quick and dirty solution. I plan to refactor it if and before merging. the if is there since failures in the explain for transaction caused the transaction to no longer be valid, so I decided for now to just avoid it

dannyzaken:
try to ignore the first commit. It will not remain as is

src/util/postgres_client.js
guymguym: got it. I think it's worth keeping instrumentation inside the code behind env var, so we can use for next performance improves. or you have a better way to collect those from `pg_stat_statements`?
perhaps keep such scripts in git too?

src/util/postgres_client.js
guymguym: ah I thought I saw you had try-catch inside

PR #6605, opened 2021-06-24
Added Quota solution design for noobaa-core component

igorpick:
### Explain the changes
doc for quota feature solution. For now the doc describes the noobaa-core part only. 
Added uml files for pluntuml vscode plugin.
Added related png images based on puml files


docs/design/quota.md
```diff
@@ -0,0 +1,121 @@
...
+            },
+        },
+        amount: {
+            type: 'object',
+            required: ['value'],
+            properties: {
+                value: {
+                    type: 'integer'
+                },
+                unit: {
```
dannyzaken, line 74: I don't think that unit is required for `amount`

docs/design/quota.md
```diff
@@ -0,0 +1,121 @@
...
+                    enum: ['GIGABYTE', 'TERABYTE', 'PETABYTE']
+                },
+            },
+        },
+        amount: {
+            type: 'object',
+            required: ['value'],
+            properties: {
+                value: {
+                    type: 'integer'
```
dannyzaken, line 72: I'm not sure if we need here `integer` or `bigint`. I think integer is enough. @nimrod-becker WDYT?

nimrod-becker: I think Integer is enough, MAX_SAFE_INTEGER even in bytes is still high enough and I Don't expect we'd overflow it

docs/design/quota.md
```diff
@@ -0,0 +1,121 @@
...
+* Provide an option to Configure Quota per Bucket on an OBC level and on the BucketClass level. If both exist, the OBC definition would override the BucketClass definition. 
+* Set a quota of object bucket claim, which limits the max number of objects per bucket (MaxObjects).
+* Set a quota of object bucket claim, which limits the max total size per bucket (MaxSize).
+* Remove quota configuration from the bucket. **?**
+* Update Quota’s on exiting buckets (updating the OBC only).
+* Update BucketClass quota, will be effective in new buckets using this BucketClass.
+* Add to NooBaa CLI new options for creating object bucket claims while providing one of the 2 types of quota.
+* Metrics and Alerts - upon calculation statistics based:
+    * Will provide the admin the quota status.
+    * Alert the admin when the number of objects / total size of objects reaches 80% and 100% of the quota. **# in the code 90% and 100% threashold?**
```
dannyzaken, line 18: @nimrod-becker what should be the right threshold? should we change the existing behavior to alert on 80%?

nimrod-becker: Sure, lets do 80 and 100.
In any case, I think the more important alerts are the mixings

igorpick: I suggest to pull the threshold's values out to config file (in json format) or env. then we can easy manage it.  
What do you think?

docs/design/quota.md
```diff
@@ -0,0 +1,124 @@
...
+While reading a bucket info, the process calculates the quota status:
+* QUOTA_NOT_SET - quota is disabled.
+* EXCEEDING_QUOTA - the bucket has exceeded one of the quota limits. (exceeded 100%). it also affets on the bucket mode.
+* APPROUCHING_QUOTA - the bucket approaches to one of the quota limits. (exceeded 90%).
+
+**The following sequence diagram illustrates the read buckets flow with quota configuration.**
+
+![Getting Started](./images/QuotaReadBucket.png)
+
+While an uploading objects, noobaa-core alerts, if the quota limits have been reached and throw an exception to client.
```
nimrod-becker, line 101:
```suggestion
While uploading objects, noobaa-core alerts, if the quota limits have been reached and throws an exception to client.
```

nimrod-becker (approved):

dannyzaken (approved):

PR #6607, opened 2021-06-24
Addition of NSFS tape files support

jeniawhite:
### Explain the changes
1. Addition of taped (sparse as well due to no POSIX command we rely on stat) files recognition and limit operations on them

1. Fixed https://issues.redhat.com/browse/MCGI-27

### Testing Instructions:
1.

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -16,6 +16,7 @@ const buffer_utils = require('../util/buffer_utils');
 const LRUCache = require('../util/lru_cache');
 const Semaphore = require('../util/semaphore');
 const nb_native = require('../util/nb_native');
+const RpcError = require('../rpc/rpc_error');
 
 const buffers_pool_sem = new Semaphore(config.NSFS_BUF_POOL_MEM_LIMIT);
 const buffers_pool = new buffer_utils.BuffersPool(config.NSFS_BUF_SIZE, buffers_pool_sem);
@@ -445,6 +446,7 @@ class NamespaceFS {
         try {
             await this._load_bucket(params, fs_account_config);
             const file_path = this._get_file_path(params);
+            await this._fail_if_archived_or_sparse_file(fs_account_config, file_path);
             file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
 
             const start = Number(params.start) || 0;
@@ -551,6 +553,7 @@ class NamespaceFS {
             ]);
             if (params.copy_source) {
                 const source_file_path = path.join(this.bucket_path, params.copy_source.key);
+                await this._fail_if_archived_or_sparse_file(fs_account_config, source_file_path);
                 try {
                     const same_inode = await this._is_same_inode(fs_account_config, source_file_path, file_path);
                     if (same_inode) return same_inode;
@@ -992,6 +995,7 @@ class NamespaceFS {
     }
 
     _translate_object_error_codes(err) {
+        if (err.rpc_code) return err;
         if (err.code === 'ENOENT') err.rpc_code = 'NO_SUCH_OBJECT';
         if (err.code === 'EEXIST') err.rpc_code = 'BUCKET_ALREADY_EXISTS';
         if (err.code === 'EPERM' || err.code === 'EACCES') err.rpc_code = 'UNAUTHORIZED';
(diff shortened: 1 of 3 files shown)


src/sdk/namespace_fs.js
```diff
@@ -153,7 +153,11 @@ class NamespaceFS {
         return bucket;
     }
 
-    is_server_side_copy(other, params) {
+    async is_server_side_copy(other, params, object_sdk) {
+        const fs_account_config = this.set_cur_fs_account_config(object_sdk);
+        await this._load_bucket(params, fs_account_config);
+        const file_path = this._get_file_path(params);
```
guymguym, line 159: Won't this be checking the target of the copy? Shouldn't we check the source?

src/sdk/namespace_fs.js
```diff
@@ -153,7 +153,11 @@ class NamespaceFS {
         return bucket;
     }
 
-    is_server_side_copy(other, params) {
+    async is_server_side_copy(other, params, object_sdk) {
```
guymguym, line 156: I can see that you had to change all to async just for this.
But I think we can do differently - in upload_object() we already have handling when link() fails to fallback to _copy_stream() so why not do that check there, where we have easier async flow to use.

src/sdk/namespace_fs.js
```diff
@@ -1127,6 +1134,20 @@ class NamespaceFS {
             throw err;
         }
     }
+
+    async _validate_data_access(fs_account_config, file_path) {
+        try {
+            const stat = await nb_native().fs.stat(fs_account_config, file_path);
+            if (isDirectory(stat)) return;
+            if (stat.blocks * 512 < stat.size) {
```
guymguym, line 1142: You can't just throw this logic in the code with zero comments.
It's impossible to understand why you are doing this.

src/sdk/namespace_fs.js
```diff
@@ -1127,6 +1134,20 @@ class NamespaceFS {
             throw err;
         }
     }
+
+    async _validate_data_access(fs_account_config, file_path) {
```
guymguym, line 1138: The name doesn't fit. I would make it explicit - _fail_if_archived_or_sparse_file()

src/sdk/namespace_fs.js
```diff
@@ -1127,6 +1134,20 @@ class NamespaceFS {
...
+        try {
+            const stat = await nb_native().fs.stat(fs_account_config, file_path);
+            if (isDirectory(stat)) return;
+            if (stat.blocks * 512 < stat.size) {
+                dbg.log0(`validate_data_access: ${file_path} rejected`, stat);
+                throw Object.assign(new Error('InvalidObjectState'), { rpc_code: 'INVALID_OBJECT_STATE' });
```
guymguym, line 1144: Why not new RpcError then?

guymguym (changes requested):

src/sdk/namespace_fs.js
```diff
@@ -1127,6 +1131,25 @@ class NamespaceFS {
...
+        try {
+            const stat = await nb_native().fs.stat(fs_account_config, file_path);
+            if (isDirectory(stat)) return;
+            // In order to verify if the file is stored in tape we compare sizes
+            // Multiple number of blocks by default block size and verify we get the size of the object
+            // If we get a size that is lower than the size of the object this means that it is taped or a spare file
+            // We had to use this logic since we do not have a POSIX call in order to verify that the file is taped
+            // This is why sparse files won't be accessible as well
+            if (stat.blocks * 512 < stat.size) {
+                dbg.log0(`validate_data_access: ${file_path} rejected`, stat);
```
guymguym, line 1145:
```suggestion
                dbg.log0(`_fail_if_archived_or_sparse_file: ${file_path} rejected`, stat);
```

src/sdk/namespace_fs.js
```diff
@@ -1127,6 +1131,25 @@ class NamespaceFS {
...
+                throw new RpcError('INVALID_OBJECT_STATE', 'Attempted to access archived or sparse file');
+            }
+        } catch (err) {
+            dbg.error(`validate_data_access: ${file_path} error`, err);
+            throw this._translate_object_error_codes(err);
+        }
```
guymguym, line 1151: Seems to me that this translate is redundant - this function is always called internally inside a try-catch that already calls translate. So I would remove this try catch inside this function altogether and let errors propagate.

src/sdk/namespace_fs.js
```diff
@@ -601,6 +603,7 @@ class NamespaceFS {
         let source_file;
         let buffer_pool_cleanup = null;
         try {
+            await this._fail_if_archived_or_sparse_file(fs_account_config, source_file_path);
```
guymguym, line 606: I think this is not the right spot - we said to check and fail even if we do link.
So this should be called in the upper level in `upload_object()` under the `if (params.copy_source) { ...` block

guymguym (changes requested):

guymguym (approved):

PR #6609, opened 2021-06-27
Setting fs debug level on startup if there is `process.env.NOOBAA_LOG_LEVEL`

liranmauda:
### Explain the changes
- Setting fs debug level on startup if there is `process.env.NOOBAA_LOG_LEVEL`

### Limitation
- This will only take effect if there is `process.env.NOOBAA_LOG_LEVEL`

### DIFF
--- a/src/util/debug_module.js
+++ b/src/util/debug_module.js
@@ -600,3 +600,7 @@ if (console_wrapper) {
     var conlogger = new DebugLogger("CONSOLE.js");
     console_wrapper.register_logger(conlogger);
 }
+
+if (Number(process.env.NOOBAA_LOG_LEVEL)) {
+    nb_native().fs.set_debug_level(Number(process.env.NOOBAA_LOG_LEVEL));
+}


src/util/debug_module.js
```diff
@@ -539,6 +539,9 @@ DebugLogger.prototype.log2_withbt = log_bt_builder(2);
 DebugLogger.prototype.log3_withbt = log_bt_builder(3);
 DebugLogger.prototype.log4_withbt = log_bt_builder(4);
 
+if (process.env.NOOBAA_LOG_LEVEL) {
```
guymguym, line 542:
```suggestion
if (Number(process.env.NOOBAA_LOG_LEVEL)) {
```

src/util/debug_module.js
```diff
@@ -539,6 +539,9 @@ DebugLogger.prototype.log2_withbt = log_bt_builder(2);
 DebugLogger.prototype.log3_withbt = log_bt_builder(3);
 DebugLogger.prototype.log4_withbt = log_bt_builder(4);
 
+if (Number(process.env.NOOBAA_LOG_LEVEL)) {
```
guymguym, line 542: I just noticed another thing - why did you choose this location in the file, where this statement is breaking the flow of defining DebugLogger class methods? I think I would push this all the way down.

guymguym (approved):

PR #6592, opened 2021-06-20
Add threshold for FS Calls and print on actions taking over those thresholds

liranmauda:
### Explain the changes

- Add threshold for FS Calls and print on actions taking over those thresholds

Signed-off-by: liranmauda <liran.mauda@gmail.com>

PR #6610, opened 2021-06-27
Changing the way we do threshold

liranmauda:
### Explain the changes
- Changing the way we do threshold

Changes to #6592 

### DIFF
--- a/src/core/nsfs.js
+++ b/src/core/nsfs.js
@@ -9,6 +9,7 @@ const dbg = require('../util/debug_module')(__filename);
 if (!dbg.get_process_name()) dbg.set_process_name('nsfs');
 dbg.original_console();
 
+const config = require('../../config');
 const nb_native = require('../util/nb_native');
 const ObjectSDK = require('../sdk/object_sdk');
 const NamespaceFS = require('../sdk/namespace_fs');
@@ -80,7 +81,8 @@ async function main(argv = minimist(process.argv.slice(2))) {
         let fs_config = {
             uid: Number(argv.uid) || process.getuid(),
             gid: Number(argv.gid) || process.getgid(),
-            backend: ''
+            backend: '',
+            warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
         };
 
         if (!fs.existsSync(fs_root)) {
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -77,7 +77,7 @@ struct FSWorker : public Napi::AsyncWorker
     std::string _backend;
     std::string _desc;
     int _errno;
-    int thresholdInMilliSec = 2;
+    int _warn_threshold_ms;
 
     FSWorker(const Napi::CallbackInfo& info)
         : AsyncWorker(info.Env())
@@ -86,11 +86,13 @@ struct FSWorker : public Napi::AsyncWorker
         , _uid(ThreadScope::orig_uid)
         , _gid(ThreadScope::orig_gid)
         , _errno(0)
+        , _warn_threshold_ms(0)
     {
         Napi::Object config = info[0].As<Napi::Object>();
         if (config.Has("uid")) _uid = config.Get("uid").ToNumber();
         if (config.Has("gid")) _gid = config.Get("gid").ToNumber();
         if (config.Has("backend")) _backend = config.Get("backend").ToString();
+        if (config.Has("warn_threshold_ms")) _warn_threshold_ms = config.Get("warn_threshold_ms").ToNumber();
     }
     void Begin(std::string desc)
     {
@@ -100,21 +102,17 @@ struct FSWorker : public Napi::AsyncWorker
     virtual void Work() = 0;
     void Execute() override
     {
-        using std::chrono::duration_cast; 
-        using std::chrono::milliseconds;
-        using std::chrono::system_clock;
-
         DBG1("FS::FSWorker::Execute: " << _desc << DVAL(_uid) << DVAL(_gid) << DVAL(_backend));
-        if (_desc.find("FileWritev") != std::string::npos ) {
-            thresholdInMilliSec = 100;
-        }
         ThreadScope tx;
         tx.set_user(_uid, _gid);
-        std::time_t timeBefore = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
+        auto start_time = std::chrono::high_resolution_clock::now();
         Work();
-        std::time_t timeAfter = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
-        if (timeAfter - timeBefore > thresholdInMilliSec) {
-            DBG0("FS::FSWorker::Execute: " << _desc << " took: " << timeAfter - timeBefore << " milliseconds");
+        auto end_time = std::chrono::high_resolution_clock::now();
+        double took_time = std::chrono::duration<double, std::milli>(end_time - start_time).count();
+        if (_warn_threshold_ms && took_time > _warn_threshold_ms) {
+            DBG0("FS::FSWorker::Execute: WARNING " << _desc << " took too long: " << took_time << " ms");
+        } else {
+            DBG1("FS::FSWorker::Execute: " << _desc << " took: " << took_time << " ms");
         }
     }
     void SetSyscallError()
--- a/src/sdk/bucketspace_nb.js
+++ b/src/sdk/bucketspace_nb.js
@@ -6,6 +6,7 @@ const P = require('../util/promise');
 const nb_native = require('../util/nb_native');
 const dbg = require('../util/debug_module')(__filename);
 const path = require('path');
+const config = require('../../config.js');
 
 /**
  * @implements {nb.BucketSpace}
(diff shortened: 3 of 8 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -77,7 +77,8 @@ struct FSWorker : public Napi::AsyncWorker
     std::string _backend;
     std::string _desc;
     int _errno;
-    int thresholdInMilliSec = 2;
+    protected:
```
guymguym, line 80: on second thought, no need to go into protected visibility.
this is a struct, so all is public by definition unless otherwise specified.
I would simply remove the protected keyword and it should be fine.

src/native/fs/fs_napi.cpp
```diff
@@ -77,7 +77,8 @@ struct FSWorker : public Napi::AsyncWorker
...
+        int _warn_threshold_ms = 2;
```
guymguym, line 81: I think the warn threshold should generally be higher.
I would aim for 100 ms.
but in any case I think it should be set in config.js and passed to the fs workers in the config param (where we send uid/gid/backend).

src/native/fs/fs_napi.cpp
```diff
@@ -105,16 +106,14 @@ struct FSWorker : public Napi::AsyncWorker
         using std::chrono::system_clock;
```
guymguym, line 106: I think these using directives are now unneeded, right?

src/core/nsfs.js
```diff
@@ -80,7 +80,8 @@ async function main(argv = minimist(process.argv.slice(2))) {
         let fs_config = {
             uid: Number(argv.uid) || process.getuid(),
             gid: Number(argv.gid) || process.getgid(),
-            backend: ''
+            backend: '',
+            warn_threshold_ms: 100
```
guymguym, line 84:
```suggestion
            warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
```

src/native/fs/fs_napi.cpp
```diff
@@ -86,11 +86,13 @@ struct FSWorker : public Napi::AsyncWorker
         , _uid(ThreadScope::orig_uid)
         , _gid(ThreadScope::orig_gid)
         , _errno(0)
+        , _warn_threshold_ms(100)
```
guymguym, line 89: I think the default should be to not warn if not provided from options
```suggestion
        , _warn_threshold_ms(0)
```

src/native/fs/fs_napi.cpp
```diff
@@ -763,6 +760,7 @@ struct FileWritev : public FSWrapWorker<FileWrap>
         auto buffers = info[1].As<Napi::Array>();
         const int buffers_len = buffers.Length();
         int total_len = 0;
+        _warn_threshold_ms = 400;
```
guymguym, line 763: is this really necessary to be hard coded here?
I would use the same value as the other ops unless it's not working.
If you still have to pass a different value - send it from js.

src/native/fs/fs_napi.cpp
```diff
@@ -100,21 +102,16 @@ struct FSWorker : public Napi::AsyncWorker
...
     {
-        using std::chrono::duration_cast; 
-        using std::chrono::milliseconds;
-        using std::chrono::system_clock;
-
         DBG1("FS::FSWorker::Execute: " << _desc << DVAL(_uid) << DVAL(_gid) << DVAL(_backend));
-        if (_desc.find("FileWritev") != std::string::npos ) {
-            thresholdInMilliSec = 100;
-        }
+        DBG0("LMLM _warn_threshold_ms is: " << _warn_threshold_ms);
```
guymguym, line 106: delete

src/sdk/namespace_fs.js
```diff
@@ -23,7 +23,8 @@ const buffers_pool = new buffer_utils.BuffersPool(config.NSFS_BUF_SIZE, buffers_
 /*const DEFAULT_FS_CONFIG = {
     uid: Number(process.env.NSFS_UID) || process.getuid(),
     gid: Number(process.env.NSFS_GID) || process.getgid(),
-    backend: ''
+    backend: '',
+    warn_threshold_ms: 100
```
guymguym, line 27: this entire comment can be deleted

src/native/fs/fs_napi.cpp
```diff
@@ -100,21 +102,16 @@ struct FSWorker : public Napi::AsyncWorker
...
+        auto start_time = std::chrono::high_resolution_clock::now();
         Work();
-        std::time_t timeAfter = duration_cast<milliseconds>(system_clock::now().time_since_epoch()).count();
-        if (timeAfter - timeBefore > thresholdInMilliSec) {
-            DBG0("FS::FSWorker::Execute: " << _desc << " took: " << timeAfter - timeBefore << " milliseconds");
+        auto end_time = std::chrono::high_resolution_clock::now();
+        double took_time = std::chrono::duration<double, std::milli>(end_time - start_time).count();
+        if (took_time > _warn_threshold_ms) {
+            DBG0("FS::FSWorker::Execute: WARNING" << _desc << " took too long: " << took_time << " milliseconds");
         }
```
guymguym, line 115:
```suggestion
        if (_warn_threshold_ms && took_time > _warn_threshold_ms) {
            DBG0("FS::FSWorker::Execute: WARNING" << _desc << " took too long: " << took_time << " ms");
        } else {
            DBG1("FS::FSWorker::Execute: " << _desc << " took: " << took_time << " ms");
        }
```

src/sdk/namespace_fs.js
```diff
@@ -138,6 +139,7 @@ class NamespaceFS {
         }
 
         fs_config_param.backend = this.fs_backend || '';
+        fs_config_param.warn_threshold_ms = config.NSFS_WARN_THRESHOLD_MS || 100;
```
guymguym, line 142:
```suggestion
        fs_config_param.warn_threshold_ms = config.NSFS_WARN_THRESHOLD_MS;
```

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -19,7 +19,8 @@ const inspect = (x, max_arr = 5) => util.inspect(x, { colors: true, depth: null,
 const DEFAULT_FS_CONFIG = {
     uid: process.getuid(),
     gid: process.getgid(),
-    backend: ''
+    backend: '',
+    warn_threshold_ms: 100
```
guymguym, line 23:
```suggestion
    warn_threshold_ms: 100,
```

src/test/unit_tests/test_nb_native_fs.js
```diff
@@ -10,7 +10,8 @@ const nb_native = require('../../util/nb_native');
 const DEFAULT_FS_CONFIG = {
     uid: process.getuid(),
     gid: process.getgid(),
-    backend: ''
+    backend: '',
+    warn_threshold_ms: 100
```
guymguym, line 14:
```suggestion
    warn_threshold_ms: 100,
```

src/server/bg_services/namespace_monitor.js
```diff
@@ -153,7 +153,10 @@ class NamespaceMonitor {
 
     async test_nsfs_resource(nsr) {
         try {
-            const fs_config_param = { backend: nsr.nsfs_config.fs_backend || '' };
+            const fs_config_param = {
+                backend: nsr.nsfs_config.fs_backend || '',
+                warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS
```
guymguym, line 158:
```suggestion
                warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
```

src/sdk/bucketspace_nb.js
```diff
@@ -216,10 +217,11 @@ class BucketSpaceNB {
...
 
             await nb_native().fs.checkAccess({
-                    uid: account.nsfs_account_config.uid,
-                    gid: account.nsfs_account_config.gid,
-                    backend: namespace_bucket_config.write_resource.resource.fs_backend
-                }, path.join(namespace_bucket_config.write_resource.resource.fs_root_path, namespace_bucket_config.write_resource.path || ''));
+                uid: account.nsfs_account_config.uid,
+                gid: account.nsfs_account_config.gid,
+                backend: namespace_bucket_config.write_resource.resource.fs_backend,
+                warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS
```
guymguym, line 223:
```suggestion
                warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
```

src/native/fs/fs_napi.cpp
```diff
@@ -86,11 +86,13 @@ struct FSWorker : public Napi::AsyncWorker
         , _uid(ThreadScope::orig_uid)
         , _gid(ThreadScope::orig_gid)
         , _errno(0)
+        , _warn_threshold_ms(100)
     {
         Napi::Object config = info[0].As<Napi::Object>();
         if (config.Has("uid")) _uid = config.Get("uid").ToNumber();
         if (config.Has("gid")) _gid = config.Get("gid").ToNumber();
         if (config.Has("backend")) _backend = config.Get("backend").ToString();
+        if (config.Has("warn_threshold_ms")) _warn_threshold_ms = config.Get("warn_threshold_ms").ToNumber();
```
guymguym, line 95: I wonder - if we send {warn_threshold_ms: undefined} does this condition evaluate to true or false?
We want to get false in that case, but not sure that Has() does that..

src/native/fs/fs_napi.cpp
liranmauda: 👍

src/native/fs/fs_napi.cpp
liranmauda: I tested it and when we pass {warn_threshold_ms: undefined}  we will keep `_warn_threshold_ms(0)`
I guess that is what we are expecting.

guymguym (approved):

PR #6638, opened 2021-07-04
BucketSpaceFS using native ops

liranmauda:
### Explain the changes

BucketSpaceFS using native ops

### Testing Instructions:
1.  Use standalone and do NSFS Bucket operations

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -1,10 +1,24 @@
 /* Copyright (C) 2020 NooBaa */
 'use strict';
 
-const fs = require('fs');
 const path = require('path');
+const config = require('../../config');
+const nb_native = require('../util/nb_native');
 const SensitiveString = require('../util/sensitive_string');
 const { S3Error } = require('../endpoint/s3/s3_errors');
+
+//TODO:  dup from namespace_fs - need to handle and not dup code
+function isDirectory(ent) {
+    if (!ent) throw new Error('isDirectory: ent is empty');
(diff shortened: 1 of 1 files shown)


src/sdk/bucketspace_fs.js
```diff
@@ -21,13 +41,13 @@ class BucketSpaceFS {
...
-            const entries = await fs.promises.readdir(this.fs_root, { withFileTypes: true });
-            const dirs_only = entries.filter(e => e.isDirectory());
+            const entries = await nb_native().fs.readdir(this.fs_config_param, this.fs_root);
+            const dirs_only = entries.filter(entree => isDirectory(entree));
             const buckets = dirs_only.map(e => ({ name: new SensitiveString(e.name) }));
             return { buckets };
         } catch (err) {
             if (err.code === 'ENOENT') {
-                console.error('nsfs: root dir not found', err);
+                console.error('BucketSpaceFS: root dir not found', err);
```
igorpick, line 50: Don't we need to use the dbg?

liranmauda: I don't think so. I think that we want this error.
@guymguym WDYT?

igorpick: @liranmauda I mean instead of using 'console.error()', use dbg.error()

liranmauda: Maybe, but in any case not in the scope of this PR. 
The dbg is not even required in this file.

igorpick (approved):

PR #6640, opened 2021-07-04
nodes monitor | Waiting a while for system_store to be set up.

liranmauda:
### Explain the changes

- In nodes monitor, waiting a while for system_store to be set up. after a while will despair.

### DIFF
--- a/src/server/node_services/nodes_monitor.js
+++ b/src/server/node_services/nodes_monitor.js
@@ -205,6 +205,26 @@ class NodesMonitor extends EventEmitter {
         this.n2n_agent.set_rpc_address('n2n://nodes_monitor');
     }
 
+    waitForValue(accessor, pullInterval = 1000) {
+        let count = 1;
+        return new Promise(resolve => {
+            function pull() {
+                const val = accessor();
+                if (_.isUndefined(val)) {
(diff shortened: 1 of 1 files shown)


src/server/node_services/nodes_monitor.js
```diff
@@ -205,6 +205,26 @@ class NodesMonitor extends EventEmitter {
         this.n2n_agent.set_rpc_address('n2n://nodes_monitor');
     }
 
+    waitForValue(accessor, pullInterval = 1000) {
+        let count = 1;
+        return new Promise(resolve => {
+            function pull() {
+                const val = accessor();
+                if (_.isUndefined(val)) {
+                    setTimeout(pull, pullInterval);
```
igorpick, line 214: Lets add a log message that we are still waiting for the system and how many retries?

liranmauda: Regrading adding logs, We can.
Regarding how many retries, I don't see a value much for that. We can retry forever but we will stop at some point from desperation 🙂 

I started to doubt if this PR is actually needed... @dannyzaken WDYT?

liranmauda:
Closing, Not liking this PR 🙂

PR #6651, opened 2021-07-07
Addition of native crypto

jeniawhite:
### Explain the changes
1. Added crypto to our native module with support of async MD5
2. TODO: FIPS, NSFS Code

1. 

### Testing Instructions:
1. Added unit_test for MD5 Async as part of the native hash test

### DIFF
--- a/src/native/tools/crypto_napi.cpp
+++ b/src/native/tools/crypto_napi.cpp
@@ -1,0 +1,176 @@
+/* Copyright (C) 2016 NooBaa */
+#include <string.h>
+#include <vector>
+#include "../third_party/isa-l_crypto/include/md5_mb.h"
+#include "../util/common.h"
+#include "../util/endian.h"
+#include "../util/napi.h"
+
+namespace noobaa
+{
+
+template <typename T>
+static Napi::Value
+api(const Napi::CallbackInfo& info)
+{
+    auto w = new T(info);
+    Napi::Promise promise = w->_deferred.Promise();
+    w->Queue();
+    return promise;
+}
+
+/**
+ * CryptoWorker is a general async worker for our crypto operations
+ */
+struct CryptoWorker : public Napi::AsyncWorker
+{
+    Napi::Promise::Deferred _deferred;
+
+    CryptoWorker(const Napi::CallbackInfo& info)
+        : AsyncWorker(info.Env())
+        , _deferred(Napi::Promise::Deferred::New(info.Env()))
+    {
+    }
+    virtual void OnOK() override
+    {
+        // LOG("CryptoWorker::OnOK: undefined");
+        _deferred.Resolve(Env().Undefined());
+    }
+    virtual void OnError(Napi::Error const& error) override
+    {
+        LOG("CryptoWorker::OnError: " << DVAL(error.Message()));
+        auto obj = error.Value();
+        _deferred.Reject(obj);
+    }
+};
+
+/**
+ * CryptoWrapWorker is meant to simplify adding async CryptoWorker instance methods to ObjectWrap types
+ * like MD5Wrap, while keeping the object referenced during that action.
+ */
+template <typename Wrapper>
+struct CryptoWrapWorker : public CryptoWorker
+{
+    Wrapper* _wrap;
+    CryptoWrapWorker(const Napi::CallbackInfo& info)
+        : CryptoWorker(info)
+    {
+        _wrap = Wrapper::Unwrap(info.This().As<Napi::Object>());
+        _wrap->Ref();
+    }
+    ~CryptoWrapWorker()
+    {
+        _wrap->Unref();
+    }
+};
+
+struct MD5Wrap : public Napi::ObjectWrap<MD5Wrap>
+{
+    size_t _NWORDS = MD5_DIGEST_NWORDS;
+    bool _WORDS_BE = false;
+    void (*_INIT)(MD5_HASH_CTX_MGR*) = md5_ctx_mgr_init;
+    MD5_HASH_CTX *(*_SUBMIT)(MD5_HASH_CTX_MGR*, MD5_HASH_CTX*, const void*, uint32_t, HASH_CTX_FLAG) = md5_ctx_mgr_submit;
+    MD5_HASH_CTX *(*_FLUSH)(MD5_HASH_CTX_MGR*) = md5_ctx_mgr_flush;
+    DECLARE_ALIGNED(MD5_HASH_CTX_MGR _mgr, 16);
+    DECLARE_ALIGNED(MD5_HASH_CTX _ctx, 16);
+
+    static Napi::FunctionReference constructor;
+    static void init(Napi::Env env)
+    {
+        constructor = Napi::Persistent(DefineClass(
+            env,
+            "MD5",
+            {
+                InstanceMethod("update", &MD5Wrap::update),
+                InstanceMethod("digest", &MD5Wrap::digest),
+            }));
+        constructor.SuppressDestruct();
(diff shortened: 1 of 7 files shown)


src/test/unit_tests/test_nb_native_hashes.js
```diff
@@ -36,8 +40,8 @@ mocha.describe('nb_native hashes', function() {
             '123456789',
             '1234567890',
         ]) {
-        mocha.it(`MD5 ${s}`, function() {
-            md5(Buffer.from(s));
+        mocha.it(`MD5 ${s}`, async function() {
+            await md5(Buffer.from(s));
```
guymguym, line 44: Why do you replace the other md5 call and not just add your calls?
We want the test to keep validating both impls for now.

src/native/tools/crypto_napi.cpp
```diff
@@ -0,0 +1,176 @@
...
+
+struct MD5Wrap : public Napi::ObjectWrap<MD5Wrap>
+{
+    size_t _NWORDS = MD5_DIGEST_NWORDS;
+    bool _WORDS_BE = false;
+    void (*_INIT)(MD5_HASH_CTX_MGR*) = md5_ctx_mgr_init;
+    MD5_HASH_CTX *(*_SUBMIT)(MD5_HASH_CTX_MGR*, MD5_HASH_CTX*, const void*, uint32_t, HASH_CTX_FLAG) = md5_ctx_mgr_submit;
+    MD5_HASH_CTX *(*_FLUSH)(MD5_HASH_CTX_MGR*) = md5_ctx_mgr_flush;
+    DECLARE_ALIGNED(MD5_HASH_CTX_MGR _mgr, 16);
+    DECLARE_ALIGNED(MD5_HASH_CTX _ctx, 16);
```
liranmauda, line 75: We can have `typedef Hasher` same as we have it here:
https://github.com/noobaa/noobaa-core/blob/8aa2d394fb716f060d950594ae720f3674d41318/src/native/tools/ssl_napi.cpp#L130-L139

liranmauda (changes requested):
We should also add SHA1 and SHA256 as interfaces

src/native/tools/crypto_napi.cpp
guymguym: Right, I don't really understand why you would copy the Hasher template code and then remove the templating and then have just md5 instead of any type?

guymguym: Also, I think that we should use OpenSSL by default and only fallback to ISA-L for fips mode.

jeniawhite: We can do that in the subsequent PR after the MVP scope.
For the MVP we need MD5 solely.
Did not add SHA1 in order to minimize the footprint and deliver today, also not to mess with FIPS.

src/test/unit_tests/test_nb_native_hashes.js
jeniawhite: It is validating both implementations, I've just transformed this to async and called mine inside as well.
I can add another function for mine that does the same thing if that is what you've meant.

liranmauda (approved):
Talked to @jeniawhite 
Approving.
We need to open a Gap for SHA1, FIPS, and rearrange this with Hasher (We should do it for the MVP)

src/sdk/namespace_fs.js
```diff
@@ -1011,6 +1042,30 @@ class NamespaceFS {
         return p.endsWith('/') ? p + config.NSFS_FOLDER_OBJECT_NAME : p;
     }
 
+    _assign_md5_to_fs_xattr(md5_digest, fs_xattr) {
+        // TODO: Assign content_md5_mtime
+        fs_xattr = Object.assign(fs_xattr || {}, {
+            [XATTR_MD5_KEY]: md5_digest
+        });
+        return fs_xattr;
```
guymguym, line 1050: Why not initialize the object from the caller, and then just update the fs_xattr object?
```
fs_xattr[XATTR_MD5_KEY] = md5_digest;
```

src/sdk/namespace_fs.js
```diff
@@ -799,10 +820,12 @@ class NamespaceFS {
                     const num = Number(e.name.slice('part-'.length));
                     const part_path = path.join(params.mpu_path, e.name);
                     const stat = await nb_native().fs.stat(fs_account_config, part_path);
+                    const fs_xattr =
+                        config.NSFS_CALCULATE_MD5 ? await this._get_fs_xattr_from_path(fs_account_config, part_path) : undefined;
```
guymguym, line 824: why from path? why not open the file and then get both stat and xattr?

src/sdk/namespace_fs.js
```diff
@@ -823,6 +846,7 @@ class NamespaceFS {
         let buffer_pool_cleanup = null;
         const fs_account_config = this.set_cur_fs_account_config(object_sdk);
         try {
+            let MD5Async = config.NSFS_CALCULATE_MD5 ? new (nb_native().crypto.MD5Async)() : undefined;
```
guymguym, line 849: This is an object, not a class name, and it should be const:
```
            const md5_hasher = config.NSFS_CALCULATE_MD5 ? new (nb_native().crypto.MD5Async)() : undefined;
```

src/sdk/namespace_fs.js
```diff
@@ -593,21 +598,21 @@ class NamespaceFS {
...
                     dbg.warn('NamespaceFS: COPY using link failed with:', e);
-                    await this._copy_stream(source_file_path, upload_path, fs_account_config, fs_xattr);
+                    fs_xattr = await this._copy_stream(source_file_path, upload_path, fs_account_config, fs_xattr);
                 }
             } else {
-                await this._upload_stream(params.source_stream, upload_path, fs_account_config, object_sdk.rpc_client, fs_xattr);
+                fs_xattr = await this._upload_stream(params.source_stream, upload_path, fs_account_config, object_sdk.rpc_client, fs_xattr);
             }
             // TODO use file xattr to store md5_b64 xattr, etc.
             const stat = await nb_native().fs.stat(fs_account_config, upload_path);
```
liranmauda, line 613: Similar to @guymguym Comments,
If we are returning `fs_xattr` from `_copy_stream` and `_upload_stream` and we have opened the file there, we should `stat` the fd (using `fstat`) and return it also, then skip this `stat`.

src/sdk/namespace_fs.js
```diff
@@ -723,9 +738,14 @@ class NamespaceFS {
                 }
             }
             await flush();
+            if (MD5Async) {
+                fs_xattr = this._assign_md5_to_fs_xattr((await MD5Async.digest()).toString('hex'), fs_xattr);
+            }
             if (fs_xattr) {
                 await target_file.setxattr(fs_account_config, fs_xattr);
             }
```
liranmauda, line 746: After we done with `target_file` we should close it and make `target_file = null` and not count on the finally.

src/sdk/namespace_fs.js
```diff
@@ -654,11 +661,15 @@ class NamespaceFS {
                 if (!bytesRead) break;
                 read_pos += bytesRead;
                 const data = buffer.slice(0, bytesRead);
+                if (MD5Async) await MD5Async.update(data);
                 await target_file.write(fs_account_config, data);
```
guymguym, line 665: Why not call these two in parallel?

guymguym:
```js
await Promise.all([
    MD5Async && MD5Async.update(data),
    target_file.write(fs_account_config, data),
]);
```

PR #6654, opened 2021-07-08
Fixing issue of duplicate buckets in DB

### LINKED ISSUE #6584: Creation bucket caused noobaa core and noobaa endpoint pods crash
### Environment info
- NooBaa Version: master-20210607
- Platform: OCP 4.6.16

### Actual behavior
1. Creation of multiple buckets caused multiple noobaa pods crash

### Expected behavior
1. No pods should crash 

### Steps to reproduce
1. Started a loop for 100's of buckets. 
2. In between, it failed creation of buckets operation. 
Overall 352 buckets were created.

**No IO was running for this test.**

Noobaa pods and noobaa endpoint pods crashed and noobaa status is failing

```
[root@ocp-akshat-1-inf ~]# podn
NAME                                               READY   STATUS             RESTARTS   AGE
noobaa-core-0                                      0/1     CrashLoopBackOff   4          24h
noobaa-db-pg-0                                     1/1     Running            0          24h
noobaa-default-backing-store-noobaa-pod-615be21a   0/1     Terminating        0          6d16h
noobaa-endpoint-74dfc4b786-468kd                   1/1     Running            7          24h
noobaa-endpoint-74dfc4b786-j9d49                   1/1     Running            0          3m33s
noobaa-operator-7996cfc767-h2lnd                   1/1     Running            0          24h
```

jackyalbo:

### Explain the changes
1.  protecting both create and delete bucket ops with semaphore - So two creates/deletes of the same bucket name won't able to run on the same time 

1.  #6584
2. 
### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -36,6 +36,8 @@ const chunk_config_utils = require('../utils/chunk_config_utils');
 const NetStorage = require('../../util/NetStorageKit-Node-master/lib/netstorage');
 const { OP_NAME_TO_ACTION } = require('../../endpoint/s3/s3_utils');
 const path = require('path');
+const KeysSemaphore = require('../../util/keys_semaphore');
+const bucket_semaphore = new KeysSemaphore(1);
 
 const VALID_BUCKET_NAME_REGEXP =
     /^(([a-z0-9]|[a-z0-9][a-z0-9-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9-]*[a-z0-9])$/;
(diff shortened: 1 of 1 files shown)


jackyalbo:
I'm making this PR as a quick fix - I think better solution is to enforce this in the DB like in mongo. It seems postgres doesn't handle a multiple field index the same way as mongo and we should fix. Maybe in another PR.

liranmauda (approved):
LGTM, 
But as mentioned in:
> I'm making this PR as a quick fix - I think better solution is to enforce this in the DB like in mongo. It seems postgres doesn't handle a multiple field index the same way as mongo and we should fix. Maybe in another PR.
We should probably handle a multiple field index As @jackyalbo mentioned in Slack:
```
CREATE UNIQUE INDEX idx_btree_buckets2_name ON buckets USING BTREE ((data ->> 'name'::text), (data ->> 'system'::text)) WHERE data ->> 'deleted' is null;
```

src/server/system_services/bucket_server.js
```diff
@@ -36,6 +36,9 @@ const chunk_config_utils = require('../utils/chunk_config_utils');
 const NetStorage = require('../../util/NetStorageKit-Node-master/lib/netstorage');
 const { OP_NAME_TO_ACTION } = require('../../endpoint/s3/s3_utils');
 const path = require('path');
+const KeysSemaphore = require('../../util/keys_semaphore');
+const bucket_create_semaphore = new KeysSemaphore(1);
+const bucket_delete_semaphore = new KeysSemaphore(1);
```
dannyzaken, line 41: shouldn't there be just one key semaphore? no reason to allow deletion while creation is in progress

liranmauda: @jackyalbo And I talked about that,
Why not? Why we should not allow deletion while creation is in progress?

dannyzaken: why should we? if the bucket creation is in progress, the behavior of the deletion is inconsistent. it might work or not. 
in any case, it's an edge case. but if we already surrounding delete_bucket with a key semaphore I think we should use the same as create

jackyalbo: ok, after talking a bit with @dannyzaken we decided to change to one semaphore

dannyzaken (approved):

src/server/system_services/bucket_server.js
liranmauda: Yes, It is an edge case, this is only when trying to delete a bucket with the same name as one that is currently being created.

PR #6657, opened 2021-07-11
Fixing db unique enforcement for system store

jackyalbo:

### Explain the changes
1. As postgres doesn't restrict uniqueness as mongo does. (If part of the key - deleted in our case is not in the new row - it will not throw the insert operation) Will remove deleted from the key, but also will remove all deleted rows from the index so new row with same name can be created if previous one was already deleted.

1.  

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/schemas/account_indexes.js
+++ b/src/server/system_services/schemas/account_indexes.js
@@ -4,9 +4,11 @@
 module.exports = [{
     fields: {
         email: 1,
-        deleted: 1
     },
     options: {
         unique: true,
+        partialFilterExpression: {
+            deleted: null,
+        }
     }
 }, ];
--- a/src/server/system_services/schemas/bucket_indexes.js
+++ b/src/server/system_services/schemas/bucket_indexes.js
@@ -5,9 +5,11 @@ module.exports = [{
     fields: {
         system: 1,
         name: 1,
-        deleted: 1
     },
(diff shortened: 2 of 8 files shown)


dannyzaken:
do we need to change also in migration_to_postgres?

https://github.com/noobaa/noobaa-core/blob/31cd9ccf2100fd74e6d38fb05f5246b1958c5fbd/src/upgrade/migration_to_postgres.js#L44-L62


Also, make sure it works after upgrading from previous versions. it should be ok but check that there are no conflicts in names or something else. I don't think there should be a problem with leaving old indexes around, but it's also worth testing

jackyalbo:
> do we need to change also in migration_to_postgres?
> 
> https://github.com/noobaa/noobaa-core/blob/31cd9ccf2100fd74e6d38fb05f5246b1958c5fbd/src/upgrade/migration_to_postgres.js#L44-L62
> 
> Also, make sure it works after upgrading from previous versions. it should be ok but check that there are no conflicts in names or something else. I don't think there should be a problem with leaving old indexes around, but it's also worth testing

Updated migration_to_db as well - also updated it to have the same order of the fields. 
Checked upgrade and it seems to just add a new index - the new one indeed blocks duplicate rows - (tried from psql)

dannyzaken (approved):

PR #6659, opened 2021-07-12
reverting package lock changes

kfir-payne:

### Explain the changes
1. reverting package lock changes

1. 

### Testing Instructions:
1.


dannyzaken (approved):

liranmauda (approved):
This `package-lock.json` change is due to npm v7
We can see that we have `lockfileVersion: 2` which implies that npm v7 was used and the package-lock.json is still backward compatible with v5 and v6
As we all still use npm v6 and also the DS relay on it we want to keep the `lockfileVersion: 1`, hance reverting.

PR #6663, opened 2021-07-13
skip tiering creation for namespace buckets

dannyzaken:

related to https://github.com/noobaa/noobaa-operator/pull/686

### Explain the changes
1. in `create_bucket` if the created bucket is a namespace bucket, skip the tiering\tier creation

1. fix for https://bugzilla.redhat.com/show_bug.cgi?id=1981331

### Testing Instructions:
1.

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -111,7 +111,7 @@ async function create_bucket(req) {
             };
             should_create_underlying_storage = true;
 
-        } else {
+        } else if (_.isUndefined(req.rpc_params.namespace) || req.rpc_params.namespace.caching) {
             // we create dedicated tier and tiering policy for the new bucket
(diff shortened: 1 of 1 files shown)


guymguym (changes requested):
@dannyzaken This makes cache buckets unusable - their policy requires both namespace and tiering.

Why is this change needed??

dannyzaken:
@guymguym, you're right. changed to include cache bucket.

this is needed since this is failing creation of NS (non-cache) buckets in OCS, where the internal storage is disabled. the create_bucket fails on `validate_pool_constraints`. a previous fix was done in the operator by sending the default bucket class policy as the tiering policy, but it then causes OBC provision to fail if the defaultBucketClass is not ready (see https://github.com/noobaa/noobaa-operator/pull/686)

romayalon (approved):

PR #6667, opened 2021-07-15
Add deletion of bucket stats during bucket delete

### LINKED ISSUE #6212: NS-CACHE: Metrics cleanup for bucket names - GAP
### Environment info
- NooBaa Version: **5.6**
- Platform: **NA**

### Actual behavior
1. Collecting cache metrics with bucket-name labels means we need to cleanup those metrics when the bucket is deleted.
2. See update_namespace_cache_stats() 
https://github.com/noobaa/noobaa-core/blob/43dd7bbb1fea7b4e772ec2a6c0276b0afbf2e810/src/sdk/endpoint_stats_collector.js#L118-L133

### Expected behavior
1. Should not leave garbage in endpoint memory/prometheus.

### Steps to reproduce
1. @nb-ohad @jeniawhite

BarakSason:
Prior to this change, bucket deletion would not trigger a deletion of
the bucket-related stats from the db which would result in leaving
redundant data in the db. This path modifies the bucket_delete flow to
include a cleaning of the relevant stats from the db.

Fixed: #6212

### DIFF
--- a/src/server/analytic_services/bucket_stats_store.js
+++ b/src/server/analytic_services/bucket_stats_store.js
@@ -71,6 +71,9 @@ class BucketStatsStore {
         return [];
     }
 
+    async delete_stats({ system, bucket }) {
+        await this._bucket_stats.deleteOne({ system, bucket });
+    }
 }
 
 // EXPORTS
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -843,7 +843,7 @@ async function delete_bucket(req) {
         var bucket = find_bucket(req);
         // TODO before deleting tier and tiering_policy need to check they are not in use
         let tiering_policy = bucket.tiering;
-        const reason = await can_delete_bucket(req.system, bucket);
+        const reason = await can_delete_bucket(bucket);
         if (reason) {
             throw new RpcError(reason, 'Cannot delete bucket');
         }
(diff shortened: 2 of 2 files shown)


src/server/system_services/bucket_server.js
```diff
@@ -1839,7 +1843,7 @@ function resolve_tiering_policy(req, policy_name) {
     return tiering_policy;
 }
 
-function can_delete_bucket(system, bucket) {
```
BarakSason, line 1842: Unused var.
Not strictly a part of the fix, but thought to fix as I came across it.

src/server/analytic_services/bucket_stats_store.js
```diff
@@ -71,6 +71,9 @@ class BucketStatsStore {
         return [];
     }
 
+    async delete_stats({ system, bucket }) {
+        await this._bucket_stats.deleteOne({ system: system, bucket: bucket });
```
jackyalbo, line 75: You don't really need this. try this:
```suggestion
        await this._bucket_stats.deleteOne({ system, bucket });
```

src/server/system_services/bucket_server.js
jackyalbo: I think we also exporting this function. If you already started with this cleanup can you remove the export as well?

jackyalbo (changes requested):

src/server/analytic_services/bucket_stats_store.js
BarakSason: Done

jackyalbo (approved):

nimrod-becker:
please squah

PR #6687, opened 2021-07-29
Replication BG worker- initial implementation

romayalon:

### Explain the changes
1. Added new BG worker called replication scanner which does the following:
    1.1. initializes noobaa endpoint connection.
    1.2. iterates on all replication policies (1:1 relation to bucket) and takes 1 rule that was updated the longest time ago.
    1.3. list objects of the src bucket and dst bucket
    1.4. compare objects 
    1.5. for each key found in diff objects, calls move_object_mixed_types
    1.6 update the status of the rule in db (advance the continuation token and time stmaps)
2. Added replication server and replication api which contains move_object_mixed_types function (there is another temporary function the set the noobaa endpoint connection to use - will need to find another solution for setting it from unit tests because it doesn't need to be exported.)
3. Added to replication store 2 functions: update_replication_status_by_id(), find_rules_updated_longest_time_ago().
4. Added status config to replication configuration schema.
5. Added unit tests.

1. Gap: continuation token when source bucket and target bucket are not on the same cloud provider

### Testing Instructions:
1.


src/server/system_services/replication_store.js
```diff
@@ -69,6 +70,40 @@ class ReplicationStore {
...
+
+    async find_rules_updated_longest_time_ago() {
+        dbg.log0('find_rules_updated_longest_time_ago: ');
+        // TODO: postgres client does not support $project with $min op,
+        // for doing an in db operation we will need to add support 
+        // for aggregation framework in postgres client 
+        const replications = await this._replicationconfigs.find({ deleted: null });
+        dbg.log0('find_rules_updated_longest_time_ago: ', replications);
+
+        const reduced_replications = _.map(replications, repl => ({
```
dannyzaken, line 99:
```suggestion
        const reduced_replications = replications.map(repl => ({
```

src/server/bg_services/replication_scanner.js
```diff
@@ -0,0 +1,283 @@
...
+                });
+            return res;
+        } catch (err) {
+            dbg.error('replication_server move_objects: semaphore error:', err, err.stack);
+            // no need to handle semaphore errors, eventually the object will be uploaded
+        }
+    }
+
+    async list_objects(bucket_name, prefix, continuation_token) {
+        try {
```
dannyzaken, line 179: I think that most of the log messages here should be log1. this message for example can print a long message

src/server/bg_services/replication_scanner.js
```diff
@@ -0,0 +1,283 @@
...
+            dbg.log0('replication_server list_objects: params:', bucket_name, prefix, continuation_token);
+            const list = await this.noobaa_connection.listObjectsV2({
+                Bucket: bucket_name.unwrap(),
+                Prefix: prefix,
+                ContinuationToken: continuation_token,
+                MaxKeys: Number(process.env.REPLICATION_MAX_KEYS) || 1000
+            }).promise();
+
+            dbg.log0('replication_server.list_objects: finished successfully', list.Contents.map(c => c.Key));
```
dannyzaken, line 188: I think that most of the log messages here should be log1. this message for example can print a long message. 

in general go over the `dbg.log0` and see what can be changed. also change error messages to `dbg.error`

romayalon: fixed

src/server/bg_services/replication_scanner.js
romayalon: fixed

src/server/system_services/replication_store.js
romayalon: fixed

src/server/server_rpc.js
```diff
@@ -73,6 +73,8 @@ class ServerRpc {
         let options = this.get_server_options();
         rpc.register_service(schema.account_api,
             require('./system_services/account_server'), options);
+        rpc.register_service(schema.replication_api,
```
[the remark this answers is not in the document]
romayalon: moved it

src/server/system_services/replication_store.js
```diff
@@ -69,6 +70,37 @@ class ReplicationStore {
         });
         return ans;
     }
+
+    async update_replication_status_by_id(_id, rule_id, status) {
+        dbg.log0('update_replication_status_by_id: ', _id, rule_id, status);
+        const parsed_id = db_client.instance().parse_object_id(_id);
+
+        const bulk = this._replicationconfigs.initializeOrderedBulkOp();
```
dannyzaken, line 78: why do you need a bulk update here?

dannyzaken (approved):
Approved.

see if you do have an easy alternative for the bulk update which seems a bit unnecessary. if not you can open a technical debt on it and we can do it later

src/server/system_services/replication_store.js
romayalon: as we discussed, didn't know there is a problem using bulk and conversion of queries with '.$.' was only exist in bulk so I used it, anyway I moved the conversion of .$. queries to a function in postgres client and called it also from updateOne (for non bulk)

romayalon:
@dannyzaken 
please take a look at postgres_client.js, I added there handling of .$. for non bulk ops.

dannyzaken (approved):

PR #6689, opened 2021-07-29
add ignore_name_already_exist property to check_external_connection api

kfir-payne:

related to https://github.com/noobaa/noobaa-operator/pull/688

### Explain the changes
1. Added ignore_name_already_exist property to check_external_connection api so we could check an external connection before updating it.

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -439,7 +439,7 @@ module.exports = {
             method: 'GET',
             params: {
                 type: 'object',
-                required: ['identity', 'secret', 'endpoint'],
+                required: ['identity', 'secret', 'endpoint', 'endpoint_type'],
                 properties: {
                     name: {
                         type: 'string'
@@ -457,6 +457,9 @@ module.exports = {
                     },
                     endpoint_type: {
                         $ref: 'common_api#/definitions/endpoint_type'
+                    },
+                    ignore_name_already_exist: {
+                        type: 'boolean'
                     }
                 }
             },
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -846,7 +846,7 @@ async function update_external_connection(req) {
 
     let check_failed = false;
     try {
-        const { status } = await _check_external_connection({
+        const { status } = await _check_external_connection_internal({
             name,
             identity,
             secret,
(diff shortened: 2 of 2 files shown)


src/api/account_api.js
```diff
@@ -439,7 +439,7 @@ module.exports = {
             method: 'GET',
             params: {
                 type: 'object',
-                required: ['identity', 'secret', 'endpoint'],
+                required: ['identity', 'secret', 'endpoint', 'endpoint_type'],
```
liranmauda, line 442: Why we made it a requirement?
I can see that all places in the core are calling it with this field, but have we checked `CheckExternalConnectionAPI` in the operator?

src/api/account_api.js
```diff
@@ -457,6 +457,9 @@ module.exports = {
                     },
                     endpoint_type: {
                         $ref: 'common_api#/definitions/endpoint_type'
+                    },
+                    ignore_name_already_exist: {
+                        type: 'boolean'
```
liranmauda, line 462: Are we calling it?

liranmauda (commented):
line 927:
the function `check_external_connection` Are we using it? 
and if we do why do we have also an internal one `_check_external_connection`?
If we actually using `check_external_connection`, can you async/await it?

src/api/account_api.js
kfir-payne: In the comment above, I added a link to a PR in the operator that will use that field.

src/api/account_api.js
kfir-payne: I added `endpoint_type` as a required field because if you call this RPC function without it, it will fail.
yes, I checked the operator and verified all the calls there would have `endpoint_type`.

kfir-payne:
@liranmauda we do use `check_external_connection` in the operator, the internal one is to check a connection before an update, I can async/await it.

src/api/account_api.js
liranmauda: 👍

src/api/account_api.js
liranmauda: 👍

liranmauda:
> @liranmauda we do use `check_external_connection` in the operator, the internal one is to check a connection before an update, I can async/await it.

For me, it is weird to have 2 functions that do the same with an extension, Let's try to avoid the code duplication and use the internal function inside the external. 
We should probably change the internal function name (maybe add `_internal` same as `_check_azure_connection_internal`)

kfir-payne:
@liranmauda agree, good point, I'll  try to see what we can reuse there.

dannyzaken (approved):

PR #6699, opened 2021-08-08
Support followSymlinks on top of NS FS

igorpick:
Signed-off-by: Igor Pick <ipick@redhat.com>

### Requirements

1. Resolve symbolic links in a target entry path.
2. if the target entry path is outside of a bucket, returns an AccessDenied error.

Implemented in the following methods:
1. read object metadata - return ACCESS ERROR if an object path is outside of the bucket.
2. read object stream -  return ACCESS ERROR if an object path is outside of the bucket.
3. upload object - return ACCESS ERROR if an object destination/source path is outside of the bucket.
4. list multiparts -  return ACCESS ERROR if provided mpu_path is outside of the bucket.
5. complete object upload - return ACCESS ERROR if a multipart destination path is outside of the bucket.
6. Delete object - return ACCESS ERROR if an object path is outside of the bucket.
7. Delete multiple objects - return ACCESS ERROR if one of objects' paths is outside of the bucket.
8. List object - if bucket entry is a symbolic link and outside of the bucket, returns info of the symbolic link himself, otherwise, returns info of the target entry.

Note: Related issue:
https://bugzilla.redhat.com/show_bug.cgi?id=1429472

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -19,6 +19,7 @@
 #include <uv.h>
 #include <sys/xattr.h>
 #include <vector>
+#include <stdlib.h>
 
 #ifdef __APPLE__
     #include <sys/param.h>
@@ -1136,6 +1137,34 @@ struct FileFsync : public FSWrapWorker<FileWrap>
     }
 };
 
+struct RealPath : public FSWorker
+{
+    std::string _path;
+    char* _full_path;
+
+    RealPath(const Napi::CallbackInfo& info)
+        : FSWorker(info)
+    {
+        _path = info[1].As<Napi::String>();
+        Begin(XSTR() << DVAL(_path));
+    }
+
+    virtual void Work()
+    {
+        _full_path = realpath(_path.c_str(), NULL);
+        if (_full_path == NULL) SetSyscallError();
+    }
+
+    virtual void OnOK()
+    {
+        DBG1("FS::RealPath::OnOK: " << DVAL(_path) << DVAL(_full_path));
+        
+        Napi::Env env = Env();
+        auto res = Napi::String::New(env, _full_path);
+        _deferred.Resolve(res);
+    }
+};
+
 Napi::Value
 FileWrap::close(const Napi::CallbackInfo& info)
 {
@@ -1362,6 +1391,7 @@ fs_napi(Napi::Env env, Napi::Object exports)
     exports_fs["link"] = Napi::Function::New(env, api<Link>);
     exports_fs["linkat"] = Napi::Function::New(env, api<Linkat>);
     exports_fs["fsync"] = Napi::Function::New(env, api<Fsync>);
+    exports_fs["realpath"] = Napi::Function::New(env, api<RealPath>);
 
     FileWrap::init(env);
     exports_fs["open"] = Napi::Function::New(env, api<FileOpen>);
@@ -1371,7 +1401,9 @@ fs_napi(Napi::Env env, Napi::Object exports)
 
     exports_fs["S_IFMT"] = Napi::Number::New(env, S_IFMT);
     exports_fs["S_IFDIR"] = Napi::Number::New(env, S_IFDIR);
+    exports_fs["S_IFLNK"] = Napi::Number::New(env, S_IFLNK);
     exports_fs["DT_DIR"] = Napi::Number::New(env, DT_DIR);
+    exports_fs["DT_LNK"] = Napi::Number::New(env, DT_LNK);
 
     exports_fs["set_debug_level"] = Napi::Function::New(env, set_debug_level);
 
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -59,6 +59,41 @@ function isDirectory(ent) {
     }
 }
 
+/**
+ * 
+ * @param {*} stat - entity stat yo check
+ * @param {*} fs_account_config - account config using to check symbolic links
+ * @param {*} entry_path - path of symbolic link
+ * @returns 
+ */
+async function is_directory_or_symlink_to_directory(stat, fs_account_config, entry_path) {
+    try {
+        let r = isDirectory(stat);
+        if (!r && is_symbolic_link(stat)) {
+            let targetStat = await nb_native().fs.stat(fs_account_config, entry_path);
+            if (!targetStat) throw new Error('is_directory_or_symlink_to_directory: targetStat is empty');
+            r = isDirectory(targetStat);
+        }
+        return r;
+    } catch (err) {
+        if (err.code !== 'ENOENT') {
+            throw err;
+        }
+    }
+}
+
+function is_symbolic_link(stat) {
+    if (!stat) throw new Error('isSymbolicLink: stat is empty');
+    if (stat.mode) {
+        // eslint-disable-next-line no-bitwise
+        return (((stat.mode) & nb_native().fs.S_IFMT) === nb_native().fs.S_IFLNK);
+    } else if (stat.type) {
+        return stat.type === nb_native().fs.DT_LNK;
+    } else {
+        throw new Error(`isSymbolicLink: stat ${stat} is not supported`);
+    }
+}
+
 function get_umasked_mode(mode) {
     // eslint-disable-next-line no-bitwise
     return mode & ~config.NSFS_UMASK;
@@ -309,9 +344,11 @@ class NamespaceFS {
                         return;
                     }
 
+                    const isDir = await is_directory_or_symlink_to_directory(ent, fs_account_config, path.join(dir_path, ent.name));
+
                     const r = {
-                        key: this._get_entry_key(dir_key, ent),
-                        common_prefix: isDirectory(ent),
+                        key: this._get_entry_key(dir_key, ent, isDir),
+                        common_prefix: isDir,
                     };
 
                     let pos;
@@ -367,8 +404,12 @@ class NamespaceFS {
                         const prev_dir = sorted_entries[marker_index - 1];
                         const prev_dir_name = path.join(prev_dir.name, '/');
                         if (marker_curr.startsWith(prev_dir_name) && dir_key !== prev_dir.name) {
-                            if (!delimiter && isDirectory(prev_dir)) {
-                                await process_dir(path.join(dir_key, prev_dir_name));
+                            if (!delimiter) {
+                                const isDir = await is_directory_or_symlink_to_directory(
+                                    prev_dir, fs_account_config, path.join(dir_path, prev_dir_name));
+                                if (isDir) {
+                                    await process_dir(path.join(dir_key, prev_dir_name));
+                                }
                             }
                         }
                     }
@@ -419,7 +460,12 @@ class NamespaceFS {
             await Promise.all(results.map(async r => {
                 if (r.common_prefix) return;
                 const entry_path = path.join(this.bucket_path, r.key);
-                r.stat = await nb_native().fs.stat(fs_account_config, entry_path);
+                //If entry is outside of bucket, returns stat of symbolic link
+                if (await this._is_path_in_bucket_boundaries(fs_account_config, entry_path)) {
+                    r.stat = await nb_native().fs.stat(fs_account_config, entry_path);
+                } else {
+                    r.stat = await nb_native().fs.lstat(fs_account_config, entry_path);
+                }
                 // This is done to get the MD5 for the response, should find a more efficient way
                 // r.fs_xattr = config.NSFS_CALCULATE_MD5 ? await this._get_fs_xattr_from_path(fs_account_config, entry_path) : undefined;
             }));
@@ -458,8 +504,9 @@ class NamespaceFS {
         let file;
         const fs_account_config = this.set_cur_fs_account_config(object_sdk);
         try {
-            await this._load_bucket(params, fs_account_config);
             const file_path = this._get_file_path(params);
+            await this._check_path_in_bucket_boundaries(fs_account_config, file_path);
+            await this._load_bucket(params, fs_account_config);
             file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
             const stat = await file.stat(fs_account_config);
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
@@ -500,6 +547,7 @@ class NamespaceFS {
         try {
             await this._load_bucket(params, fs_account_config);
             const file_path = this._get_file_path(params);
+            await this._check_path_in_bucket_boundaries(fs_account_config, file_path);
             await this._fail_if_archived_or_sparse_file(fs_account_config, file_path);
             file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
 
@@ -620,12 +668,14 @@ class NamespaceFS {
         let fs_xattr = to_fs_xattr(params.xattr);
         // dbg.log0('NamespaceFS.upload_object:', upload_path, '->', file_path);
         try {
+            await this._check_path_in_bucket_boundaries(fs_account_config, file_path);
             await Promise.all([
                 this._make_path_dirs(file_path, fs_account_config),
                 this._make_path_dirs(upload_path, fs_account_config)
             ]);
             if (params.copy_source) {
                 const source_file_path = path.join(this.bucket_path, params.copy_source.key);
+                await this._check_path_in_bucket_boundaries(fs_account_config, source_file_path);
                 await this._fail_if_archived_or_sparse_file(fs_account_config, source_file_path);
                 try {
                     const same_inode = await this._is_same_inode(fs_account_config, source_file_path, file_path, fs_xattr);
@@ -690,6 +740,7 @@ class NamespaceFS {
             target_file = await nb_native().fs.open(fs_account_config, upload_path, 'w');
             //Reading the source_file and writing into the target_file
             let read_pos = 0;
+
             for (;;) {
                 const { buffer, callback } = await buffers_pool.get_buffer();
                 buffer_pool_cleanup = callback;
@@ -844,6 +895,7 @@ class NamespaceFS {
         try {
             const fs_account_config = this.set_cur_fs_account_config(object_sdk);
             await this._load_multipart(params, fs_account_config);
+            await this._check_path_in_bucket_boundaries(fs_account_config, params.mpu_path);
             const entries = await nb_native().fs.readdir(fs_account_config, params.mpu_path);
             const multiparts = await Promise.all(
                 entries
@@ -883,6 +935,7 @@ class NamespaceFS {
             multiparts.sort((a, b) => a.num - b.num);
             await this._load_multipart(params, fs_account_config);
             const file_path = this._get_file_path(params);
+            await this._check_path_in_bucket_boundaries(fs_account_config, file_path);
             const upload_path = path.join(params.mpu_path, 'final');
             write_file = await nb_native().fs.open(fs_account_config, upload_path, 'w', get_umasked_mode(config.BASE_MODE_FILE));
             for (const { num, etag } of multiparts) {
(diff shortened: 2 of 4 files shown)


src/deploy/NVA_build/noobaa_init.sh
```diff
@@ -178,7 +178,7 @@ init_endpoint() {
   fix_non_root_user
 
   cd /root/node_modules/noobaa-core/
-  run_internal_process node ./src/s3/s3rver_starter.js
+  run_internal_process node ${ENDPOINT_NODE_OPTIONS} ./src/s3/s3rver_starter.js
```
liranmauda, line 181: Please make a comment or on how can it be used

liranmauda (dismissed):
Some general points:
1. I see that in `_setup_single_namespace` we didn't add the `follow_symlinks` is that intentioned to always be true there? 
2. `fstat` is acting like stat, only with `fd` instead of `pathname`. Do we need to handle it also? Is open following links?

src/deploy/NVA_build/noobaa_init.sh
igorpick: Done. Added to the PR comment

igorpick:
@liranmauda 
About your questions:
1. please take a look on object_sdk.js line 339. it's in _setup_single_namespace method
2. According to documentations, 'fstat' is identical to 'stat', except that the file about which information is to be retrieved is specified by a file descriptor. Since all our code uses pathname, why do you think we have to switch to fd?

src/deploy/NVA_build/noobaa_init.sh
igorpick: Added comment to the file as well

liranmauda: Greate, Just 2 notes: 
1. change the `k edit...` to `kubectl edit...`
2. We can have a CLI flag for that, WDYT?

igorpick: 1. Done
2. We can do that, but since it's not a small diff, in my opinion, it should be additional task and PR.

liranmauda: > 2. We can do that, but since it's not a small diff, in my opinion, it should be additional task and PR.

of course, as it is on the operator side...

liranmauda:
> 1. please take a look on object_sdk.js line 339. it's in _setup_single_namespace method

True, I missed that it is inside that function, but I meant that the call for `new NamespaceFS` in line 334 was missing the `follow_symlinks`. I see that it is already fixed.

> 2. According to documentations, 'fstat' is identical to 'stat', except that the file about which information is to be retrieved is specified by a file descriptor. Since all our code uses pathname, why do you think we have to switch to fd?

That is not true. 
We have a lot of places that we use `fstat` for example:
https://github.com/noobaa/noobaa-core/blob/c30d0e17a95d0501af921919ca74d57be90c7c99/src/sdk/namespace_fs.js#L438

you can see that the stat is calling from `fd` that is called file, also you can see that there is no `pathname` in that call

places that are using actual stat looks like this: 
https://github.com/noobaa/noobaa-core/blob/c30d0e17a95d0501af921919ca74d57be90c7c99/src/sdk/namespace_fs.js#L106

the are being called with `nb_native().fs.` and they have `pathname`

igorpick:
You have no option to not follow symlinks while using fd

src/native/fs/fs_napi.cpp
```diff
@@ -244,15 +244,21 @@ struct Stat : public FSWorker
...
         _path = info[1].As<Napi::String>();
         Begin(XSTR() << "Stat " << DVAL(_path));
+
+        Napi::Object config = info[0].As<Napi::Object>();
+        if (config.Has("follow_symlinks")) _follow_symlinks = config.Get("follow_symlinks").ToBoolean();
     }
     virtual void Work()
     {
-        int r = stat(_path.c_str(), &_stat_res);
+        int r = _follow_symlinks ? stat(_path.c_str(), &_stat_res) : lstat(_path.c_str(), &_stat_res);
```
liranmauda, line 261: Why do we want to do it in the native? 
We have stat and `lstat` as ops, looks weird to call a `stat` op and get `lstat` 
We can do it on `namespace_fs` instead of here

liranmauda:
> > you can see that the stat is calling from `fd` that is called file, also you can see that there is no `pathname` in that call
> 
> You have no option not following symlinks while using fd. But I am looking now for how to get file descriptor of symbolic link instead of the target file

We can just call `lstat` when we need instead of `fstat`

igorpick:
> > > you can see that the stat is calling from `fd` that is called file, also you can see that there is no `pathname` in that call
> > 
> > 
> > You have no option not following symlinks while using fd. But I am looking now for how to get file descriptor of symbolic link instead of the target file
> 
> We can just call `lstat` when we need instead of `fstat`

NO, you cannot do that. lstat gets pathname and fstat gets fd

src/native/fs/fs_napi.cpp
igorpick: Then you have to modify all the stat calls in namespace_fs. now, it's exactly like in os. when you run stat <filename>, actually os runs lstat(2). just when you run stat -L, os runs stat(2)

liranmauda:
> No, you cannot do that. lstat gets pathname and fstat gets fd

So?
pseudo code:
```
if you want to follow {
  stat = await file.stat(fs_account_config); 
} else {
  stat = await nb_native().fs.stat(fs_account_config, dir_path); 
}
```
You have the pathname when you open the file to get the `fd`

igorpick:
> dir_path

In my opinion, it should be in native. See my comment above

src/native/fs/fs_napi.cpp
liranmauda: As far as I can tell, ` stat -L` is not running `lstat` op. 
Please run on a system `strace stat -L` and see if it call the `lstat` op. 

I don't see a reason why we should not move the logic into the `namespace_fs` and keep it mixed in the native (mixed as a single op is calling 2 different calls.)
@guymguym WDYT?

liranmauda:
> > dir_path

pseud code, Any kind of path...

src/native/fs/fs_napi.cpp
igorpick: But `stat` does run `lstat`

jeniawhite (dismissed):
Just change napi and nsfs to call different stats, also please make sure to change FileWrap stat to lstat.
I.E when with followLinks do not use fstat of the wrapper and just call lstat. This will not work on FD and work on the path and be less efficient, but not sure if we have another option.

guymguym (dismissed):
@igorpick Take a look at “Treatment of symbolic links in system calls” here -
https://man7.org/linux/man-pages/man7/symlink.7.html

Few points:
1. What about open calls from fs.open and fs.checkAccess? We need to use O_NOFOLLOW and this still only handles basename.
2. Using lstat and O_NOFOLLOW will only handle the basename component of the path, what about the dirname component?? I don’t think there’s a way around just adding an explicit checking logic of the dirname components in our nsfs layer.
3. What about mkdir in _make_path_dirs in upload_object - it calls mkdir without checking anything on the paths.
4. What about link (source_file_path) in upload_object - it would follow links on the source.
5. All the calls to stat inside the temporary directory used for temporary files for upload and more meta files - don’t really need to use lstat.
6. Changing the dir_cache behavior based on bucket config feels very fishy.

So I would revert from replacing all stat calls in the namespace_fs level, which looks inconvenient both from readability and code maint perspective when we want to add more stat calls and need to notice it.

Instead I would start by checking the follow symlinks case in namespace_fs check_access() js function so that if the config says to avoid following symlinks, then it will also loop through the dir components of the path and check access to all of those (and use O_NOFOLLOW flag). Now since we only call this function today in list_objects, we will need to start calling it from object ops before allowing it to process any input as a path. We should probably refactor this function a bit if we want to use it for files, and also on upload where we want to allow non-existing paths... 

Next I would add the O_NOFOLLOW flag to our open calls and make sure to translate this to a correct error / reply. Returning the content of the link itself doesn't necessarily make the most sense.

Finally, I think that once we will check explicitly in all ops that any input paths are ok to follow (both dirs and files paths and upload file paths and source file paths...) and use the O_NOFOLLOW flag in opens, most of the stat calls in the code do not need to be replaced. For those files in our temp dir it doesn't really make all that more sense to use lstat. Next, the calls in dir_cache should not use lstat at all because we don’t want the cache reply to depend on the caller configuration, this is also why @romayalon had to add the check access call explicitly before using the cache and not embed the access check into the cache. The only case I see that we need to explicitly change a stat call is in list_objects where we call to stat on entry_path - does lstat make more sense here? not sure that returning the link attributes is correct, in any case it should be the same as the behavior of read_object_md for that entry (either the link stats, or zeroed stats, or omit from list and error on read...). I guess the link attributes are ok.

I hope my comments are clear enough and let me know what you think.
Cheers

dannyzaken (dismissed):

dannyzaken:
After a discussion with @nimrod-becker and @guymguym we suggest the following:
1. remove the `follow_symlinks` configuration. this should not be configurable for now
2. for `List` operations, if we encounter a symlink we should show the link target only if it is inside the bucket (under the path of the bucket). otherwise, we should show the link itself in the list (or filter it out?)
3. for `Get` operations, if the target is inside the bucket we return the content of the target (or NoSuchKey if it's a dir). if the target is outside of the bucket we should return AccessDenied

dannyzaken:
just to clarify regarding List Objects operations. when listing a symbolic link 
if the target is within the bucket - return the **target's** metadata (size/last modified/etc.)
otherwise, return the link and its own MD

@guymguym @nimrod-becker WDYT?

src/api/bucket_api.js
```diff
@@ -1251,7 +1251,7 @@ module.exports = {
                         },
                         should_create_underlying_storage: {
                             type: 'boolean'
-                        },
+                        }
```
dannyzaken, line 1254: remove this change

src/sdk/namespace_fs.js
```diff
@@ -144,6 +179,7 @@ class NamespaceFS {
      *  fs_backend?: string;
      *  bucket_id: string;
      *  namespace_resource_id?: string;
+     *  follow_symlinks?: boolean
```
dannyzaken, line 182: remove

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -426,11 +413,152 @@ mocha.describe('namespace_fs', function() {
...
+        tmp_fs_path = '/private/' + tmp_fs_path;
+    }
+    const bucket = 'bucket1';
+    const bucket_full_path = tmp_fs_path + '/' + bucket;
+    const expected_dirs = ['d1', 'd2', 'd3/d3d1'];
+    const expected_files = ['f1', 'f2', 'f3', 'd2/f4', 'd2/f5', 'd3/d3d1/f6'];
+    const expected_links = [{ t: 'f1', n: 'lf1' }, { t: '/etc', n: 'ld2' }];
+    const dummy_object_sdk = { requesting_account: { nsfs_account_config: { uid: process.getuid(), gid: process.getgid() } } };
+
+    const ns = new NamespaceFS({ bucket_path: bucket_full_path, bucket_id: '1', namespace_resource_id: undefined, follow_symlinks: false });
```
dannyzaken, line 429: remove `follow_symlinks: false`

src/sdk/namespace_fs.js
```diff
@@ -59,6 +59,41 @@ function isDirectory(ent) {
...
 }
 
+/**
+ * 
+ * @param {*} stat - entity stat yo check
+ * @param {*} fs_account_config - account config using to check symbolic links
+ * @param {*} entry_path - path of symbolic link
+ * @returns 
+ */
+async function isDirectoryOrSymLinkToDirectory(stat, fs_account_config, entry_path) {
```
dannyzaken, line 69: in general change the signatures of new functions to snake case to be more consistent with the rest of the code
```suggestion
async function is_directory_or_symlink_to_directory(stat, fs_account_config, entry_path) {
```

src/sdk/namespace_fs.js
```diff
@@ -59,6 +59,41 @@ function isDirectory(ent) {
...
+    try {
+        let r = isDirectory(stat);
+        if (!r && isSymbolicLink(stat)) {
+            let targetStat = await nb_native().fs.stat(fs_account_config, entry_path);
```
dannyzaken, line 73: I'm not sure I understand the logic here. why is that the targetStat? don't you need to check `realpath`?

igorpick: No, if the stat is a symbolic link, then the method tries to get a stat of the target file and check if the target file is a directory. fs.stat() returns a stat of the lowest target, even if there are a few symbolic links on the way (symlink points to symlink and so on)

src/server/system_services/bucket_server.js
```diff
@@ -570,7 +570,6 @@ function get_bucket_changes(req, update_request, bucket, tiering_policy) {
     if (update_request.versioning) {
         get_bucket_changes_versioning(req, bucket, update_request, single_bucket_update, changes);
     }
-
```
dannyzaken, line 573: remove change

src/sdk/object_sdk.js
```diff
@@ -276,7 +276,7 @@ class ObjectSDK {
                 write_resource: wr,
                 read_resources: _.map(rr, ns_info => (
                     ns_info.endpoint_type === 'MULTIPART' ? ns_info.ns :
-                    this._setup_single_namespace(_.extend({}, ns_info))
+                        this._setup_single_namespace(_.extend({}, ns_info))
```
dannyzaken, line 279: remove change

src/util/fs_utils.js
```diff
@@ -159,6 +159,14 @@ function create_path(dir, mode) {
     return fs.promises.mkdir(dir, { mode, recursive: true });
 }
 
+function create_file(file_path) {
+    return fs.promises.open(file_path, 'w');
```
dannyzaken, line 163: this will leave an open file handle

src/util/fs_utils.js
```diff
@@ -159,6 +159,14 @@ function create_path(dir, mode) {
...
+}
+
+function create_symbolic_link(target, link_path) {
```
dannyzaken, line 166: I think that the added utils here are redundant. it is a one-liner and only used in the test. 
if a function is necessary (maybe for `create_file` where you also need to close the file) you can make a local func in the test

src/sdk/namespace_fs.js
```diff
@@ -844,23 +895,24 @@ class NamespaceFS {
...
-                    const fs_xattr =
-                        config.NSFS_CALCULATE_MD5 ? await this._get_fs_xattr_from_path(fs_account_config, part_path) : undefined;
-                    return {
-                        num,
-                        size: stat.size,
-                        etag: this._get_etag(stat, fs_xattr),
-                        last_modified: new Date(stat.mtime),
-                    };
-                })
+                    .filter(e => e.name.startsWith('part-'))
```
dannyzaken, line 902: I think these whitespaces are redundant

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -426,11 +413,152 @@ mocha.describe('namespace_fs', function() {
...
+    const dummy_object_sdk = { requesting_account: { nsfs_account_config: { uid: process.getuid(), gid: process.getgid() } } };
+
+    const ns = new NamespaceFS({ bucket_path: bucket_full_path, bucket_id: '1', namespace_resource_id: undefined });
+
+    mocha.before(async () => {
+        await fs_utils.create_fresh_path(`${bucket_full_path}`);
+        await P.all(_.map(expected_dirs, async dir =>
+            fs_utils.create_fresh_path(`${bucket_full_path}/${dir}`)));
+        await P.all(_.map(expected_files, async file =>
+            fs_utils.create_file(`${bucket_full_path}/${file}`)));
```
dannyzaken, line 436: you can use `fs.promises.appendFile` to create a file

dannyzaken, line 436: you can use `fs.promises.appendFile` to create a file

src/sdk/namespace_fs.js
igorpick: Done

src/util/fs_utils.js
igorpick: Used apendFile instead...

src/util/fs_utils.js
igorpick: Done. Moved to the test.

src/test/unit_tests/test_namespace_fs.js
igorpick: Done

src/sdk/namespace_fs.js
```diff
@@ -1302,6 +1362,70 @@ class NamespaceFS {
         }
     }
 
+    /**
+     * Returns true if the entry path is symbolic link
+     * @param {*} fs_account_config 
+     * @param {*} entry_path 
+     * @returns 
+     */
+    async _is_symbolic_link_by_path(fs_account_config, entry_path) {
```
dannyzaken, line 1371: I don't see this function being used anywhere. am I missing it?

igorpick: Removed it

src/sdk/namespace_fs.js
```diff
@@ -309,9 +344,11 @@ class NamespaceFS {
                         return;
                     }
 
+                    const common_prefix = await is_directory_or_symlink_to_directory(ent, fs_account_config, path.join(dir_path, ent.name));
```
jackyalbo, line 347: I would change it to isDir like you do in other places - sounds less confusing

src/sdk/namespace_fs.js
```diff
@@ -1302,6 +1362,49 @@ class NamespaceFS {
...
+     * @returns 
+     */
+    async _is_path_in_bucket_boundaries(fs_account_config, entry_path) {
+        dbg.log0('check_bucket_boundaries: fs_account_config', fs_account_config, 'file_path', entry_path);
+        if (!entry_path.startsWith(this.bucket_path)) {
+            dbg.log0('check_bucket_boundaries: the path', entry_path, 'is not in the bucket', this.bucket_path, 'boundaries');
+                return false;
+        }
+        try {
+            // Returns tyhe real path of the entry.
```
jackyalbo, line 1378:
```suggestion
            // Returns the real path of the entry.
```

src/sdk/namespace_fs.js
```diff
@@ -1302,6 +1362,49 @@ class NamespaceFS {
...
+                return false;
+            }
+        } catch (err) {
+            // Error: No such file or directory
+            // In the upload use case, the destination file desn't exist yet, need to validate the parent dirs path.
+            if (err.code === 'ENOENT') {
+                return this._is_path_in_bucket_boundaries(fs_account_config, path.dirname(entry_path));
+            }
+            dbg.log0('check_bucket_boundaries error', err.code, entry_path, err);
+            return false;
```
jackyalbo, line 1392: Do we want to hide the error? it seems like in some cases we will return EACCESS when the fs error is different.

src/sdk/namespace_fs.js
jackyalbo: I meant: 
```suggestion
                    const isDir = await is_directory_or_symlink_to_directory(ent, fs_account_config, path.join(dir_path, ent.name));
```

src/sdk/namespace_fs.js
igorpick: The error log is for unusual errors, like "I/O", "Too many symbolic links" and so on.
So, I think we should know about that.

src/sdk/namespace_fs.js
```diff
@@ -392,7 +433,7 @@ class NamespaceFS {
                 try {
                     dbg.warn('NamespaceFS: open dir streaming', dir_path, 'size', cached_dir.stat.size);
                     dir_handle = await nb_native().fs.opendir(fs_account_config, dir_path); //, { bufferSize: 128 });
-                    for (;;) {
+                    for (; ;) {
```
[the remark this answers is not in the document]
igorpick: Done

src/sdk/namespace_fs.js
igorpick: Done

src/sdk/namespace_fs.js
igorpick: Done

src/sdk/namespace_fs.js
jackyalbo: Just asking if the user shouldn't be aware of those as you only log them but not sending them back to the user - he will get just access denied. Again, just a question - not sure if that's good or not...

igorpick: I think no, but @dannyzaken wdyt?

igorpick: Done

dannyzaken (approved):

jackyalbo (approved):

PR #6700, opened 2021-08-09
Addition of FSync

jeniawhite:
### Explain the changes
1. Added FSync for I/O flows

1. Multipart upload as temp files that are renamed
2. FSync on reads for dual access when needed
3. Marking of FSynced files in the FS

### Testing Instructions:
1.

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -612,6 +612,7 @@ class NamespaceFS {
             // TODO use file xattr to store md5_b64 xattr, etc.
             const stat = await nb_native().fs.stat(fs_account_config, upload_path);
             await nb_native().fs.rename(fs_account_config, upload_path, file_path);
+            await nb_native().fs.fsync(fs_account_config, path.dirname(upload_path));
             return { etag: this._get_etag(stat, fs_xattr) };
         } catch (err) {
             this.run_update_issues_report(object_sdk, err);
@@ -676,6 +677,7 @@ class NamespaceFS {
             //Closing the source_file and target files
             await source_file.close(fs_account_config);
             source_file = null;
+            await target_file.fsync(fs_account_config);
             await target_file.close(fs_account_config);
             target_file = null;
             // Used for etag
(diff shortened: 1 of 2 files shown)


liranmauda (approved):
Just one question:
Do you want to keep the close of the dirs in the finally?
Why not use the same logic that we use on files and close it on the try block +  make sure with null that dir (but keep the finally just in case we error on the way)?

BarakSason (commented):
Personally, I'd appreciate some more info in the commit message (why is this needed? High-level explanation on how is this implemented?).
Some in-code documentation would be appreciated as well.

guymguym (changes requested):
Since our "update directory" code paths never really have the directory open with opendir (beccause it's only useful for reading entries one-by-one), then I think it makes more sense for fsync to be implemented once as an instance method for FileWrap - which you did, and second as a top api function (like rename) that takes a file/dir path and does open+fsync+close (instead of the DirWrap function). This will also simplify the error handling code needed from js. WDYT?

jeniawhite:
@liranmauda Removed the finally scope code due to usage of a function instead of opening a wrapper.
@BarakSason Please elaborate more on the information that is missing, you can read more at https://man7.org/linux/man-pages/man2/fdatasync.2.html to understand the high-level design, why it is used, and what the method does.

liranmauda (approved):
LGTM

guymguym (approved):

src/sdk/namespace_fs.js
```diff
@@ -612,6 +612,7 @@ class NamespaceFS {
             // TODO use file xattr to store md5_b64 xattr, etc.
             const stat = await nb_native().fs.stat(fs_account_config, upload_path);
             await nb_native().fs.rename(fs_account_config, upload_path, file_path);
+            await nb_native().fs.fsync(fs_account_config, path.dirname(upload_path));
```
guymguym, line 615: @jeniawhite Wait, isn't this the wrong path? we should be calling fsync on the target dir of the rename, which would be `path.dirname(file_path)` in this scope

guymguym: There's a link() call few lines above this line, that I think needs fsync too - and there it's indeed on the dirname of upload_path. Generally it seems that we want to follow any update to the filesystem with an fsync call to the directory of the updated file... that includes all these natiive fs calls - rename, link, unlink, mkdir, rmdir, writeFile, open('w'). We also said we want to have a control flag to turn off fsync.

PR #6718, opened 2021-08-25
Exposing odf_health_status metrics

liranmauda:
### Explain the changes
Exposing odf_health_status metrics


liranmauda:
@nimrod-becker @romayalon We can change the logic and decide that of more than 0.3 We can error and not warning.

nimrod-becker (approved):

Issue #6707, opened 2021-08-13
Endpoint pod crashed with OOM and RC=137 with IO


akmithal:
### Environment info
- NooBaa Version: master-20210810
- Platform: OCP 4.7

### Actual behavior
1. Endpoint gets OOM and crashed

### Expected behavior
1. Endpoint should be handle IO and not get OOM

### Steps to reproduce

Resources set as 2 CPU, 3 GB RAM for endpoint pod
 ```
resources:
          limits:
            cpu: "2"
            memory: 3Gi
          requests:
            cpu: "2"
            memory: 3Gi
```

1. Started download of 1 million objects and listing them together from another session
2. Got OOMKilled for endpoint pod in few minutes and IO failed

`download failed: s3://bucket-1m/obj_109122 to bucket-1m/obj_109122 An error occurred (503) when calling the GetObject operation: Service Unavailable`

```
[root@ocp-akshat-1-inf ~]# podn
NAME                                               READY   STATUS        RESTARTS   AGE
noobaa-core-0                                      1/1     Running       0          2d
noobaa-db-pg-0                                     1/1     Running       0          2d
noobaa-default-backing-store-noobaa-pod-e1b8222a   0/1     Terminating   0          41h
noobaa-endpoint-75db6b8d47-v8trg                   0/1     OOMKilled     0          2d
noobaa-operator-569ffd7c9c-t9kll                   1/1     Running       10         2d
```

```
Aug-13 5:39:19.835 [Endpoint/13]    [L0] core.sdk.namespace_fs:: check_access: dir_path /nsfs/nsfs-nsr-1/bucket-1m fs_account_config { gid: 800, uid: 5000, nsfs_only: false, new_buckets_path: '/', backend: 'GPFS', warn_threshold_ms: 100 }
Aug-13 5:39:20.045 [Endpoint/13]    [L0] core.endpoint.s3.ops.s3_get_object:: request aborted: undefined
Aug-13 5:39:20.046 [Endpoint/13]    [L0] core.endpoint.s3.ops.s3_get_object:: request aborted: undefined
[...]

PR #6726, opened 2021-09-12
Semaphore improvement and addition to upload flows

jeniawhite:
### Explain the changes
1. Accounted PUT Object / Upload Part in the global semaphore.
Gap: The current implementation takes up a buffer of 8 megabytes for every PUT, Part Upload even if they are small. 
2. Changes to semaphore cleanup logic on non settled working items

1. https://github.com/noobaa/noobaa-core/issues/6707

### Testing Instructions:
1.

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -22,7 +22,7 @@ const nb_native = require('../util/nb_native');
 
 const buffers_pool_sem = new Semaphore(config.NSFS_BUF_POOL_MEM_LIMIT, {
     timeout: config.IO_STREAM_SEMAPHORE_TIMEOUT,
-    timeout_error_code: 'IO_STREAM_ITEM_TIMEOUT'
+    timeout_error_code: 'IO_STREAM_ITEM_TIMEOUT',
 });
 const buffers_pool = new buffer_utils.BuffersPool(config.NSFS_BUF_SIZE, buffers_pool_sem);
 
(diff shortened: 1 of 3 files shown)


src/sdk/namespace_fs.js
```diff
@@ -708,6 +711,10 @@ class NamespaceFS {
         }
     }
 
+    // Allocated config.NSFS_BUF_SIZE in Semaphore but in fact we can take up more inside
+    // This is due to MD5 calculation and data buffers
+    // Can be finetuned further on if needed and inserting the Semaphore logic inside
+    // Instead of wrapping the whole _upload_stream function (q_buffers lives outside of data scope of the stream)
```
liranmauda, line 717:
```suggestion
    // Allocated config.NSFS_BUF_SIZE in Semaphore but in fact, we can take up more inside
    // This is due to MD5 calculation and data buffers
    // Can be finetuned further on if needed and inserting the Semaphore logic inside
    // Instead of wrapping the whole _upload_stream function (q_buffers lives outside of the data scope of the stream)
```

liranmauda (approved):
LGTM
Just one question, Are we sure we want to add the `_work_cap` into `surround` as well? 
As we are not calling it with count we will basically get a default 2 min work timeout (as the name suggesting 🙂  )

jeniawhite:
@liranmauda 
Yes, we still need the same logic for surround as well since it just takes up 1 instead of a count that you gave it.
Even in that case we can have a Semaphore of 1 and take it all up and not release on non-settled promises.

PR #6747, opened 2021-10-06
remove "distinct on" query and flow

igorpick:
### Explain the changes
1.  During the last postgres performance improvement, detected that the "distinct on" query is very expensive and may take 30 -40 sec and a lot of cpu in scale cluster. 
Additionally, the responded data is not in use due to a bug, since the 
```
if (undeletable_bucket_set.has(bucket_name)) {
                    b.undeletable = 'NOT_EMPTY';
                }
```
is always false. (undeletable_bucket_set has jsonb objects)

Found two components which should use the bucket.undeletable flag - fronted and noobaa cli.
Fronted uses the flag in order to disable "delete" button in bucket list. For now the button is always enabled, but user is not able to delete bucket, since the core makes the validation and returns to ui error if the bucket is not empty.
Cli uses the flag to show a bucket status.

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -743,19 +743,6 @@ class MDStore {
         return Boolean(obj);
     }
 
-    async all_buckets_with_completed_objects() {
-        return this._objects.distinct('bucket', {
-            // prefix for stored blob blocks information. TODO: move somwhere like config.js
-            key: { $not: /^\.noobaa_blob_blocks/ },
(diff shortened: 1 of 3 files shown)


igorpick:
@guymguym @nimrod-becker please share you opinion about that.
if decide to remove it, will cleanup the fronted and cli as well.

dannyzaken (approved):

liranmauda (approved):
LGTM

Issue #6741, opened 2021-10-05
Segmentation fault, RC=139


akmithal:
### Environment info
- NooBaa Version: master-20211004
- Platform: OCP 4.8.9

### Actual behavior
1. IO caused endpoint panic - Segmentation fault

### Expected behavior
1. No error or panic 

### Steps to reproduce
1. Started IO using warp on newly upgraded code to master-20211004
2. Got endpoint panic 

```
[root@hpo-11 ~]# warp mixed --access-key=LFJdyLI42bsYcxtvvfAs  --secret-key=opRVOT30imM3H3lVFOlp+I/vsJDa9BziBDj0mqpg --obj.size=9M --host=s3-noobaa.apps.ocp-akshat-1.cp.fyre.ibm.com --duration=10m --bucket=bucke
t-1 --debug --tls --insecure
warp: <ERROR> upload error: 503 Service Unavailable
warp: <ERROR> upload error: 503 Service Unavailable
warp: <ERROR> upload error: Put "https://s3-noobaa.apps.ocp-akshat-1.cp.fyre.ibm.com/bucket-1/iugsXfjR/1.6M4JeagDm0YS3GsB.rnd"
```

```
2021-10-05 05:48:00.999455 [PID-14/TID-14] [L1] FS::FSWorker::OnError: Mkdir _path=/nsfs/nsfs-nsr-1/bucket-1/.noobaa-nsfs_615aeb5a264b0b0033672e11 _mode=504  error.Message()=File exists 
2021-10-05 05:48:00.999524 [PID-14/TID-14] [L1] FS::FSWorker::Begin: Mkdir _path=/nsfs/nsfs-nsr-1/bucket-1/.noobaa-nsfs_615aeb5a264b0b0033672e11/uploads _mode=504 
2021-10-05 05:48:01.016781 [PID-14/TID-22] [L1] FS::FSWorker::Execute: Fsync _path=/nsfs/nsfs-nsr-1/bucket-1/p89Ok9km  took: 18.3327 ms
2021-10-05 05:48:01.016850 [PID-14/TID-22] [L1] FS::FSWorker::Execute: Mkdir _path=/nsfs/nsfs-nsr-1/bucket-1/.noobaa-nsfs_615aeb5a264b0b0033672e11 _mode=504 _uid=5000 _gid=800 _backend=GPFS 
[...]

PR #6750, opened 2021-10-11
Handle Segmentation fault due to hash calculations in NSFS

jeniawhite:
### Explain the changes
1. Writing ChunkFS to handle NSFS data buffers

1. https://github.com/noobaa/noobaa-core/issues/6741

### Testing Instructions:
1.


liranmauda (approved):
LGTM,
Just a question, Is there a reason why we are calling the callback cb in some of the functions?

PR #6779, opened 2021-10-25
Removing googleapis module

liranmauda:
### Explain the changes
- Removing googleapis module
- Removing `_notify_latest_version` which is not used


jeniawhite (approved):
Need to make sure that we do not want this capability anymore.
We can rewrite that with AWS S3 for example.
I know that for DS we are covering the versions and updates, but for US maybe we want people to know about newer versions as well?

PR #6791, opened 2021-11-01
Bumped prom-client to 14.0.0

liranmauda:
### Explain the changes
1. Bumped prom-client to 14.0.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6220,9 +6220,9 @@
       "dev": true
     },
     "prom-client": {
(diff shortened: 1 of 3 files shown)


liranmauda:
Tested on OCP 4.9 with OCS 4.9

We can see that we can get the metrics:
Metrics_·_Red_Hat_OpenShift_Container_Platform

jackyalbo (approved):

PR #6814, opened 2021-11-22
Removing "allowed buckets" functionality

BarakSason:
The "allowed buckets" functionality interferes with
another access control mechanism which was implemented
and thus needs to be removed

### DIFF
--- a/src/server/common_services/auth_server.js
+++ b/src/server/common_services/auth_server.js
@@ -591,27 +591,14 @@ function _prepare_auth_request(req) {
         dbg.log3('load auth system:', req.system && req.system._id);
     };
 
-    req.has_bucket_permission = function(bucket) {
-        return has_bucket_permission(bucket, req.account);
-    };
-
     req.has_bucket_anonymous_permission = function(bucket) {
         return has_bucket_anonymous_permission(bucket);
-    };
-
-    req.check_bucket_permission = function(bucket) {
-        if (!req.has_bucket_permission(bucket)) {
-            throw new RpcError('UNAUTHORIZED', 'No permission to access bucket');
-        }
     };
 
     req.has_s3_bucket_permission = function(bucket) {
         // Since this method can be called both authorized and unauthorized
         // We need to check the anonymous permission only when the bucket is configured to server anonymous requests
         // In case of anonymous function but with authentication flow we roll back to previous code and not return here
-        if (req.auth_token && typeof req.auth_token === 'object') {
-            return req.has_bucket_permission(bucket);
-        }
         // If we came with a NooBaa management token then we've already checked the method permissions prior to this function
         // There is nothing specific to bucket permissions for the management credentials
         // So we allow bucket access to any valid auth token
@@ -625,13 +612,6 @@ function _prepare_auth_request(req) {
 
         return false;
     };
-
-    req.check_s3_bucket_permission = function(bucket) {
-        if (!req.has_s3_bucket_permission(bucket)) {
-            throw new RpcError('UNAUTHORIZED', 'No permission to access bucket');
-        }
-    };
-
 }
 
 function _get_auth_info(account, system, authorized_by, role, extra) {
@@ -654,14 +634,6 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
     }
 
     return response;
-}
-
-function has_bucket_permission(bucket, account) {
-    return _.get(account, 'allowed_buckets.full_permission', false) ||
-        _.find(
-            _.get(account, 'allowed_buckets.permission_list') || [],
-            allowed_bucket => String(allowed_bucket._id) === String(bucket._id)
-        );
 }
 
 // TODO: This should be changed to ACLs / Bucket Policy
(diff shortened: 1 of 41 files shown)


.settings/org.eclipse.wst.jsdt.ui.superType.container
```diff
@@ -0,0 +1 @@
+org.eclipse.wst.jsdt.launching.JRE_CONTAINER
```
nimrod-becker, line 1: please remove

.settings/org.eclipse.wst.jsdt.ui.superType.name
```diff
@@ -0,0 +1 @@
+Global
```
nimrod-becker, line 1: please remove

nimrod-becker:
Notice deepscan new issues and lint issues failing all the tests

jackyalbo:
@BarakSason Thinking maybe we should split this PR? maybe one PR for UI, that we could easily push if it doesn't break. and the rest on a second one..  what do you think? will be easier to go through...

src/server/common_services/auth_server.js
```diff
@@ -591,10 +591,6 @@ function _prepare_auth_request(req) {
         dbg.log3('load auth system:', req.system && req.system._id);
     };
 
-    req.has_bucket_permission = function(bucket) {
```
jackyalbo, line 594: has_s3_bucket_permission uses this function - not sure you can just remove. Also not sure what is our plan about this once allowed buckets are removed? to deny all requests or to allow them.

jackyalbo, line 594: has_s3_bucket_permission uses this function - not sure you can just remove. Also not sure what is our plan about this once allowed buckets are removed? to deny all requests or to allow them.

jackyalbo (changes requested):

src/server/common_services/auth_server.js
BarakSason: @jackyalbo I'm fine with splitting this. @nimrod-becker Your thoughts?
As far as I understand, there is a different mechanism which is used instead of the "allowed buckets" one, so it's not a matter of allowing/denying everything as the other mechanism will take care of it (Nimrod - If I'm mistaken, please correct me).

nimrod-becker: Regarding splitting, we can do that that is ok.

Allow/Deny will be be set according to the bucket policy, I believe, but might be mistaken, that Jacky is asking what would be the default in case there is no explicit rule. We should behave as S3 would, can easily test it on an S3 bucket

guymguym: Hey, I think that the default when there is no bucket policy should be to allow only the owner of the bucket, like AWS does ->

https://docs.aws.amazon.com/AmazonS3/latest/userguide/access-control-overview.html#about-resource-owner

>Amazon S3 bucket and object ownership
>
> Buckets and objects are Amazon S3 resources. By default, only the resource owner can access these resources. The resource owner refers to the AWS account that creates the resource. For example:
> 
> 1. The AWS account that you use to create buckets and upload objects owns those resources.
> 
> 2. If you upload an object using AWS Identity and Access Management (IAM) user or role credentials, the AWS account that the user or role belongs to owns the object.
> 
> 3. A bucket owner can grant cross-account permissions to another AWS account (or users in another account) to upload objects. In this case, the AWS account that uploads objects owns those objects. The bucket owner does not have permissions on the objects that other accounts own, with the following exceptions:
> 
>    - The bucket owner pays the bills. The bucket owner can deny access to any objects, or delete any objects in the bucket, regardless of who owns them.
> 
>     - The bucket owner can archive any objects or restore archived objects regardless of who owns them. Archival refers to the storage class used to store the objects. For more information, see Managing your storage lifecycle.
>

liranmauda (commented):
@BarakSason 
Changing the UI requires building it. as I don't see `lib.js` and `lib.js.map` files I assume that it was missed.
Please run:
`cd frontend`
`npm run build-lib`
and then commit those changes (I assume that there will be changes).

liranmauda:
> @BarakSason Thinking maybe we should split this PR? maybe one PR for UI, that we could easily push if it doesn't break. and the rest on a second one.. what do you think? will be easier to go through...

@jackyalbo Not sure how it can be split. 
this PR is removing `allowed_buckets` from the schema which will probably break the UI.
WDYT?

src/server/system_services/bucket_server.js
```diff
@@ -1551,12 +1481,6 @@ function get_bucket_info({
     info.triggers = _.map(bucket.lambda_triggers, trigger => {
         const ret_trigger = _.omit(trigger, '_id');
         ret_trigger.id = trigger._id.toString();
-        const func = _.find(func_configs, func_config =>
-            func_config.name === trigger.func_name &&
-            func_config.version === trigger.func_version
-        );
-        const exec_account = system_store.get_account_by_email(func.exec_account);
-        ret_trigger.permission_problem = !auth_server.has_bucket_permission(bucket, exec_account);
```
jackyalbo, line 1559: Again - none bucket owner should lead to permission problem

jackyalbo (changes requested):

jackyalbo:
@BarakSason also I think you are missing on upgrade script - did you test an upgrade? Not sure what you were planning to do if a system already have account with allowed bucket - but it needs to be handled somehow.

nimrod-becker:
Closing will update when a new PR replacing this one will come up

PR #6819, opened 2021-11-24
Invalidate cache on ListObjects

baum:
While listing objects compare hub and cache
and delete objects from cache if
- object is missing in hub (out of band delete)
- object is outdated (out of band object update)

Fixes:
- https://bugzilla.redhat.com/show_bug.cgi?id=2024107

### DIFF
--- a/src/sdk/namespace_cache.js
+++ b/src/sdk/namespace_cache.js
@@ -177,13 +177,61 @@ class NamespaceCache {
     // OBJECT LIST //
     /////////////////
 
-    async list_objects(params, object_sdk) {
-        const get_from_cache = params.get_from_cache;
-        params = _.omit(params, 'get_from_cache');
-        if (get_from_cache) {
-            return this.namespace_nb.list_objects(params, object_sdk);
+    // While listing objects compare hub and cache
+    // and delete objects from cache if
+    // - object is missing in hub (out of band delete)
+    // - object is outdated (out of band object update)
+    async _invalidate_on_list(cache_promise, hub_promise, object_sdk) {
+        const [hub_res, cache_res] = await Promise.allSettled([hub_promise, cache_promise]);
+
+        // if one of the promises was rejected,
+        // get out of here
+        const hub_ok = hub_res.status === 'fulfilled';
+        const cache_ok = cache_res.status === 'fulfilled';
+        if (!hub_ok || !cache_ok) {
+            return;
         }
-        return this.namespace_hub.list_objects(params, object_sdk);
+
+        // filter out objects in cache list which
+        // are not marked is_latest
+        const cobjs = cache_res.value.objects.filter(o => o.is_latest);
+        const hobjs = hub_res.value.objects;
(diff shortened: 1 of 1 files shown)


guymguym:
Hey @baum I am not sure we should use the accidental list call to invalidate the cache - our design was to **validate** on GET object based on TTL.

See here -
https://github.com/noobaa/noobaa-core/blob/60c5ced7347593b8f488ed0dd1957a46ee40c08a/src/sdk/namespace_cache.js#L233-L237

Can you explain why our validations on GET did not work in your reproduction?

baum:
Hello @guymguym,

Let me explain. The scenario is an out-of-band update of a cached object in the hub with a larger one. By design, NooBaa ListObjects() returns the hub's picture of the world.

The client (AWS CLI) is doing [sync](https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html), and issues ranged [GetObject()](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html), after performing ListObjects(). The client uses an open range `X-`, the NooBaa endpoint  throws an InvalidRange exception [here](https://github.com/noobaa/noobaa-core/blob/60c5ced7347593b8f488ed0dd1957a46ee40c08a/src/util/http_utils.js#L218), since the range start `X` is greater than the cached object size.

The root cause is inconsistency between ListObjects() and GetObject(). Please see the [bz](https://bugzilla.redhat.com/show_bug.cgi?id=2024107) for more info.

guymguym:
Hey, I saw the BZ and I believe I understand the issue.
My question is - why didn't our validation check catch it in get-object?
We check in every get object request to match the Etag/LastModifiedTime with the hub before we can use the locally cached object data.
In addition - we have a TTL setting which can be set to allow such inconsistency by design, in exchange for less validations.

dannyzaken:
@guymguym AFAIK we don't validate on get-object unless the TTL has expired. 
I don't think we should treat it as an accidental list, it is a common pattern to list and then get-object. since the list returns the etag why isn't it a good idea to use it for validation?

baum:
@dannyzaken 

Agree, let's use `etag` instead of `last_modified_time`. More robust this way.

```diff
diff --git a/src/sdk/namespace_cache.js b/src/sdk/namespace_cache.js
index 040a9f7e0..84d0919a1 100644
--- a/src/sdk/namespace_cache.js
+++ b/src/sdk/namespace_cache.js
@@ -211,7 +211,7 @@ class NamespaceCache {
                 const ho = hobjs[j]; // current hub list object
                 const missing_in_hub = (ho.key.localeCompare(co.key) > 0);
                 const same_key = (ho.key.localeCompare(co.key) === 0);
-                const outdated = (ho.last_modified_time > co.last_modified_time);
+                const outdated = (ho.etag !== co.etag);
                 should_delete = (missing_in_hub || (same_key && outdated));
                 if (same_key) {
                     j += 1; // next hub list object
```

PR #6821, opened 2021-11-25
Fix test s3 ops for namespace buckets

romayalon:

### Explain the changes
1. fix tests for namespace buckets (nightly run)

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/test/unit_tests/test_s3_ops.js
+++ b/src/test/unit_tests/test_s3_ops.js
@@ -1,5 +1,7 @@
 /* Copyright (C) 2016 NooBaa */
 /* eslint max-lines-per-function: ['error', 700] */
+/* eslint-disable no-invalid-this */
+
 'use strict';
 const _ = require('lodash');
 // setup coretest first to prepare the env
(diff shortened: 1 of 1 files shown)


igorpick (approved):

liranmauda (approved):
LGTM, 
Just one comment, as we are disabling the eslint rule `no-invalid-this` in every timeout line by doing `// eslint-disable-line no-invalid-this` we can do it for the whole file by doing `/* eslint-disable no-alert */` and then we can remove the disable of each line.
I think it is better as we will not need to worry about it every time we are using `this.` on this file.

romayalon:
> LGTM, Just one comment, as we are disabling the eslint rule `no-invalid-this` in every timeout line by doing `// eslint-disable-line no-invalid-this` we can do it for the whole file by doing `/* eslint-disable no-alert */` and then we can remove the disable of each line. I think it is better as we will not need to worry about it every time we are using `this.` on this file.

@liranmauda I added this line instead /* eslint-disable no-invalid-this */

PR #6826, opened 2021-12-02
Provide access keys on account creation

### LINKED ISSUE #6822: Feature Request: Add possibility to update s3 credentials via RPC API
We safe the s3 credentials of noobaa in our vault and when we completely reinstall the noobaa instance (with wipe of DB) we automatically recreate the accounts via the RPC API, but when we do so new credentials get automatically created. We then need to update to the new credentials everywhere we use them

Either having the possibility to provide your own access and secret key when creating an account via the RPC API (`account_api` => `create_account`) or be able to update the credentials via the RPC API after creating the account (maybe `account_api` => `update_account`)

It would be possible to directly change them in the db, but this is definitely not the preferred qay

**Additional context**

jeniawhite:
### Explain the changes
1. Allowing to provide user's custom access_keys when creating an account

1. Fixed https://github.com/noobaa/noobaa-core/issues/6822

### Testing Instructions:
1.

### DIFF
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -73,6 +73,12 @@ module.exports = {
                     nsfs_account_config: {
                         $ref: 'common_api#/definitions/nsfs_account_config'
                     },
+                    access_keys: {
+                        type: 'array',
+                        items: {
+                            $ref: 'common_api#/definitions/access_keys'
+                        }
+                    },
                 },
             },
             reply: {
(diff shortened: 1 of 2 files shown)


romayalon (approved):

liranmauda (approved):
We should also give the option to update with provided access keys. 
something like `generate_account_keys` but instead of generating, update with the provided. 

With that said, It can be in a different PR.
LGTM.

PR #6828, opened 2021-12-05
Fix lifecycle deleting in minutes instead of days

### LINKED ISSUE #6818: Files are deleted from the bucket regardless of the expiration rule defined on the lifecycle configuration
### Environment info
- NooBaa Version: 5.7.0
- Platform: OpenShift 4

### Actual behavior
1. I'm trying to set a lifecycle configuration on my bucket so objects older than 4 days will be expired (deleted).
Thing is that, for some reason, ~30 minutes after setting that cleanup rule all the files in the bucket are deleted, regardless of their creation date.

### Expected behavior
1. Only files that are older than 4 days should be deleted

### Steps to reproduce
1. Existing S3 bucket with existing files. some of the files were uploaded to the bucket more that 4 days ago, some of the files were uploaded to the bucket less than 4 days ago.
2. Call PutBucketLifecycleConfiguration REST API to set a cleanup rule that expires the files after 4 days.
3. Actual - ~30 minutes after that REST call, ALL the files are deleted from the bucket.

The payload used to setup the lifecycle configuration:

`<LifecycleConfiguration>
    <Rule>
        <ID>ExpirationRule</ID>
        <Prefix></Prefix>
        <Status>Enabled</Status>
        <Expiration>
            <Days>4</Days>
        </Expiration>
    </Rule>
</LifecycleConfiguration>`

liranmauda:
### Explain the changes
- Fix lifecycle deleting in minutes instead of days
- Async/await lifecycle

1. Fixes: #6818

### DIFF
--- a/src/server/bg_services/lifecycle.js
+++ b/src/server/bg_services/lifecycle.js
@@ -16,79 +16,66 @@ var LIFECYCLE = {
     schedule_min: 1 //run every 5 minutes
 };
 
-function background_worker() {
+async function background_worker() {
     const system = system_store.data.systems[0];
-    if (!system) return P.resolve();
-    return P.fcall(function() {
-            dbg.log0('LIFECYCLE READ BUCKETS configuration:', 'BEGIN');
-            return system_store.refresh()
-                .then(function() {
-                    _.each(system_store.data.buckets, function(bucket, i) {
-                        if (!bucket.lifecycle_configuration_rules || bucket.deleting) {
-                            return;
+    if (!system) return;
+    try {
+        dbg.log0('LIFECYCLE READ BUCKETS configuration: BEGIN');
+        await system_store.refresh();
+        for (const bucket of system_store.data.buckets) {
+            if (!bucket.lifecycle_configuration_rules || bucket.deleting) return;
+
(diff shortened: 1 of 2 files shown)


src/server/bg_services/lifecycle.js
```diff
@@ -16,80 +16,67 @@ var LIFECYCLE = {
...
-                                'status:', lifecycle_rule.status, 'last_sync',
-                                Math.floor((now - lifecycle_rule.last_sync) / 1000 / 60), 'min ago.');
+    if (!system) return;
+    try {
+        dbg.log0('LIFECYCLE READ BUCKETS configuration: BEGIN');
+        await system_store.refresh();
+        for (const bucket of system_store.data.buckets) {
+            if (!bucket.lifecycle_configuration_rules || bucket.deleting) return;
+
+            P.all(_.map(bucket.lifecycle_configuration_rules, async function(lifecycle_rule, j) {
```
dannyzaken, line 28: check if this has to be in the background or we should wait for it.

also maybe change to arrow function 
```suggestion
            P.all(_.map(bucket.lifecycle_configuration_rules, async (lifecycle_rule, j) => {
```

dannyzaken (approved):

PR #6838, opened 2021-12-21
Initial STS support

romayalon:

### Explain the changes
1. Added sts_rest, sts_utils, sts_errors, sts_sdk to the endpoint and handle requests and errors.
2. Added to endpoint the ability to accept sts.assumeRole operation and currently just generate random s3 credentials (session token will be handled in the follow-up PR)
3. Added role_config to account schema and to create_account/update_account API calls and implementation in account_server.
4. added assume_role_policy as part of the role_config, 
    4.1. this policy will be tested when a user tries to call sts.assumeRole according to the assume_role_policy of the 
           assumed role (assumed_account).
    4.2. on create_account/update_account this policy will be validated.
4. Added test_sts.js file that tests role_config in account and sts.assumeRole

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -105,7 +105,39 @@ module.exports = {
             }
         },
 
-        // TODO: Update to the relevant schema
+        assume_role_policy: {
+            type: 'object',
+            required: ['statement'],
+            properties: {
+                version: { type: 'string' },
+                statement: {
+                    type: 'array',
+                    items: {
+                        type: 'object',
+                        required: ['effect', 'action', 'principal'],
+                        properties: {
+                            effect: {
+                                enum: ['allow', 'deny'],
+                                type: 'string'
+                            },
+                            action: {
+                                type: 'array',
+                                items: {
+                                    type: 'string'
+                                }
+                            },
+                            principal: {
+                                type: 'array',
+                                items: {
+                                    $ref: '#/definitions/email',
+                                }
+                            }
+                        }
+                    }
+                },
+            }
+        },
+
         bucket_policy: {
             type: 'object',
             required: ['statement'],
@@ -848,5 +880,17 @@ module.exports = {
                 }
             }
         },
+        role_config: {
+            type: 'object',
+            required: ['role_name', 'assume_role_policy'],
+            properties: {
+                role_name: {
+                    type: 'string'
+                },
+                assume_role_policy: {
+                    $ref: '#/definitions/assume_role_policy'
+                }
+            }
+        }
     }
 };
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -20,9 +20,11 @@ const P = require('../util/promise');
 const config = require('../../config');
 const s3_rest = require('./s3/s3_rest');
 const blob_rest = require('./blob/blob_rest');
+const sts_rest = require('./sts/sts_rest');
 const lambda_rest = require('./lambda/lambda_rest');
 const endpoint_utils = require('./endpoint_utils');
 const FuncSDK = require('../sdk/func_sdk');
+const StsSDK = require('../sdk/sts_sdk');
 const ObjectIO = require('../sdk/object_io');
 const ObjectSDK = require('../sdk/object_sdk');
 const xml_utils = require('../util/xml_utils');
@@ -47,6 +49,7 @@ if (process.env.NOOBAA_LOG_LEVEL) {
  * @typedef {http.IncomingMessage & {
  *  object_sdk?: ObjectSDK;
  *  func_sdk?: FuncSDK;
+ *  sts_sdk?: StsSDK;
  *  virtual_hosts?: readonly string[];
  * }} EndpointRequest
  */
@@ -62,6 +65,7 @@ if (process.env.NOOBAA_LOG_LEVEL) {
  * @typedef {{
  *  http_port?: number;
  *  https_port?: number;
+ *  https_port_sts?: number;
  *  init_request_sdk?: EndpointHandler;
  * }} EndpointOptions
  */
@@ -73,6 +77,7 @@ async function start_endpoint(options = {}) {
     try {
         const http_port = options.http_port || Number(process.env.ENDPOINT_PORT) || 6001;
         const https_port = options.https_port || Number(process.env.ENDPOINT_SSL_PORT) || 6443;
+        const https_port_sts = options.https_port_sts || Number(process.env.ENDPOINT_SSL_PORT_STS) || 7443;
         const endpoint_group_id = process.env.ENDPOINT_GROUP_ID || 'default-endpoint-group';
 
         const virtual_hosts = Object.freeze(
@@ -124,17 +129,23 @@ async function start_endpoint(options = {}) {
         }
 
         const endpoint_request_handler = create_endpoint_handler(init_request_sdk, virtual_hosts);
+        const endpoint_request_handler_sts = create_endpoint_handler(init_request_sdk, virtual_hosts, true);
 
         const ssl_cert = await ssl_utils.get_ssl_certificate('S3');
         const ssl_options = { ...ssl_cert, honorCipherOrder: true };
         const http_server = http.createServer(endpoint_request_handler);
         const https_server = https.createServer(ssl_options, endpoint_request_handler);
+        const https_server_sts = https.createServer(ssl_options, endpoint_request_handler_sts);
 
         dbg.log0('Starting HTTP', http_port);
         await listen_http(http_port, http_server);
         dbg.log0('Starting HTTPS', https_port);
         await listen_http(https_port, https_server);
         dbg.log0('S3 server started successfully');
+
+        dbg.log0('Starting HTTPS STS', https_port_sts);
+        await listen_http(https_port_sts, https_server_sts);
+        dbg.log0('STS server started successfully');
 
         await prom_reporting.start_server(config.EP_METRICS_SERVER_PORT);
 
@@ -159,7 +170,7 @@ async function start_endpoint(options = {}) {
  * @param {readonly string[]} virtual_hosts
  * @returns {EndpointHandler}
  */
-function create_endpoint_handler(init_request_sdk, virtual_hosts) {
+function create_endpoint_handler(init_request_sdk, virtual_hosts, sts) {
     const blob_rest_handler = process.env.ENDPOINT_BLOB_ENABLED === 'true' ? blob_rest : unavailable_handler;
     const lambda_rest_handler = config.DB_TYPE === 'mongodb' ? lambda_rest : unavailable_handler;
 
@@ -168,7 +179,6 @@ function create_endpoint_handler(init_request_sdk, virtual_hosts) {
         endpoint_utils.prepare_rest_request(req);
         req.virtual_hosts = virtual_hosts;
         init_request_sdk(req, res);
-
         if (req.url.startsWith('/2015-03-31/functions')) {
             return lambda_rest_handler(req, res);
         } else if (req.headers['x-ms-version']) {
@@ -177,8 +187,14 @@ function create_endpoint_handler(init_request_sdk, virtual_hosts) {
             return s3_rest.handler(req, res);
         }
     };
+    /** @type {EndpointHandler} */
+    const endpoint_sts_request_handler = (req, res) => {
+        endpoint_utils.prepare_rest_request(req);
+        init_request_sdk(req, res);
+        return sts_rest(req, res);
+    };
 
-    return endpoint_request_handler;
+    return sts ? endpoint_sts_request_handler : endpoint_request_handler;
 }
 
 /**
@@ -191,6 +207,7 @@ function create_init_request_sdk(rpc, internal_rpc_client, object_io) {
     const init_request_sdk = (req, res) => {
         const rpc_client = rpc.new_client();
         req.func_sdk = new FuncSDK(rpc_client);
+        req.sts_sdk = new StsSDK(rpc_client, internal_rpc_client);
         req.object_sdk = new ObjectSDK(rpc_client, internal_rpc_client, object_io);
     };
     return init_request_sdk;
--- a/src/endpoint/sts/ops/sts_post_assume_role.js
+++ b/src/endpoint/sts/ops/sts_post_assume_role.js
@@ -1,0 +1,52 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const dbg = require('../../../util/debug_module')(__filename);
+const { StsError } = require('../sts_errors');
+
+/**
+ * https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html
+ */
+async function assume_role(req) {
+    dbg.log1('sts_post_assume_role body: ', req.body);
+    let assumed_role;
+    try {
+        assumed_role = await req.sts_sdk.get_assumed_role(req);
+    } catch (err) {
+        if (err.code === 'ACCESS_DENIED') {
+            throw new StsError(StsError.AccessDeniedException);
+        }
+        throw new StsError(StsError.InternalFailure);
+    }
+    // Temporary credentials are NOT stored in noobaa
+    // TODO: need to generate session token and store in it 
+    // the temporary credentials and expiry
+    const access_keys = await req.sts_sdk.generate_temp_access_keys();
+    return {
+        AssumeRoleResponse: {
+            AssumeRoleResult: {
+                AssumedRoleUser: {
+                    Arn: `arn:aws:sts::${assumed_role.access_key}:assumed-role/${assumed_role.role_config.role_name}/${req.body.role_session_name}`,
+                    AssumedRoleId: `${assumed_role.access_key}:${req.body.role_session_name}`
+                },
+                Credentials: {
+                    AccessKeyId: access_keys.access_key.unwrap(),
+                    SecretAccessKey: access_keys.secret_key.unwrap(),
+                    Expiration: '',
+                    SessionToken: ''
+                },
+                PackedPolicySize: 0
+            }
+        }
+    };
+}
+
+module.exports = {
+    handler: assume_role,
+    body: {
+        type: 'application/x-www-form-urlencoded',
+    },
+    reply: {
+        type: 'xml',
+    },
+};
--- a/src/endpoint/sts/sts_rest.js
+++ b/src/endpoint/sts/sts_rest.js
@@ -1,0 +1,234 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const _ = require('lodash');
+const dbg = require('../../util/debug_module')(__filename);
+const StsError = require('./sts_errors').StsError;
+const js_utils = require('../../util/js_utils');
+const http_utils = require('../../util/http_utils');
+const signature_utils = require('../../util/signature_utils');
+const system_store = require('../../server/system_services/system_store').get_instance();
+
+const STS_MAX_BODY_LEN = 4 * 1024 * 1024;
+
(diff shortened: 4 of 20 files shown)


src/api/common_api.js
```diff
@@ -105,7 +105,40 @@ module.exports = {
...
+                                items: {
+                                    type: 'string'
+                                }
+                            },
+                            principal: {
+                                // email of account
+                                type: 'array',
+                                items: {
+                                    wrapper: SensitiveString,
+                                }
```
guymguym, line 134: If this is an email then best ref to that common definition 
```suggestion
                                items: {
                                    $ref: '#/definitions/email',
                                }
```

src/endpoint/endpoint.js
```diff
@@ -168,9 +171,10 @@ function create_endpoint_handler(init_request_sdk, virtual_hosts) {
         endpoint_utils.prepare_rest_request(req);
         req.virtual_hosts = virtual_hosts;
         init_request_sdk(req, res);
-
         if (req.url.startsWith('/2015-03-31/functions')) {
             return lambda_rest_handler(req, res);
+        } else if (req.headers.authorization && req.headers.authorization.includes('/sts/')) {
```
guymguym, line 176: Is the authorization header the only way to identify STS requests? 
What is the format of this header for STS?
I can't find it here - https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html

The reason I think this is less desirable is that the same authorization mechanisms can be applied to multiple services. I see in the STS API examples that it always sends a query string with `?Version=2011-06-15&Action=...` perhaps we can check for those somehow? WDYT?

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -0,0 +1,51 @@
...
+            PackedPolicySize: 0
+        }
+    };
+}
+
+module.exports = {
+    handler: assume_role,
+    body: {
+        type: 'xml',
+    },
```
guymguym, line 47: does this request take any body at all?

src/sdk/sts_sdk.js
```diff
@@ -0,0 +1,100 @@
...
+
+const cloud_utils = require('../util/cloud_utils');
+const dbg = require('../util/debug_module')(__filename);
+const LRUCache = require('../util/lru_cache');
+const { RpcError } = require('../rpc');
+const signature_utils = require('../util/signature_utils');
+const db_client = require('../util/db_client').instance();
+
+// DO we need that cache here? 
+const account_cache = new LRUCache({
```
guymguym, line 12: can we combine this with the object_sdk cache?

src/endpoint/sts/sts_rest.js
```diff
@@ -0,0 +1,315 @@
...
+    const params = {};
+    const params_arr = action_and_params.split('&');
+    params_arr.map(param => {
+        const entry = param.split('=');
+        if (entry[0] === 'Action' || entry[0] === 'Version') return {};
+        params[_.snakeCase(_.lowerCase(entry[0]))] = decodeURIComponent(entry[1]);
+        return params;
+    });
+    return params;
+}
```
guymguym, line 241: Isn't `req.query` already parsed for our endpoint requests?? 
This should be handled by calling `prepare_rest_request` see - 
https://github.com/noobaa/noobaa-core/blob/31a2d8edd5ad8e46813d86bca0f5290b1312b6c7/src/endpoint/endpoint_utils.js#L7-L17

src/endpoint/endpoint.js
romayalon: nice catch :) it's actually something I discussed with @jeniawhite and we wanted your opinion about
While using the aws sdk, the query string was empty and this `?Version=2011-06-15&Action=...` string was available only after reading the body of the request.
Jenia said that when he did the POC he had this issue too and he had to read the body in the endpoint file and then after parsing this query string from the body, he was able to identify sts calls, but should we really always read the body?

src/endpoint/sts/ops/sts_post_assume_role.js
romayalon: yes, ?Version=2011-06-15&Action=... being passed through the body

src/sdk/sts_sdk.js
romayalon: another thing I discussed with @jeniawhite and we had some doubts about, do we want to share it with object_sdk? 
do we maybe want a different cache with a different expiry here since most of our operations in this sdk are on accounts

romayalon:
@guymguym another thing @jeniawhite and I had doubts about, instead of having one role_config per account should we maybe have an array of role_configs that might help us in the future if we will want to be more compatible with AWS? WDYT?

src/endpoint/endpoint.js
guymguym: Perhaps you know this but AWS uses an api definition called [Smithy](https://awslabs.github.io/smithy/index.html) which is open sourced under [github/awslabs](https://github.com/awslabs/smithy).
In Smithy terms AWS STS uses a protocol called "AWS Query" (you can see it [here](https://github.com/awslabs/smithy-rs/blob/e559e4cf592f05c58f5ac2211af9641c201b96b0/aws/sdk/aws-models/sts.json#L45) in the STS smithy model.
But you should see smithy docs for this AWS Query protocol - 
https://awslabs.github.io/smithy/1.0/spec/aws/aws-query-protocol.html

The first paragraph says:
> The AWS Query protocol uses HTTP and serializes HTTP requests using query string parameters and responses using XML. Requests can be sent using GET requests and placing parameters in the query string; however, sending requests using POST with an application/x-www-form-urlencoded Content-Type is preferred because some proxies, clients, and servers have limitations around the maximum amount of data that can be sent in a query string.

 
So POST method calls in this protocol use a body of application/x-www-form-urlencoded.
We probably want to support it not by trying to parse it as XML and get a bare string back and then parse it.
Also notice that you can parse it with the same querystring library from nodejs - 
https://nodejs.org/api/querystring.html#querystringparsestr-sep-eq-options

src/endpoint/sts/sts_rest.js
guymguym: so perhaps better to parse using [querystring.parse()](https://nodejs.org/api/querystring.html#querystringparsestr-sep-eq-options)

src/endpoint/sts/sts_rest.js
```diff
@@ -0,0 +1,315 @@
...
+    throw new StsError(StsError.NotImplemented);
+}
+
+function parse_params(action_and_params) {
+    const params = {};
+    const params_arr = action_and_params.split('&');
+    params_arr.map(param => {
+        const entry = param.split('=');
+        if (entry[0] === 'Action' || entry[0] === 'Version') return {};
+        params[_.snakeCase(_.lowerCase(entry[0]))] = decodeURIComponent(entry[1]);
```
guymguym, line 237: I don't think you mean to use  `_.lowerCase` and then `_.snakeCase`  because by lowering all letters first you don't get the typical word separation when the input is CamelCase 

`CamelCase => (lowercase) camelcase => (snakecase) camelcase`
or
`CamelCase => (snakecase) camel_case`

src/endpoint/sts/sts_rest.js
```diff
@@ -0,0 +1,315 @@
...
+
+function handle_error(req, res, err) {
+    var stserr =
+        ((err instanceof StsError) && err) ||
+        new StsError(RPC_ERRORS_TO_STS[err.rpc_code] || StsError.InternalFailure);
+
+    if (stserr.code === 'MalformedPolicy') {
+        stserr.message = err.message;
+        stserr.detail = err.rpc_data.detail;
+    }
```
guymguym, line 251: Is this copied over from s3_rest?

src/sdk/sts_sdk.js
```diff
@@ -0,0 +1,100 @@
...
+        const account_id_str = req.params.role_arn.split(':')[4];
+        let account_id;
+        try {
+            account_id = db_client.parse_object_id(account_id_str);
+        } catch (err) {
+            throw new RpcError('NO_SUCH_ACCOUNT', 'No such account with id: ' + account_id_str);
+        }
+        if (!account_id) throw new RpcError('NO_SUCH_ACCOUNT', 'No such account with id: ' + account_id);
+
+        let account = await this.rpc_client.account.read_account_by_id({ account_id });
```
guymguym, line 53: why is this not using the accounts_cache?

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -0,0 +1,51 @@
...
+        assumed_role = await req.sts_sdk.get_assumed_role(req);
+    } catch (err) {
+        if (err.code === 'ACCESS_DENIED') {
+            throw new StsError(StsError.AccessDeniedException);
+        }
+        throw new StsError(StsError.InternalFailure);
+    }
+    // Temporary credentials are NOT stored in noobaa
+    // TODO: need to generate session token and store in it 
+    // the temporary credentials and expiry
```
guymguym, line 23: Shouldn't we handle this TODO for anything to work?

src/sdk/sts_sdk.js
guymguym: I would prefer a single cache if possible. What would be the different expiry that you had in mind?

src/endpoint/sts/ops/sts_post_assume_role.js
romayalon: currently i'm just returning access keys, will generate and parse the session token in the following PR.

src/sdk/sts_sdk.js
romayalon: the key of account cache is access key and not account id, but after re-discussing it with @jeniawhite, I changed the account identifier when calling assume role to access_key since the account id is internal to the database, so I changed this line to use the account cache.

src/endpoint/sts/sts_rest.js
romayalon: removed thanks

src/sdk/sts_sdk.js
romayalon: sure, I eventually used the object_sdk accounts cache instead of this new one, I think that for now it's enough to have 1 and if in the future we will decide to do optimizations of the account cache for sts purposes we can split them. 
but regarding the expiry, I thought since we are mainly using the accounts info here we need shorter expirty.

src/endpoint/endpoint.js
romayalon: thank you, I used querystring.
should I move the actual reading of the body to the endpoint file?
the issue is that the s3 and the sts endpoint are being called on the same address, which means that if we will move it to the endpoint and will always read the body in order to identify if the service being called is STS we will do it on s3 ops too.

src/endpoint/sts/ops/sts_post_assume_role.js
guymguym: but if these are temp credentials that are not stored anywhere - how can you validate requests with those creds?

src/endpoint/endpoint.js
guymguym: not sure I understand the difficulty - we explicitly call http_utils.read_and_parse_body after identifying the API and operation, so we could just have a simple logic there to check the content-type and parse accordingly, and just make the op.body.type assert that the type is as expected.

https://github.com/noobaa/noobaa-core/blob/64cbdbaf6721ba2538e2564b177f201bc34a25c5/src/util/http_utils.js#L228-L235

src/endpoint/sts/ops/sts_post_assume_role.js
romayalon: maybe I'm missing something here..
what I'm saying is that after this PR will be merged the temporary credentials that a user will get as a response of assumeRole can not be used and will be denied. After merging the changes of the next PR the credentials can be accapted.

guymguym: Ah OK, I see now that this PR will not yet be usable until the next one. Thanks for clarifying

src/endpoint/endpoint.js
romayalon: so If I understood you correctly, you suggest that in endpoint.create_endpoint_handler() if the content-type is 'application/x-www-form-urlencoded' I'll call read_and_parse_body() and then I'll have the body parsed and by the body's properties, I'll be able to infer that this is STS API and not S3?

guymguym: No, content-type is only used for POST requests anyway, and it's not a clear differentiator between services. But actually - why don't we just listen on another separate port for STS? We don't have to use one port for both... It shouldn't be a problem to have the endpoint listen on two ports, and in kubernetes we can assign it to a whole new service/route as desired. 

My point about the content-type was different and not related to the question of how to differentiate between s3 and sts - but I don't want to confuse the two so I'll comment in the other thread

src/endpoint/sts/ops/sts_post_assume_role.js
guymguym: I think that body type here should be 'x-www-form-urlencoded' and not 'xml' (although you currently ignore that value because you have just a single operation in sts_rest and you parse the body just once in the main handle_request function.

What I would do is move the parsing of such body from sts_rest to http_utils and invoke it from [read_and_parse_body](https://github.com/noobaa/noobaa-core/blob/64cbdbaf6721ba2538e2564b177f201bc34a25c5/src/util/http_utils.js#L228) when the body.type specifies it and the request content-type header agrees...

src/endpoint/endpoint.js
romayalon: we wanted to avoid listening to another port but It seems to be necassary, np i'll add that, thanks.

src/endpoint/sts/ops/sts_post_assume_role.js
romayalon: I agree regarding the part that the type here should be 'x-www-form-urlencoded' and i'll move the parsing to read_and_parse_body..
but in order to identify the op (sts_post_assume_role) I need to read the body, so I didn't ignore the body.type because I have only one function, I did it because I had to read the body in order to identify the op

guymguym: I see so since `Action=AssumeRole` is inside the body you will have to read_and_parse_body just based on the content-type header, which will have to be called before resolving the op. Sounds ok.

src/endpoint/sts/sts_utils.js
```diff
@@ -0,0 +1,10 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+
+const OP_NAME_TO_ACTION = Object.freeze({
+    post_assume_role: 'sts:AssumeRole',
+});
+
+
+exports.OP_NAME_TO_ACTION = OP_NAME_TO_ACTION;
```
guymguym, line 10: This module seems to be a bit sad - maybe just move the code into sts_rest?

src/util/http_utils.js
```diff
@@ -287,6 +287,19 @@ async function parse_request_body(req, options) {
             throw new options.ErrorClass(options.error_invalid_body);
         }
     }
+    if (options.body.type.includes('application/x-www-form-urlencoded')) {
```
guymguym, line 290: Why `.includes()` and not just `===`?

src/endpoint/sts/sts_rest.js
```diff
@@ -0,0 +1,298 @@
...
+    res.setHeader('x-amz-id-2', req.request_id);
+
+    // note that browsers will not allow origin=* with credentials
+    // but anyway we allow it by the agent server.
+    res.setHeader('Access-Control-Allow-Origin', '*');
+    res.setHeader('Access-Control-Allow-Credentials', true);
+    res.setHeader('Access-Control-Allow-Methods', 'GET,POST,PUT,DELETE,OPTIONS');
+    res.setHeader('Access-Control-Allow-Headers',
+        'Content-Type,Content-MD5,Authorization,X-Amz-User-Agent,X-Amz-Date,ETag,X-Amz-Content-Sha256');
+    res.setHeader('Access-Control-Expose-Headers', 'ETag');
```
guymguym, line 57: maybe move to http_utils too

src/endpoint/sts/sts_rest.js
```diff
@@ -0,0 +1,298 @@
...
+    const reply = await op.handler(req, res);
+    http_utils.send_reply(req, res, reply, {
+        ...options,
+        body: op.body,
+        reply: op.reply
+    });
+}
+
+
+function check_headers(req) {
```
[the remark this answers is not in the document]
romayalon: moved it to http_utils, it caused some dependency issues do I had to change it a bit

src/util/http_utils.js
romayalon: it can contain more stuff, for instance: ```application/x-www-form-urlencoded; charset=utf-8```

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -0,0 +1,52 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const dbg = require('../../../util/debug_module')(__filename);
+const { StsError } = require("../sts_errors");
```
liranmauda, line 5:
```suggestion
const { StsError } = require('../sts_errors');
```

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html
+ */
+async function assume_role(req) {
+    dbg.log0('sts_post_assume_role body: ', req.body);
```
liranmauda, line 11: maybe slice the `req.body`? can it be too large?

src/sdk/sts_sdk.js
```diff
@@ -0,0 +1,84 @@
...
+    set_auth_token(auth_token) {
+        this.rpc_client.options.auth_token = auth_token;
+    }
+
+    get_auth_token() {
+        return this.rpc_client.options.auth_token;
+    }
+
+    async get_assumed_role(req) {
+        dbg.log0('sts_sdk.get_assumed_role body', req.body);
```
liranmauda, line 27: might be good idea to slice the body? or maybe dbg1?

src/test/unit_tests/coretest.js
```diff
@@ -175,15 +180,26 @@ function setup(options = {}) {
             default_handler: endpoint_request_handler,
         });
 
+        await announce('start_https_server');
```
liranmauda, line 183:
```suggestion
        await announce('start_https_server (sts)');
```

liranmauda (changes requested):

src/endpoint/sts/sts_rest.js
```diff
@@ -0,0 +1,234 @@
...
+    const account_info = await req.sts_sdk.get_assumed_role(req);
+    const assume_role_policy = account_info.role_config && account_info.role_config.assume_role_policy;
+    if (!assume_role_policy) throw new StsError(StsError.AccessDeniedException);
+    const method = _get_method_from_req(req);
+    const cur_account_email = req.sts_sdk.requesting_account && req.sts_sdk.requesting_account.email.unwrap();
+    // system owner by design can always assume role policy of any account
+    if ((cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
+
+    const permission = has_assume_role_permission(assume_role_policy, method, cur_account_email);
+    dbg.log0('sts_rest.authorize_request_policy permission is: ', permission);
```
liranmauda, line 137: Maybe add in the log the email? 
so we will know who this log refers to.

liranmauda (approved):
LGTM

PR #6841, opened 2021-12-23
Force `rpc_address` to use `ws` as the protocol to avoid using ICE

liranmauda:
### Explain the changes

- Force rpc_address to use ws as the protocol to avoid using ICE
- Fixing the agent port to 60101

As decided we will leave the ICE code in for now and we will do a cleanup once we are 100% about it.
We want to clean the code in a later version, probably at 4.11

### DIFF
--- a/src/server/node_services/nodes_monitor.js
+++ b/src/server/node_services/nodes_monitor.js
@@ -1303,10 +1303,15 @@ class NodesMonitor extends EventEmitter {
         if (!item.agent_info) return;
         if (!item.node_from_store) return;
         const system = system_store.data.get_by_id(item.node.system);
-        const rpc_proto = process.env.AGENTS_PROTOCOL || 'n2n';
-        const rpc_address = rpc_proto === 'n2n' ?
-            'n2n://' + item.node.peer_id :
-            rpc_proto + '://' + item.node.ip + ':' + (process.env.AGENT_PORT || 9999);
+        // TODO: remove this block as part of the clean up of ICE code. 
+        // We will set the rpc protocol to be 'ws' and the port to be 60101
+
+        // const rpc_proto = process.env.AGENTS_PROTOCOL || 'n2n';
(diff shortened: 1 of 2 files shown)


src/server/node_services/nodes_monitor.js
```diff
@@ -1303,10 +1303,15 @@ class NodesMonitor extends EventEmitter {
...
-            'n2n://' + item.node.peer_id :
-            rpc_proto + '://' + item.node.ip + ':' + (process.env.AGENT_PORT || 9999);
+        // TODO: remove this block as part of the clean up of ICE code. 
+        // We will set the rpc protocol to be 'ws' and the port to be 60101
+
+        // const rpc_proto = process.env.AGENTS_PROTOCOL || 'n2n';
+        // const rpc_address = rpc_proto === 'n2n' ?
+        //     'n2n://' + item.node.peer_id :
+        //     rpc_proto + '://' + item.node.ip + ':' + (process.env.AGENT_PORT || 60101);
+        const rpc_proto = 'ws';
```
guymguym, line 1313: shouldn't we be using `wss` for TLS?
n2n was configured to use TLS too - tcp_tls flag in n2n_config...

guymguym (changes requested):
See comments on _authenticate_agent_api and wss-vs-ws

liranmauda:
Moving to draft.
2 topics here:
1. We need to decide what to do with the agent authentication. we will have a talk regarding @guymguym suggestion
2. Unit tests are failing as we scaling the agent locally and with this change it is failing on rpc_adress which was hardcoded into the same port (60101)

dannyzaken:
let's have a discussion on that and see how we want to advance with this PR with regards to also moving to node 16

liranmauda:
Open a Jira story: https://issues.redhat.com/browse/MCGI-126
Closing.

PR #6849, opened 2022-01-03
Bucket lifecycle rule creation policies compatibility with aws new API

liranmauda:
### Explain the changes

***This PR intention is not to add any flow if there are missing, It will change the API to be compatible with AWS's new API.
due to the fact that we are not testing it, if there are missing flows they should be addressed on a different PR***

- Fixing bucket lifecycle rule creation policies to be compatible with AWS new API
- Adding Upgrade script

https://docs.aws.amazon.com/AmazonS3/latest/API/API_LifecycleRule.html

1.  https://bugzilla.redhat.com/show_bug.cgi?id=2029298

### DIFF
--- a/src/api/bucket_api.js
+++ b/src/api/bucket_api.js
@@ -614,16 +614,16 @@ module.exports = {
                         type: 'array',
                         items: {
                             type: 'object',
-                            required: ['id', 'prefix', 'status', 'expiration'],
+                            required: ['id', 'filter', 'status', 'expiration'],
                             properties: {
                                 id: {
                                     type: 'string'
                                 },
-                                prefix: {
-                                    type: 'string'
-                                },
                                 status: {
                                     type: 'string'
+                                },
+                                filter: {
+                                    $ref: 'common_api#/definitions/bucket_lifecycle_filter'
                                 },
                                 expiration: {
                                     type: 'object',
@@ -637,7 +637,6 @@ module.exports = {
                                         expired_object_delete_marker: {
                                             type: 'boolean'
                                         }
-
                                     }
                                 },
                                 abort_incomplete_multipart_upload: {
@@ -700,16 +699,16 @@ module.exports = {
                 type: 'array',
                 items: {
                     type: 'object',
-                    required: ['id', 'prefix', 'status', 'expiration'],
+                    required: ['id', 'filter', 'status', 'expiration'],
                     properties: {
                         id: {
                             type: 'string'
                         },
-                        prefix: {
-                            type: 'string'
-                        },
                         status: {
                             type: 'string'
+                        },
+                        filter: {
+                            $ref: 'common_api#/definitions/bucket_lifecycle_filter'
                         },
                         last_sync: {
                             idate: true
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -890,7 +890,44 @@ module.exports = {
                 assume_role_policy: {
                     $ref: '#/definitions/assume_role_policy'
                 }
+            },
+        },
+        bucket_lifecycle_filter: {
+            type: 'object',
+            properties: {
+                prefix: {
+                    type: 'string'
+                },
+                // TODO: the functionality is not yet implemented so it is commented out
+                // object_size_greater_than: {
+                //     $ref: 'common_api#/definitions/bigint'
+                // },
+                // object_size_less_than: {
+                //     $ref: 'common_api#/definitions/bigint'
+                // },
+                // tag: {
+                //     // It is an Array for the use of And property
+                //     // Which includes all the filter properties and and array of tags
+                //     // It is not in the api as we dont need it, we will just parse it.
+                //     type: 'array',
+                //     items: {
+                //         $ref: '#/definitions/bucket_lifecycle_filter_tag'
+                //     },
+                // },
             }
-        }
+        },
+        // TODO: the functionality is not yet implemented so it is commented out
+        // bucket_lifecycle_filter_tag: {
+        //     type: 'object',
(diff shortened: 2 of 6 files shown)


src/upgrade/upgrade_scripts/5.10.0/upgrade_bucket_lifecycle_structure.js
```diff
@@ -0,0 +1,52 @@
...
+            if (!bucket.lifecycle_configuration_rules || bucket.deleting) return;
+
+            let iteratee = 0;
+            for (const lifecycle_rule of bucket.lifecycle_configuration_rules) {
+                if (lifecycle_rule.prefix) {
+                    dbg.log0('bucket.lifecycle_configuration_rules[', iteratee, ']', lifecycle_rule);
+                    lifecycle_rule.filter = {};
+                    lifecycle_rule.filter.prefix = lifecycle_rule.prefix;
+                    delete lifecycle_rule.prefix;
+                    bucket.lifecycle_configuration_rules[iteratee] = lifecycle_rule;
```
jackyalbo, line 23: I'm not sure why you need to do it inline - I think it's better to just create a new array of changes and push when needed, without changing the system_store (at least not before the make_changes)

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -13,30 +13,42 @@ async function put_bucket_lifecycle(req) {
...
+        const current_rule = {
             id: rule_id,
-            prefix: rule.Prefix[0],
+            filter: {},
             status: rule.Status[0]
         };
+        if (rule.Prefix) {
+            throw new S3Error(S3Error.InvalidArgument);
+        }
+        if (rule.Filter) {
```
jackyalbo, line 37: what about all the other options in filter?  If we don't want to handle maybe we should throw something...

src/api/common_api.js
```diff
@@ -848,5 +848,51 @@ module.exports = {
                 }
             }
         },
+        bucket_lifecycle_filter: {
+            type: 'object',
+            properties: {
+                and: {
+                    type: 'object',
```
jackyalbo, line 855: not sure what it this and? It seems like the filter needs to be an array of this properties. take a look here: https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configuration-examples.html

jackyalbo (changes requested):

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
liranmauda: Sure, We will throw `S3Error.NotImplemented`

src/api/common_api.js
liranmauda: As we talked, removed the `and`, and we will just do it when we parse the request.
also as this PR is not containing new/missing flows commented out the unused properties

src/api/bucket_api.js
```diff
@@ -614,17 +614,17 @@ module.exports = {
                         type: 'array',
                         items: {
                             type: 'object',
-                            required: ['id', 'prefix', 'status', 'expiration'],
+                            required: ['id', 'status'],
```
dannyzaken, line 617: isn't expiration required?

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -13,30 +13,45 @@ async function put_bucket_lifecycle(req) {
...
+    //     <Prefix>logs/</Prefix> 
+    //   </Filter>
     //   <Status>Enabled</Status>
     //   <Expiration>
     //     <Days>365</Days>
     //   </Expiration>
     // </Rule>
     const lifecycle_rules = _.map(req.body.LifecycleConfiguration.Rule, rule => {
-        var rule_id = uuid().split('-')[0];
+        let rule_id = uuid().split('-')[0];
```
dannyzaken, line 25: why do you split the uuid?

src/upgrade/upgrade_scripts/5.10.0/upgrade_bucket_lifecycle_structure.js
```diff
@@ -0,0 +1,52 @@
...
+
+async function run({ dbg, system_store, system_server }) {
+
+    try {
+        dbg.log0('starting upgrade lifecycle...');
+        let should_make_change = false;
+        const buckets = [];
+        for (const bucket of system_store.data.buckets) {
+            //Do not update if there are no lifecycle_configuration_rules or if the bucket is being deleted.
+            if (!bucket.lifecycle_configuration_rules || bucket.deleting) return;
```
dannyzaken, line 14: why do you ignore deleting? I think we want to change the schema anyway.

src/upgrade/upgrade_scripts/5.10.0/upgrade_bucket_lifecycle_structure.js
```diff
@@ -0,0 +1,47 @@
...
+                expiration: lifecycle_configuration_rules.expiration,
+            }));
+
+            buckets.push({
+                _id: bucket._id,
+                lifecycle_configuration_rules: new_rules,
+            });
+        }
+
+        if (new_rules.length > 0) {
```
jackyalbo, line 30: Undefined new_rules will break your code. Check buckets length instead and then move new_rules declaration to inside the for.

jackyalbo (changes requested):

jackyalbo (approved):

Issue #6830, opened 2021-12-09
NSFS - Data integrity fails for S3 uploads


akmithal:
### Environment info
- NooBaa Version: 4.9.0.rc4 build
- Platform: OCP 4.9

### Actual behavior
1. Data integrity mismatch for upload of files

### Expected behavior
1. md5sum should match for file uploads

### Steps to reproduce
1. Created 50GB file on local filesystem and uploaded to s3 bucket
2. Check md5sum at both places and it mismatches

Run1:

```
[root@hpo-21 akshat]# md5sum big-files/big_file_50 
5a0ca502a09115c36bd51a739e72b2c1  big-files/big_file_50

[root@sc-200-x-master bucket-53]# md5sum big_file_50 
d422099144fd2ecb7b3f0f78a938571c  big_file_50
```

Run2:
Created 5x10GB files and upload to s3 bucket. md5 mismatched for all.

**Md5sum is different for all the 5 files.** 

Local FS:
```
[root@hpo-21 akshat]# md5sum medium-files/*
00a577da860b19af814a0d07a69efad1  medium-files/file_10GB_1
00a577da860b19af814a0d07a69efad1  medium-files/file_10GB_2
00a577da860b19af814a0d07a69efad1  medium-files/file_10GB_3
00a577da860b19af814a0d07a69efad1  medium-files/file_10GB_4
00a577da860b19af814a0d07a69efad1  medium-files/file_10GB_5
```

Storage backend:
```
[root@sc-200-x-master bucket-63]# md5sum *
29424e71ec499fdf8110fb1e1054594a  file_10GB_1
a2bdaf20eb8e94337f5ffa90b27ee4cc  file_10GB_2
d313286bd8516824ccc34aedc83db5af  file_10GB_3
70e7e455ae48b4b2e69c82b6cd538957  file_10GB_4
14a79070495deef7e0593d485cdae155  file_10GB_5
```

Run 3: Upload for 1K, 10M files worked correctly. md5sum worked. 

@jeniawhite , @nimrod-becker : Please screen this urgently.

akmithal:
I could see failure coming even for file upload of 1GB. md5sum worked fine for uploads of 1K,10M, 100M file sizes.

```
[root@hpo-21 small-dir]# md5sum file_1G 
a5a7630f6e224b836d9c6e09832618ec  file_1G

```

```
[root@sc-200-x-master bucket-73]# md5sum file_1G 
[...]

PR #6852, opened 2022-01-11
Ref FSWorker parameters in NSFS native

jeniawhite:
### Explain the changes
1. Holding up a ref for the parameters of the FSWorker
2. This avoids NodeJS garbage collection or movement of buffers and other parameters that are being used by the native worker

1. https://github.com/noobaa/noobaa-core/issues/6830

### Testing Instructions:
1.

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -145,6 +145,12 @@ set_statfs_res(Napi::Object res, Napi::Env env, struct statfs &statfs_res)
 struct FSWorker : public Napi::AsyncWorker
 {
     Napi::Promise::Deferred _deferred;
+    // _args_ref is used to keep refs to all the args for the worker lifetime,
+    // which is needed for workers that receive buffers,
+    // because in their ctor they copy the pointers to the buffer's memory,
+    // and if the JS caller scope does not keep a ref to the buffers until after the call,
+    // then the worker may access invalid memory...
+    Napi::ObjectReference _args_ref;
     pid_t _tid;
     uid_t _uid;
     gid_t _gid;
@@ -156,12 +162,14 @@ struct FSWorker : public Napi::AsyncWorker
     FSWorker(const Napi::CallbackInfo& info)
         : AsyncWorker(info.Env())
         , _deferred(Napi::Promise::Deferred::New(info.Env()))
+        , _args_ref(Napi::Persistent(Napi::Object::New(info.Env())))
         , _tid(0)
         , _uid(ThreadScope::orig_uid)
         , _gid(ThreadScope::orig_gid)
         , _errno(0)
         , _warn_threshold_ms(0)
     {
+        for (int i = 0; i < (int)info.Length(); ++i) _args_ref.Set(i, info[i]);
         Napi::Object config = info[0].As<Napi::Object>();
         if (config.Has("uid")) _uid = config.Get("uid").ToNumber();
         if (config.Has("gid")) _gid = config.Get("gid").ToNumber();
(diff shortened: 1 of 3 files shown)


src/util/chunk_fs.js
```diff
@@ -45,10 +45,9 @@ class ChunkFS extends stream.Transform {
 
     async _flush_buffers(callback) {
         if (this.q_buffers.length) {
-            const buffers_to_write = this.q_buffers;
+            await this.target_file.writev(this.fs_account_config, this.q_buffers);
             this.q_buffers = [];
             this.q_size = 0;
-            await this.target_file.writev(this.fs_account_config, buffers_to_write);
```
guymguym, line 51: Adding FSWorker _args_ref should already handle this case...

I think I prefer to keep resetting `this.q_buffers` **before** diving into any async code, because I worry that even if this might seem to work, it makes the code highly sensitive to any quirks in the nodejs stream/events behavior which can be a little hairy around the edges, even more than promises :) 

But we can still achieve the same and make super-duper sure that we keep ref on the buffers from this scope if you refer to it after writev:

```js
            const buffers_to_write = this.q_buffers;
            this.q_buffers = [];
            this.q_size = 0;
            await this.target_file.writev(this.fs_account_config, buffers_to_write);
            this._total_num_buffers += buffers_to_write.length;
```

src/native/fs/fs_napi.cpp
```diff
@@ -145,6 +145,7 @@ set_statfs_res(Napi::Object res, Napi::Env env, struct statfs &statfs_res)
 struct FSWorker : public Napi::AsyncWorker
 {
     Napi::Promise::Deferred _deferred;
+    Napi::ObjectReference _args_ref;
```
guymguym, line 148: We can add this to explain:
```suggestion

    // _args_ref is used to keep refs to all the args for the worker lifetime,
    // which is needed for workers that receive buffers,
    // because in their ctor they copy the pointers to the buffer's memory,
    // and if the JS caller scope does not keep a ref to the buffers until after the call,
    // then the worker may access invalid memory...
    Napi::ObjectReference _args_ref;
```

src/native/fs/fs_napi.cpp
```diff
@@ -156,12 +157,14 @@ struct FSWorker : public Napi::AsyncWorker
...
         : AsyncWorker(info.Env())
         , _deferred(Napi::Promise::Deferred::New(info.Env()))
+        , _args_ref(Napi::Persistent(Napi::Object::New(info.Env())))
         , _tid(0)
         , _uid(ThreadScope::orig_uid)
         , _gid(ThreadScope::orig_gid)
         , _errno(0)
         , _warn_threshold_ms(0)
     {
+        for (int i = 0; i < (int)info.Length(); ++i) _args_ref.Set(i, info[i]);
```
guymguym, line 167: Actually - we need to do the same for CryptoWorker !
Perhaps this can explain why we saw bad md5 calculations..?

https://github.com/noobaa/noobaa-core/blob/2b5600631c0c4964e35a966e889c76e19709606e/src/native/tools/crypto_napi.cpp#L22-L25

guymguym (changes requested):
@jeniawhite Couple of comments

src/util/chunk_fs.js
jeniawhite: The purpose of the total_num_buffers is just a dummy to hold the ref in the Javascript scope?
I do not see any other use for this variable other than that, am I right?

guymguym: Yes

guymguym (approved):

PR #6854, opened 2022-01-12
fix chunked upload

romayalon:

### Explain the changes
1. Fixed panic when destroying stream - when calling destroy(err) the error is emitted but no one listens to that event- removed it
2.  Added a flag reuse_last_stream - used when we do not want to add resume() call to the last stream in the call to pipeline in decode_chunked_upload since we will reuse the decoder stream.
3. stream_utils.pipeline - removed tap stream addition to the end of the streams array and called resume() on the last stream in the array instead.
according to this issue in node JS (https://github.com/nodejs/node/issues/40685) and according to the docs 'The readable.resume() method can be used to fully consume the data from a stream without actually processing any of that data'

1. Fixed BZ #2036211

### Testing Instructions:
1.

### DIFF
--- a/src/util/stream_utils.js
+++ b/src/util/stream_utils.js
@@ -4,7 +4,6 @@
 const util = require('util');
 const stream = require('stream');
 const events = require('events');
-const _ = require('lodash');
 
 
 /**
@@ -30,7 +29,7 @@ function get_tap_stream(func) {
 }
 
 const async_pipeline = util.promisify(stream.pipeline);
-async function pipeline(streams) {
+async function pipeline(streams, reuse_last_stream) {
     if (!streams || !streams.length) throw new Error('Pipeline called without streams');
     if (streams.find(strm => strm.destroyed)) {
         const err = new Error('Pipeline called on destroyed stream');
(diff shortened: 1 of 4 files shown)


src/util/stream_utils.js
```diff
@@ -30,19 +30,20 @@ function get_tap_stream(func) {
...
 
 const async_pipeline = util.promisify(stream.pipeline);
-async function pipeline(streams) {
+async function pipeline(streams, reuse_last_stream) {
     if (!streams || !streams.length) throw new Error('Pipeline called without streams');
     if (streams.find(strm => strm.destroyed)) {
         const err = new Error('Pipeline called on destroyed stream');
         for (const strm of streams) {
-            if (!strm.destroyed) strm.destroy(err);
+            if (!strm.destroyed) strm.destroy();
```
guymguym, line 38: @romayalon Isn't it just enough to try-catch around this call?

romayalon: try and catch won't be enough, but we can listen to the error event on the stream: 
```
const fooErr = new Error('foo error');
myStream.destroy(fooErr);
myStream.on('error', (fooErr) => console.error(fooErr.message)); // foo error****
```

but in the next line, we are throwing the same error so I'm not sure it's necessary to emit it and listen, WDYT?

jeniawhite (approved):

src/util/stream_utils.js
```diff
@@ -30,20 +29,28 @@ function get_tap_stream(func) {
...
-async function pipeline(streams) {
+async function pipeline(streams, reuse_last_stream) {
     if (!streams || !streams.length) throw new Error('Pipeline called without streams');
     if (streams.find(strm => strm.destroyed)) {
         const err = new Error('Pipeline called on destroyed stream');
         for (const strm of streams) {
-            if (!strm.destroyed) strm.destroy(err);
+            if (!strm.destroyed) {
+                strm.destroy(err);
+                strm.on('error', err1 => console.log(`stream_utils: error occured on stream: `, err1));
```
jeniawhite, line 39: This is quite a weird behavior
I would expect the streams to listen to these events in order not to crash
Doing that here means that you won't actually be able to distinguish between streams and do different logic in case of errors

jeniawhite (approved):

PR #6855, opened 2022-01-16
Fix complete_object_multiparts on empty part list

jackyalbo:

### Explain the changes
1. fix for https://bugzilla.redhat.com/show_bug.cgi?id=2040682
2. user upload an empty multipart - so no parts could be found for mp. leading to failure when trying to sort an undefined array. 

1. Fixed BZ2040682

### Testing Instructions:
1.

### DIFF
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -1927,7 +1927,7 @@ async function _complete_object_multiparts(obj, multipart_req) {
                 `multipart num=${num} etag=${etag} etag_md5_b64=${etag_md5_b64} not found in group ${util.inspect(group)}`);
         }
         md5.update(Buffer.from(mp.md5_b64, 'base64'));
-        const mp_parts = parts_by_mp[mp._id.toHexString()];
+        const mp_parts = parts_by_mp[mp._id.toHexString()] || [];
         _complete_next_parts(mp_parts, context);
         used_multiparts.push(mp);
         for (const part of mp_parts) {
--- a/src/test/unit_tests/test_s3_ops.js
+++ b/src/test/unit_tests/test_s3_ops.js
@@ -489,6 +489,65 @@ mocha.describe('s3_ops', function() {
                 }
             }).promise();
         });
+
+        mocha.it('should allow multipart with empty part', async function() {
+            this.timeout(600000);
+
+            const res1 = await s3.createMultipartUpload({
+                Bucket: bucket_name,
+                Key: text_file2
+            }).promise();
+
+            const res2 = await s3.uploadPart({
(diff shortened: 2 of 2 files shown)


src/server/object_services/object_server.js
```diff
@@ -1928,11 +1928,13 @@ async function _complete_object_multiparts(obj, multipart_req) {
         }
         md5.update(Buffer.from(mp.md5_b64, 'base64'));
         const mp_parts = parts_by_mp[mp._id.toHexString()];
-        _complete_next_parts(mp_parts, context);
-        used_multiparts.push(mp);
-        for (const part of mp_parts) {
-            used_parts.push(part);
+        if (mp_parts) { // handles case where multipart is of size 0 - so mp_parts is undefined
```
romayalon, line 1931: no error should be thrown here? AWS handles this situation the same?

jackyalbo: @romayalon Yes, same scenario works in AWS. So we shouldn't fail as well.

src/server/object_services/object_server.js
```diff
@@ -1928,11 +1928,13 @@ async function _complete_object_multiparts(obj, multipart_req) {
         }
         md5.update(Buffer.from(mp.md5_b64, 'base64'));
         const mp_parts = parts_by_mp[mp._id.toHexString()];
```
jeniawhite, line 1930: I think that you can use this instead:
```const mp_parts = parts_by_mp[mp._id.toHexString()] || [];```

My reasoning is that in other flows we have the same problem but since the return, there is an empty array we just pass the sorting without errors.

```
const map_res = req.rpc_params.multiparts ?
        await _complete_object_multiparts(obj, req.rpc_params.multiparts) :
        await _complete_object_parts(obj);
```
You can see that this returns an empty array in the parts case:
```
    /**
     * @param {nb.ObjectMD} obj
     * @returns {Promise<nb.PartSchemaDB[]>}
     */
    async find_all_parts_of_object(obj) {
        return this._parts.find({ obj: { $eq: obj._id, $exists: true }, deleted: null });
    }
```
That is the reason we aren't failing in other flows.
So I think that instead of changing the logic we can just align with the other case and pass an empty array when we do not have parts.
Another thing that we need to make sure of is that we are not breaking the context.num_parts.
A person can upload 20 parts and only 10 will have an actual size, in that case, the num_parts will be 10 instead of 20 because we won't account for them in the loops.
It is a bit confusing...

jackyalbo: @jeniawhite fixed and added a unit-test. please re-review and approve

jeniawhite (approved):

PR #6859, opened 2022-01-24
STS | Session tokens generation and parsing

romayalon:

### Explain the changes
1. Added missing session tokens generation in sts_post_assume_role op. session token consists of the temp access keys 
   and the access key of the assumed role (the access_key that is saved in the DB).
2. Added in s3_rest and sts_rest calls to authorize_session_token() function that will authorize the session token when it's defined.
3. In authenticate_request() in both s3_rest and sts_rest:
    3.1.  Swapped the auth token access key with the assumed_role_access_key (the access_key that is saved in the DB) 
            from the session token properties - done in order to get the actual permissions. 
    3.2. Saved the temp access key and secret key on the auth token in order to compare correctly the signature calculation.

3. Added unit tests 

1. 

### Testing Instructions:
1.

### DIFF
--- a/src/endpoint/sts/ops/sts_post_assume_role.js
+++ b/src/endpoint/sts/ops/sts_post_assume_role.js
@@ -3,6 +3,8 @@
 
 const dbg = require('../../../util/debug_module')(__filename);
 const { StsError } = require('../sts_errors');
+const jwt_utils = require('../../../util/jwt_utils');
+const config = require('../../../../config');
 
 /**
  * https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html
@@ -19,9 +21,9 @@ async function assume_role(req) {
         throw new StsError(StsError.InternalFailure);
     }
     // Temporary credentials are NOT stored in noobaa
-    // TODO: need to generate session token and store in it 
-    // the temporary credentials and expiry
+    // The generated session token will store in it the temporary credentials and expiry and the role's access key
     const access_keys = await req.sts_sdk.generate_temp_access_keys();
+    const expiry = config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS;
     return {
         AssumeRoleResponse: {
             AssumeRoleResult: {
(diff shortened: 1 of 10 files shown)


src/sdk/sts_sdk.js
```diff
@@ -73,7 +73,7 @@ class StsSDK {
                 }
             }
             const signature = signature_utils.get_signature_from_auth_token(
-                token, this.requesting_account.access_keys[0].secret_key.unwrap());
+                token, token.temp_secret_key || this.requesting_account.access_keys[0].secret_key.unwrap());
```
jeniawhite, line 76: In the ```auth_server``` you've used an intermediate variable for that, is there a reason not to use it here likewise?

src/sdk/object_sdk.js
```diff
@@ -137,7 +137,7 @@ class ObjectSDK {
                 }
             }
             const signature = signature_utils.get_signature_from_auth_token(
-                token, this.requesting_account.access_keys[0].secret_key.unwrap());
+                token, token.temp_secret_key || this.requesting_account.access_keys[0].secret_key.unwrap());
```
jeniawhite, line 140: Same comment here as in ```sts_sdk```

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -19,9 +21,9 @@ async function assume_role(req) {
         throw new StsError(StsError.InternalFailure);
     }
     // Temporary credentials are NOT stored in noobaa
-    // TODO: need to generate session token and store in it 
-    // the temporary credentials and expiry
+    // The generated session token will store in it the temporary credentials and expiry and the role's access key
     const access_keys = await req.sts_sdk.generate_temp_access_keys();
+    const expiry = process.env.STS_SESSION_TOKEN_EXPIRY || config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS;
```
jeniawhite, line 26: I would move this "OR" to the ```config.js```
It is better to have a single point of edits for the values instead of having "OR" in multiple points of code if we would want to know expiry in different parts of the code
Looking at a single source of truth seems like a simpler solution to maintain

src/server/common_services/auth_server.js
```diff
@@ -476,8 +476,8 @@ function _authorize_signature_token(req) {
         role: role.role,
         client_ip: auth_token_obj.client_ip,
     };
-
-    const signature = signature_utils.get_signature_from_auth_token(auth_token_obj, secret_key.unwrap());
+    const secret_key1 = auth_token_obj.temp_secret_key || secret_key.unwrap();
```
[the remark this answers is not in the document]
romayalon: renamed

src/util/http_utils.js
```diff
@@ -591,6 +591,20 @@ function parse_content_length(req, options) {
     return size;
 }
 
+function authorize_session_token(req, options) {
+    const jwt_utils = require('./jwt_utils'); // eslint-disable-line global-require
```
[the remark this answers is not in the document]
romayalon: yes, jwt_utils.js checks for JWT_SECRET env var, and we do not pass it to the default-backing-store pod, which causes a panic ```PANIC: process uncaughtException Error: JWT_SECRET env variable not found. it must exist when running in kuberentes```
since authorize_session_token() will be called only from the endpoint pod I thought that requiring from inside the function is an ok handling, let me know what you think

src/sdk/sts_sdk.js
romayalon: added a variable

src/sdk/object_sdk.js
romayalon: added a variable

src/endpoint/sts/ops/sts_post_assume_role.js
romayalon: I eventually deleted this env var since I could just change it in the config

jeniawhite (approved):

Issue #6848, opened 2022-01-03
Semaphore timeout using warp run


akmithal:
### Environment info
- NooBaa Version: master-20211214 
- Platform: OCP 4.9

### Actual behavior
1. Warp IO run failed with Connection timeout, Semaphore timeout 

### Expected behavior
1. No IO errors

### Steps to reproduce
1. Started 3 warp runs:

`warp mixed --access-key=ZUIjTle37eXW93MUxsSl  --secret-key=gZP8amlj19L65630lD0EvC2yuYEOaEQBr01q17pA --obj.size=9M --host=s3-openshift-storage.apps.ocp-akshat-2.cp.fyre.ibm.com --duration=20m --bucket=bucket-301 --debug --tls --insecure`

`warp mixed --access-key=RK3MTWAImWB5nuzH5QcW  --secret-key=WNS9UGuB9nyM2dtqYdekcbv8cWo6TrK3MDnhqYWE --obj.size=20M --host=s3-openshift-storage.apps.ocp-akshat-2.cp.fyre.ibm.com --duration=20m --bucket=bucket-302 --debug --tls --insecure
`

`warp mixed --access-key=XNKBQ9v5GktzRir3iR08  --secret-key=NMr1sSRut+qwp6Yo6v12I+iTujLIJI3XVFy9AyJJ --obj.size=100M --host=s3-openshift-storage.apps.ocp-akshat-2.cp.fyre.ibm.com --duration=10m --bucket=bucket-303 --debug --tls --insecure`

Errors:

```
warp: <ERROR> upload error: Put "https://s3-openshift-storage.apps.ocp-akshat-2.cp.fyre.ibm.com/bucket-302/u%28ZYPGmR/1.gqAXXMarEzb3Czlk.rnd": Connection closed by foreign host https://s3-openshift-storage.apps.ocp-akshat-2.cp.fyre.ibm.com/bucket-302/u%28ZYPGmR/1.gqAXXMarEzb3Czlk.rnd. Retry again.
warp: <ERROR> upload error: Put "https://s3-openshift-storage.apps.ocp-akshat-2.cp.fyre.ibm.com/bucket-302/C9mk2L6f/1.UC86nvzK5i%29C4bhI.rnd": Connection closed by foreign host https://s3-openshift-storage.apps.ocp-akshat-2.cp.fyre.ibm.com/bucket-302/C9mk2L6f/1.UC86nvzK5i%29C4bhI.rnd. Retry again.
[...]

PR #6868, opened 2022-02-03
Fixing NSFS semaphore leak

jeniawhite:
### Explain the changes
1. Handling the promises in the pipeline differently (caused unsettled promises on rejections)
2. Rewrite transform and flush async logic with the addition of try and catch to avoid unhandled promise rejections

1. https://github.com/noobaa/noobaa-core/issues/6848

### Testing Instructions:
1.

### DIFF
--- a/src/util/chunk_fs.js
+++ b/src/util/chunk_fs.js
@@ -8,7 +8,7 @@ const nb_native = require('./nb_native');
 
 /**
  *
- * ChunkNSFS
+ * ChunkFS
  * 
  * Calculates etag and writes stream data to the filesystem batching data buffers
  *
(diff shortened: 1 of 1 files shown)


src/util/chunk_fs.js
```diff
@@ -28,54 +28,50 @@ class ChunkFS extends stream.Transform {
...
+                size: chunk.length,
+                count: this.count
+            });
+            this.count = 0;
+            while (chunk && chunk.length) {
+                const available_size = config.NSFS_BUF_SIZE - this.q_size;
+                const buf = (available_size < chunk.length) ? chunk.slice(0, available_size) : chunk;
+                this.q_buffers.push(buf);
+                this.q_size += buf.length;
+                if (this.q_size === config.NSFS_BUF_SIZE) await this._flush_buffers();
```
guymguym, line 45: I'd use >= just being defensive
```suggestion
                if (this.q_size >= config.NSFS_BUF_SIZE) await this._flush_buffers();
```

guymguym (commented):
Missing details on what is the fix itself, and what is the pitfall to avoid in such a code. I'd put it in code comments or in the PR main comment, because I already forgot what was the exact fix here that we discussed before...

src/util/chunk_fs.js
```diff
@@ -28,54 +28,50 @@ class ChunkFS extends stream.Transform {
...
+                await this.target_file.writev(this.fs_account_config, buffers_to_write);
+                // Hold the ref on the buffers from the JS side
+                this._total_num_buffers += buffers_to_write.length;
+            }
+            if (callback) {
+                if (this.MD5Async) this.digest = (await this.MD5Async.digest()).toString('hex');
+                return callback();
+            }
+        } catch (error) {
+            return callback(error);
```
guymguym, line 73: Perhaps this should be printed here as warning/error? sounds like this is wouldn't be a good thing that we fail during this process of flushing data from memory to file... same for transform..
```suggestion
            console.error('ChunkFS FLUSH TO FILE FAILED', this.q_size, this._total_num_buffers, error);
            return callback(error);
```

src/util/chunk_fs.js
```diff
@@ -28,54 +28,50 @@ class ChunkFS extends stream.Transform {
...
+                const available_size = config.NSFS_BUF_SIZE - this.q_size;
+                const buf = (available_size < chunk.length) ? chunk.slice(0, available_size) : chunk;
+                this.q_buffers.push(buf);
+                this.q_size += buf.length;
+                if (this.q_size === config.NSFS_BUF_SIZE) await this._flush_buffers();
+                chunk = (available_size < chunk.length) ? chunk.slice(available_size) : null;
+            }
+            return callback();
         } catch (error) {
             return callback(error);
```
guymguym, line 50:
```suggestion
            console.error('ChunkFS TRANSFORM FAILED', this.q_size, this._total_num_buffers, error);
            return callback(error);
```

src/util/chunk_fs.js
```diff
@@ -28,54 +28,50 @@ class ChunkFS extends stream.Transform {
         this._total_num_buffers = 0;
     }
 
-    _transform(chunk, encoding, callback) {
+    async _transform(chunk, encoding, callback) {
         try {
-            this._process_chunk(chunk, callback);
+            if (this.MD5Async) await this.MD5Async.update(chunk);
+            stats_collector.instance(this.rpc_client).update_namespace_write_stats({
```
guymguym, line 34: I would add this condition to help running this class it in tests that do not care about stats:
```suggestion
            if (this.rpc_client) {
                stats_collector.instance(this.rpc_client).update_namespace_write_stats({
```

jeniawhite: I'm using a dummy RPC for that matter, you want me to revert the dummy RCP code from the unit_tests?
```
const DUMMY_RPC = {
    object: {
        update_endpoint_stats: (...params) => null
    }
};
```

guymguym:
@jeniawhite Can you provide a detailed explanation of what caused the unsettled promise in the code before?
I mainly want to learn from this - and from the info in the PR I couldn't figure it out.
Thanks

jeniawhite:
@guymguym 
There was a problem with ```writev``` writing to invalid addresses of memory
This caused an unhandled promise rejection from the pipeline which caused unsettled promises (the pipeline itself is being awaited for), which in turn caused a leak in the semaphore of the I/O of NSFS
I've wrapped things with try/catch and rewrote the methods to async at the top level (transform/flush)

guymguym:
Thanks.
The writev fix was in #6852 right?
So this fix is required as safety only?

jeniawhite:
@guymguym 
Yes, you are correct
But I need to clarify that this fix is not a safety only
Even with #6852 applied if we would have had a different problem from the native layer or any other exception inside the pipeline's ```transform/flush``` we would have a leak without this code
So this handles any future exceptions from the pipeline and cause them not to create a leak in the semaphore

guymguym (approved):

Issue #6766, opened 2021-10-18
Warp causing IO failure in noobaa endpoint log 


akmithal:
### Environment info
- NooBaa Version: v4.9.0-188.ci
- Platform: OCP 4.8.9 

### Actual behavior
1. IO failures using warp

### Expected behavior
1. IO failure shouldn't happen

### Steps to reproduce
1. Started IO using warp and got failures for all 3 runs in Noobaa endpoint logs:

For bucket-2:

Warp run:
```
[root@hpo-11 akshat]# warp mixed --access-key=MDJzrV8UnqMwsK8cgPig  --secret-key=bfNt3V0Pl4SOSjvvCAdm5pMdFLanLuOmRZxjxlxl --obj.size=1G --host=s3-openshift-storage.apps.ocp-akshat-1.cp.fyre.ibm.com --duration=20m --bucket=bucket-2 --debug --tls --insecure
warp: <ERROR> upload error: Put "https://s3-openshift-storage.apps.ocp-akshat-1.cp.fyre.ibm.com/bucket-2/Z7HyQxNj/2.quPJR2EKh6CszLfe.rnd?partNumber=2&uploadId=0d8c0328-c533-463a-9d1f-838b9a3fa989": write tcp 9.30.26.199:51626->9.30.38.152:443: write: broken pipe
warp: <ERROR> upload error: Put "https://s3-openshift-storage.apps.ocp-akshat-1.cp.fyre.ibm.com/bucket-2/roeWN10H/2.4oNfmf1gYj%299Mi%28w.rnd?partNumber=2&uploadId=30654bff-8d89-4e50-a0da-d1c2fd899bc2": write tcp 9.30.26.199:51652->9.30.38.152:443: use of closed network connection
warp: <ERROR> upload error: Put "https://s3-openshift-storage.apps.ocp-akshat-1.cp.fyre.ibm.com/bucket-2/y3ZWYpuv/2.D%28UZfeKJwmVQOOEV.rnd?partNumber=2&uploadId=49b391c1-6432-41d6-ac93-aa9986ce58d6": write tcp 9.30.26.199:51630->9.30.38.152:443: use of closed network connection
```

```
[...]

PR #6879, opened 2022-02-10
Fixing issue of NOENT when uploading object to deleted sub-dir

jackyalbo:
### Explain the changes
1. fix issue where upload fail with "The specified key does not exist" when in the same time delete happens that deletes the target directory (as part of being the last object in this sub-dir). In that case we will retry the rename and not just fail.

1. Partly fix to #6766

### Testing Instructions:
1.


liranmauda (approved):
LGTM

PR #6886, opened 2022-02-22
Defer MD5 calculation and assignment of the azure blob upload to prevent upload blocking

tangledbytes:

### Explain the changes
1. Remove MD5 calculation of the azure blob upload to prevent upload blocking

1. Fixes [BZ - 2041494](https://bugzilla.redhat.com/show_bug.cgi?id=2041494)

### Testing Instructions:
1. Deploy new image of NooBaa in the kubernetes cluster.
2. Create a  `azure-blob` namespace store.
3. Create a bucketclass for the above namespace store.
4. Create an OBC with above mentioned bucketclass.
5. Try uploading objects to newly created bucket.
6. The upload should be successful.

### Notes for the reviewers
**What was wrong?**
An attempt to compute MD5 hash from the `file_stream` was being made so that this value could be passed on to the `azure-store` client. The issue was that either we can compute the MD5 hash and exhaust the stream, hence nothing left to upload or we can attempt to generate MD5 has we upload (but this isn't possible as we have to provide MD5 before uploading).

It should be noted that providing MD5 hash to the `azure-store` client is optional and the client does no verification, hence essentially rendering the hash useless.

### Correctness Tests
- [x] NooBaa computed MD5 of the blob matches with `md5 <object>`.
- [x] `aws s3api head-object --bucket <bucket> --key <object>` ETag matches with the computed MD5.
- [x] `x-ms-content-md5` header is attached to the object.

### DIFF
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -250,17 +250,25 @@ class NamespaceBlob {
 
             this._check_valid_xattr(params);
             try {
-                const { new_stream, md5_buf } = await azure_storage.calc_body_md5(params.source_stream);
+                const { new_stream, md5_buf } = azure_storage.calc_body_md5(params.source_stream);
+                const headers = {
+                    blobContentType: params.content_type,
+                };
+                Object.defineProperty(headers, "blobContentMD5", {
+                    get: () => md5_buf()
+                });
+
                 obj = await blob_client.uploadStream(
                     new_stream.pipe(count_stream),
                     params.size,
-                    undefined, {
+                    undefined,
+                    {
                         metadata: params.xattr,
-                        blobHTTPHeaders: {
-                            blobContentType: params.content_type,
-                            blobContentMD5: md5_buf // TODO: check that before & after the change the MD5 was uploaded 
-                        }
-                    });
+                        blobHTTPHeaders: headers
+                    }
+                );
+                obj.contentMD5 = md5_buf();
+
             } catch (err) {
                 this._translate_error_code(err);
                 dbg.warn('NamespaceBlob.upload_object:', inspect(err));
--- a/src/util/new_azure_storage_wrap.js
+++ b/src/util/new_azure_storage_wrap.js
@@ -6,36 +6,40 @@ const azure_storage = require('@azure/storage-blob');
 // needed only for enerateBlockIdPrefix() and get_block_id() functions
 const old_azure_storage = require('azure-storage');
 
-
 function new_md5_stream() {
     const md5_stream = new stream.Transform({
         transform(buf, encoding, next) {
             this.md5.update(buf);
             this.size += buf.length;
-            this.push(buf);
-            next();
+            next(null, buf);
         }
     });
+
     md5_stream.md5 = crypto.createHash('md5');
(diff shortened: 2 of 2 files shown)


src/sdk/namespace_blob.js
```diff
@@ -250,15 +250,13 @@ class NamespaceBlob {
 
             this._check_valid_xattr(params);
             try {
-                const { new_stream, md5_buf } = await azure_storage.calc_body_md5(params.source_stream);
                 obj = await blob_client.uploadStream(
-                    new_stream.pipe(count_stream),
```
romayalon, line 255: @utkarsh-pro 
it's true that blobContentMD5 is optional, but since NooBaa is s3 compatible, when uploading an object we would like to return the ETag as the MD5 calculation of the object (you can see it _get_blob_object_info())

removing the MD5 calculation from this code will cause issues in the replication feature because we are testing if an object should be replicated also by its ETag.

utkarsh-pro: @romayalon I have re-added the MD5 hash computation but had to move it another Azure API blob call. Previously, we were trying to compute the md5 has before hand but this was leading to hanging the stream, now it computes the hash as the stream is written and then after the upload finishes it updates the `blobContentMD5` header value. Does this fix the replication issue?

Anything other approach that you would suggest?

romayalon: After adding the setHTTPHeaders() call there are 2 calls to Azure being done instead of 1 for each object upload, so it is less efficient. If there is a solution that still includes 1 op, it's preferred. 

Also, in the docs, they claim that calling setHTTPHeaders() will clear HTTP headers of the object that are not being set in the current call (which means blobContentType will be removed).
 
And regarding the correctness,I would verify the following:
1. MD5 calculation was done correctly.
2. After an object was uploaded to the Namespace Bucket get into Azure Blob UI and check that the MD5 was set successfully to the object.
3. Get the object from NooBaa and check that the ETag returned by NooBaa equals the MD5 calc.

utkarsh-pro: @romayalon I have changed the code so that it remains single op, however the approach is kinda "hacky". 

**Problem**: The issue is that we can compute the hash only after the stream is consumed by Azure SDK, however we also have to provide the hash in the function invocation that will consume the stream as a parameter.
**Hack**:  I have defined a custom getter for `blobContentMD5` field which internally invokes the md5 computation function. This function relies on the assumption that the stream has been consumed (so that the transform stream has seen all of the content). This is working at the moment because Azure SDK for blob storage at the moment performs the upload operation in two steps:
1. Upload the blobs and store the list of the block ids.
2. Commit the block id list and attach the http headers and metadata.
So because the headers are being accessed in step 2 (i.e. after the upload - stream consumption), we can compute the hash accurately. However if someday they decide to flip the flow (which happens in other Azure SDK) then we will start getting empty hash values.

This was the only 1 op solution I could think of, please let me know if I have missed something here. 

Also, I have updated the correctness tests in the description if the PR 😄 .

romayalon (approved):

utkarsh-pro:
@romayalon, done ✅

PR #6891, opened 2022-02-28
Disable NSFS md5 calculation

romayalon:

### Explain the changes
1. 

1. 

### Testing Instructions:
1.

### DIFF
--- a/config.js
+++ b/config.js
@@ -574,7 +574,7 @@ config.NSFS_UMASK = 0o007;
 // The threshold in milliseconds for prompting a warning
 config.NSFS_WARN_THRESHOLD_MS = 100;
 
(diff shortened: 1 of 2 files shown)


guymguym:
@romayalon Can you provide more info of why change the default to false?

nimrod-becker:
@guymguym This was requested for the initial release of NS FS

guymguym:
OK, I guess this is a temporary solution anyway.
Going forward this would have to go to the nsfs_account_config or nsfs_bucket_config because it can't really be a system wide configuration, and then we will only have to decide the default for new accounts/buckets.

nimrod-becker:
totally agree

nimrod-becker (approved):

PR #6901, opened 2022-03-07
initial lifecycle system/unit tests

baum:
- initial lifecycle system/unit tests
- new schema and get/put logic

### DIFF
--- a/.github/workflows/lifecycle.yaml
+++ b/.github/workflows/lifecycle.yaml
@@ -1,0 +1,15 @@
+name: Bucket Lifecycle Tests
+
+# Run on each new PR and each new push to existing PR
+on: [push, pull_request]
+
+jobs:
+  run-lifecycle-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: Run Unit Bucket Lifecycle Tests
+        id: run-test-bucket-lifecycle
+        run: make run-single-test testname=lifecycle_index.js || exit 1
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -73,22 +73,26 @@ module.exports = {
             }
         },
 
+        tag: {
+            type: 'object',
+            required: [
+                'key',
+                'value'
+            ],
+            properties: {
+                key: {
+                    type: 'string'
+                },
+                value: {
+                    type: 'string'
+                },
+            }
+        },
+
         tagging: {
             type: 'array',
             items: {
-                type: 'object',
-                required: [
-                    'key',
-                    'value'
-                ],
-                properties: {
-                    key: {
-                        type: 'string'
-                    },
-                    value: {
-                        type: 'string'
-                    },
-                }
+                $ref: '#/definitions/tag'
             }
         },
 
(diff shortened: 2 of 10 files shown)


liranmauda:
@baum Why do we want to run it as a single tests and not in the whole unit test suit?

baum:
@liranmauda, 

The answer is twofold:
1. Running in parallel to other tests makes the CI cycle shorter
2. Having a separate test, makes the development cycle shorter

liranmauda:
> @liranmauda,
> 
> The answer is twofold:
> 
> 1. Running in parallel to other tests makes the CI cycle shorter
> 2. Having a separate test, makes the development cycle shorter

I don't think we should break it into single tests.
Running it locally while developing is bound do forget single tests.

baum:
@liranmauda what do you suggest? could you elaborate?

src/api/common_api.js
```diff
@@ -138,6 +142,126 @@ module.exports = {
...
+                tag: {
+                    $ref: '#/definitions/tag'
+                },
+                object_size_greater_than: {
+                    type: 'number'
+                },
+                object_size_less_than: {
+                    type: 'number'
+                },
+                and: {
```
dannyzaken, line 201: we don't necessarily have to use the same structure as the REST API.
see the discussion here https://github.com/noobaa/noobaa-core/pull/6849#discussion_r790244522

src/api/common_api.js
```diff
@@ -138,6 +142,126 @@ module.exports = {
...
+                },
+            },
+        },
+        bucket_lifecycle_rule_filter: {
+            type: 'object',
+            properties: {
+                prefix: {
+                    type: 'string'
+                },
+                tag: {
```
dannyzaken, line 192: just to make my previous comment clearer - I think `tags` should be an array, and we can remove the `and` property. we should always treat any existing property as part of an `And`. unless it conflicts with a different possible behavior that I'm missing

baum: how about [get_bucket_lifecycle()](https://github.com/noobaa/noobaa-core/pull/6901/files/8c02125d28707ce2371c7378fe5b84d6dc0638ab#diff-431c35029c5e0154ce478a1f6672c076db12f50276241c058eb26cf4db63b0ab)? Should we always return `And`, even if `And` actually was not used during `put_bucket_lifecycle()`?

baum: This is an intelligent bright idea of generalization, however simplifying internal data representation beyond the point of not being able to reconstruct the original filter, is not a good idea IMHO.

jackyalbo: Maybe I'm missing something here? And doesn't just mean we have more than one filter? like one prefix and tag or more than one tags? why can't we understand that from what we have in the DB?

jackyalbo (dismissed):

src/api/common_api.js
dannyzaken: @jackyalbo, Alex's main concern is that the user can set a policy with `And` and only one element. in that case, when the user gets the policy, a different structure is returned. 
@baum correct me if I'm wrong

.github/workflows/lifecycle.yaml
```diff
@@ -0,0 +1,15 @@
+name: Bucket Lifecycle Tests
```
dannyzaken, line 1: @baum, I tend to agree with @liranmauda. you can add the tests to `src/test/unit_tests/index.js` instead of running them in a separate action.
we might want to eventually break our tests into different actions, but this is out of the scope of this PR

baum:
Closing this PR.
This logic is included in [Lifecycle PR](https://github.com/noobaa/noobaa-core/pull/6919)

baum:
See [here](https://github.com/noobaa/noobaa-core/pull/6919)

Issue #6907, opened 2022-03-09
HPO579:  make bucket (mb) BucketAlreadyOwned and then Read timeout


MonicaLemay:
This is from HPO defect 579

I get both BucketAlreadyOwned and then Read timeout

First, I get the message BucketAlreadyOwned :

Here is the workflow:

Account s50040 has 10 buckets the ESS and mmdas confirm 10 buckets

```

[root@c83f1-ems hpofs1]# ls -ali s50040
total 262
69671 drwxr-xr-x. 12 50040 50040 4096 Mar 4 16:31 .
3 drwxr-xr-x. 122 root root 262144 Mar 7 18:06 ..
4235040 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b1
4235046 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b10
4235041 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b2
4234998 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b3
4235042 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b4
4235043 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b5
4235044 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b6
4235045 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b7
4234995 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b8
4234902 drwxr-x---. 2 50040 50040 4096 Mar 4 16:31 s50040b9
```
Now,  let’s make more:

```
 s50040 mb s3://s50040b11
make_bucket failed: s3://s50040b11 An error occurred (BucketAlreadyOwnedByYou) when calling the CreateBucket operation: Your previous request to create the named bucket succeeded and you already own it. You get this error in all AWS regions except US East (N. Virginia) region, us-east-1. In us-east-1 region, you will get 200 OK, but it is no-op (if bucket exists it Amazon S3 will not do anything).

The time it takes is:
real 1m8.956s
user 0m0.504s
sys 0m0.070s
```

Again, another one:
```
s50040 mb s3://s50040b12
[...]

PR #6911, opened 2022-03-21
Fix parallel bucket creation

romayalon:

### Explain the changes
1. In an investigation of #6907 I noticed a situation in which there are 2 calls to create_bucket in the system (mb of the same bucket) in parallel or a retry (when NooBaa core is very busy).
I was able to see that on s3 mb retry (caused when NooBaa core is very busy) req.system.buckets_by_name didn't contain the newly created bucket entry (that was created on the first call to mb) and it didn't recognize that the bucket already exists on the second trial. I added req.load_auth() after aqcuiring the bucket_semaphore.surround_key(..) lock in both create_bucket and delete_bucket.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


jackyalbo (approved):
LGTM. if you can please update the details of your latest update to the code in the PR Conversation before merging.

PR #6919, opened 2022-03-23
Lifecycle feature enhancements 

baum:
###     Bucket lifecycle configuration

    - initial lifecycle system/unit tests
    - new schema and get/put logic

### Lifecycle - bg worker

    - background worker verification

### Testing
```shell  
- `make run-single-test testname=lifecycle_index.js`
- `make run-single-test-postgres testname=lifecycle_index.js`
```

### DIFF
--- a/config.js
+++ b/config.js
@@ -326,7 +326,7 @@ config.DEFAULT_S3_AUTH_METHOD = {
 // LIFECYCLE CONFIG //
 //////////////////////
 
-config.LIFECYCLE_INTERVAL = 15 * 60 * 1000;
+config.LIFECYCLE_INTERVAL = 8 * 60 * 60 * 1000; // 8h
 
 //////////////////////////
 // STATISTICS_COLLECTOR //
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -73,22 +73,26 @@ module.exports = {
             }
         },
 
+        tag: {
+            type: 'object',
+            required: [
+                'key',
+                'value'
+            ],
+            properties: {
+                key: {
+                    type: 'string'
+                },
+                value: {
+                    type: 'string'
+                },
+            }
+        },
+
         tagging: {
             type: 'array',
             items: {
-                type: 'object',
-                required: [
-                    'key',
-                    'value'
-                ],
-                properties: {
-                    key: {
-                        type: 'string'
-                    },
-                    value: {
-                        type: 'string'
-                    },
-                }
+                $ref: '#/definitions/tag'
             }
         },
 
@@ -135,6 +139,116 @@ module.exports = {
                         }
                     }
                 },
+            }
+        },
+
+        // bucket lifecycle
+
+        bucket_lifecycle_rule_expiration: {
+            type: 'object',
+            properties: {
+                days: {
+                    minimum: 1,
+                    type: 'integer'
+                },
+                date: {
+                    idate: true
+                },
+                /*
+                expired_object_delete_marker: {
+                    type: 'boolean'
+                }
+                */
+            }
+        },
+        bucket_lifecycle_rule_status: {
+            enum: ['Enabled', 'Disabled'],
+            type: 'string'
+        },
+        bucket_lifecycle_rule_filter: {
+            type: 'object',
+            properties: {
+                prefix: {
+                    type: 'string'
+                },
+                tags: {
+                    type: 'array',
+                    items: {
+                        $ref: '#/definitions/tag',
+                    },
+                },
+                object_size_greater_than: {
+                    type: 'number'
+                },
+                object_size_less_than: {
+                    type: 'number'
+                },
+                // conditional and filter operator, used to reconstruct
+                // the original XML S3 REST interface rule representation.
+                and: {
+                    type: 'boolean'
+                }
+            }
+        },
+        bucket_lifecycle_rule: {
+            type: 'object',
+            required: ['id', 'filter', 'expiration', 'status'],
+            properties: {
+                id: {
+                    type: 'string'
+                },
+                status: {
+                    $ref: '#/definitions/bucket_lifecycle_rule_status'
+                },
+                filter: {
+                    $ref: '#/definitions/bucket_lifecycle_rule_filter'
+                },
+                expiration: {
+                    $ref: '#/definitions/bucket_lifecycle_rule_expiration'
+                },
+                last_sync: {
+                    idate: true
+                },
+                /*
+                abort_incomplete_multipart_upload: {
+                    type: 'object',
+                    properties: {
+                        days_after_initiation: {
+                            type: 'integer'
+                        },
+                    }
+                },
+                transition: {
+                    type: 'object',
+                    properties: {
+                        date: {
+                            idate: true
+                        },
+                        storage_class: {
+                            $ref: '#/definitions/storage_class_enum'
+                        }
+                    }
+                },
+                noncurrent_version_expiration: {
+                    type: 'object',
+                    properties: {
+                        noncurrent_days: {
+                            type: 'integer'
+                        },
+                    }
+                },
+                noncurrent_version_transition: {
+                    type: 'object',
+                    properties: {
+                        noncurrent_days: {
+                            type: 'integer'
+                        },
+                        storage_class: {
+                            $ref: '#/definitions/storage_class_enum'
+                        }
+                    }
+                },
+                */
             }
         },
 
@@ -892,42 +1006,6 @@ module.exports = {
                 }
             },
         },
-        bucket_lifecycle_filter: {
-            type: 'object',
-            properties: {
-                prefix: {
-                    type: 'string'
-                },
-                // TODO: the functionality is not yet implemented so it is commented out
-                // object_size_greater_than: {
-                //     $ref: 'common_api#/definitions/bigint'
-                // },
-                // object_size_less_than: {
-                //     $ref: 'common_api#/definitions/bigint'
-                // },
-                // tag: {
-                //     // It is an Array for the use of And property
-                //     // Which includes all the filter properties and and array of tags
-                //     // It is not in the api as we dont need it, we will just parse it.
-                //     type: 'array',
-                //     items: {
-                //         $ref: '#/definitions/bucket_lifecycle_filter_tag'
-                //     },
-                // },
-            }
-        },
-        // TODO: the functionality is not yet implemented so it is commented out
-        // bucket_lifecycle_filter_tag: {
-        //     type: 'object',
-        //     required: ['key', 'value'],
-        //     properties: {
-        //         key: {
-        //             type: 'string'
-        //         },
-        //         value: {
-        //             type: 'string'
-        //         },
-        //     }
-        // }
+
     }
 };
--- a/src/endpoint/s3/ops/s3_get_bucket_lifecycle.js
+++ b/src/endpoint/s3/ops/s3_get_bucket_lifecycle.js
@@ -8,44 +8,50 @@ const _ = require('lodash');
  */
 async function get_bucket_lifecycle(req) {
     const reply = await req.object_sdk.get_bucket_lifecycle_configuration_rules({ name: req.params.bucket });
-    return {
-        LifecycleConfiguration: _.map(reply, rule => ({
-            Rule: [{
-                    ID: rule.id,
-                    Prefix: rule.prefix,
-                    Status: rule.status,
-                },
-                rule.transition ? {
-                    Transition: {
-                        Days: rule.transition.days,
-                        StorageClass: rule.transition.storage_class,
-                    }
-                } : {},
-                rule.expiration ? {
-                    Expiration: (rule.expiration.days ? {
-                        Days: rule.expiration.days,
-                        ExpiredObjectDeleteMarker: rule.expiration.expired_object_delete_marker ?
-                            rule.expiration.expired_object_delete_marker : {}
-                    } : {
-                        Date: rule.expiration.date,
-                        ExpiredObjectDeleteMarker: rule.expiration.expired_object_delete_marker ?
-                            rule.expiration.expired_object_delete_marker : {}
-                    })
-                } : {},
-                rule.noncurrent_version_transition ? {
-                    NoncurrentVersionTransition: {
-                        NoncurrentDays: rule.noncurrent_version_transition.noncurrent_days,
-                        StorageClass: rule.noncurrent_version_transition.storage_class,
-                    }
-                } : {},
-                rule.noncurrent_version_expiration ? {
-                    NoncurrentVersionExpiration: {
-                        NoncurrentDays: rule.noncurrent_version_expiration.noncurrent_days,
-                    }
-                } : {}
-            ]
-        }))
-    };
+    const rules = _.map(reply, rule => {
+        const current_rule = {
+            ID: rule.id,
+            Status: rule.status,
+        };
+
+
+        if (rule.filter.and) {
+            current_rule.Filter = {
+                And: [{
+                        Prefix: rule.filter.prefix,
+                        ObjectSizeGreaterThan: rule.filter.object_size_greater_than,
+                        ObjectSizeLessThan: rule.filter.object_size_less_than,
+                    },
+                    _.map(rule.filter.tags, tag => ({ Tag: { Key: tag.key, Value: tag.value } })),
+                ],
+            };
+        } else {
+            current_rule.Filter = {
+                Prefix: rule.filter.prefix,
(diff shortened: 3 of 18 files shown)


baum:
Closed [the initial configuration PR](https://github.com/noobaa/noobaa-core/pull/6901). That logic is included here.

src/server/bg_services/lifecycle.js
```diff
@@ -11,63 +11,113 @@ const system_store = require('../system_services/system_store').get_instance();
...
+        dbg.log0('LIFECYCLE SKIP bucket:', bucket.name, '(bucket id:', bucket._id, ') rule id', rule.id, 'rule', rule, 'now', now, 'last_sync', rule.last_sync, 'no expiration');
+        return;
+    }
+    dbg.log0('LIFECYCLE PROCESSING bucket:', bucket_rule);
+
+    await server_rpc.client.object.delete_multiple_objects_by_filter({
+        bucket: bucket.name,
+        create_time: get_expiration_timestamp(rule.expiration),
+        filter_delete_markers: get_expiration_filter_delete_markers(rule.expiration),
+        prefix: get_prefix(rule.filter),
```
dannyzaken, line 86: the getter seems unnecessary here. if it's just to check if `filter` is defined, you can either use optional chaining - `rule?.filter?.prefix` (this syntax should be supported in node 14) or just use `rule.filter && rule.filter.prefix`.
The same goes for the other getters

```suggestion
        prefix: rule?.filter?.prefix,
```
or 
```suggestion
        prefix: rule.filter && rule.filter.prefix,
```

src/server/bg_services/lifecycle.js
```diff
@@ -11,63 +11,113 @@ const system_store = require('../system_services/system_store').get_instance();
...
+    if (rule.expiration === undefined) {
+        dbg.log0('LIFECYCLE SKIP bucket:', bucket.name, '(bucket id:', bucket._id, ') rule id', rule.id, 'rule', rule, 'now', now, 'last_sync', rule.last_sync, 'no expiration');
+        return;
+    }
+    dbg.log0('LIFECYCLE PROCESSING bucket:', bucket_rule);
+
+    await server_rpc.client.object.delete_multiple_objects_by_filter({
+        bucket: bucket.name,
+        create_time: get_expiration_timestamp(rule.expiration),
+        filter_delete_markers: get_expiration_filter_delete_markers(rule.expiration),
```
dannyzaken, line 85: If I'm not mistaken, this does not match the expected behavior of that flag.
As we discussed, I think for now we can leave `expired_object_delete_marker` out of scope

src/server/object_services/md_store.js
```diff
@@ -447,6 +453,18 @@ class MDStore {
         let delete_marker;
         if (filter_delete_markers === true) delete_marker = null;
         else if (filter_delete_markers === false) delete_marker = true;
+        let size_filter;
```
dannyzaken, line 456: we need to test the performance of these queries and if new indexes are necessary

src/server/bg_services/lifecycle.js
```diff
@@ -11,63 +11,113 @@ const system_store = require('../system_services/system_store').get_instance();
...
+function get_size_greater_than(filter) {
+    return get_filter_field(filter, 'object_size_greater_than');
+}
+
+function get_tags(filter) {
+    return get_filter_field(filter, 'tags');
+}
+
+async function handle_bucket_rule(system, rule, j, bucket) {
+    dbg.log0('LIFECYCLE HANDLING BUCKET:', bucket.name, '(bucket id:', bucket._id, ') BEGIN');
```
dannyzaken, line 64: you added a lot of log0 messages. I'm not sure if all these log messages should be `dbg.log0`. We should be careful not to make our logs noisier than they already are.

src/server/object_services/md_store.js
baum: See [here](https://github.com/noobaa/noobaa-core/blob/cc184f7fbff7690bf8f31062b60d4d370b62df0f/src/server/object_services/md_store.js#L409)

src/server/bg_services/lifecycle.js
baum: 👍

src/server/bg_services/lifecycle.js
baum: 👍

src/server/bg_services/lifecycle.js
baum: 👍

src/endpoint/s3/ops/s3_get_bucket_lifecycle.js
```diff
@@ -8,44 +8,51 @@ const _ = require('lodash');
...
+                    const tag = rule.filter.tags[0];
+                    current_rule.Filter.Tag = {
+                        Key: tag.key,
+                        Value: tag.value
+                    };
+                }
+            }
+        }
+
+        if (rule.expiration) {
```
liranmauda, line 44: expiration is required by the api, do we need this if?

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -1,92 +1,123 @@
...
+function parse_filter(filter) {
+    const current_rule_filter = {};
+    if (filter.Tag && filter.Tag.length === 1) {
+        const tag = filter.Tag[0];
+        current_rule_filter.tags = [{ key: tag.Key[0], value: tag.Value[0] }];
+    }
+    if (filter.Prefix && filter.Prefix.length === 1) {
+        current_rule_filter.prefix = filter.Prefix[0];
+    }
+    if (filter.ObjectSizeGreaterThan &&
```
liranmauda, line 19: - Can we set ObjectSizeGreaterThan and ObjectSizeLessThan that is an empty group in s3?

src/api/common_api.js
```diff
@@ -138,6 +142,116 @@ module.exports = {
...
+                    type: 'string'
+                },
+                status: {
+                    $ref: 'common_api#/definitions/bucket_lifecycle_rule_status'
+                },
+                filter: {
+                    $ref: 'common_api#/definitions/bucket_lifecycle_rule_filter'
+                },
+                expiration: {
+                    $ref: 'common_api#/definitions/bucket_lifecycle_rule_expiration'
```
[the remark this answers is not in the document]
baum: 👍

src/endpoint/s3/ops/s3_get_bucket_lifecycle.js
```diff
@@ -8,44 +8,51 @@ const _ = require('lodash');
...
-                } : {}
-            ]
-        }))
-    };
+    const rules = _.map(reply, rule => {
+        const current_rule = {
+            ID: rule.id,
+            Status: rule.status,
+        };
+        if (rule.filter) {
```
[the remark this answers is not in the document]
baum: 👍, also validating on put

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
baum: It's a good idea to verify invalid size range

src/server/object_services/md_store.js
```diff
@@ -456,20 +474,20 @@ class MDStore {
                 $lt: new Date(moment.unix(max_create_time).toISOString()),
                 $exists: true
             } : undefined,
+            size: size_filter,
             upload_started: typeof upload_mode === 'boolean' ? {
                 $exists: upload_mode
             } : undefined,
+            tagging,
```
dannyzaken, line 481: I'm not sure this does what you intend it to do. you want to find all objects that has all of the given tags, but I think that this will search for a perfect match between the query tags and the object tags

baum: Thanks for pointing this out. See [Postgres client change](https://github.com/noobaa/noobaa-core/pull/6919/files#diff-c020c466408e4072550dbc75a3b7144063b7162ce224cdac0d3ef0d253acd794), and added test with an out-of-order subset of tags match.

dannyzaken: shouldn't that be changed to:

```suggestion
            tagging: {
               $all: tagging
            },
```

src/util/postgres_client.js
```diff
@@ -742,10 +742,27 @@ class PostgresTable {
...
     async find(query, options = {}) {
 
+        function isObject(v) {
+            return (typeof v === 'object' && !Array.isArray(v) && v !== null);
+        }
+        function calculateOptionsAndArrayFields(q) {
+            const ops = [ '$all' ];
+            const l = [];
+            // eslint-disable-next-line guard-for-in
+            for (const p in q) {
```
dannyzaken, line 752: in general `for in` is discouraged, hence the eslint rule. see here https://eslint.org/docs/rules/guard-for-in

instead, you can use this:
```suggestion
            for (const key of Object.keys(q)) {
```

src/util/postgres_client.js
```diff
@@ -742,10 +742,27 @@ class PostgresTable {
 
     async find(query, options = {}) {
 
+        function isObject(v) {
+            return (typeof v === 'object' && !Array.isArray(v) && v !== null);
+        }
+        function calculateOptionsAndArrayFields(q) {
+            const ops = [ '$all' ];
```
dannyzaken, line 749: the handling of `$all` operator should be in `handle_ops_encoding`. see for example how `$in` is handled.

src/server/object_services/md_store.js
baum: `$all` operator is applied [here](https://github.com/baum/noobaa-core/blob/45b242d714f4408b1c2558b40c1a8e04479d4a0a/src/server/object_services/object_server.js#L978-L980).

dannyzaken: I think we should have the DB related code in the md_store and not object_server

src/util/postgres_client.js
baum: 👍

src/util/postgres_client.js
baum: 👍

src/server/object_services/md_store.js
baum: Moved [here](https://github.com/baum/noobaa-core/blob/d068eea5c4ec7e54bcf19b656092203dd7672afe/src/server/object_services/md_store.js#L477-L479).

src/test/unit_tests/test_lifecycle.js
```diff
@@ -0,0 +1,222 @@
...
+            // go back in time
+            let create_time = new Date();
+            create_time.setDate(create_time.getDate() - age);
+            const update = {
+                create_time,
+            };
+            if (size) update.size = size;
+            if (tagging) update.tagging = tagging;
+
+            const putObjectPropertyResult = await rpc_client.object.put_object_property({ bucket, key, update });
```
dannyzaken, line 107: I don't think we should add RPC functions just for testing purposes. 
any reason to not call md_store directly? 

```suggestion
            await MDStore.instance().update_object_by_id(obj_id, update);
```

src/server/object_services/object_server.js
```diff
@@ -155,6 +155,27 @@ async function put_object_tagging(req) {
...
+ *
+ */
+async function put_object_property(req) {
+    throw_if_maintenance(req);
+    load_bucket(req);
+    const obj = await find_object_md(req);
+    const info = get_object_info(obj);
+
+    await MDStore.instance().update_object_by_id(
+        obj._id, req.rpc_params.update, undefined, undefined
```
dannyzaken, line 170: I suggested removing this function, but in general, you don't need to explicitly pass `undefined`

baum: 👍

Just note the style of MDStore.instance().update_object_by_id() references in object_server.js

[line 149-151](https://github.com/noobaa/noobaa-core/blob/cc184f7fbff7690bf8f31062b60d4d370b62df0f/src/server/object_services/object_server.js#L149-L151)
```js
    await MDStore.instance().update_object_by_id(
        obj._id, { tagging: req.rpc_params.tagging }, undefined, undefined
    );
```

[line 190-192](https://github.com/noobaa/noobaa-core/blob/cc184f7fbff7690bf8f31062b60d4d370b62df0f/src/server/object_services/object_server.js#L190-L192)
```js
    await MDStore.instance().update_object_by_id(
        obj._id, undefined, { 'tagging': 1 }, undefined
    );
```

[line 275-284](https://github.com/noobaa/noobaa-core/blob/cc184f7fbff7690bf8f31062b60d4d370b62df0f/src/server/object_services/object_server.js#L275-284)
```js
    await MDStore.instance().update_object_by_id(
        obj._id, {
            lock_settings: {
                legal_hold: {
                    status: req.params.legal_hold.status
                },
                retention,
            }
        }, undefined, undefined
    );
```

Those are the first three references and the list goes on...

src/test/unit_tests/test_lifecycle.js
baum: 👍

baum: Tried this one. This code causes a mongo specific RPC error throwing in mongo_client.js:

```
     Error: NO_SUCH_OBJECT
      at MongoClient.check_update_one (src/util/mongo_client.js:336:19)
      at MDStore.update_object_by_id (src/server/object_services/md_store.js:112:30)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
      at async create_mock_object (src/test/unit_tests/test_lifecycle.js:108:40)
      at async Context.<anonymous> (src/test/unit_tests/test_lifecycle.js:173:13)
```

dannyzaken: this is not mongo specific, it is coming from our code. 
in `update_object_by_id` we call `check_update_one`, which is different in mongo_client vs postgres_client. maybe there is something wrong with the obj id you pass. let me know if you need help debugging it.

src/server/object_services/object_server.js
dannyzaken: that is just a general comment, which is probably relevant to other places in the code :) 
in any case, this function should be removed, and there is no need to fix the code

dannyzaken (commented):
@baum update `LIFECYCLE_INTERVAL` to 12 hours in config.js

src/test/unit_tests/test_lifecycle.js
baum: I am looking into this issue. 


- Ran the test with postgres and all tests passed  - [postgres_lifecycle_index.log](https://github.com/noobaa/noobaa-core/files/8459406/postgres_lifecycle_index.log)
- Ran the test with mongo and some tests failed -
[mongo_lifecycle_index.log](https://github.com/noobaa/noobaa-core/files/8459402/mongo_lifecycle_index.log)

dannyzaken: @baum, the only thing the RPC adds is a server/client layer, which makes me suspect there is a timing issue here. We can test this theory by adding a small delay before `update_object_by_id`.
If the timing is indeed the issue, then It's worth examining if it is an issue in our code or something in the mongo stack. You can add log messages in the md_store right before calling the DB functions, to make sure they are executed in the correct order.

dannyzaken (approved):

liranmauda:
> @baum update `LIFECYCLE_INTERVAL` to 12 hours in config.js

I don't like 12 hours, it is a window and running twice a day this BG is not sufficient in my opinion.
Currently it is 15 min, and we can change that to be 1 or 2 hours.

IMO, I think that if we trigger a deletion of file ~12 hours after the expiration due, we will get rejects.

config.js
```diff
@@ -326,7 +326,7 @@ config.DEFAULT_S3_AUTH_METHOD = {
 // LIFECYCLE CONFIG //
 //////////////////////
 
-config.LIFECYCLE_INTERVAL = 15 * 60 * 1000;
+config.LIFECYCLE_INTERVAL = 12 * 60 * 60 * 1000; // 12h
```
liranmauda, line 329: I think we should make it interval more in a day, 1-2 hours maybe.

liranmauda (changes requested):

dannyzaken:
@liranmauda Alex and I discussed it, and 12 hours seems reasonable for a few reasons:
* the minimal time resolution of the lifecycle expiration is 1 day. there is no specific requirement to perform it in a specific hour 
* the lifecycle queries can be heavy when having a large number of object_mds. we won't necessarily want to add indexes to these queries, and running them less frequently helps us to achieve that. 

we want to merge this PR, and we can tune this delay later on if we see a reason to do that

dannyzaken:
@liranmauda @baum, let's set the delay to 8 hours

liranmauda (approved):

PR #6933, opened 2022-04-07
Adding minimal support for GPFS dynamic library

jackyalbo:

### Explain the changes
1.  As part of supporting GPFS Dynamic Library - adding functionality to be later used for encryption and SSE-S3

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -4,6 +4,7 @@
 #include "../util/napi.h"
 #include "../util/os.h"
 #include "../util/buf.h"
+#include "./gpfs_fcntl.h"
 
 #include <dirent.h>
 #include <fcntl.h>
@@ -33,6 +34,8 @@ namespace noobaa
 {
 
 DBG_INIT(0);
+
+static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg) = 0;
 
 const static std::map<std::string, int> flags_to_case = {
     { "r", O_RDONLY },
@@ -1384,6 +1387,18 @@ void
 fs_napi(Napi::Env env, Napi::Object exports)
 {
     auto exports_fs = Napi::Object::New(env);
+
+    const char* env_p = std::getenv("GPFS_DL_PATH");
+    if (env_p != NULL) {
+        LOG("FS::GPFS GPFS_DL_PATH=" << env_p);
+        uv_lib_t *lib = (uv_lib_t*) malloc(sizeof(uv_lib_t));
+        if (uv_dlopen(env_p, lib)) {
+            PANIC("Error: %s\n" << uv_dlerror(lib));
+        }
+        if (uv_dlsym(lib, "gpfs_fcntl", (void **) &dlsym_gpfs_fcntl)) {
+            PANIC("Error: %s\n" << uv_dlerror(lib));
+        }
+    }
 
     exports_fs["stat"] = Napi::Function::New(env, api<Stat>);
     exports_fs["lstat"] = Napi::Function::New(env, api<LStat>);
--- a/src/native/fs/gpfs.h
+++ b/src/native/fs/gpfs.h
@@ -1,0 +1,4183 @@
+/*                                                                              */
+/* Copyright (C) 2001 International Business Machines                           */
+/* All rights reserved.                                                         */
+/*                                                                              */
+/* This file is part of the GPFS user library.                                  */
+/*                                                                              */
+/* Redistribution and use in source and binary forms, with or without           */
+/* modification, are permitted provided that the following conditions           */
+/* are met:                                                                     */
+/*                                                                              */
+/*  1. Redistributions of source code must retain the above copyright notice,   */
+/*     this list of conditions and the following disclaimer.                    */
+/*  2. Redistributions in binary form must reproduce the above copyright        */
+/*     notice, this list of conditions and the following disclaimer in the      */
+/*     documentation and/or other materials provided with the distribution.     */
+/*  3. The name of the author may not be used to endorse or promote products    */
+/*     derived from this software without specific prior written                */
+/*     permission.                                                              */
+/*                                                                              */
+/* THIS SOFTWARE IS PROVIDED BY THE AUTHOR ``AS IS'' AND ANY EXPRESS OR         */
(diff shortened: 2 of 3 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -34,6 +35,9 @@ namespace noobaa
 
 DBG_INIT(0);
 
+int (*gpfs_fcntl)(gpfs_file_t fileDesc,
```
guymguym, line 38: Do you need to declare this yourself? Isn't it in the headers of the lib?

src/native/fs/fs_napi.cpp
```diff
@@ -1385,6 +1389,17 @@ fs_napi(Napi::Env env, Napi::Object exports)
 {
     auto exports_fs = Napi::Object::New(env);
 
+    uv_lib_t *lib = (uv_lib_t*) malloc(sizeof(uv_lib_t));
```
guymguym, line 1392: Define as null first and only malloc inside the condition?

src/native/fs/fs_napi.cpp
```diff
@@ -1385,6 +1389,17 @@ fs_napi(Napi::Env env, Napi::Object exports)
...
+    if(const char* env_p = std::getenv("GPFS_DL_PATH")) {
```
guymguym, line 1393: I prefer to take assignments out of the condition

guymguym (changes requested):

src/native/fs/fs_napi.cpp
jackyalbo: Will move both inside the condition

src/native/fs/fs_napi.cpp
```diff
@@ -1385,6 +1386,18 @@ fs_napi(Napi::Env env, Napi::Object exports)
...
     auto exports_fs = Napi::Object::New(env);
 
+    const char* env_p = std::getenv("GPFS_DL_PATH");
+    if (env_p != NULL) {
+        LOG("FS::GPFS GPFS_DL_PATH=" << env_p);
+        uv_lib_t *lib = (uv_lib_t*) malloc(sizeof(uv_lib_t));
+        if (uv_dlopen(env_p, lib)) {
+            PANIC("Error: %s\n" << uv_dlerror(lib));
+        }
+        if (uv_dlsym(lib, "gpfs_fcntl", (void **) &gpfs_fcntl)) {
```
guymguym, line 1396: OK, read this -
https://stackoverflow.com/questions/20321802/i-am-about-to-use-dlopen-to-open-shared-object-do-i-need-to-include-correspon

What I understand is that technically the exported functions from headers are not "used" when using dlsym because the addresses are not resolved during link phase. However if we use constants or structures then the header is still useful of course. The question is if we're doing the right thing when assigning to the gpfs_fcntl name here - does it really change the name from the header? Did you see an example that does this same pattern (update the function pointed by the header)?

guymguym: Adding @eshelmarc

guymguym: I guess you used this example -
http://docs.libuv.org/en/v1.x/guide/utilities.html#loading-libraries

guymguym: I wonder what actually happens with the current code.
Which pointer is updated by dlsym?

baum: @guymguym '[Which pointer is updated by dlsym?](https://github.com/noobaa/noobaa-core/pull/6933#discussion_r845147085)' I might be missing something however, [this](https://github.com/jackyalbo/noobaa-core/blob/ff417caf12679d72ae0ffa603afd080473ef1b8d/src/native/fs/fs_napi.cpp#L38) is the definition of a pointer to `gpfs_fcntl ` function **variable** and not a declaration. dlsym writes the pointer to the requested function to that variable.

eshelmarc: do you want to fail if library is not found or execute the functions in a way that don't need GPFS functions.

guymguym: > do you want to fail if library is not found or execute the functions in a way that don't need GPFS functions.

@eshelmarc I don't think it's necessary - this is why we condition it on the existence of the `GPFS_DL_PATH` env, so if you want to load it without the library you just omit that env, but if you provide it - you want to make sure it is loaded and used.

src/native/fs/fs_napi.cpp
```diff
@@ -34,6 +35,8 @@ namespace noobaa
 
 DBG_INIT(0);
 
+int (*gpfs_fcntl)(gpfs_file_t fileDesc, void* fcntlArgP);
```
guymguym, line 38: We probably want to declare this pointer static (file scoped), also better choose a name that doesnt exist in the gpfs header to avoid confusion, like dlsym_gpfs_fnctl.
```suggestion
static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg);
```

src/native/fs/fs_napi.cpp
guymguym: > guymguym 'Which pointer is updated by dlsym?' I might be missing something however, this is the definition of a pointer to function **variable** and not a declaration. dlsym writes the pointer to the requested function ( `gpfs_fcntl` ) to that variable.

@baum Yes I understan that but was asking because the same symbol name is also defined as a function in gpfs headers. Nevermind though, Jacky already answered that the code I commented on was not even compiling.

src/native/fs/fs_napi.cpp
jackyalbo: done

src/native/fs/fs_napi.cpp
```diff
@@ -34,6 +35,8 @@ namespace noobaa
 
 DBG_INIT(0);
 
+static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg);
```
guymguym, line 38: Oh we forgot to initialize this pointer to null -
```suggestion
static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg) = 0;
```

guymguym (approved):
Last comment on initialize the function pointer to 0 and LGTM

PR #6937, opened 2022-04-10
log based replication scheam changes

kfir-payne:

### Explain the changes
1. add new fields in the schema to support log based replication info
2. validate logs bucket exist
3. change bucket api in accordance

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/api/bucket_api.js
+++ b/src/api/bucket_api.js
@@ -1376,21 +1376,41 @@ module.exports = {
 
 
         replication_policy: {
-            type: 'array',
-            items: {
-                type: 'object',
-                required: ['destination_bucket', 'rule_id'],
-                properties: {
-                    destination_bucket: { $ref: 'common_api#/definitions/bucket_name' },
-                    rule_id: { type: 'string' },
-                    filter: {
+            type: 'object',
+            required: ['rules'],
+            properties: {
+                rules: {
+                    type: 'array',
+                    items: {
                         type: 'object',
+                        required: ['destination_bucket', 'rule_id'],
                         properties: {
-                            prefix: { type: 'string' },
-                            // s3 support also tag or and operator of 2 tags/ tag and prefix 
+                            destination_bucket: { $ref: 'common_api#/definitions/bucket_name' },
+                            rule_id: { type: 'string' },
+                            filter: {
+                                type: 'object',
+                                properties: {
+                                    prefix: { type: 'string' },
+                                    // s3 support also tag or and operator of 2 tags/ tag and prefix 
+                                }
+                            }
                         }
                     }
-                }
+                },
+                log_replication_info: {
+                    type: 'object',
+                    required: ['logs_location'],
+                    properties: {
+                        logs_location: {
+                            type: 'object',
+                            required: ['logs_bucket'],
+                            properties: {
+                                logs_bucket: { type: 'string' },
+                                prefix: { type: 'string' }
+                            }
+                        }
+                    }
+                },
             }
         },
     }
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -1932,10 +1932,20 @@ async function get_bucket_replication(req) {
     if (!replication_id) throw new RpcError('NO_REPLICATION_ON_BUCKET');
 
     const replication = await replication_store.instance().get_replication_by_id(replication_id);
-    const bucket_names_replication = _.map(replication, rule =>
-        ({ ...rule, destination_bucket: system_store.data.get_by_id(rule.destination_bucket).name }));
+    const bucket_names_replication = _.map(replication, rule => {
+        let named_rule = { ...rule, destination_bucket: system_store.data.get_by_id(rule.destination_bucket).name };
+        delete named_rule.rule_status;
+        return named_rule;
+    });
 
-    return bucket_names_replication;
+    const res = {
+        rules: bucket_names_replication
+    };
+    if (replication.aws_log_replication_info) {
+        res.log_replication_info.logs_location = replication.aws_log_replication_info.logs_location;
+    }
+
+    return res;
 }
 
 async function delete_bucket_replication(req) {
@@ -1957,7 +1967,7 @@ async function delete_bucket_replication(req) {
 }
 
 function validate_replication(req) {
-    const replication_rules = req.rpc_params.replication_policy;
+    const replication_rules = req.rpc_params.replication_policy.rules;
     // num of rules in configuration must be in defined limits
     if (replication_rules.length > config.BUCKET_REPLICATION_MAX_RULES ||
         replication_rules.length < 1) throw new RpcError('INVALID_REPLICATION_POLICY', 'Number of rules is invalid');
@@ -1977,6 +1987,7 @@ function validate_replication(req) {
         pref_by_dst_bucket[destination_bucket] = pref_by_dst_bucket[destination_bucket] || [];
         pref_by_dst_bucket[destination_bucket].push((filter && filter.prefix) || '');
     }
+
     // all rule_ids must be different
     if (new Set(rule_ids).size < replication_rules.length) throw new RpcError('INVALID_REPLICATION_POLICY', 'All rule ids must be unique');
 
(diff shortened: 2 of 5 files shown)


src/server/system_services/replication_store.js
```diff
@@ -84,7 +84,6 @@ class ReplicationStore {
         return ans;
     }
 
-
```
dannyzaken, line 87: remove change

src/api/bucket_api.js
```diff
@@ -1522,5 +1528,19 @@ module.exports = {
                 }
             }
         },
+
+        log_based_replication_info: {
```
dannyzaken, line 1532: maybe move `log_based_replication_info` into `replication policy` definition? they seem to always go together. 

Also, I think that `log_based_replication_info` is a bit of a mouthful. maybe rename it to `log_replication_info` or something?

src/server/system_services/bucket_server.js
```diff
@@ -1973,6 +1973,14 @@ function validate_replication(req) {
             throw new RpcError('INVALID_REPLICATION_POLICY', `Rule ${rule_id} destination bucket is invalid`);
         }
 
+        if (rule.log_based_replication_info) {
```
dannyzaken, line 1976: `log_based_replication_info` is not part of `rule` anymore. should be moved to out of the for loop

src/server/system_services/bucket_server.js
```diff
@@ -1977,6 +1977,25 @@ function validate_replication(req) {
...
+            req.system.buckets_by_name[req.rpc_params.replication_policy.log_replication_info.logs_location.logs_bucket.unwrap()];
+
+        // validate logs_bucket exist
+        if (!logs_bucket) {
+            throw new RpcError('INVALID_LOG_REPLICATION_INFO', `logs bucket is invalid`);
+        }
+
+        const logs_bucket_info = get_bucket_info({ bucket: logs_bucket });
+        if (logs_bucket_info.namespace && logs_bucket.namespace.write_resource.resource.connection.endpoint_type !== 'AWS') {
+            throw new RpcError('INVALID_LOG_REPLICATION_INFO', `logs bucket is not compatible with log based replication`);
```
romayalon, line 1992: we need to add here to the error message that logs_bucket should be a namespace bucket on AWS

src/server/system_services/bucket_server.js
```diff
@@ -1977,6 +1977,25 @@ function validate_replication(req) {
...
+        }
+        if (logs_bucket_info.mode !== 'OPTIMAL') {
```
romayalon, line 1994: if we are throwing on the operator a persistent error and rejecting, maybe for this case we shouldn't

src/server/system_services/bucket_server.js
```diff
@@ -1998,13 +2017,41 @@ function validate_replication(req) {
...
+    const validated_replication = { rules: validated_replication_rules };
+
+    if (req.rpc_params.replication_policy.log_replication_info) {
+        let validated_log_based_replication;
+        const logs_bucket = req.system.buckets_by_name &&
+            req.system.buckets_by_name[req.rpc_params.replication_policy.log_replication_info.logs_location.logs_bucket.unwrap()];
+        if (!logs_bucket) {
+            const source_bucket = find_bucket(req);
+            validated_log_based_replication = {
+                log_replication_info: { logs_location: { logs_bucket: source_bucket._id } }
```
romayalon, line 2037: why doing it?

src/server/system_services/bucket_server.js
```diff
@@ -1998,13 +2017,41 @@ function validate_replication(req) {
...
+            };
+        } else {
+            validated_log_based_replication = {
+                log_replication_info: { logs_location: { logs_bucket: logs_bucket._id } }
+            };
+        }
+        if (req.rpc_params.replication_policy.log_replication_info.logs_location.subpath !== "") {
```
romayalon, line 2044: why all this code is needed? 
I don't see why putting '/' for no subpath and '' for marker, keeping it empty won't work?

src/server/system_services/bucket_server.js
kfir-payne: worth to make a discussion about it, @dannyzaken WDYT?

src/server/system_services/bucket_server.js
kfir-payne: its one of the default behaviour we agreed on, if logs_bucket is not provided assume the logs are in the source bucket

romayalon: so add it as a comment please here

src/server/system_services/bucket_server.js
kfir-payne: putting '/' on the subpath it's just a default value if non provided, search the logs at the top of the bucket.
log marker can stay empty, planed to initiate it on the BG worker, and in the meanwhile put a placeholder in order to insert to the DB.

src/server/system_services/bucket_server.js
kfir-payne: will do

src/server/system_services/bucket_server.js
romayalon: Without '/' searching the logs will work? why do we need in the DB an empty marker? you can just check if it exists or not

kfir-payne: we could leave them empty and then when needed check existence and proceed as follows, I think that placing the default values now will save us time on always checking if they exist.

src/server/system_services/bucket_server.js
dannyzaken: I think we can skip this check for now

src/server/system_services/schemas/replication_configuration_schema.js
```diff
@@ -39,6 +39,25 @@ module.exports = {
...
                 }
+            },
+        },
+        log_replication_info: {
+            type: 'object',
+            required: ['logs_location'],
+            properties: {
+                logs_location: {
+                    type: 'object',
+                    required: ['logs_bucket', 'subpath'],
```
dannyzaken, line 50: I think `subpath` shouldn't be required. maybe rename it to prefix?

src/server/system_services/schemas/replication_configuration_schema.js
```diff
@@ -39,6 +39,29 @@ module.exports = {
...
+        aws_log_replication_info: {
+            type: 'object',
+            required: ['logs_location'],
+            properties: {
+                logs_location: {
+                    type: 'object',
+                    required: ['logs_bucket'],
+                    properties: {
+                        logs_bucket: {
+                            objectid: true // bucket id
```
dannyzaken, line 53: didn't we decide to access the logs will through the S3 bucket directly? 
of course, the user will have to provide a bucket that is accessible by the same account

kfir-payne: if remember correctly we decided that the logs bucket will also be a namespace bucket in Noobaa

src/server/system_services/bucket_server.js
```diff
@@ -1932,10 +1932,16 @@ async function get_bucket_replication(req) {
...
-    const bucket_names_replication = _.map(replication, rule =>
+    const bucket_names_replication = _.map(replication.rules, rule =>
         ({ ...rule, destination_bucket: system_store.data.get_by_id(rule.destination_bucket).name }));
 
-    return bucket_names_replication;
+    let { logs_location } = replication.aws_log_replication_info ? replication.aws_log_replication_info : undefined;
+    const res = {
+        rules: bucket_names_replication,
+        log_replication_info: logs_location
+    };
```
dannyzaken, line 1942: if `aws_log_replication_info` is undefined, I think your code will fail.
you can use optional chaining, which is supported in node 14, or you can just use a simple `if`

with optional chaining
```suggestion
    const res = {
        rules: bucket_names_replication
        log_replication_info: replication?.aws_log_replication_info?.logs_location
    }
```


or:
```suggestion
    const res = {
        rules: bucket_names_replication
    }
    if (replication.aws_log_replication_info) {
        res.log_replication_info = aws_log_replication_info.logs_location
    }
```

dannyzaken (approved):

PR #6940, opened 2022-04-12
print log message for long running db queries

dannyzaken:

### Explain the changes
1. if a query in Postgres is taking more than the defined threshold, print the query info to log.
2. added `config.DB_LOG_QUERY_THRESHOLD = 1000` as the threshold in ms

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -369,6 +369,9 @@ config.on_premise = {
     nva_part: "NVA_Upgrade.tgz"
 };
 
+// the threshold in ms for logging long running queries
+config.LONG_DB_QUERY_THRESHOLD = parseInt(process.env.LONG_DB_QUERY_THRESHOLD, 10) || 5000;
+
 /*
   Central Stats Collection & Diagnostics
 */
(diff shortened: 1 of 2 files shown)


config.js
```diff
@@ -369,6 +369,9 @@ config.on_premise = {
     nva_part: "NVA_Upgrade.tgz"
 };
 
+// the threshold in ms for logging long running queries
+config.DB_LOG_QUERY_THRESHOLD = 1000;
```
jackyalbo, line 373: Maybe this should be bigger? Did we check how much queries are bigger than 1 sec? I'm pretty sure that in minikube it will print a lot, not sure if in normal clusters.... So maybe we should disable for minikube or maybe set specific for it something bigger?

dannyzaken: I added an env to have a way to change it easily. 
I think that the majority of queries should not take longer than 1 sec. it is valuable to know of queries that do take longer than that, so we can fix them if needed. we can tune it later if we see it overloads the logs.
WDYT?

jackyalbo (approved):

PR #6946, opened 2022-04-14
Fix lifecycle size interval

baum:
Fix lifecycle size interval query.

Base PR #6919

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -453,18 +453,7 @@ class MDStore {
         let delete_marker;
         if (filter_delete_markers === true) delete_marker = null;
         else if (filter_delete_markers === false) delete_marker = true;
-        let size_filter;
-        if (max_size) {
-            size_filter = {
-                $lt: max_size,
-                $exists: true,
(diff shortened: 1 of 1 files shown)


src/server/object_services/md_store.js
```diff
@@ -453,19 +453,7 @@ class MDStore {
         let delete_marker;
         if (filter_delete_markers === true) delete_marker = null;
         else if (filter_delete_markers === false) delete_marker = true;
-        let size_filter;
-        if (max_size) {
-            size_filter = {
-                $lt: max_size,
-                $exists: true,
-            };
-        } else if (min_size) {
```
dannyzaken, line 462: I think that it's sufficient to remove the `else` and you can avoid the `$and`.

e.g.:
https://github.com/noobaa/noobaa-core/blob/efc509c9fd06df8be771ee6d87c893321da883d5/src/server/object_services/md_store.js#L811-L815

baum: 👍

dannyzaken (approved):

PR #6947, opened 2022-04-14
NSFS - Use temp file create+linkat to PUT an object in gpfs

eshelmarc:
Signed-off-by: Marc Eshel <eshel@us.ibm.com>

### Explain the changes
1. use temp files to create object for PUT

1. small file create are slow

### Testing Instructions:
1. set env variable that points to libgpfs
2. GPFS_DL_PATH={PATH}libgpfs.so

- [ ] Doc added/updated
- [ ] Tests added


src/native/fs/fs_napi.cpp
```diff
@@ -30,13 +32,68 @@
...
+// typedef struct gpfs_fssnap_handle gpfs_fssnap_handle_t;
+// typedef struct gpfs_ifile gpfs_ifile_t;
+// typedef unsigned int gpfs_ino_t;
+// typedef struct gpfs_iattr gpfs_iattr_t;
+
+// typedef int gpfs_file_t;
+// typedef struct gpfsFcntlHeader gpfsFcntlHeader_t;
+// typedef struct gpfsGetSetXAttr gpfsGetSetXAttr_t;
+
+#define ROUNDUP(X,Y) ( (Y) * ( ((X)+(Y)-1) / (Y) ) )
```
jackyalbo, line 47: I think we don't need most of this defines - probably even all of them. maybe relevant for the encryption I will push in the future

src/native/fs/fs_napi.cpp
```diff
@@ -30,13 +32,68 @@
...
+#define GPFS_FCNTL_ERR_ATTR_EXISTS         28
+#define GPFS_FCNTL_ERR_BUFFER_TOO_SMALL    29
+#define GPFS_FCNTL_ERR_NO_ATTR_SPACE       30
+
 namespace noobaa
 {
 
 DBG_INIT(0);
 
+gpfs_fssnap_handle_t *(*gpfs_get_fssnaphandle_by_path)(const char*);
```
jackyalbo, line 72: probably not needed for this flow

src/native/fs/fs_napi.cpp
```diff
@@ -30,13 +32,68 @@
...
+
+gpfs_ifile_t *(*gpfs_iopen)(gpfs_fssnap_handle_t *fssnapHandle,
```
jackyalbo, line 74: also not needed

src/native/fs/fs_napi.cpp
```diff
@@ -30,13 +32,68 @@
...
+    gpfs_ino_t ino,
+    int open_flags,
+    const gpfs_iattr_t *statxbuf,
+    const char *symLink);
+
+int (*gpfs_igetattrsx)(gpfs_ifile_t *ifile,
```
jackyalbo, line 80: not needed

src/native/fs/fs_napi.cpp
```diff
@@ -30,13 +32,68 @@
...
+    int flags,
+       void *buffer,
+       int bufferSize,
+       int *attrSize);
+
+int (*gpfs_fcntl)(gpfs_file_t fileDesc,
```
jackyalbo, line 86: old code - needs to be removed

src/native/fs/fs_napi.cpp
```diff
@@ -30,13 +32,68 @@
...
+           void* fcntlArgP);
+
 static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg) = 0;
 
+int (*gpfs_linkat)(gpfs_file_t fileDesc, const char *oldpath,
```
jackyalbo, line 91: probably also old code, needs to be using only one of those and probably the static ones like we did in gpfs_fcntl

src/native/fs/fs_napi.cpp
```diff
@@ -279,6 +337,47 @@ struct Stat : public FSWorker
     }
 };
 
+
+/**
+ * temp_file support by spectrumscale
+ */
+struct tempFile : public FSWorker
```
jackyalbo, line 344: Can't this be part of FileWrap or DirWrap? As we always do with operation interacting with file descriptors

src/sdk/namespace_fs.js
```diff
@@ -661,6 +661,46 @@ class NamespaceFS {
...
+                  } catch (e) {
+                      dbg.warn('NamespaceFS: COPY using link failed with:', e);
+                      fs_xattr = await this._copy_stream(source_file_path, upload_path, fs_account_config, fs_xattr);
+                  }
+              } else {
+                  // TODO: Take up only as much as we need (requires fine-tune of the semaphore inside the _upload_stream)
+                  // Currently we are taking config.NSFS_BUF_SIZE for any sized upload (1KB upload will take a full buffer from semaphore)
+                  fs_xattr = await buffers_pool_sem.surround_count(
+                      config.NSFS_BUF_SIZE,
+                      async () => this._upload_stream_tmp(params, this.bucket_path, file_path, fs_account_config,
```
jackyalbo, line 692: it seems like this is the only change between the flows, maybe the bug if in line 664 can move closer to this line and get less changes and smaller code

jackyalbo (changes requested):

src/native/fs/fs_napi.cpp
```diff
@@ -6,6 +6,8 @@
 #include "../util/buf.h"
 #include "./gpfs_fcntl.h"
 
+#include "./gpfs_fcntl.h"
```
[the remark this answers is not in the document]
eshelmarc: done

src/native/fs/fs_napi.cpp
```diff
@@ -30,13 +32,68 @@
     #include <sys/statfs.h>
 #endif
 
+// #define GPFS_ATTRFLAG_INCL_ENCR          0x0040  /* Include encryption attributes */
```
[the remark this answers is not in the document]
eshelmarc: done

src/native/fs/fs_napi.cpp
eshelmarc: done, I started from you code and hoped that you will use it when you finish encryption

src/native/fs/fs_napi.cpp
eshelmarc: done

src/native/fs/fs_napi.cpp
eshelmarc: done

src/native/fs/fs_napi.cpp
eshelmarc: done

src/native/fs/fs_napi.cpp
eshelmarc: is needed, and used in tempFile

src/native/fs/fs_napi.cpp
eshelmarc: it can but I think a function to get the fd will be useful in the future since most gpfs ops take a fd

src/sdk/namespace_fs.js
eshelmarc: there other changes, like spiking the need to check for upload (work) directory

src/native/fs/fs_napi.cpp
```diff
@@ -1387,6 +1467,18 @@ void
 fs_napi(Napi::Env env, Napi::Object exports)
 {
     auto exports_fs = Napi::Object::New(env);
+    uv_lib_t *lib = (uv_lib_t*) malloc(sizeof(uv_lib_t));
+
+    if(const char* env_p = std::getenv("GPFS_DL_PATH")) {
```
jackyalbo, line 1472: I think this all if is also reminders of my old code. we should stay only with the new if.

src/native/fs/fs_napi.cpp
jackyalbo: I think you should be using only this one: dlsym_gpfs_linkat (line 41)
As discussed in my previous PR, we should work with the static version (I think it's again mixes of old and new versions - before the fixes I made to my PR and after)

jackyalbo (changes requested):

src/native/fs/fs_napi.cpp
jackyalbo: will we be using an fd for files not needed to being opened and closed. If we don't then I think we should be using the wrapper, so it will help us to remember to close the file. @guymguym what do you think?

eshelmarc: ok, I will move it.

src/native/fs/fs_napi.cpp
eshelmarc: will fix

eshelmarc:
@guymguym @jackyalbo what is holding this patch?

guymguym:
Hey @eshelmarc I think the Cpp code is looking good, but I am a bit more concerned that the js side is duplicating too much code... I would like to propose a concrete direction on how to merge these paths by small refactoring, but haven't had the time yet. We can talk about it next week.

eshelmarc:
@guymguym as you know my js knowledge is limited, so I was trying to keep the changes simple, I don't mind if someone else will make those changes or if you have patience to show me exactly how to improve it that is fine too.

src/sdk/namespace_fs.js
```diff
@@ -877,6 +918,55 @@ class NamespaceFS {
         }
     }
 
+    // Allocated config.NSFS_BUF_SIZE in Semaphore but in fact we can take up more inside
+    // This is due to MD5 calculation and data buffers
+    // Can be finetuned further on if needed and inserting the Semaphore logic inside
+    // Instead of wrapping the whole _upload_stream function (q_buffers lives outside of the data scope of the stream)
+    async _upload_stream_tmp(params, upload_path, file_path, fs_account_config, rpc_client, fs_xattr) {
```
guymguym, line 925: @eshelmarc I suggest to merge `_upload_stream_tmp()`  with the existing `_upload_stream()` function, simply by adding another argument which we can name `use_tmpfile` and the callers will send as true/false depending on the path. Inside we can determine the open flags accordingly ('w' vs. 'wt') and also call the linkfileat function before closing (for that you seem to need to send also the file_path to the function which can be done for both call sites).

src/sdk/namespace_fs.js
```diff
@@ -877,6 +918,55 @@ class NamespaceFS {
...
+            return fs_xattr;
+        } catch (error) {
+            dbg.error('_upload_stream had error: ', error);
+            throw error;
+        } finally {
+            try {
+            dbg.log1('NamespaceFS.link_object:', target_file, upload_path);
+            if (typeof nb_native().fs.temp_file === 'function') {
+                await target_file.linkfileat(fs_account_config, upload_path, file_path);
+            }
```
guymguym, line 961: I think this should not be in the catch, but instead moved up into the try block, so that we will still close the file on failure to link.

src/sdk/namespace_fs.js
eshelmarc: done

eshelmarc:
@guymguym as you know my js knowledge is limited, so I was trying to keep the changes simple, I don't mind if someone else will make those changes or if you have patience to show me exactly how to improve it that is fine too.

src/sdk/namespace_fs.js
eshelmarc: BTW, you can not close the file before linking it, we use the fd to link the file.

src/sdk/namespace_fs.js
guymguym: right, it's inside a lower try block with the call to `close`, which means that it won't close if linkAt fails. But there's another try surrounding it, and there we are inside a finally block actually. So Let me resolve this thread, because the code moved with the merge of the functions. I will suggest what to move on the updated code in a moment.

guymguym:
> not sure what to do with DeepScan failure, it is not my code that is flaged.

@eshelmarc if you haven't then try rebasing from upstream -
```
git fetch --all
git rebase upstream/main
git push -f origin
```

src/sdk/namespace_fs.js
```diff
@@ -849,13 +892,18 @@ class NamespaceFS {
             throw error;
         } finally {
             try {
-                if (target_file) await target_file.close(fs_account_config);
+            dbg.log0('NamespaceFS.upload_stream:', open_mode, file_path, upload_path);
+            if (typeof nb_native().fs.temp_file === 'function' && open_mode == 'wt') {
+                await target_file.linkfileat(fs_account_config, upload_path, file_path);
+            }
```
guymguym, line 898: Cut these 3 lines (+one log line) and move it up right after the fsync call in line 887.
This way if linkfileat fails, our finally will still be called and we will close the fd (to discard it).

guymguym: BTW I would leave out this `typeof nb_native().fs.temp_file === 'function'` from here,
because open_mode=wt already checked it, and is less bloated/

src/native/fs/fs_napi.cpp
eshelmarc: not sure what was needed to be resolved

src/sdk/namespace_fs.js
eshelmarc: not sure if someone can use wt for other reason

eshelmarc:
I am not sure if I did the merge with the main tree correctly, I see another patch in in this PR, so let me know when the patch is ready how to merge it in.

eshelmarc:
not sure what is failing, it was working for my tests

src/sdk/namespace_fs.js
guymguym: and not link it? what kind of temp file can be closed without linking?

src/native/fs/fs_napi.cpp
guymguym: @jackyalbo Can you please go over your comments and see what can be resolved vs. what still requires fixing?

src/native/fs/fs_napi.cpp
```diff
@@ -1387,17 +1429,17 @@ void
 fs_napi(Napi::Env env, Napi::Object exports)
 {
     auto exports_fs = Napi::Object::New(env);
+    uv_lib_t *lib = (uv_lib_t*) malloc(sizeof(uv_lib_t));
```
guymguym, line 1432: @eshelmarc Why did you move this malloc outside of the env condition of GPFS_DL_PATH?
Is there a problem to keep it where it was?

src/native/fs/fs_napi.cpp
```diff
@@ -1387,17 +1429,17 @@ void
...
         LOG("FS::GPFS GPFS_DL_PATH=" << env_p);
-        uv_lib_t *lib = (uv_lib_t*) malloc(sizeof(uv_lib_t));
         if (uv_dlopen(env_p, lib)) {
             PANIC("Error: %s\n" << uv_dlerror(lib));
         }
-        if (uv_dlsym(lib, "gpfs_fcntl", (void **) &dlsym_gpfs_fcntl)) {
+        if (uv_dlsym(lib, "gpfs_linkat", (void **) &gpfs_linkat)) {
             PANIC("Error: %s\n" << uv_dlerror(lib));
         }
+        exports_fs["temp_file"] = Napi::Function::New(env, api<Linkat>);
```
guymguym, line 1442: Something is off here - when GPFS_DL_PATH we export the generic linkat action as a temp_file symbol ??
What's going on? Can we organize this a bit?

So the the Linkat struct is unrelated specifically to gpfs, it just uses the posix linkat function, and this is unused right now from the js code, because we don't have a use case for it.

The linkfileat function which is only implemented for gpfs is added to any file wrap.

I suggest:
1. Set the generic linkat function as exports_fs["linkat"].
2. Set a generic linkfileat function as a file wrap function, which will currently be implemented only for gpfs and panic otherwise.
3. The js code can check for the the existence of the `GPFS_DL_PATH` env and `fs_account_config.backend === 'GPFS'` to decide if it should use the tmpfile path.

src/sdk/namespace_fs.js
```diff
@@ -863,20 +907,25 @@ class NamespaceFS {
...
             // Used for etag
             return fs_xattr;
         } catch (error) {
             dbg.error('_upload_stream had error: ', error);
             throw error;
         } finally {
             try {
-                if (target_file) await target_file.close(fs_account_config);
+            if (target_file) await target_file.close(fs_account_config);
```
guymguym, line 921: bad indent, you can undo this now

src/sdk/namespace_fs.js
eshelmarc: if it was not WT than we don't need to link it and the logic is of the old _upload_stream()

src/sdk/namespace_fs.js
eshelmarc: fixed

src/native/fs/fs_napi.cpp
eshelmarc: no, moved it back

src/native/fs/fs_napi.cpp
eshelmarc: maybe in another PR ?

src/native/fs/fs_napi.cpp
```diff
@@ -991,6 +995,38 @@ struct FileSetxattr : public FSWrapWorker<FileWrap>
...
+    int _dirfd;
+    int _rc_res;
+
+
+    linkFileAt(const Napi::CallbackInfo& info)
+        : FSWrapWorker<FileWrap>(info)
+    {
+        _path = info[1].As<Napi::String>();
+        _filepath = info[2].As<Napi::String>();
+        _dirfd = _wrap->_fd;
```
guymguym, line 1011: wait, this is not a dir here, we are implementing this for FileWrap, meaning that _wrap->_fd is of the file that we opened, which is the temp file itself. So is it just an incorrect name of the variable? or did you expect to find the open directory fd here?

guymguym: also there is no need for you to copy it, you can just access _wrap->_fd inside Work()...

guymguym: You add at the end of the ctor this call like all other workers do:
```
        Begin(XSTR() << "LinkFileAt " << DVAL(_wrap->_path) << DVAL(_wrap->_fd) << DVAL(_filepath));
```

src/native/fs/fs_napi.cpp
```diff
@@ -991,6 +995,38 @@ struct FileSetxattr : public FSWrapWorker<FileWrap>
...
+    std::string _filepath;
+    int _dirfd;
+    int _rc_res;
+
+
+    linkFileAt(const Napi::CallbackInfo& info)
+        : FSWrapWorker<FileWrap>(info)
+    {
+        _path = info[1].As<Napi::String>();
+        _filepath = info[2].As<Napi::String>();
```
guymguym, line 1010: What is the difference between these? _path is the string provided to open, so for tempfile isn't that the same as the target path? or is there a case where I need to open with one path, but link it to another path?

src/native/fs/fs_napi.cpp
```diff
@@ -991,6 +995,38 @@ struct FileSetxattr : public FSWrapWorker<FileWrap>
...
+            SetSyscallError();
+            return;
+        }
+    }
+    virtual void OnOK()
+    {
+        DBG1("IBM::linkFileAt::OnOK: " << DVAL(_rc_res));
+        Napi::Env env = Env();
+        _deferred.Resolve(Napi::Number::New(env, _rc_res));
+    }
```
guymguym, line 1027: I think you should omit this entire function. We don't have anything to return as _rc_res was already checked and is 0...

src/native/fs/fs_napi.cpp
guymguym: I just mean to make a simple fix actually to just (in this scope we simply want to mark that we have loaded the library and that's it, nothing to do with the Linkat api, and nothing to do with temp_file).

```suggestion
        auto gpfs = Napi::Object::New(env);
        // for now we export an (empty) object, which can be checked to indicate that 
        // gpfs lib was loaded and its api's can be used.
        // e.g: gpfs["version"] = Napi::String::New(env,  gpfs_get_version());
        // e.g: gpfs["foo"] = Napi::Function::New(env, api<Foo>);
        gpfs.Freeze();
        exports_fs["gpfs"] = gpfs;
```

src/sdk/namespace_fs.js
```diff
@@ -682,7 +682,49 @@ class NamespaceFS {
         const fs_account_config = this.set_cur_fs_account_config(object_sdk);
         await this._load_bucket(params, fs_account_config);
 
+        if (typeof nb_native().fs.temp_file === 'function') {
```
guymguym, line 685: Based on the [this suggestion](https://github.com/noobaa/noobaa-core/pull/6947/files#r888402205), change this to:
```suggestion
        if (nb_native().fs.gpfs) {
```

guymguym: oh I see that you wanted this to represent a generic capability to use temp_file, instead of checking for backend type, but I prefer to keep most logic in the js world, and just provide a straightforward api to the native functions. if we need to support more backends with temp file we can fix it up here to allow more backends to use the same flow.

guymguym (changes requested):
@eshelmarc Hey, I posted another set of changes needed.
I think it will just take another round of changes after this (to deduplicate the remaining copied js code). Thanks

The one thing I wish we had is a test case for gpfs functions - we can't run this in our CI, but we can have it run on demand, perhaps @jackyalbo can initiate one in his next PR, and just load the library and call some gpfs specific functions.

src/native/fs/fs_napi.cpp
eshelmarc: _path was the directory name but we don't need it here, open uses the directory since we don't have a name, but link at gives that temp file the name.

src/native/fs/fs_napi.cpp
```diff
@@ -991,6 +995,38 @@ struct FileSetxattr : public FSWrapWorker<FileWrap>
...
+        _path = info[1].As<Napi::String>();
+        _filepath = info[2].As<Napi::String>();
+        _dirfd = _wrap->_fd;
+    }
+    virtual void Work()
+    {
+        _rc_res = gpfs_linkat(_dirfd, "", AT_FDCWD, _filepath.c_str(), AT_EMPTY_PATH);
+        if (_rc_res) {
+            DBG0("linkFileAt error FS::FD " << DVAL(_dirfd) << DVAL(_rc_res));
+            SetSyscallError();
```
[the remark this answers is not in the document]
eshelmarc: done

src/native/fs/fs_napi.cpp
eshelmarc: it is the file fd, I removed _dirfd

src/native/fs/fs_napi.cpp
eshelmarc: done

eshelmarc:
I made most of the change the to change exports_fs["temp_file"] did not work got the error ../src/native/fs/fs_napi.cpp:1440:14: error: ‘class Napi::Object’ has no member named ‘Freeze’

src/native/fs/fs_napi.cpp
guymguym: Ah right, I see the problem, please add this define here in [src/native/util/napi.h](https://github.com/noobaa/noobaa-core/blob/0a1c7746ad68621fa47b53fb53ecf6b7f9016442/src/native/util/napi.h#L5) right before including the node_api.h and this should include the support for Freeze():
```cpp
#define NAPI_VERSION 8
#include <node_api.h>
```

src/native/fs/fs_napi.cpp
```diff
@@ -991,6 +995,28 @@ struct FileSetxattr : public FSWrapWorker<FileWrap>
...
+    LinkFileAt(const Napi::CallbackInfo& info)
+        : FSWrapWorker<FileWrap>(info)
+    {
+        _filepath = info[2].As<Napi::String>();
+        Begin(XSTR() << "LinkFileAt " << DVAL(_wrap->_path) << DVAL(_wrap->_fd) << DVAL(_filepath));
+    }
+    virtual void Work()
+    {
+        // Link the Linux linkat() if there is no old path use empty string.
+        _rc_res = gpfs_linkat(_wrap->_fd, "", AT_FDCWD, _filepath.c_str(), AT_EMPTY_PATH);
```
guymguym, line 1012: _rc_res can be a local variable.

src/native/fs/fs_napi.cpp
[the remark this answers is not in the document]
guymguym: You left the SetSyscallError - is that intentional? Does gpfs function calls set errno correctly?

src/native/fs/fs_napi.cpp
guymguym: I see, but is there a reason not to send the target file name on open temp file ('wt')?
Is open expecting a dir name there?

src/native/fs/fs_napi.cpp
eshelmarc: done

src/native/fs/fs_napi.cpp
eshelmarc: yes, it must be the directory name

src/native/fs/fs_napi.cpp
eshelmarc: Every thing will work without the gpfs.Freeze(); should I delete it for now ?

guymguym: ok never mind, just comment that Freeze() call and I'll resolve it later.

src/native/fs/fs_napi.cpp
```diff
@@ -991,6 +995,28 @@ struct FileSetxattr : public FSWrapWorker<FileWrap>
...
+    LinkFileAt(const Napi::CallbackInfo& info)
+        : FSWrapWorker<FileWrap>(info)
+    {
+        _filepath = info[2].As<Napi::String>();
+        Begin(XSTR() << "LinkFileAt " << DVAL(_wrap->_path) << DVAL(_wrap->_fd) << DVAL(_filepath));
+    }
+    virtual void Work()
+    {
+        int _rc_res;
+        // Link the Linux linkat() if there is no old path use empty string.
```
guymguym, line 1011: can you explain in that comment the difference between the gpfs_linkat version and the linux linkat version?
The man page explains the same mechanism for linkat - https://man7.org/linux/man-pages/man2/link.2.html

>AT_EMPTY_PATH (since Linux 2.6.39)
              If oldpath is an empty string, create a link to the file
              referenced by olddirfd (which may have been obtained using
              the [open(2)](https://man7.org/linux/man-pages/man2/open.2.html) O_PATH flag).  In this case, olddirfd can
              refer to any type of file except a directory.  This will
              generally not work if the file has a link count of zero
              (files created with O_TMPFILE and without O_EXCL are an
              exception).  The caller must have the CAP_DAC_READ_SEARCH
              capability in order to use this flag.  This flag is Linux-
              specific; define _GNU_SOURCE to obtain its definition.

src/native/fs/fs_napi.cpp
guymguym: can you resolve this as well ^

src/native/fs/fs_napi.cpp
guymguym: ok, so is the _path argument still needed then? you can shift it to be arg number 1,
and call it like this - `await target_file.linkfileat(fs_account_config, file_path);`

src/native/fs/fs_napi.cpp
guymguym: I suggest to keep it like all the other code, and not name a local variable with underscore -
```
int r = gpfs_linkat(...);
if (r) ...
```

src/native/fs/fs_napi.cpp
```diff
@@ -45,6 +46,7 @@ const static std::map<std::string, int> flags_to_case = {
     { "rs+", O_RDWR | O_SYNC },
     { "sr+", O_RDWR | O_SYNC },
     { "w", O_TRUNC | O_CREAT | O_WRONLY },
+    { "wt", O_RDWR | __O_TMPFILE },
```
guymguym, line 49: why are we using `__O_TMPFILE` and not `O_TMPFILE`?
If there's a good reason, then we better explain that in a comment here.

src/native/fs/fs_napi.cpp
guymguym: yeah ok, just delete the call to Freeze() and keep the rest.
I'll take a look at how to enable the native freeze api in another PR.

guymguym: ah I see you did that already, so that's done. resolving.

src/sdk/namespace_fs.js
```diff
@@ -682,7 +682,50 @@ class NamespaceFS {
         const fs_account_config = this.set_cur_fs_account_config(object_sdk);
         await this._load_bucket(params, fs_account_config);
 
+        const can_use_temp_file = Boolean(nb_native().fs.gpfs);
```
guymguym, line 685: We want to allow mixing backends on the same system - see [backend enum](https://github.com/noobaa/noobaa-core/blob/0a1c7746ad68621fa47b53fb53ecf6b7f9016442/src/api/common_api.js#L941-L944)
Here's a suggestion with a comment to leave some reasons behind:

```js
// We define if a backend supports the tempfile+linkat flow,
// as every bucket can use a different fs backend type, we need to check 
// the type supports it, and for GPFS we check that the native library was loaded.
// We should update this as we add more backends that support this flow.
const can_use_temp_file = Boolean(fs_account_config.backend === 'GPFS' && nb_native().fs.gpfs);
```

src/native/fs/fs_napi.cpp
eshelmarc: done

src/native/fs/fs_napi.cpp
eshelmarc: Linux stat the file name and if it doesn't exist (which it will not, we create temp file with no name) it will fail the link call before calling the underling fs (in this case gpfs) that is the only reason we added the call, the interface is identical.

src/native/fs/fs_napi.cpp
eshelmarc: fixed

src/sdk/namespace_fs.js
eshelmarc: I don't know how to set fs_account_config.backend === 'GPFS' in my build, so I can not test it. I have a stand alone binary, and never got config to work.

src/native/fs/fs_napi.cpp
guymguym: I understand what you say, but the man for linkat says that with AT_EMPTY_PATH option it will allow an empty oldpath and then treat olddirfd as file fd, which is what we need, no?

src/sdk/namespace_fs.js
guymguym: It's in src/core/nsfs.js - see below.
You can get if from a flag (no need to define it, just add to usage to be nice).


https://github.com/noobaa/noobaa-core/blob/0a1c7746ad68621fa47b53fb53ecf6b7f9016442/src/core/nsfs.js#L81-L86

so you can write:
```js
  backend: argv.backend || '',
```

src/native/fs/fs_napi.cpp
eshelmarc: That is not the problem, Linux stat the file name of the name to be linked, it is a problem with Linux that the link operation don't get the to underline fs but we don't have an option here.

src/sdk/namespace_fs.js
```diff
@@ -837,11 +882,11 @@ class NamespaceFS {
     // This is due to MD5 calculation and data buffers
     // Can be finetuned further on if needed and inserting the Semaphore logic inside
     // Instead of wrapping the whole _upload_stream function (q_buffers lives outside of the data scope of the stream)
-    async _upload_stream(params, upload_path, fs_account_config, rpc_client, fs_xattr) {
+    async _upload_stream(params, upload_path, file_path, open_mode, fs_account_config, rpc_client, fs_xattr) {
```
guymguym, line 885: The unit tests failure is because you have another call to _upload_stream from upload_multipart() in this file that you need to fix.

```
2022-06-03T20:32:19.6190270Z 
2022-06-03T20:32:19.6196705Z   3319 passing (14m)
2022-06-03T20:32:19.6202078Z   78 pending
2022-06-03T20:32:19.6207777Z   1 failing
2022-06-03T20:32:19.6214785Z 
2022-06-03T20:32:19.6228031Z   1) namespace_fs
2022-06-03T20:32:19.6228432Z        multipart upload
2022-06-03T20:32:19.6228785Z          upload, read, delete a small multipart object:
2022-06-03T20:32:19.6229180Z      TypeError: Cannot convert undefined or null to object
2022-06-03T20:32:19.6229804Z       at NamespaceFS._upload_stream (src/sdk/namespace_fs.js:889:48)
2022-06-03T20:32:19.6230660Z       at /root/node_modules/noobaa-core/src/sdk/namespace_fs.js:973:34
2022-06-03T20:32:19.6231094Z       at Semaphore.surround_count (src/util/semaphore.js:90:90)
2022-06-03T20:32:19.6231516Z       at async NamespaceFS.upload_multipart (src/sdk/namespace_fs.js:971:30)
2022-06-03T20:32:19.6233647Z       at async Context.<anonymous> (src/test/unit_tests/test_namespace_fs.js:255:34)
```

src/native/fs/fs_napi.cpp
guymguym: Oh right right right, so the entire thing is because we need to support replacing an existing name, and linux linkat will not allow it to pass, correct?

eshelmarc: yes, I think I got it backward, it check that the file doesn't exist, and if it does it fails, and we need to also replace existing files.

guymguym: got it. so lets just update the comment to explain that reason.

src/sdk/namespace_fs.js
eshelmarc: fixed

src/sdk/namespace_fs.js
eshelmarc: fixed

src/native/fs/fs_napi.cpp
eshelmarc: done

guymguym: thanks!

src/native/fs/fs_napi.cpp
guymguym: in anycase, this change looks harmless, resolving.

src/sdk/namespace_fs.js
guymguym: can you add this change of nsfs.js to the PR?

eshelmarc: so just make it backend: 'GPFS',  ?

eshelmarc:
still failing, any ideas why?

src/sdk/namespace_fs.js
```diff
@@ -682,7 +682,50 @@ class NamespaceFS {
         const fs_account_config = this.set_cur_fs_account_config(object_sdk);
         await this._load_bucket(params, fs_account_config);
 
+        const can_use_temp_file = Boolean(fs_account_config.backend === 'GPFS' || nb_native().fs.gpfs);
```
guymguym, line 685: The tests fail because this needs to be &&, not ||.
And since some of our tests set backend type to GPFS this fails to upload because we don't have the library in CI.
(search 'GPFS' in src/test/unit_tests/test_bucketspace.js)

src/native/fs/fs_napi.cpp
```diff
@@ -45,6 +46,7 @@ const static std::map<std::string, int> flags_to_case = {
     { "rs+", O_RDWR | O_SYNC },
     { "sr+", O_RDWR | O_SYNC },
     { "w", O_TRUNC | O_CREAT | O_WRONLY },
+    { "wt", O_RDWR | O_TMPFILE },
```
guymguym, line 49: silly question, why do we need O_RDWR for tmpfile? isn't O_WRONLY enough?

src/sdk/namespace_fs.js
```diff
@@ -863,6 +908,10 @@ class NamespaceFS {
                 await target_file.setxattr(fs_account_config, fs_xattr);
             }
             if (config.NSFS_TRIGGER_FSYNC) await target_file.fsync(fs_account_config);
+            dbg.log1('NamespaceFS.upload_stream:', open_mode, file_path, upload_path);
+            if (open_mode === 'wt') {
+                await target_file.linkfileat(fs_account_config, file_path);
```
guymguym, line 913: almost forgot - we need to make path dirs for file_path, right before we call linkfileat:
```js
if (open_mode === 'wt') {
    await this._make_path_dirs(file_path, fs_account_config);
    await target_file.linkfileat(fs_account_config, file_path);
}
```

src/native/fs/fs_napi.cpp
eshelmarc: it is opened as temp, but it becomes a permanent file, you might want to read it ?

src/sdk/namespace_fs.js
eshelmarc: the directory must exist before the open, the link will not move it to another directory, so if the directory did not exist we should create it before open in _upload_stream()

src/sdk/namespace_fs.js
```diff
@@ -682,7 +682,8 @@ class NamespaceFS {
         const fs_account_config = this.set_cur_fs_account_config(object_sdk);
         await this._load_bucket(params, fs_account_config);
 
-        if (typeof nb_native().fs.temp_file === 'function') {
+        const can_use_temp_file = Boolean(fs_account_config.backend === 'GPFS' && nb_native().fs.gpfs);
```
eshelmarc, line 685: fs_account_config.backend === 'GPFS'  **is not working**, how do I get the backend value that was set in nsfs.js backend: backend || (process.env.GPFS_DL_PATH ? 'GPFS' : ''),

src/native/fs/fs_napi.cpp
eshelmarc: done

src/sdk/namespace_fs.js
guymguym: Oh, ok I mixed up. We take the backend only from the bucket configs, and not from the account config, which is why the previous attempt wouldn't work. We need to set the backend here:

https://github.com/noobaa/noobaa-core/blob/0a1c7746ad68621fa47b53fb53ecf6b7f9016442/src/core/nsfs.js#L126-L130

Just add another field called fs_backend like so:

```js
              fs_backend: argv.backend || (process.env.GPFS_DL_PATH ? 'GPFS' : ''),
              namespace_resource_id: undefined
        });
```

src/native/fs/fs_napi.cpp
guymguym: I see, and I just wonder - is there any benefit to use write-only when possible? or there shouldn't be much different to the filesystem? (any filesystem)

src/core/nsfs.js
```diff
@@ -75,13 +76,14 @@ async function main(argv = minimist(process.argv.slice(2))) {
         }
         const http_port = Number(argv.http_port) || 6001;
         const https_port = Number(argv.https_port) || 6443;
+        const backend = (argv.backend);
```
guymguym, line 79: you can use `argv.backend` inline so this doesn't seem to be needed

src/native/fs/fs_napi.cpp
```diff
@@ -1388,16 +1422,22 @@ fs_napi(Napi::Env env, Napi::Object exports)
...
-    const char* env_p = std::getenv("GPFS_DL_PATH");
-    if (env_p != NULL) {
-        LOG("FS::GPFS GPFS_DL_PATH=" << env_p);
+    if(const char* env_p = std::getenv("GPFS_DL_PATH")) {
         uv_lib_t *lib = (uv_lib_t*) malloc(sizeof(uv_lib_t));
+        LOG("FS::GPFS GPFS_DL_PATH=" << env_p);
         if (uv_dlopen(env_p, lib)) {
             PANIC("Error: %s\n" << uv_dlerror(lib));
         }
-        if (uv_dlsym(lib, "gpfs_fcntl", (void **) &dlsym_gpfs_fcntl)) {
```
guymguym, line 1398: You need to keep gpfs_fnctl because Jacky's next PR counts on it.
Just replicate this uv_dlsym call for linkat instead of replacing it.

src/core/nsfs.js
```diff
@@ -75,13 +76,14 @@ async function main(argv = minimist(process.argv.slice(2))) {
...
         const fs_root = argv._[0];
         if (!fs_root) return print_usage();
 
         let fs_config = {
             uid: Number(argv.uid) || process.getuid(),
             gid: Number(argv.gid) || process.getgid(),
-            backend: '',
+            backend: backend || (process.env.GPFS_DL_PATH ? 'GPFS' : ''),
```
guymguym, line 86: I wrote in the other thread but now attaching it to the code - the backend setting here is for the account, which is intentionally overridden by the bucket config, which comes from line 128 where we create new NamespaceFS, so we need to remove `backend` from here, and instead put it on new NamespaceFS as `fs_backend` like so:
```js
        const ns_fs = new NamespaceFS({
            fs_backend: argv.backend || (process.env.GPFS_DL_PATH ? 'GPFS' : ''),
            bucket_path: fs_root + '/' + bucket_name,
            bucket_id: '000000000000000000000000',
            namespace_resource_id: undefined,
        });
```

src/core/nsfs.js
guymguym: or better off set this with default already, so that the prints later will show the actual value:
```
const backend = argv.backend || (process.env.GPFS_DL_PATH ? 'GPFS' : '');
```

later when creating fs_config you can use the short field notation of `backend,` which stands for `backend: backend,` which is nice.

src/core/nsfs.js
guymguym: notice that I fixed this comment ^ to keep fs_config.backend and use it here, so that we can define the defaults once, in the main function, and then use the actual backend value inside the inner function init_request_sdk.

src/native/fs/fs_napi.cpp
```diff
@@ -35,7 +35,10 @@ namespace noobaa
 
 DBG_INIT(0);
 
-static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg) = 0;
```
guymguym, line 38: undelete dlsym_gpfs_fcntl

eshelmarc: I had this line in but testing failed because of unused define.

src/core/nsfs.js
eshelmarc: done

src/core/nsfs.js
eshelmarc: done

src/native/fs/fs_napi.cpp
eshelmarc: I don't think there is any benefit, and we might need read to compress, encrypt ...?

src/native/fs/fs_napi.cpp
eshelmarc: done

src/native/fs/fs_napi.cpp
eshelmarc: done

src/sdk/namespace_fs.js
eshelmarc: done

src/core/nsfs.js
```diff
@@ -75,13 +76,14 @@ async function main(argv = minimist(process.argv.slice(2))) {
...
         const https_port = Number(argv.https_port) || 6443;
+        const backend = argv.backend || (process.env.GPFS_DL_PATH ? 'GPFS' : '');
         const fs_root = argv._[0];
         if (!fs_root) return print_usage();
 
         let fs_config = {
             uid: Number(argv.uid) || process.getuid(),
             gid: Number(argv.gid) || process.getgid(),
-            backend: '',
+            backend: argv.backend || (process.env.GPFS_DL_PATH ? 'GPFS' : ''),
```
guymguym, line 86: no need to repeat, just use the const backend you defined before, with this shorthand syntax (this means `backend: backend,`) -
```suggestion
            backend,
```

src/native/fs/fs_napi.cpp
guymguym: ok got you. thanks for the explanation. resolving.

src/native/fs/fs_napi.cpp
guymguym: I assume it's not complaining on unused line now, right? resolving.

src/sdk/namespace_fs.js
guymguym: Let me try to understand this better - 
- What are the restrictions on path for `open(path, 'wt')`? 
- Does it need to be a directory or a file path?
- If the dir path of it has to exist, then we will have to call `_make_path_dirs` there too, which we currently don't have in the 'wt' flow at all, so I think we need to add. 
- Currently I see that we send the `bucket_path` (which is the root dir of the bucket) as the path to open tmp file. We only use the target `file_path` (which is the object key) for `linkfileat()`, so I am confused with your last comment that `link` will not move it to another directory.
- Do we have to provide the `file_path` to open instead of the `bucket_path`?
- We will need to call `_make_path_dirs` again before `linkfileat` to ensure that the dirs in `file_path` still exist after all the upload time has passed - we had a bug like this already in the rename flow that the dirs were deleted by the time we finished writing to the upload_path, and so we had to create those right before renaming to the target.

src/sdk/namespace_fs.js
```diff
@@ -682,7 +682,50 @@ class NamespaceFS {
...
+                                                      object_sdk.rpc_client, fs_xattr)
+                  );
+              }
+              // TODO use file xattr to store md5_b64 xattr, etc.
+              const stat = await nb_native().fs.stat(fs_account_config, file_path);
+              return { etag: this._get_etag(stat, fs_xattr) };
+          } catch (err) {
+              this.run_update_issues_report(object_sdk, err);
+             throw this._translate_object_error_codes(err);
+          }
```
guymguym, line 725: It seems to me like pulling out the try-catch from being copied in these two cases should be a simple way to dedup a bit.

guymguym: Also I see that you introduced a "unique" 2 spaces indentation there. Can you please fix it to 4 spaces across this code?

eshelmarc: done

src/core/nsfs.js
eshelmarc: done

src/sdk/namespace_fs.js
eshelmarc: The path for open is a directory path where the temp file will be created. We want to use the parent directory so we can inherent attributes from it. I don't think you will be able to delete the directory while a file is open in it, this different from creating a file in work directory and than rename it to a directory that might not exist anymore. I will try to fix it.

guymguym: so you mean that linkat will fail if the dir path differs from open - like this ? 
```
fd = open("dir1/", "wt");
linkat(fd, "", AT_FDCWD, "dir1/file", AT_EMPTY_PATH); // OK
linkat(fd, "", AT_FDCWD, "dir2/file", AT_EMPTY_PATH); // ERROR ?
```

guymguym: or should we provide `"dir1/file"` to open as well?

eshelmarc: temp open expects a directory path, in this case dir1 NOT dir1/file
the link will work with a different path, like dir2/file2 but it is better to create the temp file in the target directory so it can inherent the directory attributes.   
I will work on a fix to open the right director.

eshelmarc:
@guymguym let me know if there is anything else that I need to change

src/sdk/namespace_fs.js
guymguym: I think we still need to mkdirs here too because what if that dir was deleted/moved during our upload?

src/sdk/namespace_fs.js
```diff
@@ -914,13 +978,14 @@ class NamespaceFS {
 
     async upload_multipart(params, object_sdk) {
         try {
+            const open_mode = 'w';
```
guymguym, line 981: I would put a comment here that explains that we should switch to using `wt` here as well,
but it's definitely less of a priority because multipart uploads are large and the overhead of renaming a large part file is amortized in the entire upload of the part. However it still has a benefit in reducing rename contentions so we would like to close this todo at some point.

eshelmarc: done

src/sdk/namespace_fs.js
eshelmarc: the directory can not be deleted, if it is moved, which is very unlikely, the request will fail.

src/sdk/namespace_fs.js
```diff
@@ -837,11 +885,22 @@ class NamespaceFS {
...
-    async _upload_stream(params, upload_path, fs_account_config, rpc_client, fs_xattr) {
+    async _upload_stream(params, upload_path, file_path, open_mode, fs_account_config, rpc_client, fs_xattr) {
         let target_file;
         const { source_stream, md5_b64, key, bucket, upload_id } = params;
         try {
-            target_file = await nb_native().fs.open(fs_account_config, upload_path, 'w', get_umasked_mode(config.BASE_MODE_FILE));
+            if (open_mode === 'wt') {
+                var to = file_path.lastIndexOf('/');
+                to = to === -1 ? file_path.length : to;
+                var dir_path = file_path.substring(0, to);
```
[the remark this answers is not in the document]
eshelmarc: done

src/sdk/namespace_fs.js
```diff
@@ -682,44 +685,89 @@ class NamespaceFS {
...
-                    dbg.warn('NamespaceFS: COPY using link failed with:', e);
-                    fs_xattr = await this._copy_stream(source_file_path, upload_path, fs_account_config, fs_xattr);
+        const can_use_temp_file = Boolean(fs_account_config.backend === 'GPFS' && nb_native().fs.gpfs);
+        if (can_use_temp_file) {
+            const file_path = this._get_file_path(params);
+            const open_mode = 'wt';
+            // Uploading to a temp file then we will rename it.
+            let fs_xattr = to_fs_xattr(params.xattr);
+            // dbg.log0('NamespaceFS.upload_object:', this.upload_path, '->', file_path);
+            try {
```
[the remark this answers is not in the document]
eshelmarc: done

src/sdk/namespace_fs.js
```diff
@@ -509,7 +509,10 @@ class NamespaceFS {
             await this._load_bucket(params, fs_account_config);
             file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
             const stat = await file.stat(fs_account_config);
-            if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
+            // GPFS will create directories with attributes like ACLs and others, and expect them to be returned on get head.
+            if (fs_account_config.backend !== 'GPFS') {
```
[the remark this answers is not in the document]
eshelmarc: I will remove it and work on another PR for directory support needed by GPFS/AFM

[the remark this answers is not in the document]
guymguym: Created issue #6980 to track this.

src/sdk/namespace_fs.js
guymguym: OK, I wonder - since linkat is actually taking a target dirfd, which we always send as AT_FDCWD, perhaps we can open that dir and keep the dirfd while uploading, and at the end here, we will send dirfd+filename which will make it precisely refer to the opened target dir. WDYT?

src/sdk/namespace_fs.js
```diff
@@ -682,44 +682,83 @@ class NamespaceFS {
...
-                    if (same_inode) return same_inode;
-                    // Doing a hard link.
-                    await nb_native().fs.link(fs_account_config, source_file_path, upload_path);
-                } catch (e) {
-                    dbg.warn('NamespaceFS: COPY using link failed with:', e);
-                    fs_xattr = await this._copy_stream(source_file_path, upload_path, fs_account_config, fs_xattr);
+            if (can_use_temp_file) {
+                const open_mode = 'wt';
+                // Uploading to a temp file then we will rename it.
+                let fs_xattr = to_fs_xattr(params.xattr);
```
guymguym, line 691: This can be pushed up outside of `if (can_use_temp_file)`

src/sdk/namespace_fs.js
```diff
@@ -682,44 +682,83 @@ class NamespaceFS {
...
+                // dbg.log0('NamespaceFS.upload_object:', this.upload_path, '->', file_path);
+                if (params.copy_source) {
+                    const upload_id = uuidv4();
+                    const upload_path = path.join(this.bucket_path, this.get_bucket_tmpdir(), 'uploads', upload_id);
```
guymguym, line 695: This can be pushed up outside of `if (can_use_temp_file)`

src/sdk/namespace_fs.js
```diff
@@ -682,44 +682,83 @@ class NamespaceFS {
...
+                    const source_file_path = path.join(this.bucket_path, params.copy_source.key);
+                    await this._check_path_in_bucket_boundaries(fs_account_config, file_path);
```
guymguym, line 697: This can be pushed up outside of `if (can_use_temp_file)`

src/sdk/namespace_fs.js
guymguym: both upload_id and upload_path are just strings, so there is no problem to take these two lines also out of the if(copy_source) case and make it shared inside the try{} level.

src/sdk/namespace_fs.js
guymguym: This has to be called for all of our cases, both tmpfile and copy_source, so should end up in the try{} level.

src/sdk/namespace_fs.js
```diff
@@ -682,44 +682,83 @@ class NamespaceFS {
...
+                    await this._fail_if_archived_or_sparse_file(fs_account_config, source_file_path);
+                    try {
+                        const same_inode = await this._is_same_inode(fs_account_config, source_file_path, file_path, fs_xattr);
+                        if (same_inode) return same_inode;
+                        // Doing a hard link.
+                        await nb_native().fs.link(fs_account_config, source_file_path, upload_path);
+                    } catch (e) {
+                        dbg.warn('NamespaceFS: COPY using link failed with:', e);
+                        fs_xattr = await this._copy_stream(source_file_path, upload_path, fs_account_config, fs_xattr);
+                    }
```
guymguym, line 709: the remaining if(copy_source) block is exactly the same, so we should push it out of `if (can_use_temp_file)` to write it once in this flow.

src/sdk/namespace_fs.js
```diff
@@ -682,44 +682,83 @@ class NamespaceFS {
...
+                    // TODO: Take up only as much as we need (requires fine-tune of the semaphore inside the _upload_stream)
+                    // Currently we are taking config.NSFS_BUF_SIZE for any sized upload (1KB upload will take a full buffer from semaphore)
+                    fs_xattr = await buffers_pool_sem.surround_count(
+                        config.NSFS_BUF_SIZE,
+                        async () => this._upload_stream(params, this.bucket_path, file_path, open_mode, fs_account_config,
+                                                        object_sdk.rpc_client, fs_xattr)
+                    );
+                    // TODO use file xattr to store md5_b64 xattr, etc.
+                    const stat = await nb_native().fs.stat(fs_account_config, file_path);
+                    return { etag: this._get_etag(stat, fs_xattr) };
```
guymguym, line 720: The only difference in this ending is that we don't want to call _move_to_dest, but I think that fsync is missing here. I suggest to merge this with the existing ending of the else case, and just add a condition like `if(open_mode===wt)  _move_to_dest()....`

src/sdk/namespace_fs.js
eshelmarc: done

src/sdk/namespace_fs.js
eshelmarc: done

src/sdk/namespace_fs.js
eshelmarc: done

src/sdk/namespace_fs.js
eshelmarc: moved one more line

eshelmarc: had to undo some changes because of run time errors

eshelmarc:
@guymguym which Update branch should I use, Update with merge commit or  update with rebase ?

guymguym:
fetch and rebase from upstream:
```
git fetch --all
git rebase upstream/main
git push -f origin
```

eshelmarc:
</nas/eshel2-core/noobaa-core/src> git fetch --all
Fetching origin
</nas/eshel2-core/noobaa-core/src> git rebase upstream/main
fatal: Needed a single revision
invalid upstream 'upstream/main'

guymguym:
oops, sorry mixed up master/main, this repo still uses master:
```
git fetch --all
git rebase upstream/master
git push -f origin
```

eshelmarc:
@guymguym why is it failing, why can we merge it already?

romayalon:
@eshelmarc @guymguym 
I re-ran the workflow and now it passes

guymguym (approved):
@eshelmarc - I think the cpp code looks great. I still think that the js code could be reused instead of repeated - it doesn't seem needed to copy any of the code.

@nimrod-becker - In terms of quality - I believe we have a testing gap here that should be addressed - can you add this to discussion on how to add CI test coverage for the gpfs lib integration?

@jackyalbo - please complete your review as well.

@romayalon - In any case, I am ok with merging and improving in next PRs.

jackyalbo (approved):
I also agree that we can do more with code reuse in object_upload. but we can move to a following PR that will handle that

PR #6948, opened 2022-04-20
Deprecating the FE

liranmauda:
### Explain the changes

This PR is part of deprecating the FE.
- Cleaning the `lib.js` and the `lib.js.map` and also removing the FE test.
- Removing the entry of the FE from the `package.json`.
- Deleting the FE directory
- Cleaning the FE related code (mainly in web server and schemas)

There might be some leftovers... 
There is probably code that needs to be removed/changed on the operator side


guymguym (approved):
LGTM! 
I would also delete the `FE_PULL_REQUEST.md` and `SUPPORT.md` files from the root of the project.

guymguym:
FYI @nb-ohad 🥇

Issue #6950, opened 2022-05-03
PUT of file with md5 set occasionally reports TypeError: callback is not a function


MonicaLemay:
### Environment info
[root@c83f1-infa ~]# noobaa status
INFO[0000] CLI version: 5.9.0
INFO[0000] noobaa-image: quay.io/rhceph-dev/odf4-mcg-core-rhel8@sha256:8360f899f026691b9b7f8cea9a92f1589e0a7184c694ab843d61ffaa6262b49c
INFO[0000] operator-image: quay.io/rhceph-dev/odf4-mcg-rhel8-operator@sha256:eb56f52778b71a04b7df62db4ea241c65f84e2b80ddcc9b588414d173d7e086c
INFO[0000] noobaa-db-image: quay.io/rhceph-dev/rhel8-postgresql-12@sha256:be7212e938d1ef314a75aca070c28b6433cd0346704d0d3523c8ef403ff0c69e

[root@c83f1-infa ~]# oc version
Client Version: 4.9.29
Server Version: 4.9.29
Kubernetes Version: v1.22.5+a36406b
[root@c83f1-infa ~]#

[root@c83f1-infa ~]# oc get csv odf-operator.v4.9.6 -o json | grep full
            "full_version": "4.9.6-1",
### Actual behavior
This just showed up after upgrading to above mentioned SW levels.  In noobaa issue 6934, I was able to run warp tests with the S3 server configured for md5 and the warp set to md5=true.  I cannot now.

In this test that I'll show below, both the s3 server is set to MD5 enabled and the warp is set to md5=true.  The below mentioned problem does NOT occur if warp does not use md5=true.

Here is my warp launcher:
```
 ./warp mixed --host=172.20.100.24,172.20.100.23,172.20.100.22 --host-select=roundrobin --access-key=AMKwXLzOJEXzayywWPnO --secret-key=oUJLgVjRH7D0bvzHtoQ6zv8Bec6PQqLusVDlpz9D --obj.size=1G --duration=48h --bucket=s5001b7 --debug  --insecure  --tls --benchdata=/home/RW_moki/warp --warp-client=10.28.20.26 --obj.randsize --objects=1000 --concurrent=20 --md5=true
 ```

On the warp client that does the reading and writing, I see this frequently but not 100%.
```
warp: Sending benchmark_status
[...]

PR #6952, opened 2022-05-04
NSFS | chunk_fs.flush_buffers() missing callback argument

romayalon:

### Explain the changes
1. Added a check for if(callback) before using it.
2. Emitted an error to the stream when catching an error and no callback was provided.

Fixing one symptom seen in https://github.com/noobaa/noobaa-core/issues/6950
1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/chunk_fs.js
+++ b/src/util/chunk_fs.js
@@ -62,6 +62,8 @@ class ChunkFS extends stream.Transform {
         this._flush_buffers(callback);
     }
 
+    // callback will be passed only at the end of the stream by _flush()
+    // while this function is called without callback during _transform() and returns a promise.
     async _flush_buffers(callback) {
         try {
             if (this.q_buffers.length) {
@@ -78,7 +80,10 @@ class ChunkFS extends stream.Transform {
             }
         } catch (error) {
             console.error('ChunkFS _flush_buffers failed', this.q_size, this._total_num_buffers, error);
-            return callback(error);
+            if (callback) {
+                return callback(error);
+            }
+            throw error;
         }
     }
 }


src/util/chunk_fs.js
```diff
@@ -78,7 +80,10 @@ class ChunkFS extends stream.Transform {
             }
         } catch (error) {
             console.error('ChunkFS _flush_buffers failed', this.q_size, this._total_num_buffers, error);
-            return callback(error);
+            if (callback) {
+                return callback(error);
+            }
+            this.emit('error', error);
```
guymguym, line 86: This is called from _transform and is awaited for.
So I think I would expect that when this function does not receive a callback, it should be expected to return a promise rejection. Since this is async function - we can simply throw. 
```suggestion
            throw error;
```

guymguym (dismissed):

guymguym (changes requested):
see comment on throw error

src/util/chunk_fs.js
romayalon: ok, fixed!

src/util/chunk_fs.js
```diff
@@ -62,6 +62,8 @@ class ChunkFS extends stream.Transform {
         this._flush_buffers(callback);
     }
 
+    // callback will be passed only at the end of the stream
+    // while this function can be called sooner without callback
```
guymguym, line 66: just a minor comment clarification
```suggestion
    // callback will be passed only at the end of the stream by _flush()
    // while this function is called without callback during _transform() and returns a promise.
```

guymguym (approved):

guymguym:
@romayalon Lets merge this one too

PR #6955, opened 2022-05-09
NSFS | Add checks for request/response aborted in namespace_fs read_object_stream() 

romayalon:

### Explain the changes
1. Object_sdk - Added an abort controller property and added 2 functions:
   1.1. setup_abort_controller() - setup event listeners on the req/res that will call abort() of the abort_controller.
   1.2. throw_if_aborted() - will check if the abort_controller got the aborted signal and will throw error when true.
2. S3_get_object -
   2.1. Called setup_abort_controller() in order to hang the event listeners on each get_object request/response.
   2.2. Removed the previous event listeners code.
3. Namespace FS - 
   3.1. Added a call to object_sdk.throw_if_aborted() after each await in read_object_stream(). (see Note)
   3.2. Passed object_sdk.abort_controller.signal parameter to wait_finished() and to wait_drain().

NOTE - 
The whole purpose of this new abort_controller functionality was to fix the following situation -
1. A Client sends s3_get_object request, NooBaa is writing a buffer to the response, and the client reads it very slowly.
2. NooBaa in the meantime reads the next data from the fs and holds a new buffer, but NooBaa is not yet writing it to the socket and is awaiting for wait_drain() (because the client reads the data very slowly).
3. The user calls abort on the request, and the first buffer is released.
4. NooBaa still awaits on wait_drain() and therefore still holding the second buffer.

1. Fixed #https://github.com/noobaa/noobaa-core/issues/6934 hopefully

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/core/nsfs.js
+++ b/src/core/nsfs.js
@@ -93,45 +93,11 @@ async function main(argv = minimist(process.argv.slice(2))) {
         console.warn(WARNINGS);
         console.log('nsfs: setting up ...', { fs_root, http_port, https_port });
 
-        const noop = /** @type {any} */ () => {
-            // TODO
-        };
-
-        const bs = new BucketSpaceFS({ fs_root });
-        const object_sdk = new ObjectSDK(null, null, null);
-
-        // resolve namespace and bucketspace
-        const namespaces = {};
-        object_sdk._get_bucketspace = () => bs;
-        object_sdk._get_bucket_namespace = async bucket_name => {
-            const existing_ns = namespaces[bucket_name];
-            if (existing_ns) return existing_ns;
-            const ns_fs = new NamespaceFS({
-                bucket_path: fs_root + '/' + bucket_name,
-                bucket_id: '000000000000000000000000',
-                namespace_resource_id: undefined
-            });
-            namespaces[bucket_name] = ns_fs;
-            return ns_fs;
-        };
-
-        object_sdk.get_auth_token = noop;
-        object_sdk.set_auth_token = noop;
-        object_sdk.authorize_request_account = noop;
-        object_sdk.read_bucket_sdk_website_info = noop;
-        object_sdk.read_bucket_sdk_namespace_info = noop;
-        object_sdk.read_bucket_sdk_caching_info = noop;
-        object_sdk.read_bucket_sdk_policy_info = noop;
-        object_sdk.read_bucket_usage_info = noop;
-        object_sdk.requesting_account = {
-            nsfs_account_config: fs_config
-        };
-
         const endpoint = require('../endpoint/endpoint');
         await endpoint.start_endpoint({
             http_port,
             https_port,
-            init_request_sdk: (req, res) => { req.object_sdk = object_sdk; },
+            init_request_sdk: (req, res) => init_request_sdk(req, res, fs_root, fs_config),
         });
 
         console.log('nsfs: listening on', util.inspect(`http://localhost:${http_port}`));
@@ -141,6 +107,43 @@ async function main(argv = minimist(process.argv.slice(2))) {
         console.error('nsfs: exit on error', err.stack || err);
         process.exit(2);
     }
+}
+
+function init_request_sdk(req, res, fs_root, fs_config) {
+    const noop = /** @type {any} */ () => {
+        // TODO
+    };
+
+    const bs = new BucketSpaceFS({ fs_root });
+    const object_sdk = new ObjectSDK(null, null, null);
+
+    // resolve namespace and bucketspace
+    const namespaces = {};
+    object_sdk._get_bucketspace = () => bs;
+    object_sdk._get_bucket_namespace = async bucket_name => {
+        const existing_ns = namespaces[bucket_name];
+        if (existing_ns) return existing_ns;
+        const ns_fs = new NamespaceFS({
+            bucket_path: fs_root + '/' + bucket_name,
+            bucket_id: '000000000000000000000000',
+            namespace_resource_id: undefined
+        });
+        namespaces[bucket_name] = ns_fs;
+        return ns_fs;
+    };
+
+    object_sdk.get_auth_token = noop;
+    object_sdk.set_auth_token = noop;
+    object_sdk.authorize_request_account = noop;
+    object_sdk.read_bucket_sdk_website_info = noop;
+    object_sdk.read_bucket_sdk_namespace_info = noop;
+    object_sdk.read_bucket_sdk_caching_info = noop;
+    object_sdk.read_bucket_sdk_policy_info = noop;
+    object_sdk.read_bucket_usage_info = noop;
+    object_sdk.requesting_account = {
+        nsfs_account_config: fs_config
+    };
+    req.object_sdk = object_sdk;
 }
 
 exports.main = main;
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -11,6 +11,7 @@ const http_utils = require('../../../util/http_utils');
  */
 async function get_object(req, res) {
 
+    req.object_sdk.setup_abort_controller(req, res);
     const agent_header = req.headers['user-agent'];
     const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
     const encryption = s3_utils.parse_encryption(req);
@@ -100,20 +101,13 @@ async function get_object(req, res) {
         }
     }
 
-    let read_stream;
-
-    // on http disconnection close the read stream to stop from buffering more data
-    req.on('aborted', () => {
-        dbg.log0('request aborted:', req.path);
-        if (read_stream && read_stream.close) read_stream.close();
-    });
-    res.on('error', err => {
-        dbg.log0('response error:', err, req.path);
-        if (read_stream && read_stream.close) read_stream.close();
-    });
-
-    read_stream = await req.object_sdk.read_object_stream(params, res);
+    const read_stream = await req.object_sdk.read_object_stream(params, res);
     if (read_stream) {
+        // if read_stream supports closing, then we handle abort cases such as http disconnection
+        // by calling the close method to stop it from buffering more data which will go to waste.
+        if (read_stream.close) {
+            req.object_sdk.add_abort_handler(() => read_stream.close());
+        }
         read_stream.on('error', err => {
             dbg.log0('read stream error:', err, req.path);
             res.destroy(err);
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -540,13 +540,15 @@ class NamespaceFS {
     }
     */
 
+    // eslint-disable-next-line max-statements
     async read_object_stream(params, object_sdk, res) {
         let file;
         let buffer_pool_cleanup = null;
         const fs_account_config = this.set_cur_fs_account_config(object_sdk);
+        let file_path;
         try {
             await this._load_bucket(params, fs_account_config);
-            const file_path = this._get_file_path(params);
+            file_path = this._get_file_path(params);
             await this._check_path_in_bucket_boundaries(fs_account_config, file_path);
             await this._fail_if_archived_or_sparse_file(fs_account_config, file_path);
             file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
@@ -563,13 +565,17 @@ class NamespaceFS {
 
             let count = 1;
             for (let pos = start; pos < end;) {
+                object_sdk.throw_if_aborted();
 
                 // allocate or reuse buffer
-                const remain_size = Math.max(0, end - pos);
+                // TODO buffers_pool and the underlying semaphore should support abort signal
+                // to avoid sleeping inside the semaphore until the timeout while the request is already aborted.
                 const { buffer, callback } = await buffers_pool.get_buffer();
-                buffer_pool_cleanup = callback;
+                buffer_pool_cleanup = callback; // must be called ***IMMEDIATELY*** after get_buffer
+                object_sdk.throw_if_aborted();
 
                 // read from file
+                const remain_size = Math.max(0, end - pos);
                 const read_size = Math.min(buffer.length, remain_size);
 
                 // Update the read stats               
@@ -588,6 +594,7 @@ class NamespaceFS {
                     callback();
                     break;
                 }
+                object_sdk.throw_if_aborted();
                 const data = buffer.slice(0, bytesRead);
 
                 // update stats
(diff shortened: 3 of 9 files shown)


src/test/unit_tests/test_namespace_fs.js
```diff
@@ -31,6 +32,16 @@ const DEFAULT_FS_CONFIG = {
...
+function get_dummy_object_sdk() {
+    let dummy_object_sdk = { requesting_account: { nsfs_account_config: { uid: process.getuid(), gid: process.getgid() } },
+        abort_controller: new AbortController()
+    };
+    dummy_object_sdk.throw_if_aborted = () => {
+        if (dummy_object_sdk.abort_controller.signal.aborted) throw new Error('request aborted signal');
+    };
+    return dummy_object_sdk;
+}
+
```
guymguym, line 44:
```suggestion
function make_dummy_object_sdk() {
    return {
        requesting_account: {
            nsfs_account_config: {
                uid: process.getuid(),
                gid: process.getgid(),
            }
        },
        abort_controller: new AbortController(),
        throw_if_aborted() {
            if (this.abort_controller.signal.aborted) throw new Error('request aborted signal');
        }
    };
}

```

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -100,20 +101,12 @@ async function get_object(req, res) {
...
-        dbg.log0('response error:', err, req.path);
-        if (read_stream && read_stream.close) read_stream.close();
-    });
-
-    read_stream = await req.object_sdk.read_object_stream(params, res);
+    const read_stream = await req.object_sdk.read_object_stream(params, res);
     if (read_stream) {
+        // on http disconnection close the read stream to stop from buffering more data
+        if (read_stream.close) {
+            req.object_sdk.abort_controller.signal.onabort = () => read_stream.close();
```
guymguym, line 108: Don't override the onabort function, instead add an event handler to allow others to listen for it as well -
```suggestion
            req.object_sdk.abort_controller.signal.once('abort', () => read_stream.close());
```

src/sdk/object_sdk.js
```diff
@@ -64,6 +65,28 @@ class ObjectSDK {
             this.bucketspace_fs = new BucketSpaceFS({ fs_root: process.env.BUCKETSPACE_FS });
         }
         this.requesting_account = undefined;
+        this.abort_controller = new AbortController();
+    }
+
+    setup_abort_controller(req, res) {
```
guymguym, line 71:
```suggestion
    /**
     * setup_abort_controller adds event handlers to the http request and response,
     * in order to handle aborting requests gracefully. The `abort_controller` member will
     * be used to signal async flows that abort was detected.
     * @see {@link https://nodejs.org/docs/latest/api/globals.html#class-abortcontroller}
     * @param {import('http').IncomingMessage} req 
     * @param {import('http').ServerResponse} res 
     */
    setup_abort_controller(req, res) {
```

src/sdk/namespace_fs.js
```diff
@@ -563,11 +564,13 @@ class NamespaceFS {
 
             let count = 1;
             for (let pos = start; pos < end;) {
+                object_sdk.throw_if_aborted();
 
                 // allocate or reuse buffer
                 const remain_size = Math.max(0, end - pos);
                 const { buffer, callback } = await buffers_pool.get_buffer();
                 buffer_pool_cleanup = callback;
```
guymguym, line 572:
```suggestion
                buffer_pool_cleanup = callback; // must be called ***IMMEDIATELY*** after get_buffer
```

guymguym (changes requested):

src/endpoint/s3/ops/s3_get_object.js
romayalon: Added the event_listener as we discussed on DM

src/sdk/namespace_fs.js
```diff
@@ -635,15 +645,18 @@ class NamespaceFS {
             return null;
 
         } catch (err) {
+            dbg.log0('NamespaceFS: read_object_stream error', err);
             throw this._translate_object_error_codes(err);
 
         } finally {
             try {
+                dbg.log0('NamespaceFS: read_object_stream finally');
```
guymguym, line 653: If you print this specific message at least do it inside the if(file). Same for if(buffer_pool_cleanup).

src/sdk/object_sdk.js
```diff
@@ -64,6 +65,41 @@ class ObjectSDK {
...
+            this.abort_controller.abort();
+        });
+    }
+
+    throw_if_aborted() {
+        if (this.abort_controller.signal.aborted) throw new Error('request aborted signal');
+    }
+
+    add_abort_handler(handler) {
+        if (this.abort_controller.signal.aborted) return;
```
guymguym, line 101: Instead of return you want to call setUmmediate(handler)

guymguym (changes requested):

src/sdk/namespace_fs.js
```diff
@@ -635,17 +645,24 @@ class NamespaceFS {
...
 
         } catch (err) {
+            dbg.log0('NamespaceFS: read_object_stream error', err);
             throw this._translate_object_error_codes(err);
 
         } finally {
             try {
-                if (file) await file.close(fs_account_config);
+                if (file) {
+                    dbg.log0('NamespaceFS: read_object_stream finally');
```
guymguym, line 654:
```suggestion
                    dbg.log0('NamespaceFS: read_object_stream finally closing file', file_path);
```

src/sdk/namespace_fs.js
```diff
@@ -635,17 +645,24 @@ class NamespaceFS {
             return null;
 
         } catch (err) {
+            dbg.log0('NamespaceFS: read_object_stream error', err);
```
guymguym, line 648:
```suggestion
            dbg.log0('NamespaceFS: read_object_stream error file', file_path, err);
```

src/sdk/namespace_fs.js
```diff
@@ -602,27 +606,33 @@ class NamespaceFS {
...
                     drain_promise = null;
+                    object_sdk.throw_if_aborted();
                 }
 
                 // write the data out to response
                 buffer_pool_cleanup = null; // cleanup is now in the socket responsibility
                 const write_ok = res.write(data, null, callback);
                 if (!write_ok) {
-                    drain_promise = stream_utils.wait_drain(res);
+                    drain_promise = stream_utils.wait_drain(res, object_sdk.abort_controller.signal);
```
guymguym, line 616: on second thought, we better make wait_drain follow the same api as wait_finished and send an options object with a signal property...

src/sdk/namespace_fs.js
```diff
@@ -563,11 +564,13 @@ class NamespaceFS {
 
             let count = 1;
             for (let pos = start; pos < end;) {
+                object_sdk.throw_if_aborted();
 
                 // allocate or reuse buffer
                 const remain_size = Math.max(0, end - pos);
```
guymguym, line 570: not sure how this line got up here, we can push it a bit lower next to the `const read_size = ...`

src/sdk/namespace_fs.js
```diff
@@ -563,11 +564,13 @@ class NamespaceFS {
...
                 const { buffer, callback } = await buffers_pool.get_buffer();
```
guymguym, line 571:
```suggestion
                // TODO buffers_pool and the underlying semaphore should support abort signal
                // to avoid sleeping inside the semaphore until the timeout while the request is already aborted.
                const { buffer, callback } = await buffers_pool.get_buffer();
```

src/sdk/namespace_fs.js
```diff
@@ -602,27 +606,33 @@ class NamespaceFS {
...
+
             // wait for the last drain if pending.
             if (drain_promise) {
                 await drain_promise;
                 drain_promise = null;
+                object_sdk.throw_if_aborted();
             }
 
             // end the stream
             res.end();
```
guymguym, line 632: I wonder - doesn't `res.end()` emit the 'close' even on the response? If so then would our code in object_sdk.setup_abort_controller that registers on the 'close' event might not work correctly and cause our next "throw_if_aborted" (the one after wait_finished) to fail? BTW notice that originally we did not register on the 'close' event of the response (in the original code in s3_get_object), and during our testing we decided to add it because we thought it was missing - perhaps it was not missing and now that we register on request 'aborted' we are covered for that case. Need to validate it. If this is not right, then I would suggest to remove that last throw_if_aborted, and then check if wait_finished completes fine before close is emitted.

src/sdk/namespace_fs.js
```diff
@@ -602,27 +606,33 @@ class NamespaceFS {
...
-            await stream_utils.wait_finished(res);
+
+            await stream_utils.wait_finished(res, { signal: object_sdk.abort_controller.signal });
+            object_sdk.throw_if_aborted();
 
             dbg.log0('NamespaceFS: read_object_stream completed', {
```
guymguym, line 637: add file_path to this print, not sure why it's not here

src/sdk/namespace_fs.js
```diff
@@ -635,17 +645,24 @@ class NamespaceFS {
...
+                    await file.close(fs_account_config);
+                }
             } catch (err) {
                 dbg.warn('NamespaceFS: read_object_stream file close error', err);
             }
             try {
                 // release buffer back to pool if needed
-                if (buffer_pool_cleanup) buffer_pool_cleanup();
+                if (buffer_pool_cleanup) {
+                    dbg.log0('NamespaceFS: read_object_stream finally buffer_pool_cleanup');
```
guymguym, line 663:
```suggestion
                    dbg.log0('NamespaceFS: read_object_stream finally buffer_pool_cleanup', file_path);
```

src/sdk/object_sdk.js
```diff
@@ -64,6 +65,41 @@ class ObjectSDK {
...
+    }
+
+    throw_if_aborted() {
+        if (this.abort_controller.signal.aborted) throw new Error('request aborted signal');
+    }
+
+    add_abort_handler(handler) {
+        if (this.abort_controller.signal.aborted) setImmediate(handler);
+        return this.abort_controller.signal.addEventListener('abort', handler, { once: true });
     }
```
guymguym, line 103:
```suggestion
    add_abort_handler(handler) {
        const s = this.abort_controller.signal;
        if (s.aborted) {
            setImmediate(handler);
        } else {
            s.addEventListener('abort', handler, { once: true });
        }
    }
```

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -100,20 +101,12 @@ async function get_object(req, res) {
...
-    });
-    res.on('error', err => {
-        dbg.log0('response error:', err, req.path);
-        if (read_stream && read_stream.close) read_stream.close();
-    });
-
-    read_stream = await req.object_sdk.read_object_stream(params, res);
+    const read_stream = await req.object_sdk.read_object_stream(params, res);
     if (read_stream) {
+        // on http disconnection close the read stream to stop from buffering more data
```
guymguym, line 106:
```suggestion
        // if read_stream supports closing, then we handle abort cases such as http disconnection
        // by calling the close method to stop it from buffering more data which will go to waste.
```

guymguym (commented):
another round of comments. apologies for making it iterative.

src/sdk/namespace_fs.js
romayalon: it's because we wanted to put it at the beginning of every iteration,
and specifically after the open (for the first iteration) and before the get_buffer(),
if i'll move it next to the read_size it won't be checked before the get_buffer()

guymguym: no no no, I meant just the `const read_size = ...` line, not the throw if aborted...

src/sdk/object_sdk.js
guymguym: This suggested change makes sure we don't add the listener when aborted ^ as we don't want both setImmediate and addEventListener to be applied in this case.

src/core/nsfs.js
```diff
@@ -143,6 +109,43 @@ async function main(argv = minimist(process.argv.slice(2))) {
     }
 }
 
+function make_object_sdk_instance(fs_root, fs_config) {
```
guymguym, line 112: how about defining this to be `function init_request_sdk(req, res, fs_root, fs_config) {...}`

src/sdk/namespace_fs.js
```diff
@@ -602,29 +609,36 @@ class NamespaceFS {
...
+                    object_sdk.throw_if_aborted();
                 }
 
                 // write the data out to response
                 buffer_pool_cleanup = null; // cleanup is now in the socket responsibility
                 const write_ok = res.write(data, null, callback);
                 if (!write_ok) {
-                    drain_promise = stream_utils.wait_drain(res);
+                    drain_promise = stream_utils.wait_drain(res, { signal: object_sdk.abort_controller.signal })
+                        .catch(() => undefined);
```
guymguym, line 620: need to fix this to avoid swallowing errors raised from drain promise -
```suggestion
                    drain_promise = stream_utils.wait_drain(res, { signal: object_sdk.abort_controller.signal });
                    drain_promise.catch(() => undefined); // this avoids UnhandledPromiseRejection
```

src/sdk/object_sdk.js
```diff
@@ -64,6 +65,40 @@ class ObjectSDK {
...
+     * in order to handle aborting requests gracefully. The `abort_controller` member will
+     * be used to signal async flows that abort was detected.
+     * @see {@link https://nodejs.org/docs/latest/api/globals.html#class-abortcontroller}
+     * @param {import('http').IncomingMessage} req 
+     * @param {import('http').ServerResponse} res 
+     */
+    setup_abort_controller(req, res) {
+        res.once('error', err => {
+            dbg.log0('response error:', err, req.url);
+            this.abort_controller.abort();
```
guymguym, line 82:
```suggestion
            this.abort_controller.abort(err);
```

src/sdk/object_sdk.js
```diff
@@ -64,6 +65,40 @@ class ObjectSDK {
...
+        });
+
+        req.once('aborted', () => {
+            dbg.log0('request aborted', req.url);
+            this.abort_controller.abort();
+        });
```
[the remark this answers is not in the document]
guymguym: Seems like we need to do this:
```
        req.once('close', () => {
            if (req.destroyed) {
                dbg.log0('request aborted', req.url);
                this.abort_controller.abort(new Error('request aborted ' + req.url));
            }
        });
```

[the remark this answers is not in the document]
guymguym: Ok, so this is a gap for moving to node 16, because on node 14 we can't check the req.destroyed state yet.

FYI @liranmauda and @nimrod-becker - we need to track this so that once we switch to node 16 we make sure to include this fix, which @romayalon left as a comment in the code.

guymguym (approved):
Great work @romayalon. Super artistic and accurate as usual! Godspeed.

romayalon:
@guymguym Thank you for your great help and profound code review!

PR #6967, opened 2022-05-26
reduced log chattiness - changed level of log0 messages

dannyzaken:

### Explain the changes
1. fixed several error\warn messages that were printed with `dbg.log0` level
2. part of an effort to support reduced log chattiness by reducing log level to warn\err only
3. for some repeating messages increased the log level to `log1`

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/agent/block_store_services/block_store_azure.js
+++ b/src/agent/block_store_services/block_store_azure.js
@@ -194,7 +194,7 @@ class BlockStoreAzure extends BlockStoreBase {
 
                     const data_size = Number(info_arg.contentLength);
                     const noobaablockmd = info_arg.metadata.noobaablockmd || info_arg.metadata.noobaa_block_md;
-                    dbg.log0('block_store_azure._delete_blocks info: data_size: ', data_size, 'noobaablockmd: ', noobaablockmd);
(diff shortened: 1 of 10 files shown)


liranmauda (approved):
LGTM,
In General this is a good practice, even when the logs are only printed in a catch of an error, printing them in `dbg0` will cause it to be ignored in the case where we set the debug level to `warn` in the config map
https://github.com/noobaa/noobaa-core/blob/069e2f6c58e0a205a876fa8d0f84d68d3971e286/src/util/debug_config.js#L27-L31

romayalon:
@dannyzaken @liranmauda 
There are several log messages that I specifically was asked to keep as log0 and not an error/warning.
maybe we should decide on well-defined rules for it.

dannyzaken:
> @dannyzaken @liranmauda There are several log messages that I specifically was asked to keep as log0 and not an error/warning. maybe we should decide on well-defined rules for it.

@romayalon what is the reason to keep it in log0?

romayalon:
@guymguym can you please share your logging methodology?

liranmauda (approved):

PR #6972, opened 2022-05-31
Stringify error codes before update_issues_report() in namespace_s3 flows

romayalon:

### Explain the changes
1. In https://bugzilla.redhat.com/show_bug.cgi?id=2090968 we received an error from RGW and the error code was 502 which is a number and when calling update_issues_report() we pass this error code and validate that error_code is a string so we fail with INVALID_SCHEMA_PARAMS CLIENT on this rpc call.
RGW should not throw an error with a numeric error code but in order to secure this functionality, I added this casting.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


liranmauda (approved):
LGTM

PR #6973, opened 2022-06-01
NSFS - supporting GPFS DL for Encryption

jackyalbo:

### Explain the changes
1. Adding initial support for Encryption for NSFS
2. Will check for files/objects only on GPFS volumes if they are encrypted
3. Added a tool for checking encryption check on files on the FS - nsfs-enc
4. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/asm.gypi
+++ b/src/native/asm.gypi
@@ -2,14 +2,14 @@
 {
     'conditions': [
 
-        [ 'node_arch=="x64"', {
-            'defines': ['HAVE_AS_KNOWS_AVX512', 'HAVE_AS_KNOWS_SHANI'],
-        }],
-
         # LINUX
         [ 'OS=="linux"', {
             'conditions': [
                 [ 'node_arch=="x64"', {
+                    'defines': [
+                        'HAVE_AS_KNOWS_AVX512',
+                        'HAVE_AS_KNOWS_SHANI',
+                    ],
                     'rules': [{
                         'rule_name': 'assemble',
                         'extension': 'asm',
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -1,33 +1,111 @@
 /* Copyright (C) 2016 NooBaa */
 #include "../util/b64.h"
+#include "../util/buf.h"
 #include "../util/common.h"
 #include "../util/napi.h"
 #include "../util/os.h"
-#include "../util/buf.h"
 #include "./gpfs_fcntl.h"
 
 #include <dirent.h>
 #include <fcntl.h>
+#include <limits.h>
 #include <map>
 #include <math.h>
+#include <stdlib.h>
 #include <sys/stat.h>
 #include <sys/syscall.h>
 #include <sys/types.h>
 #include <sys/uio.h>
+#include <sys/xattr.h>
 #include <thread>
 #include <typeinfo>
 #include <unistd.h>
 #include <uv.h>
-#include <sys/xattr.h>
 #include <vector>
-#include <stdlib.h>
-#include <limits.h>
 
 #ifdef __APPLE__
-    #include <sys/param.h>
-    #include <sys/mount.h>
+#include <sys/mount.h>
+#include <sys/param.h>
 #else
-    #include <sys/statfs.h>
+#include <sys/statfs.h>
+#endif
+
+#define ROUNDUP(X, Y) ((Y) * (((X) + (Y)-1) / (Y)))
+
+// Should total to 256 (sizeof(buffer) 212 + sizeof(header) 16 + sizeof(payload) 28)
+#define GPFS_XATTR_BUFFER_SIZE 212
+#define GPFS_BACKEND "GPFS"
+#define GPFS_XATTR_PREFIX "gpfs"
+#define GPFS_DOT_ENCRYPTION_EA "Encryption"
+#define GPFS_ENCRYPTION_XATTR_NAME GPFS_XATTR_PREFIX "." GPFS_DOT_ENCRYPTION_EA
+
+// This macro should be used after openning a file
+// it will autoclose the file using AutoCloser and will throw an error in case of failures
+#define CHECK_OPEN_FD(fd)        \
+    AutoCloser closer(this, fd); \
+    do {                         \
+        if (fd < 0) {            \
+            SetSyscallError();   \
+            return;              \
+        }                        \
+    } while (0)
+
+// This macro should be used when we want to verify a file is open inside a wrapper function
+// it won't autoclose and will throw an error in case of the fd is not open
+#define CHECK_WRAP_FD(fd)                                                       \
+    do {                                                                        \
+        if (fd < 0) {                                                           \
+            SetError(XSTR() << _desc << ": ERROR not opened " << _wrap->_path); \
+            return;                                                             \
+        }                                                                       \
+    } while (0)
+
+// This macro should be used after calling any syscall/gpfs-syscall that uses errno
+// in case of failures - it will throw the errno using SetSyscallError
+#define SYSCALL_OR_RETURN(x)   \
+    do {                       \
+        int r = (x);           \
+        if (r) {               \
+            SetSyscallError(); \
+            return;            \
+        }                      \
+    } while (0)
+
+// This macro should be used after calling any gpfs-fcntl which doesn't use errno
+// It will handle error which number will be saved under gpfs_error
+#define GPFS_FCNTL_OR_RETURN(x)                                              \
+    do {                                                                     \
+        int gpfs_error = 0;                                                  \
+        int r = (x);                                                         \
+        if (r) {                                                             \
+            if (gpfs_error) {                                                \
+                SetError(XSTR() << "GPFS FCNTL error:" << DVAL(gpfs_error)); \
+            } else {                                                         \
+                SetSyscallError();                                           \
+            }                                                                \
+            return;                                                          \
+        }                                                                    \
+    } while (0)
+
+// This macro should be used when calling fstat in case we want to validate
+// no race condition happened changing the fd during the function run
+#define CHECK_CTIME_CHANGE(fd, stat_res, path)                    \
+    do {                                                          \
+        auto start_ctime = stat_res.st_ctime;                     \
+        SYSCALL_OR_RETURN(fstat(fd, &stat_res));                  \
+        auto end_ctime = stat_res.st_ctime;                       \
+        if (start_ctime != end_ctime) {                           \
+            SetError(XSTR() << "FileStat: " << DVAL(path)         \
+                            << " cancelled due to ctime change"); \
+        }                                                         \
+    } while (0)
+
+#ifdef __APPLE__
+#define flistxattr(a, b, c) ::flistxattr(a, b, c, 0)
+#define getxattr(a, b, c, d) ::getxattr(a, b, c, d, 0, 0)
+#define fgetxattr(a, b, c, d) ::fgetxattr(a, b, c, d, 0, 0)
+#define fsetxattr(a, b, c, d, e) ::fsetxattr(a, b, c, d, e, 0)
+#define fremovexattr(a, b) ::fremovexattr(a, b, 0)
 #endif
 
 namespace noobaa
@@ -35,12 +113,12 @@ namespace noobaa
 
 DBG_INIT(0);
 
+typedef std::map<std::string, std::string> XattrMap;
+
 static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg) = 0;
 
 static int (*dlsym_gpfs_linkat)(
-    gpfs_file_t fileDesc, const char *oldpath,
-    gpfs_file_t newdirfd, const char *newpath,
-    int flags) = 0;
+    gpfs_file_t fileDesc, const char* oldpath, gpfs_file_t newdirfd, const char* newpath, int flags) = 0;
 
 const static std::map<std::string, int> flags_to_case = {
     { "r", O_RDONLY },
@@ -50,9 +128,9 @@ const static std::map<std::string, int> flags_to_case = {
     { "rs+", O_RDWR | O_SYNC },
     { "sr+", O_RDWR | O_SYNC },
     { "w", O_TRUNC | O_CREAT | O_WRONLY },
-    #ifdef O_TMPFILE
-        { "wt", O_RDWR | O_TMPFILE },
-    #endif
+#ifdef O_TMPFILE
+    { "wt", O_RDWR | O_TMPFILE },
+#endif
     { "wx", O_TRUNC | O_CREAT | O_WRONLY | O_EXCL },
     { "xw", O_TRUNC | O_CREAT | O_WRONLY | O_EXCL },
     { "w+", O_TRUNC | O_CREAT | O_RDWR },
@@ -70,6 +148,9 @@ const static std::map<std::string, int> flags_to_case = {
     { "sa+", O_APPEND | O_CREAT | O_RDWR | O_SYNC },
 };
 
+const static std::vector<std::string> GPFS_XATTRS{ GPFS_ENCRYPTION_XATTR_NAME };
+const static std::vector<std::string> USER_XATTRS{ "user.content_md5" };
+
 struct Entry
 {
     std::string name;
@@ -77,6 +158,30 @@ struct Entry
     uint8_t type;
 };
 
+struct gpfsRequest_t
+{
+    gpfsFcntlHeader_t header;
+    gpfsGetSetXAttr_t payload;
+    char buffer[GPFS_XATTR_BUFFER_SIZE];
+};
+
+void
+build_gpfs_get_ea_request(gpfsRequest_t* reqP, std::string key)
+{
+    int nameLen = key.size();
+    int bufLen = sizeof(reqP->buffer);
+    memset(reqP->buffer, 0, bufLen);
+
+    reqP->header.totalLength = sizeof(*reqP);
+    reqP->header.fcntlVersion = GPFS_FCNTL_CURRENT_VERSION;
+    reqP->payload.structLen = reqP->header.totalLength - sizeof(reqP->header);
+    reqP->payload.structType = GPFS_FCNTL_GET_XATTR;
+    reqP->payload.nameLen = nameLen;
+    // bufferLen is the size of buffer - roundingup of the attribute name to 8 chars
+    reqP->payload.bufferLen = bufLen - ROUNDUP(nameLen, 8);
+    reqP->payload.flags = GPFS_FCNTL_XATTRFLAG_NONE;
+    memcpy(&reqP->payload.buffer[0], key.c_str(), nameLen);
+}
 
 template <typename T>
 static Napi::Value
@@ -89,7 +194,7 @@ api(const Napi::CallbackInfo& info)
 }
 
 void
-set_stat_res(Napi::Object res, Napi::Env env, struct stat &stat_res)
+set_stat_res(Napi::Object res, Napi::Env env, struct stat& stat_res, XattrMap& xattr_res)
 {
     res["dev"] = Napi::Number::New(env, stat_res.st_dev);
     res["ino"] = Napi::Number::New(env, stat_res.st_ino);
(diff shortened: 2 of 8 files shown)


guymguym:
@jackyalbo We discussed that we want to fail PUT-object if the request is for encryption, but the underlying filesystem provides non-encrypted storage.

guymguym:
@jackyalbo rebase and resolve conflicts?

jackyalbo:
> @jackyalbo rebase and resolve conflicts?

@guymguym Can you check I resolved in the way you intended for the GPFS sub module? thanks.

src/native/fs/fs_napi.cpp
```diff
@@ -285,6 +296,92 @@ struct Stat : public FSWorker
...
+        char xnambuf[256];
+        _is_encrypt_res = false;
+
+        strcpy(xnambuf,XATTR_APPID_TO_NAMESPACE(GPFS_ATTR_INTERNAL_APPL_ID));
+        strcat(xnambuf,".");
+        strcat(xnambuf,GPFS_DOT_ENCRYPTION_EA);
+
+        nameLen = strlen(xnambuf);
+
+        fd = open(_path.c_str(), O_RDONLY);
```
guymguym, line 330: This is probably a case where we would prefer to operate on an open file instead of a path, which means that instead of making this function an FSWorker it will be added as a method to FileWrap.

src/native/fs/fs_napi.cpp
```diff
@@ -285,6 +296,92 @@ struct Stat : public FSWorker
...
+        int bufLen = 256;
+        int nameLen;
+        gpfsFcntlHeader_t *hdrP;
+        gpfsGetSetXAttr_t *getAttrP;
+        char xnambuf[256];
+        _is_encrypt_res = false;
+
+        strcpy(xnambuf,XATTR_APPID_TO_NAMESPACE(GPFS_ATTR_INTERNAL_APPL_ID));
+        strcat(xnambuf,".");
+        strcat(xnambuf,GPFS_DOT_ENCRYPTION_EA);
```
guymguym, line 326: what does this do. missing comment to explain.

src/sdk/namespace_fs.js
```diff
@@ -510,10 +513,12 @@ class NamespaceFS {
             file = await nb_native().fs.open(fs_account_config, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
             const stat = await file.stat(fs_account_config);
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
+            const is_gpfs = Boolean(fs_account_config.backend === 'GPFS' && nb_native().fs.gpfs);
```
guymguym, line 516: this is repeated multiple times in the file. lets move to a function.

src/sdk/namespace_fs.js
```diff
@@ -747,9 +752,22 @@ class NamespaceFS {
                 }
                 // TODO use file xattr to store md5_b64 xattr, etc.
                 const stat = await nb_native().fs.stat(fs_account_config, upload_path);
+                const is_gpfs = Boolean(fs_account_config.backend === 'GPFS' && nb_native().fs.gpfs);
+                const is_encrypted = is_gpfs && await nb_native().fs.is_encrypted(fs_account_config, upload_path);
+                if (params.encryption && params.encryption.algorithm === 'AES256' && !is_encrypted) {
+                    dbg.error('upload_object: User requested encryption but encryption not supported for FS');
+                    throw new Error('upload_object: User requested encryption but encryption not supported for FS');
```
guymguym, line 759: This will become an internal error in S3, right. We want a more suitable error like ServerSideEncryptionConfigurationNotFoundError ([see](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#ErrorCodeList)) and also we want to add a custom message to explain the situation to the user better.

guymguym (changes requested):

src/native/fs/fs_napi.cpp
jackyalbo: @guymguym If we will add  is_encrypted to FileWrap doesn't it always to be in that class? will I just return false when it's not GPFS drvie or something like this, also adding this function as part of the gpfs submodule that you asked for also won't be needed. Do I get you right here?

src/native/fs/fs_napi.cpp
```diff
@@ -1443,6 +1540,8 @@ fs_napi(Napi::Env env, Napi::Object exports)
         // e.g: gpfs["foo"] = Napi::Function::New(env, api<Foo>);
         // gpfs.Freeze();
         exports_fs["gpfs"] = gpfs;
+        gpfs["is_encrypted"] = Napi::Function::New(env, api<IsEncrypted>);
+        gpfs.Freeze();
```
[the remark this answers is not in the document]
jackyalbo: work for me, what wasn't suppose to work?

src/native/fs/fs_napi.cpp
jackyalbo: added FileWrap is_encrypted functionality

src/native/fs/fs_napi.cpp
```diff
@@ -285,6 +296,92 @@ struct Stat : public FSWorker
...
+        bufP = (char *) malloc(bufLen);
+        if (bufP == NULL)
+        {
+            errno = -ENOMEM;
+            SetSyscallError();
+            return;
+        }
+        memset(bufP, 0, bufLen);
+        hdrP = (gpfsFcntlHeader_t *)bufP;
+        getAttrP = (gpfsGetSetXAttr_t *)&bufP[sizeof(*hdrP)];
```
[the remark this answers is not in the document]
jackyalbo: done

jackyalbo:
@guymguym Added the requests you asked on our last talk. Please re-review or remove request for changes. Thanks

src/native/fs/fs_napi.cpp
```diff
@@ -77,6 +92,27 @@ struct Entry
...
 };
 
+struct gpfsRequest_t{
+    gpfsFcntlHeader_t header;
+    gpfsGetSetXAttr_t payload;
+    char buffer[GPFS_XATTR_BUFFER_SIZE];
+};
+
+void build_gpfs_get_ea_request(gpfsRequest_t* reqP) {
+    int nameLen = strlen(GPFS_XATTR_FULL_NAME);
```
guymguym, line 102: Since this is a literal string constant, you don't need to use strlen(), and can simply use sizeof().
However, notice that sizeof returns the entire memory length including the null terminator, so it will be 1 more than strlen.
```suggestion
    int nameLen = sizeof(GPFS_XATTR_FULL_NAME) - 1; // sizeof counts the memory including the last null char
```

src/native/fs/fs_napi.cpp
```diff
@@ -30,6 +30,21 @@
...
+
+#define GPFS_DOT_ENCRYPTION_EA "Encryption"
+#define GPFS_ATTR_INTERNAL_APPL_ID   3
+#define GPFS_ATTR_EXT_NAMESPACE_ID   4
+
+#define GPFS_XATTR_PREFIX "gpfs"
+#define XATTR_APPID_TO_NAMESPACE(id) \
+  (id == GPFS_ATTR_INTERNAL_APPL_ID) ? GPFS_XATTR_PREFIX : ""
+#define XATTR_APPID_TO_NAMESPACE_LEN(id) \
+  (sizeof(XATTR_APPID_TO_NAMESPACE(id)) - 1)
```
guymguym, line 43: these macros seem unused now.

src/native/fs/fs_napi.cpp
```diff
@@ -1434,8 +1582,8 @@ void
 fs_napi(Napi::Env env, Napi::Object exports)
 {
     auto exports_fs = Napi::Object::New(env);
-
-    if(const char* env_p = std::getenv("GPFS_DL_PATH")) {
+    const char* env_p = std::getenv("GPFS_DL_PATH");
+    if (env_p != NULL) {
```
guymguym, line 1586: while you're at it - better rename this variable to gpfs_dl_path instead of env_p:

```suggestion
    const char* gpfs_dl_path = std::getenv("GPFS_DL_PATH");
    if (gpfs_dl_path) {
```

src/native/fs/fs_napi.cpp
```diff
@@ -30,6 +30,21 @@
...
+#define GPFS_ATTR_EXT_NAMESPACE_ID   4
+
+#define GPFS_XATTR_PREFIX "gpfs"
+#define XATTR_APPID_TO_NAMESPACE(id) \
+  (id == GPFS_ATTR_INTERNAL_APPL_ID) ? GPFS_XATTR_PREFIX : ""
+#define XATTR_APPID_TO_NAMESPACE_LEN(id) \
+  (sizeof(XATTR_APPID_TO_NAMESPACE(id)) - 1)
+#define GPFS_XATTR_FULL_NAME GPFS_XATTR_PREFIX "." GPFS_DOT_ENCRYPTION_EA
+// Should total to 256 (sizeof(buffer) 212 + sizeof(header) 16 + sizeof(payload) 28)
+#define GPFS_XATTR_BUFFER_SIZE   212
```
guymguym, line 46: I would make this an assert in fs_napi() that `sizeof(gpfsRequest_t)` correctly sums up to 256

src/sdk/namespace_fs.js
```diff
@@ -468,6 +469,8 @@ class NamespaceFS {
                 } else {
                     r.stat = await nb_native().fs.lstat(fs_account_config, entry_path);
                 }
+                const is_gpfs = Boolean(fs_account_config.backend === 'GPFS' && nb_native().fs.gpfs);
+                r.is_encrypted = is_gpfs && await nb_native().fs.is_encrypted(fs_account_config, entry_path);
```
guymguym, line 473: isn't it under `nb_native().fs.gpfs.is_encrypted`?

src/native/fs/fs_napi.cpp
```diff
@@ -287,6 +323,62 @@ struct Stat : public FSWorker
     }
 };
 
+/**
+ * IsEncrypted is an spectrumscale op
+ */
+struct IsEncrypted : public FSWorker
```
guymguym, line 329: I would add this to the general purpose getxattr methods.
Sync with @romayalon who is planning to enhance those functions.

guymguym (commented):
Comments from review, CC @romayalon:
1. fs_napi - make all stat calls return also xattrs - `fs.stat`/`fs.lstat` should now use `open`/`open(O_NOFOLLOW)` and then get both the stat and xattr and optionally also gpfs xattrs.
2. namespace_fs - all calls to getxattr will become unneeded, and will be taken from the stat result stat.xattr for `_get_object_info()`, and also for `_get_upload_info()` (this should be added to make sure we always return both `{etag,encryption}` on all upload responses.

romayalon:
@guymguym @jackyalbo 
2 questions about this decision -
1. What about the dir cache? Will we use the xattr of the stat result of the cache entry? if yes we need to invalidate it correctly, if no we still need to call getxattr again (and there is a performance issue here)
2. Thoughts about the performance of calling stat + getallxattr when one of them is not really needed?

src/native/fs/fs_napi.cpp
```diff
@@ -158,6 +200,110 @@ set_fs_worker_stats(Napi::Env env, Napi::Object fs_worker_stats, std::string wor
...
+        
+        value = std::string(val.cdata(), value_len);
+    } else if (value_len == 0) {
+        value = "";
+    }
+    return 0;
+}
+
+int
+get_fd_xattr(int fd, std::map<std::string, std::string> &xattr, bool skip_user_att)
```
guymguym, line 236: how about a typedef for this map type? to make it more readable

src/native/fs/fs_napi.cpp
```diff
@@ -70,13 +82,38 @@ const static std::map<std::string, int> flags_to_case = {
     { "sa+", O_APPEND | O_CREAT | O_RDWR | O_SYNC },
 };
 
+const static std::vector<std::string> gpfs_xattrs{ GPFS_XATTR_FULL_NAME };
+const static std::vector<std::string> user_xattrs{ "user.content_md5" };
```
guymguym, line 86: these are const vectors, so how about capitalizing their names GPFS_XATTRS , USER_XATTRS

src/native/fs/fs_napi.cpp
```diff
@@ -30,6 +30,18 @@
...
 
+#define ROUNDUP(X,Y) ( (Y) * ( ((X)+(Y)-1) / (Y) ) )
+
+#define GPFS_DOT_ENCRYPTION_EA "Encryption"
+#define GPFS_ATTR_INTERNAL_APPL_ID   3
+#define GPFS_ATTR_EXT_NAMESPACE_ID   4
+
+#define GPFS_BACKEND "GPFS"
+#define GPFS_XATTR_PREFIX "gpfs"
+#define GPFS_XATTR_FULL_NAME GPFS_XATTR_PREFIX "." GPFS_DOT_ENCRYPTION_EA
```
guymguym, line 41: this name is too generic, maybe GPFS_ENCRYPTION_XATTR_NAME

src/native/fs/fs_napi.cpp
```diff
@@ -284,66 +432,82 @@ struct FSWrapWorker : public FSWorker
...
         Begin(XSTR() << "Stat " << DVAL(_path));
     }
     virtual void Work()
     {
-        int r = stat(_path.c_str(), &_stat_res);
+        int flags = O_RDONLY;
+        // using O_PATH (MAC - O_SYMLINK) with O_NOFOLLOW allow us to open the symlink itself
+        // instead of openning the file it links to https://man7.org/linux/man-pages/man7/symlink.7.html
+#ifdef __APPLE__
+        if (_use_lstat) flags = O_SYMLINK | O_NOFOLLOW;
```
guymguym, line 457: did you intentionally override the flags (O_RDONLY)? or did you mean to add to it with `flags |= ...` ?

src/native/fs/fs_napi.cpp
```diff
@@ -1140,126 +1301,19 @@ struct LinkFileAt : public FSWrapWorker<FileWrap>
...
-        r = fstat(fd, &stat_res);
-        if (r) {
-            SetSyscallError();
-            return;
-        }
-        auto end_ctime = stat_res.st_ctime;
-        if (start_ctime != end_ctime) {
-            // Cancel on ctime change due to non atomic nature of this call
-            SetError(XSTR() << "FileGetxattr: " << DVAL(_wrap->_path) << " cancelled due to ctime change");
-        }
```
guymguym, line 1239: I think this was accidentally removed - this is a check that the inode ctime was not changed in between the calls to get the xattrs, to make sure we don't race with another client making changes to the attributes.

src/native/fs/fs_napi.cpp
```diff
@@ -284,66 +432,82 @@ struct FSWrapWorker : public FSWorker
...
+        }
+        int r;
+        r = fstat(fd, &_stat_res);
+        if (r) {
+            SetSyscallError();
+            r = close(fd);
+            if (r) SetSyscallError(); // Report only
+            return;
+        }
+        if (!_use_lstat) {
```
guymguym, line 474: can you add a comment to explain why use_lstat does not require (or does not support?) xattrs

src/native/fs/fs_napi.cpp
jackyalbo: yes, O_PATH is replacing O_RDONLY as is stronger than him, you can't even read the file with it.
```
perform operations that act purely at the file descriptor level.  The file itself is not opened, and
other file operations read, write ...  fail with the error EBADF.
```
When O_PATH is specified in flags, flag bits other than O_CLOEXEC, O_DIRECTORY, and O_NOFOLLOW are ignored.
```

File: src/native/fs/fs_napi.cpp [hunk 19]
```diff
@@ -158,6 +202,110 @@ set_fs_worker_stats(Napi::Env env, Napi::Object fs_worker_stats, std::string wor
...
+            return rc;
+        }
+        int error_code = gpfsGetXattrRequest.payload.errReasonCode;
+        if (error_code == GPFS_FCNTL_ERR_NONE)
+        {
+            int name_len = gpfsGetXattrRequest.payload.nameLen;
+            int buffer_len = gpfsGetXattrRequest.payload.bufferLen;
+            xattr[key] = std::string(gpfsGetXattrRequest.buffer[ROUNDUP(name_len, 8)], buffer_len);
+        } else if (error_code == GPFS_FCNTL_ERR_NO_ATTR) {
+            xattr[key] = "";
```
guymguym, line 301: is it better to include it as empty string rather than not including it at all?

src/native/fs/fs_napi.cpp
```diff
@@ -158,6 +202,110 @@ set_fs_worker_stats(Napi::Env env, Napi::Object fs_worker_stats, std::string wor
...
+        } else {
+            return error_code;
```
guymguym, line 303: maybe add a specific log here so that we know this error code comes from gpfs and not from posix?

src/native/fs/fs_napi.cpp
```diff
@@ -158,6 +202,110 @@ set_fs_worker_stats(Napi::Env env, Napi::Object fs_worker_stats, std::string wor
...
+
+static int 
+get_fd_gpfs_xattr(int fd, XattrMap &xattr)
+{
+    for (std::string key: GPFS_XATTRS) {
+        gpfsRequest_t gpfsGetXattrRequest;
+        build_gpfs_get_ea_request(&gpfsGetXattrRequest, key);
+        int rc = dlsym_gpfs_fcntl(fd, &gpfsGetXattrRequest);
+        if (rc != 0)
+        {
```
guymguym, line 291: keep the curly on same line, also another case couple of lines below

src/native/fs/fs_napi.cpp
```diff
@@ -158,6 +202,110 @@ set_fs_worker_stats(Napi::Env env, Napi::Object fs_worker_stats, std::string wor
...
+        value = "";
+    }
+    return 0;
+}
+
+static int 
+get_fd_xattr(int fd, XattrMap &xattr, bool skip_user_xattr)
+{
+    if (skip_user_xattr) { // we won't list the attributes just return the prefefined list
+        for (std::string key: USER_XATTRS) {
```
guymguym, line 241: I think you can iterate with a const ref instead like so - 
https://stackoverflow.com/questions/15176104/range-based-loop-get-item-by-value-or-reference-to-const
```suggestion
        for (auto const &key: USER_XATTRS) {
```

src/native/fs/fs_napi.cpp
guymguym: ok, worth noting it in a comment as it is not easy to notice

src/native/asm.gypi
```diff
@@ -2,14 +2,14 @@
...
-        }],
-
         # LINUX
         [ 'OS=="linux"', {
             'conditions': [
                 [ 'node_arch=="x64"', {
+                    'defines': [
+                        'HAVE_AS_KNOWS_AVX512',
+                        'HAVE_AS_KNOWS_SHANI',
+                    ],
```
guymguym, line 12: in addition to this AVX512 fix we just need another tiny fix so it can run on Mac M1 with rosetta (as x64) - there's this unwanted space that creeped in at the end of the filename string here. Fix it too and we'll be able to run it locally...
https://github.com/noobaa/noobaa-core/blob/bbb342f02571b6c6ae4fc89fcc7a5a19004d5c79/src/native/third_party/isa-l.gyp#L96

guymguym (approved):

src/sdk/namespace_fs.js
```diff
@@ -793,19 +778,19 @@ class NamespaceFS {
     // Comparing both device and inode number (st_dev and st_ino returned by stat) 
     // will tell you whether two different file names refer to the same thing.
     // If so, we will return the etag of the file_path
```
liranmauda, line 780: I guess this comment needs to be updated, maybe something like: 
```
    // If so, we will return the etag of the file_path and <something about encryption 😄 >
```

src/sdk/namespace_fs.js
```diff
@@ -1326,6 +1315,33 @@ class NamespaceFS {
...
+    _get_upload_info(stat) {
+        const etag = this._get_etag(stat);
+        const encryption = this._get_encryption_info(stat);
+        return {
+            etag,
+            encryption
+        };
+    }
+
+    _get_encryption_info(stat) {
```
liranmauda, line 1327: By the look of it, it will return only in gpfs case, if that is the case, can you add a comment on this function?

src/native/fs/fs_napi.cpp
```diff
@@ -1,46 +1,124 @@
...
+                            << " cancelled due to ctime change"); \
+        }                                                         \
+    } while (0)
+
+#ifdef __APPLE__
+#define flistxattr(a, b, c) ::flistxattr(a, b, c, 0)
+#define getxattr(a, b, c, d) ::getxattr(a, b, c, d, 0, 0)
+#define fgetxattr(a, b, c, d) ::fgetxattr(a, b, c, d, 0, 0)
+#define fsetxattr(a, b, c, d, e) ::fsetxattr(a, b, c, d, e, 0)
+#define fremovexattr(a, b) ::fremovexattr(a, b, 0)
```
liranmauda, line 108: Did the indentation slip? 

True in all cases if `#ifdef __APPLE__` in this file

src/native/fs/fs_napi.cpp
```diff
@@ -50,9 +128,9 @@ const static std::map<std::string, int> flags_to_case = {
     { "rs+", O_RDWR | O_SYNC },
     { "sr+", O_RDWR | O_SYNC },
     { "w", O_TRUNC | O_CREAT | O_WRONLY },
-    #ifdef O_TMPFILE
-        { "wt", O_RDWR | O_TMPFILE },
-    #endif
+#ifdef O_TMPFILE
+    { "wt", O_RDWR | O_TMPFILE },
+#endif
```
liranmauda, line 133: Indentation

src/native/fs/fs_napi.cpp
```diff
@@ -124,10 +229,15 @@ set_stat_res(Napi::Object res, Napi::Env env, struct stat &stat_res)
     res["ctime"] = Napi::Date::New(env, uint64_t(round(ctimeMs)));
     res["birthtime"] = Napi::Date::New(env, uint64_t(round(birthtimeMs)));
 
+    auto xattr = Napi::Object::New(env);
+    res["xattr"] = xattr;
+    for (auto it = xattr_res.begin(); it != xattr_res.end(); ++it) {
+        xattr.Set(it->first, it->second);
```
liranmauda, line 235: Do we want to use Set or we can set it as all the other places here using obj[key] = val ?

src/native/fs/fs_napi.cpp
```diff
@@ -284,63 +476,53 @@ struct FSWrapWorker : public FSWorker
 struct Stat : public FSWorker
```
liranmauda, line 476: Might be worth commenting that now this op can actually do stat and lstat.
I know that having stat and lstat that are much similar is duplicating the code, but isn't it more straightforward? @jackyalbo @guymguym WDYT?

liranmauda (changes requested):

src/native/third_party/isa-l.gyp
```diff
@@ -93,7 +93,7 @@
                         'isa-l/erasure_code/ec_highlevel_func.c',
                         'isa-l/erasure_code/gf_vect_mul_sse.asm',
                         'isa-l/erasure_code/gf_vect_mul_avx.asm',
-                        'isa-l/erasure_code/gf_vect_dot_prod_sse.asm ',
+                        'isa-l/erasure_code/gf_vect_dot_prod_sse.asm',
```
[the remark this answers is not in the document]
guymguym: WDYM isa-l issue? they don't have a gyp file for building - it's ours.

src/native/fs/fs_napi.cpp
guymguym: no it didn't - by default preprocessor directives are not indented. We just formatted the code with code formatter and it fixed the indent to be on first column.

src/native/fs/fs_napi.cpp
guymguym: Not sure what you mean by "more straighforward" here.
For example in open there is no lopen but instead there are flags to indicate how to treat symlinks.

src/native/fs/fs_napi.cpp
guymguym: It's not the same. The other cases are std::map, this one is Napi::Object.

src/sdk/namespace_fs.js
guymguym: is there a comment that will help you understand that better?

liranmauda: It is pretty understandable, I think, but maybe add a comment:
`_get_encryption_info will return undefined if it is not in gpfs` 
or something like that 😄

src/native/fs/fs_napi.cpp
liranmauda: I have missed something in reading it, Looks good.

liranmauda (approved):

src/sdk/namespace_fs.js
jackyalbo: done

PR #6974, opened 2022-06-07
Log based replication background worker and log parser

kfir-payne:

### Explain the changes
1. Add log based replication background worker, responsible for scanning all the replication rules, and based on candidates provided by the log parser, decide which objects should be replicated.
2. Add log parser, responsible for getting the next log to check, parse the log, and returning a list of keys (candidates) for the logs replication scanner to check.
3. Add a function to `replication store` that returns all the replication rules containing log based replication info.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


src/server/bg_services/log_replication_scanner.js
```diff
@@ -0,0 +1,219 @@
...
+
+            // case 1: object not found in src
+            if (!candidate.src_object_info) {
+                break;
+            }
+
+            // case 2: object not found in dst
+            if (!candidate.dst_object_info) {
+                keys_sizes_map_to_copy[candidate.key] = candidate.ContentLength;
+                break;
```
romayalon, line 114: in lines 108 and 114 it should be continute and not break if I understand correctly, right?

src/server/bg_services/log_replication_scanner.js
```diff
@@ -0,0 +1,219 @@
...
+        const src_dst_objects_list = await this.head_objects(src_bucket, dst_bucket, candidates);
+
+        // edge case 1: src_head_map = {} , nothing to replicate
+        if (!src_dst_objects_list.length) return keys_sizes_map_to_copy;
+
+        for (let candidate of src_dst_objects_list) {
+
+            // case 1: object not found in src
+            if (!candidate.src_object_info) {
+                break;
```
romayalon, line 108: this case solves a situation in which an object was written and you saw it in the logs but then it was deleted before NooBaa replicated it?

src/server/bg_services/log_replication_scanner.js
```diff
@@ -0,0 +1,219 @@
...
+
+            // case 3: object present both on the src and dst, need to check_data_or_md_change
+            if (this.check_data_or_md_changed(candidate.src_object_info, candidate.dst_object_info)) {
+                keys_sizes_map_to_copy[candidate.key] = candidate.ContentLength;
+            }
+        }
+        return keys_sizes_map_to_copy;
+    }
+
+    check_data_or_md_changed(src_info, dst_info) {
```
romayalon, line 125: all the following functions can be reused as they have almost if not identical code as in the regular replication flow.
I would recommend creating replication_utils.js file and using the same code.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -0,0 +1,219 @@
...
+        return true;
+    }
+
+    async scan() {
+        if (!this.noobaa_connection) throw new Error('noobaa endpoint connection is not started yet...');
+
+        const replications = await replication_store.find_aws_log_based_replication_rules();
+
+        await P.all(_.map(replications, async repl => {
+            _.map(repl.rules, async rule => {
```
romayalon, line 67: In my opinion - going through all rules here is incorrect because we can have many rules and the cycle will take lots of time.

src/server/system_services/replication_store.js
```diff
@@ -102,6 +102,14 @@ class ReplicationStore {
 
         return reduced_replications;
     }
+
+    async find_aws_log_based_replication_rules() {
+        dbg.log0('find_aws_log_based_replication_rules: ');
+        const replications = await this._replicationconfigs.find({ deleted: null });
+
+        const reduced_replications = _.filter(replications, repl => repl.aws_log_replication_info);
+        dbg.log0('find_aws_log_based_replication_rules: ', reduced_replications);
```
romayalon, line 111: As we discussed - return reduced_replications and filter aws logs replication in the db query

src/server/bg_services/log_replication_scanner.js
kfir-payne: correct

src/server/bg_services/log_replication_scanner.js
```diff
@@ -0,0 +1,152 @@
...
+const _ = require('lodash');
+const system_store = require('../system_services/system_store').get_instance();
+const dbg = require('../../util/debug_module')(__filename);
+const system_utils = require('../utils/system_utils');
+const config = require('../../../config');
+const P = require('../../util/promise');
+const Semaphore = require('../../util/semaphore');
+const replication_store = require('../system_services/replication_store').instance();
+const cloud_utils = require('../../util/cloud_utils');
+const log_formatter = require('./replication_log_parser');
```
dannyzaken, line 13: shouldn't this be `log_parser` instead of `log_formatter`?

dannyzaken (commented):
I have a general comment. in my opinion, the log_replication_scanner should be completely generic and should not have any references to the type of the sources (AWS in this case). 
only the parser should be created according to the type, and it should return a list of candidates. the rest should be completely the same for each cloud we will support in the future

dannyzaken:
few points following a discussion with @kfir-payne:

* the parser should get a limit value for the number of candidates to get from the log
   * the parser should try to provide the number of candidates with some limitations on it. e.g.: in AWS the log file is 100 lines long, and not all of it is write operations that provide candidates. the parser should read as much log objects as necessary to get to the `limit`. of course we should have a max number of objects that we want to read in a single iteration
* we should have configs for these values, such as max candidates per iteration, max objects to read per iteration, delay should be different from list scanner.
* use the replication store logic of sorting by `last_cycle_end`.

package.json
```diff
@@ -111,6 +111,7 @@
     "ps-node": "0.1.6",
     "request": "2.88.2",
     "rimraf": "3.0.2",
+    "s3-access-log-parser": "0.0.2",
```
dannyzaken, line 114: I prefer to avoid this dependancy. The parsing should not be that complicated, and I don't like the idea of adding a dependency just for that (security issues, maintainability, etc.). 
looking at the [github repo](https://github.com/icodeforlove/s3-access-log-parser), it looks like it is not maintained in the last 6 years

src/server/bg_services/log_replication_scanner.js
```diff
@@ -0,0 +1,154 @@
...
+        }
+        return keys_sizes_map_to_copy;
+    }
+
+    async head_objects(src_bucket_name, dst_bucket_name, candidates) {
+        const src_dst_objects_info = await P.all(_.map(candidates, async key => {
+            try {
+                dbg.log1('log_replication_scanner head_object: params:', src_bucket_name, key);
+
+                const src_dst_object_info = await P.all([await this.noobaa_connection.headObject({
```
romayalon, line 135: this array of 2 items originates from an old version of this function? I don't see why it's needed

src/server/bg_services/log_replication_scanner.js
romayalon: I'm still worrying about going through all rules here at once,
this function includes list_objects of the logs_bucket, get_object of the next log, parsing, head_object of each candidate * 2 (src & dst buckets) comparison of the results, replication of each object that needs to be replicated

src/server/bg_services/replication_log_parser.js
```diff
@@ -0,0 +1,176 @@
...
+    const { logs_bucket, logs_prefix, continuation_token } = aws_log_replication_info;
+
+    const s3 = get_source_bucket_aws_connection(source_bucket_id, aws_log_replication_info);
+
+    let candidates_map = {};
+    let logs_retrieved_count = config.AWS_LOG_CANDIDATES_LIMIT;
+
+    do {
+        const next_log_list = await aws_list_next_log(s3, logs_bucket, logs_prefix, continuation_token);
+        if (continuation_token === next_log_list.NextContinuationToken) break;
```
romayalon, line 31: need a comment

src/server/bg_services/replication_log_parser.js
```diff
@@ -0,0 +1,176 @@
...
+async function get_aws_log_candidates(source_bucket_id, aws_log_replication_info, candidates_limit) {
+    const { logs_bucket, logs_prefix, continuation_token } = aws_log_replication_info;
+
+    const s3 = get_source_bucket_aws_connection(source_bucket_id, aws_log_replication_info);
+
+    let candidates_map = {};
+    let logs_retrieved_count = config.AWS_LOG_CANDIDATES_LIMIT;
+
+    do {
+        const next_log_list = await aws_list_next_log(s3, logs_bucket, logs_prefix, continuation_token);
```
romayalon, line 30: num of list_objects can be reduced to 1 instead of 10 when using list objects version 1 and passing the key_marker instead of the continuation token as we discussed

src/server/bg_services/replication_log_parser.js
```diff
@@ -0,0 +1,176 @@
...
+        if (continuation_token === next_log_list.NextContinuationToken) break;
+
+        const next_log = await aws_get_next_log(s3, logs_bucket, next_log_list.Contents[0].Key);
+        aws_parse_log_object(candidates_map, next_log);
+        logs_retrieved_count -= 1;
+    }
+    while (Object.keys(candidates_map).length < candidates_limit && logs_retrieved_count);
```
romayalon, line 37: a reminder to add the limit of 1000 candidates

src/server/system_services/replication_store.js
```diff
@@ -86,22 +86,49 @@ class ReplicationStore {
...
     }
+
+    async find_aws_log_based_replication_rules() {
+        dbg.log1('find_aws_log_based_replication_rules: ');
+        const replications = await this._replicationconfigs.find({ deleted: null });
+
+        const aws_reduced_replications = _.filter(replications, repl => repl.log_replication_info.aws_log_replication_info);
+        const sorted_reduced_replications = _.map(aws_reduced_replications, repl => ({
+            replication_id: repl._id,
+            log_replication_info: _.minBy(repl.log_replication_info, info => (info.status && info.status.last_cycle_end) || 0) //least_recently_replicated_rule
```
romayalon, line 126: we don't need to sort if we are going through all rules, let's put it in a gap

romayalon:
Please add to the PR description that this PR was not tested and this BG worker is not called yet.

liranmauda:
Please note that the package-lock should not be with the changes, please rebase it from master, and use npm 8.

dannyzaken:
Moving the PR to https://github.com/noobaa/noobaa-core/pull/7124

PR #6998, opened 2022-06-30
Remove `allowed_buckets`

tangledbytes:
### Explain the changes

This PR attempts to remove `allowed_buckets` construct from Noobaa core:
1. The construct of `allowed_buckets.permission_list` is replaced with Bucket Policy.
2. All the files that belonged to lambdas and were touched in the PR are deleted.
3. Tests were modified to use Bucket Policies wherever either `allowed_buckets.full_permission` or `allowed_buckets.permission_list` was being used.
4. Upgrade script has been added for version 5.12 which removes `allowed_buckets` field from the accounts in the database.

### Issues: Fixed BZ1929108
1. https://bugzilla.redhat.com/show_bug.cgi?id=1929108

### Testing Instructions:
1. Run `make test`
2. Run `make test-postgres`
3.  Run `docker run --user root --rm --name test1 noobaa-tester ./src/test/unit_tests/run_npm_test_on_test_container.sh -s sudo_index.js`

- [ ] Doc added/updated
- [ ] Tests added


liranmauda (commented):
Few General notes:
1. Need to take care of upgrades
2. On places that you touch and the have promise chains, better convert to async/await
3. The lambda files can be removed completely (@nimrod-becker to keep me honest)

nimrod-becker:
I think they can

src/server/common_services/auth_server.js
```diff
@@ -660,12 +662,45 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
     return response;
 }
 
+// has_bucket_permission returns true if either the account is the owner of the bucket
+// or the account has explicit permission to access the bucket via the Bucket Policy.
+//
+// NOTE: If the owner account has is denied access as per the policy then the owner
+// also is denied access to the bucket.
 function has_bucket_permission(bucket, account) {
```
jackyalbo, line 670: it seems like we only use has_bucket_permission in lambda triggers which I think we are going to remove soon. Maybe we should just take it off

src/server/common_services/auth_server.js
```diff
@@ -660,12 +662,45 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
-        );
+    dbg.log0('has_bucket_permission:', bucket.name, account.email, bucket.owner_account.email);
+    const is_owner = bucket.owner_account.email.unwrap() === account.email.unwrap();
+    const bucket_policy = bucket.s3_policy;
+
+    if (!bucket_policy) {
+        return is_owner;
+    }
+
+    const bucket_policy_evaluation = _.flatten(
```
jackyalbo, line 679: I would add an explanation about what we are doing here. maybe we should check for an * or pass the specfic action which I think will be able to understand. so maybe we should just check that he has full control on the bucket.

src/server/system_services/bucket_server.js
```diff
@@ -715,53 +705,52 @@ async function update_buckets(req) {
  */
 function update_bucket_s3_access(req) {
```
jackyalbo, line 706: I would remove that. the user only option to update the bucket s3 access will be using the bucket policy api. I think this was used only by the UI anyhow.

src/server/system_services/bucket_server.js
```diff
@@ -1319,6 +1291,35 @@ async function claim_bucket(req) {
         }
     }
 
+    try {
```
jackyalbo, line 1294: I think we can delete claim_bucket - we don't use it to the best of my knowledge

src/server/system_services/bucket_server.js
```diff
@@ -1363,6 +1369,13 @@ async function delete_claim(req) {
         auth_token: req.auth_token
     });
 
+    // Delete policy asscociated with the bucket
```
jackyalbo, line 1372: also delete_claim can be removed

src/test/system_tests/test_bucket_access.js
```diff
@@ -109,12 +109,33 @@ async function setup() {
 
     let account = account_by_name(system_info.accounts, full_access_user.email);
     full_access_user.access_keys = account.access_keys[0];
+    await Promise.all(
```
jackyalbo, line 112: I am not sure what you are doing here. please add a comment

src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -168,25 +165,26 @@ mocha.describe('s3_bucket_policy', function() {
...
-        await s3_owner.deleteBucketPolicy({ // should work - owner can always delete the buckets policy
-            Bucket: BKT,
-        }).promise();
-        await s3_a.putBucketPolicy({ // user a have FC to bucket
-            Bucket: BKT,
-            Policy: JSON.stringify(policy)
-        }).promise();
-    });
+    // @utkarsh-pro: Should no longer work
+    // mocha.it('should be able to set bucket policy when none set', async function() {
```
jackyalbo, line 169: maybe change to owner and not any user. but in aws in a bucket with no bucket policy can't any user set one - that's what I remember - but please validate ( by testing in AWS console)

src/upgrade/upgrade_scripts/5.12.0/remove_allowed_buckets.js
```diff
@@ -0,0 +1,30 @@
+/* Copyright (C) 2022 NooBaa */
+"use strict";
+
+async function run({ dbg, system_store }) {
+    try {
+        dbg.log0('starting upgrade accounts...');
+        const accounts = system_store.data.accounts
```
jackyalbo, line 7: Is that what we want to do with upgrade? will it set that all the users have full control or no control? it seems like we need to handle such upgrades with care and at least let the users know what will be the artifacts.

jackyalbo (changes requested):
Just note with all of this PR that buckets can have no bucket policy - and in that case that we need to verify what AWS do and that we are doing the same.
I am not sure setting a bucket policy for a user without him asking for it is recommended.

src/server/system_services/bucket_server.js
utkarsh-pro: Got it, will remove this ✅ .

src/server/system_services/bucket_server.js
utkarsh-pro: Got it, will remove this ✅ .

src/server/system_services/bucket_server.js
utkarsh-pro: I will once find references to this function and if found none, I will remove it ✅ .

src/server/common_services/auth_server.js
utkarsh-pro: Actually:
- `req.has_bucket_permission` wraps this function.
    - Gets called only in 2 places but as you mentioned in context of lambdas.
- `req.has_s3_bucket_permission` also wraps this function.
    - `req.check_s3_bucket_permission` wraps the above function.
        - Gets called in `load_bucket` in `object_server.js`
        - Gets called in `find_bucket` in `bucket_server.js`
    - Gets called in `list_buckets` in `bucket_server.js`

src/server/common_services/auth_server.js
utkarsh-pro: I also wanted to perform check only for the requested action however, when I tried to see the call chain for the function, it seemed that the original action gets lost (for example when getting called from the `bucket_server.js`) which makes it difficult to pass it as one of the parameters to the function.

This method just checks if there is any policy which can be used to allow user the access to the bucket (to try to replicate `permission_list` behaviour)

src/test/system_tests/test_bucket_access.js
utkarsh-pro: This is again trying to replicate `permission_list`. Instead of manually granting permissions to each bucket, it loops over the `permission_list` and generates S3 policies which will give the user equivalent permissions over the buckets that `permission_list` was giving.

src/upgrade/upgrade_scripts/5.12.0/remove_allowed_buckets.js
utkarsh-pro: Bucket owners will still have full control over the buckets but yes, anyone relying on the `permission_list` or `full_permission` will face issues. Would you recommend generating S3 policies for them in the upgrade process?

Also, I don't think we can truly replicate `full_permission` because of the way we handle setting policies on the buckets.

src/server/common_services/auth_server.js
jackyalbo: ok, so It make sense. but let's add a comment explaining that. I think usually what we wanted to see is if the user suppose to see this bucket, so maybe we can just check if he can read the bucket/list it objects. WDYT?

src/upgrade/upgrade_scripts/5.12.0/remove_allowed_buckets.js
jackyalbo: I think anyway most users are not using allowed_buckets (as we don't have any UI anymore). but I think users that didn't play with allowed_buckets and could access all of their buckets need to still be able to do that. was the default for newly created bucket is S3 full permission?

src/server/common_services/auth_server.js
utkarsh-pro: Oh yes, that's what I tried in the first iteration. Turns out that functions like `load_bucket` (which indirectly relies on this function) is invoked in multiple operations in `object_server.js` including in `create_object_upload` etc.

Taking the example of `create_object_upload`, if a user does have upload permission (regardless of the list and get permissions), the operation should succeed right?

Assuming the above assumption is correct, the `has_bucket_permission` checks if a user can do _something_ with the bucket or not.

---
I also explored if we can get rid of permission checks in the utility functions `load_bucket`, `find_bucket`. They seemed redundant assuming that the request has already flown through the initial `authorization` checks (where we do have the information of "action" as well). However, what I realized and I might be wrong here and please correct if I am - We can invoke RPCs to these entities which will bypass the REST flow essentially bypassing the REST authorization flow, hence I left the permission checks in these methods.

src/upgrade/upgrade_scripts/5.12.0/remove_allowed_buckets.js
utkarsh-pro: Under the assumption that we are talking about users using NooBaa core in conjunction with NooBaa operator. 

When users create an OBC to get a bucket, we create it using operator's account (which has `full_permission: true`), so the bucket owner is the `operator`. Later on `operator` creates a new account which grants access to the bucket via `permission_list`.  So the `full_permission` is `true` for the account that creates the bucket but not for the account that is given to the user for accessing the bucket.

To address that, if in https://github.com/noobaa/noobaa-operator/pull/956 we replace [PermissionList](https://github.com/utkarsh-pro/noobaa-operator/blob/69896b9bd268bbac8fcf695adfb3a1633b5b636a/pkg/obc/provisioner.go#L600) with a S3 policy which allows access to the bucket, then the user should be able keep their access. But this approach has the following issues:
1. As you mentioned earlier, we would be creating policy without user's permission.
2. We will be adding policy to the bucket, which is fine but if the user would try to add some policy to the bucket (unaware of the previous policy set by the operator), they will overwrite the policy (which will block their own access). I am not sure how prevalent is this flow though where user set S3 policies on the OBC based buckets. 
    - We can kind of fix this though, by denying user (again via policy) to set any S3 policies on the bucket (again not sure how prevalent is this flow).

src/server/common_services/auth_server.js
jackyalbo: I agree that it seems that authorization needs to happen only in its own part, so maybe our goal should be to remove this checks from both load_bucket and find_bucket, and leave it only to s3_rest authorization. please check if you find any place where you can see such behavior where we authorize the request regarding one bucket and then tries to access a different one (other than the triggers that we already discussed and needs to be removed in the future)

src/upgrade/upgrade_scripts/5.12.0/remove_allowed_buckets.js
jackyalbo: I think that now the user creates this account and give if FC using bucket_claim_owner. so not sure we need any change to the noobaa-operator in that case.
but other than that, I think that we should decide on one approach - I think maybe we shouldn't add any other bucket policies but maybe worth discussing with more people. But we should let the user know that bucket-owner/system-owner will need to recreate all the removed access with new bucket policies.

utkarsh-pro:
**Streamlining discussion (TLDR of above comments)** for reviewers

_All of the "resolved" comments have been addressed_

There are 2 items still needs to be discussed in this PR:
1. `has_bucket_permission` has become fairly complicated because
    - Earlier it used to rely on `allowed_buckets` which now no longer exists. It has been now replaced with bucket policy.
    - Regardless of the intended action, `has_bucket_permission` verifies every action because this function is directly and indirectly called in multiple other portions where the intended action may or may not be accessible.
2. Removal of `allowed_buckets.permission_list` (NooBaa Operator uses the above field to grant access to accounts to the buckets created via OBC) causes the following issues :
    1. NooBaa operator will fail to give access to accounts associated with the OBC.
        - Proposed Solution: Replace `allowed_buckets.permission_list` with bucket policy.
        - Issue with proposed solution: Creating a bucket policy without user consent.
    3. Users will not be able to access their previous buckets:
        - Proposed Solution: Translate all of the `allowed_buckets.permission_list` to bucket policy in the 5.12 upgrade script.
        - Issue with proposed solution: Creating a bucket policy without user consent.

utkarsh-pro:
After discussions over the items mentioned in the previous comment:
1. `has_bucket_permission` rigorous permission checks if can be avoided or reduced must be reduced. This may involve removing redundant permission checks.
2. Removal `allowed_buckets.full_permission` may cause regression but the goal is to ensure that at least 90% of the users should be able to upgrade without any hiccups. 
    - 2 flows that needs to be fully functional are:
        1. Creation of OBC and verifying if the bucket is accessible and the user can perform operations on the bucket.
        2. Creation of S3 bucket by the user and verifying that the user can perform operations on the bucket.

src/server/common_services/auth_server.js
utkarsh-pro: You were right @jackyalbo. We should be able to get rid of authorization inside `load_bucket`, `find_bucket` and `list_buckets` because we the request gets already authorized in the s3_rest flow and we never request for buckets which were not part of original user request.

A caveat though: I might be wrong here so please correct me if I am - in our tests we often make direct RPC calls and RPC calls only go through authentication (in their middlewares), so won't the tests will fail? Also as far as I understand, RPC calls can be made independent of operator, in that case it might be problematic to skip authorization right?

jackyalbo: I still think this can go as well... for sure the bucket policy part as we didn't check it before this change as well, so probably not need for it now all of a sudden. I think commands from RPC should be handled on it's own. But If anything jumps to your eyes that some call suppose to be blocked but doesn't - let's discuss.

jackyalbo (commented):
Most of it looks clean and good. See if you can also clean has_bucket_permission or even better so get rid of it. Thanks

utkarsh-pro:
After discussion with @jackyalbo, `load_bucket` and `find_bucket` in the bucket_server and object_server no longer rely on `has_bucket_permission` which essentially means, that if an RPC is performed and the user is authenticated then they will be able to access buckets/objects without further authorization. The exception to this is `list_bucket` which will list only the buckets to which the requesting account has access. Hence, authz is now only limited to `s3_rest` `sts_rest`.

PS: For some reason the  test is failing here although passing on the local system.

jackyalbo (approved):

jackyalbo (approved):
Looks good, looks like you need to rebase.

utkarsh-pro:
Rebased the branch ✔️. Thank you @jackyalbo for reviewing this PR 😄!

Merging the PR...

PR #7012, opened 2022-07-17
Add access mode property to namespace resource

kfir-payne:

### Explain the changes
1. Add access mode property to namespace resource.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/api/pool_api.js
+++ b/src/api/pool_api.js
@@ -105,6 +105,10 @@ module.exports = {
                     nsfs_config: {
                         $ref: 'common_api#/definitions/nsfs_config'
                     },
+                    access_mode: {
+                        type: 'string',
+                        enum: ['READ_ONLY', 'READ_WRITE']
+                    },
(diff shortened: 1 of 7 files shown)


dannyzaken (commented):
* I think you should also add access_mode for NSFS (@romayalon any reason not to?)
* you should pass the access_mode from object_sdk to the namespace

romayalon:
@kfir-payne @dannyzaken Currently NSFS namespace buckets can use only one namespace resource. I guess we would like to support multi namespace resources in the future, so yes it can be added unless there is a limitation with the current implementation.

dannyzaken:
@romayalon @kfir-payne I don't see any limitation unless there are flows in NSFS that expect to write stuff to the DB, other than user data (e.g. if some conf\metadata\stats\whatever is stored in the PV)

dannyzaken (approved):

PR #7014, opened 2022-07-19
prints improvements and addition of jemalloc prints

jackyalbo:

### Explain the changes
1.  adding jemalloc prints using malloc_stats_print
2. setting down the logging threshold to 512MB
3. checking the threshold against RSS memory and not heap

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/tools/malloc_napi.cpp
+++ b/src/native/tools/malloc_napi.cpp
@@ -1,0 +1,54 @@
+/* Copyright (C) 2016 NooBaa */
+#include "../util/napi.h"
+#include "../util/common.h"
+
+#include <stdlib.h>
+#include <uv.h>
+#include <dlfcn.h>
+#include <malloc.h>
+
+namespace noobaa
+{
+
+DBG_INIT(0);
+
+static void (*malloc_stats_print)
+(
+    void (*write_cb) (void *, const char *), 
+    void *cbopaque, 
+    const char *opts
+) = 0;
+
+static void _print_malloc_stats(const Napi::CallbackInfo& info);
+static void _print_jemalloc_stats(const Napi::CallbackInfo& info);
+
+void
+malloc_napi(Napi::Env env, Napi::Object exports)
+{
+    auto exports_malloc = Napi::Object::New(env);
+    uv_lib_t *lib = (uv_lib_t*) malloc(sizeof(uv_lib_t));
+    lib->handle = RTLD_DEFAULT;
+    lib->errmsg = NULL;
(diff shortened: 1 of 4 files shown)


src/native/tools/malloc_napi.cpp
```diff
@@ -0,0 +1,25 @@
+/* Copyright (C) 2016 NooBaa */
+#include "../util/napi.h"
+
+#include <stdlib.h>
+#include <jemalloc/jemalloc.h>
```
guymguym, line 5: Where is this header coming from? I think we want to have this as an optional thing because it will fail on local builds

src/util/panic.js
```diff
@@ -36,17 +38,31 @@ function enable_heapdump(name, next_mb, step_mb) {
     return module.exports;
 }
 
+function enable_malloc_print(name, next_mb, step_mb) {
```
guymguym, line 41: Perhaps just include it in the regular memory interval instead of adding another?

jackyalbo: I can, just was thinking we maybe want to set one and not the other, like if I will turn on the heapdump it will kill the process after every 256MB.

src/native/tools/malloc_napi.cpp
jackyalbo: currently is seems to fail on none-local builds as well :) 
how will we set it optional? I was thinking maybe to add jemalloc to the builder docker...

guymguym:
Looks good @jackyalbo !!

src/native/tools/malloc_napi.cpp
```diff
@@ -0,0 +1,54 @@
...
+(
+    void (*write_cb) (void *, const char *), 
+    void *cbopaque, 
+    const char *opts
+) = 0;
+
+static void _print_malloc_stats(const Napi::CallbackInfo& info);
+static void _print_jemalloc_stats(const Napi::CallbackInfo& info);
+
+void
```
guymguym, line 25: I just wish there was a comment here or on the entire file, saying that his is for functions to the memory allocation system, and can print jemalloc stats (if loaded), or the default malloc stats otherwise.

src/util/panic.js
```diff
@@ -57,9 +73,17 @@ function memory_monitor() {
         const align = h.step - (increase % h.step);
         h.next += increase + align;
     }
+    if (mp && mp.next && mp.next <= current) {
```
guymguym, line 76: Perhaps this block should come before the heapdump one, because we know heapdump has crashed the process before, so it just better be last so we get all the info we can.

PR #7025, opened 2022-07-27
🌸🍒 mdsequence cherries

baum:
# PostgreSQL-related performance optimization.

Those changes were 🌸🍒 cherry-picked from [dannyzaken/noobaa-core/tree/danny-mdsequence](https://github.com/dannyzaken/noobaa-core/tree/danny-mdsequence)

- Use PostgreSQL sequence
   - currently mdsequence is using a counter inside a jsonb field, which has very bad performance in parallel writes. 
   - changed to use postgres native SEQUENCE.
- Increase  PostgreSQL client connections pool size if `LOCAL_MD_SERVER`
   - for the endpoints, we saw that when a large number of small writes is done in parallel, a lot of queries are waiting for clients. the DB has max connections of 600, so it should be able to handle more connections.
- Remove sort by version_enabled in find_object_null_version
   - For some reason, the generated query is causing Postgres to perform a sort according to `version_enabled`, even though the index is also on `version_enabled`.
   - `version_enabled` is always null in this case, and the sort is redundant. removing it solved the issue.

### DIFF
--- a/config.js
+++ b/config.js
@@ -596,6 +596,11 @@ config.QUOTA_MAX_OBJECTS = Number.MAX_SAFE_INTEGER;
 //////////////////////////
 config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS = 60 * 60 * 1000; // 1 hour
 
+///////////////////////////////////////////
+//      PostgreSQL client pool size      //
+///////////////////////////////////////////
+config.POSTGRES_MAX_CLIENTS = (process.env.LOCAL_MD_SERVER === 'true') ? 80 : 10;
+
 /////////////////////
 //                 //
 //    OVERRIDES    //
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -55,7 +55,7 @@ class MDStore {
             schema: data_block_schema,
             db_indexes: data_block_indexes,
         });
-        this._sequences = db_client.instance().define_collection({
+        this._sequences = db_client.instance().define_sequence({
             name: 'mdsequences' + test_suffix,
         });
     }
@@ -397,12 +397,7 @@ class MDStore {
      * @returns {Promise<number>}
      */
     async alloc_object_version_seq() {
-        // empty query, we maintain a single doc in this collection
-        const query = {};
-        const update = { $inc: { object_version_seq: 1 } };
-        const options = { upsert: true, returnOriginal: false };
-        let res = await this._sequences.findOneAndUpdate(query, update, options);
-        return res.value.object_version_seq;
+        return this._sequences.nextsequence();
     }
 
     /**
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -181,7 +181,7 @@ class MDStore {
             upload_started: null,
         }, {
             hint: 'null_version_index',
-            sort: { bucket: 1, key: 1, version_enabled: 1 },
+            sort: { bucket: 1, key: 1 },
         });
     }
 
@@ -194,7 +194,7 @@ class MDStore {
             // partialFilterExpression:
             deleted: null,
         }, {
-            sort: { bucket: 1, key: 1, version_enabled: 1 },
+            sort: { bucket: 1, key: 1 },
         });
     }
 
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -11,6 +11,7 @@ const crypto = require('crypto');
 const util = require('util');
 const EventEmitter = require('events').EventEmitter;
 const { Pool, Client } = require('pg');
+const { MongoSequence } = require('./mongo_client');
 
 const P = require('./promise');
 const dbg = require('./debug_module')(__filename);
@@ -26,6 +27,7 @@ const { RpcError } = require('../rpc');
 const SensitiveString = require('./sensitive_string');
 const time_utils = require('./time_utils');
 const config = require('../../config');
+
 mongodb.Binary.prototype[util.inspect.custom] = function custom_inspect_binary() {
     return `<mongodb.Binary ${this.buffer.toString('base64')} >`;
 };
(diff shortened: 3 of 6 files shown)


dannyzaken (commented):
@baum let's have 3 separate commits, so it will be easier to review and backport
1. mdsequence (squash relevant commits)
2. pool max client 
3. sort changes

src/util/postgres_client.js
```diff
@@ -1331,7 +1355,7 @@ class PostgresClient extends EventEmitter {
         this.new_pool_params = {
 
             // TODO: check the effect of max clients. default is 10
-            // max: 50,
+            max: (process.env.LOCAL_MD_SERVER === 'true') ? 80 : 10,
```
dannyzaken, line 1358: I suggest adding 'POSTGRES_MAX_CLIENTS` to config.js, which gives us the ability to override it with an env.

src/server/object_services/md_store.js
```diff
@@ -181,7 +181,7 @@ class MDStore {
             upload_started: null,
         }, {
             hint: 'null_version_index',
-            sort: { bucket: 1, key: 1, version_enabled: 1 },
+            sort: { bucket: 1, key: 1 },
```
dannyzaken, line 184: there's also `find_object_or_upload_null_version`, that uses a similar query. let's change that too

config.js
```diff
@@ -596,6 +596,12 @@ config.QUOTA_MAX_OBJECTS = Number.MAX_SAFE_INTEGER;
 //////////////////////////
 config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS = 60 * 60 * 1000; // 1 hour
 
+///////////////////////////////////////////
+//      PostgreSQL client pool size      //
+///////////////////////////////////////////
+config.POSTGRES_MAX_CLIENTS = 80
+config.POSTGRES_DEFAULT_CLIENTS = 10
```
dannyzaken, line 603: I suggest having one `POSTGRES_MAX_CLIENTS` config, and in postgres_client, just use it. It will make it easier to override without trying to remember which one is used in each case

```suggestion
config.POSTGRES_MAX_CLIENTS = (process.env.LOCAL_MD_SERVER === 'true') ? 80 : 10;
```

src/util/postgres_client.js
```diff
@@ -1284,6 +1310,33 @@ class PostgresClient extends EventEmitter {
             if (pg_client) await pg_client.end();
         }
     }
+
+    async upgradeSequence(name) {
```
dannyzaken, line 1314: this should be part of the upgrade script. it will not be used in any other context

src/server/object_services/md_store.js
baum: I also subscribed to this approach, however after consideration, I am leaning towards the presented approach. 

The upgrade of sequences is about the client's internal model mapping into the DB and change in its implementation. The current approach makes sense responsibility-wise and makes the code tidier and more readable on both the upgrade script and Postgres client code sides.

src/util/postgres_client.js
```diff
@@ -1284,6 +1310,33 @@ class PostgresClient extends EventEmitter {
...
+            pg_client = new Client({ ...this.new_pool_params, database: undefined });
+            await pg_client.connect();
+            const res = await pg_client.query(`SELECT count(*) FROM pg_tables WHERE tablename  = '${name}';`);
+            const count = Number(res.rows[0].count);
+            if (count === 0) {
+                dbg.log0(`⛔️ Table ${name} not found`);
+                return;
+            }
+            dbg.log0(`✅ Table ${name} is found`);
+            const mongoSeq = new MongoSequence({ name, client: this });
```
dannyzaken, line 1326: why `MongoSequence`?

baum: To calculate the [start value](https://github.com/baum/noobaa-core/blob/1e7f1aab2c3bbefc45fccb129b21eb8306bbb8bd/src/util/postgres_client.js#L1327) for the native postgres [sequence creation](https://github.com/baum/noobaa-core/blob/1e7f1aab2c3bbefc45fccb129b21eb8306bbb8bd/src/util/postgres_client.js#L1330).

src/util/postgres_client.js
baum: Initially, I also thought so 😀 however after consideration I've started leaning towards the current logic code breakdown. 

We are dealing with a change in DB client model mapping into the relational DB, specifically how sequences are mapped and their transformation.  The present approach looks cleaner responsibility-wise. Both the Postgres client and the upgrade script look cleaner and tidier, with no code duplication.

jackyalbo: I agree. We should remove

src/util/postgres_client.js
dannyzaken: but why mongo? we are upgrading from postgres table

baum: during the upgrade, the sequence is still in so to speak "mongo" format,  i.e. a single doc in this collection.... [`MongoSequence.nextsequence()`](https://github.com/baum/noobaa-core/blob/1e7f1aab2c3bbefc45fccb129b21eb8306bbb8bd/src/util/mongo_client.js#L85-L92) contains the logic which knows to get the (next) value if the sequence

src/util/postgres_client.js
dannyzaken: I prefer to have the upgrade script as self-contained as possible, with as few dependencies as possible.
Also, once the upgrade is completed, we are left with a dead code that has no other purpose. 

I suggest providing a function in the db client to perform direct SQL queries and use that for the upgrade script.

baum: I would prefer to reuse the existing code, instead of having  duplication in the implementation of:

- fetching current value from a collection-based sequence
- native Postgres sequence creation
- etc...

IMHO sequence upgrade is similar to DB [Creation](https://github.com/noobaa/noobaa-core/blob/44b8f84e0af51abe952d9c0a44891f3ad18ebe9f/src/util/postgres_client.js#L1287-L1299), [Drop](https://github.com/noobaa/noobaa-core/blob/44b8f84e0af51abe952d9c0a44891f3ad18ebe9f/src/util/postgres_client.js#L1287-L1299), collection [create](https://github.com/noobaa/noobaa-core/blob/44b8f84e0af51abe952d9c0a44891f3ad18ebe9f/src/util/postgres_client.js#L1410-L1423) or new sequence [create](https://github.com/baum/noobaa-core/blob/1e7f1aab2c3bbefc45fccb129b21eb8306bbb8bd/src/util/mongo_client.js#L577-L579): those are the responsibility of the corresponding DB Client.  The responsibility of the upgrade script is to [verify the DB Type](https://github.com/baum/noobaa-core/blob/danny-mdsequence-cherries/src/upgrade/upgrade_scripts/5.12.0/upgrade_postgres_native_sequence.js#L8-L14) and let the DB client do SQL or other DB-specific operations since the DB client is the layer that is responsible for those kind of stuff.  Finally, the client's upgrade code is self-contained, once the versions of the core which used the collection-based sequence are no longer supported, the Postgres client [upgrade function](https://github.com/baum/noobaa-core/blob/1e7f1aab2c3bbefc45fccb129b21eb8306bbb8bd/src/util/postgres_client.js#L1314-L1338) could be easily removed...

baum: Another thought is we could incorporate the logic of [if there is a table, convert to native sequence](https://github.com/baum/noobaa-core/blob/1e7f1aab2c3bbefc45fccb129b21eb8306bbb8bd/src/util/postgres_client.js#L1314-L1338) into [postgres_client.define_sequence()](https://github.com/baum/noobaa-core/blob/1e7f1aab2c3bbefc45fccb129b21eb8306bbb8bd/src/util/postgres_client.js#L1463-L1465) logic, and achieve lazy migration that way. This way we don't need the upgrade script at all.

src/util/postgres_client.js
```diff
@@ -1284,6 +1310,30 @@ class PostgresClient extends EventEmitter {
...
     }
+
+    // Lazy migration of the old mongo style collection/table based
+    // sequences. If a table with the name matching the sequence one
+    // is found, then:
+    // - fetch the current sequence value from the collection
+    // - drop the table
+    // - return the init value to be used for native sequence
+    // If no table is found, return 1 - clean install
+    async migrateFromMongoSequence(name) {
```
dannyzaken, line 1321: I think there could be some race conditions here since we have several endpoints starting at once that can run this migration.

baum: it does not matter which start value wins the race to `CREATE SEQUENCE IF NOT EXISTS`. we don't care if a couple of seq numbers are skipped. the sequence uniqueness invariant is preserved and one and only  DB sequence instance is created. 

wdyt?

baum: Good point @dannyzaken

I have reviewed the sequence creation/migration process. The drop table moved after the native sequence creation completion and `DROP TABLE` became `IF EXISTS`. The drop table becomes reentrant. The possibility of entering the race after the table is dropped but the native sequence does not exist yet and winning by creating a native sequence with init value 1 is excluded. 

So during the endpoints race, whoever wins the following invariants are preserved:
- one and only native sequence is created
- the uniqueness of sequence number in case of migration is preserved.  we don't care if a couple of sequence numbers are skipped, during migration
- in the case of migration the collection table is dropped

Let me know if you still see any issues there.

dannyzaken (approved):

PR #7028, opened 2022-07-31
Preps for the ARM image build

baum:
# Preps for the ARM image build

- Update third-party native build
- Update the docker build procedure
- Multiarch build procedure with manifest was extracted into https://github.com/noobaa/noobaa-core/pull/7038


src/deploy/NVA_build/builder.Dockerfile
```diff
@@ -7,14 +7,13 @@ LABEL maintainer="Liran Mauda (lmauda@redhat.com)"
...
-RUN dnf update -y -q --nobest && \
-    dnf clean all
-RUN dnf install -y -q wget unzip which vim python2 python3 && \
-    dnf group install -y -q "Development Tools" && \
-    dnf clean all
+RUN dnf update -y -q --nobest
+RUN dnf clean all
+RUN dnf install -y -q wget unzip which vim python2 python3
+RUN dnf group install -y -q "Development Tools"
+RUN dnf clean all
```
liranmauda, line 16: Why add the `RUN`? 
each RUN is a layer, and as we are not squashing the layers there is no point in doing a `dnf clean all` as the data remains in the previous layer. 
You can add a comment that explains why they are all in the same instruction.

src/deploy/NVA_build/install_nodejs.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -x
```
liranmauda, line 3: If adding `-x` I think it is better to give the debug prompt a better meaning:
```Suggestion
export PS4='\e[36m+ ${FUNCNAME:-main}@${BASH_SOURCE}:${LINENO} \e[0m'
set -x
```

src/deploy/NVA_build/install_nodejs.sh
```diff
@@ -8,17 +10,22 @@ then
     exit 1
 fi
 
+if [ "$(uname -m)" = "aarch64" ]; then
```
liranmauda, line 13:
```Suggestion
if [ "$(uname -m)" = "aarch64" ]
then
```

liranmauda (changes requested):
And few General Notes:
1. Is it compiling on Mac M1?
2. We don't need some subdirectories such as Docs and examples in Isa-l and Unit tests in Cm256.

liranmauda:
Also Adding @guymguym as a reviewer

src/deploy/NVA_build/builder.Dockerfile
baum: Good point, reverting.

liranmauda:
We also need to update `src/native/third_party/versions.md`

baum:
> We also need to update `src/native/third_party/versions.md`

✅ [versions.md](https://github.com/baum/noobaa-core/blob/arm-image/src/native/third_party/versions.md)

liranmauda (approved):
LGTM, We need to test it on M1 (which I assume will not work).
We can merge this and create an Issue that it is not compiling on M1

guymguym:
@liranmauda Can we fix it for M1 before merging? It would be best to avoid creating this technical debt...

.github/workflows/build-amd-image.yaml
```diff
@@ -0,0 +1,84 @@
+name: Build AMD image
+on: [ push, pull_request ]
```
dannyzaken, line 2: we don't need this to run on every push\pr. the same for the other workflows

dannyzaken (changes requested):
let's not push these workflows to run on every PR

dannyzaken:
maybe this PR should first handle the dependencies update, which is a good thing to do regardless of ARM\M1 builds. 
I suggest removing the changes regarding different archs, and only leaving the 3rd parties update

src/deploy/NVA_build/builder.Dockerfile
```diff
@@ -45,7 +44,11 @@ RUN chmod +x ./install_nodejs.sh && \
 #   Size: ~ 43 MB
 ##############################################################
 RUN stable_version=$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt) && \
-    curl -LO https://storage.googleapis.com/kubernetes-release/release/${stable_version}/bin/linux/amd64/kubectl && \
+    if [ "$(uname -m)" = "aarch64" ]; \
+    then arch=arm64; \
+    else arch=amd64; \
+    fi && \
```
liranmauda, line 50: Can we make it one-liner? 
```Suggestion
if [ "$(uname -m)" = "aarch64" ] ; then arch=arm64 ; else arch=amd64 ; fi && \
```

dannyzaken:
> @liranmauda Can we fix it for M1 before merging? It would be best to avoid creating this technical debt...

@guymguym I'm testing it on an M1 mac, and will update

dannyzaken:
@baum I was able to build a noobaa-core image on my m1 mac, and deploy it on minikube. Thanks for this PR! 😄 🎉
I think we can merge it once the workflows are fixed (see below). 

Some gaps that still exist:
* the Postgres image we use in the operator is not built for arm. running on m1 with the x86 image is not usable (system store queries take ~30 seconds). we will need to address that in the operator
* `make test` does not work. we try to get mongo in the test container build. mongo 3.6 which doesn't have an arm version.
* `make test-postgres` also doesn't work, since it uses the same test container. I guess it will also need an arm compatible postgres image. 


some comments specific to this PR:
* I think we should have one build workflow that gets the platform as a parameter. we already have the [manual-full-build](https://github.com/noobaa/noobaa-core/blob/master/.github/workflows/manual-full-build.yaml) workflow, so let's try to modify it to handle arm builds as well. once we have it, we can add arm builds to our nightly build workflows. In my opinion, we can defer the workflows part to a separate PR after we close the other gaps above. 

@guymguym @liranmauda any comments?

guymguym:
@liranmauda Are we going to use multi-arch image like `docker buildx` or `podman buildx`?

guymguym:
@dannyzaken Did you try building `node-gyp rebuild` directly on your M1 and checked that the output is arm64 and not x86_64 (which runs ok with rosetta)? because that's what I am getting without this PR:
```
$ file build/Debug/nb_native.node
build/Debug/nb_native.node: Mach-O 64-bit bundle x86_64
```

liranmauda:
> @liranmauda Are we going to use multi-arch image like `docker buildx` or `podman buildx`?

@guymguym Currently there is no such plan.
@nimrod-becker WDYT?

nimrod-becker:
@liranmauda less familiar with it, what do you say? If there is the value we can do it (not necessarily as part of this specific PR, can be a follow-up). Would that imply changes to our build pipeline?

dannyzaken:
> @dannyzaken Did you try building `node-gyp rebuild` directly on your M1 and checked that the output is arm64 and not x86_64 (which runs ok with rosetta)? because that's what I am getting without this PR:
> 
> ```
> $ file build/Debug/nb_native.node
> build/Debug/nb_native.node: Mach-O 64-bit bundle x86_64
> ```

@guymguym I didn't test building natively on M1. I only tested building an image with `make noobaa`

baum:
@dannyzaken removed the verification actions:

- '.github/workflows/build-amd-image.yaml'
- '.github/workflows/build-arm-image.yaml'

@nimrod-becker the idea is to enable multi-arch manual-full-build.yaml, see the current revision. 

However, the new action has to be tested. Could anyone trigger a controlled test run of the new manual build?

Best regards 🖖

baum:
Multiarch build procedure with manifests was extracted into https://github.com/noobaa/noobaa-core/pull/7038

dannyzaken (approved):

PR #7031, opened 2022-08-03
Fix Postgres latest version index and handle upgrade

romayalon:

### Explain the changes
1. Removed version_past from latest_version_index index because postgres considers nulls as distinct, causes multiple latest versions of objectmds to be allowed. 
2. Handled upgrade by doing: 
 A. Updated old versions of an object with version_past = true
 B. Dropped the current index - idx_btree_objectmds_latest_version_index
3. Recreation of the index will be handled lazily by the md_store constructor.
1.  On which version the upgrade should be?

### Testing Instructions:
1. Install old noobaa on pg (4.7 and up).
2. Put bucket.
3. Put versioning on the bucket.
4. Put/delete multiple objects of the same key and bucket **in parallel**.
5. Head object - you might get an old version of the object/get no such key (delete marker).
6. Upgrade NooBaa version.
7. Head object again and get the latest version.
8. Try step 4 again but this time step 5 will always return the latest version of the object.

- [ ] Doc added/updated
- [ ] Tests added


src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
```diff
@@ -0,0 +1,86 @@
...
+            dbg.log0('upgrade_objectmds: no versioned_bucket_ids, update not needed');
+        }
+
+        // 2. drop index
+        try {
+            const drop_index = `DROP INDEX idx_btree_objectmds_latest_version_index;`;
+            const drop_res = await md_store._objects.single_query(drop_index);
+            dbg.log0('upgrade_objectmds: dropped index successfully', drop_res);
+        } catch (err) {
+            if (err.code !== '42704') throw err;
```
guymguym, line 70: I would move this to be a new function in PgTable -
```js
drop_index(index_short_name) {
  try {
   await _internal_sql_query(`DROP INDEX idx_btree_${this.name}_${index_short_name};`);
  } catch (err) {
    if (err.code !== '42704') throw err; // allow "undefined_object" error
  }
}
```

src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
```diff
@@ -0,0 +1,86 @@
...
+                AND ((A.data -> 'upload_started'::text) IS NULL OR (A.data -> 'upload_started'::text) = 'null'::jsonb)
+                AND ((B.data -> 'upload_started'::text) IS NULL OR (B.data -> 'upload_started'::text) = 'null'::jsonb) 
+                AND A.data->'version_seq' < B.data->'version_seq'`;
+
+            for (let bucket of versioned_buckets) {
+                try {
+                    dbg.log0('upgrade_objectmds: updating bucket', bucket._id);
+                    const query = `${update} WHERE _id in (${select_fields} WHERE
+                        A.data->>'bucket'='${bucket._id}' AND B.data->>'bucket'='${bucket._id}'
+                        AND ${select_by});`;
```
guymguym, line 50: I don't think you need to run over the versioned buckets - you repeat the scan per bucket.
but if you just match on the A.data->>'bucket' = B.data->>'bucket' like you do for keys it will be a single query for all the buckets.

guymguym:
```
nbcore=# EXPLAIN
SELECT A._id
FROM objectmds A, objectmds B
WHERE A.data->>'bucket' = B.data->>'bucket'
AND A.data->>'key' = B.data->>'key'
AND ((A.data -> 'version_past'::text) IS NULL OR (A.data -> 'version_past'::text) = 'null'::jsonb)
AND ((B.data -> 'version_past'::text) IS NULL OR (B.data -> 'version_past'::text) = 'null'::jsonb)
AND ((A.data -> 'deleted'::text) IS NULL OR (A.data -> 'deleted'::text) = 'null'::jsonb)
AND ((B.data -> 'deleted'::text) IS NULL OR (B.data -> 'deleted'::text) = 'null'::jsonb)
AND ((A.data -> 'upload_started'::text) IS NULL OR (A.data -> 'upload_started'::text) = 'null'::jsonb)
AND ((B.data -> 'upload_started'::text) IS NULL OR (B.data -> 'upload_started'::text) = 'null'::jsonb)
AND A.data->>'version_seq' < B.data->>'version_seq'
;
                                                                                                    QUERY PLAN
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Nested Loop  (cost=0.28..16.34 rows=1 width=25)
   Join Filter: (((a.data ->> 'version_seq'::text) < (b.data ->> 'version_seq'::text)) AND ((a.data ->> 'bucket'::text) = (b.data ->> 'bucket'::text)) AND ((a.data ->> 'key'::text) = (b.data ->> 'key'::text)))
   ->  Index Scan using idx_btree_objectmds_latest_version_index on objectmds a  (cost=0.14..8.15 rows=1 width=508)
   ->  Index Scan using idx_btree_objectmds_latest_version_index on objectmds b  (cost=0.14..8.15 rows=1 width=483)
(4 rows)

```

src/server/object_services/schemas/object_md_indexes.js
```diff
@@ -8,12 +8,10 @@ module.exports = [
     {
         // find_object_latest()
         // list_objects()
+        // note that version_past is always null here by partialFilterExpression.
         fields: {
             bucket: 1,
             key: 1,
-            // we include version_past as extra index field to separate from null_version_index.
-            // note that version_past is always null here by partialFilterExpression.
-            version_past: 1,
```
guymguym, line 16: I think we need the same handling for the other two indexes - latest_version_index and null_version_index where we also used a boolean field which is null by default

guymguym (changes requested):

src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
romayalon: yes but I'm afraid it'll take so much time so we will see a session timeout, so I thought maybe we should still iterate over the buckets. WDYT?

guymguym: Can you set a custom timeout for this connection?

romayalon: I can try,
@tamireran and I are currently doing some performance tests on a 200K objects dataset and only for one bucket we can see it takes 13 seconds.
currently, I'm running it for all buckets and it didn't finish even after more than 10 minutes.

guymguym: Well but even a single bucket can contain all the objects... so we need to be prepared for a longer call. Did you try the query I suggested in the pr review or the one in the code now?

romayalon: @guymguym I tried both

guymguym: Can you try just the select without the update? Is that significantly shorter?

src/upgrade/upgrade_manager.js
```diff
@@ -142,7 +143,7 @@ async function run_upgrade() {
             for (const script of upgrade_scripts) {
                 dbg.log0(`running upgrade script ${script.file}: ${script.description}`);
                 try {
-                    await script.run({ dbg, db_client, system_store, system_server });
+                    await script.run({ dbg, db_client, system_store, system_server, md_store });
```
dannyzaken, line 146: can't we get the md_store inside the script?

src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
romayalon: ok I found the issue, there was missing > in the left null check

guymguym: and now what? is it fast?

romayalon: the select of all buckets now takes 10 seconds, which is good.
The update is very slow

romayalon: ->> - returns output but query plan doesn't contain index scan
-> - doesn't return output but query plan contains an index scan

guymguym: Afaiu data->>'key' is equivalent to data->'key'::text

guymguym: https://www.postgresql.org/docs/current/functions-json.html

guymguym: Can you show what is your final suggestion? I'm a bit lost

romayalon:
```
UPDATE objectmds SET data = jsonb_set(data, '{version_past}', '"true"') 
WHERE _id in (SELECT distinct A._id FROM objectmds A, objectmds B
WHERE (A.data->>'bucket'=B.data->>'bucket')
AND (A.data->>'key' = B.data->>'key')
AND ((A.data ->> 'version_past'::text) IS NULL OR (A.data -> 'version_past'::text) = 'null'::jsonb)
AND ((B.data ->> 'version_past'::text) IS NULL OR (B.data -> 'version_past'::text) = 'null'::jsonb)
AND ((A.data ->> 'deleted'::text) IS NULL OR (A.data -> 'deleted'::text) = 'null'::jsonb)
AND ((B.data ->> 'deleted'::text) IS NULL OR (B.data -> 'deleted'::text) = 'null'::jsonb)
AND ((A.data ->> 'upload_started'::text) IS NULL OR (A.data -> 'upload_started'::text) = 'null'::jsonb)
AND ((B.data ->> 'upload_started'::text) IS NULL OR (B.data -> 'upload_started'::text) = 'null'::jsonb)
AND (A.data->'version_seq' < B.data->'version_seq'));
```

this query runs forever and even when specifing bucket id it doesn't work for large dataset

src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
```diff
@@ -0,0 +1,99 @@
...
+        // 1. update old versions of an object with version_past = true
+        // 2. drop the current index - idx_btree_objectmds_latest_version_index
+
+        // idx_btree_objectmds_latest_version_index postgres index is incorrect since 
+        // postgres considers nulls as distinct, causes multiple latest versions of objectmds to be allowed
+        // recreation of the index is being handled lazily in md_store constructor
+
+        pg_client_cursor = await db_client.instance().pool.connect();
+        pg_client = await db_client.instance().pool.connect();
+        cursor = await db_client.instance().sql_query(pg_client_cursor,
```
guymguym, line 29: I would prefer that we move the pool connect into sql_query if possible.
Do we have to call client.release() explicitly after using it? 
For the cursor case is it enough to call close on the cursor to release back the connection to the pool?

src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
```diff
@@ -0,0 +1,99 @@
...
+
+        // 2. drop index
+        try {
+            const index_short_name = `latest_version_index;`;
+            // Do not use direclty objectmds collection of db_client 
+            // so it won't create schema changes here
+            const drop_res = await db_client.instance().drop_index(pg_client, 'objectmds', index_short_name);
+            dbg.log0('upgrade_objectmds: dropped index successfully', drop_res);
+        } catch (err) {
+            if (err.code !== '42704') throw err;
```
guymguym, line 79: handling the "index not found" error should be inside drop_index()

src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
```diff
@@ -0,0 +1,86 @@
...
+                AND A.data->'version_seq' < B.data->'version_seq'`;
+
+            for (let bucket of versioned_buckets) {
+                try {
+                    dbg.log0('upgrade_objectmds: updating bucket', bucket._id);
+                    const query = `${update} WHERE _id in (${select_fields} WHERE
+                        A.data->>'bucket'='${bucket._id}' AND B.data->>'bucket'='${bucket._id}'
+                        AND ${select_by});`;
+
+                    const res = await md_store._objects.single_query(query);
```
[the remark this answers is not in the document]
romayalon: @dannyzaken @guymguym I added sql_query, Danny - any objections to changing single_query() of PgTable to be private?

src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
romayalon: resolving since we fixed it together.

src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
romayalon: According to pg documentation: https://node-postgres.com/features/pooling
```
You must always return the client to the pool if you successfully check it out,
regardless of whether or not there was an error with the queries you ran on the client.
If you don't check in the client your application will leak them and 
eventually your pool will be empty forever and all future requests to check out a client from the pool will wait forever.

```
And regarding the cursor - in pg doc example they call client.release() right after cursor.close() so closing cursor will not release the connection back to the pool.

Also, the update query is inside a loop so moving the pool.connect() and client.release() to sql_query() will cause extra work of acquiring and releasing a connection for each iteration. On the other hand, maybe keeping the same connection for all the iterations
might be a worse behavior. WDYT? @guymguym

src/server/object_services/schemas/object_md_indexes.js
romayalon: let's do it in a different PR so we will keep PRs small

src/upgrade/upgrade_scripts/5.12.0/upgrade_objectmds_old_versions.js
guymguym: ok in that case I would prefer to have a separate method that would encapsulate it inside the pg client e.g:

```js
async sql_query_batch(query, values, batch_size, batch_handler) {
  let client;
  try {
    client = await this.pool.connect();
    let cursor;
    try {
      cursor = client.query(new Cursor(query, values));
      for (;;) {
        const rows = await cursor.read(batch_size);
        if (!rows.length) break;
        const should_continue = await batch_handler(rows);
        if (!should_continue) break;
      }
    } finally {
      if (cursor) await cursor.close();
    }
  } finally {
    if (client) await client.release();
  }
}

```

romayalon: sounds good, I updated the PR and left the update queries pg_client in the upgrade script instead of creating it and releasing it on every update.

src/server/object_services/schemas/object_md_indexes.js
romayalon: I created another PR that fixes null_version_index

romayalon:
closing this PR, since we decided to go without an automatic upgrade (https://github.com/noobaa/noobaa-core/pull/7044 was merged)
the manual upgrade script will be reviewed on slack.

PR #7027, opened 2022-07-28
Adding update_nsfs_stats to stats_api

liranmauda:
### Explain the changes
1. Adding update_nsfs_stats to stats_api

Signed-off-by: liranmauda <liran.mauda@gmail.com>

PR #7033, opened 2022-08-08
Send the nsfs stats from the endpoint to the core

liranmauda:
### Explain the changes

Send the nsfs stats from the endpoint to the core

This PR continues the work from #7026 and #7027 

### DIFF
--- a/src/sdk/endpoint_stats_collector.js
+++ b/src/sdk/endpoint_stats_collector.js
@@ -18,6 +18,7 @@ class EndpointStatsCollector {
     constructor(rpc_client) {
         this.rpc_client = rpc_client;
         this.reset_all_stats();
+        this.nsfs_counters = this._new_namespace_stats();
         this.prom_metrics_report = prom_report.get_endpoint_report();
     }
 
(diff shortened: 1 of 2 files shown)


src/sdk/endpoint_stats_collector.js
```diff
@@ -37,9 +37,9 @@ class EndpointStatsCollector {
         // clear this.send_stats to allow new updates to trigger another _send_stats
         this.send_stats = null;
         try {
-            await this.rpc_client.object.update_endpoint_stats(this.get_all_stats(), {
+            await P.all([await this.rpc_client.object.update_endpoint_stats(this.get_all_stats(), {
```
jackyalbo, line 40: no need for the awaits inside of the P.all

src/sdk/endpoint_stats_collector.js
```diff
@@ -37,9 +37,9 @@ class EndpointStatsCollector {
...
                 timeout: SEND_STATS_TIMEOUT
-            });
+            }), await this._send_nsfs_stats()]);
```
jackyalbo, line 42: I think it's better to change _send_stats to _send_endpoint_stats and to keep the old logic inside and in _send_stats just leave the P.all calling the 2 funcs, and the parts that we need for both calls.

src/sdk/endpoint_stats_collector.js
```diff
@@ -48,6 +48,25 @@ class EndpointStatsCollector {
         }
     }
 
+    async _send_nsfs_stats() {
+        // LMLM Not sure if there is a point of retry sending it, We can send it in the next iteration.
+        // this.send_nsfs_stats = null;
+        // LMLM TODO: remove the log??
+        const _nsfs_stats = { nsfs_stats: { io_stats: this.nsfs_counters } };
+        dbg.log0(`LMLM _nsfs_stats: ${JSON.stringify(_nsfs_stats)}`);
```
jackyalbo, line 56: Clean or change all the LMLM prints and remarks, or follow on them for the next review cycles.

jackyalbo (changes requested):

jackyalbo (approved):

PR #7034, opened 2022-08-08
server: Updated the error status in case of non-existent account

### LINKED ISSUE #5747: Checking external connection to azure fails with wrong status
### Environment info
- NooBaa Version: master
- Platform: local

### Actual behavior
1. When adding external connection to a storage account which doesn't exist in azure, we are return INVALID_ENDPOINT when the problem isn't in the endpoint but in the account name

### Expected behavior
1. we should return that the problem is with the account name or in the credentials not in the endpoint which we provide to the user.

### Steps to reproduce
1. try to create an azure account to a storage account which doesn't exist

RPC REQUEST SEND account_api.check_external_connection {endpoint_type: "AZURE", endpoint: "https://blob.core.windows.net", identity: "jackystorage3", secret: "secret"}

RPC REQUEST REPLY account_api.check_external_connection ==> 
{status: "INVALID_ENDPOINT", error: {…}}
error:
code: "ENOTFOUND"
message: "getaddrinfo ENOTFOUND jackystorage3.blob.core.windows.net jackystorage3.blob.core.windows.net:443"
__proto__: Object
status: "INVALID_ENDPOINT" <=== wrong!
__proto__: Object

nik-redhat:
Signed-off-by: nik-redhat <nladha@redhat.com>

### Explain the changes
1. Added a check for `ENOTFOUND` error code for external connection `azure` .

1. Fixed #5747

### DIFF
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -1055,8 +1055,11 @@ async function _check_azure_connection_internal(params) {
         if (err.code === 'AuthenticationFailed' &&
             err.authenticationerrordetail && err.authenticationerrordetail.indexOf('Request date header too old') > -1) {
             throw err_to_status(err, 'TIME_SKEW');
+        } else if (err.code === 'ENOTFOUND') {
+            throw err_to_status(err, 'UNKNOWN_FAILURE');
+        } else {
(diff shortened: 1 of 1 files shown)


nik-redhat:
@jackyalbo can you please review the PR.
TIA :)

baum (approved):
LGTM. Thank you @nik-redhat 🖖

PR #7040, opened 2022-08-11
Fixed map_common_prefixes queries

dannyzaken:

### Explain the changes
* Changed pl/sql function `map_common_prefixes` to fetch all required entries in one call.
* instead of querying for 1000 keys and then extracting the common prefixes from it (can be very few common prefixes if there are a lot of objects in that "dir"), it iterates over the objectmds until reaching the max keys or the end.
* To avoid unnecessary scanning of objects - if we find a common prefix, we set the marker to be the next possible value for a common prefix (by incrementing the last character by 1)
* These changes rely on the DB having collation of LC_COLLATE = 'C'. it will not work on DBs with utf8 collation

1. More cleanup can be done after this change. for example, it is probably unnecessary to redo the list_object queries in the object_server, since now it should return all of the required count
https://github.com/noobaa/noobaa-core/blob/5b6bcb568bbb699e735523067b83a37e622e8db9/src/server/object_services/object_server.js#L1028-L1031

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -835,29 +835,21 @@ class PostgresTable {
         let mr_q;
         sql_query.where = mongo_to_pg('data', encode_json(this.schema, options.query), { disableContainmentQuery: true });
         sql_query.order_by = options.sort && convert_sort(options.sort);
-        sql_query.limit = options.limit;
-        let query_string = `SELECT * FROM ${this.name} WHERE ${sql_query.where}`;
-        if (sql_query.order_by) {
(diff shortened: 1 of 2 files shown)


liranmauda (commented):
@dannyzaken  Have we tested Upgrade? 
In the past, we had issues when we changed sql_functions

liranmauda (commented):
Other then the comment, LGTM

liranmauda (approved):

PR #7042, opened 2022-08-15
Fix Postgres null version index and handle upgrade

romayalon:

### Explain the changes
1. Similar to https://github.com/noobaa/noobaa-core/pull/7031 but handles null version index, will drop the duplicated changes from this PR when  #7031 will be merged.

1. On which version the upgrade should be?

### Testing Instructions:
1. Install old noobaa on pg (4.7 and up).
2. Put bucket.
3. Put multiple objects of the same key and bucket in parallel.
4. Head object - you might get an old version of the object (not version as in versioning but not the most updated data/metadata that was already overridden ).
5. Upgrade NooBaa version.
6. Head object again and get the latest version. (not version as in versioning but not the most updated data/metadata).
7. Try step 4 again but this time step 5 will always return the latest version of the object.

- [ ] Doc added/updated
- [ ] Tests added


romayalon:
closing this PR, since we decided to go without an automatic upgrade (https://github.com/noobaa/noobaa-core/pull/7044 was merged)
the manual upgrade script will be reviewed on slack.

PR #7057, opened 2022-08-30
Add resource access mode validation

kfir-payne:

### Explain the changes
1. Add resource check access mode function.
2. Add access mode check to every Write privilege operation in `object_sdk`.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/object_sdk.js
+++ b/src/sdk/object_sdk.js
@@ -419,6 +419,12 @@ class ObjectSDK {
         return this.rpc_client.options.auth_token;
     }
 
+    check_is_readonly_namespace(ns) {
+        if (ns.is_readonly_namespace()) {
+            throw new RpcError('UNAUTHORIZED', `Read Only namespace bucket`);
+        }
+    }
+
     /////////////////
     // OBJECT LIST //
     /////////////////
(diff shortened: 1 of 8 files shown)


src/sdk/object_sdk.js
```diff
@@ -585,22 +593,26 @@ class ObjectSDK {
...
     async upload_multipart(params) {
         const ns = await this._get_bucket_namespace(params.bucket);
+        this.check_is_readonly_namespace(ns);
         if (params.copy_source) await this.fix_copy_source_params(params, ns);
         return ns.upload_multipart(params, this);
     }
 
     async list_multiparts(params) {
         const ns = await this._get_bucket_namespace(params.bucket);
+        this.check_is_readonly_namespace(ns);
```
liranmauda, line 603: Do we need it on a list operation? 
Or, Can a read Only list multiparts?

kfir-payne: it's a list op on the `multiparts`, so I think it shouldn't be available on a read only resource,
but I would be happy to hear what the team thinks.

dannyzaken: I think we can remove the check here. access can be blocked on the source itself

dannyzaken (approved):

PR #7065, opened 2022-09-13
[Fix]: allow admin account to perform any action on the system

tangledbytes:

### Explain the changes
`allowed_buckets` PR did not give any special privilege to system owner account which was **not** desirable. This PR restores _super_ privileges for the system owner account.

- [ ] Doc added/updated
- [ ] Tests added


jackyalbo (approved):
This is a temporary fix - we will remove admin in the future - operator account will be system owner.

PR #7077, opened 2022-09-21
Allow anonymous access to object_api.list_objects

### LINKED ISSUE #7036: Allow list_objects for anonymous users
We want to have an anonymously accessible (read-only) s3 bucket you can navigate through a webbrowser.
We noticed that while all necessary operations were allowed the call still failed with a 403.

`Aug-7 19:54:01.610 [Endpoint/13] [L0] core.endpoint.s3.s3_utils:: bucket_policy: action fit? s3:* s3:listbucket
Aug-7 19:54:01.610 [Endpoint/13] [L0] core.endpoint.s3.s3_utils:: bucket_policy: principal fit? SENSITIVE-7846cdd4c2b90527 undefined
Aug-7 19:54:01.610 [Endpoint/13] [L0] core.endpoint.s3.s3_utils:: bucket_policy: resource fit? /^arn:aws:s3:::website-foo-62044d78-39c5-4b65-9e0b-998350fda26c\/.*$/ arn:aws:s3:::website-foo-62044d78-39c5-4b65-9e0b-998350fda26c
Aug-7 19:54:01.610 [Endpoint/13] [L0] core.endpoint.s3.s3_utils:: bucket_policy: is_statements_fit true true false
Aug-7 19:54:01.610 [Endpoint/13] [ERROR] CONSOLE:: RPC._on_request: ERROR srv object_api.list_objects reqid 230@fcall://fcall(7pacxw7c) connid fcall://fcall(7pacxw7c) Error: not anonymous method list_objects
at RpcRequest.req.check_anonymous (/root/node_modules/noobaa-core/src/server/common_services/auth_server.js:528:44)
at authorize (/root/node_modules/noobaa-core/src/server/common_services/auth_server.js:436:17)
at RPC._on_request (/root/node_modules/noobaa-core/src/rpc/rpc.js:323:27)`

Troubleshooting has been done with @guymguym and he came to the conclusion that this call is not allowed for anonymous users.
[ref](https://github.com/noobaa/noobaa-core/blob/12b8a5beeb518d876f4fec22f12d714f57e47c17/src/api/object_api.js#L749)

Since the policy allows it the call to list_objects should also be allowed.

In other cases we need to share either keys (directly or indirectly) and explain to users how to setup.
There is an existing s3 explorer project however it's not built right now for authentication and with auth we need to share keys with the app, change endpoint to use ODF instead of aws, ...
For each bucket we also need to create this setup again.

tangledbytes:

### Explain the changes
This PR simply enables `anonymous` requests to the `list_objects` API. The implications of the changes being that:
1. If a user allows anonymous access to a bucket via a bucket policy then anyone will be able to access the data.
2. RPCs as have no longer any authorization requirement, will be able to access this data without offering _any_ credentials. Implying, the data won't be protected by the policy in case of RPCs.

### Issues: Fixes #7036

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -746,7 +746,10 @@ module.exports = {
                     }
                 }
             },
-            auth: { system: ['admin', 'user'] }
+            auth: {
+                system: ['admin', 'user'],
+                anonymous: true,
+            }
         },
 
         list_object_versions: {
--- a/src/test/unit_tests/test_s3_bucket_policy.js
+++ b/src/test/unit_tests/test_s3_bucket_policy.js
@@ -1,9 +1,11 @@
 /* Copyright (C) 2016 NooBaa */
+/* eslint max-lines-per-function: ['error', 650] */
 'use strict';
 
 // setup coretest first to prepare the env
 const coretest = require('./coretest');
-const { rpc_client, EMAIL, POOL_LIST } = coretest;
+const { rpc_client, EMAIL, POOL_LIST, anon_rpc_client } = coretest;
+const MDStore = require('../../server/object_services/md_store').MDStore;
 coretest.setup({ pools_to_create: POOL_LIST });
 
 const AWS = require('aws-sdk');
@@ -29,6 +31,7 @@ const user_b = 'bob';
 let s3_a;
 let s3_b;
 let s3_owner;
+let s3_anon;
 
 const anon_access_policy = {
     Version: '2012-10-17',
@@ -39,6 +42,10 @@ const anon_access_policy = {
         Resource: [`arn:aws:s3:::*`]
     }]
 };
+
+// core_test_store is a global object created with the intention
+// to store data between tests.
+const cross_test_store = {};
 
 async function setup() {
     const self = this; // eslint-disable-line no-invalid-this
@@ -79,6 +86,11 @@ async function setup() {
     s3_creds.secretAccessKey = admin_keys[0].secret_key.unwrap();
     s3_owner = new AWS.S3(s3_creds);
     await s3_owner.createBucket({ Bucket: BKT }).promise();
+    s3_anon = new AWS.S3({
+        ...s3_creds,
+        accessKeyId: undefined,
+        secretAccessKey: undefined,
+    });
 }
 
 mocha.describe('s3_bucket_policy', function() {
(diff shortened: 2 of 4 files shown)


NotABugItsAFeature:
Any update on this? :)

src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -458,4 +464,24 @@ mocha.describe('s3_bucket_policy', function() {
...
+        await s3_owner.deleteBucketPolicy({
+            Bucket: BKT,
+        }).promise();
+
+        await assert_throws_async(s3_anon.makeUnauthenticatedRequest(
+            'listObjects',
+            { Bucket: BKT }
+        ).promise());
+    });
 });
```
liranmauda, line 487: Maybe add another test that tests a policy of a deny and then 
```
anonymous user should not be able to list bucket objects when policy doesn\'t allow
```

jackyalbo:
Hi @utkarsh-pro , the PR by itself looks good, I'm just more worried about the side effect that you mention: Is that saying that with this fix anyone can list a bucket using RPC (like no need for auth) or does this means that every authenticated user can do that (which I think was right before this change as well)?

utkarsh-pro:
> Hi @utkarsh-pro , the PR by itself looks good, I'm just more worried about the side effect that you mention: Is that saying that with this fix anyone can list a bucket using RPC (like no need for auth) or does this means that every authenticated user can do that (which I think was right before this change as well)?

Former one @jackyalbo.
With the `allowed_buckets` PR, we removed the logic of permission checking from the RPC servers completely under the assumption that the request will be anyway authorized by the `s3_rest`, so in RPC servers all we care about is if `anonymous` access is allowed to an API or not, nothing more. Our assumption holds true for normal cases but won't hold true when someone initiates a RPC directly to the RPC servers.

Example: Someone creates a pod in k8s and sends a RPC request to noobaa to retrieve the list of all the buckets, they will be able to do so after this change without offering any token. This would be even more problematic if someone exposes noobaa rpc servers via LoadBalancer service to the public internet (which normal noobaa installation doesn't do) because then anyone will be able to just initiate a RPC to retrieve bucket list.

guymguym:
good points @utkarsh-pro @jackyalbo 

Ok I think we need to add to list_objects rpc handler a call to `req.has_s3_bucket_permission()`  like we do in list_buckets.

https://github.com/noobaa/noobaa-core/blob/bfab783def19e332e7d5429a8152af7184ee4290/src/server/system_services/bucket_server.js#L850-L854

This function is defined by auth_server.js and checks that if the call has no authentication info then the bucket itself has to be configured to allow anonymous access, so an outsider cannot bypass that check by going to the rpc server directly -

https://github.com/noobaa/noobaa-core/blob/bfab783def19e332e7d5429a8152af7184ee4290/src/server/common_services/auth_server.js#L602-L621

You can see that we once though all load bucket calls might want to check that here, but thought it should go into specific flows instead -

https://github.com/noobaa/noobaa-core/blob/bfab783def19e332e7d5429a8152af7184ee4290/src/server/object_services/object_server.js#L1401-L1402

But perhaps the same check is also missing in other existing anonymous api's like read_object_md, etc ?

utkarsh-pro:
That makes sense @guymguym. Yes, I think it is also missing from pre-existing anonymous APIs. I'd add the check to those APIs as well.

On a related note, it recently occurred to me that not having authorization in RPC servers might be problematic for multi-tenant deployments as well (here, by multi-tenant I mean - 1 NooBaa deployment in a K8s cluster being shared by multiple users). Not sure how valid or important this use case is though.

guymguym:
@utkarsh-pro Do you want to describe what is the exact rpc auth issue that sharing the cluster with other users has? we can try to figure out together if there's a simple way to allow it, because rpc auth is quite flexible.

utkarsh-pro:
> @utkarsh-pro Do you want to describe what is the exact rpc auth issue that sharing the cluster with other users has? we can try to figure out together if there's a simple way to allow it, because rpc auth is quite flexible.

@guymguym, I didn't realise that we do check the `role` of the requesting account if the user is authenticated. So, I think we are good in that regard.

src/api/object_api.js
```diff
@@ -746,7 +746,10 @@ module.exports = {
                     }
                 }
             },
-            auth: { system: ['admin', 'user'] }
+            auth: {
+                system: ['admin', 'user'],
+                anonymous: true
```
liranmauda, line 751:
```suggestion
                anonymous: true,
```

liranmauda (commented):
@utkarsh-pro 
Due to the fact that we added `has_s3_bucket_permission` in `list_objects`, `read_object_md`, and `read_object_mapping`, Do we want to add unit tests for those?

utkarsh-pro:
@liranmauda, would you recommend adding tests:
- Test to ensure that a user cannot access any resource but their own.
- Test to ensure that a user can keep accessing their resources and addition of `has_s3_bucket_permission` hasn't caused a regression? (I think this might be redundant test because other tests already rely on this RPC under the assumption that they have access to this resource)

liranmauda:
>@liranmauda, would you recommend adding tests:
>
>Test to ensure that a user cannot access any resource but their own.
>Test to ensure that a user can keep accessing their resources and addition of has_s3_bucket_permission hasn't caused a regression?

@utkarsh-pro I think we should add tests that per each op, a user can or cannot access a resource according to his permission.

regarding the second point, It is always welcome.

src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -1,9 +1,11 @@
 /* Copyright (C) 2016 NooBaa */
+/* eslint max-lines-per-function: ['error', 650] */
```
liranmauda, line 2: Maybe we can devide all the RPC tests here to a new mocha.describe('[RPC TEST]: s3_bucket_policy' ...)

that way we will not need to disable this rule. 
One thing that we will need to do is taking the before out of the describe a maybe use `mocha.beforeEach` and also do a `mocha.afterEach` that will remove the bucket and setup.

this way we might also be able to avoid using a global object to store data between tests.

liranmauda (approved):
Apart from the comment I wrote, LGTM, We can merge and follow the comment in later PR.

PR #7089, opened 2022-10-18
Collecting FS worker stats and sending it to the to `/metrics/nsfs_stats`

liranmauda:
### Explain the changes
- Added fs_workers_stats to common_api and stats_api
- Renamed fs_config_param to fs_context to reflect this object nature better
- Added report_fs_stats function to be called in the fs_napi
- Add timing report of the fs Worker in the fs_napi
- Publishing the report of the fs Workers in the `/metrics/nsfs_stats` route

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -151,6 +151,13 @@ set_statfs_res(Napi::Object res, Napi::Env env, struct statfs &statfs_res)
 
 }
 
+void 
+set_fs_worker_stats(Napi::Env env, Napi::Object fs_worker_stats, std::string work_name, double took_time, int error){
+    fs_worker_stats["name"] = Napi::String::New(env, work_name);
+    fs_worker_stats["took_time"] = Napi::Number::New(env, took_time);
+    fs_worker_stats["error"] = Napi::Number::New(env, error);
+}
+
 /**
  * FSWorker is a general async worker for our fs operations
  */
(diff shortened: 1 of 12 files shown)


jackyalbo:
Please add design doc

src/native/fs/fs_napi.cpp
```diff
@@ -180,17 +183,21 @@ struct FSWorker : public Napi::AsyncWorker
...
+        if (fs_context.Has("gid")) _gid = fs_context.Get("gid").ToNumber();
+        if (fs_context.Has("backend")) _backend = fs_context.Get("backend").ToString();
+        if (fs_context.Has("warn_threshold_ms")) _warn_threshold_ms = fs_context.Get("warn_threshold_ms").ToNumber();
+        if (fs_context.Has("report_fs_stats")) _report_fs_stats = Napi::Persistent(fs_context.Get("report_fs_stats").As<Napi::Function>());
+
     }
     void Begin(std::string desc)
     {
         _desc = desc;
+        _work_name = _desc.substr(0, _desc.find(" "));
```
jackyalbo, line 200: maybe instead of this hack we should just add work_name to the Begin function variables?

liranmauda: Talking to @jackyalbo we think we don't need to add a work_name to the Begin function.

src/native/fs/fs_napi.cpp
```diff
@@ -223,12 +230,27 @@ struct FSWorker : public Napi::AsyncWorker
     virtual void OnOK() override
     {
         DBG1("FS::FSWorker::OnOK: undefined " << _desc);
+        Napi::Env env = Env();
+        if (!_report_fs_stats.IsEmpty()) {
```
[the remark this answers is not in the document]
liranmauda: Great catch 😄

jackyalbo (approved):
Looks better, for next times I think we should add design doc in advance especially if the reviewer is not part of the design cycle... A side of the unclear design it looks fine, please add the design doc asap.

PR #7020, opened 2022-07-25
NSFS | Folders extended attributes 

romayalon:
Signed-off-by: Romy <romy2232@gmail.com>

### Explain the changes
1. FS Napi - added dir_content xattr to the USER_XATTRS vector.
2. test_namespace_fs - added unit tests
3. namespace_fs - 

    b. **list objects -** 
        - Skipped processing entry when entry is .folder file. 
        - Checked existence of XATTR_DIR_CONTENT xattr in the dir's xattr in process_dir and when evaluated to true the entry will be added to the objects list (when no delimiter or in other corner cases)

    c. **read_object_md -** added true to the call of get_version_path in order to get the directory path and not the .folder file when object is a directory content object.

    d. **upload_object / complete_multipart_upload -** 
d.1. upload_object - 
        * Created the .folder file only when content size > 0 when the entry is a directory.
        * Created _create_empty_dir_content function for the creation of directory objects of size 0 -  this function created 
               the directory key, sets the xattr on the directory, deleted the old directory if exists, and returns directory info as a 
               upload_info response.
       * On non-empty dir content - added a check to finish_upload() to not set the xattr on the object itself but put it on the 
       directory instead.
d.2. mpu - 
        * we always create a .folder file as we assume the size is always > 0
        * when object is a dir content object - 
               - skipped set xattr on .folder. 
               - replaced old xattr with new xattr and added XATTR_DIR_OBJ xattr to the directory. 
               - no override case - as we assume size > 0.

    e. **delete_object / delete_multiple_objects -** 
        - Unlink - skipped when the file is .folder and error code is ENOENT (needed for directory objects of size 0)
        - Called clear xattr on directories ( needed for non-empty directory objects )
        
     f. **get_object -**  called stat when the file is .folder and checked if XATTR_DIR_CONTENT xattr exists, if yes, returned {};  (needed for directory objects of size 0)

### Issues: Fixed #xxx / Gap #xxx
1. gap - upgrade script for existing directory objects (need to copy .folder xattr to the folder itself)
2. Fixed #https://github.com/noobaa/noobaa-core/issues/6980

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

PR #7090, opened 2022-10-18
NSFS | Native xattr api additions

romayalon:

### Explain the changes
1. moved out xattr changes from https://github.com/noobaa/noobaa-core/pull/7020 so it can be used by @jackyalbo for the encryption feature.
2. get_single_xattr() added.
3. setXattr(xattr) was changed to replaceXattr(xattr, prefix_of_xattrs_to_clear)
4. Added unit tests

1. GAP - ENODATA error (No data available) is thrown when getsinglexattr() is called but the corresponding attr doesn't exist, in fs_napi - err.code is set by calling uv_translate_sys_error, but ENODATA is not translated by this function and we get an unknown system error instead, opened the following to track it https://github.com/libuv/libuv/issues/3795
In the meantime, I used error.message in tests instead err.code.
### Testing Instructions:
1. 

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -761,7 +761,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
                 InstanceMethod<&FileWrap::read>("read"),
                 InstanceMethod<&FileWrap::write>("write"),
                 InstanceMethod<&FileWrap::writev>("writev"),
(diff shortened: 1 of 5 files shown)


jackyalbo (approved):
LGTM, I would maybe add a test for an attribute key that doesn't exist, and maybe for changing an existing key. WDYT?

romayalon:
Thanks @jackyalbo 
I added these tests :)

PR #7111, opened 2022-11-27
change in KeyCount field in s3 reply

shirady:
change in KeyCount field in s3 reply (is used in list object command). 

### Explain the changes
1. add the number of prefixes to the KeyCount field

1.  test_bucket_listv2_delimiter_basic was failing (1 != 3).

### Testing Instructions:
1. https://gist.github.com/dannyzaken/c85f0006e5d6582c8d8666cafce6ab76

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_get_bucket.js
+++ b/src/endpoint/s3/ops/s3_get_bucket.js
@@ -46,7 +46,7 @@ async function get_bucket(req) {
                 ...(list_type === '2' ? {
                     ContinuationToken: cont_tok,
                     StartAfter: start_after,
-                    KeyCount: reply.objects.length,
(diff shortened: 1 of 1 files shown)


guymguym (approved):
That's good catch!

liranmauda (commented):
@shirady @guymguym a question, 
if we will not get a delimiter, we will not get common_prefixes, and this field will be undefined. 
adding it will lead to NaN.
Am I correct?

shirady:
Hi @liranmauda ,
If we will not get a delimiter the common_prefix will be an empty array.

dannyzaken (approved):

PR #7117, opened 2022-12-01
Add workflow for automated NooBaa releases

tangledbytes:

### Explain the changes
This PR attempts to add a Github workflow to automate releases of **both** NooBaa core and NooBaa operator. The PR changes does the following:
1. Release NooBaa core OCI images to dockerhub and quay registries.
2. Dispatch a release in the NooBaa operator repository by triggering workflow in PR: https://github.com/noobaa/noobaa-operator/pull/998.
3. Update the version in the `package.json` and `package-lock.json`.

image

Flow:
1. Read the version to be released from `package.json`.
2. Build OCI images for the version read.
3. Push the OCI images to Quay and Dockerhub.
4. Dispatch workflow in the `sync_operator_repository` (format: `repo`).
5. Increment version in `package.json` and `package-lock.json`.
6. Commit the changes in the tree to the base branch.

Caveats:
- The entire workflow assumes that the version in the `package.json` is correct. It does not have additional logic to ensure that. Whatever the version would be in that file, it will be released.
- Rollbacks are manual for now. That is,  rollback sync operator changes, delete OCI images and revert the commits. Can be automated in the future though.
- Whenever a new version branch is created, we need to ensure that `package.json` reflects that. So when we move to 5.14 branch, `package.json` needs to be set to "5.14.0". This can be automated as well but not sure if its worth doing so.

The scripts/workflow will require a few secrets in place like:
1. `GHACTION_GH_PAT` - A github token which has write access to the repositories. This is needed because the workflow attempts to synchronize a release in the operator repository as well by dispatching the release workflow there.
8. `DOCKERHUB_USERNAME` - Username in dockerhub where the OCI image is supposed to be pushed.
9. `DOCKERHUB_TOKEN` - Dockerhub token is used to login to the registry and push the image.
10. `QUAY_USERNAME` - Username in quay where the OCI image is supposed to be pushed.
11. `QUAY_TOKEN` - Quay token is used to login to the registry and push the image.
12. `OCI_ORG` - This is used internally hence this can be anything. It can be same as `DOCKERHUB_USERNAME` or `QUAY_USERNAME` or a random string.

### DIFF
--- a/.github/workflows/releaser.yaml
+++ b/.github/workflows/releaser.yaml
@@ -1,0 +1,45 @@
+name: Releaser
+
+on:
+  workflow_dispatch:
+    inputs:
+      base_branch:
+        description: 'The base branch to release from'
+        required: true 
+      sync_operator_repository:
+        description: "Sync operator repository"
+        required: false
+        default: "noobaa-operator"
+
+permissions:
+  contents: write
+
+jobs:
+  releaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set branch
+        run: echo "BRANCH=${{ github.event.inputs.base_branch }}" >> $GITHUB_ENV
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ env.BRANCH }}
+      - name: Fetch all tags
+        run: git fetch --force --tags
+      - uses: actions/setup-node@v3
+        with:
+          node-version-file: .nvmrc
+      - name: Release NooBaa Core Image
+        env:
+          GITHUB_TOKEN: ${{ secrets.GHACTION_GH_PAT }}
+          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+          QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+          OCI_ORG: ${{ secrets.OCI_ORG }}
+          BASE_BRANCH: "${{ github.event.inputs.base_branch }}"
+        run: |
+          git config --global user.email "github-action@noobaa.io"
+          git config --global user.name "NooBaa GitHub Action"
+
+          bash tools/releaser.sh --oci-org $OCI_ORG --sync-operator-repository ${{ github.event.inputs.sync_operator_repository }} --gh-org noobaa 
--- a/tools/releaser.sh
+++ b/tools/releaser.sh
@@ -1,0 +1,206 @@
+#!/bin/bash
+
+set -euo pipefail
+set -x
+
+export PS4='\e[36m+ ${FUNCNAME:-main}@${BASH_SOURCE}:${LINENO} \e[0m'
+
+dir=$(dirname "$0")
+
+# Default values for the arguments.
+GH_ORG="noobaa"
+GH_REPO="noobaa-core"
+OCI_ORG="noobaa"
+DRY_RUN="false"
+SYNC_OPERATOR_REPOSITORY=""
+
+# Default values for the environment variables.
+GITHUB_TOKEN=${GITHUB_TOKEN:-}
+DOCKERHUB_USERNAME=${DOCKERHUB_USERNAME:-}
+DOCKERHUB_TOKEN=${DOCKERHUB_TOKEN:-}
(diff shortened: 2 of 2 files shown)


utkarsh-pro:
This PR assumes that the version is `package.json` is the version that is supposed to be released. This assumption does not hold true for our `5.12` branch where the version is set to `5.13.0`. Before we start using this workflow, we will need to ensure that the version is corrected there.

liranmauda:
Versions of the branches:
5.8.18, 5.9.14, 5.10.10, 5.11.5

dannyzaken (approved):

.github/workflows/releaser.yaml
```diff
@@ -0,0 +1,45 @@
...
+          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+          QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+          OCI_ORG: ${{ secrets.OCI_ORG }}
+          BASE_BRANCH: "${{ github.event.inputs.base_branch }}"
+        run: |
+          git config --global user.email "github-action@noobaa.io"
+          git config --global user.name "NooBaa GitHub Action"
+
+          bash tools/releaser.sh --oci-org $OCI_ORG --sync-operator-repository ${{ github.event.inputs.sync_operator_repository }} --gh-org noobaa
```
liranmauda, line 45: Add new line

liranmauda (dismissed):

.github/workflows/releaser.yaml
```diff
@@ -0,0 +1,45 @@
...
+        run: echo "BRANCH=${{ github.event.inputs.base_branch }}" >> $GITHUB_ENV
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ env.BRANCH }}
+      - name: Fetch all tags
+        run: git fetch --force --tags
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16.14.0
```
liranmauda, line 31: Change to node 18.12.1
Can we take it out of the .nvmrc?

tools/releaser.sh
```diff
@@ -0,0 +1,202 @@
...
+  if [[ $# -eq 1 ]]; then
+    if [[ "$1" == "1" ]]; then
+      version_with_v "$version"
+    else
+      version_without_v "$version"
+    fi
+  else
+    version_without_v "$version"
+  fi
+}
```
liranmauda, line 122: Samer as the operator version function.

tools/releaser.sh
```diff
@@ -0,0 +1,202 @@
...
+    echo "Skip Syncing operator, no repository was provided."
+    return
+  fi
+
+  gh workflow run releaser.yaml -R "$GH_ORG/$SYNC_OPERATOR_REPOSITORY" -f base_branch="$BASE_BRANCH"
+}
+
+function update_package_json() {
+  local version=$(bump_semver_patch $(get_noobaa_version))
+  npm version ${version} --no-git-tag-version
```
liranmauda, line 177: Is it also updating the package-lock file?

.github/workflows/releaser.yaml
utkarsh-pro: Using node-version-file now which reads the version from nvmrc file. Using https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#node-version-file.

tools/releaser.sh
utkarsh-pro: Yes, I checked that running `npm version <version> --no-git-tag-version` not only updates the version in `package.json` but also updates the `package-lock.json`.

tools/releaser.sh
```diff
@@ -0,0 +1,210 @@
...
+    if [[ "$1" == "1" ]]; then
+      version_with_v "$version"
+    else
+      version_without_v "$version"
+      echo "[Warn]: Unknown argument passed to get_noobaa_version."
+    fi
+  else
+    version_without_v "$version"
+  fi
+}
```
liranmauda, line 130: follow the get_noobaa_version in the operator.

utkarsh-pro: Actually this is from the operator.

liranmauda: So I missed it on the operator as well.
We cant use `echo "[Error]:` or `echo "[Warn]:` here unless we are handling this error on the call.
Currently, upon error or this warning, we will change the version to include this string.

liranmauda (approved):
LGTM

PR #7123, opened 2022-12-07
update tsconfig file with target es2021

shirady:

### Explain the changes
1. change target to es2021

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2020",
+    "target": "es2021",
     "module": "commonjs",
     "resolveJsonModule": true,
     "noEmit": true,


liranmauda (approved):
LGTM

tangledbytes (approved):

shirady:
I changed the target to es2021 because I wanted to use the `replaceAll` function (which is a feature in [es2021](https://www.telerik.com/blogs/learn-new-features-introduced-javascript-es2021)).
Before this change, I had this error (using Visual Studio as an IDE) :
```
[{
...
	"owner": "typescript",
	"code": "2550",
	"severity": 8,
	"message": "Property 'replaceAll' does not exist on type 'string'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2021' or later.",
	"source": "ts",
...
}]
```
I decided to change it to es2021 and not to es2022 because es2021 is fully supported by node 16, while es2022 is not yet. see [here](https://node.green/).

guymguym:
nice @shirady ! 
so far we had to use replace with `/regexp/g` in order to replace all, which is both an overkill and error-prone for replace all of an exact string...
also the logical-assignments operators are a nice addition so we can use to do async initializations, like for example -
```js
let nvmrc = '';

exports.get_nvmrc = async () => {
    nvmrc ||= await fs.readFile('.nvmrc', 'utf8');
    return nvmrc.replaceAll('.', '_');
};
```

PR #7124, opened 2022-12-07
Log based replication background worker and log parser

Neon-White:
Code originally written by Kfir Payne <kfirpayne@gmail.com>

### Explain the changes
1. Add log based replication background worker, responsible for scanning all the replication rules, and based on candidates provided by the log parser, decide which objects should be deleted from pre-existing replicas.
2. Add log parser, responsible for getting the next log to check, parse the log, and returning a list of keys (candidates) for the logs replication scanner to check.
3. Add a function to replication store that returns all the replication rules containing log based replication info.

This PR adds support for deletion of the replicas (replicated via currently existing replication scanner). AWS logs are **not** chronological in nature and also do **not** guarantee that each event will be captured in the logs (during testing it was observed that the logs kept missing crucial events such as of object deletion as well) which AWS calls "best effort log delivery". Due to these constraints we have the following caveats in the current implementation:
1. If the replication scanner encounters a delete type record, it will delete the object from the replica. However, if there was no log corresponding to the delete - the object will **not** be deleted from the replica.
2. Deliveries of the logs are excruciatingly slow. It may take more than an hour for a `DELETE` change to be replicated in a replica.
3. We most likely cannot implement a full fledged replication tool due to the fact that
    1. Events might be missed (eg. missing a `PUT` - that won't be good).
    2. Logs are not chronological. Logs do have timestamp in them which are accurate but that isn't of much help because the only guarantee that AWS gives is that timestamp in the name of the log file will always be greater than the timestamp of the events it stores (obviously, though).

- [ ] Doc added/updated
- [ ] Tests added


utkarsh-pro:
Lesson learnt. DeepScan does not understand optional chaining properly yet 🙂 .

src/server/bg_services/log_replication_scanner.js
```diff
@@ -0,0 +1,206 @@
...
+    async run_batch() {
+        if (!this._can_run()) return;
+        dbg.log0('log_replication_scanner: starting scanning bucket replications');
+        try {
+            if (!this.noobaa_connection) {
+                this.noobaa_connection = cloud_utils.set_noobaa_s3_connection(system_store.data.systems[0]);
+            }
+            await this.scan();
+        } catch (err) {
+            dbg.error('log_replication_scanner:', err, err.stack);
```
shirady, line 45: When you print `err` does it also prints the stack (`err.stack`)?
If so then you print the error stack twice.

src/server/bg_services/replication_log_parser.js
```diff
@@ -0,0 +1,320 @@
...
+const config = require('../../../config');
+const moment = require('moment');
+
+/**
+ * get_log_candidates will return an object which contains the log candidates
+ * @param {*} source_bucket_id ID of the source bucket
+ * @param {String} rule_id ID of the replication rule
+ * @param {Record<any, any>} replicationconfig Replication configuration
+ * @param {Number} candidates_limit Candidates limit
+ * @param {(log_entry: Record<string, any>) => Boolean} ignore_fn - function that returns true if the log entry should be ignored
```
shirady, line 20: In all other cases in this PR it is string (not `String`)

src/server/bg_services/replication_log_parser.js
```diff
@@ -0,0 +1,320 @@
...
+        // then no need to process anything
+        if (!next_log_entry.Contents || next_log_entry.Contents.length === 0) {
+            dbg.log0('log_parser: no more log entries');
+            break;
+        }
+
+        const next_log_data = await aws_get_next_log(s3, logs_bucket, next_log_entry.Contents[0].Key);
+        aws_parse_log_object(candidates_obj, next_log_data, ignore_fn);
+
+        logs_retrieved_count -= 1;
```
shirady, line 54: I don't understand the use of `logs_retrieved_count`, why do you change it and why is it inside the while loop?

src/server/system_services/replication_store.js
```diff
@@ -86,22 +86,69 @@ class ReplicationStore {
...
 
         return reduced_replications;
     }
+
+    async find_aws_log_based_replication_rules() {
+        dbg.log1('find_aws_log_based_replication_rules: ');
+        const replications = await this._replicationconfigs.find({ deleted: null });
+
+        const aws_reduced_replications = _.filter(replications, repl => repl.log_replication_info.aws_log_replication_info);
+        // TODO: Debug and fix
```
shirady, line 143: Why leave the code in the comment?
Are there cases where we will use it?

src/server/utils/replication_utils.js
```diff
@@ -0,0 +1,127 @@
...
+    // execute replication of object
+    const src_last_modified_recent = src_info.LastModified >= dst_info.LastModified;
+    if (!src_last_modified_recent) return false;
+
+    // data change - replicate the object
+    const data_change = src_info.ContentLength !== dst_info.ContentLength || src_info.ETag !== dst_info.ETag;
+    if (data_change) return true;
+
+    // md change - head objects and compare metadata, if metadata is different - copy object
+
```
shirady, line 22: According to your style before I think this new line is redundant.

src/server/utils/replication_utils.js
```diff
@@ -0,0 +1,127 @@
...
+                } catch (err) {
+                    dbg.error('replication_utils delete_objects: error: ', err, bucket_name, keys);
+                }
+            });
+        return res;
+    } catch (err) {
+        dbg.error('replication_utils delete_objects: semaphore error:', err, err.stack);
+    }
+}
+
```
shirady, line 121: According to your style before - add `// EXPORTS`.

src/server/bg_services/replication_log_parser.js
```diff
@@ -0,0 +1,320 @@
...
+                // ignore the log entry if it should be ignored
+                if (ignore_fn(log)) continue;
+
+                if (log.operation.includes('PUT.OBJECT') || log.operation.includes('POST.OBJECT')) {
+                    candidates_map[log.key] = {
+                        action: 'copy',
+                        time: log.time,
+                    };
+                }
+                if (log.operation.includes('DELETE.OBJECT') && log.http_status === 204) {
```
romayalon, line 128: is there a valid situation in which the heep_status will be other than 204 and the op is delete? it's not handled according to this code

src/server/system_services/replication_store.js
```diff
@@ -86,22 +86,69 @@ class ReplicationStore {
         return ans;
     }
 
+    async update_log_replication_status_by_id(_id, last_cycle_end) {
+        dbg.log1('update_log_replication_status_by_id: ', _id, last_cycle_end);
+        const parsed_id = db_client.instance().parse_object_id(_id);
+        const find = { _id: parsed_id};
```
romayalon, line 92: missing space

src/server/bg_services/log_replication_scanner.js
utkarsh-pro: Yeah, it seem redundant. Will remove it.

src/server/bg_services/replication_log_parser.js
utkarsh-pro: It is `string`, isn't that correct?

src/server/bg_services/replication_log_parser.js
utkarsh-pro: It is to ensure that only `AWS_LOG_CANDIDATES_LIMIT` logs are processed in a single iteration.

src/server/system_services/replication_store.js
utkarsh-pro: I think this came from Kfir's code. I don't know how relevant this comment is to others however it communicated to me that there is further filtering that can be done here in order to avoid multiple `ifs` in the rest of the code. Having said that, I don't know how buggy this commented code is (or if it is buggy at all).

config.js
```diff
@@ -539,6 +539,9 @@ config.BUCKET_REPLICATION_MAX_DST_BUCKETS = 100;
 config.REPLICATION_SEMAPHORE_CAP = 5000;
 config.REPLICATION_SEMAPHORE_TIMEOUT = 300 * 1000;
 config.REPLICATION_ENABLED = true;
+config.LOG_REPLICATION_ENABLED = true;
+config.AWS_LOG_CANDIDATES_LIMIT = 10;
+config.BUCKET_LOG_REPLICATOR_DELAY = 1 * 120 * 1000;
```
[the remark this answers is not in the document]
utkarsh-pro: I am not sure why this time was choses. I agree with the format as well. I think this is coming from Kfir's work on this as well. Both I and @MeridianExplorer did not know the rationale behind this number so we left it untouched I guess.

Having said that, the rationale could be that:
- Logs grow **really** quickly. Like **REALLY** quickly.
- We keep `HEAD`ing the objects in the bucket also adds the logs. I think if the delay is too much then we can take several hours to process just about 1000 logs (which can be generated really quickly). Say, we process 10 logs every 5 mins then we will take about 5 * 100 (i.e ~ 8 hours to process first 1000 logs).

src/api/replication_api.js
```diff
@@ -44,5 +44,30 @@ module.exports = {
                 system: 'admin'
             }
         },
+        delete_objects: {
```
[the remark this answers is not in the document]
utkarsh-pro: I am not sure about the first comment. I lack a bit of context there maybe @MeridianExplorer would know.

src/server/bg_services/replication_scanner.js
```diff
@@ -95,8 +95,9 @@ class ReplicationScanner {
             let move_res;
             const keys_to_copy = Object.keys(keys_sizes_map_to_copy);
             if (keys_to_copy.length) {
-                const copy_type = this._get_copy_type();
-                move_res = await this.move_objects(copy_type, src_bucket.name, dst_bucket.name, keys_to_copy);
+                const copy_type = replication_utils.get_copy_type();
+                move_res = await replication_utils.move_objects(this._scanner_sem, this.client, copy_type, src_bucket.name, dst_bucket.name,
+                    keys_to_copy);
```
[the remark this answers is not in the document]
utkarsh-pro: Yes, there seems to be a new line (accidental I believe) there. Will fix it.

src/server/bg_services/replication_log_parser.js
shirady: From the other examples in this PR, it seems `string`.
I also found [this](https://jsdoc.app/tags-param.html) and inside in the examples, it is `string`
Screenshot 2023-02-27 at 13 34 28

src/server/bg_services/replication_log_parser.js
```diff
@@ -0,0 +1,320 @@
...
+        pool_server.get_namespace_resource_extended_info(source_bucket.namespace.write_resource.resource);
+
+    const agent = s3_resource_connection_info.endpoint_type === 'AWS' ?
+        http_utils.get_default_agent(s3_resource_connection_info.endpoint) :
+        http_utils.get_unsecured_agent(s3_resource_connection_info.endpoint);
+
+    const s3 = new AWS.S3({
+        params: { Bucket: logs_bucket },
+        endpoint: s3_resource_connection_info.endpoint,
+        accessKeyId: s3_resource_connection_info.access_key.unwrap(),
```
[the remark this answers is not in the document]
utkarsh-pro: Yes, this had be wondering as well when I picked this up. I will document it somewhere.

src/server/bg_services/replication_log_parser.js
utkarsh-pro: Actually the idea was to not skip every such operation which is of type `DELETE` but the HTTP_STATUS is 204. For some reason, I have seen multiple such operations which fail with `400` errors or with `MALFORMEDHEADER` error code. So the idea is to process only those operations which resulted in a successful deletions and ignore the rest.

src/server/system_services/replication_store.js
shirady: If we don't use it or do not know when we will use it in the future, it is better to delete it.

src/server/bg_services/replication_log_parser.js
utkarsh-pro: Oh no, I meant that the line that you have pointed to uses `string`.

Also, both `String` and `string` are valid types to use (one being the class another primitive type) in JSDoc. But I agree, maybe we can just use `string` instead of mixing `String` and `string`.

config.js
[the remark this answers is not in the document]
romayalon: but we have a limit for the candidate's number also so we won't process 1000 log files, right? I wonder if running this bg worker every 2 minutes will make noobaa work harder, as you said we are heading every object every 2 minutes and it's added to the logs as well

src/server/bg_services/replication_log_parser.js
romayalon: I meant to deletion with success http codes like 202, 204, 200, https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE
not sure if we would ever get those but please check

config.js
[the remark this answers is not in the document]
utkarsh-pro: Oh no, I meant that let's say when we begin the processing the logs there are 1000 logs in the s3 bucket. Now we process 10 logs every 5 minute. So to churn through these 1k logs it will take us 5 * 100 minutes (~ 8 hours) and this churning will create more logs for us to process. 

For example say following sequence events happen:
1. User start replication via NooBaa while there are already 1k logs (not actually a lot - usually this amount of log is generated very quickly).
2. NooBaa starts processing the logs and this processing of 1k logs will take us about 8 hours.
3. While we were processing, user deleted `x` number of objects - this created more logs.

In such sequence of events, we will take at least 8 hours (no upper bound) to process the 2 deletes.

Having said that, frequent log churning will also be heavy on NooBaa but I am not sure what the right amount of delay would be.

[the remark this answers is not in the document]
romayalon: @dannyzaken thoughts about the correct delay?

src/server/bg_services/replication_log_parser.js
romayalon: interesting, so we won't support processing delete multiple objects event on this drop? probably we would like at least document it as a gap

src/server/bg_services/replication_server.js
```diff
@@ -19,6 +19,35 @@ async function move_objects_by_type(req) {
...
+    dbg.log1('replication_server delete_objects: params:', bucket_name, keys);
+    const delete_done_list = [];
+
+    const noobaa_con = cloud_utils.set_noobaa_s3_connection(system_store.data.systems[0]);
+    if (!noobaa_con) throw new Error('noobaa endpoint connection is not started yet...');
+    try {
+        const res = await noobaa_con.deleteObjects({
+            Bucket: bucket_name.unwrap(),
+            Delete: {
+                Objects: keys.map(key => ({ Key: key }))
```
romayalon, line 38: here we will have max 1000 keys?

utkarsh-pro: It is not being enforced in the code yet. @romayalon should I split the deletes into batches of 1k if there are more than 1k keys?

romayalon: yes, my next suggestion is to keep using the semaphore if we don't have this limit :)

utkarsh-pro: Got it. Thank you.

utkarsh-pro: @romayalon, done.

src/api/replication_api.js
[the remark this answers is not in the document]
dannyzaken: Forr now, this interface is sufficient. We can always add anything if necessary

config.js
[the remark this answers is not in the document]
dannyzaken: I don't think there was any deep discussion regarding this value. maybe we can change it to 5 like the scan-based and fine-tune it later

src/server/utils/replication_utils.js
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/replication_log_parser.js
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/replication_log_parser.js
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/log_replication_scanner.js
utkarsh-pro: Marking resolved - addressed.

src/server/system_services/bucket_server.js
```diff
@@ -1805,8 +1805,12 @@ async function get_bucket_replication(req) {
     const res = {
         rules: bucket_names_replication
     };
-    if (replication.aws_log_replication_info) {
-        res.log_replication_info.logs_location = replication.aws_log_replication_info.logs_location;
+    if (replication.log_replication_info?.aws_log_replication_info) {
```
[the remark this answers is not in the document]
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/replication_scanner.js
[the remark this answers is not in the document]
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -0,0 +1,206 @@
...
+                dbg.error('log_replication_scanner.head_objects: error:', err);
+                throw err;
+            }
+        }));
+
+        return src_dst_objects_info;
+    }
+}
+
+exports.LogReplicationScanner = LogReplicationScanner;
```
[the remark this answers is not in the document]
utkarsh-pro: Marking resolved - addressed.

src/server/system_services/bucket_server.js
```diff
@@ -1891,7 +1895,7 @@ function normalize_replication(req) {
 
     const validated_replication = {
         rules: validated_replication_rules,
-        aws_log_replication_info: log_replication_info
+        log_replication_info: {aws_log_replication_info: log_replication_info}
```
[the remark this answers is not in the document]
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/replication_server.js
```diff
@@ -19,6 +19,29 @@ async function move_objects_by_type(req) {
...
 }
 
+async function delete_objects(req) {
+    const { bucket_name, keys } = req.rpc_params;
+    dbg.log1('replication_server delete_objects: params:', bucket_name, keys);
+    const delete_done_list = [];
+
+    const noobaa_con = cloud_utils.set_noobaa_s3_connection(system_store.data.systems[0]);
+    if (!noobaa_con) throw new Error('noobaa endpoint connection is not started yet...');
+    await P.map_with_concurrency(100, keys, async key => {
```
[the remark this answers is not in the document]
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -0,0 +1,206 @@
...
+        // Iterate over the policies in parallel
+        await P.all(_.map(replications, async repl => {
+            _.map(repl.rules, async rule => {
+                // At the moment this scanner just supports deletions
+                // hence if sync_deletions is false we can skip the entire
+                // rule completely
+                //
+                // NOTE: Should be removed once we support other operations
+                // as well
+                if (!rule.sync_deletions) return;
```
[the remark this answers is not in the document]
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/replication_log_parser.js
```diff
@@ -0,0 +1,320 @@
...
+ * @param {*} source_bucket_id ID of the source bucket
+ * @param {String} rule_id ID of the replication rule
+ * @param {Record<any, any>} replicationconfig Replication configuration
+ * @param {Number} candidates_limit Candidates limit
+ * @param {(log_entry: Record<string, any>) => Boolean} ignore_fn - function that returns true if the log entry should be ignored
+ * @returns 
+ */
+async function get_log_candidates(source_bucket_id, rule_id, replicationconfig, candidates_limit, ignore_fn = () => false) {
+    if (replicationconfig && replicationconfig.log_replication_info && replicationconfig.log_replication_info.aws_log_replication_info) {
+        return get_aws_log_candidates(source_bucket_id, rule_id, replicationconfig, candidates_limit, ignore_fn);
```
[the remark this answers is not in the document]
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/replication_log_parser.js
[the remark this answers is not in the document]
utkarsh-pro: @MeridianExplorer - I think we should add this to the doc.

src/server/system_services/replication_store.js
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/replication_server.js
utkarsh-pro: Marking resolved - addressed.

dannyzaken (approved):

src/server/system_services/replication_store.js
utkarsh-pro: Marking resolved - addressed.

src/server/bg_services/replication_log_parser.js
utkarsh-pro: @romayalon, I performed `aws s3api delete-objects` on the bucket and finally received the logs for them. It turns out that the call results in `BATCH.DELETE.OBJECT` and a successful delete is indicated by `204`.

Attaching the relevant log here:
```
2e055fd7dc2c9eb01fe8d089e32b84c4b9e8a66b85e370793b4c7259ea477902 utk-repl [28/Feb/2023:14:15:42 +0000] 202.134.149.165 arn:aws:iam::255735097469:user/utkarsh H1SZ73V4C2X0DJXA BATCH.DELETE.OBJECT awsdata1 - 204 - - 10240 - - - - - BVQb2i2NLQHD1BcYkQZU58Ev1Elsz8+sQtOaYY9JU02SGdkuiQPHyXWcIuzUT/hC3YMS/LFNwRQ= SigV4 ECDHE-RSA-AES128-GCM-SHA256 AuthHeader utk-repl.s3.ap-south-1.amazonaws.com TLSv1.2 - -
2e055fd7dc2c9eb01fe8d089e32b84c4b9e8a66b85e370793b4c7259ea477902 utk-repl [28/Feb/2023:14:15:42 +0000] 202.134.149.165 arn:aws:iam::255735097469:user/utkarsh H1SZ73V4C2X0DJXA BATCH.DELETE.OBJECT awsdata2 - 204 - - 10240 - - - - - BVQb2i2NLQHD1BcYkQZU58Ev1Elsz8+sQtOaYY9JU02SGdkuiQPHyXWcIuzUT/hC3YMS/LFNwRQ= SigV4 ECDHE-RSA-AES128-GCM-SHA256 AuthHeader utk-repl.s3.ap-south-1.amazonaws.com TLSv1.2 - -
2e055fd7dc2c9eb01fe8d089e32b84c4b9e8a66b85e370793b4c7259ea477902 utk-repl [28/Feb/2023:14:15:42 +0000] 202.134.149.165 arn:aws:iam::255735097469:user/utkarsh H1SZ73V4C2X0DJXA REST.POST.MULTI_OBJECT_DELETE - "POST /?delete HTTP/1.1" 200 - 192 - 42 - "-" "aws-cli/2.10.2 Python/3.11.2 Darwin/22.3.0 source/arm64 prompt/off command/s3api.delete-objects" - BVQb2i2NLQHD1BcYkQZU58Ev1Elsz8+sQtOaYY9JU02SGdkuiQPHyXWcIuzUT/hC3YMS/LFNwRQ= SigV4 ECDHE-RSA-AES128-GCM-SHA256 AuthHeader utk-repl.s3.ap-south-1.amazonaws.com TLSv1.2 - -
```

Here I deleted 2 keys with delete-objects (the keys being `awsdata1` and `awsdata2`).

I think with this we have all of our bases covered.

Can I mark this as resolved now?

romayalon: great! if AWS docs say that it will return 204 on success, sure

PR #7125, opened 2022-12-08
[Backport to 5.11] Fixed map_common_prefixes queries

nimrod-becker:
* changed pl/sql function map_common_prefixes to fetch all required entries in one call.
* instead of querying for 1000 keys and then extracting the common prefixes from it (can be very few common prefixes if there are a lot of objects in that "dir"), it iterates over the objectmds untils reaching to the max keys or the end.
* To avoid unnecessary scanning of objects - if we find a common prefix, we set the marker to be the next possible value for a common prefix (by incrementing the last charecter by 1)
* These changes rely on the DB having collation of LC_COLLATE = 'C'. it will not work on DBs with utf8 collation

(cherry picked from commit 78f0ff813979371c90e7f33ac93fdbc6d928e80d)

### Explain the changes
1. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -835,29 +835,21 @@ class PostgresTable {
         let mr_q;
         sql_query.where = mongo_to_pg('data', encode_json(this.schema, options.query), { disableContainmentQuery: true });
         sql_query.order_by = options.sort && convert_sort(options.sort);
-        sql_query.limit = options.limit;
-        let query_string = `SELECT * FROM ${this.name} WHERE ${sql_query.where}`;
-        if (sql_query.order_by) {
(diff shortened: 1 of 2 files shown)


liranmauda:
@nimrod-becker We also need to bring in:
https://github.com/noobaa/noobaa-core/commit/e24e7e034238dbbefb3fb5d788c12146c9745e2c
https://github.com/noobaa/noobaa-core/commit/c86d6f621d03e1daee4b173c0271cc382e2a325b
@romayalon @dannyzaken ^^

nimrod-becker:
@liranmauda Its already in 5.11

liranmauda (approved):

PR #7133, opened 2022-12-15
NSFS | Versioning | Implement GET/HEAD object

### LINKED ISSUE #7128: NSFS | Versioning | Implement GET/HEAD object
This feature is part of NSFS versioning. 

GET object will return the latest or the requested version object.
Basically bucket can be version:

1. Disabled 

- No call flow will/behaviour will be changed

2. Enabled

- return the latest object if specific version is not requested. 
- fetch the stat of all objects and return the object which is requested for

4. Suspended
Same behaviour as in case of Version enabled

steps:
1. Funnel the application to namespaceFS.js based on bucketspace and namespace
2. if version is specified as part of the GET, then stat("version") under `.version`directory, 
 ```
   if present  
         return the file 
   else 
          return err
```
3. If no version specified then return the file under the specified path
4. Follow the same process for HEAD object, instead of data just return the metadata of the request file

@dannyzaken @romayalon Please correct if wrong

### PR DESCRIPTION
**Pull request by vh05**
### Explain the changes
This feature is part of NSFS versioning.
GET/HEAD object will return the latest or the requested version object.

Below is the behaviour of the patch.
NOTE: Lets call "file under bucket_path as regular file in this commit context"

1. with version disabled
   - no behavior change - should return the regular file

2. with version enabled/suspended
   - file created before the version is enabled - should return the regular file
   - no version id specified - should return the regular file
   - with version specified - if version id of regular file matches with version specified - should return the regular file
   - with version specified - if version id of regular file does not match with version specified - should return the versioned file
   - delete marked xattr placed - should return ENOENT

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### TBD
Optimize calls to `stat` in` _find_version_path()` 

`Signed-off-by: Vinayakswami Hariharmath <vharihar@redhat.com>

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -40,13 +40,14 @@ const XATTR_VERSION_ID = XATTR_USER_PREFIX + 'version_id';
 const XATTR_PREV_VERSION_ID = XATTR_USER_PREFIX + 'prev_version_id';
 //const XATTR_DELETE_MARKER = XATTR_USER_PREFIX + 'delete_marker';
 const HIDDEN_VERSIONS_PATH = '.versions';
+const XATTR_DELETE_MARKER = XATTR_USER_PREFIX + 'delete_marker';
 
 const versioning_status_enum = {
     VER_ENABLED: 'ENABLED',
     VER_SUSPENDED: 'SUSPENDED',
     VER_DISABLED: 'DISABLED'
 };
-
+const version_format = /^[a-z0-9]+$/;
 // describes the status of the copy that was done, default is fallback
 // LINKED = the file was linked on the server side
 // IS_SAME_INODE = source and target are the same inode, nothing to copy
@@ -507,7 +508,7 @@ class NamespaceFS {
                 if (r.common_prefix) {
                     res.common_prefixes.push(r.key);
                 } else {
-                    res.objects.push(this._get_object_info(bucket, r.key, r.stat));
+                    res.objects.push(this._get_object_info(bucket, r.key, r.stat, 'null'));
                 }
                 if (res.is_truncated) {
                     res.next_marker = r.key;
@@ -536,7 +537,8 @@ class NamespaceFS {
             await this._load_bucket(params, fs_context);
             const stat = await nb_native().fs.stat(fs_context, file_path);
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
-            return this._get_object_info(params.bucket, params.key, stat, params.version_id);
+            this._throw_if_delete_marker(stat);
+            return this._get_object_info(params.bucket, params.key, stat, params.version_id || 'null');
         } catch (err) {
             this.run_update_issues_report(object_sdk, err);
             throw this._translate_object_error_codes(err);
@@ -572,6 +574,8 @@ class NamespaceFS {
             file_path = await this._find_version_path(fs_context, params);
             await this._check_path_in_bucket_boundaries(fs_context, file_path);
             await this._fail_if_archived_or_sparse_file(fs_context, file_path);
+            const stat = await nb_native().fs.stat(fs_context, file_path);
+            this._throw_if_delete_marker(stat);
             file = await nb_native().fs.open(fs_context, file_path, undefined, get_umasked_mode(config.BASE_MODE_FILE));
 
             const start = Number(params.start) || 0;
@@ -1723,6 +1727,7 @@ class NamespaceFS {
         const cur_ver_path = this._get_file_path({ key });
         if (!version_id) return cur_ver_path;
 
+        this._throw_if_wrong_version_format(version_id);
         const versioned_path = this._get_version_path(key, version_id);
         const version_info = await this._get_version_info(fs_context, versioned_path);
         if (version_info) return versioned_path;
@@ -1731,6 +1736,28 @@ class NamespaceFS {
         const cur_ver_info = await this._get_version_info(fs_context, cur_ver_path);
         if (cur_ver_info && cur_ver_info.version_id_str === version_id) return cur_ver_path;
         throw new RpcError('NO_SUCH_OBJECT', 'version doesn\'t exist');
+    }
+
+    _throw_if_delete_marker(stat) {
+        if (this.versioning === versioning_status_enum.VER_ENABLED || this.versioning === versioning_status_enum.VER_SUSPENDED) {
+            const xattr_delete_marker = stat.xattr[XATTR_DELETE_MARKER];
+            if (xattr_delete_marker) {
+                throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
+            }
+        }
+    }
+
+    _throw_if_wrong_version_format(version_id) {
+        if (version_id === 'null') {
+            return;
+        }
+        const v_parts = version_id.split('-');
+        if (v_parts[0] !== 'mtime' || v_parts[2] !== 'ino') {
+            throw new RpcError('BAD_REQUEST', 'Bad Request');
+        }
+        if (!version_format.test(v_parts[1]) || !version_format.test(v_parts[3])) {
+            throw new RpcError('BAD_REQUEST', 'Bad Request');
+        }
     }
 }
 
--- a/src/test/unit_tests/test_namespace_fs.js
+++ b/src/test/unit_tests/test_namespace_fs.js
@@ -1,4 +1,5 @@
 /* Copyright (C) 2020 NooBaa */
+/*eslint max-lines-per-function: ["error", 700]*/
 'use strict';
 
 
@@ -17,7 +18,9 @@ const s3_utils = require('../../endpoint/s3/s3_utils');
 const config = require('../../../config');
 const fs = require('fs');
 const { AbortController } = require('node-abort-controller');
-
+const coretest = require('./coretest');
+const { rpc_client } = coretest;
+const nsfs_versioning = require('./test_bucketspace_versioning.js');
 const inspect = (x, max_arr = 5) => util.inspect(x, { colors: true, depth: null, maxArrayLength: max_arr });
 
 // TODO: In order to verify validity add content_md5_mtime as well
@@ -159,7 +162,7 @@ mocha.describe('namespace_fs', function() {
             }, dummy_object_sdk, out);
             const res = out.join().toString();
             assert.strict.equal(res.slice(13, 28), '(C) 2020 NooBaa');
-            assert.strict.equal(res.slice(37, 43), 'strict');
+            assert.strict.equal(res.slice(34, 40), 'eslint');
         });
 
         mocha.it('read range', async function() {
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: src/sdk/namespace_fs.js [hunk 1]
```diff
@@ -514,7 +517,36 @@ class NamespaceFS {
             await this._load_bucket(params, fs_context);
             const stat = await nb_native().fs.stat(fs_context, file_path);
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
-            return this._get_object_info(params.bucket, params.key, stat);
+
```
romayalon, line 520: Hi @v-harihar A great start! :)
this functionality of searching a version in .versions/ or fallback to the parent folder is needed for almost every API function.
In order to write and maintain this logic once, it should move out to a helper function outside of read_object_md.
I would create a function with similar functionality to what you did but I would leave the delete marker error handling to read_object_md/read_object_stream.
that way, instead of calling: `const file_path = this._get_file_path(params);` I would call a function that does this searching logic.
After that change, you will need to do the delete marker check logic once

v-harihar: sure .. I will rebase this patch once your patch is merged.

src/sdk/namespace_fs.js
```diff
@@ -545,6 +554,11 @@ class NamespaceFS {
         let buffer_pool_cleanup = null;
         const fs_context = this.prepare_fs_context(object_sdk);
         let file_path;
+
+        if (this.versioning === 'ENABLED' || this.versioning === 'SUSPENDED') {
```
romayalon, line 558: this is a bit hacky, I think it's preferred to have here the same behavior as in read_object_md:
get the version path and handle delete marker and version id etc

src/sdk/namespace_fs.js
```diff
@@ -509,12 +512,18 @@ class NamespaceFS {
...
         const fs_context = this.prepare_fs_context(object_sdk);
         try {
-            const file_path = this._get_file_path(params);
+            const file_path = await this._get_version_path(params, fs_context);
             await this._check_path_in_bucket_boundaries(fs_context, file_path);
             await this._load_bucket(params, fs_context);
             const stat = await nb_native().fs.stat(fs_context, file_path);
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
-            return this._get_object_info(params.bucket, params.key, stat);
+            if (this.versioning === 'ENABLED' || this.versioning === 'SUSPENDED') {
```
romayalon, line 520: please define the following consts and use them instead of comparing the versioning mode with a string:
const VER_ENABLED = 'ENABLED';
const VER_SUSPENDED = 'SUSPENDED';
const VER_DISABLED = 'DISABLED';

src/sdk/namespace_fs.js
```diff
@@ -509,12 +512,18 @@ class NamespaceFS {
...
+                const xattr_delete_marker = stat.xattr[XATTR_DELETE_MARKER];
+                if (xattr_delete_marker) {
```
romayalon, line 522: you can move this check to a  helper function named for example
_throw_if_delete_marker() since we will need it in read_object_stream

src/sdk/namespace_fs.js
```diff
@@ -1332,6 +1346,29 @@ class NamespaceFS {
         };
     }
 
+    async _get_version_path(params, fs_context) {
+        if (this.versioning === 'ENABLED' || this.versioning === 'SUSPENDED') {
+            const file_path = this._get_file_path(params);
+            const stat = await nb_native().fs.stat(fs_context, file_path);
+            if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
+            const xattr_version_id = stat.xattr[XATTR_VERSION_ID];
+            if (!xattr_version_id || !params.version_id || xattr_version_id === params.version_id) {
```
romayalon, line 1355: A few comments about that if - 
1. no need to stat if !params.version_id, you can move this check to the condition in the first line, and the versioning mode check is not needed because the user can ask for a version id even if the versioning is still disabled and we should fail since the version doesn't exist, but we shouldn't return the latest version. 
2. The first condition here !xattr_version_id is incorrect, for example, if a user asked for a specific version but it doesn't exist and there is only latest version of that object, the latest version will return but it's not what the user asked, we should fail with no such object in this case.
3. The last condition - we should fail with no such object here also if we can't find the version in the latest version path or under .versions/ so we need to add stat check for the version path as well and fail if we couldn't find this version

src/sdk/namespace_fs.js
```diff
@@ -1332,6 +1346,29 @@ class NamespaceFS {
...
+                // case 1: when the object was created but versioning was not yet enabled.
+                // Return the latest file path
+                // case 2: the object wanted is the latest version, return the latest file path
+                return file_path;
+            } else {
+                // If the object wanted is among the previous versions, return the
+                // object from $BUCKET_PATH/.versions/version_id
+                const v_path = path.normalize(path.join(this.bucket_path, VERSION_DIR, params.version_id));
```
romayalon, line 1363: the version path should be of the following form:
```
_get_version_path(key, version_id) {
        const key_version = path.basename(key) + (version_id ? '_' + version_id : '');
        return path.normalize(path.join(this.bucket_path, path.dirname(key), hidden_versions_path, key_version));
    }
```

romayalon (changes requested):

src/sdk/namespace_fs.js
v-harihar: Since we already have object md, I felt we can use the whatever information we got from md call which does save some IO operations. However I have replaced `get_file_path()` with `get_version_path()`, May be we can think of optimization here to avoid multiple IO operations

src/sdk/namespace_fs.js
v-harihar: sure

src/sdk/namespace_fs.js
v-harihar: I updated the patch to align with above comments (may not be complete, Please take look)

src/sdk/namespace_fs.js
v-harihar: Considered `key_version` to address the file inside `.versions`. Please correct me if wrong

src/sdk/namespace_fs.js
romayalon: How did it even pass test_namespace_fs.js tests? 
test_namespace_fs calls directly to the namespace runtime object and it's not going through the s3_put_object.js flow.
Optimization is always good and I guess we can optimize the 2 calls to get_version_path() to 1 but we need to take into consideration the scale and the fact that the file can be changed between the read_object_stream and read_object_md

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -229,6 +234,283 @@ mocha.describe('namespace_fs', function() {
...
+        });
+
+        mocha.after(async () => {
+            //fs_utils.folder_delete(tmp_fs_root);
+            for (let email of accounts) {
+                await rpc_client.account.delete_account({ email });
+            }
+        });
+
+        mocha.it('get object, versioning not enabled - should return latest object', async function() {
```
romayalon, line 347: could you add a test for get/head null version when it's not the latest? 
currently, it's not a different flow than what we have but maybe for future changes it's good to have it as well

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -764,3 +1046,14 @@ async function read_object_stream(ns, bucket, file_key, dummy_object_sdk) {
...
+
+async function create_object(object_path, data, version_id) {
+    const target_file = await nb_native().fs.open(DEFAULT_FS_CONFIG, object_path, 'w+');
+    await fs.promises.writeFile(object_path, data);
+    if (version_id !== 'null') {
+        const xattr_version_id = { 'user.version_id': `${version_id}` };
+        await target_file.replacexattr(DEFAULT_FS_CONFIG, xattr_version_id);
+    }
+
+    return target_file;
```
romayalon, line 1058: no newline at the end of the file here and in test_bucketspace_versioning as well

src/sdk/namespace_fs.js
```diff
@@ -509,12 +514,13 @@ class NamespaceFS {
...
         try {
-            const file_path = this._get_file_path(params);
+            const file_path = await this._get_version_path(params, fs_context);
             await this._check_path_in_bucket_boundaries(fs_context, file_path);
             await this._load_bucket(params, fs_context);
             const stat = await nb_native().fs.stat(fs_context, file_path);
             if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
-            return this._get_object_info(params.bucket, params.key, stat);
+            this._throw_if_delete_marker(stat);
+            return this._get_object_info(params.bucket, params.key, stat, params.version_id || 'null');
```
romayalon, line 523: why do we need this 'null'? 
Also, when bucket versioning is disabled we shouldn't return version_id

v-harihar: No it is the past behavior. `_get_object_info()` used to return `version_id` as `null`

src/sdk/namespace_fs.js
```diff
@@ -547,7 +553,7 @@ class NamespaceFS {
         let file_path;
         try {
             await this._load_bucket(params, fs_context);
-            file_path = this._get_file_path(params);
+            file_path = await this._get_version_path(params, fs_context);
             await this._check_path_in_bucket_boundaries(fs_context, file_path);
```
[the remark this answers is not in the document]
v-harihar: sure

src/sdk/namespace_fs.js
romayalon: you are right we used to always return '1'? so the sdk/cli removes it when getting an answer?

v-harihar: let me check this

src/sdk/namespace_fs.js
```diff
@@ -1332,6 +1341,37 @@ class NamespaceFS {
         };
     }
 
+    async _get_version_path(params, fs_context) {
+        const file_path = this._get_file_path(params);
+        if (params.version_id) {
+            const stat = await nb_native().fs.stat(fs_context, file_path);
+            if (isDirectory(stat)) throw Object.assign(new Error('NoSuchKey'), { code: 'ENOENT' });
```
romayalon, line 1348: I ran some examples on AWS and noticed that head object with version-id='1' throws BadRequest and not ENOENT (as the current implementation will throw), so we need here a check for an invalid version parameter.
It should be a helper function so it can be used in all the functions

romayalon:
@v-harihar Looks good, one last comment from my side

romayalon (approved):

liranmauda (changes requested):
@v-harihar Please squash the commits
We are not using merge from the UI, as we do not want to get the `Merge branch 'master' into <branch>` commit

liranmauda (approved):

PR #7137, opened 2022-12-18
NSFS | add list buckets op metric

### LINKED ISSUE #7136: metrics not found for s3 `list buckets` operation
### Environment info
- NooBaa Version: **5.12.0**
- NooBaa Image: **noobaa/noobaa-core:master-20220913**
- NooBaa Operator Image: **noobaa/noobaa-operator:5.12.0**
- Platform: **Kubernetes 1.25.4 | OpenShift 4.12.0-rc.4**

### Actual behavior
1. metrics are not reflecting in the I/O Operations for `aws s3 ls s3://` operation.
or it may be included under FS Ops metrics, but FS metrics are not getting for all operations.

### Expected behavior
1. There should be some metrics for bucket list operation.

### Steps to reproduce
1. Set up s3 account to with access key and secret key
2. perform s3 operations `aws s3 ls s3://` to get list of buckets
3. use noobaa route to check metrics on `/metrics/nsfs_stats`.

*List Buckets*
**aws s3 ls s3://**

image

### Please find must-gather logs:

https://ibm.ent.box.com/file/1092845779983?s=g16bxab8ujxixp9yy7q2cogh6apuiszl

liranmauda:
### Explain the changes
- Add  list bucket op metric

1. Fixes: https://github.com/noobaa/noobaa-core/issues/7136

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -845,6 +845,9 @@ module.exports = {
                 create_bucket: {
                     $ref: 'common_api#/definitions/op_stats_val'
                 },
+                list_buckets: {
+                    $ref: 'common_api#/definitions/op_stats_val'
+                },
                 delete_bucket: {
                     $ref: 'common_api#/definitions/op_stats_val'
                 },
@@ -860,6 +863,9 @@ module.exports = {
                 head_object: {
                     $ref: 'common_api#/definitions/op_stats_val'
                 },
+                read_object: {
+                    $ref: 'common_api#/definitions/op_stats_val'
+                },
                 initiate_multipart: {
                     $ref: 'common_api#/definitions/op_stats_val'
                 },
@@ -867,9 +873,6 @@ module.exports = {
                     $ref: 'common_api#/definitions/op_stats_val'
                 },
                 complete_object_upload: {
-                    $ref: 'common_api#/definitions/op_stats_val'
-                },
-                read_object: {
                     $ref: 'common_api#/definitions/op_stats_val'
                 },
             }
--- a/src/sdk/object_sdk.js
+++ b/src/sdk/object_sdk.js
@@ -801,8 +801,22 @@ class ObjectSDK {
     ////////////
 
     async list_buckets() {
-        const bs = this._get_bucketspace();
-        return bs.list_buckets(this);
+        const start_time = Date.now();
+        let error = 0;
+        try {
+            const bs = this._get_bucketspace();
+            const reply = bs.list_buckets(this);
+            return reply;
+        } catch (e) {
+            error = 1;
+            throw e;
+        } finally {
+            stats_collector.instance(this.internal_rpc_client).update_ops_counters({
+                time: Date.now() - start_time,
+                op_name: `list_buckets`,
+                error,
+            });
+        }
     }
 
     async read_bucket(params) {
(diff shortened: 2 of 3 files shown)


src/sdk/object_sdk.js
```diff
@@ -801,8 +801,22 @@ class ObjectSDK {
...
+            const bs = this._get_bucketspace();
+            const reply = bs.list_buckets(this);
+            return reply;
+        } catch (e) {
+            error = 1;
+            throw e;
+        } finally {
+            stats_collector.instance(this.internal_rpc_client).update_ops_counters({
+                time: Date.now() - start_time,
+                op_name: `list_bucket`,
```
romayalon, line 816: list bucket is not the same as list buckets,
change it to list buckets as the name of this function

src/sdk/object_sdk.js
```diff
@@ -801,8 +801,22 @@ class ObjectSDK {
     ////////////
 
     async list_buckets() {
-        const bs = this._get_bucketspace();
-        return bs.list_buckets(this);
+        const start_time = Date.now();
+        let error = 0;
```
romayalon, line 805: As discussed in the last review,
these lines should be a wrapper of each function that requires this metrics logic, instead of having this logic 1 time, it needs to be maintained in ~10 different functions.

src/api/common_api.js
```diff
@@ -845,6 +845,9 @@ module.exports = {
                 create_bucket: {
                     $ref: 'common_api#/definitions/op_stats_val'
                 },
+                list_bucket: {
+                    $ref: 'common_api#/definitions/op_stats_val'
```
romayalon, line 849: as we discussed,
this schema should be refactored to something like op_name: 'string'
and value, currently it adds a lot of redundant code and a redundant maintenance

src/sdk/object_sdk.js
romayalon: and every function has 16 rows instead of 2

liranmauda: This should be a handler, but as for now, We will keep it this way and make a PR for creating a handler later.

src/api/common_api.js
liranmauda: Currently, there is no plan to change the schema, We made it this way to be explicit. 
We wanted the schema to state each op name. 
This is actually not a redundant code, but I do agree that it needs more attention when adding items to the metrics. 

We can have a discussion later on the road to determine if this should be explicit or not. 
@guymguym @dannyzaken @nimrod-becker WDYT?

src/sdk/object_sdk.js
liranmauda: @romayalon I talked to @nimrod-becker 
We will do that for 4.13, and after the point that we don't plan to add more ops for 4.12.z

romayalon (approved):

nimrod-becker:
Following a discussion we had on the different comments on this PR we agreed to

1. Since today we send everything we collect, there is no sense in whitelisting filtering. Even if such would be needed it makes more sense to do so in the consumer (client) of this API. So in this PR, lets remove the whitelist filtering code altogether.

2. In a follow-up PR we agreed to fix the following point
  2.1 Change the schema from names structures to an array with op name as a string key
  2.2 Add a unit test to verify what ops are being exposed via this API in order to maintain compatibility with the client (fail if 
         anything is not expected is returned, and populate data to be returned via RPC calls)
  2.3 Add. handler for op collection instead of dup code

PR #7141, opened 2022-12-26
Nooba bucket Logging configuration

aspandey:
Changes :
1. We are trying to come up with the s3 based logging in nooba so that we can use the log based replication.
This patch is a work in progress patch towards this. which set the configuration of logging on nooba bucket.


src/server/system_services/schemas/bucket_schema.js
```diff
@@ -269,6 +269,22 @@ module.exports = {
...
                 writes: { type: 'integer' },
             }
-        }
+        },
+        // Do we need enable/disable option? if yes, how to do that?
+        logging: {
+            type: 'object',
+            required: ['source_bucket', 'log_bucket', 'log_prefix'],
+            properties: {
+                source_bucket: {
```
dannyzaken, line 278: I think that `source_bucket` is redundant. This is a schema for an individual bucket, which is the source bucket

src/server/system_services/schemas/bucket_schema.js
```diff
@@ -269,6 +269,22 @@ module.exports = {
                 reads: { type: 'integer' },
                 writes: { type: 'integer' },
             }
-        }
+        },
+        // Do we need enable/disable option? if yes, how to do that?
```
dannyzaken, line 273: according to [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutBucketLogging.html) you disable logging by sending an empty logging configuration, so I think it's equal to just deleting an existing one. so in my opinion there is no need to have enable\disable in this case.

in general when we do need it we can just add a boolean flag `enabled` or something similar

dannyzaken:
@aspandey notice that you have lint errors: 
https://github.com/noobaa/noobaa-core/actions/runs/3779960506/jobs/6425638866#step:3:2259

you can run `eslint` locally:
* from `noobaa-core` dir first run `npm install` to install all packages including eslint
* run `npm run lint` to see any lint errors

src/api/bucket_api.js
```diff
@@ -781,6 +781,74 @@ module.exports = {
...
+                        $ref: 'common_api#/definitions/bucket_name'
+                    },
+                    log_bucket: {
+                        $ref: 'common_api#/definitions/bucket_name'
+                    },
+                    log_prefix: {
+                        type: 'string',
+                    },
+                },
+                // auth: {
```
dannyzaken, line 801: why is the auth commented out? this property specifies which roles are allowed to perform the operation, by removing it the operation can be performed unauthenticated

src/server/system_services/schemas/bucket_schema.js
dannyzaken: you can remove the comment

dannyzaken (commented):
@aspandey other than the few small comments given, it looks good. 
you also need to sign the commit to fix the failing DCO

aspandey:
Checking it on my local machine and will resend it. Somehow it was not giving any error..

src/server/system_services/schemas/bucket_schema.js
dannyzaken: @aspandey, remove the comment before merging

dannyzaken (approved):

PR #7143, opened 2022-12-27
S3 Select Support

alphaprinz:
### Explain the changes
1. Incorporate s3select (https://github.com/ceph/s3select) to implement a "select object content"-like feature.
Feature is ideally as close as possible to s3, but it is excpeted that s3select will somewhat diverge.
If and when s3select will introduce new features which don't change its api, we can take them by changing our commit in the submodule (if api changes we also need to adapt native code accordingly).

1. None

### Testing Instructions:
1. Consider enabling ceph s3 tests related to "select object content"

- [ ] Doc added/updated
/docs/desgin/s3select.md
/docs/dev_guide/S3Select_Build.md
- [ ] Tests added
TODO

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "s3select"]
+	path = submodules/s3select
+	url = https://github.com/ceph/s3select
--- a/docs/design/s3select.md
+++ b/docs/design/s3select.md
@@ -1,0 +1,123 @@
+
+# S3 Select
+
+
+## Goal
+
+Provide a feature to run S3 Select queries as close as possible to AWS feature.
+
+
+## Scenarios
+
+A client sends an S3 Select request according to AWS specification (see [SelectObjectContent - Amazon Simple Storage Service](https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html)).
+
+The endpoint will retrieve the relevant file (AKA key), process the query, and reply according to specification (see [Appendix: SelectObjectContent Response - Amazon Simple Storage Service](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html)).
+
+
+## Technical Details
+
+
+### Query Processing Implementation
+
+Use external implementation, as implementing in-house is not a valid option.
+
+
+
+* S3Select ([https://github.com/ceph/s3select](https://github.com/ceph/s3select)) implements running SQL on character stream in C++. This requires napi, git and build changes. This is the chosen implementation.
+* Partiql ([https://partiql.org/](https://partiql.org/)) is available only in Kotlin implementation. This requires starting a JVM in the endpoint to process SQL, which is cumbersome.
+    * [https://github.com/partiql/partiql-lang-rust](https://github.com/partiql/partiql-lang-rust) can be considered in the future, but for now it doesn’t have a full implementation yet.
+
+
+### New S3 OP
+
+A new s3 op, post_object_select will be added.
+
+There are two variations of files to consider:
+
+
+
+1. csv, json - These are “stream-friendly”, they are read once from start to finish. So we can get their stream like standard get_object s3 op, pipe the stream to s3select to process, and pipe the result to http res object.
+2. Parquet
+    1. Processing a parquet file requires random seeking. In other words, at any time during the processing of the SQL, s3select can require to read any address of the file. So the stream-piping is not suitable.
+    2. A general, currently undesigned, solution is to channel the request upwards, back to a callback in post_object_select, which will ask object_sdk for the required address and send it back to s3select.
+    3. In the simple case of FS namespace buckets, if the required file resides in the endpoint host, the file path (rather than stream) can be sent to s3select for processing. This allows processing Parquet files stored in namespace buckets.
+    4. Parquet requires Arrow, a library for processing Parquet files.
+    5. Initial feature will not support Parquet.
+
+
+### HTTP Response
+
+Response should conform to specification (see [Appendix: SelectObjectContent Response - Amazon Simple Storage Service](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html)).
+
+We can adapt S3Select example code that already has implementation to add crc and headers for each chunk.
+
+
+### Napi
+
+New native code that will implement a transform stream (similar to coder/chunker). This stream will have these functions:
+
+
+
+* write(byte array) - that will reply with rows matching the query.
+* flush() - that will reply with SQL-aggregate (eg count, sum) result.
+
+Preferably, stream will not encode/decode bytes into/from strings and will not copy buffers.
+
+
+### Git and Build
+
+
+
+1. S3Select source
+
+S3Select is a git repository. It has 3 dependencies:
+
+
+
+* Fast Cpp Csv Parser ([https://github.com/ben-strasser/fast-cpp-csv-parser](https://github.com/ben-strasser/fast-cpp-csv-parser))
+* RapidJson parser ([https://github.com/Tencent/rapidjson](https://github.com/Tencent/rapidjson))
+* Boost
+
+The parser git repositories which are submodules of s3select repository.
+
+Boost is assumed by s3select make to be installed on the builder machine.
+
+
+
+2. noobaa-core repository
+
+In noobaa-core repository, a new submodules directory and .gitmodules file are added.
+
+It will have the s3select and boost repositories as submodules, and the two parsers recursively, inside the s3select submodule.
+
+
+
+3. Docker build
+
+In docker build, these repositories are fetched and node-gyp uses the sources for compilation.
+
+Only necessary boost submodules are fetched.
+
+All repositories are checkout to a specific commit/tag so updates won’t affect us directly.
+
+When updates are needed, the checkout command needs to be updated to a new commit/tag.
+
+
+
+4. Native build
+
+Running native build should still be possible.
+
+Git submodules are needed to be fetched manually (once).
+
+
+### Tests
+
+
+
+* S3 ceph tests
+* ..
+
+
+### Upstream Docs
+
--- a/src/deploy/NVA_build/Base.Dockerfile
+++ b/src/deploy/NVA_build/Base.Dockerfile
@@ -19,7 +19,14 @@ RUN npm install --production && \
 ##############################################################
 COPY ./binding.gyp .
 COPY ./src/native ./src/native/
-RUN npm run build:native
+COPY ./src/deploy/NVA_build/clone_submodule.sh ./src/deploy/NVA_build/
+COPY ./src/deploy/NVA_build/clone_s3select_submodules.sh ./src/deploy/NVA_build/
+ARG BUILD_S3SELECT=1
+#Clone S3Select and its two submodules, but only if BUILD_S3SELECT=1.
+RUN ./src/deploy/NVA_build/clone_s3select_submodules.sh
+#Pass BUILD_S3SELECT down to GYP native build.
+#S3Select will be built only if this parameter is equal to "1".
+RUN GYP_DEFINES=BUILD_S3SELECT=$BUILD_S3SELECT npm run build:native
 
 ##############################################################
 # Layers:
--- a/src/deploy/NVA_build/builder.Dockerfile
+++ b/src/deploy/NVA_build/builder.Dockerfile
@@ -11,7 +11,7 @@ LABEL maintainer="Liran Mauda (lmauda@redhat.com)"
 #     dnf clean all
 RUN dnf update -y -q --nobest && \
     dnf clean all
-RUN dnf install -y -q wget unzip which vim python2 python3 && \
+RUN dnf install -y -q wget unzip which vim python2 python3 boost-devel && \
     dnf group install -y -q "Development Tools" && \
     dnf clean all
 RUN alternatives --set python /usr/bin/python3
--- a/src/endpoint/s3/ops/s3_post_object_select.js
+++ b/src/endpoint/s3/ops/s3_post_object_select.js
@@ -1,0 +1,118 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const dbg = require('../../../util/debug_module')(__filename);
+const S3Error = require('../s3_errors').S3Error;
+const s3_utils = require('../s3_utils');
+const { S3SelectStream } = require('../../../util/s3select');
+const nb_native = require('../../../util/nb_native');
+const stream_utils = require('../../../util/stream_utils');
+
+/**
+ * https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html
+ */
+async function post_object_select(req, res) {
+
+    if (!nb_native().S3Select) {
+        throw new S3Error(S3Error.S3SelectNotCompiled);
+    }
+
+    req.object_sdk.setup_abort_controller(req, res);
+    const agent_header = req.headers['user-agent'];
+    const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
+    const encryption = s3_utils.parse_encryption(req);
+    const http_req_select_params = req.body.SelectObjectContentRequest;
+
+    const md_params = {
+        bucket: req.params.bucket,
+        key: req.params.key,
+        version_id: req.query.versionId,
+        encryption,
+    };
+    const object_md = await req.object_sdk.read_object_md(md_params);
+
+    const params = {
+        object_md,
+        obj_id: object_md.obj_id,
+        bucket: req.params.bucket,
+        key: req.params.key,
+        content_type: object_md.content_type,
+        noobaa_trigger_agent,
+        encryption,
+    };
+
+    //handle ScanRange
+    if (Array.isArray(http_req_select_params.ScanRange)) {
+        const scan_range = http_req_select_params.ScanRange[0];
+        if (scan_range.Start) {
+            params.start = Number(scan_range.Start);
+        }
+        if (scan_range.End) {
+            if (scan_range.Start) {
+                params.end = Number(scan_range.End);
+            } else {
+                //if only End is specified, start from {End} bytes from the end.
+                params.start = object_md.size - (Number(scan_range.End));
+            }
+        }
+    }
+
+    //prepare s3select stream
+    const input_serialization = http_req_select_params.InputSerialization[0];
+    let input_format = null;
+    if (input_serialization.CSV) {
+        input_format = 'CSV';
+    } else if (input_serialization.JSON) {
+        input_format = 'JSON';
+    } else {
+        throw new S3Error(S3Error.MissingInputSerialization);
+    }
+
+    //currently s3select can only output in the same format as input format
+    if (Array.isArray(http_req_select_params.OutputSerialization)) {
+        const output_serialization = http_req_select_params.OutputSerialization[0];
+        if ((output_serialization.CSV && input_format !== 'CSV') ||
+            (output_serialization.JSON && input_format !== 'JSON')) {
+                throw new S3Error(S3Error.OutputInputFormatMismatch);
+            }
+    }
+
+    const select_args = {
+        query: http_req_select_params.Expression[0],
+        input_format: input_format,
+        input_serialization_format: http_req_select_params.InputSerialization[0][input_format][0],
+        records_header_buf: S3SelectStream.records_message_headers
+    };
+    const s3select = new S3SelectStream(select_args);
+    dbg.log3("select_args = ", select_args);
+
+    //pipe s3select result into http result
+    stream_utils.pipeline([s3select, res], true /*res is a write stream, no need for resume*/);
+
+    //send s3select pipe to read_object_stream.
+    //in some cases (currently nsfs) it will pipe object stream into our pipe (s3select)
+    const read_stream = await req.object_sdk.read_object_stream(params, s3select);
+    if (read_stream) {
+        // if read_stream supports closing, then we handle abort cases such as http disconnection
+        // by calling the close method to stop it from buffering more data which will go to waste.
+        if (read_stream.close) {
+            req.object_sdk.add_abort_handler(() => read_stream.close());
+        }
+        read_stream.on('error', err => {
+            dbg.error('read stream error:', err, req.path);
+            res.destroy(err);
+        });
+        //in other cases, we need to pipe the read stream ourselves
+        stream_utils.pipeline([read_stream, s3select], true /*no need to resume s3select*/);
+    }
+}
+
+module.exports = {
+    handler: post_object_select,
+    body: {
+        type: 'xml',
+    },
+    reply: {
+        type: 'raw',
+    },
+};
--- a/src/endpoint/s3/s3_errors.js
+++ b/src/endpoint/s3/s3_errors.js
@@ -512,6 +512,26 @@ S3Error.InvalidEncodingType = Object.freeze({
     http_code: 400,
 });
 
+////////////////////////////////////////////////////////////////
+// S3 Select                                                  //
+////////////////////////////////////////////////////////////////
+
+S3Error.S3SelectNotCompiled = Object.freeze({
+    code: 'S3SelectNotCompiled',
+    message: 'This server was not compiled with S3 Select support. Recompile with BUILD_S3SELECT=1.',
+    http_code: 501,
+});
+S3Error.MissingInputSerialization = Object.freeze({
+    code: 'MissingRequiredParameter',
+    message: 'InputSerialization is required. Please check the service documentation and try again.',
+    http_code: 400,
+});
+S3Error.OutputInputFormatMismatch = Object.freeze({
+    code: 'OutputInputFormatMismatch',
+    message: 'OutputSerialization format must match InputSerializatoin format.',
+    http_code: 501,
+});
+
 S3Error.RPC_ERRORS_TO_S3 = Object.freeze({
     UNAUTHORIZED: S3Error.AccessDenied,
     BAD_REQUEST: S3Error.BadRequest,
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -52,7 +52,8 @@ const OBJECT_SUB_RESOURCES = Object.freeze({
     'uploads': 'uploads',
     'uploadId': 'uploadId',
     'legal-hold': 'legal_hold',
-    'retention': 'retention'
+    'retention': 'retention',
+    'select': 'select',
 });
 
 const S3_OPS = load_ops();
@@ -581,6 +582,7 @@ function load_ops() {
         post_bucket_delete: require('./ops/s3_post_bucket_delete'),
         post_object_uploadId: require('./ops/s3_post_object_uploadId'),
         post_object_uploads: require('./ops/s3_post_object_uploads'),
+        post_object_select: require('./ops/s3_post_object_select'),
         put_bucket: require('./ops/s3_put_bucket'),
         put_bucket_accelerate: require('./ops/s3_put_bucket_accelerate'),
         put_bucket_acl: require('./ops/s3_put_bucket_acl'),
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -80,6 +80,7 @@ const OP_NAME_TO_ACTION = Object.freeze({
     post_object: { regular: "s3:putobject" },
     post_object_uploadId: { regular: "s3:putobject" },
     post_object_uploads: { regular: "s3:putobject" },
+    post_object_select: { regular: "s3:selectobjectcontent" },
 
     put_bucket_accelerate: { regular: "s3:putaccelerateconfiguration" },
     put_bucket_acl: { regular: "s3:putbucketacl" },
--- a/src/native/nb_native.gyp
+++ b/src/native/nb_native.gyp
@@ -16,15 +16,24 @@
             }],
             [ 'node_arch=="arm64"', {
                 'cflags': ['-DUSE_NEON']
-            }],
+            }]
         ],
     },
 
     'targets': [{
+	'variables': {
+            'BUILD_S3SELECT%':0
+        },
         'target_name': 'nb_native',
         'include_dirs': [
             '<@(napi_include_dirs)',
         ],
+	'conditions': [
+	    ['BUILD_S3SELECT==1', {
+		'dependencies': ['s3select/s3select.gyp:s3select'],
+		'cflags': ['-DBUILD_S3SELECT=1']
+	    }]
+	],
         'dependencies': [
             '<@(napi_dependencies)',
             'third_party/cm256.gyp:cm256',
@@ -32,6 +41,7 @@
             'third_party/isa-l.gyp:isa-l-ec',
             'third_party/isa-l.gyp:isa-l-md5',
             'third_party/isa-l.gyp:isa-l-sha1',
+            'third_party/isa-l.gyp:isa-l-crc'
         ],
         'sources': [
             # module
--- a/src/native/s3select/s3select.gyp
+++ b/src/native/s3select/s3select.gyp
@@ -1,0 +1,19 @@
+{
+    'includes': ['../third_party/common_third_party.gypi'],
+    'targets': [{
+        'target_name': 's3select',
+        'type': 'static_library',
+        'cflags_cc!': ['-fno-rtti'],
+        'include_dirs': [
+            '<@(napi_include_dirs)',
+            '../../../submodules/s3select/include',
+            '../../../submodules/s3select/rapidjson/include/'
+        ],
+        'sources': [
+            's3select_napi.cpp'
+        ],
+	'link_settings': {
+		'libraries': ['/lib64/libboost_thread.so.1.66.0']
+    }
+    }]
+}
--- a/src/native/s3select/s3select_napi.cpp
+++ b/src/native/s3select/s3select_napi.cpp
@@ -1,0 +1,279 @@
+/* Copyright (C) 2016 NooBaa */
+
+#include "../../../src/native/third_party/isa-l/include/crc.h"
+#include "../../../submodules/s3select/include/s3select.h"
+#include "../util/common.h"
+#include "../util/napi.h"
+#include <arpa/inet.h>
+
+namespace noobaa
+{
+
+const char *CSV_FORMAT = "CSV";
+const char *JSON_FORMAT = "JSON";
+const char *PARQUET_FORMAT = "PARQUET";
+
+class S3SelectNapi;
+
+class SelectWorker : public Napi::AsyncWorker
+{
+public:
+    Napi::ObjectReference _args_ref;
+    Napi::Promise::Deferred _deferred;
+    Napi::ObjectWrap<S3SelectNapi>& _wrap;
+    s3selectEngine::csv_object *_csv_object;
+    s3selectEngine::json_object *_json_object;
+    std::string _select;
+    const char* _buffer;
+    const unsigned int _buffer_len;
+    const bool _is_flush;
+    const std::string _input_format;
+    const uint8_t* _headers_bytes;
+    const uint32_t _headers_len;
+
+    SelectWorker(const Napi::CallbackInfo& info,
+        Napi::ObjectWrap<S3SelectNapi>& wrap,
+        s3selectEngine::csv_object* csv_object,
+        s3selectEngine::json_object* json_object,
+        const char* buffer,
+        const unsigned int buffer_len,
+        const bool is_flush,
+        const std::string input_format,
+        const uint8_t* headers_bytes,
+        const uint32_t headers_len)
+        : AsyncWorker(info.Env())
+        , _args_ref(Napi::Persistent(Napi::Object::New(info.Env())))
+        , _wrap(wrap)
+        , _csv_object(csv_object)
+        , _json_object(json_object)
+        , _buffer(buffer)
+        , _buffer_len(buffer_len)
+        , _is_flush(is_flush)
+        , _input_format(input_format)
+        , _deferred(info.Env())
+        , _headers_bytes(headers_bytes)
+        , _headers_len(headers_len)
+    {
+        //take a ref on args to make sure they are not GCed until worker is done
+        uint32_t i;
+        for (i = 0; i < info.Length(); ++i) {
+            _args_ref.Set(i, info[i]);
+        }
+        //prevent GC of S3SelectNapi object by JS until worker is done
+        _wrap.Ref();
+    }
+
+    ~SelectWorker()
+    {
+        //Worker is done, unref the ref we took in ctor
+        _wrap.Unref();
+    }
+
+    void Execute() override
+    {
+        int rc;
+        // TODO - if we get total size on beginning, i think we can remove the is_flush if
+        if (CSV_FORMAT == _input_format) {
+            if (_is_flush) {
+                rc = _csv_object->run_s3select_on_stream(_select, nullptr, 0, 0);
+            } else {
+                rc = _csv_object->run_s3select_on_stream(_select, _buffer, _buffer_len, SIZE_MAX);
+            }
+        } else {
+            if (_is_flush) {
+                rc = _json_object->run_s3select_on_stream(_select, nullptr, 0, 0);
+            } else {
+                rc = _json_object->run_s3select_on_stream(_select, _buffer, _buffer_len, SIZE_MAX);
+            }
+        }
+        if (rc < 0) {
+            if (CSV_FORMAT == _input_format) {
+                SetError(_csv_object->get_error_description());
+            } else {
+                // SetError(json_object.get_error_description()); //TODO - they added a getter, use after submodule is updated
+                SetError("failed to select from json");
+            }
+        }
+
+        /*std::cout << "select res = " << this->select << std::endl;
+        std::cout.flush();*/
+    }
+
+    void OnOK() override
+    {
+        Napi::Env env = Env();
+        // in case of empty select result (ie, no rows in current buffer matched sql condition), return null
+        if (_select.empty()) {
+            _deferred.Resolve(env.Null());
+            return;
+        }
+
+        Napi::Object result = Napi::Object::New(env);
+        Napi::Value select_buf = Napi::Buffer<char>::/*New*/ Copy(env, _select.data(), _select.length());
+        result.Set("select", select_buf);
+
+        uint32_t prelude[3];
+        int32_t prelude_crc, message_crc, prelude_crc_BE;
+        prelude[0] = htonl(_select.length() + _headers_len + 16);
+        prelude[1] = htonl(_headers_len);
+        prelude_crc = crc32_gzip_refl_base(0, (uint8_t*)prelude, 8);
+        result.Set("prelude_crc", prelude_crc);
+        prelude_crc_BE = htonl(prelude_crc);
+        /*std::cout << "sz = " << sizeof(headers_bytes) << ", native prelude_crc = " << prelude_crc << ", prelude = " << std::hex << std::setfill('0') << std::setw(2) << prelude[0] << prelude[1];
+        std::cout << std::endl;
+        std::cout.flush();*/
+        message_crc = crc32_gzip_refl_base(prelude_crc, (uint8_t*)&prelude_crc_BE, 4);
+        message_crc = crc32_gzip_refl_base(message_crc, const_cast<uint8_t*>(_headers_bytes), _headers_len);
+        message_crc = crc32_gzip_refl_base(message_crc, (uint8_t*)_select.data(), _select.length());
+        result.Set("message_crc", message_crc);
+
+        _deferred.Resolve(result);
+    }
+
+    void OnError(Napi::Error const& error) override
+    {
+        Napi::Env env = Env();
+        auto obj = error.Value();
+        _deferred.Reject(obj);
+    }
+};
+
+class S3SelectNapi : public Napi::ObjectWrap<S3SelectNapi>
+{
+
+public:
+    static Napi::Object Init(Napi::Env env, Napi::Object exports);
+    S3SelectNapi(const Napi::CallbackInfo& info);
+    Napi::Value Write(const Napi::CallbackInfo& info);
+    Napi::Value Flush(const Napi::CallbackInfo& info);
+    ~S3SelectNapi();
+
+private:
+    static Napi::FunctionReference constructor;
+    Napi::ObjectReference _args_ref;
+    std::string input_format;
+    s3selectEngine::s3select s3select;
+    s3selectEngine::csv_object* csv_object = nullptr;
+    s3selectEngine::json_object* json_object = nullptr;
+    const uint8_t* headers_buf;
+    uint32_t headers_len;
+};
+
+Napi::Value
+S3SelectNapi::Write(const Napi::CallbackInfo& info)
+{
+    Napi::Buffer<char> buffer = info[0].As<Napi::Buffer<char>>();
+    SelectWorker* worker = new SelectWorker(
+        info,
+        *this,
+        csv_object,
+        json_object,
+        buffer.Data(),
+        buffer.Length(),
+        false,
+        input_format,
+        headers_buf,
+        headers_len);
+    worker->Queue();
+    return worker->_deferred.Promise();
+}
+
+Napi::Value
+S3SelectNapi::Flush(const Napi::CallbackInfo& info)
+{
+    SelectWorker* worker = new SelectWorker(
+        info,
+        *this,
+        csv_object,
+        json_object,
+        nullptr, /*No buffer for flush*/
+        0,
+        true,
+        input_format,
+        headers_buf,
+        headers_len);
+    worker->Queue();
+    return worker->_deferred.Promise();
+}
+
+Napi::FunctionReference S3SelectNapi::constructor;
+
+Napi::Object
+S3SelectNapi::Init(Napi::Env env, Napi::Object exports)
+{
+    Napi::HandleScope scope(env);
+
+    Napi::Function func = DefineClass(
+        env,
+        "S3SelectNapi",
+        {
+            InstanceMethod("write", &S3SelectNapi::Write),
+            InstanceMethod("flush", &S3SelectNapi::Flush)
+        }
+    ); //end of DefineClass
+
+    constructor = Napi::Persistent(func);
+    constructor.SuppressDestruct();
+
+    exports.Set("S3Select", func);
+    return exports;
+}
+
+void
+s3select_napi(Napi::Env env, Napi::Object exports)
+{
+    S3SelectNapi::Init(env, exports);
+}
+
+std::string
+GetStringWithDefault(Napi::Object obj, std::string name, std::string dfault)
+{
+    if (obj.Has(name)) {
+        return obj.Get(name).ToString().Utf8Value();
+    }
+    return dfault;
+}
+
+S3SelectNapi::S3SelectNapi(const Napi::CallbackInfo& info)
+    : Napi::ObjectWrap<S3SelectNapi>(info)
+    , _args_ref(Napi::Persistent(Napi::Object::New(info.Env())))
+{
+    Napi::Object context = info[0].As<Napi::Object>();
+    Napi::Object input_serialization_format = context.Get("input_serialization_format").As<Napi::Object>();
+    Napi::Buffer headers_buf_obj = context.Get("records_header_buf").As<Napi::Buffer<uint8_t>>();
+    _args_ref.Set("headers_buf", headers_buf_obj);
+    headers_buf = headers_buf_obj.Data();
+    headers_len = headers_buf_obj.Length();
+    std::string query = context.Get("query").ToString();
+    input_format = context.Get("input_format").ToString();
+
+    s3select.parse_query(query.c_str());
+    if (!s3select.get_error_description().empty()) {
+        throw Napi::Error::New(info.Env(), XSTR() << "s3select: parse_query failed " << s3select.get_error_description());
+    }
+
+    if (input_format == JSON_FORMAT) {
+        json_object = new s3selectEngine::json_object(&s3select);
+    } else if (input_format == CSV_FORMAT) {
+        s3selectEngine::csv_object::csv_defintions csv_defs;
+        csv_defs.row_delimiter = GetStringWithDefault(input_serialization_format, "RecordDelimiter", "\n").c_str()[0];
+        csv_defs.column_delimiter = GetStringWithDefault(input_serialization_format, "FieldDelimiter", ",").c_str()[0];
+        csv_defs.use_header_info = (0 == GetStringWithDefault(input_serialization_format, "FileHeaderInfo", "IGNORE").compare("USE"));
+        csv_defs.quote_fields_always = false;
+        csv_object = new s3selectEngine::csv_object(&s3select, csv_defs);
+    } else {
+        throw Napi::Error::New(info.Env(), XSTR() << "input_format is neither csv nor json");
+    }
+}
+
+S3SelectNapi::~S3SelectNapi()
+{
+    if (nullptr != csv_object) {
+        delete csv_object;
+    }
+    if (nullptr != json_object) {
+        delete json_object;
+    }
+}
+
+} // namespace noobaa
--- a/src/tools/s3select.js
+++ b/src/tools/s3select.js
@@ -1,0 +1,59 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const { S3Select } = require('../util/nb_native')();
+const { Transform } = require('readable-stream');
+const minimist = require('minimist');
+
+const argv = minimist(process.argv, {
+    default: {
+        input_format: 'CSV',
+        field_delimiter: ',',
+        record_delimiter: '\n',
+        file_header_info: 'IGNORE'
+    }
+});
+
+class S3SelectStream extends Transform {
+
+    constructor(opts = {}) {
+        super(opts);
+        const context = {
+            query: argv.query,
+            input_format: argv.input_format,
+            input_serialization_format: {
+                FieldDelimiter: argv.field_delimiter,
+                RecordDelimiter: argv.record_delimiter,
+                FileHeaderInfo: argv.file_header_info
+            },
+            records_header_buf: Buffer.from("don't care") /*Used for CRC, ignored here.*/
+        };
+        if (S3Select) {
+            this.s3select = new S3Select(context);
+        } else {
+            console.log("ERROR - nb_native doesn't have S3Select. Did you build with GYP_DEFINES=BUILD_S3SELECT=1?");
+            process.exit(1);
+        }
+    }
+
+    async _transform(chunk, encoding, cb) {
+        //console.log("got chunk ", chunk.length);
+        let res = await this.s3select.write(chunk);
+        if (res) {
+            this.push(res.select);
+        }
+        return cb();
+    }
+
+    async _flush(cb) {
+        //console.log("in flush");
+        let res = await this.s3select.flush();
+        if (res) {
+            this.push(res.select);
+        }
+        return cb();
+    }
+}
+
+const s3select = new S3SelectStream();
+process.stdin.pipe(s3select).pipe(process.stdout);
--- a/src/util/s3select.js
+++ b/src/util/s3select.js
@@ -1,0 +1,127 @@
+/* Copyright (C) 2023 NooBaa */
+'use strict';
+
+const dbg = require('./debug_module')(__filename);
+const { Transform } = require('stream');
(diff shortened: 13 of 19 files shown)


guymguym (changes requested):
As an initial comment, I would say that we shouldn't use s3lect - this is hard to read and write, and we had a similar case before (s3rver) which was quickly removed from the codebase. So Please revert to use s3select instead (both in file names and in code). Thanks

src/native/ceph_s3select/example.js
```diff
@@ -0,0 +1,30 @@
...
+'use strict';
+
+const nb_native = require('../../util/nb_native');
+const { Transform } = require('readable-stream');
+
+class S3SelectStream extends Transform {
+
+    constructor(opts = {}) {
+        super(opts);
+        var context = {query: process.argv[2], input_format: 'csv', row_delimiter: '\n', column_delimiter: ',', use_header_info: false, quote_fields_always: false};
```
guymguym, line 11: line is too long, add new lines after commas.

src/native/ceph_s3select/example.js
```diff
@@ -0,0 +1,30 @@
...
+        this.CephS3lect = new (nb_native().CephS3lect)(context);
```
guymguym, line 12: variables/members/functions should be named in snake-case - meaning `this.ceph_s3_select = ...`.

src/native/ceph_s3select/ceph_s3lect_napi.cpp
```diff
@@ -0,0 +1,428 @@
...
+        //ReportWorkerStats(1);
+        auto obj = error.Value();
+        /*if (_errno) {
+            obj.Set("code", Napi::String::New(env, uv_err_name(uv_translate_sys_error(_errno))));
+        }*/
+        //_deferred.Reject(obj);
+    }
+};
+
+class CephS3lectNapi : public Napi::ObjectWrap<CephS3lectNapi> {
```
guymguym, line 294: I don't think the class name needs to have Ceph in its name.
The fact that this library is part of ceph is not important, and it has nothing to do with ceph itself.
I would just name this S3SelectNapi.

guymguym: This is a general comment - I don't think we need to name any of our types as CephXXX. This is confusing and makes readers think that we are using ceph codebase for this functionality, but we only use the s3select library which is a completely decoupled lib.

src/native/ceph_s3select/ceph_s3lect_napi.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+        public:
```
guymguym, line 296: indentation of public/private should be at the 0 column.

src/native/ceph_s3select/ceph_s3lect_napi.cpp
```diff
@@ -0,0 +1,428 @@
...
+            ~CephS3lectNapi();
+
+        private:
+            static Napi::FunctionReference constructor;
+            InputFormat input_format;
+            s3selectEngine::s3select s3select;
+            s3selectEngine::csv_object *csv_object = NULL;
+            s3selectEngine::json_object *json_object = NULL;
+            Napi::Promise::Deferred *deferred = NULL;
+		    CephSelectWorker worker;
```
guymguym, line 310: indentation. you should use the formatting rules as defined in the repo -
https://github.com/noobaa/noobaa-core/blob/master/.clang-format

I use the vscode c++ extension to format native code, and it reads that config file.

src/native/ceph_s3select/ceph_s3lect_napi.cpp
```diff
@@ -0,0 +1,428 @@
...
+            std::string select, chunk;
+
+};
+
+Napi::Value CephS3lectNapi::Write(const Napi::CallbackInfo& info){
+	//TODO type checks?
+	Napi::Buffer<char> buffer = info[0].As<Napi::Buffer<char>>();
```
guymguym, line 317: Why char? use Buffer<uint8_t>. See other examples in the repo.

src/native/ceph_s3select/ceph_s3lect_napi.cpp
```diff
@@ -0,0 +1,428 @@
...
+
+    /*std::cout << "write " << buffer << std::endl;
+    std::cout.flush();*/
+
+    if (NULL != deferred) {
```
guymguym, line 322: what is this doing? and why?

src/native/ceph_s3select/ceph_s3lect_napi.cpp
```diff
@@ -0,0 +1,428 @@
...
+#include "../../../submodules/boost/libs/crc/include/boost/crc.hpp"
+#include <arpa/inet.h>
+
+namespace noobaa {
+
+DBG_INIT(0);
+
+enum InputFormat {csv, json, parquet};
+
+class awsCli_handler {
```
guymguym, line 15: move to js. also this name seems very confusing - how is this related to awscli??

src/endpoint/s3/s3_rest.js
```diff
@@ -52,7 +52,8 @@ const OBJECT_SUB_RESOURCES = Object.freeze({
     'uploads': 'uploads',
     'uploadId': 'uploadId',
     'legal-hold': 'legal_hold',
-    'retention': 'retention'
+    'retention': 'retention',
+    'select': 'select'
```
guymguym, line 56:
```suggestion
    'select': 'select',
```

.gitmodules
```diff
@@ -0,0 +1,6 @@
+[submodule "ceph_s3select"]
+	path = submodules/ceph_s3select
+	url = https://github.com/ceph/s3select
```
guymguym, line 3: add an exact commit/branch ref to pin the dependency.

src/deploy/NVA_build/builder.Dockerfile
```diff
@@ -56,5 +57,7 @@ RUN stable_version=$(curl -s https://storage.googleapis.com/kubernetes-release/r
     curl -LO https://storage.googleapis.com/kubernetes-release/release/${stable_version}/bin/linux/${arch}/kubectl && \
     chmod +x ./kubectl
 
+
+
```
guymguym, line 61: clean unneeded diff?

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,185 @@
...
+ */
+async function post_object_select(req, res) {
+
+    req.object_sdk.setup_abort_controller(req, res);
+    const agent_header = req.headers['user-agent'];
+    const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
+    const encryption = s3_utils.parse_encryption(req);
+    let part_number;
+    // If set, part_number should be positive integer from 1 to 10000
+    if (req.query.partNumber) {
```
guymguym, line 69: part number? how is that related here?

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,185 @@
...
+        part_number = s3_utils.parse_part_number(req.query.partNumber, S3Error.InvalidArgument);
+    }
+    const md_conditions = http_utils.get_md_conditions(req);
```
guymguym, line 72: md_conditions doesn't seem relevant.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,185 @@
...
+    };
+    if (req.query.get_from_cache !== undefined) {
+        md_params.get_from_cache = true;
+    }
+    if (part_number) {
+        md_params.part_number = part_number;
+    }
+    const object_md = await req.object_sdk.read_object_md(md_params);
+
+    s3_utils.set_response_object_md(res, object_md);
```
guymguym, line 89: is this correct?? adding these response headers doesn't seem related to this op.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,185 @@
...
+        const end = params.end === undefined ? params.object_md.size : Math.min(params.end, params.object_md.size);
+        if (params.object_md.first_range_data.length >= end) {
+            const sliced_data = params.object_md.first_range_data.slice(start, end);
+            res.end(sliced_data);
+            return;
+        }
+    }
+
+    const read_stream = await req.object_sdk.read_object_stream(params, res);
+    if (read_stream) {
```
guymguym, line 155: what if not?

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,185 @@
...
+        // by calling the close method to stop it from buffering more data which will go to waste.
+        if (read_stream.close) {
+            req.object_sdk.add_abort_handler(() => read_stream.close());
+        }
+        read_stream.on('error', err => {
+            dbg.log0('read stream error:', err, req.path);
+            res.destroy(err);
+        });
+        var http_req_select_params = req.body.SelectObjectContentRequest;
+        var select_args = {query: http_req_select_params.Expression[0], input_format: 'csv', row_delimiter: '\n', column_delimiter: ',', use_header_info: false, quote_fields_always: false};
```
guymguym, line 166: line too long.

src/native/ceph_s3select/ceph_s3lect_napi.cpp
alphaprinz: Each write has its own deferred object.
When a new write starts, the previously-used deferred object is deleted.
Or do you mean the can by "recycled"?

.gitmodules
alphaprinz: I'm went over https://git-scm.com/docs/gitmodules and didn't see relevant option.
Am I missing something?

guymguym: in gitmodules we can use the branch attribute.

if there is no upstream branch we want to track, and we need a specific sha we should use this method - https://stackoverflow.com/questions/8813249/git-submodules-specify-a-specific-sha

.gitmodules
```diff
@@ -0,0 +1,6 @@
...
+[submodule "boost"]
+	path = submodules/boost
+	url = https://github.com/boostorg/boost
```
guymguym, line 6: I want to discuss the Boost dependency as git submodule with @liranmauda and @dannyzaken too. 

I see it like this - s3select library has to come from git, because afaik there is no other way to get it from, not upstream, and not downstream, so that's simple, and in any case it's small, even with those two added csv&json libs.

Boost however is a big library, and can be complex to build (think about FIPS, architectures, etc). I don't think downstream will use this gitmodules approach, but instead will use the boost package from downstream - what do you think @liranmauda ?. For upstream container we can probably just use the boost-dev package from centos and not have to think about it much. On Mac we also have `brew install boost` which worked for me (maybe need to configure a path somewhere but easy enough).

Consider that Boost is just the beginning - we will also need Apache Arrow to process Parquet structure, and I can see in https://arrow.apache.org/install/ that both centos8/9 and brew have it, so that's another good sign that we will be able to use this approach going forward.

It's not that relying on having boost&arrow as a prerequisite on the is really a better approach, but I think it will be simpler to maintain than adding those as submodules as they are big libraries.

So basically I suggest to:
1. leave just s3select as gitmodule
2. update docs/wiki on how to install the boost prereq on mac/linux for building with s3select enabled.
3. change the upstream dockerfile to `dnf install boost...` instead of fetching it from git.

Would be great to hear your thoughts and considerations.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,261 @@
...
+    }
+
+    return Buffer.from(byteArray);
+};
+
+class S3SelectStream extends Transform {
+
+    constructor(opts) {
+        super(opts);
+        this.CephS3lect = new (nb_native().CephS3lect)(opts);
```
guymguym, line 28: when we build without BUILD_S3SELECT=0 this CephS3select would be undefined and this will fail miserably. Can you add a more explicit check in the beginning of the operation that will fail but reply a more easy to understand error that this library was not built and then fail?

src/endpoint/s3/s3_utils.js
```diff
@@ -79,6 +79,7 @@ const OP_NAME_TO_ACTION = Object.freeze({
     post_object: { regular: "s3:putobject" },
     post_object_uploadId: { regular: "s3:putobject" },
     post_object_uploads: { regular: "s3:putobject" },
+    post_object_select: { regular: "s3:selectobjectcontent" },
```
guymguym, line 82: Maybe I'm mixing up but is this going to affect bucket policy permissions? so is there a special bucket policy that needs to be allowed for s3 select?

src/native/ceph_s3select/ceph_s3lect.gyp
```diff
@@ -0,0 +1,19 @@
+{
+    'includes': ['../third_party/common_third_party.gypi'],
+    'targets': [{
+        'target_name': 'ceph_s3lect',
```
guymguym, line 4: please rename to simple s3select

guymguym (changes requested):

.gitmodules
liranmauda: Boost - we have it downstream, we can just install it.
Arrow - I am not sure what I have found, I think we have it.
s3select - We should consult Boris to see how can we build it downstream, and if RGW doing/will do that so we can get the same process.
@b-ranto It will be also great to hear your thoughts and considerations.

src/native/ceph_s3select/ceph_s3lect_napi.cpp
alphaprinz: run_s3select_on_stream() expects char*. Using another type for buffer would reuire a type conversion somewhere along the way.

src/endpoint/s3/ops/s3_post_object_select.js
alphaprinz: I will end http response with status code 500 and nothing else.

src/endpoint/s3/s3_utils.js
alphaprinz: I don't see something like this.
According to https://docs.aws.amazon.com/AmazonS3/latest/userguide/selecting-content-from-objects.html you (just) need s3:GetObject.

dannyzaken:
@alphaprinz - for future reference and to make life easier to reviewers :) 
* please edit the PR description to add more info\abstract on the PR itself
* let's add a dev doc md as we discussed, under `/docs`. also add the design doc under `/doc/design`

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,206 @@
...
+        //prelude:
+        index = buffer.writeUInt32BE(total_length_bytes);
+        index = buffer.writeUInt32BE(records_message_headers.length, index);
+        index = buffer.writeInt32BE(crc32.buf(buffer.slice(0, index)), index); //prelude crc
+        //headers:
+        index += records_message_headers.copy(buffer, index);
+        //body:
+        index += select_result.copy(buffer, index);
+        //message crc:
+        buffer.writeInt32BE(crc32.buf(buffer.slice(0, index)), index);
```
dannyzaken, line 67: according to the docs `Buffer.slice` is deprecated and we should use `subarray` instead
https://nodejs.org/api/buffer.html#bufslicestart-end

docs/design/s3select.md
```diff
@@ -0,0 +1,121 @@
...
+### New S3 OP
+
+A new s3 op, post_object_select will be added.
+
+There are two variations of files to consider:
+
+
+
+1. csv, json - These are “stream-friendly”, they are read once from start to finish. So we can get their stream like standard get_object s3 op, pipe the stream to ceph to process, and pipe the result to http res object.
+2. Parquet
```
shirady, line 40: Maybe add a comment that Parquet is not supported in our implementation yet.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,206 @@
...
+        this.push(chunk);
+        return cb();
+    }
+}
+
+/**
+ * https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html
+ */
+// eslint-disable-next-line max-statements
+async function post_object_select(req, res) {
```
shirady, line 120: I think this function should apear at the begining of the file.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,206 @@
...
+
+    const read_stream = await req.object_sdk.read_object_stream(params, res);
+    if (read_stream) {
+        // if read_stream supports closing, then we handle abort cases such as http disconnection
+        // by calling the close method to stop it from buffering more data which will go to waste.
+        if (read_stream.close) {
+            req.object_sdk.add_abort_handler(() => read_stream.close());
+        }
+        read_stream.on('error', err => {
+            dbg.log0('read stream error:', err, req.path);
```
shirady, line 166: Why `dbg.log0()` and not `dbg.error()`?

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,206 @@
...
+        });
+        const http_req_select_params = req.body.SelectObjectContentRequest;
+        const input_serialization = http_req_select_params.InputSerialization[0];
+        let input_format = null;
+        if ('CSV' in input_serialization) {
+            input_format = 'CSV';
+        } else if ('JSON' in input_serialization) {
+            input_format = 'JSON';
+        } else {
+            throw new S3Error(S3Error.InvalidArgument);
```
shirady, line 177: Did you check what AWS returns exactly?
Maybe we can improve the information in the error we return.
If so, there is a part in s3Error of "Errors actually returned by AWS S3 although not documented" you can a new error there.

src/native/nb_native.gyp
```diff
@@ -16,15 +16,24 @@
...
             [ 'node_arch=="arm64"', {
                 'cflags': ['-DUSE_NEON']
-            }],
+            }]
         ],
     },
 
     'targets': [{
+	'variables': {
+            'BUILD_S3SELECT%':0
```
shirady, line 25: What is the sign %?

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,206 @@
...
+        index += select_result.copy(buffer, index);
+        //message crc:
+        buffer.writeInt32BE(crc32.buf(buffer.slice(0, index)), index);
+
+        //dbg.log2("select content response = ", buffer.toString('hex').match(/../g).join(' '));
+        return buffer;
+    }
+
+    async handle_result(result, cb) {
+        if (result.length > 0) {
```
shirady, line 74: What if `result` is null or undefined?
I would add `if (result && ...`

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,206 @@
...
+        dbg.log2(this.header, chunk.toString());
+        this.push(chunk);
+        return cb();
+    }
+}
+
+/**
+ * https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html
+ */
+// eslint-disable-next-line max-statements
```
shirady, line 119: Can we delete this comment?

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,206 @@
...
+            query: http_req_select_params.Expression[0],
+            input_format: input_format,
+            input_serialization_format: http_req_select_params.InputSerialization[0][input_format][0]
+        };
+        const s3select = new S3SelectStream(select_args);
+        s3select.on('end', () => {
+            res.write(end_message);
+            res.end();
+        });
+        dbg.log3("select_args = ", select_args);
```
shirady, line 189: Is there a logic why is it in debug level 3?
there is also a print in debug level 2
`dbg.log2("select res = "`

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,230 @@
...
+namespace noobaa {
+
+DBG_INIT(0);
+
+enum InputFormat {csv, json, parquet};
+
+class SelectWorker : public Napi::AsyncWorker
+{
+public:
+	Napi::Promise::Deferred *deferred = NULL;
```
shirady, line 16: Why `NULL` and not `nullptr`?
(I comment on this line, but it is in general in cpp files).

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,206 @@
...
+
+    constructor(opts) {
+        super(opts);
+        this.s3select = new (nb_native().S3Select)(opts);
+    }
+
+    async encode_chunk(select_result) {
+        dbg.log2("select res = ", select_result.toString());
+
+        const total_length_bytes = select_result.length + records_message_headers.length + 16;
```
shirady, line 54: Please change the 'magic number' (16) so that it would be easier to maintain the code.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,219 @@
...
+            input_serialization_format: http_req_select_params.InputSerialization[0][input_format][0],
+            http_res: res
+        };
+        const s3select = new S3SelectStream(select_args);
+        /*s3select.on('end', () => {
+            res.write(end_message);
+            res.end();
+        });*/
+        dbg.log3("select_args = ", select_args);
+        read_stream.pipe(new EchoDebugStream({header: "before = "})).pipe(s3select).pipe(new EchoDebugStream({header: "after = "})).pipe(res);
```
guymguym, line 203: calling .pipe.pipe.pipe is not handling all the error propagation and destroyed cases well enough. for more context see this test code - [test_pipeline_stuck_on_destroyed_stream.js](https://github.com/noobaa/noobaa-core/blob/master/src/test/unrelated/test_pipeline_stuck_on_destroyed_stream.js).

this is why we have our pipeline function in [stream_utils.js](https://github.com/noobaa/noobaa-core/blob/579969198fff7c53a68a5372080040ef8b8a5ff4/src/util/stream_utils.js#L32) that wraps util.pipeline and adds some more checks.

src/native/nb_native.gyp
alphaprinz: This is GYP syntax for default variable values.
https://gyp.gsrc.io/docs/InputFormatReference.md#Providing-Default-Values-for-Variables

src/endpoint/s3/ops/s3_post_object_select.js
alphaprinz: Not unless there's a good way to reduce number of lines in that method.

src/endpoint/s3/ops/s3_post_object_select.js
alphaprinz: There's no obvious guideline. I think this is "debuggy" enough to warrant level3.

src/native/s3select/s3select.gyp
```diff
@@ -0,0 +1,19 @@
+{
+    'includes': ['../third_party/common_third_party.gypi'],
+    'targets': [{
+        'target_name': 's3select',
+        'type': 'static_library',
+        'cflags_cc!': ['-fno-rtti'],
```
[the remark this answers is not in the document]
alphaprinz: There's a bunch of error with no-rtti, so I guess yes.
../src/native/s3select/../../../submodules/s3select/include/s3select_oper.h:1846:44: error: ‘dynamic_cast’ not permitted with -fno-rtti
     l = dynamic_cast<arithmetic_operand*>(p)->l;

src/endpoint/s3/ops/s3_post_object_select.js
shirady: Easy way - to separate the function into more functions, from what I see the function length is over 70 lines (in `eslint` it was defined as a max of 60 lines with an ambition to reduce it to 40 lines).

src/endpoint/s3/ops/s3_post_object_select.js
shirady: @alphaprinz, Would you please move these classes to [s3_utils](https://github.com/noobaa/noobaa-core/blob/master/src/endpoint/s3/s3_utils.js) and just require at the top as @guymguym suggested? (I would like your main function to appear at the beginning of the file).

src/endpoint/s3/s3_errors.js
```diff
@@ -144,6 +144,11 @@ S3Error.InvalidArgument = Object.freeze({
     message: 'Invalid Argument',
     http_code: 400,
 });
+S3Error.MissingInputSerialization = Object.freeze({
+    code: 'MissingRequiredParameter',
+    message: 'InputSerialization is required. Please check the service documentation and try again.',
```
shirady, line 149: @alphaprinz, does this error appear in the [link above](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html
)? (I saw it [here](https://docs.aws.amazon.com/AmazonS3/latest/API/API_InputSerialization.html)).

src/endpoint/s3/s3_utils.js
```diff
@@ -668,6 +671,109 @@ function response_field_encoder_url(value) {
...
+        //headers:
+        this.push(S3SelectStream.records_message_headers);
+        //body:
+        this.push(select_result.select);
+        //message crc:
+        const buffer2 = Buffer.alloc(S3SelectStream.crc_length_bytes);
+        buffer2.writeInt32BE(select_result.message_crc);
+        this.push(buffer2);
+
+        //dbg.log2("select content response = ", buffer.toString('hex').match(/../g).join(' '));
```
shirady, line 743: @alphaprinz, please don't leave lines of code in the comments.
If you think it can be useful to us you can change the debug level to higher.

src/endpoint/s3/s3_utils.js
```diff
@@ -668,6 +671,109 @@ function response_field_encoder_url(value) {
...
+    Buffer.from('End'),
+    Buffer.from([13]), //again for second header
+    Buffer.from(':message-type'),
+    Buffer.from([7, 0, 5]),
+    Buffer.from('event'), //end of HEADERS
+    //no PAYLOAD (DATA ends)
+    Buffer.from([0xfe, 0x2c, 0xee, 0x99]) //MESSAGE CRC (crc of above)
+    ]);
+
+//same for headers of records message
```
shirady, line 696: I think an indentation is missing here (as I saw in your code style in the lines before).

src/native/s3select/example.js
```diff
@@ -0,0 +1,50 @@
...
+        super(opts);
+        const context = {
+            query: process.argv[2],
+            input_format: 'CSV',
+            input_serialization_format: {
+                FieldDelimiter: ',',
+                RecordDelimiter: '\n',
+                FileHeaderInfo: 'IGNORE'
+            },
+            records_header_buf: Buffer.from("dont' care") /*Used for CRC, ignored here.*/
```
shirady, line 19: I guess it's supposed to be `don't`

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,256 @@
...
+        , buffer(buffer)
+        , buffer_len(buffer_len)
+        , is_flush(is_flush)
+        , input_format(input_format)
+        , deferred(env)
+        , headers_bytes(headers_bytes)
+        , headers_len(headers_len)
+    {}
+
+    /*~SelectWorker() {
```
shirady, line 54: Was it accidentally left (to check that the object was destroyed)? @alphaprinz

src/endpoint/s3/s3_errors.js
alphaprinz: Yes. This message is what AWS returns (for this bad request).

src/endpoint/s3/s3_utils.js
alphaprinz: I think it's useful for someone trying to debug this code.

shirady:
@alphaprinz I added some comments.
in general, since this PR is in its final stage, please remove the code in the comments: you can either delete it or if it is useful you can set them to a high debug level (I'm sure a also have a solution for the cpp files).

src/endpoint/s3/s3_errors.js
shirady: I'm sure.

In this file the errors are separated:
1. `Errors documented in AWS error pages` - in the link http://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html (currently you put the error there).
3. `Errors actually returned by AWS S3 although not documented`

I asked since I didn't see it in the link of Errors documented in AWS error pages.

alphaprinz: The error code MissingRequiredParameter is documented in AWS (https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#SelectObjectContentErrorCodeList).
The message (spefically for SerlizationInput param) is copied from AWS response.
So as far as I understand the location is ok?

shirady: Thanks, yes.

src/native/s3select/s3select_napi.cpp
shirady: Maybe we can use a debug flag and when we set it we will have all the cpp code for debug use.

src/endpoint/s3/s3_utils.js
shirady: Great, we can uncomment it and set a high debug level in advance.

shirady: maybe add a comment before it, for example:
```
// Uncomment this transforms the buffer into a human-readable string.
// it is not at any debug level since of heavy CPU costs and will be useful only in specific debug scenarios
```

src/deploy/NVA_build/Base.Dockerfile
```diff
@@ -19,7 +19,15 @@ RUN npm install --production && \
 ##############################################################
 COPY ./binding.gyp .
 COPY ./src/native ./src/native/
-RUN npm run build:native
+COPY ./src/deploy/NVA_build/clone_submodule.sh ./src/deploy/NVA_build/
+ARG BUILD_S3SELECT=1
+#Clone S3Select and its two submodules, but only if BUILD_S3SELECT=1.
+RUN ./src/deploy/NVA_build/clone_submodule.sh submodules/s3select https://github.com/ceph/s3select 58875a5ee76261cc0a3d943bb168f9f9292c34a4 BUILD_S3SELECT
+RUN ./src/deploy/NVA_build/clone_submodule.sh submodules/s3select/rapidjson https://github.com/Tencent/rapidjson fcb23c2dbf561ec0798529be4f66394d3e4996d8 BUILD_S3SELECT
+RUN ./src/deploy/NVA_build/clone_submodule.sh submodules/s3select/include/csvparser https://github.com/ben-strasser/fast-cpp-csv-parser 5a417973b4cea674a5e4a3b88a23098a2ab75479 BUILD_S3SELECT
```
guymguym, line 27: How about moving all this to a script of its own so that users can call it manually too if needed?

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html
+ */
+async function post_object_select(req, res) {
+
+    if (nb_native().S3Select === undefined) {
+        dbg.error("Request for object select but compiled without BUILD_S3SELECT=1");
+        res.statusCode = 501;
+        res.end();
+        return;
```
guymguym, line 19: I wouldn't send just status code without a message - this will be very hard for users to figure out. 
What you should do here is throw an S3Error that will produce the correct code and message.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,97 @@
...
+
+    const md_params = {
+        bucket: req.params.bucket,
+        key: req.params.key,
+        version_id: req.query.versionId,
+        encryption,
+    };
+    if (req.query.get_from_cache !== undefined) {
+        md_params.get_from_cache = true;
+    }
```
guymguym, line 35: This can be removed - this is a non standard query that we added to get object just for testing/troubleshooting of namespace_cache. If we ever decide to support this here we can add, but for now it seems irrelevant.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,97 @@
...
+        obj_id: object_md.obj_id,
+        bucket: req.params.bucket,
+        key: req.params.key,
+        content_type: object_md.content_type,
+        noobaa_trigger_agent,
+        encryption,
+    };
+    if (md_params.get_from_cache) {
+        params.get_from_cache = true;
+    }
```
guymguym, line 49: This can be removed too.

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,251 @@
...
+}
+
+Napi::FunctionReference S3SelectNapi::constructor;
+
+Napi::Object
+S3SelectNapi::Init(Napi::Env env, Napi::Object exports)
+{
+    Napi::HandleScope scope(env);
+
+    Napi::Function func = DefineClass(env, "S3SelectNapi", { InstanceMethod("write", &S3SelectNapi::Write), InstanceMethod("flush", &S3SelectNapi::Flush) });
```
guymguym, line 173: add line breaks for readability -
```suggestion
    Napi::Function func = DefineClass(env, "S3SelectNapi", {
        InstanceMethod("write", &S3SelectNapi::Write),
        InstanceMethod("flush", &S3SelectNapi::Flush),
    });
```

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    SelectWorker* worker = new SelectWorker(info.Env(), *csv_object, *json_object, buffer.Data(), buffer.Length(), false, input_format, headers_buf, headers_len);
+    worker->Queue();
+    return worker->deferred.Promise();
+}
+
+Napi::Value
+S3SelectNapi::Flush(const Napi::CallbackInfo& info)
+{
+    SelectWorker* worker = new SelectWorker(info.Env(), *csv_object, *json_object, nullptr, 0, true, input_format, headers_buf, headers_len);
```
guymguym, line 161: add line breaks for readability -
```suggestion
    SelectWorker* worker = new SelectWorker(
        info.Env(),
        *csv_object,
        *json_object,
        nullptr,
        0,
        true,
        input_format,
        headers_buf,
        headers_len.
    );
```

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+Napi::Value
+S3SelectNapi::Write(const Napi::CallbackInfo& info)
+{
+    Napi::Buffer<char> buffer = info[0].As<Napi::Buffer<char>>();
+
+    /*std::cout << "write len = " << buffer.Length() <<" data = " << buffer.Data() << std::endl;
+    std::cout.flush();*/
+
+    SelectWorker* worker = new SelectWorker(info.Env(), *csv_object, *json_object, buffer.Data(), buffer.Length(), false, input_format, headers_buf, headers_len);
```
guymguym, line 153: add line breaks for readability...

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,251 @@
...
+        , csv_object(csv_object)
+        , json_object(json_object)
+        , buffer(buffer)
+        , buffer_len(buffer_len)
+        , is_flush(is_flush)
+        , input_format(input_format)
+        , deferred(env)
+        , headers_bytes(headers_bytes)
+        , headers_len(headers_len)
+    {}
```
guymguym, line 52: The worker has to hold a ref on the S3SelectNapi object to avoid gc which will make many pointers invalid here.
See how we did that in FSWrapWorker - https://github.com/noobaa/noobaa-core/blob/0a04de2f7e53289d42416d29f491940ce81ea8c7/src/native/fs/fs_napi.cpp#L488-L506

In addition, there is another lifetime issue about the args which needs to be handled like we do in FsWorker args_ref - https://github.com/noobaa/noobaa-core/blob/0a04de2f7e53289d42416d29f491940ce81ea8c7/src/native/fs/fs_napi.cpp#L378-L383

Lets discuss if this is unclear.

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    void Execute() override
+    {
+        this->select.clear();
```
guymguym, line 56: since the worker is not reused, this is unneeded (and therefore confusing).

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+        int rc;
+        // TODO - if we get total size on beginning, i think we can remove the is_flush if
+        if (csv == input_format) {
+            if (is_flush) {
+                rc = csv_object.run_s3select_on_stream(select, buffer, 0, 0);
```
guymguym, line 62: in flush case there is no buffer, so better send null instead to make it clearer.

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+        /*std::cout << "select res = " << this->select << std::endl;
+        std::cout.flush();*/
+    }
+
+    void OnOK() override
+    {
+        Napi::Env env = Env();
+        // in case of empty select result (ie, no rows in current buffer matched sql condition), return null
+        if (0 == select.length()) {
```
guymguym, line 90:
```suggestion
        if (select.empty()) {
```

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,251 @@
...
+    Napi::Promise::Deferred deferred;
+    s3selectEngine::csv_object& csv_object;
+    s3selectEngine::json_object& json_object;
+    std::string select;
+    const char* buffer;
+    const unsigned int buffer_len;
+    const bool is_flush;
+    const InputFormat input_format;
+    const u_int8_t* headers_bytes;
+    const u_int32_t headers_len;
```
guymguym, line 31: I suggest that you rename all members to start with `_` - this is the best way to identify what is a class member vs what is a local variable when reading the code flow.

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+        Napi::Object result = Napi::Object::New(env);
+        Napi::Value select_buf = Napi::Buffer<char>::/*New*/ Copy(env, select.data(), select.length());
+        result.Set("select", select_buf);
+
+        u_int32_t prelude[3];
+        int32_t prelude_crc, message_crc, prelude_crc_BE;
+        prelude[0] = htonl(select.length() + headers_len + 16);
+        prelude[1] = htonl(headers_len);
+        prelude_crc = crc32_gzip_refl_base(0, (u_int8_t*)prelude, 8);
```
guymguym, line 103: use the standard type name `uint8_t` which is defined in cstdint.h - not sure where this `u_int8_t` comes from but I would use the standard anyway.

guymguym (changes requested):

src/native/s3select/s3select_napi.cpp
shirady: @guymguym, a prefix of `_` in a member name is usually a convention of a private member, from what I know `m` or `m_` prefix is for public members (It is a separate discussion about why those members are public, and I will not start it at this point...).

guymguym: You're right, but these are all "effectively" private members, the fact their visibility is public is unrelated to their use in the code and we never really share these classes outside this module. 

To be honest I am not a fan of encapsulation (although I was once a fan of it) but the more I used javascript/python and other dynamic languages, I got used to the model that anything is effectively public, and it is the code's job to explain why something shouldn't be used directly or how it should be used. 

In any case, for c++ classes/structs (which are not plain data objects but actually types with methods), the reason for prefixing members with `_` is just to make it easier to read their methods and quickly identify that we are accessing a member and not a local variable or function argument, because in c++ there is no requirement to use `this->` when accessing members.

guymguym: Oh and I would refrain from adding `m_` mainly because this changes the pronunciation of the names and makes it harder to express logic with it. e.g. `open(m_file_name, m_mode, m_options)` - yuck :)

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,289 @@
...
+        *csv_object,
+        *json_object,
+        buffer.Data(),
+        buffer.Length(),
+        false,
+        input_format,
+        headers_buf,
+        headers_len);
+    //prevent GC of this until worker is done
+    this->Ref();
```
guymguym, line 175: better move this ref to the worker ctor

src/endpoint/s3/s3_utils.js
```diff
@@ -12,6 +12,9 @@ const crypto = require('crypto');
 const config = require('../../.././config');
 const ChunkedContentDecoder = require('../../util/chunked_content_decoder');
 const stream_utils = require('../../util/stream_utils');
+const { Transform } = require('readable-stream');
```
guymguym, line 15: use the standard `require('stream')` instead

src/endpoint/s3/s3_utils.js
```diff
@@ -668,6 +671,111 @@ function response_field_encoder_url(value) {
...
+            return cb(err);
+        }
+    }
+
+    async _flush(cb) {
+        try {
+            const select_result = await this.s3select.flush();
+            await this.handle_result(select_result);
+            this.push(S3SelectStream.end_message);
+            this.http_res.end();
```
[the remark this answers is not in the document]
guymguym: you should be able to remove http_res from here, it doesn't belong, a stream is ended by pushing null.

src/tools/s3select.js
```diff
@@ -0,0 +1,50 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const nb_native = require('../util/nb_native');
+const { Transform } = require('readable-stream');
+
+class S3SelectStream extends Transform {
```
guymguym, line 7: cool tool, but don't you prefer to use the S3SelectStream from s3_utils? Doesn't it fit here?

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+Napi::Value
+S3SelectNapi::Write(const Napi::CallbackInfo& info)
+{
+    Napi::Buffer<char> buffer = info[0].As<Napi::Buffer<char>>();
+    SelectWorker* worker = new SelectWorker(
+        info,
+        *this,
+        *csv_object,
+        *json_object,
```
guymguym, line 167: It's not a good idea to dereference a nullptr (one of csv/json is always null).
I would send the pointer instead.

src/tools/s3select.js
alphaprinz: This can be run from command line. The S3SelectStream encodes the chunks in AWS format. So maybe if we add a flag to S3SelectStream to skip the encoding.

src/endpoint/s3/s3_errors.js
```diff
@@ -512,6 +517,16 @@ S3Error.InvalidEncodingType = Object.freeze({
     http_code: 400,
 });
 
+////////////////////////////////////////////////////////////////
+// S3 Select not compiled                                     //
+////////////////////////////////////////////////////////////////
+
+S3Error.S3SelectNotCompiled = Object.freeze({
+    code: 'S3 Select was not compiled.',
```
guymguym, line 525: error codes are CamelCase by existing standard so change to something like
```suggestion
    code: 'S3SelectNotCompiled',
```

src/endpoint/s3/s3_errors.js
```diff
@@ -512,6 +517,16 @@ S3Error.InvalidEncodingType = Object.freeze({
     http_code: 400,
 });
 
+////////////////////////////////////////////////////////////////
+// S3 Select not compiled                                     //
+////////////////////////////////////////////////////////////////
+
+S3Error.S3SelectNotCompiled = Object.freeze({
+    code: 'S3SelectNotCompiled.',
```
guymguym, line 525: remove the dot -
```suggestion
    code: 'S3SelectNotCompiled',
```

src/endpoint/s3/s3_utils.js
```diff
@@ -668,6 +672,109 @@ function response_field_encoder_url(value) {
     return new URLSearchParams({ 'a': value }).toString().slice(2); // slice the leading 'a='
 }
 
+class S3SelectStream extends Transform {
```
guymguym, line 675: It's not a small class, and this s3_utils module is big enough, so I would prefer to move it to it's own file, the natural pick would be `src/util/s3select.js`

src/endpoint/s3/ops/s3_post_object_select.js
guymguym: Something is still missing - did you try to run this through nsfs? 
If read_object_stream does not return a stream, we still need to pipe the s3select stream to the response.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,86 @@
...
+        bucket: req.params.bucket,
+        key: req.params.key,
+        content_type: object_md.content_type,
+        noobaa_trigger_agent,
+        encryption,
+    };
+
+    //prepare s3select stream
+    const http_req_select_params = req.body.SelectObjectContentRequest;
+    const input_serialization = http_req_select_params.InputSerialization[0];
```
guymguym, line 45: what about OutputSerialization ?

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,289 @@
...
+    } else if (!input_format_str.compare("CSV")) {
+        input_format = InputFormat::csv;
+    } else {
+        throw Napi::Error::New(info.Env(), XSTR() << "input_format is neither csv nor json");
+    }
+
+    s3select.parse_query(query.c_str());
+    if (!s3select.get_error_description().empty()) {
+        throw Napi::Error::New(info.Env(), XSTR() << "s3select: parse_query failed " << s3select.get_error_description());
+    }
```
guymguym, line 260: What will map all the s3select library errors to the correct protocol error types?
See https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html#SelectObjectContentErrorCodeList

src/util/s3select.js
```diff
@@ -0,0 +1,115 @@
...
+    ]);
+
+    //same for headers of records message
+    static records_message_headers = Buffer.concat([
+        S3SelectStream.header_to_buffer(':event-type', 'Records'),
+        S3SelectStream.header_to_buffer(':content-type', 'application/octet-stream'),
+        S3SelectStream.header_to_buffer(':message-type', 'event'),
+    ]);
+
+    constructor(opts) {
```
guymguym, line 49: Can you defined in a jsdoc what opts can be provided:
```suggestion
    /**
     * @param {{
     *     query: string;
     *     input_format: 'CSV' | 'JSON';
     *     input_serialization_format: {
     *        FieldDelimiter: string;
     *        RecordDelimiter: string;
     *        FileHeaderInfo: string;
     *     };
     *     records_header_buf: Buffer;
     *     ...
     *  }} opts
     */
    constructor(opts) {
```

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,86 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const dbg = require('../../../util/debug_module')(__filename);
+const S3Error = require('../s3_errors').S3Error;
+const s3_utils = require('../s3_utils');
+const s3_select_utils = require('../../../util/s3select');
```
guymguym, line 7:
```suggestion
const { S3SelectStream } = require('../../../util/s3select');
```

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+S3SelectNapi::S3SelectNapi(const Napi::CallbackInfo& info)
+    : Napi::ObjectWrap<S3SelectNapi>(info)
+{
+    Napi::Object context = info[0].As<Napi::Object>();
+    Napi::Object input_serialization_format = context.Get("input_serialization_format").As<Napi::Object>();
+    Napi::Buffer headers_buf_obj = context.Get("records_header_buf").As<Napi::Buffer<uint8_t>>();
+    headers_buf = headers_buf_obj.Data();
+    headers_len = headers_buf_obj.Length();
```
guymguym, line 245: Keeping the pointer to this memory is not safe because no one is aware that you are using this buffer's memory outside this ctor call. However since you pass in a static buffer then it is safe, but I wouldn't want to rely on this in here.

There are many ways to resolve this, most notably is to send this headers buffer on the write/flush calls as arguments, because there we already have an _args_ref that keeps the reference to it. But you can also do the same handling inside the S3SelectNapi object and keep its own _args_ref if you prefer.

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,289 @@
...
+    std::string query = context.Get("query").ToString();
+    std::string input_format_str = context.Get("input_format").ToString();
+
+    if (!input_format_str.compare("JSON")) {
+        input_format = InputFormat::json;
+    } else if (!input_format_str.compare("CSV")) {
+        input_format = InputFormat::csv;
```
guymguym, line 252: Comparing std::string is better with simple `==`, no need to use the compare method.
Also - why do we need to define another enum? I would simply send the string to the worker, and just define couple of consts to be used for checking equality - 
```c++
const char *CSV_FORMAT = "CSV";
const char *JSON_FORMAT = "JSON";
const char *PARQUET_FORMAT = "PARQUET";

...
if (input_format == CSV_FORMAT) {
  ...
} else if (input_format == JSON_FORMAT) {
  ...
}
```

src/tools/s3select.js
```diff
@@ -0,0 +1,50 @@
...
+
+    constructor(opts = {}) {
+        super(opts);
+        const context = {
+            query: process.argv[2],
+            input_format: 'CSV',
```
guymguym, line 13: Use `argv = require('minimist')` (search for code examples in the repo - there are many) and then you can send argv.query, argv.input_format, argv.field_delimiter etc... as flags (--query --input_format --field_delimiter) to test different options.

src/tools/s3select.js
```diff
@@ -0,0 +1,50 @@
...
+            input_serialization_format: {
+                FieldDelimiter: ',',
+                RecordDelimiter: '\n',
+                FileHeaderInfo: 'IGNORE'
+            },
+            records_header_buf: Buffer.from("don't care") /*Used for CRC, ignored here.*/
+        };
+        const nb_native_instance = nb_native();
```
guymguym, line 21: you can usually keep the nb_native function as is and just extract what you want to check from it like this:
```suggestion
        const { S3Select } = nb_native();
        if (!S3Select) throw new Error(...)
        this.s3select = new S3Select(...);
```

src/tools/s3select.js
```diff
@@ -0,0 +1,50 @@
...
+            this.s3select = new (nb_native_instance.S3Select)(context);
+        } else {
+            console.log("ERROR - nb_native doesn't have S3Select. Did you build with GYP_DEFINES=BUILD_S3SELECT=1?");
+            process.exit(1);
+        }
+    }
+
+    async _transform(chunk, encoding, cb) {
+        //console.log("got chunk ", chunk.length);
+        let select = await this.s3select.write(chunk);
```
guymguym, line 32: I'd rename to `res`
```suggestion
        let res = await this.s3select.write(chunk);
```

package-lock.json
```diff
@@ -1324,9 +1324,9 @@
       }
     },
     "node_modules/@npmcli/fs/node_modules/semver": {
-      "version": "7.3.7",
-      "resolved": "https://registry.npmjs.org/semver/-/semver-7.3.7.tgz",
-      "integrity": "sha512-QlYTucUYOews+WeEujDoEGziz4K6c47V/Bd+LjSSYcA94p+DmINdf7ncaUinThfvZyu13lN9OY1XDxt8C0Tw0g==",
+      "version": "7.3.8",
```
guymguym, line 1327: Is the diff on package-lock needed? Seems like a leftover from running npm update or installing additional packages that are no longer part of this PR...

src/endpoint/s3/ops/s3_post_object_select.js
guymguym: Some request parameters seem to be ignored.
Any reason why we cannot support it?

- [RequestProgress](https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html#AmazonS3-SelectObjectContent-request-RequestProgress)
- [ScanRange](https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html#AmazonS3-SelectObjectContent-request-ScanRange)
- [CompressionType](https://docs.aws.amazon.com/AmazonS3/latest/API/API_InputSerialization.html#AmazonS3-Type-InputSerialization-CompressionType)
- [OutputSerialization](https://docs.aws.amazon.com/AmazonS3/latest/API/API_SelectObjectContent.html#AmazonS3-SelectObjectContent-request-OutputSerialization)

src/native/s3select/s3select_napi.cpp
alphaprinz: headers_buf is a paramter in Worker ctor, so it's in the _args_ref there.
Or did I misunderstand?

guymguym: It's not, only the args that are provided in `info` are inserted into _args_ref.

src/endpoint/s3/ops/s3_post_object_select.js
alphaprinz: Generally, this are not implemeted (yet?). I can return 501 if one of them is in the request. Do you think it warrants a mail/meeting?

* Request Progress is optional. Also seems to somewhat contradict https://docs.aws.amazon.com/AmazonS3/latest/API/RESTSelectObjectAppendix.html which states "Stats message: Amazon S3 sends this message at the end of the request. It contains statistics about the query." I think this might be one of the differences between v1 and v2 of select object content.
* ScanRange - I can add it, if needed.
* CompressionType - I guess this is only relevant for FSNS?
* OutputSerlization - Implementing this sounds problematic because I don't see support for it in the s3select library so either we add ourselves to the lib or manipulate its output.

src/native/s3select/s3select_napi.cpp
alphaprinz: I will look into mapping the. It'll probably need to hard-coded manually. Also there's no guarantee that s3select in aws error codes will match nicely.

package-lock.json
alphaprinz: Reverted.

src/native/s3select/s3select_napi.cpp
alphaprinz: Right. Added to Worker's _args_ref.

src/native/s3select/s3select_napi.cpp
guymguym: In any case I would write the error mapping in js, but here we just want to propagate the information from the library inside the thrown js error object - an error is an object, so you can set more properties on it such as an `err.code` string.

Does the library provide structure error codes? or just a description? if there are no error codes enumerated I would raise an issue to the library asking for the error info in a structured manner.

src/endpoint/s3/ops/s3_post_object_select.js
guymguym: The concern I have with this current code is if combining the two separate pipeline calls works as expected - we never had this called this way - what we should verify is that propagating errors works well both from the reader to the response, and also in case of the client closing the connection then the response will raise errors and we should propagate it back to close our reader flows (either a stream or an async loop) and make sure we correctly handle aborting and discarding all the resources. It's hard to figure it all out by reading, so I would make a test for it.

src/endpoint/s3/ops/s3_post_object_select.js
guymguym: - ScanRange sounds simple.
- CompressionType is unrelated to nsfs, it's for any object which is a compressed json/csv. We can defer that for now and return error if requested.
- OutputSerialization - I am not following how this is working then, what is the "default" output serialization format that the library responds with?

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,277 @@
...
+        , _deferred(info.Env())
+        , _headers_bytes(headers_bytes)
+        , _headers_len(headers_len)
+    {
+        //take a ref on args to make sure they are not GCed until worker is done
+        uint32_t i;
+        for (i = 0; i < info.Length(); ++i) {
+            _args_ref.Set(i, info[i]);
+        }
+        _args_ref.Set(i, _headers_bytes);
```
guymguym, line 62: This won't work, you have to keep the ref to the original js buffer, not to the memory pointer...

src/native/s3select/s3select_napi.cpp
alphaprinz: S3select lib seem to go by free hard-coded text rather than enums/something structured.
Git-grep on base_s3select_excption brings a bunch of errors with free text as description and nothing else
(eg https://github.com/ceph/s3select/blob/2e96e6feeaf242f2adde4f050bf6a7f89cfd2c0f/include/s3select.h#L1280).

src/endpoint/s3/ops/s3_post_object_select.js
alphaprinz: *OutputSerialization - s3select response format is according to input - csv for csv, json for json. This is because it has different implementations for csv and json (s3selectEngine::json_object and s3selectEngine::csv_object).

guymguym: I see, so perhaps just validate that the request does not ask for converting to a different type, and accept if either not provided or matching the input type.

src/native/s3select/s3select_napi.cpp
guymguym: better raise an issue on it and ask for proper error codes that can be mapped for s3 compatibility.

guymguym: if you can open an issue and put the link here for future reference it would be great.

src/native/s3select/s3select_napi.cpp
alphaprinz: Moved to S3Select_Napi ctor.

src/endpoint/s3/ops/s3_post_object_select.js
```diff
@@ -0,0 +1,92 @@
...
+        // by calling the close method to stop it from buffering more data which will go to waste.
+        if (read_stream.close) {
+            req.object_sdk.add_abort_handler(() => read_stream.close());
+        }
+        read_stream.on('error', err => {
+            dbg.error('read stream error:', err, req.path);
+            res.destroy(err);
+        });
+        //in other cases, we need to pipe the read stream ourselves
+        stream_utils.pipeline([read_stream, s3select], true);
```
guymguym, line 80: @romayalon - We need to check about the last bool arg `reuse_last_stream` with you.
As the code is now, the last stream in this pipeline call is the s3select stream, which was already pipelined into res above, so it might not need to be resumed, but I'm not sure after reading the comment about it in stream_utils.pipeline.

src/native/s3select/s3select_napi.cpp
```diff
@@ -0,0 +1,279 @@
...
+        info,
+        *this,
+        csv_object,
+        json_object,
+        buffer.Data(),
+        buffer.Length(),
+        false,
+        input_format,
+        headers_buf,
+        headers_len);
```
guymguym, line 176: To make the headers buffer be safely ref/unref in the worker ctor/dtor with _args_ref you can just send it here to the function (and also in flush):
```suggestion
    Napi::Buffer<char> headers = info[0].As<Napi::Buffer<char>>();
    Napi::Buffer<char> buffer = info[1].As<Napi::Buffer<char>>();
    SelectWorker* worker = new SelectWorker(
        info,
        *this,
        csv_object,
        json_object,
        buffer.Data(),
        buffer.Length(),
        false,
        input_format,
        headers.Data(),
        headers.Length());
```

guymguym: Oh oops I just saw you added _args_ref to the S3SelectNapi object too, so that works as well. Resolving.

src/endpoint/s3/ops/s3_post_object_select.js
romayalon: if s3select stream is not needed anymore we need to call resume(), so the true argument should not be passed to the function.

guymguym: i am still confused - what does it mean "not needed anymore"?

romayalon: the s3select  stream won't be used after that line

guymguym (approved):

PR #7144, opened 2022-12-27
Adding ROOT key rotate to core

jackyalbo:

### Explain the changes
1. Adding rotate_root_key to system_api to allow key rotation of the ROOT_KEY for the system.
2. Admin that wants to rotate the ROOT_KEY needs to send the new ROOT_KEY as new_secret to this api, following with the changing the ROOT_KEY in the env variables
3. The function will decrypt all system master keys with the current ROOT_KEY and then re-encrypt it with the new secret
4. After that, the system will start re-encryption of all the other lower levels master keys in the system (like accounts for example)
5. Once finished the user will need to set the env variable ROOT_KEY to the new_secret so in the case of restart, keys will be able to decrypt.

Following this PR we will create a PR for NooBaa operator that will send this command every given time, so key periodic key rotation will be able to run automatically.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/api/system_api.js
+++ b/src/api/system_api.js
@@ -115,6 +115,20 @@ module.exports = {
             }
         },
 
+        rotate_root_key: {
+            method: 'PUT',
+            params: {
+                type: 'object',
+                required: ['new_root_key'],
+                properties: {
+                    new_root_key: { $ref: 'common_api#/definitions/secret_enc_key' },
+                }
+            },
+            auth: {
+                system: 'admin'
+            }
+        },
+
         rotate_master_key: {
             method: 'PUT',
             params: {
--- a/src/server/system_services/master_key_manager.js
+++ b/src/server/system_services/master_key_manager.js
@@ -58,19 +58,23 @@ class MasterKeysManager {
         return this.resolved_master_keys_by_id[ROOT_KEY];
     }
 
-    async load_root_key() {
+    load_root_key() {
         if (this.is_initialized) return;
         const { NOOBAA_ROOT_SECRET } = process.env;
         if (!NOOBAA_ROOT_SECRET) throw new Error('NON_EXISTING_ROOT_KEY');
+        this._update_root_key(NOOBAA_ROOT_SECRET);
+        this.is_initialized = true;
+    }
+
+    _update_root_key(new_root_key) {
         const r_key = {
             _id: ROOT_KEY,
-            cipher_key: Buffer.from(NOOBAA_ROOT_SECRET, 'base64'),
+            cipher_key: Buffer.from(new_root_key, 'base64'),
             cipher_type: "aes-256-gcm",
             description: "Root Master Key",
             disabled: false
         };
         this.resolved_master_keys_by_id[ROOT_KEY] = r_key;
-        this.is_initialized = true;
     }
 
     /**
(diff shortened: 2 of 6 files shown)


src/server/system_services/master_key_manager.js
```diff
@@ -73,6 +73,17 @@ class MasterKeysManager {
         this.is_initialized = true;
     }
 
+    async update_root_key(new_secret) {
+        const r_key = {
```
romayalon, line 77: should we somehow reuse load_root_key()? looks like they have similar logic besides 1/2 lines

src/server/system_services/master_key_manager.js
```diff
@@ -73,6 +73,17 @@ class MasterKeysManager {
         this.is_initialized = true;
     }
 
+    async update_root_key(new_secret) {
```
romayalon, line 76: redundant async

src/server/system_services/system_server.js
```diff
@@ -1282,6 +1282,24 @@ async function _get_endpoint_groups() {
...
 }
 
+/**
+ *
+ * ROTATE ROOT KEY
+ *
+ */
+async function rotate_root_key(req) {
+    const new_secret = req.rpc_params.new_secret;
+    const mkm = system_store.master_key_manager;
```
romayalon, line 1292: Why add a new function instead of using rotate_master_key and adding a new entity?

src/api/system_api.js
```diff
@@ -115,6 +115,20 @@ module.exports = {
             }
         },
 
+        rotate_root_key: {
+            method: 'PUT',
+            params: {
+                type: 'object',
+                required: ['new_secret'],
```
romayalon, line 122: using the secret term and not keeping the root_key term may be a bit confusing, don't you think?

src/test/unit_tests/test_encryption.js
```diff
@@ -813,6 +814,39 @@ mocha.describe('Rotation tests', function() {
                 }
             }));
         });
+
+        [true, false].forEach(update => {
```
romayalon, line 818: this test is a bit complicated, where did you check other secrets like chunk cipher keys ?

src/server/system_services/master_key_manager.js
```diff
@@ -163,6 +174,20 @@ class MasterKeysManager {
...
+     * 
+     * @param {String} m_key_id 
+     * @param {SensitiveString} new_secret 
+     * @returns {Object}  
+     */
+    _reencrypt_master_key_by_root(m_key_id, new_secret) {
+        const decrypted_mkey = this.get_master_key_by_id(m_key_id); // returns the decrypted m_key
+        if (!decrypted_mkey) throw new Error('NO_SUCH_KEY');
+        this.update_root_key(new_secret); // update ROOT_KEY with new secret provided
+        const reencrypted_mkey = this.encrypt_buffer_with_master_key_id(decrypted_mkey.cipher_key, ROOT_KEY);
```
romayalon, line 187: just return

src/server/system_services/system_server.js
```diff
@@ -1282,6 +1282,24 @@ async function _get_endpoint_groups() {
...
+    await P.all(_.map(system_store.data.systems, async function(system) {
+        const reencrypted = mkm._reencrypt_master_key_by_root(system.master_key_id._id, new_secret);
+        await upsert_master_key({
+            _id: system.master_key_id._id,
+            update: { cipher_key: reencrypted }
+        });
+        _rotate_master_key(system.name, 'SYSTEM');
```
romayalon, line 1299: Don't we want to keep the old root key in some way?

src/server/system_services/master_key_manager.js
jackyalbo: done

src/server/system_services/system_server.js
jackyalbo: as discussed... not the same flow and don't want to just add extra code to an already huge function

src/api/system_api.js
jackyalbo: fixed to new_root_key

src/server/system_services/system_server.js
jackyalbo: not as part of core - not sure if will be saved by operator also

romayalon (approved):
Approving with a comment about the old root key.
As discussed failures will be handled by the operator so the old secret won't be deleted if the core returned errors

PR #7146, opened 2022-12-29
NSFS | get_single_user_xattr() shouldnt throw in case of missing attribute

romayalon:

### Explain the changes
1. By default we do not calculatie md5 on nsfs (config.NSFS_CALCULATE_MD5 = false)
While specifying skip_user_xattr flag on stat() we will check for a constant list of extended attributes.
currently, the list contains 'user.content_md5', but it's empty by default for every object because config.NSFS_CALCULATE_MD5 is false. 
This missing attribute caused stat() to fail on _is_same_inode() check which caused potential redundant links/uploads.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -22,12 +22,17 @@
 #include <unistd.h>
 #include <uv.h>
 #include <vector>
+#include <errno.h>
 
 #ifdef __APPLE__
 #include <sys/mount.h>
 #include <sys/param.h>
 #else
 #include <sys/statfs.h>
+#endif
+
+#ifndef __APPLE__
+#define ENOATTR ENODATA
 #endif
 
 #define ROUNDUP(X, Y) ((Y) * (((X) + (Y)-1) / (Y)))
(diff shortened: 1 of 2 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -272,7 +273,14 @@ static int
 get_single_user_xattr(int fd, std::string key, std::string& value)
 {
     ssize_t value_len = fgetxattr(fd, key.c_str(), NULL, 0);
-    if (value_len == -1) return -1;
+    if (value_len == -1) {
+        int missing_attr_errno = ENODATA;
+        #ifdef __APPLE__
```
jackyalbo, line 278: Not better to put it in define?

src/native/fs/fs_napi.cpp
```diff
@@ -272,7 +273,14 @@ static int
...
+            missing_attr_errno = ENOATTR;
+        #endif
+        if (errno != missing_attr_errno) return -1;
+        value = "";
```
jackyalbo, line 282: I'm just not sure if we can have a value of empty string "" which id different from attributes that doesn't exist. Maybe do this error check in get_fd_xattr and then just skip. WDYT?

jackyalbo (approved):

PR #7154, opened 2023-01-04
NSFS | Versioning | ENABLED | Delete object / Delete multiple objects implementation

romayalon:
### Explain the changes

1. Delete object/ Delete multiple objects support deletion by versioning status.

2. Implementation of delete object versioning enabled in general - 
2.1. If the user passed version id -
      2.1.1. The actual version should be deleted
      2.1.2. If the deleted version is the latest version we should promote the second latest to be the latest
2.2. else 
      2.2.1. The latest version (if exists) should move to .versions/
      2.2.2. A delete marker should be created in .versions/
3. Added 3 new helper functions - _delete_object_versioning_enabled(), create_delete_marker(), find_max_version_past().
4. Added unit tests in test_bucketspace_versioning.js

1. Gaps - 
    1.1. Open question regarding the prev_version xattr optimization - on deletion of version id - how we can update the prev version xattr of the following version of the deleted one
    1.2. Tests for delete_multiple_objects / test_nsfs_versioning (currently I added the tests in test_bucketspace_versioning)

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1,4 +1,5 @@
 /* Copyright (C) 2020 NooBaa */
+/*eslint max-lines: ["error", 2500]*/
 'use strict';
 
 const _ = require('lodash');
@@ -8,6 +9,7 @@ const util = require('util');
 const mime = require('mime');
 const { v4: uuidv4 } = require('uuid');
 const dbg = require('../util/debug_module')(__filename);
+const P = require('../util/promise');
 
 const config = require('../../config');
 const s3_utils = require('../endpoint/s3/s3_utils');
@@ -38,9 +40,8 @@ const XATTR_USER_PREFIX = 'user.';
 const XATTR_MD5_KEY = XATTR_USER_PREFIX + 'content_md5';
 const XATTR_VERSION_ID = XATTR_USER_PREFIX + 'version_id';
 const XATTR_PREV_VERSION_ID = XATTR_USER_PREFIX + 'prev_version_id';
-//const XATTR_DELETE_MARKER = XATTR_USER_PREFIX + 'delete_marker';
-const HIDDEN_VERSIONS_PATH = '.versions';
 const XATTR_DELETE_MARKER = XATTR_USER_PREFIX + 'delete_marker';
+const HIDDEN_VERSIONS_PATH = '.versions';
 
 const versioning_status_enum = {
     VER_ENABLED: 'ENABLED',
@@ -550,7 +551,7 @@ class NamespaceFS {
         try {
             const fs_context = this.prepare_fs_context(object_sdk);
             await this._load_bucket(params, fs_context);
-            const file_path = this._find_version_path(fs_context, params);
+            const file_path = await this._find_version_path(fs_context, params);
             return fs.createReadStream(file_path, {
                 highWaterMark: config.NSFS_BUF_SIZE,
                 start: Number.isInteger(params.start) ? params.start : undefined,
@@ -809,6 +810,9 @@ class NamespaceFS {
         const part_upload = file_path === upload_path;
         const same_inode = params.copy_source && copy_res === copy_status_enum.SAME_INODE;
 
+        let stat = await target_file.stat(fs_context);
+        this._verify_encryption(params.encryption, this._get_encryption_info(stat));
+
         let fs_xattr;
         // handle xattr
         if (!params.copy_source || !params.xattr_copy) {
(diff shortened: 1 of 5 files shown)


src/sdk/namespace_fs.js
```diff
@@ -1220,16 +1225,41 @@ class NamespaceFS {
...
+                        await this._delete_path_dirs(file_path, fs_context);
+                        res.push({ key });
+                    } catch (err) {
+                        res.push({ err_code: err.code, err_message: err.message });
+                    }
+                }
+            } else {
+                // [{key: a, version: 1}, {key: a, version: 2}, {key:b, version: 1}] => {'a': [1, 2], 'b': [1]}
+                const versions_by_key_map = {};
+                params.objects.forEach(({ key, version_id }) => {
```
jackyalbo, line 1249: maybe use for of or try to see if you can use some help function of groupBy or reduce

src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1797,197 @@ class NamespaceFS {
...
+    // 4. if deleted version is a delete marker && max_past_version older than the deleted version 
+    //    4.1. move it file path (to be latest)
+    async _delete_object_versioning_enabled(fs_context, file_path, params) {
+        return params.version_id ?
+            this._delete_version_id(fs_context, file_path, params) :
+            this._delete_latest_version(fs_context, file_path, params);
+    }
+
+    // iterate over the key's versions array
+    // if version is undefined, delete latest
```
jackyalbo, line 1819: make this comment nested

src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1797,197 @@ class NamespaceFS {
...
+    // delete version_id - 
+    // 1. get version info, if it's empty - return
+    // 2. unlink key
+    // 3. promote second latest -> latest
+    async _delete_version_id(fs_context, file_path, params) {
+        const version_info = await this._get_version_info(fs_context, file_path);
+        if (!version_info) return {};
+        await nb_native().fs.unlink(fs_context, file_path);
+        await this._delete_path_dirs(file_path, fs_context);
+        await this._promote_version_to_latest(fs_context, params, version_info);
```
jackyalbo, line 1860: as we may delete the last version here we need to try and delete the path after the promotion

src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1797,197 @@ class NamespaceFS {
...
+                    res.push(version_res);
+                }
+            } catch (err) {
+                res.push({ err_code: err.code, err_message: err.message });
+            }
+        }
+        // delete .versions/ if it's empty
+        const file_path = this._get_version_path(key);
+        await this._delete_path_dirs(file_path, fs_context);
+        await this._promote_version_to_latest(fs_context, { key }, undefined);
```
jackyalbo, line 1847: as we may delete the last version here we need to try and delete the path after the promotion

jackyalbo, line 1847: we maybe here can avoid calling promote_to_latest in the cases we know for sure we don't need to promote like if we deleted only not-latest versions, subject to that we want to the process to handle just what it created and not fix missing promotions of other objects.

romayalon: As discussed, we would like to keep the current behavior and promote max past version only when needed.
This function can be reused in the future by a background worker that will try to fix invalid FS states, while not sending deleted_version_info and always promoting max past version if needed.

src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1800,213 @@ class NamespaceFS {
             throw new RpcError('BAD_REQUEST', 'Bad Request');
         }
     }
+
+    // delete latest version -
```
jackyalbo, line 1804: Do we need this aggregated comment? it seems both function have a comment of their own about what they do, I think it's long and unneeded, maybe take this comment from here and copy to the function as it seems this ones are more informative.

jackyalbo: maybe change to comment about this function gets a single key w/o version_id and delete it. Not suer if we need this function anyway, do we use it more than once? plan to use it more than once?

src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1800,213 @@ class NamespaceFS {
...
+    // 4. if deleted version is a delete marker && max_past_version older than the deleted version 
+    //    4.1. move it file path (to be latest)
+    async _delete_object_versioning_enabled(fs_context, file_path, params) {
+        return params.version_id ?
+            this._delete_version_id(fs_context, file_path, params) :
+            this._delete_latest_version(fs_context, file_path, params);
+    }
+
+    // 1. iterate over the key's versions array
+    //    1.1 if version is undefined, delete latest
```
jackyalbo, line 1822:
```suggestion
    //    1.1 if version_id is undefined, delete latest
```

src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1800,213 @@ class NamespaceFS {
...
+    //    2.1. if max_past_version does not exist / is a delete marker - skip, nothing to move
+    //    2.2. else - move max past version -> latest version path
+    // 3. if deleted version mtime < max_past_version mtime - skip (check if deleted version is latest or latest delete marker in .versions/)
+    // 4. move max past version -> latest version path
+    // conditions 1 and 2 guard on situations where we don't want to try move max version past to latest 
+    async _promote_version_to_latest(fs_context, params, deleted_version_info) {
+        dbg.log1('Namespace_fd._promote_version_to_latest', params, deleted_version_info);
+        const latest_ver_path = this._get_file_path({ key: params.key });
+        const deleted_latest = deleted_version_info && deleted_version_info.path === latest_ver_path;
+        // 1 - if deleted file is not a delete marker and not the latest - no need to promote max - return
```
jackyalbo, line 1894: not sure if we still need 1 and 2 comments... but whatever you prefer

src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -378,35 +379,753 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
             const comp_res = await compare_version_ids(full_path, key, res.VersionId);
             assert.ok(comp_res);
         });
+
+
+        mocha.it('delete object version id - disabled key', async function() {
```
jackyalbo, line 384: I didn't go over all the tests one by one... but in general it seems you didn't put any comments in it, maybe you will want to add some. at least about what you are doing, what you expect and things like that.

src/sdk/namespace_fs.js
romayalon: let's keep it for now, we saw it's not easy to understand

src/sdk/namespace_fs.js
romayalon: I removed it, for now, we might need it when we will start working on suspended versioning flows

jackyalbo (approved):

PR #7156, opened 2023-01-08
Consume secret data from a file.

liranmauda:
### Explain the changes

- Add a util to get data from a file. will use that to fetch Kubernetes secret mounted to a file.
- Replaced env.NOOBAA_AUTH_TOKEN with config.NOOBAA_AUTH_TOKEN
- Replaced env.JWT_SECRET wite config.JWT_SECRET
- Replaced env.SECRET_SERVER with config.SECRET_SERVER
- Keeping the default behavior of using env variables for overriding if needed or testing.

1. Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2152143

### Depend on:
- https://github.com/noobaa/noobaa-operator/pull/1015

### DIFF
--- a/src/util/os_utils.js
+++ b/src/util/os_utils.js
@@ -591,16 +591,13 @@ function is_folder_permissions_set(current_path) {
 }
 
 function read_server_secret() {
-    if (process.env.SERVER_SECRET) {
-        return process.env.SERVER_SECRET;
-    } else {
-        // in kubernetes we must have SERVER_SECRET loaded from a kubernetes secret
-        if (process.env.CONTAINER_PLATFORM === 'KUBERNETES') {
-            throw new Error('SERVER_SECRET env variable not found. it must exist when running in kubernetes');
-        }
-        // for all non kubernetes platforms (docker, local, etc.) return a dummy secret
-        return '12345678';
+    if (config.SERVER_SECRET) return config.SERVER_SECRET;
+    // in kubernetes we must have SERVER_SECRET loaded from a kubernetes secret
+    if (process.env.CONTAINER_PLATFORM === 'KUBERNETES') {
+        throw new Error('server_secret mounted to a file was not found. it must exist when running in kubernetes');
     }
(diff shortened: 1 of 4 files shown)


src/util/os_utils.js
```diff
@@ -591,16 +592,14 @@ function is_folder_permissions_set(current_path) {
...
-        // for all non kubernetes platforms (docker, local, etc.) return a dummy secret
-        return '12345678';
+    const server_secret = ssl_utils.get_data_from_kubernetes_secret(`/etc/noobaa-server`, `server_secret`);
+    if (server_secret) return server_secret;
+    // in kubernetes we must have SERVER_SECRET loaded from a kubernetes secret
+    if (process.env.CONTAINER_PLATFORM === 'KUBERNETES') {
+        throw new Error('server_secret mounted to a file was not found. it must exist when running in kubernetes');
     }
+    // for all non kubernetes platforms (docker, local, etc.) return a dummy secret
+    return '12345678';
```
shirady, line 602: Why do we send a different dummy secret?
`'12345678'` from `read_server_secret`, while `'abcdefgh'` from `get_jwt_secret()`

liranmauda: This is not in the scope of this PR, and I don't want to challenge it here.
This difference was probably created to diverse between server secret and jwt

liranmauda:
Converted to draft due to:
1. We need to get the latest yamls from the operator, so the test environments can work
2. We have one more secretKeyRef to address.

liranmauda:
> Converted to draft due to:
> 
> 1. We need to get the latest yamls from the operator, so the test environments can work
> 2. We have one more secretKeyRef to address.

### Regarding 1:
@dannyzaken should we take the latest yamls from the operator, or should we modify the ones in core? 
Do we have a specific diversity that we would like to keep, or it was just unmaintained? 

### Regarding 2:
It seems that for Postgress we need the env variable so we will keep the secretKeyRef

dannyzaken (approved):

PR #7158, opened 2023-01-09
Update es after updating node version

shirady:

### Explain the changes
1. Update es after updating node version to node 18 in eslint configuration and tsconfig file, see [here](https://node.green/).
2. Match numbers in the comments to the actual set value.

Related PRs:
#7123 
#7152

### Issues: 
Gap - ability to use most updated standards.

### Testing Instructions:
I only tested that change of header to be the current year. The tests I did appear in the [comment below](https://github.com/noobaa/noobaa-core/pull/7158#discussion_r1067775952).

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,16 +1,22 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const CURRENT_YEAR = new Date().getFullYear();
+
 module.exports = {
 
     // prevent from looking up other config files in containing folders
     root: true,
 
-    // our environment is node.js with it's es6 extensions
+    // Our environment is node.js
     env: {
         node: true,
-        es2020: true,
+        es2022: true,
     },
 
+    // See https://eslint.org/docs/latest/user-guide/configuring/language-options#specifying-environments
     parserOptions: {
-        ecmaVersion: 11, // 2020
+        ecmaVersion: 13, // es2022
     },
 
 
@@ -24,6 +30,7 @@ module.exports = {
     // we turn ON all of eslint rules by extending eslint:all
     // we selectively override the rules we want to remove or change.
     // See http://eslint.org/docs/user-guide/configuring#using-eslintall
+
     extends: 'eslint:all',
 
     rules: {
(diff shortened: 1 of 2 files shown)


liranmauda:
Regarding TSlint we might want to migrate it to ESLint:
npm WARN deprecated tslint@6.1.3: TSLint has been deprecated in favor of ESLint. Please see https://github.com/palantir/tslint/issues/4534 for more information.

.eslintrc.js
```diff
@@ -49,17 +54,18 @@ module.exports = {
 
         // verify every file contains our copyright header
         'header/header': ['error', 'block', [{
-            template: " Copyright (C) 2020 NooBaa ",
+            template: " Copyright (C) 2023 NooBaa ",
```
dannyzaken, line 57: @guymguym do we still need the Copyright template?

dannyzaken:
> Regarding TSlint we might want to migrate it to ESLint: npm WARN deprecated tslint@6.1.3: TSLint has been deprecated in favor of ESLint. Please see [palantir/tslint#4534](https://github.com/palantir/tslint/issues/4534) for more information.

@liranmauda anything we should do in this PR?

liranmauda (commented):
LGTM

shirady:
You see the push one after another because I'm checking something that nor related to this PR.

.eslintrc.js
guymguym: @dannyzaken I don't know. We can check with an expert if you want to change. But I think you can safely use a regexp to allow any 20** year...

guymguym: https://www.npmjs.com/package/eslint-plugin-header

> When you use a regular expression pattern, you can also provide a template property, to provide the comment value when using eslint --fix:
```
{
    "plugins": [
        "header"
    ],
    "rules": {
        "header/header": [2, "block", [
            {"pattern": " Copyright \\d{4}", "template": " Copyright 2019"}, 
            "My Company"
        ]]
    }
}
```

guymguym (approved):

.eslintrc.js
guymguym: @shirady Can we really just change it like that to 2023? it should make all the existing headers wrong and make eslint error per file... Isn't eslint failing for you now??

guymguym (changes requested):
see comment on eslint header ^

.eslintrc.js
shirady: I change it to be automatic to the current year.

**No header case** - 2 ways to check it:
In VS COde -> create new js file and save it -> Problems "missing header" eslint(header/header) -> Fix this header/header problem -> first line will appear:` /* Copyright (C) 2023 NooBaa */`.
Install eslint globally (brew install eslint)-> `eslint --fix <js file path>` -> first line will appear:` /* Copyright (C) 2023 NooBaa */`.

**Header with the year less than 2023** - I also saved an existing file and run `eslint --fix <js file path>` to see that the year is not changed. for example: `s3rver.js`.

**Illegal header** - added an header` /* Copyright (C) zzz NooBaa */` -> Problems: "incorrenct header eslint(header/header) -> Fix this header/header problem -> first line will appear:` /* Copyright (C) 2023 NooBaa */`.
same result from running `eslint --fix <js file path>`.

guymguym: oh sorry I didn't see the full diff and missed that we already have a pattern.
in any case, looks good now.

guymguym (approved):

PR #7161, opened 2023-01-12
Ceph s3 Tests | Update Ceph S3 Tests Commit Number

shirady:

### Explain the changes
1. Update the commit number of the repo s3-tests by Ceph.
 As a result the number of the passed tests will be 387 instead of 386, the test `test_object_copy_16m` was [added](https://github.com/ceph/s3-tests/pull/473).
2. Add to the logs the printing of the number of workers. 
Note: Currently it is with 1
3. Change the action to run on push or pull request of code inside `src` directory on `master` branch (in addition to the option to run it manually).
4. Increase the memory of db pod.
5. Add printings of k8s events in the logs.
6. Change the max days from 500 to 180.

### Testing Instructions:
1. Based on the instructions [here](https://gist.github.com/dannyzaken/c85f0006e5d6582c8d8666cafce6ab76) - the steps of ‘Build images’ and ‘Deploy noobaa’.
2. I’ve noticed that after installing noobaa we need to wait before running the job, otherwise all the tests will be failed.
`kubectl wait --for=condition=available backingstore/noobaa-default-backing-store --timeout=3m`
wait a couple of minutes after this condition is met ~3m (it's just a bypass).
You can use this command to check when the default backing store is in mode OPTIMAL:
`nb api system read_system`
3. Deploy the tests job (run it from the noobaa-core repo):
` kubectl apply -f src/test/system_tests/ceph_s3_tests/test_ceph_s3_job.yml`
View the logs:
`kubectl logs job/noobaa-tests-s3 -f`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/ceph-s3-tests.yaml
+++ b/.github/workflows/ceph-s3-tests.yaml
@@ -1,8 +1,5 @@
-name: ceph-s3-tests
-on: 
-  workflow_dispatch:
-  schedule:
-    - cron: "00 22,1 * * *" # Runs every day at 22:00 and 1:00 UTC
+name: Ceph S3 Tests
+on: [push, pull_request, workflow_dispatch]
 
 jobs:
   ceph-s3-tests:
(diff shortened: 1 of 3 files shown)


dannyzaken (approved):

liranmauda (commented):
@shirady @dannyzaken In `test_ceph_s3_deploy.sh` maybe we can change the `max_days="500"` to be 180 again

shirady:
> @shirady @dannyzaken In `test_ceph_s3_deploy.sh` maybe we can change the `max_days="500"` to be 180 again

I don't mind changing it to 180.
@romayalon I saw that you changed it in #6710, what do you think?

romayalon:
@shirady 
The last commit in https://github.com/ceph/s3-tests was more than a year ago, so this condition is not really relevant anymore, why not just take the latest commit?

liranmauda (approved):

PR #7163, opened 2023-01-12
Fix the replication scanner pagination support and add a fitting unittest

Neon-White:
### Explain the changes
1. The replication scanner's pagination logic was bugged since it tried to access a rule's continuation token through the hierarchy `rule.status...` instead of `rule.rule_status...`. The PR changes the logic to use the correct hierarchy
2. Also corrected the name of the `dst_cont_token`
3. A new unittest verifies that the fix works and will also keep future regressions in check. The unittest works by writing 1100 objects to the bucket, and running the scanner twice.
4. `list_objects` was removed and replaced with `list_all_objs_in_bucket`. The original function had no pagination support, and would only list the first 1k objects. The new one supports pagination and provides an array holding data about *all* objects in a given bucket.

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/test/unit_tests/test_bucket_replication.js
+++ b/src/test/unit_tests/test_bucket_replication.js
@@ -798,29 +798,126 @@ async function put_object(s3_owner, bucket_name, key, optional_body) {
     console.log('put_object: ', util.inspect(res));
 }
 
-async function list_objects(s3_owner, bucket_name) {
-    const res = await s3_owner.listObjects({ Bucket: bucket_name }).promise();
-    console.log('list_objects: ', util.inspect(res));
-    return res;
-}
 
 async function delete_object(s3_owner, bucket_name, key) {
     const res = await s3_owner.deleteObject({ Bucket: bucket_name, Key: key }).promise();
     console.log('delete_object: ', util.inspect(res));
 }
+
+// The function lists *all* objects in a given bucket, including buckets that have more than 1k objs
+async function list_all_objs_in_bucket(s3owner, bucket, prefix) {
+    let isTruncated = true;
+    let marker;
+    const elements = [];
+    while (isTruncated) {
+      let params = { Bucket: bucket };
+      if (prefix) params.Prefix = prefix;
+      if (marker) params.Marker = marker;
+        const response = await s3owner.listObjects(params).promise();
+        elements.push.apply(elements, response.Contents);
+        isTruncated = response.IsTruncated;
+        if (isTruncated) {
+          marker = response.Contents.slice(-1)[0].Key;
+        }
+    }
+    return elements;
+  }
+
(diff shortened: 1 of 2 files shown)


src/test/unit_tests/test_bucket_replication.js
```diff
@@ -798,29 +798,122 @@ async function put_object(s3_owner, bucket_name, key, optional_body) {
...
+
+// The function lists *all* objects in a given bucket, including buckets that have more than 1k objs
+async function list_all_objs_in_bucket(s3owner, bucket, prefix) {
+    let isTruncated = true;
+    let marker;
+    const elements = [];
+    while (isTruncated) {
+      let params = { Bucket: bucket };
+      if (prefix) params.Prefix = prefix;
+      if (marker) params.Marker = marker;
```
shirady, line 815: I think we will never enter this condition.
From what I understand: `marker` was initialized and not changed.

src/test/unit_tests/test_bucket_replication.js
```diff
@@ -798,29 +798,122 @@ async function put_object(s3_owner, bucket_name, key, optional_body) {
...
+        // create buckets
+        await P.all(_.map(buckets, async bucket_name => {
+            await rpc_client.bucket.create_bucket({ name: bucket_name });
+        }));
+        await put_replication(src_bucket, [
+            { rule_id: '1k-replication-rule', destination_bucket: target_bucket, filter: { prefix: '' } }
+        ], false);
+        const admin_account = await rpc_client.account.read_account({ email: EMAIL });
+        const admin_keys = admin_account.access_keys;
+        //await create_namespace_buckets(admin_account);
```
shirady, line 872: Would you please either uncomment it or remove it? (depends on what you need)

src/test/unit_tests/test_bucket_replication.js
```diff
@@ -798,29 +798,122 @@ async function put_object(s3_owner, bucket_name, key, optional_body) {
...
+        s3_creds.accessKeyId = admin_keys[0].access_key.unwrap();
+        s3_creds.secretAccessKey = admin_keys[0].secret_key.unwrap();
+        s3_creds.endpoint = coretest.get_http_address();
+        s3_owner = new AWS.S3(s3_creds);
+
+        // populate source bucket
+        for (let i = 0; i < obj_amount; i++) {
+            let key = create_random_body();
+            src_bucket_keys.push(key);
+            await put_object(s3_owner, src_bucket, key);
```
dannyzaken, line 883: How long does it take to run it with 1100 objects? Can we do it by using a smaller number of objects and changing `process.env.REPLICATION_MAX_KEYS`?

dannyzaken (approved):

PR #7167, opened 2023-01-17
Bug Fix #7166 | Change message of error in PG client

### LINKED ISSUE #7166: Account creation give error while noobaa-db pod running node is power down
### Environment info
- ODF 4.9 and 4.10
- Platform: OCP 4.9.31

### Actual behavior
1. account create fails with in-appropriate message when noobaa-db pod is in Init state
```
error: "this.begin() must be called before sending queries on this transaction' 
```
### Expected behavior
1.  account create fails should give Appropriate message when noobaa-db pod is in Init state. like:
```
Cannot connect to noobaa-db pod
```

shirady:

### Explain the changes
1. Change the error message in case the query failed (command `query`).
2. Wrap the action `client.pool.connect` with an exception (i.e. - try... catch clause).
3.  Add a flag `should_rollback` to prevent rollback call in case the DB was not connected.

### Issues:
1. Although the issue was opened with a request to change the message to "Cannot connect to noobaa-db pod", this is still a fix to this issue ([link to the issue](https://github.com/noobaa/noobaa-core/issues/7166)).
_Note: this link was not added to the title of issues because we didn't want it to automatically close the issue - we changed the label of it._

### Testing Instructions:
1. Based on the instructions [here](https://gist.github.com/dannyzaken/c85f0006e5d6582c8d8666cafce6ab76) - the steps of ‘Build images’ and ‘Deploy noobaa’.
3. use the watch to see all 5 pods running (and observe the changes that we will do).
`watch kubectl get pods`
4. kill `noobaa-db-pg` pod and create a new account
```
kubectl delete pod noobaa-db-pg-0 && nb api account_api create_account '{
  "email": "jenia5@noobaa.io",
  "name": "jenia5",
  "has_login": false,
  "s3_access": true,
  "default_resource": "noobaa-default-backing-store"
}'
```

### Attached Screenshots:
Previous message:
Screenshot 2023-01-17 at 13 26 49

New message:
Screenshot 2023-01-23 at 14 31 13

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -742,7 +742,7 @@ class SystemStore extends EventEmitter {
         for (const res of bulk_results) {
             if (res && !res.ok) {
                 dbg.error('got error on bulk execute', res.err);
-                // should we throw here? retry?
+                throw new Error(res.err);
             }
         }
 
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -28,6 +28,7 @@ const SensitiveString = require('./sensitive_string');
 const time_utils = require('./time_utils');
 const config = require('../../config');
 
+const DB_CONNECT_ERROR_MESSAGE = 'Could not acquire client from DB connection pool';
 mongodb.Binary.prototype[util.inspect.custom] = function custom_inspect_binary() {
     return `<mongodb.Binary ${this.buffer.toString('base64')} >`;
 };
@@ -330,7 +331,12 @@ class PgTransaction {
     }
 
     async begin() {
-        this.pg_client = await this.client.pool.connect();
+        try {
+            this.pg_client = await this.client.pool.connect();
+        } catch (err) {
+            dbg.error(DB_CONNECT_ERROR_MESSAGE, err);
+            throw new Error(DB_CONNECT_ERROR_MESSAGE);
+        }
         this.pg_client.once('error', err => {
             dbg.error('got error on pg_transaction', err, this.transaction_id);
         });
@@ -339,7 +345,7 @@ class PgTransaction {
 
     async query(text, values) {
         if (!this.pg_client) {
-            throw new Error('this.begin() must be called before sending queries on this transaction');
+            throw new Error('pg_client does not exist');
         }
         return _do_query(this.pg_client, { text, values }, this.transaction_id);
     }
@@ -396,8 +402,10 @@ class BulkOp {
         let nMatched = 0;
         let nModified = 0;
         let nRemoved = 0;
+        let should_rollback = false;
         try {
             await this.transaction.begin();
+            should_rollback = true;
             const batch_query = this.queries.join('; ');
             let results = await this.transaction.query(batch_query);
             if (!Array.isArray(results)) {
@@ -418,7 +426,7 @@ class BulkOp {
         } catch (err) {
             errmsg = err;
             dbg.error('PgTransaction execute error', err);
-            await this.transaction.rollback();
+            if (should_rollback) await this.transaction.rollback();
         } finally {
             this.transaction.release();
         }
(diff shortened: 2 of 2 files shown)


src/util/postgres_client.js
```diff
@@ -91,7 +92,7 @@ function encode_json(schema, val) {
     if (!val || !schema) {
         return val;
     }
-    // eslint-disable-next-line no-use-before-define
```
dannyzaken, line 94: removing this line is causing eslint errors, at least in my env. I'm not sure why it is not failing eslint in the tests

shirady: Hi Danny,
I run `eslint` on the file and I don't get an error.
Just to be sure `eslint` works I changed the header (2016 -> zzz) and saved and got an error.
Change it back - and nothing...
Screenshot 2023-01-19 at 8 23 18

shirady: 1. `eslint` is running in the required checks and it passed (runs `npm run lint`), as mentioned above and in the CI/CD meeting.
2. After fetching the most updated code, including #7158  this was not an error on Danny's machine as well.

shirady:
Moved to draft PR because we think about adding major changes and the above screenshot with the future printing is not a better solution.

Suggested changes related to PG client DB (not in the current code changes):
1.  The origin of the printings is after a failure to publish to cluster of an RPC call.
2. Today we mock MongoDb behavior and we use `bulk_per_collection` @dannyzaken  suggested that we would want to change it since we can in PG use bulk for all collections.
3. In case it helps, here is the stack from the logs (noobaa-core pod after deleting the DB pod + creating a new account as mentioned in the description of the PR):
Screenshot 2023-01-19 at 16 21 23

src/server/system_services/system_store.js
```diff
@@ -742,7 +742,7 @@ class SystemStore extends EventEmitter {
         for (const res of bulk_results) {
             if (res && !res.ok) {
                 dbg.error('got error on bulk execute', res.err);
-                // should we throw here? retry?
+                throw new Error(res.err);
```
shirady, line 745: @dannyzaken @guymguym What do you think? (please also comment in the PR)

src/util/postgres_client.js
```diff
@@ -396,9 +402,11 @@ class BulkOp {
         let nMatched = 0;
         let nModified = 0;
         let nRemoved = 0;
+        let should_rollback = false;
         try {
             await this.transaction.begin();
             const batch_query = this.queries.join('; ');
+            should_rollback = true;
```
romayalon, line 409: Wondering about this change,
this.transaction.begin() includes -
1. initialization of this.pg_client 
2. running the first query of the transaction - BEGIN TRANSACTION

if something failed for example in this.queries.join, shouldn't we call rollback?
I know nothing happened but indeed a transaction started in the db, do we need it as a cleanup?

src/server/system_services/system_store.js
dannyzaken: for now, I think we should throw here. 
later on, we should change the behavior to perform all bulk changes in the system_store in a single transaction (and not per collection)

romayalon:
@rkomandu Is the new error message suggested by these changes informative enough?

src/util/postgres_client.js
shirady: @Danny with this change we still have the `BEGIN TRANSACTION` in our DB and we will not rollback it - from what I understand that @romayalon wrote.

romayalon: @shirady yes but after reading a bit I think that rollback won't clean the started transaction and actually we can keep committing updates to the transaction after rollbacking, so I guess this is okay.

romayalon (approved):

dannyzaken (approved):

shirady:
The flow was: RPC request -> create account (account server) -> make changes (system store) -> ... -> make changes internal (system store) -> execute (transaction error - this was failing since the db pod was down) -> rollback + query (which we don't need because we didn't execute and failed in the begin function since the db pod was down).

PR #7168, opened 2023-01-18
NSFS list_objects should only list latest objects

### LINKED ISSUE #7164: NSFS list_objects() should only list latest objects
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. list_objects, returns every object under the bucket. It should ignore objects under `.versions` where we store the versions of the object.

### Expected behavior
1. list_objects() should only list latest objects not all objects (versions)

### Steps to reproduce
1. Create a object and create versions of it
2. do list_objects(), that returns objects under `.versions` as well

vh05:
### Explain the changes
list_objects, returns every object under the bucket. It should ignore objects under .versions where we
store the versions of the object.

### Testing Instructions:
1. Create a object and create versions of it
2. do list_objects(), that should not return objects under `.versions`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -337,6 +337,9 @@ class NamespaceFS {
              * @returns {Promise<void>}
              */
             const process_dir = async dir_key => {
+                if (this._skip_version_dir(dir_key)) {
+                    return;
+                }
 
                 // /** @type {fs.Dir} */
                 let dir_handle;
@@ -2031,6 +2034,11 @@ class NamespaceFS {
         } catch (err) {
             dbg.warn('namespace_fs.find_max_version_past: .versions/ folder could not be found', err);
        }
+    }
+
+    _skip_version_dir(dir_key) {
+        const idx = dir_key.indexOf(HIDDEN_VERSIONS_PATH);
+        return ((idx === 0) || (idx > 0 && dir_key[idx - 1] === '/'));
     }
 }
 
--- a/src/test/unit_tests/test_namespace_fs.js
+++ b/src/test/unit_tests/test_namespace_fs.js
@@ -1,5 +1,5 @@
 /* Copyright (C) 2020 NooBaa */
-/*eslint max-lines-per-function: ["error", 700]*/
+/*eslint max-lines-per-function: ["error", 900]*/
 'use strict';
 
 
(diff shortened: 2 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1761,17 @@ class NamespaceFS {
             throw new RpcError('BAD_REQUEST', 'Bad Request');
         }
     }
+
+    _skip_version_dir(dir_key) {
+        const parts = dir_key.split('/');
+        let skip = false;
```
romayalon, line 1767: I think you can just return dir_key.indexOf(`/.${HIDDEN_VERSIONS_PATH}/`) > 0, no?

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -232,6 +232,113 @@ mocha.describe('namespace_fs', function() {
...
+            await create_object(`${full_path}/${key}`, body, 'null');
+            await create_object(`${full_path_version_dir}/${version_key}`, version_body, 'null');
+            await create_object(`${dir1}/${key}`, body, 'null');
+            await create_object(`${dir1_version_dir}/${version_key}`, version_body, 'null');
+            await create_object(`${dir2}/${key}`, body, 'null');
+            await create_object(`${dir1_version_dir}/${version_key}`, version_body, 'null');
+        });
+
+        mocha.after(async () => {
+            //fs_utils.folder_delete(tmp_fs_root);
```
romayalon, line 326: please add it back for cleaning, can you please uncomment it in other tests in this file as well?

src/sdk/namespace_fs.js
v-harihar: Yeah sure. Need to be bit used to these methods :)

src/test/unit_tests/test_namespace_fs.js
v-harihar: that went unnoticed. Sure will do that

src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1761,13 @@ class NamespaceFS {
             throw new RpcError('BAD_REQUEST', 'Bad Request');
         }
     }
+
+    _skip_version_dir(dir_key) {
+        if(dir_key.indexOf(`${HIDDEN_VERSIONS_PATH}`) >= 0) {
```
romayalon, line 1766: we expect at least 1 char before that so I think the condition should be > 0

v-harihar: For `.versions` under bucket path will have 0 index. Ex: for  `$BUCKET_PATH/.versions` , the `dir_key` would be `.versions` that will have 0 index.

correct me if wrong

romayalon: ah ok coool

src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1761,10 @@ class NamespaceFS {
             throw new RpcError('BAD_REQUEST', 'Bad Request');
         }
     }
+
+    _skip_version_dir(dir_key) {
+        return dir_key.indexOf(`${HIDDEN_VERSIONS_PATH}`) >= 0;
```
guymguym, line 1766: seems like the string interpolation (`${x}`) is unneeded 
```suggestion
        return dir_key.indexOf(HIDDEN_VERSIONS_PATH) >= 0;
```

romayalon: shouldn't we search for /.versions/ ? instead of .versions

v-harihar: Yes. In the previous code, the file parts were divided with `/` delimeter and here it should be `/.versions/`. I will correct it

romayalon: Let's take for example the following dir:
dir_key = a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/.versions
Definitions: 
n = dir_key.length
l = number of '/'
m = avg number of chars of every subdir
n = l + (l+1)*m

Option1 -
1. Splitting by '/' is 1 loop - O(n) - while n is the number of chars in dir_key.
2. Iterating the result array and comparing char by char of each item in the array O((l+1)*m) while m is the average number of chars of each item in the array and l is the number of '/' in the string = the number of items in the split result array.
Total complexity time - O(n+(l+1)*m)

Option 2 - 
1. searching char by char twice will take - O(2n)

So the question is (l+1)*m < n, the answer is yes and your assumption is correct, the difference is exactly l because we said that n = l + (l+1)*m
I think that it's a very slight improvement, I want to believe that l is not a big number in most of the cases unless I did a mistake in the calculation/ the algorithm of option 1, please correct me if you see a mistake.

Anyway, there is another optimization you can do for option 2 in order to iterate the string once - for example:
```
const idx = dir_key.indexOf(HIDDEN_VERSIONS_PATH);
return idx === 0 || idx > 0 && dir_key[idx-1] === ‘/’;
```
then the optimized version of option 2 is O(n).

v-harihar: This is okay :)

aspandey: @v-harihar do we have this version path like "dir_key = a/b/c/d/e/f/g/h/i/j/k/l/m/n/o/p/.versions" only?

If yes, why can't we first split this string, with "/" as delimiter, and then just compare last element with ".versions"?

I mean to say is that if we have any idea of the position of ".versions" then compare that only.

I know you would have thought of this but just mentioning.

romayalon: So Vinayak is saying that dir_key can be exactly `.versions` (without '/') or `a/b/.versions` for example
I thought that we can use dir_key.endsWith('.versions') but still we need to check for the / if the index > 0 (that's why I wrote indexOf and not endsWith) / compare to .versions or search for dir_key.lastIndexOf('/') and then slice and compare to .versions

v-harihar: @aspandey Yes logic wise both `split & compare` and the `indexOf() & check index (the added logic)` would work fine.  Also process cycle wise both results same. We had that logic earlier in this patch and overwritten because of append push which I should not have done :)

src/sdk/namespace_fs.js
```diff
@@ -1759,6 +1761,11 @@ class NamespaceFS {
             throw new RpcError('BAD_REQUEST', 'Bad Request');
         }
     }
+
+    _skip_version_dir(dir_key) {
+        const idx = dir_key.indexOf(HIDDEN_VERSIONS_PATH)
```
romayalon, line 1766: missing ;

dannyzaken (approved):

PR #7170, opened 2023-01-19
Ceph s3 Tests | Collect logs in case of failure

shirady:
### Explain the changes
1. Move the command to print the k8s events to run only in case of failure.
2. Create a new step to collect logs in case of failure and collect noobaa diagnostics and db dump.

### Issues: Gap
1. Today we print the k8s events on every run, although we need it only if the runner failed.
2. We saw runs that after noobaa install the backing store pod is not created yet ([example1](https://github.com/noobaa/noobaa-core/actions/runs/3939866458/jobs/6740234105), [example2](https://github.com/noobaa/noobaa-core/actions/runs/3952806933/jobs/6768330683)).
Today we don't have the tools to investigate it, and we would like to check this case specifically and use the logs in general as needed.
3. Freeze the version of operator to the latest current commit, to avoid a failed run due to code changes in the operator repo.
4. Add a timeout of 90 minutes according to the CI/CD discussion today.

### Testing Instructions:
1. GH Action changes on the workflow in this push, tests will happen in every push.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/ceph-s3-tests.yaml
+++ b/.github/workflows/ceph-s3-tests.yaml
@@ -4,6 +4,7 @@ on: [push, pull_request, workflow_dispatch]
 jobs:
   ceph-s3-tests:
     runs-on: ubuntu-latest
+    timeout-minutes: 90
     steps:
       - name: Checkout noobaa-core
         uses: actions/checkout@v3
@@ -27,6 +28,10 @@ jobs:
         with:
           repository: 'noobaa/noobaa-operator'
           path: 'noobaa-operator'
+         # Freeze the version of operator
+         # to avoid a failed run due to code changes in the operator repo.
+         # Need to update the commit once in a while
+          ref: 28960d4dbbba612d1a3e53245791076c11632b60
 
       - name: Change settings for k8s and minikube
         run: |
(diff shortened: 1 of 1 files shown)


shirady:
This PR is in draft because I'm testing it using pushes.
It also includes a commit that I will delete (change of pending list).

.github/workflows/ceph-s3-tests.yaml
```diff
@@ -77,3 +75,21 @@ jobs:
...
+         echo "K8S Events"
+         kubectl get events --sort-by='.metadata.creationTimestamp' -A
+         mkdir -p my_logs
+         cd ./noobaa-operator
+         ./build/_output/bin/noobaa-operator diagnose --db-dump --dir=my_logs
+         cd ../my_logs
+         pwd
+         ls
+      - name: Save logs
+        uses: actions/upload-artifact@v3
```
dannyzaken, line 92: shouldn't  `if: ${{ failure() }}` be used here too?

shirady: Should be, thanks

.github/workflows/ceph-s3-tests.yaml
```diff
@@ -27,6 +27,10 @@ jobs:
         with:
           repository: 'noobaa/noobaa-operator'
           path: 'noobaa-operator'
+         # Freeze the version of operator
+         # to avoid a failed run due to code changes in the operator repo.
+         # Need to update the commit once in a while
+          ref: 28960d4dbbba612d1a3e53245791076c11632b60
```
dannyzaken, line 33: Let's have that in a separate commit in this PR

dannyzaken (approved):

liranmauda (commented):
@shirady LGTM

PR #7172, opened 2023-01-19
[TEST] Testing suppression on draft PRs

nimrod-becker:
### Explain the changes
1. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/build-arm64-image.yaml
+++ b/.github/workflows/build-arm64-image.yaml
@@ -1,8 +1,9 @@
 name: Build arm64 image
-on: [ push, pull_request ]
+on: [ push, pull_request, pull_request_review ]
 
 jobs:
   build-arm64-image:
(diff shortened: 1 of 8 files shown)


shirady:
@nimrod-becker, Would you please change it to PR (instead of draft) so can see that it triggers the events of checks as well? 
I read about a [bug](https://github.com/orgs/community/discussions/25722#discussioncomment-3248909) on it, so I just want to check it before merging.
Then back to PR draft so the reviewers will see that the checks were skipped.

nimrod-becker:
@shirady not yet as I want to add labeling and review status, once all is there I will play around with moving the different states

dannyzaken:
@nimrod-becker anything to do with this PR?

PR #7177, opened 2023-01-19
NSFS | Versioning | GPFS Linkat & Unlinkat fd verification

romayalon:

### Explain the changes
2. Fs_napi.unlinkfileat() that calls GPFS unlinkfileat.
3. Fs_napi.getfd() an accessor for fd files.
4. Fs_napi.linkfileat() - added replace_fd and passed it to linkat() for fd verification.
5. GPFS.h - Added the new declarations for linkat & unlinkat
6. namespace_fs - Added safe_move() function that calls safe_move_posix() or safe_move_gpfs() according to the result of _is_gpfs() functions.
7. Namespace_fs.safe_move_gpfs() - 
   7.1. Wraps linkfileat() & unlinkfileat() (if requested) calls with retries
   7.2. open dir and file itself in order to unlink it and pass the needed fd
8. Namespace_fs.delete_object() - 
    8.1. delete version id & latest - called unlinkfileat with retries.
9. Namespace_fs.delete_multiple_objects() - 
    9.1. delete version id & latest - called unlinkfileat with retries.

1. optimization - delete version id - GPFS link overrides, no need to unlink before promoting, but if there is nothing to promote we should 
2. move_to_dest retries - checkaccess() when called from upload_object() will fail on temp file since it doesn't exist

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -125,6 +125,12 @@ static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg) = 0;
 static int (*dlsym_gpfs_linkat)(
     gpfs_file_t fileDesc, const char* oldpath, gpfs_file_t newdirfd, const char* newpath, int flags) = 0;
 
+static int (*dlsym_gpfs_linkatif)(
+    gpfs_file_t fileDesc, const char* oldpath, gpfs_file_t newdirfd, const char* newpath, int flags, gpfs_file_t replace_fd) = 0;
+
+static int (*dlsym_gpfs_unlinkat)(
+    gpfs_file_t fileDesc, const char* path, gpfs_file_t fd) = 0;
+
 const static std::map<std::string, int> flags_to_case = {
     { "r", O_RDONLY },
     { "rs", O_RDONLY | O_SYNC },
@@ -1145,8 +1151,10 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
                 InstanceMethod<&FileWrap::writev>("writev"),
                 InstanceMethod<&FileWrap::replacexattr>("replacexattr"),
                 InstanceMethod<&FileWrap::linkfileat>("linkfileat"),
+                InstanceMethod<&FileWrap::unlinkfileat>("unlinkfileat"),
                 InstanceMethod<&FileWrap::stat>("stat"),
                 InstanceMethod<&FileWrap::fsync>("fsync"),
+                InstanceAccessor<&FileWrap::getfd>("fd"),
             }));
         constructor.SuppressDestruct();
     }
@@ -1170,8 +1178,10 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
     Napi::Value writev(const Napi::CallbackInfo& info);
     Napi::Value replacexattr(const Napi::CallbackInfo& info);
     Napi::Value linkfileat(const Napi::CallbackInfo& info);
+    Napi::Value unlinkfileat(const Napi::CallbackInfo& info);
     Napi::Value stat(const Napi::CallbackInfo& info);
     Napi::Value fsync(const Napi::CallbackInfo& info);
+    Napi::Value getfd(const Napi::CallbackInfo& info);
 };
 
 Napi::FunctionReference FileWrap::constructor;
@@ -1372,20 +1382,49 @@ struct FileReplacexattr : public FSWrapWorker<FileWrap>
 struct LinkFileAt : public FSWrapWorker<FileWrap>
 {
     std::string _filepath;
-
+    int _replace_fd;
     LinkFileAt(const Napi::CallbackInfo& info)
         : FSWrapWorker<FileWrap>(info)
+        , _replace_fd(0)
     {
         _filepath = info[1].As<Napi::String>();
+        if (info.Length() > 2 && !info[2].IsUndefined()) {
+            _replace_fd = info[2].As<Napi::Number>();
+        }
         Begin(XSTR() << "LinkFileAt " << DVAL(_wrap->_path) << DVAL(_wrap->_fd) << DVAL(_filepath));
     }
     virtual void Work()
     {
         // gpfs_linkat() is the same as Linux linkat() but we need a new function because
         // Linux will fail the linkat() if the file already exist and we want to replace it if it existed.
-        SYSCALL_OR_RETURN(dlsym_gpfs_linkat(_wrap->_fd, "", AT_FDCWD, _filepath.c_str(), AT_EMPTY_PATH));
+        if (_replace_fd == 0){
+            SYSCALL_OR_RETURN(dlsym_gpfs_linkat(_wrap->_fd, "", AT_FDCWD, _filepath.c_str(), AT_EMPTY_PATH));
+        } else {
+            SYSCALL_OR_RETURN(dlsym_gpfs_linkatif(_wrap->_fd, "", AT_FDCWD, _filepath.c_str(), AT_EMPTY_PATH, _replace_fd));
+        }
     }
 };
+
+struct UnlinkFileAt : public FSWrapWorker<FileWrap>
+{
+    std::string _filepath;
+    int _delete_fd;
+    UnlinkFileAt(const Napi::CallbackInfo& info)
+        : FSWrapWorker<FileWrap>(info)
+        , _delete_fd(0)
+    {
+        _filepath = info[1].As<Napi::String>();
+        if (info.Length() > 2 && !info[2].IsUndefined()) {
+            _delete_fd = info[2].As<Napi::Number>();
+        }
+        Begin(XSTR() << "UnlinkFileAt" << DVAL(_wrap->_path.c_str()) << DVAL(_wrap->_fd) << DVAL(_filepath) << DVAL(_delete_fd));
+    }
+    virtual void Work()
+    {   
+        SYSCALL_OR_RETURN(dlsym_gpfs_unlinkat(_wrap->_fd, _filepath.c_str(), _delete_fd));
+    }
+};
+
 
 struct FileStat : public FSWrapWorker<FileWrap>
 {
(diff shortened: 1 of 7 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -1497,6 +1553,18 @@ FileWrap::linkfileat(const Napi::CallbackInfo& info)
...
+Napi::Value
+FileWrap::unlinkfileat(const Napi::CallbackInfo& info)
+{
+    return api<UnlinkFileAt>(info);
+}
+
+Napi::Value
+FileWrap::getfd(const Napi::CallbackInfo& info)
+{
+    return api<GetFd>(info);
```
guymguym, line 1565: you don't need to submit a worker to the threadpool to do this, FileWrap keeps the _fd and you can just return it synchronously here with `return Napi::Number::New(info.Env(), _fd);`

src/native/fs/gpfs.h
```diff
@@ -1215,10 +1219,35 @@ gpfs_stat(const char *pathname, /* File pathname */
...
  *              EINVAL  Not a GPFS file
  *              ESTALE  cached fs information was invalid
+ *              EEXIST  Destination file already exists and not the same as
+ *                      in replace fd
  *-------------------------------------------------------------------------*/
 int GPFS_API
 gpfs_linkat(gpfs_file_t olddirfd, const char *oldpath,
-            gpfs_file_t newdirfd, const char *newpath, int flags);
+            gpfs_file_t newdirfd, const char *newpath, int flags,
+            gpfs_file_t replacefd);
```
guymguym, line 1228: @eshelmarc @romayalon is there a versioning on this header? it would have to match the runtime library in order to work, right? should we make this dependency more explicit somewhere so that we will get a proper error (eg. libgpfs version mismatch...) vs just an error that a symbol was not found.

src/native/fs/fs_napi.cpp
```diff
@@ -1307,18 +1314,67 @@ struct FileReplacexattr : public FSWrapWorker<FileWrap>
...
+    }
+};
+
+struct UnlinkFileAt : public FSWrapWorker<FileWrap>
+{
+    std::string _filepath;
+    int _delete_fd;
+    UnlinkFileAt(const Napi::CallbackInfo& info)
+        : FSWrapWorker<FileWrap>(info)
+        , _delete_fd(0)
```
guymguym, line 1344: how about renaming to replace_fd like it is in libgpfs header?

also can you add a comment to document this fd argument - e.g that it's optional, and if undefined it's like 0 and mean what exactly - that any inode can be replaced? or that no file must be replaced - so it works like posix un/linkat? i am not sure which... and when fd is non 0 it will only verify the replaced inode matches before replacing it, but what about if there is no file there - does it proceed with un/linkat? or returns with error?

src/native/fs/gpfs.h
eshelmarc: The libgpfs code include a IFILE_MAGIC number that GPFS main code will verify.
Also GPFS development did not want to change en existing interface, namely gpfs_linkat() so we now have
**gpfs_linkat**(gpfs_file_t olddirfd, const char *oldpath, gpfs_file_t newdirfd, const char *newpath, int flags);
when you don't have a second fd, and
**gpfs_linkatif**(gpfs_file_t olddirfd, const char *oldpath, gpfs_file_t newdirfd, const char *newpath, int flags, gpfs_file_t replacefd);
when you do have a 2nd fd

src/native/fs/fs_napi.cpp
romayalon: Cool thanks, it's a temporary change, I wish to remove it when unlinkat will support AT_FDCWD flag

src/native/fs/fs_napi.cpp
romayalon: Sure I can change it but in unlinkat() the verified fd is actually called _fd (the verified fd called _replace_fd only in linkat) so should I use replace_fd term here also?

src/native/fs/gpfs.h
eshelmarc: no

src/native/fs/gpfs.h
```diff
@@ -1215,10 +1219,35 @@ gpfs_stat(const char *pathname, /* File pathname */
  *              EBADF   Bad file desc
  *              EINVAL  Not a GPFS file
  *              ESTALE  cached fs information was invalid
+ *              EEXIST  Destination file already exists and not the same as
+ *                      in replace fd
  *-------------------------------------------------------------------------*/
 int GPFS_API
 gpfs_linkat(gpfs_file_t olddirfd, const char *oldpath,
-            gpfs_file_t newdirfd, const char *newpath, int flags);
+            gpfs_file_t newdirfd, const char *newpath, int flags,
```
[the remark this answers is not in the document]
romayalon: updated

src/sdk/namespace_fs.js
```diff
@@ -969,24 +971,20 @@ class NamespaceFS {
     //    2.1 create .versions/ if it doesnt exist
     //    2.2 move old version (existing latest) to .versions/
     // 3. move new version to dest (key path)
-    async _move_to_dest_ver_enabled(fs_context, source_path, dest_path, { key }) {
-
+    async _move_to_dest_ver_enabled(fs_context, source_path, dest_path, target_file, { key }) {
         const cur_ver_info = await this._get_version_info(fs_context, dest_path);
```
jackyalbo, line 975: as we discussed it seems we need to retry on this whole part and not just in safe_move

src/sdk/namespace_fs.js
```diff
@@ -2053,19 +2043,9 @@ class NamespaceFS {
...
-                fs_context,
-                async ({ mtimeNsBigint, ino }) => {
-                    const unique_temp_path = await this._generate_unique_path(fs_context);
-                    await nb_native().fs.safe_unlink(fs_context, file_path, unique_temp_path, mtimeNsBigint, ino);
-                },
-                Object.assign(version_info, { from_path: file_path }),
-                'FS::SafeUnlink ERROR unlink target doesn\'t match expected inode and mtime',
-                ['ENOENT']
-            );
+            // TODO optimization - GPFS link overrides, no need to unlink before promoting, but if there is nothing to promote we should unlink
```
jackyalbo, line 2047: change comment 3 of the function to promote only if version_id is latest

src/sdk/namespace_fs.js
```diff
@@ -1989,58 +1982,55 @@ class NamespaceFS {
...
-                        latest_ver_info = version_info;
+        let latest_dir_file;
+        try {
+            const res = [];
+            let deleted_delete_marker = false;
+            let delete_marker_created = false;
+            let latest_ver_info;
+            const latest_version_path = this._get_file_path({ key });
+            await this._check_path_in_bucket_boundaries(fs_context, latest_version_path);
+            latest_dir_file = this._is_gpfs(fs_context) && (await this._open_file(fs_context, path.dirname(latest_version_path), 'r'));
```
jackyalbo, line 1993: move under delete_latest if possible

src/test/unit_tests/test_nsfs_versioning_gpfs.js
```diff
@@ -0,0 +1,323 @@
...
+// }
+
+// function get_version_id_by_stat(stat) {
+//     return 'mtime-' + stat.mtimeNsBigint.toString(36) + '-ino-' + stat.ino.toString(36);
+// }
+
+function get_version_id_by_xattr(stat, prev) {
+    if (prev) return stat && stat.xattr[XATTR_PREV_VERSION_ID];
+    return (stat && stat.xattr[XATTR_VERSION_ID]) || 'null';
+}
```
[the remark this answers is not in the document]
romayalon: done

src/sdk/namespace_fs.js
```diff
@@ -916,29 +916,27 @@ class NamespaceFS {
...
-        await this._wrap_safe_op_with_retries(
-            fs_context,
-            async ({ mtimeNsBigint, ino }) => {
-                const unique_temp_path = await this._generate_unique_path(fs_context);
-                await nb_native().fs.safe_unlink(fs_context, from_path, unique_temp_path, mtimeNsBigint, ino);
-            },
-            Object.assign(version_info || {}, { from_path }),
-            'FS::SafeUnlink ERROR unlink target doesn\'t match expected inode and mtime'
-        );
+    async safe_move_gpfs(fs_context, source_path, dest_path, dst_ver_info, upload_file, should_unlink) {
```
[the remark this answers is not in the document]
romayalon: done

src/sdk/namespace_fs.js
```diff
@@ -2208,6 +2191,110 @@ class NamespaceFS {
...
+    async safe_link_gpfs(fs_context, source_path, dest_path, dst_ver_info, upload_file) {
+        let source_file;
+        try {
+            source_file = upload_file || (await this._open_file(fs_context, source_path, 'r'));
+            dbg.log0('Namespace_fs.safe_move_gpfs source_file:', upload_file, source_file, source_file && source_file.fd);
+
+            let dest_file;
+            await this._wrap_safe_op_with_retries(
+                fs_context,
+                async ({ to_version_info }) => {
```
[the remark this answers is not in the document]
romayalon: removed it completely

src/sdk/namespace_fs.js
romayalon: done

src/sdk/namespace_fs.js
```diff
@@ -2130,7 +2112,8 @@ class NamespaceFS {
         if (version_info) {
             to_delete_version_path = this._get_version_path(params.key, version_info.version_id_str);
```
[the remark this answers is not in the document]
romayalon: done

src/sdk/namespace_fs.js
romayalon: removed it completely

src/sdk/namespace_fs.js
```diff
@@ -2208,6 +2191,110 @@ class NamespaceFS {
...
+                    if (file) await file.close(fs_context);
+                }
+            };
+            const gpfs_retry_err_msg = 'EEXIST';
+
+            // retry safe unlinking a file in case of parallel put/delete of the source path
+            await this._wrap_safe_op_with_retries(
+                fs_context,
+                gpfs_unlink_handler,
+                Object.assign(version_info || {}, { path_to_delete }),
```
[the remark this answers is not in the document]
romayalon: removed

jackyalbo (approved):

PR #7178, opened 2023-01-23
Changes to all GitHub workflows - timeout, checkout v3, concurrency, general cleanup

shirady:

### Explain the changes
According to _CI next steps_ meeting:
1. Change default timeouts for our tests to 1.5h (90 minutes).
2. Change all checkout to v3 (v2 is using node version 12).
3. Use GH Concurrency to cancel the previous run.
This was a pilot in #7159 , but after we started using it, we [discovered](https://github.com/styfle/cancel-workflow-action/issues/200) that there is a builtin option from GitHub (concurrency).

General cleanup (was not in the meeting): 
1. Remove `exit 1` since the default behavior is when a step is failing to fail the whole job.
2. Remove extra newlines.
3. Remove comments that explain the code itself.
4. Remove unused `id` fields in the steps.

### Issues: Gap
The main gap is that the default timeout of jobs is 6 hours and it is a waste of computing resources since all of our jobs run in a sixth of the time.  Another waste of computing resources is when we use successive pushes, and we would like to reduce it.

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/build-arm64-image.yaml
+++ b/.github/workflows/build-arm64-image.yaml
@@ -4,6 +4,10 @@ on: [ push, pull_request ]
 jobs:
   build-arm64-image:
     runs-on: ubuntu-22.04
+    timeout-minutes: 90
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true
     env:
       # For multiplatform use list like:
       # "linux/amd64,linux/arm64"
@@ -11,7 +15,7 @@ jobs:
       GIT_COMMIT: ${{ github.sha }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           ref: ${{ github.event.inputs.branch }}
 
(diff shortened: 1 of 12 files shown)


nimrod-becker (approved):

liranmauda (commented):
@shirady The reason that we added `|| exit 1` was that sometimes steps failed and the tests were green. In this case, we will not be aware that the test failed as we are not looking into green tests (which makes sense).  
If we want to remove the `|| exit 1` we need to be sure that failing steps will lead to red tests (failing GH Action tests).

shirady:
> @shirady The reason that we added `|| exit 1` was that sometimes steps failed and the tests were green. In this case, we will not be aware that the test failed as we are not looking into green tests (which makes sense). If we want to remove the `|| exit 1` we need to be sure that failing steps will lead to red tests (failing GH Action tests).

Hi Liran,
From previous trial and error, we know that if a step fails then all job will fail.
Anyway, I did a test: I add an [exit 1 ](https://github.com/shirady/noobaa-core-copy-for-test/commit/8b643b7cdc48e44b9b2158e4c98bf70e11e31d6e)to the file _run_npm_test_on_test_container.sh_ and the job [failed](https://github.com/shirady/noobaa-core-copy-for-test/actions/runs/3984996544/jobs/6831840980).
_Note: it's a copy of noobaa-core, this link will not be available in the future._

liranmauda (approved):
LGTM

PR #7183, opened 2023-01-26
Fix custom metadata on namespaced Azure blob backend

t-pirhonen:
### Explain the changes
1. Replace '-' with '_' when writing custom metadata headers (x-amz-meta) to a azure blob backend
2. When getting a blob replace '_' with '-' 

1. fixes retrieving S3 objects from a blob backend where custom metadata contains '_'
2. Note that this would convert the '_' back to '-' even when there were no '-' in originally
 
### Testing Instructions:
2. upload an object with custom metadata containing a '_' in the key

### DIFF
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -207,19 +207,19 @@ class NamespaceBlob {
                 if (!res.readableStreamBody) throw new Error('NamespaceBlob.read_object_stream: download response is invalid');
                 return resolve(res.readableStreamBody.pipe(count_stream));
             }).catch(err => {
-                    this._translate_error_code(err);
-                    dbg.warn('NamespaceBlob.read_object_stream:',
-                        this.container,
-                        inspect(_.omit(params, 'object_md.ns')),
-                        'callback err', inspect(err)
-                    );
-                    try {
-                        count_stream.emit('error', err);
-                    } catch (err2) {
-                        // ignore, only needed if there is no error listeners
-                    }
-                    return reject(err);
+                this._translate_error_code(err);
+                dbg.warn('NamespaceBlob.read_object_stream:',
+                    this.container,
+                    inspect(_.omit(params, 'object_md.ns')),
+                    'callback err', inspect(err)
+                );
+                try {
+                    count_stream.emit('error', err);
+                } catch (err2) {
+                    // ignore, only needed if there is no error listeners
                 }
+                return reject(err);
+            }
 
             );
         });
@@ -270,8 +270,8 @@ class NamespaceBlob {
                     new_stream.pipe(count_stream),
                     params.size,
                     undefined, {
-                        metadata: params.xattr,
-                        blobHTTPHeaders: headers
+                    metadata: params.xattr,
+                    blobHTTPHeaders: headers
                     }
                 );
                 obj.contentMD5 = md5_buf();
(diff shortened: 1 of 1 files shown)


lallinger-arbeit:
Hey @liranmauda,
it's been quite some time but hopefully i'm finally able to get our pull requests upstream with the help of Tuomas (@t-pirhonen).
If there is anything missing for the pull request to be merged, please let us know! We really want to get our (mostly minor) fixes upstream!

src/sdk/namespace_blob.js
```diff
@@ -395,6 +395,16 @@ class NamespaceBlob {
...
     _check_valid_xattr(params) {
+
+        const xattr = params.xattr;
+        for (const xattrKey in xattr) {
+            const tmp = xattrKey.replace(/-/g, '_');
+            const tmp2 = xattr[xattrKey];
+            delete xattr[xattrKey];
+            xattr[tmp] = tmp2;
+        }
+        params.xattr = xattr;
```
liranmauda, line 406: Will this work instead? 
```suggestion
        for (const xattrKey in params.xattr) {
            params.xattr[xattrKey] = params.xattr[xattrKey].replace(/-/g, '_');
        }
```

liranmauda:
Hi @lallinger-arbeit and @t-pirhonen 
Thanks for this PR 🙂 
For the lit error please read: https://eslint.org/docs/latest/rules/guard-for-in 
for the DCO you can run: `git rebase HEAD~1 --signoff`
Can you please test the loop I wrote? 
Thanks.

src/sdk/namespace_blob.js
liranmauda: maybe even:
```suggestion
        for (const [key, val] of params.xattr) {
            params.xattr[key] = val.replace(/-/g, '_');
        }
```

nimrod-becker:
Looking at azCopy, I can see that Azure provides 3 options for invalid MD keys
Fail the entire op, skip the key or rename (see https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-s3#handle-differences-in-object-metadata)

Then rename conventions are here (https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-s3#how-azcopy-renames-object-keys)

I don't think the rename conventions are good, but I do like the option to give a way for the client to choose what to do.

Would like to add a custom header with a directive of skip, fail, rename and we will act accordingly

nimrod-becker:
@lallinger-arbeit 
If you want to pick this one up (according to my latest comment) that's fine, we would also love to pick it up and fix it ourselves, so waiting on your response

lallinger-arbeit:
Hey @nimrod-becker,

> Looking at azCopy, I can see that Azure provides 3 options for invalid MD keys Fail the entire op, skip the key or rename (see https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-s3#handle-differences-in-object-metadata)
> 
> Then rename conventions are here (https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-s3#how-azcopy-renames-object-keys)
> 
> I don't think the rename conventions are good, but I do like the option to give a way for the client to choose what to do.

Oh, didn't know azCopy already adresses this problem. What do you not like about the rename convention? With the new key preserving the original format the mentioned gap regarding original request containing '_' would be fixed..


> Would like to add a custom header with a directive of skip, fail, rename and we will act accordingly

I don't think this should really be a user settable option, in anyway, as it should be transparent what underlying data storage Noobaa uses.. But if you wanted to implement this i have no issues with it!
There's only one keypoint that needs to be adressed by the solution: If no custom directive skip header is set some sort of renaming should always happen. This is mainly because you do not always have full control over the metadata that is sent, for example when using Spark, Hadoop, Deltalake, Parquet.. This is also why we stumbled across the issue and needed to fix it.

> If you want to pick this one up (according to my latest comment) that's fine, we would also love to pick it up and fix it ourselves, so waiting on your response

I would be very happy if you can pick it up! If you need anything from us, we are there to asisst! :)

nimrod-becker:
Right so that would simply mean we set default to rename :)

We will pick it up

PR #7186, opened 2023-02-01
Fix folder creation on namespaced Azure blob backend

t-pirhonen:
### Explain the changes
1. on object upload set uploadStream size to 1 if object size is 0 (= empty directory)

1. fixes folder creation on namespace Azure blob backend. For some reason the blob client throws an error when size of the uploaded object is 0.
2. maybe there is a better solution however this is the most simple at the moment

### Testing Instructions:
2. create directory

### DIFF
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -268,7 +268,7 @@ class NamespaceBlob {
 
                 obj = await blob_client.uploadStream(
                     new_stream.pipe(count_stream),
-                    params.size,
+                    params.size === 0 ? 1 : params.size,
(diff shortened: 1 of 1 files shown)


liranmauda:
Hi @t-pirhonen 
Thanks for this PR.
It looks like this PR carried another Commit from a different PR.
Please rebase from master and see that the only commit there is the relevant one. 
After rebasing squash the commits and we will review the PR

lallinger-arbeit:
@liranmauda should be ok now!

nimrod-becker:
This PR is over a year old, if still relevant please rebase and reopen

PR #7192, opened 2023-02-01
Change upload object ID encoding to hex on Azure blob backend

t-pirhonen:
### Explain the changes
1. change the encoding of object ID from base64 to hex

1. WinSCP's implementation of S3 does not support base64 encoded upload IDs as it does not urlencode the ID correctly, when it contains '/', '+', '='. using hex solves this
2. also the etag, which is hex-encoded, is used as upload ID in the noobaa namespaced s3 implementation

### Testing Instructions:
2. use WinSCP to upload multiple objects
3. => some of the uploads fail, but if you retry often enough every object can be uploaded. this is due to the base64 encoding of the upload ID sometimes contain the characters mentioned above

### DIFF
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -384,7 +384,7 @@ class NamespaceBlob {
         }
         // obj_id contains the upload obj_id that is stored in db OR  obj_id contains a random string of bytes
         // using the first option when need to store xattr and set it after complete multipart upload phase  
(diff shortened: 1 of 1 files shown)


lallinger-arbeit:
i know this technically isn't a noobaa bug but we want our external customers to be able to use WinSCP with noobaa and patching WinSCP is not really an option:/ i would greatly appreciate it if you could still merge it!

nimrod-becker:
This PR is over a year old, if still relevant please rebase and reopen

PR #7200, opened 2023-02-09
Fix for HTTP Agent no longer working with unsecure S3 backend

t-pirhonen:
### Explain the changes
1. remove checking if NODE_EXTRA_CA_CERTS is empty

1. fixes getting an HTTP Client when requesting an unsecure HTTP Agent
2. broken by https://github.com/noobaa/noobaa-core/commit/621a634abba9ec0f4fd6774fa1b2f0e22d5f8a7f#diff-edc32a1c01ddf9d0c80b785573ad7de7443829016b1260eee65c5f1e92961406R398

### Testing Instructions:
1. connect noobaa to unsecure S3 backend e.g. Minio
2. get and put requests fail

### DIFF
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -411,7 +411,7 @@ function get_unsecured_agent(endpoint) {
     const is_aws_address = cloud_utils.is_aws_endpoint(endpoint);
     const hostname = url.parse(endpoint) ? url.parse(endpoint).hostname : endpoint;
     const is_localhost = _.isString(hostname) && hostname.toLowerCase() === 'localhost';
-    return _get_http_agent(endpoint, is_localhost || (!is_aws_address && _.isEmpty(NODE_EXTRA_CA_CERTS)));
+    return _get_http_agent(endpoint, is_localhost || !is_aws_address);
 }
 
 function _get_http_agent(endpoint, request_unsecured) {


jackyalbo:
Can you explain why we are moving this? Better on the PR itself. Why do we want to return to using an unsecure clients?

lallinger-arbeit:
hey @jackyalbo,
the problem is the method is called `get_unsecured_agent` and this is correctly called when using a S3 backend without httpS (e.g. a local MinIO instance), but when `NODE_EXTRA_CA_CERTS` is not set the method still falsely returns an httpS client resulting in a failure to connect to the backend S3

jackyalbo:
Hey @lallinger-arbeit 
So the thing is we probably don't want to revert this call... We do want that in the case NODE_EXTRA_CA_CERTS is not empty we will use a secured https client. Maybe there is an issue there, that for some reason although  NODE_EXTRA_CA_CERTS is empty or even don't exist - we still return a secured https client and that's maybe need to be better handled. I would suggest adding a print to check what you get in  NODE_EXTRA_CA_CERTS in your case and fixing this condition accordingly.

lallinger-arbeit:
I don't understand why i get an https agent when i explicitly specify http e.g. `http://s3.mycloud.com`? In my opinion this shouldn't happen regardless of if there are NODE_EXTRA_CA_CERTS set or not..

jackyalbo:
request_unsecured not suppose to effect if you get http client or https, that's suppose to be effected only by the protocol in url.parse. take a look here:
https://github.com/noobaa/noobaa-core/blob/0a04de2f7e53289d42416d29f491940ce81ea8c7/src/util/http_utils.js#L420
maybe the fix is hiding there

jackyalbo:
maybe I'll add some more info, the difference between unsecured_https_agent and regular https_agent is just if we reject unauthorized/self signed certs or not. secured doesn't mean https vs http.

https://github.com/noobaa/noobaa-core/blob/0a04de2f7e53289d42416d29f491940ce81ea8c7/src/util/http_utils.js#L27-L28

lallinger-arbeit:
ah ok, then we'll need to verify if this is still happening! Will take about 2 weeks, as tuomas is currently out of office..

PR #7203, opened 2023-02-13
NSFS | versioning | Implement list-object-versions

### LINKED ISSUE #7169: NSFS | versioning | Implement list-object-versions
This feature is part of NSFS versioning.

`list-object-versions` Returns metadata about all versions of the objects in a bucket. You can also use request parameters as selection criteria to return metadata about a subset of all the object versions.

more info: https://docs.aws.amazon.com/cli/latest/reference/s3api/list-object-versions.html

**Additional context**
sample output for `list-object-versions`

> aws s3api list-object-versions --bucket my-bucket --prefix index.html
>  Output:
> 
> {
>     "DeleteMarkers": [
>         {
>             "Owner": {
>                 "DisplayName": "my-username",
>                 "ID": "7009a8971cd660687538875e7c86c5b672fe116bd438f46db45460ddcd036c32"
>             },
>             "IsLatest": true,
>             "VersionId": "B2VsEK5saUNNHKcOAJj7hIE86RozToyq",
>             "Key": "index.html",
>             "LastModified": "2015-11-10T00:57:03.000Z"
>         },
>         {
>             "Owner": {
>                 "DisplayName": "my-username",
>                 "ID": "7009a8971cd660687538875e7c86c5b672fe116bd438f46db45460ddcd036c32"
>             },
>             "IsLatest": false,
>             "VersionId": ".FLQEZscLIcfxSq.jsFJ.szUkmng2Yw6",
>             "Key": "index.html",
>             "LastModified": "2015-11-09T23:32:20.000Z"
>         }
>     ],
>     "Versions": [
>         {
>             "LastModified": "2015-11-10T00:20:11.000Z",
>             "VersionId": "Rb_l2T8UHDkFEwCgJjhlgPOZC0qJ.vpD",
>             "ETag": "\"0622528de826c0df5db1258a23b80be5\"",
>             "StorageClass": "STANDARD",
>             "Key": "index.html",
>             "Owner": {
>                 "DisplayName": "my-username",
>                 "ID": "7009a8971cd660687538875e7c86c5b672fe116bd438f46db45460ddcd036c32"
>             },
>             "IsLatest": false,
>             "Size": 38
>         },
>         {
>             "LastModified": "2015-11-09T23:26:41.000Z",
>             "VersionId": "rasWWGpgk9E4s0LyTJgusGeRQKLVIAFf",
>             "ETag": "\"06225825b8028de826c0df5db1a23be5\"",
>             "StorageClass": "STANDARD",
>             "Key": "index.html",
>             "Owner": {
>                 "DisplayName": "my-username",
>                 "ID": "7009a8971cd660687538875e7c86c5b672fe116bd438f46db45460ddcd036c32"
>             },
>             "IsLatest": false,
>             "Size": 38
>         },
>         {
>             "LastModified": "2015-11-09T22:50:50.000Z",
>             "VersionId": "null",
>             "ETag": "\"d1f45267a863c8392e07d24dd592f1b9\"",
>             "StorageClass": "STANDARD",
>             "Key": "index.html",
>             "Owner": {
>                 "DisplayName": "my-username",
>                 "ID": "7009a8971cd660687538875e7c86c5b672fe116bd438f46db45460ddcd036c32"
>             },
>             "IsLatest": false,
>             "Size": 533823
>         }
>     ]
> }
> 

@romayalon Please correct me if wrong

vh05:
### Explain the changes
list-object-versions returns metadata about all versions of the objects in a bucket. You can also use request parameters as selection criteria to return metadata about a subset of all the object versions.

Moved object sorting under .versions to dir_cache
    
    1. Kept 2 seperate cache for list-objects and list_object_versions
    2. Sorting of the all entries done in version_dir_cache
    3. handled keymarker and versions id marker
    4. Added test cases for list object versions
    
Issues: Fixed #7169

### Testing Instructions:
1.  Create few objects and append the same to create versions of the original object
2. Delete few object
3. Performat `list-object_versions` operation. This should returns the metadata of all objects including `versions` and `delete_markers`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_get_bucket_versions.js
+++ b/src/endpoint/s3/ops/s3_get_bucket_versions.js
@@ -17,6 +17,11 @@ async function get_bucket_versions(req) {
         throw new S3Error(S3Error.InvalidArgument);
     }
 
+    if (req.query['version-id-marker'] && !req.query['key-marker']) {
+        dbg.warn('A version-id marker cannot be specified without a key marker');
+        throw new S3Error(S3Error.InvalidArgument);
+    }
+
     const reply = await req.object_sdk.list_object_versions({
         bucket: req.params.bucket,
         prefix: req.query.prefix,
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1,5 +1,6 @@
 /* Copyright (C) 2020 NooBaa */
-/*eslint max-lines: ["error", 2500]*/
+/*eslint max-lines: ["error", 3000]*/
+/*eslint max-statements: ["error", 80, { "ignoreTopLevelFunctions": true }]*/
 'use strict';
 
 const _ = require('lodash');
@@ -8,7 +9,6 @@ const path = require('path');
 const util = require('util');
 const mime = require('mime');
 const { v4: uuidv4 } = require('uuid');
-
 const P = require('../util/promise');
 const dbg = require('../util/debug_module')(__filename);
 const config = require('../../config');
@@ -77,6 +77,47 @@ function sort_entries_by_name(a, b) {
     if (a.name < b.name) return -1;
     if (a.name > b.name) return 1;
     return 0;
+}
+
+function _get_mtime_from_filename(filename) {
+    if (!filename.includes('_mtime-')) {
+        // Latest file wont have time suffix which will push the latest
+        // object last in the list. So to keep the order maintained,
+        // returning the latest time. Multiplying with 1e6 to provide
+        // nano second precision
+        return BigInt(Date.now() * 1e6);
+    }
+    const file_parts = filename.split('-');
+    return size_utils.string_to_bigint(file_parts[file_parts.length - 3], 36);
+}
+
+function _get_filename(file_name) {
+    if (file_name.includes('_mtime-')) {
+        return file_name.substring(0, file_name.indexOf('_mtime-'));
+    }
+    return file_name;
+}
+/**
+ * @param {fs.Dirent} first_entry
+ * @param {fs.Dirent} second_entry
+ * @returns {Number}
+ */
+function sort_entries_by_name_and_time(first_entry, second_entry) {
+    const first_entry_name = _get_filename(first_entry.name);
+    const second_entry_name = _get_filename(second_entry.name);
+    if (first_entry_name === second_entry_name) {
+        const first_entry_mtime = _get_mtime_from_filename(first_entry.name);
+        const second_entry_mtime = _get_mtime_from_filename(second_entry.name);
+        // To sort the versions in the latest first order,
+        // below logic is followed
+        if (second_entry_mtime < first_entry_mtime) return -1;
+        if (second_entry_mtime > first_entry_mtime) return 1;
+        return 0;
+    } else {
+        if (first_entry_name < second_entry_name) return -1;
+        if (first_entry_name > second_entry_name) return 1;
+        return 0;
+    }
 }
 
 function isDirectory(ent) {
@@ -185,6 +226,62 @@ const dir_cache = new LRUCache({
         if (stat.size <= config.NSFS_DIR_CACHE_MAX_DIR_SIZE) {
             sorted_entries = await nb_native().fs.readdir(fs_context, dir_path);
             sorted_entries.sort(sort_entries_by_name);
+            for (const ent of sorted_entries) {
+                usage += ent.name.length + 4;
+            }
+        }
+        return { time, stat, sorted_entries, usage };
+    },
+    validate: async ({ stat }, { dir_path, fs_context }) => {
+        const new_stat = await nb_native().fs.stat(fs_context, dir_path);
+        return (new_stat.ino === stat.ino && new_stat.mtimeNsBigint === stat.mtimeNsBigint);
+    },
+    item_usage: ({ usage }, dir_path) => usage,
+    max_usage: config.NSFS_DIR_CACHE_MAX_TOTAL_SIZE,
+});
+
+/**
+ * @typedef {{
+ *  time: number,
+ *  stat: nb.NativeFSStats,
+ *  usage: number,
+ *  sorted_entries?: fs.Dirent[],
+ * }} ReaddirVersionsCacheItem
+ * @type {LRUCache<object, string, ReaddirCacheItem>}
+ */
+const versions_dir_cache = new LRUCache({
+    name: 'nsfs-versions-dir-cache',
+    make_key: ({ dir_path }) => dir_path,
+    load: async ({ dir_path, fs_context }) => {
+        const time = Date.now();
+        const stat = await nb_native().fs.stat(fs_context, dir_path);
+        const version_path = dir_path + "/" + HIDDEN_VERSIONS_PATH;
+        let ver_dir_stat_size;
+        let is_version_path_exists = false;
+        try {
+            const ver_dir_stat = await nb_native().fs.stat(fs_context, version_path);
+            ver_dir_stat_size = ver_dir_stat.size;
+            is_version_path_exists = true;
+        } catch (err) {
+            if (err.code === 'ENOENT') {
+                dbg.log0('NamespaceFS: Version dir not found, ', version_path);
+            } else {
+                throw err;
+            }
+            ver_dir_stat_size = 0;
+        }
+        let sorted_entries;
+        let usage = config.NSFS_DIR_CACHE_MIN_DIR_SIZE;
+        if (stat.size + ver_dir_stat_size <= config.NSFS_DIR_CACHE_MAX_DIR_SIZE) {
+            const latest_versions = await nb_native().fs.readdir(fs_context, dir_path);
+            if (is_version_path_exists) {
+                const old_versions = await nb_native().fs.readdir(fs_context, version_path);
+                const entries = latest_versions.concat(old_versions);
+                sorted_entries = entries.sort(sort_entries_by_name_and_time);
+            } else {
+                sorted_entries = latest_versions.sort(sort_entries_by_name);
+            }
+            /*eslint no-unused-expressions: ["error", { "allowTernary": true }]*/
             for (const ent of sorted_entries) {
                 usage += ent.name.length + 4;
             }
@@ -315,6 +412,27 @@ class NamespaceFS {
      * @param {ListParams} params
      */
     async list_objects(params, object_sdk) {
+        return this._list_objects(params, object_sdk, false);
+    }
+
+    /**
+     * @typedef {{
+     *  bucket: string,
+     *  prefix?: string,
+     *  delimiter?: string,
+     *  key_marker?: string,
+     *  version_id_marker?: string,
+     *  limit?: number,
+     * }} ListVersionsParams
+     */
+    /**
+     * @param {ListVersionsParams} params
+     */
+    async list_object_versions(params, object_sdk) {
+        return this._list_objects(params, object_sdk, true);
+    }
+
+    async _list_objects(params, object_sdk, list_versions) {
         try {
             const fs_context = this.prepare_fs_context(object_sdk);
             await this._load_bucket(params, fs_context);
@@ -323,13 +441,13 @@ class NamespaceFS {
                 bucket,
                 delimiter = '',
                 prefix = '',
+                version_id_marker = '',
                 key_marker = '',
             } = params;
 
             if (delimiter && delimiter !== '/') {
                 throw new Error('NamespaceFS: Invalid delimiter ' + delimiter);
             }
-
             const limit = Math.min(1000, _.isUndefined(params.limit) ? 1000 : params.limit);
             if (limit < 0) throw new Error('Limit must be a positive Integer');
             // In case that we've received max-keys 0, we should return an empty reply without is_truncated
@@ -354,25 +472,20 @@ class NamespaceFS {
              * @returns {Promise<void>}
              */
             const process_dir = async dir_key => {
-                if (this._skip_version_dir(dir_key)) {
+                if (this._is_hidden_version_path(dir_key)) {
                     return;
                 }
-
                 // /** @type {fs.Dir} */
                 let dir_handle;
-
                 /** @type {ReaddirCacheItem} */
                 let cached_dir;
-
                 const dir_path = path.join(this.bucket_path, dir_key);
-
                 const prefix_dir = prefix.slice(0, dir_key.length);
                 const prefix_ent = prefix.slice(dir_key.length);
                 if (!dir_key.startsWith(prefix_dir)) {
                     // dbg.log0(`prefix dir does not match so no keys in this dir can apply: dir_key=${dir_key} prefix_dir=${prefix_dir}`);
                     return;
                 }
-
                 const marker_dir = key_marker.slice(0, dir_key.length);
                 const marker_ent = key_marker.slice(dir_key.length);
                 // marker is after dir so no keys in this dir can apply
@@ -383,7 +496,6 @@ class NamespaceFS {
                 // when the dir portion of the marker is completely below the current dir
                 // then every key in this dir satisfies the marker and marker_ent should not be used.
                 const marker_curr = (marker_dir < dir_key) ? '' : marker_ent;
-
                 // dbg.log0(`process_dir: dir_key=${dir_key} prefix_ent=${prefix_ent} marker_curr=${marker_curr}`);
                 /**
                  * @typedef {{
@@ -393,7 +505,11 @@ class NamespaceFS {
                  */
                 const insert_entry_to_results_arr = async r => {
                     let pos;
-                    if (results.length && r.key < results[results.length - 1].key) {
+                    // Since versions are arranged next to latest object in the latest first order,
+                    // no need to find the sorted last index. Push the ".versions/#VERSION_OBJECT" as
+                    // they are in order
+                    if (results.length && r.key < results[results.length - 1].key &&
+                        !this._is_hidden_version_path(r.key)) {
                         pos = _.sortedLastIndexBy(results, r, a => a.key);
                     } else {
                         pos = results.length;
@@ -403,7 +519,6 @@ class NamespaceFS {
                         is_truncated = true;
                         return; // not added
                     }
-
                     if (!delimiter && r.common_prefix) {
                         await process_dir(r.key);
                     } else {
@@ -424,26 +539,37 @@ class NamespaceFS {
                  */
                 const process_entry = async ent => {
                     // dbg.log0('process_entry', dir_key, ent.name);
-                    if (!ent.name.startsWith(prefix_ent) ||
+                    if ((!ent.name.startsWith(prefix_ent) ||
                         ent.name < marker_curr ||
                         ent.name === this.get_bucket_tmpdir() ||
-                        ent.name === config.NSFS_FOLDER_OBJECT_NAME) {
+                        ent.name === config.NSFS_FOLDER_OBJECT_NAME) &&
+                        !this._is_hidden_version_path(ent.name)) {
                         return;
                     }
-
                     const isDir = await is_directory_or_symlink_to_directory(ent, fs_context, path.join(dir_path, ent.name));
 
-                    const r = {
-                        key: this._get_entry_key(dir_key, ent, isDir),
-                        common_prefix: isDir,
-                    };
+                    let r;
+                    if (list_versions && ent.name.includes('_mtime-')) {
+                        r = {
+                            key: this._get_version_entry_key(dir_key, ent),
+                            common_prefix: isDir,
+                        };
+                    } else {
+                        r = {
+                            key: this._get_entry_key(dir_key, ent, isDir),
+                            common_prefix: isDir,
+                        };
+                    }
                     await insert_entry_to_results_arr(r);
                 };
 
                 if (!(await this.check_access(fs_context, dir_path))) return;
-
                 try {
-                    cached_dir = await dir_cache.get_with_cache({ dir_path, fs_context });
+                    if (list_versions) {
+                        cached_dir = await versions_dir_cache.get_with_cache({ dir_path, fs_context });
+                    } else {
+                        cached_dir = await dir_cache.get_with_cache({ dir_path, fs_context });
+                    }
                 } catch (err) {
                     if (err.code === 'ENOENT') {
                         dbg.log0('NamespaceFS: no keys for non existing dir', dir_path);
@@ -462,11 +588,27 @@ class NamespaceFS {
 
                 if (cached_dir.sorted_entries) {
                     const sorted_entries = cached_dir.sorted_entries;
-                    const marker_index = _.sortedLastIndexBy(
-                        sorted_entries,
-                        make_named_dirent(marker_curr),
-                        get_entry_name
-                    );
+                    let marker_index;
+                    // Two ways followed here to find the index.
+                    // 1. When inside marker_dir: Here the entries are sorted based on time. Here
+                    //    FindIndex() is called since sortedLastIndexBy() expects sorted order by name
+                    // 2. When marker_dir above dir_path: sortedLastIndexBy() is called since entries are
+                    //     sorted by name
+                    if (list_versions && marker_curr && !marker_curr.includes('/')) {
+                        let start_marker = marker_curr;
+                        if (version_id_marker) start_marker = version_id_marker;
+                        marker_index = _.findIndex(
+                            sorted_entries,
+                            {name: start_marker}
+                        ) + 1;
+                    } else {
+                        marker_index = _.sortedLastIndexBy(
+                            sorted_entries,
+                            make_named_dirent(marker_curr),
+                            get_entry_name
+                        );
+                    }
+
                     // handling a scenario in which key_marker points to an object inside a directory
                     // since there can be entries inside the directory that will need to be pushed
                     // to results array
@@ -485,6 +627,10 @@ class NamespaceFS {
                     }
                     for (let i = marker_index; i < sorted_entries.length; ++i) {
                         const ent = sorted_entries[i];
+                        if (list_versions && marker_curr) {
+                            const ent_name = ent.name.substring(0, ent.name.indexOf('_mtime-'));
+                            if (ent_name !== marker_curr) break;
+                        }
                         // when entry is NSFS_FOLDER_OBJECT_NAME=.folder file,
                         // and the dir key marker is the name of the curr directory - skip on adding it
                         if (ent.name === config.NSFS_FOLDER_OBJECT_NAME && dir_key === marker_dir) {
@@ -497,7 +643,6 @@ class NamespaceFS {
                     }
                     return;
                 }
-
                 // for large dirs we cannot keep all entries in memory
                 // so we have to stream the entries one by one while filtering only the needed ones.
                 try {
@@ -539,26 +684,37 @@ class NamespaceFS {
                 common_prefixes: [],
                 is_truncated,
                 next_marker: undefined,
+                next_version_id_marker: undefined,
             };
             for (const r of results) {
+                let obj_info;
                 if (r.common_prefix) {
                     res.common_prefixes.push(r.key);
                 } else {
-                    res.objects.push(this._get_object_info(bucket, r.key, r.stat, 'null'));
+                    obj_info = this._get_object_info(bucket, r.key, r.stat, 'null', true);
+                    if (!list_versions && obj_info.delete_marker) {
+                        continue;
+                    }
+                    if (this._is_hidden_version_path(obj_info.key)) {
+                        obj_info.key = path.normalize(obj_info.key.replace(HIDDEN_VERSIONS_PATH + '/', ''));
+                        obj_info.key = obj_info.key.substring(0, obj_info.key.indexOf('_mtime-'));
+                    }
+                    res.objects.push(obj_info);
                 }
                 if (res.is_truncated) {
-                    res.next_marker = r.key;
+                    if (list_versions && r.key.includes('_mtime-')) {
+                        const next_version_id_marker = r.key.substring(r.key.lastIndexOf('/') + 1);
+                        res.next_version_id_marker = next_version_id_marker;
+                        res.next_marker = next_version_id_marker.substring(0, next_version_id_marker.indexOf('_mtime'));
+                    } else {
+                        res.next_marker = r.key;
+                    }
                 }
             }
             return res;
         } catch (err) {
             throw this._translate_object_error_codes(err);
         }
-    }
-
-    async list_object_versions(params, object_sdk) {
-        // for now we do not support versioning, so returning the same as list objects
-        return this.list_objects(params, object_sdk);
     }
 
     /////////////////
@@ -1547,6 +1703,16 @@ class NamespaceFS {
         return dir_key + ent.name + (isDir ? '/' : '');
     }
 
+    /**
+     * @param {string} dir_key
+     * @param {fs.Dirent} ent
+     * @returns {string}
+     */
+     _get_version_entry_key(dir_key, ent) {
+        if (ent.name === config.NSFS_FOLDER_OBJECT_NAME) return dir_key;
+        return dir_key + HIDDEN_VERSIONS_PATH + '/' + ent.name;
+    }
+
     async _make_path_dirs(file_path, fs_context) {
         const last_dir_pos = file_path.lastIndexOf('/');
         if (last_dir_pos > 0) return this._create_path(file_path.slice(0, last_dir_pos), fs_context);
@@ -1579,10 +1745,11 @@ class NamespaceFS {
      * @param {nb.NativeFSStats} stat 
      * @returns {nb.ObjectInfo}
      */
-    _get_object_info(bucket, key, stat, return_version_id) {
+     _get_object_info(bucket, key, stat, return_version_id, is_latest = true) {
         const etag = this._get_etag(stat);
         const encryption = this._get_encryption_info(stat);
         const version_id = return_version_id && this._is_versioning_enabled() && this._get_version_id_by_xattr(stat);
+
         return {
             obj_id: etag,
             bucket,
@@ -1593,8 +1760,8 @@ class NamespaceFS {
             content_type: mime.getType(key) || 'application/octet-stream',
             // temp:
             version_id: version_id,
-            is_latest: true,
-            delete_marker: false,
+            is_latest: is_latest,
+            delete_marker: stat.xattr[XATTR_DELETE_MARKER] === 'true',
             tag_count: 0,
             encryption,
             xattr: to_xattr(stat.xattr),
@@ -2216,7 +2383,7 @@ class NamespaceFS {
        }
     }
 
-    _skip_version_dir(dir_key) {
+    _is_hidden_version_path(dir_key) {
         const idx = dir_key.indexOf(HIDDEN_VERSIONS_PATH);
         return ((idx === 0) || (idx > 0 && dir_key[idx - 1] === '/'));
     }
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -79,8 +79,15 @@ function sort_entries_by_name(a, b) {
     return 0;
 }
 
+function _is_version_object(filename) {
+    const mtime_substr_index = filename.indexOf('_mtime-');
+    if (mtime_substr_index < 0) return false;
+    const ino_substr_index = filename.indexOf('-ino-');
+    return ino_substr_index > mtime_substr_index;
+}
+
 function _get_mtime_from_filename(filename) {
-    if (!filename.includes('_mtime-')) {
+    if (!_is_version_object(filename)) {
         // Latest file wont have time suffix which will push the latest
         // object last in the list. So to keep the order maintained,
         // returning the latest time. Multiplying with 1e6 to provide
@@ -92,7 +99,7 @@ function _get_mtime_from_filename(filename) {
 }
 
 function _get_filename(file_name) {
-    if (file_name.includes('_mtime-')) {
+    if (_is_version_object(file_name)) {
         return file_name.substring(0, file_name.indexOf('_mtime-'));
     }
     return file_name;
@@ -549,7 +556,7 @@ class NamespaceFS {
                     const isDir = await is_directory_or_symlink_to_directory(ent, fs_context, path.join(dir_path, ent.name));
 
                     let r;
-                    if (list_versions && ent.name.includes('_mtime-')) {
+                    if (list_versions && _is_version_object(ent.name)) {
                         r = {
                             key: this._get_version_entry_key(dir_key, ent),
                             common_prefix: isDir,
@@ -594,6 +601,8 @@ class NamespaceFS {
                     //    FindIndex() is called since sortedLastIndexBy() expects sorted order by name
                     // 2. When marker_dir above dir_path: sortedLastIndexBy() is called since entries are
                     //     sorted by name
+                    // 3. One of the below conditions, marker_curr.includes('/') checks whether
+                    //    the call is for the directory that contains marker_curr
                     if (list_versions && marker_curr && !marker_curr.includes('/')) {
                         let start_marker = marker_curr;
                         if (version_id_marker) start_marker = version_id_marker;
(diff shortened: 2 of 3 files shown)


romayalon:
@v-harihar 
Hi, wasn't here for a long time, finally, have the chance to review these changes.
I see some missing implementation here - In the design doc, few implementation details were discussed and I can't find them covered by this PR -
Sort  - all keys and within key sort all versions.
New parameters - 
version_id_marker - will start listing from that marker
New response fields - 
DeleteMarker- will tell if a  version is a delete marker / regular version.
IsLatest - boolean (will be marked when traversing on the real folder/ null for disabled buckets)
VersionIDMarker - string
NextVersionIDMarker - string

Also, we will need relevant unit tests for this to be covered.

v-harihar:
> @v-harihar Hi, wasn't here for a long time, finally, have the chance to review these changes. I see some missing implementation here - In the design doc, few implementation details were discussed and I can't find them covered by this PR - Sort - all keys and within key sort all versions. New parameters - version_id_marker - will start listing from that marker New response fields - DeleteMarker- will tell if a version is a delete marker / regular version. IsLatest - boolean (will be marked when traversing on the real folder/ null for disabled buckets) VersionIDMarker - string NextVersionIDMarker - string
> 
> Also, we will need relevant unit tests for this to be covered.

sure.

src/sdk/namespace_fs.js
```diff
@@ -299,6 +299,26 @@ class NamespaceFS {
...
+        return this._list_objects(params, object_sdk, false);
+    }
+
+    async list_object_versions(params, object_sdk) {
+        return this._list_objects(params, object_sdk, true);
+    }
+
+    async _list_version_dir_objects(results, entry_name, dir_key) {
+        const filePath = path.normalize(path.join(this.bucket_path, '/', dir_key, entry_name));
+        const versions = await readdir(filePath);
```
romayalon, line 311: you should use await nb_native().fs.readdir(fs_context, filePath);

src/sdk/namespace_fs.js
```diff
@@ -332,16 +352,16 @@ class NamespaceFS {
...
             const results = [];
+            const versions = [];
 
             /**
              * @param {string} dir_key
              * @returns {Promise<void>}
              */
             const process_dir = async dir_key => {
-                if (this._skip_version_dir(dir_key)) {
+                if (!list_versions && this._skip_version_dir(dir_key)) {
```
romayalon, line 362: should list_versions boolean check be inside skip_version_dir() ?

src/sdk/namespace_fs.js
```diff
@@ -513,7 +549,37 @@ class NamespaceFS {
...
+                    const obj_info = this._get_object_info(bucket, r.key, r.stat, 'null', true);
+                    if (!list_versions && obj_info.delete_marker) {
+                        continue;
+                    }
+                    if (obj_info.delete_marker) {
+                        obj_info.is_latest = false;
+                    }
+                    if (list_versions) {
+                        versions.forEach(v => {
+                            var search_key = v.key.replace(HIDDEN_VERSIONS_PATH, '');
```
romayalon, line 562: can you explain what this code does?

src/sdk/namespace_fs.js
```diff
@@ -402,7 +422,11 @@ class NamespaceFS {
                         return; // not added
                     }
 
-                    if (!delimiter && r.common_prefix) {
+                    if (ent.name === HIDDEN_VERSIONS_PATH && r.common_prefix) {
+                        if (list_versions) {
+                            await this._list_version_dir_objects(versions, ent.name, dir_key);
```
romayalon, line 427: why is this new function needed? can't we re-use the existing flow of processing a dir? 
and versions will be pushed inside the existing results array instead of the new versions array, the usage of the dir_cache also already sorts the entries for us

src/sdk/namespace_fs.js
```diff
@@ -513,7 +549,37 @@ class NamespaceFS {
...
+                                search_key = search_key.replace('/', '');
+                            } else {
+                                search_key = search_key.replace('//', '/');
+                            }
+                            if (search_key.startsWith(r.key)) {
+                                obj_info_list.push(this._get_object_info(bucket, r.key, v.stat, 'null', false));
+                            }
+                        });
+                    }
+                    /* arrange the object versions in ascending create time */
```
romayalon, line 573: this means that S3 sorts the keys by name but the versions by creation time?

src/sdk/namespace_fs.js
```diff
@@ -513,7 +549,37 @@ class NamespaceFS {
                 if (r.common_prefix) {
                     res.common_prefixes.push(r.key);
                 } else {
-                    res.objects.push(this._get_object_info(bucket, r.key, r.stat, 'null'));
+                    const obj_info_list = [];
+                    const obj_info = this._get_object_info(bucket, r.key, r.stat, 'null', true);
+                    if (!list_versions && obj_info.delete_marker) {
```
romayalon, line 554: can this situation happen? since we are not listing objects under .versions/ and delete_marker must be under .versions/

src/sdk/namespace_fs.js
```diff
@@ -513,7 +549,37 @@ class NamespaceFS {
...
+                        continue;
+                    }
+                    if (obj_info.delete_marker) {
```
romayalon, line 557: delete marker can be latest version if there is no current version
for example:
put object (version1 created)
delete object (delete marker was created and latest)

src/sdk/namespace_fs.js
v-harihar: sure

src/sdk/namespace_fs.js
v-harihar: Here in this case, `.versions` dir is parsed based on every `process_entry()`. So the `versions` list can be pushed on top of  the `results` stack based on the `create_time` and then `process_entry()` which is the latest object is pushed to the stack.

src/sdk/namespace_fs.js
v-harihar: This is incase of `list_objects`. When the latest object has `delete_marker`, then `list_objects` wont expect it to list the object which has `delete_marker`

src/sdk/namespace_fs.js
v-harihar: If the object is deleted then do we have to keep it `latest` ? 🤔

src/sdk/namespace_fs.js
v-harihar: Here I am trying to pull the file_path from `version` list by removing `.versions` dir from the file path.

ex:  lets consider that we want to push $BUCKET_PATH/file1 and its versions.

On `.versions` list will have list of all versions of all objects. Here we need to filter `version` list based on the `$BUCKET_PATH/file1`. But the keys on `version` list would have `.versions` in the path name. So to compare the `process_entry()`, we need to remove `.versions` from the file path.

src/sdk/namespace_fs.js
v-harihar: Yes. I observed that the o/p was not sorted based on time. So sorting the entries based on c_time and pushing on the stack

src/sdk/namespace_fs.js
romayalon: this can be checked on AWS :) but I guess that the answer is yes

src/sdk/namespace_fs.js
romayalon: in process entry we use _get_entry_key(), can you move this logic there and call _get_entry_key() here instead?

src/sdk/namespace_fs.js
```diff
@@ -402,7 +421,11 @@ class NamespaceFS {
                         return; // not added
                     }
 
-                    if (!delimiter && r.common_prefix) {
+                    if (ent.name === HIDDEN_VERSIONS_PATH && r.common_prefix) {
+                        if (list_versions) {
+                            await this._list_version_dir_objects(fs_context, versions, ent.name, dir_key);
```
romayalon, line 426: as we discussed before, can't we just use process_dir ?
in process_dir() we use the dir_cache which makes the performance much better and provides sorting as well, the readdir exists also there in the cache

src/sdk/namespace_fs.js
romayalon: wait not sure I understand, in AWS it's not sorted by time but you did sort by time? 
according to what the versions should be sorted?

src/sdk/namespace_fs.js
```diff
@@ -299,6 +298,26 @@ class NamespaceFS {
...
+        return this._list_objects(params, object_sdk, true);
+    }
+
+    async _list_version_dir_objects(fs_context, results, entry_name, dir_key) {
+        const filePath = path.normalize(path.join(this.bucket_path, '/', dir_key, entry_name));
+        const versions = await nb_native().fs.readdir(fs_context, filePath);
+
+        versions.forEach(version => {
+            const v = {
+                key: path.normalize(path.join(dir_key, entry_name, '/', version.name)),
```
romayalon, line 314: do we need the full path for stat? 
I think also it could be added to the cache with a small change

src/sdk/namespace_fs.js
```diff
@@ -332,16 +351,16 @@ class NamespaceFS {
 
             /** @type {Result[]} */
             const results = [];
+            const versions = [];
```
romayalon, line 354: why a new array is needed? can't we push the results array?

src/sdk/namespace_fs.js
```diff
@@ -496,13 +519,25 @@ class NamespaceFS {
...
+
             await Promise.all(results.map(async r => {
                 if (r.common_prefix) return;
                 const entry_path = path.join(this.bucket_path, r.key);
                 //If entry is outside of bucket, returns stat of symbolic link
                 const use_lstat = !(await this._is_path_in_bucket_boundaries(fs_context, entry_path));
                 r.stat = await nb_native().fs.stat(fs_context, entry_path, { use_lstat });
             }));
+
+            if (list_versions) {
```
romayalon, line 531: this will be redundant if versions will be kept in the results arr

src/sdk/namespace_fs.js
romayalon: but on list_objects we won't read .versions folder and delete_marker even if it's the latest will stay under .versions

v-harihar: So the latest object will be moved under .versions once deleted ?

I am bit confused here

src/sdk/namespace_fs.js
v-harihar: At line 547, the results array is traversed one by one and related version objects inside version dir are fetched, arranged based on c_time. Once  we have sorted version objects, we will push them into `object_info` at line 577 to 581

src/sdk/namespace_fs.js
v-harihar: Yes, this is taken care.

src/sdk/namespace_fs.js
v-harihar: `list_object_versions` list the objects based on the time, latest appears 1st in the o/p. We need to return the object_info sorted based on the time then we can match the behavior on AWS

src/sdk/namespace_fs.js
v-harihar: This method is not relevant since `process_dir` is used to update the version list.

src/sdk/namespace_fs.js
romayalon: on deletion of latest version (version id not specified) -
1. latest version moves to .versions/
2. new delete marker is created under .versions/ and will be the latest version

On deletion of latets version (version id specified) - 
1. the version itself will be deleted
2. If second latest version exists and not a delete marker it will move outside from .versions/ to be the latest version

v-harihar: Makes sense. Removed explicit assignment `latest = false`

src/sdk/namespace_fs.js
```diff
@@ -513,7 +534,36 @@ class NamespaceFS {
...
+                                search_key = search_key.replace('/', '');
+                            } else {
+                                search_key = search_key.replace('//', '/');
+                            }
+                            if (search_key.startsWith(r.key)) {
+                                obj_info_list.push(this._get_object_info(bucket, r.key, v.stat, 'null', false));
+                            }
+                        });
+                        /* arrange the object versions in ascending create time */
+                        obj_info_list.sort(function(o1, o2) {
```
romayalon, line 555: do you think we can add it to the dir_cache as we do for the other sort? 
sort by time will be called if the dir is some_path/.versions/ ?

src/sdk/namespace_fs.js
```diff
@@ -513,7 +534,36 @@ class NamespaceFS {
...
                 } else {
-                    res.objects.push(this._get_object_info(bucket, r.key, r.stat, 'null'));
+                    const obj_info = this._get_object_info(bucket, r.key, r.stat, 'null', true);
+                    if (!list_versions && obj_info.delete_marker) {
+                        continue;
+                    }
+                    if (list_versions) {
+                        const obj_info_list = [];
+                        versions.forEach(v => {
+                            let search_key = v.key.replace(HIDDEN_VERSIONS_PATH, '');
```
romayalon, line 544: if I understood correctly -  key is not the path anymore,  can this code be removed?

src/sdk/namespace_fs.js
v-harihar: If I am not wrong, in `process_dir()`, we enter each dir and process them into a list. 

ex: assume `dir1` `dir2` are 2 dirs, under which we have `.versions` dir. When we enter `dir1`, we process the entries. If we sort the objects under `.versions` then we sort all entries irrespective of versions of particular key. So I thought not to go for sort here and filter the versions outside based on the key, push them to `obj_info_list` and sort them based on time.

romayalon: aha I understand, can we maybe sort first by key and then by create time? 
because here we will always need to sort even though we could have these entries using cache

src/sdk/namespace_fs.js
romayalon: so on get_entry_key() dir_key is dir1/.versions/ and ent is key1_xyz? and then we remove .versions/, do you think we can just return dir1/key1_xyz directly from get_entry_key() instead ?

src/sdk/namespace_fs.js
v-harihar: We can skip the sort based on `key` for '.versions' dir.
Whether we sort here or outside number of sorting cycles wont differ. 
Once we have the list ready (cached), we can run the sorting for `.versions` once the `stat` for the objects available.

src/sdk/namespace_fs.js
romayalon: yes but you will need to remove it only from the end of dir_key so you won't have these many cases, no? and also I think it's the correct place to put it since this is the function handling how get_entry_key() should look like

src/sdk/namespace_fs.js
romayalon: the downside of not sorting it inside the cache is that the cache won't keep the entries sorted so we will need to sort the versions every time the user calls list_object_version,
A question - 
is there any specific reason for sorting by c_time (file's metadata, such as its ownership, location, file type, and permission settings change time) for sorting instead of sorting by m_time (content modification time)? 
I'm wondering if we can maybe instead of stat take the m_time from the version's key, and then you can sort it inside the cache
Tagging @guymguym if you want to share an opinion here

src/sdk/namespace_fs.js
v-harihar: One inevitable case. we need `full_path` of the `version file` to get the stat of the `versioned object`. If we modify the `r.key` in `get_entry_key()` then for `stat` will result in taking `stat` on latest object.

Please correct me if wrong,

src/sdk/namespace_fs.js
dannyzaken: Following an online review with @v-harihar and @romayalon:
* in the current implementation, there is a nested loop over all dir entries and all versions under `.versions`. this can have scale issues when there are a lot of files and versions and needs to be changed.
* As Romy suggested above, we will keep `.versions` sorted in the cache. Sort order is first by key and then by version time. 
* once we have both the dir entries and the versions in 2 sorted lists, we can merge them together, without the need for nested loops or resorting
* after merging the 2 lists, we need to take into account the `key_marker`, the `version_id_marker` and `limit` for returning the actual list result

src/sdk/namespace_fs.js
```diff
@@ -79,6 +79,22 @@ function sort_entries_by_name(a, b) {
...
+/**
+ * @param {fs.Dirent} a
+ * @param {fs.Dirent} b
+ * @returns {Number}
+ */
+function sort_entries_by_time(a, b) {
+    const a_mtime = _get_mtime_from_filename(a.name);
+    const b_mtime = _get_mtime_from_filename(b.name);
+    return b_mtime - a_mtime;
+}
```
dannyzaken, line 96: you should compare first by file name and only if filenames match then compare by mtime. something like that:

```suggestion
function sort_entries_by_name_and_time(a, b) {
    const a_name = _get_filename(a.name) // will cut the version suffix
    const b_name = _get_filename(b.name)
    if (a_name === b_name) {
        const a_mtime = _get_mtime_from_filename(a.name);
        const b_mtime = _get_mtime_from_filename(b.name);
        return b_mtime - a_mtime;
    } else {
        if (a_name < b_name) return -1;
        if (a_name > b_name) return 1;
        return 0;
    }
}
```

src/sdk/namespace_fs.js
```diff
@@ -199,6 +215,55 @@ const dir_cache = new LRUCache({
...
+ */
+const versions_dir_cache = new LRUCache({
+    name: 'nsfs-versions-dir-cache',
+    make_key: ({ dir_path }) => dir_path,
+    load: async ({ dir_path, fs_context }) => {
+        const time = Date.now();
+        const stat = await nb_native().fs.stat(fs_context, dir_path);
+        const sorted_entries = [];
+        let usage = config.NSFS_DIR_CACHE_MIN_DIR_SIZE;
+        if (stat.size <= config.NSFS_DIR_CACHE_MAX_DIR_SIZE) {
```
dannyzaken, line 235: you should check for the sum of dir size and ver size

```suggestion
        if (stat.size + ver_dir_stat.size <= config.NSFS_DIR_CACHE_MAX_DIR_SIZE) {
```

src/sdk/namespace_fs.js
```diff
@@ -199,6 +215,55 @@ const dir_cache = new LRUCache({
...
+            sorted_versions.sort(sort_entries_by_time);
+            sorted_latest_versions.forEach(entry => {
+                sorted_entries.push(entry);
+                sorted_versions.forEach(version => {
+                    if (version.name.startsWith(entry.name)) {
+                        version.name = HIDDEN_VERSIONS_PATH + "/" + version.name;
+                        sorted_entries.push(version);
+                    }
+                });
+            });
```
dannyzaken, line 251: I think we can avoid this nested loop of entries and versions. The format of the version name allows us to sort them together

something like that (I may have syntax errors here):

```suggestion
            const latest_versions = await nb_native().fs.readdir(fs_context, dir_path);
            const version_path = dir_path + "/" + HIDDEN_VERSIONS_PATH;
            const old_versions = await nb_native().fs.readdir(fs_context, version_path);
            const entries = latest_versions.concat(old_versions);
            const sorted_entries = entries.sort(sort_entries_by_name_and_time);
```

dannyzaken (changes requested):

src/sdk/namespace_fs.js
v-harihar: Yes. The latest object (not under `.versions`) wont have version suffix. So When we sort based on time, the latest object will go last in the order

src/sdk/namespace_fs.js
v-harihar: sure

src/sdk/namespace_fs.js
v-harihar: For latest object, I will always return time.now(). Just to keep it on top of the entries

src/sdk/namespace_fs.js
v-harihar: With the suggested change, the `version_path` in the sorted arrays will be wrongly set. i.e assume that
the latest file is in `$BUCKET/dir1/file1` and  its version path will be `$BUCKET/dir1/.versions/file1_mtime-*`

with the suggested change, we will have `readdir(dir1)` which gives proper path i.e `$BUCKET/dir1/file1` but
`readdir(`dir1/.versions`)` will lead to `$BUCKET/dir1/file1_mtime-*` (NOTE missing `.versions` in the path)

This is handled from line no: 247-248 in the above code

>                     if (version.name.startsWith(entry.name)) {
>                         version.name = HIDDEN_VERSIONS_PATH + "/" + version.name;
>                         sorted_entries.push(version);
>                     }
>

v-harihar: Discussed and resolved

src/sdk/namespace_fs.js
```diff
@@ -199,6 +239,57 @@ const dir_cache = new LRUCache({
...
+        const stat = await nb_native().fs.stat(fs_context, dir_path);
+        const version_path = dir_path + "/" + HIDDEN_VERSIONS_PATH;
+        let ver_dir_stat_size;
+        let is_version_path_exists = false;
+        try {
+            const ver_dir_stat = await nb_native().fs.stat(fs_context, version_path);
+            ver_dir_stat_size = ver_dir_stat.size;
+            is_version_path_exists = true;
+        } catch (err) {
+            ver_dir_stat_size = 0;
```
dannyzaken, line 265: What's the purpose of this error handling?

dannyzaken: IIUC, you want to check for the existance of `.versions`. I think you should check specifically for `ENOENT`, and not ignore other errors

src/sdk/namespace_fs.js
```diff
@@ -79,6 +79,46 @@ function sort_entries_by_name(a, b) {
...
+function _get_mtime_from_filename(filename) {
+    if (!filename.includes('_mtime-')) {
+        // Latest file wont have time suffix which will push the latest
+        // object last in the list. So to keep the order maintained,
+        // returning the latest time. Multiplying with 1e6 to provide
+        // nano second precision
+        return Date.now() * 1e6;
+    }
+    const file_parts = filename.split('-');
+    return parseInt(file_parts[1], 36);
```
dannyzaken, line 91: once you get the mtime part, you should use `size_utils.string_to_bigint` and not `parseInt`
e.g.: see `_extract_version_info_from_xattr`

src/sdk/namespace_fs.js
```diff
@@ -79,6 +79,46 @@ function sort_entries_by_name(a, b) {
     return 0;
 }
 
+function _get_mtime_from_filename(filename) {
+    if (!filename.includes('_mtime-')) {
```
dannyzaken, line 83: notice there is a gap in this PR, where we kind of ignoring `null` versions. for null versions we also don't have `_mtime` in the filename. 
let's leave it for now and open a gap on the null versions case. we can think if we want to add `mtime` to null versions as well. this will also affect @shirady when implementing suspend versioning

src/sdk/namespace_fs.js
```diff
@@ -79,6 +79,46 @@ function sort_entries_by_name(a, b) {
...
+        // Latest file wont have time suffix which will push the latest
+        // object last in the list. So to keep the order maintained,
+        // returning the latest time. Multiplying with 1e6 to provide
+        // nano second precision
+        return Date.now() * 1e6;
```
dannyzaken, line 88: not sure, but you might need to return it as a `size_utils.BigInterger` for consistency 

```suggestion
        return size_utils.BigInterger(Date.now() * 1e6);
```

src/sdk/namespace_fs.js
```diff
@@ -79,6 +79,46 @@ function sort_entries_by_name(a, b) {
...
+function sort_entries_by_name_and_time(a, b) {
+    // const a_mtime = _get_mtime_from_filename(a.name);
+    // const b_mtime = _get_mtime_from_filename(b.name);
+    // return b_mtime - a_mtime;
+    const a_name = _get_filename(a.name);
+    const b_name = _get_filename(b.name);
+    if (a_name === b_name) {
+        const a_mtime = _get_mtime_from_filename(a.name);
+        const b_mtime = _get_mtime_from_filename(b.name);
+        return b_mtime - a_mtime;
```
dannyzaken, line 114: I'm not sure native subtraction will work with BigInt. you can use `copmare`
```suggestion
        return b_mtime.compare(a_mtime);
```

src/sdk/namespace_fs.js
v-harihar: Yeah, did not consider that case.
sure, I will make a note of it

src/sdk/namespace_fs.js
```diff
@@ -79,6 +79,46 @@ function sort_entries_by_name(a, b) {
...
+}
+/**
+ * @param {fs.Dirent} a
+ * @param {fs.Dirent} b
+ * @returns {Number}
+ */
+function sort_entries_by_name_and_time(a, b) {
+    // const a_mtime = _get_mtime_from_filename(a.name);
+    // const b_mtime = _get_mtime_from_filename(b.name);
+    // return b_mtime - a_mtime;
```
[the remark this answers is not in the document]
v-harihar: sure

src/sdk/namespace_fs.js
v-harihar: This is also not supported with BigInt()

src/sdk/namespace_fs.js
v-harihar: Checking for BigInt supported operations

src/sdk/namespace_fs.js
v-harihar: sure

src/sdk/namespace_fs.js
```diff
@@ -79,6 +79,46 @@ function sort_entries_by_name(a, b) {
...
 
+function _get_mtime_from_filename(filename) {
+    if (!filename.includes('_mtime-')) {
+        // Latest file wont have time suffix which will push the latest
+        // object last in the list. So to keep the order maintained,
+        // returning the latest time. Multiplying with 1e6 to provide
+        // nano second precision
+        return Date.now() * 1e6;
+    }
+    const file_parts = filename.split('-');
```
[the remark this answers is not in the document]
v-harihar: Yes, I should consider it from the end of the string to fetch the mtime

src/sdk/namespace_fs.js
v-harihar: sure

src/sdk/namespace_fs.js
```diff
@@ -482,6 +609,7 @@ class NamespaceFS {
                     }
                     for (let i = marker_index; i < sorted_entries.length; ++i) {
                         const ent = sorted_entries[i];
+                        if (list_versions && !ent.name.includes(marker_curr)) break;
```
romayalon, line 612: can you explain this condition? what happens if marker_curr is 'c' but object name is 'ac'?

src/sdk/namespace_fs.js
```diff
@@ -459,11 +570,27 @@ class NamespaceFS {
...
-                        make_named_dirent(marker_curr),
-                        get_entry_name
-                    );
+                    let marker_index;
+                    // Two ways followed here to find the index.
+                    // 1. When inside marker_dir: Here the entries are sorted based on time. Here
+                    //    FindIndex() is called since sortedLastIndexBy() expects sorted order by name
+                    // 2. When marker_dir above dir_path: sortedLastIndexBy() is called since entries are
+                    //     sorted by name
+                    if (list_versions && marker_curr && !marker_curr.includes('/')) {
```
romayalon, line 579: shouldn't we use _.findIndex() only when version_id_marker passed?

src/sdk/namespace_fs.js
```diff
@@ -536,15 +663,31 @@ class NamespaceFS {
...
                     res.common_prefixes.push(r.key);
                 } else {
-                    res.objects.push(this._get_object_info(bucket, r.key, r.stat, 'null'));
+                    obj_info = this._get_object_info(bucket, r.key, r.stat, 'null', true);
+                    if (!list_versions && obj_info.delete_marker) {
+                        continue;
+                    }
+                    if (this._is_hidden_version_path(obj_info.key)) {
+                        obj_info.key = path.normalize(obj_info.key.replace(HIDDEN_VERSIONS_PATH + '/', ''));
+                        obj_info.key = obj_info.key.substring(0, obj_info.key.indexOf("_mtime-"));
```
romayalon, line 679:
```suggestion
                        obj_info.key = obj_info.key.substring(0, obj_info.key.indexOf('_mtime-'));
```

src/sdk/namespace_fs.js
```diff
@@ -312,6 +407,14 @@ class NamespaceFS {
      * @param {ListParams} params
      */
     async list_objects(params, object_sdk) {
+        return this._list_objects(params, object_sdk, false);
+    }
+
+    async list_object_versions(params, object_sdk) {
```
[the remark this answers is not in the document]
v-harihar: sure

src/sdk/namespace_fs.js
v-harihar: When only `key_marker` is provided then we got to list the versions of all the versions of that `key_marker`. So in this case, assigning `start_marker` to `marker_curr`. If `version_id_marker` is also provided then assigning `version_id_marker` to `marker_curr`

Correct me if wrong

src/sdk/namespace_fs.js
v-harihar: agree. It should be `match` not `includes`.

When the `key_marker` is provided then we only have to list the versions of that marker. So breaking here if the key is other than `key_marker`

src/sdk/namespace_fs.js
v-harihar: sure

src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -27,6 +27,7 @@ const DEFAULT_FS_CONFIG = {
     backend: '',
     warn_threshold_ms: 100,
 };
+coretest.setup({ pools_to_create: [coretest.POOL_LIST[1]] });
```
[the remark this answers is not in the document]
v-harihar: sure

src/sdk/namespace_fs.js
```diff
@@ -79,6 +79,45 @@ function sort_entries_by_name(a, b) {
...
+ * @param {fs.Dirent} b
+ * @returns {Number}
+ */
+function sort_entries_by_name_and_time(a, b) {
+    const a_name = _get_filename(a.name);
+    const b_name = _get_filename(b.name);
+    if (a_name === b_name) {
+        const a_mtime = _get_mtime_from_filename(a.name);
+        const b_mtime = _get_mtime_from_filename(b.name);
+        if (b_mtime < a_mtime) return -1;
```
romayalon, line 111: when a_mtime < b_mtime shouldn't we return -1?

src/sdk/namespace_fs.js
romayalon: got you, right 
and can you please add a comment about this case - `&& !marker_curr.includes('/')` 
not sure I understand it

src/sdk/namespace_fs.js
```diff
@@ -536,15 +679,31 @@ class NamespaceFS {
...
+                        obj_info.key = obj_info.key.substring(0, obj_info.key.indexOf('_mtime-'));
+                    }
+                    res.objects.push(obj_info);
                 }
                 if (res.is_truncated) {
-                    res.next_marker = r.key;
+                    if (list_versions && r.key.includes('_mtime-')) {
+                        const next_version_id_marker = r.key.substring(r.key.lastIndexOf('/') + 1);
+                        res.next_version_id_marker = next_version_id_marker;
+                        res.next_marker = key_marker;
```
romayalon, line 703: why the next_marker is the key_marker we got as parameter?

src/sdk/namespace_fs.js
v-harihar: With this, the list will be arranged in the latest first order

romayalon: right, can you add a comment on it please?

src/sdk/namespace_fs.js
v-harihar: This will make sure that the path does not include `/`. 

ex:  if `key_marker = dir1/file1`, we call `findIndex()` only when we get at file level i.e when `marker_curr = file1`. `marker_curr` is directory by directory search and traverses through `key_marker` path. We dont call findIndex() when `marker_curr = dir1/file1` instead call goest to else part.

src/sdk/namespace_fs.js
v-harihar: sure

src/sdk/namespace_fs.js
```diff
@@ -79,6 +79,45 @@ function sort_entries_by_name(a, b) {
...
+        return file_name.substring(0, file_name.indexOf('_mtime-'));
+    }
+    return file_name;
+}
+/**
+ * @param {fs.Dirent} a
+ * @param {fs.Dirent} b
+ * @returns {Number}
+ */
+function sort_entries_by_name_and_time(a, b) {
```
liranmauda, line 105: please use meaningful variable names.

src/sdk/namespace_fs.js
v-harihar: I think I need to fetch the latest_key name and assign to next_key here

`res.next_marker = next_version_id_marker.substring(0, next_version_id_marker.indexOf('_mtime'));`

src/sdk/namespace_fs.js
```diff
@@ -1,5 +1,6 @@
 /* Copyright (C) 2020 NooBaa */
-/*eslint max-lines: ["error", 2500]*/
+/*eslint max-lines: ["error", 3000]*/
+/*eslint max-statements: ["error", 80, { "ignoreTopLevelFunctions": true }]*/
```
liranmauda, line 3: I wonder, isn't there a better way then adding those?

src/endpoint/s3/ops/s3_get_bucket_versions.js
```diff
@@ -17,6 +17,11 @@ async function get_bucket_versions(req) {
         throw new S3Error(S3Error.InvalidArgument);
     }
 
+    if (req.query['version-id-marker'] && !req.query['key-marker']) {
+        dbg.warn('A version-id marker cannot be specified without a key marker');
+        throw new S3Error(S3Error.InvalidArgument);
+    }
+
```
liranmauda, line 24: Maybe worth taking it into a different PR and adding a test for it.

src/sdk/namespace_fs.js
liranmauda: variables... 
don't use `a` and `b`.
and corresponding in the function itself, `a_name` 
rather be something like `first_entry` `second_entry` or something similar

v-harihar: sure

dannyzaken (dismissed):

src/endpoint/s3/ops/s3_get_bucket_versions.js
v-harihar: sure

dannyzaken: @v-harihar @liranmauda I don't think it is worth a different PR or test, it's a small change. 

we can have it maybe in another commit in this PR, so we can backport it separately if needed. make sense?

v-harihar: sure. Adding the test in another PR in the same commit

src/sdk/namespace_fs.js
```diff
@@ -424,26 +539,37 @@ class NamespaceFS {
...
                     }
-
                     const isDir = await is_directory_or_symlink_to_directory(ent, fs_context, path.join(dir_path, ent.name));
 
-                    const r = {
-                        key: this._get_entry_key(dir_key, ent, isDir),
-                        common_prefix: isDir,
-                    };
+                    let r;
+                    if (list_versions && ent.name.includes('_mtime-')) {
```
romayalon, line 552: it's a bit overkill but in case the object name includes '_mtime-' and it's not a version we fail here

src/sdk/namespace_fs.js
romayalon: ok, can you add a comment on that case?

src/sdk/namespace_fs.js
```diff
@@ -485,6 +627,10 @@ class NamespaceFS {
                     }
                     for (let i = marker_index; i < sorted_entries.length; ++i) {
                         const ent = sorted_entries[i];
+                        if (list_versions && marker_curr) {
+                            const ent_name = ent.name.substring(0, ent.name.indexOf('_mtime-'));
```
romayalon, line 631: here you can call get_file_name() instead

src/sdk/namespace_fs.js
```diff
@@ -199,6 +240,62 @@ const dir_cache = new LRUCache({
...
+            /*eslint no-unused-expressions: ["error", { "allowTernary": true }]*/
+            for (const ent of sorted_entries) {
+                usage += ent.name.length + 4;
+            }
+        }
+        return { time, stat, sorted_entries, usage };
+    },
+    validate: async ({ stat }, { dir_path, fs_context }) => {
+        const new_stat = await nb_native().fs.stat(fs_context, dir_path);
+        return (new_stat.ino === stat.ino && new_stat.mtimeNsBigint === stat.mtimeNsBigint);
```
romayalon, line 293: do we need here validation of ver_dir_stat_size as well?

src/sdk/namespace_fs.js
```diff
@@ -462,11 +588,27 @@ class NamespaceFS {
...
+                    //    FindIndex() is called since sortedLastIndexBy() expects sorted order by name
+                    // 2. When marker_dir above dir_path: sortedLastIndexBy() is called since entries are
+                    //     sorted by name
+                    if (list_versions && marker_curr && !marker_curr.includes('/')) {
+                        let start_marker = marker_curr;
+                        if (version_id_marker) start_marker = version_id_marker;
+                        marker_index = _.findIndex(
+                            sorted_entries,
+                            {name: start_marker}
+                        ) + 1;
```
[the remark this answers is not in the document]
v-harihar: We usually list versions after start_marker (not including the start_marker). So marker_index will be start_marker + 1

src/sdk/namespace_fs.js
```diff
@@ -539,15 +684,31 @@ class NamespaceFS {
...
                     res.common_prefixes.push(r.key);
                 } else {
-                    res.objects.push(this._get_object_info(bucket, r.key, r.stat, 'null'));
+                    obj_info = this._get_object_info(bucket, r.key, r.stat, 'null', true);
+                    if (!list_versions && obj_info.delete_marker) {
+                        continue;
+                    }
+                    if (this._is_hidden_version_path(obj_info.key)) {
+                        obj_info.key = path.normalize(obj_info.key.replace(HIDDEN_VERSIONS_PATH + '/', ''));
+                        obj_info.key = obj_info.key.substring(0, obj_info.key.indexOf('_mtime-'));
```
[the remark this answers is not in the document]
v-harihar: sure

src/sdk/namespace_fs.js
v-harihar: sure

src/sdk/namespace_fs.js
```diff
@@ -539,15 +684,31 @@ class NamespaceFS {
...
+                    }
+                    res.objects.push(obj_info);
                 }
                 if (res.is_truncated) {
-                    res.next_marker = r.key;
+                    if (list_versions && r.key.includes('_mtime-')) {
+                        const next_version_id_marker = r.key.substring(r.key.lastIndexOf('/') + 1);
+                        res.next_version_id_marker = next_version_id_marker;
+                        res.next_marker = next_version_id_marker.substring(0, next_version_id_marker.indexOf('_mtime'));
```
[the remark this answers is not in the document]
v-harihar: sure

src/sdk/namespace_fs.js
v-harihar: May be we can consider it in separate PR to optimize it ? I will open up PR for this.

dannyzaken: I'm fine with handling it in a different PR, as long as there are no regressions

romayalon (approved):

romayalon:
Approved but -
1. Please squash the commits
2. Please add the gaps in the next pr

PR #7208, opened 2023-02-14
use service certificate for rpc_http_server

jackyalbo:

### Explain the changes
1. We were using our own certificate instead of the service one for all the rpc servers (bg_workers and hosted_agents)
2. Every attempt to reach this servers in openshift caused error of self sign certificate - now suppose to be fixed

1. Fixed https://bugzilla.redhat.com/show_bug.cgi?id=2168867

### Testing Instructions:
1.  use curl from endpoint/core:
 curl --cacert /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt https://noobaa-mgmt.openshift-storage.svc:8445/.    *** or 8446 for hosted_agents

See that you don't get this result:
curl: (60) SSL certificate problem: self signed certificate ===> fails

instead you should get an empty response

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/rpc/rpc_http_server.js
+++ b/src/rpc/rpc_http_server.js
@@ -42,14 +42,15 @@ class RpcHttpServer extends events.EventEmitter {
     /**
      * start_server
      */
-    start_server(options) {
+    async start_server(options) {
         const port = parseInt(options.port, 10);
         const secure = options.protocol === 'https:' || options.protocol === 'wss:';
         const logging = options.logging;
         dbg.log0('HTTP SERVER:', 'port', port, 'secure', secure, 'logging', logging);
 
+        const ssl_cert = await ssl_utils.get_ssl_certificate('MGMT');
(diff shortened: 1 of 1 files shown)


liranmauda (approved):
LGTM,
Just one comment, can we change the function name from`start_server` to `rpc_start_server`? currently, it is confusing as we also have `start_server` in the `prometheus_reporting.js`

jackyalbo:
> LGTM, Just one comment, can we change the function name from`start_server` to `rpc_start_server`? currently, it is confusing as we also have `start_server` in the `prometheus_reporting.js`

I don't think we should do it in this PR anyway, but I think you can understand what server you are starting from the context, as long as you give the server a proper name.

for me, rpc_http_server.start_server, sound better than rpc_http_server.rpc_start_server

and anyway prometheus_reporting is newer, so maybe we should change there :))

PR #7212, opened 2023-02-16
# BZ 2168010 | Bytes used by Noobaa Bucket in Prometheus metrics

naveenpaul1:
### Explain the changes
1.  showing the number of bytes used by Noobaa Bucket in Prometheus metrics

1. https://bugzilla.redhat.com/show_bug.cgi?id=2168010#

### Testing Instructions:
1. add one or more buckets to noobaa
3. upload data to that bucket and check for `NooBaa_objectbucket_used_bytes` in Prometheus it should return all the buckets with bytes used by each bucket

- [ ] Doc added/updated
- [ ] Tests added

Signed-off-by: Naveen Paul @naveenpaul1

### DIFF
--- a/src/api/stats_api.js
+++ b/src/api/stats_api.js
@@ -758,7 +758,7 @@ module.exports = {
                     type: 'array',
                     items: {
                         type: 'object',
-                        required: ['bucket_name', 'quota_size_precent', 'quota_quantity_percent', 'capacity_precent', 'is_healthy', 'tagging'],
+                        required: ['bucket_name', 'quota_size_precent', 'quota_quantity_percent', 'capacity_precent', 'is_healthy', 'tagging', 'bucket_used_bytes'],
                         properties: {
                             bucket_name: {
                                 type: 'string'
@@ -774,6 +774,9 @@ module.exports = {
                             },
                             is_healthy: {
                                 type: 'boolean'
+                            },
+                            bucket_used_bytes: {
+                                type: 'number'
                             },
                             tagging: {
                                 $ref: 'common_api#/definitions/tagging',
--- a/src/server/system_services/stats_aggregator.js
+++ b/src/server/system_services/stats_aggregator.js
@@ -518,7 +518,8 @@ async function _partial_buckets_info(req) {
                 capacity_precent: (is_capacity_relevant && bucket_total > 0) ? size_utils.bigint_to_json(bucket_used.multiply(100)
                     .divide(bucket_total)) : 0,
                 is_healthy: _.includes(OPTIMAL_MODES, bucket_info.mode),
-                tagging: bucket_info.tagging || []
+                tagging: bucket_info.tagging || [],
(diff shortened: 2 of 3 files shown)


src/server/system_services/stats_aggregator.js
```diff
@@ -518,7 +518,8 @@ async function _partial_buckets_info(req) {
                 capacity_precent: (is_capacity_relevant && bucket_total > 0) ? size_utils.bigint_to_json(bucket_used.multiply(100)
                     .divide(bucket_total)) : 0,
                 is_healthy: _.includes(OPTIMAL_MODES, bucket_info.mode),
-                tagging: bucket_info.tagging || []
+                tagging: bucket_info.tagging || [],
+                objectbucket_used_bytes: size_utils.bigint_to_json(bucket_used)
```
dannyzaken, line 522: as you probably noticed, `bucket_used` is of type `BigInteger`, which is a third-party class we are using.  
The `bigint_to_json` function is converting the BigInteger into types that we can store in the DB or serialize in the RPC. This function usually returns a primitive Number, but it can also return an object if the value is over 1 PetaBytes. In this case, the API validation will fail.  
I suggest returning the internal value of the BigInteger:

```suggestion
                objectbucket_used_bytes: bucket_used.value
```

This will return a primitive Number up to `Number.MAX_SAFE_INTEGER` which is ~7 PetaBytes

naveenpaul1: Ya that makes sense will update the code

src/api/stats_api.js
```diff
@@ -758,7 +758,7 @@ module.exports = {
                     type: 'array',
                     items: {
                         type: 'object',
-                        required: ['bucket_name', 'quota_size_precent', 'quota_quantity_percent', 'capacity_precent', 'is_healthy', 'tagging'],
+                        required: ['bucket_name', 'quota_size_precent', 'quota_quantity_percent', 'capacity_precent', 'is_healthy', 'tagging', 'objectbucket_used_bytes'],
```
liranmauda, line 761: I think that we don't need to state that it is `objectbucket` (similar to `ocs_objectbucket_used_bytes` for RADOS S3), I think we can simply call it `bucket_used_bytes` as we have `bucket_capacity` (which weirdly presents the percentage), `bucket_size_quota`, `bucket_tagging`, ect'
```suggestion
                        required: ['bucket_name', 'quota_size_precent', 'quota_quantity_percent', 'capacity_precent', 'is_healthy', 'tagging', 'bucket_used_bytes'],
```

WDYT?

liranmauda:
@naveenpaul1 Please also rebase from master and squash the commits. after that please sign the commit.

src/api/stats_api.js
naveenpaul1: I am fine with this, I gave this name to align with OCS metrics name mentioned in BZ
 @dannyzaken

liranmauda (approved):
LGTM

PR #7219, opened 2023-02-27
Namespace GCP | Adding delete multiple

liranmauda:
### Explain the changes

- Adding delete multiple

### Testing Instructions:
1.  use `aws s3api --endpoint-url=<url:port> delete-objects --bucket <bucket_name> --delete '{"Objects": [{"Key": "<keyname>"}, {"Key": "<keyname>"}]}' --no-verify-ssl`

Note: in order to see the response we need to do:
1. `aws configure`
2. Change the `Default output format [None]:` to json

### DIFF
--- a/src/sdk/namespace_gcp.js
+++ b/src/sdk/namespace_gcp.js
@@ -4,7 +4,7 @@
 const _ = require('lodash');
 const util = require('util');
 
-// const P = require('../util/promise');
+const P = require('../util/promise');
 const stream_utils = require('../util/stream_utils');
 const dbg = require('../util/debug_module')(__filename);
 const S3Error = require('../endpoint/s3/s3_errors').S3Error;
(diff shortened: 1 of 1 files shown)


src/sdk/namespace_gcp.js
```diff
@@ -234,7 +234,16 @@ class NamespaceGCP {
     async delete_multiple_objects(params, object_sdk) {
         // https://googleapis.dev/nodejs/storage/latest/File.html#delete
         dbg.log0('NamespaceGCP.delete_multiple_objects:', this.bucket, inspect(params));
-        throw new S3Error(S3Error.NotImplemented);
+
+        const res = await P.map_with_concurrency(10, params.objects, obj =>
```
dannyzaken, line 238: GCS seems to support [batch requests](https://cloud.google.com/storage/docs/batch), but it is not part of the js SDK. Maybe it's worth investing some time in working with it directly using REST API and not using the SDK (we might be able to take auth info from the SDK).  

batch requests will be very useful for many operations in the google NS. Specifically in this case it can reduce the number of requests significantly (`delete_multiple_objects` can contain up to 1000 objects)

src/sdk/namespace_gcp.js
```diff
@@ -234,7 +234,16 @@ class NamespaceGCP {
...
+            this.gcs.bucket(this.bucket).file(obj.key).delete()
+            .then(() => ({}))
```
dannyzaken, line 240: Shouldn't the response be translated to the expected result of s3 op?

dannyzaken (approved):
I think that when the SDK supports batch requests we want to use it maybe open a tech debt issue on that

PR #7221, opened 2023-03-02
System Store | Move refresh() outside of make_changes_internal() 

### LINKED ISSUE #7220: RPC requests timeouts due to _serial_load semaphore deadlock
### Environment info
- NooBaa Version: 4.13
- Platform: OpenShift 4.13

While investigating https://bugzilla.redhat.com/show_bug.cgi?id=2172624 and searching for information/errors in the logs of all NooBaa's pods, we saw many RPC requests timeouts: 
a. Operator logs - full of pings from operator to core without answers.
b. Core logs - all heartbeat() calls throws RPC requests timeout  
c. Endpoint logs - NooBaa receives new s3 requests but doesn't return an answer.

After checking the code we realized that before processing RPC request, NooBaa first runs middleware that calls refresh() that calls load() in order to reload system store data. 
load() code is surrounded by _serial_load semaphore locking (semaphore counter is 1).
We ran a node debugger and saw that this semaphore lock is taken and not being released and it has thousands of items in its waiting queue.
 
The 2 points in NooBaa's code that lock this semaphore are:
   - make_changes() 
   - load()

While reading make_changes() code again, we noticed that inside make_changes_internal() we call refresh(). In a corner case** of refresh(), NooBaa calls load() which loads the system store data under the same lock, but this lock is already acquired by make_changes() - and that's the deadlock, for more info about this change see https://github.com/noobaa/noobaa-core/pull/6066

** corner cases of refresh that call load - 
1. load() is called and awaited when the system store data last load time > 1 hour ago
2. load() is called without awaiting when the system store data last load time > 10 minutes and < 60 minutes

### Actual behavior
1.  S3 service receives new requests without responding back, RPC requests timeouts all over noobaa's pods logs.

### Expected behavior
1. S3 service receives new requests and replies back, No RPC timeouts.

### Steps to reproduce
Not easy to reproduce on a real system, we barely saw this issue.
1. Change artificially START_REFRESH_THRESHOLD & FORCE_REFRESH_THRESHOLD in order to 0 and we should see this issue again.

Some thoughts about solutions:
1. Taking out the inner lock is the immediate solution.
3. Adding a timeout for this semaphore can be considered as well.

romayalon:
### Explain the changes
1. Moved refresh() outside of make_changes_internal() because both functions try to lock the same _serial_load semaphore (counter 1), this can cause a deadlock in several situations. 
Check the attached github issue for more info.

1. Fixed https://github.com/noobaa/noobaa-core/issues/7220
3. Fixed https://bugzilla.redhat.com/show_bug.cgi?id=2172624

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -589,8 +589,11 @@ class SystemStore extends EventEmitter {
      *
      */
     async make_changes(changes) {
+        // Refreshing must be done outside the semapore lock because refresh
+        // might call load that is locking on the same semaphore.
+        const data = await this.refresh();
         const { any_news, last_update } = await this._load_serial.surround(
-            () => this._make_changes_internal(changes)
+            () => this._make_changes_internal(data, changes)
         );
 
         // Reloading must be done outside the semapore lock because the load is
@@ -610,7 +613,7 @@ class SystemStore extends EventEmitter {
         }
     }
 
-    async _make_changes_internal(changes) {
+    async _make_changes_internal(data, changes) {
         const bulk_per_collection = {};
         const now = new Date();
         const last_update = now.getTime();
@@ -632,8 +635,6 @@ class SystemStore extends EventEmitter {
             bulk_per_collection[name] = bulk;
             return bulk;
         };
-
-        const data = await this.refresh();
 
         _.each(changes.insert, (list, name) => {
             const col = get_collection(name);
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -351,7 +351,7 @@ class SystemStore extends EventEmitter {
         this.is_finished_initial_load = false;
         this.START_REFRESH_THRESHOLD = 10 * 60 * 1000;
         this.FORCE_REFRESH_THRESHOLD = 60 * 60 * 1000;
-        this._load_serial = new Semaphore(1);
+        this._load_serial = new Semaphore(1, { warning_timeout: this.START_REFRESH_THRESHOLD });
         for (const col of COLLECTIONS) {
             db_client.instance().define_collection(col);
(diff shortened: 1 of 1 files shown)


src/server/system_services/system_store.js
```diff
@@ -589,8 +589,11 @@ class SystemStore extends EventEmitter {
      *
      */
     async make_changes(changes) {
+        // Refreshing must be done outside the semapore lock because refresh
+        // might call load that is locking on the same semaphore.
+        const data = await this.refresh();
```
guymguym, line 594: I think we don't want to save the data from before the lock and send it inside, because we might wait for the lock a bit, and the data we took might be stale.

I think we can call `await refresh()` outside the lock, and inside _make_changes_internal we would just take the latest data from `const data = this.data`. 

If we want to be super duper, we can add a check inside the lock that if `Date.now()-this.data.time > this.FORCE_REFRESH_THRESHOLD` then we warn/throw? (also if > START_REFRESH_THRESHOLD we better warn...)

guymguym: also, we can add a warning timeout to this semaphore like you suggested in the issue, just in case.

romayalon: that's right, thanks!

romayalon: I had many directions about what is a correct warning timeout but I left with thinking that in the case of semaphore count is 1, work timeout should be equal waiting outside timeout, so if we start warn after START_REFRESH_THRESHOLD we assume that load() work takes less than START_REFRESH_THRESHOLD, not sure about it, let me know what you think

jackyalbo (approved):

PR #7225, opened 2023-03-08
Namespace GCP | Adding upload_object (copy) and read object metadata

liranmauda:
### Explain the changes
- Adding the missing copy of upload_object
- Adding read object metadata
- Adding xattr to the object info

### Testing Instructions:
1. Copy a file from the gcp bucket: 
    for example: `aws s3 --endpoint-url=<url:port> cp s3://<namespace_bucket_name>/<file> s3://<namespace_bucket_name>/<file_copy> --no-verify-ssl`

Note: copy will also trigger read object metadata.


jackyalbo (approved):
LGTM

PR #7234, opened 2023-03-16
Fix Log Replication Deletions

tangledbytes:
### Explain the changes
This PR ensures that even if multiple actions are performed on an object, we process all of the actions instead of processing only the last one. Processing only the last action causes issues in the scenario when the logs are not chronologically ordered, for example: `DELETE /cat.jpeg` then `PUT /cat.jpeg`, when these 2 entries are part of the same candidate batch, NooBaa would just process the last `PUT` which means that the `DELETE` will never be processed.

PS: Added JS Docs to have the static analysis "safety net".

This issue was discovered by Sagi (QE) during happy path testing. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/bg_services/log_replication_scanner.js
+++ b/src/server/bg_services/log_replication_scanner.js
@@ -120,17 +120,38 @@ class LogReplicationScanner {
         }));
     }
 
+    /**
+     * process_candidates takes the source and destination buckets and the candidates
+     * and process them depending upon the kind of candidate.
+     * @param {*} src_bucket source bucket
+     * @param {*} dst_bucket destination bucket
+     * @param {nb.ReplicationLogCandidates} candidates 
+     */
     async process_candidates(src_bucket, dst_bucket, candidates) {
         const src_dst_objects_list = await this.head_objects(src_bucket.name, dst_bucket.name, candidates);
 
         for (const candidate of src_dst_objects_list) {
-            if (candidate.data.action === 'copy') {
+            await this.process_candidate(src_bucket, dst_bucket, candidate);
+        }
+    }
+
+    async process_candidate(src_bucket, dst_bucket, candidate) {
+        // each candidate represents several action over the same key
+        // that are sorted by time.
+        const items = candidate.data || [];
+
+        for (const item of items) {
+            if (item.action === 'copy') {
                 // TODO: Cannot do it yet because it seems that the AWS logs
                 // are not ordered by time
             }
 
-            if (candidate.data.action === 'delete') {
+            if (item.action === 'delete') {
                 await this.process_delete_candidate(src_bucket, dst_bucket, candidate);
+            }
+
+            if (item.action === 'conflict') {
+                await this.process_conflict_candidate(src_bucket, dst_bucket, candidate);
             }
         }
     }
@@ -138,6 +159,32 @@ class LogReplicationScanner {
     async process_delete_candidate(src_bucket, dst_bucket, candidate) {
         if (!candidate.src_object_info && candidate.dst_object_info) {
             await this.delete_object(dst_bucket.name, candidate.key);
+        }
+    }
+
+    async process_conflict_candidate(src_bucket, dst_bucket, candidate) {
+        // If the item is present in the source bucket but not in the destination bucket
+        // then it is 'copy' action
+        if (candidate.src_object_info && !candidate.dst_object_info) {
+            // call the process_candidate again but with the 'copy' action
+            await this.process_candidate(src_bucket, dst_bucket, {
+                key: candidate.key,
+                data: { action: 'copy', time: candidate.data.time },
+                src_object_info: candidate.src_object_info,
+                dst_object_info: candidate.dst_object_info,
+            });
+        }
+
+        // If the item is present in the destination bucket but not in the source bucket
+        // then it is 'delete' action
+        if (!candidate.src_object_info && candidate.dst_object_info) {
+            // call the process_candidate again but with the 'delete' action
+            await this.process_candidate(src_bucket, dst_bucket, {
+                key: candidate.key,
+                data: { action: 'delete', time: candidate.data.time },
+                src_object_info: candidate.src_object_info,
+                dst_object_info: candidate.dst_object_info,
+            });
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/server/bg_services/replication_log_parser.js
```diff
@@ -118,16 +131,18 @@ function aws_parse_log_object(candidates_map, log_object, ignore_fn) {
                 if (ignore_fn(log)) continue;
 
                 if (log.operation.includes('PUT.OBJECT') || log.operation.includes('POST.OBJECT')) {
-                    candidates_map[log.key] = {
+                    candidates.push({
```
romayalon, line 134: do you also need to sort actions of the same key by the time in order to make it work correctly?
I wonder if we could do this kind of sorting and then skip redundant actions

src/server/bg_services/log_replication_scanner.js
```diff
@@ -120,6 +120,13 @@ class LogReplicationScanner {
         }));
     }
 
+    /**
+     * process_candidates takes the source and destination buckets and the candidates
+     * and process them depending upon the kind of candidate.
+     * @param {*} src_bucket source bucket
+     * @param {*} dst_bucket destination bucket
+     * @param {Array<{ key: string, action: 'copy' | 'delete', time: Date }>} candidates
```
romayalon, line 128: this row repeats 4 times, can you define this type?

utkarsh-pro: Makes sense. Should I add it to `nb.d.ts`?

src/server/bg_services/replication_log_parser.js
utkarsh-pro: Sorting by time could be helpful if we can perform more than 1 action. Right now the only truly actionable "action" is deletion. We won't have redundant actions because there can be only 1 successful `DELETE /<key> 204`.

But I think you are right, sorting + the key based approach looks a bit more elegant and makes more sense.

Making this change, thank you! 🙌

src/server/bg_services/log_replication_scanner.js
romayalon: hmmm I think you can

romayalon:
@utkarsh-pro https://github.com/noobaa/noobaa-core/pull/7235 is this PR will verify these changes?

src/server/bg_services/replication_log_parser.js
romayalon: yes I forgot about the fact that currently we don't support PUT,
for the future it's probably needed

src/server/bg_services/log_replication_scanner.js
utkarsh-pro: Added types ion `nb.d.ts`.

src/server/bg_services/replication_log_parser.js
utkarsh-pro: I have made the changes. The candidates are sorted by time and if there are conflicting actions (`copy` and `delete` happening at the same time) then the conflict is resolved separately). This would ensure that the sequence of operations that we perform will be correct.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -141,6 +156,32 @@ class LogReplicationScanner {
         }
     }
 
+    async process_conflict_candidate(src_bucket, dst_bucket, candidate) {
+        // If the item is present in the source bucket but not in the destination bucket
+        // then it is 'copy' action
```
romayalon, line 161: we should handle also the following scenario - if (candidate.src_object_info && candidate.dst_object_info)
when the source object is newer than the destination

src/server/bg_services/replication_log_parser.js
```diff
@@ -63,6 +66,45 @@ async function get_aws_log_candidates(source_bucket_id, rule_id, replication_con
...
+        if (!candidates_map.has(time)) {
+            candidates_map.set(time, {});
+        }
+
+        const time_map = candidates_map.get(time);
+
+        // if there is already a candidate for the same key, then we have a conflict
+        if (time_map[key]) {
+            const conflict_candidate = time_map[key];
+            conflict_candidate.action = 'conflict';
```
romayalon, line 96: should we avoid here versioning enabled buckets? this "squash" will not create the same versions as we have in the source bucket

src/server/bg_services/replication_log_parser.js
```diff
@@ -18,19 +18,22 @@ const moment = require('moment');
...
 async function get_aws_log_candidates(source_bucket_id, rule_id, replication_config, candidates_limit, ignore_fn) {
-    let aws_log_replication_info = replication_config.log_replication_info.aws_log_replication_info;
+    const aws_log_replication_info = replication_config.log_replication_info.aws_log_replication_info;
     const { logs_bucket, prefix } = aws_log_replication_info.logs_location;
-    let continuation_token = get_continuation_token_for_rule(rule_id, replication_config);
     const s3 = get_source_bucket_aws_connection(source_bucket_id, aws_log_replication_info);
+    let continuation_token = get_continuation_token_for_rule(rule_id, replication_config);
 
-    let candidates_obj = {};
+    const candidates = [];
```
romayalon, line 36: Why do you keep using an array approach instead of having an object of the following form - { key: [{time: Date, action: string] }, sorting the inner arrays and then skipping situations like put,put,delete that happened in different timestamps.

src/server/bg_services/log_replication_scanner.js
utkarsh-pro: Actually I knowingly left out this scenario in current as well as previous PR as it is handled already right now by the scan based replication.

Maybe we can bring this change in another PR as it is not specific to the fix? WDYT?

src/server/bg_services/replication_log_parser.js
utkarsh-pro: Yes, I had questions about versioned objects when I picked up the task. Unfortunately, we don't yet have a "doc" yet (not that I know of) which clarifies how we deal with the cases like:
- Source has versioning destination doesn't.
- Destination has versioning source doesn't.
- Both have versioning.

eg. What to do with delete markers when destination doesn't have versioning? Not that it can't be done, just that we don't have it documented yet.

This might be a broader discussion I think, that's why the implementation currently does not do anything special when it comes to versioned objects.

romayalon (approved):

src/server/bg_services/replication_log_parser.js
romayalon: Alright, for creating a better user experience I think that we should document it

src/server/bg_services/replication_log_parser.js
```diff
@@ -63,6 +66,55 @@ async function get_aws_log_candidates(source_bucket_id, rule_id, replication_con
...
+        const { key } = log;
+        if (!logs_per_key[key]) {
+            logs_per_key[key] = new Map();
+        }
+
+        const logs_for_key_time_map = logs_per_key[key];
+
+        // if there is already a candidate for the same key, then we have a conflict
+        if (logs_for_key_time_map.has(log.time)) {
+            // TODO: Object versioning will raise false alarms here
```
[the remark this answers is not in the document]
utkarsh-pro: @romayalon, opened an issue here: https://github.com/noobaa/noobaa-core/issues/7252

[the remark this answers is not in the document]
romayalon: perfect, thanks!

PR #7235, opened 2023-03-17
Add a basic AWS S3 server log parsing unittest

Neon-White:
### Explain the changes
1. A basic unittest that checks that log parsing works as expected was added

- [x] Tests added

### DIFF
--- a/src/test/unit_tests/test_bucket_replication.js
+++ b/src/test/unit_tests/test_bucket_replication.js
@@ -13,7 +13,7 @@ const http = require('http');
 const SensitiveString = require('../../util/sensitive_string');
 const cloud_utils = require('../../util/cloud_utils');
 const { ReplicationScanner } = require('../../server/bg_services/replication_scanner');
-
+const log_parser = require('../../server/bg_services/replication_log_parser');
 
 const db_client = require('../../util/db_client').instance();
 coretest.setup({ pools_to_create: [coretest.POOL_LIST[0]] });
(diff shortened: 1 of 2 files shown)


src/test/unit_tests/test_bucket_replication.js
```diff
@@ -920,4 +920,27 @@ mocha.describe('Replication pagination test', function() {
...
+        aaa test.bucket [13/Feb/2023:15:08:28 +0000] 1.1.1.1 arn:aws:iam::111:user/beneli AAA BATCH.DELETE.OBJECT text.txt - 204 - - 0 - - - - - AAA SigV4 ECDHE-RSA-AES128-GCM-SHA256 AuthHeader s3.us-east-2.amazonaws.com TLSv1.2 - -
+        `};
+        const action_dictionary = {'test': 'delete', 'test.js': 'delete', 'code2': 'copy', 'test2': 'delete', 'testfile.js': 'delete', 'empty': 'copy', 'text.txt': 'delete'};
+        log_parser.aws_parse_log_object(
+            candidate_map, example_log, function() { return false; }
+        );
+        for (const [object_name, value] of Object.entries(candidate_map)) {
```
romayalon, line 942: if the candidates map is empty that test will pass even though it should fail

src/test/unit_tests/test_bucket_replication.js
```diff
@@ -920,4 +920,30 @@ mocha.describe('Replication pagination test', function() {
...
         await list_objects_and_wait(s3_owner, target_bucket, obj_amount);
     });
+});
+
+mocha.describe('AWS S3 server log parsing tests', function() {
+    // Pagination test
+    mocha.it('Test AWS S3 server log parsing', async function() {
+        const logs = [];
+        const example_log = {Body: `
+        aaa test.bucket [13/Feb/2023:15:08:28 +0000] 1.1.1.1 arn:aws:iam::111:user/user AAA BATCH.DELETE.OBJECT test - 1 - - 1 - - - - - AAA SigV4 ECDHE-RSA-AES128-GCM-SHA256 AuthHeader s3.us-east-2.amazonaws.com TLSv1.2 - -
```
utkarsh-pro, line 930: I think we can also test for the following (as part of this PR or maybe future PRs):
1. If timestamp of a DELETE op on key `k` is greater than timestamp of a `PUT` op on key `k` **but** in the logs the `DELETE` shows up first - do we make sure that we process `DELETE` and not overwrite it with `PUT`? (Sagi found this)

dannyzaken (approved):

PR #7236, opened 2023-03-18
Calculate MD5 per bucket

aayushchouhan09:
### Explain the changes
1.  Added a flag to calculate MD5 calculation per bucket.

- [X] Tests added

### DIFF
--- a/src/api/bucket_api.js
+++ b/src/api/bucket_api.js
@@ -32,6 +32,7 @@ module.exports = {
                         type: 'boolean'
                     },
                     bucket_claim: { $ref: '#/definitions/bucket_claim' },
+                    force_md5_etag: { type: 'boolean' },
                 }
             },
             reply: {
@@ -874,6 +875,9 @@ module.exports = {
                 namespace: { $ref: '#/definitions/namespace_bucket_config' },
                 bucket_claim: { $ref: '#/definitions/bucket_claim' },
                 logging: {$ref: '#/definitions/logging'},
+                force_md5_etag: {
+                    type: 'boolean'
+                },
                 tiering: {
                     $ref: 'tiering_policy_api#/definitions/tiering_policy'
                 },
@@ -1177,6 +1181,9 @@ module.exports = {
                 name: { $ref: 'common_api#/definitions/bucket_name' },
                 new_name: { $ref: 'common_api#/definitions/bucket_name' },
                 tiering: { $ref: 'common_api#/definitions/tiering_name' },
+                force_md5_etag: {
+                    type: 'boolean'
+                },
                 new_tag: {
                     type: 'string',
                 },
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -214,6 +214,7 @@ class NamespaceFS {
      *  access_mode: string;
      *  versioning: 'DISABLED' | 'SUSPENDED' | 'ENABLED';
      *  stats: import('./endpoint_stats_collector').EndpointStatsCollector;
+     *  force_md5_etag: boolean;
      * }} params
      */
     constructor({
@@ -224,6 +225,7 @@ class NamespaceFS {
         access_mode,
         versioning,
         stats,
+        force_md5_etag,
     }) {
         dbg.log1('NamespaceFS: buffers_pool length',
             buffers_pool.buffers.length, buffers_pool.sem);
@@ -234,6 +236,7 @@ class NamespaceFS {
         this.access_mode = access_mode;
         this.versioning = (config.NSFS_VERSIONING_ENABLED && versioning) || versioning_status_enum.VER_DISABLED;
         this.stats = stats;
+        this.force_md5_etag = force_md5_etag;
     }
 
     /**
@@ -1050,11 +1053,13 @@ class NamespaceFS {
         const { source_stream } = params;
         try {
             // Not using async iterators with ReadableStreams due to unsettled promises issues on abort/destroy
+            const md5_enabled = config.NSFS_CALCULATE_MD5 || this.force_md5_etag;
             const chunk_fs = new ChunkFS({
                 target_file,
                 fs_context,
                 stats: this.stats,
                 namespace_resource_id: this.namespace_resource_id,
+                md5_enabled
             });
             chunk_fs.on('error', err1 => dbg.error('namespace_fs._upload_stream: error occured on stream ChunkFS: ', err1));
             await stream_utils.pipeline([source_stream, chunk_fs]);
@@ -1170,7 +1175,8 @@ class NamespaceFS {
         const fs_context = this.prepare_fs_context(object_sdk);
         const open_mode = 'w';
         try {
-            const MD5Async = config.NSFS_CALCULATE_MD5 ? new (nb_native().crypto.MD5Async)() : undefined;
+            const md5_enabled = config.NSFS_CALCULATE_MD5 || this.force_md5_etag;
+            const MD5Async = md5_enabled ? new (nb_native().crypto.MD5Async)() : undefined;
             const { multiparts = [] } = params;
             multiparts.sort((a, b) => a.num - b.num);
             await this._load_multipart(params, fs_context);
--- a/src/sdk/object_sdk.js
+++ b/src/sdk/object_sdk.js
@@ -301,7 +301,10 @@ class ObjectSDK {
                         ns: this._setup_single_namespace(
                             bucket.namespace.read_resources[0],
                             bucket._id,
-                            { versioning: bucket.bucket_info.versioning }
+                            {
+                                versioning: bucket.bucket_info.versioning,
+                                force_md5_etag: bucket.force_md5_etag
+                            },
                         ),
                         bucket,
                         valid_until: time + config.OBJECT_SDK_BUCKET_CACHE_EXPIRY_MS,
@@ -436,6 +439,7 @@ class ObjectSDK {
                 access_mode: r.access_mode,
                 versioning: options && options.versioning,
                 stats: this.stats,
+                force_md5_etag: options && options.force_md5_etag,
             });
         }
         // TODO: Should convert to cp_code and target_bucket as folder inside
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -210,6 +210,7 @@ async function create_bucket(req) {
             // TODO: Should implement validity checks
             bucket.bucket_claim = req.rpc_params.bucket_claim;
         }
+        bucket.force_md5_etag = req.rpc_params.force_md5_etag;
         changes.insert.buckets = [bucket];
         changes.insert.master_keys = [bucket_m_key];
 
@@ -606,6 +607,10 @@ function get_bucket_changes(req, update_request, bucket, tiering_policy) {
         get_bucket_changes_versioning(req, bucket, update_request, single_bucket_update, changes);
     }
 
+    if (!_.isUndefined(update_request.force_md5_etag)) {
+        single_bucket_update.force_md5_etag = update_request.force_md5_etag;
+    }
+
     if (!_.isUndefined(quota)) {
         get_bucket_changes_quota(req, bucket, quota, single_bucket_update, changes);
     }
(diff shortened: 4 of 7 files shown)


src/sdk/namespace_fs.js
```diff
@@ -1118,7 +1118,9 @@ class NamespaceFS {
         const fs_context = this.prepare_fs_context(object_sdk);
         const open_mode = 'w';
         try {
-            let MD5Async = config.NSFS_CALCULATE_MD5 ? new (nb_native().crypto.MD5Async)() : undefined;
+            const namespace_bucket_config = await object_sdk.read_bucket_sdk_namespace_info(params.bucket.name)
```
guymguym, line 1121: I don't like that we need to call this inside the namespace code itself.
I would expect that object_sdk would handle it and pass the option inside the params.

src/sdk/namespace_fs.js
```diff
@@ -1118,7 +1118,9 @@ class NamespaceFS {
...
+            let MD5enabled = namespace_bucket_config.write_resource.nsfs_md5_enabled 
+            let MD5Async = MD5enabled ? new (nb_native().crypto.MD5Async)() : undefined;
```
guymguym, line 1123: style: use `const` instead of `let` where possible, and use snake_case for identifier names.

guymguym (dismissed):

src/api/bucket_api.js
```diff
@@ -1373,7 +1373,8 @@ module.exports = {
             required: ['resource'],
             properties: {
                 resource: { type: 'string' },
-                path: { type: 'string' }
+                path: { type: 'string' },
+                nsfs_md5_enabled: { type: 'boolean' }
```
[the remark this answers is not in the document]
achouhan09: Okay, sounds good I will update the flag name.

src/sdk/namespace_fs.js
```diff
@@ -1118,7 +1118,9 @@ class NamespaceFS {
         const fs_context = this.prepare_fs_context(object_sdk);
         const open_mode = 'w';
         try {
-            let MD5Async = config.NSFS_CALCULATE_MD5 ? new (nb_native().crypto.MD5Async)() : undefined;
```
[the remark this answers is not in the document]
achouhan09: Sure, Thanks.

src/sdk/namespace_fs.js
achouhan09: ack , Thanks

src/api/bucket_api.js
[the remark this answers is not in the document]
dannyzaken: @guymguym do we want the same defaults for nsfs\noobaa buckets?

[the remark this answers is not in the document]
guymguym: Good question.
Today the default for backingstore buckets is true.
But the default for nsfs buckets is false.
We could add this option to account, and then based on the account that creates the bucket we will set the bucket option?
WDYT we should do?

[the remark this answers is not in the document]
dannyzaken: @guymguym  We do plan to also add an option to the account, but it would be optional, so we still need a default behavior. 

Several ways we can resolve it:
* We can decide to make the `calc_md5_etag` for one of those (bucket or account) mandatory, but it feels a bit too advanced to have as a mandatory field.
* Change the default for new "chunked buckets" (trying to push your term for it 😜) to false. For existing buckets, we can set `calc_md5_etag=true` on upgrades. This behavior might be a bit confusing.
* Have different defaults for different types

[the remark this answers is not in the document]
dannyzaken: Also, maybe the name for the flag should be `force_md5_etag`, since for some cases (S3 NS for example) we will not disable the md5 etag

src/server/system_services/schemas/bucket_schema.js
```diff
@@ -111,6 +111,7 @@ module.exports = {
                             objectid: true // namespace resource id
                         },
                         path: { type: 'string' },
+                        force_md5_etag: { type: 'boolean' }, // enable md5 calculation per bucket
```
dannyzaken, line 114: based on our previous discussion, this configuration should be at the top level of the bucket. it shouldn't be under `write_resource`

src/api/bucket_api.js
```diff
@@ -1373,7 +1373,8 @@ module.exports = {
             required: ['resource'],
             properties: {
                 resource: { type: 'string' },
-                path: { type: 'string' }
+                path: { type: 'string' },
+                force_md5_etag: { type: 'boolean' }
```
dannyzaken, line 1377: `force_md5_etag` should be added to get\set methods , and not `namespace_resource_config`. e.g.: `create_bucket`, `update_bucket_params`, `bucket_info`

src/endpoint/s3/ops/s3_post_object_uploads.js
dannyzaken, line 1: This op (`post_object_uploads`) is not relevant for md5 etag, it only starts a multipart upload, but no data is uploaded during this operation.
in any case, I don't think the handling of `force_md5_etag` should be added to the ops (rest) layer, only to the object_sdk layer. the relevant functions there are `upload_object`, `upload_multipart` and `complete_object_upload`.

src/sdk/namespace_fs.js
```diff
@@ -1006,11 +1006,13 @@ class NamespaceFS {
         const { source_stream } = params;
         try {
             // Not using async iterators with ReadableStreams due to unsettled promises issues on abort/destroy
+            const MD5Enabled = params.force_md5_etag
```
dannyzaken, line 1009: just to be consistent with our coding style 
```suggestion
            const md5_enabled = params.force_md5_etag
```

achouhan09: done, Thanks

src/server/system_services/bucket_server.js
```diff
@@ -539,6 +541,8 @@ async function read_bucket_sdk_info(req) {
             .then(get_bucket_info),
     };
 
+    reply.force_md5_etag = bucket.force_md5_etag;
```
dannyzaken, line 544: you should return `force_md5_etag` in `get_bucket_info` and not here. this implementation does not match the `bucket_api` definition (`force_md5_etag` is defined inder `bucket_info`)

dannyzaken (approved):

src/sdk/namespace_fs.js
```diff
@@ -1170,7 +1172,8 @@ class NamespaceFS {
         const fs_context = this.prepare_fs_context(object_sdk);
         const open_mode = 'w';
         try {
-            const MD5Async = config.NSFS_CALCULATE_MD5 ? new (nb_native().crypto.MD5Async)() : undefined;
+            const md5_enabled = params.force_md5_etag;
+            const MD5Async = md5_enabled ? new (nb_native().crypto.MD5Async)() : undefined;
```
dannyzaken, line 1176: Following up on @utkarsh-pro comment [here](https://github.com/noobaa/noobaa-core/pull/7297#discussion_r1191013617) on the account PR, maybe we should also keep the `config.NSFS_CALCULATE_MD5` as a global toggle. by default it should stay false, but it will allow us to turn it on globally when needed

src/test/unit_tests/test_system_servers.js
```diff
@@ -435,6 +436,36 @@ mocha.describe('system_servers', function() {
         });
     });
 
+    mocha.it('Calculate md5_etag for bucket works', async function() {
+        this.timeout(90000); // eslint-disable-line no-invalid-this
+        return P.resolve()
```
dannyzaken, line 441: I suggest writing it with `await` and not using a promise chain

src/test/unit_tests/test_system_servers.js
```diff
@@ -435,6 +436,36 @@ mocha.describe('system_servers', function() {
...
+            .then(() => rpc_client.bucket.create_bucket({
+                name: MD5_BUCKET,
+                force_md5_etag: true,
+            }))
+            .then(() => rpc_client.bucket.read_bucket({
+                name: MD5_BUCKET,
+            }))
+            .then(info => assert(info.force_md5_etag && info.force_md5_etag === true))
```
dannyzaken, line 449: I suggest writing it with await and not using a promise chain. it's much more readable

e.g.:
```suggestion
        await rpc_client.bucket.create_bucket({
             name: MD5_BUCKET,
             force_md5_etag: true,
         });
         const info = await rpc_client.bucket.read_bucket({
                name: MD5_BUCKET,
          })
          assert(info.force_md5_etag && info.force_md5_etag === true);
```

src/test/unit_tests/test_system_servers.js
```diff
@@ -435,6 +436,36 @@ mocha.describe('system_servers', function() {
...
+            .then(() => rpc_client.bucket.update_bucket({
+                name: MD5_BUCKET,
+                force_md5_etag: false,
+            }))
+            .then(() => rpc_client.bucket.read_bucket({
+                name: MD5_BUCKET,
+            }))
+            .then(info => assert(info.force_md5_etag && info.force_md5_etag === false))
```
dannyzaken, line 457: this condition will always be false. suggest changing to this


```suggestion
            .then(info => assert(info.force_md5_etag === false))
```

src/test/unit_tests/test_system_servers.js
achouhan09: Sure, Thanks.

src/test/unit_tests/test_system_servers.js
achouhan09: Okay, I will do the required changes here.

src/sdk/namespace_fs.js
```diff
@@ -1170,7 +1172,8 @@ class NamespaceFS {
         const fs_context = this.prepare_fs_context(object_sdk);
         const open_mode = 'w';
         try {
-            const MD5Async = config.NSFS_CALCULATE_MD5 ? new (nb_native().crypto.MD5Async)() : undefined;
+            const md5_enabled = params.force_md5_etag;
+            const MD5Async = (config.NSFS_CALCULATE_MD5 || md5_enabled) ? new (nb_native().crypto.MD5Async)() : undefined;
```
guymguym, line 1176: I would write
```suggestion
            const md5_enabled = config.NSFS_CALCULATE_MD5 || params.force_md5_etag;
            const MD5Async = md5_enabled ? new (nb_native().crypto.MD5Async)() : undefined;
```

src/sdk/object_sdk.js
```diff
@@ -160,6 +160,11 @@ class ObjectSDK {
         return bucket.namespace;
     }
 
+    async read_force_md5_etag(name) {
+        const { bucket } = await bucket_namespace_cache.get_with_cache({ sdk: this, name });
+        return bucket.force_md5_etag;
+    }
+
```
guymguym, line 167: I don't think it makes sense that we'll add a new function for every new bucket option...
Can we instead add it to the namespace object when we initialize it?

src/sdk/namespace_nb.js
```diff
@@ -129,6 +129,7 @@ class NamespaceNB {
             active_triggers,
             operation
         });
+        params = _.omit(params, 'force_md5_etag');
```
guymguym, line 132: why is this omitted from here? we want to also use it inside...

src/server/system_services/bucket_server.js
```diff
@@ -606,6 +607,10 @@ function get_bucket_changes(req, update_request, bucket, tiering_policy) {
         get_bucket_changes_versioning(req, bucket, update_request, single_bucket_update, changes);
     }
 
+    if (!_.isUndefined(update_request.force_md5_etag)) {
```
guymguym, line 610: I don't see why we need to be explicit about undefined, we can just check with regular if
```suggestion
    if (update_request.force_md5_etag) {
```

achouhan09: I think regular if will not work here, as `force_md5_etag` is a boolean flag and if the value of `update_request.force_md5_etag` is `false` then the condition is not met and value is not updated for `single_bucket_update.force_md5_etag`

src/sdk/namespace_nb.js
dannyzaken: for this PR we will only use this flag for NSFS. we can add it in nb buckets later

src/sdk/object_sdk.js
dannyzaken: @achouhan09 you can propagate it to NamespaceFS [here](https://github.com/achouhan09/noobaa-core/blob/030b5603ac43611823d0c20cfd8261af8604caf9/src/sdk/object_sdk.js#L436)

achouhan09: Done, Thanks

PR #7241, opened 2023-03-22
endpoint forks revamped

guymguym:
### Explain the changes
1. Allow running endpoints with multiple workers (forks) that share the same ports and respond to incoming http request as a "cluster".
2. This is needed when running endpoints in standalone mode for performance. 
3. This is limited to a single host.
4. Currently this is off by default.
5. When on kubernetes, this is done by starting multiple endpoint pods with a service on top, which is also a multi-host service, but for standalone this is the simplest way to get good performance from a host.

1. NA

### Testing Instructions:
1. NA

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/fork_utils.js
+++ b/src/util/fork_utils.js
@@ -1,0 +1,42 @@
+/* Copyright (C) 2023 NooBaa */
+'use strict';
+
+// these type hacks are needed because the type info from require('node:cluster') is incorrect
+const cluster = /** @type {import('node:cluster').Cluster} */ (
+    /** @type {unknown} */ (require('node:cluster'))
+);
+
+/**
+ * The cluster module allows easy creation of child processes that all share server ports.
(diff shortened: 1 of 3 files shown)


src/util/fork_utils.js
```diff
@@ -0,0 +1,42 @@
...
+ * When count > 1 the primary process will fork worker processes to process incoming http requests.
+ * In case of any worker exit, also the entire process group will exit.
+ * @see https://nodejs.org/api/cluster.html
+ * 
+ * @param {number?} count number of workers to start.
+ * @returns {boolean} true if workers were started.
+ */
+function start_workers(count = 0) {
+
+    if (cluster.isPrimary && count > 1) {
```
liranmauda, line 20: as discussed, use count > 0

liranmauda (approved):

PR #7244, opened 2023-03-26
endpoint stats refactor

guymguym:
### Explain the changes
1. endpoint_stats_collector - print stats to log, and extract triggering and error handling to utils:
   1. DelayedTrigger util - a robust trigger util so we don't miss updates.
   1. DelayedCollector util - handles errors by merging stats so we don't override updates.
1. object_sdk - cleanups and refactor common stats calling code (_call_op_and_update_stats)
1. pass stats instance to namespaces so they don't need rpc_client and it can also be null for tests.
1. namespace_cache fix calls to update_hub_latency_stats.

1. NA

### Testing Instructions:
1. NA


src/sdk/endpoint_stats_collector.js
```diff
@@ -1,258 +1,289 @@
...
+     * @param {NsfsStats} data
+     * @returns {Promise<void>}
+     */
+    async _process_nsfs_stats(data) {
+        dbg.log0('nsfs stats - IO counters :', data.io_stats);
+        for (const [k, v] of Object.entries(data.op_stats ?? {})) {
+            dbg.log0(`nsfs stats - S3 op=${k} :`, v);
+        }
+        for (const [k, v] of Object.entries(data.fs_workers_stats ?? {})) {
+            dbg.log0(`nsfs stats - FS op=${k} :`, v);
```
liranmauda, line 150: Do we want them to be always printed? 
Maybe we want it to be printed only on a higher debug level?

WDYT?

src/sdk/endpoint_stats_collector.js
```diff
@@ -1,258 +1,289 @@
...
-        } else {
-            io_stats.read_count += count;
-            io_stats.read_bytes += size;
-        }
+    update_namespace_read_stats({ namespace_resource_id, bucket_name = undefined, size = 0, count = 0, is_err = false }) {
+        this.endpoint_stats_collector.update({
+            namespace_stats: {
+                [namespace_resource_id]: is_err ? {
+                    error_read_count: count,
+                    error_read_bytes: size,
```
liranmauda, line 184: I think we will never get is_err true

src/sdk/endpoint_stats_collector.js
```diff
@@ -1,258 +1,289 @@
...
-        } else {
-            io_stats.write_count += count;
-            io_stats.write_bytes += size;
-        }
+    update_namespace_write_stats({ namespace_resource_id, bucket_name = undefined, size = 0, count = 0, is_err = false }) {
+        this.endpoint_stats_collector.update({
+            namespace_stats: {
+                [namespace_resource_id]: is_err ? {
+                    error_write_count: count,
+                    error_write_bytes: size,
```
liranmauda, line 201: I think we will never get is_err true

liranmauda (approved):
Just some comments, 
Beside, LGTM

src/util/delayed_trigger.js
```diff
@@ -0,0 +1,82 @@
...
+        this._on_trigger = on_trigger;
+        this._max_retries = max_retries;
+        this._num_retries = 0;
+        this._timer = null;
+        this._sem = new Semaphore(1);
+    }
+
+    trigger() {
+        // when the timer is set we already have a trigger in place
+        if (this._timer) return;
```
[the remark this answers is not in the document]
guymguym: yes, when the timer is scheduled, for example if the delay is 10 seconds, then during those 10 seconds we will hit this first condition and bail, because the timer is already set. After 10 seconds the timer will be called and will unset this field and immediately go and acquire the semaphore. So unsetting the timer allows new stats updates to register a new timer.

src/util/delayed_trigger.js
```diff
@@ -0,0 +1,82 @@
...
+                await this._sem.surround(this._on_trigger);
+
+                // reset retries count on success
+                this._num_retries = 0;
+
+            } catch (err) {
+
+                // check if retries allow retrigger
+                this._num_retries += 1;
+                if (this._num_retries > this._max_retries) {
```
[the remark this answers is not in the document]
guymguym: since we increment it before the check, it's correct as it is now.

src/sdk/object_sdk.js
```diff
@@ -119,9 +138,13 @@ class ObjectSDK {
      * @returns {nb.BucketSpace}
      */
     _get_bucketspace() {
-        return this.bucketspace_fs || this.bucketspace_nb;
+        return this.bucketspace;
```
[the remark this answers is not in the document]
guymguym: not really, but I decided to leave it for now in case we want to change how bucketspace is resolved. leaving for next refactor :)

src/sdk/endpoint_stats_collector.js
guymguym: notice that only existing entries will print, so if there is no activity it is quiet.
I think this is useful - we don't log on every operation, so at least print periodic stats.

src/sdk/endpoint_stats_collector.js
guymguym: yeah I see this is not used right now, but still I left the original option in case we want to add error stats.

PR #7249, opened 2023-03-29
noobaa/core: Transaction related changes

aspandey:
This patch makes changes to send bulk queries for
multiple collections in one transaction for postgres DB. In case of failure of any query, all the bulk operations will be rollbacked.

unit test has not been added for this patch

### DIFF
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -746,9 +746,8 @@ class SystemStore extends EventEmitter {
             });
         });
 
-        const bulk_results = await Promise.all(Object.values(bulk_per_collection).map(
-            bulk => bulk.length && bulk.execute({ j: true })
-        ));
+
+        const bulk_results = await db_client.instance().execute_multiple_bulks(bulk_per_collection);
 
         for (const res of bulk_results) {
             if (res && !res.ok) {
@@ -756,6 +755,7 @@ class SystemStore extends EventEmitter {
                 throw new Error(res.err);
             }
         }
+
 
         return { any_news, last_update };
     }
--- a/src/test/unit_tests/test_system_store.js
+++ b/src/test/unit_tests/test_system_store.js
@@ -5,6 +5,8 @@ const mocha = require('mocha');
 const assert = require('assert');
 const coretest = require('./coretest');
 const db_client = require('../../util/db_client');
+const config = require('../../../config');
+const _ = require('lodash');
 
 // setup coretest first to prepare the env
 coretest.setup({ pools_to_create: [coretest.POOL_LIST[0]] });
@@ -136,6 +138,52 @@ mocha.describe('system_store', function() {
         const data2 = await system_store.load();
         console.log('new_data_store', data2.systems.length);
         assert.strictEqual(data2.systems[0].name, 'new_name');
+    });
+
+    mocha.it('Check make_changes could not insert same id twice', async function() {
+        if (config.DB_TYPE === 'mongodb') this.skip(); // eslint-disable-line no-invalid-this
+
+        const system_id = system_store.new_system_store_id();
+        const orig_name = `JenTheMajesticSlothSystemStoreLoop3`;
+        const test_name = `test_account_name`;
+        await system_store.load();
+
+        await system_store.make_changes({
+            insert: {
+                systems: [{
+                    _id: system_id,
+                    name: orig_name,
+                    owner: system_store.new_system_store_id(),
+                }]
+            }
+        });
+        try {
+            await system_store.make_changes({
+                insert: {
+                    systems: [{
+                        _id: system_id,
+                        name: 'new_name',
+                        owner: system_store.new_system_store_id(),
+                    }]
+                },
+                update: {
+                    accounts: [{
+                        _id: system_store.new_system_store_id(),
+                        name: test_name,
+                        email: 'test@noobaa.com',
+                        has_login: true,
+                    }]
+                }
+            });
+        } catch (err) {
+            console.log('Could not insert duplicate key :', err.message);
+        }
+        const data2 = await system_store.load();
+        console.log('new_data_store', data2);
+        console.log('account details : ', data2.accounts[0]);
+        assert(!(_.isEqual(data2.systems[0].name, 'new_name')));
+        assert(!(_.isEqual(data2.accounts[0].name, 'test_name')));
+
     });
 
 });
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -354,48 +354,7 @@ class PgTransaction {
         await this.query('COMMIT TRANSACTION');
     }
 
-    async rollback() {
-        await this.query('ROLLBACK TRANSACTION');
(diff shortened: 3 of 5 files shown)


aspandey:
@dannyzaken

src/server/system_services/system_store.js
```diff
@@ -38,9 +38,11 @@ const size_utils = require('../../util/size_utils');
 const os_utils = require('../../util/os_utils');
 const config = require('../../../config');
 const db_client = require('../../util/db_client');
+const pgc = require('../../util/postgres_client');
 
 const { RpcError } = require('../../rpc');
 const master_key_manager = require('./master_key_manager');
+const { ok } = require('assert');
```
liranmauda, line 45: I think it is leftover, should we remove it?

src/server/system_services/system_store.js
```diff
@@ -617,6 +619,8 @@ class SystemStore extends EventEmitter {
 
     async _make_changes_internal(changes) {
         const bulk_per_collection = {};
+        let bulk_results = [];
+        let pg_bulk_results = [];
```
liranmauda, line 623: Look like it is not beeing used.

src/util/postgres_client.js
```diff
@@ -478,6 +480,79 @@ class BulkOp {
...
+        ok = true;
+    } catch (err) {
+        errmsg = err;
+        dbg.error('PgTransaction execute error', err);
+        if (should_rollback) await transaction.rollback();
+    } finally {
+        transaction.release();
+    }
+    return [errmsg, ok];
+    // return {
```
liranmauda, line 525: can we clean this comment, or do we want to keep it?

liranmauda:
> Thanks @liranmauda for comments. This is work in progress. Will make these changes in next iteration.

Yes, noticed the draft after I reviewed 🙂

src/server/system_services/system_store.js
```diff
@@ -740,17 +744,27 @@ class SystemStore extends EventEmitter {
...
 
-        const bulk_results = await Promise.all(Object.values(bulk_per_collection).map(
-            bulk => bulk.length && bulk.execute({ j: true })
-        ));
 
-        for (const res of bulk_results) {
-            if (res && !res.ok) {
-                dbg.error('got error on bulk execute', res.err);
-                throw new Error(res.err);
+        if (config.DB_TYPE === 'postgres') {
```
dannyzaken, line 748: Maybe worth hiding the different implementations in the db clients. I suggest adding a function to `DBClient` interface in [nb.d.ts](https://github.com/noobaa/noobaa-core/blob/aeb5e82334ba1526cec2016ad698045e44ba02dc/src/sdk/nb.d.ts#L649), and then you can have a different implementation for postgres\mongo. Maybe something like `execute_multiple_bulks()`.
This function should return a common `bulk_results` array that should be inspected the same for both DBs

src/util/postgres_client.js
dannyzaken: in addition to my comment [here](https://github.com/noobaa/noobaa-core/pull/7249#discussion_r1155319425), we should return here the same structure that we return for mongo and handle it in a common code

src/server/system_services/system_store.js
aspandey: Done.  Good Idea.

src/util/postgres_client.js
aspandey: Done.

src/util/postgres_client.js
```diff
@@ -478,6 +479,87 @@ class BulkOp {
...
 }
 
+async function multi_collection_bulk(bulk_per_collection) {
+
+    const client = PostgresClient.instance();
+
+    const queries = _.flatten(Object.values(bulk_per_collection).map(bulk => bulk.get_queries()));
+    const transaction = new PgTransaction(client);
+
+    let ok = false;
```
romayalon, line 489: this is the same code as execute() of bulkOp class, I don't see a value in maintaining this code in 2 different functions, easily can move to a helper function

src/server/system_services/system_store.js
```diff
@@ -740,9 +740,8 @@ class SystemStore extends EventEmitter {
             });
         });
 
-        const bulk_results = await Promise.all(Object.values(bulk_per_collection).map(
-            bulk => bulk.length && bulk.execute({ j: true })
-        ));
+
+        const bulk_results = await db_client.instance().execute_multiple_bulks(bulk_per_collection);
```
romayalon, line 744: rollbacking all collections queries on failures of related collections in a single make_changes() can be a great value for our db validity, 
but it can cause unexpected failures if a single make_changes() updates/inserts/removes non related queries (rollback one query due to failure of non related query)
I would check that every call to make_changes() make sense in that way.

src/util/postgres_client.js
aspandey: Makes sense. Done.

src/server/system_services/system_store.js
aspandey: @romayalon,
I am not sure what could be not related queries in a single make_changes() function call.
On what basis we can say that two queries are related or not. This could be very complex to handle and could be bug prone if we do not have simple way of segregating queries. 
I have a limited understanding on this so could not visualize the tradeoff. 
@dannyzaken please provide your input as well.

For the refreshed patch, I have not made changes for this comment. Once we decide what to do next, I will send  modified code.

dannyzaken (commented):
@aspandey, I discussed with @romayalon, and we agreed this change is ok. We reviewed the make_changes calls, and it seems to behave better with this change. 
We also agreed that it makes sense to refactor the common code in postgers_client

romayalon:
Hi @aspandey,
As Danny said I was worried about some specific make_changes() calls that make multiple changes and the queries are not related to each other.
For example, something like - 
```
        update: {
            buckets: buckets_updates,
            accounts: accounts_updates,
            pools: pools_updates,
            namespace_resources: namespace_resources_updates
        },
   });
```
I just wanted to make sure we won't roll back queries for no reason because this is the side effect of making one transaction per single make_changes() call instead of per collection.

src/test/unit_tests/test_system_store.js
```diff
@@ -138,4 +138,37 @@ mocha.describe('system_store', function() {
...
+                    _id: system_id,
+                    name: orig_name,
+                    owner: system_store.new_system_store_id(),
+                }]
+            }
+        });
+        try {
+            await system_store.make_changes({
+                insert: {
+                    systems: [{
```
dannyzaken, line 158: For the second `make_changes` you should add a change\insert to different tables (pools, accounts, etc.). After the failure, make sure all changes were not applied. 
of course, this is the case for Postgres only, so we should skip this step for Mongo

src/server/system_services/system_store.js
```diff
@@ -757,6 +756,7 @@ class SystemStore extends EventEmitter {
             }
         }
 
+
```
dannyzaken, line 759: Unnecessary change.

PR #7255, opened 2023-04-02
support custom config for backing stores

guymguym:
### Explain the changes
1. support disabling chunk digest/encryption/compression.
2. support disabling md5/sha256 for etag.
3. support increasing max object part size.
4. support using erasure coding by default.

1. NA

### Testing Instructions:
1. NA

### DIFF
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -379,7 +379,7 @@ async function complete_object_upload(req) {
                         })`);
         }
     }
-    if (req.rpc_params.md5_b64 !== obj.md5_b64) {
+    if (req.rpc_params.md5_b64 && req.rpc_params.md5_b64 !== obj.md5_b64) {
         if (obj.md5_b64) {
             throw new RpcError('BAD_DIGEST_MD5',
                 'md5 on complete object differs from create object', {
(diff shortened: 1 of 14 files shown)


romayalon (approved):

src/server/object_services/object_server.js
```diff
@@ -1591,6 +1591,44 @@ function check_md_conditions(conditions, obj) {
...
+   const md5_b64 = updates?.md5_b64 || entity.md5_b64;
+   if (md5_b64) return Buffer.from(md5_b64, 'base64').toString('hex');
+
+   const sha256_b64 = updates?.sha256_b64 || entity.sha256_b64;
+   if (sha256_b64) return 'sha256-' + Buffer.from(sha256_b64, 'base64').toString('hex');
+
+   const id = updates?._id || entity._id;
+   if (id) return 'id-' + id.toHexString();
+
+   return '';
```
guymguym, line 1629: @romayalon I wonder if I should throw here instead of silently "failing" ... WDYT?

romayalon: Is there a situation where we get there? I couldn't find one and I thought that id is the default

guymguym: right, we shouldn't get there, this is why I thought it might be better to throw

guymguym: right, we shouldn't get there, this is why I thought it might be better to throw

PR #7256, opened 2023-04-02
NSFS | Unset supplemental groups of the endpoint & tests improvements 

romayalon:
### Explain the changes
1. Unset supplemental groups of the endpoint when switching user in fs_napi threads. (Opened a [gap](https://github.com/noobaa/noobaa-core/issues/7274) in order to support user's supplemental groups). This change was needed only in os_linux and was not needed in os_darwin. According to https://www.unix.com/man-page/osx/2/pthread_setugid_np/ the  library call that was used to switch the uid & gid also clears the supplementary groups
2. Root permissions tests - Added root user flag and moved it to makefile instead of running it using docker run command.
3. Added new test_nsfs_access.js file unit tests
4. Closed unclosed fd in tests and some other test cleanups

***More details about the issue caused this to happen -*** 
1. a user in linux, besides having uid and gid, has also “supplemental groups”,
gid is the user’s primary group but he can be a part of other groups.
2. So we started seeing that every user can access folders/files created by gid 0,
the reason for it is that in switch user we switched only the uid and gid, but we also needed to unset the supplemental groups (which was always 0). Since we didn’t do it we started seeing other users can access root’s folders

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added


src/native/util/os_linux.cpp
```diff
@@ -24,6 +26,13 @@ get_current_uid()
 const uid_t ThreadScope::orig_uid = getuid();
 const gid_t ThreadScope::orig_gid = getgid();
 
+const std::vector<gid_t> ThreadScope::orig_groups = [] {
```
jackyalbo, line 29: As discussed - add a comment here, or near the set/unset of the groups to explain why or what the code do what it does.

jackyalbo (approved):

src/native/util/os_linux.cpp
romayalon: Added

PR #7258, opened 2023-04-03
Handle invalid Azure object metadata keys

Neon-White:
### Explain the changes
1. The new code should adhere to [azcopy's method](https://github.com/Azure/azure-storage-azcopy/blob/108dbddd5adba1ea133ad5ab391c122bd71bf731/common/fe-ste-models.go#L1235-L1244) of resolving invalid object metadata keys in Azure

The key of the newly added header is `azure-metadata-handling`, and its possible values are:
- `RenameIfInvalid` - which will rename the key in order to preserve it (which is the default behavior if the header is not supplied)
- `ExcludeIfInvalid` - which will exclude the invalid tag
- `FailIfInvalid` - which will fail the upload

We support the same naming and recovery methodologies as Microsoft, meaning we won't be able to discern between tags whose renamed form cannot be restored (e.g. -  `!@#` and `#$%` on the same object)

1. Fixes object IO when the object contains metadata keys that are invalid by Azure's standard

### Testing Instructions:
1. Upload an object with metadata keys containing non-alphanumeric characters (e.g. `!@#-`) to a bucket backed by an Azure namespacestore
2. Download the object from the bucket
3. Verify that the tags are the same as they were originally 
Bonus step:
4. After 1, download the object directly from Azure, and verify that it now contains two new metadata keys related to the invalid one, starting with `rename_` and `rename_key_`

### DIFF
--- a/src/endpoint/s3/ops/s3_put_object.js
+++ b/src/endpoint/s3/ops/s3_put_object.js
@@ -49,7 +49,8 @@ async function put_object(req, res) {
         tagging,
         tagging_copy: s3_utils.is_copy_tagging_directive(req),
         encryption,
-        lock_settings
+        lock_settings,
+        azure_invalid_md_header: req.headers['azure-metadata-handling'] || undefined
     });
 
     if (reply.version_id && reply.version_id !== 'null') {
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -12,7 +12,9 @@ const azure_storage = require('../util/new_azure_storage_wrap');
 const stream_utils = require('../util/stream_utils');
 const s3_utils = require('../endpoint/s3/s3_utils');
 const schema_utils = require('../util/schema_utils');
-const valid_attr_regex = /^[A-Za-z_][A-Za-z0-9_]+$/;
+const invalid_azure_md_regex = /[^a-zA-Z0-9_]/g;
+const XATTR_RENAME_PREFIX = 'rename_';
+const XATTR_RENAME_KEY_PREFIX = 'rename_key_';
 
 /**
  * @implements {nb.Namespace}
@@ -409,16 +411,38 @@ class NamespaceBlob {
         return { obj_id };
     }
 
+    // Adhere to azcopy's ResolveInvalidKey for handling of invalid metadata keys
+    _rename_invalid_md_key(xattr, md_key) {
+        const valid_key = md_key.replace(invalid_azure_md_regex, '_');
+        xattr[XATTR_RENAME_PREFIX + valid_key] = xattr[md_key];
+        xattr[XATTR_RENAME_KEY_PREFIX + valid_key] = md_key;
+        delete xattr[md_key];
+    }
+
+    _exclude_invalid_md_key(xattr, md_key) {
+            delete xattr[md_key];
+    }
+
     _check_valid_xattr(params) {
-        // This md validation check is a part of namespace blob because but Azure Blob 
-        // accepts C# identifiers only but S3 accepts other xattr too.
-        const is_invalid_attr = ([key, val]) => !valid_attr_regex.test(key);
-        const invalid_attr = Object.entries(params.xattr).find(is_invalid_attr);
-        if (invalid_attr) {
-            const err = new Error('InvalidMetadata: metadata keys are invalid.');
-            err.rpc_code = 'INVALID_REQUEST';
-            throw err;
-        }
+        const modified_xattr = _.cloneDeep(params.xattr);
+        for (const md_key of Object.keys(params.xattr)) {
+            invalid_azure_md_regex.lastIndex = 0;
+            if (invalid_azure_md_regex.test(md_key)) {
+                switch (params.azure_invalid_md_header) {
+                    case 'ExcludeIfInvalid':
+                        this._exclude_invalid_md_key(modified_xattr, md_key);
+                        break;
+                    case 'FailIfInvalid':
+                        throw new Error('Object metadata key ' + md_key + ' is invalid in Azure and the FailIfInvalid header value was provided, thus the upload has failed');
+                    case 'RenameIfInvalid':
+                        this._rename_invalid_md_key(modified_xattr, md_key);
+                        break;
+                    default:
+                        this._rename_invalid_md_key(modified_xattr, md_key);
+                    }
+                }
+            }
+        params.xattr = modified_xattr;
     }
 
     async upload_multipart(params, object_sdk) {
@@ -656,6 +680,19 @@ class NamespaceBlob {
         const xattr = _.extend(obj.metadata, {
             'noobaa-namespace-blob-container': this.container,
         });
+        const modified_xattr = _.cloneDeep(xattr);
+        // Restore invalid metadata keys
+        for (const md_key of Object.keys(xattr)) {
+            if (md_key.startsWith(XATTR_RENAME_KEY_PREFIX)) {
+                const restored_key = md_key.substring(11);
+                if (_.has(modified_xattr, XATTR_RENAME_PREFIX + restored_key)) {
+                    modified_xattr[modified_xattr[md_key]] = modified_xattr[XATTR_RENAME_PREFIX + restored_key];
+                    delete modified_xattr[md_key];
+                    delete modified_xattr[XATTR_RENAME_PREFIX + restored_key];
+                }
+            }
+        }
+
         return {
             obj_id: blob_etag,
             bucket,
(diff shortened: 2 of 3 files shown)


src/sdk/namespace_blob.js
```diff
@@ -410,6 +411,17 @@ class NamespaceBlob {
...
+                let valid_key = tag_key.replace(invalid_azure_tag_regex, '_');
+                tags['rename_' + valid_key] = tags[tag_key];
+                tags['rename_key_' + valid_key] = tag_key;
+                delete tags[tag_key];
+            }
+        }
+        params.xattr = tags;
         // This md validation check is a part of namespace blob because but Azure Blob 
         // accepts C# identifiers only but S3 accepts other xattr too.
         const is_invalid_attr = ([key, val]) => !valid_attr_regex.test(key);
```
romayalon, line 427: This should be removed, right? as we now handle this before

src/sdk/namespace_blob.js
```diff
@@ -410,6 +411,17 @@ class NamespaceBlob {
     }
 
     _check_valid_xattr(params) {
+        // Adhere to azcopy's ResolveInvalidKey for handling of invalid metadata keys
+        let tags = params.xattr;
+        for (let tag_key in tags) {
+            if (invalid_azure_tag_regex.test(tag_key)) {
+                let valid_key = tag_key.replace(invalid_azure_tag_regex, '_');
```
romayalon, line 418: use of const in a variable that won't be reassigned is always preferred

src/sdk/namespace_blob.js
```diff
@@ -410,6 +411,17 @@ class NamespaceBlob {
...
+                tags['rename_' + valid_key] = tags[tag_key];
```
romayalon, line 419: please move rename_ and rename_key_ to be constant at the beginning of the file

src/sdk/namespace_blob.js
```diff
@@ -656,6 +668,17 @@ class NamespaceBlob {
         const xattr = _.extend(obj.metadata, {
             'noobaa-namespace-blob-container': this.container,
         });
+
+        // Restore invalid metadata keys
+        for (let tag_key in xattr) {
+            if (tag_key.startsWith("rename_key_")) {
+                let valid_key = tag_key.substring(11);
```
romayalon, line 675: here I think it's better to use slice, and rename_key_.length() instead of hard coded 11

romayalon: you can just declare the size of it as a const at the beginning of the file/ outside of the loop
also instead of slice, have you considered using replace one with '' (empty string ? )

src/sdk/namespace_blob.js
```diff
@@ -12,7 +12,9 @@ const azure_storage = require('../util/new_azure_storage_wrap');
 const stream_utils = require('../util/stream_utils');
 const s3_utils = require('../endpoint/s3/s3_utils');
 const schema_utils = require('../util/schema_utils');
-const valid_attr_regex = /^[A-Za-z_][A-Za-z0-9_]+$/;
+const invalid_azure_tag_regex = /[^a-zA-Z0-9_]/;
+const rename_prefix = 'rename_';
```
romayalon, line 16: maybe XATTR_RENAME_PREFIX and XATTR_RENAME_KEY_PREFIX ?

src/sdk/namespace_blob.js
```diff
@@ -656,6 +685,17 @@ class NamespaceBlob {
         const xattr = _.extend(obj.metadata, {
             'noobaa-namespace-blob-container': this.container,
         });
+
+        // Restore invalid metadata keys
+        xattr.forEach(tag_key => {
```
romayalon, line 690:
```suggestion
        xattr.forEach(md_key => {
```

src/sdk/namespace_blob.js
```diff
@@ -656,6 +685,17 @@ class NamespaceBlob {
...
+            if (tag_key.startsWith(rename_key_prefix)) {
+                let valid_key = tag_key.replace(rename_key_prefix, '');
```
romayalon, line 692:
```suggestion
                const restored_key = tag_key.replace(rename_key_prefix, '');
```

src/sdk/namespace_blob.js
```diff
@@ -409,16 +411,43 @@ class NamespaceBlob {
...
-        // accepts C# identifiers only but S3 accepts other xattr too.
-        const is_invalid_attr = ([key, val]) => !valid_attr_regex.test(key);
-        const invalid_attr = Object.entries(params.xattr).find(is_invalid_attr);
-        if (invalid_attr) {
-            const err = new Error('InvalidMetadata: metadata keys are invalid.');
-            err.rpc_code = 'INVALID_REQUEST';
-            throw err;
-        }
+        let xattr = params.xattr;
+        xattr.forEach(tag_key => {
```
romayalon, line 436:
```suggestion
        xattr.forEach(md_key => {
```

src/sdk/namespace_blob.js
```diff
@@ -12,7 +12,9 @@ const azure_storage = require('../util/new_azure_storage_wrap');
 const stream_utils = require('../util/stream_utils');
 const s3_utils = require('../endpoint/s3/s3_utils');
 const schema_utils = require('../util/schema_utils');
-const valid_attr_regex = /^[A-Za-z_][A-Za-z0-9_]+$/;
+const invalid_azure_tag_regex = /[^a-zA-Z0-9_]/;
```
romayalon, line 15:
```suggestion
const invalid_azure_md_regex = /[^a-zA-Z0-9_]/;
```

romayalon: We are currently not supporting tagging in namespace_blob (see Tagging section in this file), I find it very confusing to keep using tag term when manipulating metadata.

src/endpoint/s3/ops/s3_put_object.js
```diff
@@ -49,7 +49,8 @@ async function put_object(req, res) {
         tagging,
         tagging_copy: s3_utils.is_copy_tagging_directive(req),
         encryption,
-        lock_settings
+        lock_settings,
+        azure_invalid_tag_header: req.headers['azure-metadata-handling'] || undefined
```
romayalon, line 53: Is there a situation where meta data is passed but this header isn't?

src/sdk/namespace_blob.js
```diff
@@ -12,7 +12,9 @@ const azure_storage = require('../util/new_azure_storage_wrap');
 const stream_utils = require('../util/stream_utils');
 const s3_utils = require('../endpoint/s3/s3_utils');
 const schema_utils = require('../util/schema_utils');
-const valid_attr_regex = /^[A-Za-z_][A-Za-z0-9_]+$/;
+const invalid_azure_md_regex = /[^a-zA-Z0-9_]/;
+const XATTR_RENAME_PREFIX = 'rename_';
+const XATTR_RENAME_KEY_PREFIX = 'XATTR_RENAME_PREFIX_';
```
alphaprinz, line 17: should be "rename_key_" according to 
// 'rename_key_123_invalid':'123-invalid'

src/util/http_utils.js
```diff
@@ -525,6 +525,11 @@ function check_headers(req, options) {
         throw new options.ErrorClass(options.error_bad_request);
     }
 
+    if (!_.includes([undefined, 'ExcludeIfInvalid', 'FailIfInvalid', 'RenameIfInvalid'], req.headers['azure-metadata-handling'])) {
```
dannyzaken, line 528: since this is tested on any S3 request, maybe better to only verify if defined 
```suggestion
    if (req.headers['azure-metadata-handling'] && !_.includes(['ExcludeIfInvalid', 'FailIfInvalid', 'RenameIfInvalid'], req.headers['azure-metadata-handling'])) {
```

src/sdk/namespace_blob.js
```diff
@@ -409,16 +411,39 @@ class NamespaceBlob {
...
-        // accepts C# identifiers only but S3 accepts other xattr too.
-        const is_invalid_attr = ([key, val]) => !valid_attr_regex.test(key);
-        const invalid_attr = Object.entries(params.xattr).find(is_invalid_attr);
-        if (invalid_attr) {
-            const err = new Error('InvalidMetadata: metadata keys are invalid.');
-            err.rpc_code = 'INVALID_REQUEST';
-            throw err;
-        }
+        const modified_xattr = _.cloneDeep(params.xattr);
+        Object.keys(params.xattr).forEach(md_key => {
```
dannyzaken, line 428: I think that using `for of` is a better style than `forEach`. it is more readable, and also support standard control flow keywords (`break`, `continue`, etc.)

```suggestion
        for (const md_key of Object.keys(params.xattr)) {
```

dannyzaken (commented):
squash your commits

dannyzaken (approved):

PR #7262, opened 2023-04-10
Optimize NooBaa deletion paths

tangledbytes:
### Explain the changes
This PR attempts to improve performance NooBaa delete paths by doing the following:
1. Change bucket_reclaimer approach to object deletion - instead of deleting objects 1 by 1, 1k objects are marked deleted at once, this significantly reduces DB calls.
2. Add index for objectparts table which significantly speed up the queries which DB cleaner uses to identify the object parts that needs to be deleted.
3. Add index for objectmultiparts table which significantly speed up the queries which DB cleaner uses to identify the object parts that needs to be deleted.
4. Forces use of indexes by altering selections in DB calls.

#### Analysis
[analysis.md](https://github.com/noobaa/noobaa-core/files/11262724/analysis.md)

This was a customer issue where they noticed serious NooBaa degradation when a bucket with 3M objects was deleted.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -14,7 +14,7 @@ const glob_to_regexp = require('glob-to-regexp');
 
 const P = require('../../util/promise');
 const dbg = require('../../util/debug_module')(__filename);
-const { MDStore } = require('./md_store');
+const { MDStore, make_md_id } = require('./md_store');
 const LRUCache = require('../../util/lru_cache');
 const size_utils = require('../../util/size_utils');
 const time_utils = require('../../util/time_utils');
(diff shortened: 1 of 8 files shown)


dannyzaken (commented):
@utkarsh-pro. At first look, It looks good. There are few things we need to get more info on:
* How are the new queries behave with the indexes? We should run `EXPLAIN` on the new queries and ensure they hit the indexes correctly and do not require sorting.
* Let's inspect other paths that go through the delete flows. mainly `delete_multiple_objects`, `delete_object` and lifecycle

utkarsh-pro:
Folks, I have attached a `analysis.md` in the PR description which might be helpful in understanding several delete path queries performance.

nimrod-becker:
@guymguym if you feel like taking a look

baum (approved):
lgtm

src/server/object_services/schemas/object_part_indexes.js
```diff
@@ -31,4 +32,18 @@ module.exports = [{
             }
         }
     },
+    {
+        // This index is used for queries where we want to find all the chunks of a specific object
+        // which are already marked deleted.
+        fields: {
+            obj: 1,
```
dannyzaken, line 39: @utkarsh-pro IIUC, this can also be achieved by removing the filter on `deleted`, Is that right? I wonder which approach is better, since adding indexes can affect writes performance. @guymguym any thoughts?

utkarsh-pro: @dannyzaken, I think partial indexes are faster for both reads and writes.

> One major reason for using a partial index is to avoid indexing common values. Since a query searching for a common value (one that accounts for more than a few percent of all the table rows) will not use the index anyway, there is no point in keeping those rows in the index at all. This reduces the size of the index, which will speed up those queries that do use the index. It will also speed up many table update operations because the index does not need to be updated in all cases. [Example 11.1](https://www.postgresql.org/docs/current/indexes-partial.html#INDEXES-PARTIAL-EX1) shows a possible application of this idea.

Taken from: https://www.postgresql.org/docs/current/indexes-partial.html

utkarsh-pro: Or did I... misunderstood the comment?

dannyzaken (approved):

src/server/object_services/object_server.js
```diff
@@ -983,6 +983,39 @@ async function delete_multiple_objects_by_filter(req) {
...
+    const { objects } = await MDStore.instance().find_objects({
+        bucket_id: make_md_id(bucket_id),
+        limit,
+        key: undefined,
+    });
+
+    // delete the objects
+    await MDStore.instance().remove_objects_and_unset_latest(objects);
+
+    const bucket_has_objects = await MDStore.instance().has_any_objects_for_bucket(bucket_id);
```
guymguym, line 1015: We can remove this and just check the objects.length.
Both here and also in the previous function delete_multiple_objects_by_filter().

src/server/object_services/schemas/object_part_indexes.js
guymguym: We discussed these new indexes on parts/multiparts and while there is an alternative to define the index only on the `deleted` field like in objectmds aggregate_by_delete_dates, this would mean that db_cleaner will cleanup parts and multiparts without specific relation to a deleted object, and this could be fine, but we don't have to do it now. The benefit is that the index will allow an index condition on the delete date, but the downside is that we will delete some parts and multiparts of objects and not all parts of the object, which might look a bit weird for other scanners (perhaps). In any case, this looks fine as we have to have some index for deletion, and this index will be limited by the amount of pending rows to deleted (the db_cleaner will keep it small).

PR #7264, opened 2023-04-13
Change Rules In `.eslintrc.js` + Bump `eslint` Version

shirady:
### Explain the changes
1. Add rule `prefer-const` (for more information, see [here](https://eslint.org/docs/latest/rules/prefer-const)) and turn on `no-var` rule in turn off `logical-assignment-operators` rule `.eslintrc.js`.
2. Change files according to the changes in the rules above.
3. Bump `eslint` version to `8.38.0` (change file `package.json` and automatically was generated changes in file `package-lock.json` after `npm install`).

1. none

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -169,6 +169,11 @@ module.exports = {
         // instead of expression (foo = function() {})
         'func-style': ['error', 'declaration', { allowArrowFunctions: true }],
 
+        "prefer-const": ["error", {
+            "destructuring": "all",
+            "ignoreReadBeforeAssign": false
+        }],
+
         //////////////////////////////////////////////////////////////////////
         //                                                                  //
         // WARN                                                             //
@@ -295,9 +300,6 @@ module.exports = {
         // we do allow _name or name_ as identifiers
         'no-underscore-dangle': 'off',
 
-        // prefer to use let/const instead of var
-        'no-var': 'off',
-
         // turn off todo/fixme comments - will grep it to a different report
         'no-warning-comments': 'off',
 
@@ -318,9 +320,6 @@ module.exports = {
 
         // prefer using arrow functions for callbacks, but too much to fix
         'prefer-named-capture-group': 'off',
-
-        // we prefer using const, but too much to fix
-        'prefer-const': 'off',
 
         // we prefer using destructuring, but too much to fix
         'prefer-destructuring': 'off',
@@ -356,5 +355,8 @@ module.exports = {
 
         //Allow spacing between template tags and their literals
         'template-tag-spacing': 'off',
+
+        // we prefer not to adopt the logical assignment operators from ES2020
+        'logical-assignment-operators': 'off'
     }
 };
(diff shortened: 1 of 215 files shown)


shirady:
As you can see I chose in the rule:
```
        "prefer-const": ["error", {
            "destructuring": "all",
            "ignoreReadBeforeAssign": false
        }],
```
to use  `"destructuring": "all"` and not `"destructuring": "any"` (see [here](https://eslint.org/docs/latest/rules/prefer-const#destructuring)) because we have cases in which we have destructor and part of the variables are `const` and part of them `let`, for example:
https://github.com/noobaa/noobaa-core/blob/81307b6f8806be712b0c500877aa81e5ed651823/src/endpoint/s3/s3_rest.js#L307
(`bucket` is `const`, `key` is `let`...).

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# format changes: no-var, prefer-const
+a828ce6ab7d06077231bfe6f7827f9c9cf05f4ef
```
dannyzaken, line 2: shouldn't you add it only after merging to master?

dannyzaken:
@shirady great PR! 

Following up on the slack discussion with @liranmauda, let's change the eslint version to the latest (`v8.38.0`) and set `'logical-assignment-operators': 'off'` in `eslintrc`

.git-blame-ignore-revs
shirady: (@dannyzaken I added numbers):

> 1. what is this revision? 
> 2. Shouldn't you add it only after merging to master?

1. I'm not sure I understand the first question. if you asked about the file `.git-blame-ignore-revs` it is a file that allows us to do changes that will not be reflected in `git blame` option (in Github as well). [here](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) is the link for more information.
2. I did it in a separate commit, I will do it in a separate PR (after merging this PR).

shirady:
> @shirady great PR!
> 
> Following up on the slack discussion with @liranmauda, let's change the eslint version to the latest (`v8.38.0`) and set `'logical-assignment-operators': 'off'` in `eslintrc`

I added `'logical-assignment-operators': 'off'`.
@dannyzaken Are we sure that we want to bump the version in this PR? I asked @liranmauda in #7265 if he can bump it there.

dannyzaken:
> > @shirady great PR!
> > Following up on the slack discussion with @liranmauda, let's change the eslint version to the latest (`v8.38.0`) and set `'logical-assignment-operators': 'off'` in `eslintrc`
> 
> I added `'logical-assignment-operators': 'off'`. @dannyzaken Are we sure that we want to bump the version in this PR? I asked @liranmauda in #7265 if he can bump it there.

@liranmauda PR is intended to be backported to 5.13, but we don't want to backport the eslint changes. Also, I think it's relevant to this PR

dannyzaken:
LGTM. @liranmauda waiting for your approval

liranmauda (approved):

PR #7267, opened 2023-04-17
Remove `s3_access` As Required From `update_account_s3_access`

### LINKED ISSUE #7204: Add the Option to Define Default Backing Store by the User Explicitly
Not a problem, an improvement.

I want to have an explicit command to define the default resource in the CLI (and in the operator in general).

1. Leave it as is, today I used an RPC call `update_account_s3_access` to change the default resource.
2. Create another RPC call that is quite similar to `update_account_s3_access` just more simple (`email` and `default_resource` as required fields only) and with a meaningful name.
3.  Have an explicit command to define the default resource.

shirady:
### Explain the changes
1. Remove `s3_access` as required from `update_account_s3_access`. The use of `s3_access` was historical: before the UI was deprecated we had two types of accounts: 
(1) account access S3
(2) account UI - account UI was for using the UI and it didn't have `s3_access`.
Today all accounts have `s3_access`.

### Issues: Fixed #7204 
1. In the opened issue this is a fix to the [comment](https://github.com/noobaa/noobaa-core/issues/7204#issuecomment-1427477227).

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -266,7 +266,7 @@ module.exports = {
             method: 'PUT',
             params: {
                 type: 'object',
-                required: ['email', 's3_access'],
+                required: ['email'],
                 properties: {
                     email: { $ref: 'common_api#/definitions/email' },
                     s3_access: {


liranmauda (dismissed):
LGTM

liranmauda (changes requested):
Based on the explanation, It all accounts should have s3_access then:
1. In the `create_account` we should remove the if: https://github.com/noobaa/noobaa-core/blob/50a67612d5ae343e198071aaab77d84d35a5b12c/src/server/system_services/account_server.js#L85 
2. Same goes for the `update_account_s3_access`:  https://github.com/noobaa/noobaa-core/blob/50a67612d5ae343e198071aaab77d84d35a5b12c/src/server/system_services/account_server.js#L330 and: https://github.com/noobaa/noobaa-core/blob/50a67612d5ae343e198071aaab77d84d35a5b12c/src/server/system_services/account_server.js#L384 and the else of those if should be removed.
3. Same goes for the `validate_create_account_params`: https://github.com/noobaa/noobaa-core/blob/50a67612d5ae343e198071aaab77d84d35a5b12c/src/server/system_services/account_server.js#L1294

Another point is that if we don't need it, we should probably remove the require form `create_account` and  in the `account_api`.
Moreover, we should probably remove it all completely from the api, create an upgrade script, and fix the Operator calls for those.

nimrod-becker:
I agree with Liran's comment

shirady:
@liranmauda, thank you as discussed with @dannyzaken I'll close this PR for now.

PR #7268, opened 2023-04-17
Add And Update `.git-blame-ignore-revs`

shirady:
### Explain the changes
1. Add `.git-blame-ignore-revs` file with the commit hash of the changes (for more information, see [here](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)).
2. Update `.git-blame-ignore-revs` with commit number from #7264 that contains all file changes (but unfortunately also `eslint.js` change...). 

1. none

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# format changes: no-var, prefer-const
+4dedb1245288ba9c25b0aa7fe0a244f0864fdb98


liranmauda:
@shirady 
Can you explain the motivation?

shirady:
> @shirady Can you explain the motivation?

Of course, since I change the files without adding content (formatting them) adding the commit to the file `.git-blame-ignore-revs` will not show those lines as changed.

I created a repo to demonstrate:
With `.git-blame-ignore-revs` [blame](https://github.com/shirady/ignore-format/blame/master/hello.txt).
Without `.git-blame-ignore-revs` [blame](https://github.com/shirady/ignore-format/blame/without/hello.txt).

* The only flaw is that I added `eslint.js` to the commit...

liranmauda:
> > @shirady Can you explain the motivation?
> 
> Of course, since I change the files without adding content (formatting them) adding the commit to the file `.git-blame-ignore-revs` will not show those lines as changed.
> 
> I created a repo to demonstrate: With `.git-blame-ignore-revs` [blame](https://github.com/shirady/ignore-format/blame/master/hello.txt). Without `.git-blame-ignore-revs` [blame](https://github.com/shirady/ignore-format/blame/without/hello.txt).
> 
> * The only flaw is that I added `eslint.js` to the commit...

I am not convinced it is actually necessary... 
and what if a change to immutable is leading to an issue, why should we ignore it in the blame?

shirady:
> I am not convinced it is actually necessary... and what if a change to immutable is leading to an issue, why should we ignore it in the blame?

There is an assumption that this change should create an issue. I'm fine with leaving it, and maybe creating it in the future (if we would like to).

PR #7271, opened 2023-04-17
Add getBucketPolicyStatus command

nadavMiz:
### Explain the changes
Add getBucketPolicyStatus command. this command returns if pucket policy is public. currently bucket policy is public if statement has wildcard at either either principal or resource. should change once we support more complex statements

aws getBucketPolicyStatus documentation:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketPolicyStatus.html


src/server/system_services/bucket_server.js
```diff
@@ -384,6 +384,41 @@ async function get_bucket_policy(req) {
     };
 }
 
+async function get_bucket_policy_status(req) {
+    dbg.log0('get_bucket_policy_status:', req.rpc_params);
+    const bucket = find_bucket(req, req.rpc_params.name);
+    const policy = bucket.s3_policy;
+    if (!policy) return {};
+    const isPublic = _is_policy_public(policy);
```
jackyalbo, line 392: for internal variables please use our format of is_public
```suggestion
    const is_public = _is_policy_public(policy);
```

src/endpoint/s3/ops/s3_get_bucket_policy_status.js
```diff
@@ -0,0 +1,23 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const S3Error = require('../s3_errors').S3Error;
+
+/**
+ * https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetBucketPolicyStatus.html
+ */
+async function get_bucket_policy_status(req) {
+    const reply = await req.object_sdk.get_bucket_policy_status({ name: req.params.bucket });
```
jackyalbo, line 10: If we are using the same commands as get_bucket_policy. Maybe just move _is_policy_public to here and just call req.object_sdk.get_bucket_policy. WDYT?
```suggestion
    const reply = await req.object_sdk.get_bucket_policy({ name: req.params.bucket });
    if (!reply.policy) throw new S3Error(S3Error.NoSuchBucketPolicy);
    return _is_policy_public(reply.policy);
```

nadavMiz: seems much simpler, I'll change that

jackyalbo (approved):
Nice. Looks great :)

PR #7273, opened 2023-04-19
add bucket authentication for server side encryption algorithm

nadavMiz:
### Explain the changes
expend bucket policy to enable statements containing conditions for authenticating based on server side encryption algorithm. providing base for farther adding other condition statements. enabling deny, accept bucket policies for stringEquals StringNotEquals and null conditions for server side encryptions

### Testing Instructions:
1. ceph s3 test: s3tests_boto3.functional.test_s3:test_bucket_policy_put_obj_s3_noenc

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -254,7 +254,7 @@ module.exports = {
 
         bucket_policy_principal: {
             anyOf: [{
-                    wrapper: SensitiveString,
+                    wrapper: SensitiveString
                 }, {
                     type: 'object',
                     required: ['AWS'],
@@ -285,6 +285,22 @@ module.exports = {
                 }
               ]
         },
+
+        bucket_policy_string_condition: {
+            type: 'object',
+            additionalProperties: {
+                type: 'string'
+            }
+        },
+
+        bucket_policy_null_condition: {
+            type: 'object',
+            additionalProperties: {
+                enum: ['true', 'false'],
+                type: 'string'
+            }
+        },
+
         bucket_policy: {
             type: 'object',
             required: ['Statement'],
@@ -311,6 +327,32 @@ module.exports = {
                             },
                             Resource: {
                                 $ref: '#/definitions/string_or_string_array'
+                            },
+                            Condition: {
+                                type: 'object',
+                                properties: {
+                                    StringEquals: {
+                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    },
+                                    StringNotEquals: {
+                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    },
+                                    StringEqualsIgnoreCase: {
+                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    },
+                                    StringNotEqualsIgnoreCase: {
+                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    },
+                                    StringLike: {
+                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    },
+                                    StringNotLike: {
+                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    },
+                                    Null: {
+                                        $ref: '#/definitions/bucket_policy_null_condition'
+                                    }
+                                }
                             }
                         }
                     }
--- a/src/endpoint/s3/s3_bucket_policy_utils.js
+++ b/src/endpoint/s3/s3_bucket_policy_utils.js
@@ -1,0 +1,219 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const _ = require('lodash');
+const dbg = require('../../util/debug_module')(__filename);
+const s3_utils = require('./s3_utils');
+
+const OP_NAME_TO_ACTION = Object.freeze({
+    delete_bucket_analytics: { regular: "s3:PutAnalyticsConfiguration" },
+    delete_bucket_cors: { regular: "s3:PutBucketCORS" },
+    delete_bucket_encryption: { regular: "s3:PutEncryptionConfiguration" },
+    delete_bucket_inventory: { regular: "s3:PutInventoryConfiguration" },
+    delete_bucket_lifecycle: { regular: "s3:PutLifecycleConfiguration" },
+    delete_bucket_metrics: { regular: "s3:PutMetricsConfiguration" },
+    delete_bucket_policy: { regular: "s3:DeleteBucketPolicy" },
+    delete_bucket_replication: { regular: "s3:PutReplicationConfiguration" },
+    delete_bucket_tagging: { regular: "s3:PutBucketTagging" },
+    delete_bucket_website: { regular: "s3:DeleteBucketWebsite" },
+    delete_bucket: { regular: "s3:DeleteBucket" },
+    delete_object_tagging: { regular: "s3:DeleteObjectTagging", versioned: "s3:DeleteObjectVersionTagging" },
+    delete_object_uploadId: { regular: "s3:AbortMultipartUpload" },
+    delete_object: { regular: "s3:DeleteObject", versioned: "s3:DeleteObjectVersion" },
+
+    get_bucket_accelerate: { regular: "s3:GetAccelerateConfiguration" },
+    get_bucket_acl: { regular: "s3:GetBucketAcl" },
+    get_bucket_analytics: { regular: "s3:GetAnalyticsConfiguration" },
+    get_bucket_cors: { regular: "s3:GetBucketCORS" },
+    get_bucket_encryption: { regular: "s3:GetEncryptionConfiguration" },
+    get_bucket_inventory: { regular: "s3:GetInventoryConfiguration" },
+    get_bucket_lifecycle: { regular: "s3:GetLifecycleConfiguration" },
+    get_bucket_location: { regular: "s3:GetBucketLocation" },
+    get_bucket_logging: { regular: "s3:GetBucketLogging" },
+    get_bucket_metrics: { regular: "s3:GetMetricsConfiguration" },
+    get_bucket_notification: { regular: "s3:GetBucketNotification" },
+    get_bucket_policy: { regular: "s3:GetBucketPolicy" },
+    get_bucket_policy_status: { regular: "s3:GetBucketPolicyStatus" },
+    get_bucket_replication: { regular: "s3:GetReplicationConfiguration" },
+    get_bucket_requestpayment: { regular: "s3:GetBucketRequestPayment" },
+    get_bucket_tagging: { regular: "s3:GetBucketTagging" },
+    get_bucket_uploads: { regular: "s3:ListBucketMultipartUploads" },
+    get_bucket_versioning: { regular: "s3:GetBucketVersioning" },
+    get_bucket_versions: { regular: "s3:ListBucketVersions" },
+    get_bucket_website: { regular: "s3:GetBucketWebsite" },
+    get_bucket_object_lock: { regular: "s3:GetBucketObjectLockConfiguration" },
+    get_bucket: { regular: "s3:ListBucket" },
+    get_object_acl: { regular: "s3:GetObjectAcl" },
+    get_object_tagging: { regular: "s3:GetObjectTagging", versioned: "s3:GetObjectVersionTagging" },
+    get_object_uploadId: { regular: "s3:ListMultipartUploadParts" },
+    get_object_retention: { regular: "s3:GetObjectRetention"},
+    get_object_legal_hold: { regular: "s3:GetObjectLegalHold" },
+    get_object: { regular: "s3:GetObject", versioned: "s3:GetObjectVersion" },
+    get_service: { regular: "s3:ListAllMyBuckets" },
+
+    head_bucket: { regular: "s3:ListBucket" },
+    head_object: { regular: "s3:GetObject", versioned: "s3:GetObjectVersion" },
+
+    post_bucket_delete: { regular: "s3:DeleteObject" },
+    post_object: { regular: "s3:PutObject" },
+    post_object_uploadId: { regular: "s3:PutObject" },
+    post_object_uploads: { regular: "s3:PutObject" },
+    post_object_select: { regular: "s3:GetObject" },
+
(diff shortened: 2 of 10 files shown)


src/api/common_api.js
```diff
@@ -252,6 +252,15 @@ module.exports = {
             }
         },
 
+        bucket_policy_condition: {
+            type: 'object',
+            properties: {
+                serverSideEncryption: {
```
liranmauda, line 258: Stile: 
this is a camelcase style, we are using snake_case
```suggestion
                server_side_encryption: {
```

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -23,10 +23,38 @@ async function put_bucket_policy(req) {
...
+    const parsed_condition = {};
+    for (const [condition, obj] of Object.entries(conditions)) {
+        if (!valid_conditions[condition]) {
+            console.error('put_bucket_policy: invalid condition', condition);
+            throw new S3Error(S3Error.InvalidArgument);
+        }
+        parsed_condition[valid_conditions[condition]] = {};
+        for (const [condition_key, value] of Object.entries(obj)) {
+            if (!condition_key_map[condition_key]) {
+                console.error('put_bucket_policy: invalid condition key', condition_key);
```
aspandey, line 49: Not only condition_key but also log "condition" which has this condition_key. This will have better information if it fails.

nadavMiz: changed it

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -23,10 +23,38 @@ async function put_bucket_policy(req) {
         }
         if (statement.action) statement.action = _.flatten([statement.action]);
         if (statement.resource) statement.resource = _.flatten([statement.resource]);
+        if (statement.condition) statement.condition = _parse_condition(statement.condition);
     }
     await req.object_sdk.put_bucket_policy({ name: req.params.bucket, policy });
 }
 
+function _parse_condition(conditions) {
+    const valid_conditions = {
```
liranmauda, line 32: Please add the link for the conditions in a comment:

```suggestion
    // https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html
    const valid_conditions = {
```

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -23,10 +23,38 @@ async function put_bucket_policy(req) {
...
+        'stringequals': 'string_equals',
+        'stringnotequals': 'string_not_equals',
+        'null': 'null'
+    };
+    const condition_key_map = {
```
liranmauda, line 37: Please add a link here.

src/endpoint/s3/ops/s3_put_bucket_policy.js
liranmauda: We can change the name to reflact that those are the supported contions (currently it is only a subset of the list)

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -23,10 +23,38 @@ async function put_bucket_policy(req) {
...
+        's3:x-amz-server-side-encryption': 'server_side_encryption'
+    };
+    const parsed_condition = {};
+    for (const [condition, obj] of Object.entries(conditions)) {
+        if (!valid_conditions[condition]) {
+            console.error('put_bucket_policy: invalid condition', condition);
```
liranmauda, line 43: This Error is not indicating that we are not supporting the condition, for example if we get a condition that is valid such as `StringEqualsIgnoreCase` but we are not supporting it we should prompt the error that indicate that it is not supported.
also we can use dbg and not console
```suggestion
            dbg.error('put_bucket_policy: not supported condition', condition);
```

src/api/common_api.js
```diff
@@ -252,6 +252,15 @@ module.exports = {
             }
         },
 
+        bucket_policy_condition: {
```
liranmauda, line 255: We should think if this is actually required. 
@dannyzaken WDYT?

jackyalbo (commented):
We should add unit tests to cover these changes in test_s3_bucket_policy. Please add.

nadavMiz:
> We should add unit tests to cover these changes in test_s3_bucket_policy. Please add.

I have added unit tests. also I have noticed I didn't handle returning the condition values to their original values in get _bucket_policy, so I have added that, and unit test for it

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -23,10 +23,42 @@ async function put_bucket_policy(req) {
...
+        'null': 'null'
+    };
+    //https://docs.aws.amazon.com/AmazonS3/latest/userguide/list_amazons3.html#amazons3-policy-keys
+    //TODO: this is a subset of the full list of condition keys. should add more condtion keys in the future
+    const condition_key_map = {
+        's3:x-amz-server-side-encryption': 'server_side_encryption'
+    };
+    const parsed_condition = {};
+    for (const [condition, obj] of Object.entries(conditions)) {
+        if (!supported_conditions[condition]) {
```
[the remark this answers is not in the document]
nadavMiz: the reason I added the map was to convert the s3 format to the DB format (same as all the other maps), since we convert the json to lowercase at the beginning of the function all the conditions seems to be a single word (the s3 style is camel case, so StringNotEqual will be converted to stringnotequal while I want to convert it to string_not_equal). so I don't think string manipulations will work in this case. I don't want to remove the conversion to lower case since it might hurt other functionality of the bucket policy. maybe I can just leave the conditions as single words in the DB as well.
In get-bucket-policy we already return all the policy values in lower case anyway

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -23,10 +23,42 @@ async function put_bucket_policy(req) {
...
         if (statement.resource) statement.resource = _.flatten([statement.resource]);
+        if (statement.condition) statement.condition = _parse_condition(statement.condition);
     }
     await req.object_sdk.put_bucket_policy({ name: req.params.bucket, policy });
 }
 
+function _parse_condition(conditions) {
+    // https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html
+    // TODO: this is a subset of the full list of conditions. should add more condtions in the future
+    const supported_conditions = {
```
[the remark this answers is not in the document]
nadavMiz: its worth noting that since we convert the bucket policy to lower case. IgnoreCase wouldn't be different than regular stringEqual / stringNotEqual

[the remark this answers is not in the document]
jackyalbo: so maybe we shouldn't just lowercase the whole bucket policy. we can move the lowercase to principle/action/reason and avoid doing it for conditions.

liranmauda (dismissed):
> We should think if this is actually required. 
> @dannyzaken WDYT?

I don't think that we should add bucket authentication for server-side encryption algorithm In a PR just for fixing a ceph test.

As @dannyzaken and I suggested before, this feels too big of a change as it is touching the authentication, and we should do it as a feature and not partially as part of a ceph test.

I think that we should close this PR. 

@nimrod-becker @dannyzaken @jackyalbo WDYT?

src/api/common_api.js
```diff
@@ -287,6 +321,32 @@ module.exports = {
...
+                                    string_not_equals_ignore_case: {
+                                        $ref: '#/definitions/bucket_policy_condition'
+                                    },
+                                    string_like: {
+                                        $ref: '#/definitions/bucket_policy_condition'
+                                    },
+                                    string_not_like: {
+                                        $ref: '#/definitions/bucket_policy_condition'
+                                    },
+                                    null: {
```
guymguym, line 346: we probably don't want to use a name such as `null` as a key - this will be error prone in the code.
Maybe call it null_condition or something like that?

src/api/common_api.js
```diff
@@ -267,7 +301,7 @@ module.exports = {
                                 type: 'string'
                             },
                             effect: {
-                                enum: ['allow', 'deny'],
+                                enum: ['Allow', 'Deny'],
```
guymguym, line 304: Is this change backward compatible?

src/endpoint/s3/ops/s3_get_bucket_policy.js
```diff
@@ -2,16 +2,38 @@
...
 async function get_bucket_policy(req) {
     const reply = await req.object_sdk.get_bucket_policy({ name: req.params.bucket });
     if (!reply.policy) throw new S3Error(S3Error.NoSuchBucketPolicy);
+    for (const statement of reply.policy.statement) {
+        if (statement.condition) statement.condition = _parse_condition(statement.condition);
+    }
     return reply.policy;
 }
 
+function _parse_condition(conditions) {
```
guymguym, line 19: is this needed? why can't we store the bucket policy json as is?

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -10,21 +10,79 @@ const S3Error = require('../s3_errors').S3Error;
...
+            adapted_policy.statement.push(adapted_statement);
         }
-        if (statement.action) statement.action = _.flatten([statement.action]);
-        if (statement.resource) statement.resource = _.flatten([statement.resource]);
     }
-    await req.object_sdk.put_bucket_policy({ name: req.params.bucket, policy });
+    await req.object_sdk.put_bucket_policy({ name: req.params.bucket, policy: adapted_policy });
+}
+
+function _parse_condition(conditions) {
```
guymguym, line 40: why are we parsing it? can't we store it as is?

src/endpoint/s3/s3_rest.js
```diff
@@ -210,10 +210,14 @@ async function authorize_request_policy(req) {
     const auth_token = req.object_sdk.get_auth_token();
     const arn_path = _get_arn_from_req_path(req);
     const method = _get_method_from_req(req);
+    const condition_elements = {};
+    condition_elements.encryption = s3_utils.parse_encryption(req);
+    condition_elements.bucket = req.params.bucket;
+    condition_elements.tagging = await req.object_sdk.get_object_tagging(req.params)?.tagging;
```
guymguym, line 216: why do we always get the tagging here? this is always needed? can't we get it on demand?

src/api/common_api.js
nadavMiz: Null is the key used by AWS. It should stay if we want to store the JSON as is. However, I can change it specifically if it is problematic

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -0,0 +1,217 @@
...
+    put_object_legal_hold: { regular: "s3:GetObjectLegalHold"},
+    put_object: { regular: "s3:PutObject" },
+});
+
+const qm_regex = /\?/g;
+const ar_regex = /\*/g;
+
+const predicate_map = {
+    'StringEquals': (request_value, policy_value) => request_value === policy_value,
+    'StringNotEquals': (request_val, policy_value) => request_val !== policy_value,
```
jackyalbo, line 94: please use request_value instead of request_val in all the places in the map.
```suggestion
    'StringNotEquals': (request_value, policy_value) => request_value !== policy_value,
```

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -0,0 +1,217 @@
...
+
+const supported_actions = {
+    's3:ExistingObjectTag': ['s3:DeleteObjectTagging', 's3:DeleteObjectVersionTagging', 's3:GetObject', 's3:GetObjectAcl', 's3:GetObjectTagging', 's3:GetObjectVersion', 's3:GetObjectVersionTagging', 's3:PutObjectAcl', 's3:PutObjectTagging', 's3:PutObjectVersionTagging'],
+    's3:x-amz-server-side-encryption': ['s3:PutObject']
+};
+
+async function _is_server_side_encryption_fit(req, predicate, value) {
+    const encryption = s3_utils.parse_encryption(req);
+    const algorithm = encryption ? encryption.algorithm : null;
+    dbg.log1('bucket_policy: encrytpion fit?', value, algorithm, predicate(algorithm, value));
```
jackyalbo, line 121: move calling the predicate outside of the function and print the result

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -0,0 +1,217 @@
...
+    return predicate(algorithm, value);
+}
+
+async function _is_object_tag_fit(req, predicate, value) {
+    const reply = await req.object_sdk.get_object_tagging(req.params);
+    const tag = reply?.tagging?.find(element => (element.key === value.key));
+    const tag_value = tag ? tag.value : null;
+    dbg.log1('bucket_policy: object tag fit?', value, tag);
```
jackyalbo, line 129: add here the predicate response like you did in the previous function (again without calling it twice)

src/server/system_services/bucket_server.js
```diff
@@ -513,6 +513,18 @@ function _validate_s3_policy(policy, bucket_name) {
...
             }
         }
+        if (statement.Condition) {
+            for (const condition of Object.values(statement.Condition)) {
+                for (const condition_key of Object.keys(condition)) {
+                    // some condition keys have arguments in their names.(e.g. s3:ExistingObjectTag/<key>)
+                    // parse to get only the condition key itself
+                    const key_parts = condition_key.split("/");
+                    if (!SUPPORTED_BUCKET_POLICY_CONDITIONS.includes(key_parts[0])) {
+                        throw new RpcError('MALFORMED_POLICY', 'Policy has invalid condition key', { detail: condition_key });
```
jackyalbo, line 523: Should we return "malformed policy" in case of a condition we don't support? Maybe we should at least add something like this:
```suggestion
                        throw new RpcError('MALFORMED_POLICY', 'Policy has invalid or unsupported condition key', { detail: condition_key });
```
WDYT?

jackyalbo (approved):

PR #7276, opened 2023-04-20
Fixing bad return code of NPM test

jackyalbo:
### Explain the changes
1. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -15,4 +15,4 @@ jobs:
       - name: Run Test
         run: |
           make test
-          docker run --privileged --rm --name test1 noobaa-tester ./src/test/unit_tests/run_npm_test_on_test_container.sh -s sudo_index.js
+          make privileged-test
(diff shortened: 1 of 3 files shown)


liranmauda (approved):

nimrod-becker (approved):

nimrod-becker:
This is a 4 year old code 😲

jackyalbo:
@nimrod-becker not really, part of my latest PR and me trying to get rid of mongo :)

.github/workflows/unit.yaml
```diff
@@ -15,4 +15,4 @@ jobs:
       - name: Run Test
         run: |
           make test
```
shirady, line 17: Since you created `privileged-test` with a dependency of `tester`, why do we need this line?

jackyalbo: make test will run the regular - none privileged tests
make privileged-test will run the extra privileged tests. 
We just pushed line 18 into its own makefile target, so it will start all the needed dockers.

PR #7283, opened 2023-04-27
block_store_fs enhanced to use xattrs and migrate blocks to tier2

guymguym:
### Explain the changes
1. Enhance backing_store_fs to use xattr to store block_md json instead of in separate .meta files.
2. This allows reducing the number of files needed on the file system by half.
3. Add new config option (false by default) - `config.BLOCK_STORE_FS_TIER2_ENABLED` - enable to set custom xattr trigger to signal the filesystem to migrate the data to archive tier (for filesystems that support that).
4. Added new config option (false by default) - `config.BLOCK_STORE_FS_MAPPING_INFO_ENABLED ` - enable to pass on extra recovery information (which bucket, object, offset, etc) to be added in block_md xattr, for cases when recovery without the db is needed.

1. NA

### Testing Instructions:
1. Regression of backing_store_fs (PV pools from the operator).
2. Enable the new configs to test the new options and check the xattr of the blocks (e.g `xattr -l <filename>` on mac).

### DIFF
--- a/src/agent/block_store_services/block_store_fs.js
+++ b/src/agent/block_store_services/block_store_fs.js
@@ -7,9 +7,10 @@ const path = require('path');
 
 const P = require('../../util/promise');
 const dbg = require('../../util/debug_module')(__filename);
+const config = require('../../../config.js');
 const fs_utils = require('../../util/fs_utils');
 const os_utils = require('../../util/os_utils');
-const config = require('../../../config.js');
+const nb_native = require('../../util/nb_native');
 const string_utils = require('../../util/string_utils');
 const BlockStoreBase = require('./block_store_base').BlockStoreBase;
 const get_block_internal_dir = require('./block_store_base').get_block_internal_dir;
@@ -24,9 +25,10 @@ class BlockStoreFs extends BlockStoreBase {
         this.old_blocks_path = path.join(this.root_path, 'blocks');
         this.config_path = path.join(this.root_path, 'config');
         this.usage_path = path.join(this.root_path, 'usage');
+        this.fs_context = {};
     }
 
-    init() {
+    async init() {
         // create internal directories to hold blocks by their last 3 hex digits
         // this is done to reduce the number of files in one directory which leads
         // to bad performance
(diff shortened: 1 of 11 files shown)


src/agent/block_store_services/block_store_fs.js
```diff
@@ -58,171 +60,178 @@ class BlockStoreFs extends BlockStoreBase {
...
+        /** @type {nb.NativeFSXattr} */
+        const xattr = {};
+
+        // fsync is needed before actually setting the migrate trigger
+        let xattr_need_fsync = false;
+
+        // set the block md xattr
+        const block_md_to_store = _.pick(block_md, 'id', 'digest_type', 'digest_b64', 'mapping_info');
         const block_md_data = JSON.stringify(block_md_to_store);
+        xattr[config.BLOCK_STORE_FS_XATTR_BLOCK_MD] = block_md_data;
```
romayalon, line 135: on overwrite, shouldn't we also delete the old meta_path file?

src/agent/block_store_services/block_store_fs.js
```diff
@@ -58,171 +60,178 @@ class BlockStoreFs extends BlockStoreBase {
...
-                if (this._usage && del_stat) {
-                    const usage = {
-                        size: -(del_stat.size + ((md_del_stat && md_del_stat.size) ? md_del_stat.size : 0)),
-                        count: -1
-                    };
-                    return this._update_usage(usage);
-                }
-            });
+
+        const [block_stat, meta_stat] = await Promise.all([
```
romayalon, line 202: old code guards meta_path deletion to be after block_path deletion, don't we want to keep this behavior in the new code?

src/native/fs/fs_napi.cpp
```diff
@@ -856,19 +898,33 @@ struct Rename : public FSWorker
...
         , _mode(0666)
     {
         _path = info[1].As<Napi::String>();
         auto buf = info[2].As<Napi::Buffer<uint8_t>>();
         _data = buf.Data();
         _len = buf.Length();
         if (info.Length() > 3 && !info[3].IsUndefined()) {
-            _mode = info[3].As<Napi::Number>().Uint32Value();
+        }
+        if (info.Length() > 3 && !info[3].IsUndefined()) {
```
romayalon, line 919: is this cond `if (info.Length() > 3 && !info[3].IsUndefined()) { } ` appears twice?

src/sdk/namespace_fs.js
```diff
@@ -1495,6 +1500,7 @@ class NamespaceFS {
                 config.NSFS_OPEN_READ_MODE,
                 get_umasked_mode(config.BASE_MODE_FILE),
             );
+            // @ts-ignore the next call is incorrect...
```
romayalon, line 1503: I think you can just remove this function completely

src/sdk/namespace_fs.js
```diff
@@ -329,7 +333,7 @@ class NamespaceFS {
              * @typedef {{
              *  key: string,
              *  common_prefix: boolean,
-             *  stat?: fs.Stats,
+             *  stat?: nb.NativeFSStats,
```
romayalon, line 336: this nb.NativeFSStats type needed in the LRU cache as well

src/native/fs/fs_napi.cpp
guymguym: fixed, thanks

src/sdk/namespace_fs.js
guymguym: you're right! fixed.

src/sdk/namespace_fs.js
guymguym: done

src/agent/block_store_services/block_store_fs.js
guymguym: I originally thought I can skip that because we do not actually expect block overwrites on our flows, and it can only occur during a retry of some sort (even then not sure it can happen). Looking back I think overwrite block could have returned an error, but as we chose to avoid erroring and just handle the case, I will fix. Thanks

src/agent/block_store_services/block_store_fs.js
guymguym: I don't think the old code did that - it was using Promise.all too and awaited those calls in parallel.

src/agent/block_store_services/block_store_fs.js
```diff
@@ -58,171 +60,181 @@ class BlockStoreFs extends BlockStoreBase {
...
+                    await nb_native().fs.unlink(fs_context, meta_path).catch(ignore_not_found);
                 }
-                const size = (block_md.is_preallocated ? 0 : data.length) + block_md_data.length - overwrite_size;
-                const count = (block_md.is_preallocated ? 0 : 1) - overwrite_count;
-                if (size || count) this._update_usage({ size, count });
+            }
+        }
+
+        try {
+            await nb_native().fs.writeFile(fs_context, block_path, data, {
```
utkarsh-pro, line 164: I am just trying to understand here. What happens if the unlink succeeds but this fails and `_test_root_path_exists` throws as well?

src/native/fs/fs_napi.cpp
```diff
@@ -883,6 +937,18 @@ struct Writefile : public FSWorker
         } else if ((size_t)len != _len) {
             SetError(XSTR() << "Writefile: partial write error " << DVAL(len) << DVAL(_len));
         }
+
+        if (_set_xattr) {
+            if (_xattr_need_fsync) {
+                SYSCALL_OR_RETURN(fsync(fd));
```
utkarsh-pro, line 943: Again, trying to understand - we fsync before setting xattr intentionally?

src/native/fs/fs_napi.cpp
```diff
@@ -931,13 +1011,25 @@ struct Readfile : public FSWorker
             remain -= len;
             p += len;
         }
+
+        CHECK_CTIME_CHANGE(fd, _stat_res, _path);
```
utkarsh-pro, line 1015: Can it happen that a race happens after the comparison of ctime and before function exit? Is this interleaving possible?

guymguym: Yes but the "exit" from the function is not a point of interest, what we care here is that ctime has not changed while we were reading the content and the xattrs of the file, which means we read an "atomic" state of the file. After we completed loading everything to memory, it can change, but our memory holds a consistent state of the file. Hope this is clear.

src/agent/block_store_services/block_store_fs.js
guymguym: The usage is not going to update in that case, which is indeed undesired, but since usage calculation is not atomic anyhow, any failure can cause this.

I explained to @romayalon's question in the [comment above](https://github.com/noobaa/noobaa-core/pull/7283/files#r1185765251) that the overwrite block case is not an actual flow we invoke on purpose.

The reason we decided that write_block should handle overwrites was that we didn't want to fail user uploads on one hand, and we didn't want the usage to start getting skewed because of retries so we had to check if this is an overwrite first and "reduce" that replaced size.

I guess that simply failing the request if the same block is written with different data could just be better.

src/native/fs/fs_napi.cpp
utkarsh-pro: That makes sense. But what tier2 FS does when XFS fsync fails (data block write failure)? Does tier2 filesystem fsync fails? If yes, do we not mark it for migration?

utkarsh-pro: On a second thought. Are we even supposed to handle fsync failures? If no, then my comment is useless.

guymguym: fsync can fail with EIO or ENOSPC. 
the question is what else can we do on these cases.
now we fail the block write, and the endpoint will retry - see [retry_write_block()](https://github.com/noobaa/noobaa-core/blob/dcf47166771b814524e44cafab2895ea2a5d3b7f/src/sdk/map_client.js#L251)
so for now I think this is a good starting point to fail and try again, up to some retry limit.
this is also related to the question we had on writing with O_DIRECT...
We should revise this if needed in the future.
Thanks. Resolving

src/agent/block_store_services/block_store_fs.js
guymguym: For now I will keep the existing behavior and we can revise as needed in the future.

src/agent/block_store_services/block_store_fs.js
romayalon: right, I mistakenly thought that they are chained, resolving

PR #7296, opened 2023-05-11
standalone commands

guymguym:
### Explain the changes
1. Standalone commands
    1. Moved all standalone commands like nsfs to `src/cmd/*`.
    2. Add `src/cmd/api` command to send rpc api like the operator api command.
    3. Add also the web+bg+s3+hosted_agents+s3cat commands to src/cmd.
    4. Build single-executable from src/cmd/index with pkg so it can be used to run all the core services.
2. npm scripts:
   1. Add `npm run db*` scripts for running postgres on dev env.
   2. add `--unhandled-rejections=warn` to node commands in npm scripts.
   3. Reorganize npm scripts a bit.
3. Added `docs/standalone.md` with build and run instructions.
4. Fix a small problem in system_store that caused standalone processes to try and connect to the db even if it was not needed.

1. NA

### Testing Instructions:
1. In `docs/standalone.md`


src/server/web_server.js
```diff
@@ -37,53 +38,54 @@ const stats_aggregator = require('./system_services/stats_aggregator');
 const addr_utils = require('../util/addr_utils');
 const kube_utils = require('../util/kube_utils');
 const http_utils = require('../util/http_utils');
+const server_rpc = require('./server_rpc');
 
 const rootdir = path.join(__dirname, '..', '..');
 const dev_mode = (process.env.DEV_MODE === 'true');
+const http_port = process.env.PORT || '5001';
+const https_port = process.env.SSL_PORT || '5443';
```
liranmauda, line 46: maybe worth moving it to the config.js? 
so it will be something like this: 
```suggestion
const http_port = config.PORT;
const https_port =config.SSL_PORT;
```

src/server/web_server.js
```diff
@@ -37,53 +38,54 @@ const stats_aggregator = require('./system_services/stats_aggregator');
...
 
-if (process.env.NOOBAA_LOG_LEVEL) {
-    const dbg_conf = debug_config.get_debug_config(process.env.NOOBAA_LOG_LEVEL);
-    dbg_conf.core.map(module => dbg.set_module_level(dbg_conf.level, module));
-}
+async function main() {
+    try {
 
-db_client.instance().connect();
+        if (process.env.NOOBAA_LOG_LEVEL) {
```
liranmauda, line 52: Same as above, maybe move into the config.js?

liranmauda (commented):
some small comments + Love the cmd api 🙂

guymguym:
@liranmauda Thanks for the comments, these are about re-organizing the config/env, which I am very much interested in. But I will postpone this to next PRs. Thanks

PR #7297, opened 2023-05-11
Calculate MD5 per account

aayushchouhan09:
### Explain the changes
1.  Added a flag to calculate MD5 etag per account.

- [X] Tests added

### DIFF
--- a/src/test/unit_tests/test_system_servers.js
+++ b/src/test/unit_tests/test_system_servers.js
@@ -193,6 +193,33 @@ mocha.describe('system_servers', function() {
         await rpc_client.events.read_activity_log({ limit: 2016 });
     });
 
+    mocha.it('Calculate md5_etag for account works', async function() {
+        await rpc_client.account.create_account({
+            name: EMAIL1,
+            email: EMAIL1,
+            has_login: false,
+            s3_access: true,
+            force_md5_etag: true,
+        });
+        let info = await rpc_client.account.read_account({
+            email: EMAIL1,
+        });
+        assert(info.force_md5_etag === true);
+
+        await rpc_client.account.update_account({
+            email: EMAIL1,
+            force_md5_etag: false,
+        });
+        info = await rpc_client.account.read_account({
+            email: EMAIL1,
+        });
+        assert(info.force_md5_etag === false);
+
+        await rpc_client.account.delete_account({
+            email: EMAIL1,
+        });
+    });
+
     ////////////
     //  AUTH  //
     ////////////
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -64,6 +64,9 @@ module.exports = {
                             },
                         },
                     },
+                    force_md5_etag: {
+                        type: 'boolean'
+                    },
                     nsfs_account_config: {
                         $ref: 'common_api#/definitions/nsfs_account_config'
                     },
@@ -197,6 +200,9 @@ module.exports = {
                                 enum: ['DARK', 'LIGHT']
                             }
                         }
+                    },
+                    force_md5_etag: {
+                        type: 'boolean'
                     },
                     role_config: {
                         $ref: 'common_api#/definitions/role_config'
@@ -707,6 +713,9 @@ module.exports = {
                             enum: ['DARK', 'LIGHT']
                         }
                     }
+                },
+                force_md5_etag: {
+                    type: 'boolean'
                 },
                 nsfs_account_config: {
                     $ref: 'common_api#/definitions/nsfs_account_config'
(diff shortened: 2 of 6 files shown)


src/util/chunk_fs.js
```diff
@@ -19,14 +19,15 @@ class ChunkFS extends stream.Transform {
...
      *      stats: import('../sdk/endpoint_stats_collector').EndpointStatsCollector,
      * }} params
      */
-    constructor({ target_file, fs_context, namespace_resource_id, stats }) {
+    constructor({ target_file, fs_context, namespace_resource_id, md5_enabled, stats }) {
         super();
         this.q_buffers = [];
         this.q_size = 0;
-        this.MD5Async = config.NSFS_CALCULATE_MD5 ? new (nb_native().crypto.MD5Async)() : undefined;
+        this.MD5Async = md5_enabled ? new (nb_native().crypto.MD5Async)() : undefined;
```
utkarsh-pro, line 30: Can we have `this.MD5Async = (md5_enabled || config.NSFS_CALCULATE_MD5) ? new (nb_native().crypto.MD5Async)() : undefined;` so that we can preserve the older behaviour? Or preserving the older behaviour not a requirement?

src/sdk/namespace_nb.js
```diff
@@ -129,6 +129,7 @@ class NamespaceNB {
             active_triggers,
             operation
         });
+        params = _.omit(params, ['force_md5_etag_bucket', 'force_md5_etag_account']);
```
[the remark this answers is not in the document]
achouhan09: Yes correct, currently we are supporting this feature for nsfs only.

src/util/chunk_fs.js
achouhan09: No, Actually our requirement is to remove `this.NSFS_CALCULATE_MD5` which is a global flag and replace it with our new flag `force_md5_enabled`

utkarsh-pro: Then I think we need to update the tests as well which might be setting this to true/false and then performing tests. Also if the intention is to get rid of `NSFS_CALCULATE_MD5` then maybe we can get rid of it from `config.js` as well?

src/sdk/object_sdk.js
```diff
@@ -689,6 +698,8 @@ class ObjectSDK {
                 const ns = await this._get_bucket_namespace(params.bucket);
                 this._check_is_readonly_namespace(ns);
                 if (params.copy_source) await this.fix_copy_source_params(params, ns);
+                params.force_md5_etag_bucket = await this.read_force_md5_etag_for_bucket(params.bucket);
+                params.force_md5_etag_account = await this.read_force_md5_etag_for_account();
```
dannyzaken, line 702: I'm not sure there is a reason not to send a single force_md5 flag, WDYT?
```suggestion
                params.force_md5_etag = await this.read_force_md5_etag_for_bucket(params.bucket) || await this.read_force_md5_etag_for_account();
```

src/sdk/object_sdk.js
```diff
@@ -160,6 +160,15 @@ class ObjectSDK {
         return bucket.namespace;
     }
 
+    async read_force_md5_etag_for_bucket(name) {
+        const { bucket } = await bucket_namespace_cache.get_with_cache({ sdk: this, name });
+        return bucket.force_md5_etag;
+    }
+
+    async read_force_md5_etag_for_account() {
```
dannyzaken, line 168: There is no need to make this function `async`, it only accesses memory

src/sdk/object_sdk.js
achouhan09: Yes, that would be better I think. We can remove some unecessary checks in that way.

src/sdk/object_sdk.js
achouhan09: yes

src/test/unit_tests/test_system_servers.js
```diff
@@ -193,6 +193,33 @@ mocha.describe('system_servers', function() {
...
+            email: EMAIL1,
+            has_login: false,
+            s3_access: true,
+            force_md5_etag: true,
+        });
+        let info = await rpc_client.account.read_account({
+            email: EMAIL1,
+        });
+        assert(info.force_md5_etag === true);
+
```
utkarsh-pro, line 208: Maybe we can also test if after setting this true NSFS really enforces MD5 calculation? Will be helpful to avoid regressions when someone else is working on this area.

achouhan09: Thanks, yes your are right.

We are not testing the complete flow here, this test is only to check the account part. I think, we can have one more test to check the whole flow of md5 calculation but for this PR the above test should be fine. @dannyzaken wdyt??

dannyzaken: in the system servers test we mainly tests the APIs. in this case the force_md5 flag is also not having much effect, since it is only used for NSFS flows.
I agree that we should have separate tests for the functionality of the md5 calculations in NSFS

tangledbytes (approved):

dannyzaken (approved):

tangledbytes (approved):

PR #7300, opened 2023-05-18
Removing DOCKER_BUILDKIT env variable from the Makefile

liranmauda:
### Explain the changes
Removing DOCKER_BUILDKIT env variable from the Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,12 +50,6 @@ ifdef testname
 	endif
 endif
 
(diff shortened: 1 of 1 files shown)


liranmauda:
It seems that The CI does not have it. 
Not sure if we care, but it will make the build a bit longer.

utkarsh-pro:
Yeah, it doesn't seem like we use any docker buildx specific features.

tangledbytes (approved):

Issue #5341, opened 2019-02-14
Object Tagging Gaps on Namespace buckets


jeniawhite:
### Environment info
- Version: **VERSION**
- Deployment: **AZURE | AWS | GCLOUD | ESX | VBOX | DEV**
- Customer: **NAME**

### Actual behavior
GAPS:
A) Namespace s3 as source doesn't support copy object tagging:
1. Since the S3 SDK doesn't return the TagCount in the headObject we cannot know if we should trigger a getObjectTagging in order to get the tags.
2. This works when the source namespace is NooBaa.
3. We can use the header from the readObject in order to know if there are any tags, or we can do the inefficient way of just calling getObjectTagging for every copy object call that is originated from namespace s3.

B) Namespace blob doesn't support object tagging

C) Namespace s3 doesn't return TagCount (we need to take the counter from the headers of read_object_stream and not just the body stream)

### Expected behavior
1. We should be able to populate the tags in the object_sdk.

### Steps to reproduce
1.

nimrod-becker:
https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-tags?tabs=azure-ad

PR #7303, opened 2023-05-21
Fix get_tags for copy from namespace-s3

nadavMiz:
### Explain the changes
1. Get tag_count from read_object_stream on s3 since this value is not always returned in read_object_md
2. moved _populate_source_object_tagging to only run when copying on different servers. (when copying from the same server tagging is copied already)

1. https://github.com/noobaa/noobaa-core/issues/5341

### Testing Instructions:
1. create s3 namespace bucket, and noobaa bucket
2. create object with tags on the s3 bucket
3. copy object from s3 bucket to noobaa bucket with COPY tagging_dierctive
4. run get_object_tagging on the new object. check that the tags were copied

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/object_sdk.js
+++ b/src/sdk/object_sdk.js
@@ -590,7 +590,6 @@ class ObjectSDK {
         // get the namespace for source bucket
         const source_ns = await this._get_bucket_namespace(bucket);
         const source_md = await source_ns.read_object_md(source_params, this);
-        if (params.tagging_copy) await this._populate_source_object_tagging({ source_params, source_ns, source_md });
         const ranges = http_utils.normalize_http_ranges(
             params.copy_source.ranges, source_md.size, true);
         // For ranged copy we don't have the specific range hashes
@@ -660,6 +659,10 @@ class ObjectSDK {
                 dbg.warn(`upload_object with copy_sources - copying by reading source first (not server side)
                 so it can take some time and cause client timeouts`);
             }
+            if (params.tagging_copy) {
+                await this._populate_source_object_tagging({ source_params, source_ns, source_md });
+                params.tagging = source_md.tagging;
+            }
             // reset the copy_source param
             params.copy_source = null;
         }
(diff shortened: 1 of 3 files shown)


src/sdk/object_sdk.js
```diff
@@ -656,6 +655,10 @@ class ObjectSDK {
                 dbg.warn(`upload_object with copy_sources - copying by reading source first (not server side)
                 so it can take some time and cause client timeouts`);
             }
+            if (params.tagging_copy) {
```
jackyalbo, line 658: doesn't moving this to here effect line 612: 
```if (params.tagging_copy) params.tagging = source_md.tagging;```

Comment (nadavMiz) [reply src/sdk/object_sdk.js:662]: this line only affects copying between noobaa buckets, as other buckets copy within the server is handled by the SDK. in the noobaa bucket case the tagging is already returned by read_object_md

**Comment by nadavMiz** [comment]
added unit test for copying objects with tags within the same server and different servers. for the different server case I use AWS as a different server then noobaa, for namespace buckets I use noobaa bucket. didn't bother with azure, since azurite doesn't support tagging anyway

File: src/test/unit_tests/test_s3_ops.js [hunk 2]
```diff
@@ -423,6 +477,68 @@ mocha.describe('s3_ops', function() {
...
+                Tagging: {
+                    TagSet: [{
+                        Key: 's3ops',
+                        Value: 'copy_tagging'
+                    }]
+                }
+            };
+
+            let notSupported = false;
+            await s3.putObjectTagging(params).on('error', err => {
```
jackyalbo, line 496: does every error here mean putObjectTagging is not supported? maybe we should be more specific?

src/test/unit_tests/test_s3_ops.js
```diff
@@ -423,6 +477,68 @@ mocha.describe('s3_ops', function() {
...
+                Key: text_file1,
+                Tagging: {
+                    TagSet: [{
+                        Key: 's3ops',
+                        Value: 'copy_tagging_different_platform'
+                    }]
+                }
+            };
+            await s3.putObjectTagging(params).on('error', err => {
+                notSupported = true;
```
jackyalbo, line 528: again here

src/test/unit_tests/test_s3_ops.js
nadavMiz: I have added the error. I have to mention that I never actually saw this error happens. I just saw that whenever putObjectTagging is called in the code this error is checked. whoever added it (Sanjeev Gupta) doesn't work in the company anymore. I added it to be consistent with the code, in case there are situations I have missed

jackyalbo: I don't think we should follow his logic, especially if we don't understand it. If you have some cases this test needs to be skipped - keep it, but otherwise, I think it's better to just fail on any error thrown (avoiding this on error/complete).

nadavMiz: I removed the the error checking

jackyalbo (approved):

PR #7307, opened 2023-05-23
nsfs command: simple auth + fix forks flag + fix read_account perf + warmup before buffer pool reads

guymguym:
### Explain the changes

1. Add simple auth with single access_key secret_key to nsfs command. Previously the nsfs command had no authentication. This adds a single auth for single-tenancy which is the first step. Next the plan is to follow with multi-tenant configuration from a config file, and then integrate with user mapping DBs such as AD.

2. Add --forks flag option (only had an env ENDPOINTS_FORKS option so far), and also fix a bug that prevented from forks to work properly (primary had to return and not listen on same ports).

3. Fix performance regression for small requests caused by read_account() being called every request by authorize_request_policy. Instead, use the accounts_cache which we already maintained for the signature checking. (CC @jackyalbo @romayalon).

4. NSFS reads into warmup buffer to prefetch archived/cached files which are identified by being sparse (CC @eshelmarc)
 

1. NA

### Testing Instructions:
1. Anonymous: `npm -- run nsfs .`
2. Authenticated: `npm -- run nsfs --access_key 123 --secret_key XXX .`
3. Forks: `npm -- run nsfs --forks 4 .`

- [ ] Doc added/updated
- [ ] Tests added


liranmauda (approved):
Need to fix the lint, deepscan, DCO, and squash 
Other than that, LGTM.

src/endpoint/s3/s3_rest.js
```diff
@@ -216,7 +217,7 @@ async function authorize_request_policy(req) {
         return;
     }
 
-    const account = await req.object_sdk.rpc_client.account.read_account({});
+    const account = req.object_sdk.requesting_account;
```
guymguym, line 220: @jackyalbo @romayalon FYI - this is the fix we discussed, instead of read_account, it reuses the same account_cache that object_sdk uses to check the signatures.

src/sdk/namespace_fs.js
```diff
@@ -790,12 +806,28 @@ class NamespaceFS {
...
+                // and then use just a small buffer to wait for a tiny read,
+                // which will recall the file from archive or load from remote into cache,
+                // and once it returns we can continue to the full fledged read.
+                if (is_sparse_file(stat)) {
+                    dbg.log0('NamespaceFS: read_object_stream - warmup sparse file', {
+                        file_path, pos, size: stat.size, blocks: stat.blocks,
+                    });
+                    await file.read(fs_context, this.warmup_buffer, 0, 1, pos);
+                }
+
```
guymguym, line 830: @eshelmarc This is the logic to improve buffer pool usage - 
I only read 1 byte on the current pos, so it assumes that prefetching will get a whole 8MB.

src/endpoint/s3/s3_rest.js
eshelmarc: by default AFM will fetch the 4M block that includes this byte, so if you need 8M we need to set the chunck size for AFM to 12M, the other option is to set the offset to the first byte of the 4M block, in that case we can set the chunck size to 8M, this assume the block size is 4M and that NooBaa is checking for the block size.

PR #7308, opened 2023-05-24
Implement Azure blob replication logic

Neon-White:
### Explain the changes
1. Initial implementation of Azure blob log retrieval and parsing
2. Addition of a unit test for the Azure parser
3. Replaced the `ignore_fn` lambda function with the `ignore_deletions` flag, and moved the handling to the log parsers

### TODO
1. Things marked with NWTODO (Neon White TODO)

- [x] Tests added

### DIFF
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -408,6 +408,7 @@ module.exports = {
                     },
                     identity: { $ref: 'common_api#/definitions/access_key' },
                     secret: { $ref: 'common_api#/definitions/secret_key' },
+                    azure_log_access_keys: { $ref: 'common_api#/definitions/azure_log_access_keys' },
                     aws_sts_arn: {
                         type: 'string'
                     },
@@ -464,6 +465,7 @@ module.exports = {
                     },
                     identity: { $ref: 'common_api#/definitions/access_key' },
                     secret: { $ref: 'common_api#/definitions/secret_key' },
+                    azure_log_access_keys: { $ref: 'common_api#/definitions/azure_log_access_keys' },
                     aws_sts_arn: {
                         type: 'string'
                     },
--- a/src/server/bg_services/replication_log_parser.js
+++ b/src/server/bg_services/replication_log_parser.js
@@ -10,6 +10,8 @@ const pool_server = require('../system_services/pool_server');
 const dbg = require('../../util/debug_module')(__filename);
 const config = require('../../../config');
 const moment = require('moment');
+const { LogsQueryClient, LogsQueryResultStatus } = require("@azure/monitor-query");
+const { ClientSecretCredential } = require("@azure/identity");
 
 /**
  * get_log_candidates will return an object which contains the log candidates
@@ -17,17 +19,25 @@ const moment = require('moment');
  * @param {string} rule_id ID of the replication rule
  * @param {Record<any, any>} replicationconfig Replication configuration
  * @param {Number} candidates_limit Candidates limit
- * @param {(log_entry: Record<string, any>) => boolean} ignore_fn - function that returns true if the log entry should be ignored
+ * @param {boolean} sync_deletions - Whether deletions should be synced or not
  * @returns {Promise<{
  *  items: nb.ReplicationLogCandidates,
  *  done: () => Promise<void>
  * }>} Candidates
  */
-async function get_log_candidates(source_bucket_id, rule_id, replicationconfig, candidates_limit, ignore_fn = () => false) {
-    return get_aws_log_candidates(source_bucket_id, rule_id, replicationconfig, candidates_limit, ignore_fn);
+async function get_log_candidates(source_bucket_id, rule_id, replicationconfig, candidates_limit, sync_deletions) {
+    const source_bucket = system_store.data.get_by_id(source_bucket_id);
+    const endpoint_type = source_bucket.namespace.write_resource.resource.connection.endpoint_type;
+    if (endpoint_type === "AWS") {
+        return get_aws_log_candidates(source_bucket_id, rule_id, replicationconfig, candidates_limit, sync_deletions);
+    } else if (endpoint_type === "AZURE") {
+        return get_azure_log_candidates(source_bucket_id, rule_id, replicationconfig, candidates_limit, sync_deletions);
+    } else {
+        throw new Error(`REPLICATION_LOG_PARSER: Unsupported endpoint type ${endpoint_type}`);
+    }
 }
 
-async function get_aws_log_candidates(source_bucket_id, rule_id, replication_config, candidates_limit, ignore_fn) {
+async function get_aws_log_candidates(source_bucket_id, rule_id, replication_config, candidates_limit, sync_deletions) {
     const aws_log_replication_info = replication_config.log_replication_info.aws_log_replication_info;
     const { logs_bucket, prefix } = aws_log_replication_info.logs_location;
     const s3 = get_source_bucket_aws_connection(source_bucket_id, aws_log_replication_info);
@@ -50,7 +60,7 @@ async function get_aws_log_candidates(source_bucket_id, rule_id, replication_con
         }
 
         const next_log_data = await aws_get_next_log(s3, logs_bucket, next_log_entry.Contents[0].Key);
-        aws_parse_log_object(logs, next_log_data, ignore_fn);
+        aws_parse_log_object(logs, next_log_data, sync_deletions);
 
         logs_retrieved_count -= 1;
     }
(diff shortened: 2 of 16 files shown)


src/api/account_api.js
```diff
@@ -443,6 +444,7 @@ module.exports = {
                     },
                     identity: { $ref: 'common_api#/definitions/access_key' },
                     secret: { $ref: 'common_api#/definitions/secret_key' },
+                    azure_log_access_keys: { $ref: 'common_api#/definitions/azure_log_access_keys' },
```
dannyzaken, line 447: you added the support in the API, but I didn't see the use of it in the implementation in account_server. same foe `check_external_connection`

src/server/system_services/account_server.js
```diff
@@ -1192,6 +1204,7 @@ function get_account_info(account, include_connection_cache) {
             endpoint: credentials.endpoint,
             identity: credentials.access_key,
             auth_method: credentials.auth_method,
+            azure_log_access_keys: credentials.azure_log_access_keys,
```
dannyzaken, line 1207: if we consider it as sensitive then we shouldn't return it here (same as secret_key)

src/server/system_services/account_server.js
```diff
@@ -779,7 +801,8 @@ async function update_external_connection(req) {
             endpoint_type: connection.endpoint_type,
             endpoint: connection.endpoint,
             cp_code: connection.cp_code,
-            auth_method: connection.auth_method
+            auth_method: connection.auth_method,
+            azure_log_access_keys: req.rpc_params.azure_log_access_keys || undefined,
```
utkarsh-pro, line 805: Won't this be undefined if the key is not set?

src/server/system_services/account_server.js
```diff
@@ -799,7 +822,11 @@ async function update_external_connection(req) {
         },
         $set: {
             "sync_credentials_cache.$.access_key": identity,
-            "sync_credentials_cache.$.secret_key": encrypted_secret
+            "sync_credentials_cache.$.secret_key": encrypted_secret,
+            "sync_credentials_cache.$.azure_log_access_keys.azure_tenant_id": req.rpc_params.azure_log_access_keys.azure_tenant_id || undefined,
```
utkarsh-pro, line 826: Can you please check the behaviour when these fields are set to `undefined`? Actually I think I noticed a fluke with our `encode_json` which iterates over all the keys and captures the keys even if the values are set to `undefined`. I am not sure if that applies here.

src/server/system_services/bucket_server.js
```diff
@@ -1970,6 +1970,7 @@ function validate_replication(req) {
 }
 
 function normalize_replication(req) {
+    console.log('NWHITE ENTERED NORMA');
```
utkarsh-pro, line 1973: Can probably remove these logs now or maybe converted to `dbg.log2`?

src/server/system_services/account_server.js
```diff
@@ -719,6 +722,18 @@ async function add_external_connection(req) {
...
+    if (req.rpc_params.azure_log_access_keys) {
+        info.azure_log_access_keys = {
+            azure_tenant_id: req.rpc_params.azure_log_access_keys?.azure_tenant_id,
+            azure_client_id: req.rpc_params.azure_log_access_keys?.azure_client_id,
+            azure_client_secret: system_store.master_key_manager.encrypt_sensitive_string_with_master_key_id(
+                req.rpc_params.azure_log_access_keys?.azure_client_secret, req.account.master_key_id._id),
+            azure_logs_analytics_workspace_id: req.rpc_params.azure_log_access_keys?.azure_logs_analytics_workspace_id,
+        };
+    } else {
+        info.azure_log_access_keys = undefined;
```
romayalon, line 734: no need to define a variable as undefined

src/server/system_services/account_server.js
```diff
@@ -719,6 +722,18 @@ async function add_external_connection(req) {
     info.secret_key = system_store.master_key_manager.encrypt_sensitive_string_with_master_key_id(
         req.rpc_params.secret, req.account.master_key_id._id);
 
+    if (req.rpc_params.azure_log_access_keys) {
+        info.azure_log_access_keys = {
+            azure_tenant_id: req.rpc_params.azure_log_access_keys?.azure_tenant_id,
```
romayalon, line 727: you already checked that req.rpc_params.azure_log_access_keys defined, so no need to add the question mark  on every line
for instance - 
req.rpc_params.azure_log_access_keys?.azure_tenant_id,

src/server/system_services/account_server.js
```diff
@@ -1192,6 +1240,12 @@ function get_account_info(account, include_connection_cache) {
             endpoint: credentials.endpoint,
             identity: credentials.access_key,
             auth_method: credentials.auth_method,
+            azure_log_access_keys: {
+                azure_tenant_id: "Not Accesible",
```
romayalon, line 1244: why not accessible?

src/server/system_services/replication_store.js
```diff
@@ -135,16 +135,19 @@ class ReplicationStore {
...
     }
 
-    async find_aws_log_based_replication_rules() {
-        dbg.log1('find_aws_log_based_replication_rules: ');
+    async find_log_based_replication_rules() {
+        dbg.log1('find_log_based_replication_rules: ');
         const replications = await this._replicationconfigs.find({ deleted: null });
-
         const aws_reduced_replications = _.filter(replications, repl => repl.log_replication_info?.aws_log_replication_info);
+        const azure_reduced_replications = _.filter(replications, repl => repl.log_replication_info?.azure_log_replication_info);
```
romayalon, line 142: you can iterate on the replications array once instead of twice to reduce complexity time

src/server/bg_services/replication_log_parser.js
```diff
@@ -66,6 +76,79 @@ async function get_aws_log_candidates(source_bucket_id, rule_id, replication_con
...
+    // value is valid.
+    // If a continuation token does not yet exist, we'll take all logs from the past hour
+    { duration: duration },
+    {serverTimeoutInSeconds: 300}
+    );
+
+    if (query_result.status === LogsQueryResultStatus.Success) {
+        const tables_from_result = query_result.tables;
+        if (tables_from_result && tables_from_result[0].rows.length > 0) {
+            console.log("get_azure_log_candidates: Found new Azure logs:");
```
romayalon, line 122: please change console.log prints to a higher debug level since it might be very noisy

romayalon: and errors/warns to debug.error/warn

src/server/bg_services/replication_log_parser.js
```diff
@@ -66,6 +76,79 @@ async function get_aws_log_candidates(source_bucket_id, rule_id, replication_con
...
+    let continuation_token = get_continuation_token_for_rule(rule_id, replication_config);
+    let duration;
+    if (continuation_token) {
+        duration = `${(Date.now() - parseInt(continuation_token, 10))}ms`;
+    } else {
+        duration = Durations.oneHour;
+    }
+    continuation_token = Date.now();
+    const query_result = await logs_query_client.queryWorkspace(
+    monitor_workspace_id.unwrap(),
```
romayalon, line 109: indetnation

src/server/bg_services/replication_log_parser.js
```diff
@@ -66,6 +76,79 @@ async function get_aws_log_candidates(source_bucket_id, rule_id, replication_con
...
+    kusto_query,
+    // The timespan is an ISO8601 formatted time (or interval). Some common aliases
+    // are available (like OneDay, OneHour, FoutyEightHours, etc..) but any properly formatted ISO8601
+    // value is valid.
+    // If a continuation token does not yet exist, we'll take all logs from the past hour
+    { duration: duration },
```
romayalon, line 115: What's the size of the runtime object response of one-hour logs (specifically in rush hour)? I am afraid it'll be too big and we will get oom killed

romayalon: what I'm trying to say is that we need to check if one hour can get us a very big object which we will not be able to handle/crash

src/server/system_services/account_server.js
dannyzaken: we should not return it

src/api/account_api.js
```diff
@@ -652,6 +655,7 @@ module.exports = {
                                         type: 'string'
                                     },
                                     identity: { $ref: 'common_api#/definitions/access_key' },
+                                    azure_log_access_keys: { $ref: 'common_api#/definitions/azure_log_access_keys' },
```
dannyzaken, line 658: we should not return it in account_info

src/server/bg_services/replication_log_parser.js
dannyzaken: I agree. We need to test and tune it. We can do it after this PR

src/api/account_api.js
romayalon: why not?

dannyzaken: Since we consider it as sensitive as credentials, so for the same reason we don't return secret_key in account_info, we also shouldn't return this

dannyzaken (approved):

PR #7311, opened 2023-05-28
NSFS | Versioning | Suspended Mode Put Object

shirady:
### Explain the changes
1. Add NSFS Versioning Suspended mode to support put-object operation.
 
### Issues: Gaps
1. In suspended mode changes we used only "safe" filesystem functions (safe_move, safe_unlink), later we might change it according to the specific case (we also don't have concurrency tests).
2. GPFS was not tested yet with these changes.
3. We currently don't keep `XATTR_PREV_VERSION_ID` updated in case there is a deletion of a specific version id or in case we delete an old null version from the `.versions/` directory.
4. Test cases that combine delete-object and put-object will be added after the delete-object in suspended mode will be implemented.
 
#### Manual Tests:
5. Use AWS CLI and use the put-object command (here is a link to their [doc](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/put-object.html)).
6. Set the bucket to versioning Enabled and put-object of another key (here is a link to their [doc](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/put-bucket-versioning.html)).
7. Set versioning to suspended and put versions of the same object that were uploaded before (steps 4 and 5).
 
While you add objects to the bucket you can view the changes
2. Move to the bucket path (if you did it as written in the wiki it would be ' cd /nsfs/fs1/bucket-path/').
 
#### Run Unit Tests: 
2. From the core tab in the terminal run: `make root-perm-test` (You should see "110 passing" at the end of the output).
 
- [X] Doc added/updated: updated.
- [X] Tests added: NSFS Versioning Suspended mode: put-object, copy-object, and upload multipart (based on existing tests that we have in NSFS Versioning Enabled mode).

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -42,6 +42,7 @@ const XATTR_VERSION_ID = XATTR_USER_PREFIX + 'version_id';
 const XATTR_PREV_VERSION_ID = XATTR_USER_PREFIX + 'prev_version_id';
 const XATTR_DELETE_MARKER = XATTR_USER_PREFIX + 'delete_marker';
 const HIDDEN_VERSIONS_PATH = '.versions';
+const NULL_VERSION_ID = 'null';
 
 const versioning_status_enum = {
     VER_ENABLED: 'ENABLED',
@@ -1081,7 +1082,7 @@ class NamespaceFS {
                 }
                 fs_xattr = this._assign_md5_to_fs_xattr(digest, fs_xattr);
             }
-            if (!part_upload && this._is_versioning_enabled()) {
+            if (!part_upload && (this._is_versioning_enabled() || this._is_versioning_suspended())) {
                 const cur_ver_info = await this._get_version_info(fs_context, file_path);
                 fs_xattr = await this._assign_versions_to_fs_xattr(fs_context, cur_ver_info, stat, params.key, fs_xattr);
             }
@@ -1136,10 +1137,8 @@ class NamespaceFS {
                     } else {
                         await nb_native().fs.rename(fs_context, source_path, dest_path);
                     }
-                } else if (this._is_versioning_enabled()) {
-                    await this._move_to_dest_ver_enabled(fs_context, source_path, dest_path, target_file, key, open_mode);
                 } else {
-                    await this._move_to_dest_ver_suspended();
+                    await this._move_to_dest_version(fs_context, source_path, dest_path, target_file, key, open_mode);
                 }
                 break;
             } catch (err) {
@@ -1155,21 +1154,29 @@ class NamespaceFS {
         }
     }
 
-    // 1. get latest version_id if exists
-    // 2. if latest version exists -
-    //    2.1 create .versions/ if it doesnt exist
-    //    2.2 move latest version to .versions/
-    // 3. move new version to latest_ver_path (key path)
+    // 1. get latest version_id
+    // 2. if versioning is suspended - 
+    //     2.1 if version ID of the latest version is null - 
+    //       2.1.1 remove latest version
+    //     2.2 else (version ID of the latest version is unique or there is no latest version) -
+    //       2.2.1 remove a version (or delete marker) with null version ID from .versions/ (if exists)
+    //  3. if latest version exists -
+    //     versioning is enabled
+    //     OR
+    //     versioning is suspended AND latest version is a unique id (not null version)
+    //       2.1.1 create .versions/ if it doesn't exist
+    //       2.1.2 move latest version to .versions/
+    // 4. move new version to latest_ver_path (key path)
     // retry safe linking a file in case of parallel put/delete of the source path
-    async _move_to_dest_ver_enabled(fs_context, new_ver_tmp_path, latest_ver_path, upload_file, key, open_mode) {
-        dbg.log1('Namespace_fs._move_to_dest_ver_enabled:', new_ver_tmp_path, latest_ver_path, upload_file);
-
+    async _move_to_dest_version(fs_context, new_ver_tmp_path, latest_ver_path, upload_file, key, open_mode) {
+        dbg.log1('Namespace_fs._move_to_dest_version:', new_ver_tmp_path, latest_ver_path, upload_file);
         let gpfs_options;
         const is_gpfs = this._is_gpfs(fs_context);
         let retries = config.NSFS_RENAME_RETRIES;
         for (;;) {
             try {
                 const new_ver_info = !is_gpfs && await this._get_version_info(fs_context, new_ver_tmp_path);
+                // get latest version_id if exists
                 const latest_ver_info = await this._get_version_info(fs_context, latest_ver_path);
                 const versioned_path = latest_ver_info && this._get_version_path(key, latest_ver_info.version_id_str);
                 const versioned_info = latest_ver_info && await this._get_version_info(fs_context, versioned_path);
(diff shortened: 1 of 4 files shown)


src/sdk/namespace_fs.js
```diff
@@ -2152,6 +2225,25 @@ class NamespaceFS {
         };
     }
 
+    // We can have only one versioned object with null version ID per key.
+    // It can be latest version, old version in .version/ directory or delete marker
+    // This function removes an object or delete marker with a null version ID inside .version/ directory 
+    async _delete_null_version_from_versions_directory(key, fs_context, is_gpfs) {
```
jackyalbo, line 2231: seems redundant to pass both fs_context and is_gpfs. just call is_gpfs inside with fs_context.

src/sdk/namespace_fs.js
```diff
@@ -1463,9 +1521,16 @@ class NamespaceFS {
...
-        fs_xattr = Object.assign(fs_xattr || {}, {
-            [XATTR_VERSION_ID]: this._get_version_id_by_stat(new_ver_stat)
-        });
+        if (this._is_versioning_enabled()) {
+            fs_xattr = Object.assign(fs_xattr || {}, {
+                [XATTR_VERSION_ID]: this._get_version_id_by_stat(new_ver_stat)
+            });
+        } else if (this._is_versioning_suspended()) {
+            fs_xattr = Object.assign(fs_xattr || {}, {
+                [XATTR_VERSION_ID]: this._get_version_id_suspended()
```
jackyalbo, line 1530: not sure if we must have it in a function but if so let's call something like get_version_null or get_null_version so we will know what to expect

shirady: I change it to a const variable instead of a function and change its name.

src/sdk/namespace_fs.js
```diff
@@ -2233,6 +2323,15 @@ class NamespaceFS {
         }
     }
 
+    async safe_unlink(fs_context, src_path, src_ver_info, gpfs_options) {
```
romayalon, line 2326: you can call this function from delete api of versioning enabled mode

src/sdk/namespace_fs.js
```diff
@@ -1014,12 +1015,70 @@ class NamespaceFS {
...
+                        latest_ver_info, open_mode, undefined, versioned_info) :
+                        undefined;
+
+                dbg.log1('Namespace_fs._move_to_dest_ver_suspended:', latest_ver_info, new_ver_info, gpfs_options);
+
+                if (latest_ver_info) {
+                    dbg.log1('NamespaceFS._move_to_dest_ver_suspended: latest version exist');
+                    if (latest_ver_info.version_id_str === NULL_VERSION_ID) {
+                        dbg.log1('NamespaceFS._move_to_dest_ver_suspended: version ID of the latest version is null - the file will be unlinked');
+                        await this.safe_unlink(fs_context, latest_ver_path, latest_ver_info, gpfs_options);
```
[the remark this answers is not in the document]
romayalon: whatever you want, it depends if you want to work on the safe operations changes on a different PR or here

src/sdk/namespace_fs.js
```diff
@@ -1014,12 +1015,70 @@ class NamespaceFS {
...
+    // 2. if latest version exists -
+    //    2.1 if version ID of the latest version is null - 
+    //       2.1.1 remove latest version
+    //    2.2 if version ID of the latest version is a unique ID - 
+    //       2.2.1 create .versions/ if it doesn't exist
+    //       2.2.2 move latest version to .versions/
+    // 3. remove a version (or delete marker) with null version ID from .versions/ (if exists)
+    // 4. move new version to latest_ver_path (key path)
+    // retry safe linking a file in case of parallel put/delete of the source path
+    async _move_to_dest_ver_suspended(fs_context, new_ver_tmp_path, latest_ver_path, upload_file, key, open_mode) {
```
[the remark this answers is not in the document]
romayalon: I agree that we need to think hard about how to make it readable
I'm thinking about a day that we will find a bug in this logic and we will forget to maintain both of them (we will probably miss one of them)
we can try to think of ideas on Slack

src/sdk/namespace_fs.js
```diff
@@ -1151,42 +1155,43 @@ class NamespaceFS {
...
+                        dbg.log1('NamespaceFS._move_to_dest_version: version ID of the latest version is null - the file will be unlinked');
+                        await this.safe_unlink(fs_context, latest_ver_path, latest_ver_info, gpfs_options);
+                    } else {
+                        dbg.log1('NamespaceFS._move_to_dest_version version ID of the latest version is a unique ID - the file will be moved it to .versions/ directory');
+                        await this._make_path_dirs(versioned_path, fs_context);
+                        await this.safe_move(fs_context, latest_ver_path, versioned_path, latest_ver_info,
+                            gpfs_options && gpfs_options.move_to_versions);
+                    }
                 }
+                // remove a version (or delete marker) with null version ID from .versions/ (if exists)
```
romayalon, line 1172: any reason for this structure?
If possible I would prefer 
if (versioning_enabled) {
...
} else if (versioning_suspended){
if (latest is null) unlink latest
else delete_null_version_from_versions_dir()
} else {
throw err
}

src/sdk/namespace_fs.js
```diff
@@ -1151,42 +1155,43 @@ class NamespaceFS {
...
                     if (err.message !== posix_unlink_retry_err && err.code !== gpfs_unlink_retry_catch) throw err;
-                    dbg.warn('Namespace_fs._move_to_dest_ver_enabled: unable to delete new version tmp file, ignoring...');
+                    dbg.warn('Namespace_fs._move_to_dest_version: unable to delete new version tmp file, ignoring...');
                 }
                 break;
             } catch (err) {
                 retries -= 1;
                 if (retries <= 0 || !this.should_retry_link_unlink(is_gpfs, err)) throw err;
-                dbg.warn(`NamespaceFS._move_to_dest_ver_enabled retrying retries=${retries}` +
+                dbg.warn(`NamespaceFS._move_to_dest_ver_suspended retrying retries=${retries}` +
```
romayalon, line 1187:
```suggestion
                dbg.warn(`NamespaceFS._move_to_dest_version retrying retries=${retries}` +
```

src/sdk/namespace_fs.js
```diff
@@ -2328,6 +2341,37 @@ class NamespaceFS {
...
 
+    // We can have only one versioned object with null version ID per key.
+    // It can be latest version, old version in .version/ directory or delete marker
+    // This function removes an object version or delete marker with a null version ID inside .version/ directory 
+    async _delete_null_version_from_versions_directory(key, fs_context) {
+        const is_gpfs = this._is_gpfs(fs_context);
+        let retries = config.NSFS_RENAME_RETRIES;
+        const null_versioned_path = this._get_version_path(key, NULL_VERSION_ID);
+        await this._check_path_in_bucket_boundaries(fs_context, null_versioned_path);
+        const null_versioned_path_info = await this._get_version_info(fs_context, null_versioned_path);
```
romayalon, line 2352: get_version_info() should be inside the for

src/sdk/namespace_fs.js
```diff
@@ -2409,6 +2453,15 @@ class NamespaceFS {
         }
     }
 
+    async safe_unlink(fs_context, src_path, src_ver_info, gpfs_options) {
+        if (this._is_gpfs(fs_context)) {
+            const { src_file = undefined, dir_file = undefined } = gpfs_options;
+            if (dir_file) await this.safe_unlink_gpfs(fs_context, src_path, src_file, dir_file);
```
romayalon, line 2459: why not do anything if dir_file is undefined?

src/sdk/namespace_fs.js
romayalon: is it possible?

romayalon (approved):

src/sdk/namespace_fs.js
```diff
@@ -1166,41 +1173,69 @@ class NamespaceFS {
...
                 if (gpfs_options) await this._close_files_gpfs(fs_context, gpfs_options.move_to_dst, open_mode);
             }
         }
     }
 
-    async _move_to_dest_ver_suspended() {
-        throw new Error('TODO');
-    }
+    // if (this._is_versioning_enabled()) {
+    //     if (latest_ver_info) {
```
romayalon, line 1216: just remove this comment :)

shirady: Removed, thank you

PR #7313, opened 2023-05-29
Tiering with storage class support - TTL based worker + TMFS blocks migration

tangledbytes:
### Explain the changes
Context: NooBaa in its write flows marks an object for migration so that TMFS can migrate to another tier2 and NooBaa's read flow can recall the block from the other tier using TMFS calls but this would leave a bunch of objects on the disk (cache). 

This PR adds a BG Worker which iterates over the chunks associated with a bucket (the bucket needs to have at least 2 tiers and for the case of TMFS, the second tier **HAS** to share the same pool) and select the chunks which haven't been read for longer than 30 mins (configurable) and move them to second tier and evict the associated the blocks.

#### Tested In the following way
1. Upload 2 objects, testobject1 and testobject2 of sizes 2G and 4G respectively.
2. Let the BG worker do its job.
3. Run `find storage/backingstores -name '*.data' -type f |  xargs -n1 -I{} sh -c "echo 'file: {} => '; xattr {}; echo"` which correctly shows `user._trigger.migrate` being set on the files.
4. Run `psql -d nbcore -U postgres -c "SELECT data FROM datachunks;"` shows the second tier as the tier selected for `datachunks`.
5. TTL Worker does **not** iterates over the blocks which are in second tier.

- [x] Doc added/updated
- [x] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -687,9 +687,9 @@ config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS = 60 * 60 * 1000; // 1 hour
 // BLOCK STORE FS      //
 /////////////////////////
 
-config.BLOCK_STORE_FS_TIER2_ENABLED = false;
+config.BLOCK_STORE_FS_TMFS_ENABLED = false;
 config.BLOCK_STORE_FS_MAPPING_INFO_ENABLED = false;
-config.BLOCK_STORE_FS_TIER2_ALLOW_MIGRATED_READS = true;
+config.BLOCK_STORE_FS_TMFS_ALLOW_MIGRATED_READS = true;
 
 config.BLOCK_STORE_FS_XATTR_BLOCK_MD = 'user.noobaa.block_md';
 config.BLOCK_STORE_FS_XATTR_QUERY_MIGSTAT = 'user._query.migstat';
@@ -697,7 +697,10 @@ config.BLOCK_STORE_FS_XATTR_TRIGGER_RECALL = 'user._trigger.recall';
 config.BLOCK_STORE_FS_XATTR_TRIGGER_MIGRATE = 'user._trigger.migrate';
 config.BLOCK_STORE_FS_XATTR_TRIGGER_PREMIGRATE = 'user._trigger.premigrate';
 
-
+config.TIERING_TTL_WORKER_ENABLED = false;
+config.TIERING_TTL_WORKER_BATCH_SIZE = 1000;
+config.TIERING_TTL_WORKER_BATCH_DELAY = 1 * 60 * 1000; // 1 minutes
+config.TIERING_TTL_MS = 30 * 60 * 1000; // 30 minutes
 
 /////////////////////
 //                 //
--- a/src/agent/block_store_services/block_store_fs.js
+++ b/src/agent/block_store_services/block_store_fs.js
@@ -16,6 +16,10 @@ const BlockStoreBase = require('./block_store_base').BlockStoreBase;
 const get_block_internal_dir = require('./block_store_base').get_block_internal_dir;
 const { RpcError } = require('../../rpc');
 
+const TMFS_STATE_MIGRATED = 'MIGRATED';
+const TMFS_STATE_PREMIGRATED = 'PREMIGRATED';
+const TMFS_STATE_RESIDENT = 'RESIDENT';
+
 class BlockStoreFs extends BlockStoreBase {
 
     constructor(options) {
@@ -27,7 +31,7 @@ class BlockStoreFs extends BlockStoreBase {
         this.usage_path = path.join(this.root_path, 'usage');
 
         this.fs_context = {
-            disable_ctime_check: config.BLOCK_STORE_FS_TIER2_ENABLED
+            disable_ctime_check: config.BLOCK_STORE_FS_TMFS_ENABLED
         };
     }
 
@@ -92,7 +96,7 @@ class BlockStoreFs extends BlockStoreBase {
         let block_file;
 
         try {
-            if (config.BLOCK_STORE_FS_TIER2_ENABLED) {
+            if (config.BLOCK_STORE_FS_TMFS_ENABLED) {
                 block_file = await nb_native().fs.open(fs_context, block_path);
                 const stat = await block_file.stat(
                     fs_context,
@@ -101,9 +105,9 @@ class BlockStoreFs extends BlockStoreBase {
 
                 const migstat = JSON.parse(stat.xattr[config.BLOCK_STORE_FS_XATTR_QUERY_MIGSTAT] || '{}');
 
-                if (migstat.State === 'MIGRATED') {
+                if (migstat.State === TMFS_STATE_MIGRATED) {
                     // if not yet trying to premigrate, try now.
-                    if (migstat.TargetState !== 'PREMIGRATED') {
+                    if (migstat.TargetState !== TMFS_STATE_PREMIGRATED) {
                         await block_file.replacexattr(fs_context, {
                             [config.BLOCK_STORE_FS_XATTR_TRIGGER_RECALL]: 'now'
                         });
@@ -111,7 +115,7 @@ class BlockStoreFs extends BlockStoreBase {
 
                     // if not allowed to read migrated blocks, throw error - temporary fix until
                     // we move to migstat polling.
-                    if (!config.BLOCK_STORE_FS_TIER2_ALLOW_MIGRATED_READS) {
+                    if (!config.BLOCK_STORE_FS_TMFS_ALLOW_MIGRATED_READS) {
                         throw new RpcError('MIGRATED', `block is migrated`);
                     }
                 }
@@ -174,7 +178,7 @@ class BlockStoreFs extends BlockStoreBase {
         if (!is_test_block) {
 
             // set xattr to trigger migration of file to underlying tier
-            if (config.BLOCK_STORE_FS_TIER2_ENABLED) {
+            if (config.BLOCK_STORE_FS_TMFS_ENABLED) {
                 xattr[config.BLOCK_STORE_FS_XATTR_TRIGGER_MIGRATE] = 'now';
                 xattr_need_fsync = true;
             }
@@ -254,6 +258,86 @@ class BlockStoreFs extends BlockStoreBase {
             };
             this._update_usage(usage);
         }
+    }
+
+
+    /**
+     * @param {string[]} block_ids
+     * @param {string} storage_class
+     * @returns {Promise<{ moved_block_ids: string[] }>}
+     */
+    async _move_blocks_to_storage_class(block_ids, storage_class) {
+        if (storage_class === 'GLACIER') {
+            return this._move_blocks_to_glacier(block_ids);
+        }
+        if (storage_class === 'STANDARD') {
+            // Nothing to do for now
+            return Promise.resolve({ moved_block_ids: [] });
+        }
+
+        throw new Error(`unsupported storage class ${storage_class}`);
+    }
+
+    /**
+     * @param {string[]} block_ids
+     * @returns {Promise<{ moved_block_ids: string[] }>}
+     */
+    async _move_blocks_to_glacier(block_ids) {
+        const moved_block_ids = [];
+        if (!config.BLOCK_STORE_FS_TMFS_ENABLED) return { moved_block_ids };
+
+        await P.map_with_concurrency(10, block_ids, async block_id => {
+            try {
+                if (await this._move_block_to_glacier(block_id)) moved_block_ids.push(block_id);
+            } catch (err) {
+                dbg.warn(`_move_block_to_glacier block ${block_id} failed due to`, err);
+            }
+        });
+
+        return { moved_block_ids };
+    }
+
+    /**
+     * @note Supports only Tier2/TMFS FS_ARCHIVE for others will 
+     * simply return false
+     * @param {string} block_id 
+     * @returns {Promise<boolean>}
+     */
+    async _move_block_to_glacier(block_id) {
+        return this._move_block_to_tmfs(block_id);
+    }
+
+    async _move_block_to_tmfs(block_id) {
+        const fs_context = this.fs_context;
+        const block_path = this._get_block_data_path(block_id);
+        let evicting = false;
+        let file = null;
+
+        try {
+            file = await nb_native().fs.open(fs_context, block_path);
+            const stat = await file.stat(fs_context, { xattr_get_keys: [config.BLOCK_STORE_FS_XATTR_QUERY_MIGSTAT] });
+            const migstat = JSON.parse(stat?.xattr?.[config.BLOCK_STORE_FS_XATTR_QUERY_MIGSTAT] || '{}');
+
+            // If the block is in PREMIGRATED state or is in RESIDENT state, we need to MIGRATE it back to the
+            // tape storage (eviction).
+            if (migstat.State === TMFS_STATE_PREMIGRATED || migstat.State === TMFS_STATE_RESIDENT) {
+                // it is possible that we requested this in the previous run
+                // as well so we need to make sure we do not trigger it again
+                if (migstat.TargetState !== TMFS_STATE_MIGRATED) {
+                    await file.replacexattr(fs_context, {
+                        [config.BLOCK_STORE_FS_XATTR_TRIGGER_MIGRATE]: 'now'
+                    });
+
+                    evicting = true;
+                }
+            }
+        } catch (err) {
+            ignore_not_found(err);
+        } finally {
+            if (file) await file.close(fs_context);
+        }
+
+        return evicting;
     }
 
     _get_usage() {
--- a/src/api/block_store_api.js
+++ b/src/api/block_store_api.js
@@ -252,6 +252,36 @@ module.exports = {
             },
         },
 
+        move_blocks_to_storage_class: {
+            method: 'POST',
+            params: {
+                type: 'object',
+                required: ['block_ids', 'storage_class'],
+                properties: {
+                    block_ids: {
+                        type: 'array',
+                        items: {
+                            type: 'string'
+                        }
+                    },
+                    storage_class: {
+                        $ref: 'common_api#/definitions/storage_class_enum'
+                    }
+                },
+            },
+            reply: {
+                type: 'object',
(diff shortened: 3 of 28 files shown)


src/server/bg_workers.js
```diff
@@ -227,6 +228,13 @@ function run_master_workers() {
     } else {
         dbg.warn('KEY ROATATION NOT ENABLED');
     }
+
+    if (config.BLOCK_STORE_FS_TIER2_ENABLED) {
+        register_bg_worker(new Tier2BlockEvicter({
```
guymguym, line 233: Please rename the class and file to "Tier2BlockEvictor" (with `o`)

src/server/bg_services/tier2_block_evicter.js
```diff
@@ -0,0 +1,134 @@
...
+
+        try {
+            // TODO: Improve the iteration here - Instead of iterating
+            // over all the blocks, we should iterate over the blocks
+            // that have `node_type` set to `BLOCK_STORE_FS`.
+            const blocks = await this.iterate_all_blocks(
+                this.marker,
+                config.AGENT_BLOCKS_RECLAIMER_BATCH_SIZE,
+                false /* deleted_only */
+            );
```
guymguym, line 49: The question is if we need to add time conditions to the evictor, so that it won't immediately evict blocks that were just read and premigrated. We might need to set a configurable time before it should get remigrated to avoid thrashing.

guymguym (approved):
@utkarsh-pro The structure looks good. 
Notice the comment on adding a time condition to the evictor.

src/server/bg_services/tier2_block_evicter.js
utkarsh-pro: @guymguym how about something like this?

```diff
async populate_block_evicter_blocks(blocks) {
    if (!blocks || !blocks.length) return;

    const db_blocks = await this.populate_nodes_for_blocks(blocks);

    // Filter out the blocks which have `node_type` set to `BLOCK_STORE_FS`.
-   return _.filter(db_blocks, block => block.to_block_md().node_type === 'BLOCK_STORE_FS');
+   return _.filter(db_blocks, block => {
+       const is_blocks_fs = block.to_block_md().node_type === 'BLOCK_STORE_FS';
+       if (!is_blocks_fs) return false;
+
+       // @ts-ignore
+       const last_read = block.chunk.chunk_db.tier_lru;
+       if (last_read instanceof Date) {
+           const curr_time = moment();
+           if (curr_time.diff(last_read, 'minutes') < config.TIER2_BLOCK_EVICTER_MIN_LRU_TIME) {
+               return false;
+           }
+       }
+ 
+       return true;
+     });
}
```

guymguym (changes requested):
Hey @utkarsh-pro 
This PR is superb, but I raised several changes which are important (+ some cosmetic). 
I'd love to discuss it further.
!!

src/agent/block_store_services/block_store_fs.js
```diff
@@ -249,6 +249,60 @@ class BlockStoreFs extends BlockStoreBase {
...
+
+        if (config.BLOCK_STORE_FS_TIER2_ENABLED) {
+            try {
+                file = await nb_native().fs.open(fs_context, block_path);
+                const stat = await file.stat(fs_context, { xattr_get_keys: [config.BLOCK_STORE_FS_XATTR_QUERY_MIGSTAT] });
+                const migstat = JSON.parse(stat?.xattr?.[config.BLOCK_STORE_FS_XATTR_QUERY_MIGSTAT] || '{}');
+
+                // If the block is in PREMIGRATED state, we need to MIGRATE it back to the
+                // tape storage (eviction).
+                if (migstat.State === 'PREMIGRATED') {
```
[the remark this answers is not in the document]
utkarsh-pro: @guymguym @piste-jp-ibm when is a file marked `RESIDENT`?

I had the following state transition assumptions:

`* - (write)-> MIGRATED - (recall) -> PREMIGRATED - (evict) -> MIGRATED`

src/api/block_store_api.js
```diff
@@ -252,6 +252,33 @@ module.exports = {
             },
         },
 
+        evict_blocks: {
```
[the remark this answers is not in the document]
utkarsh-pro: That.. actually does make sense. I will rename this.

utkarsh-pro:
@guymguym lemme try to summarize the proposed changes (specifically storage class) here:

- The idea is to have the construct of storage classes in our tiers. Each tier can have <= 1 storage class specified.
- Storage class will just communicate some additional details to the underlying block store. This detail will be used by functions like `move_blocks_to_storage_class`.
- `move_blocks_to_storage_class` must consume block ids and storage class and should try to move it to that storage class doing _something_ (here something would be different for each block store which plans to support this).
- TTL worker should invoke `map_client.run()` which will do all of the heavy lifting.
- Read flow already moves the chunks across the tiers. We will need to modify flow (maybe `map_client.read_object_mapping`?) such that it also starts invoking `move_blocks_to_storage_class`.

**Clarifications**
1. Are storage classes transparent to S3? By "transparent" I mean that S3 responses will always mention `StorageClass: Standard`?

guymguym:
@utkarsh-pro I think you got it, just two notes:

> - TTL worker should invoke map_client.run() which will do all of the heavy lifting.
> - Read flow already moves the chunks across the tiers. We will need to modify flow (maybe map_client.read_object_mapping?) such that it also starts invoking move_blocks_to_storage_class.

Both TTL worker and TTF worker already call build_chunks and therefore invoke map_client, so there is no need to add such a call in the workers, only to make this call also handle the storage class move as I suggested in this comment - https://github.com/noobaa/noobaa-core/pull/7313#discussion_r1223705849

> 1. Are storage classes transparent to S3? By "transparent" I mean that S3 responses will always mention StorageClass: Standard?

Yes. For now we only handle the internal mappings. Exposing tiers to the clients is a whole new challenge.

src/agent/block_store_services/block_store_fs.js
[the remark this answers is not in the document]
guymguym: That's right @utkarsh-pro AFAIU `RESIDENT` will only be file files were written but never been migrated. Currently we always migrate blocks right after write+fsync, so that's going to be needed ATM. I do think that deferring the migration of freshly written blocks to the worker to batch them might be a desirable option to use disks as a buffer for the archive tier, and in that case we would change to avoid migrating on write, and then need the worker to also handle `RESIDENT` blocks. But we can change it later too, I just noticed it now.

src/sdk/map_client.js
```diff
@@ -557,6 +560,43 @@ class MapClient {
...
+                }
+            }
+        }
+
+        if (blocks.length === 0) return;
+
+        const blocks_by_agent = _.groupBy(blocks, 'address');
+        await P.map(_.keys(blocks_by_agent), async agent_address => {
+            const blocks_for_agent = blocks_by_agent[agent_address];
+            const block_ids = _.map(blocks_for_agent, block => block._id?.toString());
```
guymguym, line 584: is there a case where _id is undefined? is that valid? if so we might want to filter it out and not leave those undefined values in the array.

src/agent/block_store_services/block_store_fs.js
```diff
@@ -98,9 +102,9 @@ class BlockStoreFs extends BlockStoreBase {
 
                 const migstat = JSON.parse(stat.xattr[config.BLOCK_STORE_FS_XATTR_QUERY_MIGSTAT] || '{}');
 
-                if (migstat.State === 'MIGRATED') {
+                if (migstat.State === TIER2_STATE_MIGRATED) {
                     // if not yet trying to premigrate, try now.
-                    if (migstat.TargetState !== 'PREMIGRATED') {
+                    if (migstat.TargetState !== TIER2_STATE_MIGRATED) {
```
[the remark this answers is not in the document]
utkarsh-pro: Whoops!

src/api/common_api.js
```diff
@@ -1188,5 +1188,10 @@ module.exports = {
                 }
             },
         },
+
+        storage_class_enum: {
+            type: 'string',
+            enum: ['STANDARD', 'FS_ARCHIVE', 'S3_GLACIER']
```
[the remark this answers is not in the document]
nimrod-becker: I think its good, going with enum means we can add other stuff later and yes we could expose this on the BackingStores at some point. I think marking as class instead of specific storage types lets you control better how you want to use class X (and maybe for me class X would be used differently).

And again, we can always add more values to the enum if there is a need for more class seperation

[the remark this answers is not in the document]
jackyalbo: @guymguym If I understand your question is whether we want a big list of all supported storage classes of all different vendors or only the storage classes levels applying to all possible. I have to admit that using levels can make it more clear. I'm seeing that Google support 4 levels: STANDARD, NEARLINE, COLDLINE and ARCHIVE. maybe we should do the same?
https://cloud.google.com/storage/docs/storage-classes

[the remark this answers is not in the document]
guymguym: Yeah @jackyalbo that was my though too. I wonder if trying to define "abstract" storage classes will be successful, or we will find ourselves struggling to decide how to map the different providers classes to our abstract classes. For example, what should "S3 Glacier Instant Retrieval" map to vs "S3 Glacier Flexible Retrieval" vs. "S3 Glacier Deep Archive"...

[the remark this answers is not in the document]
utkarsh-pro: I think maybe we can have more generic names here and then later on can introduce more specific names if a need arises.

The only issue I think that can occur with the usage of generic names like `ARCHIVE` is that if we have a backing store supporting 2 types of archiving then we will need a way to disambiguate.

How about having `STANDARD` and `ARCHIVE` for now? Here `STANDARD` and `undefined` (i.e unset) storage class can be treated similarly.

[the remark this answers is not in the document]
guymguym: @utkarsh-pro I think your proposal is good. Lets go with STANDARD and ARCHIVE enums only for now. We can always add more later.

config.js
```diff
@@ -657,7 +657,10 @@ config.BLOCK_STORE_FS_XATTR_TRIGGER_RECALL = 'user._trigger.recall';
 config.BLOCK_STORE_FS_XATTR_TRIGGER_MIGRATE = 'user._trigger.migrate';
 config.BLOCK_STORE_FS_XATTR_TRIGGER_PREMIGRATE = 'user._trigger.premigrate';
 
-
+config.TIERING_TTL_WORKER_ENABLED = true;
```
guymguym, line 660: I think this should be false by default, otherwise it will affect any existing buckets with two+ tiers. Right?

src/agent/block_store_services/block_store_fs.js
```diff
@@ -249,6 +253,88 @@ class BlockStoreFs extends BlockStoreBase {
...
+     * @returns {Promise<{ moved_block_ids: string[] }>}
+     */
+    async _move_blocks_to_storage_class(block_ids, storage_class) {
+        if (storage_class === 'FS_ARCHIVE') {
+            return this._move_blocks_to_fs_archive(block_ids);
+        }
+        if (storage_class === 'STANDARD') {
+            // Nothing to do for now
+            return Promise.resolve({ moved_block_ids: [] });
+        }
```
guymguym, line 269: Is this api being called when moving the chunks back to first tier?
If so it's better if we actually trigger a recall here, because then our actual reads will have a chance to already find the data ready.

src/sdk/map_client.js
```diff
@@ -557,6 +563,52 @@ class MapClient {
             }));
     }
 
+    async move_blocks_to_storage_class() {
+        // Block movement is not done if move_to_tier was not requested or if insufficient
+        // previous tier information is provided.
+        if (!this.move_to_tier || !this.current_tiers || this.current_tiers.length !== this.chunks.length) return;
```
guymguym, line 569: I think the last check is more of an error
```suggestion
        if (!this.move_to_tier || !this.current_tiers) return;
        if (this.current_tiers.length !== this.chunks.length) throw new Error('...');
```

src/server/bg_services/tier_ttl_worker.js
```diff
@@ -0,0 +1,123 @@
...
+                sort_direction: 1,
+                max_tier_lru: moment().subtract(config.TIERING_TTL_MS, 'ms').toDate(),
+            });
+            if (!chunk_ids.length) continue;
+
+            await this._build_chunks(
+                chunk_ids,
+                next_tier_id,
+                false,
+                chunk_ids.map(() => previous_tier._id)
```
guymguym, line 94: if we always have just a single tier here, then do we really want to send an array?
we could just send the previous_tier_id just like we do for the next_tier_id.
I think maybe this array of current_tiers feels a bit too much.

src/server/bg_services/tier_ttl_worker.js
```diff
@@ -0,0 +1,123 @@
...
+                (bucket.tiering && bucket.tiering.tiers.length > 1)
+            ));
+    }
+
+    async _ttl_move_chunks(buckets) {
+        const chunks_to_move = config.TIERING_TTL_WORKER_BATCH_SIZE;
+
+        for (const bucket of buckets) {
+            await node_allocator.refresh_tiering_alloc(bucket.tiering, this.last_run);
+            const tiering_status = node_allocator.get_tiering_status(bucket.tiering);
```
guymguym, line 72: notice that we track the usage of every tier based on its resources.
this means that for the second tier that is marked as storage_class=ARCHIVE we should fix the tier capacity calculation to be "infinite", like we do for cloud pools, so that our code will know it can use it although the FS reported available capacity is low.

src/agent/block_store_services/block_store_fs.js
utkarsh-pro: Yes it is. I also considered doing the same but then decided to skip doing it in this PR and rather address it in another PR which:
1. Removes the responsibility of "archiving" from the write flow because TTL worker can take care of that.
2. Adds the recall here.

Basically a PR dedicated to Read/Write flow. WDYT?

src/server/bg_services/tier_ttl_worker.js
utkarsh-pro: I made this an array under the assumption that the callers of the scrubber.build_chunks do not guarantee that the current tier of all the chunks is same.

If the above assumption is not true then maybe we can switch to a single value parameter.

src/agent/block_store_services/block_store_fs.js
guymguym: sure, that makes sense. lets keep that focus.

src/api/common_api.js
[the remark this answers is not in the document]
guymguym: We talked and agreed to align with the storage class enum of S3, and for now we will define just STANDARD and GLACIER and add more as needed.

src/agent/block_store_services/block_store_fs.js
```diff
@@ -256,6 +260,86 @@ class BlockStoreFs extends BlockStoreBase {
...
+
+    /**
+     * @param {string[]} block_ids
+     * @param {string} storage_class
+     * @returns {Promise<{ moved_block_ids: string[] }>}
+     */
+    async _move_blocks_to_storage_class(block_ids, storage_class) {
+        if (storage_class === 'GLACIER') {
+            return this._move_blocks_to_glacier(block_ids);
+        }
```
guymguym, line 272: Just to avoid confusion, I would change to not call these block_store_fs methods with *glacier in their name, because while the storage class is called glacier, we are translating it to TMFS. So I think I would so something like this here, and if TMFS is not enabled we should probably continue and throw the error (unsupported storage class ) because we don't have a glacier tier to move to. WDYT?

```suggestion
        if (storage_class === 'GLACIER') {
            if (config.BLOCK_STORE_FS_TMFS_ENABLED) {
                return this._move_blocks_to_tmfs(block_ids);
            }
        }
```

src/sdk/map_client.js
```diff
@@ -557,6 +563,53 @@ class MapClient {
...
+        const blocks = [];
+        const target_storage_class = this.move_to_tier.storage_class;
+        const target_attached_pools = this.move_to_tier.mirrors.map(mirror => mirror.spread_pools.map(pool => String(pool._id))).flat();
+
+        for (const [idx, chunk] of this.chunks.entries()) {
+            const tier_before_move = this.current_tiers[idx];
+            const current_attached_pools = tier_before_move.mirrors.map(mirror => mirror.spread_pools.map(pool => String(pool._id))).flat();
+
+            if (
+                !js_utils.compare_unordered(target_attached_pools, current_attached_pools, true) ||
```
guymguym, line 581: how about using _.xor() instead of creating a special function for this esoteric case?

src/server/bg_services/scrubber.js
```diff
@@ -73,7 +73,8 @@ async function build_chunks(req) {
     const chunk_ids = _.map(req.rpc_params.chunk_ids, id => MDStore.instance().make_md_id(id));
     const tier = req.rpc_params.tier && system_store.data.get_by_id(req.rpc_params.tier);
     const evict = req.rpc_params.evict;
-    const builder = new MapBuilder(chunk_ids, tier, evict);
+    const current_tiers = _.map(req.rpc_params.current_tiers, id => system_store.data.get_by_id(id));
```
guymguym, line 76: BTW a minor comment - lodash _.map is not really needed that much anymore, because there's a standard Array.map. Difference is that _.map also works on objects and nulls, but for arrays using the standard map function is enough. No need to go changing it all, but going forward we can try to use the standard js functions.

src/sdk/map_client.js
utkarsh-pro: I .. did not know about `xor`

src/server/bg_services/scrubber.js
utkarsh-pro: Yeah, I did it only for consistency but now that I think, I could have made it consistent by changing the line above to use the native function 🤦 .

guymguym (approved):
@utkarsh-pro All in all this is absolutely fantastic! I added couple of small comments, but it can also be addressed in next PRs. Godspeed.

src/server/bg_services/scrubber.js
guymguym: Don't really make changes for this comment, I just wanted to note that for the future. Resolving.

PR #7316, opened 2023-05-30
Objects versions Replication - Step 1

liranmauda:
### Explain the changes
- Adding sync_versions the bucket_api and to the replication_configuration_schema
- Adding list_versioned_buckets_and_compare which will currently only list the source
- Adding list_objects_versions
- Adding _object_grouped_by_key_and_omitted to group each key and its versions, and omit the last key if needed.
- Add a function (_get_next_key_marker) to get the desired next key marker.

*This PR is a prep, the code is not accessible. will add unit tests on a later PR, when we have the code ready*
*With this PR we should expect all the regular replication tests to work*

### DIFF
--- a/src/server/bg_services/replication_scanner.js
+++ b/src/server/bg_services/replication_scanner.js
@@ -36,7 +36,7 @@ class ReplicationScanner {
     constructor({ name, client }) {
         this.name = name;
         this.client = client;
-        this._scanner_sem = new Semaphore(config.REPLICATION_SEMAPHORE_CAP, {
+        this.scanner_semaphore = new Semaphore(config.REPLICATION_SEMAPHORE_CAP, {
             timeout: config.REPLICATION_SEMAPHORE_TIMEOUT,
             timeout_error_code: 'REPLICATION_ITEM_TIMEOUT',
             verbose: true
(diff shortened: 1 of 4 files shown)


src/server/bg_services/replication_scanner.js
```diff
@@ -194,6 +232,43 @@ class ReplicationScanner {
         }
     }
 
+    // list_objects_versions will list all the objects with the versions, continuation_token is the key marker.
+    async list_objects_versions(bucket_name, prefix, continuation_token) {
```
romayalon, line 236: add here a comment that we prefer keeping the cont_token term due to re-use of db properties but this is really a key marker

romayalon (approved):

src/server/bg_services/replication_scanner.js
```diff
@@ -88,23 +88,35 @@ class ReplicationScanner {
...
             const cur_dst_cont_token = (rule.rule_status && rule.rule_status.dst_cont_token) || '';
 
-            const { keys_sizes_map_to_copy, src_cont_token, dst_cont_token } = await this.list_buckets_and_compare(src_bucket.name,
-                dst_bucket.name, prefix, cur_src_cont_token, cur_dst_cont_token);
+            let src_cont_token;
+            let dst_cont_token;
+            let keys_sizes_map_to_copy;
+
+            // eslint-disable-next-line no-constant-condition
+            if (false) { //rule.sync_versions) {
```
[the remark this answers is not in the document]
liranmauda: Sure, will add a comment on the next PR.

src/server/bg_services/replication_scanner.js
liranmauda: Sure, will add a comment on the next PR.

PR #7319, opened 2023-06-01
Fix #7314: Use mtime instead of ctime when tier2 FS is enabled.

### LINKED ISSUE #7314: Skip `ctime` check in the Tier-2 environment

At this time, our filesystem managing file contents between tape and disk implicitly change the ctime while recalling the file on disk. On the other hand, the backing store checks it is same between before recall and after recall. So get operation against migrated file is stalled every time. 

Do not check `ctime` at read (recall) in the Tier-2 environment.

Instead we can have `mtime` check, our filesystem handles `mtime` correctly.

---

@guymguym , 

I created this based on the conversation in 5/29. Please handle this.

tangledbytes:
### Explain the changes
This PR adds a simple flag in `fs_napi` which forces using `mtime` instead of `ctime` wherever appropriate. At some places use of mtime is not appropriate so in that place the check is completely skipped.

### Fixes: #7314

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -105,6 +105,21 @@
         }                                                         \
     } while (0)
 
+// This macro is similar to `CHECK_CTIME_CHANGE` but it is used when we want to
+// instead use mtime instead of ctime. Using mtime instead of ctime is useful
+// when the inode data may change but should not interfere with the operations
+// NooBaa is doing.
+#define CHECK_MTIME_CHANGE(fd, stat_res, path)                    \
+    do {                                                          \
+        auto start_mtime = stat_res.st_mtime;                     \
+        SYSCALL_OR_RETURN(fstat(fd, &stat_res));                  \
+        auto end_mtime = stat_res.st_mtime;                       \
+        if (start_mtime != end_mtime) {                           \
+            SetError(XSTR() << "FileStat: " << DVAL(path)         \
+                            << " cancelled due to mtime change"); \
+        }                                                         \
+    } while (0)
+
 #ifdef __APPLE__
 #define flistxattr(a, b, c) ::flistxattr(a, b, c, 0)
 #define getxattr(a, b, c, d) ::getxattr(a, b, c, d, 0, 0)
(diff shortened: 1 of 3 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -619,7 +641,11 @@ struct Stat : public FSWorker
                 GPFS_FCNTL_OR_RETURN(get_fd_gpfs_xattr(fd, _xattr, gpfs_error));
             }
         }
-        CHECK_CTIME_CHANGE(fd, _stat_res, _path);
+
+        // If mtime check is requested instead of ctime check then completely
+        // skip running the check. Change in mtime does not indicate anything other
+        // than that file content.
+        if (!use_mtime_check) CHECK_CTIME_CHANGE(fd, _stat_res, _path);
```
guymguym, line 648: TBH this is rather confusing, because we have an option that explicitly say "use mtime" and then we ignore mtime in some cases, so perhaps we should just rename the option to represent the behavior better?

guymguym (commented):
@utkarsh-pro See comments, I don't have a concrete suggestion for renaming the use_mtime_check option, but I think it's a bit misleading as it is right now...

src/native/fs/fs_napi.cpp
```diff
@@ -1034,7 +1058,9 @@ struct Readfile : public FSWorker
             p += len;
         }
 
-        CHECK_CTIME_CHANGE(fd, _stat_res, _path);
+        if (!disable_ctime_check) {
+            CHECK_MTIME_CHANGE(fd, _stat_res, _path);
+        }
```
guymguym, line 1063: Need to fix this to:
```suggestion
        if (disable_ctime_check) {
            CHECK_MTIME_CHANGE(fd, _stat_res, _path);
        } else {
            CHECK_CTIME_CHANGE(fd, _stat_res, _path);
        }
```

utkarsh-pro: Thank you.

src/native/fs/fs_napi.cpp
```diff
@@ -459,6 +474,11 @@ struct FSWorker : public Napi::AsyncWorker
     double _took_time;
     Napi::FunctionReference _report_fs_stats;
 
+    // diable_ctime_check disables the ctime check in the stat and read file fuctions
+    //
+    // NOTE: If this is enabled then some functions will fallback to using mtime check
+    bool disable_ctime_check;
```
guymguym, line 480: Oops, we forgot to initialize this in ctor to false...

guymguym (approved):
!!

PR #7325, opened 2023-06-06
Bug Fix #7324 | Change Error Message of Put Bucket Versioning

shirady:
### Explain the changes
1. Change the error message of Put bucket versioning in case the versioning configuration status is not Enabled or Suspended.

### Issues: Fixed #7324 
1. Put bucket versioning error using NooBaa endpoint should be same as AWS (which is MalformedXML).

### Testing Instructions:
1. Are inside the issue mentioned above (#7324)

- [ ] Doc added/updated
- [ ] Tests added


shirady:
After changing the error message:
Screenshot 2023-06-06 at 15 23 09

nimrod-becker (approved):

PR #7329, opened 2023-06-07
Bug Fix #7328 | Remove Duplications before Sending Them to Delete Multiple Objects

### LINKED ISSUE #7328: Delete Objects Reply of Same Key Objects Using Noobaa Endpoint Is Different than AWS
### Environment info
- NooBaa Version: current master (5.14)
- Platform: minikube 1.30.1

### Actual behavior
1. Delete objects with the same key objects returns a couple of DeleteObjects in reply (although it is the same object to delete).

### Expected behavior
1.  Delete objects with the same delete request should return one DeleteObject.

#### AWS
You should have an account with an access key and secret access key (use `aws configure` with those details), and a bucket.
In the bucket please add key hello.txt, then run:
`aws s3api delete-objects --bucket <your-bucket-name> --delete file://delete.json`
`delete.json` is a JSON document in the current directory that specifies the object to delete:
```
{
    "Objects": [
      {
        "Key": "hello.txt"
      },
      {
        "Key": "hello.txt"
      },
      {
        "Key": "hello.txt"
      }
    ]
}
```
#### NooBaa
You should deploy and install NooBaa, (use `aws configure` with admin access key and secret key from the `status` command in Operator CLI), add key hello.txt to the `first.bucket`, then run:
`aws s3api delete-objects -endpoint-url=https://localhost:12443 --no-verify-ssl --bucket <your-bucket-name> --delete file://delete.json`
`delete.json` is a JSON document in the current directory that specifies the object to delete:
```
{
    "Objects": [
      {
        "Key": "hello.txt"
      },
      {
        "Key": "hello.txt"
      },
      {
        "Key": "hello.txt"
      }
    ]
}
```
Notes:
- In the example above the NooBaa bucket is the `first.bucket` that we have by default.
- The `endpoint-url` I'm using is with port forwarding (`kubectl port-forward -n default service/s3 12443:443`). 

#### AWS
In the screenshot below we see the commands that put the object hello.txt, show the `delete.json` content, and run delete objects.
Screenshot 2023-06-07 at 9 11 35

AWS reply:
Screenshot 2023-06-07 at 9 11 22

#### NooBaa
In the screenshot below we see the commands that put the object hello.txt, show the `delete.json` content, and run delete objects.
Screenshot 2023-06-07 at 9 33 03

NooBaa reply:
Screenshot 2023-06-07 at 9 29 31
Three DeleteObjects instead of one.

shirady:
### Explain the changes
1. Remove duplications before sending them to delete multiple objects.
2. Add parsing XML errors in case required parts the XML were not sent by the client.

### Issues: Fixed #7328 
1. Delete-objects reply of the same key objects should be the same as AWS.

2. If the client didn't send the XML with the `Delete` part we would have an error that we cannot read properties of undefined (tried to read 'Quiet'). _Note: This was not part of the original issue but was reviled in the discussion of the PR._

### Testing Instructions:
1. Are inside the issue mentioned above #7328 .

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_post_bucket_delete.js
+++ b/src/endpoint/s3/ops/s3_post_bucket_delete.js
@@ -1,7 +1,6 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
-const _ = require('lodash');
 const dbg = require('../../../util/debug_module')(__filename);
 const S3Error = require('../s3_errors').S3Error;
 
@@ -11,18 +10,31 @@ const S3Error = require('../s3_errors').S3Error;
  */
 async function post_bucket_delete(req) {
 
-    const quiet = req.body.Delete.Quiet && req.body.Delete.Quiet[0];
-    const objects = _.map(req.body.Delete.Object, obj => ({
-        key: obj.Key && obj.Key[0],
-        version_id: obj.VersionId && obj.VersionId[0],
-    }));
+    // The request body XML format:
+    // Delete (Required), Object (Required), Quiet (Not Required)
+    // Parsing the XML and throw an error in case the required part are not included
+    const delete_data = req.body.Delete;
+    if (!delete_data) throw new S3Error(S3Error.MalformedXML);
+    const delete_list = req.body.Delete.Object;
+    if (!delete_list) throw new S3Error(S3Error.MalformedXML);
+    const quiet = req.body.Delete.Quiet?.[0];
 
-    if (objects.length > 1000) {
+    if (delete_list.length > 1000) {
         dbg.error('The request can not contain a list of more than 1000 keys');
         throw new S3Error(S3Error.MalformedXML);
     }
 
-    dbg.log3('post_bucket_delete: objects', objects);
+    // Removing duplicated entries
+    // note: it was intentionally that first we check the list length and only then remove duplications
+    const uniq_map = new Map();
+    for (const item of delete_list) {
+        const key = item.Key?.[0];
+        const version_id = item.VersionId?.[0];
+        const key_version = (key || '') + '\0' + (version_id || ''); // using null char (\x00) as separator
+        uniq_map.set(key_version, { key, version_id });
+    }
+    const objects = Array.from(uniq_map.values());
+    dbg.log3('post_bucket_delete: objects without duplications', objects);
 
     const reply = await req.object_sdk.delete_multiple_objects({
         bucket: req.params.bucket,
--- a/src/test/unit_tests/test_bucketspace_versioning.js
+++ b/src/test/unit_tests/test_bucketspace_versioning.js
@@ -1151,13 +1151,13 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
             }
             const delete_res = await account_with_access.deleteObjects({
                     Bucket: delete_multi_object_test_bucket, Delete: { Objects: arr } }).promise();
-            assert.equal(delete_res.Deleted.length, 200);
+            assert.equal(delete_res.Deleted.length, 1);
             for (const res of delete_res.Deleted) {
                 assert.equal(res.DeleteMarker, true);
             }
             const versions_dir = path.join(full_multi_delete_path, '.versions');
             const versions = await nb_native().fs.readdir(DEFAULT_FS_CONFIG, versions_dir);
-            assert.equal(versions.length, 501);
+            assert.equal(versions.length, 302); // 1 null version + 30 dm and versions + 1 new dm
             await delete_object_versions(full_multi_delete_path, key1);
             await delete_object_versions(full_multi_delete_path, 'a');
         });
@@ -1306,19 +1306,19 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
 
             const delete_res = await account_with_access.deleteObjects({
                     Bucket: delete_multi_object_test_bucket, Delete: { Objects: arr } }).promise();
-            assert.equal(delete_res.Deleted.length, 150);
-            for (let i = 0; i < 50; i++) {
+            assert.equal(delete_res.Deleted.length, 101); // 100 dm and versions + 1 new dm
+            for (let i = 0; i < 1; i++) {
                 assert.notEqual(delete_res.Deleted[i].DeleteMarkerVersionId, undefined);
                 assert.equal(delete_res.Deleted[i].DeleteMarker, true);
             }
-            for (let i = 50; i < 150; i++) {
-                if (i % 2 === 1) assert.equal(delete_res.Deleted[i].DeleteMarker, true);
-                if (i % 2 === 0) assert.equal(delete_res.Deleted[i].DeleteMarker, undefined);
+            for (let i = 1; i < 101; i++) {
+                if (i % 2 === 0) assert.equal(delete_res.Deleted[i].DeleteMarker, true);
+                if (i % 2 === 1) assert.equal(delete_res.Deleted[i].DeleteMarker, undefined);
             }
             const versions_dir = path.join(full_multi_delete_path, '.versions');
             const versions = await nb_native().fs.readdir(DEFAULT_FS_CONFIG, versions_dir);
 
-            assert.equal(versions.length, 250);
+            assert.equal(versions.length, 201); // 200 dm and versions + 1 new dm
             await fs_utils.file_must_not_exist(path.join(full_multi_delete_path, key1));
             await delete_object_versions(full_multi_delete_path, key1);
         });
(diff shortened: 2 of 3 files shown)


aayushchouhan09 (changes requested):

src/endpoint/s3/ops/s3_post_bucket_delete.js
```diff
@@ -62,13 +67,14 @@ async function post_bucket_delete(req) {
...
         } else {
-            results[i] = {
+            results[j] = {
                 Deleted: {
                     Key: req_obj.key,
                     VersionId: req_obj.version_id,
                 }
             };
         }
+        j += j; // since we don't increase the counter of j in the for-loop header
```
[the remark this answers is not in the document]
shirady: Sorry, I put this PR as a draft without reviewers. It was a typo, you're right.

aayushchouhan09 (approved):

jackyalbo (commented):
Wouldn't it be better to remove duplicates from the request? before sending to the core? and then avoid it for the results?

shirady:
> Wouldn't it be better to remove duplicates from the request? before sending to the core? and then avoid it for the results?

Thank you, I changed it.
As far as I know, it would cause failed tests of `src/test/unit_tests/test_bucketspace_versioning.js` since there are tests that count the keys in the result.

src/endpoint/s3/ops/s3_post_bucket_delete.js
```diff
@@ -12,7 +12,7 @@ const S3Error = require('../s3_errors').S3Error;
 async function post_bucket_delete(req) {
 
     const quiet = req.body.Delete.Quiet && req.body.Delete.Quiet[0];
-    const objects = _.map(req.body.Delete.Object, obj => ({
+    let objects = _.map(req.body.Delete.Object, obj => ({
```
romayalon, line 15: I would first filter and then map

src/endpoint/s3/ops/s3_post_bucket_delete.js
```diff
@@ -23,6 +23,8 @@ async function post_bucket_delete(req) {
     }
 
     dbg.log3('post_bucket_delete: objects', objects);
+    objects = _.uniqWith(objects, _.isEqual);
```
romayalon, line 26: I'm not sure about this uniqWith()... 
googled it a bit and saw that it has bad performance, how much time it takes for 1000 objects?
I would use filter() & a helper Set() but we need to check what's faster

src/endpoint/s3/ops/s3_post_bucket_delete.js
```diff
@@ -23,6 +23,8 @@ async function post_bucket_delete(req) {
...
+    dbg.log3('post_bucket_delete: objects without duplications', objects);
 
     const reply = await req.object_sdk.delete_multiple_objects({
```
romayalon, line 29: please add a unit test

src/endpoint/s3/ops/s3_post_bucket_delete.js
```diff
@@ -23,6 +23,8 @@ async function post_bucket_delete(req) {
     }
```
romayalon, line 24: Notice that on AWS, will we get this error on the non-filtered array, I checked it on an array that contains only 2 different keys, and all the rest are duplicates
so I'm not sure you will not be able to filter first

src/endpoint/s3/ops/s3_post_bucket_delete.js
shirady: 1. The helper Set() would not preserve the order of the keys.
2. I did a simple check:

Screenshot 2023-06-13 at 8 50 12

Call delete objects with the same key that appears 1,000 times in the list.
It took less than 0.005 seconds (with printings included).
Here is a screenshot from the logs:
Screenshot 2023-06-13 at 8 46 59

src/endpoint/s3/ops/s3_post_bucket_delete.js
shirady: I agree with you.

src/endpoint/s3/ops/s3_post_bucket_delete.js
shirady: You're right in the next comment.
I cannot filter the array before since the limit of 1,000 keys is on the original objects.
I also checked it before I decided where to filter the array.

src/endpoint/s3/ops/s3_post_bucket_delete.js
shirady: Sure

src/endpoint/s3/ops/s3_post_bucket_delete.js
romayalon: why do you need to preserve the order of the keys?
I meant that it'll be a lookup Set for the duplicate check.
see the accepted answer here - https://stackoverflow.com/questions/49507420/lodash-performance-hit-with-uniqwith-to-remove-duplicate-objects

shirady: You can read about `uniqWith()` [here](https://stackoverflow.com/questions/31740155/lodash-remove-duplicates-from-array) if you want (more users mark it as a useful answer).
I improved the performance check according to the link you send by using `performance.now()`, and you can see:
Screenshot 2023-06-18 at 14 11 54

It's about ~8 ms
Screenshot 2023-06-18 at 11 14 47

shirady: `Set` is using the `===` operator (see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set#value_equality)), hence we cannot compare the whole objects (since `===` comparison on objects checks reference comparison) and we would need another tool to check the two properties of `key` and `version_id` to decide if an object is duplicated.

romayalon: try the following method:
```
// method object 
const obj_start = performance.now();
const obj = {};
const obj_result = [];
for (let i = 0; i < objects.length; i++) {
  const key_version = JSON.stringify(objects[i]);
  if (!obj[key_version]) {
    obj[key_version] = true;
    obj_result.push(key_version);
  }
}
const obj_end = performance.now();
```
I see a major difference when running on my laptop between the 3 methods: set, object and uniq.
SET:
Removed 6 duplicates from 1000 items
Operation took 0.7413411140441895 ms
OBJECT:
Removed 6 duplicates from 1000 items
Operation took 1.0831308364868164 ms
UNIQUE:
Removed 6 duplicates from 1000 items
Operation took 378.04422998428345 ms

romayalon: if you will push objects[i]  to obj_result will it work without parsing it again?

shirady: Yes

romayalon (approved):

src/endpoint/s3/ops/s3_post_bucket_delete.js
```diff
@@ -23,6 +24,23 @@ async function post_bucket_delete(req) {
...
 
     dbg.log3('post_bucket_delete: objects', objects);
+    // In this section we will remove duplication from object array
```
guymguym, line 27: @shirady Let me suggest a variations of the code, to eliminate the JSON.stringify call which is much heavier than a simple string concat. Using map/set is also a better idea in cases where the uniq keys are long and random.

```js
const delete_list = req.body.Delete?.Object || [];
if (delete_list.length > 1000) {
    throw ...
}

const uniq_map = new Map();
    const key = item.Key?.[0];
    const version_id = item.VersionId?.[0];
    const key_version = (key || '') + '\0' + (version_id || ''); // using null char (\x00) as separator
    uniq_map.set(key_version, { key, version_id });
}
const objects = Array.from(uniq_map.values());
```

shirady: I changed it by your suggestion, although I didn't understand why you use `req.body.Delete?.Object` and not just `req.body.Delete.Object` since Object is required:

> Object
> The object to delete.
> Type: Array of ObjectIdentifier data types
> Required: Yes

Would you please explain why?
To match the style (and avoid a DeepScan issue) I added the optional chaining operator (`?.`) to `quiet`.

liranmauda: `req.body.Delete?.Object` checks if `req.body.Delete` is defined, and only if it is, it goes to `Object`

shirady: Thanks, but `Delete` is also required as well...
```
Delete
Root level tag for the Delete parameters.
Required: Yes
```
And before this change, we didn't check if it is non-null or non-defined and was fine... (or not?).

guymguym: because the user might not be so dependable and might not send it in the xml...

shirady: Before the change (in addition to the bug fix) we also didn't parse the XML, so in case the client didn't send the XML with the `Delete` part we would have an error that we cannot read properties of undefined (tried to read 'Quiet').

As discussed, there are a few operations in which the body we get an XML is low (I checked - 22 operations in `s3/ops` we implemented approximately half of them).

There are two options to handle the missing `Delete` part:
1. Ignore it and use an empty array of the deleted list (the suggested change in first commit).
2. Throw an error when we start to parse the XML and don't find a required field.
I was convinced when we talked about the first option, but now that I'm looking at our examples I tend to go by the second option, for example: `put_bucket_tagging(req)` uses a function `parse_body_tagging_xml(req)` and inside it the first check is:
     
```
const tagging = req.body.Tagging;
    if (!tagging) throw new S3Error(S3Error.MalformedXML);
```
* I added the second option in a separate commit.

guymguym (approved):

PR #7338, opened 2023-06-12
fix block recall

tangledbytes:
### Explain the changes
This PR:
1. Adds an option which prevents block store from throwing `MIGRATED` error. Enabled by default. This is a temporary fix.
2. Changes standalone.md to recommend increasing IO_TIMEOUTS (used in blockstore _request calls).

- [ ] Doc added/updated
- [ ] Tests added


guymguym (approved):
Looks great!

PR #7339, opened 2023-06-13
Validate versions_dir_cache

### LINKED ISSUE #7334: Improve versions_dir_cache for NSFS
Right now `versions_dir_cache` is caching the entries only for objects and versions of objects `validity()` is not verfied and not served from the cache. This ticket is too handle the `validate()` part of `versions_dir_cache`

Description of what you want to happen.

vh05:
### Explain the changes
With this patch, versions_dir_cache can be invalidated based object modifications under HIDDEN_VERSION_PATH and cache would be refreshed

### Issues: Fixed #7334 

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -267,10 +267,11 @@ const dir_cache = new LRUCache({
  * @typedef {{
  *  time: number,
  *  stat: nb.NativeFSStats,
+ *  ver_dir_stat: nb.NativeFSStats,
  *  usage: number,
  *  sorted_entries?: fs.Dirent[],
  * }} ReaddirVersionsCacheItem
- * @type {LRUCache<object, string, ReaddirCacheItem>}
+ * @type {LRUCache<object, string, ReaddirVersionsCacheItem>}
  */
 const versions_dir_cache = new LRUCache({
     name: 'nsfs-versions-dir-cache',
@@ -281,8 +282,9 @@ const versions_dir_cache = new LRUCache({
         const version_path = dir_path + "/" + HIDDEN_VERSIONS_PATH;
         let ver_dir_stat_size;
         let is_version_path_exists = false;
+        let ver_dir_stat;
         try {
-            const ver_dir_stat = await nb_native().fs.stat(fs_context, version_path);
+            ver_dir_stat = await nb_native().fs.stat(fs_context, version_path);
             ver_dir_stat_size = ver_dir_stat.size;
             is_version_path_exists = true;
         } catch (err) {
@@ -291,6 +293,7 @@ const versions_dir_cache = new LRUCache({
             } else {
                 throw err;
             }
+            ver_dir_stat = null;
             ver_dir_stat_size = 0;
         }
         let sorted_entries;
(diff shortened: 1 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -293,11 +296,21 @@ const versions_dir_cache = new LRUCache({
...
         const new_stat = await nb_native().fs.stat(fs_context, dir_path);
-        return (new_stat.ino === stat.ino && new_stat.mtimeNsBigint === stat.mtimeNsBigint);
+        if (ver_dir_stat) {
+            const versions_dir_path = path.normalize(path.join(dir_path, '/', HIDDEN_VERSIONS_PATH));
+            const new_versions_stat = await nb_native().fs.stat(fs_context, versions_dir_path);
+            return (new_stat.ino === stat.ino &&
+                    new_stat.mtimeNsBigint === stat.mtimeNsBigint &&
+                    new_versions_stat.ino === ver_dir_stat.ino &&
+                    new_versions_stat.mtimeNsBigint === ver_dir_stat.mtimeNsBigint);
+            } else {
```
romayalon, line 310: indentation

src/sdk/namespace_fs.js
```diff
@@ -293,11 +296,21 @@ const versions_dir_cache = new LRUCache({
...
+                return (new_stat.ino === stat.ino &&
+                        new_stat.mtimeNsBigint === stat.mtimeNsBigint);
+            }
```
romayalon, line 313: worth adding some tests

src/sdk/namespace_fs.js
v-harihar: My bad

src/sdk/namespace_fs.js
```diff
@@ -293,11 +296,21 @@ const versions_dir_cache = new LRUCache({
...
             }
         }
-        return { time, stat, sorted_entries, usage };
+        return { time, stat, ver_dir_stat, sorted_entries, usage };
     },
-    validate: async ({ stat }, { dir_path, fs_context }) => {
+    validate: async ({ stat, ver_dir_stat }, { dir_path, fs_context }) => {
         const new_stat = await nb_native().fs.stat(fs_context, dir_path);
-        return (new_stat.ino === stat.ino && new_stat.mtimeNsBigint === stat.mtimeNsBigint);
+        if (ver_dir_stat) {
```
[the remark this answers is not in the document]
v-harihar: Let me present my thoughts here.

Base of the thought: In any case, when we update the object, will will end up calling `load()` which will eventually load objects and its versions.

case 1: No versions exist, only latest object CREATED: Else part will be executed (because `ver_dir_stat` can't be populated from the cache) and it will be false, so load() will be called 

case 2: version of the object CREATED, So object created in case 1 will be moved to `.versions`: Else part will be executed (because `ver_dir_stat` can't be populated from the cache) and it will return false since the objects have been modified under main dir and returns false and lead to call to `load()`

case 3: 2nd version of the object CREATED: In this `mtime` is changed on both main dir and `.versions` dir, we will end up calling if() in this case and returns false since objects have been modified and lead to call to `load()`

case 4: If we delete any version of the object under `.versions`,  then we call list_versions(), then if part will be executed because `ver_dir_stat` will be already there in the cache and returns false since the dir has been modified

`if(ver_dir_stat)` condition addresses only those cases (like case 4) in which `mtime` modification only happens to `.versions` and not  to main directory. Whenever we PUT the object both main dir and `.versions` dir's `mtime` will be modified because we make the movement of the objects between main dir and `.versions` dir

Please correct me if I am missing anything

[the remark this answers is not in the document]
romayalon: hmm got you ok

src/sdk/namespace_fs.js
v-harihar: sure

v-harihar: I will add the test and update the patch asap. Some problems with the working system

romayalon (approved):

PR #7340, opened 2023-06-13
Fix backingstore missing RPC errors

tangledbytes:
### Explain the changes
This PR attempts to fix the errors we see while running AWS commands against endpoints in standalone (non container-zed deployment of NooBaa) by ensuring that the system_store is created with `standalone: true` attribute which prevents registering the agent as cluster member hence `publish_to_cluster` call doesn't reaches out to the agent (which does not have the RPC registered).

NOTE: `require('......auth_server')` delay is intentional. `auth_server` invokes `system_store.get_instance()` as well which defeats the purpose of later on passing `standalone: true`.

- [ ] Doc added/updated
- [ ] Tests added


utkarsh-pro:
Duplicate PR. @guymguym has a better fix.

PR #7343, opened 2023-06-15
nsfs support for direct io

guymguym:
### Explain the changes
1. direct io is useful in some cases to reduce data copies.
2. nsfs buffer pool allocates its buffers with memalign of 4K to allow direct io from the FS.
3. refactor parse_open_flags instead of the flags_to_case map, to allow direct/sync options no matter in which order.

1. NA

### Testing Instructions:
1. override in config-local to set `config.NSFS_OPEN_READ_FLAGS = 'rd'` or in .env with `CONFIG_JS_NSFS_OPEN_READ_FLAGS=rd`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -7,6 +7,7 @@
 #include "./gpfs_fcntl.h"
 
 #include <dirent.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <limits.h>
 #include <map>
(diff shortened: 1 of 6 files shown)


romayalon (approved):

src/native/fs/fs_napi.cpp
```diff
@@ -1834,6 +1846,29 @@ set_debug_level(const Napi::CallbackInfo& info)
     return info.Env().Undefined();
 }
 
+static const int DIO_BUFFER_MEMALIGN = 4096;
```
utkarsh-pro, line 1849: Maybe if and when we want to support a wider range of systems we can add a C dep which queries the block devices for their block size?

Older disks (like really old disks) usually have 512 block size (while the modern 4kn ones have usually a 512 emulation). Sometimes I have noticed that using USB for drives gives them block size 512 (logical block size while physical stays at some 4kn). I have also sometimes rarely encountered 1024 as block size.

Fun fact: I discovered recently that 9p mounts have absolutely no issue with even 8 byte alignment (found via experiments).

guymguym: @utkarsh-pro but 4k align seems to cover all known filesystems, right? 

I also asked if this should be used from fsstat or stat and was told that the best is to use the stat.blksize because fsstat might have different storage beneath. However we want to allocate these buffers once and pool them, so I just used the maximum, and I dont think this will cause any memory waste.

PR #7345, opened 2023-06-15
Fixing test_s3_bucket_policy to work with pg sequences

jackyalbo:
### Explain the changes
1. we used the same key for all the tests - making the versioning test case to delete a shared key KEY, which other TC depend on
2. it still worked as in the versioning TC I was using 'nbver-1' which doesn't exist since we moved to PG sequences - so KEY didn't really get deleted although he was supposed to.
4. now versioning TC will use a new key and deleting it won't affect the other TCs
5. taking down the test setup to one pool with one host

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


liranmauda (commented):
LGTM,

Just one comment, why not get the first version into a variable also, then use the exact VersionID to delete each object? 

I feel it is more straight forward, general,  and does not rely on the fact that it is an actual sequence

romayalon (approved):

PR #7347, opened 2023-06-16
Makefile add optional CONTAINER_PLATFORM flag to container builds

guymguym:
### Explain the changes
1. See https://lima-vm.io/?file=docs/multi-arch.md
2. See https://github.com/containerd/nerdctl/blob/main/docs/multi-platform.md
3. used `CONTAINER_PLATFORM=amd64` for building x86_64 images on Mac arm64.
4. In a lima shell:
```
make noobaa CONTAINER_ENGINE=nerdctl CONTAINER_PLATFORM=amd64
```

CC @utkarsh-pro 

1. NA

### Testing Instructions:
1. NA

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,14 @@ ifeq ($(CONTAINER_ENGINE),)
 	CONTAINER_ENGINE=$(shell podman version >/dev/null 2>&1 && echo podman)
 endif
 
+# see https://github.com/containerd/nerdctl/blob/main/docs/multi-platform.md
+# e.g use CONTAINER_PLATFORM=amd64 for building x86_64 on arm.
+CONTAINER_PLATFORM_FLAG=
+ifdef CONTAINER_PLATFORM
+	CONTAINER_PLATFORM_FLAG="--platform=$(CONTAINER_PLATFORM)"
+endif
+
+
 GIT_COMMIT?="$(shell git rev-parse HEAD | head -c 7)"
 NAME_POSTFIX?="$(shell ${CONTAINER_ENGINE} ps -a | wc -l | xargs)"
 BUILDER_TAG?="noobaa-builder"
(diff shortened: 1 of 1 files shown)


liranmauda (approved):
LGTM

Makefile
```diff
@@ -3,6 +3,14 @@ ifeq ($(CONTAINER_ENGINE),)
 	CONTAINER_ENGINE=$(shell podman version >/dev/null 2>&1 && echo podman)
 endif
 
+# see https://github.com/containerd/nerdctl/blob/main/docs/multi-platform.md
+# e.g use CONTAINER_PLATFORM=amd64 for building x86_64 on arm.
```
utkarsh-pro, line 7:
```suggestion
# e.g use CONTAINER_PLATFORM_FLAG=amd64 for building x86_64 on arm.
```

tangledbytes (approved):

Makefile
guymguym: no, notice that I had to change it because it failed otherwise. The user needs to set the CONTAINER_PLATFORM=amd64 and then the Makefile will sets the CONTAINER_PLATFORM_FLAG=--platform=amd64 (approppriately) for all the container build calls...

utkarsh-pro:
@guymguym, docker supports multi platform builds as well. 
```
$ docker run --platform linux/amd64 busybox uname -a
Linux a7476d751565 5.15.82-0-virt #1-Alpine SMP Mon, 12 Dec 2022 09:15:17 +0000 x86_64 GNU/Linux

$ docker build --platform linux/arm64,linux/amd64,linux/ppc64le ...` # This requires a custom docker builder  though - docker buildx create ...
```

guymguym:
> docker supports multi platform builds as well.

@utkarsh-pro even better then! I just tried with lima and nerdctl, but I assumed this flag could be applicable for other container engines as well.

liranmauda:
@guymguym @utkarsh-pro 
we need to see that podman support that too.
Currently, we all should move to podman.

guymguym:
@liranmauda it was very simple for me to check that podman has the exact same --platform flag too. However I am not trying to make any suggestions about running this downstream. But if you want to try, now you can with a simple arg.

liranmauda:
> @liranmauda it was very simple for me to check that podman has the exact same --platform flag too. However I am not trying to make any suggestions about running this downstream. But if you want to try, now you can with a simple arg.

I remember that at some point we could not do it as podman was behind. 
In any case, those flows are not triggered downstream, so we are safe there. (we are not using our makefile there).

PR #7355, opened 2023-06-20
NSFS | Check for GPFS lib path env var before calling gpfs functions

romayalon:
### Explain the changes
**The Problem -** Creating nsfs namespacestore using fs-backend=GPFS without having GPFS lib caused the endpoint pods to get into CLBO.

**The Solution -**
1. Before reading GPFS xattr - instead of only checking the Namespacestore backend type, we will also check for the existance of the GPFS_DL_PATH environment variable.

1. 

### Testing Instructions:
1. Install noobaa
2. Create an NSFS Namespacestore of type GPFS
3. Check if the endpoint pods crashing

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -135,6 +135,8 @@ DBG_INIT(0);
 
 typedef std::map<std::string, std::string> XattrMap;
 
+static const char* gpfs_dl_path = std::getenv("GPFS_DL_PATH");
+
 static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg) = 0;
 
 static int (*dlsym_gpfs_linkat)(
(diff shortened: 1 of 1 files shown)


jackyalbo (approved):

src/native/fs/fs_napi.cpp
```diff
@@ -559,6 +559,11 @@ struct FSWorker : public Napi::AsyncWorker
             _report_fs_stats.Call({ fs_worker_stats });
         }
     }
+    bool use_gpfs_lib()
+    {
+        const char* gpfs_dl_path = std::getenv("GPFS_DL_PATH");
```
guymguym, line 564: we have this check in the initialization function on the bottom of the file, so how about moving the gpfs_dl_path variable to a static module like the dlsym_gpfs_fcntl and other symbols we load from the library? -

https://github.com/noobaa/noobaa-core/blob/590ce634e5cf0f3c9066e5064202c90d1bb82e5c/src/native/fs/fs_napi.cpp#LL1881C60-L1881C60

another option is to just check if one of those symbols is non zero...

romayalon: I moved it to be a static field, is that fine?

guymguym (approved):

PR #7357, opened 2023-06-21
fs_napi readFile fix buffer mem leak (introduced recently in pr #7283)

guymguym:
### Explain the changes
1. passing a buffer_releaser to the buffer allocated in readFile.

1. bug introduced in #7283

### Testing Instructions:
1. invoke backing store fs reads.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -146,6 +146,14 @@ static int (*dlsym_gpfs_linkatif)(
 static int (*dlsym_gpfs_unlinkat)(
     gpfs_file_t fileDesc, const char* path, gpfs_file_t fd) = 0;
 
+static const int DIO_BUFFER_MEMALIGN = 4096;
+
+static void
+buffer_releaser(Napi::Env env, uint8_t* buf)
+{
+    if (buf) free(buf);
+}
+
 static int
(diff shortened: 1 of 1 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -1058,7 +1064,11 @@ struct Readfile : public FSWorker
         }
 
         _len = _stat_res.st_size;
-        _data = new uint8_t[_len];
+        int r = posix_memalign((void**)&_data, DIO_BUFFER_MEMALIGN, _len);
+        if (r || !_data) {
```
utkarsh-pro, line 1068: I don't know if we need the ability to read 0 sized file or not but we do then I think we would error out here.

Quoting `posix_memalign` man page:
> If size is 0, then posix_memalign() returns either NULL, or a unique pointer value that can later be successfully passed to [free](https://linux.die.net/man/3/free)(3).

guymguym: oh, so I guess I should only check the return code, and not the buffer pointer, is that what you meant?

guymguym: @utkarsh-pro see updated version

tangledbytes (approved):
LGTM :)

PR #7359, opened 2023-06-22
Fix fs_napi after #7358 - use ToBoolean to check for non falsy values

guymguym:
### Explain the changes
1. In #7358 I introduced a bug that fails writes because in fs_napi we used to check if a key was provided in the options, but not if that key has a non-falsy value, meaning not undefined or null etc.

1. NA

### Testing Instructions:
1. Flows that use fs_napi writeFile such as backing_store_fs.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -474,7 +474,7 @@ clear_xattr(int fd, std::string _prefix)
 static void
 load_xattr_get_keys(Napi::Object& options, std::vector<std::string>& _xattr_get_keys)
 {
-    if (options.Has("xattr_get_keys")) {
+    if (options.Get("xattr_get_keys").ToBoolean()) {
         Napi::Array keys = options.Get("xattr_get_keys").As<Napi::Array>();
         auto keys_length = keys.Length();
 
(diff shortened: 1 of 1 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -654,9 +661,9 @@ struct Stat : public FSWorker
         : FSWorker(info)
     {
         _path = info[1].As<Napi::String>();
-        if ((int)info.Length() == 3) {
+        if (info[2].ToBoolean()) {
```
utkarsh-pro, line 664: We had the check for number of arguments passed earlier but here we are missing it. I might be wrong but I think here if the function is invoked with incorrect number of arguments then that might throw error or worse (don't know if runtime bound checks happens when node gyp compiles) we might read invalid memory?

👆 assumes that we want to fail nicely if the function is invoked with incorrect number of arguments.

guymguym: Right, but the point is that it acts just like an array in javascript, when you access an index above length, it just returns undefined. So there is no real need to check for length specifically, but just that it evaluates to a non-falsy value.

tangledbytes (approved):

src/native/fs/fs_napi.cpp
utkarsh-pro: Ah, that's neat.

PR #7360, opened 2023-06-22
[TMFS] Address minors from #7313 & support auto second tier creation & assignment if configured

tangledbytes:
### Explain the changes

This PR:
1. Adds support for auto tier 2 creation and assignment to backingstore buckets if `TMFS_BUCKET_TIER2_AUTOCONF` and `BLOCK_STORE_FS_TMFS_ENABLED` are enabled. Created another config parameter instead of linking it to `BLOCK_STORE_FS_TMFS_ENABLED` to provide slightly more control over behaviour hopefully helpful while debugging.
2. Addresses minors from #7313 by removing the js_utils function and renaming the functions.

PS: 2nd change was introduced accidentally as a result of creating a branch off my previous branch (which had the changes). I can split the PR into 2 different PRs or merge the 2 commits, I have no preference here.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -176,6 +176,7 @@ config.DEFAULT_BUCKET_NAME = 'first.bucket';
 config.INTERNAL_STORAGE_POOL_NAME = 'system-internal-storage-pool';
 // config.SPILLOVER_TIER_NAME = 'bucket-spillover-tier';
 config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
+config.BUCKET_AUTOCONF_TIER2_ENABLED = false;
 
 //////////////////////////
 // MD AGGREGATOR CONFIG //
--- a/src/agent/block_store_services/block_store_fs.js
+++ b/src/agent/block_store_services/block_store_fs.js
@@ -15,6 +15,7 @@ const string_utils = require('../../util/string_utils');
 const BlockStoreBase = require('./block_store_base').BlockStoreBase;
 const get_block_internal_dir = require('./block_store_base').get_block_internal_dir;
 const { RpcError } = require('../../rpc');
+const { STORAGE_CLASS_GLACIER, STORAGE_CLASS_STANDARD } = require('../../endpoint/s3/s3_utils');
 
 const TMFS_STATE_MIGRATED = 'MIGRATED';
 const TMFS_STATE_PREMIGRATED = 'PREMIGRATED';
@@ -271,10 +272,12 @@ class BlockStoreFs extends BlockStoreBase {
      * @returns {Promise<{ moved_block_ids: string[] }>}
      */
     async _move_blocks_to_storage_class(block_ids, storage_class) {
-        if (storage_class === 'GLACIER') {
-            return this._move_blocks_to_glacier(block_ids);
+        if (storage_class === STORAGE_CLASS_GLACIER) {
+            if (config.BLOCK_STORE_FS_TMFS_ENABLED) {
+                return this._move_blocks_to_tmfs(block_ids);
+            }
         }
-        if (storage_class === 'STANDARD') {
+        if (storage_class === STORAGE_CLASS_STANDARD) {
             // Nothing to do for now
             return Promise.resolve({ moved_block_ids: [] });
         }
@@ -286,29 +289,18 @@ class BlockStoreFs extends BlockStoreBase {
      * @param {string[]} block_ids
      * @returns {Promise<{ moved_block_ids: string[] }>}
      */
-    async _move_blocks_to_glacier(block_ids) {
+    async _move_blocks_to_tmfs(block_ids) {
         const moved_block_ids = [];
-        if (!config.BLOCK_STORE_FS_TMFS_ENABLED) return { moved_block_ids };
 
         await P.map_with_concurrency(10, block_ids, async block_id => {
             try {
-                if (await this._move_block_to_glacier(block_id)) moved_block_ids.push(block_id);
+                if (await this._move_block_to_tmfs(block_id)) moved_block_ids.push(block_id);
             } catch (err) {
-                dbg.warn(`_move_block_to_glacier block ${block_id} failed due to`, err);
+                dbg.warn(`_move_block_to_tmfs block ${block_id} failed due to`, err);
             }
         });
 
         return { moved_block_ids };
-    }
-
-    /**
-     * @note Supports only Tier2/TMFS FS_ARCHIVE for others will 
-     * simply return false
-     * @param {string} block_id 
-     * @returns {Promise<boolean>}
-     */
-    async _move_block_to_glacier(block_id) {
-        return this._move_block_to_tmfs(block_id);
     }
 
     async _move_block_to_tmfs(block_id) {
--- a/src/sdk/map_client.js
+++ b/src/sdk/map_client.js
@@ -17,7 +17,6 @@ const Semaphore = require('../util/semaphore');
 const KeysSemaphore = require('../util/keys_semaphore');
 const block_store_client = require('../agent/block_store_services/block_store_client').instance();
 const system_store = require('../server/system_services/system_store').get_instance();
-const js_utils = require('../util/js_utils');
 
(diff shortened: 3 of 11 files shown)


src/agent/block_store_services/block_store_fs.js
```diff
@@ -286,13 +289,12 @@ class BlockStoreFs extends BlockStoreBase {
...
+    async _move_blocks_to_tmfs(block_ids) {
         const moved_block_ids = [];
-        if (!config.BLOCK_STORE_FS_TMFS_ENABLED) return { moved_block_ids };
 
         await P.map_with_concurrency(10, block_ids, async block_id => {
             try {
-                if (await this._move_block_to_glacier(block_id)) moved_block_ids.push(block_id);
+                if (await this._move_block_to_tmfs(block_id)) moved_block_ids.push(block_id);
             } catch (err) {
                 dbg.warn(`_move_block_to_glacier block ${block_id} failed due to`, err);
```
guymguym, line 299: fix fn name in log print

src/sdk/map_client.js
```diff
@@ -578,7 +577,7 @@ class MapClient {
             const current_attached_pools = tier_before_move.mirrors.map(mirror => mirror.spread_pools.map(pool => String(pool._id))).flat();
 
             if (
-                !js_utils.compare_unordered(target_attached_pools, current_attached_pools, true) ||
+                !_.xor(target_attached_pools, current_attached_pools) ||
```
guymguym, line 580: _.xor returns an array, which is always evaluated as a true value, but we need to check if it is non empty.
```suggestion
                _.xor(target_attached_pools, current_attached_pools).length ||
```

src/server/system_services/bucket_server.js
```diff
@@ -151,6 +153,36 @@ async function create_bucket(req) {
             );
             changes.insert.tieringpolicies = [tiering_policy];
             changes.insert.tiers = [tier];
+
+            // Attach a `GLACIER` tier to the bucket if it is not namespace.caching
+            // and appropriate configuration is set
+            if (!req.rpc_params.namespace?.namespace?.caching) {
```
guymguym, line 159: Isn't it a double namespace there?
But also, this also needs to prevent applying this to any namespace buckets (not just ones with caching option set).

```suggestion
            if (!req.rpc_params.namespace) {
```

src/server/system_services/bucket_server.js
```diff
@@ -151,6 +153,36 @@ async function create_bucket(req) {
...
+                    tier_server.check_tier_exists(req, tier2_name);
+                    const tier2 = tier_server.new_tier_defaults(
+                        tier2_name,
+                        req.system._id,
+                        chunk_config._id,
+                        tier2_mirrors,
+                        STORAGE_CLASS_GLACIER,
+                    );
+
+                    changes.insert.tiers.push(tier2);
```
guymguym, line 177: There's also first.bucket that gets created in system_server create_system.
Would be good to share this code flow for both cases.

config.js
```diff
@@ -702,6 +702,12 @@ config.TIERING_TTL_WORKER_BATCH_SIZE = 1000;
 config.TIERING_TTL_WORKER_BATCH_DELAY = 1 * 60 * 1000; // 1 minutes
 config.TIERING_TTL_MS = 30 * 60 * 1000; // 30 minutes
 
+//////////////////////////
+//         TMFS         //
+//////////////////////////
+
+config.TMFS_BUCKET_TIER2_AUTOCONF = false;
```
guymguym, line 709: I think this config should actually be defined as a generic bucket tiering option, and not specific to TMFS.
This will allow us to use it also for other buckets such as with S3 backing stores that can automatically tier to secondary storage class.
Meaning move it to be somewhere in the // SYSTEM CONFIG // area, like:
```suggestion
config.BUCKET_AUTOCONF_TIER2_ENABLED = false;
```

guymguym (changes requested):
Few changes needed. Thanks

src/server/system_services/bucket_server.js
utkarsh-pro: Moved this into a common function. Doesn't handles any customizations yet but does extend the functionality to create_system.

Tested: `npm run web` => `first.bucket` had 2 tiers attached to it.

src/server/system_services/system_server.js
```diff
@@ -339,6 +339,8 @@ async function create_system(req) {
     try {
         const account_id = system_store.new_system_store_id();
         const changes = new_system_changes(name, account_id);
+        bucket_server.auto_setup_tier2(req, changes, true);
```
guymguym, line 342: I think it would make it a bit more robust if this call was inside new_system_changes...

src/server/system_services/bucket_server.js
```diff
@@ -76,6 +77,60 @@ function new_bucket_defaults(name, system_id, tiering_policy_id, owner_account_i
...
+
+    const default_pool_id = initial_tier.mirrors?.[0]?.spread_pools?.[0];
+    if (!default_pool_id) return;
+
+    const initial_tier_name = initial_tier.name;
+    if (!initial_tier_name) return;
+
+    const tier2_name = `${initial_tier_name}_auto_tier2`;
+
+    const system_id = req.system?._id || changes.insert?.systems?.[0]?._id;
```
guymguym, line 101: I think it's easier to get it from tier1

src/server/system_services/bucket_server.js
```diff
@@ -76,6 +77,60 @@ function new_bucket_defaults(name, system_id, tiering_policy_id, owner_account_i
...
 
+function auto_setup_tier2(req, changes, skip_check = false) {
+    if (!config.BUCKET_AUTOCONF_TIER2_ENABLED) return;
+    if (!changes.insert) return;
+
+    const tiering_policies = changes.insert.tieringpolicies;
+    if (!tiering_policies) return;
+
+    const initial_tier = changes.insert.tiers?.[0];
+    if (!initial_tier) return;
```
guymguym, line 88: It's not bad, but how about sending the tier1 and tiering_policy objects as arguments instead of assuming their location in the changes object?

src/server/system_services/bucket_server.js
```diff
@@ -76,6 +77,60 @@ function new_bucket_defaults(name, system_id, tiering_policy_id, owner_account_i
...
+
+    const chunk_config_id = initial_tier?.chunk_config;
+    if (!chunk_config_id) return;
```
guymguym, line 91: I think we better remove this silent check here, because a tier must have a chunk_config so there is no need to allow one without.

src/server/system_services/bucket_server.js
```diff
@@ -76,6 +77,60 @@ function new_bucket_defaults(name, system_id, tiering_policy_id, owner_account_i
...
+
+    const default_pool_id = initial_tier.mirrors?.[0]?.spread_pools?.[0];
+    if (!default_pool_id) return;
+
+    const initial_tier_name = initial_tier.name;
+    if (!initial_tier_name) return;
```
guymguym, line 97: no need really, a tier without a name is not a real tier :)

src/server/system_services/bucket_server.js
```diff
@@ -76,6 +77,60 @@ function new_bucket_defaults(name, system_id, tiering_policy_id, owner_account_i
...
+        system_id,
+        chunk_config_id,
+        tier2_mirrors,
+        STORAGE_CLASS_GLACIER,
+    );
+
+    changes.insert.tiers.push(tier2);
+    init_tiering_policy.tiers.push({
+        tier: tier2._id,
+        order: 1,
```
guymguym, line 128: perhaps take the previous order + 1? not critical though.

guymguym (commented):
I added few small comments. I love how you reused the code in auto_setup_tier2!

src/server/system_services/bucket_server.js
utkarsh-pro: I missed this comment actually

guymguym (approved):

PR #7362, opened 2023-06-27
Fix CVE-2019-15597 - Implement new DF command instead of library function

nadavMiz:
### Explain the changes
1. implemented a local function to replace the unmaintained node-df library. this function is based on the node-df implementation with changes to fit our code better (https://github.com/adriano-di-giovanni/node-df).
2. add '\' before all special characters in the file name so bash will read them as part of the file name itself. this is to prevent the code injection suggested in  CVE-2019-15597

1. https://bugzilla.redhat.com/show_bug.cgi?id=2124534

### Testing Instructions:
1. df is used mainly during node creation, so create a new backing store and validate that it was created successfully

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/util/os_utils.js
+++ b/src/util/os_utils.js
@@ -8,7 +8,6 @@ const ps = require('ps-node');
 const util = require('util');
 const path = require('path');
 const moment = require('moment-timezone');
-const node_df = require('node-df');
 const blockutils = require('linux-blockutils');
 const child_process = require('child_process');
 const ip_module = require('ip');
@@ -341,15 +340,66 @@ function get_disk_mount_points() {
         });
 }
 
+//https://github.com/adriano-di-giovanni/node-df/blob/master/lib/parse.js
+function _df_parse(stdout) {
+    return stdout
+            .trim()
+            .split(/\r\n|\r|\n/) // split into rows
+            .slice(1) // strip column headers away
+            .map(row => {
+                const columns = row
+                // one or more whitespaces followed by one or more digits
+                // must be interpreted as column delimiter
+                .replace(/\s+(\d+)/g, '\t$1')
+                // one or more whitespaces followed by a slash
+                // must be interpreted as the last column delimiter
+                .replace(/\s+\//g, '\t/')
+                // split into columns
+                .split(/\t/);
 
-function get_mount_of_path(file_path) {
+        return {
+            filesystem: columns[0],
+            size: parseInt(columns[1], 10),
+            used: parseInt(columns[2], 10),
+            available: parseInt(columns[3], 10),
+            capacity: parseInt(columns[4], 10) * 0.01,
+            mount: columns.slice(5).join(' '),
+        };
+    });
+}
+
+//https://github.com/adriano-di-giovanni/node-df/blob/master/lib/index.js
+async function _df(file_path, flags = '') {
+    const file_name_regex = /[^-_./a-zA-Z0-9]/g;
+    const flags_argument = '-kP' + flags;
+    const args = [flags_argument];
+    if (file_path) {
+        // CVE-2019-15597 add '\' to special charecters to prevent command injection
+        file_path = file_path.replace(file_name_regex, match => '\\' + match);
+        args.push(file_path);
+    }
+    try {
+    const execFile = util.promisify(child_process.execFile);
+    const { stdout, stderr } = await execFile('/bin/df', args);
+    if (stderr) {
+        dbg.error("df command failed:", stderr);
+        throw new Error(stderr);
+    }
+    return _df_parse(stdout);
+    } catch (err) {
+        dbg.error('df command failed:', err);
+        throw err;
+    }
+}
+
+async function get_mount_of_path(file_path) {
     console.log('get_mount_of_path');
-    return P.fromCallback(callback => node_df({
-            file: file_path
-        }, callback))
-        .then(function(drives) {
-            return drives && drives[0] && drives[0].mount;
-        });
+    try {
+        const drives = await _df(file_path);
+        return drives?.[0]?.mount;
+    } catch (err) {
+        dbg.error(err);
+    }
 }
 
 
@@ -365,11 +415,15 @@ async function get_block_device_sizes() {
 }
 
 async function get_drive_of_path(file_path) {
-    const volumes = await P.fromCallback(callback => node_df({ file: file_path }, callback));
-    const vol = volumes && volumes[0];
-    if (vol) {
-        const size_by_bd = await get_block_device_sizes();
-        return linux_volume_to_drive(vol, size_by_bd);
+    try {
+        const volumes = await _df(file_path);
+        const vol = volumes?.[0];
+        if (vol) {
+            const size_by_bd = await get_block_device_sizes();
+            return linux_volume_to_drive(vol, size_by_bd);
+        }
+    } catch (err) {
+        dbg.error(err);
     }
 }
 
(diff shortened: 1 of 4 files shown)


nimrod-becker:
lets add a unit test

src/util/os_utils.js
```diff
@@ -341,12 +340,62 @@ function get_disk_mount_points() {
...
+        if (stderr) {
+            dbg.error("df command failed:", stderr);
+            callback(new Error(stderr), null);
+            return;
+        }
+        const drives = _df_parse(stdout);
+        callback(null, drives);
+
+     });
+}
```
liranmauda, line 394: Not sure we should use callbacks anymore, maybe use promisify? 
```suggestion
async function _df(file_path, flags = '') {
    const file_name_regex = /[^-_./a-zA-Z0-9]/g;
    const flags_argument = '-kP' + flags;
    const args = [flags_argument];

    if (file_path) {
        // CVE-2019-15597 add '\' to special characters to prevent command injection
        file_path = file_path.replace(file_name_regex, match => '\\' + match);
        args.push(file_path);
    }

    try {
        const execFile = util.promisify(child_process.execFile);
        const { stdout, stderr } = await execFile('/bin/df', args);
        if (stderr) {
            throw new Error(stderr);
        }
        return _df_parse(stdout);
    } catch (err) {
        console.error('df command failed:', err);
        throw err;
    }
}
```

src/util/os_utils.js
```diff
@@ -341,12 +340,62 @@ function get_disk_mount_points() {
         });
 }
 
+function _df_parse(stdout) {
+    return stdout
+            .trim()
+            .split(/\r\n|\r|\n/) // split into rows
+            .slice(1) // strip column headers away
+            .map(function(row) {
```
liranmauda, line 348: maybe use arrow function:
```suggestion
            .map((row) =>  {
```

liranmauda (changes requested):

nadavMiz:
> lets add a unit test

since df checks the available space on the whole file system, which I have no control over, I can't test the validity of the actual values. so instead I just check that the command returns a valid response and that no errors are thrown

src/util/os_utils.js
```diff
@@ -341,12 +340,62 @@ function get_disk_mount_points() {
...
+            size: parseInt(columns[1], 10),
+            used: parseInt(columns[2], 10),
+            available: parseInt(columns[3], 10),
+            capacity: parseInt(columns[4], 10) * 0.01,
+            mount: columns.slice(5).join(' '),
+        };
+    });
+}
+
+async function df(file_path, callback, flags = '') {
```
liranmauda, line 370: It is only meant to be used here (yes, and tests) 
we should call it with a private name convention 
```suggestion
async function _df(file_path, callback, flags = '') {
```

src/test/unit_tests/test_os_utils.js
```diff
@@ -0,0 +1,44 @@
+/* Copyright (C) 2016 NooBaa */
```
liranmauda, line 1: I think that we should write the tests using `jest` and not `mocha`
@dannyzaken your input?

liranmauda (changes requested):

src/util/os_utils.js
```diff
@@ -341,12 +340,62 @@ function get_disk_mount_points() {
...
+    const file_name_regex = /[^-_./a-zA-Z0-9]/g;
+    const flags_argument = '-kP' + flags;
+    const args = [flags_argument];
+    if (file_path) {
+        // CVE-2019-15597 add '\' to special charecters to prevent command injection
+        file_path = file_path.replace(file_name_regex, match => '\\' + match);
+        args.push(file_path);
+    }
+    child_process.execFile('/bin/df', args, function(err, stdout, stderr) {
```
utkarsh-pro, line 379: On my linux VM, this path is `/usr/bin/df` while on my mac it is `/bin/df`. Maybe we can do a path lookup? WDYT?

nadavMiz: @utkarsh-pro This is the same code we used before. I just copied it from the library(https://github.com/adriano-di-giovanni/node-df/blob/master/lib/index.js#L36). maybe on linux its both /bin/df and /usr/bin/df? that's the case on my Linux

src/test/unit_tests/jest_tests/test_os_utils.js
```diff
@@ -0,0 +1,57 @@
...
+        expect(typeof(drive.filesystem)).toBe('string');
+        expect(typeof(drive.size)).toBe('number');
+        expect(typeof(drive.used)).toBe('number');
+        expect(typeof(drive.available)).toBe('number');
+        expect(typeof(drive.capacity)).toBe('number');
+        expect(typeof(drive.mount)).toBe('string');
+      });
+    });
+  });
+});
```
liranmauda, line 57: Missing new line

liranmauda (changes requested):

nadavMiz:
I have added try catch to df calls with file name since now the command throws if the file doesn't exist and does not just return a null value

src/test/unit_tests/jest_tests/test_os_utils.js
```diff
@@ -0,0 +1,60 @@
+/* Copyright (C) 2016 NooBaa */
```
liranmauda, line 1: please rename this file to `<test_name>.test.js`

src/util/os_utils.js
```diff
@@ -341,15 +340,68 @@ function get_disk_mount_points() {
...
-        }, callback))
+    try {
+        return _df(file_path)
         .then(function(drives) {
             return drives && drives[0] && drives[0].mount;
         });
+    } catch (err) {
+        dbg.error(err);
+    }
 }
```
liranmauda, line 405: we can make this function asynced, and also use optional chaining:
```suggestion
async function get_mount_of_path(file_path) {
  console.log('get_mount_of_path');
  try {
    const drives = await _df(file_path);
    return drives?.[0]?.mount;
  } catch (err) {
    dbg.error(err);
  }
}
```

src/util/os_utils.js
```diff
@@ -975,6 +1023,7 @@ exports.read_drives = read_drives;
 exports.get_raw_storage = get_raw_storage;
 exports.remove_linux_readonly_drives = remove_linux_readonly_drives;
 exports.get_main_drive_name = get_main_drive_name;
+exports.df = _df;
```
liranmauda, line 1026: this export is only meant for tests, I wonder if we should export it as an internal name
```suggestion
exports._df = _df;
```
I am not that opinionated here, but if we want to pass the intention that this function is only meant for this file's internal use, then the export should provide the same intention.

I feel it is similar to writing an internal function inside a class and exporting the class.
@guymguym WDYT?

src/util/os_utils.js
```diff
@@ -365,11 +415,15 @@ async function get_block_device_sizes() {
...
 
 async function get_drive_of_path(file_path) {
-    const volumes = await P.fromCallback(callback => node_df({ file: file_path }, callback));
-    const vol = volumes && volumes[0];
-    if (vol) {
-        const size_by_bd = await get_block_device_sizes();
-        return linux_volume_to_drive(vol, size_by_bd);
+    try {
+        const volumes = await _df(file_path);
+        const vol = volumes && volumes[0];
```
liranmauda, line 420:
```suggestion
        const vol = volumes?.[0];
```

src/util/os_utils.js
nadavMiz: @utkarsh-pro can I resolve this comment?

src/util/os_utils.js
nadavMiz: I have changed it to be _df. @guymguym any comments?

liranmauda (approved):
LGTM

PR #7369, opened 2023-07-02
NSFS | Versioning | Suspended Mode Delete Object

shirady:
### Explain the changes
1. Add NSFS Versioning Suspended mode to support delete-object operation.
2. Minor changes in comments and formatting in code lines related to NSFS | Versioning | Suspended Mode Put Object.
 
### Issues: Gaps
1. In suspended mode changes we used only "safe" filesystem functions (safe_move, safe_unlink), later we might change it according to the specific case (we also don't have concurrency tests).
3. GPFS was not tested yet with these changes.
4. We currently don't keep `XATTR_PREV_VERSION_ID` updated in case there is a deletion of a specific version id or in case we delete an old null version from the `.versions/` directory.
5. When we have a delete-object request we don't validate that it was requested from the bucket owner.
 
#### Manual Tests:
6. Use AWS CLI and use the put-object command (here is a link to their [doc](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/put-object.html)).
7. Set the bucket to versioning Enabled and put-object of another key (here is a link to their [doc](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/put-bucket-versioning.html)).
8. Set versioning to suspended and put versions of the same object that were uploaded before (steps 4 and 5).
9. Use AWS CLI and use the delete-object command (here is a link to their [doc](https://docs.aws.amazon.com/cli/latest/reference/s3api/delete-object.html)).
 
While you add and delete objects to the bucket you can view the changes
2. Move to the bucket path (if you did it as written in the wiki it would be `cd /nsfs/fs1/bucket-path/`).
 
#### Run Unit Tests:
2. From the core tab in the terminal run: `make root-perm-test` (You should see "140 passing" at the end of the output).
 
- [X] Tests added: NSFS Versioning Suspended mode: delete-object (latest),  delete-object version-ID, and delete-objects (based on existing tests that we have in NSFS Versioning Enabled mode).

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1175,8 +1175,8 @@ class NamespaceFS {
     //     versioning is enabled
     //     OR
     //     versioning is suspended AND latest version is a unique id (not null version)
-    //       2.1.1 create .versions/ if it doesn't exist
-    //       2.1.2 move latest version to .versions/
+    //       3.2 create .versions/ if it doesn't exist
+    //       3.3 move latest version to .versions/
     // 4. move new version to latest_ver_path (key path)
     // retry safe linking a file in case of parallel put/delete of the source path
     async _move_to_dest_version(fs_context, new_ver_tmp_path, latest_ver_path, upload_file, key, open_mode) {
@@ -1195,7 +1195,7 @@ class NamespaceFS {
                 gpfs_options = is_gpfs ?
                     await this._open_files_gpfs(fs_context, new_ver_tmp_path, latest_ver_path, upload_file,
                         latest_ver_info, open_mode, undefined, versioned_info) :
-                        undefined;
+                    undefined;
 
                 dbg.log1('Namespace_fs._move_to_dest_version:', latest_ver_info, new_ver_info, gpfs_options);
 
@@ -1204,13 +1204,13 @@ class NamespaceFS {
                         dbg.log1('NamespaceFS._move_to_dest_version suspended: version ID of the latest version is null - the file will be unlinked');
                         await this.safe_unlink(fs_context, latest_ver_path, latest_ver_info, gpfs_options);
                     } else {
-                    // remove a version (or delete marker) with null version ID from .versions/ (if exists)
-                    await this._delete_null_version_from_versions_directory(key, fs_context);
+                        // remove a version (or delete marker) with null version ID from .versions/ (if exists)
+                        await this._delete_null_version_from_versions_directory(key, fs_context);
                     }
                 }
                 if (latest_ver_info &&
                     ((this._is_versioning_enabled()) ||
-                    (this._is_versioning_suspended() && latest_ver_info.version_id_str !== NULL_VERSION_ID))) {
+                        (this._is_versioning_suspended() && latest_ver_info.version_id_str !== NULL_VERSION_ID))) {
                     dbg.log1('NamespaceFS._move_to_dest_version version ID of the latest version is a unique ID - the file will be moved it to .versions/ directory');
                     await this._make_path_dirs(versioned_path, fs_context);
                     await this.safe_move(fs_context, latest_ver_path, versioned_path, latest_ver_info,
@@ -1496,12 +1496,10 @@ class NamespaceFS {
             let res;
             if (this._is_versioning_disabled()) {
                 await this._delete_single_object(fs_context, file_path, params);
-            } else if (this._is_versioning_enabled()) {
+            } else {
                 res = params.version_id ?
                     await this._delete_version_id(fs_context, file_path, params) :
                     await this._delete_latest_version(fs_context, file_path, params);
-            } else {
-                throw new Error('TODO');
             }
             return res || {};
         } catch (err) {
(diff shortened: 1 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -2338,40 +2337,70 @@ class NamespaceFS {
...
 
-                    gpfs_options = is_gpfs && await this._open_files_gpfs(fs_context, latest_ver_path,
-                        undefined, undefined, undefined, undefined, true, versioned_info);
+                gpfs_options = is_gpfs ?
+                    await this._open_files_gpfs(fs_context, latest_ver_path,
+                        undefined, undefined, undefined, undefined, true, versioned_info) :
+                    undefined;
 
+                dbg.log1('Namespace_fs._delete_latest_version:', latest_ver_info, versioned_path, versioned_info, gpfs_options);
+                if (this._is_versioning_suspended()) {
```
romayalon, line 2377: suggesting the following structure instead - 
```
if (latest_ver_info) {
    const suspended_and_latest_is_not_null = this._is_versioning_suspended() && latest_ver_info.version_id_str !== NULL_VERSION_ID
    if (this._is_versioning_enabled() || suspended_and_latest_is_not_null){
        await this._make_path_dirs(versioned_path, fs_context);
        await this.safe_move(fs_context, latest_ver_path, versioned_path, latest_ver_info,
                    gpfs_options && gpfs_options.delete_version);
        if (suspended_and_latest_is_not_null) {
            // remove a version (or delete marker) with null version ID from .versions/ (if exists)
            await this._delete_null_version_from_versions_directory(params.key, fs_context);
        }        
    } else {
        // versioning suspended and version_id is null
        dbg.log1('NamespaceFS._delete_latest_version: suspended mode version ID of the latest version is null - file will be unlinked');
        await this.safe_unlink(fs_context, latest_ver_path, latest_ver_info, gpfs_options);
    }  
}
```

src/sdk/namespace_fs.js
```diff
@@ -2419,15 +2448,35 @@ class NamespaceFS {
...
-                const file_path = this._get_version_path(params.key, delete_marker_version_id);
+                if (this._is_versioning_enabled()) {
+                    // the delete marker path built from its version info (mtime + ino)
+                    delete_marker_version_id = this._get_version_id_by_stat(stat);
+                    file_path = this._get_version_path(params.key, delete_marker_version_id);
+                } else {
+                    // the delete marker file name would be with a 'null' suffix
+                    delete_marker_version_id = NULL_VERSION_ID;
+                    file_path = this._get_version_path(params.key, delete_marker_version_id);
+                }
```
romayalon, line 2461: let's move ` file_path = this._get_version_path(params.key, delete_marker_version_id);` outside of the conditions

src/sdk/namespace_fs.js
```diff
@@ -2419,15 +2448,35 @@ class NamespaceFS {
...
 
                 // finish upload part
-                const fs_xattr = await this._assign_versions_to_fs_xattr(fs_context, deleted_version_info,
+                let fs_xattr;
+                if (this._is_versioning_suspended() &&
+                    (deleted_version_info?.version_id_str === NULL_VERSION_ID)) {
+                    // the previous version will be the previous version of the deleted version
+                    const prev_version_info = await this.get_prev_version_info(fs_context, params.key,
```
romayalon, line 2468: in the design meetings, we decided to exclude this prev linked list maintenance due to concurrent deletions/puts, so maybe let's try to re think about it in a different PR/design

src/sdk/namespace_fs.js
```diff
@@ -2419,15 +2448,35 @@ class NamespaceFS {
...
+                    // the delete marker path built from its version info (mtime + ino)
+                    delete_marker_version_id = this._get_version_id_by_stat(stat);
+                    file_path = this._get_version_path(params.key, delete_marker_version_id);
+                } else {
+                    // the delete marker file name would be with a 'null' suffix
+                    delete_marker_version_id = NULL_VERSION_ID;
+                    file_path = this._get_version_path(params.key, delete_marker_version_id);
+                }
 
                 // finish upload part
```
romayalon, line 2463: not sure if it's relevant, check and if needed please remove

src/sdk/namespace_fs.js
```diff
@@ -2434,15 +2458,29 @@ class NamespaceFS {
...
+                    delete_marker_version_id = NULL_VERSION_ID;
+                }
                 const file_path = this._get_version_path(params.key, delete_marker_version_id);
 
-                // finish upload part
-                const fs_xattr = await this._assign_versions_to_fs_xattr(fs_context, deleted_version_info,
+                let fs_xattr;
+                if (this._is_versioning_suspended() &&
+                    (deleted_version_info?.version_id_str === NULL_VERSION_ID)) {
+                    fs_xattr = await this._assign_versions_to_fs_xattr(fs_context, undefined,
```
romayalon, line 2474: if the deleted version was unique, it'll move to .versions/, and in that case we do need to update the prev version of the delete_marker as we do for versioning enabled

romayalon (approved):

PR #7371, opened 2023-07-02
NSFS | Versioning | List Object Version - Support for Null Version

shirady:
### Explain the changes
1. NSFS versioning list-object-versions operation - add support for null version id (details about the [gap](https://github.com/noobaa/noobaa-core/pull/7203#discussion_r1208549398)). 
Before this change when we had a version id of null the behavior was not expected, when I test it I saw:
Screenshot 2023-06-27 at 10 38 11
2. Change the method `_get_version_id_by_stat` to function outside of the `NameSpaceFS` class.
 
### Issues: Gaps
1. GPFS was not tested yet with these changes.
 
#### Manual Tests:
6. Use AWS CLI and use the put-object command (here is a link to their [doc](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/put-object.html)).
7. Set the bucket to versioning Enabled and put-object of another key (here is a link to their [doc](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/s3api/put-bucket-versioning.html)).
8. Set versioning to suspended and put versions of the same object that were uploaded before (steps 4 and 5).
9. Use AWS CLI and use the list-object-versions command (here is a link to their [doc](https://docs.aws.amazon.com/cli/latest/reference/s3api/list-object-versions.html)).
 
While you add objects to the bucket you can view the changes
2. Move to the bucket path (if you did it as written in the wiki it would be `cd /nsfs/fs1/bucket-path/`).
 
##### Test stat error edge case (and no old_version edge case):
Status: When we have .versions directory and inside it only 1 null version, add an error after stat. Should list object version without null version.

#### Run Unit Tests:
2. From the core tab in the terminal run: `make root-perm-test` (You should see "144 passing" at the end of the output).
 
- [X] Tests added: NSFS list object version for null version cases only (using prefix to decrease the list and focus on the null version tests).

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -45,6 +45,7 @@ const XATTR_DELETE_MARKER = XATTR_USER_PREFIX + 'delete_marker';
 const XATTR_DIR_CONTENT = XATTR_USER_PREFIX + 'dir_content';
 const HIDDEN_VERSIONS_PATH = '.versions';
 const NULL_VERSION_ID = 'null';
+const NULL_VERSION_SUFFIX = '_' + NULL_VERSION_ID;
 
 const INTERNAL_XATTR = [
     XATTR_CONTENT_TYPE,
@@ -88,6 +89,23 @@ function sort_entries_by_name(a, b) {
     return 0;
 }
 
+function _get_version_id_by_stat({ino, mtimeNsBigint}) {
+    // TODO: GPFS might require generation number to be added to version_id
+    return 'mtime-' + mtimeNsBigint.toString(36) + '-ino-' + ino.toString(36);
+}
+
+function _is_version_object_including_null_version(filename) {
+    const is_version_object = _is_version_object(filename);
+    if (!is_version_object) {
+        return _is_version_null_version(filename);
+    }
+    return is_version_object;
+}
+
+function _is_version_null_version(filename) {
+    return filename.endsWith(NULL_VERSION_SUFFIX);
+}
+
 function _is_version_object(filename) {
     const mtime_substr_index = filename.indexOf('_mtime-');
     if (mtime_substr_index < 0) return false;
@@ -110,6 +128,8 @@ function _get_mtime_from_filename(filename) {
 function _get_filename(file_name) {
     if (_is_version_object(file_name)) {
         return file_name.substring(0, file_name.indexOf('_mtime-'));
+    } else if (_is_version_null_version(file_name)) {
+        return file_name.substring(0, file_name.indexOf(NULL_VERSION_SUFFIX));
     }
     return file_name;
 }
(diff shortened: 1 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -2421,7 +2448,7 @@ class NamespaceFS {
 
                 // the delete marker path built from its version info (mtime + ino)
                 const stat = await upload_params.target_file.stat(fs_context);
-                delete_marker_version_id = this._get_version_id_by_stat(stat);
+                delete_marker_version_id = NamespaceFS.get_version_id_by_stat(stat);
```
v-harihar, line 2451: Do we require call using `NamespaceFs` that `this` ?

src/sdk/namespace_fs.js
```diff
@@ -2077,13 +2104,13 @@ class NamespaceFS {
...
 
-    _get_version_id_by_stat({ino, mtimeNsBigint}) {
+    static get_version_id_by_stat({ino, mtimeNsBigint}) {
         // TODO: GPFS might require generation number to be added to version_id
         return 'mtime-' + mtimeNsBigint.toString(36) + '-ino-' + ino.toString(36);
     }
 
     _get_version_id_by_mode(stat) {
-        if (this._is_versioning_enabled()) return this._get_version_id_by_stat(stat);
+        if (this._is_versioning_enabled()) return NamespaceFS.get_version_id_by_stat(stat);
```
v-harihar, line 2113: Do we require to follow this convention ?

src/sdk/namespace_fs.js
```diff
@@ -2077,13 +2104,13 @@ class NamespaceFS {
         return this.versioning === versioning_status_enum.VER_SUSPENDED;
     }
 
-    _get_version_id_by_stat({ino, mtimeNsBigint}) {
+    static get_version_id_by_stat({ino, mtimeNsBigint}) {
```
v-harihar, line 2107: Do we require to follow this convention here as in most place we still have internal functions starting with`_`

vh05 (changes requested):

vh05 (commented):
/lgtm 

I feel one more approval would be better

src/sdk/namespace_fs.js
```diff
@@ -2697,4 +2725,23 @@ class NamespaceFS {
...
 
+    // This is helper function for list object version
+    // In order to sort the entries by name we would like to change the name of files with suffix of '_null'
+    // to have the structure of _mtime-...-ino-... as version id.
+    // This function returns a set that contains all file names that were changed (after change)
+    async function _rename_null_version(old_versions, fs_context, version_path) {
+        const renamed_null_versions = new Set();
+        for (const old_version of old_versions) {
+            if (_is_version_null_version(old_version.name)) {
+                const stat = await nb_native().fs.stat(fs_context, path.join(version_path, old_version.name));
```
romayalon, line 2736: error handling and removal from the array when the error is ENOENT,
need to rethink about other errors

src/sdk/namespace_fs.js
```diff
@@ -2085,13 +2113,13 @@ class NamespaceFS {
         return this.versioning === versioning_status_enum.VER_SUSPENDED;
     }
 
-    _get_version_id_by_stat({ino, mtimeNsBigint}) {
+    static _get_version_id_by_stat({ino, mtimeNsBigint}) {
```
romayalon, line 2116: take it outside the class

src/sdk/namespace_fs.js
```diff
@@ -309,9 +360,25 @@ const versions_dir_cache = new LRUCache({
...
             const latest_versions = await nb_native().fs.readdir(fs_context, dir_path);
             if (is_version_path_exists) {
-                const old_versions = await nb_native().fs.readdir(fs_context, version_path);
-                const entries = latest_versions.concat(old_versions);
+                let old_versions = await nb_native().fs.readdir(fs_context, version_path);
+                // In case we have a null version id inside .versions/ directory we will rename it
+                // Therefore, old_versions will not include an entry with 'null' suffix
+                const { renamed_old_null_versions, error_on_stat } = await _rename_null_version(old_versions, fs_context, version_path);
+                // Remove null versions that could not be rename (error on stat) and their name changed to undefined
+                if (error_on_stat) old_versions = old_versions.filter(old_version => old_version.name);
```
romayalon, line 368: hmm so that's another iteration on a possibly very large array, can you maybe use reduce instead?

shirady: Why would we prefer `reduce` over `filter` in this case?
It would be less readable: 
```suggestion
                if (error_on_stat) old_versions = old_versions.reduce((accumulator, old_version) => {
                    if (old_version.name) accumulator.push(old_version);
                    return accumulator;
                  }, []);
```

romayalon: I meant that in order to avoid another iteration on the array you can do the filter inside _rename_null_version in a single reduce function.

shirady: Thanks, I change it so we won't do another iteration in case stat fails.

romayalon (approved):

vh05 (approved):

PR #7373, opened 2023-07-03
Fix `INVALID_SCHEMA_DB objectstats ERRORS`

tangledbytes:
### Explain the changes
This PR fixes `INVALID_SCHEMA_DB objectstats ERRORS` that we encounter in our endpoints.

The root cause for the issue was due to the fact that postgres (by default/without `COALASCE`) does not create nested fields. This problem is being fixed in this PR via a hack field in `findOneAndUpdate`'s `options` parameter called `upsert_options` which creates the necessarily fields if the entry is new.

Other solutions to the problem were also explored like:
1. Changing the upstream dependency to handle `upserts` better. Rejected due to potentially slow dev cycle.
2. Update noobaa internal methods to be smarter with handling upserts. Rejected due to fear of regression on other part of codebase.
3. Using 2 `findOneAndUpdate`. Works but doesn't works both theoretically and practically. Too many races. Rejected.

Although the changes are kind of postgres specific and tested on postgres exclusively, I fully expect this to function well in mongo as well.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/analytic_services/endpoint_stats_store.js
+++ b/src/server/analytic_services/endpoint_stats_store.js
@@ -68,19 +68,31 @@ class EndpointStatsStore {
     async _update_s3_ops_counters(system, report) {
         const { usage = {}, errors = {} } = report.s3_ops;
         const selector = {
-            system: system._id
+            system: system._id,
         };
+        const options = {
+            upsert: true,
+            returnOriginal: false,
+
+            // Ineffective and not needed in mongo
+            //
+            // This is a postgres hack, this allows us to
+            // initialize the counters if the fields are missing
+            upsert_fields: {
+                s3_usage_info: usage,
+                s3_errors_info: errors
+            }
+        };
+
         const update = {
+            $set: {
+                system: system._id,
+            },
             $inc: {
                 ..._.mapKeys(usage, (unused, key) => `s3_usage_info.${key}`),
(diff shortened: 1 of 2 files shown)


src/server/analytic_services/endpoint_stats_store.js
```diff
@@ -68,19 +68,31 @@ class EndpointStatsStore {
...
+            returnOriginal: false,
+
+            // Ineffective and not needed in mongo
+            //
+            // This is a postgres hack, this allows us to
+            // initialize the counters if the fields are missing
+            upsert_fields: {
+                s3_usage_info: usage,
+                s3_errors_info: errors
+            }
```
guymguym, line 84: Hey @dannyzaken - PTAL
I remember you said there is something to fix for upsert - is that in the direction you where thinking?

dannyzaken: This fix looks good. In general, I think we should get rid of upserts altogether since the way we implement it in Postgres is too complex. This will require changing our stats stores that still use it

dannyzaken (approved):

src/server/analytic_services/endpoint_stats_store.js
utkarsh-pro: I agree, but we might still need `upsert` or at least an atomic `find` + `insert` (a transaction based wrapper maybe)?

PR #7375, opened 2023-07-03
passing log candidates as batch for and copy and delete

aayushchouhan09:
### Explain the changes
1. Currently one log object at a time is passed for copy and delete. In this PR we are trying to pass a batch of objects for both copy and delete individually.

### DIFF
--- a/src/server/bg_services/log_replication_scanner.js
+++ b/src/server/bg_services/log_replication_scanner.js
@@ -122,73 +122,91 @@ class LogReplicationScanner {
     async process_candidates(src_bucket, dst_bucket, candidates) {
         const src_dst_objects_list = await this.head_objects(src_bucket.name, dst_bucket.name, candidates);
 
+        dbg.log1('log_replication_scanner: process_candidates src_dst_objects_list: ', src_dst_objects_list);
+
+        const copy_keys = [];
+        const delete_keys = [];
+
         for (const candidate of src_dst_objects_list) {
-            await this.process_candidate(src_bucket, dst_bucket, candidate);
+            const action = this.process_candidate(candidate);
+            if (action === 'copy') {
+                copy_keys.push(candidate.key);
+            } else if (action === 'delete') {
+                delete_keys.push(candidate.key);
+            } else {
+                // Log skipped candidate key
+                dbg.log1('process_candidates: skipped_key: ', candidates.key);
+            }
+        }
+
+        dbg.log1('log_replication_scanner: process_candidates copy_keys: ', copy_keys);
+        dbg.log1('log_replication_scanner: process_candidates delete_keys: ', delete_keys);
+
+        // calling copy_objects and delete_objects by passing batch of keys
+        await Promise.all([
+            this.copy_objects(src_bucket, dst_bucket, copy_keys),
+            this.delete_objects(dst_bucket, delete_keys)
+        ]);
+
+        // returning copy_keys and delete_keys after processing candidates
+        return { copy_keys, delete_keys };
+    }
+
+    process_candidate(candidate) {
+        // if there is no log for candidate then we should skip
+        if (!candidate.data) {
+            return 'skip';
+        }
+
+        // each candidate represents latest action per key
+        const item = candidate.data;
+
+        if (item.action === 'copy') {
+            return this.process_copy_candidate(candidate);
+        }
+
+        if (item.action === 'delete') {
+            return this.process_delete_candidate(candidate);
+        }
+
+        if (item.action === 'conflict') {
+            return this.process_conflict_candidate(candidate);
         }
     }
 
-    async process_candidate(src_bucket, dst_bucket, candidate) {
-        // each candidate represents several action over the same key
-        // that are sorted by time.
-        const items = candidate.data || [];
-
-        for (const item of items) {
-            if (item.action === 'copy') {
-                await this.process_copy_candidate(src_bucket, dst_bucket, candidate);
-            }
-
-            if (item.action === 'delete') {
-                await this.process_delete_candidate(src_bucket, dst_bucket, candidate);
-            }
-
-            if (item.action === 'conflict') {
-                await this.process_conflict_candidate(src_bucket, dst_bucket, candidate);
-            }
-        }
-    }
-
-    async process_copy_candidate(src_bucket, dst_bucket, candidate) {
+    process_copy_candidate(candidate) {
         const src_object_info = candidate.src_object_info;
         const dst_object_info = candidate.dst_object_info;
         if (src_object_info && (!dst_object_info || (src_object_info.LastModified > dst_object_info.LastModified &&
             src_object_info.ETag !== dst_object_info.ETag))) {
-            await this.copy_object(src_bucket.name, dst_bucket.name, candidate.key);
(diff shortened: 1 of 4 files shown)


src/server/bg_services/log_replication_scanner.js
```diff
@@ -122,93 +122,79 @@ class LogReplicationScanner {
...
-            await this.process_candidate(src_bucket, dst_bucket, candidate);
+            await this.process_candidate(copy_keys, delete_keys, src_bucket, dst_bucket, candidate);
         }
+
+        this.copy_objects(src_bucket.name, dst_bucket.name, copy_keys);
+        this.delete_objects(dst_bucket.name, delete_keys);
     }
 
-    async process_candidate(src_bucket, dst_bucket, candidate) {
+    async process_candidate(copy_keys, delete_keys, src_bucket, dst_bucket, candidate) {
```
dannyzaken, line 136: `process_candidate` goes over all occurrences of a specific key and performs the operation for it. But this is redundant, and I think it can also result in wrong deletions.
I suggest processing only the most recent item in `items` and ignoring the rest

achouhan09: Agreed, I think we can store the most recent logs for the key in candidate itself.

src/server/bg_services/replication_log_parser.js
```diff
@@ -192,7 +192,8 @@ function create_candidates(logs) {
         const logs_for_key = Array.from(logs_for_key_time_map.values());
         const sorted_logs_for_key = logs_for_key.sort((a, b) => a.time.getTime() - b.time.getTime());
 
-        candidates[key] = sorted_logs_for_key;
+        // Storing only latest log per key in candidates
+        candidates[key] = [sorted_logs_for_key[sorted_logs_for_key.length - 1]];
```
utkarsh-pro, line 196: If we just care about the log entry with maximum timestamp then maybe we shouldn't be sorting them at all, just finding the log entry with maximum timestamp should do?

src/server/bg_services/log_replication_scanner.js
```diff
@@ -122,28 +122,37 @@ class LogReplicationScanner {
...
-            }
+        if (item.action === 'delete') {
+            return this.process_delete_candidate(src_bucket, dst_bucket, candidate);
+        }
 
-            if (item.action === 'conflict') {
-                await this.process_conflict_candidate(src_bucket, dst_bucket, candidate);
-            }
+        if (item.action === 'conflict') {
+            await this.process_conflict_candidate(src_bucket, dst_bucket, candidate);
```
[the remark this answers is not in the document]
achouhan09: right, Thanks.

src/server/bg_services/replication_log_parser.js
achouhan09: Yes, that will be fine.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -122,28 +122,36 @@ class LogReplicationScanner {
...
+        this.copy_objects(src_bucket, dst_bucket, copy_keys);
+        this.delete_objects(dst_bucket, delete_keys);
     }
 
     async process_candidate(src_bucket, dst_bucket, candidate) {
-        // each candidate represents several action over the same key
-        // that are sorted by time.
-        const items = candidate.data || [];
+        // each candidate represents latest action per key
+        const item = candidate.data && candidate.data[0];
```
utkarsh-pro, line 143: Not sure but here `item` can be `undefined` as well right? It might happen if we return an empty array from `create_candidates`?

src/server/bg_services/log_replication_scanner.js
```diff
@@ -122,28 +122,36 @@ class LogReplicationScanner {
...
+            const action = await this.process_candidate(src_bucket, dst_bucket, candidate);
+            if (action === 'copy') {
+                copy_keys.push(candidate.key);
+            } else if (action === 'delete') {
+                delete_keys.push(candidate.key);
+            }
         }
+
+        this.copy_objects(src_bucket, dst_bucket, copy_keys);
+        this.delete_objects(dst_bucket, delete_keys);
```
utkarsh-pro, line 138: Should we await them?

src/server/bg_services/log_replication_scanner.js
utkarsh-pro: That looks right.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -122,28 +122,44 @@ class LogReplicationScanner {
...
+            const action = await this.process_candidate(src_bucket, dst_bucket, candidate);
+            if (action === 'copy') {
+                copy_keys.push(candidate.key);
+            } else if (action === 'delete') {
+                delete_keys.push(candidate.key);
+            }
         }
+
+        await this.copy_objects(src_bucket, dst_bucket, copy_keys);
+        await this.delete_objects(dst_bucket, delete_keys);
```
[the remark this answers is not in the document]
achouhan09: Sure, Thanks. I will update the same.

tangledbytes (approved):
LGTM

src/server/bg_services/replication_log_parser.js
```diff
@@ -162,41 +162,39 @@ async function get_azure_log_candidates(source_bucket_id, rule_id, replication_c
...
+        const log_time = time.toString();
         const logs_for_key_time_map = logs_per_key[key];
-
-        // if there is already a candidate for the same key, then we have a conflict
-        if (logs_for_key_time_map.has(log.time)) {
+        // If there is already a candidate for the same key, then we have a conflict
+        if (logs_for_key_time_map.has(log_time)) {
             // TODO: Object versioning will raise false alarms here
-            const conflict_log = logs_for_key_time_map.get(log.time);
+            const conflict_log = logs_for_key_time_map.get(log_time);
```
dannyzaken, line 186: I think we should mark a conflict only if there are 2 different operations for the same key (copy vs delete). e.g.: if we have multiple uploads to the same key, we still need to copy the latest one and not treat it as a conflict

achouhan09: Sure, updated the code PTAL. Thanks

src/server/bg_services/log_replication_scanner.js
```diff
@@ -122,93 +122,111 @@ class LogReplicationScanner {
...
         const src_dst_objects_list = await this.head_objects(src_bucket.name, dst_bucket.name, candidates);
 
+        dbg.log1('log_replication_scanner: process_candidates src_dst_objects_list: ', src_dst_objects_list);
+
+        const copy_keys = [];
+        const delete_keys = [];
+
         for (const candidate of src_dst_objects_list) {
-            await this.process_candidate(src_bucket, dst_bucket, candidate);
+            const action = await this.process_candidate(candidate);
```
dannyzaken, line 131: unless I missed anything, `process_candidate` doesn't need to be async and there is no reason to wait for it.

achouhan09: yes right, no need to async updated the code. Thanks

dannyzaken (approved):

PR #7381, opened 2023-07-05
Add config option to disable node io detention

guymguym:
### Explain the changes
1. Backing store over TMFS have longer delays.
2. Added `config.NODE_IO_DETENTION_DISABLE` that can be set true to allow keeping the nodes active and not get into detention. 
3. We already had `config.NODE_IO_DETENTION_THRESHOLD` that could be tweaked separately to shorten or zero the time that detention is applied, before it gets reset, but this still allows the node to enter detention and disconnect
4. Cleaned 3 unused configs - LONG_GONE_THRESHOLD, SHORT_GONE_THRESHOLD, LONG_BUILD_THRESHOLD.
5. Moved local constants from nodes_monitor to config, just in case we want to override it.

1. NA

### Testing Instructions:
1. NA


dannyzaken (approved):

liranmauda (approved):
LGTM

PR #7384, opened 2023-07-06
add semaphore state to noobaa endpoint metrics

naveenpaul1:
### Explain the changes
1. This code change will add both namespace store and NSFS semaphore state to noobaa endpoint metrics `s3-service-monitor`

1. semaphore metrics will give more understanding of semaphore related issues

### Testing Instructions:
1. read/write data to either namespace store or NSFS S3 bucket.
2. verify Prometheus metrics, noobaa endpoint metrics should have the following metrics 
`semaphore_value,
semaphore_waiting_value,
semaphore_waiting_time,
semaphore_waiting_queue`
image

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -121,6 +121,11 @@ config.ENDPOINT_MONITOR_INTERVAL = 10 * 60 * 1000; // 10min
 // Keep connection on long requests
 config.S3_KEEP_ALIVE_WHITESPACE_INTERVAL = 15 * 1000;
 config.S3_MD_SIZE_LIMIT = 2 * 1024;
+// Semaphore monitoring execution interval
+config.SEMAPHORE_MONITOR_DELAY = 10 * 1000;
+// Semaphore metrics average calculation intervals in minutes, values need to be in ascending order
+config.SEMAPHORE_METRICS_AVERAGE_INTERVALS = Object.freeze(['1', '5', '10']);
+config.ENABLE_SEMAPHORE_MONITOR = true
 
 config.ENDPOINT_HTTP_SERVER_REQUEST_TIMEOUT = 300 * 1000;
 config.ENDPOINT_HTTP_SERVER_KEEPALIVE_TIMEOUT = 5 * 1000;
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -40,6 +40,7 @@ const prom_reporting = require('../server/analytic_services/prometheus_reporting
 const background_scheduler = require('../util/background_scheduler').get_instance();
 const endpoint_stats_collector = require('../sdk/endpoint_stats_collector');
 const { NamespaceMonitor } = require('../server/bg_services/namespace_monitor');
+const { SemaphoreMonitor } = require('../server/bg_services/semaphore_monitor');
 
 if (process.env.NOOBAA_LOG_LEVEL) {
     const dbg_conf = debug_config.get_debug_config(process.env.NOOBAA_LOG_LEVEL);
@@ -107,6 +108,7 @@ async function main(options = {}) {
         process.on('warning', e => dbg.warn(e.stack));
 
         let internal_rpc_client;
+        let object_io;
 
         let init_request_sdk = options.init_request_sdk;
         if (!init_request_sdk) {
@@ -131,8 +133,7 @@ async function main(options = {}) {
                 const n2n_agent = rpc.register_n2n_agent(((...args) => signal_client.node.n2n_signal(...args)));
                 n2n_agent.set_any_rpc_address();
             }
-
-            const object_io = new ObjectIO(location_info);
+            object_io = new ObjectIO(location_info);
 
             internal_rpc_client = rpc.new_client({
                 auth_token: await get_auth_token(process.env)
@@ -181,9 +182,16 @@ async function main(options = {}) {
                 should_monitor: nsr => Boolean(nsr.nsfs_config),
             }));
         }
+
+        if (config.ENABLE_SEMAPHORE_MONITOR) {
+            background_scheduler.register_bg_worker(new SemaphoreMonitor({
+                name: 'semaphore_monitor',
+                object_io: object_io,
+            }));
+        }
+
         // Start a monitor to send periodic endpoint reports about endpoint usage.
         start_monitor(internal_rpc_client, endpoint_group_id);
-
     } catch (err) {
         handle_server_error(err);
     }
--- a/src/sdk/endpoint_stats_collector.js
+++ b/src/sdk/endpoint_stats_collector.js
@@ -7,6 +7,7 @@ const mime = require('mime');
 const dbg = require('../util/debug_module')(__filename);
 const prom_report = require('../server/analytic_services/prometheus_reporting');
 const DelayedCollector = require('../util/delayed_collector');
+const config = require('../../config');
 
 /**
  * @typedef {{
@@ -83,6 +84,10 @@ class EndpointStatsCollector {
         this.update_fs_stats = fs_worker_stats => this._update_fs_stats(fs_worker_stats);
 
         this.prom_metrics_report = prom_report.get_endpoint_report();
+        this.semaphore_reports = {
+            object_io: [],
+            nsfs: [],
+        };
     }
 
     static instance() {
@@ -237,6 +242,35 @@ class EndpointStatsCollector {
         this.endpoint_stats_collector.update({
             bucket_counters: { [bucket_name]: { [content_type]: { write_count: 1 } } }
         });
+    }
+
+    update_semaphore_state(report, type, report_sample_sizes) {
+        if (!this.semaphore_reports[type]) {
+            throw new Error(`Semaphore report type ${type} is not valid`);
+        }
+        this.semaphore_reports[type].push(report);
+        if (this.semaphore_reports[type].length > report_sample_sizes[report_sample_sizes.length - 1]) {
+            this.semaphore_reports[type].shift();
+        }
+        const average_intervals = [];
+        for (const sample_size of report_sample_sizes) {
+            if (this.semaphore_reports[type].length < sample_size) {
+                return;
+            }
+            // Convert the sampling size to sampling average interval
+            const average_interval = (sample_size * config.SEMAPHORE_MONITOR_DELAY) / (60 * 1000);
+            average_intervals.push(average_interval);
+            const average_values = this.semaphore_reports[type].slice(this.semaphore_reports[type].length - sample_size);
+            this.update_semaphore_reports(average_values, { type, average_interval }, average_intervals);
+        }
+
+    }
+
+    update_semaphore_reports(reports, labels, average_intervals) {
+        this.prom_metrics_report.set('semaphore_value', labels, reports, average_intervals);
+        this.prom_metrics_report.set('semaphore_waiting_value', labels, reports, average_intervals);
+        this.prom_metrics_report.set('semaphore_waiting_time', labels, reports, average_intervals);
+        this.prom_metrics_report.set('semaphore_waiting_queue', labels, reports, average_intervals);
     }
 
     update_nsfs_read_stats({ namespace_resource_id, bucket_name = undefined, size = 0, count = 0, is_err = false }) {
--- a/src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
+++ b/src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
@@ -10,6 +10,7 @@ const js_utils = require('../../../util/js_utils');
 // come next to add endpoint metrics reporting.
 // -----------------------------------------
 
+const nsfs_suffix = "_nsfs";
 const NOOBAA_ENDPOINT_METRICS = js_utils.deep_freeze([{
         type: 'Counter',
         name: 'hub_read_bytes',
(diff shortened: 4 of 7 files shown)


src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
```diff
@@ -161,6 +161,38 @@ const NOOBAA_ENDPOINT_METRICS = js_utils.deep_freeze([{
                 10000, 20000, 50000,
             ],
         }
+    },
+    {
+        type: 'Gauge',
+        name: 'semaphore_value',
+        configuration: {
+            help: 'Namespace semaphore value',
+            labelNames: ['bucket_name', 'type']
```
dannyzaken, line 170: does it mean we store it per bucket? why?

dannyzaken (commented):
a few general comments:
1. is there a reason to share the same metric name for `object_io` and `namespace_fs` semaphores? We can have both NSFS and noobaa bucket workloads in the same system. @nimrod-becker I remember something about reducing the number of metrics we send to Prometheus. Is this something to consider? 
2. the metrics contain only one sample of the semaphore values taken only when there is IO. So once IO stops, the value will not be updated, am I right? 
3. I think it can be more consistent with sampling the semaphores values in constant intervals (a function running in the background) and providing an average (e.g., for the last minute).

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
naveenpaul1: Hi @dannyzaken,
1. Even if `object_io` and `namespace_fs` have the same metrics name, It has a label `type` which will have the value `ns` or `ns_fs` and we can use these metrics to identify the type of metrics.
2. Yes, sampling will happen only when there is an IO. Once IO is done and semaphore values go back to default values metrics will have that default semaphore value
3. Semaphore values will vary only when there is an IO and most of the time it will have the same value, and it will increase the metrics we push to Prometheus. We can always find the average in specific intervals by using Prometheus Query Language. I think pushing average value to metrics creates confusion at the graph level.
Please share your thoughts

naveenpaul1: I added  label `bucket_name` for debugging purposes, When there is an issue with one bucket using this label we can query only that bucket's metrics from Prometheus.

src/sdk/endpoint_stats_collector.js
```diff
@@ -239,6 +239,15 @@ class EndpointStatsCollector {
         });
     }
 
+    async update_semaphore_reports(repotrs){
```
liranmauda, line 242: why is this async? 
I dont see `setMultipleMetrics` return a promise.

src/sdk/endpoint_stats_collector.js
```diff
@@ -239,6 +239,15 @@ class EndpointStatsCollector {
...
+        var type = 'ns'
```
liranmauda, line 243: what is type ns use for? 
yes, it is a lable but for what reasone we define `ns`? 
and even if we should define the lable, can we use a better name? 
and avoid var, use let.

liranmauda (changes requested):

config.js
```diff
@@ -121,6 +121,7 @@ config.ENDPOINT_MONITOR_INTERVAL = 10 * 60 * 1000; // 10min
 // Keep connection on long requests
 config.S3_KEEP_ALIVE_WHITESPACE_INTERVAL = 15 * 1000;
 config.S3_MD_SIZE_LIMIT = 2 * 1024;
+config.SEMAPHORE_MONITOR_INTERVAL = 10 * 1000;
```
dannyzaken, line 124: As discussed, add a config for the period we want an average for (e.g. 5 min).
check if you can implement with an array of intervals, so we can get metrics for multiple averages (e.g., 1 min, 5 min, 15 min). In that case, test if `load_config_env_overrides` in this file works for a numbers array.

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
```diff
@@ -161,6 +162,38 @@ const NOOBAA_ENDPOINT_METRICS = js_utils.deep_freeze([{
                 10000, 20000, 50000,
             ],
         }
+    },
+    {
+        type: 'Gauge',
+        name: 'semaphore_value',
```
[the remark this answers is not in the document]
dannyzaken: @liranmauda notice the `'type'` labelName, which identify the semaphore

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
```diff
@@ -192,6 +225,49 @@ class NooBaaEndpointReport extends BasePrometheusReport {
...
+    set(name, labels, value) {
+        if (!this._metrics) return;
+        const metric = this._metrics[name];
+        if (!metric) throw new Error(`Unknown metric ${name}`);
+        if (metric.type !== 'Gauge') throw new Error(`Metric ${name} is not Gauge`);
+        metric.prom_instance.set(labels, value);
+    }
+
+    // setMultipleMetrics value metric
+    setMultipleMetrics(name, labels, values) {
```
dannyzaken, line 238: @naveenpaul1, as discussed, you can simplify this function by only maintaining the arrays of reports per label here and having a static collect method that averages the reports.

src/endpoint/endpoint.js
```diff
@@ -183,12 +185,47 @@ async function main(options = {}) {
...
         handle_server_error(err);
     }
 }
 
+
+/**
+ * @param {Semaphore} io_buffers_sem
+ * @param {import('../sdk/endpoint_stats_collector').EndpointStatsCollector} endpoint_stats_collector
+ */
+async function start_semaphore_monitor(io_buffers_sem, endpoint_stats_collector) {
```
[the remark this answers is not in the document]
dannyzaken: I agree. Better to use an existing infrastructure. 
maybe we can combine `start_semaphore_monitor` with [start_monitor](https://github.com/noobaa/noobaa-core/pull/7384/files#diff-4d81fd51d411b12e46ff80c1edadeb3cf1c0a5e267677106ce1ab602dd36fdbeR188) in the same bg worker. but not necessarily in this PR

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
```diff
@@ -192,6 +225,49 @@ class NooBaaEndpointReport extends BasePrometheusReport {
...
     }
 
+    // Set value metric
+    set(name, labels, value) {
+        if (!this._metrics) return;
+        const metric = this._metrics[name];
+        if (!metric) throw new Error(`Unknown metric ${name}`);
+        if (metric.type !== 'Gauge') throw new Error(`Metric ${name} is not Gauge`);
+        metric.prom_instance.set(labels, value);
+    }
```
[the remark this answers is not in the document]
naveenpaul1: @liranmauda This is a new method we added for setting  `Gauge` metrics type. Here we will set the metrics to Prometheus client if the type is Gauge. In this PR semaphore(not just for semaphore metrics, any Gauge type metrics) metrics will be set to Prometheus client.

config.js
naveenpaul1: @dannyzaken Once we get the average of multiple intervals how are we planning to push this value? In different metrics? same metrics with different labels?

src/endpoint/endpoint.js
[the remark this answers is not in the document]
dannyzaken: @naveenpaul1 let's change the code to use a bg worker for the semaphore monitor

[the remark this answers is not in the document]
naveenpaul1: done

src/endpoint/endpoint.js
```diff
@@ -107,6 +108,7 @@ async function main(options = {}) {
         process.on('warning', e => dbg.warn(e.stack));
 
         let internal_rpc_client;
+        const object_io = new ObjectIO(location_info);
```
dannyzaken, line 111: I wouldn't move the creation of `object_io`. you can declare it here and create it in the same location as before (under `if (!init_request_sdk)`). for the semaphore monitor, you can ignore the object_io if not defined

src/server/bg_services/semapore_monitor.js
```diff
@@ -0,0 +1,103 @@
...
+        } catch (err) {
+            dbg.error('semaphore_monitor:', err, err.stack);
+        }
+        return config.SEMAPHORE_MONITOR_DELAY;
+    }
+
+    async run_semaphore_monitor() {
+        if (!this.should_monitor) return;
+        try {
+            await this.start_semaphore_monitor(this.object_io);
```
dannyzaken, line 41: rename maybe to something like `sample_semphore_monitor`

src/server/bg_services/semapore_monitor.js
```diff
@@ -0,0 +1,103 @@
...
+                timestamp: Date.now(),
+                semaphore_state: {
+                    semaphore_cap: config.IO_SEMAPHORE_CAP,
+                    value: io_buffers._io_buffers_sem.value,
+                    waiting_value: io_buffers._io_buffers_sem.waiting_value,
+                    waiting_time: io_buffers._io_buffers_sem.waiting_time,
+                    waiting_queue: io_buffers._io_buffers_sem._wq.length,
+                }
+            };
+            await endpoint_stats_collector.update_semaphore_state(semaphore_report, "ns", this.report_bulk_sizes);
```
dannyzaken, line 76: This function shouldn't be `async`, and there is no need to `await` here, since `update_semaphore_state` is not async

src/server/bg_services/semapore_monitor.js
```diff
@@ -0,0 +1,103 @@
...
+                timestamp: Date.now(),
+                semaphore_state: {
+                    semaphore_cap: config.NSFS_BUF_POOL_MEM_LIMIT,
+                    value: buffers_pool_sem.value,
+                    waiting_value: buffers_pool_sem.waiting_value,
+                    waiting_time: buffers_pool_sem.waiting_time,
+                    waiting_queue: buffers_pool_sem._wq.length,
+                }
+            };
+            await endpoint_stats_collector.update_semaphore_state(semaphore_report, "nsfs", this.report_bulk_sizes);
```
dannyzaken, line 95: same as above

src/server/bg_services/semapore_monitor.js
```diff
@@ -0,0 +1,103 @@
...
+            dbg.error('semaphore_monitor:', err, err.stack);
+        }
+        return config.SEMAPHORE_MONITOR_DELAY;
+    }
+
+    async run_semaphore_monitor() {
+        if (!this.should_monitor) return;
+        try {
+            await this.start_semaphore_monitor(this.object_io);
+            await this.start_nsfs_semaphore_monitor();
```
dannyzaken, line 42: same as above

src/endpoint/endpoint.js
```diff
@@ -181,14 +181,33 @@ async function main(options = {}) {
                 should_monitor: nsr => Boolean(nsr.nsfs_config),
             }));
         }
+
+        const report_bulk_sizes = semaphore_report_bulk_sizes();
```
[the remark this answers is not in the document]
naveenpaul1: @dannyzaken  method will execute only once inside endpoint.js and avoid executing multiple times in BG worker, Since its load values only once when the endpoint pod initializes user has to restart the endpoint pod after updating `SEMAPHORE_METRICS_AVERAGE_INTERVALS` config. Should I move this method to BG worker?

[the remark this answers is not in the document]
dannyzaken: My comment was more on the location of the `semaphore_report_bulk_sizes` function. This code is more relevant inside `semaphore_monitor.js` than `endpoint.js`. You can move it to the constructor of `SemaphoreMonitor` and it will only be called once.
It's good to still accept `report_sample_sizes` as one of the constructor parameters so that you can override the config.js defaults  (e.g., for testing purposes).

src/sdk/endpoint_stats_collector.js
```diff
@@ -239,6 +242,37 @@ class EndpointStatsCollector {
         });
     }
 
+    update_semaphore_state(report, type, report_sample_sizes) {
+        if (type === 'object_io') {
+            this.ns_semaphore_reports.push(report);
+            if (this.ns_semaphore_reports.length > report_sample_sizes[report_sample_sizes.length - 1]) {
```
dannyzaken, line 248:
```suggestion
        const reports = this.semaphore_reports[type];
        if (!reports) {/*handle error*/}
        reports.push(report);
        /*...*/
```

src/server/analytic_services/prometheus_reports/noobaa_endpoint_report.js
```diff
@@ -192,6 +257,15 @@ class NooBaaEndpointReport extends BasePrometheusReport {
         metric.prom_instance.inc(labels, value);
     }
 
+    // Set value metric
+    set(name, labels, value) {
+        if (!this._metrics) return;
+        const metric = this._metrics[name];
+        if (!metric) throw new Error(`Unknown metric ${name}`);
+        if (metric.type !== 'Gauge') throw new Error(`Metric ${name} is not Gauge`);
+        if (metric.collect) metric.prom_instance.collect = metric.collect(metric.prom_instance, labels, value);
```
dannyzaken, line 266: looking at the example [here](https://github.com/siimon/prom-client#custom-metrics), the use of `collect` function seems incorrect.
IIUC, the collect function will be called when Prometheus scrapes the metrics, and this is when the average should be calculated (not every  10 seconds when a sample is taken). I just could not understand how labels are working together with a `collect` function

dannyzaken (approved):

liranmauda (approved):
Have some comments

PR #7386, opened 2023-07-09
Adding jest tests to the GH actions and the `npm run jest`

liranmauda:
### Explain the changes
Adding jest tests to the GH actions and the `npm run jest`


nadavMiz (approved):
LGTM

PR #7389, opened 2023-07-10
change bucket policy schema to fit AWS json

nadavMiz:
### Explain the changes
1. change bucket policy schema to match AWS schema, according to what we currently support. passing the json as is from the request
3. change all calling functions to use correct arguments.
4. add an upgrade script for the bucket policy
5. add unit tests for upgrade script 

1. gap - anyOf schema changes aren't dealt by decode_json / encode_json functions. since we only use string and sensitiveString in our case it works fine. issue: https://github.com/noobaa/noobaa-core/issues/7448

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -252,41 +252,65 @@ module.exports = {
             }
         },
 
+        bucket_policy_principal: {
+            anyOf: [{
+                    wrapper: SensitiveString,
+                }, {
+                    type: 'object',
+                    required: ['AWS'],
+                    properties: {
+                        AWS: {
+                            anyOf: [{
+                                    wrapper: SensitiveString
+                                }, {
+                                    type: 'array',
+                                    items: {
+                                        wrapper: SensitiveString
+                                    }
+                                }]
+                        }
+                    }
+                }
+            ]},
+        string_or_string_array: {
+            anyOf: [
+                {
+                    type: 'string'
+                },
+                {
+                    type: 'array',
+                    items: {
+                        type: 'string'
+                    }
+                }
+              ]
+        },
         bucket_policy: {
             type: 'object',
-            required: ['statement'],
+            required: ['Statement'],
             properties: {
-                version: { type: 'string' },
-                statement: {
+                Version: { type: 'string' },
+                Statement: {
                     type: 'array',
                     items: {
                         type: 'object',
-                        required: ['effect', 'action', 'principal', 'resource'],
+                        required: ['Effect', 'Action', 'Principal', 'Resource'],
                         properties: {
-                            sid: {
+                            Sid: {
                                 type: 'string'
                             },
-                            effect: {
-                                enum: ['allow', 'deny'],
+                            Effect: {
+                                enum: ['Allow', 'Deny'],
                                 type: 'string'
                             },
-                            action: {
-                                type: 'array',
-                                items: {
-                                    type: 'string'
-                                }
+                            Action: {
+                                $ref: '#/definitions/string_or_string_array'
                             },
-                            principal: {
-                                type: 'array',
-                                items: {
-                                    wrapper: SensitiveString,
-                                }
+                            Principal: {
+                                $ref: '#/definitions/bucket_policy_principal'
                             },
-                            resource: {
-                                type: 'array',
-                                items: {
-                                    type: 'string'
-                                }
+                            Resource: {
+                                $ref: '#/definitions/string_or_string_array'
                             }
                         }
                     }
--- a/src/endpoint/s3/ops/s3_put_bucket_policy.js
+++ b/src/endpoint/s3/ops/s3_put_bucket_policy.js
@@ -1,7 +1,6 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
-const _ = require('lodash');
 const S3Error = require('../s3_errors').S3Error;
 
 /**
@@ -10,21 +9,21 @@ const S3Error = require('../s3_errors').S3Error;
 async function put_bucket_policy(req) {
     let policy;
     try {
-        policy = JSON.parse(JSON.stringify(req.body).toLowerCase());
+        policy = JSON.parse(JSON.stringify(req.body));
     } catch (error) {
         console.error('put_bucket_policy: Invalid JSON provided', error);
         throw new S3Error(S3Error.InvalidArgument);
     }
-    // adapting bucket policy to our schema
-    for (const statement of policy.statement) {
-        if (statement.principal) {
-            if (statement.principal.aws) statement.principal = statement.principal.aws;
-            statement.principal = _.flatten([statement.principal]);
+    try {
+        await req.object_sdk.put_bucket_policy({ name: req.params.bucket, policy });
+    } catch (error) {
+        let err = error;
+        if (error.rpc_code === "INVALID_SCHEMA") {
+            err = new S3Error(S3Error.MalformedPolicy);
+            err.message = "Policy was not well formed or did not validate against the published schema";
         }
-        if (statement.action) statement.action = _.flatten([statement.action]);
-        if (statement.resource) statement.resource = _.flatten([statement.resource]);
+        throw err;
     }
-    await req.object_sdk.put_bucket_policy({ name: req.params.bucket, policy });
 }
 
 module.exports = {
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -364,7 +364,7 @@ function handle_error(req, res, err) {
         ((err instanceof S3Error) && err) ||
         new S3Error(S3Error.RPC_ERRORS_TO_S3[err.rpc_code] || S3Error.InternalError);
 
-    if (s3err.code === 'MalformedPolicy') {
+    if (!(err instanceof S3Error) && s3err.code === 'MalformedPolicy') {
         s3err.message = err.message;
         s3err.detail = err.rpc_data.detail;
     }
--- a/src/test/unit_tests/test_upgrade_scripts.js
+++ b/src/test/unit_tests/test_upgrade_scripts.js
@@ -1,0 +1,98 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+// setup coretest first to prepare the env
+const coretest = require('./coretest');
+const { rpc_client, EMAIL } = coretest;
+coretest.setup({ pools_to_create: [coretest.POOL_LIST[1]] });
+
+const AWS = require('aws-sdk');
+const http = require('http');
+const system_store = require('../../server/system_services/system_store').get_instance();
+const upgrade_bucket_policy = require('../../upgrade/upgrade_scripts/5.14.0/upgrade_bucket_policy');
+const dbg = require('../../util/debug_module')(__filename);
+const assert = require('assert');
+const mocha = require('mocha');
+
+const BKT = 'test-bucket';
+let s3;
+
+async function _clean_all_bucket_policies() {
+    for (const bucket of system_store.data.buckets) {
+        if (bucket.s3_policy) {
+            await s3.deleteBucketPolicy({Bucket: bucket.name.unwrap()}).promise();
+        }
+    }
+}
+
+mocha.describe('test upgrade scripts', async function() {
+    mocha.before(async function() {
+        this.timeout(600000); // eslint-disable-line no-invalid-this
+        await system_store.load();
+
+        const account_info = await rpc_client.account.read_account({ email: EMAIL, });
+        s3 = new AWS.S3({
+            endpoint: coretest.get_http_address(),
+            accessKeyId: account_info.access_keys[0].access_key.unwrap(),
+            secretAccessKey: account_info.access_keys[0].secret_key.unwrap(),
+            s3ForcePathStyle: true,
+            signatureVersion: 'v4',
+            computeChecksums: true,
+            s3DisableBodySigning: false,
+            region: 'us-east-1',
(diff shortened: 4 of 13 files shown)


src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -10,19 +10,18 @@ const S3Error = require('../s3_errors').S3Error;
...
         throw new S3Error(S3Error.InvalidArgument);
     }
-    // adapting bucket policy to our schema
-    for (const statement of policy.statement) {
-        if (statement.principal) {
-            if (statement.principal.aws) statement.principal = statement.principal.aws;
-            statement.principal = _.flatten([statement.principal]);
+    for (const statement of policy.Statement) {
+        if (statement.Principal) {
+            if (statement.Principal.AWS) statement.Principal = statement.Principal.AWS;
```
dannyzaken, line 20: Why do we still have this code to change the bucket policy? I thought the motivation for this change is that we just store the bucket policy as is.

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -10,19 +10,18 @@ const S3Error = require('../s3_errors').S3Error;
 async function put_bucket_policy(req) {
     let policy;
     try {
-        policy = JSON.parse(JSON.stringify(req.body).toLowerCase());
```
dannyzaken, line 13: IIUC, by removing this you make the bucket policy case matching sensitive. is this intended? what should be the correct behavior?

nadavMiz: yes, this is intended AWS bucket policies are case-sensitive. I have confirmed it by trying to push policies with correct and incorrect casing to AWS bucket policies

src/endpoint/s3/ops/s3_put_bucket_policy.js
nadavMiz: the reason I did it that way was because I didn't know schema can have more than one type for each field. I have changed the schema to support all types (object, array, string)

aspandey (commented):
Looks good to me. I observe that most of the changes are related to lower case to upper case. We need to make sure that this does not break other code wherever it might be using lowercase.

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -10,20 +9,11 @@ const S3Error = require('../s3_errors').S3Error;
...
-    // adapting bucket policy to our schema
-    for (const statement of policy.statement) {
-        if (statement.principal) {
-            if (statement.principal.aws) statement.principal = statement.principal.aws;
-            statement.principal = _.flatten([statement.principal]);
-        }
-        if (statement.action) statement.action = _.flatten([statement.action]);
-        if (statement.resource) statement.resource = _.flatten([statement.resource]);
-    }
     await req.object_sdk.put_bucket_policy({ name: req.params.bucket, policy });
```
dannyzaken, line 17: Since you pass the user's policy as is in the RPC, you should catch INVALID_SCHEMA errors and return it as `MalformedPolicy`. Other wise, if the user sends a malformed policy, they will get an internal error

src/upgrade/upgrade_scripts/5.3.0/update_bucket_owner_account.js
```diff
@@ -1,31 +0,0 @@
-/* Copyright (C) 2016 NooBaa */
-"use strict";
-
-const _ = require('lodash');
-
```
romayalon, line 5: what's the motivation behind removing old upgrade scripts??

src/api/common_api.js
```diff
@@ -252,41 +252,66 @@ module.exports = {
             }
         },
 
+        bucket_policy_principal: {
+            anyOf: [{
```
romayalon, line 256: I recall we didn't support anyOf/oneOf in postgresClient encode_json/decode_json, @dannyzaken is this valid now?

src/api/common_api.js
```diff
@@ -252,41 +252,66 @@ module.exports = {
...
+                    type: 'string',
+                    enum: ['*']
+                }, {
```
romayalon, line 259: writing it here but it's not related to this file, updating nsfs tool is needed since we use bucket policy there and it's breaking the tool

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -1,30 +1,25 @@
...
-        if (statement.principal) {
-            if (statement.principal.aws) statement.principal = statement.principal.aws;
-            statement.principal = _.flatten([statement.principal]);
-        }
-        if (statement.action) statement.action = _.flatten([statement.action]);
-        if (statement.resource) statement.resource = _.flatten([statement.resource]);
+    try {
+        await req.object_sdk.put_bucket_policy({ name: req.params.bucket, policy });
+    } catch (error) {
+        throw new RpcError('MALFORMED_POLICY', "Policy was not well formed or did not validate against the published schema", { detail: error.rpc_data });
```
romayalon, line 21: 2 comments - 
1. Why throw MALFORMED_POLICY for every error that occurs?
2. Throwing S3 error type instead of an RPC error type is expected in these s3_.. classes

src/upgrade/upgrade_scripts/5.14.0/upgrade_bucket_policy.js
```diff
@@ -0,0 +1,53 @@
...
+"use strict";
+
+const util = require('util');
+const _ = require('lodash');
+const { OP_NAME_TO_ACTION } = require('../../../endpoint/s3/s3_utils');
+
+async function run({ dbg, system_store, system_server }) {
+
+    try {
+        dbg.log0('Starting bucket policy...');
```
romayalon, line 11:
```suggestion
        dbg.log0('Starting bucket policy upgrade...');
```

src/upgrade/upgrade_scripts/5.14.0/upgrade_bucket_policy.js
```diff
@@ -0,0 +1,53 @@
...
+        const buckets = [];
+        const all_op_names = _.compact(_.flatMap(OP_NAME_TO_ACTION, action => [action.regular, action.versioned]));
+        for (const bucket of system_store.data.buckets) {
+            //Do not update if there are no bucket policy.
+            if (!bucket.s3_policy) return;
```
romayalon, line 16: why return? you mean continue?

src/upgrade/upgrade_scripts/5.14.0/upgrade_bucket_policy.js
```diff
@@ -0,0 +1,53 @@
...
+
+            const new_policy = {};
+            if (bucket.s3_policy.version) new_policy.Version = bucket.s3_policy.version;
+
+            new_policy.Statement = bucket.s3_policy.statement.map(statement => ({
+                Effect: statement.effect === 'allow' ? 'Allow' : 'Deny',
+                Action: statement.action.map(atatement_action =>
```
romayalon, line 23:
```suggestion
                Action: statement.action.map(statement_action =>
```

src/upgrade/upgrade_scripts/5.14.0/upgrade_bucket_policy.js
```diff
@@ -0,0 +1,53 @@
...
+                    all_op_names.find(action => action.toLowerCase() === atatement_action)),
```
romayalon, line 24: Instead of creating an array and iterate the whole array for each action, create a helper map that looks like the following: 
{ lowercasefunc: CamelCaseFuncName }

src/upgrade/upgrade_scripts/5.14.0/upgrade_bucket_policy.js
```diff
@@ -0,0 +1,53 @@
...
+            });
+        }
+
+        if (buckets.length > 0) {
+            dbg.log0(`Replacing bucket policy rules to the new API structure for buckets: ${buckets.map(bucket => util.inspect(bucket)).join(', ')}`);
+            await system_store.make_changes({ update: { buckets } });
+        } else {
+            dbg.log0('Upgrading buckets policy: no upgrade needed...');
+        }
+
```
romayalon, line 42: we wanted to start creating tests for upgrades, I think it's a great start for creating upgrade tests
(I guess no need to create a real upgrade flow, just put in the db old bucket policies, call run() function of this update script, and then validate that the bucket policies were updated correctly)

src/upgrade/upgrade_scripts/5.3.0/update_bucket_owner_account.js
liranmauda: I agree with @romayalon 
I don't think we should remove those. 
although they are old, they are usually used as a reference to writing new ones. 
@nadavMiz  Please revert the removal of all the upgrade scripts.

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -10,21 +9,22 @@ const S3Error = require('../s3_errors').S3Error;
...
-    for (const statement of policy.statement) {
-        if (statement.principal) {
-            if (statement.principal.aws) statement.principal = statement.principal.aws;
-            statement.principal = _.flatten([statement.principal]);
+    try {
+        await req.object_sdk.put_bucket_policy({ name: req.params.bucket, policy });
+    } catch (error) {
+        if (error.rpc_code === "INVALID_SCHEMA") {
+            const err = new S3Error(S3Error.MalformedPolicy);
+            err.message = "Policy was not well formed or did not validate against the published schema";
```
romayalon, line 22: why do we need to override the error message?

src/endpoint/s3/ops/s3_put_bucket_policy.js
```diff
@@ -10,21 +9,22 @@ const S3Error = require('../s3_errors').S3Error;
...
+            throw err;
+        } else {
```
romayalon, line 24: if the malformed condition is needed, you can remove the else and the first throw err and just throw err after the condition

src/endpoint/s3/s3_rest.js
```diff
@@ -364,7 +364,7 @@ function handle_error(req, res, err) {
         ((err instanceof S3Error) && err) ||
         new S3Error(S3Error.RPC_ERRORS_TO_S3[err.rpc_code] || S3Error.InternalError);
 
-    if (s3err.code === 'MalformedPolicy') {
+    if (!(err instanceof S3Error) && s3err.code === 'MalformedPolicy') {
```
romayalon, line 367: is it needed?

src/test/unit_tests/test_upgrade_scripts.js
```diff
@@ -0,0 +1,98 @@
...
+            signatureVersion: 'v4',
+            computeChecksums: true,
+            s3DisableBodySigning: false,
+            region: 'us-east-1',
+            httpOptions: { agent: new http.Agent({ keepAlive: false }) },
+        });
+        await s3.createBucket({ Bucket: BKT }).promise();
+    });
+
+    mocha.it('test upgrade bucket policy', async function() {
```
romayalon, line 48: worth adding the version number to which we upgrade

src/upgrade/upgrade_scripts/5.14.0/upgrade_bucket_policy.js
```diff
@@ -0,0 +1,53 @@
...
+const util = require('util');
+const _ = require('lodash');
+const { OP_NAME_TO_ACTION } = require('../../../endpoint/s3/s3_utils');
+
+async function run({ dbg, system_store, system_server }) {
+
+    try {
+        dbg.log0('Starting bucket policy upgrade...');
+        const buckets = [];
+        const all_op_names = new Map(_.flatMap(OP_NAME_TO_ACTION, action =>
```
romayalon, line 13: Building the Map/Object using a for loop will iterate OP_NAME_TO_ACTION only once instead of 3 times currently

src/endpoint/s3/ops/s3_put_bucket_policy.js
nadavMiz: unlike other s3 errors, the MalformedPolicy message is different depending on what failed. we default it to "Invalid principal in policy" which might not be the right error. It looks like the idea was that the message would be overwritten by the RPC error message (that is why I used RPC error originally). the reason I use this message and not a more specific description is that INVALID_SCHEMA doesn't return why it failed. and changing INVALID_SCHEMA seems out of scope for this assignment

src/endpoint/s3/s3_rest.js
nadavMiz: yes, otherwise it will fail for s3 errors because rpc_data is undefined. seems like the original intention was that MalformedPolicy will always be thrown as a RpcError and that the message and detail will be overwritten by its values

src/api/common_api.js
romayalon: @dannyzaken ?

romayalon: As we discussed, this anyOf doesn't really gets encoded, but since all we have inside the any of is string/sensitive string it works. 
@dannyzaken Are you ok with merging it and opening a gap for the anyOf issue?

nadavMiz: opened an issue for the anyOf: https://github.com/noobaa/noobaa-core/issues/7448

dannyzaken: I'm fine with the gap.

romayalon (approved):

PR #7392, opened 2023-07-11
Makefile target to build linux executable in build container

guymguym:
### Explain the changes
1. Added a `make executable` target to build the noobaa-core single-executable (with pkg for now) for linux x86_64 inside a build container.

1. NA

### Testing Instructions:
1. See standalone.md - tldr - `make executable`

- [x] Doc added/updated

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,28 +1,29 @@
+BUILDER_TAG?="noobaa-builder"
+NOOBAA_BASE_TAG?="noobaa-base"
+NOOBAA_TAG?="noobaa"
+TESTER_TAG?="noobaa-tester"
+POSTGRES_IMAGE?="centos/postgresql-12-centos7"
+MONGO_IMAGE?="centos/mongodb-36-centos7"
+
 CONTAINER_ENGINE?=$(shell docker version >/dev/null 2>&1 && echo docker)
 ifeq ($(CONTAINER_ENGINE),)
 	CONTAINER_ENGINE=$(shell podman version >/dev/null 2>&1 && echo podman)
 endif
+ifeq ($(CONTAINER_ENGINE),)
+	CONTAINER_ENGINE=$(shell lima nerdctl version >/dev/null 2>&1 && echo lima nerdctl)
+endif
 
 # see https://github.com/containerd/nerdctl/blob/main/docs/multi-platform.md
 # e.g use CONTAINER_PLATFORM=amd64 for building x86_64 on arm.
+CONTAINER_PLATFORM?=$(shell [ "`arch`" = "arm64" ] || [ "`arch`" = "aarch64" ] && echo amd64)
 CONTAINER_PLATFORM_FLAG=
-ifdef CONTAINER_PLATFORM
+ifneq ($(CONTAINER_PLATFORM),)
 	CONTAINER_PLATFORM_FLAG="--platform=$(CONTAINER_PLATFORM)"
 endif
 
-
 GIT_COMMIT?="$(shell git rev-parse HEAD | head -c 7)"
 NAME_POSTFIX?="$(shell ${CONTAINER_ENGINE} ps -a | wc -l | xargs)"
-BUILDER_TAG?="noobaa-builder"
-TESTER_TAG?="noobaa-tester"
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -1,28 +1,29 @@
...
 ifeq ($(CONTAINER_ENGINE),)
 	CONTAINER_ENGINE=$(shell podman version >/dev/null 2>&1 && echo podman)
 endif
+ifeq ($(CONTAINER_ENGINE),)
+	CONTAINER_ENGINE=$(shell lima nerdctl version >/dev/null 2>&1 && echo lima nerdctl)
+endif
 
 # see https://github.com/containerd/nerdctl/blob/main/docs/multi-platform.md
 # e.g use CONTAINER_PLATFORM=amd64 for building x86_64 on arm.
+CONTAINER_PLATFORM?="amd64"
```
liranmauda, line 18: On my old Mac, the x86 one, `CONTAINER_PLATFORM?="amd64"` lead to:
```sh
$ make executable

## Build image noobaa-builder ...
docker build "--platform="amd64""  -f src/deploy/NVA_build/builder.Dockerfile   -t noobaa-builder .
[+] Building 0.8s (3/3) FINISHED
 => [internal] load .dockerignore                                                                                                                                                                                                      0.0s
 => => transferring context: 152B                                                                                                                                                                                                      0.0s
 => [internal] load build definition from builder.Dockerfile                                                                                                                                                                           0.0s
 => => transferring dockerfile: 2.18kB                                                                                                                                                                                                 0.0s
 => ERROR [internal] load metadata for quay.io/centos/centos:stream8                                                                                                                                                                   0.7s
------
 > [internal] load metadata for quay.io/centos/centos:stream8:
------
builder.Dockerfile:1
--------------------
   1 | >>> FROM quay.io/centos/centos:stream8
   2 |     LABEL maintainer="Liran Mauda (lmauda@redhat.com)"
   3 |
--------------------
ERROR: failed to solve: quay.io/centos/centos:stream8: no match for platform in manifest sha256:76dcf5724accefc182f93ef58c849a69fdbf937f9e94e7ab2230ab972d066d41: not found
make: *** [builder] Error 1
```
Commenting it out make it work. 
Do we want the default to be `"amd64"` ?

guymguym: Thanks for testing this, I really didn't know if this would also work out on real x86 arch.
Perhaps I should just check the current arch first and only set to x86 (amd64) if the current arch is arm...
WDYT?

guymguym: @liranmauda I tried to add a fix - can you try to run again?

liranmauda: Sound Great.
Now it works for me
```sh
$ make executable

## Build image noobaa-builder ...
docker build   -f src/deploy/NVA_build/builder.Dockerfile   -t noobaa-builder .
[+] Building 0.3s (16/16) FINISHED
 => [internal] load .dockerignore                                                                                                                                                                                                      0.0s
 => => transferring context: 152B                                                                                                                                                                                                      0.0s
 => [internal] load build definition from builder.Dockerfile                                                                                                                                                                           0.1s
 => => transferring dockerfile: 2.18kB                                                                                                                                                                                                 0.0s
 => [internal] load metadata for quay.io/centos/centos:stream8                                                                                                                                                                         0.0s
 => [ 1/11] FROM quay.io/centos/centos:stream8
```

src/deploy/standalone/export.Dockerfile
```diff
@@ -0,0 +1,11 @@
+# This dockerfile exports just the single file executable from the built image.
```
liranmauda, line 1: Do we need 2 Dockerfiles? 
if this is intended to be built using the makefile we can write it as an additional stage in `executable.Dockerfile` and in the make file use a target for building.

guymguym: I had this in the same dockerfile when I started, the problem is that when using docker build --output it seems to not save the cache because it only outputs the content to a directory, and I wanted to have a proper cache. So with an intermediate step that has a named image (`noobaa-core-executable`) the cache is saved and only the export step uses the --output flag.

liranmauda (approved):
LGTM

PR #7394, opened 2023-07-11
Support Azure log credentials in `update_external_connection`

Neon-White:
### Explain the changes
1. Added support for the optional Azure log credentials in `update_external_connection`
2. Fixed a bug in the Azure log iterator

### DIFF
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -772,11 +772,10 @@ async function add_external_connection(req) {
 
 async function update_external_connection(req) {
     const connection = cloud_utils.find_cloud_connection(req.account, req.rpc_params.name);
-    const {
-        name,
-        identity = connection.access_key,
-        secret
-    } = req.rpc_params;
+    const name = req.rpc_params.name;
+    const identity = req.rpc_params.identity || connection.access_key;
+    const secret = req.rpc_params.secret;
+    const azure_log_access_keys = req.rpc_params.azure_log_access_keys;
 
     const encrypted_secret = system_store.master_key_manager.encrypt_sensitive_string_with_master_key_id(
         secret, req.account.master_key_id._id);
@@ -791,6 +790,7 @@ async function update_external_connection(req) {
             endpoint: connection.endpoint,
             cp_code: connection.cp_code,
             auth_method: connection.auth_method,
+            azure_log_access_keys: azure_log_access_keys,
         });
         check_failed = status !== 'SUCCESS';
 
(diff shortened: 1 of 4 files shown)


src/server/system_services/account_server.js
```diff
@@ -793,6 +793,7 @@ async function update_external_connection(req) {
             endpoint: connection.endpoint,
             cp_code: connection.cp_code,
             auth_method: connection.auth_method,
+            azure_log_access_keys: connection.azure_log_access_keys,
```
romayalon: I'll move the [old PR discussion](https://github.com/noobaa/noobaa-core/pull/7308#discussion_r1258238501) to here, I def agree with not printing secrets (get_account_info()), but we do return identity so we also need to return the ids from the azure_log_access_keys object

```
azure_log_access_keys: {
            properties: {
                azure_tenant_id -> should be printed to the user
                azure_client_id -> should be printed to the user
                azure_client_secret ,
                azure_logs_analytics_workspace_id  -> should be printed to the user
            }
 }
```

what's the point of letting the user update these properties without him knowing what he put there in the first place

dannyzaken: @Neon-White do we want to treat all of these the same as secret_key, or does any of these fields should be returned to identify the credentials? 
e.g., for azure connections, we use the Account Name as the "identity" (`AWS_ACCESS_KEY_ID` for S3 connections). should any of these new fields be returned as part of the account information?

src/server/system_services/account_server.js
```diff
@@ -791,6 +790,7 @@ async function update_external_connection(req) {
             endpoint: connection.endpoint,
             cp_code: connection.cp_code,
             auth_method: connection.auth_method,
+            azure_log_access_keys: azure_log_access_keys || undefined,
```
romayalon, line 793: `|| undefined` is redundant, no?

src/server/system_services/account_server.js
```diff
@@ -803,15 +803,34 @@ async function update_external_connection(req) {
         throw new RpcError('INVALID_CREDENTIALS', `Credentials are not valid ${name}`);
     }
 
+    let set_obj;
+    if (azure_log_access_keys?.azure_tenant_id?.unwrap()) {
```
romayalon, line 807: why checking for azure_tenant_id existence and not for azure_log_access_keys?

romayalon, line 807: having Azure log access keys is a very small scenario and I think that the code flow should be:
1. create set_obj // -> main flow first
2. if (azure_log_access_keys) set the azure log fields on set_obj // -> other flows

romayalon: try adding +optional to the operator code

dannyzaken: I think that here we should validate the values of `azure_log_access_keys`. we already have the [schema](https://github.com/noobaa/noobaa-core/blob/845073892c4768ec6e0edf593b0baadd20686a85/src/api/common_api.js#L754C1-L754C1) that enforces all the required fields. if these fields are invalid we should throw an error.
in the operator we should make sure we call this API correctly

romayalon (approved):

dannyzaken (approved):

PR #7395, opened 2023-07-11
Migrating to AWS SDK V3 | Add Internal Class to Decide Which AWS SDK Version to Use

shirady:
### Explain the changes
1. Adding a new supportive layer for creating an AWS S3 Client using AWS SDK V2 and AWS SDK V3 this layer will return the S3 Client base on our decision, for example: if the signature version is 'v2' then AWS SDK V2, else AWS SDK V3.
 
#### Background:
when using AWS SDK V2 we encounter the warning about AWS SDK V2 entering maintenance mode:

> We are formalizing our plans to enter AWS SDK for JavaScript (v2) into maintenance mode in 2023.
> Please migrate your code to use AWS SDK for JavaScript (v3).
> For more information, check the migration guide at https://a.co/7PzMCcy

The API is not backward compatible with the old version, so we need to change our code.  
One of the changes is related to `signatureVersion`, by using the AWS SDK V3 we can only use `signatureVersion` `v4`.
> v2: The signature version to sign requests with (overriding the API configuration).
> v3: Deprecated. Signature V2 supported in v2 SDK has been deprecated by AWS. v3 only supports signature v4.

Since the server might only support `signatureVersion` `v2`, we still need to support it in our s3-client generated from the AWS SDK.

1. None

### Testing Instructions:
set `AWS_SDK_VERSION_3_DISABLED` to false.

 **AWS Server**
1. Create a new `js` file in the repo with all needed parameters to create a new instance of s3 client (using aws sdk v3 params), for example, access key, secret access key, endpoint ('[https://s3.amazonaws.com')](https://s3.amazonaws.com%27%29/), signature version ('v2'), and region.
2. Run operations that were implemented in the class, for example: `listObjectsV2`, and print the result.
3. Repeat steps 1-2 with signature version `v4`.

**To run jest tests:**

- [ ] Doc added/updated
- [X] Tests added - add a couple of simple tests with jest framework (checks the decision on which SDK version to use, checks automated replacement of params).

### DIFF
--- a/src/sdk/noobaa_s3_client/noobaa_s3_client.js
+++ b/src/sdk/noobaa_s3_client/noobaa_s3_client.js
@@ -1,0 +1,106 @@
+/* Copyright (C) 2023 NooBaa */
+'use strict';
+
+const http = require('http');
+const https = require('https');
+const { HttpProxyAgent } = require('http-proxy-agent');
+const { HttpsProxyAgent } = require('https-proxy-agent');
+const { S3ClientSDKV2 } = require('./noobaa_s3_client_sdkv2');
+const { S3 } = require("@aws-sdk/client-s3");
+const { NodeHttpHandler } = require("@aws-sdk/node-http-handler");
+const config = require('../../../config');
+const http_utils = require('../../util/http_utils');
+
+// The params are the AWS SDK V3 params.
+// params = a map of parameters that are passed to the constructor of S3 Object in order to to bind to every request
+function get_s3_client_v3_params(params) {
+    let s3_client;
+
+    if (should_use_sdk_v2(params)) {
+        change_s3_client_params_to_v2_structure(params);
+        s3_client = new S3ClientSDKV2(params);
+    } else {
+        s3_client = new S3(params);
+    }
+    return s3_client;
+}
+
+// Here we decide which version of the sdk should we use:
+// 1. by the flag AWS_SDK_VERSION_3_DISABLED (we use it as a workaround in case we want to use only AWS SDK V2).
+// 2. signatureVersion defined? if it is 'v2' (signature version v2 is only supported in AWS SDK V2)
+function should_use_sdk_v2(params) {
+    return config.AWS_SDK_VERSION_3_DISABLED || (params.signatureVersion && params.signatureVersion === 'v2');
+}
+
+// we will use this function in debugging to print the class type easily
+function get_sdk_class_str(s3) {
+    if (s3 instanceof S3ClientSDKV2) return 'S3 Client SDK V2';
+    return 'S3 Client SDK V3';
+}
+
+// This function gets map of params as they are written in AWS SDK V3
+// and change the map of params as they appear in AWS SDK V2.
+// The comments are based on AWS SDK V3 upgrading notes:
+// https://github.com/aws/aws-sdk-js-v3/blob/main/UPGRADING.md
+function change_s3_client_params_to_v2_structure(params) {
+    // v2: s3ForcePathStyle, v3: forcePathStyle
+    replace_field(params, 'forcePathStyle', 's3ForcePathStyle');
+
+    // v2: sslEnabled, v3: tls
+    replace_field(params, 'tls', 'sslEnabled');
+
+    // v2: s3BucketEndpoint, v3: bucketEndpoint.
+    // Note that when set to true, you specify the request endpoint in the Bucket request parameter,
+    // the original endpoint will be overwritten.
+    // Whereas in v2, the request endpoint in client constructor overwrites the Bucket request parameter.
+    replace_field(params, 'bucketEndpoint', 's3BucketEndpoint');
+
+    if (params.endpoint && params.requestHandler) {
+        // v2: A set of options to pass to the low-level HTTP request.
+        // v3: These options are aggregated differently in v3. All requests use HTTPS by default.
+        // If you are passing custom endpoint which uses http, then you need to provide httpAgent
+        const agent = http_utils.get_agent_by_endpoint(params.endpoint);
+        params.httpOptions = {
+            agent
+        };
+        delete params.requestHandler;
+    }
+}
+
+// Replace in place a name of field in map params
+function replace_field(params, remove_field, add_field) {
+    if (params[remove_field] !== undefined) {
+        params[add_field] = params[remove_field];
+        delete params[remove_field];
+    }
+}
+
+// This function allows us to wrap the agent by NodeHttpHandler
+// The difference is between the property httpsAgent vs httpAgent (secure and unsecure)
+// This function was created based on the use of http_utils.get_unsecured_agent
+function get_requestHandler_with_suitable_agent(endpoint) {
+    const agent = http_utils.get_agent_by_endpoint(endpoint);
+    if (agent instanceof https.Agent || agent instanceof HttpsProxyAgent) {
+        return new NodeHttpHandler({
+            httpsAgent: agent
+        });
+    } else if (agent instanceof http.Agent || agent instanceof HttpProxyAgent) {
+        return new NodeHttpHandler({
+            httpAgent: agent
+        });
+    } else {
+        throw new Error(`Protocol of ${endpoint} doesn't have a matching agent (valid only for HTTP, HTTPS, WSS only)`);
+    }
+}
+
+// v2: code, v3: Code
+function check_error_code(err, code) {
+    return err.code === code || err.Code === code;
+}
+
+// EXPORTS
+exports.get_s3_client_v3_params = get_s3_client_v3_params;
+exports.change_s3_client_params_to_v2_structure = change_s3_client_params_to_v2_structure;
+exports.get_sdk_class_str = get_sdk_class_str;
+exports.check_error_code = check_error_code;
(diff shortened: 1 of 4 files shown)


shirady:
It's a draft PR:
* The plan is to separate the first 2 commits in a different PR and this PR would only contain the third commit.
* This PR implements our idea as discussed to have a layer before creating an S3 Client. Now this layer contains 4 operations, but it will expand as we will add all the needed operations.

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
```diff
@@ -0,0 +1,115 @@
...
+    get_credentials(params) {
+        const credentials = {
+            accessKeyId: params.accessKeyId instanceof SensitiveString ?
+                params.accessKeyId.unwrap() : params.accessKeyId,
+            secretAccessKey: params.secretAccessKey instanceof SensitiveString ?
+                params.secretAccessKey.unwrap() : params.secretAccessKey,
+        };
+        if (params.sessionToken) credentials.sessionToken = params.sessionToken;
+        return credentials;
+    }
```
liranmauda, line 87: Do we want it?

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
```diff
@@ -0,0 +1,115 @@
...
+
+class S3ClientSDKV3 {
+    constructor(params) {
+        this.params_sdk_v3 = this.get_s3_client_params_sdk_v3(params);
+    }
+
+    // this function gets map of params as they are written in AWS SDK V2
+    // and returns the map of params as they should appear in AWS SDK V3
+    get_s3_client_params_sdk_v3(params) {
+        const params_sdk_v3 = {};
```
liranmauda, line 26: Do we need new object? or do we want to change the params itself?

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
```diff
@@ -0,0 +1,115 @@
...
+                const requestHandler = this.get_requestHandler_with_suitable_agent(params.endpoint, agent);
+                params_sdk_v3.requestHandler = requestHandler;
+            }
+        }
+        if (params.region) {
+            params_sdk_v3.region = params.region;
+        } else {
+            // in AWS SDK V3 we must set region (else we have would have an error when executing an operation
+            // (Error: Region is missing)
+            const DEFAULT_REGION = 'us-east-1';
```
liranmauda, line 43: Should it move into the config?

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3rg.js
```diff
@@ -0,0 +1,50 @@
...
+ * S3 Interface, S3 Class and operations in AWS SDK V3:
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/S3/
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Class/S3/
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/
+ */
+
+// RG = Regular, internal terminology for creating S3 CLient as we used to it in AWS SDK V2
+// (using high-level API)
+
+class S3ClientSDKV3RG extends S3ClientSDKV3 {
```
liranmauda, line 19: not sure RG is catchy :)

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3rg.js
```diff
@@ -0,0 +1,50 @@
...
+    constructor(params) {
+        super(params);
+        this.s3 = new S3(this.params_sdk_v3);
+        // override the credentials after we have the client (to avoid printing of credentials)
+        this.params_sdk_v3.credentials = undefined;
```
liranmauda, line 24: Can we make it sensetive sting?

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3bm.js
dannyzaken, line 1: I think we can drop this file for now, especially if we don't use it.

We use a lot of commands already, so I don't see much value in requiring them individually. Am I missing something?

dannyzaken (commented):
There is a lot of unnecessary abstraction in this PR. 
According to `noobaa_s3_client_sdkv3rg.js`, it looks like the interface is almost unchanged (in V2 there is `.promise` and in v3 `Body` is returned as stream. 

can we instead, for V3, return the actual client (`new S3(...)`), and for only for V2 have a layer to do the necessary fixes? we still have to handle the conversion from stream to buffer in V3. we already have a util for that in `buffer_utils` (see [here](https://github.com/noobaa/noobaa-core/blob/6a2879fe6142c6df4041bb44188b9502fc61289f/src/util/buffer_utils.js#L102-L105))

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3bm.js
shirady: The file was dropped.

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -0,0 +1,238 @@
...
+const { S3 } = require("@aws-sdk/client-s3");
+const { NodeHttpHandler } = require("@aws-sdk/node-http-handler");
+const dbg = require('../../util/debug_module')(__filename);
+dbg.set_process_name('noobaa_s3_client_by_sdk');
+const config = require('../../../config');
+const http_utils = require('../../util/http_utils');
+
+// The params are the AWS SDK V2 params.
+// params = a map of parameters that are passed to the constructor of S3 Object in order to to bind to every request
+function get_s3_client_v2_params(params) {
```
dannyzaken, line 15: redundant. we can change the params to v3

shirady: Removed, as discussed we will change hard-coded the params to v3 structure.

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -0,0 +1,148 @@
...
+        if (use_sdk_v2_by_signature_version) return use_sdk_v2_by_signature_version;
+    }
+
+    if (params.maxRedirects) {
+        const use_sdk_v2_by_max_redirects = should_use_sdk_v2_by_max_redirects(params.maxRedirects);
+        return use_sdk_v2_by_max_redirects;
+    }
+
+    return false;
+}
```
dannyzaken, line 45: This function should be more straightforward. 
First, I don't think we should take `maxRedirects` as a factor. We use it only in 2 tests, and even then, we explicitly set it to 10, which is the default value. 
Also, there is no need to break each condition into its own function. It just makes the code very verbose. 

```suggestion
function should_use_sdk_v2(params) {
    return config.AWS_SDK_VERSION_3_DISABLED || params?.signatureVersion
}
```

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -0,0 +1,148 @@
...
+// max redirects is only supported in AWS SDK V2
+// v2: The maximum amount of redirects to follow for a service request (Integer, default is 10).
+// v3: Deprecated. SDK does not follow redirects to avoid unintentional cross-region requests.
+function should_use_sdk_v2_by_max_redirects(max_redirects) {
+    return max_redirects >= 0;
+}
+
+// we will use this function in debugging to print the class type easily
+function get_sdk_class_str(s3) {
+    if (is_instance_is_S3ClientSDKV2(s3)) return 'S3 Client SDK V2';
```
dannyzaken, line 63: similar comment here. there is no need to wrap each condition in a function 

```suggestion
    if (s3 instanceof S3ClientSDKV2) return 'S3 Client SDK V2';
```

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -0,0 +1,148 @@
...
+    replace_filed(params, 'forcePathStyle', 's3ForcePathStyle');
+
+    // v2: s3DisableBodySigning, v3: applyChecksum
+    replace_filed(params, 'applyChecksum', 's3DisableBodySigning');
+
+    // v2: sslEnabled, v3: tls
+    replace_filed(params, 'tls', 'sslEnabled');
+
+    // v2: maxRetries, v3: maxAttempts.
+    if (params.maxAttempts > 0) {
```
dannyzaken, line 87: do we currently use `maxRetries` anywhere in the code? couldn't find it. since there are a lot of changes, we should only map the ones we currently use

src/sdk/noobaa_s3_client/noobaa_s3_client.js
dannyzaken: I missed the `=== 'v2'`

Regarding the condition, in general, in an `or` statement, if the first conditional is true, the second one will not be evaluated. For an `and` statement, if the first is `false` the second is never evaluated.

dannyzaken (approved):

PR #7396, opened 2023-07-12
Support multipart_upload for log based replication

aayushchouhan09:
### Explain the changes
1.  Added support for multipart_upload in log based replicaiton.


Neon-White (approved):

dannyzaken (changes requested):
The decision to perform a single upload vs. multipart should be made in the replication server, not the scanner.
The scanner should figure out what to do. The replication server should figure out how to do it. 
This will also better align with the batch PR that sends multiple keys to the replication server

achouhan09:
We will handle it later. 
cc @dannyzaken @liranmauda

PR #7407, opened 2023-07-23
Allow the bucket diff constructor to get s3_params OR connection

liranmauda:
### Explain the changes
- Allow the bucket diff constructor to get s3_params OR connection

### Testing Instructions:
1. run `npx jest test_bucket_diff.test.js`

- [x] Tests added

### DIFF
--- a/src/server/utils/bucket_diff.js
+++ b/src/server/utils/bucket_diff.js
@@ -13,17 +13,31 @@ class BucketDiff {
      *   first_bucket: string;
      *   second_bucket: string;
      *   version: boolean;
-     *   s3_params: AWS.S3.ClientConfiguration
+     *   s3_params?: AWS.S3.ClientConfiguration
+     *   connection?: AWS.S3
      *   for_replication: boolean
      * }} params
      */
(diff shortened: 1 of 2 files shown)


src/server/utils/bucket_diff.js
```diff
@@ -13,17 +13,31 @@ class BucketDiff {
...
+            connection,
+            for_replication,
+        } = params;
         this.first_bucket = first_bucket;
         this.second_bucket = second_bucket;
         this.version = version;
         //We will assume at this stage that first_bucket and second_bucket are accessible via the same s3
         // If we want to do a diff on none s3 we should use noobaa namespace buckets.
-        this.s3 = new AWS.S3(s3_params);
+        if (connection) {
```
romayalon, line 35: what is the motivation for allowing both connection/s3 params? 
It's inconsistent and I don't think we should support both ways

liranmauda: When we have noobaa connection, we want to use it.
for example, on the replication scanner.
When we don't have a connection, such as when we start it as a tool, we need to establish one, by using s3 params.

romayalon (approved):

PR #7413, opened 2023-07-31
Migrating to AWS SDK V3 | noobaa_s3_client without region + implement It in analyze resource

shirady:
### Explain the changes
1. Change the default and use AWS SDK V3 (config file).
2. Add get region and edit endpoint address (for AWS only).
3. Implement the use of AWS SDK V3: s3 params, region, and edited endpoint in analyze resource.
4. Style and typo fixes.

#### Background:
In AWS SDK v3 S3 client and bucket must have the same region, the concept of a global client doesn't exist anymore [(here)](https://github.com/aws/aws-sdk-js-v3/issues/1807). Without region, you'll get an error message `Region is missing` [(here)](https://github.com/aws/aws-sdk-js-v3/issues/1845), even if the endpoint address is not AWS.

1. Gap - we don't enforce AWS endpoint address to include the region today (`region` is not required).

### Testing Instructions:
**Analyze resource**
1. Based on the instructions [here](https://gist.github.com/dannyzaken/c85f0006e5d6582c8d8666cafce6ab76) - the steps of _‘Build images’_ and _‘Deploy noobaa’._
2. For namespacestore, Run:
`nb namespacestore create <aws-s3 or s3-compatible>  <namespace-store-name>`
* It creates the log files and then deletes the created job.
3. Cases:
* s3-compatible - Noobaa bucket (with noobaa system as a server, the endpoint was taken from the field `InternalDNS` under `S3 Addresses`).
* aws-s3 with flag region in  region that is not 'us-east-1'.
* aws-s3 without flag region on bucket in region that is not 'us-east-1'.

**To run jest tests:**
`npx jest test_string_utils.test.js`

- [ ] Doc added/updated
- [ X] Tests added: basic tests for new functions (`get_region`, `edit_global_aws_endpoint`, `index_of_end`)

### DIFF
--- a/config.js
+++ b/config.js
@@ -710,7 +710,7 @@ config.TIERING_TTL_MS = 30 * 60 * 1000; // 30 minutes
 // AWS SDK VERSION     //
 /////////////////////////
 
-config.AWS_SDK_VERSION_3_DISABLED = true;
+config.AWS_SDK_VERSION_3_DISABLED = false;
 config.DEFAULT_REGION = 'us-east-1';
 
 /////////////////////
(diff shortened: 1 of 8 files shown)


shirady:
Closing this PR:
- We didn't want to create an async command to find the region (the most recommended way is to pass the region) in case we don't have it we would extract and update the region on the first call.
- If we have an AWS endpoint, we will not pass it to the constructor SDK.
- implementations of using the SDK v3 will be in different PRs.

PR #7417, opened 2023-08-01
Replace minikube with standalone deployment for Ceph Tests

tangledbytes:
### Explain the changes
This PR replaces minikube with standalone deployment (inside a container). The expectations are:
1. Have more reliable runs.
2. Runs should take lesser time.

This is done by creating a `standalone_deploy.sh` script which can be used locally as well to spin up a fully functioning NooBaa.

### Testing Instructions:
1. Run `make test-cephs3` or `CONTAINER_PLATFORM=linux/arm64 make test-cephs3` if on mac.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -246,6 +246,17 @@ test-postgres: tester
 tests: test #alias for test
 .PHONY: tests
 
+test-cephs3: tester
+	@echo "\033[1;34mRunning tests with Postgres.\033[0m"
+	@$(call create_docker_network)
+	@$(call run_postgres)
+	@echo "\033[1;34mRunning tests\033[0m"
+	$(CONTAINER_ENGINE) run $(CPUSET) --network noobaa-net --name noobaa_$(GIT_COMMIT)_$(NAME_POSTFIX) --env "SUPPRESS_LOGS=$(SUPPRESS_LOGS)" --env "POSTGRES_HOST=coretest-postgres-$(GIT_COMMIT)-$(NAME_POSTFIX)" --env "POSTGRES_USER=noobaa" --env "DB_TYPE=postgres" --env "POSTGRES_DBNAME=coretest" -v $(PWD)/logs:/logs $(TESTER_TAG) "./src/test/system_tests/ceph_s3_tests/run_ceph_test_on_test_container.sh"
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -246,6 +246,17 @@ test-postgres: tester
 tests: test #alias for test
 .PHONY: tests
 
+test-cephs3: tester
+	@echo "\033[1;34mRunning tests with Postgres.\033[0m"
+	@$(call create_docker_network)
+	@$(call run_postgres)
+	@echo "\033[1;34mRunning tests\033[0m"
+	$(CONTAINER_ENGINE) run $(CPUSET) --network noobaa-net --name noobaa_$(GIT_COMMIT)_$(NAME_POSTFIX) --env "SUPPRESS_LOGS=$(SUPPRESS_LOGS)" --env "POSTGRES_HOST=coretest-postgres-$(GIT_COMMIT)-$(NAME_POSTFIX)" --env "POSTGRES_USER=noobaa" --env "DB_TYPE=postgres" --env "POSTGRES_DBNAME=coretest" -v $(PWD)/logs:/logs $(TESTER_TAG) "./src/test/system_tests/ceph_s3_tests/run_ceph_test_on_test_container.sh" || true
```
liranmauda, line 254: why is it `|| true` ? if that fails it will mark the passing tests as green.

liranmauda (commented):
LGTM

Makefile
utkarsh-pro: You are right. I forgot to remove this, I was running it locally and wanted to remove the containers after the test finishes regardless of the results. 

Removed ✅

utkarsh-pro:
> @utkarsh-pro
I think we should do that for all the jobs.

@liranmauda, I think we are using minikube only in sanity tests now, right? Or am I missing something?

liranmauda:
> > @utkarsh-pro
> > I think we should do that for all the jobs.
> 
> @liranmauda, I think we are using minikube only in sanity tests now, right? Or am I missing something?

I think that you are correct. It will be grat to change that also

liranmauda (approved):

PR #7419, opened 2023-08-01
Allow customizing keep-alive and request timeouts for endpoint servers

tangledbytes:
### Explain the changes
This PR adds 2 configurations to endpoints, the intention is to allow the users to configure these timeouts based on their environments. For example, in a case where a NooBaa deployment is behind a horribly excruciatingly slow VPN, in those cases the user might want to tweak these values. 

**NOTE:** This still requires cooperation from the AWS client. So it is important for the client to use `--cli-read-timeout` and `--cli-connect-timeout`.

### Testing Instructions:
1. Introduce network throttling.
2. Spin up NooBaa.
3. Run `aws s3api --endpoint <val> put-object ...` should run just fine provided the timeouts are correctly set.

- [ ] Doc added/updated
- [ ] Tests added


utkarsh-pro:
@guymguym I haven't added any keepAliveEnabled because I think `keep-alive: true` can only be set for http.Agent and not for servers.

guymguym (approved):

guymguym:
@utkarsh-pro perhaps add these configs to standalone.md too with recommended values for TMFS?

guymguym:
@utkarsh-pro Notice I rebased your branch ([utkarsh-pro:utkarsh-pro/temp/timout-work](https://github.com/utkarsh-pro/noobaa-core/tree/utkarsh-pro/temp/timout-work)) you will need to use "the force" to push/pull it to your clone

utkarsh-pro:
@guymguym, I have added 15 mins as request timeout (with 3 retries this should give us about 45 mins) and 1 minute as keep alive timeout for TMFS.

PR #7420, opened 2023-08-01
Fix Azure log iteration and credentials lookup

Neon-White:
### Explain the changes
1. Fix and improve the Azure log iteration logic
2. Add `azure_log_access_keys` to the namespace resource connection for simpler access

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2228108

### DIFF
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -803,21 +803,27 @@ async function update_external_connection(req) {
         throw new RpcError('INVALID_CREDENTIALS', `Credentials are not valid ${name}`);
     }
 
-    const set_obj = {
+    const acc_update_set_obj = {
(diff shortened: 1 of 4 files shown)


src/server/system_services/account_server.js
```diff
@@ -807,17 +807,23 @@ async function update_external_connection(req) {
         "sync_credentials_cache.$.access_key": identity,
         "sync_credentials_cache.$.secret_key": encrypted_secret,
     };
+    const map_obj = {
```
dannyzaken, line 810: `map_obj` is not a very meaningful name. rename to something like `ns_resource_updates` or something similar that describes the purpose of the object

dannyzaken (approved):

PR #7424, opened 2023-08-02
Fix space check on TMFS

tangledbytes:
### Explain the changes
This PR adds a custom free space checker to fix TMFS disk space issue. The implementation relies on using the data from `df` instead of internal calculated IO.

Potential issues:
1. `df` data unavailable - falls back to base class implementation, so should be "safe"
2. Stale `df` data - in worse case scenario this might lead to a partial write problem.

### Testing Instructions:
1. Full blown ceph test should be good enough to ensure no regression.

- [ ] Doc added/updated
- [ ] Tests added


src/agent/block_store_services/block_store_fs.js
```diff
@@ -88,6 +92,31 @@ class BlockStoreFs extends BlockStoreBase {
...
+            return super._check_write_space(data_length);
+        }
+
+        const required_space = data_length + (1024 * 1024); // require some spare space
+        const free_space = this.cached_df_data.free;
+        if (free_space < required_space) {
+            throw new RpcError('NO_BLOCK_STORE_SPACE', 'used space exceeded the total capacity of ' +
+                this.usage_limit + ' bytes');
+        }
+    }
```
guymguym, line 119: How about we make `_check_write_space` async and separate it from get_storage_info(), so that we can decide when to trigger reloading of the df info based on other conditions, instead of just relying on the get_storage_info() timing, something like this:

configs -
```js
config.BLOCK_STORE_FS_CACHED_DF_MAX_TIME = 30 * 1000; // 30 seconds
config.BLOCK_STORE_FS_CACHED_DF_MIN_TIME = 1 * 1000; // 1 seconds
config.BLOCK_STORE_FS_CACHED_DF_MIN_SPACE = 1 * 1024 * 1024 * 1024; // 1 GB
```

check -
```suggestion
    async _check_write_space(data_length) {
        const now = Date.now();
        if (!this.cached_df_data || 
            (now - this.cached_df_data.time >= config.BLOCK_STORE_FS_CACHED_DF_MAX_TIME) ||
            (now - this.cached_df_data.time >= config.BLOCK_STORE_FS_CACHED_DF_MIN_TIME && 
                this.cached_df_data.free <= config.BLOCK_STORE_FS_CACHED_DF_MIN_SPACE)) {
            const drive = await os_utils.get_drive_of_path(this.root_path);
            this.cached_df_data = { ...drive.storage, time: now };
        }

        ...
    }
```

guymguym (changes requested):

guymguym (approved):

PR #7433, opened 2023-08-06
Add Region to Connection Schemas + Use Region in Check Connection

shirady:
### Explain the changes
1. Add region to connection schemas.
2. Use region in `check_aws_connection`.

1. none

### Testing Instructions:
Need to build with operator change noobaa/noobaa-operator#1188.
2. Create namespacestore/backingstore and pass the region explicitly:
`nb namespacestore create aws-s3  <namespace-store-name> --region=<target-bucket-region>`.
`nb backingstore create aws-s3  <backingstore-store-name> --region=<target-bucket-region>`.
3. Add printings to `check_external_connection`, `check_aws_connection` and see that the region passed.
Note: test on buckets in the regions 'us-east-1' and 'us-west-1'.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -445,6 +445,9 @@ module.exports = {
                     identity: { $ref: 'common_api#/definitions/access_key' },
                     secret: { $ref: 'common_api#/definitions/secret_key' },
                     azure_log_access_keys: { $ref: 'common_api#/definitions/azure_log_access_keys' },
+                    region: {
+                        type: 'string'
+                    }
                 }
             },
             auth: {
@@ -652,6 +655,9 @@ module.exports = {
                                         type: 'string'
                                     },
                                     endpoint: {
+                                        type: 'string'
+                                    },
+                                    region: {
                                         type: 'string'
                                     },
                                     identity: { $ref: 'common_api#/definitions/access_key' },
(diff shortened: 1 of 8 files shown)


Neon-White (commented):
Looks good, just one question - is there a reason that the `region` attribute is added to the connection only in `update_external_connection` and not in `add_external_connection` as well?

shirady:
> Looks good, just one question - is there a reason that the region attribute is added to the connection only in update_external_connection and not in add_external_connection as well?

Would you please mark the line where you think I didn't add the region attribute?

I'll try to explain although I'm not sure what you meant exactly:
In `add_external_connection` schema the region was already set - see `account.api` lines 415-417:
https://github.com/noobaa/noobaa-core/blob/7ffb2e6123df5ea22e778cd79fe4b20977466b07/src/api/account_api.js#L415-L417

But in the function `add_external_connection(req)` we didn't extract the region from `req`, and I added it in line 717 (see it in the files changes, this was the original line): 
https://github.com/noobaa/noobaa-core/blob/7ffb2e6123df5ea22e778cd79fe4b20977466b07/src/server/system_services/account_server.js#L717
and eventually, the info that would contain the region will be saved in the DB (we add the region to `info` and it will be saved under `sync_credentials_cache`):
https://github.com/noobaa/noobaa-core/blob/7ffb2e6123df5ea22e778cd79fe4b20977466b07/src/server/system_services/account_server.js#L752-L759

Neon-White (approved):

PR #7434, opened 2023-08-07
NSFS | MPU | optimization

### LINKED ISSUE #7348: NSFS Presto uploads large file as multiparts, calls complete with short timeout, causing retries and race on the mpu dir
### Environment info
- NooBaa Version: **5.10**
- Platform: **OpenShift 4.10**

### Actual behavior
1. NSFS complete_multipart_upload copies the parts one by one to a final file, and once it's done it moves the final file to target dir and deletes the parts files.
2. For 8 GB files, the default timeout of presto for the complete call is too short (few seconds), and then it aborts and sends second and third retries.
3. Presto will give up anyway, but we also see the retries print errors in the log of "No such file or directory", because the first complete request will eventually finish copying the data and then goes and deletes the part files, failing the other requests.

### Expected behavior
1. We should prefer to upload parts directly to the final file by offset if possible.
2. If not possible, we somehow need to "lock" the multipart directory to prevent more than one requests modifying it.

### Steps to reproduce
1. Presto write large files to S3 endpoints.

romayalon:
### Explain the changes
1. 

1. Fixed #7348 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -187,6 +187,9 @@ parse_open_flags(std::string flags)
             bits |= O_RDWR;
             bits &= ~(O_RDONLY | O_WRONLY);
             break;
+        case '*':
+            bits &= ~(O_TRUNC);
+            break;
         case 's':
             bits |= O_SYNC;
             break;
@@ -220,10 +223,13 @@ const static std::vector<std::string> GPFS_XATTRS{ GPFS_ENCRYPTION_XATTR_NAME };
 const static std::vector<std::string> USER_XATTRS{
     "user.content_type",
     "user.content_md5",
-    "user.version_id",
-    "user.prev_version_id",
-    "user.delete_marker",
-    "user.dir_content",
+    "user.noobaa.version_id",
+    "user.noobaa.prev_version_id",
+    "user.noobaa.delete_marker",
+    "user.noobaa.dir_content",
+    "user.noobaa.part_offset",
+    "user.noobaa.part_size",
+    "user.noobaa.part_etag",
     "user.storage_class",
     "user.noobaa.restore.request",
     "user.noobaa.restore.ongoing",
@@ -1401,22 +1407,31 @@ struct FileWrite : public FSWrapWorker<FileWrap>
 {
     const uint8_t* _buf;
     size_t _len;
+    off_t _offset;
     FileWrite(const Napi::CallbackInfo& info)
         : FSWrapWorker<FileWrap>(info)
         , _buf(0)
         , _len(0)
+        , _offset(-1)
     {
         auto buf = info[1].As<Napi::Buffer<uint8_t>>();
         _buf = buf.Data();
         _len = buf.Length();
-        Begin(XSTR() << "FileWrite " << DVAL(_wrap->_path) << DVAL(_len));
+        if (info.Length() > 3 && !info[3].IsUndefined()) {
+            _offset = info[3].As<Napi::Number>();
+        }
+        Begin(XSTR() << "FileWrite " << DVAL(_wrap->_path) << DVAL(_len) << DVAL(_offset));
     }
     virtual void Work()
     {
         int fd = _wrap->_fd;
         CHECK_WRAP_FD(fd);
-        // TODO: Switch to pwrite when needed
-        ssize_t bw = write(fd, _buf, _len);
+        ssize_t bw = -1;
+        if (_offset >= 0) {
+            bw = pwrite(fd, _buf, _len, _offset);
+        } else {
+            bw = write(fd, _buf, _len);
+        }
         if (bw < 0) {
             SetSyscallError();
         } else if ((size_t)bw != _len) {
@@ -1429,9 +1444,11 @@ struct FileWritev : public FSWrapWorker<FileWrap>
 {
     std::vector<struct iovec> iov_vec;
     ssize_t _total_len;
+    off_t _offset;
     FileWritev(const Napi::CallbackInfo& info)
         : FSWrapWorker<FileWrap>(info)
         , _total_len(0)
+        , _offset(-1)
     {
         auto buffers = info[1].As<Napi::Array>();
         const int buffers_len = buffers.Length();
@@ -1445,13 +1462,21 @@ struct FileWritev : public FSWrapWorker<FileWrap>
             iov_vec[i] = iov;
             _total_len += iov.iov_len;
         }
-        Begin(XSTR() << "FileWritev " << DVAL(_wrap->_path) << DVAL(_total_len) << DVAL(buffers_len));
+        if (info.Length() > 2 && !info[2].IsUndefined()) {
+            _offset = info[2].As<Napi::Number>();
+        }
+        Begin(XSTR() << "FileWritev " << DVAL(_wrap->_path) << DVAL(_total_len) << DVAL(buffers_len) << DVAL(_offset));
     }
     virtual void Work()
     {
         int fd = _wrap->_fd;
         CHECK_WRAP_FD(fd);
-        ssize_t bw = writev(fd, iov_vec.data(), iov_vec.size());
+        ssize_t bw = -1;
+        if (_offset >= 0) {
+            bw = pwritev(fd, iov_vec.data(), iov_vec.size(), _offset);
+        } else {
+            bw = writev(fd, iov_vec.data(), iov_vec.size());
+        }
         if (bw < 0) {
             SetSyscallError();
         } else if (bw != _total_len) {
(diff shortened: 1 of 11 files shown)


liranmauda (approved):
Some comments but LGTM

src/sdk/namespace_fs.js
```diff
@@ -1465,76 +1529,133 @@ class NamespaceFS {
...
+
+    async _copy_bytes(fs_context, src_file, dst_file, size, write_offset, read_offset) {
+        dbg.log0(`NamespaceFS._copy_bytes size=${size} read_offset=${read_offset} write_offset=${write_offset}`);
+        let buffer_pool_cleanup = null;
+        try {
+            let read_pos = Number(read_offset || 0);
+            let bytes_written = 0;
+            const total_bytes_to_write = Number(size);
+            let write_pos = Number(write_offset || 0);
+            for (;;) {
```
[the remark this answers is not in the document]
romayalon: I passed this one because while(true) throws eslint error on constant condition

src/sdk/namespace_fs.js
```diff
@@ -1390,35 +1398,89 @@ class NamespaceFS {
...
+
+            const md_upload_path = this._get_part_md_path(params);
+            part_md_file = await this._open_file(fs_context, md_upload_path, md_open_mode);
+            let md_upload_params = {
+                fs_context, params, object_sdk, upload_path: md_upload_path, open_mode: md_open_mode,
+                target_file: part_md_file, file_path: md_upload_path
+            };
+
+            let upload_res;
+            const pre_known_size = params.size;
```
guymguym, line 1437: change to boolean to allow size 0 parts (theoretic)
```suggestion
             pre_known_size = (size >= 0);
```

src/sdk/namespace_fs.js
```diff
@@ -1390,35 +1398,89 @@ class NamespaceFS {
...
+            if (!pre_known_size) {
+                // 2.1
+                upload_res = await buffers_pool_sem.surround_count(config.NSFS_BUF_SIZE, async () => this._upload_stream(md_upload_params));
+                params.size = upload_res.total_bytes;
+            }
+            const offset = params.size * (params.num - 1);
+            const upload_path = this._get_part_data_path(params);
+            try {
+                await nb_native().fs.stat(fs_context, upload_path);
```
guymguym, line 1446: add comment - why we stat here?

src/sdk/namespace_fs.js
```diff
@@ -1390,35 +1398,89 @@ class NamespaceFS {
...
+    //    2.1. upload data to part_md_file
+    //    2.2. calc offset
+    //    2.3. copy the bytes to by_size file
+    //    2.4. set on the part_md_file size, offset and etag xattr
     async upload_multipart(params, object_sdk) {
-        // We can use 'wt' for open mode like we do for upload_object() when
-        // we figure out how to create multipart upload using temp files.
-        const open_mode = 'w';
+        let open_mode = 'r+';
+        const md_open_mode = 'w+';
```
guymguym, line 1421: we can add a single open mode if that can help here -
```
    { "w*", O_CREAT | O_RDWR },
```

src/sdk/namespace_fs.js
```diff
@@ -1464,77 +1534,137 @@ class NamespaceFS {
...
         } catch (err) {
             dbg.error(err);
             throw this._translate_object_error_codes(err);
         } finally {
-            await this.complete_object_upload_finally(buffer_pool_cleanup, read_file, target_file, fs_context);
+            await this.complete_object_upload_finally(undefined, [...part_size_to_fd_map.values()], target_file, fs_context);
+        }
+    }
+
+    async _copy_bytes(fs_context, src_file, dst_file, size, write_offset, read_offset) {
```
guymguym, line 1609: extract to util

src/sdk/namespace_fs.js
```diff
@@ -39,6 +39,9 @@ const XATTR_USER_PREFIX = 'user.';
 // TODO: In order to verify validity add content_md5_mtime as well
 const XATTR_CONTENT_TYPE = XATTR_USER_PREFIX + 'content_type';
 const XATTR_MD5_KEY = XATTR_USER_PREFIX + 'content_md5';
+const XATTR_PART_OFFSET = XATTR_USER_PREFIX + 'part_offset';
+const XATTR_PART_SIZE = XATTR_USER_PREFIX + 'part_size';
+const XATTR_PART_ETAG = XATTR_USER_PREFIX + 'part_etag';
```
guymguym, line 44: add noobaa prefix - user.noobaa.xattrname
we want for the internal xattrs, i.e ones that we do not want others to rely on, we want to add the noobaa prefix.

guymguym (changes requested):

src/native/fs/fs_napi.cpp
```diff
@@ -1397,22 +1397,31 @@ struct FileWrite : public FSWrapWorker<FileWrap>
...
     }
     virtual void Work()
     {
         int fd = _wrap->_fd;
         CHECK_WRAP_FD(fd);
-        // TODO: Switch to pwrite when needed
-        ssize_t bw = write(fd, _buf, _len);
+        ssize_t bw = -1;
+        if (_offset == 0) {
+            bw = write(fd, _buf, _len);
```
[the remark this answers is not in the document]
romayalon: done

src/sdk/namespace_fs.js
romayalon: with the new open mode - w*, the stat is redundant, I removed it

src/sdk/namespace_fs.js
romayalon: instead, what I did it to use * as truncate remover, let me know what you think

guymguym: I'd summon @eshelmarc to comment on that. WDYT on this mode. We don't want to truncate, but keep writing to the file in offsets that are multiplies of the part size. Romy said that O_APPEND is also not suitable so we wanted to open with just O_CREAT - Does this make sense?

src/util/chunk_fs.js
```diff
@@ -57,6 +61,7 @@ class ChunkFS extends stream.Transform {
                 if (this.q_buffers.length === this.iov_max || this.q_size >= config.NSFS_BUF_SIZE) await this._flush_buffers();
                 chunk = (available_size < chunk.length) ? chunk.slice(available_size) : null;
             }
+            this.total_bytes += this.q_size;
```
guymguym, line 64: I would prefer to count it inside flush right next to where we handle _total_num_buffers and offset.
I understand that we are counting the input bytes and the output bytes are the same, only that the timing and error cases is different, so I would count bytes that we already output, and that's only known inside flush.

src/util/native_fs_utils.js
```diff
@@ -0,0 +1,50 @@
...
+
+
+async function copy_bytes(buffers_pool, fs_context, src_file, dst_file, size, write_offset, read_offset) {
+    dbg.log1(`NamespaceFS._copy_bytes size=${size} read_offset=${read_offset} write_offset=${write_offset}`);
+    let buffer_pool_cleanup = null;
+    try {
+        let read_pos = Number(read_offset || 0);
+        let bytes_written = 0;
+        const total_bytes_to_write = Number(size);
+        let write_pos = Number(write_offset || 0);
```
guymguym, line 14:
```suggestion
        let write_pos = write_offset >= 0 ? write_offset : 0;
```

src/util/native_fs_utils.js
```diff
@@ -0,0 +1,50 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const dbg = require('../util/debug_module')(__filename);
+
+
+async function copy_bytes(buffers_pool, fs_context, src_file, dst_file, size, write_offset, read_offset) {
```
guymguym, line 7: add jsdoc with types

src/sdk/namespace_fs.js
```diff
@@ -1390,35 +1393,82 @@ class NamespaceFS {
...
+            const upload_info = await this._finish_upload(md_upload_params);
             return upload_info;
         } catch (err) {
             this.run_update_issues_report(object_sdk, err);
             throw this._translate_object_error_codes(err);
         } finally {
             try {
-                if (target_file) await target_file.close(fs_context);
+                if (target_file) target_file.close(fs_context);
+                if (part_md_file) part_md_file.close(fs_context);
```
guymguym, line 1464: need to await each one individually. I think its time we have a small util function `native_fs_utils.finally_close_files(list_of_files)` that takes a list of files and calls close() on all (maybe with P.map with some small concurrency like 4) and catches the errors on each one individually to print to the log, but it should not throw so that we can call it in finally.

PR #7436, opened 2023-08-08
Add #!/bin/bash header to bash files

romayalon:
### Explain the changes
1. The changed files were missing #!/bin/bash header which caused the following error while trying to build noobaa -
```Error while loading //./src/deploy/NVA_build/install_arrow_build.sh: Exec format error```
Fun Fact - it's called Shebang - https://en.wikipedia.org/wiki/Shebang_%28Unix%29
1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


alphaprinz (approved):
Thanks

PR #7442, opened 2023-08-15
Migrating to AWS SDK V3 | Add AWS SDK V3 Wrapper

shirady:
### Explain the changes
1. add aws sdk v3 wrapper, in case we didn't receive the region (best practice is to get the region of the target bucket from the client).

1. none.

### Testing Instructions:
**Analyze resource**
We use analyze resource with a change - we set the region to be the `config.DEFAULT_REGION`, params as aws sdk v3 structure, and use `noobaa_s3_client.get_s3_client_v3_params` function.
2. Create namespacestore:
`nb namespacestore create <aws-s3 or s3-compatible>  <namespace-store-name>`
3. Run analyze resource (It creates the log files and then deletes the created job):
3. Cases:
1 - s3-compatible - Noobaa bucket (with noobaa system as a server, the endpoint was taken from the field `InternalDNS` under `S3 Addresses`).
2 - aws-s3 without flag region on a bucket in a region that is not 'us-east-1' (for example 'us-west-1').

**To run jest tests:**

- [ ] Doc added/updated
- [x] Tests added: edit existing tests in `test_noobaa_s3_client.test.js`.

### DIFF
--- a/src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
+++ b/src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
@@ -1,0 +1,39 @@
+/* Copyright (C) 2023 NooBaa */
+'use strict';
+
+const { S3 } = require('@aws-sdk/client-s3');
+const dbg = require('../../util/debug_module')(__filename);
+
+/**
+ * Same as S3Client but wraps the send method to resubmit the request with the correct region
+ * if the server responds that another region should be used.
+ * 
+ * S3 Interface, S3 Class and operations in AWS SDK V3:
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/S3/
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Class/S3/
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/
+ */
+class S3ClientAutoRegion extends S3 {
+
+    /**
+     * Overriding the send method by wrapping it and fixing the region if the error indicates it.
+     * All the methods of the base class call send() to submit an SDK command.
+     * @override
+     */
+    async send(...args) {
+        try {
+            const res = await super.send(...args);
+            return res;
+        } catch (err) {
+            const region = err.$response?.headers?.['x-amz-bucket-region'];
+            if (!region) throw err;
+            dbg.log0(`Updating region to new region ${region} in bucket ${args[0].input.Bucket}, will call ${args[0].constructor.name} again with the new region`);
+            this.config.region = region;
+            const res = await super.send(...args);
+            return res;
+        }
+    }
+}
+
+// EXPORTS
+exports.S3ClientAutoRegion = S3ClientAutoRegion;
--- a/src/sdk/noobaa_s3_client/noobaa_s3_client.js
+++ b/src/sdk/noobaa_s3_client/noobaa_s3_client.js
@@ -6,7 +6,7 @@ const https = require('https');
 const { HttpProxyAgent } = require('http-proxy-agent');
 const { HttpsProxyAgent } = require('https-proxy-agent');
 const { S3ClientSDKV2 } = require('./noobaa_s3_client_sdkv2');
-const { S3 } = require("@aws-sdk/client-s3");
+const { S3ClientAutoRegion } = require('./noobaa_s3_client_sdkv3');
 const { NodeHttpHandler } = require("@aws-sdk/node-http-handler");
 const config = require('../../../config');
 const http_utils = require('../../util/http_utils');
@@ -20,7 +20,7 @@ function get_s3_client_v3_params(params) {
         change_s3_client_params_to_v2_structure(params);
         s3_client = new S3ClientSDKV2(params);
     } else {
-        s3_client = new S3(params);
+        s3_client = new S3ClientAutoRegion(params);
     }
     return s3_client;
 }
--- a/src/test/unit_tests/jest_tests/test_noobaa_s3_client.test.js
+++ b/src/test/unit_tests/jest_tests/test_noobaa_s3_client.test.js
@@ -2,10 +2,10 @@
 /* eslint-disable no-undef */
 'use strict';
 
-const { S3 } = require("@aws-sdk/client-s3");
 const { NodeHttpHandler } = require("@aws-sdk/node-http-handler");
 const { Agent } = require("http");
 const { S3ClientSDKV2 } = require("../../../sdk/noobaa_s3_client/noobaa_s3_client_sdkv2");
+const { S3ClientAutoRegion } = require("../../../sdk/noobaa_s3_client/noobaa_s3_client_sdkv3");
 const noobaa_s3_client = require("../../../sdk/noobaa_s3_client/noobaa_s3_client");
 const config = require('../../../../config');
 
@@ -40,7 +40,7 @@ describe('noobaa_s3_client get_s3_client_v3_params', () => {
             config.AWS_SDK_VERSION_3_DISABLED = false;
             const params = {};
             const s3 = noobaa_s3_client.get_s3_client_v3_params(params);
-            expect(s3).toBeInstanceOf(S3);
+            expect(s3).toBeInstanceOf(S3ClientAutoRegion);
         });
 
         it('should choose by signatureVersion v4', () => {
@@ -50,7 +50,7 @@ describe('noobaa_s3_client get_s3_client_v3_params', () => {
                 signatureVersion: signature_version,
             };
             const s3 = noobaa_s3_client.get_s3_client_v3_params(params);
-            expect(s3).toBeInstanceOf(S3);
+            expect(s3).toBeInstanceOf(S3ClientAutoRegion);
         });
 
     });


src/util/string_utils.js
```diff
@@ -144,6 +144,14 @@ function escape_reg_exp(str) {
...
 }
 
+// indexOfEnd is the index of of string_to_search including it
+// Based on this: https://stackoverflow.com/a/18893403
+// Didn't use it as is since it creates eslint error: no-extend-native error
+function index_of_end(original_string, string_to_search) {
+    const index = original_string.indexOf(string_to_search);
+    return index === -1 ? -1 : index + string_to_search.length;
+}
+
```
guymguym, line 154: I suggest to rename it to `index_after`, add types, and shorten variables -

```suggestion
/**
 * Search the string `s` in `str` and if found return the offset right after it.
 * @type {string} str
 * @type {string} s
 * @returns {number}
 */
function index_after(str, s) {
    const i = str.indexOf(s);
    return i >= 0 ? i + s.length : i;
}

```

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
```diff
@@ -0,0 +1,192 @@
...
+}
+
+/**
+ * S3 Interface, S3 Class and operations in AWS SDK V3:
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/S3/
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Class/S3/
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/
+ */
+
+class S3ClientSDKV3 {
```
guymguym, line 64: why not extend the S3Client and override the send function as we discussed?
This will eliminate the need to redefine all the api methods.
Unless I'm missing something I think it can just be:

```js
/**
 * Same as S3Client but wraps the send method to resubmit the request with the correct region
 * if the server responds that another region should be used.
 */
class S3ClientAutoRegion extends S3Client {

    /**
     * Overriding the send method by wrapping it and fixing the region if the error indicates it.
     * All the methods of the base class call send() to submit an SDK command.
     * @override
     */
    async send(params) {
        try {
           const res = await super.send(params);
        } catch (err) {
            const region = get_region_from_aws_error(err);
            if (!region) throw err;
            this.config.region = region;
            return await super.send(params);
        }
    }

}

```

guymguym (changes requested):

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
shirady: Let's take an example: we have a client that didn't pass the region and his first action is `headObject`.
In HTTP HEAD response body is empty.
Hence in this case we would get the error code 400 without any attached information in the error, I would not have the information to extract on the region and I would throw an error.
This is the way I chose to pay for the first time in `listObjects` with the wrong region and could extract the region from the error.

Anyway, I know of an action that doesn't behave the same (for example, return a different response type) and I would use this class in similar cases.

src/util/string_utils.js
shirady: I like meaningful names (and they are easier to search when needed).
I also attached the link to the original code because it helps to understand the context (i would imagine that this kind of function would be built-in...).

guymguym: IMHO this is a function that is just trivial - I can't think of anyone going to debug it after it being written, ever. Also I don't try to predict what will be added to the js language. Feel free to keep long names if you find it clearer but in trivial cases I tend to prefer short names (i for index, s for string are just standard).

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
guymguym: TemporaryRedirect is http 302 or 307 and PermanentRedirect is http 301 or 308 and both send a Location header as a response. Is that not the case for HeadObject with a wrong region?

shirady: In `headObject` with a wrong region, I get `301: UnknownError` without an error that I can work with.

guymguym: @shirady try this:

```js
/* Copyright (C) 2023 NooBaa */
'use strict';

/**
 * Same as S3Client but wraps the send method to resubmit the request with the correct region
 * if the server responds that another region should be used.
 */
class S3ClientAutoRegion extends S3 {

    /**
     * Overriding the send method by wrapping it and fixing the region if the error indicates it.
     * All the methods of the base class call send() to submit an SDK command.
     * @override
     */
    async send(...args) {
        try {
            console.log('\n' + args[0].constructor.name);
            console.log('REGION', this.config.region);
            const res = await super.send(...args);
            console.log('OK', res.$metadata.httpStatusCode);
            return res;
        } catch (err) {
            console.log('ERROR', err.message);
            const region = err.$response?.headers?.['x-amz-bucket-region'];
            if (!region) throw err;
            console.log('NEW REGION', region);
            this.config.region = region;
            const res = await super.send(...args);
            console.log('NEW OK', res.$metadata.httpStatusCode);
            return res;
        }
    }
}

async function main() {
    const s3 = new S3ClientAutoRegion();
    const Bucket = 'guymguym';
    const Key = 'lala';
    const wrong_region = 'eu-west-2';

    s3.config.region = wrong_region;
    await s3.listBuckets({});

    s3.config.region = wrong_region;
    await s3.headBucket({ Bucket });

    s3.config.region = wrong_region;
    await s3.listObjects({ Bucket });

    s3.config.region = wrong_region;
    await s3.putBucketTagging({ Bucket, Tagging: { TagSet: [] } });

    s3.config.region = wrong_region;
    await s3.headObject({ Bucket, Key });

    s3.config.region = wrong_region;
    await s3.getObject({ Bucket, Key });

    s3.config.region = wrong_region;
    await s3.putObjectTagging({ Bucket, Key, Tagging: { TagSet: [] } });

    s3.config.region = wrong_region;
    await s3.putObject({ Bucket, Key, Body: Key });

    s3.destroy();
}

main();

```

shirady: @guymguym Thank you
I have a couple of concerns:
1) I don't know in what action I wouldn't get the region in the response header - and it will be a bug... (very specific: create a client without region first using an action that in its response doesn't have 'x-amz-bucket-region').
2) Using it I would mean losing the control that I had when I explicitly defined for every action what to run and used a flag to check if I already updated the region (I defined the actions under the class).
3) I saw a code snippet in Python ([here](https://www.programcreek.com/python/?CodeExample=get%20bucket%20region)) that does two extra checks: checks if the region appears in the error and if both of them not runs a `headBucket` and take it from its header (align to what I did at first). The last check cannot be used in this class as is now because it works for every send (and not a specific use to get the region).

guymguym: > I don't know in what action I wouldn't get the region in the response header - and it will be a bug... (very specific: create a client without region first using an action that in its response doesn't have 'x-amz-bucket-region').

I prefer this optimistic approach, because the other way would be that every operation would have to start by calling GetBucketLocation API to find the bucket location first (but definitely shouldn't call ListObjects as you have currently in the PR which is way to heavy). Also calling any other call than then intended API is prone to Bucket Policy errors since in AWS every IAM can be allowed to perform different API calls. The safest approach is to invoke only the requested operation and parse the response headers. You can see that I tested this works for a handful of operations above, and all send this header as expected. If AWS doesn't reply that header for a specific API - we can workaround it.

> Using it I would mean losing the control that I had when I explicitly defined for every action what to run and used a flag to check if I already updated the region (I defined the actions under the class).

I don't want us to have to maintain the list of S3 commands - I don't think this adds any real control. This handling can and should be dynamic. If you want to assert any behavior - we can add a test.

> I saw a code snippet in Python ([here](https://www.programcreek.com/python/?CodeExample=get%20bucket%20region)) that does two extra checks: checks if the region appears in the error and if both of them not runs a headBucket and take it from its header (align to what I did at first). The last check cannot be used in this class as is now because it works for every send (and not a specific use to get the region).

I would start just with parsing the header. If we see that calling GetBucketLocation or HeadBucket is needed we can always add that inside of our send() call as a fallback option to get the region. I don't see a problem to add such calls there as needed.

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
```diff
@@ -0,0 +1,39 @@
...
+     */
+    async send(...args) {
+        try {
+            const res = await super.send(...args);
+            return res;
+        } catch (err) {
+            const region = err.$response?.headers?.['x-amz-bucket-region'];
+            if (!region) throw err;
+            this.config.region = region;
+            const res = await super.send(...args);
```
guymguym, line 32: worth logging something here

shirady: Added, thanks.

guymguym (approved):

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
```diff
@@ -0,0 +1,41 @@
...
+
+const { S3 } = require('@aws-sdk/client-s3');
+const dbg = require('../../util/debug_module')(__filename);
+
+/**
+ * S3 Interface, S3 Class and operations in AWS SDK V3:
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Interface/S3/
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/Package/-aws-sdk-client-s3/Class/S3/
+ * https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/
+ */
```
guymguym, line 12: this is a leftover? if needed perhaps merge with the next comment of the class?

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
```diff
@@ -0,0 +1,41 @@
...
+     * @override
+     */
+    async send(...args) {
+        try {
+            const res = await super.send(...args);
+            return res;
+        } catch (err) {
+            const region = err.$response?.headers?.['x-amz-bucket-region'];
+            if (!region) throw err;
+            dbg.log0(`Updating region to new region ${region}, will call ${args[0].constructor.name} again with the new region`);
```
guymguym, line 32: also print the bucket name?

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
shirady: I merged the comments.

src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js
shirady: I added the bucket name (the assumption is that the only command that doesn't have an input of bucket is `ListBuckets`, but it accepts every region so we will not get to that printing).

PR #7447, opened 2023-08-17
Remove python dependency explicitly from RPM package

tangledbytes:
### Explain the changes
This PR removes the python dependency from the RPM package explicitly.

Why was this added in the first place?
RPM by default will detect the dependencies by looking at things like shebang, etc. We have certain python files for node-gyp config etc which are not required at runtime at all but RPM noticed the files and added the python dependencies.

### Testing Instructions:
1. Run `make rpm CONTAINER_PLATFORM=linux/arm64`
2. On a centos/RHEL system run `dnf install ./<name>`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/deploy/RPM_build/noobaa.spec
+++ b/src/deploy/RPM_build/noobaa.spec
@@ -22,6 +22,8 @@ Recommends: jemalloc
 
 %global __os_install_post %{nil}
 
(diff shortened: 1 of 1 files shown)


liranmauda (approved):
LGTM

PR #7452, opened 2023-08-21
Migrating to AWS SDK V3 | Implementation In Analyze Resource and Namespace Monitor

shirady:
### Explain the changes
1. Implementation of AWS SDK V3 In Namespace Monitor.
2. Implementation of AWS SDK V3  In Analyze Resource (use passed region).
3. Omit endpoint parameter for AWS endpoint when using AWS SDK V3.
4. Change the default in the config file to use AWS SDK v3.

1. none.

### Testing Instructions:
**Analyze Resource**
2. For backingstore, run:
`nb backingstore create aws-s3 <backing-store-name>` (optional: add the `--region` flag)
`nb backingstore create s3-compatible <backing-store-name>`
`nb diagnostics analyze backingstore <backing-store-name>`
3. For namespacestore, run:
`nb namespacestore create aws-s3 <namespace-store-name>`  (optional: add the `--region` flag)
`nb namespacestore create s3-compatible <namespace-store-name>`
3. For all resources (backingstore and namespacestore), run:
`nb diagnostics analyze resources`
It creates the log files and then deletes the created job.
5. Check the logs of noobaa-core: `kubectl logs -f` look for `test_s3_resource` printings.

Cases:
1 - s3-compatible - Noobaa bucket (with noobaa system as a server, the endpoint was taken from the field `InternalDNS` under `S3 Addresses`).
2 - aws-s3 without flag region on a bucket in a region that is not 'us-east-1' (for example 'us-west-1') and aws-s3 with flag region.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -723,7 +723,7 @@ config.TIERING_TTL_MS = 30 * 60 * 1000; // 30 minutes
 // AWS SDK VERSION     //
 /////////////////////////
 
-config.AWS_SDK_VERSION_3_DISABLED = true;
+config.AWS_SDK_VERSION_3_ENABLED = true;
 config.DEFAULT_REGION = 'us-east-1';
 
 /////////////////////
--- a/src/sdk/noobaa_s3_client/noobaa_s3_client.js
+++ b/src/sdk/noobaa_s3_client/noobaa_s3_client.js
@@ -1,6 +1,7 @@
 /* Copyright (C) 2023 NooBaa */
 'use strict';
 
+const _ = require('lodash');
 const http = require('http');
 const https = require('https');
 const { HttpProxyAgent } = require('http-proxy-agent');
@@ -10,6 +11,7 @@ const { S3ClientAutoRegion } = require('./noobaa_s3_client_sdkv3');
 const { NodeHttpHandler } = require("@smithy/node-http-handler");
 const config = require('../../../config');
 const http_utils = require('../../util/http_utils');
+const cloud_utils = require('../../util/cloud_utils');
 
 // The params are the AWS SDK V3 params.
 // params = a map of parameters that are passed to the constructor of S3 Object in order to to bind to every request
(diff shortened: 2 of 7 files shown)


Neon-White (approved):

config.js
```diff
@@ -723,7 +723,7 @@ config.TIERING_TTL_MS = 30 * 60 * 1000; // 30 minutes
 // AWS SDK VERSION     //
 /////////////////////////
 
-config.AWS_SDK_VERSION_3_DISABLED = true;
+config.AWS_SDK_VERSION_3_DISABLED = false;
```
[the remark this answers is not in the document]
shirady: It's not too late, I can refactor it (in this PR or in a separate PR it doesn't matter).
I'll share my thoughts so we will find a better suggestion.
I use the suffix DISABLED to indicate that it's a boolean, the suggestion to remove it doesn't indicate what kind of value to expect.
The main use of this variable is in case the new AWS SDK V3 would create a bug and we want to switch back to AWS SDK V2 fast - the main use of it is here:
https://github.com/noobaa/noobaa-core/blob/06d01f2e4c1ac1cd14368f82e550946ec2e5c23b/src/sdk/noobaa_s3_client/noobaa_s3_client.js#L28-L33
Changing it to ENABLED can be a solution for the double negative you're raising, but it would create a negative condition in the function above (`return !config.AWS_SDK_VERSION_3_ENABLED`...).
WDYT?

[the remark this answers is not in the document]
shirady: If it is confusing, then it should be changed.
I changed it to `AWS_SDK_VERSION_3_ENABLED`.

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -89,7 +97,7 @@ function get_requestHandler_with_suitable_agent(endpoint) {
             httpAgent: agent
         });
     } else {
-        throw new Error(`Protocol of ${endpoint} doesn't have a matching agent (valid only for HTTP, HTTPS, WSS only)`);
+        dbg.error(`Protocol of ${endpoint} doesn't have a matching agent (valid only for HTTP, HTTPS, WSS only)`);
```
dannyzaken, line 100: why did you remove the `throw`, is this a valid state? if yes then we shouldn't log an error. if it's an invalid state it's better to throw an identify it earlier

shirady: I removed it because this function meant to get `NodeHttpHandler` and I thought that there might be cases that a S3 client doesn't really need this property and it might be fine, but rethinking it - in the cases where it is needed and someone used it improperly it would create weird errors... I'll throw the error as it was.

src/tools/diagnostics/analyze_resource/analyze_resource.js
```diff
@@ -24,6 +25,7 @@ async function main() {
     try {
         const resource_type = process.env.RESOURCE_TYPE;
         check_supported_resource_type(resource_type);
+        console.info('connection_basic_details', connection_basic_details);
```
liranmauda, line 28: change it to dbg maybe?

liranmauda (approved):
LGTM

src/tools/diagnostics/analyze_resource/analyze_resource.js
shirady: I prefer leaving it with the console.log/info/error, etc.
As we checked, I didn't use `dbg.log` in this file (it only warps for the process name and time).

dannyzaken (approved):

PR #7464, opened 2023-08-28
Adding support for SSL and connection string for Postgres client + SSL tests (2nd commit)

jackyalbo:
### Explain the changes
1. Adding support for SSL/TLS for working with an external Postgres DB. in case of adding env variable POSTGRES_SSL_REQUIRED
2. Adding support for supporting self-signed server certificate, by using env variable POSTGRES_SSL_UNAUTHORIZED
3. Adding support to work with client certificate - when located in /etc/external-db-secret (like for MGMT and S3)
4. Adding support to work with a unified connection string (instead of providing host, dbname, port, and creds) with POSTGRES_CONNETION_STRING env variable

Testing:
5. Added docker builder to create an SSL-supported version for external PG (v15)
6. Added Makefile actions for building this docker image: build_ssl_postgres
7. Added Makefile for running sanity and unit test on top of the docker image for SSL Postgres
8. Added actions to run both sanity and unit test for SSL Postgres

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/sanity-ssl.yaml
+++ b/.github/workflows/sanity-ssl.yaml
@@ -1,0 +1,20 @@
+name: Build & Sanity SSL
+on: [push, pull_request]
+
+jobs:
+  run-sanity-ssl-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Run Build & SSL Sanity Tests
+        run: |
+          set -x
+          mkdir -p logs/sanity-test-logs
+          chmod 777 logs/sanity-test-logs
+          make test-external-pg-sanity
--- a/src/deploy/NVA_build/ssl_conf.sh
+++ b/src/deploy/NVA_build/ssl_conf.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+# echo ssl setting into pg_hba.conf configuration file
+echo 'local all all trust' > /var/lib/postgresql/data/pg_hba.conf
+echo 'hostssl all all all cert clientcert=verify-full' >> /var/lib/postgresql/data/pg_hba.conf
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -27,6 +27,7 @@ const { RpcError } = require('../rpc');
 const SensitiveString = require('./sensitive_string');
 const time_utils = require('./time_utils');
 const config = require('../../config');
+const ssl_utils = require('./ssl_utils');
 
 const DB_CONNECT_ERROR_MESSAGE = 'Could not acquire client from DB connection pool';
 mongodb.Binary.prototype[util.inspect.custom] = function custom_inspect_binary() {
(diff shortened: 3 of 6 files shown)


src/deploy/NVA_build/ssl_conf.sh
```diff
@@ -0,0 +1,3 @@
+# echo ssl setting into pg_hba.conf configuration file
```
liranmauda, line 1: add `!#/bin/bash` as a first line

src/util/postgres_client.js
```diff
@@ -1412,19 +1415,25 @@ class PostgresClient extends EventEmitter {
...
+                connectionString: process.env.POSTGRES_CONNETION_STRING,
+                ...params,
+            };
+        } else {
+            // TODO: This need to move to another function
+            this.new_pool_params = {
+                host: process.env.POSTGRES_HOST || '127.0.0.1',
+                user: process.env.POSTGRES_USER || 'postgres',
+                password: process.env.POSTGRES_PASSWORD || 'noobaa',
+                database: process.env.POSTGRES_DBNAME || 'nbcore',
```
liranmauda, line 1430: should we move all process.env into the config?

.github/workflows/sanity-ssl.yaml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Run Build & SSL Sanity Tests
+        run: |
+          set -x
+          mkdir -p logs/sanity-test-logs
+          chmod 777 logs/sanity-test-logs
+          make test-external-pg-sanity
```
romayalon, line 20: why do we need a different workflow? can't we use the sanity workflow and run both make test-external-pg-sanity & make test-sanity

.github/workflows/ssl-postgres-unit-tests.yaml
```diff
@@ -0,0 +1,16 @@
...
+    timeout-minutes: 90
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Run Unit Tests with External SSL Postgres
+        run: make test-external-postgres
```
romayalon, line 16: why do we need a different workflow? can't we use the sanity workflow and run both make test-postgres & make test-external-postgres

src/util/postgres_client.js
```diff
@@ -1357,6 +1358,7 @@ class PostgresClient extends EventEmitter {
     async dropDatabase() {
         let pg_client;
         try {
+            if (process.env.POSTGRES_SSL_REQUIRED) await this._load_ssl_certs();
```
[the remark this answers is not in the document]
jackyalbo: why? we want to use it in the future from the operator

.github/workflows/ssl-postgres-unit-tests.yaml
jackyalbo: why not? running one after the other will make it really lengthy... is it something we want to avoid? (limited resources in github?). let's have a discussion about it.

liranmauda (approved):

.github/workflows/sanity-ssl.yaml
jackyalbo: Decided to leave for now only test-external-pg-sanity. The unit test will run only on default Postgres for now.

.github/workflows/ssl-postgres-unit-tests.yaml
jackyalbo: Decided to leave for now only test-external-pg-sanity. The unit test will run only on default Postgres for now.

src/util/postgres_client.js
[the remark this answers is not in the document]
jackyalbo: decided to keep the old format for now

src/util/postgres_client.js
jackyalbo: decided to leave the old format for now

PR #7465, opened 2023-08-29
Add NotPrefix, NotAction and NotResource to bucket policy

nadavMiz:
### Explain the changes
1. Add NotPrefix, NotAction and NotResource augments to bucket policy
they can be used instead of the Prefix,Action,Resource counterparts 

1. https://bugzilla.redhat.com/show_bug.cgi?id=2222487

### Testing Instructions:
1. add bucket policy with one of those arguments
2. bucket policy should work

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -309,52 +309,118 @@ module.exports = {
                 Statement: {
                     type: 'array',
                     items: {
-                        type: 'object',
-                        required: ['Effect', 'Action', 'Principal', 'Resource'],
-                        properties: {
-                            Sid: {
-                                type: 'string'
-                            },
-                            Effect: {
-                                enum: ['Allow', 'Deny'],
-                                type: 'string'
-                            },
-                            Action: {
-                                $ref: '#/definitions/string_or_string_array'
-                            },
-                            Principal: {
-                                $ref: '#/definitions/bucket_policy_principal'
-                            },
-                            Resource: {
-                                $ref: '#/definitions/string_or_string_array'
-                            },
-                            Condition: {
+                        allOf: [
+                            {
                                 type: 'object',
+                                required: ['Effect'],
                                 properties: {
-                                    StringEquals: {
-                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    Sid: {
+                                        type: 'string'
                                     },
-                                    StringNotEquals: {
-                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    Principal: {
+                                        $ref: '#/definitions/bucket_policy_principal'
                                     },
-                                    StringEqualsIgnoreCase: {
-                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    NotPrincipal: {
+                                        $ref: '#/definitions/bucket_policy_principal'
                                     },
-                                    StringNotEqualsIgnoreCase: {
-                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    Action: {
+                                        $ref: '#/definitions/string_or_string_array'
                                     },
-                                    StringLike: {
-                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    NotAction: {
+                                        $ref: '#/definitions/string_or_string_array'
                                     },
-                                    StringNotLike: {
-                                        $ref: '#/definitions/bucket_policy_string_condition'
+                                    Resource: {
+                                        $ref: '#/definitions/string_or_string_array'
                                     },
-                                    Null: {
-                                        $ref: '#/definitions/bucket_policy_null_condition'
+                                    NotResource: {
+                                        $ref: '#/definitions/string_or_string_array'
+                                    },
+                                    Effect: {
+                                        enum: ['Allow', 'Deny'],
+                                        type: 'string'
+                                    },
+                                    Condition: {
+                                        type: 'object',
+                                        properties: {
+                                            StringEquals: {
+                                                $ref: '#/definitions/bucket_policy_string_condition'
+                                            },
+                                            StringNotEquals: {
+                                                $ref: '#/definitions/bucket_policy_string_condition'
+                                            },
+                                            StringEqualsIgnoreCase: {
+                                                $ref: '#/definitions/bucket_policy_string_condition'
+                                            },
+                                            StringNotEqualsIgnoreCase: {
+                                                $ref: '#/definitions/bucket_policy_string_condition'
+                                            },
+                                            StringLike: {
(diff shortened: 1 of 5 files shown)


src/api/common_api.js
```diff
@@ -310,7 +310,7 @@ module.exports = {
                     type: 'array',
                     items: {
                         type: 'object',
-                        required: ['Effect', 'Action', 'Principal', 'Resource'],
+                        required: ['Effect'],
```
shirady, line 313: I understand it is a simple way to solve it, but I'm not sure if we won't "pay" for it by validating the policies in the bucket server when we could do it before.

src/server/system_services/bucket_server.js
```diff
@@ -489,26 +489,46 @@ async function put_bucket_policy(req) {
 function _validate_s3_policy(policy, bucket_name) {
     const all_op_names = _.compact(_.flatMap(OP_NAME_TO_ACTION, action => [action.regular, action.versioned]));
     for (const statement of policy.Statement) {
-        if (statement.Principal.AWS) {
-            for (const principal of _.flatten([statement.Principal.AWS])) {
+        if (!statement.Principal && !statement.NotPrincipal) {
+            throw new RpcError('MALFORMED_POLICY', 'must have Principal or NotPrincipal in policy', { detail: "" });
+        }
+        if (statement.Principal && statement.NotPrincipal) {
+            throw new RpcError('MALFORMED_POLICY', 'The policy elements Principal and NotPrincipal cannot be used in the same statement', { detail: "" });
```
shirady, line 496: Why do you send `{ detail: "" }` to the `RpcError`?
I think you don't have to:
https://github.com/noobaa/noobaa-core/blob/c497a4f17104081b58bcafbf5b64979cbb0f9ebc/src/rpc/rpc_error.js#L21-L23

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -145,33 +145,48 @@ async function has_bucket_policy_permission(policy, account, method, arn_path, r
...
-        let action_fit = false;
-        let principal_fit = false;
-        let resource_fit = false;
-        let condition_fit = true;
-        for (const action of _.flatten([statement.Action])) {
-            dbg.log1('bucket_policy: action fit?', action, method);
-            if ((action === '*') || (action === 's3:*') || (action === method)) {
-                action_fit = true;
-            }
+function _is_action_fit(method, statement_action, isNotStatement) {
```
shirady, line 148: (related also to `_is_principal_fit`, `_is_resource_fit`):
1. Would you please calculate the `isNotStatement` inside the function itself?
It would be more readable in statements calling this function.
2. Would you please add a comment with a short explanation of what we check?
Specially in `_is_resource_fit` with the regular expression.

src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -860,4 +860,90 @@ mocha.describe('s3_bucket_policy', function() {
...
+                Resource: [`arn:aws:s3:::${BKT}/*`]
+            }
+        ]};
+        await s3_owner.putBucketPolicy({
+            Bucket: BKT,
+            Policy: JSON.stringify(auth_put_policy)
+        }).promise();
+
+        await s3_a.putObject({
+            Body: 'Some data for the file... bla bla bla... ',
```
shirady, line 881: If you can, maybe use it as a constant variable (instead of writing the string every time).

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -145,33 +145,48 @@ async function has_bucket_policy_permission(policy, account, method, arn_path, r
...
+        dbg.log1('bucket_policy: ', isNotStatement ? 'NotResource' : 'Resource', ' fit?', resource_regex, arn_path);
+        if (resource_regex.test(arn_path)) {
+            resource_fit = true;
         }
-        condition_fit = await _is_condition_fit(statement, req, method);
+    }
+    return isNotStatement ? !resource_fit : resource_fit;
+}
+
+async function _is_statements_fit(statements, account, method, arn_path, req) {
```
shirady, line 183: Would you please explain this function?
If we have one statement that fits then all statements fit?

src/server/system_services/bucket_server.js
nadavMiz: I have to for MALFORMED_POLICY errors. because this value is used in 
https://github.com/noobaa/noobaa-core/blob/c497a4f17104081b58bcafbf5b64979cbb0f9ebc/src/endpoint/s3/s3_rest.js#L369-L372

shirady: You created `RpcError` with property `rpc_code` (not `code`) of `MALFORMED_POLICY` (not `MalformedPolicy`). If you can explain in the flow what will change from where we start to the code snippet you shared.

nadavMiz: not relevant anymore after removal, but we need it in case of MALFORMED_POLICY:
(https://github.com/noobaa/noobaa-core/blob/c497a4f17104081b58bcafbf5b64979cbb0f9ebc/src/endpoint/s3/s3_rest.js#L369-L372)

src/api/common_api.js
```diff
@@ -309,52 +309,113 @@ module.exports = {
...
-                            },
-                            Effect: {
-                                enum: ['Allow', 'Deny'],
-                                type: 'string'
-                            },
-                            Action: {
-                                $ref: '#/definitions/string_or_string_array'
+                        allOf: [
+                            {
+                                oneOf: [
```
jackyalbo, line 314: Not an expert with this but can't we combine the 2 parts? the principle with the principles in the bottom and all the others (Resource and Action)? This looks really unreadable...
also, leave the type and required in the beginning if you can so at least it will look like something we know from the other APIs

src/api/common_api.js
```diff
@@ -309,52 +309,113 @@ module.exports = {
...
-                                    Null: {
-                                        $ref: '#/definitions/bucket_policy_null_condition'
+                                    Effect: {
+                                        enum: ['Allow', 'Deny'],
+                                        type: 'string'
+                                    },
+                                    Condition: {
+                                        type: 'object',
+                                        properties: {
+                                            StringEquals: {
```
jackyalbo, line 393: in condition can we get all of these options in the same condition? Maybe also add a remark about how this schema is supposed to look, and maybe also a link to AWS docs.

jackyalbo (commented):
In total, I think it looks good. If you can make the schema look a little bit better then great.

jackyalbo (approved):

shirady (approved):
LGTIM

PR #7469, opened 2023-08-30
NC | NSFS | NSFS NC Systemd example script and docs for RPM installation

romayalon:
### Explain the changes
1. Added systemd example script to the RPM build
2. Added NC NSFS docs for RPM installation and build

1. 

### Testing Instructions:
1. 

- [x] Doc added/updated
- [ ] Tests added

### DIFF
--- a/docs/non_containerized_NSFS.md
+++ b/docs/non_containerized_NSFS.md
@@ -1,0 +1,79 @@
+# Non Containerized NSFS
+
+Running nsfs non containerized is useful for deploying in linux without depending on kubernetes, but requires some steps which are described next.
+
+
+## Build
+
+### Build options - 
+1. noobaa-core S3 Select is enabled. This requires boost shared objects to be present on the machine where noobaa is supposed to be installed. This feature can be disabled if not required (due to additional dependency) by running make rpm BUILD_S3SELECT=0.
+2. In order to build RPM packages for other architectures simply run make rpm CONTAINER_PLATFORM=linux/amd64 or make rpm CONTAINER_PLATFORM=linux/ppc64le
+
+Running the following command will result with an RPM file -
+```sh
+make rpm
+```
+
+
+## Install
+
+### Pre-requisites (S3 select enabled) - 
+1. boost - 
+```sh
+yum install epel-release && yum install boost
+```
+
+### A workaround for machines that cannot install epel-release -
+Use the workaround below if `yum install epel-release` resulted in the following error - 
+```
+Updating Subscription Management repositories.
+Unable to read consumer identity
+
+This system is not registered with an entitlement server. You can use subscription-manager to register.
+
+Last metadata expiration check: <SomeTimeStamp>
+No match for argument: epel-release
+Error: Unable to find a match: epel-release
+```
+1. Install wget - 
+
+```sh
+yum install wget
+```
+2. Download and install boost -
+```sh
+wget https://rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/boost-system-1.66.0-13.el8.x86_64.rpm
+wget https://rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/boost-thread-1.66.0-13.el8.x86_64.rpm
+rpm -i boost-system-1.66.0-13.el8.x86_64.rpm
+rpm -i boost-thread-1.66.0-13.el8.x86_64.rpm
+```
+
+### Install NooBaa RPM - 
+Download the RPM to the machine and install it by running the following command - 
+
+```sh
+rpm -i noobaa-core-5.14.0-1.el8.x86_64.rpm
+```
+
+After installing NooBaa RPM, it's expected to have noobaa-core source code under /usr/local/noobaa-core and an nsfs systemd example script under /etc/systemd/system/.
+
+## Run the nsfs service - 
+This systemd example script runs noobaa non containerized (currently) single mode, while /tmp/test/ is the fs root path directory.
+
+```sh
+systemctl start nsfs
+```
+
+## Test 
+1. create a bucket directory -
+```sh
+mkdir -p /tmp/test/bucket1/
+```
+2. try list the buckets - 
+```sh
+ curl http://localhost:6001
+```
+The following is an expected valid response - 
+```xml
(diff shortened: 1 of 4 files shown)


docs/non_containerized_NSFS.md
```diff
@@ -0,0 +1,62 @@
...
+1. wget - 
+
+```sh
+yum install wget
+```
+2. boost -
+```sh
+wget https://rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/boost-system-1.66.0-13.el8.x86_64.rpm
+wget https://rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/boost-thread-1.66.0-13.el8.x86_64.rpm
+rpm -i boost-thread-1.66.0-13.el8.x86_64.rpm
```
tangledbytes, line 30: I think the order here needs to be reversed? First system and then thread or else I am getting
```
[root@a82fc3b93bd8 /]# rpm -i boost-thread-1.66.0-13.el8.x86_64.rpm
warning: boost-thread-1.66.0-13.el8.x86_64.rpm: Header V3 RSA/SHA256 Signature, key ID 8483c65d: NOKEY
error: Failed dependencies:
        boost-system(x86-64) = 1.66.0-13.el8 is needed by boost-thread-1.66.0-13.el8.x86_64
        libboost_system.so.1.66.0()(64bit) is needed by boost-thread-1.66.0-13.el8.x86_64
```

docs/non_containerized_NSFS.md
```diff
@@ -0,0 +1,62 @@
...
+rpm -i boost-system-1.66.0-13.el8.x86_64.rpm
+```
+
+### Install NooBaa RPM - 
+Download the RPM to the machine and install it by running the following command -
```
tangledbytes, line 35: I think the command following this line would just install the RPM and not download it.

romayalon: @tangledbytes Yes, it's just installation for now because I don't know where will the RPM be published.
@nimrod-becker @guymguym We need to discuss what RPM download instructions will look like,
1. Where these RPMs are published?
3. Will the RPM be based on our regular version branches?
4. Will this RPM build be automated like DS/US builds?

docs/non_containerized_NSFS.md
romayalon: hmm that's weird, installing them both like that - 
`rpm -i boost-thread-1.66.0-13.el8.x86_64.rpm boost-system-1.66.0-13.el8.x86_64.rpm` won't cry on it
i'll change it thanks

tangledbytes:
One thing that slightly concerns me (not completely related to this PR) is that I added upgrade manager as part of RPM upgrades, so basically if a user already has NooBaa installed and then the install RPM for version 5.14.1 then our upgrade scripts will kick in and will do its thing. Now that's good but the issue would occur when there would be no postgresql to connect to (NSFS case) and upgrade_script would block (infinite loop) RPM install.

I think we should add a config which disable the upgrade manager. WDYT @romayalon @guymguym?

docs/non_containerized_NSFS.md
```diff
@@ -0,0 +1,62 @@
...
+
+```sh
+yum install wget
+```
+2. boost -
+```sh
+wget https://rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/boost-system-1.66.0-13.el8.x86_64.rpm
+wget https://rpmfind.net/linux/centos/8-stream/AppStream/x86_64/os/Packages/boost-thread-1.66.0-13.el8.x86_64.rpm
+rpm -i boost-system-1.66.0-13.el8.x86_64.rpm
+rpm -i boost-thread-1.66.0-13.el8.x86_64.rpm
```
guymguym, line 31: this is only a workaround for machines that cannot use `yum install epel-release && yum install boost`.
I would keep it here but as a second resort for cases where yum cannot be used.

guymguym (approved):

PravinRanjan10:
Hi all, Posting the observation for tracking purpose:
Observations on Noobaa-service(nsfs):
Systemctl start nsfs (Looks ok)
Able to start the service, and basic CRUD on buckets done.
  2.  Systemctl stop nsfs (looks like not smooth stop)
● nsfs.service - The NooBaa nsfs service.
  Loaded: loaded (/etc/systemd/system/nsfs.service; static; vendor preset: disabled)
  Active: failed (Result: exit-code) since Sun 2023-09-10 21:12:33 PDT; 1s ago
 Process: 1859288 ExecStop=/bin/kill QUIT $MAINPID (code=exited, status=64)
 Process: 1859279 ExecStart=/usr/local/node/bin/node /usr/local/noobaa-core/src/cmd/nsfs.js /tmp/test/ (code=killed, signal=TERM)
 Main PID: 1859279 (code=killed, signal=TERM)
3.  Systemctl enable nsfs [seems not working!!]
[root@pr-ss1-worker-1 ~]# systemctl enable nsfs
The unit files have no installation config (WantedBy, RequiredBy, Also, Alias
settings in the [Install] section, and DefaultInstance for template units).
This means they are not meant to be enabled using systemctl.
Possible reasons for having this kind of units are:
1) A unit may be statically enabled by being symlinked from another unit’s
  .wants/ or .requires/ directory.
2) A unit’s purpose may be to act as a helper for some other unit which has
  a requirement dependency on it.
3) A unit may be started when needed via activation (socket, path, timer,
  D-Bus, udev, scripted systemctl call, ...).
4) In case of template units, the unit is meant to be enabled with some
  instance name specified.
4.   Systemctl disable nsfs [seems not working]
5. Sometimes nsfs is in auto-start mode (specially, run enable and start)
● nsfs.service - The NooBaa nsfs service.
  Loaded: loaded (/etc/systemd/system/nsfs.service; static; vendor preset: disabled)
  Active: activating (auto-restart) (Result: exit-code) since Sun 2023-09-10 21:17:35 PDT; 1s ago
 Process: 1865942 ExecStart=/usr/local/node/bin/node /usr/local/noobaa-core/src/cmd/nsfs.js /tmp/test/ (code=exited, status=1/FAILURE)
 Main PID: 1865942 (code=exited, status=1/FAILURE) (edited)

romayalon:
@PravinRanjan10 
Hey, this is just an example script, the install section is empty, have you tried my answer on slack?

PravinRanjan10:
> @PravinRanjan10 Hey, this is just an example script, the install section is empty, have you tried my answer on slack?

Okay. Edited the nsfs.service file and able to enable/disable the nsfs. But stop still looks failed!!.

nimrod-becker:
we'll look at stop

PR #7474, opened 2023-09-03
Build RPM - some fixes

liranmauda:
### Explain the changes
- Removed redundant downloading for nodejs
- Fix an issue when building the rpm target failed when PLATFORM was not set
- Added the ability to provide the rep build image tag externally

### DIFF
--- a/src/deploy/RPM_build/packagerpm.sh
+++ b/src/deploy/RPM_build/packagerpm.sh
@@ -6,6 +6,7 @@ set -x
 dir=$(dirname "$0")
 
 SKIP_NODE_INSTALL=1 source $dir/../install_nodejs.sh
+NODE_PATH="${NODE_PATH:-/usr/local/node}"
 
 noobaaver="$(npm pkg get version | tr -d '"')"
 revision="1"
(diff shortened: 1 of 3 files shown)


liranmauda:
@tangledbytes If you can, please run it on your Linux, lets see that it work there as well 🙂

src/deploy/RPM_build/packagerpm.sh
```diff
@@ -33,9 +34,17 @@ function move_builds() {
...
 function get_node_tar() {
+    local arch=$(get_arch)
+    local filename="node-v${nodever}-linux-${arch}.tar.xz"
+
     pushd ~/rpmbuild/SOURCES/
-    local path=$(NODEJS_VERSION=${nodever} download_node)
-    mv ${path} node-${nodever}.tar.xz
+    if [ -f ${NODE_PATH}/${filename} ]
+    then
+        mv ${NODE_PATH}/${filename} node-${nodever}.tar.xz
```
tangledbytes, line 43: Don't we have to tar it again here because we untarred it on `install_nodejs.sh:54`? Also, I think we should `cp` instead of `mv` because we symlinked previously on `install_nodejs.sh`.

liranmauda: the tar in `install_nodejs.sh:54` is not removing the tarball, just opening it, so we have the tared file. 
the symlink is a link of the open files, so we can move the tar.
also previously we rm the file, and now we are not, as we don't need it clean on the base image and we can move it wherever we want.

tangledbytes (approved):
Tested on Linux x86.

PR #7479, opened 2023-09-06
Migrating to AWS SDK V3 | Use AWS SDK V3 In Unit Tests (s3_ops)

shirady:
### Explain the changes
1. Implementation of AWS SDK V3 (without wrapper) in unit tests:
s3_ops
   - Remove request handlers in stage 'error' and 'complete' and use catch error instead.
   - Remove request handlers that added version ID to query params and use the param of the action itself.
   - Replace request handlers with middlewarestack in case of namespace cache.
   - Small fixes:
     - In test 'should version head text-file' - replace the object to match type needed (string, but could be string or Error).
     - In 'test should tag text file' - fix the message value to the right one. 

1. none

### Testing Instructions:
1. `make test` or `make run-single-test testname=test_s3_ops.js` (The test runs as a part of the CI)

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/test/unit_tests/test_s3_ops.js
+++ b/src/test/unit_tests/test_s3_ops.js
@@ -8,11 +8,11 @@ const _ = require('lodash');
 const coretest = require('./coretest');
 coretest.setup({ pools_to_create: coretest.POOL_LIST });
 const config = require('../../../config');
-const AWS = require('aws-sdk');
+const { S3 } = require('@aws-sdk/client-s3');
+const { NodeHttpHandler } = require("@smithy/node-http-handler");
 const http_utils = require('../../util/http_utils');
 const mocha = require('mocha');
 const assert = require('assert');
-const querystring = require('querystring');
 const P = require('../../util/promise');
 const azure_storage = require('@azure/storage-blob');
 
@@ -54,6 +54,7 @@ const azure_mock_connection_string = `DefaultEndpointsProtocol=http;AccountName=
 mocha.describe('s3_ops', function() {
 
     let s3;
+    let s3_client_params;
     // Bucket name for the source namespace resource
     let source_namespace_bucket;
     // Bucket name for the target namespace resource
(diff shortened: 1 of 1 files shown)


tangledbytes:
@shirady I see that in the NPM docs under v2 compatible style they mention:
> The client can also send requests using v2 compatible style. However, it results in a bigger bundle size and may be dropped in next major version.

Should we instead use `const { S3Client } = require("@aws-sdk/client-s3");`?

shirady:
@tangledbytes we can use `S3Client` instead of `S3` in the unit tests, but I thought it might be unclear since we would use it only there. In the other parts of the code, we will keep using AWS SDK V2 because there might be clients that use signature version 2, and we also created a wrapper for AWS SDK V3 in case the client didn't pass the region of the target bucket ([here](https://github.com/noobaa/noobaa-core/blob/master/src/sdk/noobaa_s3_client/noobaa_s3_client_sdkv3.js)). Since in the unit tests we know that it uses signature version 4, and the region is known I used the AWS SDK V3 directly.

In the introduction ([here](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/introduction/)) under "High-Level Concepts" they show the two options: 
1. Bare-bones clients/commands
2. Aggregated clients/commands, and it is written:

> Under the hood this calls the bare-bones commands.

So by switching to the Bare-bones clients/commands, we would have the same results.
I had an idea to create a wrapper class for each command to use the bare-bones commands while keeping the AWS SDK V2 code style, code snippet for example:
```
class S3ClientSDKV3BM {
    constructor(params) {
        this.s3 = new S3 Client(params);
    }

    async headObject(params) {
        const res = await this.s3.send(new HeadObjectCommand(params));
        return res;
    }
```
But it is another part of the code that we would need to maintain - right now we get it from AWS SDK V3.

tangledbytes:
@shirady, that makes sense. Let's hope that they don't drop the support in the future 🤞.

tangledbytes (approved):

dannyzaken (approved):

PR #7486, opened 2023-09-10
complete upload - skip db access for small object (for performance)

alphaprinz:
### Explain the changes
For small objects with one part, we know the start/end of object.
There's no need to fix values in db at the end of the put object.
(It's needed for multipart upload where each part doesn't know its start/end at single part upload).
Also, set part as committed when inserting the part (again, to skip db access for removing uncommitted at the end).

Redundant db access at the end of upload.

### Testing Instructions:

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/object_io.js
+++ b/src/sdk/object_io.js
@@ -467,7 +467,7 @@ class ObjectIO {
                     start: params.start,
                     end: params.start + chunk_info.size,
                     seq: params.seq,
-                    uncommitted: !params.complete_upload && (chunks.length > 1 || typeof params.multipart_id !== 'undefined'),
+                    uncommitted: typeof params.multipart_id === 'undefined' ? undefined : true,
                     // millistamp: time_utils.millistamp(),
                     // bucket: params.bucket,
                     // key: params.key,
--- a/src/sdk/object_io.js
+++ b/src/sdk/object_io.js
@@ -454,6 +454,11 @@ class ObjectIO {
     async _upload_chunks(params, complete_params, chunks, callback) {
         try {
             const is_using_encryption = params.encryption || (params.copy_source && params.copy_source.encryption);
+            // on multipart upload we mark the parts as uncommitted
+            // because we still need to fix their offsets on complete upload.
+            // otherwise, we immediately mark the parts as committed
+            // because no further update is needed.
+            const uncommitted = Boolean(params.multipart_id);
             params.range = {
                 start: params.start,
                 end: params.start,
@@ -467,7 +472,7 @@ class ObjectIO {
                     start: params.start,
                     end: params.start + chunk_info.size,
                     seq: params.seq,
-                    uncommitted: typeof params.multipart_id === 'undefined' ? undefined : true,
+                    uncommitted: uncommitted,
                     // millistamp: time_utils.millistamp(),
                     // bucket: params.bucket,
                     // key: params.key,
--- a/src/sdk/object_io.js
+++ b/src/sdk/object_io.js
@@ -467,7 +467,7 @@ class ObjectIO {
                     start: params.start,
                     end: params.start + chunk_info.size,
                     seq: params.seq,
-                    uncommitted: !params.complete_upload,
+                    uncommitted: !params.complete_upload && (chunks.length > 1 || typeof params.multipart_id !== 'undefined'),
                     // millistamp: time_utils.millistamp(),
                     // bucket: params.bucket,
                     // key: params.key,
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -406,10 +406,16 @@ async function complete_object_upload(req) {
         set_updates.sha256_b64 = req.rpc_params.sha256_b64;
     }
 
-    const map_res =
-        req.rpc_params.multiparts ?
-        await _complete_object_multiparts(obj, req.rpc_params.multiparts) :
-        await _complete_object_parts(obj, req.rpc_params.num_parts);
+    dbg.log0("req.rpc_params.num_parts = ", req.rpc_params.num_parts, ", req.params.num_parts = ", req.params.num_parts);
+
+    let map_res;
+    if (req.rpc_params.multiparts) {
+        map_res = await _complete_object_multiparts(obj, req.rpc_params.multiparts);
+    } else if (req.rpc_params.num_parts) {
+        map_res = {size: obj.size, num_parts: req.rpc_params.num_parts};
+    } else {
+        map_res = await _complete_object_parts(obj);
+    }
 
     if (req.rpc_params.size !== map_res.size) {
         if (req.rpc_params.size >= 0) {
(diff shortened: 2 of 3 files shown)


src/server/object_services/object_server.js
```diff
@@ -406,9 +406,17 @@ async function complete_object_upload(req) {
...
+
+    if (req.rpc_params.multiparts) {
+        map_res = await _complete_object_multiparts(obj, req.rpc_params.multiparts);
+    } else if (req?.params?.num_parts === 1) {
+        //skip db access for small objects
+        dbg.log1("skipped _complete_object_multiparts");
+        map_res = {size: req.rpc_params.size, num_parts: 1};
+    } else {
+        map_res = await _complete_object_parts(obj);
+    }
```
guymguym, line 419: I would prefer to move this optimization into `_complete_object_parts()` (need to send params.num_parts inside).

Over there we can start by identifying a single part case, and optimize the hell of it, so that we don't even read the part from the db, which is an aggressive optimization which we also need to explain why it is safe (the reason is that mpu cannot have a single part per multipart... which is a bit of a "bad" assumption to take because it could easily get broken one day by someone who just doesn't expect it calling these apis).

We can also add a simpler condition that will still read the parts, but will skip updating them by noticing that they are already committed and do not require an offset/seq update - simply surround the push to parts_updates with a condition like 
```js
if (part.uncommitted || updates.set_updates) {
    context.parts_updates.push(updates);
}
```

on this line - 
https://github.com/noobaa/noobaa-core/blob/cdd266f64904a67b985fd07c9d97abfe85f43790/src/server/object_services/object_server.js#L2099

guymguym (changes requested):

src/server/object_services/object_server.js
alphaprinz: I've moved the check whether to skip the db into _complete_next_parts().
Since we're going for performance, I figured it's best to skip the db network hop altogether.
You think it's not safe enough?

src/sdk/object_io.js
```diff
@@ -467,7 +467,7 @@ class ObjectIO {
                     start: params.start,
                     end: params.start + chunk_info.size,
                     seq: params.seq,
-                    uncommitted: !params.complete_upload,
+                    uncommitted: !params.complete_upload && (chunks.length > 1 || typeof params.multipart_id !== 'undefined'),
```
guymguym, line 470: why do we care how many chunks? if this is not multipart then we can commit all parts.

src/server/object_services/object_server.js
```diff
@@ -683,7 +684,9 @@ async function copy_object_mapping(req) {
         part.obj = obj._id;
         part.bucket = req.bucket._id;
         part.multipart = multipart ? multipart._id : undefined;
-        part.uncommitted = true;
+        if (parts.length > 1) {
+            part.uncommitted = true;
+        }
```
guymguym, line 689: again, why does it matter how many parts?

src/server/object_services/object_server.js
```diff
@@ -1974,7 +1977,14 @@ async function update_endpoint_stats(req) {
 /**
  * @param {nb.ObjectMD} obj
  */
-async function _complete_object_parts(obj) {
+async function _complete_object_parts(obj, num_parts) {
+
+    if (num_parts === 1) {
+        //skip db access if only one part
+        //(this part is already ok in db, no need to fix it)
```
guymguym, line 1984: Is a single part different than any other number of parts?
The only difference is if we are on multipart where we need to fix the offsets.

guymguym (changes requested):

src/sdk/object_io.js
```diff
@@ -467,7 +467,7 @@ class ObjectIO {
                     start: params.start,
                     end: params.start + chunk_info.size,
                     seq: params.seq,
-                    uncommitted: !params.complete_upload,
+                    uncommitted: typeof params.multipart_id === 'undefined' ? undefined : true,
```
guymguym, line 470: 1. There is no need to use typeof x === 'undefined' in modern js, you can just use x === undefined. But actually in this we don't really care to identify the undefined vs other falsy values (which is the common case) so I would just use the Boolean() function for that.
2. It makes more sense to define it once outside of this chunks.map() call and define it one for all in the top level of this function.
3. It deserves a comment explaining how this uncommitted flag is used here.

```js
// on multipart upload we mark the parts as uncommitted 
// because we still need to fix their offsets on complete upload.
// otherwise, we immediately mark the parts as committed 
// because no further update is needed.
const uncommitted = Boolean(params.multipart_id);
```

src/server/object_services/object_server.js
```diff
@@ -1969,9 +1974,6 @@ async function _complete_object_parts(obj) {
 
     const parts = await MDStore.instance().find_all_parts_of_object(obj);
     _complete_next_parts(parts, context);
-    if (context.parts_updates.length) {
-        await MDStore.instance().update_parts_in_bulk(context.parts_updates);
-    }
```
guymguym, line 1974: To make this a bit safer, Instead of completely removing this call, I would leave it here if there are part_updates, but instead we should change _complete_next_parts and only insert the part_update if the part is uncommitted or has a seq change.

src/server/object_services/object_server.js
```diff
@@ -406,9 +406,14 @@ async function complete_object_upload(req) {
...
     }
 
-    const map_res = req.rpc_params.multiparts ?
-        await _complete_object_multiparts(obj, req.rpc_params.multiparts) :
-        await _complete_object_parts(obj);
+    let map_res;
+    if (req.rpc_params.multiparts) {
+        map_res = await _complete_object_multiparts(obj, req.rpc_params.multiparts);
+    } else if (req.rpc_params.num_parts) {
+        map_res = {size: obj.size, num_parts: req.rpc_params.num_parts};
```
guymguym, line 413: add short explanation
```suggestion
    } else if (req.rpc_params.num_parts) {
        // in this case we got all the info we need and can avoid quering md_store
        map_res = { size: obj.size, num_parts: req.rpc_params.num_parts };
```

guymguym (changes requested):
Hi @alphaprinz Good to see you're back!

src/server/object_services/object_server.js
```diff
@@ -1956,10 +1963,8 @@ async function update_endpoint_stats(req) {
     ]);
 }
 
-/**
- * @param {nb.ObjectMD} obj
- */
```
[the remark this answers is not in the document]
alphaprinz: Nope, thanks :)

src/sdk/object_io.js
alphaprinz: Ok, thanks.

src/server/object_services/object_server.js
alphaprinz: Ok. I think I got it, but please review.

guymguym (changes requested):
From code perspective I think this looks good.

The main concern I have left is to test how this works in namespace_cache where we also use the upload flows to write partial ranges to the cache bucket.

Please verify that this is covered and passing by -
https://github.com/noobaa/noobaa-core/blob/master/src/test/unit_tests/test_namespace_cache.js

nimrod-becker:
@alphaprinz any chance of looking at what Guy suggested in the last comment?

alphaprinz:
Hi @nimrod-becker and @guymguym 
Circling back to this, I admit I don't remember all the details, but I did find in slack from 20/6 that I've checked and test_case_range_read_range_variations() in namespace_cache_tests_list.js. is the scenario Guy mentioned.

PR #7463, opened 2023-08-28
NSFS | moving util functions to native_fs_utils.js file and some cleanup

romayalon:
### Explain the changes
1. On nb.d NativeFSContext backend is 'GPFS',  I suggest change it to be a string
2. Removed secret logging
3. Started moving some unbounded util functions to native_fs_utils.js
4. Converted hard coded 'null' to NULL_VERSION_ID variable
5. Re-used is_versioning_enables/suspended functions
6. Added some missing parameteres to NamespaceFS object on test_namespace_fs.js
 

### Issues: Fixed #xxx / Gap #xxx
1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

PR #7488, opened 2023-09-11
Object Versions Replication - NamespaseS3 fix

liranmauda:
### Explain the changes
As we are adding a field  `noobaa-namespace-s3-bucket` MetaData in the `Metadata` with the bucket name, we will always replicate it. in order to avoid this, we will omit this added field.

### Testing Instructions:
1. Enable versioning on 2 aws buckets
2. Write a few different versions for the same object on the first bucket
3. Create 2 namespace buckets pointing to those 2 aws buckets
4. Create a replication between those buckets with syc.versions flag on
5. See that the replication is working as expected

### DIFF
--- a/src/util/cloud_utils.js
+++ b/src/util/cloud_utils.js
@@ -170,7 +170,6 @@ function set_noobaa_s3_connection(sys) {
     const endpoint = system_address[0] && system_address[0].hostname;
     const access_key = sys.owner && sys.owner.access_keys && sys.owner.access_keys[0].access_key.unwrap();
     const secret_key = sys.owner && sys.owner.access_keys && sys.owner.access_keys[0].secret_key.unwrap();
-    dbg.log0('replication_server.set_noobaa_s3_connection: ', endpoint, access_key, secret_key);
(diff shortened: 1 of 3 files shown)


src/util/cloud_utils.js
```diff
@@ -170,7 +170,7 @@ function set_noobaa_s3_connection(sys) {
     const endpoint = system_address[0] && system_address[0].hostname;
     const access_key = sys.owner && sys.owner.access_keys && sys.owner.access_keys[0].access_key.unwrap();
     const secret_key = sys.owner && sys.owner.access_keys && sys.owner.access_keys[0].secret_key.unwrap();
-    dbg.log0('replication_server.set_noobaa_s3_connection: ', endpoint, access_key, secret_key);
+    dbg.log1('replication_server.set_noobaa_s3_connection: ', endpoint, access_key, secret_key);
```
romayalon: can you delete the secret_key from this print? I remove it in https://github.com/noobaa/noobaa-core/pull/7463 but it's was not code reviewed yet

romayalon (approved):

PR #7494, opened 2023-09-14
DRAFT ssl cert reload - run in endpoint bg (bz 2237903)

alphaprinz:
### Explain the changes
The certificate for S3 endpoint can change.
Endpoint now periodically checks whether certificate file has changed.
If so, the https servers are updated with new certificate.
(Endpoint pod doesn't need to be restarted).

https://bugzilla.redhat.com/show_bug.cgi?id=2237903

### Testing Instructions:
Note currently used certificate, eg-
openssl s_client -connect localhost:10443 -showcerts 2>/dev/null </dev/null | sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p' | openssl x509 -text -noout

Create a new certificate, eg-
openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=CommonNameOrHostname"

From here, follow instructions in https://github.com/noobaa/noobaa-operator/blob/master/doc/ssl-dns-routing.md:
Delete current secret (if exists), eg-
kubectl delete secret noobaa-s3-serving-cert
Create a new secret from the new certificate. Note filenames must be tls.crt and tls.key-
kubectl create secret generic  noobaa-s3-serving-cert --from-file=tls.crt --from-file=tls.key
Wait until new files are created in endpoint's pod /etc/s3-secret.
Wait until background worker runs.
Running openssl -showcerts now should have the new certificate.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/cert_worker.js
+++ b/src/endpoint/cert_worker.js
@@ -1,0 +1,44 @@
+/* Copyright (C) 2023 NooBaa */
+'use strict';
+
+const dbg = require('../util/debug_module')(__filename);
+const config = require('../../config.js');
+const ssl_utils = require('../util/ssl_utils');
+
+/**
+ * Updates s3 https server(s) when certificate files are updated.
+ * Note ODF rotates certificates every 13 months.
+ * Some more details in operator's doc/ssl-dns-routing.md -
+ * https://github.com/noobaa/noobaa-operator/blob/master/doc/ssl-dns-routing.md
+ */
+class CertWorker {
+
+    /**
+     * @param {{
+     *   name: string;
+     *   https_server: https.Server;
+     *   sts_server: https.Server;
+     * }} params
+     */
+    constructor({ name, https_server, sts_server }) {
+        this.name = name;
+        this.https_server = https_server;
+        this.sts_server = sts_server;
+    }
+
+    async run_batch() {
+        dbg.log2('certificate bg worker: running check for certificate update');
+        if (await ssl_utils.update_certs_from_disk()) {
+            //certificate were updated. give new cert to https servers.
+            dbg.log0('certificate bg worker: updating certs');
+            const ssl_cert = await ssl_utils.get_ssl_certificate('S3');
+            const ssl_options = { ...ssl_cert, honorCipherOrder: true };
+            this.https_server.setSecureContext(ssl_options);
+            this.sts_server.setSecureContext(ssl_options);
+        }
+        //use same delay as core
+        return config.CLUSTER_HB_INTERVAL;
+    }
+}
+
+exports.CertWorker = CertWorker;
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -41,6 +41,8 @@ const background_scheduler = require('../util/background_scheduler').get_instanc
 const endpoint_stats_collector = require('../sdk/endpoint_stats_collector');
 const { NamespaceMonitor } = require('../server/bg_services/namespace_monitor');
 const { SemaphoreMonitor } = require('../server/bg_services/semaphore_monitor');
+const { CertWorker } = require('./cert_worker');
+
 
 if (process.env.NOOBAA_LOG_LEVEL) {
     const dbg_conf = debug_config.get_debug_config(process.env.NOOBAA_LOG_LEVEL);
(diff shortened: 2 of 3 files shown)


src/endpoint/endpoint.js
```diff
@@ -41,6 +41,8 @@ const background_scheduler = require('../util/background_scheduler').get_instanc
 const endpoint_stats_collector = require('../sdk/endpoint_stats_collector');
 const { NamespaceMonitor } = require('../server/bg_services/namespace_monitor');
 const { SemaphoreMonitor } = require('../server/bg_services/semaphore_monitor');
+const { CertWorker } = require('./cert_worker');
```
jackyalbo, line 44: let's keep in the same place with all the bg_workers... or at least under endpoint/bg_services/
but maybe we will use it in core as well in the future

src/endpoint/cert_worker.js
```diff
@@ -0,0 +1,44 @@
...
+        if (await ssl_utils.update_certs_from_disk()) {
+            //certificate were updated. give new cert to https servers.
+            dbg.log0('certificate bg worker: updating certs');
+            const ssl_cert = await ssl_utils.get_ssl_certificate('S3');
+            const ssl_options = { ...ssl_cert, honorCipherOrder: true };
+            this.https_server.setSecureContext(ssl_options);
+            this.sts_server.setSecureContext(ssl_options);
+        }
+        //use same delay as core
+        return config.CLUSTER_HB_INTERVAL;
```
jackyalbo, line 40: I think we should maybe use a different one than HeartBeat. I think once a day should be enough no? maybe with error handling retry it more frequently(1 min?) CERT_WORKER_DELAY and CERT_WORKER_ERROR_DELAY.

src/endpoint/cert_worker.js
```diff
@@ -0,0 +1,44 @@
...
+     * }} params
+     */
+    constructor({ name, https_server, sts_server }) {
+        this.name = name;
+        this.https_server = https_server;
+        this.sts_server = sts_server;
+    }
+
+    async run_batch() {
+        dbg.log2('certificate bg worker: running check for certificate update');
```
jackyalbo, line 30: let keep the format we usually use: 
```suggestion
        dbg.log2('cert_worker: running check for certificate update');
```

src/endpoint/cert_worker.js
```diff
@@ -0,0 +1,44 @@
...
+        if (await ssl_utils.update_certs_from_disk()) {
+            //certificate were updated. give new cert to https servers.
+            dbg.log0('certificate bg worker: updating certs');
```
jackyalbo, line 33: here as well:
```suggestion
            dbg.log0('cert_worker: updating certs');
```

src/endpoint/endpoint.js
```diff
@@ -190,6 +193,12 @@ async function main(options = {}) {
             }));
         }
 
+        background_scheduler.register_bg_worker(new CertWorker({
+            name: 'certificate_worker',
```
jackyalbo, line 197: let's decide on one name. cert_worker or ceritificate_worker both works but let's use one every where :)

src/endpoint/cert_worker.js
alphaprinz: That's what Ohad did initially.
Also if user deletes the old certificates manually (because they were hacked and can't be used anymore) you don't want to wait a day for new certificate to kick in.

jackyalbo: OK, so it can be lower (half an hour?|)... just think the 2 shouldn't be connected and it's confusing. also maybe it should also have an enabled/disabled config in case it's self-gen and we are not interested in it at all. I think the user can always delete the pod in case of an emergency.

dannyzaken:
following a discussion with @alphaprinz, we can also consider using `fs.watch` to identify any changes. the downside is to handle cases where the containing dif does not exist. we can maybe create it initially in the docker file to solve that. @jackyalbo WDYT?

jackyalbo:
> following a discussion with @alphaprinz, we can also consider using `fs.watch` to identify any changes. the downside is that it will not work when the containing dir does not exist. we can maybe create it initially in the docker file to solve that. @jackyalbo WDYT?
you want to fs.watch the dir? maybe we can make k8s create it even if the secret don't exist (not sure if it's possible - or maybe the secret always exist - just empty - again not sure)
also we maybe want to use same solution to both core and endpoint regarding rotation. So once we find a solution we should maybe move core to it as well and outside of server_monitor

alphaprinz:
Closed, we will got with fs.watch() alternative https://github.com/noobaa/noobaa-core/pull/7502.

PR #7502, opened 2023-09-20
Reload certs 2237903 watch

alphaprinz:
### Explain the changes
The certificate for S3 endpoint can change.
Endpoint now watches for changes in certificate file.
If so, the https servers are updated with new certificate.
(Endpoint pod doesn't need to be restarted).

(https://bugzilla.redhat.com/show_bug.cgi?id=2237903)

### Testing Instructions:
Note currently used certificate, eg-
openssl s_client -connect localhost:10443 -showcerts 2>/dev/null </dev/null | sed -ne '/-BEGIN CERTIFICATE-/,/-END CERTIFICATE-/p' | openssl x509 -text -noout

Create a new certificate, eg-
openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -sha256 -days 3650 -nodes -subj "/C=XX/ST=StateName/L=CityName/O=CompanyName/OU=CompanySectionName/CN=CommonNameOrHostname"

From here, follow instructions in https://github.com/noobaa/noobaa-operator/blob/master/doc/ssl-dns-routing.md:
Delete current secret (if exists), eg-
kubectl delete secret noobaa-s3-serving-cert
Create a new secret from the new certificate. Note filenames must be tls.crt and tls.key-
kubectl create secret generic noobaa-s3-serving-cert --from-file=tls.crt --from-file=tls.key
Wait until new files are created in endpoint's pod /etc/s3-secret.
Wait until background worker runs.
Running openssl -showcerts now should have the new certificate.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -81,6 +81,7 @@ dbg.log0('endpoint: replacing old umask: ', old_umask.toString(8), 'with new uma
 /**
  * @param {EndpointOptions} options
  */
+/* eslint-disable max-statements */
 async function main(options = {}) {
     try {
         // the primary just forks and returns, workers will continue to serve
@@ -145,11 +146,17 @@ async function main(options = {}) {
         const endpoint_request_handler = create_endpoint_handler(init_request_sdk, virtual_hosts);
         const endpoint_request_handler_sts = create_endpoint_handler(init_request_sdk, virtual_hosts, true);
 
-        const ssl_cert = await ssl_utils.get_ssl_certificate('S3');
-        const ssl_options = { ...ssl_cert, honorCipherOrder: true };
+        const ssl_cert_info = await ssl_utils.get_ssl_cert_info('S3');
+        const ssl_options = { ...ssl_cert_info.cert, honorCipherOrder: true };
         const http_server = http.createServer(endpoint_request_handler);
         const https_server = https.createServer(ssl_options, endpoint_request_handler);
         const https_server_sts = https.createServer(ssl_options, endpoint_request_handler_sts);
+        ssl_cert_info.on('update', updated_ssl_cert_info => {
+            dbg.log0("Setting updated S3 ssl certs for endpoint.");
+            const updated_ssl_options = { ...updated_ssl_cert_info.cert, honorCipherOrder: true };
+            https_server.setSecureContext(updated_ssl_options);
+            https_server_sts.setSecureContext(updated_ssl_options);
+        });
 
         if (http_port > 0) {
             dbg.log0('Starting S3 HTTP', http_port);
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -1393,6 +1393,9 @@ class PostgresClient extends EventEmitter {
             this.pool.end();
             this.pool = null;
         }
+        if (this.ssl_cert_info) {
+            this.ssl_cert_info.removeListener(this._update_ssl_cert);
+        }
     }
 
     async reconnect() {
@@ -1633,9 +1636,13 @@ class PostgresClient extends EventEmitter {
     }
 
     async _load_ssl_certs() {
-        const ssl_cert = await ssl_utils.get_ssl_certificate('EXTERNAL_DB') || {};
+        this.ssl_cert_info = await ssl_utils.get_ssl_cert_info('EXTERNAL_DB') || {};
         /** @type {import('tls').ConnectionOptions} */
-        this.new_pool_params.ssl = { ...ssl_cert, rejectUnauthorized: !process.env.POSTGRES_SSL_UNAUTHORIZED };
+        this.new_pool_params.ssl = { ...this.ssl_cert_info.cert, rejectUnauthorized: !process.env.POSTGRES_SSL_UNAUTHORIZED };
+        this.ssl_cert_info.on('update', () => {
+            dbg.log("EXTERNAL_DB ssl certs have changed. Reconnecting.");
+            this.reconnect();
+        });
         // this will mask out unwanted prints of user certificate
         this.print_pool_params.ssl = { rejectUnauthorized: !process.env.POSTGRES_SSL_UNAUTHORIZED };
     }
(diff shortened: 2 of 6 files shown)


src/endpoint/endpoint.js
```diff
@@ -190,6 +192,28 @@ async function main(options = {}) {
             }));
         }
 
+        fs.watch(ssl_utils.get_cert_dir('S3'), {},
+            (event_type, filename) => {
+                ssl_utils.update_certs_from_disk().then(is_updated => {
```
jackyalbo, line 197: Any reason not to use async/await? We are trying to avoid using promise chains.
would also prefer this function to be outside of fs.watch. maybe with cert name as a parameter as I will need this code also for EXTERNAL_DB certificates

alphaprinz: When I try to use await instead of then I get that await can only be use in top level.

The new version has this new functionality inside ssl_utils.

src/endpoint/endpoint.js
```diff
@@ -145,11 +146,17 @@ async function main(options = {}) {
...
 
-        const ssl_cert = await ssl_utils.get_ssl_certificate('S3');
-        const ssl_options = { ...ssl_cert, honorCipherOrder: true };
+        const ssl_cert_info = await ssl_utils.get_ssl_cert_info('S3');
+        const ssl_options = { ...ssl_cert_info.cert, honorCipherOrder: true };
         const http_server = http.createServer(endpoint_request_handler);
         const https_server = https.createServer(ssl_options, endpoint_request_handler);
         const https_server_sts = https.createServer(ssl_options, endpoint_request_handler_sts);
+        ssl_cert_info.on('update', updated_ssl_cert_info => {
+            dbg.log0("endpoint updating cert for https_servers.");
```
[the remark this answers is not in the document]
jackyalbo: maybe change to something like: "S3/MGMT/EXTERNAL_DB certs dir changes. updating cert for endpoint/web_server/RPC/postgres_client"
and the same for other directories.

src/util/ssl_utils.js
```diff
@@ -71,40 +94,19 @@ function get_ssl_certificate(service) {
...
-                    dbg.warn(`SSL certificate failed to update from dir ${cert_info.dir}:`, err.message);
-                }
-                return false;
+        try {
+            fs.watch(cert_info.dir, {}, cert_info.file_notification.bind(cert_info));
+        } catch (err) {
+            if (err.code === 'ENOENT') {
+                dbg.warn("Certificate folder ", cert_info.dir, " does not exist. New certificate won't be loaded.");
+            } else {
+                dbg.error("Failed to watch certificate dir ", cert_info.dir);
```
jackyalbo, line 104: print the error.

src/util/ssl_utils.js
```diff
@@ -71,40 +94,19 @@ function get_ssl_certificate(service) {
...
-            } catch (err) {
-                if (err.code !== 'ENOENT') {
-                    dbg.warn(`SSL certificate failed to update from dir ${cert_info.dir}:`, err.message);
-                }
-                return false;
+        try {
+            fs.watch(cert_info.dir, {}, cert_info.file_notification.bind(cert_info));
+        } catch (err) {
+            if (err.code === 'ENOENT') {
+                dbg.warn("Certificate folder ", cert_info.dir, " does not exist. New certificate won't be loaded.");
```
jackyalbo, line 102: another reason to maybe check for the directory in advance and to handle this error there. although not sure we have this split, the level of the print?

src/util/ssl_utils.js
```diff
@@ -9,18 +9,45 @@ const Semaphore = require('../util/semaphore');
...
+            this.is_generated = false;
+
+            for (const an_update_cb of this.update_cbs) {
+                dbg.log0('about to invoke = ', an_update_cb, ", type of update_cb_ = ", an_update_cb);
+                an_update_cb(cert_on_disk);
+            }
+        } catch (err) {
+            if (err.code !== 'ENOENT') {
+                dbg.warn(`SSL certificate failed to update from dir ${this.dir}:`, err.message);
+            }
```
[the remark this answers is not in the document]
alphaprinz: Yes. Such errors are logged.

src/util/postgres_client.js
```diff
@@ -1633,7 +1633,10 @@ class PostgresClient extends EventEmitter {
     }
 
     async _load_ssl_certs() {
-        const ssl_cert = await ssl_utils.get_ssl_certificate('EXTERNAL_DB') || {};
+        const ssl_cert = await ssl_utils.get_ssl_certificate('EXTERNAL_DB', new_cert => {
+            this.new_pool_params.ssl = { ...new_cert, rejectUnauthorized: !process.env.POSTGRES_SSL_UNAUTHORIZED };
```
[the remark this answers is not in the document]
alphaprinz: This is out-of-scope.
I've just adapted the existing code to be notified when certs are updated.

src/util/ssl_utils.js
```diff
@@ -71,40 +94,19 @@ function get_ssl_certificate(service) {
...
-                cert_info.is_generated = false;
-                return true;
-
-            } catch (err) {
-                if (err.code !== 'ENOENT') {
-                    dbg.warn(`SSL certificate failed to update from dir ${cert_info.dir}:`, err.message);
-                }
-                return false;
+        try {
+            fs.watch(cert_info.dir, {}, cert_info.file_notification.bind(cert_info));
```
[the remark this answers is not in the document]
alphaprinz: Personally I like that we can still update self-generated certs to "real" certs in containerized env, though I admit it's not a practical scenario.

At any rate, this is the rule-of-thumb I followed-
For containerized env, the directory should always exists.
Non-containerized envs are not relevant for this PR.

src/util/ssl_utils.js
alphaprinz: I don't understand.
This is the only place non-existing directory is handled.

If you think there's still a problem despite what I wrote in previous comment (https://github.com/noobaa/noobaa-core/pull/7502#discussion_r1356202750), let me know.

jackyalbo (approved):
Approving Under the assumptions made in the comments

PR #7503, opened 2023-09-20
Namespace GCP | Adding read_object_stream

liranmauda:
### Explain the changes
Namespace GCP | Adding read_object_stream


jackyalbo (approved):
The code itself looks good so approving. This code will be later tested and/or re-written.
I think maybe we should add a comment that this code is currently experimental.

PR #7510, opened 2023-09-27
NSFS | NC | bucket and account manage CLI

naveenpaul1:
### Explain the changes
1. Users can create, update, delete, and list buckets and accounts using CLI 

1. https://issues.redhat.com/browse/MCGI-197

### Testing Instructions:
1. Please refer the doc : non_containerized_NSFS.md

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -1,0 +1,455 @@
+/* Copyright (C) 2020 NooBaa */
+'use strict';
+
+const fs_utils = require('../util/fs_utils');
+const SensitiveString = require('../util/sensitive_string');
+const minimist = require('minimist');
+const path = require('path');
+const fs = require('fs');
+const P = require('../util/promise');
+const _ = require('lodash');
+const config = require('../../config');
+
+const HELP = `
+Help:
+
+    "nsfs" is a noobaa-core command runs a local S3 endpoint on top of a filesystem.
+    Each sub directory of the root filesystem represents an S3 bucket.
+    manage nsfs will provide a command line interface to create new accounts and map existing directories 
+    to Noobaa as buckets. For more information refer to the noobaa docs.
+`;
+
+const USAGE = `
+Usage:
+
+    node src/cmd/manage_nsfs <action> <type> [options...]
+`;
+
+const ARGUMENTS = `
+Arguments:
+
+    <type>    Set the resource type such as accounts and buckets
+    <action>  Action could be add, update, list, status and delete for accounts/buckets.
+`;
+
+const ACCOUNT_OPTIONS = `
+Account Options:
+
+    # Read account details from json file, no need to mention all the propery one by one in CLI
+    --from_file <dir>           (default none)          Set buckt/account schema full path.
+                                                        Get bucket and account details from json file                
+
+    # required for add, replace 
+    --name <name>               (default none)          Set the name for account/bucket.
+    --email <email>             (default none)          Set the email for account/bucket.
+    --uid <uid>                 (default as process)    Send requests to the Filesystem with uid.
+    --gid <gid>                 (default as process)    Send requests to the Filesystem with gid.
+    --secret_key <key>          (default none)          The secret key pair for the access key.
+    --new_buckets_path <dir>    (default none)          Set the root of the filesystem where each subdir is a bucket.
+    
+    # required for add, replace and delete
+    --access_key <key>          (default none)                                     Authenticate incoming requests for this access key only (default is no auth).
+    --config_root <dir>         (default config.NSFS_NC_DEFAULT_CONF_DIR)          Configuration files path for Noobaa standalon NSFS.
+`;
+
+const BUCKET_OPTIONS = `
+Bucket Options:
+
+    # Read Bucket details from json file, no need to mention all the propery one by one in CLI
+    --from_file <dir>           (default none)          Set buckt/account schema full path.
+                                                        Get bucket and account details from json file
+    # required for add, replace 
+    --email <email>             (default none)          Set the email for account/bucket.
+    --path <dir>                (default none)          Set the bucket path.
+
+    # required for add, replace and delete
+    --name <name>               (default none)          Set the name for account/bucket.
+    --config_root <dir>         (default config.NSFS_NC_DEFAULT_CONF_DIR)          Configuration files path for Noobaa standalon NSFS.
+`;
+
+function print_account_usage() {
+    console.warn(HELP);
+    console.warn(USAGE.trimStart());
+    console.warn(ARGUMENTS.trimStart());
+    console.warn(ACCOUNT_OPTIONS.trimStart());
+    process.exit(1);
+}
+
+function print_bucket_usage() {
+    console.warn(HELP);
+    console.warn(USAGE.trimStart());
+    console.warn(ARGUMENTS.trimStart());
+    console.warn(BUCKET_OPTIONS.trimStart());
+    process.exit(1);
+}
+
+async function check_and_create_config_dirs(config_root) {
+    const bucket_dir_stat = await fs_utils.file_exists(config_root);
+    if (!bucket_dir_stat) {
+        await fs_utils.create_path(config_root);
+        await fs_utils.create_path(config_root + '/buckets');
+        await fs_utils.create_path(config_root + '/accounts');
+        return true;
+    }
+    return false;
+}
+
+async function main(argv = minimist(process.argv.slice(2))) {
+    try {
+        const resources_type = argv._[0] || '';
+        if ((argv.help || argv.h) && resources_type === 'account') return print_account_usage();
+        if ((argv.help || argv.h) && resources_type === 'bucket') return print_bucket_usage();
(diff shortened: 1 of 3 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,411 @@
...
+    try {
+        const resourse_type = argv._[0] || '';
+        if ((argv.help || argv.h) && resourse_type === 'account') return print_account_usage();
+        if ((argv.help || argv.h) && resourse_type === 'bucket') return print_bucket_usage();
+        if (argv.debug) {
+            const debug_level = Number(argv.debug) || 5;
+            dbg.set_module_level(debug_level, 'core');
+            nb_native().fs.set_debug_level(debug_level);
+        }
+        const config_root = argv.config_root ? String(argv.config_root) : '';
```
romayalon, line 108:
```suggestion
        const config_root = argv.config_root ? String(argv.config_root) : config.NSFS_NC_DEFAULT_CONF_DIR;
```

romayalon: it's not in master yet but I added /etc/noobaa.conf.d/ as config.NSFS_NC_DEFAULT_CONF_DIR

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,411 @@
...
+            console.log("Bucket exists with access key : " + data.name);
+        } else {
+            console.log("Bucket do not exists with access key : " + data.name);
+        }
+    } else if (action === 'update') {
+        const is_valid = await validate_bucket_add_args(data);
+        if (!is_valid) {
+            print_bucket_usage();
+        }
+        create_resourse(data.name, bucket_config_path, data);
```
romayalon, line 209: why do we create the resource on update? also I saw the on fetch we update the file

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,411 @@
...
+        data.name = new SensitiveString(data.name);
+    }
+    if (!data) {
+        const bucket_path = argv.path ? String(argv.path) : '';
+        data = {
+            name: name,
+            system_owner: email,
+            bucket_owner: email,
+            tag: "",
+            versioning: "DISABLED",
```
romayalon, line 156: let's replace all double quotes (") to single quotes (')

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,411 @@
...
+        print_account_usage();
+    }
+    const data = _.merge({}, source, target);
+    return data;
+}
+
+async function manage_bucket_operations(action, data, config_root) {
+    const bucket_config_path = config_root + '/buckets';
+    if (action === 'add') {
+        const is_valid = await validate_bucket_add_args(data);
```
romayalon, line 186: I would suggest to split this function to multiple functions, each handling an action

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,411 @@
...
+    const action = argv._[1] || '';
+    const config_root = argv.config_root ? String(argv.config_root) : '';
+    const data = await fetch_account_data(argv);
+    manage_account_operations(action, data, config_root);
+}
+
+async function fetch_account_data(argv) {
+    let data;
+    const action = argv._[1] || '';
+    const from_file = argv.from_file ? String(argv.from_file) : '';
```
romayalon, line 234: it looks like from_file and config_root assignment can be done once in main, WDYT? 
const config_root = argv.config_root ? String(argv.config_root) : ''; is done many times and I don't think it's necassary

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,411 @@
...
+    } else if (action === 'list') {
+        const accounts = await list_resources(account_path);
+        console.log(accounts);
+    } else {
+        console.log("Account action not found.");
+        print_account_usage();
+    }
+}
+
+async function create_resourse(file_name, file_path, data) {
```
romayalon, line 333: I would suggest to change the create_resource to create_config_file(), resource has a different meaning in NooBaa and I find it a bit confusing

src/cmd/manage_nsfs.js
naveenpaul1: @romayalon its calling method `await fs_utils.replace_file(full_path, data);`. It creates the file if the file missing and replaces already exists. On fetch we we are just getting data from the config file and updating object with input data.

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,452 @@
...
+    # Read Bucket details from json file, no need to mention all the propery one by one in CLI
+    --from_file <dir>           (default none)          Set buckt/account schema full path.
+                                                        Get bucket and account details from json file
+    # required for add, replace 
+    --email <email>             (default none)          Set the email for account/bucket.
+    --path <dir>                (default none)          Set the bucket path.
+
+    # required for add, replace and delete
+    --name <name>               (default none)          Set the name for account/bucket.
+    --config_root <dir>         (default none)          Configuration files path for Noobaa standalon NSFS.
```
romayalon, line 69:
```suggestion
    --config_root <dir>         (default config.NSFS_NC_DEFAULT_CONF_DIR)          Configuration files path for Noobaa standalon NSFS.
```

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,452 @@
...
+    }
+    return false;
+}
+
+async function main(argv = minimist(process.argv.slice(2))) {
+    try {
+        const resourse_type = argv._[0] || '';
+        if ((argv.help || argv.h) && resourse_type === 'account') return print_account_usage();
+        if ((argv.help || argv.h) && resourse_type === 'bucket') return print_bucket_usage();
+        if (argv.debug) {
```
romayalon, line 104: so, do we support updating the debug level through this CLI? might be a good idea, we just need it documented and tested

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,452 @@
...
+
+async function get_bucket_config_file_status(data, bucket_config_path) {
+    const is_valid = await validate_minimum_bucket_args(data);
+    if (!is_valid) {
+        print_account_usage();
+        return;
+    }
+    const bucket_dir_stat = await fs_utils.file_exists(path.join(bucket_config_path, data.name + '.json'));
+    if (bucket_dir_stat) {
+        console.log('Bucket exists with access key : ' + data.name);
```
romayalon, line 198: why not print all the bucket data?

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,452 @@
...
+    const full_path = path.join(file_path, file_name.unwrap() + '.json');
+    await fs_utils.replace_file(full_path, data);
+    console.log('Resourse created : ', file_name.unwrap());
+}
+
+async function update_config_file(file_name, file_path, data) {
+    data = JSON.stringify(data);
+    const full_path = path.join(file_path, file_name.unwrap() + '.json');
+    await fs_utils.replace_file(full_path, data);
+    console.log('Resourse updated : ', file_name.unwrap());
```
romayalon, line 373:
```suggestion
    console.log('Resource updated : ', file_name.unwrap());
```

romayalon: please replace it all over the file

src/cmd/manage_nsfs.js
```diff
@@ -0,0 +1,452 @@
...
+    --new_buckets_path <dir>    (default none)          Set the root of the filesystem where each subdir is a bucket.
+    
+    # required for add, replace and delete
+    --access_key <key>          (default none)          Authenticate incoming requests for this access key only (default is no auth).
+    --config_root <dir>         (default none)          Configuration files path for Noobaa standalon NSFS.
+`;
+
+const BUCKET_OPTIONS = `
+Bucket Options:
+
```
romayalon, line 59: I think it's important here to remind the user that create bucket cli command won't create the directory and the bucket directory should exist prior to the create bucket cli command

naveenpaul1: @romayalon but we are creating the bucket for the user from cli in method `check_and_create_config_dirs()`, Should I remove this?

src/cmd/manage_nsfs.js
naveenpaul1: @romayalon When I thought about it, it doesn't look good to have debug level changes in this CLI. WDYT? Should I remove it?

src/cmd/manage_nsfs.js
naveenpaul1: @romayalon oh.. ok, I got confused it with config_root dir, Yes I will mention about bucke dir in doc, Thank you

src/cmd/manage_nsfs.js
romayalon: you can remove it from this PR and we should think how to let users do that

romayalon (approved):

src/cmd/manage_nsfs.js
naveenpaul1: done

PR #7511, opened 2023-09-27
increase timeout for bucket policy unit tests

nadavMiz:
### Explain the changes
1. increase the timeout for unit tests that are failing on exceeded timeout.
2. timeout was increased to 15000, this number was chosen based on timeout that was set for similar tests in the same file 

1. 

### Testing Instructions:
1. run unit tests for bucket policy. tests should pass

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/test/unit_tests/test_s3_bucket_policy.js
+++ b/src/test/unit_tests/test_s3_bucket_policy.js
@@ -192,6 +192,8 @@ mocha.describe('s3_bucket_policy', function() {
     });
 
     mocha.it('should be able to set bucket policy when none set', async function() {
+        const self = this; // eslint-disable-line no-invalid-this
+        self.timeout(15000);
         const policy = {
             Version: '2012-10-17',
             Statement: [{
@@ -651,6 +653,8 @@ mocha.describe('s3_bucket_policy', function() {
     });
 
     mocha.it('should be able to deny based on server side encryption', async function() {
+        const self = this; // eslint-disable-line no-invalid-this
+        self.timeout(15000);
         const auth_put_policy = {
             Version: '2012-10-17',
             Statement: [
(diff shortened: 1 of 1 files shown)


nadavMiz:
some of the bucket policy sometimes fails in the CI for timeout exceeded. this is an example of the issue observed in the logs:

```
2023-09-26T15:46:31.1166445Z   3421 passing (23m)
2023-09-26T15:46:31.1177169Z   18 pending
2023-09-26T15:46:31.1188507Z   1 failing
2023-09-26T15:46:31.1199752Z
2023-09-26T15:46:31.1224495Z   1) s3_bucket_policy
2023-09-26T15:46:31.1224830Z        should be able to set bucket policy when none set:
2023-09-26T15:46:31.1225857Z      Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/root/node_modules/noobaa-core/src/test/unit_tests/index.js)
2023-09-26T15:46:31.1226417Z       at listOnTimeout (node:internal/timers:569:17)
2023-09-26T15:46:31.1226780Z       at process.processTimers (node:internal/timers:512:7)
```bash

File: src/test/unit_tests/test_s3_bucket_policy.js [hunk 1]
```diff
@@ -192,6 +192,8 @@ mocha.describe('s3_bucket_policy', function() {
     });
 
     mocha.it('should be able to set bucket policy when none set', async function() {
+        const self = this; // eslint-disable-line no-invalid-this
```
naveenpaul1, line 195: why we can not use this.timeout()? lint error?

shirady (approved):
LGTM

src/test/unit_tests/test_s3_bucket_policy.js
nadavMiz: both ways cause lint errors. other tests in this file use ```const self = this``` so I did it this way to be consistent with the other tests

naveenpaul1 (approved):

PR #7517, opened 2023-10-01
NSFS | Fix nsfs simple fs tool

romayalon:
### Explain the changes
1. A few functions like bucket policy etc were commented out due to the recent NC changes, also the bucket policy structure was updated to Upper case on master so I fixed it.

1. 

### Testing Instructions:
1. Open 2 termial tabs - 
2. Tab 1 - run sudo node src/cmd/nsfs /path/to/fs_root/ --simple
3. Tab 2 - run s3 commands to localhost:6443/6001

- [ ] Doc added/updated
- [ ] Tests added


alphaprinz (approved):
LGTM

PR #7518, opened 2023-10-02
NSFS | NC | nsfs service additions

romayalon:
### Explain the changes
1. Added needed env vars to nsfs service
2. Added a check when creating fs_napi for GPFS lib file existence before reading the lib file - and added this check boolean in use_gpfs_lib().

1. https://issues.redhat.com/browse/MCGI-201

### Testing Instructions:
1. Run NC nsfs with the GPFS path as env variable while lib file exists
2. Run NC nsfs with the GPFS path as env variable while lib file does not exist
3. Run NC nsfs without the GPFS path as env variable

- [ ] Doc added/updated
- [ ] Tests added


src/deploy/nsfs.service
```diff
@@ -1,12 +1,15 @@
 [Unit]
 Description=The NooBaa nsfs service.
+After=gpfs-wait-mount.service
```
tangledbytes, line 3: Will this prevent the service to start if `gpfs-wait-mount.service` does not exist?

src/deploy/nsfs.service
```diff
@@ -1,12 +1,15 @@
...
 
 [Service]
 Restart=always
 RestartSec=2
 User=root
 Group=root
-Environment=
+Environment=ENDPOINT_FORKS=0
```
tangledbytes, line 10: Should we also allow another place to configure this, like `system.json` (eg. `"forks": 8`) so that consumers of this service don't have to edit the unit file to configure the forks?

romayalon: I think that we can specify an .env file instead under the config dir, WDYT?

src/deploy/nsfs.service
romayalon: no, it will work event without it

src/deploy/nsfs.service
tangledbytes: On a second thought, I think `config-local.js` is good enough for that purpose.

tangledbytes: > I think that we can specify a .env file instead under the config dir, WDYT?
Makes sense!

tangledbytes (approved):

naveenpaul1 (approved):

romayalon:
@tangledbytes I added a default nsfs_env.env file as an example that should be linked to /etc/noobaa.conf.d/.env
and added it as an environmentFile to be used in nsfs.service, please re-check it

src/deploy/nsfs.service
```diff
@@ -1,13 +1,14 @@
...
+After=gpfs-wait-mount.service
 
 [Service]
 Restart=always
 RestartSec=2
 User=root
 Group=root
-Environment=
 ExecStart=/usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/nsfs.js
+EnvironmentFile=/etc/noobaa.conf.d/.env
```
tangledbytes, line 11: We already have support for `$NOOBAA_DIR/.env` file though. Should we document somewhere that these two `.env` files are different and that `$NOOBAA_DIR/.env` will always supersede `/etc/noobaa.conf.d/.env` one?

PR #7522, opened 2023-10-04
NSFS | NC | Atomic creation of buckets and accounts config files (bucketspace FS and CLI)

romayalon:
### Explain the changes
1. 

1. https://issues.redhat.com/browse/MCGI-200

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -9,6 +9,7 @@ const fs = require('fs');
 const P = require('../util/promise');
 const _ = require('lodash');
 const config = require('../../config');
+const native_fs_utils = require('../util/native_fs_utils');
 
 const HELP = `
 Help:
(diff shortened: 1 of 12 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -168,18 +178,26 @@ async function update_bucket_object(config_root, target) {
         print_account_usage();
     }
     const data = _.merge({}, source, target);
-    data.path = source.path;
+    // TODO: ask Naveen why this line was added
+    //data.path = source.path;
```
romayalon, line 182: Hey @naveenpaul1 I saw that that source path is always taken, why is that? a user won't be able to update path according to this line

naveenpaul1: @romayalon Idea is source(new json schema for update) should be able to update all properties in the existing bucket,  `_.merge()`  was not updating the `path` on its own, So I have to do it in code explicitly.

romayalon: but this line - data.path = source.path; won't update the data with a new path, it takes the source and not the new path. or am I missing something?
and thanks i'll update the print

naveenpaul1: @romayalon got your point and you are correct, This code takes the same path from the existing bucket schema not using new path. It should be `target.path`.

src/cmd/manage_nsfs.js
```diff
@@ -420,13 +429,9 @@ async function validate_bucket_add_args(data) {
         console.error('Error: User data and bucket path should not be empty');
         return false;
     }
-    if (!data.path.includes(data.name.unwrap())) {
```
naveenpaul1, line 423: why this check removed?

naveenpaul1 (approved):
LGTM

src/cmd/manage_nsfs.js
romayalon: there is no suck restriction - to have the bucket name in the path of the exported directory

PR #7524, opened 2023-10-04
Added a check for bidirectional replication for matching prefix

aayushchouhan09:
### Explain the changes
1. Replication Validation - Block bi-directional replication for matching prefixes

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -1899,7 +1899,10 @@ async function put_bucket_replication(req) {
     dbg.log0('put_bucket_replication:', req.rpc_params);
     const bucket = find_bucket(req);
 
-    validate_replication(req);
+    // fetching all the replication rules already present in the DB
+    const db_repl_rules = await replication_store.instance().get_replication_rules();
+
+    validate_replication(req, db_repl_rules);
     const replication_rules = normalize_replication(req);
 
     const bucket_replication_id = bucket.replication_policy_id;
@@ -1956,7 +1959,7 @@ async function delete_bucket_replication(req) {
     await replication_store.instance().delete_replication_by_id(replication_id);
 }
 
-function validate_replication(req) {
+function validate_replication(req, db_repl_rules) {
     const replication_rules = req.rpc_params.replication_policy.rules;
     // num of rules in configuration must be in defined limits
     if (replication_rules.length > config.BUCKET_REPLICATION_MAX_RULES ||
@@ -1964,13 +1967,36 @@ function validate_replication(req) {
 
     const rule_ids = [];
     const pref_by_dst_bucket = {};
+    const src_bucket = req.system.buckets_by_name && req.system.buckets_by_name[req.rpc_params.name.unwrap()];
 
     for (const rule of replication_rules) {
         const { destination_bucket, filter, rule_id } = rule;
+
         const dst_bucket = req.system.buckets_by_name && req.system.buckets_by_name[destination_bucket.unwrap()];
         // rule's destination bucket must exist and not equal to the replicated bucket
         if (req.rpc_params.name.unwrap() === destination_bucket.unwrap() || !dst_bucket) {
             throw new RpcError('INVALID_REPLICATION_POLICY', `Rule ${rule_id} destination bucket is invalid`);
+        }
+
+        // validation for bidirectional replication - blocking bidirectional replication only for matching prefixes
+        if (destination_bucket.unwrap() !== req.rpc_params.name.unwrap()) {
+            const prefix = filter?.prefix || '';
+            // checking if there already a rule consisting of src_bucket as destination_bucket for matching prefix
+            for (const db_rules of db_repl_rules) {
+                // checking if db_rules belongs to dst_bucket as source_bucket
+                if (_.isEqual(db_rules._id, dst_bucket.replication_policy_id)) {
+                    const matching_rule = db_rules.rules.find(
+                        db_rule =>
+                            _.isEqual(src_bucket._id, db_rule.destination_bucket) &&
+                            (!db_rule.filter || db_rule.filter.prefix.toString().startsWith(prefix) ||
+                             prefix.toString().startsWith(db_rule.filter.prefix.toString()))
+                    );
+                    if (matching_rule) {
+                        throw new RpcError('INVALID_REPLICATION_POLICY',
+                            `Bidirectional replication found for bucket "${destination_bucket.unwrap()}" and prefix is "${prefix}"`);
+                    }
+                }
+            }
         }
 
         rule_ids.push(rule_id);
--- a/src/server/system_services/replication_store.js
+++ b/src/server/system_services/replication_store.js
@@ -54,6 +54,11 @@ class ReplicationStore {
         return parsed_id;
     }
 
+    async get_replication_rules() {
+        const repl = await this._replicationconfigs.find({ deleted: null });
+        return repl;
+    }
+
     async get_replication_by_id(replication_id) {
         dbg.log1('get_replication_by_id: ', replication_id);
         const repl = await this._replicationconfigs.findOne({ _id: db_client.instance().parse_object_id(replication_id), deleted: null });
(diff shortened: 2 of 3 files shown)


src/server/system_services/bucket_server.js
```diff
@@ -1943,17 +1943,25 @@ async function delete_bucket_replication(req) {
...
-function validate_replication(req) {
+async function validate_replication(req) {
     const replication_rules = req.rpc_params.replication_policy.rules;
     // num of rules in configuration must be in defined limits
     if (replication_rules.length > config.BUCKET_REPLICATION_MAX_RULES ||
         replication_rules.length < 1) throw new RpcError('INVALID_REPLICATION_POLICY', 'Number of rules is invalid');
 
     const rule_ids = [];
     const pref_by_dst_bucket = {};
+    const src_bucket = req.system.buckets_by_name && req.system.buckets_by_name[req.name];
```
liranmauda, line 1954: why not get it from the rule? 
At this point we do not care about the bucket, only about it's name

src/server/system_services/bucket_server.js
```diff
@@ -1943,17 +1943,25 @@ async function delete_bucket_replication(req) {
...
 
     for (const rule of replication_rules) {
         const { destination_bucket, filter, rule_id } = rule;
+
+        // check for bidirectional replication
+        const repl_rules = filter && await replication_store.instance().get_replication_by_prefix_and_destination_bucket(filter.prefix, src_bucket.id);
```
liranmauda, line 1960: if the name of the source bucket in the reqest that we are chcking is equal to the destination_bucket in this rule, we don't (and should not) need to do this validation.

src/server/system_services/bucket_server.js
```diff
@@ -1943,17 +1943,25 @@ async function delete_bucket_replication(req) {
...
+        if (repl_rules != null) {
+            throw new RpcError('INVALID_REPLICATION_POLICY', 'found bidirectional replication');
```
liranmauda, line 1962: maybe more explaining error, I am not sure that `found bidirectional replication` will help the customer/support to understand what needs to be changed.

src/server/system_services/replication_store.js
```diff
@@ -60,6 +60,16 @@ class ReplicationStore {
...
 
+    async get_replication_by_prefix_and_destination_bucket(prefix, destination_bucket) {
+        dbg.log1('get_replication_by_prefix_and_destination_bucket: ', prefix, destination_bucket);
+        const query = {
+            "rules.filter.prefix": prefix,
+            "rules.destination_bucket": destination_bucket
+        }
+        const repl = await this._replicationconfigs.findOne( query , { deleted: null });
+        return repl && repl.rules;
+    }
```
liranmauda, line 71: This will find exact match with prefix.
We want:
1. exact match with the prefix
2. subset of the prefix
3. include in this prefix

liranmauda (changes requested):
Some comments + please add some tests

src/server/system_services/bucket_server.js
achouhan09: Thanks, updated the same.

src/server/system_services/bucket_server.js
achouhan09: Done.

src/server/system_services/bucket_server.js
achouhan09: Thanks, I have updated the error msg here.

src/server/system_services/replication_store.js
achouhan09: Done.

src/server/system_services/bucket_server.js
```diff
@@ -1956,23 +1959,45 @@ async function delete_bucket_replication(req) {
...
+        if (destination_bucket.unwrap() !== req.rpc_params.name.unwrap()) {
+            const prefix = (filter && filter.prefix) || '';
+            // check if there already a rule consisting of src_bucket as destination_bucket for matching prefix
+            for (const db_rules of db_repl_rules) {
+                if (_.isEqual(db_rules._id, dst_bucket.replication_policy_id)) {
+                    const matching_rule = db_rules.rules.find(
+                        db_rule =>
+                            _.isEqual(src_bucket._id, db_rule.destination_bucket) &&
+                            (!db_rule.filter || db_rule.filter.prefix.toString().startsWith(prefix) ||
+                             prefix.toString().startsWith(db_rule.filter.prefix.toString()))
```
liranmauda, line 1991: we should also block if one way contain empty prefix.

src/server/system_services/replication_store.js
```diff
@@ -54,6 +54,12 @@ class ReplicationStore {
         return parsed_id;
     }
 
+    async get_replication_rules() {
+        dbg.log1('get_replication_rules');
+        const repl = await this._replicationconfigs.find({ deleted: null });
+        return repl && repl;
```
liranmauda, line 60:
```suggestion
        return repl;
```

src/server/system_services/replication_store.js
```diff
@@ -54,6 +54,12 @@ class ReplicationStore {
         return parsed_id;
     }
 
+    async get_replication_rules() {
+        dbg.log1('get_replication_rules');
```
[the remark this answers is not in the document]
achouhan09: sure.

src/server/system_services/bucket_server.js
achouhan09: Not sure, if I understand your question but this condition will check for empty prefix as well. Also, I have written a test case for the same.

src/server/system_services/bucket_server.js
```diff
@@ -1956,23 +1959,45 @@ async function delete_bucket_replication(req) {
...
 
+        // validation for bidirectional replication - blocking bidirectional replication only for matching prefixes
+        if (destination_bucket.unwrap() !== req.rpc_params.name.unwrap()) {
+            const prefix = (filter && filter.prefix) || '';
+            // check if there already a rule consisting of src_bucket as destination_bucket for matching prefix
+            for (const db_rules of db_repl_rules) {
+                if (_.isEqual(db_rules._id, dst_bucket.replication_policy_id)) {
+                    const matching_rule = db_rules.rules.find(
+                        db_rule =>
+                            _.isEqual(src_bucket._id, db_rule.destination_bucket) &&
```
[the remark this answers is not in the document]
achouhan09: In first if condition we are checking if the `db_rules` we are checking is the replication policy for `dst_bucket`(as source bucket) , and in second condition we are finding the rules where `src_bucket` is destination bucket with the prefix conditions applied among the `db_rules`. Sure, I will add a comment there as well.

src/server/system_services/bucket_server.js
```diff
@@ -1956,23 +1959,45 @@ async function delete_bucket_replication(req) {
...
+
         const dst_bucket = req.system.buckets_by_name && req.system.buckets_by_name[destination_bucket.unwrap()];
         // rule's destination bucket must exist and not equal to the replicated bucket
         if (req.rpc_params.name.unwrap() === destination_bucket.unwrap() || !dst_bucket) {
             throw new RpcError('INVALID_REPLICATION_POLICY', `Rule ${rule_id} destination bucket is invalid`);
         }
 
+        // validation for bidirectional replication - blocking bidirectional replication only for matching prefixes
+        if (destination_bucket.unwrap() !== req.rpc_params.name.unwrap()) {
+            const prefix = (filter && filter.prefix) || '';
```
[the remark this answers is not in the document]
achouhan09: sure, thanks for suggestion.

src/server/system_services/bucket_server.js
```diff
@@ -1956,23 +1959,45 @@ async function delete_bucket_replication(req) {
...
+            // check if there already a rule consisting of src_bucket as destination_bucket for matching prefix
+            for (const db_rules of db_repl_rules) {
+                if (_.isEqual(db_rules._id, dst_bucket.replication_policy_id)) {
```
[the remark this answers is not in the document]
achouhan09: No `id` is an object in both `db_rules` and `dst_bucket` , so need to compare it with _.isEqual here.

src/server/system_services/replication_store.js
achouhan09: Thanks for suggestion.

src/server/system_services/bucket_server.js
liranmauda: True, and great adding the tests.

liranmauda (approved):

PR #7530, opened 2023-10-10
[NC | NSFS] Add mm policy migration rule to kick in every 15 mins

tangledbytes:
### Explain the changes
This PR adds a new rule to the spectrum scale policies which migrates **resident** data from disk to tape **every** 15 minutes along with the restore trigger that happens every 15 minutes.

**Why?**
Previously, if a user would upload an object to `GLACIER` storage class and then will issue a recall, they will have to wait at least till `00:15 UTC` to be able to get their object back. That is acceptable from S3 protocol perspective IMO as restores can take upto days but in this case we are simply taking more time to restore because the **resident** files are migrated to the tape at `00:00 UTC` and only once migrated we bring them back to the tape (every 15 mins) hence the user will have to wait till `00:15 UTC`.

**What?**
With this PR, we have 3 different flows, one for migrating restored data (aka _premigrated_ data), one for data that was freshly uploaded hence not on tape (aka _resident_ data) and one for restoring data to the disk from tape.
We trigger migration of premigrated data every day at `00:00 UTC` as the granularity for restore request is in days and there is no point in issuing it more often.
We trigger the restoration every 15 mins.
We trigger migration of resident data (freshly uploaded data) every 15 mins.

With this PR if a user uploads an object and immediately requests a refresh then at max the restore will take 30 mins (15 mins to first move that data to tape and then 15 to move that data back to disk).

**Can there be race between data moving to tape and back to disk in the same flow?**
I don't think so. The flows identify the candidates based on their state on the filesystem. If the file is `resident`, it will be handed over to migration flow while if the file is `migrated` it will be handed over to the restore flow.

Tested this in the tape simulation environment.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/deploy/spectrum_archive/process_restore_candidates.sh
+++ b/src/deploy/spectrum_archive/process_restore_candidates.sh
@@ -13,9 +13,10 @@ function get_expiry() {
 		exit 1
 	fi
 
-	# This ugly piece of code converts the ISO 8601 date to a format that is similar
-	# to NodeJS's Date.toISOString() output
-	local expiry=$(date -u -Ins -d "+$requested_days days" | sed s/+00:00/Z/ | sed s/,/./)
+	# Take the current date and add "requested_days" to it and strip away the time
+	# bits just like AWS does
+	# This format is ISO 8601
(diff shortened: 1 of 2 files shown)


romayalon (approved):
LGTM

PR #7531, opened 2023-10-10
NSFS | NC | Add syslog configuration on RPM

naveenpaul1:
### Explain the changes
1. Add syslog configuration on RPM deployment/service enablement

1. https://issues.redhat.com/browse/MCGI-196

### Testing Instructions:
1. install RPM
2. check for log file `var/log/noobaa.log`
3. verify logrotate by running.
`logrotate /etc/logrotate.d/noobaa/logrotate_noobaa.conf` and check for `var/log/noobaa.log.1.gz`

- [X] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/deploy/RPM_build/noobaa.spec
+++ b/src/deploy/RPM_build/noobaa.spec
@@ -39,6 +39,7 @@ tar -xJf %{SOURCE1} -C node-%{nodever}/
 %install
 rm -rf $RPM_BUILD_ROOT
 mkdir -p $RPM_BUILD_ROOT/usr/local/
+mkdir -p $RPM_BUILD_ROOT/etc/logrotate.d/noobaa
 
 cp -R %{_builddir}/%{name}-%{version}-%{revision}/noobaa $RPM_BUILD_ROOT/usr/local/noobaa-core
 cp -R %{_builddir}/node-%{nodever}/* $RPM_BUILD_ROOT/usr/local/noobaa-core/node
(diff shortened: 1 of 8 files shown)


src/util/debug_module.js
```diff
@@ -21,6 +21,7 @@ const path = require('path');
 const util = require('util');
 const os = require('os');
 const debug_config = require('./debug_config');
+const env_config = require('../../config.js');
```
tangledbytes, line 24: Nitpick: We usually call it just `config`. So `const env_config = require('../../config.js')`.

src/deploy/RPM_build/noobaa.spec
```diff
@@ -51,13 +52,23 @@ ln -s /usr/local/noobaa-core/node/bin/npx $RPM_BUILD_ROOT/usr/local/noobaa-core/
...
 
+cp -R /etc/rsyslog.d $RPM_BUILD_ROOT/etc/rsyslog.d
+cp -R /etc/logrotate.d/noobaa $RPM_BUILD_ROOT/etc/logrotate.d/
+
 
 %files
 /usr/local/noobaa-core
 /etc/systemd/system/nsfs.service
+/etc/logrotate.d/noobaa/logrotate_noobaa.conf
+/etc/rsyslog.d
```
tangledbytes, line 63: I am not sure if I understand this correctly but does this mean that we take the ownership of the entire `/etc/rsyslog.d` directory? What if it is already placed by another program?

naveenpaul1: @tangledbytes ya, It make sense, I will give specific file names,

tangledbytes (approved):

PR #7533, opened 2023-10-16
NC | NSFS | HTTP and certificate configurations

naveenpaul1:
### Explain the changes
1. create allow/deny http configuration - take it from noobaa.conf.d/system.json
certificate file location, see - ssl_utils.js - take it from noobaa.conf.d/system.json
system.json
`{
  "current_version": "5.15.0",
  "upgrade_history": {
    "successful_upgrades": []
   },
   "allow_http": false,
   "nsfs_ssl_cert_dir": "/etc/nc_nsfs_secret"
}`

1. https://issues.redhat.com/browse/MCGI-203

### Testing Instructions:
1. 

- [x] Doc added/updated
- [ ] Tests added

### DIFF
--- a/docs/non_containerized_NSFS.md
+++ b/docs/non_containerized_NSFS.md
@@ -264,6 +264,12 @@ node src/cmd/manage_nsfs bucket add --config_root ../standalon/config_root --fro
 ```
 NSFS management CLI command will create both account and bucket dir if it's missing in the config_root path.
 
+## NSFS Certificate
+
+Non containerized NSFS certificate location is configured in system.json file under the property `nsfs_ssl_cert_dir` and the path should contain SSL files tls.key and tls.crt. System will use a cert from this dir to create a valid HTTPS connection. If cert is missing in this dir a self-signed SSL certificate will be generated. Make sure the path mentioned in `nsfs_ssl_cert_dir` is valid before running nsfs command, If the path is invalid then cert flow will fail.
+
+Non containerized NSFS allow nonsecure HTTP connection only when `allow_http` in system.json is true.
+
 ## Log and Logrotate
 Noobaa logs are configured using rsyslog and logrotate. RPM will configure rsyslog and logrotate if both are already running. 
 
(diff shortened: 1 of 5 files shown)


src/endpoint/endpoint.js
```diff
@@ -204,6 +208,16 @@ async function main(options = {}) {
     }
 }
 
+async function is_nsfs_http_denied(nsfs_config_root) {
+    if (!nsfs_config_root) {
```
romayalon, line 212: I think you can take the !nsfs_config_root condition outside instead of get into this flow.

src/util/ssl_utils.js
```diff
@@ -60,8 +61,14 @@ function verify_ssl_certificate(certificate) {
 }
 
 // Get SSL certificate (load once then serve from cache)
-function get_ssl_cert_info(service) {
-    const cert_info = certs[service];
+async function get_ssl_cert_info(service, nsfs_config_root) {
+    const nsfs_ssl_cert_dir = await get_nfs_system_property('nsfs_ssl_cert_dir', nsfs_config_root);
```
romayalon, line 65: can we read this property outside, for example on nsfs.js?

romayalon (commented):
looks good, some minor questions

docs/non_containerized_NSFS.md
```diff
@@ -264,6 +264,12 @@ node src/cmd/manage_nsfs bucket add --config_root ../standalon/config_root --fro
 ```
 NSFS management CLI command will create both account and bucket dir if it's missing in the config_root path.
 
+## NSFS Certificate
+
+Non containerized NSFS certificate location is configured in system.json file under the property `nsfs_ssl_cert_dir`. System will use a cert from this dir to create a valid HTTPS connection. If cert is missing in this dir a self-signed SSL certificate will be generated. Make sure the path mentioned in `nsfs_ssl_cert_dir` is valid before running nsfs command, If the path is invalid then cert flow will fail.
+
```
jackyalbo, line 270: maybe explain that we expect two files under the directory and that they have to be named tls.key and tls.crt.

romayalon (approved):
LGTM

PR #7540, opened 2023-10-23
NC | NSFS | Events

naveenpaul1:
event command added

### Explain the changes
1. add support for NSFS events
2. verify GPFS_DL_PATH

1. https://issues.redhat.com/browse/MCGI-192

### Testing Instructions:
verify the events in location /vars/logs/noobaa_events.log

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/docs/dev_guide/non_containerized_NSFS_events.md
+++ b/docs/dev_guide/non_containerized_NSFS_events.md
@@ -1,0 +1,84 @@
+# Non Containerized NSFS Events
+
+This document will list all the possible Noobaa non-containerized NSFS events and possible reasons and resolutions.
+
+## Events
+
+### 1. noobaa_nsfs_crashed
+#### Reseason
+- Noobaa endpoint module failed to load.
+- High Noobaa resource utilization.
+
+#### Resolutions
+- Check Noobaa resource utilization.
+
+### 2. noobaa_gpfslib_missing
+arguments: `gpfs_dl_path`
+#### Reseason
+- Missing gpfslib in `GPFS_DL_PATH` path.
+#### Resolutions
+- Add gpfslib in `GPFS_DL_PATH` path.
+
+### 3. noobaa_started
+#### Reseason
+- Noobaa started without any issues.
+#### Resolutions
+- Nil
+
+### 4. noobaa_account_created
+arguments: `account_name`
+#### Reseason
+- Noobaa user account created.
+#### Resolutions
+- Nil
+
+### 5. noobaa_bucket_creation_failed
+arguments: `bucket_name`
+#### Reseason
+- User does not have permission to update `noobaa.conf.d` dir and its redirect path if present.
+- User does not have permission to create the bucket's underlying storage directory.
+
+#### Resolutions
+- Check access rights for `noobaa.conf.d` dir and it's redirect path if present.
+- Check account `new_buckets_path` property and verify adequate permission present for this dir location. 
+
+### 6. noobaa_bucket_delete_failed
+arguments: `bucket_name`, `bucket_path`
+#### Reseason
+- User does not have permission to delete the bucket config file from `noobaa.conf.d` dir and its redirect path if present.
+- User does not have permission to delete the bucket's underlying storage directory.
+- Bucket storage dir is missing.
+
+#### Resolutions
+- Check access rights for `noobaa.conf.d` dir and it's redirect path if present.
+- Check account `new_buckets_path` property and verify adequate permission present for this dir location. 
+- Make sure both the bucket config field and underlying storage dir are present.
+
+### 7. noobaa_bucket_not_found
+arguments: `bucket_name`
+#### Reseason
+- Bucket config file in config_root path is missing.
+- Bucket config JSON schema validation failed.
+- Bucket's underlying storage directory not found 
+#### Resolutions
+- Check for the valid bucket config file in config root dir.
+- Verify bucket config JSON schema.
+- Check for Bucket's underlying storage directory present with permission.
+
+### 8. noobaa_object_get_failed
+arguments : `bucket_path`, `object_name`
+#### Reseason
+- Noobaa bucket path is missing.
+- Bucket I/O operation is failed.
+#### Resolutions
+- Verify the bucket path.
+
+### 9. noobaa_object_uplaod_failed
+arguments : `bucket_path`, `object_name`
+#### Reseason
+- Bucket path is outside the bucket boundaries.
+- Bucket storage class is not supported.
+- Object I/O operation is failed.
+#### Resolutions
+- Make sure bucket storage class is supported.
+- Check for I/O operations.
--- a/src/cmd/nsfs.js
+++ b/src/cmd/nsfs.js
@@ -265,6 +265,7 @@ async function main(argv = minimist(process.argv.slice(2))) {
             backend,
             warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
         };
+        verify_gpfslib(backend, fs_config);
         const account = {
             email: new SensitiveString('nsfs@noobaa.io'),
             nsfs_account_config: fs_config,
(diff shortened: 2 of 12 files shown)


src/cmd/nsfs.js
```diff
@@ -321,6 +321,14 @@ async function main(argv = minimist(process.argv.slice(2))) {
 
     } catch (err) {
         console.error('nsfs: exit on error', err.stack || err);
+        //noobaa crashed
+        dbg.event("{ \"code\":\"noobaa_nsfs_crashed\"," +
```
romayalon, line 325: maybe having a function in a utils file for event logging could be better?

src/cmd/event.js
```diff
@@ -0,0 +1,196 @@
...
+const ARGUMENTS = `
+Arguments:
+
+    <action>  Action could be list active and all the events.
+`;
+
+const OPTIONS = `
+Account Options:
+
+    # Read account details from json file, no need to mention all the propery one by one in CLI
```
romayalon, line 33: need to update this row

src/cmd/event.js
```diff
@@ -0,0 +1,196 @@
...
+    result_events = await sort_events(result_events);
+    return result_events;
+}
+
+async function get_all_events() {
+    try {
+        const events_logs = await fs.promises.readFile('/var/log/noobaa_events.log');
+        const content = events_logs.toString();
+        let events_info = content.split('\n');
+        events_info = events_info.slice().reverse();
```
romayalon, line 179: why reversing the events?

src/cmd/event.js
```diff
@@ -0,0 +1,196 @@
...
+            const time_diff = Date.now() - event_time.getTime();
+            if (time_diff < config.EVENT_ACTIVE_INTERVAL) {
+                message.time = new Date(event_time.getTime());
+                const repeated = structure_repeated_events(message, active_events);
+                if (!repeated || active_events.length === 0) {
+                    message.message_thread.push(message.message);
+                    active_events.push(message);
+                }
+            } else {
+                active_events = sort_events(active_events);
```
romayalon, line 126: can you use break instead?

src/cmd/event.js
```diff
@@ -0,0 +1,196 @@
...
+                return active_events;
+            }
+        }
+    });
+    active_events = sort_events(active_events);
```
romayalon, line 131: why do we need to sort the events by severity?

src/cmd/event.js
naveenpaul1: @romayalon Since we are reading from a log file,  most recent(active) event will be placed end of the file and to get the active events fast I thought reversing the array is a good option, another option is to traverse from the bottom. WDYT?

src/cmd/event.js
naveenpaul1: @romayalon scale team is expecting the sorted active events based on severity.
```If there are multiple events shown by the command, they will be listed in the order we recommend```
```they be fixed, with the most important event to fix at the top ```

**Comment by tangledbytes** [comment]
@naveenpaul1 we won't be doing file rotations?

**Comment by naveenpaul1** [comment]
@tangledbytes yes, event log file is rottated  [src/deploy/standalone/logrotate_noobaa.conf](https://github.com/noobaa/noobaa-core/pull/7540/files#diff-a7493d66d4a4148793f7ef8494bc55d347ebc9916f4b6b40febb57aea7fd51ce)

File: src/cmd/event.js (hunk 3 above)
Comment (romayalon) [reply src/cmd/event.js:178]: it's obviously nice but I'm not sure if it's too much effort,
I checked the events file we have on a regular must gather and I don't see that it's sorted/reversed
@dannyzaken WDYT?

File: src/cmd/event.js (hunk 5 above)
Comment (romayalon) [reply src/cmd/event.js:130]: got you!

File: src/util/json_utils.js [hunk 6]
```diff
@@ -67,5 +67,19 @@ class JsonObjectWrapper {
 
 }
 
+async function format_and_send_event(event) {
+    const event_json = "{ \"code\":\"" + event.code + "\"," +
```
tangledbytes, line 71: Can't we create a js object and simply `JSON.stringify(value)`?

src/cmd/event.js
tangledbytes: Yeah, if we can avoid the reversal then we can also increase the log file size and just read the file line by line instead of reading it all in memory.

src/util/json_utils.js
dannyzaken: agree. this can be simply converted to 
```javascript
    const event_json = JSON.stringify(event);
```

dannyzaken: Also, I don't think it belongs in `json_utils.js`. This module was mainly used to handle jsons stored in files.
Maybe it makes sense to move this implementation to debug_module

src/util/debug_module.js
```diff
@@ -517,6 +525,7 @@ DebugLogger.prototype.warn = log_syslog_builder('warn');
 DebugLogger.prototype.info = log_syslog_builder('info');
 DebugLogger.prototype.log = log_syslog_builder('log');
 DebugLogger.prototype.trace = log_syslog_builder('trace');
+DebugLogger.prototype.event = log_syslog_builder('event');
```
dannyzaken, line 528: I think we should avoid using this `log_syslog_builder`, especially considering you later remove all the formatting added by this function. 
I suggest adding a simple function to log the event.

Also, maybe we should add a global toggle to disable event logging, just in case?

```javascript
function log_event(event) {
    if (!config.EVENT_LOGGING_ENABLED) return;// necessary?
    syslog(int_dbg._levels_to_syslog.EVENT, JSON.stringify(event), config.EVENT_FACILITY);
}
```

and export it as the `event` function
```suggestion
DebugLogger.prototype.event = log_event;
```

src/cmd/event.js
```diff
@@ -0,0 +1,195 @@
...
+            }
+        }
+    });
+    result_events = await sort_events(result_events);
+    return result_events;
+}
+
+async function get_all_events() {
+    try {
+        const events_logs = await fs.promises.readFile('/var/log/noobaa_events.log');
```
dannyzaken, line 175: a few comments:
1. better to have the path in a constant, since you use it in multiple places. Also it's probably useful to provide a flag to override the default 
2. this file can be rotated by logrotate. shouldn't we get events from all files? what are the requirements?
3. since this file is written by syslog, won't it contain syslog headers that we need to strip before parsing as JSON?

naveenpaul1: @dannyzaken monitoring team don't need a CLI to fetch the event, they will be fetching the event directly from the event file, 
will remove this file, 
Yes, this file can rotate by logrotate, current condition rotate monthly or size in 10M, I am not sure how we can query the previous event log file in this case.
All the valid event properties are under the property `message`, only `timestamp` needs to be fetched from syslog header, and monitoring team is ok with that,
``` {"timestamp":"2023-10-31T06:43:53.127151+00:00","message":"{\"code\":\"noobaa_started\",\"entity_type\":\"NODE\",\"event_type\":\"STATE_CHANGE\",\"message\":\"Noobaa started\",\"scope\":\"NODE\",\"severity\":\"INFO\",\"state\":\"HEALTHY\",\"pid\":650}", "host":"cbf26f4d14d2","severity":"notice","facility":"local2","syslog-tag":"node[650]:","source":"node"} ```

File: src/util/debug_module.js [hunk 9]
```diff
@@ -398,7 +401,12 @@ class InternalDebugLogger {
     log_internal(msg_info) {
         if (syslog) {
             // syslog path
-            syslog(this._levels_to_syslog[msg_info.level], msg_info.message_syslog, config.DEBUG_FACILITY);
+            if (msg_info.level === 'EVENT') {
```
dannyzaken, line 404: isn't it redundant now when you have log_event as a separate function?

src/sdk/namespace_fs.js
```diff
@@ -915,6 +915,18 @@ class NamespaceFS {
                     const dir_stat = await nb_native().fs.stat(fs_context, md_path);
                     if (dir_stat && dir_stat.xattr[XATTR_DIR_CONTENT] === '0') return null;
                 } catch (err) {
+                    //failed to get object
+                    dbg.event({
+                        code: "noobaa_object_get_failed",
```
romayalon, line 920: can you please add a list of these codes on the docs as requested on the HPO meeting?

naveenpaul1: @romayalon An initial version of the doc is added. [Link](https://github.com/noobaa/noobaa-core/blob/78c49df1902eae6b978570d4a982e3efb2cc758b/docs/dev_guide/non_containerized_NSFS_events.md)

docs/dev_guide/non_containerized_NSFS_events.md
```diff
@@ -0,0 +1,81 @@
...
+
+### 8. noobaa_object_get_failed
+arguments : `bucket_path`, `object_name`
+#### Reseason
+- Noobaa bucket path is missing.
+- Bucket I/O operation is failed.
+#### Resolutions
+- Verify the bucket path.
+
+### 9. noobaa_object_uplaod_failed
```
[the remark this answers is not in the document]
romayalon: there can be other reasons for the upload's failure, maybe add `Object I/O operation is failed.` as well ?

src/cmd/nsfs.js
```diff
@@ -324,10 +325,38 @@ async function main(argv = minimist(process.argv.slice(2))) {
...
+            scope: "NODE",
+            severity: "ERROR",
+            state: "STOPPED"
+        });
         process.exit(2);
     }
 }
 
+async function verify_gpfslib(backend, fs_config) {
+    if (backend === 'GPFS' && nb_native().fs.gpfs) {
```
romayalon, line 343: I don't think we are really using backend === 'GPFS', I think you can remove this condition

tangledbytes (approved):

PR #7541, opened 2023-10-23
Logs transfer changes bucket logging

aspandey:
For now, whenever an operation happens on an object, we record that log
    and send it to endpoint logs.
    This change is responsible for sending these logs to noobaa core
    rsyslogs.


aspandey:
created by mistake

PR #7544, opened 2023-10-25
NC | NSFS | Schema validation

naveenpaul1:
### Explain the changes
1. Buckets and accounts schema validation
2. creation_date added to bucket management CLI

1. https://issues.redhat.com/browse/MCGI-204

### Testing Instructions:
1. Test with invalid schema JSON for bucket and accounts, Should throw an error
2. Try to create bucket with an invalid creation date, if the creation date missing current data use

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -150,6 +150,7 @@ async function fetch_bucket_data(argv, config_root, from_file) {
             should_create_underlying_storage: true,
         };
     }
+    data.creation_date = new Date().toISOString();
     if (action === 'update') {
         data = _.omitBy(data, _.isUndefined);
         data = await update_bucket_object(config_root, data);
@@ -284,6 +285,7 @@ async function fetch_account_data(argv, config_root, from_file) {
             }
         };
     }
+    data.creation_date = new Date().toISOString();
     if (action === 'update') {
         data = _.omitBy(data, _.isUndefined);
         data = update_account_object(config_root, data);
@@ -421,7 +423,7 @@ async function validate_bucket_add_args(data) {
         return false;
     }
     if (!data.path.includes(data.name.unwrap())) {
-        console.error('Error: bucket name should be same as dir name');
+        console.error('Error: Bucket name should be same as dir name');
         return false;
     }
     const bucket_dir_stat = await fs_utils.file_exists(data.path);
--- a/src/cmd/nsfs.js
+++ b/src/cmd/nsfs.js
@@ -317,7 +317,7 @@ async function main(argv = minimist(process.argv.slice(2))) {
             }
         });
         if (await endpoint.is_http_allowed(nsfs_config_root)) {
-            dbg.log0('nsfs: listening on', util.inspect(`http://localhost:${http_port}`));
+            console.log('nsfs: listening on', util.inspect(`http://localhost:${http_port}`));
         }
         console.log('nsfs: listening on', util.inspect(`https://localhost:${https_port}`));
 
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -14,6 +14,10 @@ const BucketSpaceSimpleFS = require('./bucketspace_simple_fs');
 const _ = require('lodash');
 const util = require('util');
 const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
+const { default: Ajv } = require('ajv');
+const bucket_schema = require('../server/object_services/schemas/nsfs_bucket_schema');
+const account_schema = require('../server/object_services/schemas/nsfs_account_schema');
+
 
 const dbg = require('../util/debug_module')(__filename);
 
@@ -21,6 +25,7 @@ const VALID_BUCKET_NAME_REGEXP =
     /^(([a-z0-9]|[a-z0-9][a-z0-9-]*[a-z0-9])\.)*([a-z0-9]|[a-z0-9][a-z0-9-]*[a-z0-9])$/;
 const ACCOUNT_PATH = 'accounts';
 const BUCKET_PATH = 'buckets';
+const ajv = new Ajv({ verbose: true, allErrors: true });
 
 
 //TODO:  dup from namespace_fs - need to handle and not dup code
@@ -89,6 +94,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             const iam_path = this._get_account_config_path(access_key);
             const { data } = await nb_native().fs.readFile(this.fs_context, iam_path);
             const account = JSON.parse(data.toString());
+            this.validate_account_schema(account);
             account.name = new SensitiveString(account.name);
             account.email = new SensitiveString(account.email);
             for (const k of account.access_keys) {
@@ -98,16 +104,29 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             return account;
         } catch (err) {
             dbg.error('BucketSpaceFS.read_account_by_access_key: failed with error', err);
-            throw new RpcError('NO_SUCH_ACCOUNT', `Account with access_key not found`, err);
+            if (err.code === 'ENOENT') {
+                throw new RpcError('NO_SUCH_ACCOUNT', `Account with access_key not found.`, err);
+            }
+            throw new RpcError('NO_SUCH_ACCOUNT', err.message);
         }
     }
 
+    validate_account_schema(account) {
+        const valid = ajv.validate(account_schema, account);
+        if (!valid) throw new RpcError('INVALID_SCHEMA', ajv.errors[0]?.message);
+    }
+
+    validate_bucket_schema(bucket) {
+        const valid = ajv.validate(bucket_schema, bucket);
+        if (!valid) throw new RpcError('INVALID_SCHEMA', ajv.errors[0]?.message);
+    }
     async read_bucket_sdk_info({ name }) {
         try {
             const bucket_config_path = this._get_bucket_config_path(name);
             dbg.log0('BucketSpaceFS.read_bucket_sdk_info: bucket_config_path', bucket_config_path);
             const { data } = await nb_native().fs.readFile(this.fs_context, bucket_config_path);
             const bucket = JSON.parse(data.toString());
+            this.validate_bucket_schema(bucket);
             const is_valid = await this.check_bucket_config(bucket);
             if (!is_valid) {
                 console.warn('BucketSpaceFS: one or more bucket config check is failed for bucket : ', name);
@@ -247,8 +266,8 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
         bucket.path = bucket_path;
         bucket.should_create_underlying_storage = true;
         bucket.force_md5_etag = params.force_md5_etag;
-        bucket.bucket_owner = sdk.requesting_account.email;
         bucket.creation_date = new Date().toISOString();
+        this.validate_bucket_schema(bucket);
         const create_bucket = JSON.stringify(bucket);
 
         // TODO: handle both bucket config json and directory creation atomically 
--- a/src/server/object_services/schemas/nsfs_account_schema.js
+++ b/src/server/object_services/schemas/nsfs_account_schema.js
@@ -1,0 +1,55 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+module.exports = {
+    $id: 'account_schema',
+    type: 'object',
+    required: [
+        'name',
+        'email',
+        'access_keys',
+        'nsfs_account_config',
+        'creation_date',
+    ],
+    properties: {
+        name: {
+            type: 'string'
+        },
+        email: {
+            type: 'string',
+        },
+        creation_date: {
+            type: 'string',
+        },
+        access_keys: {
+            type: 'array',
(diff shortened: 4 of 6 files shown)


src/sdk/bucketspace_fs.js
```diff
@@ -98,16 +99,41 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+    async validate_account_schema(account) {
+        if (!account.name || !account.email) {
+            throw new RpcError('MISSING_ACCOUNT_INFO', 'Missing account info.');
+        }
+        if (!account.access_keys[0] || !account.access_keys[0].access_key || !account.access_keys[0].secret_key) {
+            throw new RpcError('INVALID_ACCESS', 'Missing account info.');
+        }
+        if (!account.nsfs_account_config || account.nsfs_account_config.uid === undefined ||
+            account.nsfs_account_config.gid === undefined || !account.nsfs_account_config.new_buckets_path) {
+            throw new RpcError('MISSING_NSFS_ACCOUNT_CONFIGURATION', 'Missing account configuration.');
```
shirady, line 118: Why did you add "NSFS" ('MISSING_NSFS_ACCOUNT_CONFIGURATION') and in the other cases you didn't?

src/sdk/bucketspace_fs.js
```diff
@@ -98,16 +99,41 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
             dbg.error('BucketSpaceFS.read_account_by_access_key: failed with error', err);
-            throw new RpcError('NO_SUCH_ACCOUNT', `Account with access_key not found`, err);
+            if (err.code === 'ENOENT') {
+                throw new RpcError('NO_SUCH_ACCOUNT', `Account with access_key not found.`, err);
+            }
+            throw new RpcError('NO_SUCH_ACCOUNT', err.message);
+        }
+    }
+
+    async validate_account_schema(account) {
```
shirady, line 109: Why is it an async function?

src/sdk/bucketspace_fs.js
```diff
@@ -98,16 +99,41 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        if (!account.access_keys[0] || !account.access_keys[0].access_key || !account.access_keys[0].secret_key) {
+            throw new RpcError('INVALID_ACCESS', 'Missing account info.');
+        }
+        if (!account.nsfs_account_config || account.nsfs_account_config.uid === undefined ||
+            account.nsfs_account_config.gid === undefined || !account.nsfs_account_config.new_buckets_path) {
+            throw new RpcError('MISSING_NSFS_ACCOUNT_CONFIGURATION', 'Missing account configuration.');
         }
     }
 
+    async validate_bucket_schema(bucket) {
```
shirady, line 122: Why is it an async function?

src/sdk/bucketspace_fs.js
```diff
@@ -98,16 +99,41 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        if (!bucket.name || !bucket.system_owner || !bucket.bucket_owner) {
+            throw new RpcError('INVALID_BUCKET_INFO', 'Missing bucket info.');
+        }
+        if (!bucket.path || !bucket.creation_date) {
+            throw new RpcError('INVALID_BUCKET_STATE', 'Missing bucket storage details.');
```
shirady, line 127: If I see an error "'Missing bucket storage details." how can I know how to fix it?

src/sdk/bucketspace_fs.js
```diff
@@ -98,16 +99,41 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+            throw new RpcError('INVALID_ACCESS', 'Missing account info.');
+        }
+        if (!account.nsfs_account_config || account.nsfs_account_config.uid === undefined ||
+            account.nsfs_account_config.gid === undefined || !account.nsfs_account_config.new_buckets_path) {
+            throw new RpcError('MISSING_NSFS_ACCOUNT_CONFIGURATION', 'Missing account configuration.');
         }
     }
 
+    async validate_bucket_schema(bucket) {
+        if (!bucket.name || !bucket.system_owner || !bucket.bucket_owner) {
```
shirady, line 123: What is `system_owner`?
I saw it is by default `new SensitiveString(email)` (but the `bucket_owner` is also defined that way.

src/cmd/manage_nsfs.js
```diff
@@ -148,8 +150,10 @@ async function fetch_bucket_data(argv, config_root, from_file) {
             versioning: 'DISABLED',
             path: bucket_path,
             should_create_underlying_storage: true,
+            creation_date: creation_date,
         };
     }
+    update_bucket_creation_date(data);
```
shirady, line 156: Why do you set the `creation_date` in the `data` and then update it (`update_bucket_creation_date`)?
You can update it and then set it in the data, the function would return the date instead of changing the property in place, for example:
```suggestion
   creation_date: get_updated_bucket_creation_date(creation_date),
```

src/cmd/manage_nsfs.js
```diff
@@ -172,6 +176,14 @@ async function update_bucket_object(config_root, target) {
     return data;
 }
 
+async function update_bucket_creation_date(data) {
```
shirady, line 179: Why is it an async function?

shirady (commented):
Hi,
Added a couple of comments and questions.

src/cmd/manage_nsfs.js
```diff
@@ -172,6 +176,14 @@ async function update_bucket_object(config_root, target) {
...
+    if (data.creation_date) {
+        data.creation_date = Date.parse(data.creation_date) ? new Date(Date.parse(data.creation_date)).toISOString() : '';
+    } else {
+        data.creation_date = new Date().toISOString();
```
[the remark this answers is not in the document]
naveenpaul1: `toISOString()` will format the date

src/cmd/manage_nsfs.js
```diff
@@ -61,6 +61,7 @@ Bucket Options:
     # required for add, replace 
     --email <email>             (default none)          Set the email for account/bucket.
     --path <dir>                (default none)          Set the bucket path.
+    --creation_date <time>      (default current time)          Set time bucket creation. Formate YYYY-MM-DDTHH:mm:ssZ,
```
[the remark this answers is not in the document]
naveenpaul1: creation date is mandatory for S3 bucket, S3 bucket format is `YYYY-MM-DDTHH:mm:ssZ`, user will get  the format from --help argument

src/sdk/bucketspace_fs.js
```diff
@@ -98,16 +99,41 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+            throw new RpcError('NO_SUCH_ACCOUNT', err.message);
+        }
+    }
+
+    async validate_account_schema(account) {
+        if (!account.name || !account.email) {
+            throw new RpcError('MISSING_ACCOUNT_INFO', 'Missing account info.');
+        }
+        if (!account.access_keys[0] || !account.access_keys[0].access_key || !account.access_keys[0].secret_key) {
+            throw new RpcError('INVALID_ACCESS', 'Missing account info.');
```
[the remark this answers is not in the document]
naveenpaul1: access and secret key is a mix of alphanumeric characters with special characters not sure very specific validation is required here.

src/sdk/bucketspace_fs.js
naveenpaul1: this check is specific to NSFS, normal account will not have account configuration.

src/sdk/bucketspace_fs.js
naveenpaul1: a person creating the bucket, in our case both remain the same.

src/cmd/manage_nsfs.js
naveenpaul1: we can't do that because we need to do the same logic for --from-file flow too.

shirady (approved):
LGTM

@naveenpaul1, please check if the validation that that we have a value (without checking something about the value) is enough to meet the requirement.

src/cmd/manage_nsfs.js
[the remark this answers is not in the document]
shirady: As understood, the user would not need to use milliseconds, so it should be fine (probably a minute will be enough).

src/sdk/bucketspace_fs.js
```diff
@@ -98,16 +99,41 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
             dbg.error('BucketSpaceFS.read_account_by_access_key: failed with error', err);
-            throw new RpcError('NO_SUCH_ACCOUNT', `Account with access_key not found`, err);
+            if (err.code === 'ENOENT') {
+                throw new RpcError('NO_SUCH_ACCOUNT', `Account with access_key not found.`, err);
+            }
+            throw new RpcError('NO_SUCH_ACCOUNT', err.message);
+        }
+    }
+
+    validate_account_schema(account) {
```
romayalon, line 109: hmm I think we want _ajv validation and not a custom validation, while creating a schema, validating etc, @liranmauda @guymguym WDYT ?

romayalon: same for the bucket validation

romayalon (changes requested):

src/sdk/bucketspace_fs.js
romayalon: adding here a reference for ajv validation we do in postgres client - https://github.com/noobaa/noobaa-core/blob/master/src/util/postgres_client.js#L1322

src/cmd/manage_nsfs.js
romayalon, line 64: can you please update the bucket schema of the nsfs standalone doc with this, also I wonder if we want to have also creation_date on account as well, WDYT?

naveenpaul1: so the `creation_date` will be current date, right?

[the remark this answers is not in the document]
naveenpaul1: if the creation_date is empty then the current date is used, So you suggest removing the `creation_date` flag, is that right?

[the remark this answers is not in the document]
romayalon: I'm open to the option to have this flag, I just think that people will prefer to have it always created by us, no?
@PravinRanjan10 do you think the CLI should have a creation_date flag?

romayalon (commented):
I think we should also use the schema validation in bucketspace_fs on create bucket :)

src/cmd/manage_nsfs.js
romayalon: yeah, sounds good to me

src/cmd/manage_nsfs.js
```diff
@@ -61,6 +61,7 @@ Bucket Options:
     # required for add, replace 
     --email <email>             (default none)          Set the email for account/bucket.
     --path <dir>                (default none)          Set the bucket path.
+    --creation_date <time>      (default current time)  Set time bucket creation. Formate YYYY-MM-DDTHH:mm:ssZ,
```
liranmauda, line 64: Why do we need the creation_date to be exposed as a flag?

liranmauda (changes requested):
@naveenpaul1 
few comments/questions:
1. Am I correct and the creation and read of the account/bucket are not calling any RPC? 
2. please change the names of those Schema, to reflect that those are standalone (or any other name) schemas and not a RPC ones. Also, we already have files with those names so it can get us confused.

src/cmd/manage_nsfs.js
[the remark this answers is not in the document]
liranmauda: @romayalon I just commented the same without seeing this comment... 
@naveenpaul1 I also think that we should not have this flag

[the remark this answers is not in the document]
PravinRanjan10: > I'm open to the option of having this flag, I just think that people will prefer to have it always created by us, no? @PravinRanjan10 do you think the CLI should have a creation_date flag?

@romayalon CLI should not have this flag. The internal logic should generate, once bucket creation is successful.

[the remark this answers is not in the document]
naveenpaul1: Will remove the `creation_date`

src/cmd/nsfs.js
```diff
@@ -317,7 +317,7 @@ async function main(argv = minimist(process.argv.slice(2))) {
             }
         });
         if (await endpoint.is_http_allowed(nsfs_config_root)) {
-            dbg.log0('nsfs: listening on', util.inspect(`http://localhost:${http_port}`));
+            console.log('nsfs: listening on', util.inspect(`http://localhost:${http_port}`));
```
liranmauda, line 320: I wonder, why are we using console and not dbg? 
by requireing the console we "hijak" the console, but console does not have levels like 0,1 ...

liranmauda (approved):
LGTM

src/cmd/nsfs.js
naveenpaul1: @liranmauda ya totally agree with you, we need to have a discussion on this. In this PR I just reverted the previous code that I accidentally updated in my last PR.

romayalon (approved):

PR #7545, opened 2023-10-25
NSFS | NC | removing system store requirement from nsfs.js

romayalon:
### Explain the changes
1. System store requirement initiating the constructor of the system Store which defines the db collections. while defining db collection the vacumeAnalyzer is set to run once a day so the nsfs process panics when it happens.
currently, it's not being used so I suggest removing this requirement.

```Oct-25 18:13:17.778 [nsfs/90733] [ERROR] core.util.postgres_client:: vacuumAndAnalyze failed TypeError: Cannot read properties of undefined (reading 'query')
    at _do_query (/Users/romy/github/noobaa-core/src/util/postgres_client.js:250:37)
    at PostgresTable.single_query (/Users/romy/github/noobaa-core/src/util/postgres_client.js:679:16)
    at runNextTicks (node:internal/process/task_queues:60:5)
    at listOnTimeout (node:internal/timers:538:9)
    at process.processTimers (node:internal/timers:512:7)
    at async Timeout.vacuumAndAnalyze [as _onTimeout] (/Users/romy/github/noobaa-core/src/util/postgres_client.js:700:13)
/Users/romy/github/noobaa-core/src/util/postgres_client.js:250
        const res = await pg_client.query(q);
                                    ^

TypeError: Cannot read properties of undefined (reading 'query')
    at _do_query (/Users/romy/github/noobaa-core/src/util/postgres_client.js:250:37)
    at PostgresTable.single_query (/Users/romy/github/noobaa-core/src/util/postgres_client.js:679:16)
    at runNextTicks (node:internal/process/task_queues:60:5)
    at listOnTimeout (node:internal/timers:538:9)
    at process.processTimers (node:internal/timers:512:7)
    at async Timeout.vacuumAndAnalyze [as _onTimeout] (/Users/romy/github/noobaa-core/src/util/postgres_client.js:700:13)
```
1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/nsfs.js
+++ b/src/cmd/nsfs.js
@@ -14,8 +14,9 @@ const fs = require('fs');
 const util = require('util');
 const minimist = require('minimist');
 
-require('../server/system_services/system_store').get_instance({ standalone: true });
-
+if (process.env.LOCAL_MD_SERVER === 'true') {
+    require('../server/system_services/system_store').get_instance({ standalone: true });
+}
 //const js_utils = require('../util/js_utils');
 const nb_native = require('../util/nb_native');
 //const schema_utils = require('../util/schema_utils');


naveenpaul1 (approved):
LGTM

src/cmd/nsfs.js
```diff
@@ -14,8 +14,6 @@ const fs = require('fs');
 const util = require('util');
 const minimist = require('minimist');
 
-require('../server/system_services/system_store').get_instance({ standalone: true });
```
tangledbytes, line 17: @romayalon if we will remove it from here then later the `endpoint` will initialize the system_store and that too without `standalone: true` being set (if `LOCAL_MD_SERVER` is set to `true`). 

I think that's why this was added here before other things are included which can potentially bring system_store.

romayalon: @tangledbytes I thought that too, but in NC NSFS - LOCAL_MD_SERVER is false, so I guess it's related to tape only? 
I tested it (LOCAL_MD_SERVER is not set) and it worked
do you know if there are plans of using data buckets on tape soon?

tangledbytes: A recent POC that Martin did, did use data buckets. Lemme quickly check with them again...

romayalon: so sounds like we need it, or maybe in the future, will it work if I'll require systemStore when LOCAL_MD_SERVER=true?

tangledbytes: Yeah! That sounds good!

tangledbytes (approved):

PR #7552, opened 2023-10-29
Log Based Replication | Listing logs using listObjectVersions

liranmauda:
### Explain the changes
- Listing logs using listObjectVersions, We are doing that in order to use KeyMarker for pagination.

### DIFF
--- a/src/server/bg_services/replication_log_parser.js
+++ b/src/server/bg_services/replication_log_parser.js
@@ -49,28 +49,32 @@ async function get_aws_log_candidates(source_bucket_id, rule_id, replication_con
     do {
         const next_log_entry = await aws_get_next_log_entry(s3, logs_bucket, prefix, log_object_continuation_token);
 
-        // save the continuation token for the next iteration
-        log_object_continuation_token = next_log_entry.NextContinuationToken;
-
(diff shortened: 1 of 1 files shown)


src/server/bg_services/replication_log_parser.js
```diff
@@ -49,28 +49,32 @@ async function get_aws_log_candidates(source_bucket_id, rule_id, replication_con
     do {
         const next_log_entry = await aws_get_next_log_entry(s3, logs_bucket, prefix, log_object_continuation_token);
```
achouhan09, line 50: Just a small doubt, in the very first call here how we are evaluating this `log_object_continuation_token` as it must be a `log_key_name` from now and not a `continuation_token`

liranmauda: @achouhan09  you are correct, great catch.
so in the first-ever call, it will work as expected as the continuation_token field in the DB is empty.
But in the case of upgrade, the continuation_token string will go into the KeyMarker param and not work as expected. 
We need to fix that.

liranmauda:
This will not work in all cases, need to reevaluate. closing.

Issue #7551, opened 2023-10-27
HTTP Server not Starting by Default, but logging that it did start


matthewklos1:
### Environment info
- NooBaa Version: **VERSION**
- Platform: RHEL 8.8 on x86 Standalone (Linux swat10 4.18.0-477.27.1.el8_8.x86_64)

### Actual behavior
1. 
HTTP server doesn't start, but logs that it does start and is listening

With --debug 10 there is an entry stating HTTP is not allowed, but later in the log, highlighted in green is says it is listening. Have confirmed with netstat the port (6001) is not listening

```
[L0] core.endpoint.endpoint:: HTTP is not allowed for NSFS
...
[LOG] CONSOLE:: nsfs: listening on 'http://localhost:6001/'
```

### Expected behavior
1.
HTTP server either starts, or if it doesnt start to be logged correctly as not starting

### Steps to reproduce
1.
New install of Noobaa on RHEL 8.8 

### More information - Screenshots / Logs / Other output
The fix to starting the HTTP server is to set allow_http to True in the system.json in the config root. 
```
[root@swat10 config]# cat system.json 
{"current_version":"5.15.0","allow_http":"True","upgrade_history":{"successful_upgrades":[]}}
```

Full Log of the issue: 
```Oct 26 20:22:16 swat10 node[190862]: [nsfs/190862]    [L0] core.endpoint.endpoint:: endpoint: replacing old umask:  22 with new umask:  0
Oct 26 20:22:16 swat10 node[190862]: Oct-26 20:22:16.217 [nsfs/190862]    [L0] core.endpoint.endpoint:: endpoint: replacing old umask:  22 with new umask:  0
Oct 26 20:22:16 swat10 node[190862]: Oct-26 20:22:16.217 [nsfs/190862]    [L0] core.endpoint.endpoint:: endpoint: replacing old umask:  22 with new umask:  0
Oct 26 20:22:16 swat10 node[190862]: [nsfs/190862]    [L0] core.endpoint.endpoint:: Configured Virtual Hosts: []
Oct 26 20:22:16 swat10 node[190862]: Oct-26 20:22:16.218 [nsfs/190862]    [L0] core.endpoint.endpoint:: Configured Virtual Hosts: []
Oct 26 20:22:16 swat10 node[190862]: Oct-26 20:22:16.218 [nsfs/190862]    [L0] core.endpoint.endpoint:: Configured Virtual Hosts: []
[...]

PR #7553, opened 2023-10-30
NC | NSFS | HTTP Server not Starting, but logging mismatch

naveenpaul1:
### Explain the changes
1. https://github.com/noobaa/noobaa-core/issues/7551
2. when `system.json` missing `allow_http` flag, the flow skips HTTP server creation

1. https://github.com/noobaa/noobaa-core/issues/7551

### Testing Instructions:
1. try to run nsfs with or without `allow_http` flag in system.json

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -158,16 +158,15 @@ async function main(options = {}) {
             https_server.setSecureContext(updated_ssl_options);
             https_server_sts.setSecureContext(updated_ssl_options);
         });
-        const allow_http = await endpoint_utils.get_nsfs_system_property('allow_http', options.nsfs_config_root);
-        if (options.nsfs_config_root && !allow_http) {
-            dbg.log0('HTTP is not allowed for NSFS.');
-        } else {
+        if (await is_http_allowed(options.nsfs_config_root)) {
             const http_server = http.createServer(endpoint_request_handler);
             if (http_port > 0) {
                 dbg.log0('Starting S3 HTTP', http_port);
                 await listen_http(http_port, http_server);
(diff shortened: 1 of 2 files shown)


src/endpoint/endpoint.js
```diff
@@ -210,6 +209,16 @@ async function main(options = {}) {
     }
 }
 
+async function is_http_allowed(nsfs_config_root) {
+    const allow_http = await endpoint_utils.get_nsfs_system_property('allow_http', nsfs_config_root);
+        if (nsfs_config_root && allow_http === undefined) {
```
v-harihar, line 214: 214 and 216 conditions can be combined,
something like `if (nsfs_config_root && allow_http === undefined && allow_http)`, since both conditions returing same value. Is n't it ?

src/endpoint/endpoint.js
```diff
@@ -158,16 +158,15 @@ async function main(options = {}) {
             https_server.setSecureContext(updated_ssl_options);
             https_server_sts.setSecureContext(updated_ssl_options);
         });
-        const allow_http = await endpoint_utils.get_nsfs_system_property('allow_http', options.nsfs_config_root);
-        if (options.nsfs_config_root && !allow_http) {
-            dbg.log0('HTTP is not allowed for NSFS.');
-        } else {
+        if (await is_http_allowed(options.nsfs_config_root)) {
```
v-harihar, line 161: is_https_allowed() can be normal (sync) function ? without await ?

naveenpaul1: no @v-harihar `get_nsfs_system_property()` is a async function

vh05 (approved):
lgtm

PR #7555, opened 2023-10-31
Fix - bidirectional validation flow should work with operator flow

aayushchouhan09:
### Explain the changes
1. `validate_replication()` api call was made two times when we try to create obc, first independently and second with `put_bucket_replication()` api call. We updated the `validate_replication()` api code to work as expected.

1. Bidirectional validation check added in core not working as expected with operator code.

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -1853,10 +1853,7 @@ async function put_bucket_replication(req) {
     dbg.log0('put_bucket_replication:', req.rpc_params);
     const bucket = find_bucket(req);
 
-    // fetching all the replication rules already present in the DB
-    const db_repl_rules = await replication_store.instance().get_replication_rules();
-
-    validate_replication(req, db_repl_rules);
+    await validate_replication(req);
     const replication_rules = normalize_replication(req);
 
     const bucket_replication_id = bucket.replication_policy_id;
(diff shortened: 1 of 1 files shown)


src/server/system_services/bucket_server.js
```diff
@@ -1933,7 +1932,7 @@ function validate_replication(req, db_repl_rules) {
         }
 
         // validation for bidirectional replication - blocking bidirectional replication only for matching prefixes
-        if (destination_bucket.unwrap() !== req.rpc_params.name.unwrap()) {
+        if (destination_bucket.unwrap() !== req.rpc_params.name.unwrap() && src_bucket) {
```
achouhan09, line 1935: @dannyzaken @liranmauda This condition here is specific to obc, as for obc `validate_replication()` is called two times one before the creation of src_bucket and one after the src_bucket creation inside `put_bucket_replication()`. Because of this condition check only execute if src_bucket is already present/created.

Thats the best way I found here, pls feel free to provide more suggestions.

liranmauda: I suggest adding the above text as a comment in the code.

dannyzaken: @@achouhan09, I'm not sure I understand. even if the source bucket is not created yet, don't we still want to check the replication_rules against destinations?

achouhan09: Actually, here if src_bucket is not present then we cannot fetch the `src_bucket._id` which was stored in DB replication_rules, as this time src_bucket would be the destination_bucket in this validation check.
Let me know if you need more explaination

achouhan09: @liranmauda comment has beed added as per your suggestion, PTAL.

dannyzaken (approved):

Issue #7482, opened 2023-09-07
NSFS | NC | endpoint metrics should be exposed on prometheus and aggregated to parent of forks


guymguym:
### Environment info
- NooBaa Version: **master**
- Platform: **Linux/mac**

### Actual behavior
1. Currently the endpoint_stats_collector just prints the metrics to log in standalone mode (no rpc_client).
3. In addition, when combined with --forks, the metrics port does not work as expected and will just load balance each request to a different fork.

### Expected behavior
1. We want the metrics to be exposed on prometheus metrics port from the endpoint.
2. We want the forks parent to serve the metrics port and be able to collect from all its children.

### Steps to reproduce
1. node src/cmd/nsfs --metrics_port 7001 .
2. curl http://localhost:7001

naveenpaul1:
code merged to master

PR #7562, opened 2023-11-06
NSFS | NC |endpoint metrics should aggregated fork

naveenpaul1:
### Explain the changes
1. Currently the endpoint_stats_collector just prints the metrics to log in standalone mode (no rpc_client).
2. In addition, when combined with --forks, the metrics port does not work as expected and will just load balance each request to a different fork.
3. NSFS Certificate doc updated

1. https://github.com/noobaa/noobaa-core/issues/7482

### Testing Instructions:

1. start nsfs with multiple forks and Prometheus port
2. Hit Prometheus url and verify the values

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/docs/non_containerized_NSFS.md
+++ b/docs/non_containerized_NSFS.md
@@ -436,6 +436,13 @@ NSFS management CLI command will create both account and bucket dir if it's miss
 
 Non containerized NSFS certificates/ directory location will be under the config_root path. The certificates/ directory should contain SSL files tls.key and tls.crt. System will use a cert from this dir to create a valid HTTPS connection. If cert is missing in this dir a self-signed SSL certificate will be generated. Make sure the path to certificates/ directory is valid before running nsfs command, If the path is invalid then cert flow will fail.
 
+Non containerized NSFS restrict insecure HTTP connections when `allow_http` is set to false in cofig.json. This is not the default behaviour.
+
+## Monitoring
+
+Prometheus metrics port can be passed through the argument `--metrics_port` while executing the nsfs command. 
+NSFS state and output metrics can be fetched from URL `http:{host}:{metrics_port}/metrics/nsfs_stats`.
+
 ## Log and Logrotate
 Noobaa logs are configured using rsyslog and logrotate. RPM will configure rsyslog and logrotate if both are already running. 
 
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -36,11 +36,11 @@ const server_rpc = require('../server/server_rpc');
 const debug_config = require('../util/debug_config');
 const auth_server = require('../server/common_services/auth_server');
 const system_store = require('../server/system_services/system_store');
-const prom_reporting = require('../server/analytic_services/prometheus_reporting');
 const background_scheduler = require('../util/background_scheduler').get_instance();
 const endpoint_stats_collector = require('../sdk/endpoint_stats_collector');
 const { NamespaceMonitor } = require('../server/bg_services/namespace_monitor');
 const { SemaphoreMonitor } = require('../server/bg_services/semaphore_monitor');
+const prom_reporting = require('../server/analytic_services/prometheus_reporting');
 const cluster = /** @type {import('node:cluster').Cluster} */ (
     /** @type {unknown} */ (require('node:cluster'))
 );
@@ -91,12 +91,12 @@ async function main(options = {}) {
     try {
         // the primary just forks and returns, workers will continue to serve
         fork_count = options.forks ?? config.ENDPOINT_FORKS;
-        if (fork_utils.start_workers(fork_count)) return;
+        const metrics_port = options.metrics_port || config.EP_METRICS_SERVER_PORT;
+        if (fork_utils.start_workers(metrics_port, fork_count)) return;
 
         const http_port = options.http_port || Number(process.env.ENDPOINT_PORT) || 6001;
         const https_port = options.https_port || Number(process.env.ENDPOINT_SSL_PORT) || 6443;
-        const https_port_sts = options.https_port_sts || Number(process.env.ENDPOINT_SSL_STS_PORT) || 7443;
-        const metrics_port = options.metrics_port || config.EP_METRICS_SERVER_PORT;
+        const https_port_sts = options.https_port_sts || Number(process.env.ENDPOINT_SSL_PORT_STS) || 7443;
         const endpoint_group_id = process.env.ENDPOINT_GROUP_ID || 'default-endpoint-group';
 
         const virtual_hosts = Object.freeze(
@@ -182,9 +182,9 @@ async function main(options = {}) {
             await listen_http(https_port_sts, https_server_sts);
             dbg.log0('Started STS HTTPS successfully');
         }
-        if (metrics_port > 0) {
+        if (metrics_port > 0 && cluster.isPrimary) {
             dbg.log0('Starting metrics server', metrics_port);
-            await prom_reporting.start_server(metrics_port);
+            await prom_reporting.start_server(metrics_port, false);
             dbg.log0('Started metrics server successfully');
         }
 
--- a/src/sdk/endpoint_stats_collector.js
+++ b/src/sdk/endpoint_stats_collector.js
@@ -6,8 +6,12 @@ const mime = require('mime');
 
 const dbg = require('../util/debug_module')(__filename);
 const prom_report = require('../server/analytic_services/prometheus_reporting');
+const stats_aggregator = require('../server/system_services/stats_aggregator');
 const DelayedCollector = require('../util/delayed_collector');
 const config = require('../../config');
+const cluster = /** @type {import('node:cluster').Cluster} */ (
+    /** @type {unknown} */ (require('node:cluster'))
+);
 
 /**
  * @typedef {{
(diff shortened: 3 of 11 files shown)


src/endpoint/endpoint.js
```diff
@@ -178,11 +177,6 @@ async function main(options = {}) {
             await listen_http(https_port_sts, https_server_sts);
             dbg.log0('Started STS HTTPS successfully');
         }
-        if (metrics_port > 0) {
```
romayalon, line 181: won't this create a bug on containerized env?

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
```diff
@@ -6,15 +6,15 @@ const config = require('../../../../config.js');
 
 class BasePrometheusReport {
     constructor() {
-        this._registry = new this.prom_client.Registry();
+        this._register = this.prom_client.register;
```
romayalon, line 9: can you please explain about this change?

src/sdk/endpoint_stats_collector.js
```diff
@@ -350,6 +353,16 @@ class EndpointStatsCollector {
             this.prom_metrics_report.observe('hub_write_latency', { bucket_name }, hub_write_latency);
         }
     }
+    update_fork_counter() {
+        // add fork related metrics to prometheus
+        const code = `worker_${cluster.worker.id}`;
+        this.prom_metrics_report.inc('fork_counter', {code});
+    }
+}
+if (cluster.isWorker) {
```
romayalon, line 362: can you explain about this change? not sure why we increase this counter every time minute

src/endpoint/endpoint.js
naveenpaul1: @romayalon you are saying in containerized env there is no fork so we will miss the metrics server creation. Is that right?

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
naveenpaul1: @romayalon Here we are creating a new object but we should point to existing register object that hold all the metrics, `this.prom_client.register` also pointing to `Registry` class.

src/sdk/endpoint_stats_collector.js
naveenpaul1: @romayalon I added these metrics to get the number of forks running at a time. I am ok with removing these metrics.

src/endpoint/endpoint.js
romayalon: I think that forks were added specifically for non containerized env, so yes, please check that

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
romayalon: why we should point to an existing register object?

src/sdk/endpoint_stats_collector.js
romayalon: it's ok to have this metric, but wouldn't it increase the fork count every second? not sure how it's working

naveenpaul1: yes, it will increase the count each time the fork runs, in this case, every minute, and the count doesn't have any significance. I am not sure Prometheus metrics can have a key without any value.

src/server/analytic_services/prometheus_reports/base_prometheus_report.js
naveenpaul1: existing register object holds all metrics, when we create a new one in our code it does not have any metrics and returns null inside the cluster metrics call.

src/endpoint/endpoint.js
naveenpaul1: @romayalon  you are correct it was affecting the containerized prometheus deployment, I added the previous code for containerized flow.

src/sdk/endpoint_stats_collector.js
naveenpaul1: removed setIntervel() now the code runs only once.

nimrod-becker:
@liranmauda @dannyzaken can you please review?

src/util/fork_utils.js
```diff
@@ -15,8 +16,7 @@ const cluster = /** @type {import('node:cluster').Cluster} */ (
  * @param {number?} count number of workers to start.
  * @returns {boolean} true if workers were started.
  */
-function start_workers(count = 0) {
-
+function start_workers(metrics_port, count = 0) {
```
romayalon, line 19: can you please replace the order of metrics_port and count, so that count will be first?

src/server/system_services/stats_aggregator.js
```diff
@@ -1355,23 +1363,29 @@ function _new_namespace_nsfs_stats() {
 }
 
 // Will return the current nsfs_io_counters and reset it.
-function get_nsfs_io_stats() {
+function get_nsfs_io_stats(only_current = true) {
```
romayalon, line 1366: what this only_current variable means? not sure it's understood, maybe worth having more informative name

src/server/system_services/stats_aggregator.js
```diff
@@ -1246,6 +1246,14 @@ async function update_nsfs_stats(req) {
     if (_nsfs_counters.fs_workers_stats) _update_fs_stats(_nsfs_counters.fs_workers_stats);
 }
 
+async function standalon_update_nsfs_stats(nsfs_stats) {
+    dbg.log1(`standalon_update_nsfs_stats. nsfs_stats =`, nsfs_stats);
+    const _nsfs_counters = nsfs_stats || {};
```
romayalon, line 1251: can this line be deleted if you will add default {} on the signature? 
async function standalon_update_nsfs_stats(nsfs_stats = {}) {... }

docs/non_containerized_NSFS.md
```diff
@@ -268,7 +268,7 @@ NSFS management CLI command will create both account and bucket dir if it's miss
 
 Non containerized NSFS certificate location is configured in system.json file under the property `nsfs_ssl_cert_dir` and the path should contain SSL files tls.key and tls.crt. System will use a cert from this dir to create a valid HTTPS connection. If cert is missing in this dir a self-signed SSL certificate will be generated. Make sure the path mentioned in `nsfs_ssl_cert_dir` is valid before running nsfs command, If the path is invalid then cert flow will fail.
 
-Non containerized NSFS allow nonsecure HTTP connection only when `allow_http` in system.json is true.
+Non containerized NSFS restrict insecure HTTP connections when `allow_http` is set to false in system.json. This is not the default behaviour.
```
romayalon, line 271: can you please add some instructions about how to fetch the stats and the output etc?

src/endpoint/endpoint.js
```diff
@@ -36,11 +36,14 @@ const server_rpc = require('../server/server_rpc');
...
 const system_store = require('../server/system_services/system_store');
-const prom_reporting = require('../server/analytic_services/prometheus_reporting');
 const background_scheduler = require('../util/background_scheduler').get_instance();
 const endpoint_stats_collector = require('../sdk/endpoint_stats_collector');
 const { NamespaceMonitor } = require('../server/bg_services/namespace_monitor');
 const { SemaphoreMonitor } = require('../server/bg_services/semaphore_monitor');
+const prom_reporting = require('../server/analytic_services/prometheus_reporting');
+const cluster = /** @type {import('node:cluster').Cluster} */ (
+    /** @type {unknown} */ (require('node:cluster'))
+);
```
romayalon, line 46: do we need to uncomment line 47 in bucketspace? 
//if (this.stats) fs_context.report_fs_stats = this.stats.update_fs_stats;

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -94,6 +125,52 @@ async function start_server(
     }
 }
 
+async function metrics_nsfs_stats_handler() {
+    const nsfs_io_stats = {};
```
romayalon, line 129: most of this code is taken from web_server right? can we merge the common logic instead ?

naveenpaul1: @romayalon here the format is in JSON for more readability, but for `web_serve.js` it's in plain HTML format.

src/endpoint/endpoint.js
naveenpaul1: I didn't get you.

src/util/fork_utils.js
naveenpaul1: lint is throwing an error when I try to change the position, `parameter with default value should be last`

src/server/system_services/stats_aggregator.js
naveenpaul1: updated with `reset_nsfs_counters`.

src/cmd/manage_nsfs.js
```diff
@@ -452,7 +452,6 @@ async function add_account_config_file(data, accounts_path, access_keys_path, co
         print_account_usage();
         return;
     }
-
```
dannyzaken, line 455: remove

src/util/fork_utils.js
dannyzaken: I think we can remove the default value. WDYT?

dannyzaken (approved):
Approving the PR. 
since it's pretty large, I think we need more testing before merging, mainly to make sure that it does not affect prometheus reporting in Openshift\K8s

naveenpaul1:
Verified the Prometheus metrics in openshift cluster.

Screenshot from 2023-12-08 18-12-47

PR #7571, opened 2023-11-08
docker build - bump base image to centos stream 9

alphaprinz:
### Explain the changes
1. Upgrade base image to centos stream 9.
2. Adapt build.

1. 

### Testing Instructions:
1. make tests, github actions.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/s3select/s3select.gyp
+++ b/src/native/s3select/s3select.gyp
@@ -13,7 +13,7 @@
             's3select_napi.cpp'
         ],
 	'link_settings': {
-		'libraries': ['/lib64/libboost_thread.so.1.66.0']
+		'libraries': ['/lib64/libboost_thread.so.1.75.0']
         },
 	'variables': {
             'BUILD_S3SELECT_PARQUET%':0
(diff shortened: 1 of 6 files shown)


liranmauda (approved):

src/native/s3select/s3select.gyp
```diff
@@ -13,7 +13,7 @@
             's3select_napi.cpp'
         ],
 	'link_settings': {
-		'libraries': ['/lib64/libboost_thread.so.1.66.0']
+		'libraries': ['/lib64/libboost_thread.so.1.75.0']
```
dannyzaken, line 16: @alphaprinz @liranmauda, does this affect downstream builds?

alphaprinz: Good catch, I've asked Boris.
I think we need to work with a softlink so that specific version won't matter?

liranmauda: Downstream is based on RHEL 9 since 5.13
@b-ranto Will this change effect Downstream? maybe we can have a private build to test is. WDYT?

b-ranto: It won't, we use a hack that manually update the version to the one in RHEL 9:

https://pkgs.devel.redhat.com/cgit/containers/mcg-core/tree/Dockerfile?h=rhodf-4.13-rhel-9#n29

i.e. we have had this patch applied downstream for a while now. :-)

b-ranto (approved):

Issue #7454, opened 2023-08-22
NSFS Object tagging support needed


guymguym:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. NSFS does not store and retrieve object tagging -
https://github.com/noobaa/noobaa-core/blob/06d01f2e4c1ac1cd14368f82e550946ec2e5c23b/src/sdk/namespace_fs.js#L1647-L1662

### Expected behavior
1. Tagging should be supported by nsfs:
https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectTagging.html
2. Tags are mutable, which is more flexible than the immutable object meta headers.
3. Can be stored as simple key-value xattrs with some prefix e.g `user.noobaa.tag.KEY=VALUE`

### Steps to reproduce
1. NA

guymguym:
@nimrod-becker we need this for next release, can it be added to the plan?

nimrod-becker:
ofc

guymguym:
@nimrod-becker next Y

romayalon:
@naveenpaul1 is there anything else or we can close this issue?

naveenpaul1:
no @romayalon, Issue can be closed.

PR #7572, opened 2023-11-08
NSFS | NC | Object Tagging

naveenpaul1:
### Explain the changes
1. Tagging should be supported by nsfs
2. Can be stored as simple key-value xattrs with some prefix e.g `user.noobaa.tag.KEY=VALUE`

1. https://github.com/noobaa/noobaa-core/issues/7454

### Testing Instructions:
1. try to put, get and delete object tag with bellow comands
`aws s3api put-object-tagging --tagging '{"TagSet": [{ "Key": "designation", "Value": "confidential" }, { "Key": "department", "Value": "finance" }, { "Key": "team", "Value": "payroll" } ]}' --bucket {bucke-name} --key {object-key} --endpoint https://localhost:6443 --no-verify-ssl`

`aws s3api get-object-tagging --bucket {bucke-name} --key {object-key}--endpoint https://localhost:6443 --no-verify-ssl`

`aws s3api delete-object-tagging --bucket {bucke-name} --key {object-key} --endpoint https://localhost:6443 --no-verify-ssl`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -247,6 +247,8 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             _.isUndefined(account.nsfs_account_config.gid)) return false;
         try {
             dbg.log0('_has_access_to_nsfs_dir: checking access:', ns.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
+            //TODO:  Operation not permitted error on change_user() with non root user, and 
+            //proccess exit with error
             await nb_native().fs.checkAccess({
                 uid: account.nsfs_account_config.uid,
                 gid: account.nsfs_account_config.gid,
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -39,6 +39,7 @@ const buffers_pool = new buffer_utils.BuffersPool({
 
 const XATTR_USER_PREFIX = 'user.';
 const XATTR_NOOBAA_INTERNAL_PREFIX = XATTR_USER_PREFIX + 'noobaa.';
+const XATTR_NOOBAA_CUSTOM_PREFIX = XATTR_NOOBAA_INTERNAL_PREFIX + 'tag.';
 // TODO: In order to verify validity add content_md5_mtime as well
 const XATTR_MD5_KEY = XATTR_USER_PREFIX + 'content_md5';
 const XATTR_CONTENT_TYPE = XATTR_NOOBAA_INTERNAL_PREFIX + 'content_type';
@@ -1803,7 +1804,9 @@ class NamespaceFS {
         await this._delete_path_dirs(file_path, fs_context);
         // when deleting the data of a directory object, we need to remove the directory dir object xattr
         // if the dir still exists - occurs when deleting dir while the dir still has entries in it
-        if (this._is_directory_content(file_path, params.key)) await this._clear_user_xattr(fs_context, this._get_file_md_path(params));
+        if (this._is_directory_content(file_path, params.key)) {
+            await this._clear_user_xattr(fs_context, this._get_file_md_path(params), XATTR_USER_PREFIX);
+        }
     }
 
     ///////////////////////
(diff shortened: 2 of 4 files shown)


src/sdk/namespace_fs.js
```diff
@@ -1777,16 +1780,66 @@ class NamespaceFS {
     ////////////////////
 
     async get_object_tagging(params, object_sdk) {
```
liranmauda, line 1782: The function should also handle versionid, and get the versioned object tagging

src/sdk/namespace_fs.js
```diff
@@ -1777,16 +1780,66 @@ class NamespaceFS {
...
+            }
+        } catch (err) {
+            dbg.error(`namespace_fs.get_object_tagging: failed in dir ${file_path} with error: `, err);
+            throw this._translate_object_error_codes(err);
+        }
+        dbg.log0('namespace_fs.get_object_tagging: return tagging ', tag_set, 'file_path :', file_path);
+        return { tagging: tag_set };
     }
+
     async delete_object_tagging(params, object_sdk) {
```
liranmauda, line 1808: The delete_object_tagging should  handle versionid

src/sdk/namespace_fs.js
```diff
@@ -1777,16 +1780,66 @@ class NamespaceFS {
...
+        const file_path = this._get_file_path(params);
+        const fs_context = this.prepare_fs_context(object_sdk);
+        try {
+            await this._clear_user_xattr(fs_context, file_path, XATTR_NOOBAA_CUSTOM_PREFIX);
+        } catch (err) {
+            dbg.error(`namespace_fs.delete_object_tagging: failed in dir ${file_path} with error: `, err);
+            throw this._translate_object_error_codes(err);
+        }
+        //ideal case put should return version id, but nsfs do not support versioning. adding dummy version.
+        return {version_id: 'v1'};
```
liranmauda, line 1819: I don't think that this is the case, If I recall we do support versioning in NSFS
@romayalon to keep me honest

src/sdk/namespace_fs.js
```diff
@@ -1777,16 +1780,66 @@ class NamespaceFS {
...
     async put_object_tagging(params, object_sdk) {
-        // TODO
-        return { tagging: [] };
+        let fs_xattr;
+        const tagging = params.tagging && Object.fromEntries(params.tagging.map(tag => ([tag.key, tag.value])));
+        for (const [xattr_key, xattr_value] of Object.entries(tagging)) {
+            fs_xattr = Object.assign(fs_xattr || {}, {
+                [XATTR_NOOBAA_CUSTOM_PREFIX + xattr_key]: xattr_value,
+            });
+        }
```
liranmauda, line 1829: I think we can avoid Object.assign by initiating the fs_xattr as an object:
```suggestion
        let fs_xattr = {};
        const tagging = params.tagging && Object.fromEntries(params.tagging.map(tag => ([tag.key, tag.value])));
        for (const [xattr_key, xattr_value] of Object.entries(tagging)) {
              fs_xattr[XATTR_NOOBAA_CUSTOM_PREFIX + xattr_key] = xattr_value;
        }
```
Am I missing something?

src/sdk/namespace_fs.js
```diff
@@ -1777,16 +1780,66 @@ class NamespaceFS {
...
+            await this._clear_user_xattr(fs_context, file_path, XATTR_NOOBAA_CUSTOM_PREFIX);
+        } catch (err) {
+            dbg.error(`namespace_fs.delete_object_tagging: failed in dir ${file_path} with error: `, err);
+            throw this._translate_object_error_codes(err);
+        }
+        //ideal case put should return version id, but nsfs do not support versioning. adding dummy version.
+        return {version_id: 'v1'};
     }
+
     async put_object_tagging(params, object_sdk) {
```
liranmauda, line 1822: Same as in all the ops, I thnk we should handle versions.

src/sdk/namespace_fs.js
```diff
@@ -1777,16 +1780,66 @@ class NamespaceFS {
...
+        try {
+            // remove existng tag before putting new tags
+            await this._clear_user_xattr(fs_context, file_path, XATTR_NOOBAA_CUSTOM_PREFIX);
+            await this.set_fs_xattr_op(fs_context, file_path, fs_xattr, undefined);
+        } catch (err) {
+            dbg.error(`namespace_fs.put_object_tagging: failed in dir ${file_path} with error: `, err);
+            throw this._translate_object_error_codes(err);
+        }
+        //ideal case put should return version id, but nsfs do not support versioning. adding dummy version.
+        return { tagging: [], version_id: 'v1' };
```
liranmauda, line 1842: Not sure it is the case.

src/sdk/namespace_fs.js
nimrod-becker: While not part of the next version, it is available upstream and it should be handled in the code, I agree

src/sdk/namespace_fs.js
nimrod-becker: While not part of the next version, it is available upstream and it should be handled in the code, I agree

src/sdk/namespace_fs.js
nimrod-becker: While not part of the next version, it is available upstream and it should be handled in the code, I agree

naveenpaul1: @liranmauda version changes are handled in latest patch.

src/endpoint/s3/ops/s3_put_object_tagging.js
```diff
@@ -11,7 +11,8 @@ async function put_object_tagging(req, res) {
     const reply = await req.object_sdk.put_object_tagging({
         bucket: req.params.bucket,
         key: req.params.key,
-        tagging: tag_set
+        tagging: tag_set,
+        version_id: req.query.versionId,
```
liranmauda, line 15: Actually, I think I missed that one, I think that putting is not supported in was, and we should also not support it.
can you revert the version id handle in put object tagging? 
one thing to notice, is that namespace_s3 does add this into it's params, we should consider removing it from there (as a different task in a different PR).

naveenpaul1: ok, Will remove version related changes from `put_object_tagging()`

src/sdk/namespace_fs.js
```diff
@@ -1826,16 +1829,73 @@ class NamespaceFS {
...
+            file_path = this._get_file_path(params);
+        }
+        const fs_context = this.prepare_fs_context(object_sdk);
+        try {
+            await this._clear_user_xattr(fs_context, file_path, XATTR_NOOBAA_CUSTOM_PREFIX);
+        } catch (err) {
+            dbg.error(`NamespaceFS.delete_object_tagging: failed in dir ${file_path} with error: `, err);
+            throw this._translate_object_error_codes(err);
+        }
+        //ideal case put should return version id, but nsfs do not support versioning. adding dummy version.
```
liranmauda, line 1877: We can remove this comment

liranmauda (approved):
LGTM

src/sdk/bucketspace_fs.js
```diff
@@ -247,6 +247,8 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             _.isUndefined(account.nsfs_account_config.gid)) return false;
         try {
             dbg.log0('_has_access_to_nsfs_dir: checking access:', ns.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
+            //TODO:  Operation not permitted error on change_user() with non root user, and
```
romayalon, line 250: @naveenpaul1 can you explain what is the problem here? this should work with non root user

PR #7573, opened 2023-11-09
Support role config deletion

Neon-White:
### Explain the changes
This PR adds support for a role config deletion flag that can be set to `true` if a user wishes to delete an account's role config

This PR is necessary for [operator PR 1238](https://github.com/noobaa/noobaa-operator/pull/1238)

### DIFF
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -206,6 +206,9 @@ module.exports = {
                     role_config: {
                         $ref: 'common_api#/definitions/role_config'
                     },
+                    remove_role_config: {
+                        type: 'boolean'
+                    },
                 }
             },
             auth: {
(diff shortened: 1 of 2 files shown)


src/api/common_api.js
```diff
@@ -1342,6 +1342,9 @@ module.exports = {
                 },
                 assume_role_policy: {
                     $ref: '#/definitions/assume_role_policy'
+                },
+                delete: {
```
dannyzaken, line 1346: Although it was my suggestion, I think we shuldn't add the `delete` here 🙂  
The `role_config` is also used in the `account_schema` and it might cause a confusion. 
Maybe better to add a `delete_role_config` flag in the `update_account` params as you initially suggested

dannyzaken (approved):

PR #7576, opened 2023-11-10
Noobaa/Bucket Logging : Log delivery to noobaa core

aspandey:
Whenever an object operation comes to endpoint, handle_request function process it and logs it to its /var/log/messages.

We need these logs to be sent on noobaa core so that we can process it further and accordingly write it to log bucket as log objects.

In this code change we are sending logs to noobaa core via noobaa-syslog service.

Credit goes to aprinzse@redhat.com for the code contribution.

### DIFF
--- a/src/deploy/NVA_build/rsyslog.conf
+++ b/src/deploy/NVA_build/rsyslog.conf
@@ -12,8 +12,8 @@ $ModLoad imuxsock # provides support for local system logging (e.g. via logger c
 #$ModLoad immark  # provides --MARK-- message capability
 
 # Provides UDP syslog reception
-#$ModLoad imudp
-#$UDPServerRun 514
+$ModLoad imudp
+$UDPServerRun 514
 
 # Provides TCP syslog reception
 #$ModLoad imtcp
@@ -48,6 +48,11 @@ $OmitLocalLogging off
 # Log all kernel messages to the console.
 # Logging much else clutters up the screen.
 #kern.*                                                 /dev/console
+
+if $msg contains '{"noobaa_bucket_logging":"true"' then {
+    action(type="omfile" file="/var/log/bucket_logs.log")
+    stop
+}
 
 # Log anything (except mail) of level info or higher.
 # Don't log private authentication messages!
--- a/src/endpoint/s3/s3_bucket_logging.js
+++ b/src/endpoint/s3/s3_bucket_logging.js
@@ -3,6 +3,8 @@
 
 const dbg = require('../../util/debug_module')(__filename);
 const http_utils = require('../../util/http_utils');
+const dgram = require('node:dgram');
+const { Buffer } = require('node:buffer');
 
 
 async function send_bucket_op_logs(req, res) {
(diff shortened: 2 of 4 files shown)


src/deploy/NVA_build/NooBaa.Dockerfile
```diff
@@ -158,6 +158,8 @@ ENV LD_PRELOAD /usr/lib64/libjemalloc.so.2
 #RUN mkdir -p /nsfs/fs1/amitpb && chmod -R 777 /nsfs/
 #RUN mkdir -p /nsfsAA/fs1/amitpb && chmod -R 777 /nsfsAA/
 
+RUN chmod 777 /var/log/messages
```
alphaprinz, line 161: Is this needed? Feels like a leftover for testing.

src/endpoint/s3/s3_bucket_logging.js
```diff
@@ -34,6 +36,18 @@ function endpoint_bucket_op_logs(op_name, req, res, source_bucket) {
     const s3_log = get_bucket_log_record(op_name, source_bucket, req, res);
     dbg.log1("Bucket operation logs = ", s3_log);
 
+    const buffer = Buffer.from(JSON.stringify(s3_log));
+    const client = dgram.createSocket('udp4');
```
alphaprinz, line 40: Can we have a single udp socket object instead of re-creating it for each request?

src/endpoint/s3/s3_bucket_logging.js
```diff
@@ -34,6 +36,18 @@ function endpoint_bucket_op_logs(op_name, req, res, source_bucket) {
...
+    //const client = dgram.createSocket(req.object_sdk.rpc_client.rpc.router.syslog);
```
alphaprinz, line 41: Remove comments (also below)

alphaprinz (changes requested):
@dannyzaken  I'm not sure if this is a requirement for now, but I think we should redirect these logs to a specific file in /log (and not to default /var/log/messages).

src/deploy/NVA_build/NooBaa.Dockerfile
aspandey: done

src/endpoint/s3/s3_bucket_logging.js
aspandey: done

src/endpoint/s3/s3_bucket_logging.js
aspandey: done

aspandey:
> @dannyzaken I'm not sure if this is a requirement for now, but I think we should redirect these logs to a specific file in /log (and not to default /var/log/messages).

I have modified rsyslog.conf to save logs in different file, /var/log/bucket_logs.log

src/endpoint/s3/s3_bucket_logging.js
```diff
@@ -34,6 +53,15 @@ function endpoint_bucket_op_logs(op_name, req, res, source_bucket) {
     const s3_log = get_bucket_log_record(op_name, source_bucket, req, res);
     dbg.log1("Bucket operation logs = ", s3_log);
 
+    const buffer = Buffer.from(JSON.stringify(s3_log));
+    const client = create_syslog_udp_socket();
+    const url = new URL(req.object_sdk.rpc_client.rpc.router.syslog);
```
alphaprinz, line 58: Can we make this url parsing and port parsing in line below also one-time instead of for each request?

src/deploy/NVA_build/rsyslog.conf
```diff
@@ -49,6 +49,21 @@ $OmitLocalLogging off
 # Logging much else clutters up the screen.
 #kern.*                                                 /dev/console
 
+if $msg contains '{"op":"PUT"' then {
```
alphaprinz, line 52: These are the only actions we want to log?

aspandey: I think so for now we are logging these ops only.

src/endpoint/s3/s3_bucket_logging.js
aspandey: will do that.

alphaprinz (approved):

src/deploy/NVA_build/rsyslog.conf
dannyzaken, line 52: I suggest selecting messages based on a single tag that will be added to all bucket logging messages, instead of selecting only specific ops

aspandey: Makes sense. will modify it

PR #7591, opened 2023-11-13
NSFS | NC | Configuration | accounts/ and access_keys/ directories changes

romayalon:
### Explain the changes
1. replace the current accounts/access_key.json structure with accounts/name.json and the corresponding access_keys/access_key.symlink - 
```
/etc/noobaa.conf.d/accounts/alice.json  --> { "name": "alice", "access_keys": [{ "access_key": "ABCD", "secret_key": "***" }], ... }
/etc/noobaa.conf.d/access_keys/ABCD.symlink --> /etc/noobaa.conf.d/accounts/alice.json
```
3. fix account update name/access key
4. fix bucket update name
5. unit tests addition

### Testing Instructions:
1. sudo  node ./node_modules/mocha/bin/mocha /Users/username/github/noobaa-core/src/test/unit_tests/test_bucketspace_fs.js
2. sudo  node ./node_modules/mocha/bin/mocha /Users/username/github/noobaa-core/src/test/unit_tests/test_nc_nsfs_cli.js

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -10,6 +10,7 @@ const P = require('../util/promise');
 const _ = require('lodash');
 const config = require('../../config');
 const native_fs_utils = require('../util/native_fs_utils');
+const nb_native = require('../util/nb_native');
 
 const HELP = `
 Help:
@@ -44,14 +45,15 @@ Account Options:
     # required for add, update 
     --name <name>               (default none)                              Set the name for the account.
     --email <email>             (default none)                              Set the email for the account.
+    --new_name <name>           (default none)                              Set a new name for the account (update).
     --uid <uid>                 (default as process)                        Send requests to the Filesystem with uid.
     --gid <gid>                 (default as process)                        Send requests to the Filesystem with gid.
-    --user <distinguished_name> (default none)                              Send requests to the Filesystem with distinguished_name (overrides uid/gid).
     --secret_key <key>          (default none)                              The secret key pair for the access key.
     --new_buckets_path <dir>    (default none)                              Set the filesystem's root where each subdir is a bucket.
     
     # required for add, update, and delete
     --access_key <key>          (default none)                              Authenticate incoming requests for this access key only (default is no auth).
+    --new_access_key <key>      (default none)                              Set a new access key for the account.
     --config_root <dir>         (default config.NSFS_NC_DEFAULT_CONF_DIR)   Configuration files path for Noobaa standalon NSFS.
 
     # Used for list
@@ -72,6 +74,7 @@ Bucket Options:
 
     # required for add, update, and delete
     --name <name>               (default none)                              Set the name for the bucket.
+    --new_name <name>           (default none)                              Set a new name for the bucket.
     --config_root <dir>         (default config.NSFS_NC_DEFAULT_CONF_DIR)   Configuration files path for Noobaa standalon NSFS.
 `;
 
@@ -100,12 +103,17 @@ function print_bucket_usage() {
     process.exit(1);
 }
 
+const buckets_dir_name = '/buckets';
+const accounts_dir_name = '/accounts';
+const access_keys_dir_name = '/access_keys';
+
 async function check_and_create_config_dirs(config_root) {
     const bucket_dir_stat = await fs_utils.file_exists(config_root);
     if (!bucket_dir_stat) {
         await fs_utils.create_path(config_root);
-        await fs_utils.create_path(config_root + '/buckets');
-        await fs_utils.create_path(config_root + '/accounts');
+        await fs_utils.create_path(config_root + buckets_dir_name);
+        await fs_utils.create_path(config_root + accounts_dir_name);
+        await fs_utils.create_path(config_root + access_keys_dir_name);
         return true;
     }
     return false;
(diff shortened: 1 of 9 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -238,24 +260,42 @@ async function get_bucket_config_file_status(data, bucket_config_path, config_ro
...
+
+    data.name = data.new_name || cur_name;
+
+    const cur_bucket_config_path = get_config_file_path(bucket_config_path, cur_name.unwrap());
+    const new_bucket_config_path = get_config_file_path(bucket_config_path, data.name.unwrap());
+
+    data = JSON.stringify(_.omit(data, ['new_name']));
+
+    await native_fs_utils.create_config_file(fs_context, bucket_config_path, new_bucket_config_path, data);
+    await native_fs_utils.delete_config_file(fs_context, bucket_config_path, cur_bucket_config_path);
```
naveenpaul1, line 298: I think we should delete the previous file only when creation is success.

src/cmd/manage_nsfs.js
```diff
@@ -334,93 +382,155 @@ async function fetch_account_data(argv, config_root, from_file) {
...
     }
     // TODO: support non root fs context
     const fs_context = get_root_fs_context(config_root_backend);
-    const full_account_config_path = get_config_file_path(accounts_config_path, data.access_keys[0].access_key);
+    const access_key = data.access_keys[0].access_key;
+    const full_account_config_path = get_config_file_path(accounts_path, data.name);
+    const full_account_config_access_key_path = get_symlink_config_file_path(access_keys_path, access_key);
+
+    const name_exists = await config_file_exists(fs_context, full_account_config_path);
+    const access_key_exists = await config_file_exists(fs_context, full_account_config_access_key_path);
```
naveenpaul1, line 431: I think we need the same check for updating the account config file with `new_name` or `new_access_key`.

src/cmd/manage_nsfs.js
romayalon: what do you mean on success? currently it'll throw error if the creation will fail so it won't delete the old path , or I missed something?

src/cmd/manage_nsfs.js
```diff
@@ -431,24 +541,36 @@ async function manage_account_operations(action, data, config_root, config_root_
 
 
 /**
- * list_config_file will list all the config files (json) in a given path
- * @param {fs.PathLike} full_path
+ * list_config_file will list all the config files (json) in a given config directory
+ * @param {string} config_path
```
liranmauda, line 545: Why do we change it to string?
Isn't it always a path?

src/cmd/manage_nsfs.js
```diff
@@ -431,24 +541,36 @@ async function manage_account_operations(action, data, config_root, config_root_
...
  */
-async function list_config_file(full_path) {
-    const entries = await fs.promises.readdir(full_path);
+async function list_config_file(config_path) {
```
liranmauda, line 547: I think this we can change the function name:
```suggestion
async function list_config_files(config_path) {
```

src/cmd/manage_nsfs.js
```diff
@@ -431,24 +541,36 @@ async function manage_account_operations(action, data, config_root, config_root_
...
+ * get_view_config_data will read a config file and return its content ready to be printed
  * @param {fs.PathLike} config_file_path
- * @param {string} config_file_name
  */
-async function get_config_data(config_file_path, config_file_name) {
-    const full_path = get_config_file_path(config_file_path.toString(), config_file_name);
-    const data = await fs.promises.readFile(full_path);
+async function get_view_config_data(config_file_path) {
+    const data = await fs.promises.readFile(config_file_path);
     const resources = _.omit(JSON.parse(data.toString()), ['access_keys']);
```
liranmauda, line 574: We can call `get_config_data` here:

```suggestion
    const resources = _.omit(get_config_data(config_file_path), ['access_keys']);
```

src/cmd/manage_nsfs.js
```diff
@@ -476,6 +598,10 @@ async function validate_bucket_add_args(data) {
         console.error('Error: bucket path is mandatory, please use the --path flag');
         return false;
     }
+    if (!update && data.new_name) {
+        console.error('Error: Bucket new_name can not be used on add command, please remove the --new_name flag');
+        return false;
+    }
```
liranmauda, line 604: We can maybe print a warning that we will ignor this flag and keep with the add,
@romayalon WDYT?

src/test/unit_tests/test_bucketspace_fs.js
```diff
@@ -421,7 +425,16 @@ mocha.describe('bucketspace_fs', function() {
...
     const stat1 = await fs.promises.stat(bucket_config_path);
     assert.equal(stat1.nlink, 1);
 }
+
+
+function get_config_file_path(config_type_path, file_name) {
+    return path.join(config_root, config_type_path, file_name + '.json');
+}
+
+function get_access_key_config_file_path(config_type_path, file_name) {
```
liranmauda, line 438: Can you write a comment?
and also maybe rename the function?
I find the name confusing when the function reurn `.symlink`

src/cmd/manage_nsfs.js
romayalon: this is the signature of join(), - join(...paths: string[]): string;

src/cmd/manage_nsfs.js
romayalon: yes I wanted to do that but then I needed to stringify data again..

src/cmd/manage_nsfs.js
romayalon: that's what I initially wanted to do, but this may cause a confusion to the user, he won't understand why new_name is not there

src/test/unit_tests/test_bucketspace_fs.js
romayalon: rename to what? not sure what's confusing

liranmauda: By the code, I will assume we are getting the symlink path and not the actual file path.
This is the reason I think a comment here is needed, and if it is the case indeed the `get_access_key_config_file_path` is not accurate (but something that involves symlink maybe will do it)

src/cmd/manage_nsfs.js
naveenpaul1: I missed that, ya make sense

src/cmd/manage_nsfs.js
romayalon: thank you, I added it

src/cmd/manage_nsfs.js
```diff
@@ -202,10 +213,11 @@ async function fetch_bucket_data(argv, config_root, from_file) {
 }
 
 async function update_bucket_object(config_root, target) {
-    const account_path = config_root + '/buckets';
+    const bucket_path = config_root + '/buckets';
```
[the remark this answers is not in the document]
romayalon: done

liranmauda (approved):
LGTM

PR #7601, opened 2023-11-15
[NC | NSFS] WAL based tape migrations and recalls

tangledbytes:
### Explain the changes
This PR adds WAL based migrations and recalls to NooBaa. The PR adds 2 commands `manage_nsfs`:
1. `glacier restore` - Uses `eeadm` to restore all the files that were requested for restore.
2. `glacier migrate` - Uses `eeadm` to migrate all the files to the tape.
3. `glacier expiry` - Uses `mmapplypolicy` to migrate the expired files to tape.

NOTE: There were many races but now I hope I am dealing with all of them. What are the races I accounted for?
1. Multiple nodes issuing the `migrate`/`restore` at the same time (maybe cron doing it?).
2. Same nodes issue `migrate` and `restore` at the same time such that both of them interleave which is not problematic on the surface as they don't share the log but it is problematic because eventually they all deal with the same S3 objects. Imagine, (1) Upload a file to GLACIER (2) Issue a `restore-object` (3) Cron schedules `glacier restore` as well as `tape migrate` (4) a file gets "recalled" but then is migrated.
4. A single instance of `migrate` is running (probably because WAL was huge and `eeadm` is slow) while another one got scheduled. Same for recall.

NOTE: This solution does **NOT** allow running `glacier recall` and `glacier migrate` at the same time. It is intentional. One call will block another however `glacier expiry` can run alongside of the aforementioned commands.

- [x] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -722,6 +722,30 @@ config.NSFS_RENAME_RETRIES = 3;
 config.NSFS_VERSIONING_ENABLED = true;
 config.NSFS_UPDATE_ISSUES_REPORT_ENABLED = true;
 
+config.NSFS_GLACIER_LOGS_DIR = '/var/run/noobaa-nsfs/wal';
+config.NSFS_GLACIER_LOGS_MAX_INTERVAL = 15 * 60 * 1000;
+
+// NSFS_GLACIER_ENABLED can override internal autodetection and will force
+// the use of restore for all objects.
+config.NSFS_GLACIER_ENABLED = false;
+config.NSFS_GLACIER_LOGS_ENABLED = true;
+config.NSFS_GLACIER_BACKEND = 'TAPECLOUD';
+
+// TAPECLOUD Glacier backend specific configs
+config.NSFS_GLACIER_TAPECLOUD_BIN_DIR = '/opt/ibm/tapecloud/bin';
+
+// NSFS_GLACIER_MIGRATE_INTERVAL indicates the interval between runs
+// of `manage_nsfs glacier migrate`
+config.NSFS_GLACIER_MIGRATE_INTERVAL = 15 * 60 * 1000;
+
+// NSFS_GLACIER_RESTORE_INTERVAL indicates the interval between runs
+// of `manage_nsfs glacier restore`
+config.NSFS_GLACIER_RESTORE_INTERVAL = 15 * 60 * 1000;
+
+// NSFS_GLACIER_EXPIRY_INTERVAL indicates the interval between runs
+// of `manage_nsfs glacier expiry`
+config.NSFS_GLACIER_EXPIRY_INTERVAL = 12 * 60 * 60 * 1000;
+
 ////////////////////////////
 // NSFS NON CONTAINERIZED //
 ////////////////////////////
@@ -742,11 +766,8 @@ config.BASE_MODE_CONFIG_DIR = 0o700;
 
 config.NSFS_WHITELIST = [];
 
-// NSFS_RESTORE_ENABLED can override internal autodetection and will force
-// the use of restore for all objects.
-config.NSFS_RESTORE_ENABLED = false;
-config.NSFS_HEALTH_ENDPOINT_RETRY_COUNT = 3
-config.NSFS_HEALTH_ENDPOINT_RETRY_DELAY = 10
+config.NSFS_HEALTH_ENDPOINT_RETRY_COUNT = 3;
+config.NSFS_HEALTH_ENDPOINT_RETRY_DELAY = 10;
 
 //Quota
 config.QUOTA_LOW_THRESHOLD = 80;
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -18,11 +18,12 @@ const ManageCLIError = require('../manage_nsfs/manage_nsfs_cli_errors').ManageCL
 const NSFS_CLI_ERROR_EVENT_MAP = require('../manage_nsfs/manage_nsfs_cli_errors').NSFS_CLI_ERROR_EVENT_MAP;
 const ManageCLIResponse = require('../manage_nsfs/manage_nsfs_cli_responses').ManageCLIResponse;
 const NSFS_CLI_SUCCESS_EVENT_MAP = require('../manage_nsfs/manage_nsfs_cli_responses').NSFS_CLI_SUCCESS_EVENT_MAP;
+const manage_nsfs_glacier = require('../manage_nsfs/manage_nsfs_glacier');
 const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
 const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
 const { print_usage } = require('../manage_nsfs/manage_nsfs_help_utils');
 const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE,
-    LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS} = require('../manage_nsfs/manage_nsfs_constants');
+    LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS, GLACIER_ACTIONS } = require('../manage_nsfs/manage_nsfs_constants');
 const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 
 function throw_cli_error(error_code, detail, event_arg) {
@@ -105,6 +106,8 @@ async function main(argv = minimist(process.argv.slice(2))) {
             await bucket_management(argv, from_file);
         } else if (type === TYPES.IP_WHITELIST) {
             await whitelist_ips_management(argv);
+        } else if (type === TYPES.GLACIER) {
+            await glacier_management(argv);
         } else {
             // we should not get here (we check it before)
             throw_cli_error(ManageCLIError.InvalidType);
@@ -822,6 +825,8 @@ function validate_type_and_action(type, action) {
         if (!Object.values(ACTIONS).includes(action)) throw_cli_error(ManageCLIError.InvalidAction);
     } else if (type === TYPES.IP_WHITELIST) {
         if (action !== '') throw_cli_error(ManageCLIError.InvalidAction);
+    } else if (type === TYPES.GLACIER) {
+        if (!Object.values(GLACIER_ACTIONS).includes(action)) throw_cli_error(ManageCLIError.InvalidAction);
     }
 }
 
@@ -838,6 +843,8 @@ function validate_no_extra_options(type, action, input_options) {
         valid_options = VALID_OPTIONS.bucket_options[action];
     } else if (type === TYPES.ACCOUNT) {
         valid_options = VALID_OPTIONS.account_options[action];
+    } else if (type === TYPES.GLACIER) {
+        valid_options = VALID_OPTIONS.glacier_options[action];
     } else {
         valid_options = VALID_OPTIONS.whitelist_options;
     }
@@ -941,6 +948,26 @@ function _validate_access_keys(argv) {
             check_symbols: true,
         })) throw_cli_error(ManageCLIError.AccountSecretKeyFlagComplexity);
 
+}
+async function glacier_management(argv) {
+    const action = argv._[1] || '';
+    await manage_glacier_operations(action, argv);
+}
+
+async function manage_glacier_operations(action, argv) {
+    switch (action) {
+        case GLACIER_ACTIONS.MIGRATE:
+            await manage_nsfs_glacier.process_migrations();
+            break;
+        case GLACIER_ACTIONS.RESTORE:
+            await manage_nsfs_glacier.process_restores();
+            break;
+        case GLACIER_ACTIONS.EXPIRY:
+            await manage_nsfs_glacier.process_expiry();
+            break;
+        default:
+            throw_cli_error(ManageCLIError.InvalidGlacierOperation);
+    }
 }
 
 exports.main = main;
--- a/src/manage_nsfs/manage_nsfs_glacier.js
+++ b/src/manage_nsfs/manage_nsfs_glacier.js
@@ -1,0 +1,166 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+
+const path = require('path');
+const { PersistentLogger } = require('../util/persistent_logger');
+const config = require('../../config');
+const nb_native = require('../util/nb_native');
+const { GlacierBackend } = require('../sdk/nsfs_glacier_backend/backend');
+const { getGlacierBackend } = require('../sdk/nsfs_glacier_backend/helper');
+const native_fs_utils = require('../util/native_fs_utils');
+
+const CLUSTER_LOCK = 'cluster.lock';
+const SCAN_LOCK = 'scan.lock';
+
+const RESTORE_TIMESTAMP_FILE = 'restore.timestamp';
+const MIGRATE_TIMESTAMP_FILE = 'migrate.timestamp';
+const EXPIRY_TIMESTAMP_FILE = 'expiry.timestamp';
+
+async function process_migrations() {
+    const fs_context = native_fs_utils.get_process_fs_context();
+
+    await lock_and_run(fs_context, CLUSTER_LOCK, async () => {
+        if (
+            await low_free_space() ||
+            await time_exceeded(fs_context, config.NSFS_GLACIER_MIGRATE_INTERVAL, MIGRATE_TIMESTAMP_FILE)
+        ) {
+            await run_glacier_migrations(fs_context);
+            await record_current_time(fs_context, MIGRATE_TIMESTAMP_FILE);
+        }
+    });
+}
+
+/**
+ * run_tape_migrations reads the migration WALs and attempts to migrate the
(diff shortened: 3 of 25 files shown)


tangledbytes:
Ah, there is still one possible race here:
1. P1 opens a file
2. P2 renames the file
3. Someone issues a `tape migrate` which gets exclusive access to the recently renamed file
4. P1 gets blocked as `tape migrate` as the exclusive access
5. `tape migrate` unlinks the file on completion
6. P1 gets unblocked, happily performs write (the file won't be deleted as there is an open FD).
7. P1 closes the FD upon swap, the file gets deleted

RESULT: Lost writes.

Working on it now...

config.js
```diff
@@ -690,6 +690,11 @@ config.NSFS_RENAME_RETRIES = 3;
 config.NSFS_VERSIONING_ENABLED = true;
 config.NSFS_UPDATE_ISSUES_REPORT_ENABLED = true;
 
+config.NSFS_WAL_DIR = '/var/log/noobaa-nsfs/wal';
+config.NSFS_WAL_MAX_INTERVAL = 15 * 60 * 1000;
+
+config.TAPECLOUD_BIN_DIR = '/opt/ibm/tapecloud/bin';
```
guymguym, line 696:
```suggestion
config.NSFS_GLACIER_BIN_DIR = '/opt/ibm/tapecloud/bin';
```

src/cmd/manage_nsfs.js
```diff
@@ -14,19 +14,26 @@ const SensitiveString = require('../util/sensitive_string');
...
+const eeadm = require('../util/tapecloud_utils');
+const { parse_decimal_int } = require('../endpoint/s3/s3_utils');
+const os = require('os');
 
 const TYPES = {
     ACCOUNT: 'account',
     BUCKET: 'bucket',
-    IPWHITELIST: 'whitelist'
+    IPWHITELIST: 'whitelist',
+    TAPE: 'tape',
```
guymguym, line 26: Perhaps better refer to this as GLACIER instead of tape?
This can go to any media that supports migrate/restore commands...

src/cmd/manage_nsfs.js
```diff
@@ -822,6 +839,318 @@ function _validate_access_keys(argv) {
         })) throw_cli_error(ManageCLIError.AccountSecretKeyFlagComplexity);
 
 }
+async function tape_mangement(argv) {
```
guymguym, line 842: again, I think the terms in the code should be around the functionality (glacier), and not the implementation (tape)
```suggestion
async function glacier_mangement(argv) {
```

src/util/tapecloud_utils.js
```diff
@@ -0,0 +1,86 @@
...
+ * The file should be in the following format: <filename>
+ * 
+ * The function returns the names of the files which failed
+ * to migrate.
+ * @param {string} file filename
+ * @returns {Promise<string[]>} failedfiles
+ */
+async function migrate(file) {
+    try {
+        await asyncexec(`${config.TAPECLOUD_BIN_DIR}/eemigrate ${file}`);
```
guymguym, line 58: I would put each bin name in its own config option and not have it hardcoded

src/util/tapecloud_utils.js
```diff
@@ -0,0 +1,86 @@
...
+    // Find the line in the stdout which has the line 'task ID is, <id>' and extract id
+    const match = stdout.match(/task ID is (\d+)/);
+    if (match.length !== 2) {
+        throw error;
+    }
+
+    const task_id = match[1];
+
+    // Fetch task status and see what failed
+    const { stdout: taskshowstdout } = await asyncexec(`eeadm task show -r ${task_id}`);
```
guymguym, line 25: this command (eeadm task show) should also be in its own wrapped script so we don't have to hard coded it here.

src/cmd/manage_nsfs.js
```diff
@@ -822,6 +839,318 @@ function _validate_access_keys(argv) {
...
+    const action = argv._[1] || '';
+    await manage_tape_operations(action, argv.interval, argv.free_threshold);
+}
+
+async function manage_tape_operations(action, interval, free_threshold) {
```
guymguym, line 847: Can we move all these functions to another source file? it becomes too big.

src/cmd/manage_nsfs.js
```diff
@@ -822,6 +839,318 @@ function _validate_access_keys(argv) {
...
+ * 
+ * NOTE: This assumes that `NSFS_WAL_DIR` is on the same mount and uses
+ * that to determine the free space
+ * @param {nb.NativeFSContext} fs_context
+ * @param {number} interval interval in which the migration should run
+ * @param {number} free_threshold percentage of disk that needs to be free
+ * @param {string} timestamp_file name of the file which tracks timestamp
+ * @returns {Promise<boolean>}
+ */
+async function should_run_eeadm(fs_context, interval, free_threshold, timestamp_file) {
```
guymguym, line 1091: rename to should_migrate? eeadm is an implementation detail

src/cmd/manage_nsfs.js
```diff
@@ -822,6 +839,318 @@ function _validate_access_keys(argv) {
...
+}
+
+/**
+ * lock_and_run acquires a flock and calls the given callback after
+ * acquiring the lock
+ * @param {nb.NativeFSContext} fs_context 
+ * @param {Function} cb 
+ */
+async function lock_and_run(fs_context, cb) {
+    const lockfilename = 'process.lock';
```
guymguym, line 1145: not sure I follow - we need to lock the file that we load and process, not just any file.

src/cmd/manage_nsfs.js
```diff
@@ -822,6 +839,318 @@ function _validate_access_keys(argv) {
...
+ * acquire_file_lock takes a lockfile name and optionally a mode and will
+ * create/open a file and acquire lock on it and return the lock file fd.
+ * @param {nb.NativeFSContext} fs_context 
+ * @param {string} lockfile 
+ * @param {'EXCLUSIVE' | 'SHARED'} mode 
+ * @returns {Promise<nb.NativeFile>}
+ */
+async function acquire_file_lock(fs_context, lockfile, mode = 'EXCLUSIVE') {
+    const lockfd = await nb_native().fs.open(fs_context, path.join(config.NSFS_WAL_DIR, lockfile), 'rw');
+    await lockfd.flock(fs_context, mode);
```
guymguym, line 1132: if this fails we leak the open file. I would suggest to inline this into the lock_and_run function to avoid such an issue.

src/native/fs/fs_napi.cpp
```diff
@@ -1680,6 +1683,34 @@ struct FileFsync : public FSWrapWorker<FileWrap>
     }
 };
 
+struct FileFlock : public FSWrapWorker<FileWrap>
```
guymguym, line 1686: would it make sense to add this as option to open instead?
do you need the UNLOCK option?

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -459,6 +462,11 @@ mocha.describe('namespace_fs', function() {
         const restore_key = 'restore_key_1';
         const data = crypto.randomBytes(100);
 
+        mocha.before(async function() {
+            const dir = await mkdtemp(`${os.tmpdir()}${path.sep}`);
+            config.NSFS_WAL_DIR = dir;
```
guymguym, line 467: why is this needed? can't we mkdir in nsfs instead of relying on someone to create it for us?

src/cmd/manage_nsfs.js
```diff
@@ -822,6 +839,318 @@ function _validate_access_keys(argv) {
...
+ * @param {number} interval interval in which the migration should run
+ * @param {number} free_threshold percentage of disk that needs to be free
+ * @param {string} timestamp_file name of the file which tracks timestamp
+ * @returns {Promise<boolean>}
+ */
+async function should_run_eeadm(fs_context, interval, free_threshold, timestamp_file) {
+    // 1. Run every X mins
+    try {
+        const { data } = await nb_native().fs.readFile(fs_context, path.join(config.NSFS_WAL_DIR, timestamp_file));
+        const lastrun = new Date(data.toString('utf-8'));
```
guymguym, line 1095: buffer.toString() is utf8 by default, no need to specify it

config.js
```diff
@@ -690,6 +690,11 @@ config.NSFS_RENAME_RETRIES = 3;
 config.NSFS_VERSIONING_ENABLED = true;
 config.NSFS_UPDATE_ISSUES_REPORT_ENABLED = true;
 
+config.NSFS_WAL_DIR = '/var/log/noobaa-nsfs/wal';
```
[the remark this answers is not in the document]
tangledbytes: This is used for WALs in general so not specific to Glacier (although logs are used for Glacier only at the moment). I couldn't think of a default location so I just set it to this, we can (and probably should) change this.

src/util/tapecloud_utils.js
tangledbytes: When you say script, do you mean another JS file dedicated to this or a string which simply stores this command?

src/cmd/manage_nsfs.js
tangledbytes: Yeah, I'll do that.

src/cmd/manage_nsfs.js
tangledbytes: This is more like a Golang mutex which doesn't protects anything just the code flow. I use this to ensure that each execution is sequential in nature regardless of nodes/processes which are issuing the calls.

src/native/fs/fs_napi.cpp
tangledbytes: I don't need `UNLOCK` yet. I separated it out from `open` so as to not make it seem like its a one step process (its not and that results in weird races).

src/sdk/nsfs_glacier_backend/generic.js
```diff
@@ -0,0 +1,167 @@
...
+    static XATTR_RESTORE_ONGOING = 'user.noobaa.restore.ongoing';
+
+    /**
+     * XATTR_RESTORE_REQUEST_TEMP is set to the same valuue as XATTR_RESTORE_REQUEST
+     * by a backend as a means to mark the request to be in-flight.
+     * 
+     * Any backend needs to make sure that both the attributes shall NOT be set at the same
+     * time.
+     */
+    static XATTR_RESTORE_REQUEST_TEMP = 'user.noobaa.restore.request.temp';
```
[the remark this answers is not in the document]
tangledbytes: Renamed to `XATTR_RESTORE_REQUEST_STAGED` - sounds fancier.

src/sdk/nsfs_glacier_backend/generic.js
```diff
@@ -0,0 +1,167 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+
+const nb_native = require('../../util/nb_native');
+
+class GlacierBackend {
```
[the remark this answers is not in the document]
tangledbytes: Renamed this to `backend.js` and other one to `helper.js`. I understand that usually `util` and `helper` becomes dump yard for all the functions but couldn't come up with any other sensible name.

src/sdk/namespace_fs.js
```diff
@@ -1252,6 +1246,10 @@ class NamespaceFS {
             fs_xattr = Object.assign(fs_xattr || {}, {
                 [XATTR_STORAGE_CLASS_KEY]: params.storage_class
             });
+
+            if (params.storage_class === s3_utils.STORAGE_CLASS_GLACIER && config.NSFS_GLACIER_ENABLED) {
+                await NamespaceFS.migrate_wal.append(file_path);
```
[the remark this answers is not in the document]
tangledbytes: Done

src/sdk/namespace_fs.js
```diff
@@ -1252,6 +1246,10 @@ class NamespaceFS {
             fs_xattr = Object.assign(fs_xattr || {}, {
                 [XATTR_STORAGE_CLASS_KEY]: params.storage_class
             });
+
+            if (params.storage_class === s3_utils.STORAGE_CLASS_GLACIER && config.NSFS_GLACIER_ENABLED) {
```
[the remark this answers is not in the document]
tangledbytes: Done

src/sdk/namespace_fs.js
```diff
@@ -1169,7 +1160,10 @@ class NamespaceFS {
         }
         let open_path = upload_path || file_path;
 
-        const copy_res = params.copy_source && (await this._try_copy_file(fs_context, params, file_path, upload_path));
+        let copy_res;
+        if (force_copy_fallback) copy_res = copy_status_enum.FALLBACK;
+        else if (params.copy_source) copy_res = await this._try_copy_file(fs_context, params, file_path, upload_path);
```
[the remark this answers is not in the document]
tangledbytes: Done

src/sdk/namespace_fs.js
```diff
@@ -1160,6 +1133,24 @@ class NamespaceFS {
     // and opens upload_path (if exists) or file_path
     // returns upload params - params that are passed to the called functions in upload_object
     async _start_upload(fs_context, object_sdk, file_path, params, open_mode) {
+        let force_copy_fallback = false;
+
+        if (params.copy_source) {
```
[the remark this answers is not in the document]
tangledbytes: Done

src/sdk/namespace_fs.js
```diff
@@ -1160,6 +1133,24 @@ class NamespaceFS {
...
+            const storage_class = s3_utils.parse_storage_class(stat.xattr[XATTR_STORAGE_CLASS_KEY]);
+            const restore_status = this._get_object_restore_status(stat, storage_class);
+
+            if (storage_class === s3_utils.STORAGE_CLASS_GLACIER) {
+                if (restore_status?.ongoing || !restore_status?.expiry_time) {
+                    dbg.warn('_validate_upload: object is not restored yet', file_path, restore_status);
+                    throw new S3Error(S3Error.InvalidObjectState);
+                }
+
+                force_copy_fallback = true;
```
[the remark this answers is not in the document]
tangledbytes: I added it to `_start_upload` for now as I noticed that `objectsdk` removes `params.copy_source` if this returns `false` but I needed that info in the `_start_upload` to perform checks like GLACIER object is in restored state or not, etc.

I have added that `content_type` check though.

src/sdk/nsfs_glacier_backend/eeadm.js
```diff
@@ -0,0 +1,409 @@
...
+WHERE
+    FILE_SIZE > 0
+    AND PATH_NAME LIKE '${bucket_root}/%'
+    AND xattr('user.storage_class') = 'GLACIER'
+    AND (
+            is_premigrated
+            AND
+            CURRENT_TIMESTAMP - TIMESTAMP(SUBSTR(xattr('${GlacierBackend.XATTR_RESTORE_EXPIRY}'), 0, 10)) >= INTERVAL '0' DAYS
+    )
+    AND NOT user_exclude_list
```
guymguym, line 137: consider extracting this from our code into a script in NSFS_GLACIER_EEADM_BIN_DIR.
the script can take the arguments that we want to send, like expiry / bucket_name / etc and our code will not need to maintain.

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -0,0 +1,193 @@
...
+const RESTORE_TIMESTAMP_FILE = 'restore.timestamp';
+const MIGRATE_TIMESTAMP_FILE = 'migrate.timestamp';
+const EXPIRY_TIMESTAMP_FILE = 'expiry.timestamp';
+
+async function process_migrations(interval, free_threshold) {
+    const fs_context = {
+        uid: process.getuid(),
+        gid: process.getgid(),
+        warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
+    };
```
guymguym, line 23: use native_fs_utils.get_process_fs_context

src/sdk/nsfs_glacier_backend/helper.js
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * This module exists so as to export the common function `getGlacierBackend`
+ * 
+ * Keeping this in the generic.js creates cyclic dependency issue.w
+ */
+
+const config = require('../../../config');
+const { TapeCloudGlacierBackend } = require('./tapecloud');
+// eslint-disable-next-line no-unused-vars
+const { GlacierBackend } = require('./backend');
```
guymguym, line 13: still needed?

tangledbytes: Actually that's just for JSDoc so that we can have autocompletes working properly.

src/util/persistent_logger.js
```diff
@@ -0,0 +1,210 @@
...
+    _auto_rotate() {
+        this.swap_lock_file = path.join(this.dir, `swaplock.${this.file}`);
+
+        setInterval(async () => {
+            await this._swap();
+        }, this.cfg.max_interval).unref();
+    }
+
+    async _swap() {
+        if (!this.fh || !this.local_size) return;
```
guymguym, line 118: fix and only return if !this.fh because we need to release the lock

guymguym (approved):

PR #7605, opened 2023-11-16
Add experimental support for STS CLI usage

Neon-White:
### Explain the changes
Currently, NooBaa overrides various AWS signing processes, which eventually leads to some authentication failures - e.g. when trying to use AWS CLI for STS operations.
I fixed the problem by repurposing an already-existing body checksum calculation, and assigning it to `content_sha256_sig` in case it wasn't yet assigned. Since the NooBaa signature process already utilizes it, no further changes were required.

### DIFF
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -265,6 +265,7 @@ function read_request_body(req, options) {
                 }
             } else {
                 req.content_sha256_buf = sha256_buf;
+                req.content_sha256_sig ||= sha256_buf.toString('hex');
             }
             return resolve();
         });


src/util/signature_utils.js
```diff
@@ -230,7 +231,13 @@ function _aws_request(req, region, service) {
             headers_for_sdk[sdk_key] = value;
         }
     }
+    // if req.body exists, convert it to x-www-form-urlencoded
+    let body;
+    if (req.body && service === 'sts') {
```
romayalon, line 236: so do we need to add a body for all sts actions or just to assume role? can you maybe share a link for why this is the needed solution?
Also, doesn't this affect non cli requests? how things are not failing on the sts sdk tests?

src/util/signature_utils.js
```diff
@@ -242,6 +249,20 @@ function _aws_request(req, region, service) {
...
+
+    return queryString;
+  }
+
+  function _convertKey(key) {
+    // Capitalize the first letter and replace underscores with camel case
+    // eslint-disable-next-line no-shadow
+    return key.charAt(0).toUpperCase() + key.slice(1).replace(/_+(\w)/g, (_, p1) => p1.toUpperCase());
+  }
+
```
romayalon, line 265: sts cli tests are needed

dannyzaken (approved):

guymguym (changes requested):

src/util/http_utils.js
```diff
@@ -298,6 +299,10 @@ async function parse_request_body(req, options) {
         try {
             const res = querystring.parse(req.body.toString());
             const renamed = _.mapKeys(res, (value, key) => _.snakeCase(key));
+            // If x-amz-content-sha256 isn't provided, we calculate it ourselves -
+            if (!req.headers['x-amz-content-sha256']) {
+                req.body_sha256_calc_by_nb = AWS.util.crypto.sha256(req.body, 'hex');
```
guymguym, line 304: Hmm, so I see that we already calculate the incoming content sha256 in the end of `read_request_body()` and we save it in `req.content_sha256_buf` so this calculation is actually redundant in any case. 

I am not sure if this fits all cases but it seems that you could just do this generally in read_request_body:
```
                req.content_sha256_buf = sha256_buf;
                req.content_sha256_sig ||= sha256_buf;
```

guymguym (dismissed):

guymguym (approved):
👍

PR #7607, opened 2023-11-19
NSFS | NC | Actions | fix RHEL 9 rpm suffix

romayalon:
### Explain the changes
1. docker build was bumped base image to RHEL9 in https://github.com/noobaa/noobaa-core/pull/7571, so the rpm suffix should be updated as well.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/manual-build-rpm.yaml
+++ b/.github/workflows/manual-build-rpm.yaml
@@ -22,7 +22,7 @@ jobs:
       - name: Prepare Suffix
         id: suffix
         if: ${{ github.event.inputs.tag != '' }}
-        run: echo suffix="_${{ github.event.inputs.tag }}" >> $GITHUB_OUTPUT
+        run: echo suffix="-${{ github.event.inputs.tag }}" >> $GITHUB_OUTPUT
       
       - name: Build RPM
         id: build_rpm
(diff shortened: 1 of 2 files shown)


liranmauda:
@romayalon @tangledbytes 
Few questions here:
1. Is the RPM really bound to a specific RHEL version? 
2. Will the RPM work on both RHEL 8 and RHEL 9? 

In the case that it is not bound, we can remove the suffix of the OC

We need to ensure we support RHEL 9 and RHEL 8.

romayalon:
@liranmauda As I mentioned on slack, currently it only builds RHEL9, if we want to support both RHEL8 and RHEL9 this change https://github.com/noobaa/noobaa-core/pull/7571 should be reverted and we should make it configurable by the user. the issue is that there was no nightly rpm build since this change was merged.

liranmauda:
I don't think we should revert, We should only patch the single line that changes the `'libraries': ['/lib64/libboost_thread.so.1.75.0']` 
We should do it dynamically, On RHEL9 keep it, and on RHEL 8 change it to 1.66 back.
We should also use the correct base for supporting both RHEL versions.

Opened: https://github.com/noobaa/noobaa-core/issues/7608

In this PR case, we should keep the RHEL version in the RPM name, and el9 represents it correctly.

liranmauda (approved):

tangledbytes (approved):

romayalon:
@liranmauda In any case, `make rpm` and the affected actions should be changed when we have both RHEL8 and RHEL9 support

PR #7610, opened 2023-11-20
NSFS | NC | CLI | create config dirs if they don't exists

romayalon:
### Explain the changes
1. create config dirs if they don't exists at the beginning of manage_nsfs and in create_config_file and before symlinking access_key to account

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -108,15 +108,25 @@ const accounts_dir_name = '/accounts';
 const access_keys_dir_name = '/access_keys';
 
 async function check_and_create_config_dirs(config_root) {
-    const bucket_dir_stat = await fs_utils.file_exists(config_root);
-    if (!bucket_dir_stat) {
-        await fs_utils.create_path(config_root);
-        await fs_utils.create_path(config_root + buckets_dir_name);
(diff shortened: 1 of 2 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -108,15 +108,25 @@ const accounts_dir_name = '/accounts';
...
-        await fs_utils.create_path(config_root + accounts_dir_name);
-        await fs_utils.create_path(config_root + access_keys_dir_name);
-        return true;
+    const pre_req_paths = [
+        config_root,
+        path.join(config_root, buckets_dir_name),
+        path.join(config_root, accounts_dir_name),
+        path.join(config_root, access_keys_dir_name)
+    ];
+    for (const p of pre_req_paths) {
```
liranmauda, line 117: Let's make the `p` variable name `path`

liranmauda (approved):
Approve, with one comment

src/cmd/manage_nsfs.js
romayalon: path can't be re defined, it's a const at the beginning of the file

PR #7611, opened 2023-11-20
parameterize centos base image version

alphaprinz:
### Explain the changes
1. Make version of centos in base image of dockerfiles a parameter.

1. Build works for both centos 8 and 9

### Testing Instructions:
1.make noobaa CENTOS_VER=[8,9]
(default is 9) 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/deploy/NVA_build/Base.Dockerfile
+++ b/src/deploy/NVA_build/Base.Dockerfile
@@ -27,6 +27,7 @@ ARG BUILD_S3SELECT=1
 ARG BUILD_S3SELECT_PARQUET=0
 #Clone S3Select and its two submodules, but only if BUILD_S3SELECT=1.
 RUN ./src/deploy/NVA_build/clone_s3select_submodules.sh
+RUN ln -s /lib64/libboost_thread.so.1.66.0 /lib64/libboost_thread.so.1.75.0 || true
 #Pass BUILD_S3SELECT down to GYP native build.
 #S3Select will be built only if this parameter is equal to "1".
 RUN GYP_DEFINES="BUILD_S3SELECT=$BUILD_S3SELECT BUILD_S3SELECT_PARQUET=$BUILD_S3SELECT_PARQUET" npm run build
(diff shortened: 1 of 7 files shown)


src/native/s3select/s3select.gyp
```diff
@@ -13,7 +13,7 @@
             's3select_napi.cpp'
         ],
 	'link_settings': {
-		'libraries': ['/lib64/libboost_thread.so.1.75.0']
+		'libraries': ['/lib64/libboost_thread.so.link']
```
liranmauda, line 16: This change will break the downstream build
We should choose:
1. Ask Boris to change accordingly
OR
2. Not do `RUN ln -s /lib64/libboost_thread.so.1.* /lib64/libboost_thread.so.link` and explicitly do:
```
RUN sed -i -e 's|1.75.0|1.66.0|g' $REMOTE_SOURCE_DIR/noobaa-core/app/src/native/s3select/s3select.gyp
```
When the RHEL version is 8.

src/deploy/NVA_build/Base.Dockerfile
```diff
@@ -27,6 +27,7 @@ ARG BUILD_S3SELECT=1
 ARG BUILD_S3SELECT_PARQUET=0
 #Clone S3Select and its two submodules, but only if BUILD_S3SELECT=1.
 RUN ./src/deploy/NVA_build/clone_s3select_submodules.sh
+RUN ln -s /lib64/libboost_thread.so.1.* /lib64/libboost_thread.so.link
```
liranmauda, line 30: I don't think that using `ln -s` with a wildcard is ok...
I think we should explicitly do:
```
RUN sed -i -e 's|1.75.0|1.66.0|g' $REMOTE_SOURCE_DIR/noobaa-core/app/src/native/s3select/s3select.gyp
```
When the RHEL version is 8.

src/native/s3select/s3select.gyp
alphaprinz: Not sure it would break ds, I think Boris would get the same symlink.
At any rate current version has a softlink from 75 to 66, which I think is the simplest solution without sed.
Hopefully next time we will switch to symlink and then we won't need to worry about version.

src/deploy/NVA_build/Base.Dockerfile
alphaprinz: Changed to symlink from 75 to 66 specifically.
Some more details in gyp comment.

liranmauda (approved):

PR #7615, opened 2023-11-22
manage_nsfs | Generate access keys

### LINKED ISSUE #7580: Accesskey and secretkey should be generated by accepting uid and gid instead of accesskey and secretkey
### Environment info
NooBaa Version: 5.15
Platform: Linux

### Accesskey and secretkey should be generated by accepting uid and gid instead it is expecting accesskey and secretkey
Actual Behaviour:
Command :
[root@ramya-sc-worker noobaa-core]#node src/cmd/manage_nsfs account add --config_root /ibm/fs1/ces/s3-config --name noobaa --email [noobaa@gmail.com](mailto:noobaa@gmail.com) --new_buckets_path /ibm/fs2 --access_key a-abcdefghijklmn123456 --secret_key s-abcdefghijklmn123456
Oct-25 22:38:34.010 [/2817991] [LOG] CONSOLE:: Resource created : a-abcdefghijklmn123456

Expected Behaviour:
Command:
[root@ramya-sc-worker noobaa-core]# node src/cmd/manage_nsfs account add --config_root /ibm/fs1/ces/s3-config --name noobaa1 --email [noobaa@gmail.com](mailto:noobaa@gmail.com) --new_buckets_path /ibm/fs2 --uid 1023 --gid 1024

Oct-25 22:38:34.010 [/2817991] [LOG] CONSOLE:: Resource created : a-abcdefghijklmn123456

ACCESSKEY: a-abcdefghijklmn123456
SECRETKEY: s-abcdefghijklmn123456

liranmauda:
### Explain the changes
- Generate access keys if not provided by the flags
- Added `--regenerate` flag to regenerate the access keys automatically on update
- Printing the access keys after add and update

### BUG
Fixes: https://github.com/noobaa/noobaa-core/issues/7580

### Test instruction 
 1. Run `npx jest test_nc_nsfs_account_cli.test.js`

- [x] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -9,6 +9,7 @@ const fs = require('fs');
 const P = require('../util/promise');
 const _ = require('lodash');
 const config = require('../../config');
+const cloud_utils = require('../util/cloud_utils');
 const native_fs_utils = require('../util/native_fs_utils');
 const nb_native = require('../util/nb_native');
 
@@ -46,14 +47,15 @@ Account Options:
     --name <name>               (default none)                              Set the name for the account.
     --email <email>             (default none)                              Set the email for the account.
     --new_name <name>           (default none)                              Set a new name for the account (update).
-    --uid <uid>                 (default as process)                        Send requests to the Filesystem with uid.
-    --gid <gid>                 (default as process)                        Send requests to the Filesystem with gid.
+    --uid <uid>                 (default none)                              Send requests to the Filesystem with uid.
+    --gid <gid>                 (default none)                              Send requests to the Filesystem with gid.
     --secret_key <key>          (default none)                              The secret key pair for the access key.
     --new_buckets_path <dir>    (default none)                              Set the filesystem's root where each subdir is a bucket.
     
     # required for add, update, and delete
     --access_key <key>          (default none)                              Authenticate incoming requests for this access key only (default is no auth).
     --new_access_key <key>      (default none)                              Set a new access key for the account.
+    --regenerate                (default none)                              When set and new_access_key is not set, will regenerate the access_key
     --config_root <dir>         (default config.NSFS_NC_DEFAULT_CONF_DIR)   Configuration files path for Noobaa standalon NSFS.
 
     # Used for list
@@ -78,28 +80,13 @@ Bucket Options:
     --config_root <dir>         (default config.NSFS_NC_DEFAULT_CONF_DIR)   Configuration files path for Noobaa standalon NSFS.
 `;
 
-function print_usage() {
-    console.warn(HELP);
-    console.warn(USAGE.trimStart());
-    console.warn(ARGUMENTS.trimStart());
-    console.warn(ACCOUNT_OPTIONS.trimStart());
-    console.warn(BUCKET_OPTIONS.trimStart());
-    process.exit(1);
-}
 
-function print_account_usage() {
+function usage({ print_account = false, print_bucket = false }) {
     console.warn(HELP);
     console.warn(USAGE.trimStart());
     console.warn(ARGUMENTS.trimStart());
-    console.warn(ACCOUNT_OPTIONS.trimStart());
-    process.exit(1);
-}
-
-function print_bucket_usage() {
-    console.warn(HELP);
-    console.warn(USAGE.trimStart());
-    console.warn(ARGUMENTS.trimStart());
-    console.warn(BUCKET_OPTIONS.trimStart());
+    if (print_account) console.warn(ACCOUNT_OPTIONS.trimStart());
+    if (print_bucket) console.warn(BUCKET_OPTIONS.trimStart());
     process.exit(1);
 }
 
@@ -134,26 +121,26 @@ async function main(argv = minimist(process.argv.slice(2))) {
         const resources_type = argv._[0] || '';
         if (argv.help || argv.h) {
             if (resources_type === 'account') {
-                return print_account_usage();
+                return usage({ print_account: true });
             } else if (resources_type === 'bucket') {
-                return print_bucket_usage();
+                return usage({ print_bucket: true });
             }
-            return print_usage();
+            return usage({ print_account: true, print_bucket: true });
         }
         if (resources_type === 'account') {
             if (argv.uid && typeof argv.uid !== 'number') {
-                console.error('Error: UID  must be a number');
+                console.error('Error: UID must be a number');
                 return;
             }
             if (argv.gid && typeof argv.gid !== 'number') {
-                console.error('Error: GIT must be a number');
+                console.error('Error: GID must be a number');
                 return;
             }
         }
         const config_root = argv.config_root ? String(argv.config_root) : config.NSFS_NC_DEFAULT_CONF_DIR;
         if (!config_root) {
             console.error('Error: Config dir should not be empty');
-            print_account_usage();
+            usage({ print_account: true });
             return;
         }
         await check_and_create_config_dirs(config_root);
@@ -229,7 +216,7 @@ async function fetch_existing_bucket_data(config_root, target) {
         source = await get_config_data(full_bucket_config_path);
     } catch (err) {
         console.error('NSFS Manage command: Could not find bucket ' + target.name + ' to update');
-        print_bucket_usage();
+        usage({ print_bucket: true });
     }
     const data = _.merge({}, source, target);
     return data;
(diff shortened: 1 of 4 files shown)


liranmauda:
@romayalon I made it a draft as it needs tests and update the docs. 
Let's make sure that this is what we want and add the test and docs accordingly.

romayalon:
@liranmauda In the last HPO meeting we were asked for this feature, and few requests based on this PR were raised - 
1. instead of having a regenerate flag - if access_key and secret_key not provided we want to generate them.
2. We need a generation of access_keys on add account as well
3. we need to return these access keys to the user.

liranmauda:
> @liranmauda In the last HPO meeting we were asked for this feature, and few requests based on this PR were raised -
> 
> 1. We need a generation of access_keys on add account as well - if access_key and secret_key not provided we want to generate them.
> 2. we need to return these access keys to the user.

@romayalon This is exactly what this PR does...

src/cmd/manage_nsfs.js
```diff
@@ -353,25 +344,64 @@ async function account_management(argv, config_root, from_file) {
...
 async function fetch_account_data(argv, config_root, from_file) {
     let data;
+    let generate_access_keys = true;
     const action = argv._[1] || '';
     if (from_file) {
         const raw_data = await fs.promises.readFile(from_file);
         data = JSON.parse(raw_data.toString());
     }
+    let new_access_key = argv.new_access_key;
+    if (action === 'update') {
```
romayalon, line 388: can't we do this whole logic inside set_access_key()?

romayalon:
@liranmauda please add to the docs the regenerate option

src/cmd/manage_nsfs.js
liranmauda: We can, but that means that we need to get `set_access_key()` awareness for the action.
also it means that we will need to change `argv.new_access_key` or return `new_access_key` and we need to make sure we also return generate_access_keys as false due to the fact that if we do update then those keys should not be generated in the `access_keys` fields. 

I think it is more clear this way when the `set_access_key()` is not aware of the action.

liranmauda:
> @liranmauda please add to the docs the regenerate option

@romayalon I do not think we should add each flag to the docs, we have help for it 🙂

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -0,0 +1,287 @@
...
+            assert_account(account, account_options, false);
+            const access_key = account.access_keys[0].access_key;
+            const secret_key = account.access_keys[0].secret_key;
+            expect(access_key).toBeDefined();
+            expect(secret_key).toBeDefined();
+            const account_symlink = await read_config_file(config_root, access_keys_schema_dir, access_key, true);
+            assert_account(account_symlink, account_options);
+        });
+
+        it('cli create account with access_key and without secret_key', async () => {
```
romayalon, line 80: I'm not sure this a valid behavior, are you sure we want to generate only access key / secret_key?

src/cmd/manage_nsfs.js
```diff
@@ -78,28 +80,17 @@ Bucket Options:
...
-    process.exit(1);
-}
-
-function print_bucket_usage() {
+/**
+ * usage will print the usage and help
+ * @param {boolean} print_account will also print the account specific help
+ * @param {boolean} print_bucket will also print the bucket specific help
+ */
+function usage(print_account, print_bucket) {
```
romayalon, line 88:
```suggestion
function usage({ print_account=false, print_bucket = false }) {
```
I think that the suggested signature will be better understood in the calls to this function

src/cmd/manage_nsfs.js
```diff
@@ -353,25 +344,64 @@ async function account_management(argv, config_root, from_file) {
...
+/**
+ * set_access_keys will set the access keys either given as args or generated.
+ * @param {{ access_key: any; secret_key: any; }} argv
+ * @param {boolean} generate a flag for generating the access_keys automatically
+ */
+function set_access_keys(argv, generate) {
+    const { access_key, secret_key } = argv;
+    let generated_access_key;
+    let generated_secret_key;
+    if (generate) {
```
romayalon, line 356: so on add account we will always generate creds even though the user provided them? I think we should avoid it if it's not needed

src/cmd/manage_nsfs.js
liranmauda: I don't like having those defaults.
especially, when it is a boolean and all we do is if(print_account)

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
liranmauda: Well, as we are set now, it is possible regardless of the regenerate code. 
There is no limitation to update only Paris and we can do an update with only one accessKey or secretKey
This code is just following it.
We can create verification that will handle it (again regardless of this PR).
If we want to we should open an issue and do a PR accordingly.

For this PR scope, I think we should keep it (and decide later if we want to handle it).
One more thought is, IMHO, we should not allow the user to give those values at all, and we should actually always generate or regenerate both keys (and remove the access_key/secret_key/new_access_kek completely).

src/cmd/manage_nsfs.js
liranmauda: Well, that was my initial thought also, but as it is a CLI, and this one is small, and we are not calling it a lot (well manually triggering add), I think it is ok. 
We can always check if argv.access_keys was provided, but as stated in the previous comment, I think we should revoke the option to get access_keys, and only generate/regenerate them.

src/cmd/manage_nsfs.js
romayalon: I think it's better than usage `(false, true)` which is not informative. 
these are just options for usage print, I don't see a reason for having 2 parameters instead of just print_options

src/cmd/manage_nsfs.js
```diff
@@ -544,16 +570,17 @@ async function get_account_config_file_status(data, accounts_path, access_keys_p
     }
 }
 
-
 async function manage_account_operations(action, data, config_root, config_root_backend) {
     const accounts_path = path.join(config_root, accounts_dir_name);
     const access_keys_path = path.join(config_root, access_keys_dir_name);
     if (action === 'add') {
         await add_account_config_file(data, accounts_path, access_keys_path, config_root_backend);
```
naveenpaul1, line 577: I think you need to remove the validation for access_key and secret_key.

src/cmd/manage_nsfs.js
```diff
@@ -353,25 +340,64 @@ async function account_management(argv, config_root, from_file) {
...
             creation_date: new Date().toISOString(),
             wide: argv.wide,
             new_name: argv.new_name,
-            new_access_key: argv.new_access_key,
-            access_keys: [{
-                access_key: argv.access_key,
-                secret_key: argv.secret_key
-            }],
+            new_access_key,
+            access_keys: set_access_keys(argv, generate_access_keys),
```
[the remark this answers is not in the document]
liranmauda: @naveenpaul1 We are avoiding it. the `generate_access_keys` flag is changed to false when we do update.

src/cmd/manage_nsfs.js
liranmauda: The validation should be still there if those fields do not exist (bug) then it should fail.
I do agree that it should never fail.

src/cmd/manage_nsfs.js
```diff
@@ -353,25 +340,64 @@ async function account_management(argv, config_root, from_file) {
...
+        access_key: access_key || generated_access_key,
+        secret_key: secret_key || generated_secret_key,
+    }];
+}
+
+/**
+ * Will print the access_keys
+ * @param {{ access_keys: any[]; }} data
+ */
+function print_access_keys(data) {
```
romayalon, line 368: on my [PR](https://github.com/noobaa/noobaa-core/pull/7655) i'm handling writing to stdout and writing access_keys on update/add always and on list and status only if --show_secrets flag was provided so I think you can remove it

src/cmd/manage_nsfs.js
```diff
@@ -46,14 +47,15 @@ Account Options:
...
-    --gid <gid>                 (default as process)                        Send requests to the Filesystem with gid.
+    --uid <uid>                 (default none)                              Send requests to the Filesystem with uid.
+    --gid <gid>                 (default none)                              Send requests to the Filesystem with gid.
     --secret_key <key>          (default none)                              The secret key pair for the access key.
     --new_buckets_path <dir>    (default none)                              Set the filesystem's root where each subdir is a bucket.
     
     # required for add, update, and delete
     --access_key <key>          (default none)                              Authenticate incoming requests for this access key only (default is no auth).
     --new_access_key <key>      (default none)                              Set a new access key for the account.
+    --regenerate                (default none)                              When set and new_access_key is not set, will regenerate the access_key
```
naveenpaul1, line 58: I think mentioning access_key and secret_key makes it more clear.

liranmauda:
Closing in favor of https://github.com/noobaa/noobaa-core/pull/7666

PR #7617, opened 2023-11-27
NSFS | NC | remove stat from verify_gpfs_lib()

romayalon:
### Explain the changes
1. verify_gpfs_lib() had a stat(GPFS_DL_PATH) call which caused EINVAL error. The stat(GPFS_DL_PATH) is actually not needed because stat(GPFS_DL_PATH) is already being called on the initialization of fs_napi.

1. 

### Testing Instructions:
1. tested manually by @janfrode (thanks Jan!)

- [ ] Doc added/updated
- [ ] Tests added


naveenpaul1 (approved):
LGTM

PR #7618, opened 2023-11-27
NSFS | NC | config.json load implementation instead of env file usage

romayalon:
### Explain the changes
1. nsfs_config_schema.js - Added nsfs_config_schema - defines the possible customization for nsfs.js.
2. config.js - added load and reload of config.json functions in order to update config values if needed. 
3. config.js - added the missing nsfs_config_schema configs to config.
4. nsfs.js - replaced some process.env.[KEY] to config[KEY].
5. ssl_utils.js & endpoint.js - removed nsfs_ssl_cert_dir so now we will always look for certificates under /path/to/config_dir/certificates/
6. endpoint.js - moved allow_http from system.json to config.json
7. endpoint_utils.js - removed get_nsfs_system_property() as it's not being used anymore after steps 4 and 5.
8. Updated docs

1. https://issues.redhat.com/browse/MCGI-217

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -10,6 +10,10 @@ const os = require('os');
 const fs = require('fs');
 const path = require('path');
 const assert = require('assert');
+const _ = require('lodash');
+const { default: Ajv } = require('ajv');
+const nsfs_config_schema = require('./src/server/object_services/schemas/nsfs_config_schema');
+const ajv = new Ajv({ verbose: true, allErrors: true });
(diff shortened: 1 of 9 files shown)


liranmauda (approved):
LGTM, just those few docs comments.
Approving.

PR #7626, opened 2023-11-29
Draft - Node 20.9.0 upgrade

alphaprinz:
### Explain the changes
1. Upgrade to nodejs 20.9.0
2. google_cloud.storage.nodejs-common.Util.getUserAgentFromPackageJson no longer exported. Currently commented out.
3. net-ping/nan - too old. Has drop-in replacement draft in net_utils.js.
4. nan 2.18.0 - has compiliation warning, had to remove -Werror in gyp. https://github.com/nodejs/nan/issues/953.
5. encryption tests - namespace creation uses same bucket, for some reason doesn't fail on node18.

Some more deatils on 2-4: https://docs.google.com/document/d/1uDgFDP2_2TYWEwBn5JtjLcStjOmeqYzBwTctY0pRfsc/edit?usp=drive_link
Some more details on 5: https://noobaa.slack.com/archives/C0NSK9ZNE/p1700419078270099
1. Newer nodejs version.

### Testing Instructions:
1. make tests

- [ ] Doc added/updated
- [ ] Tests added


liranmauda:
Hi @alphaprinz 
Thanks for this PR 🙂 

Please divide it into several PRs.

- The modules upgrade should be in a separate PR, prior to this one.
- The modules that require node 20 (are there any?) should be with this PR.
- The replacement of net-ping/nan should also be a separate PR, prior to this one. 

Then, and only then we should have the bump to node 20. 
This is to prevent backporting issues to previous branches, and downstream issues.  

One more question:
- Does this node version bump the major version of NPM from 9 to 10? 
- If it does, we need to make sure that the downstream builds can handle it, by asking @b-ranto to create a private build from this branch.

b-ranto:
The downstream build passed.

b-ranto (approved):

alphaprinz:
Will be split into separate PRs.

PR #7639, opened 2023-12-04
NSFS - fix some small issues

jackyalbo:
### Explain the changes
1. fork will now use the total memory set for the machine divided by the number of forks
2. on fork exit for any reason an event will be sent in addition to killing the main thread. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -20,6 +20,11 @@ const ajv = new Ajv({ verbose: true, allErrors: true });
 // CONTAINER RESOURCES //
 /////////////////////////
 
+config.ENDPOINT_FORKS = Number(process.env.ENDPOINT_FORKS) || 0;
+config.OS_MEM_RESERVED = 2 * 1024 * 1024 * 1024;
+config.OS_MEM_LIMIT_TOTAL = os.totalmem() - config.OS_MEM_RESERVED;
+config.OS_MEM_LIMIT = config.ENDPOINT_FORKS > 0 ? config.OS_MEM_LIMIT_TOTAL / config.ENDPOINT_FORKS : config.OS_MEM_LIMIT_TOTAL;
+
 // The REQUEST is the "minimal" or "guaranteed" amount required for the container's operation.
 // The LIMIT is the "maximal" or "burst" amount that the container can use for extended operations.
 // In Kubernetes for "guaranteed" QOS class REQUEST and LIMIT both have to be provided and equal.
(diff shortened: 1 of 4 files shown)


src/cmd/nsfs.js
```diff
@@ -258,6 +258,7 @@ async function main(argv = minimist(process.argv.slice(2))) {
         const https_port_sts = Number(argv.https_port_sts) || config.ENDPOINT_SSL_STS_PORT;
         const metrics_port = Number(argv.metrics_port) || config.EP_METRICS_SERVER_PORT;
         const forks = Number(argv.forks) || config.ENDPOINT_FORKS;
+        if (forks > 0) process.env.ENDPOINT_FORKS = forks.toString();
```
romayalon, line 261: As I mentioned on Slack, after the new configuration, this is not needed if setting it to config.json.

config.js
```diff
@@ -650,7 +650,9 @@ assert(config.NAMESPACE_CACHING.DEFAULT_BLOCK_SIZE > config.INLINE_MAX_SIZE);
 //////////////////
 
 config.NSFS_BUF_SIZE = 8 * 1024 * 1024;
-config.NSFS_BUF_POOL_MEM_LIMIT = config.BUFFERS_MEM_LIMIT;
+config.NSFS_BUF_POOL_MEM_LIMIT = config.ENDPOINT_FORKS > 0 ?
```
romayalon, line 653: hmmm why this is the final change? asking because the JIRA description specifies having 4 gb memory limit, 3 CPU limit per fork

src/util/fork_utils.js
```diff
@@ -29,6 +31,15 @@ function start_workers(count = 0) {
         cluster.on('exit', (worker, code, signal) => {
             console.warn('WORKER exit', { id: worker.id, pid: worker.process.pid, code, signal });
             console.error('EXIT ON WORKER ERROR');
+            dbg.event({
```
romayalon, line 34: can you please add this reason to the first section of non_containzerized_NSFS_events.md please?
also please add to non_containzerized_NSFS.md the manual handling of this issue

src/cmd/nsfs.js
jackyalbo: used for argv.forks to work. resolving.

config.js
jackyalbo: added the 4GB limit

src/util/fork_utils.js
jackyalbo: not relevant anymore - won't create an event - just restart a new one

romayalon: wait why deleting the event? we still want to notify regarding that fork restart
BTW, what about killing the main thread after a few retries?

romayalon: so as we discussed on slack, let's have a discussion on the restarts and add it in a different PR

romayalon (approved):

PR #7653, opened 2023-12-08
NSFS | NC | Add white list IP table support for S3

naveenpaul1:
### Explain the changes
1. Noobaa should get the CES IP and have a white list in order to allow only incoming requests from a given list of IPs - not having it is a security issue from Spectrum scale perspective

doc : https://ibm.ent.box.com/file/1379004287366?s=bit6jatl19fbmfcov7ldyszoc4ttnbka

1. https://issues.redhat.com/browse/MCGI-209

### Testing Instructions:
1. 

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -33,6 +33,13 @@ Arguments:
     <action>  Action could be add, update, list, status and delete for accounts/buckets.
 `;
 
+const WHITELIST_OPTIONS = `
+Whitelist Options:
+
+    # Read Whitelist IPs and update the configurations.
+    --ips <ips>                       (default none)          Set whitelist ips in array format : '["127.0.0.1", "192.000.10.000", "3002:0bd6:0000:0000:0000:ee00:0033:6778"]'
+`;
+
 const ACCOUNT_OPTIONS = `
 Account Options:
 
@@ -103,6 +110,12 @@ function print_bucket_usage() {
     process.exit(1);
 }
 
+function print_whitelist_usage() {
+    process.stdout.write(HELP);
+    process.stdout.write(WHITELIST_OPTIONS.trimStart());
+    process.exit(1);
+}
+
 const buckets_dir_name = '/buckets';
 const accounts_dir_name = '/accounts';
 const access_keys_dir_name = '/access_keys';
@@ -117,7 +130,7 @@ async function check_and_create_config_dirs(config_root) {
     for (const dir_path of pre_req_dirs) {
         try {
             const fs_context = native_fs_utils.get_process_fs_context();
-            const dir_exists = await config_file_exists(fs_context, dir_path);
+            const dir_exists = await native_fs_utils.config_file_exists(fs_context, dir_path);
             if (dir_exists) {
                 dbg.log1('nsfs.check_and_create_config_dirs: config dir exists:', dir_path);
             } else {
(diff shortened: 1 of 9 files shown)


src/util/http_utils.js
```diff
@@ -671,6 +671,17 @@ function authorize_session_token(req, options) {
         throw new options.ErrorClass(options.error_invalid_token);
     }
 }
+function validate_nsfs_whitelist(req) {
+    const client_ip = parse_client_ip(req);
+    if (config.NSFS_WHITELIST.length === 0) return;
+    for (const whitelist_ip of config.NSFS_WHITELIST) {
+        if (client_ip.includes(whitelist_ip)) {
+            return
```
romayalon, line 679: missing ;

src/test/unit_tests/test_nc_nsfs_cli.js
```diff
@@ -506,6 +507,57 @@ mocha.describe('manage_nsfs cli', function() {
...
+        const ips = ['127.0.0.1', '192.000.10.000', '3002:0bd6:0000:0000:0000:ee00:0033:999'];
+        mocha.before(async () => {
+            await write_config_file(config_root, '', 'config', config_options);
+        });
+        mocha.after(async () => {
+            fs_utils.file_delete(path.join(config_root, 'config.json'));
+        });
+
+        mocha.it('cli add whitelist ips first time', async function() {
+                const cmd = `node src/cmd/manage_nsfs whitelist --config_root ${config_root} --ips '${JSON.stringify(ips)}'`;
```
romayalon, line 522: can you please use exec_manage_cli for re use of code?

src/cmd/manage_nsfs.js
```diff
@@ -733,6 +745,42 @@ function is_undefined(value) {
...
 
+async function whitelist_ips_management(args, config_root) {
+    const ips = args.ips;
+    const is_valid = await validate_whitelist_arg(ips);
+    if (!is_valid) {
+        print_whitelist_usage();
+        return;
+    }
+    const whitelist_ips = JSON.parse(ips);
+    const config_path = path.join(config_root, 'config.json');
```
romayalon, line 756: please update nsfs_config_schema, we are validating the schema of config.json on config.js, so it should fail without it.

naveenpaul1: @romayalon validation is not failing because of a new property, but still I think its a good to have a validation.

src/util/http_utils.js
```diff
@@ -671,6 +672,17 @@ function authorize_session_token(req, options) {
         throw new options.ErrorClass(options.error_invalid_token);
     }
 }
+function validate_nsfs_whitelist(req) {
+    const client_ip = parse_client_ip(req);
+    if (config.NSFS_WHITELIST.length === 0) return;
+    for (const whitelist_ip of config.NSFS_WHITELIST) {
+        if (client_ip.includes(whitelist_ip)) {
```
romayalon, line 679: shouldn't the condition be client_ip === whitelist_ip?
I mean, what if the client_ip is 11.22.33.44 and the white_list_ip is 11.22.33.4?

naveenpaul1: @romayalon, method `parse_client_ip()` returns the IPs like this `::ffff:127.0.0.1` , thats why I added includes()  instead of `===` , Do I need to get IP only by filtering out the unwanted char?

romayalon (approved):

PR #7661, opened 2023-12-17
NSFS | NC | add root permission requirement for health and manage scripts

romayalon:
### Explain the changes
1. As discussed at the latest HPO meeting, we should require root permissions for manage_nsfs execution, also sounds like we won't be able to run health without it as well.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


naveenpaul1 (approved):
LGTM

PR #7662, opened 2023-12-18
NSFS | NC | replace bucketspace_fs user fs_context with root fs_context on config files ops

romayalon:
### Explain the changes
1. config files create/update/delete should always be done by root (for this version, in the future versions we might need to have CONFIG_UID/CONFIG_GID that will do the manipulations on the configuration files)

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


naveenpaul1 (approved):
LGTM

PR #7666, opened 2023-12-19
manage_nsfs | Generate access keys

### LINKED ISSUE #7580: Accesskey and secretkey should be generated by accepting uid and gid instead of accesskey and secretkey

liranmauda:
### Explain the changes
- Generate access keys if not provided by the flags
- Added --regenerate flag to regenerate the access keys automatically on update
- Add test complexity for the access and secret keys
### BUG
Fixes: https://github.com/noobaa/noobaa-core/issues/7580

### Test instruction 
 1. Run `sudo npx jest test_nc_nsfs_account_cli.test.js`

- [x] Tests added

### DIFF
--- a/src/manage_nsfs/manage_nsfs_cli_errors.js
+++ b/src/manage_nsfs/manage_nsfs_cli_errors.js
@@ -134,15 +134,33 @@ ManageCLIError.AccountNameAlreadyExists = Object.freeze({
 //// ACCOUNT ARGUMENTS ERRORS ////
 //////////////////////////////////
 
- ManageCLIError.MissingAccountSecretKeyFlag = Object.freeze({
+ManageCLIError.MissingAccountSecretKeyFlag = Object.freeze({
     code: 'MissingAccountSecretKeyFlag',
-    message: 'Account secret key is mandatory, please use the --secret_key flag',
+    message: 'Account secret key is mandatory, please use the --secret_key flag or --regenerate on update',
     http_code: 400,
 });
 
 ManageCLIError.MissingAccountAccessKeyFlag = Object.freeze({
     code: 'MissingAccountAccessKeyFlag',
-    message: 'Account access key is mandatory, please use the --access_key flag',
+    message: 'Account access key is mandatory, please use the --access_key flag or --regenerate on update on update',
+    http_code: 400,
+});
+
+ManageCLIError.AccountSecretKeyFlagComplexity = Object.freeze({
+    code: 'AccountSecretKeyFlagComplexity',
+    message: 'Account secret length must be 40, and must contain uppercase, lowercase, numbers and symbols',
+    http_code: 400,
+});
+
+ManageCLIError.AccountAccessKeyFlagComplexity = Object.freeze({
+    code: 'AccountAccessKeyFlagComplexity',
+    message: 'Account access key length must be 20, and must contain uppercase and numbers',
+    http_code: 400,
+});
+
+ManageCLIError.NewAccountAccessKeyFlagComplexity = Object.freeze({
+    code: 'NewAccountAccessKeyFlagComplexity',
+    message: 'Account new access key length must be 20, and must contain uppercase and numbers',
     http_code: 400,
 });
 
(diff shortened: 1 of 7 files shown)


src/util/native_fs_utils.js
```diff
@@ -263,7 +263,7 @@ async function create_config_file(fs_context, schema_dir, config_path, config_da
         } catch (err) {
             if (err.code !== 'ENOENT') throw err;
         }
-        dbg.log1('native_fs_utils: create_config_file config_path:', config_path, 'config_data:', config_data, 'is_gpfs:', open_mode);
+        dbg.log1('create_config_file:: create_config_file config_path:', config_path, 'config_data:', config_data, 'is_gpfs:', open_mode);
```
romayalon, line 266: please don't remove the name of the file from the print

src/util/string_utils.js
```diff
@@ -144,6 +148,85 @@ function escape_reg_exp(str) {
...
+function validate_symbols(string, check_symbols) {
+    const symbols_test = symbols_regexp.test(string);
+    if (check_symbols && !symbols_test) {
+        return false;
+    }
+    return true;
+}
+
+
+function validate_complexity(string, options = {}) {
```
romayalon, line 212: missing jsdoc

src/manage_nsfs/manage_nsfs_cli_errors.js
```diff
@@ -134,15 +134,33 @@ ManageCLIError.AccountNameAlreadyExists = Object.freeze({
 //// ACCOUNT ARGUMENTS ERRORS ////
 //////////////////////////////////
 
- ManageCLIError.MissingAccountSecretKeyFlag = Object.freeze({
+ManageCLIError.MissingAccountSecretKeyFlag = Object.freeze({
     code: 'MissingAccountSecretKeyFlag',
-    message: 'Account secret key is mandatory, please use the --secret_key flag',
+    message: 'Account secret key is mandatory, please use the --secret_key flag or --regenerate',
```
romayalon, line 139: please add `--regenerate on update` or something like that in both places

src/util/native_fs_utils.js
liranmauda: dbg print the name of the file, so the result is something like: `...native_fs_utils native_fs_utils:`
changing this will lead to: `...native_fs_utils create_config_file::` which is indicating the function in the file.

romayalon: right, so just remove the double name of function - create_config_file, it appears twice in few different places

liranmauda: Ho, a great catch, missed that.

romayalon (approved):

liranmauda:
@romayalon we are trying CLI account update uid by access key but the code is failing when there is only access_key and no secret key.
I do not think that access_key is an identifier, and I think we should not use it as one.
I think that the test should fail.
I can make it pass by adding the secret_key but it should not be used like that (maybe even force update to use the name flag).
WDYT?

Issue #7625, opened 2023-11-28
NSFS | NC | CLI | account add - new_buckets_path should be optional and set allow_bucket_creation to false if missing


romayalon:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. Currently on account add - new_buckets_path is mandatory.

### Expected behavior
1. new_buckets_path should be optional and set allow_bucket_creation to false if missing. 
Details: 
1. allow_bucket_creation should be added to account_nsfs_schema.
2. need to check if allow_bucket_creation === true in bucketspace_fs on create bucket.

### Steps to reproduce
1. run account add without specifying new_buckets_path.

romayalon:
@ramya-c3 adding request validation label, Naveen merged a fix -  [7682](https://github.com/noobaa/noobaa-core/pull/7682) PR, please close it if verified.

PravinRanjan10:
[root@prscale-12 noobaa-core]# node src/cmd/manage_nsfs account add --name demoo-acc-1 --email abc@email.com --uid 1000 --gid 2000 2>/dev/null
{
  "response": {
    "code": "AccountCreated",
    "reply": {
      "name": "demoo-acc-1",
      "email": "abc@email.com",
      "creation_date": "2024-01-23T09:05:32.613Z",
      "access_keys": [
        {
          "access_key": "wNftlEWdL61o91F4sLdp",
          "secret_key": "M8iO22z515C4HB1iwVDpd0r7w4dDcaFogwHelxwD"
        }
      ],
      "nsfs_account_config": {
        "uid": 1000,
        "gid": 2000
      },
      "allow_bucket_creation": false,
      "_id": "65af815c42aea515175c7b44"
    }
  }
}

PravinRanjan10:
LGTM

PR #7682, opened 2023-12-28
NC | NSFS | new_buckets_path should be optional

naveenpaul1:
### Explain the changes
1. new_buckets_path should be optional and set allow_bucket_creation to false if missing.

1. https://github.com/noobaa/noobaa-core/issues/7625

### Testing Instructions:
1. run account add without specifying new_buckets_path and try to create bucket using that account and verify the error

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -252,8 +252,6 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             _.isUndefined(account.nsfs_account_config.gid)) return false;
         try {
             dbg.log0('_has_access_to_nsfs_dir: checking access:', ns.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
-            //TODO:  Operation not permitted error on change_user() with non root user, and 
-            //proccess exit with error
             await nb_native().fs.checkAccess({
                 uid: account.nsfs_account_config.uid,
                 gid: account.nsfs_account_config.gid,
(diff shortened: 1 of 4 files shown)


src/sdk/bucketspace_fs.js
```diff
@@ -278,6 +276,9 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
 
     async create_bucket(params, sdk) {
         return bucket_semaphore.surround_key(String(params.name), async () => {
+            if (!sdk.requesting_account.allow_bucket_creation) {
+                throw new RpcError('BUCKET_CREATION_NOT_ALLOWED');
```
romayalon, line 280: please change to 'throw new RpcError('UNAUTHORIZED', 'Not allowed to create new buckets');' 
found it in bucket_server :)

src/test/unit_tests/test_bucketspace_fs.js
```diff
@@ -68,6 +69,7 @@ const account_user2 = {
     name: 'user2',
```
romayalon, line 69: can you please add a test that checks a scenario where an account is not authorized to create a bucket?

naveenpaul1: sure, will add that.

romayalon (approved):

shirady:
@naveenpaul1 @romayalon it seems like a configuration change.
If needed it should be also reflected in the creation of the accounts for the Ceph test (currently in PR #7685, in the file [src/deploy/NVA_build/standalone_deploy_nsfs.sh](https://github.com/noobaa/noobaa-core/pull/7685/files#diff-71d8375cc7e11f9ced9a0a8aa2fced4b66ce67f4028748824530a8f3fd2c5c7c)).

romayalon:
@shirady it's not needed, this makes new_buckets_path optional but the ceph accounts need new_buckets_path because they are creating buckets.

liranmauda:
@naveenpaul1 as we talked in dev, tests should be written in jest
@nimrod-becker Should we block this till the tests are written in jest?

Issue #7252, opened 2023-03-30
Support versioned objects in log based replication


tangledbytes:
# Description
As of today (30th March 2023), log based replication implementation does **not** support versioned objects. This lack of support may not effect the correctness of the replication (iff scan based replication exists and is functional).

The following scenario is problematic:
1. Source bucket has versioning enabled.
2. User deletes a versioned object such that the deleted version is  _not_ the latest version.
3. Log based replication considers this a candidate for deletion and deletes the **latest** version from the destination bucket.
4. This gap is identified by the scan based replication and the replicator will replicate the missing object.

This gap has **severe implications** if and whenever we remove scan based replication.

Initial discussion on this are here: #7234

liranmauda:
Fixed By: https://github.com/noobaa/noobaa-core/pull/7508

PR #7688, opened 2024-01-02
Support delete replication flow in bucket_diff - PR 1

aayushchouhan09:
### Explain the changes
1. This is the first PR in support for the delete replication. In this PR we are identifying the keys needs to be deleted for versioning enabled buckets in replication flow inside bucket_diff.

### Testing Instructions:
1. use command : `npx jest test_bucket_diff.test.js`

- [ ] Tests added

### DIFF
--- a/src/server/utils/bucket_diff.js
+++ b/src/server/utils/bucket_diff.js
@@ -68,47 +68,67 @@ class BucketDiff {
             //         { ETag: 'etag1.2', Size: 89317, Key: '1', VersionId: 'v1.2', IsLatest: false, }
             //     ]
             // }
+            //
+            // keys_del_diff_map is {{ [key: string]: Array<object> }} where the Array consist of metadata objects.
+            // for example: {
+            //     "1": [
+            //         { Key: '1', VersionId: "v1.2", IsLatest: true, LastModified: '2022-02-27T10:45:00.000Z' },
+            //         { Key: '1', VersionId: "v1.1", IsLatest: false, LastModified: '2022-01-27T10:45:00.000Z' }
+            //     ]
+            // }
             keys_diff_map: {},
+            keys_del_diff_map: {},
             first_bucket_cont_token: '',
             second_bucket_cont_token: '',
         };
 
         let first_bucket_contents_left;
         let first_bucket_cont_token;
+        let first_bucket_delete_markers;
         //list the objects in the first bucket
         ({
             bucket_contents_left: first_bucket_contents_left,
-            bucket_cont_token: first_bucket_cont_token
+            bucket_cont_token: first_bucket_cont_token,
+            delete_markers: first_bucket_delete_markers
         } = await this.get_objects(this.first_bucket, prefix, max_keys, current_first_bucket_cont_token));
 
         if (first_bucket_cont_token) diff.first_bucket_cont_token = first_bucket_cont_token;
-        if (Object.keys(first_bucket_contents_left).length === 0) return diff;
+        if (Object.keys(first_bucket_contents_left).length === 0 &&
+            Object.keys(first_bucket_delete_markers).length === 0) return diff;
 
         let second_bucket_contents_left;
         let second_bucket_cont_token;
+        let second_bucket_delete_markers;
         let new_second_bucket_cont_token = current_second_bucket_cont_token;
         let keep_listing_second_bucket = true;
 
         while (keep_listing_second_bucket) {
             ({
                 bucket_contents_left: second_bucket_contents_left,
-                bucket_cont_token: new_second_bucket_cont_token
+                bucket_cont_token: new_second_bucket_cont_token,
+                delete_markers: second_bucket_delete_markers
             } = await this.get_objects(this.second_bucket, prefix, max_keys, new_second_bucket_cont_token));
 
             const keys_diff_response = await this.get_keys_diff(
-                first_bucket_contents_left, second_bucket_contents_left, new_second_bucket_cont_token);
+                first_bucket_contents_left, second_bucket_contents_left, new_second_bucket_cont_token,
+                first_bucket_delete_markers, second_bucket_delete_markers);
 
             first_bucket_contents_left = keys_diff_response.keys_contents_left;
+            first_bucket_delete_markers = keys_diff_response.delete_markers_left;
             keep_listing_second_bucket = keys_diff_response.keep_listing_second_bucket;
             diff.keys_diff_map = { ...diff.keys_diff_map, ...keys_diff_response.keys_diff_map };
+            diff.keys_del_diff_map = {...diff.keys_del_diff_map, ...keys_diff_response.keys_del_diff_map};
 
             const first_bucket_key_array = Object.keys(first_bucket_contents_left);
+            const first_bucket_del_array = Object.keys(first_bucket_delete_markers);
             const second_bucket_key_array = Object.keys(second_bucket_contents_left);
 
-            if (first_bucket_key_array.length !== 0 && second_bucket_key_array.length !== 0) {
+            if (first_bucket_key_array.length !== 0 && second_bucket_key_array.length !== 0 ||
+                first_bucket_del_array.length !== 0) {
                 const first_bucket_key_in_last_pos = first_bucket_key_array[first_bucket_key_array.length - 1];
                 const second_bucket_key_in_last_pos = second_bucket_key_array[second_bucket_key_array.length - 1];
-                if (first_bucket_key_in_last_pos >= second_bucket_key_in_last_pos) {
+                if (first_bucket_key_in_last_pos >= second_bucket_key_in_last_pos ||
+                    first_bucket_del_array.length !== 0) {
                     second_bucket_cont_token = new_second_bucket_cont_token;
                 }
             }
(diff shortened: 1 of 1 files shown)


src/server/utils/bucket_diff.js
```diff
@@ -69,38 +69,51 @@ class BucketDiff {
             //     ]
             // }
             keys_diff_map: {},
+            keys_del_arr: [],
```
liranmauda, line 72: I was re thinking, and i think we should use an object here instead of an array. We should carry the delete key with all it's properties as we do in the versions diff.

src/server/utils/bucket_diff.js
```diff
@@ -209,8 +222,10 @@ class BucketDiff {
         dbg.log2('BucketDiff get_objects:: bucket_response', list_objects_response);
         const bucket_contents_left = this._object_grouped_by_key_and_omitted(list_objects_response);
         const bucket_cont_token = this._get_next_key_marker(list_objects_response, bucket_contents_left);
-        dbg.log2('BucketDiff get_objects:: bucket', bucket, 'bucket_contents_left', bucket_contents_left);
-        return { bucket_contents_left, bucket_cont_token };
+        const field = 'DeleteMarker';
```
liranmauda, line 225: This will fail on a bucket that have no version enabled.

src/server/utils/bucket_diff.js
```diff
@@ -235,6 +250,41 @@ class BucketDiff {
         return ans;
     }
 
+    async get_first_bucket_keys_del(first_bucket_contents_left, first_bucket_delete_markers){
+        const first_bucket_delete_keys = [];
+        const field = 'IsLatest';
+        // Checking if there are any keys with latest delete marker in first bucket
+        Object.keys(first_bucket_contents_left).forEach(key => {
```
liranmauda, line 257: In the case that you already holds the delete marker, all we need to do is check that it is latest, if it is then none of the versions are

liranmauda (changes requested):

src/server/utils/bucket_diff.js
achouhan09: sure, will update it. Thanks

src/server/utils/bucket_diff.js
achouhan09: I need to check only for the latest delete marker and not for the version here, right?

src/server/utils/bucket_diff.js
achouhan09: Thanks, I will look into it.

src/server/utils/bucket_diff.js
```diff
@@ -69,38 +69,47 @@ class BucketDiff {
             //     ]
             // }
             keys_diff_map: {},
+            keys_del_map: {},
```
liranmauda, line 72: Need to add a comment on the stracture of `keys_del_map` same as the comment for `keys_diff_map` 
amd maybe change the name of `keys_del_map` to `keys_del_diff_map`

src/server/utils/bucket_diff.js
```diff
@@ -69,38 +69,47 @@ class BucketDiff {
...
             first_bucket_cont_token: '',
             second_bucket_cont_token: '',
         };
 
         let first_bucket_contents_left;
         let first_bucket_cont_token;
+        let first_bucket_delete_markers
```
liranmauda, line 79: Missing semicolon

src/server/utils/bucket_diff.js
```diff
@@ -69,38 +69,47 @@ class BucketDiff {
...
                 bucket_contents_left: second_bucket_contents_left,
-                bucket_cont_token: new_second_bucket_cont_token
+                bucket_cont_token: new_second_bucket_cont_token,
+                delete_markers: second_bucket_delete_markers
             } = await this.get_objects(this.second_bucket, prefix, max_keys, new_second_bucket_cont_token));
 
             const keys_diff_response = await this.get_keys_diff(
                 first_bucket_contents_left, second_bucket_contents_left, new_second_bucket_cont_token);
 
+            const delete_keys = await this.get_keys_del(
```
liranmauda, line 106: Why do we need to call it directly from here and not from the `get_keys_diff` function?

src/server/utils/bucket_diff.js
```diff
@@ -193,13 +202,36 @@ class BucketDiff {
         return list_objects_response.NextContinuationToken;
     }
 
+    /**
+     * @param {import("aws-sdk/lib/request").PromiseResult<AWS.S3.ListObjectVersionsOutput, AWS.AWSError> |
+    *         import("aws-sdk/lib/request").PromiseResult<AWS.S3.ListObjectsV2Output, AWS.AWSError>} list
```
liranmauda, line 207:
```suggestion
    *         import("aws-sdk/lib/request").PromiseResult<AWS.S3.ListObjectsV2Output, AWS.AWSError>} list_objects_response
```

src/server/utils/bucket_diff.js
```diff
@@ -193,13 +202,36 @@ class BucketDiff {
...
+    *
+    * _get_delete_markers will return the delete markers grouped by key
+    *
+    */
+    _get_delete_markers(list_objects_response) {
```
liranmauda, line 212: We should do a similar behaviour like `_object_grouped_by_key_and_omitted`
We should group the delete markers per key and omit when needed.

src/server/utils/bucket_diff.js
```diff
@@ -235,6 +268,25 @@ class BucketDiff {
         return ans;
     }
 
+    async get_keys_del(first_bucket_contents_left, first_bucket_delete_markers, second_bucket_delete_markers) {
```
liranmauda, line 271: Please add a javadoc for the typescript

src/server/utils/bucket_diff.js
```diff
@@ -235,6 +268,25 @@ class BucketDiff {
...
+            const first_bucket_delete_marker = first_bucket_delete_markers[key][field];
+            const second_bucket_delete_marker = second_bucket_delete_markers[key] && second_bucket_delete_markers[key][field];
+            // If the delete marker for a key in the first bucket is latest but is either not present or not latest in the second bucket
+            if (first_bucket_delete_marker && !second_bucket_delete_marker) {
+                delete_keys[key] = first_bucket_contents_left[key];
+            }
+        });
+
+        return delete_keys;
+    }
```
liranmauda, line 288: I think this logic belongs to the `get_keys_diff` function.
the `get_keys_diff` takes into account the possibility that the list of the first and the secound bucket are not alligned.

liranmauda (changes requested):

tangledbytes:
@achouhan09 this partially deals with https://github.com/noobaa/noobaa-core/issues/7252 right?

achouhan09:
@tangledbytes  Yes, it is partially related to it.

src/server/utils/bucket_diff.js
achouhan09: Done, Thanks.

src/server/utils/bucket_diff.js
achouhan09: Sure, will update it.

src/server/utils/bucket_diff.js
achouhan09: Resolved.

src/server/utils/bucket_diff.js
achouhan09: Done, Thanks.

src/server/utils/bucket_diff.js
achouhan09: done, thanks

src/server/utils/bucket_diff.js
```diff
@@ -68,47 +68,67 @@ class BucketDiff {
...
             diff.keys_diff_map = { ...diff.keys_diff_map, ...keys_diff_response.keys_diff_map };
+            diff.keys_del_diff_map = {...diff.keys_del_diff_map, ...keys_diff_response.keys_del_diff_map};
 
             const first_bucket_key_array = Object.keys(first_bucket_contents_left);
+            const first_bucket_del_array = Object.keys(first_bucket_delete_markers);
             const second_bucket_key_array = Object.keys(second_bucket_contents_left);
 
-            if (first_bucket_key_array.length !== 0 && second_bucket_key_array.length !== 0) {
+            if (first_bucket_key_array.length !== 0 && second_bucket_key_array.length !== 0 ||
+                first_bucket_del_array.length !== 0) {
```
liranmauda, line 127: Do we need to also add `second_bucket_delete_markers.length !== 0` for the case that we only have delete markers?

src/server/utils/bucket_diff.js
```diff
@@ -68,47 +68,67 @@ class BucketDiff {
...
                 const first_bucket_key_in_last_pos = first_bucket_key_array[first_bucket_key_array.length - 1];
                 const second_bucket_key_in_last_pos = second_bucket_key_array[second_bucket_key_array.length - 1];
-                if (first_bucket_key_in_last_pos >= second_bucket_key_in_last_pos) {
+                if (first_bucket_key_in_last_pos >= second_bucket_key_in_last_pos ||
+                    first_bucket_del_array.length !== 0) {
```
liranmauda, line 131: Do we need to also add second_bucket_delete_markers.length !== 0 for the case that we only have delete markers?

src/server/utils/bucket_diff.js
```diff
@@ -151,33 +171,42 @@ class BucketDiff {
...
-     * the results of this will be {{ [key: string]: Array<object> }} where the Array consist of metadata objects.
+     * the results of this will be two objects of type {{ [key: string]: Array<object> }} where the Array consist of metadata objects.
      *       for example: {
      *            "1": [
      *                { ETag: 'etag1.1', Size: 24599, Key: '1', VersionId: 'v1.1', IsLatest: true, },
      *                { ETag: 'etag1.2', Size: 89317, Key: '1', VersionId: 'v1.2', IsLatest: false, }
      *            ]
      *        }
-     * @returns {nb.BucketDiffKeysDiff}
+     * @returns {Object}
```
liranmauda, line 185: Maybe add the type of the items in the object (nb.BucketDiffKeysDiff)

src/server/utils/bucket_diff.js
```diff
@@ -151,33 +171,42 @@ class BucketDiff {
...
-            if (Object.keys(grouped_by_key).length > 1) {
-                grouped_by_key = _.omit(grouped_by_key, list[field][last_key_pos].Key);
+            let last_key_pos = list[field].length - 1;
+            if (Object.keys(grouped_objects).length > 1) {
+                grouped_objects = _.omit(grouped_objects, list[field][last_key_pos].Key);
+            }
+
+            last_key_pos = list[delete_marker].length - 1;
+            if (Object.keys(grouped_delete_markers).length > 1) {
+                grouped_delete_markers = _.omit(grouped_delete_markers, list[delete_marker][last_key_pos].Key);
```
liranmauda, line 206: I think we should only trancate the last delete marker key if it is equal to the last key in the version field.

achouhan09:
Closing this PR, new PR is created in replacement #7867

Issue #7683, opened 2024-01-01
NSFS | NC | CLI | Account update script change access_key without new_access_key argument in config file.


naveenpaul1:
### Environment info
- NooBaa Version: **master**

### Actual behavior
1. Account is updating with the wrong access_key. access_key in the config file got updated with the new access-key but the symlink file name remains the old name and it making the account corrupted.

### Expected behavior
1. access_key in config file change only when the `--new_access_key` added to command.

### Steps to reproduce
1.  ``` sudo node src/cmd/manage_nsfs account add --config_root /standalon/noobaa --name account1 --email z1@noobaa.com --new_buckets_path ../../standalon/nsfs_root/ --access_key Dwertyuiopasdfg12345 --secret_key Qwertyuiopasdfg12345Qwertyuiopasdfg12345 --uid 1000 --gid 1000```
2. ``` sudo node src/cmd/manage_nsfs account update --config_root /standalon/noobaa --name account1 --email z111ddddd@noobaa.com --access_key Hwertyuiopasdfg1234 ```

liranmauda:
@naveenpaul1
In this issue --access_key can be without --secret_key which does not make sense. 
#7681 is activating the validation for it.
The `--new_access_key` is not always needed, as autogenerate will not use it, and cannot use it.

liranmauda:
`--regenerate` flag is updating and `--access_key` is not.
as we remove `--access_key` as Identifier, we can remove the `--new_access_key` completely.

naveenpaul1:
@liranmauda AFIK we need both method right?  add access key using `--access_key` arg and --generate access key arg. ?

liranmauda:
@naveenpaul1 
update `--access_key` is required and indeed it has a bug that is not updating the symlink unless `--new_access_key` is there.
The way it worked before the regenerate addition was that you had to add `--new_access_key` to update that key, as access_key was the identifier for the account.
as access_key is no longer the identifier, using this flag should update the key and the symlink, and it makes the `--new_access_key` redundant or unneeded

romayalon:
@liranmauda As discussed at HPO meetings, --access_key and ---secret_key on add account /new_access_key on update needed for a specific use case of HPO and we should not remove them.

liranmauda:
Re-opening for validation

PravinRanjan10:
LGTM

romayalon:
closing per @PravinRanjan10 last comment

PR #7690, opened 2024-01-03
NC | NSFS |Account update script change access_key

naveenpaul1:
### Explain the changes
1. Account is updating with the wrong access_key. access_key in the config file got updated with the new access-key but the symlink file name remains the old name and it making the account corrupted.

1. https://github.com/noobaa/noobaa-core/issues/7683

### Testing Instructions:
1.  ``` sudo node src/cmd/manage_nsfs account add --config_root /standalon/noobaa --name account1 --email z1@noobaa.com --new_buckets_path ../../standalon/nsfs_root/ --access_key Dwertyuiopasdfg12345 --secret_key Qwertyuiopasdfg12345Qwertyuiopasdfg12345 --uid 1000 --gid 1000```
2. ``` sudo node src/cmd/manage_nsfs account update --config_root /standalon/noobaa --name account1 --email z111ddddd@noobaa.com --access_key Hwertyuiopasdfg1234 ```

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -458,6 +458,10 @@ async function fetch_existing_account_data(config_root, target) {
             throw_cli_error(ManageCLIError.NoSuchAccountName, target.name);
         }
     }
+    if (source.access_keys[0].access_key && target.access_keys[0].access_key &&
+            source.access_keys[0].access_key !== target.access_keys[0].access_key) {
+        throw_cli_error(ManageCLIError.AccountAccessKeyAndNameMismatch, target.access_keys[0].access_key);
+    }
     const data = _.merge({}, source, target);
     return data;
 }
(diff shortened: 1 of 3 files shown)


liranmauda:
@naveenpaul1 
The `--access_key` should not be an identifier for action `update`
as a result, using the `--access_key` flag on action `update` should update the account's access key.

Throwing an error when the name and the access key of the account are not matched is treating the access key as an Identifier.

IMHO, we should use `--access_key` for updates, and we should just use the `new_access_key` variable for updating it.
and we should completely revoke the `--new_access_key` flag.

naveenpaul1:
@liranmauda I agree that `access_key` should not identifier for the update in our CLI its not an identifier, We can update accounts even without `access_key`. This fix is to make sure when a user accidentally passes the wrong access_key and it corrupts an account. I don't think it's a very common case.

naveenpaul1:
closing PR, it is already taken care of in PR: https://github.com/noobaa/noobaa-core/pull/7697

PR #7702, opened 2024-01-07
NSFS | NC | Add Schema Validation to Bucket and Account Add and Update (NC NSFS CLI)

### LINKED ISSUE #7694: NSFS | NC | AJV bucket and account schemas should be validated on creation
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
1. bucket and account schemas currently validated on read account/bucket sdk info.

### Expected behavior
1. bucket and account schemas should be validated on add and update as well. bucket creation/update through CLI & bucketspace_fs. account create/update through CLI.

### Steps to reproduce
1.

shirady:
### Explain the changes
1. Move the NC NSFS schema validation that we had in file `src/sdk/bucketspace_fs.js` (from PR #7544) to a separate file (`src/manage_nsfs/nsfs_schema_utils.js`) to reuse it.
2. Edit the values of `s3_policy` and `fs_backend` to be undefined (and not empty `''`, as it passed as an argument).
3. Validate the data using schema and return an error in case it doesn't match (details would be passed according to the original message), using a mapping `ManageCLIError.RPC_ERROR_TO_MANAGE`.
4. Update the unit tests for the cases of `s3_policy` and `fs_backend` deletions (separate the argument that we pass as `''` to the actual property that would be `undefined`).
5. Change the `uid` and `gid` extract in function `fetch_account_data` to be `undefined` when the argument user was passed (and not `false`).
6. Use `schema_utils.strictify` with option  `additionalProperties: false` in the `nsfs_account_schema` and `nsfs_bucket_schema`.
7. Add the `versioning` enum to the bucket schema.
8. Add tests for the schema validation of nsfs_account_schema and nsfs_bucket_schema.

### Issues: Fixed #7694  
1. Add schema validation to bucket and account add and update - using NC NSFS CLI (fs_backend changes in a separate PR).
2. GAP - I couldn't ref bucket_api schema and had to duplicate the values, else got the error: `Error: reference "bucket_api" resolves to more than one schema` in the CI in job `run-unit-tests`.

### Testing Instructions:
1. For tests that were edited, please run:
`sudo node ./node_modules/.bin/_mocha src/test/unit_tests/test_nc_nsfs_cli.js`
`sudo node ./node_modules/.bin/_mocha src/test/unit_tests/test_bucketspace_fs.js`
3. For new tests, please run: 
`npx jest test_nc_nsfs_schema_validation.test.js`
(no need for sudo permission, can see results live also with jest extension in VSCode).

- [ ] Doc added/updated
- [X] Tests added
- [X] Tests edited (see comment above)

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -14,6 +14,7 @@ const SensitiveString = require('../util/sensitive_string');
 const ManageCLIError = require('../manage_nsfs/manage_nsfs_cli_errors').ManageCLIError;
 const ManageCLIResponse = require('../manage_nsfs/manage_nsfs_cli_responses').ManageCLIResponse;
 const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
+const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
 
 const TYPES = {
     ACCOUNT: 'account',
@@ -199,6 +200,7 @@ async function main(argv = minimist(process.argv.slice(2))) {
         dbg.log1('NSFS Manage command: exit on error', err.stack || err);
         const manage_err = ((err instanceof ManageCLIError) && err) ||
             new ManageCLIError(ManageCLIError.FS_ERRORS_TO_MANAGE[err.code] ||
+                ManageCLIError.RPC_ERROR_TO_MANAGE[err.rpc_code] ||
                 ManageCLIError.InternalError);
         throw_cli_error(manage_err, err.stack || err);
     }
@@ -254,7 +256,10 @@ async function fetch_bucket_data(argv, from_file) {
         bucket_owner: new SensitiveString(String(data.bucket_owner)),
         // update bucket identifier
         new_name: data.new_name && new SensitiveString(String(data.new_name)),
-        fs_backend: data.fs_backend || undefined
+        // fs_backend deletion specified with empty string '' (but it is not part of the schema)
+        fs_backend: data.fs_backend || undefined,
+        // s3_policy deletion specified with empty string '' (but it is not part of the schema)
+        s3_policy: data.s3_policy || undefined,
     };
 
     return data;
@@ -288,6 +293,10 @@ async function add_bucket(data) {
     if (exists) throw_cli_error(ManageCLIError.BucketAlreadyExists, data.name.unwrap());
 
     const data_json = JSON.stringify(data);
+    // We take an object that was stringify
+    // (it unwraps ths sensitive strings, creation_date to string and removes undefined parameters)
+    // for validating against the schema we need an object, hence we parse it back to object
+    nsfs_schema_utils.validate_bucket_schema(JSON.parse(data_json));
     await native_fs_utils.create_config_file(fs_context, buckets_dir_path, bucket_conf_path, data_json);
     write_stdout_response(ManageCLIResponse.BucketCreated, data_json);
 }
@@ -316,6 +325,10 @@ async function update_bucket(data) {
     if (!update_name) {
         const bucket_config_path = get_config_file_path(buckets_dir_path, data.name);
         data = JSON.stringify(data);
+        // We take an object that was stringify
+        // (it unwraps ths sensitive strings, creation_date to string and removes undefined parameters)
+        // for validating against the schema we need an object, hence we parse it back to object
+        nsfs_schema_utils.validate_bucket_schema(JSON.parse(data));
         await native_fs_utils.update_config_file(fs_context, buckets_dir_path, bucket_config_path, data);
         write_stdout_response(ManageCLIResponse.BucketUpdated, data);
         return;
@@ -330,7 +343,10 @@ async function update_bucket(data) {
     if (exists) throw_cli_error(ManageCLIError.BucketAlreadyExists, data.name.unwrap());
 
     data = JSON.stringify(_.omit(data, ['new_name']));
-
+    // We take an object that was stringify
+    // (it unwraps ths sensitive strings, creation_date to string and removes undefined parameters)
+    // for validating against the schema we need an object, hence we parse it back to object
+    nsfs_schema_utils.validate_bucket_schema(JSON.parse(data));
     await native_fs_utils.create_config_file(fs_context, buckets_dir_path, new_bucket_config_path, data);
     await native_fs_utils.delete_config_file(fs_context, buckets_dir_path, cur_bucket_config_path);
     write_stdout_response(ManageCLIResponse.BucketUpdated, data);
@@ -430,8 +446,8 @@ async function fetch_account_data(argv, from_file) {
             access_keys,
             nsfs_account_config: {
                 distinguished_name: argv.user,
-                uid: !argv.user && argv.uid,
-                gid: !argv.user && argv.gid,
+                uid: argv.user ? undefined : argv.uid,
+                gid: argv.user ? undefined : argv.gid,
                 new_buckets_path: argv.new_buckets_path,
                 fs_backend: argv.fs_backend ? String(argv.fs_backend) : config.NSFS_NC_STORAGE_BACKEND
             }
@@ -456,6 +472,7 @@ async function fetch_account_data(argv, from_file) {
             uid: data.nsfs_account_config.uid && Number(data.nsfs_account_config.uid),
             gid: data.nsfs_account_config.gid && Number(data.nsfs_account_config.gid),
             new_buckets_path: data.nsfs_account_config.new_buckets_path,
+            // fs_backend deletion specified with empty string '' (but it is not part of the schema)
             fs_backend: data.nsfs_account_config.fs_backend || undefined
         },
         allow_bucket_creation: !is_undefined(data.nsfs_account_config.new_buckets_path),
@@ -503,6 +520,10 @@ async function add_account(data) {
     }
 
     data = JSON.stringify(data);
+    // We take an object that was stringify
+    // (it unwraps ths sensitive strings, creation_date to string and removes undefined parameters)
+    // for validating against the schema we need an object, hence we parse it back to object
+    nsfs_schema_utils.validate_account_schema(JSON.parse(data));
     await native_fs_utils.create_config_file(fs_context, accounts_dir_path, account_config_path, data);
     await native_fs_utils._create_path(access_keys_dir_path, fs_context, config.BASE_MODE_CONFIG_DIR);
     await nb_native().fs.symlink(fs_context, account_config_path, account_config_access_key_path);
@@ -522,6 +543,10 @@ async function update_account(data) {
     if (!update_name && !update_access_key) {
         const account_config_path = get_config_file_path(accounts_dir_path, data.name);
         data = JSON.stringify(data);
+        // We take an object that was stringify
+        // (it unwraps ths sensitive strings, creation_date to string and removes undefined parameters)
+        // for validating against the schema we need an object, hence we parse it back to object
+        nsfs_schema_utils.validate_account_schema(JSON.parse(data));
         await native_fs_utils.update_config_file(fs_context, accounts_dir_path, account_config_path, data);
         write_stdout_response(ManageCLIResponse.AccountUpdated, data);
         return;
@@ -540,6 +565,10 @@ async function update_account(data) {
         throw_cli_error(err_code);
     }
     data = JSON.stringify(_.omit(data, ['new_name', 'new_access_key']));
+    // We take an object that was stringify
+    // (it unwraps ths sensitive strings, creation_date to string and removes undefined parameters)
+    // for validating against the schema we need an object, hence we parse it back to object
+    nsfs_schema_utils.validate_account_schema(JSON.parse(data));
     if (update_name) {
         await native_fs_utils.create_config_file(fs_context, accounts_dir_path, new_account_config_path, data);
         await native_fs_utils.delete_config_file(fs_context, accounts_dir_path, cur_account_config_path);
--- a/src/server/object_services/schemas/nsfs_bucket_schema.js
+++ b/src/server/object_services/schemas/nsfs_bucket_schema.js
@@ -28,6 +28,9 @@ module.exports = {
         },
         versioning: {
             type: 'string',
+            enum: ['DISABLED', 'SUSPENDED', 'ENABLED']
+            // GAP would like to use $ref: 'bucket_api#/definitions/versioning'
+            // but currently it creates an error Error: reference "bucket_api" resolves to more than one schema
         },
         path: {
             type: 'string',
--- a/src/manage_nsfs/manage_nsfs_cli_errors.js
+++ b/src/manage_nsfs/manage_nsfs_cli_errors.js
@@ -72,6 +72,11 @@ ManageCLIError.MissingConfigDirPath = Object.freeze({
     message: 'Config dir path should not be empty',
     http_code: 400,
 });
+ManageCLIError.InvalidSchema = Object.freeze({
+    code: 'InvalidSchema',
+    message: 'Schema invalid, please use required properties',
+    http_code: 400,
+});
 
 //////////////////////////////
 //// IP WHITE LIST ERRORS ////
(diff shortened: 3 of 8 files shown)


shirady:
### Manual Testing Instructions:
## Schema validation for buckets and accounts

First, create the `FS_ROOT` and give permissions:

```bash
mkdir -p /tmp/nsfs_root1
chmod 777 /tmp/nsfs_root1
```

- `new_buckets_path` flag that we will use in the account commands
- `path` in the buckets commands.

In each case the commands work, hence we added the code change (to sabotage the created object and make sure that it creates an error, in each case we chose one of the properties that is required to be deleted.

### Account add
1. Add account `sudo node src/cmd/manage_nsfs account add --name <account-name> --email <account-email> --new_buckets_path /tmp/nsfs_root1 --uid <uid-number> --gid <gid-number>`
2.  Add this line:
`delete data.email; // test`
before passing to the schema validation.
3. See the error: `RpcError: must have required property 'email'`

### Account update
1. Create an account (the account that we would update).
2. Update account `sudo node src/cmd/manage_nsfs account update --name <account-name> --email <change-the-account-email>`
2.  Add this line (in the first condition - it's neither a name update nor access keys update):
`delete data_without_properties_to_update.email; // test`
before passing to the schema validation.
4. See the error: `RpcError: must have required property 'email'`

### Bucket add
1. Create an account (the account that we would create the bucket for it).
2. Add bucket `sudo node src/cmd/manage_nsfs bucket add --name <bucket-name> --email <account-email> --path /tmp/nsfs_root1`
3.  Add this line:
`delete data.creation_date; // test`
before passing to the schema validation.
4. See the error: `RpcError: must have required property 'creation_date'`

### Bucket update
1. Create a bucket (the bucket that we would update).
3. Update bucket `sudo node src/cmd/manage_nsfs account update --name <account-name> --email <change-the-account-email>`
2.  Add this line (in the first condition - it's not a name update):
`delete data_without_properties_to_update.creation_date; // test`
before passing to the schema validation.
5. See the error: `RpcError: must have required property 'creation_date'`

## Remove bucket policy and fs_backend:

First, Create bucket `sudo node src/cmd/manage_nsfs bucket add --name <bucket-name> --email <account-email> --path /tmp/nsfs_root1` In the next commands please use the same `<bucket-name>` as you created here.

### Bucket policy:
1. Add bucket policy: `sudo node src/cmd/manage_nsfs bucket update --name <bucket-name> --bucket_policy '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":["*"]},"Action":["s3:*"],"Resource":["arn:aws:s3:::<bucket-name>/*","arn:aws:s3:::<bucket-name>"]}]}'` (replace the `<bucket-name>` in the bucket policy json argument as well)
2. Remove bucket policy: `sudo node src/cmd/manage_nsfs bucket update --name <same-name-above> --bucket_policy ''`

### FS backend
1. Add FS backend: `sudo node src/cmd/manage_nsfs bucket update --name <bucket-name> --fs_backend GPFS`
2. Remove FS backend: `sudo node src/cmd/manage_nsfs bucket update --name <bucket-name> --fs_backend ''`

src/api/common_api.js
```diff
@@ -1269,7 +1269,8 @@ module.exports = {
 
         fs_backend: {
             type: 'string',
-            enum: ['CEPH_FS', 'GPFS', 'NFSv4']
+            enum: ['CEPH_FS', 'GPFS', 'NFSv4', '']
```
romayalon, line 1272: this is a hacky solution, please avoid using it if possible as I did in https://github.com/noobaa/noobaa-core/pull/7698

src/server/utils/nsfs_schema_utils.js
```diff
@@ -0,0 +1,34 @@
...
+'use strict';
+
+const RpcError = require('../../rpc/rpc_error');
+const { default: Ajv } = require('ajv');
+const ajv = new Ajv({ verbose: true, allErrors: true });
+const { KEYWORDS } = require('../../util/schema_keywords');
+const common_api = require('../../api/common_api');
+
+ajv.addKeyword(KEYWORDS.methods);
+ajv.addKeyword(KEYWORDS.doc);
```
romayalon, line 11: I think that a better location for this file can be under manage_nsfs folder

src/server/object_services/schemas/nsfs_bucket_schema.js
```diff
@@ -42,7 +42,13 @@ module.exports = {
             $ref: 'common_api#/definitions/fs_backend'
         },
         s3_policy: {
-            $ref: 'common_api#/definitions/bucket_policy',
+            oneOf: [{
+                $ref: 'common_api#/definitions/bucket_policy',
+            }, {
+                type: 'string',
+                enum: ['']
```
romayalon, line 49: here again please don't use this hack and delete the bucket policy as I did in the PR attached in the previous comment

src/cmd/manage_nsfs.js
```diff
@@ -754,6 +765,50 @@ async function validate_account_args(data, action) {
...
+            }],
+            nsfs_account_config: {
+                ...data.nsfs_account_config,
+                distinguished_name: data.nsfs_account_config?.distinguished_name?.unwrap(),
+            },
+        };
+        nsfs_schema_utils.validate_account_schema(account_to_validate);
+    } catch (err) {
+        dbg.error('extract_properties_from_account_data_and_validate_schema: schema validation error:', err);
+        throw err;
```
romayalon, line 790: this will throw internal error, please add a new CLI error specifying the details of the error

src/cmd/manage_nsfs.js
```diff
@@ -754,6 +765,50 @@ async function validate_account_args(data, action) {
     }
 }
 
+// we take the data object and extract its properties according to the schema:
+// - we unwrap ths sensitive string (that we have from fetch_account_data)
+// - we change the creation_date to string
+function extract_properties_from_account_data_and_validate_schema(data) {
+    try {
+        const account_to_validate = {
```
romayalon, line 773: this is redundant, JSON.stringify will do it for you

src/cmd/manage_nsfs.js
```diff
@@ -754,6 +765,50 @@ async function validate_account_args(data, action) {
...
+        throw err;
+    }
+}
+
+// we take the data object and extract its properties according to the schema:
+// - we unwrap ths sensitive string (that we have from fetch_bucket_data)
+// - we change the creation_date to string
+function extract_properties_from_bucket_data_and_validate_schema(data) {
+    try {
+        const bucket_to_validate = {
```
romayalon, line 799: this is redundant, JSON.stringify will do it for you

src/cmd/manage_nsfs.js
```diff
@@ -754,6 +765,50 @@ async function validate_account_args(data, action) {
...
+            ...data,
+            name: data.name?.unwrap(),
+            system_owner: data.system_owner?.unwrap(),
+            bucket_owner: data.bucket_owner?.unwrap(),
+            creation_date: data.creation_date?.toString(),
+        };
+        nsfs_schema_utils.validate_bucket_schema(bucket_to_validate);
+    } catch (err) {
+        dbg.error('extract_properties_from_bucket_data_and_validate_schema: schema validation error:', err);
+        throw err;
```
romayalon, line 809: same here, please do not throw internal error

romayalon:
@shirady please add automatic tests

src/api/common_api.js
shirady: Leaving this value as an empty string does not match the schema.
It's either changing the schema to the needs (using an empty string to describe removal of fs_backend) or setting the default POSIX and adding it to the enum, as mentioned in the error:
https://github.com/noobaa/noobaa-core/blob/d18db4308db95df162ebe10b442cf110aae0f2fd/src/manage_nsfs/manage_nsfs_cli_errors.js#L281

src/cmd/manage_nsfs.js
shirady: Thanks, would change it.

src/cmd/manage_nsfs.js
shirady: Will add, thanks.

src/cmd/manage_nsfs.js
shirady: Thanks, would change it.

src/cmd/manage_nsfs.js
shirady: As suggested I will add a new CLI error, thanks.

src/server/object_services/schemas/nsfs_bucket_schema.js
shirady: I'm not sure what you're suggesting.
In case someone created a bucket with a bucket policy and tries to delete it - what do you expect the bucket policy to be? (with current code it is an empty string).

src/server/utils/nsfs_schema_utils.js
shirady: Do you mean where 'manage.nsfs.js' is now? to move this file to `src/cmd`?

src/api/common_api.js
romayalon: fs_backend is not required, please make it undefined, there is no need to add anything to the schema

src/server/utils/nsfs_schema_utils.js
romayalon: no, please put this file under manage_nsfs folder

src/server/object_services/schemas/nsfs_bucket_schema.js
romayalon: undefined

src/cmd/manage_nsfs.js
```diff
@@ -289,6 +290,7 @@ async function add_bucket(data) {
     const exists = await native_fs_utils.config_file_exists(fs_context, bucket_conf_path);
     if (exists) throw_cli_error(ManageCLIError.BucketAlreadyExists, data.name.unwrap());
 
+    extract_properties_from_bucket_data_and_validate_schema(data);
```
romayalon, line 293: @shirady 
A few days ago I noticed that when creating an account while specifying a distinguished name instead of uid and gid, we accidentally assign false to uid and gid instead of a number and this would break the schema, can you please check that?

shirady: @romayalon I checked it.
- I will suggest a small fix in this PR (separate commit for code review).
- It doesn't break the schema (if it was I had to fix it before...).

To reproduce it, we can add this print:
`console.log('SDSD res', res);`
in the test (it) `cli account create `under (describe) `cli account flow distinguished_name - happy path`.
Before fix:

> [LOG] CONSOLE:: SDSD res {
>   "response": {
>     "code": "AccountCreated",
>     "reply": {
> ...
>       "nsfs_account_config": {
>         "distinguished_name": "moti1003",
>         "uid": false,
>         "gid": false,
>         "new_buckets_path": "/private/tmp/test_bucketspace_fs/root_path_manage_nsfs/new_buckets_path_user2/"
> ...

After fix:

> [LOG] CONSOLE:: SDSD res {
>   "response": {
>     "code": "AccountCreated",
>     "reply": {
> ...
>       "nsfs_account_config": {
>         "distinguished_name": "moti1003",
>         "new_buckets_path": "/private/tmp/test_bucketspace_fs/root_path_manage_nsfs/new_buckets_path_user2/"
> ...

romayalon: setting a property to false (type boolean) instead of a number must break the schema.
I wonder if you didn't see it because of the oneOf keyword added to the schema.
please add additionalProperties: false to the schema (both account and bucket), I suppose that after adding it you should see the error.

shirady: @romayalon I added it both in the account and bucket schema as the last line and rerun the mentioned test above (using the original code before the change) and still didn't get an error from the schema.
After setting the `additionalProperties: false` to each list of the `oneOf` and re-running the test again I received 3 failing tests including the one mentioned above), therefore in nsfs_account_schema you'll see the `additionalProperties: false` appears 3 times.

romayalon: I checked what we are doing in our existing schema code and it looks like we use schema_utils.strictify(schema, options, base). This should set additionalProperties: false recursively and do some more stuff that we might want to do

shirady: I added it (`schema_utils.strictify`), thanks.

src/cmd/manage_nsfs.js
```diff
@@ -257,6 +258,15 @@ async function fetch_bucket_data(argv, from_file) {
         fs_backend: data.fs_backend || undefined
     };
 
+    // fs_backend deletion specified with empty string '' (but it is not part of the schema)
+    if (argv.fs_backend === '') {
```
romayalon, line 262: line 258 will do the same in the object assignment, so this is redundant, you can still keep the comment if you wish
I suggest also to do the same for s3_policy if possible

src/cmd/manage_nsfs.js
```diff
@@ -462,6 +475,12 @@ async function fetch_account_data(argv, from_file) {
         new_name: data.new_name && new SensitiveString(String(data.new_name)),
         new_access_key: data.new_access_key && new SensitiveString(String(data.new_access_key))
     };
+
+    // fs_backend deletion specified with empty string '' (but it is not part of the schema)
+    if (argv.fs_backend === '' && data.nsfs_account_config?.fs_backend !== undefined) {
+        data.nsfs_account_config.fs_backend = undefined;
```
romayalon, line 481: same here, line 471 handles this already

src/cmd/manage_nsfs.js
```diff
@@ -758,6 +780,35 @@ async function validate_account_args(data, action) {
     }
 }
 
+/**
+ * We take an object that was stringify
+ * (it unwraps ths sensitive strings, creation_date to string and removes undefined parameters)
+ * for validating against the schema we need an object, hence we parse it back to object
+ * @param {string} data
+ */
+function validate_account_schema_wrapper(data) {
```
romayalon, line 789: I suggest to avoid adding a new function only for calling a different function and throwing a different error, 
if you would like to throw ManageCLIError.InvalidSchema instead of InternalError (INVALID_SCHEMA) you can add a mapping between the two in a new RPC_ERROR_TO_MANAGE object similar to FS_ERRORS_TO_MANAGE
and in the main on manage_nsfs apply the mapping based on the err.rpc_code

src/cmd/manage_nsfs.js
```diff
@@ -758,6 +780,35 @@ async function validate_account_args(data, action) {
...
+
+/**
+ * We take an object that was stringify
+ * (it unwraps ths sensitive strings, creation_date to string and removes undefined parameters)
+ * for validating against the schema we need an object, hence we parse it back to object
+ * @param {string} data
+ */
+function validate_bucket_schema_wrapper(data) {
+    try {
+        nsfs_schema_utils.validate_bucket_schema(JSON.parse(data));
```
romayalon, line 806: try to avoid stringify and then parsing again, you can use the initial value

src/cmd/manage_nsfs.js
shirady: I will remove this and add the comment for `s3_policy`, thanks.

src/cmd/manage_nsfs.js
shirady: I will remove it, thanks.

src/cmd/manage_nsfs.js
shirady: I will remove the wrapper and add the mapping as suggested, thanks.

src/cmd/manage_nsfs.js
shirady: Don't forget what I earned from the `JSON.stringify` that is not in the initial value - it removes `undefined` values and `unwarp` `SensitiveString` properties (either using it or doing it my own as I did at first). I added a comment to explain this.

romayalon (approved):
Approving, as mentioned on https://github.com/noobaa/noobaa-core/issues/7694, bucketspace_fs missing validations will be added in the next PR

naveenpaul1 (approved):

Issue #7699, opened 2024-01-05
Error messages are not proper for account status command


anandhu-karattu:
### Environment info
- NooBaa Version: noobaa-core-5.15.0-20240102.el9.x86_64
- Platform: Stand alone RHEL 9.2

### Actual behavior
When I do the account status command, it is giving invalid output. 2 scenarios 
Issue 1:- Do account status without any account name. It is showing "Account already exists - access_key" in output.
Output must clearly say that "No account name provides" or something similar

```
[root@anrhel92-x-worker1 ~]# /usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/manage_nsfs account status 2>/dev/null
{
  "error": {
    "code": "AccountAccessKeyAlreadyExists",
    "message": "Account already exists - access_key",
    "detail": "xwoCw6Z7LsOInb2yVBU3"
  }
}

```

Issue 2: If I specify any invalid account name, out put is saying "Account already exists - name".
It should clearly error out saying account does not exist.
In the below example "test4" is a non existent account
```
[root@anrhel92-x-worker1 ~]# /usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/manage_nsfs account status --name test4 2>/dev/null
{
  "error": {
    "code": "AccountNameAlreadyExists",
    "message": "Account already exists - name",
    "detail": "test4"
  }
}

```

### Expected behavior
All the error messages should be proper. Details are specified in above section

### Steps to reproduce
Details are provided in issue description

romayalon:
adding request validation due to the PR of @naveenpaul1.
@anandhu-karattu please close it when verified.

ramya-c3:
[root@r-12 noobaa-core]# /usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/manage_nsfs account status --name terst 2>/dev/null
{
  "error": {
    "code": "NoSuchAccountName",
    "message": "Account does not exist - name",
    "detail": "terst"
  }
}
[root@r-12 noobaa-core]# /usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/manage_nsfs account status --name test 2>/dev/null
{
  "response": {
    "code": "AccountStatus",
    "reply": {
      "name": "test",
      "email": "test",
      "creation_date": "2024-02-12T05:13:52.106Z",
      "nsfs_account_config": {
        "uid": 23,
        "gid": 23,
[...]

PR #7703, opened 2024-01-08
NSFS | NC | CLI |bucket delete leaving objects

naveenpaul1:
### Explain the changes
1. S3 bucket delete is possible only for empty buckets.
2. Falied unit test cases after changing the `assert_error` condition to `equal`
3. https://github.com/noobaa/noobaa-core/issues/7699

1. https://github.com/noobaa/noobaa-core/issues/7699

### Testing Instructions:
1. create a bucket and copy one or more items to it and try to delete.
2. test steps mentioned issue [7699](https://github.com/noobaa/noobaa-core/issues/7699)

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -411,7 +411,8 @@ async function fetch_account_data(argv, from_file) {
     }
     if (action !== ACTIONS.LIST && action !== ACTIONS.STATUS) _validate_access_keys(argv);
     if (action === ACTIONS.ADD || action === ACTIONS.STATUS) {
-        access_keys = set_access_keys(argv, true);
+        const regenerate = action === ACTIONS.ADD
+        access_keys = set_access_keys(argv, regenerate);
     } else if (action === ACTIONS.UPDATE) {
         access_keys = set_access_keys(argv, Boolean(argv.regenerate));
         new_access_key = access_keys[0].access_key;
@@ -577,9 +578,9 @@ async function get_account_status(data, show_secrets) {
         write_stdout_response(ManageCLIResponse.AccountStatus, config_data);
     } catch (err) {
         if (is_undefined(data.name)) {
-            throw_cli_error(ManageCLIError.AccountAccessKeyAlreadyExists, data.access_keys[0].access_key.unwrap());
+            throw_cli_error(ManageCLIError.NoSuchAccountAccessKey, data.access_keys[0].access_key.unwrap());
         } else {
-            throw_cli_error(ManageCLIError.AccountNameAlreadyExists, data.name.unwrap());
+            throw_cli_error(ManageCLIError.NoSuchAccountName, data.name.unwrap());
         }
     }
 }
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -380,17 +380,21 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             try {
                 const namespace_bucket_config = await object_sdk.read_bucket_sdk_namespace_info(params.name);
                 dbg.log1('BucketSpaceFS.delete_bucket: namespace_bucket_config', namespace_bucket_config);
+                const ns = await object_sdk._get_bucket_namespace(params.name);
                 if (namespace_bucket_config && namespace_bucket_config.should_create_underlying_storage) {
-                    const ns = await object_sdk._get_bucket_namespace(params.name);
                     // delete underlying storage = the directory which represents the bucket
                     dbg.log1('BucketSpaceFS.delete_bucket: deleting uls', this.fs_root, namespace_bucket_config.write_resource.path);
                     await ns.delete_uls({
                         name,
                         full_path: path.join(this.fs_root, namespace_bucket_config.write_resource.path) // includes write_resource.path + bucket name (s3 flow)
                     }, object_sdk);
+                } else if (namespace_bucket_config){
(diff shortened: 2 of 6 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -410,7 +411,7 @@ async function fetch_account_data(argv, from_file) {
         data = JSON.parse(raw_data.toString());
     }
     if (action !== ACTIONS.LIST && action !== ACTIONS.STATUS) _validate_access_keys(argv);
-    if (action === ACTIONS.ADD || action === ACTIONS.STATUS) {
+    if (action === ACTIONS.ADD) {
```
liranmauda, line 414: removing `ACTIONS.STATUS` will lead to an error when trying to do status based on the access key.
We should keep this

src/cmd/manage_nsfs.js
```diff
@@ -663,6 +664,19 @@ async function get_config_data(config_file_path, show_secrets) {
 ///     VALIDATIONS     ///
 ///////////////////////////
 
+async function is_bucket_empty(fs_context, full_bucket_config_path) {
```
liranmauda, line 667: Please add the JSDoc comment for the TypeScript

src/test/unit_tests/test_nc_nsfs_cli.js
```diff
@@ -378,10 +396,20 @@ mocha.describe('manage_nsfs cli', function() {
             assert_account(account, account_options);
         });
 
+        mocha.it('cli account status without name and access_key', async function() {
```
liranmauda, line 399: Should be written using jest as decided a few weeks ago. adding to mocha is not something that we want
@nimrod-becker FYI

liranmauda (changes requested):

src/cmd/manage_nsfs.js
```diff
@@ -663,6 +664,19 @@ async function get_config_data(config_file_path, show_secrets) {
...
+    let data;
+    try {
+        data = await get_config_data(full_bucket_config_path);
+    } catch (err){
+        throw_cli_error(ManageCLIError.NoSuchBucket,  full_bucket_config_path);
+    }
+    const entries = await nb_native().fs.readdir(fs_context, data.path);
```
liranmauda, line 674: we can add a TODO here, once we can read only a single instance in the directory, and not iterate over the whole directory we should use it.

src/sdk/bucketspace_fs.js
```diff
@@ -388,6 +388,13 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
                         name,
                         full_path: path.join(this.fs_root, namespace_bucket_config.write_resource.path) // includes write_resource.path + bucket name (s3 flow)
                     }, object_sdk);
+                } else if (namespace_bucket_config){
+                    // TODO check only for single instance.
```
romayalon, line 392: what this TODO means?

src/test/unit_tests/test_bucketspace_fs.js
```diff
@@ -315,6 +316,21 @@ mocha.describe('bucketspace_fs', function() {
...
+        mocha.it('delete_bucket for non empty buckets', async function() {
+            const param = { name: test_not_empty_bucketbucket};
+            await create_bucket(param.name);
+            const bucket_file_path = path.join(new_buckets_path, param.name, 'dummy.txt');
+            await nb_native().fs.writeFile(ACCOUNT_FS_CONFIG, bucket_file_path,
+                Buffer.from(JSON.stringify("data")), {
+                    mode: config.BASE_MODE_FILE,
+                });
+            try {
+                await bucketspace_fs.delete_bucket(param, dummy_object_sdk);
```
romayalon, line 328: please add here assert.fail('should have failed with NOT EMPTY')

romayalon (commented):
looks good, final comments

src/sdk/bucketspace_fs.js
naveenpaul1: @romayalon bucket delete should not happen even if there is a single file in bucket, in that case we dont need to iterate through complete files, It was a suggestion from Liran. Will update the comment with more explaining words.

romayalon: that reminds me of something that we need to handle here, if the directory only contains the internal hidden directory we should allow the deletion
few of our internal dirs are - 
get_bucket_tmpdir(), versions directory

liranmauda: I asked for the TODO, when we will implement the object list optimization we can call readdir whiteout iterating over all the files. 
Then we should implement it also here, as this readdir is a performance hazard... 
If we have a lot of files/objects, then failing this delete bucket for the reason that the bucket is not empty will take a lot of time.

romayalon: currently on Namespace FS we do like follows - 
```
  const list = await this.list_objects({ ...params, limit: 1 }, object_sdk);

  if (list && list.objects && list.objects.length > 0) {
      throw new RpcError('NOT_EMPTY', 'underlying directory has files in it');
  }
```
I guess that what @liranmauda means, list_objects ignores also the hidden directories @naveenpaul1 you can do this here as well, just replace this with the correct namespace

liranmauda: @romayalon This is a great sulotion

naveenpaul1: @romayalon @liranmauda updated code with `ns.list_object()` to get the dir items

romayalon (approved):

liranmauda (approved):

PR #7713, opened 2024-01-11
NSFS | NC add nsfs_config_schema to nsfs_schema_utils and add missing properties

romayalon:
### Explain the changes
1. add missing properties to nsfs_config_schema - (NSFS_NC_STORAGE_BACKEND, NSFS_NC_CONFIG_DIR_BACKEND)
2. move nsfs_bucket_schema and nsfs_account_schema from object_services to system_services
3. add nsfs_config_schema to nsfs_utils and remove redundant code from config.json

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


shirady (approved):
LGTM - small comment that might change the default value.

shirady:
@romayalon if you want to add tests that would "protect" changes of schema you can do it in test file style this: `src/test/unit_tests/jest_tests/test_nc_nsfs_schema_validation.test.js`.
Now that I'm aware that there more schemas I would break it to account and bucket (and you can add a fie for configuration schema).

src/server/system_services/schemas/nsfs_config_schema.js
```diff
@@ -54,7 +54,17 @@ const nsfs_node_config_schema = {
         GPFS_DL_PATH: {
             type: 'string',
             default: '',
-            description: 'indicate the location of the gpfs library file, service restart required, usually should be set to /usr/lib64/libgpfs.so'
+            description: 'indicates the location of the gpfs library file, service restart required, usually should be set to /usr/lib64/libgpfs.so'
+        },
+        NSFS_NC_STORAGE_BACKEND: {
+            $ref: 'common_api#/definitions/fs_backend',
+            default: '',
```
shirady, line 61: This default does not match the schema (it is a string that is not part of the enum), might be confusing.

src/server/system_services/schemas/nsfs_config_schema.js
```diff
@@ -54,7 +54,17 @@ const nsfs_node_config_schema = {
...
+            description: 'indicates the global storage backend type, service restart required'
+        },
+        NSFS_NC_CONFIG_DIR_BACKEND: {
+            $ref: 'common_api#/definitions/fs_backend',
+            default: '',
```
shirady, line 66: same comment above (This default does not match the schema (it is a string that is not part of the enum), might be confusing.).

src/server/system_services/schemas/nsfs_config_schema.js
romayalon: np I changed it to undefined

src/server/system_services/schemas/nsfs_config_schema.js
romayalon: same answer

src/test/unit_tests/jest_tests/test_nc_nsfs_schema_validation.test.js
```diff
@@ -1258,3 +1258,93 @@ describe('schema validation NC NSFS bucket', () => {
     });
 
 });
+
+describe('schema validation NC NSFS config schema', () => {
```
shirady, line 1262: @romayalon 
I will separate the files if you want to separate them for config schema in this PR (or I'll do it in my PR).

romayalon: I think that separating the tests by different "describe" sections is enough for now.
but I noticed that all tests codes are pretty repetitive( expect(err).toBeInstanceOf(RpcError); for example repeats 50 times), I think we can add helper functions instead that will make the test file a lot shorter, for example:
```
function create_account_object(),
function create_bucket_object(),
function assert_validation(schema_type, object_to_validate, fail_options = { message: '') })
```

shirady: I wrote it this way so it will be super easy to investigate in case a test fails.
Anyway, we can take this discussion on my open PR.

PR #7719, opened 2024-01-15
NSFS | NC | NC coretest - bug fixes - PR 1/3

romayalon:
### Explain the changes
1. manage_nsfs.js - throw NoSuchAccount only when err.code === ENOENT, and other errors on different error codes.
2. bucketspace_fs.js - 
    2.1. read_account_by_access_key() - distinguished_name should we wrapped as SensitiveString() & fixed the relevant 
           tests on test_bucketspace_fs.js.
    2.2. check_bucket_config() - nb_native().fs.stat() throws an error, wrapped with try and catch instead of the 
           (!bucket_dir_stat) condition.
    2.3. get_bucket_name() - replaced the nb_native().fs.readFile() to just return the suffix of the bucket config json file 
           (without the json).
    2.4. list_buckets() - moved try catch to wrap the readdir() only, removed the buckets.filter(bucket => 
           bucket.name.unwrap()) as it looks unnecessary.
3. test_nc_nsfs_cli.js - assert_account() - added check for uid,gid = undefined when distinguished_name defined and vice 
    versa.

### Testing Instructions:
1. Tests covering these cases are a part of the NC_CORETST effort - test_bucketspace.js,

- [ ] Doc added/updated
- [ ] Tests Edited

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -492,11 +492,14 @@ async function fetch_existing_account_data(target) {
         source = await get_config_data(account_path, true);
     } catch (err) {
         dbg.log1('NSFS Manage command: Could not find account', target, err);
-        if (is_undefined(target.name)) {
-            throw_cli_error(ManageCLIError.NoSuchAccountAccessKey, target.access_keys[0].access_key);
-        } else {
-            throw_cli_error(ManageCLIError.NoSuchAccountName, target.name);
+        if (err.code === 'ENOENT') {
+            if (is_undefined(target.name)) {
+                throw_cli_error(ManageCLIError.NoSuchAccountAccessKey, target.access_keys[0].access_key);
+            } else {
+                throw_cli_error(ManageCLIError.NoSuchAccountName, target.name);
+            }
         }
+        throw err;
     }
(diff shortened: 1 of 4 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -492,11 +492,14 @@ async function fetch_existing_account_data(target) {
...
-        } else {
-            throw_cli_error(ManageCLIError.NoSuchAccountName, target.name);
+        if (err.code === 'ENOENT') {
+            if (is_undefined(target.name)) {
+                throw_cli_error(ManageCLIError.NoSuchAccountAccessKey, target.access_keys[0].access_key);
+            } else {
+                throw_cli_error(ManageCLIError.NoSuchAccountName, target.name);
+            }
         }
+        throw err;
```
naveenpaul1, line 502: @romayalon so we will catch the rest of the error in main block, right?

naveenpaul1 (approved):

src/cmd/manage_nsfs.js
romayalon: yes, exactly.
without it, even if there was an error we will say that the operation succeeded.

PR #7720, opened 2024-01-15
NSFS | NC | NC Coretests changes - PR 2/3

romayalon:
WIP - This PR is dependent on merging https://github.com/noobaa/noobaa-core/pull/7719
### Explain the changes
1. Makefile - added run-nc-tests makefile rule for running nc_index.js tests file.
2. test_utils.js - added a helper function for get_coretest_path() that chooses coretest/nc_coretest path by process.env.NC_CORETEST.
3. nc_coretest.js - config_dir, storage, and other setups and starting nsfs.js process, added mapping between rpc_client calls and manage_nsfs commands or other mocks.
4. nc_index.js - added NSFS relevant test files.
5. test_bucketspace.js - added skips by process.env.NC_CORETEST of non-relevant tests, and different error_codes.

1. Gap - nc_index.js contains commented test files that needs to be added as well.
2. github action for running nc_index.js

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/test/unit_tests/nc_coretest.js
+++ b/src/test/unit_tests/nc_coretest.js
@@ -1,0 +1,450 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const fs = require('fs');
+const p = require('path');
+const _ = require('lodash');
+const mocha = require('mocha');
+const { exec_manage_cli, nc_nsfs_manage_entity_types, nc_nsfs_manage_actions } = require('../system_tests/test_utils');
+const child_process = require('child_process');
+const argv = require('minimist')(process.argv);
+const SensitiveString = require('../../util/sensitive_string');
+
+// keep me first - this is setting envs that should be set before other modules are required.
+const NC_CORETEST = 'nc_coretest';
+process.env.DEBUG_MODE = 'true';
+process.env.ACCOUNTS_CACHE_EXPIRY = '1';
+process.env.NC_CORETEST = 'true';
+
+require('../../util/dotenv').load();
+require('../../util/panic');
+require('../../util/fips');
+
+const config = require('../../../config.js');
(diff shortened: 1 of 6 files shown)


src/test/unit_tests/nc_coretest.js
```diff
@@ -0,0 +1,450 @@
...
+    const list = await exec_manage_cli(nc_nsfs_manage_entity_types.ACCOUNT, nc_nsfs_manage_actions.LIST, options);
+    return { accounts: JSON.parse(list).response.reply };
+}
+
+/**
+ * update_account_s3_access_manage updates an account using manage_nsfs
+ * @param {object} options
+ * @returns {Promise<void>}
+ */
+async function update_account_s3_access_manage(options) {
```
naveenpaul1, line 327: @romayalon are we validating this account update anywhere else? just a doubt

naveenpaul1 (commented):
A comman doubt : - How this different from the normal test?

src/test/unit_tests/nc_coretest.js
romayalon: on the tests, this function equal to the rpc function - update_account_s3_access(), the difference here is that on Non containerized version we don't have RPC server so we do the update by running manage NSFS cli calls instead and return rpc like response.

romayalon:
@naveenpaul1 The regular Coretest runs noobaa endpoint (and other components of noobaa core) and then a test can run full s3 flow.
NC Coretest runs nsfs.js instead of running the endpoint and by that, we run the NC code. Then a test can run full s3 flow on NC environment. I hope it makes it clearer, let me know if there are more questions.

naveenpaul1 (approved):
LGTM

PR #7725, opened 2024-01-16
Add configuration to deny upload to "STANDARD" storage class

tangledbytes:
### Explain the changes
This PR adds a simple configuration which denies uploads to `STANDARD` storage class. The default value is `false` which can be overridden by `config-local.js`.

### Testing Instructions:
1. Set `config.DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD = true` in `config-local.js`
2. Try uploading a file either without specifying the storage class or after specifying `STANDARD` storage class.
3. The upload should fail. Replace the storage class with `GLACIER` and the upload should succeed (given `config.NSFS_RESTORE_ENABLED = true`).

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -172,6 +172,8 @@ config.S3_CORS_EXPOSE_HEADERS = [
 ].join(',');
 config.STS_CORS_EXPOSE_HEADERS = 'ETag';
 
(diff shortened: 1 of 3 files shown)


guymguym (approved):
small name suggestion

PR #7728, opened 2024-01-16
NSFS | NC | Verify That Bucket Owner Exists

### LINKED ISSUE #7724: NSFS | NC | CLI | Bucket Create/Update Should Check That We Have the Account
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
1. Currently users can create a bucket without having the actual account. 
2. Currently users can update a bucket without having the actual account. 

### Expected behavior
1. Should throw an error (bucket create/update should check that we have the account).

### Steps to reproduce
Bucket Create
1. `sudo node src/cmd/manage_nsfs bucket add --name<bucket-name> --email <account-email> --path <path>`.
(the `<account-email>` should be an account that was not created yet).

Bucket Update
1. Create an account `sudo node src/cmd/manage_nsfs account add --name <account-name> --email <account-email> --new_buckets_path /tmp/nsfs_root1 --uid <uid-number> --gid <gid-number>`.
(the `<account-email>` should be an account that already exists).
2. Create a bucket `sudo node src/cmd/manage_nsfs bucket add --name<bucket-name> --email <account-email> --path <path>`.
(the `<account-email>` should be an account that already exists, from step 1).
3. Update the bucket `sudo node src/cmd/manage_nsfs bucket update --name<bucket-name> --email <account-email>`.
(the `<account-email>` should be an account that was not created yet).

none

shirady:
### Explain the changes
1. Verify that the bucket owner exists when using `bucket add` and `bucket update` options from the NC NSFS CLI.
2. Change the variable names in the test `cli account list` to match the test.

### Issues: Fixed #7724
1. Currently we can create and update a bucket with a bucket owner that doesn't exist.
2. GAP - should improve performance of the implementation, iterate with concurrency, and the ability to stop the iteration once we find that the bucket owner account exists (more details, [here](https://github.com/noobaa/noobaa-core/pull/7728#discussion_r1457403937)).

### Testing Instructions:
1. Manual tests are described in the issue mentioned above.
3. For running the unit test with the new tests, please run: `sudo node ./node_modules/.bin/_mocha src/test/unit_tests/test_nc_nsfs_cli.js`.

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -185,6 +185,7 @@ function get_symlink_config_file_path(config_type_path, file_name) {
 
 async function add_bucket(data) {
     await validate_bucket_args(data, ACTIONS.ADD);
+    await verify_bucket_owner(data.bucket_owner.unwrap());
     const fs_context = native_fs_utils.get_process_fs_context(config_root_backend);
     const bucket_conf_path = get_config_file_path(buckets_dir_path, data.name);
     const exists = await native_fs_utils.config_file_exists(fs_context, bucket_conf_path);
@@ -197,6 +198,33 @@ async function add_bucket(data) {
     nsfs_schema_utils.validate_bucket_schema(JSON.parse(data_json));
     await native_fs_utils.create_config_file(fs_context, buckets_dir_path, bucket_conf_path, data_json);
     write_stdout_response(ManageCLIResponse.BucketCreated, data_json);
+}
+
+/** verify_bucket_owner will check if the bucket_owner has an account
+ * in case it does not find one, it would throw an error
+ * @param {string} bucket_owner
+ */
+async function verify_bucket_owner(bucket_owner) {
+    let is_bucket_owner_exist = false;
+    const show_secrets = false;
+    const fs_context = native_fs_utils.get_process_fs_context();
+    const entries = await nb_native().fs.readdir(fs_context, accounts_dir_path);
+    // Gap - should replace this implementation
+    // it keeps iterating even if we find that the bucket owner exist
+    await P.map_with_concurrency(10, entries, async entry => {
+        if (entry.name.endsWith('.json') && !is_bucket_owner_exist) {
+            const full_path = path.join(accounts_dir_path, entry.name);
+            const data = await get_config_data(full_path, show_secrets);
+            if (data.email === bucket_owner) {
+                is_bucket_owner_exist = true;
+            }
+        }
+    });
+
+    if (!is_bucket_owner_exist) {
+        const detail_msg = `bucket owner ${bucket_owner} does not exists`;
+        throw_cli_error(ManageCLIError.BucketSetForbiddenNoBucketOwner, detail_msg);
+    }
 }
 
 async function get_bucket_status(data) {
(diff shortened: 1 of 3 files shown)


shirady:
Waiting for #7723 to be merged first before merging this PR.

src/manage_nsfs/manage_nsfs_cli_errors.js
```diff
@@ -246,6 +246,20 @@ ManageCLIError.BucketAlreadyExists = Object.freeze({
     http_code: 409,
 });
 
+ManageCLIError.BucketCreateForbiddenNoBucketOwner = Object.freeze({
```
romayalon, line 249: IMO we can have same error for update and create

src/cmd/manage_nsfs.js
```diff
@@ -352,6 +376,17 @@ async function update_bucket(data) {
     write_stdout_response(ManageCLIResponse.BucketUpdated, data);
 }
 
+/** verify_update_bucket will check if the bucket_owner has an account
+ * in case it does not find one, it would throw an error
+ * @param {string} bucket_owner
+ */
+async function verify_update_bucket(bucket_owner) {
```
romayalon, line 383: there is no need to have a different function for update, use the same function as for create

src/test/unit_tests/test_nc_nsfs_cli.js
```diff
@@ -255,14 +315,16 @@ mocha.describe('manage_nsfs cli', function() {
             await assert_config_file_permissions(config_root, schema_dir, gpfs_bucket_options.name);
         });
 
-        mocha.it('cli bucket update owner', async function() {
+        mocha.it('cli bucket update owner - should fail bucket owner does not exist', async function() {
```
romayalon, line 318: please keep the happy path test and add a sad path test instead

src/cmd/manage_nsfs.js
```diff
@@ -302,6 +303,31 @@ async function add_bucket(data) {
...
+ * in case it does not find one, it would throw an error
+ * @param {string} bucket_owner
+ */
+async function verify_bucket_owner(bucket_owner) {
+    let is_bucket_owner_exist = false;
+    const show_secrets = false;
+    const fs_context = native_fs_utils.get_process_fs_context();
+    const entries = await nb_native().fs.readdir(fs_context, accounts_dir_path);
+    await P.map_with_concurrency(10, entries, async entry => {
+        if (entry.name.endsWith('.json') && !is_bucket_owner_exist) {
```
romayalon, line 316: why not just throw? even if found bucket_owner === data.email you are still going to iterate the array and check this condition

shirady: I will iterate the array, that's true, but I will not execute the "expensive" operation of `get_config_data`.
If I find that the bucket_owner exists I don't throw the error, I just stop opening the files and read them.

romayalon: there is no reason to keep iterating...

shirady: You're right. But map is not iterating under the hood (and hence I cannot stop/break it), it transforms every element and I'm using it for the concurrency that is already implemented.
This implementation would change after the account identifier would be ID that would pass and the account will be lookup by this ID.
In any way, I would add this gap in the PR and inside the code.

romayalon: in verify_delete_account(), you threw an error in the middle of the loop, why not throw an error here as well?

romayalon (approved):

src/cmd/manage_nsfs.js
shirady: Decided that I would add this as GAP, and we will fix it after the dev freeze (improve performance).

Issue #7715, opened 2024-01-11
NSFS | NC | Health script | add account uid gid can access new_buckets_path check 


romayalon:
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
1. Currently there is no check if an account can access its new_buckets_path during the run of the health script.

### Expected behavior
1. During the health script run, a check if an account can access its new_buckets_path will be executed.

### Steps to reproduce
1. Create a path that only root can access - for example -  sudo mkdir /tmp/root_dir ; chmod 700 /tmp/root_dir/
2. Add an account having uid=200, gid=200, new_buckets_path = /tmp/root_dir/
4. Run health script - currently you won't see an error, after this bug is resolved - expect to see this issue reported.

anandhu-karattu:
Verified on downstream build 02/12

Below is the output from health script . Error is properly reported for the accounts having no access to fs directories

```
[root@ananrhel92-x-worker1 ~]# jq '.' output_all.json
{
  "service_name": "noobaa_nsfs",
  "status": "OK",
  "memory": "338.2M",
  "checks": {
    "services": [
      {
        "name": "noobaa_nsfs",
        "service_status": "active",
        "pid": "2241747",
        "error_type": "PERSISTENT"
      },
      {
        "name": "rsyslog",
        "service_status": "active",
        "pid": "2240893",
        "error_type": "PERSISTENT"
      }
    ],
    "endpoint": {
      "endpoint_state": {
        "response": {
          "response_code": "RUNNING",
          "response_message": "Endpoint running successfuly."
        },
        "total_fork_count": 2,
        "running_workers": [
          2,
          1
        ]
      },
      "error_type": "TEMPORARY"
    },
    "accounts_status": {
      "invalid_accounts": [
        {
          "name": "ibm2",
          "storage_path": "/mnt/fs1/root_dir/",
          "code": "ACCESS_DENIED"
        }
      ],
      "valid_accounts": [
        {
          "name": "ibm1",
          "storage_path": "/mnt/fs1/ibm1/"
        }
      ],
      "error_type": "PERSISTENT"
    },
    "buckets_status": {
      "invalid_buckets": [],
      "valid_buckets": [
        {
          "name": "bucket1000",
          "storage_path": "/mnt/fs1/ibm1/bucket1000"
        }
      ],
      "error_type": "PERSISTENT"
    }
  }
}
```

@naveenpaul1 , marking this as verified

naveenpaul1:
[...]

PR #7738, opened 2024-01-19
NSFS | NC | Health | add account uid gid can access new_buckets_path check

naveenpaul1:
### Explain the changes
1. Add check if an account can access its new_buckets_path during the run of the health script.

1. https://github.com/noobaa/noobaa-core/issues/7715

### Testing Instructions:
1. Create a path that only root can access - for example - sudo mkdir /tmp/root_dir ; chmod 700 /tmp/root_dir/
2. Add an account having uid=200, gid=200, new_buckets_path = /tmp/root_dir/
3. Run health script - currently you won't see an error, after this bug is resolved - expect to see this issue reported.

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/health.js
+++ b/src/cmd/health.js
@@ -72,6 +72,14 @@ const health_errors = {
       error_code: 'INVALID_CONFIG',
       error_message: 'Schema JSON is not valid, Please check the JSON format.',
     },
+    ACCESS_DENIED: {
+      error_code: 'ACCESS_DENIED',
+      error_message: 'Account do no have access to storage path mentioned in schema.',
+    },
+    MISSING_CONFIG: {
+      error_code: 'MISSING_CONFIG',
+      error_message: 'Schema JSON is not found.',
+    }
 };
 
 const fork_response_code = {
@@ -92,7 +100,7 @@ const fork_response_code = {
 const health_errors_tyes = {
   PERSISTENT: 'PERSISTENT',
   TEMPORARY: 'TEMPORARY',
-};
+}
 
 //suppress aws sdk related commands.
 process.env.AWS_SDK_JS_SUPPRESS_MAINTENANCE_MODE_MESSAGE = '1';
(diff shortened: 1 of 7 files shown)


src/cmd/health.js
```diff
@@ -334,6 +346,11 @@ async get_storage_status(config_root, type, all_details) {
       const { data } = await nb_native().fs.readFile(fs_context, config_file_path);
       config_data = JSON.parse(data.toString());
       storage_path = type === 'bucket' ? config_data.path : config_data.nsfs_account_config.new_buckets_path;
+      // check for access in account new_buckets_path dir
+      if (type === 'account') {
+        await nb_native().fs.stat(this.get_account_fs_context(config_data.nsfs_account_config.uid,
```
romayalon, line 351: stat() does not require permissions to the file itself, accoridng to **https://linux.die.net/man/2/stat** -
```
No permissions are required on the file itself, but-in the case of stat() and lstat() - execute (search) permission is required on all of the directories in path that lead to the file.
```

use checkAccess instead -
await nb_native().fs.checkAccess(fs_context, storage_path);

romayalon (changes requested):

src/cmd/health.js
naveenpaul1: @romayalon updated with `checkAccess()`

src/cmd/health.js
```diff
@@ -349,14 +366,21 @@ async get_storage_status(config_root, type, all_details) {
           invalid_storage = {
             name: config_data.name,
             storage_path: storage_path,
-            'code': 'STORAGE_NOT_EXIST',
+            code: health_errors.STORAGE_NOT_EXIST.error_code,
```
romayalon, line 369: we can have ENOENT error on the readFile(config_file_path) also right?
same for the other errors, we need to have a separate try catch for readFile(config_file) and ceckAccess(storage_path)

naveenpaul1: @romayalon chances for getting `ENOENT` in conig file path is impossible since we consider only those files that are already there in the config root path. Missing read access for config file is something we can consider, WDYT?

romayalon: I agree it's not a common case, but it's possible if the bucket was deleted between the time we read the directory and this line, actually the read access is less worrying me because we run this script with sudo

romayalon (approved):

PR #7739, opened 2024-01-19
Support delete flow for log based replication

aayushchouhan09:
### Explain the changes
1. Updated the code for the delete flow in log based replications and to also work with versioned objects.

### DIFF
--- a/src/server/bg_services/log_replication_scanner.js
+++ b/src/server/bg_services/log_replication_scanner.js
@@ -90,7 +90,10 @@ class LogReplicationScanner {
 
                 const sync_versions = rule.sync_versions || false;
 
-                await this.process_candidates(src_bucket, dst_bucket, candidates.items, sync_versions, rule_id, replication_id);
+                const sync_deletions = rule.sync_deletions || false;
+
+                await this.process_candidates(
+                    src_bucket, dst_bucket, candidates.items, sync_versions, sync_deletions, rule_id, replication_id);
 
                 // Commit will save the continuation token for the next scan
                 // This needs to be done only after the candidates were processed.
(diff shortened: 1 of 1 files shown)


config.js
```diff
@@ -624,6 +624,7 @@ config.AWS_LOG_CANDIDATES_LIMIT = 10;
 config.BUCKET_LOG_REPLICATOR_DELAY = 5 * 60 * 1000;
 config.AZURE_QUERY_TRUNCATION_MAX_SIZE_IN_BITS = 10 * 1024 * 1024;
 config.BUCKET_DIFF_FOR_REPLICATION = true;
+config.BUCKET_DIFF_FOR_DELETION = false;
```
liranmauda, line 627: This should not be consumed via the config file

achouhan09: Thanks, updated the code.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -136,18 +140,18 @@ class LogReplicationScanner {
...
-    async process_candidates_sync_version(src_bucket, dst_bucket, candidates) {
+    async process_candidates_sync_version(src_bucket, dst_bucket, candidates, sync_deletions) {
         const bucketDiff = new BucketDiff({
             first_bucket: src_bucket.name,
             second_bucket: dst_bucket.name,
             version: true,
             connection: this.noobaa_connection,
-            for_replication: config.BUCKET_DIFF_FOR_REPLICATION
+            for_replication: config.BUCKET_DIFF_FOR_REPLICATION,
+            for_deletion: sync_deletions
```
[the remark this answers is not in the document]
achouhan09: I have added `for_deletion` in my latest PR, it was already merged so you can see the changes.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -159,22 +163,24 @@ class LogReplicationScanner {
...
                 });
-                // Currently, as get_buckets_diff is not supporting deletions, we will just pass the keys_diff_map as copy
-                // This needs to be reevaluated once the delete is supported.
-                copy_keys = { ...copy_keys, ...keys_diff_map };
+
+                if (sync_deletions) {
+                    delete_keys = { ...delete_keys, ...keys_diff_map};
+                } else {
+                    copy_keys = { ...copy_keys, ...keys_diff_map };
+                }
```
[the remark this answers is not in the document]
achouhan09: @liranmauda @Neon-White we can do something like this: 
```suggestion
              if (sync_versions) {
                  ({ copy_keys } = await this.process_candidates_sync_version(src_bucket, dst_bucket, candidates, false));
              } else if (sync_versions && sync_deletions) { // if sync deletions is enabled then we capture delete keys
                  ({ delete_keys } = await this.process_candidates_sync_version(src_bucket, dst_bucket, candidates, sync_deletions));
              } else {
                  ({ copy_keys, delete_keys } = await this.process_candidates_not_sync_version(src_bucket, dst_bucket, candidates));
              }
```

src/server/bg_services/log_replication_scanner.js
```diff
@@ -90,7 +90,14 @@ class LogReplicationScanner {
 
                 const sync_versions = rule.sync_versions || false;
 
-                await this.process_candidates(src_bucket, dst_bucket, candidates.items, sync_versions, rule_id, replication_id);
+                const sync_deletions = rule.sync_deletions || false;
+
+                if (!sync_versions && sync_deletions) {
+                    throw new Error('Cannot enable sync_deletions when sync_versions is not enabled');
+                }
```
liranmauda, line 97: This is not true.
in log based you can have sync_deletions without sync_versions

liranmauda (changes requested):

src/server/bg_services/log_replication_scanner.js
```diff
@@ -111,14 +114,15 @@ class LogReplicationScanner {
...
      * @param {string} rule_id
      * @param {string} replication_id
      */
-    async process_candidates(src_bucket, dst_bucket, candidates, sync_versions, rule_id, replication_id) {
+    async process_candidates(src_bucket, dst_bucket, candidates, sync_versions, sync_deletions, rule_id, replication_id) {
         let copy_keys;
         let delete_keys;
         if (sync_versions) {
-            ({ copy_keys, delete_keys } = await this.process_candidates_sync_version(src_bucket, dst_bucket, candidates));
+            ({ copy_keys, delete_keys } = await this.process_candidates_sync_version(src_bucket, dst_bucket, candidates, sync_deletions));
```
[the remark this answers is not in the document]
liranmauda: This is a valid flow, even today, we can sync deletion as it is based on the logs.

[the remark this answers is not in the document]
achouhan09: Thanks for clarification.

src/server/bg_services/log_replication_scanner.js
achouhan09: Removed, thanks

src/server/bg_services/log_replication_scanner.js
```diff
@@ -111,18 +118,21 @@ class LogReplicationScanner {
...
      */
-    async process_candidates(src_bucket, dst_bucket, candidates, sync_versions, rule_id, replication_id) {
+    async process_candidates(src_bucket, dst_bucket, candidates, sync_versions, sync_deletions, rule_id, replication_id) {
         let copy_keys;
         let delete_keys;
         if (sync_versions) {
-            ({ copy_keys, delete_keys } = await this.process_candidates_sync_version(src_bucket, dst_bucket, candidates));
+            copy_keys = await this.process_candidates_sync_version(src_bucket, dst_bucket, candidates, false);
+            if (sync_deletions) { // If sync_deletions is enabled, then only the deletion keys for versioned objects are captured
+                delete_keys = await this.process_candidates_sync_version(src_bucket, dst_bucket, candidates, sync_deletions);
```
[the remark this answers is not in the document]
achouhan09: Done.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -136,45 +146,41 @@ class LogReplicationScanner {
...
             connection: this.noobaa_connection,
-            for_replication: config.BUCKET_DIFF_FOR_REPLICATION
+            for_replication: config.BUCKET_DIFF_FOR_REPLICATION,
+            for_deletion: sync_deletions
         });
 
-        let copy_keys = {};
-        // TODO: support delete flow in object versions replication
-        // const delete_keys = [];
+        let diff_keys = {};
```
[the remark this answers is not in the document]
achouhan09: Done.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -136,45 +143,41 @@ class LogReplicationScanner {
...
+                diff_keys = { ...diff_keys, ...keys_diff_map };
             }
         }
 
-        dbg.log1('process_candidates_sync_version:: copy_keys', copy_keys);
-        //TODO support delete flow and return also the delete
-        // returning copy_keys and delete_keys after processing candidates
-        // return { copy_keys, delete_keys };
-        return { copy_keys, delete_keys: [] };
+        dbg.log1('process_candidates_sync_version:: diff_keys', diff_keys);
```
liranmauda, line 178: We might want to add here the value of sync_deletions

liranmauda (changes requested):
in https://github.com/noobaa/noobaa-core/pull/7739/files#diff-2c366e383f2d6a122a16b2efbfe9c12f900c6ddc54e9b5074c92265efa4b3facR137-R140
I am not sure we can do Promise all when there is a versioning.
If there is a versions and we intend on deleting, we should first do the copy of all the versions, then do the delete (we should avoid doing that in parallel)

src/server/bg_services/log_replication_scanner.js
achouhan09: Sure, will add that.

achouhan09:
> in https://github.com/noobaa/noobaa-core/pull/7739/files#diff-2c366e383f2d6a122a16b2efbfe9c12f900c6ddc54e9b5074c92265efa4b3facR137-R140 I am not sure we can do Promise all when there is a versioning. If there is a versions and we intend on deleting, we should first do the copy of all the versions, then do the delete (we should avoid doing that in parallel)

Yes, make sense.

src/server/bg_services/log_replication_scanner.js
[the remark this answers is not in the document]
achouhan09: @Neon-White Just to update here:
`process_candidates_sync_version`  is a generic function which calculates diff_keys (either copy or delete) on the basis of `sync_deletions` flag, if we create a new function like `process_candidates_sync_deletion` then we will be having the same logical function(redundant code) but only with different flag value. 

@liranmauda wdyt?

src/server/bg_services/log_replication_scanner.js
```diff
@@ -111,70 +114,74 @@ class LogReplicationScanner {
...
+            // for sync_versions enabled, deletions cannot be performed until the copying process is completed
+            await this.copy_objects(src_bucket.name, dst_bucket.name, copy_keys, rule_id, replication_id);
+
+            if (sync_deletions) { // If sync_deletions is enabled, then only the deletion keys for versioned objects are captured
+                const diff_keys = await this.process_candidates_sync_version(src_bucket, dst_bucket, candidates, sync_deletions);
+                delete_keys = Object.keys(diff_keys); // fetching keys array from diff_keys
+                await this.delete_objects(dst_bucket.name, delete_keys);
+            }
         } else {
             ({ copy_keys, delete_keys } = await this.process_candidates_not_sync_version(src_bucket, dst_bucket, candidates));
```
achouhan09, line 135: @liranmauda Do we need to put comment like:

```suggestion
            // here even if sync_deletions is disabled, we are processing delete_keys for not_sync_verison
            ({ copy_keys, delete_keys } = await this.process_candidates_not_sync_version(src_bucket, dst_bucket, candidates));
```

src/server/bg_services/log_replication_scanner.js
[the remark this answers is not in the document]
achouhan09: @Neon-White As we are only differentiating these two functions on the basis of `sync_deletions` flag value, so I don't think we need to create two separate functions(having same logic). But yes I agree with your point of changing the name, I do suggest `process_candidates_sync_version_and_deletion` or so.

src/server/bg_services/log_replication_scanner.js
achouhan09: @Neon-White wdyt?

src/server/bg_services/log_replication_scanner.js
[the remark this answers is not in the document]
achouhan09: For this PR we are going with the current name `process_candidate_sync_version` . We will change it accordingly if required in future. Resolving with this comment. Thanks

src/server/bg_services/log_replication_scanner.js
achouhan09: I am not sure, but may be it is the required to call both `copy_objects` and `delete_objects` in parallel for not_sync_version.

liranmauda (approved):

src/server/bg_services/log_replication_scanner.js
achouhan09: Putting the above comment for now. Thanks

PR #7745, opened 2024-01-22
NSFS | NC | CLI | List config files performance improvement & added new filters for list accounts/buckets

### LINKED ISSUE #7673: NSFS | NC | CLI | Extend list command with filters
### Environment info
- NooBaa Version: master

### Actual behavior
1. list currently does not have any filters.

### Expected behavior
2. list accounts while specifying filters.

Details:
A new requirement for having a better user experience is to extend the list operation with a few filters - 
1. filter accounts by uid:gid pair
2. filter accounts by access_key
3. filter accounts by name, etc ( this might make the status command redundant )
4. filter accounts by distinguished name

romayalon:
### Explain the changes
1. Check wide before reading the config files will read the config files only when needed.
2. removed redundant filter loop and added the check in the list config files itself
3. Added user, name, access_key filters to list accounts 
4. Added name filter to list buckets
5. added unit tests

1. Fixed https://github.com/noobaa/noobaa-core/issues/7673

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -19,7 +19,8 @@ const ManageCLIResponse = require('../manage_nsfs/manage_nsfs_cli_responses').Ma
 const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
 const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
 const { print_usage } = require('../manage_nsfs/manage_nsfs_help_utils');
-const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, TYPE_STRING_OR_NUMBER } = require('../manage_nsfs/manage_nsfs_constants');
+const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, TYPE_STRING_OR_NUMBER,
+    LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS} = require('../manage_nsfs/manage_nsfs_constants');
 
 function throw_cli_error(error_code, detail) {
     const err = new ManageCLIError(error_code).to_string(detail);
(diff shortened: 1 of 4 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -600,40 +599,78 @@ async function manage_account_operations(action, data, show_secrets, argv) {
...
     const fs_context = native_fs_utils.get_process_fs_context();
     const entries = await nb_native().fs.readdir(fs_context, config_path);
+    const should_filter = Object.keys(filters).length > 0;
 
     let config_files_list = await P.map_with_concurrency(10, entries, async entry => {
         if (entry.name.endsWith('.json')) {
-            const full_path = path.join(config_path, entry.name);
-            const data = await get_config_data(full_path, show_secrets);
-            return data;
+            if (wide || should_filter) {
```
naveenpaul1, line 665: @romayalon please add these changes to CLI doc,

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -372,6 +381,52 @@ describe('manage nsfs cli account flow', () => {
             expect(JSON.parse(res.stdout).error.message).toBe(ManageCLIError.InvalidArgument.message);
         });
 
+        it('cli list filter by user (account 4)', async () => {
```
naveenpaul1, line 384: Could you please add test for bucket aslo.

romayalon: added thanks

src/cmd/manage_nsfs.js
romayalon: what do you mean? we currently don't put every option of the CLI in the docs but I think we should in a different PR that takes everything from help and describes everything more deeply
I added it to the help for now

naveenpaul1 (approved):

PR #7758, opened 2024-01-24
NSFS | NC | Add Details to Error Message When Invalid Schema

### LINKED ISSUE #7752: NSFS | NC | NSFS account/ bucket/config schema validation error message is not informative enough
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
1. currently on NSFS schema validation we throw the following error - 
`if (!valid) throw new RpcError('INVALID_SCHEMA', ajv.errors[0]?.message);`
ajv.errors[0]?.message - does not state what property caused the failure, it contains the reason for failure only.

### Expected behavior
NSFS account/ bucket/config schema validation error message -  contains the property that caused the failure, it contains the reason for failure only.
Note: we might need to add more info in addition to the property

### Steps to reproduce
1. Update config.json with wrong schema property
2. run nsfs.js/manage_nsfs.js/health.js

shirady:
### Explain the changes
1. Add details to the error message when invalid schema.
2. Update the tests to check the basic message (without the additional info).
3. Add JSDoc on the functions in the file `nsfs_schema_utils`.

### Issues: Fixed #7752
1. Today we send the error message of `ajv.errors[0]?.message`, there are cases where we would like to have more details in the message - for example: if there is an additional property it is "must NOT have additional properties" without the property name.

### Testing Instructions:
1. Create an account: `sudo node src/cmd/manage_nsfs account add --name <account-name> --email <email> --new_buckets_path <path> --uid <uid> --gid <gid>`.
4. Create a bucket: `sudo node src/cmd/manage_nsfs bucket add --name <bucket-name> --email <account-email> --path <path>`.
5. Change the details insides the config: `sudo vi /etc/noobaa.conf.d/buckets/<bucket-name>.json` to invalid, examples:
(1) add additional property `"lala": "lala"`, (2) remove required property `creation_date`, (3) change type `"should_create_underlying_storage":123`
6. Update the bucket (for example unset `fs_backend`): `sudo node src/cmd/manage_nsfs bucket update --name <bucket-name> --fs_backend ''`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/manage_nsfs/nsfs_schema_utils.js
+++ b/src/manage_nsfs/nsfs_schema_utils.js
@@ -32,19 +32,53 @@ schema_utils.strictify(account_schema, {
 // NOTE - DO NOT strictify nsfs_config_schema
 // we might want to use it in the future for adding additional properties
 
+/**
+ * validate_account_schema validates an account object against the NC NSFS account schema
+ * @param {object} account
+ */
 function validate_account_schema(account) {
     const valid = ajv.validate(account_schema, account);
-    if (!valid) throw new RpcError('INVALID_SCHEMA', ajv.errors[0]?.message);
+    if (!valid) {
+        const err_msg = ajv.errors[0].message ? create_schema_err_msg(ajv.errors[0]) : undefined;
+        throw new RpcError('INVALID_SCHEMA', err_msg);
+    }
 }
 
+/**
+ * validate_bucket_schema validates a bucket object against the NC NSFS bucket schema
+ * @param {object} bucket
+ */
 function validate_bucket_schema(bucket) {
     const valid = ajv.validate(bucket_schema, bucket);
-    if (!valid) throw new RpcError('INVALID_SCHEMA', ajv.errors[0]?.message);
+    if (!valid) {
+        const err_msg = ajv.errors[0].message ? create_schema_err_msg(ajv.errors[0]) : undefined;
+        throw new RpcError('INVALID_SCHEMA', err_msg);
+    }
 }
 
+/**
+ * validate_nsfs_config_schema validates a config object against the NC NSFS config schema
+ * @param {object} config
+ */
 function validate_nsfs_config_schema(config) {
     const valid = ajv.validate(nsfs_config_schema, config);
-    if (!valid) throw new RpcError('INVALID_SCHEMA', ajv.errors[0]?.message);
+    if (!valid) {
+        const err_msg = ajv.errors[0].message ? create_schema_err_msg(ajv.errors[0]) : undefined;
+        throw new RpcError('INVALID_SCHEMA', err_msg);
+    }
+}
+
+/**
+ * create_schema_err_msg would use the original error message we got from avj
+ * and adds additional info in case we have it
+ * @param {object} err
+ */
+function create_schema_err_msg(err) {
+    if (!err || !err.message) return;
+    const err_msg = [err.message];
+    if (err.params) err_msg.push(JSON.stringify(err.params));
+    if (err.instancePath) err_msg.push(JSON.stringify(err.instancePath));
+    return err_msg.join(' | ');
 }
 
 //EXPORTS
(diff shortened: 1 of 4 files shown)


src/manage_nsfs/nsfs_schema_utils.js
```diff
@@ -32,19 +32,59 @@ schema_utils.strictify(account_schema, {
...
+
+/**
+ * create_schema_err_msg would use the original error message
+ * and adds additional info in case we have it
+ * @param {object} err
+ */
+function create_schema_err_msg(err) {
+    const original_err_msg = err?.message;
+    const err_msg = [original_err_msg];
+    let additional_info = err?.params ? JSON.stringify(ajv.errors[0]?.params) : undefined;
```
naveenpaul1, line 79: can we use method `err` here instead for getting it again in `ajv.errors[0]`

src/manage_nsfs/nsfs_schema_utils.js
```diff
@@ -32,19 +32,59 @@ schema_utils.strictify(account_schema, {
...
 // we might want to use it in the future for adding additional properties
 
+/**
+ * validate_bucket_schema validates a account object against the NC NSFS account schema
+ * @param {object} account
+ */
 function validate_account_schema(account) {
     const valid = ajv.validate(account_schema, account);
-    if (!valid) throw new RpcError('INVALID_SCHEMA', ajv.errors[0]?.message);
+    if (!valid) {
```
naveenpaul1, line 41: Please check `ajv.errors[0]` also there is no point in calling method without valid `ajv.errors[0]`, then we can remove the same check in multiple places in `create_schema_err_msg()`,

src/manage_nsfs/nsfs_schema_utils.js
```diff
@@ -32,19 +32,59 @@ schema_utils.strictify(account_schema, {
...
+ * @param {object} err
+ */
+function create_schema_err_msg(err) {
+    const original_err_msg = err?.message;
+    const err_msg = [original_err_msg];
+    let additional_info = err?.params ? JSON.stringify(ajv.errors[0]?.params) : undefined;
+    if (additional_info) {
+        err_msg.push(additional_info);
+    }
+    additional_info = err?.instancePath ? JSON.stringify(ajv.errors[0]?.instancePath) : undefined;
```
naveenpaul1, line 83: you are already checking for the `ajv.errors[0]?.instancePath` here can you move the push also inside the check.
```suggestion
   err?.params ? err_msg.push(JSON.stringify(err.params)) : undefined;
```

src/manage_nsfs/nsfs_schema_utils.js
```diff
@@ -32,19 +32,57 @@ schema_utils.strictify(account_schema, {
 // NOTE - DO NOT strictify nsfs_config_schema
 // we might want to use it in the future for adding additional properties
 
+/**
+ * validate_bucket_schema validates a account object against the NC NSFS account schema
```
romayalon, line 36:
```suggestion
 * validate_account_schema validates an account object against the NC NSFS account schema
```

src/manage_nsfs/nsfs_schema_utils.js
```diff
@@ -32,19 +32,57 @@ schema_utils.strictify(account_schema, {
...
+ * @param {object} account
+ */
 function validate_account_schema(account) {
     const valid = ajv.validate(account_schema, account);
-    if (!valid) throw new RpcError('INVALID_SCHEMA', ajv.errors[0]?.message);
+    if (!valid && ajv.errors[0]) {
```
romayalon, line 41: what should we throw if !valid but ajv.errors[0] is undefined?

shirady: I changed the code, I hope it is more clear.
As I mentioned, I don't understand in which case the schema will be invalid and there will be an error without a message, which was in the code before.

romayalon (approved):

Issue #7753, opened 2024-01-23
NSFS | NC | Events bugs & events addition to manage nsfs CLI


romayalon:
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
1. Currently there are no events created during the manage_nsfs cli usage
2. Currently there are no buckets events
3. bucketspace_fs.read_account_by_access_key() - has incorrect event
4.  bucketspace_fs.read_bucket_sdk_info() - we always throw noobaa_bucket_not_found event and it's not necessarily always the issue, need to be more specific.

PR #7759, opened 2024-01-24
NSFS | NC | Events | bugs & events addition to manage nsfs CLI

naveenpaul1:
### Explain the changes
1. Currently there are no events created during the manage_nsfs cli usage
2. Currently there are no buckets events
3. bucketspace_fs.read_account_by_access_key() - has incorrect event
4. bucketspace_fs.read_bucket_sdk_info() - we always throw noobaa_bucket_not_found event and it's not necessarily 
always the issue, need to be more specific.

1. https://github.com/noobaa/noobaa-core/issues/7753

### Testing Instructions:
1. New events are added for account/bucket create/delete. 
2. Event added for whitelist update
3. verify the events in var/logs/noobaa-events.log after running the CLI or S3 commands

- [X] Doc added/updated
- [ ] Tests added

### DIFF
--- a/docs/dev_guide/non_containerized_NSFS_events.md
+++ b/docs/dev_guide/non_containerized_NSFS_events.md
@@ -81,7 +81,7 @@ arguments : `bucket_path`, `object_name`
 #### Resolutions
 - Verify the bucket path.
 
-### 10. noobaa_object_uplaod_failed
+### 10. noobaa_object_upload_failed
 arguments : `bucket_path`, `object_name`
 #### Reasons
 - Bucket path is outside the bucket boundaries.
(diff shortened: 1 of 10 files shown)


docs/dev_guide/non_containerized_NSFS_events.md
```diff
@@ -89,4 +89,39 @@ arguments : `bucket_path`, `object_name`
...
 #### Resolutions
 - Make sure bucket storage class is supported.
-- Check for I/O operations.
\ No newline at end of file
+- Check for I/O operations.
+
+### 11. account_created
+arguments : `account_name`
+#### Reasons
+- Bucket created using CLI command
```
romayalon, line 97:
```suggestion
- Account created using CLI command
```

docs/dev_guide/non_containerized_NSFS_events.md
```diff
@@ -89,4 +89,39 @@ arguments : `bucket_path`, `object_name`
...
+arguments : `account_name`
+#### Reasons
+- Account deleted using CLI.
+#### Resolutions
+- Nil
+
+### 13. bucket_created
+arguments : `bucket_name`
+#### Reasons
+- Bucket creatd using CLI or S3.
```
romayalon, line 111:
```suggestion
- Bucket created using CLI or S3.
```

src/sdk/bucketspace_fs.js
```diff
@@ -26,6 +26,29 @@ const ACCOUNT_PATH = 'accounts';
 const ACCESS_KEYS_PATH = 'access_keys';
 const bucket_semaphore = new KeysSemaphore(1);
 
+const BUCKET_EVENT_ERROR_CODE = {
+    NO_SUCH_BUCKET: {
+        event_code: 'noobaa_no_such_bucket',
+        event_message: 'Bucket not found, bucket name : ',
+    },
+    BUCKET_ALREADY_EXISTS: {
+        event_code: 'noobaa_bucket_alread_exists',
```
romayalon, line 35:
```suggestion
        event_code: 'noobaa_bucket_already_exists',
```

src/sdk/bucketspace_fs.js
```diff
@@ -26,6 +26,29 @@ const ACCOUNT_PATH = 'accounts';
 const ACCESS_KEYS_PATH = 'access_keys';
 const bucket_semaphore = new KeysSemaphore(1);
 
+const BUCKET_EVENT_ERROR_CODE = {
```
romayalon, line 29: loved this struct, a few comments -
1. can we move it to a new manage_nsfs_events_utils.js file?
2. can we have a similar struct for accounts? and for IO events?
3. In this new events utils let's try create a function that receives the code and the arguments and runs the dbg.events since we now have many large events blocks in our code that I think we can reduce to a single line function call.

naveenpaul1: @romayalon regarding point 3, When we try to move the `dbg.event()` to util file, we need to pass most of the arguments as parameters and this will result in a very long method call with a small change in number of lines.
we need to pass code, event_type, message, description, severity, state arguments, because these aruments are dependent on error types.

romayalon: what I meant is to pre-define these in a struct and just pass the code and the arguments, WDYT?

naveenpaul1: @romayalon ya, that is something I can try out. I need to replace duplicate event code with unique code.

src/util/manage_nsfs_events_utils.js
```diff
@@ -0,0 +1,266 @@
...
+            event_type: 'ERROR',
+            scope: 'NODE',
+            severity: 'ERROR',
+            state: 'HEALTHY',
+        }
+    }
+};
+
+async function create_event(event_from, event_code, name, arg, err) {
+    const description = EVENT_ERROR_CODE[event_from][event_code].description + (err ? ', error: ' + err : '');
```
romayalon, line 250: worth adding a variable, so we won't need to evaluate this every time -
const event = EVENT_ERROR_CODE[event_from][event_code];
and then use event.event_code/ event.entity_type / etc...

src/util/manage_nsfs_events_utils.js
```diff
@@ -0,0 +1,266 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const dbg = require('../util/debug_module')(__filename);
```
romayalon, line 4: can you please move this file to be under src/manage_nsfs/ directory ?

src/cmd/manage_nsfs.js
```diff
@@ -206,6 +207,7 @@ async function add_bucket(data) {
     // for validating against the schema we need an object, hence we parse it back to object
     nsfs_schema_utils.validate_bucket_schema(JSON.parse(data_json));
     await native_fs_utils.create_config_file(fs_context, buckets_dir_path, bucket_conf_path, data_json);
+    manage_nsfs_events_utils.create_event('BUCKET', 'BUCKET_CREATED', data.name, {bucket_name: data.name});
     write_stdout_response(ManageCLIResponse.BucketCreated, data_json);
```
romayalon, line 211: It looks like we missed some events, mainly error events,
I think that we might want to consider calling the create_event() function from 
write_stdout_response() according to the response code and from throw_cli_error() according to the error we got.
this way we won't miss existing events, I would add a mapping between responses to errors to events
and if the event exists we will call create_event()

naveenpaul1: @romayalon do we need to have an event for all the error scenarios, some CLI errors due to invalid flags and values, And do not point to the state of the system. if we call create_event() from write_stdout_response() it will create an event even for simple status checks and bucket and account listings. I don't see how this will help the monitoring team. WDYT?

romayalon: ofc not, I agree that there are CLI failures that are definetly should not create an event.
I would create an event only to the errors mapped to the events we already have ( for example ManageCLIResponse.BucketCreated is mapped to 'BUCKET_CREATED' event)

naveenpaul1: oh.. ok, Sure will do that

src/manage_nsfs/manage_nsfs_events_utils.js
```diff
@@ -0,0 +1,331 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const dbg = require('../util/debug_module')(__filename);
+const _ = require('lodash');
+
+
+const EVENT_ERROR_CODE = {
+    NOOBAA: {
+        FORK_EXIT: {
```
romayalon, line 10: I think it'll be nicer to have the same format we have in errors and responses, WDYT?
ManageCLIEvent.FORK_EXIT = Object.freeze({
    event_code: 'noobaa_fork_exit',
    entity_type: 'NODE',
    event_type: 'STATE_CHANGE',
    message: 'Noobaa fork exited',
    description: 'Noobaa fork exited due to internal error',
    scope: 'NODE',
    severity: 'ERROR',
    state: 'DEGRADED',
});
also, it'll make our life easier on the mapping so we won't need the error_from, WDYT?

naveenpaul1: @romayalon `Object.freeze()` objects are not mutatable. right? For most of the events, we need to update the message, description, and arguments. I am not sure we can do that with `Object.freeze()`.

romayalon (approved):

PR #7761, opened 2024-01-24
[ESLint Fixes] moving to @stylistic plugin + fixes

jackyalbo:
### Explain the changes
1. In addition to prior PR, seems like moving to stylistic plugin is pretty straightforward

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,6 +21,8 @@ module.exports = {
 
 
     plugins: [
+        // stylistic validates style related rules: https://eslint.style/
+        '@stylistic/js',
         // eslint-plugin-header validates copyright header
         'header'
     ],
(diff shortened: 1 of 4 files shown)


liranmauda (approved):

.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
 
     plugins: [
         // eslint-plugin-header validates copyright header
+        '@stylistic/js',
```
shirady, line 25: Please move this after the 'header' (the comment above refers to it).

shirady (approved):
LGTM - minor comment

shirady:
@liranmauda should this PR separated into a couple of commits?
1. Adding the plugin + and it's changes.
2. Fixing existing errors (in file `health.js`).

liranmauda:
> @liranmauda should this PR separated into a couple of commits?
> 
> 1. Adding the plugin + and it's changes.
> 2. Fixing existing errors (in file `health.js`).


Basically, yes.
This case is where we will want to backport only the eslint fixes to a branch where we don't have the `health.js`

I missed it, Great catch.
as this PR is merged, if we need to backport we will need to drop this conflict.

PR #7762, opened 2024-01-24
Do not bundle node tar in RPM build

b-ranto:
### Explain the changes
1. We should not bundle node binaries inside the RPM build of NooBaa Core. Moving it to requires instead. This blocks multi-arch builds/deployments.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/deploy/RPM_build/noobaa.spec
+++ b/src/deploy/RPM_build/noobaa.spec
@@ -1,11 +1,9 @@
 %define revision null
 %define noobaaver null
-%define nodever null
 %define releasedate null
 %define changelogdata null
 
 %define noobaatar %{name}-%{version}-%{revision}.tar.gz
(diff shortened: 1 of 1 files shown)


tangledbytes:
Need to confirm if Tokyo team still needs the ability to perform air-gap deployment.

b-ranto:
If the Tokyo confirms they don't need it (which they shouldn't in any case, this is not how you achieve air-gaped environments...) then we can remove more code.

I still think we can merge this as is (upon some more testing) and we can create a custom spec file for Tokyo team if necessary, the node tarball is still being created, that is why I did not touch that code.

tangledbytes:
@guymguym

b-ranto:
I think we can close this now as we are heading in a different direction with the builds.

Issue #7764, opened 2024-01-25
command is showing internal server error message


ramya-c3:
### Environment info
- NooBaa Version: noobaa-core-5.15.0-20240122.el9.x86_64
- Platform: Standalone

### Actual behavior
1.[root@ramyamc-32 noobaa-core]# node src/cmd/manage_nsfs account add --name test --email pr1 --uid 1235 --gid 234 --new_buckets_path 4e34 2>/dev/null
{
  "error": {
    "code": "InternalError",
    "message": "The server encountered an internal error. Please retry the request",
    "detail": "TypeError: A string was expected\n    at Object.is_path_exists (/usr/local/noobaa-core/src/util/native_fs_utils.js:463:30)\n    at validate_account_args (/usr/local/noobaa-core/src/cmd/manage_nsfs.js:757:46)\n    at add_account (/usr/local/noobaa-core/src/cmd/manage_nsfs.js:448:11)\n    at manage_account_operations (/usr/local/noobaa-core/src/cmd/manage_nsfs.js:584:15)\n    at account_management (/usr/local/noobaa-core/src/cmd/manage_nsfs.js:329:11)\n    at async main (/usr/local/noobaa-core/src/cmd/manage_nsfs.js:97:13)"
  }
}

### Expected behavior
1. It should display proper error instead of internal server error

### Steps to reproduce
1.

shirady:
The issue is when someone runs the command to add an account with the flag `--new_buckets_path` with an invalid value, (example 4e34), he sees `InternalError` instead of an error about this specific flag.

shirady:
@ramya-c3 I removed the "help" from the title.

romayalon:
@ramya-c3 that's a cute bug actually
@shirady yes, as I mentioned on slack, this is a type-casting problem.
we need to make sure, in all arguments and not only in this specific case.

romayalon:
@shirady another instance of a similar issue, we need to think about more of these cases
```
sudo node src/cmd/manage_nsfs bucket add  --name bucket2 --email noobaa@gmail.com --path 4e34
Jan-25 16:26:02.777 [/72027]   [LOG] CONSOLE:: config_path 4e+34
Jan-25 16:26:02.779 [/72027]   [LOG] CONSOLE::config_path err 4e+34 TypeError: A string was expected
{
  "error": {
    "code": "InternalError",
    "message": "The server encountered an internal error. Please retry the request",
[...]

PR #7765, opened 2024-01-25
NSFS | NC | Health script - PR 1/2 | new_buckets_path bug fixes

romayalon:
### Explain the changes
1. Added handling for new_buckets_path access check when account has only user (distinguished name) and not uid and gid.
2. CheckAcess/Stat only checked 'r' permission, now checking also 'w' permissions.
3. Added jest unit tests for rw_check
4. Added new tests to the health tests.
5. refactored a bit the health script. 

1. In the next PR i'll add the same check to manage_nsfs create/update account.

### Testing Instructions:
1. sudo jest --testRegex=jest_tests/test_nc_nsfs_new_buckets
2. sudo node mocha noobaa-core/src/test/unit_tests/test_nc_nsfs_health.js

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/cmd/health.js
+++ b/src/cmd/health.js
@@ -2,15 +2,17 @@
 'use strict';
 
 const dbg = require('../util/debug_module')(__filename);
-const os_util = require('../util/os_utils');
-const { make_https_request } = require('../util/http_utils');
-const minimist = require('minimist');
-const config = require('../../config');
 const path = require('path');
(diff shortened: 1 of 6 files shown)


shirady:
@romayalon will this would fix issue #7764 (it is a new issue)?
If so, you can add it to the description

naveenpaul1 (approved):

PR #7779, opened 2024-01-31
NSFS | NC | Use `--from_file` Option in Account Create

shirady:
### Explain the changes
1. Add the `from_file` option in the account that would include validation. The JSON file will include options (as key-value structure) that are the same as the flags we suggest. 
2. Change the structure inside the main so that eventually it doesn't matter whether the parameters are from the flags arguments (argv) or file.
3. Move `get_access_keys` to a separate function.
4. Fix indentation in `fetch_account_data` in the part of `access_key` as `SensitiveString`.
5. Remove an unused parameter from `validate_options_type_by_value`.
6. In `native_fs_utils.js` rename the function `read_config_file` to `read_file`.
7. Update json in doc so `_id` will appear first (as fixed in this PR).

### Issues: Fixed [BZ 2260325](https://bugzilla.redhat.com/show_bug.cgi?id=2260325), [BZ 2262997](https://bugzilla.redhat.com/show_bug.cgi?id=2262997), [BZ 2262992](https://bugzilla.redhat.com/show_bug.cgi?id=2262992)
1. Currently the file that was read in QE tests was a JSON file that was similar to the account config file, so when creating an account using the flag `--from_file` and in the JSON there is no `creation_date` property we do not allow the account creation (it fails on the schema validation).
2. In addition to that, when creating an account using the flag `--from_file` the account credentials standards were not checked.
3. For [BZ 2262992](https://bugzilla.redhat.com/show_bug.cgi?id=2262992) - waiting for PR #7827 to be merged before merging this PR.
4. This PR also fixed the mentioned gap regarding "id" appearing last in the JSON output.
5. GAP - doing the same changes in bucket add (a following PR after this would be merged).
6. GAP - might need to refactor `set_access_keys()` function.

### Testing Instructions:
Added unit tests, please run: `sudo npx jest test_nc_nsfs_account_cli.test.js`

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/docs/non_containerized_NSFS.md
+++ b/docs/non_containerized_NSFS.md
@@ -468,13 +468,32 @@ sudo node src/cmd/manage_nsfs bucket delete --config_root ../standalon/config_ro
 sudo node src/cmd/manage_nsfs bucket list --config_root ../standalon/config_root 2>/dev/null
 
 ```
+NSFS management CLI run will create both accounts, access_keys, and buckets directories if they are missing under the config_root directory.
 **Important**:  All the Account/Bucket commands end with `2>/dev/null` to make sure there are no unwanted logs.
 
-
+Using `from_file` flag:
+- For account and bucket creation users can also pass account or bucket values in JSON file (hereinafter referred to as "options JSON file") instead of passing them in CLI as arguments using flags.
+- General use:
 ```
-sudo node src/cmd/manage_nsfs bucket add --config_root ../standalon/config_root --from_file /json_file/path
+sudo node src/cmd/manage_nsfs account add --config_root ../standalon/config_root --from_file <options_JSON_file_path>
+sudo node src/cmd/manage_nsfs bucket add --config_root ../standalon/config_root --from_file <options_JSON_file_path>
 ```
-NSFS management CLI command will create both account and bucket dir if it's missing in the config_root path.
+- The options are key-value, where the key is the same as suggested flags. For example:
+create JSON file for account:
+```json
+// JSON file of key-value options for creating an account
+{
+    "name": "account-1001",
+    "uid": 1001,
+    "gid": 1001,
+    "new_buckets_path": "/tmp/nsfs_root1/my-bucket"
+}
+```
+```
+sudo node src/cmd/manage_nsfs account add --config_root ../standalon/config_root --from_file <options_JSON_file_path>
+```
+- When using `from_file` flag the details about the account/bucket should be only inside the options JSON file.
+- The JSON config file and JSON options file of account are different! 
 
 ## NSFS Certificate
 
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -22,7 +22,7 @@ const manage_nsfs_glacier = require('../manage_nsfs/manage_nsfs_glacier');
 const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
 const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
 const { print_usage } = require('../manage_nsfs/manage_nsfs_help_utils');
-const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, BOOLEAN_STRING_VALUES,
+const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, FROM_FILE, BOOLEAN_STRING_VALUES,
     LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS, GLACIER_ACTIONS } = require('../manage_nsfs/manage_nsfs_constants');
 const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 
@@ -89,7 +89,8 @@ async function main(argv = minimist(process.argv.slice(2))) {
         if (argv.help || argv.h) {
             return print_usage(type, action);
         }
-        validate_flags_arguments(type, action, argv);
+        const user_input_from_file = await validate_input_types(type, action, argv);
+        const user_input = user_input_from_file || argv;
         config_root = argv.config_root ? String(argv.config_root) : config.NSFS_NC_CONF_DIR;
         if (!config_root) throw_cli_error(ManageCLIError.MissingConfigDirPath);
 
@@ -99,11 +100,12 @@ async function main(argv = minimist(process.argv.slice(2))) {
         config_root_backend = argv.config_root_backend ? String(argv.config_root_backend) : config.NSFS_NC_CONFIG_DIR_BACKEND;
 
         await check_and_create_config_dirs();
-        const from_file = argv.from_file ? String(argv.from_file) : '';
         if (type === TYPES.ACCOUNT) {
-            await account_management(argv, from_file);
+            await account_management(action, user_input);
         } else if (type === TYPES.BUCKET) {
-            await bucket_management(argv, from_file);
+            // GAP - from-file in bucket
+            const path_to_json_options = argv.from_file ? String(argv.from_file) : '';
+            await bucket_management(argv, path_to_json_options);
         } else if (type === TYPES.IP_WHITELIST) {
             await whitelist_ips_management(argv);
         } else if (type === TYPES.GLACIER) {
@@ -333,20 +335,19 @@ async function manage_bucket_operations(action, data, argv) {
     }
 }
 
-async function account_management(argv, from_file) {
-    const action = argv._[1] || '';
-    const show_secrets = get_boolean_or_string_value(argv.show_secrets);
-    const data = await fetch_account_data(argv, from_file);
-    await manage_account_operations(action, data, show_secrets, argv);
+async function account_management(action, user_input) {
+    const show_secrets = get_boolean_or_string_value(user_input.show_secrets);
+    const data = await fetch_account_data(action, user_input);
+    await manage_account_operations(action, data, show_secrets, user_input);
 }
 
 /**
- * set_access_keys will set the access keys either given as args or generated.
- * @param {{ access_key: any; secret_key: any; }} argv
+ * set_access_keys will set the access keys either given or generated.
+ * @param {string} access_key
+ * @param {string} secret_key
  * @param {boolean} generate a flag for generating the access_keys automatically
  */
-function set_access_keys(argv, generate) {
-    const { access_key, secret_key } = argv;
+function set_access_keys(access_key, secret_key, generate) {
     let generated_access_key;
     let generated_secret_key;
     if (generate) {
@@ -362,50 +363,28 @@ function set_access_keys(argv, generate) {
 }
 
 // in name and new_name we allow type number, hence convert it to string
-async function fetch_account_data(argv, from_file) {
-    let data;
-    let access_keys = [{
-        access_key: undefined,
-        secret_key: undefined,
-    }];
-    let new_access_key;
-    const action = argv._[1] || '';
-    if (from_file) {
-        const fs_context = native_fs_utils.get_process_fs_context();
-        const raw_data = (await nb_native().fs.readFile(fs_context, from_file)).data;
-        data = JSON.parse(raw_data.toString());
-        // GAP - from-file is not validated
-    }
-    if (action !== ACTIONS.LIST && action !== ACTIONS.STATUS) _validate_access_keys(argv);
-    if (action === ACTIONS.ADD || action === ACTIONS.STATUS) {
-        const regenerate = action === ACTIONS.ADD;
-        access_keys = set_access_keys(argv, regenerate);
-    } else if (action === ACTIONS.UPDATE) {
-        const regenerate = get_boolean_or_string_value(argv.regenerate);
-        access_keys = set_access_keys(argv, regenerate);
-        new_access_key = access_keys[0].access_key;
-        access_keys[0].access_key = undefined; //Setting it as undefined so we can replace the symlink
-    }
-
-    if (!data) {
-        data = _.omitBy({
-            name: _.isUndefined(argv.name) ? undefined : String(argv.name),
-            email: _.isUndefined(argv.name) ? undefined : String(argv.name), // temp, keep the email internally
-            creation_date: action === ACTIONS.ADD ? new Date().toISOString() : undefined,
-            wide: _.isUndefined(argv.wide) ? undefined : get_boolean_or_string_value(argv.wide),
-            new_name: _.isUndefined(argv.new_name) ? undefined : String(argv.new_name),
-            new_access_key,
-            access_keys,
-            nsfs_account_config: {
-                distinguished_name: argv.user,
-                uid: argv.user ? undefined : argv.uid,
-                gid: argv.user ? undefined : argv.gid,
-                new_buckets_path: argv.new_buckets_path,
-                fs_backend: argv.fs_backend ? String(argv.fs_backend) : config.NSFS_NC_STORAGE_BACKEND
-            }
-        }, _.isUndefined);
-    }
+async function fetch_account_data(action, user_input) {
+    const { access_keys, new_access_key } = get_access_keys(action, user_input);
+    let data = {
+        // added undefined values to keep the order the properties when printing the data object
+        _id: undefined,
+        name: _.isUndefined(user_input.name) ? undefined : String(user_input.name),
+        email: _.isUndefined(user_input.name) ? undefined : String(user_input.name), // temp, keep the email internally
+        creation_date: action === ACTIONS.ADD ? new Date().toISOString() : undefined,
+        wide: _.isUndefined(user_input.wide) ? undefined : get_boolean_or_string_value(user_input.wide),
+        new_name: _.isUndefined(user_input.new_name) ? undefined : String(user_input.new_name),
+        new_access_key,
+        access_keys,
+        nsfs_account_config: {
+            distinguished_name: user_input.user,
+            uid: user_input.user ? undefined : user_input.uid,
+            gid: user_input.user ? undefined : user_input.gid,
+            new_buckets_path: user_input.new_buckets_path,
+            fs_backend: user_input.fs_backend ? String(user_input.fs_backend) : config.NSFS_NC_STORAGE_BACKEND
+        }
+    };
     if (action === ACTIONS.UPDATE || action === ACTIONS.DELETE) {
+        // @ts-ignore
         data = _.omitBy(data, _.isUndefined);
         data = await fetch_existing_account_data(data);
     }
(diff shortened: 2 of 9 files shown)


shirady:
# Manual Testing Instructions:
1. First, create the `FS_ROOT` : `mkdir -p /tmp/new_buckets_path` and give permissions `chmod 777 /tmp/new_buckets_path`.
2. Create a JSON config file (without `creation_date` property), for example:

 account_config.json (change the `access_key` and `secret_key`)
```
{
    “name”: “some_account_name”,
    “email”: “some@email.com”,
    “allow_bucket_creation”: true,
    “access_keys”: [
        {
            “access_key”: “ffbffbaa9b7166ac1f6c22cc09eec55e068edbf3",
            “secret_key”: “034d6cf9ad653c31ef4dc2efadd63c56ad6a16c3"
        }
    ],
    “nsfs_account_config”: {
        “uid”: 1001,
        “gid”: 1001,
        “new_buckets_path”: “/tmp/new_buckets_path”,
        “fs_backend”: “GPFS”
    }
}
```

4. Create an account using the `--from-file` option: `sudo node src/cmd/manage_nsfs account add --from_file account_config.json` (you should see that the account was created with the current date).
5. Edit the JSON config file above change the details and add the `creation_date` property, for example: `“creation_date”: “2024-01-24T13:42:52.435Z”` and rerun the account add command: `sudo node src/cmd/manage_nsfs account add --from_file account_config2.json`

 account_config2.json (change the `access_key` and `secret_key`)
```
{
    “name”: “some_account_name2”,
    “email”: “some2@email.com”,
    “allow_bucket_creation”: true,
    “access_keys”: [
        {
            “access_key”: “<access-key-1>"
            “secret_key”: “<secret-key-1>”
        }
    ],
    “nsfs_account_config”: {
        “uid”: 1002,
        “gid”: 1002,
        “new_buckets_path”: “/tmp/new_buckets_path”,
        “fs_backend”: “GPFS”
    }
}
```

src/cmd/manage_nsfs.js
```diff
@@ -366,6 +366,14 @@ async function fetch_account_data(argv, from_file) {
         const fs_context = native_fs_utils.get_process_fs_context();
         const raw_data = (await nb_native().fs.readFile(fs_context, from_file)).data;
         data = JSON.parse(raw_data.toString());
+        if (action === ACTIONS.ADD) {
+            if (data.creation_date) {
```
romayalon, line 370: hmmm I don't think we should ignore the user's creation date, what is your motivation here?

shirady: The user doesn't define the `creation_date`, the manage nsfs CLI sets it (same as we don't have the flag to set it).
For example, now he can add that the account was created in the year 1800. I don't want to let him do it.

romayalon: hmm I was thinking about upgrade scenarios but I guess it's better to have the same behavior.
can you please add tests and maybe add to the documentation a json imput example of the from file case?

shirady: Yes, I would plan to add.

A similar question about `allow_bucket_creation` - do we want to prevent from passing it in the file?

- If yes, then the JSON example will be changed (simple solution)
- If not
  - then we need to add a flag for this option (set and unset it) 
  - and remove the option to add a bucket when it's false.
  - change the required flag for account add (doesn't need to pass 'new_buckets_path' because this account doesn't create one)
  -etc... (too many changes for this PR scope, will need to open an issue).

romayalon: No, the user can decide whether the account can create buckets or not, why the JSON should be changed?

In a short glance on the NSFS account and bucket schemas there are similar cases to creation_date issue:
1. on NSFS bucket schema - 
    a. should_create_undelying_storage, this should on account add always be false, if the user provided true we should fail.
    b. _id - self generating
    c. bucket owner and owner_account should map to the same account ( this is an issue not only on --from_file flag) 
2. on nsfs account schema - 
    a. _id - self generating

Maybe there are more things like that, please check that as well

shirady: If we prevent him from this then he cannot add this option (and the JSON would not contain the property `allow_bucket_creation`).
Currently, the manage nsfs CLI doesn't support adding this option and neither enforce it (we need to remove the option to add a bucket when it's false). I would open an issue.

romayalon: Currently, allow_bucket_creation option does not exist because its value is inferred based on the new_buckets_path value, we should have the same behavior on --from_file flow, but I don't see an issue with allowing it.
We can consider adding allow_bucket_creation explicitly to account add/update command.

We do not need to enforce it on the CLI, allow_bucket_creation is an option of a specific account while there is no specific account that runs the CLI commands.

shirady: - Util we would add the option adding allow_bucket_creation explicitly to an account add/update command, the `from-file` should be aligned with it.
- By enforcing I meant that an account that has `allow_bucket_creation` `false` (account was created without `new_buckets_path`) should not be able to create a bucket (currently it can).

romayalon: 1. I think that instead of removing an ability that we want in our code from the `from-file` flow and later bring it back, we can just add allow_bucket_creation flag to the arguments flow and use it or infer it on our own if it was not specified.

2. Creating a bucket is blocked if the account is not allowed to create a bucket, see here - https://github.com/noobaa/noobaa-core/commit/e995346640dc5e8026bf1932427677d63921b92c#diff-da078ac8593ed97975c9a5ed039a123ac81a0a1082f5645da1f62d74f9d83580L285
```
bucketspace_fs.create_bucket - 
if (!sdk.requesting_account.allow_bucket_creation) {
      throw new RpcError('UNAUTHORIZED', 'Not allowed to create new buckets')
}
```

shirady: 1. I think it would create bugs. I prefer a known gap, that an issue.
2. I agree with you and I understand it. I think you didn't understand me... so I would ask - Where is the equivalent part in `manage_nsfs` to bucket add command? (I think you will not find this in the code because it is a bug that I'm trying to point to...).

romayalon: regarding point 2 - got you now, so just add to verify_bucket_owner() another check that the owner is allowed to create buckets.

shirady: Regarding point 1 - please see issue #7786.
Regarding point 2 - Please see PR #7785.

naveenpaul1: @shirady @romayalon I think this will create some confusion [7786](https://github.com/noobaa/noobaa-core/issues/7786), What will happen when user provide `allow_bucket_creation` true without the `new-bucket-path`?

docs/non_containerized_NSFS.md
```diff
@@ -467,13 +467,32 @@ sudo node src/cmd/manage_nsfs bucket delete --config_root ../standalon/config_ro
 sudo node src/cmd/manage_nsfs bucket list --config_root ../standalon/config_root 2>/dev/null
 
 ```
+NSFS management CLI command will create both account and bucket dir if it's missing in the config_root path.
 **Important**:  All the Account/Bucket commands end with `2>/dev/null` to make sure there are no unwanted logs.
 
-
+Using `from_file` flag:
+- Users can also pass account or bucket values in JSON file (hereinafter referred to as "options JSON file") instead of passing them in CLI as arguments using flags.
```
romayalon, line 474: please mention that it's for create/update operations

docs/non_containerized_NSFS.md
```diff
@@ -467,13 +467,32 @@ sudo node src/cmd/manage_nsfs bucket delete --config_root ../standalon/config_ro
 sudo node src/cmd/manage_nsfs bucket list --config_root ../standalon/config_root 2>/dev/null
 
 ```
+NSFS management CLI command will create both account and bucket dir if it's missing in the config_root path.
```
romayalon, line 470:
```suggestion
NSFS management CLI run will create both accounts, access_keys, and buckets directories if they are missing under the config_root directory.
```

src/manage_nsfs/manage_nsfs_help_utils.js
```diff
@@ -82,6 +76,9 @@ Flags:
 --access_key <string>                                     (optional)                        Set the access key for the account (default is generated)
 --secret_key <string>                                     (optional)                        Set the secret key for the account (default is generated)
 --fs_backend <none | GPFS | CEPH_FS | NFSv4>              (optional)                        Set the filesystem type of new_buckets_path (default config.NSFS_NC_STORAGE_BACKEND)
+--from_file <string>                                      (optional)                        Use details from the JSON file, there is no need to mention all the properties individually in the CLI
```
romayalon, line 79:
```suggestion
--from_file <string>                                      (optional)                        Use details from the JSON file, other account properties will be ignored
```

src/manage_nsfs/manage_nsfs_help_utils.js
```diff
@@ -63,12 +63,6 @@ const GLOBAL_CONFIG_ROOT_ALL_FLAG = `
 --config_root <string>                                    (optional)                        Use configuration files path (default config.NSFS_NC_DEFAULT_CONF_DIR)
 `;
 
-const GLOBAL_CONFIG_FLAGS_ADD_UPDATE_FLAGS = `
```
romayalon, line 66: why removing it?

src/cmd/manage_nsfs.js
```diff
@@ -690,6 +674,56 @@ async function get_config_data(config_file_path, show_secrets = false) {
...
+ * @param {string} action
+ * @param {object} user_input
+ */
+function get_access_keys(action, user_input) {
+    let access_keys = [{
+        access_key: undefined,
+        secret_key: undefined,
+    }];
+    let new_access_key;
+    if (action !== ACTIONS.LIST && action !== ACTIONS.STATUS) _validate_access_keys(user_input.access_key, user_input.secret_key);
```
romayalon, line 716:
```suggestion
    if (action === ACTIONS.ADD || action === ACTIONS.UPDATE || action === ACTIONS.DELETE) _validate_access_keys(user_input.access_key, user_input.secret_key);
```

src/cmd/manage_nsfs.js
```diff
@@ -690,6 +674,56 @@ async function get_config_data(config_file_path, show_secrets = false) {
...
+    if (action === ACTIONS.ADD || action === ACTIONS.STATUS) {
```
romayalon, line 717: You didn't write this function but this whole algorithm of this function is a bit weird, let's try focus on this function on a different PR if needed, I'm bothered about generation/regeneration of credentials flows like this that are integrated with simple fetching from user flows

src/cmd/manage_nsfs.js
```diff
@@ -690,6 +674,56 @@ async function get_config_data(config_file_path, show_secrets = false) {
...
+ * get_options_from_file will read a JSON file that include key-value of the options 
+ * (instead of flags) and return its content
+ * @param {string} type
+ * @param {string} action
+ * @param {string} file_path
+ */
+async function get_options_from_file(type, action, file_path) {
+    const fs_context = native_fs_utils.get_process_fs_context();
+    try {
+        const raw_data = (await nb_native().fs.readFile(fs_context, file_path)).data;
```
romayalon, line 687: can you please use native_fs_utils.read_config_file() and change the function name?

src/cmd/manage_nsfs.js
```diff
@@ -360,57 +365,36 @@ function set_access_keys(argv, generate) {
...
-                uid: argv.user ? undefined : argv.uid,
-                gid: argv.user ? undefined : argv.gid,
-                new_buckets_path: argv.new_buckets_path,
-                fs_backend: argv.fs_backend ? String(argv.fs_backend) : config.NSFS_NC_STORAGE_BACKEND
-            }
-        }, _.isUndefined);
-    }
+async function fetch_account_data(action, user_input) {
+    const { access_keys, new_access_key } = get_access_keys(action, user_input);
+    let data = {
```
romayalon, line 370: don't we need the omitBy(..., _.isUndefined) ?

src/cmd/manage_nsfs.js
```diff
@@ -331,20 +331,25 @@ async function manage_bucket_operations(action, data, argv) {
...
 
-async function account_management(argv, from_file) {
+async function account_management(argv, path_to_json_options) {
+    const type = argv._[0] || '';
     const action = argv._[1] || '';
     const show_secrets = Boolean(argv.show_secrets) || false;
-    const data = await fetch_account_data(argv, from_file);
+    let user_input;
+    // currently we use from_file only in add action
+    if (path_to_json_options) user_input = await get_options_from_file(type, action, path_to_json_options);
```
romayalon, line 340: looks much better!
I think that we should call this function at the same time we call validate_flags_arguments() function and integrate the 2 functions together, so that the validations will be called once

docs/non_containerized_NSFS.md
shirady: added, thanks.

docs/non_containerized_NSFS.md
shirady: I just moved this.
I will change it according to your suggestion, thanks.

src/cmd/manage_nsfs.js
shirady: I'm not sure I'm following... `validate_flags_arguments` is called in the main only once and it validates the command (that we have the type and action, that we don't have extra flags, and the values are according to the types we allowed).

In this PR we validate the command, but we also need to validate the options inside the JSON - and we are reusing the same functions as in `validate_flags_arguments`.

src/cmd/manage_nsfs.js
shirady: Why do you think we need the omitBy(..., _.isUndefined)?
I would guess that someone would think we need this for the schema validation - but before use `JSON.parse` so the undefined are removed.

src/cmd/manage_nsfs.js
shirady: I will change.

src/cmd/manage_nsfs.js
shirady: Changed.

src/cmd/manage_nsfs.js
shirady: Do you want to open an issue?
It will allow us to focus on what might bother you.

src/manage_nsfs/manage_nsfs_help_utils.js
shirady: I was afraid it might be confusing.
I removed the 'from-file' from this, so I copied those flags in the add and update commands.

src/manage_nsfs/manage_nsfs_help_utils.js
shirady: Changed.
We not just ignore, but we throw an error, see for example case 'should fail - cli create account using from_file with additional flags (name)'.

src/cmd/manage_nsfs.js
romayalon: yes, my point here is that we need to validate the arguments at the same time, regardless of their origin (flags/file)
so if flags are checked on the main function, I don't see a reason for checking the arguments here (if they originated from a file)

shirady: I will move it to the main function.

src/cmd/manage_nsfs.js
```diff
@@ -99,11 +99,15 @@ async function main(argv = minimist(process.argv.slice(2))) {
         config_root_backend = argv.config_root_backend ? String(argv.config_root_backend) : config.NSFS_NC_CONFIG_DIR_BACKEND;
 
         await check_and_create_config_dirs();
-        const from_file = argv.from_file ? String(argv.from_file) : '';
+        const path_to_json_options = argv.from_file ? String(argv.from_file) : '';
+        let user_input; // in case of add action can be from argv or from-file
         if (type === TYPES.ACCOUNT) {
-            await account_management(argv, from_file);
+            // currently we use from_file only in add action
```
romayalon, line 105: hmm, it's still not exactly what I meant.. 
I suggest merging between get_options_from_file() and validate_flags_arguments() in order to get a single flow of user input validation

shirady: I don't understand why - the number of properties is the same if we would iterate once (outside the from-file and inside the file):
1. First, we validate the command, for example: `sudo node src/cmd/manage_nsfs account add --from_file <path>` that the command has only the flags that we allowed ('from_file', 'config_root_backend', 'config_root') with the right type.
2. Second, we validate what is inside the file "options JSON file" - all the values that we currently have in `account add`.

shirady: I think that if I will not do it in two steps I will not know the origin of the property (command-line or inside the file).

src/cmd/manage_nsfs.js
romayalon: sure, I wrote it in the tracking sheet

src/cmd/manage_nsfs.js
romayalon: @shirady why do you need to know the origin of the property? 
what I mean is that if the origin of the input is argv or from a file, it doesn't really matter and it would be great to have the validations in a single flow

src/cmd/manage_nsfs.js
romayalon: just verifying with you that it's not needed anymore

src/manage_nsfs/manage_nsfs_help_utils.js
romayalon: aha I understand, maybe just bring back the global properties so it will be re-used?

src/cmd/manage_nsfs.js
shirady: I separated the validation into 2 iterations:
1. On the command: `sudo node src/cmd/manage_nsfs account add --from_file <path>`, here I don't allow all the flags, but only `from_file`, `config_root_backend`, `config_root`). The origin of this check is on argv, if I have all the options extracted - how will I validate this?
2. On the file options - I reuse VALID_OPTIONS_ACCOUNT with the option 'add'.

src/cmd/manage_nsfs.js
shirady: Not needed anymore, but if you think of a potential bug please add the scenario and I will test it.

src/manage_nsfs/manage_nsfs_help_utils.js
shirady: I will bring back the global flags including: `config_root`, and `config_root_backend` (without `from-file`).

src/cmd/manage_nsfs.js
shirady: Resolving, it is a GAP in the tracking sheet.

romayalon (approved):

PR #7781, opened 2024-02-01
Copy rsyslog config file as regular file

piste-jp:
### Explain the changes

Current rsyslog log configuration, `noobaa_syslog.conf`, is done by symlink. But this implementation makes a problem in some systems like stand alone installed RHEL8. So need to install the config file as a regular file.

`noobaa_rsyslog.conf` has many conflict configuration in stand alone installed RHEL8. So it is removed from installation under `/etc/rsyslog.conf.d` and only time stamp config is moved to `noobaa_syslog.conf`.

In addition to this, move the unit file for systemd from `/etc/systemd/system` to `/usr/lib/systemd/system`. Because `/usr/lib/systemd/system` is the directory for additional systemd unit files. Symlinks under `/etc/systemd/system` shall be managed by `systemctl` command after install.

 1. Copy noobaa_syslog.conf as a regular file instead of a symlink
 2. Do not copy noobaa_rsyslog.conf at the rpm install
 3. Move tilestamp format definition from noobaa_rsyslog.conf to noobaa_syslog.conf
 4. Copy the unit file of systemd to /usr/lib/systemd/system


piste-jp-ibm:
@naveenpaul1 ,

Sorry, I might create a PR from different branch... I will close this and open a different oner.

piste-jp-ibm:
I would like to try again in #7789.

Issue #7766, opened 2024-01-25
NSFS | NC | CLI | Deprecate email from CLI


romayalon:
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
1. The following CLI commands require an account email address - 
Account's email address being used on account add/update - 

```
sudo node src/cmd/manage_nsfs account add --name account1 --email account1@noobaa.io --uid 0 --gid 0 --new_buckets_path /tmp/b1
```

```
sudo node src/cmd/manage_nsfs account update --name account1 --email account2@noobaa.io --uid 0 --gid 0 --new_buckets_path /tmp/b1
```

Bucket's owner email address also being used on bucket add/update - 
```
sudo node src/cmd/manage_nsfs bucket add  --name bucket2 --email account1@noobaa.io --path /tmp/b1
````

```
sudo node src/cmd/manage_nsfs bucket update  --name bucket2 --email account2@noobaa.io
```

We would like to remove these references and replace them with account name.
### Expected behavior
1. replace these email references with account's name.
This is the first step towards https://github.com/noobaa/noobaa-core/issues/7734 which is the full solution.

### Steps to reproduce
1. mentioned above.

shirady:
@romayalon, If I understand it right - the fix here is a temporary solution since we wish to use the id.
My plan is:
manage nsfs commands
1. In the account we will remove the flag `--email` (account add, account update)
2. In the bucket we will replace the `--email` with flag `--bucket_owner` which is the account name.

As I see it:
- we would need to remove email property from the account schema.
- It would affect also the bucketspace_fs (system_owner and bucket owner are from the `account.email`).

WDYT? something I need to take into account?

romayalon:
@shirady Yes exactly, it's a temp change and we might need to do more changes when replacing to ids.
we need to make sure that on s3 authentication/bucket policy flows we don't miss anything when we check the requesting account.
Also, we need to check specifically the system_owner cases there.
see for example, I had to replace the email with name for bucket policy -  https://github.com/noobaa/noobaa-core/blob/master/src/endpoint/s3/s3_rest.js#L244

shirady:
Thanks @romayalon,
A couple of open questions I have:
1) As I understand it this is not part of the change:
https://github.com/noobaa/noobaa-core/blob/ed0ba58bdba737a2ba8c47c9b06dfa303e4a0c16/src/cmd/nsfs.js#L287-L291
[...]

PR #7785, opened 2024-02-04
NSFS | NC | Fix Bucket Creation Verification

shirady:
### Explain the changes
When an account not allowed to create a bucket (has `allow_bucket_creation` false) tries to create a bucket, it throws an error:
1. Add this check to the `verify_bucket_owner` function only when the action is to add a bucket.
2. Add a new error for this case.
3. Add a test for this case.

### Issues:
1. Currently we do not check if the account has the property `allow_bucket_creation` before creating a bucket using NSFS manage.
Note: We check it when the user tries to create a bucket using AWS CLI (inside `bucketspace_fs`).
2. GAP - need to add unit test for updating a bucket (see [this comment](https://github.com/noobaa/noobaa-core/pull/7785#discussion_r1477749730) with manual description).

#### Manual test:
1. Create an account without passing the flag `new-buckets-path`: 
`sudo node src/cmd/manage_nsfs account add --name <account-name> --email <account-email> --uid <uid> --<gid>`
2. Create the path for the bucket and give it permission, for example: `mkdir -p /tmp/nsfs_root1/my-bucket; chmod 777 /tmp/nsfs_root1/my-bucket`.
3. Create a bucket (should see the error): `sudo node src/cmd/manage_nsfs bucket add --name <bucket-name> --email <account-email> --path /tmp/nsfs_root1/my-bucket` (`path` with the example mentioned above).

#### Unit tests:
For running the unit test with a new test, please run:
`sudo node ./node_modules/.bin/_mocha src/test/unit_tests/test_nc_nsfs_cli.js`

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -193,7 +193,7 @@ function get_symlink_config_file_path(config_type_path, file_name) {
 
 async function add_bucket(data) {
     await validate_bucket_args(data, ACTIONS.ADD);
-    const account_id = await verify_bucket_owner(data.bucket_owner.unwrap());
+    const account_id = await verify_bucket_owner(data.bucket_owner.unwrap(), ACTIONS.ADD);
     const fs_context = native_fs_utils.get_process_fs_context(config_root_backend);
     const bucket_conf_path = get_config_file_path(buckets_dir_path, data.name);
     const exists = await native_fs_utils.is_path_exists(fs_context, bucket_conf_path);
@@ -210,11 +210,14 @@ async function add_bucket(data) {
 }
 
 /** verify_bucket_owner will check if the bucket_owner has an account
- * in case it finds on, it returns the account id, otherwise it would throw an error
+ * after it finds one, it returns the account id, otherwise it would throw an error
+ * (in case the action is add bucket it also checks that the owner has allow_bucket_creation)
  * @param {string} bucket_owner
+ * @param {string} action
  */
-async function verify_bucket_owner(bucket_owner) {
+async function verify_bucket_owner(bucket_owner, action) {
     let is_bucket_owner_exist = false;
+    let is_allow_bucket_creation = false;
     let account_id;
     const show_secrets = false;
     const fs_context = native_fs_utils.get_process_fs_context();
@@ -227,17 +230,20 @@ async function verify_bucket_owner(bucket_owner) {
             const data = await get_config_data(full_path, show_secrets);
             if (data.email === bucket_owner) {
                 is_bucket_owner_exist = true;
+                is_allow_bucket_creation = data.allow_bucket_creation;
                 account_id = data._id;
             }
         }
     });
 
-    if (is_bucket_owner_exist) {
-        return account_id;
-    } else {
+    if (!is_bucket_owner_exist) {
         const detail_msg = `bucket owner ${bucket_owner} does not exists`;
         throw_cli_error(ManageCLIError.BucketSetForbiddenNoBucketOwner, detail_msg);
     }
+    if (action === ACTIONS.ADD && !is_allow_bucket_creation) {
+            throw_cli_error(ManageCLIError.BucketCreationNotAllowed, bucket_owner);
+    }
+    return account_id;
 }
 
 async function get_bucket_status(data) {
@@ -255,7 +261,7 @@ async function get_bucket_status(data) {
 
 async function update_bucket(data) {
     await validate_bucket_args(data, ACTIONS.UPDATE);
-    await verify_bucket_owner(data.bucket_owner.unwrap());
+    await verify_bucket_owner(data.bucket_owner.unwrap(), ACTIONS.UPDATE);
 
     const fs_context = native_fs_utils.get_process_fs_context(config_root_backend);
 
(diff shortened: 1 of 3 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -227,13 +230,22 @@ async function verify_bucket_owner(bucket_owner) {
...
                 is_bucket_owner_exist = true;
+                is_allow_bucket_creation = data.allow_bucket_creation;
                 account_id = data._id;
             }
         }
     });
 
     if (is_bucket_owner_exist) {
-        return account_id;
+        if (action === ACTIONS.ADD) {
```
romayalon, line 240: I find that the following approach is better understood - 
```
if (!is_bucket_owner_exist) {
        throw_cli_error(ManageCLIError.BucketSetForbiddenNoBucketOwner, `bucket owner ${bucket_owner} does 
        not exists`);
}
if (action === ACTIONS.ADD && !is_allow_bucket_creation) {
        throw_cli_error(ManageCLIError.BucketCreationMotAllowed, bucket_owner);
}
return account_id;
```

src/manage_nsfs/manage_nsfs_cli_errors.js
```diff
@@ -275,6 +275,12 @@ ManageCLIError.BucketSetForbiddenNoBucketOwner = Object.freeze({
     http_code: 403,
 });
 
+ManageCLIError.BucketCreationMotAllowed = Object.freeze({
```
romayalon, line 278:
```suggestion
ManageCLIError.BucketCreationNotAllowed = Object.freeze({
```

src/test/unit_tests/test_nc_nsfs_cli.js
```diff
@@ -124,6 +124,39 @@ mocha.describe('manage_nsfs cli', function() {
             await exec_manage_cli(nc_nsfs_manage_entity_types.ACCOUNT, action, account_options2);
         });
 
+        mocha.it('cli bucket create - should fail bucket owner does not have allow_bucket_creation', async function() {
```
romayalon, line 127: worth adding an update flow tests to see we don't fail there

naveenpaul1, line 127:
```suggestion
        mocha.it('cli bucket create - should fail bucket owner's allow_bucket_creation is false', async function() {
```

shirady: What is the flow that you thought of?
- Since the account cannot create buckets, it should not have buckets to update (it cannot update a bucket that he doesn't own).
- I did a manual test (which is not just using the manage nsfs commands):
1. I created an account with `new_buckets_path`
2. I created a bucket
3. I manually changed the configuration file of the account (see my comment below): I removed the `new_buckets_path` and set the `allow_bucket_creation` to false.
4. I updated the bucket I already created (step 2) `--fs_backend GPFS` instead of the default successfully.

Currently, we only check the value of `allow_bucket_creation` only when the user tries to create a new bucket (both manage nsfs and bucketspace_fs).
1) I think that we should check it for every action that the user does on a bucket - add/update/delete and a user with `allow_bucket_creation` false will be able only to list and status (and from  bucketspace_fs to use GET ops).
Unfortunately, we cannot implement it now because in manage nsfs actions on a bucket are not related to a specific account (like we have the `requesting_account` in bucketspace_fs).

2) Today the value of `allow_bucket_creation` is inferred from the flag `new_buckets_path` (if we have it then it is true, and vice versa), but we do not have the option to unset `new_buckets_path` (like we have for `fs_backend` or `bucket_policy`).

src/cmd/manage_nsfs.js
```diff
@@ -227,17 +230,20 @@ async function verify_bucket_owner(bucket_owner) {
             const data = await get_config_data(full_path, show_secrets);
             if (data.email === bucket_owner) {
                 is_bucket_owner_exist = true;
+                is_allow_bucket_creation = data.allow_bucket_creation;
```
naveenpaul1, line 233: @shirady can we use `break` here to exit the loop?

src/cmd/manage_nsfs.js
shirady: I left the variable `const detail_msg` because I thought it would be easier to understand than directly put it as an argument in the function.

src/cmd/manage_nsfs.js
shirady: No, you cannot use a `break` in a map.

> Array .map and .forEach are designed to never be stopped. Therefore, you cannot break out of the loop when using those methods.

([link](https://allaboutcode.medium.com/javascript-how-to-break-statement-in-an-array-map-method-9861b39957a4))

We are using it as a temporary solution to go over the elements with concurrency.
This implementation will be changed once we use account name (it will be with a lookup on the accounts names) - please see issue #7766.

src/test/unit_tests/test_nc_nsfs_cli.js
shirady: I will change, thanks.

shirady: Summarizing:
1. The administrator is the role that uses manage nsfs, hence we throw an error only when trying to create a bucket for an account with `allow_bucket_creation` false. The rest of the action (delete, update, etc.) can be performed on the bucket since the administrator is the role that uses the system.
2. A unit test for the described scenario will be added after we can modify the `allow_bucket_creation` to false using manage_nsfs (either directly or by unsetting the value of `new_buckets_path`.

romayalon (approved):

PR #7788, opened 2024-02-04
NSFS | NC | CLI | Account add/update when uid = 0 and gid != 0 bug fix

### LINKED ISSUE #7787: NSFS | NC | CLI | Account add/update is failing when uid = 0 and gid != 0 and vice versa
### Environment info
- NooBaa Version: Master
- Platform: NC

### Actual behavior
1. Account add/update is failing when uid = 0 but and gid != 0 and vice versa

### Expected behavior
1. An account can get created/updated  when uid = 0 but and gid != 0 and vice versa

### Steps to reproduce
```
> manage_nsfs.js account add --name foo --email foo@example.com --uid 0 --gid 10 2>/dev/null
{
  "error": {
    "code": "MissingAccountNSFSConfigUID",
    "message": "Account config should include UID",
    "detail": {
      "uid": 0,
      "gid": 10,
      "new_buckets_path": "/gpfs/bucket-root"
    }
  }
}
```

romayalon:
### Explain the changes
1. is_undefined() checks using if (!value), which return true when uid/gid is 0, causing the account add/update to fail.

1. Fixed https://github.com/noobaa/noobaa-core/issues/7787

### Testing Instructions:
1. `manage_nsfs.js account add --name foo --email foo@example.com --uid 0 --gid 10 2>/dev/null`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -424,7 +424,7 @@ async function fetch_account_data(argv, from_file) {
             // fs_backend deletion specified with empty string '' (but it is not part of the schema)
             fs_backend: data.nsfs_account_config.fs_backend || undefined
         },
-        allow_bucket_creation: !is_undefined(data.nsfs_account_config.new_buckets_path),
+        allow_bucket_creation: !is_string_undefined(data.nsfs_account_config.new_buckets_path),
         // updates of account identifiers
         new_name: data.new_name && new SensitiveString(String(data.new_name)),
         new_access_key: data.new_access_key && new SensitiveString(String(data.new_access_key))
@@ -442,7 +442,7 @@ async function fetch_existing_account_data(target) {
     } catch (err) {
         dbg.log1('NSFS Manage command: Could not find account', target, err);
         if (err.code === 'ENOENT') {
-            if (is_undefined(target.name)) {
+            if (is_string_undefined(target.name)) {
                 throw_cli_error(ManageCLIError.NoSuchAccountAccessKey, target.access_keys[0].access_key);
             } else {
                 throw_cli_error(ManageCLIError.NoSuchAccountName, target.name);
@@ -576,13 +576,13 @@ async function get_account_status(data, show_secrets) {
     await validate_account_args(data, ACTIONS.STATUS);
 
     try {
-        const account_path = is_undefined(data.name) ?
+        const account_path = is_string_undefined(data.name) ?
             get_symlink_config_file_path(access_keys_dir_path, data.access_keys[0].access_key) :
             get_config_file_path(accounts_dir_path, data.name);
         const config_data = await get_config_data(account_path, show_secrets);
         write_stdout_response(ManageCLIResponse.AccountStatus, config_data);
     } catch (err) {
-        if (is_undefined(data.name)) {
+        if (is_string_undefined(data.name)) {
             throw_cli_error(ManageCLIError.NoSuchAccountAccessKey, data.access_keys[0].access_key.unwrap());
         } else {
             throw_cli_error(ManageCLIError.NoSuchAccountName, data.name.unwrap());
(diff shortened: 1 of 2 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -774,10 +774,10 @@ async function validate_account_args(data, action) {
 
         if (is_undefined(data.access_keys[0].secret_key)) throw_cli_error(ManageCLIError.MissingAccountSecretKeyFlag);
         if (is_undefined(data.access_keys[0].access_key)) throw_cli_error(ManageCLIError.MissingAccountAccessKeyFlag);
-        if (data.nsfs_account_config.gid && (is_undefined(data.nsfs_account_config.uid))) {
```
shirady, line 777: In addition to your change, I think either the function name `is_undefined` or the implementation of the function should be changed. 
- Function name change - The current implementation doesn't check if the value is undefined but if the value is Falsy (here is a [link](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)), and 0 is Falsy.
- Change function implementation - to check if the value is_undefined explicitly (includes the SensitiveString case).
I would go on the second option.

src/test/unit_tests/test_nc_nsfs_cli.js
```diff
@@ -530,6 +531,33 @@ mocha.describe('manage_nsfs cli', function() {
             }
         });
 
+        mocha.it('cli account add - uid=0, gid!=0', async function() {
+            const action = ACTIONS.ADD;
+            const account_name = 'uid_is_0';
+            const options = { name: account_name, email: account_name, uid: 0, gid: 1001 };
```
shirady, line 537: So it won't confuse I would add a variable for the account email, for example:
```suggestion
            const account_name = 'uid_is_0';
            const account_email = `${account_name}@noobaa.com`
            const options = { name: account_name, email: account_email, uid: 0, gid: 1001 };
```

src/test/unit_tests/test_nc_nsfs_cli.js
```diff
@@ -530,6 +531,33 @@ mocha.describe('manage_nsfs cli', function() {
             }
         });
 
+        mocha.it('cli account add - uid=0, gid!=0', async function() {
```
shirady, line 534: I would change the description so it will be clear for others... (other tests as well).
```suggestion
        mocha.it('cli account add - uid is 0, gid is not 0', async function() {
```

shirady:
@romayalon please notice there is a similar issue that is not covered by this fix:
`sudo node src/cmd/manage_nsfs account add --name foo --email foo@example.com --user 0` (user name is 0).

```json
{
  "error": {
    "code": "InvalidAccountNSFSConfig",
    "message": "Account config should not be empty, should contain UID, GID or user",
    "detail": {
      "distinguished_name": 0
    }
  }
}
```

src/test/unit_tests/test_nc_nsfs_cli.js
naveenpaul1, line 534: I think it would be better if you add this in jest test `test_nc_nsfs_account_cli.test.js`

romayalon:
@shirady As discussed on Slack - this is a different type of issue that will addressed by replacing the STRING_OR_NUMBER type with string only.

> @romayalon please notice there is a similar issue that is not covered by this fix: `sudo node src/cmd/manage_nsfs account add --name foo --email foo@example.com --user 0` (user name is 0).
> 
> ```json
> {
>   "error": {
>     "code": "InvalidAccountNSFSConfig",
>     "message": "Account config should not be empty, should contain UID, GID or user",
>     "detail": {
>       "distinguished_name": 0
>     }
>   }
> }
> ```

shirady (approved):
You can leave the comment unhandled and mentioned it as a GAP (the fix is more important)

PR #7791, opened 2024-02-05
NSFS | NC | Fix Check Flag Type From Manage NSFS CLI

### LINKED ISSUE #7790: Invalid argument type error details are not clear
### Environment info
- NooBaa Version: standalone noobaa
- Platform: noobaa-core-5.15.0-20240131.el9.x86_64

### Actual behavior
[root@ramyamc-12 noobaa-core]# node /usr/local/noobaa-core/src/cmd/manage_nsfs account add --name 12 --email 2>/dev/null
{
  "error": {
    "code": "InvalidArgumentType",
    "message": "Invalid argument type",
    "detail": "type of option email should be string, received: true (type boolean)"
  }
}
[root@ramyamc-12 noobaa-core]# node /usr/local/noobaa-core/src/cmd/manage_nsfs account add --name 2>/dev/null
{
  "error": {
    "code": "InvalidArgumentType",
    "message": "Invalid argument type",
    "detail": "type of option name should be string, number, received: true (type boolean)"
  }
}
[root@ramyamc-12 noobaa-core]# 

### Expected behavior
1. "type of option name should be string, number, received: true (type boolean)"
This should be rephrased  it confusing to the customer

### Steps to reproduce
1.

shirady:
### Explain the changes
1. Remove the string or number and only handle names and allow it to be a number (there were part of #7756).
2. Check that the argument is not empty directly with argv array (and not object my minimist).
3. Rename the function from "options' to "flags" to be aligned with customer terms when referring to the raw input from the user.
4. Remove option validation from from-file (will be a GAP).

### Issues: Fixed #7790
1. In the PR we will change the message when the user passes a flag that is not boolean without a value.
2. GAP - I removed the validation in the part of "from-file", which should be handled and tested specifically.

### Testing Instructions:
1. For manual tests add a valid flag without value (except boolean flags), for example:
`sudo node src/cmd/manage_nsfs account add --name` (`--name` with no value).
2. For running the unit test with the new tests, please run:
`sudo npx jest test_nc_nsfs_account_cli.test.js`.

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -19,7 +19,7 @@ const ManageCLIResponse = require('../manage_nsfs/manage_nsfs_cli_responses').Ma
 const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
 const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
 const { print_usage } = require('../manage_nsfs/manage_nsfs_help_utils');
-const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, TYPE_STRING_OR_NUMBER,
+const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE,
     LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS} = require('../manage_nsfs/manage_nsfs_constants');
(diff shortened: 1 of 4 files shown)


romayalon (approved):
LGTM

PR #7799, opened 2024-02-06
NSFS | NC | S3 Flow | Fix list buckets creation date

### LINKED ISSUE #7795: List buckets returns CreationDate as current date instead of actual bucket create time
### Environment info
- NooBaa Version: **5.14**
- Platform: **NA**

### Actual behavior
1. We always use now for the list buckets reply -

https://github.com/noobaa/noobaa-core/blob/b04f5ab2f17b72ded93b17e8b9f431c66a8c5308/src/endpoint/s3/ops/s3_get_service.js#L11-L18

### Expected behavior
1. Should return the specific creation date of each bucket.

### Steps to reproduce
1. NA

romayalon:
### Explain the changes
1. BucketspaceFS.list_buckets() - Called read_bucket_sdk_config_info() instead of read_bucket_sdk_namespace_info() to get creation_date of the bucket. If the requesting account can access the bucket, return it together with the bucket name.
2. s3_get_service() - Returned bucket.creation_date in the correct date format if it exists, else return the fixed date.

1. Fixed https://github.com/noobaa/noobaa-core/issues/7795
2. Gaps - 
3. BucketspaceFs - 2 places where we can use concurrency for performance improvement - will add it in a follow-up PR.
4. Added read_config_file() helper function that can be re-used at many places in our code, will add it in a follow-up refactoring PR.
5. Refactor delete bucket 
### Testing Instructions:
1.  sudo node mocha src/test/unit_tests/test_bucketspace_fs.js
2. Manually - create a few buckets and list buckets and check their creation_date.

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_get_service.js
+++ b/src/endpoint/s3/ops/s3_get_service.js
@@ -15,7 +15,7 @@ async function list_buckets(req) {
             Buckets: reply.buckets.map(bucket => ({
                 Bucket: {
                     Name: bucket.name.unwrap(),
-                    CreationDate: date,
+                    CreationDate: bucket.creation_date ? s3_utils.format_s3_xml_date(bucket.creation_date) : date,
                 }
             }))
         }
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -217,22 +217,25 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             }
             throw this._translate_bucket_error_codes(err);
         }
+        // TODO - replace filter and map to map with concurrency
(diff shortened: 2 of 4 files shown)


naveenpaul1 (approved):

src/endpoint/s3/ops/s3_get_service.js
```diff
@@ -15,7 +15,7 @@ async function list_buckets(req) {
             Buckets: reply.buckets.map(bucket => ({
                 Bucket: {
                     Name: bucket.name.unwrap(),
-                    CreationDate: date,
+                    CreationDate: s3_utils.format_s3_xml_date(bucket.creation_date) || date,
```
guymguym, line 18: does format_s3_xml_date returns undefined if the passed date is undefined? otherwise this fallback || will not workout...

guymguym (approved):

PR #7801, opened 2024-02-07
NSFS | NC | Remove Email from NSFS Account and Bucket in manage_nsfs (CLI)

### LINKED ISSUE #7766: NSFS | NC | CLI | Deprecate email from CLI

shirady:
### Explain the changes
1. Remove email from NSFS account and bucket mainly in manage CLI commands:
In account command - remove it
In bucket command replace it with the flag `--owner` and set the account name in it.
3. Remove tests related to email, and update existing tests.
4. Remove the email mentioned in the documentation related to manage NSFS.
5. Increase timeouts in `test_s3_encryption.js` (not related to this PR, but could not pass the CI).

### Issues: Fixed #7766 
1. Currently we use email as a flag in account add/update and bucket add/update, this code change is for fixing the issue (it is a temporary solution towards using only the account ID).

List of GAPs I'm aware of:
1. GAP - We would still have the email in the schema and it will be the output of json account (opened issue #7811).
2. GAP - After this change, we can improve the performance of `verify_bucket_owner` and lookup the `bucket_owner` has a config file (moved to use the account name).
3. GAP - in many places, I set the value of `system_owner` as the account name (there is open issue #7793).
4. GAP - in the documentation of NSFS NC we need to explain the account creation before bucket creation.

### Testing Instructions:
1. Manual test - now do not use `--flag` email, for example:
- First, create the `FS_ROOT` and a directory for a bucket: `mkdir -p /tmp/nsfs_root1/my-bucket` and give permissions `chmod 777 /tmp/nsfs_root1/my-bucket`.
- account add: `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path /tmp/nsfs_root1 --uid <uid> --gid <gid>` (no `--email` flag).
- bucket add: `sudo node src/cmd/manage_nsfs bucket add --name <bucket-name> --owner <account-name> --path /tmp/nsfs_root1/my-bucket`
2. For running tests that were updated, please run:
- `sudo npx jest test_nc_nsfs_account_cli.test.js`
- `sudo npx jest test_nc_nsfs_bucket_cli.test.js`
- `sudo node ./node_modules/.bin/_mocha src/test/unit_tests/test_nc_nsfs_cli.js`
- `npx jest test_nc_nsfs_account_schema_validation.test.js`
- `npx jest test_nc_nsfs_bucket_schema_validation.test.js`
3. For checking the NSFS NC as it run in the CI: `make run-nc-tests CONTAINER_PLATFORM=linux/arm64` (the env `CONTAINER_PLATFORM=linux/arm64` is used because I have Mac). If you wish to run only specific tests, for example:
- `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace_fs.js`
Note: run it without the NSFS server working (ctrl + C to sudo node src/cmd/nsfs).
- `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace.js` 

- [X] Doc updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -140,8 +140,8 @@ async function fetch_bucket_data(argv, from_file) {
             _id: undefined,
             name: argv.name,
             owner_account: undefined,
-            system_owner: argv.email,
-            bucket_owner: argv.email,
+            system_owner: argv.owner, // GAP - needs to be the system_owner (currently it is the account name)
+            bucket_owner: argv.owner,
             wide: argv.wide,
             tag: undefined, // if we would add the option to tag a bucket using CLI, this should be changed
             versioning: action === ACTIONS.ADD ? 'DISABLED' : undefined,
@@ -223,7 +223,7 @@ async function add_bucket(data) {
 /** verify_bucket_owner will check if the bucket_owner has an account
  * after it finds one, it returns the account id, otherwise it would throw an error
  * (in case the action is add bucket it also checks that the owner has allow_bucket_creation)
- * @param {string} bucket_owner
+ * @param {string} bucket_owner account name
  * @param {string} action
  */
 async function verify_bucket_owner(bucket_owner, action) {
@@ -239,7 +239,7 @@ async function verify_bucket_owner(bucket_owner, action) {
         if (entry.name.endsWith('.json') && !is_bucket_owner_exist) {
             const full_path = path.join(accounts_dir_path, entry.name);
             const data = await get_config_data(full_path, show_secrets);
-            if (data.email === bucket_owner) {
+            if (data.name === bucket_owner) {
                 is_bucket_owner_exist = true;
                 is_allow_bucket_creation = data.allow_bucket_creation;
                 account_id = data._id;
@@ -398,7 +398,6 @@ async function fetch_account_data(argv, from_file) {
     if (!data) {
         data = _.omitBy({
             name: argv.name,
-            email: argv.email,
             creation_date: action === ACTIONS.ADD ? new Date().toISOString() : undefined,
             wide: argv.wide,
             new_name: argv.new_name,
@@ -421,7 +420,7 @@ async function fetch_account_data(argv, from_file) {
     data = {
         ...data,
         name: new SensitiveString(String(data.name)),
-        email: new SensitiveString(String(data.email)),
+        email: new SensitiveString(String(data.name)), // temp, keep the email internally
         access_keys: [{
             access_key: new SensitiveString(String(data.access_keys[0].access_key)),
             secret_key: new SensitiveString(String(data.access_keys[0].secret_key)),
@@ -516,6 +515,7 @@ async function update_account(data) {
     }
     const data_name = data.new_name || cur_name;
     data.name = data_name;
+    data.email = data_name; // saved internally
     data.access_keys[0].access_key = data.new_access_key || cur_access_key;
     const cur_account_config_path = get_config_file_path(accounts_dir_path, cur_name.unwrap());
     const new_account_config_path = get_config_file_path(accounts_dir_path, data.name.unwrap());
@@ -548,7 +548,7 @@ async function update_account(data) {
 
 async function delete_account(data) {
     await validate_account_args(data, ACTIONS.DELETE);
-    await verify_delete_account(data.name.unwrap(), data.email.unwrap());
+    await verify_delete_account(data.name.unwrap());
 
     const fs_context = native_fs_utils.get_process_fs_context(config_root_backend);
     const account_config_path = get_config_file_path(accounts_dir_path, data.name);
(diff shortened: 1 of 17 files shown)


src/server/system_services/schemas/nsfs_account_schema.js
```diff
@@ -7,7 +7,7 @@ module.exports = {
     required: [
         '_id',
         'name',
-        'email',
+        'email', // temp, keep the email internally
```
romayalon, line 10: why do we still need to keep it internally?

src/test/unit_tests/test_nc_nsfs_cli.js
```diff
@@ -315,17 +309,6 @@ mocha.describe('manage_nsfs cli', function() {
             }
         });
 
-        mocha.it('cli bucket update owner email', async function() {
```
romayalon, line 318: can you keep it and just replace the email with owner? 
I see I added here a validation for the creation date not getting changed on update

shirady: Sure, thanks, I changed it to owner update instead of email.
If you want we can create a test for the validation of the creation date not getting changed on update so it would be clear.

src/server/system_services/schemas/nsfs_account_schema.js
shirady: I had issues with tests `test_bucketspace_fs.js`, `test_bucketspace.js`, and `NSFS Ceph S3 tests` when I removed the email from the account totally, so I decided to remove the email in steps. Since the schedule is tight I published the PR with the first step which is removing the email flag from the end user, and keeping it internally. I will try it again.

romayalon: alright, what was the issue you saw?

shirady: I fixed it, you can see that the tests are passing in the CI (the fix is in 2 separate commits, hopefully for easier code review).

src/endpoint/s3/s3_rest.js
```diff
@@ -226,22 +226,23 @@ async function authorize_request_policy(req) {
     }
 
     const account = req.object_sdk.requesting_account;
-    const is_system_owner = account.email.unwrap() === system_owner.unwrap();
+    // in NSFS NC we don't use email as account identifier
+    const account_identifier = req.object_sdk.nsfs_config_root ? account.name.unwrap() : account.email.unwrap();
```
romayalon, line 230: I suspect that there are more places we will need to do that, for example, what about has_bucket_action_permission()?

shirady: Suspicion is good, we need to plan this change better.

In your example:
https://github.com/noobaa/noobaa-core/blob/32b799e778db9df2bff1ffeaab628d8442c11426/src/server/common_services/auth_server.js#L673-L698

There are a couple of lines that I don't understand how will they work with current configurations:
`bucket.system.owner.email.unwrap()` - in our NSFS NC bucket we don't have a system reference, but `system_owner` is a string, same with `bucket.owner_account.email.unwrap()` we don't have email as a property for the owner_account it is a string.

shirady: As discussed, I will remove the last 2 commits - the removal of email in schema and internal functions in noobaa will be a gap. In this PR I will only remove the email as a flag that I receive from the user.

romayalon (approved):

src/cmd/manage_nsfs.js
```diff
@@ -212,7 +212,7 @@ async function add_bucket(data) {
 /** verify_bucket_owner will check if the bucket_owner has an account
  * after it finds one, it returns the account id, otherwise it would throw an error
  * (in case the action is add bucket it also checks that the owner has allow_bucket_creation)
- * @param {string} bucket_owner
+ * @param {string} bucket_owner account name
```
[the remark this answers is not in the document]
shirady: @naveenpaul1, it was not an accident, I added it to describe the variable.
But if it seems like an accident, it should be removed or refactored.

PR #7802, opened 2024-02-07
add DB clean for activitylogs table

liranmauda:
### Explain the changes
- add DB clean for activitylogs table.

- [ ] Tests added


liranmauda:
@dannyzaken and I talked and we think that we should not add db cleaner, as it can be a bottleneck for the DB.
we should not log object action in the activity logs
closing in favor of https://github.com/noobaa/noobaa-core/pull/7804

PR #7808, opened 2024-02-08
NSFS | NC | Update Help `access_key` as Identifier of Account (Delete, Update)

shirady:
### Explain the changes
1. Update help printing of account update and delete.
2. Update the `VALID_OPTIONS_ACCOUNT` and remove the `access_key` from account delete.
3. Add a test to verify that we cannot use the flag `access_key` in account delete anymore.

### Issues:
1. Currently we see an unrelated error (`MissingAccountSecretKeyFlag`) when trying to delete an account using access_key` as the Identifier of the account.

### Testing Instructions:
1. For help updates run: `sudo node src/cmd/manage_nsfs account <action> --help 2>/dev/null` (replace `<action>` with update and then with delete).
2. For delete account:  (using an invalid option `--access_key` for account delete): `sudo node src/cmd/manage_nsfs account delete --access_key <access-key>`.

- [ ] Doc added/updated
- [X] Tests added


romayalon (approved):
LGTM

PR #7810, opened 2024-02-11
 NSFS | NC | Fix Issue with Flags Separator

shirady:
### Explain the changes
1. Fix the issue with the flag separator - remove the argv array manipulation.
2. Copy two existing tests and change their command line so they will be run with '=' separator instead of ' ' (space): one that passes and one that fails.

### Issues: none
1. In PR #7791 added argv manipulation for creating an error message of an empty flag, but working with that array cannot be directly used in case the separator between the flags is '=' (the flag and the argument will be the same element and not two elements) and it would result with `InternalError`.

### Testing Instructions:
1. Run any command that has an invalid argument with '=' separator, for example: `sudo node src/cmd/manage_nsfs account update --name=noobaa9 --new_buckets_path` (after name you can see the'=', the command invalid because `new_buckets_path` should be passed with <path>, a string value). 

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -882,41 +882,21 @@ function validate_no_extra_options(type, action, input_options) {
 }
 /**
  * validate_options_type_by_value check the type of the value that match what we expect.
- * in case that type of value we have is boolean (and the option type is not boolean) -
- * we check if we have a value (non empty) with array of argument (without using minimist)
  * @param {string} type
  * @param {object} input_options_with_data object with flag (key) and value
  */
 function validate_options_type_by_value(type, input_options_with_data) {
-    const argv_arr = process.argv.slice(2); // we ignore argv[0] (path to node) and argv[1] (path to app)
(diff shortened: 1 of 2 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -882,41 +882,21 @@ function validate_no_extra_options(type, action, input_options) {
 }
 /**
  * validate_options_type_by_value check the type of the value that match what we expect.
- * in case that type of value we have is boolean (and the option type is not boolean) -
```
romayalon, line 885: do we have this check somewhere else?

shirady: Not sure I understand the question...
The check for "type of value we have is boolean (and the option type is not boolean)" is part of the check of invalid type, I tried to separate this case to create a different error message (but it resulted in the issue I'm trying to fix now...), so I decided to drop it and send the same general error message in the details.

romayalon (approved):
LGTM

PR #7812, opened 2024-02-11
NSFS | NC | list buckets missing policy check & performance improvement

### LINKED ISSUE #7809: NSFS | Bucket policy authorization is skipped on List Buckets operation
### Environment info
- NooBaa Version: master
- Platform: ALL

### Actual behavior
1. Currently, we do not authorize bucket policy on list buckets - 
```
    console.log('req.params: ', req.params); // on s3 ls -> noobaa printed - { bucket: '', key: '' }
    if (!req.params.bucket) return; // we return here
```
This gap causes an issue when having multiple accounts, account1 running s3 ls can see the buckets of account2 and vice versa.
On other bucketspaces we check again on the list buckets operation for every bucket if it is allowed to be listed.

for example in bucket server - 
```
    const buckets_by_name = _.filter(
        req.system.buckets_by_name,
        async bucket => await req.has_s3_bucket_permission(bucket, "s3:ListBucket", req) && !bucket.deleting
    );
```

### Expected behavior
1. account1 running s3 ls can not see the buckets of account2 and vice versa.

### Steps to reproduce
1. create account1
2. create account2
3. create bucket1 using account1
4. create bucket2 using bucket1
5. account1 s3 ls // expect only bucket1
6. account2 s3 ls // expect only bucket2

romayalon:
### Explain the changes
1. bucketspaceFS - list_buckets - Added bucket policy check so we won't let account1 to list buckets of account2 if it's not allowed by a bucket policy even if it's allowed by fs (matching uid gid).
2. bucketspaceFS - list_buckets - refactored this function, we iterated the buckets array 4 times, now we will iterate it 2 times -
a. First iteration - map_with_concurrency with concurrency rate of 10 entries at the same time.
If entry is dir file/entry is non json / bucket is unaccessible by bucket policy / underlying storage of the bucket is unaccessible by fs we will return undefined, else - the bucket config info.
b. Second iteration - filter entries - filter will remove undefined values produced by the map_with_concurrency(). 

1. Fixed - https://github.com/noobaa/noobaa-core/issues/7809
2. Gap - Move the following 3 functions - has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir() so they can be re-used - follow up PR.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -32,12 +32,14 @@ const bucket_semaphore = new KeysSemaphore(1);
 
 /**
  * @param {nb.ObjectSDK} object_sdk 
+ * @param {string} [fs_backend]
  * @returns {nb.NativeFSContext}
  */
-function prepare_fs_context(object_sdk) {
+function prepare_fs_context(object_sdk, fs_backend) {
     const fs_context = object_sdk?.requesting_account?.nsfs_account_config;
     if (!fs_context) throw new RpcError('UNAUTHORIZED', 'nsfs_account_config is missing');
     fs_context.warn_threshold_ms = config.NSFS_WARN_THRESHOLD_MS;
+    fs_context.backend = fs_backend;
     // TODO: 
     //if (this.stats) fs_context.report_fs_stats = this.stats.update_fs_stats;
     return fs_context;
@@ -192,6 +194,18 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
     ////////////
 
     //TODO: we need to add pagination support to list buckets for more than 1000 buckets.
+    /**
+     * list_buckets will read all bucket config files, and filter them according to the requesting account's 
+     * permissions
+     * a. First iteration - map_with_concurrency with concurrency rate of 10 entries at the same time.
+     * a.1. if entry is dir file/entry is non json - we will return undefined
+     * a.2. if bucket is unaccessible by bucket policy - we will return undefined
+     * a.3. if underlying storage of the bucket is unaccessible by the account's uid/gid - we will return undefined
+     * a.4. else - return the bucket config info.
+     * b. Second iteration - filter empty entries - filter will remove undefined values produced by the map_with_concurrency().
+     * @param {nb.ObjectSDK} object_sdk
+     * @returns {Promise<object>}
+     */
     async list_buckets(object_sdk) {
         let entries;
         try {
@@ -203,55 +217,27 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             }
             throw this._translate_bucket_error_codes(err);
         }
-        // TODO - replace filter and map to map with concurrency
-        const bucket_config_files = entries.filter(entree => !native_fs_utils.isDirectory(entree) && entree.name.endsWith('.json'));
-        const bucket_names = bucket_config_files.map(bucket_config_file => this.get_bucket_name(bucket_config_file.name));
-        return this.validate_bucket_access(bucket_names, object_sdk);
-    }
 
-    async validate_bucket_access(buckets, object_sdk) {
-        // TODO - replace map & filter to map with concurrency
-        const has_access_buckets = (await P.all(_.map(
-            buckets,
-            async bucket => {
-                dbg.log1('bucketspace_fs.validate_bucket_access:', bucket.name.unwrap());
-                const bucket_config_info = await object_sdk.read_bucket_sdk_config_info(bucket.name.unwrap());
-                const ns = bucket_config_info.namespace;
-                const is_nsfs_bucket = object_sdk.is_nsfs_bucket(ns);
-                const accessible = is_nsfs_bucket ? await this._has_access_to_nsfs_dir(ns, object_sdk) : false;
-                dbg.log1('bucketspace_fs.validate_bucket_access:', bucket.name.unwrap(), is_nsfs_bucket, accessible);
-                return accessible && { ...bucket, creation_date: bucket_config_info.creation_date };
-            }))).filter(bucket_info => bucket_info);
-            return { buckets: has_access_buckets };
-    }
-
-    async _has_access_to_nsfs_dir(ns, object_sdk) {
         const account = object_sdk.requesting_account;
-        dbg.log0('_has_access_to_nsfs_dir: nsr: ', ns, 'account.nsfs_account_config: ', account && account.nsfs_account_config);
-        // nsfs bucket
-        if (!account || !account.nsfs_account_config || _.isUndefined(account.nsfs_account_config.uid) ||
-            _.isUndefined(account.nsfs_account_config.gid)) return false;
-        try {
-            dbg.log0('_has_access_to_nsfs_dir: checking access:', ns.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
-            await nb_native().fs.checkAccess({
-                uid: account.nsfs_account_config.uid,
-                gid: account.nsfs_account_config.gid,
-                backend: ns.write_resource.resource.fs_backend,
-                warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
-            }, path.join(ns.write_resource.resource.fs_root_path, ns.write_resource.path || ''));
-
-            return true;
-        } catch (err) {
-            dbg.log0('_has_access_to_nsfs_dir: failed', err);
-            if (err.code === 'ENOENT' || err.code === 'EACCES' || (err.code === 'EPERM' && err.message === 'Operation not permitted')) return false;
-            throw err;
-        }
+        const buckets = await P.map_with_concurrency(10, entries, async entry => {
+            if (native_fs_utils.isDirectory(entry) || !entry.name.endsWith('.json')) {
+                return;
+            }
+            const bucket_name = this.get_bucket_name(entry.name);
+            const bucket = await object_sdk.read_bucket_sdk_config_info(bucket_name);
+            const bucket_policy_accessible = await this.has_bucket_action_permission(bucket, account, 's3:ListBucket');
+            if (!bucket_policy_accessible) return;
+            const fs_accessible = await this.validate_fs_bucket_access(bucket, object_sdk);
+            if (!fs_accessible) return;
+            return bucket;
+        });
+        return { buckets: buckets.filter(bucket => bucket) };
     }
 
     get_bucket_name(bucket_config_file_name) {
         let bucket_name = path.basename(bucket_config_file_name);
         bucket_name = bucket_name.slice(0, bucket_name.indexOf('.json'));
-        return { name: new SensitiveString(bucket_name) };
+        return bucket_name;
     }
 
     async read_bucket(params) {
@@ -650,6 +636,68 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
 
     async put_object_lock_configuration(params, object_sdk) {
         // TODO
+    }
+
+    /////////////////
+    ///// UTILS /////
+    /////////////////
+
+    // TODO: move the following 3 functions - has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
+    // so they can be re-used
+    async has_bucket_action_permission(bucket, account, action, bucket_path = "") {
+        const account_identifier = account.name.unwrap();
+        dbg.log1('has_bucket_action_permission:', bucket.name.unwrap(), account_identifier, bucket.bucket_owner.unwrap());
+
+        const is_system_owner = account_identifier === bucket.system_owner.unwrap();
+
+        // If the system owner account wants to access the bucket, allow it
+        if (is_system_owner) return true;
+        const is_owner = (bucket.bucket_owner.unwrap() === account_identifier);
+        const bucket_policy = bucket.s3_policy;
+
+        if (!bucket_policy) return is_owner;
+        if (!action) {
+            throw new Error('has_bucket_action_permission: action is required');
+        }
+
+        const result = await bucket_policy_utils.has_bucket_policy_permission(
+            bucket_policy,
+            account_identifier,
+            action,
+            `arn:aws:s3:::${bucket.name.unwrap()}${bucket_path}`,
+            undefined
+        );
+
+        if (result === 'DENY') return false;
+        return is_owner || result === 'ALLOW';
+    }
+
+    async validate_fs_bucket_access(bucket, object_sdk) {
+        dbg.log1('bucketspace_fs.validate_fs_bucket_access:', bucket.name.unwrap());
+        const ns = bucket.namespace;
+        const is_nsfs_bucket = object_sdk.is_nsfs_bucket(ns);
+        const accessible = is_nsfs_bucket ? await this._has_access_to_nsfs_dir(ns, object_sdk) : false;
+        dbg.log1('bucketspace_fs.validate_fs_bucket_access:', bucket.name.unwrap(), is_nsfs_bucket, accessible);
+        return accessible;
+    }
+
+    async _has_access_to_nsfs_dir(ns, object_sdk) {
+        const account = object_sdk.requesting_account;
+        dbg.log1('_has_access_to_nsfs_dir: nsr: ', ns, 'account.nsfs_account_config: ', account && account.nsfs_account_config);
+        // nsfs bucket
+        if (!account || !account.nsfs_account_config || _.isUndefined(account.nsfs_account_config.uid) ||
+            _.isUndefined(account.nsfs_account_config.gid)) return false;
+        try {
+            dbg.log1('_has_access_to_nsfs_dir: checking access:', ns.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
+            const path_to_check = path.join(ns.write_resource.resource.fs_root_path, ns.write_resource.path || '');
+            const fs_context = prepare_fs_context(object_sdk, ns.write_resource.resource.fs_backend);
+            await nb_native().fs.checkAccess(fs_context, path_to_check);
+            return true;
+        } catch (err) {
+            dbg.log0('_has_access_to_nsfs_dir: failed', err);
+            if (err.code === 'ENOENT' || err.code === 'EACCES' || (err.code === 'EPERM' && err.message === 'Operation not permitted')) return false;
+            throw err;
+        }
     }
 }
 
--- a/src/test/unit_tests/nc_coretest.js
+++ b/src/test/unit_tests/nc_coretest.js
@@ -375,7 +375,7 @@ function create_namespace_resource_mock(options) {
 async function create_bucket_manage(options) {
     const { resource, path } = options.namespace.write_resource;
     const bucket_storage_path = p.join(nsrs_to_root_paths[resource], path);
-    const cli_options = { name: options.name, owner: NC_CORETEST, path: bucket_storage_path};
+    const cli_options = { name: options.name, owner: options.owner || NC_CORETEST, path: bucket_storage_path};
     await exec_manage_cli(nc_nsfs_manage_entity_types.BUCKET, nc_nsfs_manage_actions.ADD, cli_options);
 }
 
--- a/src/test/unit_tests/test_bucketspace.js
+++ b/src/test/unit_tests/test_bucketspace.js
@@ -478,10 +478,19 @@ mocha.describe('bucket operations - namespace_fs', async function() {
         assert.ok(list_ok);
     });
     mocha.it('list buckets with dn', async function() {
-        // Give s3_correct_uid access to the required buckets
         const res = await s3_correct_dn_default_nsr.listBuckets({});
         console.log(inspect(res));
-        const list_ok = bucket_in_list([bucket_name + "-s3-2"], [], res.Buckets);
+        const list_ok = bucket_in_list([bucket_name + '-s3-2'], [], res.Buckets);
+        assert.ok(list_ok);
+    });
+
+    mocha.it('delete bucket with dn', async function() {
+        const bucket = bucket_name + '-s3-2';
+        await s3_correct_dn_default_nsr.deleteBucket({ Bucket: bucket });
+        await fs_utils.file_must_not_exist(path.join(s3_new_buckets_path, bucket));
+        const res = await s3_correct_dn_default_nsr.listBuckets({});
+        console.log(inspect(res));
+        const list_ok = bucket_in_list([], [bucket_name + '-s3-2'], res.Buckets);
         assert.ok(list_ok);
     });
     mocha.it('put object with out uid gid', async function() {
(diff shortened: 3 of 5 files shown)


src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,72 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        dbg.log1('has_bucket_action_permission:', bucket.name.unwrap(), account_identifier, bucket.system_owner, bucket.owner_account);
+
+        const is_system_owner = account_identifier === bucket.system_owner.unwrap();
+
+        // If the system owner account wants to access the bucket, allow it
+        if (is_system_owner) return true;
+        const is_owner = (bucket.bucket_owner.unwrap() === account_identifier);
+        const bucket_policy = bucket.s3_policy;
+
+        if (!bucket_policy) return false;
```
naveenpaul1, line 644: @romayalon should we return true when `is_owner` true and there is no policy for buckey?

romayalon: see lines 657-658, if bucket policy result is DENY we will return false

src/sdk/bucketspace_fs.js
```diff
@@ -202,55 +202,28 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             }
             throw this._translate_bucket_error_codes(err);
```
shirady, line 203: Would you please move lines 195 to 203 to a function?
The whole logic is just for getting the entries, not the "core" of the list buckets operation.

src/sdk/bucketspace_fs.js
```diff
@@ -202,55 +202,28 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
-            if (err.code === 'ENOENT' || err.code === 'EACCES' || (err.code === 'EPERM' && err.message === 'Operation not permitted')) return false;
-            throw err;
-        }
+        const buckets = await P.map_with_concurrency(10, entries, async entry => {
+            if (native_fs_utils.isDirectory(entry) || !entry.name.endsWith('.json')) {
+                return;
+            }
+            const bucket_name = this.get_bucket_name(entry.name);
+            const bucket = await object_sdk.read_bucket_sdk_config_info(bucket_name);
+            const bucket_policy_accessible = await this.has_bucket_action_permission(bucket, account, "s3:ListBucket");
```
shirady, line 213: regarding: "s3:ListBucket":
Is there a special reason for double quotes?
I think it would be better to save it as a constant variable and reuse it whenever needed.

src/sdk/bucketspace_fs.js
```diff
@@ -202,55 +202,28 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+            if (!bucket_policy_accessible) return;
+            const fs_accessible = await this.validate_fs_bucket_access(bucket, object_sdk);
+            return fs_accessible && bucket;
+        });
+        const filteres_buckets = buckets.filter(bucket => bucket);
+        return { buckets: filteres_buckets };
     }
 
+
```
shirady, line 222: I think this new line redundant.

src/sdk/bucketspace_fs.js
shirady: @romayalon I didn't understand why it is different from:
https://github.com/noobaa/noobaa-core/blob/8ef4d07243d91c030a9501e8903413dca760d4df/src/server/common_services/auth_server.js#L683
Worth adding a comment with an explanation about this.

src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,72 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        const ns = bucket.namespace;
+        const is_nsfs_bucket = object_sdk.is_nsfs_bucket(ns);
+        const accessible = is_nsfs_bucket ? await this._has_access_to_nsfs_dir(ns, object_sdk) : false;
+        dbg.log1('bucketspace_fs.validate_fs_bucket_access:', bucket.name.unwrap(), is_nsfs_bucket, accessible);
+        return accessible;
+    }
+
+    async _has_access_to_nsfs_dir(ns, object_sdk) {
+        const account = object_sdk.requesting_account;
+        dbg.log0('_has_access_to_nsfs_dir: nsr: ', ns, 'account.nsfs_account_config: ', account && account.nsfs_account_config);
```
shirady, line 672: `log0` on purpose? (the rest of the added logs were `log1` in utils function)

src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,72 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        // nsfs bucket
+        if (!account || !account.nsfs_account_config || _.isUndefined(account.nsfs_account_config.uid) ||
```
shirady, line 674: Account name can be also a number, for example, number 0, so I would suggest changing it to `_.isUndefined(account)`.
At this point, just want to make sure - Did we already converted the distinguished_name to uid and gid? (if you can refer where).

src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,72 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+            _.isUndefined(account.nsfs_account_config.gid)) return false;
+        try {
+            dbg.log0('_has_access_to_nsfs_dir: checking access:', ns.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
+            await nb_native().fs.checkAccess({
+                uid: account.nsfs_account_config.uid,
+                gid: account.nsfs_account_config.gid,
+                backend: ns.write_resource.resource.fs_backend,
+                warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
+            }, path.join(ns.write_resource.resource.fs_root_path, ns.write_resource.path || ''));
```
shirady, line 683: I think it will be more readable to separate it into the variables `fs_context` and `path`.

src/test/unit_tests/test_bucketspace.js
```diff
@@ -41,6 +41,169 @@ const DEFAULT_FS_CONFIG = {
...
+        },
+        // 'account4': {
+        //     bucket: 'bucket4',
+        //     create_bucket_via: 'CLI',
+        //     account_info: {
+        //         uid: 1234,
+        //         gid: 1234,
+        //         new_buckets_path
+        //     }
+        // },
```
shirady, line 83: Why is it in comment? better to explain it or delete it.

src/test/unit_tests/test_bucketspace.js
```diff
@@ -41,6 +41,169 @@ const DEFAULT_FS_CONFIG = {
...
+            }
+        }
+    });
+
+    mocha.after(async function() {
+        this.timeout(60000); // eslint-disable-line no-invalid-this
+        for (const account_name of Object.keys(accounts)) {
+            const { bucket } = accounts[account_name];
+            await rpc_client.bucket.delete_bucket({ name: bucket });
+            await rpc_client.account.delete_account({ email: `${account_name}@noobaa.com` });
```
shirady, line 135: Where do we add the `@noobaa.com` to the email account?

src/test/unit_tests/test_bucketspace.js
```diff
@@ -41,6 +41,169 @@ const DEFAULT_FS_CONFIG = {
...
+        }
+        await fs_utils.folder_delete(tmp_fs_root);
+    });
+
+    mocha.it('list buckets - each account can list only its own bucket - no bucket policies applied ', async function() {
+        for (const account of Object.values(accounts)) {
+            const { s3_client } = account;
+            const res = await s3_client.listBuckets({});
+            const buckets = res.Buckets.filter(bucket => bucket.Name !== 'first.bucket');
```
shirady, line 144: Please add a comment about the filtering of 'first.bucket'

src/test/unit_tests/test_bucketspace.js
```diff
@@ -41,6 +41,169 @@ const DEFAULT_FS_CONFIG = {
...
+            .map(generated =>
+                rpc_client.bucket.put_bucket_policy({
+                    name: generated.params.bucket,
+                    policy: generated.policy,
+                })
+            )
+        );
+        let s3_client = accounts.account1.s3_client;
+        let res = await s3_client.listBuckets({});
+        let buckets = res.Buckets.filter(bucket => bucket.Name !== 'first.bucket' && bucket.Name !== accounts.account1.bucket);
```
shirady, line 195: Why did you add the second condition `bucket.Name !== accounts.account1.bucket`?
Please add an explanation in a comment

src/test/unit_tests/test_bucketspace.js
```diff
@@ -41,6 +41,169 @@ const DEFAULT_FS_CONFIG = {
...
+        }
+
+        // delete bucket policy
+        await rpc_client.bucket.put_bucket_policy({
+            name: public_bucket,
+            policy: '',
+        });
+    });
+
+    mocha.it('account2 - set allow all only account1, account1 should list bucket2 but account3 shouldnt', async function() {
```
shirady, line 177: This test seems complicated to understand, so try to simplify it (remember that it might fail to someone who touches a small part of the code and needs to understand how to debug it...).

src/test/unit_tests/test_bucketspace.js
```diff
@@ -478,12 +641,22 @@ mocha.describe('bucket operations - namespace_fs', async function() {
...
     mocha.it('list buckets with dn', async function() {
-        // Give s3_correct_uid access to the required buckets
         const res = await s3_correct_dn_default_nsr.listBuckets({});
         console.log(inspect(res));
-        const list_ok = bucket_in_list([bucket_name + "-s3-2"], [], res.Buckets);
+        const list_ok = bucket_in_list([bucket_name + '-s3-2'], [], res.Buckets);
         assert.ok(list_ok);
     });
+
+    mocha.it('delete bucket with dn', async function() {
```
shirady, line 650: Why did you add this test?

src/sdk/bucketspace_fs.js
romayalon: this needs to be done as part of a refactoring PR, I see we can have this function used by health.js and manage_nsfs.js

src/test/unit_tests/test_bucketspace.js
romayalon: why not?
this bucket was not cleaned on the teardown, so instead of removing it on teardown I added a test that checks also that we can delete this bucket using dn

shirady: I asked because I didn't understand how this test is related to the list buckets.
As I understand it now it is not related to the change, but you added a test to improve our coverage with bucket deletion using a distinguished name.

src/sdk/bucketspace_fs.js
romayalon: account is an object here

src/test/unit_tests/test_bucketspace.js
romayalon: I removed it

src/test/unit_tests/test_bucketspace.js
romayalon: I added comments

src/sdk/bucketspace_fs.js
```diff
@@ -202,55 +202,28 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+            }
+            const bucket_name = this.get_bucket_name(entry.name);
+            const bucket = await object_sdk.read_bucket_sdk_config_info(bucket_name);
+            const bucket_policy_accessible = await this.has_bucket_action_permission(bucket, account, 's3:ListBucket');
+            if (!bucket_policy_accessible) return;
+            const fs_accessible = await this.validate_fs_bucket_access(bucket, object_sdk);
+            if (!fs_accessible) return;
+            return bucket;
+        });
+        const filteres_buckets = buckets.filter(bucket => bucket);
```
shirady, line 219: I think a more meaningful variable name can be nice here (`accessible_buckets `maybe?).

src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,73 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+
+    /////////////////
+    ///// UTILS /////
+    /////////////////
+
+    // TODO: move the following 3 functions - has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
+    // so they can be re-used
+    async has_bucket_action_permission(bucket, account, action, bucket_path = "") {
+        const account_identifier = account.name.unwrap();
+        dbg.log1('has_bucket_action_permission:', bucket.name.unwrap(), account_identifier, bucket.system_owner, bucket.owner_account);
```
shirady, line 635: Don't you need to unwrap `bucket.system_owner` and `bucket.owner_account`?

src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,73 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
         // TODO
     }
+
+    /////////////////
+    ///// UTILS /////
+    /////////////////
+
+    // TODO: move the following 3 functions - has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
+    // so they can be re-used
+    async has_bucket_action_permission(bucket, account, action, bucket_path = "") {
```
shirady, line 633: Why changing the name of the parameter from `method` to `action`?

src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,73 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        if (!bucket_policy) return is_owner;
+        if (!action) {
+            throw new Error('has_bucket_action_permission: action is required');
+        }
+
+        const result = await bucket_policy_utils.has_bucket_policy_permission(
+            bucket_policy,
+            account_identifier,
+            action,
+            `arn:aws:s3:::${bucket.name.unwrap()}${bucket_path}`,
```
shirady, line 653: It is name and path without any separator between them?

src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,73 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+            undefined
```
shirady, line 654: What does it mean that we always pass undefined?
I see that in the internal function it will be `condition_fit` true, but I don't understand the whole context.

src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,73 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+    }
+
+    async _has_access_to_nsfs_dir(ns, object_sdk) {
+        const account = object_sdk.requesting_account;
+        dbg.log1('_has_access_to_nsfs_dir: nsr: ', ns, 'account.nsfs_account_config: ', account && account.nsfs_account_config);
+        // nsfs bucket
+        if (!account || !account.nsfs_account_config || _.isUndefined(account.nsfs_account_config.uid) ||
+            _.isUndefined(account.nsfs_account_config.gid)) return false;
+        try {
+            dbg.log0('_has_access_to_nsfs_dir: checking access:', ns.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
```
shirady, line 677: `dbg.log0` on purpose?

src/sdk/bucketspace_fs.js
```diff
@@ -650,6 +623,73 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        return accessible;
+    }
+
+    async _has_access_to_nsfs_dir(ns, object_sdk) {
+        const account = object_sdk.requesting_account;
+        dbg.log1('_has_access_to_nsfs_dir: nsr: ', ns, 'account.nsfs_account_config: ', account && account.nsfs_account_config);
+        // nsfs bucket
+        if (!account || !account.nsfs_account_config || _.isUndefined(account.nsfs_account_config.uid) ||
+            _.isUndefined(account.nsfs_account_config.gid)) return false;
+        try {
```
shirady, line 676: At this point we know that `account.nsfs_account_config.uid` and `account.nsfs_account_config.gid` are defined, we can define variables for easier readability:
const uid = account.nsfs_account_config.uid;
const gid = account.nsfs_account_config.gid;

src/test/unit_tests/test_bucketspace.js
```diff
@@ -478,11 +478,21 @@ mocha.describe('bucket operations - namespace_fs', async function() {
...
+        const bucket = bucket_name + '-s3-2';
+        await s3_correct_dn_default_nsr.deleteBucket({ Bucket: bucket });
+        await fs_utils.file_must_not_exist(path.join(s3_new_buckets_path, bucket));
+        const res = await s3_correct_dn_default_nsr.listBuckets({});
+        console.log(inspect(res));
+        const list_ok = bucket_in_list([], [bucket_name + '-s3-2'], res.Buckets);
         assert.ok(list_ok);
+
```
shirady, line 495: redundant newline

src/test/unit_tests/test_bucketspace.js
```diff
@@ -1532,7 +1433,294 @@ mocha.describe('nsfs account configurations', function() {
...
+        assert.equal(buckets.length, 2);
+        assert.deepStrictEqual(buckets, [accounts.account3.bucket, 'first.bucket']);
+    });
+});
+
 function get_tmp_path_by_os(_path) {
     return process.platform === MAC_PLATFORM ? '/private/' + _path : _path;
 }
 
+
```
shirady, line 1608: redundant newline

src/test/unit_tests/test_bucketspace.js
```diff
@@ -1532,7 +1433,294 @@ mocha.describe('nsfs account configurations', function() {
...
+        nsfs_account_config,
+        default_resource
+    });
+    return {
+        access_key: account.access_keys[0].access_key.unwrap(),
+        secret_key: account.access_keys[0].secret_key.unwrap(),
+        email: `${random_name}@noobaa.com`
+    };
+}
+
```
shirady, line 1658: redundant newline

src/test/unit_tests/test_bucketspace.js
```diff
@@ -1532,7 +1433,294 @@ mocha.describe('nsfs account configurations', function() {
...
+        }
+        assert.fail(`put_object failed ${err}, ${err.stack}`);
+    }
+
+    await s3_account.deleteObject({
+        Bucket: bucket,
+        Key: key,
+    });
+}
+
```
shirady, line 1683: redundant newline

shirady:
@romayalon, general questions that this PR raised:
1. Would you please attach an explanation (or refer to a link) about the scope of the bucket policy that we support in NC NSFS? From the PR I understand that we check and validate 's3:ListBucket'.
2. In manage nsfs CLI the list buckets operation is done by the admin who is always the system owner, therefore we don't check bucket policy before the bucket list, is that correct?

src/sdk/bucketspace_fs.js
romayalon: I removed this variable, it was there for debugging

src/sdk/bucketspace_fs.js
romayalon: https://github.com/noobaa/noobaa-core/blob/master/src/server/common_services/auth_server.js#L673
it wasn't changed

src/sdk/bucketspace_fs.js
romayalon: it's the same as in https://github.com/noobaa/noobaa-core/blob/master/src/server/common_services/auth_server.js#L673

src/sdk/bucketspace_fs.js
romayalon: Please notice that declaring variables might be sometimes needed for readability like in path_to_check , but we should keep in mind that declaring variables can be very wasteful in space and memory.

Eventually, it was not needed because I decided to avoid redundant code and just call a function instead-
```
 const path_to_check = path.join(ns.write_resource.resource.fs_root_path, ns.write_resource.path || '');
 const fs_context = prepare_fs_context(object_sdk, ns.write_resource.resource.fs_backend);
```

src/sdk/bucketspace_fs.js
romayalon: req is needed when conditions are passed in order to run the condition's function.
see here - https://github.com/noobaa/noobaa-core/pull/7273

src/sdk/bucketspace_fs.js
```diff
@@ -203,55 +205,27 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
-                warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
-            }, path.join(ns.write_resource.resource.fs_root_path, ns.write_resource.path || ''));
-
-            return true;
-        } catch (err) {
-            dbg.log0('_has_access_to_nsfs_dir: failed', err);
-            if (err.code === 'ENOENT' || err.code === 'EACCES' || (err.code === 'EPERM' && err.message === 'Operation not permitted')) return false;
-            throw err;
-        }
+        const buckets = await P.map_with_concurrency(10, entries, async entry => {
```
shirady, line 210: Please add a comment before `map_with_concurrency` or in the JSDOC of the function with an explanation of the procedure (you can reuse the PR description).

src/sdk/bucketspace_fs.js
```diff
@@ -203,55 +205,27 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+            if (native_fs_utils.isDirectory(entry) || !entry.name.endsWith('.json')) {
+                return;
+            }
+            const bucket_name = this.get_bucket_name(entry.name);
+            const bucket = await object_sdk.read_bucket_sdk_config_info(bucket_name);
+            const bucket_policy_accessible = await this.has_bucket_action_permission(bucket, account, 's3:ListBucket');
+            if (!bucket_policy_accessible) return;
+            const fs_accessible = await this.validate_fs_bucket_access(bucket, object_sdk);
+            if (!fs_accessible) return;
+            return bucket;
```
shirady, line 220: Do you know why in the previous implementation `creation_date` was used? (It is not used in your new implementation).

src/test/unit_tests/nc_coretest.js
```diff
@@ -375,7 +375,7 @@ function create_namespace_resource_mock(options) {
 async function create_bucket_manage(options) {
     const { resource, path } = options.namespace.write_resource;
     const bucket_storage_path = p.join(nsrs_to_root_paths[resource], path);
-    const cli_options = { name: options.name, owner: NC_CORETEST, path: bucket_storage_path};
+    const cli_options = { name: options.name, owner: options.owner || NC_CORETEST, path: bucket_storage_path};
```
shirady, line 378: Why did you add this: `options.owner || NC_CORETEST`?

src/test/unit_tests/test_bucketspace.js
```diff
@@ -1532,7 +1432,291 @@ mocha.describe('nsfs account configurations', function() {
...
+            const buckets = res.Buckets.map(bucket => bucket.Name).sort();
+            assert.equal(buckets.length, 2);
+            assert.deepStrictEqual(buckets, [account.bucket, 'first.bucket']);
+        }
+    });
+
+    mocha.it('account1 - all accounts are allowed to list bucket1', async function() {
+        // allow all accounts to list bucket1
+        const public_bucket = accounts.account1.bucket;
+        const bucket_policy = test_utils.generate_s3_policy('*', public_bucket, ['s3:*']);
```
shirady, line 1546: Why do you add all actions and not just the `'s3:ListBucket'`?

src/test/unit_tests/nc_coretest.js
romayalon: because we want to let to create buckets that their owner is not NC_CORETEST...

src/sdk/bucketspace_fs.js
romayalon: yes, it was a part of this fix - https://github.com/noobaa/noobaa-core/pull/7799
bucket contains creation date

shirady (approved):
LGTM

naveenpaul1 (approved):
LGTM

src/sdk/bucketspace_fs.js
```diff
@@ -651,6 +637,68 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        const account = object_sdk.requesting_account;
+        dbg.log1('_has_access_to_nsfs_dir: nsr: ', ns, 'account.nsfs_account_config: ', account && account.nsfs_account_config);
+        // nsfs bucket
+        if (!account || !account.nsfs_account_config || _.isUndefined(account.nsfs_account_config.uid) ||
+            _.isUndefined(account.nsfs_account_config.gid)) return false;
+        try {
+            dbg.log1('_has_access_to_nsfs_dir: checking access:', ns.write_resource, account.nsfs_account_config.uid, account.nsfs_account_config.gid);
+            const path_to_check = path.join(ns.write_resource.resource.fs_root_path, ns.write_resource.path || '');
+            const fs_context = prepare_fs_context(object_sdk, ns.write_resource.resource.fs_backend);
+            await nb_native().fs.checkAccess(fs_context, path_to_check);
```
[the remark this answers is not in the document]
romayalon: Hey @naveenpaul1 great question,
for listing buckets we need to check only read permissions so checkAccess() is the correct function to use,
But as part of https://github.com/noobaa/noobaa-core/pull/7775 we will merge the 2 functionalities so stay tuned :)

PR #7813, opened 2024-02-12
NSFS | NC | Improve Performance in `verify_bucket_owner`

shirady:
### Explain the changes
1. Improve performance in `verify_bucket_owner` - in the past, the identifier of `bucket_owner` was the account email, in PR #7801 we changed it to the account name. Therefore, we don't need to go over the config files and open each account when iterating, because the name of the account config file is `<account-name>.json`.
2. Edit the JSDoc of the function so the description of the variable will be clearer (see [this comment](https://github.com/noobaa/noobaa-core/pull/7801#discussion_r1485787849)).
3. Add constant `SHOW_SECRETS` to improve readability (see [this comment](https://github.com/noobaa/noobaa-core/pull/7813#discussion_r1487749083)).

### Issues:
1. It was a known gap that was mentioned in PR #7801 and #7728.

### Testing Instructions:
1. See comment below.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -215,40 +215,30 @@ async function add_bucket(data) {
 }
 
 /** verify_bucket_owner will check if the bucket_owner has an account
+ * bucket_owner is the account name in the account schema
  * after it finds one, it returns the account id, otherwise it would throw an error
  * (in case the action is add bucket it also checks that the owner has allow_bucket_creation)
- * @param {string} bucket_owner account name
+ * @param {string} bucket_owner
  * @param {string} action
  */
 async function verify_bucket_owner(bucket_owner, action) {
-    let is_bucket_owner_exist = false;
-    let is_allow_bucket_creation = false;
-    let account_id;
-    const show_secrets = false;
-    const fs_context = native_fs_utils.get_process_fs_context();
-    const entries = await nb_native().fs.readdir(fs_context, accounts_dir_path);
-    // Gap - should replace this implementation
-    // it keeps iterating even if we find that the bucket owner exist
-    await P.map_with_concurrency(10, entries, async entry => {
-        if (entry.name.endsWith('.json') && !is_bucket_owner_exist) {
-            const full_path = path.join(accounts_dir_path, entry.name);
-            const data = await get_config_data(full_path, show_secrets);
-            if (data.name === bucket_owner) {
-                is_bucket_owner_exist = true;
-                is_allow_bucket_creation = data.allow_bucket_creation;
-                account_id = data._id;
-            }
+    // check if bucket owner exists
+    const account_config_path = get_config_file_path(accounts_dir_path, bucket_owner);
+    let account;
+    try {
+        account = await get_config_data(account_config_path);
+    } catch (err) {
+        if (err.code === 'ENOENT') {
+            const detail_msg = `bucket owner ${bucket_owner} does not exists`;
+            throw_cli_error(ManageCLIError.BucketSetForbiddenNoBucketOwner, detail_msg, {bucket_owner: bucket_owner});
         }
-    });
-
-    if (!is_bucket_owner_exist) {
-        const detail_msg = `bucket owner ${bucket_owner} does not exists`;
-        throw_cli_error(ManageCLIError.BucketSetForbiddenNoBucketOwner, detail_msg, {bucket_owner: bucket_owner});
+        throw err;
     }
-    if (action === ACTIONS.ADD && !is_allow_bucket_creation) {
+    // check if bucket owner has the permission to create bucket (for bucket add only)
+    if (action === ACTIONS.ADD && !account.allow_bucket_creation) {
             throw_cli_error(ManageCLIError.BucketCreationNotAllowed, bucket_owner);
     }
-    return account_id;
+    return account._id;
 }
 
 async function get_bucket_status(data) {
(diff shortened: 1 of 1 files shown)


shirady:
# Manual Testing Instructions:
As described in issue #7724 and PR #7785:
- First, create the `FS_ROOT` and a directory for a bucket: `mkdir -p /tmp/nsfs_root1/my-bucket` and give permissions `chmod 777 /tmp/nsfs_root1/my-bucket`.

## Bucket Create (owner doesn't exists)
1. Create a bucket (should see an error): `sudo node src/cmd/manage_nsfs bucket add --name<bucket-name> --owner <account-name> --path  /tmp/nsfs_root1/my-bucket` (the `<account-name>` should be an account that was not created yet).

## Bucket Update (owner doesn't exists)
1. Create an account `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path /tmp/nsfs_root1 --uid <uid> --gid <gid>`.
2. Create a bucket (should create): `sudo node src/cmd/manage_nsfs bucket add --name <bucket-name> --owner <account-name> --path <path>` (the `<account-name>` should be an account that already exists, from step 1).
3. Update the bucket  (should see an error): `sudo node src/cmd/manage_nsfs bucket update --name <bucket-name> --owner <account-name>`(the `<account-name>` should be an account that was not created yet).

## Bucket Create (owner doesn't have permissions)
1. Create an account without passing the flag `new-buckets-path`: `sudo node src/cmd/manage_nsfs account add --name <account-name> --uid <uid> --<gid>`
2. Create a bucket (should see an error): `sudo node src/cmd/manage_nsfs bucket add --name <bucket-name> --owner <account-name> --path /tmp/nsfs_root1/my-bucket`.

src/cmd/manage_nsfs.js
```diff
@@ -223,38 +223,26 @@ async function add_bucket(data) {
...
-            if (data.name === bucket_owner) {
-                is_bucket_owner_exist = true;
-                is_allow_bucket_creation = data.allow_bucket_creation;
-                account_id = data._id;
-            }
-        }
-    });
-
+    // check if bucket owner exists
+    const fs_context = native_fs_utils.get_process_fs_context(config_root_backend);
```
shirady, line 231: @romayalon I was not sure in the `get_process_fs_context` if I should pass the `config_root_backend`.
I know that you planned to refactor this function and the argument will be `backend`.
In manage_nsfs there are calls to `get_process_fs_context` with `config_root_backend` and without any argument.

src/cmd/manage_nsfs.js
```diff
@@ -223,38 +223,26 @@ async function add_bucket(data) {
 /** verify_bucket_owner will check if the bucket_owner has an account
  * after it finds one, it returns the account id, otherwise it would throw an error
  * (in case the action is add bucket it also checks that the owner has allow_bucket_creation)
- * @param {string} bucket_owner account name
+ * @param {string} bucket_owner - bucket owner is the account name
```
shirady, line 226: @naveenpaul1, is that change is better? (regarding your comment [here](https://github.com/noobaa/noobaa-core/pull/7801#discussion_r1484038868)).

naveenpaul1: I think it will make more sense if you can add the parameter description in the above method description, adding a description not align with our previous coding style.
`verify_bucket_owner will check if the bucket_owner has an account`
to
`verify_bucket_owner will check if the bucket_owner(same as account name in account schema) has an account`

shirady: Thanks, I change it to be part of the description (and not inline variable description).

src/cmd/manage_nsfs.js
```diff
@@ -215,40 +215,29 @@ async function add_bucket(data) {
...
-                is_allow_bucket_creation = data.allow_bucket_creation;
-                account_id = data._id;
-            }
-        }
-    });
-
+    // check if bucket owner exists
+    const fs_context = native_fs_utils.get_process_fs_context(config_root_backend);
+    const account_config_path = get_config_file_path(accounts_dir_path, bucket_owner);
+    const is_bucket_owner_exist = await native_fs_utils.is_path_exists(fs_context, account_config_path);
```
romayalon, line 228: please try to avoid using both is_path_exists() and get_config_data() to the same path,
we can reduce these 2 FS calls to 1 to increase performance

src/cmd/manage_nsfs.js
```diff
@@ -215,40 +215,29 @@ async function add_bucket(data) {
...
     }
-    if (action === ACTIONS.ADD && !is_allow_bucket_creation) {
+    // check if bucket owner has the permission to create bucket (for bucket add only)
+    const show_secrets = false;
+    const account = await get_config_data(account_config_path, show_secrets);
+    if (action === ACTIONS.ADD && !account.allow_bucket_creation) {
             throw_cli_error(ManageCLIError.BucketCreationNotAllowed, bucket_owner);
     }
-    return account_id;
+    // return the account id
```
romayalon, line 239: why is this comment needed? it's very clear

src/cmd/manage_nsfs.js
shirady: Thank you

src/cmd/manage_nsfs.js
shirady: Removed, thanks.

src/cmd/manage_nsfs.js
```diff
@@ -215,40 +215,31 @@ async function add_bucket(data) {
...
+ * @param {string} bucket_owner
  * @param {string} action
  */
 async function verify_bucket_owner(bucket_owner, action) {
-    let is_bucket_owner_exist = false;
-    let is_allow_bucket_creation = false;
-    let account_id;
+    // check if bucket owner exists
+    const account_config_path = get_config_file_path(accounts_dir_path, bucket_owner);
     const show_secrets = false;
```
naveenpaul1, line 227: @shirady do we need a const here, we can directly pass `false`?

naveenpaul1 (commented):
small comment
LGTM

src/cmd/manage_nsfs.js
shirady: @naveenpaul1 we don't need it, I think it is more readable than just `false`.
I will try to add it in the constant file, and see if we like the result.

romayalon (approved):

Issue #7786, opened 2024-02-04
NSFS | NC | Let the User Explicitly Set `allow_bucket_creation` 


shirady:
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
1. Currently, the `allow_bucket_creation` value is inferred from the `new_buckets_path`: if the user passes a value to `new_buckets_path` then it is true, otherwise false.

### Expected behavior
1. Allow the user to pass the value of `allow_bucket_creation` explicitly and update it (from `true` to `false` and vice versa).

### Steps to reproduce
1. none

### More information - Screenshots / Logs / Other output
Please take into account that it will be a flag is manage nsfs CLI and a property inside the file when using the `from-file` option.

shirady:
Please add a unit test for updating a bucket (see [this comment](https://github.com/noobaa/noobaa-core/pull/7785#discussion_r1477749730) with manual description):

1. Create an account with `new_buckets_path` (`allow_bucket_creation` is true).
2. Created a bucket.
3. Update the account (in a way you choose) so `allow_bucket_creation` will be false.
4. Update the bucket already created (step 2) - for example `--fs_backend GPFS` instead of the default value => should pass.

shirady:
The verification will be tracked by the [BZ 2262992](https://bugzilla.redhat.com/show_bug.cgi?id=2262992).

shirady:
It was validated in mentioned BZ above.

PR #7827, opened 2024-02-18
NSFS | NC | Add a Flag to Change the Value of  `allow_bucket_creation`

shirady:
### Explain the changes
1. Add a flag to change the value of  `allow_bucket_creation`.
2. Edit details message when `BucketCreationNotAllowed` is thrown.
3. Add the option for boolean flags to pass 'true' and 'false' (`wide`, `show_secrets`, `regenerate`).
4. Add a new error `InvalidBooleanValue`.
5. Reduce the number of spaces in the help menu (without glacier changes to match the existing style, since it changes in PR #7848).

### Issues: Fixed [BZ 2262992](https://bugzilla.redhat.com/show_bug.cgi?id=2262992) #7786
1. Currently, the `allow_bucket_creation` value is inferred from the `new_buckets_path`: if the user passes a value to `new_buckets_path` then it is true, otherwise, it is false.
2. The fix is partial for the BZ 2262992, because the automated tests are using `from_file` flag, after this PR is merged we need to merge also PR #7779 as a fix.
3. GAP - I think we need to extract `wide` not part of the fetch data or extract and remove it before writing the config file (opened issue #7871).

#### Manual Tests:
1. Use the new flag `bucket_creation` and see the value of `allow_bucket_creation` changes, for example:
- First, create the `FS_ROOT` and a directory for a bucket: `mkdir -p /tmp/nsfs_root1/my-bucket` and give permissions `chmod 777 /tmp/nsfs_root1/` `chmod 777 /tmp/nsfs_root1/my-bucket`.
- Create an account with `new_buckets_path` (`allow_bucket_creation` inferred to be true): `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path /tmp/nsfs_root1 --uid <uid> --gid <gid>`
- Update the account (so `allow_bucket_creation` will be false): `sudo node src/cmd/manage_nsfs account update --name <account-name> --allow_bucket_creation false`.

4. Use the boolean flags with true/false values, for example: `sudo node src/cmd/manage_nsfs account status --name <account-name> --show_secrets=true`

#### Unit Tests:
please run:
- `sudo node ./node_modules/.bin/_mocha src/test/unit_tests/test_nc_nsfs_cli.js`.
- `sudo npx jest test_nc_nsfs_account_cli.test.js`.

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -22,7 +22,7 @@ const manage_nsfs_glacier = require('../manage_nsfs/manage_nsfs_glacier');
 const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
 const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
 const { print_usage } = require('../manage_nsfs/manage_nsfs_help_utils');
-const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE,
+const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, BOOLEAN_STRING_VALUES,
     LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS, GLACIER_ACTIONS } = require('../manage_nsfs/manage_nsfs_constants');
 const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 
@@ -146,7 +146,7 @@ async function fetch_bucket_data(argv, from_file) {
             owner_account: undefined,
             system_owner: argv.owner, // GAP - needs to be the system_owner (currently it is the account name)
             bucket_owner: argv.owner,
-            wide: argv.wide,
+            wide: _.isUndefined(argv.wide) ? undefined : get_boolean_or_string_value(argv.wide),
             tag: undefined, // if we would add the option to tag a bucket using CLI, this should be changed
             versioning: action === ACTIONS.ADD ? 'DISABLED' : undefined,
             creation_date: action === ACTIONS.ADD ? new Date().toISOString() : undefined,
@@ -239,7 +239,9 @@ async function verify_bucket_owner(bucket_owner, action) {
     }
     // check if bucket owner has the permission to create bucket (for bucket add only)
     if (action === ACTIONS.ADD && !account.allow_bucket_creation) {
-            throw_cli_error(ManageCLIError.BucketCreationNotAllowed, bucket_owner);
+            const detail_msg = `${bucket_owner} account not allowed to create new buckets. ` +
+            `Please make sure to have a valid new_buckets_path and enable the flag allow_bucket_creation`;
+            throw_cli_error(ManageCLIError.BucketCreationNotAllowed, detail_msg);
     }
     return account._id;
 }
@@ -333,7 +335,7 @@ async function manage_bucket_operations(action, data, argv) {
 
 async function account_management(argv, from_file) {
     const action = argv._[1] || '';
-    const show_secrets = Boolean(argv.show_secrets) || false;
+    const show_secrets = get_boolean_or_string_value(argv.show_secrets);
     const data = await fetch_account_data(argv, from_file);
     await manage_account_operations(action, data, show_secrets, argv);
 }
@@ -379,7 +381,8 @@ async function fetch_account_data(argv, from_file) {
         const regenerate = action === ACTIONS.ADD;
         access_keys = set_access_keys(argv, regenerate);
     } else if (action === ACTIONS.UPDATE) {
-        access_keys = set_access_keys(argv, Boolean(argv.regenerate));
+        const regenerate = get_boolean_or_string_value(argv.regenerate);
+        access_keys = set_access_keys(argv, regenerate);
         new_access_key = access_keys[0].access_key;
         access_keys[0].access_key = undefined; //Setting it as undefined so we can replace the symlink
     }
@@ -389,7 +392,7 @@ async function fetch_account_data(argv, from_file) {
             name: _.isUndefined(argv.name) ? undefined : String(argv.name),
             email: _.isUndefined(argv.name) ? undefined : String(argv.name), // temp, keep the email internally
             creation_date: action === ACTIONS.ADD ? new Date().toISOString() : undefined,
-            wide: argv.wide,
+            wide: _.isUndefined(argv.wide) ? undefined : get_boolean_or_string_value(argv.wide),
             new_name: _.isUndefined(argv.new_name) ? undefined : String(argv.new_name),
             new_access_key,
             access_keys,
@@ -414,8 +417,14 @@ async function fetch_account_data(argv, from_file) {
     // secret_key as SensitiveString
     data.access_keys[0].secret_key = _.isUndefined(data.access_keys[0].secret_key) ? undefined :
         new SensitiveString(String(data.access_keys[0].secret_key));
-    // allow_bucket_creation infer from new_buckets_path
-    data.allow_bucket_creation = !_.isUndefined(data.nsfs_account_config.new_buckets_path);
+    // allow_bucket_creation either set by user or infer from new_buckets_path
+    if (_.isUndefined(argv.allow_bucket_creation)) {
+        data.allow_bucket_creation = !_.isUndefined(data.nsfs_account_config.new_buckets_path);
+    } else if (typeof argv.allow_bucket_creation === 'boolean') {
+        data.allow_bucket_creation = Boolean(argv.allow_bucket_creation);
+    } else { // string of true or false
+        data.allow_bucket_creation = argv.allow_bucket_creation.toLowerCase() === 'true';
+    }
     // fs_backend deletion specified with empty string '' (but it is not part of the schema)
     data.nsfs_account_config.fs_backend = data.nsfs_account_config.fs_backend || undefined;
 
@@ -811,7 +820,7 @@ function validate_flags_arguments(type, action, argv) {
     validate_no_extra_options(type, action, input_options);
     const input_options_with_data = { ...argv };
     delete input_options_with_data._;
-    validate_options_type_by_value(type, input_options_with_data);
+    validate_options_type_by_value(input_options_with_data);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/manage_nsfs/manage_nsfs_help_utils.js
```diff
@@ -82,6 +82,7 @@ Flags:
 --access_key <string>                                     (optional)                        Set the access key for the account (default is generated)
 --secret_key <string>                                     (optional)                        Set the secret key for the account (default is generated)
 --fs_backend <none | GPFS | CEPH_FS | NFSv4>              (optional)                        Set the filesystem type of new_buckets_path (default config.NSFS_NC_STORAGE_BACKEND)
+--bucket_creation <allow | block>                         (optional)                        Set the account to explicitly allow or block bucket creation
```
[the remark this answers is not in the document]
shirady: I will change it.
The open question is about the rest of the boolean flags that we have - if we wish to have the same solution.
If so, it will be in a different PR (after this one will be merged).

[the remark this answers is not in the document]
romayalon: np thanks!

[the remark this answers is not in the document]
naveenpaul1: @shirady is not related to this, Can we reduce the space between the `flag`, option and the description? In terminal because of the huge space description is coming in the next line.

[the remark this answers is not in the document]
shirady: @naveenpaul1 we can change it of course.
The space between the flag and the (optional) was set due to this line:
https://github.com/noobaa/noobaa-core/blob/563e0ab66a9292da2071e3cc5849f3c861e6ed63/src/manage_nsfs/manage_nsfs_help_utils.js#L69

We can reduce the spaces between the (optional) and the description (in a different PR).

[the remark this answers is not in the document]
shirady: @romayalon I didn't understand you - do we want the other boolean option also to be able to send true/false and not just the option? (in a different PR as mentioned).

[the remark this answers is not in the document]
romayalon: I think it's a bit weird to fail the command when  specifying --show_secrets=true

src/cmd/manage_nsfs.js
```diff
@@ -873,8 +881,18 @@ function validate_options_type_by_value(type, input_options_with_data) {
...
-            const err_msg = `type of flag ${option} should be ${type_of_option}`;
-            throw_cli_error(ManageCLIError.InvalidArgumentType, err_msg);
+            // special case for allow_bucket_creation
+            if (option === 'allow_bucket_creation' && type_of_value === 'boolean') {
+                continue;
+            }
+            details = `type of flag ${option} should be ${type_of_option}`;
+            throw_cli_error(ManageCLIError.InvalidArgumentType, details);
+        }
+        if (input_options_with_data.allow_bucket_creation &&
```
romayalon, line 891: can we do all allow_bucket_creation checks in a single condition?

src/manage_nsfs/manage_nsfs_constants.js
```diff
@@ -78,6 +79,8 @@ const OPTION_TYPE = {
     ips: 'string',
 };
 
+const ALLOW_BUCKET_CREATION_VALUES = ['true', 'false'];
```
romayalon, line 82: I suggest changing it to BOOLEAN_STRING_VALUES because we will soon support it on regenerate, show_secrets and wide

src/cmd/manage_nsfs.js
```diff
@@ -414,8 +415,14 @@ async function fetch_account_data(argv, from_file) {
     // secret_key as SensitiveString
     data.access_keys[0].secret_key = _.isUndefined(data.access_keys[0].secret_key) ? undefined :
         new SensitiveString(String(data.access_keys[0].secret_key));
-    // allow_bucket_creation infer from new_buckets_path
-    data.allow_bucket_creation = !_.isUndefined(data.nsfs_account_config.new_buckets_path);
+    // allow_bucket_creation either set by user or infer from new_buckets_path
+    if (_.isUndefined(argv.allow_bucket_creation)) {
+        data.allow_bucket_creation = !_.isUndefined(data.nsfs_account_config.new_buckets_path);
+    } else if (typeof argv.allow_bucket_creation === 'string') {
```
romayalon, line 421:
```suggestion
    } else if (ALLOW_BUCKET_CREATION_VALUES.includes(argv.allow_bucket_creation.toLowerCase()) {
      data.allow_bucket_creation = argv.allow_bucket_creation.toLowerCase();
      }
```

src/manage_nsfs/manage_nsfs_help_utils.js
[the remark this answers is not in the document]
shirady: I will change that, thanks.

src/cmd/manage_nsfs.js
romayalon: I suggested to explicitly check it so the reader will understand the values and not just the type

src/cmd/manage_nsfs.js
romayalon: I saw you added `&& check_boolean_string_value(input_options_with_data.show_secrets)` to the other string/boolean options, can't you add it to allow_bucket_creation as well?

src/cmd/manage_nsfs.js
shirady: We are checking in in the function `validate_options_type_by_value` - type and value.
The above suggested change is in the `fetch_account_data`, which is called later.

src/cmd/manage_nsfs.js
shirady: The `check_boolean_string_value` checks:
1. `allow_bucket_creation` is defined.
2. If it is defined, then we check that the type is `string`.
3. If it is a string we check that it is the strings array that we defined <- here is the change, in the condition we check that it is NOT matched (and add a suitable detail message).

romayalon: You are still throwing InvalidArgument when it's not matching inside the loop. so I still don't see the difference between the checks. If you want a different error I suggested to change  check_boolean_string_value() to throw the errors you want.  
Also, you should check the other options once (check that option is included in an array of the boolean/string options = ['regenerate', 'wide' ...]).

shirady: Now I understand that you're suggesting adding errors, like we have `InvalidFSBackend` and not using the general `InvalidArgument`. I will add this, thanks.

src/cmd/manage_nsfs.js
```diff
@@ -239,7 +239,8 @@ async function verify_bucket_owner(bucket_owner, action) {
     }
     // check if bucket owner has the permission to create bucket (for bucket add only)
     if (action === ACTIONS.ADD && !account.allow_bucket_creation) {
-            throw_cli_error(ManageCLIError.BucketCreationNotAllowed, bucket_owner);
+            const detail_msg = `bucket owner ${bucket_owner}`;
+            throw_cli_error(ManageCLIError.BucketCreationNotAllowed, detail_msg);
```
naveenpaul1, line 243: @shirady could you please add why bucket creation is not allowed for the owner in `BucketCreationNotAllowed`
This message does not give the actual real reason why bucket creation is not allowed for user. Something like
``` Account not allowed to create new buckets, Please add valid new_buckets_path and enable the flag allow_bucket_creation ``` (Assuming new_buckets_path needed)

File: src/cmd/manage_nsfs.js [hunk 6]
```diff
@@ -414,8 +416,14 @@ async function fetch_account_data(argv, from_file) {
...
         new SensitiveString(String(data.access_keys[0].secret_key));
-    // allow_bucket_creation infer from new_buckets_path
-    data.allow_bucket_creation = !_.isUndefined(data.nsfs_account_config.new_buckets_path);
+    // allow_bucket_creation either set by user or infer from new_buckets_path
+    if (_.isUndefined(argv.allow_bucket_creation)) {
+        data.allow_bucket_creation = !_.isUndefined(data.nsfs_account_config.new_buckets_path);
+    } else if (typeof argv.allow_bucket_creation === 'string') {
+        data.allow_bucket_creation = argv.allow_bucket_creation.toLowerCase() === 'true';
+    } else {
+        data.allow_bucket_creation = Boolean(argv.allow_bucket_creation);
```
naveenpaul1, line 425: @shirady @romayalon Are we considering only `allow_bucket_creation` flag to decide for enabling bucket creation, not with `new_buckets_path`, So what will happen when user set `allow_bucket_creation` true and `new_buckets_path` empty and try to create a bucket using S3, What will be the bucket's storage path.

src/cmd/manage_nsfs.js
shirady: It is not part of the changes of this PR, but I will change it, thank you.

src/cmd/manage_nsfs.js
shirady: Looking at the current code it should be fine:
https://github.com/noobaa/noobaa-core/blob/7d2248c73f665060fc43b6c84de3ca38506c76ef/src/sdk/bucketspace_fs.js#L247-L252
In this case the `allow_bucket_creation` is true so it will not throw the first error, but `nsfs_account_config.new_buckets_path` is not defined so it should throw the second error 'MISSING_NSFS_ACCOUNT_CONFIGURATION'.

src/cmd/manage_nsfs.js
shirady: @naveenpaul1 I noticed that changing the message fails some tests in the CI.
I would edit the detailed message instead.
Anyway, the original message is aligned with what we would encounter when using the NSFS server.

naveenpaul1: @shirady This is a request I got from Ramya, May be we can add it in different PR

shirady: The details will appear in the detail message as you can see in the latest commit.

shirady: Attached output example:
```
  "error": {
    "code": "BucketCreationNotAllowed",
    "message": "Not allowed to create new buckets",
    "detail": "shira-1077 account not allowed to create new buckets. Please make sure to have a valid new_buckets_path and enable the flag allow_bucket_creation"
  }
```

romayalon (approved):

Issue #7784, opened 2024-02-01
Signature mismatch with aws cpp sdk that sends `=` unencoded


guymguym:
### Environment info
- NooBaa Version: **5.12** (but the same in current master)
- Platform: **NA**

### Actual behavior
1. AWS cpp SDK sends s3 request with key containing unencoded `=` (equal sign) in the key itself (not in the query string of the URL which is after the first `?`).
2. Other SDK's such as Java/JS/python all encode this as character as `%3D` but cpp SDK sends it as is in the URL.
3. However the SDK calculates the signature using the encoded URL.
4. NooBaa endpoint calculates the signature assuming it was encoded and therefore decides to reject the signature.
5. However other S3 endpoints accept this discrepency.

### Expected behavior
1. Signature should be compatible also with the discrepencies of AWS.

### Steps to reproduce
1. Use AWS CPP SDK to send a request for a key with reserved characters.
2. Can use https://github.com/noobaa/noobaa-core/tree/master/src/native/aws-cpp-sdk as a base.

### More information - Screenshots / Logs / Other output

Here is the code in AWS CPP SDK that causes this behavior -

https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/http/URI.cpp#L54-L60

I tried a quick fix that would support both the encoded URL and unencoded URL cases which seemed to work but I am not so sure if this is covering all cases, and it might be an overkill -

https://github.com/guymguym/noobaa-core/commit/7d5d120eefc6f4a972aa0aeced56f8f49fd1693e

guymguym:
@nimrod-becker can we priorities this for next available fix version?

PR #7829, opened 2024-02-18
fixed signature calculations for cpp aws-sdk

dannyzaken:
### Explain the changes
* AWS SDK for C++ sends the '=' character unencoded when it is in the "key" string but calculates the signature for the encoded '%3D' value.
* fixed `function _aws_request` to first decode the pathname (to avoid reencoding of '%' characters) and then encode the path again.
* Added signature tests for different names with special symbols (taken from [here](https://github.com/aws/aws-sdk-cpp/blob/f07aeab3d044f09d32c3b9d92b7e6dba8573fc45/src/aws-cpp-sdk-core/source/http/URI.cpp#L54-L60)). Also added test for name with `%3D` (escaped `=`). The new tests fails without the fix and pass after.

1. #7784 

### Testing Instructions:
1. try to perform an S3 operation (e.g. deleteObject is a simple example) with a key containing `=` (e.g: `delete=me`)
2. before the fix, you get a signature mismatch error.
3. after the fix, it should succeed. 

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/util/signature_utils.js
+++ b/src/util/signature_utils.js
@@ -202,8 +202,11 @@ const HEADERS_MAP_FOR_AWS_SDK = {
 function _aws_request(req, region, service) {
     const v2_signature = _.isUndefined(region) && _.isUndefined(service);
     const u = url.parse(req.originalUrl.replace(/%2F/g, '/'), true);
+    // for S3 we decode and escape the URI components to handle cpp sdk behaior, which for a few charecters (e.g. =,?$@) 
+    // it does not escape it in the request but escapes it for the signature calculations.
+    // see https://github.com/noobaa/noobaa-core/issues/7784 
     const pathname = service === 's3' ?
-        u.pathname :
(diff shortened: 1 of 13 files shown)


src/util/signature_utils.js
```diff
@@ -203,7 +203,7 @@ function _aws_request(req, region, service) {
     const v2_signature = _.isUndefined(region) && _.isUndefined(service);
     const u = url.parse(req.originalUrl.replace(/%2F/g, '/'), true);
     const pathname = service === 's3' ?
-        u.pathname :
+        u.pathname.split('/').map(c => AWS.util.uriEscape(decodeURIComponent(c))).join('/') :
```
guymguym, line 206: can you please just add a comment that explains this logix and why its needed? it's going to be hard to remember the details when reading the code later. tnx

guymguym (approved):
just one comment but looks good. great tests. tnx

nimrod-becker (approved):

PR #7836, opened 2024-02-20
Complete GCP namespace support

Neon-White:
### Explain the changes
1. This PR implements several functions (namely multipart uploads and tag support) required for proper namespace support over Google Cloud Storage.
I opted to use AWS SDK v3 for posterity, and went with barebones commands to speed things up a bit,
2. Since GCP only supports multipart uploads via the S3-compatible XML API, it is necessary to generate an HMAC key to use an S3 client with GCP. The PR also takes care of generating the HMAC key, as well as storing it in our DB.

1.  Up until now, our GCP interface supported resumable uploads but not multipart uploads - this PR addresses that gap

### Testing Instructions:
1. Use multipart upload to upload an object with tags to a GCP namespace bucket
3. Read the uploaded object and compare checksums to verify both write and read were successful
4. Read the object's tags and compare them to the ones from before the upload
5. Add additional tags to the object
6. Make sure the additional tags are present as well

I tested the PR locally with multipart uploads, as well as addition and removal of object metadata

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/namespace_gcp.js
+++ b/src/sdk/namespace_gcp.js
@@ -47,6 +47,13 @@ class NamespaceGCP {
         this.project_id = project_id;
         this.client_email = client_email;
         this.private_key = private_key;
+        this.gcs = new GoogleCloudStorage({
+            projectId: this.project_id,
+            credentials: {
+                client_email: this.client_email,
+                private_key: this.private_key,
+            }
+        });
         this.s3_client = new AWS.S3({
             endpoint: 'https://storage.googleapis.com',
             accessKeyId: hmac_key.access_id,
--- a/src/sdk/namespace_gcp.js
+++ b/src/sdk/namespace_gcp.js
@@ -11,7 +11,7 @@ const s3_utils = require('../endpoint/s3/s3_utils');
 const S3Error = require('../endpoint/s3/s3_errors').S3Error;
 // we use this wrapper to set a custom user agent
 const GoogleCloudStorage = require('../util/google_storage_wrap');
-const aws_sdk = require('aws-sdk');
+const AWS = require('aws-sdk');
 
 /**
  * @implements {nb.Namespace}
@@ -53,7 +53,7 @@ class NamespaceGCP {
             this.hmac_key = res[0];
             this.hmac_secret = res[1];
         });
-        this.s3_client = new aws_sdk.S3({
+        this.s3_client = new AWS.S3({
             endpoint: 'https://storage.googleapis.com',
             accessKeyId: this.hmac_key,
             secretAccessKey: this.hmac_secret
--- a/src/sdk/namespace_gcp.js
+++ b/src/sdk/namespace_gcp.js
@@ -27,6 +27,10 @@ class NamespaceGCP {
     *      private_key: string,
     *      access_mode: string,
     *      stats: import('./endpoint_stats_collector').EndpointStatsCollector,
+    *      hmac_key: {
+    *         access_id: string,
+    *         secret_key: string,
+    *      }
     * }} params
     */
     constructor({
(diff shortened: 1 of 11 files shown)


src/sdk/namespace_gcp.js
```diff
@@ -7,9 +7,11 @@ const util = require('util');
 const P = require('../util/promise');
 const stream_utils = require('../util/stream_utils');
 const dbg = require('../util/debug_module')(__filename);
+const s3_utils = require('../endpoint/s3/s3_utils');
 const S3Error = require('../endpoint/s3/s3_errors').S3Error;
 // we use this wrapper to set a custom user agent
 const GoogleCloudStorage = require('../util/google_storage_wrap');
+const aws_sdk = require('aws-sdk');
```
shirady, line 14: Why do you use the old SDK (v2 instead of v3) with the new code?
```suggestion
const { S3 } = require('@aws-sdk/client-s3');
```

If you have a good reason for using AWS SDK v2 please keep the same name (easier to search and aligned with all our files):
  ```suggestion
const AWS = require('aws-sdk');
```

Comment (shirady) [reply src/sdk/namespace_gcp.js]: In my opinion, it is better to write new code with the new AWS SDK (or at least try...).
In this PR you're using an s3 client with basic properties, and specific functions (`createMultipartUpload`, `uploadPart`, `completeMultipartUpload`, `abortMultipartUpload`).
This code would be tested so it will be easier to handle the bugs (shorter cycle).

@liranmauda @dannyzaken What do you think?

Comment (nimrod-becker) [reply src/sdk/namespace_gcp.js]: Need to consider this is done on top of GCP. We know there is compatibility with the previous SDK, and there might be with the new versions as well, but we do need to keep it in mind

File: src/sdk/namespace_gcp.js [hunk 2]
```diff
@@ -289,27 +320,133 @@ class NamespaceGCP {
...
+            res = await this.s3_client.send(
+                new UploadPartCopyCommand({
+                    Bucket: this.bucket,
+                    Key: params.key,
+                    UploadId: params.obj_id,
+                    PartNumber: params.num,
+                    CopySource: copy_source,
+                    CopySourceRange: copy_source_range,
+            }));
+            etag = s3_utils.parse_etag(res.CopyPartResult.ETag);
```
tangledbytes, line 355: We can do an early exit here if we move the `return { tag }` here. Will remove the `else` branch.

src/sdk/namespace_gcp.js
```diff
@@ -324,12 +461,13 @@ class NamespaceGCP {
     */
     async get_object_acl(params, object_sdk) {
         dbg.log0('NamespaceGCP.get_object_acl:', this.bucket, inspect(params));
-        throw new S3Error(S3Error.NotImplemented);
+        await this.read_object_md(params, object_sdk);
```
tangledbytes, line 464: So ACL APIs are essentially no-ops?

tangledbytes: I guess yes.

src/sdk/namespace_gcp.js
```diff
@@ -47,6 +64,20 @@ class NamespaceGCP {
                 private_key: this.private_key,
             }
         });
+        /* An S3 client is needed for multipart upload since GCP only supports multipart uploads via the S3-compatible XML API
+        *  https://cloud.google.com/storage/docs/multipart-uploads
+        *  This is also the reason an HMAC key is generated as part of `add_external_connection` - since the standard GCP credentials
+        *  cannot be used in conjunction with the S3 client.
+        */
+        this.s3_client = new S3Client({
```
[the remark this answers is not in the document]
tangledbytes: Yeah, the public interface is pretty damn constraint. `https://github.com/googleapis/nodejs-storage/blob/main/src/transfer-manager.ts#L180` they do expose the granular controls as well BUT they mention it is not intended for public use (then why export it? 🤦).

Yeah, you are right I guess. Let's stick to S3 ones.

liranmauda:
Keep active...

PR #7837, opened 2024-02-20
bucket_logging: Bucket log processing in side car container

aspandey:
Firstly,
rsyslogd and logrotate.sh script has been moved to side car container. Now, these two processes will be started by "command" rather than superviserd

Secondly,
we have added following bash script and changes in logrotate_noobaa.conf 1 - Added an entry in conf file to rotate /var/log/bucket_logs.log

2 - Rotated files will be moved to /var/log/noobaa_logs/ and each rotated file will have dat and time attached to its name. Example: bucket_logs.log2024-02-19-10-09-16

3 - These files will have logs for all the buckets from all the endpoints.

4 - Added a bash script noobaa_log_segregate.sh, which will keep reading rotated files in /var/log/noobaa_logs/ and segregate the bucket logs in its respective log file names. These will be kept at /log/noobaa_bucket_logs/

For example:
log.bucket-1-2024-02-19-10-05-58.341870679.log
log.bucket-2-2024-02-19-09-49-37.242243366.log

5 - These are the final log objects which will be sent to respective bucket logs.

### Testing Instructions:

./noobaa_logs_create.sh <log_bucket_range>
This script can be used to test log parsing. It takes number of buckets,
which you want to have logs from. Logs will be sent to /var/log/bucket_logs.log,
which is actually the path where rsyslogd will save the logs. Log buckets will be
having name like log.bucket-1, log.bucket-2, etc. So, to test if the BG worker is
uploading the logs correctly, you need to create buckets with similar names.
Example: ./noobaa_logs_create.sh 10

### DIFF
--- a/src/deploy/NVA_build/NooBaa.Dockerfile
+++ b/src/deploy/NVA_build/NooBaa.Dockerfile
@@ -74,7 +74,8 @@ RUN dnf install -y -q bash \
     python3-setuptools \
     jemalloc \
     xz \
-    python3-pip && \
+    python3-pip \
+    cronie && \
     dnf clean all
 
 COPY ./src/deploy/NVA_build/install_arrow_run.sh ./src/deploy/NVA_build/install_arrow_run.sh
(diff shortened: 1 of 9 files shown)


dannyzaken (commented):
@aspandey we should not have two instances of rsyslog in the core pod. the log container should handle both the core logs and the bucket logging

aspandey:
@dannyzaken You are right.
I left supervisord conf file as it is just to make sure that this change works even when PR for operator (which create side car container) is not there.
I will remove the rsyslog and logrotate from its conf file.

dannyzaken (approved):

PR #7845, opened 2024-02-23
Fix cross platform RPM builds

b-ranto:
### Explain the changes
1. This PR restructures how we do the RPM builds today.

What changed?

 *  Instead of fetching NooBaa runtime dependencies in the Dockerfile, we fetch them in the RPM build process.
 * We make sure the environment has all the dependencies that NooBaa requires for the build process

Why changed?

 *  This approach allows generating a true architecture independent SRPM, hence a SRPM can be generated once (per RHEL version) and then rpmbuild --rebuild <rpm> can be used to generate RPMS for different architectures.
 * This approach allows us to use `mock` to build the RPMs

**Note**: Usage remains same, that is use, make rpm. This will generate a SRPM and a RPM. Use the SRPM to generate the RPMs for other architectures. The command also supports make rpm SRPM_ONLY=true which is significantly faster and jsut generates SRPMs.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,10 +165,10 @@ nbdev:
 	@echo ""
 .PHONY: nbdev
 
-rpm: base
+rpm: builder
 	echo "\033[1;34mStarting RPM build for $${CONTAINER_PLATFORM}.\033[0m"
 	mkdir -p build/rpm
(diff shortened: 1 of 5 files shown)


b-ranto:
FYI: I had to disable `BUILD_S3SELECT` to get this to pass when testing locally.

b-ranto:
FYI: I had to disable `BUILD_S3SELECT` to get this to pass when testing locally.

@tangledbytes I have invited you to my fork of the repo, you should be able to update this PR if necessary.

PR #7848, opened 2024-02-26
[NC | NSFS] Fix WAL processing failure

tangledbytes:
### Explain the changes
This PR attempts to fix WAL processing. This issue was identified with the help of Abe-san. It was noticed that PersistentLogger's processor was trying to consume the timestamp file as well because of sharing the same prefix. This PR puts the `timestamp` files into the `timestamp` "namespace" such that the processor will not confuse that file with a WAL.

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/manage_nsfs/manage_nsfs_help_utils.js
+++ b/src/manage_nsfs/manage_nsfs_help_utils.js
@@ -184,23 +184,19 @@ Flags:
 
 const GLACIER_OPTIONS = `
 Usage:
-    manage_nsfs glacier <migrate | restore | expiry> [options]
+glacier <action> [flags]
+
+List of actions supported:
+migrate
+restore
+expiry
 `;
 
-const GLACIER_MIGRATE_OPTIONS = `
-Glacier Migrate Options:
-    --interval <interval>                         (default none)            Run the operation if "interval" milliseconds have passed since last run
-`;
+const GLACIER_MIGRATE_OPTIONS = ``;
 
-const GLACIER_RESTORE_OPTIONS = `
-Glacier Restore Options:
-    --interval <interval>                         (default none)            Run the operation if "interval" milliseconds have passed since last run
-`;
+const GLACIER_RESTORE_OPTIONS = ``;
 
-const GLACIER_EXPIRY_OPTIONS = `
-Glacier Expiry Options:
-    --interval <interval>                         (default none)            Run the operation if "interval" milliseconds have passed since last run
-`;
+const GLACIER_EXPIRY_OPTIONS = ``;
 
 /** 
  * print_usage would print the help according to the arguments that were passed
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -647,7 +647,7 @@ struct AutoCloser
         : _worker(worker), _fd(fd) {}
     ~AutoCloser()
     {
-        if (_fd) {
+        if (_fd >= 0) {
             int r = close(_fd);
             if (r) _worker->SetSyscallError();
         }
@@ -1341,7 +1341,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
     }
     FileWrap(const Napi::CallbackInfo& info)
         : Napi::ObjectWrap<FileWrap>(info)
-        , _fd(0)
+        , _fd(-1)
     {
     }
     ~FileWrap()
@@ -1350,7 +1350,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
             LOG("FS::FileWrap::dtor: file not closed " << DVAL(_path) << DVAL(_fd));
             int r = ::close(_fd);
             if (r) LOG("FS::FileWrap::dtor: file close failed " << DVAL(_path) << DVAL(_fd) << DVAL(r));
-            _fd = 0;
+            _fd = -1;
         }
     }
     Napi::Value close(const Napi::CallbackInfo& info);
@@ -1376,7 +1376,7 @@ struct FileOpen : public FSWorker
     mode_t _mode;
     FileOpen(const Napi::CallbackInfo& info)
         : FSWorker(info)
-        , _fd(0)
+        , _fd(-1)
         , _flags(0)
         , _mode(0666)
     {
@@ -1417,10 +1417,12 @@ struct FileClose : public FSWrapWorker<FileWrap>
     virtual void Work()
     {
         int fd = _wrap->_fd;
-        std::string path = _wrap->_path;
-        int r = close(fd);
-        if (r) SetSyscallError();
-        _wrap->_fd = 0;
+        if (fd >= 0) {
+            std::string path = _wrap->_path;
+            int r = close(fd);
+            if (r) SetSyscallError();
+            _wrap->_fd = -1;
+        }
     }
 };
 
@@ -1570,6 +1572,7 @@ struct FileReplacexattr : public FSWrapWorker<FileWrap>
     virtual void Work()
     {
         int fd = _wrap->_fd;
+        CHECK_WRAP_FD(fd);
 
         if (_prefix != "") {
             SYSCALL_OR_RETURN(clear_xattr(fd, _prefix));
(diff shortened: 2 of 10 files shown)


src/util/file_reader.js
```diff
@@ -97,7 +97,7 @@ class NewlineReader {
             if (this.lock) await fh.flock(this.fs_context, this.lock);
 
             this.fh = fh;
-        } finally {
+        } catch {
```
guymguym, line 100: First - is this valid try-catch syntax? I think it needs to have `catch (err) {...}`.
Second - do we mean to "swallow" this error? shouldn't we rethrow it up to the caller?

src/sdk/nsfs_glacier_backend/backend.js
```diff
@@ -4,8 +4,14 @@
 const nb_native = require('../../util/nb_native');
 
 class GlacierBackend {
+    // These names start with the word 'timestamp' so as to assure
+    // that it acts like a 'namespace' for the these kind of files.
+    //
+    // It also helps in making sure that the persistent logger does not
+    // confuses these files with the WAL files.
```
guymguym, line 11: where is that filtering of timestamp files being done?

src/util/file_reader.js
tangledbytes: Yes, the syntax is valid - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch. `exceptionVar` is optional.

But I think yeah, maybe we should bubble up the error.

src/sdk/nsfs_glacier_backend/backend.js
tangledbytes: In `PersistentLogger` `process_inactive` iterates over all the directory entries and filters the files based on the prefix. We have the a `PersistentLogger` named `migrate` so that instance selects all the files which start with `migrate...`

src/native/fs/fs_napi.cpp
```diff
@@ -1350,7 +1350,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
             LOG("FS::FileWrap::dtor: file not closed " << DVAL(_path) << DVAL(_fd));
             int r = ::close(_fd);
             if (r) LOG("FS::FileWrap::dtor: file close failed " << DVAL(_path) << DVAL(_fd) << DVAL(r));
-            _fd = 0;
+            _fd = -1;
```
guymguym, line 1353: need to fix the condition in the dtor from if (_fd) to if(_fd >= 0) ...
Also fix the same condition in AutoCloser.
Also I would add a check to FileClose, so if (_fd >= 0) it can close it, otherwise ignore (already closed or not opened).
Also I see that FileReplacexattr is missing the CHECK_WRAP_FD(fd) call.
Also I see that LinkFileAt and UnlinkFileAt initialize _delete_fd/_replace_fd to 0 and better initialize to -1 and check accordingly.

src/sdk/nsfs_glacier_backend/backend.js
guymguym: to be honest this is a bit accidental. I would prefer a more clear separation of which files are logs and which files are not. why not use the suffix `.log` to filter them? I also think that a file that is named `timestamp.expiry` appears to have a suffix of `.expiry` which is a bit meaningless as a suffix, so I would rather have it with a `.timestamp` suffix still, and filter the logs based on their suffix. WDYT?

src/sdk/nsfs_glacier_backend/tapecloud.js
```diff
@@ -181,7 +181,7 @@ class TapeCloudGlacierBackend extends GlacierBackend {
             const [precount, preres] = await walreader.forEach(async entry => {
                 let fh = null;
                 try {
-                    fh = await nb_native().fs.open(fs_context, entry, 'rw');
+                    fh = await nb_native().fs.open(fs_context, entry, 'rw*');
```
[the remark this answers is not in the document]
tangledbytes: I don't think opening the file by default would be better. This piece of code opens the file but `migrate` doesn't so we would be opening each file in the log for no reason. Instead I propose making `entry` an instance of a class which exposes methods like `open`, `read`, `path`, etc.? WDYT?

src/sdk/nsfs_glacier_backend/backend.js
tangledbytes: I can see why it might be a little strange to have `expiry` as a suffix when looked as files. I actually just viewed them as key-value pairs, hence my usage of the term `namespace`.

Yeah, we can enforce that each log file must have the prefix `.log`. But that in itself won't be enough. We still need to filter out the files that are meant for **this** `PersistentLogger` instance. So maybe something like `<timestamp>.<type>.log` could be the format. The filtering could be for `<type>.log`. How does that sound?

src/sdk/nsfs_glacier_backend/tapecloud.js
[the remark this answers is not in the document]
guymguym: Sure, sounds like a reasonable api

src/manage_nsfs/manage_nsfs_help_utils.js
```diff
@@ -185,23 +185,19 @@ Flags:
 
 const GLACIER_OPTIONS = `
 Usage:
-    manage_nsfs glacier <migrate | restore | expiry> [options]
-`;
+glacier <action> [options]
```
shirady, line 188: minor comment - to match the style of the help.
```suggestion
glacier <action> [flags]
```

src/sdk/nsfs_glacier_backend/backend.js
```diff
@@ -149,6 +155,8 @@ class GlacierBackend {
             stat.xattr[GlacierBackend.XATTR_RESTORE_REQUEST_STAGED]) {
             return false;
         }
+
+        return true;
```
guymguym, line 159: lets add a check that if number of blocks is 0 we also filter out this file.

src/native/fs/fs_napi.cpp
```diff
@@ -1597,12 +1600,15 @@ struct LinkFileAt : public FSWrapWorker<FileWrap>
...
     {
+        int fd = _wrap->_fd;
+        CHECK_WRAP_FD(fd);
+
         // gpfs_linkat() is the same as Linux linkat() but we need a new function because
         // Linux will fail the linkat() if the file already exist and we want to replace it if it existed.
-        if (_replace_fd == 0) {
-            SYSCALL_OR_RETURN(dlsym_gpfs_linkat(_wrap->_fd, "", AT_FDCWD, _filepath.c_str(), AT_EMPTY_PATH));
+        if (_replace_fd == -1) {
+            SYSCALL_OR_RETURN(dlsym_gpfs_linkat(fd, "", AT_FDCWD, _filepath.c_str(), AT_EMPTY_PATH));
```
guymguym, line 1609: is it really valid to use fd=-1 in this call ?? seems like an error unless we explain why it's correct

guymguym: sorry my bad - this fd is not the same as _replace_fd. resolving this comment.

src/sdk/nsfs_glacier_backend/tapecloud.js
```diff
@@ -133,16 +132,22 @@ class TapeCloudGlacierBackend extends GlacierBackend {
...
 
             // If the result of the above is false then it indicates that we concluded
             // to exit early hence the file shouldn't be processed further, exit
             if (!result) return false;
 
+            // If we didn't read even one line then it most likely indicates that the WAL is
+            // empty - this case is unlikely given the mechanism of WAL but still needs to be
+            // handled.
+            // Still return `false` so as to not insist file deletion
+            if (processed === 0) return false;
```
guymguym, line 147: Ok, so this file will keep being read every time we process migrations?
I think it's best to clean up here.
If you are worried about this case hiding some other bug so lets add a log warn message that we can look out for.
In this case you can also print some more info about the file - like it's stat attributes (size, dates, etc).
But I don't think that leaving it is better than not.

guymguym (approved):

PR #7846, opened 2024-02-26
namespace_monitor.js | Fixing the error object values to start with a lowercase

liranmauda:
### Explain the changes
When using aws sdk v3  we get the error object with values that start with an uppercase instead of lowercase we want to fix this value to start with a lowercase.

### BUGS:
1. Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2265109

PR #7860, opened 2024-03-03
Migrating to AWS SDK V3 | Following PR on a Fix That Was Added

shirady:
#### Background:
In #7846 revealed a bug that we had in the code - when using the AWS SDK v3 the `error` properties were changed to uppercase, so instead of `err.code` it was `err.Code`. We had a function `check_error_code`, but if the catch is using `throw err` and the caller is looking for `err.code` - it will not find it. Hence whenever we use the `noobaa_s3_client` and in the catch clause we have `throw err` we need to add the function `fix_error_object` so we will have the `code` property.

The changes are:
1. Use `noobaa_s3_client.fix_error_object(err)` in catch clause whenever there is `throw err`.
2. Remove the function `check_error_code` and its usages.

1. none

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/bucketspace_s3.js
+++ b/src/sdk/bucketspace_s3.js
@@ -35,6 +35,7 @@ class BucketSpaceS3 {
             const buckets_with_name_change = buckets.map(b => ({ name: new SensitiveString(b.Name) }));
             return { buckets_with_name_change };
         } catch (err) {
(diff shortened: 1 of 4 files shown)


liranmauda (approved):
LGTM

Issue #7701, opened 2024-01-05
NSFS | NC| CLI | noobaa node command taking longer time to exit


PravinRanjan10:
### Environment info

* NooBaa Version: noobaa-core-5.15.0-20240102.el8.x86_64.rpm

### Actual behavior
1. All node commands taking longer time to exit after printing the status

### Expected behavior
1. it should exit immediately after printing the status

### Steps to reproduce
1. run any node command

naveenpaul1:
intermittent issue, try to reproduce it in @PravinRanjan10  machine but could not reproduce it, will try again once it is reproducible in his system.

romayalon:
@PravinRanjan10 @naveenpaul1 do you still see this issue? or can we close this bug?

naveenpaul1:
Did a test with @PravinRanjan10 in his rhel 9 and 8 machine, We could reproduce the issue with rhel8 but not with rhel9, And confirm that it is an issue with rhel8, I am trying to fix this in rhel8 with Pravin.

romayalon:
@PravinRanjan10 @naveenpaul1 I tried it with @udaysk23 (Thanks Uday!) on a RHEL8 VM (noobaa-core-5.15.0-20240207.el8.x86_64.rpm upstream) and we couldn't reproduce it..

PravinRanjan10:
This issue is still reproducible on our cluster of rhel8. @naveenpaul1 is looking into it. FYI @romayalon

naveenpaul1:
Hi @PravinRanjan10 we identified a possible cause for this issue and the fix would be to disable random seed read.
A temporary workaround for disabling random seed read is adding env variable `DISABLE_DEV_RANDOM_SEED` to true for each manage nsfs command. Something similar to below.
`DISABLE_DEV_RANDOM_SEED='true' usr/local/noobaa-core/bin/node usr/local/noobaa-core/src/cmd/manage_nsfs account list`
This will be a temporary solution for setting env `DISABLE_DEV_RANDOM_SEED`, We will come up with a proper solution ASAP. 
special thanks to @guymguym @romayalon

romayalon:
@PravinRanjan10 @naveenpaul1 
The issue was fixed on #7864, @PravinRanjan10 can you please validate it?
The fix is currently on master, and should be backported to 4.15.1 in the near future

PravinRanjan10:
[...]

PR #7864, opened 2024-03-05
NSFS | NC | Add disable dev random seed flag

naveenpaul1:
### Explain the changes
1. add configuration for disabling random seed for CLI, the default value is `true`, new flag added `DISABLE_CLI_RANDOM_SEED`
2. `DISABLE_DEV_RANDOM_SEED` flag value will be fetched from config if no value is set on `process.env.DISABLE_DEV_RANDOM_SEED`. And the default value would be `false`

1. https://github.com/noobaa/noobaa-core/issues/7701

### Testing Instructions:
1. update `DISABLE_CLI_RANDOM_SEED` and `DISABLE_DEV_RANDOM_SEED` to check whether random seed is enabled or not.

- [X] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -96,6 +96,7 @@ config.AGENT_TEST_CONNECTION_TIMEOUT = 1 * 60 * 1000;
 config.STORE_PERF_TEST_INTERVAL = 60 * 60 * 1000; // perform test_store_perf every 1 hour
 config.CLOUD_MAX_ALLOWED_IO_TEST_ERRORS = 3;
 
+config.ENABLE_DEV_RANDOM_SEED = process.env.DISABLE_DEV_RANDOM_SEED === 'false' || false;
 
 ////////////////
 // RPC CONFIG //
--- a/docs/dev_guide/NonContainerizedDeveloperCustomizations.md
+++ b/docs/dev_guide/NonContainerizedDeveloperCustomizations.md
@@ -320,6 +320,23 @@ Example:
 ```
 
 
+## 18. Disable random seeding -
+**Description -** This flag will enable the random seeding for the application.
+
+**Configuration Key -** ENABLE_DEV_RANDOM_SEED
+
+**Type -** boolean
+
+**Default -**  false
+**Steps -**
+```
+1. Open /path/to/config_dir/config.json file.
+2. Set the config key -
+Example:
+"ENABLE_DEV_RANDOM_SEED": false
+3. systemctl restart noobaa_nsfs
+```
+
 ## Config.json example 
 ```
 > cat /path/to/config_dir/config.json
--- a/src/util/nb_native.js
+++ b/src/util/nb_native.js
@@ -8,6 +8,7 @@ const events = require('events');
 const chance = require('chance')();
 const bindings = require('bindings');
 const child_process = require('child_process');
+const config = require('../../config');
 
 const async_exec = util.promisify(child_process.exec);
 const async_delay = util.promisify(setTimeout);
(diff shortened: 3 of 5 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -7,8 +7,9 @@ const path = require('path');
 const minimist = require('minimist');
 const net = require('net');
 const config = require('../../config');
+// disable random seed only for CLI
+process.env.DISABLE_DEV_RANDOM_SEED = config.DISABLE_CLI_RANDOM_SEED;
```
romayalon, line 11: do we need to add this line also on health.js?

docs/dev_guide/NonContainerizedDeveloperCustomizations.md
```diff
@@ -320,6 +320,40 @@ Example:
 ```
 
 
+## 18. Disable random seeding -
```
romayalon, line 323: should we add these new configs to nsfs_config_schema?
and can we add a short test to 
`noobaa-core/src/test/unit_tests/jest_tests/test_nc_nsfs_config_schema_validation.test.js` ?

src/util/nb_native.js
```diff
@@ -82,7 +83,8 @@ async function read_rand_seed(seed_bytes) {
     while (offset < buf.length) {
         try {
             const count = buf.length - offset;
-            const random_dev = process.env.DISABLE_DEV_RANDOM_SEED ? '/dev/urandom' : '/dev/random';
+            if (process.env.DISABLE_DEV_RANDOM_SEED === undefined) process.env.DISABLE_DEV_RANDOM_SEED = config.DISABLE_DEV_RANDOM_SEED;
+            const random_dev = process.env.DISABLE_DEV_RANDOM_SEED === 'true' ? '/dev/urandom' : '/dev/random';
```
romayalon, line 87: Will this work on containerized NooBaa as well?
I'm asking because the previous check was -
`const random_dev = process.env.DISABLE_DEV_RANDOM_SEED ? '/dev/urandom' : '/dev/random';`
and now it was changed to -
`rocess.env.DISABLE_DEV_RANDOM_SEED = config.DISABLE_DEV_RANDOM_SEED;
            const random_dev = process.env.DISABLE_DEV_RANDOM_SEED === 'true' ? '/dev/urandom' : '/dev/random';`
so we no longer check for boolean value but for string 'true'

naveenpaul1: @romayalon ` process.env.DISABLE_DEV_RANDOM_SEED` is a type of `string`,  This condition `process.env.DISABLE_DEV_RANDOM_SEED ?` always  return `true` when  `DISABLE_DEV_RANDOM_SEED` is not empty, So even if `process.env.DISABLE_DEV_RANDOM_SEED` is `false`, This condition will assign `/dev/urandom` to random_dev.

romayalon: I was thinking about a situation where - 
process.env.DISABLE_DEV_RANDOM_SEED = 'false';
Before the change- 
random_dev = '/dev/urandom' ;
After - 
random_dev = '/dev/random';

romayalon: @naveenpaul1 but I'm starting to think that before the change it was a bug...

config.js
```diff
@@ -97,6 +97,8 @@ config.STORE_PERF_TEST_INTERVAL = 60 * 60 * 1000; // perform test_store_perf eve
 config.CLOUD_MAX_ALLOWED_IO_TEST_ERRORS = 3;
 
 
+config.DISABLE_DEV_RANDOM_SEED = 'false';
```
romayalon, line 100: @naveenpaul1 @guymguym 
Guy suggested to convert the config to - 
config.ENABLE_DEV_RANDOM_SEED = 'false'; 
because we always (on containerized and non containerized) won't enable it.
@naveenpaul1 WDYT? this will make the code on manage_nsfs.js and health.js redundant,
and we should adjust the nb_native.js code

romayalon (approved):

PR #7867, opened 2024-03-05
Support delete replication flow in bucket_diff - PR 2

aayushchouhan09:
### Explain the changes
1. This is the first PR in support for the delete replication. In this PR we are identifying the keys needs to be deleted for versioning enabled buckets in replication flow inside bucket_diff.

### Testing Instructions:
1. use command : `npx jest test_bucket_diff.test.js`

- [x] Tests added

### DIFF
--- a/src/server/utils/bucket_diff.js
+++ b/src/server/utils/bucket_diff.js
@@ -18,6 +18,7 @@ class BucketDiff {
      *   s3_params?: AWS.S3.ClientConfiguration
      *   connection?: AWS.S3
      *   for_replication: boolean
+     *   for_deletion: boolean
      * }} params
      */
     constructor(params) {
@@ -28,6 +29,7 @@ class BucketDiff {
             s3_params,
             connection,
             for_replication,
+            for_deletion,
         } = params;
         this.first_bucket = first_bucket;
         this.second_bucket = second_bucket;
@@ -42,6 +44,8 @@ class BucketDiff {
         }
         // special treatment when we want the diff for replication purpose.
         this.for_replication = for_replication;
+        // will set the bucket diff to return only the diff of delete markers.
+        this.for_deletion = for_deletion;
     }
 
     /**
@@ -62,10 +66,17 @@ class BucketDiff {
 
         const diff = {
             // keys_diff_map is {{ [key: string]: Array<object> }} where the Array consist of metadata objects.
-            // for example: {
+            // for example(disabled for_deletion): {
             //     "1": [
             //         { ETag: 'etag1.1', Size: 24599, Key: '1', VersionId: 'v1.1', IsLatest: true, },
             //         { ETag: 'etag1.2', Size: 89317, Key: '1', VersionId: 'v1.2', IsLatest: false, }
+            //     ]
+            // }
+            //
+            // for example(enabled for_deletion): {
+            //     "1": [
+            //         { Key: '1', VersionId: "v1.2", IsLatest: true, LastModified: '2022-02-27T10:45:00.000Z', },
+            //         { Key: '1', VersionId: "v1.1", IsLatest: false, LastModified: '2022-01-27T10:45:00.000Z', }
             //     ]
             // }
             keys_diff_map: {},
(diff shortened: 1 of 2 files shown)


src/server/utils/bucket_diff.js
```diff
@@ -165,9 +168,12 @@ class BucketDiff {
      * @returns {nb.BucketDiffKeysDiff}
      */
     _object_grouped_by_key_and_omitted(list) {
-        if (!list) return {};
+        if (!list || (this.for_deletion && !this.version)) return {};
```
liranmauda, line 171: Let's add a comment here.
also, maybe we want to return an error on this setup (deletion + non version)

src/server/utils/bucket_diff.js
```diff
@@ -165,9 +168,12 @@ class BucketDiff {
...
         dbg.log1('_object_grouped_by_key_and_omitted list:', list);
-        const field = this.version ? "Versions" : "Contents";
+        let field = this.version ? "Versions" : "Contents";
+        if (this.for_deletion && this.version) {
```
liranmauda, line 174: Isn't it redundant with `if (!list || (this.for_deletion && !this.version)) return {};` ?
If we are here and` if (this.for_deletion)`, then `this.version` must be true.
```suggestion
        if (this.for_deletion) {
```

src/server/utils/bucket_diff.js
```diff
@@ -454,5 +470,16 @@ class BucketDiff {
...
+    /**
+     * @param {any[]} bucket_obj
+     */
+    _get_latest_delete_marker(bucket_obj) {
+        // Extract the latest delete marker from the bucket
+        const latest_delete_marker = bucket_obj.reduce(function(prev, current) {
+            return (new Date(prev.LastModified) > new Date(current.LastModified)) ? prev : current;
+        });
+        return latest_delete_marker;
+    }
```
liranmauda, line 482: I think we do not need that.

src/server/utils/bucket_diff.js
```diff
@@ -40,8 +42,9 @@ class BucketDiff {
             if (!s3_params) throw new Error('Expected s3_params');
             this.s3 = new AWS.S3(s3_params);
         }
-        // special treatment when we want the diff for replication purpose.
+        // special treatment when we want the diff for replication/deletion purpose.
         this.for_replication = for_replication;
+        this.for_deletion = for_deletion;
```
liranmauda, line 47:
```suggestion
        // special treatment when we want the diff for replication purposes.
        this.for_replication = for_replication;
        // will set the bucket diff to return only a diff of delete markers.
        this.for_deletion = for_deletion;
```

liranmauda (changes requested):

src/server/utils/bucket_diff.js
achouhan09: Sure, will update it.

src/server/utils/bucket_diff.js
achouhan09: Will update the comment. Thanks

src/server/utils/bucket_diff.js
achouhan09: You mean we need not to fetch latest delete marker in different function? We should do it in the code itself instead of using different function, right?

src/server/utils/bucket_diff.js
achouhan09: Sure, will do this.

src/server/utils/bucket_diff.js
achouhan09: @liranmauda as per this aws s3 [api](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteMarkerEntry.html) 

Delete markers only contains fields - `IsLatest`, `Key`, `LastModified`, `VersionID`, `Owner`. And it doesn't contain the `etag` field, so to identify the keys on the basis of etags similar to version replication is not possible in delete replication. We cannot do it even with the `VersionID` as it can contain two different objects in different buckets.
This code here will get the diff keys from delete markers on the basis of `LastModified` and `IsLatest`. So, with this we are only able to replicate the latest delete markers.

src/server/utils/bucket_diff.js
```diff
@@ -454,5 +486,16 @@ class BucketDiff {
         return target_pos;
     }
 
+    /**
+     * @param {any[]} bucket_obj
+     */
+    _get_latest_delete_marker(bucket_obj) {
+        // Extract the latest delete marker from the bucket
+        const latest_delete_marker = bucket_obj.reduce(function(prev, current) {
+            return (new Date(prev.LastModified) > new Date(current.LastModified)) ? prev : current;
```
liranmauda, line 495: Why do we need to check for the date? 
why not simply check the `IsLatest` field?

liranmauda (approved):

PR #7879, opened 2024-03-10
NSFS | NC | Use `--from_file` Option in Bucket Create

shirady:
### Explain the changes
1. Add the `from_file` option in the bucket that would include validation. The JSON file will include options (as key-value structure) that are the same as the flags we suggest.
2. Add a special case of `bucket_policy` as an object in the JSON file.
3. Edit documentation of `from_file` with an example of bucket creation in addition to account creation.
4. Edit comments in file `test_nc_nsfs_account_cli.test.js` of typo (‘tha’ instead of ‘the’) and remove the ‘and check the details’ in case the tests should fail.
6. In Add bucket tests to file `test_nc_nsfs_bucket_cli.test.js`:
- Add bucket tests using `from_file` option, and copy functions from the account test file.
- Edit the tests not to have a global variable of `bucket_storage_path`, and `bucket_temp_dir_path` but local in the `describe` scope (else it creates issues since Jest running the tests in concurrency)

### Issues:
Currently, we had a gap in `from_file` when adding a new bucket.

### Testing Instructions:
Added unit tests, please run: `sudo npx jest test_nc_nsfs_bucket_cli.test.js`

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -103,9 +103,7 @@ async function main(argv = minimist(process.argv.slice(2))) {
         if (type === TYPES.ACCOUNT) {
             await account_management(action, user_input);
         } else if (type === TYPES.BUCKET) {
-            // GAP - from-file in bucket
-            const path_to_json_options = argv.from_file ? String(argv.from_file) : '';
-            await bucket_management(argv, path_to_json_options);
+            await bucket_management(action, user_input);
         } else if (type === TYPES.IP_WHITELIST) {
             await whitelist_ips_management(argv);
         } else if (type === TYPES.GLACIER) {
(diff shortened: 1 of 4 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -933,6 +927,10 @@ function validate_options_type_by_value(input_options_with_data) {
             if (['allow_bucket_creation', 'regenerate', 'wide', 'show_secrets'].includes(option) && validate_boolean_string_value(value)) {
                 continue;
             }
+            // special case for bucket_policy (from_file)
+            if (option === 'bucket_policy' && type_of_value === 'object') {
```
romayalon, line 931: so this adds a new type for bucket policy? now bucket policy is string | object?
If so, this should be added to consts & help file instead of as a special case here

shirady: It is just for `from_file` option to accept another type, since we use a JSON file, we don't need it to be a string but can be an object.
In all cases, we set one type:
https://github.com/noobaa/noobaa-core/blob/d1bf8c465af745dedbff9d3803ec13da344884ba/src/manage_nsfs/manage_nsfs_constants.js#L69-L90

But may accept more types in the function `validate_options_type_by_value` (`name`, boolean types that we accept `true` and `false`, etc.):
https://github.com/noobaa/noobaa-core/blob/d1bf8c465af745dedbff9d3803ec13da344884ba/src/cmd/manage_nsfs.js#L929-L935

romayalon: @shirady Yes, exactly the same as with the booleans/string booleans.
My point is that -
We have a list of constant types for the CLI arguments but it's not full, we add more special types on several cases on this function..
We need to consider refactoring this function since the missing types on the types list is confusing

romayalon (approved):

Issue #7873, opened 2024-03-07
NC | NSFS | Creating an Account with an Existing Access Key Doesn’t Return `Accountaccesskeyalreadyexists` Error


shirady:
### Environment info
- NooBaa Version: master
- Platform: NSFS NC

### Actual behavior
1. Creating an account with an existing access key returns `InternalError`:
```json
{
  "error": {
    "code": "InternalError",
    "message": "The server encountered an internal error. Please retry the request",
    "detail": "Error: Too many levels of symbolic links"
  }
}
```

### Expected behavior
1. Creating an account with an existing access key should return `Accountaccesskeyalreadyexists`.

### Steps to reproduce
1. Create an account using `access_key` and `secret_key` flags: `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path <path> --access_key <access_key> --secret_key <secret_key> --uid <uid> --gid <gid>`.
2. Create another account and use the same values for `access_key` and `secret_key` flags: `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path <path> --access_key <same_access_key> --secret_key <same_secret_key> --uid <uid> --gid <gid>`.

### More information - Screenshots / Logs / Other output
According to the code we would expect this:
https://github.com/noobaa/noobaa-core/blob/8363cf721afdc2968a324b85751bf9f4717c11ce/src/cmd/manage_nsfs.js#L465-L471

I think the root cause is here:
https://github.com/noobaa/noobaa-core/blob/8363cf721afdc2968a324b85751bf9f4717c11ce/src/util/native_fs_utils.js#L477-L491
We can add `if (err.code === 'ELOOP') return true;` in the catch clause.

romayalon:
@shirady Have you tried it on linux? 
As far as I know it happens only on mac

shirady:
@romayalon I tested it on my Mac, but I haven't tried it on Linux yet.
Would we solve it for Mac? (it seems link a minor change) so we can enjoy all passing tests in `sudo node ./node_modules/.bin/_mocha src/test/unit_tests/test_nc_nsfs_cli.js` in local run.

romayalon:
@shirady Yes, just wanted to mention the Mac vs Linux because if it happens on Linux it's a blocker, else it's a lower priority bug.
I saw it also when running it,  and I thought that we will handle it on the id/email replacement, but if it's a small fix, so it'll be great to fix it before that!

PR #7882, opened 2024-03-10
NSFS | Fix lstat on mac

romayalon:
### Explain the changes
1. Running stat(fs_context, path, { use_lstat: true }) on MAC, where path is a symbolic link threw ELOOP error. According to [mac docs of open()] ELOOP described as the following - 
```
[ELOOP]            Too many symbolic links
[ELOOP]            O_NOFOLLOW was specified and the target is a symbolic link.
```

2. According to [mac docs of open()](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/open.2.html) -
```
If O_NOFOLLOW is used in the mask and the target file passed to open() is
a symbolic link then the open() will fail.

If O_SYMLINK is used in the mask and the target file passed to open() is
a symbolic link then the open() will be for the symbolic link itself, not
what it links to.
```

Therefore, the O_NOFOLLOW flag should not be a part of the flags passed when calling open() of a symbolic link file and O_SYMLINK is enough.

3. Fixed the lstat test to run lstat on a symbolic link instead of a regular file.

1. Fixed #https://github.com/noobaa/noobaa-core/issues/7873

### Testing Instructions:
```
Automatic tests - 
On MAC, run the following tests and expect to pass - 
* sudo NC_CORETEST=true node  ./node_modules/mocha/bin/mocha noobaa-core/src/test/unit_tests/test_bucketspace.js
* sudo node ./node_modules/.bin/_mocha noobaa-core/src/test/unit_tests/test_nb_native_fs.js

Manual test using manage_nsfs.js - 
1. Create an account name=account1.
2. Create another account, name=account2, same access_key as of account1.
```
- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/test/unit_tests/test_nb_native_fs.js
+++ b/src/test/unit_tests/test_nb_native_fs.js
@@ -1,5 +1,6 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
+/*eslint max-lines-per-function: ["error", 500]*/
 
 const _ = require('lodash');
 const fs = require('fs');
@@ -12,9 +13,9 @@ const { get_process_fs_context } = require('../../util/native_fs_utils');
 
 const DEFAULT_FS_CONFIG = get_process_fs_context();
 
-mocha.describe('nb_native fs', function() {
+mocha.describe('nb_native fs', async function() {
 
-    mocha.describe('stat', function() {
+    mocha.describe('stat', async function() {
         mocha.it('works', async function() {
             const path = 'package.json';
             const res = await nb_native().fs.stat(DEFAULT_FS_CONFIG, path);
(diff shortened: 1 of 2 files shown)


src/test/unit_tests/test_nb_native_fs.js
```diff
@@ -12,9 +13,9 @@ const { get_process_fs_context } = require('../../util/native_fs_utils');
 
 const DEFAULT_FS_CONFIG = get_process_fs_context();
 
-mocha.describe('nb_native fs', function() {
+mocha.describe('nb_native fs', async function() {
```
shirady, line 16: General question - why do you add the `async` in the describe part? did you see an issue with this?
From what I saw in the docs the function inside `it` is async, but the `describe` doesn't contain it (see [USING ASYNC / AWAIT](https://mochajs.org/#using-async-await))

src/test/unit_tests/test_nb_native_fs.js
```diff
@@ -32,11 +33,28 @@ mocha.describe('nb_native fs', function() {
...
-            const res2 = await fs.promises.stat(path);
+    mocha.describe('lstat', async function() {
+        const link_name = 'link.json';
+        const file_name = 'file.json';
+        const PATH = `/tmp/lstat${file_name}`;
+        const LINK_PATH = `/tmp/lstat${link_name}`;
+        mocha.before('lstat-before', async function() {
+            const fd = await fs.promises.open(PATH, 'w');
+            fd.close();
+            await fs.promises.symlink(PATH, LINK_PATH);
```
shirady, line 44: Why do you need to open and close the file before creating the symbolic link?
If you can add a comment related to this.

src/test/unit_tests/test_nb_native_fs.js
```diff
@@ -32,11 +33,28 @@ mocha.describe('nb_native fs', function() {
...
+        mocha.after('lstat-after', async function() {
+            await fs.promises.rm(PATH);
+            await fs.promises.unlink(LINK_PATH);
+        });
+
+        mocha.it('lstat on symbolic link', async function() {
+            const res = await nb_native().fs.stat(DEFAULT_FS_CONFIG, LINK_PATH, { use_lstat: true });
+            const res2 = await fs.promises.lstat(LINK_PATH);
+            // res3 is the stat of the regular file
+            const res3 = await fs.promises.stat(LINK_PATH);
```
shirady, line 56: Maybe we can rename the variables (`res`, `res2`, `res3`) for readability? (in instead of or in addition to the comments).

src/test/unit_tests/test_nb_native_fs.js
```diff
@@ -47,6 +65,11 @@ mocha.describe('nb_native fs', function() {
                 _.omit(res, 'xattr', 'mtimeNsBigint', 'atimeNsBigint', 'ctimeNsBigint'), // we need to remove xattr, mtimeSec, mtimeNsec from fs_napi response as the node JS stat doesn't return them
                 _.omitBy(res2, v => typeof v === 'function'),
             );
+
+            assert.notDeepStrictEqual(
+                _.omit(res, 'xattr', 'mtimeNsBigint', 'atimeNsBigint', 'ctimeNsBigint'), // we need to remove xattr, mtimeSec, mtimeNsec from fs_napi response as the node JS stat doesn't return them
```
shirady, line 70: If find this comment confusing since the properties are not the same (I know it was copied from the code above).
```suggestion
                _.omit(res, 'xattr', 'mtimeNsBigint', 'atimeNsBigint', 'ctimeNsBigint'), // we need to remove those properties from fs_napi response as the node JS stat doesn't return them
```

src/test/unit_tests/test_nb_native_fs.js
```diff
@@ -47,6 +65,11 @@ mocha.describe('nb_native fs', function() {
                 _.omit(res, 'xattr', 'mtimeNsBigint', 'atimeNsBigint', 'ctimeNsBigint'), // we need to remove xattr, mtimeSec, mtimeNsec from fs_napi response as the node JS stat doesn't return them
                 _.omitBy(res2, v => typeof v === 'function'),
             );
+
+            assert.notDeepStrictEqual(
```
shirady, line 69: I would add a comment about this check and the check above so it would be clearer.

shirady (commented):
The code change in `fs_napi` looks fine.
I have a few questions and comments regarding the changes in the test file.

jackyalbo (approved):

PR #7889, opened 2024-03-12
NSFS | NC | Changes in `nsfs_config_schema` Related to `strictify`

### LINKED ISSUE #7874: NSFS | NC | Schema Changes (Mainly in `nsfs_config_schema`)
### Environment info
- NooBaa Version: master
- Platform: NSFS | NC

### Actual behavior
1. Currently we don't use `strictify` on `nsfs_config_schema`.
2. Currently we use `ajv.validate(schema_name, object_to_validate)` and it doesn't align with the avj documentation ([here](https://ajv.js.org/guide/getting-started.html#basic-data-validation)), please notice the comment "Best performance: compile and getSchema methods".

### Expected behavior
1. Use `strictify` on `nsfs_config_schema` with `additionalProperties: true`.
2. Compile the validate function once and re-use it (details in the link attached above).

### Steps to reproduce
1. none (internal change).

Notes:
Currently, we would get errors trying to `strictify` on `nsfs_config_schema`. Partial things I found:
- The schema uses `default` keyword which we didn't have in our past schemas (probably need to edit `strictify` for that case).
- The schema uses `description` which is unknown keyword, and can be replaced with `doc`.
- The schema uses `array`, but does not use `items` under it.
- Might need general refactoring in the `nsfs_config_schema` so we would not get errors on `strictify`.

shirady:
### Explain the changes
1. Add `strictify` on `nsfs_config_schema`.
2. Change the structure of the schema `nsfs_config_schema`:
    - Replace the `description` with `doc`.
    - Remove `default`.
    - In `NSFS_WHITELIST` which is type `array` add `items` under it.
    - Remove the `definitions` in `nsfs_config_schema` and set the `additionalProperties` to `nsfs_node_config_schema`.
3. Edit the printings in `load_nsfs_nc_config`.

### Issues: Fixed #7874
1. Those changes are originally from after setting `strictify` on `nsfs_config_schema` and handling the errors.

#### GAPS:
1. GAP - I think we should add validations to some of the properties, for example: `NSFS_WHITELIST` is validated when the user passes the command `sudo node src/cmd/manage_nsfs whitelist  --ips <list-og-ips>` using `verify_whitelist_ips`:
https://github.com/noobaa/noobaa-core/blob/d1bf8c465af745dedbff9d3803ec13da344884ba/src/cmd/manage_nsfs.js#L993-L1000
2. GAP (suggestion) - We can translate our recommendations in the schema, for example: ports values 1-65535 using a definition that we have:
https://github.com/noobaa/noobaa-core/blob/d1bf8c465af745dedbff9d3803ec13da344884ba/src/api/common_api.js#L1204-L1208
In other cases, for example `ENDPOINT_FORKS` to be `minimum` 0 and `maximum` 64 and remove the suggestions from the `doc` part.

#### Unit Tests
1. Run the schema unit tests: `npx jest test_nc_nsfs_config_schema_validation.test.js`

#### Manual Tests
To see that the config is used when running the nsfs server.
1. Check your hostname, run: `hostname` in your terminal (this will be the `<hostname-output>` in the next step).
2. Add a config file: `sudo vi /etc/noobaa.conf.d/config.json` In this example we added 2 opposite values of `ALLOW_HTTP`
```json
 {
     "ALLOW_HTTP": false,
     "host_customization": {
         "<hostname-output>": {
             "ALLOW_HTTP": true
         }
     }
 }
```
3. Start the server: `sudo node src/cmd/nsfs --debug 5`
5. Notice the printings (`my-mac` is an example of a hostname):
```bash
nsfs: config_dir_path=/etc/noobaa.conf.d
nsfs: config.json= {
  ALLOW_HTTP: false,
  host_customization: { 'my-mac': { ALLOW_HTTP: true } }
}
nsfs: merged config.json= { ALLOW_HTTP: true }
```

```bash
[nsfs/75607]    [L0] core.endpoint.endpoint:: Starting S3 HTTP 6001
[nsfs/75607]    [L0] core.endpoint.endpoint:: Started S3 HTTP successfully
```

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -1057,7 +1057,9 @@ function load_nsfs_nc_config() {
             }
             config[key] = merged_config[key];
         });
-        console.warn(`nsfs: config_dir_path=${config.NSFS_NC_CONF_DIR} config.json= ${util.inspect(merged_config)}`);
+        console.warn(`nsfs: config_dir_path=${config.NSFS_NC_CONF_DIR}`);
+        console.warn(`nsfs: config.json= ${util.inspect(config_data)}`);
+        console.warn(`nsfs: merged config.json= ${util.inspect(merged_config)}`);
         validate_nc_master_keys_config(config);
     } catch (err) {
         if (err.code !== 'MODULE_NOT_FOUND' && err.code !== 'ENOENT') throw err;
--- a/src/manage_nsfs/nsfs_schema_utils.js
+++ b/src/manage_nsfs/nsfs_schema_utils.js
@@ -36,12 +36,11 @@ schema_utils.strictify(account_schema, {
     additionalProperties: false
 });
 
+schema_utils.strictify(nsfs_config_schema, {});
+
 const validate_account = ajv.compile(account_schema);
 const validate_bucket = ajv.compile(bucket_schema);
 const validate_nsfs_config = ajv.compile(nsfs_config_schema);
-
-// NOTE - DO NOT strictify nsfs_config_schema
-// we might want to use it in the future for adding additional properties
 
 /**
  * validate_account_schema validates an account object against the NC NSFS account schema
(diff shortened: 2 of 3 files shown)


romayalon (approved):
Awesome! 
Please open a single issue for the suggested extra validations :)

PR #7895, opened 2024-03-13
[NSFS | NC] Simplify GLACIER xattr + minor fixes

tangledbytes:
### Explain the changes
This PR:
1. Reduces the total number of `xattr` in GLACIER from 4 to 2.
    1. Removed `user.noobaa.restore.ongoing` attr.
    2. Removed `user.noobaa.restore.staged` attr.
    3. Refer `get_restore_status` function if interested in state resolution logic.
4. Fixes invalid fd check is `fs_napi.cpp`.
5. Adds logging to `tapecloud.js`.
6. Prevents `storage_class` being shown as part of metadata.
7. Handle the case where filtered persistent log is empty.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -86,6 +86,9 @@ const copy_status_enum = {
     FALLBACK: 'FALLBACK'
 };
 
+const XATTR_METADATA_IGNORE_LIST = [
+    XATTR_STORAGE_CLASS_KEY,
+];
 
 /**
  * @param {fs.Dirent} a
@@ -266,9 +269,19 @@ function make_named_dirent(name) {
 }
 
 function to_xattr(fs_xattr) {
-    const xattr = _.mapKeys(fs_xattr, (val, key) =>
-        (key.startsWith(XATTR_USER_PREFIX) && !key.startsWith(XATTR_NOOBAA_INTERNAL_PREFIX) ? key.slice(XATTR_USER_PREFIX.length) : '')
-    );
+    const xattr = _.mapKeys(fs_xattr, (val, key) => {
+        // Prioritize ignore list
+        if (XATTR_METADATA_IGNORE_LIST.includes(key)) return '';
+
+        // Fallback to rules
+
+        if (key.startsWith(XATTR_USER_PREFIX) && !key.startsWith(XATTR_NOOBAA_INTERNAL_PREFIX)) {
+            return key.slice(XATTR_USER_PREFIX.length);
+        }
+
+        return '';
+    });
+
     // keys which do not start with prefix will all map to the empty string key, so we remove it once
     delete xattr[''];
     // @ts-ignore
@@ -1173,7 +1186,7 @@ class NamespaceFS {
             const src_file_path = await this._find_version_path(fs_context, params.copy_source);
             const stat = await nb_native().fs.stat(fs_context, src_file_path);
             const src_storage_class = s3_utils.parse_storage_class(stat.xattr[XATTR_STORAGE_CLASS_KEY]);
-            const src_restore_status = this._get_object_restore_status(stat, src_storage_class);
+            const src_restore_status = GlacierBackend.get_restore_status(stat.xattr, new Date(), src_file_path);
 
             if (src_storage_class === s3_utils.STORAGE_CLASS_GLACIER) {
                 if (src_restore_status?.ongoing || !src_restore_status?.expiry_time) {
@@ -1946,9 +1959,8 @@ class NamespaceFS {
      * restore_object simply sets the restore request xattr
      * which should be picked by another mechanism.
      * 
-     * restore_object internally relies on 3 xattrs:
+     * restore_object internally relies on 2 xattrs:
      * - XATTR_RESTORE_REQUEST
-     * - XATTR_RESTORE_ONGOING
      * - XATTR_RESTORE_EXPIRY
      * @param {*} params 
      * @param {nb.ObjectSDK} object_sdk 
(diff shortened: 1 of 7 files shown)


src/sdk/namespace_fs.js
```diff
@@ -1964,69 +1974,81 @@ class NamespaceFS {
...
 
-            // Total 8 states
+            const now = new Date();
+
+            // Total 6 states (restore_request - 2, restore_expiry - 3)
             const restore_request = stat.xattr[GlacierBackend.XATTR_RESTORE_REQUEST] ?
                 parseInt(stat.xattr[GlacierBackend.XATTR_RESTORE_REQUEST], 10) : null;
-            const restore_ongoing = stat.xattr[GlacierBackend.XATTR_RESTORE_ONGOING] === 'true';
             const restore_expiry = stat.xattr[GlacierBackend.XATTR_RESTORE_EXPIRY] ?
                 new Date(stat.xattr[GlacierBackend.XATTR_RESTORE_EXPIRY]) : null;
```
guymguym, line 1983: How do we handle bad date format, say if someone edits this xattr manually, which is also relevant for restore_request if it has a bad number - lets just be clear on how this behaves so that we can explain it. Perhaps it's best to log the issue and consider it as undefined?

src/sdk/namespace_fs.js
```diff
@@ -1964,69 +1974,81 @@ class NamespaceFS {
...
-                if (!restore_expiry) {
-                    dbg.log0('namespace_fs.restore_object: state - (!restore_request, restore_ongoing, !restore_expiry)');
-                    throw new S3Error(S3Error.RestoreAlreadyInProgress);
-                }
-            } else {
-                const now = new Date();
+            // 2 states
+            if (
+                (restore_request && !restore_expiry) ||
+                (restore_request && restore_expiry <= now)
```
guymguym, line 1988: Do we need to differentiate between undefined and 0 for restore_request? Right now 0 behaves just like undefined, and if we need to differentiate we can check restore_request > 0 vs. >=0.

src/sdk/namespace_fs.js
```diff
@@ -1964,69 +1974,81 @@ class NamespaceFS {
...
+                    '(restore_request, restore_expiry <= now)'
+                );
+
+                // First add it to the log and then add the extended attribute as if we fail after
+                // this point then the restore request can be triggered again without issue but
+                // the reverse doesn't works.
+                await this.append_to_restore_wal(file_path);
+
+                await file.replacexattr(fs_context, {
+                    [GlacierBackend.XATTR_RESTORE_REQUEST]: params.days.toString(),
```
guymguym, line 2027: Pehaps better to use restore_request value here which was already parsed and checked for validity instead of param.days

src/sdk/namespace_fs.js
```diff
@@ -1964,69 +1974,81 @@ class NamespaceFS {
...
+                // Should result in HTTP: 200 OK
+                return false;
             }
 
-            // 3 Invalid States
-            // 1. Restore request is set along with 1 or more than 1 attribute (2 different states)
-            // 2. Restore request is unset but both request ongoing and expiry are set
+            // Should be unreachable
+            dbg.error('namespace_fs.restore_object: unexpected state - reached unreachable');
             throw new RpcError('INTERNAL_ERROR');
```
guymguym, line 2052: Can we do something better other than error 500? This would just result a user complaint so if there is any fix we suggest lets add it to be automatic.

tangledbytes: I think for this case it is better that we do throw error. There should be just 6 combination possible, any 7th combination should not be possible and probably it is better that we get to take a look at the logs if and whenever we hit this rather than something silently keep failing and then it gets reported later on presenting itself as a totally different issue. WDYT?

src/sdk/namespace_fs.js
```diff
@@ -86,6 +86,7 @@ const copy_status_enum = {
     FALLBACK: 'FALLBACK'
 };
 
+const XATTR_METADATA_IGNORE_LIST = [XATTR_STORAGE_CLASS_KEY];
```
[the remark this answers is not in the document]
tangledbytes: Our tests aren't liking the attr missing. I have added it back in for now.

src/sdk/namespace_fs.js
```diff
@@ -2040,6 +2038,82 @@ class NamespaceFS {
...
+        if (xattr[GlacierBackend.XATTR_RESTORE_REQUEST]) {
+            const request = parseInt(xattr[GlacierBackend.XATTR_RESTORE_REQUEST], 10);
+            if (request === Number(xattr[GlacierBackend.XATTR_RESTORE_REQUEST])) {
+                restore_request = request;
+            }
+        }
+        if (xattr[GlacierBackend.XATTR_RESTORE_EXPIRY]) {
+            const expiry = new Date(xattr[GlacierBackend.XATTR_RESTORE_EXPIRY]);
+            // @ts-ignore - Checks for `Invalid Date Object`
+            if (!isNaN(expiry)) {
```
guymguym, line 2062: I think linter would prefer this check -
```suggestion
            if (!isNaN(expiry.getTime())) {
```

src/sdk/nsfs_glacier_backend/tapecloud.js
```diff
@@ -71,7 +71,10 @@ async function tapecloud_failure_handler(error) {
  */
 async function migrate(file) {
     try {
-        await exec(`${get_bin_path(MIGRATE_SCRIPT)} ${file}`);
+        dbg.log1("Starting migration for file", file);
+        const out = await exec(`${get_bin_path(MIGRATE_SCRIPT)} ${file}`, { return_stdout: true });
```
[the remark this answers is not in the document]
tangledbytes: Actually I just realized that for most of the cases, 1MB buffer is more than enough. Normal eeadm commands do not output a lot. I think we need spawn only for task output as that can be HUGE. WDYT?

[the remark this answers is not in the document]
guymguym: Maybe we should redirect the output to a file?

src/sdk/namespace_fs.js
```diff
@@ -2040,6 +2029,62 @@ class NamespaceFS {
     // INTERNALS //
     ///////////////
 
+    /**
+     * _get_restore_status is a pure function which consumes the xattr
+     * of the target file and returns its current state.
+     * @param {nb.NativeFSXattr} xattr xattr of the target file
+     * @param {Date} now - current time
+     * @param {string} [file] - file name for which the xattr is provided
+     * @returns {{ongoing?: boolean; state: nb.RestoreState; expiry_time?: Date}}
```
guymguym, line 2038: minor: move this typedef to nb.d.ts too:
```ts
interface RestoreStatus {
  state: nb.RestoreState;
  ongoing?: boolean;
  expiry_time?: Date;
}
```

src/sdk/namespace_fs.js
```diff
@@ -2040,6 +2029,62 @@ class NamespaceFS {
...
+            if (restore_expiry > now) {
+                dbg.warn('unexpected restore state - (restore_request, request_expiry > now) for', file);
+            }
+
+            return {
+                ongoing: true,
+                state: RESTORE_STATUS_ONGOING,
+            };
+        } else {
+            if (!restore_expiry || restore_expiry <= now) {
```
guymguym, line 2073: minor:
```suggestion
        } else if (!restore_expiry || restore_expiry <= now) {
```

src/sdk/namespace_fs.js
```diff
@@ -2254,29 +2299,11 @@ class NamespaceFS {
...
-                restore_status = {
-                    ongoing: true,
-                };
-            } else if (restore_expiry && restore_expiry > new Date()) {
-                restore_status = {
-                    ongoing: false,
-                    expiry_time: restore_expiry,
-                };
-            }
+            const status = this._get_restore_status(stat.xattr, new Date());
```
guymguym, line 2302: minor: better send a file_path here too

src/sdk/namespace_fs.js
```diff
@@ -2040,6 +2029,62 @@ class NamespaceFS {
     // INTERNALS //
     ///////////////
 
+    /**
+     * _get_restore_status is a pure function which consumes the xattr
+     * of the target file and returns its current state.
+     * @param {nb.NativeFSXattr} xattr xattr of the target file
+     * @param {Date} now - current time
+     * @param {string} [file] - file name for which the xattr is provided
```
guymguym, line 2037: rename to file_path to avoid confusion and also why is it optional ?

src/sdk/nsfs_glacier_backend/backend.js
```diff
@@ -151,8 +133,7 @@ class GlacierBackend {
         // in races
         if (
             stat.xattr[GlacierBackend.XATTR_RESTORE_REQUEST] ||
-            stat.xattr[GlacierBackend.XATTR_RESTORE_EXPIRY] ||
-            stat.xattr[GlacierBackend.XATTR_RESTORE_REQUEST_STAGED]) {
+            stat.xattr[GlacierBackend.XATTR_RESTORE_EXPIRY]) {
```
guymguym, line 136: I think what we want is to move the _get_restore_status function from namspace_fs to here, and reuse it also for resolving these states of should_restore and should_migrate - WDYT?

guymguym: Another small note - a few lines above this one there is a hardcoded 'GLACIER' string being used - I think we should avoid repeating these literals to avoid typos and use a constant. Perhaps look for this literal in the rest of the code and see if we can replace all those uses.

src/sdk/namespace_fs.js
```diff
@@ -1964,70 +1980,43 @@ class NamespaceFS {
...
+            }
 
-                    // First add it to the log and then add the extended attribute as if we fail after
-                    // this point then the restore request can be triggered again without issue but
-                    // the reverse doesn't works.
-                    await this.append_to_restore_wal(file_path);
+            if (restore_state.state === RESTORE_STATUS_RESTORED) {
+                const expires_on = now;
+                expires_on.setUTCDate(expires_on.getUTCDate() + params.days);
+                expires_on.setUTCHours(0, 0, 0, 0);
```
guymguym, line 2011: Should we really use midnight in UTC instead of midnight in local TZ?
Perhaps we could have the expiry time-of-day as a config.

src/sdk/namespace_fs.js
```diff
@@ -2040,6 +2029,62 @@ class NamespaceFS {
...
+     */
+    _get_restore_status(xattr, now, file = '') {
+        // Total 6 states (2x restore_request, 3x restore_expiry)
+        let restore_request;
+        let restore_expiry;
+
+        const restore_request_xattr = xattr[GlacierBackend.XATTR_RESTORE_REQUEST];
+        if (restore_request_xattr) {
+            const num = Number(restore_request_xattr);
+            if (num > 0 && Number.isInteger(num)) {
```
guymguym, line 2048: The main question I have about this format is whether having a simple restore_request that only captures the number of days is enough. I see the restore request in S3 can have a lot of parameters and I expect we might need to include those in the request xattr as well (think for example on the Tier field which can be Standard | Bulk | Expedited). Can we think if it is worth storing the actual json of the requested request in this xattr instead of just the raw days? This will also allow us to evolve the format easier.

src/sdk/namespace_fs.js
```diff
@@ -1964,70 +1980,43 @@ class NamespaceFS {
...
-            } else {
-                const now = new Date();
-
-                // The only entity that can remove the restore request is the underlying
-                // restore mechanism, implying that at some point it must have picked
-                // up the request to restore and finished the process.
-                if (restore_expiry && restore_expiry > now) {
-                    dbg.log0('namespace_fs.restore_object: state - (!restore_request, !restore_ongoing, restore_expiry > now)');
+            const now = new Date();
+            const restore_state = this._get_restore_status(stat.xattr, now, file_path);
```
guymguym, line 1984: minor: the const restore_state name is confusing as it actually returns an object with a state inside, maybe call it just restore_status or just status?

src/sdk/namespace_fs.js
tangledbytes: Yes, in s3_post_object_restore.js

src/sdk/namespace_fs.js
tangledbytes: This was intentional so as to not rely on all the nodes (nsfs nodes) having the same TZ configured. Do we need the TZ info?

src/sdk/namespace_fs.js
tangledbytes: Yeah, I don't see any issue with having JSON stored here. Our policy based solution on the other hand might need some updates to accommodate this change.

guymguym: We discussed few options for the request structure:

days float - e.g.
1
0.25

millis- e.g.
24*3600*1000

json - e.g.
{"days":1, "tier": "expedited" }
{"hours": 6}

For now we think days float is the simplest one to follow.

src/sdk/nsfs_glacier_backend/backend.js
guymguym: get_restore_status() should check the storage_class too
and should_migrate can call it to check if state is CAN_RESTORE
and should_restore can call it to check if state is ONGOING

src/sdk/nsfs_glacier_backend/tapecloud.js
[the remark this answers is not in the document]
guymguym: Deferred to next PR.

guymguym (approved):

src/sdk/namespace_fs.js
tangledbytes: Went for days as float for now. Resolving.

PR #7897, opened 2024-03-13
CI | NSFS | NC | Jest Tests - Change the value of `tmp_fs_path` For Concurrent Run 

shirady:
### Explain the changes
1.  Change the value of `tmp_fs_path` in each test (`test_nc_nsfs_account_cli.test.js`, `test_nc_nsfs_bucket_cli.test.js`) for the concurrent run that Jest is using, so we will not use the same path between tests (even if they are in different files), and `new_buckets_path` value.

1. Currently, sometimes the Jest CI is failing (it didn't happen in the past, probably as a result of tests that were added).

### Testing Instructions:
1. When running locally please run `npm run jest`.

- [ ] Doc added/updated
- [ ] Tests added


nimrod-becker:
Why are we skipping instead of fixing?

shirady:
@nimrod-becker This test is passing on my machine and used to pass in the past.
I opened this PR to check that now the Jest test is passing and planned to open an issue and investigate it.

naveenpaul1 (approved):

Issue #7894, opened 2024-03-13
Noobaa health command giving trimmed output in pipe


Roushan45:
### Environment info
- NooBaa Version: **noobaa-core-5.15.0-20240306.el9.x86_64**
- Platform: RHEL9**

### Actual behavior
1. bucket/account details output getting trimmed when using with pipe
usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/health --all_bucket_details |  cat

### Expected behavior
1. it should give full output in pipe

### Steps to reproduce
1. usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/health --all_bucket_details |  cat

shirady:
@Roushan45 would you please share the outputs?
cc: @naveenpaul1

naveenpaul1:
@shirady json output is no complete, JSON is getting trimmed middle of the complete output and json is in invalid state

rkomandu:
Upgraded on one of the CES S3 node and could see the account is displayed in jq when piped 


upgraded noobaa rpm node showed as follows 
/usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/health --all_account_details | jq |grep 10002
          "name": "s3user-10002",
          "storage_path": "/mnt/fs1/s3user-10002-dir"


on non-upgraded node, it continued to show the output as follows

/usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/health --all_account_details | jq
**parse error: Unfinished string at EOF at line 1, column 65536**

[root@rkomandu-ip-cls-x-worker2 ras]# rpm -qa |grep noobaa
noobaa-core-5.15.0-20240313.el9.x86_64


looks good for now, please merge into the ODF 4.15.z stream @naveenpaul1 @nimrod-becker

rkomandu:
Looks like @nimrod-becker , will be trying this into the 4.15.1 as per email communication dated 18th Mar

nimrod-becker:
Closing per verification [comment](https://github.com/noobaa/noobaa-core/issues/7894#issuecomment-2003359200)

PR #7898, opened 2024-03-13
NSFS | NC | health giving trimmed output in pipe

naveenpaul1:
### Explain the changes
1. if the size of health output is bucket/account details output getting trimmed when using with pipe 
usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/health --all_bucket_details | cat
2. Update timeout for test_nc_nsfs_account_cli.test.js
3. remove unused `iam_json_schema` and `iam_ttl`

with pipe `cat` or `tee` operation getting first batch of output and before flushing second set of data process is getting exited for health command. This change will remove the `process.exit()` from code and make sure complete output is printed/saved to terminal/file

1. https://github.com/noobaa/noobaa-core/issues/7894

### Testing Instructions:
1. run below command and make sure output size more than 64 bytes
usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/health --all_bucket_details | cat

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/health.js
+++ b/src/cmd/health.js
@@ -424,8 +424,9 @@ async function main(argv = minimist(process.argv.slice(2))) {
         if (deployment_type === 'nc') {
             const health = new NSFSHealth({ https_port, config_root, all_account_details, all_bucket_details, check_syslog_ng });
             const health_status = await health.nc_nsfs_health();
-            process.stdout.write(JSON.stringify(health_status) + '\n');
-            process.exit(0);
+            process.stdout.write(JSON.stringify(health_status) + '\n', () => {
+                process.exit(0);
+            });
         } else {
             dbg.log0('Health is not supported for simple nsfs deployment.');
         }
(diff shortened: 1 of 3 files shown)


src/cmd/health.js
```diff
@@ -425,7 +425,6 @@ async function main(argv = minimist(process.argv.slice(2))) {
             const health = new NSFSHealth({ https_port, config_root, all_account_details, all_bucket_details, check_syslog_ng });
             const health_status = await health.nc_nsfs_health();
             process.stdout.write(JSON.stringify(health_status) + '\n');
```
guymguym, line 427: @naveenpaul1 Can you try this approach using the write callback to exit only after flushing the output?
```suggestion
            process.stdout.write(JSON.stringify(health_status) + '\n', () => {
                process.exit(0);
            });
```

guymguym (changes requested):
Please check if my suggestion above works

src/cmd/health.js
naveenpaul1: Sure @guymguym

naveenpaul1: did one round of testing with monitoring team and this fix is working

guymguym (approved):

Issue #7828, opened 2024-02-18
NSFS | NC | Allow unset 'new_buckets_path'


shirady:
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
1. Currently we don't have the option to unset the property `new_buckets_path` after the account was created.

### Expected behavior
1. To allow users to unset the property `new_buckets_path` after the account was created when using Manage NSFS CLI.

### Steps to reproduce
1. none

### More information - Screenshots / Logs / Other output
Like we have the option to unset `fs_backend` and `s3_policy`.

nimrod-becker:
Update to account, disabling creation of new buckets for an account

romayalon:
@nadavMiz closing per your fix

PR #7899, opened 2024-03-13
NSFS | NC | Allow unset 'new_buckets_path'

nadavMiz:
### Explain the changes
1. Add a condition to set `new_buckets_path` to be undefined

1. https://github.com/noobaa/noobaa-core/issues/7828

### Testing Instructions:
1. Create an account with new_buckets_path (allow_bucket_creation inferred to be true): `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path /tmp/nsfs_root1 --uid <uid> --gid <gid>`
2. update account to have empty new_buckets_path: `node src/cmd/manage_nsfs account update --name=<account-name> --new_buckets_path=''`
3. account json should not have `new_buckets_path` field and `allow_bucket_creation` should be set to false

Unit Tests:
`sudo npx jest test_nc_nsfs_account_cli.test.js`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -390,6 +390,10 @@ async function fetch_account_data(action, user_input) {
     // secret_key as SensitiveString
     data.access_keys[0].secret_key = _.isUndefined(data.access_keys[0].secret_key) ? undefined :
         new SensitiveString(String(data.access_keys[0].secret_key));
+    // fs_backend deletion specified with empty string '' (but it is not part of the schema)
+    data.nsfs_account_config.fs_backend = data.nsfs_account_config.fs_backend || undefined;
+    // new_buckets_path deletion specified with empty string ''
+    data.nsfs_account_config.new_buckets_path = data.nsfs_account_config.new_buckets_path || undefined;
     // allow_bucket_creation either set by user or infer from new_buckets_path
     if (_.isUndefined(user_input.allow_bucket_creation)) {
         data.allow_bucket_creation = !_.isUndefined(data.nsfs_account_config.new_buckets_path);
@@ -398,8 +402,6 @@ async function fetch_account_data(action, user_input) {
     } else { // string of true or false
         data.allow_bucket_creation = user_input.allow_bucket_creation.toLowerCase() === 'true';
     }
-    // fs_backend deletion specified with empty string '' (but it is not part of the schema)
-    data.nsfs_account_config.fs_backend = data.nsfs_account_config.fs_backend || undefined;
 
     return data;
 }
--- a/src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
+++ b/src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
@@ -557,6 +557,25 @@ describe('manage nsfs cli account flow', () => {
             expect(JSON.parse(res.stdout).error.message).toBe(ManageCLIError.InvalidArgumentType.message);
         });
 
+        it('cli update account unset new_buckets_path', async () => {
+            const { name } = defaults;
+            //in exec_manage_cli an empty string is passed as no input. so operation fails on invalid type. use the string '' instead 
+            const empty_string = '\'\'';
+            const account_options = { config_root, name, new_buckets_path: empty_string};
+            const action = ACTIONS.UPDATE;
+            await exec_manage_cli(type, action, account_options);
+            let new_account_details = await read_config_file(config_root, CONFIG_SUBDIRS.ACCOUNTS, name);
+            expect(new_account_details.nsfs_account_config.new_buckets_path).toBeUndefined();
+            expect(new_account_details.allow_bucket_creation).toBe(false);
+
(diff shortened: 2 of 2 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -388,6 +388,10 @@ async function fetch_account_data(action, user_input) {
     // secret_key as SensitiveString
     data.access_keys[0].secret_key = _.isUndefined(data.access_keys[0].secret_key) ? undefined :
         new SensitiveString(String(data.access_keys[0].secret_key));
+    // allow to unset new_buckets_path
+    if (data.nsfs_account_config.new_buckets_path === '') {
+        data.nsfs_account_config.new_buckets_path = undefined;
+    }
```
shirady, line 394: Just for the code style to match I would move lines 403-404 to this part (unset values) and add in the same way the unset.

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -557,6 +557,23 @@ describe('manage nsfs cli account flow', () => {
             expect(JSON.parse(res.stdout).error.message).toBe(ManageCLIError.InvalidArgumentType.message);
         });
 
+        it('cli unset new_buckets_path', async () => {
```
shirady, line 560: Minor suggestion so it will be easier to understand the test when only viewing the test's name.
```suggestion
        it('cli update account unset new_buckets_path', async () => {
```

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -557,6 +557,23 @@ describe('manage nsfs cli account flow', () => {
...
+            const { name } = defaults;
+            const account_options = { config_root, name, new_buckets_path: '\'\''};
```
shirady, line 562: Why not use the `new_buckets_path` from defaults?
```suggestion
            const { name, new_buckets_path } = defaults;
            const account_options = { config_root, name, new_buckets_path};
```

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -557,6 +557,23 @@ describe('manage nsfs cli account flow', () => {
...
+            let new_account_details = await read_config_file(config_root, CONFIG_SUBDIRS.ACCOUNTS, name);
+            expect(new_account_details.nsfs_account_config.new_buckets_path).toBeUndefined();
+            expect(new_account_details.allow_bucket_creation).toBe(false);
+
+            //set new_buckets_path value again
+            account_options.new_buckets_path = defaults.new_buckets_path;
+            await exec_manage_cli(type, action, account_options);
+            new_account_details = await read_config_file(config_root, CONFIG_SUBDIRS.ACCOUNTS, name);
+            expect(new_account_details.nsfs_account_config.new_buckets_path).toBe(defaults.new_buckets_path);
+            expect(new_account_details.allow_bucket_creation).toBe(true);
```
shirady, line 574: I'm not sure you need this inside the same test.
It is matching a test where the account was created without `new_buckets_path` and a user is adding it and the test checks that it was added, right?

I think you can check if we have this test, and if we don't you can add a new test.

shirady (commented):
A few minor comments

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
nadavMiz: I want to update new_buckets_path as empty ('') to see that it unsets. new_buckets_path from defaults is a valid value

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
nadavMiz: I couldn't find a test that account was created without new_buckets_path and a user is adding it so I added one.
However, for this test we need to get the value of new_buckets_path to its original value for future tests. so I left this code as is.

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -332,7 +332,6 @@ describe('manage nsfs cli account flow', () => {
             const res = await exec_manage_cli(type, action, account_options);
             expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidArgumentType.code);
         });
-
```
shirady, line 335: redundant

shirady (approved):
LGTM

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -332,8 +332,8 @@ describe('manage nsfs cli account flow', () => {
...
     });
+
```
shirady, line 336: redundant

PR #7901, opened 2024-03-14
bucket_logging: Change bucket API's and functions as per schema

aspandey:
We do not store name of the bucket, on which we are setting bucket logging, in bucket schema.
There is no need to return that name as part of get_bucket_logging and store it as part of put_bucket_logging

Also, made changed to return a correct value in case of bucket logging is configured for a bucket.

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -393,7 +393,6 @@ async function put_bucket_logging(req) {
     dbg.log0('put_bucket_logging:', req.rpc_params);
     const bucket = find_bucket(req);
     const logging = {
-                        "name": bucket.name,
                         "log_bucket": req.rpc_params.log_bucket,
                         "log_prefix": req.rpc_params.log_prefix
                     };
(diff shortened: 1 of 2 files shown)


src/server/system_services/bucket_server.js
```diff
@@ -412,12 +411,19 @@ async function get_bucket_logging(req) {
...
+        logging = {
+            "log_bucket": bucket.logging.log_bucket,
+            "log_prefix": bucket.logging.log_prefix
+        };
+    } else {
+        logging = {
+            "log_bucket": "Not Configured",
+            "log_prefix": "Not Configured"
+        };
+    }
```
liranmauda, line 426: Maybe use conditional formatting? 
```suggestion
    const logging = {
        "log_bucket": bucket.logging?.log_bucket || "Not Configured",
        "log_prefix": bucket.logging?.log_prefix || "Not Configured"
    };
```

aspandey: done

liranmauda (commented):
LGTM

liranmauda (approved):

PR #7907, opened 2024-03-18
NSFS | NC | Manage NSFS | Remove `wide` Property from Fetch Data

### LINKED ISSUE #7871: NSFS | NC | `wide` Property Should Be Removed from `fetch_bucket_data` And `fetch_account_data`
### Environment info
- NooBaa Version: master
- Platform: NSFS NC

### Actual behavior
1. Currently, the `wide` property is fetched inside `fetch_bucket_data` and `fetch_account_data` while it is not a property of an account or bucket.

### Expected behavior
1. The `wide` property should be removed from `fetch_bucket_data` and `fetch_account_data` and fetch separately when needed (which is when using account/bucket list).

### Steps to reproduce
1. none, it is internal change.

Relevant parts in the code:
https://github.com/noobaa/noobaa-core/blob/8363cf721afdc2968a324b85751bf9f4717c11ce/src/cmd/manage_nsfs.js#L149

https://github.com/noobaa/noobaa-core/blob/8363cf721afdc2968a324b85751bf9f4717c11ce/src/cmd/manage_nsfs.js#L395

shirady:
### Explain the changes
1. Remove wide property from ` fetch_bucket_data` `fetch_account_data`.
2. Edit test with list wide to make sure there is a difference between list and list with wide flag.

### Issues: Fixed #7871 
1. Currently, the `wide` property is fetched inside `fetch_bucket_data` and `fetch_account_data` while it is not a property of an account or bucket.

#### Manual Tests:
1. Create a couple of accounts and a couple of buckets (using Manage NSFS).
3. List accounts using wide flag: `sudo node src/cmd/manage_nsfs account list --wide`
4. List buckets using wide flag: `sudo node src/cmd/manage_nsfs bucket list --wide`

#### Unit Tests:
- `sudo npx jest test_nc_nsfs_account_cli.test.js` (for example: 'cli list wide').
- `sudo node ./node_modules/.bin/_mocha src/test/unit_tests/test_nc_nsfs_cli.js` (for example: 'cli bucket list - wide')

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -136,7 +136,6 @@ async function fetch_bucket_data(action, user_input) {
         owner_account: undefined,
         system_owner: user_input.owner, // GAP - needs to be the system_owner (currently it is the account name)
         bucket_owner: user_input.owner,
-        wide: _.isUndefined(user_input.wide) ? undefined : get_boolean_or_string_value(user_input.wide),
         tag: undefined, // if we would add the option to tag a bucket using CLI, this should be changed
         versioning: action === ACTIONS.ADD ? 'DISABLED' : undefined,
         creation_date: action === ACTIONS.ADD ? new Date().toISOString() : undefined,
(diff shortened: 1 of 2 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -572,7 +571,8 @@ async function manage_account_operations(action, data, show_secrets, user_input)
         await delete_account(data);
     } else if (action === ACTIONS.LIST) {
         const account_filters = _.pick(user_input, LIST_ACCOUNT_FILTERS);
-        const accounts = await list_config_files(TYPES.ACCOUNT, accounts_dir_path, data.wide, show_secrets, account_filters);
+        const wide = get_boolean_or_string_value(user_input.wide);
```
naveenpaul1, line 574: @shirady do we need this check here? `_.isUndefined(user_input.wide)`

shirady: @naveenpaul1 I don't need this check because when the value is `undefined` then the function `get_boolean_or_string_value` will return `false`.
https://github.com/noobaa/noobaa-core/blob/83166c69ab308d7b37b2656a5561c972ca720889/src/cmd/manage_nsfs.js#L1042-L1044
Before I had to keep it as `undefined` because it was part of the data and it would add `wide: false` to the bucket and the account and would fail the schema validation.
It was part of the motivation for opening the issue.

naveenpaul1 (approved):

PR #7908, opened 2024-03-18
[NC | NSFS] Improve time bounds for glacier operations

tangledbytes:
### Explain the changes
This PR attempts to:
1. Improve time bounds for glacier operations by delegating the responsibility of log rotation on the log consumer (manage_nsfs.js).
2. Add the concept of failure logging. This helps in recording the previously failed entries in a separate log whose size will be always be at most the size of the processing log. This allows for more efficient retires and at the same time provides more visibility into the functioning as the failed entries are stored in `migrate.failure.[...].log` and `restore.failure.[...].log` files.
3. Improves CLI logging

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/manage_nsfs/manage_nsfs_glacier.js
+++ b/src/manage_nsfs/manage_nsfs_glacier.js
@@ -35,14 +35,7 @@ async function process_migrations() {
  * @param {import('../sdk/nsfs_glacier_backend/backend').GlacierBackend} backend
  */
 async function run_glacier_migrations(fs_context, backend) {
-    // This WAL is getting opened only so that we can process all the prcess WAL entries
-    const wal = new PersistentLogger(
-        config.NSFS_GLACIER_LOGS_DIR,
-        GlacierBackend.MIGRATE_WAL_NAME,
-        { disable_rotate: true, locking: 'EXCLUSIVE' },
-    );
-
-    await wal.process_inactive(async file => backend.migrate(fs_context, file));
+    await run_glacier_operation(fs_context, GlacierBackend.MIGRATE_WAL_NAME, backend.migrate.bind(backend));
 }
 
 async function process_restores() {
@@ -69,14 +62,7 @@ async function process_restores() {
  * @param {import('../sdk/nsfs_glacier_backend/backend').GlacierBackend} backend
  */
 async function run_glacier_restore(fs_context, backend) {
-    // This WAL is getting opened only so that we can process all the prcess WAL entries
-    const wal = new PersistentLogger(
-        config.NSFS_GLACIER_LOGS_DIR,
-        GlacierBackend.RESTORE_WAL_NAME,
-        { disable_rotate: true, locking: 'EXCLUSIVE' },
-    );
-
-    await wal.process_inactive(async file => backend.restore(fs_context, file));
+    await run_glacier_operation(fs_context, GlacierBackend.RESTORE_WAL_NAME, backend.restore.bind(backend));
 }
 
 async function process_expiry() {
@@ -106,7 +92,7 @@ async function time_exceeded(fs_context, interval, timestamp_file) {
 
         if (lastrun.getTime() + interval < Date.now()) return true;
     } catch (error) {
-        console.error('failed to read last run timestamp:', error);
+        console.error('failed to read last run timestamp:', error, 'timestamp_file:', timestamp_file);
         if (error.code === 'ENOENT') return true;
 
         throw error;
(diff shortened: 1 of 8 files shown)


src/util/persistent_logger.js
```diff
@@ -21,9 +21,9 @@ class PersistentLogger {
      * @param {string} dir parent directory
      * @param {string} namespace file prefix
      * @param {{
-     *  max_interval?: Number,
+     *  poll_interval?: Number,
      *  locking?: "SHARED" | "EXCLUSIVE",
-     *  disable_rotate?: boolean,
+     *  disable_poll_active?: boolean,
```
guymguym, line 26: I think we can remove this and rely on poll_interval only so we don't forget to set it by the various callers that only read the log or write a temp log.

src/sdk/nsfs_glacier_backend/tapecloud.js
```diff
@@ -21,11 +23,49 @@ function get_bin_path(bin_name) {
...
+            if (parsed[1] !== ERROR_DUPLICATE_TASK) {
+                // Column 5 is the filename (refer tapecloud [eeadm] manual)
+                await recorder(parsed[5]);
+            }
+        });
+
+        let errcode = 0;
+
+        proc.on('exit', code => { errcode = code; });
+        reader.on('close', () => {
```
tangledbytes, line 52: As per the docs of readline:
> The input stream receives its 'end' event;

I expect that this event will occur only once we are done going over all the lines. @guymguym how fair do you think is this assumption?

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -36,13 +36,38 @@ async function process_migrations() {
...
+    }
+
+    try {
+        // Finally replace the current active so as to consume them in the next iteration
+        await failure_wal._replace_active();
+    } catch (error) {
+        console.error('failed to replace active failure log:', error);
+    }
+
+    await failure_wal.close();
```
guymguym, line 70: This entire flow seems to be identical to the run_glacier_restore with just few changed variables/callbacks.
Do you think we can write it as a one function that gets called with different args?

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -36,13 +36,38 @@ async function process_migrations() {
...
     );
 
-    await wal.process_inactive(async file => backend.migrate(fs_context, file));
+    try {
+        // Process all the inactive and currently active log
+        await wal.process_inactive(async file => backend.migrate(fs_context, file, failure_wal.append.bind(failure_wal)));
+    } catch (error) {
+        console.error('failed to process restore logs:', error);
+    } finally {
+        await wal.close();
```
guymguym, line 52: what about closing failure_wal? perhaps we need to wrap this entire flow with a finally that will handle closing the both of them.

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -36,13 +36,38 @@ async function process_migrations() {
...
 async function run_glacier_migrations(fs_context, backend) {
     // This WAL is getting opened only so that we can process all the prcess WAL entries
-    const wal = new PersistentLogger(
+    const wal = new PersistentLogger(config.NSFS_GLACIER_LOGS_DIR, GlacierBackend.MIGRATE_WAL_NAME, { locking: 'EXCLUSIVE' });
+    const failure_wal = new PersistentLogger(
         config.NSFS_GLACIER_LOGS_DIR,
-        GlacierBackend.MIGRATE_WAL_NAME,
-        { disable_rotate: true, locking: 'EXCLUSIVE' },
+        `${GlacierBackend.MIGRATE_WAL_NAME}.failure`,
+        { locking: 'EXCLUSIVE' },
```
guymguym, line 43: I think the term wal here is unrelated anymore. I would just call it log and failure_log perhaps

src/sdk/nsfs_glacier_backend/tapecloud.js
```diff
@@ -7,6 +7,8 @@ const { GlacierBackend } = require("./backend");
 const config = require('../../../config');
 const path = require("path");
 const { exec } = require('../../util/os_utils');
+const { spawn } = require("child_process");
+const { createInterface } = require("readline/promises");
```
guymguym, line 11: I would name this const readline = require(...) because and then use readline.createInterface because without the module name as context it is less clear.

src/sdk/nsfs_glacier_backend/tapecloud.js
guymguym: https://nodejs.org/docs/latest/api/readline.html#example-read-file-stream-line-by-line

reader.on('line', async ...) will not really wait for the internal promise - I think you need to use for await for that backpressure of the input stream, and will only finish the async iterator loop once the input stream ends.

guymguym (approved):

Issue #7902, opened 2024-03-14
node bucket delete command should delete the bucket only if the bucket is empty.


anandhu-karattu:
### Environment info
- NooBaa Version: noobaa-core-5.15.0-20240313.el9.x86_64 (4.15 down stream)
- Platform: Standalone . RHEL 9.2

### Actual behavior
node bucket delete command is not validating if the bucket is empty or not.
It deletes the bucket with out checking if there is an object inside the bucket.

### Expected behavior
Bucket deletion should not be allowed if the bucket has an object inside it.

### Steps to reproduce
1. Create a bucket using node command "bucket add"

```
[root@anan4node-x-worker1 s3-config]# /usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/manage_nsfs bucket add --name bucket1000 --owner account-36503 --path /mnt/fs1/account_36503/bucket1000/
{
  "response": {
    "code": "BucketCreated",
    "reply": {
      "_id": "65f2c6fe37f6848f788cb42e",
      "name": "bucket1000",
      "owner_account": "65f2c613daf6183e5f4dad8c",
      "system_owner": "account-36503",
      "bucket_owner": "account-36503",
      "versioning": "DISABLED",
      "creation_date": "2024-03-14T09:44:30.800Z",
      "path": "/mnt/fs1/account_36503/bucket1000/",
      "should_create_underlying_storage": false,
      "fs_backend": "GPFS"
    }
  }
}
```

2. Upload an object in to the bucket using AWS CLI

```
[root@anan4node-x-worker1 ~]# AWS_ACCESS_KEY_ID=vt5PPN2sf8jRA73Sntbr AWS_SECRET_ACCESS_KEY=v59RplIV26oEa2ID2X46hZ8NrP/wCBY7OiEOfEv7 aws --endpoint https://10.11.57.178:6443 --no-verify-ssl s3 cp awscliv2.zip s3://bucket1000
urllib3/connectionpool.py:1061: InsecureRequestWarning: Unverified HTTPS request is being made to host '10.11.57.178'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
[...]

PR #7911, opened 2024-03-19
NSFS | CLI | Delete the bucket only if the bucket is empty or with --force flag

naveenpaul1:
### Explain the changes
1. Deletes the bucket with out checking if there is an object inside the bucket. This change will check whether the bucket is empty or not
2. check for temp folder and do not consider temp folder for empty condition
3. add --force flag to delete bucket even if its not empty

1. https://github.com/noobaa/noobaa-core/issues/7902

### Testing Instructions:
1. create a bucket add objects to it and try to delete with bucket CLI command, should throw error `BucketDeleteForbiddenHasObjects`
2. create a bucket add objects to it and try to delete with --force flag, bucket should get delete.

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -292,21 +292,27 @@ async function update_bucket(data) {
     write_stdout_response(ManageCLIResponse.BucketUpdated, data);
 }
 
-async function delete_bucket(data) {
+async function delete_bucket(data, force) {
     await validate_bucket_args(data, ACTIONS.DELETE);
     // we have fs_contexts: (1) fs_backend for bucket temp dir (2) config_root_backend for config files
     const fs_context_config_root_backend = native_fs_utils.get_process_fs_context(config_root_backend);
     const fs_context_fs_backend = native_fs_utils.get_process_fs_context(data.fs_backend);
     const bucket_config_path = get_config_file_path(buckets_dir_path, data.name);
     try {
-        const bucket_temp_dir_path = path.join(data.path, config.NSFS_TEMP_DIR_NAME + "_" + data._id);
-        await native_fs_utils.folder_delete(bucket_temp_dir_path, fs_context_fs_backend, true);
-        await native_fs_utils.delete_config_file(fs_context_config_root_backend, buckets_dir_path, bucket_config_path);
+        const temp_dir_name = config.NSFS_TEMP_DIR_NAME + "_" + data._id;
+        const bucket_temp_dir_path = path.join(data.path, temp_dir_name);
+        const entries = await nb_native().fs.readdir(fs_context_fs_backend, data.path);
+        const object_entries = entries.filter(element => !element.name.endsWith(temp_dir_name));
+        if (object_entries.length === 0 || force) {
+            await native_fs_utils.folder_delete(bucket_temp_dir_path, fs_context_fs_backend, true);
+            await native_fs_utils.delete_config_file(fs_context_config_root_backend, buckets_dir_path, bucket_config_path);
+            write_stdout_response(ManageCLIResponse.BucketDeleted, '', {bucket: data.name});
+        }
+        throw_cli_error(ManageCLIError.BucketDeleteForbiddenHasObjects, data.name);
     } catch (err) {
         if (err.code === 'ENOENT') throw_cli_error(ManageCLIError.NoSuchBucket, data.name);
         throw err;
     }
-    write_stdout_response(ManageCLIResponse.BucketDeleted, '', {bucket: data.name});
 }
 
 async function manage_bucket_operations(action, data, user_input) {
@@ -317,7 +323,8 @@ async function manage_bucket_operations(action, data, user_input) {
     } else if (action === ACTIONS.UPDATE) {
         await update_bucket(data);
     } else if (action === ACTIONS.DELETE) {
-        await delete_bucket(data);
+        const force = get_boolean_or_string_value(user_input.force);
+        await delete_bucket(data, force);
     } else if (action === ACTIONS.LIST) {
         const bucket_filters = _.pick(user_input, LIST_BUCKET_FILTERS);
         const wide = get_boolean_or_string_value(user_input.wide);
@@ -932,7 +939,7 @@ function validate_options_type_by_value(input_options_with_data) {
                 continue;
             }
             // special case for boolean values
-            if (['allow_bucket_creation', 'regenerate', 'wide', 'show_secrets'].includes(option) && validate_boolean_string_value(value)) {
+            if (['allow_bucket_creation', 'regenerate', 'wide', 'show_secrets', 'force'].includes(option) && validate_boolean_string_value(value)) {
                 continue;
             }
             // special case for bucket_policy (from_file)
--- a/src/manage_nsfs/manage_nsfs_help_utils.js
+++ b/src/manage_nsfs/manage_nsfs_help_utils.js
@@ -163,6 +163,7 @@ bucket delete [flags]
 
 Flags:
 --name <string>                                                         The name of the bucket
+--force                                               (optional)        Forcefully delete bucket if the bucket is not empty
 `;
 
 const BUCKET_FLAGS_STATUS = `
--- a/src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
+++ b/src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
@@ -15,6 +15,7 @@ const { set_path_permissions_and_owner, TMP_PATH, generate_s3_policy } = require
 const { ACTIONS, TYPES, CONFIG_SUBDIRS } = require('../../../manage_nsfs/manage_nsfs_constants');
 const { get_process_fs_context, is_path_exists } = require('../../../util/native_fs_utils');
 const ManageCLIError = require('../../../manage_nsfs/manage_nsfs_cli_errors').ManageCLIError;
+const { ManageCLIResponse } = require('../../../manage_nsfs/manage_nsfs_cli_responses');
 
 const tmp_fs_path = path.join(TMP_PATH, 'test_nc_nsfs_bucket_cli');
 const DEFAULT_FS_CONFIG = get_process_fs_context();
@@ -161,6 +162,43 @@ describe('manage nsfs cli bucket flow', () => {
             path_exists = await is_path_exists(DEFAULT_FS_CONFIG, bucket_temp_dir_path);
             expect(path_exists).toBe(false);
         });
+
+        it('cli delete bucket with force flag, when bucket path is not empty', async () => {
+            //here a dummy file is creating in bucket storage location(bucket_defaults.path) 
+            await create_json_file(bucket_defaults.path, 'test.json', {test: 'data'});
+            const delete_bucket_options = { config_root, name: bucket_defaults.name, force: true};
+            const resp = await exec_manage_cli(TYPES.BUCKET, ACTIONS.DELETE, delete_bucket_options);
+            expect(JSON.parse(resp.trim()).response.code).toBe(ManageCLIResponse.BucketDeleted.code);
+            const config_path = path.join(config_root, CONFIG_SUBDIRS.BUCKETS, bucket_defaults.name + '.json');
+            await fs_utils.file_must_not_exist(config_path);
+        });
+
+        it('should fail - cli delete bucket when bucket path is not empty', async () => {
+            //here a dummy file is creating in bucket storage location(bucket_defaults.path), 
+            await create_json_file(bucket_defaults.path, 'test1.json', {test: 'data'});
+            const delete_bucket_options = { config_root, name: bucket_defaults.name};
+            const resp = await exec_manage_cli(TYPES.BUCKET, ACTIONS.DELETE, delete_bucket_options);
+            expect(JSON.parse(resp.stdout).error.code).toBe(ManageCLIError.BucketDeleteForbiddenHasObjects.code);
+            const config_path = path.join(config_root, CONFIG_SUBDIRS.BUCKETS, bucket_defaults.name + '.json');
+            await fs_utils.file_must_exist(config_path);
+        });
+
+        it('cli delete bucket force flag with valid boolean string(\'true\')', async () => {
+            //here a dummy file is creating in bucket storage location(bucket_defaults.path) 
+            await create_json_file(bucket_defaults.path, 'test.json', {test: 'data'});
+            // force wth valid string value 'true'
+            const delete_bucket_options = { config_root, name: bucket_defaults.name, force: 'true'};
+            const resp = await exec_manage_cli(TYPES.BUCKET, ACTIONS.DELETE, delete_bucket_options);
+            expect(JSON.parse(resp.trim()).response.code).toBe(ManageCLIResponse.BucketDeleted.code);
+            const config_path = path.join(config_root, CONFIG_SUBDIRS.BUCKETS, bucket_defaults.name + '.json');
+            await fs_utils.file_must_not_exist(config_path);
+        });
+
+        it('should fail - cli delete bucket force flag with invalid boolean string(\'nottrue\')', async () => {
+            const delete_bucket_options = { config_root, name: bucket_defaults.name, force: 'nottrue'};
+            const resp = await exec_manage_cli(TYPES.BUCKET, ACTIONS.DELETE, delete_bucket_options);
+            expect(JSON.parse(resp.stdout).error.code).toBe(ManageCLIError.InvalidBooleanValue.code);
+        });
     });
 });
 
@@ -212,7 +250,7 @@ describe('cli create bucket using from_file', () => {
         // write the json_file_options
         const path_to_option_json_file_name = await create_json_bucket_options(path_to_json_bucket_options_dir, bucket_options);
         const command_flags = {config_root, from_file: path_to_option_json_file_name};
-        // create tha bucket and check the details
+        // create the bucket and check the details
         await exec_manage_cli(type, action, command_flags);
         // compare the details
         const bucket = await read_config_file(config_root, CONFIG_SUBDIRS.BUCKETS, bucket_defaults.name);
@@ -227,7 +265,7 @@ describe('cli create bucket using from_file', () => {
         // write the json_file_options
         const path_to_option_json_file_name = await create_json_bucket_options(path_to_json_bucket_options_dir, bucket_options);
         const command_flags = {config_root, from_file: path_to_option_json_file_name};
-        // create tha bucket and check the details
+        // create the bucket and check the details
         await exec_manage_cli(type, action, command_flags);
         // compare the details
         const bucket = await read_config_file(config_root, CONFIG_SUBDIRS.BUCKETS, bucket_defaults.name);
@@ -243,7 +281,7 @@ describe('cli create bucket using from_file', () => {
         // write the json_file_options
         const path_to_option_json_file_name = await create_json_bucket_options(path_to_json_bucket_options_dir, bucket_options);
         const command_flags = {config_root, from_file: path_to_option_json_file_name};
-        // create tha bucket and check the details
+        // create the bucket and check the details
         await exec_manage_cli(type, action, command_flags);
         // compare the details
         const bucket = await read_config_file(config_root, CONFIG_SUBDIRS.BUCKETS, bucket_defaults.name);
@@ -320,7 +358,7 @@ describe('cli create bucket using from_file', () => {
         await fs.promises.writeFile(path_to_option_json_file_name, content);
         // write the json_file_options
         const command_flags = {config_root, from_file: path_to_option_json_file_name};
-        // create tha bucket
+        // create the bucket
         await exec_manage_cli(type, action, command_flags);
         // compare the details
         const res = await exec_manage_cli(type, action, command_flags);
@@ -379,8 +417,19 @@ async function read_config_file(config_root, schema_dir, config_file_name, is_sy
  */
 async function create_json_bucket_options(path_to_json_bucket_options_dir, bucket_options) {
     const option_json_file_name = `${bucket_options.name}_options.json`;
-    const path_to_option_json_file_name = path.join(path_to_json_bucket_options_dir, option_json_file_name);
-    const content = JSON.stringify(bucket_options);
+    const path_to_option_json_file_name = await create_json_file(path_to_json_bucket_options_dir, option_json_file_name, bucket_options);
+    return path_to_option_json_file_name;
+}
+
+/** 
+ * create_json_file would create a JSON file with the data
+ * @param {string} path_to_dir
+ * @param {string} file_name
+ * @param {object} data
+ */
+async function create_json_file(path_to_dir, file_name, data) {
+    const path_to_option_json_file_name = path.join(path_to_dir, file_name);
+    const content = JSON.stringify(data);
     await fs.promises.writeFile(path_to_option_json_file_name, content);
     return path_to_option_json_file_name;
 }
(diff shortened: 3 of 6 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -143,7 +143,8 @@ async function fetch_bucket_data(action, user_input) {
         path: user_input.path,
         should_create_underlying_storage: action === ACTIONS.ADD ? false : undefined,
         new_name: _.isUndefined(user_input.new_name) ? undefined : String(user_input.new_name),
-        fs_backend: _.isUndefined(user_input.fs_backend) ? config.NSFS_NC_STORAGE_BACKEND : String(user_input.fs_backend)
+        fs_backend: _.isUndefined(user_input.fs_backend) ? config.NSFS_NC_STORAGE_BACKEND : String(user_input.fs_backend),
+        force: _.isUndefined(user_input.force) ? undefined : get_boolean_or_string_value(user_input.force),
```
shirady, line 147: `force` is not a property of a bucket.
you can define this option when you use it in `delete_bucket`.

src/cmd/manage_nsfs.js
```diff
@@ -301,13 +302,18 @@ async function delete_bucket(data) {
     const bucket_config_path = get_config_file_path(buckets_dir_path, data.name);
     try {
         const bucket_temp_dir_path = path.join(data.path, config.NSFS_TEMP_DIR_NAME + "_" + data._id);
-        await native_fs_utils.folder_delete(bucket_temp_dir_path, fs_context_fs_backend, true);
-        await native_fs_utils.delete_config_file(fs_context_config_root_backend, buckets_dir_path, bucket_config_path);
+        const entries = await nb_native().fs.readdir(fs_context_fs_backend, data.path);
+        const filter_entries = entries.filter(element => !element.name.endsWith(config.NSFS_TEMP_DIR_NAME + "_" + data._id));
```
shirady, line 306: I think it will be more readable to change the variable name to a more meaningful name (then `filter_entries`).

shirady, line 306: What about defining this as a variable `config.NSFS_TEMP_DIR_NAME + "_" + data._id`? you using it twice

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -161,6 +165,44 @@ describe('manage nsfs cli bucket flow', () => {
...
+
+        it('cli delete bucket not fail with force flag, when bucket path is not empty', async () => {
+            let action = ACTIONS.ADD;
+            const bucket_options = { name: "bucket4", owner: bucket_defaults.owner, path: bucket_defaults.path };
+            // write the json_file_options
+            const path_to_option_json_file_name = await create_json_bucket_options(path_to_json_bucket_options_dir, bucket_options);
+            const command_flags = {config_root, from_file: path_to_option_json_file_name};
+            // create the bucket and check the details
+            await exec_manage_cli(TYPES.BUCKET, action, command_flags);
+            const bucket = await read_config_file(config_root, CONFIG_SUBDIRS.BUCKETS, "bucket4");
```
shirady, line 177: Why are you using the from-file bucket creation?
I think it is better to create the bucket using the command either in the `beforeEach`.

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -161,6 +165,44 @@ describe('manage nsfs cli bucket flow', () => {
...
+            assert_bucket(bucket, bucket_options);
+            //here a dummy file is creating in bucket storage location(bucket_defaults.path), 
+            //using existing config file creation method(create_json_bucket_options) and bucket options code for this
+            await create_json_bucket_options(bucket_defaults.path, bucket_options);
+            const delet_bucket_options = { config_root, name: "bucket4", force: true};
+            action = ACTIONS.DELETE;
+            const resp = await exec_manage_cli(TYPES.BUCKET, action, delet_bucket_options);
+            expect(JSON.parse(resp.trim()).response.code).toBe(ManageCLIResponse.BucketDeleted.code);
```
shirady, line 185: What about checking that the folder doesn't exist anymore? (like in "cli delete bucket and delete temp dir")
```
            path_exists = await is_path_exists(DEFAULT_FS_CONFIG, bucket_temp_dir_path);
            expect(path_exists).toBe(false)
```

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -161,6 +165,44 @@ describe('manage nsfs cli bucket flow', () => {
             path_exists = await is_path_exists(DEFAULT_FS_CONFIG, bucket_temp_dir_path);
             expect(path_exists).toBe(false);
         });
+
+        it('cli delete bucket not fail with force flag, when bucket path is not empty', async () => {
```
shirady, line 169:
```suggestion
        it('cli delete bucket with force flag, when bucket path is not empty', async () => {
```

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -161,6 +165,44 @@ describe('manage nsfs cli bucket flow', () => {
...
+            //here a dummy file is creating in bucket storage location(bucket_defaults.path), 
+            //using existing config file creation method(create_json_bucket_options) and bucket options code for this
+            await create_json_bucket_options(bucket_defaults.path, bucket_options);
+            const delet_bucket_options = { config_root, name: "bucket4", force: true};
+            action = ACTIONS.DELETE;
+            const resp = await exec_manage_cli(TYPES.BUCKET, action, delet_bucket_options);
+            expect(JSON.parse(resp.trim()).response.code).toBe(ManageCLIResponse.BucketDeleted.code);
+        });
+
+        it('should fail - cli delete bucket fail when bucket path is not empty', async () => {
```
shirady, line 188:
```suggestion
        it('should fail - cli delete bucket when bucket path is not empty (not using force flag)', async () => {
```

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -161,6 +165,44 @@ describe('manage nsfs cli bucket flow', () => {
...
+            // write the json_file_options
+            const path_to_option_json_file_name = await create_json_bucket_options(path_to_json_bucket_options_dir, bucket_options);
+            const command_flags = {config_root, from_file: path_to_option_json_file_name};
+            // create the bucket and check the details
+            await exec_manage_cli(TYPES.BUCKET, action, command_flags);
+            const bucket = await read_config_file(config_root, CONFIG_SUBDIRS.BUCKETS, "bucket4");
+            assert_bucket(bucket, bucket_options);
+            //here a dummy file is creating in bucket storage location(bucket_defaults.path), 
+            //using existing config file creation method(create_json_bucket_options) and bucket options code for this
+            await create_json_bucket_options(bucket_defaults.path, bucket_options);
```
shirady, line 181: I think it is confusing to use the bucket option as a dummy file.
Create a separate function for this purpose so it will be clear.

src/manage_nsfs/manage_nsfs_help_utils.js
```diff
@@ -180,6 +180,7 @@ bucket list [flags]
 Flags:
 --wide                                                (optional)        Print the additional details for each bucket
 --name <string>                                       (optional)        Filter the list based on the provided bucket name
+--force                                            (optional)        Forcefully delete bucket if the bucket is not empty
```
shirady, line 183: Please make sure it is aligned (bracket under bracket in "(optional)" and the first letter in "Forcefully" under "Filter".

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -106,6 +108,7 @@ describe('manage nsfs cli bucket flow', () => {
             await fs_utils.create_fresh_path(`${config_root}/${CONFIG_SUBDIRS.BUCKETS}`);
             await fs_utils.create_fresh_path(root_path);
             await fs_utils.create_fresh_path(bucket_storage_path);
+            await fs_utils.create_fresh_path(path_to_json_bucket_options_dir);
```
shirady, line 111: "json bucket options" is related to "from_file" in bucket add and not related to delete, please change it.

shirady:
@naveenpaul1, I have more questions/comments:
1. What happens to the objects when using the `--force` option? are they deleted?
2. The function `folder_delete` is only "delete bucket specific temp folder from bucket storage path, config.NSFS_TEMP_DIR_NAME_<bucket_id>" if not, better update the description as well:
https://github.com/noobaa/noobaa-core/blob/83166c69ab308d7b37b2656a5561c972ca720889/src/util/native_fs_utils.js#L529-L534
3. Related to the new boolean variable `force`, please add it here as well:
https://github.com/noobaa/noobaa-core/blob/83166c69ab308d7b37b2656a5561c972ca720889/src/cmd/manage_nsfs.js#L934-L937
and add these tests:
a. 'cli bucket delete - use explicit flag force true' --> pass `force` as 'true' string
b. 'cli bucket delete - use explicit flag force true (case insensitive)' --> pass `force` as 'TRUE' string
c. 'cli bucket delete - use explicit flag force false' --> pass `force` as 'false' string (you can do it on an empty bucket so it won't gets an error)
d. 'cli bucket delete - use explicit flag force false (case insensitive)' --> pass `force` as 'FALSE' string (you can do it on an empty bucket so it won't gets an error)

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
naveenpaul1: We do not delete the folder in cli delete flow, we can check for schema file exist or not along with existing check

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
naveenpaul1: Dont see much change in meaning

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
naveenpaul1: force flag is always optional and no need to mention it not using

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
naveenpaul1: used exising code for creating a dummy file, dont wanted add code that do same thing again, I will try to update function name to more generic name

naveenpaul1:
@shirady 
1. method will delete only the temp dir for us, no need to update description
3. will add more test cases

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
shirady: @naveenpaul1  I'm not sure I understand this, I copied it from an existing test, would you please explain it in more detail?

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
shirady: o.k., so please remove the duplicate fail.

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
shirady: I don't see why you need the function for a "dummy" file you can just use ` await fs.promises.writeFile`.

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
shirady: It is not in the meaning but aligns with the style of the `it` that we have: starts with "should fail" a test that gets an error, otherwise the test is passing (not failing).
[Here](https://github.com/noobaa/noobaa-core/pull/7913/files#diff-978f6ca26a95a48408c53fd0362cbd1bc795a15ee504d248f198a4576a3e745a) I'm trying to add it where tests do not have it.

naveenpaul1: @shirady I dont see any issue with explicitly mentioning a test is not failing in description of a test.

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
naveenpaul1: updated the code , code will check for the bucket schema

> const config_path = path.join(config_root, CONFIG_SUBDIRS.BUCKETS, bucket_five + '.json');
> await fs_utils.file_must_not_exist(config_path);

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
shirady: Not an issue,  just aligning with the code style of the file.

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -161,6 +162,62 @@ describe('manage nsfs cli bucket flow', () => {
             path_exists = await is_path_exists(DEFAULT_FS_CONFIG, bucket_temp_dir_path);
             expect(path_exists).toBe(false);
         });
+
+        it('cli delete bucket do not fail with force flag, when bucket path is not empty', async () => {
+            const bucket_four = 'bucket4';
+            const bucket_options = { config_root, name: bucket_four, owner: bucket_defaults.owner, path: bucket_defaults.path };
```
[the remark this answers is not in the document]
naveenpaul1: we are deleting the bucket in this flow dont wanted to have issues when multiple test case running parallel.

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
naveenpaul1: Deleting temp dir is covered in different test case, why we need to add that scenario here again?

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
naveenpaul1: its ok

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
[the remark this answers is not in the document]
shirady: Did you see an issue with this?
After every test, we have the `afterEach` which deletes the `config_root` and the `root_path`.
https://github.com/noobaa/noobaa-core/blob/99fc38f3060c776a6f89dd5cd241c1860d587c7d/src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js#L133-L136

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
shirady: No need to, if you think you already covered both things, then it is fine just check the response (as you did and remove the additional check of the config if you think it is already covered).

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -161,6 +162,62 @@ describe('manage nsfs cli bucket flow', () => {
...
+            //here a dummy file is creating in bucket storage location(bucket_defaults.path), 
+            await create_json_file(bucket_defaults.path, 'test1.json', {test: 'data'});
+            const delet_bucket_options = { config_root, name: bucket_three};
+            const resp = await exec_manage_cli(TYPES.BUCKET, ACTIONS.DELETE, delet_bucket_options);
+            expect(JSON.parse(resp.stdout).error.code).toBe(ManageCLIError.BucketDeleteForbiddenHasObjects.code);
+            const config_path = path.join(config_root, CONFIG_SUBDIRS.BUCKETS, bucket_three + '.json');
+            await fs_utils.file_must_exist(config_path);
+        });
+
+        it('cli delete bucket force flag with valid boolean string(\'true\')', async () => {
```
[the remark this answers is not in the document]
naveenpaul1: 1. temp folder delete scenario is already tested in above test case not sure why you wanted test it again.
2. here you wanted to check the type cast is working properly or not, right? its covered in this two test cases.

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -15,10 +15,11 @@ const { set_path_permissions_and_owner, TMP_PATH, generate_s3_policy } = require
 const { ACTIONS, TYPES, CONFIG_SUBDIRS } = require('../../../manage_nsfs/manage_nsfs_constants');
 const { get_process_fs_context, is_path_exists } = require('../../../util/native_fs_utils');
 const ManageCLIError = require('../../../manage_nsfs/manage_nsfs_cli_errors').ManageCLIError;
+const { ManageCLIResponse } = require('../../../manage_nsfs/manage_nsfs_cli_responses');
 
 const tmp_fs_path = path.join(TMP_PATH, 'test_nc_nsfs_bucket_cli');
 const DEFAULT_FS_CONFIG = get_process_fs_context();
-
+const timeout = 50000;
```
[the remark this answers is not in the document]
naveenpaul1: I was getting time out error for local test also

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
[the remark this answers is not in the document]
naveenpaul1: @shirady ya that make sence

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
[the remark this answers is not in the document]
shirady: 1. No need for the delete checks (temp dir delete, bucket config remove).
2. Yes, I meant for more cases regarding the new boolean variable:
   - 'cli bucket delete - use explicit flag force true (case insensitive)' --> pass force as 'TRUE' string
   - 'cli bucket delete - use explicit flag force false' --> pass force as 'false' string
   - 'cli bucket delete - use explicit flag force false (case insensitive)' --> pass force as 'FALSE' string 

The `false` string check was here because if someone would use converting of `Boolean()` he would get `true` instead of `false`, those tests are for protecting the manual changes that we allowed (if one day the CLI framework will change).

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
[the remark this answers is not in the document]
shirady: interesting, maybe we need to at it because the whole tests are currently running in less than the defined timeout.

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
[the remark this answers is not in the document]
naveenpaul1: I think you are already doing it for `wide` and same function will be used for force also, dont see the value its going to add here

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
naveenpaul1: no issues, I will keep both checks

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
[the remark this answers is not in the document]
naveenpaul1: removed timeout

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -161,6 +161,44 @@ describe('manage nsfs cli bucket flow', () => {
...
+            const delete_bucket_options = { config_root, name: bucket_defaults.name, force: 'true'};
+            const resp = await exec_manage_cli(TYPES.BUCKET, ACTIONS.DELETE, delete_bucket_options);
+            expect(JSON.parse(resp.trim()).response.code).toBe(ManageCLIResponse.BucketDeleted.code);
+            const config_path = path.join(config_root, CONFIG_SUBDIRS.BUCKETS, bucket_defaults.name + '.json');
+            await fs_utils.file_must_not_exist(config_path);
+        });
+
+        it('should fail - cli delete bucket force flag with invalid boolean string(\'nottrue\')', async () => {
+            const bucket_six = 'bucket6';
+            const delete_bucket_options = { config_root, name: bucket_six, force: 'nottrue'};
```
shirady, line 198: Why not use the default bucket name?
```suggestion
            const delete_bucket_options = { config_root, bucket_defaults.name, force: 'nottrue'};
```

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -212,7 +250,7 @@ describe('cli create bucket using from_file', () => {
         // write the json_file_options
         const path_to_option_json_file_name = await create_json_bucket_options(path_to_json_bucket_options_dir, bucket_options);
         const command_flags = {config_root, from_file: path_to_option_json_file_name};
-        // create tha bucket and check the details
+        // create the bucket and check the details
```
shirady, line 253: Thanks!

src/cmd/manage_nsfs.js
```diff
@@ -292,21 +292,27 @@ async function update_bucket(data) {
...
     const bucket_config_path = get_config_file_path(buckets_dir_path, data.name);
     try {
-        const bucket_temp_dir_path = path.join(data.path, config.NSFS_TEMP_DIR_NAME + "_" + data._id);
-        await native_fs_utils.folder_delete(bucket_temp_dir_path, fs_context_fs_backend, true);
-        await native_fs_utils.delete_config_file(fs_context_config_root_backend, buckets_dir_path, bucket_config_path);
+        const temp_dir_name = config.NSFS_TEMP_DIR_NAME + "_" + data._id;
+        const bucket_temp_dir_path = path.join(data.path, temp_dir_name);
+        const entries = await nb_native().fs.readdir(fs_context_fs_backend, data.path);
+        const object_entries = entries.filter(element => !element.name.endsWith(temp_dir_name));
+        if (object_entries.length === 0 || (force && true)) {
```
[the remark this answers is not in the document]
naveenpaul1: I accidentally added `true`, thanks

src/manage_nsfs/manage_nsfs_help_utils.js
```diff
@@ -180,6 +180,7 @@ bucket list [flags]
 Flags:
 --wide                                                (optional)        Print the additional details for each bucket
 --name <string>                                       (optional)        Filter the list based on the provided bucket name
+--force                                               (optional)        Forcefully delete bucket if the bucket is not empty
```
shirady, line 183: Please move this to `BUCKET_FLAGS_DELETE` (currently it is in `BUCKET_FLAGS_LIST`).

shirady (approved):
LGTM
only minor change in help

romayalon:
Hey @naveenpaul1  can we add some tests to the no-force option? I don't see any

Issue #6615, opened 2021-06-28
Improve list objects performance on top of NS FS


nimrod-becker:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. All directory entries are brought to memory and ordered (FS doesn't return it in order), saved in cache 
2. If the directory size is larger than the cache size (64MB defaulted). then on each list command we bring everything to mem and organize

### Expected behavior
1. Add a config per account to return un-ordered list object (list v2 only)
2. Still need to support pagination and delimiter, prefix 

### Steps to reproduce
1.

### More information - Screenshots / Logs / Other output
Should / Can use 
https://man7.org/linux/man-pages/man3/telldir.3.html
https://man7.org/linux/man-pages/man3/seekdir.3.html

PR #7912, opened 2024-03-19
NSFS | Improve list objects performance on top of NS FS (PR 1/3)

naveenpaul1:
### Explain the changes
1. This is first PR for listing unordered objects.
2. This Pr contains implementation of `telldir` and `seekdir` methods in napi layer 
3. Global flag is added in config for controlling bucket list flow

1. https://github.com/noobaa/noobaa-core/issues/6615

### Testing Instructions:
1. Not possible with current changes.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -243,6 +243,7 @@ struct Entry
     std::string name;
     ino_t ino;
     uint8_t type;
+    long int off;
 };
 
 struct gpfsRequest_t
@@ -1231,6 +1232,7 @@ struct Readdir : public FSWorker
                     std::string(e->d_name),
                     e->d_ino,
                     e->d_type,
+                    e->d_off,
                 });
             } else {
                 if (errno) SetSyscallError();
@@ -1253,6 +1255,7 @@ struct Readdir : public FSWorker
             dir_rec["name"] = Napi::String::New(env, it->name);
             dir_rec["ino"] = Napi::Number::New(env, it->ino);
             dir_rec["type"] = Napi::Number::New(env, it->type);
+            dir_rec["off"] = Napi::BigInt::New(env, it->off);
             res[index] = dir_rec;
             index += 1;
         }
@@ -1889,6 +1892,8 @@ struct DirWrap : public Napi::ObjectWrap<DirWrap>
             {
                 InstanceMethod("close", &DirWrap::close),
                 InstanceMethod("read", &DirWrap::read),
+                InstanceMethod("telldir", &DirWrap::telldir),
+                InstanceMethod("seekdir", &DirWrap::seekdir),
             }));
         constructor.SuppressDestruct();
     }
@@ -1908,6 +1913,8 @@ struct DirWrap : public Napi::ObjectWrap<DirWrap>
     }
     Napi::Value close(const Napi::CallbackInfo& info);
     Napi::Value read(const Napi::CallbackInfo& info);
+    Napi::Value telldir(const Napi::CallbackInfo& info);
+    Napi::Value seekdir(const Napi::CallbackInfo& info);
 };
 
 Napi::FunctionReference DirWrap::constructor;
@@ -1938,6 +1945,59 @@ struct DirOpen : public FSWorker
         w->_dir = _dir;
         _deferred.Resolve(res);
         ReportWorkerStats(0);
+    }
+};
+
+struct TellDir : public FSWrapWorker<DirWrap>
+{
+    long int _tell_res = -1;
+    TellDir(const Napi::CallbackInfo& info)
+        : FSWrapWorker<DirWrap>(info)
+    {
+        Begin(XSTR() << "TellDir " << DVAL(_wrap->_path));
+    }
+    virtual void Work()
+    {
+        DIR* dir = _wrap->_dir;
+        if (!dir) {
+            SetError(XSTR() << "FS::TellDir::Execute: ERROR not opened " << _wrap->_path);
+            return;
+        }
+        DBG1("FS::Telldir::Work: " << DVAL(_wrap->_path));
+        _tell_res = telldir(dir);
+        if (_tell_res == -1) SetSyscallError();
+    }
+    virtual void OnOK()
+    {
+        DBG0("FS::Telldir::OnOK: " << DVAL(_wrap->_path) << DVAL(_tell_res));
+        Napi::Env env = Env();
+        auto res = Napi::BigInt::New(env, _tell_res);
+        _deferred.Resolve(res);
+    }
+};
+
+struct SeekDir : public FSWrapWorker<DirWrap>
+{
+    long int _seek_pos;
+    SeekDir(const Napi::CallbackInfo& info)
+        : FSWrapWorker<DirWrap>(info)
+    {
+        bool lossless = true;
+        _seek_pos = info[1].As<Napi::BigInt>().Int64Value(&lossless);
+        if (!lossless) {
+            SetError("seekdir bigint value must convert lossless");
+        }
+        Begin(XSTR() << "SeekDir " << DVAL(_wrap->_path) << DVAL(_seek_pos));
+    }
+    virtual void Work()
+    {
+        DIR* dir = _wrap->_dir;
+        if (!dir) {
+            SetError(XSTR() << "FS::SeekDir::Execute: ERROR not opened " << _wrap->_path);
+            return;
+        }
+        DBG1("FS::SeekDir::Work: " << DVAL(_wrap->_path));
+        seekdir(dir, _seek_pos);
     }
 };
 
@@ -1988,6 +2048,7 @@ struct DirReadEntry : public FSWrapWorker<DirWrap>
             _entry.name = std::string(e->d_name);
             _entry.ino = e->d_ino;
             _entry.type = e->d_type;
+            _entry.off = e->d_off;
         } else {
             if (errno) {
                 SetSyscallError();
@@ -2006,6 +2067,7 @@ struct DirReadEntry : public FSWrapWorker<DirWrap>
             res["name"] = Napi::String::New(env, _entry.name);
             res["ino"] = Napi::Number::New(env, _entry.ino);
             res["type"] = Napi::Number::New(env, _entry.type);
+            res["off"] = Napi::BigInt::New(env, _entry.off);
             _deferred.Resolve(res);
         }
         ReportWorkerStats(0);
@@ -2022,6 +2084,18 @@ Napi::Value
 DirWrap::read(const Napi::CallbackInfo& info)
 {
     return api<DirReadEntry>(info);
+}
+
+Napi::Value
+DirWrap::telldir(const Napi::CallbackInfo& info)
+{
+    return api<TellDir>(info);
+}
+
+Napi::Value
+DirWrap::seekdir(const Napi::CallbackInfo& info)
+{
+    return api<SeekDir>(info);
 }
 
 static Napi::Value
--- a/src/test/unit_tests/jest_tests/test_list_object.test.js
+++ b/src/test/unit_tests/jest_tests/test_list_object.test.js
@@ -1,0 +1,176 @@
+/* Copyright (C) 2016 NooBaa */
+/* eslint-disable no-undef */
+'use strict';
+
+
+const fs = require('fs');
+const path = require('path');
+const fs_utils = require('../../../util/fs_utils');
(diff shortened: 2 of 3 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -1939,6 +1943,52 @@ struct DirOpen : public FSWorker
...
+        auto res = Napi::Object::New(env);
+        res["telldir_res"] = Napi::Number::New(env, _tell_res);
+        _deferred.Resolve(res);
+    }
+};
+
+struct SeekDir : public FSWrapWorker<DirWrap>
+{
+    std::string _dir_path;
+    long int _seek_pos;
```
romayalon, line 1976: why long int?
I see it's a long - https://man7.org/linux/man-pages/man3/seekdir.3.html

src/native/fs/fs_napi.cpp
```diff
@@ -1939,6 +1943,52 @@ struct DirOpen : public FSWorker
...
+        DBG1("FS::Telldir::Work: " << DVAL(_dir_path));
+        _tell_res = telldir(dir);
+        if (_tell_res == -1) SetSyscallError();
+    }
+    virtual void OnOK()
+    {
+        DBG0("FS::Telldir::OnOK: " << DVAL(_dir_path) << DVAL(_tell_res));
+        Napi::Env env = Env();
+        auto res = Napi::Object::New(env);
+        res["telldir_res"] = Napi::Number::New(env, _tell_res);
```
romayalon, line 1968: Do we want to return a number or an object?
if a number, the return value should look like - 
```
Napi::Env env = Env();
auto res = Napi::Number::New(env, _tell_res);
_deferred.Resolve(res);
```

romayalon (changes requested):

romayalon:
@naveenpaul1 let's add some tests? :)

src/native/fs/fs_napi.cpp
guymguym: int is optional in these types - long=long int , short=short int, signed=signed int, unsigned=unsigned int.

src/native/fs/fs_napi.cpp
guymguym: @romayalon @naveenpaul1 @eshelmarc 

First I want to verify if we might need to support 64bit dir cookie - I see some traces of such implementations, but no clear documentation so might not be posix:
- [IBM AIX-telldir64](https://www.ibm.com/docs/en/aix/7.3?topic=o-opendir-readdir-telldir-seekdir-rewinddir-closedir-opendir64-readdir64-telldir64-seekdir64-rewinddir64-closedir64-fdopendir-subroutine)
- [Ext4 64 bit hash breaks 32 bit glibc 2.28+](https://bugzilla.kernel.org/show_bug.cgi?id=205957) 

However libc [telldir](https://sourceware.org/git/?p=glibc.git;a=blob;f=dirent/dirent.h;h=b33d6296e3b1889414a604eb9e99f96ee60c04b6;hb=HEAD#l220)/[seekdir](https://sourceware.org/git/?p=glibc.git;a=blob;f=dirent/dirent.h;h=b33d6296e3b1889414a604eb9e99f96ee60c04b6;hb=HEAD#l217) has fixed this type on 32bit long int.

In any case, I think our interface should return this as an opaque instead of a js number - I would go with a hex string of the bytes that compose this cookie number, and parse back on seekdir. We can also return a Buffer, but I don't know if this is really better - js is notoriously good with managing short strings.

config.js
```diff
@@ -703,6 +703,8 @@ config.NSFS_TEMP_DIR_NAME = '.noobaa-nsfs';
 
 config.NSFS_FOLDER_OBJECT_NAME = '.folder';
 
+config.NSFS_UNSORTED_OBJECTS_LIST_ENABLED = true;
```
guymguym, line 706: Is this already used by default in this PR? I think this is just confusing right now, so I'd keep it for the next PR improving the list performance, then we can decide how to pick best defaults.

src/native/fs/fs_napi.cpp
```diff
@@ -1887,6 +1887,8 @@ struct DirWrap : public Napi::ObjectWrap<DirWrap>
             {
                 InstanceMethod("close", &DirWrap::close),
                 InstanceMethod("read", &DirWrap::read),
+                InstanceMethod("telldir", &DirWrap::tell),
+                InstanceMethod("seekdir", &DirWrap::seek),
```
guymguym, line 1891: I would keep the DirWrap methods named the same to avoid having to do a "lookup" when trying to locate a function name in the code, so DirWrap::telldir/seekdir is better IMHO (even that it can be understood already that this is about a directory from the class...).

guymguym (changes requested):
@naveenpaul1 This is great!
It would be great to see the performance improvement from the full impl.
See few comments.

src/native/fs/fs_napi.cpp
```diff
@@ -1941,6 +1945,53 @@ struct DirOpen : public FSWorker
...
+        std::string _dir_path = _wrap->_path;
+        DBG1("FS::Telldir::Work: " << DVAL(_dir_path));
+        _tell_res = telldir(dir);
+        if (_tell_res == -1) SetSyscallError();
+    }
+    virtual void OnOK()
+    {
+        DBG0("FS::Telldir::OnOK: " << DVAL(_dir_path) << DVAL(_tell_res));
+        char telldir_loc[256];
+        sprintf(telldir_loc, "%ld", _tell_res);
```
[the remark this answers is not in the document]
naveenpaul1: @guymguym I reverted the changes,
I didnt get the your suggestion there, After getting more understating I will implement it in next related PRs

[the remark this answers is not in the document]
guymguym: sure, no worries, it was mostly a concern for supporting different platforms where the dir pos might be 64 bit, but for now since we use libc for that the type is fixed on long so no real concern there. we can followup if needed on next PR's like you said.

src/native/fs/fs_napi.cpp
```diff
@@ -1941,6 +1945,51 @@ struct DirOpen : public FSWorker
     }
 };
 
+struct TellDir : public FSWrapWorker<DirWrap>
+{
+    std::string _dir_path;
```
guymguym, line 1950: I dont think storing this as a member is needed, you can just access it from _wrap->_path

guymguym (approved):
Just two small comments but otherwise looks good.

src/native/fs/fs_napi.cpp
```diff
@@ -1941,6 +1945,51 @@ struct DirOpen : public FSWorker
...
+    long int _tell_res;
```
guymguym, line 1951: Better initialize this to -1 in ctor

src/native/fs/fs_napi.cpp
guymguym: OK, not sure why but I got confused here and was corrected - long int is int64 on a 64bit system.
So this is why libc used it, and this is also the reason why we can't just stick it in js number (which is a float and represents ints with up to 53 bits).

So I suggest is to change this to return js BigInt which can hold 64bit int without a problem.
See https://github.com/nodejs/node-addon-api/blob/main/doc/bigint.md

guymguym (changes requested):
First - we need to switch to BigInt - see https://github.com/noobaa/noobaa-core/pull/7912#discussion_r1560062442

Second - I see that readdir also returns the same position that telldir returns as `d_off` -
https://man7.org/linux/man-pages/man3/readdir.3.html

So we should also return this number (as bigint) from our dir.read() call so that we don't have to call telldir again if we just read the last entry where we want to pause. This means we should add it to our struct Entry (in this file).

src/native/fs/fs_napi.cpp
```diff
@@ -1941,6 +1948,49 @@ struct DirOpen : public FSWorker
     }
 };
 
+struct TellDir : public FSWrapWorker<DirWrap>
+{
+    std::string _dir_path;
```
guymguym, line 1953: remove unneeded member

src/native/fs/fs_napi.cpp
```diff
@@ -1941,6 +1948,49 @@ struct DirOpen : public FSWorker
...
+    }
+};
+
+struct SeekDir : public FSWrapWorker<DirWrap>
+{
+    long int _seek_pos;
+    SeekDir(const Napi::CallbackInfo& info)
+        : FSWrapWorker<DirWrap>(info)
+    {
+        _seek_pos = info[1].As<Napi::Number>();
```
guymguym, line 1983: change to bigint

src/test/unit_tests/jest_tests/test_list_object.test.js
```diff
@@ -0,0 +1,155 @@
...
+            let tell_dir;
+            let dir_marker;
+            let total_dir_entries = 0;
+            let dir_entry;
+            let dir_handle = await nb_native().fs.opendir(DEFAULT_FS_CONFIG, list_dir_root);
+            // reak first read after 3 entries.
+            for (let i = 0; i <= 2; i++) {
+                dir_entry = await dir_handle.read(DEFAULT_FS_CONFIG);
+                if (!dir_entry) break;
+                tell_dir = await dir_handle.telldir(DEFAULT_FS_CONFIG);
```
guymguym, line 46: check here that the off returned in the direntry is the same as returned by telldir

src/test/unit_tests/jest_tests/test_list_object.test.js
```diff
@@ -0,0 +1,155 @@
...
+                dir_marker = {
+                    dir_path: list_dir_root,
+                    pos: dir_entry.off,
+                };
+                total_dir_entries += 1;
+            }
+            // Continue the read using dir location fetch from Dir.read()
+            try {
+                dir_handle = await nb_native().fs.opendir(DEFAULT_FS_CONFIG, dir_marker.dir_path);
+                await dir_handle.seekdir(DEFAULT_FS_CONFIG, Number(dir_marker.pos));
```
guymguym, line 89: should pass as bigint, not number

guymguym (changes requested):

src/native/fs/fs_napi.cpp
```diff
@@ -1941,6 +1948,48 @@ struct DirOpen : public FSWorker
...
+};
+
+struct SeekDir : public FSWrapWorker<DirWrap>
+{
+    long int _seek_pos;
+    bool lossless = true;
+    SeekDir(const Napi::CallbackInfo& info)
+        : FSWrapWorker<DirWrap>(info)
+    {
+        _seek_pos = info[1].As<Napi::BigInt>().Int64Value(&lossless);
```
guymguym, line 1982: lossless should be local var, not member.
and it should be checked that it was indeed converted to int64 without losing data.
see proposed code:
```suggestion
        bool lossless = true;
        _seek_pos = info[1].As<Napi::BigInt>().Int64Value(&lossless);
        if (!lossless) {
            SetError("seekdir bigint value must convert lossless");
        }
```

src/native/fs/fs_napi.cpp
```diff
@@ -1941,6 +1948,51 @@ struct DirOpen : public FSWorker
...
+{
+    long int _tell_res = -1;
+    TellDir(const Napi::CallbackInfo& info)
+        : FSWrapWorker<DirWrap>(info)
+    {
+        Begin(XSTR() << "TellDir " << DVAL(_wrap->_path));
+    }
+    virtual void Work()
+    {
+        DIR* dir = _wrap->_dir;
```
guymguym, line 1961: add a check like we do in DirReadEntry for the case that dir was closed - if (!dir) ...

src/native/fs/fs_napi.cpp
```diff
@@ -1941,6 +1948,51 @@ struct DirOpen : public FSWorker
...
+        bool lossless = true;
+        _seek_pos = info[1].As<Napi::BigInt>().Int64Value(&lossless);
+        if (!lossless) {
+            SetError("seekdir bigint value must convert lossless");
+        }
+        Begin(XSTR() << "SeekDir " << DVAL(_wrap->_path) << DVAL(_seek_pos));
+    }
+    virtual void Work()
+    {
+        DIR* dir = _wrap->_dir;
```
guymguym, line 1990: add a check like we do in DirReadEntry for the case that dir was closed - if (!dir) ...

guymguym (approved):
Looks good. Lets get it merged.

romayalon (approved):

Issue #7616, opened 2023-11-23
NSFS | NC | CLI | Add force_md5_etag option for account and bucket schema/create/update 


romayalon:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. Currently a user can not set force_md5_etag property to account/bucket.

### Expected behavior
1. A user can not set force_md5_etag property to account/bucket.

### Steps to reproduce
1.

nimrod-becker:
Similar to what we have in regular deployments, add this option to an account config

nimrod-becker:
Ref to usage in namespace_fs
https://github.com/noobaa/noobaa-core/blob/d1bf8c465af745dedbff9d3803ec13da344884ba/src/sdk/namespace_fs.js#L1607

shirady:
Hi,
I have an open question (also described in the PR [comment](https://github.com/noobaa/noobaa-core/pull/7915#discussion_r1531909183)).
If I have a general configuration with `config.NSFS_CALCULATE_MD5` `true`.
Someone use the Manage NSFS CLI and decides that there is an account that will be a test account and creates it with `--force_md5_etag false`, what do we think that the value of `md5_enabled` should be?
Currently, the `md5_enabled` is still `true`.

cc: @romayalon @naveenpaul1 @nadavMiz

nimrod-becker:
Specific Bucket/Account settings override global settings

shirady:
@nimrod-becker @nadavMiz  it doesn't happen currently, do you want this change in the PR or a different one?
The current PR can be approved if we know it is a GAP to close (since it doesn't relate to the manage NSFS CLI).

shirady:
Closing, @nadavMiz fixed it in PR #7915.
Adding request validation label.
cc: @romayalon

PR #7915, opened 2024-03-20
NSFS | NC | CLI |  Add force_md5_etag option to account and bucket

nadavMiz:
### Explain the changes
1. add an option to set force_md5_etag to both account bucket add and update in the nsfs manager cli. add this option to the account and bucket schema
2. change force_md5_etag logic to first check if it is defined for the bucket, if not check if it is defined for the account, and if it isn't defined for both of them use default config.NSFS_CALCULATE_MD5.
3. add an option to unset the force_md5_etag flag to both the account and bucket

1. https://github.com/noobaa/noobaa-core/issues/7616

### Testing Instructions:
1. for bucket tests run: `sudo npx jest test_nc_nsfs_bucket_cli.test.js`
4. for account tests run `sudo npx jest test_nc_nsfs_account_cli.test.js`
5. for schema tests run ` sudo npx jest test_nc_nsfs_bucket_schema_validation.test.js`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -23,7 +23,7 @@ const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
 const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
 const { print_usage } = require('../manage_nsfs/manage_nsfs_help_utils');
 const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, FROM_FILE, BOOLEAN_STRING_VALUES,
-    LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS, GLACIER_ACTIONS } = require('../manage_nsfs/manage_nsfs_constants');
+    LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS, GLACIER_ACTIONS, LIST_UNSETABLE_OPTIONS } = require('../manage_nsfs/manage_nsfs_constants');
 const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 
 function throw_cli_error(error_code, detail, event_arg) {
@@ -143,7 +143,8 @@ async function fetch_bucket_data(action, user_input) {
         path: user_input.path,
         should_create_underlying_storage: action === ACTIONS.ADD ? false : undefined,
         new_name: _.isUndefined(user_input.new_name) ? undefined : String(user_input.new_name),
-        fs_backend: _.isUndefined(user_input.fs_backend) ? config.NSFS_NC_STORAGE_BACKEND : String(user_input.fs_backend)
+        fs_backend: _.isUndefined(user_input.fs_backend) ? config.NSFS_NC_STORAGE_BACKEND : String(user_input.fs_backend),
+        force_md5_etag: _.isUndefined(user_input.force_md5_etag) || user_input.force_md5_etag === '' ? user_input.force_md5_etag : get_boolean_or_string_value(user_input.force_md5_etag)
         };
 
     if (user_input.bucket_policy !== undefined) {
@@ -169,6 +170,8 @@ async function fetch_bucket_data(action, user_input) {
     data.fs_backend = data.fs_backend || undefined;
     // s3_policy deletion specified with empty string '' (but it is not part of the schema)
     data.s3_policy = data.s3_policy || undefined;
+    // force_md5_etag deletion specified with empty string '' checked against user_input because data.force_md5_etag is boolean
+    data.force_md5_etag = data.force_md5_etag === '' ? undefined : data.force_md5_etag;
 
     return data;
 }
@@ -376,6 +379,7 @@ async function fetch_account_data(action, user_input) {
         new_name: _.isUndefined(user_input.new_name) ? undefined : String(user_input.new_name),
         new_access_key,
         access_keys,
+        force_md5_etag: _.isUndefined(user_input.force_md5_etag) || user_input.force_md5_etag === '' ? user_input.force_md5_etag : get_boolean_or_string_value(user_input.force_md5_etag),
         nsfs_account_config: {
             distinguished_name: user_input.user,
             uid: user_input.user ? undefined : user_input.uid,
@@ -401,6 +405,8 @@ async function fetch_account_data(action, user_input) {
     data.nsfs_account_config.fs_backend = data.nsfs_account_config.fs_backend || undefined;
     // new_buckets_path deletion specified with empty string ''
     data.nsfs_account_config.new_buckets_path = data.nsfs_account_config.new_buckets_path || undefined;
+    // force_md5_etag deletion specified with empty string '' checked against user_input.force_md5_etag because data.force_md5_etag is boolean
+    data.force_md5_etag = data.force_md5_etag === '' ? undefined : data.force_md5_etag;
     // allow_bucket_creation either set by user or infer from new_buckets_path
     if (_.isUndefined(user_input.allow_bucket_creation)) {
         data.allow_bucket_creation = !_.isUndefined(data.nsfs_account_config.new_buckets_path);
@@ -937,12 +943,16 @@ function validate_options_type_by_value(input_options_with_data) {
         const type_of_option = OPTION_TYPE[option];
         const type_of_value = typeof value;
         if (type_of_value !== type_of_option) {
+            // special case for unset value (specified by '').
+            if (LIST_UNSETABLE_OPTIONS.includes(option) && value === '') {
+                continue;
+            }
             // special case for names, although the type is string we want to allow numbers as well
             if ((option === 'name' || option === 'new_name') && (type_of_value === 'number')) {
                 continue;
             }
             // special case for boolean values
-            if (['allow_bucket_creation', 'regenerate', 'wide', 'show_secrets', 'force'].includes(option) && validate_boolean_string_value(value)) {
+            if (['allow_bucket_creation', 'regenerate', 'wide', 'show_secrets', 'force', 'force_md5_etag'].includes(option) && validate_boolean_string_value(value)) {
                 continue;
             }
             // special case for bucket_policy (from_file)
--- a/src/manage_nsfs/manage_nsfs_constants.js
+++ b/src/manage_nsfs/manage_nsfs_constants.js
@@ -33,16 +33,16 @@ const GLOBAL_CONFIG_OPTIONS = new Set([GLOBAL_CONFIG_ROOT, 'config_root_backend'
 const FROM_FILE = 'from_file';
 
 const VALID_OPTIONS_ACCOUNT = {
-    'add': new Set(['name', 'uid', 'gid', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', FROM_FILE, ...GLOBAL_CONFIG_OPTIONS]),
-    'update': new Set(['name', 'uid', 'gid', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'new_name', 'regenerate', ...GLOBAL_CONFIG_OPTIONS]),
+    'add': new Set(['name', 'uid', 'gid', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', FROM_FILE, ...GLOBAL_CONFIG_OPTIONS]),
+    'update': new Set(['name', 'uid', 'gid', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'new_name', 'regenerate', ...GLOBAL_CONFIG_OPTIONS]),
     'delete': new Set(['name', ...GLOBAL_CONFIG_OPTIONS]),
     'list': new Set(['wide', 'show_secrets', 'gid', 'uid', 'user', 'name', 'access_key', ...GLOBAL_CONFIG_OPTIONS]),
     'status': new Set(['name', 'access_key', 'show_secrets', ...GLOBAL_CONFIG_OPTIONS]),
 };
 
 const VALID_OPTIONS_BUCKET = {
-    'add': new Set(['name', 'owner', 'path', 'bucket_policy', 'fs_backend', FROM_FILE, ...GLOBAL_CONFIG_OPTIONS]),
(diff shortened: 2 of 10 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -142,7 +142,8 @@ async function fetch_bucket_data(action, user_input) {
         path: user_input.path,
         should_create_underlying_storage: action === ACTIONS.ADD ? false : undefined,
         new_name: _.isUndefined(user_input.new_name) ? undefined : String(user_input.new_name),
-        fs_backend: _.isUndefined(user_input.fs_backend) ? config.NSFS_NC_STORAGE_BACKEND : String(user_input.fs_backend)
+        fs_backend: _.isUndefined(user_input.fs_backend) ? config.NSFS_NC_STORAGE_BACKEND : String(user_input.fs_backend),
+        force_md5_etag: _.isUndefined(user_input.force_md5_etag) ? undefined : get_boolean_or_string_value(user_input.force_md5_etag)
```
shirady, line 146: Question - what about a default value (`config.NSFS_CALCULATE_MD5`)?
like we have in `fs_backend` - I saw in the issue a reference to:
https://github.com/noobaa/noobaa-core/blob/d1bf8c465af745dedbff9d3803ec13da344884ba/src/sdk/namespace_fs.js#L1607

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -329,6 +338,69 @@ describe('cli create bucket using from_file', () => {
 
 });
 
+describe('cli update bucket', () => {
+    const config_root = path.join(tmp_fs_path, 'config_root_manage_nsfs2');
+    const root_path = path.join(tmp_fs_path, 'root_path_manage_nsfs2/');
```
shirady, line 343: I think it is better to make sure that these paths were not defined in the file (I think it appears in the 'cli delete bucket`). You can simply change the number in the suffix.

src/server/system_services/schemas/nsfs_account_schema.js
```diff
@@ -29,6 +29,9 @@ module.exports = {
         allow_bucket_creation: {
             type: 'boolean',
         },
+        force_md5_etag: {
+            type: 'boolean',
+        },
```
shirady, line 34: Since it is a schema change suggesting to add a test in the file `test_nc_nsfs_account_schema_validation.test.js`.
For example:
https://github.com/noobaa/noobaa-core/blob/d1bf8c465af745dedbff9d3803ec13da344884ba/src/test/unit_tests/jest_tests/test_nc_nsfs_account_schema_validation.test.js#L33-L38

src/server/system_services/schemas/nsfs_bucket_schema.js
```diff
@@ -59,6 +59,9 @@ module.exports = {
         },
         website: {
             $ref: 'common_api#/definitions/bucket_website',
+        },
+        force_md5_etag: {
+            type: 'boolean',
```
shirady, line 64: Since it is a schema change suggesting to add a test in the file `test_nc_nsfs_bucket_schema_validation.test.js`.
For example:
https://github.com/noobaa/noobaa-core/blob/d1bf8c465af745dedbff9d3803ec13da344884ba/src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_schema_validation.test.js#L56-L60

src/cmd/manage_nsfs.js
shirady: @naveenpaul1, you're right. When I wrote this comment I thought that maybe we need to use a default value in this property, but we use it in the file `namespace_fs.js` (hence I decided to resolve it).

src/test/unit_tests/jest_tests/test_nc_nsfs_account_schema_validation.test.js
```diff
@@ -329,6 +336,16 @@ describe('schema validation NC NSFS account', () => {
             const message = 'must be equal to one of the allowed values';
             assert_validation(account_data, reason, message);
         });
+
+        it('nsfs_account_config with force_md5_etag as a string (instead of boolean)', () => {
+            const account_data = get_account_data();
+            // @ts-ignore
+            account_data.force_md5_etag = ""; // added
+            const reason = 'Test should have failed because of wrong type for' +
```
shirady, line 344: I think there is a missing space before the bracket

src/cmd/manage_nsfs.js
shirady: @nadavMiz, the open question I have regarding this line:
https://github.com/noobaa/noobaa-core/blob/d1bf8c465af745dedbff9d3803ec13da344884ba/src/sdk/namespace_fs.js#L1607-L1608

`md5_enabled` will be `true` in one of the cases of the condition to `true`:
1. It was set in the config file (`config.NSFS_CALCULATE_MD5`).
2. It was set in the namespace_fs level (`this.force_md5_etag`).
3. It was set in the account (`object_sdk?.requesting_account?.force_md5_etag`).

The missing case is when it was set in the bucket (as you added in this PR).


Another thing that I don't understand is that if someone sets it in the config, and also adds for a specific account `--force_md5_etag false`, with this logical OR we would ignore it, is that what we want?

src/manage_nsfs/manage_nsfs_constants.js
```diff
@@ -33,16 +33,16 @@ const GLOBAL_CONFIG_OPTIONS = new Set([GLOBAL_CONFIG_ROOT, 'config_root_backend'
...
+    'update': new Set(['name', 'uid', 'gid', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'new_name', 'regenerate', ...GLOBAL_CONFIG_OPTIONS]),
     'delete': new Set(['name', ...GLOBAL_CONFIG_OPTIONS]),
     'list': new Set(['wide', 'show_secrets', 'gid', 'uid', 'user', 'name', 'access_key', ...GLOBAL_CONFIG_OPTIONS]),
     'status': new Set(['name', 'access_key', 'show_secrets', ...GLOBAL_CONFIG_OPTIONS]),
 };
 
 const VALID_OPTIONS_BUCKET = {
-    'add': new Set(['name', 'owner', 'path', 'bucket_policy', 'fs_backend', FROM_FILE, ...GLOBAL_CONFIG_OPTIONS]),
-    'update': new Set(['name', 'owner', 'path', 'bucket_policy', 'fs_backend', 'new_name', ...GLOBAL_CONFIG_OPTIONS]),
+    'add': new Set(['name', 'owner', 'path', 'bucket_policy', 'fs_backend', 'force_md5_etag', FROM_FILE, ...GLOBAL_CONFIG_OPTIONS]),
```
naveenpaul1, line 44: Nimrod added a [comment](https://github.com/noobaa/noobaa-core/issues/7616#issuecomment-1987915766) to is issue, about adding the `force_md5_etag` only to account. may be we wanted clarify that.

src/cmd/manage_nsfs.js
nadavMiz: this logic was added by @achouhan09, so maybe he can way in. but the idea is that `this.force_md5_etag` is the case where it was set in the bucket. so you have three cases:

1. It was set in the config file (config.NSFS_CALCULATE_MD5).
2. It was set in the account (object_sdk?.requesting_account?.force_md5_etag).
3. It was set in the bucket level (this.force_md5_etag).

regarding your other question the way I see it `force_md5_etag` is defaulted to be false unless it is said otherwise anywhere else, either for everybody in the config file, or for a specific bucket / account. 
@achouhan09 is this correct?

nadavMiz: this is the PR for adding force_md5_etag for buckets: https://github.com/noobaa/noobaa-core/pull/7236
note that this PR only exposes this feature to nsfs cli, it doesn't add new functionality

src/manage_nsfs/manage_nsfs_constants.js
nadavMiz: this comment is just a clarification of what needs to be done in the task. the task itself is about adding 'force_md5_etag' to both account and bucket schema as in the task description

src/cmd/manage_nsfs.js
achouhan09: yes @shirady , as @nadavMiz described in above comment that was the logic before. You can also refer to my PRs #7236 & #7297  for more info.

shirady: @achouhan09 thank you, I just still don't understand this scenario: if in the config file, it was set to `true` but in a specific account it is `false` (not `undefined` the user explicitly set a `false` value) - why do we still `md5_enabled` to be `true`?

achouhan09: @shirady here, I think we are only focusing if any of the above 3 cases is true then we need to calculate the MD5 etag, see Danny's comment [here](https://github.com/noobaa/noobaa-core/pull/7236#discussion_r1194156967). He mentioned that we need `config.NSFS_CALCULATE_MD5` as a global toggle, it will allow us to calculate MD5 etag globally.

shirady: @achouhan09 thank you.
@nadavMiz I will link this discussion in the issue as my open question.
resolving.

src/sdk/namespace_fs.js
```diff
@@ -1426,6 +1426,15 @@ class NamespaceFS {
         }
     }
 
+    _is_force_md5_enabled(object_sdk) {
+        let md5_enabled = config.NSFS_CALCULATE_MD5;
+        // If exists specific Bucket/Account settings should override global settings
+        if (!_.isUndefined(this.force_md5_etag) || !_.isUndefined(object_sdk?.requesting_account?.force_md5_etag)) {
+            md5_enabled = (this.force_md5_etag || object_sdk?.requesting_account?.force_md5_etag);
```
shirady, line 1433: If somehow the `NSFS_CALCULATE_MD5` in the account was set to `true` and the bucket was set to `false` what do we expect the value to be?
I thought it should be determined by the bucket (if exists), then by account (if exists), and then by config (must exist).

src/sdk/namespace_fs.js
```diff
@@ -1426,6 +1426,15 @@ class NamespaceFS {
         }
     }
 
+    _is_force_md5_enabled(object_sdk) {
```
shirady, line 1429: If you can please add JSDoc to the function (we try to add it in new functions that we add).

shirady, line 1429: I think you added this function under the "title" `// MULTIPART UPLOAD //` we also have the title `// INTERNALS //` which I think will help us in keep the file organized.

shirady, line 1429: Now that we know the use, I think that we need to have the option to "unset" it (to make it `undefined`) using the CLI. WDYT?
The scenario: bucket was added with the `--force_md5_etag true` by mistake, and one wishes to use the defaults like other buckets, how can it be done? (changing it to `false` back might not be aligned when the config changes).

shirady: I think it sounds good, this part 

> "maybe we should set it to be a string and allow only true/false and '' value to unset it"


will be only for this flag, right? because in other boolean flags, I'm not sure we have (or want) the option to unset them.

src/cmd/manage_nsfs.js
```diff
@@ -945,6 +951,10 @@ function validate_options_type_by_value(input_options_with_data) {
             if (['allow_bucket_creation', 'regenerate', 'wide', 'show_secrets', 'force'].includes(option) && validate_boolean_string_value(value)) {
                 continue;
             }
+            // special case for booleans that can be unset (allows '')
+            if (option === 'force_md5_etag' && validate_boolean_string_value_with_unset(value)) {
+                continue;
+            }
```
shirady, line 957: Why did you prefer adding a case for all the options (true, false, '') and not just the ''?
I'm suggestiong that in the previous condition add the `force_md5_etag` (the boolean true and false), and add a special case for the ''.
Having the two functions `validate_boolean_string_value` `validate_boolean_string_value_with_unset` might be confusing.
WDYT?

src/cmd/manage_nsfs.js
```diff
@@ -143,7 +143,8 @@ async function fetch_bucket_data(action, user_input) {
         path: user_input.path,
         should_create_underlying_storage: action === ACTIONS.ADD ? false : undefined,
         new_name: _.isUndefined(user_input.new_name) ? undefined : String(user_input.new_name),
-        fs_backend: _.isUndefined(user_input.fs_backend) ? config.NSFS_NC_STORAGE_BACKEND : String(user_input.fs_backend)
+        fs_backend: _.isUndefined(user_input.fs_backend) ? config.NSFS_NC_STORAGE_BACKEND : String(user_input.fs_backend),
+        force_md5_etag: _.isUndefined(user_input.force_md5_etag) ? undefined : get_boolean_or_string_value(user_input.force_md5_etag)
```
shirady, line 147: Maybe you can eliminate the unset ('') here instead of lines 173-174, for example:
```suggestion
        force_md5_etag: _.isUndefined(user_input.force_md5_etag) || user_input.force_md5_etag === ''? undefined : get_boolean_or_string_value(user_input.force_md5_etag)
```
* same comment in `fetch_account_data`.

src/cmd/manage_nsfs.js
```diff
@@ -958,9 +968,15 @@ function validate_options_type_by_value(input_options_with_data) {
...
+ * if has_unset_option is true allow '' value as an unset option
  * @param {boolean|string} value
+ * @param {boolean} has_unset_option
  */
-function validate_boolean_string_value(value) {
+function validate_boolean_string_value(value, has_unset_option) {
+    // unset value ''
+    if (has_unset_option && !value) {
+        return true;
+    }
```
shirady, line 979: I would write the condition '' explicitly (`value === ''`).

src/cmd/manage_nsfs.js
nadavMiz: I don't think it will work here since if we put an undefined value here it will be overridden by fetch_existing_bucket_data. also, I feel like lines 173-174 is a better place since it is where we unset fs_backend and s3_policy

src/cmd/manage_nsfs.js
nadavMiz: changed to have a different condition for unset items. in case it will not pass the condition it will be validated here as a regular boolean value

src/cmd/manage_nsfs.js
shirady: I can understand that you prefer to do the override in the same section where we unset fs_backend and s3_policy.
I find it confusing that in the unset of `force_md5_etag` we are using the `user_input.force_md5_etag` in condition and not `data.force_md5_etag`.

romayalon (approved):

Issue #7914, opened 2024-03-19
Symlinks under "s3-config/access_keys" should contain a relative path rather absolute path


anandhu-karattu:
### Environment info
- NooBaa Version: noobaa-core-5.15.0-20240313.el9.x86_64 (4.15 down stream)
- Platform: Standalone RHEL 9.2

### Actual behavior
We have a DR scenario with configuration backup/restore:

When an account is created, it creates a symlink under “s3-config/access_keys” directory.
```
[root@scale-2node-x-worker1 s3-config]# ls -lrt access_keys/
lrwxrwxrwx 1 root root 65 Mar 18 22:55 wwqNVMVquPBOfXIud5hK.symlink -> /gpfs/fs-ces-shared/new_path/ces/s3-config/accounts/account1.json
```
Use case:
Scale has a backup/restore feature which takes backup of configuration (everything under “s3-config” dir).
When user change the configuration path to a different FS, he can restore the old configuration.
However after restore, the “access_keys” directory still has the symlinks to old configuration path (absolute path). Thus access keys are no more working.
If I fix the symlinks to new absolute path, accounts become accessible again.
`[root@scale-2node-x-worker1 access_keys]# ln -sfn /gpfs/fs-ces-shared_new/ces/s3-config/accounts/account1.json wwqNVMVquPBOfXIud5hK.symlink`

### Expected behavior
Symlinks should contain a relative path rather absolute path

naveenpaul1:
@anandhu-karattu I dont think adding relative path will fix this specific scenario you mentioned here.
Suppose user already have a symlink 
`wwqNVMVquPBOfXIud5hK.symlink -> /gpfs/fs-ces-shared/new_path/ces/s3-config/accounts/account1.json` and we convert link to relative path to `../../../../../s3-config/accounts/account1.json` but after restoration if user copy the account config files to folder `/gpfs/fs-ces-shared/new_path/new_dir/ces/s3-config/accounts/` then the symlink will fail because new relative path should be `../../../../../s3-config/accounts/account1.json`, Correct me if I am wrong.
cc : @guymguym @romayalon

madhuthorat:
[...]

PR #7916, opened 2024-03-20
NSFS | NC | Symlinks should contain a relative path

naveenpaul1:
### Explain the changes
1. Symling target path is changed to relative path (`../accounts/account_name.json`)

1. https://github.com/noobaa/noobaa-core/issues/7914

### Testing Instructions:
1. create account and check the symlink should have relative path instead of absolute path
> root:access_keys# ls -lrt access_keys/
> total 8
> lrwxrwxrwx. 1 root root 29 Mar 20 12:36 kuDO2k8ahqvrju8clh2P.symlink -> ../accounts/acc_symlink1.json

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -49,6 +49,7 @@ function write_stdout_response(response_code, detail, event_arg) {
 const buckets_dir_name = '/buckets';
 const accounts_dir_name = '/accounts';
 const access_keys_dir_name = '/access_keys';
+const acounts_dir_relative_path = '../accounts/';
 
 let config_root;
 let accounts_dir_path;
@@ -440,6 +441,7 @@ async function add_account(data) {
     const fs_context = native_fs_utils.get_process_fs_context(config_root_backend);
     const access_key = data.access_keys[0].access_key;
     const account_config_path = get_config_file_path(accounts_dir_path, data.name);
+    const account_config_relative_path = get_config_file_path(acounts_dir_relative_path, data.name);
     const account_config_access_key_path = get_symlink_config_file_path(access_keys_dir_path, access_key);
 
     const name_exists = await native_fs_utils.is_path_exists(fs_context, account_config_path);
@@ -458,7 +460,7 @@ async function add_account(data) {
     nsfs_schema_utils.validate_account_schema(JSON.parse(data));
     await native_fs_utils.create_config_file(fs_context, accounts_dir_path, account_config_path, data);
     await native_fs_utils._create_path(access_keys_dir_path, fs_context, config.BASE_MODE_CONFIG_DIR);
-    await nb_native().fs.symlink(fs_context, account_config_path, account_config_access_key_path);
+    await nb_native().fs.symlink(fs_context, account_config_relative_path, account_config_access_key_path);
     write_stdout_response(ManageCLIResponse.AccountCreated, data, {account: event_arg});
 }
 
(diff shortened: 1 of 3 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -49,6 +49,7 @@ function write_stdout_response(response_code, detail, event_arg) {
 const buckets_dir_name = '/buckets';
 const accounts_dir_name = '/accounts';
 const access_keys_dir_name = '/access_keys';
+const acounts_dir_relative_path = '../accounts/';
```
shirady, line 52: What do you think about reusing the string from variable `accounts_dir_name`?

src/cmd/manage_nsfs.js
```diff
@@ -451,7 +453,7 @@ async function add_account(data) {
     nsfs_schema_utils.validate_account_schema(JSON.parse(data));
     await native_fs_utils.create_config_file(fs_context, accounts_dir_path, account_config_path, data);
     await native_fs_utils._create_path(access_keys_dir_path, fs_context, config.BASE_MODE_CONFIG_DIR);
-    await nb_native().fs.symlink(fs_context, account_config_path, account_config_access_key_path);
+    await nb_native().fs.symlink(fs_context, account_config_relative_path, account_config_access_key_path);
```
shirady, line 456: These changes - defining the variable `account_config_relative_path` and `symlink` to this path -  currently only appear in `add_account`, what about `update_account`?

src/cmd/manage_nsfs.js
naveenpaul1: `../accounts/` used in multiple places

src/cmd/manage_nsfs.js
naveenpaul1: Thanks, code updated

src/cmd/manage_nsfs.js
shirady: I think we need to move the definitions of the dir names to `manage_nsfs_constants` and the relative as well and use it in multiple places as needed:
https://github.com/noobaa/noobaa-core/blob/e5a28ef5cea44f0f2c9dccd5c2c3a26a762fac1e/src/cmd/manage_nsfs.js#L49-L51

We can do it in a different PR.

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -576,6 +592,24 @@ describe('manage nsfs cli account flow', () => {
             expect(new_account_details.allow_bucket_creation).toBe(true);
         });
 
+        it('cli account update account by name - check for the symlink relative path, not absolute path', async function() {
+            const { name } = defaults;
+            const new_name = 'account1_new_name';
+            const access_key = 'GIGiFAnjaaE7OKD5N123';
+            const secret_key = 'U3AYaMpU3zRDcRFWmvzgQr9MoHIAsD+3oEXAM123';
```
shirady, line 599: Are you updating the account name and also the `access_key` and `secret_key`?
If so you can edit the test name to reflect this change.

Regarding `secret_key` we add the suffix of `EXAMPLE` to avoid automatic alerts of secret leaking if you can please use it also.

src/cmd/manage_nsfs.js
guymguym: I totally agree with @shirady - keep the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).
I also don't think these const dir names should have `/` in the beginning, this is just confusing and appears like a root FS path. Instead, using path.join() will add dir separators as needed. But we can defer to next PR to avoid delaying this fix.

guymguym (approved):

src/cmd/manage_nsfs.js
naveenpaul1: @guymguym @shirady sure will add this changes in my next PR

PR #7923, opened 2024-03-22
Bucket_logging: Reset rsyslogd after logrotation

aspandey:
When logrotate rotates bucket_logs.log file after
it reaches a given size, rsyslogd could not send logs to bucket_logs.log file as it still has old fd opened for this file.

To fix this we need to send HUP signal after every rotation of logs.

### DIFF
--- a/src/deploy/NVA_build/logrotate_noobaa.conf
+++ b/src/deploy/NVA_build/logrotate_noobaa.conf
@@ -9,7 +9,7 @@
         create 660 noob root
         sharedscripts
         postrotate
-                supervisorctl signal HUP rsyslog 2> /dev/null || true
+                kill -HUP $(pidof rsyslogd) >/dev/null 2>&1 || true
         endscript
 }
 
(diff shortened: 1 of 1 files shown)


src/deploy/NVA_build/logrotate_noobaa.conf
```diff
@@ -40,4 +40,7 @@
         dateformat %Y-%m-%d-%H-%M-%S
         sharedscripts
         olddir /var/log/noobaa_logs
-}
\ No newline at end of file
+        postrotate
+                kill -HUP $(pidof rsyslogd) >/dev/null 2>&1 || true
```
liranmauda, line 44: why not do something like `/bin/systemctl restart rsyslog` ? 
```suggestion
                /bin/systemctl restart rsyslog
```

aspandey: While going through the man page of signal and rsyslogd, I understood that SIGHUP will not terminate the rsyslogd and will just re load its conf file. While in restart process there would be some time when rsyslogd will be in shutdown state. That plus the way we were doing it previously suggested to use HUP instead of restart.

While writing this, I just realized that I should also make this change for noobaa.log and client_noobaa.log files in this conf file. Will refresh it again.

liranmauda (approved):

PR #7925, opened 2024-03-25
Removed the dependency on request npm module

dannyzaken:
### Explain the changes
* removed some dead code that used `request`. I tried to limit the cleanup mainly to code that uses the request package. in some cases removed entire files that are not relevant anymore.
* for cases where I didn't want to remove the using code - implemented a simple http get in http_utils which should cover most of the cases we still need.
* agent.js - removed `public_ip` which was resolved using ipify API. This is not relevant anymore and also caused issues and questions from users
* removed agent_wrap.js - dead code
* removed read_server_config in cluster_server_api - dead code
* removed attempt_server_resolve in system_api - dead code
* removed prepare_aws_platform - dead code
* removed consolidate_test_reports - dead code
* removed test_env_builder_kubernetes - dead code
* os_utils - removed discover_virtual_appliance_address - dead code

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


nimrod-becker (approved):

liranmauda (approved):
LGTM, 
Just one small comment.

Approving.

src/test/system_tests/test_md_aggregator.js
```diff
@@ -273,10 +273,7 @@ function wait_for_server_to_start(max_seconds_to_wait, port) {
                 return isNotListening;
             },
             function() {
-                return P.ninvoke(request, 'get', {
-                        url: 'http://localhost:' + port,
-                        rejectUnauthorized: false,
-                    })
+                return http_utils.get('http://localhost:' + port)
```
jackyalbo, line 276: not http_get?
```suggestion
                return http_utils.http_get('http://localhost:' + port)
```
maybe we should also pass the options?

dannyzaken: url is passed as the first param, and `rejectUnauthorized` is irrelevant since it's HTTP and not HTTPS

dannyzaken: you are right regarding `http_get`. good catch :)

jackyalbo (approved):

PR #7942, opened 2024-04-02
[NC | NSFS] Fix GLACIER issues

tangledbytes:
### Explain the changes
This PR attempts to do the following to resolve issues reported by TC team:
1. Removes buffer pool logging
2. Throws invalid storage class error whenever any storage class other than `STANDARD` or `GLACIER` is used.
3. Displays correct storage class in the output of `list-parts` S3 API call instead of hardcoded `STANDARD`.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1575,6 +1575,11 @@ class NamespaceFS {
             const fs_context = this.prepare_fs_context(object_sdk);
             await this._load_multipart(params, fs_context);
             await this._check_path_in_bucket_boundaries(fs_context, params.mpu_path);
+            const { data } = await nb_native().fs.readFile(
+                fs_context,
+                path.join(params.mpu_path, 'create_object_upload')
+            );
+            const create_multipart_upload_params = JSON.parse(data.toString());
             const entries = await nb_native().fs.readdir(fs_context, params.mpu_path);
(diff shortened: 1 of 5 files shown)


src/sdk/namespace_fs.js
```diff
@@ -1575,6 +1575,14 @@ class NamespaceFS {
...
             await this._load_multipart(params, fs_context);
             await this._check_path_in_bucket_boundaries(fs_context, params.mpu_path);
+            let create_multipart_upload_params = {};
+            try {
+                const { data } = await nb_native().fs.readFile(fs_context, path.join(params.mpu_path, 'create_object_upload'));
+                create_multipart_upload_params = JSON.parse(data.toString());
+            } catch (error) {
+                // Let reading/parsing fail silently
+                dbg.warn('failed to read/parse create_object_upload for mpu_path:', params.mpu_path, 'error:', error);
+            }
```
guymguym, line 1585: I don't see why should we fail silently. I would keep it just like in complete_multipart_upload:
```suggestion
            const { data: create_params_buffer } = await nb_native().fs.readFile(
                fs_context,
                path.join(params.mpu_path, 'create_object_upload')
            );
```

guymguym (changes requested):

guymguym (approved):

PR #7943, opened 2024-04-02
[NC | NSFS] Add basic support for write thresholds

tangledbytes:
### Explain the changes
This PR attempts to add a basic support for disk space thresholds for write operations. It essentially means introducing a few configurable thresholds for NooBaa:
1. NSFS_LOW_FREE_SPACE_MB controls that how much mb in bytes does NooBaa consider to be too low to perform a write operation safely.
2. NSFS_LOW_FREE_SPACE_PERCENT controls how much space in terms of percentage does NooBaa consider to be too low to perform a write operation safely.
3.  NSFS_LOW_FREE_SPACE_MB_UNLEASH controls how muc space in mb does NooBaa consider to be enough to perform a write operation  safely.
4. NSFS_LOW_FREE_SPACE_PERCENT_UNLEASH controls how much much space in of percentage does NooBaa consider to be enough to perform write operations safely.

Difference between `LOW` and `OK` can be summarized as follows:
- space \in [0, low) => Deny
- space \in [low, ok)
    - prev_stable_space_state \in [0, low) => Deny
    - prev_stable_space_state \in [ok, \inifinity) => Approve
- space \in [ok, \infinity) => Approve

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -768,6 +768,34 @@ config.NSFS_GLACIER_EXPIRY_TZ = 'LOCAL';
 // Format must be HH:MM:SS
 config.NSFS_GLACIER_EXPIRY_TIME_OF_DAY = '00:00:00';
 
+config.NSFS_STATFS_CACHE_SIZE = config.NSFS_DIR_CACHE_MAX_TOTAL_SIZE;
+config.NSFS_STATFS_CACHE_EXPIRY_MS = 1 * 1000;
+
+// NSFS_LOW_FREE_SPACE_CHECK_ENABLED if set to true will use the below mentioned
+// thresholds to determine if the writes should be denied even
+// before we hit ENOSPC more filesystem.
+config.NSFS_LOW_FREE_SPACE_CHECK_ENABLED = false;
+
+// NSFS_LOW_FREE_SPACE_MB controls that how much space in
+// bytes does NooBaa consider to be too low to perform `PUT` operations
+// safely.
+config.NSFS_LOW_FREE_SPACE_MB = 8 * 1024;
+
+// NSFS_LOW_FREE_SPACE_PERCENT controls how much space in terms of
+// percentage does NooBaa consider to be too low to perform `PUT`
+// operations safely.
+config.NSFS_LOW_FREE_SPACE_PERCENT = 0.08;
+
+// NSFS_LOW_FREE_SPACE_MB_UNLEASH controls how much much space in bytes
+// does NooBaa consider to be enough to perform `PUT` operations
+// safely.
+config.NSFS_LOW_FREE_SPACE_MB_UNLEASH = 10 * 1024;
+
+// NSFS_LOW_FREE_SPACE_PERCENT_UNLEASH controls how much much space in of
+// percentage does NooBaa consider to be enough to perform `PUT`
+// operations safely.
+config.NSFS_LOW_FREE_SPACE_PERCENT_UNLEASH = 0.10;
+
 ////////////////////////////
 // NSFS NON CONTAINERIZED //
 ////////////////////////////
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -416,6 +416,28 @@ const versions_dir_cache = new LRUCache({
 });
 
 /**
+ * @typedef {{
+ *  statfs: Record<string, number>
+ * }} NsfsBucketStatFsCache
+ * @type {LRUCache<object, string, NsfsBucketStatFsCache>}
+ */
+const nsfs_bucket_statfs_cache = new LRUCache({
+    name: 'nsfs-bucket-statfs',
+    make_key: ({ bucket_path }) => bucket_path,
+    load: async ({ bucket_path, fs_context }) => {
+        const statfs = await nb_native().fs.statfs(fs_context, bucket_path);
+        return { statfs };
+    },
+    // validate - no need, validation will be as costly as `load`,
+    // instead let the item expire
+    expiry_ms: config.NSFS_STATFS_CACHE_EXPIRY_MS,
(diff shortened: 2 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -2990,6 +3004,65 @@ class NamespaceFS {
...
 
+    /**
+     * @param {nb.NativeFSContext} fs_context
+     * @param {number} [size_hint]
+     * @returns {Promise<void>}
+     */
+    async _throw_if_low_space(fs_context, size_hint) {
+        if (!config.NSFS_SPC_CHECK_ENABLED) return;
+
+        const statfs = await nb_native().fs.statfs(fs_context, this.bucket_path);
```
guymguym, line 3015: I think calling this before every operation is too much when we are serving hundreds+ operations per second.

Can we call it up to once every second (config value), and keep the last value in memory?

Perhaps changing nsfs_fs_space_cache to LRUCache would be the best way here.

src/sdk/namespace_fs.js
```diff
@@ -415,6 +415,16 @@ const versions_dir_cache = new LRUCache({
...
 
+/**
+ * @typedef {{
+ *  state: 'LOW' | 'OK' | undefined
+ * }} NsfsFsSpaceStateCache
+ * @type {NsfsFsSpaceStateCache}
+ */
+const nsfs_fs_space_cache = {
+    state: undefined,
+};
```
guymguym, line 426: this is global for all buckets but different buckets could be on different filesystems so we might need a map where the key is devid or fsid ...

guymguym (changes requested):

config.js
```diff
@@ -763,6 +763,34 @@ config.NSFS_GLACIER_EXPIRY_TZ = 'LOCAL';
...
+
+// NSFS_LOW_SPC_THRESHOLD_PERC controls how much space in terms of
+// percentage does NooBaa consider to be too low to perform `PUT`
+// operations safely.
+config.NSFS_LOW_SPC_THRESHOLD_PERC = 8;
+
+// NSFS_OK_SPC_THRESHOLD_BYT controls how much much space in bytes
+// does NooBaa consider to be enough to perform `PUT` operations
+// safely.
+config.NSFS_OK_SPC_THRESHOLD_BYT = 10 * 1024 * 1024 * 1024;
```
guymguym, line 787:
```suggestion
config.NSFS_LOW_FREE_SPACE_BYTES_UNLEASH = 10 * 1024 * 1024 * 1024;
```

config.js
```diff
@@ -763,6 +763,34 @@ config.NSFS_GLACIER_EXPIRY_TZ = 'LOCAL';
...
+
+// NSFS_OK_SPC_THRESHOLD_PERC controls how much much space in of
+// percentage does NooBaa consider to be enough to perform `PUT`
+// operations safely.
+config.NSFS_OK_SPC_THRESHOLD_PERC = 10;
```
guymguym, line 792:
```suggestion
config. NSFS_LOW_FREE_SPACE_PERCENT_UNLEASH = 0.10;
```

src/sdk/namespace_fs.js
```diff
@@ -2990,6 +3020,65 @@ class NamespaceFS {
...
+     * @param {nb.NativeFSContext} fs_context
+     * @param {number} [size_hint]
+     * @returns {Promise<void>}
+     */
+    async _throw_if_low_space(fs_context, size_hint) {
+        if (!config.NSFS_SPC_CHECK_ENABLED) return;
+
+        const { statfs } = await nsfs_bucket_statfs_cache.get_with_cache({ bucket_path: this.bucket_path, fs_context });
+        const free_space = (statfs.bfree * statfs.bsize) - (size_hint || 0);
+        const total_space = statfs.blocks * statfs.bsize;
```
guymguym, line 3033: I worry that these numbers might not fit into js number for large systems.
js number has up to 53 bits for integer values which can fit up to 8 PB (Number.MAX_SAFE_INTEGER), but practically doing calculations with such large numbers is risky and can produce unexpected results.
We could either do all the calculations with BigInt or just convert all units to MB.

src/sdk/namespace_fs.js
```diff
@@ -2996,6 +3022,59 @@ class NamespaceFS {
...
+     * @param {number} in_bytes free space threshold in mb
+     * @param {number} in_percentage free space threshold in percentage
+     * @param {number} total_space total space in mb
+     * @returns {number}
+     */
+    _get_free_space_threshold(in_bytes, in_percentage, total_space) {
+        const free_from_percentage = in_percentage * total_space;
+
+        if (in_bytes > free_from_percentage) return in_bytes;
+        return free_from_percentage;
```
guymguym, line 3075: return Math.max(...)

config.js
```diff
@@ -768,6 +768,34 @@ config.NSFS_GLACIER_EXPIRY_TZ = 'LOCAL';
 // Format must be HH:MM:SS
 config.NSFS_GLACIER_EXPIRY_TIME_OF_DAY = '00:00:00';
 
+// NSFS_LOW_FREE_SPACE_CHECK_ENABLED if set to true will use the below mentioned
+// thresholds to determine if the writes should be denied even
+// before we hit ENOSPC more filesystem.
+config.NSFS_LOW_FREE_SPACE_CHECK_ENABLED = false;
+
+config.NSFS_STATFS_CACHE_SIZE = config.NSFS_DIR_CACHE_MAX_TOTAL_SIZE;
+config.NSFS_STATFS_CACHE_EXPIRY_MS = 1 * 1000;
```
guymguym, line 777: prefer to keep the NSFS_LOW_* configs next to eachother and not break them with this

guymguym (approved):
few more comments, but overall this looks good!

PR #7950, opened 2024-04-04
bucket_logging: Store log objects inside log prefix directory

aspandey:
We provide log prefix while setting bucket logging for a noobaa bucket. When log objects are uploaded on log bucket, it should be saved under a directory named as log prefix.


liranmauda (approved):
LGTM

PR #7952, opened 2024-04-04
[NC | NSFS] use fixed `hh:mm` for running expiry script

tangledbytes:
### Explain the changes
This PR moves away from running expiry script on certain interval and instead use a configurable `hh:mm` or when the disk is running low on space.

The granularity is minutes instead of seconds to account for scheduler mess ups.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -758,9 +758,19 @@ config.NSFS_GLACIER_MIGRATE_INTERVAL = 15 * 60 * 1000;
 // of `manage_nsfs glacier restore`
 config.NSFS_GLACIER_RESTORE_INTERVAL = 15 * 60 * 1000;
 
-// NSFS_GLACIER_EXPIRY_INTERVAL indicates the interval between runs
-// of `manage_nsfs glacier expiry`
-config.NSFS_GLACIER_EXPIRY_INTERVAL = 12 * 60 * 60 * 1000;
+// NSFS_GLACIER_EXPIRY_RUN_TIME must be of the format hh:mm which specifies
+// when NooBaa should allow running glacier expiry process
+// NOTE: This will also be in the same timezone as specified in
+// NSFS_GLACIER_EXPIRY_TZ
+config.NSFS_GLACIER_EXPIRY_RUN_TIME = '03:00';
+
+// NSFS_GLACIER_EXPIRY_RUN_TIME_TOLERANCE_MINS configures the delay
+// tolerance in minutes.
+//
+// eg. If the expiry run time is set to 03:00 and the tolerance is
+// set to be 2 mins then the expiry can trigger till 03:02 (unless
+// already triggered between 03:00 - 03:02
+config.NSFS_GLACIER_EXPIRY_RUN_DELAY_LIMIT_MINS = 2 * 60;
 
 /** @type {'UTC' | 'LOCAL'} */
 config.NSFS_GLACIER_EXPIRY_TZ = 'LOCAL';
--- a/src/manage_nsfs/manage_nsfs_glacier.js
+++ b/src/manage_nsfs/manage_nsfs_glacier.js
@@ -69,12 +69,65 @@ async function process_expiry() {
     const fs_context = native_fs_utils.get_process_fs_context();
 
     await lock_and_run(fs_context, SCAN_LOCK, async () => {
-        if (!(await time_exceeded(fs_context, config.NSFS_GLACIER_EXPIRY_INTERVAL, GlacierBackend.EXPIRY_TIMESTAMP_FILE))) return;
-
-
-        await getGlacierBackend().expiry(fs_context);
-        await record_current_time(fs_context, GlacierBackend.EXPIRY_TIMESTAMP_FILE);
+        const backend = getGlacierBackend();
+        if (
+            await backend.low_free_space() ||
+            await is_desired_time(
+                    fs_context,
+                    new Date(),
+                    config.NSFS_GLACIER_EXPIRY_RUN_TIME,
+                    config.NSFS_GLACIER_EXPIRY_RUN_DELAY_LIMIT_MINS,
+                    GlacierBackend.EXPIRY_TIMESTAMP_FILE,
+            )
+        ) {
+            await backend.expiry(fs_context);
+            await record_current_time(fs_context, GlacierBackend.EXPIRY_TIMESTAMP_FILE);
+        }
     });
+}
+
+/**
+ * is_desired_time returns true if the given time matches with
+ * the desired time or if 
+ * @param {nb.NativeFSContext} fs_context 
+ * @param {Date} current
+ * @param {string} desire time in format 'hh:mm'
+ * @param {number} delay_limit_mins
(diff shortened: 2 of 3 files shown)


src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -69,12 +69,38 @@ async function process_expiry() {
...
+        (desired_hour < 0 || desired_hour >= 24) ||
+        (desired_min < 0 || desired_min >= 60)
+    ) {
+        throw new Error('invalid `NSFS_GLACIER_EXPIRY_RUN_TIME` - must be hh:mm');
+    }
 
-        await getGlacierBackend().expiry(fs_context);
-        await record_current_time(fs_context, GlacierBackend.EXPIRY_TIMESTAMP_FILE);
-    });
+    return desired_hour === current_hour && desired_min === current_min;
```
guymguym, line 103: So in case we are not scheduled to run during the desired minute, we miss the daily opportunity completely?
Perhaps we should be a bit more tolerant and have a maximum interval after that desired time where we would still run the expiry? In order to do that we still need to maintain the timestamp file to prevent running it multiple times in that window.

config.js
```diff
@@ -758,9 +758,9 @@ config.NSFS_GLACIER_MIGRATE_INTERVAL = 15 * 60 * 1000;
 // of `manage_nsfs glacier restore`
 config.NSFS_GLACIER_RESTORE_INTERVAL = 15 * 60 * 1000;
 
-// NSFS_GLACIER_EXPIRY_INTERVAL indicates the interval between runs
-// of `manage_nsfs glacier expiry`
-config.NSFS_GLACIER_EXPIRY_INTERVAL = 12 * 60 * 60 * 1000;
+// NSFS_GLACIER_EXPIRY_RUN_TIME must be of the format hh:mm which specifies
+// when NooBaa should allow running glacier expiry process
+config.NSFS_GLACIER_EXPIRY_RUN_TIME = '03:00';
```
guymguym, line 763: worth noting here that this is also in the timezone option as set by NSFS_GLACIER_EXPIRY_TZ and NSFS_GLACIER_EXPIRY_TIME_OF_DAY

config.js
```diff
@@ -758,9 +758,19 @@ config.NSFS_GLACIER_MIGRATE_INTERVAL = 15 * 60 * 1000;
...
+
+// NSFS_GLACIER_EXPIRY_RUN_TIME_TOLERANCE_MINS configures the delay
+// tolerance in minutes.
+//
+// eg. If the expiry run time is set to 03:00 and the tolerance is
+// set to be 2 mins then the expiry can trigger till 03:02 (unless
+// already triggered between 03:00 - 03:02
+// NOTE: This will also be in the same timezone as specified in
+// NSFS_GLACIER_EXPIRY_TZ
```
guymguym, line 772: this comment should be moved to NSFS_GLACIER_EXPIRY_RUN_TIME

config.js
```diff
@@ -758,9 +758,19 @@ config.NSFS_GLACIER_MIGRATE_INTERVAL = 15 * 60 * 1000;
...
+config.NSFS_GLACIER_EXPIRY_RUN_TIME_TOLERANCE_MINS = 2;
```
guymguym, line 773: Can this be set to 120 minutes? I think the default of 2 mins is anyway too low. I would expect this to be in the scale of 1-2 hours.

guymguym: Also I would prefer to change the term "tolerance" to "delay_limit" or "max_delay"

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -69,12 +69,70 @@ async function process_expiry() {
...
+    ) {
+        throw new Error('invalid desired_time - must be hh:mm');
+    }
 
-        await getGlacierBackend().expiry(fs_context);
-        await record_current_time(fs_context, GlacierBackend.EXPIRY_TIMESTAMP_FILE);
-    });
+    const min_time = new Date();
+    min_time.setHours(desired_hour);
+    min_time.setMinutes(desired_min);
```
guymguym, line 112: don't you need to check config.NSFS_GLACIER_EXPIRY_TZ and change to use setUTCHours etc?

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -69,12 +69,80 @@ async function process_expiry() {
...
+ * is_desired_time returns true if the given time matches with
+ * the desired time or if 
+ * @param {nb.NativeFSContext} fs_context 
+ * @param {Date} current
+ * @param {string} desire time in format 'hh:mm'
+ * @param {number} tolerate_mins
+ * @param {string} timestamp_file 
+ * @returns {Promise<boolean>}
+ */
+async function is_desired_time(fs_context, current, desire, tolerate_mins, timestamp_file) {
```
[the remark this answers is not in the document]
tangledbytes: Done

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -69,12 +69,80 @@ async function process_expiry() {
...
+    }
+
+    const max_time = new Date();
+    if (config.NSFS_GLACIER_EXPIRY_TZ === 'UTC') {
+        max_time.setUTCHours(desired_hour);
+        max_time.setUTCMinutes(desired_min + tolerate_mins);
+    } else {
+        max_time.setHours(desired_hour);
+        max_time.setMinutes(desired_min + tolerate_mins);
+    }
```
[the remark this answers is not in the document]
tangledbytes: I have set the secs to 0 now but not millis. Considering the granularity is minutes, does it matter?

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -69,12 +69,80 @@ async function process_expiry() {
...
+
+    if (current >= min_time && current <= max_time) {
+        try {
+            const { data } = await nb_native().fs.readFile(fs_context, path.join(config.NSFS_GLACIER_LOGS_DIR, timestamp_file));
+            const lastrun = new Date(data.toString());
+
+            // Last run should NOT be in this window
+            if (lastrun >= min_time && lastrun <= max_time) return false;
+        } catch (error) {
+            console.error('failed to read last run timestamp:', error, 'timestamp_file:', timestamp_file);
```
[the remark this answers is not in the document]
tangledbytes: Done

src/manage_nsfs/manage_nsfs_glacier.js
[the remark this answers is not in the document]
guymguym: I would prefer it to be clear and not have random millis.

guymguym (approved):
looks good, but i would still zero the millis.

src/manage_nsfs/manage_nsfs_glacier.js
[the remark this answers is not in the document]
tangledbytes: Done

PR #7954, opened 2024-04-07
Fix Bug BZ 2272900 | NSFS | Remove Issues Updates with System Store

shirady:
### Explain the changes
1. Remove issues updates with the system store.
2. Avoid updating the issues report in case of no such key when using a head object in NSFS.
3. Move the update last monitoring to the memory instead of the system store, since it is also used by the issues report.

### Issues: Fixed [BZ 2272900](https://bugzilla.redhat.com/show_bug.cgi?id=2272900)
1. Currently we update the issues report with the system store, and if we encounter 10/10 issues it's changing the namespace mode to `IO_ERRORS`. In this fix, we suggest saving this information in memory in the pool server and not updating the issue report when we don't find the key metadata (head object).
2. GAP - avoid updating the issues report was done only in NSFS.

### Testing Instructions:
General:
1. Deploy noobaa on a local cluster (MInikube or Rancher Desktop) (see [guide](https://github.com/noobaa/noobaa-operator/blob/master/doc/dev_guide/deply_noobaa_on_minikube_or_rancher_desktop.md)).
3. Deploy NSFS on a local cluster (Based on the instructions [here](https://github.com/noobaa/noobaa-core/wiki/NSFS-on-Kubernetes/8fa95c3b827c40f2435a63f60fae5f7285e6b221)).

NSFS Change:
1. Use AWS CLI and use the head-object command on a noobaa bucket in NSFS on non-existing key.
You can use a loop for this, for example: `for i in {1..10}; do s3-nb-user-1 s3api head-object --bucket fs1-jenia-bucket --key non_exist.txt; done` (`s3-nb-user-1` is an alias `alias s3-nb-user-1='AWS_ACCESS_KEY=<access_key_nsfs_account> AWS_SECRET_ACCESS_KEY=<secret_access_key_nsfs_account> aws --no-verify-ssl --endpoint-url https://localhost:12443'`.
2. Before: the namespace phase was changed from `Ready` to `Rejected` with mode `IO_ERRORS` after 10 failure.
After: the namespace phase is still `Ready`.

System store change:
1. Set higher debug level (at least 3): `noobaa system set-debug-level 5`.
2. Search for the logs `update_last_monitoring: update_last_monitoring` in the core logs (`kubectl logs pod noobaa-core-0 -f`).

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -537,6 +537,24 @@ class NamespaceFS {
         }
     }
 
+    /**
+     * _should_update_issues_report is intended to avoid updating the namespace issues report in case:
+     * 1. The key doesn't exist and the path is not internal -
+     *    internal path is created for specific cases, for example in version.
+     *    Note: it also covers the delete marker case (since it is in a versioned path)
+     * IMPORTANT: This function is correct only for read_object_md!
+     * @param {object} params
+     * @param {string} file_path
+     * @param {object} err
+     */
+    _should_update_issues_report(params, file_path, err) {
+        const { key } = params;
+        const md_file_path = this._get_file_md_path({ key });
+        const non_internal_path = file_path === md_file_path;
+        const no_such_key_condition = err.code === `ENOENT` && non_internal_path;
+        return !no_such_key_condition;
+    }
+
     is_readonly_namespace() {
         return this.access_mode === 'READ_ONLY';
     }
@@ -872,8 +890,9 @@ class NamespaceFS {
 
     async read_object_md(params, object_sdk) {
         const fs_context = this.prepare_fs_context(object_sdk);
+        let file_path;
         try {
-            const file_path = await this._find_version_path(fs_context, params, true);
+            file_path = await this._find_version_path(fs_context, params, true);
             await this._check_path_in_bucket_boundaries(fs_context, file_path);
             await this._load_bucket(params, fs_context);
             let stat = await nb_native().fs.stat(fs_context, file_path);
@@ -894,7 +913,9 @@ class NamespaceFS {
             this._throw_if_delete_marker(stat);
             return this._get_object_info(params.bucket, params.key, stat, params.version_id || 'null', isDir);
         } catch (err) {
-            this.run_update_issues_report(object_sdk, err);
+            if (this._should_update_issues_report(params, file_path, err)) {
+                this.run_update_issues_report(object_sdk, err);
+            }
             throw this._translate_object_error_codes(err);
         }
     }
--- a/src/server/system_services/pool_server.js
+++ b/src/server/system_services/pool_server.js
@@ -48,6 +48,11 @@ const POOL_HOSTS_INFO_DEFAULTS = Object.freeze({
     by_service: {},
 });
 
+
+// key: namespace_resource_id, value: { last_monitoring: date, issues: array of issues } 
+// (see namespace_resource_schema)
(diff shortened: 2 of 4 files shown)


src/sdk/namespace_fs.js
```diff
@@ -537,6 +537,25 @@ class NamespaceFS {
...
+     */
+    _should_update_issues_report(params, file_path, err) {
+        let res = true;
+        const { key } = params;
+        const md_file_path = this._get_file_md_path({ key });
+        const non_internal_path = file_path === md_file_path;
+        const no_such_key_condition = err.code === `ENOENT` && non_internal_path;
+        if (no_such_key_condition) res = false;
+        return res;
+    }
```
liranmauda, line 557:
```suggestion
    _should_update_issues_report(params, file_path, err) {
        const { key } = params;
        const md_file_path = this._get_file_md_path({ key });
        const non_internal_path = file_path === md_file_path;
        const no_such_key_condition = err.code === `ENOENT` && non_internal_path;
        return !no_such_key_condition;
    }
```

liranmauda (commented):
I am missing it probably, but how do we set the report in the DB? 
It seems that the make_changes was removed.

shirady:
> I am missing it probably, but how do we set the report in the DB? It seems that the make_changes was removed.

@liranmauda the issues report will be saved in memory (using a global variable) in the pool server.
The motivation was that the update of the system store is "heavy" and should be for changes in the system like configuration. The disadvantage of this change is that if someone deletes the core pod this information will be deleted.

src/sdk/namespace_fs.js
shirady: I can change it, the reason I tried to write it this way was to avoid negative conditions and easily expand this for other cases.

src/server/system_services/pool_server.js
```diff
@@ -48,6 +48,12 @@ const POOL_HOSTS_INFO_DEFAULTS = Object.freeze({
     by_service: {},
 });
 
+
+// key: namespace_resource_id, value: array of issues (see namespace_resource_schema)
+const map_issues_report = new Map();
+// key: namespace_resource_id, value: last_monitoring (see namespace_resource_schema)
```
romayalon, line 54: do we really need this map?
I think that having a single map that captures all the namespace_resource_monitoring info is less complex - 
// key: namespace_resource_id, value:  { last_monitoring : date, issues: array of issues }

src/sdk/namespace_fs.js
```diff
@@ -537,6 +537,23 @@ class NamespaceFS {
...
+     *    internal path is created for specific cases, for example in version.
+     *    Note: it also covers the delete marker case (since it is in a versioned path)
+     * @param {object} params
+     * @param {string} file_path
+     * @param {object} err
+     */
+    _should_update_issues_report(params, file_path, err) {
+        const { key } = params;
+        const md_file_path = this._get_file_md_path({ key });
+        const non_internal_path = file_path === md_file_path;
```
romayalon, line 552: please add a note that this function is correct only for read_object_md, if someone wants to use it on read_object_stream it has a gap regarding directory_content objects- when we get ENOENT error on the content path of the object...

src/server/system_services/pool_server.js
shirady: I removed it, hoping it is less complex now, thanks.

romayalon (approved):

Issue #7941, opened 2024-04-02
CopyObject option from copying object from bucket to bucket with url escape not working as expected


ramya-c3:
### Environment info
- NooBaa Version:  noobaa-core-5.15.0-20240306.el8.x86_64
- Platform: standalone

### Actual behavior
1.CopyObject             : Bucket does not exists                                          : url.QueryEscape() 

If we pass the query with url.queryescape it is not finding the bucket it is throwing bucket doesnot exist

### Expected behavior
1. It should copy the object from bucket  to bucket

### Steps to reproduce
1.
input := &s3.CopyObjectInput{
--
  | /* Source includes both source bucket and source object name */
  | CopySource: aws.String(url.QueryEscape(Config.bucketName+ "/" + sourceKey)),
  | /* Below are the bucket and new object name */
  | Bucket: aws.String(Config.bucketName),
  | Key:    aws.String(destKey),
  | }

shirady:
Hi @ramya-c3,
Would you please add more details to the issue?

For example:
- It looks like you're using an AWS SDK, which one? would you share the link to the SDK doc you're using?
- Did you perform the same operation on AWS buckets and it works and now you tried the same with NSFS buckets?
- If you can attach what you tried to run, what is the output? maybe share the logs as well.
- Did you try to perform the same action in AWS CLI? Did it work as expected?

shirady:
Hi,
Adding more information regarding the issue:
- The code attached is a test file using Go SDK (see below).
- The issue is in general about a client that uses `url.QueryEscape` in `Go`, the example is when it was tested by using `CopyObject`, but it is probably relevant to more actions.
- 

Attachments:
Note: some of the values were removed, you can see the `<>` sign.

- Successful response (not using `url.QueryEscape` in `Go`) - they add that If they use without `url.QueryEscape` it works but they want to support with `url.QueryEscape` for object name with special characters.
```
c80f3m5n01 240403-08:10:28 5] go run ~/test.go 
0xc00039eea02024/04/03 08:10:34 DEBUG: Request s3/CopyObject Details:
---[ REQUEST POST-SIGN ]-----------------------------
PUT /noobaa2/AFM_FILE_RENAME HTTP/1.1
Host: <>
User-Agent: aws-sdk-go/1.51.8 (go1.20.10; linux; amd64)
Content-Length: 0
Authorization: <>,
SignedHeaders=host;x-amz-content-sha256;x-amz-copy-source;x-amz-date, 
[...]

PR #7965, opened 2024-04-10
fix url decoding for copyObject

nadavMiz:
### Explain the changes
1. when url part of bucket is encoded, the '/' is encoded as well. change the order so we un-escaped the URL before we search for '/'. still search for ? in encoded state, since it isn't encoded as part of the bucket/key

1. https://github.com/noobaa/noobaa-core/issues/7941

### Testing Instructions:
1. call copy-object-input with escaped parameter combining bucket name and key (<bucket>/<key>). 
3. command should work 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/endpoint_utils.js
+++ b/src/endpoint/endpoint_utils.js
@@ -17,21 +17,24 @@ function prepare_rest_request(req) {
 }
 
 function parse_source_url(source_url) {
-    let slash_index = source_url.indexOf('/');
-    let start_index = 0;
-    if (slash_index === 0) {
-        start_index = 1;
-        slash_index = source_url.indexOf('/', 1);
-    }
-    let query_index = source_url.indexOf('?', slash_index);
+    let query_index = source_url.indexOf('?');
     let query;
     if (query_index < 0) {
         query_index = source_url.length;
     } else {
         query = querystring.parse(source_url.slice(query_index + 1));
     }
-    const bucket = decodeURIComponent(source_url.slice(start_index, slash_index));
-    const key = decodeURIComponent(source_url.slice(slash_index + 1, query_index));
+
+    // '/' may be encoded, first decde the bucket/key part
+    const decoded_source_url = decodeURIComponent(source_url.slice(0, query_index));
+    let slash_index = decoded_source_url.indexOf('/');
+    let start_index = 0;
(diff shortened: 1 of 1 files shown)


src/endpoint/endpoint_utils.js
```diff
@@ -17,21 +17,24 @@ function prepare_rest_request(req) {
...
 
 function parse_source_url(source_url) {
-    let slash_index = source_url.indexOf('/');
-    let start_index = 0;
-    if (slash_index === 0) {
-        start_index = 1;
-        slash_index = source_url.indexOf('/', 1);
-    }
-    let query_index = source_url.indexOf('?', slash_index);
+    let query_index = source_url.indexOf('?');
```
romayalon, line 20: so we don't need to start the searching after the slash_index?

src/endpoint/endpoint_utils.js
```diff
@@ -17,21 +17,24 @@ function prepare_rest_request(req) {
...
+    const decoded_source_url = decodeURIComponent(source_url.slice(0, query_index));
+    let slash_index = decoded_source_url.indexOf('/');
+    let start_index = 0;
+    if (slash_index === 0) {
+        start_index = 1;
+        slash_index = decoded_source_url.indexOf('/', 1);
+    }
+    const bucket = decoded_source_url.slice(start_index, slash_index);
+    const key = decoded_source_url.slice(slash_index + 1);
     return { query, bucket, key };
```
romayalon, line 38: can you please add tests?

src/endpoint/endpoint_utils.js
```diff
@@ -17,21 +17,24 @@ function prepare_rest_request(req) {
...
     let query;
     if (query_index < 0) {
         query_index = source_url.length;
     } else {
         query = querystring.parse(source_url.slice(query_index + 1));
     }
-    const bucket = decodeURIComponent(source_url.slice(start_index, slash_index));
-    const key = decodeURIComponent(source_url.slice(slash_index + 1, query_index));
+
+    // '/' may be encoded, first decde the bucket/key part
```
romayalon, line 28: this comment is not so understood, would you mind rephrasing it, please?

shirady:
Closing this PR because its commit was cherry-picked to PR #7984

PR #7972, opened 2024-04-15
Support to update credentials manually to an account

aayushchouhan09:
### Explain the changes
1. Added an API `update_account_keys` to update the account credentials (access key and secret key) manually. It requires `email` and `access_keys` as the input parameters.

### DIFF
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -299,6 +299,49 @@ async function generate_account_keys(req) {
     });
 }
 
+/**
+ *
+ * UPDATE_ACCOUNT_KEYS
+ *
+ */
+async function update_account_keys(req) {
+    const account = system_store.get_account_by_email(req.rpc_params.email);
+    if (!account) {
+        throw new RpcError('NO_SUCH_ACCOUNT', 'No such account email: ' + req.rpc_params.email);
+    }
+    if (req.system && req.account) {
+        if (!is_support_or_admin_or_me(req.system, req.account, account)) {
+            throw new RpcError('UNAUTHORIZED', 'Cannot update account');
+        }
+    }
+    if (account.is_support) {
+        throw new RpcError('FORBIDDEN', 'Cannot update support account');
+    }
+    const access_keys = req.rpc_params.access_keys;
+
+    access_keys.secret_key = system_store.master_key_manager.encrypt_sensitive_string_with_master_key_id(
+        access_keys.secret_key, account.master_key_id._id);
+
+    await system_store.make_changes({
+        update: {
+            accounts: [{
+                _id: account._id,
+                access_keys: [
+                    access_keys
+                ]
+            }]
+        }
+    });
+
+    Dispatcher.instance().activity({
+        event: 'account.update_credentials',
+        level: 'info',
+        system: req.system && req.system._id,
+        actor: req.account && req.account._id,
+        account: account._id,
+        desc: `Credentials for ${account.email.unwrap()} were updated ${req.account && 'by ' + req.account.email.unwrap()}`,
+    });
+}
 
 /**
  *
@@ -1402,8 +1445,6 @@ function validate_create_account_params(req) {
     }
 }
 
-
-
 async function verify_authorized_account(req) {
     //operator connects by token and doesn't have the password property.
     if (req.role === 'operator') {
@@ -1472,6 +1513,7 @@ exports.reset_password = reset_password;
 exports.delete_account = delete_account;
 exports.delete_account_by_property = delete_account_by_property;
 exports.generate_account_keys = generate_account_keys;
+exports.update_account_keys = update_account_keys;
 exports.update_account_s3_access = update_account_s3_access;
 exports.list_accounts = list_accounts;
 exports.accounts_status = accounts_status;
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -251,6 +251,22 @@ module.exports = {
             }
         },
 
+        update_account_keys: {
+            doc: 'Update account keys of the authorized account',
+            method: 'PUT',
+            params: {
+                type: 'object',
+                required: ['email', 'access_keys'],
+                properties: {
+                    email: { $ref: 'common_api#/definitions/email' },
+                    access_keys: { $ref: 'common_api#/definitions/access_keys' },
+                },
+            },
+            auth: {
+                system: 'admin'
+            }
+        },
+
         verify_authorized_account: {
             doc: 'Verify authorized account password',
             method: 'PUT',


src/server/system_services/account_server.js
```diff
@@ -299,6 +299,48 @@ async function generate_account_keys(req) {
...
+ *
+ */
+async function update_account_keys(req) {
+    const account = system_store.get_account_by_email(req.rpc_params.email);
+    if (!account) {
+        throw new RpcError('NO_SUCH_ACCOUNT', 'No such account email: ' + req.rpc_params.email);
+    }
+    if (req.system && req.account) {
+        if (!is_support_or_admin_or_me(req.system, req.account, account)) {
+            throw new RpcError('UNAUTHORIZED', 'Cannot update account');
```
shirady, line 314: Maybe give more details in the message - why "Cannot update account"?

src/server/system_services/account_server.js
```diff
@@ -299,6 +299,48 @@ async function generate_account_keys(req) {
...
+ * UPDATE_ACCOUNT_KEYS
+ *
+ */
+async function update_account_keys(req) {
+    const account = system_store.get_account_by_email(req.rpc_params.email);
+    if (!account) {
+        throw new RpcError('NO_SUCH_ACCOUNT', 'No such account email: ' + req.rpc_params.email);
+    }
+    if (req.system && req.account) {
+        if (!is_support_or_admin_or_me(req.system, req.account, account)) {
```
shirady, line 313: The "req.system && req.account" condition is only to verify that these values are defined?
So you can continue the same condition with `&& !is_support_or_admin_or_me(req.system, req.account, account)`, for less nested condition.

shirady:
@achouhan09 what about the complexity of the credentials?
Can I send access_key "123" and secret_key "123"?

src/server/system_services/account_server.js
achouhan09: The logic for this msg here is that the requesting account is none of the authorized user(support, admin or the user itself). So we are simply saying that `UNAUTHORIZED`. I don't think we need to put any additional details here.

src/server/system_services/account_server.js
achouhan09: Yes agree but I think it will make it more complex to read as there is another else if for `!req.system`

src/server/system_services/account_server.js
```diff
@@ -299,6 +300,52 @@ async function generate_account_keys(req) {
...
+    access_keys.secret_key = system_store.master_key_manager.encrypt_sensitive_string_with_master_key_id(
+        access_keys.secret_key, account.master_key_id._id);
+
+    await system_store.make_changes({
+        update: {
+            accounts: [{
+                _id: account._id,
+                access_keys: [
+                    access_keys
+                ]
```
shirady, line 335: After running this the whole array `access_keys` is replaced, right?
Do you think there is a case where you might want to save the old credentials somehow?

src/server/system_services/account_server.js
```diff
@@ -299,6 +300,52 @@ async function generate_account_keys(req) {
...
+        }
+    });
+
+    Dispatcher.instance().activity({
+        event: 'account.update_credentials',
+        level: 'info',
+        system: req.system && req.system._id,
+        actor: req.account && req.account._id,
+        account: account._id,
+        desc: `Credentials for ${account.email.unwrap()} were updated ${req.account && 'by ' + req.account.email.unwrap()}`,
```
shirady, line 346: In which cases the update of the credentials happens from one account to a different account?

src/server/system_services/account_server.js
```diff
@@ -299,6 +300,52 @@ async function generate_account_keys(req) {
...
+                access_keys: [
+                    access_keys
+                ]
+            }]
+        }
+    });
+
+    Dispatcher.instance().activity({
+        event: 'account.update_credentials',
+        level: 'info',
```
shirady, line 342: Why did you choose `info`? Did you consider `alert`?
I looked at this example:
https://github.com/noobaa/noobaa-core/blob/ed421ac19606f090fc4bf1c65e07bdb1d085b003/src/server/system_services/account_server.js#L613-L619

src/server/system_services/account_server.js
```diff
@@ -1402,7 +1449,27 @@ function validate_create_account_params(req) {
...
+    const access_key = access_keys.access_key;
+    const secret_key = access_keys.secret_key;
+    // checking the complexity of access_key
+    if (!_.isUndefined(access_key) && !string_utils.validate_complexity(access_key, {
+        require_length: 20,
+        check_uppercase: true,
+        check_lowercase: false,
+        check_numbers: true,
+        check_symbols: false,
+    })) throw new RpcError('INVALID_CREDENTIALS', 'Access key must meet complexity requirements');
```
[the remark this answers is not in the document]
achouhan09: Make sense. Sure, will add something similar. Thanks

src/server/system_services/account_server.js
achouhan09: Admin and User can only update the credentials. So its either `user by admin `or `user by user`.

src/server/system_services/account_server.js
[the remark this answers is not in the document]
achouhan09: I think we can put `ACCESS_KEY_ERROR` and `SECRET_KEY_ERROR` as error_code and in error message we could explain in details about complexity.
```suggestion
    })) throw new RpcError('ACCESS_KEY_ERROR', 'Account access key length must be 20, and must contain uppercase and numbers');
```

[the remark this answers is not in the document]
shirady: I also think it would be more clear.

src/server/system_services/account_server.js
achouhan09: As per my understanding I think `alert` would be in the case of error and `info` for activity logs.
Pls check this update function: [account_server.js](https://github.com/noobaa/noobaa-core/blob/ed421ac19606f090fc4bf1c65e07bdb1d085b003/src/server/system_services/account_server.js#L484-L501)

src/server/system_services/account_server.js
achouhan09: I was looking at the wrong code [here](https://github.com/noobaa/noobaa-core/blob/ed421ac19606f090fc4bf1c65e07bdb1d085b003/src/server/system_services/account_server.js#L319-L321). I think for now we can keep it as same because everywhere it was like that.

liranmauda:
> > @achouhan09 what about the complexity of the credentials? Can I send access_key "123" and secret_key "123"?
> 
> @shirady Yes needs to be added, I will put the validation for complexity. Thanks!

I am not convinced that we need to protect it.
it is an internal API.
We do not protect it on `create` nor `update`
@dannyzaken WDYT?

dannyzaken:
@liranmauda @achouhan09 why do we need `update_account_keys`?

liranmauda:
> @liranmauda @achouhan09 why do we need `update_account_keys`?

We need to set credentials manually to an account
This is part of an Epic for 5.17

src/server/system_services/account_server.js
```diff
@@ -299,6 +299,48 @@ async function generate_account_keys(req) {
...
+        update: {
+            accounts: [{
+                _id: account._id,
+                access_keys: [
+                    access_keys
+                ]
+            }]
+        }
+    })
+    .then(() => Dispatcher.instance().activity({
```
shirady, line 335: In which cases do we post events in the Dispatcher and where can we see them?
Can we write it in `async/await` code style instead of promise chaining `.then(`?

achouhan09: Not sure about other things, but we can see those logs inside `noobaa-core-0` pod in `core` container.
For example(tried it for generate_credentials):
```$ kubectl logs noobaa-core-0 -c core| grep generate_credentials
May-13 13:34:05.669 [WebServer/38]    [L0] core.server.notifications.dispatcher:: Adding ActivityLog entry { event: 'account.generate_credentials', level: 'info', system: 6642162b2aeb8e0026fcec53, actor: 6642162d2aeb8e0026fcec62, account: 664216592aeb8e0026fcec6b, desc: SENSITIVE-d0a52793ea7cebdb }```

shirady (approved):
LGTM

PR #7979, opened 2024-04-16
NC | Secret Keys Encryption Gaps

romayalon:
### Explain the changes
1. **Master Key Manager -**  
    - MasterKey & MasterKeysByID type declarations and some other JSDoc updates
    - create_master_keys_exec() - handle MASTER_KEYS_ALREADY_EXIST error with regular return.
    - get_master_keys_json() - refactoring (see GAP 1).
    
2. **Config.js -** 
    - Added master keys configuration validation - validate_nc_master_keys_config() - 
     Checked for the correctness of the master keys configuration.
     On executable flow - Checked that the executable files exist and can be executed.
    
4. **Unit Tests additions -** 
    - Executable flow tests.
    - Config tests for master keys related configurations.

1. GAP 1- get_master_keys_json() - should be built based on master_keys_by_id and not just the active key.
2. GAP 2 - Need to update the upstream doc.
3. GAP 3 - Add more unit tests.

### Testing Instructions:
1. `sudo jest --testRegex=jest_tests/test_nc_master_keys_exec`
2. `sudo jest --testRegex=jest_tests/test_nc_nsfs_config_schema`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -95,6 +95,7 @@ async function main(argv = minimist(process.argv.slice(2))) {
         const user_input = user_input_from_file || argv;
         config_root = argv.config_root ? String(argv.config_root) : config.NSFS_NC_CONF_DIR;
         if (!config_root) throw_cli_error(ManageCLIError.MissingConfigDirPath);
+        if (argv.config_root) config.NSFS_NC_CONF_DIR = String(argv.config_root);
 
         accounts_dir_path = path.join(config_root, accounts_dir_name);
         access_keys_dir_path = path.join(config_root, access_keys_dir_name);
--- a/src/manage_nsfs/nc_master_key_manager.js
+++ b/src/manage_nsfs/nc_master_key_manager.js
@@ -17,11 +17,36 @@ const native_fs_utils = require('../util/native_fs_utils');
 const TYPE_FILE = 'file';
 const TYPE_EXEC = 'executable';
 const ACTIVE_MASTER_KEY = 'active_master_key';
-const exec_key_suffix = 'master_keys';
+const EXEC_KEY_SUFFIX = 'master_keys';
+
+/**
+* @typedef {{
+*    id: nb.ID, 
+*    cipher_key: Buffer,
+*    cipher_iv: Buffer, 
+*    encryption_type: string
+* }} MasterKey
+*/
+
+/**
+* @typedef {Object.<nb.ID, MasterKey>} MasterKeysByID
+*/
+
+/**
+ * The following constants describe the expected statuses from executable script
+ * OK - both GET/PUT exectuables return on success
+ * VERSION_MISMATCH - PUT exectuable return if the version of master keys already exist
+ * NOT_FOUND - GET exectuable return if the version of master keys was not found
+ */
+const EXEC_STATUS_OK = 'OK';
+const EXEC_STATUS_VERSION_MISMATCH = 'VERSION_MISMATCH';
+const EXEC_STATUS_NOT_FOUND = 'NOT_FOUND';
 
 class NCMasterKeysManager {
     constructor() {
+        /** @type {MasterKey} */
         this.active_master_key = undefined;
+        /** @type {MasterKeysByID} */
         this.master_keys_by_id = {};
         this.last_init_time = 0;
     }
@@ -33,6 +58,7 @@ class NCMasterKeysManager {
 
     /**
      * init inits master_key_manager by the store type
+     * @returns {Promise<void>}
      */
     async init() {
         const store_type = config.NC_MASTER_KEYS_STORE_TYPE;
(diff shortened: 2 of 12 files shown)


src/manage_nsfs/nc_master_key_manager.js
```diff
@@ -123,20 +153,35 @@ class NCMasterKeysManager {
...
     }
 
+
+    /**
+     * get_master_keys_json return master keys as JSON object
+     * @param {MasterKey} new_master_key
+     * @returns {string}
+     */
+    _get_master_keys_json(new_master_key) {
+        return JSON.stringify({ active_master_key: new_master_key });
```
guymguym, line 163: 1. I would add a small comment saying that this is currently limited to one key, but will be changed to support previous inactive keys for rotation.
2. I would add a timestamp inside that json just in case so we could track when it was updated.
3. Do you prefer to write the master_keys_by_id as is and then [ACTIVE_MASTER_KEY] will only contain the id of the active one?
4. Use the const key name -

```suggestion
        return JSON.stringify({ [ACTIVE_MASTER_KEY]: new_master_key });
```

src/manage_nsfs/nc_master_key_manager.js
```diff
@@ -147,36 +192,56 @@ class NCMasterKeysManager {
...
-                const master_keys = await os_util.exec(command, { return_stdout: true });
-                const parsed_master_keys = JSON.parse(master_keys);
-                this._set_keys(parsed_master_keys);
-                this.last_init_time = (new Date()).getTime();
-                return;
-            } catch (err) {
-                if (err.stderr && err.stderr.trim() === 'NO_SUCH_KEY') {
-                    dbg.warn('init_from_exec: couldn\'t find master keys', err, err.code);
+                const get_master_keys_res = await os_util.exec(command, { return_stdout: true });
+                const { status, version, data } = JSON.parse(get_master_keys_res);
```
guymguym, line 229: We are reloading from executable every 5 minutes, and I worry that this behavior is error prone and could block access.
Maybe we should keep this config option but set the default to -1 in order to disable automatic refresh, WDYT?

src/manage_nsfs/nc_master_key_manager.js
```diff
@@ -19,9 +19,36 @@ const TYPE_EXEC = 'executable';
 const ACTIVE_MASTER_KEY = 'active_master_key';
 const exec_key_suffix = 'master_keys';
```
guymguym, line 20: didn't we want to add noobaa to the prefix? and this const deserves to be capitalized
```suggestion
const EXEC_KEY_SUFFIX = 'noobaa.master_keys';
```

src/cmd/manage_nsfs.js
```diff
@@ -95,6 +95,7 @@ async function main(argv = minimist(process.argv.slice(2))) {
         const user_input = user_input_from_file || argv;
         config_root = argv.config_root ? String(argv.config_root) : config.NSFS_NC_CONF_DIR;
         if (!config_root) throw_cli_error(ManageCLIError.MissingConfigDirPath);
+        if (argv.config_root) config.NSFS_NC_CONF_DIR = String(argv.config_root);
```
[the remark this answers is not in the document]
romayalon: the argv.config_root? 
I would love to remove it, it's very annoying

src/manage_nsfs/nc_master_key_manager.js
romayalon: noobaa was the prefix that was mentioned in the script itself

src/manage_nsfs/nc_master_key_manager.js
romayalon: 1 + 4 - np, changed.
Regarding the other topics - 
2. id contains timestamp, is that enough, or you mean other id?
3. make sense, it might take me a while to do these changes

src/manage_nsfs/nc_master_key_manager.js
guymguym: ok then, so without the noobaa prefix is ok

src/manage_nsfs/nc_master_key_manager.js
romayalon: np, changed to -1

src/cmd/manage_nsfs.js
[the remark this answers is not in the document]
romayalon: I meant to the option of passing config_root, 
I added it because if we won't update it the master_keys.json will get written to /etc/noobaa.conf.d/

src/manage_nsfs/nc_master_key_manager.js
romayalon: alright, added timetsamp as well

src/cmd/manage_nsfs.js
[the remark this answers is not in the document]
guymguym: why would it be created always in etc? our config.js code already handles config_dir_redirect no?

guymguym (approved):

src/cmd/manage_nsfs.js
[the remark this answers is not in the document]
romayalon: yes, but this was initially introduced before we had anything, and for cases when there is not config_dir_redirect.
I think we use it only for tests

PR #7983, opened 2024-04-17
NSFS | Improve list objects performance on top of NS FS (PR 2/3)

naveenpaul1:
### Explain the changes
1. This is Second PR for listing unordered objects.
2. This PR contains changes in list_object(), TellDir value is passed to next list object result and same value is used to seekdir dir location.
3. Global flag is added in config for controlling bucket list flow.
NB : New parameter `next_pos` added to API `list_objects` to pass the seek position. I tried to pass this along with `next_marker` as a object but Ceph test failed when comparing `next_marker`.

1. https://github.com/noobaa/noobaa-core/issues/6615

### Testing Instructions:
1. Run ListObject API after enabling global property `ALLOW_NSFS_UNSORTED_OBJECTS_LIST`, Result should retun with unsorted list.

- [ ] Doc added/updated
- [X] Tests added


src/sdk/namespace_fs.js
```diff
@@ -665,7 +667,8 @@ class NamespaceFS {
                 /**
                  * @typedef {{
                  *  key: string,
-                 *  common_prefix: boolean
+                 *  common_prefix: boolean,
+                 *  pos: string,
```
romayalon, line 671: I saw on the PR description you mentioned that you tried to avoid this next_pos and put this value on next_marker as the original solution suggested but you changed it because ceph tests were failing.
But existing ceph tests should not fail because ALLOW_NSFS_UNSORTED_OBJECTS_LIST should be false.
Have you tried to revert it? we should try to solve the issue so we can use marker/next_marker

src/sdk/namespace_fs.js
```diff
@@ -815,10 +824,16 @@ class NamespaceFS {
                 try {
                     dbg.warn('NamespaceFS: open dir streaming', dir_path, 'size', cached_dir.stat.size);
                     dir_handle = await nb_native().fs.opendir(fs_context, dir_path); //, { bufferSize: 128 });
+                    // fix : --recursive listing failes when paggination stop at last elemet of the the dir.
+                    if (next_pos && BigInt(next_pos) !== BigInt(2147483647)) {
```
romayalon, line 828: Why we need to call the same function twice here? what was the error that caused the failure?

src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -42,8 +42,9 @@ async function get_bucket(req) {
         prefix: req.query.prefix,
         delimiter: req.query.delimiter,
         limit: Math.min(max_keys_received, 1000),
+        next_pos: cont_tok_to_key_marker(cont_tok, 'next_pos'),
```
romayalon, line 45: will be redundant when changing to next_marker

romayalon (changes requested):

shirady:
Hi @naveenpaul1,
I found that when running the tests on MacOS some of them don't pass:
I tried: `npx jest test_list_object.test.js`
I see:

>   manage unsorted list objcts flow
>     Unsorted List objects
>       ✓ List object unsorted with one object upload (12 ms)
>       ✓ List object unsorted with multiple object upload (671 ms)
>     Telldir and Seekdir implementation
>       ✓ telldir returns bigint (1 ms)
>       ✓ seekdir expects bigint (13 ms)
>       ✕ list dir files - telldir and seekdir. (2 ms)
>       ✕ list dir files -  Dir.read() and seekdir() (1 ms)
>       ✕ list 10000 dir files - telldir and seekdir (1074 ms)

Details:

>   ● manage unsorted list objcts flow › Telldir and Seekdir implementation › list dir files - telldir and seekdir.
> 
>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: 5
>     Received: 8
> 
>       134 |             }
>       135 |             //total number of dir and files inside list_dir_root is 5
>     > 136 |             expect(total_dir_entries).toBe(total_files + 1);
>           |                                       ^
>       137 |         });
>       138 |
>       139 |         it('list dir files -  Dir.read() and seekdir()', async () => {
> 
>       at Object.toBe (src/test/unit_tests/jest_tests/test_list_object.test.js:136:39)
> 
>   ● manage unsorted list objcts flow › Telldir and Seekdir implementation › list dir files -  Dir.read() and seekdir()
> 
>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: 0n
>     Received: 1n
> 
>       148 |                 const tell_dir = await dir_handle.telldir(DEFAULT_FS_CONFIG);
>       149 |                 //verify tell_dir and dir_entry.off return same value
>     > 150 |                 expect(tell_dir).toBe(dir_entry.off);
>           |                                  ^
>       151 |                 dir_marker = {
>       152 |                     dir_path: list_dir_root,
>       153 |                     pos: dir_entry.off,
> 
>       at Object.toBe (src/test/unit_tests/jest_tests/test_list_object.test.js:150:34)
> 
>   ● manage unsorted list objcts flow › Telldir and Seekdir implementation › list 10000 dir files - telldir and seekdir
> 
>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: 10000
>     Received: 10501
> 
>       208 |             }
>       209 |             //total number of dir and files inside list_dir_root is 5
>     > 210 |             expect(total_dir_entries).toBe(10000);
>           |                                       ^
>       211 |         }, 10000);
>       212 |     });
>       213 | });
> 
>       at Object.toBe (src/test/unit_tests/jest_tests/test_list_object.test.js:210:39)

src/sdk/namespace_fs.js
naveenpaul1: @romayalon checking this

naveenpaul1:
@shirady could you please try the test after running 
`npm run build:native`

config.js
```diff
@@ -795,6 +795,9 @@ config.NSFS_GLACIER_MIGRATE_INTERVAL = 15 * 60 * 1000;
 // of `manage_nsfs glacier restore`
 config.NSFS_GLACIER_RESTORE_INTERVAL = 15 * 60 * 1000;
 
+// enable/disable unsorted listing application level
+config.ENABLE_NSFS_UNSORTED_OBJECTS_LIST = false;
```
guymguym, line 799: I would use "NSFS_LIST_OBJECTS_V2_UNSORTED_" as prefix and add "ENABLED" as suffix to the config name, so we can also add other suffixes for options of this feature -
```suggestion
config.NSFS_LIST_OBJECTS_V2_UNSORTED_ENABLED = false; 
```

src/test/unit_tests/jest_tests/test_list_object.test.js
```diff
@@ -1,175 +0,0 @@
...
-const nb_native = require('../../../util/nb_native');
-const {TMP_PATH} = require('../../system_tests/test_utils');
-const { get_process_fs_context } = require('../../../util/native_fs_utils');
-
-const tmp_fs_path = path.join(TMP_PATH, 'test_list_object');
-const DEFAULT_FS_CONFIG = get_process_fs_context();
-
-// eslint-disable-next-line max-lines-per-function
-describe('manage list objct flow', () => {
-    describe('Telldir and Seekdir implementation', () => {
```
guymguym, line 17: Did you remove this test? or just moved it elsewhere? It just seems like a good test to have..

src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -70,7 +70,7 @@ async function get_bucket(req) {
                     reply.next_marker, reply.objects, reply.is_truncated),
             } : { // list_type v1
                 Marker: req.query.marker || '',
-                NextMarker: req.query.delimiter ? reply.next_marker : undefined,
+                NextMarker: reply.next_marker?.marker_pos ? JSON.stringify(reply.next_marker) : reply.next_marker,
```
guymguym, line 73: I don't think we can change that for list_objects v1 at all, this will break compatibility.
We should do that only for `NextContinuationToken` in list_type === '2' above.

src/sdk/namespace_fs.js
```diff
@@ -609,6 +609,7 @@ class NamespaceFS {
     }
 
     async _list_objects(params, object_sdk, list_versions) {
+        let list_unsorted = false;
```
guymguym, line 612: it seems that you forced this to list_objects v1, which we cannot really do because the clients assume that NextMarker is only returned in some cases of abiguity (when both objects and prefixes are returned), but otherwise in flat listing the clients typically have to assume that the marker is the Key of the last entry.

So I think we want to apply this only for list_type === '2'. Some of the code can be shared, but preferably not use the same _list_objects() method for both cases (instead _list_objects should have some common code and the call _list_objects_v1 or _list_objects_v2) because I think this is too messy to read, even now it is a long and hard function to maintain.

guymguym (changes requested):

src/test/unit_tests/jest_tests/test_list_object.test.js
naveenpaul1: Its not removed it moved inside same file.

PR #7990, opened 2024-04-21
github stale action

guymguym:
### Explain the changes
1. We have too many stale issues and PR's.
2. Added the [stale action](https://github.com/marketplace/actions/close-stale-issue).

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,33 @@
+# This workflow warns and then closes issues and PRs that have had no activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see:
+# https://github.com/actions/stale
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+    - cron: '0 11 * * *'
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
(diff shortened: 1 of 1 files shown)


nimrod-becker:
Ack.
Didnt notice the stale bot was deprecated

.github/workflows/stale.yml
```diff
@@ -0,0 +1,33 @@
...
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v5
+      with:
+        days-before-stale: 60
+        days-before-pr-stale: 60
+        days-before-issue-stale: 60
```
nimrod-becker, line 25: too little of an interval. Lets set to 180 days

guymguym: I think 6 months is too much. I would keep it 90 days, and if it gets stale we can always post a comment or remove the stale label to avoid closing it.

nimrod-becker: 90 days in the issue list are still things we don't want to lose. It's too short.
You will still get your value with 180 days

guymguym: we don't close after 90, we only mark stale.

guymguym: currently in this config it will close if there is no update for another 30 days after labeling as stale

nimrod-becker (approved):

PR #7991, opened 2024-04-22
[NC | NSFS | Glacier] Support max restore days + auto logs dir creation

### LINKED ISSUE #7976: Automatic creation of /var/run/noobaa-nsfs/wal folder not happening on enabling glacier
### Environment info
- NooBaa Version: noobaa-core-5.15.0-20240320.el9.x86_64
- Platform: standalone

### Actual behavior
1. When we are trying to enable the glacier support using noobaa by adding the below configuration parameters is expecting manual creation of /var/run/noobaa-nsfs/wal folder for uploading the object.
NSFS_GLACIER_ENABLED=true
NSFS_GLACIER_LOGS_ENABLED=false

### Expected behavior
1. It should create the /var/run/noobaa-nsfs/wal folder automatically on adding these 2 params to config.json

### Steps to reproduce
1.

tangledbytes:
### Explain the changes
Fix: Create persistent logs directory if dont exist
Feat: Support max restore days

1. Fixes #7976

### Testing Instructions:
2. `./node_modules/.bin/jest --testRegex=jest_tests/test_s3_utils`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/test/unit_tests/test_nsfs_glacier_backend.js
+++ b/src/test/unit_tests/test_nsfs_glacier_backend.js
@@ -90,8 +90,8 @@ mocha.describe('nsfs_glacier', async () => {
 		const backend = new TapeCloudGlacierBackend();
 
 		// Patch backend for test
-		backend._migrate = async () => [];
-		backend._recall = async () => [];
+		backend._migrate = async () => { /**noop */ };
+		backend._recall = async () => { /**noop */ };
 		backend._process_expired = async () => { /**noop*/ };
 
 		mocha.it('upload to GLACIER should work', async () => {
--- a/config.js
+++ b/config.js
@@ -176,6 +176,17 @@ config.STS_CORS_EXPOSE_HEADERS = 'ETag';
 
 config.DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD = false;
 
+// S3_RESTORE_MAX_DAYS controls that for how many maximum number
+// of days an object can be restored using `restore-object` call.
+config.S3_RESTORE_REQUEST_MAX_DAYS = 30;
+
+/**
+ * S3_RESTORE_MAX_DAYS_BEHAVIOUR controls whether to truncate the
+ * requested number of days in restore request or whether to deny the request.
+ * @type {'DENY' | 'TRUNCATE'}
+ */
+config.S3_RESTORE_REQUEST_MAX_DAYS_BEHAVIOUR = 'TRUNCATE';
+
 /////////////////////
 // SECRETS CONFIG  //
 /////////////////////
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -63,8 +63,13 @@ async function check_and_create_config_dirs() {
         config_root,
         buckets_dir_path,
         accounts_dir_path,
-        access_keys_dir_path
+        access_keys_dir_path,
     ];
+
+    if (config.NSFS_GLACIER_LOGS_ENABLED) {
+        pre_req_dirs.push(config.NSFS_GLACIER_LOGS_DIR);
+    }
+
     for (const dir_path of pre_req_dirs) {
         try {
             const fs_context = native_fs_utils.get_process_fs_context(config_root_backend);
--- a/src/endpoint/s3/ops/s3_post_object_restore.js
+++ b/src/endpoint/s3/ops/s3_post_object_restore.js
@@ -1,9 +1,7 @@
 /* Copyright (C) 2023 NooBaa */
 'use strict';
 
-const { S3Error } = require('../s3_errors');
 const s3_utils = require('../s3_utils');
-const dbg = require('../../../util/debug_module')(__filename);
 
 /**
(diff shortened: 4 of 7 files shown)


src/test/unit_tests/test_nsfs_glacier_backend.js
```diff
@@ -168,5 +169,30 @@ mocha.describe('nsfs_glacier', async () => {
...
+				storage_class: s3_utils.STORAGE_CLASS_GLACIER,
+                xattr,
+				days: config.NSFS_GLACIER_RESTORE_MAX_DAYS + 1,
+                source_stream: buffer_utils.buffer_to_read_stream(data)
+            };
+
+            const upload_res = await glacier_ns.upload_object(params, dummy_object_sdk);
+            console.log('upload_object response', inspect(upload_res));
+
+			try {
```
guymguym, line 187: check your tabs vs spaces?

src/sdk/namespace_fs.js
```diff
@@ -2021,6 +2021,12 @@ class NamespaceFS {
         dbg.log0('namespace_fs.restore_object:', params);
 
         await this._throw_if_storage_class_not_supported(s3_utils.STORAGE_CLASS_GLACIER);
+        if (params.days > config.NSFS_GLACIER_RESTORE_MAX_DAYS) {
+            throw new S3Error({
+                ...S3Error.BadRequest,
+                detail: `restore days cannot be more than ${config.NSFS_GLACIER_RESTORE_MAX_DAYS}`},
+            );
+        }
```
guymguym, line 2029: I am not sure about failing. Perhaps we should just truncate the number?

tangledbytes: Yeah, I think adding another option for the behaviour is better.

src/test/unit_tests/test_nsfs_glacier_backend.js
tangledbytes: I will shift the blame on VSCode 😆 .

Fixed it.

src/sdk/namespace_fs.js
```diff
@@ -2054,6 +2054,17 @@ class NamespaceFS {
...
+            if (config.NSFS_GLACIER_RESTORE_MAX_DAYS_BEHAVIOUR === 'DENY') {
+                throw new S3Error({
+                    ...S3Error.BadRequest,
+                    detail: `restore days cannot be more than ${config.NSFS_GLACIER_RESTORE_MAX_DAYS}`},
+                );
+            }
+
+            dbg.warn(`restore_object days exceed max configured days (${config.NSFS_GLACIER_RESTORE_MAX_DAYS}) - truncating`);
+            params.days = config.NSFS_GLACIER_RESTORE_MAX_DAYS;
+        }
```
guymguym, line 2067: Logic looks good now, but now that I see this I would like to generalize the check for S3, not just nsfs, as this is a pure protocol check that is not specific to nsfs. So I suggest to rename the configs to have S3_ prefix instead of NSFS_ and refactor. Notice I changed below to return S3Error.InvalidArgument perhaps it fits better than BadRequest?

In s3_post_object_restore.js to extract [this code block](https://github.com/noobaa/noobaa-core/blob/5914557f0ee3e932093d22e92ec96be068d3618b/src/endpoint/s3/ops/s3_post_object_restore.js#L14-L23) and just call a utility function:
```js
const days = s3_utils.parse_restore_request_days(req);
...
```

in s3_utils:
```js
function parse_restore_request_days(req) {
    if (!req.body?.RestoreRequest?.Days?.[0]) {
        dbg.warn('parse_restore_request_days: missing Days in body');
        throw new S3Error(S3Error.MalformedXML);
    }

    const days = s3_utils.parse_decimal_int(req.body.RestoreRequest.Days[0]);
    if (days < 1) {
        dbg.warn('parse_restore_request_days: days cannot be less than 1');
    }

    if (days > config.S3_RESTORE_REQUEST_MAX_DAYS) {
        if (config.S3_RESTORE_REQUEST_MAX_DAYS_BEHAVIOR === 'DENY') {
            throw new S3Error({
                ...S3Error.InvalidArgument,
                detail: `Restore request days ${days} is above max ${config.S3_RESTORE_REQUEST_MAX_DAYS}`},
            );
        }

        dbg.log0(`Restore request days ${days} is above max ${config.S3_RESTORE_REQUEST_MAX_DAYS} - truncating`);
        return config.S3_RESTORE_REQUEST_MAX_DAYS;
    }

    return days;
}
```

tangledbytes: Actually I did initially consider to place this at the more generic location but then figured that probably we might not need this config everywhere. For example any cloud backed storages. WDYT?

guymguym: You are right that for buckets with unlimited cloud archive tier it might be preferrable to have the limit higher, but I still don't think that functionally it makes sense to have no limit - notice that S3 client has no control over premature expiry so once you set it to 100 years you are kind of stuck.

So I would say this - eventually this might make sense as a per-bucket option, but for now I am ok with this being a system wide config, and I would apply it to all types of buckets.

guymguym (approved):

PR #7996, opened 2024-04-24
Bucket_logging: Time bound bucket logs delivery

aspandey:
BZ: [2272932](https://bugzilla.redhat.com/show_bug.cgi?id=2272932)
Log rotation, segregation and upload of log objects for different buckets happens based on size and time. If the size of logs are less than the configured value, it might take time to upload the log objects.

We need to consider the following factors behind log collection and processing. 1 - logrotation depends on size of the log file, bucket_logs.log, or time of the day. Currently, the size is set to 100KB. logrotate wakes up every 1 minute. That means, once the bucket_logs.log file size reaches more than 100KB, logrotates will come up and rotate this file. If the size has not reached 100K, it still rotates the file after every 24hr.

2 - After rotation, bucket_logs.log gets processed by logsegregate.sh which gets executed every 5 min. This creates log objects for individual buckets. We have a limit of size of log objects which is 50K. Once a log object reaches 50K, it will be transferred to the log bucket.

3 - Background process, which actually copies the log object in log buckets, scans the log path every 5 minutes. If it finds any log object, it copies it to a log bucket.

If log is not delivered because it is not rotated in step 1 based on size, then that can only be rotated on daily basis. We can not do anything there.

However, we have control on step 2. If the size of the log object for a bucket is less than 50K, it is not being uploaded even after 2-3 hours or even 24 hr.

Solution: We need to make sure that even if the size of log object for a bucket is less than 50K, it get uploaded once the threshold of time reaches, which is set to be 5 minutes.

### DIFF
--- a/src/deploy/NVA_build/noobaa_log_segregate.sh
+++ b/src/deploy/NVA_build/noobaa_log_segregate.sh
@@ -8,10 +8,23 @@
 # into respective log files. These files will be moved to the final location
 # /log/noobaa_bucket_logs/ which will be scanned by the background worker.
 
-
 log_dir="/var/log/noobaa_logs"
+NOOBAA_BUCKET_LOGS_DIR="/var/log/noobaa_bucket_logs/"
+NOOBAA_BUCKET_LOGS_FINAL="/log/noobaa_bucket_logs/"
+BUCKET_LOG_FILE_SIZE_LIMIT=51200
+# Delimiter for the file name. To separate the file name from the timestamp.
+# S3 Bucket name are not allowed to have "_" in its name.
+File_Name_Del="_"
+
 # Lock file to make sure that the script is not running multiple times.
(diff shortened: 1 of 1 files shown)


src/deploy/NVA_build/noobaa_log_segregate.sh
```diff
@@ -49,28 +57,43 @@ oldest_log_file() {
     echo "$earliest_filename"
 }
 
+move_old_small_bucket_logs() {
```
vh05, line 60: Does this loop run continously ?

aspandey: NO. It will check if there are files in /var/log/noobaa_bucket_logs/ with .log extension. If it is then it will move it. If not then it will come out of loop.

vh05: okay

vh05 (approved):
lgtm

PR #7997, opened 2024-04-24
NSFS | Support anonymous NSFS requests

### LINKED ISSUE #7519: standalone nsfs: process getting crashed when we hit anonymous tests
Process getting crashed when we trigger anonumous object operations. I think the error handling at object_sdk.js i.e 

`throw new RpcError(`......`);` needs to be checked. 

When we run 

s3tests_boto3/functional/test_s3.py::test_bucket_list_objects_anonymous_fail
s3tests_boto3/functional/test_s3.py::test_bucket_listv2_objects_anonymous_fail

Stack trace:
```

Oct-3 11:24:11.486 [nsfs/15] [ERROR] core.endpoint.s3.s3_rest:: S3 ERROR <?xml version="1.0" encoding="UTF-8"?><Error><Code>AccessDenied</Code><Message>Access Denied</Message><Resource>/ceph-xa7ico4c0ht6nc6ar2fcep6v-1?encoding-type=url</Resource><RequestId>lna8fa6z-1vbxv7-14sj</RequestId></Error> GET /ceph-xa7ico4c0ht6nc6ar2fcep6v-1?encoding-type=url {"host":"localhost","accept-encoding":"identity","user-agent":"Boto3/1.23.10 Python/3.6.8 Linux/6.2.8-100.fc36.x86_64 Botocore/1.26.10","amz-sdk-invocation-id":"cdb17cda-7926-4d60-a19e-5c367aa39098","amz-sdk-request":"attempt=1"} Error: Anonymous access to namespace buckets not supported
    at NsfsObjectSDK.authorize_request_account (/root/node_modules/noobaa-core/src/sdk/object_sdk.js:226:27)
    at async Promise.all (index 0)
    at async authorize_request (/root/node_modules/noobaa-core/src/endpoint/s3/s3_rest.js:200:5)
    at async handle_request (/root/node_modules/noobaa-core/src/endpoint/s3/s3_rest.js:113:5)
    at async Object.s3_rest [as handler] (/root/node_modules/noobaa-core/src/endpoint/s3/s3_rest.js:64:9)
Oct-3 11:24:11.487 [nsfs/15] [ERROR] CONSOLE:: PANIC: process uncaughtException Error: Access Denied
    at authorize_anonymous_access (/root/node_modules/noobaa-core/src/endpoint/s3/s3_rest.js:249:27)
    at authorize_request_policy (/root/node_modules/noobaa-core/src/endpoint/s3/s3_rest.js:219:9)
    at async Promise.all (index 1)
    at async authorize_request (/root/node_modules/noobaa-core/src/endpoint/s3/s3_rest.js:200:5)
    at async handle_request (/root/node_modules/noobaa-core/src/endpoint/s3/s3_rest.js:113:5)
    at async Object.s3_rest [as handler] (/root/node_modules/noobaa-core/src/endpoint/s3/s3_rest.js:64:9)


```

naveenpaul1:
### Explain the changes
1. Support anonymous NSFS requests
2. add CLI for add/updat/delete/status the anonymous account
3. Ceph test `test_post_object_wrong_bucket` added to black list, `post_bucket` action is not supported
4. load_requesting_account() will load account from anonymous account. UID and GID or distinguished_name from anonymous account use for this.
5. anonymous user to account_cache with empty access key and for that read_account_by_access_key() method is updated 

1. Changes added to support anonymous request in namespace store
2. remove namesapce  store check in code that prevent anonymous request

### Testing Instructions:
1. Add anonymous account 
`node src/cmd/manage_nsfs account add --anonymous --uid {uid} --gid {gid}`
2. Create Bucket with policy that support anonymous request
`{
            Effect: 'Allow',
            Principal: { AWS: ["*"] },
            Action: ['s3:GetObject', 's3:ListBucket'],
            Resource: ['arn:aws:s3:::*']
        } `
4. access object item using anonymous s3 request

- [X] Doc added/updated
- [X] Tests added


romayalon:
Hi @naveenpaul1 :)
This week I added the test_s3_bucket_policy.js to the Non Containerized tests, I saw some anonymous tests in this tests file and skipped them, can you please check it and unskip the relevant tests?
reference - https://github.com/noobaa/noobaa-core/pull/8022

src/sdk/object_sdk.js
```diff
@@ -207,7 +207,16 @@ class ObjectSDK {
...
             const token = this.get_auth_token();
-            if (!token) return;
+            const is_anon = !(token && token.access_key);
+            if (is_anon) {
+                this.requesting_account = {
+                    nsfs_account_config: {
+                        uid: process.getuid(),
+                        gid: process.getgid(),
+                    }
+                };
```
guymguym, line 217: I think this is unsafe - we are creating a "partial" account for this request, and also letting it use the process elevated FS permissions.

I suggest a different approach - how about that all anonymous access will look up a "fixed" account that would have to be created in the system to allow annonymous access. That account will not have access keys off course, but will have its own nsfs_account_config provided so that it can be configured to run as a regular user on the FS. We need to hash out the details of that, but I think this approach will provide much more flexibility and security. 

WDYT @naveenpaul1 @romayalon @madhuthorat ?

guymguym (changes requested):

src/cmd/nsfs.js
```diff
@@ -235,6 +239,30 @@ async function init_nsfs_system(config_root) {
         throw error;
     }
 }
+// Create an anonymous account for anonymous request. Use this account UID and GID for bucket access.
+async function add_anonymous_account() {
```
romayalon, line 243: @naveenpaul1 @guymguym 
I had a quick sync with Guy today, and we also discussed this account creation. Guy suggests that for now we won't create this account automatically, instead we will - 
1. Let the user create an anonymous account using the CLI (uid/gid will be determined on the create/update request, no access keys are attached to this account)
2. Add a documentation section for anonymous and explain that before running an anonymous s3 request there are steps to take - 
- Create an account using the CLI while specifying --anonymous flag
- Create bucket policy on a bucket specifying that anonymous access is allowed
- Run the s3 anonymous request on the bucket

This solution makes our life a bit easier, but we need to pay attention to the fact that currently we can not have multiple anonymous users. An easy way to restrict this is to use a hard-coded name as you did. WDYT?

src/endpoint/s3/s3_rest.js
```diff
@@ -264,7 +264,8 @@ function _get_method_from_req(req) {
     const s3_op = s3_bucket_policy_utils.OP_NAME_TO_ACTION[req.op_name];
     if (!s3_op) {
         dbg.error(`Got a not supported S3 op ${req.op_name} - doesn't suppose to happen`);
-        throw new S3Error(S3Error.InternalError);
+        // Added access denied error to align with Ceph test test_post_object_wrong_bucket
+        throw new S3Error(S3Error.AccessDenied);
```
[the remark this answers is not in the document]
naveenpaul1: There are two things
1. test_post_object_wrong_bucket ceph test should be black listed because `post_bucket` action is not supported. I will add this test to black list.
2. This test was working before because request was considered anonymous and anonymous request was returning 403 error before.
Ceph is passing `AWSAccessKeyId` and `signature` as payload and But `make_auth_token_from_request()` have a [check](https://github.com/noobaa/noobaa-core/blob/master/src/util/signature_utils.js#L269) `if (req.query.AWSAccessKeyId && req.query.Signature)` where `req.query` is empty.

src/cmd/nsfs.js
naveenpaul1: changes are commited, Please review

src/util/native_fs_utils.js
```diff
@@ -419,6 +416,20 @@ async function get_user_by_distinguished_name({ distinguished_name }) {
     }
 }
 
+async function get_anonymous_account_by_name(anony_account_name) {
+    try {
+        if (!anony_account_name) throw new Error('no anonymous account name');
+        const context = get_process_fs_context();
+        const iam_path = path.join(config.NSFS_NC_CONF_DIR, 'accounts', anony_account_name + '.json');
```
romayalon, line 423: This function can do some re-uses :) 
1.  call get_config_file_path(config_type_path, file_name) for getting the file path
2. instead of 'accounts' use manage_nsfs_constants.CONFIG_SUBDIRS.ACCOUNTS
3. call get_config_data() for reading and parsing

romayalon: also, I prefer keeping the anon shortname as in object_sdk if you don't mind (instead of anony)

src/sdk/object_sdk.js
```diff
@@ -245,18 +271,11 @@ class ObjectSDK {
...
         if (bucket && req.op_name !== 'put_bucket') {
-            // ANONYMOUS: cannot work without bucket, cannot work on namespace bucket (?)
+            // ANONYMOUS: cannot work without bucket.
+            // Return if the acount is anonymous
+            const is_anon = !(token && token.access_key);
+            if (is_anon) return;
             const ns = await this.read_bucket_sdk_namespace_info(bucket);
-            if (!token) {
-                // TODO: Anonymous access to namespace buckets not supported
-                if (ns) {
```
romayalon, line 252: anonymous access is still not supported on containerized environments, no?
IIRC this should not be removed, but adding the NC env check to the condition

src/manage_nsfs/manage_nsfs_anonymous.js
```diff
@@ -0,0 +1,82 @@
...
+const mongo_utils = require('../util/mongo_utils');
+const ManageCLIError = require('../manage_nsfs/manage_nsfs_cli_errors').ManageCLIError;
+const ManageCLIResponse = require('../manage_nsfs/manage_nsfs_cli_responses').ManageCLIResponse;
+const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
+const { ACTIONS} = require('../manage_nsfs/manage_nsfs_constants');
+const { throw_cli_error, write_stdout_response, get_config_file_path,
+    get_config_data} = require('../manage_nsfs/manage_nsfs_cli_utils');
+const { validate_anonymous_account_args } = require('../manage_nsfs/manage_nsfs_validations');
+
+
```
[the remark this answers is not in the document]
romayalon: also, anonymous account can not create buckets, is that correct? this is another difference in the configs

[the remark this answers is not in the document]
naveenpaul1: @romayalon Both anon and normal account dont have that much commen inside add/update/delete methods, that because of the missing access key. These methods have access key and encryption code that is not relevent to anon account and controlling it using the check will add complexity to the code. If you want I can try that.

docs/non_containerized_NSFS.md
```diff
@@ -642,3 +642,28 @@ Access is restricted to these whitelist IPs. If there are no IPs mentioned all I
...
+```
+
+Delete: 
+```
+node src/cmd/manage_nsfs account delete --anonymous
+```
+Status
+```
+node src/cmd/manage_nsfs account status --anonymous
+```
```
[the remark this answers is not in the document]
naveenpaul1: @romayalon what you mean by policy command for anonymous requests?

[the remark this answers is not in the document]
romayalon: usually is the docs we add s3 client commands, so I meant to add s3api put-bucket-policy and specify the policy to allow anonymous requests (allow * as written in tests)

src/server/system_services/schemas/nsfs_account_schema.js
```diff
@@ -8,11 +8,9 @@ module.exports = {
         '_id',
         'name',
         'email', // temp, keep the email internally
-        'access_keys',
```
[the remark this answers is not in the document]
shirady: @naveenpaul1 if it helps I added a test here:
https://github.com/noobaa/noobaa-core/pull/8055/files#diff-a2598e99155f4104aee08363d420621d1489dcba5085a218b120741db65c67ac

Since the anonymous account and IAM account user (without access key) have similar attributes I will also share the way I create a new user (all were not reviewed yet):
https://github.com/noobaa/noobaa-core/pull/8055/files#diff-e17fde3b7bbe324918c2a7a422770715bd439ab6de21cc76f6679a0b7f31ecd0R442-R464
(if the link doesn't work, search `_new_user_defaults`).
cc: @romayalon @guymguym

docs/non_containerized_NSFS.md
```diff
@@ -642,3 +642,55 @@ Access is restricted to these whitelist IPs. If there are no IPs mentioned all I
...
+Anonymous requests are those sent without access and secret key and Noobaa NSFS allow it with supporting bucket policy. Users have to create an anonymous account before accessing the resources. Noobaa uses anonymous account's uid and gid or user(distinguished_name) for accessing bucket storage paths. Noobaa can have one anonymous account. Commands for creating anonumous account are
+```
+node src/cmd/manage_nsfs account add --anonymous --uid {uid} --gid {gid}
+```
+or
+```
+node src/cmd/manage_nsfs account add --anonymous --user {user}
+```
+
+Update :
```
shirady, line 657: I think you can add the or option as well for `--user` flag (like you have in account add).

src/manage_nsfs/manage_nsfs_constants.js
```diff
@@ -6,6 +6,7 @@ const TYPES = {
     BUCKET: 'bucket',
     IP_WHITELIST: 'whitelist',
     GLACIER: 'glacier',
+    ANONYMOUS_ACCOUNT: 'anonymous_account'
```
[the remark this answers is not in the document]
naveenpaul1: @romayalon I am using this option to verify anonymous input values. I have added specif validation for anonymous account based on ANONYMOUS_ACCOUNT option

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -91,6 +91,8 @@ function validate_no_extra_options(type, action, input_options, is_options_from_
         valid_options = VALID_OPTIONS.account_options[action];
     } else if (type === TYPES.GLACIER) {
         valid_options = VALID_OPTIONS.glacier_options[action];
+    } else if (type === TYPES.ANONYMOUS_ACCOUNT) {
+        valid_options = VALID_OPTIONS.anonymouse_account_options[action];
```
shirady, line 95: If it helps you in case you remove the `TYPE.ANONYMOUS_ACCOUNT` we have the `from_file_condition`, you can do something similar.

src/util/native_fs_utils.js
```diff
@@ -419,6 +418,20 @@ async function get_user_by_distinguished_name({ distinguished_name }) {
     }
 }
 
+async function get_anonymous_account_by_name(anon_account_name) {
```
shirady, line 421: The name is constant (`config.NC_ANON_ACCOUNT_NAME`), maybe you can remove this argument from the function signature.

src/server/system_services/schemas/nsfs_account_schema.js
```diff
@@ -8,11 +8,9 @@ module.exports = {
...
         'nsfs_account_config',
         'creation_date',
         'allow_bucket_creation',
-        'master_key_id'
```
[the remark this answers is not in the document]
naveenpaul1: We are removing the `master_key_id`, not leftovers, WDYT?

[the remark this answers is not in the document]
romayalon: let's not remove it 👍🏻

src/manage_nsfs/manage_nsfs_cli_utils.js
```diff
@@ -109,6 +109,15 @@ async function get_options_from_file(file_path) {
     }
 }
 
+/**
+ * is_anonymous will return anonymous or not depending on the access key length 
+ * (instead of flags) and return its content
+ * @param {object[]} access_keys
+ */
+function is_anonymous(access_keys) {
+    return access_keys.length > 0;
```
romayalon, line 118:
```suggestion
    return access_keys.length === 0;
```

src/cmd/manage_nsfs.js
```diff
@@ -342,11 +346,13 @@ async function fetch_account_data(action, user_input) {
 
     // override values
     // access_key as SensitiveString
-    data.access_keys[0].access_key = _.isUndefined(data.access_keys[0].access_key) ? undefined :
+    if (is_anonymous(data.access_keys)) {
```
romayalon, line 349:
```suggestion
    if (!is_anonymous(data.access_keys)) {
```

src/server/system_services/schemas/nsfs_account_schema.js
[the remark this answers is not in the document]
romayalon: @naveenpaul1 I meant that we need to still have master_key_id in schema and not remove master_key_id from schema and when creating anonymous account

src/cmd/manage_nsfs.js
```diff
@@ -406,16 +413,25 @@ async function add_account(data) {
...
-    const encrypted_account = await nc_mkm.encrypt_access_keys(data);
-    data.master_key_id = encrypted_account.master_key_id;
-    const encrypted_data = JSON.stringify(encrypted_account);
+    data.master_key_id = await nc_mkm.get_active_master_key();
+    let encrypted_data;
+    if (!is_anonymous(data.access_keys)) {
+        const encrypted_account = await nc_mkm.encrypt_access_keys(data);
+        data.master_key_id = encrypted_account.master_key_id;
+        encrypted_data = JSON.stringify(encrypted_account);
+    }
```
romayalon, line 422: Can we maybe avoid this condition here and always encrypt?
IIUC we will init() and return the active_master_key even if the array is empty - 
```
async encrypt_access_keys(account) {
        await this.init();
        const master_key_id = this.active_master_key.id;
        const encrypted_access_keys = await P.all(_.map(account.access_keys, async access_keys => ({
            access_key: access_keys.access_key,
            encrypted_secret_key: await this.encrypt(access_keys.secret_key, master_key_id)
        })));
        return { ...account, access_keys: encrypted_access_keys, master_key_id };
    }
```

romayalon (approved):

src/manage_nsfs/manage_nsfs_cli_utils.js
```diff
@@ -109,6 +109,15 @@ async function get_options_from_file(file_path) {
     }
 }
 
+/**
+ * is_anonymous will return anonymous or not depending on the access key length 
+ * (instead of flags) and return its content
+ * @param {object[]} access_keys
+ */
+function is_anonymous(access_keys) {
```
romayalon, line 117: rename to has_access_keys please

docs/non_containerized_NSFS.md
```diff
@@ -642,3 +642,60 @@ Access is restricted to these whitelist IPs. If there are no IPs mentioned all I
 ```
 sudo noobaa-cli whitelist --ips '["127.0.0.1", "192.000.10.000", "3002:0bd6:0000:0000:0000:ee00:0033:000"]'  2>/dev/null
 ```
+
+## Anonymous request
+
+Anonymous requests are those sent without access and secret key and Noobaa NSFS allow it with supporting bucket policy. Users have to create an anonymous account before accessing the resources. Noobaa uses anonymous account's uid and gid or user(distinguished_name) for accessing bucket storage paths. Noobaa can have one anonymous account. Commands for creating anonumous account are
+```
+node src/cmd/manage_nsfs account add --anonymous --uid {uid} --gid {gid}
```
romayalon, line 650: can someone create/update a regular account called anonymous? we need to block it if so

src/sdk/object_sdk.js
```diff
@@ -245,18 +256,20 @@ class ObjectSDK {
         }
         // check for a specific bucket
         if (bucket && req.op_name !== 'put_bucket') {
-            // ANONYMOUS: cannot work without bucket, cannot work on namespace bucket (?)
+            // ANONYMOUS: cannot work without bucket.
+            // Return if the acount is anonymous
+            const is_anon = !(token && token.access_key);
```
romayalon, line 261: this causing s3 requests with access key empty and secret key non empty to make anon requests instead of failing like happens in AWS

src/sdk/bucketspace_fs.js
```diff
@@ -96,8 +96,8 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
 
     async read_account_by_access_key({ access_key }) {
         try {
-            if (!access_key) throw new Error('no access key');
-            const iam_path = this._get_access_keys_config_path(access_key);
+            const iam_path = access_key === '' ? this._get_account_config_path(config.NC_ANON_ACCOUNT_NAME) :
```
guymguym, line 99: Using empty string or even undefined can be accidental and we really don't want to allow that.
Lets do this - define a new `const anonymous_access_key = Symbol('anonymous_access_key')` in object_sdk.
Then we should use this symbol as the key to the cache, but be explicit about it and not implicit.

src/sdk/object_sdk.js
```diff
@@ -207,19 +207,17 @@ class ObjectSDK {
     async load_requesting_account(req) {
         try {
             const token = this.get_auth_token();
-            if (!token) return;
+            const is_anon = !(token && token.access_key);
```
guymguym, line 210: also here we want to only check !token and not assume empty access key is also anon, right?

docs/non_containerized_NSFS.md
```diff
@@ -642,3 +642,60 @@ Access is restricted to these whitelist IPs. If there are no IPs mentioned all I
...
+```
+or
+```
+node src/cmd/manage_nsfs account add --anonymous --user {user}
```
guymguym, line 654: change all calls from manage_nsfs to noobaa-cli -
```suggestion
noobaa-cli account add --anonymous --user {user}
```

romayalon (approved):

src/sdk/object_sdk.js
```diff
@@ -207,19 +207,16 @@ class ObjectSDK {
...
                 const distinguished_name = this.requesting_account.nsfs_account_config.distinguished_name.unwrap();
-                const user = await dn_cache.get_with_cache({
-                    bucketspace: this._get_bucketspace(),
-                    distinguished_name,
-                });
-                this.requesting_account.nsfs_account_config.uid = user.uid;
-                this.requesting_account.nsfs_account_config.gid = user.gid;
+                const account_config = await this.get_distinguished_user_details(distinguished_name);
+                this.requesting_account.nsfs_account_config.uid = account_config.nsfs_account_config.uid;
+                this.requesting_account.nsfs_account_config.gid = account_config.nsfs_account_config.gid;
```
guymguym, line 219: this change seems to be unneeded now right? it feels a bit weird that the get_distinguished_user_details function returns a pseudo user structure when it actually only returns the uid and gid, so if not needed I would revert it.

src/sdk/bucketspace_fs.js
```diff
@@ -648,7 +649,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
     // TODO: move the following 3 functions - has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
     // so they can be re-used
     async has_bucket_action_permission(bucket, account, action, bucket_path = "") {
-        const account_identifier = account.name.unwrap();
+        const account_identifier = account.name?.unwrap();
```
guymguym, line 652: I think now this is not strictly needed right? because the anonymous user will have an identifier.

config.js
```diff
@@ -824,6 +824,10 @@ config.NSFS_LOW_FREE_SPACE_MB_UNLEASH = 10 * 1024;
 // operations safely.
 config.NSFS_LOW_FREE_SPACE_PERCENT_UNLEASH = 0.10;
 
+// anonymous account name
+config.NC_ANON_ACCOUNT_NAME = 'anonymous'
```
guymguym, line 828:
```suggestion
config.ANONYMOUS_ACCOUNT_NAME = 'anonymous';
```

config.js
```diff
@@ -824,6 +824,10 @@ config.NSFS_LOW_FREE_SPACE_MB_UNLEASH = 10 * 1024;
...
+config.ANONYMOUS_ACCESS_KEY = Symbol('anonymous_access_key');
```
guymguym, line 829: As a config this can be overridden, which is not really meant to be used.
I would prefer to have this as a const in object_sdk, and export it to be shared for bucketspace_fs.

src/sdk/object_sdk.js
```diff
@@ -207,19 +207,16 @@ class ObjectSDK {
     async load_requesting_account(req) {
         try {
             const token = this.get_auth_token();
-            if (!token) return;
+            if (!token && !process.env.NC_NSFS_NO_DB_ENV) return;
```
guymguym, line 210: How about we add a bucketspace method that will decide if to allow anonymous requests?
I think this will encapsulate this check for env NC_NSFS_NO_DB_ENV which is quite weird in this context.

src/sdk/object_sdk.js
```diff
@@ -245,18 +255,19 @@ class ObjectSDK {
         }
         // check for a specific bucket
         if (bucket && req.op_name !== 'put_bucket') {
-            // ANONYMOUS: cannot work without bucket, cannot work on namespace bucket (?)
+            // ANONYMOUS: cannot work without bucket.
+            // Return if the acount is anonymous
+            if (!token && process.env.NC_NSFS_NO_DB_ENV) return;
```
guymguym, line 260: Same as above - How about we add a bucketspace method that will decide if to allow anonymous requests?

src/sdk/object_sdk.js
```diff
@@ -1107,8 +1109,20 @@ class ObjectSDK {
...
-module.exports = ObjectSDK;
-module.exports.account_cache = account_cache;
-module.exports.dn_cache = dn_cache;
+// EXPORT
+module.exports = {
+    ObjectSDK,
+    anonymous_access_key: anonymous_access_key,
+    account_cache: account_cache,
+    dn_cache: dn_cache,
+};
```
guymguym, line 1128: Prefer to use this form to exports multiple things:
```
exports.ObjectSDK = ObjectSDK;
exports.anonymous_access_key = anonymous_access_key;
```
The rational is that replacing the exports object is not safe for callers that use const object_sdk = require('object_sdk') and then later on try to use object_sdk.ObjectSDK as the original exports object is replaced and not used. This is a bit anoying but is more robust.

config.js
```diff
@@ -824,6 +824,9 @@ config.NSFS_LOW_FREE_SPACE_MB_UNLEASH = 10 * 1024;
 // operations safely.
 config.NSFS_LOW_FREE_SPACE_PERCENT_UNLEASH = 0.10;
 
+// anonymous account name
+config.NC_ANON_ACCOUNT_NAME = 'anonymous';
```
guymguym, line 828:
```suggestion
config. ANONYMOUS_ACCOUNT_NAME = 'anonymous';
```

src/sdk/object_sdk.js
```diff
@@ -1107,8 +1109,20 @@ class ObjectSDK {
...
         return ns.put_object_acl(params, this);
     }
+
+    is_nsfs_containerized_user_anonymous(token) {
+        return !token && !process.env.NC_NSFS_NO_DB_ENV;
+    }
+
+    is_nsfs_non_containerized_user_anonymous(token) {
+        return !token && process.env.NC_NSFS_NO_DB_ENV;
+    }
```
guymguym, line 1119: I meant that these will be calls to bucketspace, so that in the bucketspace_fs level this is clearly going to allowed, but in bucketspace_nb it will not

guymguym (approved):
just few comments left, but all in all this is looks great to me.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -53,6 +54,7 @@ async function validate_input_types(type, action, argv) {
         validate_no_extra_options(type, action, input_options_from_file, true);
         validate_options_type_by_value(input_options_with_data_from_file);
         validate_flags_combination(type, action, input_options_from_file);
+        validate_flags_value_combination(type, action, input_options_with_data);
```
shirady, line 57: It is fine not to support the anonymous creation with `from_file` - just better to test that behavior (you can do it manually or in the unit tests).

docs/non_containerized_NSFS.md
```diff
@@ -642,3 +642,60 @@ Access is restricted to these whitelist IPs. If there are no IPs mentioned all I
 ```
 sudo noobaa-cli whitelist --ips '["127.0.0.1", "192.000.10.000", "3002:0bd6:0000:0000:0000:ee00:0033:000"]'  2>/dev/null
 ```
+
+## Anonymous request
+
+Anonymous requests are those sent without access and secret key and Noobaa NSFS allow it with supporting bucket policy. Users have to create an anonymous account before accessing the resources. Noobaa uses anonymous account's uid and gid or user(distinguished_name) for accessing bucket storage paths. Noobaa can have one anonymous account. Commands for creating anonumous account are
```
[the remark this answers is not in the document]
naveenpaul1: that should depend on the uid/gid or user set in anonymous account. cant give  specific one

naveenpaul1:
Regarding ceph test `test_post_object_wrong_bucket`
This test was working before because request was considered anonymous and anonymous request was returning 403 error before.
Ceph is passing `AWSAccessKeyId` and `signature` to this request and this request should not be anonymous But `make_auth_token_from_request()` have a [check](https://github.com/noobaa/noobaa-core/blob/master/src/util/signature_utils.js#L269) `if (req.query.AWSAccessKeyId && req.query.Signature)` where `s` in signature is capital and this condition is failing.
@shirady

Issue #7993, opened 2024-04-23
logrotate is not performing for noobaa logs if it is inside the noobaa dir under logrotate.d directory


rkomandu:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

noobaa-4.15.0 
Standalone. 

### Actual behavior
1.

logrotate should perform the compression once the size is reached

```
[root@rkomandu-ip-cls-x-worker2 ~]# ls -lrh /var/log/noobaa*
-rw-r--r-- 1 root root 1.4G Apr 23 00:40 /var/log/noobaa.log
-rw-r--r-- 1 root root  19M Apr 22 16:30 /var/log/noobaa_events.log
[root@rkomandu-ip-cls-x-worker2 ~]# logrotate /etc/logrotate.d/noobaa/logrotate_noobaa.conf
[root@rkomandu-ip-cls-x-worker2 ~]# ls -lrh /var/log/noobaa*
-rw-r--r-- 1 root root  74M Apr 23 00:40 /var/log/noobaa.log.1.gz
-rw-r--r-- 1 root root 1.3K Apr 23 00:41 /var/log/noobaa.log
-rw-r--r-- 1 root root  19M Apr 22 16:30 /var/log/noobaa_events.log.1
-rw-r--r-- 1 root root    0 Apr 23 00:41 /var/log/noobaa_events.log
[root@rkomandu-ip-cls-x-worker2 ~]# cd /var/log/noobaa.log.1.gz /tmp
-bash: cd: too many arguments
[root@rkomandu-ip-cls-x-worker2 ~]# cp /var/log/noobaa.log.1.gz /tmp
[root@rkomandu-ip-cls-x-worker2 ~]# cd /tmp

[root@rkomandu-ip-cls-x-worker2 tmp]# file noobaa.log.1.gz
noobaa.log.1.gz: gzip compressed data, last modified: Tue Apr 23 07:40:50 2024, from Unix, original size modulo 2^32 1428749230

[root@rkomandu-ip-cls-x-worker2 tmp]# gunzip noobaa.log.1.gz

[root@rkomandu-ip-cls-x-worker2 tmp]# ls -lh noobaa.log.1
-rw-r--r-- 1 root root 1.4G Apr 23 00:41 noobaa.log.1


```
@naveenpaul1 and I did this over the webex 

Note: We had an issue opened in RTC 38551, 

[...]

PR #7998, opened 2024-04-25
NSFS | Logrotate is not performing for noobaa logs

naveenpaul1:
### Explain the changes
1. add script to run logrotate in regular interval

1. https://github.com/noobaa/noobaa-core/issues/7993

### Testing Instructions:
1. Verify the noobaa log is getting zipped once log size is more than configure limit.

- [X] Doc added/updated
- [ ] Tests added

### DIFF
--- a/docs/non_containerized_NSFS.md
+++ b/docs/non_containerized_NSFS.md
@@ -612,7 +612,9 @@ Noobaa logs are pushed to `var/log/noobaa.log` and the log is rotated and compre
 
 Verify the rsyslog and logrotate rpm configuration is complete by checking the files `etc/rsyslog.d/noobaa_syslog.conf` for rsyslog and `etc/logrotate.d/noobaa/logrotate_noobaa.conf` for logrotate.These files contain the noobaa specific configuration for rsyslog and logrotate.
 
-Rotate the logs manually.
+Logrotate configuration is set up under `/etc/logrotate.d/noobaa/`. In order to trigger rotation when log files reache size threshold, each log file (`/var/log/noobaa.log` and `/var/log/noobaa_events.log`) uses its own rsyslog `outchannel` that triggers logrotate once file size reaches a limit (105MB). Logrotate is also typically called from cron on daily schedule.
+
+To rotate the logs manually run.
 
 ```
 logrotate /etc/logrotate.d/noobaa/logrotate_noobaa.conf
--- a/src/deploy/RPM_build/noobaa.spec
+++ b/src/deploy/RPM_build/noobaa.spec
@@ -89,6 +89,7 @@ mv $RPM_BUILD_ROOT/usr/local/noobaa-core/src/deploy/standalone/noobaa_syslog.con
 
 mkdir -p $RPM_BUILD_ROOT/etc/logrotate.d/noobaa
 ln -s /usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.conf $RPM_BUILD_ROOT/etc/logrotate.d/noobaa/logrotate_noobaa.conf
+chmod +x $RPM_BUILD_ROOT/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh
 
 %files
 /usr/local/noobaa-core
--- a/src/deploy/standalone/logrotate_noobaa.conf
+++ b/src/deploy/standalone/logrotate_noobaa.conf
@@ -1,7 +1,25 @@
 /var/log/noobaa.log
 {
     daily
-    size 100M
+    maxsize 100M
+    minsize 50M
+    start 1
+    missingok
+    rotate 100
+    compress
+    create 644 root root
+    sharedscripts
+    postrotate 
+        killall -HUP rsyslogd || true
+        killall -HUP syslogd || true
+    endscript
+}
+
+/var/log/noobaa_events.log
+{
+    daily
+    maxsize 100M
+    minsize 10M
     start 1
     missingok
     rotate 100
@@ -9,14 +27,13 @@
     create 644 root root
     sharedscripts
     postrotate
-        /bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
-        /bin/kill -HUP `cat /var/run/rsyslogd.pid 2> /dev/null` 2> /dev/null || true
+        killall -HUP rsyslogd || true
+        killall -HUP syslogd || true
     endscript
 }
 
 /var/log/client_noobaa.log
 {
-    daily
     size 100M
     start 1
     missingok
@@ -25,22 +42,7 @@
     create 644 root root
     sharedscripts
     postrotate
-        /bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
-        /bin/kill -HUP `cat /var/run/rsyslogd.pid 2> /dev/null` 2> /dev/null || true
-    endscript
-}
-
-/var/log/noobaa_events.log
-{
-    monthly
-    size 10M
-    start 1
-    missingok
-    rotate 10
-    create 644 root root
-    sharedscripts
-    postrotate
-        /bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
-        /bin/kill -HUP `cat /var/run/rsyslogd.pid 2> /dev/null` 2> /dev/null || true
+        killall -HUP rsyslogd || true
+        killall -HUP syslogd || true
     endscript
 }
--- a/src/deploy/standalone/logrotate_noobaa.sh
+++ b/src/deploy/standalone/logrotate_noobaa.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+/usr/sbin/logrotate -v /etc/logrotate.d/noobaa/logrotate_noobaa.conf || true
--- a/src/deploy/standalone/noobaa_syslog.conf
+++ b/src/deploy/standalone/noobaa_syslog.conf
@@ -28,14 +28,17 @@ template(name="LogseneFormat" type="list" option.json="on") {
   constant(value="\"}\n")
 }
 
-# For servers
-local0.*        /var/log/noobaa.log;RSYSLOG_FileFormat
+# 105M limit(110100480), 5MB bigger than logrotate value. When updating this value update logrotate also.
+$outchannel log_rotation,/var/log/noobaa.log,110100480,/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh
+local0.* :omfile:$log_rotation
+&stop
+
+# For events
+$outchannel event_rotation,/var/log/noobaa_events.log,110100480,/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh
+local2.* :omfile:$event_rotation;LogseneFormat
 &stop
 
 # For clients
 local1.*        /var/log/client_noobaa.log;RSYSLOG_FileFormat
 &stop
 
-# For events
-local2.*        /var/log/noobaa_events.log;LogseneFormat
-&stop
--- a/src/deploy/RPM_build/RPM.Dockerfile
+++ b/src/deploy/RPM_build/RPM.Dockerfile
@@ -36,6 +36,7 @@ COPY ./binding.gyp .
 COPY ./src/deploy/standalone/noobaa_rsyslog.conf ./src/deploy/standalone/noobaa_rsyslog.conf
 COPY ./src/deploy/standalone/noobaa_syslog.conf ./src/deploy/standalone/noobaa_syslog.conf
 COPY ./src/deploy/standalone/logrotate_noobaa.conf ./src/deploy/standalone/logrotate_noobaa.conf
+COPY ./src/deploy/standalone/logrotate_noobaa.sh ./src/deploy/standalone/logrotate_noobaa.sh
 COPY ./src/manage_nsfs ./src/manage_nsfs
 
 WORKDIR /build


src/deploy/standalone/logrotate_noobaa.conf
```diff
@@ -9,8 +9,7 @@
     create 644 root root
     sharedscripts
     postrotate
-        /bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
-        /bin/kill -HUP `cat /var/run/rsyslogd.pid 2> /dev/null` 2> /dev/null || true
+        kill -HUP $(pidof rsyslogd) >/dev/null 2>&1 || true
```
guymguym, line 12: 1. Do you know why is it needed to redirect the output of the kill command to /dev/null ? what happens if we just leave the output as is?

2. Redhat article suggests for RHEL8  and above to use systemd to send this signal to the rsyslog.service as described here - https://access.redhat.com/solutions/1294. However we might be a bit more portable by using the suggested killall option as described in this rsyslog issue - https://github.com/rsyslog/rsyslog/issues/3952

```
postrotate 
    killall -HUP rsyslogd || true
endscript
```

src/deploy/RPM_build/noobaa.spec
```diff
@@ -101,6 +101,11 @@ if [ "${state}" == "active" ]; then
   service rsyslog restart
 fi
 
+while true; do
+    /usr/sbin/logrotate /etc/logrotate.d/noobaa 2>&1
+    sleep 30
+done &
+
```
guymguym, line 108: This is not a good place to set up long running tasks, it will not persist after reboot or failures.

The recommended way to change logrotate schedule is to change it's system wide cron schedule which is daily by default using something like - 
```
sudo mv /etc/cron.daily/logrotate /etc/cron.hourly
```
(see https://stackoverflow.com/questions/25485047/hourly-rotation-of-files-using-logrotate and https://betterstack.com/community/guides/logging/how-to-manage-log-files-with-logrotate-on-ubuntu-20-04/#changing-the-system-logrotate-schedule)

But this is not something that makes sense to add to our RPM and should be handled by the system admin.

For our use case, what we want is to trigger rotation by size, which could be done using rsyslog outchannel trigger that will invoke logrotate - see reference https://www.rsyslog.com/doc/tutorials/log_rotation_fix_size.html which would be added to rsyslog config like:
```
$outchannel noobaa_log,/var/log/noobaa.log, 52428800,/path/to/logrotate_noobaa
```
where logrotate_noobaa script calls `/usr/sbin/logrotate /etc/logrotate.d/noobaa` ...

guymguym (changes requested):

docs/non_containerized_NSFS.md
```diff
@@ -612,7 +612,9 @@ Noobaa logs are pushed to `var/log/noobaa.log` and the log is rotated and compre
 
 Verify the rsyslog and logrotate rpm configuration is complete by checking the files `etc/rsyslog.d/noobaa_syslog.conf` for rsyslog and `etc/logrotate.d/noobaa/logrotate_noobaa.conf` for logrotate.These files contain the noobaa specific configuration for rsyslog and logrotate.
 
-Rotate the logs manually.
+Logrotate is configured in rsyslog `outchannel`. And script in `logrotate_noobaa.sh` will get executed once `var/log/noobaa.log` size reach 100MB. Both `outchannel` and `size` configured in logrotate `/var/log/noobaa.log` should be same.
+
```
aspandey, line 616: Size for outchannel and the size configured in logrotate configuration file
for /var/log/noobaa.log should be same.

src/deploy/standalone/logrotate_noobaa.conf
```diff
@@ -40,7 +38,6 @@
...
     endscript
 }
```
aspandey, line 43: @naveenpaul1 Its not your change but I think you should modify the "size" and "daily" options also.

Remove "daily" and keep "size" option only. As per man page of logrotate , which I have also tested, "daily|hourly|monthly" will not matter if you have provided "size" option. Specially in your case, you should anyway only use rotation based on "size" only.

src/deploy/standalone/logrotate_noobaa.conf
naveenpaul1: @guymguym code you suggested is is working.

src/deploy/standalone/logrotate_noobaa.conf
naveenpaul1: @aspandey I think rotating daily basis will organise the logs more readable format.

docs/non_containerized_NSFS.md
naveenpaul1: @aspandey I didnt get you

src/deploy/RPM_build/noobaa.spec
naveenpaul1: code updated

docs/non_containerized_NSFS.md
aspandey: I just wanted to rephrased following line 
"Both `outchannel` and `size` configured in logrotate `/var/log/noobaa.log` should be same."

src/deploy/standalone/logrotate_noobaa.conf
aspandey: I think decision taken based on  "size" would be better. What I wanted to say in previous comment is that, "size"  option will override "daily"  and only "size" will be applicable so why to mention "daily" at all.

src/deploy/standalone/logrotate_noobaa.conf
```diff
@@ -1,46 +1,41 @@
...
     rotate 10
+    compress
     create 644 root root
     sharedscripts
     postrotate
-        /bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
-        /bin/kill -HUP `cat /var/run/rsyslogd.pid 2> /dev/null` 2> /dev/null || true
-    endscript
+        killall -HUP rsyslogd || true
+    endscriptt
```
aspandey, line 40: extra "t"

naveenpaul1: Thanks @aspandey

aspandey (approved):
LGTM

src/deploy/standalone/logrotate_noobaa.sh
```diff
@@ -0,0 +1 @@
+/usr/sbin/logrotate -v /etc/logrotate.d/noobaa/logrotate_noobaa.conf
```
guymguym, line 1: add hasbang so that you don't need to specify bash to call it:
```suggestion
#!/bin/bash
/usr/sbin/logrotate -v /etc/logrotate.d/noobaa/logrotate_noobaa.conf
```

src/deploy/standalone/noobaa_syslog.conf
```diff
@@ -4,6 +4,16 @@ $umask 0000
 $FileCreateMode 0644
 $EscapeControlCharactersOnReceive off
 
+$imjournalRatelimitInterval 0
+$imjournalRatelimitBurst 0
```
guymguym, line 8: why are these needed? can you add a comment to explain?

src/deploy/standalone/noobaa_syslog.conf
```diff
@@ -4,6 +4,16 @@ $umask 0000
...
+
+# start log rotation via outchannel
+# outchannel definition, 100M limit, If updating this value update logrotate also.
+$outchannel log_rotation,/var/log/noobaa.log,104857600,bash /etc/logrotate.d/noobaa/logrotate_noobaa.sh
+#  activate the channel and log everything to it
+*.* :omfile:$log_rotation
```
guymguym, line 14: this seems to apply the same outchannel to all interfaces - isn't that messing up the other 2 interfaces we have? should we define separate outchannels for them? or maybe we can just set this channel only to local0.* ?

src/deploy/standalone/logrotate_noobaa.conf
```diff
@@ -1,46 +1,41 @@
...
+        killall -HUP rsyslogd || true
     endscript
 }
 
-/var/log/client_noobaa.log
+/var/log/noobaa_events.log
 {
-    daily
-    size 100M
+    size 10M
```
guymguym, line 17: why do we need to have a different setting for noobaa_events?
others use size 100M - isn't it a good fit for this as well?

src/deploy/RPM_build/noobaa.spec
```diff
@@ -89,12 +89,14 @@ mv $RPM_BUILD_ROOT/usr/local/noobaa-core/src/deploy/standalone/noobaa_syslog.con
...
 mkdir -p $RPM_BUILD_ROOT/etc/logrotate.d/noobaa
 ln -s /usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.conf $RPM_BUILD_ROOT/etc/logrotate.d/noobaa/logrotate_noobaa.conf
+mv $RPM_BUILD_ROOT/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh $RPM_BUILD_ROOT/etc/logrotate.d/noobaa/logrotate_noobaa.sh
 
 %files
 /usr/local/noobaa-core
 %{_unitdir}/noobaa.service
 /etc/logrotate.d/noobaa/logrotate_noobaa.conf
 /etc/rsyslog.d/noobaa_syslog.conf
+/etc/logrotate.d/noobaa/logrotate_noobaa.sh
```
guymguym, line 99: is this a good place to put scripts? I am not so sure. Why not leave it in /usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh ?

src/deploy/standalone/noobaa_syslog.conf
naveenpaul1: Without these configuration rsyslog have a log rate limit, And because of this there is a chance that we might miss few logs with higher log levels. 
https://www.rsyslog.com/doc/configuration/modules/imjournal.html
_By default, ratelimiting is activated and permits to process 20,000 messages within 10 minutes,_
This is suggested in our doc also [Link](https://github.com/noobaa/noobaa-core/blob/0e9160dbf3c45e16a46dd3f3f8ff6aa0aa9e9d6b/docs/dev_guide/standalone_with_db.md#31-additional-syslog-configuration-for-rhel8)

src/deploy/standalone/logrotate_noobaa.conf
naveenpaul1: Number of Noobaa event logs created is very less compare to normal log files because the event log will have only one entry for each events. To avoid large file with very old events entries I decided to keep a small size for the event log file rotation.

src/deploy/standalone/noobaa_syslog.conf
naveenpaul1: Ya, will set only for the `local0.*`

src/deploy/RPM_build/noobaa.spec
naveenpaul1: sure we can keep it in `/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh `

src/deploy/standalone/logrotate_noobaa.sh
naveenpaul1: @guymguym  I tried with hasbang but script is not getting executed 
error
`program '/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh' (pid 518) exited with status 1 [v8.2310.0-4.el9]
May 06 12:51:14 d310af06fbec rsyslogd[512]: file size limit cmd for file '/var/log/noobaa.log' did no resolve situation  [v8.2310.0-4.el9] `

src/deploy/RPM_build/noobaa.spec
```diff
@@ -95,6 +95,7 @@ ln -s /usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.conf $RPM_BU
 %{_unitdir}/noobaa.service
 /etc/logrotate.d/noobaa/logrotate_noobaa.conf
 /etc/rsyslog.d/noobaa_syslog.conf
+/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh
```
guymguym, line 98: is it really needed to specify? all the files in /usr/local/noobaa-core are part of the RPM...

src/deploy/standalone/logrotate_noobaa.sh
guymguym: so add `|| true` at the end...

naveenpaul1: outchannel is giving issues when script is not executable

src/deploy/standalone/noobaa_syslog.conf
```diff
@@ -4,6 +4,17 @@ $umask 0000
...
 
+# Remove defaul ratelimiting, Setting this value to 0 turns off ratelimiting
+$imjournalRatelimitInterval 0
+$imjournalRatelimitBurst 0
+
+# start log rotation via outchannel
+# outchannel definition, 100M limit, If updating this value update logrotate also.
+$outchannel log_rotation,/var/log/noobaa.log,104857600,/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa
+#  activate the channel and log everything to it
+local0.* :omfile:$log_rotation
```
guymguym, line 15: If I understand correctly, having both this line and the line 43 would produce the messages twice in the file - can you check if that's the case? I expected the outchannel to replace line 43.

src/deploy/standalone/noobaa_syslog.conf
guymguym: thanks for the info, but I am still not sure I understand - 
how is systemd journal related here?
aren't we sending log messages directly to syslog??

src/deploy/standalone/logrotate_noobaa.conf
guymguym: but why did you remove the daily option then? that would have handled it

src/deploy/standalone/noobaa_syslog.conf
```diff
@@ -4,6 +4,17 @@ $umask 0000
 $FileCreateMode 0644
 $EscapeControlCharactersOnReceive off
 
+# Remove defaul ratelimiting, Setting this value to 0 turns off ratelimiting
+$imjournalRatelimitInterval 0
+$imjournalRatelimitBurst 0
+
+# start log rotation via outchannel
+# outchannel definition, 100M limit, If updating this value update logrotate also.
+$outchannel log_rotation,/var/log/noobaa.log,104857600,/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa
```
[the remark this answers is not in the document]
naveenpaul1: It will work without `.sh` also

src/deploy/standalone/logrotate_noobaa.conf
naveenpaul1: I am ok with daily option. Like @aspandey suggested, having a size based log rotation make more sense.

src/deploy/standalone/noobaa_syslog.conf
naveenpaul1: @guymguym I am not sure about that

src/deploy/standalone/noobaa_syslog.conf
naveenpaul1: You are right same logs are generated twice in log file. Will remove the line 43

src/deploy/standalone/logrotate_noobaa.conf
aspandey: @naveenpaul1 My point was not that size will make sense more than "daily" . 
I wanted to say is that if you mention both "daily" and after that "size", then only "size" will be applicable and logrotate will not consider "daily" option.

You can check it on your laptop. Just have this conf for some file and then run logrotate command. It will say something like - "daily" is discarded/skipped as size is mentiond.
 
If you really want to include both then you have to use "min-size", "max-size" with frequency option like "daily", "monthly" or "hourly"

I have sent the patch for my use case -
https://github.com/noobaa/noobaa-core/pull/8006

src/deploy/standalone/noobaa_syslog.conf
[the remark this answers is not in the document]
naveenpaul1: I reverted back to `bash /usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh` because rsyslog still giving same error even after adding `|| true`
```
May 07 05:33:44 d310af06fbec rsyslogd[1026]: program '/usr/local/noobaa-core/src/deploy/standalone/logrotate_noobaa.sh' (pid 1065) exited with status 1 [v8.2310.0-4.el9]
May 07 05:33:44 d310af06fbec rsyslogd[1026]: file size limit cmd for file '/var/log/noobaa.log' did no resolve situation  [v8.2310.0-4.el9]
```

src/deploy/standalone/logrotate_noobaa.conf
aspandey: Testing with following conf file - 
/var/log/bucket_logs.log
{
        hourly
        size 10k
        start 1
        missingok
        rotate 100
        create 666 noob root
        dateext
        dateformat %Y-%m-%d-%H-%M-%S
        sharedscripts
        olddir /var/log/noobaa_logs
        postrotate
                kill -HUP $(pidof rsyslogd) >/dev/null 2>&1 || true
        endscript
}

noob@noobaa-core-0 /]$     /usr/sbin/logrotate -v /etc/logrotate.d/noobaa/logrotate_noobaa.conf >/dev/stdout 2>&1
reading config file /etc/logrotate.d/noobaa/logrotate_noobaa.conf
**warning: 'size' overrides previously specified 'daily'
warning: 'size' overrides previously specified 'daily'**
olddir is now /var/log/noobaa_logs

src/deploy/standalone/noobaa_syslog.conf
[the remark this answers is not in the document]
guymguym: did you chmod +x that script? it doesn't make sense

src/deploy/standalone/noobaa_syslog.conf
guymguym: but this means you also need a separate channel for the noobaa_events too for the local2 facility

src/deploy/standalone/logrotate_noobaa.conf
guymguym: so why not use max-size?

src/deploy/standalone/logrotate_noobaa.sh
guymguym: chmod +x?
Also I think the channel size should also be a bit higher than the logrotate max-size to assure that logrotate will not skip it and cause rsyslog to stop writing.

src/deploy/standalone/logrotate_noobaa.conf
```diff
@@ -1,46 +1,41 @@
...
     start 1
     missingok
     rotate 10
+    compress
     create 644 root root
     sharedscripts
     postrotate
-        /bin/kill -HUP `cat /var/run/syslogd.pid 2> /dev/null` 2> /dev/null || true
-        /bin/kill -HUP `cat /var/run/rsyslogd.pid 2> /dev/null` 2> /dev/null || true
+        killall -HUP rsyslogd || true
```
guymguym, line 39: btw now that I look at this - is there a reason we removed sending the HUP signal to syslogd? did that cause any trouble or can we keep both? I think there are systems that still run syslog daemon as syslogd (eg [AIX](https://www.ibm.com/docs/en/aix/7.3?topic=s-syslogd-daemon))
```suggestion
        killall -HUP rsyslogd || true
        killall -HUP syslogd || true
```

src/deploy/standalone/noobaa_syslog.conf
guymguym: so why should we add this?

src/deploy/standalone/noobaa_syslog.conf
```diff
@@ -29,8 +40,8 @@ template(name="LogseneFormat" type="list" option.json="on") {
 }
 
 # For servers
-local0.*        /var/log/noobaa.log;RSYSLOG_FileFormat
-&stop
+# local0.*        /var/log/noobaa.log;RSYSLOG_FileFormat
```
guymguym, line 43: this file is getting full of commented out lines and things are unordered and messy. rsyslog.conf depends on order of statements, so I suggest that you move the outchannel definitions to this section where we actually define the log output for each facility rather than having it spread around the file.

src/deploy/standalone/noobaa_syslog.conf
naveenpaul1: No, Its happening for noobaa logs because we mention it in two places and its not the case for noobaa event logs

src/deploy/standalone/logrotate_noobaa.conf
naveenpaul1: from local testing what I can see is It didnt cause any issue new logs are getting added to newly created log files

src/deploy/standalone/noobaa_syslog.conf
naveenpaul1: Previously in logs I could see message like `some messages are skipped due to rate limit`, Thats why I added this code.

src/deploy/standalone/noobaa_syslog.conf
guymguym: but how would noobaa_events get rotated by size then?

src/deploy/standalone/noobaa_syslog.conf
guymguym: We need to understand this - perhaps this is because our process is sending log messages both to stdout and to syslog, and the systemd journal captures the stdout and then tries to use that also as input to rsyslog? are we even enabling imjournal?

I see from docs - https://www.rsyslog.com/doc/configuration/modules/imjournal.html
> Warning: Some versions of systemd journal have problems with database corruption, which leads to the journal to return the same data endlessly in a tight loop. This results in massive message duplication inside rsyslog probably resulting in a denial-of-service when the system resources get exhausted. This can be somewhat mitigated by using proper rate-limiters, but even then there are spikes of old data which are endlessly repeated. By default, ratelimiting is activated and permits to process 20,000 messages within 10 minutes, what should be well enough for most use cases. If insufficient, use the parameters described below to adjust the permitted volume. It is strongly recommended to use this plugin only if there is hard need to do so.

which makes me think that using this input is not something we want at all.
maybe we need to disable injournal?

src/deploy/standalone/noobaa_syslog.conf
guymguym: can you figure out why?

docs/non_containerized_NSFS.md
```diff
@@ -612,7 +612,9 @@ Noobaa logs are pushed to `var/log/noobaa.log` and the log is rotated and compre
 
 Verify the rsyslog and logrotate rpm configuration is complete by checking the files `etc/rsyslog.d/noobaa_syslog.conf` for rsyslog and `etc/logrotate.d/noobaa/logrotate_noobaa.conf` for logrotate.These files contain the noobaa specific configuration for rsyslog and logrotate.
 
-Rotate the logs manually.
+Logrotate is configured in rsyslog `outchannel` and both `var/log/noobaa.log` and `var/log/noobaa.log` have different `outchannel`. Script in `logrotate_noobaa.sh` will get executed once `var/log/noobaa.log` and `var/log/noobaa.log` size reach 105MB limit.
```
guymguym, line 615: minor fix -
```suggestion
Logrotate configuration is set up under `/etc/logrotate.d/noobaa/`. In order to trigger rotation when log files reache size threshold, each log file (`/var/log/noobaa.log` and `/var/log/noobaa_events.log`) uses its own rsyslog `outchannel` that triggers logrotate once file size reaches a limit (105MB). Logrotate is also typically called from cron on daily schedule.
```

guymguym (approved):
Minor fix suggested to docs.
Looks good! 
Thanks for the hard work on this.

Remember to create an issue for technical gap to "NC | syslogs are sent twice" - Since debug_module sends every log message to two targets (but native code logs only to stderr at the moment) - 
1. syslog facility local0 - then [rsyslog omfile](https://www.rsyslog.com/doc/configuration/modules/omfile.html) copies it to `/var/log/noobaa.log`
2. stdout/stderr - then [systemd copies it into its journal](https://www.freedesktop.org/software/systemd/man/latest/systemd-journald.service.html#) and then [rsyslog imjournal](https://www.rsyslog.com/doc/configuration/modules/imjournal.html) copies it to the file `/var/log/messages`.

PR #8005, opened 2024-05-02
Fix ENOTSUP by falling back to meta files

### LINKED ISSUE #7843: ENOTSUP when using noobaa pv-pool with AWS EFS storage class due to xattr not supported
### Environment info
- NooBaa Version: master
- Platform: OpenShift AWS

### Actual behavior
1. when we upload a file to a pv-pool we currently save the metadata as xattr attached to the file this can be problematic when using storage classes that don't support xattr like AWS EFS

### Expected behavior
1. we should understand that xattr are not supported by the storage and failback to using 1 file for data and a second for metadata as we did in older versions

### Steps to reproduce
1. install ocp in AWS
2.  install noobaa with storage class set to AWS EFS: efs-nfs-client and try to create a pv-pool backingstore
3. see that backingstore pod is constantly failing with ENOTSUP

### LINKED ISSUE #7932: BackingStore Pod throws test_store_perf error (Operation not supported)
### Environment info
- NooBaa Version: **5.15.0-9b709d9**
- Platform: **Kubernetes v1.28.7+6e2789b | OpenShift 4.15.3**

### Actual behavior
1. Create PV Pool BackingStore with StorageClass of an external CSI provider
2. BackingStore Pod successfully mounts PV and is up and running
3. BackingStore Pod successfully writes directory structure to mounted PVC (/noobaa_storage)
4. BackingStore is in BackingStorePhaseRejected state due to reason: 'Backing store mode: NO_CAPACITY'

### Expected behavior
1. Create PV Pool BackingStore with StorageClass of an external CSI provider
2. BackingStore Pod successfully mounts PV and is up and running
3. BackingStore Pod successfully writes directory structure to mounted PVC (/noobaa_storage)
4. BackingStore is in BackingStorePhaseReady state

### Steps to reproduce
1. Install external CSI provider (Netapp Trident / Dell CSI Powerscale) that utilize NFS for mounting of external volumes
2. Create PV Pool BackingStore with StorageClass of above mentioned CSI provider
3. BackingStore is in BackingStorePhaseRejected state due to reason: 'Backing store mode: NO_CAPACITY'
   1. The BackingStore is freshly initilized with 50GB and has enough space.
    ```
    > df -h
    Filesystem                                                     Size  Used Avail Use% Mounted on
    ...
    10.3.3.104:/phch_ocp_pvc_594fdfdb_0ab3_4292_9df6_813c9a4d3fe7   50G   23M   50G   1% /noobaa_storage
    ...
    ```

#### What is working
1. I tried to use the Local Storage Operator with it's StorageClass for provisioning the backing store, which works out.
2. This Operator utilizes local disks instead of NFS shares mounted on the corresponding node where the backing store runs.  

#### Log Analysis

Storage is correctly recognized with enough space and is also communicated upwards
```
Mar-29 9:11:02.156 [Agent/20]    [L0] core.agent.agent_cli:: Loading agents [ { mount: '/noobaa_storage', drive_id: '10.3.3.104:/phch_ocp_pvc_594fdfdb_0ab3_4292_9df6_813c9a4d3fe7', storage: { total: 53687091200, free: 53666185216 } } ]
``` 

The following problematic behaviour repeats constantly, compared to the working BackingStore that utilizes the local storage operator. 
``` 
Mar-29 9:12:10.564 [Agent/20]    [L0] Agent.netapp-wait-noobaa-pod-c2d90df8-noobaa_storage-907cadfb:: get_node_ip: using fallback node ip 10.128.2.53
Mar-29 9:12:10.564 [Agent/20]    [L0] Agent.netapp-wait-noobaa-pod-c2d90df8-noobaa_storage-907cadfb:: Recieved potential servers list [ { address: 'wss://localhost:8443' } ]
Mar-29 9:12:11.457 [Agent/20]    [L0] Agent.netapp-wait-noobaa-pod-c2d90df8-noobaa_storage-907cadfb:: storage_info: { total: 53687091200, free: 53666119680, used: 0 }
Mar-29 9:12:11.461 [Agent/20]    [L0] Agent.netapp-wait-noobaa-pod-c2d90df8-noobaa_storage-907cadfb:: used storage is under 200 KB (0), treat as 0
Mar-29 9:12:11.857 [Agent/20]    [L0] Agent.netapp-wait-noobaa-pod-c2d90df8-noobaa_storage-907cadfb:: DRIVES: [ { mount:  '/' , drive_id:  'overlay' , storage: { total:  128243970048 , free:  96740057088  } }, { mount:  '' , drive_id:  'proc' , storage: { total:  0 , free:  0  } }, ... { mount:  '/noobaa_storage' , drive_id:  '10.3.3.104:/phch_ocp_pvc_594fdfdb_0ab3_4292_9df6_813c9a4d3fe7' , storage: { total:  53687091200 , free:  53666185216 , used:  0  } }, ... ]
Mar-29 9:12:11.863 [Agent/20]    [L0] Agent.netapp-wait-noobaa-pod-c2d90df8-noobaa_storage-907cadfb:: got _enable_service on storage agent
Mar-29 9:12:11.957 [Agent/20]  [WARN] core.agent.block_store_services.block_store_base:: encountered unknown error in test_store_perf [Error: Operation not supported] { code: 'ENOTSUP' }
Mar-29 9:12:11.958 [Agent/20] [ERROR] CONSOLE:: RPC._on_request: ERROR srv agent_api.test_store_perf reqid 8@ws://[::ffff:10.128.2.53]:45650/(4exhokcbf.zv) connid wss://noobaa-mgmt.openshift-storage.svc.cluster.local/(1g3katq4.01rd) Error: Operation not supported
Mar-29 9:12:12.145 [Agent/20]    [L0] core.rpc.rpc_base_conn:: RPC CONNECTION CLOSED. got event from connection: n2n://nodes_monitor(1y28j8wd.063) N2N ICE CLOSED
Mar-29 9:12:12.149 [Agent/20]    [L0] core.rpc.rpc_n2n:: _close undefined
Mar-29 9:12:12.149 [Agent/20]    [L0] core.rpc.ice:: TLS CLOSED: tcp://[10.128.2.53]:60383=>tcp://[::ffff:10.129.2.118]:47510
```

Unfortunately I could not find any details or bugs on test_store_perf errors. Would be great if someone could help me out here.

#### BackingStore PV data

BackingStore Pod seems to have already written it's correct directory structure to the PV:
```
ls -l /noobaa_storage/
total 12
-rw-r--r--. 1 noob 99  677 Mar 29 09:11 agent_conf.json
drwx------. 3 noob 99 4096 Mar 29 08:48 netapp-wait-noobaa-pod-c2d90df8-noobaa_storage-907cadfb
```

``` 
ls -l /noobaa_storage/netapp-wait-noobaa-pod-c2d90df8-noobaa_storage-907cadfb/
total 320
drwxr-xr-x. 4099 noob 99 319488 Mar 29 08:47 blocks_tree
-rw-r--r--.    1 noob 99    273 Mar 29 08:48 token
-rw-r--r--.    1 noob 99     20 Mar 29 08:47 usage
``` 

``` 
ls -l /noobaa_storage/netapp-wait-noobaa-pod-c2d90df8-noobaa_storage-907cadfb/blocks_tree/
total 16388
drwxr-xr-x. 2 noob 99 4096 Mar 29 08:47 000.blocks
drwxr-xr-x. 2 noob 99 4096 Mar 29 08:47 001.blocks
drwxr-xr-x. 2 noob 99 4096 Mar 29 08:47 002.blocks
drwxr-xr-x. 2 noob 99 4096 Mar 29 08:47 003.blocks
drwxr-xr-x. 2 noob 99 4096 Mar 29 08:47 004.blocks
...
drwxr-xr-x. 2 noob 99 4096 Mar 29 08:47 fff.blocks
drwxr-xr-x. 2 noob 99 4096 Mar 29 08:49 other.blocks
``` 

The only difference I can see is the lost+found directory is missing on the not working BackingStore pod.

#### Logs

As mentioned the same issue occours on 2 different environments with 2 different CSI provisioners that utilize NFS for mounting PVs. 

Attached you can find the [exported logs](https://github.com/noobaa/noobaa-core/files/14804267/noobaa_logs.zip) from Openshift
- netapp-wait-noobaa-pod        - not working BackingStore pod utilizing PVs from Dell / Netapp CSI Operator
- local-volume-test-noobaa-pod  - working BackingStore pod utilizing PVs from Local Storage Operator 
- noobaa-core-0
- noobaa-endpoint
- noobaa-operator

tangledbytes:
### Explain the changes
This PR re-adds the meta files to store metadata in the cases when extended attributes are not supported on the filesytem (`ENOTSUP`).

Fixes #7843
Fixes #7932

### Testing Instructions:
The way I tested it was -
1. Set `this.xattr_fallback = true`.
2. Upload an object
3. Check that both data as well as meta files are created.
4. Read the object.
5. Shutdown NooBaa and restart with `this.xattr_fallback = false`.
6. Read the object.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/agent/block_store_services/block_store_fs.js
+++ b/src/agent/block_store_services/block_store_fs.js
@@ -38,6 +38,10 @@ class BlockStoreFs extends BlockStoreBase {
         // stores the cached df data for the root path
         this.cached_df_data = null;
 
+        // If this is set to false then metadata files
+        // are used instead of xattrs
+        this.xattr_enabled = true;
+
         this.fs_context = {
             disable_ctime_check: config.BLOCK_STORE_FS_TMFS_ENABLED
         };
@@ -158,6 +162,18 @@ class BlockStoreFs extends BlockStoreBase {
                 }
             }
 
+            if (!this.xattr_enabled) {
+                const meta_path = this._get_block_meta_path(block_md.id);
+                const [block, meta] = await Promise.all([
+                    nb_native().fs.readFile(fs_context, block_path),
+                    nb_native().fs.readFile(fs_context, meta_path),
+                ]);
+
+                const block_md_from_fs = try_parse_block_md(meta.data.toString());
+
+                return { block_md: block_md_from_fs || block_md, data: block.data };
+            }
+
             const { data, stat } = await nb_native().fs.readFile(fs_context, block_path, { read_xattr: true });
 
             // read md from xattr
@@ -175,9 +191,15 @@ class BlockStoreFs extends BlockStoreBase {
             }
 
             return { block_md: block_md_from_fs || block_md, data };
-
         } catch (err) {
             if (err.rpc_code === 'MIGRATED') throw err; // Don't want to catch this error
+            if (err.code === 'ENOTSUP' && this.xattr_enabled) {
+                // Retry operation with xattr_fallback set to true
+                dbg.warn('detected ENOTSUP - turning off xattr_enabled and retrying');
+                this.xattr_enabled = false;
+
+                return this._read_block(block_md);
+            }
 
             this._test_root_path_exists(err);
         } finally {
@@ -204,9 +226,7 @@ class BlockStoreFs extends BlockStoreBase {
         // set the block md xattr
         const block_md_to_store = _.pick(block_md, 'id', 'digest_type', 'digest_b64', 'mapping_info');
         const block_md_data = JSON.stringify(block_md_to_store);
-
-        /** @type {nb.NativeFSXattr} */
-        const xattr = { [config.BLOCK_STORE_FS_XATTR_BLOCK_MD]: block_md_data };
+        const meta_path = this._get_block_meta_path(block_md.id);
 
         // a map of xattrs which will not fail the operation, but only warn
         /** @type {nb.NativeFSXattr} */
@@ -216,7 +236,6 @@ class BlockStoreFs extends BlockStoreBase {
         let xattr_need_fsync = false;
 
         if (!is_test_block) {
-
             // set xattr to trigger migration of file to underlying tier
             if (config.BLOCK_STORE_FS_TMFS_ENABLED) {
                 xattr_try = { [config.BLOCK_STORE_FS_XATTR_TRIGGER_MIGRATE]: 'now' };
@@ -230,7 +249,6 @@ class BlockStoreFs extends BlockStoreBase {
                 usage.count -= 1;
 
                 // also make sure we do not leave old .meta files on overwrite
-                const meta_path = this._get_block_meta_path(block_md.id);
                 const overwrite_meta_stat = await nb_native().fs.stat(fs_context, meta_path).catch(ignore_not_found);
                 if (overwrite_meta_stat) {
                     usage.size -= overwrite_meta_stat.size;
@@ -240,12 +258,30 @@ class BlockStoreFs extends BlockStoreBase {
         }
 
         try {
-            await nb_native().fs.writeFile(fs_context, block_path, data, {
-                xattr,
-                xattr_try,
-                xattr_need_fsync,
-            });
+            if (this.xattr_enabled) {
+                await nb_native().fs.writeFile(fs_context, block_path, data, {
+                    xattr: { [config.BLOCK_STORE_FS_XATTR_BLOCK_MD]: block_md_data },
+                    xattr_try,
+                    xattr_need_fsync,
+                });
+            } else {
+                await Promise.all([
+                    nb_native().fs.writeFile(fs_context, block_path, data),
+                    nb_native().fs.writeFile(fs_context, meta_path, Buffer.from(block_md_data, 'utf8')),
+                ]);
+
+                // MD file gets deleted above so add the md size back again to usage
+                usage.size += block_md_data.length;
+            }
         } catch (err) {
+            if (err.code === 'ENOTSUP' && this.xattr_enabled) {
+                // Retry operation with xattr_fallback set to true
+                dbg.warn('detected ENOTSUP - turning off xattr_enabled and retrying');
+                this.xattr_enabled = false;
+
+                return this._write_block(block_md, data, options);
+            }
+
             this._test_root_path_exists(err);
         }
 


romayalon:
@tangledbytes 
It should also fix https://github.com/noobaa/noobaa-core/issues/7932, right?

tangledbytes:
> @tangledbytes It should also fix #7932, right?

Yes! And in fact this is close replica of the issue we are seeing right now.

src/agent/block_store_services/block_store_fs.js
```diff
@@ -133,7 +137,7 @@ class BlockStoreFs extends BlockStoreBase {
         let block_file;
 
         try {
-            if (config.BLOCK_STORE_FS_TMFS_ENABLED) {
+            if (!this.xattr_fallback && config.BLOCK_STORE_FS_TMFS_ENABLED) {
```
guymguym, line 140: I would leave this out - we don't plan to support it without xattr at the moment, so it would be best to fail if configured with tmfs and xattr is not supported rather than just ignoring the config.
```suggestion
            if (config.BLOCK_STORE_FS_TMFS_ENABLED) {
```

src/agent/block_store_services/block_store_fs.js
```diff
@@ -158,24 +162,47 @@ class BlockStoreFs extends BlockStoreBase {
...
+            if (this.xattr_fallback) {
+                const meta_path = this._get_block_meta_path(block_md.id);
+                const [block, meta] = await Promise.all([
+                    nb_native().fs.readFile(fs_context, block_path),
+                    nb_native().fs.readFile(fs_context, meta_path),
+                ]);
+
+                data = block.data;
+                block_md_from_fs = try_parse_block_md(meta.data.toString());
+            } else {
```
guymguym, line 177: I would prefer that this if-block returns at the end without else - this way we don't need to share variables with the else block and it can stay const, and also we can keep the normal logic flat.

src/agent/block_store_services/block_store_fs.js
```diff
@@ -158,24 +162,47 @@ class BlockStoreFs extends BlockStoreBase {
...
+                        try {
+                            const meta_path = this._get_block_meta_path(block_md.id);
+                            const { data: meta_data } = await nb_native().fs.readFile(fs_context, meta_path);
+                            block_md_from_fs = try_parse_block_md(meta_data);
+                        } catch (err) {
+                            // noop
+                        }
+                    }
+                } catch (error) {
+                    if (error.code === 'ENOTSUP') {
```
guymguym, line 195: we should check that xattr_fallback is not already set so that we don't get into an infinite loop in case of any other kind of ENOTSUP case -

```suggestion
                    if (error.code === 'ENOTSUP' && !this.xattr_fallback) {
```

src/agent/block_store_services/block_store_fs.js
```diff
@@ -158,24 +162,47 @@ class BlockStoreFs extends BlockStoreBase {
...
+                    if (!block_md_from_fs) {
+                        try {
+                            const meta_path = this._get_block_meta_path(block_md.id);
+                            const { data: meta_data } = await nb_native().fs.readFile(fs_context, meta_path);
+                            block_md_from_fs = try_parse_block_md(meta_data);
+                        } catch (err) {
+                            // noop
+                        }
+                    }
+                } catch (error) {
```
guymguym, line 194: This catch swallows errors that it was not intended to.

I think it would be best to move this check to the outer try-catch, and then check if ENOTSUP and also !xattr_fallback and only then set the fallback and recurse just once.

src/agent/block_store_services/block_store_fs.js
```diff
@@ -38,6 +38,10 @@ class BlockStoreFs extends BlockStoreBase {
         // stores the cached df data for the root path
         this.cached_df_data = null;
 
+        // If this is set to true then metadata files
+        // are used instead of xattrs
+        this.xattr_fallback = false;
```
guymguym, line 43: I prefer this to be called xattr_enabled because this might be changed later to come from a configuration of the backing store for example, instead of only detected on runtime, and it makes more sense to me as a "positive" condition.

src/agent/block_store_services/block_store_fs.js
```diff
@@ -218,7 +242,7 @@ class BlockStoreFs extends BlockStoreBase {
         if (!is_test_block) {
 
             // set xattr to trigger migration of file to underlying tier
-            if (config.BLOCK_STORE_FS_TMFS_ENABLED) {
+            if (!this.xattr_fallback && config.BLOCK_STORE_FS_TMFS_ENABLED) {
```
guymguym, line 245: again, I would leave this out - we don't plan to support it without xattr at the moment, so it would be best to fail if configured with tmfs and xattr is not supported rather than just ignoring the config.

```suggestion
            if (config.BLOCK_STORE_FS_TMFS_ENABLED) {
```

src/agent/block_store_services/block_store_fs.js
```diff
@@ -230,22 +254,39 @@ class BlockStoreFs extends BlockStoreBase {
...
+                ]);
+            } else {
+                await nb_native().fs.writeFile(fs_context, block_path, data, {
+                    xattr: { [config.BLOCK_STORE_FS_XATTR_BLOCK_MD]: block_md_data },
+                    xattr_try,
+                    xattr_need_fsync,
+                });
+            }
         } catch (err) {
+            if (err.code === 'ENOTSUP') {
```
guymguym, line 282: again, we should check that xattr_fallback is not already set so that we don't get into an infinite loop in case of any other kind of ENOTSUP case -

```suggestion
            if (err.code === 'ENOTSUP' && !this.xattr_fallback) {
```

src/agent/block_store_services/block_store_fs.js
```diff
@@ -230,22 +254,39 @@ class BlockStoreFs extends BlockStoreBase {
                 usage.count -= 1;
 
                 // also make sure we do not leave old .meta files on overwrite
-                const meta_path = this._get_block_meta_path(block_md.id);
                 const overwrite_meta_stat = await nb_native().fs.stat(fs_context, meta_path).catch(ignore_not_found);
                 if (overwrite_meta_stat) {
                     usage.size -= overwrite_meta_stat.size;
-                    await nb_native().fs.unlink(fs_context, meta_path).catch(ignore_not_found);
+                    if (!this.xattr_fallback) {
```
guymguym, line 260: I think this check is undesired - we already found the meta file, so we should delete it unconditionally.

src/agent/block_store_services/block_store_fs.js
```diff
@@ -230,22 +254,39 @@ class BlockStoreFs extends BlockStoreBase {
...
         try {
-            await nb_native().fs.writeFile(fs_context, block_path, data, {
-                xattr,
-                xattr_try,
-                xattr_need_fsync,
-            });
+            if (this.xattr_fallback) {
+                await Promise.all([
+                    nb_native().fs.writeFile(fs_context, block_path, data),
+                    nb_native().fs.writeFile(fs_context, meta_path, Buffer.from(block_md_data, 'utf8')),
```
guymguym, line 272: I think we need to fix the usage.size for this case. see the code from before the change -
https://github.com/noobaa/noobaa-core/blob/e726c111d3389e8afaa350a8d4a4d79c85fa0bec/src/agent/block_store_services/block_store_fs.js#L150

src/agent/block_store_services/block_store_fs.js
```diff
@@ -328,6 +369,9 @@ class BlockStoreFs extends BlockStoreBase {
      */
     async _move_blocks_to_tmfs(block_ids) {
         const moved_block_ids = [];
+        if (this.xattr_fallback) {
```
guymguym, line 372: again, not sure we should change how this config behaves, I would leave this out.

guymguym (changes requested):

src/agent/block_store_services/block_store_fs.js
tangledbytes: That would be redundant check though, no? This catch will catch exceptions from the else branch will be chosen only when `this.xattr_fallback === false`. We exit the loop because in the recursive call `this.xattr_fallback = true` which means we won't enter this catch.

Am I missing something here?

src/agent/block_store_services/block_store_fs.js
tangledbytes: Ah, I see. I think the above suggestion was in conjunction with this suggestion. Now it makes sense.

guymguym (approved):

PR #8008, opened 2024-05-05
NC | CLI | Restrict Access Keys chars + performance improvement

### LINKED ISSUE #7931: SignatureDoesNotMatch for presigned URL when access key contains special character (`+`)
### Environment info
- NooBaa Version: **noobaa-core-5.16.0-20240310.el8.x86_64**
- Platform: **rhel-8**

### Actual behavior
1. Created an account, set the access key manually to `abc+abc`.
2. **NOTICE** - this happens when manually setting the access key, because [when we generate access keys we only pick alphanumeric chars](https://github.com/guymguym/noobaa-core/blob/dfd93ea1bcae0893700291d48e9fc489a4935c4a/src/util/cloud_utils.js#L191).
3. Used aws cli and cyberduck with that access key and secret without any issues (uses sigv4 in authentication header).
4. Tried to create a presigned URL with `aws s3 presign s3://bucket/key` but getting the URL failed with SignatureDoesNotMatch error.
5. Changed the access key to contain no `+`, and presign works too.
6. Tried also to get signed URL with cyberduck - the duck did not encode the + sign in the url and it messed up even before getting to the server.

### Expected behavior
1. Either block using access keys with invalid chars, or make it work ok also for presigned URL.

### Steps to reproduce
1. See above.

romayalon:
### Explain the changes
1. Change the access keys constraints per the following rules - 
- access_key -
   a. Length = 20.
   b. Contains only alphanumeric chars.
- secret_key -
  a. Length = 40.
  b. Contains only alphanumeric chars, "+", "/".

2. Performance improvements -
 Before the change, we tested 4 different regexp on the same access_key/secret_key, now we do it once.

1. Fixed #7931

### Testing Instructions:
1. `sudo jest --testRegex=jest_tests/test_nc_nsfs_account_cli`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/manage_nsfs/manage_nsfs_validations.js
+++ b/src/manage_nsfs/manage_nsfs_validations.js
@@ -316,23 +316,15 @@ function _validate_access_keys(access_key, secret_key) {
     if (!_.isUndefined(secret_key) && _.isUndefined(access_key)) {
         throw_cli_error(ManageCLIError.MissingAccountAccessKeyFlag);
     }
-    // checking the complexity of access_key
-    if (!_.isUndefined(access_key) && !string_utils.validate_complexity(access_key, {
-            require_length: 20,
-            check_uppercase: true,
-            check_lowercase: false,
-            check_numbers: true,
-            check_symbols: false,
-        })) throw_cli_error(ManageCLIError.AccountAccessKeyFlagComplexity);
-    // checking the complexity of secret_key
-    if (!_.isUndefined(secret_key) && !string_utils.validate_complexity(secret_key, {
-            require_length: 40,
-            check_uppercase: true,
-            check_lowercase: true,
-            check_numbers: true,
-            check_symbols: true,
-        })) throw_cli_error(ManageCLIError.AccountSecretKeyFlagComplexity);
 
+    // checking access_key length=20 and contains only alphanumeric chars
+    if (access_key && !string_utils.access_key_regexp.test(access_key)) {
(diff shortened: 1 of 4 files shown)


src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -16,6 +15,9 @@ const { throw_cli_error, get_config_file_path, get_bucket_owner_account,
 const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, FROM_FILE, BOOLEAN_STRING_VALUES,
     GLACIER_ACTIONS, LIST_UNSETABLE_OPTIONS } = require('../manage_nsfs/manage_nsfs_constants');
 
+const valid_access_key_regex = /^[a-zA-Z0-9]+$/;
+const valid_secret_key_regex = /^[a-zA-Z0-9+/]+$/;
```
guymguym, line 19: I would move these regex to be in a common module like string_utils where we have the other common regex defined.
Also it is possible to set the exact length in the regex so it doesn't require more checks.
```suggestion
const valid_access_key_regex = /^[a-zA-Z0-9]{20}$/;
const valid_secret_key_regex = /^[a-zA-Z0-9+/]{40}$/;
```

guymguym (changes requested):

src/manage_nsfs/manage_nsfs_validations.js
romayalon: nice! will do

guymguym (approved):

PR #8009, opened 2024-05-05
NSFS | NC | IAM Service - Phase 1 (dummy impls)

shirady:
### Explain the changes
1. Add the boilerplate implementation of the IAM service based on the S3 and STS services that were already implemented.
2. Add basic tests (only test the return structure).

### Issues: 
Gaps:
1.  Check the error by AWS when the request doesn't contain the header 'application/x-www-form-urlencoded'.
2. Move the requesting account to be called from the accountSpace (currently it is copied).
3. Refactor endpoint.js and nsfs.js so the both will reuse the config.js values (and not process.env in some cases).
4. Consolidate the calls of the functions `http_utils.authorize_session_token(req, headers_options);` and `authenticate_request(req);` in one place and reuse in S3, STS and IAM.
5. In `nb.d` define the params in every action (even as an object inline).
6. Implement the simple flow in account SDK.
7. Parsing and validating the params (for example: `username`).
8. Remove unused errors that were copied from STS errors.

#### Unit Tests
Those tests are basic and test the return structure of the API.
`sudo npx jest test_accountspace_fs.test.js` (without sudo will also work, but later tests in future PR will need the root permissions).

#### Manual Tests
These tests will only test the flow of the service (without any changes in the config files).
9. Start the NSFS server with: `sudo node src/cmd/nsfs --debug 5 --https_port_iam 7005`
10. Create the alias for IAM service: `alias s3-nc-user-1-iam='AWS_ACCESS_KEY_ID=<acess-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:7005'`.
11. Use AWS CLI to send requests to the IAM service, for example: `s3-nc-user-1-iam iam create-user --user-name Bob --path '/division_abc/subdivision_xyz/'` (more examples in the comment below)

Note: For checking the error parsing: comment out the `return` part in `create_user` and add `throw new IamError(IamError.AccessDeniedException);`.

- [X] Doc added/updated
- [X] Tests added


shirady:
#### Manual Tests examples:
Note: assuming that you already created the alias `s3-nc-user-1-iam` as mentioned in the PR description in the Testing Instructions part.

**user (CRUD)**
- `s3-nc-user-1-iam iam create-user --user-name Bob --path '/division_abc/subdivision_xyz/'`
  - `s3-nc-user-1-iam iam get-user` (without `--user-name` flag)
- `s3-nc-user-1-iam iam get-user --user-name Bob`
- `s3-nc-user-1-iam iam update-user --user-name Bob --new-user-name Robert --new-path '/division_abc/subdivision_abc/'`
- `s3-nc-user-1-iam iam delete-user --user-name Bob`
- `s3-nc-user-1-iam iam list-users`

**access keys (CRUD)**
- `s3-nc-user-1-iam iam create-access-key --user-name Bob`
- `s3-nc-user-1-iam iam get-access-key-last-used --access-key-id <access-key>` (currently of the root account)
- `s3-nc-user-1-iam iam update-access-key --access-key-id <access-key> --user-name Bob --status Inactive`
- `s3-nc-user-1-iam iam delete-access-key --access-key-id <access-key> --user-name Bob`
- `s3-nc-user-1-iam iam list-access-keys --user-name Bob`

src/endpoint/iam/ops/iam_get_user.js
```diff
@@ -0,0 +1,47 @@
...
+    dbg.log1('IAM GET USER', params);
+    const reply = await req.account_sdk.get_user(params);
+    dbg.log2('get_user reply', reply);
+
+    return {
+        GetUserResponse: {
+            GetUserResult: {
+                User: {
+                    UserId: reply.user_id,
+                    Path: reply.path ?? iam_utils.AWS_EMPTY_PATH,
```
guymguym, line 25: are you sure that `??` should be used here and not `||` ?
in other words - if `reply.path === ''` should we return `''` or `AWS_EMPTY_PATH`?

src/endpoint/iam/iam_utils.js
```diff
@@ -0,0 +1,41 @@
...
+const AWS_EMPTY_PATH = '/';
+const AWS_NOT_USED = 'N/A'; // can be used in case the region or the service name were not used
+
+/**
+ * format_iam_xml_date return the date without milliseconds
+ * @param {any} input
+ */
+function format_iam_xml_date(input) {
+    const date_iso = s3_utils.format_s3_xml_date(input);
+    const date_iso_no_zeros = date_iso.replace(/\.\d+/, ""); // remove the milliseconds zeros:
```
guymguym, line 15:
```suggestion
    const date_iso_no_millis = date_iso.replace(/\.\d+/, ""); // remove the milliseconds part
```

src/util/http_utils.js
```diff
@@ -23,6 +23,7 @@ const cloud_utils = require('./cloud_utils');
 
 const UNSIGNED_PAYLOAD = 'UNSIGNED-PAYLOAD';
 const STREAMING_PAYLOAD = 'STREAMING-AWS4-HMAC-SHA256-PAYLOAD';
+const FORM_URL_ENCODED = 'application/x-www-form-urlencoded';
```
guymguym, line 26: This eventually should be a list of constants for the content types we commonly use in our code base, see suggestion.
Also notice there's another occurance of 'x-www-form-urlencoded' in this file in function `parse_request_body()`, better replace it with this constant too.

```suggestion
const CONTENT_TYPE_TEXT_PLAIN = 'text/plain';
const CONTENT_TYPE_APP_OCTET_STREAM = 'application/octet-stream';
const CONTENT_TYPE_APP_JSON = 'application/json';
const CONTENT_TYPE_APP_XML = 'application/xml';
const CONTENT_TYPE_APP_FORM_URLENCODED = 'application/x-www-form-urlencoded';
```

guymguym (approved):
approving with few small comments.
also I think this feature deserves its own design doc in docs/design/iam.md
we don't need all the details in that doc, but we should start pouring content into it to describe our rational, implementation, limitations, roadmap. I would already create it with this PR, even with a single paragraph if no more time.

src/endpoint/iam/ops/iam_get_user.js
shirady: I will change the name of `AWS_EMPTY_PATH` to `AWS_DEFAULT_PATH` so it will be clearer.
The suggested path `''` is illegal, the path should either start with `/` or start and end with `/` (see [here](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateUser.html)):

> a string of characters consisting of either a forward slash (/) by itself or a string that must begin and end with forward slashes

src/endpoint/iam/iam_utils.js
shirady: Thanks

src/util/http_utils.js
shirady: I added the constants, I will reuse the new `CONTENT_TYPE_APP_FORM_URLENCODED` in other places in the code (that are not IAM) in the PR where I'm going to refactor S3 and STS.
In this PR I'm trying to keep the changes only to IAM.

src/endpoint/iam/ops/iam_get_user.js
guymguym: so should you also use || instead of ?? so that it will fallback also on empty string?

shirady: Only because I didn't add the user validation in this PR... I will change it from `??` to `||`.

PR #8014, opened 2024-05-06
Rework S3 object version ID and marker check

Neon-White:
### Explain the changes
1. Move the version ID check to happen before `read_object_md` is called, which takes care of both namespacestore and backingstore queries - while also potentially eliminating the case in which the store would enter a Rejected state
2. Remove the max-line linter error in `test_s3_ops`
3. Check for empty strings in all API calls containing a version ID
4. Check for empty strings in all API calls containing a version ID marker

1. [2265562](https://bugzilla.redhat.com/show_bug.cgi?id=2265562)

### Testing Instructions:
1. Upload an object to a bucket
5. Use the AWS CLI (`s3api` with `--version-id ""`) to try and access an existing object's version ID by providing an empty string
6. Verify that NooBaa throws the appropriate error (`An error occurred (InvalidArgument) when calling the GetObject operation: Version id cannot be the empty string`)

- [x] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -15,6 +15,9 @@ async function get_object(req, res) {
     const agent_header = req.headers['user-agent'];
     const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
     const encryption = s3_utils.parse_encryption(req);
+    if (s3_utils.check_obj_version_id(req.query.versionId)) {
+        throw new S3Error(S3Error.InvalidArgumentEmptyVersionId);
+    }
     let part_number;
     // If set, part_number should be positive integer from 1 to 10000
     if (req.query.partNumber) {
@@ -34,10 +37,6 @@ async function get_object(req, res) {
     }
     if (part_number) {
         md_params.part_number = part_number;
-    }
-
-    if (req.query.versionId?.length === 0) {
-        throw new S3Error(S3Error.InvalidArgumentEmptyVersionId);
     }
 
     const object_md = await req.object_sdk.read_object_md(md_params);
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -35,6 +35,11 @@ async function get_object(req, res) {
     if (part_number) {
         md_params.part_number = part_number;
     }
+
+    if (req.query.versionId?.length === 0) {
+        throw new S3Error(S3Error.InvalidArgumentEmptyVersionId);
+    }
+
     const object_md = await req.object_sdk.read_object_md(md_params);
 
     s3_utils.set_response_object_md(res, object_md);
@@ -59,9 +64,7 @@ async function get_object(req, res) {
         md_conditions,
         encryption,
     };
-    if (params.version_id?.length === 0) {
-        throw new S3Error(S3Error.InvalidArgumentEmptyVersionId);
-    }
+
     if (md_params.get_from_cache) {
         params.get_from_cache = true;
     }
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -15,7 +15,7 @@ async function get_object(req, res) {
     const agent_header = req.headers['user-agent'];
     const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
     const encryption = s3_utils.parse_encryption(req);
-    if (s3_utils.check_obj_version_id(req.query.versionId)) {
+    if (s3_utils.version_id_empty_string_check(req.query.versionId)) {
         throw new S3Error(S3Error.InvalidArgumentEmptyVersionId);
     }
     let part_number;
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -15,7 +15,7 @@ async function get_object(req, res) {
     const agent_header = req.headers['user-agent'];
     const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
     const encryption = s3_utils.parse_encryption(req);
-    const version_id = s3_utils.parse_version_id(req);
+    const version_id = s3_utils.parse_version_id(req.query.versionId);
     let part_number;
     // If set, part_number should be positive integer from 1 to 10000
     if (req.query.partNumber) {
@@ -26,7 +26,7 @@ async function get_object(req, res) {
     const md_params = {
         bucket: req.params.bucket,
         key: req.params.key,
-        version_id: version_id,
+        version_id,
         md_conditions,
         encryption,
     };
@@ -55,7 +55,7 @@ async function get_object(req, res) {
         obj_id: object_md.obj_id,
         bucket: req.params.bucket,
         key: req.params.key,
-        version_id: version_id,
+        version_id,
         content_type: object_md.content_type,
         noobaa_trigger_agent,
         md_conditions,
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -15,9 +15,7 @@ async function get_object(req, res) {
     const agent_header = req.headers['user-agent'];
     const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
     const encryption = s3_utils.parse_encryption(req);
-    if (s3_utils.version_id_empty_string_check(req.query.versionId)) {
-        throw new S3Error(S3Error.InvalidArgumentEmptyVersionId);
-    }
+    s3_utils.version_id_empty_string_check(req.query.versionId);
     let part_number;
     // If set, part_number should be positive integer from 1 to 10000
     if (req.query.partNumber) {
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -15,7 +15,7 @@ async function get_object(req, res) {
     const agent_header = req.headers['user-agent'];
     const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
     const encryption = s3_utils.parse_encryption(req);
-    s3_utils.version_id_empty_string_check(req.query.versionId);
+    const version_id = s3_utils.parse_version_id(req);
     let part_number;
     // If set, part_number should be positive integer from 1 to 10000
     if (req.query.partNumber) {
@@ -26,7 +26,7 @@ async function get_object(req, res) {
     const md_params = {
         bucket: req.params.bucket,
         key: req.params.key,
-        version_id: req.query.versionId,
+        version_id: version_id,
         md_conditions,
         encryption,
     };
@@ -55,7 +55,7 @@ async function get_object(req, res) {
         obj_id: object_md.obj_id,
         bucket: req.params.bucket,
         key: req.params.key,
-        version_id: req.query.versionId,
+        version_id: version_id,
         content_type: object_md.content_type,
         noobaa_trigger_agent,
         md_conditions,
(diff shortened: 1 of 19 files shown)


src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -35,6 +35,11 @@ async function get_object(req, res) {
     if (part_number) {
         md_params.part_number = part_number;
     }
+
+    if (req.query.versionId?.length === 0) {
```
guymguym, line 39: The same check should also be in head object.
Best add it as a common function in s3_utils to avoid copy-pasta.
Also move it a bit up so it's with the other request query parsing cases.
```js
const version_id = s3_utils.parse_version_id(req.query.versionId);
```

guymguym (changes requested):

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -15,6 +15,9 @@ async function get_object(req, res) {
     const agent_header = req.headers['user-agent'];
     const noobaa_trigger_agent = agent_header && agent_header.includes('exec-env/NOOBAA_FUNCTION');
     const encryption = s3_utils.parse_encryption(req);
+    if (s3_utils.version_id_empty_string_check(req.query.versionId)) {
+        throw new S3Error(S3Error.InvalidArgumentEmptyVersionId);
+    }
```
guymguym, line 20: the throw should be inside that function ...

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -35,6 +36,7 @@ async function get_object(req, res) {
     if (part_number) {
         md_params.part_number = part_number;
     }
+
```
shirady, line 39: if you didn't plan to add the newlines, please remove them.

src/endpoint/s3/s3_utils.js
```diff
@@ -622,6 +622,16 @@ function encode_uri_unless_already_encoded(uri = '') {
     return is_uri_already_encoded(uri) ? uri : encodeURIComponent(uri);
 }
 
+/**
+ * version_id_empty_string_check throws an error if version_id is an empty string
+ * @param {string} version_id
+ */
+function version_id_empty_string_check(version_id) {
+    if (version_id?.length === 0) {
```
shirady, line 630: Why did you prefer the length check over explicitly checking if the version ID is an empty string?
```suggestion
    if (version_id === '') {
```

src/endpoint/s3/s3_utils.js
```diff
@@ -622,6 +622,16 @@ function encode_uri_unless_already_encoded(uri = '') {
...
+        throw new S3Error(S3Error.InvalidArgumentEmptyVersionId);
+    }
+}
```
guymguym, line 633: I think this function is "leaking" its internal behavior and it is unneeded. As I suggested before, I personally would actually make this a "parse" function so that the caller passes the request and gets back the value of the version it from it after validating it. Similar to s3_utils parse_encryption/storage_class/part_number/etc. like this:
```suggestion
function parse_version_id(req) {
    const version_id = req.query.versionId;
    if (version_id?.length === 0) {
        throw new S3Error(S3Error.InvalidArgumentEmptyVersionId);
    }
    return version_id;
}
```

src/endpoint/s3/ops/s3_get_object.js
shirady: @guymguym why just head object and get object?
The function `s3_utils.parse_version_id(req);` is general and can be everywhere there is a parameter of `version_id`, for example: DeleteObject, DeleteObjectTagging, GetObjectAcl, PutObjectRetention, PutObjectTagging, RestoreObject (advance case ListObjectVersions - VersionIdMarker), etc.

We can search all `req.query.versionId` in the `src/endpoint/s3/ops`.

src/endpoint/s3/s3_utils.js
guymguym: oh I agree here too, the explicit empty string check is better than fishing for the length.

src/endpoint/s3/s3_utils.js
guymguym: I guess I changed my suggestion :)

src/endpoint/s3/ops/s3_get_object.js
guymguym: But I don't think VersionIdMarker in listObjectVersions is the same.

src/endpoint/s3/s3_utils.js
guymguym: checking for explicit empty string is better and more readable imho, and I'm even ready to fight with google about it.

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -53,15 +55,13 @@ async function get_object(req, res) {
         obj_id: object_md.obj_id,
         bucket: req.params.bucket,
         key: req.params.key,
-        version_id: req.query.versionId,
+        version_id: version_id,
```
guymguym, line 58: short property notation FTW -
```suggestion
        version_id,
```

src/endpoint/s3/ops/s3_head_object.js
```diff
@@ -11,6 +11,7 @@ const S3Error = require('../s3_errors').S3Error;
  */
 async function head_object(req, res) {
     const encryption = s3_utils.parse_encryption(req);
+    s3_utils.parse_version_id(req.query.versionId);
```
guymguym, line 14: also grab the returned value here and replace refs to req.query.versionId in this function
```suggestion
    const version_id = s3_utils.parse_version_id(req.query.versionId);
```

src/endpoint/s3/ops/s3_get_object.js
shirady: > But I don't think VersionIdMarker in listObjectVersions is the same.

@Neon-White I think It is something that we would need to check (from [AWS SDK V2 docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listObjectVersions-property)):

> VersionIdMarker — (String)
> Specifies the object version you want to start listing from.

guymguym: NP, I think we can go ahead and replace all refs to req.query.versionId in s3 ops.

But for versionId in xml fields or as marker - I would first check how aws handles empty string there - I don't want to assume I know the answer.

src/endpoint/s3/s3_rest.js
shirady, line 1: @Neon-White Why are you changing this file `s3_rest` also? Are you sure?

src/endpoint/s3/s3_utils.js
```diff
@@ -622,6 +622,17 @@ function encode_uri_unless_already_encoded(uri = '') {
     return is_uri_already_encoded(uri) ? uri : encodeURIComponent(uri);
 }
 
+/**
+ * parse_version_id throws an error if version_id is an empty string, and returns it otherwise
+ * @param {string} version_id
```
guymguym, line 627:
```suggestion
 * @param {string|undefined} version_id
```

src/endpoint/s3/s3_rest.js
```diff
@@ -118,9 +119,7 @@ async function handle_request(req, res) {
     usage_report.s3_usage_info.total_calls += 1;
     usage_report.s3_usage_info[op_name] = (usage_report.s3_usage_info[op_name] || 0) + 1;
 
-
-
-    if (req.query && req.query.versionId) {
+    if (s3_utils.parse_version_id(req.query?.versionId)) {
```
guymguym, line 122: I don't think this should be changed. the entire point in this code block was to prevent versioning for caching bucket. I agree that this check as a whole should have been added differently, perhaps just in a separate function than inline like that in the rest api flow, but still, we do not need to affect it with validation of the version id.

src/endpoint/s3/s3_rest.js
```diff
@@ -266,7 +265,7 @@ function _get_method_from_req(req) {
         dbg.error(`Got a not supported S3 op ${req.op_name} - doesn't suppose to happen`);
         throw new S3Error(S3Error.InternalError);
     }
-    if (req.query && req.query.versionId && s3_op.versioned) {
+    if (s3_utils.parse_version_id(req.query?.versionId) && s3_op.versioned) {
```
guymguym, line 268: again also here, the actual value of version id is not needed and should not be validated. the check was only if version id is set for the request. if you want you can add a separate `s3_utils.is_versioning_request(req)` which will just return `Boolean(req.query?.versionId)`

src/endpoint/s3/ops/s3_post_bucket_delete.js
```diff
@@ -29,7 +30,7 @@ async function post_bucket_delete(req) {
     const uniq_map = new Map();
     for (const item of delete_list) {
         const key = item.Key?.[0];
-        const version_id = item.VersionId?.[0];
+        const version_id = s3_utils.parse_version_id(item.VersionId?.[0]);
```
guymguym, line 33: did you check that AWS also errors in this api when sent with `<VersionId></VersionId>`?

src/endpoint/s3/ops/s3_get_object.js
guymguym: @Neon-White 

I tested on aws bucket -

1. delete-objects - does not return InvalidArgument at all with VersionId="", instead it just returns an array of errors with NoSuchVersion per requested version it cannot find - so we should undo the call to parse_version_id() in that op, and let our code continue with that empty string which will return the array of errors as it usually does -

```
❯ aws s3api delete-objects --bucket my-aws-bucket --delete '{"Objects":[ {"Key":"watman.png","VersionId":""}, {"Key":"README.md","VersionId":""} ]}'
{
    "Errors": [
        {
            "Key": "watman.png",
            "VersionId": "",
            "Code": "NoSuchVersion",
            "Message": "The specified version does not exist."
        },
        {
            "Key": "README.md",
            "VersionId": "",
            "Code": "NoSuchVersion",
            "Message": "The specified version does not exist."
        }
    ]
}
```

2. list-object-versions - VersionIdMarker is an input argument to this op and it does contain a versionId so it's not so different than other ops where versionId when passed as argument - I tested it with VersionIdMarker="" and it does reply with InvalidArgument with a similar yet more specific error message "A version-id marker cannot be empty". So this should actually be checked just like any other version id argument  -

```
❯ aws s3api list-object-versions --bucket my-aws-bucket --key-marker "watman.png" --version-id-marker ""

An error occurred (InvalidArgument) when calling the ListObjectVersions operation: A version-id marker cannot be empty.
```

src/endpoint/s3/ops/s3_post_bucket_delete.js
guymguym: I checked this one and this should be reverted, see other comment for details.

shirady (approved):
LGTM

src/endpoint/s3/s3_utils.js
```diff
@@ -622,6 +622,28 @@ function encode_uri_unless_already_encoded(uri = '') {
...
+ * parse_version_id_marker throws an error if version_id_marker is an empty string, and returns it otherwise
+ * @param {string|undefined} version_id_marker
+ */
+function parse_version_id_marker(version_id_marker) {
+    if (version_id_marker === '') {
+        throw new S3Error(S3Error.InvalidArgumentEmptyVersionIdMarker);
+    }
+    return version_id_marker;
+}
+
```
guymguym, line 646: I don't think we need two functions, this is handling the exact same type of information - a version_id. Since only the error is different you can have a default argument for all cases and for that different case of version_id_marker to pass the desired error spec from the caller as S3Error.InvalidArgumentEmptyVersionIdMarker

```suggestion
/**
 * parse_version_id throws an error if version_id is an empty string, and returns it otherwise
 * @param {string|undefined} version_id
 * @param {import('./s3_errors').S3ErrorSpec} [empty_err]
 */
function parse_version_id(version_id, empty_err = S3Error.InvalidArgumentEmptyVersionId) {
    if (version_id === '') {
        throw new S3Error(empty_err);
    }
    return version_id;
}

```

guymguym (approved):
Looks good!

nimrod-becker:
@guymguym 14 commits! not squashed :(

guymguym:
didnt notice. will be squashed by backports if needed.

PR #8021, opened 2024-05-07
Fixed issue for wrong creation_time for non-nsfs buckets while listin…

aayushchouhan09:
### Explain the changes
1. The fix has been implemented to display of the correct creation_time when listing non-NSFS buckets.

1. [Issue](https://github.ibm.com/ProjectAbell/abell-tracking/issues/35309)

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -985,7 +985,10 @@ async function list_buckets(req) {
     );
     return {
         buckets: _.map(buckets_by_name, function(bucket) {
-            return _.pick(bucket, 'name');
+            return {
+                name: bucket.name,
+                creation_date: bucket._id.getTimestamp().getTime()
+            };
         })
     };
 }
--- a/src/api/bucket_api.js
+++ b/src/api/bucket_api.js
@@ -399,6 +399,9 @@ module.exports = {
                             required: ['name'],
                             properties: {
                                 name: { $ref: 'common_api#/definitions/bucket_name' },
+                                creation_date: {
+                                    idate: true
+                                },
                             }
                         }
                     }


src/server/system_services/schemas/bucket_schema.js
```diff
@@ -139,6 +140,9 @@ module.exports = {
             type: 'string',
             enum: ['DISABLED', 'SUSPENDED', 'ENABLED']
         },
+        creation_date: {
+            type: 'string',
+        },
```
guymguym, line 145: We don't store dates as strings in the DB, look at how we do it for other DB schemas.
Also this kind of change would require an upgrade script to work for existing buckets.
However I think we can skip the DB schema for this purpose since we already have a date for every entity that can be extracted from `bucket._id.getTimestamp().getTime()`,  so you could just add it to the bucket api and make bucket_server reply with that date from the _id.

guymguym (changes requested):

src/server/system_services/schemas/bucket_schema.js
liranmauda: Nice, Thanks!
I was wondering how we could get the create time on a bucket that already exists.
This is the reason we did not add an upgrade script.

achouhan09: Updated and ready for review now. Thanks

src/server/system_services/bucket_server.js
```diff
@@ -985,7 +985,8 @@ async function list_buckets(req) {
     );
     return {
         buckets: _.map(buckets_by_name, function(bucket) {
-            return _.pick(bucket, 'name');
+            bucket.creation_date = bucket._id.getTimestamp().getTime();
+            return _.pick(bucket, ['name', 'creation_date']);
```
guymguym, line 989: this bucket object comes from system_store, and it is not a good idea not update it in-place.
instead just construct the answer object -
```suggestion
            return {
                name: bucket.name,
                creation_date: bucket._id.getTimestamp().getTime(),
            };
```

guymguym (changes requested):

src/server/system_services/bucket_server.js
achouhan09: Done, Thanks.

guymguym (approved):

guymguym:
please squash commits

PR #8028, opened 2024-05-08
[NC | NSFS] add support for DMAPI xattr based GLACIER storage class

tangledbytes:
### Explain the changes
This PR adds support expands the GLACIER storage class by introducing 4 new features which rely on DMAPI extended attributes. When NooBaa is configured to read DMAPI EA `NSFS_GLACIER_DMAPI_ENABLE = true` then the following features are available (each individually configurable):
1. Implicit restore status - Allows NooBaa to derive file state from DMAPI EA (`NSFS_GLACIER_DMAPI_IMPLICIT_RESTORE_STATUS = true`).
2. Implicit GLACIER - Allows NooBaa to derive storage class based on file state (`NSFS_GLACIER_DMAPI_IMPLICIT_SC = true`).
3. Tape header info parsing - Read and transmit tape info in HTTP header (`NSFS_GLACIER_DMAPI_TPS_HTTP_HEADER_ENABLE = true`).
4. Force Takeover - Allows NooBaa to takeover the lifecycle of objects which were not originally uploaded to NooBaa (`NSFS_GLACIER_DMAPI_ALLOW_NOOBAA_TAKEOVER = true`).

### Testing Instructions:

- [ ] Doc added/updated
- [x] Tests added


src/sdk/nsfs_glacier_backend/backend.js
```diff
@@ -153,7 +169,22 @@ class GlacierBackend {
...
     static get_restore_status(xattr, now, file_path) {
-        if (xattr[GlacierBackend.STORAGE_CLASS_XATTR] !== s3_utils.STORAGE_CLASS_GLACIER) return;
+		if (GlacierBackend.storage_class_from_xattr(xattr, config.NSFS_GLACIER_USE_DMAPI) !== s3_utils.STORAGE_CLASS_GLACIER) {
+			return;
+		}
+
+        if (config.NSFS_GLACIER_USE_DMAPI) {
+            if (xattr[GlacierBackend.GPFS_DMAPI_XATTR_TAPE_PREMIG]) {
+                const expiry = new Date();
+                expiry.setDate(expiry.getDate() + config.S3_RESTORE_REQUEST_MAX_DAYS);
```
guymguym, line 179: I think we should have a specific config for this amount of days `config.NSFS_GLACIER_DMAPI_PMIG_DAYS` which will be initialized by default to `config.S3_RESTORE_REQUEST_MAX_DAYS` (which is 30), but could be overridden in config.json. The reason is that I'm not sure that the default we pick now would make sense to all cases. Perhaps in some cases replying with now+1day make better sense for example.

Also we can add comment in the code to explain:

```js
// we do not know for how long the file is going to remain available,
// the expiry is set to now + fixed config, which means it's always appears
// to the user with the same amount of time left before it expires.
expiry.setDate(expiry.getDate() + config.NSFS_GLACIER_DMAPI_PMIG_DAYS);
```

src/sdk/nsfs_glacier_backend/backend.js
```diff
@@ -271,6 +303,29 @@ class GlacierBackend {
...
+	 * NOTE: For `use_dmapi` to work, the xattr must have been retrieved using fs_context
+	 * where backend is set to 'GPFS'.
+	 *
+	 * @param {nb.NativeFSXattr} xattr
+	 * @param {Boolean} [use_dmapi]
+	 * @returns {nb.StorageClass}
+	 */
+	static storage_class_from_xattr(xattr, use_dmapi) {
+		if (
+            use_dmapi &&
```
guymguym, line 320: I think you can simply use `config.NSFS_GLACIER_USE_DMAPI` inside here and avoid passing it in which will also reduce the clutter in calling this function.
```suggestion
            config.NSFS_GLACIER_USE_DMAPI &&
```

src/sdk/nsfs_glacier_backend/backend.js
```diff
@@ -153,7 +169,22 @@ class GlacierBackend {
...
      */
     static get_restore_status(xattr, now, file_path) {
-        if (xattr[GlacierBackend.STORAGE_CLASS_XATTR] !== s3_utils.STORAGE_CLASS_GLACIER) return;
+		if (GlacierBackend.storage_class_from_xattr(xattr, config.NSFS_GLACIER_USE_DMAPI) !== s3_utils.STORAGE_CLASS_GLACIER) {
+			return;
+		}
+
+        if (config.NSFS_GLACIER_USE_DMAPI) {
+            if (xattr[GlacierBackend.GPFS_DMAPI_XATTR_TAPE_PREMIG]) {
+                const expiry = new Date();
```
guymguym, line 178: this expiry might deserve a distinct name, like
```suggestion
                const premig_expiry = new Date();
```

src/sdk/nsfs_glacier_backend/backend.js
```diff
@@ -39,6 +40,20 @@ class GlacierBackend {
 
     static STORAGE_CLASS_XATTR = 'user.storage_class';
 
+	/**
```
guymguym, line 43: indentation

src/sdk/namespace_fs.js
```diff
@@ -2064,6 +2063,20 @@ class NamespaceFS {
...
+                // If the item is not already premigrated then throw error
+                if (!stat.xattr[GlacierBackend.GPFS_DMAPI_XATTR_TAPE_PREMIG]) {
+                    throw new Error('cannot restore externally managed object');
+                }
+
+                // If the item is premigrated then its a no-op
+                // Should result in HTTP: 200 OK
+                return false;
+            }
+
```
guymguym, line 2079: How about extracting this code to a function in GlacierBackend

guymguym (changes requested):

src/sdk/nsfs_glacier_backend/backend.js
tangledbytes: Using global configs makes testing quite tricky that's why I deliberately added it as a parameter wherever I could.

guymguym: ok, so maybe just set this config as the default for the argument so you can still override for testing but don't need to pass in normal calls
```js
static storage_class_from_xattr(xattr, use_dmapi = config.NSFS_GLACIER_USE_DMAPI) {
```

src/native/fs/fs_napi.cpp
```diff
@@ -230,7 +235,11 @@ parse_open_flags(std::string flags)
     return bits;
 }
 
-const static std::vector<std::string> GPFS_XATTRS{ GPFS_ENCRYPTION_XATTR_NAME };
+const static std::vector<std::string> GPFS_XATTRS{
+	GPFS_ENCRYPTION_XATTR_NAME,
+	GPFS_DMAPI_XATTR_TAPE_INDICATOR,
+	GPFS_DMAPI_XATTR_TAPE_PREMIG,
+};
```
guymguym, line 242: this list is going to be used in every stat call... I think we want to optimize it so that it will not include dmapi xattrs when config.NSFS_GLACIER_USE_DMAPI is false. 

For this I would add this to namespace_fs in prepare_fs_context to set fs_context.use_dmapi = config.NSFS_GLACIER_USE_DMAPI. This is also needed in manage_nsfs_glacier in force_gpfs_fs_context ().

Then back here in fs_napi you can store it in FSWorker and pass it to get_fd_gpfs_xattr... a long way... but it would save a these two calls when the config is false which I think is worth the coding effort.

src/sdk/nsfs_glacier_backend/backend.js
```diff
@@ -5,6 +5,7 @@ const nb_native = require('../../util/nb_native');
 const s3_utils = require('../../endpoint/s3/s3_utils');
 const { round_up_to_next_time_of_day } = require('../../util/time_utils');
 const dbg = require('../../util/debug_module')(__filename);
+const config = require('../../../config');
 
 class GlacierBackend {
```
guymguym, line 10: What do you think about renaming this class to be called simply Glacier?
This is now used in so many places, that I wish it was more concise...
I also think this file can move to `src/sdk/glacier.js` and `src/sdk/glacier_tapecloud.js`

guymguym (changes requested):

src/sdk/nsfs_glacier_backend/backend.js
tangledbytes: Yep, done.

src/native/fs/fs_napi.cpp
tangledbytes: Done, PTAL.

src/sdk/glacier.js
```diff
@@ -136,31 +152,50 @@ class GlacierBackend {
...
      * @returns {nb.RestoreStatus | undefined}
      */
     static get_restore_status(xattr, now, file_path) {
-        if (xattr[GlacierBackend.STORAGE_CLASS_XATTR] !== s3_utils.STORAGE_CLASS_GLACIER) return;
+        const storage_class = Glacier.storage_class_from_xattr(xattr);
+        if (storage_class !== s3_utils.STORAGE_CLASS_GLACIER) {
+            return;
+        }
+
+        if (Glacier.is_externally_managed(xattr)) {
```
guymguym, line 178: if xattr.storage_class!=GLACIER we don't really want to consider it as externally managed and skip this case.

src/sdk/glacier.js
```diff
@@ -185,20 +220,20 @@ class GlacierBackend {
...
-                    state: GlacierBackend.RESTORE_STATUS_CAN_RESTORE,
+                    state: Glacier.RESTORE_STATUS_CAN_RESTORE,
                 };
             }
 
             return {
                 ongoing: false,
                 expiry_time: restore_expiry,
-                state: GlacierBackend.RESTORE_STATUS_RESTORED,
+                state: Glacier.RESTORE_STATUS_RESTORED,
```
guymguym, line 236: in this case, the object is not externally managed (meaning xattr.storage_class=GLACIER), but if config.USE_DMAPI is on, we can still look at the dmapi state to verify that indeed the file is premigrated, and print a warning.

src/sdk/namespace_fs.js
```diff
@@ -2085,34 +2085,44 @@ class NamespaceFS {
...
 
-            if (restore_status.state === GlacierBackend.RESTORE_STATUS_CAN_RESTORE) {
+            if (Glacier.is_externally_managed(stat.xattr)) {
+                if (restore_status.state === Glacier.RESTORE_STATUS_RESTORED) {
+                    // If the item is premigrated then its a no-op
+                    // Should result in HTTP: 200 OK
+                    return false;
+                }
+
+                throw new Error('cannot restore externally managed object');
```
guymguym, line 2095: should we allow restore on externally managed? maybe add an optional config to allow that?

src/sdk/glacier.js
```diff
@@ -271,8 +307,62 @@ class GlacierBackend {
...
+     * is set to true).
+     *
+     * @param {nb.NativeFSXattr} xattr
+     * @returns {boolean}
+     */
+    static is_externally_managed(xattr) {
+        return Boolean(
+            config.NSFS_GLACIER_USE_DMAPI && (
+                xattr[Glacier.GPFS_DMAPI_XATTR_TAPE_INDICATOR] ||
+                xattr[Glacier.GPFS_DMAPI_XATTR_TAPE_PREMIG]
```
guymguym, line 348: externally managed should also check that xattr.storage_class != GLACIER

guymguym (changes requested):

nimrod-becker:
@tangledbytes still relevant?

tangledbytes:
@guymguym we still want this right?

src/endpoint/s3/s3_utils.js
```diff
@@ -325,6 +325,12 @@ function set_response_object_md(res, object_md) {
 
         res.setHeader('x-amz-restore', restore);
     }
+    if (storage_class === STORAGE_CLASS_GLACIER) {
+        object_md.restore_status?.backend_meta?.forEach?.((meta, idx) => {
+            const header = Object.keys(meta).map(key => `${key}=${meta[key]}`).join('&');
```
tangledbytes, line 330: Use query strings instead of doing it manually.

```
const querystring = require('node:querystring');
```

tangledbytes:
I just noticed that this PR is extremely ugly (thanks to my editor). I will fix it ASAP.

config.js
```diff
@@ -873,6 +882,21 @@ config.NSFS_GLACIER_EXPIRY_TZ = 'LOCAL';
...
+
+// NSFS_GLACIER_DMAPI_ALLOW_NOOBAA_TAKEOVER allows NooBaa to take over lifecycle
+// management of an object which was originally NOT managed by NooBaa.
+config.NSFS_GLACIER_DMAPI_ALLOW_NOOBAA_TAKEOVER = false;
+
+// NSFS_GLACIER_DMAPI_TPS_HTTP_HEADER if enabled will add additional HTTP headers
+// `x-tape-meta-copy-n` based on `dmapi.IBMTPS` EA.
+//
+// For this to work, NSFS_GLACIER_USE_DMAPI must be set to `true`.
+config.NSFS_GLACIER_DMAPI_TPS_HTTP_HEADER = config.NSFS_GLACIER_USE_DMAPI || false;
```
tangledbytes, line 898: Have this check within the code as well. config-local.js/env overrides are later.

src/endpoint/s3/s3_utils.js
```diff
@@ -325,6 +326,13 @@ function set_response_object_md(res, object_md) {
 
         res.setHeader('x-amz-restore', restore);
     }
+    if (config.NSFS_GLACIER_DMAPI_TPS_HTTP_HEADER && storage_class === STORAGE_CLASS_GLACIER) {
+        object_md.restore_status?.backend_meta?.forEach?.((meta, idx) => {
+            // @ts-ignore - For some TS check doesn't like "meta" being passed to querystring
+            const header = querystring.stringify(meta);
+            res.setHeader(`x-tape-meta-copy-${idx}`, header);
```
tangledbytes, line 333: Make this header configurable as well.

src/sdk/glacier.js
```diff
@@ -271,8 +330,93 @@ class GlacierBackend {
...
+     * extended attribute. It will use DMAPI EA if `use_dmapi` is set to true.
+     *
+     * NOTE: For `use_dmapi` to work, the xattr must have been retrieved using fs_context
+     * where backend is set to 'GPFS'.
+     *
+     * @param {nb.NativeFSXattr} xattr
+     * @param {Boolean} [use_dmapi]
+     * @returns {nb.StorageClass}
+     */
+   static storage_class_from_xattr(xattr, use_dmapi = config.NSFS_GLACIER_USE_DMAPI) {
```
tangledbytes, line 347: Consider `NSFS_GLACIER_USE_DMAPI` to just enable reading DMAPI xattr and have a different config which controls the implicit `GLACIER` storage class.

`config.NSFS_GLACIER_USE_DMAPI` can enable 4 features but the base should be to just read DMAPI xattr:
1. Implicit restore status
2. Implicit GLACIER
3. Tape header info
4. Force Takeover

src/sdk/glacier.js
```diff
@@ -271,8 +330,93 @@ class GlacierBackend {
...
+        if (
+            use_dmapi &&
+            xattr[Glacier.GPFS_DMAPI_XATTR_TAPE_INDICATOR] &&
```
tangledbytes, line 350: Re-verify why we need separate checks here.

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -212,12 +213,34 @@ function get_tz_date(hours, mins, secs, tz) {
...
+function force_gpfs_fs_context(fs_context) {
+    if (config.NSFS_GLACIER_USE_DMAPI) {
+        if (!nb_native().fs.gpfs) {
+            throw new Error('cannot use DMAPI EA: gpfs dl not loaded');
+        }
+
+        return { ...fs_context, backend: 'GPFS', use_dmapi: true };
+    }
+
+    return { ...fs_context };
```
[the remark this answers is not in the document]
tangledbytes: We can but would make testing locally more troublesome though, no?

src/endpoint/s3/s3_utils.js
```diff
@@ -325,6 +326,13 @@ function set_response_object_md(res, object_md) {
 
         res.setHeader('x-amz-restore', restore);
     }
+    if (config.NSFS_GLACIER_DMAPI_TPS_HTTP_HEADER_ENABLE && storage_class === STORAGE_CLASS_GLACIER) {
```
guymguym, line 329: can we return those if exist regardless of the object storage class?

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -135,6 +136,28 @@ async function run_glacier_operation(fs_context, log_namespace, cb) {
...
+ * force_gpfs_fs_context returns a shallow copy of given
+ * fs_context with backend set to 'GPFS'.
+ *
+ * NOTE: The function will throw error if it detects that libgfs
+ * isn't loaded.
+ *
+ * @param {nb.NativeFSContext} fs_context
+ * @returns {nb.NativeFSContext}
+ */
+function force_gpfs_fs_context(fs_context) {
```
guymguym, line 149: since this doesn't always force a gpfs backend (see last case), perhaps better rename to prepare_galcier_fs_context

src/sdk/glacier.js
```diff
@@ -154,13 +179,35 @@ class GlacierBackend {
...
     static get_restore_status(xattr, now, file_path) {
-        if (xattr[GlacierBackend.STORAGE_CLASS_XATTR] !== s3_utils.STORAGE_CLASS_GLACIER) return;
+        const storage_class = Glacier.storage_class_from_xattr(xattr);
+        if (storage_class !== s3_utils.STORAGE_CLASS_GLACIER) {
+            return;
+        }
+
+        const tape_info = Glacier.parse_tape_info(xattr);
+
+        if (Glacier.is_externally_managed(xattr)) {
```
guymguym, line 189: I think we should have the option to set restore status from dmapi even when IMPLICIT_SC is not enabled.

guymguym (approved):

PR #8055, opened 2024-05-20
NSFS | NC | IAM Service - Users CRUD API Implementation

shirady:
### Explain the changes
1. Implement the functions related to users in `AccountSpaceFS` (CRUD) - all the functions here will only work on an IAM user without access keys (no encryption is involved at this point).
2. Add more properties to `nsfs_account_schema` (all of them are not required):
- `owner` = owner is the account ID that owns this account (permission-wise).
- `creator` = creator is the account ID that created this account (internal information).
- `path` = AWS path (identifier).
3. Add function `generate_id` in `manage_nsfs_cli_utils.js` to avoid importing MongoDB code in new files, and reuse this function only.
4. Add errors in `IamError` from the doc of users' action API.
5. Add functions in `iam_utils`.
6. Add unit tests for the function in `iam_utils`.
7. Update unit test for user function in `accountspace_fs`.
8. Add unit test in `account_schema_validation` for the case of an account without access keys (empty array), and for added properties in the schema.
9. Rename:
  - `path` to `iam_path`
  - `AWS_DEFAULT_PATH` to `IAM_DEFAULT_PATH`
  - Rename `path_prefix` to `iam_path_prefix`
10. Add `ResponseMetadata` protection in `reply`.
11. Add more details in IAM documentation:
  - Add documentation related to users in the design docs (`docs/design/iam.md`)
  - Add `Get Started` section that would be the demo for IAM user management (`docs/dev_guide/nc_nsfs_iam_developer_doc.md`)

### Issues: 
Gaps:
1. Currently naming scope of `username` is global scope - all config files names, it should be only under the root account in the future.
2. Currently the files are written in the config file at the same place as the root account, meaning by default in `/etc/noobaa.conf.d/accounts/`, no hierarchy in FS (under discussion if it would be changed in the future).
3. Data is read without encryption/decryption in users CRUD API implementation in accountspace_fs.
4. Parsing and validating the params (for example: username).
5. No pagination in the `list_users` implementation.
6. No `NoobaaEvent` at this point.
7. `_translate_error_codes` was copied from `NameSpaceFS` and `BucketSpaceFS`, but might need to be refactored to match the needs.
8. In some cases the ARN that we use will not contain the user's path in errors.
9. In `password_last_used` we send a dummy value.
10. Currently we copy the `master_key_id` from the root account, we should take the most updated `master_key_id`.
11. Clean the account cache after `update_user` (we will add this in the PR of IAM access keys).

#### Unit Tests
Please run:
1. `sudo npx jest test_accountspace_fs.test.js`
2. `npx jest test_iam_utils.test.js`
3. `npx jest test_nc_nsfs_account_schema_validation.test.js`

#### Manual Test
Currently, we do not validate the input, so the test should use only valid input.
4. Use AWS CLI to send requests to the IAM service, for example:
 `s3-nc-user-1-iam iam create-user --user-name Bob --path '/division_abc/subdivision_xyz/'`
 `s3-nc-user-1-iam iam get-user --user-name Bob`
 `s3-nc-user-1-iam iam update-user --user-name Bob --new-user-name Robert --new-path '/division_abc/subdivision_abc/'`
 `s3-nc-user-1-iam iam delete-user --user-name Bob`
 `s3-nc-user-1-iam iam list-users`

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/endpoint/iam/iam_rest.js
+++ b/src/endpoint/iam/iam_rest.js
@@ -112,6 +112,7 @@ async function handle_request(req, res) {
     }
 
     const reply = await op.handler(req, res);
+    add_response_metadata_if_not_exists(reply, req.request_id); // unique to IAM
     http_utils.send_reply(req, res, reply, {
         ...options,
         body: op.body,
@@ -174,6 +175,27 @@ function verify_op_request_body_type(req) {
             `in request, ${headers}`);
         // GAP - need to make sure which error we need to throw
         throw new IamError(IamError.InvalidParameterValue);
+    }
+}
+
+// this function was added to protect the structure of IAM reply
+// and make sure that every op has ResponseMetadata inside it 
+// (in case developers add new ops in endpoint/iam/ops and accidentally did not add it)
+// structure of reply: { <op>Response { <op>Result (if exists), ResponseMetadata } }
+function add_response_metadata_if_not_exists(reply, request_id) {
+    const reply_keys = Object.keys(reply);
+    if (reply_keys.length !== 1) {
+        dbg.error('add_response_metadata_if_not_exists: reply structure does not meet standards' +
+        'should have only one object inside reply');
+        throw new IamError(IamError.InternalFailure);
+    }
+    const reply_response_key = reply_keys[0];
+    const reply_response_object = reply[reply_response_key];
+    const response_metadata = reply_response_object.ResponseMetadata;
+    if (!response_metadata) {
+        reply_response_object.ResponseMetadata = {
+            RequestId: request_id,
+        };
     }
 }
 
--- a/src/endpoint/iam/iam_utils.js
+++ b/src/endpoint/iam/iam_utils.js
@@ -1,10 +1,28 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
+const _ = require('lodash');
 const s3_utils = require('../s3/s3_utils');
 
-const AWS_DEFAULT_PATH = '/';
+const IAM_DEFAULT_PATH = '/';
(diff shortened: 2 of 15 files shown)


src/endpoint/iam/iam_utils.js
```diff
@@ -19,20 +34,39 @@ function format_iam_xml_date(input) {
...
+        return `${basic_structure}${iam_path}${username}`;
     }
     return `${basic_structure}/${username}`;
 }
 
+/**
+ * get_full_action_name returns the full action name
+ * @param {string} action (The action name as it is in AccountSpace)
+ */
+function get_full_action_name(action) {
```
guymguym, line 54: get_action_message_title and ACTION_MESSAGE_TITLE

src/endpoint/iam/iam_rest.js
```diff
@@ -187,5 +189,26 @@ function verify_op_request_body_type(req) {
...
+    const reply_keys = Object.keys(reply);
+    if (reply_keys.length !== 1) {
+        dbg.error('add_response_metadata_if_not_exists: reply structure does not meet standards' +
+        'should have only one object inside replay');
+        throw new IamError(IamError.InternalFailure);
+    }
+    const reply_response_key = reply_keys[0];
+    const reply_response_object = reply[reply_response_key];
+    const response_metadata = reply_response_object.ResponseMetadata;
+    if (_.isUndefined(response_metadata)) {
```
guymguym, line 206:
```suggestion
    if (!response_metadata) {
```

guymguym (approved):

guymguym:
@shirady Please do not forget to squash commits. thanks

PR #8062, opened 2024-05-21
[MCG] Store relevant Owner metadata for MCG objects

vh05:
### Goal

Allow customers to keep track of object owner of an object via the S3 API

### Problem

Currently when using boto3.client.list_objects, the metadata in the response under the "Owner" property is always set to

{'DisplayName': 'NooBaa', 'ID': '123'}

regardless of the name of the account that created the bucket.

### Why is this important?

Currently when using boto3.client.list_objects, the metadata in the response under the "Owner" property is always set to

{'DisplayName': 'NooBaa', 'ID': '123'}

regardless of the name of the account that created the bucket.

### Prioritized Scenarios
**In Scope**

Assign obc name or access key as the owner

### Documentation Requirements

Yes. document the approach we take

### Customers

No

### Customer Facing Story

As an administrator, I should be able to check which account uploaded a given object to an MCG bucket.

### Gap: 
Since ACLs are not suported, we are assigning obc owner as the owner of the object

**Fixes**: https://issues.redhat.com/browse/RHSTOR-5501

### DIFF
--- a/src/endpoint/s3/ops/s3_get_bucket.js
+++ b/src/endpoint/s3/ops/s3_get_bucket.js
@@ -72,7 +72,7 @@ async function get_bucket(req) {
                 NextMarker: req.query.delimiter ? reply.next_marker : undefined,
             }),
         },
-        _.map(reply.objects, obj => ({
+        await Promise.all(_.map(reply.objects, async obj => ({
             Contents: {
                 Key: field_encoder(obj.key),
                 // if the object specifies last_modified_time we use it, otherwise take create_time.
@@ -81,11 +81,11 @@ async function get_bucket(req) {
                 LastModified: s3_utils.format_s3_xml_date(obj.last_modified_time || obj.create_time),
                 ETag: `"${obj.etag}"`,
                 Size: obj.size,
-                Owner: (!list_type || req.query['fetch-owner']) && s3_utils.get_object_owner(reply),
+                Owner: (!list_type || req.query['fetch-owner']) && (await s3_utils.get_object_owner(obj, req.object_sdk)),
                 StorageClass: s3_utils.parse_storage_class(obj.storage_class),
                 RestoreStatus: get_object_restore_status(obj, restore_status_requested)
             }
-        })),
+        }))),
         _.map(reply.common_prefixes, prefix => ({
             CommonPrefixes: {
                 Prefix: field_encoder(prefix) || ''
--- a/src/endpoint/s3/ops/s3_get_bucket.js
+++ b/src/endpoint/s3/ops/s3_get_bucket.js
@@ -81,7 +81,7 @@ async function get_bucket(req) {
                 LastModified: s3_utils.format_s3_xml_date(obj.last_modified_time || obj.create_time),
                 ETag: `"${obj.etag}"`,
                 Size: obj.size,
-                Owner: (!list_type || req.query['fetch-owner']) && s3_utils.DEFAULT_S3_USER,
+                Owner: (!list_type || req.query['fetch-owner']) && s3_utils.get_object_owner(reply),
                 StorageClass: s3_utils.parse_storage_class(obj.storage_class),
                 RestoreStatus: get_object_restore_status(obj, restore_status_requested)
             }
--- a/src/sdk/namespace_nb.js
+++ b/src/sdk/namespace_nb.js
@@ -61,7 +61,6 @@ class NamespaceNB {
     async list_objects(params, object_sdk) {
         if (this.target_bucket) params = _.defaults({ bucket: this.target_bucket }, params);
         const object_info = await object_sdk.rpc_client.object.list_objects(params);
-        object_info.object_owner = await this.get_object_owner(object_sdk, params.bucket);
         return object_info;
     }
 
@@ -75,14 +74,6 @@ class NamespaceNB {
         return object_sdk.rpc_client.object.list_object_versions(params);
     }
 
-    async get_object_owner(object_sdk, bucket) {
-        // TODO: in the future we will add extra implementation per namespace
-        const info = await object_sdk.read_bucket_sdk_config_info(bucket);
-        return {
-              name: info.bucket_owner.unwrap(),
-              id: info.bucket_info.owner_account.id,
-          };
-    }
     /////////////////
     // OBJECT READ //
     /////////////////
--- a/src/sdk/namespace_nb.js
+++ b/src/sdk/namespace_nb.js
@@ -58,9 +58,11 @@ class NamespaceNB {
     // OBJECT LIST //
     /////////////////
 
-    list_objects(params, object_sdk) {
+    async list_objects(params, object_sdk) {
         if (this.target_bucket) params = _.defaults({ bucket: this.target_bucket }, params);
-        return object_sdk.rpc_client.object.list_objects(params);
+        const object_info = await object_sdk.rpc_client.object.list_objects(params);
+        object_info.object_owner = await this.get_object_owner(object_sdk, params.bucket);
+        return object_info;
     }
 
     list_uploads(params, object_sdk) {
@@ -73,6 +75,14 @@ class NamespaceNB {
         return object_sdk.rpc_client.object.list_object_versions(params);
     }
 
+    async get_object_owner(object_sdk, bucket) {
+        // TODO: in the future we will add extra implementation per namespace
+        const info = await object_sdk.read_bucket_sdk_config_info(bucket);
+        return {
+              name: info.bucket_owner.unwrap(),
+              id: info.bucket_info.owner_account.id,
+          };
+    }
     /////////////////
     // OBJECT READ //
     /////////////////
--- a/src/sdk/namespace_s3.js
+++ b/src/sdk/namespace_s3.js
@@ -94,18 +94,16 @@ class NamespaceS3 {
             objects: _.map(res.Contents, obj => this._get_s3_object_info(obj, params.bucket)),
             common_prefixes: _.map(res.CommonPrefixes, 'Prefix'),
             is_truncated: res.IsTruncated,
-            next_marker: res.NextMarker,
-            object_owner: await this.get_object_owner(object_sdk, params.bucket),
+            next_marker: res.NextMarker
         };
     }
 
-    async get_object_owner(object_sdk, bucket) {
-        // TODO: in the future we will add extra implementation per namespace
-        const info = await object_sdk.read_bucket_sdk_config_info(bucket);
-        return {
-              name: info.bucket_owner.unwrap(),
-              id: info.bucket_info.owner_account.id,
-          };
+    /**
+     * _get_object_owner in the future we will return object owner
+     * currently not implemented because ACLs are not implemented as well
+     */
+    _get_object_owner() {
+        return undefined;
     }
 
     async list_uploads(params, object_sdk) {
@@ -788,6 +786,7 @@ class NamespaceS3 {
             sha256_b64: undefined,
             stats: undefined,
             tagging: undefined,
+            object_owner: this._get_object_owner()
         };
     }
 
--- a/src/sdk/namespace_s3.js
+++ b/src/sdk/namespace_s3.js
@@ -95,7 +95,17 @@ class NamespaceS3 {
             common_prefixes: _.map(res.CommonPrefixes, 'Prefix'),
             is_truncated: res.IsTruncated,
             next_marker: res.NextMarker,
+            object_owner: await this.get_object_owner(object_sdk, params.bucket),
         };
+    }
+
+    async get_object_owner(object_sdk, bucket) {
+        // TODO: in the future we will add extra implementation per namespace
+        const info = await object_sdk.read_bucket_sdk_config_info(bucket);
+        return {
+              name: info.bucket_owner.unwrap(),
+              id: info.bucket_info.owner_account.id,
+          };
     }
 
     async list_uploads(params, object_sdk) {
--- a/src/sdk/nb.d.ts
+++ b/src/sdk/nb.d.ts
@@ -399,10 +399,16 @@ interface ObjectMD {
     lock_settings: { retention: { mode: string; retain_until_date: Date; }, legal_hold: { status: string } };
 }
 
+interface ObjectOwner {
+    id: string,
+    name: string;
+}
+
 interface ObjectInfo {
     obj_id: string;
     bucket: string;
     key: string;
+    object_owner?: ObjectOwner;
     version_id: string;
     lock_settings: { retention: { mode: string; retain_until_date: Date; }, legal_hold: { status: string } };
     is_latest: boolean;
--- a/src/test/unit_tests/test_s3_ops.js
+++ b/src/test/unit_tests/test_s3_ops.js
@@ -527,6 +527,34 @@ mocha.describe('s3_ops', function() {
             assert.strictEqual(res2.Contents.length, (res1.Contents.length + 2));
         });
 
+        mocha.it('list-objects should return proper object owner and id', async function() {
+            if (is_azure_mock) this.skip();
+            this.timeout(120000);
+            await s3.copyObject({
+                Bucket: bucket_name,
+                Key: text_file2,
+                CopySource: `/${bucket_name}/${text_file1}`,
+            });
+            await s3.copyObject({
+                Bucket: bucket_name,
+                Key: text_file3,
+                CopySource: `/${BKT5}/${text_file5}`,
+            });
+            const res = await s3.listObjects({ Bucket: bucket_name });
+
+            assert.strictEqual(res.Contents[0].Key, text_file1);
+            assert.strictEqual(res.Contents[0].Owner.DisplayName, "coretest@noobaa.com");
+            assert.notEqual(res.Contents[0].Owner.ID, undefined);
+
+            assert.strictEqual(res.Contents[1].Key, text_file2);
+            assert.strictEqual(res.Contents[1].Owner.DisplayName, "coretest@noobaa.com");
+            assert.notEqual(res.Contents[1].Owner.ID, undefined);
+
+            assert.strictEqual(res.Contents[2].Key, text_file3);
+            assert.strictEqual(res.Contents[2].Owner.DisplayName, "coretest@noobaa.com");
+            assert.notEqual(res.Contents[2].Owner.ID, undefined);
+        });
+
         mocha.it('should copy text-file tagging', async function() {
             if (is_azure_mock) this.skip();
             this.timeout(120000);
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -1444,6 +1444,12 @@ module.exports = {
                 s3_signed_url: { type: 'string' },
                 lock_settings: { $ref: 'common_api#/definitions/lock_settings' },
                 storage_class: { $ref: 'common_api#/definitions/storage_class_enum' },
+                // currently no properties for the object as there is no implementation
+                object_owner: {
+                    type: 'object',
+                    properties: {
+                    }
+                },
             }
         },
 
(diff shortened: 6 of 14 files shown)


src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -84,6 +84,12 @@ async function get_bucket(req) {
     };
 }
 
+function get_object_owner(params) {
+    return Object.freeze({
+        DisplayName: params.bucket,
```
romayalon, line 89: Hey @vh05 :)
Is this a fix for https://issues.redhat.com/browse/RHSTOR-5501 ? if so can you please attach it to the PR's description?
Originally @sagihirshfeld opened this RFE in order to get the object owner on NC environemnt, but according to https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-ownership-retrieving.html
There are 2 options for object owner - 
1. Bucket owner (default, ACL disabled)
2. Object writer (ACLs enabled)

Since ACLs are not supported - 
1. We should add a comment here about it/add it to the ACLs gap.
2. Return the bucket owner and not the bucket name (on containerized env currently obc name = obc owner account name but this won't work on NC buckets)

WDYT?

src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -84,6 +84,12 @@ async function get_bucket(req) {
     };
 }
 
+function get_object_owner(params) {
+    return Object.freeze({
```
romayalon, line 88: As per today's code and according to [AWS](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Owner.html), owner has also ID - 
const DEFAULT_S3_USER = Object.freeze({
    ID: '123',
    DisplayName: 'NooBaa'
});

ID is missing in the new response

src/endpoint/s3/ops/s3_get_bucket.js
vh05: sure.

About account owner. 

If we list the accounts, then the obc owner's name is synonymous with bucket name. Though just use the bucket name itself since it is same as obc owner's name.

```
- access_keys:
  - access_key: 6FiNkA6WTGvMMHlMlkqC
    secret_key: QqkBPoGEQp0GlPJrzJ/E6zOjFafBi4xxsqBH2ArD
  bucket_claim_owner: n1-3d645c14-c014-4eda-a5c2-be5c81de5433
```

romayalon: @vh05 that's true when using containerized NooBaa.
On Non Containerized deployment we don't have OBCs, this bug was originally opened for NC env, so I think you should fix on both deployments :)

src/endpoint/s3/ops/s3_get_bucket.js
guymguym: why are we not getting this info from req.object_sdk.requesting_account?? we should have both id and name there. and in case of anonymous request we should check how aws returns it.

guymguym: ah sorry i mixed up the requesting user and the object owner... so instead, how about we let the namespace return this info in obj.owner, and then we can really consider how to store and retrieve this info correctly per ns type.

vh05: > how about we let the namespace return this info in obj.owner, and then we can really consider how to store and retrieve this info correctly per ns type.

Checking this

vh05: @guymguym Just wanted to know, what namespace we should search for this information. The bucket namespace doesn't have this information. Where else to find this.

guymguym: I meant that the s3 op will not make this decision in this module.
Instead this information should be returned from the object_sdk->namespace_xx level so that every namespace can implement this as desired.
Currently all namespaces can return the bucket owner information, but I can see how this can be enhanced later to allow different owners per object.

vh05: Makes sense :)

src/sdk/nb.d.ts
```diff
@@ -403,6 +403,8 @@ interface ObjectInfo {
     obj_id: string;
     bucket: string;
     key: string;
+    object_owner_id: string,
```
romayalon, line 406: can you change it to `owner: object` ? and return id + name together?

src/sdk/namespace_nb.js
```diff
@@ -58,9 +58,18 @@ class NamespaceNB {
...
-        return object_sdk.rpc_client.object.list_objects(params);
+        const object_info = await object_sdk.rpc_client.object.list_objects(params);
+        let bucket_info;
+        try {
+            bucket_info = await object_sdk.read_bucket_sdk_config_info(params.bucket);
+        } catch (err) {
+            if (err.rpc_code !== 'NO_SUCH_BUCKET') throw err;
+        }
+        object_info.object_owner_name = this.get_object_owner_name(bucket_info.bucket_info.name.unwrap());
+        object_info.object_owner_id = bucket_info._id;
```
romayalon, line 71: As we discussed, the owner should be the bucket owner account name and not the bucket name itself.

src/test/system_tests/ceph_s3_tests/s3-tests-lists/s3_tests_pending_list.txt
```diff
@@ -130,4 +130,5 @@ s3tests/functional/test_headers.py::test_object_create_bad_date_none_aws2
 s3tests/functional/test_headers.py::test_bucket_create_bad_authorization_invalid_aws2
 s3tests/functional/test_headers.py::test_bucket_create_bad_date_none_aws2
 s3tests_boto3/functional/test_s3.py::test_get_object_ifnonematch_good
-s3tests_boto3/functional/test_s3.py::test_get_object_ifmodifiedsince_failed
\ No newline at end of file
+s3tests_boto3/functional/test_s3.py::test_get_object_ifmodifiedsince_failed
```
romayalon, line 133: please also add the relevant unit tests

romayalon (changes requested):

src/sdk/namespace_nb.js
vh05: Sure. will return the obect owner name directly coming from the list objects

src/test/system_tests/ceph_s3_tests/s3-tests-lists/s3_tests_pending_list.txt
vh05: You mean to say remove the relevant tests  from the pending list ?

romayalon: no sorry, the comment isn't relevant to this file, please just add regular unit tests for this issue.

src/sdk/nb.d.ts
vh05: sure. incorporated the changes

src/test/system_tests/ceph_s3_tests/test_ceph_s3_deploy.sh
```diff
@@ -38,7 +38,7 @@ commit_epoch=$(git show -s --format=%ci ${CEPH_TESTS_VERSION} | awk '{print $1}'
 commit_date=$(date -d ${commit_epoch} +%s)
 current_date=$(date +%s)
 
-max_days="180"
+max_days="360"
```
romayalon, line 41: I think it was fixed already - https://github.com/noobaa/noobaa-core/pull/8191
so it can be removed

src/sdk/namespace_s3.js
```diff
@@ -90,14 +90,31 @@ class NamespaceS3 {
...
             is_truncated: res.IsTruncated,
             next_marker: res.NextMarker,
+            object_owner: this.get_object_owner(bucket_info),
         };
     }
 
+    get_object_owner(info) {
+        return {
+              name: info.bucket_owner.unwrap(),
+              id: info._id,
```
romayalon, line 114: This is the id of the bucket? we need to have the id of the bucket_owner account

src/sdk/namespace_s3.js
```diff
@@ -90,14 +90,31 @@ class NamespaceS3 {
...
         return {
             objects: _.map(res.Contents, obj => this._get_s3_object_info(obj, params.bucket)),
             common_prefixes: _.map(res.CommonPrefixes, 'Prefix'),
             is_truncated: res.IsTruncated,
             next_marker: res.NextMarker,
+            object_owner: this.get_object_owner(bucket_info),
         };
     }
 
+    get_object_owner(info) {
```
romayalon, line 111: For re-use of code I suggest having a single function in object sdk that brings the info about the bucket owner and on these namespace*.js functions we will add in the future extra implementation per namespace, but for now these functions will look like - 

```
// in every namespace*.js file - 
get_object_owner(object_sdk) {
    // TODO: in the future we will add extra implementation per namespace
    const bucket_owner = object_sdk.get_bucket_owner();
    return {
        name: info.bucket_owner.unwrap(),
        id: info._id,
    };
}
```

```
// in object_sdk.js file - 
get_bucket_owner() {
    .....
}

File: src/sdk/nb.d.ts [hunk 9]
```diff
@@ -399,10 +399,16 @@ interface ObjectMD {
...
+interface ObjectOwner {
+    object_owner_id?: string,
+    object_owner_name?: string;
+}
+
 interface ObjectInfo {
     obj_id: string;
     bucket: string;
     key: string;
+    object_owner: ObjectOwner;
```
romayalon, line 411:
```suggestion
    object_owner?: ObjectOwner;
```

src/sdk/nb.d.ts
```diff
@@ -399,10 +399,16 @@ interface ObjectMD {
     lock_settings: { retention: { mode: string; retain_until_date: Date; }, legal_hold: { status: string } };
 }
 
+interface ObjectOwner {
+    object_owner_id?: string,
```
romayalon, line 403: here you want these properties required

src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -95,6 +95,17 @@ async function get_bucket(req) {
     };
 }
 
+function get_object_owner(obj) {
```
romayalon, line 98: this function should be in s3_utils IMO

romayalon (changes requested):

src/sdk/namespace_nb.js
```diff
@@ -73,6 +81,12 @@ class NamespaceNB {
         return object_sdk.rpc_client.object.list_object_versions(params);
     }
 
+    get_object_owner(info) {
```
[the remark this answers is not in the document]
vh05: NamespaceBlob, not really sure the format of list objects response. Also do not have a setup to verify that. So left it intentionally.

NamespaceFs, Let me reverify.. I saw some out put

src/test/unit_tests/test_s3_ops.js
```diff
@@ -486,6 +486,35 @@ mocha.describe('s3_ops', function() {
...
+                Key: text_file2,
+                CopySource: `/${bucket_name}/${text_file1}`,
+            });
+            await s3.copyObject({
+                Bucket: bucket_name,
+                Key: text_file3,
+                CopySource: `/${BKT5}/${text_file5}`,
+            });
+            const res = await s3.listObjects({ Bucket: bucket_name });
+            console.log(`VINAYAK: ${JSON.stringify(res)}`);
```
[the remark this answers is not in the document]
vh05: done

src/test/system_tests/ceph_s3_tests/test_ceph_s3_deploy.sh
vh05: done

src/sdk/namespace_s3.js
vh05: Both ids are coming same. Let me assing owner id only

src/sdk/namespace_nb.js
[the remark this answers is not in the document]
romayalon: Not sure I understood your point about NamespaceBlob, we will eventually return S3 response

src/sdk/namespace_s3.js
romayalon: owner account id and bucket id is the same?

src/sdk/namespace_s3.js
vh05: I thought of this. Since the suggestion was to source it from namespace cache. Kept like this. In future if we need namespace specific handling, we can do that.

any thoughts here ? or move to object_sdk ?

src/sdk/nb.d.ts
vh05: missed it

src/endpoint/s3/ops/s3_get_bucket.js
vh05: done

src/sdk/nb.d.ts
vh05: done

src/sdk/namespace_nb.js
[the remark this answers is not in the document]
vh05: Okay make sense. I thought it is all together different protocol

src/sdk/namespace_s3.js
vh05: Yes.

`"_id":"668d0822dafa670026921b64","name":"tito-1f04b041-673f-4883-a83c-056bcc6a99af","active_triggers":[],"system_owner":"admin@noobaa.io","bucket_owner":"operator@noobaa.io","bucket_info":{"name":"tito-1f04b041-673f-4883-a83c-056bcc6a99af","owner_account":{"email":"operator@noobaa.io","id":"668d0693dafa670026921b54"}`

src/sdk/namespace_s3.js
romayalon: I just don't like to having the same functionality in multiple places, if one place has a bug we will need to fix it 3 times, I don't see a reason for that..

src/sdk/namespace_s3.js
romayalon: bucket id - `668d0822dafa670026921b64`
owner_account._id - `668d0693dafa670026921b54`
it's not the same

src/sdk/namespace_s3.js
vh05: sure.. I will move this to object_sdk

src/sdk/object_sdk.js
```diff
@@ -181,6 +181,17 @@ class ObjectSDK {
...
 
+    async get_bucket_owner(bucket) {
+        // TODO: in the future we will add extra implementation per namespace
+        let bucket_info;
+        try {
+            bucket_info = await this.read_bucket_sdk_config_info(bucket);
+        } catch (err) {
+            if (err.rpc_code !== 'NO_SUCH_BUCKET') throw err;
+        }
+        return bucket_info;
```
romayalon, line 192: you need to return here the bucket_owner info and not the bucket info
```suggestion
        return {
              name: bucket_info.bucket_owner.unwrap(),
              id: bucket_info.owner_account.id,
          };
```

src/sdk/namespace_s3.js
```diff
@@ -95,9 +95,19 @@ class NamespaceS3 {
...
 
+    async get_object_owner(object_sdk, bucket) {
+        // TODO: in the future we will add extra implementation per namespace
+        const info = await object_sdk.get_bucket_owner(bucket);
+        return {
+              name: info.bucket_owner.unwrap(),
+              id: info.bucket_info.owner_account.id,
+          };
+    }
+
```
romayalon, line 110:
```suggestion
    async get_object_owner(object_sdk, bucket) {
        // TODO: in the future we will add extra implementation per namespace
        return object_sdk.get_bucket_owner(bucket);
    }

```

romayalon: please do that on all namespace

romayalon: please also add JSdoc on all new functions

src/test/system_tests/ceph_s3_tests/s3-tests-lists/s3_tests_pending_list.txt
```diff
@@ -136,4 +136,5 @@ s3tests_boto3/functional/test_s3.py::test_cors_presigned_put_object
 s3tests_boto3/functional/test_s3.py::test_cors_presigned_put_object_with_acl
 s3tests_boto3/functional/test_s3.py::test_cors_presigned_put_object_tenant
 s3tests_boto3/functional/test_s3.py::test_cors_presigned_put_object_tenant_with_acl
-s3tests_boto3/functional/test_s3.py::test_object_presigned_put_object_with_acl_tenant
\ No newline at end of file
+s3tests_boto3/functional/test_s3.py::test_object_presigned_put_object_with_acl_tenant
```
romayalon, line 139: can you please remove this change as well

src/sdk/namespace_s3.js
```diff
@@ -95,9 +95,19 @@ class NamespaceS3 {
             common_prefixes: _.map(res.CommonPrefixes, 'Prefix'),
             is_truncated: res.IsTruncated,
             next_marker: res.NextMarker,
+            object_owner: await this.get_object_owner(object_sdk, params.bucket),
         };
     }
 
+    async get_object_owner(object_sdk, bucket) {
```
romayalon, line 102: please also add to namespace_cache

src/sdk/object_sdk.js
vh05: In case of `namespace_fs`, the `id` is  not part of `bucket_info`. 

get_object_owner function  in `namespace_fs` 

https://github.com/noobaa/noobaa-core/pull/8062/files#diff-74a1674f40effb24da2459330c005cd5ab9fda4e1bca500a4387607926a2af1f

```
    async get_object_owner(object_sdk, bucket) {
        // TODO: in the future we will add extra implementation per namespace
        const info = await object_sdk.get_bucket_owner(bucket);
        return {
              name: info.bucket_owner.unwrap(),
              id: info.bucket_info.owner_account.id,  ================> this part (info.bucket_info.owner_account)
          };
    }
```


get_object_owner function in other namespace_***
https://github.com/noobaa/noobaa-core/pull/8062/files#diff-2243d9c761a2ed2a1a7b8a60ef26f68d78a66cc11851067db6cbe389279b23ea

```
    async get_object_owner(object_sdk, bucket) {
        // TODO: in the future we will add extra implementation per namespace
        const info = await object_sdk.get_bucket_owner(bucket);
        return {
              name: info.bucket_owner.unwrap(),
              id: info.owner_account.id,  =========================>  this part (info.ower_account)
          };
    }
```

So I thought to return general bucket_info from object_sdk and format it in namespacefs.js

src/test/system_tests/ceph_s3_tests/s3-tests-lists/s3_tests_pending_list.txt
vh05: sure

src/sdk/namespace_s3.js
vh05: Sure. I will add the JSdoc.

code changes depend on https://github.com/noobaa/noobaa-core/pull/8062#discussion_r1673766705

src/sdk/namespace_s3.js
vh05: sure

romayalon (approved):
Squash the 2 commits to 1 and merge :)
Thanks for your hard work!

vh05:
> Squash the 2 commits to 1 and merge :) Thanks for your hard work!

Thank you very much @Romy for helping me on this.. It was not possible without you :)

PR #8063, opened 2024-05-21
NSFS | NC | IAM | IAM, S3, And STS Mutual Code Reuse

shirady:
### Explain the changes
1. Reuse constant `CONTENT_TYPE_APP_FORM_URLENCODED` in STS and http_utils
2. Reuse config environment variables in the endpoint file (except `ENDPOINT_SSL_PORT_STS` - see gaps).
3. Move the `authorize_request_account` and `authenticate_request` mutual code and reuse it.

### Issues:
Gaps:
1. In `endpomit.js`, we could not reuse config because `ENDPOINT_SSL_PORT_STS` in config defaults to -1 while in the endpoint it is 7443.
2. Was asked to combine `http_utils.authorize_session_token(req, headers_options);` and `authenticate_request(req);` - but they do not not seems related.

### Testing Instructions:
These tests will only test the flow of the service:
Note: before creating the account, you need to give permission to the `new_buckets_path`: `chmod 777 /tmp/nsfs_root1`.
4. Use AWS CLI to send requests to the IAM service, for example: `s3-nc-user-1-iam iam create-user --user-name Bob --path '/division_abc/subdivision_xyz/'` (more examples in the comment below)

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -301,7 +301,7 @@ async function parse_request_body(req, options) {
             throw new options.ErrorClass(options.error_invalid_body);
         }
     }
-    if (options.body.type.includes('application/x-www-form-urlencoded')) {
+    if (options.body.type.includes(CONTENT_TYPE_APP_FORM_URLENCODED)) {
         try {
             const res = querystring.parse(req.body.toString());
             const renamed = _.mapKeys(res, (value, key) => _.snakeCase(key));
--- a/src/util/signature_utils.js
+++ b/src/util/signature_utils.js
@@ -8,6 +8,8 @@ const url = require('url');
 const path = require('path');
 const crypto = require('crypto');
 const S3Error = require('../endpoint/s3/s3_errors').S3Error;
+const http_utils = require('./http_utils');
+const { RpcError } = require('../rpc');
 
 
 ///////////////////////////////////////
@@ -343,7 +345,36 @@ function authorize_client_request(req) {
     req.headers.Authorization = v4.authorization(req.credentials, req.headers['X-Amz-Date']);
 }
 
+// mutual code of S3, STS and IAM services
+// sdk - this can be: object_sdk, sts_sdk or account_sdk
+function authenticate_request_by_service(req, sdk) {
+    const auth_token = make_auth_token_from_request(req);
+    if (auth_token) {
+        auth_token.client_ip = http_utils.parse_client_ip(req);
+    }
+    if (req.session_token) {
+        auth_token.access_key = req.session_token.assumed_role_access_key;
+        auth_token.temp_access_key = req.session_token.access_key;
+        auth_token.temp_secret_key = req.session_token.secret_key;
+    }
+    sdk.set_auth_token(auth_token);
+    check_request_expiry(req);
+}
+
+// mutual code of S3, STS and IAM services
+function authorize_request_account_by_token(token, requesting_account, is_secret_optional) {
+    const signature_secret = token.temp_secret_key || requesting_account?.access_keys?.[0]?.secret_key?.unwrap();
+    const should_check_signature = !is_secret_optional || signature_secret;
+    if (should_check_signature) {
+        const signature = get_signature_from_auth_token(token, signature_secret);
+        if (token.signature !== signature) throw new RpcError('SIGNATURE_DOES_NOT_MATCH', `Signature that was calculated did not match`);
+    }
+}
+
+
 exports.make_auth_token_from_request = make_auth_token_from_request;
 exports.check_request_expiry = check_request_expiry;
 exports.get_signature_from_auth_token = get_signature_from_auth_token;
 exports.authorize_client_request = authorize_client_request;
+exports.authenticate_request_by_service = authenticate_request_by_service;
+exports.authorize_request_account_by_token = authorize_request_account_by_token;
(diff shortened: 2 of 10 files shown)


src/util/http_utils.js
```diff
@@ -681,6 +682,20 @@ function authorize_session_token(req, options) {
     }
 }
 
+function authenticate_request_by_service(req, sdk_type) {
```
guymguym, line 685: perhaps just pass in the sdk object itself instead of its type?

src/util/services_utils.js
```diff
@@ -0,0 +1,23 @@
...
+const signature_utils = require('./signature_utils');
+const { RpcError } = require('../rpc');
+
+const sdk_type_enum = {
+    OBJECT_SDK: 'object_sdk',
+    STS_SDK: 'sts_sdk',
+    ACCOUNT_SDK: 'account_sdk',
+};
+
+function authorize_request_account_by_token(token, requesting_account) {
```
guymguym, line 13: I think it's better to move this function to signature_utils or endpoint_utils

src/util/services_utils.js
```diff
@@ -0,0 +1,23 @@
...
+    const signature_secret = token.temp_secret_key || requesting_account?.access_keys?.[0]?.secret_key?.unwrap();
+    if (signature_secret) {
```
guymguym, line 15: this condition allows unauthenticated access. I see this is not allowed for iam and sts, only for s3 because s3 also checks bucket policy. So lets add a flag to this function if to allow anonymous requests, and only send true from s3...

src/util/services_utils.js
```diff
@@ -0,0 +1,23 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+
+const signature_utils = require('./signature_utils');
+const { RpcError } = require('../rpc');
+
+const sdk_type_enum = {
+    OBJECT_SDK: 'object_sdk',
+    STS_SDK: 'sts_sdk',
+    ACCOUNT_SDK: 'account_sdk',
```
guymguym, line 10: I would avoid this enum if not needed and just send the actual sdk object to authenticate_request_by_service

src/util/http_utils.js
guymguym: I think it's better to move this function to signature_utils or endpoint_utils

guymguym (changes requested):

src/util/http_utils.js
shirady: 1. I would pass it, although I think it is less readable than the enum.
2. I moved it to `signature_utils` ( I didn't choose `endpoint_utils` is not under `src/util` but `src/endpoint` and its current functions are only used by `endpoint.js`).

src/util/services_utils.js
shirady: I moved it to `signature_utils` ( I didn't choose `endpoint_utils` is not under `src/util` but `src/endpoint` and its current functions are only used by `endpoint.js`).

src/util/services_utils.js
shirady: Thanks, I added a boolean and changed the condition for this.

src/util/services_utils.js
shirady: Changed, thanks.

src/util/signature_utils.js
```diff
@@ -343,7 +345,35 @@ function authorize_client_request(req) {
...
+        auth_token.temp_secret_key = req.session_token.secret_key;
+    }
+    sdk.set_auth_token(auth_token);
+    check_request_expiry(req);
+}
+
+// mutual code of S#, STS and IAM services
+function authorize_request_account_by_token(token, requesting_account, is_anonymous_requests_are_allowed = false) {
+    const signature_secret = token.temp_secret_key || requesting_account?.access_keys?.[0]?.secret_key?.unwrap();
+    if ((is_anonymous_requests_are_allowed && signature_secret) || !is_anonymous_requests_are_allowed) {
```
guymguym, line 367: const should_check_signature = !is_secret_optional || signature_secret;

if (should_check_signature) {
    ...
}

src/util/signature_utils.js
```diff
@@ -343,7 +345,35 @@ function authorize_client_request(req) {
...
+        auth_token.access_key = req.session_token.assumed_role_access_key;
+        auth_token.temp_access_key = req.session_token.access_key;
+        auth_token.temp_secret_key = req.session_token.secret_key;
+    }
+    sdk.set_auth_token(auth_token);
+    check_request_expiry(req);
+}
+
+// mutual code of S#, STS and IAM services
+function authorize_request_account_by_token(token, requesting_account, is_anonymous_requests_are_allowed = false) {
```
guymguym, line 365: rename and no default -
```suggestion
function authorize_request_account_by_token(token, requesting_account, is_secret_optional) {
```

guymguym (commented):
two comments + rebase

src/util/signature_utils.js
```diff
@@ -343,7 +345,35 @@ function authorize_client_request(req) {
     req.headers.Authorization = v4.authorization(req.credentials, req.headers['X-Amz-Date']);
 }
 
+// mutual code of S#, STS and IAM services
```
naveenpaul1, line 348: small comment: change s# to s3, comments

guymguym (changes requested):
looks good. just squash commits before merging

guymguym (approved):

PR #8068, opened 2024-05-23
Block access_keys updation if duplicate access_key is found

aayushchouhan09:
### Explain the changes
1.  Added a function to check if access_key provided for updation is already available for an account, throws an Rpc error if duplication is found.

1. Gap #7972

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -804,6 +804,12 @@ class SystemStore extends EventEmitter {
         }
     }
 
+    get_account_by_access_key(access_key_id) {
+        return _.find(this.data.accounts, acc =>
+            _.find(acc.access_keys, key => key.access_key.unwrap() === access_key_id.unwrap())
+        );
+    }
+
     get_accounts_by_nsfs_account_config(nsfs_account_config) {
         if (this.data && !_.isEmpty(this.data.accounts)) {
             return this.data.accounts.filter(account => account.nsfs_account_config &&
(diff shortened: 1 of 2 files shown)


src/server/system_services/system_store.js
```diff
@@ -804,6 +804,12 @@ class SystemStore extends EventEmitter {
         }
     }
 
+    check_for_access_key_duplication(access_key_id) {
+        return _.some(this.data.accounts, acc =>
+            _.some(acc.access_keys, key => key.access_key.unwrap() === access_key_id.unwrap())
+        );
+    }
```
dannyzaken, line 811: I think a function like `check_for_access_key_duplication` is too specific to implement in the system_store. 
you can change it to `get_account_by_access_key` that is more generic, or use `find` directly in `account_server` without adding a function.

```suggestion
    get_account_by_access_key(access_key_id) {
        return _.find(this.data.accounts, acc =>
        acc.access_keys && acc.access_keys[0].access_key.unwrap() === access_key_id.unwrap()
    );
    }
```

shirady: @dannyzaken For more general code we can iterate all the access keys in the array of `access_keys` and not just the first element (although today we replace the whole array with an array with one `access_keys`). WDYT?

dannyzaken (approved):

PR #8074, opened 2024-05-23
LIFECYCLE - fix issue with delete only 3K objects a day

jackyalbo:
### Explain the changes
1. delete_multiple_objects_by_filter will return the number of objects deleted
2. lifecycle will run every 5 minutes if any of its rules has deleted the max objects it can (1K - meaning he can delete more and we shouldn't wait to the next cycle)

1. Fixed https://bugzilla.redhat.com/show_bug.cgi?id=2279964

### Testing Instructions:
1. create more than 3K of objects on a bucket (let's say 3.5K)
2. set this lifecycle rule: 
```
{
    "Rules": [{
          "Expiration": {
                "Days": 1
            },
            "ID": "data-expire-withoutprefix",
            "Filter": {
                "Prefix": ""
            },
            "Status": "Enabled"
        }
    ]
}
```
3. wait till LIFECYCLE runs a day after - see that it deleted all the objects and not just 1K object every 8 hours.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -1154,8 +1154,8 @@ module.exports = {
             reply: {
                 type: 'object',
                 properties: {
-                    is_empty: {
-                        type: 'boolean'
+                    num_objects_deleted: {
+                        type: 'integer'
                     }
                 }
             },
--- a/src/server/bg_services/lifecycle.js
+++ b/src/server/bg_services/lifecycle.js
@@ -9,10 +9,7 @@ const dbg = require('../../util/debug_module')(__filename);
 const server_rpc = require('../server_rpc');
 const system_store = require('../system_services/system_store').get_instance();
 const auth_server = require('../common_services/auth_server');
-
-const LIFECYCLE = {
-    schedule_min: 5 * 1000 * 60 // run every 5 minutes
-};
+const config = require('../../../config');
 
 function get_expiration_timestamp(expiration) {
     if (!expiration) {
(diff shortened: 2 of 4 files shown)


src/api/object_api.js
```diff
@@ -1156,6 +1156,9 @@ module.exports = {
                 properties: {
                     is_empty: {
                         type: 'boolean'
+                    },
+                    objects_deleted: {
```
liranmauda, line 1160: maybe call it `num_objects_deleted` ? 
```suggestion
                    num_objects_deleted: {
```

src/server/bg_services/lifecycle.js
```diff
@@ -57,9 +60,12 @@ async function handle_bucket_rule(system, rule, j, bucket) {
     });
 
     bucket.lifecycle_configuration_rules[j].last_sync = Date.now();
+    if (res.objects_deleted >= limit) should_rerun = true;
```
liranmauda, line 63: Can it be bigger then?

src/server/object_services/object_server.js
```diff
@@ -970,7 +970,7 @@ async function delete_multiple_objects_by_filter(req) {
         }
     }));
     const bucket_has_objects = await MDStore.instance().has_any_objects_for_bucket(bucket_id);
-    return { is_empty: !bucket_has_objects };
+    return { is_empty: !bucket_has_objects, objects_deleted: objects.length };
```
liranmauda, line 973: I wonder, Do we even need the is_epmty? 
If we don't we can keep it and open a gap. (or remove it in this PR).

liranmauda (approved):
LGTM, just some small comments.

src/server/object_services/object_server.js
jackyalbo: I will just remove

src/server/bg_services/lifecycle.js
jackyalbo: not supposed to - just in case

PR #8090, opened 2024-05-28
NC | CLI | CLI errors handling when master key is invalid

romayalon:
### Explain the changes
1. **manage_nsfs.js -** 
- Account Delete - skipped the decryption in this case as we might want to delete accounts even if master key has issues.
- Account status --show_secrets / Account update - 

2. **manage_nsfs_cli_errors.js -** Created a new CLI error InvalidMasterKey and added a mapping between RPC INVALID_MASTER_KEY to InvalidMasterKey.

3. **nc_master_key_manager.js -** threw INVALID_MASTER_KEY and changed the error messages to be more accurate.

4. **test_nc_account_invalid_mkm_intergration.test.js -** added integration tests for scenarios of CLI account functions and missing/corrupted master key file.

5. **test_nc_master_keys.test.js** - updated errors on tests and added a new empty file case.

### Issues: Fixed #xxx / Gap #xxx
1. Fixed #8083
2. Fixed #8084 

### Testing Instructions:
1. Automatic -
```
sudo jest --testRegex=jest_tests/test_nc_master_keys.test.js
sudo jest --testRegex=jest_tests/test_nc_account_invalid_mkm
```
Manual - 
There are many scenarios for testing it - 
1. `sudo node src/cmd/manage_nsfs account add --name=account1 --allow_bucket_creation=true --uid=0 --gid=0  --new_buckets_path=/tmp/new_buckets_path1/`
2. delete or corrupt /etc/noobaa.conf.d/master_keys.json
3. Run `sudo node src/cmd/manage_nsfs account update --name account1 --uid=1111` and expect the following error instead of an internal error - 
```
{
  "error": {
    "code": "InvalidMasterKey",
    "message": "Master key manager had issues loading master key, can not decrypt/encrypt secrets.",
[...]

PR #8097, opened 2024-05-29
NC | Avoid concurrency master key manager initialization

romayalon:
### Explain the changes
1. manage_nsfs.js -

- account_management() - call nc_mkm.init() - in order to avoid concurrent calls master key manager initialization that might cause race conditions. call init() only if action is account add/update or show_secrets flag was provided (show_secrets can be passed in account list/status).
- list_config_files() - inside the loop - decrypt access keys only if show_secrets is set to true - reduce calls to nc_mkm.init() which will reduce calls for reading/getting the master keys json.

2. nc_master_key_manager.js - add prints for debugging.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -292,6 +292,10 @@ async function account_management(action, user_input) {
         user_input.name = config.ANONYMOUS_ACCOUNT_NAME;
         user_input.email = config.ANONYMOUS_ACCOUNT_NAME;
     }
+    // init nc_mkm here to avoid concurrent initializations
+    // init if actions is add/update (require encryption) or show_secrets = true (require decryption)
+    if ([ACTIONS.ADD, ACTIONS.UPDATE].includes(action) || show_secrets) await nc_mkm.init();
+
     const data = await fetch_account_data(action, user_input);
     await manage_account_operations(action, data, show_secrets, user_input);
 }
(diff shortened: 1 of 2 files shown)


src/manage_nsfs/nc_master_key_manager.js
```diff
@@ -79,6 +79,8 @@ class NCMasterKeysManager {
      */
     _set_keys(master_keys) {
         if (!master_keys.active_master_key || !master_keys.master_keys_by_id) throw new RpcError('INVALID_MASTER_KEYS_FILE', 'Invalid master_keys.json file');
+        dbg.log0('_set_keys: master_keys', util.inspect(master_keys));
```
guymguym, line 82: I don't think we want to log the actual keys, right? 
or is it wrapped in SensitiveString?
Even if so I would still avoid it to prevent mistakes, and only selectively print information about the master keys such as their ids / timestamps etc.

guymguym (changes requested):

src/cmd/manage_nsfs.js
```diff
@@ -622,6 +622,9 @@ async function list_config_files(type, config_path, wide, show_secrets, filters)
     const entries = await nb_native().fs.readdir(fs_context, config_path);
     const should_filter = Object.keys(filters).length > 0;
 
+    // we init nc_mkm here to avoid concurrent initializations
+    if (type === TYPES.ACCOUNT) await nc_mkm.init();
```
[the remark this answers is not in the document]
romayalon: yes I did it in [8090 PR](https://github.com/noobaa/noobaa-core/pull/8090/files#diff-1b4d8c2974ba0dfcf75223cbe7d90911c8bf1397b544895390f8f9d039a0e60fR341)

src/manage_nsfs/nc_master_key_manager.js
romayalon: @guymguym ofc not, all these prints are temporary debugging prints for the RPM I built for scale team, removing it
(see PR description - temporary prints - will remove before merging)

src/cmd/manage_nsfs.js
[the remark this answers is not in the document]
guymguym: but it should affect this init call too otherwise it will init on delete accounts here?

src/manage_nsfs/nc_master_key_manager.js
romayalon: @guymguym thanks to your comment I found a place where we printed the active_master_key so I removed it!

src/cmd/manage_nsfs.js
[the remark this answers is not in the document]
romayalon: not sure I understand how this is related to delete account, can you please explain?
BTW, I added show_secrets = true to the conditions of init/descrypt here so we will avoid it if not needed

src/manage_nsfs/nc_master_key_manager.js
romayalon: BTW, I decided to keep some of the prints eventually

guymguym (approved):

PR #8101, opened 2024-05-30
Update stale actions with exempt rules

nimrod-becker:
### Explain the changes
1. Add a 2nd more lenient threshold for getting issues and PRs to a stale state (from stale to close does not change)
2. Add exempt rules to our stale action
  2.1 Special Labels (Question, Feature Technical Debt) will not be handled by the regular threshold.
  2.2 Issues and PRs with Milestones attached will not be handled by the regular threshold.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -18,8 +18,12 @@ jobs:
       pull-requests: write
 
     steps:
+    # Handle non exempted issues
     - uses: actions/stale@v5
       with:
+        exempt-issue-labels: 'Type:Enhancement, Type:Question, Type:Technical Debt'
+        exempt-all-milestones: true
+
         days-before-stale: 90
         days-before-pr-stale: 90
         days-before-issue-stale: 90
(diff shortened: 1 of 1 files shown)


.github/workflows/stale.yml
```diff
@@ -31,3 +35,16 @@ jobs:
         days-before-issue-close: 30
         close-pr-message: 'This PR is stale and had no activity for too long - it will now be closed.'
         close-issue-message: 'This issue is stale and had no activity for too long - it will now be closed.'
+
+  # Handle exempted issues
+    - uses: actions/stale@v5
+      with:
+
+        days-before-stale: 180        
+        days-before-issue-stale: 180
```
guymguym, line 44: where is the days-before-pr-stale? is it intentionally missing here?

nimrod-becker: PRs don't have the labels to make them exempted so they will always be processed by thee 1st rule

guymguym (approved):

PR #8106, opened 2024-06-03
[NSFS | Glacier] Add support for returning RestoreStatus in listObjects

### LINKED ISSUE #8048: NSFS GLACIER optional return RestoreStatus in ListObjects

ListObjects and ListObjectsV2 have added an optional header: 

`x-amz-optional-object-attributes`
   Specifies the optional fields that you want returned in the response. Fields that you do not specify are not returned.
   Valid Values: RestoreStatus

 This allows the caller to get back a RestoreStatus per object in the list response.

NSFS GLACIER should allow to return the restore status info.

NA

tangledbytes:
### Explain the changes
This PR adds support for returning `RestoreStatus` in `ListObjects` APIs when requested via the header `x-amz-optional-object-attributes`.

Fixes: #8048 

### Testing Instructions:

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_get_bucket.js
+++ b/src/endpoint/s3/ops/s3_get_bucket.js
@@ -7,10 +7,10 @@ const S3Error = require('../s3_errors').S3Error;
 const s3_utils = require('../s3_utils');
 
 /**
- * list objects and list objects V2: 
+ * list objects and list objects V2:
  * https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjects.html
  * https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html
- * 
+ *
  * note: the original documentation was in the below link:
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketGET.html
  * (but anyway it is permanently redirected to list object link above)
@@ -26,6 +26,16 @@ async function get_bucket(req) {
     const list_type = req.query['list-type'];
     const cont_tok = req.query['continuation-token'];
     const start_after = req.query['start-after'];
+
+    const optional_object_attributes = req.headers['x-amz-optional-object-attributes'];
+    const restore_status_requested = optional_object_attributes === 'RestoreStatus';
+
+    // Only RestoreStatus is a valid attribute for now
+    if (optional_object_attributes && !restore_status_requested) {
+        // S3 API fails with `InvalidArgument` and this message
+        throw new S3Error({ ...S3Error.InvalidArgument, message: 'Invalid attribute name specified' });
+    }
+
 
     const params = {
         bucket: req.params.bucket,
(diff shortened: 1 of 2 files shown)


src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -27,6 +27,16 @@ async function get_bucket(req) {
...
+    const optional_object_attributes = req.headers['x-amz-optional-object-attributes'];
+
+    // Only RestoreStatus is a valid attribute for now
+    if (optional_object_attributes && optional_object_attributes !== 'RestoreStatus') {
+        // S3 API fails with `InvalidArgument` and this message
+        throw new S3Error({ ...S3Error.InvalidArgument, message: 'Invalid attribute name specified' });
+    }
+
+    const restore_status_requested = optional_object_attributes === 'RestoreStatus';
+
```
guymguym, line 39:
```suggestion
    const optional_object_attributes = req.headers['x-amz-optional-object-attributes'];
    const restore_status_requested = optional_object_attributes === 'RestoreStatus';

    // Only RestoreStatus is a valid attribute for now
    if (optional_object_attributes && !restore_status_requested) {
        // S3 API fails with `InvalidArgument` and this message
        throw new S3Error({ ...S3Error.InvalidArgument, message: 'Invalid attribute name specified' });
    }


```

src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -73,8 +83,21 @@ async function get_bucket(req) {
...
+            if (restore_status_requested && obj.restore_status) {
+                const restore_status = obj.restore_status;
+                contents.RestoreStatus = {
+                    IsRestoreInProgress: restore_status.ongoing,
+                };
+
+                if (!restore_status.ongoing && restore_status.expiry_time) {
+                    contents.RestoreStatus.RestoreExpiryDate = new Date(restore_status.expiry_time).toUTCString();
+                }
```
guymguym, line 96: can you move this logic to its own function?
it can simply return undefined to have no RestoreStatus element.
this will just leak less details to the calling context - something like
```
RestoreStatus: get_object_restore_status(obj, restore_status_requested),
```

guymguym (changes requested):
Also, is there a test that can be added to call with this header?

src/endpoint/s3/ops/s3_get_bucket.js
tangledbytes: A separate function in `s3_get_bucket`? Or something in `object_sdk` which forwards the request to `namespace_fs`?

src/endpoint/s3/ops/s3_get_bucket.js
```diff
@@ -63,7 +73,6 @@ async function get_bucket(req) {
             }),
         },
         _.map(reply.objects, obj => ({
-            Contents: {
```
guymguym, line 66: is this deletion correct??

guymguym (approved):
looks great! 

worth peeking if there is any ceph s3 test case that can now be enabled...

PR #8117, opened 2024-06-06
NC | CLI | trimmed output fix

### LINKED ISSUE #8111: Noobaa command to list bucket/account giving trimmed output in pipe
### Environment info
- NooBaa Version: **5.15.0/1/2/3**
- Platform: **other: RHEL9**

### Actual behavior
1. Bucket/account details output getting trimmed when using with pipe

### Expected behavior
1. it should give full output in pipe

### Steps to reproduce
1.

romayalon:
### Explain the changes
1. A follow up to https://github.com/noobaa/noobaa-core/pull/7898, doing the same changes but for noobaa_cli.
2. Had to move the stdout write on error to the catch scope of main() in manage_nsfs due to https://github.com/noobaa/noobaa-core/issues/7894.

1. Fixed #8111 
2. We should consider making the same changes of bullet 2 for success flow (write_stdout_output).

### Testing Instructions:
1. `env LC_ALL=C /usr/local/bin/noobaa-cli bucket 2>/dev/null list --wide`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -79,7 +79,11 @@ async function main(argv = minimist(process.argv.slice(2))) {
         const user_input = user_input_from_file || argv;
         config_root = argv.config_root ? String(argv.config_root) : config.NSFS_NC_CONF_DIR;
         if (!config_root) throw_cli_error(ManageCLIError.MissingConfigDirPath);
-        if (argv.config_root) config.NSFS_NC_CONF_DIR = String(argv.config_root);
+        if (argv.config_root) {
+            config.NSFS_NC_CONF_DIR = String(argv.config_root);
+            config.load_nsfs_nc_config();
+            config.reload_nsfs_nc_config();
+        }
 
         accounts_dir_path = path.join(config_root, accounts_dir_name);
         access_keys_dir_path = path.join(config_root, access_keys_dir_name);
@@ -102,10 +106,13 @@ async function main(argv = minimist(process.argv.slice(2))) {
     } catch (err) {
         dbg.log1('NSFS Manage command: exit on error', err.stack || err);
         const manage_err = ((err instanceof ManageCLIError) && err) ||
-            new ManageCLIError(ManageCLIError.FS_ERRORS_TO_MANAGE[err.code] ||
+            new ManageCLIError({
+                ...(ManageCLIError.FS_ERRORS_TO_MANAGE[err.code] ||
                 ManageCLIError.RPC_ERROR_TO_MANAGE[err.rpc_code] ||
-                ManageCLIError.InternalError);
-        throw_cli_error(manage_err, err.stack || err);
+                ManageCLIError.InternalError), cause: err });
+        process.stdout.write(manage_err.to_string() + '\n', () => {
+            process.exit(1);
+        });
     }
 }
 
@@ -256,7 +263,8 @@ async function delete_bucket(data, force) {
         if (object_entries.length === 0 || force) {
             await native_fs_utils.folder_delete(bucket_temp_dir_path, fs_context_fs_backend, true);
             await native_fs_utils.delete_config_file(fs_context_config_root_backend, buckets_dir_path, bucket_config_path);
-            write_stdout_response(ManageCLIResponse.BucketDeleted, '', {bucket: data.name});
+            write_stdout_response(ManageCLIResponse.BucketDeleted, '', { bucket: data.name });
+            return;
         }
         throw_cli_error(ManageCLIError.BucketDeleteForbiddenHasObjects, data.name);
     } catch (err) {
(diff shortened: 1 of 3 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -102,10 +102,13 @@ async function main(argv = minimist(process.argv.slice(2))) {
...
         const manage_err = ((err instanceof ManageCLIError) && err) ||
-            new ManageCLIError(ManageCLIError.FS_ERRORS_TO_MANAGE[err.code] ||
+            new ManageCLIError({
+                ...(ManageCLIError.FS_ERRORS_TO_MANAGE[err.code] ||
                 ManageCLIError.RPC_ERROR_TO_MANAGE[err.rpc_code] ||
-                ManageCLIError.InternalError);
-        throw_cli_error(manage_err, err.stack || err);
+                ManageCLIError.InternalError), detail: err.stack || err });
+        process.stdout.write(manage_err.to_string() + '\n', () => {
+            process.exit(0);
```
guymguym, line 110: exit 1

src/cmd/manage_nsfs.js
```diff
@@ -102,10 +102,13 @@ async function main(argv = minimist(process.argv.slice(2))) {
     } catch (err) {
         dbg.log1('NSFS Manage command: exit on error', err.stack || err);
         const manage_err = ((err instanceof ManageCLIError) && err) ||
-            new ManageCLIError(ManageCLIError.FS_ERRORS_TO_MANAGE[err.code] ||
+            new ManageCLIError({
+                ...(ManageCLIError.FS_ERRORS_TO_MANAGE[err.code] ||
                 ManageCLIError.RPC_ERROR_TO_MANAGE[err.rpc_code] ||
-                ManageCLIError.InternalError);
-        throw_cli_error(manage_err, err.stack || err);
+                ManageCLIError.InternalError), detail: err.stack || err });
```
guymguym, line 108: i think the detail here is not needed because it repeats the message

guymguym (changes requested):

src/cmd/manage_nsfs.js
guymguym: oh you want it for the stacktrace? in that case, perhaps just add the stack trace to the to_string() function itself - it can get the stack from this.stack and then it can go into a stack field and not details...

src/cmd/manage_nsfs.js
romayalon: oops thanks

src/cmd/manage_nsfs.js
romayalon: yes 👍🏻 I added it to the to_string()

src/manage_nsfs/manage_nsfs_cli_errors.js
```diff
@@ -16,18 +18,21 @@ class ManageCLIError extends Error {
     /**
      * @param {ManageCLIErrorSpec} error_spec 
      */
-    constructor({ code, message, http_code }) {
+    constructor({ code, message, http_code, detail, stack }) {
         super(message); // sets this.message
         this.code = code;
         this.http_code = http_code;
+        this.detail = detail;
+        this.stack = stack;
```
guymguym, line 26: oh now I understand that you needed the original error stack and not this error stack.
so in js Errors have a `cause` attribute that can be set with new Error(message, { cause: another_error }).
we can pass that optional cause to `super(message, { cause })` in beginning of ctor,
and then to_string() can return `cause: this.cause.stack || this.cause.message`

romayalon: sure, stack vs cause - 
```
cause -

{
  "error": {
    "code": "InternalError",
    "message": "The server encountered an internal error. Please retry the request",
    "cause": "ReferenceError: validate_account_args1 is not defined\n    at get_account_status (noobaa-core/src/cmd/manage_nsfs.js:539:5)\n    at manage_account_operations (noobaa-core/src/cmd/manage_nsfs.js:561:15)\n    at account_management (noobaa-core/src/cmd/manage_nsfs.js:308:11)\n    at async main (noobaa-core/src/cmd/manage_nsfs.js:95:13)"
  }
}

stack -

{
  "error": {
    "code": "InternalError",
    "message": "The server encountered an internal error. Please retry the request",
    "stack": "ReferenceError: validate_account_args1 is not defined\n    at get_account_status (noobaa-core/src/cmd/manage_nsfs.js:539:5)\n    at manage_account_operations (noobaa-core/src/cmd/manage_nsfs.js:561:15)\n    at account_management (noobaa-core/src/cmd/manage_nsfs.js:308:11)\n    at async main (noobaa-core/src/cmd/manage_nsfs.js:95:13)"
  }
}
```

src/manage_nsfs/manage_nsfs_cli_errors.js
```diff
@@ -16,18 +18,21 @@ class ManageCLIError extends Error {
...
      * @param {ManageCLIErrorSpec} error_spec 
      */
-    constructor({ code, message, http_code }) {
-        super(message); // sets this.message
+    constructor({ code, message, http_code, detail, cause }) {
+        super(message, { cause });
         this.code = code;
         this.http_code = http_code;
+        this.detail = detail;
+        this.cause = cause;
```
guymguym, line 26: No need to set this.cause because the standard Error ctor which is called in super() will do that.

guymguym (approved):
one small comment, otherwise looks good

PR #8119, opened 2024-06-09
NSFS | NC | IAM Service - Access Keys CRUD API Implementation

shirady:
### Explain the changes
1. Implement the functions related to access keys in `AccountSpaceFS` (CRUD).
2. Add more properties to `nsfs_account_schema` inside `access_key` object (all of them are not required):
`creation_date` = the date the access-key object was created.
`deactivated` = boolean (we need to translate it to status which is 'Active' or 'Inactive').
3. Clean the account cache after updating the account config using the IAM API.
4. Remove unused errors that were copied from STS' errors in `IamErrors`.
5. Change the param name from `user_name` to `username` in the ops.
6. Remove the mock variables that we used in the IAM boilerplate implementation.
7. Update unit test for access_keys function in `accountspace_fs`.
8. Add unit test in `account_schema_validation` related to the schema changes.
9. Update `Get Started` section that would be the demo for IAM access keys management (`docs/dev_guide/nc_nsfs_iam_developer_doc.md`).
10. Changes in `nc_master_key_manager` to allow an access key object with more properties and add tests for the edited functions (`encrypt_access_keys` and `decrypt_access_keys` of account).
11. Add the support for denying a request whose access key status is `Inactive` (in all services: S3, STS, IAM).
12. When creating a new user instead of copying the `master_key_id` from the root account, we now take the active `master_key_id`.
13. Add a new RPC code for a deactivated access key.
14. Edit the function `authorize_request_account_by_token` adding the access key deactivated check, but also throwing errors on every missing component instead of using optional chaining.

### Issues:
List of GAPs:
1. Parsing and validating the params (for example: username).
2. No pagination in the `list_access_keys` implementation.
3. No `NoobaaEvent` at this point.
4. In `get_access_key_last_used` we send dummy values (`region`, `last_used_date`, `service_name`).
5. Change the `IamError` class to have a template message.

#### Unit Tests
Please run:
1. `sudo npx jest test_accountspace_fs.test.js`
2. `npx jest test_iam_utils.test.js`
3. `npx jest test_nc_nsfs_account_schema_validation.test.js`
4. `sudo npx jest test_nc_master_keys.test.js` (the teardown removes what you have in `/etc/noobaa.conf.d`).

#### Manual Tests
IAM changes in NC NSFS
Currently, we do not validate the input, so the test should use only valid input.
5. Use AWS CLI to send requests to the IAM service, for example:
 `s3-nc-user-1-iam iam create-user --user-name Bob --path '/division_abc/subdivision_xyz/'`
 `s3-nc-user-1-iam iam create-access-key --user-name Bob`
 `s3-nc-user-1-iam iam get-access-key-last-used --access-key-id <access-key>`
 `s3-nc-user-1-iam iam update-access-key --access-key-id <access-key> --user-name Bob --status Inactive`
 `s3-nc-user-1-iam iam delete-access-key --access-key-id <access-key> --user-name Bob`
 `s3-nc-user-1-iam iam list-access-keys --user-name Bob`

Other
Since I changed the function `authorize_request_account_by_token` I also tested it in a containerized environment with the operator.

- [X] Doc added/updated
- [X] Tests added


guymguym (changes requested):
please squash commits

src/sdk/accountspace_fs.js
```diff
@@ -510,30 +639,158 @@ class AccountSpaceFS {
...
+        if (this._check_root_account(requesting_account)) {
+            requester = {
+                name: requesting_account_name,
+                identity: identity_enum.ROOT_ACCOUNT
+            };
+            is_root_account_or_user_on_itself = true;
+            return { is_root_account_or_user_on_itself, requester};
+        }
+        // user (on himself) - username can be undefined
+        if (_.isUndefined(username) || requesting_account_name === username) {
```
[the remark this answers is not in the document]
shirady: I changed it (even though I like the  `_.isUndefined` more than `=== undefined` because I think it is more readable and covers null case...).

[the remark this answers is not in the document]
guymguym: it doesn't cover null afaik

guymguym:
@shirady please squash to be ready to merge

src/sdk/account_sdk.js
```diff
@@ -73,11 +74,18 @@ class AccountSDK {
...
-    // copied from function in sts_sdk
     authorize_request_account(req) {
         const token = this.get_auth_token();
         // If the request is signed (authenticated)
         if (token) {
+            try {
+                signature_utils.check_access_key_status(token.access_key, this.requesting_account);
+            } catch (err) {
+                // GAP - until we change the IamError class to have templates
+                const message_with_details = `The security token included in the request is invalid`;
```
guymguym, line 85: there are no details in this message... why not have it in the error spec and just map it through our usual error code mappings?

src/server/system_services/schemas/nsfs_account_schema.js
```diff
@@ -60,6 +60,13 @@ module.exports = {
                     encrypted_secret_key: {
                         type: 'string',
                     },
+                    creation_date: {
+                        type: 'string',
+                    },
+                    // we would need to translate the is_active to 'Active' and 'Inactive' strings
+                    is_active: {
+                        type: 'boolean',
+                    },
```
guymguym, line 69: since we consider this property being absent to mean it is implicitly active, I would prefer this prop to have the opposite meaning - such as `disabled` or `deactivated`.

src/util/signature_utils.js
```diff
@@ -372,9 +372,34 @@ function authorize_request_account_by_token(token, requesting_account, is_secret
...
+// and is_active is false we will not continue to work on the request and immediately throw an error
+// access_key_id_to_find is a string (not SensitiveString)
+// inside requesting_account the access keys ids are SensitiveStrings
+function check_access_key_status(access_key_id_to_find, requesting_account) {
+    const access_keys = requesting_account.access_keys;
+    const index = _.findIndex(access_keys, item => item.access_key.unwrap() === access_key_id_to_find);
+    if (index < 0) {
+        // we should never get here
+        // in case we did, we need to debug the issue
+        dbg.warn(`check_access_key_status: could not find access_key_id ${access_key_id_to_find}`);
```
guymguym, line 385: so why are we just continuing? if this is not mean to be allowed it should throw in such a sensitive flow.

src/util/signature_utils.js
```diff
@@ -372,9 +372,34 @@ function authorize_request_account_by_token(token, requesting_account, is_secret
 }
 
 
+// In case the object access_keys in requesting_account has the is_active property
+// and is_active is false we will not continue to work on the request and immediately throw an error
+// access_key_id_to_find is a string (not SensitiveString)
+// inside requesting_account the access keys ids are SensitiveStrings
+function check_access_key_status(access_key_id_to_find, requesting_account) {
+    const access_keys = requesting_account.access_keys;
+    const index = _.findIndex(access_keys, item => item.access_key.unwrap() === access_key_id_to_find);
```
guymguym, line 381: using find only in this flow seems missing because the other functions like `authorize_request_account_by_token` are just looking at key at index 0. shouldn't we add a function to look up an access key pair from the account and use it instead in the other flows?

src/util/signature_utils.js
```diff
@@ -372,9 +372,34 @@ function authorize_request_account_by_token(token, requesting_account, is_secret
 }
 
 
+// In case the object access_keys in requesting_account has the is_active property
+// and is_active is false we will not continue to work on the request and immediately throw an error
+// access_key_id_to_find is a string (not SensitiveString)
+// inside requesting_account the access keys ids are SensitiveStrings
+function check_access_key_status(access_key_id_to_find, requesting_account) {
```
guymguym, line 379: Not sure why we need this function being called separately from `authorize_request_account_by_token`, it would be best if it would call it inside to check the access key it uses for the signature checks.

guymguym (changes requested):
see comments posted above

src/manage_nsfs/nc_master_key_manager.js
```diff
@@ -340,13 +351,29 @@ class NCMasterKeysManager {
...
+        // we will iterate the array and add the secret_key
+        // and remove the encrypted_secret_key
+        const decrypted_access_keys = [];
+        const size = account.access_keys?.length ?? 0;
+        for (let index = 0; index < size; index++) {
+            const access_keys = { ...account.access_keys[index] };
+            delete access_keys.encrypted_secret_key;
+            access_keys.secret_key = decrypted_access_keys_basic[index].secret_key;
+            decrypted_access_keys[index] = access_keys;
+        }
```
romayalon, line 368: @shirady why another iteration is needed? can't we just put it inside P.all() line?
```
const decrypted_access_keys = await P.all(_.map(account.access_keys, async access_keys => {
      const decrypted_access_keys = { ...account.access_keys };
      decrypted_access_keys.secret_key = await this.decrypt(access_keys.encrypted_secret_key, 
                  account.master_key_id);
      delete decrypted_access_keys.encrypted_secret_key;
      return decrypted_access_keys;
}));

File: src/util/signature_utils.js (hunk 4 above)
Comment (shirady) [reply src/util/signature_utils.js:385]: I will change it and throw an error.

File: src/manage_nsfs/nc_master_key_manager.js (hunk 7 above)
Comment (shirady) [reply src/manage_nsfs/nc_master_key_manager.js:368]: I didn't find the way yet...
I want to avoid adding the properties directly in the first loop, and when I tried to move it all to a separate function I couldn't use the `this`.

Comment (romayalon) [reply src/manage_nsfs/nc_master_key_manager.js:368]: for what reason you want to avoid it?

Comment (shirady) [reply src/manage_nsfs/nc_master_key_manager.js:368]: I want to avoid adding the properties directly in the first loop for 2 reasons:
1. In case it is `undefined` (the account doesn't have these properties) I do not want to set the key with `undefined` (if I do I would need to omit those properties).
2. To let us the flexibility to add properties in the schema without changing these functions again.

File: src/server/system_services/schemas/nsfs_account_schema.js [hunk 8]
```diff
@@ -50,6 +50,7 @@ module.exports = {
         },
         access_keys: {
             type: 'array',
+            maxItems: 2,
```
guymguym, line 53: I wouldn't put this in schema. We might want to have more deactivated items in the list.

src/util/schema_utils.js
```diff
@@ -4,7 +4,7 @@
 const _ = require('lodash');
 const util = require('util');
 
-const COMMON_SCHEMA_KEYWORDS = ['doc', '$id'];
+const COMMON_SCHEMA_KEYWORDS = ['doc', '$id', 'maxItems'];
```
guymguym, line 7: doesn't seem like this is in the right level, it should be in the array type case only

src/sdk/account_sdk.js
```diff
@@ -73,11 +74,15 @@ class AccountSDK {
...
 
-    // copied from function in sts_sdk
     authorize_request_account(req) {
         const token = this.get_auth_token();
         // If the request is signed (authenticated)
         if (token) {
+            try {
+                signature_utils.check_access_key_status(token.access_key, this.requesting_account);
+            } catch (err) {
+                throw new IamError(IamError.InvalidClientTokenIdInactiveAccessKey);
```
guymguym, line 84: why is this here and not in RPC_ERRORS_TO_IAM in iam_rest? I don't think we need to catch it here at all.

src/server/system_services/schemas/nsfs_account_schema.js
shirady: I will remove it.
I added it because I want to restrict the access_keys array from having more than 2 items, in what case do you think would we want to have more than 2 objects?

src/manage_nsfs/nc_master_key_manager.js
romayalon: @shirady 
1. just skip the description if secret_key and access_key are not defined
2. this row will do the flexibility - `const decrypted_access_keys = { ...account.access_keys };`

```
const decrypted_access_keys = await P.all(_.map(account.access_keys, async access_keys => {
      const decrypted_access_keys = { ...account.access_keys };
      if (decrypted_access_keys.encrypted_secret_key) {
            decrypted_access_keys.secret_key = await this.decrypt(access_keys.encrypted_secret_key, 
                   account.master_key_id);
            delete decrypted_access_keys.encrypted_secret_key;
      }
      return decrypted_access_keys;
}));
```

src/util/signature_utils.js
shirady: I will change it to be part of `authorize_request_account_by_token`.

src/util/schema_utils.js
shirady: Since I will not add this property to the schema `maxItems`, this change will be canceled anyway.

src/sdk/account_sdk.js
shirady: It is there, but I mapped the `INVALID_ACCESS_KEY_ID` to `IamError.InvalidClientTokenId` and not to `IamError.InvalidClientTokenIdInactiveAccessKey`, and then the message will look different.

shirady:
moving to draft since there is a commit in master that I had to revert (investigating).

src/manage_nsfs/nc_master_key_manager.js
shirady: I will use it, thank you.

src/sdk/account_sdk.js
guymguym: So just throw another RPC error code for it.

src/server/system_services/schemas/nsfs_account_schema.js
guymguym: It's a superficial limit. we can limit it in the api if needed, but overall there shouldn't be anything that prevents us from increasing that limit in the api and everything should work fine.

src/util/signature_utils.js
guymguym: this case might be related to the rpc auth flow.

src/sdk/account_sdk.js
shirady: I will, thanks.

src/manage_nsfs/nc_master_key_manager.js
shirady: I will note that the `encrypted_secret_key` will appear at the end of the printed access object when looking at the config (for humans).

src/util/signature_utils.js
```diff
@@ -362,13 +362,47 @@ function authenticate_request_by_service(req, sdk) {
...
+        throw new RpcError('SIGNATURE_DOES_NOT_MATCH', `Signature that was calculated did not match`);
+    }
+}
+
+/**
+ * @param {string} component
+ */
+function _throw_error_authorize_request_account_by_token(component) {
+    dbg.error(`authorize_request_account_by_token: we don't have ${component}`);
+    throw new RpcError('INTERNAL ERROR', `Needs ${component} for authorize request account`);
```
guymguym, line 405:
```suggestion
    throw new RpcError('INTERNAL_ERROR', `Needs ${component} for authorize request account`);
```

src/sdk/accountspace_fs.js
```diff
@@ -259,73 +231,233 @@ class AccountSpaceFS {
...
-            secret_key: dummy_access_key.secret_key,
-            create_date: new Date(),
-        };
+        const action = 'create_access_key';
+        dbg.log1(`AccountSpaceFS.${action}`, params, account_sdk);
+        try {
+            const requesting_account = account_sdk.requesting_account;
+            const requester = this._check_if_requesting_account_is_root_account_or_user_om_himself(action,
+                requesting_account, params.username);
+            const name_for_access_key = params.username ?? requester.name;
```
guymguym, line 252: this ?? operator will only fallback if undefined, but not if empty string. is that really what you meant??

guymguym: I think using the ?? operator is quite esoteric and are probably not intended for common cases.
It is mostly useful to avoid number 0 from being considered as falsy value (much less useful for empty strings or boolean false values...)

src/sdk/accountspace_fs.js
```diff
@@ -259,73 +231,233 @@ class AccountSpaceFS {
...
+            const requesting_account = account_sdk.requesting_account;
+            const access_key_id = params.access_key;
+            const requested_account_path = get_symlink_config_file_path(this.access_keys_dir, access_key_id);
+            await this._check_if_account_exists_by_access_key_symlink(action, requesting_account, requested_account_path, access_key_id);
+            const requested_account = await get_config_data(this.config_root_backend, requested_account_path, true);
+            this._check_if_requested_account_same_root_account_as_requesting_account(action,
+                requesting_account, requested_account);
+            return {
+                region: dummy_region, // GAP
+                last_used_date: new Date(Date.now() - 30 * MS_PER_MINUTE), // GAP
```
guymguym, line 305: why not just new Date() for simplicity? is there any need to make this date appear in the past??

src/sdk/accountspace_fs.js
```diff
@@ -497,43 +640,174 @@ class AccountSpaceFS {
...
-        res.dummy_access_key = dummy_access_key1;
-    }
-    return res;
+    _get_available_index_for_access_key(access_keys) {
+        // empty array or array with 1 access keys in index 1
+        if (access_keys.length === 0 || access_keys[0] === undefined) {
+            return 0;
+        }
+        return 1;
+    }
```
guymguym, line 719: this is really rigid. just push into the array.

src/sdk/accountspace_fs.js
```diff
@@ -259,73 +231,233 @@ class AccountSpaceFS {
...
+            const requesting_account = account_sdk.requesting_account;
+            const access_key_id = params.access_key;
+            const requester = this._check_if_requesting_account_is_root_account_or_user_om_himself(action,
+                requesting_account, params.username);
+            const requested_account_path = get_symlink_config_file_path(this.access_keys_dir, params.access_key);
+            await this._check_if_account_exists_by_access_key_symlink(action, requesting_account, requested_account_path, access_key_id);
+            const requested_account = await this._get_account_decrypted_data_optional(requested_account_path, true);
+            this._check_if_requested_account_same_root_account_as_requesting_account(action,
+                requesting_account, requested_account);
+            const { index_for_access_key, access_key } = this._get_access_key(requested_account, params.access_key);
```
guymguym, line 338: better return the actual access_key_obj rather than index

guymguym (changes requested):
posted some comments above. is the signature_utils change stable?

guymguym (approved):

src/util/signature_utils.js
```diff
@@ -362,13 +362,47 @@ function authenticate_request_by_service(req, sdk) {
...
 // mutual code of S3, STS and IAM services
-function authorize_request_account_by_token(token, requesting_account, is_secret_optional) {
-    const signature_secret = token.temp_secret_key || requesting_account?.access_keys?.[0]?.secret_key?.unwrap();
-    const should_check_signature = !is_secret_optional || signature_secret;
-    if (should_check_signature) {
-        const signature = get_signature_from_auth_token(token, signature_secret);
-        if (token.signature !== signature) throw new RpcError('SIGNATURE_DOES_NOT_MATCH', `Signature that was calculated did not match`);
+// 1 - check access key status (in case deactivated property true)
+// 2 - check secret_key signature (in case we should check the signature)
+function authorize_request_account_by_token(token, requesting_account) {
```
shirady, line 367: @nimrod-becker This is the function I'm changing that is mutual to all services (S3, STS, and IAM).
It doesn't add new functionality in the containerized environment.

shirady:
### Testing Instructions - Manual Testing in Containerized Environment:
#### Background
Since I changed the function `authorize_request_account_by_token` I also tested it in a containerized environment with the operator.
Tested on the default backing store (in the local cluster it is PVC) using the `first.bucket`.

#### Debugging addition:
We added printings in `signature_utils` inside `authorize_request_account_by_token`, and also warp this line inside the RPC auth server check signature ([here](https://github.com/noobaa/noobaa-core/blob/ed3cd1c5ba4009aee1aa391076a3761f03119e8d/src/server/common_services/auth_server.js#L485))

#### Requirements:
1. Build the image and deploy noobaa using Rancher desktop (see [guide](https://github.com/noobaa/noobaa-operator/blob/master/doc/dev_guide/deply_noobaa_on_minikube_or_rancher_desktop.md)).
Notes:
- `nb` is an alias that runs the local operator from build/_output/bin (alias created by `devenv` script).
- The namespace noobaa is installed on is `-n test1` in this example.

2. (optional) User port-forward for the services (because of an issue I have on my MacOS M1):
 - `kubectl port-forward -n test1 service/s3 12443:443` (12443 port in localhost for S3 service)
 - `kubectl port-forward -n test1 service/sts 12444:443` (12444 port in localhost for STS service)
3. Copy the noobaa admin access key and secret key from the output of `nb status --show-secrets -n test1` (we will use it later)

#### S3 Test
1. Create the alias for sending the S3 requests using the admin credentials (taken from previous step): `alias s3-user-1='AWS_ACCESS_KEY_ID=<admin-access-key> AWS_SECRET_ACCESS_KEY=<admin-secret-key> aws --no-verify-ssl --endpoint-url https://localhost:12443'`
2. Send a request, for example, to list the buckets: `s3-user-1 s3 ls` (expected result is to see the first.bucket in the output and see the debug added printings).

#### STS test
 1. Create a new noobaa user without the ability to create buckets (username is user1 in this example):
`nb account create user1 -n test1 --allow_bucket_create=false –show-secrets`.
In the output you will see the access key and secret key of the user:
```bash
INFO[0000] ✅ Exists: NooBaaAccount "user1"
INFO[0000] ✅ NooBaaAccount "user1" Phase is Ready

# NooBaaAccount spec:
allow_bucket_creation: false
default_resource: noobaa-default-backing-store
force_md5_etag: false

INFO[0000] ✅ Exists: Secret "noobaa-account-user1"
Connection info:
  AWS_ACCESS_KEY_ID      : <user1 access key id>
  AWS_SECRET_ACCESS_KEY  : <user1 secret key id>
```
2. Assign a new role policy to the user (username is user1 from the previous step and role name is my-role in this example):
`nb sts assign-role --email user1 --role_config '{"role_name": "my-role", "assume_role_policy": {"version": "2012-10-17", "statement": [ {"action": ["sts:AssumeRole"], "effect": "allow", "principal": ["*"]}
]}}' -n test1`.
3. Use the noobaa admin credentials to assume the role of the user (username is user1 , role name is my-role and seesion name is my-session in this example):
 - `alias sts-user-1='AWS_ACCESS_KEY_ID=<admin-credentials> AWS_SECRET_ACCESS_KEY=<admin-credentials> aws --no-verify-ssl --endpoint-url https://localhost:12444' # 12444 for sts service`
 - `sts-user-1 sts assume-role --role-arn arn:aws:sts::<copy the access key of the assumed role>:role/my-role --role-session-name my-seesion`, you will see an output like this (I removed the information):
 ```bash
{
    “Credentials”: {
        “AccessKeyId”: “<>”,
        “SecretAccessKey”: “<>”,
        “SessionToken”: “<>”,
        “Expiration”: “<>”
    },
    “AssumedRoleUser”: {
        “AssumedRoleId”: “<>:my-session”,
        “Arn”: “arn:aws:sts::<>:assumed-role/my-role/my-session”
    },
    “PackedPolicySize”: 0
}
```
4. With the credentials that the CLI returned in the previous step:
- `alias s3-user-assumed='AWS_ACCESS_KEY_ID=<previous step> AWS_SECRET_ACCESS_KEY=<previous step> AWS_SESSION_TOKEN="<previous step>" aws --no-verify-ssl --endpoint-url https://localhost:12443/' # 12443 for s3 service`
- try to list the buckets (should succeed, see the first.bucket in the output): `s3-user-assumed s3 ls`.
- try to create a new S3 bucket (should be blocked): `S3-user-assumed s3 mb:test-bucket`
output:
```bash
make_bucket failed: s3://test-bucket An error occurred (AccessDenied) when calling the CreateBucket operation: Access Denied
```
Endpoint logs:
- [RpcError: Not allowed to create new buckets] { rpc_code: 'UNAUTHORIZED' }

Note related to STS: 
Currently, there is an issue in STS (it uses RPC token instead of creating its token), this is the workaround for it (until we merge the fix - put this line in the comment: `if (config.NOOBAA_AUTH_TOKEN) return config.NOOBAA_AUTH_TOKEN;` in the file `jwt_utils`):
```js
function make_auth_token(object = {}, jwt_options = {}) {
    // Remote services/endpoints should not sign tokens
    // if (config.NOOBAA_AUTH_TOKEN) return config.NOOBAA_AUTH_TOKEN; //SDSD try
    // create and return the signed token
    return jwt.sign(object, get_jwt_secret(), jwt_options);
}
```

PR #8121, opened 2024-06-09
[Guaranteed Bucket Logging][NSFS]PR3 - Persistent logging

jackyalbo:
### Explain the changes
1. Adding the option for persistent bucket logging
2. Logging twice in case of persistent logging
3. Adding a command to export the logging for NSFS

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/s3/s3_bucket_logging.js
+++ b/src/endpoint/s3/s3_bucket_logging.js
@@ -5,16 +5,16 @@ const dbg = require('../../util/debug_module')(__filename);
 const http_utils = require('../../util/http_utils');
 const dgram = require('node:dgram');
 const { Buffer } = require('node:buffer');
-
+const config = require('../../../config');
 
 async function send_bucket_op_logs(req, res) {
-    if (req.params && req.params.bucket) {
+    if (req.params && req.params.bucket && req.op_name !== 'put_bucket') {
         const bucket_info = await req.object_sdk.read_bucket_sdk_config_info(req.params.bucket);
         dbg.log2("read_bucket_sdk_config_info =  ", bucket_info);
 
         if (is_bucket_logging_enabled(bucket_info)) {
             dbg.log2("Bucket logging is enabled for Bucket : ", req.params.bucket);
-            endpoint_bucket_op_logs(req.op_name, req, res, bucket_info);
+            await endpoint_bucket_op_logs(req.op_name, req, res, bucket_info);
         }
     }
 }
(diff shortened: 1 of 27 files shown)


src/endpoint/s3/s3_bucket_logging.js
```diff
@@ -54,15 +56,30 @@ const create_syslog_udp_socket = (() => {
...
 
-function endpoint_bucket_op_logs(op_name, req, res, source_bucket) {
+async function endpoint_bucket_op_logs(op_name, req, res, source_bucket) {
 
     // 1 - Get all the information to be logged in a log message.
     // 2 - Format it and send it to log bucket/syslog.
     const s3_log = get_bucket_log_record(op_name, source_bucket, req, res);
     dbg.log1("Bucket operation logs = ", s3_log);
 
+    switch (config.BUCKET_LOG_TYPE) {
```
tangledbytes, line 66: It seems that this function will be invoked for each request. Can we create a persistent logger in a higher scope so that we don't have to `open` and `close` the log file every time?

The log file seems to be getting processed using `.process` which should take care of replacing the old file as well as the logger should be able to detect the replacement and give up the locks.

src/manage_nsfs/manage_nsfs_logging.js
```diff
@@ -0,0 +1,115 @@
...
+                    Key: `${log_prefix}${format_aws_date(date)}-${sha.slice(0, 16).toUpperCase()}`,
+                    Body: upload_stream,
+                }).promise()));
+                bucket_streams[source_bucket] = upload_stream;
+            }
+            dbg.log1('uploading entry to target bucket', entry);
+            bucket_streams[source_bucket].write(entry + '\n');
+        });
+        Object.values(bucket_streams).forEach(st => st.end());
+        await Promise.all(promises);
```
tangledbytes, line 107: Do we need to support the scenarios where uploads are failing to just a selected few buckets but not others? Right now I think we would keep retrying it for the entire log file if the upload fails even just once.

src/endpoint/s3/s3_bucket_logging.js
jackyalbo: Will take a look on that

src/manage_nsfs/manage_nsfs_logging.js
jackyalbo: maybe it's a good idea to cover the promises with retries as well...thanks

src/endpoint/s3/s3_bucket_logging.js
jackyalbo: done! moved up to req.bucket_logger.

src/server/bg_services/bucket_logs_upload.js
```diff
@@ -7,6 +7,15 @@ const system_utils = require('../utils/system_utils');
...
 const fs = require('fs');
+const stream = require('stream');
+const crypto = require('crypto');
+const { PersistentLogger, LogFile } = require('../../util/persistent_logger');
+const { format_aws_date } = require('../../util/time_utils');
+const nsfs_schema_utils = require('../../manage_nsfs/nsfs_schema_utils');
+const Semaphore = require('../../util/semaphore');
+const P = require('../../util/promise');
+
+const sem = new Semaphore(10);
```
dannyzaken, line 18: better to define the semaphore size in a config

src/server/bg_services/bucket_logs_upload.js
```diff
@@ -178,4 +187,76 @@ class BucketLogUploader {
...
+                })));
+                bucket_streams[source_bucket] = upload_stream;
+            }
+            dbg.log2(`uploading entry: ${entry} to target bucket: ${target_bucket}`);
+            bucket_streams[source_bucket].write(entry + '\n');
+        });
+        Object.values(bucket_streams).forEach(st => st.end());
+        await Promise.all(promises);
+    } catch (error) {
+        dbg.error('unexpected error in upload to bucket:', error, 'for:', log_file);
```
dannyzaken, line 255: do we want to have a cleanup of incomplete logs in case of an error?

jackyalbo: As we seen together no cleanup is needed

src/manage_nsfs/manage_nsfs_logging.js
jackyalbo: Added a check on if the bucket doesn't exist and 3 retries

dannyzaken (approved):

Issue #8032, opened 2024-05-09
NC | Noobaa syslogs are sent to two different files


naveenpaul1:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
debug_module sends every log message to two targets but but native code logs only to stderr at the moment.
1. syslog facility local0 - [rsyslog omfile](https://www.rsyslog.com/doc/configuration/modules/omfile.html) copies it to `/var/log/noobaa.log`
2. stdout/stderr - [systemd copies it into its journal](https://www.freedesktop.org/software/systemd/man/latest/systemd-journald.service.html#) and then [rsyslog imjournal](https://www.rsyslog.com/doc/configuration/modules/imjournal.html) copies it to the file `/var/log/messages`.
3. send strerr to different log message.

### Expected behavior
1. Message should not be duplicated to `/var/log/noobaa.log` and `/var/log/messages`
2. journal have rate limit for the messages(20,000 message per 10mnts), Once we stop sending message to journal this bottleneck can be avoided.

### Steps to reproduce
1. Install RPM  and verify noobaa logs in `/var/log/messages` and `/var/log/noobaa.log`

naveenpaul1:
@guymguym @romayalon all the log messages are duplicated in both /logs/message and /logs/noobaa.log file. `debug_module.log_internal()` method writing logs to console using `stderr` or `console` after passing it to syslog. And these console logs are collected by  `imjournal` and added to `/logs/message` log file. its fine for local testing but RPM its not that needed.
One solution would be set a config property to control these logs writing to the console. Logs will print added to console if its enabled, Something like this.
```
log_internal(msg_info) {
        if (syslog) {
            // syslog path
            syslog(this._levels_to_syslog[msg_info.level], msg_info.message_syslog, config.DEBUG_FACILITY);
        } else if (this._log_file) {
            // rotating file steam path (non browser)
            this._log_file.write(msg_info.message_file + os.EOL);
        }
        if (!config.ENABLE_WRITE_LOGS_TO_CONSOLE) return;   <---- New property to avoid log printing to console.
        // This is also used in order to log to the console
        // browser workaround, don't use rotating file steam. Add timestamp and level
        const logfunc = LOG_FUNC_PER_LEVEL[msg_info.level] || 'log';
        if (this._log_console_silent) {
            // noop
        } else if (console_wrapper) {
[...]

PR #8122, opened 2024-06-10
NC | Respect log configuration from native code

naveenpaul1:
### Explain the changes
1. Native layer logs are moved to `var/log/message/stderr` if property `LOG_TO_STDERR_ENABLED` true.
2. Native layer logs are moved to `var/log/noobaa.log` if property `LOG_TO_SYSLOG_ENABLED` true.
3. Repeated log `semaphore_monitor: START` removed

1. https://github.com/noobaa/noobaa-core/issues/8032

### Testing Instructions:
1. Install rsyslog and Noobaa RPM
2. And verify noobaa specific log are pushed only to `/var/log/noobaa.log` if property `LOG_TO_SYSLOG_ENABLED` true.
3. Noobaa logs are not duplicated in `/var/log/noobaa.log` and `/var/log/message`

- [X] Doc added/updated
- [ ] Tests added


romayalon:
@naveenpaul1 we discussed also about changing the debug_module to not print to console when syslog is true, right?

naveenpaul1:
@romayalon We are silenting console log when syslog is enabled
` this._log_console_silent = process.env.SYSLOG_ENABLED; `

src/native/util/common.h
```diff
@@ -29,18 +30,37 @@ namespace noobaa
...
 #define DBG_VISIBLE(level) (level <= __module_debug_var__)
-#define DBG(level, x)                        \
-    do {                                     \
-        if (DBG_VISIBLE(level)) {            \
-            LOG("[L" << level << "] " << x); \
-        }                                    \
+#define DBG(level, x)                                                   \
+    do {                                                                \
+        if (DBG_VISIBLE(level)) {                                       \
+            const char* syslog_enabled = std::getenv("SYSLOG_ENABLED"); \
```
romayalon, line 57: could you move it to be a global?

naveenpaul1: You mean `SYSLOG_ENABLED`?

romayalon: yes, like we have
const char* gpfs_dl_path = std::getenv("GPFS_DL_PATH");

src/native/util/common.h
```diff
@@ -29,18 +30,37 @@ namespace noobaa
 #define DVAL(x) #x "=" << x << " "
 
 #define LOG(x) std::cerr << LOG_PREFIX() << x << std::endl
+#define napi_log(x)                                    \
```
romayalon, line 33: there 14 LOG() calls in fs_napi, should we convert them to DBG() calls as well?

romayalon, line 33: why do we need this function? can we move it to be inside SYS_LOG()?

src/native/util/common.h
```diff
@@ -34,17 +35,32 @@ namespace noobaa
...
-        if (DBG_VISIBLE(level)) {            \
-            LOG("[L" << level << "] " << x); \
-        }                                    \
+#define DBG(level, x)                                                   \
+    do {                                                                \
+        if (DBG_VISIBLE(level)) {                                       \
+            const char* syslog_enabled = std::getenv("LOG_TO_SYSLOG_ENABLED"); \
+            if (strcmp(syslog_enabled,"true") == 0) {                          \
+                SYS_LOG("[L" << level << "] " << x);                    \
+            } else {                                                    \
```
romayalon, line 60: we should also check here if we should print to stderr, no?

src/util/debug_module.js
```diff
@@ -41,7 +41,7 @@ util.inspect.defaultOptions.depth = 10;
 util.inspect.defaultOptions.colors = true;
 util.inspect.defaultOptions.breakLength = Infinity;
 
-
+process.env.LOG_TO_SYSLOG_ENABLED = config.LOG_TO_SYSLOG_ENABLED ? 'true': 'false';
```
romayalon, line 44: this should be inside config.js like we do for GPFS_DL_PATH

src/native/fs/fs_napi.cpp
```diff
@@ -2174,13 +2174,13 @@ fs_napi(Napi::Env env, Napi::Object exports)
...
-        LOG("FS::GPFS GPFS_DL_PATH=" << gpfs_dl_path);
+        DBG1("FS::GPFS GPFS_DL_PATH=" << gpfs_dl_path);
         struct stat _stat_res;
         gpfs_lib_file_exists = stat(gpfs_dl_path, &_stat_res); //SYSCALL_OR_RETURN
         if (gpfs_lib_file_exists == -1) {
-            LOG("FS::GPFS WARN couldn't find GPFS lib file GPFS_DL_PATH=" << gpfs_dl_path);
+            DBG1("FS::GPFS WARN couldn't find GPFS lib file GPFS_DL_PATH=" << gpfs_dl_path);
         } else {
-            LOG("FS::GPFS found GPFS lib file GPFS_DL_PATH=" << gpfs_dl_path);
+            DBG1("FS::GPFS found GPFS lib file GPFS_DL_PATH=" << gpfs_dl_path);
```
romayalon, line 2183: DBG1 is probably too high, DBG0 maybe better
But @guymguym said that we need to keep it LOG, maybe I miss-understood?

src/util/debug_module.js
naveenpaul1: @romayalon In that case user should add the `LOG_TO_SYSLOG_ENABLED` and `LOG_TO_STDERR_ENABLED` to config.js. Otherwise `process.env` wont have these values.

src/native/fs/fs_napi.cpp
naveenpaul1: I will revert back to `LOG`

romayalon (approved):

src/util/debug_module.js
```diff
@@ -41,7 +41,9 @@ util.inspect.defaultOptions.depth = 10;
 util.inspect.defaultOptions.colors = true;
 util.inspect.defaultOptions.breakLength = Infinity;
 
-
+//TODO : Need to finalize the solution for process.env, adding it to avoid system crash
+process.env.LOG_TO_SYSLOG_ENABLED = config.LOG_TO_SYSLOG_ENABLED ? 'true' : 'false';
+process.env.LOG_TO_STDERR_ENABLED = config.LOG_TO_STDERR_ENABLED ? 'true' : 'false';
```
guymguym, line 46: I think this should be in config.js so it will live reload -
https://github.com/noobaa/noobaa-core/blob/00513d49bc10c86f6cbd52e73d67fede40de8c7c/config.js#L1090-L1097

src/native/util/common.h
```diff
@@ -34,17 +35,34 @@ namespace noobaa
...
-        }                                    \
+#define DBG(level, x)                                                   \
+    do {                                                                \
+        if (DBG_VISIBLE(level)) {                                       \
+            const char* syslog_enabled = std::getenv("LOG_TO_SYSLOG_ENABLED"); \
+            const char* stderr_enabled = std::getenv("LOG_TO_STDERR_ENABLED"); \
+            if (syslog_enabled && strcmp(syslog_enabled, "true") == 0) {       \
+                SYS_LOG("[L" << level << "] " << x);                           \
+            }                                                                  \
+            if (stderr_enabled && strcmp(stderr_enabled, "true") == 0) {       \
```
guymguym, line 62: I suggest to move this logic inside the LOG macro, and add a separate STD_LOG macro, and then the usage of the LOG and DNG macros will stay the same, like this for example
```cpp
#define STD_LOG(x) std::cerr << LOG_PREFIX() << x << std::endl

#define SYS_LOG(x)                                      \
  ...

#define LOG(x) do { \
            if (syslog_enabled && strcmp(syslog_enabled, "true") == 0) {       \
                SYS_LOG("[L" << level << "] " << x);                           \
            }                                                                  \
            if (stderr_enabled && strcmp(stderr_enabled, "true") == 0) {       \
                STD_LOG("[L" << level << "] " << x);                        \
            } \
    } while (0)

```

guymguym (changes requested):

src/util/debug_module.js
guymguym: should we also change the default in config.js to output to STDERR only unless overridden by runtime config?
https://github.com/noobaa/noobaa-core/blob/00513d49bc10c86f6cbd52e73d67fede40de8c7c/config.js#L475-L476

to:
```js
config.LOG_TO_SYSLOG_ENABLED = false;

```

src/native/util/common.h
```diff
@@ -34,17 +35,34 @@ namespace noobaa
...
-#define DBG(level, x)                        \
-    do {                                     \
-        if (DBG_VISIBLE(level)) {            \
-            LOG("[L" << level << "] " << x); \
-        }                                    \
+#define DBG(level, x)                                                   \
+    do {                                                                \
+        if (DBG_VISIBLE(level)) {                                       \
+            const char* syslog_enabled = std::getenv("LOG_TO_SYSLOG_ENABLED"); \
+            const char* stderr_enabled = std::getenv("LOG_TO_STDERR_ENABLED"); \
```
guymguym, line 58: OK on second thought I think this is too excessive - getenv is doing linear search over all the variables and calling it twice per debug log print will be bad for performance. lets figure out how to push this values down from js on demand - for example with fs_napi set_debug_level() or maybe better just add a new method.

src/util/debug_module.js
```diff
@@ -586,10 +585,14 @@ if (console_wrapper) {
...
-    native_log_level = dbg_conf.level;
-}
-if (Number(native_log_level)) {
-    nb_native().fs.set_debug_level(Number(native_log_level));
+
+function set_log_config() {
+    const dbg_native_conf = debug_config.get_debug_config(process.env.NOOBAA_LOG_LEVEL);
+    if (typeof(dbg_native_conf.level) === "number") {
+        nb_native().fs.set_debug_level(Number(dbg_native_conf.level));
+    }
```
guymguym, line 593:
```suggestion
    nb_native().fs.set_debug_level(dbg_native_conf.level);
```

src/native/util/common.h
```diff
@@ -32,19 +33,61 @@ namespace noobaa
 
 #define DVAL(x) #x "=" << x << " "
 
-#define LOG(x) std::cerr << LOG_PREFIX() << x << std::endl
+extern bool LOG_TO_STDERR_ENABLED;
+extern bool LOG_TO_SYSLOG_ENABLED;
+extern std::stringstream* _thread_log_stream;
```
guymguym, line 38: lets remove this for now

guymguym (changes requested):

guymguym (approved):

PR #8123, opened 2024-06-10
Utilize the correct certificates for the STS HTTPS server

Neon-White:
### Explain the changes
Up until now, the STS HTTPS server reused the certificate that was generated for the S3 service. This has led to a problem with Openshift's internal HAProxy not letting requests through since it was set to `verifyhost` - which subsequently failed since the requests contained the cert for `s3.` instead of `sts.` This PR modifies the code to use the appropriate STS service cert for STS.

The PR also adds support for the user to provide a custom cert path via env vars,

This PR is the other half of https://github.com/noobaa/noobaa-operator/pull/1370

1. https://bugzilla.redhat.com/show_bug.cgi?id=2009627

### Testing Instructions:
1. Deploy NooBaa over Openshift
2. Try to use the STS route (not the service) for STS actions
3. Make sure they work

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/ssl_utils.js
+++ b/src/util/ssl_utils.js
@@ -44,6 +44,7 @@ const certs = {
     MGMT: new CertInfo('/etc/mgmt-secret'),
     S3: new CertInfo('/etc/s3-secret'),
     EXTERNAL_DB: new CertInfo('/etc/external-db-secret'),
+    STS: new CertInfo('/etc/sts-secret'),
 };
 
 function generate_ssl_certificate() {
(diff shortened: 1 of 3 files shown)


src/util/ssl_utils.js
```diff
@@ -44,6 +44,7 @@ const certs = {
     MGMT: new CertInfo('/etc/mgmt-secret'),
     S3: new CertInfo('/etc/s3-secret'),
     EXTERNAL_DB: new CertInfo('/etc/external-db-secret'),
+    STS: new CertInfo('/etc/sts-secret'),
```
dannyzaken, line 47: As long as we add code here, I suggest moving the certs paths to `config.js`, so it will be easier to override if necessary. 
```suggestion
    STS: new CertInfo(config.STS_SECRET_PATH),
```

dannyzaken (approved):

PR #8126, opened 2024-06-10
Bucket_logging: Change log object name format.

aspandey:
At present we are creating log objects and giving them name based on log_prefix and log_bucket.

There are some constraint to that and one is that a user can set same log bucket and log prefix for two different source buckets. That will cause logs from two different source buckets to go to same log objects.

In addition to that, source bucket name in log file name will help to get it easily without processing log file.

This source bucket name can be used to get more logging configuration, which can be helpful in further processing.

---
Note: As we are not providing log_prefix in log object name, it should be retrieved and processed from system_store by bg service at the time of uploading objects on log bucket.


aspandey:
will resend it again

PR #8128, opened 2024-06-10
Bucket_logging: Change log object name format.

aspandey:
At present we are creating log objects and giving them name based on log_prefix and log_bucket.

There are some constraint to that and one is that a user can set same log bucket and log prefix for two different source buckets. That will cause logs from two different source buckets to go to same log objects.

In addition to that, source bucket name in log file name will help to get it easily without processing log file.

This source bucket name can be used to get more logging configuration, which can be helpful in further processing.

Note: As we are not providing log_prefix in log object name, it should be retrieved and processed from system_store by bg service at the time of uploading objects on log bucket.

### DIFF
--- a/src/deploy/NVA_build/noobaa_log_segregate.sh
+++ b/src/deploy/NVA_build/noobaa_log_segregate.sh
@@ -14,6 +14,8 @@ NOOBAA_BUCKET_LOGS_FINAL="/log/noobaa_bucket_logs/"
 BUCKET_LOG_FILE_SIZE_LIMIT=51200
 # Delimiter for the file name. To separate the file name from the timestamp.
 # S3 Bucket name are not allowed to have "_" in its name.
+# https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html
+
 File_Name_Del="_"
 
 # Lock file to make sure that the script is not running multiple times.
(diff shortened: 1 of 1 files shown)


jackyalbo (approved):
LGTM

src/deploy/NVA_build/noobaa_log_segregate.sh
```diff
@@ -82,39 +84,43 @@ while [ -n "$(ls -A "$log_dir")" ]; do
...
             fi
 
-            if [[ "$log_field" =~ "log_prefix" ]]; then
-                IFS=: read -r -a prefix_values <<< "$log_field"
-                log_prefix_value=$(echo "${prefix_values[1]}" | tr -d '"' | sed 's/^[ \t]*//; s/[ \t]*$//')
+            if [[ "$log_field" =~ "source_bucket" ]]; then
+                IFS=: read -r -a source_values <<< "$log_field"
+                source_bucket_name=$(echo "${source_values[1]}" | tr -d '"' | sed 's/^[ \t]*//; s/[ \t]*$//')
+                continue
             fi
```
liranmauda, line 111: Why do we need to duplicate it?
can't it be a function?

aspandey: It could be a function. I modified the code and can see that it is working. However, I find the current code more readable and less prone to errors. Moreover, we are not saving any lines of code; all it is doing is adding more complexity to understand the code.

liranmauda (approved):

PR #7986, opened 2024-04-18
S3 | EncodeURIComponent / DecodeURIComponent xattr

romayalon:
### Explain the changes
1. Revert the `__dot__` convertion introduced on - https://github.com/noobaa/noobaa-core/pull/7408.
2. PUT Object / Create MPU - Call decodeURIComponent(xattr_val) before processing the request.
3. Head Objcet - If not encoded - Call encodeURIComponent(xattr_val). We shouldn't encode twice an already encoded xattr values (for example - In NSFS flow, in a case where the xattr is set directly on the file) 

### Issues: Fixed #xxx / Gap #xxx
1. Fixed #https://github.com/noobaa/noobaa-core/issues/7404
2. Fixed #https://github.com/noobaa/noobaa-core/issues/7896

### Testing Instructions:
1. `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha noobaa-core/src/test/unit_tests/test_bucketspace.js`

- [ ] Doc added/updated
- [x] Tests added

PR #8129, opened 2024-06-10
NC | Revert url encode/decode of xattr values

romayalon:
### Explain the changes
1. Revert code added in https://github.com/noobaa/noobaa-core/pull/7986
2. Added try & catch that skips xattr headers that can not be set during setHeader() by assuming that xattrs that can be returned are coming through s3. If invalid xattrs were set directly on the file system and can not return through http header, the best behavior is not to fail the head object but to not return the invalid xattr. If customer wants to get invalid headers via s3 the xattr should be encoded on the file system if they are not us-ascii.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -64,7 +64,7 @@ function get_request_xattr(req) {
         if (!hdr.startsWith(X_AMZ_META)) return;
         const key = hdr.slice(X_AMZ_META.length);
         if (!key) return;
-        xattr[key] = decodeURIComponent(val);
+        xattr[key] = val;
     });
     return xattr;
 }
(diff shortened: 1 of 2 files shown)


src/endpoint/s3/s3_utils.js
```diff
@@ -97,7 +95,11 @@ function set_response_xattr(res, xattr) {
         }
         returned_keys += 1;
         size_for_md_left -= md_header_size;
-        res.setHeader(X_AMZ_META + key, val);
+        try {
+            res.setHeader(X_AMZ_META + key, val);
+        } catch (err) {
+            dbg.warn('s3_utils.set_response_xattr set_header failed, skipping... header value', val);
```
guymguym, line 101: need more info in this log print - at least the key, and the request url...

guymguym (changes requested):

guymguym (approved):

PR #8144, opened 2024-06-15
Fix log-based replication Prometheus field

Neon-White:
### Explain the changes
`size_of_objects` is used to update the Prometheus status of log-based replication, specifically for the values of `last_cycle_writes_size` and `last_cycle_error_writes_size`. However, the attribute (`Size`) that was allocated to `size_of_objects` does not exist, which has led to errors in our logs, and possibly malfunctioning Prometheus reports.
This PR replaces `Size` with `ContentLength`, which also resolves the exceptions that were observed in the logs.

### Testing Instructions:
1. Set up an environment with log-based replication optimization
2. Disable the regular replication via `config.js`
3. Verify log-based replication works
4. Make sure no related errors were printed in the core logs

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/bg_services/replication_server.js
+++ b/src/server/bg_services/replication_server.js
@@ -85,7 +85,10 @@ async function copy_objects_mixed_types(req) {
             try {
                 await noobaa_con.copyObject(params).promise();
                 copy_res.num_of_objects += 1;
-                copy_res.size_of_objects += keys_diff_map[key][0].ContentLength;
+                // The size of the object can be in either Size or ContentLength, depending on whether
+                // the request was ListObjectVersions or HeadObject
+                const size_of_objects = keys_diff_map[key][0].ContentLength || keys_diff_map[key][0].Size;
+                copy_res.size_of_objects += size_of_objects;
             } catch (err) {
                 dbg.error('replication_server copy_objects_mixed_types: got error:', err);
             }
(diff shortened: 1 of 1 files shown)


src/server/bg_services/replication_server.js
```diff
@@ -100,7 +100,7 @@ async function copy_objects_mixed_types(req) {
                 try {
                     await noobaa_con.copyObject(params).promise();
                     copy_res.num_of_objects += 1;
-                    copy_res.size_of_objects += keys_diff_map[key][i].Size;
+                    copy_res.size_of_objects += keys_diff_map[key][i].ContentLength;
```
liranmauda, line 103: This will not work, The only valid option here is when we have versions (key.length > 1) 
In "version" we do not have `ContentLength`, we have `Size`. 
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html

src/server/bg_services/replication_server.js
```diff
@@ -85,7 +85,7 @@ async function copy_objects_mixed_types(req) {
             try {
                 await noobaa_con.copyObject(params).promise();
                 copy_res.num_of_objects += 1;
-                copy_res.size_of_objects += keys_diff_map[key][0].Size;
+                copy_res.size_of_objects += keys_diff_map[key][0].ContentLength;
```
liranmauda, line 88: In case of `(keys_diff_map[key].length === 1)` we should probably check if it is version or not.
If it is version, we should use `Size`
if it is not then we can use `ContentLength`
when we do not have version we are doing `head_object`: 
https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html

We can check if the key object here has`VersionId`

liranmauda (changes requested):
This will not work when we have version enabled.

liranmauda (approved):

PR #8149, opened 2024-06-17
Add context to fs_napi errors

### LINKED ISSUE #8065: NSFS on syscall error should log the operation info and not just the error code (even if debug level is low)
### Environment info
- NooBaa Version: **5.15.3**
- Platform: **rhel**

### Actual behavior
1. on syscall error we only see the err code unless we turn on higher debug log -
```
May 21 16:19:48 host node[2941584]: May-21 16:19:48.654 [nsfs/2941584] [ERROR] core.sdk.namespace_fs::  [Error: Operation not permitted] { code: 'EPERM' }
```

### Expected behavior
1. Should log the operation info with the error code.

### Steps to reproduce
1. Change fs permissions to cause EPERM.

tangledbytes:
### Explain the changes
This PR attempts to add `context` to the errors coming from `fs_napi`. The `context` field is then used in the S3 error reporting if it's available. The PR also reduces the log level for `OnError` method. 

This sometimes means duplicate logging, however it was still kept this way as not all the errors coming from the `fs_napi` necessarily fails an S3 operation. It could be helpful to see all the errors that we encountered and the error that resulted in the operation failure.

Another approach that was tested but I decided against it was to add the context info in the error message itself. It removes the construct of `context` but that also means that the error message will be more verbose and hard/fragile `===` tests against them.

Example of current fix:
image

Fixes #8065 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -655,6 +655,9 @@ struct FSWorker : public Napi::AsyncWorker
         if (_errno) {
             obj.Set("code", Napi::String::New(env, uv_err_name(uv_translate_sys_error(_errno))));
         }
+
+        obj.Set("context", Napi::String::New(env, _desc));
+
         _deferred.Reject(obj);
     }
 };
@@ -1311,7 +1314,7 @@ struct Fsync : public FSWorker
 };
 
 /**
- * GetPwName is an os op 
+ * GetPwName is an os op
  */
 struct GetPwName : public FSWorker
 {
@@ -2206,7 +2209,7 @@ fs_napi(Napi::Env env, Napi::Object exports)
             if (sizeof(struct gpfsRequest_t) != 256) {
                 PANIC("The gpfs get extended attributes is of wrong size" << sizeof(struct gpfsRequest_t));
             }
-            
+
             auto gpfs = Napi::Object::New(env);
             gpfs["register_gpfs_noobaa"] = Napi::Function::New(env, register_gpfs_noobaa);
             // we export the gpfs object, which can be checked to indicate that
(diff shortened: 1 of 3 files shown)


src/native/util/common.h
```diff
@@ -23,7 +23,11 @@ namespace noobaa
 {
 
 #ifndef __func__
-#define __func__ __FUNCTION__
+    #ifndef __PRETTY_FUNCTION__
+    #define __func__ __PRETTY_FUNCTION__
```
guymguym, line 27: should be ifdef

src/native/fs/fs_napi.cpp
```diff
@@ -650,11 +650,15 @@ struct FSWorker : public Napi::AsyncWorker
     {
         Napi::Env env = Env();
         DBG1("FS::FSWorker::OnError: " << _desc << " " << DVAL(error.Message()));
+        DBG0("FS::FSWorker::OnError: " << _desc << " " << DVAL(error.Message()));
```
guymguym, line 653: This might add a lot of log prints, is it needed to show the error desc? isn't the context field enough?

guymguym (changes requested):

src/native/fs/fs_napi.cpp
tangledbytes: The reason I added it was to be able to print all of the errors that we encounter on the napi layer. Not all the failures directly lead to a failure of a S3 request which means these can get lost. Having that said, it's a little bit between "debug" logging and "info" logging.

WDYT? Should I change it back to `DBG1`?

guymguym: dilemma indeed, but I think we cannot afford to always log this by default because this is something that happens often in normal flows. so unless you see that this is not the case, I would revert it to DBG1.

tangledbytes: Reverted log level.

guymguym (approved):

PR #8150, opened 2024-06-17
Fix IAM and STS endpoint servers

Neon-White:
### Explain the changes
#8123 has introduced two new bugs to the core
1. A missing `true` parameter has caused the STS server to malfunction, this PR fixes it
2. A missing `await` keyword has caused the IAM server to malfunction, this PR fixes that as well

### Testing Instructions:
For STS:
1. `rsh` into an endpoint pod
3. Run `curl -k https://localhost:7443` (change the port in case you're not using the default value)
4. Make sure you receive an error - `<?xml version="1.0" encoding="UTF-8"?><Error><Code>InternalFailure</Code><Message>The request processing has failed because of an unknown error, exception or failure.</Message><Resource>/</Resource><RequestId>...</RequestId></Error>`

For IAM:
1. Run `sudo node src/cmd/nsfs --debug 5 --https_port_iam 7005`
2. Make sure it doesn't crash with any errors


shirady (approved):

shirady:
@Neon-White please squash before merging

PR #8157, opened 2024-06-19
Fix Throw Error in `load_requesting_account` (object_sdk, account_sdk)

shirady:
### Explain the changes
1. Fix Throw Error in `load_requesting_account` (object_sdk, account_sdk) - if the `error.rpc_code` exists and it neither 
'NO_SUCH_ACCOUNT' nor 'NO_SUCH_USER' (for example error.rpc_code INTERNAL) the error still should be thrown. 
Note: `account_sdk` was copied from `object_sdk` (we don't have this issue in `sts_sdk`).
2. Simplify the condition and remove the `error.rpc_code && ` from the condition since it is strings comparison (in `account_sdk`, `object_sdk`, and `sts_sdk`).
 
### Issues:
1. Currently, errors when  `load_requesting_account` with error.rpc_code different than 'NO_SUCH_ACCOUNT' and 'NO_SUCH_USER' will be cached and not thrown.
2. Known GAP - we currently have the `load_requesting_account` implementation in all sdk service files and would like to reuse mutual code.

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/account_sdk.js
+++ b/src/sdk/account_sdk.js
@@ -64,12 +64,9 @@ class AccountSDK {
             }
         } catch (error) {
             dbg.error('load_requesting_account error:', error);
-            if (error.rpc_code) {
-                if (error.rpc_code === 'NO_SUCH_ACCOUNT') throw new RpcError('INVALID_ACCESS_KEY_ID', `Account with access_key not found`);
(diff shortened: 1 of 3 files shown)


shirady:
@liranmauda @romayalon 
It is in 2 commits in case you want to backport the object_sdk change.

romayalon (approved):

guymguym (approved):

PR #8169, opened 2024-06-26
Fix STS SessionToken generation

Neon-White:
### Explain the changes
1. This PR fixes STS functionality by separating `jwt_utils.make_auth_token` into two functions - `make_auth_token` which always signs the request, and  `make_internal_auth_token` that's used for RPC calls and only falls back to `make_auth_token` in case `NOOBAA_AUTH_TOKEN` isn't provided. Now, STS requests use `make_auth_token` directly, while internal callers were changed to use `make_internal_auth_token`

1. https://bugzilla.redhat.com/show_bug.cgi?id=2009627

### Testing Instructions:
1. See [the following comment](https://github.com/noobaa/noobaa-core/pull/8119#issuecomment-2180089580)

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/jwt_utils.js
+++ b/src/util/jwt_utils.js
@@ -16,10 +16,15 @@ function get_jwt_secret() {
 }
 
 function make_auth_token(object = {}, jwt_options = {}) {
-    // Remote services/endpoints should not sign tokens - unless required for STS
-    if (config.NOOBAA_AUTH_TOKEN && object.assumed_role_access_key === undefined) return config.NOOBAA_AUTH_TOKEN;
     // create and return the signed token
     return jwt.sign(object, get_jwt_secret(), jwt_options);
+}
+
+function make_internal_auth_token(object = {}, jwt_options = {}) {
+    // Remote services/endpoints should not sign tokens
+    if (config.NOOBAA_AUTH_TOKEN) return config.NOOBAA_AUTH_TOKEN;
+    // If an auth token isn't provided, fall back to signing normally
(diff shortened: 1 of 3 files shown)


guymguym (changes requested):
I prefer to separate it to two different functions, one for simply make_auth_token that is simply signing the input, and a separate one to be used for rpc tokens which can be called make_internal_auth_token than can check the config for remote endpoints and if not then fallback to call the other function. This means replacing the callers to use the right variant, which will make it more explicit and safer

src/util/jwt_utils.js
```diff
@@ -16,8 +16,8 @@ function get_jwt_secret() {
 }
 
 function make_auth_token(object = {}, jwt_options = {}) {
-    // Remote services/endpoints should not sign tokens
-    if (config.NOOBAA_AUTH_TOKEN) return config.NOOBAA_AUTH_TOKEN;
+    // Remote services/endpoints should not sign tokens - unless required for STS
+    if (config.NOOBAA_AUTH_TOKEN && object.assumed_role_access_key === undefined) return config.NOOBAA_AUTH_TOKEN;
```
shirady, line 20: @guymguym you had an idea for a fix that includes separating the functions, would you please give more details about it?
Anyway, I think this comment about "Remote services/endpoints" is not updated.

shirady:
@Neon-White I deleted my comment because after sending it I saw @guymguym's  comment.

guymguym (approved):

PR #8172, opened 2024-06-30
NSFS | NC | IAM Service - Input Validation

shirady:
### Explain the changes
1. Add input validation before every IAM operation.
2. Rename the file of op `get_access_key_last_used` (missing prefix `iam_`).

### Issues: 
List of GAPs:
1. Move constants to the file `iam_constants.js` (mainly from `iam_utils.js`).
2. Special handle for anonymous - currently we block all IAM User API and Access keys API on it (in the future we want to allow IAM User API on it).

#### Unit Tests
Please run:
1. `npx jest test_iam_utils.test.js`
3. `npx jest test_iam_ops_input_validation.test.js`

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/endpoint/iam/iam_constants.js
+++ b/src/endpoint/iam/iam_constants.js
@@ -1,0 +1,37 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+// TODO - move more constants here (mainly from iam_utils)
+const DEFAULT_MAX_ITEMS = 100;
+
+// TODO - reuse
+// function name in snake case style
+const IAM_ACTIONS = Object.freeze({
+    CREATE_USER: 'create_user',
+    GET_USER: 'get_user',
+    DELETE_USER: 'delete_user',
+    UPDATE_USER: 'update_user',
+    LIST_USERS: 'list_users',
+    CREATE_ACCESS_KEY: 'create_access_key',
+    GET_ACCESS_KEY_LAST_USED: 'get_access_key_last_used',
+    UPDATE_ACCESS_KEY: 'update_access_key',
+    DELETE_ACCESS_KEY: 'delete_access_key',
+    LIST_ACCESS_KEYS: 'list_access_keys'
+});
+
+
+// parameter names
+const IAM_PATH = 'Path';
+const NEW_IAM_PATH = 'NewPath';
+const IAM_PATH_PREFIX = 'PathPrefix';
+const USERNAME = 'UserName';
+const NEW_USERNAME = 'NewUserName';
+
+
+// EXPORTS
+exports.IAM_ACTIONS = IAM_ACTIONS;
+exports.DEFAULT_MAX_ITEMS = DEFAULT_MAX_ITEMS;
+exports.IAM_PATH = IAM_PATH;
+exports.NEW_IAM_PATH = NEW_IAM_PATH;
+exports.IAM_PATH_PREFIX = IAM_PATH_PREFIX;
+exports.USERNAME = USERNAME;
+exports.NEW_USERNAME = NEW_USERNAME;
--- a/src/endpoint/iam/ops/iam_list_users.js
+++ b/src/endpoint/iam/ops/iam_list_users.js
@@ -3,6 +3,7 @@
 
 const dbg = require('../../../util/debug_module')(__filename);
 const iam_utils = require('../iam_utils');
+const iam_constants = require('../iam_constants');
 const { CONTENT_TYPE_APP_FORM_URLENCODED } = require('../../../util/http_utils');
 
 /**
@@ -12,10 +13,12 @@ async function list_users(req, res) {
 
     const params = {
         marker: req.body.marker,
-        max_items: req.body.max_items,
+        // ISSUE - AWS CLI doesn't pass max_items in req.body
+        max_items: req.body.max_items ?? iam_constants.DEFAULT_MAX_ITEMS,
         iam_path_prefix: req.body.path_prefix,
     };
     dbg.log1('IAM LIST USERS', params);
+    iam_utils.validate_params(iam_constants.IAM_ACTIONS.LIST_USERS, params);
     const reply = await req.account_sdk.list_users(params);
     dbg.log2('list_users reply', reply);
 
--- a/src/util/string_utils.js
+++ b/src/util/string_utils.js
@@ -10,6 +10,11 @@ const email_regexp = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\
 const escape_regexp = /[\\^$.*+?()[\]{}|]/g;
 const access_key_regexp = /^[a-zA-Z0-9]{20}$/;
 const secret_key_regexp = /^[a-zA-Z0-9+/]{40}$/;
+// regex for IAM service
+const AWS_IAM_PATH_REGEXP = /(\u002F)|(\u002F[\u0021-\u007E]+\u002F)/;
+const AWS_USERNAME_REGEXP = /[\w+=,.@-]+/;
+const AWS_IAM_LIST_MARKER = /[\u0020-\u00FF]+/;
+const AWS_IAM_ACCESS_KEY_INPUT_REGEXP = /[\w]+/;
 
 function crypto_random_string(len, charset = ALPHA_NUMERIC_CHARSET) {
     // In order to not favor any specific chars over others we limit the maximum random value
@@ -156,3 +161,7 @@ exports.is_email_address = is_email_address;
 exports.escape_reg_exp = escape_reg_exp;
 exports.access_key_regexp = access_key_regexp;
 exports.secret_key_regexp = secret_key_regexp;
+exports.AWS_IAM_PATH_REGEXP = AWS_IAM_PATH_REGEXP;
+exports.AWS_USERNAME_REGEXP = AWS_USERNAME_REGEXP;
+exports.AWS_IAM_LIST_MARKER = AWS_IAM_LIST_MARKER;
+exports.AWS_IAM_ACCESS_KEY_INPUT_REGEXP = AWS_IAM_ACCESS_KEY_INPUT_REGEXP;
--- a/docs/design/iam.md
+++ b/docs/design/iam.md
@@ -103,3 +103,18 @@ root accounts manager can run IAM access keys create/update/delete/list - only o
 
 Here attached a diagram with all the accounts that we have in our system:
 ![All accounts diagram](https://github.com/noobaa/noobaa-core/assets/57721533/c4395c06-3ab3-4425-838b-c020ef7cc38a)
+
+## Supported Actions and their request parameters
+### Supported IAM User Operations
+- IAM CreateUser: Path, UserName (not supported: PermissionsBoundary, Tags.member.N)
+- IAM GetUser: UserName
+- IAM UpdateUser: NewPath, NewUserName, UserName
+- IAM DeleteUser: UserName
+- IAM ListUsers: PathPrefix (not supported: Marker, MaxItems)
+
+### Supported IAM Access Keys Operations
+- IAM CreateAccessKey: UserName
+- IAM GetAccessKeyLastUsed: AccessKeyId
+- IAM UpdateAccessKey: AccessKeyId, Status, UserName
+- IAM DeleteAccessKey: AccessKeyId, UserName
+- IAM ListAccessKeys: UserName (not supported: Marker, MaxItems)
--- a/src/endpoint/iam/iam_rest.js
+++ b/src/endpoint/iam/iam_rest.js
@@ -45,7 +45,7 @@ const IAM_OPS = js_utils.deep_freeze({
     post_list_users: require('./ops/iam_list_users'),
     // access key CRUD
     post_create_access_key: require('./ops/iam_create_access_key'),
-    post_get_access_key_last_used: require('./ops/get_access_key_last_used'),
+    post_get_access_key_last_used: require('./ops/iam_get_access_key_last_used'),
     post_update_access_key: require('./ops/iam_update_access_key'),
     post_delete_access_key: require('./ops/iam_delete_access_key'),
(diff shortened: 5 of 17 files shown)


src/endpoint/iam/iam_constants.js
```diff
@@ -0,0 +1,37 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+// TODO - move more constants here (mainly from iam_utils)
+const DEFAULT_MAX_ITEMS = 100;
+
+// TODO - reuse
+// function name in snake case style
+const IAM_ACTIONS = {
```
guymguym, line 8: Object.freeze

src/endpoint/iam/ops/iam_list_users.js
```diff
@@ -12,10 +13,11 @@ async function list_users(req, res) {
 
     const params = {
         marker: req.body.marker,
-        max_items: req.body.max_items,
+        max_items: req.body.max_items ?? iam_constants.DEFAULT_MAX_ITEMS,
```
guymguym, line 16: i am not sure that nullish condition is better than simple logical or. what should be done if body max items is 0?

src/util/string_utils.js
```diff
@@ -10,6 +10,11 @@ const email_regexp = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\
 const escape_regexp = /[\\^$.*+?()[\]{}|]/g;
 const access_key_regexp = /^[a-zA-Z0-9]{20}$/;
 const secret_key_regexp = /^[a-zA-Z0-9+/]{40}$/;
+// regex for IAM service
+const AWS_IAM_PATH_REGEXP = /(\u002F)|(\u002F[\u0021-\u007E]+\u002F)/;
```
guymguym, line 14: using unicode code numbers for ascii characters makes it harder to read. to use forward-slash in regex you can either use \/ or my preferred [/]

guymguym: and what is this range \u0021-\u007e? if this is needed then better document it and explain what this means.

src/util/string_utils.js
```diff
@@ -10,6 +10,11 @@ const email_regexp = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\
...
+const AWS_USERNAME_REGEXP = /[\w+=,.@-]+/;
+const AWS_LIST_MARKER = /[\u0020-\u00FF]+/;
+const AWS_ACCESS_KEY_ID = /[\w]+/;
```
guymguym, line 17: is \w really matching the correct character set for access keys?

src/util/string_utils.js
```diff
@@ -10,6 +10,11 @@ const email_regexp = /^((([a-z]|\d|[!#\$%&'\*\+\-\/=\?\^_`{\|}~]|[\u00A0-\uD7FF\
 const escape_regexp = /[\\^$.*+?()[\]{}|]/g;
 const access_key_regexp = /^[a-zA-Z0-9]{20}$/;
 const secret_key_regexp = /^[a-zA-Z0-9+/]{40}$/;
+// regex for IAM service
+const AWS_IAM_PATH_REGEXP = /(\u002F)|(\u002F[\u0021-\u007E]+\u002F)/;
+const AWS_USERNAME_REGEXP = /[\w+=,.@-]+/;
+const AWS_LIST_MARKER = /[\u0020-\u00FF]+/;
```
guymguym, line 16: unclear. why is this the range?

guymguym (changes requested):

src/endpoint/iam/iam_constants.js
shirady: Thanks, added.

src/endpoint/iam/ops/iam_list_users.js
shirady: I will throw an error because in the docs of [ListUsers](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListUsers.html#API_ListUsers_RequestParameters): 

> Valid Range: Minimum value of 1. Maximum value of 1000.

I tested it with 0 and saw an empty array (also tested it with a negative value -1 and got the items).
In the code, I decide to act as it is supposed to be according to the docs, and I will throw an error on values out of the range (including 0).
Using the nullish coalescing operator operator gives exactly what I want - according to the docs:

> If you do not include this parameter, the number of items defaults to 100.

Since you raised it now I noticed that for this parameter they used "Valid Range" While in places I tested and saw errors they used "Length Constraints". I still think it is better to throw an error and enforce the values to use as in the expected range.

Reminder - currently we haven't implemented pagination yet, so the parameter is not supported yet

src/util/string_utils.js
shirady: I took the pattern for the docs, for example ([CreateUser](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateUser.html#API_CreateUser_RequestParameters)):

> Pattern: (\u002F)|(\u002F[\u0021-\u007E]+\u002F)

Explanation of the pattern:

> a string of characters consisting of either a forward slash (/) by itself or a string that must begin and end with forward slashes. In addition, it can contain any ASCII character from the ! (\u0021) through the DEL character (\u007F), including most punctuation characters, digits, and upper and lowercased letters.

I can add the link as an inline comment, will it be easier?

I didn't understand this comment:
> using unicode code numbers for ascii characters makes it harder to read. to use forward-slash in regex you can either use / or my preferred [/]

Would you please add a suggestion?
For the forward and backward slash I'm testing it in the code itself (the pattern doesn't include it):
`const valid_aws_path = input_path.startsWith('/') && input_path.endsWith('/') && AWS_IAM_PATH_REGEXP.test(input_path);
`

src/util/string_utils.js
shirady: I took the pattern for the docs, for example ([ListUsers](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListUsers.html#API_ListUsers_RequestParameters)):

> Pattern: [\u0020-\u00FF]+

It was hard to find something invalid in this pattern - the example I found it:

> const invalid_characters = '\uD83D\uDE0A'; //emoji is not part of the pattern

src/util/string_utils.js
shirady: I took the pattern from the docs, for example ([UpdateAccessKey](https://docs.aws.amazon.com/IAM/latest/APIReference/API_UpdateAccessKey.html#API_UpdateAccessKey_RequestParameters)):

> Pattern: [\w]+

Explanation of the pattern:
> a string of characters that can consist of any upper or lowercased letter or digit.

src/util/string_utils.js
guymguym: Perhaps this is very specific for list users, so better rename this to AWS_LIST_USERS_MARKER

src/util/string_utils.js
guymguym: in javascript \w includes also underscore

guymguym: and we already have an access key regex defined - 
https://github.com/noobaa/noobaa-core/blob/7b0d3e3ccbcbf45365877ef619e52fa0bac5626e/src/util/string_utils.js#L11-L12

src/util/string_utils.js
shirady: It is for list users and list access keys.
So I will suggest renaming it to `AWS_IAM_LIST_MARKER`, WDYT?

src/util/string_utils.js
shirady: 1. Why underscore should be included?
I did a test in AWS IAM - I sent access key with _ and received an error `InvalidClientTokenId`:

> The security token included in the request is invalid

and then I sent with # and received an error `ValidationError`:

> The specified value for accessKeyId is invalid. It must contain only alphanumeric characters

Note: I used `aws iam get-access-key-last-used --access-key-id <>`. took a valid access key id and changed one of its characters.

2. It is not the same regex - for example, it sets the length to be exactly 20 while for an input it can be:

> Length Constraints: Minimum length of 16. Maximum length of 128.

src/util/string_utils.js
guymguym: sounds great

src/endpoint/iam/ops/iam_list_users.js
guymguym: ok, you definitely understand the difference of || and ?? well but this is always subtle for many people to read unless carefully focusing on it, imho. in any case we can use it, but I am trying to minimize it.

src/util/string_utils.js
guymguym: ok nevermind, better leave it as as.

src/util/string_utils.js
guymguym: so how about we keep the pattern as is but to avoid confusion with other definitions we rename this const to something like AWS_IAM_ACCESS_KEY_INPUT_REGEXP ?

src/endpoint/iam/ops/iam_list_users.js
shirady: I can add a comment in the code: "If one did not include max_items parameter (req.body.max_items value is undefined), we use a default number DEFAULT_MAX_ITEMS", but I don't like it (it explains the operator ?? and does not give an added value), let me know what you think.

src/util/string_utils.js
shirady: I will change it.
I still think this name will be confusing because later we might use this validation in the CLI.

src/util/string_utils.js
shirady: I will change it also, thanks.

src/util/string_utils.js
guymguym: but you have to agree that having two different regexp that are names aws_access_key is way more confusing

src/endpoint/iam/ops/iam_list_users.js
guymguym: oh but what is the type of req.body.max_items? string? if so then we actually want to parse it to a number and not send once a string and once a number (when undefined)

src/util/string_utils.js
shirady: I agree with this, I comment on it in case we have a better name.

src/endpoint/iam/ops/iam_list_users.js
shirady: It should be a number, but now I noticed that there is an issue with AWS CLI in which it doesn't pass it.

I ran: 
`s3-nc-user-1-iam iam list-users --path-prefix /abc/ --max-items 7 --debug` 
I added the `--path-prefix` flag so we can see a flag that passes (and a flag that I think doesn't pass).
(s3-nc-user-1-iam is an alias 
for `AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:7005`).

In the debug I see AWS CLI list the argument of max-items:
```bash
2024-07-07 08:43:35,751 - MainThread - awscli.clidriver - DEBUG - Arguments entered to CLI: ['--no-verify-ssl', '--endpoint-url', 'https://localhost:7005', 'iam', 'list-users', '--path-prefix', '/abc/', '--max-items', '7', '--debug']
```

In the debug I don't see the max params passing to the request's body (PathPrefix we can see):
```bash
2024-07-07 08:43:35,836 - MainThread - botocore.endpoint - DEBUG - Making request for OperationModel(name=ListUsers) with params: {'url_path': '/', 'query_string': '', 'method': 'POST', 'headers': {'Content-Type': 'application/x-www-form-urlencoded; charset=utf-8', 'User-Agent': 'aws-cli/2.15.36 Python/3.11.9 Darwin/23.4.0 source/arm64 prompt/off command/iam.list-users'}, 'body': {'Action': 'ListUsers', 'Version': '2010-05-08', 'PathPrefix': '/abc/'}, 'url': 'https://localhost:7005/', 'context': {'client_region': 'us-east-1', 'client_config': <botocore.config.Config object at 0x1083b49d0>, 'has_streaming_input': False, 'auth_type': None}}
```

Same issue in list-access-keys:
`s3-nc-user-1-iam iam list-access-keys --max-items 1 --user-name shira-1001 --debug`

In the debug I see AWS CLI list the argument of max-items:
```bash
2024-07-07 09:03:15,120 - MainThread - awscli.clidriver - DEBUG - Arguments entered to CLI: ['--no-verify-ssl', '--endpoint-url', 'https://localhost:7005', 'iam', 'list-access-keys', '--max-items', '1', '--user-name', 'shira-1001', '--debug']
```

In the debug I don't see the max params passing to the request's body (UserName we can see):
```bash
2024-07-07 09:03:15,228 - MainThread - botocore.endpoint - DEBUG - Making request for OperationModel(name=ListAccessKeys) with params: {'url_path': '/', 'query_string': '', 'method': 'POST', 'headers': {'Content-Type': 'application/x-www-form-urlencoded; charset=utf-8', 'User-Agent': 'aws-cli/2.15.36 Python/3.11.9 Darwin/23.4.0 source/arm64 prompt/off command/iam.list-access-keys'}, 'body': {'Action': 'ListAccessKeys', 'Version': '2010-05-08', 'UserName': 'shira-1001'}, 'url': 'https://localhost:7005/', 'context': {'client_region': 'us-east-1', 'client_config': <botocore.config.Config object at 0x106fa89d0>, 'has_streaming_input': False, 'auth_type': None}}
```

Meanwhile, I will mention the issue in the code.

guymguym: just making sure - do you have the latest aws cli? perhaps submit an issue to its github project?

guymguym (approved):

src/endpoint/iam/ops/iam_list_users.js
shirady: I updated my `awscli` (using brew) and still face this issue.
I opened a Github issue in their repo:
https://github.com/aws/aws-cli/issues/8800

PR #8173, opened 2024-07-01
Avoid throwing an error in tier_server.update_bucket_class

dannyzaken:
### Explain the changes
1. If a newly generated tier name already exists, regenerate it instead of failing the entire operation.
2. This could happen if `update_bucket_class` is called too frequently and `new_name` is generated in the exact second.
3. The name itself has no significance, except that it is used as a unique DB index key. This is the main reason to keep it unique 

1. https://bugzilla.redhat.com/show_bug.cgi?id=2294723

### Testing Instructions:
1. perform frequent updates to the a bucketclass or the backingstores that are in its policy 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/system_services/tier_server.js
+++ b/src/server/system_services/tier_server.js
@@ -17,6 +17,7 @@ const node_allocator = require('../node_services/node_allocator');
 const system_store = require('./system_store').get_instance();
 const chunk_config_utils = require('../utils/chunk_config_utils');
 const SensitiveString = require('../../util/sensitive_string');
+const P = require('../../util/promise');
 
 function new_tier_defaults(name, system_id, chunk_config, mirrors, storage_class) {
     return _.omitBy({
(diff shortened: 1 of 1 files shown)


src/server/system_services/tier_server.js
```diff
@@ -357,7 +357,11 @@ async function update_bucket_class(req) {
                 .map(obj => obj.tier.name);
             for (const [index, tier] of tiers.entries()) {
                 const old_tier = old_tiers[index];
-                const new_name = `${bucket.name.unwrap()}.${Math.round(Date.now() / 1000)}.${index}`;
+                let new_name = `${bucket.name.unwrap()}.${Math.round(Date.now() / 1000)}.${index}`;
```
jackyalbo, line 360: Cant we just remove the division by 1k? Any chance for 2 requests in the same millisec(more than a 1/100 random?) In anyway i think we should use the same format. So if you use this random, add it also to the initial new_name

jackyalbo (approved):

PR #8174, opened 2024-07-01
[NC | NSFS] Replace flock with fcntl based locks

tangledbytes:
### Explain the changes
This PR replaces the use of `flock` with `fcntl` based locks. This was done as it was observed that `flock` does **not** work in GPFS.

### Testing Instructions:
Run regular tests and they should pass.

Some tests were performed (on GPFS, CephFS and AWS EFS) using the following code:
```c
#include <sys/fcntl.h>
#include <unistd.h>
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

void lock(int fd, int mode) {
        struct flock fl = {};
        fl.l_type = mode;
        fl.l_whence = SEEK_SET;
        fl.l_start = 0;
        fl.l_len = 0;

        printf("acquiring lock\n");
       // 38 is the value for macro F_OFD_SETLKW
        if (fcntl(fd, 38, &fl) == -1) {
                perror("fcntl");
                exit(1);
        }
        printf("acquired lock\n");
}

int main(int argc, char **argv) {
        if (argc != 3) {
                fprintf(stderr, "usage: %s file mode\n", argv[0]);
                exit(1);
        }

        if (strcmp(argv[2], "s") == 0) {
                // try to acquire a shared a lock
                int fd1 = open(argv[1], O_RDWR);
                if (fd1 == -1) {
                        perror("open");
                        exit(1);
                }
                lock(fd1, F_RDLCK);

                // get another shared lock
                int fd2 = open(argv[1], O_RDWR);
                if (fd2 == -1) {
                        perror("open");
                        exit(1);
                }
                lock(fd2, F_RDLCK);

                // close the first descriptor
                close(fd1);

                sleep(30);
        }

        if (strcmp(argv[2],"x") == 0) {
                // get an exclusive lock
                int fd = open(argv[1], O_RDWR);
                if (fd == -1) {
                        perror("open");
                        exit(1);
                }
                lock(fd, F_WRLCK);
                sleep(30);
        }

    exit(0);
}
```

The result is the following and is desirable:

https://github.com/user-attachments/assets/af68a0d4-b5c3-464c-b619-334691959895

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -14,6 +14,7 @@
 #include <map>
 #include <math.h>
 #include <stdlib.h>
+#include <sys/fcntl.h>
 #include <sys/stat.h>
 #include <sys/syscall.h>
 #include <sys/types.h>
@@ -1363,6 +1364,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
                 InstanceMethod<&FileWrap::stat>("stat"),
                 InstanceMethod<&FileWrap::fsync>("fsync"),
                 InstanceMethod<&FileWrap::flock>("flock"),
+                InstanceMethod<&FileWrap::fcntllock>("fcntllock"),
                 InstanceAccessor<&FileWrap::getfd>("fd"),
             }));
         constructor.SuppressDestruct();
@@ -1392,6 +1394,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
     Napi::Value fsync(const Napi::CallbackInfo& info);
     Napi::Value getfd(const Napi::CallbackInfo& info);
     Napi::Value flock(const Napi::CallbackInfo& info);
+    Napi::Value fcntllock(const Napi::CallbackInfo& info);
 };
 
 Napi::FunctionReference FileWrap::constructor;
@@ -1732,8 +1735,10 @@ struct FileFlock : public FSWrapWorker<FileWrap>
                 lock_mode = LOCK_EX;
             } else if (mode == "UNLOCK") {
                 lock_mode = LOCK_UN;
-            } else {
+            } else if (mode == "SHARED") {
                 lock_mode = LOCK_SH;
+            } else {
+                SetError("invalid lock type");
             }
         }
 
@@ -1744,6 +1749,45 @@ struct FileFlock : public FSWrapWorker<FileWrap>
         int fd = _wrap->_fd;
         CHECK_WRAP_FD(fd);
         SYSCALL_OR_RETURN(flock(fd, lock_mode));
+    }
+};
+
+struct FileFcntlLock : public FSWrapWorker<FileWrap>
+{
+    struct flock fl;
+    FileFcntlLock(const Napi::CallbackInfo& info)
+        : FSWrapWorker<FileWrap>(info)
+        , fl()
+    {
+        // lock entire file
+        fl.l_whence = SEEK_SET;
+        fl.l_start = 0;
+        fl.l_len = 0;
+        fl.l_pid = 0;
+        fl.l_type = F_RDLCK;
+
+        if (info.Length() > 1 && !info[1].IsUndefined()) {
+            auto mode = info[1].As<Napi::String>().Utf8Value();
+            if (mode == "EXCLUSIVE") {
+                fl.l_type = F_WRLCK;
+            } else if (mode == "UNLOCK") {
+                fl.l_type = F_UNLCK;
+            } else if (mode == "SHARED") {
+                fl.l_type = F_RDLCK;
+            } else {
+                SetError("invalid lock type");
+            }
+        }
+
+        Begin(XSTR() << "FileFcntlLock" << DVAL(_wrap->_path));
+    }
+    virtual void Work()
+    {
+        int fd = _wrap->_fd;
+        CHECK_WRAP_FD(fd);
+        // This uses F_OFD_SETLKW instead for discussion related to this choice
+        // refer: https://github.com/noobaa/noobaa-core/pull/8174
+        SYSCALL_OR_RETURN(fcntl(fd, F_OFD_SETLKW, &fl));
     }
 };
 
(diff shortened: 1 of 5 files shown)


tangledbytes:
Using `F_OFD_SETLKW` instead of `F_SETLKW` as they are file object based (per descriptor) instead of per inode per process.

Although they aren't in POSIX but they seem to be supported in MacOS as well: https://opensource.apple.com/source/xnu/xnu-6153.81.5/bsd/sys/fcntl.h.auto.html

src/util/file_reader.js
```diff
@@ -115,8 +115,8 @@ class NewlineReader {
     async init() {
         let fh = null;
         try {
-            fh = await nb_native().fs.open(this.fs_context, this.path, 'r');
-            if (this.lock) await fh.flock(this.fs_context, this.lock);
+            fh = await nb_native().fs.open(this.fs_context, this.path, '+');
```
guymguym, line 118: is RW needed for fcntl locking?

src/native/fs/fs_napi.cpp
```diff
@@ -1744,6 +1747,41 @@ struct FileFlock : public FSWrapWorker<FileWrap>
...
+        fl.l_pid = 0;
+        fl.l_type = F_RDLCK;
+
+        if (info.Length() > 1 && !info[1].IsUndefined()) {
+            auto mode = info[1].As<Napi::String>().Utf8Value();
+            if (mode == "EXCLUSIVE") {
+                fl.l_type = F_WRLCK;
+            } else if (mode == "UNLOCK") {
+                fl.l_type = F_UNLCK;
+            } else {
```
guymguym, line 1770: maybe better to also check explicitly for mode == "SHARE" and then else fail

src/util/persistent_logger.js
```diff
@@ -11,10 +11,10 @@ const dbg = require('./debug_module')(__filename);
 
 /**
  * PersistentLogger is a logger that is used to record data onto disk separated by newlines.
- * 
+ *
```
guymguym, line 14: please turn off the space changes in comments - this is causing infinit line changes as most other developers do not have this trimming for comment lines.

src/util/file_reader.js
tangledbytes: From man page:
```
       EBADF  fd is not an open file descriptor

       EBADF  op is F_SETLK or F_SETLKW and the file descriptor open
              mode doesn't match with the type of lock requested.
```

The second one causes a EBADF if the descriptor is not "compatible". That's why I had to change permissions at a few places.

src/native/fs/fs_napi.cpp
tangledbytes: Yeah, that works as well. Will update the PR.

src/util/persistent_logger.js
tangledbytes: This is due to the `.editorconfig`. I will remove these changed from this PR and will update the `.editorconfig` in another one. Sounds good?

guymguym: yes thanks

src/util/file_reader.js
guymguym: so is O_RDRW only required only for exclusive lock, and O_RDONLY can work with shared? 
just trying to understand the limitations, because in that scope we are just trying the read the file...

martin-petermann:
> Using `F_OFD_SETLKW` instead of `F_SETLKW` as they are file object based (per descriptor) instead of per inode per process.
> 
> Although they aren't in POSIX but they seem to be supported in MacOS as well: https://opensource.apple.com/source/xnu/xnu-6153.81.5/bsd/sys/fcntl.h.auto.html

Also OFD locks are listed on the fcntl man page of MacOS. Therefore, some kind of official.

src/util/file_reader.js
guymguym: Ok - so I would put a code comment to explain that we open RW for this reading flow because the need for exclusive locking

src/util/persistent_logger.js
```diff
@@ -216,7 +216,7 @@ class PersistentLogger {
 
     async _open() {
         await native_fs_utils._create_path(this.dir, this.fs_context);
-        return nb_native().fs.open(this.fs_context, this.active_path, 'as');
+        return nb_native().fs.open(this.fs_context, this.active_path, 'as+');
```
guymguym, line 219: how does a and + work together? append with rw fits?

src/native/fs/fs_napi.cpp
guymguym: please apply the same also to the existing flock code

src/util/persistent_logger.js
tangledbytes: I think so. I tested it and it seemed to work just fine. I relied on [this](https://stackoverflow.com/a/50875791) stackoverflow answer.

src/native/fs/fs_napi.cpp
```diff
@@ -1744,6 +1747,43 @@ struct FileFlock : public FSWrapWorker<FileWrap>
...
+            }
+        }
+
+        Begin(XSTR() << "FileFcntlLock" << DVAL(_wrap->_path));
+    }
+    virtual void Work()
+    {
+        int fd = _wrap->_fd;
+        CHECK_WRAP_FD(fd);
+        SYSCALL_OR_RETURN(fcntl(fd, F_OFD_SETLKW, &fl));
```
guymguym, line 1783: Lets switch to F_SETLKW and add a code comment that explains that the scope of locking we can expect to use is per process and per file, but not per fd, which means that closing any other fd of that file from the same process will unlock it. we will have to workaround this restriction to avoid having two locking flows of the same file in the same process running at the same time.

guymguym (changes requested):

src/native/fs/fs_napi.cpp
guymguym: ok, so based on testing this does seem to work with F_OFD_SETLKW.

guymguym (approved):

src/native/fs/fs_napi.cpp
guymguym: @tangledbytes it would be nice if you could add the test code to the repo or maybe as a comment on this PR, for future reference.

PR #8175, opened 2024-07-01
NSFS | NC | IAM Service - Accounts Permission When No Bucket Policy

shirady:
### Explain the changes
1. Add more properties to `nsfs_bucket_schema `(not required) and update the `new_bucket_defaults` in `BucketSpaceFS`:
- `creator` = creator is the account ID that created this bucket (internal information).
Notes:  Currently we do not allow IAM accounts to create a bucket (temporary), it will be changed after we have the config structure, therefore in the future, we could see the IAM account ID in the creator property.
2. Change the condition in `authorize_request_policy` (in `s3_rest`) and `has_bucket_action_permission` (in `bucketspace_fs`) for the same root account (alternative for ownership when there is no bucket policy).

Those next changes are not related to IAM, but were raised as a part of the code review:
3. In `authorize_request_policy` (in `s3_rest`) remove the condition `req.object_sdk.nsfs_config_root` from the owner condition.
4. In `has_bucket_action_permission` (in `bucketspace_fs`) change the condition of `is_owner` from account name to id.

### Issues: 
List of GAPs:
1. Move the mutual code of AccountSpaceFS, BucketSpaceFS, and Manage NSFS to a specific file.
2. Allow IAM users to create buckets (see [comment below](https://github.com/noobaa/noobaa-core/pull/8175#issuecomment-2254463451)).

#### Unit Tests
Please run:
1. `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace_fs.js`.
3. `npx jest test_nc_nsfs_bucket_schema_validation.test.js`.

#### Manual Tests:
Notes: 
- Before starting the server please add this line: `process.env.NOOBAA_LOG_LEVEL = 'nsfs';` in the endpoint.js (before the condition `if (process.env.NOOBAA_LOG_LEVEL) {`)
- I Change the `config.NSFS_CHECK_BUCKET_BOUNDARIES = false; //SDSD` because I'm using the `/tmp/` and not `/private/tmp/`.
4. Create the alias for IAM service: `alias s3-nc-user-1-iam='AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:7005'`.
5. Use AWS CLI to create 2 IAM accounts: 
 user1: `s3-nc-user-1-iam iam create-user --user-name Bob` + `s3-nc-user-1-iam iam create-access-key --user-name Bob`
 user2: `s3-nc-user-1-iam iam create-user --user-name Robert` + `s3-nc-user-1-iam iam create-access-key --user-name Robert`
6. Create an alias for each one of the accounts for using S3 endpoint, for example: `alias s3-nc-user-1-s3-regular='AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443'` (using different port). (the second one will be `s3-nc-user-2-s3-regular`).
7. List the buckets from user1 and user2 (expect no buckets): `s3-nc-user-1-s3-regular s3 ls`, `s3-nc-user-2-s3-regular s3 ls`
8. Create a bucket from the root account: `s3-nc-user-1-s3 s3 mb s3://bucket-01`
9. List the buckets from user1 and user2 (expect to see the bucket from both users, same root account): `s3-nc-user-1-s3-regular s3 ls`, `s3-nc-user-2-s3-regular s3 ls`.
10. Delete the bucket by user2: `s3-nc-user-2-s3-regular s3 rb s3://bucket-01/`.

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -216,6 +216,7 @@ async function authorize_request_policy(req) {
     if (!req.params.bucket) return;
     if (req.op_name === 'put_bucket') return;
 
+    // owner_account is { id: bucket.owner_account, email: bucket.bucket_owner };
     const { s3_policy, system_owner, bucket_owner, owner_account } = await req.object_sdk.read_bucket_sdk_policy_info(req.params.bucket);
     const auth_token = req.object_sdk.get_auth_token();
     const arn_path = _get_arn_from_req_path(req);
@@ -236,13 +237,17 @@ async function authorize_request_policy(req) {
 
     const is_owner = (function() {
         if (account.bucket_claim_owner && account.bucket_claim_owner.unwrap() === req.params.bucket) return true;
-        if (req.object_sdk.nsfs_config_root && account._id === owner_account.id) return true; // NC NSFS case
+        if (owner_account && owner_account.id === account._id) return true;
         if (account_identifier === bucket_owner.unwrap()) return true;
         return false;
     }());
 
     if (!s3_policy) {
-        if (is_owner) return;
+        // in case we do not have bucket policy
+        // we allow IAM account to access a bucket that is owned by their root account
+        const is_iam_account_and_same_root_account_owner = account.owner !== undefined &&
+            owner_account && account.owner === owner_account.id;
+        if (is_owner || is_iam_account_and_same_root_account_owner) return;
         throw new S3Error(S3Error.AccessDenied);
     }
     const permission = await s3_bucket_policy_utils.has_bucket_policy_permission(
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -265,6 +265,11 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             if (!sdk.requesting_account.allow_bucket_creation) {
                 throw new RpcError('UNAUTHORIZED', 'Not allowed to create new buckets');
             }
+            // currently we do not allow IAM account to create a bucket (temporary)
+            if (sdk.requesting_account.owner !== undefined) {
+            dbg.warn('create_bucket: account is IAM account (currently not allowed to create buckets)');
+                throw new RpcError('UNAUTHORIZED', 'Not allowed to create new buckets');
+            }
             if (!sdk.requesting_account.nsfs_account_config || !sdk.requesting_account.nsfs_account_config.new_buckets_path) {
                 throw new RpcError('MISSING_NSFS_ACCOUNT_CONFIGURATION');
             }
@@ -320,6 +325,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             name,
             tag: js_utils.default_value(tag, undefined),
             owner_account: account._id,
+            creator: account._id,
             system_owner: new SensitiveString(account.name),
             bucket_owner: new SensitiveString(account.name),
             versioning: config.NSFS_VERSIONING_ENABLED && lock_enabled ? 'ENABLED' : 'DISABLED',
@@ -746,10 +752,16 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
 
         // If the system owner account wants to access the bucket, allow it
         if (is_system_owner) return true;
-        const is_owner = (bucket.bucket_owner.unwrap() === account_identifier);
+        const is_owner = (bucket.owner_account && bucket.owner_account.id === account._id);
         const bucket_policy = bucket.s3_policy;
 
-        if (!bucket_policy) return is_owner;
+        if (!bucket_policy) {
+        // in case we do not have bucket policy
+        // we allow IAM account to access a bucket that that is owned by their root account
+        const is_iam_and_same_root_account_owner = account.owner !== undefined &&
+            account.owner === bucket.owner_account.id;
+            return is_owner || is_iam_and_same_root_account_owner;
+        }
         if (!action) {
             throw new Error('has_bucket_action_permission: action is required');
         }
--- a/docs/design/iam.md
+++ b/docs/design/iam.md
@@ -89,6 +89,14 @@ Source: AccessKeys
     - root account
     - all IAM users only for themselves (except the first creation that can be done only by the root account).
 
+### No Bucket Policy
+If the resource doesn’t have a bucket policy the IAM user accounts can have access to the resources of the same root account.
+For example: 
+- root account creates 2 users (both are owned by it): user1, user2 and a bucket (bucket owner: <root-account-id>, bucket creator: <account-id-user1>).
+- user1 upload a file to the bucket 
+- user2 can delete this bucket (after it is empty): although user2 is not the creator, without a bucket policy his root account is the owner so he can delete the bucket.
+Note: Currently, we do not allow users to create a bucket.
+
 ### Root Accounts Manager
 The root accounts managers are a solution for creating root accounts using the IAM API.
 
--- a/src/server/system_services/schemas/nsfs_bucket_schema.js
+++ b/src/server/system_services/schemas/nsfs_bucket_schema.js
@@ -23,6 +23,10 @@ module.exports = {
         owner_account: {
             type: 'string',
         },
+        // creator is the account id that created this bucket (internal information)
+        creator: {
+            type: 'string',
+        },
         name: {
             type: 'string',
         },
--- a/src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_schema_validation.test.js
+++ b/src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_schema_validation.test.js
@@ -100,6 +100,11 @@ describe('schema validation NC NSFS bucket', () => {
             nsfs_schema_utils.validate_bucket_schema(bucket_data);
         });
 
+        it('nsfs_bucket with creator', () => {
+            const bucket_data = get_bucket_data();
+            bucket_data.creator = '65a62e22ceae5e5f1a758ab1';
+            nsfs_schema_utils.validate_bucket_schema(bucket_data);
+        });
     });
 
     describe('bucket with additional properties', () => {
@@ -434,6 +439,14 @@ describe('schema validation NC NSFS bucket', () => {
             assert_validation(bucket_data, reason, message);
         });
 
+        it('bucket with creator as a number (instead of string)', () => {
+            const bucket_data = get_bucket_data();
+            bucket_data.creator = 123; // number instead of string
+            const reason = 'Test should have failed because of wrong type for' +
+                'creator with number (instead of string)';
+            const message = 'must be string';
+            assert_validation(bucket_data, reason, message);
+        });
     });
 
     describe('skip schema check by config test', () => {
--- a/src/test/unit_tests/test_bucketspace_fs.js
+++ b/src/test/unit_tests/test_bucketspace_fs.js
@@ -25,6 +25,7 @@ const test_bucket = 'bucket1';
 const test_not_empty_bucket = 'notemptybucket';
 const test_bucket_temp_dir = 'buckettempdir';
 const test_bucket_invalid = 'bucket_invalid';
+const test_bucket_iam_account = 'bucket-iam-account-can-access';
 
 const tmp_fs_path = path.join(TMP_PATH, 'test_bucketspace_fs');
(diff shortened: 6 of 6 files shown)


romayalon (approved):

src/manage_nsfs/manage_nsfs_cli_utils.js
```diff
@@ -203,6 +203,17 @@ function check_root_account_owns_user(root_account, account) {
 }
 
 
+/**
+ * @param {object} account
+ */
+function check_root_account(account) {
```
guymguym, line 209: you mean is_root_account

src/manage_nsfs/nc_utils.js
```diff
@@ -0,0 +1,44 @@
...
+const nb_native = require('../util/nb_native');
+const native_fs_utils = require('../util/native_fs_utils');
+
+/**
+ * get_account_by_id will return the account
+ * TODO: when we have the config files by id, we can account config by id
+ * @param {object} global_config
+ * @param {string} account_id
+ */
+async function get_account_by_id(global_config, account_id) {
```
guymguym, line 16: I think this should be an abstract method of Bucketspace, and implement it differently for bucketspace_fs/nb.

src/sdk/bucketspace_fs.js
```diff
@@ -24,6 +25,7 @@ const { get_umasked_mode, isDirectory, validate_bucket_creation,
     create_config_file, delete_config_file, get_bucket_tmpdir_full_path, folder_delete } = require('../util/native_fs_utils');
 const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 const { anonymous_access_key } = require('./object_sdk');
+const manage_nsfs_cli_utils = require('../manage_nsfs/manage_nsfs_cli_utils');
```
guymguym, line 28: this dependency doesn't seem good, we don't want the bucketspace to depend on the cli code.

src/sdk/bucketspace_fs.js
```diff
@@ -314,14 +316,30 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
     }
 
 
-    new_bucket_defaults(account, { name, tag, lock_enabled, force_md5_etag }, create_uls, bucket_storage_path) {
+    async new_bucket_defaults(account, { name, tag, lock_enabled, force_md5_etag }, create_uls, bucket_storage_path) {
```
guymguym, line 319: new_XX_defaults is typically a sync function, not async. if you need to fetch anything async for the defaults we usually do that before the call and pass it in.

src/sdk/bucketspace_fs.js
```diff
@@ -314,14 +316,30 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        const owner_account_id = is_account_root_account ? account._id : account.owner;
+        let owner_account_config;
+        // in case it is IAM account we need the name of account (we have only the id in owner property)
+        if (!is_account_root_account) {
+            // global_config specific for this function
+            const global_config = {
+                config_root_backend: this.fs_context.fs_backend,
+                accounts_dir: this.accounts_dir,
+            };
+            owner_account_config = await nc_utils.get_account_by_id(global_config, account.owner);
```
guymguym, line 332: this is bad, we are going to iterate all accounts just to find the root account name here?? this is bad, can't we just avoid saving the root account name in the bucket??

src/sdk/bucketspace_fs.js
```diff
@@ -747,9 +765,10 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
         // If the system owner account wants to access the bucket, allow it
         if (is_system_owner) return true;
         const is_owner = (bucket.bucket_owner.unwrap() === account_identifier);
+        const is_same_root_account_owner = account.owner === bucket.owner_account.id; // case for IAM account
```
guymguym, line 768: how is bucket.owner_account an object with id field? isn't it just the id itself?

guymguym (changes requested):

src/endpoint/s3/s3_rest.js
```diff
@@ -241,8 +241,10 @@ async function authorize_request_policy(req) {
         return false;
     }());
 
+    const is_same_root_account_owner = req.object_sdk.nsfs_config_root && account.owner === owner_account.id; // NC NSFS case for IAM account
```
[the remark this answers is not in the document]
shirady: 1) > why is this check separated from all the other checks in is_owner? I would keep all ownership checks together.

I thought it would be more readable to write it in different conditions: we have 1)` is_owner` (root account case) and 2) `is_same_root_account_owner` IAM account.

2) > also I don't know why we check for nsfs_config_root at all - this seems unrelated - why wouldn't we use the same check in containers?

In the function, we already had a couple of implementations related to NC NSFS specifically (see above code). Since we didn't implement the owner in containerized this value will be undefined, and anyway the comparison can be false, I can change it.

src/manage_nsfs/manage_nsfs_cli_utils.js
shirady: I don't like functions that start with `_is` (this prefix I save for a boolean), and I prefer an action within the name.
So I can change it to `check_is_root_account`.

src/manage_nsfs/nc_utils.js
shirady: This is an interesting idea, but we need to think about it more, for example:
1) Will it still be relevant after we have the new config structure?
2) How do we want to improve out code and write in one-place functions that will serve: namespace_fs, bucketspace_fs, accountspace_js and the manage_nsfs?

src/sdk/bucketspace_fs.js
shirady: I would remove it, the use is currently minor.

src/sdk/bucketspace_fs.js
shirady: Thanks, It will be changed.

src/sdk/bucketspace_fs.js
```diff
@@ -314,14 +316,30 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+        const is_account_root_account = manage_nsfs_cli_utils.check_root_account(account);
+        const owner_account_id = is_account_root_account ? account._id : account.owner;
+        let owner_account_config;
+        // in case it is IAM account we need the name of account (we have only the id in owner property)
+        if (!is_account_root_account) {
+            // global_config specific for this function
+            const global_config = {
+                config_root_backend: this.fs_context.fs_backend,
+                accounts_dir: this.accounts_dir,
+            };
```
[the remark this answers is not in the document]
shirady: I tried to reuse a function from manage_nsfs, so I named the object as same way it was named in manage_nsfs.

src/sdk/bucketspace_fs.js
shirady: The whole changes in the last commit were because I didn't have the root account name and only the id.
Those properties are planned to be removed in next version (`system_owner`, `bucket_owner`).
I can add to the schema a non-required property with the root account name on IAM creation and in the next version we can remove it (like we remove `bucket_owner`). WDYT?

src/sdk/bucketspace_fs.js
shirady: You are probably right, I will check it.

src/manage_nsfs/nc_utils.js
guymguym: 1. yes.
2. so actually I mixed up and I think it should be an abstract method of the AccountSpace from responsibility perspective. we should be able to pass the accountspace instance to bucketspace/namespace as needed if they need to call it. for manage_nsfs it should be able to use accountspace as well to share the logic.

src/sdk/bucketspace_fs.js
guymguym: so why not keep using bucket_owner field for now? isn't that the account name?

src/manage_nsfs/manage_nsfs_cli_utils.js
guymguym: perhaps you have something against it, but is_XXX is the common standard pattern for both boolean variables and boolean functions. check_XXX is a more commonly a pattern used when you need to do actual work to verify some expectation, which also typically can throw to report the check failed.

src/endpoint/s3/s3_rest.js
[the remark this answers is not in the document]
guymguym: But we want the same to apply for containerized code as well, so the separation does not help.

src/manage_nsfs/nc_utils.js
shirady: Thanks for the explanation, the implementation of this would be in a separate PR.

src/endpoint/s3/s3_rest.js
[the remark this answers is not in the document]
shirady: But currently containerized code doesn't have the IAM accounts, why should I combine it with a condition that is not relevant for it?

src/sdk/bucketspace_fs.js
shirady: In this current fix, we set the `system_owner` and `bucket_owner` with the root account name.
As mentioned I don't have the account name, so at first suggestion I iterate and searched it (on all the accounts), in this suggestion I added this property in the schema.

I don't want to use the current account name because I want to validate for 2 IAM accounts that the can access the buckets of the same root account.

src/sdk/bucketspace_fs.js
```diff
@@ -747,9 +756,11 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
         // If the system owner account wants to access the bucket, allow it
         if (is_system_owner) return true;
         const is_owner = (bucket.bucket_owner.unwrap() === account_identifier);
+        const is_iam_and_same_root_account_owner = account.owner !== undefined &&
+            account.owner_name.unwrap() === bucket.bucket_owner.unwrap(); // case for IAM account
```
guymguym, line 760: use bucket.owner_account (id) instead of name

src/sdk/bucketspace_fs.js
```diff
@@ -315,13 +316,18 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
             _id: mongo_utils.mongoObjectId(),
             name,
             tag: js_utils.default_value(tag, undefined),
-            owner_account: account._id,
-            system_owner: new SensitiveString(account.name),
-            bucket_owner: new SensitiveString(account.name),
+            owner_account: owner_account,
+            creator: account._id,
+            system_owner: new SensitiveString(account.name), // TODO - change it (for IAM account it is wrong)
+            bucket_owner: new SensitiveString(account.name), // TODO - change it (for IAM account it is wrong)
```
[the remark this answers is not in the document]
shirady: In the last code changes, we removed the option to create a bucket from the IAM account.
After we have the account by ID structure, we will return this option.

src/endpoint/s3/s3_rest.js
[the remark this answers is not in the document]
shirady: The part `req.object_sdk.nsfs_config_root` was removed from the condition.

src/sdk/bucketspace_fs.js
shirady: changed, thanks.

guymguym (approved):

shirady:
### Intern Documentation of the PR:
Before adding the limitation of bucket creation on IAM accounts - the options were discussed in this PR related to the function `new_bucket_defaults` (in the file `bucketspace_fs.js`):
1.  Copy the `system_owner` and the `bucket_owner` from the IAM account - the worst solution, even if we want to remove those properties in the future, attaching them with the IAM account details is wrong.
2. Copy the `system_owner` and the `bucket_owner` from the root account - those fields are the names, and when we have the IAM account config we have the property of `owner` which is ID so to get it we thought of:
(1) Iterate all the accounts and search for the account that has this ID - this would damage the performance of this operation.
(2) Add temporary property `owner_name` in the account schema and config - we would still have issues with accounts that change their name.

Therefore, we add the limitation for creating a bucket by the IAM account, that we will remove once we could get the root account config by ID.

PR #8184, opened 2024-07-04
Replace yamljs with yaml

liranmauda:
### Explain the changes
1. Replace yamljs with yaml


liranmauda:
Replacing `yamljs` with `yaml` as `yaml` is maintained and `yamljs` is less maintained.
This is the first step, in the second step we should avoid using any kubernetes calls in the core 
@dannyzaken @jackyalbo WDYT?

jackyalbo (approved):

PR #8185, opened 2024-07-04
NC | Noobaa debug and normal log sent to two different files.

naveenpaul1:
### Explain the changes
1. Current implementation all the logs are pushed to both `iamjournal` and `syslog`

- syslog facility local0 - [rsyslog omfile](https://www.rsyslog.com/doc/configuration/modules/omfile.html) copies it to /var/log/noobaa.log
- stdout/stderr - [systemd copies it into its journal](https://www.freedesktop.org/software/systemd/man/latest/systemd-journald.service.html#) and then [rsyslog imjournal](https://www.rsyslog.com/doc/configuration/modules/imjournal.html) copies it to the file /var/log/messages

1. all the low level logs(ERROR, INFO, LOG, TRACE, L0) are pushed to var/log/message. Can be configured using property `LOG_TO_STDERR_LEVEL`), 
2. When debug enabled(LOG_TO_SYSLOG_ENABLED) all the messages will pushed to var/log/noobaa.log, Log level can be configued in using property LOG_TO_DEBUGLOG_LEVEL

1. https://github.com/noobaa/noobaa-core/issues/8032

### Testing Instructions:
1. 

- [] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/system_services/schemas/nsfs_config_schema.js
+++ b/src/server/system_services/schemas/nsfs_config_schema.js
@@ -35,8 +35,13 @@ const nsfs_node_config_schema = {
         },
         NOOBAA_LOG_LEVEL: {
             type: 'string',
-            enum: ['warn', 'default', 'nsfs', 'all'],
+            enum: ['warn', 'default', 'nsfs', 'all', 'off'],
             doc: 'logging verbosity level for the NooBaa system, service restart required'
+        },
+        NOOBAA_DEBUG_LEVEL: {
+            type: 'string',
+            enum: ['warn', 'default', 'nsfs', 'all', 'off'],
+            doc: 'logging verbosity level for the NooBaa debug logs, service restart required'
         },
         UV_THREADPOOL_SIZE: {
             type: 'number',
--- a/src/util/debug_module.js
+++ b/src/util/debug_module.js
@@ -35,6 +35,15 @@ if (process.env.NOOBAA_LOG_LEVEL) {
     }
 }
 
+let dbg_syslog_conf;
+let LOG_DEBUG_LEVEL = 0;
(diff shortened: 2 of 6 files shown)


src/util/debug_module.js
```diff
@@ -379,7 +389,8 @@ class InternalDebugLogger {
     }
 
     log_internal(msg_info) {
-        if (syslog && config.LOG_TO_SYSLOG_ENABLED) {
+        if (syslog && config.LOG_TO_SYSLOG_ENABLED &&
+            this._levels[msg_info.level] < LOG_DEBUG_LEVEL) {
```
romayalon, line 393: @naveenpaul1 
On syslog condition shouldn't it be <= ? 
Asking because on stderr we skipped every msg above LOG_LEVEL, so we printed msgs if this._levels[msg_info.level] <= LOG_LEVEL

src/server/system_services/schemas/nsfs_config_schema.js
```diff
@@ -35,7 +35,7 @@ const nsfs_node_config_schema = {
         },
         NOOBAA_LOG_LEVEL: {
```
romayalon, line 36: should we expose NOOBAA_DEBUG_LEVEL on the schema?

src/util/debug_module.js
naveenpaul1: @romayalon you are right, We need all the logs till `LOG_DEBUG_LEVEL` , Will update,

src/server/system_services/schemas/nsfs_config_schema.js
naveenpaul1: yes, On it

PR #8189, opened 2024-07-07
NC | CLI | Add Diagnose command

romayalon:
### Explain the changes
1. Add NooBaa diagnose CLI command - 
```
noobaa-cli diagnose health
noobaa-cli diagnose gather-logs
noobaa-cli diagnose metrics
```
2. `Health script` moved to be a part of `noobaa-cli` and it's no longer a standalone script. Output is wrapped by `noobaa-cli` response/errors.
3. Added `gather-logs` action (logic unimplemented yet).
4. Added `metrics` action for fetching NooBaa NC metrics.

1. gather-logs logic.

### Testing Instructions:
1. Manual - 
Install NooBaa RPM, enable the service and run the following commands - 

```
noobaa-cli diagnose health
noobaa-cli diagnose gather-logs
noobaa-cli diagnose metrics
```
2. Automatic - 
` sudo jest --testRegex=jest_tests/test_cli_diagnose`
`sudo ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nc_nsfs_health.js`

- [x] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/manage_nsfs/diagnose.js
+++ b/src/manage_nsfs/diagnose.js
@@ -1,0 +1,71 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+
+const dbg = require('../util/debug_module')(__filename);
+const health = require('./health');
+const config = require('../../config');
+const http_utils = require('../util/http_utils');
+const buffer_utils = require('../util/buffer_utils');
+const { DIAGNOSE_ACTIONS } = require('./manage_nsfs_constants');
+const ManageCLIError = require('./manage_nsfs_cli_errors').ManageCLIError;
+const { throw_cli_error, write_stdout_response } = require('./manage_nsfs_cli_utils');
+const ManageCLIResponse = require('../manage_nsfs/manage_nsfs_cli_responses').ManageCLIResponse;
+
+/**
+ * manage_diagnose_operations handles cli diagnose operations
+ * @param {string} action 
+ * @param {Object} user_input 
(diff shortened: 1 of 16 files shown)


src/manage_nsfs/diagnose.js
```diff
@@ -0,0 +1,64 @@
...
+async function gather_logs() {
+    throw_cli_error(ManageCLIError.NotImplemented);
+}
+
+/**
+ * gather_metrics handles cli diagnose metrics operation
+ * @returns {Promise<Void>}
+ */
+async function gather_metrics() {
+    const metrics_query = `curl -s http://127.0.0.1:${config.EP_METRICS_SERVER_PORT}/metrics/nsfs_stats`;
```
guymguym, line 49: I think relying on external curl on the host is not needed and we should use http.request where possible.

we should add a function like this in http_utils (like the make_https_request we already have):
```js
async function make_http_request(options, body) {
    return new Promise((resolve, reject) => {
        http.request(options, resolve)
            .on('error', reject)
            .end(body);
    });
}
```

and then can call -
```js
const res = await http_utils.make_http_request({
    hostname: 'localhost',
    port: config.EP_METRICS_SERVER_PORT,
    path: '/metrics/nsfs_stats',
});
const metrics_output = (await buffer_utils.read_stream_join(res)).toString();

```

guymguym (approved):
This is great!

PR #8192, opened 2024-07-10
NC | NSFS | Stop Setting `system_owner` in Bucket Config

### LINKED ISSUE #7793: NSFS | NC | CLI | Remove system owner from bucket config files and bucket schema
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
The system owner is a bucket property, kept on the bucket level and a part of the bucket schema.

### Expected behavior
1. The system owner is not a bucket property, and therefore it should not be kept on the bucket config file. We should consider having a system owner kept on system.json if needed. The system owner is often needed as a "root" account to be able to get access across the system. 
2. Keep attention that on bucketspace_fs.js -  read_bucket_sdk_info() we will need to "load" a system owner but not from the bucket config file.

shirady:
### Explain the changes
1. Remove the property `system_owner` as required property in `nsfs_bucket_schema`.
2. On the newly created bucket, do not set the `system_owner` value (on `manage_nsfs` and `bucketspace_fs`).
3. Remove the `system_owner` (with the value of the account name) from the current docs in NC NSFS.
4. Update unit test tests and remove the property `system_owner`.
5. Add the function `modify_bucket_on_read` in class `ConfigFS`, and use it from `get_bucket_by_name` and inside the bucket list in noobaa-cli.

### Issues: Fixed #7793 [BZ 2280212](https://bugzilla.redhat.com/show_bug.cgi?id=2280212)
1. Currently, in the NC NSFS bucket config the `system_owner` is the account name, and therefore bucket policies can't limit a bucket owner's access to buckets that he created.

#### Unit Tests:
Please run:
1. `sudo NC_CORETEST=true node --trace-warnings ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_s3_bucket_policy.js` (for NC environment, currently you will 3 failing tests - not related to this PR)
3. `make CONTAINER_PLATFORM=linux/arm64 run-single-test testname=test_s3_bucket_policy.js` (for Containerized environment)
Note: I have MacOS M1 so I need to add the flag `CONTAINER_PLATFORM=linux/arm64`.
4. `npx jest test_nc_nsfs_bucket_schema_validation.test.js`
6. `sudo node ./node_modules/.bin/_mocha src/test/unit_tests/test_nc_nsfs_cli.js`
7. `npx jest test_config_fs.test.js`

#### Manual Tests:
(1) On new buckets:
1. First, create the `FS_ROOT` and a directory for a `new_buckets_path`: `mkdir -p /tmp/nsfs_root1/` and give permissions `chmod 777 /tmp/nsfs_root1/`.
2. Create an account: `sudo node src/cmd/manage_nsfs account add --name <account-name> --uid <uid> --gid <gid> --new_buckets_path <new_buckets_path>` (use those credentials in the alias step below).
3. Start the NSFS server with: `sudo node src/cmd/nsfs --debug 5`
4. Create the alias for the created account in the S3 service: `alias s3-nc-user-1='AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443/'`
Note: I Change the `config.NSFS_CHECK_BUCKET_BOUNDARIES = false; //SDSD` because I'm using the `/tmp/` and not `/private/tmp/`.
5. Check the connection to the server (should not have buckets): `s3-nc-user-1 s3 ls`
8. Create a bucket: `s3-nc-user-1 s3 mb s3://my-bucket-to-check`
9. Create a file: `touch denied_test_obj.txt`
10. Upload the file to the bucket: `s3-nc-user-1 s3 cp denied_test_obj.txt s3://my-bucket-to-check`
11. See that the user can download the file: `s3-nc-user-1 s3 cp s3://my-bucket-to-check/denied_test_obj.txt ~/Downloads/`
12. Put bucket policy: `s3-nc-user-1 s3api put-bucket-policy --bucket my-bucket-to-check --policy file://policy.json`, the policy is:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Deny",
            "Action": "s3:*",
            "Principal": {
                "AWS": "*"
            },
            "Resource": "arn:aws:s3:::my-bucket-to-check/denied_test_obj.txt"
        }
    ]
}
```
(this policy would deny a bucket owner account from accessing an object inside its bucket)
13. See that you cannot download the file: `s3-nc-user-1 s3 cp s3://my-bucket-to-check/denied_test_obj.txt ~/Downloads/` (I saw: fatal error: An error occurred (403) when calling the HeadObject operation: Forbidden).
14. (optional) Delete the bucket policy and try again (should download the file):
  - `s3-nc-user-1 s3api delete-bucket-policy --bucket my-bucket-to-check`
  - `s3-nc-user-1 s3 cp s3://my-bucket-to-check/denied_test_obj.txt ~/Downloads/`

(2) On buckets that already exist and had the `system_owner` property - they would still have the mentioned issue.
So want to test that we can operate on buckets that has the property `system_owner`:
1. First, create the `FS_ROOT` and a directory for a `new_buckets_path`: `mkdir -p /tmp/nsfs_root1/` and give permissions `chmod 777 /tmp/nsfs_root1/`.
2. Create an account: `sudo node src/cmd/manage_nsfs account add --name shira-1001 --uid 1001 --gid 1001 --new_buckets_path /tmp/nsfs_root1/` (use those credentials in the alias step below).
3. Start the NSFS server with: `sudo node src/cmd/nsfs --debug 5`
4. Create the alias for the created account in the S3 service: `alias s3-nc-user-1='AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443/'`
Note: I Change the `config.NSFS_CHECK_BUCKET_BOUNDARIES = false; //SDSD` because I'm using the `/tmp/` and not `/private/tmp/`.
5. Check the connection to the server (should not have buckets): `s3-nc-user-1 s3 ls`
6. Create a bucket using the CLI (it doesn't matter if CLI or S3 flow for this purpose): ` sudo node src/cmd/manage_nsfs bucket add --name my-bucket --path /tmp/nsfs_root1/my-bucket --owner shira-1001`
7. List the bucket: `s3-nc-user-1 s3 ls`
8. Edit the bucket config file using `sudo vi /etc/noobaa.conf.d/buckets/my-bucket.json` and add to it: `"system_owner":"shira-1001"`
9. List the bucket: `s3-nc-user-1 s3 ls` (to test that we can list the bucket even if we have this property).
10. Create a file: `touch denied_test_obj.txt`
8. Upload the file to the bucket: `s3-nc-user-1 s3 cp denied_test_obj.txt s3://my-bucket`
9. See that the user can download the file: `s3-nc-user-1 s3 cp s3://my-bucket/denied_test_obj.txt ~/Downloads/`
* As mentioned on buckets that were already created in the system, could not use bucket policy to limit a bucket owner's access to buckets that he created.

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -230,7 +230,7 @@ async function authorize_request_policy(req) {
 
     const account = req.object_sdk.requesting_account;
     const account_identifier = req.object_sdk.nsfs_config_root ? account.name.unwrap() : account.email.unwrap();
-    const is_system_owner = account_identifier === system_owner.unwrap();
+    const is_system_owner = Boolean(system_owner) && system_owner.unwrap() === account_identifier;
 
     // @TODO: System owner as a construct should be removed - Temporary
     if (is_system_owner) return;
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -121,7 +121,6 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             };
 
             bucket.name = new SensitiveString(bucket.name);
-            bucket.system_owner = new SensitiveString(bucket.system_owner);
             bucket.bucket_owner = new SensitiveString(bucket.bucket_owner);
             bucket.owner_account = {
                 id: bucket.owner_account,
@@ -291,7 +290,6 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             tag: js_utils.default_value(tag, undefined),
             owner_account: account._id,
             creator: account._id,
-            system_owner: new SensitiveString(account.name),
             bucket_owner: new SensitiveString(account.name),
             versioning: config.NSFS_VERSIONING_ENABLED && lock_enabled ? 'ENABLED' : 'DISABLED',
             object_lock_configuration: config.WORM_ENABLED ? {
@@ -675,7 +673,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
         const account_identifier = account.name.unwrap();
         dbg.log1('has_bucket_action_permission:', bucket.name.unwrap(), account_identifier, bucket.bucket_owner.unwrap());
 
-        const is_system_owner = account_identifier === bucket.system_owner.unwrap();
+        const is_system_owner = Boolean(bucket.system_owner) && bucket.system_owner.unwrap() === account_identifier;
 
         // If the system owner account wants to access the bucket, allow it
         if (is_system_owner) return true;
--- a/src/sdk/object_sdk.js
+++ b/src/sdk/object_sdk.js
@@ -194,7 +194,7 @@ class ObjectSDK {
         const { bucket } = await bucket_namespace_cache.get_with_cache({ sdk: this, name });
         const policy_info = {
             s3_policy: bucket.s3_policy,
-            system_owner: bucket.system_owner,
+            system_owner: bucket.system_owner, // note that bucketspace_fs currently doesn't return system_owner
             bucket_owner: bucket.bucket_owner,
             owner_account: bucket.owner_account, // in NC NSFS this is the account id that owns the bucket
         };
--- a/src/server/system_services/schemas/nsfs_bucket_schema.js
+++ b/src/server/system_services/schemas/nsfs_bucket_schema.js
@@ -7,7 +7,6 @@ module.exports = {
     required: [
         '_id',
         'name',
-        'system_owner',
         'bucket_owner',
         'owner_account',
         'versioning',
@@ -30,6 +29,10 @@ module.exports = {
         name: {
             type: 'string',
         },
+        /** 
+         * @deprecated system_owner is kept for backward compatibility,
+         * but will no longer be included in new / updated bucket json.
+         */
         system_owner: {
             type: 'string',
         },
--- a/docs/NooBaaNonContainerized/CI&Tests.md
+++ b/docs/NooBaaNonContainerized/CI&Tests.md
@@ -103,6 +103,7 @@ The following is a list of `NC jest tests` files -
 8. `test_nc_nsfs_bucket_schema_validation.test.js` - Tests NC bucket schema validation.  
 9. `test_nc_nsfs_account_schema_validation.test.js` - Tests NC account schema validation.  
 10. `test_nc_nsfs_new_buckets_path_validation.test.js` - Tests new_buckets_path RW access.  
+11. `test_config_fs.test.js` - Tests ConfigFS methods.
 
 #### nc_index.js File
 * The `nc_index.js` is a file that runs several NC and NSFS mocha related tests.  
--- a/docs/NooBaaNonContainerized/GettingStarted.md
+++ b/docs/NooBaaNonContainerized/GettingStarted.md
@@ -187,7 +187,7 @@ make_bucket: s3bucket
 #### 3. Check that the bucket configuration file was created successfully -
 ```sh
 sudo cat /etc/noobaa.conf.d/buckets/s3bucket.json
-{"_id":"65cb1efcbec92b33220112d7","name":"s3bucket","owner_account":"65cb1e7c9e6ae40d499c0ae4","system_owner":"account1","bucket_owner":"account1","versioning":"DISABLED","creation_date":"2023-09-26T05:56:16.252Z","path":"/tmp/fs1/s3bucket","should_create_underlying_storage":true}
+{"_id":"65cb1efcbec92b33220112d7","name":"s3bucket","owner_account":"65cb1e7c9e6ae40d499c0ae4","bucket_owner":"account1","versioning":"DISABLED","creation_date":"2023-09-26T05:56:16.252Z","path":"/tmp/fs1/s3bucket","should_create_underlying_storage":true}
 ```
 
 #### 4. Check that the underlying file system bucket directory was created successfully -
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -94,7 +94,6 @@ async function fetch_bucket_data(action, user_input) {
         _id: undefined,
         name: user_input.name === undefined ? undefined : String(user_input.name),
         owner_account: undefined,
-        system_owner: user_input.owner, // GAP - needs to be the system_owner (currently it is the account name)
         bucket_owner: user_input.owner,
         tag: undefined, // if we would add the option to tag a bucket using CLI, this should be changed
         versioning: action === ACTIONS.ADD ? 'DISABLED' : undefined,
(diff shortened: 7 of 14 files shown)


src/server/system_services/schemas/nsfs_bucket_schema.js
```diff
@@ -26,6 +25,7 @@ module.exports = {
         name: {
             type: 'string',
         },
+        // temp, keep the system_owner where it was already added (remove it in the future)
```
romayalon, line 28: @shirady @guymguym 
Are we all ok with not removing system_owner instead of upgrading the existing bucket config files?
Should we upgrade the files?

romayalon (changes requested):

src/sdk/bucketspace_fs.js
```diff
@@ -320,7 +320,6 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             name,
             tag: js_utils.default_value(tag, undefined),
             owner_account: account._id,
-            system_owner: new SensitiveString(account.name),
```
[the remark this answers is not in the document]
shirady: The implementation changed, if you think something is still missing, please let me know.

src/sdk/object_sdk.js
```diff
@@ -194,7 +194,7 @@ class ObjectSDK {
         const { bucket } = await bucket_namespace_cache.get_with_cache({ sdk: this, name });
         const policy_info = {
             s3_policy: bucket.s3_policy,
-            system_owner: bucket.system_owner,
+            system_owner: bucket.system_owner, // in NC NSFS it would be undefined (wrapped in a SensitiveString)
```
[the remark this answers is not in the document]
shirady: I deleted the wrapping an undefined with a sensitive string, it will be just undefined.

src/server/system_services/schemas/nsfs_bucket_schema.js
guymguym: Regarding removal for existing buckets - INHO we should have a single function in our codebase (in config_dir module or a bucketspace method...) that reads a bucket.json and it will immediately delete any old properties right there during load, so that then any read-modify-write flow that uses it, will always write the updated json without it.

Also we can use jsdoc @deprecated -
```suggestion
        /** 
         * @deprecated system_owner is kept for backward compatibility, 
         * but will no longer be included in new / updated bucket json.
         */
```

src/sdk/object_sdk.js
```diff
@@ -194,7 +194,7 @@ class ObjectSDK {
         const { bucket } = await bucket_namespace_cache.get_with_cache({ sdk: this, name });
         const policy_info = {
             s3_policy: bucket.s3_policy,
-            system_owner: bucket.system_owner,
+            system_owner: bucket.system_owner, // in NC NSFS we don't have this value
```
guymguym, line 197: minor: we might add this to bucketspace_fs in the future, so I would say -
```suggestion
            system_owner: bucket.system_owner, // note that bucketspace_fs currently doesn't return system_owner
```

src/sdk/bucketspace_fs.js
```diff
@@ -156,7 +156,8 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             };
 
             bucket.name = new SensitiveString(bucket.name);
-            bucket.system_owner = new SensitiveString(bucket.system_owner);
+            // we want to remove the system_owner, but we might have buckets that were created in the past that had this property
+            if (bucket.system_owner) bucket.system_owner = new SensitiveString(bucket.system_owner);
```
guymguym, line 160: I think this code is not needed now. 
We can just ignore/remove it from the object when loading bucket info from json.

guymguym (changes requested):

src/server/system_services/schemas/nsfs_bucket_schema.js
shirady: @guymguym Regarding the deprecated tag - in the JSDOC example I can see that they mention since which version it was deprecated, don't we want to add it as well? ("deprecated since version 5.18").

A couple of questions regarding the suggestion to add the read-modify-write:
1. What about the entries in the `bucket_namespace_cache` that already have this property? Can we assume that the cache is empty because each node was upgraded?
2. The function `read_bucket_sdk_info` is used when listing the buckets - it would mean that we will read and write all the buckets on the first list-buckets, and I'm not sure we would like that.
3. In general - why not leave it read-modify and the write will happen only when updating a bucket?
cc: @romayalon

src/sdk/bucketspace_fs.js
shirady: I need it until I will implement the read-modify-write

romayalon (approved):

src/server/system_services/schemas/nsfs_bucket_schema.js
guymguym: sure, adding the version doesn't hurt.
1 - the cache is in memory, it will be cleared on upgrade.
2,3 - that's what I meant, we shouldn't write unless there is a need to update the bucket. in list/read flows we should only modify the object in memory so in case we need to write that it will be updated correctly, and also that any code that looks at this object in memory will see the correct strcture.

src/endpoint/s3/s3_rest.js
```diff
@@ -230,7 +230,7 @@ async function authorize_request_policy(req) {
 
     const account = req.object_sdk.requesting_account;
     const account_identifier = req.object_sdk.nsfs_config_root ? account.name.unwrap() : account.email.unwrap();
-    const is_system_owner = account_identifier === system_owner.unwrap();
+    const is_system_owner = Boolean(system_owner) && system_owner.unwrap() === account_identifier;
 
     // @TODO: System owner as a construct should be removed - Temporary
```
guymguym, line 235: this todo is still needed?

src/server/system_services/schemas/nsfs_bucket_schema.js
shirady: I assumed the answers to this comment yesterday the code is updated - you can review it.
regarding the version - I added it in internal function  `adjust_bucket_with_schema_updates` and I don't want to write it twice, so it will appear there.

guymguym (approved):

src/endpoint/s3/s3_rest.js
shirady: I need to check, if needed will be handled in a different PR.

tangledbytes: I added this comment when I was working on removing `allowed_buckets`. This was added as a way to fix multiple authorization issue that started popping up primarily because some flows expected system owner to have "superuser" privileges so this particular fix addressed that.

If this PR is addressing the same concerns then maybe we can remove this comment. Moreover, I no longer remember why we thought that we can totally get rid of the `system_owner` construct.

shirady: @tangledbytes In this PR we remove the `system_owner` as a property in the bucket config - so it addresses it only in the NC NSFS environment. If I understand you, your change was in the Containerized environment, so I cannot remove this comment, right?

BTW, what does it mean "System owner as a construct"?

PR #8196, opened 2024-07-10
Implement pool ownership

Neon-White:
### Explain the changes
1. Each pool and namespace resource now contain a new `owner_id` field that's used for access enforcement
2. Only a resource's owner and system owners (with the `operator` role) can delete resources with the field
3. The deletion check from the previous point can be toggled on and off with the `POOL_DELETION_ONLY_BY_OWNER_AND_OPERATOR` environment variable
4. Implement a `config.js` variable that allows toggling the check in point 3

1. This implementation supports [RHSTOR-5187 - Support Provider & Consumer mode](https://issues.redhat.com/browse/RHSTOR-5187)

### Testing Instructions:
1. WIP

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/server/system_services/pool_server.js
+++ b/src/server/system_services/pool_server.js
@@ -88,12 +88,25 @@ function set_pool_controller_factory(pool_controller_factory) {
 
 }
 
-function new_pool_defaults(name, system_id, resource_type, pool_node_type) {
+// The function checks whether the pool/resource has an owner
+// and only allows deletion in case that the owner is also the requester of the deletion
+function check_deletion_ownership(req, resource_owner_id) {
+    if (config.RESTRICT_RESOURCE_DELETION) {
+        const requester_is_sys_owner = String(req.account._id) === String(req.system.owner._id);
+        if (!requester_is_sys_owner && String(resource_owner_id) !== String(req.account._id)) {
+            dbg.error('check_deletion_ownership: requester (', req.account._id, ') is not the owner (', resource_owner_id, ') of the resource');
+            throw new RpcError('UNAUTHORIZED', 'The pool or resource can be deleted only by its owner or the system administrator');
+        }
+    }
+}
+
+function new_pool_defaults(name, system_id, resource_type, pool_node_type, owner_id) {
     const now = Date.now();
     return {
         _id: system_store.new_system_store_id(),
         system: system_id,
         name: name,
+        owner_id,
         resource_type: resource_type,
         pool_node_type: pool_node_type,
         storage_stats: {
@@ -127,7 +140,7 @@ async function create_hosts_pool(req) {
         dbg.error('create_hosts_pool: get_partial_stats failed with', error);
     }
     const { system, rpc_params, account } = req;
-    const pool = new_pool_defaults(rpc_params.name, system._id, 'HOSTS', 'BLOCK_STORE_FS');
+    const pool = new_pool_defaults(rpc_params.name, system._id, 'HOSTS', 'BLOCK_STORE_FS', req.account._id);
     const MIN_PV_SIZE_GB = 16;
     const PV_SIZE_GB = 20;
     const GB = 1024 ** 3;
@@ -148,6 +161,7 @@ async function create_hosts_pool(req) {
             }
         })
     );
+
     dbg.log0('create_hosts_pool: Creating new pool', pool);
     await system_store.make_changes({
         insert: {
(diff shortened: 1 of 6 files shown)


src/server/system_services/pool_server.js
```diff
@@ -88,6 +88,14 @@ function set_pool_controller_factory(pool_controller_factory) {
 
 }
 
+// The function checks whether the pool/resource has an owner
+// and only allows deletion in case that the owner is also the requester of the deletion
+function check_deletion_ownership(req, resource_owner_id) {
+    if (resource_owner_id && resource_owner_id !== String(req.account._id)) {
+        throw new RpcError('UNAUTHORIZED', 'The namespace resource can be deleted only by its owner');
```
liranmauda, line 95: namespace?

src/api/pool_api.js
```diff
@@ -121,6 +129,10 @@ module.exports = {
                             },
                         }
                     },
+                    // This boolean field dictates whether the RPC requester should be saved in the DB as the owner of the resource
```
dannyzaken, line 132: Why can't we store it for every resource and eliminate this flag?

src/server/system_services/pool_server.js
```diff
@@ -88,6 +88,14 @@ function set_pool_controller_factory(pool_controller_factory) {
 
 }
 
+// The function checks whether the pool/resource has an owner
+// and only allows deletion in case that the owner is also the requester of the deletion
+function check_deletion_ownership(req, resource_owner_id) {
```
dannyzaken, line 93: If I recall correctly, the plan was to add the `owner_id` to the schema and not enforce any access control at this stage. Did anything in the plan change? 
I don't mind adding it now unless there are other concerns. In any case, we should probably at least have a flag in `config.js` to toggle it, which defaults to false (don't check). 

@nimrod-becker @liranmauda?

liranmauda: I agree with having a flag on the `config.js`
I do think that the ask was to enforce it.
@nimrod-becker ?

src/api/pool_api.js
liranmauda: Upgrades?

src/server/system_services/pool_server.js
nimrod-becker: I think we said in the meeting that we will set only for remote and then check only if exists.
However after a conversation with Danny, I think we can go with setting always and checking according to config.js (which actually means an env set to replace if needed)

src/server/system_services/pool_server.js
```diff
@@ -88,12 +88,26 @@ function set_pool_controller_factory(pool_controller_factory) {
...
 }
 
-function new_pool_defaults(name, system_id, resource_type, pool_node_type) {
+// The function checks whether the pool/resource has an owner
+// and only allows deletion in case that the owner is also the requester of the deletion
+function check_deletion_ownership(req, resource_owner_id) {
+    if (config.RESTRICT_RESOURCE_DELETION) {
+        const requester_is_sys_owner = String(req.account._id) === String(req.system.owner._id);
+        if (!requester_is_sys_owner && String(resource_owner_id) !== String(req.account._id)) {
+            throw new RpcError('UNAUTHORIZED', 'The pool or resource can be deleted only by its owner or the system administrator');
```
dannyzaken, line 97: Print an error message to log including info on the req account vs the owner account

dannyzaken:
one small comment, and don't forget to squash. Let me know once the tests pass, I will approve

src/server/system_services/pool_server.js
```diff
@@ -88,12 +88,27 @@ function set_pool_controller_factory(pool_controller_factory) {
...
-function new_pool_defaults(name, system_id, resource_type, pool_node_type) {
+// The function checks whether the pool/resource has an owner
+// and only allows deletion in case that the owner is also the requester of the deletion
+function check_deletion_ownership(req, resource_owner_id) {
+    if (config.RESTRICT_RESOURCE_DELETION) {
+        const requester_is_sys_owner = String(req.account._id) === String(req.system.owner._id);
+        if (!requester_is_sys_owner && String(resource_owner_id) !== String(req.account._id)) {
+            dbg.error('check_deletion_ownership: requester (', req.account._id, ') is not the owner (', resource_owner_id, ') of the resource');
+            throw new RpcError('UNAUTHORIZED', 'The pool or resource can be deleted only by its owner or the system administrator');
+        }
```
dannyzaken, line 99: no reason for a return value here. in the tests you can either call the function that will throw if failed, or use `assert.doesNotThrow()` to be more explicit

dannyzaken (approved):

Issue #7226, opened 2023-03-08
NSFS | MPU | GET/HEAD OBEJCT | Support of part-number parameter


romayalon:
### Environment info
- NooBaa Version: **master**
- Platform: All

### Actual behavior
1. Currently when a user passes part-number parameter on Get/Head object requests, NooBaa doesn't support it and returns the whole object instead of the wanted part. This is applicable to completed multipart uploads.

### Expected behavior
1. NooBaa should return the correct response when receiving part-number parameter from the user on Get/Head object requests.

### Steps to reproduce
1. install noobaa
2. use an s3 client and call s3 get object while specifying part number - 
Example using s3 cli - s3 get-object --bucket=<BUCKET> --key=<KEY> --part-number=<PART-NUM> outfile

romayalon:
Closing, @nadavMiz fixed it on #8214

PR #8214, opened 2024-07-18
NSFS | MPU | Support of part-number parameter for get/head object for parts with the same size

nadavMiz:
### Explain the changes
1. add object xattr for multipart sizes
2. add get and head object for specific part using the part size

1. https://github.com/noobaa/noobaa-core/issues/7226 

### Testing Instructions:
1. run `sudo node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_namespace_fs.js`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -55,6 +55,7 @@ const XATTR_NOOBAA_INTERNAL_PREFIX = XATTR_USER_PREFIX + 'noobaa.';
 const XATTR_NOOBAA_CUSTOM_PREFIX = XATTR_NOOBAA_INTERNAL_PREFIX + 'tag.';
 // TODO: In order to verify validity add content_md5_mtime as well
 const XATTR_MD5_KEY = XATTR_USER_PREFIX + 'content_md5';
+const XATTR_MULTIPART_SIZE = XATTR_NOOBAA_INTERNAL_PREFIX + 'multipart_size';
 const XATTR_CONTENT_TYPE = XATTR_NOOBAA_INTERNAL_PREFIX + 'content_type';
 const XATTR_PART_OFFSET = XATTR_NOOBAA_INTERNAL_PREFIX + 'part_offset';
 const XATTR_PART_SIZE = XATTR_NOOBAA_INTERNAL_PREFIX + 'part_size';
@@ -869,7 +870,7 @@ class NamespaceFS {
                 if (r.common_prefix) {
                     res.common_prefixes.push(r.key);
                 } else {
-                    obj_info = this._get_object_info(bucket, r.key, r.stat, 'null', false, true);
+                    obj_info = this._get_object_info(bucket, r.key, r.stat, 'null', false, null, true);
                     if (!list_versions && obj_info.delete_marker) {
                         continue;
                     }
@@ -921,7 +922,9 @@ class NamespaceFS {
                 }
             }
             this._throw_if_delete_marker(stat);
-            return this._get_object_info(params.bucket, params.key, stat, params.version_id || 'null', isDir);
+            this._handle_read_part_number(params, stat);
+            const range = isNaN(params.start) ? null : {start: params.start, end: params.end};
+            return this._get_object_info(params.bucket, params.key, stat, params.version_id || 'null', isDir, range);
         } catch (err) {
             if (this._should_update_issues_report(params, file_path, err)) {
                 this.run_update_issues_report(object_sdk, err);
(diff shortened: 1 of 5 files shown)


src/sdk/namespace_fs.js
```diff
@@ -1111,6 +1112,27 @@ class NamespaceFS {
         }
     }
 
+    // handle read of specific part number from a multipart upload
+    // see _handle_complete_multipart_size()
+    _handle_read_part_number(params, res, stat) {
```
romayalon, line 1117: Can you please add jsdoc on new functions? on _handle_complete_multipart_size() as well

src/sdk/namespace_fs.js
```diff
@@ -1111,6 +1112,27 @@ class NamespaceFS {
...
+            const multipart_size = Number(stat.xattr?.[XATTR_MULTIPART_SIZE]);
+            if (multipart_size > 0) {
+                params.start = multipart_size * (params.part_number - 1);
+                // last part may be smaller
+                params.end = Math.min(multipart_size * params.part_number, stat.size);
+                // TODO s3_get_object has a similar handling which better be written once
+                const content_range = `bytes ${params.start}-${params.end - 1}/${stat.size}`;
+                res.statusCode = 206;
+                dbg.log0(params.start, params.end, stat.size, multipart_size, content_range);
+                res.setHeader?.('Content-Range', content_range);
```
romayalon, line 1128: setting properties on res should not be on this file but on the s3 layer and probably should be written once (as written in the TODO :), check s3_get_object.js file)

src/sdk/namespace_fs.js
```diff
@@ -2320,6 +2369,8 @@ class NamespaceFS {
             delete_marker,
             storage_class,
             restore_status: GlacierBackend.get_restore_status(stat.xattr, new Date(), this._get_file_path({key})),
+            multipart_size,
```
romayalon, line 2372: why do we return multipart_size? where it's being used?

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -335,13 +336,29 @@ mocha.describe('namespace_fs', function() {
...
+            assert.strictEqual(Buffer.compare(part_data, data.subarray(0, part_size)), 0);
+
             const md = await ns_tmp.read_object_md({
                 bucket: upload_bkt,
                 key: mpu_key,
             }, dummy_object_sdk);
             console.log('read_object_md response', inspect(md));
             assert.deepStrictEqual(xattr, md.xattr);
 
+            const part_md = await ns_tmp.read_object_md({
```
romayalon, line 355: let's add more tests, these are some examples - 
1. 3 parts, all same size, get/head all and expect the same size but different data
2. 3 parts, last one is a different size, check the data, get/head all and expect different sizes, different data
3. 3 parts, each one different size, expect a failure

romayalon (changes requested):

src/sdk/namespace_fs.js
guymguym: perhaps add this to a new function in `http_utils.set_response_range()` which can use the existing `http_utils.format_http_ranges()` function

src/sdk/namespace_fs.js
guymguym: probably my idea was to use it in s3_utils.set_response_object_md() but eventually I didn't.

src/sdk/namespace_fs.js
```diff
@@ -968,7 +969,7 @@ class NamespaceFS {
             const stat = await file.stat(fs_context);
             this._throw_if_delete_marker(stat);
             // await this._fail_if_archived_or_sparse_file(fs_context, file_path, stat);
-
+            this._handle_read_part_number(params, res, stat);
```
guymguym, line 972: OK this should also be called in read_object_md to return the correct range for a part_number on HeadObject... I think I had it in another commit that was lost into a limbo between git branches...

See 
> [partNumber](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html#API_HeadObject_RequestSyntax)
Part number of the object being read. This is a positive integer between 1 and 10,000. Effectively performs a 'ranged' HEAD request for the part specified. Useful querying about the size of the part and the number of parts in this object.

guymguym (changes requested):

src/sdk/namespace_fs.js
```diff
@@ -1111,6 +1114,29 @@ class NamespaceFS {
         }
     }
 
+    /**
+     * handle read of specific part number from a multipart upload
+     * see _handle_complete_multipart_size()
+     * @param {object} params - upload object params
+     * @param {object} stat - file stats
```
romayalon, line 1121: nb.NativeFSStats

src/sdk/namespace_fs.js
```diff
@@ -968,7 +971,7 @@ class NamespaceFS {
             const stat = await file.stat(fs_context);
             this._throw_if_delete_marker(stat);
             // await this._fail_if_archived_or_sparse_file(fs_context, file_path, stat);
-
+            this._handle_read_part_number(params, stat);
```
romayalon, line 974: when file is a directory object the stat here is taken from the .folder file and not from the md path, therefore we won't have the xattr, we need to add another stat for having this info.

romayalon (approved):

PR #8221, opened 2024-07-22
Add get object owner for list object versions

vh05:
Making the changes to list object versions to return the result with the relevant object owner info from object info.


romayalon (approved):
LGTM
@vh05 Which other S3 operations should return object_owner?

PR #8224, opened 2024-07-22
NC | NSFS | S3 Flow `delete_bucket` When ULS Was Deleted + Refactor Function `translate_error_codes`

### LINKED ISSUE #8180: Unable to reuse buckets names from buckets previously deleted
### Environment info
- NooBaa Version: 5.15.4
- Platform: standalone

### Actual behavior
1. Create a bucket bucket1 with filesystempath /mnt/gpfs1/tes/test1
2. Delete the filesystempath /mnt/gpfs1/tes/test1
3. Now try to delete the bucket bucket1. It fails but bucket list displays it.

# noobaa-cli bucket delete bucket1 2>/dev/null
{
  "error": {
    "code": "NoSuchBucket",
    "message": "Bucket does not exist"
  }
}
# noobaa-cli bucket list 2>/dev/null
{
  "response": {
    "code": "BucketList",
    "reply": [
      {
        "name": "bucket1"
      }
    ]
  }
}

### Expected behavior
1. It should error out properly instead of saying bucket doesnot exist it should say bucket path missing

### Steps to reproduce
1.

shirady:
### Explain the changes
1. Wrap try-catch in the S3 flow `delete_bucket` in case the ULS (Underline Storage) was deleted (even though we are the ones that are supposed to delete it): for both bases - the ULS was created by NooBaa (S3 flow create_bucket) and when the ULS was created by the user (noobaa-cli flow bucket add).
2. Move the function `translate_error_codes` to `native_fs_utils` as mutual code for `accountspace_fs`, `bucketspace_fs`, and `namespace_fs`, the function was refactored to serve all the namespaces.
3. Changes in error codes entity:
- `_load_bucket` to bucket (from object)
- `set_bucket_versioning` to bucket (from object)
- `create_uls` and `delete_uls` to bucket (from object)
- The error `ENOTDIR` is now mapped to general `NO_SUCH_${entity}`, not just `NO_SUCH_OBJECT`.

### Issues: Fixed #8180 
1. The fix is only for S3 flow (in PR #8187 it was for the case where the bucket was created and deleted using the noobaa-cli).

#### Unit Tests:
Please run: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace_fs.js`

#### Manual Tests:
I saw 2 cases:
(1) The bucket was created using an S3 request and deleted using an S3 request.
(2) The bucket was created using the CLI and deleted using an S3 request.

(1) The bucket was created using an S3 request and deleted using an S3 request (based on the [comment](https://github.com/noobaa/noobaa-core/issues/8180#issuecomment-2230787483))
5. Start the NSFS server with: `sudo node src/cmd/nsfs --debug 5`
6. Create the alias for S3 service: `alias s3-nc-user-1='AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443'`.
7. Create the bucket using S3: `s3-nc-user-1-s3 s3 mb s3://bucket-1001-16-07`
8. Test that you are able to list the newly created bucket: `s3-nc-user-1-s3 s3 ls`
9. Remove the directory that was created for the bucket: `sudo rmdir /tmp/nsfs_root1/bucket-1001-16-07`
10. Try to delete the bucket: `s3-nc-user-1-s3 s3 rb s3://bucket-1001-16-07`, after this fix it should be: `remove_bucket: bucket-1001-16-07`, in the endpoint logs you can see:

> [nsfs/68124]  [WARN] core.sdk.bucketspace_fs:: delete_bucket: bucket name bucket-1001-19-07 with bucket_storage_path /tmp/nsfs_root1/bucket-1001-16-07 got an error while trying to delete_uls [Error: No such file or directory] { code: 'ENOENT', rpc_code: 'NO_SUCH_BUCKET' }

(2) The bucket was created using the CLI and deleted using an S3 request
3. Create the alias for S3 service: `alias s3-nc-user-1='AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443'`.
5.  Create the bucket using the CLI: first create the `path` for the bucket and give permissions: `sudo mkdir -p /tmp/nsfs_root1/bucket-1001-16-07` and `sudo chmod -R 777 /tmp/nsfs_root1/bucket-1001-16-07`
6. Test that you are able to list the newly created bucket: `s3-nc-user-1-s3 s3 ls`
7. Remove the directory that was created for the bucket: `sudo rmdir /tmp/nsfs_root1/bucket-1001-16-07`
8. Try to delete the bucket: `s3-nc-user-1-s3 s3 rb s3://bucket-1001-16-07`, after this fix it should be: `remove_bucket: bucket-1001-16-07`, in the endpoint logs you can see:

> Jul-23 9:33:22.258 [nsfs/51199]  [WARN] core.sdk.namespace_fs:: _load_bucket failed, on bucket_path /tmp/nsfs_root1/bucket-1001-16-07 got error [Error: No such file or directory] { code: 'ENOENT' }
> Jul-23 9:33:22.258 [nsfs/51199]  [WARN] core.sdk.bucketspace_fs:: delete_bucket: bucket name bucket-1001-16-07 got an error while trying to list_objects [Error: No such file or directory] { code: 'ENOENT', rpc_code: 'NO_SUCH_BUCKET' }

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -21,7 +21,8 @@ const { CONFIG_SUBDIRS } = require('../manage_nsfs/manage_nsfs_constants');
 
 const KeysSemaphore = require('../util/keys_semaphore');
 const { get_umasked_mode, isDirectory, validate_bucket_creation,
-    create_config_file, delete_config_file, get_bucket_tmpdir_full_path, folder_delete } = require('../util/native_fs_utils');
+    create_config_file, delete_config_file, get_bucket_tmpdir_full_path, folder_delete,
+    entity_enum, translate_error_codes } = require('../util/native_fs_utils');
 const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 const { anonymous_access_key } = require('./object_sdk');
 
@@ -84,16 +85,6 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
         } catch (err) {
             return false;
         }
-    }
-
-    _translate_bucket_error_codes(err) {
-        if (err.rpc_code) return err;
-        if (err.code === 'ENOENT') err.rpc_code = 'NO_SUCH_BUCKET';
-        if (err.code === 'EEXIST') err.rpc_code = 'BUCKET_ALREADY_EXISTS';
-        if (err.code === 'EPERM' || err.code === 'EACCES') err.rpc_code = 'UNAUTHORIZED';
-        if (err.code === 'IO_STREAM_ITEM_TIMEOUT') err.rpc_code = 'IO_STREAM_ITEM_TIMEOUT';
-        if (err.code === 'INTERNAL_ERROR') err.rpc_code = 'INTERNAL_ERROR';
-        return err;
     }
 
     async read_account_by_access_key({ access_key }) {
@@ -178,7 +169,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             }
             return bucket;
         } catch (err) {
-            const rpc_error = this._translate_bucket_error_codes(err);
+            const rpc_error = translate_error_codes(err, entity_enum.BUCKET);
             if (err.rpc_code === 'INVALID_SCHEMA') err.rpc_code = 'INVALID_BUCKET_STATE';
             new NoobaaEvent(NoobaaEvent[rpc_error.rpc_code]).create_event(name, {bucket_name: name}, err);
             throw rpc_error;
@@ -222,7 +213,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
                 dbg.error('BucketSpaceFS: root dir not found', err, this.bucket_schema_dir);
                 throw new S3Error(S3Error.NoSuchBucket);
             }
-            throw this._translate_bucket_error_codes(err);
+            throw translate_error_codes(err, entity_enum.BUCKET);
         }
 
         const account = object_sdk.requesting_account;
@@ -302,7 +293,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
                 await create_config_file(this.fs_context, this.bucket_schema_dir, bucket_config_path, bucket_config);
             } catch (err) {
                 new NoobaaEvent(NoobaaEvent.BUCKET_CREATION_FAILED).create_event(name, {bucket_name: name}, err);
-                throw this._translate_bucket_error_codes(err);
+                throw translate_error_codes(err, entity_enum.BUCKET);
             }
 
             // create bucket's underlying storage directory
@@ -313,7 +304,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
                 dbg.error('BucketSpaceFS: create_bucket could not create underlying directory - nsfs, deleting bucket', err);
                 new NoobaaEvent(NoobaaEvent.BUCKET_DIR_CREATION_FAILED).create_event(name, {bucket: name, path: bucket_storage_path}, err);
                 await nb_native().fs.unlink(this.fs_context, bucket_config_path);
-                throw this._translate_bucket_error_codes(err);
+                throw translate_error_codes(err, entity_enum.BUCKET);
             }
         });
     }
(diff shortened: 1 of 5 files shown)


src/sdk/namespace_fs.js
```diff
@@ -2386,7 +2388,10 @@ class NamespaceFS {
         try {
             await nb_native().fs.stat(fs_context, this.bucket_path);
         } catch (err) {
-            throw this._translate_object_error_codes(err);
+            const err_to_throw = this._translate_object_error_codes(err);
+            // this a temp solution so on the function related to bucket we will change the error to the relevant one (bucket instead of object)
+            if (err_to_throw && err_to_throw.rpc_code === 'NO_SUCH_OBJECT') err_to_throw.rpc_code = 'NO_SUCH_BUCKET';
```
naveenpaul1, line 2393: @shirady Will it fail for object? for object error code will be `NO_SUCH_BUCKET`?

shirady: As I understand it we are in a function that it's name is `_load_bucket`, when it received the `ENOENT` error code it used `_translate_object_error_codes` and it was `NO_SUCH_OBJECT` and I don't think it supposed to be the rpc_code, because the action it on a bucket and not object.

> @shirady Will it fail for object? for object error code will be `NO_SUCH_BUCKET`?

I'm not sure I understand the question, I can take an example maybe - if someone tries to delete an object and the path of the bucket was deleted the flow should be: delete_object -> _load_bucket (failed, cannot stat the path) -> throw err with rpc_code 'NO_SUCH_BUCKET' -> won't be translated again in `_translate_object_error_codes`. I hope this answers your question.

naveenpaul1: @shirady my doubt is,  When try to list the object which dosn't exists inside bucket, will this code return `NO_SUCH_BUCKET`?

src/sdk/namespace_fs.js
```diff
@@ -2372,6 +2372,8 @@ class NamespaceFS {
         }
     }
 
+    // TODO - improve this function so it will be mutual to namespace_fs and bucketspace_fs
+    // (like in accountspace_fs added argument for entity)
```
romayalon, line 2376: @shirady sounds like this function exists in namespace_fs, bucketspace_fs, and accountspace_fs, can you move accountspace_fs._translate_error_codes() to a native_fs_utils.js and call it with entity from bucketspace_fs and namespace_fs?

src/sdk/namespace_fs.js
```diff
@@ -2386,7 +2388,11 @@ class NamespaceFS {
         try {
             await nb_native().fs.stat(fs_context, this.bucket_path);
         } catch (err) {
-            throw this._translate_object_error_codes(err);
+            dbg.warn('_load_bucket failed, on bucket_path', this.bucket_path, 'got error', err);
+            const err_to_throw = this._translate_object_error_codes(err);
```
romayalon, line 2392: With the change of the mutual function, you will just need to call it with 'BUCKET' entity instead of doing this manual change.

src/sdk/bucketspace_fs.js
```diff
@@ -353,13 +353,27 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+                        // in case the ULS was deleted - we will continue
+                        if (err.rpc_code !== 'NO_SUCH_BUCKET') throw err;
+                    }
                 } else {
                     // 2. delete only bucket tmpdir
-                    const list = await ns.list_objects({ ...params, bucket: name, limit: 1 }, object_sdk);
+                    let list;
+                    try {
+                        list = await ns.list_objects({ ...params, bucket: name, limit: 1 }, object_sdk);
+                    } catch (err) {
```
romayalon, line 372: I suggest moving the check to the outer catch instead of having it twice here

src/sdk/namespace_fs.js
shirady: I'm not sure I understand.
I tested listing a non-existing object in AWS and it doesn't throw an error (and we also don't throw an error).
I used this command: `aws s3 ls s3://<bucket-name>/<non-existing-object>`.

src/sdk/namespace_fs.js
shirady: I thought doing it as a part of fixing the issue would be hard for backport, hence I added this comment and hoped to do it in the future, but since you suggested I changed the code to what I planned in this PR.

src/sdk/namespace_fs.js
shirady: I changed it, thanks.

src/sdk/bucketspace_fs.js
shirady: If I would remove this internal try-catch I would skip the next step which is removing the config file, and I don't want that to happen.

src/sdk/namespace_fs.js
shirady: BTW, what is the difference between the utils in `native_fs_utils` and `fs_utils`?

romayalon (approved):

src/util/native_fs_utils.js
```diff
@@ -580,6 +588,25 @@ function get_bucket_tmpdir_full_path(bucket_path, bucket_id) {
...
+
+/**
+ * translate_error_codes we translate FS error codes to rpc_codes (strings)
+ * and add the rpc_code property to the original error object
+ * @param {object} err
+ * @param {('OBJECT'|'BUCKET'|'USER'|'ACCESS_KEY')} entity
+ */
+function translate_error_codes(err, entity) {
+    if (err.rpc_code) return err;
+    if (err.code === 'ENOENT' || err.code === 'ENOTDIR') {
```
shirady, line 600: @romayalon I just want to make sure that you noticed that the error code `ENOTDIR` will be mapped to `NO_SUCH_${entity}`, and not only 'NO_SUCH_OBJECT' (as was added in PR #7995)

PR #8234, opened 2024-07-26
[NC? | NSFS] Add support for bucket tagging in NSFS

tangledbytes:
### Explain the changes
This PR attempts to add support for bucket tagging in NSFS.

The tags are simply stored in the config file for now. `xattrs` were also considered for the same purpose but decided against it at least in this iteration as some file systems (eg. AWS EFS) do not support extended attributes.

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -22,7 +22,7 @@ const { CONFIG_SUBDIRS } = require('../manage_nsfs/manage_nsfs_constants');
 const KeysSemaphore = require('../util/keys_semaphore');
 const { get_umasked_mode, isDirectory, validate_bucket_creation,
     create_config_file, delete_config_file, get_bucket_tmpdir_full_path, folder_delete,
-    entity_enum, translate_error_codes, update_config_file} = require('../util/native_fs_utils');
+    entity_enum, translate_error_codes, update_config_file } = require('../util/native_fs_utils');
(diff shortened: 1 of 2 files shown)


guymguym (approved):
@tangledbytes is there any test coverage for this that we can enable in CI?

src/sdk/bucketspace_fs.js
```diff
@@ -440,15 +440,56 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
     }
 
     async get_bucket_tagging(params) {
-        // TODO
+        try {
+            const { name } = params;
+            dbg.log0('BucketSpaceFS.get_bucket_tagging: Bucket name', name);
+            const bucket_config_path = this._get_bucket_config_path(name);
+            const { data } = await nb_native().fs.readFile(this.fs_context, bucket_config_path);
+            const bucket = JSON.parse(data.toString());
```
guymguym, line 488: would be nice to move this common code that repeats throughout this file to a method like `const bucket = await this._load_bucket(name);`. maybe in a separate pr? CC @romayalon

guymguym (approved):

tangledbytes:
I have added tests here. Merging this PR for now.

src/sdk/bucketspace_fs.js
romayalon: it was fixes as part of the configFS changes, thanks

PR #8236, opened 2024-07-29
NC | NSFS | CLI | Make Update Account Tolerance to Missing Master Key

### LINKED ISSUE #8104: NC | MAke list accounts and update account tolerant to missing master key
### Environment info
- NooBaa Version: **5.15.3**
- Platform: **RHEL**

### Actual behavior
1. Currently, if an account master_key_id points to a missing master key, the list and update account commands in noobaa-cli (manage_nsfs) will fail with an error.
2. This can happen when the master keys that were previously used cannot be recovered due to some issue.

### Expected behavior
1. We should allow to list accounts and show that those access keys cannot be used.
2. We should allow to update an account to new access key using the current master key.

### Steps to reproduce
1. Delete the master keys file when there are already accounts existing that use it.

CC @romayalon @ramya-c3 @rkomandu

shirady:
### Explain the changes
1. Allow account update (only `regenerate` and `access_key` and `secret_key`) when we cannot decrypt the current access keys.
2. Edit the bucket list tests to validate that we receive an array in the response (as we did in PR #8272).
3. In `config_fs` fix JSDoc of the function `get_identity_config_data`.
4. In `mamage_nsfs` add JSDoc to the function `fetch_existing_account_data`.

### Issues: Fixed partially #8104 
1. Currently, if an account's `master_key_id` points to a missing master key, the commands: `account update` and `account status`  in noobaa-cli (manage_nsfs) will fail with an error.
Note: A missing master key can happen when the master keys that were previously used cannot be recovered due to some issue.
2. GAP: in this PR we didn't make list accounts tolerance to missing master key and list buckets/accounts with Invalid JSON (it was in past of this PR but due to code changes that happened this code fix will be added separately).

#### Unit Tests
Please run: 
1. `sudo npx jest test_nc_nsfs_bucket_cli.test.js` (for the change in point number 2 in the changes above).
2. `sudo npx jest test_nc_account_invalid_mkm_integration.test.js`

#### Manual Testing
A. Missing master key
1. Create an account with the CLI: `sudo node src/cmd/manage_nsfs account add --name shira-1001 --new_buckets_path /tmp/nsfs_root1 --access_key <access-key> --secret_key <secret-key> --uid <uid> --gid <gid>`
2. Make the master key invalid by renaming the file `master_key.json`: `sudo mv /etc/noobaa.conf.d/master_keys.json /etc/noobaa.conf.d/temp_master_keys.json`
3. Account list with decryption: `sudo node src/cmd/manage_nsfs account list --wide --show_secrets` (doesn't fail, but you will see the property `encrypted_secret_key`).
7. Account status with decryption: `sudo node src/cmd/manage_nsfs account status --name shira-1001 --show_secrets` (will fail).
8. Account update - can update the access keys (other update options will result in an error): `sudo node src/cmd/manage_nsfs account update --name shira-1001 --regenerate` or `sudo node src/cmd/manage_nsfs account update --access_key <access-key-id> --secret_key <secret-key>`.

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/sdk/config_fs.js
+++ b/src/sdk/config_fs.js
@@ -187,26 +187,30 @@ class ConfigFS {
     }
 
     /**
-     * get_config_data reads a config file and returns its content 
+     * get_identity_config_data reads a config file and returns its content 
      * while omitting secrets if show_secrets flag was not provided
      * and decrypts the account's secret_key if decrypt_secret_key is true
      * if silent_if_missing is true -      
      *   if the config file was deleted (encounter ENOENT error) - continue (returns undefined)
      * @param {string} config_file_path
-     * @param {{show_secrets?: boolean, decrypt_secret_key?: boolean, silent_if_missing?: boolean}} [options]
+     * @param {{show_secrets?: boolean, decrypt_secret_key?: boolean, silent_if_missing?: boolean,
+     *          return_encrypted_if_decryption_fails?:boolean}} [options]
      * @returns {Promise<Object>}
      */
     async get_identity_config_data(config_file_path, options = {}) {
-        const { show_secrets = false, decrypt_secret_key = false, silent_if_missing = false } = options;
+        const { show_secrets = false, decrypt_secret_key = false, silent_if_missing = false,
+            return_encrypted_if_decryption_fails = false } = options;
+         let config_data;
         try {
             const data = await this.get_config_data(config_file_path, options);
             if (!data && silent_if_missing) return;
-            const config_data = _.omit(data, show_secrets ? [] : ['access_keys']);
+            config_data = _.omit(data, show_secrets ? [] : ['access_keys']);
             if (decrypt_secret_key) config_data.access_keys = await nc_mkm.decrypt_access_keys(config_data);
             return config_data;
         } catch (err) {
             dbg.warn('get_identity_config_data: with config_file_path', config_file_path, 'got an error', err);
             if (err.code === 'ENOENT' && silent_if_missing) return;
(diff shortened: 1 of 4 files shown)


shirady:
@romayalon I added the suggested changes.
I was wondering if you also want to add it in the docs somewhere (the file `NooBaaCLI.md` is very general).

shirady:
I Converted it to draft since there was an issue after rebase with the health tests.

src/sdk/config_fs.js
```diff
@@ -187,26 +187,30 @@ class ConfigFS {
...
             const data = await this.get_config_data(config_file_path, options);
             if (!data && silent_if_missing) return;
-            const config_data = _.omit(data, show_secrets ? [] : ['access_keys']);
+            config_data = _.omit(data, show_secrets ? [] : ['access_keys']);
             if (decrypt_secret_key) config_data.access_keys = await nc_mkm.decrypt_access_keys(config_data);
             return config_data;
         } catch (err) {
             dbg.warn('get_identity_config_data: with config_file_path', config_file_path, 'got an error', err);
             if (err.code === 'ENOENT' && silent_if_missing) return;
+            if (return_encrypted_if_decryption_fails && err.rpc_code === 'INVALID_MASTER_KEY') return config_data;
```
romayalon, line 213: As we discussed,
This solution creates a new master key under the hood, which means it tries to solve a big issue (master key disappeared) without letting the customer know about it. I don't think this is how we should do that, let's take some time for design. 
I'm worried about temporary failures that we will create a new master key instead of fixing the issue.

shirady:
Closing for now - would need design (see [this comment](https://github.com/noobaa/noobaa-core/pull/8236#discussion_r1798901662)).

PR #8242, opened 2024-07-30
NC | NSFS | Remove Extra Information From The Logs + Refactor `iam _constants`

shirady:
### Explain the changes
1. Remove extra information from the logs:
 - The `account_sdk` was printed at the beginning of an action, in case we need something, we will add only what we need from it).
 - We used to print the account with all its details, to avoid any printing of plain secret_key, only print the ID and the name - we can check in the config the other details if needed.
3. Move the constants that we had in iam_utils to iam_constants:
- also move the constants that we had for parameter names to an object named `IAM_PARAMETER_NAME`.
4. Fix the year to 2024 in files related to IAM.
5. Remove extra newline between functions in accountspace_fs.

1. Currently, we printed (in log1 level) the `account_sdk` which filled the logs with a lot of unneeded pritings.

#### Unit Tests:
Please run:
1. `npx jest test_iam_utils.test.js`
2. `sudo npx jest test_accountspace_fs.test.js`

#### Manual Tests:
- Before starting the server please add this line: `process.env.NOOBAA_LOG_LEVEL = 'nsfs';` in the endpoint.js (before the condition `if (process.env.NOOBAA_LOG_LEVEL) {`)
- I Change the `config.NSFS_CHECK_BUCKET_BOUNDARIES = false; //SDSD` because I'm using the `/tmp/` and not `/private/tmp/`.
3. Create the alias for IAM service: `alias nc-user-1-iam='AWS_ACCESS_KEY_ID=<acess-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:7005'`.
4. Use AWS CLI to send requests to the IAM service (and look at the logs), for example:
 `s3-nc-user-1-iam iam create-user --user-name Bob`
 `s3-nc-user-1-iam iam get-user --user-name Bob`
 `s3-nc-user-1-iam iam update-user --user-name Bob new-path '/division_abc/subdivision_abc/'`
 `s3-nc-user-1-iam iam delete-user --user-name Bob`
 `s3-nc-user-1-iam iam list-users`
 `s3-nc-user-1-iam iam create-user --user-name Bob'`
 `s3-nc-user-1-iam iam create-access-key --user-name Bob`
 `s3-nc-user-1-iam iam get-access-key-last-used --access-key-id <access-key>`
 `s3-nc-user-1-iam iam update-access-key --access-key-id <access-key> --user-name Bob --status Inactive`
 `s3-nc-user-1-iam iam delete-access-key --access-key-id <access-key> --user-name Bob`
 `s3-nc-user-1-iam iam list-access-keys --user-name Bob`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/iam/iam_constants.js
+++ b/src/endpoint/iam/iam_constants.js
@@ -1,9 +1,6 @@
 /* Copyright (C) 2024 NooBaa */
 'use strict';
-// TODO - move more constants here (mainly from iam_utils)
-const DEFAULT_MAX_ITEMS = 100;
 
-// TODO - reuse
 // function name in snake case style
 const IAM_ACTIONS = Object.freeze({
     CREATE_USER: 'create_user',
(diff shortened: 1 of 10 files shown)


src/endpoint/iam/iam_constants.js
```diff
@@ -18,20 +15,58 @@ const IAM_ACTIONS = Object.freeze({
...
+    'list_access_keys': 'ListAccessKeys',
+});
 
-// parameter names
-const IAM_PATH = 'Path';
-const NEW_IAM_PATH = 'NewPath';
-const IAM_PATH_PREFIX = 'PathPrefix';
-const USERNAME = 'UserName';
-const NEW_USERNAME = 'NewUserName';
+const access_key_status_enum = Object.freeze({
```
guymguym, line 34: minor: I would make it upper case. these are just nested constants, but still constants.

guymguym (approved):

src/endpoint/iam/iam_constants.js
shirady: I can change it, I was inspired by this enum:
https://github.com/noobaa/noobaa-core/blob/909e29f92ad0d65fe65a9176058d62444b843fb8/src/sdk/namespace_fs.js#L71-L75

guymguym: yeah but i think we should be consistent to define constants with uppercase and Object.freeze() - this is a (semi)standard across many languages, and helps readability.

PR #8252, opened 2024-08-05
Noobaa Account: Replace bcrypt password hashing by crypto

aspandey:
As bcrypt is not under active maintenance, we need to replace it with node.js crypto hashing module.

This PR is in line with https://github.com/noobaa/noobaa-core/pull/8213

IN this PR I tried to implement in-built node.js crypto module. Over all approach is same. 
We just need to implement two addition functions to create hash and verify that hash.

### DIFF
--- a/src/server/common_services/auth_server.js
+++ b/src/server/common_services/auth_server.js
@@ -3,6 +3,7 @@
 
 const _ = require('lodash');
 const bcrypt = require('bcrypt');
+const node_hash = require('../../util/account_password_hash');
 const ip_module = require('ip');
 
 const P = require('../../util/promise');
@@ -20,6 +21,14 @@ const jwt_utils = require('../../util/jwt_utils');
 const config = require('../../../config');
 const s3_bucket_policy_utils = require('../../endpoint/s3/s3_bucket_policy_utils');
 
+
+
+function compare_password_hash(password, target_account) {
+    if (bcrypt.compare(password.unwrap(), target_account.password.unwrap())) {
+        return true;
+    }
+    return (node_hash.verify_node_password_hash(password.unwrap(), target_account.password.unwrap()));
+}
 
 /**
  *
@@ -66,7 +75,7 @@ function create_auth(req) {
             if (!password) return;
 
             return P.resolve()
-                .then(() => bcrypt.compare(password.unwrap(), target_account.password.unwrap()))
+                .then(() => compare_password_hash(password, target_account))
                 .then(match => {
                     if (!match) {
                         dbg.log0('password mismatch', email, system_name);
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -6,7 +6,8 @@ const _ = require('lodash');
 const net = require('net');
 const chance = require('chance')();
 const GoogleStorage = require('../../util/google_storage_wrap');
-const bcrypt = require('bcrypt');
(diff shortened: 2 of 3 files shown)


src/server/common_services/auth_server.js
```diff
@@ -21,6 +22,14 @@ const config = require('../../../config');
 const s3_bucket_policy_utils = require('../../endpoint/s3/s3_bucket_policy_utils');
 
 
+
+function compare_password_hash(password, target_account) {
+    if (bcrypt.compare(password.unwrap(), target_account.password.unwrap())) {
+        return true;
+    }
```
guymguym, line 29: What is this supposed to do - are we still allowing previous bcyrpt?
This needs to be at least explained in a code comment.

src/server/system_services/account_server.js
```diff
@@ -6,7 +6,8 @@ const _ = require('lodash');
 const net = require('net');
 const chance = require('chance')();
 const GoogleStorage = require('../../util/google_storage_wrap');
-const bcrypt = require('bcrypt');
+const node_hash = require('../../util/account_password_hash');
```
guymguym, line 9: I would call this new module `password_utils`
```suggestion
const password_utils = require('../../util/password_utils');
```

src/server/system_services/account_server.js
```diff
@@ -1358,9 +1359,9 @@ function ensure_support_account() {
         });
 }
 
-function bcrypt_password(password) {
-    return P.resolve()
-        .then(() => password && bcrypt.hash(password, 10));
+function crypto_password(password) {
+        return P.resolve()
+        .then(() => password && node_hash.create_node_password_hash(password));
 }
```
guymguym, line 1365: Better get rid of this P.resolve and use proper async function

src/server/common_services/auth_server.js
```diff
@@ -21,6 +22,14 @@ const config = require('../../../config');
 const s3_bucket_policy_utils = require('../../endpoint/s3/s3_bucket_policy_utils');
 
 
+
+function compare_password_hash(password, target_account) {
+    if (bcrypt.compare(password.unwrap(), target_account.password.unwrap())) {
```
guymguym, line 27: is bcrypt.compare a sync function call or async?

guymguym (changes requested):

src/util/account_password_hash.js
```diff
@@ -0,0 +1,38 @@
...
+                return err;
+            }
+            resolve(key === derivedKey.toString('hex'));
+        });
+    });
+}
+
+
+exports.create_node_password_hash = create_node_password_hash;
+exports.verify_node_password_hash = verify_node_password_hash;
```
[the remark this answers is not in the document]
guymguym: I am not so sure about the choice of numbers - iteration count, salt size, and key size...

I was looking at the nodejs docs - https://nodejs.org/api/crypto.html#cryptopbkdf2password-salt-iterations-keylen-digest-callback

> The iterations argument must be a number set as high as possible. The higher the number of iterations, the more secure the derived key will be, but will take a longer amount of time to complete.

> The salt should be as unique as possible. It is recommended that a salt is random and at least 16 bytes long. See [NIST SP 800-132](https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-132.pdf) for details.

PR #8253, opened 2024-08-05
[NC | NSFS | Glacier] handle `ENOENT` in `_finalize_restore` and fix glacier tests not running

### LINKED ISSUE #8251: NSFS | NC | GLACIER restore flow needs to handle `ENOENT`
### Environment info
- NooBaa Version: **master**
- Platform: **DBS3**

### Actual behavior
If `finalize_restore` fails because of `ENOENT` due to any reason including an object getting deleted after a restore request is submitted. This will force NooBaa to retry the restore over and over again.

### Expected behavior
NooBaa must handle `ENOENT` cases in it's restore flow.

tangledbytes:
### Explain the changes
This PR simply adds the handler in `_finalize_restore` to make sure that it doesn't errors out if encounters `ENOENT`.

Fixes #8251

### Testing Instructions:

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/nsfs_glacier_backend/tapecloud.js
+++ b/src/sdk/nsfs_glacier_backend/tapecloud.js
@@ -359,7 +359,17 @@ class TapeCloudGlacierBackend extends GlacierBackend {
         const entry = new NewlineReaderEntry(fs_context, entry_path);
         let fh = null;
         try {
-            fh = await entry.open("r");
+            try {
+                fh = await entry.open("r");
+            } catch (error) {
+                // restore does check if a file exist or not before triggering eeadm call but the file could get deleted
+                // after `restore` performs that check so check it here once again
+                if (error.code === 'ENOENT') {
+                    dbg.warn(`TapeCloudGlacierBackend._finalize_restore: log entry unexpectedly not found: ${entry.path}`);
+                    return;
+                }
+                throw error;
+            }
 
             const stat = await fh.stat(fs_context, {
                 xattr_get_keys: [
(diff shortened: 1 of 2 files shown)


src/sdk/nsfs_glacier_backend/tapecloud.js
```diff
@@ -396,6 +396,13 @@ class TapeCloudGlacierBackend extends GlacierBackend {
 
             await fh.replacexattr(fs_context, undefined, GlacierBackend.XATTR_RESTORE_REQUEST);
         } catch (error) {
+            // restore does check if a file exist or not before triggering eeadm call but the file could get deleted
+            // after `restore` performs that check so check it here once again
+            if (error.code === 'ENOENT') {
+                dbg.warn(`TapeCloudGlacierBackend._finalize_restore: log entry unexpectedly not found: ${entry.path}`);
+                return;
+            }
```
guymguym, line 404: it would be better if this was in a separate try-catch which is only on the open() call, to make sure we don't add any other fs calls in the flow that would be interpreted as missing file accidentally

guymguym (changes requested):

src/sdk/nsfs_glacier_backend/tapecloud.js
tangledbytes: Updated the PR

tangledbytes:
@guymguym, I have added 2 commits in this PR. One PR is the actual fix while the other one fixes the glacier tests not running (my guess since about 3 weeks now). I have split the commits for the ease of backport at least as one would be picked up by DBS3.

guymguym (approved):

PR #8254, opened 2024-08-05
Add `Date.now()` to `expiry` when passing it to `Expiration`

Neon-White:
### Explain the changes
1. The `Expiration` field in the response to `AssumeRole` returned the value in milliseconds; it should be in seconds, and added to the current time; this is now fixed
2. `generate_session_token()` passed `expiry` in milliseconds when `make_auth_token` expects seconds, this is now fixed
3. It is now possible to provide an expiration in seconds by providing the AWS CLI flag `--duration-seconds`

1. Fixes [BZ#2299801](https://bugzilla.redhat.com/show_bug.cgi?id=2299801)

### Testing Instructions:
1. Follow the `STS test` section [here](https://github.com/noobaa/noobaa-core/pull/8119#issuecomment-2180089580), observe the return value of `Expiration:`
2. Try to utilize the credentials and observe whether they expire in the timestamp shown in the previous step

- [ ] Doc added/updated
- [x] Tests updated

### DIFF
--- a/src/endpoint/sts/ops/sts_post_assume_role.js
+++ b/src/endpoint/sts/ops/sts_post_assume_role.js
@@ -6,12 +6,16 @@ const { StsError } = require('../sts_errors');
 const jwt_utils = require('../../../util/jwt_utils');
 const config = require('../../../../config');
 const { CONTENT_TYPE_APP_FORM_URLENCODED } = require('../../../util/http_utils');
+const s3_utils = require('../../s3/s3_utils');
 
 /**
  * https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html
  */
 async function assume_role(req) {
     dbg.log1('sts_post_assume_role body: ', req.body);
+    const duration_ms = _parse_sts_duration(req.body.duration_seconds);
+    const duration_sec = Math.ceil(duration_ms / 1000);
+    const expiration_time = Date.now() + duration_ms;
     let assumed_role;
     try {
         assumed_role = await req.sts_sdk.get_assumed_role(req);
@@ -21,10 +25,11 @@ async function assume_role(req) {
         }
         throw new StsError(StsError.InternalFailure);
     }
+
     // Temporary credentials are NOT stored in noobaa
     // The generated session token will store in it the temporary credentials and expiry and the role's access key
     const access_keys = await req.sts_sdk.generate_temp_access_keys();
-    const expiry = config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS;
+
     return {
         AssumeRoleResponse: {
             AssumeRoleResult: {
@@ -35,12 +40,12 @@ async function assume_role(req) {
                 Credentials: {
                     AccessKeyId: access_keys.access_key.unwrap(),
                     SecretAccessKey: access_keys.secret_key.unwrap(),
-                    Expiration: expiry,
+                    Expiration: s3_utils.format_s3_xml_date(expiration_time),
                     SessionToken: generate_session_token({
                         access_key: access_keys.access_key.unwrap(),
                         secret_key: access_keys.secret_key.unwrap(),
                         assumed_role_access_key: assumed_role.access_key
-                    }, expiry)
+                    }, duration_sec)
                 },
                 PackedPolicySize: 0
             }
@@ -49,9 +54,48 @@ async function assume_role(req) {
 }
 
 // create and return the signed token
+/**
+ * @param {Object} auth_options
+ * @param {Number} expiry in seconds
+ * @returns {String}
+*/
 function generate_session_token(auth_options, expiry) {
     dbg.log1('sts_post_assume_role.make_session_token: ', auth_options, expiry);
     return jwt_utils.make_auth_token(auth_options, { expiresIn: expiry });
+}
+
+// TODO: Generalize and move to a utils file in the future
+/**
+ * @param {String|undefined} duration_input duration in seconds
+ * @returns {Number} duration in milliseconds
+ */
+function _parse_sts_duration(duration_input) {
+    if (duration_input === undefined) {
+        return config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS;
+    }
+
+    const duration_sec = Number(duration_input);
+
+    if (!Number.isInteger(duration_sec)) {
+        throw new StsError(StsError.InvalidParameterValue);
+    }
+
+    if (duration_sec < config.STS_MIN_DURATION_SECONDS) {
+        throw new StsError(_sts_duration_validation_error(duration_input, 'greater', config.STS_MIN_DURATION_SECONDS));
+    }
+    if (duration_sec > config.STS_MAX_DURATION_SECONDS) {
+        throw new StsError(_sts_duration_validation_error(duration_input, 'less', config.STS_MAX_DURATION_SECONDS));
+    }
+
+    const duration_ms = duration_sec * 1000;
+    return duration_ms;
+}
+
+function _sts_duration_validation_error(duration_input, constraint, constraint_value) {
+    return {
+        ...StsError.ValidationError,
+        message: `Value ${duration_input} for durationSeconds failed to satisfy constraint: Member must have value ${constraint} than or equal to ${constraint_value}`,
+    };
 }
 
 module.exports = {
--- a/src/test/unit_tests/test_sts.js
+++ b/src/test/unit_tests/test_sts.js
@@ -16,6 +16,8 @@ const jwt_utils = require('../../util/jwt_utils');
 const config = require('../../../config');
 const { S3Error } = require('../../endpoint/s3/s3_errors');
 
+const defualt_expiry_seconds = Math.ceil(config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS / 1000);
+
 const errors = {
     expired_token_s3: {
         code: S3Error.ExpiredToken.code,
@@ -52,6 +54,10 @@ const errors = {
     invalid_action: {
         code: stsErr.InvalidAction.code,
         message: stsErr.InvalidAction.message
+    },
+    validation_error: {
+        code: stsErr.ValidationError.code,
+        message: stsErr.ValidationError.message
     },
     invalid_schema_params: {
         code: 'INVALID_SCHEMA_PARAMS',
@@ -177,7 +183,7 @@ mocha.describe('STS tests', function() {
         };
         const json = await assume_role_and_parse_xml(sts_admin, params);
         validate_assume_role_response(json, `arn:aws:sts::${user_b_key}:assumed-role/${role_b}/${params.RoleSessionName}`,
-            `${user_b_key}:${params.RoleSessionName}`, user_b_key);
+            `${user_b_key}:${params.RoleSessionName}`, user_b_key, defualt_expiry_seconds);
     });
 
     mocha.it('admin assume non existing role of user b - should be rejected', async function() {
@@ -208,10 +214,10 @@ mocha.describe('STS tests', function() {
         };
         const json = await assume_role_and_parse_xml(sts_c, params);
         validate_assume_role_response(json, `arn:aws:sts::${user_b_key}:assumed-role/${role_b}/${params.RoleSessionName}`,
-            `${user_b_key}:${params.RoleSessionName}`, user_b_key);
+            `${user_b_key}:${params.RoleSessionName}`, user_b_key, defualt_expiry_seconds);
 
         const temp_creds = validate_assume_role_response(json, `arn:aws:sts::${user_b_key}:assumed-role/${role_b}/${params.RoleSessionName}`,
-            `${user_b_key}:${params.RoleSessionName}`, user_b_key);
+            `${user_b_key}:${params.RoleSessionName}`, user_b_key, defualt_expiry_seconds);
         const s3 = new AWS.S3({
             ...sts_creds,
             accessKeyId: temp_creds.access_key,
@@ -255,7 +261,7 @@ mocha.describe('STS tests', function() {
         };
         const json = await assume_role_and_parse_xml(sts, params);
         validate_assume_role_response(json, `arn:aws:sts::${user_b_key}:assumed-role/${role_b}/${params.RoleSessionName}`,
-            `${user_b_key}:${params.RoleSessionName}`, user_b_key);
+            `${user_b_key}:${params.RoleSessionName}`, user_b_key, defualt_expiry_seconds);
     });
 
     mocha.it('update assume role policy of user b to allow user a', async function() {
@@ -296,7 +302,7 @@ mocha.describe('STS tests', function() {
         };
         const json = await assume_role_and_parse_xml(sts_c, params);
         validate_assume_role_response(json, `arn:aws:sts::${user_b_key}:assumed-role/${role_b}/${params.RoleSessionName}`,
-            `${user_b_key}:${params.RoleSessionName}`, user_b_key);
+            `${user_b_key}:${params.RoleSessionName}`, user_b_key, defualt_expiry_seconds);
     });
 
     mocha.it('update assume role policy of user b to allow user a sts:*', async function() {
@@ -337,7 +343,7 @@ mocha.describe('STS tests', function() {
         };
         const json = await assume_role_and_parse_xml(sts_c, params);
         validate_assume_role_response(json, `arn:aws:sts::${user_b_key}:assumed-role/${role_b}/${params.RoleSessionName}`,
-            `${user_b_key}:${params.RoleSessionName}`, user_b_key);
+            `${user_b_key}:${params.RoleSessionName}`, user_b_key, defualt_expiry_seconds);
     });
 
     mocha.it('update assume role policy of user b to allow user a *', async function() {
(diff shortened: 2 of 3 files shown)


src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -35,7 +35,7 @@ async function assume_role(req) {
                 Credentials: {
                     AccessKeyId: access_keys.access_key.unwrap(),
                     SecretAccessKey: access_keys.secret_key.unwrap(),
-                    Expiration: expiry,
+                    Expiration: `in ${expiry.toString()}ms`,
```
guymguym, line 38: what is this expiration format? from docs I see it should be specific time, and not this duration.

guymguym (changes requested):

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -35,7 +35,7 @@ async function assume_role(req) {
                 Credentials: {
                     AccessKeyId: access_keys.access_key.unwrap(),
                     SecretAccessKey: access_keys.secret_key.unwrap(),
-                    Expiration: expiry,
+                    Expiration: Date.now() + expiry,
```
shirady, line 38: I understand the fix - to have a timestamp as in the API  [AssumeRole](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) (in the type [Credentials](https://docs.aws.amazon.com/STS/latest/APIReference/API_Credentials.html)).

I'm concerned about something else - if you create the expiration timestamp in the json reply part - where do you store it? how do you verify that after an hour this role does not have access?

src/test/unit_tests/test_sts.js
```diff
@@ -392,7 +392,8 @@ function validate_assume_role_response(json, expected_arn, expected_role_id, ass
     // validate credentials
     const credentials = result.Credentials[0];
     assert.ok(credentials && credentials.AccessKeyId[0] && credentials.SecretAccessKey[0]);
-    assert.equal(credentials.Expiration[0], config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS);
+    const creds_generation_time = credentials.Expiration[0] - config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS;
+    assert(creds_generation_time < Date.now());
```
shirady, line 396: Be aware that tests based on time comparison tend to create problems in the CI.

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -35,12 +35,13 @@ async function assume_role(req) {
...
```
shirady, line 38: How will the expiration look like? I mean will we see the milli seconds? (we should not as I understand it for example: `<Expiration>2019-11-09T13:34:41Z</Expiration>`)

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -35,12 +35,13 @@ async function assume_role(req) {
...
                     SessionToken: generate_session_token({
                         access_key: access_keys.access_key.unwrap(),
                         secret_key: access_keys.secret_key.unwrap(),
                         assumed_role_access_key: assumed_role.access_key
-                    }, expiry)
+                    // generate_session_token expects an expiry in seconds
```
shirady, line 43: You can move this comment to a JDoc explanation in `generate_session_token` and rename the `expiry` to `expiry_ms` (in milliseconds) hoping it will be easier to understand.

src/endpoint/sts/ops/sts_post_assume_role.js
shirady, line 1: Related to the `expiry` - currently you only use the default, what about parsing the DurationSeconds from the request? (In AWS CLI it is `--duration-seconds`), related links:
https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html
https://docs.aws.amazon.com/cli/latest/reference/sts/assume-role.html

src/endpoint/sts/ops/sts_post_assume_role.js
shirady: I think there is a difference - this is the API of "AssumeRole" and you send me the "GetSessionToken".
- GetSessionToken - from the output (with milliseconds): <Expiration>2011-07-11T19:55:29.611Z</Expiration>
- but in AssumeRole - from the output (without milliseconds): <Expiration>2019-11-09T13:34:41Z</Expiration> (it is part of the data inside the Credentials data type).

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -21,10 +22,28 @@ async function assume_role(req) {
...
+        const duration_seconds = Number(req.body.duration_seconds);
+        // If it was provided, check if it's within the allowed range dictated by Amazon
+        if (duration_seconds < 900 || duration_seconds > 43200) {
+            dbg.error('Value', duration_seconds, 'for duration_seconds is not within the allowed range of 900-43200 seconds');
+            throw new StsError(StsError.ValidationError);
+        }
+        seconds_until_expiration = duration_seconds;
+    } else {
+        seconds_until_expiration = Math.ceil(config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS / 1000);
+    }
```
shirady, line 39: Please take it to another function (or functions) that would parse and validate the input and return what you need (in this case `expiration`).

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -21,10 +22,28 @@ async function assume_role(req) {
...
+
+    let seconds_until_expiration;
+    // Check if duration_seconds was provided via the CLI, otherwise resort to the default
+    if (req.body.duration_seconds) {
+        // duration_seconds is a string, convert it to a number
+        const duration_seconds = Number(req.body.duration_seconds);
+        // If it was provided, check if it's within the allowed range dictated by Amazon
+        if (duration_seconds < 900 || duration_seconds > 43200) {
+            dbg.error('Value', duration_seconds, 'for duration_seconds is not within the allowed range of 900-43200 seconds');
+            throw new StsError(StsError.ValidationError);
```
shirady, line 34: I think you can improve the error message - you can leave it as a gap and handle it in the future.

The general message is 'The input fails to satisfy the constraints specified by an AWS service.'
But in real cases, you see it is a specific message to the case, in your example:
"An error occurred (ValidationError) when calling the AssumeRole operation: 1 validation error detected: Value '43201' at 'durationSeconds' failed to satisfy constraint: Member must have a value less than or equal to 43200".
I remembered that @guymguym told me that we are compatible with AWS, but being compatible with the error message is not something we have to... In my opinion is it more user-friendly.

I did something like that for IAM in case you want. In the future I can move the function to a higher module in the stuck for reuse.
https://github.com/noobaa/noobaa-core/blob/8b634ddf35aaf3789539bd7f215c0bc09640d218/src/endpoint/iam/iam_utils.js#L104-L119

shirady (approved):
LGTM

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -49,11 +53,62 @@ async function assume_role(req) {
...
+        constraint = 'less';
+        constraint_value = max_value;
+    }
+    if (should_throw) {
+        const message_with_details = `Value ${input_value} for durationSeconds failed to satisfy constraint:
+            Member must have value ${constraint} than or equal to ${constraint_value}`;
+        const { code, http_code } = StsError.ValidationError;
+        throw new StsError({ code, message: message_with_details, http_code });
+    }
+}
```
guymguym, line 92: I suggest to separate to a parse function like we do for other api's.
Also should have the min/max as configs too.
Also notice to check that this is even parseable as a valid number or else error.
Also no need for moment for some date calculations.

```js
/**
 * @param {string|undefined} duration_input
 * @returns {number} duration in milliseconds
 */
function parse_sts_duration(duration_input) {
    if (duration_input === undefined) {
        return config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS;
    }

    const duration_sec = Number(duration_input);
    if (!Number.isInteger(duration_sec)) {
        throw new StsError( ?? );  // what error when input is not a number? ValidationError?
    }
    if (duration_sec < config.STS_MIN_DURATION_SECONDS) {
        throw new StsError(sts_duration_validation_error(duration_input, 'greater', config.STS_MIN_DURATION_SECONDS));
    }
    if (duration_sec > config.STS_MAX_DURATION_SECONDS) {
        throw new StsError(sts_duration_validation_error(duration_input, 'less', config.STS_MAX_DURATION_SECONDS));
    }

    const duration_ms = duration_sec * 1000;
    return duration_ms;
}

function sts_duration_validation_error(duration_input, constraint, constraint_value) {
    return {
        ... StsError.ValidationError,
        message: `Value ${duration_input} for durationSeconds failed to satisfy constraint: Member must have value ${constraint} than or equal to ${constraint_value}`,
    };
}

function get_expiration_date(duration_ms) {
    return new Date(Date.now() + duration_ms);
}
```

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -21,10 +22,13 @@ async function assume_role(req) {
         }
         throw new StsError(StsError.InternalFailure);
     }
+
+    const { expiration, seconds_until_expiration } = _get_expiration_data(req);
```
guymguym, line 26: I think this parsing should be done before calling sts_sdk.get_assumed_role.
I would call like this:
```js
const duration_ms = parse_sts_duration(req.body.duration_seconds);
const expiration_date = new Date(Date.now() + duration_ms);
```

guymguym: Also I see that our http_utils parses CONTENT_TYPE_APP_FORM_URLENCODED and converts all keys with _.snakeCase. I think this is a bad idea which can allow unintended keys to be used. Can we get rid of it? This is only used by sts and iam so not a lot of cases to fix.

guymguym (changes requested):

src/endpoint/sts/ops/sts_post_assume_role.js
shirady: @guymguym why do you set the constants (`STS_MIN_DURATION_SECONDS`, `STS_MAX_DURATION_SECONDS`, `STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS`) in the config file?
and not a file with other constants because those values should not be configurable.

src/endpoint/sts/ops/sts_post_assume_role.js
shirady: @guymguym IAM service was written based on STS, so if it's in STS it's in IAM...
I didn't see any problem with this function (I printed the keys during debugging), for example:

> Aug-27 7:41:24.015 [nsfs/3080]   [LOG] CONSOLE:: SDSD req.body Action=CreateUser&Version=2010-05-08&Path=%2Fabc%2Fdef%2F&UserName=Benny
> Aug-27 7:41:24.016 [nsfs/3080]   [LOG] CONSOLE:: SDSD res [Object: null prototype] { Action: ‘CreateUser’, Version: ‘2010-05-08’, Path: ‘/abc/def/’, UserName: ‘Benny’ }
> Aug-27 7:41:24.017 [nsfs/3080]   [LOG] CONSOLE:: SDSD renamed { action: ‘CreateUser’, version: ‘2010-05-08’, path: ‘/abc/def/’, user_name: ‘Benny’ }

If you decide to change it, it should be in a different PR.

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -49,11 +52,52 @@ async function assume_role(req) {
...
+*/
 function generate_session_token(auth_options, expiry) {
     dbg.log1('sts_post_assume_role.make_session_token: ', auth_options, expiry);
     return jwt_utils.make_auth_token(auth_options, { expiresIn: expiry });
 }
 
+// TODO: Generalize and move to a utils file in the future
+/**
+ * @param {String|undefined} duration_input
+ * @returns {Number} duration in milliseconds
```
guymguym, line 68: Need to be consistent because you made this function return duration in seconds and still return milliseconds from config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS in the default case.

I would naturally prefer durations to be in millis internally, and only convert to duration_sec on demand - which would be when calling  generate_session_token() above.

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -49,11 +52,52 @@ async function assume_role(req) {
...
+function _sts_duration_validation_error(duration_input, constraint, constraint_value) {
+    return {
+        ...StsError.ValidationError,
+        message: `Value ${duration_input} for durationSeconds failed to satisfy constraint: Member must have value ${constraint} than or equal to ${constraint_value}`,
+    };
+}
+
+function _calculate_expiration_time(seconds_until_expiration) {
+    const duration_ms = seconds_until_expiration * 1000;
+    return Math.ceil((Date.now() + duration_ms) / 1000);
```
guymguym, line 98: I don't understand why you divide by 1000 here? We are supposed to return a time from this function, so shouldn't it be in ms from epoc? it goes back to the previous comment - I think all times should be in ms to prevent confution, and only for specific functions that check for seconds we should convert - which in our case would be for generate_session_token(), and naturally also when parsing the duration input which is in seconds, but then I would expect to convert it to ms for working with it further.

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -12,6 +12,9 @@ const { CONTENT_TYPE_APP_FORM_URLENCODED } = require('../../../util/http_utils')
  */
 async function assume_role(req) {
     dbg.log1('sts_post_assume_role body: ', req.body);
+    const ms_until_expiration = _parse_sts_duration(req.body.duration_seconds);
+    const seconds_until_expiration = Math.ceil(ms_until_expiration / 1000);
+    const expiration = _calculate_expiration_time(ms_until_expiration);
```
guymguym, line 17: Lets keep the same terms
```suggestion
    const duration_ms = _parse_sts_duration(req.body.duration_seconds);
    const duration_sec = Math.ceil(duration_ms / 1000);
    const expiration_time = Date.now() + duration_ms;
```

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -49,11 +53,53 @@ async function assume_role(req) {
...
+    if (duration_input === undefined) {
+        return config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS;
+    }
+
+    const duration_ms = Number(duration_input) * 1000;
+    const duration_seconds = Math.ceil(duration_ms / 1000);
+
+    if (!Number.isInteger(duration_ms)) {
+        throw new StsError(StsError.InvalidParameterValue);
+    }
```
guymguym, line 81: Better check the result of parsing the number before making any arithmetics on it that might change the interpretation of the input. And then I would add the `const duration_ms = duration_sec * 1000` right before the return.

```suggestion
    const duration_sec = Number(duration_input);
    if (!Number.isInteger(duration_sec)) {
        throw new StsError(StsError.InvalidParameterValue);
    }
```

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -49,11 +53,53 @@ async function assume_role(req) {
...
+function _sts_duration_validation_error(duration_input, constraint, constraint_value) {
+    return {
+        ...StsError.ValidationError,
+        message: `Value ${duration_input} for durationSeconds failed to satisfy constraint: Member must have value ${constraint} than or equal to ${constraint_value}`,
+    };
+}
+
+function _calculate_expiration_time(ms_until_expiration) {
+    return Date.now() + ms_until_expiration;
+}
```
guymguym, line 101: This is trivial and can just be inlined

guymguym (commented):
final minor comments

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -35,12 +39,12 @@ async function assume_role(req) {
                 Credentials: {
                     AccessKeyId: access_keys.access_key.unwrap(),
                     SecretAccessKey: access_keys.secret_key.unwrap(),
-                    Expiration: expiry,
+                    Expiration: expiration_time_sec,
```
guymguym, line 42: This [API](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html#API_AssumeRole_Examples)  should return this element as ISO date string, like - `<Expiration>2019-11-09T13:34:41Z</Expiration>`, so we can't just return it as a number (regardless if the units is sec or ms).

Use the same function that S3 APIs uses - `s3_utils.format_s3_xml_date(time)` and send it the expiration time (in ms which is the standard units for js dates).

```suggestion
                    Expiration: s3_utils.format_s3_xml_date(expiration_time),
```

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -12,6 +12,9 @@ const { CONTENT_TYPE_APP_FORM_URLENCODED } = require('../../../util/http_utils')
  */
 async function assume_role(req) {
     dbg.log1('sts_post_assume_role body: ', req.body);
+    const duration_ms = _parse_sts_duration(req.body.duration_seconds);
+    const duration_sec = Math.ceil(duration_ms / 1000);
+    const expiration_time_sec = Math.ceil((Date.now() + duration_ms) / 1000);
```
guymguym, line 17: see previous comment for reason
```suggestion
    const expiration_time = Date.now() + duration_ms;
```

guymguym (changes requested):

guymguym (approved):
The code looks ok to me, but did you test it with aws sdk?

PR #8255, opened 2024-08-05
Add support for "Server" and "x-noobaa-available-storage-classes" headers

### LINKED ISSUE #8229: S3 head-bucket should return a header with the service identifier (config option)
## Is your feature request related to a problem? Please describe.

An S3 client needs to identify the endpoint is noobaa, and in addition what are the supported storage-classes it allows per bucket, in order to detect that using an S3 request.

Specifically when we configure to accept only GLACIER storage-class we want the client to be able to autodetect that using an API.

## Describe the solution you'd like

The [common header](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html) "Server" with value "NooBaa" is probably a good idea to be returned in all S3 API requests -

> Server | The name of the server that created the response.Type: StringDefault: AmazonS3

This value should be a config option so that downstream products can set accordingly.

In addition, head-bucket should return a custom header to specify which storage classes are supported - something like this:
```
"x-noobaa-available-storage-classes": "STANDARD,GLACIER"
```

This reply would be based on the config options to [enable GLACIER](https://github.com/noobaa/noobaa-core/blob/60b2be77071156bef42e619375f53a66fa8c25f6/config.js#L776) and [disable STANDARD](https://github.com/noobaa/noobaa-core/blob/60b2be77071156bef42e619375f53a66fa8c25f6/config.js#L188) class that we already have.

## Describe alternatives you've considered
Considered using a non S3 API to expose this information, but this would make it hard to consume by clients.

tangledbytes:
### Explain the changes
This PR adds support for:
1. `Server` HTTP header whose value is constant and is sent as part of every response.
2. `x-noobaa-available-storage-classes` which is sent as part of `HeadBucket` response. The value could be any combination of storage classes `GLACIER` and `STANDARD` based on NooBaa config.

image
image

Fixed #8229 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -269,6 +269,7 @@ function create_endpoint_handler(init_request_sdk, virtual_hosts, sts, logger) {
 
     /** @type {EndpointHandler} */
     const endpoint_request_handler = (req, res) => {
+        endpoint_utils.set_noobaa_server_header(res);
         endpoint_utils.prepare_rest_request(req);
         req.virtual_hosts = virtual_hosts;
         if (logger) req.bucket_logger = logger;
@@ -287,6 +288,7 @@ function create_endpoint_handler(init_request_sdk, virtual_hosts, sts, logger) {
     };
     /** @type {EndpointHandler} */
     const endpoint_sts_request_handler = (req, res) => {
+        endpoint_utils.set_noobaa_server_header(res);
         endpoint_utils.prepare_rest_request(req);
         init_request_sdk(req, res);
         return sts_rest(req, res);
@@ -298,6 +300,7 @@ function create_endpoint_handler(init_request_sdk, virtual_hosts, sts, logger) {
 function create_endpoint_handler_iam(init_request_sdk) {
     /** @type {EndpointHandler} */
     const endpoint_iam_request_handler = (req, res) => {
+        endpoint_utils.set_noobaa_server_header(res);
         endpoint_utils.prepare_rest_request(req);
         init_request_sdk(req, res);
         return iam_rest(req, res);
--- a/src/endpoint/s3/ops/s3_head_bucket.js
+++ b/src/endpoint/s3/ops/s3_head_bucket.js
@@ -1,13 +1,14 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
+const s3_utils = require("../s3_utils");
+
 /**
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketHEAD.html
  */
-async function head_bucket(req) {
-    await req.object_sdk.read_bucket({ name: req.params.bucket });
-    // only called to check for existence
-    // no headers or reply needed
+async function head_bucket(req, res) {
+    const bucket_info = await req.object_sdk.read_bucket({ name: req.params.bucket });
+    s3_utils.set_response_supported_storage_classes(res, bucket_info.supported_storage_classes);
 }
 
 module.exports = {
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -36,6 +36,8 @@ const DEFAULT_OBJECT_ACL = Object.freeze({
 const XATTR_SORT_SYMBOL = Symbol('XATTR_SORT_SYMBOL');
 const base64_regex = /^(?:[A-Za-z0-9+/]{4})*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?$/;
 
+const X_NOOBAA_AVAILABLE_STORAGE_CLASSES = 'x-noobaa-available-storage-classes';
+
  /**
  * get_default_object_owner returns bucket_owner info if exists
  * else it'll return the default owner
@@ -320,6 +322,14 @@ function set_response_object_md(res, object_md) {
 
         res.setHeader('x-amz-restore', restore);
     }
+}
+
+/**
+ * @param {nb.S3Response} res 
+ * @param {Array<string>} [supported_storage_classes]
+ */
+function set_response_supported_storage_classes(res, supported_storage_classes = []) {
+    res.setHeader(X_NOOBAA_AVAILABLE_STORAGE_CLASSES, supported_storage_classes);
 }
 
 /**
@@ -752,4 +762,4 @@ exports.parse_restore_request_days = parse_restore_request_days;
 exports.parse_version_id = parse_version_id;
 exports.get_object_owner = get_object_owner;
 exports.get_default_object_owner = get_default_object_owner;
-
+exports.set_response_supported_storage_classes = set_response_supported_storage_classes;
--- a/src/endpoint/endpoint_utils.js
+++ b/src/endpoint/endpoint_utils.js
@@ -3,6 +3,7 @@
 
 const querystring = require('querystring');
 const http_utils = require('../util/http_utils');
+const pkg = require('../../package.json');
 
 function prepare_rest_request(req) {
     // generate request id, this is lighter than uuid
@@ -38,6 +39,10 @@ function parse_source_url(source_url) {
     return { query, bucket, key };
 }
 
+function set_noobaa_server_header(res) {
+    res.setHeader('Server', `NooBaa/${pkg.version}`);
+}
 
 exports.prepare_rest_request = prepare_rest_request;
 exports.parse_source_url = parse_source_url;
+exports.set_noobaa_server_header = set_noobaa_server_header;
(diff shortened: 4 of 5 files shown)


src/endpoint/s3/s3_utils.js
```diff
@@ -322,6 +322,21 @@ function set_response_object_md(res, object_md) {
...
 
+/**
+ * @param {nb.S3Response} res 
+ */
+function set_response_supported_storage_classes(res) {
+    const storage_classes = [];
+    if (!config.DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD) {
+        storage_classes.push('STANDARD');
+    }
+    if (config.NSFS_GLACIER_ENABLED && config.NSFS_GLACIER_LOGS_ENABLED) {
```
guymguym, line 333: no need to require logs enabled, it might be that we enable glacier with fs policies to handle the transitions rather than logging.
```suggestion
    if (config.NSFS_GLACIER_ENABLED) {
```

src/endpoint/s3/ops/s3_head_bucket.js
```diff
@@ -1,13 +1,14 @@
...
 
+const s3_utils = require("../s3_utils");
+
 /**
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketHEAD.html
  */
-async function head_bucket(req) {
+async function head_bucket(req, res) {
+    s3_utils.set_response_supported_storage_classes(res);
     await req.object_sdk.read_bucket({ name: req.params.bucket });
```
guymguym, line 11: it would be better if this response was also responding based on the actual bucket namespace type so for namespace_fs buckets it would check config.NSFS_GLACIER_ENABLED, but for others it won't, which means that instead of setting this generically here, we should return the supported classes from read_bucket...
it might seem unneeded at the moment because currently this config is not used elsewhere, but when we do want to enable it for nsfs buckets in container deployments it would report the wrong classes for buckets.

guymguym (changes requested):

src/endpoint/s3/s3_utils.js
tangledbytes: Actually I added this intentionally because I found myself tripping over it. I enabled `NSFS_GLACIER_ENABLED` but my restores were not working. The reason was that there were no logs being written as the default is `false`. If we report `GLACIER` being supported then at least should truly be working, right?

Maybe we can make it backend specific (we get backend anyway from config) and then report it but I am inclined to keep this check here so as to not cause confusion.

guymguym: I still don't want to require enabling logging for that. NSFS_GLACIER_ENABLED is enough to tell if we support this class or not.

src/endpoint/s3/s3_utils.js
```diff
@@ -322,6 +322,21 @@ function set_response_object_md(res, object_md) {
     }
 }
 
+/**
+ * @param {nb.S3Response} res 
+ */
+function set_response_supported_storage_classes(res) {
+    const storage_classes = [];
+    if (!config.DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD) {
+        storage_classes.push('STANDARD');
```
guymguym, line 331:
```suggestion
        storage_classes.push(STORAGE_CLASS_STANDARD);
```

src/endpoint/s3/s3_utils.js
```diff
@@ -322,6 +322,21 @@ function set_response_object_md(res, object_md) {
...
+    }
+    if (config.NSFS_GLACIER_ENABLED && config.NSFS_GLACIER_LOGS_ENABLED) {
+        storage_classes.push('GLACIER');
```
guymguym, line 334:
```suggestion
        storage_classes.push(STORAGE_CLASS_GLACIER);
```

src/endpoint/s3/s3_utils.js
```diff
@@ -322,6 +322,21 @@ function set_response_object_md(res, object_md) {
...
+    }
+
+    res.setHeader('x-noobaa-available-storage-classes', storage_classes);
```
guymguym, line 337: I know it's a one appearance, but for headers we return this is like an API so I would prefer to set them as const
```js
const X_NOOBAA_AVAILABLE_STORAGE_CLASSES = 'x-noobaa-available-storage-classes';
...
    res.setHeader(X_NOOBAA_AVAILABLE_STORAGE_CLASSES, storage_classes);
```

src/util/http_utils.js
```diff
@@ -318,6 +318,8 @@ async function parse_request_body(req, options) {
 }
 
 function send_reply(req, res, reply, options) {
+    res.setHeader('Server', 'NooBaa');
```
guymguym, line 321: 1. I think this should be called at the beginning of the request, not at the end, so that it also applies to errors or any other forms of replying, so I would move it to endpoint.js or maybe add it inside the existing calls to endpoint_utils.prepare_rest_request().
2. I think we should return also the server version - like `Server: Apache/2.4.1 (Unix)` which you can get from `const pkg = require('../../package.json'); ... pkg.version`

src/endpoint/s3/s3_utils.js
```diff
@@ -322,6 +324,18 @@ function set_response_object_md(res, object_md) {
     }
 }
 
+/**
+ * @param {nb.S3Response} res 
+ * @param {Array<string>} [supported_storage_classes]
+ */
+function set_response_supported_storage_classes(res, supported_storage_classes) {
+    if (!supported_storage_classes) {
+        supported_storage_classes = [STORAGE_CLASS_STANDARD];
```
guymguym, line 333: For completeness, since this config is not specific to nsfs, I would check it here too -
```
        if (!config.DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD) {
            storage_classes.push(s3_utils.STORAGE_CLASS_STANDARD);
        }
```

src/endpoint/s3/s3_utils.js
```diff
@@ -322,6 +324,18 @@ function set_response_object_md(res, object_md) {
...
+    }
+
+    res.setHeader(X_NOOBAA_AVAILABLE_STORAGE_CLASSES, supported_storage_classes);
```
guymguym, line 336: Will this header be sent if the configuration is such that we get an empty array? 
Lets just make sure we behave in a determined way in case so users will not be confused.

tangledbytes: The headers aren't sent in that case I believe (at least I couldn't see them in the outputs)

PS: I have added screenshots in the PR description.

src/endpoint/endpoint.js
```diff
@@ -269,6 +270,8 @@ function create_endpoint_handler(init_request_sdk, virtual_hosts, sts, logger) {
 
     /** @type {EndpointHandler} */
     const endpoint_request_handler = (req, res) => {
+        res.setHeader('Server', `NooBaa/${pkg.version}`);
```
guymguym, line 273: but what about the other handlers - sts/iam?
can you add it as function in entpoint_utils and call from all handlers?

src/endpoint/s3/ops/s3_head_bucket.js
```diff
@@ -1,13 +1,14 @@
...
 /**
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketHEAD.html
  */
-async function head_bucket(req) {
-    await req.object_sdk.read_bucket({ name: req.params.bucket });
-    // only called to check for existence
-    // no headers or reply needed
+async function head_bucket(req, res) {
+    const bucket_info = await req.object_sdk.read_bucket({ name: req.params.bucket });
+    s3_utils.set_response_supported_storage_classes(res, bucket_info.supported_storage_classes || []);
```
guymguym, line 11:
```suggestion
    s3_utils.set_response_supported_storage_classes(res, bucket_info.supported_storage_classes);
```

src/endpoint/s3/s3_utils.js
```diff
@@ -322,6 +324,18 @@ function set_response_object_md(res, object_md) {
     }
 }
 
+/**
+ * @param {nb.S3Response} res 
+ * @param {Array<string>} [supported_storage_classes]
+ */
+function set_response_supported_storage_classes(res, supported_storage_classes = []) {
+    if (!config.DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD) {
+        supported_storage_classes.push(STORAGE_CLASS_STANDARD);
```
guymguym, line 333: so now the storage classes returned by the bucket always do not include STANDARD, which just becomes confusing. I didn't mean that. I only meant that if the bucketspace does not return anything, that we set the default based on the config too, but separating the responsibility like this is too weird. So I think we should just do this instead - have the bucketspace return the list, and if it doesn't return anything, just not set the header.

guymguym (approved):
just missing the call to set_noobaa_server_header to iam handler. approving

PR #8258, opened 2024-08-05
NC | ConfigFS Refactoring

romayalon:
### Explain the changes
1. Added ConfigFS file for config directory manipulation.
2. Removed functions from manage_nsfs_utils.js.
3. Updated callers to use configFS - 
    a. manage_nsfs.js (noobaa-cli) including manage_nsfs_validations.js.
    b. bucketspaceFS.js.
    c. accountspaceFS.js.
    d. Health CLI.
    e. NC Bucket Logging CLI.

### Issues: Fixed #xxx / Gap #xxx

1. Gap - ConfigFS API is not finalized.
2. Gap - Tests were not updated to use ConfigFS, this way we can also test the correctness of ConfigFS itself without adding new tests. but we might want eventually to do it.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

PR #8260, opened 2024-08-06
NC | NSFS | Rename Constants With Prefix `GLOBAL_CONFIG`

shirady:
### Explain the changes
As part of adding the class `ConfigFS` (in PR #8258) and removing the term `global_config` in manage_nsfs - since this term was not clear enough and might cause confusion - we will change the constants that had `GLOBAL_CONFIG_` as a prefix:
1. rename `GLOBAL_CONFIG_ROOT_ALL_FLAG` to `CLI_MUTUAL_FLAGS`
2. rename `GLOBAL_CONFIG_ROOT` to `CONFIG_ROOT_FLAG`
3. rename `GLOBAL_CONFIG_OPTIONS` to `CLI_MUTUAL_OPTIONS`

### Issues:
1. None, it is a small step in the plan that was mentioned in [this comment](https://github.com/noobaa/noobaa-core/pull/8161#discussion_r1660695401).

### Testing Instructions:
1. None (would be tested in the CI)

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/manage_nsfs/manage_nsfs_constants.js
+++ b/src/manage_nsfs/manage_nsfs_constants.js
@@ -36,50 +36,50 @@ const CONFIG_SUBDIRS = {
     ACCESS_KEYS: 'access_keys'
 };
 
-const GLOBAL_CONFIG_ROOT = 'config_root';
-const GLOBAL_CONFIG_OPTIONS = new Set([GLOBAL_CONFIG_ROOT, 'config_root_backend', 'debug']);
+const CONFIG_ROOT_FLAG = 'config_root';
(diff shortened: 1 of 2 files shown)


romayalon:
@shirady IMO we want to avoid using too general terms like global/general flags/options in CLI context, I would just add CLI_ prefix, WDYT?

shirady:
> @shirady IMO we want to avoid using too general terms like global/general flags/options in CLI context, I would just add CLI_ prefix, WDYT?

@romayalon I would accept the suggestion of `CLI` prefix and add `CLI_MUTUAL_` hopefully clearer than the global/general terms.

romayalon (approved):

PR #8269, opened 2024-08-08
NC | NSFS | CLI | Update `fetch_existing_account_data`

shirady:
### Explain the changes
1. In the past, for accout update the identifier could be name or access_key, but today it is only name. Therefore, suggesting to remove the other option (`get_account_by_access_key`) as it will always use `get_account_by_name`.

### Issues:
1. None

### Testing Instructions:
1.  None (will be tested in the CI).

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -351,17 +351,11 @@ async function fetch_existing_account_data(action, target, decrypt_secret_key) {
     const options = { show_secrets: true, decrypt_secret_key };
     let source;
     try {
-        source = target.name ?
-            await config_fs.get_account_by_name(target.name, options) :
-            await config_fs.get_account_by_access_key(target.access_keys[0].access_key, options);
+        source = await config_fs.get_account_by_name(target.name, options);
     } catch (err) {
-        dbg.log1('NSFS Manage command: Could not find account', target, err);
+        dbg.warn(`fetch_existing_account_data: account with name ${target.name} got an error:`, err);
         if (err.code === 'ENOENT') {
-            if (target.name === undefined) {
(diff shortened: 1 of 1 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -352,17 +352,11 @@ async function fetch_existing_account_data(action, target, decrypt_secret_key) {
     const options = { show_secrets: true, decrypt_secret_key };
     let source;
     try {
-        source = target.name ?
-            await config_fs.get_account_by_name(target.name, options) :
-            await config_fs.get_account_by_access_key(target.access_keys[0].access_key, options);
+        source = await config_fs.get_account_by_name(target.name, options);
     } catch (err) {
-        dbg.log1('NSFS Manage command: Could not find account', target, err);
+        dbg.warn('fetch_existing_account_data: got an error on', target.name, err);
```
naveenpaul1, line 357: maybe more readable message,
`fetch_existing_account_data: account with name %s could not find`

shirady: I will change the string to a template, hoping it will be a more readable message, but I can't add the "could not find" because it is an error that we receive after "get_account_by_name" and "ENOENT" is only one option, can be more causes.

naveenpaul1: templet looks good to me

naveenpaul1 (approved):

romayalon (commented):
@shirady Before merging this change we need to validate with CES team they don't use it so we won't break their code.

romayalon (approved):
As we confirmed it on Slack, approving :)

PR #8279, opened 2024-08-12
NC | Config Dir Restructure

romayalon:
### Explain the changes

This PR is based on @alphaprinz https://github.com/noobaa/noobaa-core/pull/8167 PR.

1. Added to the new structure the following paths - 
```
identities/
     {account_id}/account.json
     
accounts_by_name/
     {account_name}.symlink -> ../identities/{account_id}/account.json
```
3. Re-implemented create/update/delete account in configFS to adhere the new config dir structure.
4. Backwards compatibility - 
- Added backward compatibility to the following accounts getter functions - 
- get_account_by_name()
- is_account_exists({name: string})
- list_accounts()

5. Updated getter callers to these functions to have { fallback:true } options.
6. Converted tests to use config_fs.
7. Added backward compatibility tests.

1. Add structure tests.
2. Add list account backward compatibility tests after we agree on the expected result.
3. Update Docs.
4. Gap - When starting to use the new mapping 

### Testing Instructions:
1. `sudo npx jest --testRegex=jest_tests/test_config_fs_backward_compatibility.test.js`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -8,7 +8,7 @@ const minimist = require('minimist');
 const config = require('../../config');
 const P = require('../util/promise');
 const nb_native = require('../util/nb_native');
-const { ConfigFS, JSON_SUFFIX } = require('../sdk/config_fs');
+const { ConfigFS } = require('../sdk/config_fs');
 const cloud_utils = require('../util/cloud_utils');
 const native_fs_utils = require('../util/native_fs_utils');
 const mongo_utils = require('../util/mongo_utils');
@@ -387,8 +387,8 @@ async function add_account(data) {
     await manage_nsfs_validations.validate_account_args(config_fs, data, ACTIONS.ADD, undefined);
 
     const access_key = has_access_keys(data.access_keys) ? data.access_keys[0].access_key : undefined;
-    const name_exists = await config_fs.is_account_exists({ name: data.name });
-    const access_key_exists = access_key && await config_fs.is_account_exists({ access_key });
+    const name_exists = await config_fs.is_account_exists_by_name(data.name);
+    const access_key_exists = access_key && await config_fs.is_account_exists_by_access_key(access_key);
 
     const event_arg = data.name ? data.name : access_key;
     if (name_exists || access_key_exists) {
@@ -405,7 +405,7 @@ async function add_account(data) {
     // for validating against the schema we need an object, hence we parse it back to object
     const account = encrypted_data ? JSON.parse(encrypted_data) : data;
     nsfs_schema_utils.validate_account_schema(account);
-    await config_fs.create_account_config_file(data.name, account, true);
+    await config_fs.create_account_config_file(account);
     write_stdout_response(ManageCLIResponse.AccountCreated, data, { account: event_arg });
 }
 
@@ -434,7 +434,7 @@ async function update_account(data, is_flag_iam_operate_on_root_account) {
         // for validating against the schema we need an object, hence we parse it back to object
         const account = encrypted_data ? JSON.parse(encrypted_data) : data;
         nsfs_schema_utils.validate_account_schema(account);
-        await config_fs.update_account_config_file(data.name, account, undefined, undefined);
+        await config_fs.update_account_config_file(account);
         write_stdout_response(ManageCLIResponse.AccountUpdated, data);
         return;
     }
@@ -446,9 +446,9 @@ async function update_account(data, is_flag_iam_operate_on_root_account) {
         secret_key: data.access_keys[0].secret_key,
     };
 
-    const name_exists = update_name && await config_fs.is_account_exists({ name: data.name });
+    const name_exists = update_name && await config_fs.is_account_exists_by_name(data.name, undefined);
     const access_key_exists = update_access_key &&
-        await config_fs.is_account_exists({ access_key: data.access_keys[0].access_key.unwrap() });
+        await config_fs.is_account_exists_by_access_key(data.access_keys[0].access_key.unwrap());
 
     if (name_exists || access_key_exists) {
         const err_code = name_exists ? ManageCLIError.AccountNameAlreadyExists : ManageCLIError.AccountAccessKeyAlreadyExists;
@@ -465,18 +465,17 @@ async function update_account(data, is_flag_iam_operate_on_root_account) {
     // for validating against the schema we need an object, hence we parse it back to object
     const parsed_data = JSON.parse(encrypted_data);
     nsfs_schema_utils.validate_account_schema(parsed_data);
-    if (update_name) {
-        await config_fs.create_account_config_file(new_name, parsed_data, true, [cur_access_key]);
-        await config_fs.delete_account_config_file(cur_name, data.access_keys);
-    } else if (update_access_key) {
-        await config_fs.update_account_config_file(cur_name, parsed_data, parsed_data.access_keys, [cur_access_key]);
-    }
+    await config_fs.update_account_config_file(parsed_data, {
+        old_name: update_name && cur_name,
+        new_access_keys_to_link: update_access_key && parsed_data.access_keys,
+        access_keys_to_delete: update_access_key && [{ access_key: cur_access_key }]
+    });
     write_stdout_response(ManageCLIResponse.AccountUpdated, data);
 }
 
 async function delete_account(data) {
     await manage_nsfs_validations.validate_account_args(config_fs, data, ACTIONS.DELETE, undefined);
-    await config_fs.delete_account_config_file(data.name, data.access_keys);
+    await config_fs.delete_account_config_file(data);
     write_stdout_response(ManageCLIResponse.AccountDeleted, '', { account: data.name });
 }
 
@@ -578,19 +577,10 @@ function filter_bucket(bucket, filters) {
  * @param {object} [filters]
  */
 async function list_config_files(type, wide, show_secrets, filters = {}) {
-    let entries;
-    // in case we have a filter by name, we don't need to read all the entries and iterate them
-    // instead we "mock" the entries array to have one entry and it is the name by the filter (we add it for performance)
-    const is_filter_by_name = filters.name !== undefined;
-    if (is_filter_by_name) {
-        entries = [{'name': filters.name + JSON_SUFFIX}];
-    } else {
-        entries = type === TYPES.ACCOUNT ?
-        await config_fs.list_root_accounts() :
-        await config_fs.list_buckets();
-    }
-
+    let entries = [];
     const should_filter = Object.keys(filters).length > 0;
+    const is_filter_by_name = filters.name !== undefined;
+
     // decryption causing mkm initalization
     // decrypt only if data has access_keys and show_secrets = true (no need to decrypt if show_secrets = false but should_filter = true)
     const options = {
@@ -599,19 +589,27 @@ async function list_config_files(type, wide, show_secrets, filters = {}) {
         silent_if_missing: true
     };
 
+    // in case we have a filter by name, we don't need to read all the entries and iterate them
+    // instead we "mock" the entries array to have one entry and it is the name by the filter (we add it for performance)
+    if (is_filter_by_name) {
+        entries = [filters.name];
+    } else if (type === TYPES.ACCOUNT) {
+        entries = await config_fs.list_accounts();
+    } else if (type === TYPES.BUCKET) {
+        entries = await config_fs.list_buckets();
+    }
+
     let config_files_list = await P.map_with_concurrency(10, entries, async entry => {
-        if (entry.name.endsWith(JSON_SUFFIX)) {
-            if (wide || should_filter) {
-                const data = type === TYPES.ACCOUNT ?
-                    await config_fs.get_account_by_name(entry.name, options) :
-                    await config_fs.get_bucket_by_name(entry.name, options);
-                if (!data) return undefined;
-                if (should_filter && !filter_list_item(type, data, filters)) return undefined;
-                // remove secrets on !show_secrets && should filter
-                return wide ? _.omit(data, show_secrets ? [] : ['access_keys']) : { name: entry.name.slice(0, entry.name.indexOf(JSON_SUFFIX)) };
-            } else {
-                return { name: entry.name.slice(0, entry.name.indexOf(JSON_SUFFIX)) };
-            }
+        if (wide || should_filter) {
+            const data = type === TYPES.ACCOUNT ?
+                await config_fs.get_account_by_name(entry, options) :
+                await config_fs.get_bucket_by_name(entry, options);
+            if (!data) return undefined;
+            if (should_filter && !filter_list_item(type, data, filters)) return undefined;
+            // remove secrets on !show_secrets && should filter
+            return wide ? _.omit(data, show_secrets ? [] : ['access_keys']) : { name: entry };
+        } else {
+            return { name: entry };
         }
     });
     // it inserts undefined for the entry '.noobaa-config-nsfs' and we wish to remove it
--- a/src/manage_nsfs/health.js
+++ b/src/manage_nsfs/health.js
@@ -10,7 +10,6 @@ const nb_native = require('../util/nb_native');
 const native_fs_utils = require('../util/native_fs_utils');
 const { read_stream_join } = require('../util/buffer_utils');
 const { make_https_request } = require('../util/http_utils');
-const { JSON_SUFFIX } = require('../sdk/config_fs');
 const { TYPES } = require('./manage_nsfs_constants');
 const { get_boolean_or_string_value, throw_cli_error, write_stdout_response } = require('./manage_nsfs_cli_utils');
 const { ManageCLIResponse } = require('./manage_nsfs_cli_responses');
@@ -327,7 +326,7 @@ class NSFSHealth {
             config_root_type_exists = await this.config_fs.validate_config_dir_exists(config_dir_path);
         } else if (type === TYPES.ACCOUNT) {
             // TODO - handle iam accounts when directory structure changes - read_account_by_id
-            config_dir_path = this.config_fs.accounts_dir_path;
+            config_dir_path = this.config_fs.accounts_by_name_dir_path;
             config_root_type_exists = await this.config_fs.validate_config_dir_exists(config_dir_path);
         }
         // TODO - this is not a good handling for that - we need to take it to an upper level
@@ -339,15 +338,16 @@ class NSFSHealth {
             };
         }
 
-        const entries = type === TYPES.BUCKET ?
-            await this.config_fs.list_buckets() :
-            await this.config_fs.list_root_accounts();
-
-        const config_files = entries.filter(entree => !native_fs_utils.isDirectory(entree) && entree.name.endsWith(JSON_SUFFIX));
+        let config_files;
+        if (type === TYPES.BUCKET) {
+            config_files = await this.config_fs.list_buckets();
+        } else {
+            config_files = await this.config_fs.list_accounts();
+        }
         for (const config_file of config_files) {
             // config_file get data or push error
             const { config_data = undefined, err_obj = undefined } =
-                await this.get_config_file_data_or_error_object(type, config_file.name);
+                await this.get_config_file_data_or_error_object(type, config_file);
             if (!config_data && err_obj) {
                 invalid_storages.push(err_obj.invalid_storage);
                 continue;
@@ -395,9 +395,9 @@ class NSFSHealth {
         } catch (err) {
             let err_code;
             const config_file_path = type === TYPES.BUCKET ?
-                await this.config_fs.get_bucket_path_by_name(config_file_name) :
+                this.config_fs.get_bucket_path_by_name(config_file_name) :
                 // TODO - should be changed to id when moving to new structure for supporting iam accounts
-                await this.config_fs.get_account_path_by_name(config_file_name);
+                this.config_fs.get_account_path_by_name(config_file_name);
 
             if (err.code === 'ENOENT') {
                 dbg.log1(`Error: Config file path should be a valid path`, config_file_path, err);
--- a/src/sdk/accountspace_fs.js
+++ b/src/sdk/accountspace_fs.js
@@ -5,7 +5,7 @@ const _ = require('lodash');
 const config = require('../../config');
 const dbg = require('../util/debug_module')(__filename);
 const P = require('../util/promise');
-const { ConfigFS, JSON_SUFFIX } = require('./config_fs');
+const { ConfigFS } = require('./config_fs');
 const native_fs_utils = require('../util/native_fs_utils');
 const { create_arn, get_action_message_title, check_iam_path_was_set } = require('../endpoint/iam/iam_utils');
 const { IAM_ACTIONS, MAX_NUMBER_OF_ACCESS_KEYS, IAM_DEFAULT_PATH,
@@ -153,7 +153,7 @@ class AccountSpaceFS {
                 const requested_account_encrypted = await nc_mkm.encrypt_access_keys(requested_account);
                 const account_string = JSON.stringify(requested_account_encrypted);
                 nsfs_schema_utils.validate_account_schema(JSON.parse(account_string));
-                await this.config_fs.update_account_config_file(params.username, JSON.parse(account_string));
+                await this.config_fs.update_account_config_file(JSON.parse(account_string));
             }
             this._clean_account_cache(requested_account);
             return {
@@ -193,7 +193,7 @@ class AccountSpaceFS {
             this._check_if_requested_account_is_root_account_or_IAM_user(action, requesting_account, account_to_delete);
             this._check_if_requested_is_owned_by_root_account(action, requesting_account, account_to_delete);
             await this._check_if_user_does_not_have_resources_before_deletion(action, account_to_delete);
-            await this.config_fs.delete_account_config_file(params.username);
+            await this.config_fs.delete_account_config_file(account_to_delete);
         } catch (err) {
             dbg.error(`AccountSpaceFS.${action} error`, err);
             throw native_fs_utils.translate_error_codes(err, native_fs_utils.entity_enum.USER);
@@ -268,8 +268,10 @@ class AccountSpaceFS {
             const requested_account_encrypted = await nc_mkm.encrypt_access_keys(requested_account);
             const account_to_create_access_keys_string = JSON.stringify(requested_account_encrypted);
             nsfs_schema_utils.validate_account_schema(JSON.parse(account_to_create_access_keys_string));
-            await this.config_fs.update_account_config_file(name_for_access_key, JSON.parse(account_to_create_access_keys_string),
-                [created_access_key_obj]);
+            await this.config_fs.update_account_config_file(
+                JSON.parse(account_to_create_access_keys_string),
+                { new_access_keys_to_link: [created_access_key_obj] }
+            );
             return {
                 username: requested_account.name,
                 access_key: created_access_key_obj.access_key,
@@ -356,8 +358,7 @@ class AccountSpaceFS {
             const requested_account_encrypted = await nc_mkm.encrypt_access_keys(requested_account);
             const account_string = JSON.stringify(requested_account_encrypted);
             nsfs_schema_utils.validate_account_schema(JSON.parse(account_string));
-            const name_for_access_key = params.username ?? requester.name;
-            await this.config_fs.update_account_config_file(name_for_access_key, JSON.parse(account_string));
+            await this.config_fs.update_account_config_file(JSON.parse(account_string));
             this._clean_account_cache(requested_account);
         } catch (err) {
             dbg.error(`AccountSpaceFS.${action} error`, err);
@@ -387,7 +388,7 @@ class AccountSpaceFS {
                 requesting_account, params.username);
             const username = params.username ?? requester.name; // username is not required
             await this._check_if_account_exists_by_access_key_symlink(action, access_key_id);
-            const requested_account = await this.config_fs.get_account_by_access_key(params.access_key, { show_secrets: true});
+            const requested_account = await this.config_fs.get_account_by_access_key(access_key_id, { show_secrets: true});
             this._check_username_match_to_requested_account(action, username, requested_account);
             this._check_access_key_belongs_to_account(action, requested_account, access_key_id);
             this._check_if_requested_account_same_root_account_as_requesting_account(action,
@@ -400,12 +401,9 @@ class AccountSpaceFS {
             const requested_account_encrypted = await nc_mkm.encrypt_access_keys(requested_account);
             const account_string = JSON.stringify(requested_account_encrypted);
             nsfs_schema_utils.validate_account_schema(JSON.parse(account_string));
-            const name_for_access_key = params.username ?? requester.name;
             await this.config_fs.update_account_config_file(
-                name_for_access_key,
                 JSON.parse(account_string),
-                undefined,
-                [params.access_key]
+                { access_keys_to_delete: [{ access_key: access_key_id }] }
             );
             this._clean_account_cache(requested_account);
         } catch (err) {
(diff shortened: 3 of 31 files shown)


src/sdk/config_fs.js
```diff
@@ -151,88 +218,184 @@ class ConfigFS {
...
+       return path.join('../', CONFIG_SUBDIRS.IDENTITIES, account_id, this.json('identity'));
+    }
+
+    /**
+     * get_account_or_user_path_by_id returns the full account/user path by id as following
+     * {config_dir}/identities/{account_id}/identity.json
+     * @param {string} account_id
+     * @returns {string} 
+    */
+    get_identity_path_by_id(account_id) {
```
shirady, line 277: I think you need to change the argument from "account_id" to "id" and update the JSDoc description.

src/sdk/config_fs.js
```diff
@@ -151,88 +218,184 @@ class ConfigFS {
...
+    /**
+     * get_identity_by_id returns the full account/user data by id from the following path
+     * {config_dir}/identities/{account_id}/identity.json
+     * @param {string} id
+     * @param {string} [type]
+     * @param {{show_secrets?: boolean, decrypt_secret_key?: boolean, silent_if_missing?: boolean}} [options]
+     * @returns {Promise<Object>} 
+    */
+    async get_identity_by_id(id, type, options = {}) {
+        const account_path = this.get_identity_path_by_id(id);
```
shirady, line 290:
```suggestion
        const identity_path = this.get_identity_path_by_id(id);
```

src/sdk/config_fs.js
```diff
@@ -151,88 +218,184 @@ class ConfigFS {
...
+        let identity = await this.get_identity_config_data(account_path, { ...options, silent_if_missing: true });
+
+        if (!identity && type === TYPES.ACCOUNT) {
+            identity = await this.search_accounts_by_id(id, options);
+        }
+        if (!identity && !options.silent_if_missing) {
+            const err = new Error(`Could not find identity by id ${id}`);
+            err.code = 'ENOENT';
```
shirady, line 298: I don't think it is a good idea to return an error as it returned from the FS.

src/sdk/config_fs.js
```diff
@@ -151,88 +218,184 @@ class ConfigFS {
...
+     * @param {string} [type]
+     * @param {{show_secrets?: boolean, decrypt_secret_key?: boolean, silent_if_missing?: boolean}} [options]
+     * @returns {Promise<Object>} 
+    */
+    async get_identity_by_id(id, type, options = {}) {
+        const account_path = this.get_identity_path_by_id(id);
+        let identity = await this.get_identity_config_data(account_path, { ...options, silent_if_missing: true });
+
+        if (!identity && type === TYPES.ACCOUNT) {
+            identity = await this.search_accounts_by_id(id, options);
```
shirady, line 294: If we didn't find the identity by the ID, why do we read again in all the account configs by name and not just the old accounts?

src/sdk/config_fs.js
```diff
@@ -151,88 +218,184 @@ class ConfigFS {
...
+     * search_accounts_by_id searches all accounts and finds an account that its _id matches the given id param
+     * @param {string} id
+     * @param {{show_secrets?: boolean, decrypt_secret_key?: boolean, silent_if_missing?: boolean}} [options]
+     * @returns {Promise<Object>} 
+    */
+    async search_accounts_by_id(id, options = {}) {
+        const old_account_names = await this.list_accounts();
+        for (const account_name of old_account_names) {
+            const account_data = await this.get_account_by_name(account_name, { ...options, silent_if_missing: true });
+            if (account_data._id === id) return account_data;
```
shirady, line 314: You are with `silent_if_missing: true`, hence the `account_data` can be undefined.

src/sdk/config_fs.js
```diff
@@ -151,88 +218,184 @@ class ConfigFS {
...
+     * {config_dir}/identities/{id}/identity.json
+     * @param {string[]} ids
+     * @returns {Promise<Object>} 
+    */
+    async get_identities_by_id(ids, options = {}) {
+        const res = [];
+        for (const id of ids) {
+            const id_path = this.get_identity_path_by_id(id);
+            const id_data = await this.get_identity_config_data(id_path, { ...options, silent_if_missing: true });
+            res.push(id_data);
```
shirady, line 329: Since `silent_if_missing: true` `id_data` can be undefined, do you want to push undefined?
If so, please add a comment.

src/sdk/config_fs.js
```diff
@@ -151,88 +218,184 @@ class ConfigFS {
...
-        } else if (identifier_object.access_key) {
-            path_to_check = this.get_account_path_by_access_key(identifier_object.access_key);
-            use_lstat = true;
-       } else {
-            throw new Error(`config_fs.is_account_exists - invalid identifier provided ${identifier_object}`);
+    async is_account_exists_by_name(account_name, owner_account_id) {
+        const path_to_check = this.get_account_or_user_path_by_name(account_name, owner_account_id);
+        let res = await native_fs_utils.is_path_exists(this.fs_context, path_to_check);
+
+        if (!res && account_name) {
```
shirady, line 374: I think you need to expand this condition:
1. the path doesn't exist
2. the account name is defined (account_name is not enough, I think we allowed it to be 0)
3. you don't need to search IAM accounts in the old account path, right?

src/sdk/config_fs.js
```diff
@@ -151,88 +218,184 @@ class ConfigFS {
...
-        return native_fs_utils.is_path_exists(this.fs_context, path_to_check, use_lstat);
+        return res;
+    }
+
+    /**
+     * is_identity_exists returns true if identity config path exists in config dir
+     * @param {string} id
+     * @returns {Promise<boolean>} 
+    */
+    async is_identity_exists(id) {
```
shirady, line 386: Why it is not searching the account in the old accounts directory?

src/sdk/config_fs.js
```diff
@@ -244,11 +407,23 @@ class ConfigFS {
...
         return account;
     }
 
+    /**
+     * _remove_config_file_suffix removes suffix from a list entry name
+     * @param {string} entry
+     * @param {string} suffix
+     * @returns {string}
+     */
+    _remove_config_file_suffix(entry, suffix) {
```
shirady, line 421:
```suggestion
    _remove_config_file_suffix(entry_name, suffix) {
```

You have a builtin function:
path.parse(entry_name).name
For information:
https://stackoverflow.com/questions/4250364/how-to-trim-a-file-extension-from-a-string-in-javascript

src/sdk/config_fs.js
```diff
@@ -265,104 +440,222 @@ class ConfigFS {
...
-     * list_root_accounts returns the root accounts array exists under the config dir
-     * @returns {Promise<Dirent[]>} 
-     */
-    async list_root_accounts(options) {
-        return nb_native().fs.readdir(this.fs_context, this.accounts_dir_path);
+    * @param {fs.Dirent} entry
+    * @param {string} suffix
+    * @returns {boolean}
+    */
+    _is_config_file(entry, suffix) {
```
shirady, line 461: I think this name might be confusing, it doesn't guarantee that the file blabla.json is indeed a config file.

src/sdk/config_fs.js
```diff
@@ -265,104 +440,222 @@ class ConfigFS {
...
      */
-    async create_account_config_file(account_name, account_data, symlink_new_access_keys, old_access_keys = []) {
-        const account_path = this.get_account_path_by_name(account_name);
-        await native_fs_utils.create_config_file(this.fs_context, this.accounts_dir_path, account_path, JSON.stringify(account_data));
+    async list_accounts() {
+        const new_entries = await nb_native().fs.readdir(this.fs_context, this.accounts_by_name_dir_path);
+        const new_accounts_names = new_entries
+            .filter(entry => this._is_config_file(entry, SYMLINK_SUFFIX))
+            .map(entry => this._remove_config_file_suffix(entry.name, SYMLINK_SUFFIX));
+
```
shirady, line 482: I can suggest that you do it in 1 iteration instead of 3 (filter, map, convert to set).

I Can share something (in case it helps):
```
        // we want a set with the account names without the temp config file and without name extension
        const account_names_entries_set = entries.reduce((acc, entry) => {
            if (!entry.name.includes(config.NSFS_TEMP_CONF_DIR_NAME)) {
                acc.add(path.parse(entry.name).name);
            }
            return acc;
        }, new Set());
```

src/sdk/config_fs.js
```diff
@@ -265,104 +440,222 @@ class ConfigFS {
...
-                await nb_native().fs.unlink(this.fs_context, cur_access_key_path);
-            }
+    async update_account_config_file(account_new_data, options = {}) {
+        const { name, _id, owner = undefined } = account_new_data;
+        const data_string = JSON.stringify(account_new_data);
+        const account_path = this.get_identity_path_by_id(_id);
+        const account_dir_path = this.get_identity_dir_path_by_id(_id);
+        await native_fs_utils.update_config_file(this.fs_context, account_dir_path, account_path, data_string);
+
+        if (options.old_name) {
```
shirady, line 552: old_name? in the CLI we used curr_name - both of them sound confusing.

shirady (commented):
Adding comments on the file config_fs

src/test/system_tests/ceph_s3_tests/test_ceph_s3_constants.js
```diff
@@ -1,40 +1,58 @@
...
+
 const CEPH_TEST = {
     test_dir: 'src/test/system_tests/ceph_s3_tests/',
     s3_test_dir: 's3-tests/',
     ceph_config: 'test_ceph_s3_config.conf',
     tox_config: 'tox.ini',
     new_account_params: {
-        name: 'cephalt',
+        name: cephalt_name,
         email: 'ceph.alt@noobaa.com',
```
shirady, line 18: You can reuse the new variable `cephalt_name` in the email property.

src/test/system_tests/ceph_s3_tests/test_ceph_s3_constants.js
```diff
@@ -1,40 +1,58 @@
...
+    nc_cephtenant_account_params: {
+        name: cephtenant_name,
+        uid: 2000,
+        gid: 2000,
+        new_buckets_path: FS_ROOT_2
+    },
+    nc_anonymous_account_params: {
+        anonymous: true,
+        uid: process.getuid(),
+        gid: process.getgid()
```
shirady, line 43: We can use constant uid and gid? (3000 for example)

src/test/system_tests/test_utils.js
```diff
@@ -422,6 +425,67 @@ function get_new_buckets_path_by_test_env(new_buckets_full_path, new_buckets_dir
...
+// /**
+//  * create_fresh_config_sub_dirs_manual deletes and creates config dir subdirectories manualy 
+//  * @param {string} config_root 
+//  * @returns {Promise<Void>}
+//  */
+// async function create_fresh_config_sub_dirs_manual(config_root) {
+//     await P.all(_.map([CONFIG_SUBDIRS.IDENTITIES, CONFIG_SUBDIRS.ACCOUNTS_BY_NAME,
+//         CONFIG_SUBDIRS.BUCKETS, CONFIG_SUBDIRS.ACCESS_KEYS], async dir =>
+//             fs_utils.create_fresh_path(`${config_root}/${dir}`)));
+// }
```
shirady, line 437: Why is this function in a comment?

src/test/system_tests/test_utils.js
```diff
@@ -422,6 +425,67 @@ function get_new_buckets_path_by_test_env(new_buckets_full_path, new_buckets_dir
...
+
+/**
+ * write_manual_config_file writes config file directly to the file system without using config FS
```
shirady, line 440: Could you please add in the description in which cases we would use it instead of ConfigFS?
(for example - creating invalid JSON...)

src/test/unit_tests/jest_tests/test_accountspace_fs.test.js
```diff
@@ -665,12 +661,13 @@ describe('Accountspace_FS tests', () => {
...
                 expect(res.arn).toBeDefined();
-                const user_account_config_file = await read_config_file(accountspace_fs.config_fs.accounts_dir_path, params.new_username);
+                const user_account_config_file = await accountspace_fs.config_fs.get_account_by_name(params.new_username,
+                    config_fs_account_options);
                 expect(user_account_config_file.name).toBe(params.new_username);
-                const symlink_config_path = path.join(accountspace_fs.config_fs.access_keys_dir_path, access_key + SYMLINK_SUFFIX);
-                await fs_utils.file_must_exist(symlink_config_path);
-                const user_account_config_file_from_symlink = await read_config_file(
-                    accountspace_fs.config_fs.access_keys_dir_path, access_key, true);
+                const is_path_exists = await accountspace_fs.config_fs.is_account_exists_by_access_key(access_key);
```
shirady, line 667: I refer `.toBe(true)` (since I'm expecting explicit `true` value).

src/test/unit_tests/jest_tests/test_config_dir_structure.js
```diff
@@ -0,0 +1,136 @@
...
+
+const buckets_dir_name = '/buckets/';
+const identities_dir_name = '/identities/';
+const accounts_by_name = '/accounts_by_name/';
+const access_keys_dir_name = '/access_keys/';
+const old_accounts_dir_name = '/accounts/';
+
+// WARNING:
+// The following test file will check the directory structure created using create_config_dirs_if_missing()
+// which is called when using noobaa-cli, this will
```
shirady, line 34: I didn't understand the WARNING, maybe it is only part of the text?

src/test/unit_tests/jest_tests/test_config_dir_structure.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    // old accounts directory does not exist - (5.17)
+    const path_to_check = path.join(config_dir_path, old_accounts_dir_name);
+    path_exists = await is_path_exists(fs_context, path_to_check);
+    expect(path_exists).toBeFalsy();
+}
+
+
+async function clean_config_dir() {
```
shirady, line 113: Could you add JSDoc for the new functions in this file?

src/test/unit_tests/jest_tests/test_nc_account_invalid_mkm_integration.test.js
```diff
@@ -270,8 +267,6 @@ function fail(reason) {
 }
 
 async function setup_nc_system_and_first_account() {
-    await P.all(_.map([CONFIG_SUBDIRS.ACCOUNTS, CONFIG_SUBDIRS.ACCESS_KEYS], async dir =>
-        fs_utils.create_fresh_path(`${config_root}/${dir}`)));
```
shirady, line 274: What is replacing this?

shirady (commented):
Adding additional comments.

shirady:
@romayalon, a question - after changing the file name to `identity.json` I thought we would also have `identity_type` in the JSONs, did I miss something?

src/sdk/config_fs.js
shirady: How will we know if it is an error that was actually from the FS or that we created? (I'm thinking in the future if we see a bug, and someone gives us partial information that he saw "ENOENT" error...).
I would like to have separation and if needed we can expand the mapping to include this error.

src/sdk/config_fs.js
romayalon: 1. I would just return false, do you see any logic that is missing?
2. Added
3. I'm not sure IAM users will use account functions but I added anyway, please check that this is what you meant

src/sdk/config_fs.js
romayalon: I did that initially with for.. of but I was asked on the meeting with you and Guy to use filter and map...
If you remember Guy said that running this array is a lot shorter than all those FS ops we do... anyway, i'll try to enhance it a bit but the provided I would like to avoid creating a Set() if I don't need to, and also I need to go over the old accounts as well and add them to the set..

src/sdk/config_fs.js
romayalon: do you have a better suggestion?

src/cmd/manage_nsfs.js
```diff
@@ -465,18 +465,17 @@ async function update_account(data, is_flag_iam_operate_on_root_account) {
...
     nsfs_schema_utils.validate_account_schema(parsed_data);
-    if (update_name) {
-        await config_fs.create_account_config_file(new_name, parsed_data, true, [cur_access_key]);
-        await config_fs.delete_account_config_file(cur_name, data.access_keys);
-    } else if (update_access_key) {
-        await config_fs.update_account_config_file(cur_name, parsed_data, parsed_data.access_keys, [cur_access_key]);
-    }
+    await config_fs.update_account_config_file(parsed_data, {
+        old_name: update_name && cur_name,
+        new_access_keys_to_link: update_access_key && parsed_data.access_keys,
```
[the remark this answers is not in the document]
romayalon: Did you add a type definition for access keys object? I couldn't find any

src/cmd/manage_nsfs.js
```diff
@@ -578,19 +577,10 @@ function filter_bucket(bucket, filters) {
  * @param {object} [filters]
  */
 async function list_config_files(type, wide, show_secrets, filters = {}) {
```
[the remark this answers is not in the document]
romayalon: we need to consider it, and it requires delicate treatment, can you please add it to the small refactoring issue?

src/sdk/accountspace_fs.js
```diff
@@ -153,7 +153,7 @@ class AccountSpaceFS {
                 const requested_account_encrypted = await nc_mkm.encrypt_access_keys(requested_account);
                 const account_string = JSON.stringify(requested_account_encrypted);
                 nsfs_schema_utils.validate_account_schema(JSON.parse(account_string));
-                await this.config_fs.update_account_config_file(params.username, JSON.parse(account_string));
+                await this.config_fs.update_account_config_file(JSON.parse(account_string));
```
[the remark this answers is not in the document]
romayalon: yes, as we discussed, I already did this change as part of a follow-up PR, this is refactoring that it unrelated to the config dir restructure

src/test/system_tests/ceph_s3_tests/test_ceph_s3_constants.js
romayalon: hmm I think you have an old version of the code, it's already there

src/manage_nsfs/health.js
```diff
@@ -358,7 +358,7 @@ class NSFSHealth {
             let res;
             const storage_path = type === TYPES.BUCKET ?
                 config_data.path :
-                config_data.nsfs_account_config.new_buckets_path;
+                config_data.nsfs_account_config?.new_buckets_path;
```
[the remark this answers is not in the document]
romayalon: hmm I think you have an old version of the code, it's already there

src/test/unit_tests/jest_tests/test_nc_account_invalid_mkm_integration.test.js
romayalon: the first time the CLI will be called will create the config directory

src/sdk/config_fs.js
romayalon: On other errors we throw other errors... only on silent ENOENT we will get to this if...

src/test/system_tests/ceph_s3_tests/test_ceph_s3_constants.js
romayalon: I kept it as it was, if you feel this needs to be changed, please add it to the small refactoring issue

romayalon:
@shirady I understood from @guymguym suggestion that we won't add it to account JSON, and when it's missing it's saying the identity is account, @guymguym did I understand correctly?

src/sdk/config_fs.js
```diff
@@ -9,12 +9,37 @@ const nb_native = require('../util/nb_native');
...
+   On 5.18 - 
+   1. accounts/ will be deprecated
+   2. A new identities/ directory will be created that represents an account identity, 
+      for example -
+      when an account called alice is created (id = 111) and an IAM user called bob (id = 222) 
+      was created for alice, in the FS it will look as the folloing - 
+      2.1. config_dir/identities/111/identity.json -> represents the account properties (alice)
+      2.2. config_dir/identities/111/users/ -> represents the account's iam users
+      2.3. config_dir/identities/111/users/bob.symlink -> config_dir/identities/222/identity.json (an index to another identity)
+      2.4. config_dir/identities/222/bob.json -> represents the account properties (bob)
```
shirady, line 22: I think it should be
```suggestion
      2.4. config_dir/identities/222/identity.json -> represents the account properties (bob)     
```

src/sdk/config_fs.js
shirady: > only on silent ENOENT we will get to this if...
But I see "if (!identity && !options.silent_if_missing) {" - which means that we didn't find the identity and it is not silent, am I missing something?

src/sdk/config_fs.js
```diff
@@ -151,88 +224,200 @@ class ConfigFS {
...
-           // TODO - when users/ move to be symlink we need to use_lstat by name, id is not using lstat
-           path_to_check = this.get_account_path_by_name(identifier_object.name);
-        } else if (identifier_object.access_key) {
-            path_to_check = this.get_account_path_by_access_key(identifier_object.access_key);
-            use_lstat = true;
-       } else {
-            throw new Error(`config_fs.is_account_exists - invalid identifier provided ${identifier_object}`);
+    async is_account_exists_by_name(account_name, owner_account_id) {
+        const path_to_check = this.get_account_or_user_path_by_name(account_name, owner_account_id);
+        let res = await native_fs_utils.is_path_exists(this.fs_context, path_to_check);
```
shirady, line 382: For easier readability of the next condition, I would change the name of the variable "res" to a more meaningful name.

src/sdk/config_fs.js
shirady: It is, thanks.

src/sdk/config_fs.js
shirady: > only on silent ENOENT we will get to this if...

But I see "if (!identity && !options.silent_if_missing) {" - which means that we didn't find the identity and it is not silent, am I missing something?

src/sdk/config_fs.js
shirady: No, leave as is, thanks.

src/cmd/manage_nsfs.js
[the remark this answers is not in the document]
shirady: No, I have not created it yet (changing it to object is enough and we can improve it later).

src/cmd/manage_nsfs.js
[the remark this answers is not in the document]
shirady: Added.

src/test/system_tests/ceph_s3_tests/test_ceph_s3_constants.js
shirady: I meant:
```suggestion
        email: `${cephalt_name}@noobaa.com`,
```
but now I noticed that it is not the same... leave it as is.

src/sdk/config_fs.js
```diff
@@ -9,12 +9,37 @@ const nb_native = require('../util/nb_native');
...
-    ACCOUNTS: 'accounts',
     BUCKETS: 'buckets',
-    ACCESS_KEYS: 'access_keys'
+    ACCESS_KEYS: 'access_keys',
+    IDENTITIES: 'identities',
+    ACCOUNTS_BY_NAME: 'accounts_by_name',
+    ACCOUNTS: 'accounts', // deprecated on 5.18
 });
 
+const CONFIG_TYPES = {
```
shirady, line 38: Please add Object.freeze

src/sdk/config_fs.js
```diff
@@ -265,104 +450,217 @@ class ConfigFS {
...
+            const new_access_key_path = this.get_account_or_user_path_by_access_key(access_keys.access_key);
+            await nb_native().fs.symlink(this.fs_context, account_config_relative_path, new_access_key_path);
         }
     }
 
     /**
-     * unlink_access_key_symlink unlinks the access key from the file system
+     * unlink_access_key_index unlinks the access key from the config directory
+     * 1. get the account access_key path
+     * 2. check realpath on the account name path
```
shirady, line 637: I think it was copied
```suggestion
     * 2. check realpath on the access key path
```
Could you add in the comment why we checked that? (for example - to make sure that it belongs to an existing account)

src/sdk/config_fs.js
```diff
@@ -9,12 +9,37 @@ const nb_native = require('../util/nb_native');
...
     BUCKETS: 'buckets',
-    ACCESS_KEYS: 'access_keys'
+    ACCESS_KEYS: 'access_keys',
+    IDENTITIES: 'identities',
+    ACCOUNTS_BY_NAME: 'accounts_by_name',
+    ACCOUNTS: 'accounts', // deprecated on 5.18
 });
 
+const CONFIG_TYPES = {
+    ACCOUNT: 'account',
```
shirady, line 39: It is worth adding a comment that this ACCOUNT type is for account and user config.

src/sdk/config_fs.js
```diff
@@ -453,6 +758,46 @@ class ConfigFS {
...
+    * @returns {string | undefined}
+    */
+    _get_config_entry_name(entry, suffix) {
+        return (this._has_config_file_name_format(entry, suffix)) ?
+            path.parse(entry.name).name :
+            undefined;
+    }
+
+    /**
+    * _get_config_entries_names returns config file entry names array
```
shirady, line 787:
```suggestion
    * _get_config_entries_names returns config file names array
```

src/test/system_tests/test_utils.js
```diff
@@ -6,11 +6,13 @@ const _ = require('lodash');
 const http = require('http');
 const P = require('../../util/promise');
 const os_utils = require('../../util/os_utils');
+const nb_native = require('../../util/nb_native');
 const native_fs_utils = require('../../util/native_fs_utils');
 const config = require('../../../config');
 const { S3 } = require('@aws-sdk/client-s3');
 const { NodeHttpHandler } = require("@smithy/node-http-handler");
 const path = require('path');
+const { TYPES } = require('../../manage_nsfs/manage_nsfs_constants');
```
shirady, line 15: I think it should be changed to `CONFIG_TYPES` (because the functions are abut the config file) and not add an import from the CLI.

src/test/unit_tests/jest_tests/test_config_fs_backward_compatibility.test.js
```diff
@@ -0,0 +1,177 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+
+const path = require('path');
+const fs_utils = require('../../../util/fs_utils');
+const { ConfigFS } = require('../../../sdk/config_fs');
+const { TYPES } = require('../../../manage_nsfs/manage_nsfs_constants');
```
shirady, line 7: I think it should be changed to CONFIG_TYPES.

src/sdk/accountspace_fs.js
```diff
@@ -620,7 +615,7 @@ class AccountSpaceFS {
     }
 
     async _check_username_already_exists(action, username) {
-        const name_exists = await this.config_fs.is_account_exists({ name: username });
+        const name_exists = await this.config_fs.is_account_exists_by_name(username, undefined);
```
shirady, line 618: `const name_exists = await this.config_fs.is_account_exists_by_name(username, undefined);`
`const is_user_account_exists = await this.config_fs.is_account_exists_by_name(username);`
Why sometimes do I see in this file a call with a second argument with undefined and sometimes no?

src/sdk/config_fs.js
```diff
@@ -151,88 +224,200 @@ class ConfigFS {
...
-       let use_lstat = false;
-       if (identifier_object.name) {
-           // TODO - when users/ move to be symlink we need to use_lstat by name, id is not using lstat
-           path_to_check = this.get_account_path_by_name(identifier_object.name);
-        } else if (identifier_object.access_key) {
-            path_to_check = this.get_account_path_by_access_key(identifier_object.access_key);
-            use_lstat = true;
-       } else {
-            throw new Error(`config_fs.is_account_exists - invalid identifier provided ${identifier_object}`);
+    async is_account_exists_by_name(account_name, owner_account_id) {
```
shirady, line 380: Could you please explain what the decision was about the last check that was before? (as I understand it we has lstat true that checked the symlink and not the target where it was pointed to).

src/sdk/config_fs.js
```diff
@@ -265,104 +450,217 @@ class ConfigFS {
...
+     * @returns {Promise<void>}
+     */
+    async delete_account_config_file(data) {
+        const { _id, name, access_keys = [] } = data;
+        const account_id_config_path = this.get_identity_path_by_id(_id);
+        const account_dir_path = this.get_identity_dir_path_by_id(_id);
+
+        await this.unlink_access_keys_indexes(access_keys, account_id_config_path);
+        await this.unlink_account_name_index(name, account_id_config_path);
+        await native_fs_utils.delete_config_file(this.fs_context, account_dir_path, account_id_config_path);
```
shirady, line 577: If the file somehow doesn't exist will it throw an error? (because we don't want that and want to continue to the directory deletion).

shirady:
> @shirady I understood from @guymguym suggestion that we won't add it to account JSON, and when it's missing it's saying the identity is account, @guymguym did I understand correctly?

@romayalon I understand that we need to add it, but for the existing config files we will not have to add it because a missing property there means account.
Anyway, it is something that we will add when we will expand the identities (role, policy, etc.).

src/sdk/config_fs.js
romayalon: exactly what you said

src/sdk/config_fs.js
```diff
@@ -645,7 +653,15 @@ class ConfigFS {
         const access_key_path = this.get_account_or_user_path_by_access_key(access_key);
         const full_path = await nb_native().fs.realpath(this.fs_context, access_key_path);
         if (full_path === account_id_config_path) {
-            await nb_native().fs.unlink(this.fs_context, access_key_path);
+            try {
+                await nb_native().fs.unlink(this.fs_context, access_key_path);
+            } catch (err) {
+                if (err.code === 'ENOENT') {
+                    dbg.warn(`config_fs.unlink_access_key_index: account access)keys already unlinked ${access_key} ${account_id_config_path}`);
```
shirady, line 661: "access)keys" typo?

src/sdk/config_fs.js
```diff
@@ -679,6 +680,19 @@ class ConfigFS {
...
+     * _is_symlink_pointing_to_identity checks if the index symlink (name/access_key)
+     * is pointing to the identity file path
+     * @param {string} symlink_path 
+     * @param {string} identity_path 
+     * @returns {Promise<Boolean>}
+     */
+    async _is_symlink_pointing_to_identity(symlink_path, identity_path) {
+        const full_path = await nb_native().fs.realpath(this.fs_context, symlink_path);
+        return (full_path === identity_path ||
+            (os_utils.IS_MAC && full_path === path.join('/private/', identity_path)));
```
shirady, line 693: the /private/ is only for /tmp case?

src/util/native_fs_utils.js
```diff
@@ -336,8 +336,15 @@ async function delete_config_file(fs_context, schema_dir, config_path) {
...
         const tmp_dir_path = path.join(schema_dir, get_config_files_tmpdir());
         // TODO: add retry? should we fail deletion if the config file was updated at the same time?
-        await safe_unlink(fs_context, config_path, stat, gpfs_options, tmp_dir_path);
-
+        try {
+            await safe_unlink(fs_context, config_path, stat, gpfs_options, tmp_dir_path);
+        } catch (err) {
+            if (err.code === 'ENOENT') {
+                dbg.warn(`delete_config_file: config file already deleted ${config_path}`);
+                return;
```
shirady, line 344: Why doesn't it include the `silent_if_missing` condition? (like we have in `folder_delete`).

shirady (approved):
LGTM

src/util/native_fs_utils.js
romayalon: as we discussed, we always want on this case to not fail if the config file is missing

shirady:
@romayalon 
2 things that we might want to discuss in the future:
1) Do we want to add a property the accounts schema that will define the identity type for the account ([link](https://github.com/noobaa/noobaa-core/pull/8279#issuecomment-2295884577)) - we would probably need it when we will plan to add additional property.
2) In `get_identity_by_id` adding manually the error code `ENOENT` ([link](https://github.com/noobaa/noobaa-core/pull/8279#discussion_r1718314840)) - we need to examine it during the use if it creates somehow any confusion.

PR #8281, opened 2024-08-13
noobaa/core: Modify debug module to chnge log color

aspandey:
We need to provide an options so that a user can disable/enable color of debug logs generated by core, endpoint and backing store.

### Fixes:
- Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2297435

### DIFF
--- a/config.js
+++ b/config.js
@@ -480,6 +480,8 @@ config.EVENT_LEVEL = 5;
 config.LOG_TO_STDERR_ENABLED = true;
 config.LOG_TO_SYSLOG_ENABLED = false;
 
+config.LOG_COLOR_ENABLED = process.env.NOOBAA_LOG_COLOR ? process.env.NOOBAA_LOG_COLOR === 'true' : true;
+
 // TEST Mode
 config.test_mode = false;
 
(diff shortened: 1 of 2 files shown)


config.js
```diff
@@ -480,6 +480,8 @@ config.EVENT_LEVEL = 5;
 config.LOG_TO_STDERR_ENABLED = true;
 config.LOG_TO_SYSLOG_ENABLED = false;
 
+config.LOG_COLOR_ENABLED = true;
```
liranmauda, line 483: I did not see any use of this config, Do we need it?

src/util/debug_module.js
```diff
@@ -289,7 +290,45 @@ class InternalDebugLogger {
         return tmp_mod.__level;
     }
 
+    message_format_without_color(level, args) {
```
liranmauda, line 293: I do not think we need 2 functions, we can set the `level_color` in the orig function to be an empty string, and check if we want colors or not.

liranmauda (changes requested):

config.js
aspandey: done

src/util/debug_module.js
aspandey: done

liranmauda (approved):

PR #8286, opened 2024-08-14
Fixing bucket logging after refactor

jackyalbo:
### Explain the changes
1. Adapting the changes to work with containerized and bg_workers and not only NC, bucket_to_owner_keys functions should only get one variable.
2. Adding the bucket_logging test that was left out probably by mistake
3. Adding a second test checking the feature e2e for NC - test_nc_bucket_logging

1. Fixed ODF from not uploading the logs to the target buckets

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added


src/manage_nsfs/manage_nsfs_logging.js
```diff
@@ -10,8 +10,10 @@ const { export_logs_to_target } = require('../util/bucket_logs_utils');
 const http_utils = require('../util/http_utils');
 const AWS = require('aws-sdk');
 
+let config_fs;
 // This command goes over the logs in the persistent log and move the entries to log objects in the target buckets 
-async function export_bucket_logging(config_fs) {
+async function export_bucket_logging(global_config_fs) {
```
romayalon, line 15: we were asked to avoid giving variable names a "global_" prefix

src/util/bucket_logs_utils.js
```diff
@@ -50,17 +50,16 @@ async function export_logs_to_target(fs_context, s3_connection, bucket_to_owner_
...
  * @param {function} bucket_to_owner_keys_func
  * @returns {Promise<Boolean>}
  */
-async function _upload_to_targets(config_fs, s3_connection, log_file, bucket_to_owner_keys_func) {
+async function _upload_to_targets(fs_context, s3_connection, log_file, bucket_to_owner_keys_func) {
     const bucket_streams = {};
     const promises = [];
     try {
-        const fs_context = config_fs.fs_context;
         const file = new LogFile(fs_context, log_file);
```
romayalon, line 63: why not just pass config_fs.fs_context? will you need to do other configuration file usage in the future?

jackyalbo: file is shared between containerized and none - as we don't have config_fs there. I need to use only fs_context.

romayalon (approved):

Issue #8080, opened 2024-05-27
Account update with account name is not updating bucket_name.json file automatically


anandhu-karattu:
###Environment info

NooBaa Version: noobaa-core-5.15.3-20240514.el9.x86_64
Platform: Standalone Noobaa

### Actual behavior
1.Account update with account name is not updating bucket_name.json file automatically. 
I/O is failing due to this issue.

### Expected behavior
If the account name is updated with a new name, the corresponding buckets should get auto updated with the new account name.
Otherwise it is a tedious task to manually update all the buckets with owner details

### Steps to reproduce
1. Create an account with noobaa-cli
```
[root@3node-x-worker1 ~]# noobaa-cli account add --name account_test_old --uid 7000 --gid 7000 --new_buckets_path /mnt/fs1/account_test_old 2>/dev/null
{
  "response": {
    "code": "AccountCreated",
    "reply": {
      "_id": "66544e9f738d7528d7f9e70b",
      "name": "account_test_old",
      "email": "account_test_old",
      "creation_date": "2024-05-27T09:13:03.211Z",
      "access_keys": [
        {
          "access_key": "BjLLk92yXAEFxOKaOFyA",
          "secret_key": "xYPDVxbTaH6g8b42T2ecOk/j3I53spnA/JkDtEi2"
        }
      ],
      "nsfs_account_config": {
        "uid": 7000,
        "gid": 7000,
        "new_buckets_path": "/mnt/fs1/account_test_old",
        "fs_backend": "GPFS"
      },
      "allow_bucket_creation": true,
      "master_key_id": "664e107282dba0d07d51b455"
    }
  }
}
```

2. Create a bucket with noobaa-cli.

```
[root@3node-x-worker1 ~]# noobaa-cli bucket add --name bucket-old --owner account_test_old --path /mnt/fs1/account_test_old/bucket_old 2>/dev/null
{
  "response": {
    "code": "BucketCreated",
    "reply": {
      "_id": "66544f33b1260396610d32cc",
      "name": "bucket-old",
      "owner_account": "66544e9f738d7528d7f9e70b",
      "system_owner": "account_test_old",
      "bucket_owner": "account_test_old",
      "versioning": "DISABLED",
      "creation_date": "2024-05-27T09:15:31.368Z",
      "path": "/mnt/fs1/account_test_old/bucket_old",
      "should_create_underlying_storage": false,
      "fs_backend": "GPFS"
    }
  }
}
```

[...]

PR #8289, opened 2024-08-18
NC | Bucket Owner Removal

romayalon:
### Disclaimers - 
1. This PR is based on the bucket owner removal part of https://github.com/noobaa/noobaa-core/pull/8167 written by @alphaprinz.
2. ~~This PR is dependent on merging - https://github.com/noobaa/noobaa-core/pull/8279 and currently, cherry-picking its only commit, which will be removed after its merge.~~

### Explain the changes
1. Schema - Removal of bucket_owner property requirement from nsfs_bucket_schema.js. Didn't remove it completely from the schema for backward compatibility.
2. CLI - Create/Update flows - removal of bucket_owner property usage from the schema, but left it on the CLI API which means that on create/update bucket `--owner` flag is still the name of the account, and it'll be mapped to an id internally - owner_account.
3. S3 API (BucketspaceFS) - 
- read_bucket_sdk_info() - since we don't keep the bucket owner name we now read the account config file by its id (owner account) and set the bucket_owner run time object.
- create_bucket() - removed bucket_owner from new_bucket_defaults.
4. IAM API (AccountspaceFS) - 
- _check_if_root_account_does_not_have_buckets_before_deletion() - Updated usage by id instead of by name.  
6. Tests/Docs - Removed and updated bucket_owner related changes.

1. Fixed #https://github.com/noobaa/noobaa-core/issues/8080
2. Gap - we need to remove [the following line](https://github.com/noobaa/noobaa-core/pull/8167/files#diff-11cc61806e3f6607fe678e081751019f21db0f09d48a56a26713203b3ada1db5L241) when having IAM users in users/ directory (this will mean that a root account can have the same name as an IAM user) @shirady 

```diff
-   if (account_identifier === bucket_owner.unwrap()) return true;
+   if (owner_account && account._id === owner_account.id) return true;
+   //name check - only for root accounts
+   if (!account.owner && account_identifier_name === bucket_owner.unwrap()) return true;
```
### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -21,7 +21,7 @@ const noobaa_cli_diagnose = require('../manage_nsfs/diagnose');
 const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
 const { print_usage } = require('../manage_nsfs/manage_nsfs_help_utils');
 const { TYPES, ACTIONS, LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS, GLACIER_ACTIONS } = require('../manage_nsfs/manage_nsfs_constants');
-const { throw_cli_error, write_stdout_response, get_boolean_or_string_value, has_access_keys, set_debug_level } = require('../manage_nsfs/manage_nsfs_cli_utils');
+const { throw_cli_error, get_bucket_owner_account, write_stdout_response, get_boolean_or_string_value, has_access_keys, set_debug_level } = require('../manage_nsfs/manage_nsfs_cli_utils');
 const manage_nsfs_validations = require('../manage_nsfs/manage_nsfs_validations');
 const nc_mkm = require('../manage_nsfs/nc_master_key_manager').get_instance();
 
@@ -97,7 +97,6 @@ async function fetch_bucket_data(action, user_input) {
         _id: undefined,
         name: user_input.name === undefined ? undefined : String(user_input.name),
         owner_account: undefined,
-        bucket_owner: user_input.owner,
         tag: undefined, // if we would add the option to tag a bucket using CLI, this should be changed
         versioning: action === ACTIONS.ADD ? 'DISABLED' : undefined,
         creation_date: action === ACTIONS.ADD ? new Date().toISOString() : undefined,
@@ -124,6 +123,13 @@ async function fetch_bucket_data(action, user_input) {
         // @ts-ignore
         data = _.omitBy(data, _.isUndefined);
         data = await fetch_existing_bucket_data(data);
+    }
+
+    //if we're updating the owner, needs to override owner in file with the owner from user input.
+    //if we're adding a bucket, need to set its owner id field
+    if ((action === ACTIONS.UPDATE && user_input.owner) || (action === ACTIONS.ADD)) {
+        const account = await get_bucket_owner_account(config_fs, String(user_input.owner));
+        data.owner_account = account._id;
     }
 
     // override values
@@ -159,15 +165,17 @@ async function add_bucket(data) {
     // for validating against the schema we need an object, hence we parse it back to object
     nsfs_schema_utils.validate_bucket_schema(JSON.parse(data_json));
     await config_fs.create_bucket_config_file(data.name, data_json);
-    write_stdout_response(ManageCLIResponse.BucketCreated, data_json, { bucket: data.name });
+    await set_bucker_owner(data);
+    write_stdout_response(ManageCLIResponse.BucketCreated, data, { bucket: data.name });
 }
 
 async function get_bucket_status(data) {
     await manage_nsfs_validations.validate_bucket_args(config_fs, data, ACTIONS.STATUS);
 
     try {
-        const config_data = await config_fs.get_bucket_by_name(data.name);
-        write_stdout_response(ManageCLIResponse.BucketStatus, config_data);
+        const bucket_data = await config_fs.get_bucket_by_name(data.name);
+        await set_bucker_owner(bucket_data);
+        write_stdout_response(ManageCLIResponse.BucketStatus, bucket_data);
     } catch (err) {
         const err_code = err.code === 'EACCES' ? ManageCLIError.AccessDenied : ManageCLIError.NoSuchBucket;
         throw_cli_error(err_code, data.name);
@@ -185,9 +193,11 @@ async function update_bucket(data) {
         // We take an object that was stringify
         // (it unwraps ths sensitive strings, creation_date to string and removes undefined parameters)
         // for validating against the schema we need an object, hence we parse it back to object
-        nsfs_schema_utils.validate_bucket_schema(JSON.parse(data));
+        const parsed_data = JSON.parse(data);
+        nsfs_schema_utils.validate_bucket_schema(parsed_data);
         await config_fs.update_bucket_config_file(cur_name, data);
-        write_stdout_response(ManageCLIResponse.BucketUpdated, data);
+        await set_bucker_owner(parsed_data);
+        write_stdout_response(ManageCLIResponse.BucketUpdated, parsed_data);
         return;
     }
 
@@ -203,6 +213,8 @@ async function update_bucket(data) {
     nsfs_schema_utils.validate_bucket_schema(JSON.parse(data));
     await config_fs.create_bucket_config_file(new_name, data);
     await config_fs.delete_bucket_config_file(cur_name);
+    data = JSON.parse(data);
+    await set_bucker_owner(data);
     write_stdout_response(ManageCLIResponse.BucketUpdated, data);
 }
 
@@ -577,7 +589,7 @@ function filter_bucket(bucket, filters) {
  * @param {object} [filters]
  */
 async function list_config_files(type, wide, show_secrets, filters = {}) {
-    let entries = [];
+    let entry_names = [];
     const should_filter = Object.keys(filters).length > 0;
     const is_filter_by_name = filters.name !== undefined;
 
@@ -592,24 +604,35 @@ async function list_config_files(type, wide, show_secrets, filters = {}) {
     // in case we have a filter by name, we don't need to read all the entries and iterate them
     // instead we "mock" the entries array to have one entry and it is the name by the filter (we add it for performance)
     if (is_filter_by_name) {
-        entries = [filters.name];
+        entry_names = [filters.name];
     } else if (type === TYPES.ACCOUNT) {
-        entries = await config_fs.list_accounts();
+        entry_names = await config_fs.list_accounts();
     } else if (type === TYPES.BUCKET) {
-        entries = await config_fs.list_buckets();
+        entry_names = await config_fs.list_buckets();
     }
 
-    let config_files_list = await P.map_with_concurrency(10, entries, async entry => {
+    // temporary cache for mapping bucker owner_account (id) -> bucket_owner (name)
+    const bucket_owners_map = {};
+    let config_files_list = await P.map_with_concurrency(10, entry_names, async entry_name => {
         if (wide || should_filter) {
             const data = type === TYPES.ACCOUNT ?
-                await config_fs.get_account_by_name(entry, options) :
-                await config_fs.get_bucket_by_name(entry, options);
+                await config_fs.get_account_by_name(entry_name, options) :
+                await config_fs.get_bucket_by_name(entry_name, options);
             if (!data) return undefined;
             if (should_filter && !filter_list_item(type, data, filters)) return undefined;
             // remove secrets on !show_secrets && should filter
-            return wide ? _.omit(data, show_secrets ? [] : ['access_keys']) : { name: entry };
+            if (!wide) return { name: entry_name };
+            if (type === TYPES.ACCOUNT) return _.omit(data, show_secrets ? [] : ['access_keys']);
+            if (type === TYPES.BUCKET) {
+                data.bucket_owner = bucket_owners_map[data.owner_account];
+                if (!data.bucket_owner) {
+                    await set_bucker_owner(data);
+                    bucket_owners_map[data.owner_account] = data.bucket_owner;
+                }
+                return data;
+            }
         } else {
-            return { name: entry };
+            return { name: entry_name };
         }
     });
     // it inserts undefined for the entry '.noobaa-config-nsfs' and we wish to remove it
@@ -644,6 +667,15 @@ function get_access_keys(action, user_input) {
         access_keys[0].access_key = undefined; //Setting it as undefined so we can replace the symlink
     }
     return { access_keys, new_access_key };
+}
+
+/**
+ * set_bucker_owner gets bucket owner from cache by its id and sets bucket_owner name on the bucket data
+ * @param {object} bucket_data 
+ */
+async function set_bucker_owner(bucket_data) {
+    const account_data = await config_fs.get_identity_by_id(bucket_data.owner_account, TYPES.ACCOUNT, { silent_if_missing: true});
+    bucket_data.bucket_owner = account_data?.name;
 }
 
 async function whitelist_ips_management(args) {
--- a/src/manage_nsfs/manage_nsfs_cli_utils.js
+++ b/src/manage_nsfs/manage_nsfs_cli_utils.js
@@ -37,15 +37,20 @@ function write_stdout_response(response_code, detail, event_arg) {
  * get_bucket_owner_account will return the account of the bucket_owner
  * otherwise it would throw an error
  * @param {import('../sdk/config_fs').ConfigFS} config_fs
- * @param {string} bucket_owner
+ * @param {string} [bucket_owner]
+ * @param {string} [owner_account_id]
  */
-async function get_bucket_owner_account(config_fs, bucket_owner) {
+async function get_bucket_owner_account(config_fs, bucket_owner, owner_account_id) {
     try {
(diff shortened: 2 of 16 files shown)


src/test/unit_tests/jest_tests/test_config_fs.test.js
```diff
@@ -22,9 +22,10 @@ describe('adjust_bucket_with_schema_updates', () => {
     });
 
     it('should return bucket without deprecated properties (system_owner)', () => {
```
shirady, line 24: Please update the `it`
```suggestion
    it('should return bucket without deprecated properties', () => {
```

src/sdk/bucketspace_fs.js
```diff
@@ -17,6 +17,7 @@ const nsfs_schema_utils = require('../manage_nsfs/nsfs_schema_utils');
 const nc_mkm = require('../manage_nsfs/nc_master_key_manager').get_instance();
 const { ConfigFS, JSON_SUFFIX } = require('./config_fs');
 const mongo_utils = require('../util/mongo_utils');
+const { TYPES } = require('../manage_nsfs/manage_nsfs_constants');
```
shirady, line 20: Please use `CONFIG_TYPES` instead.

src/sdk/bucketspace_fs.js
```diff
@@ -121,11 +122,22 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             };
 
             bucket.name = new SensitiveString(bucket.name);
-            bucket.bucket_owner = new SensitiveString(bucket.bucket_owner);
+            // Backward compatibility will take longer without bucket owner name
```
shirady, line 125: Why do you add this comment "will take longer without bucket owner name"?
There is no other option that we can use...

src/sdk/bucketspace_fs.js
```diff
@@ -121,11 +122,22 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+            const account_config = await this.config_fs.get_identity_by_id(
+                bucket.owner_account,
+                TYPES.ACCOUNT,
+                { silent_if_missing: true }
+            );
+
+            if (!account_config) {
+                dbg.warn(`Bucket Owner does not exist ${bucket.owner_account}`);
+            }
+            bucket.bucket_owner = new SensitiveString(account_config?.name);
```
shirady, line 135: if the `account_config` is undefined then you wrap the undefined value, is it what you want?

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
```diff
@@ -511,7 +511,8 @@ describe('manage nsfs cli bucket flow', () => {
             await set_path_permissions_and_owner(bucket_defaults.path, account_defaults2, 0o700);
             await exec_manage_cli(TYPES.BUCKET, action, bucket_options);
             const bucket = await config_fs.get_bucket_by_name(bucket_defaults.name);
-            expect(bucket.bucket_owner).toBe(account_defaults2.name);
+            const owner_data = await config_fs.get_identity_by_id(bucket.owner_account);
```
shirady, line 514: Why did you change it from "bucket_owner" to "owner_data"?

src/test/unit_tests/test_bucketspace.js
```diff
@@ -373,8 +373,8 @@ mocha.describe('bucket operations - namespace_fs', function() {
         const s3_bucket = bucket_name + '-s3';
         const res = await s3_correct_uid_default_nsr.getBucketAcl({ Bucket: s3_bucket });
         const bucket_info = await rpc_client.bucket.read_bucket({ name: s3_bucket });
-        assert.equal(res.Owner.DisplayName, bucket_info.owner_account.email);
         assert.equal(res.Owner.ID, bucket_info.owner_account.id);
+        assert.equal(res.Owner.DisplayName, bucket_info.owner_account.email);
```
shirady, line 377: Does this change have a meaning?
If not, then better restore this file.

src/manage_nsfs/manage_nsfs_cli_utils.js
```diff
@@ -38,14 +38,17 @@ function write_stdout_response(response_code, detail, event_arg) {
  * otherwise it would throw an error
  * @param {import('../sdk/config_fs').ConfigFS} config_fs
  * @param {string} bucket_owner
```
shirady, line 40: also optional:
```suggestion
 * @param {string} [bucket_owner]
```

src/cmd/manage_nsfs.js
```diff
@@ -96,8 +96,6 @@ async function fetch_bucket_data(action, user_input) {
         // added undefined values to keep the order the properties when printing the data object
         _id: undefined,
         name: user_input.name === undefined ? undefined : String(user_input.name),
-        owner_account: undefined,
```
shirady, line 99: I think we had this "owner_account" on purpose to have the property saved as undfined.

src/sdk/bucketspace_fs.js
romayalon: there is, but it might cause a mistake, I thought of using the old bucket_owner property, read the account config file and check if it has the same id...

src/sdk/bucketspace_fs.js
romayalon: initially I threw an error if the owner account config does not exist, but later on I saw it doesn't make sense to fail on that, so since we expect a sensitive string I need to wrap something.. do you have a better suggestion?

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
romayalon: because it's the owner data, bucket_owner IMO should be the bucket owner name as we use on bucketspaceFS

src/cmd/manage_nsfs.js
```diff
@@ -194,6 +196,8 @@ async function update_bucket(data) {
         // for validating against the schema we need an object, hence we parse it back to object
         nsfs_schema_utils.validate_bucket_schema(JSON.parse(data));
         await config_fs.update_bucket_config_file(cur_name, data);
+        data = JSON.parse(data);
```
shirady, line 199: You can parse once and save it in a variable.

src/cmd/manage_nsfs.js
```diff
@@ -606,17 +612,22 @@ async function list_config_files(type, wide, show_secrets, filters = {}) {
         entries = await config_fs.list_buckets();
```
shirady, line 612: I would rename it also (less confusing) to `entries_names`

src/cmd/manage_nsfs.js
```diff
@@ -653,6 +664,15 @@ function get_access_keys(action, user_input) {
     return { access_keys, new_access_key };
 }
 
+/**
+ * set_bucker_owner gets bucket owner from cache by its id and sets bucket_owner name on the bucket data
+ * @param {object} bucket_data 
+ */
+async function set_bucker_owner(bucket_data) {
```
shirady, line 671: A couple of questions related to the account_cache:
1. Now we should use the CLI with a working server in the background? (before that we didn't have to)
2. The cache it per endpoint, and could be many endpoints, how does it work?
3. What happens if the user updates the bucket owner? (`sudo node src/cmd/manage_nsfs bucket update --name <bucket name> --owner <another_account>`) will he see the previous one?
4. Why did you decide to use it in all the operations and not just in bucket list?

src/sdk/accountspace_fs.js
```diff
@@ -26,6 +27,18 @@ const { account_cache } = require('./object_sdk');
 const dummy_region = 'us-west-2';
 const dummy_service_name = 's3';
 
+const account_id_cache = new LRUCache({
```
shirady, line 30: We already had imported the `account_cache` from object sdk, what is the difference between `account_id_cache` and `account_cache`?

src/sdk/bucketspace_fs.js
shirady: I didn't understand this yet.

src/sdk/bucketspace_fs.js
shirady: No, I want to make sure this is what you planned.
I suggest to add a comment in the code about this.

src/test/unit_tests/test_nc_nsfs_cli.js
```diff
@@ -20,17 +20,17 @@ const { TYPES, ACTIONS } = require('../../manage_nsfs/manage_nsfs_constants');
 const tmp_fs_path = path.join(TMP_PATH, 'test_bucketspace_fs');
 const DEFAULT_FS_CONFIG = get_process_fs_context();
 const config_fs_account_options = { show_secrets: true, decrypt_secret_key: true };
+const root_path = path.join(tmp_fs_path, 'root_path_manage_nsfs/');
+const config_root = path.join(tmp_fs_path, 'config_root_manage_nsfs');
+const config_fs = new ConfigFS(config_root);
+// TODO: needed for NC_CORETEST FLOW - should be handled better
+const nc_coretes_location = config.NC_MASTER_KEYS_FILE_LOCATION;
```
shirady, line 27: typo in the variable name, and I would add more details in the name.
```suggestion
const nc_coretest_master_key_location = config.NC_MASTER_KEYS_FILE_LOCATION;
```

src/test/unit_tests/jest_tests/test_nc_nsfs_bucket_cli.test.js
shirady: We use the `bucket_owner` as a tern of the owner's name only, and `owner data` and the account (objects) which is the owner of the bucket?
I'm fine with it, I think a short comment can help.

shirady:
@romayalon could you please update the the PR description (and attach the issue)?

src/sdk/accountspace_fs.js
romayalon: The difference is -  
 - Regular accounts cache - access_key -> account mapping
 - new Accounts ID Cache id -> account mapping.

src/cmd/manage_nsfs.js
```diff
@@ -592,24 +604,35 @@ async function list_config_files(type, wide, show_secrets, filters = {}) {
...
             // remove secrets on !show_secrets && should filter
-            return wide ? _.omit(data, show_secrets ? [] : ['access_keys']) : { name: entry };
+            if (!wide) return { name: entry_name };
+            if (type === TYPES.ACCOUNT) return _.omit(data, show_secrets ? [] : ['access_keys']);
+            if (type === TYPES.BUCKET) {
+                data.bucket_owner = bucket_owners_map[data.owner_account];
+                if (!data.bucket_owner) {
+                    await set_bucker_owner(data);
+                    bucket_owners_map[data.owner_account] = data.bucket_owner;
+                }
```
shirady, line 631: I think that reusing the `bucket_owner` as it was saved in the bucket config will not be valid in case the account was renamed, so I'm not sure that you can really reduce the system call (to read the account config).

src/manage_nsfs/manage_nsfs_cli_utils.js
```diff
@@ -37,15 +37,20 @@ function write_stdout_response(response_code, detail, event_arg) {
  * get_bucket_owner_account will return the account of the bucket_owner
  * otherwise it would throw an error
  * @param {import('../sdk/config_fs').ConfigFS} config_fs
- * @param {string} bucket_owner
+ * @param {string} [bucket_owner]
+ * @param {string} [owner_account_id]
  */
-async function get_bucket_owner_account(config_fs, bucket_owner) {
+async function get_bucket_owner_account(config_fs, bucket_owner, owner_account_id) {
```
shirady, line 43: I thought that you are using it only with `bucket_owner` as the name and could remove the argument `owner_account_id` and simplify the function.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -316,7 +316,7 @@ async function validate_bucket_args(config_fs, data, action) {
     if (action === ACTIONS.ADD || action === ACTIONS.UPDATE) {
         if (action === ACTIONS.ADD) native_fs_utils.validate_bucket_creation({ name: data.name });
         if ((action === ACTIONS.UPDATE) && (data.new_name !== undefined)) native_fs_utils.validate_bucket_creation({ name: data.new_name });
-        if ((action === ACTIONS.ADD) && (data.bucket_owner === undefined)) throw_cli_error(ManageCLIError.MissingBucketOwnerFlag);
+        if ((action === ACTIONS.ADD) && (data.owner_account === undefined)) throw_cli_error(ManageCLIError.MissingBucketOwnerFlag);
```
shirady, line 319: I think you can remove this line, we will never get here (we failed it before in the `get_bucket_owner_account`).

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -328,30 +328,33 @@ async function validate_bucket_args(config_fs, data, action) {
         if (!exists) {
             throw_cli_error(ManageCLIError.InvalidStoragePath, data.path);
         }
-        const account = await get_bucket_owner_account(config_fs, data.bucket_owner);
-        const account_fs_context = await native_fs_utils.get_fs_context(account.nsfs_account_config, data.fs_backend);
+
+        // bucket owner validations 
+        const owner_account = await get_bucket_owner_account(config_fs, undefined, data.owner_account);
```
shirady, line 333: I think the variable name `owner_account` is a bit confusing because in the schema we have the same property name with string type (I understand that in the schema it should have been named `owner_account_id`... but we will probably not change it).

src/sdk/accountspace_fs.js
```diff
@@ -26,6 +26,7 @@ const { account_cache } = require('./object_sdk');
 const dummy_region = 'us-west-2';
 const dummy_service_name = 's3';
 
+
```
shirady, line 29: redundent newline

src/cmd/manage_nsfs.js
romayalon: we can say it also when inside the loop, we checked the first account but the account name was changed or new accounts were added... I think it's fine since it's list

src/manage_nsfs/manage_nsfs_validations.js
romayalon: I would keep it to be on the safe side, also since we know that these flows will hopefully change in the future, i'm not sure what will be the order then

src/cmd/manage_nsfs.js
shirady: You are right, but a property of `bucket_owner` in a bucket might be not updated for years/months, unlike a bucket that was created during the loop which is not updated for seconds.

For example, the account and bucket were created in version 5.15 and in version 5.16 the account was renamed (bucket config was not updated), in version 5.18 when listing the buckets I think we should expect to see the bucket owner's updated name (and not the one which was originally saved).

romayalon: I think there is a missunderstnding, i'm not using the data.bucket_owner old property, inside set_bucket_owner(data) I read the account config by id and use it

src/cmd/manage_nsfs.js
```diff
@@ -675,7 +675,7 @@ function get_access_keys(action, user_input) {
  */
 async function set_bucker_owner(bucket_data) {
     const account_data = await config_fs.get_identity_by_id(bucket_data.owner_account, TYPES.ACCOUNT, { silent_if_missing: true});
-    bucket_data.bucket_owner = account_data.name;
+    bucket_data.bucket_owner = account_data?.name;
```
shirady, line 678: I want to make sure I understand you - do you prefer to print undefined in the property `bucket_owner` than not printing it at all?

src/cmd/manage_nsfs.js
shirady: Now I saw the map and the comment, thanks.

src/cmd/manage_nsfs.js
romayalon: It removes undefined on the list, here is an example of a manual test - 
1. Create account `account1`
2. Create 11 buckets owned by `account1`
3. List buckets - see `bucket_owner: account1`
4. Delete manually on the FS `account1` identity
5. List buckets again - see `bucket_owner` is missing
```
"response": {
    "code": "BucketList",
    "reply": [
      {
        "_id": "66c59b3c40ae366b6414ffc5",
        "name": "bucket1",
        "owner_account": "66c59b21adca3d4be6b9f556",
        "versioning": "DISABLED",
        "creation_date": "2024-08-21T07:46:04.155Z",
        "path": "/private/tmp/dir2",
        "should_create_underlying_storage": false
      },
      {
        "_id": "66c59b579d6b1c9c554db214",
        "name": "bucket6",
        "owner_account": "66c59b21adca3d4be6b9f556",
        "versioning": "DISABLED",
        "creation_date": "2024-08-21T07:46:31.907Z",
        "path": "/private/tmp/dir2",
        "should_create_underlying_storage": false
      },
      {
    .....
      }
    ]
```

shirady: Thank you, I checked the flow in the code for this case:
list_config_files-> inside the bucket loop -> in the set_bucker_owner the bucker_owner is undefined -> in write_stdout_response inside it the `to_string` of list we will JSON.parse and JSON.stringify and this is what eliminates the `undefined` value as I understand it.

shirady (approved):
LGTM

PR #8299, opened 2024-08-20
NSFS | NC | versioning - don't delete bucket with delete marker on top of versioned-object

nadavMiz:
### Explain the changes
1. on versioned mode check if bucket is empty using list-object-versions instead of list-objects

1. 

### Testing Instructions:
1. sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_bucketspace.js

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -341,7 +341,11 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
                     // 2. delete only bucket tmpdir
                     let list;
                     try {
-                        list = await ns.list_objects({ ...params, bucket: name, limit: 1 }, object_sdk);
+                        if (ns._is_versioning_disabled()) {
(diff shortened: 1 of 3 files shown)


src/sdk/bucketspace_fs.js
```diff
@@ -337,7 +337,11 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
                     // 2. delete only bucket tmpdir
                     let list;
                     try {
-                        list = await ns.list_objects({ ...params, bucket: name, limit: 1 }, object_sdk);
+                        if (bucket.versioning === 'DISABLED') {
```
romayalon, line 340:
```suggestion
                        if (ns._is_versioning_disabled()) {
```

romayalon (approved):
LGTM, just a minor suggestion

PR #8304, opened 2024-08-22
NC | Accounts ID cache addition

romayalon:
### Explain the changes
1. A follow-up PR to https://github.com/noobaa/noobaa-core/pull/8289.
2. Added in accountspaceFS accounts id cache (mapped to account config)
3. Changed the calls from get_account_by_idnetity() to use the cache.

1. 

### Testing Instructions:

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/manage_nsfs/manage_nsfs_validations.js
+++ b/src/manage_nsfs/manage_nsfs_validations.js
@@ -9,11 +9,12 @@ const string_utils = require('../util/string_utils');
 const native_fs_utils = require('../util/native_fs_utils');
 const ManageCLIError = require('../manage_nsfs/manage_nsfs_cli_errors').ManageCLIError;
 const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
-const { throw_cli_error, get_bucket_owner_account, get_options_from_file, get_boolean_or_string_value,
-    check_root_account_owns_user, is_name_update, is_access_key_update } = require('../manage_nsfs/manage_nsfs_cli_utils');
+const { throw_cli_error, get_options_from_file, get_boolean_or_string_value, get_bucket_owner_account_by_id,
+    is_name_update, is_access_key_update } = require('../manage_nsfs/manage_nsfs_cli_utils');
 const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, FROM_FILE, BOOLEAN_STRING_VALUES, BOOLEAN_STRING_OPTIONS,
     GLACIER_ACTIONS, LIST_UNSETABLE_OPTIONS, ANONYMOUS, DIAGNOSE_ACTIONS, UPGRADE_ACTIONS } = require('../manage_nsfs/manage_nsfs_constants');
 const iam_utils = require('../endpoint/iam/iam_utils');
+const { check_root_account_owns_user } = require('../nc/nc_utils');
 
 /////////////////////////////
 //// GENERAL VALIDATIONS ////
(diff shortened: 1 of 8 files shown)


src/sdk/accountspace_fs.js
```diff
@@ -26,6 +27,26 @@ const { account_cache } = require('./object_sdk');
 const dummy_region = 'us-west-2';
 const dummy_service_name = 's3';
 
+const account_id_cache = new LRUCache({
+    name: 'AccountIDCache',
+    // TODO: Decide on a time that we want to invalidate
+    expiry_ms: Number(config.ACCOUNTS_ID_CACHE_EXPIRY),
+    /**
+     * @param {{ _id: string, config_fs: import('./config_fs').ConfigFS, silent_if_missing?: boolean }} params
+     */
```
shirady, line 36: I don't see the param `silent_if_missing` used.
I think it will be easier to read if every param will be in a new line, for example:
```suggestion
    /**
     * @param {{ 
                   _id: string,
                   config_fs: import('./config_fs').ConfigFS,
         }} params
     */
```

shirady:
@romayalon The main open points I have in this PR are:
- The `account_id_cache` is based on `account_cache` with 2 differences: (1) expiry by 3 minutes (instead of 10 minutes by default) (2) use `get_identity_by_id` (read account by ID instead of `read_account_by_access_key` read account by access key). I remembered that @guymguym had in mind that were aspects that `bucket_namespace_cache` was better than `account_cache`.
- More general about the cache we use (both of the cache we have) - the cache is saved in memory per process, so changes won't be reflected in other processes - for example, any change to an account that will be made by the CLI will not be reflected in endpoint and they will still read the account with the previous changes - can be with S3 API and IAM API. As I understood creating this communication was discussed in the past (it was thought it was complex).

src/sdk/accountspace_fs.js
romayalon: thanks I removed it, it's a leftover of code I removed

romayalon:
@shirady @guymguym I suggest discussing it next week 
1. Bucket cache is better because we invalidate the items of the cache when doing changes.
2. Yes, this is a change @guymguym and I discussed in the past for master keys, we can schedule a meeting for considering it again but this is out of scope for this PR.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -369,7 +370,14 @@ async function validate_bucket_args(config_fs, data, action) {
         }
 
         // bucket owner account validations 
-        const owner_account_data = await get_bucket_owner_account(config_fs, undefined, data.owner_account);
+        let owner_account_data;
```
shirady, line 373: If you want you can get it to a small function (it is not one line anymore).

src/sdk/accountspace_fs.js
```diff
@@ -25,6 +26,26 @@ const { account_cache } = require('./object_sdk');
...
 
+const account_id_cache = new LRUCache({
+    name: 'AccountIDCache',
+    // TODO: Decide on a time that we want to invalidate
+    expiry_ms: Number(config.ACCOUNTS_ID_CACHE_EXPIRY),
+    /**
+     * @param {{ _id: string, config_fs: import('./config_fs').ConfigFS }} params
+     */
+    make_key: ({ _id }) => _id,
+    load: async ({ _id, config_fs }) => config_fs.get_identity_by_id(_id, CONFIG_TYPES.ACCOUNT,
```
shirady, line 37: Why did you prefer not to use the flag `silent_if_missing`?
I see it was used in many cases (and therefore it changes the code from checking it it is undefined to try-catch clause).

shirady (commented):
A few small comments

src/sdk/accountspace_fs.js
romayalon: because it'll create undefined values inside the cache

src/sdk/accountspace_fs.js
```diff
@@ -190,6 +211,7 @@ class AccountSpaceFS {
             this._check_if_requested_is_owned_by_root_account(action, requesting_account, account_to_delete);
             await this._check_if_user_does_not_have_resources_before_deletion(action, account_to_delete);
             await this.config_fs.delete_account_config_file(account_to_delete);
+            _clean_account_id_cache(requesting_account);
```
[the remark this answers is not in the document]
romayalon: we don't have requested_account in this function, I used account_to_delete

shirady (approved):
LGTM

Issue #8197, opened 2024-07-11
aws s3 rm with --recursive option does not delete all the objects from the bucket


anandhu-karattu:
### Environment info
Standalone **noobaa noobaa-core-5.15.4-20240704.el9.x86_64**
Platform RHEL 9.4

### Actual behavior
1. S3 upload is successful for a data set (mix of all type of objects, sparse files, files with long name, compressed files etc)
2. When download via "aws-alias s3 rm s3://bucket-10909 --recursive" , this still displays the left over directories and objects in the bucket. I checked from the FS that the dataset still exists there.

Attaching noobaa.log with this ticket. There is no error found during upload or download

Time stamps for upload:
Jul 11 15:33:13 —> upload start 
Jul 11 15:34:10  --> upload completed

Timestamp for download
Jul 11 15:35:02. --> delete start
Jul 11 15:35:13 --> delete completed

### Expected behavior
 --recursive option should always delete all the uploaded objects and directories

anandhu-karattu:
Attaching noobaa.log now
[Uploading noobaa.log…]()

anandhu-karattu:
To create the test data, I used an internal tool called populatefs which creates data (files and directories) with all ossible combinations.
Then tried to upload the dataset in to S3 bucket. No error during upload and download found

anandhu-karattu:
[noobaa.log](https://github.com/user-attachments/files/16179138/noobaa.log)

romayalon:
Updating that I tried @anandhu-karattu's dataset and indeed I see files are not getting deleted.
Seems to be a bug in List Objects.
During the initial investigation, I noticed that -
1. The first list object response is 1000 objects, IsTruncated:true and a continuation token. 
2. The second list objects command params contains the continuation token from bullet 1 and the response contains 0+ objects, IsTruncated:false. I also noticed that on some cases (maybe there are more issues) the key marker is inside a directory.

naveenpaul1:
@anandhu-karattu @romayalon issue is with list_object. Because of the uniqe folder structure test data have object listing are falling for data set, and it's affecting the object_delete call, where object delete use the list_object output to get the key. Trying to fix the issue in list_object()

naveenpaul1:
Waiting for review

PR #8317, opened 2024-08-29
NSFS : S3 rm with --recursive option does not delete all the objects

naveenpaul1:
### Explain the changes
1. Issue is with list_object api, list_object is failing with specific dir strucute and not returning complete items in that bucket. 
2. dir_key and marker_dir comparison will fail when there are dir with structure slimier to  "-/migrateDir.popFSDir.3803140/" and "/migrateDir.popFSDir.3803140.OldSet/" because "/" considered greater than "."  to fix this all the backslash is replaced with space. This updated marker and dir_key used only for comparison.

Dir structure like this is will fail
> /populatefs_dir
  --/migrateDir.popFSDir.3803140
  ----/sparseDir
   -----lnk.0.qqqqqqqqqqqqqq
   -----lnk.1.qqqqqqqqqqqqqq
   -----lnk.2.qqqqqqqqqqqqqq
--/migrateDir.popFSDir.3803140.OldSet
 ----/sparseDir
  -----lnk.0.qqqqqqqqqqqqqq
  -----lnk.1.qqqqqqqqqqqqqq
  -----lnk.2.qqqqqqqqqqqqqq
    

1. S3 upload is successful for a data set (mix of all type of objects, sparse files, files with long name, compressed files etc)
4.  When download via "aws-alias s3 rm s3://bucket-10909 --recursive" , this still displays the left over directories and objects in the bucket. I checked from the FS that the dataset still exists there.

https://github.com/noobaa/noobaa-core/issues/8197

### Testing Instructions:
1. run NC_CORETEST=true node  ./node_modules/.bin/mocha ./src/test/unit_tests/test_namespace_fs.js

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -257,6 +257,11 @@ function get_entry_name(e) {
     return e.name;
 }
 
+function get_entry_for_sorting(e) {
+    const check = e.key.includes('.') ? e.key.slice(e.key.indexOf(".")).includes('/') : false;
+    return check ? e.key.slice('/')[0] : e.key;
+}
+
 /**
  * @param {string} name
  * @returns {fs.Dirent}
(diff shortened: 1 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -665,14 +670,20 @@ class NamespaceFS {
                 }
                 const marker_dir = key_marker.slice(0, dir_key.length);
                 const marker_ent = key_marker.slice(dir_key.length);
+                // dir_key and marker_dir comparison will fail when there are folder with structure slimier to 
+                // "dir_prfix.12345/" and "dir_prfix.12345.old/" because "/" considered greater than "."  to fix this 
+                // all the backslash is replaced with space. This updated marker and dir_key used only for comparison.
+                const updated_marker_dir = marker_dir.replaceAll('/', ' ');
```
romayalon, line 676: I think that instead of replacing the value of / we need to have a better comparison function, instead of comparing dir_key < marker_dir and so on, we will call that comparison function. WDYT?

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -138,21 +138,23 @@ mocha.describe('namespace_fs', function() {
         test_ns_list_objects(ns_tmp, dummy_object_sdk, 'test_ns_list_objects');
 
         function assert_sorted_list(res) {
-            let prev_key = '';
+            //let prev_key = '';
```
romayalon, line 141: we should create here a test with the exact scenario found in the github issue

src/sdk/namespace_fs.js
```diff
@@ -258,6 +258,11 @@ function get_entry_name(e) {
     return e.name;
 }
 
+function get_entry_for_sorting(e) {
+    const check = e.key.includes('.') ? e.key.slice(e.key.indexOf(".")).includes('/') : false;
```
romayalon, line 262: is '.' the only special char that causing this issue?

romayalon (changes requested):

naveenpaul1:
PR : [8324](https://github.com/noobaa/noobaa-core/pull/8324) have few overlapping changes, Will update this PR once 8324 is merged,

PR #8331, opened 2024-09-05
Block delete_bucket requests by an OBC account

dannyzaken:
### Explain the changes
1. Prevent the deletion of an OBC bucket by the OBC account.
2. A regular account can delete an OBC bucket according to the access policy or via RPC

1. https://bugzilla.redhat.com/show_bug.cgi?id=2301657

### Testing Instructions:
1. Create an OBC bucket
2. use the credentials associated with the OBC (in the OBC secret) and attempt to delete the bucket via s3 API (e.g. aws-cli)
3. The expected status is `AccessDenied`
4. Attempt to delete the bucket using the admin credentials. deletion should work.
5. Deletion of the OBC should also work and delete the underlying bucket

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -230,9 +230,16 @@ async function authorize_request_policy(req) {
     const account = req.object_sdk.requesting_account;
     const account_identifier_name = req.object_sdk.nsfs_config_root ? account.name.unwrap() : account.email.unwrap();
     const account_identifier_id = req.object_sdk.nsfs_config_root ? account._id : undefined;
-    const is_system_owner = Boolean(system_owner) && system_owner.unwrap() === account_identifier_name;
+
+    // deny delete_bucket permissions from bucket_claim_owner accounts (accounts that were created by OBC from openshift\k8s)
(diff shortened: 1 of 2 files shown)


tangledbytes (approved):
Maybe we can add a small test in unit_tests/test_s3_ops to ensure no regression in future.

src/endpoint/s3/s3_rest.js
```diff
@@ -232,6 +232,15 @@ async function authorize_request_policy(req) {
...
+    // deny delete_bucket permissions from bucket_claim_owner accounts (accounts that were created by OBC from openshift\k8s)
+    // the OBC bucket can still be delete by normal accounts according to the access policy which is checked below
+    if (req.op_name === 'delete_bucket' && account.bucket_claim_owner) {
+        dbg.error(`delete bucket request by an OBC account is restricted. an attempt to delete bucket by account ${account_identifier_name}`);
+        throw new S3Error(S3Error.AccessDenied);
+    }
+
+
     // @TODO: System owner as a construct should be removed - Temporary
     if (is_system_owner) return;
```
jackyalbo, line 245: please move line 233 close to the check in line 245

jackyalbo (approved):

PR #8337, opened 2024-09-09
Bucket Notifications

alphaprinz:
### Explain the changes
Add S3 notification support.
AWS info https://docs.aws.amazon.com/AmazonS3/latest/userguide/EventNotifications.html

Limitations:
-No LifeCycle events.

1. Implement an s3 feature.

### Testing Instructions:
-Run local Kafka with 
docker run -p 9092:9092 apache/kafka-native:3.8.0
Connect cli consumer to Kafka:
./kafka/kafka_2.13-3.8.0/bin/kafka-console-consumer.sh --topic <topic> --from-beginning --bootstrap-server localhost:9092

- Create a notification either by standard s3api put-bucket-notification-configuration.
Alternatively, for nsfs, you can use nsfs_manage to add "notifications" field to bucket.

Example of an nsfs bucket with notifications:
aprinzse:noobaa-core$ cat bucket_notif_event.json 
{
  "name": "notif-both",
  "owner": "amit1",
  "path": "/home/aprinzse/work/standalone/storage_root/notif_both",
  "notifications": [
	  {"Id": "notif_event_kafka", "Events": ["s3:ObjectCreated:*"], "Connect": "/home/aprinzse/work/standalone/connect.kv"},
  	  {"Id": "notif_event_http",  "Events": ["s3:ObjectCreated:*"], "Connect": "/home/aprinzse/work/standalone/connect_http.kv"}
  ]
}

Where connect.kv for Kafka looks like-
aprinzse:noobaa-core$ cat /home/aprinzse/work/standalone/connect.kv
metadata.broker.list=localhost:9092
notification_protocol=kafka
topic=<topic name here>
name=kafka_notif

And for http:
aprinzse:noobaa-core$ cat /home/aprinzse/work/standalone/connect_http.kv 
http_object={"host": "localhost", "port": 9999, "timeout": 100}
notification_protocol=http
name=http_notif

-Run NSFS server with NOTIFICATION_LOG_DIR env, eg
NOTIFICATION_LOG_DIR=/home/aprinzse/work/standalone/notifications_log node src/cmd/nsfs

-Kicking-off notification file processor to send notification with "notification" operation (also need the NOTIFICATION_LOG_DIR env param):
NOTIFICATION_LOG_DIR=/home/aprinzse/work/standalone/notifications_log node src/cmd/manage_nsfs notification

- [ ] Doc added/updated
- [ ] Tests added - to follow in future PR

### DIFF
--- a/src/endpoint/s3/ops/s3_put_object.js
+++ b/src/endpoint/s3/ops/s3_put_object.js
@@ -34,6 +34,9 @@ async function put_object(req, res) {
 
     dbg.log0('PUT OBJECT', req.params.bucket, req.params.key,
         req.headers['x-amz-copy-source'] || '', encryption || '');
+    req.s3event = "ObjectCreated";
+    //for copy, use correct s3event_op. otherwise, just use default (req.method)
+    req.s3event_op = copy_source ? 'Copy' : undefined;
 
     const source_stream = req.chunked_content ? s3_utils.decode_chunked_upload(req) : req;
     const reply = await req.object_sdk.upload_object({
--- a/src/endpoint/s3/ops/s3_put_object.js
+++ b/src/endpoint/s3/ops/s3_put_object.js
@@ -79,6 +79,11 @@ async function put_object(req, res) {
         };
     }
     res.setHeader('ETag', `"${reply.etag}"`);
+
+    if (reply.seq) {
+        res.seq = reply.seq;
+        delete reply.seq;
+    }
 }
 
 
--- a/src/endpoint/s3/ops/s3_put_object.js
+++ b/src/endpoint/s3/ops/s3_put_object.js
@@ -34,9 +34,8 @@ async function put_object(req, res) {
 
     dbg.log0('PUT OBJECT', req.params.bucket, req.params.key,
         req.headers['x-amz-copy-source'] || '', encryption || '');
-    req.s3event = "ObjectCreated";
-    //for copy, use correct s3event_op. otherwise, just use default (req.method)
-    req.s3event_op = copy_source ? 'Copy' : undefined;
+    //for copy, use correct s3_event_method. otherwise, just use default (req.method)
+    req.s3_event_method = copy_source ? 'Copy' : undefined;
 
     const source_stream = req.chunked_content ? s3_utils.decode_chunked_upload(req) : req;
     const reply = await req.object_sdk.upload_object({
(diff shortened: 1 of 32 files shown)


src/endpoint/s3/ops/s3_put_object.js
```diff
@@ -76,6 +79,11 @@ async function put_object(req, res) {
         };
     }
     res.setHeader('ETag', `"${reply.etag}"`);
+
+    if (reply.seq) {
+        res.seq = reply.seq;
```
dannyzaken, line 84: Seems like seq is only returned in namespace_nb flows. This will not work for other namespaces

src/util/notifications_util.js
```diff
@@ -0,0 +1,379 @@
...
+    //eslint-disable-next-line
+    if (eTag && eTag.startsWith('\"') && eTag.endsWith('\"')) {
+        eTag = eTag.substring(2, eTag.length - 2);
+    }
+
+    const notif = {
+        eventVersion: '2.3',
+        eventSource: _get_system_name(req) + ':s3',
+        eventTime: new Date().toISOString,
+        eventName: req.s3event + ':' + (req.s3_event_op || req.method),
```
dannyzaken, line 312: I suggest instead of setting `req.s3event` in each op you can maintain an object to map `req.op_name` to the `s3event`. similar to `OP_NAME_TO_ACTION` in `s3_bucket_policy_utils.js`

src/endpoint/s3/ops/s3_put_object.js
alphaprinz: Right.
As we've discussed we can add a sequence to standalone, at the cost of adding a synchronized step to all s3 ops, so it's preferred for now to see if this is actually needed.

src/util/notifications_util.js
alphaprinz: Ok.

This was done to s3event, the second element in s3 event notation.
For the third element, called method in my code, it was possible in some cases to use the OP_NAME_TO_ACTION object (eg post_object_uploadId), but some cases (eg DeleteMarkerCreated) have to be differentiated manually in code.

I've also changed terminology in the filtering code to (try to) improve readability.

src/endpoint/s3/ops/s3_put_object.js
dannyzaken: But what about other namespaces (S3, Azure, FS) that are not in standalone?

dannyzaken (approved):

PR #8347, opened 2024-09-11
Prevent issue report on read_object_md if the object doesn't exist

tangledbytes:
### Explain the changes
This PR prevents issue report when an object doesn't exist in a target bucket. It is quite common and OK for clients to issue `HeadObject` against an object that doesn't exist in the bucket. This should not be counted as an issue with the namespace store. Counting it as an issue leads to it being marked as `REJECTED` in k8s.

This situation gets even stranger when we account for the fact that our `namespace_monitor` does take into account such scenarios which leads to NamespaceStore continuously switching between `OPTIMAL` and `REJECTED`.

And it gets worse when NooBaa internal replication engine is the entity issuing `HeadObject`. In such scenarios it isn't clear what is causing the NS to get into `REJECTED` state at all as the customer might not be even touching the bucket by themselves.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2277298

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -170,11 +170,17 @@ class NamespaceBlob {
         } catch (err) {
             this._translate_error_code(err);
             dbg.warn('NamespaceBlob.read_object_md:', inspect(err));
-            object_sdk.rpc_client.pool.update_issues_report({
-                namespace_resource_id: this.namespace_resource_id,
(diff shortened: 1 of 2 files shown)


jackyalbo (approved):

tangledbytes:
Tested entire flow on K8s. Merged.

PR #8351, opened 2024-09-11
NSFS | remove previous version external attribute

nadavMiz:
### Explain the changes
currently we have external attribute to denote the previous version. it is an optimization to not look for the latest version for promoting the previous version after deleting the latest version. However this is not well maintain and can be wrong on certain concurrency scenarios such as putting multiple objects with the same key at the same time. this can cause that the wrong version is being promoted. so for now remove the previous version external attribute. we should consider returning this xattr in the future if this optimization is necessary.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added
- [ ]


shirady (approved):
LGTM

shirady:
@nadavMiz could you please add in the PR description the details that you mentioned in the Weekly meeting about the motivation and the reasons for this change? (concurrency issues that it might create so we will have it documented)

romayalon (approved):

PR #8355, opened 2024-09-12
NSFS | fix copy_object issues

### LINKED ISSUE #8315: NSFS | S3 | Versioning: CopyObject after multipart upload of a versioned object leads to invalid data content
### Environment info
- NooBaa Version: **5.15.4**
- Platform: **CentOS 8, deployed using [https://github.com/IBM/StorageScaleVagrant](https://github.com/IBM/StorageScaleVagrant)**

### Actual behavior

Content of the copied object whose source was uploaded to a version-enabled bucket using multi-part upload is not identical to the source object:

```
E       AssertionError: assert 'bIMcUvjvnUme...cjoUdmaMHmNZD' == 'avLZIzDyrPsQ...cjoUdmaMHmNZD'
E         
E         - avLZIzDyrPsQTmrmiOZHrVdSZCtRFVsosTSLGjluRPBZGbXZNTKAKSWyhuegjyevNmYTqVxjrYlMllwCdCjefarPnMGBgGnnQebJjAdaIVSqUbhSkwakohDylIGMTJKgpjxJlmHEzycMbQdOvGdimZxcOFJnkBYBSvQEXYifdmBKfrgJOZXdVcvzbjsZFBOxvOQNqsAKqfUYXUBEBWzfQdGMgFBnXfiJWUrWOwLWsBaAPIAmkjzFVtHXETyLiLgZkKvjtbBkfDcFLxGBgnIeMjQBJYNzCVkzidUdaPqzaVYawKkIbRKqZsyjPSrpXrqIIECWMzhPQNkjyuEHfCSFeJPIqfBfxpPBooDkhsmDzmIMllnrYugjdonucmiOXipomGLWIshDfqhTWCZbjBXNusYadCgZWxyBpyloWHbDHGBGKFzqiFTvRemtGPKJSYwyDEFhjiJPTVeBGfbIaUcHoMDoIpSPFSkGgmOiNhHAluJAvymOtuOhvpobXaUbGoQCgLLAXBJvpBumNZJNBGsEyIoqnSxzoQlgqnZxGwZMvsGxoKkIOXlzLoAsscwzDB...
```

### Expected behavior

Content of the copied object should be identical to the source object, assertion should pass.

### Steps to reproduce

Run Ceph s3-tests `test_object_copy_versioning_multipart_upload()`:

```
S3TEST_CONF=s3tests.conf tox -- s3tests_boto3/functional/test_s3.py::test_object_copy_versioning_multipart_upload
```

See "all" logging level noobaa log attached - look at log entries at 15:34:00 and later:

[noobaa_20240828_1748.log.gz](https://github.com/user-attachments/files/16785846/noobaa_20240828_1748.log.gz)

nadavMiz:
### Explain the changes
1. change _upload_stream to call read_object_stream directly during fallback copy instead of piping the stream to passthroughStream
2. add content_type to multipart upload
3. always copy xattr during copy object
4. remove old copy for source xattr and change copy_object tests
5. enable copy_object ceph_tests

### Testing Instructions:
1. run ceph_test: s3tests_boto3/functional/test_s3.py::test_object_copy_versioning_multipart_upload 

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -483,7 +483,7 @@ class NamespaceFS {
     }
 
     /**
-     * @param {nb.ObjectSDK} object_sdk 
+     * @param {nb.ObjectSDK} object_sdk
      * @returns {nb.NativeFSContext}
      */
     prepare_fs_context(object_sdk) {
@@ -1090,7 +1090,9 @@ class NamespaceFS {
             // end the stream
             res.end();
 
-            await stream_utils.wait_finished(res, { signal: object_sdk.abort_controller.signal });
+            // in case of transform streams such as ChunkFS there is also a readable part. since we expect write stream
+            // and don't care about the readable part, set readable: false
+            await stream_utils.wait_finished(res, { readable: false, signal: object_sdk.abort_controller.signal });
             object_sdk.throw_if_aborted();
 
             dbg.log0('NamespaceFS: read_object_stream completed file', file_path, {
@@ -1209,9 +1211,7 @@ class NamespaceFS {
         }
 
         if (copy_res) {
-            if (copy_res === copy_status_enum.FALLBACK) {
-                params.copy_source.nsfs_copy_fallback();
-            } else {
+            if (copy_res !== copy_status_enum.FALLBACK) {
                 // open file after copy link/same inode should use read open mode
                 open_mode = config.NSFS_OPEN_READ_MODE;
                 if (copy_res === copy_status_enum.SAME_INODE) open_path = file_path;
@@ -1294,10 +1294,8 @@ class NamespaceFS {
         const stat = await target_file.stat(fs_context);
         this._verify_encryption(params.encryption, this._get_encryption_info(stat));
 
-        // handle xattr
-        // assign user xattr on non copy / copy with xattr_copy header provided
         const copy_xattr = params.copy_source && params.xattr_copy;
-        let fs_xattr = copy_xattr ? undefined : to_fs_xattr(params.xattr);
+        let fs_xattr = to_fs_xattr(params.xattr);
 
         // assign noobaa internal xattr - content type, md5, versioning xattr
         if (params.content_type) {
@@ -1339,7 +1337,6 @@ class NamespaceFS {
 
         // when object is a dir, xattr are set on the folder itself and the content is in .folder file
         if (is_dir_content) {
-            if (params.copy_source) fs_xattr = await this._get_copy_source_xattr(params, fs_context, fs_xattr);
             await this._assign_dir_content_to_xattr(fs_context, fs_xattr, { ...params, size: stat.size }, copy_xattr);
         }
         stat.xattr = { ...stat.xattr, ...fs_xattr };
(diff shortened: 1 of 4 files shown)


src/sdk/namespace_fs.js
```diff
@@ -1209,9 +1209,7 @@ class NamespaceFS {
         }
 
         if (copy_res) {
-            if (copy_res === copy_status_enum.FALLBACK) {
-                params.copy_source.nsfs_copy_fallback();
```
romayalon, line 1213: if this nsfs_copy_fallbacj() is not being used any more, can we delete it?

src/sdk/namespace_fs.js
```diff
@@ -1294,10 +1292,8 @@ class NamespaceFS {
         let stat = await target_file.stat(fs_context);
         this._verify_encryption(params.encryption, this._get_encryption_info(stat));
 
-        // handle xattr
-        // assign user xattr on non copy / copy with xattr_copy header provided
         const copy_xattr = params.copy_source && params.xattr_copy;
-        let fs_xattr = copy_xattr ? undefined : to_fs_xattr(params.xattr);
+        let fs_xattr = to_fs_xattr(params.xattr);
```
romayalon, line 1296: on copy xattr you are supposed to also set the new xatter together with the copied ones?

nadavMiz: yes, for example we need to set new version-id, what xattr needs to be copied should be determined by fix_copy_source_params:
`  if (params.xattr) params.xattr = _.omitBy(params.xattr, (val, name) => name.startsWith?.('noobaa-namespace'));`

src/sdk/object_sdk.js
```diff
@@ -642,7 +642,7 @@ class ObjectSDK {
             params.content_type = source_md.content_type;
         }
         try {
-            if (params.xattr) params.xattr = _.omitBy(params.xattr, (val, name) => name.startsWith('noobaa-namespace'));
+            if (params.xattr) params.xattr = _.omitBy(params.xattr, (val, name) => name.startsWith?.('noobaa-namespace'));
```
romayalon, line 645: this happens when the name is SYMBOL  and not a string right?

nadavMiz: right

romayalon (approved):

src/sdk/namespace_fs.js
guymguym: right, but you meant to refer to params.xattr = source_md.xattr which does this copy in fix_copy_source_params.

but don't we have the same case to fix also in _create_empty_dir_content ?
which seems to mean that _get_copy_source_xattr() is unneeded completely?

guymguym (changes requested):

src/sdk/object_sdk.js
```diff
@@ -658,12 +658,6 @@ class ObjectSDK {
             params.copy_source.bucket = actual_source_ns.get_bucket(bucket);
             params.copy_source.obj_id = source_md.obj_id;
             params.copy_source.version_id = source_md.version_id;
-            if (source_ns instanceof NamespaceFS) {
-                params.copy_source.nsfs_copy_fallback = () => {
-                    this._populate_nsfs_copy_fallback({ source_params, source_ns, params });
```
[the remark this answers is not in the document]
nadavMiz: as discussed, added a fix for both source and destination nsfs buckets. currently option to copy from nsfs to other types of buckets is disabled and will return an error

guymguym (approved):
@nadavMiz I think it looks OK, but due to the sensitivity I still pointed two things for @romayalon and @nimrod-becker to notice, so would be good to check with them

src/sdk/object_sdk.js
```diff
@@ -684,7 +681,13 @@ class ObjectSDK {
 
             // if the source namespace is NSFS then we need to pass the read_object_stream the read_stream
             if (source_ns instanceof NamespaceFS) {
-                this._populate_nsfs_copy_fallback({ source_params, source_ns, params });
+                if (target_ns instanceof NamespaceFS) {
+                    params.source_ns = actual_source_ns;
+                    params.source_params = source_params;
+                } else {
+                    //this._populate_nsfs_copy_fallback({ source_params, source_ns, params });
+                    throw new Error('TODO fix _populate_nsfs_copy_fallback');
```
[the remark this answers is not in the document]
nadavMiz: I have opened a follow up issue #8372

src/sdk/namespace_fs.js
```diff
@@ -1373,13 +1369,6 @@ class NamespaceFS {
         return upload_info;
     }
 
-    async _get_copy_source_xattr(params, fs_context, fs_xattr) {
```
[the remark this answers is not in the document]
nadavMiz: I have reviewed it with @romayalon seems like this was added by mistake

PR #8386, opened 2024-09-19
NSFS | GPFS | fix put object over null version in suspended mode

### LINKED ISSUE #8379: NSFS | S3 | Versioning: InternalError upon PutObject over existing key in version-suspended bucket
### Environment info
- NooBaa Version: **5.17.0-20240904**
- Platform: **CentOS8, deployed using [https://github.com/IBM/StorageScaleVagrant](https://github.com/IBM/StorageScaleVagrant), Scale 5.2.2.0 240913.105913**

### Actual behavior

Second PutBucket causes InternalError.

See the following test code:

```
    def test_s3cever102():
        client = get_client()
        bucket_name = get_new_bucket(client=client)
    
        # The following operation uses another client instance -> following operations will fail
    #    check_configure_versioning_retry(bucket_name, "Enabled", "Enabled")
        _ces_safe_check_configure_versioning_retry(client, bucket_name, "Enabled", "Enabled")
    
        _ces_safe_check_configure_versioning_retry(client, bucket_name, "Suspended", "Suspended")
    
        key_name = 'obj'
        data = 'object-1'
        response = client.put_object(Bucket=bucket_name, Key=key_name, Body=data)
        version_id = None
        try:
            version_id = response['VersionId']
        except KeyError:
            version_id = None
        assert version_id == None
    
        response = client.list_object_versions(Bucket=bucket_name)
        versions = response['Versions']
        assert versions[0]['VersionId'] == "null"
    
        data = 'object-2'
>       response = client.put_object(Bucket=bucket_name, Key=key_name, Body=data)
...
E           botocore.exceptions.ClientError: An error occurred (InternalError) when calling the PutObject operation (reached max retries: 4): We encountered an internal error. Please try again.
```

### Expected behavior

PutBucket works and overwrites the previous object with the new content.

### Steps to reproduce

Execute the above test code added to Ceph s3-tests.

After the error occurred, it is not possible to delete the object using a simple `aws s3 rm s3://bucket/key` operation, these attempts also return InternalError.

Noobaa log with "all" log level attached
[noobaa_20240917_4.log.gz](https://github.com/user-attachments/files/17032654/noobaa_20240917_4.log.gz)

nadavMiz:
### Explain the changes
1. on Suspeded mode we need to delete the latest null version if a new object is created (as there can be only one object with null version). this delete fails on GPFS since we use the wrong gps_options. however since in GPFS link overrides existing file anyway, we don't need to call unlink at all in such scenario. changed so unlink will not be called to delete latest null version on Suspended mode for GPFS
2. add missing use of gpfs_options.delete_version instead of gpfs_options for deleting old null version

### Testing Instructions:
Run `test_nsfs_versioning_gpfs.js` tests on gpfs cluster
- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1430,9 +1430,11 @@ class NamespaceFS {
 
                 if (this._is_versioning_suspended()) {
                     if (latest_ver_info?.version_id_str === NULL_VERSION_ID) {
-                        dbg.log1('NamespaceFS._move_to_dest_version suspended: version ID of the latest version is null - the file will be unlinked');
-                        await native_fs_utils.safe_unlink(fs_context, latest_ver_path, latest_ver_info,
-                            gpfs_options?.delete_version, bucket_tmp_dir_path);
+                        //on GPFS safe_move overrides the latest object so no need to unlink
+                        if (!is_gpfs) {
+                            dbg.log1('NamespaceFS._move_to_dest_version suspended: version ID of the latest version is null - the file will be unlinked');
+                            await native_fs_utils.safe_unlink(fs_context, latest_ver_path, latest_ver_info, undefined, bucket_tmp_dir_path);
+                        }
                     } else {
                         // remove a version (or delete marker) with null version ID from .versions/ (if exists)
                         await this._delete_null_version_from_versions_directory(key, fs_context);
@@ -2981,9 +2983,9 @@ class NamespaceFS {
                         undefined;
                     const bucket_tmp_dir_path = this.get_bucket_tmpdir_full_path();
                     await native_fs_utils.safe_unlink(fs_context, null_versioned_path, null_versioned_path_info,
-                        gpfs_options, bucket_tmp_dir_path);
+                        gpfs_options?.delete_version, bucket_tmp_dir_path);
 
-                    if (gpfs_options) await this._close_files_gpfs(fs_context, gpfs_options.delete_version, undefined, true);
+                    if (gpfs_options) await this._close_files_gpfs(fs_context, gpfs_options?.delete_version, undefined, true);
                 }
                 break;
             } catch (err) {
--- a/src/test/unit_tests/test_nsfs_versioning_gpfs.js
+++ b/src/test/unit_tests/test_nsfs_versioning_gpfs.js
@@ -218,6 +218,16 @@ mocha.describe('namespace_fs gpfs- versioning', async function() {
         assert.ok(delete_objects_res.created_version_id !== undefined);
     });
 
+    mocha.it('Suspended mode - add object null version twice', async function() {
+        await ns_obj.set_bucket_versioning('SUSPENDED', dummy_object_sdk);
+        const key3 = 'key3';
+        const put_res = await put_object(dummy_object_sdk, ns_obj, gpfs_bucket, key3);
+        assert.equal(put_res.version_id, 'null');
+        //issue #8379, overwriting null value should not fail on GPFS
+        const put_res2 = await put_object(dummy_object_sdk, ns_obj, gpfs_bucket, key3);
+        assert.equal(put_res2.version_id, 'null');
+    });
+
 });
 
 async function put_object(dummy_object_sdk, ns, bucket, key) {


src/sdk/namespace_fs.js
```diff
@@ -1419,23 +1419,22 @@ class NamespaceFS {
...
 
-                gpfs_options = is_gpfs ?
-                    await this._open_files_gpfs(fs_context, new_ver_tmp_path, latest_ver_path, upload_file,
-                        latest_ver_info, open_mode, undefined, versioned_info) :
-                    undefined;
                 const bucket_tmp_dir_path = this.get_bucket_tmpdir_full_path();
-                dbg.log1('Namespace_fs._move_to_dest_version:', latest_ver_info, new_ver_info, gpfs_options);
-
-                if (this._is_versioning_suspended()) {
+                if (!is_gpfs && this._is_versioning_suspended()) {
```
romayalon, line 1423: can you just add a comment, something like - 
on GPFS & while versioing is suspended & latest version is a null version, we can skip unlinking the old version because linkatif overrides.

src/sdk/namespace_fs.js
```diff
@@ -1419,23 +1419,22 @@ class NamespaceFS {
...
                         await this._delete_null_version_from_versions_directory(key, fs_context);
                     }
                 }
+                gpfs_options = is_gpfs ?
+                await this._open_files_gpfs(fs_context, new_ver_tmp_path, latest_ver_path, upload_file,
+                    latest_ver_info, open_mode, undefined, versioned_info) :
+                undefined;
+                dbg.log1('Namespace_fs._move_to_dest_version:', latest_ver_info, new_ver_info, gpfs_options);
+
                 if (latest_ver_info &&
```
romayalon, line 1438: Not related to this line, on the github issue there is a test called test_s3cever102, is it already running in our ceph tests run?

src/sdk/namespace_fs.js
```diff
@@ -1419,23 +1419,22 @@ class NamespaceFS {
                 const versioned_path = latest_ver_info && this._get_version_path(key, latest_ver_info.version_id_str);
                 const versioned_info = latest_ver_info && await this._get_version_info(fs_context, versioned_path);
 
-                gpfs_options = is_gpfs ?
```
romayalon, line 1422: can you move it back here? I think we wanted to open it at the beginning of the function

src/sdk/namespace_fs.js
nadavMiz: no this test was made by herald. I think there is already test for that on our test. I'll check and add one if its missing

src/sdk/namespace_fs.js
```diff
@@ -1419,23 +1419,22 @@ class NamespaceFS {
...
-                    await this._open_files_gpfs(fs_context, new_ver_tmp_path, latest_ver_path, upload_file,
-                        latest_ver_info, open_mode, undefined, versioned_info) :
-                    undefined;
                 const bucket_tmp_dir_path = this.get_bucket_tmpdir_full_path();
-                dbg.log1('Namespace_fs._move_to_dest_version:', latest_ver_info, new_ver_info, gpfs_options);
```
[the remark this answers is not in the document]
nadavMiz: might as well keep the gpfs options here instead of having two comments

src/sdk/namespace_fs.js
nadavMiz: there is already a test in test_bucketspace_versioning.js:
https://github.com/noobaa/noobaa-core/blob/3c8a29a2153fe51e46b859617c8a193758e8abe7/src/test/unit_tests/test_bucketspace_versioning.js#L453-L478

nadavMiz: there is already a test for this in test_bucketspace_versioning.js:
https://github.com/noobaa/noobaa-core/blob/3c8a29a2153fe51e46b859617c8a193758e8abe7/src/test/unit_tests/test_bucketspace_versioning.js#L366-L372

src/sdk/namespace_fs.js
```diff
@@ -1427,15 +1427,16 @@ class NamespaceFS {
                 dbg.log1('Namespace_fs._move_to_dest_version:', latest_ver_info, new_ver_info, gpfs_options);
 
                 if (this._is_versioning_suspended()) {
-                    if (latest_ver_info?.version_id_str === NULL_VERSION_ID) {
+                    //on GPFS safe_move overrides the latest object so no need to unlink
+                    if (!is_gpfs && latest_ver_info?.version_id_str === NULL_VERSION_ID) {
```
romayalon, line 1431: I think that this condition is not good because if it's GPFS but it's also NULL_VERSION_ID, we will do the else, I think that on the GPFS machine you didn't have this issue

nadavMiz: right I changed it to be a condition inside the is latest version null condition. this way if the version isn't null we will act the same, and if it is null we will no unlink only on GPFS machine

src/sdk/namespace_fs.js
romayalon: Hey can you move it back here?

romayalon (approved):

Issue #8380, opened 2024-09-18
NOOBAA_LOG_LEVEL is not working as expected


ramya-c3:
### Environment info
- NooBaa Version: noobaa-core-5.17.0-20240904.el9.x86_64
- Platform: Standalone

### Actual behavior
config.LOG_TO_STDERR_ENABLED = true;
config.LOG_TO_SYSLOG_ENABLED = true; 
In this case it is logging all the unwanted messages in the console,noobaa.log and messages.

### Expected behavior
1. we should not see messages in the console. 

### Steps to reproduce
1.config.LOG_TO_STDERR_ENABLED = true;
config.LOG_TO_SYSLOG_ENABLED = true; 

Broadcast message from systemd-journald@ramyamc-42 (Wed 2024-09-18 10:20:09 CEST):

[3267897]: 2024-09-18 10:20:09.491233 [PID-3267897/TID-3267897] FS::set_log_config: LOG_TO_STDERR_ENABLED=1 LOG_TO_SYSLOG_ENABLED=1

romayalon:
@naveenpaul1 can you please take a look?

ramya-c3:
@naveenpaul1 So you mean to say we should not use these 2 config parameters?

naveenpaul1:
issue tested and working as expected
cc : @ramya-c3

PR #8389, opened 2024-09-20
NSFS | NOOBAA_LOG_LEVEL is not working as expected

naveenpaul1:
### Explain the changes
1. log in nappi layer which is getting printed every time config.json reloaded, Its adding longs to the console. Solution would be update priority for message to info(5)

1. https://github.com/noobaa/noobaa-core/issues/8380

### Testing Instructions:
1. update config.json
config.LOG_TO_SYSLOG_ENABLED = true;
and check below logs printing on the console log, every time config.json relaod

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -2176,7 +2176,7 @@ set_debug_level(const Napi::CallbackInfo& info)
 {
     int level = info[0].As<Napi::Number>();
     DBG_SET_LEVEL(level);
-    LOG("FS::set_debug_level " << level);
+    DBG1("FS::set_debug_level " << level);
     return info.Env().Undefined();
 }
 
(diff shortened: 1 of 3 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -2176,7 +2176,6 @@ set_debug_level(const Napi::CallbackInfo& info)
 {
     int level = info[0].As<Napi::Number>();
     DBG_SET_LEVEL(level);
-    LOG("FS::set_debug_level " << level);
```
romayalon, line 2179: @naveenpaul1 why LOG() prints to console? we have other calls to LOG(), do the other calls be printed to console?
Is there an option to no remove it but print it to the logs instead?

naveenpaul1: @romayalon Other calls not printing to console, 
Two methods are getting called every time config file update, I can try with DBG1()

romayalon (approved):

ramya-c3:
config.LOG_TO_SYSLOG_ENABLED = true;  Incase of syslog-ng even if both of them are set to false i still see the messages

PR #8393, opened 2024-09-23
NC | Copy based on link/same inode overrides existing xattr

romayalon:
### Explain the changes
1. This is a follow-up fix to the comment https://github.com/noobaa/noobaa-core/pull/8355#discussion_r1762018815 - the recent changes removed an old code that tried to handle skipping setting the xattr on copy object when the copy is being done by link/same inode. Changing the object xattrs like user md, content type and storage_class,  initiate a non server side copy, therefore, setting the xattr again on the source/dst is not corrupting the xattr of the source but just doing a redundant FS operation.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


romayalon:
@guymguym this is a follow-up PR to your question on the copy object fix PR

nadavMiz (approved):
LGTM

PR #8398, opened 2024-09-24
NC | NSFS | Versioning | Suspended Mode | Optimization of System Calls

shirady:
### Explain the changes
1. Find the cases where we can use fewer system calls, for example: 
- `unlink` instead of `safe_unlink`: `safe_unlink` under the hood is `rename`, `stat`, and in case of error also `link` (in `fs_napi.cpp` see `SafeUnlink`).
- `rename` instead of `safe_move`: `safe_move` under the hood is `safe_link` and `safe_unlink` (in `native_fs_utils.js` see `safe_move`.

The cases that were changed:
   1. In `_move_to_dest_version` with suspended mode with latest version null - no need for `safe_unlink`, in any way the last version will be saved as a null version, therefore we will use `unlink` (decided to use `unlink_ignore_enoent` in case of concurrent calls that might lead to unlink in one process (before added the new version) and then context switch to a second process that also tries to unlink the same null version).
   2. In `_move_to_dest_version` with suspended mode with latest version null - no need for `safe_move`, `rename` should be enough,  in any way the last version will be saved as a null version, and rename is atomic (will end up with either the old or the new version and nothing else).
   3. In `_delete_latest_version` with suspended mode and the version id is null - no need for `safe_unlink` using `unlink` instead (as mentioned in point 1).
   4. In `_delete_null_version_from_versions_directory` we want to remove the null version from `.versions/` directory, no need for `safe_unlink` and use `unlink` instead (as mentioned in point 1). Since we didn't use the `safe_unlink` and this function is only used in suspended mode we could remove additional code lines that were related to the arguments of `safe_unlink`.

### Issues: 
1. This was a gap mentioned in suspended mode PRs, for example in #7311 in [this comment](https://github.com/noobaa/noobaa-core/pull/7311#discussion_r1221263701).

### Testing Instructions:
Unit tests:
1. Please run: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace_versioning.js`

Note: These changes have not been tested on a GPFS machine yet.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1413,12 +1413,20 @@ class NamespaceFS {
         let retries = config.NSFS_RENAME_RETRIES;
         for (;;) {
             try {
-                const new_ver_info = !is_gpfs && await this._get_version_info(fs_context, new_ver_tmp_path);
                 // get latest version_id if exists
                 const latest_ver_info = await this._get_version_info(fs_context, latest_ver_path);
+                if (this._is_versioning_suspended()) {
+                    if (latest_ver_info?.version_id_str === NULL_VERSION_ID) {
+                        dbg.log1('NamespaceFS._move_to_dest_version suspended: version ID of the latest version is null - the file will be unlinked');
+                        await native_fs_utils.unlink_ignore_enoent(fs_context, latest_ver_path);
+                    } else {
+                        // remove a version (or delete marker) with null version ID from .versions/ (if exists)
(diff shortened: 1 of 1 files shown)


shirady:
I am closing this PR due to changes in priorities.
The current thought is to add these changes with bucket cache changes (to avoid one process with suspended mode while the bucket configuration changed to Enabled and, a second process operates on the bucket and during its operation).

PR #8401, opened 2024-09-24
NSFS | GPFS | fix gpfs test file

nadavMiz:
### Explain the changes
fix small issues with gpfs tests file:
1. await bucket creation
2. use correct key for test
3. get created_delete_marker from object inside list instead of the list itself

### Testing Instructions:
need to run test file with the GPFS_ROOT_PATH enviorment variable. this variable should should point to the root of the gpfs file system where buckets are create (in this example `/ibm/fs1/ `). note that the test file will create his own tests directory, so no need to worry about it changing existing buckets.
1. run `GPFS_ROOT_PATH=/ibm/fs1/ node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nsfs_versioning_gpfs.js`
on GPFS cluster

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/test/unit_tests/test_nsfs_versioning_gpfs.js
+++ b/src/test/unit_tests/test_nsfs_versioning_gpfs.js
@@ -53,7 +53,7 @@ mocha.describe('namespace_fs gpfs- versioning', async function() {
             console.log('No Root permissions found in env. Skipping test');
             this.skip(); // eslint-disable-line no-invalid-this
         }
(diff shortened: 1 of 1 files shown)


shirady (approved):
LGTM

shirady:
@nadavMiz would you please add in testing instruction the env that we need to define before running the test? (`GPFS_ROOT_PATH` in this example it was `ibm/fs1`.

romayalon (approved):

PR #8405, opened 2024-09-24
NSFS | Versioning | Concurrency tests

romayalon:
### Explain the changes
1. Added 7 concurrency tests - 
   a. concurrent puts & delete latest objects
   b. concurrent puts & delete objects by version id
   c. concurrent delete objects by version id/latest - **currently being skipped - there is probably a bug in that flow, Opened an issue - https://github.com/noobaa/noobaa-core/issues/8414**
   d. nested key - concurrent delete multiple objects.
   e. nested key - concurrent puts & deletes.
   f. concurrent puts & list versions
   g. concurrent puts & list versions - version id paging - **currently being skipped - there is probably a bug in that flow, opened an issue -  https://github.com/noobaa/noobaa-core/issues/8406**
3. Fixed indentation and moved up a printing.
4. Refactored some existing concurrency tests.

1. Remove the skip from these tests.
2. List object versions + version id marker not working - https://github.com/noobaa/noobaa-core/issues/8406
3. Delete latest + delete object version id which is also the latest bug - https://github.com/noobaa/noobaa-core/issues/8414

### Testing Instructions:
1. `sudo jest --testRegex=jest_tests/test_versioning_conc`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/test/unit_tests/jest_tests/test_versioning_concurrency.test.js
+++ b/src/test/unit_tests/jest_tests/test_versioning_concurrency.test.js
@@ -1,4 +1,5 @@
 /* Copyright (C) 2016 NooBaa */
+/* eslint-disable max-lines-per-function */
 'use strict';
 
 const path = require('path');
@@ -56,16 +57,20 @@ describe('test versioning concurrency', () => {
         const bucket = 'bucket1';
         const key = 'key1';
         const failed_operations = [];
-        for (let i = 0; i < 5; i++) {
+        const successful_operations = [];
+        const num_of_concurrency = 5;
+        for (let i = 0; i < num_of_concurrency; i++) {
             const random_data = Buffer.from(String(i));
             const body = buffer_utils.buffer_to_read_stream(random_data);
             nsfs.upload_object({ bucket: bucket, key: key, source_stream: body }, DUMMY_OBJECT_SDK)
-                .catch(err => failed_operations.push(err));
+                .catch(err => failed_operations.push(err))
+                .then(res => successful_operations.push(res));
         }
-        await P.delay(1000);
-        expect(failed_operations.length).toBe(0);
+        await P.delay(2000);
+        expect(successful_operations).toHaveLength(num_of_concurrency);
+        expect(failed_operations).toHaveLength(0);
         const versions = await nsfs.list_object_versions({ bucket: bucket }, DUMMY_OBJECT_SDK);
-        expect(versions.objects.length).toBe(5);
+        expect(versions.objects.length).toBe(num_of_concurrency);
     });
 
     it('multiple delete version id and key', async () => {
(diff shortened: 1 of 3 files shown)


src/test/unit_tests/jest_tests/test_versioning_concurrency.test.js
```diff
@@ -181,6 +188,277 @@ describe('test versioning concurrency', () => {
...
+        expect(versions.objects).toHaveLength(1000);
+        const versions1 = await nsfs.list_object_versions({ bucket, version_id_marker: versions.next_version_id_marker }, DUMMY_OBJECT_SDK);
+        expect(versions1.objects).toHaveLength(1000);
+        const versions2 = await nsfs.list_object_versions({ bucket, version_id_marker: versions1.next_version_id_marker}, DUMMY_OBJECT_SDK);
+        expect(versions2.objects).toHaveLength(520);
+
+        const num_of_delete_markers = (versions.objects.filter(version => version.delete_marker === true)).length;
+        expect(num_of_delete_markers).toBe(0);
+        const num_of_latest_versions = (versions.objects.filter(version => version.is_latest === true)).length;
+        expect(num_of_latest_versions).toBe(initial_num_of_objects);
```
tangledbytes, line 460: Merge the versions*.objects into single array and then check. The first versions output has only 1k entries, it won't have all the delete markers (if any) and all the latest entries.

tangledbytes: Also let's consider adding something like:

```js
        const version_set = new Set(merged_versions.map(version => version.version_id));
        expect(version_set.size).toBe(2520);
```

To ensure that we do get exact number of unique versions.

src/test/unit_tests/jest_tests/test_versioning_concurrency.test.js
```diff
@@ -181,6 +187,274 @@ describe('test versioning concurrency', () => {
...
+        let merged_versions = [];
+        const expected_num_of_versions = initial_num_of_objects * initial_num_of_versions + num_of_concurrency;
+
+        for (;;) {
+            const versions = await nsfs.list_object_versions({ bucket, key_marker, version_id_marker }, DUMMY_OBJECT_SDK);
+            merged_versions = merged_versions.concat(versions.objects);
+            version_id_marker = versions.next_version_id_marker;
+            key_marker = versions.next_marker;
+            if (!version_id_marker) break;
+        }
```
shirady, line 451: Is this part is duplicate of lines 415-421? (only different array?) - can we move it to a function?
Could you please provide an explanation of this part in the function JSDoc?

src/test/unit_tests/jest_tests/test_versioning_concurrency.test.js
```diff
@@ -193,11 +467,10 @@ describe('test versioning concurrency', () => {
 async function _upload_versions(bucket, key, number_of_versions) {
     const versions_arr = [];
     for (let i = 0; i < number_of_versions; i++) {
-        const random_data = Buffer.from(String(i));
+        const random_data = Buffer.from(String(crypto_random_string(7)));
```
shirady, line 470: Would you please explain this small change?

shirady (approved):
LGTM

src/test/unit_tests/jest_tests/test_versioning_concurrency.test.js
romayalon: instead of creating a buffer from the number we create a random string

shirady: Thanks, what was the reason for the change?

romayalon: it looked like a mistake

PR #8410, opened 2024-09-25
NSFS | versioning | fix GET/HEAD object concurrency issues

### LINKED ISSUE #8397: warp versioned concurrent=100
### Environment info
- NooBaa Version: **VERSION**
-c83f2-dan9-hs200.test.net:  noobaa-core-5.17.0-20240921.el8.x86_64
c83f2-dan10-hs200.test.net:  noobaa-core-5.17.0-20240921.el8.x86_64
c83f2-dan8-hs200.test.net:  noobaa-core-5.17.0-20240921.el8.x86_64

### Actual behavior
1.using warp versioned with concurrent=100 fails due to this errors

```
warp: <ERROR> download error: The specified key does not exist.
warp: <ERROR> unexpected download size. want:287, got:454
warp: <ERROR> stat error:The specified key does not exist.
warp: <ERROR> unexpected download size. want:377, got:51
warp: <ERROR> unexpected stat size. want:944, got:591
warp: <ERROR> unexpected download size. want:79, got:353
warp: <ERROR> unexpected download size. want:52, got:68
warp: <ERROR> stat error:The specified key does not exist.
warp: <ERROR> unexpected download size. want:469, got:367
warp: <ERROR> unexpected stat size. want:26, got:258
warp: <ERROR> unexpected download size. want:758, got:701
warp: <ERROR> unexpected stat size. want:51, got:431
```
 `warp versioned --host=172.20.100.6{0...9}:6443 --access-key="$access_key" --secret-key="$secret_key" --obj.randsize --obj.size=1k  --duration=1m --disable-multipart --concurrent=100 --bucket="mixwarp-new-bucket-vers" --insecure --tls`

with concurrent =1 it doesn't fail
 `warp versioned --host=172.20.100.6{0...9}:6443 --access-key="$access_key" --secret-key="$secret_key" --obj.randsize --obj.size=1k  --duration=1m --disable-multipart --concurrent=1 --bucket="mixwarp-new-bucket-vers" --insecure --tls`
### Expected behavior
1.warp versioned complete

### Steps to reproduce
1.`warp versioned --host=172.20.100.6{0...9}:6443 --access-key="$access_key" --secret-key="$secret_key" --obj.randsize --obj.size=1k  --duration=1m --disable-multipart --concurrent=100 --bucket="mixwarp-new-bucket-vers" --insecure --tls`

nadavMiz:
### Explain the changes
1. in read_object_stream, read_object_md we first check where the file is and then read the file. the file might move from .versions or from .versions to bucket folder causing us to either fail on EONT because the file doesn't exist. or to fail because we are reading the wrong version (when the path is the latest version, but the latest version changed). add try catch file not exist errors, and add check to check if we are reading the correct version
2. we also have another issue where two put operations move the latest version to the .version folder. one version will put the latest version to the .versions dir and put itself as the latest. the other put will put the now new version in the versions die as if it were the previous latest version and overriding the old version with the wrong file. use posix_safe_link in this case as it validates both that we didn't override and second that we moved the correct version  
### Testing Instructions:
1. run `warp versioned --host=localhost:6443 --access-key="$access_key" --secret-key="$secret_key" --obj.randsize --obj.size=1k  --duration=1m --disable-multipart --concurrent=100 --bucket="mixwarp-new-bucket-vers" --insecure --tls`

- [ ] Doc added/updated
- [ ] Tests added


src/sdk/namespace_fs.js
```diff
@@ -924,22 +924,41 @@ class NamespaceFS {
...
+                        } else if (stat.xattr?.[XATTR_DIR_CONTENT] !== '0') {
+                            // find dir object content file path  and return its stat + xattr of its parent directory
+                            const dir_content_path = await this._find_version_path(fs_context, params);
+                            const dir_content_path_stat = await nb_native().fs.stat(fs_context, dir_content_path);
+                            const xattr = stat.xattr;
+                            stat = { ...dir_content_path_stat, xattr };
+                        }
+                    }
+                    if (params.version_id && !this._is_versioning_disabled() && this._get_version_id_by_xattr(stat) !== params.version_id) {
+                        dbg.warn('NamespaceFS.read_object_md mismatch version_id', params.version_id, this._get_version_id_by_xattr(stat));
```
romayalon, line 952: can you instead just throw an error and add the error code to the should_retry_link_unlink() check?
with this change, we will handle the retries once

src/sdk/namespace_fs.js
```diff
@@ -960,34 +996,39 @@ class NamespaceFS {
...
+                    if (params.version_id && !this._is_versioning_disabled() && this._get_version_id_by_xattr(stat) !== params.version_id) {
+                        retries -= 1;
+                        if (retries <= 0) throw new Error('to many retries - mismatch version_id');
+                        dbg.warn('NamespaceFS.read_object_stream mismatch version_id', params.version_id, this._get_version_id_by_xattr(stat));
+                        continue;
+                    }
+                    break;
                 } catch (err) {
-                    //failed to get object
-                    new NoobaaEvent(NoobaaEvent.OBJECT_GET_FAILED).create_event(params.key,
```
romayalon, line 977: can you bring back this event when throwing an error?

src/sdk/namespace_fs.js
```diff
@@ -960,34 +996,39 @@ class NamespaceFS {
...
+                    if (await this._is_empty_directory_content(file_path, fs_context, params)) return null;
+
+                    file = await nb_native().fs.open(
+                        fs_context,
+                        file_path,
+                        config.NSFS_OPEN_READ_MODE,
+                        native_fs_utils.get_umasked_mode(config.BASE_MODE_FILE),
+                    );
+                    stat = await file.stat(fs_context);
+                    if (params.version_id && !this._is_versioning_disabled() && this._get_version_id_by_xattr(stat) !== params.version_id) {
```
romayalon, line 1019: could you move this condition to a function as well?

src/sdk/namespace_fs.js
nadavMiz: this even still exists I just moved it in _is_empty_directory_content with the rest of the code

romayalon: I saw it, but it should be here, we should emit this event not only on the catch of _is_empty_directory_content () and when is_dir_content is true

src/util/native_fs_utils.js
```diff
@@ -199,10 +199,17 @@ async function safe_move_posix(fs_context, src_path, dst_path, src_ver_info, tmp
...
 // src_file exists on uploads (open mode = 'w' ) or deletions
 // on uploads (open mode 'wt') the dir_file is used as the link source
-async function safe_move_gpfs(fs_context, src_path, dst_path, gpfs_options) {
-    const { src_file = undefined, dst_file = undefined, dir_file = undefined, should_unlink = false } = gpfs_options;
+async function safe_move_gpfs(fs_context, src_path, dst_path, gpfs_options, src_ver_info) {
+    const { src_file = undefined, dst_file = undefined, dir_file = undefined, should_unlink = false,
+        is_move_to_versions = false } = gpfs_options;
     dbg.log1('Namespace_fs.safe_move_gpfs', src_path, dst_path, dst_file, should_unlink);
-    await safe_link_gpfs(fs_context, dst_path, src_file || dir_file, dst_file);
+    if (is_move_to_versions) {
```
romayalon, line 206: I'm not convinced this function should be aware of the destination... 
also you used safe_move_posix() directly, so why doing this change?

src/sdk/namespace_fs.js
```diff
@@ -2930,9 +2971,9 @@ class NamespaceFS {
                     const bucket_tmp_dir_path = this.get_bucket_tmpdir_full_path();
                     if (this._is_versioning_enabled() || suspended_and_latest_is_not_null) {
                         await native_fs_utils._make_path_dirs(versioned_path, fs_context);
-                        await native_fs_utils.safe_move(fs_context, latest_ver_path, versioned_path, latest_ver_info,
-                            gpfs_options && gpfs_options.delete_version, bucket_tmp_dir_path);
-                        if (suspended_and_latest_is_not_null) {
+                         await native_fs_utils.safe_move_posix(fs_context, latest_ver_path, versioned_path, latest_ver_info,
```
romayalon, line 2974: we said we need that also on _delete_latest_version, right? I couldn't find it

nadavMiz: right we need it whenever we move to .version

src/util/native_fs_utils.js
nadavMiz: on delete we call safe_move_posix() directly because we also want to unlink the source. but on move_to_dest we dont want to unlink because the file we be overwritten by the move to latest. 
The reason I have this if here instead of in _move_to_dest_version is because its seemed to be in line with how we use safe_move in combination with open_files_gpfs. where open_files_gpfs decides how the move should be handled (e.g should we verify the dst file, should we unlink ), and safe_move_gpfs does the move

src/sdk/namespace_fs.js
nadavMiz: looks like the move in _delete_latest_version is already safe_move_posix

guymguym (changes requested):

PR #8413, opened 2024-09-26
[NSFS] Fix version listing

### LINKED ISSUE #8406: NSFS | Versioning | ListObjectVersions | Version ID marker is not working
### Environment info
- NooBaa Version: 5.18
- Platform: All

### Actual behavior
1. In a test added lately `concurrent puts & list versions - version id paging` it looks like version ID marker is not working, I don't see any other functionality tests about this flow.

### Expected behavior
1. When running List versions while providing version ID marker, another 1000 versions should be returned to the user and not the old 1000 entries.

### Steps to reproduce
1. `sudo jest --testRegex=jest_tests/test_versioning_conc -t 'concurrent puts & list versions'`

tangledbytes:
### Explain the changes
This PR attempts to fix version listing. Previously, there was an assumption that when both `KeyMarker` and `NextIdVersion` is provided then only the items belonging to that `KeyMarker` starting from the given `IdVersion` should be returned. This behaviour sounds sane but upon checking with AWS S3 it turns out that AWS would keep on listing items beyond `KeyMarker` as well (and the way to control the key is by using `Prefix` and `MaxKeys`).

This PR simply removes that additional check.

Fixes https://github.com/noobaa/noobaa-core/issues/8406

### Testing Instructions:
1. sudo jest --testRegex=jest_tests/test_versioning_conc -t 'concurrent puts & list versions' (After https://github.com/noobaa/noobaa-core/pull/8405 gets merged)

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -811,10 +811,6 @@ class NamespaceFS {
                     }
                     for (let i = marker_index; i < sorted_entries.length; ++i) {
                         const ent = sorted_entries[i];
-                        if (list_versions && marker_curr) {
-                            const ent_name = _get_filename(ent.name);
-                            if (ent_name !== marker_curr) break;
-                        }
                         // when entry is NSFS_FOLDER_OBJECT_NAME=.folder file,
                         // and the dir key marker is the name of the curr directory - skip on adding it
                         if (ent.name === config.NSFS_FOLDER_OBJECT_NAME && dir_key === marker_dir) {


romayalon:
@tangledbytes 
1. While taking this fix and putting it in my test, and passing in the listObjectVersions request only a NextVersionIdMarker (KeyMarker is not provided) I experience still the same issue (always getting the first 1000 objects) is this the expected behavior? 
2. I think we need functionality tests here and we shouldn't rely on the concurrency test

tangledbytes:
> @tangledbytes
> 
> 1. While taking this fix and putting it in my test branch, and passing in the listObjectVersions request only a NextVersionIdMarker (KeyMarker is not provided) I experience still the same issue (always getting the first 1000 objects) is this the expected behavior?
> 2. I think we need functionality tests here and we shouldn't rely on the concurrency test

@romayalon,

1. I think the implementation expects it, we have the check for this at the s3 layer and AWS S3 throws error as well if we don't KeyMarker with version id. And that makes sense right? The fact that keymarker and versionid share the same prefix is a consequence of our implementation and not inherent to S3 so it does makes sense that both are required at the same time as version id without the key name might not make much sense.
2. Yes, I can add the test for the same here.

romayalon (approved):
Let's merge it and add functionality tests on a follow-up PR, i'll remove the skip on my test after you merge it.

Issue #8400, opened 2024-09-24
NSFS | Tagging bugs


romayalon:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. Tagging is not supported on Upload Object/Copy object (non server side copy/taggingDirective Replace) - it's being ignored.
2. Get Object tagging leaves the file opened without closing it - 
```
2024-09-24 11:21:08.704203 [PID-16479/TID-259] FS::FileWrap::dtor: file not closed _path=/private/tmp/dir2/obj-with-tagging _fd=25
2024-09-24 11:21:08.704508 [PID-16479/TID-259] FS::FileWrap::dtor: file not closed _path=/private/tmp/dir2/obj-with-tagging-copy _fd=24
```

### Expected behavior
1. Tagging should be supported on Upload/Copy Object.
2. Get Object Tagging should close all files opened by the operation.

### Steps to reproduce
**Tagging is not supported on Upload Object/Copy object - it's being ignored.**
Upload object -
1. Start NooBaa NC - `sudo node src/cmd/nsfs  --debug=5`
2. Upload an object - 
```bash 
s3api put-object --bucket bucket1 --key obj-with-tagging2 --body obj1 --tagging=`tag1=val1`
```
4. Get object Tagging of the object - 
```bash 
s3api get-object-tagging --bucket bucket1 --key obj-with-tagging77
{
    "TagSet": []
}
```

**Get Object tagging leaves the file opened without closing it**
1. Start NooBaa NC - `sudo node src/cmd/nsfs  --debug=5`
2. Upload an object - `s3api put-object --bucket bucket1 --key obj1 --body obj1`
5. Get Tagging of the object - `s3api get-object-tagging --bucket bucket1 --key obj1`
6. Wait a minute and check the logs of nsfs.js - and search for the following message - 
`2024-09-24 11:33:13.711420 [PID-29422/TID-259] FS::FileWrap::dtor: file not closed _path=/private/tmp/dir2/obj77 _fd=24`

naveenpaul1:
code merged

PR #8417, opened 2024-09-27
NSFS | Tagging bugs

naveenpaul1:
### Explain the changes
1. tag is added to file/folder xattr from request
2. file close after open for tagging
3. Object tag work with slash for folder objects
4. get last object version specific tag using version id or without.

1. https://github.com/noobaa/noobaa-core/issues/8400

### Testing Instructions:
1. please check above issue

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1165,7 +1165,6 @@ class NamespaceFS {
             await this._throw_if_storage_class_not_supported(params.storage_class);
 
             upload_params = await this._start_upload(fs_context, object_sdk, file_path, params, open_mode);
-
             if (!params.copy_source || upload_params.copy_res === copy_status_enum.FALLBACK) {
                 // We are taking the buffer size closest to the sized upload
                 const bp = multi_buffer_pool.get_buffers_pool(params.size);
@@ -1330,6 +1329,13 @@ class NamespaceFS {
 
             if (params.storage_class === s3_utils.STORAGE_CLASS_GLACIER) {
                 await this.append_to_migrate_wal(file_path);
+            }
+        }
+        if (params.tagging) {
+            for (const { key, value } of params.tagging) {
+                fs_xattr = Object.assign(fs_xattr || {}, {
+                    [XATTR_TAG + key]: value
+                });
             }
         }
         if (fs_xattr && !is_dir_content && should_replace_xattr) await target_file.replacexattr(fs_context, fs_xattr);
@@ -1952,15 +1958,17 @@ class NamespaceFS {
     async get_object_tagging(params, object_sdk) {
         const tag_set = [];
         let file_path;
-        if (params.version_id && this._is_versioning_enabled()) {
-            file_path = this._get_version_path(params.key, params.version_id);
+        let file;
+        const fs_context = this.prepare_fs_context(object_sdk);
+        // Version specific tag will return even if versioning suspended.
+        if (params.version_id) {
+            file_path = await this._find_version_path(fs_context, params, true);
         } else {
-            file_path = this._get_file_path(params);
+            file_path = this._get_file_md_path(params);
         }
         try {
-            const fs_context = this.prepare_fs_context(object_sdk);
             dbg.log0('NamespaceFS.get_object_tagging: param ', params, 'file_path :', file_path);
-            const file = await nb_native().fs.open(fs_context, file_path);
+            file = await nb_native().fs.open(fs_context, file_path);
             const stat = await file.stat(fs_context);
             if (stat.xattr) {
                 for (const [xattr_key, xattr_value] of Object.entries(stat.xattr)) {
@@ -1975,6 +1983,8 @@ class NamespaceFS {
         } catch (err) {
             dbg.error(`NamespaceFS.get_object_tagging: failed in dir ${file_path} with error: `, err);
             throw native_fs_utils.translate_error_codes(err, native_fs_utils.entity_enum.OBJECT);
+        } finally {
+            if (file) await file.close(fs_context);
         }
         dbg.log0('NamespaceFS.get_object_tagging: return tagging ', tag_set, 'file_path :', file_path);
         return { tagging: tag_set };
--- a/src/test/unit_tests/test_namespace_fs.js
+++ b/src/test/unit_tests/test_namespace_fs.js
@@ -1,5 +1,6 @@
 /* Copyright (C) 2020 NooBaa */
 /*eslint max-lines-per-function: ["error", 900]*/
+/*eslint max-lines: ["error", 2100]*/
 'use strict';
 
 const _ = require('lodash');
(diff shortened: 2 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -811,6 +811,10 @@ class NamespaceFS {
                     }
                     for (let i = marker_index; i < sorted_entries.length; ++i) {
                         const ent = sorted_entries[i];
+                        if (list_versions && marker_curr) {
+                            const ent_name = _get_filename(ent.name);
+                            if (ent_name !== marker_curr) break;
+                        }
```
shirady, line 817: I guess that after rebase this part would be deleted, right? (it is added in PR #8413 ).

src/sdk/namespace_fs.js
```diff
@@ -1321,6 +1326,11 @@ class NamespaceFS {
                 await this.append_to_migrate_wal(file_path);
             }
         }
+        if (params.tagging) {
+            for (const { key, value } of params.tagging) {
+                fs_xattr[XATTR_TAG + key] = value;
+            }
+        }
```
shirady, line 1333: I saw that in previous assigns they used `fs_xattr = Object.assign(fs_xattr || {}, ...` since there might be a case where `fs_xattr` is `undefined`.

src/sdk/namespace_fs.js
```diff
@@ -1941,15 +1951,18 @@ class NamespaceFS {
     async get_object_tagging(params, object_sdk) {
         const tag_set = [];
         let file_path;
+        let file;
+        // In case of object as a folder. folder tag can access using exact object key `folder1/` or folder name `folder1`
+        params.key = params.key.replace(/\/$/, '');
```
shirady, line 1956: Could you please save it in a separate variable?
Usually, we don't manipulate the parameters we receive directly.

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -1764,8 +1764,111 @@ mocha.describe('namespace_fs copy object', function() {
...
+        const ns_tmp_tagging = new NamespaceFS({ bucket_path: ns_tmp_bucket_path_tagging, bucket_id: '3', namespace_resource_id: undefined });
+        mocha.before(async () => {
+            await P.all(_.map([src_bkt, upload_bkt_tagging], async buck =>
+                fs_utils.create_fresh_path(`${tmp_fs_path_tagging}/${buck}`)));
+        });
+        mocha.after(async () => {
+            await P.all(_.map([src_bkt, upload_bkt_tagging], async buck =>
+                fs_utils.folder_delete(`${tmp_fs_path_tagging}/${buck}`)));
+        });
+        mocha.after(async () => fs_utils.folder_delete(tmp_fs_path_tagging));
```
shirady, line 1781: You can add the command in the previous `mocha.after` if you want.

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -1764,8 +1764,111 @@ mocha.describe('namespace_fs copy object', function() {
...
+                console.log('upload_object (tag) response', inspect(upload_res));
+                const tag_res = await ns_tmp_tagging.get_object_tagging({
+                    bucket: upload_bkt_tagging,
+                    key: upload_key,
+                }, dummy_object_sdk);
+                console.log('get_object_tagging response', inspect(tag_res));
+                await validate_tagging(tag_res, tagging_1);
+            });
+
+            mocha.it('Upload object with tagging, update object and tag', async function() {
```
shirady, line 1808: Could you please explain where is the part of "update object" (I see only the "update tag" where it is `tagging_2` instead of `tagging_1`)?

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -1764,8 +1764,111 @@ mocha.describe('namespace_fs copy object', function() {
...
+                console.log('upload_object (tag) response', inspect(upload_res));
+                const tag_res = await ns_tmp_tagging.get_object_tagging({
+                    bucket: upload_bkt_tagging,
+                    key: upload_folder_key,
+                }, dummy_object_sdk);
+                console.log('get_object_tagging response', inspect(tag_res));
+                await validate_tagging(tag_res, tagging_1);
+            });
+
+            mocha.it('Upload folder object with tagging, get key wthout slash', async function() {
```
shirady, line 1845: Adding a minor suggestion if you want
```suggestion
            mocha.it('Upload folder object with tagging, get object tagging key without slash', async function() {
```

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -1764,8 +1764,111 @@ mocha.describe('namespace_fs copy object', function() {
...
+                    key: upload_folder_key_without_slash,
+                }, dummy_object_sdk);
+                console.log('get_object_tagging response', inspect(tag_res));
+                await validate_tagging(tag_res, tagging_2);
+            });
+        });
+    });
 });
 
+async function validate_tagging(tag_res, tag_req) {
```
shirady, line 1867: Could you change it to be an array and not just the first element? (in case someone would like to test TagSet with more than 1 element).

shirady:
@naveenpaul1 CopyObject (tagging) is not part of this fix, right?

naveenpaul1:
@shirady What I know is `copy-object` internally uses `put-object` code for copy. `upload_object()` method getting called for `copy-object`  request with tag info.
@romayalon correct me if I am wrong.

romayalon (changes requested):

src/sdk/namespace_fs.js
```diff
@@ -1941,15 +1948,18 @@ class NamespaceFS {
     async get_object_tagging(params, object_sdk) {
         const tag_set = [];
         let file_path;
+        let file;
+        // In case of object as a folder. folder tag can access using exact object key `folder1/` or folder name `folder1`
+        const key = params.key.replace(/\/$/, '');
         if (params.version_id && this._is_versioning_enabled()) {
```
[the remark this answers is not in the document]
naveenpaul1: yes @romayalon , I tested this scenario, After suspending the version calling `get-object-tagging` with `--version-id` returns the last tag. It starts to return version-specific tags after enabling the version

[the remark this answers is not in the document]
romayalon: so we need to fix it for making sure we are compatible with AWS

src/sdk/namespace_fs.js
```diff
@@ -1941,15 +1948,18 @@ class NamespaceFS {
     async get_object_tagging(params, object_sdk) {
         const tag_set = [];
         let file_path;
+        let file;
+        // In case of object as a folder. folder tag can access using exact object key `folder1/` or folder name `folder1`
+        const key = params.key.replace(/\/$/, '');
```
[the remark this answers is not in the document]
romayalon: if the key ends with a / - you should use  _get_file_md_path({ key }) because the xattrs/tags stay on the folder

src/sdk/namespace_fs.js
```diff
@@ -1941,15 +1947,17 @@ class NamespaceFS {
     async get_object_tagging(params, object_sdk) {
         const tag_set = [];
         let file_path;
-        if (params.version_id && this._is_versioning_enabled()) {
+        let file;
+        // Version specific tag will return even if versioning suspended.
+        if (params.version_id) {
             file_path = this._get_version_path(params.key, params.version_id);
```
romayalon, line 1953: this line does not make sense to me, 
let's say you have a versioning enabled bucket and you have only one version of an object, the version is located at /bucket_path/obj1.txt
but get_version_path() will always return the path of the file inside .versions/ - this means you will get ENOENT although the version exists but it's located at the location of the latest version.
check usage of _find_version_path(fs_context, { key, version_id }, return_md_path), this should fit better

naveenpaul1: make sense, PR updated

src/sdk/namespace_fs.js
```diff
@@ -1952,15 +1958,17 @@ class NamespaceFS {
     async get_object_tagging(params, object_sdk) {
         const tag_set = [];
         let file_path;
-        if (params.version_id && this._is_versioning_enabled()) {
-            file_path = this._get_version_path(params.key, params.version_id);
+        let file;
+        const fs_context = this.prepare_fs_context(object_sdk);
+        // Version specific tag will return even if versioning suspended.
+        if (params.version_id) {
```
romayalon, line 1964: you don't need this if else, you can just 
`file_path = await this._find_version_path(fs_context, params, true);`

romayalon (approved):

PR #8419, opened 2024-09-29
NC | Concurrency & refactoring | Add delay, version move checks and GPFS refactoring

### LINKED ISSUE #8414: NSFS | Versioning | Concurrent delete latest & delete object by ID which is also the latest
### Environment info
- NooBaa Version: 4.18
- Platform: All

### Actual behavior
1. During an effort of addition of concurrency tests (PR - https://github.com/noobaa/noobaa-core/pull/8405), a test called `concurrent delete objects by version id/latest` is failing and currently being skipped, this test/scenario should be invesitigated and fixed.
The failure is that we are not deleting the version and we just move it to .versions/.

### Expected behavior
1. A user creates one version of an object.
2. After concurrent delete latest & delete object by ID which is also the latest NooBaa should create a delete marker and delete the existing version id.

### Steps to reproduce
1. Check the steps mentioned on `concurrent delete objects by version id/latest`.

romayalon:
### Explain the changes
1. NamespaceFS - 
    1.1. Added to the retries a delay, the delay is the sum of a base of 70 + random(0,50). Delay is common when using retries mechanism, specifically in multithreading, if another request moved the files in a way that caused a failure to our current request, retrying right away might still have the issue if the second request didn't finish its work, therefore we wait a bit and let the other request finish its work. I saw that usually above 100 ms is good, and having an addition of random ms is nice to have.
    1.2. moved is_gpfs ? check to inside _open_files_gpfs() function
    1.3. _delete_single_object_versions() - added 2 calls to _check_version_moved() for checking if the version moved between the latest version location and .versions/ at the time of the deletion in order to make sure that we will delete the version even if it moved. the check will throw VERSION_MOVED error that will trigger a retry, in the next try we will locate the new location of the version and remove it.
    1.4. _check_version_moved() function receives key, version_id and cur_path, if cur_path is the latest version location, we will check if it was moved to .versions/, if cur_path is in .versions/ we will check if the latest version has the same version_id as the version_id param.

1. Fixed https://github.com/noobaa/noobaa-core/issues/8414

### Testing Instructions:
1. `sudo jest --testRegex=jest_tests/test_versioning_conc  -t 'concurrent delete objects by version id/latest'`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -12,6 +12,7 @@ const { v4: uuidv4 } = require('uuid');
 const P = require('../util/promise');
 const dbg = require('../util/debug_module')(__filename);
 const config = require('../../config');
+const crypto = require('crypto');
 const s3_utils = require('../endpoint/s3/s3_utils');
 const error_utils = require('../util/error_utils');
 const stream_utils = require('../util/stream_utils');
@@ -293,6 +294,16 @@ function to_fs_xattr(xattr) {
     return _.mapKeys(xattr, (val, key) => XATTR_USER_PREFIX + key);
 }
 
+/**
+ * get_random_delay returns a random delay number between base + min and max
+ * @param {number} base 
+ * @param {number} min 
+ * @param {number} max 
+ * @returns {number}
+ */
+function get_random_delay(base, min, max) {
+    return base + crypto.randomInt(min, max);
+}
 
 /**
  * @typedef {{
@@ -1391,9 +1402,10 @@ class NamespaceFS {
     }
 
     // 1. get latest version_id
-    // 2. if versioning is suspended - 
+    // 2. if versioning is suspended -
     //     2.1 if version ID of the latest version is null - 
-    //       2.1.1 remove latest version
+    //       2.1.1. if it's POSIX backend - unlink the null version
+    //       2.1.2. if it's GPFS backend - nothing to do, the linkatif will override it
     //     2.2 else (version ID of the latest version is unique or there is no latest version) -
     //       2.2.1 remove a version (or delete marker) with null version ID from .versions/ (if exists)
     //  3. if latest version exists -
@@ -1417,10 +1429,8 @@ class NamespaceFS {
                 const versioned_path = latest_ver_info && this._get_version_path(key, latest_ver_info.version_id_str);
                 const versioned_info = latest_ver_info && await this._get_version_info(fs_context, versioned_path);
 
-                gpfs_options = is_gpfs ?
-                    await this._open_files_gpfs(fs_context, new_ver_tmp_path, latest_ver_path, upload_file,
-                        latest_ver_info, open_mode, undefined, versioned_info) :
-                    undefined;
+                gpfs_options = await this._open_files_gpfs(fs_context, new_ver_tmp_path, latest_ver_path, upload_file,
+                    latest_ver_info, open_mode, undefined, versioned_info);
                 const bucket_tmp_dir_path = this.get_bucket_tmpdir_full_path();
                 dbg.log1('Namespace_fs._move_to_dest_version:', latest_ver_info, new_ver_info, gpfs_options);
 
@@ -1442,7 +1452,7 @@ class NamespaceFS {
                     dbg.log1('NamespaceFS._move_to_dest_version version ID of the latest version is a unique ID - the file will be moved it to .versions/ directory');
                     await native_fs_utils._make_path_dirs(versioned_path, fs_context);
                     await native_fs_utils.safe_move(fs_context, latest_ver_path, versioned_path, latest_ver_info,
-                        gpfs_options && gpfs_options.move_to_versions, bucket_tmp_dir_path);
+                        gpfs_options?.move_to_versions, bucket_tmp_dir_path);
                 }
                 try {
                     // move new version to latest_ver_path (key path)
@@ -1457,9 +1467,10 @@ class NamespaceFS {
             } catch (err) {
                 retries -= 1;
                 const should_retry = native_fs_utils.should_retry_link_unlink(is_gpfs, err);
-                dbg.warn(`NamespaceFS._move_to_dest_version retrying retries=${retries} should_retry=${should_retry}` +
+                dbg.warn(`NamespaceFS._move_to_dest_version error: retries=${retries} should_retry=${should_retry}` +
                     ` new_ver_tmp_path=${new_ver_tmp_path} latest_ver_path=${latest_ver_path}`, err);
                 if (!should_retry || retries <= 0) throw err;
+                await P.delay(get_random_delay(config.NSFS_RANDOM_DELAY_BASE, 0, 50));
             } finally {
                 if (gpfs_options) await this._close_files_gpfs(fs_context, gpfs_options.move_to_dst, open_mode);
             }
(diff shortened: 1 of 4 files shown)


src/sdk/namespace_fs.js
```diff
@@ -3133,6 +3157,29 @@ class NamespaceFS {
...
+    /**
+     * _check_version_moved recieves cur_path and checks if the version moved
+     * 1. if cur_path is in .versions/ - the function will check if the verison moved to latest
+     * 2. if cur_path is the latest version - the function will check if the version moved to .versions/
+     * @param {import('./nb').NativeFSContext} fs_context 
+     * @param {string} key 
+     * @param {string} version_id 
+     * @param {string} cur_path 
+     */
+    async _check_version_moved(fs_context, key, version_id, cur_path) {
```
nadavMiz, line 3169: this seems to be a partial solution since the object can be moved again before checking, maybe its worth checking both paths to see if it exists at all? even though it still not complete, as we can't check both places at the same time.

src/sdk/namespace_fs.js
```diff
@@ -2965,29 +2986,29 @@ class NamespaceFS {
...
+                await native_fs_utils.safe_unlink(fs_context, null_versioned_path, null_versioned_path_info,
+                    gpfs_options?.delete_version, bucket_tmp_dir_path);
                 break;
             } catch (err) {
                 retries -= 1;
                 if (retries <= 0 || !native_fs_utils.should_retry_link_unlink(is_gpfs, err)) throw err;
                 dbg.warn(`NamespaceFS._delete_null_version_from_versions_directory Retrying retries=${retries} null_versioned_path=${null_versioned_path}`, err);
+                await P.delay(get_random_delay(config.NSFS_RANDOM_DELAY_BASE, 0, 50));
+            } finally {
+                if (gpfs_options) await this._close_files_gpfs(fs_context, gpfs_options.delete_version, undefined, true);
```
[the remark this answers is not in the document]
romayalon: yes I saw it, it's because move_to_versions files contained in move_to_dst - 
```
move_to_versions: { src_file: dst_file, dir_file, dst_file: versioned_file },
move_to_dst: { src_file, dst_file, dir_file, versioned_file }
```

src/sdk/namespace_fs.js
```diff
@@ -1439,10 +1449,10 @@ class NamespaceFS {
                 if (latest_ver_info &&
                     ((this._is_versioning_enabled()) ||
                         (this._is_versioning_suspended() && latest_ver_info.version_id_str !== NULL_VERSION_ID))) {
-                    dbg.log1('NamespaceFS._move_to_dest_version version ID of the latest version is a unique ID - the file will be moved it to .versions/ directory');
+                    dbg.log1('NamespaceFS._move_to_dest_version version ID of the latet version is a unique ID - the file will be moved it to .versions/ directory');
```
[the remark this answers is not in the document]
nadavMiz: sorry, it should be latest and not latet. really small issue

shirady:
@romayalon Could you please add a short explanation about "Added to the retries a delay, the delay is the sum of a base of 70 + random(0,50)." in the PR description? (Why the delay solves the issue, Why we chose it, etc.).

src/sdk/namespace_fs.js
romayalon: why partial? the retires cover future moves, but I guess since we don't need this function for move_to_dest_version() we can go with your approach.

src/test/unit_tests/jest_tests/test_versioning_concurrency.test.js
```diff
@@ -181,6 +178,40 @@ describe('test versioning concurrency', () => {
         const versions = await nsfs.list_object_versions({ bucket: bucket }, DUMMY_OBJECT_SDK);
         expect(versions.objects.length).toBe(number_of_iterations + 1); // 1 version before + 10 versions concurrent
     });
+
+    it('concurrent delete objects by version id/latest', async () => {
```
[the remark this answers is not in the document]
romayalon: yep, see now

romayalon:
> @romayalon Could you please add a short explanation about "Added to the retries a delay, the delay is the sum of a base of 70 + random(0,50)." in the PR description? (Why the delay solves the issue, Why we chose it, etc.).

Delay is common when using retries mechanism, specifically in multithreading, if another request moved the files in a way that caused a failure to our current request, retrying right away might still have the issue if the second request didn't finish its work, therefore we wait a bit and let the other request finish its work. I saw that usually above 100 ms is good, and having an addition of random ms is nice to have.

shirady:
> > @romayalon Could you please add a short explanation about "Added to the retries a delay, the delay is the sum of a base of 70 + random(0,50)." in the PR description? (Why the delay solves the issue, Why we chose it, etc.).
> 
> Delay is common when using retries mechanism, specifically in multithreading, if another request moved the files in a way that caused a failure to our current request, retrying right away might still have the issue if the second request didn't finish its work, therefore we wait a bit and let the other request finish its work. I saw that usually above 100 ms is good, and having an addition of random ms is nice to have.

@romayalon Why the base is 70 and not 100?

src/sdk/namespace_fs.js
```diff
@@ -2680,9 +2691,18 @@ class NamespaceFS {
...
 
-    // 1. if version exists in .versions/ folder - return its path
-    // 2. else if version is latest version - return latest version path
-    // 3. throw ENOENT error
+    /**
+     * _find_version_path returns the path of the version
+     * 1. if version_id is not defined, it returns the key file
+     * 2. else, 
+     *    2.1. check version format
+     *    2.2. check if latest version exists and it matches the versio_id parameter the latest version path returns
```
shirady, line 2699: minor typo
```suggestion
     *    2.2. check if the latest version exists and it matches the version_id parameter the latest version path returns
```

src/sdk/namespace_fs.js
```diff
@@ -2945,9 +2965,10 @@ class NamespaceFS {
                 }
                 break;
             } catch (err) {
-                dbg.warn(`NamespaceFS._delete_latest_version: Retrying retries=${retries} latest_ver_path=${latest_ver_path}`, err);
+                dbg.warn(`NamespaceFS._delete_latest_version error: retries=${retries} latest_ver_path=${latest_ver_path}`, err);
```
shirady, line 2968: We have this printing in more places; if you change it here, it will be better in all the rest as well.

src/sdk/namespace_fs.js
```diff
@@ -3133,6 +3157,22 @@ class NamespaceFS {
...
+    /**
+     * _check_version_moved recieves key and version_id and checks if the version still exists in one of the optional locations
+     * latest version location or .versions
+     * @param {import('./nb').NativeFSContext} fs_context 
+     * @param {string} key 
+     * @param {string} version_id 
+     */
+    async _check_version_moved(fs_context, key, version_id) {
+        const latest_version_path = this._get_file_path({ key });
+        const verisoned_path = this._get_version_path(key, version_id);
```
shirady, line 3169: A typo in the variables names: `verisoned_path` -> `versioned_path` (also `verisoned_path_info`).

src/sdk/namespace_fs.js
```diff
@@ -3133,6 +3157,22 @@ class NamespaceFS {
         }
     }
 
+    /**
+     * _check_version_moved recieves key and version_id and checks if the version still exists in one of the optional locations
```
shirady, line 3161: Worth mentioning that it is out best effort?
(might be a case where it moved after this function returned and before the caller returned the value).

shirady (approved):
LGTM

Issue #8376, opened 2024-09-17
NSFS | S3 | Versioning: PutBucketLifecycleConfiguration on version-enabled bucket fails with ParamValidationError


hseipp:
### Environment info
- NooBaa Version: **5.17.0-20240904**
- Platform: **CentOS8, deployed using [https://github.com/IBM/StorageScaleVagrant](https://github.com/IBM/StorageScaleVagrant), Scale 5.2.2.0 240913.105913**

### Actual behavior

The following test code fails at `put_bucket_lifecycle_configuration()`:

```
    def test_s3cever13():
        client = get_client()
        bucket_name = get_new_bucket(client=client)
    
        # The following operation uses another client instance -> following operations will fail
    #    check_configure_versioning_retry(bucket_name, "Enabled", "Enabled")
        _ces_safe_check_configure_versioning_retry(client, bucket_name, "Enabled", "Enabled")
    
        key = "test"
        num_versions = 3
        create_multiple_versions(client, bucket_name, key, num_versions)
    
        rules= [{'Expiration': {'Days': 1},
              'ID': 'rule1',
              'Filter': { 'Prefix': 'content-0' },
              'Status':'Enabled'}]
        lifecycle = json.dumps({'Rules': rules})
        print(lifecycle)
>       client.put_bucket_lifecycle_configuration(Bucket=bucket_name, LifecycleConfiguration=lifecycle)
...
E               botocore.exceptions.ParamValidationError: Parameter validation failed:
E               Invalid type for parameter LifecycleConfiguration, value: {"Rules": [{"Expiration": {"Days": 1}, "ID": "rule1", "Filter": {"Prefix": "content-0"}, "Status": "Enabled"}]}, type: <class 'str'>, valid types: <class 'dict'>
```

### Expected behavior

The test case should pass, the policy should be accepted and `get_bucket_lifecycle_configuration()` should return the above policy.

### Steps to reproduce

Execute the above listed test code - added to Ceph s3-tests.

### More information - Screenshots / Logs / Other output

Policy printed with the `print(lifecycle)` command above:
```
{"Rules": [{"Expiration": {"Days": 1}, "ID": "rule1", "Filter": {"Prefix": "content-0"}, "Status": "Enabled"}]}
```

Nobaa log with debuglevel set to all attached
[noobaa_20240917_1.log.gz](https://github.com/user-attachments/files/17032212/noobaa_20240917_1.log.gz)

dannyzaken:
[...]

PR #8424, opened 2024-10-01
Lifecycle changes - Allow all actions in put lifecycle configuration

### LINKED ISSUE #8341: Bucket lifecycle rules should allow top level Prefix and NoncurrentVersionExpiration and AbortIncompleteMultipartUpload elements
### Environment info
- NooBaa Version: **5.16**
- Platform: **NC**

### Actual behavior
1. We currently reject rules with top level Prefix (outside the Filter), which is allowed on AWS.
2. We currently reject rules with actions NoncurrentVersionExpiration and AbortIncompleteMultipartUpload.

### Expected behavior
1. We need to allow these.
2. This is a followup to #8300

### Steps to reproduce
1. Try to put the following lifecycle policy file which is valid for AWS:

```json
{
    "Rules": [
        {
            "ID": "Expiration",
            "Status": "Enabled",
            "Prefix": "p1",
            "Expiration": { "Days": 1 },
            "NoncurrentVersionExpiration": { "NoncurrentDays": 2 },
            "AbortIncompleteMultipartUpload": { "DaysAfterInitiation": 10 }
        },
        {
            "ID": "NoncurrentVersionExpiration",
            "Status": "Enabled",
            "Filter": {
                "Prefix": "p2"
            },
            "NoncurrentVersionExpiration": { "NoncurrentDays": 2 }
        },
        {
            "ID": "AbortIncompleteMultipartUpload",
            "Status": "Enabled",
            "Filter": {
                "And": {
                    "Prefix": "p3",
                    "ObjectSizeGreaterThan": 11
                }
            },
            "AbortIncompleteMultipartUpload": { "DaysAfterInitiation": 10 }
        }
    ]
}
```

dannyzaken:
### Explain the changes
* although not all of the actions are implemented in the lifecycle bg worker, we still want to store the provided configuration and not fail it
* in NC case, the lifecycle work is performed outside of noobaa, so we want to provide all fields
* In addition, for now, allow passing the deprecated Prefix field. This commit leaves a gap regarding validation of the configuration, which should be more aligned with what AWS allows

1. Fixes #8341
2. #8376
3. opened Gap #8426

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
+++ b/src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
@@ -6,22 +6,22 @@ const { v4: uuid } = require('uuid');
 const dbg = require('../../../util/debug_module')(__filename);
 const S3Error = require('../s3_errors').S3Error;
 
+const true_regex = /true/i;
+
 // parse lifecycle rule filter
 function parse_filter(filter) {
     const current_rule_filter = {};
-    if (filter.Tag && filter.Tag.length === 1) {
+    if (filter.Tag?.length === 1) {
         const tag = filter.Tag[0];
         current_rule_filter.tags = [{ key: tag.Key[0], value: tag.Value[0] }];
     }
-    if (filter.Prefix && filter.Prefix.length === 1) {
+    if (filter.Prefix?.length === 1) {
         current_rule_filter.prefix = filter.Prefix[0];
     }
-    if (filter.ObjectSizeGreaterThan &&
-        filter.ObjectSizeGreaterThan.length === 1) {
+    if (filter.ObjectSizeGreaterThan?.length === 1) {
         current_rule_filter.object_size_greater_than = parseInt(filter.ObjectSizeGreaterThan[0], 10);
     }
-    if (filter.ObjectSizeLessThan &&
-        filter.ObjectSizeLessThan.length === 1) {
+    if (filter.ObjectSizeLessThan?.length === 1) {
         current_rule_filter.object_size_less_than = parseInt(filter.ObjectSizeLessThan[0], 10);
     }
     if (current_rule_filter.object_size_greater_than !== undefined &&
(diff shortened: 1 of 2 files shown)


src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -95,39 +101,54 @@ async function put_bucket_lifecycle(req) {
...
-            throw new S3Error(S3Error.NotImplemented);
+
+
+        if (rule.Expiration && rule.Expiration.length === 1) {
+            current_rule.expiration = parse_expiration(rule.Expiration[0]);
         }
-        if (rule.Transition) {
-            dbg.error('Transition is not implemented, rule:', rule);
-            throw new S3Error(S3Error.NotImplemented);
+        if (rule.AbortIncompleteMultipartUpload && rule.AbortIncompleteMultipartUpload.length === 1) {
```
romayalon, line 124: style, maybe worth in all the conditions make the conditions easier to read by using ? check - 
`if (rule?.AbortIncompleteMultipartUpload.length === 1)`

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -95,39 +101,54 @@ async function put_bucket_lifecycle(req) {
...
-            throw new S3Error(S3Error.NotImplemented);
-        }
-        current_rule.expiration = parse_expiration(rule.Expiration[0]);
 
-        if (rule.AbortIncompleteMultipartUpload) {
-            dbg.error('AbortIncompleteMultipartUpload is not implemented, rule:', rule);
-            throw new S3Error(S3Error.NotImplemented);
+
+
+        if (rule.Expiration && rule.Expiration.length === 1) {
```
romayalon, line 121: expiration is still required in the schema, we probably want to remove it if it's not required anymore
same for filter

romayalon (approved):

PR #8425, opened 2024-10-01
NC | NSFS | List Object Versions - Add Tests

shirady:
### Explain the changes
1. Add Tests of `ListObjectVersions` in NSFS.
Note: some of them are based on the expected example in the AWS documentation ([link](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectVersions.html#API_ListObjectVersions_Examples)).

### Issues: 
1. none

### Testing Instructions:
1. Please run: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace_versioning.js -g 'list object versions'`

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/test/unit_tests/test_bucketspace_versioning.js
+++ b/src/test/unit_tests/test_bucketspace_versioning.js
@@ -1,12 +1,14 @@
 /* Copyright (C) 2020 NooBaa */
 /* eslint-disable max-lines-per-function */
-/*eslint max-lines: ["error",3300]*/
+/*eslint max-lines: ["error",4500]*/
+/* eslint-disable max-statements */
 'use strict';
 
 const fs = require('fs');
 const path = require('path');
 const mocha = require('mocha');
 const assert = require('assert');
+const P = require('../../util/promise');
 const fs_utils = require('../../util/fs_utils');
 const nb_native = require('../../util/nb_native');
 const size_utils = require('../../util/size_utils');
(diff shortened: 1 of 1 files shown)


src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -3192,6 +3220,626 @@ mocha.describe('List-objects', function() {
         });
         assert.equal(count, 6);
     });
+
+    mocha.it('list object versions - no objects in the bucket (versioning enabled)', async function() {
+        await s3_client.putBucketVersioning({ Bucket: bucket_name2, VersioningConfiguration: { MFADelete: 'Disabled', Status: 'Enabled' } });
+        const res = await s3_client.listObjectVersions({ Bucket: bucket_name2 });
+        assert.equal(res.$metadata.httpStatusCode, 200);
```
romayalon, line 3227: I suggest checking the number of objects equals to 0

romayalon: same for the equivalent suspended case

romayalon (approved):

src/test/unit_tests/test_bucketspace_versioning.js
shirady: We don't have any versions yet (so the array is undefined), I added a check for that, thanks.

PR #8446, opened 2024-10-09
CI | Increase Timeouts

shirady:
### Explain the changes
1. Increase the default timeout in `test_bucketspace_versioning.js` which uses mocha (default is 2000 milliseconds), according to the [documentation](https://mochajs.org/#timeouts):
>   use this.timeout(0) to disable the timeout for a hook.
2. Increase the default timeout in `test_versioning_concurrency.test.js` which uses jest (default is 5000 milliseconds, see [here](https://jestjs.io/docs/api#testname-fn-timeout)) - I based it on [this](https://stackoverflow.com/questions/68811529/how-can-i-increase-the-test-time-out-value-in-jest).

1. We saw a couple of failed attempts in the CI related to before hook of list-objects ([here](https://github.com/noobaa/noobaa-core/actions/runs/11235732499/job/31243514489)):

>   2 failing
> 
>   1) bucketspace namespace_fs - versioning
>        List-objects
>          "before all" hook for "list objects - should return only latest object":
>      Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/root/node_modules/noobaa-core/src/test/unit_tests/nc_index.js)
>       at listOnTimeout (node:internal/timers:573:17)
>       at process.processTimers (node:internal/timers:514:7)
> 
>   2) List-objects
>        "before all" hook for "list objects - should return only latest object":
>      Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/root/node_modules/noobaa-core/src/test/unit_tests/nc_index.js)
>       at listOnTimeout (node:internal/timers:573:17)
>       at process.processTimers (node:internal/timers:514:7)

2. We want to avoid any timeouts in the  test_versioning_concurrency.

### Testing Instructions:
Please run:
1. `sudo npx jest test_versioning_concurrency.test.js`
2. `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace_versioning.js`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/test/unit_tests/jest_tests/test_versioning_concurrency.test.js
+++ b/src/test/unit_tests/jest_tests/test_versioning_concurrency.test.js
@@ -10,6 +10,8 @@ const buffer_utils = require('../../../util/buffer_utils');
 const { TMP_PATH } = require('../../system_tests/test_utils');
 const { crypto_random_string } = require('../../../util/string_utils');
 const endpoint_stats_collector = require('../../../sdk/endpoint_stats_collector');
+const SECONDS = 1000;
+const test_timeout = SECONDS * 60;
 
 function make_dummy_object_sdk(nsfs_config, uid, gid) {
     return {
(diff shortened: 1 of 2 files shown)


nadavMiz (approved):
LGTM

PR #8453, opened 2024-10-10
NC | NSFS | Versioning | Add Tests With Nested Keys (more than 1 level)

shirady:
### Explain the changes
1. Add Tests With Nested Keys (more than 1 level) in versioning Enabled bucket: put-object, get-object, head-object, delete-object, delete multiple objects, and list object versions.

1. none

### Testing Instructions:
1. Please run: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace_versioning.js`

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/test/unit_tests/test_bucketspace_versioning.js
+++ b/src/test/unit_tests/test_bucketspace_versioning.js
@@ -1,6 +1,6 @@
 /* Copyright (C) 2020 NooBaa */
 /* eslint-disable max-lines-per-function */
-/*eslint max-lines: ["error",4500]*/
+/*eslint max-lines: ["error",5500]*/
 /* eslint-disable max-statements */
 'use strict';
 
@@ -25,6 +25,7 @@ const XATTR_INTERNAL_NOOBAA_PREFIX = 'user.noobaa.';
 const XATTR_VERSION_ID = XATTR_INTERNAL_NOOBAA_PREFIX + 'version_id';
 const XATTR_DELETE_MARKER = XATTR_INTERNAL_NOOBAA_PREFIX + 'delete_marker';
 const NULL_VERSION_ID = 'null';
+const HIDDEN_VERSIONS_PATH = '.versions';
 
 const DEFAULT_FS_CONFIG = get_process_fs_context();
 let CORETEST_ENDPOINT;
@@ -39,6 +40,7 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
     const bucket_name = 'versioned-enabled-bucket';
     const disabled_bucket_name = 'disabled-bucket'; // be aware that this bucket would become versioned in the copy object tests
     const suspended_bucket_name = 'suspended-bucket';
+    const nested_keys_bucket_name = 'bucket-with-nested-keys';
 
     const bucket_path = '/bucket';
     const full_path = tmp_fs_root + bucket_path;
@@ -46,6 +48,8 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
     const disabled_full_path = tmp_fs_root + disabled_bucket_path;
     const suspended_bucket_path = '/suspended_bucket';
     const suspended_full_path = tmp_fs_root + suspended_bucket_path;
+    const nested_keys_bucket_path = '/bucket_with_nested_keys';
+    const nested_keys_full_path = path.join(tmp_fs_root, nested_keys_bucket_path);
     const versions_path = path.join(full_path, '.versions/');
     const suspended_versions_path = path.join(suspended_full_path, '.versions/');
     let s3_uid5;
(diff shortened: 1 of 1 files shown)


src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -283,11 +306,25 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
...
             });
+
+            mocha.it('put object - versioning enabled - nested key (more than 1 level)', async function() {
+                await s3_uid6.putBucketVersioning({ Bucket: nested_keys_bucket_name, VersioningConfiguration: { MFADelete: 'Disabled', Status: 'Enabled' } });
+                await s3_uid6.putObject({ Bucket: nested_keys_bucket_name, Key: nested_key_level3, Body: body1 });
+                // only after the second PUT object we expect to have the .versions under the parent directory of the file
+                await s3_uid6.putObject({ Bucket: nested_keys_bucket_name, Key: nested_key_level3, Body: body2});
+                await s3_uid6.putObject({ Bucket: nested_keys_bucket_name, Key: nested_key_level3, Body: body3});
+                const version_path_nested = path.join(nested_keys_full_path, dir_path_nested, HIDDEN_VERSIONS_PATH);
+                const exist = await fs_utils.file_exists(version_path_nested);
```
romayalon, line 325: I see this check is repeating on all tests, is this the only thing we want to check here?

src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -682,6 +722,24 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
             const is_dm = await is_delete_marker(full_path, '', key1, max_version2);
             assert.ok(is_dm);
         });
+
+        mocha.it('delete object - versioning enabled - nested key (more than 1 level)', async function() {
+            const res = await s3_uid6.deleteObject({ Bucket: nested_keys_bucket_name, Key: nested_key_level3 });
+            assert.equal(res.DeleteMarker, true);
+            const exist = await fs_utils.file_not_exists(path.join(nested_keys_full_path, nested_key_level3,));
```
romayalon, line 729:
```suggestion
            const exist = await fs_utils.file_not_exists(path.join(nested_keys_full_path, nested_key_level3));
```

src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -1907,6 +1965,36 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
...
+
+            // with version-id
+            const delete_res2 = await account_with_access.deleteObjects({
+                Bucket: delete_multi_object_test_bucket, Delete: { Objects: arr_keys_and_version_id } });
+            assert.equal(delete_res2.Deleted.length, num_of_versions);
+
+            // delete the delete marker of latest version
+            await s3_uid6.deleteObject({ Bucket: delete_multi_object_test_bucket,
+                Key: delete_res.Deleted[0].Key, VersionId: delete_res.Deleted[0].DeleteMarkerVersionId});
+        });
```
romayalon, line 1997: I would add here a check of the fs directly using list object versions, so we can validate that the object was deleted completely not only through the responses

src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -2635,6 +2774,36 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
...
+            assert.equal(head_res.ContentLength, body3.length);
+
+            // by version-id
+            const head_res2 = await s3_client.headObject({
+                Bucket: nested_keys_bucket_name, Key: nested_key_level4, VersionId: put_res2.VersionId});
+            assert.equal(head_res2.ContentLength, body2.length);
+
+            // partial path of (directory without the key)
+            try {
+                await s3_client.headObject({Bucket: nested_keys_bucket_name, Key: dir_path_nested});
```
romayalon, line 2800: I suggest start using `assert.rejects`, I think it would cleanup our code

romayalon (approved):

src/test/unit_tests/test_bucketspace_versioning.js
shirady: It would be: `await assert.rejects(async () => await s3_client.headObject({Bucket: nested_keys_bucket_name, Key: dir_path_nested}));` instead of the try..catch clause, although it is one-liner, it is less flexible to debugging in my opinion.

src/test/unit_tests/test_bucketspace_versioning.js
romayalon: I think it's better to check that the version ids resulted from the putObject responses are really there, and not only having .versions/ dir in the path.

src/test/unit_tests/test_bucketspace_versioning.js
romayalon: per our discussion on slack, I don't think there are missing debugging features, here are some examples for debugging - https://www.geeksforgeeks.org/node-js-assert-rejects-function/
but it's a matter of styling

src/test/unit_tests/test_bucketspace_versioning.js
shirady: I added a check on the version names in the .version/ directory to match the version id in the putObject response.

src/test/unit_tests/test_bucketspace_versioning.js
shirady: Added, thanks.

src/test/unit_tests/test_bucketspace_versioning.js
shirady: At this point I didn't change the code, maybe in the future I will try it again.

PR #8455, opened 2024-10-10
NSFS | versioning | use linkat in gpfs safe_link to not override existing versions

nadavMiz:
### Explain the changes
1. add option to fail if target file exist in gpfs linkat function using posix linkat
2. add CAP_DAC_READ_SEARCH capability when changing uid/gid so user to permit use of linkat for non-root user (see ENOENT in  https://man7.org/linux/man-pages/man2/link.2.html)
3. get latest version from fd opened using open_files_gpfs rather then from path to avoid concurrency issues with mismatch version between the version we got and the actual version of the file. change open_gpfs to not use version_id

### Testing Instructions:
1. on gpfs machine run: `GPFS_ROOT_PATH=/ibm/fs1/tmp multiple puts of the same key'`
2. run following ceph_test: `S3TEST_CONF=s3tests3.conf tox -- s3tests_boto3/functional/test_s3.py::test_versioned_concurrent_object_create_concurrent_remove`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -559,6 +559,7 @@ struct FSWorker : public Napi::AsyncWorker
     int _warn_threshold_ms;
     double _took_time;
     Napi::FunctionReference _report_fs_stats;
+    bool _should_add_thread_capabilities;
 
     // executes the ctime check in the stat and read file fuctions
     // NOTE: If _do_ctime_check = false, then some functions will fallback to using mtime check
@@ -574,6 +575,7 @@ struct FSWorker : public Napi::AsyncWorker
         , _errno(0)
         , _warn_threshold_ms(0)
         , _took_time(0)
+        , _should_add_thread_capabilities(false)
         , _do_ctime_check(false)
     {
         for (int i = 0; i < (int)info.Length(); ++i) _args_ref.Set(i, info[i]);
@@ -607,6 +609,9 @@ struct FSWorker : public Napi::AsyncWorker
         tx.set_user(_uid, _gid);
         DBG1("FS::FSWorker::Execute: " << _desc << DVAL(_uid) << DVAL(_gid) << DVAL(geteuid()) << DVAL(getegid()) << DVAL(getuid()) << DVAL(getgid()));
 
+        if(_should_add_thread_capabilities) {
+            tx.add_thread_capabilities();
+        }
         auto start_time = std::chrono::high_resolution_clock::now();
         Work();
         auto end_time = std::chrono::high_resolution_clock::now();
@@ -640,6 +645,9 @@ struct FSWorker : public Napi::AsyncWorker
     bool use_gpfs_lib()
     {
         return gpfs_dl_path != NULL && gpfs_lib_file_exists > -1 && _backend == GPFS_BACKEND;
+    }
+    void AddThreadCapabilities() {
+        _should_add_thread_capabilities = true;
     }
     virtual void OnOK() override
     {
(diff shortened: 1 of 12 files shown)


src/util/native_fs_utils.js
```diff
@@ -252,9 +248,12 @@ async function unlink_ignore_enoent(fs_context, to_delete_path) {
...
 // safe_link_gpfs links source_path to dest_path while verifing dest.fd
-async function safe_link_gpfs(fs_context, dst_path, src_file, dst_file) {
-    dbg.log1('Namespace_fs.safe_link_gpfs source_file:', src_file, src_file.fd, dst_file, dst_file && dst_file.fd);
-    await src_file.linkfileat(fs_context, dst_path, dst_file && dst_file.fd);
+async function safe_link_gpfs(fs_context, dst_path, src_file, dst_file, should_override) {
+    dbg.log1('Namespace_fs.safe_link_gpfs source_file:', src_file, src_file.fd, dst_file, dst_file && dst_file.fd, should_override);
+    if (!should_override && dst_file) {
+        dbg.warn("safe_link_gpfs: both should_override is false and dest file is set, ignoring dst_file");
+    }
+    await src_file.linkfileat(fs_context, dst_path, should_override ? (dst_file && dst_file.fd) : -2);
```
romayalon, line 256: is there a situation where fd will be undefined and we will pass undefined?

src/native/util/os_linux.cpp
```diff
@@ -64,6 +65,27 @@ ThreadScope::restore_user()
...
 }
 
+int
+ThreadScope::set_capabilities_linkat() {
+    cap_t caps = cap_get_proc();
+    if(caps == NULL) {
+        return -1;
+    }
+    cap_value_t newcaps[1] = { CAP_DAC_READ_SEARCH, };
+    if(cap_set_flag(caps, CAP_EFFECTIVE, 1, newcaps, CAP_SET) < 0) {
```
guymguym, line 75: check if already set with cap_get_flag()

src/native/fs/fs_napi.cpp
```diff
@@ -1638,6 +1643,11 @@ struct LinkFileAt : public FSWrapWorker<FileWrap>
         // Linux will fail the linkat() if the file already exist and we want to replace it if it existed.
         if (_replace_fd >= 0) {
             SYSCALL_OR_RETURN(dlsym_gpfs_linkatif(fd, "", AT_FDCWD, _filepath.c_str(), AT_EMPTY_PATH, _replace_fd));
+        } else if (_should_not_override){
+            //set thread capabilities to allow linkat from user other than root.
+            ThreadScope tx;
+            SYSCALL_OR_RETURN(tx.set_capabilities_linkat());
```
guymguym, line 1649: Move to method on FSWorker::AddThreadCapabilities()

src/native/util/os.h
```diff
@@ -38,6 +38,8 @@ class ThreadScope
         change_user();
     }
 
+    int set_capabilities_linkat();
```
guymguym, line 41: add_thread_capabilities

src/native/util/os_darwin.cpp
```diff
@@ -54,6 +54,13 @@ ThreadScope::restore_user()
     }
 }
 
+int
+ThreadScope::set_capabilities_linkat()
+{
+    //set capabilities not used in darwin
```
guymguym, line 60: add LOG("unsupported...")

src/native/util/os_linux.cpp
```diff
@@ -64,6 +65,27 @@ ThreadScope::restore_user()
     }
 }
 
+int
+ThreadScope::set_capabilities_linkat() {
+    cap_t caps = cap_get_proc();
+    if(caps == NULL) {
+        return -1;
```
guymguym, line 72: add LOG before every return -1

romayalon (approved):

PR #8456, opened 2024-10-10
[NSFS] Fix Newline Reader to work with partial reads and improve its memory usage

tangledbytes:
### Explain the changes
This PR intends to fix the parsing issue present in our NewLine reader. Previously, the reader would perform partial reads (4096 bytes) and would force covert it into string, this works most of the times but falls apart if there was a UTF8 character at the boundary of the 4096 buffer.

This PR removes any such conversions while scanning through the buffer and also ensures to efficiently use the buffer, doesn't creates unnecessary strings and buffers and rather tries to rely on a pre-allocated buffers.

1. A DBS3 issue

### Testing Instructions:
1. `./node_modules/.bin/jest src/test/unit_tests/jest_tests/test_newline_reader.test.js`

- [ ] Doc added/updated
- [ ] Tests added


src/util/file_reader.js
```diff
@@ -21,20 +21,38 @@ class NewlineReader {
...
+     *  return_leftover?: boolean;
+     *  warn_too_long?: boolean;
+     * }} [cfg]
+     **/
+    constructor(fs_context, filepath, cfg) {
         this.path = filepath;
-        this.lock = lock;
-        this.buf = Buffer.alloc(4096);
+        this.lock = cfg?.lock;
+        this.return_leftover = Boolean(cfg?.return_leftover);
```
guymguym, line 34: I think the default should be to return the leftover, so we can rename this to "skip_leftover_line" which will be default false, but will be set to true for the persistent logger use case.

src/util/file_reader.js
```diff
@@ -21,20 +21,38 @@ class NewlineReader {
...
+        this.warn_too_long = Boolean(cfg?.warn_too_long);
```
guymguym, line 35: I think this should be renamed to "skip_overflow_lines" with default false, but will be set to true for the persistent logger use case.

src/util/file_reader.js
```diff
@@ -44,29 +63,66 @@ class NewlineReader {
...
+            if (!read) {
+                this.eof = true;
 
-        // readresults will contain >= 1 result or else we would have kept looping above
-        this.readresults = this._partialread.split('\n').slice(0, -1);
+                // what to do with the leftover in the buffer on eof
+                if (this.end > this.start) {
+                    if (this.skip_leftover_line) {
+                        console.warn("leftover at eof:", this.info());
+                    } else {
```
guymguym, line 113: we need to check also the overflow state here and not return it as a line
```suggestion
                    } else if (this.overflow_state) {
                        console.warn('line too long finally terminated at eof:', this.info());
                    } else {
```

guymguym (commented):
found a small fix

guymguym (approved):

PR #8467, opened 2024-10-15
NC | Online upgrade process small fix & refactoring

romayalon:
### Explain the changes
1. fix -  
    a. changed init_nc_system() to set config_directory properties only on a completely new system. otherwise, config_directory data will be set only via nc_upgrade_manager. Due to this change, I had to check config_directory data exists also on throw_if_config_dir_locked(), and if not, this is an old system.json and we do not allow updating config directory in that case, hence we will throw an error.
3. refactoring - 
    a. Moved init_nc_system from nsfs.js to config_fs.js, re-arranged the function and added some helper functions.
    b. Added missing JSdocs.

1. Open question - should we allow using the CLI without having system.json?
Wondering if there can be scenarios where the user wants to create a bucket/account without starting the S3 service ever, if so, we don't have any version information about the already existing accounts/buckets, we won't know if updates on the config directory should be allowed or not in that case.  - optional solution, create system.json via the CLI.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/nsfs.js
+++ b/src/cmd/nsfs.js
@@ -41,7 +41,6 @@ const BucketSpaceFS = require('../sdk/bucketspace_fs');
 const SensitiveString = require('../util/sensitive_string');
 const endpoint_stats_collector = require('../sdk/endpoint_stats_collector');
 //const { RPC_BUFFERS } = require('../rpc');
-const pkg = require('../../package.json');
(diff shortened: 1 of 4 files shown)


nadavMiz (approved):
LGTM

PR #8474, opened 2024-10-19
nsfs | wait for endpoint startup before namespace monitor registration

alphaprinz:
### Explain the changes
Wait for endpoint startup before registering namespace resource monitor.

Nsr can enter "Rejected" status if endpoint is deleted by kubernetes before it is in "Ready" state.
https://bugzilla.redhat.com/show_bug.cgi?id=2284585.

### Testing Instructions:
Repoduction
Sagie details scenario for ODS in bz.

I've reproduced with this scenario on minikube:
Start with a nsfs nsr on a pvc. A single endpoint A is in "Ready" state.
-Delete nsr. A new endpoint B is being spun.
-While endpoint B is done creating but NOT ready yet (endpoint A is still in "Ready" state), create nsr.
-Kubernetes will delete endpoint B and will leave endpoint A running.
-Endpoint B loads nsr from system store, but nsr is not mounted on it. Endpoint B issues NO_SUCH_BUCKET report on nsr (and then it is deleted by kubernetes).

- [ ] Doc added/updated
- [ ] Tests added
[local_nsfs.yaml.txt](https://github.com/user-attachments/files/17442342/local_nsfs.yaml.txt)

### DIFF
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -231,12 +231,15 @@ async function main(options = {}) {
         if (internal_rpc_client && config.NAMESPACE_MONITOR_ENABLED) {
             endpoint_stats_collector.instance().set_rpc_client(internal_rpc_client);
 
-            // Register a bg monitor on the endpoint
-            background_scheduler.register_bg_worker(new NamespaceMonitor({
-                name: 'namespace_fs_monitor',
-                client: internal_rpc_client,
-                should_monitor: nsr => Boolean(nsr.nsfs_config),
-            }));
+            //wait with monitoring until pod has started
+            setTimeout(() => {
+                // Register a bg monitor on the endpoint
+                background_scheduler.register_bg_worker(new NamespaceMonitor({
+                    name: 'namespace_fs_monitor',
+                    client: internal_rpc_client,
+                    should_monitor: nsr => Boolean(nsr.nsfs_config),
+                }));
+            }, 1000 * 60);
         }
 
         if (config.ENABLE_SEMAPHORE_MONITOR) {


src/endpoint/endpoint.js
```diff
@@ -231,12 +231,14 @@ async function main(options = {}) {
...
-                should_monitor: nsr => Boolean(nsr.nsfs_config),
-            }));
+            setTimeout(() => {
+                // Register a bg monitor on the endpoint
+                background_scheduler.register_bg_worker(new NamespaceMonitor({
+                    name: 'namespace_fs_monitor',
+                    client: internal_rpc_client,
+                    should_monitor: nsr => Boolean(nsr.nsfs_config),
+                }));
+            }, 1000 * 60);
```
shirady, line 241: I think you can add a comment why we decide to use a delay (and specifically this delay).

src/endpoint/endpoint.js
```diff
@@ -231,12 +231,14 @@ async function main(options = {}) {
         if (internal_rpc_client && config.NAMESPACE_MONITOR_ENABLED) {
             endpoint_stats_collector.instance().set_rpc_client(internal_rpc_client);
 
-            // Register a bg monitor on the endpoint
-            background_scheduler.register_bg_worker(new NamespaceMonitor({
-                name: 'namespace_fs_monitor',
-                client: internal_rpc_client,
-                should_monitor: nsr => Boolean(nsr.nsfs_config),
-            }));
+            setTimeout(() => {
```
romayalon, line 234: why not add a retry in the namespace monitor if the error was ENOENT and start time of the endpoint is lower than 60 seconds? also why 60 seconds?

alphaprinz: Sounds like more code with equivalent solution.

60 seconds to allow pod to stabilize (get to 'ready' or be deleted).

romayalon: the reason is understood, but how do you know it takes a minute? is it always a minute? should it be configurable?

alphaprinz: AFAIK, there's no way for node to know status of the pod.
I assume we don't want to make such a dependency.

60 seconds is more than enough on my minikube and is low enough not to bother other deployments in other envs, but I can make it an env variable.

romayalon (approved):

PR #8475, opened 2024-10-19
bucket_api: Pagination support for list_bucket API

aspandey:
### Explain the changes
This change provides support for the pagination of list_buckets.

GET /?continuation-token=ContinuationToken&max-buckets=MaxBuckets HTTP/1.1
Host: s3.amazonaws.com

Important links - 
https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html

https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listBuckets-property

We can call list buckets using API like this-
nb api bucket_api list_buckets '{
  "max_buckets" : 10                                       
}'

Or using S3 call like - 

const input = {
                MaxBuckets: 20
            };
const command = new ListBucketsCommand(input);

### DIFF
--- a/src/api/bucket_api.js
+++ b/src/api/bucket_api.js
@@ -441,6 +441,17 @@ module.exports = {
 
         list_buckets: {
             method: 'GET',
+            params: {
+                type: 'object',
+                properties: {
+                    continuation_token: { $ref: 'common_api#/definitions/continuation_token' },
+                    max_buckets: {
+                        type: 'integer',
+                        minimum: 1,
+                        maximum: 1000
+                    }
+                }
+            },
             reply: {
                 type: 'object',
                 required: ['buckets'],
@@ -454,10 +465,11 @@ module.exports = {
                                 name: { $ref: 'common_api#/definitions/bucket_name' },
                                 creation_date: {
                                     idate: true
-                                },
+                                }
                             }
                         }
-                    }
+                    },
+                    continuation_token: { $ref: 'common_api#/definitions/continuation_token' }
                 }
             },
             auth: {
--- a/src/endpoint/s3/ops/s3_get_service.js
+++ b/src/endpoint/s3/ops/s3_get_service.js
@@ -7,8 +7,17 @@ const s3_utils = require('../s3_utils');
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTServiceGET.html
  */
 async function list_buckets(req) {
-    const reply = await req.object_sdk.list_buckets();
+
+    const params = {
+        continuation_token: s3_utils.cont_tok_to_key_marker(req.query['continuation-token']),
+        max_buckets: req.query['max-buckets'] ? Number(req.query['max-buckets']) : undefined
+    };
+
+    const reply = await req.object_sdk.list_buckets(params);
     const date = s3_utils.format_s3_xml_date(new Date());
+    const bucket_cont_token = s3_utils.key_marker_to_cont_tok(reply.continuation_token, null, false);
+    // const bucket_cont_token = bucket_name_to_cont_token(reply.continuation_token);
+
     return {
         ListAllMyBucketsResult: {
             Owner: s3_utils.DEFAULT_S3_USER,
@@ -17,7 +26,8 @@ async function list_buckets(req) {
                     Name: bucket.name.unwrap(),
                     CreationDate: bucket.creation_date ? s3_utils.format_s3_xml_date(bucket.creation_date) : date,
                 }
-            }))
+            })),
+            ContinuationToken: bucket_cont_token,
         }
     };
 }
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -999,19 +999,60 @@ async function delete_bucket_lifecycle(req) {
  * LIST_BUCKETS
  *
  */
+
 async function list_buckets(req) {
-    const buckets_by_name = _.filter(
-        req.system.buckets_by_name,
+
+    let next_index = 0;
+    let is_truncated = false;
+
+    let continuation_token = req.rpc_params?.continuation_token;
+    const max_buckets = req.rpc_params?.max_buckets;
+
+    const accessible_bucket_list = system_store.data.buckets.filter(
         async bucket => await req.has_s3_bucket_permission(bucket, "s3:ListBucket", req) && !bucket.deleting
     );
-    return {
-        buckets: _.map(buckets_by_name, function(bucket) {
-            return {
-                name: bucket.name,
-                creation_date: bucket._id.getTimestamp().getTime()
-            };
-        })
-    };
+
+    accessible_bucket_list.sort((a, b) => a.name.unwrap().localeCompare(b.name.unwrap()));
+
+    if (!max_buckets) {
+        const buckets = accessible_bucket_list.map(b => ({
+            name: b.name,
+            creation_date: b._id.getTimestamp().getTime()
+        }));
+        return {
+            buckets,
+        };
+    }
+
+    if (continuation_token) {
+        const index = accessible_bucket_list.findIndex(
+            bucket => bucket.name.unwrap().localeCompare(continuation_token.unwrap()) >= 0
+        );
+        if (index !== -1) {
+            next_index = index + 1;
+        }
+    }
+
+    const paged_bucket_list = accessible_bucket_list.slice(next_index, next_index + max_buckets);
+
+    const buckets = paged_bucket_list.map(bucket => ({
+        name: bucket.name,
+        creation_date: bucket._id.getTimestamp().getTime()
+    }));
+
+    is_truncated = accessible_bucket_list.length > next_index + max_buckets;
+    continuation_token = is_truncated ? accessible_bucket_list[next_index + max_buckets - 1].name : undefined;
+
+    if (continuation_token) {
+        return {
+            buckets,
+            continuation_token,
+        };
+    } else {
+        return {
+            buckets,
+        };
+    }
 }
 
 
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -1210,6 +1210,10 @@ module.exports = {
             wrapper: SensitiveString,
         },
 
+        continuation_token: {
+            wrapper: SensitiveString,
+        },
+
         port: {
             type: 'integer',
             minimum: 0,
--- a/src/endpoint/s3/ops/s3_get_bucket.js
+++ b/src/endpoint/s3/ops/s3_get_bucket.js
@@ -43,7 +43,7 @@ async function get_bucket(req) {
         delimiter: req.query.delimiter,
         limit: Math.min(max_keys_received, 1000),
         key_marker: list_type === '2' ?
-            (cont_tok_to_key_marker(cont_tok) || start_after) : req.query.marker,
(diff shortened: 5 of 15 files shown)


src/server/system_services/bucket_server.js
```diff
@@ -972,19 +972,65 @@ async function delete_bucket_lifecycle(req) {
...
+    if (!max_buckets) {
+        const buckets = _.map(accessible_bucket_list, b => ({
+            name: b.name,
+            creation_date: b._id.getTimestamp().getTime()
+        }));
+        return {
+            buckets,
+        };
+    }
+    accessible_bucket_list.sort((a, b) => a._id.getTimestamp().getTime() - b._id.getTimestamp().getTime());
```
dannyzaken, line 1003: If I remember correctly, the sort order should be lexicographic and not by creation date

src/server/system_services/bucket_server.js
```diff
@@ -972,19 +972,65 @@ async function delete_bucket_lifecycle(req) {
...
+    if (continuation_token) {
+        let i = 0;
+        for (i = 0; i < accessible_bucket_list.length; i++) {
+            if (accessible_bucket_list[i]._id.toString() === continuation_token) {
```
dannyzaken, line 1007: If the sort order is indeed lexicographic, then `_id` is not a good fit for a continuation token. better to encode the bucket name into the continuation token, similar to how we do it in list objects (see [here](https://github.com/dannyzaken/noobaa-core/blob/7dfe555d046cb4fffed90d04a860da194cf35f3f/src/endpoint/s3/ops/s3_get_bucket.js#L99-L117))

Also, you should not search for the exact match of the continuation token, since the bucket can be deleted between calls to list_buckets.

dannyzaken (changes requested):

src/server/system_services/bucket_server.js
aspandey: @dannyzaken We did talk about sorting  in lexicographical order but there are few cases which will not be covered.
1 - if there are few buckets like following 
a
aa
b
bb
c
cc
d
dd

Consider max length = 2
so after first call the continuation token would be "b" and max token would be 2. now in next call it will return "b" and "bb". Let say between these two call user created some buckets like "aaa" "aa1" "aaaa3". Then these buckets will never be listed in subsequent calls.

However, if we sort it with creation time then it will list all the above created buckets as the creation time for these buckets are latest. and we are sorting in increasing order of creation time.

I think this case is a valid test case to avoid using lexicographical order. As we are not saving the sorted order anywhere and not using it in any other usage, this would not break any other part.

dannyzaken: The problem you are trying to solve is not a requirement. 
I think we should be as consistent as possible with S3. The bucket list is always dynamic, and buckets are deleted and created constantly. You should also consider the application's needs—I believe that a lexicographic order is more what the UI expects to get from the server.

src/endpoint/s3/ops/s3_get_service.js
```diff
@@ -2,12 +2,19 @@
...
  */
 async function list_buckets(req) {
-    const reply = await req.object_sdk.list_buckets();
+
+    const params = {
+        continuation_token: req.query['continuation-token'],
+        max_buckets: req.query['max-buckets'] ? Number(req.query['max-buckets']) : undefined
+    };
+
+    dbg.error("params.max_buckets:", params.max_buckets, "params.continuation_token:", params.continuation_token);
```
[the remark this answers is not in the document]
aspandey: done.

src/server/system_services/bucket_server.js
```diff
@@ -972,19 +972,65 @@ async function delete_bucket_lifecycle(req) {
...
+    const accessible_bucket_list = [];
+    const paged_bucket_list = [];
+    let next_index = 0;
+    let is_truncated = false;
+    const all_buckets = _.values(req.system.buckets_by_name);
+
+    let continuation_token = req.rpc_params?.continuation_token;
+    const max_buckets = req.rpc_params?.max_buckets;
+
+    dbg.error("max_buckets:", max_buckets, "continuation_token:", continuation_token);
```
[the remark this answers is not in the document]
aspandey: done

src/server/system_services/bucket_server.js
aspandey: done.

src/server/system_services/bucket_server.js
aspandey: done

src/server/system_services/bucket_server.js
```diff
@@ -972,19 +972,69 @@ async function delete_bucket_lifecycle(req) {
...
+            }
+        }
+        next_index = i;
+    }
+
+    for (let i = next_index; (i < accessible_bucket_list.length && i < next_index + max_buckets); i++) {
+            paged_bucket_list.push(accessible_bucket_list[i]);
+    }
+
+    const buckets = _.map(paged_bucket_list, bucket => ({
```
dannyzaken, line 1021: prefer using native functions when possible 


```suggestion
    const buckets = paged_bucket_list.map(bucket => ({
```

src/server/system_services/bucket_server.js
```diff
@@ -972,19 +972,69 @@ async function delete_bucket_lifecycle(req) {
...
+        name: bucket.name,
+        creation_date: bucket._id.getTimestamp().getTime()
+    }));
+
+    is_truncated = accessible_bucket_list.length > next_index + max_buckets;
+    continuation_token = is_truncated ? accessible_bucket_list[next_index + max_buckets].name : undefined;
```
dannyzaken, line 1027: use the last bucket as a continuation token instead of the next one. That way you will "catch" new buckets that are created between list_buckets calls

src/api/bucket_api.js
```diff
@@ -388,6 +388,19 @@ module.exports = {
 
         list_buckets: {
             method: 'GET',
+            params: {
+                type: 'object',
+                properties: {
+                    continuation_token: {
+                        type: 'string'
```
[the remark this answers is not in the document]
aspandey: done

src/endpoint/s3/ops/s3_get_service.js
```diff
@@ -2,13 +2,22 @@
...
-    const reply = await req.object_sdk.list_buckets();
+
+    const params = {
+        continuation_token: cont_token_to_bucket_name(req.query['continuation-token']),
+        max_buckets: req.query['max-buckets'] ? Number(req.query['max-buckets']) : undefined
+    };
+
+    const reply = await req.object_sdk.list_buckets(params);
     const date = s3_utils.format_s3_xml_date(new Date());
+    const bucket_cont_token = bucket_name_to_cont_token(reply.continuation_token ?? undefined);
```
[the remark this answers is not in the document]
aspandey: done

src/server/system_services/bucket_server.js
```diff
@@ -972,19 +972,69 @@ async function delete_bucket_lifecycle(req) {
...
+    const all_buckets = _.values(req.system.buckets_by_name);
+
+    let continuation_token = req.rpc_params?.continuation_token;
+    const max_buckets = req.rpc_params?.max_buckets;
+
+    for (const bucket of all_buckets) {
+        if (await req.has_s3_bucket_permission(bucket, "s3:ListBucket", req) && !bucket.deleting) {
+            accessible_bucket_list.push(bucket);
+        }
+    }
```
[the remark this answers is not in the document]
aspandey: done.

src/server/system_services/bucket_server.js
```diff
@@ -972,19 +972,69 @@ async function delete_bucket_lifecycle(req) {
...
+        let i = 0;
+        let name = '';
+        for (i = 0; i < accessible_bucket_list.length; i++) {
+            name = accessible_bucket_list[i].name.unwrap();
+            if (name === continuation_token || name > continuation_token) {
+                break;
+            }
+        }
+        next_index = i;
+    }
```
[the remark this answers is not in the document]
aspandey: done

src/server/system_services/bucket_server.js
```diff
@@ -972,19 +972,69 @@ async function delete_bucket_lifecycle(req) {
...
+
+    for (let i = next_index; (i < accessible_bucket_list.length && i < next_index + max_buckets); i++) {
+            paged_bucket_list.push(accessible_bucket_list[i]);
+    }
```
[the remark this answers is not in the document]
aspandey: done

src/server/system_services/bucket_server.js
aspandey: done

src/server/system_services/bucket_server.js
aspandey: I have made the changes to use last bucket pf previous set as continuation token. 
However, I can not imagine how it would it catch the buckets created in between the calls.

src/sdk/config_fs.js
```diff
@@ -912,7 +912,7 @@ class ConfigFS {
      * list_buckets returns the array of buckets that exists under the config dir
      * @returns {Promise<string[]>} 
      */
-    async list_buckets() {
+    async list_buckets(params) {
```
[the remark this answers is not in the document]
aspandey: done

src/endpoint/s3/ops/s3_get_service.js
```diff
@@ -17,11 +26,32 @@ async function list_buckets(req) {
...
+        throw new S3Error(S3Error.InvalidArgument);
+    }
+}
+
+function bucket_name_to_cont_token(bucket_name) {
+    if (!bucket_name) return;
+    const next_marker = bucket_name;
+    const j = JSON.stringify({ key: next_marker });
+    return Buffer.from(j).toString('base64');
+}
```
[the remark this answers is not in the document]
aspandey: done

src/server/system_services/bucket_server.js
```diff
@@ -972,19 +972,65 @@ async function delete_bucket_lifecycle(req) {
...
+    let continuation_token = req.rpc_params?.continuation_token;
+    const max_buckets = req.rpc_params?.max_buckets;
+
+    const accessible_bucket_list = (
+        await Promise.all(
+            system_store.data.buckets.map(async bucket => (
+                (await req.has_s3_bucket_permission(bucket, "s3:ListBucket", req) && !bucket.deleting) ? bucket : null
+            ))
+        )
+    ).filter(Boolean);
```
dannyzaken, line 990: Why are you running it with `Promise.all`? it is possible to have a **very large** number of buckets.  This can lead to noobaa-core getting stuck or crash

below suggestion should do it serially (check for correctness)

```suggestion
    const accessible_bucket_list = system_store.data.buckets.filter(
        async bucket => await req.has_s3_bucket_permission(bucket, "s3:ListBucket", req) && !bucket.deleting
    );
```

aspandey: Done.

src/server/system_services/bucket_server.js
dannyzaken: Let' say you have a bucket list: `bucket1, bucket2, ... bucket1000, bucket2000...`
in the first call you return the list `bucket1..bucket1000`. if you use `bucket2000` as the continuation token, any new bucket between `bucket1000 .. bucket2000` will not be included in the next call

dannyzaken (approved):

Issue #8288, opened 2024-08-18
NC | NSFS | Bucket Policy Should Be Managed Only by Bucket Owner


shirady:
### Environment info
- NooBaa Version: master (5.18)
- Platform: NC

### Actual behavior
1. A user with a bucket policy (to all actions) can add or remove a policy although he is not the owner.

### Expected behavior
1. Only the owner can add or remove the bucket policy, as I understand it from the docs (see [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-policies.html)):

> Only the bucket owner can associate a policy with a bucket.

### Steps to reproduce
Create 2 accounts and give the account2 bucket policy full access. Then, try to put the bucket policy by account2 and remove the bucket policy.
1. Create account1 with the CLI: `sudo node src/cmd/manage_nsfs account add --name shira-1001 --new_buckets_path /tmp/nsfs_root1 --access_key <access-key> --secret_key <secret-key> --uid <uid> --gid <gid>`
Create account2 with the CLI: `sudo node src/cmd/manage_nsfs account add --name shira-1002 --new_buckets_path /tmp/nsfs_root2 --access_key <access-key> --secret_key <secret-key> --uid <uid> --gid <gid>`
Note: before creating the account need to give permission to the `new_buckets_path`: `chmod 777 /tmp/nsfs_root2`.
2. Create a bucket owned by account1 with the CLI: `sudo node src/cmd/manage_nsfs bucket add --name my-bucket --path /tmp/nsfs_root1/my-bucket --owner shira-1001` 
3. Start the NSFS server with: `sudo node src/cmd/nsfs --debug 5`
Notes: 
- Before starting the server please add this line: `process.env.NOOBAA_LOG_LEVEL = 'nsfs';` in the endpoint.js (before the condition `if (process.env.NOOBAA_LOG_LEVEL) {`)
- I Change the `config.NSFS_CHECK_BUCKET_BOUNDARIES = false; //SDSD` because I'm using the `/tmp/` and not `/private/tmp/`.
4. Create the alias for S3 service: `alias nc-user-1-s3='AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443'`.
[...]

PR #8477, opened 2024-10-22
NSFS | Bucket Policy Should Be Managed Only by Bucket Owner

naveenpaul1:
### Explain the changes
1. add bucket owner id and account id validation for both nsfs and containerized deployment(bucket_server)

1. https://github.com/noobaa/noobaa-core/issues/8288

### Testing Instructions:
1. please find it in issue

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/endpoint/s3/s3_bucket_policy_utils.js
+++ b/src/endpoint/s3/s3_bucket_policy_utils.js
@@ -281,6 +281,19 @@ async function validate_s3_policy(policy, bucket_name, get_account_handler) {
     }
 }
 
+/**
+ * CRUD action on bucket policy only allow to bucket owner.
+ * @param {string} bucket_owner_id
+ * @param {string} account_id
+ */
+async function validate_s3_policy_owner(bucket_owner_id, account_id) {
+    dbg.log0('validate_s3_policy_owner : bucket_owner_id ', bucket_owner_id, 'account_id : ', account_id);
+    if (bucket_owner_id !== account_id) {
+        throw new RpcError('METHOD_NOT_ALLOWED');
+    }
+}
+
 exports.OP_NAME_TO_ACTION = OP_NAME_TO_ACTION;
 exports.has_bucket_policy_permission = has_bucket_policy_permission;
 exports.validate_s3_policy = validate_s3_policy;
+exports.validate_s3_policy_owner = validate_s3_policy_owner;
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -479,6 +479,7 @@ async function put_bucket_encryption(req) {
 async function get_bucket_policy(req) {
     dbg.log0('get_bucket_policy:', req.rpc_params);
     const bucket = find_bucket(req, req.rpc_params.name);
+    await bucket_policy_utils.validate_s3_policy_owner(bucket.owner_account._id, req.account._id);
     return {
         policy: bucket.s3_policy,
     };
@@ -488,6 +489,7 @@ async function get_bucket_policy(req) {
 async function put_bucket_policy(req) {
     dbg.log0('put_bucket_policy:', req.rpc_params);
     const bucket = find_bucket(req, req.rpc_params.name);
+    await bucket_policy_utils.validate_s3_policy_owner(bucket.owner_account._id, req.account._id);
     await bucket_policy_utils.validate_s3_policy(req.rpc_params.policy, bucket.name,
         principal => system_store.get_account_by_email(principal));
     await system_store.make_changes({
@@ -505,6 +507,7 @@ async function put_bucket_policy(req) {
 async function delete_bucket_policy(req) {
     dbg.log0('delete_bucket_policy:', req.rpc_params);
     const bucket = find_bucket(req, req.rpc_params.name);
+    await bucket_policy_utils.validate_s3_policy_owner(bucket.owner_account._id, req.account._id);
     await system_store.make_changes({
         update: {
             buckets: [{
--- a/src/test/unit_tests/test_bucketspace_fs.js
+++ b/src/test/unit_tests/test_bucketspace_fs.js
@@ -32,6 +32,7 @@ const XATTR_DELETE_MARKER = XATTR_INTERNAL_NOOBAA_PREFIX + 'delete_marker';
 
 const test_bucket = 'bucket1';
 const test_bucket2 = 'bucket2';
+const test_bucket3 = 'bucket3';
 const test_not_empty_bucket = 'notemptybucket';
 const test_bucket_temp_dir = 'buckettempdir';
 const test_bucket_invalid = 'bucket_invalid';
@@ -148,7 +149,7 @@ const dummy_ns = {
 function make_dummy_object_sdk() {
     return {
         requesting_account: {
-            _id: '65b3c68b59ab67b16f98c26e',
+            _id: account_user2._id,
             force_md5_etag: false,
             name: new SensitiveString('user2'),
             email: new SensitiveString('user2@noobaa.io'),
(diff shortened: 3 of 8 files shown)


src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -281,6 +281,17 @@ async function validate_s3_policy(policy, bucket_name, get_account_handler) {
     }
 }
 
+/*
+    CRUD action on bucket policy only allow to bucket owner 
+*/
+async function validate_s3_policy_owner(bucket_owner_id, account) {
+    if (bucket_owner_id !== account._id) {
+        const err_msg = 'The specified method is not allowed against this resource.';
```
shirady, line 289: I think you can remove this `err_msg`, the RPC call `METHOD_NOT_ALLOWED` is mapped to `S3Error.MethodNotAllowed` and you have this message inside, see:
https://github.com/noobaa/noobaa-core/blob/bfaa477af56a995a6fc0238f5a41e668f43840d6/src/endpoint/s3/s3_errors.js#L612
and:
https://github.com/noobaa/noobaa-core/blob/bfaa477af56a995a6fc0238f5a41e668f43840d6/src/endpoint/s3/s3_errors.js#L292-L296

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -281,6 +281,17 @@ async function validate_s3_policy(policy, bucket_name, get_account_handler) {
     }
 }
 
+/*
+    CRUD action on bucket policy only allow to bucket owner 
+*/
+async function validate_s3_policy_owner(bucket_owner_id, account) {
```
shirady, line 287: Could you please add JSDoc to the new function?

shirady, line 287: I think you can pass the account_id instead of the whole account object.

src/server/system_services/bucket_server.js
```diff
@@ -479,6 +479,7 @@ async function put_bucket_encryption(req) {
 async function get_bucket_policy(req) {
     dbg.log0('get_bucket_policy:', req.rpc_params);
     const bucket = find_bucket(req, req.rpc_params.name);
+    await bucket_policy_utils.validate_s3_policy_owner(bucket.owner_account._id, req.account);
```
shirady, line 482: In case the function `validate_s3_policy_owner` arguments are strings I think you will need to convert those when passing them (in case you change the `account` to `account_id` as well).

src/test/unit_tests/test_bucketspace_fs.js
```diff
@@ -954,6 +955,65 @@ mocha.describe('bucketspace_fs', function() {
...
+    mocha.describe('bucket policy ownership validations', function() {
+        mocha.before(async function() {
+            const param = { name: test_bucket3 };
+            await bucketspace_fs.create_bucket(param, dummy_object_sdk);
+        });
+        mocha.after(async function() {
+            await fs_utils.folder_delete(`${new_buckets_path}/${test_bucket}`);
+            const file_path = get_config_file_path(CONFIG_SUBDIRS.BUCKETS, test_bucket3);
+            await fs_utils.file_delete(file_path);
+        });
```
shirady, line 968: Could you please a new line here? (after the `mocha.after` and before the test `mocha.it`)

src/test/unit_tests/test_bucketspace_fs.js
```diff
@@ -954,6 +955,65 @@ mocha.describe('bucketspace_fs', function() {
...
+        mocha.it('get bucket policy not allowed non owner user', async function() {
+            const policy = generate_s3_policy(account_user1.name, test_bucket3, ['s3:*']).policy;
+            const param = { name: test_bucket3, policy: policy };
+            await bucketspace_fs.put_bucket_policy(param, dummy_object_sdk);
+            const res = await bucketspace_fs.get_bucket_policy(param, dummy_object_sdk);
+            assert_bucket_policies(policy, res.policy);
```
shirady, line 974: I think you can move these lines to the `mocha.before`, they are repeated in all the tests (`it`) under the `describe`.

src/test/unit_tests/test_bucketspace_fs.js
```diff
@@ -954,6 +955,65 @@ mocha.describe('bucketspace_fs', function() {
...
+            const dummy_object_sdk_not_owner = make_dummy_object_sdk_for_account(dummy_object_sdk, account_user1);
+            try {
+                await bucketspace_fs.delete_bucket_policy(param, dummy_object_sdk_not_owner);
+                assert.fail('should have failed with METHOD_NOT_ALLOWED');
+            } catch (err) {
+                assert.ok(err.rpc_code === 'METHOD_NOT_ALLOWED');
+            }
+            await bucketspace_fs.delete_bucket_policy(param, dummy_object_sdk);
+            const resp = await bucketspace_fs.get_bucket_policy(param, dummy_object_sdk);
+            assert_bucket_policies(undefined, resp.policy);
```
shirady, line 1014: I would add a comment that this step is for cleaning the bucket policy.

src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -221,7 +221,8 @@ mocha.describe('s3_bucket_policy', function() {
             Bucket: BKT,
             Policy: JSON.stringify(policy)
         });
-        const res_a = await s3_a.getBucketPolicy({ // should work - user a has get_bucket_policy permission
+        // Bucket Policy Should Be Managed Only by Bucket Owner : #8477
```
shirady, line 224: I'm not sure we want to leave the issue link in our code (#8477).

shirady (commented):
I added a few small comments

src/server/system_services/bucket_server.js
naveenpaul1: I am not sure about it, I could see lots of places without string conversion and its working there

src/test/unit_tests/test_bucketspace_fs.js
naveenpaul1: lint error

src/test/unit_tests/test_bucketspace_fs.js
naveenpaul1: its more like testing the delete bucket policy working for owner or not, not exactly a clean up

src/test/unit_tests/test_s3_bucket_policy.js
naveenpaul1: added this comment because of the removed comment, I can remove the issue id

src/server/system_services/bucket_server.js
shirady: As I understand it the _id origin is from Mongo in which it was an object, so one of the ways for that comparison is converting the ID to string and then use the `===` operator if you want (anyway, it will also match the type of strings that you defined in the JSDoc.
reference I saw:
https://stackoverflow.com/questions/11637353/comparing-mongoose-id-and-strings

src/test/unit_tests/test_bucketspace_fs.js
shirady: Good, thanks.
I think it can be helpful if you would add a comment that this part was to test that the deletion of the bucket policy works for the bucket owner (in the test title "delete bucket policy not allowed non owner user").

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -281,6 +281,19 @@ async function validate_s3_policy(policy, bucket_name, get_account_handler) {
     }
 }
 
+/**
+ * CRUD action on bucket policy only allow to bucket owner.
+ * @param {string} bucket_owner_id - bucket owner id
+ * @param {string} account_id - account id
```
shirady, line 287: I think you can remove the details, the type and the variable name to give us the needed information.
```suggestion
 * @param {string} bucket_owner_id
 * @param {string} account_id
```

src/test/unit_tests/test_upgrade_scripts.js
```diff
@@ -31,12 +31,13 @@ mocha.describe('test upgrade scripts', async function() {
...
         await system_store.load();
 
-        const account_info = await rpc_client.account.read_account({ email: EMAIL, });
+        const account_info = await rpc_client.account.read_account({ email: EMAIL });
         s3 = new S3({
             endpoint: coretest.get_http_address(),
             credentials: {
                 accessKeyId: account_info.access_keys[0].access_key.unwrap(),
                 secretAccessKey: account_info.access_keys[0].secret_key.unwrap(),
+                accountId: account_info._id
```
shirady, line 40: I don't think you need to add the `accountId` as a part of the S3 client in the `credentials` part.

naveenpaul1:
closing this PR after Nimrod and Danny confirm this could create issues for ODF installations that we already have. 
```
Currently, the operator account is actually the bucket owner for all OBC-created buckets. 
So if we only allow the bucket owner to change the bucket policy
```

PR #8478, opened 2024-10-23
Set default bucket encryption during bucket creation

vh05:
All S3 buckets have encryption configured by default, and objects are automatically encrypted by using server side encryption. When we do get-bucker-encryption on any bucket we get the the default encryption configuration.

With this patch we set default encryption on bucket while creating the bucket and follow the behavior of S3 bucket

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2318715

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -73,6 +73,9 @@ function new_bucket_defaults(name, system_id, tiering_policy_id, owner_account_i
         object_lock_configuration: config.WORM_ENABLED ? {
             object_lock_enabled: lock_enabled ? 'Enabled' : 'Disabled',
         } : undefined,
+        encryption: {
+            "algorithm": "AES256",
+        },
     };
 }
 
--- a/src/test/system_tests/ceph_s3_tests/s3-tests-lists/s3_tests_pending_list.txt
+++ b/src/test/system_tests/ceph_s3_tests/s3-tests-lists/s3_tests_pending_list.txt
@@ -141,3 +141,5 @@ s3tests_boto3/functional/test_s3.py::test_object_lock_delete_multipart_object_wi
 s3tests_boto3/functional/test_s3.py::test_object_lock_delete_multipart_object_with_legal_hold_on
 s3tests_boto3/functional/test_s3.py::test_get_undefined_public_block
 s3tests_boto3/functional/test_s3.py::test_get_public_block_deny_bucket_policy
+s3tests_boto3/functional/test_s3.py::test_get_bucket_encryption_s3
+s3tests_boto3/functional/test_s3.py::test_get_bucket_encryption_kms
--- a/src/test/unit_tests/test_s3_encryption.js
+++ b/src/test/unit_tests/test_s3_encryption.js
@@ -71,14 +71,22 @@ mocha.describe('Bucket Encryption Operations', async () => {
         await local_s3.createBucket({ Bucket: BKT });
     });
 
-    mocha.it('should get bucket encryption error without encryption configured', async () => {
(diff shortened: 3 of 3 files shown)


src/endpoint/s3/ops/s3_put_bucket.js
```diff
@@ -9,6 +9,15 @@ async function put_bucket(req, res) {
     const lock_enabled = config.WORM_ENABLED ? req.headers['x-amz-bucket-object-lock-enabled'] &&
         req.headers['x-amz-bucket-object-lock-enabled'].toUpperCase() === 'TRUE' : undefined;
     await req.object_sdk.create_bucket({ name: req.params.bucket, lock_enabled: lock_enabled });
+
+    // Set default server side bucket encryption
+    // More details: https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucket-encryption.html
+    await req.object_sdk.put_bucket_encryption({
```
dannyzaken, line 15: Should this be set for any bucket, even namespace buckets (S3, blob)? It's worth checking the current support for bucket encryption in NS buckets and mapping the gaps. I think that if there are any gaps, we should at least make sure we are not setting a configuration that might conflict with the target configuration (e.g. S3 bucket)

vh05: Tested it on aws-s3 namespace bucket. The bucket created out of aws-s3 namespace, is not applied with default server side encryption. Only the parent bucket is encrypted.

The bucket created here is through obc. Do we need to apply the encryption here as well and check ?

nimrod-becker: For NS S3, since we don't create the bucket on AWS itself, we should rely on the default of AWS, which should be exactly this unless someone else has set it.
When receiving putBucketEncryption for NS S3 we should simply pass it through to the target bucket and return the result, same for getBucketEncryption, simply return it from the target.

Azure, we might want to skip if it's not the same for now.

For NS FS, I would also not touch for now. Especially not as a default.

That leaves our data buckets, for them, we encrypt all the objects anyway, it's part of the pipeline. So we simply want to set this config so we can return it on a get operation.

@dannyzaken @vh05 wdyt?

vh05: Okay. For NS S3 we will show the underlying bucket's encryption configuration.

nimrod-becker:
@jackyalbo @dannyzaken to look as well

src/server/system_services/bucket_server.js
```diff
@@ -73,6 +73,9 @@ function new_bucket_defaults(name, system_id, tiering_policy_id, owner_account_i
         object_lock_configuration: config.WORM_ENABLED ? {
             object_lock_enabled: lock_enabled ? 'Enabled' : 'Disabled',
         } : undefined,
+        encryption: {
```
jackyalbo, line 76: The problem with setting it like this here is that it also applies for namespace buckets - which we don't know for sure if they have encryption set or not. So I think you need to check here if it's data buckets only (no namespace). For namespace that can be a bigger discussion on what we should return, as it can be a mix of resources of different kinds. Would handling only data buckets be enough for now?

nimrod-becker:
What about the missing BucketKeyEnabled ?

vh05:
> What about the missing BucketKeyEnabled ?

aws documentation for `get-bucket-encryption (`https://docs.aws.amazon.com/cli/latest/reference/s3api/get-bucket-encryption.html ), the o/p does not explicitly provide `bucketEnabled` information as o/p. Should we include that as `default as false` ?

nimrod-becker:
Superseded by #8716

PR #8480, opened 2024-10-25
NSFS | Versioning | Delete of partial directory of nested key results in AccessDeniedError

### LINKED ISSUE #8454: NC | NSFS | Versioning | Delete of partial directory of nested key results in `AccessDeniedError`
### Environment info
- NooBaa Version: 5.18.0 (current master)
- Platform: NC

### Actual behavior
1. When a bucket with versioning enabled and there is a nested key inside and the client tries to delete the parent directory of the file without the last slash '/' (let's say it was by accident) will result in an error `AccessDenied` and in FS layer it is `Error: Operation not permitted, code: 'EPERM'`

### Expected behavior
1. Create a delete marker without an error.

### Steps to reproduce

1. Create an account with the CLI: `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path /tmp/nsfs_root1 --access_key <access-key> --secret_key <secret-key> --uid <uid> --gid <gid>`
Note: before creating the account need to give permission to the `new_buckets_path`: `chmod 777 /tmp/nsfs_root1`, `chmod 777 /tmp/nsfs_root2`.
Notes: 
- I Change the `config.NSFS_CHECK_BUCKET_BOUNDARIES = false; //SDSD` because I’m using the `/tmp/` and not `/private/tmp/`.
5. Add bucket to the account using AWS CLI: `nc-user-1-s3 s3 mb s3://bucket-v` (`bucket-v` is the bucket name in this example)
6. Enable versioning: `nc-user-1-s3 s3api put-bucket-versioning --bucket bucket-v --versioning-configuration Status=Enabled`
7. Put a nested  object: `nc-user-1-s3 s3api put-object --bucket bucket-v --key /a/b/c/lala.txt`
8. Delete the key /a/b/c (without the last slash): `nc-user-1-s3 s3api delete-object --bucket bucket-v --key  /a/b/c`
Note: deleting the directory creates the delete marker as expected (`nc-user-1-s3 s3api delete-object --bucket bucket-v --key /a/b/c/`).

Might be that this issue is with the same root cause of the issue - #8320

```
Oct-10 16:13:38.960 [nsfs/79233]  [WARN] core.sdk.namespace_fs:: NamespaceFS._delete_latest_version error: retries=10 latest_ver_path=/tmp/nsfs_root1/bucket-v/a/b/c [Error: Operation not permitted] { code: 'EPERM', context: 'SafeLink _link_from.c_str()=/tmp/nsfs_root1/bucket-v/a/b/c _link_to.c_str()=/tmp/nsfs_root1/bucket-v/a/b/.versions/c_null _link_expected_mtime=1728566008914826496 _link_expected_inode=152582414 ' }
Oct-10 16:13:38.961 [nsfs/79233] [ERROR] core.endpoint.s3.s3_rest:: S3 ERROR <?xml version="1.0" encoding="UTF-8"?><Error><Code>AccessDenied</Code><Message>Access Denied</Message><Resource>/bucket-v//a/b/c</Resource><RequestId>m23bksck-csss5c-135r</RequestId></Error> DELETE /bucket-v//a/b/c {"host":"localhost:6443","accept-encoding":"identity","user-agent":"aws-cli/2.17.11 md/awscrt#0.20.11 ua/2.0 os/macos#24.0.0 md/arch#arm64 lang/python#3.11.9 md/pyimpl#CPython cfg/retry-mode#standard md/installer#source md/prompt#off md/command#s3api.delete-object","x-amz-date":"20241010T131338Z","x-amz-content-sha256":"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855","authorization":"AWS4-HMAC-SHA256 Credential=<>/20241010/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=4dc8ae840909447ab17ef08ce5c6d25c62a422ed1ba7fa79900a210ac1cfa3b5","content-length":"0"} Error: Operation not permitted - context: SafeLink _link_from.c_str()=/tmp/nsfs_root1/bucket-v/a/b/c _link_to.c_str()=/tmp/nsfs_root1/bucket-v/a/b/.versions/c_null _link_expected_mtime=1728566008914826496 _link_expected_inode=152582414
```

naveenpaul1:
### Explain the changes
1. Version bucket does not support dir versioning, Deleting dir key(`a/b/c/`) just deletes the .folder file and does not create delete_marker.
2.  If the bucket has key `a/b/c/test.txt` and then tries to delete key `a/b/c` code will throw `AccessDenied` error.

1. https://github.com/noobaa/noobaa-core/issues/8454

### Testing Instructions:
1. Please check the issue

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1926,7 +1926,10 @@ class NamespaceFS {
             await this._check_path_in_bucket_boundaries(fs_context, file_path);
             dbg.log0('NamespaceFS: delete_object', file_path);
             let res;
-            if (this._is_versioning_disabled()) {
+            const is_key_dir_path = await this._is_key_dir_path(fs_context, params.key);
+            if (this._is_versioning_disabled() || is_key_dir_path) {
+                // TODO- Directory object (key/) is currently can't co-exist while key (without slash) exists. see -https://github.com/noobaa/noobaa-core/issues/8320
+                // Also, Directory object (key/) is currently not supported combined with versioning - see - https://github.com/noobaa/noobaa-core/issues/8531
                 await this._delete_single_object(fs_context, file_path, params);
             } else {
                 res = params.version_id ?
@@ -2783,6 +2786,23 @@ class NamespaceFS {
 
         const versioned_path = this._get_version_path(key, version_id);
         return versioned_path;
+    }
+    /** 
+    * _is_key_dir_path will check if key is pointing to a directory or a file
+    * @param {nb.NativeFSContext} fs_context
+    * @param {string} key
+    * @returns {Promise<boolean>}
+    */
+    async _is_key_dir_path(fs_context, key) {
+        try {
+            const key_path = path.normalize(path.join(this.bucket_path, key));
+            const key_stat = await nb_native().fs.stat(fs_context, key_path, { skip_user_xattr: true });
+            const is_dir = native_fs_utils.isDirectory(key_stat);
+            return is_dir;
+        } catch (err) {
+            dbg.warn('NamespaceFS._is_key_dir_path : error while getting state for key ', key, err);
+        }
+        return false;
     }
 
     _throw_if_delete_marker(stat, params) {
--- a/src/test/unit_tests/test_namespace_fs.js
+++ b/src/test/unit_tests/test_namespace_fs.js
@@ -1,6 +1,6 @@
 /* Copyright (C) 2020 NooBaa */
 /*eslint max-lines-per-function: ["error", 900]*/
-/*eslint max-lines: ["error", 2100]*/
+/*eslint max-lines: ["error", 2200]*/
 'use strict';
 
 const _ = require('lodash');
@@ -426,6 +426,7 @@ mocha.describe('namespace_fs', function() {
 
         const dir_1 = '/a/b/c/';
         const dir_2 = '/a/b/';
+        const dir_3 = '/x/y/z/';
         const upload_key_1 = dir_1 + 'upload_key_1';
         const upload_key_2 = dir_1 + 'upload_key_2';
         const upload_key_3 = dir_2 + 'upload_key_3';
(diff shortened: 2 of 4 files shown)


src/test/unit_tests/test_namespace_fs.js
```diff
@@ -472,6 +474,22 @@ mocha.describe('namespace_fs', function() {
 
         });
 
+        mocha.it('delete dir object without last slash - a/b/c', async function() {
```
shirady, line 477: minor (missing the first slash in the title)
```suggestion
        mocha.it('delete dir object without last slash - /a/b/c', async function() {
```

src/sdk/namespace_fs.js
```diff
@@ -2778,6 +2779,24 @@ class NamespaceFS {
...
     }
+    /** 
+    * method append slash at the end of dir key when key misses last slash.
+    * @param {import('./nb').NativeFSContext} fs_context
+    * @param {string} key
+    */
+    async _get_correct_key_path(fs_context, key) {
+        try {
+            const key_path = path.normalize(path.join(this.bucket_path, key));
+            const key_state = await nb_native().fs.stat(fs_context, key_path, { skip_user_xattr: true });
```
shirady, line 2790: Is it `key_state` or `key_stat`? (we are running a stat on the key)

shirady (commented):
I added minor comments

shirady:
@naveenpaul1 could you add the test in file `src/test/unit_tests/test_bucketspace_versioning.js`?
I found this issue while I worked on PR #8453

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -472,6 +474,22 @@ mocha.describe('namespace_fs', function() {
...
+            await upload_object(ns_tmp, upload_bkt, upload_key_4, dummy_object_sdk, source);
+            await delete_object(ns_tmp, upload_bkt, '/x/y/z', dummy_object_sdk);
+
+            let entries;
+            try {
+                entries = await nb_native().fs.readdir(DEFAULT_FS_CONFIG, ns_tmp_bucket_path + dir_2);
+            } catch (e) {
+                assert.ifError(e);
+            }
+            console.log('stop when not empty - entries', entries);
```
shirady, line 488: Do we need this printing in the CI?
I'm not sure I understand it, could you please a comment with an explnantion?

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -472,6 +474,22 @@ mocha.describe('namespace_fs', function() {
 
         });
 
+        mocha.it('delete dir object without last slash - /x/y/z/', async function() {
+            const source = buffer_utils.buffer_to_read_stream(data);
+            await upload_object(ns_tmp, upload_bkt, upload_key_4, dummy_object_sdk, source);
+            await delete_object(ns_tmp, upload_bkt, '/x/y/z', dummy_object_sdk);
```
shirady, line 480: I think it would be easier to read with additional variable:
```suggestion
           const partial_dir_3 = dir_3.slice(0, -1); // the path without the last slash
            await delete_object(ns_tmp, upload_bkt, partial_dir_3, dummy_object_sdk);
```

shirady (approved):
LGTM
added 2 minor comments

romayalon (changes requested):
I want to request changes because the original issue seems incorrect to me, can you please check that throwing AccessDenied is also not thrown by AWS in the specified reproduction steps?

naveenpaul1:
@romayalon added the AWS same behaviour finding in [issue](https://github.com/noobaa/noobaa-core/issues/8454#issuecomment-2454316675) , For AWS aslo partial dir key is creating delete_marker for object.
cc: @shirady

romayalon:
@naveenpaul1 thanks for checking it, 
can you run get_object before deleting the object?
my assumption is that you would receive ENOENT, because there is no such object called a/b/c.. later please check on NooBaa if this is still the behavior with the code of this PR
When deleting an object when versioning is enabled, AWS don't care if the object exists or not, they just create a delete marker.

naveenpaul1:
@romayalon updated the get-object behavior in [issue](https://github.com/noobaa/noobaa-core/issues/8454#issuecomment-2454595083) please check

src/sdk/namespace_fs.js
```diff
@@ -2772,6 +2772,7 @@ class NamespaceFS {
      * @returns {Promise<string>}
      */
     async _find_version_path(fs_context, { key, version_id }, return_md_path) {
+        key = await this._get_correct_key_path(fs_context, key);
```
romayalon, line 2775: can you explain this change please?

naveenpaul1: @romayalon if the key is a partial dir path such as `a/b/c` this method will check whether the key is dir or not if its dir, then a backslash is added to the key to avoid the `AccessDeniedError` error. file path will leave key as it is

romayalon: But notice that we expect to fail with ENOENT when key doesn't end with a slash, and the folder is not a content dir - 
```
if (!stat.xattr?.[XATTR_DIR_CONTENT] || !params.key.endsWith('/')) {
throw error_utils.new_error_code('ENOENT', 'NoSuchKey');
```
if you change the key to have a / at the end, will we fail on this situation as we should per AWS?

src/sdk/namespace_fs.js
```diff
@@ -3017,8 +3033,12 @@ class NamespaceFS {
         const is_gpfs = native_fs_utils._is_gpfs(fs_context);
         let retries = config.NSFS_RENAME_RETRIES;
         let latest_ver_info;
+        const is_key_dir_path = await this._is_key_dir_path(fs_context, params.key);
         for (;;) {
             try {
+                if (is_key_dir_path && !this._is_directory_content(latest_ver_path, params.key)) {
```
romayalon, line 3039: @naveenpaul1 @shirady Have you checked also what happens when trying to delete a/b/c/?

romayalon: we probably want to avoid versioning flow on this case as well, right?

src/sdk/namespace_fs.js
```diff
@@ -2784,6 +2784,22 @@ class NamespaceFS {
...
+    * @param {string} key
+    */
+    async _is_key_dir_path(fs_context, key) {
+        try {
+            const key_path = path.normalize(path.join(this.bucket_path, key));
+            const key_stat = await nb_native().fs.stat(fs_context, key_path, { skip_user_xattr: true });
+            const is_dir = native_fs_utils.isDirectory(key_stat);
+            return is_dir;
+        } catch (err) {
+            dbg.warn('NamespaceFS._correct_key_path : error while getting state for key ', key, err);
```
romayalon, line 2799:
```suggestion
            dbg.warn('NamespaceFS. _is_key_dir_path : error while getting state for key ', key, err);
```

src/sdk/namespace_fs.js
shirady: I don't remember that I checked this case - delete a/b/c without the beginning slash instead of /ab/c/ (I thought only about deleting/a/b/c without the ending slash instead of /a/b/c/ as mentioned in the issue).

naveenpaul1: @romayalon @shirady for`a/b/c/`, with this code we are just deleting the `.folder` file not creating any delet_marker,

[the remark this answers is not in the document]
naveenpaul1: @romayalon we are returning empty response for deleted scenario also, if we do the same for error case also it will create confusion.

romayalon: @naveenpaul1 can you point me to the code that does that?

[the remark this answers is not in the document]
romayalon: maybe it'll create a confusion, but in any case access denied is not an appropriate error code

naveenpaul1: @romayalon here : [link](https://github.com/noobaa/noobaa-core/pull/8480/files#diff-74a1674f40effb24da2459330c005cd5ab9fda4e1bca500a4387607926a2af1fR3058) we are skipping copy file to new version when customer try to delete the dir. and else block get executed which do simple deletion and e are also skipping delete_marker creation if is_key_dir_path true.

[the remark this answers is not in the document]
naveenpaul1: @romayalon even if we dont throw error here try catch block will throw a `EPERM` error and clent will convert the error to `AccessDenied` error,
response after removing the code that throws S3Error.AccessDenied error 
```
naveenpaul:noobaa-core$ s3 s3api delete-object --bucket bucket1 --key /a/b/c
An error occurred (AccessDenied) when calling the DeleteObject operation: Access Denied
```

naveenpaul1: @romayalon we are deleting the `.folder` file when user pass key `a/b/c/` we are not creating the version and delete_marker

src/sdk/namespace_fs.js
```diff
@@ -1926,7 +1926,9 @@ class NamespaceFS {
             await this._check_path_in_bucket_boundaries(fs_context, file_path);
             dbg.log0('NamespaceFS: delete_object', file_path);
             let res;
-            if (this._is_versioning_disabled()) {
+            const is_key_dir_path = await this._is_key_dir_path(fs_context, params.key);
+            if (this._is_versioning_disabled() || is_key_dir_path) {
+                // TODO : Dir versioning not implemented, This should be fixed as part of #8320
```
romayalon, line 1931:
```suggestion
            // TODO- Directory object (key/) is currently can't co-exist while key (without slash) exists. see -https://github.com/noobaa/noobaa-core/issues/8320
            // Also, Directory object (key/) is currently not supported combined with versioning - see - <add link>
            if (this._is_versioning_disabled() || is_key_dir_path) {
```

please mention the 2 issues here - https://github.com/noobaa/noobaa-core/issues/8531

src/sdk/namespace_fs.js
```diff
@@ -3035,7 +3054,7 @@ class NamespaceFS {
                     const suspended_and_latest_is_not_null = this._is_versioning_suspended() &&
                         latest_ver_info.version_id_str !== NULL_VERSION_ID;
                     const bucket_tmp_dir_path = this.get_bucket_tmpdir_full_path();
-                    if (this._is_versioning_enabled() || suspended_and_latest_is_not_null) {
+                    if ((this._is_versioning_enabled() || suspended_and_latest_is_not_null)) {
```
romayalon, line 3057: left overs from old code

romayalon (approved):

PR #8482, opened 2024-10-27
Avoid dedup chunks that are relatively new

dannyzaken:
### Explain the changes
1. To prevent deduplicating chunks from a previous failed upload attempt, only consider deduping chunks that are older than a specific threshold.
2. This is only a partial fix for the more general case. It only addresses the issue of failed upload retries. There is still a race condition between chunk deduplication and the object reclaimer flow.

1. https://bugzilla.redhat.com/show_bug.cgi?id=2256223

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


liranmauda (approved):
LGTM

PR #8484, opened 2024-10-28
Add support for configurable max bucket & key length

tangledbytes:
### Explain the changes
This PR adds support for configurable max key length in NooBaa. Amazon S3 supports maximum of 1024 bytes long key name while NooBaa does not have any such limits. This sometimes become problematic hence this PR adds support for a config variable named `S3_MAX_KEY_LENGTH` which defaults to `Infinity` for now implying no limit, however this can be changed to 1024 in `config-local.js`.

Some known DBS3 issues.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -362,6 +362,14 @@ function get_bucket_and_key(req) {
             key = suffix;
         }
     }
+
+    if (key?.length > config.S3_MAX_KEY_LENGTH) {
+        throw new S3Error(S3Error.KeyTooLongError);
+    }
+    if (bucket?.length > config.S3_MAX_BUCKET_NAME_LENGTH) {
+        throw new S3Error(S3Error.InvalidBucketName);
+    }
+
     return {
         bucket,
         // decode and replace hadoop _$folder$ in key
--- a/config.js
+++ b/config.js
@@ -198,6 +198,22 @@ config.S3_RESTORE_REQUEST_MAX_DAYS = 30;
  */
 config.S3_RESTORE_REQUEST_MAX_DAYS_BEHAVIOUR = 'TRUNCATE';
 
+/**
+ * S3_MAX_KEY_LENGTH controls the maximum key length that will be accepted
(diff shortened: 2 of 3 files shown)


src/endpoint/sts/sts_rest.js
```diff
@@ -90,6 +90,8 @@ async function handle_request(req, res) {
     }
     req.op_name = op_name;
 
+    http_utils.validate_bucket_name_object_key_length(req);
```
guymguym, line 93: I don't think this is relevant to sts/iam requests at all - we don't parse a bucket/key params in those servers.

src/endpoint/iam/iam_rest.js
```diff
@@ -105,6 +105,8 @@ async function handle_request(req, res) {
     }
     req.op_name = op_name;
 
+    http_utils.validate_bucket_name_object_key_length(req);
```
guymguym, line 108: I don't think this is relevant to sts/iam requests at all - we don't parse a bucket/key params in those servers.

src/endpoint/s3/s3_rest.js
```diff
@@ -121,6 +121,8 @@ async function handle_request(req, res) {
     }
     req.op_name = op_name;
 
+    http_utils.validate_bucket_name_object_key_length(req);
```
guymguym, line 124: it seems more relevant in my opinion that this validation will be part of get_bucket_and_key()

src/util/http_utils.js
```diff
@@ -702,6 +702,15 @@ function validate_server_ip_whitelist(req) {
     throw new S3Error(S3Error.AccessDenied);
 }
 
+function validate_bucket_name_object_key_length(req) {
```
guymguym, line 705: I don't think this fits http_utils, it can just be in s3_rest, or if we need to share it then maybe in s3_utils

guymguym (changes requested):

guymguym (approved):

PR #8497, opened 2024-10-31
notifications | unit tests

alphaprinz:
### Explain the changes
1. Add unit tests for notifications

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/test/unit_tests/index.js
+++ b/src/test/unit_tests/index.js
@@ -98,6 +98,7 @@ require('./test_tiering_ttl_worker');
 // require('./test_tiering_upload');
 //require('./test_s3_worm');
 require('./test_bucket_logging');
+require('./test_notifications');
 
 // UPGRADE
 // require('./test_postgres_upgrade'); // TODO currently working with mongo -> once changing to postgres - need to uncomment  
--- a/src/test/unit_tests/test_notifications.js
+++ b/src/test/unit_tests/test_notifications.js
@@ -1,0 +1,206 @@
+/* Copyright (C) 2025 NooBaa */
+'use strict';
+
+// disabling init_rand_seed as it takes longer than the actual test execution
+process.env.DISABLE_INIT_RANDOM_SEED = "true";
+
+const fs = require('fs');
+const path = require('path');
+
+const config = require('../../../config');
+
+// setup coretest first to prepare the env
+const { get_coretest_path, generate_s3_client, TMP_PATH } = require('../system_tests/test_utils');
+const coretest_path = get_coretest_path();
+const coretest = require(coretest_path);
+
+const { rpc_client, EMAIL, POOL_LIST } = coretest;
+
+console.log('POOL_LIST =', POOL_LIST);
+
+const tmp_path = path.join(TMP_PATH, 'test_notifications');
+//dir for connect files
+const tmp_connect = path.join(tmp_path, 'connect');
+//dir for notification persistent files
+const notif_logs_path = path.join(tmp_path, 'notif_logs');
+
+coretest.setup({pools_to_create: [POOL_LIST[1]]});
+const mocha = require('mocha');
+const http = require('http');
+const assert = require('assert');
+const timer = require('node:timers/promises');
+const notifications_util = require('../../util/notifications_util');
+
+const http_connect_filename = 'http_connect.json';
+const http_connect_path = path.join(tmp_connect, http_connect_filename);
+//content of connect file, will be written to a file in before()
+const http_connect = {
+    agent_request_object: {"host": "localhost", "port": 9998, "timeout": 1500},
+    request_options_object: {"auth": "amit:passw", "timeout": 1500},
+    notification_protocol: 'http',
+    name: 'http_notif'
+};
+
+const bucket = 'bucket-notif';
+
+config.NOTIFICATION_LOG_DIR = notif_logs_path;
+config.NOTIFICATION_CONNECT_DIR = tmp_connect;
+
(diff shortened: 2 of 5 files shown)


src/test/unit_tests/index.js
```diff
@@ -98,6 +98,7 @@ require('./test_tiering_ttl_worker');
 // require('./test_tiering_upload');
 //require('./test_s3_worm');
 require('./test_bucket_logging');
+require('./test_notifications');
```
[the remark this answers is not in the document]
alphaprinz: Not for now.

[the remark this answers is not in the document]
shirady: Please mention this as a gap because in the NC PRs I think there are missing tests (if I understand you correctly, you gave a reference to this PR, but it doesn't include NC tests at this point).

src/test/unit_tests/test_notifications.js
```diff
@@ -0,0 +1,203 @@
...
+const http_connect_path = path.join(tmp_connect, http_connect_filename);
+//content of connect file, will be written to a file in before()
+const http_connect = {
+    agent_request_object: {"host": "localhost", "port": 9999, "timeout": 1500},
+    request_options_object: {"auth": "amit:passw", "timeout": 1500},
+    notification_protocol: 'http',
+    name: 'http_notif'
+};
+
+const Bucket = 'notif';
```
shirady, line 45: Why Uppercase variable name?
```suggestion
const bucket = 'bucket-notif';
```
Could you change the name so it will be clear that it is a bucket, for example:

src/test/unit_tests/test_notifications.js
```diff
@@ -0,0 +1,203 @@
...
+
+config.NOTIFICATION_CONNECT_DIR = tmp_connect;
```
shirady, line 47: Could you move all the config changes to one block (so we won't miss them).

src/test/unit_tests/test_notifications.js
```diff
@@ -0,0 +1,203 @@
...
+            });
+
+            //there shouldn't be a notification for the delete, wait 2 seconds to validate this
+            await notify_await_result({timeout: 2000});
+        });
+    });
+
+});
+
+const step_wait = 100;
```
shirady, line 175: Why global variable?
It can be moved as a local variable in the function, right?

src/test/unit_tests/test_notifications.js
```diff
@@ -0,0 +1,203 @@
...
+    notification_protocol: 'http',
+    name: 'http_notif'
+};
+
+const Bucket = 'notif';
+
+config.NOTIFICATION_CONNECT_DIR = tmp_connect;
+
+let http_server = null;
+let server_done = false;
```
shirady, line 50: Please add comments on those variables (so we can easy understand their exact use).

src/test/unit_tests/test_notifications.js
```diff
@@ -0,0 +1,203 @@
...
+            await s3.deleteObject({
+                Bucket,
+                Key: 'f1',
+            });
+
+            await notify_await_result({bucket: Bucket, event_name: 'ObjectRemoved:Delete'});
+        });
+
+
+        mocha.it('event notif', async () => {
```
shirady, line 140: I didn't understand the title here (we already tested put and delete notification), maybe you add more explanation here?

src/test/unit_tests/test_notifications.js
```diff
@@ -0,0 +1,203 @@
...
+        }).process_notification_files();
+        await timer.setTimeout(step_wait);
+        if (timeout !== undefined) {
+            timeout -= step_wait;
+            //timeout if we're validating notification did not arrive
+            if (timeout < 0) break;
+        }
+    }
+
+    //if we were not expecting to get a notification (time is undefined),
```
shirady, line 199: typo?
```suggestion
    //if we were not expecting to get a notification (timeout is undefined),
```

src/test/unit_tests/test_notifications.js
shirady, line 1: If there are more test cases you plan to test you can list them as gaps (so we won't miss them).

src/test/unit_tests/test_notifications.js
```diff
@@ -0,0 +1,203 @@
+/* Copyright (C) 2016 NooBaa */
```
shirady, line 1: You can update the year
```suggestion
/* Copyright (C) 2025 NooBaa */
```

src/test/unit_tests/test_notifications.js
```diff
@@ -0,0 +1,203 @@
...
+        });
+
+        mocha.it('simple notif put', async () => {
+            await s3.putObject({
+                Bucket,
+                Key: 'f1',
+                Body: 'this is the body',
+            });
+
+            await notify_await_result({bucket: Bucket, event_name: 'ObjectCreated:Put'});
```
shirady, line 126: You can consider moving the events as constants and reuse them.

shirady (approved):
LGTM

src/test/unit_tests/test_notifications.js
alphaprinz: I can use it so in json literals-

            await s3.createBucket({
                Bucket,
            });

src/test/unit_tests/test_notifications.js
alphaprinz: Yes, but a const makes more sense. It's not a parameter.

PR #8508, opened 2024-11-04
Add a separate clients pool for md_store queries

dannyzaken:
### Explain the changes
This PR aims to prevent the unavailability of Postgres clients in the client pool for system queries when there is a high load of md_store queries. 

#### The Problem
1. In `postgres_client.js,` we access the DB through a single client pool. By default, this pool's max size is 10, and the endpoints' max size is 80. 
2. In systems with a high load and many objects, the number of md_store queries and the time it takes to perform them cause the pool to reach its maximum. As a result, non-md queries are also stuck in the waiting queue, and the entire system may become unresponsive.

#### The Solution
1. Maintain two separate pools in `postgres_client.js`, a default pool and a dedicated md_store pool
2. If the md_store pool is overused, the default pool is left to handle all other, typically short queries, so the default pool should remain relatively available.  

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -199,7 +199,7 @@ config.S3_RESTORE_REQUEST_MAX_DAYS_BEHAVIOUR = 'TRUNCATE';
 /**
  * S3_MAX_KEY_LENGTH controls the maximum key length that will be accepted
  * by NooBaa endpoints.
- * 
+ *
  * This value is 1024 bytes for S3 but the default is `Infinity`
  */
 config.S3_MAX_KEY_LENGTH = Infinity;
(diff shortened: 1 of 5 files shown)


config.js
```diff
@@ -231,7 +231,8 @@ config.ROOT_KEY_MOUNT = '/etc/noobaa-server/root_keys';
 
 config.DB_TYPE = /** @type {nb.DBType} */ (process.env.DB_TYPE || 'postgres');
 
-config.POSTGRES_MAX_CLIENTS = (process.env.LOCAL_MD_SERVER === 'true') ? 80 : 10;
+config.POSTGRES_DEFAULT_MAX_CLIENTS = 10;
+config.POSTGRES_MD_MAX_CLIENTS = (process.env.LOCAL_MD_SERVER === 'true') ? 70 : 10;
```
jackyalbo, line 235: I'm just not sure if we should take the default max clients out of the MD ones... as we saw in a lot of cases that even 80 wasn't enough. other than that it looks good.

jackyalbo (approved):

PR #8514, opened 2024-11-07
noobaa/core:  remove global websocket

aspandey:
**Prep PR for node 22**

In node v22, websocket is enabled by default
https://github.com/nodejs/node/pull/51594

This will cause an error on following line when the global.Websocket would be true.
const WS = global.WebSocket || require('ws');

global.WebSocket: This will points to Node.js's native WebSocket, which does not include a Server class.
require('ws'): The ws package includes both WebSocket and WebSocket.Server, but this fallback is skipped because global.WebSocket is already defined in Node.js v22

### DIFF
--- a/src/rpc/rpc_ws.js
+++ b/src/rpc/rpc_ws.js
@@ -1,7 +1,7 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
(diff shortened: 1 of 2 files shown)


naveenpaul1 (approved):
LGTM

liranmauda (approved):

PR #7237, opened 2023-03-19
CI | Update Ceph S3 Tests Workflow

shirady:
### Explain the changes
1. Remove the comments about an open issue that was closed (noobaa/noobaa-operator#1007).
2. Run a script to wait until we have the backingstore pod (in k8s we cannot wait for a resource that doesn't exist (more information [here](https://github.com/kubernetes/kubectl/issues/1516), based on the change we did in noobaa/noobaa-operator#1063).
3. Update operator repo to the latest commit.
4. Minor changes (adding a new line between 2 steps, removing redundant quotation marks).

### Issues: Fixed #xxx / Gap #xxx
1. none.

### Testing Instructions:
1. none, will be tested in the workflow.

- [ ] Doc added/updated
- [ ] Tests added

PR #8517, opened 2024-11-11
PR1 - To install and upgrade noobaa version for running the tests 

aayushchouhan09:
### Explain the changes
1. This PR is mainly to install and upgrade noobaa version from master to latest.

### Testing Instructions:
1. In PR2 we will the the required tests.

### DIFF
--- a/.github/workflows/upgrade-tests.yaml
+++ b/.github/workflows/upgrade-tests.yaml
@@ -79,8 +79,8 @@ jobs:
       - name: Upgrade noobaa to latest
         run: |
           cd ./noobaa-operator
-          VERSION=$(go run cmd/version/main.go)
-          ./build/_output/bin/noobaa-operator upgrade --noobaa-image='noobaa-core:upgrade-tests' --operator-image='noobaa/noobaa-operator:$VERSION'
+          OPERATOR_VERSION=$(go run cmd/version/main.go)
+          ./build/_output/bin/noobaa-operator upgrade --noobaa-image='noobaa-core:upgrade-tests' --operator-image='noobaa/noobaa-operator:$OPERATOR_VERSION'
 
       - name: Run Upgrade Tests
         run: |
--- a/.github/workflows/upgrade-tests.yaml
+++ b/.github/workflows/upgrade-tests.yaml
@@ -25,7 +25,7 @@ jobs:
           cd ./noobaa-core
           sudo bash ./.travis/deploy_minikube.sh
 
-      - name: Build noobaa (from master)
+      - name: Build noobaa
         run: |
           cd ./noobaa-core
           make noobaa NOOBAA_TAG=noobaa-core:master
@@ -42,7 +42,7 @@ jobs:
           make noobaa NOOBAA_TAG=noobaa-core:upgrade-tests
 
       - name: Checkout noobaa-operator
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           repository: 'noobaa/noobaa-operator'
           path: 'noobaa-operator'
@@ -79,15 +79,10 @@ jobs:
       - name: Upgrade noobaa to latest
         run: |
           cd ./noobaa-operator
-          ./build/_output/bin/noobaa-operator upgrade --noobaa-image='noobaa-core:upgrade-tests'
+          VERSION=$(go run cmd/version/main.go)
+          ./build/_output/bin/noobaa-operator upgrade --noobaa-image='noobaa-core:upgrade-tests' --operator-image='noobaa/noobaa-operator:$VERSION'
 
-      - name: Wait for phase Ready in the backingstore pod
-        run: |
-          cd ./noobaa-operator
-          ./.travis/number_of_pods_in_system.sh --pods 5
-          kubectl wait --for=condition=available backingstore/noobaa-default-backing-store --timeout=5m 
-
-      - name: Run Upgrade Tests (PR code)
+      - name: Run Upgrade Tests
         run: |
           set -x
           cd ./noobaa-core
--- a/.github/workflows/upgrade-tests.yaml
+++ b/.github/workflows/upgrade-tests.yaml
@@ -1,0 +1,94 @@
+name: Upgrade Tests
+on: [push, pull_request, workflow_dispatch]
+
+jobs:
+  upgrade-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true
+
+    env:
+      BRANCH_NAME: ${{ github.event.inputs.branch || 'master' }}
+
+    steps:
+      - name: Checkout noobaa-core (master/input branch)
+        uses: actions/checkout@v4
+        with:
+          repository: 'noobaa/noobaa-core'
+          path: 'noobaa-core'
+          ref: ${{ env.BRANCH_NAME }}
+
+      - name: Deploy minikube
+        run: |
+          cd ./noobaa-core
+          sudo bash ./.travis/deploy_minikube.sh
+
+      - name: Build noobaa (from master)
+        run: |
+          cd ./noobaa-core
+          make noobaa NOOBAA_TAG=noobaa-core:master
+
+      - name: Checkout noobaa-core (latest)
+        uses: actions/checkout@v4
+        with:
+          repository: 'noobaa/noobaa-core'
+          path: 'noobaa-core'
+
+      - name: Build noobaa (latest)
+        run: |
+          cd ./noobaa-core
+          make noobaa NOOBAA_TAG=noobaa-core:upgrade-tests
+
+      - name: Checkout noobaa-operator
+        uses: actions/checkout@v3
+        with:
+          repository: 'noobaa/noobaa-operator'
+          path: 'noobaa-operator'
+
+      - name: Change settings for k8s and minikube
+        run: |
+          sudo mv /root/.kube /root/.minikube $HOME
+          sudo chown -R $USER $HOME/.kube $HOME/.minikube
+          sed "s/root/home\/$USER/g" $HOME/.kube/config > tmp; mv tmp $HOME/.kube/config
+
+      - name: Build operator
+        run: |
+          cd ./noobaa-operator
+          make all
+
+      - name: Install noobaa system (from master)
+        run: |
+          cd ./noobaa-operator
+          ./build/_output/bin/noobaa-operator crd create
+          ./build/_output/bin/noobaa-operator operator install
+          ./build/_output/bin/noobaa-operator system create  \
+          --db-resources='{ "limits": {"cpu": "200m","memory": "2G"}, "requests": {"cpu": "200m","memory": "2G"}}' \
+          --core-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
+          --endpoint-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
+          --noobaa-image='noobaa-core:master'
+          ./build/_output/bin/noobaa-operator status
+
+      - name: Wait for phase Ready in the backingstore pod
+        run: |
+          cd ./noobaa-operator
+          ./.travis/number_of_pods_in_system.sh --pods 5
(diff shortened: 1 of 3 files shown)


.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,94 @@
+name: Upgrade Tests
+on: [push, pull_request, workflow_dispatch]
```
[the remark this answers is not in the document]
achouhan09: Yes I think, here we want to run the tests after upgrading to the latest code pushed.

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,94 @@
...
+      - name: Wait for phase Ready in the backingstore pod
+        run: |
+          cd ./noobaa-operator
+          ./.travis/number_of_pods_in_system.sh --pods 5
+          kubectl wait --for=condition=available backingstore/noobaa-default-backing-store --timeout=5m
+
+      - name: Upgrade noobaa to latest
+        run: |
+          cd ./noobaa-operator
+          ./build/_output/bin/noobaa-operator upgrade --noobaa-image='noobaa-core:upgrade-tests'
```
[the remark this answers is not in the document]
achouhan09: If its required here to provide both the flags then I will add --operator-image also. Thanks

Makefile
```diff
@@ -332,6 +332,12 @@ test-external-pg-sanity: build-ssl-postgres
 	@$(call remove_docker_network)
 .PHONY: test-external-pg-sanity
 
+test-upgrade: tester
+	@echo "\033[1;34mRunning Upgrade Tests - after upgrading from master to latest\033[0m"
+	chmod +x ./src/test/system_tests/run_upgrade_tests.sh
+	./src/test/system_tests/run_upgrade_tests.sh || exit 1
```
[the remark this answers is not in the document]
achouhan09: It is just for phase1 testing purpose. I will remove it. Thanks

.github/workflows/upgrade-tests.yaml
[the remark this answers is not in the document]
achouhan09: The idea here is that we are first installing operator with master code and then we are upgrading the images with the latest code pushed.

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,94 @@
...
+          repository: 'noobaa/noobaa-core'
+          path: 'noobaa-core'
+
+      - name: Build noobaa (latest)
+        run: |
+          cd ./noobaa-core
+          make noobaa NOOBAA_TAG=noobaa-core:upgrade-tests
+
+      - name: Checkout noobaa-operator
+        uses: actions/checkout@v3
```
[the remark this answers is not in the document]
achouhan09: Its a typo, I will update it. Thanks

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,94 @@
...
+          cd ./noobaa-operator
+          ./.travis/number_of_pods_in_system.sh --pods 5
+          kubectl wait --for=condition=available backingstore/noobaa-default-backing-store --timeout=5m
+
+      - name: Upgrade noobaa to latest
+        run: |
+          cd ./noobaa-operator
+          ./build/_output/bin/noobaa-operator upgrade --noobaa-image='noobaa-core:upgrade-tests'
+
+      - name: Wait for phase Ready in the backingstore pod
```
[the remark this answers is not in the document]
achouhan09: I think you are right it is not required after upgrading. I will verify it once and update accordingly. Thanks

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+      - name: Deploy minikube
+        run: |
+          cd ./noobaa-core
+          sudo bash ./.travis/deploy_minikube.sh
+
+      - name: Build noobaa
+        run: |
+          cd ./noobaa-core
+          make noobaa NOOBAA_TAG=noobaa-core:master
```
shirady, line 31: You have the option to choose the branch, right?
So the image tag would not be always "noobaa-core:master".

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Install noobaa system (from master)
+        run: |
+          cd ./noobaa-operator
+          ./build/_output/bin/noobaa-operator crd create
+          ./build/_output/bin/noobaa-operator operator install
+          ./build/_output/bin/noobaa-operator system create  \
+          --db-resources='{ "limits": {"cpu": "200m","memory": "2G"}, "requests": {"cpu": "200m","memory": "2G"}}' \
+          --core-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
+          --endpoint-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
+          --noobaa-image='noobaa-core:master'
```
shirady, line 70: If the image name will be changed, then it should be changed also here.

.github/workflows/upgrade-tests.yaml
achouhan09: Sure, I will update it. Thanks

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Checkout noobaa-core (latest)
+        uses: actions/checkout@v4
+        with:
+          repository: 'noobaa/noobaa-core'
+          path: 'noobaa-core'
+
+      - name: Build noobaa (latest)
+        run: |
+          cd ./noobaa-core
+          make noobaa NOOBAA_TAG=noobaa-core:upgrade-tests
```
shirady, line 42: If we are on master branch, then the image of `noobaa-core:master` and `noobaa-core:upgrade-tests` are the same? (maybe I missed something - why should we continue with the job?)

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Install noobaa system (from master)
+        run: |
+          cd ./noobaa-operator
+          ./build/_output/bin/noobaa-operator crd create
+          ./build/_output/bin/noobaa-operator operator install
+          ./build/_output/bin/noobaa-operator system create  \
+          --db-resources='{ "limits": {"cpu": "200m","memory": "2G"}, "requests": {"cpu": "200m","memory": "2G"}}' \
+          --core-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
+          --endpoint-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
+          --noobaa-image='noobaa-core:${{ env.BRANCH_NAME || 'master' }}'
```
shirady, line 70: In the step you wrote: "Install noobaa system (from master)" - did you mean master in the core repo, right? Then why `--noobaa-image='noobaa-core:${{ env.BRANCH_NAME || 'master' }}'`? I think it should be `noobaa-core:upgrade-tests` (which according to the steps represents "Build noobaa (latest)".
Are you sure the system should be installed in the latest version?

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,89 @@
...
+      - name: Wait for phase Ready in the backingstore pod
+        run: |
+          cd ./noobaa-operator
+          ./.travis/number_of_pods_in_system.sh --pods 5
+          kubectl wait --for=condition=available backingstore/noobaa-default-backing-store --timeout=5m
+
+      - name: Upgrade noobaa to latest
+        run: |
+          cd ./noobaa-operator
+          VERSION=$(go run cmd/version/main.go)
```
shirady, line 82: I'm suggesting to add that it is the operator repo version, so it won't be confusing.
```suggestion
          OPERATOR_VERSION=$(go run cmd/version/main.go)
```

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,89 @@
...
+          ./build/_output/bin/noobaa-operator upgrade --noobaa-image='noobaa-core:upgrade-tests' --operator-image='noobaa/noobaa-operator:$VERSION'
```
shirady, line 83: In case you would accept the suggestion then it needs to be changed also here:
```suggestion
          ./build/_output/bin/noobaa-operator upgrade --noobaa-image='noobaa-core:upgrade-tests' --operator-image='noobaa/noobaa-operator:$OPERATOR_VERSION'
```

.github/workflows/upgrade-tests.yaml
achouhan09: Here, the flow is like in step 1 we are building noobaa core from master/input branch for installation of noobaa and in step 4 we are building noobaa core using PR code to upgrade the noobaa.

.github/workflows/upgrade-tests.yaml
achouhan09: I have also added pull request referece in latest build (in step 4) in my last commit, it would work I think.

.github/workflows/upgrade-tests.yaml
```diff
@@ -1,5 +1,5 @@
 name: Upgrade Tests
-on: [push, pull_request, workflow_dispatch]
+on: [pull_request]
```
shirady, line 2: Why did you change it?

shirady:
@achouhan09 I see the last run failed ([link](https://github.com/noobaa/noobaa-core/actions/runs/12600336806/job/35119145597?pr=8517)).

.github/workflows/upgrade-tests.yaml
achouhan09: No need of it, removed. thanks

.github/workflows/upgrade-tests.yaml
achouhan09: Removed the pull request reference as we don't need it I can see the latest code logs inside the upgrade tests pipeline, this proves that we checked out to latest code. Thanks

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,89 @@
...
+        run: |
+          cd ./noobaa-core
+          sudo bash ./.travis/deploy_minikube.sh
+
+      - name: Build noobaa
+        run: |
+          cd ./noobaa-core
+          make noobaa NOOBAA_TAG=noobaa-core:${{ env.BRANCH_NAME || 'master' }}
+
+      - name: Checkout noobaa-core (latest)
```
shirady, line 33: I think "latest" might be confusing - as it is according to the branch that was set in the input, right?

.github/workflows/upgrade-tests.yaml
shirady: I think I still don't understand it - I think that the title should be updated "Install noobaa system (from master)" It is not from master as we passed --noobaa-image='noobaa-core:${{ env.BRANCH_NAME || 'master' }}'

.github/workflows/upgrade-tests.yaml
```diff
@@ -0,0 +1,89 @@
...
+          --db-resources='{ "limits": {"cpu": "200m","memory": "2G"}, "requests": {"cpu": "200m","memory": "2G"}}' \
+          --core-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
+          --endpoint-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
+          --noobaa-image='noobaa-core:${{ env.BRANCH_NAME || 'master' }}'
+          ./build/_output/bin/noobaa-operator status
+
+      - name: Wait for phase Ready in the backingstore pod
+        run: |
+          cd ./noobaa-operator
+          ./.travis/number_of_pods_in_system.sh --pods 5
```
shirady, line 76: I'm not sure we still need this.
In PR #7237 I mentioned that:

> Run a script to wait until we have the backingstore pod (in k8s we cannot wait for a resource that doesn't exist (more information https://github.com/kubernetes/kubectl/issues/1516, based on the change we did in https://github.com/noobaa/noobaa-operator/pull/1063).

But it was on March 2023 and since that time I think the issue was closed with a solution that we can use.
Anyway, in version 4.19 the number of pods would probably change, so better avoid it if we can.

Issue #8471, opened 2024-10-17
NSFS | Concurrency Tests | Warp fails with error - connect: connection refused


javieramirez1:
### Environment info
- NooBaa Version: **VERSION**
c83f2-dan8-hs200.test.net:  noobaa-core-5.17.0-20241016.el9.x86_64
c83f2-dan10-hs200.test.net:  noobaa-core-5.17.0-20241016.el9.x86_64
### Actual behavior
1.The following workload was executed with a bucket previously put in suspend mode (the errors are also seen when executing warp stat and warp list)
warp get --host=172.20.100.6{0...9}:6443 --access-key="$access_key" --secret-key="$secret_key" --obj.size=1k  --concurrent=1000  --duration=30m --bucket=bucket1$j --insecure --tls

```
arp: <ERROR> download error:Get "https://172.20.100.66:6443/bucket11/RS12F%28xN/3.VOaX3zJIp805QLN%29.rnd": dial tcp 172.20.100.66:6443: connect: connection refused
warp: <ERROR> download error:Get "https://172.20.100.67:6443/bucket11/AE5otFFM/1.6WqmwEJd0re19jG6.rnd": dial tcp 172.20.100.67:6443: connect: connection refused
warp: <ERROR> download error:Get "https://172.20.100.67:6443/bucket11/UgiXqpIj/5.Wct8ORJhXx5iyO20.rnd": dial tcp 172.20.100.67:6443: connect: connection refused
warp: <ERROR> download error:Get "https://172.20.100.64:6443/bucket11/YzjuGWbj/1.gHPA46P9I3pKkUO0.rnd": dial tcp 172.20.100.64:6443: connect: connection refused
----------------------------------------
Operation: PUT. Concurrency: 1000
* Average:0.09 MiB/s, 93.16 obj/s

Throughput by host:
 * https://172.20.100.60:6443: Avg: 0.01 MiB/s, 10.09 obj/s
 * https://172.20.100.61:6443: Avg: 0.01 MiB/s, 9.58 obj/s
 * https://172.20.100.62:6443: Avg: 0.01 MiB/s, 9.50 obj/s
 * https://172.20.100.63:6443: Avg: 0.01 MiB/s, 9.58 obj/s
 * https://172.20.100.64:6443: Avg: 0.01 MiB/s, 9.07 obj/s
 * https://172.20.100.65:6443: Avg: 0.01 MiB/s, 9.46 obj/s
[...]

PR #8518, opened 2024-11-12
NC | NSFS | Panic Printings Added + Try-Catch `memory_monitor` + Change Default Event Logs + Increase `LimitNOFILE`

shirady:
### Explain the changes
1. Panic printings added - in the initial investigation of the issue mentioned below we noticed that we didn't have any "ERROR" or "PANIC" printings and noobaa exited. Therefore, based on previous experience (for example in the noobaa-cli):
https://github.com/noobaa/noobaa-core/blob/fc266c3e274f37d73d35135bdcadbe4f3cab8e18/src/cmd/manage_nsfs.js#L86-L88.
We changed the to use it with a callback (from the function description): 

> "The optional callback parameter will be executed when the data is finally written out, which may not be immediately."

2. Add try-catch clause in `memory_monitor` function - during the investigation of the issue mentioned below we noticed that there were cases when we used a high number of connections, noobaa was down due to error "EMFILE" and in the error stack, we saw that the `memory_monitor` function throw it from the `process.memoryUsage()` which is a node js function. The error stack for example:

> Nov-11 8:14:10.250 [nsfs/1901450] [ERROR] CONSOLE:: memory_monitor got an error Error: EMFILE: too many open files, uv_resident_set_memory
> at process.memoryUsage (node:internal/process/per_thread:168:5)
> at Timeout.memory_monitor [as _onTimeout] (/usr/local/noobaa-core/src/util/panic.js:51:27)
> at listOnTimeout (node:internal/timers:573:17)
> at process.processTimers (node:internal/timers:514:7) {
> errno: -24,
> code: 'EMFILE',
> syscall: 'uv_resident_set_memory'
>  }

3. Change default event logging - we default it to false and enable it only on NC NSFS. We wanted to add an event in the `panic.js` in case the main process of NC NSFS only, and we also noticed that currently, we have some of the "EVENTS" printed in the endpoint pod, which might need to be more accurate (for example - "host" field is with the pod's name):

> [Endpoint/12] [EVENT]{"timestamp":"2024-10-31T06:02:17.105Z","host":"noobaa-endpoint-6f9745857-6dznn","event":{"code":"noobaa_started","message":"Noobaa started","description":"Noobaa started running","entity_type":"NODE","event_type":"STATE_CHANGE","scope":"NODE","severity":"INFO","state":"HEALTHY","pid":12}}

and also remove the message "Event logging not enabled" in case the events are not enabled (to avoid printing it in a containerized environment).
4. Increase the number of `LimitNOFILE` in `noobaa.service` from 2^16 to 600k to set higher limitations on the open files - which are open files descriptors for reading files, websockets, etc.

### Issues: Related to issue #8471
1. Currently, with a high number of connections to one single node noobaa might be down due to "EMFILE" that probably more FDs are open during the `memory_monitor` and we want to avoid those cases.

GAPS:
1. We still don't know in which cases the number of the the number of FDs on noobaa service main process (`ls -al /proc/<PID number>/fd | wc -l`) is increasing - using `--concurrent=1000` there are cases when there is no major change, and cases when it is increasing to high numbers.
2. We might want to change the printed events in s3 failures to be printed periodically to manage our logs printings better.
3. Add the noobaa crash as an event in case it comes from the panic function.

### Testing Instructions:
A. On GPFS machine we used a case that knows that leads to crash - with `LimitNOFILE=15536` on the service and warp with `--concurrent=3000` and `--duration=10`: 
1. Without the try-catch in `memory_monitor` - noobaa was down and we could see the "PANIC" printings and the error of "connect: connection refused", the number of FDs on noobaa service main process (`ls -al /proc/<PID number>/fd | wc -l`) was stuck around the defined limit "LimitNOFILE" and we killed the warp process and it was back to its initial number.
2. With the try-catch in `memory_monitor` - noobaa was not down, we could see the "memory_monitor got an error Error: EMFILE: too many open files, uv_resident_set_memory" printings, but had warp issues "connection reset by peer" (the test ended with errors because not all requests were served, and the number of FDs on noobaa service main process (`ls -al /proc/<PID number>/fd | wc -l`) was back to its initial number.

B. On GPFS cluster change the `LimitNOFILE` to 600k:
1. `systemctl edit noobaa` add the line `LimitNOFILE=600000`.
2. `systemctl daemon-reload`.
3. Restart the service in the cluster, we use `mms3 config change DEBUGLEVEL="all"`.
4. You can check with `systemctl show noobaa | grep LimitNOFILE`.

C. On Containerized:
2. Check in the endpoint logs that you don't see "EVENT" printings.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/debug_module.js
+++ b/src/util/debug_module.js
@@ -509,7 +509,6 @@ function log_syslog_builder(syslevel) {
 */
 function log_event(event) {
     if (!config.EVENT_LOGGING_ENABLED) {
-        console.info('Event logging not enabled');
         return;
     }
     event.pid = process.pid;
--- a/config.js
+++ b/config.js
@@ -493,7 +493,7 @@ config.DEBUG_MODE_PERIOD = 10 * 60 * 1000; // 10 minutes for increased debug lev
 config.dbg_log_level = 0;
 config.DEBUG_FACILITY = 'LOG_LOCAL0';
 config.EVENT_FACILITY = 'LOG_LOCAL2';
-config.EVENT_LOGGING_ENABLED = true;
+config.EVENT_LOGGING_ENABLED = false; // should be changed in NC NSFS configuration
 config.EVENT_LEVEL = 5;
 
 config.LOG_TO_STDERR_ENABLED = true;
--- a/src/cmd/nsfs.js
+++ b/src/cmd/nsfs.js
@@ -244,6 +244,7 @@ class NsfsAccountSDK extends AccountSDK {
 async function main(argv = minimist(process.argv.slice(2))) {
     try {
         config.DB_TYPE = 'none';
+        config.EVENT_LOGGING_ENABLED = true;
         config.NSFS_VERSIONING_ENABLED = true;
         // when using data buckets on noobaa standalone we should set it to true
         config.ENABLE_OBJECT_IO_SEMAPHORE_MONITOR = false;
--- a/src/deploy/noobaa.service
+++ b/src/deploy/noobaa.service
@@ -11,7 +11,7 @@ ExecStart=/usr/local/noobaa-core/bin/node /usr/local/noobaa-core/src/cmd/nsfs.js
 EnvironmentFile=-/etc/sysconfig/noobaa
 ExecStop=/bin/kill $MAINPID
 WorkingDirectory=/usr/local/noobaa-core/
-LimitNOFILE=65536
+LimitNOFILE=600000
 
 [Install]
 WantedBy=multi-user.target
--- a/src/util/panic.js
+++ b/src/util/panic.js
@@ -8,12 +8,16 @@ const child_process = require('child_process');
 process.on('uncaughtException', err => panic('process uncaughtException', err));
 
 function panic(message, err) {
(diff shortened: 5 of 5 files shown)


src/util/debug_module.js
```diff
@@ -508,6 +508,7 @@ function log_syslog_builder(syslevel) {
  * log_builder will format the message with prefix. 
 */
 function log_event(event) {
+    if (config.DB_TYPE !== 'none') return; // currently event logging is only in NC NSFS
```
naveenpaul1, line 511: Event logging option to other deployment is not a bad decision, It will just print one more log in log file and it will help debugging at the end.
@romayalon @shirady WDYT?

shirady: @naveenpaul1 currently the logs are not accurate for containerized env, if we want them we need a bit change them (for example "host" field), and currently some of the logs are on S3 errors and we have those printed in the log, having it twice might fill the logs with those printings (for example `OBJECT_GET_FAILED` on a client that crawls many files name and tries to GET them).

romayalon (approved):
LGTM

PR #8535, opened 2024-11-18
remove link-local suffix from remoteAddress in ice.js

dannyzaken:
### Explain the changes
1. in node js 22, in some cases, `socket.remoteAddress` contains a link-local suffix, e.g.: `fe80::858:aff:fe83:2e%eth0`
2. This address format is not parsed correctly by `ip_module.isPrivate()` (npm `ip` module), and throws an error `invalid ipv4 address`
3. changed `ice.js` only to remove the %[interface] suffix.
4. wrapped `ip_module.isPrivate()` with a non-throwing function

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/rpc/ice.js
+++ b/src/rpc/ice.js
@@ -1,5 +1,5 @@
 /* Copyright (C) 2016 NooBaa */
-/* eslint max-lines: ['error', 1550] */
+/* eslint max-lines: ['error', 1650] */
 'use strict';
 
 module.exports = Ice;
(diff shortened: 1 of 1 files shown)


nimrod-becker (approved):

dannyzaken:
I am not sure what the correct address value should be used. The link-local format is valid but is not handled correctly by some applications. 
The previous behavior of node.js 20 (remoteAddress without the link-local suffix) worked for us, so I patched it, but only on ice.js. If we will see other failures as a result of this address suffix we will consider a more general fix.

liranmauda (approved):

PR #8529, opened 2024-11-14
Add AWS API compatibility table

Neon-White:
### Explain the changes
1. Add a table documenting which AWS API calls are supported and over which stores

PR #8538, opened 2024-11-18
CORS - support AWS APIs (NC and containerized)

### LINKED ISSUE #6080: Support S3 CORS API
When using aws s3 sdk to set cors rules, error occurred.
code:
```go
func setCors(bucketName string) error {
 region := "cn-northwest-1"
 if tools.Config.S3Region != "" {
  region = tools.Config.S3Region
 }
 sess := session.Must(session.NewSessionWithOptions(session.Options{
  SharedConfigState: session.SharedConfigEnable,
  Config: aws.Config{
   Region:   aws.String(region),
   Endpoint: aws.String(Key.EndPoint), //Bucket所在Endpoint
   HTTPClient: func() *http.Client {
    return &http.Client{
     Transport: func() *http.Transport {
      return &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}
     }(),
     Timeout: 3 * time.Second,
    }
   }(),
   //S3ForcePathStyle: aws.Bool(true),
  },
 }))
 svc := s3.New(sess)
 rule := s3.CORSRule{
  AllowedHeaders: aws.StringSlice([]string{"*"}),
  AllowedOrigins: aws.StringSlice([]string{"*"}),
  MaxAgeSeconds:  aws.Int64(3000),
  // Add HTTP methods CORS request that were specified in the CLI.
  AllowedMethods: aws.StringSlice([]string{"PUT", "POST", "GET"}),
 }
 params := s3.PutBucketCorsInput{
  Bucket: aws.String(bucketName),
  CORSConfiguration: &s3.CORSConfiguration{
   CORSRules: []*s3.CORSRule{&rule},
  },
 }
 _, err := svc.PutBucketCors(&params)
 if err != nil {
  log.Println("setCors, invoke s3 PutBucketCors error:", err)
  return err
 }
 log.Printf("setCors, Updated bucket %q CORS finished\n", bucketName)
 return nil
}
```
error message from noobaa:
> A header you provided implies functionality that is not implemented.
        status code: 501, request id: kc1muixr-as3c6z-12kx, host id: kc1muixr-as3c6z-12kx

Support S3 CORS API, just like S3 BucketPolicy API support in v5.3 'Bomberman'.

Still thinking about it. If Noobaa developers know about any workarounds, please let me know.

**Additional context**
NA.

jackyalbo:
### Explain the changes
1. Adding support to the S3 CORS API: PUT/GET/DELETE
2. Changing default CORS behavior from a set of defaults to supporting the API following AWS documentation:
https://docs.aws.amazon.com/AmazonS3/latest/userguide/cors.html
with some ideas from Azure documentation here:  https://learn.microsoft.com/en-us/rest/api/storageservices/cross-origin-resource-sharing--cors--support-for-the-azure-storage-services
3. Adding an option to create an anonymous bucket policy for read when getting a public-read acl, used by ceph-s3 tests

1. Fixed #6080 

### Testing Instructions:
1.  Added a few tests from ceph-tests suite.

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -64,7 +64,7 @@ config.BUFFERS_MEM_LIMIT_MIN = 32 * 1024 * 1024; // just some workable minimum s
 config.BUFFERS_MEM_LIMIT_MAX = 4 * 1024 * 1024 * 1024;
 config.BUFFERS_MEM_LIMIT = Math.min(
     config.BUFFERS_MEM_LIMIT_MAX,
-    Math.max(Math.floor(config.CONTAINER_MEM_LIMIT / 4), config.BUFFERS_MEM_LIMIT_MIN, )
+    Math.max(Math.floor(config.CONTAINER_MEM_LIMIT / 4), config.BUFFERS_MEM_LIMIT_MIN)
 );
 
 ////////////////////////
@@ -152,9 +152,8 @@ config.ENDPOINT_HTTP_SERVER_REQUEST_TIMEOUT = 300 * 1000;
 config.ENDPOINT_HTTP_SERVER_KEEPALIVE_TIMEOUT = 5 * 1000;
 config.ENDPOINT_HTTP_MAX_REQUESTS_PER_SOCKET = 0; // 0 = no limit
 
-// For now we enable fixed CORS for all buckets
-// but this should become a setting per bucket which is configurable
-// with the s3 put-bucket-cors api.
+// For now we enable fixed CORS only for sts
+// for S3 per bucket is configurabl with the s3 put-bucket-cors api.
 // note that browsers do not really allow origin=* with credentials,
 // but we just allow both from our side for simplicity.
 config.S3_CORS_ENABLED = true;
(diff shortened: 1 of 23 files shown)


nimrod-becker:
I know it's still in draft, but remember to update the table from #8529 once @Neon-White merged it.
It could also be worth adding to bucketspace_s3 since it's a simple proxy in this case.

jackyalbo:
> I know it's still in draft, but remember to update the table from #8529 once @Neon-White merged it. It could also be worth adding to bucketspace_s3 since it's a simple proxy in this case.

@nimrod-becker, about your second remark, it seems we don't support most of the bucket-level operations in bucketspace_s3 (lifecycle, tagging, encryption, bucket-policy, and so on), so I think we shouldn't add it also in this case, it can be added once we will support bucket-level operations in our APIs in the future (I remember that we wanted to work on it in the past)

nimrod-becker:
We can, we can either do it all together or add this one,

liranmauda (approved):
LGTM

PR #8552, opened 2024-11-21
NSFS | NC | add option to set account supplemental groups

### LINKED ISSUE #7274: Supporting supplemental groups for accounts
### Environment info
- NooBaa Version: **master**

### Actual behavior
1. Currently, nsfs config setting of account creation/update does not include supplemental groups configuration.
When switching the user in FS_NAPI we only switch and restore to the user's uid gid.

### Expected behavior
1. Support supplemental groups on account creation/update with nsfs account config.
2. When switching user in FS_NAPI, we should also switch groups to the supplemental groups received by the user using setgroups, and restore when restoring the uid gid.

nadavMiz:
### Explain the changes
supplemental groups are additional groups an account can be part of besides his main group (gid). add option to set a users supplemental groups through the nsfs cli. see `Supplementary group IDs` in https://man7.org/linux/man-pages/man7/credentials.7.html. having addition groups allows account to access files and directory that allow access to one of the accounts supplemental groups (similar to main group access ). supplementary group IDs are used mainly for adding permissions. for other purposes the accounts main GID will be used (for example determining the group of a file created by the account). note that this enables access only on file system level, s3 commands will still require bucket policy permissions. in the same manner account will still be block from accessing the file system even if it has bucket policy permissions
1. add option to set account supplemental groups:
2. supplemental groups can be added by nsfs cli either by passing a single positive number (202) or string of comma separated positive number ("200,101")
3. supplemental groups can be added using from_file as wither one of the previously mentioned types or as an array of positive numbers ([212,313])
4. supplemental groups will be saved as part of the account configuration in the section `nsfs_account_config`
5. set supplemental groups as part of setting the user id for both linux and darwin (see os_linux.cpp and os_darwin.cpp)
6. add tests for both adding and updating account configuration (including from_file, and anonymous access)
7. change documentation to include data about supplemental_groups option 
8. since mac-os creates new directories with parent directory gid and not the process gid. change new directories owner in test_nsfs_access.js. opened a new issue for this problem #8665 
### Testing Instructions:
1. `sudo npx jest test_nc_nsfs_account_cli.test.js`
2. `sudo npx jest test_nc_nsfs_anonymous_cli.test.js`
3. `sudo node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nsfs_access.js`
4. `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nsfs_integration.js`

### Gaps
#8665 
https://github.com/noobaa/noobaa-operator/issues/1505

- [x] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -24,7 +24,7 @@ const { print_usage } = require('../manage_nsfs/manage_nsfs_help_utils');
 const { TYPES, ACTIONS, LIST_ACCOUNT_FILTERS, LIST_BUCKET_FILTERS, GLACIER_ACTIONS } = require('../manage_nsfs/manage_nsfs_constants');
 const { throw_cli_error, get_bucket_owner_account_by_name,
     write_stdout_response, get_boolean_or_string_value, has_access_keys, set_debug_level,
-    is_name_update, is_access_key_update } = require('../manage_nsfs/manage_nsfs_cli_utils');
+    is_name_update, is_access_key_update, parse_comma_delimited_string } = require('../manage_nsfs/manage_nsfs_cli_utils');
 const manage_nsfs_validations = require('../manage_nsfs/manage_nsfs_validations');
 const nc_mkm = require('../manage_nsfs/nc_master_key_manager').get_instance();
 const notifications_util = require('../util/notifications_util');
@@ -359,6 +359,11 @@ async function fetch_account_data(action, user_input) {
         data.access_keys[0].secret_key = data.access_keys[0].secret_key === undefined ? undefined :
             new SensitiveString(String(data.access_keys[0].secret_key));
     }
+    //since supplemental_groups is an array, new list will merge with the old one instead of replacing it in fetch_existing_account_data
+    //so we need to replace this value after merging the data
+    data.nsfs_account_config.supplemental_groups = user_input.supplemental_groups === undefined ?
+        data.nsfs_account_config.supplemental_groups : parse_comma_delimited_string(user_input.supplemental_groups);
+
     if (data.new_access_key) data.new_access_key = new SensitiveString(data.new_access_key);
     // fs_backend deletion specified with empty string '' (but it is not part of the schema)
     data.nsfs_account_config.fs_backend = data.nsfs_account_config.fs_backend || undefined;
--- a/src/manage_nsfs/manage_nsfs_cli_errors.js
+++ b/src/manage_nsfs/manage_nsfs_cli_errors.js
@@ -346,6 +346,11 @@ ManageCLIError.InvalidGlacierOperation = Object.freeze({
     message: 'only "migrate", "restore" and "expiry" subcommands are supported',
     http_code: 400,
 });
+ManageCLIError.InvalidSupplementalGroupsList = Object.freeze({
+    code: 'InvalidSupplementalGroupsList',
+    message: 'supplemental groups must be a list of group ids (group id is zero or a positive integer)',
+    http_code: 400,
+});
 
 
 ////////////////////////
--- a/src/manage_nsfs/manage_nsfs_cli_utils.js
+++ b/src/manage_nsfs/manage_nsfs_cli_utils.js
@@ -91,7 +91,25 @@ function get_boolean_or_string_value(value) {
 }
 
 /**
- * get_options_from_file will read a JSON file that include key-value of the options 
+ * This function parse a comma delimited string of numbers ('0,212,111') to an array of numbers.
+ * This function assumes string format was validated before calling the function, wrong string format can
+ * lead to unexpected output (usually array of NaN)
+ * 1. if the value is a number return array with this number (3 => [3])
+ * 2. if the value is a string return an array of numbers ('0,212,111' => [0,212,111])
+ * 3. for all other types (including object and undefined) return the value itself
+ */
+function parse_comma_delimited_string(value) {
+    if (typeof value === 'number') {
+        return [value];
+    }
+    if (typeof value === 'string') {
+        return value.split(',').map(val => Number(val));
+    }
+    return value;
+}
+
+/**_
+ * get_options_fromfile will read a JSON file that include key-value of the options
  * (instead of flags) and return its content
  * @param {string} file_path
  */
@@ -156,6 +174,7 @@ function is_access_key_update(data) {
 exports.throw_cli_error = throw_cli_error;
 exports.write_stdout_response = write_stdout_response;
 exports.get_boolean_or_string_value = get_boolean_or_string_value;
+exports.parse_comma_delimited_string = parse_comma_delimited_string;
 exports.get_bucket_owner_account_by_name = get_bucket_owner_account_by_name;
 exports.get_bucket_owner_account_by_id = get_bucket_owner_account_by_id;
 exports.get_options_from_file = get_options_from_file;
--- a/src/manage_nsfs/manage_nsfs_constants.js
+++ b/src/manage_nsfs/manage_nsfs_constants.js
@@ -44,16 +44,16 @@ const FROM_FILE = 'from_file';
 const ANONYMOUS = 'anonymous';
 
 const VALID_OPTIONS_ACCOUNT = {
-    'add': new Set(['name', 'uid', 'gid', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', FROM_FILE, ...CLI_MUTUAL_OPTIONS]),
-    'update': new Set(['name', 'uid', 'gid', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', 'new_name', 'regenerate', ...CLI_MUTUAL_OPTIONS]),
+    'add': new Set(['name', 'uid', 'gid', 'supplemental_groups', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', FROM_FILE, ...CLI_MUTUAL_OPTIONS]),
+    'update': new Set(['name', 'uid', 'gid', 'supplemental_groups', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', 'new_name', 'regenerate', ...CLI_MUTUAL_OPTIONS]),
     'delete': new Set(['name', ...CLI_MUTUAL_OPTIONS]),
     'list': new Set(['wide', 'show_secrets', 'gid', 'uid', 'user', 'name', 'access_key', ...CLI_MUTUAL_OPTIONS]),
     'status': new Set(['name', 'access_key', 'show_secrets', ...CLI_MUTUAL_OPTIONS]),
 };
 
 const VALID_OPTIONS_ANONYMOUS_ACCOUNT = {
-    'add': new Set(['uid', 'gid', 'user', 'anonymous', ...CLI_MUTUAL_OPTIONS]),
-    'update': new Set(['uid', 'gid', 'user', 'anonymous', ...CLI_MUTUAL_OPTIONS]),
+    'add': new Set(['uid', 'gid', 'user', 'supplemental_groups', 'anonymous', ...CLI_MUTUAL_OPTIONS]),
+    'update': new Set(['uid', 'gid', 'user', 'supplemental_groups', 'anonymous', ...CLI_MUTUAL_OPTIONS]),
     'delete': new Set(['anonymous', ...CLI_MUTUAL_OPTIONS]),
     'status': new Set(['anonymous', ...CLI_MUTUAL_OPTIONS]),
 };
@@ -105,6 +105,7 @@ const OPTION_TYPE = {
     owner: 'string',
     uid: 'number',
     gid: 'number',
+    supplemental_groups: 'string',
     new_buckets_path: 'string',
     user: 'string',
     access_key: 'string',
--- a/src/manage_nsfs/manage_nsfs_validations.js
+++ b/src/manage_nsfs/manage_nsfs_validations.js
@@ -20,7 +20,7 @@ const { check_root_account_owns_user } = require('../nc/nc_utils');
 //// GENERAL VALIDATIONS ////
 /////////////////////////////
 
-/** 
+/**
  * validate_input_types checks if input option are valid.
  * if the the user uses from_file then the validation is on the file (in different iteration)
  * @param {string} type
@@ -41,6 +41,7 @@ async function validate_input_types(type, action, argv) {
     validate_flags_combination(type, action, input_options);
     validate_flags_value_combination(type, action, input_options_with_data);
     validate_account_name(type, action, input_options_with_data);
+    validate_supplemental_groups(input_options_with_data);
     if (action === ACTIONS.UPDATE) validate_min_flags_for_update(type, input_options_with_data);
 
     // currently we use from_file only in add action
@@ -58,6 +59,7 @@ async function validate_input_types(type, action, argv) {
         validate_flags_combination(type, action, input_options_from_file);
         validate_flags_value_combination(type, action, input_options_with_data_from_file);
         validate_account_name(type, action, input_options_with_data_from_file);
+        validate_supplemental_groups(input_options_with_data_from_file);
         return input_options_with_data_from_file;
     }
 }
@@ -175,7 +177,11 @@ function validate_options_type_by_value(input_options_with_data) {
             if ((option === 'bucket_policy' || option === 'notifications') && type_of_value === 'object') {
                 continue;
             }
-            const details = `type of flag ${option} should be ${type_of_option}`;
+            //special case for supplemental groups
+            if (option === 'supplemental_groups' && ((type_of_value === 'object') || (type_of_value === 'number'))) {
+                continue;
+            }
+            const details = `type of flag ${option} should be ${type_of_option} (and the received value is ${value})`;
             throw_cli_error(ManageCLIError.InvalidArgumentType, details);
         }
     }
@@ -195,6 +201,36 @@ function validate_boolean_string_value(value) {
         return true;
     }
     return false;
+}
+
+/**
+ * validates supplemental groups array.
+ * string type: is comma seperated positive numbers. should not begin or end with a comma.
+ * number type: the number should be positive
+ * object type: should be array of possitive integers
+ * @param {object} input_options_with_data
+ */
+function validate_supplemental_groups(input_options_with_data) {
+    const value = input_options_with_data.supplemental_groups;
+    if (!value) {
+        return;
+    }
+    if (typeof value === 'string') {
+        const regex = /^[0-9]+(,[0-9]+)+$/;
+        if (!regex.test(value)) {
+            throw_cli_error(ManageCLIError.InvalidSupplementalGroupsList);
+        }
+    } else if (typeof value === 'number') {
+        if (value < 0) {
+            throw_cli_error(ManageCLIError.InvalidSupplementalGroupsList);
+        }
+    } else if (typeof value === 'object') {
+        for (const entry of Object.values(value)) {
+            if (isNaN(Number(entry)) || Number(entry) < 0) {
+                throw_cli_error(ManageCLIError.InvalidSupplementalGroupsList);
+            }
+        }
+    }
 }
 
 /**
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -538,6 +538,43 @@ load_xattr_get_keys(Napi::Object& options, std::vector<std::string>& _xattr_get_
 }
 
 /**
+* converts Napi::Array of numbers to std::vector
+* typename T - type of the vector to convert to (e.g int, uint, gid_t)
+* warning: function will only work on vector with numeric types. should not be used with other types
+*/
+template<typename T>
+static std::vector<T>
+convert_napi_number_array_to_number_vector(const Napi::Array& arr) {
+    std::vector<T> new_vector;
+    const std::size_t arr_length = arr.Length();
+    for (std::size_t i = 0; i < arr_length; ++i) {
+        new_vector.push_back(static_cast<Napi::Value>(arr[i]).ToNumber());
+    }
+    return new_vector;
+}
+
+/**
+ * converts std::vector to comma seperated string so it can be printed to logs
+ */
+template<typename T>
+static std::string
+stringfy_vector(std::vector<T>& vec) {
+    std::stringstream ss;
+    std::size_t size = vec.size();
+    for(std::size_t i = 0; i < size; ++i) {
+        if (i > 0) ss << ',';
+        ss << vec[i];
+    }
+    return ss.str();
+}
+
+
+static std::string get_groups_as_string() {
+    std::vector<gid_t> groups = ThreadScope::get_process_groups();
+    return stringfy_vector(groups);
+}
+
+/**
  * FSWorker is a general async worker for our fs operations
  */
 struct FSWorker : public Napi::AsyncWorker
@@ -560,6 +597,7 @@ struct FSWorker : public Napi::AsyncWorker
     double _took_time;
     Napi::FunctionReference _report_fs_stats;
     bool _should_add_thread_capabilities;
+    std::vector<gid_t> _supplemental_groups;
 
     // executes the ctime check in the stat and read file fuctions
     // NOTE: If _do_ctime_check = false, then some functions will fallback to using mtime check
@@ -576,6 +614,7 @@ struct FSWorker : public Napi::AsyncWorker
         , _warn_threshold_ms(0)
         , _took_time(0)
         , _should_add_thread_capabilities(false)
+        , _supplemental_groups()
         , _do_ctime_check(false)
     {
         for (int i = 0; i < (int)info.Length(); ++i) _args_ref.Set(i, info[i]);
@@ -585,6 +624,9 @@ struct FSWorker : public Napi::AsyncWorker
             if (fs_context.Get("gid").IsNumber()) _gid = fs_context.Get("gid").ToNumber();
             if (fs_context.Get("backend").ToBoolean()) {
                 _backend = fs_context.Get("backend").ToString();
+            }
+            if (fs_context.Has("supplemental_groups")) {
+                _supplemental_groups = convert_napi_number_array_to_number_vector<gid_t>(fs_context.Get("supplemental_groups").As<Napi::Array>());
             }
             if (fs_context.Get("warn_threshold_ms").ToBoolean()) {
                 _warn_threshold_ms = fs_context.Get("warn_threshold_ms").ToNumber();
@@ -604,10 +646,12 @@ struct FSWorker : public Napi::AsyncWorker
     virtual void Work() = 0;
     void Execute() override
     {
-        DBG1("FS::FSWorker::Execute: " << _desc << DVAL(_uid) << DVAL(_gid) << DVAL(_backend));
+        const std::string supplemental_groups = stringfy_vector(_supplemental_groups);
+        DBG1("FS::FSWorker::Execute: " << _desc << DVAL(_uid) << DVAL(_gid) << DVAL(_backend) << DVAL(supplemental_groups));
         ThreadScope tx;
-        tx.set_user(_uid, _gid);
-        DBG1("FS::FSWorker::Execute: " << _desc << DVAL(_uid) << DVAL(_gid) << DVAL(geteuid()) << DVAL(getegid()) << DVAL(getuid()) << DVAL(getgid()));
+        tx.set_user(_uid, _gid, _supplemental_groups);
+        std::string new_supplemental_groups = get_groups_as_string();
+        DBG1("FS::FSWorker::Execute: " << _desc << DVAL(_uid) << DVAL(_gid) << DVAL(geteuid()) << DVAL(getegid()) << DVAL(getuid()) << DVAL(getgid()) << DVAL(new_supplemental_groups));
 
         if(_should_add_thread_capabilities) {
             tx.add_thread_capabilities();
--- a/src/native/util/os_darwin.cpp
+++ b/src/native/util/os_darwin.cpp
@@ -3,6 +3,9 @@
 
 #include "common.h"
 #include <sys/kauth.h> // for KAUTH_UID_NONE
+#include <sys/param.h>
+#include <unistd.h>
+#include <algorithm>
 
 namespace noobaa
 {
@@ -37,12 +40,34 @@ get_current_uid()
 
 const uid_t ThreadScope::orig_uid = getuid();
 const gid_t ThreadScope::orig_gid = getgid();
+const std::vector<gid_t> ThreadScope::orig_groups = get_process_groups();
 
+
+/**
+ * set the effective uid/gid/supplemental_groups of the current thread using pthread_getugid_np and setgroups
+ * we have to bypass the libc wrappers because posix requires it to syncronize
+ * uid, gid & supplemental_groups to all threads which is undesirable in our case.
+ * this way is deprecated, but we don't seem to have another way.
+ * pthread_setugid_np sets the effective uid/gid and clears the supplementary groups of the current thread
+ * See https://www.unix.com/man-page/osx/2/pthread_setugid_np/
+ * regarding setgroups When called from a thread running under an assumed per-thread identity,
+ * this function will operate against the per-thread. so we have to run it after _mac_thread_setugid.
+ * also _mac_thread_setugid will reset the groups array anyway
+ * https://github.com/apple/darwin-xnu/blame/main/bsd/kern/kern_prot.c (see setgroups1)
+ */
 void
 ThreadScope::change_user()
 {
     if (_uid != orig_uid || _gid != orig_gid) {
         MUST_SYS(_mac_thread_setugid(_uid, _gid));
+        if (!_groups.empty()) {
+            /*accourding to BSD Manual https://man.freebsd.org/cgi/man.cgi?query=setgroups
+            which darwin is occasionally compliant to setgroups changes the effective gid according to
+            the first element on the list. add the effective gid as the first element to prevent issues*/
+            _groups.push_back(_gid);
+            std::swap(_groups.front(), _groups.back());
+            MUST_SYS(setgroups(_groups.size(), &_groups[0]));
+        }
     }
 }
 
@@ -51,6 +76,7 @@ ThreadScope::restore_user()
 {
     if (_uid != orig_uid || _gid != orig_gid) {
         MUST_SYS(_mac_thread_setugid(KAUTH_UID_NONE, KAUTH_UID_NONE));
+        MUST_SYS(setgroups(orig_groups.size(), &orig_groups[0]));
     }
 }
 
@@ -60,6 +86,14 @@ ThreadScope::add_thread_capabilities()
     //set capabilities not used in darwin
     LOG("function set_capabilities_linkat is unsupported in darwin");
     return -1;
+}
+
+std::vector<gid_t>
+ThreadScope::get_process_groups() {
+    std::vector<gid_t> groups(NGROUPS_MAX);
+    int r = getgroups(NGROUPS_MAX, &groups[0]);
+    groups.resize(r);
+    return groups;
 }
 
 } // namespace noobaa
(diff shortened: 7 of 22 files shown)


src/native/util/os_linux.cpp
```diff
@@ -43,9 +43,14 @@ const std::vector<gid_t> ThreadScope::orig_groups = [] {
 void
 ThreadScope::change_user()
 {
+    if (_groups == NULL || _groups->empty()) {
+        MUST_SYS(syscall(SYS_setgroups, 0, NULL));
+    }
+    else if (_groups != NULL && *_groups != orig_groups) {
+        MUST_SYS(syscall(SYS_setgroups, _groups->size(), &_groups->at(0)));
```
naveenpaul1, line 50: are we setting only the first item?

src/native/util/os_linux.cpp
```diff
@@ -61,7 +66,9 @@ ThreadScope::restore_user()
         // must restore uid first otherwise will fail on permission
         MUST_SYS(syscall(SYS_setresuid, -1, orig_uid, -1));
         MUST_SYS(syscall(SYS_setresgid, -1, orig_gid, -1));
-        MUST_SYS(syscall(SYS_setgroups, ThreadScope::orig_groups.size(), &ThreadScope::orig_groups[0]));
+        if(_groups == NULL || orig_groups != *_groups) {
```
romayalon, line 69: why do we need this condition? we always want to restore the original groups

src/test/unit_tests/test_nsfs_access.js
```diff
@@ -66,29 +75,22 @@ mocha.describe('new tests check', async function() {
...
         }
     });
     mocha.it('ROOT readdir - dir created by non dir - success', async function() {
-        try {
-            const root_entries = await nb_native().fs.readdir(ROOT_FS_CONFIG, full_path_non_root);
-            assert.fail(`root has access to a folder that should not exist - ${root_entries}`);
-        } catch (err) {
-            assert.equal(err.code, 'ENOENT');
-        }
+            await nb_native().fs.readdir(ROOT_FS_CONFIG, full_path_non_root);
```
romayalon, line 88: why did you change this test?

src/test/unit_tests/test_nsfs_access.js
```diff
@@ -66,29 +75,22 @@ mocha.describe('new tests check', async function() {
...
     });
 
     mocha.it('NON ROOT 1 readdir - success', async function() {
-        try {
-            const non_root_entries = await nb_native().fs.readdir(NON_ROOT1_FS_CONFIG, full_path_non_root);
-            assert.fail(`non root 1 has access to a folder created by root with 770 perm - ${p} - ${non_root_entries}`);
-        } catch (err) {
-            assert.equal(err.code, 'EACCES');
-        }
+        const non_root_entries = await nb_native().fs.readdir(NON_ROOT1_FS_CONFIG, full_path_non_root);
```
romayalon, line 92: please don't change existing tests that were supposed to fail, add new tests for the supplemental groups tests

src/cmd/manage_nsfs.js
```diff
@@ -338,6 +338,7 @@ async function fetch_account_data(action, user_input) {
             distinguished_name: user_input.user,
             uid: user_input.user ? undefined : user_input.uid,
             gid: user_input.user ? undefined : user_input.gid,
+            supplemental_groups: _.isUndefined(user_input.supplemental_groups) ? undefined : String(user_input.supplemental_groups),
```
romayalon, line 341: please make sure when adding a new flag to the cli to add it to the help and the documentation as well

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -197,6 +201,19 @@ function validate_boolean_string_value(value) {
     return false;
 }
 
+function validate_supplemental_groups(value) {
```
romayalon, line 204: please add type script annotations for new functions, it helps with finding bugs on js

src/cmd/manage_nsfs.js
romayalon, line 341: we should consider adding supplemental groups to accountspace_fs._new_user_defaults()
@shirady @guymguym WDYT?

romayalon, line 341: looks like supplemental_groups loading is missing in bucketspace_fs.read_account_by_access_key()

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -197,6 +201,19 @@ function validate_boolean_string_value(value) {
...
+    if (value && typeof value === 'string') {
```
romayalon, line 205: need to add the same validation on account_server/accountspace_FS
CC: @shirady

src/native/fs/fs_napi.cpp
```diff
@@ -560,6 +571,7 @@ struct FSWorker : public Napi::AsyncWorker
     double _took_time;
     Napi::FunctionReference _report_fs_stats;
     bool _should_add_thread_capabilities;
+    std::vector<gid_t> _supplemental_groups;
```
romayalon, line 574: looks like the loading supplemental_groups to the fs_context is missing in the CLI, did I miss anything? 
it should be added in native_fs_utils.get_fs_context()

romayalon: please make sure it's being passed to NAPI correctly in all places, especially in NamespaceFS, if you could add tests on these flows as well it'll be even better

romayalon (dismissed):

src/manage_nsfs/manage_nsfs_validations.js
shirady: I'm not sure I understand - in which context do you want to add those validations?

src/native/util/os_linux.cpp
nadavMiz: no, cpp vectoe implementation is essentially wrapped c array. so sending a pointer to the first element is the same as sending pointer to a c array

nadavMiz: in c arrays are just contiguous memory. so it handles it by sending pointer to the beginning of the array, and it infers the offset by the type. maybe I can convert it to c array first, just to make it more readable

src/server/system_services/schemas/nsfs_account_schema.js
```diff
@@ -81,6 +81,7 @@ module.exports = {
                 properties: {
                     uid: { type: 'number' },
                     gid: { type: 'number' },
+                    supplemental_groups: { type: 'string' },
```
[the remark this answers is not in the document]
nadavMiz: I began from that, but it will cause issues with updating the configuration as it is a list. also we will need to convert on the native side anyway as we will convert between napi array to std::vector. I thought it was better this way as we don't need the list outside the native and it simplifies the process. also it looks better when looking at the configuration

src/test/unit_tests/test_nsfs_access.js
nadavMiz: it was wrong. by test name it was supposed to test that root is able to access folder that was created by non-root user. and it should succeed. in practice it tested that root couldn't access a folder that doesn't exist

src/test/unit_tests/test_nsfs_access.js
nadavMiz: sure, but by test name it was suppose to succeed. I assume that the test purpose was to test that non root 1 should be able to access folders it was allowed to

src/native/fs/fs_napi.cpp
```diff
@@ -537,6 +537,17 @@ load_xattr_get_keys(Napi::Object& options, std::vector<std::string>& _xattr_get_
     }
 }
 
+template<typename T>
+static std::vector<T> convert_delimited_string_to_number_vector(const std::string str, const char delimiter) {
```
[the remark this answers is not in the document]
nadavMiz: sure we would still need to convert here

src/native/util/os_linux.cpp
nadavMiz: we can. I just though we wouldn't need to if its already original groups

shirady (commented):
I added my comments (without the native files).

I would list things that you can add in this PR:
- Tests for the noobaa-cli that combines the supplemental groups with the `--from_file` flag.
- Tests for the noobaa-cli that combines the supplemental groups with the `--anonymous` flag
- Update the `S3Ops.md` about permission - currently only UID, GID is mentioned.

src/native/fs/fs_napi.cpp
```diff
@@ -537,6 +537,55 @@ load_xattr_get_keys(Napi::Object& options, std::vector<std::string>& _xattr_get_
...
+        new_vector.push_back(static_cast<Napi::Value>(arr[i]).ToNumber());
+    }
+    return new_vector;
+}
+
+/**
+ * converts std::vector to comma seperated string so it can be printed to logs
+ */
+template<typename T>
+static std::string stringfy_vector(std::vector<T>& vec) {
```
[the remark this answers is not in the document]
nadavMiz: it is, and I might have been able to combine them with more general template. anyway there was also a problem with calling getgroups from this file as it is the same for both os. instead I moved getgroups to os.h, added linux and mac implementation that gets it as a vector. and use this implementation to initialize the orig list as it is practically the same

src/native/fs/fs_napi.cpp
```diff
@@ -537,6 +537,55 @@ load_xattr_get_keys(Napi::Object& options, std::vector<std::string>& _xattr_get_
...
+    std::stringstream ss;
+    for(std::size_t i = 0; i < size; ++i) {
+        if (i > 0) ss << ',';
+        ss << typed_arr[i];
+    }
+    return ss.str();
+}
+
+static std::string get_groups_as_string() {
+    const size_t max_group_size = 12; // arbitary size. c does not allow variable length array
```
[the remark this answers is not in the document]
nadavMiz: the idea is that you can't create an array with length that is determined at runtime. the standard way to handle this is the allocate an array with the maximum size, which is 65 KB. I didn't want to allocate such a large array just for an informative log. but maybe its not such a big deal. I made a new function to get the groups that returns an std::vector. I use the max size instead. and this function is also used to get orig_groups as it is the same

src/manage_nsfs/manage_nsfs_cli_utils.js
```diff
@@ -90,8 +90,18 @@ function get_boolean_or_string_value(value) {
     }
 }
 
+function parse_comma_delimited_string(value) {
```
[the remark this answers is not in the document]
nadavMiz: in case its undefined I assign the previous value. so I need to check it outside, but I add a case for undefined in here anyway

src/manage_nsfs/manage_nsfs_cli_utils.js
```diff
@@ -90,8 +90,18 @@ function get_boolean_or_string_value(value) {
...
+    if (typeof value === 'object') {
+        return value;
+    }
+    if (typeof value === 'number') {
+        return [value];
+    }
+    return value.split(',').map(val => Number(val));
```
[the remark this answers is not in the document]
nadavMiz: considering we by this point we validate that this value can only be string of the correct format (comma delimited list of positive numbers). it can't fail. even if its a single number it will be correct. if by any chance we get a value that doesn't fit the format this function will return 'nan'

src/util/native_fs_utils.js
```diff
@@ -520,7 +520,8 @@ async function get_fs_context(nsfs_account_config, fs_backend) {
         uid: (account_ids_by_dn && account_ids_by_dn.uid) ?? nsfs_account_config.uid,
         gid: (account_ids_by_dn && account_ids_by_dn.gid) ?? nsfs_account_config.gid,
         warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
-        backend: fs_backend
+        backend: fs_backend,
+        ...(nsfs_account_config.supplemental_groups && {supplemental_groups: nsfs_account_config.supplemental_groups})
```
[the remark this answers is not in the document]
nadavMiz: because napi doesn't accept undefined value for array. so this value needs to be either a list or doesn't exist at all. napi accepts undefined for int type, if the other values can even be undefined

src/manage_nsfs/manage_nsfs_cli_utils.js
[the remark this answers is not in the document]
nadavMiz: I changed it to also run split only if its a string type. for all other types not covered by this function, including undefined I change it to just return the value

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -1004,6 +1070,35 @@ describe('manage nsfs cli account flow', () => {
...
+                const expected_groups = [303];
+                const account_options = { config_root, name, supplemental_groups};
+                await exec_manage_cli(type, ACTIONS.UPDATE, account_options);
+                const account = await config_fs.get_account_by_name(name, config_fs_account_options);
+                expect(account.nsfs_account_config.supplemental_groups).toStrictEqual(expected_groups);
+            });
+
+            it('cli account add - cli update account with supplemental groups - single group (number))', async function() {
+                const { name } = defaults;
+                const supplemental_groups = 303;
```
shirady, line 1096: If it is a number 303 don't we still expect an array [303] as a saved result?

src/util/native_fs_utils.js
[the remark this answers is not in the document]
shirady: Could you add a comment about this?
Maybe an if can be more readable? (not sure)

src/manage_nsfs/manage_nsfs_cli_errors.js
```diff
@@ -346,6 +346,11 @@ ManageCLIError.InvalidGlacierOperation = Object.freeze({
     message: 'only "migrate", "restore" and "expiry" subcommands are supported',
     http_code: 400,
 });
+ManageCLIError.InvalidSupplementalGroupsList = Object.freeze({
+    code: 'InvalidSupplementalGroupsList',
+    message: 'supplemental groups must be a list of group ids (positive integers)',
```
shirady, line 351: I think we need to address the case of 0 in this message.
0 is neither positive and negative, but can be one of the groups, right?
Something like:
```suggestion
    message: 'supplemental groups must be a list of group ids (group id is zero or a positive integer)',
```

src/manage_nsfs/manage_nsfs_constants.js
```diff
@@ -44,16 +44,16 @@ const FROM_FILE = 'from_file';
 const ANONYMOUS = 'anonymous';
 
 const VALID_OPTIONS_ACCOUNT = {
-    'add': new Set(['name', 'uid', 'gid', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', FROM_FILE, ...CLI_MUTUAL_OPTIONS]),
-    'update': new Set(['name', 'uid', 'gid', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', 'new_name', 'regenerate', ...CLI_MUTUAL_OPTIONS]),
+    'add': new Set(['name', 'uid', 'gid', 'supplemental_groups', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', FROM_FILE, ...CLI_MUTUAL_OPTIONS]),
```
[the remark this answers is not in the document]
nadavMiz: supplemental group correct type should be a positive numbers array. but in javascript this just evaluates as 'object'. so in this case I will always wan to go to special validation. I can set it to be 'number[]' just as information. even though there is no such type

src/native/fs/fs_napi.cpp
```diff
@@ -586,6 +624,9 @@ struct FSWorker : public Napi::AsyncWorker
             if (fs_context.Get("backend").ToBoolean()) {
                 _backend = fs_context.Get("backend").ToString();
             }
+            if (fs_context.Has("supplemental_groups")) {
+                _supplemental_groups = convert_napi_number_array_to_number_vector<gid_t>(fs_context.Get("supplemental_groups").As<Napi::Array>());
```
shirady, line 628: I just want to make sure - in case somehow you gey empty array (`[]`) what will happen?

src/native/util/os_darwin.cpp
```diff
@@ -37,12 +39,20 @@ get_current_uid()
 
 const uid_t ThreadScope::orig_uid = getuid();
 const gid_t ThreadScope::orig_gid = getgid();
+const std::vector<gid_t> ThreadScope::orig_groups = get_process_groups();
+
 
 void
 ThreadScope::change_user()
```
shirady, line 46: Why only in Linux we have a function description? (on the same function `change_user`).

src/native/util/os_linux.cpp
```diff
@@ -27,25 +27,24 @@ get_current_uid()
...
-}();
+const std::vector<gid_t> ThreadScope::orig_groups = get_process_groups();
 
 /**
- * set the effective uid/gid of the current thread using direct syscalls
- * unsets the supplementary group IDs for the current thread using direct syscall
+ * set the effective uid/gid/supplemental_groups of the current thread using direct syscalls
  * we have to bypass the libc wrappers because posix requires it to syncronize
- * uid & gid to all threads which is undesirable in our case.
+ * uid, gid & supplemental_groups to all threads which is undesirable in our case.
```
shirady, line 35: Do you know if this comment is still relevant?
If so, could you explain it?

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -504,6 +504,72 @@ describe('manage nsfs cli account flow', () => {
...
+
+        it('cli account add - cli create account with supplemental groups)', async function() {
+            const { type, new_buckets_path, uid, gid, name } = defaults;
+            const supplemental_groups = '303,211';
+            const expected_groups = [303, 211];
+            const account_options = { config_root, name, new_buckets_path, uid, gid, supplemental_groups};
+            const action = ACTIONS.ADD;
+            await fs_utils.create_fresh_path(new_buckets_path);
+            await fs_utils.file_must_exist(new_buckets_path);
+            await set_path_permissions_and_owner(new_buckets_path, account_options, 0o700);
```
shirady, line 516: In all these tests we are on the `/tmp` path, right? It could be problematic with the sticky bit (as mentioned in your docs) once someone will try to use it with bucket policy?
If we can, we might want to use another `new_buckets_path` and explain why, or add a comment about it so others will understand.

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -504,6 +504,72 @@ describe('manage nsfs cli account flow', () => {
...
+            await exec_manage_cli(type, action, account_options);
+            const account = await config_fs.get_account_by_name(name, config_fs_account_options);
+            expect(account.nsfs_account_config.supplemental_groups).toStrictEqual(expected_groups);
+        });
+
+        it('cli account add - cli create account with supplemental groups - single group (string))', async function() {
+            const { type, new_buckets_path, uid, gid, name } = defaults;
+            const supplemental_groups = "303";
```
shirady, line 524: Please change single quotes to double quotes.
```suggestion
            const supplemental_groups = '303';
```

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -504,6 +504,72 @@ describe('manage nsfs cli account flow', () => {
...
+            await fs_utils.file_must_exist(new_buckets_path);
+            await set_path_permissions_and_owner(new_buckets_path, account_options, 0o700);
+            await exec_manage_cli(type, action, account_options);
+            const account = await config_fs.get_account_by_name(name, config_fs_account_options);
+            expect(account.nsfs_account_config.supplemental_groups).toStrictEqual(expected_groups);
+        });
+
+        it('cli account add - cli create account with supplemental groups - invalid list', async function() {
+            const { type, new_buckets_path, uid, gid, name } = defaults;
+            const supplemental_groups = '303l,201'; //group cannot contain charecters
```
shirady, line 552: Typo
```suggestion
            const supplemental_groups = '303l,201'; //group cannot contain characters
```

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -1004,6 +1070,36 @@ describe('manage nsfs cli account flow', () => {
                 const res = await exec_manage_cli(type, ACTIONS.UPDATE, account_options);
                 expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidAccountName.code);
             });
+
+            it('cli account add - cli update account with supplemental groups)', async function() {
```
shirady, line 1074: Now you're testing `account update`, please update the `it` title in this part.
Why do you need the last bracket?
```suggestion
            it('cli account update - cli update account with supplemental groups', async function() {
```

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -1004,6 +1070,36 @@ describe('manage nsfs cli account flow', () => {
...
+                const { name } = defaults;
+                const supplemental_groups = '303,211';
+                const expected_groups = [303, 211];
+                const account_options = { config_root, name, supplemental_groups};
+                await exec_manage_cli(type, ACTIONS.UPDATE, account_options);
+                const account = await config_fs.get_account_by_name(name, config_fs_account_options);
+                expect(account.nsfs_account_config.supplemental_groups).toStrictEqual(expected_groups);
+            });
+
+            it('cli account add - cli update account with supplemental groups - single group (string))', async function() {
```
shirady, line 1084: Same - 
Now you're testing account update, please update the it title in this part.
Why do you need the last bracket?
```suggestion
            it('cli account update - cli update account with supplemental groups - single group (string)', async function() {
```

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -1004,6 +1070,36 @@ describe('manage nsfs cli account flow', () => {
...
+                const { name } = defaults;
+                const supplemental_groups = '303';
+                const expected_groups = [303];
+                const account_options = { config_root, name, supplemental_groups};
+                await exec_manage_cli(type, ACTIONS.UPDATE, account_options);
+                const account = await config_fs.get_account_by_name(name, config_fs_account_options);
+                expect(account.nsfs_account_config.supplemental_groups).toStrictEqual(expected_groups);
+            });
+
+            it('cli account add - cli update account with supplemental groups - single group (number))', async function() {
```
shirady, line 1094: Same - 
Now you're testing account update, please update the it title in this part.
Why do you need the last bracket?
```suggestion
            it('cli account update - cli update account with supplemental groups - single group (number)', async function() {
```

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -1765,6 +1861,47 @@ describe('manage nsfs cli account flow', () => {
...
+            const command_flags = {config_root, from_file: path_to_option_json_file_name};
+            // create the account and check the details
+            const res = await exec_manage_cli(type, action, command_flags);
+            expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidSupplementalGroupsList.code);
+        });
+
+        it('should fail - cli create account using from_file with invalid supplemental_groups (string)', async () => {
+            const action = ACTIONS.ADD;
+            const { name, new_buckets_path, uid, gid } = defaults;
+            const supplemental_groups = ["212d", 112]; // gid cannot a non number type
```
shirady, line 1896: Please change single quotes to double quotes.
```suggestion
            const supplemental_groups = ['212d', 112]; // gid cannot a non number type
```

src/test/unit_tests/jest_tests/test_nc_nsfs_anonymous_cli.test.js
```diff
@@ -123,6 +123,58 @@ describe('manage nsfs cli anonymous account flow', () => {
...
             expect(JSON.parse(resp.stdout).error.message).toBe(ManageCLIError.InvalidArgument.message);
         });
+
+        it('cli create anonymous account with supplemental_groups - string', async () => {
+            let action = ACTIONS.DELETE;
+            const { type, anonymous, uid, gid } = defaults;
+            let account_options = { anonymous, config_root };
+            const resp = await exec_manage_cli(type, action, account_options);
+            const res_json = JSON.parse(resp.trim());
+            expect(res_json.response.code).toBe(ManageCLIResponse.AccountDeleted.code);
```
shirady, line 133: Could you please add a comment about it? (it is part of the setup, remove existing anonymous account).
This is relevant also for the next tests you added here.

src/test/unit_tests/jest_tests/test_nc_nsfs_anonymous_cli.test.js
```diff
@@ -123,6 +123,58 @@ describe('manage nsfs cli anonymous account flow', () => {
...
+
+            action = ACTIONS.ADD;
+            const supplemental_groups = "0,202,101";
```
shirady, line 136: Please change single quotes to double quotes.
```suggestion
            const supplemental_groups = '0,202,101';
```

src/test/unit_tests/jest_tests/test_nc_nsfs_anonymous_cli.test.js
```diff
@@ -123,6 +123,58 @@ describe('manage nsfs cli anonymous account flow', () => {
...
+            const supplemental_groups = '303g,202,101';
+            const account_options = { anonymous, config_root, uid, gid, supplemental_groups };
+            const resp = await exec_manage_cli(type, action, account_options);
+            expect(JSON.parse(resp.stdout).error.message).toBe(ManageCLIError.InvalidSupplementalGroupsList.message);
+        });
+
+        it('should fail - cli create anonymous account with invalid supplemental_groups - negative number', async () => {
+            const action = ACTIONS.ADD;
+            const { type, uid, gid, anonymous } = defaults;
+            const supplemental_groups = "0,-202,101";
```
shirady, line 173: Please change single quotes to double quotes.
```suggestion
            const supplemental_groups = '0,-202,101';
```

src/test/unit_tests/test_nsfs_access.js
```diff
@@ -18,8 +18,11 @@ mocha.describe('new tests check', async function() {
     const p = '/tmp/dir/';
     const root_dir = 'root_dir';
     const non_root_dir = 'non_root_dir';
+    const non_root_dir2 = 'non_root_dir2';
     const full_path_root = path.join(p, root_dir);
-    const full_path_non_root = path.join(p, non_root_dir);
+    const full_path_non_root = path.join(full_path_root, non_root_dir);
```
shirady, line 23: Why did you change it? Now it is `/tmp/dir/root_dir/non_root_dir`?
```suggestion
    const full_path_non_root = path.join(p, non_root_dir);
```

src/test/unit_tests/test_nsfs_access.js
```diff
@@ -40,13 +43,20 @@ mocha.describe('new tests check', async function() {
...
+        uid: 1574,
+        gid: 1574,
+        backend: '',
+        supplemental_groups: [1572, 1577], //gid of non-root1 and unrelated gid
+        warn_threshold_ms: 100,
+    };
     mocha.before(async function() {
         if (test_utils.invalid_nsfs_root_permissions()) this.skip(); // eslint-disable-line no-invalid-this
-        await fs_utils.create_fresh_path(p, 0o770);
+        await fs_utils.create_fresh_path(p, 0o777);
```
shirady, line 56: Why do you change the permission?

src/test/unit_tests/test_nsfs_access.js
```diff
@@ -99,6 +109,24 @@ mocha.describe('new tests check', async function() {
...
+        await nb_native().fs.mkdir(NON_ROOT1_FS_CONFIG, full_path_non_root1, 0o770);
+        //non root3 has non-root1 group as supplemental group, so it should succeed
+        const non_root_entries = await nb_native().fs.readdir(NON_ROOT3_FS_CONFIG, full_path_non_root1);
+        assert.equal(non_root_entries && non_root_entries.length, 0);
+    });
+
+    mocha.it('NON ROOT 3 without the files gid inside the suplemental group - failure', async function() {
+        await nb_native().fs.mkdir(NON_ROOT2_FS_CONFIG, full_path_non_root2, 0o770);
+        try {
+            //non root3 doesn't have non-root1 group as supplemental group, so it should fail
```
shirady, line 123: I think there is a typo in the comment
```suggestion
            //non root3 doesn't have non-root2 group as supplemental group, so it should fail
```

src/test/unit_tests/test_nsfs_access.js
```diff
@@ -99,6 +109,24 @@ mocha.describe('new tests check', async function() {
...
     });
+
+    mocha.it('NON ROOT 3 with suplemental group - success', async function() {
+        await nb_native().fs.mkdir(NON_ROOT1_FS_CONFIG, full_path_non_root1, 0o770);
+        //non root3 has non-root1 group as supplemental group, so it should succeed
+        const non_root_entries = await nb_native().fs.readdir(NON_ROOT3_FS_CONFIG, full_path_non_root1);
+        assert.equal(non_root_entries && non_root_entries.length, 0);
+    });
+
+    mocha.it('NON ROOT 3 without the files gid inside the suplemental group - failure', async function() {
```
shirady, line 120: I think the title is not clear

src/util/native_fs_utils.js
```diff
@@ -516,12 +516,18 @@ async function get_fs_context(nsfs_account_config, fs_backend) {
     if (nsfs_account_config.distinguished_name) {
         account_ids_by_dn = await get_user_by_distinguished_name({ distinguished_name: nsfs_account_config.distinguished_name });
     }
-    return {
+    const fs_context = {
         uid: (account_ids_by_dn && account_ids_by_dn.uid) ?? nsfs_account_config.uid,
         gid: (account_ids_by_dn && account_ids_by_dn.gid) ?? nsfs_account_config.gid,
         warn_threshold_ms: config.NSFS_WARN_THRESHOLD_MS,
-        backend: fs_backend
+        backend: fs_backend,
```
shirady, line 523: You can remove the last comma
```suggestion
        backend: fs_backend
```

src/manage_nsfs/manage_nsfs_constants.js
```diff
@@ -105,6 +105,7 @@ const OPTION_TYPE = {
     owner: 'string',
     uid: 'number',
     gid: 'number',
+    supplemental_groups: 'number[]', //positive integer array. but will be evaluated as 'object' by javascript. so always need special validation
```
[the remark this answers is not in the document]
nadavMiz: yes it will always be false. which is fine because we always want to go to the validation function

src/native/fs/fs_napi.cpp
nadavMiz: it will be fine, it will just create an empty vector

src/native/util/os_darwin.cpp
nadavMiz: I can add mac description as well

src/native/util/os_linux.cpp
nadavMiz: yes it is. I assume the libc functions change the uid, gid, supplemental groups for all threads. we just want to change it for one thread. so we use special functions. we need to validate it works on mac as well

src/test/unit_tests/test_nsfs_access.js
nadavMiz: because in order that my new tests will work i needed the root directory to have full permissions. but the existing test expect it to have partial permissions. so I moved it into a folder with partial permissions. maybe I could do it reverse where my tests were part of different directory, but I don't think it matters that much

src/test/unit_tests/test_nsfs_access.js
nadavMiz: same reason as the comment above. for my tests I need root to have full permissions. otherwise non root users cant create new directories

src/native/util/os_darwin.cpp
```diff
@@ -37,12 +40,30 @@ get_current_uid()
...
+ */
 void
 ThreadScope::change_user()
 {
     if (_uid != orig_uid || _gid != orig_gid) {
         MUST_SYS(_mac_thread_setugid(_uid, _gid));
+        if (!_groups.empty()) {
+            _groups.push_back(_gid);
+            std::swap(_groups.front(), _groups.back());
+            MUST_SYS(setgroups(_groups.size(), &_groups[0]));
```
shirady, line 65: Could you add a comment about this change?

src/test/unit_tests/test_nsfs_access.js
```diff
@@ -99,6 +110,28 @@ mocha.describe('new tests check', async function() {
...
+        //TODO on mac new directories are created with the parents directory GID and not with the process GID. manually change the gid
+        fs.promises.chown(full_path_non_root1, NON_ROOT1_FS_CONFIG.uid, NON_ROOT1_FS_CONFIG.gid);
+        //non root3 has non-root1 group as supplemental group, so it should succeed
+        const non_root_entries = await nb_native().fs.readdir(NON_ROOT3_FS_CONFIG, full_path_non_root1);
+        assert.equal(non_root_entries && non_root_entries.length, 0);
+    });
+
+    mocha.it('NON ROOT 3 suplemental group without the files gid - failure', async function() {
+        //TODO on mac new directories are created with the parents directory GID and not with the process GID. manually change the gid
+        fs.promises.chown(full_path_non_root1, NON_ROOT2_FS_CONFIG.uid, NON_ROOT2_FS_CONFIG.gid);
```
shirady, line 125: Why is it here and not after the `nb_native().fs.mkdir`?
Why the `fs.promises.chown` is on `full_path_non_root1` and not `full_path_non_root2`?

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
nadavMiz: these are cli tests  and not functionality tests. so this will not be an issue in this case. we tested functionality tests and saw they work on mac with private/tmp

src/test/unit_tests/test_nsfs_access.js
nadavMiz: changed it, thanks

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -197,6 +201,35 @@ function validate_boolean_string_value(value) {
     return false;
 }
 
+/**
+ * validates supplemental groups array. verifies that string is comma seperated positive numbers. string should not
+ * begin or end with a comma. if only one number it should be positive
```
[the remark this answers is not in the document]
nadavMiz: changed it

src/native/fs/fs_napi.cpp
```diff
@@ -537,6 +537,42 @@ load_xattr_get_keys(Napi::Object& options, std::vector<std::string>& _xattr_get_
     }
 }
 
+/**
+* converts Napi::Array of numbers to std::vector
+* typename T - type of the vector to convert to (e.g int, uint, gid_t)
```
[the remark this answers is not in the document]
nadavMiz: there is no real good way to staticly restrict this. there is a way to staticly restrict to arithmetic types, but not all arithmetic types will work, so it might just be confusing. I could add type check for the napi value isNumber, or add list of supported types. but at best it will replace the runtime c exception to my own exception. I added a warning in the function description in case the function name is not clear enough, also if someone try to run this with wrong types it will either not compile because napi number cant convert to that time, or will throw an exception at runtime

src/test/unit_tests/jest_tests/test_nc_nsfs_anonymous_cli.test.js
nadavMiz: this is just to delete the existing account so a new one can be created. it is done in other tests as well. I have added a comment about this

src/native/util/os_darwin.cpp
```diff
@@ -37,12 +40,34 @@ get_current_uid()
...
+ * https://github.com/apple/darwin-xnu/blame/main/bsd/kern/kern_prot.c (see setgroups1)
+ */
 void
 ThreadScope::change_user()
 {
     if (_uid != orig_uid || _gid != orig_gid) {
         MUST_SYS(_mac_thread_setugid(_uid, _gid));
+        if (!_groups.empty()) {
+            /*accourding to BSD Manual https://man.freebsd.org/cgi/man.cgi?query=setgroups
+            which darwin is occationally complient to setgroups changes the effective gid according to
```
shirady, line 65: typo
```suggestion
            which darwin is occasionally compliant to setgroups changes the effective gid according to
```

shirady (approved):
LGTM

PR #8556, opened 2024-11-25
Removing K-means and test performance from node allocator

### LINKED ISSUE #8496: Node allocator latency_groups causes just single node to be used for writes
### Environment info
- NooBaa Version: **5.15**
- Platform: **ODF-4.15**

### Actual behavior
1. Using a PVPool of 2 or 4 we can see that just a single PV writes at a time.
2. This happens due to this code - https://github.com/noobaa/noobaa-core/blob/b27e63bff351a3f23789e5ac512cc87bf633d4a1/src/server/node_services/nodes_monitor.js#L3494
3. We can workaround this issue by setting the following config which eliminated this code path:
```
oc set env sts noobaa-core CONFIG_JS_NODE_ALLOCATOR_NUM_CLUSTERS=99999
```

### Expected behavior
1. All PV's should be writing in parallel.

### Steps to reproduce
1. Write to a bucket that uses a PVpool.

This is a benchmark running before the workaround - in the write stage you can see a single PV active, and in the read stage both are active:
image

After the workaround both are writing together.

jackyalbo:
### Explain the changes
1. Removing K-means and test performance from node allocator

1.  Fixed #8496 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/node_services/nodes_monitor.js
+++ b/src/server/node_services/nodes_monitor.js
@@ -8,7 +8,6 @@ const chance = require('chance')();
 // const dclassify = require('dclassify');
 const EventEmitter = require('events').EventEmitter;
 
-const kmeans = require('../../util/kmeans');
 const P = require('../../util/promise');
 const api = require('../../api');
 const pkg = require('../../../package.json');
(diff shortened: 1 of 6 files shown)


nimrod-becker:
I agree since we don't expect a single pv pool to be spread across GEOs, or have different storage types. That can ofc be a difference between two different pv pools, but then this code is no relevant so much.

src/server/node_services/nodes_monitor.js
```diff
@@ -1386,44 +1382,10 @@ class NodesMonitor extends EventEmitter {
         }
     }
 
-    async _test_store_perf(item) {
```
dannyzaken, line 1389: instead of this function that stored the results to DB, add a log message in `blcok_store_base.test_store_perf` that prints the results to log1

dannyzaken (approved):

PR #8561, opened 2024-11-28
nsfs - monitor only nsrs that are mounted. DFBUGS-153

alphaprinz:
### Explain the changes
1. Namespace monitor should not monitor nsfs nsrs that are not mounted on the endpoint.
Use the new NSFS_NSR_ env variable to test whether nsr should be mounted.
(see https://github.com/noobaa/noobaa-operator/pull/1481)

This commit reverts 2789d60757b925cba05d0062ef677ca0bcc4d16b.

1. https://issues.redhat.com/browse/DFBUGS-153

### Testing Instructions:
Reduce config.NAMESPACE_MONITOR_DELAY to 1000ms.
Create nsfs nsr
nsr should not be rejected (by endpoints existing before its creation).

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -213,16 +213,12 @@ async function main(options = {}) {
         // there for namespace monitor won't be registered
         if (internal_rpc_client && config.NAMESPACE_MONITOR_ENABLED) {
             endpoint_stats_collector.instance().set_rpc_client(internal_rpc_client);
-
-            //wait with monitoring until pod has started
-            setTimeout(() => {
-                // Register a bg monitor on the endpoint
-                background_scheduler.register_bg_worker(new NamespaceMonitor({
-                    name: 'namespace_fs_monitor',
-                    client: internal_rpc_client,
-                    should_monitor: nsr => Boolean(nsr.nsfs_config),
-                }));
-            }, 1000 * 60);
+            // Register a bg monitor on the endpoint
+            background_scheduler.register_bg_worker(new NamespaceMonitor({
+                name: 'namespace_fs_monitor',
+                client: internal_rpc_client,
+                should_monitor: nsr => Boolean(nsr.nsfs_config && process.env['NSFS_NSR_' + nsr.name]),
+            }));
         }
 
         if (config.ENABLE_SEMAPHORE_MONITOR) {


src/endpoint/endpoint.js
```diff
@@ -207,16 +207,12 @@ async function main(options = {}) {
...
-                    name: 'namespace_fs_monitor',
-                    client: internal_rpc_client,
-                    should_monitor: nsr => Boolean(nsr.nsfs_config),
-                }));
-            }, 1000 * 60);
+            // Register a bg monitor on the endpoint
+            background_scheduler.register_bg_worker(new NamespaceMonitor({
+                name: 'namespace_fs_monitor',
+                client: internal_rpc_client,
+                should_monitor: nsr => Boolean(nsr.nsfs_config && process.env['NSFS_NSR_' + nsr.name]),
```
romayalon, line 214: A few questions - 
1. If the endpoint got up before the namespace resource was mounted, when is the next time we will get to this flow for start monitoring? why not add a retry after 60 seconds -
https://github.com/noobaa/noobaa-core/pull/8474#discussion_r1818144231
2. Why avoid start monitoring instead of externalizing that the value of process.env['NSFS_NSR_' + nsr.name] is undefined which means that the PV was not mounted yet?

alphaprinz: 1. An endpoint that was started before the mount will be deleted after a new endpoint is created with the new mount.
The retry will not help as the endpoint that opened the report is removed when the new nsfs nsr mount is added (after nsfs nsr was created in kubernetes cluster).
2. There could be other nsfs nsrs that should be monitored (correct me if I'm wrong).

Maybe I'll make the scenario more concrete-
1. Operator install a system in a cluster.
2. There is endpoint A. It does NOT have any nsfs nsr mounts.
3. At some point, an nsfs nsr is created in the cluster.
4. In reconcile:
4. a. operator adds a mount for the nsfs nsr to endpoints' container.
4. b. operator creates an nsr object in system store.
5. A new endpoint B with the new mount is created by kubernetes.
6. While B is being created, A updates its system store, reads the nsfs nsr. The new nsfs nsr is NOT mounted in A. A reports NOENT on the nsfs nsr. Note since default interval for nsfs nsr monitoring is less than creating a new endpoint, this doesn't necessarily happen. Reducing config.NAMESPACE_MONITOR_DELAY will ensure bug reproduction.
7. Endpoint B is ready. Endpoint A is deleted. Nsr status is stuck in rejected.

romayalon: As I mentioned on Slack, I think that the correct path is not to avoid monitoring a namespace resource that is still not mounted but add this check to the monitoring process.

alphaprinz: The nsr will be monitored by the new endpoint.
The old endpoint is about to be deleted.
The only difference this commit makes is that old endpoints won't mistakenly report nsr as rejected.

romayalon: "About-to-be-deleted" is the happy path :) 
There is also the sad path where there is an issue with the mounting and it takes a while/never happens - that's exactly why I think it's important and avoiding monitoring it if it's not mounted is a partial solution.

alphaprinz: I'm not trying to solve monitoring, but rather to fix a bug in monitoring.
I'm not removing any feature that we currently have.

Again, I would like a more specific way to proceed.
If you think a different fix or an enhancement to the monitoring is needed, please specify it explicitly.

romayalon: @alphaprinz 
I already explained it in the above comment, but I'll be happy to summarize my comments - 
 
**My specific idea for solving it  -** 
Instead of not monitoring unmounted namespace resources, I think you should move the new condition you added inside the monitoring check, and externalize that this is the current issue that the namespace resource has.
[Comment 1, bullet 2](https://github.com/noobaa/noobaa-core/pull/8561#discussion_r1867261370)
[Comment 3](https://github.com/noobaa/noobaa-core/pull/8561#discussion_r1868789955)

**Why I think that my suggestion is a better behavior / user experience -** 
It will behave better in cases where the re-start of the endpoint takes time/won't happen at all.
[Comment 5](https://github.com/noobaa/noobaa-core/pull/8561#discussion_r1870716003)

**How to proceed -** 
The above summary of comments is my opinion/suggestion/how I would fix it.
IMO, You shall proceed from here as how you see it, fix it, open an issue and call it enhancement, document this gap or anything else **you** feel appropriate.

romayalon (approved):
I'm approving, 
for the protocol, I had an idea which I think will be better UX but let's let QA find the theoretical issue I suggested.

PR #8571, opened 2024-12-03
PublicAccessBlock Support

tangledbytes:
### Explain the changes
This PR attempts to add basic support for PublicAccessBlock.

1. https://issues.redhat.com/browse/RHSTOR-6623

### Testing Instructions:
1. `./node_modules/.bin/jest src/test/unit_tests/jest_tests/test_s3_utils.test.js`
2. `./node_modules/.bin/mocha src/test/unit_tests/test_public_access_block.js`
3. `NC_CORETEST=true NC_NSFS_NO_DB_ENV=true ./node_modules/.bin/mocha src/test/unit_tests/test_public_access_block.js`

- [ ] Doc added/updated
- [ ] Tests added


src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -168,14 +173,19 @@ function _is_action_fit(method, statement) {
     return statement.Action ? action_fit : !action_fit;
 }
 
-function _is_principal_fit(account, statement) {
+function _is_principal_fit(account, statement, ignore_anon_principal = false) {
```
jackyalbo, line 176: why not call it "disallow_public_access" as in the parts of the flow? I think this name is more clear; we know why we should ignore the fit.

jackyalbo (commented):
In general looks ok to me, maybe also ask someone from nc to take a look.
adding tests will be very valuable here.

src/endpoint/s3/ops/s3_put_bucket_public_access_block.js
```diff
@@ -0,0 +1,25 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+
+const s3_utils = require('../s3_utils');
+
+/**
+ * https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutPublicAccessBlock.html
+ * @param {*} req
```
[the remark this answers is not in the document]
tangledbytes: Done

src/endpoint/s3/s3_bucket_policy_utils.js
jackyalbo: I'm just not sure I fully get the flow. If we want to disallow public access. Shouldn't we have it only for `Allow` statement and not for `Deny`? It seems to me that we should not ignore a fit in case of a Deny statement.
Also, If the issue is with semantics, we need a better name for this as * is not just for anonymous principals it is for any principal. 
Maybe some more comments in this area will be good to make me and future code readers better understand its use.

jackyalbo: @tangledbytes is this relevant also for deny?

tangledbytes: Ah! I am actually sorry. I actually thought that I responded here but turns out I forgot. Yes, you are right, I should not ignore it in case of `DENY`.

Also yes, it sounds right that `anon` isn't the right word. I will replace it with `public_principla`? How's that?

Again. apologies, missed this 😅

jackyalbo (approved):

PR #6928, opened 2022-03-29
Chunk fs | EINVAL error when writing more than 1024 buffers at one writev fix

romayalon:
Signed-off-by: Romy <romy2232@gmail.com>

### Explain the changes
1. According to https://linux.die.net/man/2/writev - calling writev() with iovcnt (buffer_len /  iov_vec.size() ) larger than 1024 will cause EINVAL (Invalid argument) error which we saw in https://github.com/noobaa/noobaa-core/issues/6896. 
2. In chunk_fs _transform() added ```this.q_buffers.length === config.NSFS_IOV_MAX``` to the if clause that will result flush_buffers() when evaluated to true.
3. Added a NSFS_IOV_MAX 1024 to config.
4. Added a unit test.

### Issues: Fixed #xxx / Gap #xxx
1. will fix https://github.com/noobaa/noobaa-core/issues/6896

### Testing Instructions:
1. 

- [ ] Doc added/updated
- Tests added

PR #8577, opened 2024-12-04
NSFS | Replace ChunkFS with FileWriter

### LINKED ISSUE #8574: NSFS | ChunkFS | Inconsistent data corruption when copying using Spark
### Environment info
- NooBaa Version: 5.18
- Platform: All

### Actual behavior
1. When running Spark for copying parquet files in a loop, sometimes the parquet file appeared to be corrupted.
2. Examples - 
a. See the parquet file having `PAR1PAR1` at the end
```bash
$ cat /buck1/p607/part-00000-4754ebf0-626a-45ae-b7f3-130bba23db92-c000.snappy.parquet
PAR1BF����
         !�This is the first line.This is the first line.This is the first linet,H
                                                                                  spark_schema
                                                                                              %value%L&
                                                                                                      valuept<This is the first line.This is the first line.(This is the first line.This is the first line.�|zpt,org.apache.spark.version3.5.3)org.apache.spark.sql.parquet.row.metadata[{"type":"struct","fields":[{"name":"value","type":"string","nullable":true,"metadata":{}}]}Jparquet-mr version 1.13.1 (build db4183109d5b734ec5930d870cdae161e408ddba)�PAR1PAR1
```

b. See the parquet file having `PAR11` at the end
```bash
$ cat /buck1/p18/part-00001-5802a93c-f27c-4b57-9613-1e763b77a9e4-c000.snappy.parquet
PAR1BF���!�This is the third line.This is the third line.This is the third liner,H
                                                                                  spark_schema
                                                                                              %value%L&
                                                                                                      valuenr<This is the third line.This is the third line.(This is the third line.This is the third line.�zznr,org.apache.spark.version3.5.3)org.apache.spark.sql.parquet.row.metadata[{"type":"struct","fields":[{"name":"value","type":"string","nullable":true,"metadata":{}}]}Jparquet-mr version 1.13.1 (build db4183109d5b734ec5930d870cdae161e408ddba)�PAR11
```
Additional details of Spark internal "COPY" work - 
1. Spark writes the data to a file called /buck1/.../attempt...../part-0000...
4. copies the data (internal fallback - read and write)  to a file called /buck1/.../"task"..... - the corruption happens here - read 3. correctly, the ChunkFS,flush_buffers() shows corrupted last 8 bytes
5. copies the data (internal fallback - read and write)  to a file called /buck1/part-0000... - last 8 bytes look the same as in step 2

### Expected behavior
1. All files copied using Spark through NooBaa should have their expected data as their source.

### Steps to reproduce
1. The following steps were executed on a node having Spark shell (Scala) running and NooBaa installed - 
2. Open 2 tabs
4. First tab - where NooBaa is installed - 4. Run `journalctl -u noobaa -f > noobaa_during_spark_run.log` to get the logs of noobaa during the reproduction
6. Second tab - on the Spark shell run the following commands - 
```scala
// 1. Create a test txt file - 
val data = Seq("This is the first line.", "This is the second line.", "This is the third line.")
val rdd = spark.sparkContext.parallelize(data)
val textFilePath = "output/textFile"
rdd.saveAsTextFile(textFilePath)

// 2. Read the text file and convert it to a parquet file, write the parquet file to S3 (NooBaa) -
 val df = spark.read.text(textFilePath)

val structuredDF = df.map(row => {
val line = row.getString(0) // Extract the text line
val parts = line.split(",") // Example: split by commas
(parts(0), parts(1)) // Example: assume two parts
}).toDF("column1", "column2")

structuredDF.write.parquet("s3a://buck1/p")

// 3. Copy the parquet file in a loop -

for (i <- 1 to 300) {
  val newPath = "s3a://buck1/p" + i;
  spark.read.parquet("s3a://buck1/p").write.parquet(newPath)
}

// 4. Read the new files in a loop - expect a failure in the reading of the parquet files -  
for (i <- 1 to 300) {
  val newPath = "s3a://buck1/p" + i;
  println(s"i (input: $newPath):")
  val df = spark.read.parquet(newPath)
  df.show()
}
```

romayalon:
### Explain the changes
1. Added @guymguym's changes to Replace ChunkFS with FileWriter.
2. Applied the existing copy changes, removed the readable resume().
3. Replaced ChunkFS usage in other files to FileWriter usage.

1. Fixed https://github.com/noobaa/noobaa-core/issues/8574

### Testing Instructions:
1. See detailed instructions in the issue.
4. test_namespace_fs
5. test_file_writer
6. node /noobaa-core/src/tools/file_writer_hashing.js --hash
7. node /noobaa-core/src/tools/file_writer_hashing.js --file

- [ ] Doc added/updated
- [ ] Tests added


src/sdk/namespace_fs.js
```diff
@@ -1564,36 +1564,35 @@ class NamespaceFS {
...
-        const { source_stream, copy_source } = params;
+        const { copy_source } = params;
         try {
-            // Not using async iterators with ReadableStreams due to unsettled promises issues on abort/destroy
-            const md5_enabled = this._is_force_md5_enabled(object_sdk);
-            const chunk_fs = new ChunkFS({
+            const md5_enabled =
+                config.NSFS_CALCULATE_MD5 ||
+                this.force_md5_etag ||
+                object_sdk?.requesting_account?.force_md5_etag;
```
guymguym, line 1572: better use _is_force_md5_enabled() - it probably didn't exist when I wrote this code

src/sdk/namespace_fs.js
```diff
@@ -1564,36 +1564,35 @@ class NamespaceFS {
...
-                large_buf_size: multi_buffer_pool.get_buffers_pool(undefined).buf_size
+                large_buf_size: multi_buffer_pool.get_buffers_pool(undefined).buf_size,
+                namespace_resource_id: this.namespace_resource_id,
             });
-            chunk_fs.on('error', err1 => dbg.error('namespace_fs._upload_stream: error occured on stream ChunkFS: ', err1));
-            chunk_fs.on('finish', arg => dbg.error('namespace_fs._upload_stream: finish occured on stream ChunkFS: ', arg));
-            chunk_fs.on('close', arg => dbg.error('namespace_fs._upload_stream: close occured on stream ChunkFS: ', arg));
+            file_writer.on('error', err => dbg.error('namespace_fs._upload_stream: error occured on FileWriter: ', err));
+            file_writer.on('finish', arg => dbg.error('namespace_fs._upload_stream: finish occured on stream FileWriter: ', arg));
+            file_writer.on('close', arg => dbg.error('namespace_fs._upload_stream: close occured on stream FileWriter: ', arg));
```
guymguym, line 1585: are these error prints attempts to debug the Transform stream issues? still needed?

src/util/file_writer.js
```diff
@@ -0,0 +1,145 @@
...
+            });
+            await Promise.all([
+                this.MD5Async && this._update_md5(buffers, size),
+                this._write_all_buffers(buffers, size),
+            ]);
+            this._update_stats(size);
+            for (const buf of buffers) {
+                // @ts-ignore
+                if (buf.buffer_pool_release_callback) buf.buffer_pool_release_callback();
+            }
```
guymguym, line 117: this was a special code I added to test the nodejs changes I proposed for incoming http buffer pooling.
we can remove it for now.

src/sdk/namespace_fs.js
romayalon: yes because we wanted to see when the readable finishes,
if you don't think it can help in the future I can remove it, but anyway I'll change the dbg.error to dbg1

src/util/file_writer.js
```diff
@@ -0,0 +1,140 @@
...
+            await this._write_to_file(buffers, size);
+        } else {
+            let iov_start = 0;
+            while (iov_start < buffers.length) {
+                const iov_end = Math.min(buffers.length, iov_start + this.iov_max);
+                const buffers_to_write = buffers.slice(iov_start, iov_end);
+                const size_to_write = buffers_to_write.reduce((s, b) => s + b.length, 0);
+                await this._write_to_file(buffers_to_write, size_to_write);
+                iov_start = iov_end;
+            }
```
guymguym, line 83: since this case is very rare, and I don't know if our tests will easily get to it, I would like to suggest that in our unit tests we explicitly set iov_max to a low value such as 1 or 2 and make sure we test also this case.

romayalon: I added a test a few years ago for this iov_max case - https://github.com/noobaa/noobaa-core/pull/6928/files#diff-a5e4de68b0630a9cd0aab52d062a381429b3e890302696803ceadc2b1acb552aR22-R33
but anyway, I added in test_file_writer 4 new cases with your suggestion.

config.js
```diff
@@ -882,6 +882,8 @@ config.NSFS_LOW_FREE_SPACE_PERCENT_UNLEASH = 0.10;
 // anonymous account name
 config.ANONYMOUS_ACCOUNT_NAME = 'anonymous';
 
+config.NSFS_UPLOAD_STREAM_MAX_BUFFER_SIZE = 64 * 1024 * 1024;
```
guymguym, line 885: I am not sure if this is too much. The default in nodejs is 64KB and we do want to increase but maybe not so much. Perhaps it would be best that we set the default to 8MB and if needed we can instruct users how to increase further.

Also I think the name I used ..._MAX_BUFFER_SIZE might be thought to be something else so better rename to something like NFSF_UPLOAD_STREAM_MEM_THRESHOLD

see https://nodejs.org/docs/latest-v20.x/api/stream.html#buffering

guymguym (approved):

PR #8593, opened 2024-12-11
Bucket Notifications - PR notes, take 2

alphaprinz:
### Explain the changes
1. Switch connect file format kv -> json
2. Use only filename to specify connect file.
3. Use AWS s3api schema to store notifications conf.

1. Usability/ configuration burden
2. Schema s3 compatibility

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -1428,15 +1428,21 @@ module.exports = {
         },
         bucket_notification: {
             type: 'object',
-            required: ['Id', 'Connect'],
+            required: ['id', 'topic'],
             properties: {
-                Id: {
-                    type: 'string'
+                id: {
+                    type: 'array',
+                    items: {
+                        type: 'string',
+                    },
                 },
-                Connect: {
-                    type: 'string'
+                topic: {
+                    type: 'array',
+                    items: {
+                        type: 'string',
+                    },
                 },
-                Events: {
+                event: {
                     type: 'array',
                     items: {
                         type: 'string',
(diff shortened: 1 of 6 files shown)


src/util/notifications_util.js
```diff
@@ -140,6 +144,13 @@ class Notificator {
         //we can remove the currently processed persistent file
         return true;
     }
+
+    parse_connect_file(connect_filename) {
+        const connect_str = fs.readFileSync(path.join(this.connect_files_dir, connect_filename), 'utf-8');
```
romayalon, line 149: when interacting with config files in NC, we should read/write the file using config_fs functionalities
I would try use config_fs.get_config_data() and if it's not exactly what you need, we shall implement it in config_fs.

alphaprinz: For nsfs, config_fs will be used to load the conf file.
Otherwise (for containerized), same file api is used.

src/api/common_api.js
```diff
@@ -1428,15 +1428,21 @@ module.exports = {
         },
         bucket_notification: {
             type: 'object',
-            required: ['Id', 'Connect'],
+            required: ['Id', 'Topic'],
```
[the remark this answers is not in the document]
alphaprinz: The note from @guymguym was to use the same structure from aws cli all the way to the db.

romayalon (approved):

PR #8633, opened 2024-12-27
Fix for not accepting the past dates as expiration for bucket lifecycle rules

### LINKED ISSUE #8551: NSFS | S3 | Lifecycle: Expiration Date in the past gets accepted
### Environment info
- NooBaa Version: noobaa-core-5.17.1-20241104.el9.x86_64
- Platform: RHEL 9.2 x86

### Actual behavior
1. The expiration date in the past is getting accepted, example of the lifecycle rule which gets accepted:

{
    "Rules": [
        {
            "ID": "Expiration-date",
            "Status": "Enabled",
            "Filter": {
               "Prefix": "small"
            },
            "Expiration": { "Date": "2024-01-01T00:00:00.000Z" }
        }
     ]
}

### Expected behavior
1. If expiration date is in the past relative to current date the lifecycle rule should get rejected.

### Steps to reproduce
1. Please use the lifecycle rule above to recreate it.

aayushchouhan09:
### Explain the changes
1. Added a fix for not accepting past dates as expiration for bucket lifecycle rules also will reject if expiration date is not in correct format.

1. Fixed: #8551 

### Testing Instructions:
- [X] Tests added

### DIFF
--- a/src/test/lifecycle/common.js
+++ b/src/test/lifecycle/common.js
@@ -43,9 +43,9 @@ function empty_filter_marker_lifecycle_configuration(Bucket) {
 exports.empty_filter_marker_lifecycle_configuration = empty_filter_marker_lifecycle_configuration;
 
 function date_lifecycle_configuration(Bucket, Key) {
-    const now = new Date(Date.now());
-    now.setDate(now.getDate() + 1); // Increases date by 1 day; past/current dates can't be set
-    const midnight = new Date(now.setUTCHours(0, 0, 0, 0));
+    const date = new Date(Date.now());
+    date.setDate(date.getDate() + 1); // Increases date by 1 day; past/current dates can't be set
+    const midnight = new Date(date.setUTCHours(0, 0, 0, 0));
 
     return {
         Bucket,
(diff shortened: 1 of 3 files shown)


romayalon:
Hey @achouhan09 
1. Can you please link to the Issues section the bug you are fixing?
2. Please add tests

achouhan09:
Hi @romayalon 
Added issue link and tests. Thanks

src/test/lifecycle/common.js
```diff
@@ -44,6 +44,7 @@ exports.empty_filter_marker_lifecycle_configuration = empty_filter_marker_lifecy
 
 function date_lifecycle_configuration(Bucket, Key) {
     const now = new Date(Date.now());
+    now.setDate(now.getDate() + 1); // Increases date by 1 day; past/current dates can't be set
```
shirady, line 47: Could you change the name of the variable? as it is not "now" anymore... (I saw it in more places).

src/test/lifecycle/common.js
```diff
@@ -512,3 +539,13 @@ exports.test_and_prefix_size = async function(Bucket, Key, s3) {
...
 };
+
+exports.test_expiration_date_in_past = async function(Bucket, Key, s3) {
+    const putLifecycleParams = past_date_lifecycle_configuration(Bucket, Key);
+
+    await s3.putBucketLifecycleConfiguration(putLifecycleParams)
+        .catch(error => {
+            assert(error.code === 'InvalidArgument', 'Expected InvalidArgument: expiration date in the past');
+            console.log('Expected error received, expiration date in the past');
+        });
```
shirady, line 550: Is there a special reason for using the promise chain here?

src/test/lifecycle/common.js
```diff
@@ -63,8 +64,31 @@ function date_lifecycle_configuration(Bucket, Key) {
 }
 exports.date_lifecycle_configuration = date_lifecycle_configuration;
 
+function past_date_lifecycle_configuration(Bucket, Key) {
```
shirady, line 67: Could you reuse the previous definition of `date_lifecycle_configuration` and make the change you need on this function's output?

src/test/lifecycle/common.js
achouhan09: changed to "date"

src/test/lifecycle/common.js
achouhan09: Added try/catch instead of promise chain. Thanks

src/test/lifecycle/common.js
achouhan09: Done. Thanks

src/test/lifecycle/common.js
```diff
@@ -512,3 +517,20 @@ exports.test_and_prefix_size = async function(Bucket, Key, s3) {
...
 };
+
+exports.test_expiration_date_in_past = async function(Bucket, Key, s3) {
+    const putLifecycleParams = date_lifecycle_configuration(Bucket, Key);
+
+    // decrease expiration date by 2 days to set it in the past
+    const date = new Date(Date.now());
+    date.setDate(date.getDate() - 2);
+    const past_date = new Date(date.setUTCHours(0, 0, 0, 0));
+    putLifecycleParams.LifecycleConfiguration.Rules[0].Expiration.Date = past_date;
```
shirady, line 528: Maybe clone the object `putLifecycleParams` and then change it?
Because someone might write another test after and the object would be changed.

achouhan09: For every test we are initializing it, so I think we are good here. Thanks

shirady: Thanks, Where?

achouhan09: you can check the other test functions in the same file `common.js` , for every function we are initializing `putLifecycleParams` first and then we are calling the put_get_lifecycle_configuration(putLifecycleParams). Pls correct me if I am wrong

shirady: You're right, thank you.

achouhan09:
Closing the pull request as it is not a bug, for reference check aws documentation [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/intro-lifecycle-rules.html) . Thanks

PR #8637, opened 2024-12-30
CI | Reduce run time | Convert test_file_writer to jest

liranmauda:
### Explain the changes
- Create test_file_writer.test.js
- Delete test_file_writer.js
- Removed the entries of test_file_writer.js from the indexes

The tests take ~6 min.
by moving it to jest we increase the time jest tests are running and reduce the time the mocha are running.

This will help reduce the overall PR test time and balance the time between jest and mocha

the test is not affected by NC or containerized setups, as it runs locally.

### DIFF
--- a/src/test/unit_tests/index.js
+++ b/src/test/unit_tests/index.js
@@ -57,7 +57,6 @@ require('./test_bucket_chunks_builder');
 require('./test_mirror_writer');
 require('./test_namespace_fs');
 require('./test_ns_list_objects');
-require('./test_file_writer');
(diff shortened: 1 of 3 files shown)


romayalon:
@liranmauda awesome
The only concern I raised a few months ago and it's also a bit related to this PR was that we have a little bit of nsfs/nc tests on jest/mocha and we need to spend some time on organizing it, also we saw that the docs about it are not updated. I'll open a github issue about it.

romayalon (approved):

PR #8585, opened 2024-12-08
NC | NSFS | Add stat to `account_cache`

shirady:
### Explain the changes
1. In `account_cache` move the value that we had in `expiry_ms` to the `config.js` (with its comment of "TODO"), and add `_validate_account` function (like in `bucket_namespace_cache` we have the method `_validate_bucket_namespace`).
2. In `read_account_by_access_key` add the property `stat` so we can save the stat of the account config.
3. Rename `check_same_stat` to `check_same_stat_bucket` so we can have 2 methods for buckets and accounts. Also rename `ns_allow_stat_bucket` to `bs_allow_stat_bucket` (as it is implemented in the bucketspace and not in the namespace).
4. Add `stat_account_config_file` in `config_fs`.

### Issues: Fixed #8071
1. Currently, when updating access keys of an account using the noobaa-cli they are validate until the item is expired (in the current config 10 mminutes). With the suggested code change, updating the access keys results in changing the config file, and therefore the stat is changing and the item is no longer valid in the `account_cache`.

GAPS:
1. There are many functions that should move to accountspace and not bucketspace (and also the `account_cache` itself should be in `account_sdk`). Please see [comment1](https://github.com/noobaa/noobaa-core/pull/8585#discussion_r1890212811) [comment2](https://github.com/noobaa/noobaa-core/pull/8585#discussion_r1888280651).
2. We would probably want to do the same on `account_id_cache`.

Note: the idea for the solution was planned as part of issue #8391 

#### Automatic Test:
Please run: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_nc_with_a_couple_of_forks.js`

#### Manual Test:
1. Add the line: `dbg.log('SDSD same_stat', same_stat);` after declaring the variable `same_stat`, and restart the server.
##### Regular account (not anonymous)
2. Create an account with the CLI: `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path /Users/buckets/ --access_key <access-key> --secret_key <secret-key> --uid <uid> --gid <gid>`
Note: before creating the account need to give permission to the `new_buckets_path`: `chmod 777 /Users/buckets/`
4. Start the NSFS server with: `sudo node src/cmd/nsfs --debug 5`
[...]

PR #8527, opened 2024-11-13
NC | NSFS | Add `stat` to `bucket_namespace_cache`

shirady:
### Explain the changes
1. Create config option `NC_ENABLE_BUCKET_NS_CACHE_STAT_VALIDATION` to enable/disable this addition.
2. Add in `read_bucket_sdk_info` inside bucketspace FS property `stat`, and add a method `stat_bucket` that would only implemented in FS (as we `stat` cofig files, and in NB with DB we don't need this).
3. Edit the `_validate_bucket_namespace` so before checking the validation by time, in case it is bucketspace FS we will compare the `stat` and return `false` (invalidate) in case the config file was changed.
4. Changes in `nc_coretest` create the functions `start_nsfs_process` and `stop_nsfs_process` based on existing implementation to allow us to restart nsfs.
5. Rename the function `check_bucket_config` to `check_stat_bucket_storage_path` that it would be less confusing.
6. Edit the printings of "Could not log bucket operation" to have the details of where they were printed.
7. Adding a basic test of running with a couple of forks (based on [this comment](https://github.com/noobaa/noobaa-core/pull/8527#discussion_r1842148932)).
8. Change the default number of forks in the Ceph NSFS tests in the CI to 2 (see [comment](https://github.com/noobaa/noobaa-core/pull/8527#discussion_r1848506858)).
9. Rename file `test_bucketspace.js` to `test_nsfs_integration.js` and the nc_core test log files from `src/logfile` to `nsfs_integration_test_log`.
10. Small changes in `s3_bucket_logging.js` regarding bucket deletion: bucket existed and bucket that did not exist  (see [comment](https://github.com/noobaa/noobaa-core/pull/8527#issuecomment-2485344631)) - see #8540.

Note: the changes are based on the suggestions written as a [comment](https://github.com/noobaa/noobaa-core/issues/8391#issuecomment-2410609077).

### Issues: Fixed #8391 
1. Currently, bucket configuration is saved in a file, but when using the `bucket_namespace_cache` it expires after 1 minute, see https://github.com/noobaa/noobaa-core/blob/e1bf29e82128a9e0d359b262978d208ccff228e5/config.js#L637-L638 We want to invalidate a bucket config not only by the time criteria.

GAPs:
1. We would probably want to do the same on `account_cache` and `account_id_cache`, but currently it has lower priority than the `bucket_namespace_cache`.
[...]

Issue #8391, opened 2024-09-22
NC | NSFS | `bucket_namespace_cache` Is Not Updated Between Forks


shirady:
### Environment info
- NooBaa Version: 5.18
- Platform: NC

### Actual behavior
1. When using child process configuration (`ENDPOINT_FORKS` in the `config.json`) the `bucket_namespace_cache` is not shared between the forks. Currently, we need to wait for one minute before the item is cleaned from the cache.

### Expected behavior
1. The cache will be updated and we will not have to wait.

### Steps to reproduce
This issue was raised as part of issue #8368 
1. Change the configuration to have more than 1 fork, for example: `sudo vi /etc/noobaa.conf.d/config.json` and `{"ENDPOINT_FORKS": 2}`).
2. Change the bucket configuration (for example change the existing bucket policy from: list object versions of a versioned bucket of all principals on the bucket and its objects to: allow get-object of a specific key with principal of account2).
3. Use the configuration as it was changed (try to get-object of this key by account2).

more details about the steps can be found in [this comment](https://github.com/noobaa/noobaa-core/issues/8368#issuecomment-2357765607).
It is not guaranteed that you will reproduce it since it might be that all the changes would be from the same child process.

### More information - Screenshots / Logs / Other output
 Example of policies:
from policy1:
 ```json
{
  "Version": "2012-10-17",
  "Statement": [ 
    { 
     "Effect": "Allow", 
     "Principal": { "*" }, 
     "Action": [ "s3:ListBucketVersions" ], 
     "Resource": [ "arn:aws:s3:::<bucket-name>/*", "arn:aws:s3:::<bucket-name>" ] 
    }
  ]
}
```

to policy2:
```json
{
  "Version": "2012-10-17",
  "Statement": [ 
    { 
     "Effect": "Allow", 
     "Principal": { "AWS": [ "<account name-2" ] }, 
     "Action": [ "s3:GetObject", "s3:GetObjectVersion" ], 
     "Resource": [ "arn:aws:s3:::<bucket-name>/<key-name>" ] 
    }
  ]
}
```

shirady:
Hi,
Please notice that after adding a fix for this issue we need to go back to a couple of closed issues and make sure that they are fixed (otherwise reopen):
1. #8368 (related to `test_s3cever11` in GPFS in @hseipp repo).
[...]

PR #8642, opened 2025-01-01
NC | NSFS | Add `stat` to `account_id_cache`

shirady:
### Explain the changes
1. In `config.js` move the definition of `ACCOUNTS_ID_CACHE_EXPIRY` near the the configurations of `OBJECT_SDK_BUCKET_CACHE_EXPIRY_MS` and `OBJECT_SDK_ACCOUNT_CACHE_EXPIRY_MS`.
2. In `accountspace_fs.js`:
   - Remove the Number conversion as the value is number (was `Number(config.ACCOUNTS_ID_CACHE_EXPIRY)`).
   - Change the JSDoc of params so it will be easier to read.
   - Change the `load` inside function to be `get_identity_by_id_and_stat_file` so we will have the identity with the `stat` as property inside it.
   - Add the `validate` method, and add the functions `_validate_account_id` and `check_same_stat_account` that were partially copied from object_sdk.
3. In `config_fs` add the function `get_identity_by_id_and_stat_file` and `stat_account_config_file_by_identity ` that would try to stat the account config file (covering the identity, accounts_by_name - new path, accounts - old path).
4. In `object_sdk.js` add a new line between `anonymous_access_key` `bucket_namespace_cache` for easier readability, fix a typo in a comment.

Note: the changes are based on the suggesting written as a [comment](https://github.com/noobaa/noobaa-core/issues/8391#issuecomment-2410609077).

### Issues: 
1. It is a mentioned GAP in #8527 and #8585.

#### Automatic Test:
Please run: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_nc_with_a_couple_of_forks.js`

#### Manual Test:
2. Add the line: `dbg.log0('SDSD same_stat', same_stat);` before `if (!same_stat) { // config file of bucket was changed` and start the NSFS server with: `sudo node src/cmd/nsfs --debug 5`
5. Add bucket to the account using AWS CLI: `nc-user-1-s3 s3 mb s3://bucket-01` (`bucket-01` is the bucket name in this example)
6. Head the bucket: `nc-user-1-s3 s3api head-bucket --bucket bucket-01` (expect to see "SDSD same_stat true")
7. Edit the account config file, for example: `sudo node src/cmd/manage_nsfs account update --name <account-name> --fs_backend GPFS` (it was '' and we change it to GPFS).
8. Head the bucket again: `nc-user-1-s3 s3api head-bucket --bucket bucket-01` (expect to see "SDSD same_stat false").

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/sdk/accountspace_fs.js
+++ b/src/sdk/accountspace_fs.js
@@ -28,22 +28,61 @@ const dummy_service_name = 's3';
 
 const account_id_cache = new LRUCache({
     name: 'AccountIDCache',
-    // TODO: Decide on a time that we want to invalidate
-    expiry_ms: Number(config.ACCOUNTS_ID_CACHE_EXPIRY),
+    expiry_ms: config.ACCOUNTS_ID_CACHE_EXPIRY,
     make_key: ({ _id }) => _id,
     /**
      * Accounts are added to the cache based on id, Default value for show_secrets and decrypt_secret_key will be true, 
      * and show_secrets and decrypt_secret_key `false` only when we load cache from the health script, 
      * health script doesn't need to fetch or decrypt the secret.
-     * @param {{ _id: string, 
-     * show_secrets?: boolean,
-     * decrypt_secret_key?: boolean, 
-     * config_fs: import('./config_fs').ConfigFS }} params
+     * @param {{
+        * _id: string,
+        * show_secrets?: boolean,
+        * decrypt_secret_key?: boolean,
+        * config_fs: ConfigFS
+     * }} params
     */
-    load: async ({ _id, show_secrets = true, decrypt_secret_key = true, config_fs}) =>
-        config_fs.get_identity_by_id(_id, CONFIG_TYPES.ACCOUNT, { show_secrets: show_secrets, decrypt_secret_key: decrypt_secret_key }),
+    load: async ({ _id, show_secrets = true, decrypt_secret_key = true, config_fs }) =>
+        config_fs.get_identity_by_id_and_stat_file(
+            _id, CONFIG_TYPES.ACCOUNT, { show_secrets: show_secrets, decrypt_secret_key: decrypt_secret_key }),
+    validate: (params, data) => _validate_account_id(params, data),
 });
 
+/** _validate_account_id is an additional layer (to expiry_ms)
+ * to checks the stat of the config file
+ * @param {object} data
+ * @param {object} params
+ * @returns Promise<{boolean>}
+ */
+async function _validate_account_id(params, data) {
+    if (config.NC_ENABLE_ACCOUNT_ID_CACHE_STAT_VALIDATION) {
+        const same_stat = await check_same_stat_account(params._id, params.name, params.stat, data.config_fs);
+        if (!same_stat) { // config file of account was changed
+            return false;
+        }
+    }
+    return true;
+}
+
+    /**
+     * check_same_stat_account will return true the config file was not changed
+     * in case we had any issue (for example error during stat) the returned value will be undefined
+     * @param {string} id
+     * @param {string} account_name
+     * @param {nb.NativeFSStats} account_stat
+     * @param {ConfigFS} config_fs
+     * @returns Promise<{boolean|undefined>}
+     */
+    async function check_same_stat_account(id, account_name, account_stat, config_fs) {
+        if (!config_fs) return;
+        try {
+            const current_stat = await config_fs.stat_account_config_file_by_identity(id, account_name);
+            if (current_stat) {
+                return current_stat.ino === account_stat.ino && current_stat.mtimeNsBigint === account_stat.mtimeNsBigint;
+            }
+        } catch (err) {
+            dbg.warn('check_same_stat_account: current_stat got an error', err, 'ignoring...');
+        }
+    }
 
 /**
  * @param {Object} requested_account
--- a/src/sdk/config_fs.js
+++ b/src/sdk/config_fs.js
@@ -379,6 +379,29 @@ class ConfigFS {
     }
 
     /**
+     * get_identity_by_id_and_stat_file returns the full account/user data and stat the file:
+     * @param {string} id
+     * @param {string} [type]
+     * @param {{show_secrets?: boolean, decrypt_secret_key?: boolean, silent_if_missing?: boolean}} [options]
+     * @returns {Promise<Object>} 
+    */
+    async get_identity_by_id_and_stat_file(id, type, options = {}) {
+        let identity;
+        try {
+            identity = await this.get_identity_by_id(id, type, options);
+            if (!identity && options.silent_if_missing) {
+                return undefined; // we don't have the identity, so we can't add the property of stat to it
+            }
+            const stat = await this.stat_account_config_file_by_identity(id, identity.name);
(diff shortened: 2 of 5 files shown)


shirady:
Moving to draft since there is unexpected failure in tests "nsfs nc health -> health check".

src/sdk/accountspace_fs.js
```diff
@@ -28,22 +28,61 @@ const dummy_service_name = 's3';
...
+        * show_secrets?: boolean,
+        * decrypt_secret_key?: boolean, 
+        * config_fs: ConfigFS
+     * }} params
     */
     load: async ({ _id, show_secrets = true, decrypt_secret_key = true, config_fs}) =>
-        config_fs.get_identity_by_id(_id, CONFIG_TYPES.ACCOUNT, { show_secrets: show_secrets, decrypt_secret_key: decrypt_secret_key }),
+        config_fs.get_identity_by_id_and_stat_file(_id, CONFIG_TYPES.ACCOUNT,
+            { show_secrets: show_secrets, decrypt_secret_key: decrypt_secret_key }),
+    validate: (params, data) => _validate_account_id(params, data),
```
nadavMiz, line 47: looks like there is something weird with the indentation

src/sdk/config_fs.js
```diff
@@ -468,6 +496,46 @@ class ConfigFS {
...
+     * @returns {Promise<nb.NativeFSStats>}
+     */
+    async stat_account_config_file_by_identity(id, account_name) {
+        const identity_path = this.get_identity_path_by_id(id);
+        let stat;
+        try {
+            const stat_by_identity_path = await nb_native().fs.stat(this.fs_context, identity_path);
+            stat = stat_by_identity_path;
+        } catch (err_stat_by_identity_path) {
+            if (err_stat_by_identity_path.code === 'ENOENT') {
```
nadavMiz, line 512: this nested levels of try-catch are a bit confusing. maybe we can move those condition outside the try-catch, and return when a condition is met. what do you think?

src/sdk/config_fs.js
```diff
@@ -468,6 +496,46 @@ class ConfigFS {
...
+                        const account_name_old_path = this._get_old_account_path_by_name(account_name);
+                        try {
+                            const stat_by_account_name_old = await nb_native().fs.stat(this.fs_context, account_name_old_path);
+                            stat = stat_by_account_name_old;
+                        } catch (err_stat_by_account_name_old_path) {
+                            dbg.warn('get_identity_by_id_and_stat_file: could not stat by by account name (old accounts path)');
+                            // eslint-disable-next-line max-depth
+                            const error_to_throw = new Error(`Could not stat identity by id ${id} or by account name ${account_name}`);
+                            error_to_throw.code = 'ENOENT';
+                            throw error_to_throw;
```
nadavMiz, line 530: this seems to be the only place we actually throw. in all the previous try catch there are no other cases that needs to throw. say when the error isn't ENOENT

src/sdk/config_fs.js
```diff
@@ -378,6 +378,34 @@ class ConfigFS {
...
+    */
+    async get_identity_by_id_and_stat_file(id, type, options = {}) {
+        let identity;
+        try {
+            identity = await this.get_identity_by_id(id, type, options);
+            const stat = await this.stat_account_config_file_by_identity(id, identity.name);
+            identity.stat = stat;
+        } catch (err) {
+            dbg.error('get_identity_by_id_and_stat_file: got an error', err);
+            if (!identity && !options.silent_if_missing) {
```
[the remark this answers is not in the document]
shirady: I copied the logic of this exception from the previous function (`get_identity_by_id`) since I only wanted to add the `stat` property.

src/sdk/config_fs.js
```diff
@@ -468,6 +496,46 @@ class ConfigFS {
...
+     * stat_account_config_file_by_identity will return the stat output on account config file by id or by account name
+     * @param {string} id
+     * @param {string} account_name
+     * @returns {Promise<nb.NativeFSStats>}
+     */
+    async stat_account_config_file_by_identity(id, account_name) {
+        const identity_path = this.get_identity_path_by_id(id);
+        let stat;
+        try {
+            const stat_by_identity_path = await nb_native().fs.stat(this.fs_context, identity_path);
```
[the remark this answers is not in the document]
shirady: I have the `stat` variable that I would return, and it's value is depended on the config structure (identity, account by name, old account path). I can assign each one to this stat and return it, but I thought that by naming each one it will be easier to understand - if you think that it is confusing, I can change it.

src/sdk/config_fs.js
shirady: I didn't understand it, could you give an example or add your suggestion?

src/sdk/config_fs.js
shirady: It is the last way that we have to stat the file after we tried all the other ways, so we will throw the error.

src/sdk/config_fs.js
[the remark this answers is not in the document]
nadavMiz: right, but you call get_identity_by_id inside this function, so this code already runs

src/sdk/config_fs.js
nadavMiz: maybe you can do something like:
```
try {
    stat = await nb_native().fs.stat(this.fs_context, identity_path);
    return stat;
catch(err) { 
    if (err_stat_by_identity_path.code !== 'ENOENT') {
        throw err
    }
}
try {
     stat = await nb_native().fs.stat(this.fs_context, account_name_new_path);
    return stat;
} catch (err_stat_by_account_name_new_path
...
```
something like this
that way the conditions are separated from each other.

src/sdk/config_fs.js
nadavMiz: right, but you ignore all errors that are not ENOENT along the way. I don't think any of them should be raised. but still if you refer to ENOENT specifically might as well handle the others as well.

src/sdk/config_fs.js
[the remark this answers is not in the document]
shirady: I changed it, thanks.

src/sdk/config_fs.js
shirady: I changed it according to your suggestion, thanks.

src/sdk/config_fs.js
```diff
@@ -378,6 +378,30 @@ class ConfigFS {
...
+     * silent_if_missing is used only on get identity (if stat fails, it will throw an error)
+     * @param {string} id
+     * @param {string} [type]
+     * @param {{show_secrets?: boolean, decrypt_secret_key?: boolean, silent_if_missing?: boolean}} [options]
+     * @returns {Promise<Object>} 
+    */
+    async get_identity_by_id_and_stat_file(id, type, options = {}) {
+        let identity;
+        try {
+            identity = await this.get_identity_by_id(id, type, options);
```
nadavMiz, line 395: can you verify identity can't be undefined. and if it can please handle this case

shirady: I added, thanks.

nadavMiz (approved):
LGTM

PR #8657, opened 2025-01-06
bucket notifications - get notif should clone result before altering it (dfsbugs 1286)

alphaprinz:
### Explain the changes
1. s3_get_bucket_notification adapts the names of fields in result to match s3 spec.
This should be done on a clone. 

1. https://issues.redhat.com/browse/DFBUGS-1286

### Testing Instructions:
1. Perform s3_get_bucket_notification on a bucket.
2. Perform event (eg s3 cp) that has notification on the bucket (on the same endpoint as step 1)

- [ ] Doc added/updated
- [ ] Tests added


aspandey (approved):
Looks good to me.

Issue #8548, opened 2024-11-20
NC | NSFS | Organize Modules/Dependecies


shirady:
### Environment info
- NooBaa Version: current, master
- Platform: NC

This is the place for organizing the dependencies:
1.  Move the file `src/endpoint/s3/s3_bucket_policy_utils.js` to `src/util` (it is used by the noobaa-cli in `src/manage_nsfs/manage_nsfs_validations.js`).
2. Move the content of `validate_username` from `src/endpoint/iam/iam_utils.js` (it is used by noobaa-cli in `src/manage_nsfs/manage_nsfs_validations.js`).
3. Use the function `generate_id` from file `src/nc/nc_utils.js` in places that used directly `mongo_utils.mongoObjectId()` (for example: `src/cmd/manage_nsfs.js`). I think we can copy the content and add a reference and credit instead of any use of the module `src/util/mongo_utils.js` in `src/nc/nc_utils.js`.

PR #8661, opened 2025-01-07
NC | NSFS | IAM | Tech Debts (IAM Integration Tests, Username Validation Move module and Allow IAM User to Create Bucket)

shirady:
### Explain the changes
IAM Tech Debts (from an agreed list to wrap up the feature for now) including: 
1. IAM Integration Tests: add the file `test_nc_iam_basic_integration.js` and make the needed changes in the files `nc_coretest.js` (add the IAM port), `nc_index` (add the new test in the CI) and `test_utils.js` (add the IAM client - like we have S3 client) - the IAM integration tests the APIs of IAM that we support today.
Please Notice that it contains only the happy path, as the goal was to ensure nothing was broken between the IAM request to the server and our response (internal validation was already implemented in unit tests).
2. Username Validation Move the Module: we have 2 flows noobaa-cli and API (S3, IAM), and don't want to import modules between the flows and only from an above level. Therefore, I moved the function `validate_username` from the `iam_utils` to `validation_utils`, since it used other functions I also had to move them and move the testing file.
3. Allow IAM Users to Create Bucket - we temporarily didn't allow IAM users to create buckets (see [comment](https://github.com/noobaa/noobaa-core/pull/8175#issuecomment-2254463451)).

### Issues:
1. Fix partially issue #8548.

This PR was part of [MCGI-282](https://issues.redhat.com/browse/MCGI-282) IAM Tech Debts and the GAPS mentioned in the comments here were added in the Jira ticket (for example: adding interrogation tests for IAM and S3 together).

### Testing Instructions:
**1)  IAM Integration Tests:**
#### Automatic Tests
Please run the test: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_nc_iam_basic_integration.js`
If you want to look at the logs in noobaa, you can: `cat nsfs_integration_test_log.txt`
#### Manual Tests
In this PR I didn’t run manual tests, but if you want you can use [this guide](https://github.com/noobaa/noobaa-core/blob/master/docs/dev_guide/nc_nsfs_iam_developer_doc.md)

**2) Username Validation Move the Module:**
#### Automatic Tests
Please run the tests: 
  - `npx jest test_iam_utils.test.js` (`iam_utils` is the previous file that the function was taken from)
  - `npx jest test_nc_utils.test.js` (`validation_utils` is the current file, where I moved the functions to)
  - `sudo npx jest test_nc_nsfs_account_cli.test.js` (contains the tests that were written in the past, search the test titles that include “invalid name”)

**3) Allow IAM Users to Create Bucket**
#### Automatic Tests
Please run the test: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace_fs.js`

#### Manual Tests
3. Start the NSFS server with: `sudo node src/cmd/nsfs --debug 5 --https_port_iam 7005`
4. Create the alias for S3 service:`alias nc-user-1-s3=‘AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443’`.
5. Check the connection to the endpoint and try to list the buckets (should be empty): `nc-user-1-s3 s3 ls; echo $?`
6. Create the alias for the IAM service: `nc-user-1-iam='AWS_ACCESS_KEY_ID=<access-key-account> AWS_SECRET_ACCESS_KEY=<secret-key-account> aws --no-verify-ssl --endpoint-url https://localhost:7005/'`
7. Check the connection to the endoint and try to list the users (should be empty): `nc-user-1-iam iam list-users`
8. Create a IAM user with access key: `nc-user-1-iam iam create-user --user-name Bob` and 
`nc-user-1-iam iam create-access-key --user-name Bob`
9. Create the alias for the IAM user to S3 service: `alias nc-user-1b-s3='AWS_ACCESS_KEY_ID=<access-key-user> AWS_SECRET_ACCESS_KEY=<secret-key-user> aws --no-verify-ssl --endpoint-url https://localhost:6443/'`
10. Check the the connection to the endpoint and try to list the buckets (should be empty): `nc-user-1b-s3 s3 ls; echo $?`
11. Create a bucket by the IAM user: `nc-user-1b-s3 s3 mb s3://iam-bucket`
12. Check that the `owner_account` property is the account’s ID (the owner of the user): `sudo cat /etc/noobaa.conf.d/buckets/iam-bucket.json | jq .`
13. Delete the bucket: `nc-user-1b-s3 s3 rb s3://iam-bucket`

- [ ] Doc added/updated
- [X] Tests added


liranmauda:
@shirady could you create a new PR that only adds the @aws-sdk/client-iam

shirady:
> @shirady could you create a new PR that only adds the @aws-sdk/client-iam
> 
> and if we are using it only for tests, make it a dev dependency

Yes, thanks.

src/endpoint/iam/iam_utils.js
```diff
@@ -79,67 +80,6 @@ function parse_max_items(input_max_items) {
...
 }
 
-
-/**
- * _type_check_input checks that the input is the same as needed
- * @param {string} input_type
- * @param {string | number} input_value
- * @param {string} parameter_name
- */
-function _type_check_input(input_type, input_value, parameter_name) {
```
romayalon, line 89: i'm not quite sure that moving these functions to nc_utils.js make sense, because we are going to support IAM on containerized flow as well

src/nc/nc_utils.js
```diff
@@ -24,7 +28,136 @@ function check_root_account_owns_user(root_account, account) {
...
+ * validate_username will validate:
+ * 1. type
+ * 2. length
+ * 3. regex (from AWS docs)
+ * 4. additional internal restrictions
+ * @param {string} input_username
+ * @param {string} parameter_name
+ * @param {boolean} [use_iam_error]
+ */
+function validate_username(input_username, parameter_name = iam_constants.IAM_PARAMETER_NAME.USERNAME, use_iam_error = true) {
```
romayalon, line 114: if you remember our discussion about Internal vs external errors, use_iam_error is the probably the exact reason why we used RPC internal errors and then converted them to external errors. why not consider using RPC errors?

src/sdk/bucketspace_fs.js
```diff
@@ -345,7 +340,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             _id: mongo_utils.mongoObjectId(),
             name,
             tag: js_utils.default_value(tag, undefined),
-            owner_account: account._id,
+            owner_account: account.owner ? account.owner : account._id,
```
romayalon, line 343: can you add a comment why the owner of the bucket is the account owner and not the user? where is this behavior documented?

src/test/unit_tests/nc_coretest.js
```diff
@@ -267,6 +274,13 @@ function get_https_address() {
     return https_address;
 }
 
+/**
+ * get_iam_https_address return nc coretest https_address_iam variable
```
romayalon, line 278: missing return

src/test/unit_tests/test_nc_iam_basic_integration.js
```diff
@@ -0,0 +1,234 @@
...
+
+    mocha.describe('IAM User API', async function() {
+        const username = 'Asahi';
+        const new_username = 'Botan';
+
+        mocha.it('list users - should be empty', async function() {
+            const input = {};
+            const command = new ListUsersCommand(input);
+            const response = await iam_account.send(command);
+            assert.equal(response.$metadata.httpStatusCode, 200);
```
romayalon, line 64: expect the status code, can you check the actual response?

src/test/unit_tests/test_nc_iam_basic_integration.js
```diff
@@ -0,0 +1,234 @@
...
+        });
+
+        mocha.it('create a user', async function() {
+            const input = {
+                UserName: username
+            };
+            const command = new CreateUserCommand(input);
+            const response = await iam_account.send(command);
+            assert.equal(response.$metadata.httpStatusCode, 200);
+            assert.equal(response.User.UserName, username);
```
romayalon, line 74: any other properties of the user you should check?

src/test/unit_tests/test_nc_iam_basic_integration.js
```diff
@@ -0,0 +1,234 @@
...
+        });
+
+        mocha.it('list users - should have 1 user in the list', async function() {
+            const input = {};
+            const command = new ListUsersCommand(input);
+            const response = await iam_account.send(command);
+            assert.equal(response.$metadata.httpStatusCode, 200);
+            assert.equal(response.Users.length, 1);
+            assert.equal(response.Users[0].UserName, username);
```
romayalon, line 83: I know we did that a lot in our tests in the past, but for future tests it's better to not have a dependancy between tests

src/test/unit_tests/test_nc_iam_basic_integration.js
```diff
@@ -0,0 +1,234 @@
...
+            assert.equal(response.Users[0].UserName, new_username);
+        });
+
+        mocha.it('delete a user', async function() {
+            const input = {
+                UserName: new_username // delete a user after its username was updated
+            };
+            const command = new DeleteUserCommand(input);
+            const response = await iam_account.send(command);
+            assert.equal(response.$metadata.httpStatusCode, 200);
```
romayalon, line 121: worth creating a function with assert.equal(response.$metadata.httpStatusCode, 200); instead of repeating it

src/test/unit_tests/test_bucketspace_fs.js
```diff
@@ -431,16 +431,13 @@ mocha.describe('bucketspace_fs', function() {
...
-                await bucketspace_fs.create_bucket(param, dummy_object_sdk_for_iam_account);
-                assert.fail('should have failed with UNAUTHORIZED bucket creation');
-            } catch (err) {
-                assert.ok(err.rpc_code === 'UNAUTHORIZED');
-            }
+        mocha.it('create bucket by iam account', async function() {
+            const param = { name: test_bucket_iam_account };
+            const dummy_object_sdk_for_iam_account = make_dummy_object_sdk_for_account(dummy_object_sdk, account_iam_user1);
+            await bucketspace_fs.create_bucket(param, dummy_object_sdk_for_iam_account);
+            const bucket_config_path = get_config_file_path(CONFIG_SUBDIRS.BUCKETS, param.name);
```
romayalon, line 438: should you also check the owner identity?

src/test/unit_tests/test_nc_iam_basic_integration.js
```diff
@@ -0,0 +1,234 @@
...
+        });
+
+        mocha.it('delete access keys', async function() {
+            const input = {
+                UserName: username2,
+                AccessKeyId: access_key_id
+            };
+            const command = new DeleteAccessKeyCommand(input);
+            const response = await iam_account.send(command);
+            assert.equal(response.$metadata.httpStatusCode, 200);
```
romayalon, line 231: besides the response was successful, should you check that they really were deleted and that another usage with these keys will throw access denied?

src/test/unit_tests/test_nc_iam_basic_integration.js
```diff
@@ -0,0 +1,234 @@
...
+const CORETEST_ENDPOINT_IAM = coretest.get_iam_https_address();
+
+const config_root = path.join(TMP_PATH, 'test_nc_iam');
+// on NC - new_buckets_path is full absolute path
+// on Containerized - new_buckets_path is the directory
+const new_bucket_path_param = get_new_buckets_path_by_test_env(config_root, '/');
+
+let iam_account;
+
+mocha.describe('IAM basic integration tests - happy path', async function() {
```
romayalon, line 32: do we need also integration tests between both IAM and S3 services?
for example, I would expect a created user can do I/O etc

src/test/unit_tests/test_nc_iam_basic_integration.js
shirady: In the `test_accountspace_fs.test.js` I check that I can't find the path by the deleted access key.
I don't have full usage as the integration is IAM to the server and I don't have a full IAM_S3 integration as you asked before.

src/test/unit_tests/test_nc_iam_basic_integration.js
romayalon: okay, so please add a sub-task

src/test/unit_tests/test_nc_iam_basic_integration.js
romayalon: I don't see a point in a test that is sending list accounts and just checking that the server responded 200 without checking what's the actual response

src/nc/nc_utils.js
romayalon: I don't know if it's small or not, I think using `use_iam_error = true` is not designed well

src/endpoint/iam/iam_utils.js
romayalon: I think that iam_utils is the correct place for these functions, why can't the CLI use it from here?

src/test/unit_tests/test_nc_iam_basic_integration.js
shirady: We have a basic check `assert.equal(response.Users.length, 0);` so there is always as I understand a basic check on the response.

romayalon (approved):

PR #8664, opened 2025-01-08
Added a fix for handling invalid_schema_error returning with invalid status value for bucket lifecycle configuration rule

### LINKED ISSUE #8553: NSFS | S3 | Lifecycle: Invalid rule status returns http error 500
### Environment info
- NooBaa Version: noobaa-core-5.17.1-20241104.el9.x86_64
- Platform: RHEL 9.2 x86

### Actual behavior
1. When posting a lifecycle rule with Status=enabled instead of Status=Enabled, Noobaa returns status 500. Here is an example of the policy:

{
    "Rules": [
        {
            "ID": "Expiration-simple",
            "Status": "enabled",
            "Filter": {
               "Prefix": "small"
            },
            "Expiration": { "Days": 20 }
        }
     ]
}

Here is the status code from the aws cli debug response 
2024-11-21 17:25:40,746 - MainThread - botocore.parsers - DEBUG - Response body:
b'<?xml version="1.0" encoding="UTF-8"?><Error><Code>InternalError</Code><Message>We encountered an internal error. Please try again.</Message><Resource>/files?lifecycle</Resource><RequestId>m3rixim4-hnhpv-1102</RequestId></Error>'
2024-11-21 17:25:40,748 - MainThread - botocore.hooks - DEBUG - Event needs-retry.s3.PutBucketLifecycleConfiguration: calling handler <botocore.retryhandler.RetryHandler object at 0x7fd7d7f07760>
2024-11-21 17:25:40,748 - MainThread - botocore.retryhandler - DEBUG - retry needed: retryable HTTP status code received: 500

### Expected behavior
1. Invalid status should return 400 (bad request) instead of 500 (internal server error)

### Steps to reproduce
1. Use the lifecycle rule above.

aayushchouhan09:
### Explain the changes
1. This fix ensures that when an INVALID_SCHEMA_PARAMS error occurs due to an incorrect Status value (other than "Enabled" or "Disabled") in a bucket lifecycle configuration rule, the appropriate AWS S3 error (MalformedXML) is returned.

Below is the aws output:
Screenshot from 2025-01-08 19-12-26

1. Fixed: #8553 

### Testing Instructions:
1. create s3 backingstore>bucketclass>obc.
2. Apply below policy on the bucket created with obc:
`$ AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws s3api --no-verify-ssl --endpoint-url <endpoint-url> put-bucket-lifecycle-configuration --bucket <bucket-name> --lifecycle-configuration '{
  "Rules": [
    {
      "ID": "rule_id",
      "Status": "enabled",
      "Filter": {
        "Prefix": "test/"
      },
      "Expiration": {
        "Date": "2020-01-02T00:00:00.000Z"
      }
    }
  ]
}'
`
3. Check for the similar error as AWS S3:
`An error occurred (MalformedXML) when calling the PutBucketLifecycleConfiguration operation: The XML you provided was not well-formed or did not validate against our published schema`

- [X] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
+++ b/src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
@@ -108,9 +108,10 @@ async function put_bucket_lifecycle(req) {
         }
         id_set.add(current_rule.id);
 
-        if (rule.Status?.length !== 1) {
-            dbg.error('Rule should have status', rule);
-            throw new S3Error(S3Error.InvalidArgument);
+        if (!rule.Status || rule.Status.length !== 1 ||
+            (rule.Status[0] !== s3_const.LIFECYCLE_STATUS.STAT_ENABLED && rule.Status[0] !== s3_const.LIFECYCLE_STATUS.STAT_DISABLED)) {
+            dbg.error(`Rule should have a status value of "${s3_const.LIFECYCLE_STATUS.STAT_ENABLED}" or "${s3_const.LIFECYCLE_STATUS.STAT_DISABLED}".`, rule);
+            throw new S3Error(S3Error.MalformedXML);
         }
         current_rule.status = rule.Status[0];
 
--- a/src/endpoint/s3/s3_constants.js
+++ b/src/endpoint/s3/s3_constants.js
@@ -10,3 +10,7 @@ const s3_const = exports;
 ///////////////
 
 s3_const.MAX_RULE_ID_LENGTH = 255;
+s3_const.LIFECYCLE_STATUS = Object.freeze({
+    STAT_ENABLED: 'Enabled',
+    STAT_DISABLED: 'Disabled'
+});
--- a/src/endpoint/s3/s3_errors.js
+++ b/src/endpoint/s3/s3_errors.js
@@ -271,7 +271,8 @@ S3Error.MalformedPOSTRequest = Object.freeze({
 });
 S3Error.MalformedXML = Object.freeze({
     code: 'MalformedXML',
-    message: 'This happens when the user sends malformed xml (xml that doesn\'t conform to the published xsd) for the configuration. The error message is, "The XML you provided was not well-formed or did not validate against our published schema."',
+    // This happens when the user sends malformed xml (xml that doesn't conform to the published xsd) for the configuration.
+    message: 'The XML you provided was not well-formed or did not validate against our published schema.',
     http_code: 400,
 });
 S3Error.InvalidTag = Object.freeze({
--- a/src/test/lifecycle/common.js
+++ b/src/test/lifecycle/common.js
@@ -569,3 +569,17 @@ exports.test_rule_duplicate_id = async function(Bucket, Key, s3) {
         assert(error.code === 'InvalidArgument', 'Expected InvalidArgument: duplicate ID found in the rules');
     }
 };
+
+exports.test_rule_status_value = async function(Bucket, Key, s3) {
+    const putLifecycleParams = id_lifecycle_configuration(Bucket, Key);
+
+    // set the status value to an invalid value - other than 'Enabled' and 'Disabled'
+    putLifecycleParams.LifecycleConfiguration.Rules[0].Status = 'enabled';
+
+    try {
+        await s3.putBucketLifecycleConfiguration(putLifecycleParams);
+        assert.fail('Expected MalformedXML error due to wrong status value, but received a different response');
+    } catch (error) {
+        assert(error.code === 'MalformedXML', `Expected MalformedXML error: due to invalid status value`);
+    }
+};
--- a/src/test/system_tests/test_lifecycle.js
+++ b/src/test/system_tests/test_lifecycle.js
@@ -56,6 +56,7 @@ async function main() {
     await commonTests.test_and_prefix_size(Bucket, Key, s3);
     await commonTests.test_rule_id_length(Bucket, Key, s3);
     await commonTests.test_rule_duplicate_id(Bucket, Key, s3);
+    await commonTests.test_rule_status_value(Bucket, Key, s3);
 
     const getObjectParams = {
         Bucket,


src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -171,10 +171,19 @@ async function put_bucket_lifecycle(req) {
...
+            rules: lifecycle_rules
+        });
+    } catch (error) {
+        let err = error;
+        if (error.rpc_code === "INVALID_SCHEMA" || error.rpc_code === "INVALID_SCHEMA_PARAMS") {
+            err = new S3Error(S3Error.MalformedXML);
+            err.message = "The XML you provided was not well-formed or did not validate against our published schema";
+        }
+        throw err;
+    }
```
shirady, line 186: I prefer "Fail First" concept ([reference](https://stackoverflow.com/questions/2807241/what-does-the-expression-fail-early-mean-and-when-would-you-want-to-do-so)).
Did you try to go over the rules and validate their status value?
In the [AWS S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/API/API_LifecycleRule.html), I can see that we have 2 options only ("Valid Values: Enabled | Disabled").

src/test/lifecycle/common.js
```diff
@@ -569,3 +569,17 @@ exports.test_rule_duplicate_id = async function(Bucket, Key, s3) {
         assert(error.code === 'InvalidArgument', 'Expected InvalidArgument: duplicate ID found in the rules');
     }
 };
+
+exports.test_rule_status_value = async function(Bucket, Key, s3) {
+    const putLifecycleParams = id_lifecycle_configuration(Bucket, Key);
+
+    // set the status value to an invalid value - other then 'Enabled' and 'Disabled'
```
shirady, line 576: typo
```suggestion
    // set the status value to an invalid value - other than 'Enabled' and 'Disabled'
```

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
achouhan09: Yes, you are correct. I am failing it at the time of validation. But I think we can add still return a different error from here just like [this](https://github.com/noobaa/noobaa-core/blob/0f3d341bdb36c6a22db401da7d43696e2a22647f/src/endpoint/s3/ops/s3_put_bucket_policy.js#L17C5-L27C2)

shirady: I'm suggesting to:
1. Validate the status in the rules before it passes inside.
2. Fix the message of `MalformedXML` to be the actual message and move the text before as a comment (so we will not need to rewrite it every time).
3. Change the translation you copied as a function that will be reused
WDYT?

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -111,6 +111,11 @@ async function put_bucket_lifecycle(req) {
         if (rule.Status?.length !== 1) {
             dbg.error('Rule should have status', rule);
             throw new S3Error(S3Error.InvalidArgument);
+        } else if (rule.Status?.length === 1) {
+            if (rule.Status[0] !== "Enabled" && rule.Status[0] !== "Disabled") {
```
shirady, line 115: Why do you access it as an array? I thought we were expecting a string...
Please move "Enabled" and "Disabled" to an enum (you can use the s3_constant file).

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
achouhan09: Updated, Thanks

src/endpoint/s3/s3_constants.js
```diff
@@ -10,3 +10,7 @@ const s3_const = exports;
 ///////////////
 
 s3_const.MAX_RULE_ID_LENGTH = 255;
+s3_const.STATUS_VAL = Object.freeze({
```
shirady, line 13: I'm suggesting to rename it in case we will have more enum status in the future:
```suggestion
s3_const.LIFECYCLE_STATUS = Object.freeze({
```

src/endpoint/s3/s3_errors.js
```diff
@@ -266,7 +266,7 @@ S3Error.MalformedPOSTRequest = Object.freeze({
 });
 S3Error.MalformedXML = Object.freeze({
     code: 'MalformedXML',
-    message: 'This happens when the user sends malformed xml (xml that doesn\'t conform to the published xsd) for the configuration. The error message is, "The XML you provided was not well-formed or did not validate against our published schema."',
+    message: 'The XML you provided was not well-formed or did not validate against our published schema.',
```
shirady, line 269: You can save what was deleted as a comment (in case it helped someone).
```suggestion
   // This happens when the user sends malformed XML (XML that doesn't conform to the published xsd) for the configuration
    message: 'The XML you provided was not well-formed or did not validate against our published schema.',
```

src/endpoint/s3/s3_utils.js
```diff
@@ -785,6 +785,15 @@ function key_marker_to_cont_tok(key_marker, objects_arr, is_truncated) {
     return Buffer.from(j).toString('base64');
 }
 
+function invalid_schema_to_aws_error(error) {
```
shirady, line 788: Please add JSDoc to the new function with a short explanation of why/when to use it.

src/endpoint/s3/s3_utils.js
```diff
@@ -785,6 +785,15 @@ function key_marker_to_cont_tok(key_marker, objects_arr, is_truncated) {
...
+    if (["INVALID_SCHEMA", "INVALID_SCHEMA_PARAMS"].includes(error.rpc_code)) {
+        const err = new S3Error(S3Error.MalformedPolicy);
+        err.message = "Policy was not well-formed or did not validate against the published schema";
```
shirady, line 791: Why do you need to edit the message? you already edited the message property in the error

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -111,6 +112,11 @@ async function put_bucket_lifecycle(req) {
         if (rule.Status?.length !== 1) {
             dbg.error('Rule should have status', rule);
             throw new S3Error(S3Error.InvalidArgument);
+        } else if (rule.Status?.length === 1) {
```
shirady, line 115: Something that is not clear from this logic is what if I have a rule without a Status, where will it fail?

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -111,6 +112,11 @@ async function put_bucket_lifecycle(req) {
...
+            if (!s3_const.STATUS_VAL.ENUM.includes(rule.Status[0])) {
+                dbg.error('Rule should not have status value other than "Enabled" and "Disabled" ', rule);
```
shirady, line 117: You can reuse the enum in the valid values printings.

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
shirady: You can check what you get with log printings, you can add `typeof <variable-name>` to check it.

achouhan09: It is of `object` type.

src/endpoint/s3/s3_constants.js
```diff
@@ -10,3 +10,7 @@ const s3_const = exports;
...
+    ENUM: Object.freeze(['Enabled', 'Disabled']),
+    TYPE: 'string'
+});
```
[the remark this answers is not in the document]
achouhan09: Updated, Thanks.

src/endpoint/s3/s3_constants.js
achouhan09: Make sense, Done.

src/endpoint/s3/s3_utils.js
achouhan09: Added, Thanks

src/endpoint/s3/s3_utils.js
achouhan09: Updated, Thanks

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
achouhan09: Added a condition to check for `Status` value `null` or `undefined`. Thanks

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
achouhan09: Done, Thanks

src/endpoint/s3/s3_utils.js
```diff
@@ -785,6 +785,20 @@ function key_marker_to_cont_tok(key_marker, objects_arr, is_truncated) {
...
 }
 
+/**
+ * invalid_schema_to_aws_error handles specific AWS error responses related to invalid schema or parameters
+ * If the error's `rpc_code` is either `INVALID_SCHEMA` or `INVALID_SCHEMA_PARAMS`,
+ * the function throws a `MalformedPolicy` error. Otherwise, it throws the original error.
+ *
+ * @param {Object} error - The error object to handle
+ */
+function invalid_schema_to_aws_error(error) {
```
shirady, line 795: Just to be on the safe side, you can also throw the error if (someone used this function improperly) `error.rpc_code` is `undefined`.

src/endpoint/s3/s3_errors.js
```diff
@@ -512,7 +513,7 @@ S3Error.XAmzContentSHA256Mismatch = Object.freeze({
 });
 S3Error.MalformedPolicy = Object.freeze({
     code: 'MalformedPolicy',
-    message: 'Invalid principal in policy',
+    message: 'Policy was not well-formed or did not validate against the published schema.',
```
shirady, line 516: Why did you change this?
In the attached image, I can see that the message is from the error `MalformedXML` (not `MalformedXML`).

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -108,11 +109,16 @@ async function put_bucket_lifecycle(req) {
         }
         id_set.add(current_rule.id);
 
-        if (rule.Status?.length !== 1) {
+        if (!rule.Status || rule.Status.length !== 1) {
             dbg.error('Rule should have status', rule);
             throw new S3Error(S3Error.InvalidArgument);
```
shirady, line 114: Since you touch it, I would check that it is `InvalidArgument` (or open a gap to validate it).

src/endpoint/s3/s3_errors.js
achouhan09: Previously, this customized msg was added for `MalformedPolicy` error we are throwing in case of `invalid_schema_params` error, for which we have now created a function `invalid_schema_to_aws_error`. So, I have update the common error msg here. wdyt, do we need to update here?

src/endpoint/s3/s3_utils.js
achouhan09: Sure, Thanks

src/endpoint/s3/s3_errors.js
```diff
@@ -266,7 +266,8 @@ S3Error.MalformedPOSTRequest = Object.freeze({
 });
 S3Error.MalformedXML = Object.freeze({
     code: 'MalformedXML',
-    message: 'This happens when the user sends malformed xml (xml that doesn\'t conform to the published xsd) for the configuration. The error message is, "The XML you provided was not well-formed or did not validate against our published schema."',
+    // This happens when the user sends malformed xml (xml that doesn\'t conform to the published xsd) for the configuration.
```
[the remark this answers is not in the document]
achouhan09: Sure, Thanks

src/endpoint/s3/s3_errors.js
shirady: You need to see what AWS does and be compatible with it.

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
shirady: The first error is from the AWS client as I understand it (not from the server).
I agree with your thought.
Anyway, in the current code we can see it is `InvalidArgument` and in the screenshot it is `MalformedXML`.

src/endpoint/s3/s3_errors.js
achouhan09: as per aws documentation [here](https://docs.aws.amazon.com/AmazonS3/latest/API/ErrorResponses.html) , we can update the error msg to `"Your policy contains a principal that is not valid."`

shirady: I think it's a description, not the actual message.

achouhan09: Screenshot from 2025-01-15 21-45-03

yes you are correct, removed these changes. Thanks

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
achouhan09: I think aws cli itself preventing the api call if the policy doesn't contains the required params here.

src/test/lifecycle/common.js
```diff
@@ -569,3 +569,17 @@ exports.test_rule_duplicate_id = async function(Bucket, Key, s3) {
         assert(error.code === 'InvalidArgument', 'Expected InvalidArgument: duplicate ID found in the rules');
     }
 };
+
+exports.test_rule_status_value = async function(Bucket, Key, s3) {
+    const putLifecycleParams = id_lifecycle_configuration(Bucket, Key);
+
+    // set the status value to an invalid value - other than 'Enabled' and 'Disabled'
+    putLifecycleParams.LifecycleConfiguration.Rules[0].Status = 'enabled';
```
shirady, line 577: Since you checked (and fixed it as well) you can add a test with empty status (`""`).

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -171,10 +173,14 @@ async function put_bucket_lifecycle(req) {
...
-        name: req.params.bucket,
-        rules: lifecycle_rules
-    });
+    try {
+        await req.object_sdk.set_bucket_lifecycle_configuration_rules({
+            name: req.params.bucket,
+            rules: lifecycle_rules
+        });
+    } catch (error) {
+        s3_utils.invalid_schema_to_aws_error(error);
```
shirady, line 182: I didn't understand what is the example of getting this error in AWS? (as we are trying to be compatible to them), If you can reference the screenshot or attach it here.

src/test/lifecycle/common.js
achouhan09: I think it is not required, as the test is to check for the values "Enabled" and "Disabled" so it would fail for all the other values(including empty string ""). Also we are handling it with a single condition in `s3_put_bucket_lifecycle.js` so that code is covered with this test.

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
achouhan09: Its an extra check where we get `invalid_schema_params` or `invalid_schema` error for any edge case, we are returning an aws error(MalformedPolicy). Not sure, if we need it or not but it was there already in `put_bucket_policy` so I have added it here in `put_bucket_lifecycle_configuration` as well.

shirady: I'm not sure you need to add it as you saw it in `put_bucket_policy` and not in `put_bucket_lifecycle_configuration `.

achouhan09: I will remove it from here then and will add it if needed in future. Thanks

shirady (approved):
LGTM

Issue #8649, opened 2025-01-06
put-bucket notification should check the connection before saving the information to the bucket


ramya-c3:
### Environment info
- NooBaa Version 5.18.0-20250101
- Platform: standalone

### Actual behavior
1. 
# cat connection1.json 
{
        "agent_request_object": {"host": "ramyamc-24.vmlocal", "port": 9001, "timeout": 100},
        "request_options_object": {"auth": "root:Cluster4scale","path": "/webhook"},
        "notification_protocol": "http",
        "name": "notify_event1"
}
 
#cat notification.json
 
{
  "TopicConfigurations": [
    {
      "Id": "notify_event",
      "TopicArn": "connection10.json",
      "Events": [
        "s3:ObjectCreated:*"
      ]
    }
}

I have placed wrong json 

now i try to 

# s3cp1 s3api put-bucket-notification-configuration --bucket bucket123  --notification-configuration file://notification.
# s3cp1 s3api get-bucket-notification-configuration --bucket bucket123
{
    "TopicConfigurations": [
        {
            "Id": "notify_event",
            "TopicArn": "connection10.json",
            "Events": [
                "s3:ObjectRemoved:*"
            ]
        }
    ]
}

it fails when i try to do noobaa-cli notification

### Expected behavior
1. it should fail in this call
s3cp1 s3api put-bucket-notification-configuration --bucket bucket123  --notification-configuration file://notification.

### Steps to reproduce
1.

alphaprinz:
Closed with mentioned PR.

PR #8667, opened 2025-01-09
bucket notifications - validate notifications on change (gh issue 8649)

alphaprinz:
### Explain the changes
1. As per aws spec, when bucket's notification change, all of the bucket's notification configuration are validated.
If one of them fails, the user's request is failed.

1. https://github.com/noobaa/noobaa-core/issues/8649

### Testing Instructions:
1. Create a new bucket with a non-existing connect file or with a wrong connect info.
2. Bucket creation should fail.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -203,7 +203,7 @@ async function get_bucket_status(data) {
  * @param {Object} data
  * @returns { Promise<{ code: typeof ManageCLIResponse.BucketUpdated, detail: Object }>} 
  */
-async function update_bucket(data, user_input) {
+async function update_bucket(data) {
     const cur_name = data.name;
     const new_name = data.new_name;
     const name_update = is_name_update(data);
@@ -212,13 +212,6 @@ async function update_bucket(data, user_input) {
 
     let parsed_bucket_data;
 
-    if (user_input.notifications) {
-        //notifications are tested before they can be updated
-        const test_notif_err = await notifications_util.test_notifications(data, config_fs.connect_dir_path);
-        if (test_notif_err) {
-            throw_cli_error(ManageCLIError.InvalidArgument, "Failed to update notifications", test_notif_err);
-        }
-    }
     if (name_update) {
         parsed_bucket_data = await config_fs.create_bucket_config_file({ ...data, name: new_name });
         await config_fs.delete_bucket_config_file(cur_name);
@@ -279,6 +272,7 @@ async function delete_bucket(data, force) {
 async function bucket_management(action, user_input) {
     const data = action === ACTIONS.LIST ? undefined : await fetch_bucket_data(action, user_input);
     await manage_nsfs_validations.validate_bucket_args(config_fs, data, action);
+    await manage_nsfs_validations.validate_bucket_notifications(config_fs, user_input.notifications);
 
     let response = {};
     if (action === ACTIONS.ADD) {
@@ -286,7 +280,7 @@ async function bucket_management(action, user_input) {
     } else if (action === ACTIONS.STATUS) {
         response = await get_bucket_status(data);
     } else if (action === ACTIONS.UPDATE) {
-        response = await update_bucket(data, user_input);
+        response = await update_bucket(data);
     } else if (action === ACTIONS.DELETE) {
         const force = get_boolean_or_string_value(user_input.force);
         response = await delete_bucket(data, force);
--- a/src/endpoint/s3/ops/s3_put_bucket_notification.js
+++ b/src/endpoint/s3/ops/s3_put_bucket_notification.js
@@ -1,6 +1,9 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
+const S3Error = require('../s3_errors').S3Error;
+const notif_util = require('../../../util/notifications_util');
+
 /**
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTnotification.html
  */
@@ -20,6 +23,19 @@ async function put_bucket_notification(req) {
         }
     } else {
         topic_configuration = [];
+    }
+
+    //test new notifications.
+    //if test notification fails, fail the put op
+    const err = await notif_util.test_notifications(topic_configuration,
+        req.object_sdk.nsfs_config_root);
+    if (err) {
+        throw new S3Error({
+            code: 'InvalidArgument',
+            message: JSON.stringify(err.message),
+            http_code: 400,
+            detail: err.toString()
+        });
     }
 
     const reply = await req.object_sdk.put_bucket_notification({
--- a/src/manage_nsfs/manage_nsfs_validations.js
+++ b/src/manage_nsfs/manage_nsfs_validations.js
@@ -15,6 +15,7 @@ const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, FROM_FILE, BOOLEAN_STRING_VA
     GLACIER_ACTIONS, LIST_UNSETABLE_OPTIONS, ANONYMOUS, DIAGNOSE_ACTIONS, UPGRADE_ACTIONS } = require('../manage_nsfs/manage_nsfs_constants');
 const { check_root_account_owns_user } = require('../nc/nc_utils');
 const { validate_username } = require('../util/validation_utils');
+const notifications_util = require('../util/notifications_util');
 
 /////////////////////////////
 //// GENERAL VALIDATIONS ////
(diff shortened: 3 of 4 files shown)


src/cmd/manage_nsfs.js
```diff
@@ -275,7 +275,7 @@ async function delete_bucket(data, force) {
  */
 async function bucket_management(action, user_input) {
     const data = action === ACTIONS.LIST ? undefined : await fetch_bucket_data(action, user_input);
-    await manage_nsfs_validations.validate_bucket_args(config_fs, data, action);
+    await manage_nsfs_validations.validate_bucket_args(config_fs, data, action, user_input);
```
shirady, line 278: Can we avoid adding a new argument `user_input` in the function signature?

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -403,6 +405,15 @@ async function validate_bucket_args(config_fs, data, action) {
             }
         }
     }
+
+    //if there's a change to the bucket's notifications, we need to test them
+    //if one of the specified notifications fail, we need to fail the user's request
+    if (user_input.notifications) {
+        const test_notif_err = await notifications_util.test_notifications(user_input, config_fs.connections_dir_path);
```
shirady, line 412: Can we pass only the `notifications` property from the bucket?

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -403,6 +405,15 @@ async function validate_bucket_args(config_fs, data, action) {
...
+        if (test_notif_err) {
+            throw_cli_error(ManageCLIError.InvalidArgument, "Failed to update notifications", test_notif_err);
```
shirady, line 414: Maybe we can use a more specific error than `InvalidArgument`?

src/cmd/manage_nsfs.js
alphaprinz: I didn't find a way.
The new input is required, the merged is not good enough.
When bucket name is changed, it is passed through a specific parameter ("new_name").

src/manage_nsfs/manage_nsfs_validations.js
alphaprinz: S3 errors don't have a specific error for this, I'd like to follow suit.

shirady: Sure, just notice that It's a `ManageCLIError`, so you have more flexibility in giving the precise error you need.

src/endpoint/s3/ops/s3_put_bucket_notification.js
```diff
@@ -20,6 +24,23 @@ async function put_bucket_notification(req) {
         }
     }
 
+    //test new notifications.
+    //if test notification fails, fail the put op
+    let config_fs = null;
+    if (req.object_sdk.nsfs_config_root) {
+        config_fs = new ConfigFS(req.object_sdk.nsfs_config_root);
```
shirady, line 31: I'm not sure adding `ConfigFS` as a dependency in an S3 op is something that we would like.
It might create coupling between the deployment way, and I would like to avoid it - the implementation according to the deployment (NC / Containerized) is in inner layers and not here.
Could you think of another alternative?

src/cmd/manage_nsfs.js
shirady: But I don't see where you use the bucket name exactly.
Another option is maybe to check it later in the flow.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -384,8 +385,9 @@ async function check_new_access_key_exists(config_fs, action, data) {
  * @param {import('../sdk/config_fs').ConfigFS} config_fs
  * @param {object} data
  * @param {string} action
+ * @param {object} user_input
  */
-async function validate_bucket_args(config_fs, data, action) {
+async function validate_bucket_args(config_fs, data, action, user_input) {
```
shirady, line 390: We want to avoid adding the argument in the function signature (it is not clear when to access `data` and when `user_input`).

src/endpoint/s3/ops/s3_put_bucket_notification.js
alphaprinz: Not exactly, but once/if we merge https://github.com/noobaa/noobaa-core/pull/8684, then I will only need req.object_sdk.nsfs_config_root because connection file directory will not be any directory specified in the config file, but rather only under the config root directory.

src/manage_nsfs/manage_nsfs_validations.js
alphaprinz: Bucket name? It is not used. The new notifications field is used (if it is provided).

There's a distinction between data and user_input - data is user_input merged with file's content.
Someone who changes this is supposed to understand this.
I will add a comment to clarify.

src/endpoint/s3/ops/s3_put_bucket_notification.js
shirady: Can it be moved to the `notification_utils`? (where I see that you already have modules related to FS.

src/endpoint/s3/ops/s3_put_bucket_notification.js
```diff
@@ -20,6 +24,23 @@ async function put_bucket_notification(req) {
...
+    }
+    const err = await notif_util.test_notifications(topic_configuration,
+        config_fs?.connections_dir_path);
+    if (err) {
+        throw new S3Error({
```
shirady, line 36: Why not `throw new S3Error(S3Error.InvalidArgument);`? (use existing object)
Why did you add the `detail` field? where do we see it? (could you add a print screen of what the user will see?)

shirady:
@alphaprinz could you add an automatic test for the fix?
If I understand correctly you can use the file `test_nc_nsfs_bucket_cli.test.js` as the non-existing connection is while creating or updating a new bucket.

nimrod-becker:
@dannyzaken can you PTAL ?

src/endpoint/s3/ops/s3_put_bucket_notification.js
alphaprinz: I just figures the error would be useful. Looks like this:

aprinzse:noobaa-core$ s3nsfsapi put-bucket-notification-configuration --bucket notif --notification-configuration file:////home/aprinzse/work/standalone/put_conf/put_configuration.json 
urllib3/connectionpool.py:1061: InsecureRequestWarning: Unverified HTTPS request is being made to host 'localhost'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings

An error occurred (InvalidArgument) when calling the PutBucketNotificationConfiguration operation: "ENOENT: no such file or directory, open '/etc/noobaa.conf.d/connections/connect_http.json'"

src/cmd/manage_nsfs.js
shirady: Maybe you can add the `notifications` argument only? (I want to avoid confusion between `data` and `user_input`).

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -445,6 +447,15 @@ async function validate_bucket_args(config_fs, data, action) {
             }
         }
     }
+
+    //if there's a change to the bucket's notifications, we need to test them
+    //if one of the specified notifications fail, we need to fail the user's request
+    if (user_input.notifications) {
+        const test_notif_err = await notifications_util.test_notifications(user_input.notifications, config_fs.connections_dir_path);
+        if (test_notif_err) {
+            throw_cli_error(ManageCLIError.InvalidArgument, "Failed to update notifications", test_notif_err);
```
shirady, line 456: Could you move this part to a separate function? (the function `validate_bucket_args` and then called it (the function has many lines before your change).

alphaprinz:
Re test - The [test PR](https://github.com/noobaa/noobaa-core/pull/8497) goes through this code path.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -447,6 +448,23 @@ async function validate_bucket_args(config_fs, data, action) {
     }
 }
 
+/**
+ * When setting notifications, we are supposed to send a test notification.
+ * If this test fails, we fail the user's request.
+ * @param {object} config_fs
+ * @param {object} notifications bucket's notification conf
+ * @returns Error encountered during notification test, if any
```
shirady, line 456: Minor - I think it doesn't match the use of JSDoc `@returns` ([reference](https://jsdoc.app/tags-returns)).

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -447,6 +448,23 @@ async function validate_bucket_args(config_fs, data, action) {
...
+ */
+async function validate_bucket_notifications(config_fs, notifications) {
+    if (!notifications) {
+        return;
+    }
+    const test_notif_err = await notifications_util.test_notifications(notifications, config_fs.config_root);
```
shirady, line 462: If an error is thrown how come we didn't wrap it with try-catch clause?
Maybe worth adding a comment here so it will be more clear and also a JSDoc on the function `test_notifications`.

src/util/notifications_util.js
```diff
@@ -337,22 +338,43 @@ function get_connection(connect) {
 }
 
 
-async function test_notifications(bucket, connect_files_dir) {
-    if (!bucket.notifications) {
+async function test_notifications(notifs, nc_config_dir) {
```
shirady, line 341: Could you add a JSDoc to this function? (to help us understand better `validate_bucket_notifications`).

shirady (approved):
LGTM

shirady:
> Re test - The [test PR](https://github.com/noobaa/noobaa-core/pull/8497) goes through this code path.

@alphaprinz Please feel free to mention this in the PR description.

src/manage_nsfs/manage_nsfs_validations.js
alphaprinz: Error is not thrown, it is returned.
I will add a comment.

Issue #8198, opened 2024-07-11
nsfs metrics from metrics port 7004 are only implemented over http. This should be changed to https as default. 


ps20renar:
**Is your feature request related to a problem? Please describe.**
Metrics port are only http. This is a script related issue that should be fixed

**Describe the solution you'd like**
For connecting to the metrics of nsfs the transport should be changed to https only.

**Describe alternatives you've considered**
Description of alternative solutions or features you've considered.

**Additional context**

ps20renar:
@romayalon, Any news about this issue?

romayalon:
Hey @ps20renar, still no news unfortunantly, we had other features as top priorities in the last months 
@madhuthorat @nimrod-becker let's discuss it in the next meeting

naveenpaul1:
@ps20renar we are planning to keep the non-secure port as default since our contanrized application(ODF) will use use non secure port for accessing metrics internaly. But we will provide the option to control the http and https port using configuration. In NC NSFS deploymnet customer will have option to disabe http and enable only ssl port.
cc @romayalon

naveenpaul1:
Backport planned for the verision 5.18

romayalon:
@madhuthorat Naveen merged a fix, can you start validating this fix?

PR #8673, opened 2025-01-10
SSL | Secure service for Noobaa metric

naveenpaul1:
### Explain the changes
1. add secure port for promethus service
2. control prometesu http and https service creation with flags.  non-secure service creation can be prevented using flag (ALLOW_HTTP_METRICS) in NC NSFS deployment only, and secure prometheus service will start only if ALLOW_HTTPS_METRICS flag is enabled for all the deployents.
3. Same S3 ssl certs are used by metrics server
4. Metrics https_port added in CLI and config for flexibility.

1. https://github.com/noobaa/noobaa-core/issues/8198

### Testing Instructions:
1. start nsfs application
2. verify metrics server is started by hitting url https://localhost:9443/
3. try the same with valid local certificate, steps can be found in [doc](https://ibm.ent.box.com/file/1389669232030?s=jj98n81p3tpfan60vuaqbrj7xk0s2ct2)

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -594,6 +594,7 @@ config.WS_METRICS_SERVER_PORT = 7001;
 config.BG_METRICS_SERVER_PORT = 7002;
 config.HA_METRICS_SERVER_PORT = 7003;
 config.EP_METRICS_SERVER_PORT = 7004;
+config.EP_METRICS_SERVER_SSL_PORT = 9443;
 
 //////////////////////////////
 // OAUTH RELATES            //
@@ -916,6 +917,8 @@ config.ENDPOINT_SSL_PORT = Number(process.env.ENDPOINT_SSL_PORT) || 6443;
 config.ENDPOINT_SSL_STS_PORT = Number(process.env.ENDPOINT_SSL_STS_PORT) || (process.env.NC_NSFS_NO_DB_ENV === 'true' ? -1 : 7443);
 config.ENDPOINT_SSL_IAM_PORT = Number(process.env.ENDPOINT_SSL_IAM_PORT) || -1;
 config.ALLOW_HTTP = false;
+config.ALLOW_HTTP_METRICS = true;
+config.ALLOW_HTTPS_METRICS = true;
 // config files should allow access to the owner of the files
 config.BASE_MODE_CONFIG_FILE = 0o600;
 config.BASE_MODE_CONFIG_DIR = 0o700;
--- a/src/cmd/nsfs.js
+++ b/src/cmd/nsfs.js
@@ -71,13 +71,14 @@ Arguments:
 const OPTIONS = `
 Options:
 
-    --http_port <port>         (default 6001)   Set the S3 endpoint listening HTTP port to serve.
-    --https_port <port>        (default 6443)   Set the S3 endpoint listening HTTPS port to serve.
-    --https_port_sts <port>    (default -1)     Set the S3 endpoint listening HTTPS port for STS.
-    --https_port_iam <port>    (default -1)     Set the endpoint listening HTTPS port for IAM.
-    --metrics_port <port>      (default -1)     Set the metrics listening port for prometheus.
-    --forks <n>                (default none)   Forks spread incoming requests (config.ENDPOINT_FORKS used if flag is not provided).
-    --debug <level>            (default 0)      Increase debug level.
+    --http_port <port>          (default 6001)      Set the S3 endpoint listening HTTP port to serve.
+    --https_port <port>         (default 6443)      Set the S3 endpoint listening HTTPS port to serve.
+    --https_port_sts <port>     (default -1)        Set the S3 endpoint listening HTTPS port for STS.
+    --https_port_iam <port>     (default -1)        Set the endpoint listening HTTPS port for IAM.
+    --http_metrics_port <port>      (default 7004)    Set the metrics listening HTTP port for prometheus.
+    --https_metrics_port <port>     (default 9443)    Set the metrics listening HTTPS port for prometheus.
+    --forks <n>                     (default none)  Forks spread incoming requests (config.ENDPOINT_FORKS used if flag is not provided).
+    --debug <level>                 (default 0)     Increase debug level.
 
     ## single user mode
 
@@ -265,7 +266,8 @@ async function main(argv = minimist(process.argv.slice(2))) {
         const https_port = Number(argv.https_port) || config.ENDPOINT_SSL_PORT;
         const https_port_sts = Number(argv.https_port_sts) || config.ENDPOINT_SSL_STS_PORT;
         const https_port_iam = Number(argv.https_port_iam) || config.ENDPOINT_SSL_IAM_PORT;
-        const metrics_port = Number(argv.metrics_port) || config.EP_METRICS_SERVER_PORT;
+        const http_metrics_port = Number(argv.http_metrics_port) || config.EP_METRICS_SERVER_PORT;
+        const https_metrics_port = Number(argv.https_metrics_port) || config.EP_METRICS_SERVER_SSL_PORT;
         const forks = Number(argv.forks) || config.ENDPOINT_FORKS;
         if (forks > 0) process.env.ENDPOINT_FORKS = forks.toString(); // used for argv.forks to take effect
         const uid = Number(argv.uid) || process.getuid();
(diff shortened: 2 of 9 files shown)


config.js
```diff
@@ -907,6 +909,8 @@ config.ENDPOINT_SSL_PORT = Number(process.env.ENDPOINT_SSL_PORT) || 6443;
 config.ENDPOINT_SSL_STS_PORT = Number(process.env.ENDPOINT_SSL_STS_PORT) || (process.env.NC_NSFS_NO_DB_ENV === 'true' ? -1 : 7443);
 config.ENDPOINT_SSL_IAM_PORT = Number(process.env.ENDPOINT_SSL_IAM_PORT) || -1;
 config.ALLOW_HTTP = false;
+config.ALLOW_HTTPS_METRICS = false;
```
romayalon, line 912: why false by default?

src/cmd/nsfs.js
```diff
@@ -71,13 +71,14 @@ Arguments:
...
-    --https_port_sts <port>    (default -1)     Set the S3 endpoint listening HTTPS port for STS.
-    --https_port_iam <port>    (default -1)     Set the endpoint listening HTTPS port for IAM.
-    --metrics_port <port>      (default -1)     Set the metrics listening port for prometheus.
-    --forks <n>                (default none)   Forks spread incoming requests (config.ENDPOINT_FORKS used if flag is not provided).
-    --debug <level>            (default 0)      Increase debug level.
+    --http_port <port>              (default 6001)   Set the S3 endpoint listening HTTP port to serve.
+    --https_port <port>             (default 6443)   Set the S3 endpoint listening HTTPS port to serve.
+    --https_port_sts <port>         (default -1)     Set the S3 endpoint listening HTTPS port for STS.
+    --https_port_iam <port>         (default -1)     Set the endpoint listening HTTPS port for IAM.
+    --metrics_port       <port>     (default -1)     Set the metrics listening port for prometheus.
```
romayalon, line 78: 1. indentation of <port>
2. Add HTTP so it'll be clear like we have in the S3 HTTP port

src/cmd/nsfs.js
```diff
@@ -71,13 +71,14 @@ Arguments:
...
+    --metrics_https_port <port>     (default -1)     Set the metrics listening https port for prometheus.
```
romayalon, line 79: 1. make sure HTTPS is capital letters like the other HTTPS usage
2. why default is -1?

config.js
```diff
@@ -586,6 +586,7 @@ config.MIN_MEMORY_FOR_UPGRADE = 1200 * 1024 * 1024;
 config.SEND_EVENTS_REMOTESYS = true;
 config.PROMETHEUS_ENABLED = true;
 config.PROMETHEUS_SERVER_RETRY_COUNT = Infinity;
+config.PROMETHEUS_SSL_SERVER_RETRY_COUNT = Infinity;
```
romayalon, line 589: why do we need another one?

src/util/ssl_utils.js
```diff
@@ -46,7 +46,8 @@ const certs = {
     S3: new CertInfo(config.S3_SERVICE_CERT_PATH),
     EXTERNAL_DB: new CertInfo(config.EXTERNAL_DB_SERVICE_CERT_PATH),
     STS: new CertInfo(config.STS_SERVICE_CERT_PATH),
-    IAM: new CertInfo(config.IAM_SERVICE_CERT_PATH)
+    IAM: new CertInfo(config.IAM_SERVICE_CERT_PATH),
+    METRICS: new CertInfo(config.S3_SERVICE_CERT_PATH) // metric server will use the S3 cert.
```
romayalon, line 50: I know you have been told in the design meeting to use the S3 cert but I know @Neon-White had some work on changing STS cert to not re-use the S3 cert due to an issue mentioned in https://github.com/noobaa/noobaa-core/pull/8123
so per what Ben wrote, we might have an issue with this re-use on containerized env, Ben WDYT?
CC: @nimrod-becker

romayalon: on the other hand, on non containerized - we currently have the certificates under certificates/ dir, and adding a new certificate will require some changes in the code/structure for allowing different certificates for different services.

src/util/fork_utils.js
```diff
@@ -28,11 +28,13 @@ const fs_workers_stats = {};
  * In case of any worker exit, also the entire process group will exit.
  * @see https://nodejs.org/api/cluster.html
  * 
- * @param {number?} count number of workers to start.
  * @param {number?} metrics_port prometheus metris port.
+ * @param {number?} metrics_ssl_port prometheus metris https port.
+ * @param {number?} count number of workers to start.
+ * @param {string?} nsfs_config_root nsfs configuration path
```
romayalon, line 34:
```suggestion
 * @param {number} [metrics_port]
 * @param {number} [metrics_ssl_port]
 * @param {number} [count] number of workers to start.
 * @param {string} [nsfs_config_root] nsfs configuration path
```

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -144,6 +159,50 @@ async function start_server(
             }
         }
     }
+
+    while (retry_https_count) {
+        try {
+            if (ssl_port > 0 && config.ALLOW_HTTPS_METRICS) {
+                const ssl_cert_info = await ssl_utils.get_ssl_cert_info('METRICS', nsfs_config_root);
```
romayalon, line 166: Can we re-use start_server_and_cert() like the other services?

src/util/ssl_utils.js
romayalon: yep, thanks Ben! so I looks like we can re-use the S3-cert, but this must be documented

config.js
naveenpaul1: @romayalon ODF enabling https dosn't make sense without the `Service` changes. But we can enable it by default and can be used with service yaml changes and non-containerized can use http port without any config changes.

src/util/ssl_utils.js
naveenpaul1: Sure will update the doc

src/server/analytic_services/prometheus_reporting.js
naveenpaul1: @romayalon We can reuse `start_server_and_cert()` only because of the `endpoint_request_handler` hold all the endpoint for s3, sts and iam. That is not the case for prometheus, prometheus handler is defined in `prometheus_reporting.js` and for reusing the `start_server_and_cert()` we need to move the `metrics_request_handler` to endpoint.js but that again add the condition block inside `start_server_and_cert()` and mix prometheus and endpoint code.

config.js
romayalon: we wanted to support it on ODF and on NC, so I don't understand why it's false
which https changes are you referring to?

naveenpaul1: yes got your point, Will update value with `true` and enable HTTPS. ODF also can fetch metrics from SSL port without any config change in Noobaa side.

src/server/analytic_services/prometheus_reporting.js
romayalon: moving `metrics_request_handler` to the `endpoint.js` is not the solution we should pursue...
I can give you an idea of how to start the changes that - 
1. rename `start_server_and_cert()` to `start_endpoint_server_and_cert()`
2. extract from `start_server_and_cert()` the HTTP start server, and the HTTPS start server into 2 new functions and locate them in http_utils.
3. rename `promehteus_reporting.start_server()` to `start_metrics_server()`
4. call the new 2 HTTP and HTTPS servers start from `start_metrics_server()`

let me know what you think and if you need help in this refactoring

src/util/http_utils.js
```diff
@@ -750,7 +751,115 @@ function http_get(uri, options) {
...
+/**
+ * Listen server for http/https ports
+ * @param {number} port
+ * @param {http.Server} server prometheus metris serer.
+ * @param {('S3'|'IAM'|'STS'|'METRICS')} server_type 
+*/
+function listen_http(port, server, server_type) {
+    return new Promise((resolve, reject) => {
+        if (server_type !== 'METRICS') {
+            setup_http_server(server);
```
romayalon, line 797: @dannyzaken Do you remember maybe why we didn't have a server setup like this for metrics? do you think we should add it for metrics as well?

src/util/http_utils.js
```diff
@@ -750,7 +751,115 @@ function http_get(uri, options) {
...
+                reject(err);
+            } else {
+                resolve();
+            }
+        });
+    });
+}
+
+
+function setup_http_server(server) {
```
romayalon, line 811: both functions here are endpoint servers, please add it to a jsdoc or even to the functions names so no one will set non endpoint server using these functions

romayalon (approved):

PR #8679, opened 2025-01-13
Bumping deps to avoid CVE (13/01/2025)

liranmauda:
### Explain the changes
- Bumping deps to avoid CVE (13/01/2025)


shirady (approved):
LGTM

PR #8680, opened 2025-01-13
NC | NSFS | Fix Issue DFBUGS-1307 | Bucket Policy With Principal as ID

shirady:
### Explain the changes
1. Change the condition in `rest_s3` and `bucketspace_fs` to check the permission by the principal on account name when the previous check was only not DENY, so in case someone puts several statements using both account ID and account name, we won't skip it after the checks on the account ID.
2. Add logs in level 3 to help investigate bucket policy issues.

### Issues: Fixed [Jira-1307](https://issues.redhat.com/browse/DFBUGS-1307)
Currently, we had an issue when someone put a bucket policy with two statements:
1. Allow by name to all principals ("*").
2. Deny by account ID to a specific account.
We used the first one only as it was not "IMPLICIT_DENY" and didn't check the second option.

GAPs:
1. I think we need to create constants for the bucket policy and reuse them instead of a raw string (from example `Deny`. `Allow`) - opened #8697.
2. I think we need to add error logs in case of failure to help us with an investigation - opened #8699, for example:
https://github.com/noobaa/noobaa-core/blob/a5b95e75515ac461659cb20a91558c5c58f30df1/src/sdk/bucketspace_fs.js#L672-L674

### Testing Instructions:
Automatic test:
1. In NC Please run: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_s3_bucket_policy.js`
4. In containerized Please run: `make run-single-test testname=test_s3_bucket_policy.js CONTAINER_PLATFORM=linux/arm64` (I have MacOS so I use the flag `CONTAINER_PLATFORM=linux/arm64`).

Manual Test:
5. Add bucket to the account using AWS CLI: `nc-user-1-s3 s3 mb s3://bucket-1307` (`bucket-1307` is the bucket name in this example)
6. Put an object in the bucket: `echo 'test_data' | nc-user-1-s3 s3 cp - s3://bucket-1307/test_object.txt`
7. Create two additional account with the CLI: `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path /Users/buckets/ --access_key <access-key> --secret_key <secret-key> --uid <uid> --gid <gid>` (use the same `uid` and `gid` as the first created account).
8. Create the alias for the S3 service: alias nc-user-2-s3=‘AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443’`.
9. Check the connection to the endpoint and try to list the buckets (should be empty): `nc-user-2-s3 s3 ls; echo $?`

**Statement with * allow, statement with account name deny**
10. Put bucket policy (statement with account name): `nc-user-1-s3 s3api put-bucket-policy --bucket bucket-1307 --policy file://policy1.json`
policy1.json (replace the principal with the account name in the Deny statement):
```
{
    "Statement": [
       {
          "Effect": "Allow",
          "Principal": {
            "AWS": [
                "*"
            ]
          },
          "Action": "s3:*",
          "Resource": [
          "arn:aws:s3:::*",
          "arn:aws:s3:::bucket-1307/*"
        ]
       },
       {
          "Effect": "Deny",
          "Action": "s3:GetObject",
          "Principal": {
            "AWS": [
                "shira-1003-1"
            ]
          },          
          "Resource": [
            "arn:aws:s3:::bucket-1307/*"
          ]
        }
    ]
 }
```
11. Account 2 tries to get the object: `nc-user-2-s3 s3api get-object --bucket bucket-1307 --key test_object.txt /dev/stdout` (should get `AccessDenied` error)

**Statement with * allow, statement with account ID deny**
12. Put bucket policy (statement with account ID): `nc-user-1-s3 s3api put-bucket-policy --bucket bucket-1307 --policy file://policy2.json`
policy2.json (replace the principal with the account ID in the Deny statement)::
```
{
    "Statement": [
       {
          "Effect": "Allow",
          "Principal": {
            "AWS": [
                "*"
            ]
          },
          "Action": "s3:*",
          "Resource": [
          "arn:aws:s3:::*",
          "arn:aws:s3:::bucket-1307/*"
        ]
       },
       {
          "Effect": "Deny",
          "Action": "s3:GetObject",
          "Principal": {
            "AWS": [
                "6784c11903ab3d2123b2a1c2"
            ]
          },          
          "Resource": [
            "arn:aws:s3:::bucket-1307/*"
          ]
        }
    ]
 }
```
13. Account 2 tries to get the object: `nc-user-2-s3 s3api get-object --bucket bucket-1307 --key test_object.txt /dev/stdout` (should get `AccessDenied` error)

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -277,23 +277,27 @@ async function authorize_request_policy(req) {
         if (is_owner || is_iam_account_and_same_root_account_owner) return;
         throw new S3Error(S3Error.AccessDenied);
     }
-    let permission;
+    // in case we have bucket policy
+    let permission_by_id;
+    let permission_by_name;
     // In NC, we allow principal to be:
     // 1. account name (for backwards compatibility)
     // 2. account id
     // we start the permission check on account identifier intentionally
     if (account_identifier_id) {
-        permission = await s3_bucket_policy_utils.has_bucket_policy_permission(
+        permission_by_id = await s3_bucket_policy_utils.has_bucket_policy_permission(
             s3_policy, account_identifier_id, method, arn_path, req);
+        dbg.log3('authorize_request_policy: permission_by_id', permission_by_id);
     }
+    if (permission_by_id === "DENY") throw new S3Error(S3Error.AccessDenied);
 
-    if ((!account_identifier_id || permission === "IMPLICIT_DENY") && account.owner === undefined) {
-        permission = await s3_bucket_policy_utils.has_bucket_policy_permission(
+    if ((!account_identifier_id || permission_by_id !== "DENY") && account.owner === undefined) {
+        permission_by_name = await s3_bucket_policy_utils.has_bucket_policy_permission(
             s3_policy, account_identifier_name, method, arn_path, req);
+        dbg.log3('authorize_request_policy: permission_by_name', permission_by_name);
     }
-
-    if (permission === "DENY") throw new S3Error(S3Error.AccessDenied);
-    if (permission === "ALLOW" || is_owner) return;
+    if (permission_by_name === "DENY") throw new S3Error(S3Error.AccessDenied);
+    if ((permission_by_id === "ALLOW" || permission_by_name === "ALLOW") || is_owner) return;
 
     throw new S3Error(S3Error.AccessDenied);
 }
--- a/src/test/unit_tests/test_s3_bucket_policy.js
+++ b/src/test/unit_tests/test_s3_bucket_policy.js
@@ -10,6 +10,7 @@ const { rpc_client, EMAIL, POOL_LIST, anon_rpc_client } = coretest;
 const MDStore = require('../../server/object_services/md_store').MDStore;
 coretest.setup({ pools_to_create: process.env.NC_CORETEST ? undefined : [POOL_LIST[1]] });
 const path = require('path');
+const _ = require('lodash');
 const fs_utils = require('../../util/fs_utils');
 
 const { S3 } = require('@aws-sdk/client-s3');
@@ -33,6 +34,7 @@ async function assert_throws_async(promise, expected_message = 'Access Denied') 
 const BKT = 'test2-bucket-policy-ops';
 const BKT_B = 'test2-bucket-policy-ops-1';
 const BKT_C = 'test2-bucket-policy-ops-2';
+const BKT_D = 'test2-bucket-policy-ops-3';
 const KEY = 'file1.txt';
 const user_a = 'alice';
 const user_b = 'bob';
@@ -134,6 +136,7 @@ async function setup() {
     s3_owner = new S3(s3_creds);
     await s3_owner.createBucket({ Bucket: BKT });
     await s3_owner.createBucket({ Bucket: BKT_C });
+    await s3_owner.createBucket({ Bucket: BKT_D });
     s3_anon = new S3({
         ...s3_creds,
         credentials: {
@@ -147,7 +150,7 @@ async function setup() {
     });
 }
 
-/*eslint max-lines-per-function: ["error", 1600]*/
+/*eslint max-lines-per-function: ["error", 2000]*/
 mocha.describe('s3_bucket_policy', function() {
     mocha.before(setup);
     mocha.it('should fail setting bucket policy when user doesn\'t exist', async function() {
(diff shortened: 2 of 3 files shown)


src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -335,6 +337,100 @@ mocha.describe('s3_bucket_policy', function() {
         }));
     });
 
+    mocha.it('should not allow principal get object bucket policy with 2 statements: ' +
+        '(1) DENY principal by account ID (2) ALLOW account name as *', async function() {
+        if (!is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+        const policy = {
+            Statement: [{
+                    Sid: `Allow all s3 actions on bucket ${BKT_D} to all principals`,
```
romayalon, line 345: this policy statement is being re-used later, please put in a const

src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -335,6 +337,100 @@ mocha.describe('s3_bucket_policy', function() {
...
+            Body: BODY,
+            Bucket: BKT_D,
+            Key: key2
+        });
+        assert.equal(res_get_object.$metadata.httpStatusCode, 200);
+    });
+
+    mocha.it('should not allow principal get object bucket policy with 2 statements: ' +
+        '(1) DENY principal by account name (2) ALLOW account name as *', async function() {
+        if (!is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
```
romayalon, line 389: why skip this test on containerized?

src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -335,6 +337,100 @@ mocha.describe('s3_bucket_policy', function() {
         }));
     });
 
+    mocha.it('should not allow principal get object bucket policy with 2 statements: ' +
```
romayalon, line 340: could you think about more tests cases that can cause conflicts? 
I could think of for example - 
1. 2 statements same account - first statement by id = deny, second statement by name = allow
2. 2 statements - first deny all, second - allow account

shirady: I can try, thanks.

src/test/unit_tests/test_s3_bucket_policy.js
shirady: It is not the same policy, there's one mutual statement (allow to all participants with `*`), I'll try to reuse it.

src/test/unit_tests/test_s3_bucket_policy.js
shirady: I will try to remove this skip, thanks.

src/test/unit_tests/test_s3_bucket_policy.js
romayalon: I meant to the allow all statement, not the whole policy

src/test/unit_tests/test_s3_bucket_policy.js
shirady: I didn't add the "deny all" statement as it would not allow me any operation on it (even the preparation step of adding the object and the policy deletion in the end).
So I added your first idea as two additional tests, thanks.

romayalon: you can add deny all user from a specific operation, not necessary all operations...

shirady: I can try, thanks.

romayalon (approved):

src/endpoint/s3/s3_rest.js
```diff
@@ -277,23 +277,27 @@ async function authorize_request_policy(req) {
...
     if (account_identifier_id) {
-        permission = await s3_bucket_policy_utils.has_bucket_policy_permission(
+        permission_by_id = await s3_bucket_policy_utils.has_bucket_policy_permission(
             s3_policy, account_identifier_id, method, arn_path, req);
+        dbg.log3('authorize_request_policy: permission_by_id', permission_by_id);
     }
 
-    if ((!account_identifier_id || permission === "IMPLICIT_DENY") && account.owner === undefined) {
-        permission = await s3_bucket_policy_utils.has_bucket_policy_permission(
+    if ((!account_identifier_id || permission_by_id !== "DENY") && account.owner === undefined) {
```
alphaprinz, line 293: Somewhat nitpicking, but you can remove the 
`|| permission_by_id !== 'DENY'`
because if permission_by_id === 'DENY' the result would be S3Error.AccessDenied anyway.

If you're concerned about efficiency then it would be more efficient to throw the S3Error.AccessDenied in line 292.

src/test/unit_tests/test_s3_bucket_policy.js
shirady: I added your second idea as well, thanks.

src/endpoint/s3/s3_rest.js
shirady: I changed the code to your suggestion.

alphaprinz:
LGTM :)

romayalon (approved):

PR #8682, opened 2025-01-13
NC | Online Upgrade | Integration tests | Add CLI and S3 integration tests

romayalon:
### Explain the changes
1. Added integration tests for CLI and S3 commands during a mocked online upgrade (before the upgrade finished).
2. These tests assert that - 
    - config directory updates are not allowed via the CLI nor via S3.
    - S3 Objects operations are allowed.
    - CLI list/status commands are allowed.

### Testing Instructions:
1. online upgrade + manage CLI integration tests - `sudo  jest --testRegex=jest_tests/test_nc_online_upgrade_cli_integrations`
4. online upgrade + S3 operations - `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nc_online_upgrade_s3_integrations.js`

- [x] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/test/unit_tests/test_nc_online_upgrade_s3_integrations.js
+++ b/src/test/unit_tests/test_nc_online_upgrade_s3_integrations.js
@@ -1,0 +1,193 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const path = require('path');
+const http = require('http');
+const mocha = require('mocha');
+const assert = require('assert');
+const config = require('../../../config');
+const { get_coretest_path, TMP_PATH, TEST_TIMEOUT, update_system_json } = require('../system_tests/test_utils');
+const { S3 } = require('@aws-sdk/client-s3');
+const { NodeHttpHandler } = require("@smithy/node-http-handler");
+const { folder_delete } = require('../../util/fs_utils');
+const coretest_path = get_coretest_path();
+const coretest = require(coretest_path);
+const { rpc_client, EMAIL, NC_CORETEST_CONFIG_FS } = coretest;
+coretest.setup({});
+
+const new_buckets_path = path.join(TMP_PATH, 'test_nc_online_upgrade_new_buckets_path');
+const bucket_storage_path = path.join(new_buckets_path, 'bucket1');
+const mock_old_config_dir_version = '0.0.0';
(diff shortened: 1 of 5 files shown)


src/test/unit_tests/test_nc_online_upgrade_s3_integrations.js
```diff
@@ -0,0 +1,190 @@
...
+        await s3_client.deleteBucket({ Bucket: bucket_defaults.name });
+    }, TEST_TIMEOUT);
+
+    mocha.it('head bucket', async function() {
+        await create_default_bucket(s3_client, bucket_defaults.name);
+        await update_system_json(NC_CORETEST_CONFIG_FS, mock_old_config_dir_version);
+        await s3_client.headBucket({ Bucket: bucket_defaults.name });
+    }, TEST_TIMEOUT);
+
+    mocha.it('PUT object - host is blocked for config dir updates', async function() {
```
naveenpaul1, line 141: only bucket specific actions will get restricted while upgrading, not the object operations, Is that right?

src/test/unit_tests/test_nc_online_upgrade_s3_integrations.js
```diff
@@ -0,0 +1,190 @@
...
+        await create_default_bucket(s3_client, bucket_defaults.name);
+        await update_system_json(NC_CORETEST_CONFIG_FS, mock_old_config_dir_version);
+        await s3_client.putObject({ Bucket: bucket_defaults.name, Key: mock_obj_key, Body: mock_obj_body });
+    }, TEST_TIMEOUT);
+
+    mocha.it('GET object - host is blocked for config dir updates', async function() {
```
naveenpaul1, line 147: may be something like 
`GET object - host is blocked for config dir updates - Should not fail object action`
Will help to understand the test better.

src/test/unit_tests/test_nc_online_upgrade_s3_integrations.js
romayalon: create/update/delete bucket ops are restricted, 
list/get/head bucket is allowed
object operations are allowed

naveenpaul1 (approved):

PR #8685, opened 2025-01-14
bucket notification - introduce batch param

alphaprinz:
### Explain the changes
1. Processing a large enough number of pending notifications results in timeouts.
This fix sends notifications in batches to avoid too many inflight requests.

1. 

### Testing Instructions:
1. Create a lot of pending notifications (I got timeouts with about 300).
2. Processing the notifications shouldn't result in timeout (assuming target server has reasonable response time).

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -712,6 +712,7 @@ config.BUCKET_LOG_CONCURRENCY = 10;
 ////////////////////////////////
 config.NOTIFICATION_LOG_NS = 'notification_logging';
 config.NOTIFICATION_LOG_DIR = process.env.NOTIFICATION_LOG_DIR;
+config.NOTIFICATION_BATCH = process.env.BATCH || 10;
 
 ///////////////////////////
 //      KEY ROTATOR      //
--- a/src/util/notifications_util.js
+++ b/src/util/notifications_util.js
@@ -35,13 +35,14 @@ class Notificator {
      * @param {Object} options
      */
 
-    constructor({name, fs_context, connect_files_dir, nc_config_fs}) {
+    constructor({name, fs_context, connect_files_dir, nc_config_fs, batch_size}) {
         this.name = name;
         this.connect_str_to_connection = new Map();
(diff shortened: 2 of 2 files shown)


config.js
```diff
@@ -704,6 +704,7 @@ config.BUCKET_LOG_CONCURRENCY = 10;
 ////////////////////////////////
 config.NOTIFICATION_LOG_NS = 'notification_logging';
 config.NOTIFICATION_LOG_DIR = process.env.NOTIFICATION_LOG_DIR;
+config.NOTIFICATION_BATCH = process.env.BATCH;
```
aspandey, line 707: Why can't we do it here instead of providing default value ,10 at the time of usage ?
config.NOTIFICATION_BATCH = process.env.BATCH || 10;

src/util/notifications_util.js
```diff
@@ -35,13 +35,14 @@ class Notificator {
...
 
-    constructor({name, fs_context, connect_files_dir, nc_config_fs}) {
+    constructor({name, fs_context, connect_files_dir, nc_config_fs, batch_size}) {
         this.name = name;
         this.connect_str_to_connection = new Map();
         this.notif_to_connect = new Map();
         this.fs_context = fs_context ?? get_process_fs_context();
         this.connect_files_dir = connect_files_dir ?? DEFAULT_CONNECT_FILES_DIR;
         this.nc_config_fs = nc_config_fs;
+        this.batch_size = batch_size || config.NOTIFICATION_BATCH || 10;
```
aspandey, line 45: I thought you will obviously remove it so I did not give comment on this.
Here now you can remove 10. right?

alphaprinz: Might be over-protective in practice.
In theory someone might nullify config.NOTIFICATION_BATCH.

aspandey (approved):

PR #8687, opened 2025-01-14
NSFS | set supplemental groups dynamically to users groups

### LINKED ISSUE #8712: NSFS | macos | add dynamic supplemental groups for macos
continuation of #8687 

should be similar to #8687 but require implementation and testing with mac platform.

Description of alternative solutions or features you've considered.

**Additional context**
since all functions used are defined by posix it should be the same. but still require testing
also added helper testing functions for adding new group that require implementation for mac. should probably be done with `sudo dscl . ` like adding new user, but again need to test this

nadavMiz:
### Explain the changes
1. get user supplemental dynamically by calling [getgrouplist](https://man7.org/linux/man-pages/man3/getgrouplist.3.html). getgrouplist will scan the system for all groups that the user is a part of. that way will be able to set the users groups without specifying them in the account configuration. if specified, supplemental groups set by the account configuration will take precedence, and be set as the users groups. also in case no record of the user can be found in the os and no supplemental groups were defined in the account the supplemental groups list will be set as empty  
2. some tests were using the uid 5, which is a system reserved uid, causing the feature to give them root group permissions. changed the uid and gid to ones not reserved by the OS
3. there was a suggestion to use [`getlogin_r`](https://linux.die.net/man/3/getlogin_r) to get the user name for getgrouplist. this function gets the name of the user logged to the terminal. this function uses the `/var/run/utmp`. in my implementation I used 
`getpwuid` instead which gets the user name by uid using the `/etc/passwd` file. I used it because its more flexible and according to documentation `getlogin_r` uses the `/etc/passwd` file anyways. also `getlogin_r` might not work because the uid was switched to be different then the one used to run the process(root)
4. added implementation and tests for mac-os. run the tests same as on linux. only run on macos environment

#### Gaps
#8710 - list objects fail when using supplemental groups

continue #8552 
### Testing Instructions:
before running all tests on the code changes please run:
`npm run build:native`
manual tests (examples for linux):
1. create new group in the os:
`groupadd -g 1022 new_group`
2. create new user in the os:
`useradd -c test_user2 -m test_user2 -p -u 1033 -g 1034`
3. add user to the new group:
`usermod -a -G new_group test_user2`
4. create new noobaa account with the new groups gid:
`sudo node src/cmd/manage_nsfs.js account add --name test_user1 --uid 1022 --gid 1022`
5. create a second account with the new users gid and uid:
`sudo node src/cmd/manage_nsfs.js account add --name test_user1 --uid 1033 --gid 1034`
6. create a new bucket using the forst user
    `test_user1_s3api create-bucket --bucket test_bcuket`
7. add bucket policy to allow putObject to test_user2
8. put new object to the new bucket using the second user test_user2:
    `test_user2_s3api put-object --bucket test_bcuket --key key1 --body <path_to_body>`
9. operation should succeed 

automatic tests:
`sudo node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nsfs_access.js`

Note: building and testing was done on Linux and Mac machines

- [x] Doc added/updated
- [x] Tests added


src/test/unit_tests/test_nsfs_access.js
```diff
@@ -52,16 +54,27 @@ mocha.describe('new tests check', async function() {
...
+        backend: '',
+        warn_threshold_ms: 100,
+    };
     mocha.before(async function() {
         if (test_utils.invalid_nsfs_root_permissions()) this.skip(); // eslint-disable-line no-invalid-this
         await fs_utils.create_fresh_path(p, 0o777);
         await fs_utils.file_must_exist(p);
         await fs_utils.create_fresh_path(full_path_root, 0o770);
         await fs_utils.file_must_exist(full_path_root);
+        await test_utils.create_fs_user_by_platform(test_user_name, test_user_name, 1575, 1575); //non root 4
```
shirady, line 70: Instead of passing the numbers as hard coded, you can pass them from the objectNON_ROOT4_FS_CONFIG directly:
```suggestion
        await test_utils.create_fs_user_by_platform(test_user_name, test_user_name, NON_ROOT4_FS_CONFIG.uid, NON_ROOT4_FS_CONFIG.gid);
```

src/test/unit_tests/test_nsfs_access.js
```diff
@@ -52,16 +54,27 @@ mocha.describe('new tests check', async function() {
...
+        await test_utils.create_fs_group_by_platform(test_group_name, 1572, test_user_name); //non root 1 group
```
shirady, line 71: Same here:
```suggestion
        await test_utils.create_fs_group_by_platform(test_group_name, NON_ROOT1_FS_CONFIG.gid, test_user_name);
```

src/test/system_tests/test_utils.js
```diff
@@ -335,7 +335,40 @@ async function delete_fs_user_by_platform(name) {
...
 }
 
-/** 
+/**
+ * creates a new file system group. if user_name is define add it to the group
+ * @param {string} group_name
+ * @param {number} gid
+ * @param {string} user_name
+ */
+async function create_fs_group_by_platform(group_name, gid, user_name) {
```
shirady, line 344: Do you care only about the platform (Linux/MacOS)? Running on another FS would be fine? (GPFS for example)

src/test/unit_tests/test_nsfs_access.js
```diff
@@ -127,7 +140,31 @@ mocha.describe('new tests check', async function() {
         try {
             //non root3 doesn't have non-root2 group as supplemental group, so it should fail
             const non_root_entries = await nb_native().fs.readdir(NON_ROOT3_FS_CONFIG, full_path_non_root2);
-            assert.fail(`non root 3 has access to a folder created by user with gid not insupplemental groups - ${p} ${non_root_entries}`);
+            assert.fail(`non root 4 has access to a folder created by user with gid not insupplemental groups - ${p} ${non_root_entries}`);
+        } catch (err) {
+            assert.equal(err.code, 'EACCES');
+        }
```
shirady, line 146: Why this test was changed?

src/native/util/os_linux.cpp
```diff
@@ -29,6 +30,50 @@ const gid_t ThreadScope::orig_gid = getgid();
...
+    struct passwd* pw = getpwuid(uid);
+    if (pw == NULL) {
+        if (errno == 0) {
+            LOG("get_supplemental_groups_by_uid: no record for uid " << uid);
+        } else {
+            LOG("WARNING: get_supplemental_groups_by_uid: getpwuid failed: " << strerror(errno));
+        }
+        return -1;
+    }
+    int ngroups = NGROUPS_MAX;
```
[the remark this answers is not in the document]
nadavMiz: NGROUPS_MAX is a system defined macro that defines the maximum size of the suplemental group list. looks like on linux it is defined to be 65536. but I can regardless if it has different meaning on macos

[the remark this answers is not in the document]
shirady: I probably saw something else, we'll check again.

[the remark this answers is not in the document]
shirady: I clicked on it and saw this:
Screenshot 2025-01-21 at 12 04 09

src/test/system_tests/test_utils.js
nadavMiz: GPFS is a linux platform

src/test/unit_tests/test_nsfs_access.js
nadavMiz: I accidentally changed the comment. I changed it again now because it had a typo. the change itself was not changed. it just look like it because of git

src/native/util/os_linux.cpp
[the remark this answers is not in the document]
nadavMiz: then I guess this is the value defined by the system. anyway this PR is for linux. we can test this on mac, and if it doesn't work handle this in a different PR

shirady (approved):
LGTM

src/test/unit_tests/test_bucketspace_versioning.js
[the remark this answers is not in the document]
nadavMiz: I guess there are no permission checks in this files, so its tests doesn't fail. I guess we can change it in a different PR just so we would not use system reserved UID

PR #8700, opened 2025-01-20
Clean md store objects

vh05:
Clean md store objects once the max deleted objects reach the limit.

Fixes: https://issues.redhat.com/browse/DFBUGS-1339

### DIFF
--- a/src/server/bg_services/db_cleaner.js
+++ b/src/server/bg_services/db_cleaner.js
@@ -58,23 +58,39 @@ async function clean_md_store(last_date_to_remove) {
         ${total_objects_count} objects - Skipping...`);
         return;
     }
+    const objects_to_remove = await clean_md_store_objects(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
+    const blocks_to_remove = await clean_md_store_blocks(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
+    const filtered_chunks = await clean_md_store_chunks(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
+
+    dbg.log0(`DB_CLEANER: removed ${objects_to_remove.length + blocks_to_remove.length + filtered_chunks.length} documents from md-store`);
+}
+
+async function clean_md_store_objects(last_date_to_remove, limit) {
     dbg.log0('DB_CLEANER: checking md-store for documents deleted before', new Date(last_date_to_remove));
-    const objects_to_remove = await MDStore.instance().find_deleted_objects(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
+    const objects_to_remove = await MDStore.instance().find_deleted_objects(last_date_to_remove, limit);
     dbg.log2('DB_CLEANER: list objects:', objects_to_remove);
-    if (objects_to_remove.length) {
+    if (objects_to_remove.length > config.MD_STORE_MAX_DELETED_OBJECTS_LIMIT) {
         await P.map_with_concurrency(10, objects_to_remove, obj => db_delete_object_parts(obj));
         await MDStore.instance().db_delete_objects(objects_to_remove);
     }
-    const blocks_to_remove = await MDStore.instance().find_deleted_blocks(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
+    return objects_to_remove;
+}
+
+async function clean_md_store_blocks(last_date_to_remove, limit) {
+    const blocks_to_remove = await MDStore.instance().find_deleted_blocks(last_date_to_remove, limit);
     dbg.log2('DB_CLEANER: list blocks:', blocks_to_remove);
     if (blocks_to_remove.length) await MDStore.instance().db_delete_blocks(blocks_to_remove);
-    const chunks_to_remove = await MDStore.instance().find_deleted_chunks(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
+    return blocks_to_remove;
+}
+
+async function clean_md_store_chunks(last_date_to_remove, limit) {
+    const chunks_to_remove = await MDStore.instance().find_deleted_chunks(last_date_to_remove, limit);
     const filtered_chunks = chunks_to_remove.filter(async chunk =>
         !(await MDStore.instance().has_any_blocks_for_chunk(chunk)) &&
         !(await MDStore.instance().has_any_parts_for_chunk(chunk)));
     dbg.log2('DB_CLEANER: list chunks with no blocks and no parts to be removed from DB', filtered_chunks);
     if (filtered_chunks.length) await MDStore.instance().db_delete_chunks(filtered_chunks);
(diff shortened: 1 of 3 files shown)


src/server/bg_services/db_cleaner.js
```diff
@@ -58,23 +58,39 @@ async function clean_md_store(last_date_to_remove) {
         ${total_objects_count} objects - Skipping...`);
         return;
     }
+    const objects_to_remove = await clean_md_store_objects(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
```
jackyalbo, line 61: Did you rearrange the same code in smaller functions? I'm not sure if it's needed.

src/server/bg_services/db_cleaner.js
```diff
@@ -58,23 +58,39 @@ async function clean_md_store(last_date_to_remove) {
...
+    dbg.log0(`DB_CLEANER: removed ${objects_to_remove.length + blocks_to_remove.length + filtered_chunks.length} documents from md-store`);
+}
+
+async function clean_md_store_objects(last_date_to_remove, limit) {
     dbg.log0('DB_CLEANER: checking md-store for documents deleted before', new Date(last_date_to_remove));
-    const objects_to_remove = await MDStore.instance().find_deleted_objects(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
+    const objects_to_remove = await MDStore.instance().find_deleted_objects(last_date_to_remove, limit);
     dbg.log2('DB_CLEANER: list objects:', objects_to_remove);
-    if (objects_to_remove.length) {
+    if (objects_to_remove.length > config.MD_STORE_MAX_DELETED_OBJECTS_LIMIT) {
```
jackyalbo, line 72: why are we adding this limit? why not deleting less?

src/server/bg_services/objects_reclaimer.js
```diff
@@ -42,6 +43,9 @@ class ObjectsReclaimer {
         if (has_errors) {
             return config.OBJECT_RECLAIMER_ERROR_DELAY;
         }
+
+        await clean_md_store_objects(Date.now());
```
jackyalbo, line 47: don't think this belongs to here... why should we run db_cleaner inside object_reclaimer - db_cleaner will run on objects with no regard to the list that was reclaimed (and of course we don't want to db-delete objects that was just deleted the DB!). I think what we wanted here is a new functionality that for each of the objects will check how many deleted objects we have with the same key and if it's more than X - to delete the older copies - @dannyzaken to keep me honest here...

jackyalbo (changes requested):

src/server/bg_services/db_cleaner.js
vh05: Yes. looking to clean only deleted md objects and other calls in the function not needed. Divided the function into 3 sub functions.

src/server/bg_services/db_cleaner.js
vh05: What would be the good number ?

src/server/bg_services/objects_reclaimer.js
vh05: The jira issue (https://issues.redhat.com/browse/DFBUGS-1339) is specifically mentioned that we can call clean the deleted objects from the md store in object reclaimer. Thought this is the place. @dannyzaken Please correct me here

dannyzaken: @vh05 I didn't intend for the description in Jira to be specific instructions on how to fix it. I wanted to provide some general context. 
I don't see a reason to call from object_reclaimer to the db_cleaner. These are two separate bg_workers.

dannyzaken:
@vh05. I think this PR does not do what it is supposed to do. We want to keep up to `SOME_CONSTANT` deleted object_md **per key**, not keep `SOME_CONSTANT` of total deleted object_mds. 
This is mainly to handle the example of overwriting an object. Let's say you upload an object with the key `foo` every few seconds. You will have many rows with the same key in a short time (all except one should be deleted). We want to limit the number of these rows (I think 100 is a reasonable limit to start with).

In your implementation, you delete any object_md that is deleted and only keep 100 in total. This is a bit too aggressive, in my opinion. 

Another thing to consider is that you probably want to ignore objectmds that are not marked as `reclaimed`.

I strongly suggest creating a similar case on a local deployment. You can easily produce a dataset with >100 overwrites of the same key, so you can test your code.

dannyzaken:
You will probably need to implement new functions in md_store. 
you can take this function as an example

https://github.com/noobaa/noobaa-core/blob/f32827bba83a2ecfd403fbfb83cd81e50fbe0c2c/src/server/object_services/md_store.js#L805-L826

It is probably not sufficient, and you need to aggregate more data than this (for example, you will need to keep track of the _ids of objects to delete\keep). 
Notice that the md_store is written in Mongo query language and is translated to Postgres JSONB queries in the postgres_client. 
https://github.com/noobaa/noobaa-core/blob/f32827bba83a2ecfd403fbfb83cd81e50fbe0c2c/src/util/postgres_client.js#L1180-L1200

This translation is far from perfect, and we should verify that the generated queries work as expected and perform reasonably well.

src/server/bg_services/db_cleaner.js
nimrod-becker: I think number is arbitrary, but we can go with 10 or 20 for example

vh05:
> > 1. If we delete the objects that are already deleted (keeping only reclaimed), would that harm us or there is any reason we keep the deleted objects in md_store ?
> 
> We perform soft delete (mark the `deleted` timestamp) since it can help debug or fix issues with deleted objects. the data is still in the DB and not gone forever.
> 
> > 2. If the deleted objects are unusually high on regular basis, there is high chance of overwrite on a particular key and if we delete them after certain limit, actually we are deleting the deleted objects as result of overwrite. Is n't it ?
> 
> I'm not sure I understand your point. We don't want to remove deleted rows arbitrarily after reaching some limit. We specifically want to handle this overwrite use case, where a single object is frequently overwritten. for the general cleanup of deleted rows, the db_cleaner deletes them by date, which is good enough for now and we can tweak if necessary. this is not the scope of this bug.
> 
> > 3. As you said, deleting object mds of a particular key requires tracking the bucket and object there on. Is it not too costly operation ?
> 
> Of course we need to profile it once we have the code ready. We can run EXPLAIN queries once we have a working query, and analyze the performance.

Sure Danny. That clarifies my queries.

vh05:
@dannyzaken 

Please correct my approaches

**Approach 1:**
1. Count objects per bucket (including deleted). This will give total objects incl deleted object of each bucket
2. Count objects of bucket (active object).
3. If Count from point 1 are more than point 2 + 100, then we can walk inside the bucket and `Find deleted objects based on key`, if a key has deleted objects > 0, then clean them from md_store.

**Approach 2**
We can keep the overwrite information @ bucket level (Every time it is overwritten, the bucket info can be updated). If the object is overwritten, Key of the object and overwrite count should be added and the buckets are checked for this and deleted objects can be deleted if there are any overwritten objects.

**Approach 3**
We can have a table of the deleted objects. Everytime when we overwrite, we add the entries to this table. This table is referred to clean the md_store on regular basis if the count goes above certain number

PR #8708, opened 2025-01-21
bucket notifications - events for scale

alphaprinz:
### Explain the changes
1. Adds events to scenarios the scale wishes to monitor.
2. New event for a failed notification.
3. New event for low space in monitoring logs FS.

Free space below threshold sample:
Jan-26 10:56:06.747 [nsfs/573237] [EVENT]{"timestamp":"2025-01-26T18:56:06.746Z","host":"li-56a4e0cc-353c-11b2-a85c-9bc45442ed4b.ibm.com","event":{"code":"noobaa_notification_low_space","message":"Pending notification log dir low on space","description":"Low space","entity_type":"NODE","event_type":"WARN","scope":"NODE","severity":"WARN","state":"HEALTHY","arguments":{"fs_stat":{"type":1481003842,"bsize":4096,"blocks":483087906,"bfree":411722363,"bavail":411722363,"files":193329536,"ffree":185506484}},"pid":573237}}

Failed notification sample:
Jan-26 11:05:28.657 [/576243] [EVENT]{"timestamp":"2025-01-26T19:05:28.657Z","host":"li-56a4e0cc-353c-11b2-a85c-9bc45442ed4b.ibm.com","event":{"code":"noobaa_notification_failed","message":"Failed to send notification., value: first","description":"Notification failed., error: Error: No such file or directory","entity_type":"NODE","event_type":"WARN","scope":"NODE","severity":"WARN","state":"HEALTHY","arguments":{"code":"ENOENT","context":"Readfile _path=/etc/noobaa.conf.d/connections/connect_http.json "},"pid":576243}}

1. -

### Testing Instructions:
1. Achieve or simulate low FS space. Alternatively, set a high threshold for event (eg NOTIFICATION_SPACE_CHECK_THRESHOLD = 0.99)
2. Set NOTIFICATION_REQ_PER_SPACE_CHECK to a low value, but more than 0 (eg 1).
3. Setup notification, send a notification
5. Event should appear in logs.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/docs/NooBaaNonContainerized/ConfigFileCustomizations.md
+++ b/docs/NooBaaNonContainerized/ConfigFileCustomizations.md
@@ -478,6 +478,34 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
     3. systemctl restart noobaa
     ```
 
+### 33. Notification space monitor frequency flag -
+* <u>Key</u>: `NOTIFICATION_REQ_PER_SPACE_CHECK`
+* <u>Type</u>: Positive integer
+* <u>Default</u>: 0
+* <u>Description</u>: Free space check will run per this many notifications (per node). 0 to disable.
+* <u>Steps</u>:
+    ```
+    1. Open /path/to/config_dir/config.json file.
+    2. Set the config key to run free space check each 1000 notifications -
+    Example:
+    "NOTIFICATION_REQ_PER_SPACE_CHECK": 1000
+    3. systemctl restart noobaa
+    ```
+
+### 34. Notification space monitor threshold flag -
+* <u>Key</u>: `NOTIFICATION_SPACE_CHECK_THRESHOLD`
+* <u>Type</u>: Number
+* <u>Default</u>: 0.1
+* <u>Description</u>: Fraction (more than zero, less than 1) of availble block, below which event is logged.
+* <u>Steps</u>:
+    ```
+    1. Open /path/to/config_dir/config.json file.
+    2. Set the config key -
+    Example:
+    "NOTIFICATION_SPACE_CHECK_THRESHOLD": 0.1
+    3. systemctl restart noobaa
+    ```
+
 ## Config.json File Examples
 The following is an example of a config.json file - 
 
--- a/src/manage_nsfs/manage_nsfs_events_utils.js
+++ b/src/manage_nsfs/manage_nsfs_events_utils.js
@@ -384,4 +384,30 @@ NoobaaEvent.CONFIG_DIR_UPGRADE_FAILED = Object.freeze({
     state: 'DEGRADED',
 });
 
+///////////////////////////////
+//   NOTIFICATION EVENTS     //
+///////////////////////////////
+
+NoobaaEvent.NOTIFICATION_LOW_SPACE = Object.freeze({
+    event_code: 'noobaa_notification_low_space',
+    message: 'Pending notification log dir low on space',
+    description: 'Low space',
+    entity_type: 'NODE',
+    event_type: 'WARN',
+    scope: 'NODE',
+    severity: 'WARN',
+    state: 'HEALTHY',
+});
+
+NoobaaEvent.NOTIFICATION_FAILED = Object.freeze({
+    event_code: 'noobaa_notification_failed',
+    message: 'Failed to send notification.',
+    description: 'Notification failed.',
+    entity_type: 'NODE',
+    event_type: 'WARN',
+    scope: 'NODE',
+    severity: 'WARN',
+    state: 'HEALTHY',
+});
+
 exports.NoobaaEvent = NoobaaEvent;
--- a/src/util/notifications_util.js
+++ b/src/util/notifications_util.js
@@ -14,6 +14,7 @@ const { get_process_fs_context } = require('./native_fs_utils');
 const nb_native = require('../util/nb_native');
 const http_utils = require('../util/http_utils');
 const nc_mkm = require('../manage_nsfs/nc_master_key_manager').get_instance();
+const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 
 const OP_TO_EVENT = Object.freeze({
     put_object: { name: 'ObjectCreated' },
@@ -141,9 +142,7 @@ class Notificator {
             } catch (err) {
                 dbg.error("Failed to notify. err = ", err, ", str =", str);
                 //re-write the failed notification if it's still configured on the bucket
-                if (notif) {
-                    this.handle_failed_notification(notif, failure_append);
-                }
+                this.handle_failed_notification(notif, failure_append, err);
             }
         });
         //note we can't reject promises here, since Promise.all() is rejected on
@@ -155,22 +154,28 @@ class Notificator {
         return true;
     }
 
-    async handle_failed_notification(notif, failure_append) {
-        let bucket;
-        //re-write the failed notification in the persitent log, unless
-        //it is no longer configured on the bucket
+    async handle_failed_notification(notif, failure_append, err) {
         if (this.nc_config_fs) {
-            bucket = await this.nc_config_fs.get_bucket_by_name(notif.meta.bucket);
-        } else {
-            const system = this.system_store.data.systems[0];
-            bucket = system.buckets_by_name && system.buckets_by_name[notif.meta.bucket];
+            new NoobaaEvent(NoobaaEvent.NOTIFICATION_FAILED).create_event(notif?.meta?.name, err, err.toString());
         }
-        if (bucket.notifications) {
-            for (const notif_conf of bucket.notifications) {
-                if (notif_conf.id[0] === notif.meta.name) {
-                    //notification is still configured, rewrite it
-                    await failure_append(JSON.stringify(notif));
-                    break;
+
+        if (notif) {
+            let bucket;
+            //re-write the failed notification in the persitent log, unless
+            //it is no longer configured on the bucket
+            if (this.nc_config_fs) {
+                bucket = await this.nc_config_fs.get_bucket_by_name(notif.meta.bucket);
+            } else {
+                const system = this.system_store.data.systems[0];
+                bucket = system.buckets_by_name && system.buckets_by_name[notif.meta.bucket];
+            }
+            if (bucket.notifications) {
+                for (const notif_conf of bucket.notifications) {
+                    if (notif_conf.id[0] === notif.meta.name) {
+                        //notification is still configured, rewrite it
+                        await failure_append(JSON.stringify(notif));
+                        break;
+                    }
                 }
             }
         }
@@ -224,7 +229,7 @@ class HttpNotificator {
                     return;
                 }
                 dbg.error("Notify err =", err);
-                promise_failure_cb(notif, failure_ctxt).then(resolve);
+                promise_failure_cb(notif, failure_ctxt, err).then(resolve);
             });
             req.on('timeout', () => {
                 dbg.error("Notify timeout");
@@ -281,7 +286,7 @@ class KafkaNotificator {
                 Date.now(),
                 (err, offset) => {
                     if (err) {
-                        promise_failure_cb(notif, failure_ctxt).then(resolve);
+                        promise_failure_cb(notif, failure_ctxt, err).then(resolve);
                     } else {
                         resolve();
                     }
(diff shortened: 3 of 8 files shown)


docs/NooBaaNonContainerized/ConfigFileCustomizations.md
```diff
@@ -478,6 +478,34 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
...
+* <u>Key</u>: `NOTIFICATION_REQ_PER_SPACE_CHECK`
+* <u>Type</u>: Number
+* <u>Default</u>: 0
+* <u>Description</u>: Free space check will run per this many notifications. 0 to disable.
+* <u>Steps</u>:
+    ```
+    1. Open /path/to/config_dir/config.json file.
+    2. Set the config key -
+    Example:
+    "NOTIFICATION_REQ_PER_SPACE_CHECK": 1000
```
shirady, line 491: What does this 1000 mean? Could you write it explicitly in the example description?

docs/NooBaaNonContainerized/ConfigFileCustomizations.md
```diff
@@ -478,6 +478,34 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
...
+    3. systemctl restart noobaa
```
shirady, line 492: Are you sure you need to restart the service for this?

src/util/notifications_util.js
```diff
@@ -494,10 +499,31 @@ function get_notification_logger(locking, namespace, poll_interval) {
...
+function check_free_space(req) {
+    if (!req.object_sdk.nsfs_config_root || !config.NOTIFICATION_REQ_PER_SPACE_CHECK) {
+        return;
+    }
+    req.notification_logger.writes_counter += 1;
+
+    if (req.notification_logger.writes_counter > config.NOTIFICATION_REQ_PER_SPACE_CHECK) {
+        req.notification_logger.writes_counter = 0;
+        const fs_stat = fs.statfsSync(config.NOTIFICATION_LOG_DIR);
+        if (fs_stat.bavail / fs_stat.blocks < config.NOTIFICATION_SPACE_CHECK_THRESHOLD) {
```
shirady, line 523: Why at start do you init the `writes_counter` to 0?
Could you explain in a comment the logic for the calc?

src/util/notifications_util.js
```diff
@@ -145,6 +147,9 @@ class Notificator {
                 }
             } catch (err) {
                 dbg.error("Failed to notify. err = ", err, ", str =", str);
+                if (this.nc_config_fs) {
+                    new NoobaaEvent(NoobaaEvent.NOTIFICATION_FAILED).create_event(null, {name: notif?.meta?.name}, err);
```
shirady, line 151: I saw in other implementations that they used `undefined` instead of `null`.
```suggestion
                    new NoobaaEvent(NoobaaEvent.NOTIFICATION_FAILED).create_event(undefined, {name: notif?.meta?.name}, err);
```

In case the name is `undefined` (`notif?.meta?.name`) would you want to pass additional arguments?

src/util/notifications_util.js
```diff
@@ -494,10 +499,31 @@ function get_notification_logger(locking, namespace, poll_interval) {
...
+    if (!req.object_sdk.nsfs_config_root || !config.NOTIFICATION_REQ_PER_SPACE_CHECK) {
+        return;
+    }
+    req.notification_logger.writes_counter += 1;
+
+    if (req.notification_logger.writes_counter > config.NOTIFICATION_REQ_PER_SPACE_CHECK) {
+        req.notification_logger.writes_counter = 0;
+        const fs_stat = fs.statfsSync(config.NOTIFICATION_LOG_DIR);
+        if (fs_stat.bavail / fs_stat.blocks < config.NOTIFICATION_SPACE_CHECK_THRESHOLD) {
+            new NoobaaEvent(NoobaaEvent.NOTIFICATION_LOW_SPACE).create_event(null, {fs_stat});
```
shirady, line 524: I saw in other implementations that they used `undefined` instead of `null`.
```suggestion
            new NoobaaEvent(NoobaaEvent.NOTIFICATION_LOW_SPACE).create_event(undefined, {fs_stat});
```
Do you need all the stat information?

docs/NooBaaNonContainerized/ConfigFileCustomizations.md
```diff
@@ -478,6 +478,34 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
     3. systemctl restart noobaa
     ```
 
+### 33. Notification space monitor frequency flag -
+* <u>Key</u>: `NOTIFICATION_REQ_PER_SPACE_CHECK`
+* <u>Type</u>: Number
```
shirady, line 483: Any limitation on this number? (only positive, not higher than...) you can write in the description.

docs/NooBaaNonContainerized/ConfigFileCustomizations.md
```diff
@@ -478,6 +478,34 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
...
+    2. Set the config key -
+    Example:
+    "NOTIFICATION_REQ_PER_SPACE_CHECK": 1000
+    3. systemctl restart noobaa
+    ```
+
+### 34. Notification space monitor threshold flag -
+* <u>Key</u>: `NOTIFICATION_SPACE_CHECK_THRESHOLD`
+* <u>Type</u>: Number
+* <u>Default</u>: 0.1
```
shirady, line 498: Possible values that can be configured?

src/util/notifications_util.js
```diff
@@ -115,9 +116,10 @@ class Notificator {
     async _notify(fs_context, log_file, failure_append) {
         const file = new LogFile(fs_context, log_file);
         let send_promises = [];
+        let notif;
         await file.collect_and_process(async str => {
             try {
-                const notif = JSON.parse(str);
```
nadavMiz, line 120: why did you change this? where do you write to this value?

src/manage_nsfs/manage_nsfs_events_utils.js
```diff
@@ -384,4 +384,30 @@ NoobaaEvent.CONFIG_DIR_UPGRADE_FAILED = Object.freeze({
...
+    message: 'Pending notification log dir low on space',
+    description: 'Low space',
+    entity_type: 'NODE',
+    event_type: 'WARN',
+    scope: 'NODE',
+    severity: 'WARN',
+    state: 'HEALTHY',
+});
+
+NoobaaEvent.NOTIFICATION_FAILED = Object.freeze({
```
nadavMiz, line 402: would this be relevant when other notifications will fail as well?

src/util/notifications_util.js
aspandey, line 524: Okay, this we need to think...

Once we reach the space limit where we have to raise an event, means we are in danger of running out of space?
If so, this write onward, we should raise this event after every write and not after "NOTIFICATION_REQ_PER_SPACE_CHECK"

so I think we sould reset req.notification_logger.writes_counter = 0; 
if we are okay with the space and 

if (fs_stat.bavail / fs_stat.blocks < config.NOTIFICATION_SPACE_CHECK_THRESHOLD)

is not true.

docs/NooBaaNonContainerized/ConfigFileCustomizations.md
alphaprinz: How else config.js will be reloaded? Is there a way to reload it while endpoint is running?

src/util/notifications_util.js
alphaprinz: writes_counter counts how many writes has been performed since last check.
it starts at zero, and then zeroed on each check.

src/util/notifications_util.js
alphaprinz: It's just for logs (ie, not needed), might as well include it, no?

src/util/notifications_util.js
alphaprinz: The `notif` declaration is outside the `try` block so that is could be used for to get the name for the event in the `catch` block.

src/manage_nsfs/manage_nsfs_events_utils.js
alphaprinz: I'm not sure I understand.
You mean notifications other than bucket event notifications?
If so, maybe? It's possible, other places would use it for different purposes?
I can change the name to BUCKET_NOTIFICATION_FAILED if that would help.

src/util/notifications_util.js
alphaprinz: I'm not sure, sounds excessive logging for a situation that is not necessarily wrong or avoidable, where functionality is not necessarily degrade.
I would guess that if users ignore the warn event writing a lot more warn event will probably won't help.

src/util/notifications_util.js
alphaprinz: I've switched to `underfined`.

I don't think more details are needed. `name` is not available in an unlikely case.

src/manage_nsfs/manage_nsfs_events_utils.js
nadavMiz: you can change it to BUCKET_NOTIFICATION_FAILED if you think it needs to be used only in this context. but maybe we should send this notification in other places. we can also do it in the future on a different PR

docs/NooBaaNonContainerized/ConfigFileCustomizations.md
```diff
@@ -478,6 +478,34 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
     3. systemctl restart noobaa
     ```
 
+### 33. Notification space monitor frequency flag -
+* <u>Key</u>: `NOTIFICATION_REQ_PER_SPACE_CHECK`
+* <u>Type</u>: Positive integer
```
shirady, line 483: I think it was fine "Number" (I looked at existing example like `GPFS_DOWN_DELAY` and `NSFS_DIR_CACHE_MAX_TOTAL_SIZE`), and mention that it is a positive integer.

docs/NooBaaNonContainerized/ConfigFileCustomizations.md
shirady: I'm just asking because there are cases in this configs where it was not mentioned.

docs/NooBaaNonContainerized/ConfigFileCustomizations.md
```diff
@@ -478,6 +478,34 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
...
+    Example:
+    "NOTIFICATION_REQ_PER_SPACE_CHECK": 1000
+    3. systemctl restart noobaa
+    ```
+
+### 34. Notification space monitor threshold flag -
+* <u>Key</u>: `NOTIFICATION_SPACE_CHECK_THRESHOLD`
+* <u>Type</u>: Number
+* <u>Default</u>: 0.1
+* <u>Description</u>: Fraction (more than zero, less than 1) of availble block, below which event is logged.
```
shirady, line 499: Does it include 0 and 1? (from what I understand here no...).

src/util/notifications_util.js
shirady: Whatever you think.
Could you please attach to the PR description the printings of the new events (so we can see a full example).

docs/NooBaaNonContainerized/ConfigFileCustomizations.md
shirady, line 1: Since you add configuration - I think you should update the NSFS configuration schema and we also have tests for this schema changes, see:
`nsfs_config_schema.js` file
https://github.com/noobaa/noobaa-core/blob/97f8b75e6dd786ae4242bdc0d43c26ebe0d03dd3/src/server/system_services/schemas/nsfs_config_schema.js#L4-L11
`test_nc_nsfs_config_schema_validation.test.js` file, for example:
https://github.com/noobaa/noobaa-core/blob/97f8b75e6dd786ae4242bdc0d43c26ebe0d03dd3/src/test/unit_tests/jest_tests/test_nc_nsfs_config_schema_validation.test.js#L52-L60

docs/NooBaaNonContainerized/ConfigFileCustomizations.md
alphaprinz: No, range is open.

shirady:
@alphaprinz looking at the examples in the PR description:
1. In the first example I was expecting to see "Free space below threshold sample" but in the message I see "Noobaa started", so we probably need to paste another event here.
2. In the second event "Failed notification" in the message I see "description":"Notification failed., error: Error: No such file or directory", do you think it is enough? there is no information about the file name.

alphaprinz:
@shirady 
1. I have fixed the sample.
2. I changed the notification failure event to include the err.toString() with more details (sample was also updated).

src/util/notifications_util.js
```diff
@@ -514,10 +519,36 @@ function get_notification_logger(locking, namespace, poll_interval) {
...
+        //free space check is disabled. nothing to do.
+        return;
+    }
+    req.notification_logger.writes_counter += 1;
+
+    //is it time to check?
+    if (req.notification_logger.writes_counter > config.NOTIFICATION_REQ_PER_SPACE_CHECK) {
+        //yes. remember we've just ran  the check by zero-ing the counter.
+        req.notification_logger.writes_counter = 0;
+        const fs_stat = fs.statfsSync(config.NOTIFICATION_LOG_DIR);
```
shirady, line 545: I know that in NSFS we use: `await nb_native().fs.stat` to get the stat.
Why did you choose to use stat from fs and in its sync way?
I think it's worth adding a comment about it.

src/util/notifications_util.js
```diff
@@ -514,10 +519,36 @@ function get_notification_logger(locking, namespace, poll_interval) {
...
+        //is the ratio of available blocks less than the configures threshold?
+        if (fs_stat.bavail / fs_stat.blocks < config.NOTIFICATION_SPACE_CHECK_THRESHOLD) {
```
shirady, line 547: Why did you choose to use `bavail` (Free blocks available to unprivileged users) and not `bfree` (Free blocks in file system) ([nodejs documentation](https://nodejs.org/api/fs.html#statfsbavail))?
I think it's worth adding a comment about it (you can expand the existing comment).

shirady (approved):
LGTM

PR #8709, opened 2025-01-22
bucket notification - ignore failure if notification was removed

alphaprinz:
### Explain the changes
1. Before this PR, failed notifications were always re-written to persistent log.
Now they are re-written only if they are still configured on the bucket.

1. Remove unwanted notification from persistent logs.

### Testing Instructions:
1. Configure notification to an unavailable external connection.
2. Upload a file to the bucket.
3. Remove notification configuration from bucket.
4. Run notifications (if in NC).
5. Notification fails, but is not re-written to persistent logs.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/notifications_util.js
+++ b/src/util/notifications_util.js
@@ -44,6 +44,12 @@ class Notificator {
         this.connect_files_dir = connect_files_dir ?? DEFAULT_CONNECT_FILES_DIR;
         this.nc_config_fs = nc_config_fs;
         this.batch_size = batch_size || config.NOTIFICATION_BATCH || 10;
+
+        //requiring system_store takes about 2 seconds for the first time
+        //this time is unnecesarily added to manage_nsfs runtime, which fails nsfs tests on timeout
+        //so I'm trying to reduce places were system store is required to minimum
+        this.system_store = require('../server/system_services/system_store').get_instance();
+        this.system_utils = require('../server/utils/system_utils');
     }
 
     async run_batch() {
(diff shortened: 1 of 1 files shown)


src/util/notifications_util.js
```diff
@@ -127,25 +129,21 @@ class Notificator {
...
                 if (send_promise) send_promises.push(send_promise);
                 if (send_promises.length > this.batch_size) {
                     await Promise.all(send_promises);
                     send_promises = [];
                 }
             } catch (err) {
                 dbg.error("Failed to notify. err = ", err, ", str =", str);
-                await failure_append(str);
+                //re-write the failed notification if it's still configured on the bucket
+                if (notif) {
```
aspandey, line 144: No need of this check. 
"notif" will not be null at this point. If it could be, then we should have checked it before line no 124.

src/util/notifications_util.js
```diff
@@ -157,6 +155,27 @@ class Notificator {
...
     }
 
+    async handle_failed_notification(notif, failure_append) {
+        let bucket;
+        //re-write the failed notification in the persitent log, unless
+        //it is no longer configured on the bucket
+        if (this.nc_config_fs) {
+            bucket = await this.nc_config_fs.get_bucket_by_name(notif.meta.bucket);
+        } else {
+            const system = this.system_store.data.systems[0];
```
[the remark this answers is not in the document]
alphaprinz: If system is `null`, then _can_run() will return false and bg worker will not run (and we will not get to handle_failed_notification() where I don't check is system is not null).

src/util/notifications_util.js
alphaprinz: notif can be null in case json parsing fails (currently line 123).
If json parsing is ok, notif will not be null on 124 and forward.

aspandey (approved):

PR #8714, opened 2025-01-22
nope-ip removal - Phase 1

liranmauda:
nope-ip removal - Phase 1

# `net_utils`
- Will no longer use the deprecated `url.parse()` and use `new URL()` instead. It is possible because we use it with .hostname, which is equal for both.
- Create ip_toLong to replace the node-ip toLong

# Others
- Replace the use of isV4Format to net.isIPv4
- Replace the use of isV6Format to net.isIPv6
- Calling ip_toLong instead of toLong

### Explain the changes
1. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/common_services/auth_server.js
+++ b/src/server/common_services/auth_server.js
@@ -3,7 +3,6 @@
 
 const _ = require('lodash');
 const bcrypt = require('bcrypt');
-const ip_module = require('ip');
 
 const P = require('../../util/promise');
 const dbg = require('../../util/debug_module')(__filename);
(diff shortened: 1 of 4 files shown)


src/server/common_services/auth_server.js
```diff
@@ -578,10 +577,10 @@ function _prepare_auth_request(req) {
             const client_ip = net_utils.unwrap_ipv6(req.auth.client_ip);
             if (client_ip) {
                 let is_allowed = false;
-                const client_ip_val = ip_module.toLong(client_ip);
+                const client_ip_val = net_utils.ip_toLong(client_ip);
```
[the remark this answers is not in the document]
liranmauda: @jackyalbo I have looked at it but it is completely out of scope. 
We should have a run on looking at deprecated fields in the schema and remove them, including an upgrade script to verify they are no longer there.

src/util/http_utils.js
```diff
@@ -3,7 +3,8 @@
 /* eslint-disable no-control-regex */
 
 const _ = require('lodash');
-const ip = require('ip');
+const ip_module = require('ip');
```
jackyalbo, line 6: why are we just changing the name here? consistency? maybe explain here that this change is still in progress, and we plan to remove this module. 
Is the effort of totally removing this module that big? the isEqual sounds easy also, CIDR doesn't suppose to be hard if we have the ip_to_long conversion

liranmauda: This is Phase 1, as the title suggests.
The change of this name is to make it easy for us to look for `ip_module.` 
in later PR we will address the other functions, CIDR is not that easy...

jackyalbo: the question is if we need 2 phases :). But sure, if you prefer to split.

jackyalbo (approved):

PR #8715, opened 2025-01-22
Fix key and bucket length checks

tangledbytes:
### Explain the changes
This PR fixes the key and bucket length checks by using the byte length instead of UTF8 character length.

1. An issue found by DBS3 folks

### Testing Instructions:

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -785,6 +785,24 @@ function key_marker_to_cont_tok(key_marker, objects_arr, is_truncated) {
     return Buffer.from(j).toString('base64');
 }
 
+/**
+ * Returns true if the byte length of the key
+ * is within the range [0, max_length]
+ * @param {string} key 
+ * @param {number} max_length 
+ * @returns 
+ */
+function verify_string_byte_length(key, max_length) {
+    // Fast path
+    const MAX_UTF8_WIDTH = 4;
+    if (key.length * MAX_UTF8_WIDTH <= max_length) {
+        return true;
+    }
+
+    // Slow path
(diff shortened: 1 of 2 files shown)


src/endpoint/s3/s3_utils.js
```diff
@@ -785,6 +785,30 @@ function key_marker_to_cont_tok(key_marker, objects_arr, is_truncated) {
...
+    }
+
+    // Fast path
+    const MAX_UTF8_WIDTH = 4;
+    if (key.length * MAX_UTF8_WIDTH <= max_length) {
+        return true;
+    }
+
+    // Slow path
+    return Buffer.from(key, 'utf8').byteLength <= max_length;
```
guymguym, line 809: since we don't need the actual buffer then calling Buffer.byteLength() should be better here

src/endpoint/s3/s3_utils.js
```diff
@@ -785,6 +785,30 @@ function key_marker_to_cont_tok(key_marker, objects_arr, is_truncated) {
...
+ * Returns true if the byte length of the key
+ * is within the range [0, max_length]
+ * @param {string} key 
+ * @param {number} max_length 
+ * @returns 
+ */
+function verify_string_byte_length(key, max_length) {
+    // If max_length is infintiy then no need to do
+    // any further checks
+    if (!Number.isFinite(max_length)) {
```
guymguym, line 798: This check is covered by the next check of len*4<=max, so I would remove this one just to reduce redundant logic (which is what the Infinity value was meant for)

guymguym (approved):

PR #8722, opened 2025-01-24
NC | CLI | Added a validation to check if whitespace is not trimmed from string value for a flag

### LINKED ISSUE #8319: NC | CLI | Weird behavior of the CLI when adding space before access_key/secret_key flag
### Environment info
- NooBaa Version: Master
- Platform: NC

### Actual behavior
1. On account creation/update - when providing `--access_key` flag and `--secret_key` and there is a space before the value, Instead of failing or trimming the space at the beginning, the access key and secret key get generated automatically.

### Expected behavior
1. Failing or trimming the space at the beginning and create an account with the user's access_key and secret_key.

### Steps to reproduce
1. Run `sudo node src/cmd/manage_nsfs.js account add --name account15 --user root --new_buckets_path=/private/tmp/dir2/ --access_key= 12345678912345678912 --secret_key= a234567891234567891212345678912345678912  --debug 5`

```bash
sudo node src/cmd/manage_nsfs.js account add --name account15 --user root --new_buckets_path=/private/tmp/dir2/ --access_key= 12345678912345678912 --secret_key= a234567891234567891212345678912345678912  --debug 5

# Output - 
{
  "response": {
    "code": "AccountCreated",
    "reply": {
      "_id": "66d04174c03e0b51480e4ddd",
      "name": "account15",
      "email": "account15",
      "creation_date": "2024-08-29T09:37:56.615Z",
      "access_keys": [
        {
          "access_key": "QBxpvZvehMsbptZmlauR",
          "secret_key": "axXfXsBzpW2VeJ/2VibpioZcC+TLVSpMcTRyivC8"
        }
      ],
      "nsfs_account_config": {
        "distinguished_name": "root",
        "new_buckets_path": "/private/tmp/dir2/"
      },
      "allow_bucket_creation": true
    }
  }
}
```

aayushchouhan09:
### Explain the changes
1. Problem:  Leading spaces before a flag value, when using `=`, caused the value to be split into an empty string ('') and the actual value, resulting in faulty flag assignments.
2. Solution: A validation has been added to check if the flag value become empty due to incorrect formatting. It checks if `argv._` has more than two elements and throws an error if invalid values are detected.

This prevents incorrect assignments and ensures flag values are valid.

1. Fixed: #8319 

### Testing Instructions:
1. Try creating account with noobaa-cli using below command:
`$ sudo node src/cmd/manage_nsfs account add --name aayush2 --uid=9 --gid=9 --new_buckets_path= tmp/` 

- [X] Tests added

### DIFF
--- a/src/manage_nsfs/manage_nsfs_validations.js
+++ b/src/manage_nsfs/manage_nsfs_validations.js
@@ -29,6 +29,7 @@ const notifications_util = require('../util/notifications_util');
  * @param {object} argv
  */
 async function validate_input_types(type, action, argv) {
+    validate_no_extra_args(argv._);
     validate_type_and_action(type, action);
     // when we use validate_no_extra_options we don't care about the value, only the flags
     const input_options = Object.keys(argv);
@@ -82,6 +83,35 @@ function validate_type_and_action(type, action) {
         if (!Object.values(DIAGNOSE_ACTIONS).includes(action)) throw_cli_error(ManageCLIError.InvalidDiagnoseAction);
     } else if (type === TYPES.UPGRADE) {
         if (!Object.values(UPGRADE_ACTIONS).includes(action)) throw_cli_error(ManageCLIError.InvalidUpgradeAction);
+    }
+}
+
+/**
+ * validate_no_extra_args ensures that the parsed arguments contain only the expected type and action
+ * if `argv_` contains more than two elements (type and action), it indicate extra argument which may be due to:
+ *  - Leading spaces in flag values (e.g., `--new_buckets_path= abc/`)
+ *  - Additional unexpected arguments passed from the command line
+ *
+ * ### CLI Parsing Behavior:
+ * When passing arguments via CLI, if a flag's value starts with a space or consists only of whitespace,
+ * the shell may treat it differently:
+ *  - Leading spaces - might be preserved when quoted (`--new_buckets_path=" abc/"`).
+ *  - Unquoted values with spaces - may cause incorrect parsing (`--new_buckets_path= abc/` is interpreted as an empty value).
+ *  - The parsed argument structure typically follows:
+ *      ```json
+ *      { _: [ "command", "action" ], flag1: "value1", flag2: "value2", ... }
+ *      ```
+ *  - The `_` array should contain only the command and action. If additional elements appear,
+ *    it suggests an issue with flag values (e.g., an incorrectly formatted or empty value).
+ *
+ * @param {string[]} argv_
+ * @throws {ManageCLIError}
+ */
+function validate_no_extra_args(argv_) {
+    // checking if 'argv_' contains more then 2 values (i.e, type and action)
+    if (argv_.length > 2) {
+        const details = `unexpected extra arguments detected: ${JSON.stringify(argv_.slice(2))}, ensure flag values are correctly formatted.`;
+        throw_cli_error(ManageCLIError.InvalidArgument, details);
     }
 }
 
@@ -718,3 +748,4 @@ exports.validate_whitelist_arg = validate_whitelist_arg;
 exports.validate_whitelist_ips = validate_whitelist_ips;
 exports.validate_flags_combination = validate_flags_combination;
 exports.validate_connection_args = validate_connection_args;
+exports.validate_no_extra_args = validate_no_extra_args;
--- a/src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
+++ b/src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
@@ -23,6 +23,12 @@ const timeout = 50000;
 const config = require('../../../../config');
 const config_fs_account_options = { show_secrets: true, decrypt_secret_key: true };
 
+const quoted_type = Object.freeze({
+    SPACE_ONLY: "s", // space only
+    QUOTE_SPACE: "qs", // quote then space
+    SPACE_QUOTE: "sq", // space then quote
+});
+
 // eslint-disable-next-line max-lines-per-function
 describe('manage nsfs cli account flow', () => {
     describe('cli create account', () => {
@@ -101,6 +107,33 @@ describe('manage nsfs cli account flow', () => {
             assert_account(account, account_options, true);
             const account_symlink = await config_fs.get_account_by_access_key(access_key, config_fs_account_options);
             assert_account(account_symlink, account_options);
+        });
+
+        it('should fail - cli create account invalid new_buckets_path (leading space in new_buckets_path)', async () => {
+            const { type, name, uid, gid, new_buckets_path } = defaults;
+            const account_options = { config_root, name, uid, gid}; // will add new_buckets_path with leading space later
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options);
+            const res = await exec_manage_cli_add_leading_space_option(command, 'new_buckets_path', new_buckets_path, quoted_type.SPACE_ONLY);
+            expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidArgument.code);
+        });
+
+        it('should fail - cli create account invalid new_buckets_path (leading space in quoted new_buckets_path)', async () => {
+            const { type, name, uid, gid, new_buckets_path } = defaults;
+            const account_options = { config_root, name, uid, gid}; // will add quoted new_buckets_path with leading space later
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options);
+            const res = await exec_manage_cli_add_leading_space_option(command, 'new_buckets_path', new_buckets_path, quoted_type.SPACE_QUOTE);
+            expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidArgument.code);
+        });
+
+        it('should fail - cli create account invalid new_buckets_path (quoted leading space in new_buckets_path)', async () => {
+            const { type, name, uid, gid, new_buckets_path } = defaults;
+            const account_options = { config_root, name, uid, gid}; // will add new_buckets_path with quoted leading space later
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options);
+            const res = await exec_manage_cli_add_leading_space_option(command, 'new_buckets_path', new_buckets_path, quoted_type.QUOTE_SPACE);
+            expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidAccountNewBucketsPath.code); // should get this error if space is also quoted with the value
         });
 
         it('should fail - cli update account invalid access_key - invalid size', async () => {
@@ -2209,6 +2242,52 @@ async function exec_manage_cli(type, action, options) {
  */
 async function exec_manage_cli_add_empty_option(command, option) {
     const changed_command = command + ` --${option}`;
+    let res;
+    try {
+        res = await os_util.exec(changed_command, { return_stdout: true });
+    } catch (e) {
+        res = e;
+    }
+    return res;
+}
+
+/**
+ * exec_manage_cli_add_leading_space_option modifies the command by appending a flag with a value(quoted or non-quoted) that includes a leading space to test CLI parsing behavior
+ * @param {string} command
+ * @param {string} option
+ * @param {string} value
+ * @param {"s" | "qs" | "sq" | "default"} quoted
+ *
+ * @examples
+ * - await exec_manage_cli_add_leading_space_option("cli_command", "flag", "val", quoted_type.SPACE_ONLY);
+ *   // cli_command --flag= val  (Space before value)
+ *
+ * - await exec_manage_cli_add_leading_space_option("cli_command", "flag", "val", quoted_type.QUOTE_SPACE);
+ *   // cli_command --flag=" val"  (quote then space)
+ *
+ * - await exec_manage_cli_add_leading_space_option("cli_command", "flag", "val", quoted_type.SPACE_QUOTE);
+ *   // cli_command --flag= "val"  (space then quote)
+ *
+ * - await exec_manage_cli_add_leading_space_option("cli_command", "flag", "val");
+ *   // cli_command --flag=val  (No special formatting)
+ */
+async function exec_manage_cli_add_leading_space_option(command, option, value, quoted = "default") {
+    let changed_command;
+
+    switch (quoted) {
+        case quoted_type.SPACE_ONLY:
+            changed_command = `${command} --${option}= ${value}`;
+            break;
+        case quoted_type.QUOTE_SPACE:
+            changed_command = `${command} --${option}=" ${value}"`;
+            break;
+        case quoted_type.SPACE_QUOTE:
+            changed_command = `${command} --${option}= "${value}"`;
+            break;
+        default:
+            changed_command = `${command} --${option}=${value}`;
+    }
+
     let res;
     try {
         res = await os_util.exec(changed_command, { return_stdout: true });


src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -191,6 +191,11 @@ function validate_options_type_by_value(input_options_with_data) {
             const details = `type of flag ${option} should be ${type_of_option} (and the received value is ${value})`;
             throw_cli_error(ManageCLIError.InvalidArgumentType, details);
         }
+        // ensure flag values are not empty or whitespace only strings
+        if (type_of_value === 'string' && value.trim() === '') {
+            const details = `flag ${option} cannot have an empty value or contain only whitespace. Please provide a valid value for ${option}`;
+            throw_cli_error(ManageCLIError.InvalidArgument, details);
+        }
```
shirady, line 198: We use empty string as a way to unset some properties - you can search in the file `manage_nsfs_help_utils.js` "unset" and you will a couple of examples of use.

src/manage_nsfs/manage_nsfs_validations.js
shirady, line 1: Please add a unit test for the case you test, probably will be in the file: `test_nc_nsfs_account_cli.test.js`

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: Yes, you are correct. In that case we can make this condition to specifically check for `access_keys` faulty values as it can't be null if provided.
```suggestion
        // ensure access_keys values are not empty or whitespace only strings
        if (['access_key', 'secret_key'].includes(option) && value.trim() === '') {
            const details = `flag ${option} cannot have an empty value or contain only whitespace. Please provide a valid value for ${option}`;
            throw_cli_error(ManageCLIError.InvalidArgument, details);
        }
```

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: Sure, Thanks

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -191,6 +191,11 @@ function validate_options_type_by_value(input_options_with_data) {
             const details = `type of flag ${option} should be ${type_of_option} (and the received value is ${value})`;
             throw_cli_error(ManageCLIError.InvalidArgumentType, details);
         }
+        // ensure access_keys values are not empty or whitespace only strings
+        if (['access_key', 'secret_key'].includes(option) && value.trim() === '') {
```
nadavMiz, line 195: this function tests the type of the variables. I am not sure if its the best place to put this check. maybe move it to validate_account_identifier. @shirady WDYT

nadavMiz: also does is this case only a problem for access_key, secret_key, or for all string options? if not maybe its better to have a more general check

achouhan09: or we can put it to `validate_account_args` , I found it more suitable for this case.

Currently we are facing this only for access_key and secret_key as this values cannot be unset with ''(empty string). @shirady pls correct me if I am wrong.

shirady: I agree with @nadavMiz comment about moving the implementation from the function `validate_options_type_by_value`.

@achouhan09 from what I see in the issue and know those are the cases, but maybe there are more.
You can move the core implementation to a function, and if we will have more cases we can reuse it.

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -103,6 +103,28 @@ describe('manage nsfs cli account flow', () => {
...
+            expect(JSON.parse(res.stdout).error.message).toBe(ManageCLIError.InvalidArgument.message);
+        });
+
+        it('should fail - cli create account invalid secret_key(leading space in secret_key)', async () => {
+            const { type, name, uid, gid, access_key } = defaults;
+            const account_options = { config_root, name, uid, gid, access_key};
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options)
+            const flag = 'secret_key';
+            const value = ' a234567891234567891212345678912345678912'; // leading space for secret_key
```
shirady, line 123: I think you need to have EXAMPLE suffix for secret_key property (to avoid creating alerts on the published commit).

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: Done, Thanks

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
achouhan09: Could you explain it more?

shirady: I'll send you here an example:
https://github.com/noobaa/noobaa-core/blob/f32827bba83a2ecfd403fbfb83cd81e50fbe0c2c/src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js#L40

I'll send you in Slack the title of the mail you might have got so you can handle it.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -211,6 +211,20 @@ function validate_boolean_string_value(value) {
...
+/**
+ * validate_no_leading_whitespace validates that a given value is not empty or whitespace-only
+ * @param {string} option
+ * @param {string} value
+ * @throws {ManageCLIError}
+ */
+function validate_no_leading_whitespace(option, value) {
+    // ensure values are not empty or whitespace only strings
+    if (value.trim() === '') {
+        const details = `flag ${option} cannot have an empty value or contain only whitespace. Please provide a valid value for ${option}`;
```
shirady, line 223: I think the message might be confusing as the user did not passed an empty value or only whitespace characters (in you example: `--access_key= 12345678912345678ABC --secret_key= a234567891234567891212345678912345678912`

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -574,6 +588,12 @@ function _validate_access_keys(access_key, secret_key) {
         throw_cli_error(ManageCLIError.MissingAccountAccessKeyFlag);
     }
 
+    // checking access_key and secret_key do not contain only whitespace
```
shirady, line 591: I think the comment should match what you check in the function - validate_no_leading_whitespace.

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -103,6 +103,28 @@ describe('manage nsfs cli account flow', () => {
             assert_account(account_symlink, account_options);
         });
 
+        it('should fail - cli create account invalid access_key(leading space in access_key)', async () => {
```
shirady, line 106:
```suggestion
        it('should fail - cli create account invalid access_key (leading space in access_key)', async () => {
```

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -103,6 +103,28 @@ describe('manage nsfs cli account flow', () => {
...
+            const account_options = { config_root, name, uid, gid, secret_key};
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options);
+            const flag = 'access_key';
+            const value = ' 12345678912345678ABC'; // leading space for access_key
+            const res = await exec_manage_cli_add_leading_space_option(command, flag, value);
+            expect(JSON.parse(res.stdout).error.message).toBe(ManageCLIError.InvalidArgument.message);
+        });
+
+        it('should fail - cli create account invalid secret_key(leading space in secret_key)', async () => {
```
shirady, line 117:
```suggestion
        it('should fail - cli create account invalid secret_key (leading space in secret_key)', async () => {
```

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -103,6 +103,28 @@ describe('manage nsfs cli account flow', () => {
...
 
+        it('should fail - cli create account invalid access_key(leading space in access_key)', async () => {
+            const { type, name, uid, gid, secret_key } = defaults;
+            const account_options = { config_root, name, uid, gid, secret_key};
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options);
+            const flag = 'access_key';
+            const value = ' 12345678912345678ABC'; // leading space for access_key
+            const res = await exec_manage_cli_add_leading_space_option(command, flag, value);
+            expect(JSON.parse(res.stdout).error.message).toBe(ManageCLIError.InvalidArgument.message);
```
shirady, line 114: If you want you can check the code is the same and not the message (I know that we have checks of message in the code).
* If you fix, then in both cases.

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
achouhan09: I have updated the logic and we need not to assign `secret_key` value like this any more. Thanks for raising this point.

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: Updated, Thanks

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: sure, thanks

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
achouhan09: fixed, Thanks

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -103,6 +103,24 @@ describe('manage nsfs cli account flow', () => {
             assert_account(account_symlink, account_options);
         });
 
+        it('should fail - cli create account invalid access_key (leading space in access_key)', async () => {
+            const { type, name, uid, gid, access_key, secret_key } = defaults;
+            const account_options = { config_root, name, uid, gid, secret_key};
```
shirady, line 108: I would add a short comment that you didn't pass the `access_key` intentionally.

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
```diff
@@ -103,6 +103,24 @@ describe('manage nsfs cli account flow', () => {
...
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options);
+            const res = await exec_manage_cli_add_leading_space_option(command, 'access_key', access_key);
+            expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidArgument.code);
+        });
+
+        it('should fail - cli create account invalid secret_key (leading space in secret_key)', async () => {
+            const { type, name, uid, gid, access_key, secret_key } = defaults;
+            const account_options = { config_root, name, uid, gid, access_key};
```
shirady, line 117: Same here - I would add a short comment that you didn't pass the `secret_key` intentionally.

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
shirady, line 1: If I understand correctly, those are the cases of "Unquoted values with spaces" (from the function description), can you add the second option cases as well? ("Leading spaces"), for example: `--access_key=" <access-key>"`

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -574,6 +595,12 @@ function _validate_access_keys(access_key, secret_key) {
         throw_cli_error(ManageCLIError.MissingAccountAccessKeyFlag);
     }
 
+    // checking access_key and secret_key do not contain leading space
+    if (access_key !== undefined && secret_key !== undefined) {
```
romayalon, line 599: did you check in which other cases this could happen? I reported on access_key/secret_key but this could happen on other flags as well
we should add this check to all flags in my opinion

romayalon: also, how do we behave on trailing spaces?

src/test/unit_tests/jest_tests/test_nc_nsfs_account_cli.test.js
achouhan09: Sure, We can add this in the same tests.

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: I think for every `string` type of flag we are facing this issue(as the value got splitted in two empty string and actual value), I have verified this for another flag `--user` getting the same issue of leading space. 
But for other types of flags if we put leading space we will get `InvalidArgumentType` error(due to empty string assignment). 
So, we should apply this check on every string type flag.

Trailing space does not create any issue here.

achouhan09: We can put this check in `validate_option_type_by_value`(as before), if we want to apply this check on every string flag.

shirady: If you move it, please remember that there are flags where the empty string is our way to unset them.

achouhan09: We cannot differenciate if the empty string is provided to unset it or is it the leading space.
We need to think of another solution in that case(below 2 in my mind):
1. We are having the list of flags(that can be unset using empty string), we can remove this flags from this validation --> but after this there is no way to check leading space for this flag(that can be unset).
2. We should merge/trim the wrongly parsed flag value(because of leading space).

which approach should we follow? @romayalon @shirady @nadavMiz

romayalon: why not just add another clause to your condition checking if the whole string is not ""?

achouhan09: I am not sure if I got your point but we cannot do that because when we put leading spaces then also empty string is assigned to the flag instead of whitespace. We cannot differenciate because in both case we are getting empty string here.

achouhan09: Updated the code, please check. Thanks

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -211,6 +211,37 @@ function validate_boolean_string_value(value) {
...
+ *  Unquoted values with spaces - may cause incorrect parsing (`--access_key= 123abc` is interpreted as an empty value).
+ *
+ * @param {string[]} argv
+ * @throws {ManageCLIError}
+ */
+function validate_no_leading_whitespace(argv) {
+    for (let i = 0; i < argv.length; i++) {
+        if (!argv[i].startsWith('--') && !argv[i].includes('=')) continue;
+
+        const [option, value] = argv[i].split('=');
```
romayalon, line 230: this is not working correctly when value contains '=' in it

src/cmd/manage_nsfs.js
```diff
@@ -33,6 +33,7 @@ let config_fs;
 
 async function main(argv = minimist(process.argv.slice(2))) {
     try {
+        manage_nsfs_validations.validate_no_leading_whitespace(process.argv.slice(2));
```
romayalon, line 36: when doing it on the arguments we won't handle --from_file option

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -211,6 +211,37 @@ function validate_boolean_string_value(value) {
...
+
+        if (value === undefined) continue;
+
+        const type_of_option = OPTION_TYPE[option.slice(2)];
+
+        if (type_of_option === 'string') {
+            if (value.trim() === '' && i + 1 < argv.length && !argv[i + 1].startsWith('--')) {
```
romayalon, line 237: why not just check if (value.trim() === '' && value !== '' ) ?

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: Can you give me some example for which case it is not working. Actually, we are only evaluating argv containing '=' in it.

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: I tried to print both `process.argv` and `argv` (after slicing and using minimist) for below command:

`sudo node src/cmd/manage_nsfs account add --name aayush2 --uid=9 --gid=9  --access_key= 12345678912345678ABC --secret_key=U2AYaMpU3zRDcRFWmvzgQr9MoHIAsD+3oEXAMPL`

`process.argv:  [ '/home/aayush/.nvm/versions/node/v22.11.0/bin/node', '/home/aayush/github.com/noobaa-core/src/cmd/manage_nsfs', 'account', 'add', '--name', 'aayush2', '--uid=9', '--gid=9', '--access_key=', '12345678912345678ABC', '--secret_key=U2AYaMpU3zRDcRFWmvzgQr9MoHIAsD+3oEXAMPL' ]`

`CONSOLE:: argv:  { _: [ 'account', 'add', '12345678912345678ABC' ], name: 'aayush2', uid: 9, gid: 9, access_key: '', secret_key: 'U2AYaMpU3zRDcRFWmvzgQr9MoHIAsD+3oEXAMPL' }`

Here, `access_key` doesn't contain whitespace it is an empty string after parsing. So, I think above condition would not work. Pls correct me if I am wrong.

src/manage_nsfs/manage_nsfs_validations.js
romayalon: just try to think about any possible value that can have = in it, here is an example - 
```
noobaa-cli account add --name=aayush====1 --uid=9 --gid=9
```
when I print option and value I get the following -
```
[LOG] CONSOLE:: ROMY option: --name
[LOG] CONSOLE:: ROMY value: aayush
[LOG] CONSOLE:: ROMY option: --uid
[LOG] CONSOLE:: ROMY value: 9
[LOG] CONSOLE:: ROMY option: --gid
[LOG] CONSOLE:: ROMY value: 9
```
value of name is incorrect

romayalon: also you can have no = between the option and the value and have it only in the value
for example - 
noobaa-cli account add --name aayush====1 --uid 9 --gid 9

src/manage_nsfs/manage_nsfs_validations.js
romayalon: ok so the root cause is that minimist is parsing this space incorrectly and then we think access key is '' and we auto generate a new access key instead of using the one given us... now the problem is understood better thanks for the details, please add the full explanation of the problem to the PR description
now I understand why you do it directly on process.argv, just notice this is not enough.
why not checking if there is an expected arg in the _ array - _: [ 'account', 'add', '12345678912345678ABC' ]
for example here you have an additional and unexpected property
I think that trying to manipulate process.argv is not that simple, If you really want to do it I would check minimist code

src/cmd/manage_nsfs.js
achouhan09: We don't need to validate arguments we are getting from file as they should be in JSON format already

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -30,6 +30,7 @@ const notifications_util = require('../util/notifications_util');
  */
 async function validate_input_types(type, action, argv) {
     validate_type_and_action(type, action);
+    validate_no_leading_space(argv._);
```
shirady, line 33: Why do you call the new function second and not first?

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -65,6 +66,34 @@ async function validate_input_types(type, action, argv) {
...
+ *      ```
+ *  - The `_` array should contain only the command and action. If additional elements appear,
+ *    it suggests an issue with flag values (e.g., an incorrectly formatted or empty value).
+ *
+ * @param {string[]} argv_
+ * @throws {ManageCLIError}
+ */
+function validate_no_leading_space(argv_) {
+    // checking if 'argv_' contains more then 2 values (i.e, type and action)
+    if (argv_.length > 2) {
```
shirady, line 91: From what I understand from the Slack discussion, checking the size can be improved.
If you need in the file `manage_nsfs_constants.js` you have the list of types and actions:
https://github.com/noobaa/noobaa-core/blob/23aec8dcd195c95b121611f45ee210d280baabb8/src/manage_nsfs/manage_nsfs_constants.js#L4-L40

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -65,6 +66,34 @@ async function validate_input_types(type, action, argv) {
...
+        const details = 'flag value is invalid - leading spaces or an empty value detected, ensure flag values are properly formatted.';
+        throw_cli_error(ManageCLIError.InvalidArgument, details);
```
shirady, line 93: If you'll have a more strict check you will be able to add in the details the exact invalid value.
From you example:
on `sudo node src/cmd/manage_nsfs account add --name aayush3 --uid=9 --gid=9 --allow_bucket_creation=true --new_buckets_path= tmp/`

it is:
`_: [ 'account', 'add', 'tmp/' ]`

And you can explicitly mention the "tmp/" as the invalid value that was passed.

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -103,6 +103,33 @@ describe('manage nsfs cli account flow', () => {
             assert_account(account_symlink, account_options);
         });
 
+        it('should fail - cli create account invalid new_buckets_path (leading space in new_buckets_path)', async () => {
+            const { type, name, uid, gid, new_buckets_path } = defaults;
+            const account_options = { config_root, name, uid, gid}; // will add new_buckets_path with leading space later
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options);
+            const res = await exec_manage_cli_add_leading_space_option(command, 'new_buckets_path', new_buckets_path, 's'); // only space
```
shirady, line 111: Could you make the 's', qs' and 'sq' to appear in uppercase constants (I would add as enum) for readability? (I would avoid the use of short-names as it harder to understand).

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -103,6 +103,33 @@ describe('manage nsfs cli account flow', () => {
...
+            expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidArgument.code);
+        });
+
+        it('should fail - cli create account invalid new_buckets_path (quoted leading space in new_buckets_path)', async () => {
+            const { type, name, uid, gid, new_buckets_path } = defaults;
+            const account_options = { config_root, name, uid, gid}; // will add new_buckets_path with quoted leading space later
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options);
+            const res = await exec_manage_cli_add_leading_space_option(command, 'new_buckets_path', new_buckets_path, 'qs'); // quote then space
+            expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidAccountNewBucketsPath.code);
```
shirady, line 130: The error here is about "InvalidAccountNewBucketsPath", but the added code is about the new error "InvalidArgument", so I find this case confusing.

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -2218,6 +2245,35 @@ async function exec_manage_cli_add_empty_option(command, option) {
     return res;
 }
 
+/**
+ * exec_manage_cli_add_leading_space_option modifies the command by appending a flag with a value(quoted or non-quoted) that includes a leading space to test CLI parsing behavior
+ * @param {string} command
+ * @param {string} option
+ * @param {string} value
+ * @param {string} quoted
```
shirady, line 2253: Could you add examples in the function description for the different cases?

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: Here we want to validate the `time` and `action` args first so that we need not to do that in our new function and we can simply check the length

cc @romayalon

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: I think `validate_type_and_action` function doing the same thing so calling new function after that. wdyt?

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: make sense, will update it. Thanks

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
achouhan09: Sure, I will try to make it more readable.

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
achouhan09: Here, I have added a test just to verify if the space is also quoted with the value then we should get different error.

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
achouhan09: sure, thanks

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -23,6 +23,13 @@ const timeout = 50000;
 const config = require('../../../../config');
 const config_fs_account_options = { show_secrets: true, decrypt_secret_key: true };
 
+const quoted_type = Object.freeze({
+    SPACE_ONLY: "s", // space only
+    QUOTE_SPACE: "qs", // quote then space
+    SPACE_QUOTE: "sq", // space then quote
+    NONE: "none" // none
```
shirady, line 30: Do you need the "NONE"? I think it goes to the defaults.

src/manage_nsfs/manage_nsfs_validations.js
shirady: Sure (I guess you meant `type` and not `time`).
As long as the case that was discussed is covered -` _: [ 'account', 'tmp/' ]` (you can add it as a case).

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -65,6 +66,34 @@ async function validate_input_types(type, action, argv) {
...
+ *  - The `_` array should contain only the command and action. If additional elements appear,
+ *    it suggests an issue with flag values (e.g., an incorrectly formatted or empty value).
+ *
+ * @param {string[]} argv_
+ * @throws {ManageCLIError}
+ */
+function validate_no_leading_space(argv_) {
+    // checking if 'argv_' contains more then 2 values (i.e, type and action)
+    if (argv_.length > 2) {
+        const details = `flag value is invalid - leading spaces detected in: ${JSON.stringify(argv_.slice(2))}, ensure flag values are properly formatted.`;
```
shirady, line 92: Why `JSON.stringify(argv_.slice(2)`? (could you add a comment about it?)
Did you want to ignore the `type` and `action`? I would mention that not every command has `action` - for example, `whitelist`.

src/manage_nsfs/manage_nsfs_validations.js
shirady: I think you're right.

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -103,6 +110,33 @@ describe('manage nsfs cli account flow', () => {
             assert_account(account_symlink, account_options);
         });
 
+        it('should fail - cli create account invalid new_buckets_path (leading space in new_buckets_path)', async () => {
+            const { type, name, uid, gid, new_buckets_path } = defaults;
+            const account_options = { config_root, name, uid, gid}; // will add new_buckets_path with leading space later
+            const action = ACTIONS.ADD;
+            const command = create_command(type, action, account_options);
+            const res = await exec_manage_cli_add_leading_space_option(command, 'new_buckets_path', new_buckets_path, quoted_type.SPACE_ONLY); // only space
```
shirady, line 118: Now, you don't need the comment (it is clear from the enum) - you can leave it or delete it, whatever you think.
```suggestion
            const res = await exec_manage_cli_add_leading_space_option(command, 'new_buckets_path', new_buckets_path, quoted_type.SPACE_ONLY);
```

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
shirady: I would add a comment as it is a bit confusing.

src/manage_nsfs/manage_nsfs_validations.js
romayalon, line 92: checking argv_.length > 2 doesn't necassarly means that it's a leading spaces issue
this is a general issue that we found an unexpected "free" argument, not sure what is the term

for example - 
```
>  noobaa-cli account delete --name romy  blabla
```
[LOG] CONSOLE:: ROMY argv { _: [ 'account', 'delete', 'blabla' ], name: 'romy' }
I would rename the function and explain in the comments that there are several cases of this situation and one of them is leading spaces, and ofc throw a more general error and maybe add " or a leading spaces issues blablalal"

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
achouhan09: Yes we can remove it. Thanks

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
achouhan09: Sure, Thanks.

src/manage_nsfs/manage_nsfs_validations.js
achouhan09: after validating type and action we can remove the first two args - anyway we are printing/returning the error  when args length is more than `2`. So I think the check is good enough.
It would be printed like this below(prints the value only and not the type and action):
`"detail": "flag value is invalid - leading spaces detected in: [\"12345678912345678ABC\"], ensure flag values are properly formatted."`

achouhan09: Make sense, sure I will update it. Thanks

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -2218,6 +2251,52 @@ async function exec_manage_cli_add_empty_option(command, option) {
...
+ * - await exec_manage_cli_add_leading_space_option("cli_command", "flag", "val", quoted_type.QUOTE_SPACE);
+ *   // cli_command --flag=" val"  (quote then space)
+ *
+ * - await exec_manage_cli_add_leading_space_option("cli_command", "flag", "val", quoted_type.SPACE_QUOTE);
+ *   // cli_command --flag= "val"  (space then quote)
+ *
+ * - await exec_manage_cli_add_leading_space_option("cli_command", "flag", "val");
+ *   // cli_command --flag=val  (No special formatting)
+ */
+async function exec_manage_cli_add_leading_space_option(command, option, value, quoted = "default") {
```
shirady, line 2274: I'm not sure you need to add this default value in the function argument
```suggestion
async function exec_manage_cli_add_leading_space_option(command, option, value, quoted) {
```

shirady (approved):
LGTM

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
achouhan09: We need this to make the `quoted` argument optional. Thanks

PR #8723, opened 2025-01-24
NC | CLI | Updated the CLI response to have more info when account/bucket has been deleted

### LINKED ISSUE #8200: NC | NSFS | CLI | Improve Response (To have Details)
### Environment info
- NooBaa Version: 5.17.0
- Platform: NC

### Actual behavior
1. Currently when an account is deleted or the response contains only the code `"code": "AccountDeleted"`.
2. Currently when a bucket is deleted or the response contains only the code `"code": "BucketDeleted"`.

### Expected behavior
1. Add to the response details about the account account name that was deleted.
2. Add to the response details about the bucket name that was deleted.

### Steps to reproduce
1. Use noobaa-cli and create an account and a bucket (look at the output).

This is useful when you create accounts and buckets in a loop and want to easily find the details of the deleted account/bucket.

aayushchouhan09:
### Explain the changes
1.  Improved the response to provide a clearer message about the operations.
2.  Now, the name of the object would also be printed in the response message for more info.

1. Fixed: #8200 

### Testing Instructions:
1. Try creating/deleting(or any other operation) on objects(account, bucket, etc.) using noobaa-cli and check the response returned.

- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -259,7 +259,7 @@ async function delete_bucket(data, force) {
         }
         await native_fs_utils.folder_delete(bucket_temp_dir_path, fs_context_fs_backend, true);
         await config_fs.delete_bucket_config_file(data.name);
-        return { code: ManageCLIResponse.BucketDeleted, detail: '', event_arg: { bucket: data.name } };
+        return { code: ManageCLIResponse.BucketDeleted, detail: { name: data.name }, event_arg: { bucket: data.name } };
     } catch (err) {
         if (err.code === 'ENOENT') throw_cli_error(ManageCLIError.NoSuchBucket, data.name);
         throw err;
@@ -528,7 +528,7 @@ async function update_account(data) {
  */
 async function delete_account(data) {
     await config_fs.delete_account_config_file(data);
-    return { code: ManageCLIResponse.AccountDeleted, detail: '', event_arg: { account: data.name } };
+    return { code: ManageCLIResponse.AccountDeleted, detail: { name: data.name }, event_arg: { account: data.name } };
 }
 
 /**
@@ -791,12 +791,12 @@ async function connection_management(action, user_input) {
             break;
         case ACTIONS.DELETE:
             await config_fs.delete_connection_config_file(user_input.name);
-            response = { code: ManageCLIResponse.ConnectionDeleted };
+            response = { code: ManageCLIResponse.ConnectionDeleted, detail: {name: user_input.name} };
             break;
         case ACTIONS.UPDATE:
             await notifications_util.update_connect_file(user_input.name, user_input.key,
                 user_input.value, user_input.remove_key, config_fs);
-            response = { code: ManageCLIResponse.ConnectionUpdated };
+            response = { code: ManageCLIResponse.ConnectionUpdated, detail: {name: user_input.name} };
             break;
         case ACTIONS.STATUS:
             data = await new notifications_util.Notificator({
--- a/src/manage_nsfs/manage_nsfs_cli_responses.js
+++ b/src/manage_nsfs/manage_nsfs_cli_responses.js
@@ -7,9 +7,10 @@ const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEve
 /**
  * @typedef {{
  *      code?: string, 
+ *      message?: string,
  *      http_code: number,
  *      list?: object,
- *      status?: object,  
+ *      status?: object
  * }} ManageCLIResponseSpec
  */
 
@@ -18,17 +19,19 @@ class ManageCLIResponse {
     /**
      * @param {ManageCLIResponseSpec} response_spec 
      */
-    constructor({ code, status, list }) {
+    constructor({ code, status, list, message }) {
         this.code = code;
         this.http_code = 200;
         this.status = status;
         this.list = list;
+        this.message = message;
     }
 
     to_string(detail) {
         const json = {
             response: {
                 code: this.code,
+                message: detail?.name ? `${this.message}: ${detail.name}` : this.message,
             }
         };
         if (this.list || this.status) json.response.reply = typeof detail === 'string' ? JSON.parse(detail) : detail;
@@ -44,11 +47,13 @@ class ManageCLIResponse {
 
 ManageCLIResponse.HealthStatus = Object.freeze({
     code: 'HealthStatus',
+    message: 'Health status retrieved successfully',
     status: {}
 });
 
 ManageCLIResponse.MetricsStatus = Object.freeze({
     code: 'MetricsStatus',
+    message: 'Metrics status retrieved successfully',
     status: {}
 });
 
@@ -57,6 +62,7 @@ ManageCLIResponse.MetricsStatus = Object.freeze({
 ///////////////////////////////
 ManageCLIResponse.WhiteListIPUpdated = Object.freeze({
     code: 'WhiteListIPUpdated',
+    message: 'WhiteListIP has been updated successfully',
     status: {}
 });
 
@@ -66,25 +72,30 @@ ManageCLIResponse.WhiteListIPUpdated = Object.freeze({
 
 ManageCLIResponse.AccountCreated = Object.freeze({
     code: 'AccountCreated',
+    message: 'Account has been created successfully',
     status: {}
 });
 
 ManageCLIResponse.AccountDeleted = Object.freeze({
     code: 'AccountDeleted',
+    message: 'Account has been deleted successfully'
 });
 
 ManageCLIResponse.AccountUpdated = Object.freeze({
     code: 'AccountUpdated',
+    message: 'Account has been updated successfully',
     status: {}
 });
 
 ManageCLIResponse.AccountStatus = Object.freeze({
     code: 'AccountStatus',
+    message: 'Account status retrieved successfully',
     status: {}
 });
 
 ManageCLIResponse.AccountList = Object.freeze({
     code: 'AccountList',
+    message: 'Account list retrieved successfully',
     list: {}
 });
 
@@ -94,25 +105,30 @@ ManageCLIResponse.AccountList = Object.freeze({
 
 ManageCLIResponse.BucketCreated = Object.freeze({
     code: 'BucketCreated',
+    message: 'Bucket has been created successfully',
     status: {}
 });
 
 ManageCLIResponse.BucketDeleted = Object.freeze({
     code: 'BucketDeleted',
+    message: 'Bucket has been deleted successfully'
 });
 
 ManageCLIResponse.BucketUpdated = Object.freeze({
     code: 'BucketUpdated',
+    message: 'Bucket has been updated successfully',
     status: {}
 });
 
 ManageCLIResponse.BucketStatus = Object.freeze({
     code: 'BucketStatus',
+    message: 'Bucket status retrieved successfully',
     status: {}
 });
 
 ManageCLIResponse.BucketList = Object.freeze({
     code: 'BucketList',
+    message: 'Bucket list retrieved successfully',
     list: {}
 });
 
@@ -122,6 +138,7 @@ ManageCLIResponse.BucketList = Object.freeze({
 
 ManageCLIResponse.LoggingExported = Object.freeze({
     code: 'LoggingExported',
+    message: 'Logging data exported successfully',
     status: {}
 });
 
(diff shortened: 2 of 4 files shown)


src/manage_nsfs/manage_nsfs_cli_responses.js
```diff
@@ -31,7 +31,11 @@ class ManageCLIResponse {
                 code: this.code,
             }
         };
-        if (this.list || this.status) json.response.reply = typeof detail === 'string' ? JSON.parse(detail) : detail;
+        if (this.list || this.status) {
+            json.response.reply = typeof detail === 'string' ? JSON.parse(detail) : detail;
+        } else {
+            json.response.message = typeof detail === 'string' ? detail : `${detail}`;
+        }
```
shirady, line 38: I think you need to explain this change in a comment.
From what I understand you're going to add a `ManageCLIResponse.` to all the responses which are not `list` or `status`, when the change is currently only related to `ManageCLIResponse.AccountDeleted` and `ManageCLIResponse.BucketDeleted`.

In my opinion, and maybe others would think otherwise I would add a property of `delete` (like list and status) and make the change only under this condition. WDYT?

src/cmd/manage_nsfs.js
```diff
@@ -259,7 +259,7 @@ async function delete_bucket(data, force) {
         }
         await native_fs_utils.folder_delete(bucket_temp_dir_path, fs_context_fs_backend, true);
         await config_fs.delete_bucket_config_file(data.name);
-        return { code: ManageCLIResponse.BucketDeleted, detail: '', event_arg: { bucket: data.name } };
+        return { code: ManageCLIResponse.BucketDeleted, detail: `Bucket ${data.name} has been successfully deleted.`, event_arg: { bucket: data.name } };
```
shirady, line 262: You can probably add a small line (as I assume we test the bucket deletion) to test the change in file: test_nc_nsfs_bucket_cli.test.js

src/cmd/manage_nsfs.js
```diff
@@ -467,7 +467,7 @@ async function update_account(data) {
  */
 async function delete_account(data) {
     await config_fs.delete_account_config_file(data);
-    return { code: ManageCLIResponse.AccountDeleted, detail: '', event_arg: { account: data.name } };
+    return { code: ManageCLIResponse.AccountDeleted, detail: `Account ${data.name} has been successfully deleted.`, event_arg: { account: data.name } };
```
shirady, line 470: You can probably add a small line (as I assume we test the account deletion) to test the change in file: test_nc_nsfs_account_cli.test.js

src/manage_nsfs/manage_nsfs_cli_responses.js
achouhan09: We can add message property here.

src/cmd/manage_nsfs.js
achouhan09: Sure.

src/cmd/manage_nsfs.js
achouhan09: Sure.

src/manage_nsfs/manage_nsfs_cli_responses.js
```diff
@@ -31,7 +31,11 @@ class ManageCLIResponse {
                 code: this.code,
             }
         };
-        if (this.list || this.status) json.response.reply = typeof detail === 'string' ? JSON.parse(detail) : detail;
+        if (this.list || this.status) {
+            json.response.reply = typeof detail === 'string' ? JSON.parse(detail) : detail;
+        } else {
+            json.response.message = typeof detail === 'string' ? detail : `${detail}`;
```
nadavMiz, line 37: what is the difference between reply and message?

nadavMiz, line 37: can you add tests for this change

achouhan09: reply is of JSON type and message should be one liner string explaination for the operation

shirady: Maybe you can add this as a comment.

shirady: From what I know if the type is `object` then `${detail}` will look like: [object Object]
For example nodejs REPL:

> > let aa = {a: 1}
> > console.log(aa)
> { a: 1 }
> > console.log(`${aa}`)
> [object Object]

If you add a small test for this change as @nadavMiz you can check it.

achouhan09: @nadavMiz could you please explain what kind of tests you are suggesting here. Also, I have already modified some tests please check.

achouhan09: Sure.

nadavMiz: I think the test you modified are good enough

achouhan09: sure, thanks

src/manage_nsfs/manage_nsfs_cli_responses.js
```diff
@@ -31,7 +33,12 @@ class ManageCLIResponse {
                 code: this.code,
             }
         };
-        if (this.list || this.status) json.response.reply = typeof detail === 'string' ? JSON.parse(detail) : detail;
+        if (this.list || this.status) {
+            json.response.reply = typeof detail === 'string' ? JSON.parse(detail) : detail;
+        }
+        if (this.message) {
+            json.response.message = typeof detail === 'string' ? detail : `${detail}`;
```
shirady, line 40: If `message` it a string, why not set as is?

src/manage_nsfs/manage_nsfs_cli_responses.js
```diff
@@ -71,6 +78,7 @@ ManageCLIResponse.AccountCreated = Object.freeze({
 
 ManageCLIResponse.AccountDeleted = Object.freeze({
     code: 'AccountDeleted',
+    message: {}
```
shirady, line 81: @naveenpaul1 @romayalon 
I want to verify that it is fine to add the property in the response (I don't know if you've got it structure from Scale and I want to know we have the flexibility to add it).

src/cmd/manage_nsfs.js
```diff
@@ -252,7 +252,7 @@ async function delete_bucket(data, force) {
         }
         await native_fs_utils.folder_delete(bucket_temp_dir_path, fs_context_fs_backend, true);
         await config_fs.delete_bucket_config_file(data.name);
-        return { code: ManageCLIResponse.BucketDeleted, detail: '', event_arg: { bucket: data.name } };
+        return { code: ManageCLIResponse.BucketDeleted, detail: `Bucket ${data.name} has been successfully deleted.`, event_arg: { bucket: data.name } };
```
romayalon, line 255: I don't think it makes sense to add a message on every bucket operation response, we should add a this to the manage_nsfs_cli_responses file and set message for every op, we do the same on manage_nsfs_cli_errors 
and please add to all operations and not only to bucket/account delete

romayalon (changes requested):

src/cmd/manage_nsfs.js
achouhan09: Please correct me if I am wrong, if we add message on every bucket operation(then it should be something general like `Account deleted successfully.`). Here, still we need to customize it to put account/bucket etc. name in the message description.
As per the issue, we want more info on which account/bucket etc. was deleted and not a general description.

romayalon: the bucket name/account can be added via the argument, this issue is just an example of a bigger problem and I don't see a reason for not solving it for all the cases

achouhan09: Sure, I will try apply this to all operations. Thanks

achouhan09: Updated the code, please check now. Thanks

src/manage_nsfs/manage_nsfs_cli_responses.js
```diff
@@ -131,16 +148,19 @@ ManageCLIResponse.LoggingExported = Object.freeze({
 
 ManageCLIResponse.UpgradeSuccessful = Object.freeze({
     code: 'UpgradeSuccessful',
+    message: 'Upgrade completed successfully',
```
romayalon, line 151: Can you add to all the Upgrade functions Config Directory Upgrade prefix? for example - Config Directory Upgrade completed successfully etc

src/manage_nsfs/manage_nsfs_cli_responses.js
```diff
@@ -150,25 +170,30 @@ ManageCLIResponse.UpgradeHistory = Object.freeze({
 
 ManageCLIResponse.ConnectionCreated = Object.freeze({
     code: 'ConnectionCreated',
+    message: 'Connection has been successfully created',
```
romayalon, line 173: I think that here as well maybe Notification Connection
@alphaprinz probably can help with this response details

src/manage_nsfs/manage_nsfs_cli_responses.js
```diff
@@ -94,25 +105,30 @@ ManageCLIResponse.AccountList = Object.freeze({
 
 ManageCLIResponse.BucketCreated = Object.freeze({
     code: 'BucketCreated',
+    message: 'Bucket has been successfully created',
```
romayalon, line 108: I would put the 'successfully' in the end of the sentance

src/manage_nsfs/manage_nsfs_cli_responses.js
```diff
@@ -18,17 +19,19 @@ class ManageCLIResponse {
...
         this.status = status;
         this.list = list;
+        this.message = message;
     }
 
     to_string(detail) {
         const json = {
             response: {
                 code: this.code,
+                message: detail.name ? `${this.message}: ${detail.name}` : this.message,
```
romayalon, line 34: does adding the detail after the message makes sense in all cases?
and also, do we always have detail and detail.name?

src/cmd/manage_nsfs.js
```diff
@@ -252,7 +252,7 @@ async function delete_bucket(data, force) {
         }
         await native_fs_utils.folder_delete(bucket_temp_dir_path, fs_context_fs_backend, true);
         await config_fs.delete_bucket_config_file(data.name);
-        return { code: ManageCLIResponse.BucketDeleted, detail: '', event_arg: { bucket: data.name } };
+        return { code: ManageCLIResponse.BucketDeleted, detail: { name: data.name }, event_arg: { bucket: data.name } };
```
romayalon, line 255: are there more cases where we want to add detail?

src/manage_nsfs/manage_nsfs_cli_responses.js
achouhan09: Sure

src/manage_nsfs/manage_nsfs_cli_responses.js
achouhan09: I think I should do this for all the operations. Thanks

src/manage_nsfs/manage_nsfs_cli_responses.js
achouhan09: whenever we want to print the name of the object in the operation then only we are adding detail at the end of message otherwise only the set message would be printed for every operation

src/cmd/manage_nsfs.js
achouhan09: for now I have added this in delete_account() and delete_bucket() operation only. We can print the name of object in message by setting the name property in detail(for any operation) otherwise the set message would be printed

src/manage_nsfs/manage_nsfs_cli_responses.js
romayalon: the issue I see here is with accessing directly to detail.name while detail can be undefined

src/cmd/manage_nsfs.js
romayalon: please fix all the cases in this PR **or** open another gap for the rest of the cases so we will handle it in the future

src/manage_nsfs/manage_nsfs_cli_responses.js
achouhan09: Updated to optional chain. Thanks

src/cmd/manage_nsfs.js
achouhan09: Sure, I will open gap if there are more operations.

achouhan09: @romayalon I just verified other then `connection_management` there is no need add details explicitly. I have updated it. Thanks

For every operation we are printing the name of the object in the message(the code parsed the name automatically if `detail.name` is defined).

romayalon (approved):

PR #8730, opened 2025-01-28
NSFS | versioning | add direcory content versioning PUT action

nadavMiz:
### Explain the changes
1. part of the effort to enable versioning to content directory objects. enable PUT commend for versioned directory objects both with and without a body. when versioning is disabled directory objects work the same as before. after enabling versioning, always create .folder object (even for directory objects without a body). save xattr on .folder instead of directory itself. put object works the same as with other keys only that the .folder is treated as the file
2. if there was already directory content for this key in the old format: move the .folder(if it exist, if it doesn't create a new one) to the .version dir and move the noobaa external attributes to the file.
3. having the directory content external attribute on the directory is used to check if directory keys with the old format existed. for this reason after moving the old key to the .versions directory, remove the noobaa xattr from the directory 
4.  in case there was a key in the old format and a new object was created in suspended mode. override the old .folder and remove the old external attributes from the directory
5. added a configuration value to enable versioning for directory objects. should be false by default
6. added tests

### Issues: 
####GAPS
implement other commends

### Testing Instructions:
automatic:
`sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_bucketspace_versioning.js`

manual:
1. create new bucket
2. add directory objects with alternating versioning status
3. new objects should be in the bucket directory. old versions should be in the .versions directory under the bucket

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -294,6 +294,10 @@ function to_fs_xattr(xattr) {
     return _.mapKeys(xattr, (val, key) => XATTR_USER_PREFIX + key);
 }
 
+function filter_fs_xattr(xattr) {
+    return _.pickBy(xattr, (val, key) => key?.startsWith(XATTR_NOOBAA_INTERNAL_PREFIX));
+}
+
 /**
  * get_random_delay returns a random delay number between base + min and max
  * @param {number} base 
@@ -1199,7 +1203,7 @@ class NamespaceFS {
         try {
             await this._check_path_in_bucket_boundaries(fs_context, file_path);
 
-            if (this.empty_dir_content_flow(file_path, params)) {
+            if (this.should_use_empty_content_dir_optimization() && this.empty_dir_content_flow(file_path, params)) {
                 const content_dir_info = await this._create_empty_dir_content(fs_context, params, file_path);
                 return content_dir_info;
             }
@@ -1337,7 +1341,8 @@ class NamespaceFS {
         const part_upload = file_path === upload_path;
         const same_inode = params.copy_source && copy_res === COPY_STATUS_ENUM.SAME_INODE;
         const should_replace_xattr = params.copy_source ? copy_res === COPY_STATUS_ENUM.FALLBACK : true;
-        const is_dir_content = this._is_directory_content(file_path, params.key);
+        const is_dir_content_unoptimized_flow = this._is_directory_content(file_path, params.key) &&
+            this.should_use_empty_content_dir_optimization();
 
         const stat = await target_file.stat(fs_context);
         this._verify_encryption(params.encryption, this._get_encryption_info(stat));
@@ -1380,18 +1385,21 @@ class NamespaceFS {
                 });
             }
         }
-        if (fs_xattr && !is_dir_content && should_replace_xattr) await target_file.replacexattr(fs_context, fs_xattr);
+        if (fs_xattr && !is_dir_content_unoptimized_flow && should_replace_xattr) {
+            await target_file.replacexattr(fs_context, fs_xattr);
+        }
         // fsync
         if (config.NSFS_TRIGGER_FSYNC) await target_file.fsync(fs_context);
         dbg.log1('NamespaceFS._finish_upload:', open_mode, file_path, upload_path, fs_xattr);
 
         if (!same_inode && !part_upload) {
-            await this._move_to_dest(fs_context, upload_path, file_path, target_file, open_mode, params.key, is_dir_content);
+            await this._move_to_dest(fs_context, upload_path, file_path, target_file, open_mode, params.key,
+                is_dir_content_unoptimized_flow);
         }
 
         // when object is a dir, xattr are set on the folder itself and the content is in .folder file
         // we still should put the xattr if copy is link/same inode because we put the xattr on the directory
-        if (is_dir_content) {
+        if (is_dir_content_unoptimized_flow) {
             await this._assign_dir_content_to_xattr(fs_context, fs_xattr, { ...params, size: stat.size }, copy_xattr);
         }
         stat.xattr = { ...stat.xattr, ...fs_xattr };
@@ -1419,16 +1427,14 @@ class NamespaceFS {
     }
 
     // move to dest GPFS (wt) / POSIX (w / undefined) - non part upload
-    async _move_to_dest(fs_context, source_path, dest_path, target_file, open_mode, key, is_dir_content) {
-        dbg.log2('_move_to_dest', fs_context, source_path, dest_path, target_file, open_mode, key, is_dir_content);
+    async _move_to_dest(fs_context, source_path, dest_path, target_file, open_mode, key, is_dir_content_unoptimized_flow) {
+        dbg.log2('_move_to_dest', fs_context, source_path, dest_path, target_file, open_mode, key, is_dir_content_unoptimized_flow);
         let retries = config.NSFS_RENAME_RETRIES;
         // will retry renaming a file in case of parallel deleting of the destination path
         for (;;) {
             try {
                 await native_fs_utils._make_path_dirs(dest_path, fs_context);
-                if (this._is_versioning_disabled() ||
-                    (this._is_versioning_enabled() && is_dir_content)) {
-                // dir_content is not supported in versioning, hence we will treat it like versioning disabled
+                if (this._is_versioning_disabled() || is_dir_content_unoptimized_flow) {
                     if (open_mode === 'wt') {
                         await target_file.linkfileat(fs_context, dest_path);
                     } else {
@@ -1471,6 +1477,7 @@ class NamespaceFS {
         dbg.log1('Namespace_fs._move_to_dest_version:', new_ver_tmp_path, latest_ver_path, upload_file);
         let gpfs_options;
         const is_gpfs = native_fs_utils._is_gpfs(fs_context);
+        const is_dir_content = this._is_directory_content(latest_ver_path, key);
         let retries = config.NSFS_RENAME_RETRIES;
         for (;;) {
             try {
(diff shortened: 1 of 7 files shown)


src/sdk/namespace_fs.js
```diff
@@ -2730,9 +2755,11 @@ class NamespaceFS {
     }
 
     // returns version path of the form bucket_path/dir/.versions/{key}_{version_id}
-    _get_version_path(key, version_id) {
```
romayalon, line 2733: please add in the comment above the new case you are handling

src/sdk/namespace_fs.js
```diff
@@ -2730,9 +2755,11 @@ class NamespaceFS {
...
-        const key_version = path.basename(key) + (version_id ? '_' + version_id : '');
-        return path.normalize(path.join(this.bucket_path, path.dirname(key), HIDDEN_VERSIONS_PATH, key_version));
+    _get_version_path(key, version_id, is_dir = false) {
+        const key_name = is_dir ? '.folder' : path.basename(key);
```
romayalon, line 2759:
```suggestion
        const key_name = is_dir ? config.NSFS_FOLDER_OBJECT_NAME : path.basename(key);
```

src/sdk/namespace_fs.js
```diff
@@ -1504,14 +1512,27 @@ class NamespaceFS {
...
+                    if (latest_ver_info) {
+                        dbg.log1('NamespaceFS._move_to_dest_version version ID of the latest version is a unique ID - the file will be moved it to .versions/ directory');
+                        await native_fs_utils._make_path_dirs(versioned_path, fs_context);
+                        await native_fs_utils.safe_move(fs_context, latest_ver_path, versioned_path, latest_ver_info,
                         gpfs_options?.move_to_versions, bucket_tmp_dir_path);
+                    }
+                    if (is_disabled_dir_content) {
+                        dbg.log1('NamespaceFS._move_to_dest_version latest object is a directory object with attributes on the directory. move the xattr to the new .version file');
+                        //this scenario happens only after moving from disabled to enabled mode or after upgrade. version-id is always null
+                        versioned_path = versioned_path || this._get_version_path(key, "null", is_dir_content);
```
romayalon, line 1530:
```suggestion
                        versioned_path = versioned_path || this._get_version_path(key, NULL_VERSION_ID, is_dir_content);
```

src/sdk/namespace_fs.js
```diff
@@ -490,6 +494,7 @@ class NamespaceFS {
         this.stats = stats;
         this.force_md5_etag = force_md5_etag;
         this.warmup_buffer = nb_native().fs.dio_buffer_alloc(4096);
+        this.content_dir_versioning = config.NSFS_CONTENT_DIRECTORY_VERSIONING_ENABLED;
```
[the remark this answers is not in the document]
nadavMiz: it isn't needed. just wanted to be consistent with the flag for versioning. but I for versioning it changes. here it doesn't. I can remove it

src/sdk/namespace_fs.js
```diff
@@ -1504,14 +1512,27 @@ class NamespaceFS {
...
                 }
-                if (latest_ver_info &&
-                    ((this._is_versioning_enabled()) ||
-                        (this._is_versioning_suspended() && latest_ver_info.version_id_str !== NULL_VERSION_ID))) {
-                    dbg.log1('NamespaceFS._move_to_dest_version version ID of the latest version is a unique ID - the file will be moved it to .versions/ directory');
-                    await native_fs_utils._make_path_dirs(versioned_path, fs_context);
-                    await native_fs_utils.safe_move(fs_context, latest_ver_path, versioned_path, latest_ver_info,
+                if (((this._is_versioning_enabled()) ||
+                        (this._is_versioning_suspended() && latest_ver_info && latest_ver_info.version_id_str !== NULL_VERSION_ID))) {
+                    if (latest_ver_info) {
```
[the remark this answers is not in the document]
nadavMiz: I moved it inside. the reason was that for empty content directory we don't need to move the .folder because it doesn't exist, but we do need to create a new one and move the xattr to it. however it still only happens if versioning is enabled or suspended and latest is null. however now that I think about it, if there are xattr then they have to be null. I can think of a different way to organize it. to mitigate the change to the old format

src/sdk/namespace_fs.js
```diff
@@ -1504,14 +1512,27 @@ class NamespaceFS {
...
+                        dbg.log1('NamespaceFS._move_to_dest_version version ID of the latest version is a unique ID - the file will be moved it to .versions/ directory');
+                        await native_fs_utils._make_path_dirs(versioned_path, fs_context);
+                        await native_fs_utils.safe_move(fs_context, latest_ver_path, versioned_path, latest_ver_info,
                         gpfs_options?.move_to_versions, bucket_tmp_dir_path);
+                    }
+                    if (is_disabled_dir_content) {
```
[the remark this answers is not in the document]
nadavMiz: because its part of moving the latest to the latest dir. in the old, optimized way the xattrs and the .folder file are separated. when moving to the .versions we now need to combine them

src/sdk/namespace_fs.js
```diff
@@ -1504,14 +1512,27 @@ class NamespaceFS {
                         // remove a version (or delete marker) with null version ID from .versions/ (if exists)
                         await this._delete_null_version_from_versions_directory(key, fs_context);
                     }
+                    if (is_disabled_dir_content) {
+                        await this._clear_user_xattr(fs_context, path.dirname(latest_ver_path), XATTR_USER_PREFIX);
```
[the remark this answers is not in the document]
nadavMiz: same reason. here this is not entirely needed, because here we just clear the xattr from the directory file. but having the xattr on the directory file is the way I use to identify that we have the optimized content directory. so clearing signals that we no longer have the optimized content dir. it needs to be here because its after we delete the latest

src/sdk/namespace_fs.js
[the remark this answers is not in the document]
romayalon: I see latest_ver_info in the outer and the inner conditions

src/sdk/namespace_fs.js
```diff
@@ -1504,14 +1512,27 @@ class NamespaceFS {
...
+                        (this._is_versioning_suspended() && latest_ver_info && latest_ver_info.version_id_str !== NULL_VERSION_ID))) {
+                    if (latest_ver_info) {
+                        dbg.log1('NamespaceFS._move_to_dest_version version ID of the latest version is a unique ID - the file will be moved it to .versions/ directory');
+                        await native_fs_utils._make_path_dirs(versioned_path, fs_context);
+                        await native_fs_utils.safe_move(fs_context, latest_ver_path, versioned_path, latest_ver_info,
                         gpfs_options?.move_to_versions, bucket_tmp_dir_path);
+                    }
+                    if (is_disabled_dir_content) {
+                        dbg.log1('NamespaceFS._move_to_dest_version latest object is a directory object with attributes on the directory. move the xattr to the new .version file');
+                        //this scenario happens only after moving from disabled to enabled mode or after upgrade. version-id is always null
```
romayalon, line 1529: let's move it to a function and explain exactly what is done here

romayalon: in general if you choose keeping xattr implementation in content flow, let's try to organize it better

nadavMiz: I moved all handling of the xattr to a single function, separate from the original flow, since the original flow is the same for directory objects. it can remain the same. its good, because this will also be relevant for delete latest operation

romayalon (approved):

src/sdk/namespace_fs.js
```diff
@@ -1419,16 +1427,14 @@ class NamespaceFS {
     }
 
     // move to dest GPFS (wt) / POSIX (w / undefined) - non part upload
-    async _move_to_dest(fs_context, source_path, dest_path, target_file, open_mode, key, is_dir_content) {
-        dbg.log2('_move_to_dest', fs_context, source_path, dest_path, target_file, open_mode, key, is_dir_content);
+    async _move_to_dest(fs_context, source_path, dest_path, target_file, open_mode, key, is_old_dir_flow) {
```
romayalon, line 1430: I still think is_old_dir_flow is not an understood name, please change it to explicit name and add comments on what it is

src/sdk/namespace_fs.js
```diff
@@ -1471,6 +1477,7 @@ class NamespaceFS {
         dbg.log1('Namespace_fs._move_to_dest_version:', new_ver_tmp_path, latest_ver_path, upload_file);
         let gpfs_options;
         const is_gpfs = native_fs_utils._is_gpfs(fs_context);
+        const is_dir = this._is_directory_content(latest_ver_path, key);
```
romayalon, line 1480: same here - the naming is confusing, is_dir_content is more accurate, there can be a dir that is not a content dir

src/sdk/namespace_fs.js
```diff
@@ -1536,6 +1546,31 @@ class NamespaceFS {
         }
     }
 
+    // handle xattr of content dir when moving from disabled to enabled mode.
+    // in the disabled version of content dir, the xattr is on the directory itself. so need to move it seperatly from the obejct
+    // in case of enabled mode we need to move the xattr to the new object
+    // both for suspended and enabled mode we need to clear the user xattr from the directory
+    async _handle_latest_disabled_dir_content_xattr(fs_context, key, versioned_path, latest_ver_path) {
```
romayalon, line 1553: make sure to add jsdoc and not just comments, it helps a lot to the linter with finding issues

src/sdk/namespace_fs.js
```diff
@@ -1536,6 +1546,31 @@ class NamespaceFS {
...
+        const directory_stat = await native_fs_utils.stat_ignore_enoent(fs_context, path.dirname(latest_ver_path));
```
romayalon, line 1554: I would create a variable for the path since it's used multiple times
`const latest_version_dir_path = path.dirname(latest_ver_path);`
or something similar

src/sdk/namespace_fs.js
```diff
@@ -2730,9 +2769,12 @@ class NamespaceFS {
     }
 
     // returns version path of the form bucket_path/dir/.versions/{key}_{version_id}
-    _get_version_path(key, version_id) {
-        const key_version = path.basename(key) + (version_id ? '_' + version_id : '');
-        return path.normalize(path.join(this.bucket_path, path.dirname(key), HIDDEN_VERSIONS_PATH, key_version));
+    // in case of directory, the path is bucket_path/dir/key/.versions/.folder_{version_id}
```
romayalon, line 2772: a directory content, I would rephrase comment and is_dir variable

PR #8733, opened 2025-01-29
Return 405 for Get/Head Specific Delete-Marker

### LINKED ISSUE #8369: Containerized | Versioning | Return 405 for Get/Head Specific Delete-Marker
### Environment info
- NooBaa Version: master 5.18
- Platform: Containerized

### Background
In PR #8338 added a fix in NSFS.
Since I added the mapping between RPC error and S3 error: `METHOD_NOT_ALLOWED: S3Error.MethodNotAllowed` it might be that we probably have this issue in containerized.

### Actual behavior
1. If the specified version in the request is a delete marker the returned is 404 Not Found.

### Expected behavior
1. If the specified version in the request is a delete marker the returned is 405 Method Not Allowed.

### Steps to reproduce
1. Create a data bucket with versioning Enabled.
2. Put an object and then delete an object (save the version-id).
3. Try to head/get with the key by the version-id.

1. According to the AWS docs [GetObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html) and [HeadObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html):
`If the specified version in the request is a delete marker, the response returns a 405 Method Not Allowed error and the Last-Modified: timestamp response header.`

vh05:
If the specified version in the request is a delete marker the returned is 405 Method Not Allowed.

1. Create a versioned bucket and multiple versions of the object by copying the same object multiple times
 ```
   # nb-s3 mb s3://test-16
   # nb-s3 cp Makefile s3://test-16
```

2. delete the object
`nb-s3api delete-object --bucket test-16 --key Makefile`

3. Now head/get object with version id of the delete marker (list-object-versions will give the delete version marker id)

```
# nb-s3api head-object --bucket test-16 --key Makefile --version-id nbver-9
An error occurred (405) when calling the HeadObject operation: Method Not Allowed

```

Output:

Fixes: https://github.com/noobaa/noobaa-core/issues/8369

### DIFF
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -1568,6 +1568,12 @@ function get_obj_id(req, rpc_code) {
  * @param {string} rpc_code
  */
 function check_object_mode(req, obj, rpc_code) {
+    if (obj && obj.delete_marker && req.rpc_params.version_id) {
+        throw new RpcError('METHOD_NOT_ALLOWED',
+            'Method not allowed, delete object id of entry delete marker',
+            { last_modified: obj.last_modified_time, delete_marker: true });
+    }
+
     if (!obj || obj.deleted || obj.delete_marker) {
         throw new RpcError(rpc_code,
             `No such object: obj_id ${req.rpc_params.obj_id} bucket ${req.rpc_params.bucket} key ${req.rpc_params.key}`);

### CODE REVIEW

File: src/server/object_services/object_server.js [hunk 1]
```diff
@@ -1568,7 +1568,13 @@ function get_obj_id(req, rpc_code) {
  * @param {string} rpc_code
  */
 function check_object_mode(req, obj, rpc_code) {
-    if (!obj || obj.deleted || obj.delete_marker) {
+    if (obj && obj.delete_marker) {
```
shirady, line 1571: How can we be sure that in this case the object is a delete_marker and also the version ID was specified in the request? From reading this condition I only know that it is a delete marker.

vh05: The bucket with version enabled and the object is deleted, we can see the delete marker object as true. In other case the delete marker is false or obj it self is null.

Correct me if wrong

shirady: What about the request?
The difference here that it was on delete marker with version id.

According to the AWS docs [GetObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html) and [HeadObject](https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html):(https://docs.aws.amazon.com/AmazonS3/latest/API/API_GetObject.html#API_GetObject_ResponseElements):
If the specified version in the request is a delete marker, the response returns a 405 Method Not Allowed error and the Last-Modified: timestamp response header.

vh05: This is the behavior what I saw on aws s3 
head object
```
# aws s3api head-object --bucket vh-22 --key Makefile --version-id aqGt6sJfDntex6GPXfpl2RZwa5fUQYgV

An error occurred (405) when calling the HeadObject operation: Method Not Allowed
```

get-object
```
# aws s3api get-object --bucket vh-22 --key Makefile --version-id aqGt6sJfDntex6GPXfpl2RZwa5fUQYgV a

An error occurred (MethodNotAllowed) when calling the GetObject operation: The specified method is not allowed against this resource.

```

vh05: I think now I got what you are saying,

If we dont specify version id on delete marker, we should return ENOENT and version id is given we should return Method Not allowed. Am I right ?

shirady: We need to be compliant with the S3 protocol.
So in the previous check that you did on AWS you can also do it without "--version-id" flag and look at the errors.
Then you need to test the same cases on noobaa (on the a data bucket) and ensure the behavior is the same.
`ENOENT` is the term in filesystem, we need to return the right RPC error (that would be translated to the S3 error we need to return).
I hope it is clear.

vh05:
> @vh05, Could you attach the manual testing instructions you performed and the screenshot of the result? (to help us to see the result) Do you plan to add to this PR automatic tests? If you need I added test cases in NSFS in PR #8338 and #8402



> @vh05, Could you attach the manual testing instructions you performed and the screenshot of the result? (to help us to see the result) Do you plan to add to this PR automatic tests? If you need I added test cases in NSFS in PR #8338 and #8402

Yes, will add test

PR #8744, opened 2025-01-31
xml utils (for bucket notifications) - option for including array's key

alphaprinz:
### Explain the changes
1. Currently xml_utils.encode_xml() does not add a key per item in the array.
This causes mangled output for get-bucket-notif when more than once event is configured:
`<NotificationConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><TopicConfiguration><Event>s3:ObjectCreated:*s3:ObjectRemoved:Delete</Event><Topic>notif-secret/connect_http_c.json</Topic><Id>created_from_s3op_secret2</Id></TopicConfiguration></NotificationConfiguration>`

And aws cli output is:
```
{
    "TopicConfigurations": [
        {
            "Id": "created_from_s3op_secret2",
            "TopicArn": "notif-secret/connect_http_c.json",
            "Events": [
                "s3:ObjectCreated:*s3:ObjectRemoved:Delete"
            ]
        }
    ]
}
```

After change xml is with a tag per item in Event array:
`<NotificationConfiguration xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><TopicConfiguration><Event>s3:ObjectCreated:*</Event><Event>s3:ObjectRemoved:Delete</Event><Topic>notif-secret/connect_http_c.json</Topic><Id>created_from_s3op_secret2</Id></TopicConfiguration></NotificationConfiguration>
`
And aws cli output is:

```
{
    "TopicConfigurations": [
        {
            "Id": "created_from_s3op_secret2",
            "TopicArn": "notif-secret/connect_http_c.json",
            "Events": [
                "s3:ObjectCreated:*
                 s3:ObjectRemoved:Delete"
            ]
        }
    ]
}
```

1. https://issues.redhat.com/browse/DFBUGS-1507

### Testing Instructions:
1. Configure more than one event filtering in bucket notification.
2. Invoke get-bucket-notif.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/xml_utils.js
+++ b/src/util/xml_utils.js
@@ -18,7 +18,7 @@ const XML_HEADER = '<?xml version="1.0" encoding="UTF-8"?>';
  *
  * the object will be traveresed recursively:
  *  - any object will be encoded such that keys are xml tags.
- *  - any array will be encode without introducing new keys, meaning items are flattened.
+ *  - any array will be encode without introducing new keys, meaning items are flattened, unless repeat_array_tags is true.
  *  - string/number/boolean/date will be converted to strings.
  *  - {key: null} will encode an empty tag - <key></key>
  *  - {key: undefined} will not encode the tag at all.
@@ -31,22 +31,24 @@ const XML_HEADER = '<?xml version="1.0" encoding="UTF-8"?>';
  *   <root> <a>1</a> <b>2</b> <a>3</a> <b>4</b> <z>42</z> <c d="5"></c> </root>
  *
  */
-function encode_xml(object, ignore_header) {
+function encode_xml(object, ignore_header, repeat_array_tags) {
     let output = ignore_header ? '' : XML_HEADER;
     append_object(s => {
         output += s;
-    }, object);
+    }, object, repeat_array_tags);
     return output;
 }
 
-function append_object(append, object) {
+function append_object(append, object, repeat_array_tags, father) {
     if (typeof(object) !== 'object') {
         append(encode_xml_str(object));
     } else if (Array.isArray(object)) {
         // arrays are encoded without adding new tags
         // which allows repeating keys with the same name
         for (let i = 0; i < object.length; ++i) {
-            append_object(append, object[i]);
+            if (repeat_array_tags && i !== 0) append('<' + father + '>');
+            append_object(append, object[i], repeat_array_tags, father);
+            if (repeat_array_tags && object.length !== i + 1) append('</' + father + '>');
         }
     } else {
         // skip any keys from the prototype
@@ -63,7 +65,7 @@ function append_object(append, object) {
             // otherwise ignore the key and value altogether
             if (key[0] === '_') {
                 if (key === '_content') {
-                    append_object(append, val);
+                    append_object(append, val, repeat_array_tags, key);
                 }
                 continue;
             }
@@ -79,7 +81,7 @@ function append_object(append, object) {
                 } else {
                     append('<' + key + '>');
                 }
-                append_object(append, val);
+                append_object(append, val, repeat_array_tags, key);
                 append('</' + key + '>');
             } else {
                 append('<' + key + '>');
--- a/src/endpoint/s3/ops/s3_get_bucket_notification.js
+++ b/src/endpoint/s3/ops/s3_get_bucket_notification.js
@@ -49,5 +49,6 @@ module.exports = {
     },
     reply: {
         type: 'xml',
+        repeat_array_tags: true,
     },
 };
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -350,8 +350,8 @@ function send_reply(req, res, reply, options) {
             reply;
         // TODO: Refactor later on to support potential headers response and delayed XML body
         // This is done for the complete since we assign the XML header only in body prior to responding
-        const xml_reply = xml_utils.encode_xml(xml_root, /* ignore_header */ res.headersSent);
-        dbg.log1('HTTP REPLY XML', req.method, req.originalUrl,
+        const xml_reply = xml_utils.encode_xml(xml_root, /* ignore_header */ res.headersSent, options.reply.repeat_array_tags);
+        dbg.log0('HTTP REPLY XML', req.method, req.originalUrl,
             JSON.stringify(req.headers),
             xml_reply.length <= 2000 ?
             xml_reply : xml_reply.slice(0, 1000) + ' ... ' + xml_reply.slice(-1000));


guymguym (changes requested):
Hey, I don't think we need new options. Let me explain - our (homemade) xml encoder is really simple and fast and only encodes an xml tag when it encounter an object property. Also it completely ignores arrays - meaning it flattens their contents into the same parent tag, even deeply nested arrays will generate no xml tags at all. To repeat the same tag we make an array of objects like `[ {Event:"..."}, {Event:"...'} ]` and each object can use the same property name to repeat the same tag. It should be capable to construct any xml structure. However arrays of strings are just assumed to be appended as raw content.

See how we build the [list-objects reply](https://github.com/noobaa/noobaa-core/blob/b225590eb6f6fb5acc22901c98105338c64849a5/src/endpoint/s3/ops/s3_get_bucket.js#L57-L96) for example which also has repeating Contents and CommonPrefixes tags.

Here's examples -

```
node -p 'x=require("./src/util/xml_utils").encode_xml({ NotificationConfiguration: [{ TopicConfiguration: [ { Id: "11", Topic: "T", Filter: "F" }, Array(5).fill().map((v,i) => ({ Event: i })) ] } , { TopicConfiguration: [{ Id: "22" }] }] })' | xmllint --format -

<?xml version="1.0" encoding="UTF-8"?>
<NotificationConfiguration>
  <TopicConfiguration>
    <Id>11</Id>
    <Topic>T</Topic>
    <Filter>F</Filter>
    <Event>0</Event>
    <Event>1</Event>
    <Event>2</Event>
    <Event>3</Event>
    <Event>4</Event>
  </TopicConfiguration>
  <TopicConfiguration>
    <Id>22</Id>
  </TopicConfiguration>
</NotificationConfiguration>
```

A single js property makes just a single tag, and the contents of the tag simply appends the bare strings as "body" of the tag, which just looks like a long string -
```
node -p 'x=require("./src/util/xml_utils").encode_xml({ Reply: { Event: ["EEE","FFF","GGG"] } })' | xmllint --format -

<?xml version="1.0" encoding="UTF-8"?>
<Reply>
  <Event>EEEFFFGGG</Event>
</Reply>
```

arrays of objects allow to repeat xml tags -
```
node -p 'x=require("./src/util/xml_utils").encode_xml({ Reply: [{Event:"EEE"},{Event:"FFF"},{Event:"GGG"}] })' | xmllint --format -

<?xml version="1.0" encoding="UTF-8"?>
<Reply>
  <Event>EEE</Event>
  <Event>FFF</Event>
  <Event>GGG</Event>
</Reply>
```

guymguym:
This comment is also useful - 

https://github.com/noobaa/noobaa-core/blob/b225590eb6f6fb5acc22901c98105338c64849a5/src/util/xml_utils.js#L15-L34

alphaprinz:
can't say that format makes sense to me.
Why would TopicConfiguration be an array?
It should be an object with named fields.
At any rate, I'll move this to draft and I've opened https://github.com/noobaa/noobaa-core/pull/8745

guymguym:
Yeah, I agree with you it's not super intuitive, but it does simplify the encoder flow. There are some corner cases that we would need to consider - we have a object structure to specify xml tag attributes which is `{_content: ... , _attr: {key:"value"}}` which isn't handled well in your change. We can add this new encoder structure, bit I am just worried it will create more confusions having two very different behaviors. Notice we use an xml parser library (xml2js) that parses repeating tags to arrays (which is like you suggest) but it actually always creates arrays even for a single tag, which is also not very convenient, but still pragmatic enough. If you want to invest in this refactoring a bit more we can make it work but it might be simpler to write as a new encoder function.

Here is how we encode tags with attributes:
```
node -p 'x=require("./src/util/xml_utils").encode_xml({ Reply: [{Event:{_content:"EEE",_attr:{key:"val"}}},{Event:"FFF"},{Event:"GGG"}] })' | xmllint --format -

<?xml version="1.0" encoding="UTF-8"?>
<Reply>
  <Event key="val">EEE</Event>
  <Event>FFF</Event>
  <Event>GGG</Event>
</Reply>
```

Here is how the parser builds the js structure with arrays:
```
node -e 'x=require("xml2js"); x.parseStringPromise(`<?xml version="1.0" encoding="UTF-8"?>
<NotificationConfiguration>
  <TopicConfiguration>
    <Id>id</Id>
    <Topic>topic</Topic>
    <Event>ev1</Event>
    <Event>ev2</Event>
  </TopicConfiguration>
  <TopicConfiguration>
    <Id>22</Id>
  </TopicConfiguration>
</NotificationConfiguration>`).then(x => console.log(util.inspect(x, {depth:null})))'
{
  NotificationConfiguration: {
    TopicConfiguration: [
      { Id: [ 'id' ], Topic: [ 'topic' ], Event: [ 'ev1', 'ev2' ] },
      { Id: [ '22' ] }
    ]
  }
}
```

And now with tag attributes:
```
node -e 'x=require("xml2js"); x.parseStringPromise(`<?xml version="1.0" encoding="UTF-8"?>
<NotificationConfiguration>
  <TopicConfiguration>
    <Id>id</Id>
    <Topic>topic</Topic>
    <Event key="val">ev1</Event>
    <Event>ev2</Event>
  </TopicConfiguration>
</NotificationConfiguration>`).then(x => console.log(util.inspect(x, {depth:null})))'
{
  NotificationConfiguration: {
    TopicConfiguration: [
      {
        Id: [ 'id' ],
        Topic: [ 'topic' ],
        Event: [ { _: 'ev1', '$': { key: 'val' } }, 'ev2' ]
      }
    ]
  }
}
```

src/util/xml_utils.js
```diff
@@ -63,7 +65,7 @@ function append_object(append, object) {
             // otherwise ignore the key and value altogether
             if (key[0] === '_') {
                 if (key === '_content') {
-                    append_object(append, val);
+                    append_object(append, val, repeat_array_tags, key);
```
guymguym, line 68: key here is the literal string "_content", but the parent tag name is probably still father (or mother, or better - parent).

src/util/xml_utils.js
```diff
@@ -31,22 +31,24 @@ const XML_HEADER = '<?xml version="1.0" encoding="UTF-8"?>';
...
     if (typeof(object) !== 'object') {
         append(encode_xml_str(object));
     } else if (Array.isArray(object)) {
         // arrays are encoded without adding new tags
         // which allows repeating keys with the same name
         for (let i = 0; i < object.length; ++i) {
-            append_object(append, object[i]);
+            if (repeat_array_tags && i !== 0) append('<' + father + '>');
+            append_object(append, object[i], repeat_array_tags, father);
+            if (repeat_array_tags && object.length !== i + 1) append('</' + father + '>');
```
guymguym, line 51: this does not support tag attributes, for this encoder mode I would change the level of recursion and encode the arrays tags by the caller

alphaprinz:
Superseded by https://github.com/noobaa/noobaa-core/pull/8745

PR #8745, opened 2025-01-31
bucket notifications - adapt to xml conversion

alphaprinz:
### Explain the changes
1. Adapt get-bucket-notif reply to be converted to xml.

1. https://issues.redhat.com/browse/DFBUGS-1507 (partly)

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_get_bucket_notification.js
+++ b/src/endpoint/s3/ops/s3_get_bucket_notification.js
@@ -8,38 +8,25 @@ const _ = require('lodash');
  */
 async function get_bucket_notification(req) {
 
-    let result = await req.object_sdk.get_bucket_notification({
+    const result = await req.object_sdk.get_bucket_notification({
         bucket_name: req.params.bucket,
     });
 
-    result = _.cloneDeep(result);
-
-    const TopicConfiguration = [];
-
-    //adapt to aws cli structure
-    if (result && result.length > 0) {
-        for (const conf of result) {
-            conf.Event = conf.event;
-            conf.Topic = conf.topic;
-            conf.Id = conf.id;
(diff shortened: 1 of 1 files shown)


guymguym (commented):
just a suggestion of how we typically build these xml results from js objects with a more functional style.

we use _.map(arr, fn) just because it won't fail if the array is undefined, while for arr.map(fn) will fail, but if you prefer you can also just do (arr || []).map(fn).

```js
const topics = await req.object_sdk.get_bucket_notification({
    bucket_name: req.params.bucket,
});

return {
    NotificationConfiguration: _.map(topics, t => ({
        TopicConfiguration: [
            {
                Id: t.id,
                Topic: t.topic,
            },
            _.map(t.event, e => ({
                Event: e
            }))
        ]
    }))
};

File: src/endpoint/s3/ops/s3_get_bucket_notification.js [hunk 1]
```diff
@@ -12,34 +12,21 @@ async function get_bucket_notification(req) {
...
-                TopicConfiguration
-        } :
-        //if there's no notification, return empty NotificationConfiguration tag
-        { NotificationConfiguration: {} };
-
-    return reply;
+    //adapt to structure that xml_utils.encode_xml() will then
+    //encode into aws compliant reply
+    return {
+        NotificationConfiguration: [_.map(result, t => ({
```
guymguym, line 18: no need to wrap the map with [] because it always returns an array
```suggestion
        NotificationConfiguration: _.map(result, t => ({
```

guymguym (approved):
approved, just one minor comment

PR #8751, opened 2025-02-03
NC | Multi Protocol Access | List object with conflicting ownership

### LINKED ISSUE #8735: NC | Multi Protocol Access | If there is an object inside the bucket directory with a conflicting ownership (other than account owner), list objects throws access denied
### Environment info
- NooBaa Version: **noobaa-core-5.18.0-20250125.el9.x86_64**
- Platform: **Standalone noobaa**

### Actual behavior
1. If there is an object inside the bucket directory with a conflicting ownership (other than account owner), the entire bucket will become access denied.

### Expected behavior
1. Ideally if the user tries to list the bucket via AWScli, the bucket owner should be able to list all the objects inside the bucket directory which he has ownership. In multi protocol scenario, same directory can be used as s3 bucket directory or smb/nfs share. Thus if the files created inside the directory might have a different UID:GID than s3 account owner.

Will explain in detail in next section

### Steps to reproduce

1.Create a bucket and upload two objects in to it. Account owner is 10000:10000
```
[root@5221new-11 bucket001]# pwd
/mnt/gpfs0/account001/bucket001

[root@5221new-11 bucket001]# ls -lrt
total 10248
-rw-rw----. 1 10000 10000 10485760 Jan 29 06:23 10MB_file.bin
-rw-rw----. 1 10000 10000        5 Jan 29 06:24 abcd.txt
```
Able to list the objects successfully via AWScli

```
[root@akarattuparambil-scale-host ~]# account001 s3 ls s3://bucket001 2>/dev/null
2025-01-29 06:23:53   10485760 10MB_file.bin
2025-01-29 06:24:15          5 abcd.txt
```
2. Create another object inside the bucket with a different user. This is possible since the same directory can be used for NFS/SMB as well.

```
[root@5221new-11 bucket001]# touch test.txt ; chmod 660 test.txt; ls -lrt
total 10248
-rw-rw----. 1 10000 10000 10485760 Jan 29 06:23 10MB_file.bin
-rw-rw----. 1 10000 10000        5 Jan 29 06:24 abcd.txt
-rw-rw----. 1 root  root         0 Jan 29 09:37 test.txt   ---> Different owner. Might be from an NFS user
```

3. Now the bucket is inaccessible. 

```
[root@akarattuparambil-scale-host ~]# account001 s3 ls s3://bucket001
urllib3/connectionpool.py:1061: InsecureRequestWarning: Unverified HTTPS request is being made to host '5221new-11'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings

An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied
```

naveenpaul1:
### Explain the changes
1. Object losting will skip inaccessible item when adding it to result array
2. test cases added

1. https://github.com/noobaa/noobaa-core/issues/8735

### Testing Instructions:
1. Check issue [8735](https://github.com/noobaa/noobaa-core/issues/8735)

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -721,7 +721,10 @@ class NamespaceFS {
                         if (pos < results.length) {
                             results.splice(pos, 0, r);
                         } else {
-                            results.push(r);
+                            const stat = await native_fs_utils.stat_ignore_eacces(this.bucket_path, r, fs_context);
+                            if (stat) {
+                                results.push(r);
+                            }
                         }
                         if (results.length > limit) {
                             results.length = limit;
--- a/src/test/unit_tests/test_nsfs_access.js
+++ b/src/test/unit_tests/test_nsfs_access.js
@@ -9,6 +9,11 @@ const fs_utils = require('../../util/fs_utils');
 const nb_native = require('../../util/nb_native');
 const test_utils = require('../system_tests/test_utils');
 const fs = require('fs');
+const { TMP_PATH } = require('../system_tests/test_utils');
+const NamespaceFS = require('../../sdk/namespace_fs');
+const buffer_utils = require('../../util/buffer_utils');
+const endpoint_stats_collector = require('../../sdk/endpoint_stats_collector');
+const SensitiveString = require('../../util/sensitive_string');
 
 const new_umask = process.env.NOOBAA_ENDPOINT_UMASK || 0o000;
 const old_umask = process.umask(new_umask);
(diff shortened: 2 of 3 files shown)


src/sdk/namespace_fs.js
```diff
@@ -924,6 +919,24 @@ class NamespaceFS {
         }
     }
 
+    async validate_results(results, fs_context) {
```
romayalon, line 922: if you remove from the list right before it's being returned to the user, you might be left with less than 1000 objects, even if you have access to more than 1000 objects, can you please check that?

src/sdk/namespace_fs.js
```diff
@@ -924,6 +919,24 @@ class NamespaceFS {
...
+            const entry_path = path.join(this.bucket_path, r.key);
+            //If entry is outside of bucket, returns stat of symbolic link
+            const use_lstat = !(await this._is_path_in_bucket_boundaries(fs_context, entry_path));
+            try {
+                r.stat = await nb_native().fs.stat(fs_context, entry_path, { use_lstat });
+            } catch (err) {
+                dbg.warn('NamespaceFS: validate_results : couldnt access file entry_path', entry_path, ", skipping...");
+                const index = results.indexOf(r);
+                if (index > -1) {
+                    results.splice(index, 1);
```
romayalon, line 934: what about bucket owner, do we always want to remove this entry from the list? this should be compatible to S3

naveenpaul1: @romayalon When removing an item from the list, there is a chance the list object returns with less than 1000 items One solution would be to check the access while adding the item to result in an array and return 1000 items

src/sdk/namespace_fs.js
```diff
@@ -877,13 +878,7 @@ class NamespaceFS {
...
             const prefix_dir_key = prefix.slice(0, prefix.lastIndexOf('/') + 1);
             await process_dir(prefix_dir_key);
-            await Promise.all(results.map(async r => {
-                if (r.common_prefix) return;
-                const entry_path = path.join(this.bucket_path, r.key);
-                //If entry is outside of bucket, returns stat of symbolic link
-                const use_lstat = !(await this._is_path_in_bucket_boundaries(fs_context, entry_path));
-                r.stat = await nb_native().fs.stat(fs_context, entry_path, { use_lstat });
-            }));
+            await this.validate_results(results, fs_context);
```
nadavMiz, line 881: I am not sure this is a good name for this function. there is no much validation going on here. maybe change it to something like _get_stats_for_list_object. also, in my opinion, if you like to move it to a function it would be better to move the lambda function to a desperate function, instead of the whole Promise.all. WDYT?

src/sdk/namespace_fs.js
```diff
@@ -924,6 +919,24 @@ class NamespaceFS {
...
 
+    async validate_results(results, fs_context) {
+        await Promise.all(results.map(async r => {
+            if (r.common_prefix) return;
+            const entry_path = path.join(this.bucket_path, r.key);
+            //If entry is outside of bucket, returns stat of symbolic link
+            const use_lstat = !(await this._is_path_in_bucket_boundaries(fs_context, entry_path));
+            try {
+                r.stat = await nb_native().fs.stat(fs_context, entry_path, { use_lstat });
+            } catch (err) {
```
nadavMiz, line 930: notice that you throw for all errors and not just EACCES( permission denied) is this intentional? maybe it would be better to create a separate function that doesn't throw for selected errors. like we do in unlink_ignore_enoent. that way we can reuse in case we would like to handle permission issues in other places. see [stat](https://man7.org/linux/man-pages/man2/stat.2.html) for possible error list

nadavMiz: also that way we will be able to handle what happens in case of EACCES outside the catch. something like:
```
const stat = stat_ignore_eaccess();
if (!stat) {
    ....
}

romayalon:
@naveenpaul1 I feel we need here a short design, with attention to the small details of this behavioral change mentioned above, the requester, the error coming from the file system, how we are not breaking S3 features etc, I think it'll be great if we can discuss it internally and also with Scale.

src/sdk/namespace_fs.js
naveenpaul1: @nadavMiz Here only error we will get is an access error, files are added to the result array only because they already exist in the folder. But I like the idea of util method for access check.

romayalon (approved):

Issue #8762, opened 2025-02-04
Error messages in noobaa.log are misguiding


ramya-c3:
### Environment info
- NooBaa Version: noobaa-core-5.18.0-20250201.el9.x86_64
- Platform: standalone

### Actual behavior
Even though the notification is succesful it is logging unwanted error message so it should be removed

an 6 11:27:19 \node[1123933]: [/1123933] [WARN]
core.util.persistent_logger:: failed to rename active file: [Error: No such file or directory] {
code: 'ENOENT', context: 'Rename _old_path=/mnt/cesSharedRoot/ces/s3-
config/notifications_log/rtest_notification_logging.failure.log
_new_path=/mnt/cesSharedRoot/ces/s3-config/notifications_log/ramyamc-
22_notification_logging.failure.1736159239159.log ' }

### Expected behavior
Even though the notification is succesful it is logging unwanted error message so it should be removed

### Steps to reproduce
1.

PR #8765, opened 2025-02-04
persistent logger (for bucket notifications) - don't log rename ENOENT if there were no failures

alphaprinz:
### Explain the changes
1. If there were no failures during active persistent logger processing, the rename for the active failure file fails (because it's not there). Scale have opened a bug for this log.
This PR changes the rename to not log ENOENT in case there were no failures.
Let me know if you have other ideas.

1. https://github.com/noobaa/noobaa-core/issues/8762

### Testing Instructions:
1. Run notifications with one successful notifications. No WARN for renaming of active failure log file should be logged.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/persistent_logger.js
+++ b/src/util/persistent_logger.js
@@ -198,7 +198,7 @@ class PersistentLogger {
 
             try {
                 // Finally replace the current active so as to consume them in the next iteration
-                await failure_log._replace_active();
+                await failure_log._replace_active(!result);
             } catch (error) {
                 dbg.error('failed to replace active failure log:', error, 'log_namespace:', this.namespace);
             }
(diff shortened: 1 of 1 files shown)


src/util/persistent_logger.js
```diff
@@ -198,7 +198,7 @@ class PersistentLogger {
 
             try {
                 // Finally replace the current active so as to consume them in the next iteration
-                await failure_log._replace_active();
+                await failure_log._replace_active(!result);
```
tangledbytes, line 201: We should log `ENOENT` if the last processed journal (aka log) in a journal namespace was not deleted?

alphaprinz: Yes, the not-delete file needs to be replaced. (That's what I understand, correct me if I'm wrong. In my use case I always delete the the processed files.)

tangledbytes: Actually, I think I missed something when I was looking at this last time. This makes sense.

tangledbytes (approved):

PR #8769, opened 2025-02-05
NSFS | add configuration flag to enable dynamic supplemental groups allocation

nadavMiz:
### Explain the changes
1. add environment variable to enable/disable dynamic supplemental groups allocation
2. in case this value is set in the configuration file. the configuration file value will override the default value
3. change os_linux, os_darwing to get supplemental groups dinamicly only when this flag is not false

#### GAP
implement and test on containerized environment: dynamic supplemental groups should work. need to test environment variable passes correctly to fs_worker

### Testing Instructions:
1. automatic tests: `sudo node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nsfs_access.js`
2. manual tests:
    a. add new user with uid that matches an fs user with supplemental groups
    b. add new user that his GID matches a group user 1 is a part of
    c. create new bucket by user 2
    d. add `"NSFS_ENABLE_DINAMIC_SUPPLEMENTAL_GROUPS" : false` to config.json
    e. put object using user 1
    f. operation should fail. should get `permission denied`

- [ ] Doc added/updated
- [x] Tests added


config.js
```diff
@@ -822,6 +822,8 @@ config.NSFS_UPDATE_ISSUES_REPORT_ENABLED = true;
 config.NSFS_EXIT_EVENTS_TIME_FRAME_MIN = 24 * 60; // per day
 config.NSFS_MAX_EXIT_EVENTS_PER_TIME_FRAME = 10; // allow max 10 failed forks per day
 
+process.env.NSFS_ENABLE_DINAMIC_SUPPLEMENTAL_GROUPS = 'true';
```
romayalon, line 825:
```suggestion
process.env.NSFS_ENABLE_DYNAMIC_SUPPLEMENTAL_GROUPS = 'true';
```

src/native/util/os_darwin.cpp
```diff
@@ -80,7 +80,8 @@ static void
 set_supplemental_groups(uid_t uid, gid_t gid, std::vector<gid_t>& groups) {
     //first check if groups were defined in the account configuration
     if (groups.empty()) {
-        if (get_supplemental_groups_by_uid(uid, groups) < 0) {
+        const char* is_enabled = getenv("NSFS_ENABLE_DINAMIC_SUPPLEMENTAL_GROUPS");
+        if ((is_enabled == NULL) || (strcmp(is_enabled, "false") == 0) || get_supplemental_groups_by_uid(uid, groups) < 0) {
```
romayalon, line 84: I think we want here explicit not "true", this way if someone accidentally puts 'false1' we will still use it

config.js
romayalon: can you set it to be a regular config? we don't set here envs

config.js
```diff
@@ -822,6 +822,8 @@ config.NSFS_UPDATE_ISSUES_REPORT_ENABLED = true;
 config.NSFS_EXIT_EVENTS_TIME_FRAME_MIN = 24 * 60; // per day
 config.NSFS_MAX_EXIT_EVENTS_PER_TIME_FRAME = 10; // allow max 10 failed forks per day
```
romayalon, line 823: please update NonContainerizedDeveloperCustomizations.md file with this new config and maybe also the doc where you wrote about supplemental groups last PR

config.js
nadavMiz: yes, from what I could see we currently, only override env, when there is a configuration that overrides it. I can add a section that convert default configurations as well. also I saw that there is an env file. but looks like its used only for GPFS

romayalon: hmm not only for GPFS, currently a custom config.json is used by NC
but I think that default should work with these lines - 
``` 
       Object.keys(merged_config).forEach(function(key) {
            const config_to_env = ['NOOBAA_LOG_LEVEL', 'UV_THREADPOOL_SIZE', 'GPFS_DL_PATH'];
            if (config_to_env.includes(key)) {
                process.env[key] = merged_config[key];
                return;
            }
            config[key] = merged_config[key];
        });
```
can you please check?

config.js
```diff
@@ -1134,6 +1136,16 @@ function validate_nc_master_keys_config(config_data) {
     }
 }
 
+function set_nc_nsfs_config_to_env(config_obj) {
```
romayalon, line 1139: please make sure to add jsdoc, it helps the linter to find issues

romayalon, line 1139: please remove the nsfs from the funsion name, we use nc as the name of the deployment type and were asked to remove `nsfs` usage in general on nc related files/functions/vars etc
in NSFS_ENABLE_DYNAMIC_SUPPLEMENTAL_GROUPS it's ok to use because it's a feature used by both deployment types

config.js
```diff
@@ -1153,12 +1165,8 @@ function load_nsfs_nc_config() {
         const node_config = config_data.host_customization?.[node_name];
         const merged_config = _.merge(shared_config, node_config || {});
 
+        set_nc_nsfs_config_to_env(merged_config);
         Object.keys(merged_config).forEach(function(key) {
-            const config_to_env = ['NOOBAA_LOG_LEVEL', 'UV_THREADPOOL_SIZE', 'GPFS_DL_PATH'];
-            if (config_to_env.includes(key)) {
```
romayalon, line 1158: when taking it out of this loop, you iterate the merged_config twice, is there a real reason to do so?

config.js
```diff
@@ -1218,6 +1226,7 @@ function _validate_executable(file_name) {
 module.exports.load_nsfs_nc_config = load_nsfs_nc_config;
 module.exports.reload_nsfs_nc_config = reload_nsfs_nc_config;
 
+set_nc_nsfs_config_to_env(config);
```
romayalon, line 1229: if it's called from load_nsfs_nc_config(), why doing it here?

nadavMiz: the was a misunderstanding from my part. I thought this function was called whenever we reload the config. and I didn't want to go over the entire config object each time. but now I see there is the reload_nsfs_nc_config for that

romayalon (approved):

PR #8774, opened 2025-02-06
NC | CLI | Fix for unsettable flags issues

romayalon:
### Explain the changes
This PR fixes 3 issues with CLI unsettable flags and a tiny test rename refactoring - 
1. We checked if a flag can be unsettable only if the expected type of the flag value and the actual type of value didn't match, but this is not the case when the flag type and value's type match - we should always check it.
2. The list `LIST_UNSETABLE_OPTIONS` was missing some flags that were supposed to be included in it and be checked, their check passed because their type was string (See bullet 1).
The missing options were new_buckets_path (was missing also in the help) and supplemental_groups.
3. The list `LIST_UNSETABLE_OPTIONS` consisted s3_policy flag which does not exist, changed it to bucket_policy.
4. Renamed `test_nc_nsfs_account_cli.test.js -> test_nc_account_cli.test.js` and it's references.

** Found this bug during the reviewing of - https://github.com/noobaa/noobaa-core/pull/8722, hopefully it can help @achouhan09 with the comment https://github.com/noobaa/noobaa-core/pull/8722#discussion_r1944544541

1. 

### Testing Instructions:
1. 

- [x] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/manage_nsfs/manage_nsfs_constants.js
+++ b/src/manage_nsfs/manage_nsfs_constants.js
@@ -164,8 +164,21 @@ const BOOLEAN_STRING_VALUES = ['true', 'false'];
 const BOOLEAN_STRING_OPTIONS = new Set(['allow_bucket_creation', 'regenerate', 'wide', 'show_secrets', 'force',
     'force_md5_etag', 'iam_operate_on_root_account', 'all_account_details', 'all_bucket_details', 'anonymous']);
 
-//options that can be unset using ''
-const LIST_UNSETABLE_OPTIONS = ['fs_backend', 's3_policy', 'force_md5_etag'];
+// CLI UNSET VALUES
+const CLI_EMPTY_STRING = '';
+const CLI_EMPTY_STRING_ARRAY = '[]';
+
+const CLI_EMPTY_VALUES = new Set([CLI_EMPTY_STRING, CLI_EMPTY_STRING_ARRAY]);
+
+// options that can be unset using '' / '[]'
+const UNSETTABLE_OPTIONS_OBJ = Object.freeze({
+    'fs_backend': CLI_EMPTY_STRING,
+    'bucket_policy': CLI_EMPTY_STRING,
+    'force_md5_etag': CLI_EMPTY_STRING,
+    'supplemental_groups': CLI_EMPTY_STRING,
+    'new_buckets_path': CLI_EMPTY_STRING,
+    'ips': CLI_EMPTY_STRING_ARRAY,
+});
 
 const LIST_ACCOUNT_FILTERS = ['uid', 'gid', 'user', 'name', 'access_key'];
 const LIST_BUCKET_FILTERS = ['name'];
@@ -181,7 +194,10 @@ exports.OPTION_TYPE = OPTION_TYPE;
 exports.FROM_FILE = FROM_FILE;
 exports.BOOLEAN_STRING_VALUES = BOOLEAN_STRING_VALUES;
 exports.BOOLEAN_STRING_OPTIONS = BOOLEAN_STRING_OPTIONS;
-exports.LIST_UNSETABLE_OPTIONS = LIST_UNSETABLE_OPTIONS;
+exports.UNSETTABLE_OPTIONS_OBJ = UNSETTABLE_OPTIONS_OBJ;
+exports.CLI_EMPTY_VALUES = CLI_EMPTY_VALUES;
+exports.CLI_EMPTY_STRING = CLI_EMPTY_STRING;
+exports.CLI_EMPTY_STRING_ARRAY = CLI_EMPTY_STRING_ARRAY;
 
 exports.LIST_ACCOUNT_FILTERS = LIST_ACCOUNT_FILTERS;
 exports.LIST_BUCKET_FILTERS = LIST_BUCKET_FILTERS;
--- a/src/manage_nsfs/manage_nsfs_help_utils.js
+++ b/src/manage_nsfs/manage_nsfs_help_utils.js
@@ -98,7 +98,7 @@ Usage:
 
 Flags:
 
-    --ips <string>                                          Set configuring for allowed IP addresses; format: '["127.0.0.1", "192.0.10.0", "3002:0bd6:0000:0000:0000:ee00:0033:6778"]'
+    --ips <string>                                          Set configuring for allowed IP addresses; format: '["127.0.0.1", "192.0.10.0", "3002:0bd6:0000:0000:0000:ee00:0033:6778"]', unset using '[]'
 
 `;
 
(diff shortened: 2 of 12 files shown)


src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -192,6 +193,11 @@ function validate_options_type_by_value(input_options_with_data) {
             const details = `type of flag ${option} should be ${type_of_option} (and the received value is ${value})`;
             throw_cli_error(ManageCLIError.InvalidArgumentType, details);
         }
+        // special case for unset value (specified by '').
+        if (!LIST_UNSETTABLE_OPTIONS.includes(option) && value === '') {
```
achouhan09, line 197: This condition would also works/validates the leading space check for the flags and we can put the error msg accordingly

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -192,6 +193,11 @@ function validate_options_type_by_value(input_options_with_data) {
...
+            const details = `flag ${option} is '' ${value} but it can't be unset.`;
+            throw_cli_error(ManageCLIError.UnsetArgumentIsInvalid, details);
+        }
```
achouhan09, line 200: Still we are missing one check here, we are not checking leading space for unsettable options and faulty values assigned to unsettable flags due to the leading space.

Please check for below command(`new_buckets_path` will get assigned empty value here):

`$ sudo node src/cmd/manage_nsfs account add --name aayush2 --uid=9 --gid=9  --new_buckets_path= tmp/`

romayalon: I was not aiming to fix the leading spaces issue, please fix that on your PR

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -192,6 +193,11 @@ function validate_options_type_by_value(input_options_with_data) {
             const details = `type of flag ${option} should be ${type_of_option} (and the received value is ${value})`;
             throw_cli_error(ManageCLIError.InvalidArgumentType, details);
         }
+        // special case for unset value (specified by '').
+        if (!LIST_UNSETTABLE_OPTIONS.includes(option) && value === '') {
+            const details = `flag ${option} is '' ${value} but it can't be unset.`;
```
shirady, line 198: The `${value}`is '', and you already printed it, I'm not sure you need it.
```suggestion
            const details = `flag value of ${option} is '' but this option can't be unset.`;
```

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -22,6 +22,7 @@ const tmp_fs_path = path.join(TMP_PATH, 'test_nc_nsfs_account_cli');
 const timeout = 50000;
 const config = require('../../../../config');
 const config_fs_account_options = { show_secrets: true, decrypt_secret_key: true };
+const empty_string = "''";
```
shirady, line 25: You add this in the file `manage_nsfs_constants.js` and we reuse it there and in the tests.

src/test/unit_tests/test_nc_cli.js
```diff
@@ -991,16 +991,25 @@ mocha.describe('manage_nsfs cli', function() {
             assert_whitelist(config_data, config_options);
         });
 
-        mocha.it('cli whitelist ips is empty', async function() {
+        mocha.it('should fail - cli whitelist ips is empty', async function() {
```
shirady, line 994: I would expand the `it` and add a comment so it will be clear.
```suggestion
        mocha.it('should fail - cli whitelist ips is an empty string', async function() {
        // we allow unset the ips list with empty array '[]' and not an empty string ''
```

src/test/unit_tests/test_nc_cli.js
```diff
@@ -991,16 +991,25 @@ mocha.describe('manage_nsfs cli', function() {
...
             try {
                 await exec_manage_cli(type, '', { config_root, ips: '' });
                 assert.fail('should have failed with whitelist ips should not be empty.');
             } catch (err) {
-                assert_error(err, ManageCLIError.InvalidWhiteListIPFormat);
+                assert_error(err, ManageCLIError.UnsetArgumentIsInvalid);
```
shirady, line 999: I think the error we're throwing here is confusing `UnsetArgumentIsInvalid` ("Argument can not be unset, its value must be set or check there are no leading spaces") vs `InvalidWhiteListIPFormat` ("Whitelist IP body format is invalid") that we used to have, as the option `--ips` can be unset using `[]` and not `''`.

src/test/unit_tests/test_nc_cli.js
```diff
@@ -991,16 +991,25 @@ mocha.describe('manage_nsfs cli', function() {
...
+        mocha.it('cli whitelist ips is empty array', async function() {
+            const ips = [];
+            const res = await exec_manage_cli(type, '', { config_root, ips: JSON.stringify(ips) });
+            config_options.S3_SERVER_IP_WHITELIST = ips;
+            const config_data = await config_fs.get_config_json();
+            await assert_response('', type, res, ips);
+            assert_whitelist(config_data, config_options);
+        });
+
+        mocha.it('should fail - cli whitelist formate is invalid', async function() {
```
shirady, line 1012: We have a small typo (it was there before your changes)
```suggestion
        mocha.it('should fail - cli whitelist format is invalid', async function() {
```

src/manage_nsfs/manage_nsfs_help_utils.js
shirady, line 1: We would need to update those changes in the docs as well in the file `NooBaaCLI.md`

src/test/unit_tests/sudo_index.js
shirady, line 1: Would you prefer to move the renaming to another PR?
It would make the changes here more clear.
I need to create a PR with some changes, I can also add these.

romayalon: no, it's clear enough, thanks

src/manage_nsfs/manage_nsfs_constants.js
```diff
@@ -165,7 +165,7 @@ const BOOLEAN_STRING_OPTIONS = new Set(['allow_bucket_creation', 'regenerate', '
     'force_md5_etag', 'iam_operate_on_root_account', 'all_account_details', 'all_bucket_details', 'anonymous']);
 
 //options that can be unset using ''
-const LIST_UNSETABLE_OPTIONS = ['fs_backend', 's3_policy', 'force_md5_etag'];
+const LIST_UNSETTABLE_OPTIONS = ['fs_backend', 'bucket_policy', 'force_md5_etag', 'supplemental_groups', 'new_buckets_path'];
```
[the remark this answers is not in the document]
romayalon: yes, this is the correct spelling

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
romayalon: I don't see where it can be re-used in this form

src/test/unit_tests/test_nc_cli.js
romayalon: sure, do you have a better suggestion?

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
shirady: ok, so it is just in the test files (account. bucket).
I would rename it with uppercase, for example:
```suggestion
const UNSET_WITH_EMPTY_STRING = "''";
```

src/test/unit_tests/test_nc_cli.js
shirady: I think the simple one will be to add the `ips` to the `LIST_UNSETTABLE_OPTIONS` (and you can add a comment there), and we will end with the same error that we had before.

romayalon: @shirady I eventually thought that we need to still throw the same error code but I added more details and a resolution
> noobaa-cli whitelist --ips ''
{
  "error": {
    "code": "UnsetArgumentIsInvalid",
    "message": "Argument can not be unset or it was unset incorrectly, its value must be set correctly or check there are no leading spaces",
    "detail": "flag value of ips is '' but this option can't be unset via ''. Please use '[]' instead."
  }
}

src/manage_nsfs/manage_nsfs_help_utils.js
shirady, line 1: I think in the file "NooBaaCLI.md" you added more cases of unset with '' that were not updated here.

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -23,6 +23,7 @@ const timeout = 50000;
 const config = require('../../../../config');
 const config_fs_account_options = { show_secrets: true, decrypt_secret_key: true };
 
+
```
shirady, line 26: I think it is redundant newline

shirady (approved):
LGTM

added a few minor comments

src/manage_nsfs/manage_nsfs_help_utils.js
romayalon: please add what you think should be added, I added to update action only on purpose

shirady: In the update section is fine, I just thought we're not covered at all the places (for example - `fs_backend`) in the help.

PR #8781, opened 2025-02-10
NC | CLI | List Accounts When Decrypt Access Keys Fails

shirady:
### Explain the changes
1. In account list add additional option of `return_encrypted_on_decryption_error` and we will neither return the `secret_key` nor the `encrypted_secret_key` in case of failure of access keys decryption.
2. In the function `get_identity_config_data` add a case in  try-catch clause for the decryption failure and add the property of `decryption_err` in case there was failure of access keys decryption that is related to `INVALID_MASTER_KEY` and remove the `encrypted_secret_key` property (added a helper function for that).
3. Small refactors in file `test_nc_account_invalid_mkm_integration.test.js` so we can add additional account and rename the file `master_keys.json` for the test.

### Issues:
1. It is a partial fix related to issue #8104 - this fix only handles the list account CLI with decrypt issue.

#### Automatic Tests:
Please run:
- `sudo npx jest test_nc_account_invalid_mkm_integration.test.js -t 'cli with renamed master key file'`
- `npx jest test_config_fs.test.js -t 'remove_encrypted_secret_key'`

#### Manual Tests
3. Make the master key invalid by renaming the file `master_key.json`: `sudo mv /etc/noobaa.conf.d/master_keys.json /etc/noobaa.conf.d/temp_master_keys.json`
4. Account list with decryption: `sudo node src/cmd/manage_nsfs account list --wide --show_secrets` (should not fail, but you will see the property `encrypted_secret_key` and `decryption_err`).
5. Account status with decryption: `sudo node src/cmd/manage_nsfs account status --name <account-name> --show_secrets` (will fail - we didn't handle this case).

Before this fix we had this error (added partial output):
```
  "error": {
    "code": "InvalidMasterKey",
    "message": "Master key manager had issues loading master key, can not decrypt/encrypt secrets.",
    "cause": "Error: master key id is missing in master_keys_by_id\n    at NCMasterKeysManager._validate_master_key_manager ...
...
  }
```

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -597,6 +597,10 @@ async function list_account_config_files(wide, show_secrets, filters = {}) {
     const options = {
         show_secrets: show_secrets || should_filter,
         decrypt_secret_key: show_secrets,
+        // in case we have an error on secret_key decryption 
+        // we will neither return the secret_key nor the encrypted_secret_key
+        // and add the property of decryption_err with the error we had
+        return_on_decryption_error: true,
         silent_if_missing: true
     };
 
--- a/src/sdk/config_fs.js
+++ b/src/sdk/config_fs.js
@@ -231,24 +231,35 @@ class ConfigFS {
      * and decrypts the account's secret_key if decrypt_secret_key is true
      * if silent_if_missing is true -      
      *   if the config file was deleted (encounter ENOENT error) - continue (returns undefined)
+     * if decrypt_secret_key is true and the decryption failed with rpc error of INVALID_MASTER_KEY
+     *    and return_on_decryption_error is true - 
+     *        add a property decryption_err with the error we've got
+     *        if  return_on_decryption_error is false - throw the error (as it was before)
      * @param {string} config_file_path
-     * @param {{show_secrets?: boolean, decrypt_secret_key?: boolean, silent_if_missing?: boolean}} [options]
+     * @param {{show_secrets?: boolean, decrypt_secret_key?: boolean, silent_if_missing?: boolean,
+     *          return_on_decryption_error?: boolean}} [options]
      * @returns {Promise<Object>}
      */
     async get_identity_config_data(config_file_path, options = {}) {
         const { show_secrets = false, decrypt_secret_key = false, silent_if_missing = false } = options;
+        let config_data;
         try {
             const data = await this.get_config_data(config_file_path, options);
             if (!data && silent_if_missing) return;
-            const config_data = _.omit(data, show_secrets ? [] : ['access_keys']);
+            config_data = _.omit(data, show_secrets ? [] : ['access_keys']);
             if (decrypt_secret_key) config_data.access_keys = await nc_mkm.decrypt_access_keys(config_data);
             return config_data;
         } catch (err) {
             dbg.warn('get_identity_config_data: with config_file_path', config_file_path, 'got an error', err);
+            if (err.rpc_code === 'INVALID_MASTER_KEY' && options.return_on_decryption_error) {
+                config_data.decryption_err = err.message;
+                return this.remove_encrypted_secret_key(config_data);
+            }
             if (err.code === 'ENOENT' && silent_if_missing) return;
             throw err;
         }
     }
+
     /**
      * get_config_data reads a config file and returns its content
      * @param {string} config_file_path
@@ -820,7 +831,6 @@ class ConfigFS {
         await native_fs_utils.folder_delete(account_dir_path, this.fs_context, undefined, true);
     }
 
-
     /**
      * _prepare_for_account_schema processes account data before writing it to the config dir and does the following -
      * 1. encrypts its access keys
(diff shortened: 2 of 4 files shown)


src/sdk/config_fs.js
```diff
@@ -241,7 +246,18 @@ class ConfigFS {
             const data = await this.get_config_data(config_file_path, options);
             if (!data && silent_if_missing) return;
             const config_data = _.omit(data, show_secrets ? [] : ['access_keys']);
-            if (decrypt_secret_key) config_data.access_keys = await nc_mkm.decrypt_access_keys(config_data);
+            if (decrypt_secret_key) {
+                try {
```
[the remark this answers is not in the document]
shirady: I moved it, thanks.

src/sdk/config_fs.js
```diff
@@ -241,7 +246,18 @@ class ConfigFS {
...
+                    config_data.access_keys = await nc_mkm.decrypt_access_keys(config_data);
+                } catch (err) {
+                    dbg.warn('get_identity_config_data: with config_file_path', config_file_path, 'got an error', err);
+                    if (err.rpc_code === 'INVALID_MASTER_KEY' && options.return_encrypted_on_decryption_error) {
```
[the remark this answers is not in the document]
shirady: > here we are pushing the access_keys to the console, right?

Do you mean the warn printing? no, it's printing the path.
I didn't understand what exactly to check.

src/cmd/manage_nsfs.js
```diff
@@ -597,6 +597,10 @@ async function list_account_config_files(wide, show_secrets, filters = {}) {
     const options = {
         show_secrets: show_secrets || should_filter,
         decrypt_secret_key: show_secrets,
+        // in case we have an error on secret_key decryption 
+        // we will return the encrypted_secret_key
```
romayalon, line 601: I wouldn't return the encrypted secret key, why do you want to return it to the user?

src/test/unit_tests/jest_tests/test_nc_account_invalid_mkm_integration.test.js
```diff
@@ -269,11 +306,34 @@ function fail(reason) {
...
+
+/**
+ * master_key_file_rename will rename the master_keys.json file
+ * to mock a situation where master_key_id points to a missing master key
+ * use the to_rename_temp false to rename it back (after the test)
+ * @param {boolean} to_rename_temp
+ */
+async function master_key_file_rename(to_rename_temp) {
+    const default_fs_config = get_process_fs_context();
+    const source_path = path.join(config_root, 'master_keys.json');
```
romayalon, line 331: would it be easier to change the value of config.NC_MASTER_KEYS_FILE_LOCATION?

src/cmd/manage_nsfs.js
shirady: I thought it was another indication for the user that we had a decryption issue.
I changed it to remove this property, thanks.

src/test/unit_tests/jest_tests/test_nc_account_invalid_mkm_integration.test.js
shirady: From what I tried, no - the `config.NC_MASTER_KEYS_FILE_LOCATION` sis an empty string and the master key file was already written, so I need to rename the existing one.

src/sdk/config_fs.js
```diff
@@ -839,6 +849,20 @@ class ConfigFS {
         return { parsed_account_data, string_account_data };
     }
 
+    /**
+     * remove_encrypted_secret_key will remove the encrypted_secret_key property from an identity
+     * since we allow to have more than one access key, we hold the index (but most cases are with 1)
+     * @param {object} config_data
```
romayalon, line 855: please add returns, it helps the linter find issues

src/cmd/manage_nsfs.js
```diff
@@ -597,6 +597,10 @@ async function list_account_config_files(wide, show_secrets, filters = {}) {
     const options = {
         show_secrets: show_secrets || should_filter,
         decrypt_secret_key: show_secrets,
+        // in case we have an error on secret_key decryption 
+        // we will neither return the secret_key nor the encrypted_secret_key
+        // and also decryption_err with the error we had
+        return_encrypted_on_decryption_error: true,
```
romayalon, line 603: var name is irrelevant because you no longer return encrypted

src/sdk/config_fs.js
```diff
@@ -839,6 +849,20 @@ class ConfigFS {
...
+     */
+    remove_encrypted_secret_key(config_data) {
+        let index = 0;
+        for (const access_key of config_data.access_keys) {
+            config_data.access_keys[index] = _.omit(access_key, ['encrypted_secret_key']);
+            index += index;
+        }
+        return config_data;
```
romayalon, line 863: it's not really clear what is going on here with the index, especially the line `index += index;`
why not use a regular for (const i=0; ....) instead

src/sdk/config_fs.js
shirady: Added, thanks.

src/sdk/config_fs.js
shirady: Changed, thanks.
I thought it might be confusing with the standard loop.

romayalon (approved):

src/sdk/config_fs.js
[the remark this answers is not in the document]
shirady: @naveenpaul1 please let me know if this comment was in the scope of this PR code changes or in general.

romayalon:
@shirady So this is a partial fix of https://github.com/noobaa/noobaa-core/issues/8104, right?
only for list + decryption issue

shirady:
> @shirady So this is a partial fix of #8104, right? only for list + decryption issue

@romayalon , yes, I fixed the wrong link in the PR description, thanks.

Issue #7734, opened 2024-01-18
NSFS | NC | Bucket owner & Principal of bucket policies should be ids and not email/name


romayalon:
### Environment info
- NooBaa Version: master
- Platform: NC

### Actual behavior
1. Bucket policy principal identified by account name. An account with a specific name can be deleted. In the future someone can create a different account with the same specific name, and the bucket policy will refer to the new account. This might confuse or even cause a security issue.
2. bucket_owner identified by account email, although the security issue from bullet 1 couldn't happen because we are blocking deletion of accounts that own buckets. This property should be considered as well to be deprecated and on some flows, we should use owner_account which is an id instead.

### Expected behavior
1. foreign keys between bucket and account should be random ids.

### Steps to reproduce
Bucket policy issue - 
1. create an account, let's call it account1.
2. create a bucket while specifying a bucket policy that allows account1 to run all s3 commands on this bucket.
3. delete account1.
4. create a new account having the same name.
5. run s3 commands using the new account identified by the name account1 without access denial issues.

shirady:
When someone is assigned to this issue, please also add a test for the scenario mentioned in #7797

romayalon:
Updating for stale bot, still not fixed.
Will be addressed soon as part of the IAM effort.

nimrod-becker:
keep alive, probably related to the fixes for IAM

shirady:
Updating the status:
What we did:
1. In NC we added the option to add the account ID as a principle in PR #8280
2. We also removed the bucket owner property (as a name) #8289
3. Started a design discussion about the security issue raised in this issue and how we suggest solving it in NC and Containerized environment (will open a Jira ticket for this) - which still open.

PR #8782, opened 2025-02-10
Account with NotPrincipal should explicitly provide permission

vh05:
This is the bucket policy with below combination
```
    
    Effect : DENY,
    Action: $OPERATION,
    NotPrincipal: $ACCOUNT

``` 
The permission on the "Action" for the $ACCOUNT mentioned in the "NotPrincipal" should be explicitly given.
    
Example: For "Action: get_object", if we want to DENY permission for * (all accounts) and we want to give explicit permission to any one or few accounts, then we can include that account as part of part of "NotPrincipal"
    
Check DFBUGS-1519 for steps to reproduce the issue
    
Fixes: https://issues.redhat.com/browse/DFBUGS-1519

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -293,16 +293,19 @@ async function authorize_request_policy(req) {
             s3_policy, account_identifier_id, method, arn_path, req, public_access_block_cfg?.public_access_block?.restrict_public_buckets);
         dbg.log3('authorize_request_policy: permission_by_id', permission_by_id);
     }
-    if (permission_by_id === "DENY") throw new S3Error(S3Error.AccessDenied);
 
-    if ((!account_identifier_id || permission_by_id !== "DENY") && account.owner === undefined) {
+    if (permission_by_id === "ALLOW") return;
+
+    if ((!account_identifier_id || permission_by_id === "DENY" || permission_by_id === "IMPLICIT_DENY") && account.owner === undefined) {
         permission_by_name = await s3_bucket_policy_utils.has_bucket_policy_permission(
             s3_policy, account_identifier_name, method, arn_path, req, public_access_block_cfg?.public_access_block?.restrict_public_buckets
         );
         dbg.log3('authorize_request_policy: permission_by_name', permission_by_name);
     }
+
     if (permission_by_name === "DENY") throw new S3Error(S3Error.AccessDenied);
-    if ((permission_by_id === "ALLOW" || permission_by_name === "ALLOW") || is_owner) return;
+    // Allow access if either ID or Name has ALLOW permission, or if the user is the owner
+    if (permission_by_id === "ALLOW" || permission_by_name === "ALLOW" || is_owner) return;
 
     throw new S3Error(S3Error.AccessDenied);
 }


shirady:
@vh05 please add tests in the file "test_s3_bucket_policy.js"

src/endpoint/s3/s3_rest.js
```diff
@@ -287,14 +287,14 @@ async function authorize_request_policy(req) {
     // we start the permission check on account identifier intentionally
     if (account_identifier_id) {
         permission_by_id = await s3_bucket_policy_utils.has_bucket_policy_permission(
-            s3_policy, account_identifier_id, method, arn_path, req);
+            s3_policy, {account_identifier_name, account_identifier_id}, method, arn_path, req);
```
shirady, line 290: The account argument is a ID/name (string), not an object.
Continue with the flow, and you will see in `_is_principal_fit` the check of `principal.unwrap() === account`. You can add logs to better understand where you are (before the fix).

This check is only in NC and it checks the principal by ID (see "if (account_identifier_id) {").

shirady:
@vh05 I would also start with testing on AWS the steps that are expected to happen before we change it our logs.

Please also add the full testing instructions that you manually did so we can understand what was done (you can add as comment if you prefer).

src/endpoint/s3/s3_rest.js
vh05: @shirady any reason why we are passing id instead of name ? However the account name will be unique and we can use the name instead of id

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -174,8 +174,8 @@ function _is_principal_fit(account, statement) {
     let principal_fit = false;
     statement_principal = statement_principal.AWS ? statement_principal.AWS : statement_principal;
     for (const principal of _.flatten([statement_principal])) {
-        dbg.log1('bucket_policy: ', statement.Principal ? 'Principal' : 'NotPrincipal', ' fit?', principal, account);
-        if ((principal.unwrap() === '*') || (principal.unwrap() === account)) {
+        dbg.log1('bucket_policy: ', statement.Principal ? 'Principal' : 'NotPrincipal', ' fit?', principal, account.account_identifier_name);
```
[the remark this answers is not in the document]
vh05: I thought to pass it as object, now I am passing it as the seperate argument.

src/endpoint/s3/s3_rest.js
shirady: It is an additional way that we allow to add Principals, you can see in the [S3Ops.md](https://github.com/noobaa/noobaa-core/blob/master/docs/NooBaaNonContainerized/S3Ops.md#principal-field), the reason was to have it like in AWS for accounts, and the initial thought that was that t is part of solving the issue about account names - see issue #7734

src/endpoint/s3/s3_bucket_policy_utils.js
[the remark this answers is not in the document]
shirady: It may not be necessary, and we also have a plan to pass the Principal as ARN to support future IAM - I can share the Jira with details ([link](https://issues.redhat.com/browse/MCGI-279)) if you want to understand the future steps that might be on these utils.
I would rethink on it, anyway, keep in mind that it should work for containerized and NC.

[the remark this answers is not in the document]
vh05: @shirady Should we continue in this direction and make NC compatible with ID or we take ARN approach ? According to [this](https://github.com/noobaa/noobaa-core/issues/7734), account name is an issue.

src/endpoint/s3/s3_rest.js
vh05: So we should have account id work for NC as well rather than going for Account name ?

nadavMiz:
@vh05 from what I understand this issue is not related to "deny statement using NotPrincipal". can you pleas change the PR name and description to reflect what you are actually doing here

src/endpoint/s3/s3_bucket_policy_utils.js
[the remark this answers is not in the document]
shirady: I don't think you need to change this, just fix the issue in the link.
The issue about the names that you attached the link known and we have a solution for it - you can see the design of it here ([MCGI-281](https://issues.redhat.com/browse/MCGI-281)), but I don't think it is relevant.

src/endpoint/s3/s3_rest.js
shirady: No, we support both options. We support the account name for backward compatibility.

vh05:
> @vh05 from what I understand this issue is not related to "deny statement using NotPrincipal". can you pleas change the PR name and description to reflect what you are actually doing here

Sorry for the delayed response. I have updated the patch. Now the fix and the issue description matches I think

src/test/unit_tests/test_s3_bucket_policy.js
shirady, line 1: What about fixing the test `should be able to use notPrincipal` it is in the scope of this PR?

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -146,18 +146,31 @@ async function _is_object_version_fit(req, predicate, value) {
     return res;
 }
 
-async function has_bucket_policy_permission(policy, account, method, arn_path, req) {
+async function has_bucket_policy_permission(policy, account, method, arn_path, req, account_identifier_type = 'id') {
```
shirady, line 149: Why did you choose to have id as default?
the id is currently only in NC.
Please add JSDoc to this function or a comment about the options of `account_identifier_type`. I think using a enum can help us.

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -189,7 +201,23 @@ function _is_principal_fit(account, statement) {
             break;
         }
     }
-    return statement.Principal ? principal_fit : !principal_fit;
+
+    if (statement.NotPrincipal) {
```
shirady, line 205: So after this change, the previous code is only relevant for `statement.Principal`?
If yes, then why did you keep this `let statement_principal = statement.Principal || statement.NotPrincipal;`?

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -189,7 +201,23 @@ function _is_principal_fit(account, statement) {
...
+        // If the account is categorized as "Not Principal," we need to ALLOW operations
+        // for that account under the associated principal.
+        //
+        // To maintain backward compatibility, we also support name-based comparison for principal matching.
+        //
+        // - If `account_identifier_type` is "id" but the `account` parameter contains a name,
+        //   we return `false` to proceed/bypass to the next step, where we check for a name match.
+        //
+        // - In the next step, if `account_identifier_type` is "name" and the `account` also contains a name,
+        //   we return the actual value.
```
shirady, line 215: What if the allow is by ids and the deny is by names?
You can add it a test case.
What will happen when we support ARN, it will be more complex.

shirady (commented):
Added a couple of comments and questions.

src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -1647,6 +1647,37 @@ mocha.describe('s3_bucket_policy', function() {
         }));
     });
 
+    mocha.it.skip('should be able to use notPrincipal with Effect Deny', async function() {
```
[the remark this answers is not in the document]
vh05: Copy paste error

src/test/unit_tests/test_s3_bucket_policy.js
```diff
@@ -1647,6 +1647,37 @@ mocha.describe('s3_bucket_policy', function() {
...
+            Policy: JSON.stringify(auth_put_policy)
+        });
+
+        await s3_b.putObject({
+            Body: BODY,
+            Bucket: BKT,
+            Key: KEY,
+        });
+
+        await assert_throws_async(s3_a.putObject({
```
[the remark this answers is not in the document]
vh05: The account added in `NotPrincipal` is allowed on the `Action`. It is like excluding everyone other than  account added as `NotPrincipal`

src/test/unit_tests/test_s3_bucket_policy.js
vh05: `NotPrincipal` with `Allow` is not expected operation as per aws doc 

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_notprincipal.html 

from AWS:
`NotPrincipal must be used with "Effect":"Deny". Using it with "Effect":"Allow" is not supported. `

src/endpoint/s3/s3_bucket_policy_utils.js
vh05: IIUC, the id is considered to solve issues regarding account name and thats the updated one. We are giving exception to `name` parameter.

> I think using a enum can help us
I feel it is for default argument, string is fine

src/endpoint/s3/s3_bucket_policy_utils.js
vh05: If you see `statement_principal` is has in general considered to iterate through it irrespective of `Principal` or `NotPrincipal`. It is not required to be changes I feel

src/endpoint/s3/s3_bucket_policy_utils.js
shirady: As you can see in the [Jira ticket](https://issues.redhat.com/browse/MCGI-281), it is not solving it completely.
I didn't understand the quote; was it not clear? I meant to add enum so it will be easier to understand the options that we can use.

src/endpoint/s3/s3_bucket_policy_utils.js
shirady: The name is another option, we decided to not remove this to be backward compatible.
Anyway, it is a valid option, and in the containerized it is the only option.

src/test/unit_tests/test_s3_bucket_policy.js
shirady: I agree, it is something that would be fixed in your suggested fix?

src/test/unit_tests/test_s3_bucket_policy.js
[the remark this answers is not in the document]
shirady: When you use `assert_throws_async` it is expected to throw an error.
As you mentioned the `user_a` is allowed on the action, then you probably need to run it without this function on `s3_a` and can run it on another user.

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -189,7 +201,23 @@ function _is_principal_fit(account, statement) {
             break;
         }
     }
-    return statement.Principal ? principal_fit : !principal_fit;
+
+    if (statement.NotPrincipal) {
+        // If the account is categorized as "Not Principal," we need to ALLOW operations
```
nadavMiz, line 206: have you looked at https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_notaction.html?
note that there is an implicit Deny for actions that does not any statement. combination of Deny with NotPrincipal is not the same as Allow

src/test/unit_tests/test_s3_bucket_policy.js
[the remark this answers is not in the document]
vh05: Yeah. That expression should be with user_b not user_a. My bad

PR #7110, opened 2022-11-24
Pre-compiled regexp optimization

baum:
# Pre-compiled regexp optimization

Improve performance by precompiling regular expressions that are used repeatedly.

### COSBench

|Op-Type	|Op-Count	|Byte-Count	|Avg-ResTime	|Avg-ProcTime	|Throughput	|Bandwidth	|Succ-Ratio | 
|---|---|---|---|---|---|---|---|
|BASELINE op1: read	| 294.63 kops	| 2.95 GB	| 232.11 ms	| 232.11 ms	| 327.42 op/s	| 3.27 MB/S	| 100%
| PR op1: read	|339.79 kops	|3.4 GB	|201.28 ms	|201.28 ms	|377.58 op/s	|3.78 MB/S	|100%

https://bugzilla.redhat.com/show_bug.cgi?id=2106742

Signed-off-by: Alexander Indenbaum <aindenba@redhat.com>

PR #8783, opened 2025-02-12
[MCG] Using put-bucket-policy with wrong syntax under Resource results in InternalError instead of MalformedPolicy

vh05:
The malformed syntax should give malformed systax error.

Issue: Square brackets ([ ]) in resource_bucket_part were misinterpreted in regex.

Fix: Escape all regex special characters before inserting into RegExp().

Fixes: https://issues.redhat.com/browse/DFBUGS-1517

### DIFF
--- a/src/endpoint/s3/s3_bucket_policy_utils.js
+++ b/src/endpoint/s3/s3_bucket_policy_utils.js
@@ -91,6 +91,7 @@ const OP_NAME_TO_ACTION = Object.freeze({
 
 const qm_regex = /\?/g;
 const ar_regex = /\*/g;
+const esc_regex = /[-/^$+?.()|[\]{}]/g;
 
 const predicate_map = {
     'StringEquals': (request_value, policy_value) => request_value === policy_value,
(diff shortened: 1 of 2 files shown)


romayalon (commented):
Hey @vh05 
1. Please add unit tests
2. Initializing regex in a function is unrecommended due to performance issues and it should be pre-compiled as a global variable, read more about it here - https://github.com/noobaa/noobaa-core/pull/7110
3. To me It looks like this issue can happen in every bucket policy field, can you please check if we need this fix in more places?  
4. @nadavMiz since you are the last one to change this area, I would appreciate your review as well.

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -269,7 +269,12 @@ async function validate_s3_policy(policy, bucket_name, get_account_handler) {
         }
         for (const resource of _.flatten([statement.Resource || statement.NotResource])) {
             const resource_bucket_part = resource.split('/')[0];
-            const resource_regex = RegExp(`^${resource_bucket_part.replace(qm_regex, '.?').replace(ar_regex, '.*')}$`);
+            const resource_regex = RegExp(
+                `^${resource_bucket_part
+                .replace(/[-/^$+?.()|[\]{}]/g, '\\$&')
```
nadavMiz, line 274: notice that we are modifying the ? character right after your change. did you check it doesn't effect this? worth adding a test as well if one doesn't already exists

vh05: @nadavMiz I think the question mark (?) will be replaced with `\?` in the 1st part then \? will be replace by `\.?` . Have limited or no knowledge on this. Checking this

PR #8786, opened 2025-02-12
Add http://{endpoint_address}:{s3_port}/_/version api to the endpoint

romayalon:
### Explain the changes
1. Create an API for receiving the package version of NooBaa via the endpoint.

1. Gap - add a CLI call for that? 
2. move other internal apis to be under /_/

### Testing Instructions:
Happy path manual Test - 
1. Tab 1 - `sudo node noobaa-core/src/cmd/nsfs.js --debug 5`
3. Tab 2 - call the new /_/version api - 
```
> curl http://localhost:6001/_/version
5.19.0%
```

Bad path manual test - 
1. Tab 1 - `sudo node noobaa-core/src/cmd/nsfs.js --debug 5`
3. Tab 2 - call the new /_/version1 api - 
```
> curl http://localhost:6001/_/version1
{"error":"Internal Server Error","message":"Unknown API call version1"}%
```
- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -6,6 +6,7 @@
 require('../util/dotenv').load();
 require('../util/panic');
 require('../util/fips');
+const pkg = require('../../package.json');
 
 const dbg = require('../util/debug_module')(__filename);
 if (!dbg.get_process_name()) dbg.set_process_name('Endpoint');
@@ -294,6 +295,14 @@ function create_endpoint_handler(server_type, init_request_sdk, { virtual_hosts,
                 return fork_count_handler(req, res);
             } else if (req.url.startsWith('/endpoint_fork_id')) {
                 return endpoint_fork_id_handler(req, res);
+            } else if (req.url.startsWith('/_/')) {
+                // internals non S3 requests
+                const api = req.url.slice('/_/'.length);
+                if (api === 'version') {
+                    return version_handler(req, res);
+                } else {
+                    return internal_api_error(req, res, `Unknown API call ${api}`);
+                }
             } else {
                 return s3_rest.handler(req, res);
             }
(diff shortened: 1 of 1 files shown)


src/endpoint/endpoint.js
```diff
@@ -349,6 +355,21 @@ function fork_count_handler(req, res) {
     res.end(JSON.stringify(reply));
 }
 
+/**
+ * version_handler return the version of noobaa package
+ * @param {*} req 
+ * @param {*} res
```
guymguym, line 361: Use these types taken from the EndpointHandler definition -
```suggestion
 *  @param {EndpointRequest} req
 *  @param {import('http').ServerResponse} res
```

src/endpoint/endpoint.js
```diff
@@ -349,6 +355,21 @@ function fork_count_handler(req, res) {
...
+ * @param {*} res 
+ */
+function version_handler(req, res) {
+    const reply = {
+        version: pkg.version,
+    };
+    res.statusCode = 200;
+    res.setHeader('Content-Type', 'application/json');
+    res.setHeader('Content-Length', Buffer.byteLength(JSON.stringify(reply)));
+    res.end(JSON.stringify(reply));
```
guymguym, line 370: When returning a json it would be better to prepare the buffer once and not call stringify twice
```suggestion
    const buffer = Buffer.from(JSON.stringify(reply));
    res.statusCode = 200;
    res.setHeader('Content-Type', 'application/json');
    res.setHeader('Content-Length', buffer.length);
    res.end(buffer);
```

src/endpoint/endpoint.js
```diff
@@ -349,6 +355,21 @@ function fork_count_handler(req, res) {
...
 
+/**
+ * version_handler return the version of noobaa package
+ * @param {*} req 
+ * @param {*} res 
+ */
+function version_handler(req, res) {
+    const reply = {
+        version: pkg.version,
+    };
```
guymguym, line 366: I think a simple api call for "version" should return a single semver string, and not a json reply.
We can add a separate "versions" api call if we will want to return a more complete versions info which is also how nodejs models it:
```
> process.version
'v22.11.0'
> process.versions
{
  node: '22.11.0',
  acorn: '8.12.1',
  ada: '2.9.0',
  amaro: '0.1.8',
  ares: '1.33.1',
  brotli: '1.1.0',
  cjs_module_lexer: '1.4.1',
  cldr: '45.0',
  icu: '75.1',
  llhttp: '9.2.1',
  modules: '127',
  napi: '9',
  nbytes: '0.1.1',
  ncrypto: '0.0.1',
  nghttp2: '1.63.0',
  nghttp3: '0.7.0',
  ngtcp2: '1.3.0',
  openssl: '3.0.15+quic',
  simdjson: '3.10.0',
  simdutf: '5.5.0',
  sqlite: '3.46.1',
  tz: '2024b',
  undici: '6.20.0',
  unicode: '15.1',
  uv: '1.48.0',
  uvwasi: '0.0.21',
  v8: '12.4.254.21-node.21',
  zlib: '1.3.0.1-motley-71660e1'
}
```

src/endpoint/endpoint.js
```diff
@@ -294,6 +295,11 @@ function create_endpoint_handler(server_type, init_request_sdk, { virtual_hosts,
                 return fork_count_handler(req, res);
             } else if (req.url.startsWith('/endpoint_fork_id')) {
                 return endpoint_fork_id_handler(req, res);
+            } else if (req.url.startsWith('/_')) {
+                // internals non S3 requests
+                if (req.url === '/_version') {
+                    return version_handler(req, res);
+                }
```
guymguym, line 302: How about this - I think it is easier to add more "internal api" calls under a common internal path prefix such as `/_/<api>` rather than think of each of those api names as starting with an underscore.
Also need to handle the case of unrecognized api call.
```suggestion
            } else if (req.url.startsWith('/_/')) {
                // internals non S3 requests
                const api = req.url.slice('/_/'.length);
                if (api === 'version') {
                    return version_handler(req, res);
                } else {
                    return internal_api_error(req, res, `Unknown API call ${api}`);
                }
```

guymguym (changes requested):

src/endpoint/endpoint.js
romayalon: sure, I also fixed the places where I copied it from

guymguym (approved):

PR #8787, opened 2025-02-12
Adding ading functions to `net_utils.js` | nope-ip removal - Phase 4

liranmauda:
### Explain the changes

`net_utils.js`

- Adding `normalize_family` that will normalize the family
- Adding `is_loopback` that will check if the address is a loopback address
- Adding `is_private` that will check if the address is private
- Adding `is_public` that will check if the address is public

- Adding jest tests


nadavMiz (approved):
LGTM

Issue #8794, opened 2025-02-13
Connection status is throwing internal server error for non existing connection name


ramya-c3:
### Environment info
NooBaa Version: noobaa-core-5.18.0-20250201.el9.x86_64
Platform: standalone
### Actual behavior
1.# /usr/local/bin/noobaa-cli connection status --name http_notify23 2>/dev/null
{
  "error": {
    "code": "InternalError",
    "message": "The server encountered an internal error. Please retry the request",
    "cause": "Error: No such file or directory"
  }
}

### Expected behavior
1.# /usr/local/bin/noobaa-cli connection status --name http_notify23 2>/dev/null
{
  "error": {
    "code": "InternalError",
    "message": "The server encountered an internal error. Please retry the request",
    "cause": "Error: No such connection"
  }
}

### Steps to reproduce
1.

alphaprinz:
https://github.com/noobaa/noobaa-core/pull/8801

Issue #8793, opened 2025-02-13
Connection delete is throwing internal server error if connection name provided doesnot exist


ramya-c3:
### Environment info
- NooBaa Version: VERSION
- Platform: standalone

### Actual behavior
1. /usr/local/bin/noobaa-cli connection delete --name test 2>/dev/null
{
  "error": {
    "code": "InternalError",
    "message": "The server encountered an internal error. Please retry the request",
    "cause": "Error: No such file or directory"
  }
}

### Expected behavior
1. /usr/local/bin/noobaa-cli connection delete --name test 2>/dev/null
{
  "error": {
    "code": "Does not exist", test
  }
}

### Steps to reproduce
1.

alphaprinz:
https://github.com/noobaa/noobaa-core/pull/8801

Issue #8792, opened 2025-02-13
Connection add should throw proper  error instead of InternalError


ramya-c3:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**](https://github.com/noobaa/noobaa-core/issues/8791)

### Actual behavior
1.# /usr/local/bin/noobaa-cli connection add --name conee3 --notification_protocol http --agent_request_object '{"host":"localhost","port":8080,"timeout":100}' --request_options_object '{"auth":"vYoaR9c9tU1pjmv+2wn2I9zO","path":"/webhook"}' 2>/dev/null
{
  "error": {
    "code": "InternalError",
    "message": "The server encountered an internal error. Please retry the request",
    "cause": "Error: configuration file already exists\n    at Object.create_config_file (/usr/local/noobaa-core/src/util/native_fs_utils.js:316:25)\n    at async ConfigFS.create_connection_config_file (/usr/local/noobaa-core/src/sdk/config_fs.js:1364:9)\n    at async Object.add_connect_file (/usr/local/noobaa-core/src/util/notifications_util.js:599:5)\n    at async connection_management (/usr/local/noobaa-core/src/cmd/manage_nsfs.js:752:20)\n    at async main (/usr/local/noobaa-core/src/cmd/manage_nsfs.js:78:13)"
  }
}

### Expected behavior
1. Error should be proper

### Steps to reproduce
1.

alphaprinz:
https://github.com/noobaa/noobaa-core/pull/8801

Issue #8791, opened 2025-02-13
Connection add command output is not proper


ramya-c3:
### Environment info
- NooBaa Version: noobaa-core-5.18.0-20250201.el9.x86_64
- Platform: standalone
- 
### Actual behavior
1.# /usr/local/bin/noobaa-cli connection add --name conee3 --notification_protocol http --agent_request_object '{"host":"localhost","port":8080,"timeout":100}' --request_options_object '{"auth":"vYoaR9c9tU1pjmv+2wn2I9zO","path":"/webhook"}' 2>/dev/null
{
  "response": {
    "code": "ConnectionCreated",
    "reply": {
      "_": [
        "connection",
        "add"
      ],
      "name": "conee3",
      "notification_protocol": "http",
      "agent_request_object": {
        "host": "localhost",
        "port": 8080,
        "timeout": 100
      },
      "request_options_object": {
        "auth": "ubwaUr9HugFurz/8hRfjaYLcX/wfZKNO",
        "path": "/webhook"
      },
      "master_key_id": "67a1fb8e0dc50d8a6d182689"
    }
  }
}

### Expected behavior
1.{
  "response": {
    "code": "ConnectionCreated",
    "reply": {
      "name": "conee3",
      "notification_protocol": "http",
      "agent_request_object": {
        "host": "localhost",
        "port": 8080,
        "timeout": 100
      },
      "request_options_object": {
        "auth": "ubwaUr9HugFurz/8hRfjaYLcX/wfZKNO",
        "path": "/webhook"
      },
      "master_key_id": "67a1fb8e0dc50d8a6d182689"
    }
  }
}

### Steps to reproduce
1.

alphaprinz:
https://github.com/noobaa/noobaa-core/pull/8801

PR #8796, opened 2025-02-13
bucket notification - harden bad name scenario. Also, don't reply internal '_' field.

alphaprinz:
### Explain the changes
1. Some cases of bad user input resulted in internal error.
Reply with an error code instead.
2. Reply for add connection included the internal '_' field.

1. https://github.com/noobaa/noobaa-core/issues/8794
2. https://github.com/noobaa/noobaa-core/issues/8793
3. https://github.com/noobaa/noobaa-core/issues/8792
4. https://github.com/noobaa/noobaa-core/issues/8791

### Testing Instructions:
1. See issues.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -781,40 +781,49 @@ async function notification_management() {
 async function connection_management(action, user_input) {
     manage_nsfs_validations.validate_connection_args(user_input, action);
 
+    //don't reply the internal '_' field.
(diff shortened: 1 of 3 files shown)


aayushchouhan09 (commented):
LGTM, only you can add tests for this.

aayushchouhan09 (approved):

PR #8806, opened 2025-02-18
NC | CLI | List bucket dose not return full bucket name with `.json` suffix

### LINKED ISSUE #8802: bucket name create with extn (.json) , noobaa cli doesn't list the bucket name completely
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

Noobaa d/s 4.18.0 or 4.17.0 

### Actual behavior
1.

Create S3 account
Create S3 bucket (s3 mb s3://newbucket.json)
list the buckets 

[root@523-31 ~]# s3-u2 mb s3://acc-2-bkt-55.json

urllib3/connectionpool.py:1064: InsecureRequestWarning: Unverified HTTPS request is being made to host '192.168.3.102'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
make_bucket: acc-2-bkt-55.json  -->  -->here bucket is created 
[root@523-31 ~]#
[root@523-31 ~]#
[root@523-31 ~]# /usr/local/bin/noobaa-cli bucket list 2>/dev/null
{
  "response": {
    "code": "BucketList",
    "reply": [
      {
        "name": "acc-2-bkt-55"
      },
      {
        "name": "acc-2-bkt-2"
      },
      {
        "name": "acc-2-bkt-56.txt"
      },
      {
        "name": "acc-1-bkt-1"
      }
    ]
  }
}

### Expected behavior
1.
It should print the bucket name appropriately 

### Steps to reproduce
1.

As mentioned above 

Note: even the account name can have .json as an extension on the name , it should work or not ?

naveenpaul1:
### Explain the changes
1. remove suffix remove code from `add_config_file_suffix()`

1. https://github.com/noobaa/noobaa-core/issues/8802

### Testing Instructions:
1. check issue https://github.com/noobaa/noobaa-core/issues/8802

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/docs/NooBaaNonContainerized/NooBaaCLI.md
+++ b/docs/NooBaaNonContainerized/NooBaaCLI.md
@@ -839,3 +839,56 @@ sudo noobaa-cli connection add --from_file <options_connection_JSON_file_path>
 ```
 sudo noobaa-cli whitelist --ips ["127.0.0.1", "192.0.10.0", "3002:0bd6:0000:0000:0000:ee00:0033:6778"]'
 ```
+
+### Known issues
+
+1. Bucket with suffix `.json`
+    
+    When the customer creates a bucket with the suffix `.json`, suffix is removed from the bucket config file name. 
+    
+    For example 
+    ```
+    sudo noobaa-cli bucket add --name my-bucket1.json --owner <owner_name> --path <path>
+    ```
+    Actual file name : `{config_root}\buckets\my-test-bucket.json`
+
+    Expected file name : `{config_root}\buckets\my-test-bucket.json.json`
+    
+    I/O is not possible in this bucket. This issue is only reproducible in older versions(5.18.0 and older).
+    Issue fixed version is <TBD>
+    
+    solution :
+    * Recreate the bucket with same name in new version
+    ```
+    sudo noobaa-cli bucket add --name my-bucket1.json --owner <owner_name> --path <path> 
+    ```
+    * Delete invalid bucket
+    ```
+    bucket delete --name my-bucket1 (bucket name without suffix `.json`)
+    ```
+
+2. Account with suffix `.symlink`
+    
+    When the customer creates an account with the suffix `.symlink`, suffix is removed from the account config file name. 
+    
+    For example 
+    ```
+    sudo noobaa-cli account add --name acc_symlink.symlink --new_buckets_path <path> --uid <uid> --gid <gid>
+    ```
+    Actual file name : `{config_root}\accounts_by_name\acc_symlink.symlink`
+    
+    Expected file name : `{config_root}\accounts_by_name\acc_symlink.symlink.symlink`
+    
+    This issue is only reproducible in older versions(5.18.0 and older).
+    Issue fixed version is <TBD>
+    
+    solution :
+    * Recreate the account with same name in new version
+    ```
+    sudo noobaa-cli account add --name acc_symlink.symlink --new_buckets_path <path> --uid <uid> --gid <gid>
+    ```
+    * Delete invalid account
+    ```
+    sudo noobaa-cli account delete --name acc_symlink (account name without suffix `.symlink`)
+    ```
+
--- a/src/sdk/config_fs.js
+++ b/src/sdk/config_fs.js
@@ -103,13 +103,13 @@ class ConfigFS {
 
     /**
      * add_config_file_suffix returns the config_file_name follwed by the given suffix
+     * Bucket name can have suffix `.json`. This will make sure the bucket name and bucket config file have the same name.
      * @param {string} config_file_name
      * @param {string} suffix
      * @returns {string} 
      */
     add_config_file_suffix(config_file_name, suffix) {
         if (!config_file_name) dbg.warn(`Config file name is missing - ${config_file_name}`);
-        if (String(config_file_name).endsWith(suffix)) return config_file_name;
         return config_file_name + suffix;
     }
 
--- a/src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
+++ b/src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
@@ -603,6 +603,46 @@ describe('manage nsfs cli account flow', () => {
             const res = await exec_manage_cli(type, action, account_options);
             expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidSupplementalGroupsList.code);
         });
+
+        it('cli account add - with account name suffix `.json`', async function() {
+            const action = ACTIONS.ADD;
+            const { type, new_buckets_path, uid, gid } = defaults;
+            await fs_utils.create_fresh_path(new_buckets_path);
+            await fs_utils.file_must_exist(new_buckets_path);
+            const name = 'account_suffix.json';
+            const account_options = { config_root, name, new_buckets_path, uid, gid };
+            const res = await exec_manage_cli(type, action, account_options);
+            const parse_res = JSON.parse(res);
+            expect(parse_res.response.code).toEqual(ManageCLIResponse.AccountCreated.code);
+            expect(parse_res.response.reply.name).toEqual(name);
+            const account = await config_fs.get_account_by_name(name);
+            expect(account.name).toBe(name);
+            const res1 = await exec_manage_cli(TYPES.ACCOUNT, ACTIONS.STATUS, { name, config_root });
+            expect(JSON.parse(res1).response.reply.name).toBe(name);
+            const res_list = await exec_manage_cli(TYPES.ACCOUNT, ACTIONS.LIST, { config_root });
+            expect(JSON.parse(res_list).response.reply.map(item => item.name))
+                .toEqual(expect.arrayContaining([name]));
+        });
+
+        it('cli account add - with account name suffix `.symlink`', async function() {
+            const action = ACTIONS.ADD;
+            const { type, new_buckets_path, uid, gid } = defaults;
+            await fs_utils.create_fresh_path(new_buckets_path);
+            await fs_utils.file_must_exist(new_buckets_path);
+            const name = 'account_suffix.symlink';
+            const account_options = { config_root, name, new_buckets_path, uid, gid };
+            const res = await exec_manage_cli(type, action, account_options);
+            const parse_res = JSON.parse(res);
+            expect(parse_res.response.code).toEqual(ManageCLIResponse.AccountCreated.code);
+            expect(parse_res.response.reply.name).toEqual(name);
+            const account = await config_fs.get_account_by_name(name);
+            expect(account.name).toBe(name);
+            const res1 = await exec_manage_cli(TYPES.ACCOUNT, ACTIONS.STATUS, { name, config_root });
+            expect(JSON.parse(res1).response.reply.name).toBe(name);
+            const res_list = await exec_manage_cli(TYPES.ACCOUNT, ACTIONS.LIST, { config_root });
+            expect(JSON.parse(res_list).response.reply.map(item => item.name))
+                .toEqual(expect.arrayContaining([name]));
+        });
     });
 
     describe('cli update account', () => {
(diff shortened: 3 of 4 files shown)


src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -603,6 +603,26 @@ describe('manage nsfs cli account flow', () => {
...
             expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.InvalidSupplementalGroupsList.code);
         });
+
+        it('cli account add - with account name suffix `.json`', async function() {
+            const action = ACTIONS.ADD;
+            const { type, new_buckets_path, uid, gid } = defaults;
+            await fs_utils.create_fresh_path(new_buckets_path);
+            await fs_utils.file_must_exist(new_buckets_path);
+            const name = 'account_suffix.json';
+            const allow_bucket_creation = 'true'; // root accounts manager is not allowed to create buckets
```
shirady, line 613: I think it's relevant for your test.

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
```diff
@@ -603,6 +603,26 @@ describe('manage nsfs cli account flow', () => {
...
+            const account_options = { config_root, name, new_buckets_path, uid, gid,
+                allow_bucket_creation };
+            const res = await exec_manage_cli(type, action, account_options);
+            const parse_res = JSON.parse(res);
+            expect(parse_res.response.code).toEqual(ManageCLIResponse.AccountCreated.code);
+            expect(parse_res.response.reply.name).toEqual(name);
+            const account = await config_fs.get_account_by_name(name);
+            expect(account).not.toBeNull();
```
shirady, line 621: Are you sure you want to check if the value is `null`?

src/test/unit_tests/jest_tests/test_nc_bucket_cli.test.js
```diff
@@ -54,6 +54,12 @@ describe('manage nsfs cli bucket flow', () => {
             path: bucket_storage_path,
         };
 
+        const bucket_suffix_json = {
+            name: 'bucket3.json',
+            owner: account_defaults.name,
+            path: bucket_storage_path,
```
shirady, line 60: I think you need to create a different `bucket_storage_path` as this one is:
`const bucket_storage_path = path.join(tmp_fs_path, 'root_path_manage_nsfs', 'bucket1');` and you want it with your bucket name ("bucket3.json").

src/test/unit_tests/jest_tests/test_nc_account_cli.test.js
naveenpaul1: It should be fine, it won't break any code.

src/sdk/config_fs.js
```diff
@@ -103,13 +103,13 @@ class ConfigFS {
...
     /**
      * add_config_file_suffix returns the config_file_name follwed by the given suffix
+     * Bucket name can have suffix `.json`. This will make sure the bucket name and bucket config file have the same name.
      * @param {string} config_file_name
      * @param {string} suffix
      * @returns {string} 
      */
     add_config_file_suffix(config_file_name, suffix) {
         if (!config_file_name) dbg.warn(`Config file name is missing - ${config_file_name}`);
-        if (String(config_file_name).endsWith(suffix)) return config_file_name;
```
romayalon, line 112: will it fix an already existing bucket?

src/sdk/config_fs.js
```diff
@@ -103,13 +103,13 @@ class ConfigFS {
 
     /**
      * add_config_file_suffix returns the config_file_name follwed by the given suffix
+     * Bucket name can have suffix `.json`. This will make sure the bucket name and bucket config file have the same name.
      * @param {string} config_file_name
      * @param {string} suffix
      * @returns {string} 
      */
     add_config_file_suffix(config_file_name, suffix) {
         if (!config_file_name) dbg.warn(`Config file name is missing - ${config_file_name}`);
```
romayalon, line 112: do we have the same/similar issue with symlink? like account name that ends with symlink?

src/sdk/config_fs.js
naveenpaul1: no, It won't fix existing bucket, It can fix only new buckets. To fix the existing bucket we need to rename bucket config files. 
Main problem here is not listing. If the bucket.name and bucket config file name are different then S3 ops is not possible in that bucket.

src/sdk/config_fs.js
naveenpaul1: yes, same issue was there for `.symlink` suffix account names too, but with this change it got fixed.

src/test/unit_tests/jest_tests/test_nc_bucket_cli.test.js
```diff
@@ -242,6 +249,19 @@ describe('manage nsfs cli bucket flow', () => {
...
         });
+
+        it('cli create bucket - with bucket name suffix `.json`', async () => {
+            await fs_utils.create_fresh_path(bucket_json_storage_path);
+            await fs_utils.file_must_exist(bucket_json_storage_path);
+            const action = ACTIONS.ADD;
+            const bucket_options = { config_root, ...bucket_suffix_json };
+            await exec_manage_cli(TYPES.BUCKET, action, bucket_options);
+            const bucket = await config_fs.get_bucket_by_name(bucket_suffix_json.name);
+            expect(bucket).not.toBeNull();
```
shirady, line 260: Same here, if you want to validate that the `bucket` object is defined before accessing its `name` property you might want to change it to:
```suggestion
            expect(bucket).toBeDefined();
```

src/test/unit_tests/jest_tests/test_nc_bucket_cli.test.js
```diff
@@ -242,6 +249,19 @@ describe('manage nsfs cli bucket flow', () => {
...
+            expect(bucket.name).toBe(bucket_options.name);
+            const res = await exec_manage_cli(TYPES.BUCKET, ACTIONS.STATUS, { name: bucket_suffix_json.name, config_root });
+            expect(JSON.parse(res).response.reply.name).toBe(bucket_options.name);
```
shirady, line 263: To be aligned with the issue steps you can also test it using the `ACTIONS.LIST`.

shirady (commented):
minor comments

src/sdk/config_fs.js
romayalon: Can you please add to the PR description how it behaves on already existing buckets? it's not clear to me
also a related issue of existing buckets/config files is that if you provide on aws s3 cli tool the bucket_name or bucket_name.json it works on both, so we need to handle it as part of this PR

romayalon:
After fixing this issue we might want to backport it to old versions 
CC: @nimrod-becker

src/sdk/config_fs.js
naveenpaul1: @romayalon There are two buckets with suffice `.json`, one created before code changes and another after a code change.
```
root:noobaa-core# node src/cmd/manage_nsfs bucket list 2>/dev/null
{
  "response": {
    "code": "BucketList",
    "message": "Bucket list retrieved successfully",
    "reply": [
      {
        "name": "my-bucket1" ---> bucket with old code
      },
      {
        "name": "my-bucket2.json" ---> bucket with new code
      }
    ]
  }
}
```
But for S3 CLI bucket list both bucktes returns with suffix `.json`, this is because name is fetch from bucket.name property
```
naveenpaul:noobaa-core$ s3-nav1 s3 ls
2025-02-28 11:48:47 my-bucket1.json
2025-02-28 11:49:20 my-bucket2.json
```
And the config files in the `bucket` dir look like this
```
root:buckets#  ls -lrt
total 8
-rw-------. 1 root root 253 Feb 28 11:48 my-bucket1.json  ---> missing suffix `.json`
-rw-------. 1 root root 253 Feb 28 11:49 my-bucket2.json.json
```
In case of S3 ops, Object-put work for bucket `my-bucket1.json` only if the we pass bucket name without suffix `.json` otherwise we are getting `NoSuchBucket` error
```
naveenpaul:noobaa-core$ s3-nav1 s3 cp ./config.js s3://my-bucket1
upload: ./config.js to s3://my-bucket1/config.js
```
For bucket `my-bucket2.json`, it works as expected with bucket name `my-bucket2.json`
```
naveenpaul:noobaa-core$ s3-nav1 s3 cp ./config.js s3://my-bucket2.json
upload: ./config.js to s3://my-bucket2.json/config.js
```

romayalon: @naveenpaul1 what's your suggestion?

naveenpaul1: @romayalon One option is to write an upgrade script for the bucket/account that cross-verifies the bucket/account name with a config file name and if there is some mismatch then rename the bucket/account config file with the correct name. But I don't suggest this upgrade option because upgrading without proper consent is always risky and we are trying to fix one issue that is not reported by a customer and there is no chance for this issue to happen after merging and backporting this fix. We can also document this known issue.
cc : @rkomandu

romayalon: @naveenpaul1 I agree that we would like to avoid upgrades of the config directory and since it's a non trivial bucket name we can document it, we will need to make it extra clear with a full documentation of the issue, a workaround for existing buckets (will be to re-create the bucket using the CLI on higher versions, right? can you pls check it) same for account name that ends with symlink, which is also a very esoteric account name..

rkomandu:
@romayalon , w/r/t documentation, are you saying, either we should allow the bucket-name/account-name to have the .json as the suffix. We shouldn't touch via upgrade path, as it would get us into unknown issues. Are we trying to block the creation of bucket-name/account-name or just create and rework the listing ?

romayalon:
@rkomandu 
1. We don't want to block buckets/accounts that ends with .json / .symlink because blocking it will make us incompatible to S3.
2. Naveen will fix it for new buckets/accounts
3. Naveen will document the known issue of buckets/accounts with these suffixes when created on older versions and will document a workaround.

docs/NooBaaNonContainerized/NooBaaCLI.md
```diff
@@ -838,4 +838,55 @@ sudo noobaa-cli connection add --from_file <options_connection_JSON_file_path>
 ### White List Server IP command example
 ```
 sudo noobaa-cli whitelist --ips ["127.0.0.1", "192.0.10.0", "3002:0bd6:0000:0000:0000:ee00:0033:6778"]'
-```
\ No newline at end of file
+```
+
+### Known issues
+
+1. Bucket with suffix `.json`
```
romayalon, line 845: please write to which version this is a known issue and when it was fixed

romayalon (approved):

rkomandu:
@naveenpaul1 @romayalon  , are you going to backport this into any 5.18.z stream ? or leave it as is for now

romayalon:
@rkomandu yes, we aim to backport it to 5.18.1

rkomandu:
@naveenpaul1 
>>Naveen will document the known issue of buckets/accounts with these suffixes when created on older versions and will document a workaround.

could you share the documentation  to add in the scale documentation 

@madhuthorat FYI

naveenpaul1:
@rkomandu please find the doc here. https://github.com/noobaa/noobaa-core/blob/master/docs/NooBaaNonContainerized/NooBaaCLI.md#known-issues
You can also find a open doc update PR : https://github.com/noobaa/noobaa-core/pull/8941

rkomandu:
verified in the noobaa-core-5.18.2-20250410 build , could list with scale mm command as well the listing. Upload an object to the bucket that was created 

/usr/local/bin/noobaa-cli bucket list

{
  "response": {
    "code": "BucketList",
    "message": "Bucket list retrieved successfully",
    "reply": [
      {
        "name": "newbucket-17apr-warp"
      },
      {
        "name": "newbucket-rkomandu.json"
      }
    ]
  }
}

PR #8808, opened 2025-02-18
NSFS | Fix delete last object deletes directory object if its size is 0

### LINKED ISSUE #8805: NSFS | Delete last object in a directory that is also a directory object deletes the directory if the object directory size is 0 (no internal .folder optimization)
### Environment info
- NooBaa Version: 5.19
- Platform: All

### Actual behavior
1. When creating a 0 size directory object (a/) and creating another object in it, and later deleting the object, the directory object will be deleted including the directory.

### Expected behavior
1. When creating a 0 size directory object (a/) and creating another object in it, and later deleting the object, the directory object will not be deleted including the directory as it's on AWS.

### Steps to reproduce
1. Tab 1 - Init NooBaa service - `sudo node noobaa-core/src/cmd/nsfs.js --debug 5`
2. Tab 2 - 
3. Create a directory for the bucket - `mkdir /private/tmp/dir1/`
4. Create an account - `noobaa-cli account add --name account1 --user root --new_buckets_path /private/tmp/`
5. Create a bucket - `noobaa-cli bucket add --name bucket1 --owner account1 --path /private/tmp/dir1/`
6. Create an alias to S3 client - `alias s3api='AWS_ACCESS_KEY_ID=<access-key>  AWS_SECRET_ACCESS_KEY=<secret-key> aws --endpoint https://localhost:6443 --no-verify-ssl s3api'`
7. Create an empty directory object - `s3api put-object --bucket bucket1 --key=a0/`
8. Create a regular file - `echo "blblblbl" > a1.txt`
9. Create a regular object in a0/ directory - `s3api put-object --bucket bucket1 --key=a0/obj.txt --body a1.txt`
10. Delete the regular object from a0/ - `s3api delete-object --bucket bucket1 --key=a0/obj.txt`
11. Check if a0/ still exists using S3 api - 
```
s3api get-object --bucket bucket1 --key a0/ dev/stdout
An error occurred (NoSuchKey) when calling the GetObject operation (reached max retries: 0): The specified key does not exist. 
``` 
13. Check if a0/ still exists using the file system commands- 
``` 
sudo ls -la  /private/tmp/a0/
ls: /private/tmp/a0/: No such file or directory
```

romayalon:
### Explain the changes
1. Added stat inside the _delete_path_dirs in order to check before we delete the directory if any dir in the hierarchy is a directory object and if yes - break the loop.

1. Fixed https://github.com/noobaa/noobaa-core/issues/8805

### Testing Instructions:
1. `sudo node node_modules/mocha/bin/mocha src/test/unit_tests/test_namespace_fs -g "delete inner object in directory object size 0 - no .folder file"`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -2618,12 +2618,32 @@ class NamespaceFS {
         }
     }
 
+    /**
+     * _delete_path_dirs deletes all the paths in the hierarchy that are empty after a successful delete
+     * if the original file_path to be deleted is a regular object which means file_path is not a directory and it's not a directory object path - 
+     * before deletion of the parent directory  - 
+     * if the parent directory is a directory object (has CONTENT_DIR xattr) - stop the deletion loop
+     * else - delete the directory - if dir is not empty it will stop at the first non empty dir
+     * NOTE - the directory object check is needed because when object size is zero we won't create a .folder file and the dir will be empty
+     * therefore the deletion will succeed although we shouldn't delete the directory object
+     * @param {String} file_path 
+     * @param {nb.NativeFSContext} fs_context 
+     */
     async _delete_path_dirs(file_path, fs_context) {
         try {
-            let dir = path.dirname(file_path);
-            while (dir !== this.bucket_path) {
-                await nb_native().fs.rmdir(fs_context, dir);
-                dir = path.dirname(dir);
+            let dir_path = path.dirname(file_path);
+            const deleted_file_is_dir = file_path.endsWith('/');
+            const deleted_file_is_dir_object = file_path.endsWith(config.NSFS_FOLDER_OBJECT_NAME);
+            let should_check_dir_path_is_content_dir = !deleted_file_is_dir && !deleted_file_is_dir_object;
+            while (dir_path !== this.bucket_path) {
+                if (should_check_dir_path_is_content_dir) {
+                    const dir_stat = await nb_native().fs.stat(fs_context, dir_path);
+                    const file_is_disabled_dir_content = dir_stat.xattr && dir_stat.xattr[XATTR_DIR_CONTENT] !== undefined;
+                    if (file_is_disabled_dir_content) break;
+                }
+                await nb_native().fs.rmdir(fs_context, dir_path);
+                dir_path = path.dirname(dir_path);
+                should_check_dir_path_is_content_dir = true;
             }
         } catch (err) {
             if (err.code !== 'ENOTEMPTY' &&
@@ -3068,10 +3088,18 @@ class NamespaceFS {
         return res;
     }
 
-    // delete version_id -
-    // 1. get version info, if it's empty - return
-    // 2. unlink key
-    // 3. if version is latest version - promote second latest -> latest
+    /**
+     * delete version_id does the following - 
+     * 1. get version info, if it's empty - return
+     * 2. unlink key
+     * 3. if version is latest version - promote second latest -> latest
+     * 4. if it's the latest version - delete the directory hirerachy of the key if it's empty
+     *    if it's a past version - delete .versions/ and the directory hirerachy if it's empty
+     * @param {nb.NativeFSContext} fs_context 
+     * @param {String} file_path 
+     * @param {Object} params 
+     * @returns {Promise<{deleted_delete_marker?: string, version_id?: string}>}
+     */
     async _delete_version_id(fs_context, file_path, params) {
         // TODO optimization - GPFS link overrides, no need to unlink before promoting, but if there is nothing to promote we should unlink
         const del_obj_version_info = await this._delete_single_object_versioned(fs_context, params.key, params.version_id);
--- a/src/test/unit_tests/test_namespace_fs.js
+++ b/src/test/unit_tests/test_namespace_fs.js
@@ -1201,6 +1201,36 @@ mocha.describe('namespace_fs folders tests', function() {
 
         });
 
+        mocha.it('delete inner object in directory object size 0 - no .folder file but directory still exists', async function() {
+            const inner_key = '/inner_obj';
+            const key = upload_key_3 + inner_key;
+            const source = buffer_utils.buffer_to_read_stream(data);
+            await upload_object(ns_tmp, upload_bkt, key, dummy_object_sdk, source);
+            const p1 = path.join(ns_tmp_bucket_path, upload_key_3);
+            const p2 = path.join(ns_tmp_bucket_path, key);
+            await fs_utils.file_must_not_exist(path.join(p1, config.NSFS_FOLDER_OBJECT_NAME));
+            const full_xattr1 = await get_xattr(p1);
+            assert.deepEqual(full_xattr1, { ...user_md_and_dir_content_xattr, [XATTR_DIR_CONTENT]: obj_sizes_map[upload_key_3] });
+            await ns_tmp.delete_object({ bucket: upload_bkt, key: key }, dummy_object_sdk);
+            await fs_utils.file_must_exist(p1);
+            await fs_utils.file_must_not_exist(p2);
+        });
+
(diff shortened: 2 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -2578,10 +2578,19 @@ class NamespaceFS {
...
+     * if dir is also a directory object (disabled mode optimization - CONTENT_DIR xattr is on the directory) we stop the loop
+     * @param {String} file_path 
+     * @param {nb.NativeFSContext} fs_context 
+     */
     async _delete_path_dirs(file_path, fs_context) {
         try {
             let dir = path.dirname(file_path);
             while (dir !== this.bucket_path) {
+                const dir_stat = await nb_native().fs.stat(fs_context, dir);
+                if (dir_stat.xattr && dir_stat.xattr[XATTR_DIR_CONTENT]) break;
```
shirady, line 2593: I think that adding the explicit compare to undefined can help reading, as it is might be confusing if were not focus that the dir_stat.xattr[XATTR_DIR_CONTENT] of empty directory object is '0' (string with 0).
```suggestion
                if (dir_stat.xattr && dir_stat.xattr[XATTR_DIR_CONTENT] !== undefined) break;
```

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -1177,6 +1177,22 @@ mocha.describe('namespace_fs folders tests', function() {
 
         });
 
+        mocha.it('delete inner object in directory object size 0 - no .folder file', async function() {
+            const inner_key = '/inner_obj';
+            const key = upload_key_3 + inner_key;
+            const datar = crypto.randomBytes(100);
```
shirady, line 1183: typo
```suggestion
            const data = crypto.randomBytes(100);
```

src/sdk/namespace_fs.js
```diff
@@ -2578,10 +2578,19 @@ class NamespaceFS {
...
+     * if dir is not empty it will stop at the first non empty dir
+     * if dir is also a directory object (disabled mode optimization - CONTENT_DIR xattr is on the directory) we stop the loop
+     * @param {String} file_path 
+     * @param {nb.NativeFSContext} fs_context 
+     */
     async _delete_path_dirs(file_path, fs_context) {
         try {
             let dir = path.dirname(file_path);
             while (dir !== this.bucket_path) {
+                const dir_stat = await nb_native().fs.stat(fs_context, dir);
```
nadavMiz, line 2592: we have a lot of the same checks in different ways. I think we should start standardizing the way we handle content dir and paths in general. I have made a function _is_disabled_content_dir that does this check. maybe you can see if you can use it here. or maybe create a new function that is called both in here and in _is_disabled_content_dir. also I think its better to use stat_ignore_enoent, because if this directory is deleted by a different process we will get an exception

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -1177,6 +1177,22 @@ mocha.describe('namespace_fs folders tests', function() {
...
+            const source = buffer_utils.buffer_to_read_stream(datar);
+            await upload_object(ns_tmp, upload_bkt, key, dummy_object_sdk, source);
+            const p1 = path.join(ns_tmp_bucket_path, upload_key_3);
+            const p2 = path.join(ns_tmp_bucket_path, key);
+            const full_xattr1 = await get_xattr(p1);
+            assert.deepEqual(full_xattr1, { ...user_md_and_dir_content_xattr, [XATTR_DIR_CONTENT]: obj_sizes_map[upload_key_3] });
+            await ns_tmp.delete_object({ bucket: upload_bkt, key: key }, dummy_object_sdk);
+            await fs_utils.file_must_not_exist(path.join(p1, config.NSFS_FOLDER_OBJECT_NAME));
```
shirady, line 1191: I would change the position of this assertion:
- `await fs_utils.file_must_not_exist(path.join(p1, config.NSFS_FOLDER_OBJECT_NAME));` -> `.folder` exists after the `upload_object` (and the `p1` variable declaration).

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -1177,6 +1177,22 @@ mocha.describe('namespace_fs folders tests', function() {
...
+            await fs_utils.file_must_exist(p1);
+            await fs_utils.file_must_not_exist(p2);
```
shirady, line 1193: I think we can either add a comment about this or edit the test title.

src/sdk/namespace_fs.js
```diff
@@ -2578,12 +2578,24 @@ class NamespaceFS {
...
+     */
     async _delete_path_dirs(file_path, fs_context) {
         try {
-            let dir = path.dirname(file_path);
-            while (dir !== this.bucket_path) {
-                await nb_native().fs.rmdir(fs_context, dir);
-                dir = path.dirname(dir);
+            let dir_path = path.dirname(file_path);
+            while (dir_path !== this.bucket_path) {
+                if (file_path !== path.join(dir_path, '/') && file_path !== path.join(dir_path, config.NSFS_FOLDER_OBJECT_NAME)) {
```
nadavMiz, line 2593: what about nested directory objects. if there are two empty content dir keys: one key dir/, and another dir/key1/.
if we delete dir/key1/ won't we delete dir/

src/sdk/namespace_fs.js
nadavMiz: no need for stat_ignore_enoent because error is handled in catch. using _is_disabled_content_dir is not good here. we can do path handle refactoring in different PR if needed

src/sdk/namespace_fs.js
```diff
@@ -2618,12 +2618,31 @@ class NamespaceFS {
...
+                const file_is_dir_content_hidden_file = file_path === path.join(dir_path, config.NSFS_FOLDER_OBJECT_NAME);
+                const file_is_regular_obj = !file_is_dir && !file_is_dir_content_hidden_file;
+                if (file_is_regular_obj) {
+                    const dir_stat = await nb_native().fs.stat(fs_context, dir_path);
+                    const file_is_disabled_dir_content = dir_stat.xattr && dir_stat.xattr[XATTR_DIR_CONTENT] !== undefined;
+                    if (file_is_disabled_dir_content) break;
+                }
+                await nb_native().fs.rmdir(fs_context, dir_path);
+                dir_path = path.dirname(dir_path);
             }
```
nadavMiz, line 2646: I think what confuses me about this is that this condition can be true only for the first iteration. maybe you can assign it before something like:
```suggestion
            let dir_path = path.dirname(file_path);
            const file_is_dir = file_path.endsWith('/') || file_path.endsWith(config.NSFS_FOLDER_OBJECT_NAME);
            const first = true;
            while (dir_path !== this.bucket_path) {
                if (first && !file_is_dir) {
                const dir_stat = await nb_native().fs.stat(fs_context, dir_path);
                    const file_is_disabled_dir_content = dir_stat.xattr && dir_stat.xattr[XATTR_DIR_CONTENT] !== undefined;
                    if (file_is_disabled_dir_content) break;
                }
                await nb_native().fs.rmdir(fs_context, dir_path);
                dir_path = path.dirname(dir_path);
                first = false;
```
hopefully I understand the intention correctly. its still not ideal with the first argument always assigned to be false. but maybe easier to understand? WDYT?

nadavMiz: another thing to consider is that I think dirname of file that ends with "/" is to directory above it. so maybe file_path.endsWith('/') will cause problems. if this is correct, it worked in the first time because  `file_is_dir = file_path === path.join(dir_path, '/');` can never be true. again sorry if I am completely wrong

romayalon: it's the opposite, only on the first iteration we might want to skip the check and delete the folder directly (if the file is a directory object)
on all the other iterations we must check if the folder is a directory object

I'll try a different approach so you can understand it better

nadavMiz (approved):
LGTM

PR #8809, opened 2025-02-19
NSFS | Fix Bug | Race Between List Object and Delete Object

shirady:
### Explain the changes
1. Replace the function `stat_ignore_eacces` with `stat_if_exists` and add ignore of `ENOENT` and `ENOTDIR` (in addition to `EACCES` with a flag of `config.EACCES_IGNORE_ENTRY`).
2. Call `stat` on the `entry_path` one time and save is instead of calling it in concurrency so that we can be sure that the value was not deleted from the result array. Move the `stat` call before `results.push(r)` and `results.splice(pos, 0, r)` as they both add a result to the array.
3. Add a comment before the `check_access`, see also in PR description of #6576.

This change is continuing PR #8751 by adding a case to ignore a `stat` failure entry.

### Issues: Fixed [DFBUGS-1582](https://issues.redhat.com/browse/DFBUGS-1582)
1. Currently, during a list object operation there is a call to stat on the `entry_path` (which is the path of the key in the bucket in the FS). If a concurrent delete happens, this `stat` will return an error (`ENOENT` as the object was deleted). In this fix, we suggest that the `stat` would not return any value and, as a result, would not appear in the list object result.

GAPs - are mentioned in issue #8845 (discussions were documented [here](https://docs.google.com/document/d/1WpfNBLzcZQvT4DThm70xUYGRRaIF4ymw80Rpy8sk-OI/edit?usp=sharing)).

#### Automatic Test:
Please run: `sudo npx jest test_nsfs_concurrency.test.js`

For example we could see the lines:
```
Mar-3 14:10:13.679 [/69906]    [L0] core.util.native_fs_utils:: stat_if_exists: Could not access file entry_path /private/tmp/test_nsfs_concurrency/my-key-2 error code ENOENT , skipping...
```

```
Mar-3 14:10:43.534 [/69906]    [L0] core.util.native_fs_utils:: stat_if_exists: Could not access file entry_path /private/tmp/test_nsfs_concurrency/my-key-1000 error code ENOENT , skipping...
```

#### Manual Testing instructions:
We will add code changes to simulate the problem:
Add the sleep function:
```js
sleep(ms) {
    return new Promise(resolve => {
        setTimeout(resolve, ms);
    });
}
```
we will add the following lines:
```js
console.log('SDSD before stat', this.bucket_path, r.key, fs_context);
console.log('SDSD sleep 8000');
await this.sleep(8000);
stat = await nb_native().fs.stat(fs_context, entry_path, { use_lstat });
```

5. Start the NSFS server with: `sudo node src/cmd/nsfs --debug 5`
6. Create the alias for S3 service:`alias nc-user-1-s3=‘AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443’`
7. Check the connection to the endpoint and try to list the buckets (should be empty): `nc-user-1-s3 s3 ls; echo $?`
8. Add bucket to the account using AWS CLI: `nc-user-1-s3 s3 mb s3://bucket-race` (`bucket-race` is the bucket name in this example)
9. Put objects in the bucket: `echo 'hello_world' | nc-user-1-s3 s3 cp - s3://bucket-race/hello_world1.txt` (repeat this and change the key name to `hello_world2.txt`, `hello_world3.txt`, `hello_world4.txt`
10. Tab1: Start the list objects: `nc-user-1-s3 s3api list-objects-v2 --bucket bucket-race`
    Tab2: Delete an object while you see the "SDSD before stat" on the key: `nc-user-1-s3 s3api delete-object --bucket bucket-race --key hello_world3.txt` (for example)
Expect to see the list without the deleted key, and also to see in the logs printings.

##### Before the change
It was: `An error occurred (NoSuchKey) when calling the ListObjectsV2 operation: The specified key does not exist.`
and in the logs:
```
Feb-19 9:03:11.560 [nsfs/22301] [ERROR] core.endpoint.s3.s3_rest:: S3 ERROR <?xml version="1.0" encoding="UTF-8"?><Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Resource>/bucket-race?list-type=2&amp;encoding-type=url</Resource><RequestId>m7bkghze-8fc8b7-e0y</RequestId></Error> GET /bucket-race?list-type=2&encoding-type=url {"host":"localhost:6443","accept-encoding":"identity","user-agent":"aws-cli/2.17.11 md/awscrt#0.20.11 ua/2.0 os/macos#24.2.0 md/arch#arm64 lang/python#3.11.10 md/pyimpl#CPython cfg/retry-mode#standard md/installer#source md/prompt#off md/command#s3api.list-objects-v2","x-amz-date":"20250219T070256Z","x-amz-content-sha256":"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855","authorization":"AWS4-HMAC-SHA256 Credential=Dwertyuiopasdfg11001/20250219/us-east-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=98e78f14c8a60ea9c1442face8f569e040bab077beeff61a124b07349e8eaa29"} Error: No such file or directory - context: Stat _path=/Users/buckets/bucket-race/hello_world3.txt
```
[Logs before the change](https://github.com/user-attachments/files/18864153/local_reproduce_logs_before.txt)

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -718,11 +718,17 @@ class NamespaceFS {
                     if (!delimiter && r.common_prefix) {
                         await process_dir(r.key);
                     } else {
-                        if (pos < results.length) {
-                            results.splice(pos, 0, r);
-                        } else {
-                            const stat = await native_fs_utils.stat_ignore_eacces(this.bucket_path, r, fs_context);
-                            if (stat) {
+                        const entry_path = path.join(this.bucket_path, r.key);
+                        // If entry is outside of bucket, returns stat of symbolic link
+                        const use_lstat = !(await this._is_path_in_bucket_boundaries(fs_context, entry_path));
+                        const stat = await native_fs_utils.stat_if_exists(fs_context, entry_path,
+                            use_lstat, config.NSFS_LIST_IGNORE_ENTRY_ON_EACCES);
+                        if (stat) {
+                            r.stat = stat;
+                            // add the result only if we have the stat information
+                            if (pos < results.length) {
+                                results.splice(pos, 0, r);
+                            } else {
                                 results.push(r);
                             }
                         }
@@ -764,6 +770,11 @@ class NamespaceFS {
                     await insert_entry_to_results_arr(r);
                 };
 
+                // our current mechanism - list the files and skipping inaccessible directory (invisible in the list).
+                // We use this check_access in case the directory is not accessible inside a bucket.
+                // In a directory if we don’t have access to the directory, we want to skip the directory and its sub directories from the list.
+                // We did it outside to avoid undefined values in the cache.
+                // Note: It is not the same case as a file without permission.
                 if (!(await this.check_access(fs_context, dir_path))) return;
                 try {
                     if (list_versions) {
@@ -885,13 +896,6 @@ class NamespaceFS {
 
             const prefix_dir_key = prefix.slice(0, prefix.lastIndexOf('/') + 1);
             await process_dir(prefix_dir_key);
-            await Promise.all(results.map(async r => {
-                if (r.common_prefix) return;
-                const entry_path = path.join(this.bucket_path, r.key);
-                //If entry is outside of bucket, returns stat of symbolic link
-                const use_lstat = !(await this._is_path_in_bucket_boundaries(fs_context, entry_path));
-                r.stat = await nb_native().fs.stat(fs_context, entry_path, { use_lstat });
-            }));
             const res = {
                 objects: [],
                 common_prefixes: [],
--- a/src/test/unit_tests/jest_tests/test_nsfs_concurrency.test.js
+++ b/src/test/unit_tests/jest_tests/test_nsfs_concurrency.test.js
@@ -6,7 +6,7 @@ const P = require('../../../util/promise');
 const fs_utils = require('../../../util/fs_utils');
 const NamespaceFS = require('../../../sdk/namespace_fs');
 const buffer_utils = require('../../../util/buffer_utils');
-const { TMP_PATH } = require('../../system_tests/test_utils');
+const { TMP_PATH, TEST_TIMEOUT } = require('../../system_tests/test_utils');
 const { crypto_random_string } = require('../../../util/string_utils');
 const endpoint_stats_collector = require('../../../sdk/endpoint_stats_collector');
 
@@ -27,18 +27,20 @@ function make_dummy_object_sdk(nsfs_config, uid, gid) {
 }
 
 const DUMMY_OBJECT_SDK = make_dummy_object_sdk(true);
-describe('test nsfs concurrency', () => {
-    const tmp_fs_path = path.join(TMP_PATH, 'test_nsfs_concurrency');
 
-    const nsfs = new NamespaceFS({
-        bucket_path: tmp_fs_path,
-        bucket_id: '1',
-        namespace_resource_id: undefined,
-        access_mode: undefined,
-        versioning: 'DISABLED',
-        force_md5_etag: false,
-        stats: endpoint_stats_collector.instance(),
-    });
+const tmp_fs_path = path.join(TMP_PATH, 'test_nsfs_concurrency');
+
+const nsfs = new NamespaceFS({
+    bucket_path: tmp_fs_path,
+    bucket_id: '1',
+    namespace_resource_id: undefined,
+    access_mode: undefined,
+    versioning: 'DISABLED',
+    force_md5_etag: false,
+    stats: endpoint_stats_collector.instance(),
+});
+
+describe('test nsfs concurrency', () => {
 
     beforeEach(async () => {
         await fs_utils.create_fresh_path(tmp_fs_path);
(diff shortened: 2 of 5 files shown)


src/sdk/namespace_fs.js
```diff
@@ -721,7 +721,7 @@ class NamespaceFS {
                         if (pos < results.length) {
                             results.splice(pos, 0, r);
                         } else {
-                            const stat = await native_fs_utils.stat_ignore_eacces(this.bucket_path, r, fs_context);
+                            const stat = await native_fs_utils.stat_ignore_eacces_enoent(this.bucket_path, r.key, fs_context);
```
naveenpaul1, line 724: @shirady At the end of `list_object()` method code again checks the `stat` of each entry of `results` array. If a entry is added to `results` and deleted before checking the `results` array `stat`  it might cause an issue.

naveenpaul1: @shirady Permission change is not as common as deletion, deletion we can do through S3 API, but permission change something has to be done at dir level. 
I am not sure do we really need to check the file path and access again. We added it to the `results` because it was available at the movement.

shirady: I can use `stat_ignore_enoent` instead, but still, I would have fewer results.

nadavMiz: @shirady notice that if entry will not have a stat it might fail in `_get_object_info`. also better check that there are no issues with the limit. see https://github.com/noobaa/noobaa-core/pull/8751#discussion_r1940674457

nadavMiz, line 724: @shirady notice that if entry will not have a stat it might fail in _get_object_info. also better check that there are no issues with the limit. see https://github.com/noobaa/noobaa-core/pull/8751#discussion_r1940674457

nadavMiz, line 724: what will happen if file will get deleted after this check?

shirady: I think it is related to the discussion [here](https://github.com/noobaa/noobaa-core/pull/8809#discussion_r1961276737), and I made some changes after thinking about it.

shirady: I added a commit with a change.
You can share what you think.

src/sdk/namespace_fs.js
```diff
@@ -890,7 +892,8 @@ class NamespaceFS {
                 const entry_path = path.join(this.bucket_path, r.key);
                 //If entry is outside of bucket, returns stat of symbolic link
                 const use_lstat = !(await this._is_path_in_bucket_boundaries(fs_context, entry_path));
-                r.stat = await nb_native().fs.stat(fs_context, entry_path, { use_lstat });
+                const result_stat = await native_fs_utils.stat_ignore_enoent(fs_context, entry_path, { use_lstat });
+                r.stat = result_stat || r.previous_stat; // must have stat for _get_object_info - we will use the previous stat in case of failure
```
nadavMiz, line 896: if its already a part of r, why do we need to stat it again? if its not a part of r, we might not have stat at all, which will cause an issue with _get_object_info

src/sdk/namespace_fs.js
nadavMiz: the idea is that we do the stat check after the entry was created in a function `insert_entry_to_results_arr`. so essentially the function doesn't do its intended purpose.we already have a list of checks to do in order to decide if we want to create the entry at the beginning of process_entry. we can add this check there. and if the stat pass we can add it to the r structure there. in my opinion this will keep us consistent with how the process is working, making it more readable

nadavMiz: this suggestion is purely cosmetic, we should probably do it in a different PR

src/test/unit_tests/jest_tests/test_nsfs_concurrency.test.js
```diff
@@ -68,5 +71,65 @@ describe('test nsfs concurrency', () => {
...
+        const key_to_delete = `my-key-${random_num}`;
+
+        nsfs.list_objects({ bucket: bucket }, DUMMY_OBJECT_SDK)
+            .catch(err => {
+                console.log('error during list_objects', err);
+                throw err;
+            }).then(res => {
+                console.log('list was successful');
+            });
+        nsfs.delete_object({ bucket: bucket, key: key_to_delete }, DUMMY_OBJECT_SDK)
```
romayalon, line 90: how do you know here that the list objects got an ENOENT in the middle and that the delete didn't happen before reading the directory?

src/sdk/namespace_fs.js
```diff
@@ -890,7 +892,8 @@ class NamespaceFS {
                 const entry_path = path.join(this.bucket_path, r.key);
                 //If entry is outside of bucket, returns stat of symbolic link
                 const use_lstat = !(await this._is_path_in_bucket_boundaries(fs_context, entry_path));
-                r.stat = await nb_native().fs.stat(fs_context, entry_path, { use_lstat });
+                const result_stat = await native_fs_utils.stat_ignore_enoent(fs_context, entry_path, { use_lstat });
```
romayalon, line 895: why ignore just enoent? if we got a different error we will fail the entire list op again

src/sdk/namespace_fs.js
```diff
@@ -721,9 +722,10 @@ class NamespaceFS {
                         if (pos < results.length) {
                             results.splice(pos, 0, r);
                         } else {
-                            const stat = await native_fs_utils.stat_ignore_eacces(this.bucket_path, r, fs_context);
+                            const stat = await native_fs_utils.stat_ignore_eacces_enoent(this.bucket_path, r.key, fs_context);
```
romayalon, line 725: I don't feel we want to ignore only these 2 errors... we need to decide on a robust behavior... we will just keep adding error codes all the time

src/test/unit_tests/jest_tests/test_nsfs_concurrency.test.js
shirady: In the automatic test I can't know for sure where the deletion happens.
In the manual test, I added the `sleep` and can control exactly when I execute the deletion.

src/sdk/namespace_fs.js
shirady: Do you want to on the second `stat` to ignore all errors? I can do that

src/sdk/namespace_fs.js
shirady: Do you suggest that also for the first `stat` we will ignore all the errors?

src/test/unit_tests/jest_tests/test_nsfs_concurrency.test.js
romayalon: so it's not a good test... try changing the number of initial objects to 1000 or more so it'll be a longer and delete the last one and check if you always see that enoent error

shirady: I think it is a good test and to increase the probability of it I run it a couple of times in a loop.
On a local run with printing of stats I could see it happens.
I also added your suggestion.
Examples from the logs:

- `Feb-20 10:57:42.141 [/39986]    [L0] core.util.native_fs_utils:: stat_ignore_eacces_enoent: Could not access file entry_path /private/tmp/test_nsfs_concurrency/my-key-1 error code ENOENT , skipping...
Feb-20 10:57:42.141 [/39986]   [LOG] CONSOLE:: SDSD stat undefined`

- `Feb-20 10:58:11.560 [/39986]    [L0] core.util.native_fs_utils:: stat_ignore_eacces_enoent: Could not access file entry_path /private/tmp/test_nsfs_concurrency/my-key-1000 error code ENOENT , skipping...
Feb-20 10:58:11.560 [/39986]   [LOG] CONSOLE:: SDSD stat undefine`

To keep it modular I created it as a function so creating more cases should be easier.

src/sdk/namespace_fs.js
shirady: We discussed the error code list, and we divided them into groups (changes in the bucket, permissions/structure, and unexpected errors). In this PR, we will add only the first group.

romayalon (approved):

guymguym (approved):

src/util/native_fs_utils.js
```diff
@@ -303,6 +303,30 @@ async function stat_ignore_enoent(fs_context, file_path) {
     }
 }
 
+/**
+ * stat_if_exists execute stat on entry_path and ignores on certain error codes.
+ * @param {nb.NativeFSContext} fs_context
+ * @param {string} entry_path
+ * @param {boolean} use_lstat
+ * @param {boolean} should_ignore_eacces
+ * @returns {Promise<nb.NativeFSStats>}
```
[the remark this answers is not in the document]
shirady: Thanks, I will change.

src/sdk/namespace_fs.js
shirady: I tried to move the filtering `stat` from `insert_entry_to_results_arr` to `process_entry`, the tests are passing, but we had a discussion, and saw there are a couple of disadvantages: on nested keys (for example dir1/dir2/dir3/my_key.txt) we would stat every dir level, on directory we would stat the entries and would not reuse the stat that was saved in the cache and more (if I missed something important please add).

We also figured that even if we thought that `insert_entry_to_results_arr` only inserted entries and `process_entry` filters them - it would not be true, and we now have filtering steps in `insert_entry_to_results_arr`. 

Anyway, it seems at this point that it is not a small refactor of moving the code block that can be done on a bug fix, but we would need more thorough changes in `_list_objects`.
cc: @nadavMiz @romayalon @naveenpaul1

shirady:
**Backport comment** - We would like the backport to be with `config.NSFS_LIST_IGNORE_ENTRY_ON_EACCES = false;` (and not `true`).

PR #8811, opened 2025-02-19
NSFS | close stream when getting empty content dir

nadavMiz:
### Explain the changes
1. currently we don't close the write stream before returning from read_object_md for empty content dir case. this causes the client to wait for the stream to close until it times out. added closing of the stream in that case

1. 

### Testing Instructions:
manual:
1. put empty content dir object: `s3api put-object --bucket test-bucket --key key5/`
2. get the new object: 's3api get-object --bucket test-bucket --key key5 outfile`
3. get should succeed with empty file

automatic:
`sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_namespace_fs.js`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1022,7 +1022,12 @@ class NamespaceFS {
                     // NOTE: don't move this code after the open
                     // this can lead to ENOENT failures due to file not exists when content size is 0
                     // if entry is a directory object and its content size = 0 - return empty response
-                    if (await this._is_empty_directory_content(file_path, fs_context, params)) return null;
+                    if (await this._is_empty_directory_content(file_path, fs_context, params)) {
+                        res.end();
+                        // since we don't write anything to the stream wait_finished might not be needed. added just in case there is a delay
+                        await stream_utils.wait_finished(res, { signal: object_sdk.abort_controller.signal });
+                        return null;
+                    }
 
                     file = await nb_native().fs.open(
                         fs_context,
--- a/src/test/unit_tests/test_namespace_fs.js
+++ b/src/test/unit_tests/test_namespace_fs.js
@@ -547,6 +547,7 @@ mocha.describe('namespace_fs', function() {
         const dir_2 = '/a/b/';
         const upload_key_1 = dir_1 + 'upload_key_1/';
         const upload_key_2 = dir_2 + 'upload_key_2/';
+        const upload_key_empty = 'empty_key/';
         const data = crypto.randomBytes(100);
 
         mocha.before(async function() {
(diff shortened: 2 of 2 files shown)


src/test/unit_tests/test_namespace_fs.js
```diff
@@ -558,6 +559,22 @@ mocha.describe('namespace_fs', function() {
             console.log('upload_object with trailing / response', inspect(upload_res));
         });
 
+        mocha.it('get empty content dir', async function() {
+            await ns_tmp.upload_object({
```
shirady, line 563: Just make sure that you want to use the `ns_tmp` (I think we have another option).

src/test/unit_tests/test_namespace_fs.js
```diff
@@ -574,11 +591,17 @@ mocha.describe('namespace_fs', function() {
...
                 bucket: upload_bkt,
                 key: upload_key_2,
             }, dummy_object_sdk);
             console.log('delete_object with trailing / (key 2) response', inspect(delete_res));
+
+            delete_res = await ns_tmp.delete_object({
+                bucket: upload_bkt,
+                key: upload_key_empty,
+            }, dummy_object_sdk);
+            console.log('delete_object with trailing / (empty dir) response', inspect(delete_res));
```
shirady, line 604: Is this printing updated?  the directory is empty but has body size of 0.

shirady (approved):
LGTM
minor comments

src/sdk/namespace_fs.js
```diff
@@ -1021,8 +1021,12 @@ class NamespaceFS {
 
             // NOTE: don't move this code after the open
             // this can lead to ENOENT failures due to file not exists when content size is 0
-            // if entry is a directory object and its content size = 0 - return empty response
-                    if (await this._is_empty_directory_content(file_path, fs_context, params)) return null;
+            // if entry is a directory object and its content size = 0 - close the stream and return empty response
+            if (await this._is_empty_directory_content(file_path, fs_context, params)) {
+                res.end();
+                await stream_utils.wait_finished(res, { signal: object_sdk.abort_controller.signal });
```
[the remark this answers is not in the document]
shirady: adding a comment sounds good.

romayalon (approved):

src/test/unit_tests/test_namespace_fs.js
nadavMiz: I guess it prints the delete_object response. it should contain the version-id, and delete-marker. it isn't related to object size. not sure why we need it. but we do it for the previous object. so might as well do it here too

shirady: I meant only for the text - I thought you would like to mention it is empty content dir (no just empty dir).

src/test/unit_tests/test_namespace_fs.js
nadavMiz: looks like the only difference is that ns_tmp uses bucket path that is in TMP_PATH while ns_src uses bucket path in the current directory for some reason (I guess the intention is the src directory?). besides that they are identical. ns_tmp is used in the previous tests. so might as well use it here.also using the tmp path seems more appropriate in general

PR #8812, opened 2025-02-19
S3 | GET/HEAD Object | Add Response Headers Support 

romayalon:
### Explain the changes
1. Added support in S3 GET/HEAD Object of the following response headers:
- response-cache-control
- response-content-disposition
- response-content-language
- response-content-type
- response-expires

1. Gap - Support response-content-encoding 

### Testing Instructions:
Auto - 
1. 
```
sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nsfs_integration.js -g "Presigned URL tests"
```
2. 
```
sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nsfs_integration.js -g "response headers"
```

Manual - 
CLI -
```
s3api get-object --bucket bucket1 --key=key1 --response-content-disposition "attachment;filename=test.txt" --response-content-type="app/plaintext" outfile
{
    "AcceptRanges": "bytes",
    "LastModified": "2025-02-19T13:12:00+00:00",
    "ContentLength": 21,
    "ETag": "\"mtime-d7wg56a1ucjk-ino-1pvkd4\"",
    "ContentDisposition": "attachment;filename=test.txt",
    "ContentType": "app/plaintext",
    "Metadata": {}
}

s3api head-object --bucket bucket1 --key=key1 --response-content-disposition "attachment;filename=test.txt" --response-content-type="app/plaintext"
{
    "AcceptRanges": "bytes",
    "LastModified": "2025-02-19T13:12:00+00:00",
    "ContentLength": 21,
    "ETag": "\"mtime-d7wg56a1ucjk-ino-1pvkd4\"",
    "ContentDisposition": "attachment;filename=test.txt",
    "ContentType": "app/plaintext",
    "Metadata": {}
}
```

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -48,7 +48,7 @@ async function get_object(req, res) {
             throw new S3Error(S3Error.InvalidObjectState);
         }
     }
-
+    http_utils.set_response_headers_from_request(req, res);
     const obj_size = object_md.size;
     const params = {
         object_md,
--- a/src/util/cloud_utils.js
+++ b/src/util/cloud_utils.js
@@ -58,7 +58,8 @@ async function generate_aws_sts_creds(params, roleSessionName) {
     );
 }
 
-function get_signed_url(params, expiry = 604800) {
+function get_signed_url(params, expiry = 604800, custom_operation = 'getObject') {
+    const op = custom_operation;
     const s3 = new AWS.S3({
         endpoint: params.endpoint,
         credentials: {
@@ -76,12 +77,14 @@ function get_signed_url(params, expiry = 604800) {
             agent: http_utils.get_unsecured_agent(params.endpoint)
         }
     });
+    const response_queries = params.response_queries || {};
     return s3.getSignedUrl(
-        'getObject', {
+        op, {
             Bucket: params.bucket.unwrap(),
             Key: params.key,
             VersionId: params.version_id,
-            Expires: expiry
+            Expires: expiry,
+            ...response_queries
         }
     );
 }
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -3,6 +3,7 @@
 /* eslint-disable no-control-regex */
 
 const _ = require('lodash');
+const util = require('util');
 const ip_module = require('ip');
 const net = require('net');
 const url = require('url');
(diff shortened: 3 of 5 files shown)


src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -112,7 +112,7 @@ async function get_object(req, res) {
             return;
         }
     }
-
+    http_utils.set_response_override(req, res);
```
guymguym, line 115: this is called too late, as there are also returns before.
move to just below the s3_utils.set_response_object_md() - I would add in line 51.

src/util/cloud_utils.js
```diff
@@ -76,12 +77,14 @@ function get_signed_url(params, expiry = 604800) {
             agent: http_utils.get_unsecured_agent(params.endpoint)
         }
     });
+    const response_headers = params.response_headers || {};
```
guymguym, line 80: this is actually more of a "response_queries" rather than headers...

src/util/http_utils.js
```diff
@@ -908,6 +908,21 @@ function handle_server_error(err) {
     process.exit(1);
 }
 
+/**
+ * set_response_override sets the response headers based on the request headers
+ * gap - response-content-encoding needs to be added with a more complex logic
+ * @param {*} req 
+ * @param {import('http').ServerResponse} res 
+ */
+function set_response_override(req, res) {
```
guymguym, line 917:
```suggestion
function set_response_headers_from_request(req, res) {
```

src/util/http_utils.js
```diff
@@ -908,6 +908,21 @@ function handle_server_error(err) {
     process.exit(1);
 }
 
+/**
+ * set_response_override sets the response headers based on the request headers
+ * gap - response-content-encoding needs to be added with a more complex logic
+ * @param {*} req 
+ * @param {import('http').ServerResponse} res
```
guymguym, line 915:
```suggestion
 * @param {http.IncomingMessage} req
 * @param {http.ServerResponse} res
```

src/util/http_utils.js
```diff
@@ -908,6 +908,21 @@ function handle_server_error(err) {
...
 
+/**
+ * set_response_override sets the response headers based on the request headers
+ * gap - response-content-encoding needs to be added with a more complex logic
+ * @param {*} req 
+ * @param {import('http').ServerResponse} res 
+ */
+function set_response_override(req, res) {
+    dbg.log2('set_response_override req.query', req.query);
+    if (req.query['response-cache-control']) res.setHeader('Cache-Control', req.query['response-cache-control']);
```
guymguym, line 919: one pitfall to look for is that both query strings can be repeated by the sender, which will make an array in our req.query. Similarly, headers can also be repeated by passing an array to res.setHeader. However I am not sure if this is desired to output the response headers more than once in this case - can you check the http spec if repeating these query headers has any meaning?

guymguym: this cli might check specifically, but any client might be a bit different and querystring is working like this:
```
> querystring.parse('lala=lulu&dispo=1&dispo=2')
[Object: null prototype] { lala: 'lulu', dispo: [ '1', '2' ] }
```

romayalon: okay, I tried multiple clients for reproducing the scenario, so far without success but keeping to try
while googling it I found the following [stackoverflow question](https://stackoverflow.com/questions/48012994/is-it-fine-to-use-duplicate-response-header-with-same-value) and a reference to the [2616 RFC ](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2)


> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma. The order in which header fields with the same field-name are received is therefore significant to the interpretation of the combined field value, and thus a proxy MUST NOT change the order of these field values when a message is forwarded

Does this answer your question?

guymguym (approved):

PR #8773, opened 2025-02-06
NSFS | versioning | content directory versioning - GET/HEAD/DELETE operations

nadavMiz:
### Explain the changes
1. change _get_file_md_path to return the correct path for content dir for both versioning disabled and enabled layouts
2. change read_object_md to work with content directory with versioning
3. following above changes read_object_stream (get operation) works with directory content with versioning
4. modify delete version and delete latest functions to work with content directory with versioning
5. fix find_max_version_past to work with nested object and content directory objects
6. fix identinations in read_object_md and read_object_stream
7. add tests 

### Issues: Fixed #xxx / Gap #xxx
continuation of #8730 

####GAPS
implement ListObject and ListObjectVersions to work with new content dir structure 
GetObject operation doesn't work with empty directory content (unrelated to this PR)

### Testing Instructions:
automatic tests:
`sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_bucketspace_versioning.js`

- [ ] Doc added/updated
- [x] Tests added

Issue #8531, opened 2024-11-14
Directory object (key/) is not supported combined with versioning Enabled/Suspended


romayalon:
### Environment info
- NooBaa Version: 5.18
- Platform: All

### Actual behavior
1. A .folder file that gets created by uploading a directory object is currently not being versioned when versioning is enabled, on a second PUT of the same directory object the .folder will get overridden, and on delete we won't create a delete marker for this file.
same for MPU, delete latest/delete version id/lists/get/head etc, we need to support all flows with directory object combined with versioning.
### Expected behavior
1. Content directory object can be versioned.

### Steps to reproduce
Regular put example - 
1. aws s3 api put-object --bucket bucket1 --key a/b/c/ --body some_file_v1.txt
2. aws s3 api put-object --bucket bucket1 --key a/b/c/ --body some_file_v2.txt
3. aws s3 api list-object-versions - result - got only the second version

romayalon:
@nadavMiz What's left on this bug? can we close it?

nadavMiz:
closing as all issues were implemented in above PRs

PR #8814, opened 2025-02-19
NSFS | content dir versioning LIST and tagging operations

nadavMiz:
### Explain the changes
1. change list-object-versions to work with content directory objects
2. change tagging functions to use _find_version_path in order to get the correct path
3. remove configuration variable to enable content dir versioning, as it should be completed
4. add documentation about content dir versioning
5. add unit test

Content-dir versioning issue: #8531 
continuation of #8773 

####GAPS

### Testing Instructions:
1. 

- [x] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -763,12 +763,12 @@ class NamespaceFS {
                 /**
                  * @param {fs.Dirent} ent
                  */
-                const process_entry = async ent => {
+                const process_entry = async (ent, is_disabled_dir_content) => {
                     // dbg.log0('process_entry', dir_key, ent.name);
-                    if ((!ent.name.startsWith(prefix_ent) ||
+                    if (!ent.name.startsWith(prefix_ent) ||
                         ent.name < marker_curr ||
                         ent.name === this.get_bucket_tmpdir_name() ||
-                        ent.name === config.NSFS_FOLDER_OBJECT_NAME) ||
+                        (ent.name === config.NSFS_FOLDER_OBJECT_NAME && is_disabled_dir_content) ||
                         this._is_hidden_version_path(ent.name)) {
                         return;
                     }
@@ -783,7 +783,7 @@ class NamespaceFS {
                         };
                     } else {
                         r = {
-                            key: this._get_entry_key(dir_key, ent, isDir),
+                            key: this._get_entry_key(dir_key, ent, isDir, is_disabled_dir_content),
                             common_prefix: isDir,
                             is_latest: true
                         };
@@ -813,8 +813,8 @@ class NamespaceFS {
 
                 // insert dir object to objects list if its key is lexicographicly bigger than the key marker &&
                 // no delimiter OR prefix is the current directory entry
-                const is_dir_content = cached_dir.stat.xattr && cached_dir.stat.xattr[XATTR_DIR_CONTENT];
-                if (is_dir_content && dir_key > key_marker && (!delimiter || dir_key === prefix)) {
+                const is_disabled_dir_content = cached_dir.stat.xattr && cached_dir.stat.xattr[XATTR_DIR_CONTENT];
+                if (is_disabled_dir_content && dir_key > key_marker && (!delimiter || dir_key === prefix)) {
                     const r = { key: dir_key, common_prefix: false };
                     await insert_entry_to_results_arr(r);
                 }
@@ -867,7 +867,7 @@ class NamespaceFS {
                         if (ent.name === config.NSFS_FOLDER_OBJECT_NAME && dir_key === marker_dir) {
                             continue;
                         }
-                        await process_entry(ent);
+                        await process_entry(ent, is_disabled_dir_content);
                         // since we traverse entries in sorted order,
                         // we can break as soon as enough keys are collected.
                         if (is_truncated) break;
@@ -915,6 +915,16 @@ class NamespaceFS {
                 }
             };
 
+            const format_key_name = obj_info => {
+                if (this._is_hidden_version_path(obj_info.key)) {
+                    obj_info.key = path.normalize(obj_info.key.replace(HIDDEN_VERSIONS_PATH + '/', ''));
+                    obj_info.key = _get_filename(obj_info.key);
+                }
+                if (obj_info.key.endsWith(config.NSFS_FOLDER_OBJECT_NAME)) {
+                    obj_info.key = obj_info.key.slice(0, -config.NSFS_FOLDER_OBJECT_NAME.length);
+                }
+            };
+
             const prefix_dir_key = prefix.slice(0, prefix.lastIndexOf('/') + 1);
             await process_dir(prefix_dir_key);
             const res = {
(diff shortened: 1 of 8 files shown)


src/sdk/namespace_fs.js
```diff
@@ -756,7 +756,7 @@ class NamespaceFS {
                         };
                     } else {
                         r = {
-                            key: this._get_entry_key(dir_key, ent, isDir),
+                            key: this._get_entry_key(dir_key, ent, isDir, is_disabled_dir_content),
```
romayalon, line 759: this handles versioning disabled, where is the change for list_versions? do we need to have in _get_version_entry_key an extended implementation?

src/sdk/namespace_fs.js
```diff
@@ -2111,13 +2116,8 @@ class NamespaceFS {
...
     async delete_object_tagging(params, object_sdk) {
         dbg.log0('NamespaceFS.delete_object_tagging:', params);
-        let file_path;
-        if (params.version_id && this._is_versioning_enabled()) {
-            file_path = this._get_version_path(params.key, params.version_id);
-        } else {
-            file_path = this._get_file_path(params);
-        }
         const fs_context = this.prepare_fs_context(object_sdk);
+        const file_path = await this._find_version_path(fs_context, params, true);
```
romayalon, line 2120: have you checked that this change is compatible to S3? especially when versioning is suspended, delete tagging from the version id file is expected per the protocol?

src/sdk/namespace_fs.js
```diff
@@ -2450,7 +2445,6 @@ class NamespaceFS {
      * @returns {string}
      */
      _get_version_entry_key(dir_key, ent) {
-        if (ent.name === config.NSFS_FOLDER_OBJECT_NAME) return dir_key;
```
romayalon, line 2453: where is the fix of the object name when running list_versions and it's a directory object?

src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -639,7 +640,63 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
...
+            version_ids.add(push_res.VersionId);
+            push_res = await s3_uid6.putObject({ Bucket: content_dir_bucket_name, Key: key, Body: body1 });
+            version_ids.add(push_res.VersionId);
+            push_res = await s3_uid6.putObject({ Bucket: content_dir_bucket_name, Key: key, Body: body1 });
+            version_ids.add(push_res.VersionId);
+            const list_res = await s3_uid6.listObjectVersions({ Bucket: content_dir_bucket_name });
+            const key_list_versions = list_res.Versions.filter(v => v.Key === key);
+            assert.equal(key_list_versions.length, version_ids.size);
+            key_list_versions.forEach(v => {
+                assert(version_ids.has(v.VersionId));
```
romayalon, line 695: can you add an assert of the key and the version id do not contain .folder?

src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -639,7 +640,63 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
...
             const get_res = await s3_uid6.getObject({ Bucket: content_dir_bucket_name, Key: key });
             assert.equal(get_res.VersionId, res2.VersionId);
+        });
+
+        mocha.it('content directory - object tagging', async function() {
+            const dir_tagging_key = 'dir_tagging/';
+            const tag_set1 = { TagSet: [{ Key: "key1", Value: "Value1" }] };
+            const tag_set2 = { TagSet: [{ Key: "key2", Value: "Value2" }] };
 
+            await s3_uid6.putBucketVersioning({ Bucket: content_dir_bucket_name, VersioningConfiguration: { MFADelete: 'Disabled', Status: 'Enabled' } });
```
romayalon, line 650: after validating the behavior on AWS, can you add a test for suspended mode as well?

src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -639,7 +640,63 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
...
+            assert.equal(res.TagSet.length, 0);
+        });
+
+        mocha.it('content directory - list object', async function() {
+            const res = await s3_uid6.listObjects({ Bucket: content_dir_bucket_name });
+            assert.equal(res.Contents.length, 10);
+        });
+
+        mocha.it('content directory - list object versions', async function() {
+            const key = "list_key/";
```
romayalon, line 683: do we need a test of list object versions versioning suspended?

src/test/unit_tests/test_bucketspace_versioning.js
```diff
@@ -639,7 +640,63 @@ mocha.describe('bucketspace namespace_fs - versioning', function() {
...
+            const version_ids = new Set();
+            let push_res = await s3_uid6.putObject({ Bucket: content_dir_bucket_name, Key: key, Body: body1 });
+            version_ids.add(push_res.VersionId);
```
romayalon, line 686: can you also add a delete marker of content dir in this test / another test?

src/sdk/namespace_fs.js
nadavMiz: there is no special case for versioned object. we get the .folder file. in the end we format the key to remove the ,folder from the name. also its already in the correct directory level in case of delimiter so no need to change anything

src/sdk/namespace_fs.js
nadavMiz: the change is inside the function. essentially there is no special case for the extended implementation. we use the .folder file as if it is a regular key

src/sdk/namespace_fs.js
nadavMiz: there shouldn't be a difference between enabled and suspended mode. but I will validate this. I will also check what happens if we are in disabled mode and have version as a paramater

romayalon:
@nadavMiz please add a link in the description to all the content directory PRs the original issue, I think it's 8531

src/sdk/namespace_fs.js
nadavMiz: the solution is in _get_version_entry_key. removing the line so the object is processed. we format the name in the format key function at the end of list-objects

src/sdk/namespace_fs.js
```diff
@@ -915,6 +915,17 @@ class NamespaceFS {
                 }
             };
 
+            const format_key_name = obj_info => {
+                if (this._is_hidden_version_path(obj_info.key)) {
+                    obj_info.key = path.normalize(obj_info.key.replace(HIDDEN_VERSIONS_PATH + '/', ''));
+                    obj_info.key = _get_filename(obj_info.key);
+                    set_latest_delete_marker(obj_info);
```
romayalon, line 922: should set_latest_delete_marker() function be a part of a function called format_key_name() ?

src/sdk/namespace_fs.js
romayalon: hey did you check this?

romayalon (approved):
some minor comments, but looks good

src/sdk/namespace_fs.js
nadavMiz: right, it shouldn't. also putting it inside _is_hidden_version_path is not important. I moved it outside

Issue #8816, opened 2025-02-20
notification log doesn't have the etag field entered correctly as compared to the put/copy object operation


rkomandu:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

noobaa d/s build -->noobaa-core-5.18.0-20250218.el9.x86_64 

### Actual behavior
1.
```

upload an object (using put-object) API and then the Etag is the response. In the notification log the "m" is missing, time-xxxx-ino-5vn is only present. can we fix the typo, as this is "mtime-xxxx-ino-5vn" when MD5 is disabled. 

 s3u20k1-api put-object --bucket newbucket-notify-api  --key file-3G --body /root/file_3G
urllib3/connectionpool.py:1061: InsecureRequestWarning: Unverified HTTPS request is being made to host '192.168.0.6'. Adding certificate verification is strongly advised. See: https://urllib3.readthedocs.io/en/1.26.x/advanced-usage.html#ssl-warnings
{
    "ETag": "\"mtime-d7x7np2kvchs-ino-5vn\""
}


{"meta":{"connect":"notify_event.json","name":"notify_event","bucket":"newbucket-notify-api"},"notif":{"Records":[{"eventVersion":"2.3","eventSource":"rk523-13:s3","eventTime":"2025-02-20T10:45:49.499Z","s3":{"s3SchemaVersion":"1.0","object":{"sequencer":"19522f6b93b","key":"file-3G","eTag":"time-d7x7np2kvchs-ino-5v"},"bucket":{"name":"newbucket-notify-api","ownerIdentity":{"principalId":"s3user20k1"},"arn":"arn:aws:s3:::newbucket-notify-api"}},"eventName":"ObjectCreated:Put","userIdentity":{"principalId":"s3user20k1"},"requestParameters":{"sourceIPAddress":"::ffff:192.168.0.50"},"responseElements":{"x-amz-request-id":"m7d7u0xx-23lhnn-73w","x-amz-id-2":"m7d7u0xx-23lhnn-73w"}}]}}

```

### Expected behavior
1.

correct the typo for the notification log 

### Steps to reproduce
1.

posted above

rkomandu:
@alphaprinz 
Another observation as posted below, you are truncating the etag id (at the end).. for ex:

```
[...]

PR #8819, opened 2025-02-20
bucket notifications - wrong etag, size

alphaprinz:
### Explain the changes
1. Etag was mistakenly cut off of first and last character.
2. size was wrong.

1. https://github.com/noobaa/noobaa-core/issues/8816

### Testing Instructions:
1. Upload an object to a bucket with notifications configured.
3. Get its etag and size (eg using head-object).
4. Compare with etag and size field in notification.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_put_object.js
+++ b/src/endpoint/s3/ops/s3_put_object.js
@@ -65,6 +65,8 @@ async function put_object(req, res) {
     }
     s3_utils.set_encryption_response_headers(req, res, reply.encryption);
 
+    res.size_for_notif = size || reply.size;
+
     if (copy_source) {
         // TODO: This needs to be checked regarding copy between diff namespaces
         // In that case we do not have the copy_source property and just read and upload the stream
--- a/src/util/notifications_util.js
+++ b/src/util/notifications_util.js
@@ -422,7 +422,7 @@ function compose_notification_req(req, res, bucket, notif_conf) {
     let eTag = res.getHeader('ETag');
     //eslint-disable-next-line
     if (eTag && eTag.startsWith('\"') && eTag.endsWith('\"')) {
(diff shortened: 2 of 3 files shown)


src/endpoint/s3/ops/s3_put_object.js
```diff
@@ -65,6 +65,8 @@ async function put_object(req, res) {
     }
     s3_utils.set_encryption_response_headers(req, res, reply.encryption);
 
+    res.size_for_notif = size || reply.size;
```
aspandey, line 68: Under which circumstances the size would be 0 and reply.size would be non zero?
The opposite is possible but if we have called putobject to upload some data then "size"sould be there if not then reply.size should also be zero. 
It is possible that the call to upload was not successful so reply.size might be less than size or zero.

alphaprinz: size is zero (and reply.size is not zero) when copy_source is true.

src/util/notifications_util.js
```diff
@@ -422,7 +422,7 @@ function compose_notification_req(req, res, bucket, notif_conf) {
     let eTag = res.getHeader('ETag');
     //eslint-disable-next-line
     if (eTag && eTag.startsWith('\"') && eTag.endsWith('\"')) {
```
aspandey, line 424: It starts with '\"' ?  shouldn't it be '"\' ?    For example: "\test-etag\"
Please keep one example in comment to avoid confusion for code reader.

alphaprinz: The only option currently for quotation is with double quote marks, "".
That is the only test performed in 424.
I have not seen an etag quoted with single quotes '' yet, so I didn't test for it.

aspandey (approved):

PR #8824, opened 2025-02-25
NC | GPFS | Config Files | Create Config Files Size 0 Bug Fix

romayalon:
### Explain the changes
Background to the bug - 
In the current create_config_file() function logic, due to small files rename() low performance we have tried to use GPFS linkat(), GPFS linkat() overrides the link target. Due to this GPFS linkat() optimization, we used the safe_move() mechanism for making the config file creation atomic, before using GPFS safe_move() the destination file should be open (size 0).
In a customer bug we found that when there is an error thrown from one of the steps in the function, we are not cleaning the destination file.

Changes - 
1. On first creation of a file, we will remove the creation of the destination file and the linkat() usage since on GPFS linkat() overrides the link target and will use POSIX link(), link() on race condition will fail if target already exists. By doing this change we will get atomic creation without leaving destination files.

1. [DFBUGS-1522](https://issues.redhat.com/browse/DFBUGS-1522)

### Testing Instructions:
Manual and Artificial reproduction on GPFS - 

a. Add to `/usr/local/noobaa-core/src/util/native_fs_utils.js` the following artificial error throw - 
`throw new Error('Error test');`
Notice that you need to write that error after this line - 
`if (is_gpfs) gpfs_dst_file = await open_file(fs_context, schema_dir, config_path, 'w*', config.BASE_MODE_CONFIG_FILE);`

b. Create a bucket/account using the CLI/S3 - 
`noobaa-cli bucket add --name=bucket1 --path=/path/to/bucket/storage/path --owner account1`

c. List buckets directory and check for buckets size 0 - 
`sudo ls -la /{config-dir-path}/buckets/`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/native_fs_utils.js
+++ b/src/util/native_fs_utils.js
@@ -319,10 +319,9 @@ function get_config_files_tmpdir() {
  * @param {string} config_data 
  */
 async function create_config_file(fs_context, schema_dir, config_path, config_data) {
-    const is_gpfs = _is_gpfs(fs_context);
-    const open_mode = is_gpfs ? 'wt' : 'w';
+    const open_mode = 'w';
+    let open_path;
     let upload_tmp_file;
-    let gpfs_dst_file;
     try {
         // validate config file doesn't exist
         try {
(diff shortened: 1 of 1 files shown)


nadavMiz (approved):
LGTM

PR #8829, opened 2025-02-26
NSFS | versioning | fix nested object concurrency issue

nadavMiz:
### Explain the changes
1. when putting and deleting nested objects at the same time, the delete operation might delete the objects directories inside move_to_dest_versions. in that case we don't recreate the paths if they don't exist. added line to recreate them inside move_to_dest_versions

### Testing Instructions:
1. fixes test issues for test that already exists: `content dir multiple puts of the same key - enabled`
2. run `sudo npx jest test_versioning_concurrency.test.js`

- [ ] Doc added/updated
- [ ] Tests added


romayalon (approved):
LGTM

PR #8843, opened 2025-03-03
Fix RPM build on RHEL8 & friends

tangledbytes:
### Explain the changes
This PR attempts to fix the broken builds of NooBaa RPM on RHEL8.

### Testing Instructions:
1. `make rpm CENTOS_VER=8` should work.
2. `make rpm CENTOS_VER=9` should also work.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/deploy/NVA_build/Base.Dockerfile
+++ b/src/deploy/NVA_build/Base.Dockerfile
@@ -32,6 +32,7 @@ RUN ./src/deploy/NVA_build/clone_s3select_submodules.sh
 RUN ln -s /lib64/libboost_thread.so.1.66.0 /lib64/libboost_thread.so.1.75.0 || true
 #Pass BUILD_S3SELECT down to GYP native build.
 #S3Select will be built only if this parameter is equal to "1".
+#Next step would fail for RHEL8 and derivatives
 RUN GYP_DEFINES="BUILD_S3SELECT=$BUILD_S3SELECT BUILD_S3SELECT_PARQUET=$BUILD_S3SELECT_PARQUET" npm run build
 
 ##############################################################
--- a/src/deploy/RPM_build/RPM.Dockerfile
+++ b/src/deploy/RPM_build/RPM.Dockerfile
@@ -38,6 +38,9 @@ COPY ./src/deploy/standalone/noobaa-logrotate ./src/deploy/standalone/noobaa-log
 COPY ./src/manage_nsfs ./src/manage_nsfs
 COPY ./src/nc ./src/nc
 
+# Install GCC11 toolchain on Centos8 to match the default toolchain of Centos9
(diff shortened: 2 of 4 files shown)


src/deploy/RPM_build/RPM.Dockerfile
```diff
@@ -54,4 +57,5 @@ ENV BUILD_S3SELECT_PARQUET=${BUILD_S3SELECT_PARQUET}
 ENV CENTOS_VER=${CENTOS_VER}
 ENV SRPM_ONLY=${SRPM_ONLY}
 RUN mkdir -p /export
-CMD ./packagerpm.sh /export /build
+# Set GCC Toolset in path - won't exist in RHEL8 but that's OK
```
romayalon, line 60: I believe you meant RHEL9 right?

src/native/s3select/s3select.gyp
```diff
@@ -4,6 +4,7 @@
         'target_name': 's3select',
         'type': 'static_library',
         'cflags_cc!': ['-fno-rtti'],
+        'cflags_cc': ['-std=c++17'],
```
romayalon, line 7: what are the implications of this change?

src/deploy/RPM_build/RPM.Dockerfile
tangledbytes: Ah, you are right.

src/deploy/NVA_build/Base.Dockerfile
```diff
@@ -32,6 +32,7 @@ RUN ./src/deploy/NVA_build/clone_s3select_submodules.sh
 RUN ln -s /lib64/libboost_thread.so.1.66.0 /lib64/libboost_thread.so.1.75.0 || true
 #Pass BUILD_S3SELECT down to GYP native build.
 #S3Select will be built only if this parameter is equal to "1".
+#Next step would fail for RHEL8 and derivates
```
alphaprinz, line 35: Is this comment correct?
I thought the point is to fix RHEL8?

tangledbytes: Actually this dockerfile is not used for RPM generation so doesn't effect this PR. The reason I still left this comment here so that when someday someone tries to build a NooBaa image with Centos8 as base image and when it fails, it won't be a total surprise.

I am not sure if we want to fix it. This dockerfile is used for generating noobaa-core image and test images but none of them _need_ centos8/rhel8 as its base (while the same isn't true for RPMs, both scale and DBS3 need RPMs specifically generated for RHEL8).

The gist is, we should probably fix it iff we plan to keep on supporting Centos8 based docker/OCI images for NooBaa core, which I am not sure if we want to.

alphaprinz: OK, makes sense.
Can you just fix typo derivates -> derivatives

alphaprinz (approved):

PR #8850, opened 2025-03-05
bucket notifications - add notif storage test to health

alphaprinz:
### Explain the changes
1. Add test for notification storage

New output's field looks like this: 
"checks": {
  "notif_storage_threshold_details": {
    "threshold": 0.2,
    "ratio": 0.815,
    "result": "above threshold"
  }
}

1. 

### Testing Instructions:
1. NOTIFICATION_LOG_DIR=/notifications_log NOTIFICATION_SPACE_CHECK_THRESHOLD=0.2  node src/cmd/manage_nsfs.js diagnose health --notif_storage_threshol

- [x] Doc added/updated
- [x] Tests added

### DIFF
--- a/docs/NooBaaNonContainerized/Health.md
+++ b/docs/NooBaaNonContainerized/Health.md
@@ -55,7 +55,7 @@ The `health` command is used to analyze NooBaa health with customizable options.
 
 ```sh
 noobaa-cli diagnose health [--deployment_type][--https_port]
-[--all_account_details][--all_bucket_details][--config_root][--debug]
+[--all_account_details][--all_bucket_details][--all_connection_details][--notif_storage_threshold][--config_root][--debug]
 ```
 ### Flags -
 
@@ -83,6 +83,11 @@ noobaa-cli diagnose health [--deployment_type][--https_port]
     - Type: Boolean
     - Default: false
     - Description: Indicates if health output should contain connection test result.
+
+- `notif_storage_threshold`
+    - Type: Boolean
+    - Default: false
+    - Description: Whether health ouput should check if notification storage FS is below threshold.
 
 - `config_root`
     - Type: String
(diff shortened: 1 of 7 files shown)


src/manage_nsfs/health.js
```diff
@@ -66,6 +66,10 @@ const health_errors = {
         error_code: 'INVALID_CONFIG_DIR',
         error_message: 'Config directory is invalid',
     },
+    LOW_STORAGE: {
+        error_code: 'LOW_STORAGE',
+        error_message: 'FS storage is low.'
```
naveenpaul1, line 71: If this error code only used for notification log file you can add more details like `FS storage is low for notification log directory`

docs/NooBaaNonContainerized/Health.md
```diff
@@ -269,6 +274,15 @@ Output:
         }
       ]
     },
+    "notif_storage_threshold_details": {
+      "invalid_storages": [],
```
naveenpaul1, line 278: do we need `invalid_storages`? I am not sure invalid and valid are applicable here, Since we will have only one path for the log directory. Usually follow this structure when there are possible invalid and valid scenarios, But here I dont think thats the case.

src/manage_nsfs/health.js
```diff
@@ -428,6 +436,25 @@ class NSFSHealth {
         };
     }
 
+    get_notif_storage_threshold_status() {
+        const invalid_storages = [];
```
romayalon, line 440: why it's an array if it's being checked and pushed once?

src/manage_nsfs/health.js
```diff
@@ -66,6 +66,10 @@ const health_errors = {
         error_code: 'INVALID_CONFIG_DIR',
         error_message: 'Config directory is invalid',
     },
+    NOTIF_LOW_STORAGE: {
```
naveenpaul1, line 69: @alphaprinz are we using this const somewhere?

alphaprinz: No, I've removed it.

naveenpaul1 (approved):
LGTM

PR #8839, opened 2025-02-27
NC | CLI | Lifecycle worker

romayalon:
### Explain the changes
This PR adds the NC CLI command used for running the lifecycle worker. 
1.  Added a new CLI command type called lifecycle and options that currently contain only the global options.
2. Started creating a frame for the nc_lifecycle worker process - 
- Block noobaa nc_lifecycle worker from working if NooBaa service is not running.
- list buckets, iterate them and on every bucket that contains a lifecycle policy - handle the rules.
- handle_bucket_rule logic - 
  *  check that the rule is enabled
  * get delete candidates
  * call delete multiple objects 
  * update last_sync of the processed lifecycle rule
- the actual rules processing will be added in the next PRs, currently its implementation is empty.

3. Added 2 new manage_nsfs_cli errors - SystemJsonIsMissing and NooBaaServiceIsNotActive.

Refactoring - 
1. moved nsfsObjectSDK from nsfs.js to its own file src/sdk/nsfs_object_sdk.js and updated its callers.
2. moved NOOBAA_SERVICE_NAME  and service state status check by systectl call from health.js to manage_nsfs_cli_utils.js.  

### Issues: Fixed #xxx / Gap #xxx
Gaps -
1. Add implementation of the processing of the 4 rules in POSIX.
2. Add implementation of the processing of expiration and nonCurrenytays
3. Add notifications usage.
4.  Decide on batch size
5. Endpoints Stats should be supported? currently we will not collect stats on the lifecycle worker process.
6. Add CLI unit tests 
7. Add CLI documentation. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

PR #8853, opened 2025-03-05
NC | lifecycle | add option to abort multiple uploads

nadavMiz:
### Explain the changes
1. added builder function for lifecycle filter
2. added lifecycle rule to abort multipart upload
3. changed delete-candidates to be an object (candidate) containing both delete candidates and abort mpu candidates
4. added function to abort list of filtered mpus
5. added test file test_nc_lifecycle_cli.test.js

### Issues: Fixed #xxx /
continuation of #8839 

### Testing Instructions:

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/manage_nsfs/nc_lifecycle.js
+++ b/src/manage_nsfs/nc_lifecycle.js
@@ -9,6 +9,7 @@ const config = require('../../config');
 const nb_native = require('../util/nb_native');
 const NsfsObjectSDK = require('../sdk/nsfs_object_sdk');
 const ManageCLIError = require('./manage_nsfs_cli_errors').ManageCLIError;
+const path = require('path');
 const { throw_cli_error, get_service_status, NOOBAA_SERVICE_NAME } = require('./manage_nsfs_cli_utils');
 
 // TODO: 
@@ -22,21 +23,23 @@ const { throw_cli_error, get_service_status, NOOBAA_SERVICE_NAME } = require('./
  * @param {import('../sdk/config_fs').ConfigFS} config_fs 
  * @returns {Promise<Void>}
  */
-async function run_lifecycle(config_fs) {
+async function run_lifecycle(config_fs, disable_service_validation) {
     const options = { silent_if_missing: true };
     const system_json = await config_fs.get_system_config_file(options);
-    await throw_if_noobaa_not_active(config_fs, system_json);
+    if (!disable_service_validation) await throw_if_noobaa_not_active(config_fs, system_json);
 
     const bucket_names = await config_fs.list_buckets();
-    const concurrency = 10; // TODO - think about it 
+    const concurrency = 10; // TODO - think about it
     await P.map_with_concurrency(concurrency, bucket_names, async bucket_name => {
         const bucket_json = await config_fs.get_bucket_by_name(bucket_name, options);
         const account = { email: '', nsfs_account_config: config_fs.fs_context, access_keys: [] };
         const object_sdk = new NsfsObjectSDK('', config_fs, account, bucket_json.versioning, config_fs.config_root, system_json);
+        //TODO temporary - need to check if we want to use a real account
+        object_sdk._simple_load_requesting_account();
         await P.all(_.map(bucket_json.lifecycle_configuration_rules,
             async (lifecycle_rule, j) => {
                 dbg.log0('NC LIFECYCLE READ BUCKETS configuration handle_bucket_rule bucket name:', bucket_json.name, 'rule', lifecycle_rule, 'j', j);
-                return handle_bucket_rule(lifecycle_rule, j, bucket_json, object_sdk);
+                return await handle_bucket_rule(config_fs, lifecycle_rule, j, bucket_json, object_sdk);
             }
         ));
     });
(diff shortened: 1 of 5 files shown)


src/manage_nsfs/nc_lifecycle.js
```diff
@@ -60,11 +62,57 @@ async function throw_if_noobaa_not_active(config_fs, system_json) {
...
+}
+
+function _build_lifecycle_filter(params) {
+    /**
+     * @param {*} object_info
+     */
+    return function(object_info) {
+        if (params.filter?.prefix && !object_info.key.startsWith(params.filter.prefix)) return false;
+        if (params.expiration && object_info.age < params.expiration) return false;
+        if (params.filter?.tags && !_file_contain_tags(object_info, params.tags)) return false;
```
romayalon, line 106: missing size filter IIUC

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -170,20 +224,52 @@ async function get_candidates_by_expiration_delete_marker_rule(lifecycle_rule, b
...
+    const nsfs = await object_sdk._get_bucket_namespace(bucket_json.name);
+    const mpu_path = nsfs._mpu_root_path();
+    const filter = lifecycle_rule.filter;
+    const expiration = lifecycle_rule.abort_incomplete_multipart_upload.days_after_initiation;
+    const res = [];
+
+    const filter_func = _build_lifecycle_filter({filter, expiration});
+    let dir_handle;
+    try {
+        dir_handle = await nb_native().fs.opendir(fs_context, mpu_path);
```
romayalon, line 249: let's convert it to list_uploads?

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -73,10 +121,13 @@ async function handle_bucket_rule(config_fs, lifecycle_rule, j, bucket_json, obj
     const should_process_lifecycle_rule = validate_rule_enabled(lifecycle_rule, bucket_json, now);
     if (!should_process_lifecycle_rule) return;
     dbg.log0('LIFECYCLE PROCESSING bucket:', bucket_json.name, '(bucket id:', bucket_json._id, ') rule', util.inspect(lifecycle_rule));
-    const delete_candidates = await get_delete_candidates(bucket_json, lifecycle_rule);
+    const delete_candidates = await get_delete_candidates(bucket_json, lifecycle_rule, object_sdk, config_fs.fs_context);
```
romayalon, line 124:
```suggestion
    const candidates = await get_delete_candidates(bucket_json, lifecycle_rule, object_sdk, config_fs.fs_context);
```

src/test/unit_tests/jest_tests/test_nc_lifecycle_cli.test.js
```diff
@@ -0,0 +1,99 @@
...
+    });
+
+    afterAll(async () => {
+        await fs_utils.folder_delete(`${root_path}/${test_bucket}`);
+        await fs_utils.folder_delete(`${root_path}/${test_bucket2}`);
+        await fs_utils.folder_delete(root_path);
+        await fs_utils.folder_delete(config_root);
+    }, TEST_TIMEOUT);
+
+    it('lifecycle_cli - abort mpu by number of days ', async () => {
```
romayalon, line 73: can we do have a similar test that has both existing prefix and days_after_initiation?
and in the file system we will have one upload key that pass the filter and one that does not?

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -22,21 +23,22 @@ const { throw_cli_error, get_service_status, NOOBAA_SERVICE_NAME } = require('./
...
-    await throw_if_noobaa_not_active(config_fs, system_json);
+    if (!disable_service_validation) await throw_if_noobaa_not_active(config_fs, system_json);
 
     const bucket_names = await config_fs.list_buckets();
     const concurrency = 10; // TODO - think about it 
     await P.map_with_concurrency(concurrency, bucket_names, async bucket_name => {
         const bucket_json = await config_fs.get_bucket_by_name(bucket_name, options);
         const account = { email: '', nsfs_account_config: config_fs.fs_context, access_keys: [] };
         const object_sdk = new NsfsObjectSDK('', config_fs, account, bucket_json.versioning, config_fs.config_root, system_json);
+        object_sdk._simple_load_requesting_account();
```
romayalon, line 37: can you add a TODO here that this is temporary probably and we need to check if it's really want we want to do
Maybe we should take the bucket owner/ create a system owner account/ create a nc lifecycle worker account
CC: @guymguym

src/manage_nsfs/nc_lifecycle.js
nadavMiz: its because its not needed for abort mpu. I will add it

src/manage_nsfs/nc_lifecycle.js
nadavMiz: I think that at this point its better to leave as is. this function is pretty much the same as list_uploads. but it filters the entries while going over the mpus and it allows better to resume going over the mpus if we want to work with batches. its probably better to modify list_uploads and use the same function in both places. but since we don't have much time, and its already implemented, I think we should leave this with a TODO. WDYT?

src/manage_nsfs/nc_lifecycle.js
nadavMiz: added

src/manage_nsfs/nc_lifecycle.js
romayalon: ok so please make sure to add a TODO for merging this double logic

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -60,11 +63,60 @@ async function throw_if_noobaa_not_active(config_fs, system_json) {
...
+
+function _build_lifecycle_filter(params) {
+    /**
+     * @param {*} object_info
+     */
+    return function(object_info) {
+        if (params.filter?.prefix && !object_info.key.startsWith(params.filter.prefix)) return false;
+        if (params.expiration && object_info.age < params.expiration) return false;
+        if (params.filter?.tags && !_file_contain_tags(object_info, params.filter.tags)) return false;
+        //TODO what should happen for abort mpus?
```
romayalon, line 108: why do we need this TODO? if tags and size are not allowed we shouldn't get into the condition

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -60,11 +63,60 @@ async function throw_if_noobaa_not_active(config_fs, system_json) {
     }
 }
 
+function _get_file_age_days(stat) {
```
romayalon, line 66: can you pls add appropriate jsdoc in all the functions added to this file/PR?

romayalon (approved):
minor comments, please fix before merging

PR #8858, opened 2025-03-09
NC | Lifecycle | Add lock to lifecycle worker

romayalon:
### Describe the Problem
NC lifecycle worker should run by a cron job from multiple nodes - 
Currently, we would like only one process/worker will do the lifecycle worker work.
In the future, the work will split by all the different lifecycle workers.
NooBaa will synchronize the cron jobs running on the different nodes by using a file system lock (like NooBaa does for Glacier).

### Explain the Changes
1. nc_lifecycle - created a wrapper function called run_lifecycle_under_lock() that takes a file system lock before running the lifecycle work, used native_fs_utils.lock_and_run() function.
the lock called cluster.lock (similarly to glacier's terms) but it's stored under the config directory. 
3. moved lock_and_run() from manage_nsfs_glacier to native_fs_utils.
4. manage_nsfs - called run_lifecycle_under_lock() instead of run_lifecycle().

1. Gap - add automatic unit tests assets - currently added tests but blocked on [8860](https://github.com/noobaa/noobaa-core/pull/8860) for asserting the response.

### Testing Instructions:
Manual & Artificial test without run time validation - 
1. Add `await P.delay(10000);` inside the lock callback.
2. Tab 1 - Run `noobaa-cli lifecycle --disable_runtime_validation`.
3. Tab 2 - Run `noobaa-cli lifecycle --disable_runtime_validation`.
4. Checked that in the logs of tab 1 I see `run_lifecycle_under_lock acquired lock - start lifecycle'` and after 10 seconds I see `run_lifecycle_under_lock done lifecycle - released lock` and only after the release of the lock I see these log messages in Tab 2.
5. Run `ls /etc/noobaa.conf.d/` - check that the lock file called cluster.lock was created.  

Manual & Artificial test without run time validation - 
1. Change in config.json NC_LIFECYLE_RUN_TIME to be the time now in format HH:MM.
2. Add `await P.delay(10000);` inside the lock callback.
3. Tab 1 - Run `noobaa-cli lifecycle --disable_runtime_validation`.
4. Tab 2 - Run `noobaa-cli lifecycle --disable_runtime_validation`.
5. Checked that in the logs of tab 1 I see `run_lifecycle_under_lock acquired lock - start lifecycle'` and after 10 seconds I see `run_lifecycle_under_lock done lifecycle - released lock` and only after the release of the lock I see these log messages in Tab 2.
6. Run `ls /etc/noobaa.conf.d/` - check that the lock file called cluster.lock was created.  
- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -961,6 +961,27 @@ config.NC_DISABLE_HEALTH_ACCESS_CHECK = false;
 config.NC_DISABLE_POSIX_MODE_ACCESS_CHECK = true;
 config.NC_DISABLE_SCHEMA_CHECK = false;
 
+////////// NC LIFECYLE  //////////
+
+config.LIFECYCLE_LOGS_DIR = '/var/log/noobaa/lifecycle';
+
+// NC_LIFECYCLE_RUN_TIME must be of the format hh:mm which specifies
+// when NooBaa should allow running nc lifecycle process
+// NOTE: This will also be in the same timezone as specified in
+// NC_LIFECYCLE_TZ
+config.NC_LIFECYCLE_RUN_TIME = '01:00';
+
+// NC_LIFECYCLE_RUN_DELAY_LIMIT_MINS configures the delay
+// tolerance in minutes.
+//
+// eg. If the expiry run time is set to 01:00 and the tolerance is
+// set to be 2 mins then the expiry can trigger till 01:02 (unless
+// already triggered between 01:00 - 01:02
+config.NC_LIFECYCLE_RUN_DELAY_LIMIT_MINS = 2;
+
+/** @type {'UTC' | 'LOCAL'} */
+config.NC_LIFECYCLE_TZ = 'LOCAL';
+
 ////////// GPFS //////////
 config.GPFS_DOWN_DELAY = 1000;
 
--- a/src/manage_nsfs/manage_nsfs_glacier.js
+++ b/src/manage_nsfs/manage_nsfs_glacier.js
@@ -8,14 +8,15 @@ const nb_native = require('../util/nb_native');
 const { GlacierBackend } = require('../sdk/nsfs_glacier_backend/backend');
 const { getGlacierBackend } = require('../sdk/nsfs_glacier_backend/helper');
 const native_fs_utils = require('../util/native_fs_utils');
+const { is_desired_time, record_current_time } = require('./manage_nsfs_cli_utils');
 
 const CLUSTER_LOCK = 'cluster.lock';
 const SCAN_LOCK = 'scan.lock';
 
 async function process_migrations() {
     const fs_context = native_fs_utils.get_process_fs_context();
-
-    await lock_and_run(fs_context, CLUSTER_LOCK, async () => {
+    const lock_path = path.join(config.NSFS_GLACIER_LOGS_DIR, CLUSTER_LOCK);
+    await native_fs_utils.lock_and_run(fs_context, lock_path, async () => {
         const backend = getGlacierBackend();
 
         if (
@@ -23,7 +24,8 @@ async function process_migrations() {
             await time_exceeded(fs_context, config.NSFS_GLACIER_MIGRATE_INTERVAL, GlacierBackend.MIGRATE_TIMESTAMP_FILE)
         ) {
             await run_glacier_migrations(fs_context, backend);
-            await record_current_time(fs_context, GlacierBackend.MIGRATE_TIMESTAMP_FILE);
+            const timestamp_file_path = path.join(config.NSFS_GLACIER_LOGS_DIR, GlacierBackend.MIGRATE_TIMESTAMP_FILE);
+            await record_current_time(fs_context, timestamp_file_path);
         }
     });
 }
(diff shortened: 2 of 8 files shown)


nadavMiz (approved):
LGTM

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -3,20 +3,38 @@
...
 
+const CLUSTER_LOCK = 'cluster.lock';
+
+/**
+ * run_lifecycle_under_lock runs the lifecycle workflow under a file system lock
+ * lifecycle workflow is being locked to prevent multiple instances from running the lifecycle workflow
+ * @param {import('../sdk/config_fs').ConfigFS} config_fs 
+ */
+async function run_lifecycle_under_lock(config_fs) {
+    const lock_path = path.join(config_fs.config_root, CLUSTER_LOCK);
```
tangledbytes, line 30: Will configfs always be on a distrubuted FS like GPFS, Ceph, etc? This is true for DBS3 but I am unaware of other deployments.

src/native/fs/fs_napi.cpp
```diff
@@ -1846,6 +1848,16 @@ struct FileFcntlLock : public FSWrapWorker<FileWrap>
                 SetError("invalid lock type");
             }
         }
+        if (info.Length() > 2 && !info[2].IsUndefined()) {
+            auto op_string = info[2].As<Napi::String>().Utf8Value();
+            if (op_string == "THROW") {
+                op = F_OFD_SETLK;
```
tangledbytes, line 1854: We might want to quickly confirm that this works as expected on GPFS and CephFS. Although most likely it would but just to be on the safe side. We had some surprises with flock behaviour on GPFS initially.

src/util/native_fs_utils.js
```diff
@@ -712,6 +712,30 @@ function translate_error_codes(err, entity) {
     return err;
 }
 
+/**
+ * lock_and_run acquires a flock and calls the given callback after
```
tangledbytes, line 716:
```suggestion
 * lock_and_run acquires a fcntl and calls the given callback after
```

Initially we had `flock` but later we replaced it with `fcntl` based lock and I totally forgot to update it in the comment here.

src/manage_nsfs/nc_lifecycle.js
romayalon: you are right, I did this change on https://github.com/noobaa/noobaa-core/pull/8860/files and now took it to this PR as well

src/native/fs/fs_napi.cpp
romayalon: I think we will try to avoid surprises and use the functionality you implemented for Glacier, I just updated the PR with the changes, I had to take some of your code to util file, can you please check?
Also as I asked on Slack, I saw that on Glacier, for restore and migrate you check the timestamp file with exceeded_time() and in expiry you check is_desired_time(), can you please elaborate on the differences? I thought that in my case is_desired_time() should fit but I would be glad to get your opinion here

config.js
```diff
@@ -961,6 +961,27 @@ config.NC_DISABLE_HEALTH_ACCESS_CHECK = false;
...
 config.NC_DISABLE_SCHEMA_CHECK = false;
 
+////////// NC LIFECYLE  //////////
+
+config.LIFECYCLE_LOGS_DIR = '/var/log/noobaa/lifecycle';
+
+// NC_LIFECYLE_RUN_TIME must be of the format hh:mm which specifies
+// when NooBaa should allow running nc lifecycle process
+// NOTE: This will also be in the same timezone as specified in
+// NC_LIFECYLE_RUN_TIME
```
tangledbytes, line 971:
```suggestion
// NC_LIFECYCLE_TZ
```

src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -8,22 +8,24 @@ const nb_native = require('../util/nb_native');
...
+    await native_fs_utils.lock_and_run(fs_context, lock_path, async () => {
         const backend = getGlacierBackend();
 
         if (
             await backend.low_free_space() ||
             await time_exceeded(fs_context, config.NSFS_GLACIER_MIGRATE_INTERVAL, GlacierBackend.MIGRATE_TIMESTAMP_FILE)
         ) {
             await run_glacier_migrations(fs_context, backend);
-            await record_current_time(fs_context, GlacierBackend.MIGRATE_TIMESTAMP_FILE);
+            const timestamp_file_path = path.join(config.NSFS_GLACIER_LOGS_DIR, GlacierBackend.MIGRATE_TIMESTAMP_FILE);
```
tangledbytes, line 27: We need to do the same for restore as well.

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -3,24 +3,63 @@
...
+const { is_desired_time, record_current_time } = require('./manage_nsfs_cli_utils');
 
-// TODO: 
+// TODO:
 // implement 
 // 1. notifications
 // 2. POSIX scanning and filtering per rule
 // 3. GPFS ILM policy and apply for scanning and filtering optimization
 
+const CLUSTER_LOCK = 'cluster.lock';
```
tangledbytes, line 23: Can we add a check to make sure that `config.LIFECYCLE_LOGS_DIR` is never the same as `GLACIER_LOGS_DIR`. Because if someone somehow makes them same then all the glacier operations and lifecycle operations will start to run sequentially because they will have the same path.

This might be ignored as well and we can just say that the admin configured the system inccorrectly.

tangledbytes (approved):
LGTM.

src/manage_nsfs/nc_lifecycle.js
romayalon: This change was a mistake, we must have the lock and timestamp files in the config directory which is a shared root between all the nodes

PR #8862, opened 2025-03-10
NC | lifecycle - add expire rule

nadavMiz:
### Describe the Problem
part of the effort to support lifecycle in NC. add expire rule for lifecycle policy

### Explain the Changes
1. added filter for non versioned objects to be expired based on namespace_fs list_object function
2. added tagging to get_object_info and made a common function to get tags from xattr
3. added tests

continuation of #8853 

### Gaps
list object works in batches. need to support filtering and deleting in batches 

### Testing Instructions:

- [ ] Doc added/updated
- [x] Tests added


romayalon (approved):
LGTM

PR #8867, opened 2025-03-12
NC | Online Upgrade CLI | Validate `expected_version` on `upgrade start`

### LINKED ISSUE #8851: Error message cause should not throw internal error message
### Environment info
- NooBaa Version: noobaa-core-5.18.0-20250222.el9.x86_66
- Platform: Standalone

### Actual behavior
1. # noobaa-cli upgrade start --expected_version 5.13.3 2>/dev/null
{
 "error": {
  "code": "UpgradeFailed",
  "message": "Upgrade request failed",
  "cause": "Error: system does not exist\n  at NCUpgradeManager.upgrade_config_dir (/usr/local/noobaa-core/src/upgrade/nc_upgrade_manager.js:102:33)\n  at async start_config_dir_upgrade (/usr/local/noobaa-core/src/manage_nsfs/upgrade.js:55:29)\n  at async Object.manage_upgrade_operations (/usr/local/noobaa-core/src/manage_nsfs/upgrade.js:23:13)\n  at async main (/usr/local/noobaa-core/src/cmd/manage_nsfs.js:74:13)"
 }
}

### Expected behavior
# noobaa-cli upgrade start --expected_version 5.13.3 2>/dev/null
{
 "error": {
  "code": "UpgradeFailed",
  "message": "Upgrade request failed",
  "cause": "The version provided is lower than the installed version"
 }
}
### Steps to reproduce
1.

### LINKED ISSUE #8852: If the version is not supported proper error should be thrown
### Environment info

NooBaa Version: noobaa-core-5.18.0-20250222.el9.x86_66
Platform: Standalone

### Actual behavior
# noobaa-cli upgrade start --expected_version 5.19.0 2>/dev/null
{
  "response": {
    "code": "UpgradeSuccessful",
    "reply": {
      "message": "config_dir_version on system.json and config_fs.config_dir_version match, nothing to upgrade"
    }
  }
}

### Expected behavior
1. It should be error instead of success

### Steps to reproduce
1.

shirady:
### Describe the Problem
Currently, we do not validate the value of the argument of `expected_version` in noobaa-cli when using `upgrade start` command.

### Explain the Changes
1. Add the function of `validate_expected_version` that would check that we have the flag of `expected_version` and it is from the supported versions we allow. Also, add specific errors to the CLI (`MissingExpectedVersionFlag` and `InvalidArgumentType`) so that eventually we would not report it as `UpgradeFailed`.
2. Move functions that are related to versions to `versions_utils`, and added a helper function `is_valid_sematic_version` to check the structure of the string.
3. Throw an error instead of the message "config_dir_version on system.json and config_fs.config_dir_version match, nothing to upgrade" and change the tests accordingly.
4. Add the `skip_verification` to `BOOLEAN_STRING_OPTIONS` and use it to in `validate_expected_version`. 

### Issues: Fixed #8852, Fixed #8851
1. In the described issues, the example is of using `upgrade start --expected_version <expected_version>` when `expected_version` is not supported: either lower than expected, or at the same version.

#### Automatic Testing
Please run:
- `sudo rm -rf /etc/noobaa.conf.d/; sudo npx jest test_cli_upgrade.test` (only in a test environment).
- `sudo npx jest test_nc_upgrade_manager.test.js`

#### Manual Testing
You can recreate the failing flag options, for example:
-  `sudo node src/cmd/manage_nsfs upgrade start #invalid, missing --expected_version`
- `sudo node src/cmd/manage_nsfs upgrade start --expected_version  9  #invalid (number and not string)`
-  `sudo node src/cmd/manage_nsfs upgrade start --expected_version  bla #invalid not semver structure`
- `sudo node src/cmd/manage_nsfs upgrade start --expected_version '6.18.0' #invalid version (higher)`
- `sudo node src/cmd/manage_nsfs upgrade start --expected_version '4.18.0' #invalid version (lower)`
- `sudo node src/cmd/manage_nsfs upgrade start --expected_version '5.19.0' #nothing to upgrade`, for the following case, we need to have `system.json` created, therefore before running it please create an account and start the NSFS server: (1) Create an account with the CLI: `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path /Users/buckets/ --access_key <access-key> --secret_key <secret-key> --uid <uid> --gid <gid>` - before creating the account need to give permission to the `new_buckets_path`: `chmod 777 /Users/buckets/`
(2) Start the NSFS server with: `sudo node src/cmd/nsfs --debug 5`

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/manage_nsfs/manage_nsfs_validations.js
+++ b/src/manage_nsfs/manage_nsfs_validations.js
@@ -1,6 +1,7 @@
 /* Copyright (C) 2024 NooBaa */
 'use strict';
 
+const pkg = require('../../package.json');
 const config = require('../../config');
 const dbg = require('../util/debug_module')(__filename);
 const net = require('net');
@@ -16,6 +17,7 @@ const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, FROM_FILE, BOOLEAN_STRING_VA
 const { check_root_account_owns_user } = require('../nc/nc_utils');
 const { validate_username } = require('../util/validation_utils');
 const notifications_util = require('../util/notifications_util');
+const version_utils = require('../util/versions_utils');
 
 /////////////////////////////
 //// GENERAL VALIDATIONS ////
@@ -738,6 +740,43 @@ function validate_connection_args(user_input, action) {
     }
 }
 
+////////////////////////////
+//// UPGRADE VALIDATION ////
+///////////////////////////
+
+/**
+ * validate_expected_version checks that expected_version is valid
+ * if it's not valid it throws an error
+ * Note: the cases where we used ManageCLIError errors will 
+ *        not be reported as UpgradeFailed and without adding an event
+ * @param {string} expected_version
+ * @param {boolean} [skip_verification]
+ */
+function validate_expected_version(expected_version, skip_verification = false) {
+    if (!expected_version) {
+        throw_cli_error(ManageCLIError.MissingExpectedVersionFlag);
+    }
+    if (!version_utils.is_valid_semantic_version(expected_version)) {
+        const detail = 'expected_version must have sematic version structure (major.minor.patch)';
+        throw_cli_error(ManageCLIError.InvalidArgumentType, detail);
+    }
+    if (!skip_verification && !version_match_to_current_version(expected_version)) {
+        throw new Error(`expected_version must match the package version ${pkg.version}`);
+    }
+}
+
+/**
+ * version_match_to_current_version checks that the version
+ * is the same as the one that was defined in the pkg
+ * @param {string} expected_version
+ * @returns {boolean}
+ */
+function version_match_to_current_version(expected_version) {
+    const package_version = pkg.version;
+    const diff_from_package_version = version_utils.version_compare(expected_version, package_version);
+    return diff_from_package_version === 0;
+}
+
 
 // EXPORTS
 exports.validate_input_types = validate_input_types;
@@ -751,3 +790,4 @@ exports.validate_whitelist_ips = validate_whitelist_ips;
 exports.validate_flags_combination = validate_flags_combination;
 exports.validate_connection_args = validate_connection_args;
 exports.validate_no_extra_args = validate_no_extra_args;
+exports.validate_expected_version = validate_expected_version;
--- a/src/upgrade/nc_upgrade_manager.js
+++ b/src/upgrade/nc_upgrade_manager.js
@@ -7,7 +7,8 @@ const util = require('util');
 const pkg = require('../../package.json');
 const dbg = require('../util/debug_module')(__filename);
 const { CONFIG_DIR_PHASES } = require('../sdk/config_fs');
-const { should_upgrade, run_upgrade_scripts, version_compare } = require('./upgrade_utils');
+const { should_upgrade, run_upgrade_scripts } = require('./upgrade_utils');
+const { version_compare } = require('../util/versions_utils');
 
 const hostname = os.hostname();
 // prior to 5.18.0 - there is no config dir version, the config dir version to be used on the first upgrade is 0.0.0 (5.17.0 -> 5.18.0)
@@ -108,7 +109,11 @@ class NCUpgradeManager {
         const package_to_version = this.package_version;
         const this_upgrade_versions = { config_dir_from_version, config_dir_to_version, package_from_version, package_to_version };
 
-        if (!should_upgrade(config_dir_from_version, config_dir_to_version)) return { message: 'config_dir_version on system.json and config_fs.config_dir_version match, nothing to upgrade' };
+        if (!should_upgrade(config_dir_from_version, config_dir_to_version)) {
+            const err_message = 'config_dir_version on system.json and config_fs.config_dir_version match, nothing to upgrade';
+            dbg.error(`upgrade_config_dir: ${err_message}`);
+            throw new Error(err_message);
+        }
 
         if (!skip_verification) await this._verify_config_dir_upgrade(system_data, expected_version, hosts_list);
 
--- a/src/util/versions_utils.js
+++ b/src/util/versions_utils.js
@@ -1,0 +1,47 @@
+/* Copyright (C) 2025 NooBaa */
+"use strict";
+
+/**
+ * @param {string} ver
+ */
+function parse_ver(ver) {
+    const stripped_ver = ver.split('-')[0];
+    return stripped_ver.split('.').map(i => Number.parseInt(i, 10));
+}
+
+/**
+ * version_compare compares 2 versions. returns positive if ver1 is larger, negative if ver2, 0 if equal
+ * @param {string} ver1
+ * @param {string} ver2
+ */
+function version_compare(ver1, ver2) {
+    const ver1_arr = parse_ver(ver1);
+    const ver2_arr = parse_ver(ver2);
+    const max_length = Math.max(ver1_arr.length, ver2_arr.length);
+    for (let i = 0; i < max_length; ++i) {
+        const comp1 = ver1_arr[i] || 0;
+        const comp2 = ver2_arr[i] || 0;
+        const diff = comp1 - comp2;
+        // if version component is not the same, return the difference
+        if (diff) return diff;
+    }
+    return 0;
+}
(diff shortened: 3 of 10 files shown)


src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -738,6 +740,39 @@ function validate_connection_args(user_input, action) {
...
+ * @param {string} expected_version
+ */
+function validate_expected_version(expected_version) {
+    if (!expected_version) {
+        throw new Error('expected_version flag is required');
+    }
+    if (!version_utils.is_valid_sematic_version(expected_version)) {
+        throw new Error('expected_version must have sematic version structure (major.minor.patch)');
+    }
+    if (check_if_expected_version_newer_that_package_version(expected_version)) {
```
naveenpaul1, line 759: method name is confusing, It gives a feeling that an error is thrown when the expected version is latest.
And I am not sure if a long method name like this is really helpful or not.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -751,3 +786,5 @@ exports.validate_whitelist_ips = validate_whitelist_ips;
 exports.validate_flags_combination = validate_flags_combination;
 exports.validate_connection_args = validate_connection_args;
 exports.validate_no_extra_args = validate_no_extra_args;
+exports.validate_expected_version = validate_expected_version;
+exports.check_if_expected_version_newer_that_package_version = check_if_expected_version_newer_that_package_version;
```
naveenpaul1, line 790: no need to export.

src/util/versions_utils.js
```diff
@@ -0,0 +1,47 @@
+/* Copyright (C) 2016 NooBaa */
```
naveenpaul1, line 1: 2025

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -738,6 +740,39 @@ function validate_connection_args(user_input, action) {
...
+////////////////////////////
+//// UPGRADE VALIDATION ////
+///////////////////////////
+
+/**
+ * validate_expected_version checks that expected_version is valid
+ * if it's not valid it throws an error
+ * @param {string} expected_version
+ */
+function validate_expected_version(expected_version) {
```
naveenpaul1, line 752: can we move this method also to new util file?

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -738,6 +740,39 @@ function validate_connection_args(user_input, action) {
...
+    }
+}
+
+/**
+ * compare_version_to_current_version checks that the version
+ * is not later that the one that was defined in the pkg
+ * @param {string} expected_version
+ * @returns {boolean}
+ */
+function check_if_expected_version_newer_that_package_version(expected_version) {
```
naveenpaul1, line 770: can we move this method also to new util file?

romayalon:
@shirady 
That's a nice change but I don't see how this answers the bug, the reporter got confused by the error code "SuccessfulUpgrade" although there was no upgrade because 5.19.0 is the current version, this is the error message they got - `"message": "config_dir_version on system.json and config_fs.config_dir_version match, nothing to upgrade"` - this is an error of the user but not NooBaa's, we need to decide if we want to report it as an error or as a response saying nothing to upgrade.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -738,6 +740,39 @@ function validate_connection_args(user_input, action) {
...
+ */
+function validate_expected_version(expected_version) {
+    if (!expected_version) {
+        throw new Error('expected_version flag is required');
+    }
+    if (!version_utils.is_valid_sematic_version(expected_version)) {
+        throw new Error('expected_version must have sematic version structure (major.minor.patch)');
+    }
+    if (check_if_expected_version_newer_that_package_version(expected_version)) {
+        throw new Error(`expected_version cannot be later that the package version ${pkg.version}`);
```
romayalon, line 760: what happens if the expected version equals to pkg.version?

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -738,6 +740,39 @@ function validate_connection_args(user_input, action) {
...
+
+/**
+ * compare_version_to_current_version checks that the version
+ * is not later that the one that was defined in the pkg
+ * @param {string} expected_version
+ * @returns {boolean}
+ */
+function check_if_expected_version_newer_that_package_version(expected_version) {
+    const package_version = pkg.version;
+    const diff_from_package_version = version_utils.version_compare(expected_version, package_version);
```
romayalon, line 772: in the bug, the error is about config_dir_version and not source code version
so that's a nice change but we probably should check how we behave on source code valid upgrade while in the config dir version there is nothing to upgrade because they match

src/manage_nsfs/manage_nsfs_validations.js
shirady: I tried to change it, if you think you have a better suggestion please suggest.

src/manage_nsfs/manage_nsfs_validations.js
shirady: Removed, thanks.

src/util/versions_utils.js
shirady: Updated, thanks.

src/manage_nsfs/manage_nsfs_validations.js
shirady: It's about an argument, the `versions_utils.js` is more general (like a library that we can use).

src/manage_nsfs/manage_nsfs_validations.js
shirady: It's about an argument, the `versions_utils.js` is more general (like a library that we can use).

shirady:
> @shirady That's a nice change but I don't see how this answers the bug, the reporter got confused by the error code "SuccessfulUpgrade" although there was no upgrade because 5.19.0 is the current version, this is the message they got - `"message": "config_dir_version on system.json and config_fs.config_dir_version match, nothing to upgrade"` - this is an error of the user but not NooBaa's, we need to decide if we want to report it as an error or as a response saying nothing to upgrade.

@romayalon I added a case for throwing an error on the message "config_dir_version on system.json and config_fs.config_dir_version match, nothing to upgrade".

In their case in the issue they used RPM `noobaa-core-5.18.0-20250222.el9.x86_66` in it the pkg.version is `5.18.0` and in the suggested code change they would get `UpgradeFailed` when trying to upgrade using the CLI.

src/manage_nsfs/manage_nsfs_validations.js
shirady: I do not fail it from the CLI perspective (I saw in the tests that adding it fails parts of the tests).
I had an assumption that there are cases where the upgrade did not finish, and we wanted to rerun it, and didn't want the CLI to block it on the argument.

src/manage_nsfs/manage_nsfs_validations.js
shirady: I added the error thrown in a separate commit.

romayalon: I think there is a misunderstanding here, 
when upgrading the config dir, the package.version must be equal to the expected_version 
we already throw on this issue in the first check of _verify_config_dir_upgrade(), and this verification can be disabled, this is something we should consider

romayalon (approved):

src/upgrade/nc_upgrade_manager.js
```diff
@@ -108,7 +109,11 @@ class NCUpgradeManager {
         const package_to_version = this.package_version;
         const this_upgrade_versions = { config_dir_from_version, config_dir_to_version, package_from_version, package_to_version };
 
-        if (!should_upgrade(config_dir_from_version, config_dir_to_version)) return { message: 'config_dir_version on system.json and config_fs.config_dir_version match, nothing to upgrade' };
```
romayalon, line 111: @jackyalbo - https://issues.redhat.com/browse/DFBUGS-2801
original change

PR #8871, opened 2025-03-18
Change Log Printing in `namespace_monitor`

shirady:
### Describe the Problem
When we have a couple of namespacestores (for example 2) and the `namespace_monitor` has issues with one of them, I added the bucket name to the error printing so it would be easier to understand which one fails.
In addition, the printing of "has error as expected" seems wrong as we don't expect issues in the motoring.

### Explain the Changes
1. Add the bucket name in case of an error in S3, Azure, and GCP.
2. Change the printing of "has error as expected" to "has an unexpected error".

### Issues:
1. This issue was raised as part of the investigation of the issue [ProjectAbell/abell-tracking #45758](https://github.ibm.com/ProjectAbell/abell-tracking/issues/45758) (GitHub enterprise link).

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/bg_services/namespace_monitor.js
+++ b/src/server/bg_services/namespace_monitor.js
@@ -76,7 +76,7 @@ class NamespaceMonitor {
                 this.update_last_monitoring(nsr._id, nsr.name, endpoint_type);
             } catch (err) {
                 this.run_update_issues_report(err, nsr);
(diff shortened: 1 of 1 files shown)


romayalon (approved):
if you fix it for the s3 flow, you should do that for all other namespace resource types
LGTM

PR #8872, opened 2025-03-18
NC | lifecycle | add expire non current and expire delete marker rules

nadavMiz:
### Explain the Changes
1. add noncurrent_version_expiration rule
2. add expired_object_delete_marker rule
3. restructure tests and add lifecycle versioned bucket tests

continuation of #8862 

### Testing Instructions:

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/manage_nsfs/nc_lifecycle.js
+++ b/src/manage_nsfs/nc_lifecycle.js
@@ -410,15 +410,18 @@ async function throw_if_noobaa_not_active(config_fs, system_json) {
 async function get_candidates(bucket_json, lifecycle_rule, object_sdk, fs_context) {
     const candidates = { abort_mpu_candidates: [], delete_candidates: [] };
     const rule_state = lifecycle_run_status.buckets_statuses[bucket_json.name].rules_statuses[lifecycle_rule.id]?.state || {};
+    let versions_list;
     if (lifecycle_rule.expiration) {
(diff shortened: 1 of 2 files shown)


src/manage_nsfs/nc_lifecycle.js
```diff
@@ -149,7 +149,8 @@ async function process_buckets(config_fs, bucket_names, system_json) {
  */
 async function process_bucket(config_fs, bucket_name, system_json) {
     const bucket_json = await config_fs.get_bucket_by_name(bucket_name, config_fs_options);
-    const account = { email: '', nsfs_account_config: config_fs.fs_context, access_keys: [] };
+    const account_path = config_fs.get_account_path_by_id(bucket_json.owner_account);
```
romayalon, line 152: 1. There is already a function that does it for you - get_identity_by_id(), notice that this function is backwards compatible, therefore it's always the function you should choose
2. Please handle errors, for example the account owner was deleted - we don't want all the lifecycle run to fail

nadavMiz: the case where account is deleted might be tricky. we currently don't handle well any uid/gid which doesn't belong to the bucket owner. we wouldn't be able to add delete markers properly

nadavMiz:
seprated to two different PRs:
#8961 
#8949

Issue #8857, opened 2025-03-06
NSFS | S3 | Presigned URL: Incorrect return code for out-of-range ExpiresIn parameter


hseipp:
### Environment info
- NooBaa Version: **5.17.6-20250228**
- Platform: **CentOS 8 deployed using [https://github.com/IBM/StorageScaleVagrant](https://github.com/IBM/StorageScaleVagrant), Scale 5.2.2**

### Actual behavior

When creating a presigned URL with an expiration date bigger than the [allowed](https://github.com/aws/aws-sdk-net/issues/528#issuecomment-273286734) 7 days (604800 seconds), noobaa returns HTTP status code 400 while it should return 403 as it does with similar invalid requests like when providing a negative `ExpiresIn` parameter. 

```
    def test_object_raw_get_x_amz_expires_out_max_range():
        bucket_name = _setup_bucket_object_acl('public-read', 'public-read')
        client = get_client()
        params = {'Bucket': bucket_name, 'Key': 'foo'}
    
        url = client.generate_presigned_url(ClientMethod='get_object', Params=params, ExpiresIn=609901, HttpMethod='GET')
    
        res = requests.get(url, verify=get_config_ssl_verify()).__dict__
>       assert res['status_code'] == 403
E       assert 400 == 403
```

BTW: 604801 is still allowed as `ExpiresIn` value, an error is only returned with 604802 and bigger.

### Expected behavior

Request returns HTTP error 403.

### Steps to reproduce

Use Ceph s3-tests with the method

```
S3TEST_CONF=s3tests.conf tox -- s3tests_boto3/functional/test_s3.py::test_object_raw_get_x_amz_expires_out_max_range
```

### More information - Screenshots / Logs / Other output

Noobaa log with "all" debuglevel attached

[noobaa-20250306_1.log.gz](https://github.com/user-attachments/files/19105337/noobaa-20250306_1.log.gz)

shirady:
Hi @hseipp,
I tested this against AWS and saw that the error code is 400.
The steps I did (using AWS CLI and curl):
1. Create a bucket in AWS: `aws s3 mb s3://<bucket-name>`
2. Put an object: `aws s3api put-object --bucket s<bucket-name>--key <key-name> --body <path-to-file>`
3. Create presign-URL with `--expires-in` flag: `aws s3 presign s3://s<bucket-name>/<key-name> --expires-in 12000000000000 #should be invalid`
[...]

PR #8493, opened 2024-10-31
Fix pre-signed url issues

romayalon:
### Explain the changes
1. Fixed a bug in _check_expiry_query_v4(), new Date(request_date) is an InvalidDate because requestDate is an amz format and needs to be converted to ISO string before converting to a Date, this issue caused (new Date(request_date).getTime()) always NaN - therefore, we never checked correctly the expiry of the presigned URL, URLs where never expired on this flow.
2. Added a correct S3 error when presigned URL expired - RequestExpired.
3. Fixed a bug that caused us to throw TimeTooSkewed on presigned URLs that were signed more than 15 minutes ago.
4. Added Request is not yet valid error when request time is bigger than now.
5. Added _check_expiry_limit() that checks if the expiry is more than 7 days, throws AuthorizationQueryParametersError if true.

### Issues: Fixed #xxx / Gap #xxx
1. Fixed https://github.com/noobaa/noobaa-core/issues/8490
3. Gap - Automtic tests 
4.  Future request Gap
aws s3 presign s3://bucket1/obj1.txt --expires-in 120 --endpoint-url https://127.0.0.1:6443
change date to be in 2 hours + 
curl --insecure '<url-recieved>'
Curl Output -> 
we have a different internal error currently 'Mismatching date in X-Amz-Credential and X-Amz-Date

### Testing Instructions:
1. Tested manually - 
Pre-requisites - create account and a bucket, start noobaa service, upload an object to the bucket.
```
## Case 1 - expiry > 7 days
aws s3 presign s3://bucket1/obj1.txt --expires-in 12000000000000 --endpoint-url https://127.0.0.1:6443
curl --insecure '<url-recieved>'

# Curl Output
<?xml version="1.0" encoding="UTF-8"?><Error><Code>AuthorizationQueryParametersError</Code><Message>X-Amz-Expires must be less than a week (in seconds); that is, the given X-Amz-Expires must be less than 604800 seconds</Message><Resource>/bucket1/obj1.txt...</RequestId></Error>

## Case 2 - request expired
aws s3 presign s3://bucket1/obj1.txt --expires-in 1 --endpoint-url https://127.0.0.1:6443
curl --insecure '<url-recieved>'
# Curl Output
<?xml version="1.0" encoding="UTF-8"?><Error><Code>AccessDenied</Code><Message>Request has expired</Message>...</Error>

## Case 3 - Future request 
aws s3 presign s3://bucket1/obj1.txt --expires-in 120 --endpoint-url https://127.0.0.1:6443
[...]

PR #8873, opened 2025-03-18
Presigned URL | Add an Error on Expiry with a Negative Number

shirady:
### Describe the Problem
Currently, when a presigned URL is used with a negative number, NooBaa returns an error of `AccessDenied` with a message of `Request has expired`, while in AWS it is `AuthorizationQueryParametersError` with a message of `X-Amz-Expires must be non-negative`.

### Explain the Changes
1. Add the error `AuthorizationQueryParametersErrorNonNegative` (and rename `AuthorizationQueryParametersError` to `AuthorizationQueryParametersErrorWeek`).
2. Add a function `_check_expiry_non_negative` that is called in `check_request_expiry`.
3. Add a test and edit an existing test with the error renaming.
4. Add docs in NC related to the presign URL as an alternative feature to the bucket policy feature.

### Issues:
1. I noticed it as part of testing on issue #8857.
2. This PR continues PRs: #8493, #8503.

### Automatic Tests:
1. Please run: `sudo NC_CORETEST=true NOOBAA_LOG_LEVEL=all node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_nsfs_integration.js -g 'Presigned URL tests'`

### Manual Tests:
5. Add bucket to the account using AWS CLI: `nc-user-1-s3 s3 mb s3://bucket-01` (`bucket-01` is the bucket name in this example)
6. Put an object: `nc-user-1-s3 s3api put-object --bucket s<bucket-name>--key <key-name> --body <path-to-file>`
7. Create presign-URL with `--expires-in` flag: `aws s3 presign s3://<bucket-name>/<key-name> --expires-in -7 #should be invalid`
8. Use the presign-url with curl with `-v` flag to see the HTTP status code: `curl --insecure -v <the URL we received>`

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/util/signature_utils.js
+++ b/src/util/signature_utils.js
@@ -286,6 +286,7 @@ function make_auth_token_from_request(req) {
  */
 function check_request_expiry(req) {
     if (req.query['X-Amz-Date'] && req.query['X-Amz-Expires']) {
+        _check_expiry_non_negative(req.query['X-Amz-Expires']);
         _check_expiry_limit(req.query['X-Amz-Expires']);
         _check_expiry_query_v4(req.query['X-Amz-Date'], req.query['X-Amz-Expires']);
     } else if (req.query.Expires) {
@@ -298,7 +299,20 @@ function check_request_expiry(req) {
 // expiry_seconds limit is 7 days = 604800 seconds
 function _check_expiry_limit(expiry_seconds) {
     if (Number(expiry_seconds) > 604800) {
-        throw new S3Error(S3Error.AuthorizationQueryParametersError);
+        throw new S3Error(S3Error.AuthorizationQueryParametersErrorWeek);
+    }
+}
+
+/**
+ * _check_expiry_non_negative converts the expiry_seconds that we got
+ * from eq.query['X-Amz-Expires'] to number and checks that it is non negative number
+ * (throws an error otherwise)
+ * Note: we don't run it in the condition of epoch time (req.query.expires)
+ * @param {string} expiry_seconds 
+ */
+function _check_expiry_non_negative(expiry_seconds) {
+    if (Number(expiry_seconds) < 0) {
+        throw new S3Error(S3Error.AuthorizationQueryParametersErrorNonNegative);
     }
 }
 
--- a/docs/NooBaaNonContainerized/S3Ops.md
+++ b/docs/NooBaaNonContainerized/S3Ops.md
@@ -99,6 +99,14 @@ Currently we support a couple of options:
 2. Principal by account name: `"Principal": { "AWS": [ "<account-name-1>", "<account-name-2>", ... ,"<account-name-n>"] }`
 3. Principal by account ID: `"Principal": { "AWS": [ "<account-ID-1>", "<account-ID-2>", ... ,"<account-ID-n>"] }`
 
+### Presigned URL Support
+An account can use presigned URLs to grant time-limited access to objects in its bucket without updating the bucket policy. The credentials used by the presigned URL are those of the user who generated the URL (more information in [AWS docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/using-presigned-url.html)).  
+
+#### Simple Test Using AWS CLI and curl:
+1. Pre-requisites: Create account and a bucket, noobaa service should be running, and upload an object to the bucket. 
+2. Create a presigned URL for a minute (the expiration is in seconds): `aws s3 presign s3://<bucket-name>/<key-name> --expires-in 60 --endpoint-url <endpoint-address>` (if needed you can add `--region us-east-1`).
+3. Then you can use: `curl --insecure <URL>` to test the output (`-v` if you want verbose output).
+
 ### Anonymous Requests Support
 
 Anonymous requests are S3 requests made without an access key or a secret key - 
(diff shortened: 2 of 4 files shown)


src/util/signature_utils.js
```diff
@@ -286,6 +286,7 @@ function make_auth_token_from_request(req) {
  */
 function check_request_expiry(req) {
     if (req.query['X-Amz-Date'] && req.query['X-Amz-Expires']) {
+        _check_expiry_non_negative(req.query['X-Amz-Expires']);
```
romayalon, line 289: Did you eventually check if AuthorizationQueryParametersErrorWeek error should be thrown in req.query.expires case? what was the result?

romayalon (approved):

src/util/signature_utils.js
shirady: I mentioned it as a gap, I didn't find a way to manually test it.

romayalon: A manual way to do it is using the sdk
```
'use strict';

const cloud_utils = require('../../util/cloud_utils');
const SensitiveString = require('../../util/sensitive_string');
const dbg = require('../../util/debug_module')(__filename);

const ACCESS_KEY = '';
const SECRET_KEY = '';
const BUCKET = '';
const KEY = '';

function main() {
    const presigned_url_params = {
        endpoint: 'https://s3-us-east-1.amazonaws.com/',
        bucket: new SensitiveString(BUCKET),
        key: KEY,
        access_key: new SensitiveString(ACCESS_KEY),
        secret_key: new SensitiveString(SECRET_KEY)
    };

    const expiry = 10;
    const presigned_url = cloud_utils.get_signed_url(presigned_url_params, expiry);
    dbg.log0('presigned_url', presigned_url);
}

main();
```

I tried it with a week + 10 and it throws an error - 'InvalidExpiryTime: Presigning does not support expiry time greater than a week with SigV4 signing.', so the sdk won't allow it
but you can use this script for checking things with Expire

shirady: I used your suggested script with a minor changes:
1. In `get_signed_url` I changed the region to `us-east-1` to match the region of the AWS bucket.
2. In the endpoint address I changed it to `https://s3.us-east-1.amazonaws.com` (it was `s3-us-east-1`).
3. All arguments are passed as environment variables.
4. The script was added in the path: `src/test/unit_tests/manual_test.js`.
5. I added the environment variables (with `export`) in 2 tabs - AWS and NooBaa.
6. I run: `node ./src/test/unit_tests/manual_test.js`.
7. With the URL I run: `curl --insecure -v <URL>`.

Cases:
on AWS I saw what you saw with `const expiry = 604800 + 10; // week + 10`.
But when I changed it to `const expiry = -7; // negative number` in AWS it is still `X-Amz-Date\=20250319T071138Z\&X-Amz-Expires\=-7\` and the error is `<Error><Code>AuthorizationQueryParametersError</Code><Message>X-Amz-Expires must be non-negative</Message>`.

while in NooBaa it is `Expires\=1742368866\` and the error is `<Error><Code>AccessDenied</Code><Message>Request has expired</Message>` (we added the change only in the `X-Amz-Expires`).
Which means that I didn't crack the way to get `Expires` in the AWS URL.

Attaching the script:
```
'use strict';

const cloud_utils = require('../../util/cloud_utils');
const SensitiveString = require('../../util/sensitive_string');
const dbg = require('../../util/debug_module')(__filename);

const ACCESS_KEY = process.env.AWS_ACCESS_KEY_ID;
const SECRET_KEY = process.env.AWS_SECRET_ACCESS_KEY;
const BUCKET = process.env.BUCKET;
const KEY = process.env.KEY;
const is_aws = process.env.IS_AWS;
console.log('SDSD is_aws', is_aws);
const ENDPOINT = is_aws ? 'https://s3.us-east-1.amazonaws.com/' : 'https://localhost:6443';

function main() {
    const presigned_url_params = {
        endpoint: ENDPOINT,
        bucket: new SensitiveString(BUCKET),
        key: KEY,
        access_key: new SensitiveString(ACCESS_KEY),
        secret_key: new SensitiveString(SECRET_KEY)
    };

    const expiry = -7; // negative number
    // const expiry = 604800 + 10; // week + 10
    const presigned_url = cloud_utils.get_signed_url(presigned_url_params, expiry);
    dbg.log0('presigned_url', presigned_url);
}

main();

```

Another thing I noticed is that AWS closes the connection (`Closing connection`) while NooBaa not (`Connection #0 to host localhost left intact`) I wasn't sure if it was an issue.

shirady: Now I understand that the query string parameter name `Expires` is in sigv2 and `X-Amz-Expires` is in sigv4.
Signature Version 2 is being turned off (deprecated), therefore I cannot test the behavior in AWS.

Anyway, after reading the doc about the `Expires`:

> The time when the signature expires, specified as the number of seconds since the epoch (00:00:00 UTC on January 1, 1970). A request received after this time (according to the server) will be rejected.

So I'm fine with what we suggested, to check it only in the `X-Amz-Expires` condition, as the `Expires` is calculated and will be "Request has expired" error.

links:
[Authenticating Requests: Using Query Parameters (AWS Signature Version 4)](https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-query-string-auth.html)
[Signing and authenticating REST requests (AWS signature version 2)
](https://docs.aws.amazon.com/AmazonS3/latest/API/RESTAuthentication.html)

PR #8874, opened 2025-03-18
Fix id parse for s3_put_bucket_cors

jackyalbo:
### Describe the Problem
Issue while running put-bucket-cors with id - we expected string and got array - now we will support array of single ID

### Explain the Changes
1. 

1. 

### Testing Instructions:
1. run put-bucket-core with id
for example:
```
cors.json:
{
  "CORSRules": [
    {
      "ID": "rule-1",
      "AllowedOrigins": ["http://www.example.com"],
      "AllowedHeaders": ["*"],
      "AllowedMethods": ["PUT", "POST", "DELETE"],
      "MaxAgeSeconds": 3000,
      "ExposeHeaders": ["x-amz-server-side-encryption"]
    },
  ]
}
aws s3api put-bucket-cors --bucket cors-bucket --cors-configuration file://cors.json
```

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/endpoint/s3/ops/s3_put_bucket_cors.js
+++ b/src/endpoint/s3/ops/s3_put_bucket_cors.js
@@ -13,7 +13,7 @@ async function put_bucket_cors(req) {
             allowed_methods: rule.AllowedMethod,
             allowed_origins: rule.AllowedOrigin,
             expose_headers: rule.ExposeHeader,
-            id: rule.ID,
+            id: rule.ID?.[0],
             max_age_seconds: rule.MaxAgeSeconds && parseInt(rule.MaxAgeSeconds, 10),
         }, _.isUndefined)
     );
--- a/src/test/unit_tests/test_s3_ops.js
+++ b/src/test/unit_tests/test_s3_ops.js
@@ -461,6 +461,59 @@ mocha.describe('s3_ops', function() {
         });
     });
 
+    mocha.describe('bucket-cors', function() {
+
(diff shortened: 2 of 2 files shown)


src/endpoint/s3/ops/s3_put_bucket_cors.js
```diff
@@ -13,7 +13,7 @@ async function put_bucket_cors(req) {
             allowed_methods: rule.AllowedMethod,
             allowed_origins: rule.AllowedOrigin,
             expose_headers: rule.ExposeHeader,
-            id: rule.ID,
+            id: rule.ID?.[0],
```
liranmauda, line 16: Not sure I understand why it is an array.
Looking at https://docs.aws.amazon.com/AmazonS3/latest/API/API_CORSRule.html it should be a sting.
will it always be array? 
we can make sure by doing:
```suggestion
            id: (typeof rule.ID === "string") ? rule.ID : rule.ID?.[0],
```
WDYT?

liranmauda (approved):

src/endpoint/s3/ops/s3_put_bucket_cors.js
jackyalbo: I agree about the documentation, that's what confused us in the 1st place. This was tested with AWS cli and with node.js SDK (Added also a test) and as they can do all sorts of future changes we will support only this format for now.

PR #8860, opened 2025-03-09
NC | Lifecycle | Status, Events, Timeout

romayalon:
### Describe the Problem
The NC lifecycle process should -
1. Write lifecycle run status in a log file on finish/fail.
2. Write events on start, finish, failure.
3. Have configurable timeout that is by default 6 hours.

### Explain the Changes
1. Added config.NC_LIFECYLE_TIMEOUT_MS configuration that is set by default to be 6 hours, added P.race[lifecyle_run(), lifecycle_timeout()] in lifecycle_run_under_lock().
2. Added 3 new events - LIFECYCLE_STARTED, LIFECYCLE_SUCCESSFUL and LIFECYCLE_FAILED that will be logged to events.log.
3. Added lifecycle run status capture in nc_lifecycle, per run, bucket and rule. After every run a JSON file will be written under `/var/log/noobaa/lifecycle/`.
The name of the file will be lifecycle_run_1741525171954.json while 1741525171954 is the lifecycle run start timestamp.
The following is an example of the times other statuses captured in a lifecycle run status.
```JSON
{
   "running_host": "my-hostname1",
    "lifecycle_run_times": {
      "start_time": 1741529394579,
      "end_time": 1741529394597,
      "took_ms": 18,
      "buckets_list_start_time": 1741529394584,
      "buckets_list_end_time": 1741529394584,
      "buckets_list_took_ms": 0,
      "buckets_process_took_ms": 13
    },
    "errors": [],
    "buckets_statuses": [
      {
        "bucket_name": "bucket1",
        "bucket_process_times": {
          "start_time": 1741529394585,
          "end_time": 1741529394597,
          "took_ms": 12
          "error": { "code": "", "message": "", "stack": ""}
        },
        "rules_statuses": [
          {
            "lifecycle_rule": "rule1",
            "rule_process_times": {
              "start_time": 1741529394587,
              "end_time": 1741529394597,
              "took_ms": 10,
              "list_candidates_start_time": 1741529394589,
              "list_candidates_took_ms": 0,
              "delete_candidates_start_time": 1741529394589,
              "delete_candidates_took_ms": 0,
              "update_last_sync_time": 1741529394589,
              "update_last_sync_took_ms": 8
            },
            "num_objects_deleted": 0,
            "num_objects_delete_failed": 0,
            "objects_delete_errors": []
            "error": { "code": "", "message": "", "stack": ""} 
          }
        ]
      }
    ]
  }
}
```

### Issues: Fixed #xxx / Gap #xxx
[...]

PR #8875, opened 2025-03-18
NC | Lifecycle | Small configuration adjustments/ fixes

romayalon:
### Describe the Problem
Per feedback received in a design meeting adding some changes + tiny CLI fix.
 
### Explain the Changes
1. Changed the following configurations - 
```
config.NC_LIFECYCLE_TIMEOUT_MS = 8 * 60 * 60;
config.NC_LIFECYCLE_RUN_TIME = 00:00;
config.NC_LIFECYCLE_CONFIG_DIR_NAME = 'lifecycle';
```
2. Moved back the lifecycle cluster lock and the timestamp file to be inside a directory in the config directory which is shared between all the nodes - `/etc/noobaa.conf.d/lifecycle/`. Initially was the case in [8858](https://github.com/noobaa/noobaa-core/pull/8858) but changed it mistakenly on [8860](https://github.com/noobaa/noobaa-core/pull/8860).
3. Renamed cluster.lock to be lifecycle_cluster.lock.
4. Moved the write_stdout_response/throw_cli_error to manage_nsfs because it was wrongly under the lock and these functions do process.exit(). So changed the order to first release the lock and then write the response to stdout.

1. 

### Testing Instructions:
1. Run `noobaa-cli lifecycle` and expect to see successful + status
2. Run `sudo cat /{path_to_config_path}/lifecycle/lifecycle.timestamp` - should contain last lifecycle run timestamp 
3. Run `sudo cat /{path_to_config_path}/lifecycle/lifecycle_cluster.lock` - should be empty
4. Run `sudo cat /var/log/noobaa/lifecycle/lifecycle_run_{lifecycle_run_timestamp}.json | jq` - should have the status

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -1012,20 +1012,21 @@ config.NC_DISABLE_SCHEMA_CHECK = false;
 ////////// NC LIFECYLE  //////////
 
 config.NC_LIFECYCLE_LOGS_DIR = '/var/log/noobaa/lifecycle';
-config.NC_LIFECYCLE_TIMEOUT_MS = 6 * 60 * 60 * 1000;
+config.NC_LIFECYCLE_CONFIG_DIR_NAME = 'lifecycle';
+config.NC_LIFECYCLE_TIMEOUT_MS = 8 * 60 * 60 * 1000;
 
 // NC_LIFECYCLE_RUN_TIME must be of the format hh:mm which specifies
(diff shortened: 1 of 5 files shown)


nadavMiz (approved):
minor comments

PR #8876, opened 2025-03-19
Remove `uuid` module and replace it with the native `crypto.randomUUID()`

liranmauda:
### Explain the Changes
- Remove `uuid` module and replace it with the native `crypto.randomUUID()`


nadavMiz (approved):
LGTM

PR #8877, opened 2025-03-19
NC | LIFECYCLE | Add temporary NC_LIFECYCLE_GPFS_ILM_ENABLED configuration

romayalon:
### Describe the Problem
Currently, we always return an empty array if the underlying file system is GPFS.

### Explain the Changes
1. Add NC_LIFECYCLE_GPFS_ILM_ENABLED configuration and set it to false by default so we will fallback to POSIX directory scan.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


nadavMiz (approved):
LGTM

PR #8878, opened 2025-03-19
[NC | Glacier] Fix migrate.log stat issue

tangledbytes:
### Describe the Problem
This PR fixes the stat bug which was preventing `migrate` calls from running.

### Explain the Changes
Rely on the `fh` from the open call instead of the internal `fh` and add a try catch to ensure no crashes.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


naveenpaul1 (approved):
LGTM

PR #8880, opened 2025-03-19
[NC | NSFS] Fix GPFS xattr read

tangledbytes:
### Describe the Problem
The issue arises from incorrect pointer arithmetic when accessing the buffer containing the extended attribute's value. The code mistakenly uses array indexing to get a single character at an offset, then treats it as a pointer, leading to undefined behaviour.

### Explain the Changes
Use pointer addition to get the starting address of the value: `(gpfsGetXattrRequest.buffer + name_len)` without ROUNDUP as we need the offset to the start of the value and not the end of the buffer.
- [ ] Doc added/updated
- [ ] Tests added


tangledbytes:
@eshelmarc ^ This PR

guymguym (approved):

Issue #8884, opened 2025-03-19
Lifecycle + Notifications | Notifications are being written even if delete operation failed


romayalon:
### Environment info
- NooBaa Version: 5.19
- Platform: Containerized

### Actual behavior
1. Per the code written in handle_bucket_rule() in lifecycle.js, the lifecycle worker writes events even if the delete operation failed.

### Expected behavior
1. The lifecycle worker should write events only if the delete operation succeeded.
2. we need to check that deleted_obj.err_code === undefined before writing the notification. 

### Steps to reproduce
1.

alphaprinz:
Fixed pushed to master (19) and backported to 18

PR #8886, opened 2025-03-20
bucket notifications - lifecycle - ignore failures (issue #8884)

alphaprinz:
### Describe the Problem
Lifecycle notification sends a notification for failed deletions.

### Explain the Changes
1. For each deletion, include the result.
2. If deletion result is failure, don't send a notification.

[1. ](https://github.com/noobaa/noobaa-core/issues/8884)

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -1166,9 +1166,23 @@ module.exports = {
                     deleted_objects: {
                         type: 'array',
                         items: {
-                            $ref: '#/definitions/object_info'
+                            oneOf: [
+                                {
+                                    $ref: '#/definitions/object_info',
+                                },
+                                {
+                                    type: 'object',
+                                    properties: {
+                                        err_code: {
+                                            type: 'string',
+                                            enum: ['AccessDenied', 'InternalError']
+                                        },
+                                        err_message: { type: 'string' }
+                                    }
+                                }
+                            ]
                         }
-                    }
+                    },
                 }
             },
             auth: { system: 'admin' }
--- a/src/server/bg_services/lifecycle.js
+++ b/src/server/bg_services/lifecycle.js
@@ -68,13 +68,13 @@ async function handle_bucket_rule(system, rule, j, bucket) {
         })
     });
 
-    //dbg.log0("LIFECYCLE PROCESSING res =", res);
-
     if (res.deleted_objects) {
 
         const writes = [];
 
         for (const deleted_obj of res.deleted_objects) {
+            //if deletion has failed, don't send a notification
+            if (deleted_obj.err_code) continue;
             for (const notif of bucket.notifications) {
                 if (check_notif_relevant(notif, {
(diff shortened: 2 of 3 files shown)


src/server/bg_services/lifecycle.js
```diff
@@ -68,13 +68,13 @@ async function handle_bucket_rule(system, rule, j, bucket) {
...
 
-    //dbg.log0("LIFECYCLE PROCESSING res =", res);
-
     if (res.deleted_objects) {
 
         const writes = [];
 
-        for (const deleted_obj of res.deleted_objects) {
+        for (let i = 0; i < res.deleted_objects.length; ++i) {
+            if (res.delete_fail[i]) continue;
```
romayalon, line 76: why do we need this delete_fail? and not checking `if (deleted_obj.err_code !== undefined) continue;` here?

romayalon: if possible to do what I suggested, we want NC lifecycle and lifecycle to share this code, see [comment](https://github.com/noobaa/noobaa-core/pull/8883#discussion_r2006074393)

romayalon: I don't think that returning 2 arrays is a good practice, one with all object infos (even if they are not deleted) and the other with booleans if the delete succeeded or failed - IMO it's very confusing
We need to fix the return value of the RPC request which is incorrect - (was added here the return value was added here - https://github.com/noobaa/noobaa-core/commit/7bc42c2638de859cd35f53cada7b4a8ec8e3e4a5) 
please check the reply of delete_multiple_objects https://github.com/noobaa/noobaa-core/blob/master/src/api/object_api.js#L686-L701
I expected to see a similar reply and not have a different reply structure

src/api/object_api.js
```diff
@@ -1465,6 +1465,8 @@ module.exports = {
                     properties: {
                     }
                 },
+                err_code: {type: 'string'},
```
romayalon, line 1468: errors or object infos should be returned inside an array of the reply of delete_multiple_objects functions, not to object_info itself - because there are a lot of functions that return object_info 
please check the reply of delete_multiple_objects https://github.com/noobaa/noobaa-core/blob/master/src/api/object_api.js#L686-L701

romayalon: we should follow the [delete_multiple_objects() reply](https://github.com/noobaa/noobaa-core/blob/master/src/api/object_api.js#L686-L701), 
which means that the array item will be object_info or error, this is the other solution I mentioned in the last comment

alphaprinz: I've used oneOf to include both object info and the error in the same object of the reply array.
The only other option for same-array reply (I can think of) is explicitly writing all needed fields in object_api reply and in implementation in object_server (which would be unnecessarily verbose).

romayalon (approved):
LGTM

PR #8897, opened 2025-03-24
NC | lifecycle | expire objects with batching

nadavMiz:
### Describe the Problem
we populate the delete candidates in memory which might cause issues when the delete candidates list very large. also we want to promote fairness between the lifecycle rules, so one process / bucket will not take most of the processing time. this PR adds populating expiry candidates with batching so we process NC_LIFECYCLE_LIST_BATCH_SIZE of objects each time for every rule in a loop. there is also a loop between bucket so a bucket will process only NC_LIFECYCLE_BUCKET_BATCH_SIZE objects each time

### Explain the Changes
1. add two config variables:
    *  NC_LIFECYCLE_LIST_BATCH_SIZE - determines how much objects we process on each loop for each rule. (it is the limit used in list_objects function) 
    * NC_LIFECYCLE_BUCKET_BATCH_SIZE - determines the amount of objects that each bucket process every time in the loop before moving to the next bucket
2. added two loops, one outer loop for going over the buckets in batches. one inner loop to go over the rules for every bucket
3. added stat property to the lifecycle status both to every rule, and to every bucket. this property saves the key marker where the rule stopped on the last batch, so we can resume in the next batch. and it saves the is_finished property that determines if we finished processing the rule/ bucket and we don't need to resume
4. once all buckets are finished, exit the loop 

### Issues: Gaps 
1. add unit tests
2. change bucket state inside update_status function
3. restructure object counting per bucket. currently its done in the bucket state
4. need better way to decide if we should not have another batch in case expire statement doesn't exists. currently check if rule state is empty  

### Testing Instructions:
manual testing:
1. configure the config variables to be small values for testing. NC_LIFECYCLE_BUCKET_BATCH_SIZE should be bigger than NC_LIFECYCLE_LIST_BATCH_SIZE
5. create a bucket and add lifecycle rule
6. add nuber of objects larger than NC_LIFECYCLE_BUCKET_BATCH_SIZE
7. rules should apply to all buckets in a cyclic order with batching
8. lifecycle run (`sudo cat /var/log/noobaa/lifecycle/lifecycle_run_<timetamp>.json`) should show state of all buckets and rules with `is_finished = true`. in case run finished before all objects were processed, state should include key_marker or where the rule last stopped. you can make the process end prematurely by setting `NC_LIFECYCLE_TIMEOUT_MS` in the timeout to be a small enough value so the run will not finish, but high enough value so at least on cycle will finish (one list-object operation) 

- [ ] Doc added/updated
- [ ] Tests added


src/manage_nsfs/nc_lifecycle.js
```diff
@@ -137,13 +137,21 @@ async function run_lifecycle(config_fs, disable_service_validation) {
...
-    await P.map_with_concurrency(buckets_concurrency, bucket_names, async bucket_name =>
-        await _call_op_and_update_status({
-            bucket_name,
-            op_name: TIMED_OPS.PROCESS_BUCKET,
-            op_func: async () => process_bucket(config_fs, bucket_name, system_json)
-        })
-    );
+    let is_all_buckets_finished = false;
+    while (!is_all_buckets_finished) {
+        dbg.log0("process_buckets");
```
romayalon, line 142: remove log or change it to log2 + something more informative

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -170,16 +181,33 @@ async function process_bucket(config_fs, bucket_name, system_json) {
...
+                            index,
+                            bucket_json,
+                            object_sdk,
+                            should_notify
+                        )
+                    })
+                )
+            );
+            num_objects_processed += config.NC_LIFECYCLE_LIST_BATCH_SIZE;
+            bucket_state.is_finished = Object.values(lifecycle_run_status.buckets_statuses[bucket_json.name].rules_statuses).reduce(
```
romayalon, line 205: please add to the gaps list to move it to the update_status() function

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -170,16 +181,33 @@ async function process_bucket(config_fs, bucket_name, system_json) {
...
-                    rule_id: lifecycle_rule.id,
-                    op_name: TIMED_OPS.PROCESS_RULE,
-                    op_func: async () => process_rule(config_fs, lifecycle_rule, index, bucket_json, object_sdk, should_notify)
-                })
-            )
-        );
+        const bucket_state = lifecycle_run_status.buckets_statuses[bucket_json.name].state || {};
+        let num_objects_processed = 0;
+        while (!bucket_state.is_finished && num_objects_processed < config.NC_LIFECYCLE_BUCKET_BATCH_SIZE) {
+            dbg.log0("process_rules");
```
romayalon, line 187: remove log or change it to log2 + something more informative

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -170,16 +181,33 @@ async function process_bucket(config_fs, bucket_name, system_json) {
...
+                            lifecycle_rule,
+                            index,
+                            bucket_json,
+                            object_sdk,
+                            should_notify
+                        )
+                    })
+                )
+            );
+            num_objects_processed += config.NC_LIFECYCLE_LIST_BATCH_SIZE;
```
romayalon, line 204: something here does not fit
after iterating all the rules of the bucket we add only 1000, seems like it's not adding the correct amount

src/manage_nsfs/nc_lifecycle.js
nadavMiz: I removed it entirely

src/manage_nsfs/nc_lifecycle.js
nadavMiz: I removed it entirely

src/manage_nsfs/nc_lifecycle.js
nadavMiz: done

src/manage_nsfs/nc_lifecycle.js
nadavMiz: I did it by adding the num_of_objects to the bucket stat and updating it inside the expiry. it probably need restructuring. also I didn't count the mpu object to not make it more disorganized. but I can count them if you think its good to have

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -137,13 +137,20 @@ async function run_lifecycle(config_fs, disable_service_validation) {
...
+            await _call_op_and_update_status({
+                bucket_name,
+                op_name: TIMED_OPS.PROCESS_BUCKET,
+                op_func: async () => process_bucket(config_fs, bucket_name, system_json)
+            })
+        );
+        is_all_buckets_finished = Object.values(lifecycle_run_status.buckets_statuses).reduce(
+            (acc, bucket) => acc && (bucket.state?.is_finished),
+            true
+        );
```
romayalon, line 152: do you mind setting is_finished on the run as well? 
so we can know if we processed all the buckets or not

romayalon (approved):

PR #8913, opened 2025-03-27
NC | lifecycle | fix notifications

nadavMiz:
### Describe the Problem
unlike the object server namespace_fs doesn't return object_info as part of the delete_multiple_objects result. object_info is needed to compose delete notification. change to return the entire object_info for the delete candidates object instead of just the key. then combine the object info with the result of delete_multiple_objects to create a new delete_object to pass to create_notification_delete_object. the implementation assumes that delete_multiple_objects return list has the same object order as the object list it got as an input.(same assumption is done in s3_post_bucket_delete)
also unlike the bucket object in containerized lifecycle, bucket_json does not contain the account name. added the account name from object sdk to the bucket_json before passing it to create_notification_delete_object.

### Explain the Changes
1. return object_info instead of just key for expire delete candidates
2. create deleted_obj object that contain both object_info and info from delete_multiple_objects result and pass it to create_notification_delete_object
3. use bucket owner instead of root user as the object sdk account. (this is from #8872 but its needed here as well)
4. add account name from  object_sdk as bucket_owner in bucket_json before passing it to create_notification_delete_object

### Issues
Fixes issues with #8883 

####GAPS
1. lifecycle and nc lifecycle notification are similer, should have a joint function in notifations_util and call it from both
2. modifying bucket_json before passing it to create_notification_delete_object is a bit hacky. should probably modify create_notification_delete_object to account for different structure. should probably be done when combining the code

### Testing Instructions:
1. `sudo npx jest test_nc_lifecycle_cli`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/manage_nsfs/nc_lifecycle.js
+++ b/src/manage_nsfs/nc_lifecycle.js
@@ -17,6 +17,7 @@ const ManageCLIError = require('./manage_nsfs_cli_errors').ManageCLIError;
 const { throw_cli_error, get_service_status, NOOBAA_SERVICE_NAME,
     is_desired_time, record_current_time } = require('./manage_nsfs_cli_utils');
 const SensitiveString = require('../util/sensitive_string');
+const {CONFIG_TYPES} = require('../sdk/config_fs');
 
 // TODO:
 // implement
@@ -161,7 +162,11 @@ async function process_buckets(config_fs, bucket_names, system_json) {
  */
 async function process_bucket(config_fs, bucket_name, system_json) {
     const bucket_json = await config_fs.get_bucket_by_name(bucket_name, config_fs_options);
-    const account = { email: '', nsfs_account_config: config_fs.fs_context, access_keys: [] };
+    const account = await config_fs.get_identity_by_id(bucket_json.owner_account, CONFIG_TYPES.ACCOUNT, {silent_if_missing: true});
+    if (!account) {
+        dbg.warn(`process_bucket - bucket owner ${bucket_json.owner_account} does not exist for bucket ${bucket_name}. skipping lifecycle for this bucket`);
+        return;
+    }
     const object_sdk = new NsfsObjectSDK('', config_fs, account, bucket_json.versioning, config_fs.config_root, system_json);
     await object_sdk._simple_load_requesting_account();
     const should_notify = notifications_util.should_notify_on_event(bucket_json, notifications_util.OP_TO_EVENT.lifecycle_delete.name);
@@ -210,40 +215,6 @@ async function process_rules(config_fs, bucket_json, object_sdk, should_notify) 
     }
 }
 
-async function send_lifecycle_notifications(delete_res, bucket_json, object_sdk) {
-    const writes = [];
-    for (const deleted_obj of delete_res) {
-        if (delete_res.err_code) continue;
-        for (const notif of bucket_json.notifications) {
-            if (notifications_util.check_notif_relevant(notif, {
-                op_name: 'lifecycle_delete',
-                s3_event_method: deleted_obj.delete_marker_created ? 'DeleteMarkerCreated' : 'Delete',
-            })) {
-                //remember that this deletion needs a notif for this specific notification conf
-                writes.push({notif, deleted_obj});
-            }
-        }
-    }
-
-    //required format by compose_notification_lifecycle
-    bucket_json.bucket_owner = new SensitiveString(bucket_json.bucket_owner);
-
-    //if any notifications are needed, write them in notification log file
-    //(otherwise don't do any unnecessary filesystem actions)
-    if (writes.length > 0) {
-        let logger;
-        try {
-            logger = notifications_util.get_notification_logger('SHARED');
-            await P.map_with_concurrency(100, writes, async write => {
-                const notif = notifications_util.compose_notification_lifecycle(write.deleted_obj, write.notif, bucket_json, object_sdk);
-                logger.append(JSON.stringify(notif));
-            });
-        } finally {
-            if (logger) logger.close();
-        }
-    }
-}
-
 /**
  * process_rule processes the lifecycle rule for a bucket
  * TODO - implement notifications for the deleted objects (check if needed for abort mpus as well)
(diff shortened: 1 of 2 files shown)


src/manage_nsfs/nc_lifecycle.js
```diff
@@ -161,7 +161,11 @@ async function process_buckets(config_fs, bucket_names, system_json) {
  */
 async function process_bucket(config_fs, bucket_name, system_json) {
     const bucket_json = await config_fs.get_bucket_by_name(bucket_name, config_fs_options);
-    const account = { email: '', nsfs_account_config: config_fs.fs_context, access_keys: [] };
+    const account = await config_fs.get_identity_by_id(bucket_json.owner_account, undefined, {silent_if_missing: true});
```
romayalon, line 164: type should be account so it'll be backwards compatible

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -210,23 +214,31 @@ async function process_rules(config_fs, bucket_json, object_sdk, should_notify)
     }
 }
 
-async function send_lifecycle_notifications(delete_res, bucket_json, object_sdk) {
+function create_notification_delete_object(delete_res, delete_obj_info) {
```
romayalon, line 217: 1. please move to notifications util section
2. please add jsdoc

romayalon: do we need to pass the version id of the delete marker in the notification as well? what about deleted_delete_marker? will we need this info?

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -210,23 +214,31 @@ async function process_rules(config_fs, bucket_json, object_sdk, should_notify)
...
+    return {
+        ...delete_obj_info,
+        created_delete_marker: delete_res.created_delete_marker
+    };
+}
+
+async function send_lifecycle_notifications(delete_res, delete_candidates, bucket_json, object_sdk) {
```
romayalon, line 224: please add jsdoc

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -210,23 +214,31 @@ async function process_rules(config_fs, bucket_json, object_sdk, should_notify)
...
-        if (delete_res.err_code) continue;
+    for (let i = 0; i < delete_res.length; ++i) {
+        if (delete_res[i].err_code) continue;
         for (const notif of bucket_json.notifications) {
             if (notifications_util.check_notif_relevant(notif, {
                 op_name: 'lifecycle_delete',
-                s3_event_method: deleted_obj.delete_marker_created ? 'DeleteMarkerCreated' : 'Delete',
+                s3_event_method: delete_res[i].delete_marker_created ? 'DeleteMarkerCreated' : 'Delete',
             })) {
+                const deleted_obj = create_notification_delete_object(delete_res[i], delete_candidates[i]);
```
romayalon, line 233: are you sure that the delete_res[i] and delete_candidates[i] are the same object?

src/manage_nsfs/nc_lifecycle.js
nadavMiz: right its not part of the object_info because we get the object info of the object to delete. I can pass it from the res. there are many nuances I am adding tests now

src/manage_nsfs/nc_lifecycle.js
nadavMiz: they should match. it is what we do in s3_post_bucket_delete. the alternative is to modify the return object of of delete_multiple_objects to be like in containerized. for this PR I wanted to match as much as possible to s3_post_bucket_delete. as I don't know why this way was decided. currently it shouldn't matter as we can't have both delete marker and regular delete at the same time. but I would still validate this

src/manage_nsfs/nc_lifecycle.js
nadavMiz: changed the type to account

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -323,6 +294,70 @@ async function abort_mpus(candidates, object_sdk) {
...
+}
+
+/**
+ *
+ * @param {Object[]} delete_res array of delete results
+ * @param {Object[]} delete_candidates array of delete candidates object info
+ * @param {Object} bucket_json
+ * @param {Object} object_sdk
+ * @returns {Promise<Void>}
+ * NOTE implementation assumes delete_candidates and delete_res uses the same index. this assumtion is also made in
```
alphaprinz, line 322: typo - assumption

src/test/unit_tests/jest_tests/test_nc_lifecycle_cli.test.js
```diff
@@ -255,8 +270,8 @@ describe('noobaa cli - lifecycle', () => {
             }
         ];
         await bucketspace_fs.set_bucket_lifecycle_configuration_rules({ name: test_bucket, rules: lifecycle_rule });
-        create_object(test_bucket, test_key1, 100, true);
-        create_object(test_bucket, test_key2, 100, false);
+        create_object_lifecycle(test_bucket, test_key1, 100, true);
```
alphaprinz, line 273: create_object_lifecycle is async. Don't you need an await when you call it?

alphaprinz (commented):
Had some comments, otherwise LGTM

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -161,7 +162,11 @@ async function process_buckets(config_fs, bucket_names, system_json) {
  */
 async function process_bucket(config_fs, bucket_name, system_json) {
     const bucket_json = await config_fs.get_bucket_by_name(bucket_name, config_fs_options);
-    const account = { email: '', nsfs_account_config: config_fs.fs_context, access_keys: [] };
+    const account = await config_fs.get_identity_by_id(bucket_json.owner_account, CONFIG_TYPES.ACCOUNT, {silent_if_missing: true});
+    if (!account) {
+        dbg.warn("process_bucket - bucket owner does not exist. skipping lifecycle for this bucket");
```
romayalon, line 167: please add to the log the bucket name and owner_account so we will know which bucket had this issue

romayalon (approved):

PR #8917, opened 2025-03-28
bucket notifications - fixes for CompleteMultipartUploadResult etag and statusCode check

alphaprinz:
### Describe the Problem
1. etag is missing from CompleteMultipartUploadResult notification
2. notification are created for failed requests

### Explain the Changes
1. Add reply of CompleteMultipartUploadResult to compose_notification_req so we can take etag.
2. Don't write notification logs is http status code is not OK (ie in the 200s).

1. 

### Testing Instructions:
1. Run multipart req. eg (assuming bucket has notification configured)-
s3 api create-multipart-upload --bucket <bucket> --key <key>
(get multipart id)
s3 api upload-part --bucket <bucket> --key <key> --part-number 1 --body <filename> --upload-id "<upload id>"
(get etag for part1, edit parts.json)
complete-multipart-upload --bucket <bucket> --key <key> --multipart-upload file://parts.json --upload-id  "<upload id>"
(examine notification)
2. Perform failed s3 req. Eg above example with bad etag in parts.json.

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/test/unit_tests/test_notifications.js
+++ b/src/test/unit_tests/test_notifications.js
@@ -52,6 +52,8 @@ let http_server = null;
 let server_done = false;
 let expected_bucket;
 let expected_event_name;
+let expected_key;
+let expected_eTag;
 
 // eslint-disable-next-line max-lines-per-function
 mocha.describe('notifications', function() {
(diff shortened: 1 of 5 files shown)


src/test/unit_tests/test_notifications.js
```diff
@@ -171,16 +189,53 @@ mocha.describe('notifications', function() {
...
+                        PartNumber: 1
+                    }]
+                }
+            });
+
+            await notify_await_result({
+                bucket_name: bucket,
+                event_name: 'ObjectCreated:CompleteMultipartUpload',
+                key: "mp1",
+                etag: res_complete.ETag
```
aspandey, line 223: res_complete.Etag and res_upload.Etag should be same.
Do we need to put some assert condition here before notify result?

alphaprinz: No, that's not correct.
res_upload.Etag is md5(part body), similar to put_object's etag.
res_complete.Etag is md5 of concatenated part's etags.
AWS docs don't explain this explicitly, [this](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html) only has - 
"This ETag is not necessarily an MD5 hash of the object data."

aspandey (approved):

PR #8920, opened 2025-03-30
NC | IAM | Return Empty List On Unimplemented Operations (`list-group-for-user`)

shirady:
### Describe the Problem
Currently, we have implemented a couple of operations in IAM that are related to Users and Access keys, but for the rest of the cases, we fail the request with the error `NotImplemented`. There are automated tools that use it and under the hood use the operation `list-group-for-user`.

### Explain the Changes
1. Add the operation `list-group-for-user` to the constants and add a handler that always returns an empty list on existing user.

### Issues:
1. Currently we throw the error `NotImplemented` on `list-group-for-user` (and on every unsupported operation).

GAP - we would like to do this kind of implementation for other list operation in IAM as well (in a separate PR).

#### Automatic Tests:
Please run: `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_nc_iam_basic_integration.js`

#### Manual Tests
1. Create an account with noobaa CLI: `sudo node src/cmd/manage_nsfs account add --name <account-name> --new_buckets_path /Users/buckets/ --access_key <access-key> --secret_key <secret-key> --uid <uid> --gid <gid>`
2. Start the NSFS server (using debug mode and the port for IAM): `sudo node src/cmd/nsfs --debug 5 --https_port_iam 7005`
3. Create the alias for IAM service: 
`alias nc-user-1-iam='AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:7005'`
4. Create a user: `nc-user-1-iam iam create-user --user-name Bob` 
5. Check the `list-group-for-user`: `nc-user-1-iam iam list-groups-for-user --user-name Bob` (and expect to see empty list)
6. Check the `list-group-for-user` on non-exsiting-user: `nc-user-1-iam iam list-groups-for-user --user-name non-existing-user` (and expect to see an error `NoSuchEntity`)

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/test/unit_tests/test_nc_iam_basic_integration.js
+++ b/src/test/unit_tests/test_nc_iam_basic_integration.js
@@ -11,8 +11,11 @@ const { TMP_PATH, generate_nsfs_account, get_new_buckets_path_by_test_env, gener
          get_coretest_path } = require('../system_tests/test_utils');
 const { ListUsersCommand, CreateUserCommand, GetUserCommand, UpdateUserCommand, DeleteUserCommand,
         ListAccessKeysCommand, CreateAccessKeyCommand, GetAccessKeyLastUsedCommand,
-        UpdateAccessKeyCommand, DeleteAccessKeyCommand } = require('@aws-sdk/client-iam');
+        UpdateAccessKeyCommand, DeleteAccessKeyCommand,
+        ListGroupsForUserCommand } = require('@aws-sdk/client-iam');
 const { ACCESS_KEY_STATUS_ENUM } = require('../../endpoint/iam/iam_constants');
+const IamError = require('../../endpoint/iam/iam_errors').IamError;
+
 
 const coretest_path = get_coretest_path();
 const coretest = require(coretest_path);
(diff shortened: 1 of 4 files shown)


src/test/unit_tests/test_nc_iam_basic_integration.js
```diff
@@ -229,6 +230,40 @@ mocha.describe('IAM basic integration tests - happy path', async function() {
...
+            // delete a user
+            const input = {
+                UserName: username3
+            };
+            const command = new DeleteUserCommand(input);
+            const response = await iam_account.send(command);
+            _check_status_code_ok(response);
+        });
+
+        mocha.it('list groups for user - should be empty', async function() {
```
alphaprinz, line 257: I would add this check also in between of user create and delete.
Otherwise LGTM

alphaprinz (approved):

PR #8922, opened 2025-04-01
Refactor | Remove deprecated pwhile from the code in favor of async/await

liranmauda:
### Explain the Changes

Refactor | Remove deprecated pwhile from the code in favor of async/await

### DIFF
--- a/src/test/system_tests/test_cloud_pools.js
+++ b/src/test/system_tests/test_cloud_pools.js
@@ -83,7 +83,7 @@ function init_s3() {
 }
 
 
-function list_all_s3_objects(bucket_name) {
+async function list_all_s3_objects(bucket_name) {
     // Initialization of IsTruncated in order to perform the first while cycle
     const listObjectsResponse = {
         is_truncated: true,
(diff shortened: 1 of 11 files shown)


src/tools/events_generator.js
```diff
@@ -238,37 +234,34 @@ EventsGenerator.prototype.generate_audit = function(num, cat) {
...
-    let count = 0;
-    return P.pwhile(() => count < num,
-        () => {
-            count += 1;
-            const chosen = Math.floor(Math.random() * (logs_size));
-            return P.resolve()
-                .then(() => Dispatcher.instance().activity(_.clone(events_pool[chosen])))
-                .then(() => P.delay(1000));
-        });
+    for (let count = 0; count < num; count++) {
```
jackyalbo, line 237: We usually use ++count
```suggestion
    for (let count = 0; count < num; ++count) {
```

jackyalbo: fix in the rest of the code as well

jackyalbo (approved):
LGTM. also here there is a lot of code we can get rid of in the future.

src/test/system_tests/test_cloud_pools.js
```diff
@@ -92,37 +92,33 @@ function list_all_s3_objects(bucket_name) {
...
-                        console.error(err);
-                        throw err;
-                    });
-            })
-        .then(() => {
-            console.log('list_all_s3_objects - current cloud objects:', listObjectsResponse.objects);
-            return listObjectsResponse.objects;
-        });
+    while (listObjectsResponse.is_truncated) {
+        listObjectsResponse.is_truncated = false;
```
jackyalbo, line 96: does this need to be inside the while?

PR #8927, opened 2025-04-02
NC | NSFS | Fix the `get_bucket_encryption` to Return Object

shirady:
### Describe the Problem
In NC, the request of `get-bucket-encryption` will fail on `TypeError: Cannot read properties of undefined (reading 'encryption')` if we didn't `put-bucket-encryption` before.

### Explain the Changes
1. By default, we will fail the request with `ServerSideEncryptionConfigurationNotFoundError`, aligned with PR #8716 . 
2. Edit existing unit tests in `test_bucketspace_fs.js`.
3. Not related to NC changes:
  - Add a test (only runs in containerized).
  - Improve docs related to running `core_tests` locally.

### Issues:
1. Currently, when a client tries to use the API of `get-bucket-encryption` in NC it fails with `InternalError` and in the logs, we can see that `TypeError: Cannot read properties of undefined (reading 'encryption')`.

#### Unit Test:
Please run:
`sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha ./src/test/unit_tests/test_bucketspace_fs.js`

Note: I also added a test in containerized (although the changes are not related to it):
Use this [guide](https://github.com/noobaa/noobaa-core/blob/master/docs/dev_guide/run_coretest_locally.md) for local run and run: `./node_modules/mocha/bin/mocha.js src/test/unit_tests/test_s3_ops.js NOOBAA_LOG_LEVEL=all`

#### Manual Testing Instructions
3. Create the alias for S3 service:`alias nc-user-1-s3=‘AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:6443’`
5. Add bucket to the account using AWS CLI: `nc-user-1-s3 s3 mb s3://bucket-01-04` (`bucket-01-04` is the bucket name in this example)
6. Get bucket encryption: `nc-user-1-s3 s3api get-bucket-encryption --bucket bucket-01-04` (should throw `ServerSideEncryptionConfigurationNotFoundError`).

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -594,7 +594,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             const { name } = params;
             dbg.log0('BucketSpaceFS.get_bucket_encryption: Bucket name', name);
             const bucket = await this.config_fs.get_bucket_by_name(name);
-            return bucket.encryption;
+            return { encryption: bucket.encryption };
         } catch (err) {
             throw translate_error_codes(err, entity_enum.BUCKET);
         }
--- a/docs/dev_guide/run_coretest_locally.md
+++ b/docs/dev_guide/run_coretest_locally.md
@@ -58,6 +58,13 @@ Another option, which is less recommended, is to connect to postgres container a
 Notes:
 * For running `psql` commands you would need to install it on your machine (better with the matching version of postgres).
 * In case you already run postgres on your machine (not the mentioned container) it might raise some issues, better remove it (in MacOs using the Activity Monitor and search for postgres).
+Example: you stopped the docker run, and you tried to connect to postgres, but got an error (probably password authentication failed). As it expected to see the following message when we don't have a postgres container:
+
+```
+psql: error: connection to server at "127.0.0.1", port 5432 failed: Connection refused
+	Is the server running on that host and accepting TCP/IP connections?
+```
+
 * In case a test failed with "error: database "coretest" already exists" it means that you probably forgot to drop the DB table, and you can stop (ctrl + c) and rerun the `docker run` command mentioned above and then run the test again.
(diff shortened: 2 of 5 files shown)


src/sdk/bucketspace_fs.js
```diff
@@ -594,6 +594,16 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             const { name } = params;
             dbg.log0('BucketSpaceFS.get_bucket_encryption: Bucket name', name);
             const bucket = await this.config_fs.get_bucket_by_name(name);
+            // on undefined return a mock
+            if (bucket.encryption === undefined) {
+                const mock_encryption = {
+                    encryption: {
+                        algorithm: 'AES256',
```
romayalon, line 601: Hi few questions - 
1. Is this matching the default encryption on AWS? 
2. Can you define these values in a common place? I see it's also the default for example on bucketspace_nb? why default is decided on bucketspace layer ? looks like if it was on a higher level we wouldn't have this bug initially
3. Can you call it default instead of mock?

shirady: 1. Yes, I created a bucket and then run `aws s3api get-bucket-encryption --bucket <bucket-name>`
```
{
    "ServerSideEncryptionConfiguration": {
        "Rules": [
            {
                "ApplyServerSideEncryptionByDefault": {
                    "SSEAlgorithm": "AES256"
                },
                "BucketKeyEnabled": false
            }
        ]
    }
}
```
2. I didn't find a good place. I thought about `src/endpoint/s3/s3_constants.js` but it is not suitable as it in the endpoint. I'm open to suggestions. It is the default for data buckets, as you can see in PR #8716. I didn't want to change the behavior of what was so far in NC, so just for users that need this request not to fail I will return the default, else it will return an error (`ServerSideEncryptionConfigurationNotFoundError` instead of the `InternalError`, which was a bug).
3. I changed it from "mock" to "default" as suggested.

romayalon: I found similar S3 default values in s3_utils.js

shirady: I removed the mock, they are fine with the error `ServerSideEncryptionConfigurationNotFoundError` (not the `InternalError`).

romayalon (approved):

PR #8929, opened 2025-04-02
NC | Lifecycle | Convert NCLifecycle to a class

romayalon:
### Explain the Changes
1. Converted NClifecycle to be a class and changed the calls accordingly.
2. native_fs_utils.read_file() - for tests purposes - this function always parsed the file as a json, added a param so it can return the string data without parsing it.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/test/system_tests/test_utils.js
+++ b/src/test/system_tests/test_utils.js
@@ -737,12 +737,15 @@ async function clean_config_dir(config_fs, custom_config_dir_path) {
  * @param {nb.NativeFSContext} fs_context 
  * @param {String} file_path 
  * @param {Object} file_data 
+ * @param {{stringify_json?: Boolean}} [options={}] 
  */
-async function create_file(fs_context, file_path, file_data) {
+async function create_file(fs_context, file_path, file_data, options = {}) {
+    const buf = Buffer.from(options?.stringify_json ? JSON.stringify(file_data) : file_data);
     await nb_native().fs.writeFile(
         fs_context,
         file_path,
-        Buffer.from(JSON.stringify(file_data)), {
+        buf,
+        {
             mode: native_fs_utils.get_umasked_mode(config.BASE_MODE_FILE)
         }
     );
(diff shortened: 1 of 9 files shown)


src/test/system_tests/test_utils.js
```diff
@@ -737,12 +737,15 @@ async function clean_config_dir(config_fs, custom_config_dir_path) {
  * @param {nb.NativeFSContext} fs_context 
  * @param {String} file_path 
  * @param {Object} file_data 
+ * @param {{stringify_json?: Boolean}} [options={}] 
  */
-async function create_file(fs_context, file_path, file_data) {
+async function create_file(fs_context, file_path, file_data, options = {}) {
+    const buf = Buffer.from(options?.stringify_json ? JSON.stringify(file_data) : file_data);
```
nadavMiz, line 743: since previously we always called  stringify, I assume it the more common option. maybe we should set the default to be stringify_json = true

src/test/unit_tests/jest_tests/test_nc_lifecycle_cli.test.js
```diff
@@ -793,6 +794,7 @@ describe('noobaa cli - lifecycle notifications', () => {
 
         notification_log_entries.forEach(notification => {
             const record = notification.notif.Records[0];
+            console.log('ROMY record', record);
```
nadavMiz, line 797: remove

src/test/unit_tests/jest_tests/test_nc_lifecycle_cli.test.js
```diff
@@ -809,8 +811,9 @@ describe('noobaa cli - lifecycle notifications', () => {
         await create_object(object_sdk, test_bucket, test_key1, 100, false);
         await create_object(object_sdk, test_bucket, test_key2, 100, false);
 
-        await exec_manage_cli(TYPES.LIFECYCLE, '', {disable_service_validation: 'true', disable_runtime_validation: 'true', config_root}, undefined,
-            {NOTIFICATION_LOG_DIR: tmp_lifecycle_logs_dir_path});
+        await config_fs.create_config_json_file(JSON.stringify({ NOTIFICATION_LOG_DIR: tmp_lifecycle_logs_dir_path }));
```
nadavMiz, line 814: if its the same for all notifications logs, maybe its better to move the beforeEach?

src/test/system_tests/test_utils.js
romayalon: I don't think so.. on a function that is called create_file() we can't assume that the default it a json file..
what I thought is to create another function that is called create_json_file() that will pass a striginified data, but I had a feeling it's redundant, WDYT?

nadavMiz: having it as an option is fine

nadavMiz (approved):
LGTM

PR #8932, opened 2025-04-03
bucket notifications - facilitate notif conf to override connection

alphaprinz:
### Describe the Problem
We want to allow user to describe an external Kafka server in a connection file, but allow to specify to specify Kafka topic per notification conf (instead of in the connection).

### Explain the Changes
TopicConfigurations' TopicArn field can be of this syntax "kafka:::topic/<connection>/<Kafka topic>", where:
-if connection is not specified, the default account's connection is used.
-Kafka topic in TopicArn takes precedence over Kafka topic in connection.
-backward compatible - you can still put just the connection in TopicArn, like now

1. 

### Testing Instructions:
1. Set "default_connection" field in NC user.

2. Create a connection file for kafka (with or without topic field).
{
	"notification_protocol": "kafka",
	"name": "k",
	"topic": "mytopic",
	"kafka_options_object": "{\"metadata.broker.list\":\"localhost:9092\"}"
}

3. Configure a notification with override object:
{
    "TopicConfigurations": [
        {
	    "Id": "first",
            "TopicArn": "kafka:::topic//override",
	    "Events": []
        }
    ]
}
4. Create an event (ie upload object to bucket with notification).
5. The notification will be sent with user's default connection with topic "override".

- [ ] Doc added/updated - TBD
- [ ] Tests added

### DIFF
--- a/docs/NooBaaNonContainerized/NooBaaCLI.md
+++ b/docs/NooBaaNonContainerized/NooBaaCLI.md
@@ -75,7 +75,7 @@ The `account add` command is used to create a new account with customizable opti
 ```sh
 noobaa-cli account add --name <account_name> --uid <uid> --gid <gid> [--user]
 [--new_buckets_path][--access_key][--secret_key][--fs_backend]
-[--allow_bucket_creation][--force_md5_etag][--anonymous][--from_file][--iam_operate_on_root_account]
+[--allow_bucket_creation][--force_md5_etag][--anonymous][--from_file][--iam_operate_on_root_account][--default_connection]
 ```
 #### Flags -
 - `name` (Required)
@@ -135,6 +135,10 @@ noobaa-cli account add --name <account_name> --uid <uid> --gid <gid> [--user]
     - Type: Boolean
     - Description: Specifies if the account allowed to create root accounts using the IAM API (the default behavior is to create of IAM accounts). See - [IAM - Root Accounts Manager](./../design/iam.md#root-accounts-manager).
 
+- `default_connection`
+    - Type: String
+    - Description: A default account for Kafka external servers. See bucket-notifications.md.
+
 ### Update Account
 
 The `account update` command is used to update an existing account with customizable options.
@@ -143,7 +147,7 @@ The `account update` command is used to update an existing account with customiz
 ```sh
 noobaa-cli account update --name <account_name> [--new_name][--uid][--gid][--user]
 [--new_buckets_path][--access_key][--secret_key][--regenerate][--fs_backend]
-[--allow_bucket_creation][--force_md5_etag][--anonymous][--iam_operate_on_root_account]
+[--allow_bucket_creation][--force_md5_etag][--anonymous][--iam_operate_on_root_account][--default_connection]
 ```
 #### Flags -
 - `name` (Required)
@@ -206,6 +210,10 @@ noobaa-cli account update --name <account_name> [--new_name][--uid][--gid][--use
 - `iam_operate_on_root_account`
     - Type: Boolean
     - Description: Specifies if the account allowed to create root accounts using the IAM API (the default behavior is to create of IAM accounts). See - [IAM - Root Accounts Manager](./../design/iam.md#root-accounts-manager).
+
+- `default_connection`
+    - Type: String
+    - Description: A default account for Kafka external servers. See bucket-notifications.md.
 
 ### Account Status
 
--- a/src/endpoint/s3/ops/s3_put_bucket_notification.js
+++ b/src/endpoint/s3/ops/s3_put_bucket_notification.js
@@ -10,6 +10,7 @@ const notif_util = require('../../../util/notifications_util');
 async function put_bucket_notification(req) {
 
     let topic_configuration = req.body.NotificationConfiguration?.TopicConfiguration;
+    const default_connection = req.object_sdk.requesting_account.default_connection;
 
     //adapt to db shcema
     if (topic_configuration) {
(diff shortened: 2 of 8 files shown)


guymguym (changes requested):
I think we should follow the arn structure format, see comments posted above.

src/util/notifications_util.js
```diff
@@ -186,15 +186,24 @@ class Notificator {
         }
     }
 
-    async parse_connect_file(connect_filename, decrypt = false) {
+    async parse_connect_file(connect_filename_with_overrides, decrypt = false) {
         let connect;
+        const filename_parts = connect_filename_with_overrides.split('?');
+        const connect_filename_no_overrides = filename_parts[0];
+        const overrides_str = filename_parts[1];
```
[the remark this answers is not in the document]
alphaprinz: I've made the topic for kafka to optionally be "kafka:::topic".
Still haven't found any mention of cluster UUID in the client.
Note that using '/' separator in containerized is somewhat confusing because I also need secret name in addition to connection name (currently, at least).

src/endpoint/s3/ops/s3_put_bucket_notification.js
```diff
@@ -10,13 +10,37 @@ const notif_util = require('../../../util/notifications_util');
...
 
     //adapt to db shcema
     if (topic_configuration) {
         for (const conf of topic_configuration) {
             conf.id = conf.Id;
             conf.event = conf.Event;
             conf.topic = conf.Topic;
+            //handle Kafka's topic synax, if present
+            if (conf.Topic && conf.Topic.length > 0 && conf.Topic[0].startsWith('kafka:::topic/')) {
+                //kafka_topic_parts[0] = 'kafka:::topic'
```
nadavMiz, line 23: I assume this is the structure of the kafka_topic_parts after the split. maybe its just me but it looks like code that was commented out. can you maybe add a line above that that says something like: `kafka_topic_parts object pasts:`. maybe it will be clearer that way

src/manage_nsfs/manage_nsfs_constants.js
```diff
@@ -46,8 +46,8 @@ const FROM_FILE = 'from_file';
 const ANONYMOUS = 'anonymous';
 
 const VALID_OPTIONS_ACCOUNT = {
-    'add': new Set(['name', 'uid', 'gid', 'supplemental_groups', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', FROM_FILE, ...CLI_MUTUAL_OPTIONS]),
-    'update': new Set(['name', 'uid', 'gid', 'supplemental_groups', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', 'new_name', 'regenerate', ...CLI_MUTUAL_OPTIONS]),
+    'add': new Set(['name', 'uid', 'gid', 'supplemental_groups', 'new_buckets_path', 'user', 'access_key', 'secret_key', 'fs_backend', 'allow_bucket_creation', 'force_md5_etag', 'iam_operate_on_root_account', 'default_connection', FROM_FILE, ...CLI_MUTUAL_OPTIONS]),
```
nadavMiz, line 49: why is this part of the account configuration. shouldn't it be per bucket?

docs/NooBaaNonContainerized/NooBaaCLI.md
```diff
@@ -75,7 +75,7 @@ The `account add` command is used to create a new account with customizable opti
 ```sh
 noobaa-cli account add --name <account_name> --uid <uid> --gid <gid> [--user]
 [--new_buckets_path][--access_key][--secret_key][--fs_backend]
-[--allow_bucket_creation][--force_md5_etag][--anonymous][--from_file][--iam_operate_on_root_account]
+[--allow_bucket_creation][--force_md5_etag][--anonymous][--from_file][--iam_operate_on_root_account][--default_connection]
```
nadavMiz, line 78: can you add a test for the cli path in `test_nc_account_cli.test.js`

src/manage_nsfs/manage_nsfs_constants.js
alphaprinz: Apparently in Marc's project the user and connection creation would be done by one admin user (with manage_nsfs cli), while buckets and their notif conf would be done by other (not-admin) users (with s3 rest api). So the admin would be able to do a once-per-system users and connection creation (including user's default connection), and then the other users wouldn't need to know about connections at all.

src/endpoint/s3/ops/s3_put_bucket_notification.js
```diff
@@ -10,13 +10,38 @@ const notif_util = require('../../../util/notifications_util');
...
+    const default_connection = req.object_sdk.requesting_account.default_connection;
 
     //adapt to db shcema
     if (topic_configuration) {
         for (const conf of topic_configuration) {
             conf.id = conf.Id;
             conf.event = conf.Event;
             conf.topic = conf.Topic;
+            //handle Kafka's topic synax, if present
+            if (conf.Topic && conf.Topic.length > 0 && conf.Topic[0].startsWith('kafka:::topic/')) {
```
nadavMiz, line 22: wouldn't it be also good to have a validation on the cli level. that you wouldn't be able to configure default_connection of this format?

alphaprinz: I'm not sure I understand.
Account's default_connection is just a string. It doesn't have a "special" syntax (as opposed to conf.Topic's optional "kafka:::topic" syntax).

nadavMiz: ok, if it doesn't have a  "special" syntax, its fine

nadavMiz (approved):
LGTM

nimrod-becker:
Guy's comments were resolved, we can merge

PR #8914, opened 2025-03-27
Added required validations in lifecycle rules

aayushchouhan09:
### Describe the Problem
The following validations are missing from PutBucketLifecycleConfiguration:

- Expiration Date Must Be Midnight UTC
    - `Date` must be exactly 00:00:00 UTC, or AWS rejects it.
- AbortIncompleteMultipartUpload Cannot Be Used with Size/Tags
    - AWS does not allow it with `Tags` or `ObjectSizeGreaterThan/LessThan`.
- Expiration Rule Conflict
     - `Days/Date` cannot be used with `ExpiredObjectDeleteMarker`.
- Multiple Filter Actions Require And
     - AWS requires multiple `Filters` to be inside an `And` block.

### Explain the Changes
1. Added validations for expiration date, AbortIncompleteMultipartUpload, expiration rule conflicts, and multiple filters requiring And.

### Issues: Fixed #xxx / Gap #xxx
1. Fix: #8870 

### Testing Instructions:
1.  run lifecycle tests: `make run-single-test-postgres testname=lifecycle_index.js`
2.  removed the faulty test from ceph s3 tests and created [ceph s3 tests bug](https://github.com/ceph/s3-tests/issues/638)

- [X] Tests added

PR #8940, opened 2025-04-07
Added a fix for bucket lifecycle having tagging as an empty array in filter

aayushchouhan09:
### Describe the Problem
While debugging a failing test, it was found that bucket lifecycle incorrectly include `tagging: []` (as an empty array) when tagging is not provided.

Below lifecycle rule doesn't work correctly, here tagging is an empty array:

`{ bucket_id: 67efcc4869121b0008c9ea14, key: /^/, max_create_time: 1743724800, tagging: [], latest_versions: true, filter_delete_markers: true, max_size: 3, min_size: 1, limit: 1000 }`

### Explain the Changes
1. Added a fix to handle the edge case where tagging is defined as an empty array in the lifecycle filter.
2. The fix ensures that if tagging is not provided or is an empty array, it is set to undefined, so it is excluded from the final query.

1. Gap: #8914 

### Testing Instructions:
1. Tests covered in the above PR where the gap is found.

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -669,7 +669,7 @@ class MDStore {
                 $lt: new Date(moment.unix(max_create_time).toISOString()),
                 $exists: true
             } : undefined,
-            tagging: tagging ? {
+            tagging: (tagging?.length > 0) ? {
                 $all: tagging,
             } : undefined,
             size: (max_size || min_size) ?


src/server/object_services/md_store.js
```diff
@@ -669,7 +669,7 @@ class MDStore {
                 $lt: new Date(moment.unix(max_create_time).toISOString()),
                 $exists: true
             } : undefined,
-            tagging: tagging ? {
+            tagging: (tagging && tagging.length) ? {
```
liranmauda, line 672: maybe be more explicit:
```suggestion
            tagging: (tagging && tagging.length > 0) ? {
```

achouhan09: sure, thanks

src/server/object_services/md_store.js
```diff
@@ -669,7 +669,7 @@ class MDStore {
                 $lt: new Date(moment.unix(max_create_time).toISOString()),
                 $exists: true
             } : undefined,
-            tagging: tagging ? {
+            tagging: (tagging && tagging.length > 0) ? {
```
tangledbytes, line 672:
```suggestion
            tagging: (tagging?.length > 0) ? {
```

tangledbytes (approved):

src/server/object_services/md_store.js
achouhan09: Thanks

PR #8949, opened 2025-04-08
NC | lifecycle | add newer noncurrent versions rule

nadavMiz:
### Describe the Problem
add lifecycle rule to delete noncurrent objects based on number of newer version. save progress in state struck to enable work in batches. change the state structure to enable working on both expire and noncurrent actions in the same rule

### Explain the Changes
1. implement function to filter based on newer noncurrent versions
2. separate the state object to non-current and expire state to keep track of keyh_markers for both list_object and list-object-versions
3. don't run action if is_finished is true. when both expire and noncurrent state is_finished flag is set to true, the rule is set to be finished
4. move handling of the states and getting next batch to new functions: `load_versions_list` and `load_object_list`
5. added tests
6. since while using timeout we might not delete the batch objects before moving to the next batch causing the test to fail occasionally. reduce the timeout so no action will happen

### Gaps
1. move the accumulation of num_processed_objects to update_status()

### Testing Instructions:

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/manage_nsfs/nc_lifecycle.js
+++ b/src/manage_nsfs/nc_lifecycle.js
@@ -145,7 +145,7 @@ class NCLifecycle {
     }
 
     /**
-     * process_buckets does the following - 
+     * process_buckets does the following -
      * 1. if it's a GPFS optimization - create candidates files
      * 2. iterates over buckets and handles their rules
      * @param {String[]} bucket_names
@@ -364,6 +364,7 @@ class NCLifecycle {
      */
     async get_candidates(bucket_json, lifecycle_rule, object_sdk) {
         const candidates = { abort_mpu_candidates: [], delete_candidates: [] };
+        const versions_list = undefined;
         if (lifecycle_rule.expiration) {
             candidates.delete_candidates = await this.get_candidates_by_expiration_rule(lifecycle_rule, bucket_json,
                 object_sdk);
@@ -373,7 +374,12 @@ class NCLifecycle {
             }
         }
         if (lifecycle_rule.noncurrent_version_expiration) {
-            const non_current_candidates = await this.get_candidates_by_noncurrent_version_expiration_rule(lifecycle_rule, bucket_json);
+            const non_current_candidates = await this.get_candidates_by_noncurrent_version_expiration_rule(
+                lifecycle_rule,
+                bucket_json,
+                object_sdk,
+                {versions_list}
+            );
             candidates.delete_candidates = candidates.delete_candidates.concat(non_current_candidates);
         }
         if (lifecycle_rule.abort_incomplete_multipart_upload) {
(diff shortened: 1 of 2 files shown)


src/manage_nsfs/nc_lifecycle.js
```diff
@@ -553,11 +604,31 @@ class NCLifecycle {
...
+        }
+
+        const filter_func = this._build_lifecycle_filter({filter: lifecycle_rule.filter, expiration: 0});
+        const num_newer_versions = lifecycle_rule.noncurrent_version_expiration.newer_noncurrent_versions;
+        const num_non_current_days = lifecycle_rule.noncurrent_version_expiration.noncurrent_days;
+        const delete_candidates = [];
+
+        for (const entry of versions_list.objects) {
+            const lifecycle_info = this._get_lifecycle_object_info_from_list_object_entry(entry);
+            if ((num_newer_versions !== undefined && this.filter_newer_versions(entry, rule_state, num_newer_versions)) ||
```
romayalon, line 622: shouldn't it be && ?

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -498,6 +509,63 @@ class NCLifecycle {
...
+            version_id_marker: rule_state.version_id_marker
+        });
+        if (list_versions.is_truncated) {
+            rule_state.is_finished = false;
+            rule_state.key_marker_versioned = list_versions.next_marker;
+            rule_state.version_id_marker = list_versions.next_version_id_marker;
+        } else {
+            rule_state.is_finished = true;
+        }
+        const bucket_state = this.lifecycle_run_status.buckets_statuses[bucket_json.name].state;
```
romayalon, line 534: not necessarily on this PR if you don't want, I think we can move the accumulation of num_processed_objects to update_status(), can you add it as a gap?

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -428,29 +434,42 @@ class NCLifecycle {
         }
     }
 
+    async load_objects_list(object_sdk, lifecycle_rule, bucket_json, rule_state) {
```
romayalon, line 437: jsdoc

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -428,29 +434,42 @@ class NCLifecycle {
...
     /**
      *
      * @param {*} lifecycle_rule
      * @param {Object} bucket_json
-     * @param {nb.ObjectSDK} object_sdk 
+     * @param {nb.ObjectSDK} object_sdk
      * @returns {Promise<Object[]>}
      */
     async get_candidates_by_expiration_rule_posix(lifecycle_rule, bucket_json, object_sdk) {
+        const rule_state = this.lifecycle_run_status.buckets_statuses[bucket_json.name].rules_statuses[lifecycle_rule.id].state.expire;
```
romayalon, line 463: seems like my GPFS code is not aligned with the state.expire change, 
let's talk on our meeting today what we need to do there

romayalon (approved):

PR #8950, opened 2025-04-08
NC | Delete object filter verification on regular delete object 

romayalon:
### Describe the Problem
On NC lifecycle - getting the deletion candidates might take a lot of time (especially when using GPFS optimization), therefore we should re-check the filter and expiration before deleting the object.

### Explain the Changes
1. NCLifecycle - Since we call delete_multiple_objects per rule - passed the filter func to the delete_multiple_objects function.
2. NamespaceFS - Changed delete_multiple_objects to delete based on file descriptors if the flow is gpfs/lifecycle_deletion + apply the filter func and check if we should still delete the object. if not - throw an error.
3. refactoring - moved some functions to utils so they can be used by both NCLifecycle and NamespaceFS.

1.  

### Testing Instructions:
1. `sudo jest --testRegex=jest_tests/test_nc_lifecycle_posix`
2.  `sudo jest --testRegex=jest_tests/test_nc_lifecycle`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -24,6 +24,7 @@ const LRUCache = require('../util/lru_cache');
 const nb_native = require('../util/nb_native');
 const RpcError = require('../rpc/rpc_error');
 const { S3Error } = require('../endpoint/s3/s3_errors');
+const lifecycle_utils = require('../util/lifecycle_utils');
 const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 const { PersistentLogger } = require('../util/persistent_logger');
 const { Glacier } = require('./glacier');
@@ -298,6 +299,11 @@ function filter_fs_xattr(xattr) {
     return _.pickBy(xattr, (val, key) => key?.startsWith(XATTR_NOOBAA_INTERNAL_PREFIX));
 }
 
+/**
+ * get_tags_from_xattr converts relevant xattr to tags format
+ * @param {Object} xattr 
+ * @returns {Object}
+ */
 function get_tags_from_xattr(xattr) {
     const tag_set = [];
     for (const [xattr_key, xattr_value] of Object.entries(xattr)) {
@@ -736,6 +742,8 @@ class NamespaceFS {
                         const use_lstat = !(await this._is_path_in_bucket_boundaries(fs_context, entry_path));
                         const stat = await native_fs_utils.stat_if_exists(fs_context, entry_path,
                             use_lstat, config.NSFS_LIST_IGNORE_ENTRY_ON_EACCES);
+                        // TODO - GAP of .folder files - we return stat of the directory for the 
+                        // xattr, but the creation time should be of the .folder files (and maybe more )
                         if (stat) {
                             r.stat = stat;
                             // add the result only if we have the stat information
@@ -1523,8 +1531,10 @@ class NamespaceFS {
                 await native_fs_utils._make_path_dirs(latest_ver_path, fs_context);
                 if (is_gpfs) {
                     const latest_ver_info_exist = await native_fs_utils.is_path_exists(fs_context, latest_ver_path);
-                    gpfs_options = await this._open_files_gpfs(fs_context, new_ver_tmp_path, latest_ver_path, upload_file,
-                        latest_ver_info_exist, open_mode, undefined, undefined);
+                    gpfs_options = await this._open_files(fs_context, {
+                        src_path: new_ver_tmp_path, dst_path: latest_ver_path, upload_or_dir_file: upload_file,
+                        dst_ver_exist: latest_ver_info_exist, open_mode
+                    });
 
                     //get latest version if exists
                     const latest_fd = gpfs_options?.move_to_dst?.dst_file;
@@ -1580,7 +1590,7 @@ class NamespaceFS {
                 if (!should_retry || retries <= 0) throw err;
                 await P.delay(get_random_delay(config.NSFS_RANDOM_DELAY_BASE, 0, 50));
             } finally {
-                if (gpfs_options) await this._close_files_gpfs(fs_context, gpfs_options.move_to_dst, open_mode);
+                if (gpfs_options) await this._close_files(fs_context, gpfs_options.move_to_dst, open_mode);
             }
         }
     }
@@ -2048,7 +2058,7 @@ class NamespaceFS {
                         const file_path = this._get_file_path({ key });
                         await this._check_path_in_bucket_boundaries(fs_context, file_path);
                         dbg.log1('NamespaceFS: delete_multiple_objects', file_path);
-                        await this._delete_single_object(fs_context, file_path, { key });
+                        await this._delete_single_object(fs_context, file_path, { key, filter_func: params.filter_func });
                         res.push({ key });
                     } catch (err) {
                         res.push({ err_code: err.code, err_message: err.message });
@@ -2063,7 +2073,7 @@ class NamespaceFS {
                 }
                 dbg.log3('NamespaceFS: versions_by_key_map', versions_by_key_map);
                 for (const key of Object.keys(versions_by_key_map)) {
-                    const key_res = await this._delete_objects_versioned(fs_context, key, versions_by_key_map[key]);
+                    const key_res = await this._delete_objects_versioned(fs_context, key, versions_by_key_map[key], params.filter_func);
                     res = res.concat(key_res);
                 }
             }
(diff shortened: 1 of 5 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -1736,21 +1712,50 @@ struct UnlinkFileAt : public FSWrapWorker<FileWrap>
 {
     std::string _filepath;
     int _delete_fd;
+    int64_t _verify_inode;
+    int _flags;
     UnlinkFileAt(const Napi::CallbackInfo& info)
```
nadavMiz, line 1717: the new unlink function structure is very confusing.
from what I understand we use to have 4 unlink functions functions:
SafeUnlink - posix safe unlink function
UnlinkFileAt - GPFS safe unlink function
unlink - unsafe unlink fucntion - simple POSIX unlink
unlinkat - unsafe unlink fucntion - simple POSIX unlinkat

now we have:
UnlinkFileAt - which acts as both safe unlink for posix and GPFS. but also as an unsafe function just for POSIX
while we still have SafeUnlink which is a safe function just for POSIX versioning
and unlink which is a general unsafe function for both POSIX and GPFS, and the only GPFS unsafe option

nadavMiz: in my opinion we should ideally combine the safe options together and not allow unsafe usage. but we should at least add comment for each function on when it should be used

src/native/fs/fs_napi.cpp
```diff
@@ -1736,21 +1712,50 @@ struct UnlinkFileAt : public FSWrapWorker<FileWrap>
...
+        } else {
+            DBG1("Unlinkat " << DVAL(fd) << DVAL(path.c_str()) << DVAL(_filepath.c_str()) << DVAL(_flags) << DVAL(_verify_inode));
+            if (_verify_inode > 0) {
+                struct stat _stat_res;
+                std::string full_path = path + "/" + _filepath;
+                DBG1("Unlinkat " << DVAL(full_path));
+                SYSCALL_OR_RETURN(stat(full_path.c_str(), &_stat_res));
+                int64_t stat_actual_ino = _stat_res.st_ino;
+                DBG1("Unlinkat " << DVAL(stat_actual_ino) << DVAL(_verify_inode));
+                if (stat_actual_ino != _verify_inode) {
```
nadavMiz, line 1751: we should probably add comment about this being best effort. file could still change between stat and unlink

src/sdk/namespace_fs.js
```diff
@@ -2069,9 +2077,24 @@ class NamespaceFS {
...
+    /**
+     * _delete_single_object does the following before deleting the object
+     * 1. open dir_file and src_file fd
+     * 2. _unlink_file_ignore_enoent - which means it stats the to be deleted file, validate filter if exists, unlink safely
+     * 3. deleted parent directories if they are empty 
+     * 4. clears directory object xattr if relevant
+     * 5. closes file and dir_file
+     */
     async _delete_single_object(fs_context, file_path, params) {
-        await native_fs_utils.unlink_ignore_enoent(fs_context, file_path);
```
nadavMiz, line 2074: I think we should have a separation between the lifecycle path that requires safe_unlink and general path that doesn't. the extra file openings and validations are not required for cases other than lifecycle

src/sdk/namespace_fs.js
```diff
@@ -3223,33 +3262,30 @@ class NamespaceFS {
...
                 dbg.log1('Namespace_fs._delete_latest_version:', latest_ver_info);
                 if (latest_ver_info) {
-                    if (is_gpfs) {
-                    gpfs_options = await this._open_files_gpfs(fs_context, latest_ver_path, undefined, undefined, undefined,
-                            undefined, true);
-                        const latest_fd = gpfs_options?.delete_version?.src_file;
-                        latest_ver_info = latest_fd && await this._get_version_info(fs_context, undefined, latest_fd);
-                        if (!latest_ver_info) break;
-                    }
+                    files = await this._open_files(fs_context, { src_path: latest_ver_path, delete_version: true });
```
nadavMiz, line 3275: again, should probably not open on non GPFS non lifecycle paths

src/sdk/namespace_fs.js
```diff
@@ -2912,28 +2935,35 @@ class NamespaceFS {
...
 
             const version_id_str = this._get_version_id_by_xattr(stat);
             const ver_info_by_xattr = this._extract_version_info_from_xattr(version_id_str);
             return {
                 ...(ver_info_by_xattr || stat),
                 version_id_str,
                 delete_marker: stat.xattr[XATTR_DELETE_MARKER],
-                path: version_path
+                path: version_path,
+                stat
```
nadavMiz, line 2966: its just my opinion. but adding the stat here doesn't seem like the proper way of passing the stat. this function is used to get version specific information, some of it from the stat itself. this function is used in many places. adding the stat here for a specific purpose seems kind of forced, especially if we have both the stat and the stats content in the same object

src/sdk/namespace_fs.js
```diff
@@ -3476,6 +3513,32 @@ class NamespaceFS {
...
+     * _unlink_file_ignore_enoent does the following - 
+     * 1. stat the to be deleted file
+     * 2. checks that the file should be deleted based on lifecycle filter (if the flow is not lifecycle the check will be skipped)
+     * 3. calls unlinkfileat that inside checks that the path to be deleted has the same inode/fd of the file that should be deleted
+     * @param {nb.NativeFSContext} fs_context 
+     * @param {Object} params 
+     * @param {String} file_path 
+     * @param {{dir_file: nb.NativeFile, src_file: nb.NativeFile }} files 
+     */
+    async _unlink_file_ignore_enoent(fs_context, params, file_path, { dir_file, src_file }) {
```
nadavMiz, line 3526: this function has almost the same name as native_fs_utils.unlink_ignore_enoent. should use a different name to prevent confusion

src/native/fs/fs_napi.cpp
```diff
@@ -1736,21 +1712,50 @@ struct UnlinkFileAt : public FSWrapWorker<FileWrap>
...
+        }
+        if (info.Length() > 4 && !info[4].IsUndefined()) {
+            _flags = info[4].As<Napi::Number>();
+        }
+        Begin(XSTR() << "UnlinkFileAt" << DVAL(_wrap->_path.c_str()) << DVAL(_wrap->_fd) << DVAL(_filepath) << DVAL(_delete_fd) << DVAL(_verify_inode));
     }
     virtual void Work()
     {
         int fd = _wrap->_fd;
+        std::string path = _wrap->_path;
```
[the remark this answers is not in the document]
romayalon: _path is coming from the file warp struct (line 1423), which is an object we use for doing operations on opened files.
unlinkfileat worker is not a regular worker, it's a fileWrap worker, therefore unlinkfileat is being called from the filewrap object (first we open the file, get a file wrap and later on able to do fs_napi operations on the file wrap).
[**UnlinkAt**](https://linux.die.net/man/2/unlinkat) as you probably know if being executed on the dirfd and not on the file itself 
```
int unlinkat(int dirfd, const char *pathname, int flags);
```
I understand your confusion with the naming, will change it to dir path for extra clarity

src/native/fs/fs_napi.cpp
romayalon: After my change we have - 
1. unlink() - Unsafe regular worker - I don't agree we don't need it, sometimes like in tmp file we don't care about safety because the file is tmp/have a unique id. Also it's used by many of non NC nsfs components which I'm not going to touch.
2.  SafeUnlink() - Safe regular worker -  used for posix, I agree that should be merged with unlinkfileat() but it's a part of the safe_move() and all the versioning move version mechanism which requires the fs opening changes we discussed about in the past, this PR opens fd for the delete operations but has nothing to do with PUT + versioning, therefore it's not the PR to do the merge between those ops.
3. UnlinkfileAt - fd safe file wrap worker - requires opening of files before the operation can be used. now contains GPFS and POSIX implementation.

src/sdk/namespace_fs.js
romayalon: I did this because I saw multiple stat calls for MISMATCH VERSION after calls to this function 
these stat calls are redundant file system calls so I wanted to refactor it, 
I agree that we should do it on a different PR though
also the same for checking MISMATCH on posix scenario will use this

src/sdk/namespace_fs.js
romayalon: I agree, I did a lot of refactoring in this PR and I think we don't want to take a risk at this point of time
moving it to a different PR

src/native/fs/fs_napi.cpp
romayalon: eventually I removed all changes from this PR and used SafeUnlink()
when we will move to use open files on posix we will need those changes

src/sdk/namespace_fs.js
romayalon: so I removed it for now and called stat again when needed as we do in MISMATCH VERSION check
we should still fix it on a different PR though

nadavMiz (approved):
LGTM

PR #8953, opened 2025-04-09
File Reader | Add next_line_file_offset

romayalon:
### Describe the Problem
On https://github.com/noobaa/noobaa-core/pull/8923, we would like to preserve the next_line_file_offset in order to start reading in the next iteration from the middle of the file and not read the whole file again.
Moved the fileReader changes from 8923 to a new PR.

### Explain the Changes
1. Added next_line_file_offset to file Reader.
2. Added unit tests.

1. 

### Testing Instructions:
1. sudo jest --testRegex=jest_tests/test_newline_reader --t "next_line_file_offset"

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/util/file_reader.js
+++ b/src/util/file_reader.js
@@ -30,6 +30,7 @@ class NewlineReader {
      *  bufsize?: number;
      *  skip_leftover_line?: boolean;
      *  skip_overflow_lines?: boolean;
+     *  read_file_offset?: number;
      * }} [cfg]
      **/
     constructor(fs_context, filepath, cfg) {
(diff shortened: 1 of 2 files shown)


src/util/file_reader.js
```diff
@@ -118,13 +120,15 @@ class NewlineReader {
                         console.warn('line too long finally terminated at eof:', this.info());
                     } else {
                         const line = this.buf.toString('utf8', this.start, this.end);
+                        this.start = this.end;
```
tangledbytes, line 123: I know the intention here to set `this.start` was to make sure that we can keep reading while the file was being written but I think to enable that we would need more than just this as you would notice that we enter the read loops only if `this.eof === false`. Not something to be addressed in this PR probably, or maybe just a comment on top of the loop stating that after each read the reader must set `reader.eof = false` if they intend to keep on reading from a file while it is being written.

tangledbytes (approved):
LGTM

PR #8961, opened 2025-04-09
NC | lifecycle | add expire delete marker rule

nadavMiz:
### Describe the Problem
added expire delete marker rule. see https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configuration-examples.html#lifecycle-config-conceptual-ex7

### Explain the Changes
1. add expire delete marker rule
2. in case last item is a delete marker, change the state to start from this delete marker in the next run

1. 

### Testing Instructions:
1. `sudo npx jest test_nc_lifecycle_cli`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/manage_nsfs/nc_lifecycle.js
+++ b/src/manage_nsfs/nc_lifecycle.js
@@ -369,7 +369,12 @@ class NCLifecycle {
             candidates.delete_candidates = await this.get_candidates_by_expiration_rule(lifecycle_rule, bucket_json,
                 object_sdk);
             if (lifecycle_rule.expiration.days || lifecycle_rule.expiration.expired_object_delete_marker) {
-                const dm_candidates = await this.get_candidates_by_expiration_delete_marker_rule(lifecycle_rule, bucket_json);
+                const dm_candidates = await this.get_candidates_by_expiration_delete_marker_rule(
+                    lifecycle_rule,
+                    bucket_json,
+                    object_sdk,
+                    {versions_list}
+                );
                 candidates.delete_candidates = candidates.delete_candidates.concat(dm_candidates);
             }
         }
@@ -439,21 +444,21 @@ class NCLifecycle {
      * @param {Object} object_sdk
      * @param {Object} lifecycle_rule
      * @param {Object} bucket_json
-     * @param {Object} rule_state
+     * @param {Object} expire_state
      * @returns
      */
-    async load_objects_list(object_sdk, lifecycle_rule, bucket_json, rule_state) {
+    async load_objects_list(object_sdk, lifecycle_rule, bucket_json, expire_state) {
         const objects_list = await object_sdk.list_objects({
             bucket: bucket_json.name,
             prefix: lifecycle_rule.filter?.prefix,
-            key_marker: rule_state.key_marker,
+            key_marker: expire_state.key_marker,
             limit: config.NC_LIFECYCLE_LIST_BATCH_SIZE
         });
         if (objects_list.is_truncated) {
-            rule_state.key_marker = objects_list.next_marker;
-            rule_state.is_finished = false;
+            expire_state.key_marker = objects_list.next_marker;
+            expire_state.is_finished = false;
         } else {
-            rule_state.is_finished = true;
+            expire_state.is_finished = true;
         }
         const bucket_state = this.lifecycle_run_status.buckets_statuses[bucket_json.name].state;
         bucket_state.num_processed_objects += objects_list.objects.length;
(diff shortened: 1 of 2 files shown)


src/manage_nsfs/nc_lifecycle.js
```diff
@@ -518,32 +544,89 @@ class NCLifecycle {
...
-        bucket_state.num_processed_objects += objects_list.objects.length;
-        if (objects_list.is_truncated) {
-            rule_state.key_marker = objects_list.next_marker;
-        } else {
-            rule_state.is_finished = true;
-        }
         return filtered_objects;
     }
 
+    filter_expired_delete_marker(object, next_object, filter_func) {
```
romayalon, line 550: jsdoc

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -518,32 +544,89 @@ class NCLifecycle {
...
+        const lifecycle_info = this._get_lifecycle_object_info_from_list_object_entry(object);
+        if (!filter_func(lifecycle_info)) return false;
+        return object.is_latest && object.delete_marker && object.key !== next_object.key;
+    }
+
     /**
      * get_candidates_by_expiration_delete_marker_rule processes the expiration delete marker rule
      * @param {*} lifecycle_rule
      * @param {Object} bucket_json
```
romayalon, line 559: update

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -518,32 +544,89 @@ class NCLifecycle {
...
-        // TODO - implement
-        return [];
+    async get_candidates_by_expiration_delete_marker_rule(lifecycle_rule, bucket_json, object_sdk, {versions_list}) {
+        const rule_state = this.lifecycle_run_status.buckets_statuses[bucket_json.name].rules_statuses[lifecycle_rule.id].state.noncurrent;
+        if (rule_state.is_finished) return [];
+        if (!versions_list) {
+            versions_list = await this.load_versions_list(object_sdk, bucket_json, rule_state);
+        }
+        const candidates = [];
+        const filter_func = this._build_lifecycle_filter({filter: lifecycle_rule.filter, expiration: 0});
```
romayalon, line 569: when having an expired delete marker we don't care if there is also expiration by days/date?

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -518,32 +544,89 @@ class NCLifecycle {
...
+            }
+        }
+        return candidates;
     }
 
 /////////////////////////////////////////////
 //////// NON CURRENT VERSION HELPERS ////////
 /////////////////////////////////////////////
 
+    async load_versions_list(object_sdk, bucket_json, rule_state) {
```
romayalon, line 592: add jsdoc in all new functions please

src/manage_nsfs/nc_lifecycle.js
nadavMiz: its a separate rule than both of them

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -518,32 +544,89 @@ class NCLifecycle {
...
+                candidates.push(versions_list.objects[i]);
+            }
+        }
+        const last_item = versions_list.objects.length > 0 && versions_list.objects[versions_list.objects.length - 1];
+        const lifecycle_info = this._get_lifecycle_object_info_from_list_object_entry(last_item);
+        if (last_item.is_latest && last_item.delete_marker && filter_func(lifecycle_info)) {
+            if (rule_state.is_finished) {
+                candidates.push(last_item);
+            } else {
+                //need the next item to decide if we need to delete. start the next cycle from this key latest
```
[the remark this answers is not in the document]
nadavMiz: version_id marker is undefined, key_marker is still the same. that way we will start at the first key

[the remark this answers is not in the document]
romayalon: got it, can you add it in a comment?

src/manage_nsfs/nc_lifecycle.js
nadavMiz: according to aws:
```
You can't specify both a Days and an ExpiredObjectDeleteMarker tag on the same rule. When you specify the Days tag, Amazon S3 automatically performs ExpiredObjectDeleteMarker cleanup when the delete markers are old enough to satisfy the age criteria. To clean up delete markers as soon as they become the only version, create a separate rule with only the ExpiredObjectDeleteMarker tag.
```
so I guess you are correct. I'll change this

src/manage_nsfs/nc_lifecycle.js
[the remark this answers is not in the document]
nadavMiz: right there was a bug, in key_marker is set to be the next key. in case the next key is different it wouldn't work. I corrected this

src/manage_nsfs/nc_lifecycle.js
nadavMiz: trying to put date and expire delete marker together returns MalformedXML error. there is nothing on AWS documentation about about implicit ExpiredObjectDeleteMarker when having date rule. so I guess we can't have both of them together

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -490,26 +501,50 @@ class NCLifecycle {
...
+     */
+    async load_objects_list(object_sdk, lifecycle_rule, bucket_json, rule_state) {
+        const objects_list = await object_sdk.list_objects({
+            bucket: bucket_json.name,
+            prefix: lifecycle_rule.filter?.prefix,
+            key_marker: rule_state.key_marker,
+            limit: config.NC_LIFECYCLE_LIST_BATCH_SIZE
+        });
+        if (objects_list.is_truncated) {
+            rule_state.key_marker = objects_list.next_marker;
```
romayalon, line 520: shouldn't this be under rule_state.expire.key_marker? where do we set noncurrent/expire on the rule state?

nadavMiz: those are noncurrent/expire. the I changed the variable names

romayalon (approved):

PR #8967, opened 2025-04-10
[NC | DBS3] Add support for reserved bucket tags

tangledbytes:
### Explain the Changes
This PR adds the concept of "reserved" bucket tags which are bucket tags which can be set via noobaa cli or via `put-bucket-tagging` but they can come with further restrictions like a schema, immutability, etc.

DBS3 Example:
```
// In config-local.js or in config.json
config.NSFS_GLACIER_RESERVED_BUCKET_TAGS = {
    'deep-archive-copies': {
        schema: {
            $id: 'deep-archive-copies-schema-v0',
            enum: ['1', '2']
        }, // JSON Schema
        immutable: 'if-data',
        default: '1',
        event: true,
    }
}
```

Usage Example:
```
$ noobaa-cli bucket update --tag '[{ "key": "k1", "value": "v1" }]' ...
$ noobaa-cli bucket update --merge_tag '[{ "key": "k1", "value": "v1" }]' ...
$ aws s3api put-bucket-tagging TagSet=[{Key=deep-archive-copies,Value=2}]'
```

1. 

### Testing Instructions:
1. 

- [x] Doc added/updated
- [ ] Tests added

### DIFF
--- a/config.js
+++ b/config.js
@@ -954,6 +954,33 @@ config.NSFS_GLACIER_FORCE_EXPIRE_ON_GET = false;
 // interval
 config.NSFS_GLACIER_MIGRATE_LOG_THRESHOLD = 50 * 1024;
 
+/** 
+ * NSFS_GLACIER_RESERVED_BUCKET_TAGS defines an object of bucket tags which will be reserved
+ * by the system and PUT operations for them via S3 API would be limited - as in they would be
+ * mutable only if specified and only under certain conditions.
+ *
+ * @type {Record<string, {
+ *  schema: Record<any, any> & { $id: string },
+ *  immutable: true | false | 'if-data',
+ *  default: any,
+ *  event: boolean
+ * }>}
+ * 
+ * @example
+ * {
+    'deep-archive-copies': {
+        schema: {
+            $id: 'deep-archive-copies-schema-v0',
+            enum: ['1', '2']
+        }, // JSON Schema
+        immutable: 'if-data',
+        default: '1',
+        event: true
+    }
+ * }
+ */
+config.NSFS_GLACIER_RESERVED_BUCKET_TAGS = {};
+
 // anonymous account name
 config.ANONYMOUS_ACCOUNT_NAME = 'anonymous';
 
--- a/src/cmd/manage_nsfs.js
+++ b/src/cmd/manage_nsfs.js
@@ -39,6 +39,8 @@ const { throw_cli_error, get_bucket_owner_account_by_name,
 const manage_nsfs_validations = require('../manage_nsfs/manage_nsfs_validations');
 const nc_mkm = require('../manage_nsfs/nc_master_key_manager').get_instance();
 const notifications_util = require('../util/notifications_util');
+const BucketSpaceFS = require('../sdk/bucketspace_fs');
+const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 
 ///////////////
 //// GENERAL //
@@ -135,7 +137,6 @@ async function fetch_bucket_data(action, user_input) {
         force_md5_etag: user_input.force_md5_etag === undefined || user_input.force_md5_etag === '' ? user_input.force_md5_etag : get_boolean_or_string_value(user_input.force_md5_etag),
         notifications: user_input.notifications
     };
-
     if (user_input.bucket_policy !== undefined) {
         if (typeof user_input.bucket_policy === 'string') {
             // bucket_policy deletion specified with empty string ''
@@ -152,6 +153,27 @@ async function fetch_bucket_data(action, user_input) {
         // @ts-ignore
         data = _.omitBy(data, _.isUndefined);
         data = await merge_new_and_existing_config_data(data);
+    }
+
+    if ((action === ACTIONS.UPDATE && user_input.tag) || (action === ACTIONS.ADD)) {
+        const tags = JSON.parse(user_input.tag || '[]');
+        data.tag = BucketSpaceFS._merge_reserved_tags(
+            data.tag || BucketSpaceFS._default_bucket_tags(),
+            tags,
+            action === ACTIONS.ADD ? true : await _is_bucket_empty(data),
+        );
+    }
+
+    if ((action === ACTIONS.UPDATE && user_input.merge_tag) || (action === ACTIONS.ADD)) {
+        const merge_tags = JSON.parse(user_input.merge_tag || '[]');
+        data.tag = _.merge(
+            data.tag,
+            BucketSpaceFS._merge_reserved_tags(
+                data.tag || BucketSpaceFS._default_bucket_tags(),
+                merge_tags,
+                action === ACTIONS.ADD ? true : await _is_bucket_empty(data),
+            )
+        );
     }
 
     //if we're updating the owner, needs to override owner in file with the owner from user input.
@@ -200,7 +222,14 @@ async function add_bucket(data) {
     data._id = mongo_utils.mongoObjectId();
     const parsed_bucket_data = await config_fs.create_bucket_config_file(data);
     await set_bucker_owner(parsed_bucket_data);
-    return { code: ManageCLIResponse.BucketCreated, detail: parsed_bucket_data, event_arg: { bucket: data.name }};
+
+    const [reserved_tag_event_args] = BucketSpaceFS._generate_reserved_tag_event_args({}, data.tag);
+
+    return {
+        code: ManageCLIResponse.BucketCreated,
+        detail: parsed_bucket_data,
+        event_arg: { ...(reserved_tag_event_args || {}), bucket: data.name, account: parsed_bucket_data.bucket_owner },
+    };
 }
 
 /**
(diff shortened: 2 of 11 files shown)


src/sdk/bucketspace_fs.js
```diff
@@ -933,6 +987,25 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
 
         return storage_classes;
     }
+
+    static async #_is_bucket_empty(name, params, ns, object_sdk) {
```
shirady, line 991: Why does the function start with the sign `#`?
Please add JSDoc to the new function

src/sdk/bucketspace_fs.js
```diff
@@ -485,12 +487,21 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+     * @param {nb.ObjectSDK} object_sdk 
+     */
+    async put_bucket_tagging(params, object_sdk) {
         try {
             const { name, tagging } = params;
             dbg.log0('BucketSpaceFS.put_bucket_tagging: Bucket name, tagging', name, tagging);
             const bucket = await this.config_fs.get_bucket_by_name(name);
-            bucket.tag = tagging;
+            const { ns } = await object_sdk.read_bucket_full_info(name);
+            const is_bucket_empty = await BucketSpaceFS.#_is_bucket_empty(name, null, ns, object_sdk);
```
shirady, line 500: Why do you pass `null` as the params?

src/sdk/bucketspace_fs.js
shirady, line 1: Could you add JSDoc to the new functions?

src/sdk/bucketspace_fs.js
```diff
@@ -933,6 +987,25 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
 
         return storage_classes;
     }
+
+    static async #_is_bucket_empty(name, params, ns, object_sdk) {
+        params = params || {};
+
+        let list;
+        try {
+            if (ns._is_versioning_disabled()) {
```
shirady, line 996: `versioning` is a property of the bucket, I think you can check if the bucket versioning is disabled in a simpler way.
https://github.com/noobaa/noobaa-core/blob/779c0bf9d3c8e7a30c6d192a59924238a24bde67/src/server/system_services/schemas/nsfs_bucket_schema.js#L48-L49

src/sdk/bucketspace_fs.js
```diff
@@ -933,6 +987,25 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+                list = await ns.list_objects({ ...params, bucket: name, limit: 1 }, object_sdk);
+            } else {
+                list = await ns.list_object_versions({ ...params, bucket: name, limit: 1 }, object_sdk);
```
shirady, line 999: Is this code used more than once?
We can move it to a function and reuse it.

config.js
```diff
@@ -956,6 +956,31 @@ config.NSFS_GLACIER_FORCE_EXPIRE_ON_GET = false;
 // interval
 config.NSFS_GLACIER_MIGRATE_LOG_THRESHOLD = 50 * 1024;
 
+/** 
+ * NSFS_GLACIER_RESERVED_BUCKET_TAGS defines an object of bucket tags which will be reserved
+ * by the system and PUT operations for them via S3 API would be disallowed even by the bucket
+ * owners
+ *
```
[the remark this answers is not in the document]
tangledbytes: I have modified the comment to make it make more sense. These tags can be changed but only under certain circumstances. Also users can never delete them.

src/cmd/manage_nsfs.js
```diff
@@ -39,6 +39,7 @@ const { throw_cli_error, get_bucket_owner_account_by_name,
 const manage_nsfs_validations = require('../manage_nsfs/manage_nsfs_validations');
 const nc_mkm = require('../manage_nsfs/nc_master_key_manager').get_instance();
 const notifications_util = require('../util/notifications_util');
+const BucketSpaceFS = require('../sdk/bucketspace_fs');
```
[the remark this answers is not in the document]
tangledbytes: I did, I don't see any circular dep here.

src/manage_nsfs/manage_nsfs_constants.js
```diff
@@ -62,7 +62,7 @@ const VALID_OPTIONS_ANONYMOUS_ACCOUNT = {
 
 const VALID_OPTIONS_BUCKET = {
     'add': new Set(['name', 'owner', 'path', 'bucket_policy', 'fs_backend', 'force_md5_etag', 'notifications', FROM_FILE, ...CLI_MUTUAL_OPTIONS]),
```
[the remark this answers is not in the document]
tangledbytes: Wanted the behaviour to be consistent with S3. Don't think setting user specified tags when the bucket is created is either a use case or desired.

src/sdk/bucketspace_fs.js
```diff
@@ -4,6 +4,7 @@
 const _ = require('lodash');
 const util = require('util');
 const path = require('path');
+const { default: Ajv } = require('ajv');
```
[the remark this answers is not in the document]
tangledbytes: Yeah but here the use case is different. The schema can be changed even while noobaa is running and the schema is not part of code rather comes from config.

src/sdk/bucketspace_fs.js
tangledbytes: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

src/sdk/bucketspace_fs.js
tangledbytes: Because I wanted to not provide and params in this case.

`undefined` vs `null` - this is quite opinionated. I prefer using `null` over `undefined` when I intentionally want to have a zero value. It helps me tell when the zero value was intentional and when the zero value is system generated.

I am aware that Guy disagrees with this opinion as `null` is an object.

src/sdk/bucketspace_fs.js
tangledbytes: I wanted to keep the check consistent with the rest of `bucketspace_fs`.

config.js
```diff
@@ -956,6 +956,33 @@ config.NSFS_GLACIER_FORCE_EXPIRE_ON_GET = false;
...
+            $id: 'deep-archive-copies-schema-v0',
+            enum: ['1', '2']
+        }, // JSON Schema
+        immutable: 'if-data',
+        default: '1',
+        event: true
+    }
+ * }
+ */
+config.NSFS_GLACIER_RESERVED_BUCKET_TAGS = {};
```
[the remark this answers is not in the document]
tangledbytes: I wanted to keep it simple but at the same time I didn't want to write hard-coded things.

This structure just made sure that I don't have to hard-code few things like -
1. Validation for the tags - Different reserved tags will have different validation requirements.
2. Default values - Different tags would have different values, some might not even have one.
3. Immutability - Different tags would want to behave differently. Some tags might want to allow mutability, some might not and some might want to do conditionally.

Do you have a suggestion for a simpler structure but make sure that we don't have some specific code (because this generic code ain't too complicated)?

src/cmd/manage_nsfs.js
```diff
@@ -729,6 +814,7 @@ async function set_bucker_owner(bucket_data) {
     } catch (err) {
         dbg.warn(`set_bucker_owner.couldn't find bucket owner data by id ${bucket_data.owner_account}`);
     }
+    console.log(account_data);
```
romayalon, line 817: delete

src/sdk/bucketspace_fs.js
```diff
@@ -485,13 +496,49 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
         try {
             const { name, tagging } = params;
             dbg.log0('BucketSpaceFS.put_bucket_tagging: Bucket name, tagging', name, tagging);
-            const bucket = await this.config_fs.get_bucket_by_name(name);
-            bucket.tag = tagging;
-            await this.config_fs.update_bucket_config_file(bucket);
+            const bucket_path = this.config_fs.get_bucket_path_by_name(name);
+
+            const bucket_lock_file = `${bucket_path}.lock`;
+            await native_fs_utils.lock_and_run(this.fs_context, bucket_lock_file, async () => {
```
romayalon, line 510: so lock_and_run() will create .lock file per bucket, do we want to use it in all our config updates (accounts/buckets etc)? 
if so, can you open a github issue for that? mentioning this requires changes in the CLI and in bucketspace_fs.

src/cmd/manage_nsfs.js
```diff
@@ -361,7 +416,37 @@ async function bucket_management(action, user_input) {
     } else if (action === ACTIONS.STATUS) {
         response = await get_bucket_status(data);
     } else if (action === ACTIONS.UPDATE) {
-        response = await update_bucket(data);
+        const bucket_path = config_fs.get_bucket_path_by_name(user_input.name);
+        const bucket_lock_file = `${bucket_path}.lock`;
+        await native_fs_utils.lock_and_run(config_fs.fs_context, bucket_lock_file, async () => {
```
romayalon, line 421: can you move await native_fs_utils.lock_and_run() to be before line 409? 
we tried to keep this function minimal and simple and with these changes it adds an inline complexity 
also, adding lock_and_run before all functions (except LIST) will be better as I see it, WDYT?

src/cmd/manage_nsfs.js
```diff
@@ -200,7 +222,24 @@ async function add_bucket(data) {
     data._id = mongo_utils.mongoObjectId();
     const parsed_bucket_data = await config_fs.create_bucket_config_file(data);
     await set_bucker_owner(parsed_bucket_data);
-    return { code: ManageCLIResponse.BucketCreated, detail: parsed_bucket_data, event_arg: { bucket: data.name }};
+
+    const reserved_tag_event_args = data.tag?.reduce((curr, tag) => {
```
romayalon, line 226: can you move it to a function?

src/sdk/bucketspace_fs.js
```diff
@@ -485,13 +496,49 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+                const bucket = await this.config_fs.get_bucket_by_name(name);
+                const { ns } = await object_sdk.read_bucket_full_info(name);
+                const is_bucket_empty = await BucketSpaceFS.#_is_bucket_empty(name, null, ns, object_sdk);
+
+                const tagging_object = bucket.tag?.reduce((curr, tag) => Object.assign(curr, { [tag.key]: tag.value }), {});
```
romayalon, line 515: I suggest moving the tagging logic to a function to keep this function simple and clean

src/sdk/bucketspace_fs.js
tangledbytes: I don't think all of the ops require locks. Most of them are indeed overwrites however for some of them, especially the ones which are both CLI and AWS S3 API controllable, I think we must have a lock to make sure no races result in reverting the other operation.

Yes, I will open an issue.

src/cmd/manage_nsfs.js
tangledbytes: I don't think even other ops would require locking but yeah, won't do any harm either. I will add it and hopefully that brings down the visual complexity.

romayalon: so at least mention it in the issue, we should think about it and since it'll be backported we shouldn't do this change in this PR

tangledbytes: Sure. I will mention it in the issue created for the previous comment.

romayalon (approved):

PR #8968, opened 2025-04-10
fix ceph CORS tests failure

aspandey:
Following Tests were failing -
s3tests_boto3/functional/test_s3.py::test_cors_presigned_put_object s3tests_boto3/functional/test_s3.py::test_cors_presigned_put_object_with_acl s3tests_boto3/functional/test_s3.py::test_cors_presigned_put_object_tenant s3tests_boto3/functional/test_s3.py::test_cors_presigned_put_object_tenant_with_acl

Issue -
Even after matching, we were returning the wrong origin and method for the cors calls.
We were mainly returning the configuration rather then sending back the match.


src/util/http_utils.js
```diff
@@ -719,8 +719,8 @@ function set_cors_headers_s3(req, res, cors_rules) {
         // https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html
         dbg.log0('set_cors_headers_s3: found matching CORS rule:', matched_rule);
         set_cors_headers(req, res, {
-            allow_origin: matched_rule.allowed_origins.includes('*') ? '*' : req.headers.origin,
-            allow_methods: matched_rule.allowed_methods.join(','),
+            allow_origin: match_origin,
```
jackyalbo, line 722: this is not correct. in case you have * in the allowed_origin list - you need to return * and not the sent origin - see here:
```
curl -X GET "${ENDPOINT}" \
  -H "Host: ${BUCKET_NAME}.s3.${AWS_REGION}.amazonaws.com" \
  -H "Date: ${DATE_NOW}" \
  -H "Authorization: AWS ${ACCESS_KEY}:${SIGNATURE}" \
  -H "Origin: https://allowed-origin.com" \
  -D - \
  --silent
HTTP/1.1 404 Not Found
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: GET, POST, PUT, DELETE
Access-Control-Expose-Headers: ETag, X-Amz-Version-Id
Vary: Origin, Access-Control-Request-Headers, Access-Control-Request-Method
x-amz-request-id: RVP2Z31CTC8AH42M
x-amz-id-2: YmZDrdYXE2wG/tcQ6qgCqGdF26qu2hDpTtVKd4vJF2vTUoPWZGHho5Cy061FVvuK67sRHq2nChc=
Content-Type: application/xml
Transfer-Encoding: chunked
Date: Thu, 10 Apr 2025 12:06:44 GMT
Server: AmazonS3```

File: src/util/http_utils.js [hunk 2]
```diff
@@ -719,8 +719,8 @@ function set_cors_headers_s3(req, res, cors_rules) {
...
+            allow_methods: match_method,
```
jackyalbo, line 723: This is also not correct - take a look at the above example
Allow-Methods in the rule are: GET, POST, PUT, DELETE
, and match_method is GET

src/util/http_utils.js
jackyalbo: this is tested with AWS. let me know if you need the script for running request with AWS bucket

jackyalbo (changes requested):

src/util/http_utils.js
jackyalbo: also see your change cause another s3_test to fail about CORS:
test_cors_origin_wildcard
please check in advance all of the tests are passing.

src/util/http_utils.js
aspandey: I agree, this was not the correct solution to the problem. 
Will revert back. I have found more issues and will discuss it with you.

PR #8964, opened 2025-04-09
NSFS | content dir, ignore failure to delete xattr from latest, if already deleted

nadavMiz:
### Describe the Problem
when adding to content dir at the same time after enabling versioning. both of them might try to remove the xattr from the directory. one of them will fail with ENODATA (The named attribute does not exist). ignore this error because attribute already deleted. see [removexattr](https://man7.org/linux/man-pages/man2/removexattr.2.html)

### Explain the Changes
1. ignore xattr clear error for data already deleted

### Issues: Fixed #8846 
NOTE - issue doesn't replicate on my environment. ran jest tests on CI 15 times without test failing but since it is a concurrency test, there may be other problems that didn't happen by chance

### Testing Instructions:
1. run `sudo npx jest test_versioning_concurrency -t "content dir multiple puts of the same key"`

- [ ] Doc added/updated
- [ ] Tests added

PR #8975, opened 2025-04-20
NC | Add non current timestamp xattr support

romayalon:
### Describe the Problem
We need a new xattr that describes when the version became non current to be assigned on uploads (move to .versions/) and removed on deletions (promote to latest). The new xattr will be used as a hint for the nc lifecycle non current days rule. 

### Explain the Changes
1. assigned Date.now() to XATTR_NON_CURRENT_TIMESTASMP xattr on the following cases - 
a. _move_to_dest_version() - when moving latest version to version_path.
b. _move_directory_content_xattr_to_versioned_file() - when moving directory object xattr to version file.
c. _assign_versions_to_fs_xattr() - when version is a delete marker only.

2. Removed XATTR_NON_CURRENT_TIMESTASMP value from file on _promote_version_to_latest().

3. Updated test_bucketspace_versioning.js tests to test these cases as well.
1. Gap - we should catch error on clear if ENODATA in general (not only for this change) - see comment https://github.com/noobaa/noobaa-core/pull/8964/files#r2051772675

### Testing Instructions:
1. sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_bucketspace_versioning.js

- [ ] Doc added/updated
- [x] Tests added


nadavMiz (approved):
LGTM

nadavMiz:
I don't think there is a problem, but we should add concurrency tests just to be safe

PR #8983, opened 2025-04-22
bucket notifications - use node name in eventSource

alphaprinz:
### Describe the Problem
Ramya from scale asked that eventSource will have the node's name (not system's name).

### Explain the Changes
1. Use correct node's name in eventSource instead of first key in system.json.

1. https://issues.redhat.com/browse/DFBUGS-2088

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/notifications_util.js
+++ b/src/util/notifications_util.js
@@ -513,10 +513,11 @@ function compose_notification_test(req) {
 
 function _get_system_name(req) {
 
+    //in NC case - return node name
     if (req && req.object_sdk && req.object_sdk.nsfs_system) {
-        const name = Object.keys(req.object_sdk.nsfs_system)[0];
+        const name = process.env.NODE_NAME || os.hostname();
         return name;
-    } else {
+    } else { //in containerized - return system's name
         //see comment on Notificator._can_run() for the require here
         const system_store = require('../server/system_services/system_store').get_instance();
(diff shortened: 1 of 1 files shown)


src/util/notifications_util.js
```diff
@@ -514,7 +514,7 @@ function compose_notification_test(req) {
 function _get_system_name(req) {
 
     if (req && req.object_sdk && req.object_sdk.nsfs_system) {
-        const name = Object.keys(req.object_sdk.nsfs_system)[0];
+        const name = process.env.NODE_NAME || os.hostname();
```
aspandey, line 517: If it is about returning the NODE name or hostname then why do we need to check this "if" condition?
We were having this "if" condition previously because we were using req.object_sdk inside "if" block.

But now, we do not need to check it.
I feel that just these two line should be enough -
        const name = process.env.NODE_NAME || os.hostname();
        return name;

aspandey (commented):
minor comment given above

src/util/notifications_util.js
alphaprinz: The if is to distinguish between NC and containerized (they have different handling).
I've added comments for clarification.

aspandey (approved):
Looks good to me

Issue #8870, opened 2025-03-17
lifecycle | PutBucketLifecycleConfiguration - missing validations


nadavMiz:
### Environment info
- NooBaa Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

The following validations are missing from PutBucketLifecycleConfiguration:
* according to AWS Date argument can only be at midnight UTC. in order to prevent time related errors this should be validated on policy level at PutBucketLifecycleConfiguration. AWS error: 
```
An error occurred (InvalidArgument) when calling the PutBucketLifecycleConfiguration operation: 'Date' must be at midnight GMT
```
* AbortIncompleteMultipartUpload can't be defined with either size or tags. error:
```
An error occurred (InvalidRequest) when calling the PutBucketLifecycleConfiguration operation: AbortIncompleteMultipartUpload cannot be specified with Tags.
```
* expire rule: can't have day/date with ExpiredObjectDeleteMarker

* AWS do not allow to specify multiple filter actions without And and on NooBaa it's allowed.

### Actual behavior
operation succeed
### Expected behavior
1. operation should fail with errors:

### More information - Screenshots / Logs / Other output
https://docs.aws.amazon.com/AmazonS3/latest/API/API_LifecycleExpiration.html#API_LifecycleExpiration_Contents

PR #8990, opened 2025-04-27
NC | lifecycle | Add Tests in POSIX Integration Tests - Part 1

shirady:
### Describe the Problem
We want to expand our test coverage in the NC lifecycle test.

### Explain the Changes
1. Create an enum for the status for easier reuse.
2. Fix the description mismatch (`id`) in the rule `lifecycle_rule_delete_all`.
3. Rename the file: remove the suffix `cli` and replace it with `posix_integration`, also inside the test titles, remove the `cli` and replace it with `nc`.
4. Separate with additional `describe` parts.
5. Remove the test "abort mpu by tags" as according to the [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configuration-examples.html):

> When you use the ExpiredObjectDeleteMarker S3 Lifecycle action, the rule cannot specify a tag-based filter.

Note: We added a validation for this - see issue #8870

7. Add cases where the status is `Disabled` and the rule doesn't run (in partially picked cases).
8. In versioning:
  - `DISABLED`, add the test cases for nested key and content dir
  - `ENABLED` only for nested key (as content dir is an unsupported case - see our [design doc of versioning](https://github.com/noobaa/noobaa-core/blob/master/docs/design/NsfsVersioning.md#out-of-scope)).
using Jest feature `it.each` (see [Jest docs](https://jestjs.io/docs/api#each) - we used array of objects and in the title we used the placeholder of `$description` to add information in the title).
Note: The tests were picked (we didn't want to run them on all the existing tests, although we can).
9. Add a warning to the JSDoc function `update_file_mtime`.
10. Format document.
11. Add the name test in our docs with a short explanation.

### Issues:
1. GAPS - we still have more tests that we plan to add to improve our coverage.

#### Automatic Tests
1. Please run: `sudo npx jest test_nc_lifecycle_posix_integration.test`

- [X] Doc added/updated
- [X] Tests added

### DIFF
--- a/src/test/unit_tests/jest_tests/test_nc_lifecycle_posix_integration.test.js
+++ b/src/test/unit_tests/jest_tests/test_nc_lifecycle_posix_integration.test.js
@@ -21,6 +21,11 @@ const crypto = require('crypto');
 const NsfsObjectSDK = require('../../../sdk/nsfs_object_sdk');
 const nb_native = require('../../../util/nb_native');
 
+const LIFECYCLE_RULE_STATUS_ENUM = Object.freeze({
+    ENABLED: 'Enabled',
+    DISABLED: 'Disabled'
+});
+
 const new_umask = process.env.NOOBAA_ENDPOINT_UMASK || 0o000;
 const old_umask = process.umask(new_umask);
 console.log('test_nc_lifecycle_cli: replacing old umask: ', old_umask.toString(8), 'with new umask: ', new_umask.toString(8));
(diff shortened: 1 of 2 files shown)


src/test/unit_tests/jest_tests/test_nc_lifecycle_posix_integration.test.js
```diff
@@ -1164,6 +1170,7 @@ async function create_object(sdk, bucket, key, size, is_old, tagging) {
 
 /**
  * update_file_mtime updates the mtime of the target path
+ * Warning - this operation would change the mtime of the file to 5 days ago - which means that it changes the etag / obj_id of the object
```
nadavMiz, line 1173: maybe also add a warning about it not changing the version_id so it will cause problems with versioning

nadavMiz (approved):
LGTM, minor comment

src/test/unit_tests/jest_tests/test_nc_lifecycle_posix_integration.test.js
shirady: Thanks, I will add.

Issue #8861, opened 2025-03-09
NC | Lifecycle | Gaps / TODOs from PRs


romayalon:
### Environment info
- NooBaa Version: 5.19
- Platform: NC

### Gaps/TODOs from PRs list - 

- [x] Add automatic unit tests to nc lifecycle lock mechanism - [8858](https://github.com/noobaa/noobaa-core/pull/8858).
- [x] Add automatic unit tests to events, timout, and status - [8860](https://github.com/noobaa/noobaa-core/pull/8860).
- [x] Add implementation of the processing of the 4 rules in POSIX. [8839](https://github.com/noobaa/noobaa-core/pull/8839).
- [x] Add implementation of the processing of expiration GPFS 
- [ ] Add implementation of the processing of nonCurrentDays GPFS.
- [x] Add notifications usage. [8839](https://github.com/noobaa/noobaa-core/pull/8839)
- [x] Decide on batch size. [8839](https://github.com/noobaa/noobaa-core/pull/8839)
- [x] Endpoints Stats should be supported? currently, we will not collect stats on the lifecycle worker process. [8839](https://github.com/noobaa/noobaa-core/pull/8839)
- [x] Add CLI unit tests [8839](https://github.com/noobaa/noobaa-core/pull/8839)
- [x] Add CLI documentation. [8839](https://github.com/noobaa/noobaa-core/pull/8839)
  - [x] specific example - health changes see [comment](https://github.com/noobaa/noobaa-core/pull/8893#discussion_r2010179516).
- [x] currently object-sdk is request-account uses _simple_load_requesting_account. need to check what account to use [8853](https://github.com/noobaa/noobaa-core/pull/8853)
- [x] add validation that policy doesn't have both AbortIncompleteMultipartUpload and tags / sizes in filter [8853](https://github.com/noobaa/noobaa-core/pull/8853) - fixed in #8914 
- [ ] modify list-parts for better error handling and add pagination. after use list-parts for lifecycle instead of current implementation [8853](https://github.com/noobaa/noobaa-core/pull/8853) - think maybe we should use here ILM policies as well.
- [x] enable working in batches. [8862](https://github.com/noobaa/noobaa-core/pull/8862)
- [ ] Decide when we want to write timestamp file and when we don't.
- [ ] Take account from cache in process_bucket()
- [ ] Check if we need to add DIRECTIRY_PLUS to our ILM policies or not.
[...]

PR #8991, opened 2025-04-27
NC | lifecycle | remove key_marker and version_marker from state when finished

nadavMiz:
set key_marker and version_marker to be undefined when list_object is is finished. do it so it wouldn't be confusing when looking at the status after the run has finished
 
### Explain the Changes
1. 

1. 

### Testing Instructions:

manual tests:
1. run lifecycle with batches
2. check status of the rules. should only have is_finished set to true

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/manage_nsfs/nc_lifecycle.js
+++ b/src/manage_nsfs/nc_lifecycle.js
@@ -458,6 +458,7 @@ class NCLifecycle {
             expire_state.key_marker = objects_list.next_marker;
             expire_state.is_finished = false;
         } else {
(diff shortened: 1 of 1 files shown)


shirady:
@noobaa/dev if you want you can link to https://github.com/noobaa/noobaa-core/issues/8861 in the section “Gaps/TODOs from testing”, and you can edit the test where it was found (one `expect` line).

romayalon (approved):

PR #9001, opened 2025-04-29
NC | lifecycle | Add Tests in POSIX Integration Tests - Part 3

shirady:
### Describe the Problem
We want to expand our test coverage in the NC lifecycle.

### Explain the Changes
1. Add tests related to versioning SUSPENDED based on the tests that we had in versioning ENABLED.
2. Rename the titles `versioning ENABLE` to `versioning ENABLED`.
3. Remove redundant `set_bucket_versioning` as it is set in the `beforeAll`.
4. Move the test `nc lifecycle - noncurrent expiration rule - expire versioning enabled bucket - regular key` under the `describe` of `describe noobaa nc - lifecycle versioning ENABLED` directly and rename it to `nc lifecycle - expiration rule - expire versioning ENABLED bucket - regular key`.
5. Copy the new tests that were added in PR #8992 and run them when the bucket is in versioning SUSPENDED, edit the "id" in the rule in those tests to match them, and edit the comments so it would be clearer.
6. Edit the titles of the test to have versioning ENABLED / SUSPENDED in them.

### Issues:
1. GAPS - we still have more tests that we plan to add to improve our coverage.

#### Automatic Tests
1. Please run: `sudo npx jest test_nc_lifecycle_posix_integration`

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/test/unit_tests/jest_tests/test_nc_lifecycle_posix_integration.test.js
+++ b/src/test/unit_tests/jest_tests/test_nc_lifecycle_posix_integration.test.js
@@ -1,5 +1,6 @@
 /* Copyright (C) 2016 NooBaa */
 /* eslint-disable max-lines-per-function */
+/* eslint-disable  max-lines */
 'use strict';
 
 // disabling init_rand_seed as it takes longer than the actual test execution
@@ -421,7 +422,7 @@ describe('noobaa nc - lifecycle versioning DISABLED', () => {
     }
 });
 
-describe('noobaa nc - lifecycle versioning ENABLE', () => {
+describe('noobaa nc - lifecycle versioning ENABLED', () => {
     const test_bucket_path = `${root_path}/${test_bucket}`;
     const test_key1_regular = 'test_key1';
     const test_key2_regular = 'test_key2';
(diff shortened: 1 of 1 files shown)


shirady:
Moved to draft, I have issues after rebase.

src/test/unit_tests/jest_tests/test_nc_lifecycle_posix_integration.test.js
```diff
@@ -558,36 +589,336 @@ describe('noobaa nc - lifecycle versioning ENABLE', () => {
...
+        await fs_utils.create_fresh_path(config_root, 0o777);
+        set_nc_config_dir_in_config(config_root);
+        await fs_utils.create_fresh_path(root_path, 0o777);
+        const res = await exec_manage_cli(TYPES.ACCOUNT, ACTIONS.ADD, account_options1);
+        const json_account = JSON.parse(res).response.reply;
+        console.log(json_account);
+        object_sdk = new NsfsObjectSDK('', config_fs, json_account, "DISABLED", config_fs.config_root, undefined);
+        object_sdk.requesting_account = json_account;
+        await object_sdk.create_bucket({ name: test_bucket });
+        await object_sdk.set_bucket_versioning({ name: test_bucket, versioning: "SUSPENDED" });
```
nadavMiz, line 812: since you set it to be Suspended from the beginning. wouldn't it mean we will never have more then one version per object?

nadavMiz: it is good to have a single test that we test that we actually don't do anything

shirady: There are tests where the bucket versioning was set to SUSPENDED at the beginning, for example:
- `nc lifecycle - versioning SUSPENDED - expiration rule - regular key`
- `nc lifecycle - versioning SUSPENDED - noncurrent expiration rule - expire older versions - `
in the rest of the tests, the bucket versioning is set to ENABLED and then to SUSPENDED, so they are just another validation of the original case.

nadavMiz (approved):
LGTM

PR #8982, opened 2025-04-22
NSFS | FS Napi | getpwnam concurrency issue

romayalon:
### Describe the Problem
While debugging https://issues.redhat.com/browse/DFBUGS-2227, it was found that getpwnam() sometimes return incorrect uid/gid per the username provided. 
After checking the getpwnam https://man7.org/linux/man-pages/man3/getpwnam.3.html man page, I saw that in fs_napi we use an unsafe multi threaded operation getpwnam() instead of the safe operation getpwnam_r().
image

### Explain the Changes
1. Changed fs_napi GetPwName() to use getpwnam_r() instead of getpwnam().
2. Added concurrency getPwName() tests.

### Issues: Fixed #xxx / Gap #xxx
1. https://issues.redhat.com/browse/DFBUGS-2227

### Testing Instructions:
1. Auto tests - sudo jest --testRegex=jest_tests/test_fs_napi_concurrency.test.js
2. Manual tests for reproducing the original bug - 
1. install warp
2. create at least 2 file system users, create 2 noobaa accounts, configured with distinguished names and create bucket for each account.
you can use the following script - 
```
storage_root_path="/nsfs_storage" ;
mkdir $storage_root_path/ -m 777;
for i in {1..2}; do 
groupadd -g 700$i user-$i ; 
useradd -c user-$i -m user-$i -p $(openssl passwd -1 pass-$i) -u 700$i -g 700$i ; 
mkdir $storage_root_path/user-$i-buckets/ -m 770; 
chown user-$i:user-$i $storage_root_path/user-$i-buckets/ ;  
noobaa-cli account add --name user-$i --new_buckets_path=$storage_root_path/user-$i-buckets/ --user=user-$i ;
for j in {1..1} ; do 
    mkdir $storage_root_path/user-$i-buckets/user-$i-bucket-$j/ -m 770 ;
    chown user-$i:user-$i $storage_root_path/user-$i-buckets/user-$i-bucket-$j/ ; 
    noobaa-cli bucket add --name=user.$i.bucket.$j --path=$storage_root_path/user-$i-buckets/user-$i-bucket-$j/ --owner=user-$i ; 
    done
done
```
3. Open 2 tabs, and run from each the following warp command with different accounts - 
4. tab 1 - 
```
warp put --host=localhost:6443 --access-key=<access_key of account 1> --secret-key=<secret_key of account 1>--obj.size=1k --duration=1m --disable-multipart  --bucket=<bucket of account 1> --tls --insecure --noclear --concurrent 1
[...]

PR #9010, opened 2025-05-05
NC | supplemental groups - getpwuid concurrency issue

### LINKED ISSUE #8985: Supplemental groups thread safety issue + dynamic supplemental groups PANIC in a container
### Environment info
- NooBaa Version: 4.19
- Platform: NC in oowy container

### Actual behavior
1. We use `getpwuid(uid)` in our os_linux.cpp and os_darwin.cpp code while it's an unsafe for multi threaded usage per the [getpwuid man page](https://man7.org/linux/man-pages/man3/getpwuid.3p.html), while the thread safe function we should use is `getpwuid_r()`.
```

The getpwuid() function need not be thread-safe.
...
The getpwuid_r() function is thread-safe and returns values in a
user-supplied buffer instead of possibly using a static data area
that may be overwritten by each call.
```
image
One more issue that might be related (or not) is the following - 

4.19 - 
```
Apr 22 13:28:37 node[1234]: PANIC: MUST SYSCALL FAILED: syscall(SYS_setgroups, groups.size(), &groups[0]) = -1 Operation not permitted (1) set_supplemental_groups() at ../src/native/util/os_linux.cpp:75
```
4.18.2 - 
```
Apr 24 11:10:23 d93e136f730c node[442]: PANIC: MUST SYSCALL FAILED: syscall(SYS_setgroups, orig_groups.size(), &orig_groups[0]) = -1 Op
eration not permitted (1) restore_user() at ../src/native/util/os_linux.cpp:63
Apr 24 11:10:23 d93e136f730c node[442]: qemu: uncaught target signal 6 (Aborted) - core dumped
```

### Expected behavior
1. All napi functions in use should be thread safe and we should not fail on PANIC. 

### Steps to reproduce
1. Run RHEL 9 container - `docker run --privileged --user root -d --platform=linux/amd64 oowy/redhat:ubi9`
2. login to the container - docker exec -it <container_id> bash
3. install NooBaa RPM per the instructions mentioned in https://github.com/noobaa/noobaa-core/blob/master/docs/NooBaaNonContainerized/GettingStarted.md#install
4. Check NooBaa restarts in the logs - `journalctl -u noobaa`

nadavMiz:
### Describe the Problem
as noted in #8982 : getpwuid function is not thread safe. so need to replace with thread safe getpwuid_r

### Explain the Changes
1. create a static variable for password buffer size in ThreadScope scope. initialize it from buffer length we got in fs_napi CTOR
2. replace getpwuid in both dynamic supplemental group usage for both linux and darwin.

### Testing Instructions:
1. for general supplemental groups tests run:
`sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nsfs_access.js`

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -168,7 +168,6 @@ typedef std::map<std::string, std::string> XattrMap;
 const char* gpfs_dl_path = std::getenv("GPFS_DL_PATH");
 
 int gpfs_lib_file_exists = -1;
-static long PASSWD_BUF_SIZE = -1;
 
 static int (*dlsym_gpfs_fcntl)(gpfs_file_t file, void* arg) = 0;
 
@@ -1410,12 +1409,13 @@ struct GetPwName : public FSWorker
     }
     virtual void Work()
     {
-        _buf.reset(new char[PASSWD_BUF_SIZE]);
+        const long passwd_buf_size = ThreadScope::get_passwd_buf_size();
+        _buf.reset(new char[passwd_buf_size]);
         if (!_buf) {
             SetSyscallError();
             return;
         }
-        int rc = getpwnam_r(_user.c_str(), &_pwd, _buf.get(), PASSWD_BUF_SIZE, &_getpwnam_res);
+        int rc = getpwnam_r(_user.c_str(), &_pwd, _buf.get(), passwd_buf_size, &_getpwnam_res);
         if (rc != 0) {
             SetSyscallError();
             return;
@@ -2415,11 +2415,7 @@ fs_napi(Napi::Env env, Napi::Object exports)
     exports_fs["DT_DIR"] = Napi::Number::New(env, DT_DIR);
     exports_fs["DT_LNK"] = Napi::Number::New(env, DT_LNK);
     exports_fs["PLATFORM_IOV_MAX"] = Napi::Number::New(env, IOV_MAX);
-    long passwd_bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
-    if (passwd_bufsize == -1) {
-        passwd_bufsize = 16384;
-    }
-    PASSWD_BUF_SIZE = passwd_bufsize;
+    ThreadScope::init_passwd_buf_size();
 
 
 #ifdef O_DIRECT
--- a/src/native/util/os.h
+++ b/src/native/util/os.h
@@ -41,6 +41,20 @@ public:
         change_user();
     }
 
+    static void init_passwd_buf_size()
+    {
+        long passwd_bufsize = sysconf(_SC_GETPW_R_SIZE_MAX);
+        if (passwd_bufsize == -1) {
+            passwd_bufsize = 16384;
+        }
+        passwd_buf_size = passwd_bufsize;
+    }
+
+    static long get_passwd_buf_size()
+    {
+        return passwd_buf_size;
+    }
+
     int add_thread_capabilities();
 
     const static uid_t orig_uid;
@@ -55,6 +69,8 @@ private:
     uid_t _uid;
     gid_t _gid;
     std::vector<gid_t> _groups;
+
+    static long passwd_buf_size;
 };
 
 } // namespace noobaa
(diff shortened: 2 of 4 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -2420,6 +2420,7 @@ fs_napi(Napi::Env env, Napi::Object exports)
         passwd_bufsize = 16384;
     }
     PASSWD_BUF_SIZE = passwd_bufsize;
+    ThreadScope::set_passwd_buf_size(PASSWD_BUF_SIZE);
```
guymguym, line 2423: Maintaining the same in two places is messy for no good reason. Let choose one place. It can be the ThreadScope for ease of sharing it. Just add a getter and use it also from fs_napi.

src/native/util/os.h
```diff
@@ -41,11 +41,17 @@ class ThreadScope
...
+    {
+        passwd_buf_size = buf_size;
+    }
+
     int add_thread_capabilities();
 
     const static uid_t orig_uid;
     const static gid_t orig_gid;
     const static std::vector<gid_t> orig_groups;
+    static int passwd_buf_size;
```
guymguym, line 54: use type long to match the sysconf output.

src/native/util/os_darwin.cpp
```diff
@@ -43,15 +43,18 @@ get_current_uid()
...
 get_supplemental_groups_by_uid(uid_t uid, std::vector<gid_t>& groups)
 {
-    // getpwuid will only indicate if an error happened by setting errno. set it to 0, so will know if there is a change
-    errno = 0;
-    struct passwd* pw = getpwuid(uid);
-    if (pw == NULL) {
-        if (errno == 0) {
+    std::unique_ptr<char[]> buf(new char[ThreadScope::passwd_buf_size]);
+    struct passwd pwd;
+    struct passwd *getpwnam_res;
```
guymguym, line 53: initialize pointer to null.
also there is no real need to change the name to be long and anoying, it can stay pw.

src/native/util/os_darwin.cpp
```diff
@@ -61,7 +64,7 @@ get_supplemental_groups_by_uid(uid_t uid, std::vector<gid_t>& groups)
     int ngroups = NGROUPS_MAX;
     //for some reason on mac getgrouplist accepts an array of int instead of gid_t. so need to create a vector of int and then insert it into groups
     std::vector<int> tmp_groups(ngroups);
-    if (getgrouplist(pw->pw_name, pw->pw_gid, &tmp_groups[0], &ngroups) < 0) {
+    if (getgrouplist(pwd.pw_name, pwd.pw_gid, &tmp_groups[0], &ngroups) < 0) {
```
guymguym, line 67: I think it's better to keep using the pointer pw->pw_name to access the info. the pwd struct is only a "storage" variable, and the pointer is the actual response.

src/native/fs/fs_napi.cpp
```diff
@@ -2419,7 +2419,7 @@ fs_napi(Napi::Env env, Napi::Object exports)
     if (passwd_bufsize == -1) {
         passwd_bufsize = 16384;
     }
-    PASSWD_BUF_SIZE = passwd_bufsize;
+    ThreadScope::set_passwd_buf_size(passwd_bufsize);
```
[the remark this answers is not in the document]
guymguym: it's not per thread. it's completely static in the process. we just needed somewhere to manage this that would be available to both fs_napi and the thread_scope code. I wouldn't be too worried about this integer  appearing too much like a per-thread constant because it's just a number we keep.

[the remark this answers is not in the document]
romayalon: exactly it's not really per thread because it's a static constant, it's not worrying :) it's just a bit weird it's there
let's move on with this solution

src/native/util/os_darwin.cpp
```diff
@@ -43,15 +43,19 @@ get_current_uid()
 const uid_t ThreadScope::orig_uid = getuid();
 const gid_t ThreadScope::orig_gid = getgid();
 const std::vector<gid_t> ThreadScope::orig_groups = get_process_groups();
+long ThreadScope::passwd_buf_size = -1;
```
romayalon, line 46: do we still need this line?

romayalon (approved):

PR #9019, opened 2025-05-12
Read postgres connection details from a file

dannyzaken:
### Explain the Changes
- Reading the DB connection details from a file. This supports passing the secret data from the operator in a more secure way, by volume mount instead of an env
- The existing environment variables are still inspected, allowing overrides if needed.
- The operator should be updated to use a volume mount following this PR.

1. Partial fix to https://issues.redhat.com/browse/DFBUGS-1466. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -28,6 +28,7 @@ const SensitiveString = require('./sensitive_string');
 const time_utils = require('./time_utils');
 const config = require('../../config');
 const ssl_utils = require('./ssl_utils');
+const fs_utils = require('./fs_utils');
 
 const DB_CONNECT_ERROR_MESSAGE = 'Could not acquire client from DB connection pool';
 mongodb.Binary.prototype[util.inspect.custom] = function custom_inspect_binary() {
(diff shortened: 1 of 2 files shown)


src/util/postgres_client.js
```diff
@@ -1496,22 +1497,27 @@ class PostgresClient extends EventEmitter {
...
+        if (process.env.POSTGRES_CONNECTION_STRING_PATH) {
             /** @type {import('pg').PoolConfig} */
             this.new_pool_params = {
-                connectionString: process.env.POSTGRES_CONNECTION_STRING,
+                connectionString: fs.readFileSync(process.env.POSTGRES_CONNECTION_STRING_PATH, "utf8"),
                 ...params,
             };
         } else {
+            // get the connection configuration. first from env, then from file, then default
+            const host = process.env.POSTGRES_HOST || fs_utils.try_read_file_sync(process.env.POSTGRES_HOST_PATH) || '127.0.0.1';
```
jackyalbo, line 1509: Shouldn't they be all in the same file? we will have different file for every env?

dannyzaken: This is the way secrets are mounted in kubernetes. Each key in the secret is mounted in a separate file

jackyalbo: We can save a a JSON string under the same key if we want, but I guess that as it's only few of them, it's not that bad.
Maybe don't use 4 different envs, and instead use one for all, and under them we can use user/password/dbname and port

jackyalbo: after a discussion with @dannyzaken decided to approve.

jackyalbo (approved):

PR #9029, opened 2025-05-16
Authentication for metrics and version endpoint

naveenpaul1:
### Describe the Problem

Noobaa provides multiple metrics endpoints for customers. The first one is Noobaa management service and the other one is Noobaa endpoint service. In the case of Noobaa management metrics endpoint, Noobaa provides a route to fetch the metrics. And for Noobaa endpoint metrics, metrics can be fetched by loadbalancer IP and port. Anyone with a valid URL can access these metrics endpoints without any kind of authentication.

### Explain the Changes
**noobaa -core**
1. JWT token will be authenticated Inside metrics_request_handler() and get_version_handler() 
2. JWT token will be verified against signature and role `metrics-auth`, with the role validation we can make sure other JWT tokens signed using the same secret won’t work.
3. Invalid auth token will return 403 error

1. [RHSTOR-7202](https://issues.redhat.com/browse/RHSTOR-7202)
4. [DFBUGS-1802](https://issues.redhat.com/browse/DFBUGS-1802)

### Testing Instructions:
**containerized deployment**
1. The customer should be able to access the bearer token from the secret `metrics-auth-secret`, secret can be used for accessing noobaa management and endpoint metrics/version.   

```
JWT_TOKEN=$(oc get secret/{token-secret-name} -n {namespace} -o jsonpath={.data.metrics_token} | base64 -d)    
curl -k -H "Authorization: Bearer ${JWT_TOKEN "https://{endpoint-loadbalancer-ingress-ip}:{endpoint-port}
curl -k -H "Authorization: Bearer ${JWT_TOKEN}" https://$(oc -n {namespace} get route noobaa-mgmt -o jsonpath='{.status.ingress[*].host}/version or metrics endpoints')
```
Design doc : https://ibm.ent.box.com/notes/1853310270159

- [X] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **New Features**
  - Added optional bearer token authorization for version and metrics endpoints.
  - Introduced configuration flags to enable or disable authentication for these endpoints.

- **Bug Fixes**
  - Corrected numbering sequence in configuration documentation for improved clarity.

- **Documentation**
  - Updated configuration documentation to reflect accurate numbering and new authentication options.

- **Style**
  - Enhanced code comments to improve readability and maintainability.

### DIFF
--- a/config.js
+++ b/config.js
@@ -1126,6 +1126,9 @@ config.DEFAULT_REGION = 'us-east-1';
 
 config.VACCUM_ANALYZER_INTERVAL = 86400000;
 
+config.NOOBAA_METRICS_AUTH_ENABLED = process.env.NOOBAA_METRICS_AUTH_ENABLED === 'true';
+config.NOOBAA_VERSION_AUTH_ENABLED = process.env.NOOBAA_VERSION_AUTH_ENABLED === 'true';
+
 /////////////////////
 //                 //
 //    OVERRIDES    //
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -349,6 +349,7 @@ function create_endpoint_handler(server_type, init_request_sdk, { virtual_hosts,
  * @param {import('http').ServerResponse} res 
  */
 function version_handler(req, res) {
+    if (config.NOOBAA_VERSION_AUTH_ENABLED && !http_utils.authorize_bearer(req, res)) return;
     const noobaa_package_version = pkg.version;
     res.statusCode = 200;
     res.setHeader('Content-Type', 'text/plain');
--- a/src/server/analytic_services/prometheus_reporting.js
+++ b/src/server/analytic_services/prometheus_reporting.js
@@ -21,6 +21,9 @@ const reports = Object.seal({
     endpoint: null // optional
 });
 
+const ROLE_METRICS = 'metrics';
+const ROLE_ADMIN = 'admin';
+
 let io_stats_complete = {};
 let ops_stats_complete = {};
 let fs_worker_stats_complete = {};
@@ -69,6 +72,12 @@ async function start_server(
         return;
     }
     const metrics_request_handler = async (req, res) => {
+        if (config.NOOBAA_METRICS_AUTH_ENABLED && !http_utils.authorize_bearer(req, res, [ ROLE_METRICS, ROLE_ADMIN ])) {
+            // Authorize bearer token metrics endpoint
+            // Role 'metrics' is used in operator RPC call, 
+            // Update operator RPC call first before changing role
+            return;
+        }
         // Serve all metrics on the root path for system that do have one or more fork running.
         if (fork_enabled) {
             // we would like this part to be first as clusterMetrics might fail.
--- a/src/server/web_server.js
+++ b/src/server/web_server.js
@@ -217,6 +217,8 @@ async function get_log_level_handler(req, res) {
 }
 
 async function get_version_handler(req, res) {
+    // Authorize bearer token version endpoint
+    if (config.NOOBAA_VERSION_AUTH_ENABLED && !http_utils.authorize_bearer(req, res)) return;
     const { status, version } = await getVersion(req.url);
     if (version) res.send(version);
     if (status !== 200) res.status(status);
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -946,6 +946,57 @@ function set_response_headers_from_request(req, res) {
     if (req.query['response-expires']) res.setHeader('Expires', req.query['response-expires']);
 }
 
+/**
+ * Authenticate JWT bearer token for metrics / version endpoints.
+ * Returns `true` on success, `false` after the function already sent an HTTP
+ * response (401/403) and the caller should terminate the handler early.
+ * 
+ * @param {import('http').IncomingMessage} req
+ * @param {import('http').ServerResponse} res
+ * @param {string[]} [roles]
+ */
+function authorize_bearer(req, res, roles = undefined) {
+    const { authorization, ...rest_headers } = req.headers;
+
+    const populate_response = () => {
+        res.statusCode = 403;
+        res.setHeader('Content-Type', 'text/plain');
+        res.end('Forbidden');
+    };
+
+
+    if (!authorization) {
+        dbg.error('Authentication required:', req.method, req.url, rest_headers);
+        // request lacks authentication, let the client know it's required with 401 Unauthorized
+        res.statusCode = 401;
+        res.setHeader('WWW-Authenticate', 'Bearer');
+        res.setHeader('Content-Type', 'text/plain');
+        res.end('Unauthorized');
+        return false;
+    }
+    if (!authorization.startsWith('Bearer ')) {
+        dbg.error('Authentication scheme must be Bearer:', req.method, req.url, rest_headers);
+        // authentication was provided but is invalid, return 403 Forbidden.
+        populate_response();
+        return false;
+    }
+    const token = authorization.slice('Bearer '.length);
+    let auth;
+    try {
+        auth = jwt_utils.authorize_jwt_token(token);
+    } catch (err) {
+        dbg.error('Authentication failed to verify JWT token:', req.method, req.url, rest_headers, err);
+        populate_response();
+        return false;
+    }
+    if (roles && !roles.includes(auth.role)) {
+        dbg.error('Authentication role is not allowed:', auth, roles, req.method, req.url, rest_headers);
+        populate_response();
+        return false;
+    }
+    return true;
+}
+
 exports.parse_url_query = parse_url_query;
 exports.parse_client_ip = parse_client_ip;
 exports.get_md_conditions = get_md_conditions;
@@ -984,3 +1035,4 @@ exports.CONTENT_TYPE_APP_JSON = CONTENT_TYPE_APP_JSON;
 exports.CONTENT_TYPE_APP_XML = CONTENT_TYPE_APP_XML;
 exports.CONTENT_TYPE_APP_FORM_URLENCODED = CONTENT_TYPE_APP_FORM_URLENCODED;
 exports.set_response_headers_from_request = set_response_headers_from_request;
+exports.authorize_bearer = authorize_bearer;
--- a/src/util/jwt_utils.js
+++ b/src/util/jwt_utils.js
@@ -27,6 +27,10 @@ function make_internal_auth_token(object = {}, jwt_options = {}) {
     return make_auth_token(object, jwt_options);
 }
 
+/**
+ * authorize jwt token by verifying it against the jwt secret
+ * @param {string} token
+ */
 function authorize_jwt_token(token) {
     try {
         return jwt.verify(token, get_jwt_secret());
--- a/docs/NooBaaNonContainerized/ConfigFileCustomizations.md
+++ b/docs/NooBaaNonContainerized/ConfigFileCustomizations.md
@@ -449,7 +449,7 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
     "LOG_TO_STDERR_ENABLED": false
     3. systemctl restart noobaa
     ```
-### 30. Notification log directory
+### 31. Notification log directory
 * <u>Key</u> `NOTIFICATION_LOG_DIR`
 * <u>Type</u> String
 * <u>Default</u> empty
@@ -462,7 +462,7 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
     "NOTIFICATION_LOG_DIR": "/etc/notif"
     3. systemctl restart noobaa
 
-### 31. Prometheus HTTP enable flag -
+### 32. Prometheus HTTP enable flag -
 * <u>Key</u>: `ALLOW_HTTP_METRICS`  
 * <u>Type</u>: Boolean  
 * <u>Default</u>: true  
@@ -476,7 +476,7 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
     3. systemctl restart noobaa
     ```
 
-### 32. Prometheus HTTPS enable flag -
+### 33. Prometheus HTTPS enable flag -
 * <u>Key</u>: `ALLOW_HTTPS_METRICS`  
 * <u>Type</u>: Boolean  
 * <u>Default</u>: true  
@@ -490,7 +490,7 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
     3. systemctl restart noobaa
     ```
 
-### 33. Notification space monitor frequency flag -
+### 34. Notification space monitor frequency flag -
 * <u>Key</u>: `NOTIFICATION_REQ_PER_SPACE_CHECK`
 * <u>Type</u>: Positive integer
 * <u>Default</u>: 0
@@ -504,7 +504,7 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
     3. systemctl restart noobaa
     ```
 
-### 34. Notification space monitor threshold flag -
+### 35. Notification space monitor threshold flag -
 * <u>Key</u>: `NOTIFICATION_SPACE_CHECK_THRESHOLD`
 * <u>Type</u>: Number
 * <u>Default</u>: 0.1
@@ -518,7 +518,7 @@ Warning: After setting this configuration, NooBaa will skip schema validations a
     3. systemctl restart noobaa
     ```
 
-### 34. Dynamic supplemental groups allocation flag -
+### 36. Dynamic supplemental groups allocation flag -
 * <u>Key</u>: `NSFS_ENABLE_DYNAMIC_SUPPLEMENTAL_GROUPS`
 * <u>Type</u>: boolean
 * <u>Default</u>: true
--- a/src/endpoint/endpoint_utils.js
+++ b/src/endpoint/endpoint_utils.js
@@ -4,6 +4,7 @@
 const querystring = require('querystring');
 const http_utils = require('../util/http_utils');
 const pkg = require('../../package.json');
+const config = require('../../config');
 
 function prepare_rest_request(req) {
     // generate request id, this is lighter than uuid
@@ -40,7 +41,9 @@ function parse_source_url(source_url) {
 }
 
 function set_noobaa_server_header(res) {
-    res.setHeader('Server', `NooBaa/${pkg.version}`);
+    if (!config.NOOBAA_VERSION_AUTH_ENABLED) {
+        res.setHeader('Server', `NooBaa/${pkg.version}`);
+    }
 }
 
 exports.prepare_rest_request = prepare_rest_request;


src/util/jwt_utils.js
```diff
@@ -27,11 +27,38 @@ function make_internal_auth_token(object = {}, jwt_options = {}) {
...
 }
 
-function authorize_jwt_token(token) {
+/**
+ * authorize jwt token. Validate the jwt token inside passed validate_jwt method
+ * @param {string} token
+ * @param {Function} validate_jwt_func
+ */
+function authorize_jwt_token(token, validate_jwt_func = undefined) {
+    if (token && token.includes('Bearer')) {
```
alphaprinz, line 36: Might be worth to change 'Bearer' into 'Bearer ' to make sure space is there.

src/util/jwt_utils.js
```diff
@@ -27,11 +27,38 @@ function make_internal_auth_token(object = {}, jwt_options = {}) {
...
+            if (err) {
+                throw err;
+            }
+            if (decoded.role !== "metrics-auth") {
+                throw new Error("Role based authorization failed");
+            }
+          });
     } catch (err) {
-        dbg.error('JWT VERIFY FAILED', token, err);
+        dbg.error('JWT verification failed', err);
```
alphaprinz, line 61: token was removed on purpose?

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -69,6 +70,21 @@ async function start_server(
         return;
     }
     const metrics_request_handler = async (req, res) => {
+        // TODO: This is a temporary condition to avoid the token authentication for metrics, 
+        // Need to add authentication for NSFS NC also after confirming flow with the Scale team
+        if (!process.env.NC_NSFS_NO_DB_ENV) {
```
tangledbytes, line 75: What if we make the authentication itself behind a config which is `true` by default but can be disabled via config (and consequently via env)?

Note: I lack all the context here so if this is meaningless please ignore.

tangledbytes: I still believe that it could be beneficial if we have a config dedicated to toggling on and off this feature.

src/util/jwt_utils.js
```diff
@@ -27,11 +27,35 @@ function make_internal_auth_token(object = {}, jwt_options = {}) {
...
+    const token = req.headers.authorization;
     try {
-        return jwt.verify(token, get_jwt_secret());
+        if (!token) {
+            throw new Error('Missing authorization header');
+        }
+        const decoded = authorize_jwt_token(token);
+        // Same role 'metrics-auth' is used in operator RPC call, 
+        // if needs to change update operator RPC call first
+        if (decoded.role !== "metrics-auth") {
```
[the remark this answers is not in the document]
naveenpaul1: @tangledbytes makes sense will update

src/util/jwt_utils.js
```diff
@@ -27,11 +27,35 @@ function make_internal_auth_token(object = {}, jwt_options = {}) {
...
 }
 
+/**
+ * authorize jwt token. Validate the jwt token inside passed validate_jwt method
+ * @param {string} token
+ */
 function authorize_jwt_token(token) {
+    if (token && token.includes('Bearer ')) {
+        token = token.split(' ')[1];
+    }
```
guymguym, line 37: this better move to the calling function, so that a token here is just a jwt token and not the full header.

src/util/jwt_utils.js
```diff
@@ -27,11 +27,35 @@ function make_internal_auth_token(object = {}, jwt_options = {}) {
...
+    return jwt.verify(token, get_jwt_secret());
+}
+
+/**
+ * autheticate jwt token for prometheus metrics and version request
+ * @param {nb.S3Request} req
+ */
+function authenticate_jwt_token(req) {
```
guymguym, line 45: this function doesn't belong in jwt_utils. lets move it to http_utils and rename it to something like authorize_bearer(req, res) so that it handles extracting the token from the header, calling jwt_utils.authorize_jwt_token(token) and then handles the errors to return a proper http AccessDenied 403 error.

src/util/jwt_utils.js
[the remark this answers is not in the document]
guymguym: This is a generic function, why should it allow specific roles like that.
Instead I would let it accept an array of roles that are allowed for the http request so that the caller can decide which roles the http request should be valid for.

src/server/web_server.js
```diff
@@ -218,6 +219,17 @@ async function get_log_level_handler(req, res) {
 }
 
 async function get_version_handler(req, res) {
+    try {
+        jwt_utils.authenticate_jwt_token(req);
+    } catch (err) {
+        res.writeHead(403, { 'Content-Type': 'text/plain' });
```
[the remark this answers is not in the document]
guymguym: First this error handling shouldn't be here. as noted below this should be in a function in http_utils.
Second - why are we authenticating the get_version call ??

guymguym (changes requested):

src/server/web_server.js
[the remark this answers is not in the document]
naveenpaul1: [Parent link](https://issues.redhat.com/browse/ODFRFE-26) for this [issue](https://issues.redhat.com/browse/RHSTOR-7202) mentioned that version also needs to be authenticated.

src/util/http_utils.js
```diff
@@ -946,6 +946,40 @@ function set_response_headers_from_request(req, res) {
...
+ * @param {string[]} roles
+ */
+function authorize_bearer(req, res, roles) {
+    let token = req.headers.authorization;
+    try {
+        if (!token) {
+            dbg.error('Missing authorization header : ', token);
+            throw new S3Error(S3Error.AccessDenied);
+        }
+        if (token.includes('Bearer ')) {
```
guymguym, line 962: this bearer header format should be required, not optional, if it doesnt have the bearer prefix, thats an error

src/util/http_utils.js
```diff
@@ -946,6 +946,40 @@ function set_response_headers_from_request(req, res) {
...
+            throw new S3Error(S3Error.AccessDenied);
+        }
+    } catch (err) {
+        dbg.error('JWT verification failed for token : ', token, err);
+        res.writeHead(403, { 'Content-Type': 'application/json' });
+        const reply = JSON.stringify({
+            error: 'AccessDenied',
+            message: err.message,
+        }, null, 2);
+        res.end(reply);
```
guymguym, line 979: you need to return something like false from here or throw, otherwise the caller will continue to process the request too

src/server/web_server.js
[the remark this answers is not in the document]
guymguym: We better add a config option that allow us to turn on/off the auth for both metrics and version routes (one config for each because different clients use it)

src/util/http_utils.js
```diff
@@ -946,6 +946,43 @@ function set_response_headers_from_request(req, res) {
...
+ * @param {string[]} roles
+ */
+function authorize_bearer(req, res, roles) {
+    let token = req.headers.authorization;
+    try {
+        if (!token) {
+            dbg.error('Missing authorization header : ', token);
+            throw new S3Error(S3Error.AccessDenied);
+        }
+        if (!token.includes('Bearer ')) {
```
guymguym, line 962: use startsWith instead of includes

src/util/http_utils.js
```diff
@@ -946,6 +946,43 @@ function set_response_headers_from_request(req, res) {
...
+            dbg.error('JWT token is missing the Bearer prefix, ', token);
+            throw new S3Error(S3Error.AccessDenied);
+        }
+        token = token.split(' ')[1];
```
guymguym, line 966: use token.slice('Bearer '.length)

src/util/http_utils.js
```diff
@@ -946,6 +946,43 @@ function set_response_headers_from_request(req, res) {
...
+        const decoded = jwt_utils.authorize_jwt_token(token);
+        // Role 'metrics-auth' is used in operator RPC call, 
+        // If needs to change update operator RPC call first
+        if (!roles.includes(decoded.role)) {
+            dbg.error('Bearer token authorization failed : ', token);
+            throw new S3Error(S3Error.AccessDenied);
+        }
+    } catch (err) {
+        dbg.error('JWT verification failed for token : ', token, err);
```
guymguym, line 975: better log the req.url and req.headers info and not just the opaque token

src/util/http_utils.js
```diff
@@ -946,6 +946,43 @@ function set_response_headers_from_request(req, res) {
...
+ * @param {nb.S3Request} req
+ * @param {nb.S3Response} res
+ * @param {string[]} roles
+ */
+function authorize_bearer(req, res, roles) {
+    let token = req.headers.authorization;
+    try {
+        if (!token) {
+            dbg.error('Missing authorization header : ', token);
+            throw new S3Error(S3Error.AccessDenied);
```
guymguym, line 960: why are we throwing an s3 error here? seems unrelated

src/util/http_utils.js
```diff
@@ -946,6 +946,43 @@ function set_response_headers_from_request(req, res) {
...
+            dbg.error('JWT token is missing the Bearer prefix, ', token);
+            throw new S3Error(S3Error.AccessDenied);
+        }
+        token = token.split(' ')[1];
+        const decoded = jwt_utils.authorize_jwt_token(token);
+        // Role 'metrics-auth' is used in operator RPC call, 
+        // If needs to change update operator RPC call first
+        if (!roles.includes(decoded.role)) {
+            dbg.error('Bearer token authorization failed : ', token);
+            throw new S3Error(S3Error.AccessDenied);
```
guymguym, line 972: again, s3 error is not related

src/server/web_server.js
```diff
@@ -218,6 +218,10 @@ async function get_log_level_handler(req, res) {
 }
 
 async function get_version_handler(req, res) {
+    // Authorize bearer token version endpoint
+    if (config.VERSION_AUTH_ENABLED) {
+        http_utils.authorize_bearer(req, res, [ "metrics-auth", "admin" ]);
```
guymguym, line 223: why is metrics-auth here? or should this be allowed for any role?

config.js
```diff
@@ -1118,6 +1118,9 @@ config.DEFAULT_REGION = 'us-east-1';
 
 config.VACCUM_ANALYZER_INTERVAL = 86400000;
 
+config.METRICS_AUTH_ENABLED = true;
+config.VERSION_AUTH_ENABLED = true;
```
guymguym, line 1122: i am not sure about the default being true. will that break anything?

src/util/http_utils.js
```diff
@@ -946,6 +946,43 @@ function set_response_headers_from_request(req, res) {
...
+        }
+    } catch (err) {
+        dbg.error('JWT verification failed for token : ', token, err);
+        res.writeHead(403, { 'Content-Type': 'application/json' });
+        const reply = JSON.stringify({
+            error: 'AccessDenied',
+            message: err.message,
+        }, null, 2);
+        res.end(reply);
+        return false;
```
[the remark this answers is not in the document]
guymguym: If auth fails, we need to respond, and then stop processing further. There are many reasons why continue to process the request if auth failed can lead to unexpected behavior and security issues, and that's currently what the code does. It's not a behavior we want to have even if you can't detect an issue on your tests.

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -69,6 +69,12 @@ async function start_server(
         return;
     }
     const metrics_request_handler = async (req, res) => {
+        // TODO: This is a temporary condition to avoid the token authentication for metrics, 
+        // Need to add authentication for NSFS NC also after confirming flow with the Scale team
+        if (!process.env.NC_NSFS_NO_DB_ENV && config.METRICS_AUTH_ENABLED) {
+            // Authorize bearer token metrics endpoint
+            http_utils.authorize_bearer(req, res, [ "metrics-auth", "admin" ]);
```
guymguym: I would rename the role to just "metrics" because that "-auth" doesn't make sense as a role name.

config.js
guymguym: What did you test for? did you check for noobaa-cli health, and upgrades, etc?

naveenpaul1: @guymguym In this PR we are skipping token authentication for NSFS NC. Metrics and version authentication only for the containerized Noobaa.

naveenpaul1: @guymguym  I did the upgrades test in my local cluster, 
1. Deployed Noobaa operator and core based on the 5.19 master branch
2. updated the operator, core and endpoint pods images with my custom changes image 
3. verified the Prometheus targets are up and srcapping.
also will ask the testing team to include upgrade test in their test plan.

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -69,6 +69,24 @@ async function start_server(
...
+                http_utils.authorize_bearer(req, res, [ "metrics", "admin" ]);
+            } catch (err) {
+                res.writeHead(403, { 'Content-Type': 'application/json' });
+                const reply = JSON.stringify({
+                    error: 'AccessDenied',
+                    message: err.message,
+                }, null, 2);
+                res.end(reply);
+                return;
+            }
```
guymguym: I would like to move the error handling into authorize_bearer because there are different cases there.
here what we should do is just call it and then return if it returned false.
```suggestion
            // Authorize bearer token metrics endpoint
            // Role 'metrics' is used in operator RPC call, 
            // Update operator RPC call first before changing role
            if (!http_utils.authorize_bearer(req, res, [ "metrics", "admin" ])) return;
```

src/server/web_server.js
```diff
@@ -218,6 +218,20 @@ async function get_log_level_handler(req, res) {
 }
 
 async function get_version_handler(req, res) {
+    // Authorize bearer token version endpoint
+    if (config.VERSION_AUTH_ENABLED) {
```
guymguym: I think we should skip this on NC for now, until we make a conscious decision to opt in to using auth for this.
```suggestion
    if (!process.env.NC_NSFS_NO_DB_ENV && config.VERSION_AUTH_ENABLED) {
```

src/server/web_server.js
```diff
@@ -218,6 +218,20 @@ async function get_log_level_handler(req, res) {
...
+            http_utils.authorize_bearer(req, res);
+        } catch (err) {
+            res.writeHead(403, { 'Content-Type': 'application/json' });
+            const reply = JSON.stringify({
+                error: 'AccessDenied',
+                message: err.message,
+            }, null, 2);
+            res.end(reply);
+            return;
+        }
```
guymguym: I would like to move the error handling into authorize_bearer because there are different cases there.
here what we should do is just call it and then return if it returned false.
```suggestion
            if (!http_utils.authorize_bearer(req, res)) return;
```

src/server/web_server.js
guymguym: oh, now I get it, this is only in web_server, which is not used for nc, but this also means that our endpoints have a separate version api that we did not handle, which is in endpoint.js version_handler() which will be served on noobaa-endpoint containers.

src/util/http_utils.js
```diff
@@ -946,6 +946,36 @@ function set_response_headers_from_request(req, res) {
...
+        const decoded = jwt_utils.authorize_jwt_token(token);
+        if (roles && !roles.includes(decoded.role)) {
+            dbg.error('Bearer token authorization failed : ', token);
+            throw new Error('Access Denied');
+        }
+    } catch (err) {
+        dbg.error('JWT verification failed for token: ', token, " req.url: ", req.url, " req.headers: ",
+             req.headers, err);
+        throw err;
+    }
```
guymguym: I suggest to handle the errors inside so that the error handling is unified. I also believe that for these errors the reply should be as simple as possible, as the error code is the only relevant information, and we do not want to reply any extra information.

```suggestion
    const auth_hdr = req.headers.authorization;
    if (!auth_hdr) {
        dbg.error('Authentication required:', req.method, req.url, req.headers);
        // request lacks authentication, let the client know it's required with 401 Unauthorized
        res.statusCode = 401;
        res.setHeader('WWW-Authenticate', 'Bearer');
        res.setHeader('Content-Type', 'text/plain');
        res.end('Unauthorized');
        return false;
    }

    if (!auth_hdr.startsWith('Bearer ')) {
        dbg.error('Authentication scheme must be Bearer:', req.method, req.url, req.headers);
        // authentication was provided but is invalid, return 403 Forbidden.
        res.statusCode = 403;
        res.setHeader('Content-Type', 'text/plain');
        res.end('Forbidden');
        return false;
    }

    const token = auth_hdr.slice('Bearer '.length);
    let auth;
    try {
        auth = jwt_utils.authorize_jwt_token(token);
    } catch (err) {
        dbg.error('Authentication failed to verify JWT token:', req.method, req.url, req.headers, err);
        res.statusCode = 403;
        res.setHeader('Content-Type', 'text/plain');
        res.end('Forbidden');
        return false;
    }

    if (roles && !roles.includes(auth.role)) {
        dbg.error('Authentication role is not allowed:', auth, roles, req.method, req.url, req.headers);
        res.statusCode = 403;
        res.setHeader('Content-Type', 'text/plain');
        res.end('Forbidden');
        return false;
    }

    return true;
```

guymguym (dismissed):
@naveenpaul1 see comments above. In addition to the version_handler in endpoint.js that I mentioned in a comment, there is another path which is set_noobaa_server_header() in endpoint_utils. This is called from endpoint.js and sets the response header very early in the handler, because we wanted to always return it. to disable this handling we need a separate config option and make sure that we don't change the current behavior to NC users.

src/util/http_utils.js
```diff
@@ -946,6 +946,52 @@ function set_response_headers_from_request(req, res) {
...
+        res.end('Forbidden');
+        return false;
+    }
+    if (roles && !roles.includes(auth.role)) {
+        dbg.error('Authentication role is not allowed:', auth, roles, req.method, req.url, req.headers);
+        res.statusCode = 403;
+        res.setHeader('Content-Type', 'text/plain');
+        res.end('Forbidden');
+        return false;
+    }
```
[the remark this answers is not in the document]
guymguym: nice catch rabbit. FYI @naveenpaul1

config.js
```diff
@@ -1125,6 +1125,9 @@ config.DEFAULT_REGION = 'us-east-1';
 
 config.VACCUM_ANALYZER_INTERVAL = 86400000;
 
+config.METRICS_AUTH_ENABLED = process.env.METRICS_AUTH_ENABLED || false;
+config.VERSION_AUTH_ENABLED = process.env.VERSION_AUTH_ENABLED || false;
```
guymguym: 1. envs are strings
2. we need to stop using generic env names and make sure to prefix with NOOBAA_
3. we already have a general purpose env mechanism to override configs by using the env.CONFIG_JS_ prefix, which I wish we have changed to be a simple env.NOOBAA_ prefix...

```suggestion
config.METRICS_AUTH_ENABLED = process.env.NOOBAA_METRICS_AUTH_ENABLED === 'true';
config.VERSION_AUTH_ENABLED = process.env.NOOBAA_VERSION_AUTH_ENABLED === 'true';
```

src/endpoint/endpoint.js
```diff
@@ -355,6 +355,10 @@ function create_endpoint_handler(server_type, init_request_sdk, { virtual_hosts,
  * @param {import('http').ServerResponse} res 
  */
 function version_handler(req, res) {
+    // TODO : update after auth added to NSFS NC
+    if (!process.env.NC_NSFS_NO_DB_ENV && config.VERSION_AUTH_ENABLED) {
```
guymguym: when the default is false, lets remove all these conditions of process.env.NC_NSFS_NO_DB_ENV. we don't want them, and once nsfs will want to enable this we will not require a code change.

src/endpoint/endpoint.js
```diff
@@ -355,6 +355,10 @@ function create_endpoint_handler(server_type, init_request_sdk, { virtual_hosts,
...
+        if (!http_utils.authorize_bearer(req, res)) return;
+    }
     const noobaa_package_version = pkg.version;
```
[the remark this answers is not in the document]
guymguym: this is wrong. authorize_bearer is not an async function. resolving.

guymguym:
> Well, both approaches are viable, depending on how you look. Which is the large deployment base, and which is a deviation. You're looking at NC as the base, with ODF needing to change these params; Naveen and Liran are looking at ODF as the base with NC needing to change. From a deployment number, I don't think there is even a question.

@nimrod-becker it's not about who is more important, it's about introducing breaking changes gradually. the core code should be backwards compatible unless we can't, so even if the new version of the operator should handle this correctly, there is no reason why older versions of the operator should break, and off course also non-operator versions, wether it's nsfs or devenv.

src/util/http_utils.js
```diff
@@ -946,6 +946,55 @@ function set_response_headers_from_request(req, res) {
...
 }
 
+/**
+ * Authenticate JWT bearer token for metrics / version endpoints.
+ * Returns `true` on success, `false` after the function already sent an HTTP
+ * response (401/403) and the caller should terminate the handler early.
+ * 
+ * @param {import('http').IncomingMessage} req
+ * @param {import('http').ServerResponse} res
+ * @param {string[]} roles
```
tangledbytes: Nitpick: 
```suggestion
 * @param {string[]} [roles]
```

src/server/web_server.js
```diff
@@ -217,6 +217,10 @@ async function get_log_level_handler(req, res) {
 }
 
 async function get_version_handler(req, res) {
+    // Authorize bearer token version endpoint
+    if (config.NOOBAA_VERSION_AUTH_ENABLED) {
+        if (!http_utils.authorize_bearer(req, res)) return;
+    }
```
tangledbytes: Nitpick:
```suggestion
    if (config.NOOBAA_VERSION_AUTH_ENABLED && !http_utils.authorize_bearer(req, res)) {
        return;
    }
```

src/util/http_utils.js
```diff
@@ -946,6 +946,55 @@ function set_response_headers_from_request(req, res) {
...
+ */
+function authorize_bearer(req, res, roles = undefined) {
+    const { authorization, ...rest_headers } = req.headers;
+    if (!authorization) {
+        dbg.error('Authentication required:', req.method, req.url, rest_headers);
+        // request lacks authentication, let the client know it's required with 401 Unauthorized
+        res.statusCode = 401;
+        res.setHeader('WWW-Authenticate', 'Bearer');
+        res.setHeader('Content-Type', 'text/plain');
+        res.end('Unauthorized');
```
tangledbytes, line 974: Maybe just create a function in the scope of this function (a const function) and call it all the three places to avoid dups.

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -69,6 +69,12 @@ async function start_server(
         return;
     }
     const metrics_request_handler = async (req, res) => {
+        if (config.NOOBAA_METRICS_AUTH_ENABLED) {
+            // Authorize bearer token metrics endpoint
+            // Role 'metrics' is used in operator RPC call, 
+            // Update operator RPC call first before changing role
+            if (!http_utils.authorize_bearer(req, res, [ "metrics", "admin" ])) return;
```
tangledbytes: Let's move these roles to package wide constants.

src/server/analytic_services/prometheus_reporting.js
```diff
@@ -69,6 +69,12 @@ async function start_server(
...
+        }
```
tangledbytes: Nitpick
```suggestion
        if (config.NOOBAA_METRICS_AUTH_ENABLED && !http_utils.authorize_bearer(req, res, [ROLE_METRICS, ROLE_ADMIN])) {
            // Authorize bearer token metrics endpoint
            // Role 'metrics' is used in operator RPC call, 
            // Update operator RPC call first before changing role
           return;
        }
```

tangledbytes (approved):

PR #9040, opened 2025-05-26
Added a fix to not include the temp dir in list objects

### LINKED ISSUE #8908: NSFS | List Objects with `s3 ls` of a Bucket Shows an Hidden Directory (Probably not in Normal Flow)
### Environment info
- NooBaa Version: 5.19
- Platform: NC

### Actual behavior
1. When using the AWS CLI of list objects with `s3 ls` I could see the hidden internal directory (`PRE .noobaa-nsfs_67d7ed5f51c5043305fee625/`). Note that the bucket ID is `67e2accc57cf31f3f0d80f31`.

### Expected behavior
1. When using the AWS CLI of list objects with `s3 ls` and not see any hidden internal directory.

### Steps to reproduce
I'm not sure what exactly I did but I can guess that I reused the same path for creating a bucket with the name `bucket-01`, and then deleting it (configuration-wise - delete with `sudo rm -rf /etc/noobaa.conf.d/`) and created the bucket again `bucket-01` with noobaa-cli In the same path with the same name I had the previous hidden folder so it shows it as an item.

`ls -al /Users/buckets/bucket-01`
```
total 8
drwxrwxrwx   5 shiradymnik  admin  160 Mar 25 21:39 .
drwxrwxrwx  19 root         admin  608 Mar 26 17:39 ..
drwxrwx---   3 1001         admin   96 Mar 17 11:40 .noobaa-nsfs_67d7ed5f51c5043305fee625
drwxrwx---   3 1001         admin   96 Mar 25 15:36 .noobaa-nsfs_67e2accc57cf31f3f0d80f31
-rw-rw----@  1 1001         admin   13 Mar 25 21:39 hello_world1.txt
```
`67e2accc57cf31f3f0d80f31` the bucket ID that was created with noobaa-cli in that time,
`67d7ed5f51c5043305fee625` probably old bucket ID that was deleted from the config, but it's underlying storage was not deleted.

`nc-user-1-s3 s3 ls s3://bucket-01`

```
                           PRE .noobaa-nsfs_67d7ed5f51c5043305fee625/
2025-03-25 21:39:28         13 hello_world1.txt
```

The outputs of `list-objects-v2` and `list-objects` are as expected (no hidden folder appears):
`nc-user-1-s3 s3api list-objects --bucket bucket-01`

```
{
    "Contents": [
        {
            "Key": "hello_world1.txt",
            "LastModified": "2025-03-25T19:39:28+00:00",
            "ETag": "\"mtime-d8ploctaemf4-ino-3v0xtu\"",
            "Size": 13,
            "StorageClass": "STANDARD",
            "Owner": {
                "DisplayName": "shira-1001",
                "ID": "67e2acb6a091ac33941ee437"
            }
        }
    ],
    "RequestCharged": null
}
```

`nc-user-1-s3 s3api list-objects-v2 --bucket bucket-01`

```
{
    "Contents": [
        {
            "Key": "hello_world1.txt",
            "LastModified": "2025-03-25T19:39:28+00:00",
            "ETag": "\"mtime-d8ploctaemf4-ino-3v0xtu\"",
            "Size": 13,
            "StorageClass": "STANDARD"
        }
    ],
    "RequestCharged": null
}
```

aayushchouhan09:
### Describe the Problem
Currently, when we do `list_objects` for a bucket then for non versioned bucket in the response it will print the temp/hidden dir (starts with `.noobaa-nsfs`) that is already present inside the directory but for versioned bucket the tmp/hidden dir cannot be printed in the response and we can found the `Version dir not found` logs.

### Explain the Changes
1.  Added a fix to exclude temp dir (starts with `.noobaa-nsfs`) from the object listing response.

1.  Fixed: #8908 

### Testing Instructions:
1. To run test: `sudo node node_modules/mocha/bin/mocha src/test/unit_tests/test_namespace_fs.js`
- [X] Tests added

### DIFF
--- a/src/test/unit_tests/test_ns_list_objects.js
+++ b/src/test/unit_tests/test_ns_list_objects.js
@@ -3,8 +3,10 @@
 
 const mocha = require('mocha');
 const assert = require('assert');
+const config = require('../../../config');
 
 const buffer_utils = require('../../util/buffer_utils');
+const fs_utils = require('../../util/fs_utils');
 
 // eslint-disable-next-line max-lines-per-function
 function test_ns_list_objects(ns, object_sdk, bucket) {
@@ -313,6 +315,33 @@ function test_ns_list_objects(ns, object_sdk, bucket) {
         });
     });
 
+    mocha.describe('list objects - hidden dir', function() {
+
+        this.timeout(10 * 60 * 1000); // eslint-disable-line no-invalid-this
+
+        const hidden_dir_bkt = 'test_ns_hidder_dir';
+        const dummy_file1 = 'dummy1';
+        const dummy_file2 = 'dummy2';
+        const hidden_dir = config.NSFS_TEMP_DIR_NAME + '-buck_id';
+
+        mocha.before(async function() {
+            fs_utils.create_fresh_path(`${ns.bucket_path}/${hidden_dir}`); // creating new hidden dir starting with .noobaa-nsfs
+            await create_keys([dummy_file1, dummy_file2]);
+        });
+        mocha.after(async function() {
+            await delete_keys([dummy_file1, dummy_file2]);
+        });
+
+        mocha.it('should not list the hidden/internal dir as common_prefixes', async function() {
+            const r = await ns.list_objects({
+                bucket: hidden_dir_bkt,
+                delimiter: '/'
+            }, object_sdk);
+            assert.deepStrictEqual(r.is_truncated, false);
+            assert.deepStrictEqual(r.common_prefixes, []);
+            assert.deepStrictEqual(r.objects.length, 2);
+        });
+    });
     mocha.describe('max keys test', function() {
 
         this.timeout(10 * 60 * 1000); // eslint-disable-line no-invalid-this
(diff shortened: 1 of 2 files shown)


romayalon:
@achouhan09 
Add tests please

achouhan09:
@romayalon Added tests. Thanks

src/test/unit_tests/test_ns_list_objects.js
```diff
@@ -313,6 +313,32 @@ function test_ns_list_objects(ns, object_sdk, bucket) {
         });
     });
 
+    mocha.describe('list objects - hidden dir', function() {
```
romayalon, line 318: how this test is relevant to the added fix? per my understanding this test is not creating a hidden directory that starts with .noobaa-nsfs except the one that is getting created during the object upload it won't list it as well

aayushchouhan09: Yes, you are correct updated it. Thanks

nadavMiz (approved):

romayalon (approved):

PR #8800, opened 2025-02-16
NC | S3 internal APIs gaps | Add versions CLI command | S3 internal API refactoring

romayalon:
### Explain the changes
1. Added a new CLI command called versions that returns the following versions status - 
```
> noobaa-cli versions
{
  "response": {
    "code": "VersionsStatus",
    "message": "Versions status retrieved successfully",
    "reply": {
      "rpm_source_code_versions": {
        "package_version": "5.19.0",
        "config_fs_version": "1.0.0"
      },
      "host_running_service_versions": {
        "package_version": "5.19.0",
        "config_fs_version": "1.0.0"
      },
      "config_dir_version": "1.0.0"
    }
  }
}
```

- rpm_source_code_versions consists of the package_version and the config_fs version.
- host_running_service_versions consists of the running service package_version and config_fs version.
- config_dir_version is the current config_dir_version registered in system.json.

2. added endpoint_fork_id and total_fork_count to be under the internal endpoint apis prefix - /_/ and did some refactoring there.
3. Added unit tests to the new CLI command.

### Issues: Fixed #xxx / Gap #xxx
1. Fixed gaps mentioned in https://github.com/noobaa/noobaa-core/pull/8786
2. Open question - is rpm_source_code_versions fetched correctly.

### Testing Instructions:
1. `sudo jest --testRegex=jest_tests/test_cli_versions`
2. Manual - 

Service is up - 
a. Tab 1 - start noobaa service - `sudo node noobaa-core/src/cmd/nsfs.js --debug`
b. Tab 2 - run noobaa-cli versions command - `noobaa-cli versions`

Service is down - 
a. Tab 1 - start noobaa service - `sudo node noobaa-core/src/cmd/nsfs.js --debug`, stop noobaa service - Ctrl + C
b. Tab 1 - run noobaa-cli versions command - `noobaa-cli versions`

system.json was never created - 
a. Tab 1 - run noobaa-cli versions command - `noobaa-cli versions`

- [x] Doc added/updated
- [x] Tests added

PR #9045, opened 2025-05-28
NC Health | add dedicated servers for each forks for health checks

nadavMiz:
### Describe the Problem
since we can't predict node forwarding of connections, we can't make sure health requests gets to all forks. currently a retry mechanism is used. this is unreliable, and might fail in certain cases. also under heavy system loads health requests can timeout. both those cases cause health to detect the system as being down and as a result causing a failover in CES systems. add a dedicated server for each fork for health requests

### Explain the Changes
1. add a new server for each fork for health requests. the port for each server will be in range between a defined base port and port + number of forks
2. add a value in config ENDPOINT_FORK_PORT_BASE that define the base for the dedicated ports for each fork
3. change health script to use the new dedicated servers instead of the endpoint main server to detect forks health
4. add a list of port offsets in the master fork. if fork is replaced, give his server the port of the replaced fork
5. add an argument for the health script to disable server validation - `disable_service_validation`. this value is to enable the use for testing both lacal and on CI

how the mechanism for distributing the different ports to each fork works:
Each fork is assigned a unique port based on its index offset from a base port. To allow a new fork to reuse the same port as a dead one, the primary process maintains a list of port offsets. When spawning a fork, the primary first waits for a "ready" message from the fork (indicating it's ready to receive instructions). Only then does the primary send the assigned port. This ensures the fork doesn't miss the port assignment message. The fork then starts its server on the received port. It's crucial that the primary starts listening for the "ready" message immediately after spawning to avoid missing it.

1. https://issues.redhat.com/browse/DFBUGS-2284

### Testing Instructions:
1. `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/unit_tests/test_nc_with_a_couple_of_forks.js`

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

- **New Features**
  - Introduced a dedicated health server for forked processes in non-containerized environments, improving health check reliability and visibility.
  - Added new configuration options for fork server port base, timeout, and retry attempts.
  - Health checks now query each fork individually, providing more accurate status reporting.

- **Bug Fixes**
  - Improved handling of worker process restarts to maintain consistent port assignments and robust server startup.

- **Tests**
  - Added comprehensive tests for fork health server behavior, including concurrency and fork restart scenarios.
  - Centralized and enhanced test utilities for mocking health check functions.

- **Chores**
  - Updated documentation and parameter types for server-related functions to reflect new health server support.

### DIFF
--- a/src/endpoint/endpoint.js
+++ b/src/endpoint/endpoint.js
@@ -43,6 +43,7 @@ const { SemaphoreMonitor } = require('../server/bg_services/semaphore_monitor');
 const prom_reporting = require('../server/analytic_services/prometheus_reporting');
 const { PersistentLogger } = require('../util/persistent_logger');
 const { get_notification_logger } = require('../util/notifications_util');
+const { is_nc_environment } = require('../nc/nc_utils');
 const NoobaaEvent = require('../manage_nsfs/manage_nsfs_events_utils').NoobaaEvent;
 const cluster = /** @type {import('node:cluster').Cluster} */ (
     /** @type {unknown} */ (require('node:cluster'))
@@ -57,7 +58,8 @@ const SERVICES_TYPES_ENUM = Object.freeze({
     S3: 'S3',
     STS: 'STS',
     IAM: 'IAM',
-    METRICS: 'METRICS'
+    METRICS: 'METRICS',
+    FORK_HEALTH: 'FORK_HEALTH',
 });
 
 const new_umask = process.env.NOOBAA_ENDPOINT_UMASK || 0o000;
@@ -117,11 +119,11 @@ async function main(options = {}) {
         const https_metrics_port = options.https_metrics_port || config.EP_METRICS_SERVER_SSL_PORT;
         /**
         * Please notice that we can run the main in 2 states:
-        * 1. Only the primary process runs the main (fork is 0 or undefined) - everything that 
+        * 1. Only the primary process runs the main (fork is 0 or undefined) - everything that
         *    is implemented here would be run by this process.
-        * 2. A primary process with multiple forks (IMPORTANT) - if there is implementation that 
-        *    in only relevant to the primary process it should be implemented in 
-        *    fork_utils.start_workers because the primary process returns after start_workers 
+        * 2. A primary process with multiple forks (IMPORTANT) - if there is implementation that
+        *    in only relevant to the primary process it should be implemented in
+        *    fork_utils.start_workers because the primary process returns after start_workers
         *    and the forks will continue executing the code lines in this function
         *  */
         const is_workers_started_from_primary = await fork_utils.start_workers(http_metrics_port, https_metrics_port,
@@ -202,14 +204,29 @@ async function main(options = {}) {
             { ...options, https_port: https_port_s3, http_port: http_port_s3, virtual_hosts, bucket_logger, notification_logger });
         await start_endpoint_server_and_cert(SERVICES_TYPES_ENUM.STS, init_request_sdk, { https_port: https_port_sts, virtual_hosts });
         await start_endpoint_server_and_cert(SERVICES_TYPES_ENUM.IAM, init_request_sdk, { https_port: https_port_iam });
-
+        const is_nc = is_nc_environment();
+        // fork health server currently runs only on non containerized enviorment
+        if (is_nc) {
+            // current process is the primary and only fork. start the fork server directly with the base port
+            if (cluster.isPrimary) {
+                await fork_message_request_handler({
+                    nsfs_config_root: options.nsfs_config_root,
+                    health_port: config.ENDPOINT_FORK_PORT_BASE
+                });
+            // current process is a worker so we listen to get the port from the primary process.
+            } else {
+                process.on('message', fork_message_request_handler);
+                //send a message to the primary process that we are ready to receive messages
+                process.send({ready_to_start_fork_server: true});
+            }
+        }
 
         // START METRICS SERVER
         if ((http_metrics_port > 0 || https_metrics_port > 0) && cluster.isPrimary) {
             await prom_reporting.start_server(http_metrics_port, https_metrics_port, false, options.nsfs_config_root);
         }
 
-        // TODO: currently NC NSFS deployments don't have internal_rpc_client nor db, 
+        // TODO: currently NC NSFS deployments don't have internal_rpc_client nor db,
         // there for namespace monitor won't be registered
         if (internal_rpc_client && config.NAMESPACE_MONITOR_ENABLED) {
             endpoint_stats_collector.instance().set_rpc_client(internal_rpc_client);
@@ -289,8 +306,6 @@ function create_endpoint_handler(server_type, init_request_sdk, { virtual_hosts,
                 return blob_rest_handler(req, res);
             } else if (req.url.startsWith('/total_fork_count')) {
                 return fork_count_handler(req, res);
-            } else if (req.url.startsWith('/endpoint_fork_id')) {
-                return endpoint_fork_id_handler(req, res);
             } else if (req.url.startsWith('/_/')) {
                 // internals non S3 requests
                 const api = req.url.slice('/_/'.length);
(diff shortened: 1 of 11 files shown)


src/util/fork_utils.js
```diff
@@ -101,6 +109,25 @@ function nsfs_io_stats_handler(msg) {
...
+ * NOTE - currently runs only on non containerized enviorment
+ * @param {*} worker
+ * @param {*} offset
+ * @param {*} nsfs_config_root
+ */
+async function _send_fork_server_message(worker, offset, nsfs_config_root) {
+    const is_nc = process.env.NC_NSFS_NO_DB_ENV && process.env.NC_NSFS_NO_DB_ENV === 'true';
+    if (is_nc && offset >= 0) {
+        //wait for the worker to be ready to receive messages
+        await once(worker, 'message');
```
romayalon, line 123: add timeout so we won't wait forever?

src/util/fork_utils.js
```diff
@@ -66,6 +71,9 @@ async function start_workers(metrics_port, https_metrics_port, nsfs_config_root,
             console.warn(`${exit_events.length} exit events in the last ${config.NSFS_EXIT_EVENTS_TIME_FRAME_MIN} minutes,` +
                 ` max allowed are: ${config.NSFS_MAX_EXIT_EVENTS_PER_TIME_FRAME}`);
             const new_worker = cluster.fork();
+            const offset = fork_port_offsets.findIndex(id => id === worker.id);
+            await _send_fork_server_message(new_worker, offset, nsfs_config_root);
+            fork_port_offsets[offset] = new_worker.id;
             console.warn('WORKER re-started', { id: new_worker.id, pid: new_worker.process.pid });
```
romayalon, line 77: should we print also the new port for each fork?

src/endpoint/endpoint.js
```diff
@@ -531,8 +543,25 @@ function unavailable_handler(req, res) {
     res.end(reply);
 }
 
+function fork_main_handler(req, res) {
+    endpoint_utils.set_noobaa_server_header(res);
+    endpoint_utils.prepare_rest_request(req);
+    if (req.url.startsWith('/endpoint_fork_id')) {
+        return endpoint_fork_id_handler(req, res);
+    }
```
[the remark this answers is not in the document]
nadavMiz: we don't have anything else to do. we don't actually need a request in the url. made it mostly as preperation in case we want to expand this function. I can throw an exception for invalid request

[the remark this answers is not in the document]
romayalon: we should throw something like internal_api_error - https://github.com/noobaa/noobaa-core/blob/master/src/endpoint/endpoint.js#L300C28-L300C46 like we do for internal apis of the endpoint
also, I started refactoring internal calls to be under /_/ - https://github.com/noobaa/noobaa-core/pull/8800
since you are doing this change i'll wait for you to finish with it and will rebase after

src/util/fork_utils.js
nadavMiz: added 5 minute timeout. its worth mentioning that if the fork is killed the listener is canceled and the code will not run entirely

src/endpoint/endpoint.js
```diff
@@ -531,8 +546,37 @@ function unavailable_handler(req, res) {
     res.end(reply);
 }
 
+/**
+ * handler for the inidivdual fork server. used to handle requests the get the worker id
+ * currently used to check if fork is alive by the health script
+ * @type {EndpointHandler}
+ */
```
romayalon:
```suggestion
 * @param {EndpointRequest} req 
 * @param {import('http').ServerResponse} res 
 * @returns {EndpointHandler}
 */
```

src/nc/nc_utils.js
```diff
@@ -24,7 +24,15 @@ function check_root_account_owns_user(root_account, account) {
     return root_account._id === account.owner;
 }
 
+/**
+ * @returns {boolean} true if the current environment is a NooBaa non containerized environment
+ */
+function is_nc_enviorment() {
```
romayalon:
```suggestion
function is_nc_environment() {
```

src/util/fork_utils.js
```diff
@@ -101,6 +111,29 @@ function nsfs_io_stats_handler(msg) {
...
+    if (is_nc && offset >= 0) {
+        //wait for the worker to be ready to receive messages
+        try {
+            await once(worker, 'message', { signal: AbortSignal.timeout(5 * 60000) });
+            worker.send({
+                nsfs_config_root: nsfs_config_root,
+                health_port: config.ENDPOINT_FORK_PORT_BASE + offset
+            });
+        } catch (err) {
+            dbg.warn(`Timeout: It took more than 5 minute to get a message from worker ${worker.id} do not send start server message`);
```
romayalon, line 155: is this all the error handling we really need here? what will happen to forks that had issues to communicate with the parent process

src/util/fork_utils.js
```diff
@@ -66,7 +72,11 @@ async function start_workers(metrics_port, https_metrics_port, nsfs_config_root,
             console.warn(`${exit_events.length} exit events in the last ${config.NSFS_EXIT_EVENTS_TIME_FRAME_MIN} minutes,` +
                 ` max allowed are: ${config.NSFS_MAX_EXIT_EVENTS_PER_TIME_FRAME}`);
             const new_worker = cluster.fork();
-            console.warn('WORKER re-started', { id: new_worker.id, pid: new_worker.process.pid });
+            const offset = fork_port_offsets.findIndex(id => id === worker.id);
+            _send_fork_server_message(new_worker, offset, nsfs_config_root);
```
romayalon: why the await is removed here as well?

src/util/fork_utils.js
```diff
@@ -101,6 +111,29 @@ function nsfs_io_stats_handler(msg) {
...
+ * @param {*} worker
+ * @param {*} offset
+ * @param {*} nsfs_config_root
+ */
+async function _send_fork_server_message(worker, offset, nsfs_config_root) {
+    const is_nc = is_nc_enviorment();
+    if (is_nc && offset >= 0) {
+        //wait for the worker to be ready to receive messages
+        try {
+            await once(worker, 'message', { signal: AbortSignal.timeout(5 * 60000) });
```
romayalon: you mentioned in the description the following - 
```
primary first waits for a "ready" message from the fork (indicating it's ready to receive instructions). Only then does the primary send the assigned port. This ensures the fork doesn't miss the port assignment message. The fork then starts its server on the received port. It's crucial that the primary starts listening for the "ready" message immediately after spawning to avoid missing it.

```
where is this logic? shouldn't process await for ready message specifically?

nadavMiz: I can add it. I wanted to make it simple, as this message is printed at the start of the run, so it is the first message anyway

src/util/fork_utils.js
nadavMiz: I found an issue that if the fork fails before it sends the message, the once await will be canceled so we will never finish this function. this causes that the next line the assigned the worker id in fork_port_offsets line never runs. so the worker that replace the one that fail will not know which port to give it. removing the await ensures we assign the port while still waiting for the `once`, only  without waiting for it here

src/util/fork_utils.js
nadavMiz: the fork that issue the communications keeps running. right, this will cause a broken scenario where the fork runs without his designated fork server.I though that if he didn't send the message, it will be equivalent to the case that it didn't start the server at all. I guess I can instead kill it, and hope that it will run the message in the next run. we also need to think on the case that it just didn't run for 5 minutes because the system is overloaded. I gave it 5 minute purposefully because I wanted to give it a long time. for example for 1 minutes it did occasionally not send the message in time

nadavMiz: OK, in order to solve all of those issues, I removed the once. instead I moved calling of the function to the existing fork message handler, and renamed it. for the specific ready message. in case the fork will die, it will just be cancelled

src/util/fork_utils.js
nadavMiz: moved the handling to the existing handler (used before only for metrics). will handle this specific message

src/util/fork_utils.js
```diff
@@ -84,20 +100,55 @@ async function start_workers(metrics_port, https_metrics_port, nsfs_config_root,
     return false;
 }
 
-function nsfs_io_stats_handler(msg) {
-    if (msg.io_stats) {
+//global variable as it is shared between all workers
+let fork_server_timeout = config.NC_FORK_SERVER_TIMEOUT_MIN;
+function create_worker_message_handler(params) {
+    let fork_server_timer;
+    if (is_nc_environment()) {
```
romayalon: per my understanding we want to set this timeout when the fork is getting created and not when the message event is getting called, how this works with my expectation?

src/util/fork_utils.js
```diff
@@ -84,20 +100,55 @@ async function start_workers(metrics_port, https_metrics_port, nsfs_config_root,
...
+        fork_server_timer = setTimeout(async () => {
+            dbg.error(`Timeout: It took more than ${fork_server_timeout} minutes to get a ready message from worker ${params.worker.id}, killing the worker`);
+            await params.worker.kill();
+            fork_server_timeout += 1; // increment the timeout for the next worker
```
romayalon: why should we add time for the next worker?

src/util/fork_utils.js
```diff
@@ -84,20 +100,55 @@ async function start_workers(metrics_port, https_metrics_port, nsfs_config_root,
...
+        }, fork_server_timeout * 60 * 1000);
```
romayalon: per the slack conversation, I was expecting to see that after x numbers of workers restart we will stop, where is this change?

src/util/fork_utils.js
nadavMiz: its when the message is getting created. when we start listening to it. its just for organization, so it would be easier to understand what this timeout relates to, and so it will always happen if start waiting at different points. I can move it for after the fork start if you like.

src/util/fork_utils.js
nadavMiz: in case the worker just took more then 3 minutes to start. we wait more the next time. note that I made the forks share the timeout, so if two forks takes more then 3 minutes to load, they will both increase the timeout.

src/util/fork_utils.js
nadavMiz: right I'll add it

src/util/fork_utils.js
romayalon: I just don't see how it helps to start setting the timeout after we got the message event, since this timeout is our tool for making sure we got the event message

nadavMiz: sorry I didn't explain well. its not after we got the message event. its when we set the listener for the said message event. right before we start listening to it

romayalon (approved):

PR #9057, opened 2025-06-04
Code Refactor | Remove Lambda/func_services

naveenpaul1:
### Explain the Changes
1. Remove Lambda and func_services relate code

1. not using Lambdaand func_services in noobaa code, fix will remove unwanted code from codebase

### Testing Instructions:
1. test complete noobaa core functionalities

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **Removals**
  - Removed all support for serverless functions (Lambda-like features), including function creation, management, invocation, and related APIs.
  - Eliminated bucket event triggers and their management from buckets and objects.
  - Deleted all user interface and REST endpoints related to functions and triggers.
  - Removed all related system tests, unit tests, and example/test Lambda functions.

- **Chores**
  - Cleaned up codebase by deleting unused schemas, indexes, migration logic, and trigger dispatching code related to functions and triggers.
  - Simplified asynchronous code flows by replacing promise chains with async/await in notification dispatchers.
  - Removed all references and registrations of function services from server and test setups.


src/upgrade/migration_to_postgres.js
```diff
@@ -24,8 +24,6 @@ const object_part_indexes = require('../server/object_services/schemas/object_pa
 const data_chunk_indexes = require('../server/object_services/schemas/data_chunk_indexes');
 const data_block_indexes = require('../server/object_services/schemas/data_block_indexes');
 const node_schema = require('../server/node_services/node_schema');
-const func_schema = require('../server/func_services/func_schema');
```
jackyalbo, line 27: let's just get rid of this whole file and also in general, try and remove as much as you can from the mongo migration as in next tasks we will want to remove all of mongo anyway.

src/sdk/namespace_merge.js
```diff
@@ -181,28 +144,14 @@ class NamespaceMerge {
...
         // TODO: Decide which one to return (currently we do not support versioning on our namespaces)
         return _.first(reply);
     }
 
 
     async delete_multiple_objects(params, object_sdk) {
-        const operation = 'ObjectRemoved';
-        const load_for_trigger = object_sdk.should_run_triggers({ active_triggers: this.active_triggers, operation });
-        const head_res = load_for_trigger && await this._ns_map(ns => P.map(params.objects, async obj => {
+        const head_res = await this._ns_map(ns => P.map(params.objects, async obj => {
```
jackyalbo: I think we can just remove this no? as load_for_trigger will always be false

src/server/notifications/dispatcher.js
```diff
@@ -241,7 +240,7 @@ class Dispatcher {
                     l.account = _.pick(account.record, 'email');
                     l.account.linkable = account.linkable;
                 }
-                return P.resolve(log_item.func && func_store.get_by_id_include_deleted(log_item.func, 'funcs'));
+                return P.resolve(log_item.func);
```
jackyalbo: I think we don't need this anymore? we can remove. Also here good chance to change to async/await.
also we probably need to clean 245-249.

src/upgrade/migration_to_postgres.js
naveenpaul1: @jackyalbo Maybe I will create a different PR for Mongo changes. It will make the changes and review simpler, and if anything goes wrong we will have much more chance to find the issue. WDYT?

src/sdk/namespace_merge.js
naveenpaul1: @jackyalbo head_res is used inside _merge_multiple_delete_responses() to identify all the request objects are deleted or not.

jackyalbo: yes, but it seems that now load_for_trigger will always be false. so we are probably handling also false option in our code. Make sure our code handles false; if so, we can remove this map.

jackyalbo: @naveenpaul1 So can we remove all of this code? lines 154-167 and set head_res to false?
I think you can also remove handling of head _res form _merge_multiple_delete_responses and _handle_single_namespace_deletes as well

naveenpaul1: @jackyalbo Inside method [_pick_ns_obj_reply()](https://github.com/noobaa/noobaa-core/blob/master/src/sdk/namespace_merge.js#L277) we are accessing `obj`(object.obj), it is accessible only from `head_res` and we check which object to return using the `create_time`.

naveenpaul1: Inside method  [_pick_ns_obj_reply](https://github.com/noobaa/noobaa-core/blob/master/src/sdk/namespace_merge.js#L277) we are accessing obj, it is accessible only from head_res and we check which object to return using the create_time. We need to remove this reference aslo.

jackyalbo: Looks like you handled it correctly. I'm approving

jackyalbo (approved):

PR #9066, opened 2025-06-05
persistent logger - check fd before write in append() (dfs 2624)

alphaprinz:
### Describe the Problem
Scenario is:
There are two processes.
One is the nsfs s3 endpoint server. This process writes to the active log file.
Second is the "notificator". This process reads the active file and deletes it.

While handling an s3 op, nsfs calls PersistentLogger.append().
The fd is inited to the active file.
After fd is inited, but before write() is invoked, the notificator processes and deletes the active file.
nsfs process' _poll_active_file_change() closes the fd.
When nsfs invokes write() in a closed fd, error is thrown and nsfs panics.

Note that this is not a likely scenario, and the heavy load used for dfs2624 correlates with this.

### Explain the Changes
Check fd before writing. If closed, reinit.

1. https://issues.redhat.com/browse/DFBUGS-2624

### Testing Instructions:
Because above scenario is is unlikely, I was able to reproduce it with looping the write() for ~100k times and simultaneously running the notificator.

Another option is with debug-
-add await timer between fd init and write
-break nsfs process at the timer
-run notificator ("manage_nsfs notification").
-wait for notifcator to end
-wait for _poll_active_file_change interval (default is 10 seconds)
-step over await timer
-the async _poll_active_file_change should run and close the fd
-now fd is closed and next line for nsfs is the write.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of log writing to handle cases where log files are deleted or closed by other processes during operation.

### DIFF
--- a/src/util/persistent_logger.js
+++ b/src/util/persistent_logger.js
@@ -8,6 +8,7 @@ const P = require('./promise');
 const semaphore = require('./semaphore');
 const { NewlineReader } = require('./file_reader');
 const dbg = require('./debug_module')(__filename);
+const APPEND_ATTEMPTS_LIMIT = 5;
 
 /**
  * PersistentLogger is a logger that is used to record data onto disk separated by newlines.
@@ -105,10 +106,20 @@ class PersistentLogger {
      * @param {string} data 
      */
     async append(data) {
-        const fh = await this.init();
-
         const buf = Buffer.from(data + '\n', 'utf8');
-        await fh.write(this.fs_context, buf, buf.length);
+
+        for (let attempt = 0; attempt < APPEND_ATTEMPTS_LIMIT; ++attempt) {
+            const fh = await this.init();
+            //if another process has deleted the active file,
+            //this process' _poll_active_file_change might have closed the fd
+            //in that case fd is -1
+            //in order to avoid inter-process locking, we just re-init
+            //the fd to the new active file.
+            if (fh.fd === -1) continue;
+            await fh.write(this.fs_context, buf, buf.length);
+            break;
+        }
+
         this.local_size += buf.length;
     }
 


guymguym:
we have used file locks for that. producers hold a share lock, and the consumer first renames the file and then takes an exclusive lock to ensure no more producers use the active file. please sync with @tangledbytes on how this is handled fitb glacier logs.

nimrod-becker:
@coderabbitai review

tangledbytes:
@alphaprinz the way we handle this on the tape ops is that the writers must hold shared locks while a reader must acquire an exclusive lock.

src/util/persistent_logger.js
```diff
@@ -105,10 +106,20 @@ class PersistentLogger {
...
      */
     async append(data) {
-        const fh = await this.init();
-
         const buf = Buffer.from(data + '\n', 'utf8');
-        await fh.write(this.fs_context, buf, buf.length);
+
+        for (let attempt = 0; attempt < APPEND_ATTEMPTS_LIMIT; ++attempt) {
+            const fh = await this.init();
+            //if another process has deleted the active file,
```
tangledbytes, line 113: The other process should not be able to delete the file if there is >= 1 process that exists as a writer. The reason is that the delete would happen only after `_process` would finish processing the active log file and in order to finish this the `collect_and_process` must have finished which must acquire an `EXCLUSIVE` lock before doing so.

tangledbytes:
It seems that we are deleting the log file after releasing the lock. The following sequence can cause the issue:
- Reader (R) acquires EX
- `collect_and_process` of R closes the log file
- Writer (W) which had already opened the FD but was waiting for the lock acquires SH.
- W checks the `nlink` which is still >= 1.
- R's `_process` deletes the file post processing **after** it has already lost the lock due to `collect_and_process` closing the file.
- W succeeds in init but the underlying file got deleted.
- W's file replace poller detects the deletion and closes the FD.
- W writes and panics.

alphaprinz:
replacement PR https://github.com/noobaa/noobaa-core/pull/9183

PR #9069, opened 2025-06-09
NC | Lifecycle | GPFS | fix skip of incorrect noobaa uploads internal directory

romayalon:
### Describe the Problem
As part of code reviewing, I noticed we are skipping `.noobaa_nsfs%/%` directory instead of `.noobaa-nsfs%/%` directory during the nc lifecycle on GPFS flow. Although when setting empty prefix filter I couldn't make this directory get deleted using the lifecycle process.

### Explain the Changes
1. changed to skip the correct internal uploads directory.

### Issues: Fixed #xxx / Gap #xxx
1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **Refactor**
  - Updated internal directory path construction to use a configurable directory name instead of a hardcoded value for lifecycle policy generation.

- **Tests**
  - Adjusted tests to align with the new configurable directory name for temporary NSFS directories.

PR #9073, opened 2025-06-10
NC | Lifecycle | An internal file/folder should be blocked for deletion by the lifecycle worker

romayalon:
### Describe the Problem
An internal file/folder should be blocked for deletion by the lifecycle worker, we are not supposed to see an issue due to the whole flow but I added it in order to be extra safe on the filter func as well.

1. POSIX flow - we shouldn't see this issue at all because list_objects() is used and should not be return an internal file.
2. GPFS flow  - also shouldn't happen because on the ILM policy we avoid internal directory/past versions and when seeing .folder file we convert the key to be the directory name.

### Explain the Changes
1. lifecycle_utils.build_lifecycle_filter() - returned false if the object key is inside the bucket temp uploads dir / any .versions/ dir or directory object (.folder).

1. 

### Testing Instructions:
1. `sudo jest --testRegex=jest_tests/test_nc_lifecycle.test`

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved lifecycle filtering to prevent deletion of temporary files, folder marker objects, and versioned objects.
- **Tests**
  - Added new test cases to verify deletion fails for internal and special system files and directories.
  - Enhanced test assertions for more flexible error checking.
- **Chores**
  - Removed unnecessary debug logging from a CORS test case.

### DIFF
--- a/src/util/lifecycle_utils.js
+++ b/src/util/lifecycle_utils.js
@@ -148,6 +148,11 @@ function build_lifecycle_filter(params) {
      * @param {Object} object_info
      */
     return function(object_info) {
+        // fail if object is a temp file/part or a folder object or a versioned object
+        if (object_info.key.startsWith(config.NSFS_TEMP_DIR_NAME)) return false;
+        if (object_info.key.includes(config.NSFS_FOLDER_OBJECT_NAME)) return false;
+        if (object_info.key.includes('.versions/')) return false;
+
         if (params.filter?.prefix && !object_info.key.startsWith(params.filter.prefix)) return false;
         if (params.expiration && object_info.age < params.expiration) return false;
         if (params.filter?.tags && !file_contain_tags(object_info, params.filter.tags)) return false;
--- a/src/test/unit_tests/jest_tests/test_nc_lifecycle.test.js
+++ b/src/test/unit_tests/jest_tests/test_nc_lifecycle.test.js
@@ -92,6 +92,44 @@ describe('delete_multiple_objects + filter', () => {
             await assert_object_deletion_failed(delete_res);
         });
 
+        it('delete_multiple_objects - filter should fail on internal dir - empty prefix - versioning DISABLED bucket', async () => {
+            const object_upload_res = await nsfs.create_object_upload({ key: key, bucket: bucket_name }, dummy_object_sdk);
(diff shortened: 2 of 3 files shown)


src/util/lifecycle_utils.js
```diff
@@ -148,6 +148,11 @@ function build_lifecycle_filter(params) {
      * @param {Object} object_info
      */
     return function(object_info) {
+        // fail if object is a temp file/part or a folder object or a versioned object
```
nadavMiz, line 151: this logic is shared with list-object. and we would need to change in both cases in case there will be a new hidden file/folder added. maybe we should add a utility function to check if file is internal directory?

nadavMiz, line 151: i think it is a good check. but we should probably filter those files on the ilm layer if possible

romayalon: it's already being filtered in the ILM layer as mentioned in the description
see what I wrote in the description - 
```
we are not supposed to see an issue due to the whole flow but I added it in order to be extra safe on the filter func as well.

POSIX flow - we shouldn't see this issue at all because list_objects() is used and should not be return an internal file.
GPFS flow - also shouldn't happen because on the ILM policy we avoid internal directory/past versions and when seeing .folder file we convert the key to be the directory name.
```

romayalon: I added these code changed because of potential bugs in code like I saw here - https://github.com/noobaa/noobaa-core/pull/9069

romayalon: can you explain why would we need to change in 2 different places if this function is shared?

nadavMiz: they are both intended to ignore internal directories when going over files. if there is a new internal directory. we would need to ignore it in both

romayalon: indeed, since both places using this function why would we need to change in 2 different places? or I don't get your point...

nadavMiz: I mean list-object in general regardless of lifecycle:
```
if (!ent.name.startsWith(prefix_ent) ||
    ent.name < marker_curr ||
    ent.name === this.get_bucket_tmpdir_name() ||
    (ent.name === config.NSFS_FOLDER_OBJECT_NAME && is_disabled_dir_content) ||
    this._is_hidden_version_path(ent.name)) {
    return;
}

```

PR #9081, opened 2025-06-11
NC | Lifecycle | GPFS | Apply policy per mount point while catching and logging mount point errors on status

romayalon:
### Explain the Changes
1. Wrapped create_candidates_file_by_gpfs_ilm_policy() with try and catch that skips failures in case of one mount point has any issues.
2. Added status of the create_candidates_file_by_gpfs_ilm_policy() per mount point that includes times and errors.
3. Added config.NC_LIFECYCLE_GPFS_MMAPPLY_ILM_POLICY_CONCURRENCY in order to increase the concurrency of mmapplypolicy if we would like to do it in the future.

1. Fixed https://issues.redhat.com/browse/DFBUGS-2729

Gaps - 
I noticed we need to do a similar change for buckets and rules. 

### Testing Instructions:
1. Manual testing - mentioned in the bug.
2. Manual + artificial test - 
a. Add to line 1222 the following error throw - `if (mount_point === '/mnt/gpfs0') throw new Error('ROMY Mount point /mnt/gpfs0 is not supported for lifecycle ILM policies');`
b. create 2 buckets, one on /mnt/gpfs0/romy_storage1/ and one on /mnt/cesSharedRoot/romy_storage1/ and add files 10 days old under the directories.
```
noobaa-cli bucket add --name bucket.ces.mount --path /mnt/cesSharedRoot/romy_storage1/ --owner account1
noobaa-cli bucket add --name bucket.gpfs.mount --path /mnt/gpfs0/romy_storage1/ --owner account1
for i in {1..10} ; do touch -d  "10 days ago"  /mnt/cesSharedRoot/romy_storage1/file${i}.txt ; done
for i in {1..10} ; do touch -d  "10 days ago"  /mnt/gpfs0/romy_storage1/file${i}.txt ; done
```
c. Attach a lifecycle policy with filter/prefix: "" to both of the buckets.
```
cat ~/demo/policies/delete_all_policy.js
{
    "Rules": [
        {
            "Expiration":{ "Days": 1},
            "ID": "expire-data",
            "Filter":{ "Prefix": ""},
            "Status": "Enabled"
        }
    ]
}
s3-account-demo1 put-bucket-lifecycle-configuration --bucket bucket.ces.mount --lifecycle-configuration file://~/demo/policies/delete_all_policy.js
```
d. run lifecycle worker - `noobaa-cli lifecycle --disable_runtime_validation --debug 5 &> lifecycle.logs`
e. check that the objects under /mnt/cesSharedRoot/romy_storage1/ were deleted and under /mnt/gpfs0/romy_storage1/  the objects still exist.
check the logs of the latest run and see it was completed
```
ls -la  /mnt/cesSharedRoot/romy_storage1/  // -should be empty
ls -la /mnt/gpfs0/romy_storage1/ // should still contain all objects
cat lifecycle.logs |  less -R
```

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **New Features**
	- Added detailed status tracking and error reporting for lifecycle operations at the mount point level.
	- Introduced concurrent processing for candidate file creation by mount point, improving efficiency and visibility into individual mount point operations.

- **Chores**
	- Added a new configuration option to control concurrency for lifecycle policy application.


src/manage_nsfs/nc_lifecycle.js
```diff
@@ -807,21 +812,23 @@ class NCLifecycle {
      * } params
      * @returns {Void}
     */
-    update_status({ bucket_name, rule_id, op_name, op_times, reply = [], error = undefined }) {
+    update_status({ mount_point, bucket_name, rule_id, op_name, op_times, reply = [], error = undefined }) {
```
nadavMiz, line 815: not for this PR. but I think this function is getting messier with time. I think we should consider better Polymorphism. maybe even use objects?

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -1190,11 +1213,22 @@ class NCLifecycle {
...
+                await this._call_op_and_update_status({
+                    mount_point,
+                    op_name: TIMED_OPS.CREATE_GPFS_CANDIDATE_FILE_BY_ILM_POLICY,
+                    op_func: async () => {
+                        const ilm_policy_path = await this.write_tmp_ilm_policy(mount_point, policy);
+                        await this.create_candidates_file_by_gpfs_ilm_policy(mount_point, ilm_policy_path);
+                    }
+                });
+            } catch (err) {
+                dbg.error('create_candidates_file_by_gpfs_ilm_policy failed with error', err, err.code, err.message);
```
nadavMiz, line 1229: wouldn't we still try to parse the file in get_candidates? do we have something to protect us there?

romayalon: if the candidates file was not created we won't parse it, see here - https://github.com/noobaa/noobaa-core/blob/master/src/manage_nsfs/nc_lifecycle.js#L1429-L1434
the reading is wrapped with a try catch clause - 
```
 } catch (err) {
            if (err.code === 'ENOENT') {
                dbg.log2(`parse_candidates_from_gpfs_ilm_policy ilm_candidates_file_exists does not exist, no candidates to delete`);
                this._set_rule_state(bucket_json, lifecycle_rule, finished_state);
                return [];
            }
```

nadavMiz (approved):

PR #9095, opened 2025-06-12
Added metrices to show the replication status per bucket

aayushchouhan09:
### Describe the Problem
Previously, there was no metric available to show the replication progress for each individual bucket. As a result, users had no way to monitor how much data had been successfully replicated.

### Explain the Changes
1.  Added new metrices to expose replication per bucket: `bucket_last_cycle_total_objects_num`, `bucket_last_cycle_replicated_objects_num`,  `bucket_last_cycle_error_objects_num`

Screenshot from 2025-06-25 18-41-29

1.  Issue: https://issues.redhat.com/browse/RHSTOR-7208

### Testing Instructions:
1. To view the metrices:

> $ JWT_TOKEN=$(kubectl get secret/noobaa-metrics-auth-secret -o jsonpath={.data.metrics_token} | base64 -d)
$ kubectl exec -it noobaa-core-0 -- curl -k -H "Authorization: Bearer ${JWT_TOKEN}" http://localhost:8080/metrics/bg_workers

## Summary by CodeRabbit

- **New Features**
  - Enhanced replication metrics to include per-bucket counts for total objects scanned, replicated objects, and replication errors in the last cycle.
- **Improvements**
  - Replication status reporting now distinguishes between rule-level and bucket-level metrics for more detailed insights.
- **Style**
  - Minor formatting improvements for log readability.

### DIFF
--- a/src/server/analytic_services/prometheus_reports/noobaa_core_report.js
+++ b/src/server/analytic_services/prometheus_reports/noobaa_core_report.js
@@ -384,6 +384,27 @@ const NOOBAA_CORE_METRICS = js_utils.deep_freeze([{
         }
     }, {
         type: 'Gauge',
+        name: 'bucket_last_cycle_total_objects_num',
+        configuration: {
+            help: 'Total number of objects scanned per bucket in last replication cycle',
+            labelNames: ['bucket_name']
+        }
+    }, {
+        type: 'Gauge',
+        name: 'bucket_last_cycle_replicated_objects_num',
+        configuration: {
+            help: 'Number of objects replicated per bucket in last replication cycle',
+            labelNames: ['bucket_name']
+        }
+    }, {
+        type: 'Gauge',
+        name: 'bucket_last_cycle_error_objects_num',
+        configuration: {
+            help: 'Number of objects failed to replicate per bucket in last replication cycle',
+            labelNames: ['bucket_name']
+        }
+    }, {
+        type: 'Gauge',
         name: 'bucket_used_bytes',
         configuration: {
             help: 'Object Bucket Used Bytes',
@@ -606,9 +627,12 @@ class NooBaaCoreReport extends BasePrometheusReport {
     set_replication_status(repl_info) {
         if (!this._metrics) return;
         const replication_id = repl_info.replication_id;
+        const bucket_name = repl_info.bucket_name;
         delete this._metrics.replication_status.hashMap[String(repl_info.replication_id)];
         this._metrics.replication_status.set(_.omit(repl_info, ['last_cycle_writes_size',
-            'last_cycle_writes_num', 'last_cycle_error_writes_size', 'last_cycle_error_writes_num'
+            'last_cycle_writes_num', 'last_cycle_error_writes_size', 'last_cycle_error_writes_num',
+            'bucket_last_cycle_total_objects_num', 'bucket_last_cycle_replicated_objects_num',
+            'bucket_last_cycle_error_objects_num'
         ]), Date.now());
 
         delete this._metrics.replication_last_cycle_writes_size.hashMap[String(repl_info.replication_id)];
@@ -622,6 +646,15 @@ class NooBaaCoreReport extends BasePrometheusReport {
 
         delete this._metrics.replication_last_cycle_error_writes_num.hashMap[String(repl_info.replication_id)];
         this._metrics.replication_last_cycle_error_writes_num.set({ replication_id }, repl_info.last_cycle_error_writes_num);
+
+        delete this._metrics.bucket_last_cycle_total_objects_num.hashMap[String(bucket_name)];
+        this._metrics.bucket_last_cycle_total_objects_num.set({ bucket_name }, repl_info.bucket_last_cycle_total_objects_num);
+
+        delete this._metrics.bucket_last_cycle_replicated_objects_num.hashMap[String(bucket_name)];
+        this._metrics.bucket_last_cycle_replicated_objects_num.set({ bucket_name }, repl_info.bucket_last_cycle_replicated_objects_num);
+
+        delete this._metrics.bucket_last_cycle_error_objects_num.hashMap[String(bucket_name)];
+        this._metrics.bucket_last_cycle_error_objects_num.set({ bucket_name }, repl_info.bucket_last_cycle_error_objects_num);
     }
 }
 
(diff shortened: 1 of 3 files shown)


config.js
```diff
@@ -699,7 +699,7 @@ config.REPLICATION_SEMAPHORE_TIMEOUT = 300 * 1000;
 config.REPLICATION_ENABLED = true;
 config.LOG_REPLICATION_ENABLED = true;
 config.AWS_LOG_CANDIDATES_LIMIT = 10;
-config.BUCKET_LOG_REPLICATOR_DELAY = 5 * 60 * 1000;
+config.BUCKET_LOG_REPLICATOR_DELAY = 5 * 1000;
```
Neon-White: I don't think we're performant enough (or receive enough resources) to perform a scan every 5 seconds.
5 minutes may be too much, but in my opinion, five seconds is too little.

src/server/analytic_services/prometheus_reports/noobaa_core_report.js
```diff
@@ -623,6 +630,14 @@ class NooBaaCoreReport extends BasePrometheusReport {
         delete this._metrics.replication_last_cycle_error_writes_num.hashMap[String(repl_info.replication_id)];
         this._metrics.replication_last_cycle_error_writes_num.set({ replication_id }, repl_info.last_cycle_error_writes_num);
     }
+
+    update_bucket_replication_percentage(repl_info) {
+        if (!this._metrics) return;
```
Neon-White: In what case is `this._metrics` missing?
Regardless, please add a log line to communicate this state

src/server/analytic_services/prometheus_reports/noobaa_core_report.js
```diff
@@ -623,6 +630,14 @@ class NooBaaCoreReport extends BasePrometheusReport {
...
+        const { bucket_name, repl_percentage } = repl_info;
+
+        delete this._metrics.bucket_replication_percentage.hashMap[String(bucket_name)];
+        this._metrics.bucket_replication_percentage.set({bucket_name}, repl_percentage);
+    }
```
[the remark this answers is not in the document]
Neon-White: +1 to CodeRabbit

config.js
liranmauda: @achouhan09 I agree, revert this one.
(I guess it was for testing the change, but in that case we should keep a comment next to it to revert it, so we will not forget)

config.js
```diff
@@ -690,7 +690,7 @@ config.NS_MAX_ALLOWED_IO_ERRORS = 9;
 //      BUCKET REPLICATOR     //
 ////////////////////////////////
 
-config.BUCKET_REPLICATOR_DELAY = 5 * 60 * 1000;
+config.BUCKET_REPLICATOR_DELAY = 5 * 1000;
```
liranmauda: Revert this change

src/server/bg_services/replication_scanner.js
```diff
@@ -129,6 +129,11 @@ class ReplicationScanner {
                 const replication_status = replication_utils.get_rule_status(rule.rule_id, src_cont_token, keys_diff_map, copy_res);
 
                 replication_utils.update_replication_prom_report(src_bucket.name, replication_id, replication_status);
+
+                const repl_obj = Object.keys(keys_diff_map).length;
+                const total_obj = bucketDiff.get_objects_count(src_bucket.name, prefix);
+                const replication_percentage = replication_utils.get_replication_percentage(repl_obj, total_obj);
+                replication_utils.update_replication_prom_report_per_bucket(src_bucket.name, replication_percentage);
```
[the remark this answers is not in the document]
liranmauda: @achouhan09 I agree on that one. 
The cycle deliberately ignores new written files on a cycle where the continuation token is passed.
This means that if you have 10 files with names a-j you you replicated a-c and started a new cycle, if someone writes a file name a2, it will be ignored till the next cycle.
If the file is e2, then it will be caught by the next run on this cycle and get replicated.

As a result, we should count the number of files  in the beginning of this loop (right after `const bucketDiff = new BucketDiff` )  and count based on this the percentage, and in a case that it exceed the 100% report 100% 

We should also change the name of this metric to reflect that it is reporting the replication percentage in a certain cycle (maybe under the help: `help: 'Replication percentage per bucket in the current cycle',` )

src/server/analytic_services/prometheus_reports/noobaa_core_report.js
```diff
@@ -343,6 +343,13 @@ const NOOBAA_CORE_METRICS = js_utils.deep_freeze([{
         configuration: {
             help: 'Health status'
         }
+    }, {
+        type: 'Gauge',
+        name: 'bucket_replication_percentage',
+        configuration: {
+            help: 'Replication percentage per bucket',
```
liranmauda: It should reflect the fact that this percentage is true only to the current cycle
```suggestion
            help: 'Replication percentage per bucket in the current cycle',
```

src/server/utils/bucket_diff.js
```diff
@@ -233,6 +233,37 @@ class BucketDiff {
...
     }
 
+    /**
+     * @param {string} bucket
+     * @param {string} prefix
+     * @param {number} max_keys
+     *
+     * get_objects_count will return the total number of objects present in the specified bucket and prefix
+     */
+    async get_objects_count(bucket, prefix, max_keys = 1000) {
```
liranmauda: Why is it here in the bucket diff? 
Why do we have max keys and iterating on the list? 
This should run as little as possible to avoid egress

liranmauda (changes requested):

src/server/analytic_services/prometheus_reports/noobaa_core_report.js
aayushchouhan09: we are doing a null check here for `this._metrics`  if  `config.PROMETHEUS_ENABLED` is false we are retuning and not updating the metrics.

src/server/utils/bucket_diff.js
aayushchouhan09: We are using already defined functions in bucket diff which are designed to handle the versioning and deletions. Yes, we can remove the max_keys.

src/server/bg_services/replication_scanner.js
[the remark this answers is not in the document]
aayushchouhan09: Done, Thanks

config.js
aayushchouhan09: Done, Thanks

achouhan09:
https://github.com/coderabbitai

liranmauda:
@coderabbitai review

liranmauda (approved):

PR #9096, opened 2025-06-12
system store - reduce load time (part of performance effort 4.20)

alphaprinz:
### Describe the Problem
Reduce system_store.load() cpu load.

### Explain the Changes
1. REMOVED - Introduce "fresh enough" - if load was performed recently, don't load again.
2. Perform db queries concurrently (instead of sequentially).
3. REMOVED - Reduce sempahore's block
4. Remove publish changes from hb and md_aggregator.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

Here are the updated release notes:

- **New Features**
  - Added a new configuration option to control concurrency during database data loading.
  - Introduced a new API endpoint to retrieve system store data.
  - Enabled selection between database and core system sources for system data loading.
- **Improvements**
  - Optimized data loading by limiting parallel database reads based on configurable concurrency.
  - Enhanced system store refresh with detailed debug logging and conditional cluster-wide notifications.
  - Improved control over system updates by allowing suppression of cluster-wide notifications.

### DIFF
--- a/config.js
+++ b/config.js
@@ -235,8 +235,8 @@ config.ROOT_KEY_MOUNT = '/etc/noobaa-server/root_keys';
 
 config.DB_TYPE = /** @type {nb.DBType} */ (process.env.DB_TYPE || 'postgres');
 
-config.POSTGRES_DEFAULT_MAX_CLIENTS = 10;
-config.POSTGRES_MD_MAX_CLIENTS = (process.env.LOCAL_MD_SERVER === 'true') ? 70 : 10;
+config.POSTGRES_DEFAULT_MAX_CLIENTS = 5;
+config.POSTGRES_MD_MAX_CLIENTS = (process.env.LOCAL_MD_SERVER === 'true') ? 70 : 5;
 
 ///////////////////
 // SYSTEM CONFIG //
--- a/src/api/system_api.js
+++ b/src/api/system_api.js
@@ -459,6 +459,13 @@ module.exports = {
             auth: {
                 system: 'admin'
             }
+        },
+
+        get_system_store: {
+            method: 'GET',
+            auth: {
+                system: false
+            }
         }
     },
 
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -394,17 +394,21 @@ class SystemStore extends EventEmitter {
         if (this.data) {
             load_time = this.data.time;
         }
+        let res;
         const since_load = Date.now() - load_time;
         if (since_load < this.START_REFRESH_THRESHOLD) {
-            return this.data;
+            res = this.data;
         } else if (since_load < this.FORCE_REFRESH_THRESHOLD) {
             dbg.warn(`system_store.refresh: system_store.data.time > START_REFRESH_THRESHOLD, since_load = ${since_load}, START_REFRESH_THRESHOLD = ${this.START_REFRESH_THRESHOLD}`);
             this.load().catch(_.noop);
-            return this.data;
+            res = this.data;
         } else {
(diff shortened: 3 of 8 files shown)


src/server/system_services/system_store.js
```diff
@@ -407,30 +408,33 @@ class SystemStore extends EventEmitter {
     }
 
     async load(since) {
+        const new_data = new SystemStoreData();
```
dannyzaken: Is the purpose of it to save operations under lock? I think it can mess with the `.time` property of `SystemStoreData`, and might lead to unknown behavior. not sure if it's worth the savings.

config.js
```diff
@@ -250,6 +250,7 @@ config.INTERNAL_STORAGE_POOL_NAME = 'system-internal-storage-pool';
 // config.SPILLOVER_TIER_NAME = 'bucket-spillover-tier';
 config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
 config.BUCKET_AUTOCONF_TIER2_ENABLED = false;
+config.READ_NEW_DATA_CONCURRENCY = parseInt(process.env.READ_NEW_DATA_CONCURRENCY, 10) || 5;
```
dannyzaken: The name is a bit generic in my opinion, it's not clear what it does. Maybe it should be something like `SYSTEM_STORE_LOAD_CONCURRENT_COLLECTION` or just `SYSTEM_STORE_LOAD_CONCURRENCY`

src/server/system_services/system_store.js
```diff
@@ -393,44 +394,51 @@ class SystemStore extends EventEmitter {
...
             this.load().catch(_.noop);
-            return this.data;
+            res = this.data;
         } else {
             dbg.warn(`system_store.refresh: system_store.data.time > FORCE_REFRESH_THRESHOLD, since_load = ${since_load}, FORCE_REFRESH_THRESHOLD = ${this.FORCE_REFRESH_THRESHOLD}`);
-            return this.load();
+            res = this.load();
         }
+        //call refresh periodically
+        P.delay_unblocking(this.START_REFRESH_THRESHOLD).then(this.refresh);
```
dannyzaken: this can cause a lot of `refresh()` calls to be queued.  I believe that `refresh()` is called on every RPC request as a middleware (see [here](https://github.com/alphaprinz/noobaa-core/blob/36b98920c74a32b42ac35e4f55579656abcbed2a/src/server/server_rpc.js#L32)), so I don't think we should schedule another refresh.

config.js
```diff
@@ -236,8 +236,8 @@ config.ROOT_KEY_MOUNT = '/etc/noobaa-server/root_keys';
 
 config.DB_TYPE = /** @type {nb.DBType} */ (process.env.DB_TYPE || 'postgres');
 
-config.POSTGRES_DEFAULT_MAX_CLIENTS = 10;
-config.POSTGRES_MD_MAX_CLIENTS = (process.env.LOCAL_MD_SERVER === 'true') ? 70 : 10;
+config.POSTGRES_DEFAULT_MAX_CLIENTS = 5;
+config.POSTGRES_MD_MAX_CLIENTS = (process.env.LOCAL_MD_SERVER === 'true') ? 70 : 5;
```
dannyzaken, line 239: Why are you changing this?

src/api/system_api.js
```diff
@@ -459,6 +459,27 @@ module.exports = {
...
+        },
+
+        get_system_store: {
+            method: 'GET',
+            /*params: {
+                type: 'object'
+            },*/
+            reply: {
+                type: 'object',
+                /*properites: {
```
dannyzaken: we should myabe use `RPC_BUFFERS` here. see example in block_store_api

PR #9100, opened 2025-06-15
NC | Lifecycle | GPFS | Skip bucket processing if mount point has errors and add to the bucket status

romayalon:
### Describe the Problem
Currently, mount point errors are externalized on the mount point status, it could benefit to have this also on the bucket status level.

### Explain the Changes
1. In process_bucket() - in GPFS flow -  if bucket's mount point has any errors it will skip its processing and add throw an error for the bucket.

1. Fixed https://issues.redhat.com/browse/DFBUGS-2729

### Testing Instructions:

1. Manual testing - mentioned in the bug.
2. Manual + artificial test - 
a. Add to line 1222 the following error throw - `if (mount_point === '/mnt/gpfs0') throw new Error('ROMY Mount point /mnt/gpfs0 is not supported for lifecycle ILM policies');`
b. create 2 buckets, one on /mnt/gpfs0/romy_storage1/ and one on /mnt/cesSharedRoot/romy_storage1/ and add files 10 days old under the directories.
```
noobaa-cli bucket add --name bucket.ces.mount --path /mnt/cesSharedRoot/romy_storage1/ --owner account1
noobaa-cli bucket add --name bucket.gpfs.mount --path /mnt/gpfs0/romy_storage1/ --owner account1
for i in {1..10} ; do touch -d  "10 days ago"  /mnt/cesSharedRoot/romy_storage1/file${i}.txt ; done
for i in {1..10} ; do touch -d  "10 days ago"  /mnt/gpfs0/romy_storage1/file${i}.txt ; done
```
c. Attach a lifecycle policy with filter/prefix: "" to both of the buckets.
```
cat ~/demo/policies/delete_all_policy.js
{
    "Rules": [
        {
            "Expiration":{ "Days": 1},
            "ID": "expire-data",
            "Filter":{ "Prefix": ""},
            "Status": "Enabled"
        }
    ]
}
s3-account-demo1 put-bucket-lifecycle-configuration --bucket bucket.ces.mount --lifecycle-configuration file://~/demo/policies/delete_all_policy.js
```
d. run lifecycle worker - `noobaa-cli lifecycle --disable_runtime_validation --debug 5 &> lifecycle.logs`
e. check that the objects under /mnt/cesSharedRoot/romy_storage1/ were deleted and under /mnt/gpfs0/romy_storage1/  the objects still exist.
check the logs of the latest run and see it was completed 
```
ls -la  /mnt/cesSharedRoot/romy_storage1/  // -should be empty
ls -la /mnt/gpfs0/romy_storage1/ // should still contain all objects
cat lifecycle.logs |  less -R
```

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **New Features**
  - Added an early error check to prevent processing of buckets with mount point errors when GPFS optimization is enabled.

- **Bug Fixes**
  - Improved reliability by ensuring lifecycle processing is aborted for buckets with problematic mount points.
  - Enhanced error handling to prevent retries on buckets that encounter processing errors.

### DIFF
--- a/src/manage_nsfs/nc_lifecycle.js
+++ b/src/manage_nsfs/nc_lifecycle.js
@@ -90,6 +90,9 @@ class NCLifecycle {
         this.disable_service_validation = options.disable_service_validation || false;
         this.disable_runtime_validation = options.disable_runtime_validation || false;
         this.should_continue_last_run = options.should_continue_last_run || false;
+        // used for GPFS optimization - maps bucket names to their mount points
+        // example - { 'bucket1': '/gpfs/mount/point1', 'bucket2': '/gpfs/mount/point2' }
+        this.bucket_to_mount_point_map = {};
     }
 
     /**
@@ -183,13 +186,19 @@ class NCLifecycle {
         }
 
         while (!this.lifecycle_run_status.state.is_finished) {
-            await P.map_with_concurrency(buckets_concurrency, bucket_names, async bucket_name =>
-                await this._call_op_and_update_status({
-                    bucket_name,
-                    op_name: TIMED_OPS.PROCESS_BUCKET,
-                    op_func: async () => this.process_bucket(bucket_name, system_json)
-                })
-            );
+            await P.map_with_concurrency(buckets_concurrency, bucket_names, async bucket_name => {
+                try {
+                    await this._call_op_and_update_status({
+                        bucket_name,
+                        op_name: TIMED_OPS.PROCESS_BUCKET,
+                        op_func: async () => this.process_bucket(bucket_name, system_json)
+                    });
+                } catch (err) {
+                    dbg.error('process_bucket failed with error', err, err.code, err.message);
+                    if (!this.lifecycle_run_status.buckets_statuses[bucket_name]) this.init_bucket_status(bucket_name);
+                    this.lifecycle_run_status.buckets_statuses[bucket_name].state.is_finished = true;
+                }
+            });
             this.lifecycle_run_status.state.is_finished = Object.values(this.lifecycle_run_status.buckets_statuses).reduce(
                 (acc, bucket) => acc && (bucket.state?.is_finished),
                 true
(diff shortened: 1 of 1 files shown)


src/manage_nsfs/nc_lifecycle.js
```diff
@@ -204,6 +213,12 @@ class NCLifecycle {
      */
     async process_bucket(bucket_name, system_json) {
         const bucket_json = await this.config_fs.get_bucket_by_name(bucket_name, config_fs_options);
+        if (this._should_use_gpfs_optimization()) {
```
nadavMiz: I don't really like this. it breaks the general flow of the bucket processing unnecessarily. from what I can see we check the bucket status inside process_rules, that means that if there is a mount error we will keep throwing for each loop inside process_buckets. also if we already know the status has errors, why wouldn't we just  set the bucket status to is_finished there. we can add a check for the bucket status in the beginning of the process_bucket function to save the operations we do here before checking for the status in process_rules

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -183,13 +186,19 @@ class NCLifecycle {
...
+            await P.map_with_concurrency(buckets_concurrency, bucket_names, async bucket_name => {
+                try {
+                    await this._call_op_and_update_status({
+                        bucket_name,
+                        op_name: TIMED_OPS.PROCESS_BUCKET,
+                        op_func: async () => this.process_bucket(bucket_name, system_json)
+                    });
+                } catch (err) {
+                    dbg.error('process_bucket failed with error', err, err.code, err.message);
+                    if (!this.lifecycle_run_status.buckets_statuses[bucket_name]) this.init_bucket_status(bucket_name);
```
nadavMiz, line 198: I think that for readability we should have a get function for the state. its a very long line we need to do in a lot of places. its a bit nitpicking and we can also do it in a different PR

romayalon: added them to https://github.com/noobaa/noobaa-core/issues/8861 so we will address it not in the backport timeline.

nadavMiz (approved):
LGTM

PR #9135, opened 2025-07-08
Removed counters queries from md_store find_objects

dannyzaken:
### Describe the Problem
- `md_store.find_objects` performed additional count queries to support the old ui `list_objects_admin` function
- These queries can be very expensive for certain operations, specifically in lifecycle and obc deletion flows.

### Explain the Changes
- Removed the counters that are returned from find_objects and are not used anywhere in the code.

1.  https://issues.redhat.com/browse/DFBUGS-2232

### Testing Instructions:
1. Create an OBC bucket and upload millions of objects
2. Delete the OBC to trigger `delete_multiple_objects_unordered` that calls `find_objects`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed display of object counters in the admin object list view.

* **Refactor**
  * Simplified object listing to show only the list of objects without additional counters.
  * Minor formatting and code cleanup for improved readability.

### DIFF
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -901,25 +901,6 @@ module.exports = {
                 type: 'object',
                 required: ['objects'],
                 properties: {
-                    counters: {
-                        type: 'object',
-                        properties: {
-                            by_mode: {
-                                type: 'object',
-                                properties: {
-                                    completed: {
-                                        type: 'integer'
-                                    },
-                                    uploading: {
-                                        type: 'integer'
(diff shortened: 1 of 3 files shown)


tangledbytes (approved):
LGTM.

PS: The PR description made me wonder why we first do `find_objects` and then perform the updates. Maybe we could have just gotten away with one DB call (especially if we are writing SQL).

PR #9142, opened 2025-07-14
[NSFS | GLACIER] Support newline and backslash character in Glacier logs

tangledbytes:
### Describe the Problem
NooBaa Glacier cannot work with object keys which have newline characters in their name.

### Explain the Changes
This PR adds/changes the following:
1. Adds 2 new methods to the base Glacier class `encode_log` and `decode_log`. The idea is to be able to encode a log before writing it to the disk and decode a log after reading it from the disk. This allows to encode special characters like `\n`.

1. 

### Testing Instructions:

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of special characters (including newlines) in object keys and log entries across migration and restore flows by adding transparent encoding/decoding for logging, WAL entries, and failure recording — increases robustness of migrate/restore operations.

* **Tests**
  * Added unit tests covering objects with special-character keys to validate successful and failed restore/migration paths and expiry behavior.

### DIFF
--- a/src/sdk/glacier_tapecloud.js
+++ b/src/sdk/glacier_tapecloud.js
@@ -191,6 +191,8 @@ class TapeCloudUtils {
 }
 
 class TapeCloudGlacier extends Glacier {
+    static LOG_DELIM = ' -- ';
+
     /**
      * @param {nb.NativeFSContext} fs_context
      * @param {LogFile} log_file
@@ -200,8 +202,14 @@ class TapeCloudGlacier extends Glacier {
     async stage_migrate(fs_context, log_file, failure_recorder) {
         dbg.log2('TapeCloudGlacier.stage_migrate starting for', log_file.log_path);
 
+        // Wrap failure recorder to make sure we correctly encode the entries
+        // before appending them to the failure log
+        const encoded_failure_recorder = async failure => failure_recorder(this.encode_log(failure));
+
         try {
             await log_file.collect(Glacier.MIGRATE_STAGE_WAL_NAME, async (entry, batch_recorder) => {
+                entry = this.decode_log(entry);
+
                 let entry_fh;
                 let should_migrate = true;
                 try {
@@ -230,7 +238,7 @@ class TapeCloudGlacier extends Glacier {
                     // Can't really do anything if this fails - provider
                     // needs to make sure that appropriate error handling
                     // is being done there
-                    await failure_recorder(entry);
+                    await encoded_failure_recorder(entry);
                     return;
                 }
 
@@ -240,14 +248,14 @@ class TapeCloudGlacier extends Glacier {
                 // Mark the file staged
                 try {
                     await entry_fh.replacexattr(fs_context, { [Glacier.XATTR_STAGE_MIGRATE]: Date.now().toString() });
-                    await batch_recorder(entry);
+                    await batch_recorder(this.encode_log(entry));
                 } catch (error) {
                     dbg.error('failed to mark the entry migrate staged', error);
 
                     // Can't really do anything if this fails - provider
                     // needs to make sure that appropriate error handling
                     // is being done there
-                    await failure_recorder(entry);
+                    await encoded_failure_recorder(entry);
                 } finally {
                     entry_fh?.close(fs_context);
                 }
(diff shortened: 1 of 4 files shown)


guymguym:
@tangledbytes few comments:

What if the filename contains this new compound delimiter?
How do we send paths with newlines to eeadm - does it use the same delimiter?

I feel like this solution is a bit random, and I might consider using encodeURI on the paths and then decodeURI when reading it:

```
> encodeURI('/path/to/file with spaces and percents %%%')
'/path/to/file%20with%20spaces%20and%20percents%20%25%25%25'
```

Another option is to use null character '\0' as a line terminator, which is very stable, though it makes the file a bit less readable.

tangledbytes:
> @tangledbytes few comments:
> 
> What if the filename contains this new compound delimiter? How do we send paths with newlines to eeadm - does it use the same delimiter?
> 
> I feel like this solution is a bit random, and I might consider using encodeURI on the paths and then decodeURI when reading it:
> 
> ```
> > encodeURI('/path/to/file with spaces and percents %%%')
> '/path/to/file%20with%20spaces%20and%20percents%20%25%25%25'
> ```
> 
> Another option is to use null character '\0' as a line terminator, which is very stable, though it makes the file a bit less readable.

@guymguym, just want to mention that I am still working on this PR, it is not clean yet.

Regarding the solution, I agree it is a little strange however it was chose not only because our reader suffers from newline delimiter problem but the same problem is with eeadm as well. It cannot handle file with '\n' in path names. If we don want it to support that, we need to use `\n-- ` as delimiter and only then it can handle '\n' characters in the path names. Now I am not sure what it does in case this compound delim is present in the path name.

src/sdk/glacier_tapecloud.js
```diff
@@ -425,6 +453,25 @@ class TapeCloudGlacier extends Glacier {
         return result.toLowerCase().trim() === 'true';
     }
 
+    /**
+     * 
+     * @param {string} data 
+     * @returns
```
jackyalbo: Can you add an explanation of what the line will look like after the encoding? Maybe also add an example.

src/sdk/glacier_tapecloud.js
```diff
@@ -195,6 +195,8 @@ class TapeCloudUtils {
 }
 
 class TapeCloudGlacier extends Glacier {
+    static LOG_DELIM = ' -- ';
```
jackyalbo, line 194: Will we have an issue if we have this delimiter inside the entry itself?

src/sdk/glacier_tapecloud.js
tangledbytes: Added an example and some description to the method.

src/sdk/glacier_tapecloud.js
tangledbytes: I think we won't have issue on the NooBaa side because the reader delimiter is still `\n` however `eeadm` doesn't really support this delimiter within the path either so I think there we would encounter some issues.

jackyalbo (approved):

src/sdk/glacier_tapecloud.js
```diff
@@ -240,14 +248,14 @@ class TapeCloudGlacier extends Glacier {
                 // Mark the file staged
                 try {
                     await entry_fh.replacexattr(fs_context, { [Glacier.XATTR_STAGE_MIGRATE]: Date.now().toString() });
-                    await batch_recorder(entry);
+                    await batch_recorder(this.encode_log(entry));
                 } catch (error) {
```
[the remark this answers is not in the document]
tangledbytes: The encoding is inline with what eeadm expects as well. This is OK and intentional.

src/sdk/glacier_tapecloud.js
```diff
@@ -379,25 +400,32 @@ class TapeCloudGlacier extends Glacier {
...
+
         try {
             const success = await this._recall(
                 log_file.log_path,
                 async entry_path => {
+                    entry_path = this.decode_log(entry_path);
                     dbg.log2('TapeCloudGlacier.restore.partial_failure - entry:', entry_path);
-                    await failure_recorder(entry_path);
+                    await encoded_failure_recorder(entry_path);
                 },
```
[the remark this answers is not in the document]
tangledbytes: It should not be raw, it must be escaped as well.

src/sdk/glacier_tapecloud.js
```diff
@@ -421,6 +449,29 @@ class TapeCloudGlacier extends Glacier {
...
+    }
+
+    /**
+     * 
+     * @param {string} data 
+     * @returns {string}
+     */
+    decode_log(data) {
+        return data.substring(TapeCloudGlacier.LOG_DELIM.length).replace(/\\n/g, '\n').replace(/\\\\/g, '\\');
+    }
```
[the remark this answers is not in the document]
tangledbytes: Good suggestion. Done.

PR #9148, opened 2025-07-16
System store phase1

alphaprinz:
### Describe the Problem
Reduce system store load time in order to prevent long queries and improve performance.

### Explain the Changes
1. Remove system store load trigger from hb.
2. Don't eval log params if log is below log level.
3. Add concurrency to db queries.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added a configuration option to control concurrency for system data loading.

* **Improvements**
  * Enhanced debug logging for system data loading with conditional detail based on log level.
  * Optimized internal update handling to optionally suppress cluster-wide reload notifications.

### DIFF
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -351,6 +351,7 @@ class SystemStore extends EventEmitter {
         this.is_finished_initial_load = false;
         this.START_REFRESH_THRESHOLD = 10 * 60 * 1000;
         this.FORCE_REFRESH_THRESHOLD = 60 * 60 * 1000;
+        this.SYSTEM_STORE_LOAD_CONCURRENCY = config.SYSTEM_STORE_LOAD_CONCURRENCY || 5;
         this._load_serial = new semaphore.Semaphore(1, { warning_timeout: this.START_REFRESH_THRESHOLD });
         for (const col of COLLECTIONS) {
             db_client.instance().define_collection(col);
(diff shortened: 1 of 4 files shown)


dannyzaken (approved):

src/server/system_services/system_store.js
```diff
@@ -351,6 +351,7 @@ class SystemStore extends EventEmitter {
         this.is_finished_initial_load = false;
         this.START_REFRESH_THRESHOLD = 10 * 60 * 1000;
         this.FORCE_REFRESH_THRESHOLD = 60 * 60 * 1000;
+        this.SYSTEM_STORE_LOAD_CONCURRENCY = config.SYSTEM_STORE_LOAD_CONCURRENCY || 5;
```
liranmauda, line 354: Why do we need this default if we have it in the config.js ?

src/server/system_services/system_store.js
```diff
@@ -411,26 +412,29 @@ class SystemStore extends EventEmitter {
...
                 await this._register_for_changes();
                 await this._read_new_data_from_db(new_data);
                 const secret = await os_utils.read_server_secret();
                 this._server_secret = secret;
-                dbg.log1('SystemStore: fetch took', time_utils.millitook(millistamp));
-                dbg.log1('SystemStore: fetch size', size_utils.human_size(JSON.stringify(new_data).length));
-                dbg.log1('SystemStore: fetch data', util.inspect(new_data, {
-                    depth: 4
-                }));
+                if (dbg.should_log(1)) { //param should match below logs' level
```
liranmauda, line 431: Why do we need this here? 
I think the purpose of this was not to log if the param is lower, but it is internally used in the debug_mudule.js
We can also use it when we have a logic that is strictly for the logs, and we do not want to do it if the param is lower, but here I am not convinced that it is the case.

src/server/system_services/system_store.js
alphaprinz: Just in case.

src/server/system_services/system_store.js
alphaprinz: The purpose is to save cpu cost of evaluation of the log params (eg inspect()).

PR #9150, opened 2025-07-20
system store - endpoints load from core instead of db

alphaprinz:
### Describe the Problem
Reduce db load spikes when all endpoints load from db simultaneously by having only core load from db and endpoints load from core instead.

### Explain the Changes
1. Introduce new RPC that allows endpoints to get system store from core.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Public GET endpoint to retrieve the system store without system-level auth and to obtain the latest in-memory DB snapshot.

* **Configuration**
  * New SYSTEM_STORE_SOURCE setting (defaults to "DB") to choose data source (DB or CORE).

* **Behavior Changes**
  * System store can load from CORE with automatic fallback to DB and a two-phase publish flow; load accepts an explicit load_source (DB or CORE).

* **Tests**
  * Integration test added to verify CORE vs DB load parity.

### DIFF
--- a/config.js
+++ b/config.js
@@ -255,7 +255,9 @@ config.INTERNAL_STORAGE_POOL_NAME = 'system-internal-storage-pool';
 config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
 config.BUCKET_AUTOCONF_TIER2_ENABLED = false;
 config.SYSTEM_STORE_LOAD_CONCURRENCY = parseInt(process.env.SYSTEM_STORE_LOAD_CONCURRENCY, 10) || 5;
-config.SYSTEM_STORE_SOURCE = process.env.SYSTEM_STORE_SOURCE || "DB";
+// SYSTEM_STORE_SOURCE determines the preffered source for loading system_store data
+// This can be either "DB" to load from the DB or "CORE" to load from the system_server in noobaa-core
+config.SYSTEM_STORE_SOURCE = process.env.SYSTEM_STORE_SOURCE?.toUpperCase() || "DB";
 //////////////////////////
 // MD AGGREGATOR CONFIG //
 //////////////////////////
--- a/config.js
+++ b/config.js
@@ -255,7 +255,7 @@ config.INTERNAL_STORAGE_POOL_NAME = 'system-internal-storage-pool';
 config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
 config.BUCKET_AUTOCONF_TIER2_ENABLED = false;
 config.SYSTEM_STORE_LOAD_CONCURRENCY = parseInt(process.env.SYSTEM_STORE_LOAD_CONCURRENCY, 10) || 5;
-
+config.SYSTEM_STORE_SOURCE = process.env.SYSTEM_STORE_SOURCE || "db";
 //////////////////////////
 // MD AGGREGATOR CONFIG //
 //////////////////////////
--- a/config.js
+++ b/config.js
@@ -255,7 +255,7 @@ config.INTERNAL_STORAGE_POOL_NAME = 'system-internal-storage-pool';
 config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
 config.BUCKET_AUTOCONF_TIER2_ENABLED = false;
 config.SYSTEM_STORE_LOAD_CONCURRENCY = parseInt(process.env.SYSTEM_STORE_LOAD_CONCURRENCY, 10) || 5;
-config.SYSTEM_STORE_SOURCE = process.env.SYSTEM_STORE_SOURCE || "db";
+config.SYSTEM_STORE_SOURCE = process.env.SYSTEM_STORE_SOURCE || "DB";
 //////////////////////////
 // MD AGGREGATOR CONFIG //
 //////////////////////////
--- a/src/api/server_inter_process_api.js
+++ b/src/api/server_inter_process_api.js
@@ -19,7 +19,8 @@ module.exports = {
                 properties: {
                     since: { idate: true },
                     load_from_core_step: {
-                        type: 'string'
+                        type: 'string',
+                        enum: ['CORE', 'ENDPOINT']
                     }
                 }
             },
--- a/src/api/server_inter_process_api.js
+++ b/src/api/server_inter_process_api.js
@@ -17,7 +17,10 @@ module.exports = {
             params: {
                 type: 'object',
                 properties: {
-                    since: { idate: true }
+                    since: { idate: true },
+                    load_from_core_step: {
+                        type: 'string'
+                    }
                 }
             },
             auth: {
--- a/src/api/server_inter_process_api.js
+++ b/src/api/server_inter_process_api.js
@@ -20,7 +20,7 @@ module.exports = {
                     since: { idate: true },
                     load_from_core_step: {
                         type: 'string',
-                        enum: ['CORE', 'ENDPOINT']
+                        enum: ['DB', 'CORE']
                     }
                 }
             },
--- a/src/api/server_inter_process_api.js
+++ b/src/api/server_inter_process_api.js
@@ -18,7 +18,7 @@ module.exports = {
                 type: 'object',
                 properties: {
                     since: { idate: true },
-                    load_from_core_step: {
+                    load_source: {
                         type: 'string',
                         enum: ['DB', 'CORE']
                     }
--- a/src/api/system_api.js
+++ b/src/api/system_api.js
@@ -459,6 +459,13 @@ module.exports = {
             auth: {
                 system: 'admin'
             }
+        },
+
+        get_system_store: {
+            method: 'GET',
+            auth: {
+                system: false
+            }
         }
     },
 
--- a/src/api/system_api.js
+++ b/src/api/system_api.js
@@ -463,6 +463,12 @@ module.exports = {
 
         get_system_store: {
             method: 'GET',
+            reply: {
+                type: 'object',
+                properties: {
+                    // [RPC_BUFFERS].data
+                },
+            },
             auth: {
                 system: false
             }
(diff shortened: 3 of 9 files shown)


src/api/system_api.js
```diff
@@ -459,6 +459,13 @@ module.exports = {
             auth: {
                 system: 'admin'
             }
+        },
+
+        get_system_store: {
+            method: 'GET',
+            auth: {
```
dannyzaken, line 472: Better to return the `RPC_BUFFERS` under a reply object, like in [block_store_api](https://github.com/alphaprinz/noobaa-core/blob/766302a61a550aef4b63b4a02da18ed10d3ec711/src/api/block_store_api.js#L191-L201) for example.


```suggestion
            reply: {
                type: 'object',
                properties: {
                    // [RPC_BUFFERS].data
                },
            },
            auth: {
```

src/endpoint/endpoint.js
```diff
@@ -171,7 +171,7 @@ async function main(options = {}) {
                 // Load a system store instance for the current process and register for changes.
                 // We do not wait for it in becasue the result or errors are not relevent at
                 // this point (and should not kill the process);
-                system_store.get_instance().load();
+                system_store.get_instance({source: system_store.SOURCE.CORE}).load();
```
dannyzaken: Maybe I missed something, but I think the passed `source` is not used in get_instance

src/server/system_services/system_server.js
```diff
@@ -303,6 +303,11 @@ function get_system_status(req) {
     };
 }
 
+async function get_system_store() {
+    return {
```
dannyzaken: add try-catch to log an error on the core side

src/server/system_services/system_store.js
```diff
@@ -352,6 +357,8 @@ class SystemStore extends EventEmitter {
         this.START_REFRESH_THRESHOLD = 10 * 60 * 1000;
         this.FORCE_REFRESH_THRESHOLD = 60 * 60 * 1000;
         this.SYSTEM_STORE_LOAD_CONCURRENCY = config.SYSTEM_STORE_LOAD_CONCURRENCY || 5;
+        this.source = (process.env.HOSTNAME && process.env.HOSTNAME.indexOf("endpoint") === -1) ? SOURCE.DB : SOURCE.CORE;
```
dannyzaken: It's better to pass the source from outside rather than infer it inside the constructor. It will give us better control later if needed. 
Another option is to add a const in config.js, which is controlled by an env var. It will allow us to tweak it if any bugs are found in the field.

src/server/system_services/system_store.js
```diff
@@ -414,18 +421,23 @@ class SystemStore extends EventEmitter {
...
                 this.master_key_manager.load_root_key();
+                const new_data = new SystemStoreData();
                 let millistamp = time_utils.millistamp();
                 await this._register_for_changes();
-                await this._read_new_data_from_db(new_data);
+                if (this.source === SOURCE.DB) {
+                    await this._read_new_data_from_db(new_data);
+                } else {
+                    this.data = new SystemStoreData();
+                    await this._read_new_data_from_core(this.data);
```
dannyzaken: handle errors and maybe fallback to read_from_db (?)

src/api/server_inter_process_api.js
```diff
@@ -17,7 +17,10 @@ module.exports = {
             params: {
                 type: 'object',
                 properties: {
-                    since: { idate: true }
+                    since: { idate: true },
+                    load_from_core_step: {
+                        type: 'string'
+                    }
```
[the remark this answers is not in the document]
dannyzaken: @alphaprinz, I agree with the rabbit, better to add an enum. Also use upper case for convention
Also, I think the meaning of the flag should be different. instead `load_core_step` which is `core` or `endpoint`, it should be `load_source` that can be `core` or `db`


```suggestion
                    since: { idate: true },
                    load_source: {
                        type: 'string'
                        enum: ['CORE', 'DB']
                    }
```

src/server/system_services/system_store.js
```diff
@@ -352,6 +357,10 @@ class SystemStore extends EventEmitter {
         this.START_REFRESH_THRESHOLD = 10 * 60 * 1000;
         this.FORCE_REFRESH_THRESHOLD = 60 * 60 * 1000;
         this.SYSTEM_STORE_LOAD_CONCURRENCY = config.SYSTEM_STORE_LOAD_CONCURRENCY || 5;
+        //load from core if enabled and this is an endpoint
+        const is_endpoint = process.env.HOSTNAME && process.env.HOSTNAME.indexOf("endpoint") !== -1;
```
dannyzaken: why do we need this? I think we should only look at `config.SYSTEM_STORE_SOURCE`, regardless of the hostname. we don't want to restrict it to endpoints, as we might have other scale-out services outside of core that we might want to load from core (e.g.: some bg_workers, etc.)

src/server/system_services/system_store.js
```diff
@@ -616,13 +668,28 @@ class SystemStore extends EventEmitter {
...
                     target: '',
-                    request_params: { since: last_update }
+                    request_params: { since: last_update, load_from_core_step: 'core' }
                 });
+
+                //if endpoints are loading system store from core, we need to wait until
+                //above publish_to_cluster() will update core's in-memory db.
+                //the next publist_to_cluster() will make endpoints load the updated
+                //system store from core
+                if (config.SYSTEM_STORE_SOURCE.toLowerCase() === 'core') {
```
dannyzaken: the second phase should not be conditional. e.g., we can have make_changes in the core, where `config.SYSTEM_STORE_SOURCE == 'db'`. we want to notify on this change to any process that should load from core.

src/server/common_services/server_inter_process.js
```diff
@@ -13,11 +13,20 @@ const dbg = require('../../util/debug_module')(__filename);
...
 async function load_system_store(req) {
+    //if endpoints load from core, and this load is for core
+    //(ie, the first load_system_store() out of two),
+    //then endpoints skip it.
+    //endpoints will be updated in the next load_system_store()
+    //once core's in memory system store is updated.
+    const is_endpoint = process.env.HOSTNAME && process.env.HOSTNAME.indexOf("endpoint") !== -1;
+    if (is_endpoint && req?.rpc_params?.load_from_core_step === 'core') {
+        return;
+    }
```
dannyzaken: I think we should perform this check inside load, and returen if `req?.rpc_params?.load_source !== this.source`

src/server/system_services/system_store.js
```diff
@@ -458,6 +472,11 @@
         });
     }
 
+    //return the latest copy of in-memory data
+    async recent_db_data() {
+        return this._load_serial.surround(async () => this.old_db_data);
+    }
```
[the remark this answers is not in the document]
dannyzaken: I agree with the rabbit comment. better to validate that we are in a system_store that has `this.source == 'DB'`

config.js
```diff
@@ -250,7 +250,7 @@ config.INTERNAL_STORAGE_POOL_NAME = 'system-internal-storage-pool';
 config.ALLOW_BUCKET_CREATE_ON_INTERNAL = true;
 config.BUCKET_AUTOCONF_TIER2_ENABLED = false;
 config.SYSTEM_STORE_LOAD_CONCURRENCY = parseInt(process.env.SYSTEM_STORE_LOAD_CONCURRENCY, 10) || 5;
-
+config.SYSTEM_STORE_SOURCE = process.env.SYSTEM_STORE_SOURCE || "db";
```
dannyzaken: maybe use upper case to conform with the rest of the code.
Also add a comment

```suggestion
// SYSTEM_STORE_SOURCE determines the preffered data source for loading system_store data
// This can be either "DB" to load the data from the DB or "CORE" to load the data from the system_server in noobaa-core
config.SYSTEM_STORE_SOURCE = process.env.SYSTEM_STORE_SOURCE || "DB";
```

src/server/system_services/system_store.js
alphaprinz: Core's publish behaviour is different in case load-from-core - there are two steps instead of one.
This means core needs to know endpoint's system store source.
So some parameter in core needs to indicate which publish to use.
Currently I'm reusing the same config.SYSTEM_STORE_SOURCE param.
We can switch core to use a different, new param.
I think that would be more confusing - client will need to enable two different params in two different kube objects (endpoint deployment and core stateful set).
We can discuss this further, or I can just add a new param to tell core to use two steps load.

src/server/system_services/system_store.js
alphaprinz: The first phase is not conditional and always happen, whether we are in one or two steps publish.
The second phase is only for endpoints in case we use two steps publish.
Or have I not understood you?

src/server/common_services/server_inter_process.js
alphaprinz: I've moved the check, but I think your check will skip core's load?

src/api/server_inter_process_api.js
```diff
@@ -17,7 +17,11 @@ module.exports = {
             params: {
                 type: 'object',
                 properties: {
-                    since: { idate: true }
+                    since: { idate: true },
+                    load_from_core_step: {
```
dannyzaken: I suggest renaming it for clarity to something like `load_source` or something similar

src/server/system_services/system_store.js
```diff
@@ -851,3 +929,4 @@ SystemStore._instance = undefined;
 // EXPORTS
 exports.SystemStore = SystemStore;
 exports.get_instance = SystemStore.get_instance;
+exports.SOURCE = SOURCE;
```
dannyzaken, line 932: didn't see that it is used outside of system_store. if not remove the export.

dannyzaken (commented):
2 small comments. also, squash the commits.

dannyzaken (approved):

PR #9171, opened 2025-07-30
[NSFS | Glacier] Disallow UploadPartCopy if the source is not restored

tangledbytes:
### Describe the Problem
This PR addresses the problem where NooBaa NSFS would allow copy of parts even if they are in Glacier storage class but aren't restored.
This doesn't throws an error in case to LTFS but results in a blocking read.

### Explain the Changes
Changes the `read_object_stream` to ensure that we check the object state before performing the read.

1. https://github.ibm.com/Tier2-test/Test-2025/issues/39

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented reading of objects stored in Glacier storage class unless they are fully restored, ensuring proper access restrictions and error messaging for Glacier objects.

### DIFF
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1072,6 +1072,16 @@ class NamespaceFS {
                         dbg.warn('NamespaceFS.read_object_stream mismatch version_id', params.version_id, this._get_version_id_by_xattr(stat));
                         throw error_utils.new_error_code('MISMATCH_VERSION', 'file version does not match the version we asked for');
                     }
+
+                    // Disallow read if the object is in Glacier storage class and isn't restored
+                    const obj_storage_class = Glacier.storage_class_from_xattr(stat.xattr);
+                    const obj_restore_status = Glacier.get_restore_status(stat.xattr, new Date(), file_path);
+                    if (obj_storage_class === s3_utils.STORAGE_CLASS_GLACIER) {
+                        if (obj_restore_status?.ongoing || !obj_restore_status?.expiry_time) {
+                            dbg.warn('read_object_stream: object is not restored yet', obj_restore_status);
+                            throw new S3Error(S3Error.InvalidObjectState);
+                        }
+                    }
                     break;
                 } catch (err) {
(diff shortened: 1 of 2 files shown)


src/sdk/namespace_fs.js
```diff
@@ -1072,6 +1072,16 @@ class NamespaceFS {
                         dbg.warn('NamespaceFS.read_object_stream mismatch version_id', params.version_id, this._get_version_id_by_xattr(stat));
                         throw error_utils.new_error_code('MISMATCH_VERSION', 'file version does not match the version we asked for');
                     }
+
+                    // Disallow read if the object is in Glacier storage class and isn't restored
+                    const obj_storage_class = Glacier.storage_class_from_xattr(stat.xattr);
+                    const obj_restore_status = Glacier.get_restore_status(stat.xattr, new Date(), file_path);
+                    if (obj_storage_class === s3_utils.STORAGE_CLASS_GLACIER) {
+                        if (obj_restore_status?.ongoing || !obj_restore_status?.expiry_time) {
```
romayalon, line 1080: A few questions - 
1. Why aren't you checking obj_restore_status.state === Glacier.RESTORE_STATUS_RESTORED ? worth adding it as a function to glacier.js? is it being checked / should be checked anywhere else?
2. And just to be sure, we would like to throw this error in all functions that call read_object_stream, IIUC UploadPartCopy was the original buggy scenario but this is the required behavior in general, right?

tangledbytes: @romayalon 
1. We just want to make sure that if a restore is ongoing or if there is no active expiry on the object then we must not allow a read to happen. I think an equivalent test could be `if (obj_restore_status.status !== Glacier.RESTORE_STATUS_RESTORED)`. I will take another look and would replace this check with the simpler test. I just want to make sure that it is compatible with out implicit restore statuses.
2. Yeah, we would want to prevent reads everywhere if this condition satisfies. For deep archive, if we don't prevent this then it results in a blocking read.

romayalon (approved):
Approving 
Do you think it's worth adding tests for this?

tangledbytes:
@romayalon, I have added some tests. Thank you 🙂 .

PR #9197, opened 2025-08-25
performance - read only pool POC

alphaprinz:
### Describe the Problem
Add read only pool that connects to read only replica's host.

### Explain the Changes
1. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- New Features
  - Added a dedicated read-only database pool and optional read-only host configuration (env fallback) so read and write pools can target separate hosts.

- Bug Fixes
  - Connection/disconnection messages now reference the requested pool for clearer diagnostics.

- Refactor
  - Connection pooling updated for flexible per-pool selection and per-pool host overrides without impacting other pools.

### DIFF
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -626,10 +626,10 @@ class PostgresTable {
     }
 
 
-    get_pool() {
-        const pool = this.client.get_pool(this.pool_key);
+    get_pool(key = this.pool_key) {
+        const pool = this.client.get_pool(key);
         if (!pool) {
-            throw new Error(`The postgres clients pool ${this.pool_key} disconnected`);
+            throw new Error(`The postgres clients pool ${key} disconnected`);
         }
         return pool;
     }
@@ -1494,6 +1494,10 @@ class PostgresClient extends EventEmitter {
             md: {
                 instance: null,
                 size: config.POSTGRES_MD_MAX_CLIENTS
+            },
+            read_only: {
+                instance: null,
+                size: config.POSTGRES_DEFAULT_MAX_CLIENTS
             }
         };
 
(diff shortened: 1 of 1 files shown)


src/util/postgres_client.js
```diff
@@ -1520,6 +1525,9 @@ class PostgresClient extends EventEmitter {
                 port,
                 ...params,
             };
+            if (host.endsWith("-rw")) {
```
dannyzaken: I suggest using an env to pass the read-only host, e.g. `process.env.POSTGRES_RO_HOST`. This will be more consistent with the rest of the postgres info we pass.

src/util/postgres_client.js
```diff
@@ -1520,6 +1525,9 @@ class PostgresClient extends EventEmitter {
...
+                this.pools.read_only.host = host.replace("-rw", "-r");
```
dannyzaken: I think we should use the `-ro` service. The `-r` service routes to any instance that can perform read, either secondary or primary. the `-ro` service will only route to the secondary, and allow us to control which instance we want to communicate with.

src/util/postgres_client.js
alphaprinz: I don't mind the change, it's just that the other params come from the postgres mounted secret path /etc/postgres-secret/, which will make this param different than the others.

src/util/postgres_client.js
```diff
@@ -1507,6 +1511,7 @@ class PostgresClient extends EventEmitter {
         } else {
             // get the connection configuration. first from env, then from file, then default
             const host = process.env.POSTGRES_HOST || fs_utils.try_read_file_sync(process.env.POSTGRES_HOST_PATH) || '127.0.0.1';
+            const host_ro = process.env.POSTGRES_HOST_RO || fs_utils.try_read_file_sync(process.env.POSTGRES_HOST_RO_PATH) || host.replace("-rw", "-ro");
```
dannyzaken: Remove the `host.replace(..` as default. If there is no POSTGRES_HOST_RO envs, we should not use the RO pool. 
We should not assume that we always have a RO server e.g., if the user uses an external postgres.

dannyzaken (approved):

PR #9198, opened 2025-08-26
NC | Fix fs_napi syslog facility

### LINKED ISSUE #9165: Make SYS_LOG() respect DEBUG_FACILITY instead of hardcoding LOCAL0
### Environment info
- NooBaa Version: 4.20
- Platform: All

### Actual behavior
Currently, our native layer supports only the local0 syslog facility, as shown in the following code snippet:

https://github.com/noobaa/noobaa-core/blob/25db33f68f31694143fccfbf2d2155412b4c04fd/src/native/util/common.h#L44-L50

### Expected behavior
A useful enhancement to this function would be to allow setting the syslog facility via the already existing DEBUG_FACILITY configuration, `config.DEBUG_FACILITY = 'LOG_LOCAL0';`, instead of using the hardcoded local0 value.

### Steps to reproduce
1.

jackyalbo:
### Describe the Problem
fs_napi didn't support changing the facility of the syslog - it was always sending to LOCAL0. 
while for syslog-napi we did support different facilities

### Explain the Changes
1.  Took the facility logic from syslog-napi and moved to common.cpp
2. both syslog-napi and fs-napi will be using the shared facility control logic

1. Fixed #9165 

### Testing Instructions:
1. Change DEBUG_FACILITY to "LOG_LOCAL1",
2. See that logs are being written to the correct syslog facility `/etc/rsyslog.d/noobaa_syslog.conf` (default is /var/log/client_noobaa.log)

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Logging config now supports selecting a syslog facility (e.g., LOG_LOCAL0/1/2) for flexible debug routing.
  * App passes the selected facility to native logging when configuring logs.

* **Breaking Changes**
  * SDK: set_log_config now accepts an additional debug facility string argument.

* **Chores**
  * Linting updated to ignore TypeScript declaration (*.d.ts) files.

### DIFF
--- a/src/util/debug_module.js
+++ b/src/util/debug_module.js
@@ -506,7 +506,7 @@ function log_syslog_builder(syslevel) {
 /* When logging events message should not have any prefix such as time and processes id and 
  *the message should be in json format because of this events are logged directly without using log_builder; 
  * log_builder will format the message with prefix. 
-*/
+ */
 function log_event(event) {
     if (!config.EVENT_LOGGING_ENABLED) {
         return;
@@ -610,7 +610,7 @@ if (console_wrapper) {
 function set_log_config() {
     const dbg_native_conf = debug_config.get_debug_config(process.env.NOOBAA_LOG_LEVEL);
     nb_native().fs.set_debug_level(dbg_native_conf.level);
-    nb_native().fs.set_log_config(config.LOG_TO_STDERR_ENABLED, config.LOG_TO_SYSLOG_ENABLED);
+    nb_native().fs.set_log_config(config.LOG_TO_STDERR_ENABLED, config.LOG_TO_SYSLOG_ENABLED, config.DEBUG_FACILITY);
 }
 
 config.event_emitter.on("config_updated", set_log_config);
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -20,6 +20,7 @@ module.exports = {
         ecmaVersion: 13, // es2022
     },
 
+    ignorePatterns: ['*.d.ts', '**/*.d.ts'],
 
     plugins: [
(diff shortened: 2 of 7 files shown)


src/util/debug_module.js
```diff
@@ -610,7 +610,7 @@ if (console_wrapper) {
 function set_log_config() {
```
naveenpaul1, line 610: Facility should be specific to each message, In our case normal message will have facility `local0` and event will have `local1`, I doint think it shoudl not be a deployment wise configuration like `LOG_TO_STDERR_ENABLED` and `LOG_TO_SYSLOG_ENABLED`.
I think we might need to update the hardcoded value in [here](https://github.com/noobaa/noobaa-core/blob/master/src/native/util/common.h#L47C24-L47C34), I am not 100% sure, Need to test check.

jackyalbo: @naveenpaul1 This code updates this hardcode to this: _convert_facility(SYSLOG_DEBUG_FACILITY);
We already have these configs in the code in config.js, so this code will just let us change the facility we will use for debug instead of harcoded to what we have in SYSLOG_DEBUG_FACILITY.

naveenpaul1: Ok, got it, then it should be fine.
But in my last discussion with Romy, we agreed on updating the rsyslog config with noobaa-specific facility, because what we have now is generic, and multiple rsyslog configs can have the same facility, and it will create issues like this one [DFBUGS-3651](https://issues.redhat.com/browse/DFBUGS-3651)(not orginal issue)
If this issue is not fixing that issue it it should be fine.

liranmauda (approved):

PR #9204, opened 2025-09-02
NC | Update noobaa.spec python requirement in order to support RHEL 10

jackyalbo:
### Describe the Problem
We don't have Python 3.9 in RHEL 10. So we will use 3.12 instead.

### Explain the Changes
1. 

1.  Fixed https://issues.redhat.com/browse/DFBUGS-3868

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Chores**
  * RPM packaging updated to select the system Python on RHEL 9+ and retain Python 3.9 on RHEL 8 and earlier during builds.
  * Preserves existing build tools and runtime behavior; no user-facing changes or upgrade impact.

### DIFF
--- a/src/deploy/RPM_build/noobaa.spec
+++ b/src/deploy/RPM_build/noobaa.spec
@@ -26,13 +26,17 @@ URL:  https://www.noobaa.io/
 Source0:  %{noobaatar}
 
 BuildRequires:  systemd
-BuildRequires:  python3.9
 BuildRequires:  make
 BuildRequires:  gcc-c++
 BuildRequires:  boost-devel
 BuildRequires:  libcap-devel
 %if 0%{?rhel} == 8
 BuildRequires:  gcc-toolset-11
+%endif
+%if 0%{?rhel} > 8
+BuildRequires:  python3 # We can use default version in RHEL 9+
+%else
+BuildRequires:  python3.9 # We need at least 3.9 in RHEL 8
 %endif
 
 Recommends:     jemalloc


jackyalbo:
@b-ranto Can you see if I missed anything here

src/deploy/RPM_build/noobaa.spec
```diff
@@ -26,14 +26,18 @@ URL:  https://www.noobaa.io/
...
 BuildRequires:  gcc-c++
 BuildRequires:  boost-devel
 BuildRequires:  libcap-devel
 %if 0%{?rhel} == 8
 BuildRequires:  gcc-toolset-11
 %endif
+%if 0%{?rhel} == 10
+BuildRequires:  python3.12
+%else
+BuildRequires:  python3.9
```
tangledbytes: I am not sure why it was `python3.9` instead of just `python3` but I think it was probably because one of the built-in module had different behaviour in older python version than in 3.9 🤔 .

We might need to re-assess the above though, as in if we really need to fix it to 3.9 or if just `python3` works as well. If it does then we might be able to get away without using the conditional. WDYT?

b-ranto: It might not even get built with python version older than 3.9 (a lot of projects already dropped support for older versions). The default version in rhel 8 is older than 3.9 though so that is why we probably needed this.

b-ranto:
This did seem to have helped us get rhel 10 builds. Considering all of the above, we may want to revert the conditional and use something like:
```
%if 0%{?rhel} > 8
BuildRequires:  python3 # We can use default version in RHEL 9+
%else
BuildRequires:  python3.9 # We need at least 3.9 in RHEL 8
%fi
```

src/deploy/RPM_build/noobaa.spec
tangledbytes: Yeah @b-ranto, I think that was the reason we fixed it earlier.

tangledbytes (approved):

PR #9209, opened 2025-09-04
Added a fix for POSTGRES_CONNECTION_STRING_PATH handling

aayushchouhan09:
### Describe the Problem
The issue is with the NooBaa core code handling of the `POSTGRES_CONNECTION_STRING_PATH` parameter. When the NooBaa operator sets this environment variable but the connection string file doesn't exist or is not readable, the code uses `fs.readFileSync()` directly which throws an unhandled exception. This causes the NooBaa core pod to crash during operator installation, leading to CI job failures with CrashLoopBackOff errors.

### Explain the Changes
1.  Added proper error handling for `POSTGRES_CONNECTION_STRING_PATH` by wrapping `fs.readFileSync()` in a try-catch block.

1. gap: https://github.com/noobaa/noobaa-operator/pull/1689

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- New Features
  - Support configuring the database via a single Postgres connection string from an environment variable or a file, with automatic fallback to host/user/password/port/database settings. No breaking changes.

- Bug Fixes
  - Improved error reporting when reading the connection string from a file, providing clearer failure messages to aid troubleshooting.

### DIFF
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -1500,10 +1500,14 @@ class PostgresClient extends EventEmitter {
 
         if (process.env.POSTGRES_CONNECTION_STRING_PATH) {
             /** @type {import('pg').PoolConfig} */
-            this.new_pool_params = {
-                connectionString: fs.readFileSync(process.env.POSTGRES_CONNECTION_STRING_PATH, "utf8"),
-                ...params,
-            };
+            try {
+                this.new_pool_params = {
+                    connectionString: fs.readFileSync(process.env.POSTGRES_CONNECTION_STRING_PATH, "utf8"),
+                    ...params,
+                };
+            } catch (err) {
+                throw new Error(`Failed to read connection string file '${process.env.POSTGRES_CONNECTION_STRING_PATH}': ${err.message}`);
+            }
         } else {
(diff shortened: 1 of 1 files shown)


dannyzaken (changes requested):
I commented [here](https://github.com/noobaa/noobaa-operator/pull/1689#pullrequestreview-3185693292) on the necessary changes in the mentioned operator PR. If there is an environment variable for `POSTGRES_CONNECTION_STRING_PATH` but the path does not exist, the code should fail instead of silently falling back to the `else` code. 
We might want to catch the error from `fs.readFileSync` and throw a more descriptive error, so it is clear why the process crashes.

src/util/postgres_client.js
```diff
@@ -1498,10 +1498,13 @@ class PostgresClient extends EventEmitter {
         };
 
 
-        if (process.env.POSTGRES_CONNECTION_STRING_PATH) {
+        const connection_string = process.env.POSTGRES_CONNECTION_STRING ||
```
dannyzaken: `POSTGRES_CONNECTION_STRING` is no longer used.

PR #9216, opened 2025-09-12
Replaced mongodb.ObjectId() with the function mongo_utils.mongoObjectId()

aayushchouhan09:
### Describe the Problem
Part of mongo reference removal task.

### Explain the Changes
1. Replaced `mongodb.ObjectId()` with `mongo_utils.mongoObjectId()` --> using class and constructor.
3. Fixed test files to use new ObjectId generation approach.

1. Replaced `mongodb.ObjectId()` with `mongo_utils.mongoObjectId()`.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Refactor**
  * Centralized identifier generation/validation across services, replacing direct driver ObjectId usage.

* **SDK / Type Declarations**
  * ID types and parsing/generation updated to the centralized identifier representation.

* **Utilities**
  * Identifier helper expanded to validate/generate IDs and exposes a compatible ObjectId class.

* **Database / Postgres**
  * Postgres client and related code now use centralized ID handling; some ID returns normalized to strings.

* **Tools**
  * Developer tools and lifecycle scripts updated to use the standardized ID utility.

* **Tests**
  * Unit and integration tests updated to use the centralized ID helper.

### DIFF
--- a/src/util/mongo_utils.js
+++ b/src/util/mongo_utils.js
@@ -159,7 +159,7 @@
 // }
 
 function is_object_id(id, generate = false) {
-    const ERROR_MSG = 'Argument passed must be a single string of 12 bytes or a string of 24 hex characters';
+    const err_msg = 'Argument passed must be a string of 24 hex characters';
 
     if (id === null || id === undefined) {
         return generate ? mongoObjectId() : false;
@@ -181,7 +181,7 @@ function is_object_id(id, generate = false) {
         return generate ? hex_string.toLowerCase() : true;
     }
 
-    if (generate) throw new Error(ERROR_MSG);
+    if (generate) throw new Error(err_msg);
     return false;
 }
 
@@ -272,7 +272,7 @@ class ObjectID {
     }
 
     valueOf() {
-        return this;
+        return this._id;
     }
 
     toJSON() {
--- a/src/util/mongo_utils.js
+++ b/src/util/mongo_utils.js
@@ -5,7 +5,7 @@
 
 // const _ = require('lodash');
 // const util = require('util');
-const mongodb = require('mongodb');
+// const mongodb = require('mongodb');
 
 // const { RpcError } = require('../rpc');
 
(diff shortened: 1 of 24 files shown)


src/util/mongo_utils.js
```diff
@@ -5,7 +5,7 @@
 
 // const _ = require('lodash');
 // const util = require('util');
-const mongodb = require('mongodb');
+// const mongodb = require('mongodb');
```
naveenpaul1, line 8: remove it

src/util/postgres_client.js
```diff
@@ -1868,28 +1875,28 @@ class PostgresClient extends EventEmitter {
...
         if (_.isArray(doc)) {
             _.each(doc, d => this.fix_id_type(d));
         } else if (doc && doc._id) {
-            doc._id = new mongodb.ObjectId(doc._id);
+            doc._id = new mongo_utils.ObjectId(doc._id);
         }
         return doc;
     }
 
     is_object_id(id) {
```
naveenpaul1, line 1898: did you get a chance to test this flow in upgrade case, 
Just wanted to confirm id changes works also with previously created mongo ids

src/util/mongo_utils.js
```diff
@@ -158,10 +158,34 @@ const mongodb = require('mongodb');
 //     return doc;
 // }
 
-function is_object_id(id) {
-    return (id instanceof mongodb.ObjectId);
+function is_object_id(id, generate = false) {
```
naveenpaul1, line 161: Why need this change?
Are these validations aligned as a MongoDB object id?

src/util/postgres_client.js
aayushchouhan09: Sure, I will test the changes for upgrade.

src/util/mongo_utils.js
aayushchouhan09: Yes we are trying to make it work like mongodb objectid()

src/util/mongo_utils.js
aayushchouhan09: I think we can remove it later at the time of cleaning this file as there are many other functions also present which are commented as well.

Issue #9184, opened 2025-08-06
Add options for standard storage class modes for archive tier


guymguym:
We now have just a global boolean option config.DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD=false that deep archive changes to true to force just GLACIER class uploads.

For various clients we want to have a few modes for upload with storage class standard, and make it per account (or bucket), these are the options we need:
- Always translate to glacier
- Keep standard (resident/premigrated)
- Deny upload (today)

PR #9219, opened 2025-09-15
[NSFS | GLACIER] Add support for GLACIER_DA and forced GLACIER

tangledbytes:
### Explain the Changes
This PR adds support for `GLACIER_DA` storage class and adds support forced `GLACIER` storage class. With the changes in this PR, if `NSFS_GLACIER_FORCE_STORAGE_CLASS = true` then when no storage class is provided OR when `STANDARD` storage class is provided, it will be treated as `GLACIER` storage class object. The following combinations are expected and supported:
1. `NSFS_GLACIER_FORCE_STORAGE_CLASS = false; DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD = false` - No storage class or storage class `STANDARD` would work.
2. `NSFS_GLACIER_FORCE_STORAGE_CLASS = false; DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD = true` - No storage class or storage class `STANDARD` would **not** work.
3. `NSFS_GLACIER_FORCE_STORAGE_CLASS = true; DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD = true` - No storage class or storage class `STANDARD` would work **however** the storage class would be implicitly changed to `GLACIER`.
4. `NSFS_GLACIER_FORCE_STORAGE_CLASS = true; DENY_UPLOAD_TO_STORAGE_CLASS_STANDARD = false` - _Same as above_.

The above tries to achive the ask which was:
> So, just to summarize,
> - if customer specified Glacier FR, Glacier FR is set in Deep Archive (as of today)
> - Glacier DA -> Glacier DA (we are rejecting today)
> - Standard -> Glacier FR
>     - For a specific customer, a special setting will be enabled so that Standard -> Standard
> - (nothing specified) -> Glacier FR
>     - For a specific customer, a special setting will be enabled so that the unspecified case -> Standard
> - any other classes -> error with InvalidStorageClass

https://github.com/noobaa/noobaa-core/issues/9184

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added support for a new GLACIER_DA storage class.
  * NamespaceFS now recognizes multiple Glacier classes (GLACIER, GLACIER_IR, GLACIER_DA).

* **Configuration**
  * New optional setting to force Glacier storage class when none or STANDARD is requested (opt-in, default: off).

* **Improvements**
  * Storage-class normalization/override now happens earlier in request handling for more consistent behavior.

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -107,6 +107,9 @@ async function handle_request(req, res) {
     }
     http_utils.check_headers(req, headers_options);
 
+    // Will override the storage class if configured
+    s3_utils.override_storage_class(req);
+
     const redirect = await populate_request_additional_info_or_redirect(req);
     if (redirect) {
         res.setHeader('Location', redirect);
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -20,6 +20,8 @@ const STORAGE_CLASS_STANDARD = 'STANDARD';
 const STORAGE_CLASS_GLACIER = 'GLACIER'; // "S3 Glacier Flexible Retrieval"
 /** @type {nb.StorageClass} */
 const STORAGE_CLASS_GLACIER_IR = 'GLACIER_IR'; // "S3 Glacier Instant Retrieval"
+/** @type {nb.StorageClass} */
+const STORAGE_CLASS_GLACIER_DA = 'GLACIER_DA'; // "DBS3 specific Storage Class"
 
 const DEFAULT_S3_USER = Object.freeze({
     ID: '123',
@@ -382,6 +384,7 @@ function parse_storage_class(storage_class) {
     if (!storage_class) return STORAGE_CLASS_STANDARD;
     if (storage_class === STORAGE_CLASS_STANDARD) return STORAGE_CLASS_STANDARD;
     if (storage_class === STORAGE_CLASS_GLACIER) return STORAGE_CLASS_GLACIER;
+    if (storage_class === STORAGE_CLASS_GLACIER_DA) return STORAGE_CLASS_GLACIER_DA;
     if (storage_class === STORAGE_CLASS_GLACIER_IR) return STORAGE_CLASS_GLACIER_IR;
     throw new Error(`No such s3 storage class ${storage_class}`);
 }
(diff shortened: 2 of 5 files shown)


src/sdk/namespace_fs.js
```diff
@@ -3612,11 +3612,15 @@ class NamespaceFS {
     }
 
     async _is_storage_class_supported(storage_class) {
+        const glacier_storage_classes = [
+            s3_utils.STORAGE_CLASS_GLACIER,
+            s3_utils.STORAGE_CLASS_GLACIER_DA,
+            s3_utils.STORAGE_CLASS_GLACIER_IR,
```
naveenpaul1, line 3618: Newly added storage class in STORAGE_CLASS_GLACIER_DA, And STORAGE_CLASS_GLACIER_IR was missing in previous check? Is there any reason for adding it now?

tangledbytes: Yep, these are newly added. `GLACIER_DA` is something DBS3 specific and they want to support this one as well on the other hand our implementation is quite close to `GLACIER_IR` (instant retrieval) because DBS2 usually triggers the retrieval within 10 minutes and the restores usually finish within 20 minutes of submitting the request unlike AWS S3 which can take couple of hours.

naveenpaul1 (approved):

src/endpoint/s3/s3_rest.js
```diff
@@ -107,6 +107,9 @@ async function handle_request(req, res) {
     }
     http_utils.check_headers(req, headers_options);
 
+    // Will override the storage class if configured
+    s3_utils.override_storage_class(req);
+
```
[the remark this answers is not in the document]
guymguym: @tangledbytes I think this comment on override_storage_class() is indeed concerning - do we have a test that covers this path with auth??

src/sdk/namespace_fs.js
```diff
@@ -3612,11 +3612,15 @@ class NamespaceFS {
     }
 
     async _is_storage_class_supported(storage_class) {
+        const glacier_storage_classes = [
```
guymguym, line 3615: Maybe old school, but I find it weird that we define this constant inside the check function. I would define it in module scope probably...

src/endpoint/s3/s3_utils.js
```diff
@@ -819,9 +822,19 @@ function parse_body_public_access_block(req) {
...
 }
 
+function override_storage_class(req) {
+    if (
+        config.NSFS_GLACIER_FORCE_STORAGE_CLASS &&
+        parse_storage_class_header(req) === STORAGE_CLASS_STANDARD
+    ) {
+        req.headers['x-amz-storage-class'] = STORAGE_CLASS_GLACIER;
+    }
+}
```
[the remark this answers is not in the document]
guymguym: @tangledbytes I agree with rabbit's comment , and would also reconsider if we can achieve this without overriding the input headers, because this can be a source of issues.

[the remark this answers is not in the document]
tangledbytes: @guymguym addressed it here - https://github.com/noobaa/noobaa-core/pull/9278

src/sdk/namespace_fs.js
tangledbytes: @guymguym I just did it so that I don't have to chain a lot of ifs and could simply use `.include`. There was no other reason for defining that variable 🙂 .

guymguym: yeah that's a fair point!

PR #9222, opened 2025-09-18
[DB-Cleaner] fixing find_deleted_objects to hit index

jackyalbo:
### Describe the Problem
old find_deleted_object didn't to_ts when looking for the last deleted object - avoiding md-aggregator index and also any defaulting to sequance scanning the table.

### Explain the Changes
1. As this query is used only by db-cleaner - used directly executeSQL so we can better control the query structure
2. We will now also check for recalimed - in case the object is deleted but not yet reclaimed, we won't remove from the DB
3. It will now hit the md-aggregator index 
```
EXPLAIN ANALYZE SELECT '_id' FROM objectmds WHERE (to_ts(data->>'deleted')<to_ts('"2025-09-19T14:20:21.570Z"') and data ? 'deleted' and data ? 'reclaimed') LIMIT 1000;
                                                                                  QUERY PLAN
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.43..1086.30 rows=1000 width=32) (actual time=0.073..2.493 rows=1000 loops=1)
   ->  Index Scan using idx_btree_objectmds_aggregate_by_delete_dates on objectmds  (cost=0.43..1110691.74 rows=1022860 width=32) (actual time=0.071..2.415 rows=1000 loops=1)
         Index Cond: (to_ts((data ->> 'deleted'::text)) < '2025-09-19 14:20:21.57'::timestamp without time zone)
         Filter: (data ? 'reclaimed'::text)
 Planning Time: 1.263 ms
 Execution Time: 2.557 ms
(6 rows)
```
just to compare:
```
EXPLAIN ANALYSE SELECT jsonb_build_object('_id', data->'_id', 'deleted', data->'deleted') as data FROM objectmds WHERE (data->'deleted'<'"2025-08-28T14:20:21.570Z"'::jsonb and data ? 'deleted') LIMIT 1000;
                                                        QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=0.00..340.76 rows=1000 width=32) (actual time=2702.703..2702.707 rows=0 loops=1)
   ->  Seq Scan on objectmds  (cost=0.00..352103.40 rows=1033297 width=32) (actual time=2702.702..2702.702 rows=0 loops=1)
         Filter: ((data ? 'deleted'::text) AND ((data -> 'deleted'::text) < '"2025-08-28T14:20:21.570Z"'::jsonb))
         Rows Removed by Filter: 3132279
 Planning Time: 3.832 ms
 Execution Time: 2702.738 ms
(6 rows)
```

1. https://issues.redhat.com/browse/DFBUGS-2232

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Can list deleted-and-reclaimed objects before a time cutoff, returning unique IDs up to a configurable limit.

* **Refactor**
  * Retrieval switched to PostgreSQL-backed queries for consistency.
  * Time handling standardized to ISO strings for time-based operations.

* **Tests**
  * Added PostgreSQL-only integration test validating deletion-by-query, reclamation marking, and limited retrieval.

* **Documentation**
  * Added JSDoc for the new retrieval method.

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -720,7 +720,7 @@ class MDStore {
         limit,
         return_results = false,
     }) {
-        const params = [new Date()];
+        const params = [new Date().toISOString()];
         const sql_conditions = [];
         if (key) {
             params.push(key.source);
@@ -1110,20 +1110,21 @@ class MDStore {
         return buckets;
     }
 
+    /**
+     * Find deleted objects that were deleted before max_delete_time and are reclaimed.
+     *
+     * @param {number} max_delete_time - timestamp in milliseconds
+     * @param {number} limit
+     * @returns {Promise<nb.ID[]>}
+     */
     async find_deleted_objects(max_delete_time, limit) {
-        const objects = await this._objects.find({
-            deleted: {
-                $lt: new Date(max_delete_time),
-                $exists: true // This forces the index to be used
-            },
-        }, {
-            limit: Math.min(limit, 1000),
-            projection: {
-                _id: 1,
-                deleted: 1
-            }
-        });
-        return db_client.instance().uniq_ids(objects, '_id');
+        const query_limit = limit || 1000;
+        const query = `SELECT _id 
+        FROM ${this._objects.name}
+        WHERE (to_ts(data->>'deleted')<to_ts($1) and data ? 'deleted' and data ? 'reclaimed') 
+        LIMIT ${query_limit};`;
+        const result = await this._objects.executeSQL(query, [new Date(max_delete_time).toISOString()]);
+        return db_client.instance().uniq_ids(result.rows, '_id');
     }
 
     async db_delete_objects(object_ids) {
--- a/src/test/integration_tests/db/test_md_store.js
+++ b/src/test/integration_tests/db/test_md_store.js
@@ -1,4 +1,5 @@
 /* Copyright (C) 2016 NooBaa */
+/* eslint max-lines-per-function: ["error", 2000]*/
 'use strict';
 
 // setup coretest first to prepare the env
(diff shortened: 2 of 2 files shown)


src/server/object_services/md_store.js
```diff
@@ -1110,20 +1110,20 @@ class MDStore {
...
-                _id: 1,
-                deleted: 1
-            }
-        });
-        return db_client.instance().uniq_ids(objects, '_id');
+        const query = `SELECT _id 
+        FROM ${this._objects.name}
+        WHERE (to_ts(data->>'deleted')<to_ts($1) and data ? 'deleted' and data ? 'reclaimed') 
+        LIMIT ${Math.min(limit || 0, 1000)};`;
+        const result = await this._objects.executeSQL(query, [new Date(max_delete_time).toISOString()]);
```
dannyzaken, line 1126: The default limit should not be 0. Also, to make it more readable, calculate the limit outside of the string literal or use a default value for the argument. 


```suggestion
        const query_limit = limit || 1000;
        const query = `SELECT _id 
        FROM ${this._objects.name}
        WHERE (to_ts(data->>'deleted')<to_ts($1) and data ? 'deleted' and data ? 'reclaimed') 
        LIMIT ${query_limit};`;
        const result = await this._objects.executeSQL(query, [new Date(max_delete_time).toISOString()]);
```

jackyalbo: It's better to ensure the limit is a maximum of 1K... So I will use: 
```suggestion
        const query_limit = Math.min(limit || 1000, 1000);
```

src/server/object_services/md_store.js
```diff
@@ -1110,20 +1110,21 @@ class MDStore {
...
-            },
-        }, {
-            limit: Math.min(limit, 1000),
-            projection: {
-                _id: 1,
-                deleted: 1
-            }
-        });
-        return db_client.instance().uniq_ids(objects, '_id');
+        const query_limit = Math.min(limit || 1000, 1000);
```
dannyzaken: Why do we need to limit the `limit` to 1000? It's very limiting 😄 
As long as we have a default, let's allow any value, so we can change it in config.js if we want. WDYT?

jackyalbo: As currently we don't use the default - I don't really mind :)

dannyzaken (approved):

PR #9237, opened 2025-10-07
Read only uses

alphaprinz:
### Describe the Problem
Use new 'read only' pool (connected to cnpg replicated read-only cluster) in order to alleviate load from main, writable pg cluster.

### Explain the Changes
1. Move some background queries to read-only pool:
-db cleancer
-object reclaimer's first find
-scrubber's first find

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Query API accepts an optional preferred-pool parameter to route individual queries to a specific database pool.
  * New config flag POSTGRES_USE_READ_ONLY to toggle use of read-only replicas (defaults to true).

* **Improvements**
  * Read operations prefer read-only replicas to distribute load.
  * Queries fall back and retry on the primary pool when a preferred pool is unavailable.

* **Observability**
  * Added debug logging for query host selection and retry behavior.

### DIFF
--- a/config.js
+++ b/config.js
@@ -251,6 +251,10 @@ config.DB_TYPE = /** @type {nb.DBType} */ (process.env.DB_TYPE || 'postgres');
 config.POSTGRES_DEFAULT_MAX_CLIENTS = 10;
 config.POSTGRES_MD_MAX_CLIENTS = (process.env.LOCAL_MD_SERVER === 'true') ? 70 : 10;
 
+//whether to use read-only postgres replica cluster
+//ro host is set by operator in process.env.POSTGRES_HOST_RO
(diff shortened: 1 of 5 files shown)


dannyzaken (commented):
In general, this looks good. A few general comments\questions 
* We should consider the case where the secondary is down. Maybe we should fallback to the primary?
* We should make sure that the rest of the code behaves nicely if the secondary returns stale data (e.g., object_mds that were marked as reclaimed but were not reflected yet in the secondary). From a quick look at the code, it looks ok, but I didn't look too deeply into it.

dannyzaken (approved):

PR #9243, opened 2025-10-14
Support AGENT_CONFIG from volume mounts and remove secret logging

aayushchouhan09:
### Describe the Problem
`AGENT_CONFIG` was supported as environment variable, not as volume-mounted secrets. Also the agent logs exposed sensitive configuration data.

### Explain the Changes
1. Added support for reading AGENT_CONFIG from volume-mounted secrets
2. Removed sensitive data logging from agent pod logs

1. Bug: https://issues.redhat.com/browse/DFBUGS-2687
2. Operator PR: https://github.com/noobaa/noobaa-operator/pull/1718

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved agent configuration initialization with stricter validation and clearer error messages.
  * Now exits early if a valid configuration already exists, and resolves config from environment or default locations.
  * Supports raw JSON or base64-encoded JSON safely: only writes a new config after successful validation/decoding, and fails fast on invalid or missing config.

### DIFF
--- a/src/deploy/NVA_build/noobaa_init.sh
+++ b/src/deploy/NVA_build/noobaa_init.sh
@@ -122,16 +122,24 @@ run_internal_process() {
 
 prepare_agent_conf() {
   AGENT_CONF_FILE="/noobaa_storage/agent_conf.json"
-  if [ -z ${AGENT_CONFIG} ]
-  then
-    echo "AGENT_CONFIG is required ENV variable. AGENT_CONFIG is missing. Exit"
+
+  [ -f "$AGENT_CONF_FILE" ] && return 0
+
+  # get AGENT_CONFIG from env var or file
+  AGENT_CONFIG_PATH=${AGENT_CONFIG_PATH:-"/etc/agent-config/agent_config"}
+  AGENT_CONFIG=${AGENT_CONFIG:-$(cat "$AGENT_CONFIG_PATH" 2>/dev/null || echo "")}
+
+  if [ -z "${AGENT_CONFIG}" ]; then
+    echo "AGENT_CONFIG is required. AGENT_CONFIG is not found in env or $AGENT_CONFIG_PATH. Exit"
     exit 1
-  else
-    echo "Got base64 agent_conf: ${AGENT_CONFIG}"
-    if [ ! -f $AGENT_CONF_FILE ]; then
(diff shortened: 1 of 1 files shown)


src/deploy/NVA_build/noobaa_init.sh
```diff
@@ -122,17 +122,27 @@ run_internal_process() {
...
 prepare_agent_conf() {
   AGENT_CONF_FILE="/noobaa_storage/agent_conf.json"
-  if [ -z ${AGENT_CONFIG} ]
-  then
-    echo "AGENT_CONFIG is required ENV variable. AGENT_CONFIG is missing. Exit"
+
+  # get AGENT_CONFIG from env var or config.js (volume mount)
+  if [ -z ${AGENT_CONFIG} ]; then
+    cd /root/node_modules/noobaa-core/
+    AGENT_CONFIG=$(node -e "const config = require('./config'); console.log(config.AGENT_CONFIG || '')")
```
tangledbytes: Maybe `node -p 'require("config").AGENT_CONFIG || ""'` would do?

aayushchouhan09: Done, Thanks

config.js
```diff
@@ -242,6 +242,12 @@ if (process.env.CONTAINER_PLATFORM || process.env.LOCAL_MD_SERVER) {
 
 config.ROOT_KEY_MOUNT = '/etc/noobaa-server/root_keys';
 
+//////////////////
+// AGENT CONFIG //
+//////////////////
+config.AGENT_CONFIG_PATH = process.env.AGENT_CONFIG_PATH || '/etc/agent-config/agent_config';
+config.AGENT_CONFIG = process.env.AGENT_CONFIG || _get_data_from_file(config.AGENT_CONFIG_PATH);
```
tangledbytes: Is the only reason these config variables are defined here to use in `noobaa_init.sh`? If yes, then I think it might be better to simply add the code there itself. Something like `AGENT_CONFIG=${AGENT_CONFIG:-$(cat "$AGENT_CONFIG_PATH")}`?

aayushchouhan09: yes, make sense. Removed and updated, Thanks!

tangledbytes (approved):
LGTM.

PR #9260, opened 2025-10-30
Added support for tagging for IAM users

aayushchouhan09:
### Describe the Problem
Support tagging for IAM users.

### Explain the Changes
1. Added three IAM tag operations: `TagUser`, `UntagUser`, and `ListUserTags` APIs.

1. JIRA: https://issues.redhat.com/browse/RHSTOR-8032

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * User tagging APIs: TagUser and UntagUser endpoints plus SDK methods; ListUserTags supported where implemented.

* **Validation**
  * IAM-style tag and tag-key validators with AWS-compatible pattern and length checks.

* **Configuration Updates**
  * Maximum tags per user set to 50; listings return sorted, non-paginated results when available.

* **Behavior**
  * Tag limits enforced; some storage backends implement tagging fully while others provide stubs.

* **Tests**
  * Updated tests to assert length-related validation messages.


src/util/string_utils.js
```diff
@@ -15,6 +15,7 @@ const AWS_IAM_PATH_REGEXP = /^(\u002F|\u002F[\u0021-\u007E]+\u002F)$/;
 const AWS_USERNAME_REGEXP = /^[\w+=,.@-]+$/;
 const AWS_IAM_LIST_MARKER = /^[\u0020-\u00FF]+$/;
 const AWS_IAM_ACCESS_KEY_INPUT_REGEXP = /^[\w]+$/;
+const AWS_IAM_TAG_KEY_AND_VALUE_REGEXP = /^[\p{L}\p{Z}\p{N}_.:/=+\-@]+$/u;
```
shirady, line 18: Why do we need this `/u`?
You can add a test in `test_string_utils.test.js` to help us understand.

src/sdk/nb.d.ts
```diff
@@ -901,6 +901,10 @@ interface AccountSpace {
     update_user(params: object, account_sdk: AccountSDK): Promise<any>;
     delete_user(params: object, account_sdk: AccountSDK): Promise<any>;
     list_users(params: object, account_sdk: AccountSDK): Promise<any>;
+    // user tags
+    tag_user(params: object, account_sdk: AccountSDK): Promise<any>;
+    untag_user(params: object, account_sdk: AccountSDK): Promise<any>;
+    list_user_tags(params: object, account_sdk: AccountSDK): Promise<any>;
```
shirady: I would move it after access keys (we don't have the implementation in NC).

src/endpoint/iam/ops/iam_tag_user.js
```diff
@@ -0,0 +1,49 @@
...
+const iam_utils = require('../iam_utils');
+const iam_constants = require('../iam_constants');
+const { CONTENT_TYPE_APP_FORM_URLENCODED } = require('../../../util/http_utils');
+
+/**
+ * https://docs.aws.amazon.com/IAM/latest/APIReference/API_TagUser.html
+ */
+async function tag_user(req, res) {
+
+    const tags = [];
```
shirady: Could you move it to a function?
What are you trying to achieve here?

src/endpoint/iam/ops/iam_untag_user.js
```diff
@@ -0,0 +1,46 @@
...
+ * https://docs.aws.amazon.com/IAM/latest/APIReference/API_UntagUser.html
+ */
+async function untag_user(req, res) {
+
+    const tag_keys = [];
+    let tag_index = 1;
+    while (req.body[`tag_keys_member_${tag_index}`]) {
+        tag_keys.push(req.body[`tag_keys_member_${tag_index}`]);
+        tag_index += 1;
+    }
```
shirady: If you had a function, then you can reuse it here also.

src/endpoint/iam/iam_rest.js
```diff
@@ -63,6 +63,8 @@ const ACTIONS = Object.freeze({
     'ListUserPolicies': 'list_user_policies',
     'ListUserTags': 'list_user_tags',
     'ListVirtualMFADevices': 'list_virtual_mfa_devices',
+    'TagUser': 'tag_user',
+    'UntagUser': 'untag_user',
```
shirady: Could you move the tags operations after access keys (and also move the list user tags near)?

src/endpoint/iam/iam_utils.js
```diff
@@ -117,6 +117,15 @@ function validate_user_params(action, params) {
...
           break;
+        case iam_constants.IAM_ACTIONS.TAG_USER:
+            validate_tag_user_params(params);
+          break;
+        case iam_constants.IAM_ACTIONS.UNTAG_USER:
+            validate_untag_user_params(params);
+          break;
+        case iam_constants.IAM_ACTIONS.LIST_USER_TAGS:
+            validate_list_user_tags_params(params);
+          break;
```
shirady: It is not USER API, it is in the scope of TAGS, could you move it to a separate function?

src/endpoint/iam/iam_utils.js
```diff
@@ -478,6 +487,49 @@ function translate_rpc_error(err) {
...
 }
 
+/**
+ * validate_tag_user_params checks the params for tag_user action
+ * @param {object} params
+ */
+function validate_tag_user_params(params) {
+    try {
+        check_required_username(params);
+        validation_utils.validate_iam_tags(params.tags);
```
shirady: Could you please switch the order? Validate the username (required + rest) and then validate tags?

src/endpoint/iam/iam_utils.js
```diff
@@ -478,6 +487,49 @@ function translate_rpc_error(err) {
...
+
+/**
+ * validate_untag_user_params checks the params for untag_user action
+ * @param {object} params
+ */
+function validate_untag_user_params(params) {
+    try {
+        check_required_username(params);
+        validation_utils.validate_iam_tag_keys(params.tag_keys);
+        validation_utils.validate_username(params.username, iam_constants.IAM_PARAMETER_NAME.USERNAME);
```
shirady, line 688: Same here - Could you please switch the order? Validate the username (required + rest) and then validate tags?

src/endpoint/iam/iam_utils.js
```diff
@@ -478,6 +487,49 @@ function translate_rpc_error(err) {
...
+/**
+ * validate_list_user_tags_params checks the params for list_user_tags action
+ * @param {object} params
+ */
+function validate_list_user_tags_params(params) {
+    try {
+        check_required_username(params);
+        validate_marker(params.marker);
+        validate_max_items(params.max_items);
+        validation_utils.validate_username(params.username, iam_constants.IAM_PARAMETER_NAME.USERNAME);
```
shirady: Same here - Could you please switch the order? Validate the username (required + rest)

src/endpoint/iam/ops/iam_list_user_tags.js
```diff
@@ -14,21 +14,26 @@ async function list_user_tags(req, res) {
...
+    }
+    dbg.log2('list_user_tags reply', result);
 
     return {
         ListUserTagsResponse: {
-            ListUserTagsResult: {
-                Tags: [],
-                IsTruncated: false,
-            },
+            ListUserTagsResult: result,
```
shirady: Could you keep the structure as it was?
You would need to add the implementation of returning an empty list in accountspace_fs.

src/util/validation_utils.js
```diff
@@ -102,8 +102,98 @@ function validate_username(input_username, parameter_name = iam_constants.IAM_PA
...
+        const tag_position = `Tags.member.${i + 1}`;
+
+        _type_check_input('object', tag, tag_position);
+        // reject null and arrays
+        if (tag === null || Array.isArray(tag)) {
+            const message_with_details = `1 validation error detected: Value ${tag} at ` +
+                `'${tag_position}'  failed to satisfy constraint: Member must be object`;
+            throw new RpcError('VALIDATION_ERROR', message_with_details);
+        }
+        _type_check_input('string', tag.key, `${tag_position}.Key`);
```
shirady: Could you separate this check for the key tag and value tag? (so the function would not be that long).

src/sdk/accountspace_nb.js
```diff
@@ -247,6 +247,136 @@ class AccountSpaceNB {
...
+
+        const tags = paginated_tags.map(tag => ({
+            member: {
+                Key: tag.key,
+                Value: tag.value
+            }
+        }));
+
+        const is_truncated = end_index < sorted_tags.length;
+        const marker = is_truncated && paginated_tags.length > 0 ? paginated_tags[paginated_tags.length - 1].key : undefined;
```
shirady: Could you move it to a separate function?
We didn't implement pagination so far, and if you move it to a function it can be reusable.

shirady (commented):
Adding a couple of comments

src/endpoint/iam/ops/iam_list_user_tags.js
aayushchouhan09: We are returning the same structure as it was before, here we are adding marker only if it was present. I will add a proper comment and default return values.

src/endpoint/iam/ops/iam_tag_user.js
aayushchouhan09: We are fetching the key and value pairs for tagging from AWS-styled indexing, will create a function for it.

src/sdk/accountspace_nb.js
aayushchouhan09: We will not support pagination in this PR, removing it. Thanks

src/endpoint/iam/ops/iam_list_user_tags.js
aayushchouhan09: removed pagination support for now, Thanks

src/util/string_utils.js
aayushchouhan09: As per my understanding `/u` here stands for Unicode, if we don't put it in the end then `\p{L}` or `\p{Z}` would not work, they would be treated as plain characters otherwise.

\p{L} : Any letter (Unicode property: Letter)
\p{Z} : space, non-breaking space, etc.
\p{N} : Any number (Unicode property: Number)

shirady: So it is relevant only for the tag regex?

src/util/validation_utils.js
```diff
@@ -170,8 +170,96 @@ function validate_policy_document(input_policy_document, parameter_name = iam_co
...
+/**
+ * _validate_tag_pattern validates tag key or value against AWS pattern
+ * @param {string} value - tag key or value to validate
+ * @param {string} position - position identifier for error message
+ */
+function _validate_tag_pattern(value, position) {
+    const is_valid = AWS_IAM_TAG_KEY_AND_VALUE_REGEXP.test(value);
+    if (!is_valid) {
+        const message_with_details = `1 validation error detected: Value at '${position}' ` +
+            `failed to satisfy constraint: Member can only contain letters, numbers, spaces, and the following characters: _.:/=+\\-@`;
```
shirady: Is it the error from AWS?

src/endpoint/iam/ops/iam_list_user_tags.js
```diff
@@ -14,20 +14,19 @@ async function list_user_tags(req, res) {
...
+    const reply = await req.account_sdk.list_user_tags(params);
+    dbg.log2('list_user_tags reply', reply);
 
     return {
         ListUserTagsResponse: {
             ListUserTagsResult: {
-                Tags: [],
-                IsTruncated: false,
+                Tags: reply.tags || [],
+                IsTruncated: reply.is_truncated || false
```
shirady: Please don't default the value here and only return what you had in the reply.
```suggestion
                Tags: reply.tags,
                IsTruncated: reply.is_truncated
```

For the NC default value, you can do something similar to this:
https://github.com/noobaa/noobaa-core/blob/a566cda9239262765570ed38a35fb6a218cc3576/src/sdk/accountspace_fs.js#L550-L559

src/endpoint/iam/iam_utils.js
```diff
@@ -81,12 +81,15 @@ function parse_max_items(input_max_items) {
 }
 
 /**
- * validate_params will call the aquivalent function in user or access key
+ * validate_params will call the aquivalent function in user, access key, or tagging
```
shirady: Good comment, just suggesting to make it an example so we will not need to maintain it.
```suggestion
 * validate_params will call the equivalent function (for example: user, access key, tagging, etc.).
```

src/endpoint/iam/ops/iam_tag_user.js
```diff
@@ -0,0 +1,40 @@
...
+const { CONTENT_TYPE_APP_FORM_URLENCODED } = require('../../../util/http_utils');
+
+/**
+ * https://docs.aws.amazon.com/IAM/latest/APIReference/API_TagUser.html
+ */
+async function tag_user(req, res) {
+
+    const params = {
+        username: req.body.user_name,
+        tags: iam_utils.parse_tags_from_request_body(req.body),
```
shirady, line 19: Why do we parse the body and not the tags in the body?
If you can give an example - what we had before and what we have after it can be useful.

aayushchouhan09: Here we need to pass `req.body` because of aws encoding format. I have added a comment for input/output example.

AWS encoded req.body: `{ action: 'TagUser', version: '2010-05-08', user_name: 'test-user', tags_member_1_key: 'key', tags_member_1_value: 'test' }`

after parsing params.tags: `[ { key: 'key', value: 'test' } ]`

src/util/string_utils.js
aayushchouhan09: yes right

src/util/validation_utils.js
aayushchouhan09: updated the error as expected

src/util/validation_utils.js
```diff
@@ -170,8 +176,136 @@ function validate_policy_document(input_policy_document, parameter_name = iam_co
...
 
 
+/**
+ * _validate_length_constraint validates length constraints with AWS-style error messages
+ * @param {string|Array} value - value to validate
+ * @param {string} position - position identifier for error message
+ * @param {number} min_length - minimum length (optional)
+ * @param {number} max_length - maximum length (optional)
+ */
+function _validate_length_constraint(value, position, min_length, max_length) {
```
shirady: Why create a new function when we already have `_length_check_input`?

src/util/validation_utils.js
```diff
@@ -170,8 +176,136 @@ function validate_policy_document(input_policy_document, parameter_name = iam_co
...
+    }
+
+    _validate_length_constraint(tag_keys, 'tagKeys', undefined, iam_constants.DEFAULT_MAX_TAGS);
+
+    for (let i = 0; i < tag_keys.length; i++) {
+        _type_check_input('string', tag_keys[i], 'tagKeys');
+
+        if (_has_tag_key_violation(tag_keys[i])) {
+            throw new RpcError('VALIDATION_ERROR',
+                `1 validation error detected: Value at 'tagKeys' failed to satisfy constraint: Member must satisfy constraint: [${TAG_KEY_CONSTRAINTS.join(', ')}]`);
```
shirady, line 288: Did you check this error against AWS?
It seems like it prints the array on constraints.

src/endpoint/iam/iam_constants.js
```diff
@@ -51,6 +57,7 @@ const IDENTITY_ENUM = Object.freeze({
 
 // miscellaneous
 const DEFAULT_MAX_ITEMS = 100;
+const DEFAULT_MAX_TAGS = 50;
```
shirady: Why default?

```suggestion
const MAX_TAGS = 50;
```

src/endpoint/iam/ops/iam_list_user_tags.js
```diff
@@ -14,20 +14,19 @@ async function list_user_tags(req, res) {
     const params = {
         username: req.body.user_name,
         marker: req.body.marker,
-        max_items: iam_utils.parse_max_items(req.body.max_items) ?? iam_constants.DEFAULT_MAX_ITEMS,
+        max_items: iam_utils.parse_max_items(req.body.max_items) ?? iam_constants.DEFAULT_MAX_TAGS,
```
shirady: There no need to change, according to the AWS docs ([link](https://docs.aws.amazon.com/IAM/latest/APIReference/API_ListUserTags.html)):

> If you do not include this parameter, the number of items defaults to 100.

src/endpoint/iam/ops/iam_tag_user.js
```diff
@@ -0,0 +1,43 @@
...
+ * https://docs.aws.amazon.com/IAM/latest/APIReference/API_TagUser.html
+ */
+async function tag_user(req, res) {
+
+    // parse tags from AWS encoded format
+    // input: { tags_member_1_key: 'env', tags_member_1_value: 'prod', ... }
+    // output: [{ key: 'env', value: 'prod' }, ...]
+    const params = {
+        username: req.body.user_name,
+        tags: iam_utils.parse_tags_from_request_body(req.body),
```
shirady, line 19: Did you test that you get the expected error when parsing fails?

src/sdk/accountspace_nb.js
```diff
@@ -328,6 +328,117 @@ class AccountSpaceNB {
...
+
+        const tags_map = new Map();
+        for (const tag of existing_tags) {
+            tags_map.set(tag.key, tag.value);
+        }
+        for (const tag of params.tags) {
+            tags_map.set(tag.key, tag.value);
+        }
+
+        // enforce AWS 50-tag limit after merging
```
shirady: Removing the number so we won't need to maintain it
```suggestion
        // enforce AWS tag limit after merging
```

shirady (approved):
Approving with a couple of comments

src/util/validation_utils.js
aayushchouhan09: yes, it should look like (verified in aws):

`An error occurred (ValidationError) when calling the UntagUser operation: 1 validation error detected: Value at 'tagKeys' failed to satisfy constraint: Member must satisfy constraint: [Member must have length less than or equal to 128, Member must have length greater than or equal to 1, Member must satisfy regular expression pattern: [\p{L}\p{Z}\p{N}_.:/=+\-@]+, Member must not be null]`

src/util/validation_utils.js
aayushchouhan09: We can remove it after some minimal modification to `_length_check_input()`

src/endpoint/iam/ops/iam_tag_user.js
aayushchouhan09: Added error handling, Thanks

naveenpaul1 (approved):

PR #9269, opened 2025-11-05
IAM | Account Schema Changes for Supporting Inline User Policy

shirady:
### Describe the Problem
Adding the schema for IAM User Policy inline policy.
The API implementation would be in a separate PR.

### Explain the Changes
1. Define in `common_api` `iam_user_policy_document` based on `bucket_policy` structure without `Principal` and `NotPrincipal` (and `Condition`), and create `iam_user_policy`.
2. Add the property `iam_user_policies` in `account` schema (which also serves for IAM users).

Note: the explanation why I could not just `ref` `bucket_policy` in the comments ([link](https://github.com/noobaa/noobaa-core/pull/9269#discussion_r2495179110)).

### Issues:
1. none

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * IAM user policies are now supported, enabling inline policy definition and validation for individual users.
  * Account profiles can store and reference multiple user policies, extending per-user access control options.
  * User policies include configurable statements with action and resource specifications and a simplified schema for common inline policy use.

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -487,6 +487,90 @@ module.exports = {
             }
         },
 
+        // based on bucket policy without Principal and NotPrincipal since are not used in inline policies
+        // removed the Condition as we don't support it yet
+        iam_user_policy_document: {
+            type: 'object',
+            required: ['Statement'],
+            properties: {
+                Version: { type: 'string' },
+                Statement: {
+                    type: 'array',
+                    items: {
+                        allOf: [{
+                                type: 'object',
+                                required: ['Effect'],
+                                properties: {
+                                    Sid: {
+                                        type: 'string'
+                                    },
+                                    Action: {
+                                        $ref: '#/definitions/string_or_string_array'
+                                    },
+                                    NotAction: {
+                                        $ref: '#/definitions/string_or_string_array'
+                                    },
+                                    Resource: {
+                                        $ref: '#/definitions/string_or_string_array'
+                                    },
+                                    NotResource: {
+                                        $ref: '#/definitions/string_or_string_array'
+                                    },
+                                    Effect: {
+                                        enum: ['Allow', 'Deny'],
+                                        type: 'string'
+                                    },
+                                }
+                            },
+                            // see the comment in bucket_policy about these schemas
+                            // here we removed the Principal / NotPrincipal schemas
+                            {
+                                oneOf: [{
+                                        type: 'object',
+                                        required: ["Action"],
+                                        additionalProperties: true,
+                                        properties: {}
+                                    },
+                                    {
+                                        type: 'object',
+                                        required: ["NotAction"],
+                                        additionalProperties: true,
+                                        properties: {}
+                                    }
+                                ],
+                            },
+                            {
+                                oneOf: [{
+                                        type: 'object',
+                                        required: ["Resource"],
+                                        additionalProperties: true,
+                                        properties: {}
+                                    },
+                                    {
+                                        type: 'object',
+                                        required: ["NotResource"],
+                                        additionalProperties: true,
+                                        properties: {}
+                                    }
+                                ],
+                            },
+                        ]
+                    }
+                },
+            }
+        },
+
+        iam_user_policy: {
+            type: 'object',
+            required: ['policy_name', 'policy_document'],
+            properties: {
+                policy_name: { type: 'string' },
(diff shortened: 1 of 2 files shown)


liranmauda (approved):
LGTM

src/api/common_api.js
```diff
@@ -487,6 +487,90 @@ module.exports = {
             }
         },
 
+        // based on bucket policy without Principal and NotPrincipal since are not used in inline policies
+        // removed the Condition as we don't support it yet
+        iam_user_policy_document: {
```
naveenpaul1, line 492: Will it work if bucket and IAM use same schema and blocking the principle for IAM in code?
Do we need to make changes in two places if there are any changes in future?

shirady: I'm reusing the same elements in the schema of bucket policy (with minor changes).
The schema allows us to avoid saving in a DB something that we know it is not in the structure (Principal, or currently not supported - Conditions for example).

What I had in mind is that In IAM user inline policy from what I checked in AWS there are no validations (expect the structure), so now we don't need to go over the policy in code again, and would enjoy it from the schema check when putting a user policy.

shirady: I thought about my explanation again - 
`Principal` is required in `bucket_policy` schema, hence I cannot `ref` it directly.

I will add a demonstration:

With this change:
```
        iam_user_policy: {
            type: 'object',
            required: ['policy_name', 'policy_document'],
            properties: {
                policy_name: { type: 'string' },
                policy_document: {
                    $ref: 'common_api#/definitions/bucket_policy',
                }
            }
        },
```

This policy:
```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:GetBucketLocation"
            ],
            "Resource": "*"
        }
    ]
}
```
When trying to put user policy (without `Principal`) - adding partial logs:

```
Nov-5 15:45:04.098 [Endpoint/13]  [WARN] core.util.postgres_client:: INVALID_SCHEMA_DB accounts ERRORS [
  { instancePath: '/iam_user_policies/0/policy_document/Statement/0', schemaPath: '#/properties/Statement/items/allOf/1/oneOf/0/required', keyword: 'required', params: { missingProperty: 'Principal' }, message: "must have required property 'Principal'", schema: [ 'Principal', [length]: 1 ], parentSchema: { type: 'object', required: [ 'Principal', [length]: 1 ], additionalProperties: true, properties: {} }, data: { Action: [ 's3:GetBucketLocation', [length]: 1 ], Effect: 'Allow', Resource: '*' } },
  { instancePath: '/iam_user_policies/0/policy_document/Statement/0', schemaPath: '#/properties/Statement/items/allOf/1/oneOf/1/required', keyword: 'required', params: { missingProperty: 'NotPrincipal' }, message: "must have required property 'NotPrincipal'", schema: [ 'NotPrincipal', [length]: 1 ], parentSchema: { type: 'object', required: [ 'NotPrincipal', [length]: 1 ], additionalProperties: true, properties: {} }, data: { Action: [ 's3:GetBucketLocation', [length]: 1 ], Effect: 'Allow', Resource: '*' } },
  {
    instancePath: '/iam_user_policies/0/policy_document/Statement/0',
    schemaPath: '#/properties/Statement/items/allOf/1/oneOf',
    keyword: 'oneOf',
    params: { passingSchemas: null },
    message: 'must match exactly one schema in oneOf',
    schema: [ { type: 'object', required: [ 'Principal', [length]: 1 ], additionalProperties: true, properties: {} }, { type: 'object', required: [ 'NotPrincipal', [length]: 1 ], additionalProperties: true, properties: {} }, [length]: 2 ],
    parentSchema: {
      oneOf: [ { type: 'object', required: [ 'Principal', [length]: 1 ], additionalProperties: true, properties: {} }, { type: 'object', required: [ 'NotPrincipal', [length]: 1 ], additionalProperties: true, properties: {} }, [length]: 2 ]
    },
    data: { Action: [ 's3:GetBucketLocation', [length]: 1 ], Effect: 'Allow', Resource: '*' }
  },
```

@naveenpaul1 WDYT?

naveenpaul1: Ya, I am good here

naveenpaul1 (approved):

PR #9270, opened 2025-11-06
IAM | User Inline Policy Implementation

shirady:
### Describe the Problem
Add the implementation for the user inline policy.
Note: At this point, we don't use this policy; it is just stored in the DB.

### Explain the Changes
1. Add the implementation for `put_user_policy`, `get_user_policy`, `delete_user_policy`, `list_user_policies`.
2. Fix `iam_list_user_policies` so it will show the members by user policy name.
3. In `delete_user` add the step `_check_if_user_does_not_have_resources_before_deletion` - partial copy from accountspace_fs, only the part on IAM users (without the case of root accounts deletion), add there access key and user policy as resources that should be checked.
4. Remove the test `test_accountspace_nb.test.js` as the plan it to add a layer in design.

### Issues: 
list of GAPS:
- Added the implementation to the current design. The plan is to add a layer to the flow: accountspaceNB -> **account_server** -> DB (writing to the DB occurs through the system store).
- The tests are only manual at this point; there is a plan to add automated tests.
- There are a couple of places where the `id` was sent an `ObjectId` instead of a string, which would need to track all those cases, as it defects the detailed error message (prints Object object).

### Testing Instructions:
3. I'm using port-forward (in a different tab): `kubectl port-forward -n test1 service/iam 12443:443`
4. Create the alias for the admin - first, need to get the credentials: `nb status --show-secrets -n test1` and then `alias user-1-iam='AWS_ACCESS_KEY=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:12443'`
5. Check the connection to the endpoint and try to list the users (should be an empty list): `user-1-iam iam list-users; echo $?`
6. Create a user: `user-1-iam iam create-user --user-name Robert`
Note: To validate user creation, you can rerun `user-1-iam iam list-users` and expect 1 user in the list
7. Run user policy API commands using AWS CLI:
- `user-1-iam iam put-user-policy --user-name Robert --policy-name mypolicy --policy-document file://policy_01.json`
An example of a policy that was used:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:GetBucketLocation"
            ],
            "Resource": "*"
        }
    ]
}
```
- `user-1-iam iam get-user-policy --user-name Robert --policy-name mypolicy`
- `user-1-iam iam list-user-policies --user-name Robert`
- `user-1-iam iam delete-user-policy --user-name Robert --policy-name mypolicy`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Full IAM user policy management: create, retrieve, update, delete, and sorted listing.
  * New policy-related action grouping and an inline policy size limit.

* **Bug Fixes**
  * Policy name listing payload shape changed to object entries.
  * Safer handling when access keys are absent; improved deletion-conflict and missing-policy errors and logging.
  * Pre-deletion checks block account/user deletion if resources, keys, or policies exist.
  * Policy document returned consistently as JSON where applicable.

* **Tests**
  * Removed obsolete tests for previously unimplemented policy methods.


src/sdk/accountspace_nb.js
```diff
@@ -440,27 +441,77 @@ class AccountSpaceNB {
...
+        const updated_user_policies = iam_user_policies.splice(iam_user_policy_index, 1);
+
+        await system_store.make_changes({
+            update: {
+                accounts: [{
+                    _id: requested_account._id,
+                    $updated_user_policies: updated_user_policies,
+                }]
+            }
+        });
```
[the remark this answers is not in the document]
shirady: Why didn't you resolve this?

src/sdk/accountspace_nb.js
```diff
@@ -440,27 +441,77 @@ class AccountSpaceNB {
...
+        dbg.log1(`AccountSpaceNB.${action}`, params);
+        const requesting_account = system_store.get_account_by_email(account_sdk.requesting_account.email);
+        const requested_account = validate_and_return_requested_account(params, action, requesting_account, account_sdk);
+        const iam_user_policies = requested_account.iam_user_policies || [];
+        const iam_user_policy_index = account_util._check_user_policy_exists(action, iam_user_policies, params.policy_name);
+        return {
+            username: params.username,
+            policy_name: params.policy_name,
+            policy_document: JSON.stringify(iam_user_policies[iam_user_policy_index].policy_document),
+        };
```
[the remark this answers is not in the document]
shirady: Are you sure?
The original `policy_document` was a string, and we parsed it and then stored it in the DB.
We also mentioned in the schema that it is an object.
In what case exactly will I have it as a plain string?

src/sdk/accountspace_nb.js
```diff
@@ -440,27 +441,77 @@ class AccountSpaceNB {
...
+        const iam_user_policies = requested_account.iam_user_policies || [];
+        const index_of_iam_user_policy = account_util._get_iam_user_policy_index(iam_user_policies, params.policy_name);
+        const iam_user_policy_to_add = {
+            policy_name: params.policy_name,
+            policy_document: params.policy_document,
+        };
+        if (index_of_iam_user_policy === -1) {
+            iam_user_policies.push(iam_user_policy_to_add);
+        } else {
+            iam_user_policies[index_of_iam_user_policy] = iam_user_policy_to_add;
```
naveenpaul1, line 457: Are we updating the existing policy? I think we should not update existing one. WDYT? Could  you please check AWS behaviour.

src/util/account_util.js
```diff
@@ -459,7 +465,7 @@ function _throw_access_denied_error(action, requesting_account, details, entity)
     let message_with_details;
     if (entity === 'USER') {
         let user_message;
-        if (action === IAM_ACTIONS.LIST_ACCESS_KEYS) {
+        if (action === IAM_ACTIONS.LIST_ACCESS_KEYS || action.includes('policy') || action.includes('policies')) {
```
naveenpaul1: Can you add this to constants.

src/util/account_util.js
```diff
@@ -569,6 +583,44 @@ function _list_access_keys_from_account(requesting_account, account, on_itself)
...
+function _check_if_user_does_not_have_resources_before_deletion(action, account_to_delete) {
+    _check_if_user_does_not_have_access_keys_before_deletion(action, account_to_delete);
+    _check_if_user_does_not_have_user_policy_before_deletion(action, account_to_delete);
+}
+
+function _check_if_user_does_not_have_access_keys_before_deletion(action, account_to_delete) {
+    const resource_name = 'access keys';
+    const access_keys = account_to_delete.access_keys || [];
+    const is_access_keys_removed = access_keys.length === 0;
+    if (!is_access_keys_removed) {
```
naveenpaul1: `is_access_keys_removed` is confusing.
Can you change to something like `is_access_key_empty`

src/test/unit_tests/internal/test_accountspace_nb.test.js
```diff
@@ -1,64 +0,0 @@
-/* Copyright (C) 2025 NooBaa */
-/* eslint-disable max-lines-per-function */
```
naveenpaul1, line 2: why this file removed?

src/sdk/accountspace_nb.js
shirady: This is how AWS behaves - here are the steps:
1. Create a user: `aws iam create-user --user-name shira-check03`
2. Put user inline policy: `aws iam put-user-policy --user-name shira-check03 --policy-name mypolicy --policy-document file://policy_01.json`
3. Get the user inline policy: `aws iam get-user-policy --user-name shira-check03 --policy-name mypolicy`, here is the output:
```
{
    "UserName": "shira-check03",
    "PolicyName": "mypolicy",
    "PolicyDocument": {
        "Version": "2012-10-17",
        "Statement": [
            {
                "Effect": "Allow",
                "Action": [
                    "s3:GetBucketLocation"
                ],
                "Resource": "*"
            }
        ]
    }
}
```
4. Put user inline policy - same policy name, but different policy document: `aws iam put-user-policy --user-name shira-check03 --policy-name mypolicy --policy-document file://policy_02.json`
5. Get the user inline policy: `aws iam get-user-policy --user-name shira-check03 --policy-name mypolicy`, here is the output:
```
{
    "UserName": "shira-check03",
    "PolicyName": "mypolicy",
    "PolicyDocument": {
        "Version": "2012-10-17",
        "Statement": [
            {
                "Effect": "Allow",
                "Action": [
                    "s3:HeadBucket"
                ],
                "Resource": "*"
            }
        ]
    }
}
```
7. LIst the user inline policies: `aws iam list-user-policies --user-name shira-check03`
```
{
    "PolicyNames": [
        "mypolicy"
    ]
}
```
You can see that the user has one policy and the content is the last policy that was added (action of `HeadBucket` and not `GetBucketLocation` as it was at first).

I hope it is clear, if not please ask.

src/util/account_util.js
shirady: I changed it, thanks.

src/util/account_util.js
shirady: I changed it, thanks.

src/test/unit_tests/internal/test_accountspace_nb.test.js
shirady: In the future design the accountspace_nb would probably just pass the parameters to the account server.

naveenpaul1 (approved):

src/util/account_util.js
```diff
@@ -569,6 +585,63 @@ function _list_access_keys_from_account(requesting_account, account, on_itself)
...
+
+// each char is  byte and not including whitespaces
+function _get_total_size_of_policies(iam_user_policies) {
+    let total_size = 0;
+    for (const iam_user_policy of iam_user_policies) {
+        const policy_as_string = JSON.stringify(iam_user_policy);
+        total_size += policy_as_string.length;
+    }
+    return total_size;
+}
```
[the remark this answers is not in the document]
shirady: Are you sure?
I read the document of AWS - see this [link](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html):

> You can add as many inline policies as you want to an IAM user, role, or group. But the total aggregate policy size (the sum size of all inline policies) per entity can't exceed the following limits:
> User policy size can't exceed 2,048 characters.

I also tested it by adding a policy with ~2000 chars and then adding the another policy with the same file and it failed with `LimitExceeded`.

PR #9271, opened 2025-11-06
NC | Adding support of user bucket path

jackyalbo:
### Describe the Problem
Adding support for an out-of-S3 header to let the user control where the bucket is supposed to be created.
x-noobaa-custom-bucket-path - will point to where the user wants their new bucket to be saved in the file system.

### Explain the Changes
1.  Adapting APIs to support x-noobaa-custom-bucket-path http header - once provided, instead of using the account new_bucket_path directory added to the bucket name, the name in the header will be used.
2.  Accounts can define an allow-list for permitted custom bucket paths; CLI options added to set or unset this allow-list.
3. Adding a test to check this scenario.

config.NSFS_CUSTOM_BUCKET_PATH_ALLOWED_LIST = '/ibm1/:/ibm3/'
or as part of the account, nsfs_account_config, for example:

nsfs_account_config: {
        uid: 100,
        gid: 100,
        new_buckets_path: '/gpfs/new_buckets',
        nsfs_only: false,
        custom_bucket_path_allowed_list: `/gpfs1/:/gpfs2/`,
}
If it exists, the account's custom_bucket_path_allowed_list will override the system's one.
If neither does, we will drop the request with 'access denied'.
1. 

### Testing Instructions:
1. Apply `custom_bucket_path_allowed_list` in config.js and/or in your account
2. Send `x-noobaa-custom-bucket-path` header as part of S3 put_bucket request (using account creds)
3. Verify that the bucket was created in the provided path and not under the account's new_buckets_path/bucket_name

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Bucket creation can specify a custom storage path via a special HTTP header.
  * Accounts can define an allow-list of permitted custom bucket path prefixes and a CLI flag to set/unset it.
  * Public configuration entries added to expose the header name and the allow-list setting.

* **Tests**
  * Integration tests added for allowed/denied creation, existing-path conflicts, and account-based path scenarios.

* **Documentation**
  * CLI and account docs updated to describe the new flag and behavior.

### DIFF
--- a/config.js
+++ b/config.js
@@ -1022,6 +1022,9 @@ config.NSFS_LIST_IGNORE_ENTRY_ON_EACCES = true;
 // we will for now handle the same way also EINVAL error - for gpfs stat issues on list (.snapshots)
 config.NSFS_LIST_IGNORE_ENTRY_ON_EINVAL = true;
 
+config.NSFS_CUSTOM_BUCKET_PATH_HTTP_HEADER = 'x-noobaa-custom-bucket-path';
+config.NSFS_CUSTOM_BUCKET_PATH_ALLOWED_LIST = ''; // colon separated list of paths prefixes
+
 ////////////////////////////
 // NSFS NON CONTAINERIZED //
 ////////////////////////////
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -303,9 +303,19 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             const fs_context = this.prepare_fs_context(sdk);
             validate_bucket_creation(params);
 
-            const { name } = params;
+            const { name, custom_bucket_path } = params;
             const bucket_config_path = this.config_fs.get_bucket_path_by_name(name);
-            const bucket_storage_path = path.join(sdk.requesting_account.nsfs_account_config.new_buckets_path, name);
+            if (custom_bucket_path) {
+                const allowed_list = sdk.requesting_account.nsfs_account_config.custom_bucket_path_allowed_list ||
+                    config.NSFS_CUSTOM_BUCKET_PATH_ALLOWED_LIST;
+                const allowed_path_prefixes = allowed_list ? allowed_list.split(':').map(p => p.trim()).filter(p => p) : [];
+                if (!allowed_path_prefixes.length || !allowed_path_prefixes.some(prefix => custom_bucket_path.startsWith(prefix))) {
+                    const message = `Not allowed to create new buckets: ${custom_bucket_path} outside of the custom_bucket_path_allowed_list: ${allowed_list}`;
+                    dbg.error(`BucketSpaceFS.create_bucket: ${message}`);
+                    throw new RpcError('UNAUTHORIZED', message);
+                }
+            }
+            const bucket_storage_path = custom_bucket_path || path.join(sdk.requesting_account.nsfs_account_config.new_buckets_path, name);
 
             dbg.log0(`BucketSpaceFS.create_bucket
                 requesting_account=${util.inspect(sdk.requesting_account)},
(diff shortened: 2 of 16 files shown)


config.js
```diff
@@ -1016,6 +1016,8 @@ config.NSFS_LIST_IGNORE_ENTRY_ON_EACCES = true;
 // we will for now handle the same way also EINVAL error - for gpfs stat issues on list (.snapshots)
 config.NSFS_LIST_IGNORE_ENTRY_ON_EINVAL = true;
 
+config.NSFS_USER_BUCKET_PATH_HTTP_HEADER = 'x-nsfs-bucket-path';
```
guymguym: @jackyalbo I would use x-noobaa- as a prefix for anything that is noobaa specific
```suggestion
config.NSFS_USER_BUCKET_PATH_HTTP_HEADER = 'x-noobaa-custom-bucket-path';
```

src/sdk/bucketspace_fs.js
```diff
@@ -305,7 +305,8 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
 
             const { name } = params;
             const bucket_config_path = this.config_fs.get_bucket_path_by_name(name);
-            const bucket_storage_path = path.join(sdk.requesting_account.nsfs_account_config.new_buckets_path, name);
+            const bucket_storage_path = params.user_bucket_path ||
+                path.join(sdk.requesting_account.nsfs_account_config.new_buckets_path, name);
```
guymguym: First - the term "user_bucket_path" is confusing - the user has a new_buckets_path where buckets gets created by default. So in this new feature we are actually allowing to override that path, which I would name as "custom_bucket_path".

Second - I think we need some minimal protection for the custom path - Otherwise the request might specify very sensitive paths in the host such as /etc or /root. We need to have some whitelist of paths that the custom_bucket_path is required to be a subpath of, this can be a config such as:
```
config.NSFS_CUSTOM_BUCKET_PATHS_WHITELIST = ['/ibm', '/gpfs']
```

I am even contemplating if we can restrict the custom paths to always be a subpath of the account's new_buckets_path to allow more flexibility in restricting every account to a specific subtree of the filesystem.

FYI @madhuthorat @jackyalbo @nimrod-becker

jackyalbo: We decided to go with two levels of allowed_bucket_paths. per system, using config.js:
`config.NSFS_CUSTOM_BUCKET_PATH_ALLOWED_LIST = '/ibm1/:/ibm3/'`
or as part of the account, nsfs_account_config, for example:
```
nsfs_account_config: {
        uid: 100,
        gid: 100,
        new_buckets_path: '/gpfs/new_buckets',
        nsfs_only: false,
        custom_bucket_path_allowed_list: `/gpfs1/:/gpfs2/`,
}
```
If it exists, the account's custom_bucket_path_allowed_list will override the system's one.
If neither does, we will drop the request with 'access denied'.
@guymguym @madhuthorat for any last feedback

nimrod-becker:
@nadavMiz if you could take a look please

src/test/integration_tests/nc/cli/test_nc_account_cli.test.js
```diff
@@ -657,6 +657,21 @@ describe('manage nsfs cli account flow', () => {
...
+            const account_options = { config_root, name, new_buckets_path, uid, gid, custom_bucket_path_allowed_list };
+            await fs_utils.create_fresh_path(new_buckets_path);
+            await fs_utils.file_must_exist(new_buckets_path);
+            await set_path_permissions_and_owner(new_buckets_path, account_options, 0o700);
+            await exec_manage_cli(type, action, account_options);
+            const account = await config_fs.get_account_by_name(name, config_fs_account_options);
+            console.log('account=', account);
+            assert_account(account, account_options, false);
+            expect(account.nsfs_account_config.custom_bucket_path_allowed_list).toBe(custom_bucket_path_allowed_list);
+        });
```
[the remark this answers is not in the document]
jackyalbo: cool. thanks

nadavMiz (approved):
LGTM

Issue #8931, opened 2025-04-02
Data Buckets | List Buckets Can Return Buckets in Deleting State


shirady:
### Environment info
- NooBaa Version: 5.19
- Platform: Happened in the operator CI

### Actual behavior
1. In the CI we sometimes have buckets in deleting state.

### Expected behavior
1. We should not have buckets in deleting state.

### Steps to reproduce
1. It runs as a part of the operator CI.

### More information - Screenshots / Logs / Other output
This a [link](https://github.com/noobaa/noobaa-operator/actions/runs/14154326946/job/39651579645?pr=1570) example of a failed run without code changes (below the downloaded file).
You can see:
```
09:06:05 💬 Deleting buckets
2025-03-30 09:05:52 multinamespace-obc-deleting-1743325553731
2025-03-30 09:05:43 multinamespace-obc-deleting-1743325546281
2025-03-30 08:49:31 second.bucket
2025-03-30 08:54:24 failns.bucket1
2025-03-30 08:45:05 first.bucket
2025-03-30 08:53:48 target.bucket2
2025-03-30 08:53:46 target.bucket1
2025-03-30 08:53:43 fails3.bucket
```
We see buckets whose names contain `deleting-<timestamp>`.

And the failure comes from:
```
remove_bucket failed: Unable to delete all objects in the bucket, bucket will not be deleted.
09:06:07 ❌  aws s3 rb s3://multinamespace-obc-deleting-1743325553731 --force failed, Exiting
```

The downloaded file: [0_run-cli-tests.txt](https://github.com/user-attachments/files/19550729/0_run-cli-tests.txt)

PR #9272, opened 2025-11-07
Fix - list_buckets allowing unauthorized bucket access

aayushchouhan09:
### Describe the Problem
The `list_buckets()` function was using `.filter()` with an async function, which doesn't wait for promises to resolve. This caused all buckets to pass through the filter ingnoring the actual permissions and allowing users to see all buckets in the system instead of only their authorized buckets.

### Explain the Changes
1. Handle system owner and bucket owner permissions while listing buckets.
2. Special case handled for obc check.
3. Special case for noobaa cli.

**Note** : Previously, we were allowing bucket policy to interfere with the list bucket operations which is not aws complaint.

1. Issue (fixed partially for containerized): https://issues.redhat.com/browse/DFBUGS-4449
2. Fixed issue: #8931 

(Gaps)
3. Gap: Handle IAM user (small fix)
4. Related PR (gap): #9301 

### Testing Instructions:

(admin test)
1. Try listing buckets with admin/system credentials, it should lists all the buckets.

(owner test)
5. Create an account (in case of OBC create OBC bucket and it will create OBC account)
6. Create bucket for the account (not for obc case)
7. Create S3 client using account credentials.
8. Try to list the buckets:
      `aws s3 ls` --> It should list only the authorized buckets for the account.

(NooBaa cli)
9. Use noobaa cli to list buckets `noobaa list buckets` --> it should list all the buckets

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Bucket listing now excludes deleting buckets and shows users only buckets they own; admins still see all buckets.
  * Permission checks updated to consider multiple ownership types (system, direct, claim, IAM-root) for more accurate access decisions.

* **Tests**
  * Added integration tests verifying per-account bucket isolation and admin visibility; updated session-token tests to assert specific bucket inclusion.

### DIFF
--- a/src/server/common_services/auth_server.js
+++ b/src/server/common_services/auth_server.js
@@ -496,6 +496,10 @@ function _prepare_auth_request(req) {
     req.has_bucket_action_permission = async function(bucket, action, bucket_path, req_query) {
         return has_bucket_action_permission(bucket, req.account, action, req_query, bucket_path);
     };
+
+    req.has_bucket_ownership_permission = function(bucket) {
+        return has_bucket_ownership_permission(bucket, req.account, req.auth && req.auth.role);
+    };
 }
 
 function _get_auth_info(account, system, authorized_by, role, extra) {
@@ -521,6 +525,73 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
 }
 
 /**
+ * is_system_owner checks if the account is the system owner
+ * @param {Record<string, any>} bucket
+ * @param {Record<string, any>} account
+ * @returns {boolean}
+ */
+function is_system_owner(bucket, account) {
+    if (!bucket?.system?.owner?.email || !account?.email) return false;
+    return bucket.system.owner.email.unwrap() === account.email.unwrap();
+}
+
+/**
+ * is_bucket_owner checks if the account is the direct owner of the bucket
+ * @param {Record<string, any>} bucket
+ * @param {Record<string, any>} account
+ * @returns {boolean}
+ */
+function is_bucket_owner(bucket, account) {
+    if (!bucket?.owner_account?.email || !account?.email) return false;
+    return bucket.owner_account.email.unwrap() === account.email.unwrap();
+}
+
+/**
+ * is_bucket_claim_owner checks if the account is the OBC (ObjectBucketClaim) owner of the bucket
+ * @param {Record<string, any>} bucket
+ * @param {Record<string, any>} account
+ * @returns {boolean}
+ */
+function is_bucket_claim_owner(bucket, account) {
+    if (!account?.bucket_claim_owner || !bucket?.name) return false;
+    return account.bucket_claim_owner.name.unwrap() === bucket.name.unwrap();
+}
+
+/**
+ * has_bucket_ownership_permission returns true if the account can list the bucket in ListBuckets operation
+ *
+ * aws-compliant behavior:
+ * - System owner can list all the buckets
+ * - Operator account (noobaa cli) can list all the buckets
+ * - Root accounts can list buckets they own
+ * - OBC owner can list their buckets
+ * - IAM users can list their owner buckets
+ *
+ * @param {Record<string, any>} bucket
+ * @param {Record<string, any>} account
+ * @param {string} role
+ * @returns {Promise<boolean>}
+ */
+async function has_bucket_ownership_permission(bucket, account, role) {
+    // system owner can list all the buckets
+    if (is_system_owner(bucket, account)) return true;
+
+    // operator account (noobaa cli) can list all the buckets
+    if (role === 'operator') return true;
+
+    // check direct ownership
+    if (is_bucket_owner(bucket, account)) return true;
+
+    // special case: check bucket claim ownership (OBC)
+    if (is_bucket_claim_owner(bucket, account)) return true;
+
+    // special case: iam user can list the buckets of their owner
+    // TODO: handle iam user
+
+    return false;
+}
+
+/**
  * has_bucket_action_permission returns true if the requesting account has permission to perform
  * the given action on the given bucket.
  *
@@ -538,19 +609,21 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
  */
 async function has_bucket_action_permission(bucket, account, action, req_query, bucket_path = "") {
     dbg.log1('has_bucket_action_permission:', bucket.name, account.email, bucket.owner_account.email);
-    // If the system owner account wants to access the bucket, allow it
-    if (bucket.system.owner.email.unwrap() === account.email.unwrap()) return true;
 
-    const is_owner = (bucket.owner_account.email.unwrap() === account.email.unwrap()) ||
-        (account.bucket_claim_owner && account.bucket_claim_owner.name.unwrap() === bucket.name.unwrap());
+    // system owner can access all buckets
+    if (is_system_owner(bucket, account)) return true;
+
+    // check ownership: direct owner or OBC
+    const has_owner_access = is_bucket_owner(bucket, account) || is_bucket_claim_owner(bucket, account);
+
     const bucket_policy = bucket.s3_policy;
 
     if (!bucket_policy) {
         // in case we do not have bucket policy
-        // we allow IAM account to access a bucket that that is owned by their root account
+        // we allow IAM account to access a bucket that is owned by their root account
         const is_iam_and_same_root_account_owner = account.owner !== undefined &&
             account.owner._id.toString() === bucket.owner_account._id.toString();
-        return is_owner || is_iam_and_same_root_account_owner;
+        return has_owner_access || is_iam_and_same_root_account_owner;
     }
     if (!action) {
         throw new Error('has_bucket_action_permission: action is required');
@@ -566,7 +639,8 @@ async function has_bucket_action_permission(bucket, account, action, req_query, 
     );
 
     if (result === 'DENY') return false;
-    return is_owner || result === 'ALLOW';
+
+    return has_owner_access || result === 'ALLOW';
 }
 
 /**
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -1096,9 +1096,13 @@ async function list_buckets(req) {
     let continuation_token = req.rpc_params?.continuation_token;
     const max_buckets = req.rpc_params?.max_buckets;
 
-    const accessible_bucket_list = system_store.data.buckets.filter(
-        async bucket => await req.has_s3_bucket_permission(bucket, "s3:ListBucket", req) && !bucket.deleting
-    );
+    // filter buckets based on ownership
+    const bucket_permissions = await P.map(system_store.data.buckets, async bucket => {
+        if (bucket.deleting) return null;
+        const has_permission = await req.has_bucket_ownership_permission(bucket);
+        return has_permission ? bucket : null;
+    });
+    const accessible_bucket_list = bucket_permissions.filter(bucket => bucket !== null);
 
     accessible_bucket_list.sort((a, b) => a.name.unwrap().localeCompare(b.name.unwrap()));
 
--- a/src/test/integration_tests/api/s3/test_s3_list_buckets.js
+++ b/src/test/integration_tests/api/s3/test_s3_list_buckets.js
@@ -127,5 +127,53 @@ mocha.describe('s3_ops', function() {
         });
     });
 
+    mocha.describe('list_buckets permissions', function() {
+        this.timeout(60000);
+        let s3_account_a;
+        let s3_account_b;
+
+        async function create_account_and_client(name) {
+            const account = await rpc_client.account.create_account({
+                name, email: name, has_login: false, s3_access: true,
+                default_resource: coretest.POOL_LIST[0].name
+            });
+            return new S3Client({
+                ...s3_client_params,
+                credentials: {
+                    accessKeyId: account.access_keys[0].access_key.unwrap(),
+                    secretAccessKey: account.access_keys[0].secret_key.unwrap(),
+                }
+            });
+        }
+
+        mocha.before(async function() {
+            s3_account_a = await create_account_and_client('account-a');
+            s3_account_b = await create_account_and_client('account-b');
+            await s3_account_a.send(new CreateBucketCommand({ Bucket: 'bucket-a' }));
+            await s3_account_b.send(new CreateBucketCommand({ Bucket: 'bucket-b' }));
+            await s3.send(new CreateBucketCommand({ Bucket: 'admin-buck' }));
+        });
+
+        mocha.after(async function() {
+            await s3_account_a.send(new DeleteBucketCommand({ Bucket: 'bucket-a' }));
+            await s3_account_b.send(new DeleteBucketCommand({ Bucket: 'bucket-b' }));
+            await s3.send(new DeleteBucketCommand({ Bucket: 'admin-buck' }));
+            await rpc_client.account.delete_account({ email: 'account-a' });
+            await rpc_client.account.delete_account({ email: 'account-b' });
+        });
+
+        mocha.it('accounts should list only owned buckets', async function() {
+            const buckets_a = (await s3_account_a.send(new ListBucketsCommand())).Buckets.map(b => b.Name);
+            const buckets_b = (await s3_account_b.send(new ListBucketsCommand())).Buckets.map(b => b.Name);
+            assert.deepStrictEqual(buckets_a, ['bucket-a']);
+            assert.deepStrictEqual(buckets_b, ['bucket-b']);
+        });
+
+        mocha.it('admin should lists all the buckets', async function() {
+            const buckets = (await s3.send(new ListBucketsCommand())).Buckets.map(b => b.Name);
+            assert(buckets.includes('bucket-a') && buckets.includes('bucket-b') && buckets.includes('admin-buck'));
+        });
+    });
+
 });
 
--- a/src/test/integration_tests/api/sts/test_sts.js
+++ b/src/test/integration_tests/api/sts/test_sts.js
@@ -457,6 +457,7 @@ function verify_session_token(session_token, access_key, secret_key, assumed_rol
 mocha.describe('Session token tests', function() {
     const { rpc_client } = coretest;
     const alice2 = 'alice2';
+    const alice2_buck = 'alice2-test-bucket';
     const bob2 = 'bob2';
     const charlie2 = 'charlie2';
     const accounts = [{ email: alice2 }, { email: bob2 }, { email: charlie2 }];
@@ -466,6 +467,8 @@ mocha.describe('Session token tests', function() {
     mocha.after(async function() {
         const self = this; // eslint-disable-line no-invalid-this
         self.timeout(60000);
+
+        await accounts[0].s3.deleteBucket({ Bucket: alice2_buck }).promise();
         for (const account of accounts) {
             await rpc_client.account.delete_account({ email: account.email });
         }
@@ -542,6 +545,10 @@ mocha.describe('Session token tests', function() {
             name: 'first.bucket',
             policy: s3accesspolicy,
         });
+
+        // create a bucket owned by alice2 for ListBuckets to work
+        // Note: bucket policy is not related to ListBuckets operation
+        await accounts[0].s3.createBucket({ Bucket: alice2_buck }).promise();
     });
 
     mocha.it('user b assume role of user a - default expiry - list s3 - should be allowed', async function() {
@@ -564,7 +571,7 @@ mocha.describe('Session token tests', function() {
         });
 
         const buckets1 = await temp_s3_with_session_token.listBuckets().promise();
-        assert.ok(buckets1.Buckets.length > 0);
+        assert.ok(buckets1.Buckets[0].Name === alice2_buck);
     });
 
     mocha.it('user b assume role of user a - valid expiry via durationSeconds - list s3 - should be allowed', async function() {
@@ -589,7 +596,7 @@ mocha.describe('Session token tests', function() {
         });
 
         const buckets1 = await temp_s3_with_session_token.listBuckets().promise();
-        assert.ok(buckets1.Buckets.length > 0);
+        assert.ok(buckets1.Buckets[0].Name === alice2_buck);
     });
 
     mocha.it('user b assume role of user a - invalid expiry via durationSeconds - should be rejected', async function() {
--- a/src/test/integration_tests/nsfs/test_nsfs_integration.js
+++ b/src/test/integration_tests/nsfs/test_nsfs_integration.js
@@ -249,14 +249,12 @@ mocha.describe('bucket operations - namespace_fs', function() {
         s3_wrong_uid = new S3(s3_creds);
     });
     mocha.it('list buckets with wrong uid, gid', async function() {
-        // Give s3_wrong_uid access to the required buckets
-        const s3_policy = generate_s3_policy('*', first_bucket, ['s3:*']);
-        await rpc_client.bucket.put_bucket_policy({ name: first_bucket, policy: s3_policy.policy });
-
+        await s3_wrong_uid.createBucket({ Bucket: 'bucket-wrong-uid' });
         const res = await s3_wrong_uid.listBuckets({});
         console.log(inspect(res));
-        const list_ok = bucket_in_list([first_bucket], [bucket_name], res.Buckets);
+        const list_ok = bucket_in_list(['bucket-wrong-uid'], [], res.Buckets);
         assert.ok(list_ok);
+        await s3_wrong_uid.deleteBucket({ Bucket: 'bucket-wrong-uid' });
     });
     mocha.it('update account', async function() {
         this.timeout(600000); // eslint-disable-line no-invalid-this
@@ -390,6 +388,17 @@ mocha.describe('bucket operations - namespace_fs', function() {
     });
 
     mocha.it('get bucket acl - rpc bucket', async function() {
+        // Grant s3_correct_uid_default_nsr permission to get bucket ACL on first_bucket
+        let account_principal;
+        if (is_nc_coretest) {
+            account_principal = 'account_s3_correct_uid';
+        } else {
+            const account_info = await rpc_client.account.read_account({ email: 'account_s3_correct_uid@noobaa.com' });
+            account_principal = iam_utils.create_arn_for_root(account_info._id.toString());
+        }
+        const generated = generate_s3_policy(account_principal, first_bucket, ['s3:GetBucketAcl']);
+        await rpc_client.bucket.put_bucket_policy({ name: first_bucket, policy: generated.policy });
+
         const res = await s3_correct_uid_default_nsr.getBucketAcl({ Bucket: first_bucket });
         const bucket_info = await rpc_client.bucket.read_bucket({ name: first_bucket });
         assert.equal(res.Owner.DisplayName, bucket_info.owner_account.email);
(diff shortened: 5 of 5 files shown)


src/server/common_services/auth_server.js
```diff
@@ -536,7 +536,12 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
  * @returns  {Promise<boolean>} true if the account has permission to perform the action on the bucket
  */
 async function has_bucket_action_permission(bucket, account, action, req_query, bucket_path = "") {
+    if (!account || !account.email) {
+        dbg.warn('has_bucket_action_permission: account or account.email is missing');
+        return false;
+    }
```
shirady: What was the case you saw where this happens?

src/server/system_services/bucket_server.js
```diff
@@ -1062,8 +1062,13 @@ async function list_buckets(req) {
...
     const max_buckets = req.rpc_params?.max_buckets;
 
-    const accessible_bucket_list = system_store.data.buckets.filter(
-        async bucket => await req.has_s3_bucket_permission(bucket, "s3:ListBucket", req) && !bucket.deleting
+    const accessible_bucket_list = [];
+    await Promise.all(
+        system_store.data.buckets.map(async bucket => {
+            if (bucket.deleting) return;
+            const has_permission = await req.has_s3_bucket_permission(bucket, "s3:ListBucket");
+            if (has_permission) accessible_bucket_list.push(bucket);
```
shirady: You're adding an item (`bucket`) to the same data structure (array `accessible_bucket_list`) in parallel, which seems to me not safe (case - race condition).
WDYT?

src/server/system_services/bucket_server.js
shirady, line 1091: Could you document the `list_buckets` behaviour?
As from I understand from the code it is not as AWS behaves.

src/test/integration_tests/api/s3/test_s3_list_buckets.js
```diff
@@ -127,5 +127,78 @@ mocha.describe('s3_ops', function() {
         });
     });
 
+    mocha.describe('list_buckets permissions', function() {
+        this.timeout(60000);
+        let s3_user;
+        const test_user = 'test-user';
```
shirady: Could you change the variable name to `test_account` as this fix is related to accounts and not users.
```suggestion
        const test_account = 'test-account';
```

src/server/system_services/bucket_server.js
```diff
@@ -1062,8 +1062,13 @@ async function list_buckets(req) {
     let continuation_token = req.rpc_params?.continuation_token;
     const max_buckets = req.rpc_params?.max_buckets;
 
-    const accessible_bucket_list = system_store.data.buckets.filter(
-        async bucket => await req.has_s3_bucket_permission(bucket, "s3:ListBucket", req) && !bucket.deleting
+    const accessible_bucket_list = [];
+    await Promise.all(
+        system_store.data.buckets.map(async bucket => {
+            if (bucket.deleting) return;
+            const has_permission = await req.has_s3_bucket_permission(bucket, "s3:ListBucket");
```
shirady: `s3:ListBucket` is the policy for listing objects in a bucket, while you're trying to check permission for the list buckets.

AWS Doc:
**ListObjectsV2** ([link](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html))
>  To use this operation, you must have READ access to the bucket. You must have permission to perform the s3:ListBucket action

**ListBuckets** ([link](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListBuckets.html))

> Returns a list of all buckets owned by the authenticated sender of the request. To grant IAM permission to use this operation, you must add the s3:ListAllMyBuckets policy action.

WDYT?

src/test/integration_tests/api/s3/test_s3_list_buckets.js
```diff
@@ -127,5 +127,78 @@ mocha.describe('s3_ops', function() {
...
+        });
+
+        mocha.it('admin should list all buckets', async function() {
+            const buckets = (await s3.send(new ListBucketsCommand())).Buckets.map(b => b.Name);
+            assert(buckets.length >= 2);
+            assert(buckets.includes(user_bucket));
+            assert(buckets.includes(admin_bucket));
+        });
+
+        mocha.it('bucket policy grants list access', async function() {
```
shirady: I think this test `it` should not appear, see the comment on the bucket policy I added.

src/test/integration_tests/api/s3/test_s3_list_buckets.js
```diff
@@ -127,5 +127,78 @@ mocha.describe('s3_ops', function() {
...
+            let buckets = (await s3_user.send(new ListBucketsCommand())).Buckets.map(b => b.Name);
+            assert(!buckets.includes(admin_bucket));
+
+            await rpc_client.bucket.put_bucket_policy({
+                name: admin_bucket,
+                policy: {
+                    Version: '2012-10-17',
+                    Statement: [{
+                        Effect: 'Allow',
+                        Principal: { AWS: test_user },
```
naveenpaul1: We might need to change this after this [PR](https://github.com/noobaa/noobaa-core/pull/9277)

src/server/common_services/auth_server.js
```diff
@@ -558,7 +563,9 @@ async function has_bucket_action_permission(bucket, account, action, req_query,
     );
 
     if (result === 'DENY') return false;
-    return is_owner || result === 'ALLOW';
+    if (result === 'ALLOW') return true;
```
naveenpaul1: previous condition aslo doing the same thing right? is thera nything wrong with that?

nimrod-becker:
@aayushchouhan09 @shirady what about bucketspace_fs ?

src/server/common_services/auth_server.js
aayushchouhan09: We just added it as a safe check, not for particular case. It might be when account being corrupted by any means.

src/server/common_services/auth_server.js
aayushchouhan09: I put it at the time of debugging, we can move it back as before. Thanks

src/server/system_services/bucket_server.js
aayushchouhan09: Yes, you are correct we should use `LIstAllMyBuckets` instead. Thanks

src/server/system_services/bucket_server.js
aayushchouhan09: Need to update it. Thanks

src/test/integration_tests/api/s3/test_s3_list_buckets.js
aayushchouhan09: I didn't get you, Do you want to remove/update this test?

src/server/system_services/bucket_server.js
shirady: I can suggest that you create a helper function that would fetch the bucket by owner.
It is not related to bucket policy - bucket policy is on a bucket.

src/test/integration_tests/api/s3/test_s3_list_buckets.js
shirady: I don't see the connection between the bucket policy and the ability to list buckets.
You can test it on AWS and check

src/test/integration_tests/api/s3/test_s3_list_buckets.js
```diff
@@ -127,5 +127,78 @@ mocha.describe('s3_ops', function() {
         });
     });
 
+    mocha.describe('list_buckets permissions', function() {
```
shirady, line 130: I would add another account and bucket - to make sure every account sees only its buckets.
- account a - only bucket a
- account b - only bucket b
- admin account - admin bucket, bucket a, and bucket b

src/test/integration_tests/api/s3/test_s3_list_buckets.js
aayushchouhan09: we can remove this tests, you are right bucket policy is not related to list buckets.

src/server/system_services/bucket_server.js
aayushchouhan09: As per my understanding (currently):

1. Returns all the buckets for admin
2. Returns only buckets the account owns
3. NSFS buckets need both ownership + filesystem access
4. Bucket policies do not affect this operation (AWS s3 behaviour)

src/server/system_services/bucket_server.js
```diff
@@ -1095,9 +1095,11 @@ async function list_buckets(req) {
     let continuation_token = req.rpc_params?.continuation_token;
     const max_buckets = req.rpc_params?.max_buckets;
 
-    const accessible_bucket_list = system_store.data.buckets.filter(
-        async bucket => await req.has_s3_bucket_permission(bucket, "s3:ListBucket", req) && !bucket.deleting
-    );
+    // filter buckets based on ownership
+    const accessible_bucket_list = system_store.data.buckets.filter(bucket => {
+        if (bucket.deleting) return false;
+        return req.has_list_bucket_permission(bucket);
```
shirady: Why do you need to add a function in the request for listing the buckets?
WDYT about creating a helper function that would fetch the buckets according to the owner in the bucket server?

src/server/common_services/auth_server.js
```diff
@@ -496,6 +496,10 @@ function _prepare_auth_request(req) {
     req.has_bucket_action_permission = async function(bucket, action, bucket_path, req_query) {
         return has_bucket_action_permission(bucket, req.account, action, req_query, bucket_path);
     };
+
+    req.has_list_bucket_permission = function(bucket) {
+        return has_list_bucket_permission(bucket, req.account, req.system);
+    };
```
shirady: If you can kindly share why you need to have a function that is related only to list buckets on the request?

src/server/common_services/auth_server.js
shirady, line 1: I think in some conditions of the optional chaining that might result in `undefined` === `undefined` `true`.

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,51 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+ * @returns {boolean}
+ */
+function has_list_bucket_permission(bucket, account, system) {
+    // system owner can list all the buckets
+    if (is_system_owner(bucket.system, account)) return true;
+
+    // bucket owner can list their buckets only
+    if (is_bucket_owner(bucket, account)) return true;
+
+    // TODO: Add IAM policy support for user's s3:ListAllMyBuckets action
```
shirady: I think that in the flow the IAM policy was already checked, so you need to list the buckets of the account that owns the user.
You have something similar in `bucketspace_fs.js‎`, right?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -542,6 +545,10 @@ mocha.describe('Session token tests', function() {
             name: 'first.bucket',
             policy: s3accesspolicy,
         });
+
+        // create a bucket owned by alice2 for ListBuckets to work
+        // Note: bucket policy alone is not sufficient for ListBuckets operation
```
shirady: bucket policy is not related to list buckets.

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -564,7 +571,7 @@ mocha.describe('Session token tests', function() {
         });
 
         const buckets1 = await temp_s3_with_session_token.listBuckets().promise();
-        assert.ok(buckets1.Buckets.length > 0);
+        assert.ok(buckets1.Buckets[0].Name === alice2_buck);
```
shirady, line 574: Why do you need to change this?

src/test/integration_tests/nsfs/test_nsfs_integration.js
```diff
@@ -387,6 +385,16 @@ mocha.describe('bucket operations - namespace_fs', function() {
...
+
+        let account_principal;
+        if (is_nc_coretest) {
+            account_principal = 'account_s3_correct_uid';
+        } else {
+            const account_info = await rpc_client.account.read_account({ email: 'account_s3_correct_uid@noobaa.com' });
+            account_principal = iam_utils.create_arn_for_root(account_info._id.toString());
+        }
+        const generated = generate_s3_policy(account_principal, first_bucket, ['s3:GetBucketAcl']);
+        await rpc_client.bucket.put_bucket_policy({ name: first_bucket, policy: generated.policy });
```
shirady, line 400: What is this addition?
In which case did we need the action of `GetBucketAcl`?

src/sdk/bucketspace_fs.js
```diff
@@ -921,25 +921,52 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
         const bucket_policy = bucket.s3_policy;
 
-        if (!bucket_policy) {
-            // in case we do not have bucket policy
-            // we allow IAM account to access a bucket that that is owned by their root account
-            const is_iam_and_same_root_account_owner = account.owner !== undefined &&
-                account.owner === bucket.owner_account.id;
-            return is_owner || is_iam_and_same_root_account_owner;
-        }
+        if (!bucket_policy) return owner;
```
shirady: Why did you remove this part?

src/sdk/bucketspace_fs.js
```diff
@@ -921,25 +921,52 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+    // TODO: move the following 4 functions - is_bucket_owner(), has_list_bucket_permission(), has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
     // so they can be re-used
+
+    /**
+     * is_bucket_owner checks if the account owns the bucket
+     * @param {Record<string, any>} bucket
+     * @param {Record<string, any>} account
+     * @returns {boolean}
+     */
+    is_bucket_owner(bucket, account) {
```
shirady: I think it is confusing; the IAM user does not own the bucket.

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,51 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+/**
+ * has_list_bucket_permission returns true if the account can list the bucket in ListBuckets operation
+ * this is a synchronous function that only checks system ownership and bucket ownership
+ *
+ * @param {Record<string, any>} bucket
+ * @param {Record<string, any>} account
+ * @param {Record<string, any>} system
+ * @returns {boolean}
+ */
+function has_list_bucket_permission(bucket, account, system) {
```
shirady: I think the name of the function is confusing, as the operation S3 is not list bucket, but list buckets.

src/sdk/bucketspace_fs.js
aayushchouhan09: This function is just checking the ownership of the bucket, as per its working I think it looks good.
And for IAM user we can have another function which only checks for user.

src/sdk/bucketspace_fs.js
aayushchouhan09: I will revert it, I thought it was only checking the owner again. Thanks

src/server/common_services/auth_server.js
aayushchouhan09: If you talking about this line:
`return owner || (permission_by_id === 'ALLOW' || permission_by_name === 'ALLOW');`

It would never go to `undefined` === `undefined` --> `true` , because we are comparing with string `'ALLOW'`

src/server/common_services/auth_server.js
aayushchouhan09: Previously, we are using `has_bucket_action_permission()` to check the permission of the bucket but that function is generic checking for bucket policy which is not related to list buckets.
So, we have introduced another function called `has_list_bucket_permission()` which is specific to list bucket only.

src/server/common_services/auth_server.js
aayushchouhan09: I think like we are checking for the permission of one bucket at a time in this fuction. But yes I agree we should rename it. Thanks.

src/server/common_services/auth_server.js
aayushchouhan09: Sure, I will try to add that check. Thanks

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: before our changes tests is expecting `first.bucket` to be listed because of bucket policy applied to it. But, as we remove the bucket policy check from list buckets it cannot be listed. And to maintain the tests to work as expected we created another bucket in the account and listing it.

src/test/integration_tests/nsfs/test_nsfs_integration.js
aayushchouhan09: `'get bucket acl - rpc bucket'`
this test was failing as it is not having the `GetBucketAcl` permission on `first.bucket` (admin bucket), so we added a policy here to pass this test.

src/server/system_services/bucket_server.js
aayushchouhan09: Can you explain where you want this function and changes?
I tried to keep the flow similar to `has_bucket_action_permission()`  for list buckets permission checks.

src/sdk/bucketspace_fs.js
aayushchouhan09: I will update this. Thanks

src/server/common_services/auth_server.js
aayushchouhan09: updated at other places, Thanks

src/server/system_services/bucket_server.js
shirady: If you can add a comment to document what you implemented, it would be clear for others as well.

> As per my understanding (currently):
> 
> 1. Returns all the buckets for admin
> 2. Returns only buckets the account owns
> 3. NSFS buckets need both ownership + filesystem access
> 4. Bucket policies do not affect this operation (AWS s3 behaviour)

Also the OBC case.

src/test/integration_tests/api/s3/test_s3_list_buckets.js
shirady: In general, I would prefer modifying a test rather than removing a test.

src/server/system_services/bucket_server.js
shirady: I see that you tried the same flow as `has_bucket_action_permission`, but I thought it would be easier if you had just a helper function in the bucket server for that purpose.
Anyway, I would review it as is now.

src/server/common_services/auth_server.js
shirady: I still think it exists in parts of the code, I will try to point out where I see it.

src/test/integration_tests/api/sts/test_sts.js
shirady: `buckets1.Buckets[0].Name` there is an element of ordering here (a bucket in index 0) - if `buckets1.Buckets` has only 1 bucket? add a comment, or change the validation to check if it is included.

src/server/common_services/auth_server.js
```diff
@@ -496,6 +496,10 @@ function _prepare_auth_request(req) {
     req.has_bucket_action_permission = async function(bucket, action, bucket_path, req_query) {
         return has_bucket_action_permission(bucket, req.account, action, req_query, bucket_path);
     };
+
+    req.has_list_bucket_permission = function(bucket) {
```
shirady: I still see `list_bucket` where you mean `list_buckets`.
I think it is confusing because:
- List bucket = list objects in the buckets.
- List buckets = list buckets under the account.

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,100 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
 }
 
+/**
+ * is_system_owner checks if the account is the system owner
+ * @param {Record<string, any>} system
+ * @param {Record<string, any>} account
+ * @returns {boolean}
+ */
+function is_system_owner(system, account) {
+    return system?.owner?.email?.unwrap() === account?.email?.unwrap();
```
shirady: If I pass `system` as `undefined` and `account` as `undefined`, the function will return `true`?

I would use the original statement, and you can return false on `undefined` arguments if you want.
This was the original line:
`bucket.system.owner.email.unwrap() === account.email.unwrap()`

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,100 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+}
+
+/**
+ * is_bucket_owner checks if the account is the direct owner of the bucket
+ * @param {Record<string, any>} bucket
+ * @param {Record<string, any>} account
+ * @returns {boolean}
+ */
+function is_bucket_owner(bucket, account) {
+    return bucket?.owner_account?.email?.unwrap() === account?.email?.unwrap();
```
shirady: If I pass `bucket` as `undefined` and `account` as `undefined`, the function will return `true`?

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,100 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+
+/**
+ * has_iam_list_buckets_permission checks if the account has IAM policy granting s3:ListAllMyBuckets
+ * for buckets owned by their root account
+ * @param {Record<string, any>} bucket
+ * @param {Record<string, any>} account
+ * @returns {Promise<boolean>}
+ */
+async function has_iam_list_buckets_permission(bucket, account) {
+    if (!account?.owner) return false;
```
shirady: Why do you use optional chaining here, and why do you return `false`?
I thought about `if (account.owner === undefined)...`

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,100 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+    const iam_policies = account.iam_user_policies || [];
+    if (iam_policies.length === 0) return false;
+
+    // check each IAM policy for s3:ListAllMyBuckets permission
+    for (const iam_policy of iam_policies) {
+        const result = await s3_bucket_policy_utils.has_bucket_policy_permission(
+            iam_policy.policy_document,
+            undefined,
+            's3:ListAllMyBuckets',
+            'arn:aws:s3:::*',
```
shirady: Did you test this in AWS?
I think it should be:

```suggestion
            '*',
```

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,100 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+}
+
+/**
+ * has_iam_list_buckets_permission checks if the account has IAM policy granting s3:ListAllMyBuckets
+ * for buckets owned by their root account
+ * @param {Record<string, any>} bucket
+ * @param {Record<string, any>} account
+ * @returns {Promise<boolean>}
+ */
+async function has_iam_list_buckets_permission(bucket, account) {
```
shirady: I don't understand why do you need this function exactly it looks complicated.
Just to make sure that we understand each other the IAM policy is only for users, not accounts in general.

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,100 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+            'arn:aws:s3:::*',
+            undefined,
+            { should_pass_principal: false }
+        );
+
+        if (result === 'DENY') return false;
+
+        if (result === 'ALLOW') {
+            const bucket_owner_id = bucket.owner_account?._id;
+            const account_owner_id = account.owner?._id;
```
shirady: Why do you use optional chaining here (`bucket.owner_account?._id`, `account.owner?._id`)?

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,100 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+
+    // check direct ownership
+    if (is_bucket_owner(bucket, account)) return true;
+
+    // special case: check bucket claim ownership (OBC)
+    if (account.bucket_claim_owner?.name?.unwrap() === bucket.name.unwrap()) return true;
+
+    // check IAM policy for s3:ListAllMyBuckets
+    // Note: IAM users need this policy to list buckets owned by their root account
+    if (await has_iam_list_buckets_permission(bucket, account)) return true;
```
shirady: Before you call it, you want this only for users (where `account.owner !== undefined`), right?

src/server/common_services/auth_server.js
```diff
@@ -560,7 +666,8 @@ async function has_bucket_action_permission(bucket, account, action, req_query,
     );
 
     if (result === 'DENY') return false;
-    return is_owner || result === 'ALLOW';
+
+    return owner || result === 'ALLOW';
```
shirady: In case there is a bucket policy, we want to follow it and not override it for users under the account.

src/server/common_services/auth_server.js
```diff
@@ -538,14 +636,22 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+    if (is_system_owner(bucket.system, account)) return true;
+
+    // check ownership: direct owner, IAM user inheritance, or OBC
+    const is_owner = is_bucket_owner(bucket, account);
+    const bucket_owner_id = bucket.owner_account?._id;
+    const account_owner_id = account.owner?._id;
+    const has_iam_access = account_owner_id !== undefined && bucket_owner_id !== undefined &&
+                           String(bucket_owner_id) === String(account_owner_id);
+    const has_obc_access = account.bucket_claim_owner?.name?.unwrap() === bucket.name.unwrap();
+    const owner = is_owner || has_iam_access || has_obc_access;
```
shirady: I think this variable name is too general and might be confusing.

src/test/integration_tests/nsfs/test_nsfs_integration.js
```diff
@@ -1125,10 +1131,11 @@ mocha.describe('list objects - namespace_fs', async function() {
         await rpc_client.bucket.put_bucket_policy({ name: s3_root_b_name, policy: s3_policy.policy });
     });
 
-    mocha.it('list buckets - uid & gid match - account with permission', async function() {
+    mocha.it('list buckets - uid & gid match - account lists owned bucket', async function() {
         const { s3_client } = accounts.account4;
         const res = await s3_client.listBuckets({});
-        assert.equal(res.Buckets.length, 4);
+        assert.ok(res.Buckets.length >= 1);
+        assert.ok(res.Buckets.some(b => b.Name === s3_root_b_name));
```
shirady, line 1138: If you can add a comment here, it can help

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,68 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+
+    // check direct ownership
+    if (is_bucket_owner(bucket, account)) return true;
+
+    // special case: check bucket claim ownership (OBC)
+    if (is_bucket_claim_owner(bucket, account)) return true;
+
+    // special case: iam user can list the buckets of their owner
+    if (account.owner) {
+        const account_owner = system_store.data.get_by_id(account.owner._id);
```
shirady: I would try to avoid the `system_store` call and reuse what we have, this is a reference that might help:

```
        const is_iam_and_same_root_account_owner = account.owner !== undefined &&
            account.owner._id.toString() === bucket.owner_account._id.toString();
```

aayushchouhan09: We will create a separate PR for this fix, added a comment to add this fix later. Thanks

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,69 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+ * - OBC owner can list their buckets
+ * - IAM users can list their owner buckets
+ *
+ * @param {Record<string, any>} bucket
+ * @param {Record<string, any>} account
+ * @returns {Promise<boolean>}
+ */
+async function has_bucket_ownership_permission(bucket, account) {
+    // system owner can list all the buckets
+    if (is_system_owner(bucket, account)) return true;
```
shirady, line 577: @aayushchouhan09 
I think that for the command `noobaa bucket list` in the CLI that you mentioned in the call, you need an additional condition.
In the help of the CLI:

> List NooBaa buckets

Otherwise, it will list only the buckets under the operator account - which are the OBC buckets that were created with the command `noobaa obc create <obc-name>`.

cc: @naveenpaul1

src/server/common_services/auth_server.js
```diff
@@ -520,6 +524,69 @@ function _get_auth_info(account, system, authorized_by, role, extra) {
...
+
+    // check direct ownership
+    if (is_bucket_owner(bucket, account)) return true;
+
+    // special case: check bucket claim ownership (OBC)
+    if (is_bucket_claim_owner(bucket, account)) return true;
+
+    // special case: iam user can list the buckets of their owner
+    // TODO: handle iam user
+    // if (account.owner && is_bucket_owner(bucket, account.owner)) return true;
```
shirady: no need for code in comments

shirady (approved):
Approving not to block you.
Known gaps:
1. List buckets requested by the operator (to support `noobaa bucket list` in the operator CLI and see all the NooBaa buckets).
2. List buckets requested by IAM user (in a following PR).

src/server/common_services/auth_server.js
aayushchouhan09: Added a fix for it. Thanks!

PR #9281, opened 2025-11-13
IAM | Authorize Requests for IAM Users According to IAM User Inline Policy

shirady:
### Describe the Problem
Add a step in `s3_rest` to authorize requests for IAM users according to the IAM user inline policy.

### Explain the Changes
1. Add the function `authorize_request_iam_policy` and call it from `authorize_request`.
2. Add properties that I need for this step in `account_info` (`owner` and `iam_user_policies`).

### Issues:
List of GAPs:
1. The tests are only manual at this point; there is a plan to add automated tests after we have the design change.
2. I might want to refactor the code so that `s3_bucket_policy_utils.js` so the terms will be policy in general and can be used for bucket policy and IAM policy.
3. I might want to move the created function somewhere else (perhaps `iam_rest`) so the thrown error will be `IamError.AccessDeniedException` instead of `S3Error.AccessDenied`

Design Decisions:
- When there is no IAM user policy, it is authorized for this step (didn't want to change default behavior).
- The IAM policy authorization is done in **parallel** - I had the option to do it **sequentially**, meaning policy after policy - get the resolved promise answer and work on it (`DENY` throws an error), but I decided to do it in parallel, assuming that the total user policy is limited. If I were to do it sequentially, I would need to wait for every policy check, instead of running all the checks and waiting for all the promises to be resolved and work on the result (`DENY`, `ALLOW`, `IMPLICIT_DENY`). I also decided to check in parallel to the bucket policy, and I had the option to do it after, as in the case where there is an IAM user policy with `DENY`, we would like it to be thrown as part of the calculation.

### Testing Instructions:
4. Wait for the default backing store pod to be in state Ready before starting the tests: `kubectl wait --for=condition=available backingstore/noobaa-default-backing-store --timeout=6m -n test1`
5. I'm using port-forward (in a different tab):
6. Create the alias for the admin - first, need to get the credentials: `nb status --show-secrets -n test1` and then:
7. Check the connection to the endpoint:
8. Create a user: `admin-iam iam create-user --user-name Robert`
9. Create access keys: `admin-iam iam create-access-keys --user-name Robert`
10. Create the alias for the user (like in step 4 with alias `user-2-s3`).
11. The case with no bucket policy, no IAM policy - the user will not be restricted from S3 operation, for example:
- The user creates a bucket: `user-2-s3 s3 mb s3://mybucket` (should work).
- The user put an object in that bucket: `echo 'test_data' | user-2-s3 s3 cp - s3://first.bucket/test_object.txt` (should work).
12. The admin adds an inline policy so the user cannot create a bucket, but can put an object.
policy_deny_create_bucket_allow_put_object.json
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Deny",
            "Action": [
                "s3:CreateBucket"
            ],
            "Resource": "*"
        },
        {
            "Effect": "Allow",
            "Action": [
                "s3:PutObject"
            ],
            "Resource": "*"
        }
    ]
}
```
13. `admin-iam iam put-user-policy --user-name Robert --policy-name policy_deny_create_bucket_allow_put_object --policy-document file://policy_deny_create_bucket_allow_put_object.json`
14. Repeat step 10:
- The user put an object in that bucket: `echo 'test_data' | user-2-s3 s3 cp - s3://first.bucket/test_object2.txt` (should work).
- The user creates a bucket: `user-2-s3 mb mybucket2` (should throw `AccessDenied` error).

Code changes for testing:
```diff
const account_cache = new LRUCache({
    name: 'AccountCache',
-    expiry_ms: config.OBJECT_SDK_ACCOUNT_CACHE_EXPIRY_MS,
+   expiry_ms: 1, //SDSD 
```
2. (temporarily) Not to be blocked on authorization by bucket policy `src/endpoint/s3/s3_rest.js` in the `if (!s3_policy) {` remove the error thrown as currently `owner_account` in NC deployment and is undefined. rest of the code in `else` condition.

Notes:
- I used the admin account, but for IAM operations, there is no difference between a created account and an admin account. I tested with the admin only because it saves steps (I have the admin account upon system creation and a bucket).

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Accounts can include an owner field and associated IAM user policies surfaced in account info.
  * S3 authorization now evaluates IAM user inline policies in parallel with existing bucket/account policies.
  * Bucket permission checks honor public-access restrictions when evaluating permissions.

### DIFF
--- a/src/endpoint/s3/s3_bucket_policy_utils.js
+++ b/src/endpoint/s3/s3_bucket_policy_utils.js
@@ -149,7 +149,8 @@ async function _is_object_version_fit(req, predicate, value) {
     return res;
 }
 
-async function has_bucket_policy_permission(policy, account, method, arn_path, req, disallow_public_access = false) {
+async function has_bucket_policy_permission(policy, account, method, arn_path, req,
+    { disallow_public_access = false, should_pass_principal = true } = {}) {
     const [allow_statements, deny_statements] = _.partition(policy.Statement, statement => statement.Effect === 'Allow');
 
     // the case where the permission is an array started in op get_object_attributes
@@ -157,12 +158,19 @@ async function has_bucket_policy_permission(policy, account, method, arn_path, r
 
     // look for explicit denies
     const res_arr_deny = await is_statement_fit_of_method_array(
-        deny_statements, account, method_arr, arn_path, req); // No need to disallow in "DENY"
+        deny_statements, account, method_arr, arn_path, req, {
+            disallow_public_access: false, // No need to disallow in "DENY"
+            should_pass_principal
+        }
+    );
     if (res_arr_deny.every(item => item)) return 'DENY';
 
     // look for explicit allows
     const res_arr_allow = await is_statement_fit_of_method_array(
-        allow_statements, account, method_arr, arn_path, req, disallow_public_access);
+        allow_statements, account, method_arr, arn_path, req, {
+            disallow_public_access,
+            should_pass_principal
+        });
     if (res_arr_allow.every(item => item)) return 'ALLOW';
 
     // implicit deny
@@ -217,15 +225,19 @@ function _is_resource_fit(arn_path, statement) {
     return statement.Resource ? resource_fit : !resource_fit;
 }
 
-async function is_statement_fit_of_method_array(statements, account, method_arr, arn_path, req, disallow_public_access = false) {
+async function is_statement_fit_of_method_array(statements, account, method_arr, arn_path, req,
+    { disallow_public_access = false, should_pass_principal = true } = {}) {
     return Promise.all(method_arr.map(method_permission =>
-        _is_statements_fit(statements, account, method_permission, arn_path, req, disallow_public_access)));
+        _is_statements_fit(statements, account, method_permission, arn_path, req, { disallow_public_access, should_pass_principal })));
 }
 
-async function _is_statements_fit(statements, account, method, arn_path, req, disallow_public_access = false) {
+async function _is_statements_fit(statements, account, method, arn_path, req,
+    { disallow_public_access = false, should_pass_principal = true} = {}) {
     for (const statement of statements) {
         const action_fit = _is_action_fit(method, statement);
-        const principal_fit = _is_principal_fit(account, statement, disallow_public_access);
+        // When evaluating IAM user inline policies, should_pass_principal is false since these policies
+        // don't have a Principal field (the principal is implicitly the user)
+        const principal_fit = should_pass_principal ? _is_principal_fit(account, statement, disallow_public_access) : true;
         const resource_fit = _is_resource_fit(arn_path, statement);
         const condition_fit = await _is_condition_fit(statement, req, method);
 
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -224,7 +224,9 @@ async function authorize_request(req) {
         req.object_sdk.authorize_request_account(req),
         // authorize_request_policy(req) is supposed to
         // allow owners access unless there is an explicit DENY policy
-        authorize_request_policy(req)
+        authorize_request_policy(req),
+        // authorize_request_iam_policy(req) is for users only
+        authorize_request_iam_policy(req),
     ]);
 }
 
@@ -299,14 +301,17 @@ async function authorize_request_policy(req) {
     // we start the permission check on account identifier intentionally
     if (account_identifier_id) {
         permission_by_id = await s3_bucket_policy_utils.has_bucket_policy_permission(
-            s3_policy, account_identifier_id, method, arn_path, req, public_access_block?.restrict_public_buckets);
+            s3_policy, account_identifier_id, method, arn_path, req,
+            { disallow_public_access: public_access_block?.restrict_public_buckets }
+        );
         dbg.log3('authorize_request_policy: permission_by_id', permission_by_id);
     }
     if (permission_by_id === "DENY") throw new S3Error(S3Error.AccessDenied);
 
     if ((!account_identifier_id || permission_by_id !== "DENY") && account.owner === undefined) {
         permission_by_name = await s3_bucket_policy_utils.has_bucket_policy_permission(
-            s3_policy, account_identifier_name, method, arn_path, req, public_access_block?.restrict_public_buckets
+            s3_policy, account_identifier_name, method, arn_path, req,
+            { disallow_public_access: public_access_block?.restrict_public_buckets }
         );
         dbg.log3('authorize_request_policy: permission_by_name', permission_by_name);
     }
@@ -316,11 +321,50 @@ async function authorize_request_policy(req) {
     throw new S3Error(S3Error.AccessDenied);
 }
 
+async function authorize_request_iam_policy(req) {
+    const auth_token = req.object_sdk.get_auth_token();
+    const is_anonymous = !(auth_token && auth_token.access_key);
+    if (is_anonymous) return;
+
+    const account = req.object_sdk.requesting_account;
+    const is_iam_user = account.owner !== undefined;
+    if (!is_iam_user) return; // IAM policy is only on IAM users (account root user is authorized here)
+
+    const resource_arn = _get_arn_from_req_path(req);
+    const method = _get_method_from_req(req);
+    const iam_policies = account.iam_user_policies || [];
+    if (iam_policies.length === 0) return;
+
+    // parallel policy check
+    const promises = [];
+    for (const iam_policy of iam_policies) {
+        // We are reusing the bucket policy util function as it checks the policy document
+        const promise = s3_bucket_policy_utils.has_bucket_policy_permission(
+            iam_policy.policy_document, undefined, method, resource_arn, req,
+            { should_pass_principal: false }
+        );
+        promises.push(promise);
+    }
+    const permission_result = await Promise.all(promises);
+    let has_allow_permission = false;
+    for (const permission of permission_result) {
+        if (permission === "DENY") throw new S3Error(S3Error.AccessDenied);
+        if (permission === "ALLOW") {
+            has_allow_permission = true;
+        }
+    }
+    if (has_allow_permission) return;
+    dbg.log1('authorize_request_iam_policy: user have inline policies but none of them matched the method');
+    throw new S3Error(S3Error.AccessDenied);
+}
+
 async function authorize_anonymous_access(s3_policy, method, arn_path, req, public_access_block) {
     if (!s3_policy) throw new S3Error(S3Error.AccessDenied);
 
     const permission = await s3_bucket_policy_utils.has_bucket_policy_permission(
-        s3_policy, undefined, method, arn_path, req, public_access_block?.restrict_public_buckets);
+        s3_policy, undefined, method, arn_path, req,
+        { disallow_public_access: public_access_block?.restrict_public_buckets }
+    );
     if (permission === "ALLOW") return;
 
     throw new S3Error(S3Error.AccessDenied);
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -1049,6 +1049,14 @@ function get_account_info(account, include_connection_cache) {
     };
     info.role_config = account.role_config;
     info.force_md5_etag = account.force_md5_etag;
+
+    if (account.iam_user_policies) {
+        info.iam_user_policies = account.iam_user_policies;
+    }
+    if (account.owner) {
+        info.owner = account.owner._id.toString();
+    }
+
     return info;
 }
 
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -749,6 +749,15 @@ module.exports = {
                 },
                 nsfs_account_config: {
                     $ref: 'common_api#/definitions/nsfs_account_config'
+                },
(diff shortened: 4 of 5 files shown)


src/server/system_services/account_server.js
```diff
@@ -1049,6 +1049,14 @@ function get_account_info(account, include_connection_cache) {
     };
     info.role_config = account.role_config;
     info.force_md5_etag = account.force_md5_etag;
+
+    if (account.iam_user_policies) {
+        info.iam_user_policies = account.iam_user_policies;
+    }
+    if (account.owner) {
+        info.owner = account.owner._id.toString();
+    }
```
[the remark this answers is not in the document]
shirady: The function `authorize_request_policy` handles the account object from 2 deployments:
1. File system (NC in our docs), and the account schema for it is the file `src/server/system_services/schemas/nsfs_account_schema.js`, and there you can see the `owner` is a string.
2. Container deployment - in which the `owner` was added lately, but we didn't have it in the `account` in this function, as it was not defined in the `account_info` and hence was always `undefined` - until the last few weeks we didn't have users, so this is a change and this code is work in progress in a different PR.

To clarify, as I understand it, the `account_server` will check if we have the property `owner` in the account. This property is defined in the schema `src/server/system_services/schemas/account_schema.js` and serves as a reference to the owner. It contains the `_id` field, which we pass in the API as a string.

Please let me know if I missed something.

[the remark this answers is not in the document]
shirady: In NC deployment we only have the endpoint, we don't have the nobaa-core.
The account server is in noobaa-core, so it is only relevant in a containerized deployment.

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -217,15 +218,17 @@ function _is_resource_fit(arn_path, statement) {
...
+        _is_statements_fit(statements, account, method_permission, arn_path, req, disallow_public_access, should_pass_principal)));
 }
 
-async function _is_statements_fit(statements, account, method, arn_path, req, disallow_public_access = false) {
+async function _is_statements_fit(statements, account, method, arn_path, req, disallow_public_access = false,
+        should_pass_principal = true) {
     for (const statement of statements) {
         const action_fit = _is_action_fit(method, statement);
-        const principal_fit = _is_principal_fit(account, statement, disallow_public_access);
+        const principal_fit = should_pass_principal ? _is_principal_fit(account, statement, disallow_public_access) : true;
```
naveenpaul1, line 240: In which flow we skipping _is_principal_fit()?

src/endpoint/s3/s3_rest.js
```diff
@@ -316,6 +318,42 @@ async function authorize_request_policy(req) {
...
+    for (const iam_policy of iam_policies) {
+        // We reusing the bucket policy util function as it is checks the policy document
+        const promise = s3_bucket_policy_utils.has_bucket_policy_permission(
+            iam_policy.policy_document, undefined, method, resource_arn, req, undefined, false
+        );
+        promises.push(promise);
+    }
+    const permission_result = await Promise.all(promises);
+    let has_allow_permission = false;
+    for (const permission of permission_result) {
```
naveenpaul1, line 350: can we permission_result.includes('DENY')? if `yes` then throw S3Error?

src/endpoint/s3/s3_rest.js
```diff
@@ -316,6 +318,42 @@ async function authorize_request_policy(req) {
...
+    const method = _get_method_from_req(req);
+    const iam_policies = account.iam_user_policies || [];
+    if (iam_policies.length === 0) return;
+
+    // parallel policy check
+    const promises = [];
+    for (const iam_policy of iam_policies) {
+        // We reusing the bucket policy util function as it is checks the policy document
+        const promise = s3_bucket_policy_utils.has_bucket_policy_permission(
+            iam_policy.policy_document, undefined, method, resource_arn, req, undefined, false
```
naveenpaul1: Why are not checking the principal here? As my understanding IAM policy should have the constructed principal that consists of the iam user ARN, Correct me if I am worng.

src/endpoint/s3/s3_rest.js
```diff
@@ -224,7 +224,9 @@ async function authorize_request(req) {
         req.object_sdk.authorize_request_account(req),
         // authorize_request_policy(req) is supposed to
         // allow owners access unless there is an explicit DENY policy
-        authorize_request_policy(req)
+        authorize_request_policy(req),
```
naveenpaul1, line 227: What will happen if there is no `explicit deny` in authorize_request_policy() and there is `implicit allow` in iam_policy?

src/endpoint/s3/s3_bucket_policy_utils.js
shirady: The user's inline policy is embedded to a user; there is no principal field in the JSON policy.
If it helps, you can see in the example in `PutUserPolicy` the IAM policy attached ([link](https://docs.aws.amazon.com/IAM/latest/APIReference/API_PutUserPolicy.html)):

> PolicyDocument={"Version":"2012-10-17",		 	 	 "Statement":{"Effect":"Allow","Action":["iam:Get*", "iam:List*", "iam:Generate*"],"Resource":"*"}}

src/endpoint/s3/s3_rest.js
shirady: There is no `implicit allow` case.
The options are:

1. `DENY` - we throw an error immediately 
2. `ALLOW` - we authorize the request
3. `IMPLICIT DENY` - we will throw an error if we didn't encounter `ALLOW`

I will try to answer - case:
`authorize_request_policy` - `DENY` -> will throw error
`authorize_request_iam_policy` - `ALLOW` -> the function will return
The user will encounter the `AccessDenied` error due to the bucket policy check.

src/endpoint/s3/s3_rest.js
shirady: As mentioned, the IAM inline user policy that was added does not have the principal field.
I attached an example, and I can also share links from the docs, for example ([link](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html)):

> You cannot use the Principal element in an identity-based policy. Identity-based policies are permissions policies that you attach to IAM identities (users, groups, or roles). In those cases, the principal is implicitly the identity where the policy is attached.

src/endpoint/s3/s3_rest.js
shirady: We can use `Array.prototype.includes()` instead of a `for` loop, but since we check 2 options changing the loop to `include` would mean to use it twice, iterating twice instead of once.

src/endpoint/s3/s3_bucket_policy_utils.js
naveenpaul1: Its don't have the principal in policy because the principal is the same account that it owns. It can not have principal because it will never have multiple principals, will have one principal that is itself. WDYT?

src/endpoint/s3/s3_rest.js
naveenpaul1: Since bucket policy with ARN is not handled in this PR I am fine with this changes,
And we need to handle it in next PR

naveenpaul1 (approved):

src/endpoint/s3/s3_bucket_policy_utils.js
shirady: We don't have a `Principal` field in the user inline policy.
I only reused the function of the bucket policy.
Perhaps, I will create a refactoring PR in the future so it's clear (I mentioned this in the PR description in the list of Gaps).

src/endpoint/s3/s3_rest.js
shirady: In this story of IAM policy, it only deals with IAM policy.
The cases of the bucket policy are different tasks (a different story that related to bucket policy) - as discussed should be added in the task list, and were not part of this story.

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -149,20 +149,21 @@ async function _is_object_version_fit(req, predicate, value) {
     return res;
 }
 
-async function has_bucket_policy_permission(policy, account, method, arn_path, req, disallow_public_access = false) {
+async function has_bucket_policy_permission(policy, account, method, arn_path, req, disallow_public_access = false,
+        should_pass_principal = true) {
```
liranmauda: Giving 2 options with Boolean default value can lead to bugs
```suggestion
async function has_bucket_policy_permission(policy, account, method, arn_path, req, 
{
disallow_public_access = false,
    should_pass_principal = true,
  } = {}
```
Then, when calling it, you can: 
```
has_bucket_policy_permission(policy, account, method, arn, req, {
  should_pass_principal: true,
});
```

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -149,20 +149,21 @@ async function _is_object_version_fit(req, predicate, value) {
...
     const [allow_statements, deny_statements] = _.partition(policy.Statement, statement => statement.Effect === 'Allow');
 
     // the case where the permission is an array started in op get_object_attributes
     const method_arr = Array.isArray(method) ? method : [method];
 
     // look for explicit denies
     const res_arr_deny = await is_statement_fit_of_method_array(
-        deny_statements, account, method_arr, arn_path, req); // No need to disallow in "DENY"
+        deny_statements, account, method_arr, arn_path, req, undefined, should_pass_principal); // No need to disallow in "DENY"
```
liranmauda: Update according to the first comment

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -149,20 +149,21 @@ async function _is_object_version_fit(req, predicate, value) {
...
     if (res_arr_deny.every(item => item)) return 'DENY';
 
     // look for explicit allows
     const res_arr_allow = await is_statement_fit_of_method_array(
-        allow_statements, account, method_arr, arn_path, req, disallow_public_access);
+        allow_statements, account, method_arr, arn_path, req, disallow_public_access, should_pass_principal);
```
liranmauda: Update according to the first comment

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -217,15 +218,17 @@ function _is_resource_fit(arn_path, statement) {
     return statement.Resource ? resource_fit : !resource_fit;
 }
 
-async function is_statement_fit_of_method_array(statements, account, method_arr, arn_path, req, disallow_public_access = false) {
+async function is_statement_fit_of_method_array(statements, account, method_arr, arn_path, req,
```
liranmauda, line 228: Same comment regarding the options

src/endpoint/s3/s3_bucket_policy_utils.js
liranmauda: @shirady add a comment here that explains

src/endpoint/s3/s3_rest.js
```diff
@@ -316,6 +318,42 @@ async function authorize_request_policy(req) {
     throw new S3Error(S3Error.AccessDenied);
 }
 
+async function authorize_request_iam_policy(req) {
+    const auth_token = req.object_sdk.get_auth_token();
+    const is_anon = !(auth_token && auth_token.access_key);
```
liranmauda: Personally, I do not like name shortcuts
```suggestion
    const is_anonymous = !(auth_token && auth_token.access_key);
```

src/endpoint/s3/s3_rest.js
```diff
@@ -316,6 +318,42 @@ async function authorize_request_policy(req) {
...
+    if (is_anon) return;
+
+    const account = req.object_sdk.requesting_account;
+    const is_iam_user = account.owner !== undefined;
+    if (!is_iam_user) return;
```
liranmauda: Add a comment

liranmauda (changes requested):

src/endpoint/s3/s3_bucket_policy_utils.js
shirady: Changed, thanks.

src/endpoint/s3/s3_bucket_policy_utils.js
shirady: The code was updated according to the function signature, thanks.

src/endpoint/s3/s3_bucket_policy_utils.js
shirady: The code was updated according to the function signature, thanks.

src/endpoint/s3/s3_bucket_policy_utils.js
shirady: The code was updated according to the function signature, thanks.

src/endpoint/s3/s3_rest.js
shirady: I agree, I copied this from another function and wanted to use it as is.
Anyway, I changed it in the new code, thanks.

src/endpoint/s3/s3_rest.js
shirady: I added a comment, thanks.

liranmauda (approved):

PR #9287, opened 2025-11-18
use exact read size to acquire from io semaphore

dannyzaken:
### Describe the Problem
- In read_object_stream, we used the requested_size passed to the _read function as the value to acquire from the io semaphore. By default, this is 32 MB (the stream's highWaterMark).
- For datasets with mostly small objects, this limits the number of concurrent reads more than necessary.

### Explain the Changes
- Changed io_sem_size to reflect the actual size requested by the current read.
- Also, avoid entering the code under the semaphore if there is nothing more to read.
- changed debug level of some read\upload messages to log1 instead of log0

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed early termination handling for read streams and ensured caught errors are rethrown after logging.
  * Improved completion signaling when reads finish and added explicit end-of-stream behavior when requested range is already satisfied.

* **Refactor**
  * Reduced logging verbosity across upload and read flows to minimize noise.
  * Adjusted read resource sizing to better match requested ranges.
  * Added a log entry indicating streaming target and reader position progress.

### DIFF
--- a/src/sdk/object_io.js
+++ b/src/sdk/object_io.js
@@ -173,7 +173,7 @@ class ObjectIO {
         params.bucket_master_key_id = obj_upload.bucket_master_key_id;
 
         try {
-            dbg.log0('upload_object_range: start upload stream', upload_params);
+            dbg.log1('upload_object_range: start upload stream', upload_params);
             return this._upload_stream(params, complete_params);
         } catch (err) {
             dbg.error('upload_object_range: object part upload failed', upload_params, err);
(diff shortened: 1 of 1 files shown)


src/sdk/object_io.js
```diff
@@ -592,13 +592,19 @@ class ObjectIO {
...
                 return;
             }
-            const io_sem_size = _get_io_semaphore_size(requested_size);
 
             // TODO we dont want to use requested_size as end, because we read entire chunks
             // and we are better off return the data to the stream buffer
             // instead of getting multiple calls from the stream with small slices to return.
 
             const requested_end = Math.min(params.end, reader.pos + requested_size);
+            if (requested_end <= reader.pos) {
```
jackyalbo, line 601: Just thinking here, maybe the logic will be clearer:
`if (requested_size === 0 || reader.pos >= params.end) {`
As this will give us more understanding that reading is finished

alphaprinz (approved):
Nice. LGTM.

re Jacky's comment, can't say I have a preference. Options are equivalent and both make sense.

tangledbytes (approved):

PR #9290, opened 2025-11-19
Added support for second access key s3 access

aayushchouhan09:
### Describe the Problem
Currently, we do not support second access key for s3 access.

### Explain the Changes
1.  Previously, authentication only checked the first access key (`access_keys[0]`), preventing IAM users from using their second access key for S3 operations. This fix updates `auth_server.js`, `account_server.js` and `account_util.js` to iterate through all access keys in the array.

1. JIRA: https://issues.redhat.com/browse/RHSTOR-8114

### Testing Instructions:
1. Create account: `nb account create <account-name>`
2. Create user under the above account created.
3. Create two access keys for the user.
4. Try accessing s3 apis using the user credentials (first and second both)
5. Also try deactivating the access keys.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Authentication now correctly matches the provided access key and uses that key's associated secret, enabling proper per-key secret resolution for accounts with multiple keys.
  * Account lookup by access key is more reliable — read requests now return the correct account information when queried by access key.

### DIFF
--- a/src/server/common_services/auth_server.js
+++ b/src/server/common_services/auth_server.js
@@ -178,7 +178,7 @@ function unauthorized_error(reason) {
  *
  */
 function create_access_key_auth(req) {
-    const access_key = req.rpc_params.access_key.unwrap();
+    const access_key = req.rpc_params.access_key;
     const string_to_sign = req.rpc_params.string_to_sign;
     const signature = req.rpc_params.signature;
 
@@ -186,19 +186,17 @@ function create_access_key_auth(req) {
         throw new RpcError('UNAUTHORIZED', 'signature error');
     }
 
-    const account = _.find(system_store.data.accounts, function(acc) {
-        if (acc.access_keys) {
-            return acc.access_keys[0].access_key.unwrap().toString() === access_key.toString();
-        } else {
-            return false;
-        }
-    });
+    const account = system_store.get_account_by_access_key(access_key);
 
     if (!account || account.deleted) {
         throw new RpcError('UNAUTHORIZED', 'account not found');
     }
 
-    const secret = account.access_keys[0].secret_key.unwrap().toString();
+    const key_pair = account.access_keys.find(key =>
+        key.access_key.unwrap() === access_key.unwrap()
+    );
+
+    const secret = key_pair.secret_key.unwrap();
     const signature_test = signature_utils.get_signature_from_auth_token({ string_to_sign: string_to_sign }, secret);
     if (signature_test !== signature) {
         throw new RpcError('UNAUTHORIZED', 'signature error');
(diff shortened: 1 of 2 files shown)


src/server/common_services/auth_server.js
```diff
@@ -187,18 +187,25 @@ function create_access_key_auth(req) {
...
     }
 
-    const secret = account.access_keys[0].secret_key.unwrap().toString();
+    const key_pair = account.access_keys.find(key =>
+        key.access_key.unwrap().toString() === access_key.toString()
+    );
+
+    if (key_pair.deactivated) {
+        throw new RpcError('UNAUTHORIZED', 'access key is deactivated');
+    }
```
shirady: What is the flow to get here?
When creating a new access key it is active by default.

src/server/common_services/auth_server.js
```diff
@@ -187,18 +187,25 @@ function create_access_key_auth(req) {
     }
 
     const account = _.find(system_store.data.accounts, function(acc) {
```
shirady: I think you could use the function you created in the past: `get_account_by_access_key`
I saw this in PR #8068

src/server/common_services/auth_server.js
aayushchouhan09: Root account will be able to deactivate the access_key manually.

aayushchouhan09: Not required check, we already checking this condition in `signature_utils.authorize_request_account_by_token` removed it. Thanks

Got confused because of the delay in account cache config loading

src/server/common_services/auth_server.js
```diff
@@ -178,27 +178,25 @@ function unauthorized_error(reason) {
...
     if (!account || account.deleted) {
         throw new RpcError('UNAUTHORIZED', 'account not found');
     }
 
-    const secret = account.access_keys[0].secret_key.unwrap().toString();
+    const key_pair = account.access_keys.find(key =>
+        key.access_key.unwrap() === access_key.unwrap()
+    );
+
+    const secret = key_pair.secret_key.unwrap().toString();
```
naveenpaul1: `.toString()` really needed?

src/util/account_util.js
```diff
@@ -283,11 +283,11 @@ function create_access_key_auth(req, account, is_iam) {
             account.access_keys = access_keys;
         }
         decrypted_access_keys = _.cloneDeep(account.access_keys);
-        account.access_keys[0] = {
-            access_key: account.access_keys[0].access_key,
```
naveenpaul1, line 287: I think we dont need to change here, this method is getting called from `create_account()` and access key always will be first item in that array.

src/server/common_services/auth_server.js
aayushchouhan09: we can remove it. Thanks

src/util/account_util.js
aayushchouhan09: Sure, Thanks

naveenpaul1 (approved):

PR #9291, opened 2025-11-19
IAM | No Bucket Policy Authorization for IAM Users

shirady:
### Describe the Problem
In case there is no bucket policy, we want users under the account to be authorized to perform S3 operations.

### Explain the Changes
1. In case there is no bucket policy - permit users under the account for S3 operations.

### Issues: 
List of GAPs:
1. This code does not handle a case where there is a user under an OBC account, in case we decide to add this layer.
2. The tests are only manual at this point; there is a plan to add automated tests after we have the design change.

### Testing Instructions:
1. Build the images and install the NooBaa system on Rancher Desktop (see [guide](https://github.com/noobaa/noobaa-operator/blob/master/doc/dev_guide/deply_noobaa_on_minikube_or_rancher_desktop.md)).
3. I'm using port-forward (in a different tab):
4. Create the alias for the admin - first, need to get the credentials: `nb status --show-secrets -n test1` and then:
5. Check the connection to the endpoint:
6. Create a user: `admin-iam iam create-user --user-name Robert`
7. Create access keys: `admin-iam iam create-access-keys --user-name Robert`
8. Create the alias for the user (like in step 4 with alias `user-2-s3`).
9. The case with no bucket policy, for example:
- The user can put an object in the bucket of the owner:
`echo 'test_data' | user-2-s3 s3 cp - s3://first.bucket/test_object.txt` (should work).
- The user can list objects in the bucket (to check the flow of the RPC call of `has_bucket_action_permission` function that was created in the `auth_server`):
`user-1-s3 s3 ls s3://first.bucket`

Code changes for testing:
```diff
const account_cache = new LRUCache({
    name: 'AccountCache',
-    expiry_ms: config.OBJECT_SDK_ACCOUNT_CACHE_EXPIRY_MS,
+   expiry_ms: 1, //SDSD 
```

Notes:
- I used the admin account, but for IAM operations, there is no difference between a created account and an admin account. I tested with the admin only because it saves steps (I have the admin account upon system creation and a bucket).

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added a public bucket_owner_id field to bucket info and policy responses.

* **Improvements**
  * Owner info now includes both account id and display name.
  * Authorization expanded so IAM accounts sharing a root owner can access buckets when no policy exists.
  * Ownership checks unified and made aware of containerized vs. non-containerized deployments.

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -238,6 +238,7 @@ async function authorize_request_policy(req) {
         s3_policy,
         system_owner,
         bucket_owner,
+        bucket_owner_id,
         owner_account,
         public_access_block,
     } = await req.object_sdk.read_bucket_sdk_policy_info(req.params.bucket);
@@ -253,8 +254,9 @@ async function authorize_request_policy(req) {
     }
 
     const account = req.object_sdk.requesting_account;
-    const account_identifier_name = req.object_sdk.nsfs_config_root ? account.name.unwrap() : account.email.unwrap();
-    const account_identifier_id = req.object_sdk.nsfs_config_root ? account._id : undefined;
+    const is_nc_deployment = req.object_sdk.nsfs_config_root;
+    const account_identifier_name = is_nc_deployment ? account.name.unwrap() : account.email.unwrap();
+    const account_identifier_id = is_nc_deployment ? account._id : undefined;
     const account_identifier_arn = s3_bucket_policy_utils.get_bucket_policy_principal_arn(account);
 
     // deny delete_bucket permissions from bucket_claim_owner accounts (accounts that were created by OBC from openshift\k8s)
@@ -286,8 +288,11 @@ async function authorize_request_policy(req) {
     if (!s3_policy) {
         // in case we do not have bucket policy
         // we allow IAM account to access a bucket that is owned by their root account
-        const is_iam_account_and_same_root_account_owner = account.owner !== undefined &&
-            owner_account && account.owner === owner_account.id;
+        let is_iam_account_and_same_root_account_owner = false;
+        if (account.owner !== undefined) {
+            const owner_account_to_compare = is_nc_deployment ? (owner_account && owner_account.id) : bucket_owner_id;
+            is_iam_account_and_same_root_account_owner = account.owner === owner_account_to_compare;
+        }
         if (is_owner || is_iam_account_and_same_root_account_owner) return;
         throw new S3Error(S3Error.AccessDenied);
     }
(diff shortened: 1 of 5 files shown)


src/server/common_services/auth_server.js
```diff
@@ -545,7 +545,13 @@ async function has_bucket_action_permission(bucket, account, action, req_query,
...
     const bucket_policy = bucket.s3_policy;
 
-    if (!bucket_policy) return is_owner;
+    if (!bucket_policy) {
+        // in case we do not have bucket policy
+        // we allow IAM account to access a bucket that that is owned by their root account
+        const is_iam_and_same_root_account_owner = account.owner !== undefined &&
+            account.owner._id.toString() === bucket.owner_account._id.toString();
+        return is_owner || is_iam_and_same_root_account_owner;
+    }
```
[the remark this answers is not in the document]
shirady: in the `auth_server.js` the account has a property of the `owner` is object ID and it can be referenced.
`account.owner` is a string when we read in the endpoint like you see in the file of `s3_rest.js`.

src/endpoint/s3/s3_rest.js
```diff
@@ -286,8 +287,8 @@ async function authorize_request_policy(req) {
     if (!s3_policy) {
         // in case we do not have bucket policy
         // we allow IAM account to access a bucket that is owned by their root account
-        const is_iam_account_and_same_root_account_owner = account.owner !== undefined &&
-            owner_account && account.owner === owner_account.id;
+        const owner_id_to_compare = req.object_sdk.nsfs_config_root ? owner_account?.id : bucket_owner_id;
```
naveenpaul1: I think its better to move this `req.object_sdk.nsfs_config_root` to a variable like `is_nsfs_nc`, I have seen repeated use of this, and will improve the readbility.

src/sdk/object_sdk.js
```diff
@@ -218,6 +218,7 @@ class ObjectSDK {
             s3_policy: bucket.s3_policy,
             system_owner: bucket.system_owner, // note that bucketspace_fs currently doesn't return system_owner
             bucket_owner: bucket.bucket_owner,
+            bucket_owner_id: bucket.bucket_owner_id,
```
naveenpaul1: please add comment

src/endpoint/s3/s3_rest.js
shirady: I added this, thank.

src/sdk/object_sdk.js
shirady: I added a comment.

naveenpaul1 (approved):

PR #9300, opened 2025-11-21
IAM | IAM integration tests

naveenpaul1:
### Describe the Problem

Both NC and containerized Noobaa will share the same IAM unit test.

### Explain the Changes
1. coretest updated with IAM server and port.
2. `test_nc_iam_basic_integration.js` renamed to `test_iam_basic_integration.js`
3. `test_iam_basic_integration` added to index.js

1. 

### Testing Instructions:
1. Run test in NC
    `NC_CORETEST=true ./node_modules/.bin/mocha src/test/integration_tests/api/iam/test_iam_basic_integration.js`
2. Run test
    `./node_modules/.bin/mocha src/test/integration_tests/api/iam/test_iam_basic_integration.js`

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * IAM endpoints now pass only the username when looking up users.
  * Access-key validation tightened to ensure keys exist before authenticating.
  * Account ownership checks made more reliable via a centralized owner-ID resolver (now exported).

* **Tests**
  * Enhanced IAM integration tests with environment-aware setup, conditional NC-specific flows, and credential handling.
  * Added isolated IAM HTTPS endpoint support and updated test entry points.

### DIFF
--- a/src/endpoint/iam/ops/iam_list_attached_user_policies.js
+++ b/src/endpoint/iam/ops/iam_list_attached_user_policies.js
@@ -21,7 +21,7 @@ async function list_attached_user_policies(req, res) {
 
     dbg.log1('To check that we have the user we will run the IAM GET USER', params);
     iam_utils.validate_params(iam_constants.IAM_ACTIONS.GET_USER, params);
-    await req.account_sdk.get_user(params);
+    await req.account_sdk.get_user({ username: params.username });
 
     dbg.log1('IAM LIST ATTACHED USER POLICIES (returns empty list on every request)', params);
 
--- a/src/sdk/accountspace_nb.js
+++ b/src/sdk/accountspace_nb.js
@@ -63,7 +63,6 @@ class AccountSpaceNB {
     }
 
     async get_user(params, account_sdk) {
-
         const requesting_account = system_store.get_account_by_email(account_sdk.requesting_account.email);
         return await account_sdk.rpc_client.account.get_user(params, requesting_account);
     }
--- a/src/test/integration_tests/api/iam/test_iam_basic_integration.js
+++ b/src/test/integration_tests/api/iam/test_iam_basic_integration.js
@@ -7,9 +7,10 @@ const path = require('path');
 const _ = require('lodash');
 const mocha = require('mocha');
 const assert = require('assert');
+const SensitiveString = require('../../../../util/sensitive_string');
 const fs_utils = require('../../../../util/fs_utils');
 const { TMP_PATH, generate_nsfs_account, get_new_buckets_path_by_test_env, generate_iam_client,
-         require_coretest } = require('../../../system_tests/test_utils');
+         require_coretest, is_nc_coretest } = require('../../../system_tests/test_utils');
 const { ListUsersCommand, CreateUserCommand, GetUserCommand, UpdateUserCommand, DeleteUserCommand,
         ListAccessKeysCommand, CreateAccessKeyCommand, GetAccessKeyLastUsedCommand,
         UpdateAccessKeyCommand, DeleteAccessKeyCommand,
@@ -28,37 +29,47 @@ const IamError = require('../../../../endpoint/iam/iam_errors').IamError;
 
 
 const coretest = require_coretest();
-const setup_options = { should_run_iam: true, https_port_iam: 7005, debug: 5 };
+let setup_options;
+if (is_nc_coretest) {
+    setup_options = { should_run_iam: true, https_port_iam: 7005, debug: 5 };
+}
 coretest.setup(setup_options);
 const { rpc_client, EMAIL, get_current_setup_options, stop_nsfs_process, start_nsfs_process } = coretest;
 
-const CORETEST_ENDPOINT_IAM = coretest.get_iam_https_address();
-
-const config_root = path.join(TMP_PATH, 'test_nc_iam');
-// on NC - new_buckets_path is full absolute path
-// on Containerized - new_buckets_path is the directory
-const new_bucket_path_param = get_new_buckets_path_by_test_env(config_root, '/');
-
 let iam_account;
+let account_res;
+let config_root;
 
 mocha.describe('IAM basic integration tests - happy path', async function() {
     this.timeout(50000); // eslint-disable-line no-invalid-this
 
     mocha.before(async () => {
         // we want to make sure that we run this test with a couple of forks (by default setup it is 0)
-        const current_setup_options = get_current_setup_options();
-        const same_setup = _.isEqual(current_setup_options, setup_options);
-        if (!same_setup) {
-            console.log('current_setup_options', current_setup_options, 'same_setup', same_setup);
-            await stop_nsfs_process();
-            await start_nsfs_process(setup_options);
+        if (is_nc_coretest) {
+            config_root = path.join(TMP_PATH, 'test_nc_iam');
+            // on NC - new_buckets_path is full absolute path
+            // on Containerized - new_buckets_path is the directory
+            const new_bucket_path_param = get_new_buckets_path_by_test_env(config_root, '/');
+
+            const current_setup_options = get_current_setup_options();
+            const same_setup = _.isEqual(current_setup_options, setup_options);
+            if (!same_setup) {
+                console.log('current_setup_options', current_setup_options, 'same_setup', same_setup);
+                await stop_nsfs_process();
+                await start_nsfs_process(setup_options);
+            }
+            await fs_utils.create_fresh_path(new_bucket_path_param);
+            await fs_utils.file_must_exist(new_bucket_path_param);
+            account_res = await generate_nsfs_account(rpc_client, EMAIL, new_bucket_path_param, { admin: true });
+        } else {
+            account_res = (await rpc_client.account.read_account({ email: EMAIL })).access_keys[0];
         }
 
         // needed details for creating the account (and then the client)
-        await fs_utils.create_fresh_path(new_bucket_path_param);
-        await fs_utils.file_must_exist(new_bucket_path_param);
-        const res = await generate_nsfs_account(rpc_client, EMAIL, new_bucket_path_param, { admin: true });
-        iam_account = generate_iam_client(res.access_key, res.secret_key, CORETEST_ENDPOINT_IAM);
+        const coretest_endpoint_iam = coretest.get_https_address_iam();
+        const access_key = account_res.access_key instanceof SensitiveString ? account_res.access_key.unwrap() : account_res.access_key;
+        const secret_key = account_res.secret_key instanceof SensitiveString ? account_res.secret_key.unwrap() : account_res.secret_key;
+        iam_account = generate_iam_client(access_key, secret_key, coretest_endpoint_iam);
     });
 
     mocha.after(async () => {
@@ -196,6 +207,8 @@ mocha.describe('IAM basic integration tests - happy path', async function() {
         });
 
         mocha.it('get access key (last used)', async function() {
+            // Skipping for containerized noobaa
+            if (!is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
             const input = {
                 AccessKeyId: access_key_id
             };
--- a/src/test/utils/coretest/coretest.js
+++ b/src/test/utils/coretest/coretest.js
@@ -66,6 +66,8 @@ let http_server;
 let https_address;
 let https_server_sts;
 let https_address_sts;
+let https_server_iam;
+let https_address_iam;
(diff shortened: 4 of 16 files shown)


src/sdk/accountspace_nb.js
```diff
@@ -63,7 +63,11 @@ class AccountSpaceNB {
     }
 
     async get_user(params, account_sdk) {
-
+        // Added to fix the coretest, ListAttachedUserPoliciesCommand 
+        // calling get_user with more param we allow and its failing RPC schema validation.
+        params = {
+            username: params.username,
+        };
```
shirady: It is not clear, in the file `src/endpoint/iam/ops/iam_get_user.js` we pass only the `username` why do you see more params?
https://github.com/noobaa/noobaa-core/blob/d3c091e0b95197f772c7703c478c56947f2bf325/src/endpoint/iam/ops/iam_get_user.js#L14-L20

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -32,33 +33,39 @@ const setup_options = { should_run_iam: true, https_port_iam: 7005, debug: 5 };
...
 
-const CORETEST_ENDPOINT_IAM = coretest.get_iam_https_address();
-
 const config_root = path.join(TMP_PATH, 'test_nc_iam');
 // on NC - new_buckets_path is full absolute path
 // on Containerized - new_buckets_path is the directory
 const new_bucket_path_param = get_new_buckets_path_by_test_env(config_root, '/');
 
 let iam_account;
+let res;
```
shirady: If you moved this variable as a global one, I think it would be better to give it a more meaningful name.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -32,33 +33,39 @@ const setup_options = { should_run_iam: true, https_port_iam: 7005, debug: 5 };
 coretest.setup(setup_options);
 const { rpc_client, EMAIL, get_current_setup_options, stop_nsfs_process, start_nsfs_process } = coretest;
```
shirady, line 38: Why did you move this `CORETEST_ENDPOINT_IAM`?

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -32,33 +33,39 @@ const setup_options = { should_run_iam: true, https_port_iam: 7005, debug: 5 };
...
+        } else {
+            res = (await rpc_client.account.read_account({ email: EMAIL })).access_keys[0];
         }
 
         // needed details for creating the account (and then the client)
-        await fs_utils.create_fresh_path(new_bucket_path_param);
-        await fs_utils.file_must_exist(new_bucket_path_param);
-        const res = await generate_nsfs_account(rpc_client, EMAIL, new_bucket_path_param, { admin: true });
-        iam_account = generate_iam_client(res.access_key, res.secret_key, CORETEST_ENDPOINT_IAM);
+        const CORETEST_ENDPOINT_IAM = coretest.get_iam_https_address();
```
shirady: I wasn't sure if you still wanted it with capital letters when it moved from global variable to local variable.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -196,6 +203,10 @@ mocha.describe('IAM basic integration tests - happy path', async function() {
         });
 
         mocha.it('get access key (last used)', async function() {
+            // Test case is wrong, get-access-key-last-used should have 
+            // access key belongs to same account that requesting.
+            // Skipping for containerized noobaa
+            if (!is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
             const input = {
                 AccessKeyId: access_key_id
             };
```
shirady: We can fix the test by adding `username` param, no?
Otherwise, we would not have a test case for `GetAccessKeyLastUsedCommand`.

```suggestion
            const input = {
            AccessKeyId: access_key_id
            UserName: username2
            };
```

src/test/utils/coretest/coretest.js
```diff
@@ -66,6 +66,9 @@ let http_server;
 let https_address;
 let https_server_sts;
 let https_address_sts;
+
```
shirady: You can remove the newline

src/test/utils/index/index.js
```diff
@@ -109,3 +109,6 @@ require('../../integration_tests/api/s3/test_lifecycle');
 // MD Sequence
 require('../../integration_tests/db/test_mdsequence');
 
+// Running with IAM port
+require('../../integration_tests/api/iam/test_iam_basic_integration'); // please notice that we use a different setup
```
shirady: > please notice that we use a different setup

It was in NC because there changes in the NSFS server.
Are you sure you need this comment in containerized?

src/test/utils/coretest/coretest.js
```diff
@@ -459,6 +478,10 @@ function get_https_address() {
 function get_https_address_sts() {
     return https_address_sts;
 }
+
+function get_iam_https_address() {
```
shirady: I would change the function name to `get_https_address_iam` (to be like STS).
It would need to change in the file `src/test/utils/coretest/nc_coretest.js` as well.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -32,33 +33,39 @@ const setup_options = { should_run_iam: true, https_port_iam: 7005, debug: 5 };
...
+            if (!same_setup) {
+                console.log('current_setup_options', current_setup_options, 'same_setup', same_setup);
+                await stop_nsfs_process();
+                await start_nsfs_process(setup_options);
+            }
+            await fs_utils.create_fresh_path(new_bucket_path_param);
+            await fs_utils.file_must_exist(new_bucket_path_param);
+            res = await generate_nsfs_account(rpc_client, EMAIL, new_bucket_path_param, { admin: true });
+        } else {
+            res = (await rpc_client.account.read_account({ email: EMAIL })).access_keys[0];
```
shirady: This would test the users under the admin only.
Do you want to test a user under a general account? (not admin) then we need to create this account.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
naveenpaul1: It was returned with the port empty for containerised flow. Not getting initialised when the flow reaches here.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
naveenpaul1: `get-access-key-last-used` dont have username param.
https://docs.aws.amazon.com/cli/latest/reference/iam/get-access-key-last-used.html
we need to add it when we add more test

src/test/integration_tests/api/iam/test_iam_basic_integration.js
naveenpaul1: This PR to create the basic test, We need to add more test later.

src/sdk/accountspace_nb.js
naveenpaul1: Methods like `ListAttachedUserPoliciesCommand` and `ListMFADevicesCommand` are calling `GetUser` internally, with parameter.
```
{ username: 'non-existing-user', marker: undefined, max_items: 100, iam_path_prefix: undefined }
````

Comment (shirady) [reply src/sdk/accountspace_nb.js]: I added this to return an empty list on many APIs.
We can change it so that the only parameter that they would send is the username.
We have this:
https://github.com/noobaa/noobaa-core/blob/2af53d07a98e4bc9b2d26af5ba77e96f5f67c27c/src/endpoint/iam/ops/iam_list_attached_user_policies.js#L24

you can change it to:
```diff
- await req.account_sdk.get_user(params);
+ await req.account_sdk.get_user(params.username);
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
shirady: If it is a matter of timing, this might also happen in the CI, maybe we need to wait for the endpoint to have this address (although it sound a bit weird that we didn't have it at this point...).

src/test/integration_tests/api/iam/test_iam_basic_integration.js
shirady: Thanks, you can skip this for both.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -32,33 +33,39 @@ const setup_options = { should_run_iam: true, https_port_iam: 7005, debug: 5 };
 coretest.setup(setup_options);
```
shirady: I think you need to change here, as some of the options are related to NC and not containerized.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
naveenpaul1: I havent seen this issue in CI, What are you suggesting? A wait loop?
I think it will happen only when you test only IAM test locally, I will keep it before.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
shirady: I don't think that the test case is wrong, you can run `get-access-key-last-used` on another user (from the same account, I just tested in AWS).
In case it fails in containerized then we need to fix it (in another PR). 
Suggesting to change the comment:
```suggestion
            // Skipping for containerized noobaa temporary
            if (!is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
            const input = {
                AccessKeyId: access_key_id
            };
```

src/endpoint/iam/ops/iam_list_attached_user_policies.js
```diff
@@ -21,7 +21,7 @@ async function list_attached_user_policies(req, res) {
 
     dbg.log1('To check that we have the user we will run the IAM GET USER', params);
     iam_utils.validate_params(iam_constants.IAM_ACTIONS.GET_USER, params);
-    await req.account_sdk.get_user(params);
+    await req.account_sdk.get_user({username: params.username});
```
shirady: missing spaces:

```suggestion
    await req.account_sdk.get_user({ username: params.username });
```

This change was in a couple of files, so there the spaces should be added too.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -28,37 +29,47 @@ const IamError = require('../../../../endpoint/iam/iam_errors').IamError;
 
 
 const coretest = require_coretest();
-const setup_options = { should_run_iam: true, https_port_iam: 7005, debug: 5 };
-coretest.setup(setup_options);
+let setup_options;
+if (is_nc_coretest) {
+    setup_options = { should_run_iam: true, https_port_iam: 7005, debug: 5 };
+    coretest.setup(setup_options);
+}
```
shirady, line 35: We do not need to call the `coretest.setup();` in coretest? (non nc_coretest)

src/util/account_util.js
```diff
@@ -333,19 +333,23 @@ function _check_if_account_exists(action, email_wrapped) {
...
+    let owner_account_id;
     if (typeof user_account.owner === 'object') {
         owner_account_id = String(user_account.owner._id);
     } else {
-        owner_account_id = user_account.owner._id;
+        owner_account_id = user_account.owner;
     }
-    return root_account_id === owner_account_id;
+    return owner_account_id;
 }
```
shirady: Could you move this function to `iam_utils` so we can reuse it?

src/test/integration_tests/api/iam/test_iam_basic_integration.js
shirady, line 76: This one also needs to be run only in `if (is_nc_coretest) `

src/util/account_util.js
```diff
@@ -333,21 +333,16 @@ function _check_if_account_exists(action, email_wrapped) {
...
-    if (typeof user_account.owner === 'object') {
-        owner_account_id = String(user_account.owner._id);
-    } else {
-        owner_account_id = user_account.owner._id;
-    }
+    const owner_account_id = get_owner_account_id(user_account);
     return root_account_id === owner_account_id;
 }
 
+
```
shirady, line 345: redundant newline

shirady (approved):
I added last 2 comment

PR #9301, opened 2025-11-21
Fix (NC | NSFS) - list_buckets allowing unauthorized bucket access

aayushchouhan09:
### Describe the Problem
Currently, bucket policy interfere with the list buckets which is not aws complaint.

### Explain the Changes
1. Handled `list_buckets()` with proper permissions and make it aws complaint.

NOTE: We do not support inline policy for iam users in NC and NSFS, after we support it we need to put iam policy to user to list the root account buckets.
`AWS_ACCESS_KEY_ID=<root-account-id> AWS_SECRET_ACCESS_KEY=<root-account-key> aws --no-verify-ssl --endpoint-url https://localhost:13443 iam put-user-policy   --user-name iam-aayush   --policy-name S3ListBucketsPolicy   --policy-document '{ "Version": "2012-10-17", "Statement": [ { "Effect": "Allow", "Action": [ "s3:ListAllMyBuckets", "s3:ListBucket", "s3:GetObject", "s3:PutObject" ], "Resource": "*" } ]}'
`

1. Issue: https://issues.redhat.com/browse/DFBUGS-4449
2. Related PR: #9272 

### Testing Instructions:
(NC | NSFS)
1.  Create account
2.  Create bucket using the above account credentials
3.  Try listing bucket --> it should list the owned buckets for the user
4.  Create iam user under above account and create access key
5.  Using iam user credentials try listing buckets --> it should list the above root account buckets (currently we do not support inline policy for NC/NSFS, after we support it `s3:ListAllMyBuckets` permission is required with inline policy to list root account buckets)

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Bucket listing and access now use ownership-aware checks (including IAM root inheritance) to gate visibility and actions.

* **New Features**
  * Account and bucket metadata include owner identifiers in the SDK/API.
  * NSFS directory access validation added to enforce filesystem-based access rules.

* **Refactor**
  * Permission logic consolidated into reusable ownership-first checks, with clearer ownership vs policy flow.

* **Tests**
  * Removed two bucket-listing policy tests; added/updated bucket-tagging tests with cross-account scenarios and setup/cleanup.

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -241,7 +241,7 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
      * permissions
      * a. First iteration - map_with_concurrency with concurrency rate of 10 entries at the same time.
      * a.1. if entry is dir file/entry is non json - we will return undefined
-     * a.2. if bucket is unaccessible by bucket policy - we will return undefined
+     * a.2. if bucket is not owned by the account (or their root account for IAM users) - we will return undefined
      * a.3. if underlying storage of the bucket is unaccessible by the account's uid/gid - we will return undefined
      * a.4. else - return the bucket config info.
      * b. Second iteration - filter empty entries - filter will remove undefined values produced by the map_with_concurrency().
@@ -271,9 +271,9 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
                 if (err.rpc_code !== 'NO_SUCH_BUCKET') throw err;
             }
             if (!bucket) return;
-            const bucket_policy_accessible = await this.has_bucket_action_permission(bucket, account, 's3:ListBucket');
-            dbg.log2(`list_buckets: bucket_name ${bucket_name} bucket_policy_accessible`, bucket_policy_accessible);
-            if (!bucket_policy_accessible) return;
+
+            if (!this.has_bucket_ownership_permission(bucket, account)) return;
+
             const fs_accessible = await this.validate_fs_bucket_access(bucket, object_sdk);
             if (!fs_accessible) return;
             return bucket;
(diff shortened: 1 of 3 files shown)


src/sdk/bucketspace_fs.js
```diff
@@ -921,24 +921,66 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
     ///// UTILS /////
     /////////////////
 
-    // TODO: move the following 3 functions - has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
+    // TODO: move the following 5 functions - is_bucket_owner(), is_iam_and_same_root_account_owner(), has_bucket_ownership_permission(), has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
```
shirady: If the functions are not used anymore, you can remove them.

src/test/unit_tests/nsfs/test_bucketspace_fs.js
```diff
@@ -519,28 +519,6 @@ mocha.describe('bucketspace_fs', function() {
             const res = await bucketspace_fs.list_buckets({}, dummy_object_sdk_for_iam_account);
             assert.equal(res.buckets.length, 0);
         });
-        mocha.it('list buckets - different account with bucket policy (principal by name)', async function() {
```
shirady, line 522: I suggest changing the test instead of removing it.
Change to another action (not list buckets).
I think it was mainly for checking different principals in bucket policy.

src/sdk/bucketspace_fs.js
aayushchouhan09: These are util functions and all are used, comment is about to move them in future for code reusability

src/test/unit_tests/nsfs/test_bucketspace_fs.js
aayushchouhan09: Sure I will edit them to work for different action.

src/sdk/bucketspace_fs.js
```diff
@@ -921,24 +921,66 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+     *
+     * aws-compliant behavior:
+     * - Root accounts can list buckets they own
+     * - IAM users can list their owner buckets
+     *
+     * @param {Record<string, any>} bucket
+     * @param {Record<string, any>} account
+     * @returns {Promise<boolean>}
+     */
+    async has_bucket_ownership_permission(bucket, account) {
```
jackyalbo: Am I missing something? why is this async? I guess you can also remove the await

src/sdk/bucketspace_fs.js
```diff
@@ -921,24 +921,66 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
     /////////////////
 
-    // TODO: move the following 3 functions - has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
+    // TODO: move the following 6 functions - is_bucket_owner(), is_iam_and_same_root_account_owner(), has_bucket_ownership_permission(), has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
     // so they can be re-used
+
+    /**
+     * is_bucket_owner checks if the account is the direct owner of the bucket
+     * @param {Record<string, any>} bucket
+     * @param {Record<string, any>} account
```
jackyalbo: Any reason not to use Account and Bucket from nb.d.ts

```suggestion
     * @param {nb.Bucket} bucket
     * @param {nb.Account} account
```

src/sdk/bucketspace_fs.js
aayushchouhan09: Removed, Thanks

src/sdk/bucketspace_fs.js
```diff
@@ -921,24 +921,66 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
     ///// UTILS /////
     /////////////////
 
-    // TODO: move the following 3 functions - has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
+    // TODO: move the following 6 functions - is_bucket_owner(), is_iam_and_same_root_account_owner(), has_bucket_ownership_permission(), has_bucket_action_permission(), validate_fs_bucket_access(), _has_access_to_nsfs_dir()
```
jackyalbo: I don't like keeping this updated after each PR, if we don't want to move it right now, let's create a new section // SHARED UTILS - with the original explanation - and keep everything that we think we want to move from this file there

aayushchouhan09: Moved, Thanks

src/sdk/nb.d.ts
```diff
@@ -63,6 +63,7 @@ interface System extends Base {
 
 interface Account extends Base {
     _id: ID;
+    owner: ID;
```
jackyalbo: do you know if owner is required for account? maybe @shirady can help
in the nsfs_account_schema it doesn't. but let's keep it the same there and here.

jackyalbo (commented):
LGTM, but I will let @shirady approve this PR as she have more knowledge about the expected list_buckets behavior.

src/sdk/nb.d.ts
shirady: not required in both schemas:
https://github.com/noobaa/noobaa-core/blob/1382037a986722a59a07346bd140243f21e11ead/src/server/system_services/schemas/nsfs_account_schema.js#L7-L16

https://github.com/noobaa/noobaa-core/blob/1382037a986722a59a07346bd140243f21e11ead/src/server/system_services/schemas/account_schema.js#L9-L14

```suggestion
    owner?: ID;
```

src/sdk/nb.d.ts
```diff
@@ -179,6 +180,7 @@ interface MirrorStatus {
 
 interface Bucket extends Base {
     _id: ID;
+    owner_account: ID;
```
shirady:
```suggestion
    owner_account?: ID;
```

Although it is required in NC bucket schema
https://github.com/noobaa/noobaa-core/blob/1382037a986722a59a07346bd140243f21e11ead/src/server/system_services/schemas/nsfs_bucket_schema.js#L7-L15

It is not required in the bucket schema 
https://github.com/noobaa/noobaa-core/blob/1382037a986722a59a07346bd140243f21e11ead/src/server/system_services/schemas/bucket_schema.js#L27-L32

shirady (approved):

PR #9302, opened 2025-11-23
IAM | Change Default Behavior of Users Without IAM User Policy

shirady:
### Describe the Problem
Currently in Containerised NooBaa deployment, when a user is created under an account (and has access keys), they can perform all S3 operations. Since we added the IAM User Policy APIs, we want to change the default so the user must have a policy for any S3 operations.

### Explain the Changes
1. Change `authorize_request` in  `s3_rest.js` so that the IAM policy check will be after `authorize_request_account` and before `authorize_request_policy`.
2. Check for IAM user policy in case it is not NC deployment.

### Issues: 
List of GAPs:
1. The tests are only manual at this point; there is a plan to add automated tests after we have the design change.
2. I might want to refactor the code so that `s3_bucket_policy_utils.js` so the terms will be policy in general and can be used for bucket policy and IAM policy.
3. I might want to move the created function somewhere else (perhaps `iam_rest`) so the thrown error will be `amError.AccessDeniedException` instead of `S3Error.AccessDenied`

### Testing Instructions:
3. I'm using port-forward (in a different tab):
4. Create the alias for the admin - first, need to get the credentials: `nb status --show-secrets -n test1` and then:
5. Check the connection to the endpoint:
6. Create a user: `admin-iam iam create-user --user-name Robert`
7. Create access keys: `admin-iam iam create-access-keys --user-name Robert`
8. Create the alias for the user (like in step 4 with alias `user-1-s3`): `alias user-1-s3='AWS_ACCESS_KEY=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:12443'`
9. The case with no bucket policy, no IAM policy - the user would be restricted from S3 operation, for example:
- The user lists objects in a bucket: `user-1-s3 s3 ls s3://first.bucket` (should not work - should throw `AccessDenied` error).
12. The admin adds an inline policy so the user can list objects in the bucket.
policy_allow_list_bucket.json
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:ListBucket"
            ],
            "Resource": "*"
        }
    ]
}
```
13. `admin-iam iam put-user-policy --user-name Robert --policy-name policy_allow_list_bucket --policy-document file://policy_allow_list_bucketjson`
14. Repeat step 9:
- The user lists objects in a bucket: `user-1-s3 s3 ls s3://first.bucket` (should work).

Code changes for testing:
```diff
const account_cache = new LRUCache({
    name: 'AccountCache',
-    expiry_ms: config.OBJECT_SDK_ACCOUNT_CACHE_EXPIRY_MS,
+   expiry_ms: 1, //SDSD 
```

Notes:
- I used the admin account, but for IAM operations, there is no difference between a created account and an admin account. I tested with the admin only because it saves steps (I have the admin account upon system creation and a bucket `first.bucket`).

- [X] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Documentation**
  * Clarified that an IAM user policy is required for S3 operations and updated ordering to show the IAM user policy evaluated before the bucket policy; minor wording tweaks.

* **Refactor**
  * Authorization checks now run sequentially rather than in parallel, changing control flow and error timing.
  * Tightened conditions under which IAM policy validation may be bypassed for specific deployments and added notes for future IAM handling.

### DIFF
--- a/docs/design/IamUserInlinePolicy.md
+++ b/docs/design/IamUserInlinePolicy.md
@@ -4,7 +4,7 @@ When used, it adds a layer of permission to the users under the account.
 We decided that IAM user inline policies are checked for authorization only in S3 operations (`src/endpoint/s3/s3_rest.js`).
 
 ## User Without IAM User Policy
-We decided that when a user is created under the account (and has access keys), it can operate all S3 operations (unless there is a bucket policy which do not authorize it).
+User must have IAM policy to be authorized for S3 operations.
 
 ## User With IAM User Policy
 The user’s inline policy is embedded in the user.  
@@ -29,8 +29,8 @@ If a user has a user policy, the ability to perform an S3 operation is based on 
 For every S3 request, authorization (`authorize_request` in `src/endpoint/s3/s3_rest.js`) is performed.
 The authorization now will have:
 1. Authorization handle for signed request and anonymous requests.  
-2. Authorization handle according to bucket policy.  
-3. Authorization handle according to the user IAM policy (the new added layer - only for IAM users).  
+2. Authorization handle according to the user IAM policy (the new added layer - only for IAM users).  
+3. Authorization handle according to bucket policy.  
 
 If one of the layers does not permit it would result in `AccessDenied` error.
 
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -217,14 +217,11 @@ function authenticate_request(req) {
 
 async function authorize_request(req) {
     await req.object_sdk.load_requesting_account(req);
-    await Promise.all([
-        req.object_sdk.authorize_request_account(req),
-        // authorize_request_policy(req) is supposed to
-        // allow owners access unless there is an explicit DENY policy
-        authorize_request_policy(req),
-        // authorize_request_iam_policy(req) is for users only
-        authorize_request_iam_policy(req),
-    ]);
+    await req.object_sdk.authorize_request_account(req);
+    await authorize_request_iam_policy(req); // authorize_request_iam_policy(req) is for users only
+    // authorize_request_policy(req) is supposed to
+    // allow owners access unless there is an explicit DENY policy
+    await authorize_request_policy(req);
 }
 
 async function authorize_request_policy(req) {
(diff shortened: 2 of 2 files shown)


src/endpoint/s3/s3_rest.js
```diff
@@ -348,7 +346,7 @@ async function authorize_request_iam_policy(req) {
     const resource_arn = _get_arn_from_req_path(req);
     const method = _get_method_from_req(req);
     const iam_policies = account.iam_user_policies || [];
-    if (iam_policies.length === 0) return;
+    if (iam_policies.length === 0 && req.object_sdk.nsfs_config_root) return; // We do not have IAM policies in NC yet
```
naveenpaul1, line 346: This PR more like disabling iam_policy authorize check for NC?

naveenpaul1 (approved):

src/endpoint/s3/s3_rest.js
shirady: Yes, we do not have an IAM user policy in NC, so the default behavior in NC is that the user can perform S3 operations after it has access keys.
In this PR, we change the default behavior in containerized deployments that must have an IAM user policy for S3 operations.

docs/design/IamUserInlinePolicy.md
```diff
@@ -4,7 +4,7 @@ When used, it adds a layer of permission to the users under the account.
 We decided that IAM user inline policies are checked for authorization only in S3 operations (`src/endpoint/s3/s3_rest.js`).
 
 ## User Without IAM User Policy
-We decided that when a user is created under the account (and has access keys), it can operate all S3 operations (unless there is a bucket policy which do not authorize it).
+User must have IAM policy to be authorized for S3 operations.
```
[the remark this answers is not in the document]
shirady: We don't have an IAM policy in NC, I thought it might be confusing for the reader.
IF he wants to read about IAM in NC the user can access it:
https://github.com/noobaa/noobaa-core/blob/c22df7108daab8391c435bb76ce41b191b06e458/docs/design/iam.md

PR #9309, opened 2025-11-24
IAM | Principal validation and S3 permission updated with ID

naveenpaul1:
### Describe the Problem

1. Bucket policy "Principal" validation was based on the email before and it updated with account/user ID
2. put_bucket_policy validate principal account/user ID
3. Bucket policy ARN related test added

#### Principal validation 
1. Validate account/user ID

#### S3 Rest authorize request policy
1. Account/user ID is added to request policy validation replacing account email. account/user ID based validation for containerized deployments NSFS NC.
2.  Updated test cases(test_s3_bucket_policy.js), previousily these tests where using email for principal for both NSFS non-containerized and acontainerized, With latest changes containerized will use account/user ID and ARN. Enabled account/user ID based test cases also for containerized integration test

GAP: IAM User test cases for ARN and user ID test cases are missing

### Testing Instructions:
Un integration test `test_s3_bucket_policy.js`
1. Create a bucket(test1) from account(account1)
2. Put bucket policy to newly created bucket()
```
s3api put-bucket-policy --bucket test1 --policy '{"Version": "2012-10-17","Statement": [ { "Effect": "Allow", "Principal": { "AWS": [ "${user2_id}" ] }, "Action": [ "s3:*" ], "Resource": [ "arn:aws:s3:::test1/*", "arn:aws:s3:::test1" ] } ]}'
```
`user2_id` could be ID of different account or IAM user ID of IAM user(not belongs to account1)
4. Verify the `user2` have `bucket1` access.

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate bucket policy enforcement: principal matching accepts multiple identifier forms and evaluates denies before allows to reduce incorrect denials.
  * Owner-account access checks added so owner-based permissions are correctly considered.

* **New Features**
  * Improved principal resolution across deployments so policies using root/user identifiers and ARNs are recognized.

* **Tests**
  * Expanded bucket policy tests to cover ARN principals and unified account-ID usage; teardown now conditional for NC core tests.

* **Chores**
  * Minor logging and formatting clarifications.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #9312, opened 2025-11-25
IAM | Principal in Bucket Policy of the Account Root User When the Requesting Account Is IAM User

shirady:
### Describe the Problem
Currently, bucket policy principal of ARN was for root user account only, after this change it would be effective for IAM users under this account as well.

### Explain the Changes
1. Add additional check after the check of bucket policy by principal ARN of the user to also check the permission in account level.
2. Fix `permission_by_arn` call to `has_bucket_policy_permission` so the last argument will be an object.

### Issues: 
List of GAPs:
1. This code does not handle a case where there is a user under an OBC account, in case we decide to add this layer.
3. The tests are only manual at this point; there is a plan to add automated tests after we have the design change.

### Testing Instructions:
1. Build the images and install the NooBaa system on Rancher Desktop (see [guide](https://github.com/noobaa/noobaa-operator/blob/master/doc/dev_guide/deply_noobaa_on_minikube_or_rancher_desktop.md)).
4. I'm using port-forward (in a different tab):
5. create 2 accounts:
- `nb account create shira-acc01 -n test1 --show-secrets`
- `nb account create shira-acc02 -n test1 --show-secrets`
6. Create the alias for the accounts:
`account-1-s3='AWS_ACCESS_KEY=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:12443'`
`account-2-s3='AWS_ACCESS_KEY=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:12443'`
`account-2-iam='AWS_ACCESS_KEY=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:14443'`
7. Check the connection to the endpoint:
- try to list the buckets: `account-1-s3 s3 ls; echo $?` `account-2-s3 s3 ls; echo $?`
8. Create a bucket for each account:
- `account-1-s3 s3 mb s3://buc-acc1`
- `account-2-s3 s3 mb s3://buc-acc2`
7. Create a user under account 2: `account-2-iam iam create-user --user-name Robert`
Note: To validate user creation, you can run `account-2-iam iam list-users` and expect 1 user in the list
9. Create access keys: `account-2-iam iam create-access-key --user-name Robert`
10. Add IAM policy to the user: `account-2-iam iam put-user-policy --user-name Robert --policy-name policy_allow_s3 --policy-document file://~/Documents/iam-tests/iam_policies/policy_allow_s3.json`

policy_allow_s3.json
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Resource": "*"
        }
    ]
}
```
11. Create the alias for the user (like in step 5 with alias `user-2-s3`): `alias user-1-s3='AWS_ACCESS_KEY=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://localhost:12443'`
12. The case with no bucket policy, for example:
- The user cannot put an object in the bucket in the second account:
`echo 'test_data' | user-1-s3  s3 cp - s3://buc-acc1/test_object.txt` (should not work)
- The user cannot list objects in the bucket (to check the flow of the RPC call of `has_bucket_action_permission` function that was created in the `auth_server`):
`user-1-s3 s3 ls s3://buc-acc1` (should not work)
13. Find the ID of account 2 for the ARN, for example by running: `account-2-s3 s3api get-bucket-acl --bucket buc-acc2` and taking the ID.
14. Add bucket policy of the root account user as principal:
`account-1-s3 s3api put-bucket-policy --bucket buc-acc1 --policy file://policy_principal_arn_owner.json`

policy_principal_arn_owner.json
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": [
                    "arn:aws:iam::69259fff8b07c200228c4fb3:root"
                ]
            },
            "Action": [
                "s3:*"
            ],
            "Resource": "*"
        }
    ]
}
```

15. The case no bucket policy, for example:
- The user can put an object in the bucket in the second account:
`echo 'test_data' | user-1-s3  s3 cp - s3://buc-acc1/test_object.txt` (should work)
- The user can list objects in the bucket (to check the flow of the RPC call of `has_bucket_action_permission` function that was created in the `auth_server`):
`user-1-s3 s3 ls s3://buc-acc1` (should not work)

Code changes for testing:
```diff
const account_cache = new LRUCache({
    name: 'AccountCache',
-    expiry_ms: config.OBJECT_SDK_ACCOUNT_CACHE_EXPIRY_MS,
+   expiry_ms: 1, //SDSD 
```

Notes:

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced bucket permission validation to include additional owner-based authorization checks, ensuring more comprehensive access control for S3 operations.

### DIFF
--- a/src/endpoint/s3/s3_rest.js
+++ b/src/endpoint/s3/s3_rest.js
@@ -250,7 +250,7 @@ async function authorize_request_policy(req) {
     }
 
     const account = req.object_sdk.requesting_account;
-    const is_nc_deployment = req.object_sdk.nsfs_config_root;
+    const is_nc_deployment = Boolean(req.object_sdk.nsfs_config_root);
     const account_identifier_name = is_nc_deployment ? account.name.unwrap() : account.email.unwrap();
     const account_identifier_id = is_nc_deployment ? account._id : undefined;
     const account_identifier_arn = s3_bucket_policy_utils.get_bucket_policy_principal_arn(account);
@@ -296,6 +296,7 @@ async function authorize_request_policy(req) {
     let permission_by_id;
     let permission_by_name;
     let permission_by_arn;
+    let permission_by_arn_owner;
 
     // In NC, we allow principal to be:
     // 1. account name (for backwards compatibility)
@@ -321,13 +322,26 @@ async function authorize_request_policy(req) {
     // Policy permission is validated by account arn
     if (!account_identifier_id) {
         permission_by_arn = await s3_bucket_policy_utils.has_bucket_policy_permission(
-            s3_policy, account_identifier_arn, method, arn_path, req, public_access_block?.restrict_public_buckets
+            s3_policy, account_identifier_arn, method, arn_path, req,
+            { disallow_public_access: public_access_block?.restrict_public_buckets }
         );
         dbg.log3('authorize_request_policy: permission_by_arn', permission_by_arn);
     }
     if (permission_by_arn === "DENY") throw new S3Error(S3Error.AccessDenied);
 
-    if ((permission_by_id === "ALLOW" || permission_by_name === "ALLOW" || permission_by_arn === "ALLOW") || is_owner) return;
+    // ARN check for users under the account
+    // ARN check is not implemented in NC yet
+     if (!is_nc_deployment && account.owner !== undefined) {
+        const owner_account_identifier_arn = s3_bucket_policy_utils.create_arn_for_root(account.owner);
+        permission_by_arn_owner = await s3_bucket_policy_utils.has_bucket_policy_permission(
+            s3_policy, owner_account_identifier_arn, method, arn_path, req,
+            { disallow_public_access: public_access_block?.restrict_public_buckets }
+        );
+        dbg.log3('authorize_request_policy permission_by_arn_owner', permission_by_arn_owner);
+        if (permission_by_arn_owner === "DENY") throw new S3Error(S3Error.AccessDenied);
+     }
+    if ((permission_by_id === "ALLOW" || permission_by_name === "ALLOW" ||
+        permission_by_arn === "ALLOW" || permission_by_arn_owner === "ALLOW") || is_owner) return;
 
     throw new S3Error(S3Error.AccessDenied);
 }
--- a/src/server/common_services/auth_server.js
+++ b/src/server/common_services/auth_server.js
@@ -640,7 +640,19 @@ async function has_bucket_action_permission(bucket, account, action, req_query, 
 
     if (result === 'DENY') return false;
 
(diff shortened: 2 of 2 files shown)


src/endpoint/s3/s3_rest.js
```diff
@@ -321,13 +322,26 @@ async function authorize_request_policy(req) {
...
+            s3_policy, account_identifier_arn, method, arn_path, req,
+            { disallow_public_access: public_access_block?.restrict_public_buckets }
         );
         dbg.log3('authorize_request_policy: permission_by_arn', permission_by_arn);
     }
     if (permission_by_arn === "DENY") throw new S3Error(S3Error.AccessDenied);
 
-    if ((permission_by_id === "ALLOW" || permission_by_name === "ALLOW" || permission_by_arn === "ALLOW") || is_owner) return;
+    // ARN check for users under the account
+    // ARN check is not implemented in NC yet
```
naveenpaul1, line 333: Can we move this to a different method?

src/endpoint/s3/s3_rest.js
```diff
@@ -321,13 +322,26 @@ async function authorize_request_policy(req) {
...
+     if (!is_nc_deployment && account.owner !== undefined) {
+        const owner_account_identifier_arn = s3_bucket_policy_utils.create_arn_for_root(account.owner);
+        permission_by_arn_owner = await s3_bucket_policy_utils.has_bucket_policy_permission(
+            s3_policy, owner_account_identifier_arn, method, arn_path, req,
```
naveenpaul1, line 337: FYI: Now we are checking against ARN, but we are supporting account ID aslo in bucket policy principal and I will add this also in  this PR [#9309](https://github.com/noobaa/noobaa-core/pull/9309)

src/endpoint/s3/s3_rest.js
shirady: I can, but I think that we need to refactor this function in general, I'm not sure that taking this part to a different function is enough.

naveenpaul1: That make sence, we will have refactoring task later.

naveenpaul1 (approved):

PR #9317, opened 2025-11-26
vectors - design doc draft

alphaprinz:


nimrod-becker (approved):

nimrod-becker (approved):

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+### Vector Buckets and Indexes
+
+1. Needed for minimal POC - create vector bucket, create index, put vector and query vector.
+2. MVP requires all get/list/delete API implemented.
+1. BucketPolicy is not a must for MVP.
+4. Hopefully will be able to reuse current bucket/object implementation with some adaptation.
+   1. Eg add "bucket content" field to indicate whether bucket is object or vector instead of creating a new “vector bucket” entity.
+
+### Terminology
+AWS S3 vectors and lance have the use the same term with different meaning.
```
guymguym, line 30: seems like some context is missing for the reader - better add some basic info about lancedb and that we plan to use it to implement s3vectors api before jumping into these details.

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+In AWS: a "Vector Bucket" hold several "Indexes". An "Index" holds several vectors.
+In Lance: a "Table" holds several vectors. An "Index" is created on data to accelerate search, similar to RDBM index.
+
+In the following description of S3 vectors api design, whenever "Bucket" or "Index" is referred, it should be explicitly mention whether it refers to the AWS or Lance meaning.
```
guymguym, line 35: from my perspective it's not really that different - you could just say that lance db doesn't have a bucket notion.

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+## S3 API
+### Vector Buckets and Indexes
+
+1. Needed for minimal POC - create vector bucket, create index, put vector and query vector.
+2. MVP requires all get/list/delete API implemented.
+1. BucketPolicy is not a must for MVP.
+4. Hopefully will be able to reuse current bucket/object implementation with some adaptation.
+   1. Eg add "bucket content" field to indicate whether bucket is object or vector instead of creating a new “vector bucket” entity.
+
+### Terminology
```
guymguym, line 12: add also the noobaa terminology here for a reader that is less familiar with it - 
NB vs FS bucketspace
NB vs FS namespace
refer to https://github.com/noobaa/noobaa-core/blob/master/docs/bucket-types.md

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+
+In the following description of S3 vectors api design, whenever "Bucket" or "Index" is referred, it should be explicitly mention whether it refers to the AWS or Lance meaning.
+
+### Vector Bucket
+
+#### Create
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_CreateVectorBucket.html
+
+This is a NOP for Lance implementation.
+Lance requires either an explicit schema or a specific vector in order to create a table.
```
guymguym, line 86: I think this kind of statement can be confusing without more context - by "Lance implementation" here you probably wanted to say that it doesn't need to do anything on the Lance plugin itself, but clearly I still need to use this op when using S3vectors API with Lance, and since there was no mention of the plugin system here, it's quite vague.

Perhaps it would be easier to understand if you add an "Architecture" section that describes these topics before getting into the API details:
1. Plugin system to support variety of backend solutions
2. LanceDB
3. GPU acceleration

A simple [mermaid diagram](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-diagrams) will be very useful here, here is something to start from:
```mermaid
flowchart TD
    A1[RAG]
    A2[Inference/Agent]
    B[S3 Vector Endpoint]
    B1[BucketSpace]
    C{{Vector Plugin System}}
    D1[LanceDB Plugin]
    D2[LanceDB API]
    E1[TBD... e.g. OpenSearch Plugin]
    E2[TBD... e.g. OpenSearch Vector API]
    F[FS Backend]
    G[S3 Backend]
    
    A2-->|Query|B
    A1-->|Ingest|B
    B-->C
    B-->|Bucket configs and policies|B1
    C-->D1
    C-->E1
    D1-->D2
    D2-->F
    D2-->G
    E1-->E2
    E2-.->F
    E2-.->G
```

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+
+For NB bucketspace, insert a new row in the VectorBuckets DB table.
+
+For FS bucketspace, create a new json file in the vector_buckets table.
+
+Currently only "vectorBucketName" is used. "encryptionConfiguration" and "tags" are ignored.
```
guymguym, line 92: notice that the api returns an ARN. worth mentioning what we'll return.

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### Get
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_GetVectorBucket.html
+
+Returns relevant data according to the relevant bucketspace.
```
guymguym:
```suggestion
Returns bucket information according to the relevant bucketspace.
```

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+Pagination with "nextToken" is not currently implemented. Evaluating ROI of implementing it should be evaluated at a later stage.
+
+#### Delete
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_DeleteVectorBucket.html
+
+Shall propagate delete request to relevant bucketspace.
+NB shall add "deleted" field to the relevant db row.
+FS shall delete the relevant json file.
+
+For NB, actual deletion will be implemeted in standard "DB Cleaner" BG worker.
```
guymguym, line 117: Is vector buckets also required to be empty to be deleted, like object buckets?

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+
+#### Delete
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_DeleteVectors.html
+
+In Lance:
+-translate keys into Lance "ids in" filter.
+-execute delete
+
+### Index
+TODO - needed for lance POC?
```
guymguym: I don't think it's possible to use a bucket without an index in S3 vectors API.

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+### Tags TODO
+### Policy TODO
+
+## Noobaa Middle layer
```
guymguym:
```suggestion
## Noobaa Rest API Layer
```

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+The operator will publish a new endpoint and service.
+
+### Op Handling
+#### Op Handler
+Each AWS S3 vectors api action listed in
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations_Amazon_S3_Vectors.html
+that we decide to support shall have a corresponding OP handler to handle it.
+An OP handler is supposed to be a thin layer translating HTTP req parameters into an object_sdk method.
+
+#### ObjectSDK
```
guymguym: I don't think ObjectSDK is related. 
If needed we should have a VectorSDK to route the operations to the configured plugin.

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+### Policy TODO
+
+## Noobaa Middle layer
+
+### Vector REST
+As a first layer of handling a vector request, this new http listener will read, parse, validate, authenticate, send down to op handler, and write out result (if any).
+Essentially similar to existing s3/iam/sts REST listeners, with different OP handlers.
+Particularities of vectors api should be handled here, specifically lack of body's sha256 header, and URL parsing (ie how to get op and vector bucket name from URL).
+The endpoint will register this new https listener.
+The operator will publish a new endpoint and service.
```
guymguym, line 211: Is this what the operator does also for sts/iam? a separate endpoint and service??

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+ObjectSDK shall propagate request to either one or both of-
+1. Relevant Bucketspace
+2. VectorUtils
+
+#### VectorUtils
+VectorUtils shall determine the VectorConnection used to handle the request.
+VectorConnections shall be stored in an LRU cache, creating and connecting a new connection lazily.
+VectorUtils shall propagate request to a connected VectorConnection.
+
+#### VectorConnection
```
guymguym:
```suggestion
#### VectorPlugin
```

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+VectorConnection is abstract, with a concrete implementation per vector backend (Lance, Davince, etc).
```
guymguym:
```suggestion
VectorPlugin is abstract, with a concrete implementation per vector backend (Lance, OpenSearch, etc).
```

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+VectorConnection translates parameters to vector backend api and calls appropriate apis on vector backend client.
+
+## BackingStorage/Storage?/VectorStorage?
+### CRDs
+We need to specify, at least in Lance client case, two kinds of independent parameters:
+-A storage connection - Can be an S3 account (endpoint url, secret id, secret key), or a FS based.
+This can be the alread-existing CRDs Backingstore and NamespaceStore.
```
guymguym: I think this should be based on a BucketClass, and not on the underlying BackingStore/NSStore.

docs/design/vectors.md
```diff
@@ -0,0 +1,170 @@
...
+For FS we can also use a pvc directly, without NamespaceStore?
```
guymguym: I think this is eventually what has to happen, because the DB itself will need to be initialized with the configuration of the storage which is either a FS path or an S3 connection.

docs/design/vectors.md
alphaprinz: In s3- Not exactly. It should have no Indexes.
BTW It is not specified if an index needs to be empty in order to be deleted.

docs/design/vectors.md
alphaprinz: Yes. They all appear in the list of services and endpoints in kubectl.
They list in system reconciler here
https://github.com/alphaprinz/noobaa-operator/blob/master/pkg/system/reconciler.go#L81-L83

docs/design/vectors.md
alphaprinz: IIUC BucketClass has a bunch of stuff relevant to objects but not for the vector use case, eg tiers and mirroring?
I just need a directory mounted on the endpoints, or an s3 connection credentials.

docs/design/vectors.md
alphaprinz: Thanks, copied into the doc :)

docs/design/vectors.md
alphaprinz: Added more explanations, hope it's better and enough.

docs/design/vectors.md
alphaprinz: Added that note explicitly, just to be safe

docs/design/vectors.md
alphaprinz: Yes, but specifically for Lance we said that CreateIndex s3 API will translate to creating an index in a Lance table.
At least for a minimal POC that is not a requirement.

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+
+### Architecture
+
+```mermaid
+flowchart TD
+    A1[RAG]
+    A2[Inference/Agent]
+    B[S3 Vector Endpoint]
+    B1[BucketSpace]
+    C{{Vector Plugin System}}
```
guymguym, line 58:
```suggestion
    C{{VectorSDK}}
```

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+    C-->E1
+    D1-->D2
+    D2-->F
+    D2-->G
+    E1-->E2
+    E2-.->F
+    E2-.->G
+```
+
+### Vector Bucket
```
guymguym, line 80:
```suggestion
### Operations on Vector Buckets
```

docs/design/vectors.md
guymguym: Empirically we can check how aws responds to deleting non-empty index. The doc might not be complete.

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+#### Delete
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_DeleteVectorBucket.html
+
+Shall propagate delete request to relevant bucketspace.
+NB shall add "deleted" field to the relevant db row.
+FS shall delete the relevant json file.
+
+For NB, actual deletion will be implemeted in standard "DB Cleaner" BG worker.
+
+### Vector
```
guymguym, line 119:
```suggestion
### Operations on Vectors
```

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+-translate vector format from Lance to aws
+
+#### Delete
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_DeleteVectors.html
+
+In Lance:
+-translate keys into Lance "ids in" filter.
+-execute delete
+
+### Index
```
guymguym, line 158:
```suggestion
### Operations on Vector Indexes
```

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+In Lance case, the technical difficulty here is to translate find a place for the Lance parameters in the AWS request body type.
+Since 
+-Lance index and S3 vectors index are essentially different AND
+-S3 vectors' CreateIndex request type is restricted,
+we can utilize the generic Tags field to put all data.
+
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_CreateIndex.html
+https://lancedb.github.io/lancedb/js/interfaces/IndexOptions/
+
+Other than that, Index API is a simple CRUD.
```
guymguym: For now we should add configs to pick the index type (HNSW / IVF) and configs to other default parameters in addition to the ones sent by S3 API (dataType, dimensions, distanceMetric).

Can we call [createIndex](https://lancedb.github.io/lancedb/js/classes/Table/#createindex) eachtime?

guymguym: Here is what I found on when to use LanceDB without an Index - 

Brute-force (no-index) search is suitable in specific scenarios: 
1. Small datasets: For datasets with up to around 100,000 records, a vector index is usually not necessary as the latency is acceptable (often under 100ms).
2. Guaranteed 100% recall: When you need to find the exact nearest neighbors without any approximation.
3. Immediate searchability of new data: When new data is added to a table, it is immediately searchable via brute-force even before a background indexing job has run. 

I think these are not our goals in S3 Vectors, so by default I would define an index.

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+    C{{Vector Plugin System}}
+    D1[LanceDB Plugin]
+    D2[LanceDB API]
+    E1[TBD... e.g. OpenSearch Plugin]
+    E2[TBD... e.g. OpenSearch Vector API]
+    F[FS Backend]
+    G[S3 Backend]
+    
+    A2-->|Query|B
+    A1-->|Ingest|B
```
guymguym, line 67:
```suggestion
    A1-->|Put|B
```

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+Similar to https://github.com/noobaa/noobaa-core/pull/9271
+Not compatible with AWS cli.
+
+3. Use account-level default (similar to account's default resource).
+This enables account-level granularity control (not per-bucket as needed).
+Can be combine as a default fall-back with above "Pure s3-compatible" option.
+
+4. New actions in operator cli (similar to OB in ODF, manage_nsfs in NSFS). Allows control on parameter names and values. Eg
+nb vector-bucket create --vector-storage=NS1 --vector-storage-path='/vectors1' --vector-plugin Lance
+We can add a new "VectorBucket" CRD or add fields to existing bucket's CRD.
```
guymguym, line 278: these two should be the first options to implement.

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+-add maxResults as limit to Lance query.
+-execute query
+-translate vectors format from Lance to aws. Add data or metadata as necessary.
+
+Currently without segments (pagination).
+
+#### Query
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_QueryVectors.html
+
+In Lance:
```
guymguym, line 144: missing "filter" "returnMetadata"

docs/design/vectors.md
guymguym: S3 CreateVectorIndex should be translated to Lance createTable + createIndex.

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+### Terminology
+#### AWS S3 Vectors
+As part of their S3 offering, AWS implements API for interacting with vectors.
+This includes creating a "vector bucket", which contains "indexes".
+Vectors can be inserted into indexes, and then queried to get a list of vectors closest to query vector.
+More in
+https://aws.amazon.com/s3/features/vectors/
+https://docs.aws.amazon.com/AmazonS3/latest/API/API_Operations_Amazon_S3_Vectors.html
+
+#### LanceDB
```
guymguym, line 21: need to mention the gap that lance metadata is columns with schema, whereas s3vectors metadata is a json (free schema), and we need to decide which solution to use to map between those.

docs/design/vectors.md
```diff
@@ -0,0 +1,255 @@
...
+#### VectorUtils
+VectorUtils shall determine the VectorPlugin used to handle the request.
+VectorPlugins instances shall be stored in an LRU cache, creating and connecting a new connection lazily.
+VectorUtils shall propagate request to a connected VectorPlugin.
+
+#### VectorPlugin
+VectorPlugin is abstract, with a concrete implementation per vector backend (Lance, Davinci, etc).
+VectorPlugin translates parameters to vector backend api and calls appropriate apis on vector backend client.
+
+## Vector storage
```
guymguym, line 237:
```suggestion
## Vector bucket configurations
```

PR #9318, opened 2025-11-27
Generate optimized SQL query to improve performance of find_chunks_by_dedup_key method

karthikps97:
### Describe the Problem
The mongo to pg query converter uses IN operator for array. But in case of postgres, using ANY operator improves performance as the query size is reduced and it also allows for plan caching. The sort is also applied on '_id' JSON field of data column, but there is _id column already present (primary key). 

### Explain the Changes
1. Refactored the find_chunks_by_dedup_key method to self generate SQL queries. Improvements are:
 - Using ANY instead of IN improves query planning time for larger array size. (Postgres internally converts IN to ANY, but it requires an additional parsing step) 
 - Using _id instead of data._id for sorting improves query execution time as we are utilizing the primary key index.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **Improvements**
  * Chunk deduplication retrieval now uses a SQL-backed path for broader DB support and improved resilience.
  * Collection objects now expose schema metadata for easier inspection.

* **Tests**
  * Added Postgres-only integration tests for deduplicated chunk lookup, missing-key and empty-key cases.
  * Many integration suites now skip Postgres-specific tests when not running against Postgres.

### DIFF
--- a/src/server/object_services/map_server.js
+++ b/src/server/object_services/map_server.js
@@ -85,8 +85,8 @@ class GetMapping {
         if (!config.DEDUP_ENABLED) return;
         await Promise.all(Object.values(this.chunks_per_bucket).map(async chunks => {
             const bucket = chunks[0].bucket;
-            const dedup_keys = _.compact(_.map(chunks,
-                chunk => chunk.digest_b64 && Buffer.from(chunk.digest_b64, 'base64')));
+            const dedup_keys = chunks.map(chunk => chunk.digest_b64).filter(Boolean);
+
             if (!dedup_keys.length) return;
             dbg.log0('GetMapping.find_dups: found keys', dedup_keys.length);
             const dup_chunks_db = await MDStore.instance().find_chunks_by_dedup_key(bucket, dedup_keys);
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -13,6 +13,7 @@ const mime = require('mime-types');
 const P = require('../../util/promise');
 const dbg = require('../../util/debug_module')(__filename);
 const db_client = require('../../util/db_client');
+const { decode_json } = require('../../util/postgres_client.js');
 
 const mongo_functions = require('../../util/mongo_functions');
 const object_md_schema = require('./schemas/object_md_schema');
@@ -1537,27 +1538,34 @@ class MDStore {
 
     /**
      * @param {nb.Bucket} bucket
-     * @param {nb.DBBuffer[]} dedup_keys
+     * @param {string[]} dedup_keys
      * @returns {Promise<nb.ChunkSchemaDB[]>}
      */
     async find_chunks_by_dedup_key(bucket, dedup_keys) {
-        // TODO: This is temporary patch because of binary representation in MongoDB and PostgreSQL
-        /** @type {nb.ChunkSchemaDB[]} */
-        const chunks = await this._chunks.find({
-            system: bucket.system._id,
-            bucket: bucket._id,
-            dedup_key: {
-                $in: dedup_keys,
-                $exists: true
-            },
-            deleted: null,
-        }, {
-            sort: {
-                _id: -1 // get newer chunks first
-            }
-        });
-        await this.load_blocks_for_chunks(chunks);
-        return chunks;
+        if (!dedup_keys?.length) return [];
+
+        const query = `
+            SELECT * 
+            FROM ${this._chunks.name} 
+            WHERE 
+                (data ->> 'system' = $1 
+                AND data ->> 'bucket' = $2 
+                AND (data ->> 'dedup_key' = ANY($3) 
+                AND data ? 'dedup_key') 
+                AND (data->'deleted' IS NULL OR data->'deleted' = 'null'::jsonb)) 
+            ORDER BY _id DESC;
+            `;
+        const values = [`${bucket.system._id}`, `${bucket._id}`, dedup_keys];
+
+        try {
+            const res = await this._chunks.executeSQL(query, values);
+            const chunks = res.rows.map(row => decode_json(this._chunks.schema, row.data));
+            await this.load_blocks_for_chunks(chunks);
+            return chunks;
+        } catch (err) {
+            dbg.error('Error while finding chunks by dedup_key. error is ', err);
+            throw err;
+        }
     }
 
     iterate_all_chunks_in_buckets(lower_marker, upper_marker, buckets, limit) {
--- a/src/test/integration_tests/db/test_md_store.js
+++ b/src/test/integration_tests/db/test_md_store.js
@@ -380,6 +380,42 @@ mocha.describe('md_store', function() {
             return md_store.delete_chunks_by_ids(_.map(chunks, '_id'));
         });
 
+        mocha.it('find_chunks_by_dedup_key()', async () => {
+            if (config.DB_TYPE !== 'postgres') return; // feature uses SQL path
+            const bucket = { _id: md_store.make_md_id(), system: { _id: system_id } };
+            const chunk = {
+                _id: md_store.make_md_id(),
+                system: system_id,
+                bucket: bucket._id,
+                frags: [{ _id: md_store.make_md_id() }],
+                size: 10,
+                frag_size: 10,
+                dedup_key: Buffer.from('noobaa')
+            };
+            await md_store.insert_chunks([chunk]);
+            const chunksArr = await md_store.find_chunks_by_dedup_key(bucket, [Buffer.from('noobaa').toString('base64')]);
+            assert(Array.isArray(chunksArr));
+            assert(chunksArr.length >= 1);
+            assert(chunksArr[0].frags[0]?._id?.toString() === chunk.frags[0]._id.toString());
+        });
+
+        mocha.it('test find_chunks_by_dedup_key - dedup_key doesnt exist in DB', async () => {
+            if (config.DB_TYPE !== 'postgres') return; // feature uses SQL path
+            const bucket = { _id: md_store.make_md_id(), system: { _id: system_id } };
+            const chunksArr = await md_store.find_chunks_by_dedup_key(bucket, [Buffer.from('unknownkey').toString('base64')]);
+            assert(Array.isArray(chunksArr));
+            assert(chunksArr.length === 0);
+        });
+
+        mocha.it('find_chunks_by_dedup_key empty dedup_key array passed', async () => {
+            if (config.DB_TYPE !== 'postgres') return; // feature uses SQL path
+            const bucket = { _id: md_store.make_md_id(), system: { _id: system_id } };
+
+            const chunksArr = await md_store.find_chunks_by_dedup_key(bucket, []);
+            assert(Array.isArray(chunksArr));
+            assert(chunksArr.length === 0);
+        });
+
     });
 
 
--- a/src/test/integration_tests/internal/test_encryption.js
+++ b/src/test/integration_tests/internal/test_encryption.js
@@ -31,6 +31,10 @@ const key_rotator = new KeyRotator({ name: 'kr'});
 
 config.MIN_CHUNK_AGE_FOR_DEDUP = 0;
 
+// @ts-ignore
+// Do not run below tests if DB is not PostgreSQL
+if (config.DB_TYPE !== 'postgres') return;
+
 mocha.describe('Encryption tests', function() {
     const { rpc_client, EMAIL, SYSTEM } = coretest;
     let response_account;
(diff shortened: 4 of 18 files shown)


liranmauda (changes requested):
On test_encription.js Is there any test that should run?
If not, we can do in line 33 (so you will not have to wrap the whole test.
Also we could add a comment for it that explain thta this test should not run if the DB is not SQL

if (config.DB_TYPE !== 'postgres') return;

src/test/integration_tests/internal/test_encryption.js
```diff
@@ -31,6 +31,10 @@ const key_rotator = new KeyRotator({ name: 'kr'});
 
 config.MIN_CHUNK_AGE_FOR_DEDUP = 0;
 
+// @ts-ignore
+// Do not run below tests if DB is not PostgreSQL
+if (config.DB_TYPE !== 'postgres') return;
+
```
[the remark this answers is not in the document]
liranmauda: It was a psedo-code
the actuall code should be something like:
```
mocha.describe('Encryption tests', function() {
  before(function () {
    if (config.DB_TYPE !== 'postgres') {
      this.skip(); 
    }
   ...
  });
 ```
 On any top level describe.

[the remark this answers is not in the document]
Comment (karthikps97) [reply src/test/integration_tests/internal/test_encryption.js:37]: Skipping the test using before hook would still run all the "after" hooks. This would throw errors specifically in the `unpopulate_system` function call as the initilaization would not have taken place. I would have to repeat the same skip statements in each 'after' hook. Wouldn't a single `return` or conditional `if` be better?

File: src/server/object_services/md_store.js [hunk 2]
```diff
@@ -1541,23 +1542,28 @@ class MDStore {
...
-        await this.load_blocks_for_chunks(chunks);
-        return chunks;
+        const values = [];
+        let query = `SELECT * FROM ${this._chunks.name} WHERE (data ->> 'system' = $1 AND data ->> 'bucket' = $2`;
+        values.push(`${bucket.system._id}`, `${bucket._id}`);
+
+        if (dedup_keys.length) {
+            query += ` AND (data ->> 'dedup_key' = ANY($3) AND data ? 'dedup_key')`;
+            values.push(dedup_keys);
+        } else {
```
jackyalbo: I don't think we suppose to have empty dedup keys.
https://github.com/noobaa/noobaa-core/blob/dae43d2eb8ae38e166a1bcc0be56ef9f1ae8d153/src/server/object_services/map_server.js#L90
we can avoid running the query in this case. just return in the beginning.

src/server/object_services/map_server.js
```diff
@@ -85,8 +85,12 @@ class GetMapping {
...
         await Promise.all(Object.values(this.chunks_per_bucket).map(async chunks => {
             const bucket = chunks[0].bucket;
-            const dedup_keys = _.compact(_.map(chunks,
-                chunk => chunk.digest_b64 && Buffer.from(chunk.digest_b64, 'base64')));
+            const dedup_keys = [];
+            chunks.forEach(chunk => {
+                if (chunk?.digest_b64) {
+                    dedup_keys.push(chunk.digest_b64);
+                }
+            });
```
jackyalbo:
```suggestion
            const dedup_keys = chunks.map(chunk => chunk.digest_b64).filter(Boolean);
```

src/server/object_services/md_store.js
```diff
@@ -1541,23 +1542,28 @@ class MDStore {
...
-            deleted: null,
-        }, {
-            sort: {
-                _id: -1 // get newer chunks first
-            }
-        });
-        await this.load_blocks_for_chunks(chunks);
-        return chunks;
+        const values = [];
+        let query = `SELECT * FROM ${this._chunks.name} WHERE (data ->> 'system' = $1 AND data ->> 'bucket' = $2`;
```
jackyalbo: I'm thinking maybe working with @Neon-White format for this queries making them more readable. @dannyzaken WDYT?

jackyalbo (commented):
Also can you share a comparison of before and after to this PR description?

src/server/object_services/md_store.js
jackyalbo: https://github.com/noobaa/noobaa-core/blob/dae43d2eb8ae38e166a1bcc0be56ef9f1ae8d153/src/server/object_services/md_store.js#L307-L317
for example

src/server/object_services/md_store.js
karthikps97:
```text
Result: Tested with 1M+ rows in datachunks, with the query filtering 100 dedup_keys.

nbcore=# select count(*) from datachunks
  count  

---------
 1088000

**Current query:**

EXPLAIN ANALYZE SELECT * FROM datachunks WHERE (data->>'system'='693124b23742600022f30a64' and data->>'bucket'='693124b23742600022f30a6c' and (data->>'dedup_key' IN ('YLE95BklbH7n0c5/0i4OlEn2v6QPsweb8IcZf8BmXHc=', 
'YLE95BklbH7n0c5/0i4OlEn2v6QPsweb8IcZf8BmXHc=',
....(removed for readability)....
'zsViCcfiavZaRGB4G5fbGq6DzPccEIprV8JbESaB8zE=') and data ? 'dedup_key') and (data->'deleted' IS NULL OR data->'deleted' = 'null'::jsonb)) ORDER BY data->>'_id' DESC;

Sort  (cost=169.30..169.30 rows=1 width=1454) (actual time=0.448..0.452 rows=100 loops=1)
   Sort Key: ((data ->> '_id'::text)) DESC
   Sort Method: quicksort  Memory: 169kB
   ->  Index Scan using idx_btree_datachunks_dedup_key_system_bucket on datachunks  (cost=0.55..169.29 rows=1 width=1454) (actual time=0.030..0.375 rows=100 loops=1)
         Index Cond: (((data ->> 'dedup_key'::text) = ANY ('{YLE95BklbH7n0c5/0i4OlEn2v6QPsweb8IcZf8BmXHc=,YLE95BklbH7n0c5/0i4OlEn2v6QPsweb8IcZf8BmXHc=,wntv6uTo3IhN1BwepJqGTe493C0pB88j4SMsB92UrvQ=,
....(removed for readability)....
aO3VxeMSCz/041jhXpKEc+CPFTCO++Q9Q9WdaTF093k=,zsViCcfiavZaRGB4G5fbGq6DzPccEIprV8JbESaB8zE=}'::text[])) AND ((data ->> 'system'::text) = '693124b23742600022f30a64'::text) AND ((data ->> 'bucket'::text) = '693124b23742600022f30a6c'::text))
 Planning Time: 0.125 ms
 Execution Time: 0.471 ms
(8 rows)


**With changes:**

EXPLAIN ANALYZE SELECT * FROM datachunks WHERE (data->>'system'='693124b23742600022f30a64' and data->>'bucket'='693124b23742600022f30a6c' and (data->>'dedup_key'=ANY(ARRAY['YLE95BklbH7n0c5/0i4OlEn2v6QPsweb8IcZf8BmXHc=', 
'YLE95BklbH7n0c5/0i4OlEn2v6QPsweb8IcZf8BmXHc=',
....(removed for readability)....
'zsViCcfiavZaRGB4G5fbGq6DzPccEIprV8JbESaB8zE=']) 
and data ? 'dedup_key') and (data->'deleted' IS NULL OR data->'deleted' = 'null'::jsonb)) ORDER BY _id DESC;


Sort  (cost=169.29..169.30 rows=1 width=1422) (actual time=0.433..0.437 rows=100 loops=1)
   Sort Key: _id DESC
   Sort Method: quicksort  Memory: 167kB
   ->  Index Scan using idx_btree_datachunks_dedup_key_system_bucket on datachunks  (cost=0.55..169.28 rows=1 width=1422) (actual time=0.030..0.368 rows=100 loops=1)
         Index Cond: (((data ->> 'dedup_key'::text) = ANY ('{YLE95BklbH7n0c5/0i4OlEn2v6QPsweb8IcZf8BmXHc=,YLE95BklbH7n0c5/0i4OlEn2v6QPsweb8IcZf8BmXHc=,wntv6uTo3
IhN1BwepJqGTe493C0pB88j4SMsB92UrvQ=,z8R8VzzFUaX+10H+JDzb/5qGh10IPUuGbVInOuuTz3Q=,FzlSOeTfPmi1upqzMe6ZfhlYa+V0MzB2nKQqPZ9pj9Q=,
....(removed for readability)....
aO3VxeMSCz/041jhXpKEc+CPFTCO++Q9Q9WdaTF093k=,zsViCcfiavZaRGB4G5fbGq6DzPccEIprV8JbESaB8zE=}'::text[])) AND ((data ->> 'system'::text) = '693124b23742600022f30a64'::text) AND ((data ->> 'bucket'::text) = '693124b23742600022f30a6c'::text))
 Planning Time: 0.119 ms
 Execution Time: 0.450 ms
```

src/server/object_services/md_store.js
karthikps97: Formatted the query as suggested.

liranmauda (approved):

src/server/object_services/md_store.js
```diff
@@ -1541,23 +1542,30 @@ class MDStore {
...
+        const values = [`${bucket.system._id}`, `${bucket._id}`, dedup_keys];
+
+        try {
+            const res = await this._chunks.executeSQL(query, values);
+            const chunks = res?.rows?.map(row => decode_json(this._chunks.schema, row.data));
+            await this.load_blocks_for_chunks(chunks);
+            return chunks;
+        } catch (err) {
+            dbg.error('Error while finding chunks by dedup_key. error is ', err);
+            return [];
```
dannyzaken: We should not catch all errors here and ignore them, as it can cause silent failures that are hidden in the logs. It's a good idea to log the error and rethrow it.

src/test/integration_tests/internal/test_encryption.js
```diff
@@ -31,6 +31,10 @@ const key_rotator = new KeyRotator({ name: 'kr'});
 
 config.MIN_CHUNK_AGE_FOR_DEDUP = 0;
 
+// @ts-ignore
+// Do not run below tests if DB is not PostgreSQL
+if (config.DB_TYPE !== 'postgres') return;
```
dannyzaken, line 36: @karthikps97 @liranmauda, why is this needed?

src/test/integration_tests/db/test_md_store.js
```diff
@@ -380,6 +380,34 @@ mocha.describe('md_store', function() {
...
+                dedup_key: Buffer.from('noobaa')
+            };
+            await md_store.insert_chunks([chunk]);
+            const chunksArr = await md_store.find_chunks_by_dedup_key(bucket, [Buffer.from('noobaa').toString('base64')]);
+            assert(Array.isArray(chunksArr));
+            assert(chunksArr.length >= 1);
+            assert(chunksArr[0].frags[0]?._id?.toString() === chunk.frags[0]._id.toString());
+        });
+
+        mocha.it('find_chunks_by_dedup_key empty dedup_key array passed', async () => {
```
dannyzaken, line 410: Why does this specific case need to be tested? We don't expect an empty array, IIRC

src/test/integration_tests/db/test_md_store.js
```diff
@@ -380,6 +380,34 @@ mocha.describe('md_store', function() {
             return md_store.delete_chunks_by_ids(_.map(chunks, '_id'));
         });
 
+        mocha.it('find_chunks_by_dedup_key()', async () => {
```
dannyzaken, line 383: worth adding more tests for different cases. e.g.: test that we return an empty array when there is no match, etc.

src/test/integration_tests/internal/test_encryption.js
karthikps97: These tests are executed as part of both PostgreSQL tests and MongoDB tests. We are skipping this for the MongoDB run as it will fail since executeSQL is not implemented by mongo client.

src/test/integration_tests/db/test_md_store.js
karthikps97: We return early from the function in case an empty dedup_keys array is passed. So, added this test case.

src/test/integration_tests/db/test_md_store.js
karthikps97: Added one test for this scenario.

src/server/object_services/md_store.js
karthikps97: I'm rethrowing the error now rather than ignoring.

dannyzaken (approved):

PR #9329, opened 2025-12-02
IAM | User ID for principal is not supported

naveenpaul1:
### Describe the Problem

User ID for principal is not supported

### Explain the Changes
1. Block put pucket policy if the principal is IAM user ID

1. 

### Testing Instructions:
1. 

GAP: Missing integration test for IAM users, Will add in upcoming PRs

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added a deployment-aware helper to determine when an account identifier should be used for bucket policy evaluation.

* **Bug Fixes**
  * Improved bucket policy principal resolution: IAM users are no longer treated as valid principals when looked up by account ID, reducing unintended principal matches.
  * Policy evaluation now respects deployment-specific account identification rules for more accurate access control.

### DIFF
--- a/src/endpoint/s3/s3_bucket_policy_utils.js
+++ b/src/endpoint/s3/s3_bucket_policy_utils.js
@@ -378,6 +378,19 @@ function get_bucket_policy_principal_arn(account) {
 }
 
 /**
+ *  Both NSFS NC and containerized will validate bucket policy against acccount id 
+ *  but in containerized deplyment not against IAM user ID.
+ * 
+ * @param {boolean} is_nc_deployment
+ * @param {object} account
+ */
+function get_account_identifier_id(is_nc_deployment, account) {
+    if (is_nc_deployment || account.owner === undefined) {
+        return account._id;
+    }
+}
+
+/**
  * create_arn_for_root creates the AWS ARN for root account user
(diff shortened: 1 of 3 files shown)


shirady:
@naveenpaul1,
What about the `s3_rest` file?
It still checks the `account_id` for IAM user

src/endpoint/s3/s3_bucket_policy_utils.js
```diff
@@ -377,6 +377,19 @@ function get_bucket_policy_principal_arn(account) {
...
 
+/**
+ *  Both NSFS NC and containerized will validate bucket policy against acccount id 
+ *  but in containerized deplyment not against IAM user ID.
+ * 
+ * @param {boolean} is_nc_deployment
+ * @param {object} account
+ */
+function get_account_identifier_id(is_nc_deployment, account) {
+    if (is_nc_deployment || (!is_nc_deployment && account.owner === undefined)) {
```
shirady: I'm not sure that you need this part `!is_nc_deployment`:
in case: `is_nc_deployment = true`, enter the condition on the first part
in case: `is_nc_deployment = false` enter the condition on the second part only

shirady (approved):

PR #9330, opened 2025-12-02
IAM | Remove Unneeded `system_store` Calls - Part 1

shirady:
### Describe the Problem
Currently, there are many calls to `system_store` in functions related to IAM.

### Explain the Changes
1. Reduce the calls to `system_store`, mainly where it was to get the `requesting_account` in the file `accountspace_nb.js`.

### Issues:
List of GAPs:
- I think we need to move the `create_user` inside the `account_server` so we would have the `req.account` access and keep this file with small functionality.
- Planned to work on the files: `account_server` and `account_utils` as well.
- There is code duplication that I found; probably we can reduce more in the future.

### Testing Instructions:
I rerun the API tests to make sure it doesn't break.

#### Automatic test
Containerized
1. Use the guide "Run Tests From coretest Locally" ([link](https://github.com/noobaa/noobaa-core/blob/master/docs/dev_guide/run_coretest_locally.md)) for running the core tests.
4. Run first tab: `docker run -p 5432:5432 -e POSTGRESQL_ADMIN_PASSWORD=noobaa  quay.io/sclorg/postgresql-15-c9s`
5. Run second tab: `NOOBAA_LOG_LEVEL=all ./node_modules/.bin/mocha src/test/integration_tests/api/iam/test_iam_basic_integration.js`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal account/user handling and streamlined backend request flow by removing redundant internal argument passing.
  * Consolidated ownership/resource resolution and cleaned up related labels and comments.
  * No changes to public APIs, method signatures, or any user-visible behavior.

### DIFF
--- a/src/sdk/accountspace_nb.js
+++ b/src/sdk/accountspace_nb.js
@@ -1,17 +1,8 @@
-/* Copyright (C) 2024 NooBaa */
+/* Copyright (C) 2025 NooBaa */
 'use strict';
 
 const account_util = require('../util/account_util');
-const system_store = require('..//server/system_services/system_store').get_instance();
 const { IAM_DEFAULT_PATH} = require('../endpoint/iam/iam_constants');
-
-/* 
-    TODO: DISCUSS: 
-    1. IAM API only for account created using IAM API and OBC accounts not from admin, support, 
-       operator and account created using noobaa.
-    2. Do we need to have two access keys
-    3. get_access_key_last_used() API call could return dummy values?
-*/
 
 /**
  * @implements {nb.AccountSpace}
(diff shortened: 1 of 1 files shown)


src/sdk/accountspace_nb.js
```diff
@@ -35,22 +26,21 @@ class AccountSpaceNB {
     //////////////////////
 
     async create_user(params, account_sdk) {
-
-        const requesting_account = system_store.get_account_by_email(account_sdk.requesting_account.email);
-        const account_email_wrapped = account_util.get_account_email_from_username(params.username, requesting_account._id.toString());
+        const requesting_account = account_sdk.requesting_account;
+        const account_email_wrapped = account_util.get_account_email_from_username(params.username, requesting_account._id);
```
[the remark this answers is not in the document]
shirady: Is it the requesting_account object that was read, so the ID is a string.
In case it helps you. I printed those lines:
```
        console.log('SDSD requesting_account._id', requesting_account._id);
        console.log('SDSD check', requesting_account._id.toString() === requesting_account._id);
```
And you can see:
```
Dec-4 10:43:44.900 [Endpoint/13]   [LOG] CONSOLE:: SDSD requesting_account._id 6931655fbd35d3002209e3cc
Dec-4 10:43:44.900 [Endpoint/13]   [LOG] CONSOLE:: SDSD check true
```
Anyway, if you find this more readable to run the `toString()` on a string I can do that...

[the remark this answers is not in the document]
shirady: Adding `.toString()` would help readability for others?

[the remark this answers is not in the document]
shirady: ok, so I will remove this.

naveenpaul1 (approved):

src/sdk/accountspace_nb.js
```diff
@@ -63,110 +51,86 @@ class AccountSpaceNB {
     }
 
     async get_user(params, account_sdk) {
-        const requesting_account = system_store.get_account_by_email(account_sdk.requesting_account.email);
-        return await account_sdk.rpc_client.account.get_user(params, requesting_account);
+        return account_sdk.rpc_client.account.get_user(params);
```
naveenpaul1, line 54: why `await` removed?

shirady: It is redundant.
In more detail ([link](https://github.com/noobaa/noobaa-core/pull/9330#pullrequestreview-3549016071)):

> …Since they don’t add logic around the call, you can simplify slightly by dropping the redundant await
> …
> (and similarly for update_user, delete_user, list_users)
> 78-98: Access-key RPC wrappers match the new API shape; consider DRYing or dropping await
> ...
> Remove await and just return account_sdk.rpc_client.account.<method>(params);…

shirady: in case it helps, link in StackOverFlow about the case where it might e used (not our case) -
https://stackoverflow.com/questions/38708550/difference-between-return-await-promise-and-return-promise

naveenpaul1: sincw all the unit test case is passing, LGTM

PR #9331, opened 2025-12-03
Added a fix for authentication of root and iam user account

aayushchouhan09:
### Describe the Problem
When an IAM user tried to use list-access-keys on itself with the `--user-name` flag, it resulted in an AccessDenied error. However, the same operation without --user-name flag worked fine.

### Explain the Changes
1. Added logic to correctly identify when an IAM user is operating on themselves, whether or not they provide the `--user-name` flag.

1. Issue: `list-access-keys` was not working as expected for user with `--user-name` flag.

### Testing Instructions:
1. Create account
2. Create iam user and its access key
3. Using the iam user credentials try listing access keys(using below command) --> should work
`$ AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url https://192.168.58.2:31309 iam list-access-keys  --user-name <user-name>`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Unified and strengthened account-access validation so self-service and cross-account operations are resolved consistently, reducing authorization errors and preventing improper access.
  * More reliable resolution logic with consistent return behavior across scenarios, while preserving public API behavior (no signature changes).

### DIFF
--- a/src/util/account_util.js
+++ b/src/util/account_util.js
@@ -696,25 +696,31 @@ function validate_create_account_params(req) {
 }
 
 function validate_and_return_requested_account(params, action, requesting_account) {
-    const on_itself = !params.username;
-        let requested_account;
-        if (on_itself) {
-            // When accesskeyt API called without specific username, action on the same requesting account.
-            // So in that case requesting account and requested account is same.
-            requested_account = requesting_account;
-            // we do not allow for AWS account root user to perform IAM action on itself
-            if (requesting_account.owner === undefined) {
-                throw new RpcError('NOT_AUTHORIZED', 'You do not have permission to perform this action.');
-            }
-        } else {
-            _check_if_requesting_account_is_root_account(action, requesting_account, { username: params.username });
-            const account_email = get_account_email_from_username(params.username, requesting_account._id.toString());
-            _check_if_account_exists(action, account_email, params.username);
-            requested_account = system_store.get_account_by_email(account_email);
-            _check_if_requested_account_is_root_account_or_IAM_user(action, requesting_account, requested_account);
-            _check_if_requested_is_owned_by_root_account(action, requesting_account, requested_account);
+    const requester_username = requesting_account.name.unwrap();
+    // check if root account or IAM user is operating on themselves (with or without --user-name flag)
+    const no_username_or_self_operation = !params.username; // can be root account or IAM user
+    const is_iam_user_operating_on_itself = !_check_root_account(requesting_account) &&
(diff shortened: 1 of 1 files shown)


src/util/account_util.js
```diff
@@ -695,25 +695,30 @@ function validate_create_account_params(req) {
...
+    const on_itself = no_username_or_self_operation || is_iam_user_operating_on_itself;
+
+    let requested_account;
+    if (on_itself) {
+        // When accesskeyt API called without specific username, action on the same requesting account.
+        // So in that case requesting account and requested account is same.
+        requested_account = requesting_account;
+        // we do not allow for AWS account root user to perform IAM action on itself
+        if (requesting_account.owner === undefined) {
+            throw new RpcError('NOT_AUTHORIZED', 'You do not have permission to perform this action.');
```
shirady, line 713: As discussed in Slack, I'm fine with removing this as you suggested and guard the admin credentials in a different way.

shirady (approved):

src/util/account_util.js
```diff
@@ -695,25 +695,30 @@ function validate_create_account_params(req) {
...
-        return requested_account;
+    } else {
+        _check_if_requesting_account_is_root_account(action, requesting_account, { username: params.username });
+        const account_email = get_account_name_from_username(params.username, requesting_account._id.toString());
+        _check_if_account_exists(action, account_email);
+        requested_account = system_store.get_account_by_email(account_email);
+        _check_if_requested_account_is_root_account_or_IAM_user(action, requesting_account, requested_account);
+        _check_if_requested_is_owned_by_root_account(action, requesting_account, requested_account);
+    }
+    return requested_account;
```
naveenpaul1, line 723: Is there any chance`requested_account ` is undefinde when try with IAM user on itself? I see we getting accesskye from `requested_account` withour null check

src/util/account_util.js
aayushchouhan09: I think, we should do this in separate PR and as per the discussion (ongoing) in slack

src/util/account_util.js
aayushchouhan09: added undefined check in the start, Thanks

naveenpaul1 (approved):

src/util/account_util.js
aayushchouhan09: removed it as it is not required.

PR #9332, opened 2025-12-04
IAM | Doc | Update file `docs/design/AWS_API_Compatibility.md`

shirady:
### Describe the Problem
Update the `docs/design/AWS_API_Compatibility.md` and take the IAM and STS to a different table where the comparison is between deployments instead of namespacestore/backingstore types.

### Explain the Changes
1. Move the STS API to the deployment comparison - change the NC column to not implemented.
2. Move the IAM API to the deployment comparison.
3. Add new actions that were implemented in Containerized deployment.

### Issues:
1. none

### Testing Instructions:
1. none

- [X] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Documentation**
  * Reorganized AWS API compatibility guide with a new "IAM and STS API" section and added "S3 API" subsection for clearer discovery.
  * Consolidated IAM and STS action listings into a single, reformatted table showing per-deployment support (containerized vs non-containerized), including pagination and partial-implementation notes.
  * Clarified deployment port differences, manual enablement for non-containerized deployments, headings, footnotes, and overall readability.
  * No changes to public API behavior or capabilities.

### DIFF
--- a/docs/design/AWS_API_Compatibility.md
+++ b/docs/design/AWS_API_Compatibility.md
@@ -1,10 +1,13 @@
 [NooBaa Core](../../README.md) /
 
 # AWS APIs Compatibility Table
-The chart below strives to provide an up-to-date overview of which AWS API calls are supported by NooBaa, and to what extent.
-Actions that are absent from the table are entirely unsupported*.
-The table is split into categories, each containing a list of features and their corresponding API actions.
-NooBaa utilizes different implementations of API calls for different storage providers, so the table is split into columns for each provider to indicate whether it's supported or not.
+The chart below strives to provide an up-to-date overview of which AWS API calls are supported by NooBaa, and to what extent.  
+Actions that are absent from the table are entirely *unsupported*.  
+
+## S3 API
+The table is split into categories, each containing a list of features and their corresponding API actions.  
+NooBaa utilizes different implementations of API calls for different storage providers, so the table is split into columns for each provider to indicate whether it's supported or not.  
+
 The store types currently included in the table are backingstore (regardless of storage provider), namespace filesystem (NSFS), namespace for Amazon Web Services, and namespace for Microsoft Azure.
 For more information, see [S3 Compatibility](https://github.com/noobaa/noobaa-operator/tree/master/doc/s3-compatibility.md), [Bucket Types](https://github.com/noobaa/noobaa-operator/tree/master/doc/bucket-types.md), [Backingstore CRD](https://github.com/noobaa/noobaa-operator/blob/master/doc/backing-store-crd.md) and [Namespacestore CRD](https://github.com/noobaa/noobaa-operator/blob/master/doc/namespace-store-crd.md).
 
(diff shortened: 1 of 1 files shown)


docs/design/AWS_API_Compatibility.md
```diff
@@ -71,16 +74,32 @@ _* Note that it is also possible for actions to be supported but absent because
...
+|             | Users          | UpdateUser           | ✅ | ✅ | |
+|             | Users          | DeleteUser           | ✅ | ✅ | |
+|             | Users          | ListUsers            | ✅ | ✅ | *No pagination support|
+|             | Access Keys    | CreateAccessKey      | ✅ | ✅ | |
+|             | Access Keys    | GetAccessKeyLastUsed | ✅ | ✅ | *Partially implemented|
+|             | Access Keys    | UpdateAccessKey      | ✅ | ✅ | |
+|             | Access Keys    | DeleteAccessKey      | ✅ | ✅ | |
+|             | Access Keys    | ListAccessKeys       | ✅ | ✅ | *No pagination support|
+|             | Tag User       | TagUser              | ✅ | ❌ | | |
+|             | Tag User       | UntagUser            | ✅ | ❌ | | |
```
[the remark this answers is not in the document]
shirady: Why is it still open?
Should be resolved, right?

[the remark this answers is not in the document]
shirady: What about the `Comments` column? It is saved for this.

docs/design/AWS_API_Compatibility.md
```diff
@@ -71,16 +74,32 @@ _* Note that it is also possible for actions to be supported but absent because
...
-|                       | Access Keys                     | CreateAccessKey                   | ❌            | ✅    | ❌      | ❌        |                                                                         |
-|                       | Access Keys                     | UpdateAccessKey                   | ❌            | ✅    | ❌      | ❌        |                                                                         |
-|                       | Access Keys                     | DeleteAccessKey                   | ❌            | ✅    | ❌      | ❌        |                                                                         |
-|                       | Access Keys                     | ListAccessKeys                    | ❌            | ✅*   | ❌      | ❌        | *No pagination support                                                  |
 
-_* IAM API uses a different port than the S3 API, and needs to be manually enabled prior to use._
\ No newline at end of file
+
+## IAM and STS API
```
naveenpaul1, line 79: Do we need to add supported parameters?

shirady: We can add those in the IAM doc (like we have in NC - [link](https://github.com/noobaa/noobaa-core/blob/master/docs/design/iam.md#supported-actions-and-their-request-parameters)) and then add a link here if you want.
I like that this table is simple to understand with minimal information.

docs/design/AWS_API_Compatibility.md
```diff
@@ -71,16 +74,32 @@ _* Note that it is also possible for actions to be supported but absent because
...
+The following table compare the actions of managing entities between NooBaa deployments:
+1. NooBaa non-containerized deployment (NC)
+2. Containerized deployment
+
+**IAM API uses a different port than the S3 API (and needs to be manually enabled before use in NC deployments).**
+
+| Category              | Feature                         | API Action                        | Containerized  | NC   | Comments                                                                |
+|:---------------------:|:-------------------------------:|:---------------------------------:|:-------------:|:------:|-------------------------------------------------------------------------|
+| **STS API** | Session Tokens | AssumeRole           | ✅ | ❌ | |
```
shirady, line 86: @jackyalbo could you look at this line?

1. I also moved STS (and not just IAM) to a deployment comparison, if you don't think it makes sense, I would move only IAM.
2. I also thought that you added `AssumeRoleWithWebIdentity` and maybe you want to add this one also.

jackyalbo: As we are using AssumeRoleWithWebIdentity just for LDAP which is not supported by AWS, I think we can skip it here. 
About 1, I don't really mind putting STS here also - and as you know much more about how much of STS we support, I'll let you take the call :).

shirady: Thanks, I will leave it as I suggested here.

docs/design/AWS_API_Compatibility.md
naveenpaul1: yes, It would be better to share a link here.

shirady: I added a link in the notes.

naveenpaul1 (approved):

docs/design/AWS_API_Compatibility.md
```diff
@@ -71,16 +74,34 @@ _* Note that it is also possible for actions to be supported but absent because
...
-|                       | Access Keys                     | DeleteAccessKey                   | ❌            | ✅    | ❌      | ❌        |                                                                         |
-|                       | Access Keys                     | ListAccessKeys                    | ❌            | ✅*   | ❌      | ❌        | *No pagination support                                                  |
 
-_* IAM API uses a different port than the S3 API, and needs to be manually enabled prior to use._
\ No newline at end of file
+
+## IAM and STS API
+The following table compare the actions of managing entities between NooBaa deployments:
+1. NooBaa non-containerized deployment (NC)
+2. Containerized deployment
```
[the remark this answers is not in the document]
shirady: Why is it still open?

PR #9336, opened 2025-12-07
IAM | Refactor `src/endpoint/s3/s3_bucket_policy_utils.js` (Used in Bucket Policy and IAM User Inline Policy)

shirady:
### Describe the Problem
Currently, our `src/endpoint/s3/s3_bucket_policy_utils.js` is used for bucket policy in several places, and not all of them are in the endpoint S3 stack (for example, NC CLI). In addition, when I added the IAM user inline policy, I reused those functions.

### Explain the Changes
1. Renaming:
- In the logs: `bucket_policy -> access_policy`
- The function: `has_bucket_policy_permission -> has_access_policy_permission`
2.  Move `s3_bucket_policy_utils.js` from `src/endpoint/s3/` to `src/util/`.
3.  Renaming:
- The file: `s3_bucket_policy_utils.js` to `access_policy_utils.js`
- The variable name: `s3_bucket_policy_utils` to `acess_policy_utils`

### Issues: 
List of GAPs:
- We'd better also move the `const s3_utils = require('../endpoint/s3/s3_utils');` as it is in the endpoint S3 stack.

### Testing Instructions:
1. Tested in the CI.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Refactor**
  * Unified and renamed internal policy utilities across the system to consolidate policy validation and permission checks; public APIs and user-facing behavior unchanged.

* **Tests**
  * Updated tests to reference the consolidated policy utilities; test behavior and outcomes remain the same.

### DIFF
--- a/src/manage_nsfs/manage_nsfs_validations.js
+++ b/src/manage_nsfs/manage_nsfs_validations.js
@@ -9,7 +9,7 @@ const P = require('../util/promise');
 const string_utils = require('../util/string_utils');
 const native_fs_utils = require('../util/native_fs_utils');
 const ManageCLIError = require('../manage_nsfs/manage_nsfs_cli_errors').ManageCLIError;
-const bucket_policy_utils = require('../endpoint/s3/s3_bucket_policy_utils');
+const access_policy_utils = require('../util/access_policy_utils');
 const { throw_cli_error, get_options_from_file, get_boolean_or_string_value, get_bucket_owner_account_by_id,
     is_name_update, is_access_key_update } = require('../manage_nsfs/manage_nsfs_cli_utils');
 const { TYPES, ACTIONS, VALID_OPTIONS, OPTION_TYPE, FROM_FILE, BOOLEAN_STRING_VALUES, BOOLEAN_STRING_OPTIONS,
@@ -487,7 +487,7 @@ async function validate_bucket_args(config_fs, data, action) {
         }
         if (data.s3_policy) {
             try {
-                await bucket_policy_utils.validate_s3_policy(data.s3_policy, data.name,
+                await access_policy_utils.validate_bucket_policy(data.s3_policy, data.name,
                     async principal => config_fs.is_account_exists_by_principal(principal, { silent_if_missing: true })
                 );
             } catch (err) {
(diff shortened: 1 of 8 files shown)


liranmauda:
Why do we have 4 commits?

shirady:
> Why do we have 4 commits?

@liranmauda I broke it into a couple of commits to make it easier to review.
Anyway, I didn't invite reviewers yet.
When I merge, I plan to have a single commit.

jackyalbo:
@shirady Any specific reason for this name? document_policy? why not just policy? or access_policy?

shirady:
> @shirady Any specific reason for this name? document_policy? why not just policy? or access_policy?

Since it is a JSON file that we check.
I took the name from the IAM User inline put operation ([link](https://docs.aws.amazon.com/IAM/latest/APIReference/API_PutUserPolicy.html)).
In bucket policy it is a single JSON that is attached to a bucket, in IAM user inline policy it can be multiple JSONs on a single user.
I searched for a general name.
* policy  - seems too short.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -487,7 +487,7 @@ async function validate_bucket_args(config_fs, data, action) {
         }
         if (data.s3_policy) {
             try {
-                await bucket_policy_utils.validate_s3_policy(data.s3_policy, data.name,
+                await access_policy_utils.validate_s3_policy(data.s3_policy, data.name,
```
jackyalbo: maybe 
```suggestion
                await access_policy_utils.validate_bucket_policy(data.s3_policy, data.name,
```
WDYT?

jackyalbo (approved):
LGTM

src/manage_nsfs/manage_nsfs_validations.js
jackyalbo: I think one of two options: if you will share this code with IAM, remove the S3. If not, I think validate_bucket_policy will sound better than s3_policy.

shirady: The user's inline policy is not validated.
I renamed the function as you suggested.

PR #9359, opened 2026-01-05
Added a fix for 'NotPrincipal' bucket policy check

aayushchouhan09:
### Describe the Problem
Bucket policy `NotPrincipal` was not working correctly. When an account was specified in `NotPrincipal`, it was incorrectly denied access because the authorization code checked account identifiers (ID, Name, ARN) separately. If the policy used ID format but the second check used ARN format, the string comparison failed and the account was not excluded from the Deny statement.
`Principal` worked correctly because the code used OR logic for ALLOW (if ANY check returned ALLOW, access was granted), but used short-circuit logic for DENY (throw on first DENY). 
This issue caused `NotPrincipal` to fail when format mismatch occurred in any single check.

### Explain the Changes
1. Combined all account identifiers (ID + Name for NC, ID + ARN for containerized) into a single array and pass them together to `has_access_policy_permission()` instead of making separate calls for each identifier format.

1. Fixed: https://issues.redhat.com/browse/DFBUGS-1519

### Testing Instructions:
1. Create three accounts - test-a, test-b, test-c
2. Create bucket `test-buck` with `test-a` creds.
3. Apply policy to the `test-buck` (using `test-a` creds, also please update command as per your resources) : 

`$ AWS_ACCESS_KEY_ID=<access-key-a> AWS_SECRET_ACCESS_KEY=<secret-key-a> aws --no-verify-ssl --endpoint-url <url> s3api put-bucket-policy --bucket test-buck --policy '{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "s3:*"
            ],
            "Principal": {
                "AWS": ["<test-b>", "<test-c>"]
            },
            "Resource": ["arn:aws:s3:::test-buck", "arn:aws:s3:::test-buck/*"]
        },
        {
            "Effect": "Deny",
            "Action": ["s3:GetObject"],
            "NotPrincipal": { "AWS": "<test-b>"},
            "Resource": ["arn:aws:s3:::test-buck", "arn:aws:s3:::test-buck/*"]
        }
    ]
}'`

4. Upload an object to the bucket (using `test-a` creds).
5. Try copying (GetObject op) that object using creds of all three accounts --> except for `test-b` GetObject op should fail (after the fix)

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * DENY rules are now consistently enforced across deployments; owner IAM access is evaluated correctly.
  * Policies using NotPrincipal with Effect Allow are rejected as malformed.

* **Refactor**
  * Consolidated permission evaluation into a single, consistent flow with unified account identifier handling.

* **Tests**
  * Added/activated NotPrincipal Deny tests and adjusted test cleanup to delete buckets before account teardown.

### DIFF
--- a/src/test/integration_tests/api/s3/test_s3_bucket_policy.js
+++ b/src/test/integration_tests/api/s3/test_s3_bucket_policy.js
@@ -1630,40 +1630,61 @@ mocha.describe('s3_bucket_policy', function() {
         }));
     });
 
-    mocha.it.skip('should be able to use notPrincipal', async function() {
-        // This test is broken - Effect Allow can't be used with NotPrincipal
-        // copied from the docs:
-        // "NotPrincipal must be used with "Effect":"Deny".
-        // Using it with "Effect":"Allow" is not supported."
+    mocha.it('should deny access using NotPrincipal with Effect Deny', async function() {
+        // NotPrincipal must be used with "Effect":"Deny"
         // source: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_notprincipal.html
         const self = this; // eslint-disable-line no-invalid-this
         self.timeout(15000);
-        const auth_put_policy = {
-        Version: '2012-10-17',
-        Statement: [
-            {
-                Effect: 'Allow',
-                NotPrincipal: { AWS: a_principal },
-                Action: ['s3:PutObject'],
-                Resource: [`arn:aws:s3:::${BKT}/*`]
+
+        const policy = {
+            Version: '2012-10-17',
+            Statement: [
+                {
+                    Effect: 'Allow',
+                    Principal: { AWS: [a_principal, b_principal] },
+                    Action: ['s3:*'],
+                    Resource: [`arn:aws:s3:::${BKT_D}`, `arn:aws:s3:::${BKT_D}/*`]
+                },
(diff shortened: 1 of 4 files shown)


shirady (approved):
LGTM

src/test/integration_tests/api/s3/test_s3_bucket_policy.js
```diff
@@ -1630,40 +1630,34 @@ mocha.describe('s3_bucket_policy', function() {
         }));
     });
 
-    mocha.it.skip('should be able to use notPrincipal', async function() {
```
nadavMiz, line 1633: why did you remove this test. if Allow can't be used with NotPrincipal, wouldn't it be better to change it so we check that we don't allow it, instead of removing it altogether

nadavMiz (approved):
small comment, other than that LGTM

src/test/integration_tests/api/s3/test_s3_bucket_policy.js
aayushchouhan09: Yes, make sense. Added a check and test for `Allow` with `NotPrincipal`. Thanks

PR #9360, opened 2026-01-07
[NC | GLACIER] Fix no progress in processing Glacier logs when multiple processes are involved

tangledbytes:
### Describe the Problem
When for any reason (most common when and tested one is `glacier migrate` taking too long), there are a lot of processes trying to acquire the lock to process the log files, they tend to get stuck. 

### Explain the Changes
This PR removes the locks completely and simply adds a test by checking whether the lock is placed or not. If it is placed all the processes would SKIP processing that log file assuming that there MUST be another process already consuming the log file and hence they all would move forward to attempt to process other log files.

This not only ensures that migrate doesn't gets stuck but also ensures that we can keep processing new migrate requests that are coming while the older ones are getting processed.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added an asynchronous file-lock query API (returns EXCLUSIVE, SHARED, or null) and updated types.

* **Bug Fixes**
  * Log processing now skips files that are exclusively locked.
  * File close is now awaited to ensure consistent async behavior.

* **Refactor**
  * Removed per-stage locking; simplified to cluster-wide lock handling.

* **Chores**
  * Build trigger updated to consider native filesystem changes for base-image rebuilds.

### DIFF
--- a/.github/workflows/run-pr-tests.yaml
+++ b/.github/workflows/run-pr-tests.yaml
@@ -85,7 +85,7 @@ jobs:
       - name: Should build noobaa base image
         id: should_build_base
         run: |
-          base_files=("package.json" "base.dockerfile" ".nvmrc")
+          base_files=("package.json" "src/deploy/NVA_build/Base.Dockerfile" ".nvmrc" "src/native/fs/fs_napi.cpp")
           output=false
           for file in ${{ steps.changed_files.outputs.all_changed_files }}; do
             if printf '%s\n' "${base_files[@]}" | grep -x -q "$file"; then
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -1471,6 +1471,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
                 InstanceMethod<&FileWrap::fsync>("fsync"),
                 InstanceMethod<&FileWrap::flock>("flock"),
                 InstanceMethod<&FileWrap::fcntllock>("fcntllock"),
+                InstanceMethod<&FileWrap::fcntlgetlock>("fcntlgetlock"),
                 InstanceAccessor<&FileWrap::getfd>("fd"),
             }));
         constructor.SuppressDestruct();
@@ -1503,6 +1504,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
     Napi::Value getfd(const Napi::CallbackInfo& info);
     Napi::Value flock(const Napi::CallbackInfo& info);
     Napi::Value fcntllock(const Napi::CallbackInfo& info);
+    Napi::Value fcntlgetlock(const Napi::CallbackInfo& info);
 };
 
 Napi::FunctionReference FileWrap::constructor;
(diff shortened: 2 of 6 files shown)


src/util/persistent_logger.js
```diff
@@ -112,16 +112,44 @@ class PersistentLogger {
...
         } catch (error) {
             dbg.error('failed reading dir:', this.dir, 'with error:', error);
             return;
         }
 
+        /**
+         * @param {import('fs').Dirent<string>} file 
+         * @returns {Promise<boolean>}
+         */
+        const check_locked = async file => {
```
guymguym, line 130: if you're worried that this might have false positives perhaps instead of a one shot check, try to check 3 times with 1 second delay.

tangledbytes: Sounds good! Added it. I added a bit of randomness to ensure multiple processes if stuck here don't check (probabilistically very low chance) at the same time.

.github/workflows/run-pr-tests.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
       - name: Should build noobaa base image
         id: should_build_base
         run: |
-          base_files=("package.json" "base.dockerfile" ".nvmrc")
+          base_files=("package.json" "src/deploy/NVA_build/Base.Dockerfile" ".nvmrc" "src/native/fs/fs_napi.cpp")
```
guymguym, line 88: This makes me wonder - does base image has all the native modules built already?
if so then this list might need more than just fs_napi..?

src/sdk/nb.d.ts
```diff
@@ -1041,6 +1041,7 @@ interface NativeFile {
     fd: number;
     flock(fs_context: NativeFSContext, operation: "EXCLUSIVE" | "SHARED" | "UNLOCK"): Promise<void>;
     fcntllock(fs_context: NativeFSContext, operation: "EXCLUSIVE" | "SHARED" | "UNLOCK"): Promise<void>;
+    fcntlgetlock(fs_context: NativeFSContext): Promise<"EXCLUSIVE" | "SHARED" | null>;
```
guymguym, line 1044: I think we should avoid mixing null ("object") and string in return type by either 
1. using undefined is better than null because at least it doesn't carry a misleading "object" type
2. using literal string "UNLOCK" to match the previous flock/fcntllock signatures

src/native/fs/fs_napi.cpp
```diff
@@ -2011,6 +2013,51 @@ struct FileFcntlLock : public FSWrapWorker<FileWrap>
...
+
+        Begin(XSTR() << "FileFcntlGetLock" << DVAL(_wrap->_path));
+    }
+    virtual void Work()
+    {
+        int fd = _wrap->_fd;
+        CHECK_WRAP_FD(fd);
+        if (fcntl(fd, F_OFD_GETLK, &fl) == -1) {
+            SetSyscallError();
+        };
```
guymguym, line 2045: does SYSCALL_OR_RETURN macro not work here?

src/native/fs/fs_napi.cpp
```diff
@@ -1471,6 +1471,7 @@ struct FileWrap : public Napi::ObjectWrap<FileWrap>
                 InstanceMethod<&FileWrap::fsync>("fsync"),
                 InstanceMethod<&FileWrap::flock>("flock"),
                 InstanceMethod<&FileWrap::fcntllock>("fcntllock"),
+                InstanceMethod<&FileWrap::fcntlgetlock>("fcntlgetlock"),
```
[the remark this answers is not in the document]
guymguym: Good point, adding a jsdoc on the typescript function is the best place for that.

src/sdk/glacier.js
```diff
@@ -229,25 +227,18 @@ class Glacier {
...
             await native_fs_utils.lock_and_run(fs_context, lock_path(Glacier.GLACIER_CLUSTER_LOCK), async () => {
                 await process_glacier_logs(primary_log_ns, process_staged_fn);
             });
 
-            // Acquire a type specific lock to consume staged logs
-            await native_fs_utils.lock_and_run(
-                fs_context, lock_path(stage_lock_file), async () => {
-                    await process_glacier_logs(staged_log_ns, process_primary_fn);
-                }
-            );
```
guymguym, line 250: Isn't this cluster lock removal going to create more races?
Is this required? or can we split to a separate fix?

PR #9371, opened 2026-01-11
NC | object lock - implement object lock configuration function

nadavMiz:
### Explain the Changes
1. implement the following functions to set object locking configuration in nsfs:
  - `get_object_legal_hold`
  -  `put_object_legal_hold`
  - `get_object_retention` 
  - `put_object_retention`
  - `get_object_lock_configuration`
  - `put_object_lock_configuration`
2. move object lock schema to common_api and add it to nsfs_bucket schema
3. add `lock_status` data to head object (`get_object_info` function) so locking information will be available in GET and HEAD object functions
4. don't allow to set object lock configuration when versioning is not enabled
5. don't allow to set versioning when object lock is enabled
6. modify `test_s3_worm` to work in non containerized environment and add it to the nc tests list
7. fix tests and disable tests that test features that are not yet implemented 

### Gaps:
1. don't allow to suspend versioning if object lock configuration is enabled
2. we currently don't allow to set bucket object locking configuration without a default lock. aws does not make this restriction
3. more robust testings  

### Testing Instructions:
1. run `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/integration_tests/api/s3/test_s3_worm.js`

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Object Lock: bucket-level configuration, get/put configuration endpoints, legal-hold and retention operations; lock settings now appear in object metadata and listings.

* **Behavior Changes**
  * Stricter defaults for Object Lock enablement; preventing disabling of versioning when Object Lock is configured.

* **API**
  * Object Lock schema centralized and standardized across related endpoints; removed duplicated inline schema.

* **Tests**
  * Added/updated integration tests and test config to cover Object Lock and environment-specific differences.

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -1589,6 +1589,47 @@ module.exports = {
                 block_public_policy: { type: 'boolean' },
                 restrict_public_buckets: { type: 'boolean' },
             },
-        }
+        },
+        object_lock_configuration: {
+            type: 'object',
+            properties: {
+                object_lock_enabled: {
+                    type: 'string',
+                    enum: ['Enabled']
+                },
+                rule: {
+                    type: 'object',
+                    properties: {
+                        default_retention: {
+                            oneOf: [{
+                                    type: 'object',
+                                    required: ['years', 'mode'],
+                                    additionalProperties: false,
+                                    properties: {
+                                        years: { type: 'integer' },
+                                        mode: {
+                                            type: 'string',
+                                            enum: ['GOVERNANCE', 'COMPLIANCE']
+                                        }
+                                    }
+                                },
+                                {
+                                    type: 'object',
+                                    required: ['days', 'mode'],
+                                    additionalProperties: false,
+                                    properties: {
+                                        days: { type: 'integer' },
+                                        mode: {
+                                            type: 'string',
+                                            enum: ['GOVERNANCE', 'COMPLIANCE']
+                                        }
+                                    }
+                                }
+                            ]
+                        }
+                    }
+                }
+            }
+        },
     }
 };
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -356,8 +356,8 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             owner_account: account.owner ? account.owner : account._id, // The account is the owner of the buckets that were created by it or by its users.
             creator: account._id,
             versioning: config.NSFS_VERSIONING_ENABLED && lock_enabled ? 'ENABLED' : 'DISABLED',
-            object_lock_configuration: config.WORM_ENABLED ? {
-                object_lock_enabled: lock_enabled ? 'Enabled' : 'Disabled',
+            object_lock_configuration: (config.WORM_ENABLED && config.NSFS_VERSIONING_ENABLED && lock_enabled) ? {
+                object_lock_enabled: 'Enabled',
             } : undefined,
             creation_date: new Date().toISOString(),
             force_md5_etag: force_md5_etag,
@@ -483,10 +483,17 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
     async set_bucket_versioning(params, object_sdk) {
         try {
             const ns = await object_sdk._get_bucket_namespace(params.name);
-            await ns.set_bucket_versioning(params.versioning, object_sdk);
             const { name, versioning } = params;
             dbg.log0('BucketSpaceFS.set_bucket_versioning: Bucket name, versioning', name, versioning);
             const bucket = await this.config_fs.get_bucket_by_name(name);
+            if (bucket.object_lock_configuration) {
+                if (versioning !== 'ENABLED') {
+                    const err = new S3Error(S3Error.InvalidBucketState);
+                    err.message = 'An Object Lock configuration is present on this bucket, so the versioning state cannot be changed.';
+                    throw err;
+                }
+            }
+            await ns.set_bucket_versioning(params.versioning, object_sdk);
             bucket.versioning = versioning;
             await this.config_fs.update_bucket_config_file(bucket);
         } catch (err) {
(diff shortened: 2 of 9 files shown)


src/sdk/namespace_fs.js
```diff
@@ -66,6 +66,8 @@ const XATTR_DELETE_MARKER = XATTR_NOOBAA_INTERNAL_PREFIX + 'delete_marker';
 const XATTR_DIR_CONTENT = XATTR_NOOBAA_INTERNAL_PREFIX + 'dir_content';
 const XATTR_NON_CURRENT_TIMESTASMP = XATTR_NOOBAA_INTERNAL_PREFIX + 'non_current_timestamp';
 const XATTR_TAG = XATTR_NOOBAA_INTERNAL_PREFIX + 'tag.';
+const XATTR_LEGAL_HOLD = XATTR_NOOBAA_INTERNAL_PREFIX + 'legal_hold';
+const XATTR_RETENTION_MODE = XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_mode';
```
jackyalbo, line 71: @nadavMiz As we don't use JSON.parse for any of our xattrs, I think maybe we should split retention to mode and date, and then we can avoid all the parse handling and everything, also it will be more easily read from the file system. WDYT?

```suggestion
const XATTR_RETENTION_MODE = XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_mode';
const XATTR_RETENTION_DATE= XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_date';
```

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -38,7 +42,7 @@ mocha.describe('s3 worm', function() {
     let version_id2;
     let version_id3;
     let version_id4;
-    const user_a = 'alice';
+    const user_a = 'alicia';
```
jackyalbo, line 46: why? :)

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -282,29 +307,46 @@ mocha.describe('s3 worm', function() {
...
             }), 'InvalidRequest', 'SOAP requests must be made over an HTTPS connection.');
+            }
         });
         mocha.it('should put legal hold on OBJ3', async function() {
             const conf = await s3_owner.putObjectLegalHold({ Bucket: BKT, Key: OBJ3, LegalHold: { Status: 'OFF' } });
+            delete conf.$metadata; // sdk returns metadata we dont care about here
             assert.deepEqual(conf, {});
         });
         mocha.it('should get object', async function() {
+            //TODO: nc - aplly object lock from bucket configuration not implemented yet
```
jackyalbo:
```suggestion
            //TODO: nc - apply object lock from bucket configuration not implemented yet
```
I think you have some of those in the file

nadavMiz:
@coderabbitai help

src/test/integration_tests/api/s3/test_s3_worm.js
nadavMiz: because alice already exists in a different test. we get an error for user already exists when trying to create this user

src/sdk/namespace_fs.js
nadavMiz: I don't think that it will improve it much. we will still need to handle parsing in the sense that both fields need to exist and be valid. and on readability level it will mean that we will have to fields we will need to look at rather than one. I like the idea of JSON better because it fits the s3 format in get and set. but I don't have strong feeling regarding either way. in any case, we validate the fields before assigning the xattr, so the shouldn't be any problems unless the someone sets the xattr manually

src/sdk/bucketspace_fs.js
```diff
@@ -920,11 +920,30 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
...
+                throw new RpcError('INVALID_BUCKET_STATE');
+            }
+            dbg.log0('BucketSpaceFS.put_object_lock_configuration: Bucket name, configuraion', name, object_lock_configuration);
+            const bucket = await this.config_fs.get_bucket_by_name(name);
+            bucket.object_lock_configuration = object_lock_configuration;
+            await this.config_fs.update_bucket_config_file(bucket);
+        } catch (error) {
+            throw translate_error_codes(error, entity_enum.BUCKET);
+        }
     }
```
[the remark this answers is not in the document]
nadavMiz: schema is already validated inside update_bucket_config_file. so validation here is not required

src/sdk/namespace_fs.js
nadavMiz: done

src/api/common_api.js
```diff
@@ -1589,6 +1589,40 @@ module.exports = {
...
+                                            enum: ['GOVERNANCE', 'COMPLIANCE']
+                                        }
+                                    }
+                                }
+                            ]
+                        }
+                    }
+                }
+            }
+        },
```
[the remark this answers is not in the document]
nadavMiz: this case is already validated in parse_body_object_lock_conf_xml. but added this check just in case

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -282,30 +310,47 @@ mocha.describe('s3 worm', function() {
...
+                    VersionId: version_id3
+                }), 'NoSuchObjectLockConfiguration', 'The specified object does not have a ObjectLock configuration');
+            } else {
+                //TODO: error should be NoSuchObjectLockConfiguration
+                await assert_throws_async(s3_owner.getObjectLegalHold({
                 Bucket: BKT,
                 Key: OBJ3,
                 VersionId: version_id3
             }), 'InvalidRequest', 'SOAP requests must be made over an HTTPS connection.');
+            }
```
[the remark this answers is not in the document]
nadavMiz: @coderabbitai there already is a TODO

jackyalbo (approved):

src/sdk/namespace_fs.js
```diff
@@ -2200,17 +2203,110 @@ class NamespaceFS {
...
+            if (now < retain_until_date) {
+                //TODO should check for bypass_governance permission
+                if (retention.mode === 'COMPLIANCE' ||
+                    (retention.mode === 'GOVERNANCE' && !bypass_governance)) {
+                    const err = new S3Error(S3Error.AccessDenied);
+                    err.message = 'Access Denied because object protected by object lock.';
+                    throw err;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
nadavMiz: known issue there is a TODO

PR #9380, opened 2026-01-16
[GLACIER | NC] Fix fcntlgetlock INVAL issue

tangledbytes:
### Describe the Problem
GPFS doesn't support `flock.l_type` value to be `F_UNLCK` which caused fcntlgetlock to throw exception when run on GPFS.

### Explain the Changes
This PR replaces `F_UNLCK` with `F_WRLCK` which has the same result as `F_UNLCK`. This PR also adds `nb_native().fs.fcntlgetlock` which is safer than the previous `fcntlgetlock` as the previous one was easy to incorrectly use by acquiring the lock from an file handler and then accidentally reusing the same handler to check for lock which would always result in `F_UNLCK`. The new API eliminates that possibility by managing the lifecycle of the FD internally.

1. 

### Testing Instructions:
1. `./node_modules/.bin/mocha src/test/unit_tests/nsfs/test_nsfs_glacier_backend.js`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Public lock-query API added to check file lock state (undefined/unlocked, SHARED, EXCLUSIVE) from file objects and top-level FS.

* **Bug Fixes**
  * Standardized unlocked reporting and improved lock-detection semantics.
  * Background logger now queries lock state directly, simplifying retries and removing explicit handle management.

* **Documentation**
  * Added LockType and JSDoc clarifying lock semantics and return values.

* **Tests**
  * Increased timeout for a restore-object test to improve reliability.

* **Chores**
  * New configurable logger lock-check interval option.

### DIFF
--- a/src/native/fs/fs_napi.cpp
+++ b/src/native/fs/fs_napi.cpp
@@ -1447,6 +1447,43 @@ struct GetPwName : public FSWorker
     }
 };
 
+struct FcntlGetLock: public FSWorker
+{
+    std::string _path;
+    struct flock fl;
+    FcntlGetLock(const Napi::CallbackInfo& info)
+        : FSWorker(info)
+        , fl()
+    {
+        fl.l_whence = SEEK_SET;
+        fl.l_start = 0;
+        fl.l_len = 0;
+        fl.l_pid = 0;
+        fl.l_type = F_WRLCK;
+
+        _path = info[1].As<Napi::String>();
+        Begin(XSTR() << "FcntlGetLock" << DVAL(_path));
+    }
+    virtual void Work()
+    {
+        int fd = open(_path.c_str(), 0);
+        CHECK_OPEN_FD(fd);
+        SYSCALL_OR_RETURN(fcntl(fd, F_OFD_GETLK, &fl));
+    }
+    virtual void OnOK() {
+        DBG1("FS::FileFcntlGetLock::OnOK: " << DVAL(fl.l_type));
+        Napi::Env env = Env();
+
+        if (fl.l_type == F_UNLCK) {
+            _deferred.Resolve(env.Undefined());
+        } else if (fl.l_type == F_RDLCK) {
+            _deferred.Resolve(Napi::String::New(env, "SHARED"));
+        } else {
+            _deferred.Resolve(Napi::String::New(env, "EXCLUSIVE"));
+        }
+    }
+};
+
 struct FileWrap : public Napi::ObjectWrap<FileWrap>
 {
     std::string _path;
@@ -2013,10 +2050,6 @@ struct FileFcntlLock : public FSWrapWorker<FileWrap>
     }
 };
 
-// NOTE: This function in most cases must NOT be called from the
-// same fd from which a lock was acquired earlier as the locks
-// are associated with the FD itself and the result of this
-// function in those cases would be meaningless.
 struct FileFcntlGetLock : public FSWrapWorker<FileWrap>
 {
     struct flock fl;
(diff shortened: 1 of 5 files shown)


src/native/fs/fs_napi.cpp
```diff
@@ -2613,6 +2640,7 @@ fs_napi(Napi::Env env, Napi::Object exports)
     exports_fs["getsinglexattr"] = Napi::Function::New(env, api<GetSingleXattr>);
     exports_fs["getpwname"] = Napi::Function::New(env, api<GetPwName>);
     exports_fs["symlink"] = Napi::Function::New(env, api<Symlink>);
+    exports_fs["fcntlgetlock"] = Napi::Function::New(env, api<FcntlGetLock>);
```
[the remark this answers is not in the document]
tangledbytes: The tests use PersistentLogger.process which now use this updated API. I don't think there are more tests needed.

src/native/fs/fs_napi.cpp
```diff
@@ -1447,6 +1447,43 @@ struct GetPwName : public FSWorker
...
+
+        _path = info[1].As<Napi::String>();
+        Begin(XSTR() << "FcntlGetLock" << DVAL(_path));
+    }
+    virtual void Work()
+    {
+        int fd = open(_path.c_str(), 0);
+        CHECK_OPEN_FD(fd);
+        SYSCALL_OR_RETURN(fcntl(fd, F_OFD_GETLK, &fl));
+    }
```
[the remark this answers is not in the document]
tangledbytes: Not really. Checking lock doesn't require write flag.

src/native/fs/fs_napi.cpp
```diff
@@ -2028,28 +2061,22 @@ struct FileFcntlGetLock : public FSWrapWorker<FileWrap>
         fl.l_start = 0;
         fl.l_len = 0;
         fl.l_pid = 0;
-        // F_UNLCK is a trick, by default GETLK doesn't report
-        // lock status rather checks for conflicts.
-        // However, if F_UNLCK is used then conflict detection
-        // acts similar lock status fetching.
-        fl.l_type = F_UNLCK;
+        fl.l_type = F_WRLCK;
```
alphaprinz, line 2064: According to this doc https://man7.org/linux/man-pages/man2/F_OFD_GETLK.2const.html,
When using F_OFD_GETLK, l_type is used as output field.
If so, why is set here?

src/native/fs/fs_napi.cpp
```diff
@@ -1447,6 +1447,43 @@ struct GetPwName : public FSWorker
...
+
+        _path = info[1].As<Napi::String>();
+        Begin(XSTR() << "FcntlGetLock" << DVAL(_path));
+    }
+    virtual void Work()
+    {
+        int fd = open(_path.c_str(), 0);
+        CHECK_OPEN_FD(fd);
+        SYSCALL_OR_RETURN(fcntl(fd, F_OFD_GETLK, &fl));
```
alphaprinz, line 1471: Why do we need the OFD variety?
Can we use F_GETLK?

alphaprinz (approved):
Had some minor questions, otherwise LGTM.

src/native/fs/fs_napi.cpp
tangledbytes: Actually it is both input and output. When specified in input it is used for conflict detection. So for example, if l_type is set as F_RDLCK then after the syscall returns, l_type could contain any of the 3 values depending upon whether there was any conflicting lock in place or not and if there was then which one it was.

src/native/fs/fs_napi.cpp
tangledbytes: I don't think if we acquire lock using F_OFD_SETLKW then it will be returned if we query using F_GETLK. And we use OFD one for the former to tie the lock to the fd instead of the process. F_SETLKW is process scoped.

PR #9389, opened 2026-01-23
Optimize bulk delete operation for non versioned buckets

karthikps97:
### Describe the Problem
For delete-objects S3 API, the internal implementation performs separate DB update queries (soft-delete in objectmds) for each object. This introduces additional round trip delay between the endpoint and PostgreSQL. It can be avoided by performing single update for all objects (Non versioned bucket only).

### Explain the Changes
1. Created a new function to perform bulk update for all the objects to be deleted.
2. Added checks in appropriate places in code to skip individual updates.

NOTE: This improvement is only for non versioned buckets. Buckets with versioning suspended are not part of this improvement.

1. https://issues.redhat.com/browse/DFBUGS-5137

### Testing Instructions:
1. Create a json file with all the objects to be deleted.
2. Use the S3 delete-objects API with the json file.
3. There is a limit of upto 1000 objects in a single request.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized bulk-delete path for buckets with versioning disabled to improve throughput and ensure consistent per-item results.

* **New Features**
  * Added unordered bulk update support to accelerate large-scale object deletions.

* **Infrastructure**
  * Introduced a centralized constants module to standardize common configuration values.

### DIFF
--- a/src/common/constants.js
+++ b/src/common/constants.js
@@ -1,0 +1,14 @@
+/* Copyright (C) 2026 NooBaa */
+'use strict';
+
+const COMMON_CONSTANTS = {
+  S3: {
+    VERSIONING: {
+      ENABLED: 'ENABLED',
+      SUSPENDED: 'SUSPENDED',
+      DISABLED: "DISABLED"
+    }
+  }
+};
+
+module.exports = COMMON_CONSTANTS;
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -34,6 +34,7 @@ const { EndpointStatsStore } = require('../analytic_services/endpoint_stats_stor
 const { IoStatsStore } = require('../analytic_services/io_stats_store');
 const { ChunkAPI } = require('../../sdk/map_api_types');
 const config = require('../../../config');
+const CONSTANTS = require('../../common/constants');
 const Quota = require('../system_services/objects/quota');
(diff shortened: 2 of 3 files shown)


src/common/constants.js
```diff
@@ -0,0 +1,14 @@
+/* Copyright (C) 2026 NooBaa */
```
tangledbytes, line 1: I think s3_constants.js is a better place for this.

tangledbytes (approved):

src/common/constants.js
karthikps97: But the file itself is present in endpoint directory. So, thought of creating a common constants module.

src/server/object_services/object_server.js
```diff
@@ -2130,6 +2163,60 @@ function _sort_parts_by_seq(a, b) {
...
+                deleted: now
+            }
+        };
+
+        bulk.find(filter).updateOne(payload);
+
+        dbg.log1(`${obj_md.key} was deleted by ${req.account && req.account.email.unwrap()}`);
+        const reply = _get_delete_obj_reply(obj_md);
+        reply.seq = await MDStore.instance().alloc_object_version_seq();
+        results[idx] = reply;
```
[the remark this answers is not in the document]
karthikps97: If there's an error entire operation is marked as failed. As the delete request is idempotent, the client retries the request.

PR #9420, opened 2026-02-09
NC | implement object lock functionality

nadavMiz:
### Describe the Problem
implement missing object lock functionality

### Explain the Changes
1. implement put object lock when creating new object using the following logic:
    a. if legal_hold or retention flags were specified in the header use the provided values
    b. if the flags were not defined used the default object lock policy in the defined for the bucket
    c. if there are no headers and no default policy don't put any lock
2. check for object lock during delete object version. in case there is a lock and no bypass, throw AcessDenied
3. fix xml parsing for object lock configuration to allow to enable object lock without setting a default policy. also to throw MalformedXml error in case of a malformed policy
4. change put_object_retention to allow to extend the retention time
5. allow to set object lock during create multipart upload
6. enable all tests in `test_s3_worm.js` for nc and add new tests

1. continuation of #9371 

### Testing Instructions:
1. run `sudo NC_CORETEST=true node ./node_modules/mocha/bin/mocha src/test/integration_tests/api/s3/test_s3_worm.js`

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Object Lock expanded: bucket default lock config exposed; lock settings flow through uploads (including multipart) and are stored with object metadata; enforcement during reads/deletes/versions.

* **Bug Fixes**
  * Validation tightened: retention durations enforce minimums (1 day / 1 year).
  * Improved parsing and error mapping for malformed lock configuration inputs; safer error handling.

* **Tests**
  * Broad new and updated tests covering retention, legal hold, versioning, multipart uploads, and default-bucket behavior.

* **Chores**
  * Minor test comment/reference updates and mock enhancements to support bucket-info reads.

### DIFF
--- a/src/endpoint/s3/ops/s3_put_bucket_object_lock.js
+++ b/src/endpoint/s3/ops/s3_put_bucket_object_lock.js
@@ -17,10 +17,19 @@ async function put_bucket_object_lock(req, res) {
     }
     const lock_configuration = s3_utils.parse_body_object_lock_conf_xml(req);
 
-    await req.object_sdk.put_object_lock_configuration({
-        name: req.params.bucket,
-        object_lock_configuration: lock_configuration,
-    });
+    try {
+        await req.object_sdk.put_object_lock_configuration({
+            name: req.params.bucket,
+            object_lock_configuration: lock_configuration,
+        });
+    } catch (err) {
+        let error = err;
+        if (err.rpc_code === 'INVALID_SCHEMA' || err.rpc_code === 'INVALID_SCHEMA_PARAMS') {
+            console.error('put_bucket_object_lock: Invalid schema provided', err);
+            error = new S3Error(S3Error.MalformedXML);
+        }
+        throw error;
+    }
 }
 
 module.exports = {
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -538,35 +538,50 @@ function parse_body_encryption_xml(req) {
 
 function parse_body_object_lock_conf_xml(req) {
     const configuration = req.body.ObjectLockConfiguration;
-    const retention = configuration.Rule[0].DefaultRetention[0];
+    try {
+        const retention = configuration?.Rule?.[0]?.DefaultRetention?.[0];
+        const object_lock_enabled = configuration?.ObjectLockEnabled?.[0];
+        if (object_lock_enabled !== 'Enabled') {
+            dbg.error('invalid object_lock_enabled, can only be Enabled', object_lock_enabled);
+            throw new S3Error(S3Error.MalformedXML);
+        }
+        const conf = {
+            object_lock_enabled,
+        };
 
-    if ((retention.Days && retention.Years) || (!retention.Days && !retention.Years) ||
-        (retention.Mode[0] !== 'GOVERNANCE' && retention.Mode[0] !== 'COMPLIANCE')) throw new S3Error(S3Error.MalformedXML);
-
-    const conf = {
-        object_lock_enabled: configuration.ObjectLockEnabled[0],
-        rule: { default_retention: { mode: retention.Mode[0], } }
-    };
-
-    if (retention.Days) {
-        const days = parseInt(retention.Days[0], 10);
-        if (days <= 0) {
-            const err = new S3Error(S3Error.InvalidArgument);
-            err.message = 'Default retention period must be a positive integer value';
+        if (retention) {
+            conf.rule = { default_retention: { mode: retention.Mode[0] } };
+            if ((!retention.Days && !retention.Years) || (retention.Mode[0] !== 'GOVERNANCE' && retention.Mode[0] !== 'COMPLIANCE')) {
+                dbg.error('invalid object lock confugration retention', retention);
+                throw new S3Error(S3Error.MalformedXML);
+            }
+            if (retention.Days) {
+                const days = parseInt(retention.Days[0], 10);
+                if (!Number.isInteger(days) || days <= 0) {
+                    const err = new S3Error(S3Error.InvalidArgument);
+                    err.message = 'Default retention period must be a positive integer value';
+                    throw err;
+                }
+                conf.rule.default_retention.days = days;
+            }
+            if (retention.Years) {
+                const years = parseInt(retention.Years[0], 10);
+                if (!Number.isInteger(years) || years <= 0) {
+                    const err = new S3Error(S3Error.InvalidArgument);
+                    err.message = 'Default retention period must be a positive integer value';
+                    throw err;
+                }
+                conf.rule.default_retention.years = years;
+            }
+        }
+        return conf;
+    } catch (err) {
+        dbg.error('parse_body_object_lock_conf_xml failed', err);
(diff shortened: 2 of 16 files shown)


src/sdk/namespace_fs.js
```diff
@@ -2219,10 +2281,55 @@ class NamespaceFS {
...
+            if (new_date > retain_until_date && new_retention.mode === current_retention.mode) return;
+            //TODO should check for bypass_governance permission
+            if (current_retention.mode === 'COMPLIANCE' ||
+                (current_retention.mode === 'GOVERNANCE' && !bypass_governance)) {
+                const err = new S3Error(S3Error.AccessDenied);
+                err.message = 'Access Denied because object protected by object lock.';
+                throw err;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
nadavMiz: @coderabbitai can you give a reference to that. as far as I know you can't change the retention mode. wouldn't that also be dangerous, that would mean someone can make the lock weaker by just increasing the time. also I tried it on aws in it seems I can't do it. but I might be missing something

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
```diff
@@ -17,10 +17,18 @@ async function put_bucket_object_lock(req, res) {
...
-        object_lock_configuration: lock_configuration,
-    });
+    try {
+        await req.object_sdk.put_object_lock_configuration({
+            name: req.params.bucket,
+            object_lock_configuration: lock_configuration,
+        });
+    } catch (err) {
+        let error = err;
+        if (err.rpc_code === 'INVALID_SCHEMA' || err.rpc_code === 'INVALID_SCHEMA_PARAMS') {
```
jackyalbo, line 27: We should pass the original error somewhere, I think maybe as the error message passed to the end-user, or at least write it to the logs

src/endpoint/s3/s3_utils.js
```diff
@@ -538,35 +538,41 @@ function parse_body_encryption_xml(req) {
...
-    const retention = configuration.Rule[0].DefaultRetention[0];
-
-    if ((retention.Days && retention.Years) || (!retention.Days && !retention.Years) ||
-        (retention.Mode[0] !== 'GOVERNANCE' && retention.Mode[0] !== 'COMPLIANCE')) throw new S3Error(S3Error.MalformedXML);
-
-    const conf = {
-        object_lock_enabled: configuration.ObjectLockEnabled[0],
-        rule: { default_retention: { mode: retention.Mode[0], } }
-    };
+    try {
```
jackyalbo, line 541: Why do you need this? what suppose to be thrown in the code other than what you are throwing yourself?

src/endpoint/s3/s3_utils.js
```diff
@@ -538,35 +538,41 @@ function parse_body_encryption_xml(req) {
 
 function parse_body_object_lock_conf_xml(req) {
     const configuration = req.body.ObjectLockConfiguration;
-    const retention = configuration.Rule[0].DefaultRetention[0];
-
-    if ((retention.Days && retention.Years) || (!retention.Days && !retention.Years) ||
```
jackyalbo, line 543: Why are we removing it? handled by the schema?

src/sdk/namespace_fs.js
```diff
@@ -1368,6 +1368,9 @@ class NamespaceFS {
         if (part_upload) {
             fs_xattr = this._assign_part_props_to_fs_xattr(params.size, digest, offset, fs_xattr);
         }
+        if (!part_upload) {
```
jackyalbo: Can't it just be else?

src/sdk/namespace_fs.js
```diff
@@ -2203,7 +2210,62 @@ class NamespaceFS {
...
+    /**
+     * assigns object lock settings to fs xattr, if lock_settings is not provided, will try to get default object lock configuration of the bucket and assign it to fs xattr
+     * @param {Object} params
+     * @param {nb.ObjectSDK} object_sdk
+     * @param {Object} fs_xattr  
+     */
+    async _assign_object_lock_to_fs_xattr(params, object_sdk, fs_xattr) {
+        let lock_settings = params.lock_settings;
+        if (lock_settings) {
+            await this._check_bucket_lock_enabled(params.bucket, object_sdk);
```
jackyalbo: can't you just call `const { bucket } = await object_sdk.read_bucket_full_info(bucket_name);` and just pass bucket to the internal functions? also then they won't need to be async functions

src/sdk/namespace_fs.js
```diff
@@ -2203,7 +2210,62 @@ class NamespaceFS {
...
+        const retain_until_date = default_retention.days ? new Date(today.setDate(today.getDate() + default_retention.days)) :
+            new Date(today.setFullYear(today.getFullYear() + default_retention.years));
+
+        return { retention: { mode: default_retention.mode, retain_until_date } };
+    }
+
+    /**
+     * check if the object deletion should be blocked by retention lock. if the object is blocked will throw AccessDenied error
+     * @param {Object} retention - object retention lock settings
+     * @param {boolean} bypass_governance - if true, and user has permission to use this flag, will allow to bypass governance mode retention lock. compliance mode retention lock cannot be bypassed.
```
jackyalbo, line 2323: add @throws

src/sdk/namespace_fs.js
```diff
@@ -2219,10 +2281,55 @@ class NamespaceFS {
...
     }
 
+    /**
+     * check if the object retention lock settings can be updated to the new retention settings. if not, will throw AccessDenied error
+     * the rules are:
+     * 1. if the new retention is longer than the current retention, it can be updated (can increase retention time)
+     * 2. if the new retention is shorter than the current retention, it cannot be updated and will throw error, unless the user has bypass_governance permission and the current retention mode is GOVERNANCE
+     * @param {Object} current_retention - current object retention lock settings
+     * @param {Object} new_retention - new object retention lock settings
+     * @param {boolean} bypass_governance - if true, and user has permission to use this flag, will allow to bypass governance mode retention lock. compliance mode retention lock cannot be bypassed.
```
jackyalbo, line 2349: also here and everywhere else its missing add @throws

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
nadavMiz: I can add log for that. I think that the s3 error should match aws

src/endpoint/s3/s3_utils.js
nadavMiz: yes, its handled by the schema. but I'll return it anyway

src/endpoint/s3/s3_utils.js
nadavMiz: the xml might not match what we expect. for example Rule can be object instead of a list. in that case we will get internal error, but we want to return malformed xml. shouldn't happen because client probably align itself with aws. but just in case

src/sdk/namespace_fs.js
nadavMiz: right, it can

jackyalbo (approved):

PR #9423, opened 2026-02-12
IAM | Upgrade from 4.20 to 4.21 failed when configuring an single account in bucket policy

naveenpaul1:
### Describe the Problem
Upgrade from 4.20 to 4.21 failed when configuring an account in bucket policy prior to upgrade 

### Explain the Changes
1. Previously principals.AWS only upgraded type of array, this code change also consider string type principle.

1. https://issues.redhat.com/browse/DFBUGS-5592

### Testing Instructions:
1. On one tab I have docker run -it --rm --network host -v $(pwd):/noobaa -v noobaa_node_modules:/noobaa/node_modules/ -v noobaa_build:/noobaa/build/ -e container=docker -u root noobaa-base bash. Then on the very first install you must do npm ci; npm run build. This will save these assets in a docker volume, whenever you would run the docker command next time, you won't have to do anything, source code is mapped locally so local changes are reflected immediately without any slow image creation.

2. On another tab I have docker run -it --rm -v noobaapg:/var/lib/postgresql --network host -e POSTGRES_PASSWORD=noobaa -e POSTGRES_DB=nbcore postgres:16 .

3. On third tab I simply have docker exec -u postgres -it $(docker ps | awk 'NR>1' | fzf | awk '{print $1}') psql -d nbcore (where I select the pg container - this is fancier you can trim down the command). I keep this tab on to inspect the database post run- the db is called coretest.

4. run ./node_modules/.bin/mocha src/test/integration_tests/internal/test_upgrade_scripts.js in fouth tab.

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bucket policy upgrade to more reliably resolve IAM principals, preserve wildcard entries, and retain unresolved/invalid principals so access settings are not lost across upgrades.
  * Better handles both single and multiple principal formats to prevent misconfiguration during policy migration.

* **Tests**
  * Updated integration tests to validate expanded policy shapes, wildcard preservation, arrays vs single principals, and handling of unresolved principal values.

### DIFF
--- a/src/test/integration_tests/internal/test_upgrade_scripts.js
+++ b/src/test/integration_tests/internal/test_upgrade_scripts.js
@@ -222,6 +222,50 @@ mocha.describe('test upgrade scripts', async function() {
                     Action: ['s3:PutObject'],
                     Resource: [`arn:aws:s3:::*`]
                 },
+                {
+                    Effect: 'Allow',
+                    Principal: {
+                        "AWS": new SensitiveString(iam_username),
+                    },
+                    Action: ['s3:PutObject'],
+                    Resource: [`arn:aws:s3:::*`]
+                },
+                {
+                    Effect: 'Allow',
+                    Principal: {
+                        "AWS": [
+                            '*',
+                            new SensitiveString(iam_username),
+                            new SensitiveString(iam_username)
+                        ],
+                    },
+                    Action: ['s3:PutObject'],
+                    Resource: [`arn:aws:s3:::*`]
+                },
+                {
+                    Effect: 'Allow',
+                    Principal: {
+                        "AWS": '*',
+                    },
+                    Action: ['s3:PutObject'],
+                    Resource: [`arn:aws:s3:::*`]
+                },
+                {
+                    Effect: 'Allow',
+                    Principal: {
+                        "AWS": [new SensitiveString('invalid')],
+                    },
+                    Action: ['s3:PutObject'],
+                    Resource: [`arn:aws:s3:::*`]
+                },
+                {
+                    Effect: 'Allow',
+                    Principal: {
+                        "AWS": new SensitiveString('invalid_second'),
+                    },
+                    Action: ['s3:PutObject'],
+                    Resource: [`arn:aws:s3:::*`]
+                },
             ]
         };
         // clean all leftover bucket policies as upgrade script doesn't work on updated policies 
@@ -273,6 +317,13 @@ mocha.describe('test upgrade scripts', async function() {
 
         assert.strictEqual(new_policy.Statement[0].Principal.AWS[0], `arn:aws:iam::${account._id.toString()}:root`);
         assert.strictEqual(new_policy.Statement[1].Principal.AWS[0], `arn:aws:iam::${iam_account._id.toString()}:user/${iam_account.email.unwrap()}`);
+        assert.strictEqual(new_policy.Statement[2].Principal.AWS, `arn:aws:iam::${iam_account._id.toString()}:user/${iam_account.email.unwrap()}`);
+        assert.strictEqual(new_policy.Statement[3].Principal.AWS[0], '*');
+        assert.strictEqual(new_policy.Statement[3].Principal.AWS[1], `arn:aws:iam::${iam_account._id.toString()}:user/${iam_account.email.unwrap()}`);
+        assert.strictEqual(new_policy.Statement[3].Principal.AWS[2], `arn:aws:iam::${iam_account._id.toString()}:user/${iam_account.email.unwrap()}`);
+        assert.strictEqual(new_policy.Statement[4].Principal.AWS, '*');
+        assert.strictEqual(new_policy.Statement[5].Principal.AWS[0], 'invalid');
+        assert.strictEqual(new_policy.Statement[6].Principal.AWS, 'invalid_second');
     });
 
     mocha.after(async function() {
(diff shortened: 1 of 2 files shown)


src/upgrade/upgrade_scripts/5.21.0/upgrade_bucket_policy_principal.js
```diff
@@ -3,15 +3,12 @@
...
-    if (!principals.AWS) return;
-    const principal_arns = [];
-    for (const principal of principals.AWS) {
-        if (principal === '*') continue;
-        const account = system_store.data.accounts.find(acc => acc.email.unwrap() === principal.unwrap());
+
+function get_arn(dbg, principal, system_store) {
+    const account = system_store.data.accounts.find(acc => acc.email.unwrap() === principal.unwrap());
         if (!account) {
             dbg.log0(`Could not find the account with email: ${principal}`);
```
shirady: If you want to see the account in logs you would need to `principal.unwrap()`.

src/upgrade/upgrade_scripts/5.21.0/upgrade_bucket_policy_principal.js
shirady, line 18: I think you need the name and not the email.
https://github.com/noobaa/noobaa-core/blob/c33b18c2d06bd434ebd41bf93fb53e724e320433/src/server/system_services/account_server.js#L43

src/upgrade/upgrade_scripts/5.21.0/upgrade_bucket_policy_principal.js
```diff
@@ -20,9 +17,33 @@ function _create_arn(dbg, principals, system_store) {
...
+            principal_arns.push(arn);
+        }
+        return { AWS: principal_arns };
+    } else {
+        const principal = principals.AWS;
+        if (principal.unwrap() === '*') {
+            return { AWS: '*' };
+        }
+        const arn = get_arn(dbg, principal, system_store);
+        if (!arn) return { AWS: principal };
```
shirady: I think in case there was 1 account in the policy and it was deleted, then it might create issues since the bucket policy is not valid (you can check).
I will not block the PR as it fixes more critical issue.

src/test/integration_tests/internal/test_upgrade_scripts.js
```diff
@@ -222,6 +222,34 @@ mocha.describe('test upgrade scripts', async function() {
...
                     Resource: [`arn:aws:s3:::*`]
                 },
+                {
+                    Effect: 'Allow',
+                    Principal: {
+                        "AWS": new SensitiveString(iam_username),
+                    },
+                    Action: ['s3:PutObject'],
+                    Resource: [`arn:aws:s3:::*`]
+                },
```
shirady, line 232: If you can add an example of Statement of account (not IAM user) I think it can be useful.

shirady (approved):
Adding a few comment on the script
Approving not to block you

src/upgrade/upgrade_scripts/5.21.0/upgrade_bucket_policy_principal.js
naveenpaul1: didnt get you.

shirady: You can change this later.

I saw:
```js
arn = `arn:aws:iam::${account._id.toString()}:user${iam_path}${account.email.unwrap()}`;
```
username place with `email`, so I mentioned it should be `name` and referred to the code where it is set.

src/test/integration_tests/internal/test_upgrade_scripts.js
```diff
@@ -222,6 +222,50 @@ mocha.describe('test upgrade scripts', async function() {
...
+                {
+                    Effect: 'Allow',
+                    Principal: {
+                        "AWS": [
+                            '*',
```
liranmauda, line 237: Is this valid? 
if a value of `*` which is a string is valid, then will `if (principal.unwrap() === '*') {` work?

src/test/integration_tests/internal/test_upgrade_scripts.js
```diff
@@ -222,6 +222,50 @@ mocha.describe('test upgrade scripts', async function() {
...
+                            new SensitiveString(iam_username)
+                        ],
+                    },
+                    Action: ['s3:PutObject'],
+                    Resource: [`arn:aws:s3:::*`]
+                },
+                {
+                    Effect: 'Allow',
+                    Principal: {
+                        "AWS": '*',
```
liranmauda, line 248: Is this valid?
if a value of * which is a string is valid, then will if (principal.unwrap() === '*') { work?

src/test/integration_tests/internal/test_upgrade_scripts.js
naveenpaul1: It is valid, even if we save the principle as string, It is saved as `SensitiveString` in DB [bucket schema](https://github.com/noobaa/noobaa-core/blob/master/src/api/common_api.js#L321)
please find the log I added
```
console.log("_create_arn, principal: ", principal, principal.unwrap())
```
And Output
```
Feb-16 18:48:25.095 [Endpoint/79915]   [LOG] CONSOLE:: _create_arn, principal:  SENSITIVE-fdc569c3f3202c04 iam_username
Feb-16 18:48:25.095 [Endpoint/79915]   [LOG] CONSOLE:: _create_arn, principal:  SENSITIVE-7846cdd4c2b90527 *
Feb-16 18:48:25.095 [Endpoint/79915]   [LOG] CONSOLE:: _create_arn, principal: SENSITIVE-fdc569c3f3202c04 iam_username
Feb-16 18:48:25.095 [Endpoint/79915]   [LOG] CONSOLE:: _create_arn, principal:  SENSITIVE-fdc569c3f3202c04 iam_username
Feb-16 18:48:25.095 [Endpoint/79915]   [LOG] CONSOLE:: _create_arn, principal:  SENSITIVE-ec1f46354deeb3c6 invalid
```

PR #9427, opened 2026-02-12
Fixing few issues with sdkv3

jackyalbo:
### Describe the Problem
The main issue here is we can't just set s3DisableBodySigning to false in sdkv3. 
Setting his replacement applyChecksum - will stop calculating the checksum and will leave x-amz-content-sha256
to be empty and not UNSIGNED-PAYLOAD like it was in sdkv2.
So if we do want to set applyChecksum to false - we will need to patch this headr ourselves like this:
```
params.applyChecksum = cloud_utils.disable_s3_compatible_bodysigning(endpoint);
const client = noobaa_s3_client.get_s3_client_v3_params(params);
client.middlewareStack.add(
  (next) => (args) => {
    // Only add if missing to avoid double-hashing
    if (!args.request.headers["x-amz-content-sha256"]) {
      args.request.headers["x-amz-content-sha256"] = "UNSIGNED-PAYLOAD";
    }
    return next(args);
  },
  { step: "build" }
);
```
So, we won't use it as the default for now

### Explain the Changes
1. Fixed the main issue - removing applyChecksum
2. returning the signatureVersion part - for cases we will revert to sdkv2
3. fixing putObject in blockstore_s3 to use empty string instead of undefined: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3/command/PutObjectCommand/
4. Adding an auto translate from applyChecksum to s3DisableBodySigning for sdkv2, which was missing. 

1. Fixed https://issues.redhat.com/browse/DFBUGS-4461
2. Opened an issue with AWS SDK about this issue: https://github.com/aws/aws-sdk-js-v3/issues/7740

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved S3 signature version handling for better compatibility with various storage endpoints.
  * Corrected checksum/body-signing mapping for non-AWS endpoints and ensured consistent behavior when metadata is disabled.

* **Tests**
  * Added unit test verifying body-signing/applyChecksum mapping between client parameter formats.

* **Chores**
  * Clarifying comments added around S3 client options to aid future SDK migration.

### DIFF
--- a/src/agent/block_store_services/block_store_s3.js
+++ b/src/agent/block_store_services/block_store_s3.js
@@ -45,6 +45,7 @@ class BlockStoreS3 extends BlockStoreBase {
                         secretAccessKey: this.cloud_info.access_keys.secret_key.unwrap(),
                     },
                     forcePathStyle: true,
+                    signatureVersion: cloud_utils.get_s3_endpoint_signature_ver(endpoint, this.cloud_info.auth_method),
                     region: config.DEFAULT_REGION,
                     requestHandler: new NodeHttpHandler({
                         httpsAgent: http_utils.get_default_agent(endpoint)
@@ -59,12 +60,15 @@ class BlockStoreS3 extends BlockStoreBase {
             this.s3cloud = noobaa_s3_client.get_s3_client_v3_params({
                 endpoint: endpoint,
                 forcePathStyle: true,
+                signatureVersion: cloud_utils.get_s3_endpoint_signature_ver(endpoint, this.cloud_info.auth_method),
                 credentials: {
                     accessKeyId: this.cloud_info.access_keys.access_key.unwrap(),
                     secretAccessKey: this.cloud_info.access_keys.secret_key.unwrap(),
                 },
                 region: config.DEFAULT_REGION,
-                applyChecksum: cloud_utils.disable_s3_compatible_bodysigning(endpoint),
+                // We will disable setting applyChecksum as sdkv3 doesn't always "know" to fall back to UNSIGNED-PAYLOAD automatically 
+                // in x-amz-content-sha256 for every command - and this header is a must for v4 signature.
+                // applyChecksum: cloud_utils.disable_s3_compatible_bodysigning(endpoint),
                 requestHandler: new NodeHttpHandler({
                     httpsAgent: http_utils.get_unsecured_agent(endpoint)
                 }),
@@ -145,6 +149,7 @@ class BlockStoreS3 extends BlockStoreBase {
             accessKeyId: this.cloud_info.access_keys.access_key,
             secretAccessKey: this.cloud_info.access_keys.secret_key.unwrap(),
             signatureVersion: cloud_utils.get_s3_endpoint_signature_ver(endpoint, this.cloud_info.auth_method),
+            // IMPORTANT - we had issues with applyChecksum - when migrating to sdkv3 check if this option is needed.
             s3DisableBodySigning: cloud_utils.disable_s3_compatible_bodysigning(endpoint),
         };
         if (this.cloud_info.aws_sts_arn) {
@@ -240,7 +245,7 @@ class BlockStoreS3 extends BlockStoreBase {
         const res = await this.s3cloud.putObject({
             Bucket: this.cloud_info.target_bucket,
             Key: this.usage_path,
-            Body: this.disable_metadata ? usage_data : undefined,
+            Body: this.disable_metadata ? usage_data : '',
             Metadata: this.disable_metadata ? undefined : {
                 [this.usage_md_key]: usage_data
             },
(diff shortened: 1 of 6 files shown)


naveenpaul1 (approved):
A few unit tests are failing, otherwise LGTM

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -53,6 +53,10 @@ function change_s3_client_params_to_v2_structure(params) {
     // v2: s3ForcePathStyle, v3: forcePathStyle
     replace_field(params, 'forcePathStyle', 's3ForcePathStyle');
 
+    // v2: s3DisableBodySigning, v3: applyChecksum
+    replace_field(params, 'applyChecksum', 's3DisableBodySigning');
+    params.s3DisableBodySigning = cloud_utils.disable_s3_compatible_bodysigning(params.endpoint);
+
```
shirady, line 58: Could you add a test case in file `test_noobaa_s3_client.test.js` with this case?
https://github.com/noobaa/noobaa-core/blob/b4194add80f8ff4e898c88302111164501cfae07/src/test/unit_tests/internal/test_noobaa_s3_client.test.js#L59

src/agent/block_store_services/block_store_s3.js
```diff
@@ -59,12 +60,15 @@ class BlockStoreS3 extends BlockStoreBase {
...
+                signatureVersion: cloud_utils.get_s3_endpoint_signature_ver(endpoint, this.cloud_info.auth_method),
                 credentials: {
                     accessKeyId: this.cloud_info.access_keys.access_key.unwrap(),
                     secretAccessKey: this.cloud_info.access_keys.secret_key.unwrap(),
                 },
                 region: config.DEFAULT_REGION,
-                applyChecksum: cloud_utils.disable_s3_compatible_bodysigning(endpoint),
+                // We will disable setting applyChecksum as sdkv3 doesn't always "know" to fall back to UNSIGNED-PAYLOAD automatically 
+                // in x-amz-content-sha256 for every command - and this header is a must for v4 signature.
+                // applyChecksum: cloud_utils.disable_s3_compatible_bodysigning(endpoint),
```
shirady, line 71: I think we would like the removal of the field be part of `get_s3_client_v3_params` so we can avoid future bugs (in case someone else migrates existing code with `applyChecksum` field and use this function, WDYT?

src/agent/block_store_services/block_store_s3.js
```diff
@@ -240,7 +244,7 @@ class BlockStoreS3 extends BlockStoreBase {
         const res = await this.s3cloud.putObject({
             Bucket: this.cloud_info.target_bucket,
             Key: this.usage_path,
-            Body: this.disable_metadata ? usage_data : undefined,
+            Body: this.disable_metadata ? usage_data : '',
```
shirady, line 248: In the PR description, we didn't explain why we did this - 

> fixing putObject in blockstore_s3 to use empty string instead of undefined

src/agent/block_store_services/block_store_s3.js
shirady, line 71: If we had a client with `signatureVersion` with value of `v2`:
So in the flow, we would choose `AWS SDK V2` for it,
but when we replace the fields back - in this code, you removed the field `applyChecksum`, so in `change_s3_client_params_to_v2_structure`, this client would not have `s3DisableBodySigning` as it originally had.

jackyalbo: I prefer not to, as it's not always relevant - the issue I got was with empty body for get and head and always

jackyalbo: yes, I checked with v2 and it works also without s3DisableBodySigning - I guess we are losing some sort of performance benefit, but as we hope that we will eventually stop having v2 clients - I don't think it's a big issue.

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -53,6 +53,9 @@ function change_s3_client_params_to_v2_structure(params) {
     // v2: s3ForcePathStyle, v3: forcePathStyle
     replace_field(params, 'forcePathStyle', 's3ForcePathStyle');
 
+    // v2: s3DisableBodySigning, v3: applyChecksum
+    replace_field(params, 'applyChecksum', 's3DisableBodySigning');
```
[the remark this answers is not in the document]
jackyalbo: from what we see the logic is the same and it's only a rename of this param. see here: https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/migrating/notable-changes/

src/agent/block_store_services/block_store_s3.js
jackyalbo: Added the refernce

shirady (approved):
LGTM

PR #8958, opened 2025-04-09
Added `x-amz-expiration` missing HTTP header in the response of object operations

aayushchouhan09:
### Describe the Problem
S3-compatible clients expect the `x-amz-expiration` response header when an object is subject to a lifecycle expiration rule. However, NooBaa was not setting this header in GET, PUT, or HEAD responses, even when valid expiration rules were configured.

### Explain the Changes
1. Added `set_expiration_header()` function to include the `x-amz-expiration` header in GET, PUT, and HEAD object responses.
2. Implemented `parse_expiration_header()` to convert lifecycle expiration rules (days or date) into proper s3 style header format.

Below is the output for header of putObject, headObject and getObject (after the fix):

Screenshot from 2025-04-30 18-30-45

### Issues: Fixed #xxx / Gap #xxx
1.  Partially Fixed: #8554 
2.  Ceph-s3 tests are failing because in ceph tests they are calculating number of days without converting `start_time` to midnigt UTC (inside check_expiration_header() function) which is causing error and returning wrong no. of expiration days.
[ceph-s3 tests issue](https://github.com/ceph/s3-tests/issues/652) has been created for the same.

### Testing Instructions:
1. Ceph-s3 tests are failing.
3. Added unit tests.
- [X] Tests added

PR #9437, opened 2026-02-18
Added lifecycle-configuration-rules in bucket SDK cache for x-amz-expiration header

aayushchouhan09:
### Describe the Problem
`set_expiration_header()` was loading bucket lifecycle rules on every GET/HEAD/PUT, which is expensive on NC/NSFS (it re-reads bucket config), for small objects this is noticeable and should be avoided.

### Explain the Changes
1. `set_expiration_header()` now fetches lifecycle rules via object_sdk.read_bucket_sdk_lifecycle_rules(), which uses the existing bucket SDK cache and returns early when no rules exist.

1.  Fixed: #9418
2.  Related PR: #8958 

### Testing Instructions:
1.  Tests not required, its a small fix and the tests were already added in the above mentioned related PR.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Bucket info now exposes lifecycle configuration rules so clients can view lifecycle settings.
  * Added API to retrieve bucket lifecycle rules for use when evaluating expirations.

* **Bug Fixes**
  * Tightened validation for expiration header processing to avoid setting headers without matching rules.
  * Improved handling when lifecycle rules are missing, empty, or invalid to ensure consistent responses.

### DIFF
--- a/src/sdk/object_sdk.js
+++ b/src/sdk/object_sdk.js
@@ -244,6 +244,11 @@ class ObjectSDK {
         }
     }
 
+    async read_bucket_sdk_lifecycle_rules(name) {
+        const { bucket } = await bucket_namespace_cache.get_with_cache({ sdk: this, name });
+        return bucket?.lifecycle_configuration_rules;
+    }
+
     async load_requesting_account(req) {
         try {
             const token = this.get_auth_token();
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -745,7 +745,11 @@ function set_amz_headers(req, res) {
 async function set_expiration_header(req, res, object_info) {
     if (!config.S3_LIFECYCLE_EXPIRATION_HEADER_ENABLED) return;
 
-    const rules = req.params.bucket && await req.object_sdk.get_bucket_lifecycle_configuration_rules({ name: req.params.bucket });
+    const bucket_name = req.params?.bucket;
+    if (!bucket_name) return;
+
+    const rules = await req.object_sdk.read_bucket_sdk_lifecycle_rules(bucket_name);
+    if (!Array.isArray(rules) || !rules.length) return;
 
     const matched_rule = lifecycle_utils.get_lifecycle_rule_for_object(rules, object_info);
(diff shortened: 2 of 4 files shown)


src/sdk/object_sdk.js
```diff
@@ -244,6 +244,13 @@ class ObjectSDK {
         }
     }
 
+    async read_bucket_sdk_lifecycle_rules(name) {
+        const { bucket } = await bucket_namespace_cache.get_with_cache({ sdk: this, name });
+        if (bucket?.lifecycle_configuration_rules) return bucket.lifecycle_configuration_rules;
+        // TODO: remove this fallback once containerized NooBaa (non-NSFS) bucket SDK info includes lifecycle rules
```
romayalon: Hey :) 2 suggestions - 
1. add lifecycle configuration rules to bucket_sdk_info on this PR (for containerized buckets as well) - should be a small change.
2. on NC/NSFS bucket you will still trigger another call to read the bucket configuration rules even if we already know it's empty so I suggest to avoid it

aayushchouhan09: Make sense, I have updated the code as per the comments. Thanks

src/util/http_utils.js
```diff
@@ -689,7 +689,11 @@ function set_amz_headers(req, res) {
  * @param {Object} object_info
  */
 async function set_expiration_header(req, res, object_info) {
-    const rules = req.params.bucket && await req.object_sdk.get_bucket_lifecycle_configuration_rules({ name: req.params.bucket });
+    const bucket_name = req.params?.bucket;
+    if (!bucket_name) return;
+
+    const rules = await req.object_sdk.read_bucket_sdk_lifecycle_rules(bucket_name);
+    if (!rules.length) return;
```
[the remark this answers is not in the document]
aayushchouhan09: rules cannot be null or undefined, it can either be some rules or []

romayalon (approved):
LGTM
Do we need here a new test? or is it already tested in an existing test

PR #9458, opened 2026-03-03
Update nodejs from v24.4.1 to 24.13.0

romayalon:
### Describe the Problem
[In order to make node.js support PQC(Post-Quantum Cryptography) we need to upgrade to node.js 24.7.0](https://github.com/nodejs/node/releases/tag/v24.7.0), was asked by Liran to jump directly to 24.13.0. 

### Explain the Changes
1. Update nodejs from v24.4.1 to 24.13.0

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Chores**
  * Updated the required Node.js version to 24.7.0 across the project.
  * CI configuration now sources the Node.js version from the repository configuration file instead of using a fixed, hard-coded value, ensuring consistency between local and CI environments.


.nvmrc
```diff
@@ -1 +1 @@
-24.4.1
+24.7.0
```
liranmauda: This is a good comment, as the `Validate-package-lock.yaml` should be aligned to the node version in the `.nvmrc` 

The test itself should not fail.
We only test the top-level dependencies (`Validate dependencies top-level versions in package-lock.json`)  to avoid a mismatch with the tree dependencies, as they can be changed when they are not pinned to a certain version. 

This test comes to verify that the `npm install` was running before the PR, and avoid merging a `package.json` change before running `npm install` and committing the change.

liranmauda (approved):

PR #9464, opened 2026-03-04
Fixes for S3A Hadoop tests gh action

jackyalbo:
### Describe the Problem
GitHub Actions have very weak runners, so I had to add some more fixes to make the action run in an acceptable time

### Explain the Changes
1. perbuilt s3a-tester image - so we won't need to re-download the test dependencies for hours for every run. the image is now in quay.io -  quay.io/jalbo/s3a-tester:latest
2. Skipping this test ITestS3AContractMultipartUploader - as it hangs due to very low resources for the run
3. Added some fixes of naming conventions 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Chores**
  * Centralized NC container naming and improved lifecycle handling for consistent start/stop/wait behavior and clearer logs.
  * Switched to a purpose-built S3A tester image and moved Hadoop dependency setup into that image to speed and stabilize test runs.
  * Updated test scripts and XML config, added new S3A test properties and an exclusion mechanism to streamline which tests execute.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,8 +165,10 @@ MINT_NOOBAA_HTTP_ENDPOINT_PORT=6001
 # HADOOP S3A VARIABLES #
 #######################
 
-HADOOP_S3A_IMAGE?="maven:3.9.6-eclipse-temurin-11"
+# HADOOP_S3A_IMAGE?="maven:3.9.6-eclipse-temurin-11"
+HADOOP_S3A_IMAGE?="quay.io/noobaa/s3a-tester:v1"
 HADOOP_S3A_ENDPOINT_PORT?=6001
+HADOOP_S3A_NC_CONTAINER:=noobaa-s3a-$(GIT_COMMIT)
 HADOOP_S3A_SCRIPT?="$(REPO_ROOT)/src/test/external_tests/hadoop_s3a_tests/run_hadoop_s3a_tests.sh"
 
 ###############
@@ -385,10 +387,10 @@ run-nc-tests: tester
 hadoop-s3a-nc-tests: tester
 	@$(call create_docker_network)
 	@echo "\033[1;34mRunning Hadoop S3A NC tests\033[0m"
-	@$(call run_nc_endpoint)
-	@$(call wait_nc_endpoint)
+	@$(call run_s3a_nc_endpoint)
+	@$(call wait_s3a_nc_endpoint)
 	@$(call run_hadoop_s3a_tests)
-	@$(call stop_nc_endpoint)
+	@$(call stop_s3a_nc_endpoint)
 	@$(call remove_docker_network)
 .PHONY: hadoop-s3a-nc-tests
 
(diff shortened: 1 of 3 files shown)


src/test/external_tests/hadoop_s3a_tests/run_hadoop_s3a_tests.sh
```diff
@@ -62,9 +63,34 @@ cat <<EOF > /tmp/hadoop/hadoop-tools/hadoop-aws/src/test/resources/auth-keys.xml
...
+    <value>true</value>
+  </property>
 </configuration>
 EOF
 
-cd /tmp/hadoop
-echo "Running: mvn -pl hadoop-tools/hadoop-aws -DskipTests=false -DskipITs=false -Dit.test=ITestS3A* -Dtest=TestS3A* verify"
-mvn -pl hadoop-tools/hadoop-aws -DskipTests=false -DskipITs=false -Dit.test=ITestS3A* -Dtest=TestS3A* verify
+echo "Running: mvn -pl :hadoop-aws -am -DskipTests=false -DskipITs=false -Dit.test='ITestS3A*,!ITestS3AContractMultipartUploader' -Dtest=TestS3A* verify"
+mvn -pl :hadoop-aws -am -DskipTests=false -DskipITs=false "-Dit.test=ITestS3A*,!ITestS3AContractMultipartUploader" -Dtest=TestS3A* verify
```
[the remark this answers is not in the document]
jackyalbo: good idea

Makefile
```diff
@@ -165,8 +165,10 @@ MINT_NOOBAA_HTTP_ENDPOINT_PORT=6001
 # HADOOP S3A VARIABLES #
 #######################
 
-HADOOP_S3A_IMAGE?="maven:3.9.6-eclipse-temurin-11"
+# HADOOP_S3A_IMAGE?="maven:3.9.6-eclipse-temurin-11"
+HADOOP_S3A_IMAGE?="quay.io/jalbo/s3a-tester:latest"
```
liranmauda: We should probably have it in a noobaa quay register and not in a private one

liranmauda (approved):

Makefile
nadavMiz: since this image will be in quay. and its dockerbuild is in this repository. we need to think of how to maintain the target image so they will always match.
but since the s3a workflow runs on demand anyway and not part of the CI. why not keep it simple and just build everything as part of the run_hadoop_s3a_tests.sh script. and just have it run slow?

jackyalbo: it's taking too long. If I were to build the image, I could have just downloaded the tests during the build. I agree it's not amazing, but it's a hack for now.

nadavMiz: then we should at least add a comment in the dockerfile that the image is in quay. and new images should be uploaded there.

jackyalbo: Will do it as part of moving the image to the noobaa repository in a following PR. Thanks

jackyalbo: done

jackyalbo: Added a comment to the dockerfile

PR #9023, opened 2025-05-13
Added a validation for deletion of anonymous account with --name

aayushchouhan09:
### Describe the Problem
Currently, we are allowing user to delete the anonymous account with below command:
`noobaa-cli account delete --name anonymous`

### Explain the Changes
1. Added a validation to not delete the anonymous account with --name.

### Issues: Fixed #xxx / Gap #xxx
1. Fixed: #8856 

### Testing Instructions:
1.  `sudo npx jest noobaa-core/src/test/unit_tests/jest_tests/test_nc_account_cli.test.js`
- [X] Tests added

PR #9465, opened 2026-03-05
NC | fix anonymous user delete test

nadavMiz:
### Describe the Problem
There were a few issues with the test added in #9023:
1. `config_root` was not passed to the CLI command. This occasionally caused failures in CI when the default config root was modified by other test suites.
2. The test used a fixed name instead of `uid`/`gid`. While this may not always cause issues, it could fail in certain environments where these values differ (e.g., when modified by other test suites).
3. The test was placed in the wrong location. It should be in `test_nc_anonymous_cli.js` alongside the other anonymous user CLI tests.

### Explain the Changes
1. Added a new test in `test_nc_anonymous_cli.js` with the correct CLI parameters.
2. changed second `before_all` to be `after_all` as it should have been

1. 

### Testing Instructions:
1. `sudo npx jest src/test/integration_tests/nc/cli/test_nc_anonymous_cli.test.js`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Tests**
  * Refined anonymous-account test suite: removed a redundant case, added explicit failure and success scenarios for anonymous-account deletion, and retained nickname-based deletion checks.
  * Improved test isolation by moving account setup into individual tests and adding global cleanup to ensure reliable teardown and lifecycle separation.

### DIFF
--- a/src/test/integration_tests/nc/cli/test_nc_account_cli.test.js
+++ b/src/test/integration_tests/nc/cli/test_nc_account_cli.test.js
@@ -1783,24 +1783,6 @@ describe('manage nsfs cli account flow', () => {
             expect(JSON.parse(res.stdout).error.code).toBe(ManageCLIError.AccountDeleteForbiddenHasIAMAccounts.code);
         });
 
-        it('should fail when deleting anonymous account with option (name), and succeed with option (anonymous)', async function() {
-            const { type } = defaults;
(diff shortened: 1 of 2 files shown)


liranmauda (approved):
LGTM

naveenpaul1 (approved):

PR #9506, opened 2026-03-23
vectors - make lance dependency optional, it not supported on ppc64

alphaprinz:
### Describe the Problem
Lance dependency fails on ppc64, make it optional.

### Explain the Changes
1. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Chores**
  * Made a vector-related package optional in dependency configuration so the app can run without it.
* **Bug Fixes**
  * Vector REST endpoints now fail fast and return a clear "Not Available" response (HTTP 501) when the optional vector package is missing, preventing further request processing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -145,5 +145,8 @@
     "pkg-fetch": "3.5.2",
     "sinon": "21.0.3",
     "wtfnode": "0.10.1"
+  },
+  "optionalDependencies": {
+    "@lancedb/lancedb": "0.27.1"
   }
 }
--- a/src/endpoint/vector/vector_rest.js
+++ b/src/endpoint/vector/vector_rest.js
@@ -7,6 +7,7 @@ const VectorSDK = require('../../sdk/vector_sdk');
 const js_utils = require('../../util/js_utils');
 const http_utils = require('../../util/http_utils');
 const signature_utils = require('../../util/signature_utils');
+const lance = js_utils.require_optional('@lancedb/lancedb');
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -145,5 +145,8 @@
     "pkg-fetch": "3.5.2",
     "sinon": "21.0.3",
     "wtfnode": "0.10.1"
+  },
+  "optionalDependencies": {
+    "@lancedb/lancedb": "0.27.1"
```
liranmauda, line 150: This will not work at the downstream. 
@b-ranto to keep me honest

src/endpoint/vector/vector_rest.js
```diff
@@ -7,6 +7,12 @@ const VectorSDK = require('../../sdk/vector_sdk');
 const js_utils = require('../../util/js_utils');
 const http_utils = require('../../util/http_utils');
 const signature_utils = require('../../util/signature_utils');
+let lance;
+try {
+    lance = require('@lancedb/lancedb');
+} catch (err) {
+    console.log("Failed to load lance. Vector API is not avilable. err =", err);
+}
```
liranmauda: We can use `js_utils.require_optional`

package.json
b-ranto: Yes, we should avoid defining optional dependencies.

alphaprinz: @liranmauda @b-ranto @nimrod-becker
What is the alternative?

liranmauda (approved):

PR #9513, opened 2026-03-24
vectors - build - skip building lance for ppc64 without optionalDependencies

alphaprinz:
### Describe the Problem
Lance npm package doesn't support ppc64, and downstream build doesn't support optionalDependencies.
As a workaround, delete lance dependency for ppe64 from package.json.

### Explain the Changes
1. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Vector API now returns a proper HTTP 501 error when the service is unavailable, instead of failing silently.
  * Improved platform-specific build handling to ensure consistent installation across different system architectures.


liranmauda (approved):
We need to make sure that the doesnstream build also has this in its Dockerfile.
@b-ranto FYI

Other than that LGTM.

PR #9514, opened 2026-03-25
combine db querries for get command

nadavMiz:
### Describe the Problem

1. During the GET command on the PostgreSQL path, parts, chunks, and blocks were fetched in three separate DB queries, adding redundant round-trips for every object read.
2. For qualifying GET requests (non-folder, non-part-number), fetching object metadata and then streaming its blocks required multiple sequential DB queries and RPC calls, making small object reads unnecessarily slow.
3.The can_use_get_inline decision was computed inside namespace_s3.js rather than at the request entry point, so other namespaces couldn't benefit from it.

### Explain the Changes
1. Added find_parts_chunks_blocks_by_range() to md_store.js — a PostgreSQL-specific combined query that fetches parts, chunks, and aggregated blocks in a single round-trip using jsonb_agg.
2. Added find_object_with_mapping_by_key() to md_store.js — a PostgreSQL-specific query that fetches the object row together with its first INLINE_NUM_PARTS parts, chunks, and blocks in one round-trip. read_object_md in object_server.js uses this when can_use_get_inline is set, then reads the block data via MapClient and returns it as first_range_data in the response. s3_get_object can then serve small objects (or write the leading bytes of larger objects) without a separate streaming call
3. Moved the can_use_get_inline = true assignment from namespace_s3.js to s3_get_object.js so it is set once at the request entry point for all namespaces. namespace_s3 now reads it from params.can_use_get_inline instead of computing it locally.
4. reuse object_md from s3_get_object in read_object_mapping instead of calling read_object_md again. other paths that uses this function such as from namespace_context will still call get_object_md as before. also don't return object_md as it is never used

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Optional inline metadata prefetching returns initial bytes and inline chunk info with object metadata to accelerate eligible reads.
  * New configuration to control inline-read eligibility by part count.

* **Improvements**
  * GET object flow and SDK favor inline-prefetched data to reduce latency for eligible reads.
  * SDK read_object_md is now asynchronous (returns a Promise).

* **Tests**
  * Added integration tests covering inline prefetching and mapping queries.

### DIFF
--- a/config.js
+++ b/config.js
@@ -682,6 +682,11 @@ config.REMOTE_NOOAA_NAMESPACE = `remote-${config.KUBE_APP_LABEL}`;
 // FILES RELATED             //
 ///////////////////////////////
 config.INLINE_MAX_SIZE = 4096;
+// Maximum number of parts to prefetch in a single DB round-trip alongside read_object_md.
+// Matches the number of parts that read_object_stream fetches per range-aligned window
+// (IO_OBJECT_RANGE_ALIGN / CHUNK_SPLIT_AVG_CHUNK). Objects with more parts fall back to
+// the standard read_object_mapping RPC.
+config.MAPPINGS_PREFETCH_NUM_PARTS = config.IO_OBJECT_RANGE_ALIGN / config.CHUNK_SPLIT_AVG_CHUNK;
 
 ///////////////////////////////
 // CACHE (ACCOUNT, BUCKET)   //
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -472,16 +472,20 @@ module.exports = {
                     bucket: { $ref: 'common_api#/definitions/bucket_name' },
                     key: { type: 'string' },
                     obj_id: { objectid: true },
+                    size: { type: 'integer' },
                     start: { type: 'integer' },
                     end: { type: 'integer' },
                     location_info: { $ref: 'common_api#/definitions/location_info' },
+                    prefetched_chunks: {
+                        type: 'array',
+                        items: { $ref: '#/definitions/chunk_info' },
+                    },
                 },
             },
             reply: {
                 type: 'object',
-                required: ['object_md', 'chunks'],
+                required: ['chunks'],
                 properties: {
-                    object_md: { $ref: '#/definitions/object_info' },
                     chunks: {
                         type: 'array',
                         items: { $ref: '#/definitions/chunk_info' },
@@ -571,6 +575,7 @@ module.exports = {
                     key: { type: 'string' },
                     md_conditions: { $ref: '#/definitions/md_conditions' },
                     encryption: { $ref: 'common_api#/definitions/object_encryption' },
+                    should_prefetch_mappings: { type: 'boolean' },
                     adminfo: {
                         type: 'object',
                         properties: {
@@ -1567,6 +1572,10 @@ module.exports = {
                 object_owner: {
                     type: 'object',
                     properties: {}
+                },
+                prefetched_mappings: {
+                    type: 'array',
+                    items: { $ref: '#/definitions/chunk_info' }
                 },
             }
         },
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -42,6 +42,9 @@ async function get_object(req, res) {
     if (part_number) {
         md_params.part_number = part_number;
     }
+    if (!part_number) {
+        md_params.should_prefetch_mappings = true;
+    }
 
     const object_md = await req.object_sdk.read_object_md(md_params);
 
--- a/src/sdk/map_client.js
+++ b/src/sdk/map_client.js
@@ -89,6 +89,7 @@ class MapClient {
      * @param {Partial<nb.ObjectInfo>} [props.object_md]
      * @param {number} [props.read_start]
      * @param {number} [props.read_end]
+     * @param {nb.ChunkInfo[]} [props.prefetched_chunks]
      * @param {nb.LocationInfo} [props.location_info]
      * @param {nb.Tier} [props.move_to_tier]
      * @param {boolean} [props.check_dups]
@@ -103,6 +104,7 @@ class MapClient {
         this.object_md = props.object_md;
         this.read_start = props.read_start;
         this.read_end = props.read_end;
+        this.prefetched_chunks = props.prefetched_chunks;
         this.location_info = props.location_info;
         this.move_to_tier = props.move_to_tier;
         this.check_dups = Boolean(props.check_dups);
@@ -390,9 +392,11 @@ class MapClient {
             obj_id: this.object_md.obj_id,
             bucket: this.object_md.bucket,
             key: this.object_md.key,
+            size: this.object_md.size,
             start: this.read_start,
             end: this.read_end,
             location_info: this.location_info,
+            prefetched_chunks: this.prefetched_chunks,
         });
         return res.chunks.map(chunk_info => {
             // TODO: Maybe move this to map_reader?
--- a/src/server/object_services/map_reader.js
+++ b/src/server/object_services/map_reader.js
@@ -12,8 +12,48 @@ const MDStore = require('./md_store').MDStore;
 const map_server = require('./map_server');
 const db_client = require('../../util/db_client');
 const { ChunkDB } = require('./map_db_types');
+const { ChunkAPI } = require('../../sdk/map_api_types');
 const server_rpc = require('../server_rpc');
 const auth_server = require('../common_services/auth_server');
+const system_store = require('../system_services/system_store').get_instance();
+
+
+/**
+ * Builds prepared ChunkDB instances from pre-fetched parts and chunks_db rows.
+ * Groups each part with its matching chunk and calls prepare_chunks to resolve
+ * block locations before returning.
+ *
+ * @param {nb.PartSchemaDB[]} parts
+ * @param {nb.ChunkSchemaDB[]} chunks_db
+ * @returns {Promise<nb.Chunk[]>}
+ */
+async function assemble_chunks_from_parts(parts, chunks_db) {
+    const chunks_db_by_id = _.keyBy(chunks_db, '_id');
+    const chunks = parts.map(part =>
+        new ChunkDB({ ...chunks_db_by_id[part.chunk.toHexString()], parts: [part] })
+    );
+    await map_server.prepare_chunks({ chunks });
+    return chunks;
+}
+
+/**
+ * Fetch parts and chunks by range, build and return ChunkDB array.
+ * @param {nb.ID} obj_id
+ * @param {{ start: number, end: number }} rng
+ * @param {(a: any, b: any) => number} sorter
+ * @returns {Promise<nb.Chunk[]>}
+ */
+async function read_parts_mapping_postgres(obj_id, rng, sorter) {
+    const { parts, chunks_db } = await MDStore.instance().find_parts_chunks_blocks_by_range({
+        obj_id,
+        start_gte: rng.start - config.MAX_OBJECT_PART_SIZE,
+        start_lt: rng.end,
+        end_gt: rng.start,
+        sorter,
+    });
+    if (parts.length === 0) return [];
+    return assemble_chunks_from_parts(parts, chunks_db);
+}
 
 /**
  *
@@ -27,13 +67,30 @@ const auth_server = require('../common_services/auth_server');
  * @param {number} [start]
  * @param {number} [end]
  * @param {nb.LocationInfo} [location_info]
+ * @param {nb.ChunkInfo[]} [prefetched_chunks_info]
  * @returns {Promise<nb.Chunk[]>}
  */
-async function read_object_mapping(obj, start, end, location_info) {
+async function read_object_mapping(obj, start, end, location_info, prefetched_chunks_info) {
     // check for empty range
     const rng = sanitize_object_range(obj, start, end);
     if (!rng) return [];
 
+    let chunks;
+    if (config.DB_TYPE === 'postgres') {
+        // Combined query: parts + chunks + blocks in one round-trip.
+        // If pre-fetched chunk data was passed from read_object_md, use it to skip the DB
+        const sorter = location_info ? _block_sorter_local(location_info) : _block_sorter_basic;
+        if (prefetched_chunks_info && prefetched_chunks_info.length) {
+            chunks = prefetched_chunks_info.map(chunk_info => new ChunkAPI(chunk_info, system_store));
+        } else {
+            chunks = await read_parts_mapping_postgres(obj._id, rng, sorter);
+        }
+        if (chunks.length === 0) return [];
+        if (await update_chunks_on_read(chunks, location_info)) {
+            chunks = await read_parts_mapping_postgres(obj._id, rng, sorter);
+        }
+        return chunks;
+    }
     // find parts intersecting the [start,end) range
     const parts = await MDStore.instance().find_parts_by_start_range({
         obj_id: obj._id,
@@ -47,9 +104,7 @@ async function read_object_mapping(obj, start, end, location_info) {
     });
 
     if (parts.length === 0) return [];
-
-    let chunks = await read_parts_mapping(parts, location_info);
-
+    chunks = await read_parts_mapping(parts, location_info);
     if (await update_chunks_on_read(chunks, location_info)) {
         chunks = await read_parts_mapping(parts, location_info);
     }
@@ -230,6 +285,7 @@ function _block_sorter_local(location_info) {
 }
 
 // EXPORTS
+exports.assemble_chunks_from_parts = assemble_chunks_from_parts;
 exports.read_object_mapping = read_object_mapping;
 exports.read_object_mapping_admin = read_object_mapping_admin;
 exports.read_node_mapping = read_node_mapping;
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -1,5 +1,5 @@
 /* Copyright (C) 2016 NooBaa */
-/*eslint max-lines: ["error", 2300]*/
+/*eslint max-lines: ["error", 2400]*/
 'use strict';
 
 require('../../util/fips');
@@ -704,17 +704,21 @@ async function copy_object_mapping(req) {
  *
  */
 async function read_object_mapping(req) {
-    const { start, end, location_info } = req.rpc_params;
+    const { obj_id, key, size, start, end, location_info, prefetched_chunks } = req.rpc_params;
 
-    const obj = await find_object_md(req);
+    load_bucket(req);
+    const obj = {
+        _id: MDStore.instance().make_md_id(obj_id),
+        key,
+        size,
+    };
 
     // Check if the requesting account is authorized to read the object
     if (!await req.has_s3_bucket_permission(req.bucket, 's3:GetObject', '/' + obj.key, undefined)) {
         throw new RpcError('UNAUTHORIZED', 'requesting account is not authorized to read the object');
     }
 
-    const chunks = await map_reader.read_object_mapping(obj, start, end, location_info);
-    const object_md = get_object_info(obj);
+    const chunks = await map_reader.read_object_mapping(obj, start, end, location_info, prefetched_chunks);
 
 
     // only update the object read stats if the bucket has more than one tier
@@ -734,7 +738,6 @@ async function read_object_mapping(req) {
     }
 
     return {
-        object_md,
         chunks: chunks.map(chunk => chunk.to_api()),
     };
 }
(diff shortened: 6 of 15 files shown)


src/sdk/namespace_s3.js
```diff
@@ -199,15 +199,14 @@ class NamespaceS3 {
...
                 request.Range = `bytes=0-${config.INLINE_MAX_SIZE - 1}`;
             }
             this._set_md_conditions(params, request);
             this._assign_encryption_to_request(params, request);
             let res;
             try {
-                res = can_use_get_inline ?
+                res = params.can_use_get_inline ?
                     await this.s3.getObject(request) :
                     await this.s3.headObject(request);
```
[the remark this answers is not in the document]
nadavMiz: the whole point was to move this logic to s3_get_object so it would be in a centralized location. I removed the comment to avoid misunderstandings

src/server/object_services/object_server.js
```diff
@@ -850,10 +873,94 @@ async function read_object_md(req) {
...
+        object_md: info,
+        read_start: 0,
+        read_end: size,
+        rpc_client: server_rpc.client,
+        report_error: (block_md, action, err) => dbg.error('read_object_md inline block error', block_md, err),
+    });
+    await mc.read_chunks();
+    if (!mc.had_errors) {
+        const slices = _slice_buffers_in_range(chunks_api, 0, size);
+        if (slices) info.first_range_data = Buffer.concat(slices.map(s => s.data).filter(Boolean));
```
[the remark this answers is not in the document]
nadavMiz: @coderabbitai  is this scenario possible? MDStore.instance().find_object_with_mapping_by_key return the first parts of an object. so wouldn't it always start at 0 and be contentious? in what scenarios can this happaen?

dannyzaken:
I measured the new queries on a local DB with ~3 Mil objectmds. The results are very good for the execution time.
for `find_parts_chunks_blocks_by_range`:
```sql
nbcore=# EXPLAIN ANALYZE SELECT
                jsonb_agg(
                    jsonb_build_object(
                        'part', p.data,
                        'chunk', c.data,
                        'blocks', (
                            SELECT jsonb_agg(b.data)
                            FROM datablocks b
                            WHERE b.data->>'chunk' = c.data->>'_id'
                                AND b.data ? 'chunk'
                                AND (b.data->'deleted' IS NULL OR b.data->'deleted' = 'null'::jsonb)
                        )
                    )
                    ORDER BY (p.data->>'start')::bigint ASC
                ) AS mapping
            FROM objectparts p
            JOIN datachunks c ON c.data->>'_id' = p.data->>'chunk'
                AND (c.data->'deleted' IS NULL OR c.data->'deleted' = 'null'::jsonb)
            WHERE p.data->>'obj' = '68b125ecb0f069000d98856f'
                AND p.data ? 'obj'
                AND (p.data->'deleted' IS NULL OR p.data->'deleted' = 'null'::jsonb)
                AND (p.data->'uncommitted' IS NULL OR p.data->'uncommitted' = 'null'::jsonb)
                AND (p.data->>'start')::bigint >= -33554432
                AND (p.data->>'start')::bigint < 40225340
                AND (p.data->>'end')::bigint > 0;
                                                                                                                                      QUERY PLAN
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=45420.31..45420.32 rows=1 width=32) (actual time=0.557..0.558 rows=1 loops=1)
   ->  Nested Loop  (cost=0.58..562.32 rows=1 width=966) (actual time=0.332..0.333 rows=1 loops=1)
         ->  Index Scan using idx_btree_objectparts_obj_1_start_1 on objectparts p  (cost=0.29..554.01 rows=1 width=292) (actual time=0.303..0.304 rows=1 loops=1)
               Index Cond: ((data ->> 'obj'::text) = '68b125ecb0f069000d98856f'::text)
               Filter: ((((data -> 'uncommitted'::text) IS NULL) OR ((data -> 'uncommitted'::text) = 'null'::jsonb)) AND (((data ->> 'start'::text))::bigint >= '-33554432'::integer) AND (((data ->> 'start'::text))::bigint < 40225340) AND (((data ->> 'end'::text))::bigint > 0))
         ->  Index Scan using idx_btree_datachunks_id_desc on datachunks c  (cost=0.29..8.31 rows=1 width=674) (actual time=0.026..0.026 rows=1 loops=1)
               Index Cond: ((data ->> '_id'::text) = (p.data ->> 'chunk'::text))
   SubPlan 1
     ->  Aggregate  (cost=44857.96..44857.97 rows=1 width=32) (actual time=0.155..0.155 rows=1 loops=1)
           ->  Bitmap Heap Scan on datablocks b  (cost=472.16..44857.59 rows=148 width=364) (actual time=0.123..0.132 rows=1 loops=1)
                 Recheck Cond: (((data ->> 'chunk'::text) = (c.data ->> '_id'::text)) AND (data ? 'chunk'::text))
                 Rows Removed by Filter: 1
                 Heap Blocks: exact=2
                 ->  Bitmap Index Scan on idx_btree_datablocks_chunk  (cost=0.00..472.13 rows=14876 width=0) (actual time=0.109..0.110 rows=2 loops=1)
                       Index Cond: ((data ->> 'chunk'::text) = (c.data ->> '_id'::text))
 Planning Time: 2.683 ms
 Execution Time: 0.771 ms
(18 rows)
```

for `find_object_with_mapping_by_key`:
```sql
nbcore=# EXPLAIN ANALYZE WITH obj AS (
    SELECT o._id, o.data
    FROM objectmds o
    WHERE o.data->>'bucket' = '68c93cdfcc5792002252cab6'
        AND o.data->>'key' = 'application/90b436580744908b/1995254b5b0:19952c2936d:250260c5'
        AND (o.data->'deleted' IS NULL OR o.data->'deleted' = 'null'::jsonb)
        AND (o.data->'upload_started' IS NULL OR o.data->'upload_started' = 'null'::jsonb)
        AND (o.data->'version_past' IS NULL OR o.data->'version_past' = 'null'::jsonb)
    LIMIT 1
),
parts AS (
    SELECT p.data
    FROM obj
    JOIN objectparts p ON p.data->>'obj' = obj._id
        AND p.data ? 'obj'
        AND (p.data->'deleted' IS NULL OR p.data->'deleted' = 'null'::jsonb)
        AND (p.data->'uncommitted' IS NULL OR p.data->'uncommitted' = 'null'::jsonb)
    ORDER BY (p.data->>'start')::bigint ASC
    LIMIT 10
)
SELECT
    obj.data AS obj_data,
    jsonb_agg(
        jsonb_build_object(
            'part', parts.data,
            'chunk', c.data,
            'blocks', (
                SELECT jsonb_agg(b.data)
                FROM datablocks b
                WHERE b.data->>'chunk' = c.data->>'_id'
                    AND b.data ? 'chunk'
                    AND (b.data->'deleted' IS NULL OR b.data->'deleted' = 'null'::jsonb)
            )
        )
        ORDER BY (parts.data->>'start')::bigint ASC
    ) FILTER (WHERE parts.data IS NOT NULL) AS mapping
FROM obj
LEFT JOIN parts ON true
LEFT JOIN datachunks c ON c.data->>'_id' = parts.data->>'chunk'
    AND (c.data->'deleted' IS NULL OR c.data->'deleted' = 'null'::jsonb)
GROUP BY obj._id, obj.data;
                                                                                            QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 GroupAggregate  (cost=566.42..45424.43 rows=1 width=164) (actual time=0.127..0.129 rows=1 loops=1)
   Group Key: obj._id, obj.data
   CTE obj
     ->  Limit  (cost=0.41..8.43 rows=1 width=693) (actual time=0.022..0.022 rows=1 loops=1)
           ->  Index Scan using idx_btree_objectmds_latest_version_index on objectmds o  (cost=0.41..8.43 rows=1 width=693) (actual time=0.021..0.021 rows=1 loops=1)
                 Index Cond: (((data ->> 'bucket'::text) = '68c93cdfcc5792002252cab6'::text) AND ((data ->> 'key'::text) = 'application/90b436580744908b/1995254b5b0:19952c2936d:250260c5'::text))
   ->  Sort  (cost=557.98..557.99 rows=1 width=1098) (actual time=0.062..0.063 rows=1 loops=1)
         Sort Key: obj._id, obj.data
         Sort Method: quicksort  Memory: 26kB
         ->  Nested Loop Left Join  (cost=549.91..557.97 rows=1 width=1098) (actual time=0.054..0.056 rows=1 loops=1)
               ->  CTE Scan on obj  (cost=0.00..0.02 rows=1 width=132) (actual time=0.024..0.024 rows=1 loops=1)
               ->  Nested Loop Left Join  (cost=549.91..557.94 rows=1 width=966) (actual time=0.029..0.030 rows=1 loops=1)
                     ->  Limit  (cost=549.62..549.63 rows=1 width=300) (actual time=0.021..0.022 rows=1 loops=1)
                           ->  Sort  (cost=549.62..549.63 rows=1 width=300) (actual time=0.021..0.022 rows=1 loops=1)
                                 Sort Key: (((p.data ->> 'start'::text))::bigint)
                                 Sort Method: quicksort  Memory: 25kB
                                 ->  Nested Loop  (cost=0.29..549.61 rows=1 width=300) (actual time=0.015..0.016 rows=1 loops=1)
                                       ->  CTE Scan on obj obj_1  (cost=0.00..0.02 rows=1 width=100) (actual time=0.000..0.000 rows=1 loops=1)
                                       ->  Index Scan using idx_btree_objectparts_obj_1_start_1 on objectparts p  (cost=0.29..549.57 rows=1 width=292) (actual time=0.009..0.010 rows=1 loops=1)
                                             Index Cond: ((data ->> 'obj'::text) = (obj_1._id)::text)
                                             Filter: (((data -> 'uncommitted'::text) IS NULL) OR ((data -> 'uncommitted'::text) = 'null'::jsonb))
                     ->  Index Scan using idx_btree_datachunks_id_desc on datachunks c  (cost=0.29..8.31 rows=1 width=674) (actual time=0.006..0.006 rows=1 loops=1)
                           Index Cond: ((data ->> '_id'::text) = (p.data ->> 'chunk'::text))
   SubPlan 2
     ->  Aggregate  (cost=44857.96..44857.97 rows=1 width=32) (actual time=0.025..0.026 rows=1 loops=1)
           ->  Bitmap Heap Scan on datablocks b  (cost=472.16..44857.59 rows=148 width=364) (actual time=0.013..0.013 rows=1 loops=1)
                 Recheck Cond: (((data ->> 'chunk'::text) = (c.data ->> '_id'::text)) AND (data ? 'chunk'::text))
                 Heap Blocks: exact=1
                 ->  Bitmap Index Scan on idx_btree_datablocks_chunk  (cost=0.00..472.13 rows=14876 width=0) (actual time=0.008..0.008 rows=1 loops=1)
                       Index Cond: ((data ->> 'chunk'::text) = (c.data ->> '_id'::text))
 Planning Time: 1.127 ms
 Execution Time: 0.196 ms
(33 rows)
```

The planning time for these queries is a bit long (over 1.5 ms. sometimes I got over 10 ms). This is expected for a relatively complex query, but I think we can still optimize it in a follow-up PR. 
We can use a prepared statement (see [here](https://node-postgres.com/features/queries#prepared-statements) from node-postgres docs), which can be done by simply sending the query with a `name` parameter. This will cause the node-postgres module to create a prepared statement in the server. After 5 calls to this prepared statement, postgres will use the cached planning and skip the planning step. I tested it manually in psql and got these results after the first 5 queries:
```
nbcore=# PREPARE find_parts_chunks_blocks_by_range(text, bigint, bigint, bigint) AS SELECT
    jsonb_agg(
        jsonb_build_object(
            'part', p.data,
            'chunk', c.data,
            'blocks', (
                SELECT jsonb_agg(b.data)
                FROM datablocks b
                WHERE b.data->>'chunk' = c.data->>'_id'
                    AND b.data ? 'chunk'
                    AND (b.data->'deleted' IS NULL OR b.data->'deleted' = 'null'::jsonb)
            )
        )
        ORDER BY (p.data->>'start')::bigint ASC
    ) AS mapping
FROM objectparts p
JOIN datachunks c ON c.data->>'_id' = p.data->>'chunk'
    AND (c.data->'deleted' IS NULL OR c.data->'deleted' = 'null'::jsonb)
WHERE p.data->>'obj' = $1
    AND p.data ? 'obj'
    AND (p.data->'deleted' IS NULL OR p.data->'deleted' = 'null'::jsonb)
    AND (p.data->'uncommitted' IS NULL OR p.data->'uncommitted' = 'null'::jsonb)
    AND (p.data->>'start')::bigint >= $2
    AND (p.data->>'start')::bigint < $3
    AND (p.data->>'end')::bigint > $4;

nbcore=# EXPLAIN ANALYZE EXECUTE find_parts_chunks_blocks_by_range('68b125ecb0f069000d98856f', -33554432, 40225340, 0);
                                                                                                                          QUERY PLAN
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Aggregate  (cost=45420.31..45420.32 rows=1 width=32) (actual time=0.120..0.121 rows=1 loops=1)
   ->  Nested Loop  (cost=0.58..562.32 rows=1 width=966) (actual time=0.037..0.038 rows=1 loops=1)
         ->  Index Scan using idx_btree_objectparts_obj_1_start_1 on objectparts p  (cost=0.29..554.01 rows=1 width=292) (actual time=0.027..0.028 rows=1 loops=1)
               Index Cond: ((data ->> 'obj'::text) = $1)
               Filter: ((((data -> 'uncommitted'::text) IS NULL) OR ((data -> 'uncommitted'::text) = 'null'::jsonb)) AND (((data ->> 'start'::text))::bigint >= $2) AND (((data ->> 'start'::text))::bigint < $3) AND (((data ->> 'end'::text))::bigint > $4))
         ->  Index Scan using idx_btree_datachunks_id_desc on datachunks c  (cost=0.29..8.31 rows=1 width=674) (actual time=0.007..0.007 rows=1 loops=1)
               Index Cond: ((data ->> '_id'::text) = (p.data ->> 'chunk'::text))
   SubPlan 1
     ->  Aggregate  (cost=44857.96..44857.97 rows=1 width=32) (actual time=0.036..0.036 rows=1 loops=1)
           ->  Bitmap Heap Scan on datablocks b  (cost=472.16..44857.59 rows=148 width=364) (actual time=0.018..0.021 rows=1 loops=1)
                 Recheck Cond: (((data ->> 'chunk'::text) = (c.data ->> '_id'::text)) AND (data ? 'chunk'::text))
                 Rows Removed by Filter: 1
                 Heap Blocks: exact=2
                 ->  Bitmap Index Scan on idx_btree_datablocks_chunk  (cost=0.00..472.13 rows=14876 width=0) (actual time=0.013..0.013 rows=2 loops=1)
                       Index Cond: ((data ->> 'chunk'::text) = (c.data ->> '_id'::text))
 Planning Time: 0.039 ms
 Execution Time: 0.191 ms
(18 rows)
```
See the difference in the planning time. I got similar results for the other query. 

One caveat is that a prepared query is stored in the scope of a connection, so it will only be beneficial after 5 statement calls on the same client. For the MD store queries, we use a pool of up to 70 clients, so we can expect many "cold" clients under low load. But under high loads, which is what we want to optimize, we can expect to have many "hot" clients that will use the cached planning. 
In any case, we should address that in the next PRs and not this one.

src/server/object_services/object_server.js
```diff
@@ -850,10 +873,98 @@ async function read_object_md(req) {
         }
     }
 
+    // Populate first_range_data for the pre-fetched inline parts.
+    if (can_use_get_inline && inline_parts) {
+        await _read_inline_data(info, inline_parts, inline_chunks_db);
```
dannyzaken, line 878: I think that the real gain is the combined DB query. Reading the actual data here may have some negligible performance gain, but it adds unnecessary code duplication. I suggest adding the mappings info (`inline_parts` and `inline_chunk_db`) to the returned object_md and continuing with the regular read_object_stream. In `MapClient.run_read_object`, you can skip reading the parts you already have. With this approach, we can also tweak `config.INLINE_NUM_PARTS` to test the combined query for larger objects decoupled from the actual read.

src/api/object_api.js
```diff
@@ -1568,6 +1570,11 @@ module.exports = {
                     type: 'object',
                     properties: {}
                 },
+                first_range_data: { binary: true },
+                inline_chunks: {
```
dannyzaken, line 1574: rename to `prefetched_mappings`

src/api/object_api.js
```diff
@@ -571,6 +572,7 @@ module.exports = {
                     key: { type: 'string' },
                     md_conditions: { $ref: '#/definitions/md_conditions' },
                     encryption: { $ref: 'common_api#/definitions/object_encryption' },
+                    can_use_get_inline: { type: 'boolean' },
```
dannyzaken, line 575: rename to `prefetch_mappings`

config.js
```diff
@@ -682,6 +682,9 @@ config.REMOTE_NOOAA_NAMESPACE = `remote-${config.KUBE_APP_LABEL}`;
 // FILES RELATED             //
 ///////////////////////////////
 config.INLINE_MAX_SIZE = 4096;
+// Maximum number of parts to fetch in a single DB round-trip for the inline read fast path.
+// Objects with more parts than this fall back to the standard read_object_stream path.
+config.INLINE_NUM_PARTS = 1;
```
dannyzaken, line 687: rename to `MAPPINGS_PREFETCH_NUM_PARTS`

I think that this value should be `config.IO_OBJECT_RANGE_ALIGN / config.CHUNK_SPLIT_AVG_CHUNK`. This will give us 8 by default. This value will match the number of parts we usually get in read_object_stream`

src/server/object_services/object_server.js
```diff
@@ -688,24 +688,21 @@ async function copy_object_mapping(req) {
...
 /**
- *
  * read_object_mapping
- *
+ * @param {Object} req
  */
 async function read_object_mapping(req) {
     const { start, end, location_info } = req.rpc_params;
 
     const obj = await find_object_md(req);
```
dannyzaken, line 697: I think that this is also a DB query that we can eliminate

src/sdk/map_client.js
```diff
@@ -377,11 +377,47 @@ class MapClient {
...
     }
 
+    /**
+     * Runs a full object read: fetches chunk mapping then reads block data.
+     * Uses inline_chunks from object_md when available and covering the requested
+     * range, avoiding a round-trip to read_object_mapping.
+     */
     async run_read_object() {
-        this.chunks = await this.read_object_mapping();
+        this.chunks = this._use_inline_chunks() ?? await this.read_object_mapping();
```
dannyzaken, line 386: After following the flow again from object_io `read_object_stream`, I believe we should handle this in the layer of map_reader. We should pass the mappings down all the way to `map_reader.read_object_mapping` and then follow the flow as if we just read it from the DB. That way we will not skip code that is related to stats (in object server) and tiering. 

Also, after completing the read of these chunks, we should dereference the mappings object to avoid holding it in memory for the rest of the read.

dannyzaken (commented):
The naming comments are mostly a recommendation. I feel like using `inline` is not accurate. I'm open to other names as well :)

src/server/object_services/object_server.js
```diff
@@ -797,14 +794,36 @@ function _convert_rpc_params_to_bucket_policy_req(rpc_params) {
...
+        if (result) {
+            obj = result.obj;
+            prefetched_parts = result.parts;
+            prefetched_chunks_db = result.chunks_db;
+            // find_object_with_mapping_by_key does not call check_object_mode,
+            // so replicate the same guard that find_object_md applies.
+            check_object_mode(req, obj, 'NO_SUCH_OBJECT');
+        }
+    }
+    if (!obj) obj = await find_object_md(req);
```
dannyzaken: maybe we should put this in `else`.

src/server/object_services/object_server.js
dannyzaken: skip `find_object_md` if we already have it

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -42,6 +42,9 @@ async function get_object(req, res) {
     if (part_number) {
         md_params.part_number = part_number;
     }
+    if (!part_number && !req.params.key.endsWith('/')) {
```
dannyzaken: We can remove this condition and always prefetch

src/api/object_api.js
```diff
@@ -1568,6 +1574,11 @@ module.exports = {
                     type: 'object',
                     properties: {}
                 },
+                first_range_data: { binary: true },
```
dannyzaken: do we still need `first_range_data`?

src/sdk/map_client.js
```diff
@@ -377,22 +377,59 @@ class MapClient {
...
+    _take_prefetched_chunks_if_covering_range() {
+        const { prefetched_mappings } = this.object_md;
+        if (!prefetched_mappings || !prefetched_mappings.length) return undefined;
+
+        const read_start = this.read_start ?? 0;
+        const read_end = this.read_end ?? this.object_md.size;
+
+        // prefetched_mappings are ordered by part.start from offset 0; the last part's
+        // end must reach read_end for the range to be fully covered.
+        const last_part = prefetched_mappings[prefetched_mappings.length - 1].parts?.[0];
```
dannyzaken: Not sure I understand this function. Why do we need to check if the prefetched data covers the entire range? Can't we just use what we have as long as we start from the same offset? (should be 0 currently for prefetched data)

src/server/object_services/map_reader.js
```diff
@@ -27,31 +78,46 @@ const auth_server = require('../common_services/auth_server');
...
+            if (chunks.length === 0) return [];
+            if (await update_chunks_on_read(chunks, location_info)) {
+                chunks = await read_parts_mapping_postgres(obj._id, rng, sorter);
+            }
+        } catch (err) {
+            dbg.warn('read_object_mapping: postgres fast path failed, falling back to standard path', err);
+            chunks = undefined;
+        }
+    }
+    if (!chunks) {
```
dannyzaken: why do we need this fallback?

src/api/object_api.js
nadavMiz: right we don't its a leftover from the change before. its now only used in namespace_s3 that don't need this value in the schema

src/server/object_services/map_reader.js
nadavMiz: we don't its also a leftover from a test I made because the ci was failling I removed it. now we just let it throw in case it fails

src/sdk/map_client.js
```diff
@@ -382,17 +382,39 @@ class MapClient {
         await this.read_chunks();
     }
 
+    /**
+     * Filters prefetched chunk mappings to the requested range and clears the reference.
+     * @returns {nb.ChunkInfo[] | undefined}
+     */
+    _take_prefetched_chunks_for_range() {
```
dannyzaken: I suggest doing the prefetch calculation in object_io.read_object_stream. 
see here:
https://github.com/noobaa/noobaa-core/blob/f96dc21e1a17f49c403d6c5001459f4a10a4754f/src/sdk/object_io.js#L582-L614
If `params.start` is within the prefetched chunks range, adjust `requested_end` based on the chunks we already have and limit it by `params.end`. Then continue with read_object normally, passing the prefetched chunks all the way to map_reader.
The main reason for doin it in read_object_stream, is that you can acquire from the `_io_buffers_sem` the exact size that we have prefetched.

dannyzaken (approved):

PR #9515, opened 2026-03-25
Vectors - LanceDb table index for vectors and filterable MD fields

alphaprinz:
### Describe the Problem
User is expected to query vectors and filter by (filterable) MD fields.
This requires creating indexes in the LanceDb table.

### Explain the Changes
1. Create a vector index
2. Create an index for each filterable MD field.

1. Auto reindexing after table has enough row changes (additions/deletion)

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Vector index creation now accepts a metadata configuration to mark metadata fields as non-filterable and returns that configuration when fetching an index.
  * Uploads now use the full index metadata so non-filterable keys are respected; table/index creation is more robust and tolerates "not enough rows" for vector index setup.

* **Tests**
  * Integration test added to verify index creation with non-filterable metadata keys.

### DIFF
--- a/src/api/bucket_api.js
+++ b/src/api/bucket_api.js
@@ -1053,8 +1053,8 @@ module.exports = {
                     dimension: {
                         type: 'integer',
                         minimum: 1,
-                    }
-                    //TODO - non filterable MD keys
+                    },
+                    metadata_configuration: { $ref: 'common_api#/definitions/metadata_configuration' },
                 }
             },
             reply: {
--- a/src/endpoint/vector/ops/vector_index_get.js
+++ b/src/endpoint/vector/ops/vector_index_get.js
@@ -26,6 +26,7 @@ async function post_get_index(req, res) {
             dataType: vector_index_info.data_type,
             distanceMetric: vector_index_info.distance_metric,
             creationTime: vector_index_info.creation_time / 1000,
+            metadataConfiguration: vector_index_info.metadata_configuration,
         }
     };
 }
--- a/src/util/vectors_util.js
+++ b/src/util/vectors_util.js
@@ -60,11 +60,11 @@ class LanceConn extends VectorConn {
 
     }
 
-    async put_vectors(vector_bucket, vector_index, vectors, is_retry = false) {
+    async put_vectors(vector_bucket_name, vector_index, vectors, is_retry = false) {
 
-        dbg.log0("put_vectors vector_bucket =", vector_bucket, ", vector_index =", vector_index, ", vectors =", vectors);
+        dbg.log0("put_vectors vector_bucket =", vector_bucket_name, ", vector_index =", vector_index.name, ", vectors =", vectors);
         //note underscore is not allowed in vector bucket name
-        const table_name = vector_bucket + "_" + vector_index;
+        const table_name = vector_bucket_name + "_" + vector_index.name;
 
         let lance_vectors;
         if (is_retry) {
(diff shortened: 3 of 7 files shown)


src/util/vectors_util.js
```diff
@@ -180,6 +189,25 @@ class LanceConn extends VectorConn {
...
+        //schema is implicitly defined by the first inserted vectors, get md fields from a vector
+        const keys = Object.keys(vectors[0]);
+
+        for (const key of keys) {
+            //skip 'id' and 'vector' fields
+            if (key === 'id' || key === 'vector') continue;
+            //skip non-filterable metadata keys (ie, fields which user explicitly asked not to index)
+            if (non_filterable_metadata_keys.indexOf(key) > -1) continue;
+            //md field is searchable and should be indexed:
+            await table.createIndex(key);
```
romayalon: if this is a retry and we already created some of the indexes but not all of them, should we handle this situation?

src/api/bucket_api.js
```diff
@@ -1053,8 +1053,19 @@ module.exports = {
                     dimension: {
                         type: 'integer',
                         minimum: 1,
-                    }
-                    //TODO - non filterable MD keys
+                    },
+                    metadata_configuration: {
```
romayalon: 1. metadata_configuration should also be returned in getIndex() resonse
https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/s3vectors/command/GetIndexCommand/
2. Are there more APIs that should contain metadataConfigutation in the params/reply?
3. create_vector_index and other index related functions should be in bucket_api? is it a bucket related function?

romayalon (changes requested):

src/util/vectors_util.js
alphaprinz: A put_vector retry?
If so, in a put_vector retry, the table has already been created (by another, concurrent put_vectors), and we go into the `if (table)` branch (currently line 89).

src/api/bucket_api.js
alphaprinz: create_vector_index mistakenly didn't use the metadata_configuration in common_api. I will push a fix.
1. Added to vector_index_get.js.
2. Yes. Get/List vector index return vector index info, which contains the metadata_configutation field.
3. I would say yes.

src/endpoint/vector/ops/vector_index_get.js
```diff
@@ -26,6 +26,7 @@ async function post_get_index(req, res) {
             dataType: vector_index_info.data_type,
             distanceMetric: vector_index_info.distance_metric,
             creationTime: vector_index_info.creation_time / 1000,
+            metadataConfiguration: vector_index_info.metadata_configuration,
```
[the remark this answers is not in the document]
alphaprinz: This is correct. See https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_GetIndex.html

src/util/vectors_util.js
romayalon: table exists but not all of the indexes because some failed to get created

src/util/vectors_util.js
```diff
@@ -180,6 +192,32 @@ class LanceConn extends VectorConn {
...
+            if (key === 'vector') continue;
+            //skip non-filterable metadata keys (ie, fields which user explicitly asked not to index)
+            if (non_filterable_metadata_keys.indexOf(key) > -1) continue;
+            //md field is searchable and should be indexed:
+            try {
+                await table.createIndex(key);
+            } catch(err) {
+                //revert so retry is possible
+                dbg.error("Failed to created index for", key, ", err =", err);
+                this.lance.dropTable(name);
```
romayalon, line 246: why dropping the table, I would just skip the creation of existing indexes/add a try catch clause and if index already exist, continue, WDYT?

romayalon (approved):

PR #9516, opened 2026-03-25
vectors - list vectors - add segments and nextToken

alphaprinz:
### Describe the Problem
Allow object (vector buckets, vector indices, vectors) to be listed in several independent s3 op invocations.

### Explain the Changes
1. Implement nextToken and segment params.
2. For db objects (vector bucket and indices), skip to nextToken and return from there.
3. For vector segments, split up the table and use lance table.offset to get to next staring point.

1. nextToken and segments params were ignored.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **New Features**
  * Vector listing: pagination (maxResults now defaults to 500), segmentation (segmentCount/segmentIndex), and combined paged-segment retrieval; responses include nextToken until a segment is exhausted.

* **Bug Fixes**
  * Validation added for nextToken, segmentCount, and segmentIndex; malformed inputs return structured validation errors with field-level details.

* **Tests**
  * Integration tests added for pagination, segmentation, and combined paged-segment flows.

### DIFF
--- a/src/endpoint/vector/ops/vector_list_vectors.js
+++ b/src/endpoint/vector/ops/vector_list_vectors.js
@@ -1,6 +1,10 @@
 /* Copyright (C) 2025 NooBaa */
 'use strict';
-//const config = require('../../../../config');
+
+const _ = require('lodash');
+
+const { VectorError } = require('../vector_errors');
+
 const dbg = require('../../../util/debug_module')(__filename);
 
 /**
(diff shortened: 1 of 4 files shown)


src/util/vectors_util.js
```diff
@@ -106,21 +106,50 @@ class LanceConn extends VectorConn {
...
+        dbg.log0("lance list vectors offset =", offset, ", end =", end);
+
         const query = table.query();
-        if (limit) query.limit(limit);
-        //TODO - this won't work for large tables.
-        //support aws segments/tokens?
+        query.offset(offset);
+        //limit can't get us over end
+        const limit = offset + params.max_results >= end ? end - offset : params.max_results;
+        query.limit(limit);
```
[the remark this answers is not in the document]
alphaprinz: max_results has a default set in list vectors s3 op handler.

src/endpoint/vector/ops/vector_list_vectors.js
```diff
@@ -10,17 +13,45 @@ async function post_list_vectors(req, res) {
...
+     const delim = next_token.indexOf('_');
+    if (delim === -1) return false;
+    const split = next_token.split('_');
+    if (split.length !== 2) return false;
+    if (split[0] === '' || split[1] === '') return false;
+    const start = Number(split[0]);
+    const end = Number(split[1]);
+    if (Number.isNaN(start) || Number.isNaN(end)) return false;
+    if (start > end) return false;
+    return true;
```
tangledbytes, line 94: 1. We are allowing all the numbers are here, including decimals and negatives which won't make sense for offset calculation. I am not sure how lancedb deals with negative and float offsets.
2. [Minor] line 45 indentation.

tangledbytes: IMO should add `Number.isIntegner(start) && Number.isInteger(end) && start >= 0 && end > 0` .

src/util/vectors_util.js
```diff
@@ -106,21 +106,50 @@ class LanceConn extends VectorConn {
...
+        const end = Number(next_token_parts[1]);
+        dbg.log0("lance list vectors offset =", offset, ", end =", end);
+
         const query = table.query();
-        if (limit) query.limit(limit);
-        //TODO - this won't work for large tables.
-        //support aws segments/tokens?
+        query.offset(offset);
+        //limit can't get us over end
+        const limit = offset + params.max_results >= end ? end - offset : params.max_results;
```
tangledbytes, line 143: What happens if `params.mx_results >= end` but end and offset are same? In the sanity check we are OK with `start == end`. In that case `limit = 0`, what does lance do in that case?

src/util/vectors_util.js
```diff
@@ -106,21 +106,50 @@ class LanceConn extends VectorConn {
...
 
         const table = await this.get_table(table_name);
         //TODO - check if(!table)
+
+        //determine this request place in the table
+        if (!params.next_token) {
+            //if there's no next_token, we're either in a segmented list_vector op or not segmented
+            //for not-segmented, create a default dummy segment which will include all rows
+            params.segment_count = params.segment_count || 1;
+            params.segment_index = params.segment_index || 0;
```
tangledbytes, line 122: What if these are invalid? We don't seem to be checking them anywhere. Let's say segment_count is negative, that makes segment negative which makes start negative. I am not sure how lance reacts to negative offsets.

src/util/vectors_util.js
```diff
@@ -106,21 +106,50 @@ class LanceConn extends VectorConn {
...
+        const offset = Number(next_token_parts[0]);
+        const end = Number(next_token_parts[1]);
+        dbg.log0("lance list vectors offset =", offset, ", end =", end);
+
         const query = table.query();
-        if (limit) query.limit(limit);
-        //TODO - this won't work for large tables.
-        //support aws segments/tokens?
+        query.offset(offset);
+        //limit can't get us over end
```
tangledbytes, line 142: We can't go beyond end but we seem to have no limits on the params.next_token. A client can construct a request like `0_9999999` right?

src/util/vectors_util.js
alphaprinz: I have changed validation to fail end === offset.

src/util/vectors_util.js
alphaprinz: Added validation.

src/util/vectors_util.js
alphaprinz: Well, yes. Although note max_results limit is 1000.

src/endpoint/vector/ops/vector_list_vectors.js
alphaprinz: Added, thanks.

src/util/vectors_util.js
tangledbytes: max_results isn't being validated though.. right? It can be more than 1000 or negative too right?

tangledbytes (approved):
LGTM in general. Just one max_results validation related comment.

PR #9519, opened 2026-03-26
vectors - get vector bucket op

alphaprinz:
### Describe the Problem
Vectors API describe a get vector bucket op, which we are missing.

### Explain the Changes
1. Implement missing op.
2. Align vector_bucket_name param

1. 

### Testing Instructions:
1. https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_GetVectorBucket.html

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added a Get Vector Bucket endpoint to retrieve bucket metadata.

* **Refactor**
  * Standardized vector-bucket parameter name to "vector_bucket_name" across APIs, server handlers, and utilities.

* **SDK**
  * Added get_vector_bucket support in the SDK surface.

* **Tests**
  * Added an integration test validating the GetVectorBucket operation.

### DIFF
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -1529,15 +1529,15 @@ async function delete_public_access_block(req) {
 
 // UTILS //////////////////////////////////////////////////////////
 
-function validate_bucket_creation(req, buckets_by_name) {
-    if (req.rpc_params.name.unwrap().length < 3 ||
-        req.rpc_params.name.unwrap().length > 63 ||
-        net.isIP(req.rpc_params.name.unwrap()) ||
-        !VALID_BUCKET_NAME_REGEXP.test(req.rpc_params.name.unwrap())) {
+function validate_bucket_creation(req, buckets_by_name, name = req.rpc_params.name) {
+    if (name.unwrap().length < 3 ||
+        name.unwrap().length > 63 ||
+        net.isIP(name.unwrap()) ||
+        !VALID_BUCKET_NAME_REGEXP.test(name.unwrap())) {
         throw new RpcError('INVALID_BUCKET_NAME');
     }
     buckets_by_name = buckets_by_name || req.system.buckets_by_name;
-    const bucket = buckets_by_name && buckets_by_name[req.rpc_params.name.unwrap()];
+    const bucket = buckets_by_name && buckets_by_name[name.unwrap()];
 
     if (bucket) {
         if (system_store.has_same_id(bucket.owner_account, req.account)) {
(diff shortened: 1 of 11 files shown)


src/server/system_services/bucket_server.js
tangledbytes, line 2611: Duplicates?

src/server/system_services/bucket_server.js
```diff
@@ -2333,7 +2340,7 @@ async function list_vector_buckets(req) {
     return await list_vector_objects(req, system_store.data.vector_buckets, get_vector_bucket_info);
 }
 
-function find_vector_bucket(req, vector_bucket_name = req.rpc_params.name) {
+function find_vector_bucket(req, vector_bucket_name = req.rpc_params.vector_bucket_name) {
     dbg.log0("vector_bucket_name = ", vector_bucket_name, ", unw =", vector_bucket_name.unwrap());
     const vector_bucket = req.system.vector_buckets_by_name && req.system.vector_buckets_by_name[vector_bucket_name.unwrap()];
```
tangledbytes, line 2344: Will `req.system` be populated if we don't do `req.load_auth()`?

alphaprinz: Sorry, I'm not with you.
Why would req.load_auth() refresh req.system?
In auth_server.js I only see it's setting req.role.

tangledbytes: Maybe I am misunderstanding but I though the flow is:
1. Load auth data from JWT
2. JWT will have system name or system_id
3. `load_auth` will read that and load req.system (along with req.account)

```js
    req.load_auth = function() {
        if (req.auth) {
            if (req.auth.account_id) req.account = system_store.data.get_by_id(req.auth.account_id);
            else if (req.auth.email) req.account = system_store.data.accounts_by_email[req.auth.email];

            if (req.auth.system_id) req.system = system_store.data.get_by_id(req.auth.system_id);
            else if (req.auth.system) req.system = system_store.data.systems_by_name[req.auth.system];

            req.role = req.auth.role;
        }
    };
```

alphaprinz: I admit I don't know where req.system is populated during rpc handling.
But it's definitely there and a bunch of other function use it without load_auth() (eg get_bucket_tagging()->find_bucket().

tangledbytes: Actually you are right. I just checked, we have a middleware for loading auth in the RPC server.

tangledbytes (approved):

PR #9520, opened 2026-03-27
Added support for bucket error objects num metrics when target is unreachable

aayushchouhan09:
### Describe the Problem
When the target bucket is unreachable (network disconnection or deletion), the replication cycle aborts early and the bucket-level cycle metrics (`bucket_last_cycle_total_objects_num`, `bucket_last_cycle_replicated_objects_num`, `bucket_last_cycle_error_objects_num`) are never updated, leaving stale values from the last successful cycle.

### Explain the Changes
1. For log-based replication, moved candidate fetching before the target bucket check and report the candidate count as errors when the target is unreachable or missing.
2. Fixed pre-existing bug where copy_objects called non-existent `get_rule_status` instead of `get_rule_and_bucket_status` with missing bucket_status argument.
3. For scan-based replication, added `report_failed_replication_cycle()` that updates all replication metrics when the target is unreachable - sets `bucket_last_cycle_total_objects_num` to `storage_stats.objects_count`, `replicated to 0`, and `increments error counts by 1 per failed cycle`.

Screenshot from 2026-04-01 20-14-00

1. JIRA: https://redhat.atlassian.net/browse/DFBUGS-3827

### Testing Instructions:
1. Put replication policy for buckets (src-bucket and dst-bucket).
2. Somehow make target bucket unreachable for dst-bucket (example delete the target bucket for dst-bucket namespacestore).
3. Check replication bg_worker metrics for multiple replication cycles, verify if error metrics are increasing by 1 when target is unreachable:
`$ JWT_TOKEN=$(kubectl get secret/noobaa-metrics-auth-secret -o jsonpath={.data.metrics_token} | base64 -d)`
`$ kubectl exec -it noobaa-core-0 -- curl -k -H "Authorization: Bearer ${JWT_TOKEN}" http://localhost:8080/metrics/bg_workers`

Note: Try similar steps for both scan and log based.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Differentiate missing source vs destination buckets: missing sources stop processing without changing metrics; missing destinations mark targets unreachable and record failed replication cycles with totals, zero replicated, and errors equal to total.
  * Ensure errors during candidate processing or bucket-diffing update replication metrics to reflect failed cycles before surfacing the error.

* **New Features**
  * Added explicit reporting for failed replication cycles to improve operational visibility.

### DIFF
--- a/src/server/bg_services/log_replication_scanner.js
+++ b/src/server/bg_services/log_replication_scanner.js
@@ -72,12 +72,8 @@ class LogReplicationScanner {
                 const rule_id = rule.rule_id;
 
                 const { src_bucket, dst_bucket } = replication_utils.find_src_and_dst_buckets(rule.destination_bucket, replication_id);
-                if (!src_bucket || !dst_bucket) {
-                    dbg.error('log_replication_scanner: can not find src_bucket or dst_bucket object', src_bucket, dst_bucket);
-                    // mark target bucket as unreachable (uses bucket id since bucket was deleted/not found)
-                    if (src_bucket && !dst_bucket) {
-                        replication_utils.update_replication_target_status(src_bucket.name, String(rule.destination_bucket), false);
-                    }
+                if (!src_bucket) {
+                    dbg.error('log_replication_scanner: src_bucket not found for replication_id:', replication_id);
                     return;
                 }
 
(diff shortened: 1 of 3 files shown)


src/server/bg_services/log_replication_scanner.js
```diff
@@ -72,12 +72,8 @@ class LogReplicationScanner {
...
 
                 const { src_bucket, dst_bucket } = replication_utils.find_src_and_dst_buckets(rule.destination_bucket, replication_id);
-                if (!src_bucket || !dst_bucket) {
-                    dbg.error('log_replication_scanner: can not find src_bucket or dst_bucket object', src_bucket, dst_bucket);
-                    // mark target bucket as unreachable (uses bucket id since bucket was deleted/not found)
-                    if (src_bucket && !dst_bucket) {
-                        replication_utils.update_replication_target_status(src_bucket.name, String(rule.destination_bucket), false);
-                    }
+                if (!src_bucket) {
+                    dbg.error('log_replication_scanner: can not find src_bucket object', src_bucket);
```
alphaprinz: This would just print 'undefined'.
Maybe change to source bucket's name?

src/server/bg_services/log_replication_scanner.js
```diff
@@ -90,14 +86,35 @@ class LogReplicationScanner {
                 );
                 if (!candidates.items || !candidates.done) return;
 
+                const total = Object.keys(candidates.items).length;
```
alphaprinz, line 89: Is Objects.keys() needed?
What is candidates type?
Are you sure it it doesn't have a way to get its size?

src/server/bg_services/log_replication_scanner.js
```diff
@@ -90,14 +86,35 @@ class LogReplicationScanner {
...
+                if (!dst_bucket) {
+                    dbg.error('log_replication_scanner: can not find dst_bucket object', dst_bucket);
```
alphaprinz: Again, log would only print 'undefined'.
Change to destination bucket's name.

alphaprinz (commented):
All in all looks good.
Had some minor comments.

src/server/bg_services/log_replication_scanner.js
aayushchouhan09: `candidates.items` is `Record<string, { action, time }>` - a plain object. There's no `.size` property. `Object.keys()` is the right way to get the count.
It is called only once per replication cycle, I think we can keep it.

src/server/bg_services/log_replication_scanner.js
```diff
@@ -90,14 +86,35 @@ class LogReplicationScanner {
...
 
                 const sync_deletions = rule.sync_deletions || false;
 
-                await this.process_candidates(
-                    src_bucket, dst_bucket, candidates.items, sync_versions, sync_deletions, rule_id, replication_id);
+                try {
+                    await this.process_candidates(
+                        src_bucket, dst_bucket, candidates.items, sync_versions, sync_deletions, rule_id, replication_id);
+                } catch (err) {
+                    replication_utils.update_replication_prom_report(src_bucket.name, replication_id, {}, {
```
alphaprinz, line 114: Is err logged anywhere?
I would add a log here, just to be on the safe side.

src/server/utils/replication_utils.js
```diff
@@ -86,6 +86,23 @@ function update_replication_prom_report(bucket_name, replication_policy_id, rule
     core_report.set_replication_status(last_cycle_status);
 }
 
+function report_failed_replication_cycle(bucket_name, replication_id, rule_id, total) {
+    const core_report = prom_reporting.get_core_report();
```
alphaprinz, line 91: Maybe a log here also, just in case, can't hurt.

alphaprinz (approved):
LGTM, I would add some more logs to ease debugging in case of failures.

PR #9522, opened 2026-03-30
Vectors load bucket

alphaprinz:
### Describe the Problem
Some vector ops run on a specific vector bucket and vector index.
Load these object at the start of the rest request, to have a unified check whether op needed object are available.
(Similar to how s3 ops validate requested bucket is available).

Also some fixes to request authorization.

### Explain the Changes
1. Load vector bucket and vector index before op's handler.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added an operation to retrieve vector bucket information (metadata and creation time).

* **API Changes**
  * Vector bucket create and delete operations now use `vector_bucket_name` instead of `name`.
  * SDK surface updated to expose a get-vector-bucket call.

* **Tests**
  * Added an integration test validating vector bucket retrieval and response structure.

### DIFF
--- a/src/util/vectors_util.js
+++ b/src/util/vectors_util.js
@@ -300,17 +300,17 @@ async function create_fs_db(path = '/tmp/lance') {
     return db;
 }
 
-async function create_vector_bucket({name}) {
-    dbg.log0("create_vector_bucket name = ", name);
+async function create_vector_bucket({vector_bucket_name}) {
+    dbg.log0("create_vector_bucket name = ", vector_bucket_name);
     const vc = await getVectorConn();
     await vc.create_vector_bucket();
     dbg.log0("create_vector_bucket done");
 }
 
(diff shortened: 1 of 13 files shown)


src/util/vectors_util.js
```diff
@@ -230,17 +230,17 @@ async function create_fs_db(path = '/tmp/lance') {
     return db;
 }
 
-async function create_vector_bucket({name}) {
-    dbg.log0("create_vector_bucket name = ", name);
+async function create_vector_bucket({vector_bucket_name}) {
+    dbg.log0("create_vector_bucket name = ", vector_bucket_name);
     const vc = await getVectorConn();
```
romayalon, line 305: I would expect getVectorConn() to connect to the target bucket based on the namespace_resource of the loaded bucket

PR #9523, opened 2026-03-31
Remove lambda trigger upgrade script

tangledbytes:
### Explain the Changes
1. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Chores**
  * Automatic cleanup during system upgrade: deprecated bucket lambda configuration is removed and legacy lambda-related database tables are dropped. This runs as part of the upgrade with no user intervention, reducing unused metadata, simplifying storage, and removing obsolete infrastructure.

### DIFF
--- a/src/upgrade/upgrade_scripts/5.22.0/remove_lambda_triggers.js
+++ b/src/upgrade/upgrade_scripts/5.22.0/remove_lambda_triggers.js
@@ -1,0 +1,28 @@
+/* Copyright (C) 2025 NooBaa */
+"use strict";
+
+async function run({ dbg, system_store }) {
+    try {
+        dbg.log0('starting remove lambda_triggers from buckets...');
+        const buckets = system_store.data.buckets
+            .map(b => b.lambda_triggers && ({
+                _id: b._id,
+                $unset: { lambda_triggers: true }
+            }))
+            .filter(Boolean);
+        if (buckets.length > 0) {
+            dbg.log0(`removing "lambda_triggers" from buckets: ${buckets.map(b => b._id).join(', ')}`);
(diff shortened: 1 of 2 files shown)


nimrod-becker (approved):

src/upgrade/upgrade_scripts/5.22.0/remove_lambda_triggers.js
```diff
@@ -0,0 +1,28 @@
...
+        dbg.log0('starting remove lambda_triggers from buckets...');
+        const buckets = system_store.data.buckets
+            .map(b => b.lambda_triggers && ({
+                _id: b._id,
+                $unset: { lambda_triggers: true }
+            }))
+            .filter(Boolean);
+        if (buckets.length > 0) {
+            dbg.log0(`removing "lambda_triggers" from buckets: ${buckets.map(b => b._id).join(', ')}`);
+            await system_store.make_changes({ update: { buckets } });
```
naveenpaul1, line 15: Do we need to remove the actual lambda_triggers or it will also get deleted?

tangledbytes: I think removing them from the buckets and dropping the other 2 tables should be enough. Do you have any suggestion here? I haven't worked on the original lambda_triggers removal so it is possible that I might be missing something here.

naveenpaul1: it should be fine

naveenpaul1 (approved):

PR #9525, opened 2026-04-01
Vectors lance path

alphaprinz:
### Describe the Problem
Implement using NSR from vector bucket as the path for lance.

### Explain the Changes
1. Use NSR in vector bucket for lance path.
2. Remove work around for lance to use 'lance' object bucket.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added an API to retrieve vector bucket details (includes creationTime) and support creating vector buckets with custom namespace/subpath headers.

* **Refactor**
  * Streamlined vector request routing and SDK initialization; improved per-bucket connection management and lazy loading.

* **Bug Fixes**
  * Tightened bucket-scoped authorization and improved null-safety for request parameters.

* **Tests**
  * Added integration test validating vector bucket retrieval and creation timestamps.

### DIFF
--- a/src/endpoint/vector/ops/vector_bucket_create.js
+++ b/src/endpoint/vector/ops/vector_bucket_create.js
@@ -2,6 +2,9 @@
 'use strict';
 //const config = require('../../../../config');
 const dbg = require('../../../util/debug_module')(__filename);
+const config = require('../../../../config');
+
+const { VectorError } = require('../vector_errors');
 
 /**
  * https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_CreateVectorBucket.html
@@ -10,8 +13,19 @@ async function post_vector_bucket(req, res) {
 
     dbg.log0("post_vector_bucket body =", req.body, ", headers =", req.headers);
 
-    const ns_name = req.headers['x-noobaa-custom-ns'];
-    const subpath = req.headers['x-noobaa-custom-subpath'];
+    const ns_name = req.headers[config.VECTORS_NSR_HEADER];
+    const subpath = req.headers[config.NSFS_CUSTOM_BUCKET_PATH_HTTP_HEADER];
+    const vector_db_type = req.headers[config.VECTORS_DB_TYPE_HEADER] || 'lance';
+
+    //for now NS header is mandatory
+    if (!ns_name) {
+        throw new VectorError({
+            code: VectorError.ValidationException.code,
+            http_code: VectorError.ValidationException.http_code,
+            message: VectorError.ValidationException.message,
+            fieldList: [{path: config.VECTORS_NSR_HEADER, message: "Missing"}],
+        });
+    }
 
     const namespace_resource = {
         resource: ns_name,
@@ -22,6 +36,7 @@ async function post_vector_bucket(req, res) {
     await req.vector_sdk.create_vector_bucket({
         vector_bucket_name,
         namespace_resource,
+        vector_db_type
     });
 }
 
--- a/src/endpoint/vector/ops/vector_bucket_create.js
+++ b/src/endpoint/vector/ops/vector_bucket_create.js
@@ -8,10 +8,21 @@ const dbg = require('../../../util/debug_module')(__filename);
  */
 async function post_vector_bucket(req, res) {
 
-    dbg.log0("post_vector_bucket body =", req.body);
+    dbg.log0("post_vector_bucket body =", req.body, ", headers =", req.headers);
+
+    const ns_name = req.headers['x-noobaa-custom-ns'];
+    const subpath = req.headers['x-noobaa-custom-subpath'];
+
+    const namespace_resource = {
+        resource: ns_name,
+        path: subpath //not to be confused with nsr path
+    };
 
     const vector_bucket_name = req.body.vectorBucketName;
-    await req.vector_sdk.create_vector_bucket({ vector_bucket_name });
+    await req.vector_sdk.create_vector_bucket({
+        vector_bucket_name,
+        namespace_resource,
+    });
 }
 
 exports.handler = post_vector_bucket;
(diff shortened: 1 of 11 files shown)


src/util/vectors_util.js
tangledbytes: The signature has changed - retry carshes

src/server/system_services/bucket_server.js
```diff
@@ -2355,14 +2362,21 @@ function find_vector_index(req, vector_bucket_name, vector_index_name) {
 }
 
 function get_vector_bucket_info(vector_bucket) {
+    const nsr = pool_server.get_namespace_resource_info(vector_bucket.namespace_resource.resource);
```
tangledbytes, line 2364: We might want to add a null guard here? Like `if (vector_bucket.namespace_resource) { ... }`

src/endpoint/vector/ops/vector_bucket_get.js
```diff
@@ -0,0 +1,23 @@
...
+async function post_get_vector_bucket(req, res) {
+
+    dbg.log0("post_get_vector_bucket body =", req.body);
+
+    const vb = req.vector_bucket;
+
+    return {
+        vectorBucket: {
+            vectorBucketName: vb.name,
+            creationTime: vb.creation_time / 1000,
```
tangledbytes, line 17: Are we doing this ms to sec conversion in list as well?

src/util/vectors_util.js
```diff
@@ -182,103 +172,68 @@ class LanceConn extends VectorConn {
...
-            awsAccessKeyId: account.access_keys[0].access_key.unwrap(),
-            awsSecretAccessKey: account.access_keys[0].secret_key.unwrap(),
-            endpoint: `http://localhost:${config.ENDPOINT_PORT}`,
-            allowHttp: "true"
-        },
-    };
-}
-
-let lanceConn;
+const vector_bucket_name_to_conn = new Map();
```
tangledbytes: How about using a LRU so that this doesn't grows infinitely?

src/endpoint/vector/ops/vector_bucket_create.js
```diff
@@ -8,10 +8,20 @@ const dbg = require('../../../util/debug_module')(__filename);
  */
 async function post_vector_bucket(req, res) {
 
-    dbg.log0("post_vector_bucket body =", req.body);
+    dbg.log0("post_vector_bucket body =", req.body, ", headers =", req.headers);
+
+    const ns_name = req.headers['x-noobaa-custom-ns'];
+    const subpath = req.headers['x-noobaa-custom-subpath'];
```
tangledbytes: What if these are undefined?

src/server/system_services/bucket_server.js
alphaprinz: This is mandatory and validated at creation.

src/endpoint/vector/ops/vector_bucket_get.js
alphaprinz: Yes.

src/endpoint/vector/ops/vector_bucket_create.js
alphaprinz: ns_name is mandatory (or taken from account default, but this is not implemented yet), subpath is optinal.

tangledbytes: So will we create a broken vector bucket if aws s3vectors is used? Or will it work even with `resource: undefined`?

tangledbytes: From a quick check it seems that might breaak bucket listing as `get_vector_bucket_info` expects non-undefined nsr.

tangledbytes (commented):
My only remaining concern would be that we might break list-vector-buckets when a bucket is created with `aws s3vectors create-vector-bucket` as that won't provide the required ns in the header and we don't have a fallback.

src/endpoint/vector/ops/vector_bucket_create.js
```diff
@@ -8,10 +8,21 @@ const dbg = require('../../../util/debug_module')(__filename);
...
```
romayalon: @alphaprinz @jackyalbo @nimrod-becker 
I wonder if there is confusion here, I'm familiar with Jacky's changes - https://github.com/noobaa/noobaa-core/pull/9271
should these 2 be the same? is this a different requirement?
Jacky's changes are implementing a header for s3 bucket path (x-noobaa-custom-bucket-path) and not nsr + subpath headers

src/util/vectors_util.js
```diff
@@ -252,135 +243,81 @@ class LanceConn extends VectorConn {
...
-        lanceConn = new LanceConn({
-            path: lance_s3_opts ? "s3://lance" : '/tmp/lance',
-            opts: lance_s3_opts
+//const vector_bucket_name_to_conn = new Map();
+
+const vector_conns = new LRUCache({
+    name: 'VectorConn',
+    expiry_ms: 5000,
+    make_key: ({ owner_account, name }) => owner_account.id + name.unwrap(),
+    load: async vector_bucket => {
```
romayalon: this inline implementation is loading a very **specific** connection to a vector bucket on top of lanceDB and connection of type FS path in a **general** vector connection LRU cache, I think we should move it out to a function and handle it based on vector db type and if having path return path else the s3 connection
this could be a refactoring too if you prefer

src/endpoint/vector/ops/vector_bucket_create.js
alphaprinz: I don't think there's a definite answer.
Personally I can go either way.
I've changed the subpath to use the same name as Jacky's custom path.

romayalon: @alphaprinz ahh I think this requirement was coming from CES team, right @nimrod-becker  @jackyalbo?

tangledbytes (approved):

PR #9526, opened 2026-04-02
optimized load_blocks_for_chunks and find_chunks_by_dedup_key queries

dannyzaken:
### Describe the Problem

Two functions in `md_store.js` (`load_blocks_for_chunks` and `find_chunks_by_dedup_key`) that query the `datablocks` table (~3M rows) suffered from PostgreSQL choosing sequential scans when the number of chunk IDs in the `IN` clause exceeded ~10 values. The default `random_page_cost = 4.0` caused the planner to prefer seq scan over the existing `idx_btree_datablocks_chunk` B-tree index. `load_blocks_for_chunks` took ~2600ms for 38 chunks, and `find_chunks_by_dedup_key` added an extra DB round-trip by calling `load_blocks_for_chunks` separately after fetching chunks.

### Explain the Changes
1. **`load_blocks_for_chunks`**: Replaced the mongo-to-pg translation layer (`this._blocks.find(...)`) with a raw SQL query using `UNNEST + LATERAL + OFFSET 0`. The `OFFSET 0` acts as an optimization barrier that forces PostgreSQL into a nested loop with per-chunk Bitmap Index Scan on `idx_btree_datablocks_chunk`, reducing execution from ~2600ms to ~2ms for 38 chunks (~1000x faster).
2. **`find_chunks_by_dedup_key`**: Combined the two-step approach (find chunks by dedup_key, then call `load_blocks_for_chunks`) into a single `INNER JOIN` query between `datachunks` and `datablocks`. The dedup_key index filters chunks to a small set, so the planner reliably picks Nested Loop + index scans on both tables. The denormalized result set is deduplicated in JS using a `Map` keyed by chunk `_id`. This eliminates one DB round-trip entirely.
3. Bumped the eslint `max-lines` directive from 2200 to 2210 to accommodate the additional lines from the optimized implementations.

1. https://redhat.atlassian.net/browse/DFBUGS-5326

### Testing Instructions:
1. Run existing integration tests: `node src/test/integration_tests/db/test_md_store.js` — covers `find_chunks_by_dedup_key` with matching keys, non-matching keys, and empty arrays.
2. Verify `load_blocks_for_chunks` via any flow that reads object data (e.g. `map_reader`, `map_builder`, `map_deleter`).
3. On a production-scale DB (~3M datablocks rows), run `EXPLAIN ANALYZE` on both queries to confirm index scans are used instead of seq scans.

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Refactor**
  * Chunk fragments are now returned with their associated blocks populated in a single step, with optional ordering applied.
* **Bug Fixes**
  * Soft-deleted chunks and deleted blocks are excluded from results.
* **Tests**
  * Integration tests extended to verify fragment grouping, block population, ordering, and deletion handling.
* **Style**
  * Linting rule adjusted to allow a slightly larger file length.

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -1,5 +1,5 @@
 /* Copyright (C) 2016 NooBaa */
-/*eslint max-lines: ["error", 2200]*/
+/*eslint max-lines: ["error", 2210]*/
 'use strict';
 
 /** @typedef {typeof import('../../sdk/nb')} nb */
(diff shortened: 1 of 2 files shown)


src/server/object_services/md_store.js
```diff
@@ -1545,22 +1545,49 @@ class MDStore {
...
+            JOIN ${this._blocks.name} b
+              ON b.data ->> 'chunk' = c.data ->> '_id'
+             AND b.data ? 'chunk'
+             AND (b.data->'deleted' IS NULL OR b.data->'deleted' = 'null'::jsonb)
+            WHERE c.data ->> 'system' = $1
+              AND c.data ->> 'bucket' = $2
+              AND c.data ->> 'dedup_key' = ANY($3)
+              AND c.data ? 'dedup_key'
+              AND (c.data->'deleted' IS NULL OR c.data->'deleted' = 'null'::jsonb)
+            ORDER BY c._id DESC
```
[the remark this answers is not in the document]
dannyzaken: For dedup we don't really care about chunks with no blocks. in map_server.js, after getting the results from `find_chunks_by_dedup_key` we mark such frags with `is_accessible = false` and not consider them for dedup. Please verify me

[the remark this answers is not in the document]
dannyzaken: I'll update the test

karthikps97 (approved):

nimrod-becker:
@jackyalbo @tangledbytes @nadavMiz can you PTAL?

src/server/object_services/md_store.js
```diff
@@ -1545,22 +1545,49 @@ class MDStore {
...
-                AND (data ->> 'dedup_key' = ANY($3) 
-                AND data ? 'dedup_key') 
-                AND (data->'deleted' IS NULL OR data->'deleted' = 'null'::jsonb)) 
-            ORDER BY _id DESC;
-            `;
+            SELECT
+                c.data AS chunk_data,
+                b.data AS block_data
+            FROM ${this._chunks.name} c
+            JOIN ${this._blocks.name} b
```
tangledbytes, line 1552: Would `LEFT JOIN` be better here to include the chunks which have no associated blocks (due to any reason)?

src/server/object_services/md_store.js
```diff
@@ -1545,22 +1545,49 @@ class MDStore {
...
+              ON b.data ->> 'chunk' = c.data ->> '_id'
+             AND b.data ? 'chunk'
+             AND (b.data->'deleted' IS NULL OR b.data->'deleted' = 'null'::jsonb)
```
tangledbytes, line 1555: Are there cases where we set JSON `null` to the fields? Or is it just for precaution?

tangledbytes (approved):
LGTM in general

src/server/object_services/md_store.js
dannyzaken: commented [here](https://github.com/noobaa/noobaa-core/pull/9526#discussion_r3029162394) on this. Please verify me.

src/server/object_services/md_store.js
dannyzaken: just a precaution

PR #9530, opened 2026-04-08
vectors - require lance should be optional

alphaprinz:
### Describe the Problem
Noobaa can't start if lance is not present, fails on ppc64.

### Explain the Changes
1. Change require to require_optional.

1. https://redhat.atlassian.net/browse/DFBUGS-6299

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Chores**
  * Improved handling of an optional database integration so import failures are deferred until the integration is actually used.
  * If the optional database package is not installed, the app now reports a clear runtime error when attempting to connect instead of failing on startup.

### DIFF
--- a/src/util/vectors_util.js
+++ b/src/util/vectors_util.js
@@ -6,7 +6,8 @@ const config = require('../../config');
 const path = require('path');
 const LRUCache = require('./lru_cache');
 const system_store = require('../server/system_services/system_store').get_instance();
-const lance = require('@lancedb/lancedb');
+const js_utils = require('../util/js_utils');
+const lance = js_utils.require_optional('@lancedb/lancedb');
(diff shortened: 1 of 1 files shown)


src/util/vectors_util.js
tangledbytes, line 26: Shouldn't we add a check here for `if (!lance) throw new Error("...")`?

alphaprinz: It's already checked in vector_rest, which should be the beginning of all vector requests.
I will add here also, just in case.

tangledbytes (approved):

nimrod-becker (approved):

PR #9541, opened 2026-04-14
Refactor - moved executeSQL to PostgresClient class

dannyzaken:
### Explain the Changes
1. moved `executeSQL` from `PostgresTable` to `PostgresClient`
2. `executeSQL` accepts general SQL queries that can be unrelated to the actual table that it was being called on. make more sense as a general `PostgresClient` method.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Refactor**
  * Centralized SQL execution through the database client for more consistent pool usage.
  * Improved pool resolution with fallbacks for more reliable query routing.
  * Added safe no-op behavior when no database is configured to prevent failures in DB-less environments.

### DIFF
--- a/src/util/postgres_client.js
+++ b/src/util/postgres_client.js
@@ -689,43 +689,6 @@ class PostgresTable {
         return _do_query(client || this.get_pool(), q, 0);
     }
 
-    /**
-     * executeSQL takes a raw SQL query and params and runs it against
-     * the database. If `query_name` is passed then it prepares a
-     * statement on the first execution while the further executions
-     * will re-utilize the prepared statement (pre-parsed).
-     *
-     * @template T
-     *
-     * @param {string} query
-     * @param {Array<any>} params
-     * @param {{
-     *  query_name?: string,
-     *  preferred_pool?: string,
-     * }} [options = {}]
-     *
-     * @returns {Promise<import('pg').QueryResult<T>>}
-     */
-    async executeSQL(query, params, options = {}) {
-        /** @type {Pool} */
-        const pool = this.get_pool(options.preferred_pool);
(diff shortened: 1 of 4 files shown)


src/util/postgres_client.js
```diff
@@ -1361,6 +1324,58 @@ class PostgresClient extends EventEmitter {
...
+            values: params,
+        };
+
+        if (query_name) {
+            q.name = query_name;
+        }
+
+        const res = await _do_query(client, q, 0);
+        client.release();
+
```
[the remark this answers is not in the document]
dannyzaken: fixed by passing the pool as the client rather then explictly acqiring one with `connect`

src/util/postgres_client.js
```diff
@@ -1361,6 +1324,56 @@ class PostgresClient extends EventEmitter {
...
+     * @param {Array<any>} params
+     * @param {{
+     *   query_name?: string,
+     *   preferred_pool?: string,
+     * }} [options={}]
+     * @returns {Promise<import('pg').QueryResult<T>>}
+     */
+    async executeSQL(query, params, options = {}) {
+        const { query_name, preferred_pool = 'default' } = options;
+        /** @type {import('pg').Pool} */
```
alphaprinz, line 1360: import is needed here?
I think /** @type {Pool} */ should be ok.

alphaprinz (approved):
LGTM, just a little nit-picking about the 'import' in the Pool type.

src/util/postgres_client.js
dannyzaken: the `@type` here is redundant. removed it

PR #9545, opened 2026-04-16
[S3Vectors] Supporting bucket policies - Part III - Authorisation

jackyalbo:
### Describe the Problem
We will now enforce vector bucket policy authorisation for both NC and NB  

### Explain the Changes
1. Adding authorize request to vector_rest
2. Adding tests
3. changed APIs to use vector_bucket_name instead of name to better align with other vector ops

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Enforced authorization and policy-based access control for vector bucket policy operations.
  * Vector bucket policies are now returned as part of bucket info.
  * IAM action matching extended to support service wildcards (e.g., s3:*).
  * Vector bucket policy endpoints now use a unified vector bucket identifier parameter.

* **Tests**
  * Added comprehensive integration tests covering vector bucket policy CRUD, validation, and authorization scenarios.


src/test/integration_tests/api/vectors/test_vector_bucket_policy.js
```diff
@@ -0,0 +1,558 @@
...
+            try {
+                await s3_vectors_client.send(new s3vectors.GetVectorBucketPolicyCommand({
+                    vectorBucketName: VECTOR_BUCKET,
+                }));
+                assert.fail('Expected NoSuchVectorBucketPolicy after delete');
+            } catch (err) {
+                assert.strictEqual(err.name, 'NoSuchVectorBucketPolicy');
+            }
+        });
+    });
```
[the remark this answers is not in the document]
jackyalbo: As we can have more than one endpoint, but still want to have cache - I think we can't avoid cases of stale enforcement. I think it's the same to what we have in s3 policy.

src/endpoint/vector/vector_rest.js
```diff
@@ -212,6 +215,112 @@ function authenticate_request(req) {
     }
 }
 
+/**
```
nadavMiz, line 218: this entire process looks very similar to what we already do in s3_rest. maybe we can combine these paths and use vector specific functions when necessary. for example use authorize_request_vector_policy inside authorize_request. or use some kind of polymorphism. this can save us a lot of effort if we find issues that applies to both paths. or if we want to add changes that applies to both paths

src/util/access_policy_utils.js
```diff
@@ -188,12 +188,18 @@ async function has_access_policy_permission(policy, account, method, arn_path, r
     return 'IMPLICIT_DENY';
 }
 
+function _is_service_wildcard_match(action, method) {
```
nadavMiz, line 191: good idea. maybe we can also add action === '*' inside here so all wildcard matching logic is in one place, rather than keeping it as a separate condition in the caller.

alphaprinz (approved):
LTGM, though I admit I'm not familiar with all details.
I'll leave Nadav's comment to your consideration.
Good catch on the index cache name typo 👍

src/endpoint/vector/vector_rest.js
jackyalbo: As it's 2 different, separate flows I will keep it like this - we can maybe in the future think of some functionalities we can take out - most here are just different in the exception thrown

PR #9556, opened 2026-04-20
Fix default CA to be based on system certs instead of none

jackyalbo:
### Describe the Problem
When, for some reason, OCP failed to inject the CA list to our pods, noobaa-core/endpoint would start with an empty list of CAs instead of the system defaults - this fix is supposed to fix this.

### Explain the Changes
1. base ca list on the system default instead of empty list

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTPS certificate validation by updating how trusted CA certificates are assembled. When internal and external CA bundles are available, they’re combined; when the external bundle is missing or empty, the app retains system trust and adds the internal CA (when present).
* **Tests**
  * Added unit tests covering trusted CA bundle construction for both “both bundles present” and “external bundle missing/empty” scenarios, including certificate file and environment handling.

### DIFF
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -7,6 +7,7 @@ const util = require('util');
 const net = require('net');
 const url = require('url');
 const http = require('http');
+const tls = require('tls');
 const https = require('https');
 const crypto = require('crypto');
 const xml2js = require('xml2js');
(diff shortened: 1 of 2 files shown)


src/util/http_utils.js
```diff
@@ -44,10 +45,16 @@ const EXTERNAL_CA_CERTS = process.env.EXTERNAL_CA_CERTS || '/etc/ocp-injected-ca
...
-        fs_utils.try_read_file_sync(INTERNAL_CA_CERTS),
-        fs_utils.try_read_file_sync(EXTERNAL_CA_CERTS),
-    ].filter(Boolean))
+    ca: (() => {
+        const custom_certs = [
+            fs_utils.try_read_file_sync(INTERNAL_CA_CERTS),
+            fs_utils.try_read_file_sync(EXTERNAL_CA_CERTS),
+        ].filter(Boolean);
+        return custom_certs.length ? [
+            ...tls.getCACertificates('default'),
```
liranmauda, line 54: As far as I can tell, it was added in Node.js v22.15.0
Here is the list of nodes we are using (based on the `.nvmrc`)

5.16 -> 20.11.0
5.17 -> 20.17.0
5.18 -> 22.11.0
5.19 -> 22.11.0
5.20 -> 22.11.0
5.21 -> 24.4.1

For the older versions, we will have to bump nodeJS

tangledbytes (approved):

tangledbytes (approved):

PR #9561, opened 2026-04-22
Add error handler to the Azure stream to catch AbortError

naveenpaul1:
### Describe the Problem

In ODF 4.16.21, noobaa-endpoint pods backed by Azure Blob Storage are entering a CrashLoopBackOff. When the underlying network connection to Azure is interrupted mid-stream (likely due to timeouts or throttling), the @azure/storage-blob SDK throws an AbortError. NooBaa fails to catch this exception, resulting in an uncaughtException that fatally panics the Node.js process.

### Explain the Changes
1. Add error handler to the Azure stream to catch AbortError

1. https://redhat.atlassian.net/browse/DFBUGS-5838

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved blob download reliability: errors from the storage stream are now caught, logged (error name and message), and propagated into the existing failure path so download failures are consistently detected and cause the operation to fail.

### DIFF
--- a/src/sdk/namespace_blob.js
+++ b/src/sdk/namespace_blob.js
@@ -248,6 +248,12 @@ class NamespaceBlob {
                     this.container,
                     inspect(_.omit(params, 'object_md.ns')),
                     'callback res');
+                // Add error handler to the Azure stream to catch AbortError
+                res.readableStreamBody?.on('error', stream_err => {
+                    dbg.error('NamespaceBlob.read_object_stream: Azure stream error:', stream_err.name, stream_err.message);
+                    // Destroy count_stream with error - properly cleans up and propagates error
+                    count_stream.destroy(stream_err);
+                });
                 if (!res.readableStreamBody) throw new Error('NamespaceBlob.read_object_stream: download response is invalid');
                 return resolve(res.readableStreamBody.pipe(count_stream));
             }).catch(err => {


src/sdk/namespace_blob.js
```diff
@@ -244,11 +244,17 @@ class NamespaceBlob {
                 params.start,
                 params.end - params.start,
             ).then(res => {
-                dbg.log0('NamespaceBlob.read_object_stream:',
+                dbg.log0('NamespaceBlob.read_object_stream1111:',
```
tangledbytes: Do we want this change?

src/sdk/namespace_blob.js
```diff
@@ -244,11 +244,17 @@ class NamespaceBlob {
...
                     this.container,
                     inspect(_.omit(params, 'object_md.ns')),
                     'callback res');
                 if (!res.readableStreamBody) throw new Error('NamespaceBlob.read_object_stream: download response is invalid');
```
tangledbytes: Do we need to place this after adding the handler? I am just wondering if there could be a scenario where we throw here and we encounter the error on the stream as well, will nodejs panic again?

naveenpaul1: @tangledbytes Ya thats make sence, Will move empty check after the handler.

src/sdk/namespace_blob.js
```diff
@@ -248,6 +248,12 @@ class NamespaceBlob {
                     this.container,
                     inspect(_.omit(params, 'object_md.ns')),
                     'callback res');
+                // Add error handler to the Azure stream to catch AbortError
+                res.readableStreamBody.on('error', stream_err => {
+                    dbg.error('NamespaceBlob.read_object_stream: Azure stream error:', stream_err.name, stream_err.message);
+                    // Destroy count_stream with error - properly cleans up and propagates error
+                    count_stream.destroy(stream_err);
+                });
```
liranmauda, line 256: We should probably write this block after the throw 
```
if (!res.readableStreamBody) throw new Error('NamespaceBlob.read_object_stream: download response is invalid');
```

tangledbytes (approved):

src/sdk/namespace_blob.js
naveenpaul1: @liranmauda with this code `res.readabaleStreamBody?.on(...)` we can avoid the issue here

liranmauda: @naveenpaul1, this means that you check it twice. 
Why do you need to check it twice and not just throw when it is not there? 
This is redundant code.

src/sdk/namespace_blob.js
liranmauda: @tangledbytes In this case, if we use optional chaining, we will not place the handler regardless, and we will throw anyway.
If there is no stream, then there is no handler that we need to attach

PR #9562, opened 2026-04-23
vectors - bg delete, reindex

alphaprinz:
### Describe the Problem
Background maintenance for vectors for containerized.

### Explain the Changes
1. DB cleaner should clean deleted vector buckets and indices.
2. Vector index's vector index should be re-created after enough changes.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Configurable automatic reindexing for vector indexes (enable/disable, delay, threshold).
  * Background reindex scheduler that detects over-threshold indexes and runs reindexes.
  * API/SDK hooks to report and update "rows since index"; vector put/delete now contribute to this tracking.
  * Vector index creation initializes a rows-since-index counter; integrated reindex operation for the vector backend.

* **Bug Fixes / Maintenance**
  * System-store cleanup now includes vector-related collections.

### DIFF
--- a/src/server/bg_services/vector_buckets_reindexer.js
+++ b/src/server/bg_services/vector_buckets_reindexer.js
@@ -1,0 +1,69 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const dbg = require('../../util/debug_module')(__filename);
+const config = require('../../../config');
+const auth_server = require('../common_services/auth_server');
+const system_store = require('../system_services/system_store').get_instance();
+const vector_utils = require('../../util/vectors_util');
+
+
+class VectorBucketsReindexer {
+    constructor({client, name}) {
+        this.name = name;
+        this.client = client;
+    }
+
+    async run_batch() {
+        dbg.log0('Vector Bucket Reindexer: BEGIN');
+        if (!system_store.is_finished_initial_load) {
+            dbg.log0('waiting for system store to load');
+            return;
+        }
+        if (!system_store.data.systems[0]) {
+            dbg.log0('system does not exist, skipping');
+            return;
+        }
+
+        if (!system_store.data.vector_indices) {
+            dbg.log0("Vector Bucket Reindexer: no vector indices.");
+            return config.REINDEX_VECTOR_BUCKETS_DELAY;
+        }
+
+        for (const vi of system_store.data.vector_indices) {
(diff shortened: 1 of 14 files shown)


src/server/bg_services/vector_buckets_reindexer.js
```diff
@@ -0,0 +1,67 @@
...
+                        value: 0
+                    }, {
+                        auth_token: auth_server.make_auth_token({
+                            system_id: system_store.data.systems[0]._id,
+                            account_id: system_store.data.systems[0].owner._id,
+                            role: 'admin'
+                        })
+                    });
+
+                    vector_utils.reindex(vi.vector_bucket, vi);
```
tangledbytes: Does this need `await`?

src/server/bg_services/vector_buckets_reindexer.js
```diff
@@ -0,0 +1,67 @@
...
+            dbg.log0("vi.name =", vi.name, ", vi.rows_since_index =", vi.rows_since_index);
+
+            if (vi.rows_since_index > config.VECTOR_INDEX_ROWS_REINDEX) {
+
+                try {
+                    await this.client.bucket.update_rows_since_index({
+                        vector_bucket_name: vi.vector_bucket.name,
+                        vector_index_name: vi.name,
+                        op: 'SET',
+                        value: 0
```
tangledbytes, line 79: But what if there are new inserts in between of the runs?

src/server/bg_services/vector_buckets_reindexer.js
alphaprinz: Reindex can take a lot of time. I rather leave it without await.

src/server/bg_services/vector_buckets_reindexer.js
alphaprinz: This is mainly heuristic. It's not critical to be exact. I'd rather just leave it as is.

src/server/bg_services/vector_buckets_reindexer.js
tangledbytes: Then let's at least add a `.catch` or else in case of an exception it would either be caught by the runtime in older nodes or would cause a crash in the newer version (unhandled exceptions).

alphaprinz: Ok, added.

tangledbytes (approved):

PR #9575, opened 2026-05-01
Fix InternalError on concurrent PUT uploads to the same key

dannyzaken:
### Describe the Problem
Concurrent PUT uploads to the same bucket and key fail with `InternalError`. When two PUT requests target the same object key simultaneously, both call `find_object_null_version` which returns `null` for each (no existing completed object yet), so both enter the `update_object_by_id` code path to promote their upload to "latest". The first UPDATE succeeds and satisfies the `latest_version_index` unique partial index (`bucket, key` where `version_past=null`). The second UPDATE then violates that same unique constraint (Postgres error code `23505`), which was not handled and propagated as an unrecoverable `InternalError`. The same race can also occur on the bulk path (`complete_object_upload_latest_mark_remove_current_and_delete`) when both requests find an existing object and attempt the mark-and-replace transaction concurrently. S3 allows concurrent uploads to the same key — the resulting object content is undefined when the data differs between uploads, but the operation must not fail.

### Explain the Changes

The changes in this PR are meant to reduce the probability of getting an InternalError by introducing retries in case of duplicate key errors. This does not completely eliminate the problem, which can still occur if there is an unreasonably high number of concurrent uploads to the same key. 

1. **Retry logic in `complete_object_upload`** (`object_server.js`): Wrapped `_put_object_handle_latest` in `P.retry` (up to 3 attempts, 50ms delay) with `should_retry_func` that only retries on duplicate key errors. On retry, `_put_object_handle_latest` re-queries `find_object_null_version` and sees the state left by the winning transaction, taking the correct code path for the updated state.
2. **`should_retry_func` parameter for `P.retry`** (`promise.js`): Added an optional callback to selectively decide whether an error should trigger a retry. Non-matching errors are thrown immediately without consuming retry attempts. Also added a guard to skip `delay()` when `delay_ms` is `0`.
3. **Error propagation in md_store bulk operations** (`md_store.js`): Changed `throw new Error(...)` to `throw res.err || new Error(...)` in `remove_object_move_latest`, `insert_object_delete_marker_move_latest`, `complete_object_upload_latest_mark_remove_current`, and `complete_object_upload_latest_mark_remove_current_and_delete` so the original Postgres error (with `.code`) is preserved for retry logic inspection.
4. **`update_parts_in_bulk` error handling** (`md_store.js`): Made the function `async`, execute the bulk op with `await`, and throw `res.err` when present instead of silently returning a failed result.
5. **`MDStore.is_err_duplicate_key()`** (`md_store.js`): Added a convenience method that delegates to `db_client.instance().is_err_duplicate_key(err)`, keeping the error-detection logic close to the store layer.
6. **Rename `errmsg` → `pg_error`** (`postgres_client.js`): Renamed the catch variable in `BulkOp.execute()` for clarity — it holds a full Error object, not a string message.

1. https://redhat.atlassian.net/browse/DFBUGS-6548

### Testing Instructions:

1. Run the unit tests that exercise `P.retry`:  
   `npm run mocha` (full harness loads `src/test/unit_tests/util_functions_tests/test_promise_utils.js`), or run that file directly with mocha after `npm run build` if your environment requires the native addon for loading `promise.js`.
2. The **`P.retry`** suite in `src/test/unit_tests/util_functions_tests/test_promise_utils.js` covers:
   - Success on first attempt.
   - Multiple failures then success (retry until recovery).
   - Exhaustion of attempts when errors persist.
   - **`should_retry_func`**: retries when it returns true; stops after one attempt when it returns false.
   - **`DO_NOT_RETRY`**: no retry when the error carries that flag.
   - **`error_logger`**: invoked for errors that lead to a retry.
   - **`func(attempts)`**: remaining attempts passed through as documented.
   - **`delay_ms`**: when non-zero, elapsed wall time is at least one delay interval between attempts (single-delay scenario).
3. To reproduce the original bug manually: run two concurrent `aws s3 cp` commands to the same bucket/key against a NooBaa endpoint backed by Postgres — this creates a race that, without this fix, can yield `InternalError`. Timing-dependent; not consistently reproducible.

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of bulk-write failures and duplicate-key conflicts to surface the underlying database error and reduce silent failures.
  * Added automatic retry for object uploads that encounter duplicate-key conflicts (up to 3 attempts with short backoff).
  * Made database write-error reporting more consistent across operations.

* **Tests**
  * Added comprehensive tests covering retry behavior, conditional retrying, error handling, logging, and delay behavior.

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -1,5 +1,5 @@
 /* Copyright (C) 2016 NooBaa */
-/*eslint max-lines: ["error", 2210]*/
+/*eslint max-lines: ["error", 2220]*/
 'use strict';
 
 /** @typedef {typeof import('../../sdk/nb')} nb */
@@ -10,7 +10,6 @@ const moment = require('moment');
 const mongodb = require('mongodb');
 const mime = require('mime-types');
 
-const P = require('../../util/promise');
 const dbg = require('../../util/debug_module')(__filename);
 const db_client = require('../../util/db_client');
 const { decode_json } = require('../../util/postgres_client.js');
@@ -105,6 +104,10 @@ class MDStore {
 
     is_valid_md_id(id_str) {
         return mongodb.ObjectId.isValid(id_str);
+    }
+
+    is_err_duplicate_key(err) {
+        return db_client.instance().is_err_duplicate_key(err);
     }
 
     /////////////
@@ -469,7 +472,7 @@ class MDStore {
         if (!res.ok || res.nMatched !== 2 || res.nModified !== 2) {
             dbg.error('remove_object_move_latest: partial bulk update',
                 _.clone(res), old_latest_obj, new_latest_obj);
-            throw new Error('remove_object_move_latest: partial bulk update');
+            throw res.err || new Error('remove_object_move_latest: partial bulk update');
         }
     }
 
@@ -505,7 +508,7 @@ class MDStore {
         if (!res.ok || res.nMatched !== 1 || res.nModified !== 1 || res.nInserted !== 1) {
             dbg.error('insert_object_delete_marker_move_latest: partial bulk update',
                 _.clone(res), obj, delete_marker);
-            throw new Error('insert_object_delete_marker_move_latest: partial bulk update');
+            throw res.err || new Error('insert_object_delete_marker_move_latest: partial bulk update');
         }
         return delete_marker;
     }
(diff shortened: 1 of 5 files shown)


dannyzaken:
A more complete solution to this bug is to identify this race condition on duplicate key error, and complete the put successfully while marking the new object_md as deleted. I think that this solution requires a lot more testing and is riskier for backports. The retries solution should be sufficient for real-world scenarios.

liranmauda (approved):

PR #9585, opened 2026-05-06
vectors - create vector index - add validation to dataType and dimension

alphaprinz:
### Describe the Problem
Vector index can be created with wrong dimension.
Create vector index ignores dataType given by user (there's only one applicable value).

### Explain the Changes
1. Add validation to the fields.

1. https://redhat.atlassian.net/browse/DFBUGS-6758
2. https://redhat.atlassian.net/browse/DFBUGS-6757

### Testing Instructions:
1. See bugs.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Optional data_type parameter for vector index creation (supports float32).
* **Bug Fixes / Improvements**
  * Input validation for vector index dimension (1–4096) and data_type with clearer validation responses.

### DIFF
--- a/src/endpoint/vector/ops/vector_index_create.js
+++ b/src/endpoint/vector/ops/vector_index_create.js
@@ -3,6 +3,8 @@
 //const config = require('../../../../config');
 const dbg = require('../../../util/debug_module')(__filename);
 
+const { VectorError } = require('../vector_errors');
+
 /**
  * https://docs.aws.amazon.com/AmazonS3/latest/API/API_S3VectorBuckets_CreateIndex.html
  */
@@ -14,6 +16,8 @@ async function post_create_index(req, res) {
     const vector_bucket_name = req.body.vectorBucketName;
     const dimension = req.body.dimension;
     const distance_metric = req.body.distanceMetric;
+    const data_type = req.body.dataType;
+
     let metadata_configuration;
     if (req.body.metadataConfiguration) {
         metadata_configuration = {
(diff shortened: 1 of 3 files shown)


src/endpoint/vector/ops/vector_index_create.js
```diff
@@ -14,16 +16,37 @@ async function post_create_index(req, res) {
...
+
     let metadata_configuration;
     if (req.body.metadataConfiguration) {
         metadata_configuration = {
             non_filterable_metadata_keys: req.body.metadataConfiguration?.nonFilterableMetadataKeys
         };
     }
 
+    if (dimension < 1 || dimension > 4096) {
+        throw new VectorError({
```
shirady, line 29: I noticed that you don't send the `type` - 
https://github.com/noobaa/noobaa-core/blob/1040f36a76b2263f57af2975cbab053bfa1f54fd/src/endpoint/vector/vector_errors.js#L113-L118

shirady (approved):
LGTM
Would you want to move the supported type ('float32') to const so we can use it instead of writing it in a couple of places?

src/endpoint/vector/ops/vector_index_create.js
alphaprinz: I don't see it being used (correct me if I'm wrong).

shirady: If vectors have the S3 errors structure, then you probably don't need the `type` at all as part of the error.

PR #9589, opened 2026-05-07
fix: noobaa-log-processor Permission denied

sakshimunjal:
### Describe the Problem
Bug: https://redhat.atlassian.net/browse/DFBUGS-6697
The noobaa-log-processor sidecar container in the noobaa-core-0 StatefulSet fails on every startup with Permission denied errors. The container runs under OpenShift's restricted Security Context Constraints (non-root UID 10001, allowPrivilegeEscalation: false, all Linux capabilities dropped). The entrypoint scripts were making three assumptions that are incompatible with this security posture:

- logrotate.sh used crontab - to schedule log segregation, which writes to /var/spool/cron/ — a root-owned directory

- noobaa_logs.sh started crond, which tries to create /var/run/crond.pid — not writable by a non-root user

- noobaa_logs.sh started rsyslogd using default system settings; rsyslog.conf loaded imuxsock (requires creating /dev/log), used /var/lib/rsyslog as work directory (not writable), and local logging is enabled with $OmitLocalLogging off. These locations and operations are not allowed for a restricted non-root container.

This is a persistent configuration bug that reproduces on every pod start on any OpenShift cluster enforcing restricted SCCs.

### Explain the Changes

- Replaced crontab with a loop that runs every 5 minutes, removing the need to write to /var/spool/cron/, and removed crond entirely.

- Updated rsyslogd to use writable PID and work directories accessible to the non-root container user.

- Disabled imuxsock and local socket logging so rsyslogd no longer tries to create /dev/log, which is not permitted in restricted containers.

1. Fixed: noobaa-log-processor container crashes with Permission denied on /var/spool/cron/, /var/run/crond.pid, /dev/log, and /var/lib/rsyslog under OpenShift restricted SCC

### Testing Instructions:
Deploy from the branch/merged code on any OpenShift cluster with default restricted SCC:
1. Deploy MCG operator and NooBaa
2. Check the noobaa-log-processor container logs and verify Permission denied errors are not appearing
3. Confirm logrotate and rsyslogd are running

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Chores**
  * Updated logging infrastructure and log rotation scheduling mechanisms.
  * Refined logging system initialization and configuration to enhance system reliability.


nimrod-becker:
@tangledbytes can you PTAL?

src/deploy/NVA_build/rsyslog.conf
```diff
@@ -6,7 +6,7 @@
 #### MODULES ####
 
 # The imjournal module bellow is now used as a message source instead of imuxsock.
-$ModLoad imuxsock # provides support for local system logging (e.g. via logger command)
+#$ModLoad imuxsock # provides support for local system logging (e.g. via logger command)
```
naveenpaul1, line 9: why this comment out?

sakshimunjal: imuxsock was cauing permission denied error since it was trying to access /dev/log, which is admin only

tangledbytes:
I am wondering what caused this issue suddenly? One PR that comes to mind is #9394 . Could you probably check if that's the case?

naveenpaul1 (approved):

src/deploy/NVA_build/noobaa_logs.sh
```diff
@@ -21,8 +21,8 @@ mkdir -p /log/noobaa_bucket_logs
 chmod 777 /log/noobaa_bucket_logs
 
 
-/usr/sbin/crond  &
-/usr/sbin/rsyslogd  &
+mkdir -p /var/log/rsyslog
+/usr/sbin/rsyslogd -i /var/log/rsyslogd.pid &
```
tangledbytes, line 25: /usr/sbin/rsyslogd -i /var/log/rsyslog/rsyslogd.pid

tangledbytes (approved):

PR #9590, opened 2026-05-07
use prepared statements for db get querries

nadavMiz:
### Describe the Problem
following #9514 the two new statements still take a lot of time to prepare in the postgress DB

### Explain the Changes
1. add a statement name so postgress will use the statements as prepare statements and cache the plan for subsequent calls

1. https://redhat.atlassian.net/browse/RHSTOR-8846

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Chores**
  * Added a configuration toggle to enable or disable named prepared statements for PostgreSQL (enabled by default).
* **Improvements**
  * When enabled, database queries can be sent with stable, reusable names so the database can cache query plans, reducing latency and improving query stability and predictability.

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -2074,7 +2074,12 @@ class MDStore {
                 AND (p.data->>'end')::bigint > $4
         `;
         const values = [String(obj_id), start_gte, start_lt, end_gt];
-        const res = await db_client.instance().executeSQL(query, values, { preferred_pool: this._postgres_pool });
+        const res = await db_client.instance().executeSQL(query, values, {
+            preferred_pool: this._postgres_pool,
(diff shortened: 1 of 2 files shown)


src/server/object_services/md_store.js
```diff
@@ -2068,7 +2068,10 @@ class MDStore {
                 AND (p.data->>'end')::bigint > $4
         `;
         const values = [String(obj_id), start_gte, start_lt, end_gt];
-        const res = await db_client.instance().executeSQL(query, values, { preferred_pool: this._postgres_pool });
+        const res = await db_client.instance().executeSQL(query, values, {
```
dannyzaken, line 2077: We should put it behind a config, so we can toggle it off if needed. Other than that, looks good

dannyzaken (approved):

PR #9592, opened 2026-05-08
fix race condition when loading system_store in CORE source path

dannyzaken:
### Describe the Problem

Identified during warp test. When `SYSTEM_STORE_SOURCE=CORE`, uploads to S3 endpoints fail with:
```
TypeError: Cannot read properties of undefined (reading '<tier_id>')
    at SystemStoreData.get_by_id (system_store.js)
    at get tier (map_api_types.js)
```

The CORE load path in `system_store.js` eagerly replaced `this.data` with a fresh `SystemStoreData()` instance before calling `rebuild()`. Since the constructor does not initialize `idmap`, any concurrent request (e.g. an upload calling `get_by_id`) that lands during the async gaps between the assignment and `rebuild()` hits `undefined.idmap`, causing the TypeError. The DB path did not have this issue because it reads into a local variable and only assigns `this.data` at the last moment.

### Explain the Changes
1. Read CORE data into the existing `new_data` local variable instead of directly into `this.data`, keeping the old (valid) `this.data` alive during all async operations.
2. Defer `this.data` assignment to just before `rebuild()`, eliminating the race window, matching the pattern already used by the DB path.
3. Maintain `this.old_db_data` in the CORE success path so that if a subsequent load falls back to DB, the incremental merge has a valid base state.
4. Unrelated change - Remove unused second argument (`true`) passed to `map_deleter.delete_object_mappings()` in `objects_reclaimer.js`.

1. 

### Testing Instructions:
1. Set `SYSTEM_STORE_SOURCE=CORE` on the endpoint pod.
2. Run concurrent S3 uploads (e.g. with `warp`) while triggering a system store reload (e.g. by creating/editing a bucket or pool to trigger `publish_to_cluster`).
3. Verify uploads no longer fail with `TypeError: Cannot read properties of undefined` in `get_by_id`.
4. Verify that after a CORE load failure (e.g. simulate by temporarily stopping the core pod), the endpoint falls back to DB loading successfully.

- [ ] Doc added/updated
- [ ] Tests added

## Release Notes

* **Bug Fixes**
  * Improved object cleanup process for better reliability during garbage collection operations.
  * Enhanced system data loading with more robust fallback handling to ensure data integrity and availability during initialization.

### DIFF
--- a/src/server/bg_services/objects_reclaimer.js
+++ b/src/server/bg_services/objects_reclaimer.js
@@ -30,7 +30,7 @@ class ObjectsReclaimer {
         const reclaimed_objects_ids = [];
         await P.all(unreclaimed_objects.map(async obj => {
             try {
-                await map_deleter.delete_object_mappings(obj, true);
+                await map_deleter.delete_object_mappings(obj);
                 reclaimed_objects_ids.push(obj._id);
             } catch (err) {
                 dbg.error(`got error when trying to delete object ${obj.key} mappings :`, err);
--- a/src/server/system_services/system_store.js
+++ b/src/server/system_services/system_store.js
@@ -468,14 +468,14 @@ class SystemStore extends EventEmitter {
 
                 if (this.source === SOURCE.CORE) {
                     try {
-                        this.data = new SystemStoreData();
-                        await this._read_new_data_from_core(this.data);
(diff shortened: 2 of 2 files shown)


alphaprinz (approved):
Nice  👍

PR #9597, opened 2026-05-10
NC | Logging Enhancement

jackyalbo:
Add new missing logs and added more info to existing logs

### Describe the Problem
Missing internal server errors (data ingestion slowdown) caused by: invalid credentials, ECONNRESET during reads, aborted requests, and fast-failures that suggest timeouts.
Current gaps: no request_id in abort/reset logs; ECONNRESET silently dropped at HTTP layer; signature mismatch never logged; no request-level timing; errors not classified by type.

1. Fixed https://redhat.atlassian.net/browse/DFBUGS-5194#
2. Fixed https://redhat.atlassian.net/browse/ODFRFE-174

### Testing Instructions:
1. See that you can see more logs in cases of reset connections and authorizarion flows errors both in normal and high debug levels

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Chores**
  * Implemented comprehensive diagnostic logging with centralized request timing and ID tracking throughout S3 request handling.
  * Enhanced structured error reporting for read operations, authentication events, and connection failures with detailed metadata including error codes and operation duration.
  * Improved error categorization for read stream operations to support better troubleshooting and system monitoring.

[Review Change Stack](https://app.coderabbit.ai/change-stack/noobaa/noobaa-core/pull/9597)

### DIFF
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -124,6 +124,14 @@ async function get_object(req, res) {
         }
     }
 
+    dbg.log2('GET read start', {
+        request_id: req.request_id,
+        bucket: req.params.bucket,
+        key: req.params.key,
+        start: params.start,
+        end: params.end,
+        content_length: req.headers['content-length'],
+    });
     const read_stream = await req.object_sdk.read_object_stream(params, res);
     if (read_stream) {
         // if read_stream supports closing, then we handle abort cases such as http disconnection
@@ -132,7 +140,12 @@ async function get_object(req, res) {
             read_stream.destroy(new Error('abort read stream'));
         });
         read_stream.on('error', err => {
-            dbg.log0('read stream error:', err, req.path);
+            dbg.log0('GET read stream error', {
+                request_id: req.request_id,
+                code: err.code,
+                path: req.path,
+                duration_ms: req.start_time ? Date.now() - req.start_time : undefined,
+            }, err.message);
             res.destroy(err);
         });
         read_stream.pipe(res);
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -479,6 +479,17 @@ const nsfs_bucket_statfs_cache = new LRUCache({
 const nsfs_low_space_fsids = new Set();
 
 /**
+ * Classify the type of a read stream error for logging purposes.
+ * @param {Error & { code?: string }} err
+ * @returns {string}
+ */
+function _classify_read_stream_error(err) {
+    if (err.name === 'AbortError') return 'ABORTED';
+    if (err.code === 'ECONNRESET' || err.code === 'EPIPE') return 'CONNECTION_RESET';
+    return err.code || 'UNKNOWN';
+}
+
+/**
  * NamespaceFS map objets to files in a filesystem.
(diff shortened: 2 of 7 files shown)


liranmauda (approved):
LGTM

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -132,7 +140,12 @@ async function get_object(req, res) {
             read_stream.destroy(new Error('abort read stream'));
         });
         read_stream.on('error', err => {
-            dbg.log0('read stream error:', err, req.path);
+            dbg.error('GET read stream error', {
```
nadavMiz: I think we should keep it as log0. this can happen if client disconnect. which is the most common way it will show. and an error log will unnecessarily scare support in case of client disconnect.

src/sdk/namespace_fs.js
```diff
@@ -1201,7 +1211,11 @@ class NamespaceFS {
             return null;
 
         } catch (err) {
-            dbg.log0('NamespaceFS: read_object_stream error file', file_path, err);
```
nadavMiz, line 1204: again and in all cases where it was already log0. I think we should retain log0 just because how much disconnects are common

src/sdk/object_sdk.js
```diff
@@ -147,17 +147,32 @@ class ObjectSDK {
...
      * be used to signal async flows that abort was detected.
      * @see {@link https://nodejs.org/docs/latest/api/globals.html#class-abortcontroller}
-     * @param {import('http').IncomingMessage} req
+     * @param {nb.S3Request} req
      * @param {import('http').ServerResponse} res
      */
     setup_abort_controller(req, res) {
         res.once('error', err => {
-            dbg.log0('response error:', err, req.url);
+            const sys_err = /** @type {NodeJS.ErrnoException} */ (err);
```
nadavMiz, line 155: we have a lot of duplicate logs along the abort path already. so I am afraid we are making it very messy by making them all so verbose. but I am not sure how much larger we are making them. maybe having all the data at one log will be more useful? WDYT?

src/util/http_utils.js
```diff
@@ -1011,6 +1011,7 @@ function setup_endpoint_server(server) {
         (err, socket) => {
 
             if (err.code === 'ECONNRESET' || !socket.writable) {
+                dbg.error('ENDPOINT CLIENT CONNECTION RESET', err.code, socket.remoteAddress);
```
nadavMiz: wouldn't that just be another duplicate log along the path of abort?

src/sdk/namespace_fs.js
jackyalbo: changed back to log0

src/endpoint/s3/ops/s3_get_object.js
jackyalbo: back to log0

src/sdk/object_sdk.js
jackyalbo: let's not revert what we currently have - moved it back to log0 like all the rest

src/util/http_utils.js
jackyalbo: good call - will take this out

nadavMiz (approved):

PR #9611, opened 2026-05-13
Added md_store queries plan checker and fixed queries

dannyzaken:
### Describe the Problem

Several `md_store` PostgreSQL queries were not hitting their intended indexes, falling back to sequential scans. The root causes were:

1. **Predicate mismatches** -- WHERE clause expressions like `data->>'field' IS NULL` don't match partial index predicates defined as `data->'field' IS NULL OR data->'field' = 'null'::jsonb`, causing the planner to skip the index.
2. **Expression mismatches** -- `find_deleted_chunks` and `find_deleted_blocks` used the ORM `find()` method which generates `data->'deleted' < ...::jsonb`, but the `aggregate_by_delete_dates` indexes are built on `to_ts(data->>'deleted')`.
3. **Subquery mismatches** -- `delete_orphaned_delete_marker` used an OR condition (`deleted IS NULL OR delete_marker IS NOT NULL`) in its NOT EXISTS subquery that spanned across partial index boundaries, forcing a seq scan.
4. There was no automated way to detect these regressions -- query plan correctness was never verified in CI.

### Explain the Changes
1. **Aligned query predicates with partial index definitions** in `find_object_latest`, `delete_and_insert_deferred`, `remove_pending_multiparts`, `remove_noncurrent_versions`, and other functions -- changed `data->>'field' IS NULL` to `(data->'field' IS NULL OR data->'field' = 'null'::jsonb)` and `data->>'upload_started' IS NOT NULL` to `data ? 'upload_started'` to match the index predicates.
2. **Rewrote `find_deleted_chunks` and `find_deleted_blocks`** from ORM `find()` calls to raw SQL using `to_ts(data->>'deleted')` to match the `aggregate_by_delete_dates` expression index, following the same pattern already used by `find_deleted_objects`.
3. **Rewrote `delete_orphaned_delete_marker`** -- restructured the NOT EXISTS subquery to use `deleted IS NULL AND upload_started IS NULL` predicates that match `version_seq_index`, added bucket filter for efficient index lookup, and removed the OR condition that prevented index usage.
4. **Added `ExplainPlanChecker` utility** (`src/test/utils/explain_plan_checker.js`) that intercepts queries through `postgres_client`, runs `EXPLAIN (FORMAT JSON)` with `enable_seqscan=off`, and captures plans for verification.
5. **Added `test_query_plans` integration test** (`src/test/integration_tests/db/test_query_plans.js`) covering all `md_store` functions with parameter permutations matching real production callers. Tests assert that no sequential scan occurs (unless explicitly marked as `allow_miss`), and print a summary report with HIT/MISS/SKIP status.
6. **Fixed optional chaining bug** in `server_inter_process.js` that caused a TypeError during coretest startup.

1.

### Testing Instructions:
1. Run the query plan verification test: `node --allow-natives-syntax ./node_modules/.bin/_mocha src/test/integration_tests/db/test_query_plans.js`
2. Verify all test cases pass (no unexpected Seq Scans) and review the printed report for HIT/MISS status.
3. Run the full test suite (`npm test`) to confirm no regressions in existing `test_md_store` or other tests.
4. Optionally, verify individual query plans manually with `EXPLAIN (FORMAT JSON)` on a populated database for the rewritten functions: `find_deleted_chunks`, `find_deleted_blocks`, `delete_orphaned_delete_marker`.

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented a runtime error when a missing load-source parameter is encountered.

* **Performance Improvements**
  * Improved cleanup/delete queries to correctly handle null JSON fields and use timestamp-based filtering for chunk/block cleanup.

* **Tests**
  * Added a comprehensive query-plan integration suite and a test utility to capture and assert PostgreSQL EXPLAIN plans, ensuring index usage and preventing sequential scans.

[Review Change Stack](https://app.coderabbit.ai/change-stack/noobaa/noobaa-core/pull/9611)

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -153,16 +153,17 @@ class MDStore {
         if (delete_obj_id) {
             bulk.add_query(
                 `UPDATE ${this._objects.name} SET data = data || ${escapeLiteral(deleted_json)}::jsonb` +
-                ` WHERE _id = ${escapeLiteral(String(delete_obj_id))} AND data->>'deleted' IS NULL`
+                ` WHERE _id = ${escapeLiteral(String(delete_obj_id))}` +
+                ` AND (data->'deleted' IS NULL OR data->'deleted' = 'null'::jsonb)`
             );
         } else if (bucket_id && key) {
             bulk.add_query(
                 `UPDATE ${this._objects.name} SET data = data || ${escapeLiteral(deleted_json)}::jsonb` +
                 ` WHERE data->>'bucket' = ${escapeLiteral(String(bucket_id))}` +
                 ` AND data->>'key' = ${escapeLiteral(key)}` +
-                ` AND data->>'deleted' IS NULL` +
-                ` AND data->>'upload_started' IS NULL` +
-                ` AND data->>'version_enabled' IS NULL`
+                ` AND (data->'deleted' IS NULL OR data->'deleted' = 'null'::jsonb)` +
+                ` AND (data->'upload_started' IS NULL OR data->'upload_started' = 'null'::jsonb)` +
+                ` AND (data->'version_enabled' IS NULL OR data->'version_enabled' = 'null'::jsonb)`
             );
         }
         bulk.insert_many([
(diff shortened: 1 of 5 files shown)


src/server/object_services/md_store.js
```diff
@@ -1926,21 +1928,17 @@ class MDStore {
...
-                projection: {
-                    _id: 1,
-                    deleted: 1
-                },
-                preferred_pool: 'read_only'
-            })
-            .then(objects => db_client.instance().uniq_ids(objects, '_id'));
+    async find_deleted_chunks(max_delete_time, limit) {
+        const query_limit = limit || 1000;
+        const query = `SELECT _id
```
alphaprinz, line 1933: Projection has 'deleted' field. Is it not needed?

alphaprinz (approved):
Great work :)

Had a minor comment about the projection/select being different.
Not necessarily wrong, if field was not used.

src/server/object_services/md_store.js
dannyzaken: it was never used

PR #9620, opened 2026-05-15
KeyCloak | assume role with web identity KeyCloak

naveenpaul1:
### Describe the Problem
 Initial commit for Support Keycloak with MCG client STS
Its create a skelton for the KeyCloak integration with nessesary files and design

### Explain the Changes
1. Refactor assume_role_with_web_identity method to include both LDAP and KeyCloak
2. KeyCloak configuration mounted in endpoint pod
3. Create KeyCloak clinet and providers using KeyCloak configuration from mounted path
4. Verify KeyCloak access token inside noobaa endpoint
5. Return temporary access tokens and session token

1. https://redhat.atlassian.net/browse/RHSTOR-8075

### Testing Instructions:
1. Will do later with mode 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Keycloak/OIDC-based authentication for STS web-identity role assumption
  * Support for multiple OIDC providers with auto-discovery and runtime config reload
  * Session tags from OIDC tokens are extracted and included in generated credentials
  * Source identity now uses email/sub when available, not just DN
* **Chores**
  * Added JWKS handling dependency for OIDC token verification
* **Behavior**
  * Falls back to existing LDAP flow when no matching OIDC provider is configured

### DIFF
--- a/src/endpoint/sts/ops/sts_post_assume_role_with_web_identity.js
+++ b/src/endpoint/sts/ops/sts_post_assume_role_with_web_identity.js
@@ -18,7 +18,8 @@ async function assume_role_with_web_identity(req) {
     const expiration_time = Date.now() + duration_ms;
     let assumed_role;
     try {
-        assumed_role = await req.sts_sdk.get_assumed_ldap_user(req);
+        // CHANGED: Use unified method that supports both LDAP and OIDC/Keycloak
+        assumed_role = await req.sts_sdk.get_assumed_web_identity_role(req);
     } catch (err) {
         if (err.rpc_code === 'ACCESS_DENIED') {
             throw new StsError(StsError.AccessDeniedException);
@@ -29,12 +30,23 @@ async function assume_role_with_web_identity(req) {
         if (err.rpc_code === 'INVALID_WEB_IDENTITY_TOKEN') {
             throw new StsError({ ...StsError.InvalidIdentityToken, message: err.message });
         }
-        dbg.error('get_assumed_ldap_user error:', err);
+        dbg.error('get_assumed_web_identity_role error:', err);
         throw new StsError(StsError.InternalFailure);
     }
     // Temporary credentials are NOT stored in noobaa
     // The generated session token will store in it the temporary credentials and expiry and the role's access key
     const access_keys = await req.sts_sdk.generate_temp_access_keys();
+
+    // CHANGED: Include session tags in session token if present (for OIDC/Keycloak)
+    const session_token_data = {
+        access_key: access_keys.access_key.unwrap(),
+        secret_key: access_keys.secret_key.unwrap(),
+        assumed_role_access_key: assumed_role.access_key
+    };
+    // Add session tags if present (from Keycloak/OIDC tokens)
+    if (assumed_role.session_tags && Object.keys(assumed_role.session_tags).length > 0) {
+        session_token_data.session_tags = assumed_role.session_tags;
+    }
 
     return {
         AssumeRoleWithWebIdentityResponse: {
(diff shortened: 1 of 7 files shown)


src/util/keycloak_utils.js
```diff
@@ -0,0 +1,147 @@
...
+    static async discover(issuer_url) {
+        const well_known_url = new URL('.well-known/openid-configuration', issuer_url);
+        try {
+            const response = await make_https_request(
+                {
+                    method: 'GET',
+                    hostname: well_known_url.hostname,
+                    port: well_known_url.port || '443',
+                    path: well_known_url.pathname,
+                    rejectUnauthorized: false
```
VershaAgrawal: this should return true .false disables certificate validation

src/util/keycloak_utils.js
```diff
@@ -0,0 +1,147 @@
...
+            // Decode header to get kid
+            const decoded_header = jwt.decode(token, { complete: true });
+            if (!decoded_header) {
+                throw new Error('Invalid token format');
+            }
+
+            // Get signing key
+            const signing_key = await this.get_signing_key(decoded_header.header.kid);
+            // Verify token
+            const verified = jwt.verify(token, signing_key, {
```
VershaAgrawal, line 68: Should we add client id as well to check the token against the specific client?

src/util/keycloak_utils.js
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    /**
+     * Discover KeyCloak configuration from well-known endpoint
+     * @param {string} issuer_url - The access token to introspect
+     * @returns {Promise<Object>} - .well-known OIDC provider configuration object
+     */
+    static async discover(issuer_url) {
+        const well_known_url = new URL('.well-known/openid-configuration', issuer_url);
+        try {
```
VershaAgrawal, line 107: Please check statusCode first and fail with a clear error before parsing JSON.

src/sdk/sts_sdk.js
```diff
@@ -96,7 +98,10 @@ class StsSDK {
             role_config: account.role_config
         };
     }
-
+    /**
+     * Get aassumed LDAP user
```
VershaAgrawal: typo "assumed"

src/endpoint/sts/ops/sts_post_assume_role_with_web_identity.js
```diff
@@ -49,13 +61,9 @@ async function assume_role_with_web_identity(req) {
...
-                    SessionToken: sts_utils.generate_session_token({
-                        access_key: access_keys.access_key.unwrap(),
-                        secret_key: access_keys.secret_key.unwrap(),
-                        assumed_role_access_key: assumed_role.access_key
-                    }, duration_sec)
+                    SessionToken: sts_utils.generate_session_token(session_token_data, duration_sec)
                 },
-                SourceIdentity: assumed_role.dn,
+                SourceIdentity: assumed_role.dn || assumed_role.email || assumed_role.sub,
                 Provider: assumed_role.iss,
```
VershaAgrawal, line 67: The LDAP path only returns `dn` (. So for LDAP, `SubjectFromWebIdentityToken`, `Audience` and `Provider` will be empty even when assume-role succeeds. Is that fine ?

VershaAgrawal (changes requested):

src/endpoint/sts/ops/sts_post_assume_role_with_web_identity.js
naveenpaul1: assumed_role.email || assumed_role.sub added for KeyCloak

src/util/keycloak_utils.js
naveenpaul1: we are getting signing key based on kid, SO I think we dont need to send the client ID

VershaAgrawal (approved):

PR #9622, opened 2026-05-18
Wait for async operations to complete in db_cleaner

karthikps97:
### Describe the Problem
In db_cleaner, we have multiple instances where `filter` function is used with an `async` callback but we don't wait for the promise to resolve. The callbacks return a promise which is always a `true` for `filter`. The resultant array is invalid as the conditions were never checked.

To test this scenario,  update datachunks with `deleted` timestamp. db_cleaner still deletes the chunks with entries in objectparts table for it.

Initial state:
```text
nbcore=# SELECT COUNT(*)
            FROM datachunks d 
            where data->'deleted' is NULL;
 count 
-------
 12800
(1 row)
```

Datachunk with its corresponding objectpart:
```text
nbcore=# select _id from datachunks d where _id='6a0aa1d0846bb90022557f6a';
           _id            
--------------------------
 6a0aa1d0846bb90022557f6a
(1 row)

nbcore=#   select * from objectparts o where data->>'chunk'='6a0aa1d0846bb90022557f6a';
           _id            |                                                                                                                   data                         
                                                                                           
--------------------------+------------------------------------------------------------------------------------------------------------------------------------------------
-------------------------------------------------------------------------------------------
 6a0aa1d0846bb90022557f6b | {"_id": "6a0aa1d0846bb90022557f6b", "end": 133169152, "obj": "6a0aa1d0846bb90022557b67", "seq": 126, "chunk": "6a0aa1d0846bb90022557f6a", "star
t": 132120576, "bucket": "6a060c82868e31002276144d", "system": "6a060c82868e310022761445"}
```

Final state after db_cleaner runs:
```text
nbcore=# SELECT COUNT(*)
            FROM datachunks d 
            where data->'deleted' is not NULL;
 count 
-------
     0
(1 row)
```

db_cleaner log for each cycle:
`May-15 7:23:02.372 [BGWorkers/1042]    [L0] core.server.bg_services.db_cleaner:: DB_CLEANER: removed 1000 documents from md-store`

### Explain the Changes
1. Refactor the code for datachunks, pools and buckets to wait for the promise to resolve, so that the conditional expressions can be evaluated.
2. Use a single DB query for checking whether blocks or parts are present for a chunk. This removes an extra round trip for additional DB call.
3. Skip deleting block records whose blocks are unreclaimed even after they have expired and log such blocks.

db_cleaner only removes chunks which don't have reference:
image

1. 

### Testing Instructions:

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved background database cleanup to more reliably identify blocks/chunks and buckets/pools that are safe to remove.
  * Added a more accurate eligibility check to ensure candidates still have no remaining block/part references before deletion.
* **Refactor**
  * Updated cleanup selection to compute deletion candidates up front and avoid async predicate patterns.
  * De-duplicated deletion targets prior to applying database changes, reducing unnecessary work.

### DIFF
--- a/src/server/bg_services/db_cleaner.js
+++ b/src/server/bg_services/db_cleaner.js
@@ -67,13 +67,25 @@ async function clean_md_store(last_date_to_remove) {
         await P.map_with_concurrency(10, objects_to_remove, obj => db_delete_object_parts(obj));
         await MDStore.instance().db_delete_objects(objects_to_remove);
     }
-    const blocks_to_remove = await MDStore.instance().find_deleted_blocks(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
+    let blocks_to_remove = await MDStore.instance().find_deleted_blocks(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
+    blocks_to_remove = blocks_to_remove.filter(block => {
+        const reclaimed = Boolean(block?.data?.reclaimed);
+        if (!reclaimed) {
+            dbg.warn("DB_CLEANER: found block which is not reclaimed yet, ", JSON.stringify(block?.data));
+        }
+        return reclaimed;
+    });
+    blocks_to_remove = db_client.instance().uniq_ids(blocks_to_remove, '_id');
     dbg.log2('DB_CLEANER: list blocks:', blocks_to_remove);
     if (blocks_to_remove.length) await MDStore.instance().db_delete_blocks(blocks_to_remove);
     const chunks_to_remove = await MDStore.instance().find_deleted_chunks(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
-    const filtered_chunks = chunks_to_remove.filter(async chunk =>
-        !(await MDStore.instance().has_any_blocks_for_chunk(chunk)) &&
-        !(await MDStore.instance().has_any_parts_for_chunk(chunk)));
+    const filtered_chunks = (
+        await Promise.all(chunks_to_remove.map(async chunk => {
+            const has_blocks_or_parts = await MDStore.instance().has_any_blocks_or_parts_for_chunk(chunk);
+            return has_blocks_or_parts ? null : chunk;
+        }))
+    ).filter(Boolean);
+
     dbg.log2('DB_CLEANER: list chunks with no blocks and no parts to be removed from DB', filtered_chunks);
     if (filtered_chunks.length) await MDStore.instance().db_delete_chunks(filtered_chunks);
     dbg.log0(`DB_CLEANER: removed ${objects_to_remove.length + blocks_to_remove.length + filtered_chunks.length} documents from md-store`);
@@ -157,10 +169,25 @@ async function clean_system_store(last_date_to_remove) {
     dbg.log2('DB_CLEANER: list vector_buckets:', vector_buckets);
     dbg.log2('DB_CLEANER: list vector_indices:', vector_indices);
     dbg.log2('DB_CLEANER: list pools:', pools);
-    const filtered_buckets = buckets.filter(async bucket =>
-        !(await MDStore.instance().has_any_objects_for_bucket_including_deleted(bucket)));
-    const filtered_pools = pools.filter(async pool =>
-        !(await nodes_store.has_any_nodes_for_pool(pool)));
+    const filtered_buckets = (
+        await Promise.all(
+            buckets.map(async bucket => {
+                const hasObjects =
+                    await MDStore.instance().has_any_objects_for_bucket_including_deleted(bucket);
+                return hasObjects ? null : bucket;
+            })
+        )
+    ).filter(Boolean);
+
+    const filtered_pools = (
+        await Promise.all(
+            pools.map(async pool => {
+                const hasNodes = await nodes_store.has_any_nodes_for_pool(pool);
+                return hasNodes ? null : pool;
+            })
+        )
+    ).filter(Boolean);
+
     if (accounts.length || filtered_buckets.length || filtered_pools.length || vector_buckets.length || vector_indices.length) {
         await system_store.make_changes({
             db_delete: {
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -1959,6 +1959,20 @@ class MDStore {
             .then(obj => Boolean(obj));
     }
 
+    async has_any_blocks_or_parts_for_chunk(chunk_id) {
+        const query = `
+        SELECT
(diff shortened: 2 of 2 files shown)


src/server/object_services/md_store.js
```diff
@@ -1959,6 +1959,20 @@ class MDStore {
...
+        SELECT
+            EXISTS (SELECT 1 from ${this._parts.name} where data->>'chunk' = $1)
+            OR
+            EXISTS (SELECT 1 FROM ${this._blocks.name} where data->>'chunk' = $2)
+        AS has_reference;
+        `;
+        const result = await db_client.instance().executeSQL(query, [chunk_id, chunk_id], {
+            preferred_pool: 'read_only',
+        });
+        return Boolean(result.rows[0]?.has_reference);
```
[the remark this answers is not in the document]
karthikps97: This uses PostgreSQL and the deleted_chunks would be of type string.

src/server/bg_services/db_cleaner.js
```diff
@@ -71,9 +71,13 @@ async function clean_md_store(last_date_to_remove) {
     dbg.log2('DB_CLEANER: list blocks:', blocks_to_remove);
     if (blocks_to_remove.length) await MDStore.instance().db_delete_blocks(blocks_to_remove);
     const chunks_to_remove = await MDStore.instance().find_deleted_chunks(last_date_to_remove, config.DB_CLEANER_DOCS_LIMIT);
-    const filtered_chunks = chunks_to_remove.filter(async chunk =>
-        !(await MDStore.instance().has_any_blocks_for_chunk(chunk)) &&
-        !(await MDStore.instance().has_any_parts_for_chunk(chunk)));
+    const filtered_chunks = (
+        await Promise.all(chunks_to_remove.map(async chunk => {
```
tangledbytes, line 83: I would suggest to use something like `map_with_concurrency` (or just split the array manually -- too much work though). I think it might overwhelm node if there are 1000 concurrent requests.

dannyzaken: I agree with Utkarsh. This will not scale, and can take up all the clients in postgres_client. 
I think we can filter `chunks_to_remove` on the DB side in one or two queries, WDYT?

src/server/bg_services/db_cleaner.js
```diff
@@ -157,10 +161,25 @@ async function clean_system_store(last_date_to_remove) {
     dbg.log2('DB_CLEANER: list vector_buckets:', vector_buckets);
     dbg.log2('DB_CLEANER: list vector_indices:', vector_indices);
     dbg.log2('DB_CLEANER: list pools:', pools);
-    const filtered_buckets = buckets.filter(async bucket =>
-        !(await MDStore.instance().has_any_objects_for_bucket_including_deleted(bucket)));
-    const filtered_pools = pools.filter(async pool =>
-        !(await nodes_store.has_any_nodes_for_pool(pool)));
+    const filtered_buckets = (
```
dannyzaken, line 172: Same comment as above. This will not scale

src/server/bg_services/db_cleaner.js
karthikps97: Yes, but wasn't the earlier code similar with all the queries being executed simultaneously? The earlier code used to have 2 DB calls per chunk, which is reduced to a single query now.

dannyzaken: @karthikps97, you are correct that this is not a new problem, but I think that it is a major one. Since you're already fixing this code path, what is the effort to make it more scalable?

src/server/bg_services/db_cleaner.js
dannyzaken: @karthikps97, we should avoid deleting blocks that are not `reclaimed`, but `find_deleted_blocks` does not check that. I think that under normal circumstances, we should expect all deleted blocks that meet the age cutoff to already be reclaimed. In that case, I would not change the query in `find_deleted_blocks`, and instead filter out the returned blocks that are not reclaimed. We should probably log if such blocks are found.

src/server/object_services/md_store.js
```diff
@@ -1959,6 +1959,20 @@ class MDStore {
...
+        const query = `
+        SELECT
+            EXISTS (SELECT 1 from ${this._parts.name} where data->>'chunk' = $1)
+            OR
+            EXISTS (SELECT 1 FROM ${this._blocks.name} where data->>'chunk' = $2)
+        AS has_reference;
+        `;
+        const result = await db_client.instance().executeSQL(query, [chunk_id, chunk_id], {
+            preferred_pool: 'read_only',
+        });
```
dannyzaken: I think that these queries will perform sequential scans and not use the indexes.

dannyzaken (commented):
Hi @karthikps97, I want to refresh this PR. See my additional comments regarding the deleted blocks cleanup and the indexes.

src/server/object_services/md_store.js
karthikps97: Yes, these will do seq scans. I will update the query to include the full index conditions, so that it will use the index. Thanks.

src/server/bg_services/db_cleaner.js
karthikps97: Yes, this also seems to be an issue. I'll filter the blocks which are not yet reclaimed and log them under `warn` level.

src/server/object_services/md_store.js
karthikps97: Updated the query so that it would perform index scan. Result:
`Result  (cost=1.91..1.92 rows=1 width=1) (actual time=0.150..0.150 rows=1 loops=1)
  InitPlan 1 (returns $0)`
   ` ->  Index Scan using idx_btree_objectparts_chunk on objectparts  (cost=0.42..35.14 rows=512 width=0) (actual time=0.065..0.065 rows=0 loops=1)`
          `Index Cond: ((data ->> 'chunk'::text) = '6a0aa1d2846bb900225615de'::text)
  InitPlan 2 (returns $1)`
    `->  Index Scan using idx_btree_datablocks_chunk on datablocks  (cost=0.42..3083.08 rows=3074 width=0) (actual time=0.082..0.082 rows=0 loops=1)`
          `Index Cond: ((data ->> 'chunk'::text) = '6a0aa1d2846bb900225615de'::text)`
`Planning Time: 1.565 ms
Execution Time: 0.165 ms`

src/server/bg_services/db_cleaner.js
karthikps97: @dannyzaken, As you mentioned, under normal conditions the blocks should be reclaimed. But if they go unreclaimed, could we end up fetching the same unreclaimed records in each cycle rendering the db_cleaner ineffective?

dannyzaken (approved):

PR #9593, opened 2026-05-08
Added fix for resetting of replication_target_status to default

aayushchouhan09:
### Describe the Problem
replication_target_status could stay scraped after replication was removed or labels changed, because the gauge was only updated with set() and never cleared.

### Explain the Changes
1. Replication metrics: Call reset_replication_target_status() at the start of ReplicationScanner.scan() (same reset-then-repaint idea as set_bucket_status), with helpers on NooBaaCoreReport and replication_utils. Log replication scanner stays set-only so both scanners do not wipe the same gauge.

### Issues: Fixed #xxx / Gap #xxx
1.  Fixed: https://redhat.atlassian.net/browse/DFBUGS-6398

### Testing Instructions:
1. Create two obc and set bucket replication, then copy some objects to src bucket.
2.  Check if replication is working fine, then delete the destination obc and wait for `NooBaaReplicationTargetUnreachable` alert firing.
3.  Now, delete the source obc and on after the next replication cycle check whether the alert is removed.

`prometheus dashboard flow:

0) Prereqs: kubectl and helm work against your cluster. Set once:
export NS=default

1) Add Helm repo
helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
helm repo update

2) Install or upgrade kube-prometheus-stack (match NooBaa labels):
helm upgrade prometheus prometheus-community/kube-prometheus-stack --install --namespace $NS --create-namespace --set prometheus.prometheusSpec.serviceMonitorSelector.matchLabels.app=noobaa --set prometheus.prometheusSpec.ruleSelector.matchLabels.prometheus=k8s --set prometheus.prometheusSpec.ruleSelector.matchLabels.role=alert-rules

Check it applied:
helm -n $NS get values prometheus
kubectl -n $NS get prometheus -o yaml | grep -A30 serviceMonitorSelector

3) Fix TLS on NooBaa ServiceMonitors (minimal tlsConfig)
Replace tlsConfig so there is no caFile, no ca: {}, no cert: {} — only skip-verify and SNI:

kubectl -n "$NS" patch servicemonitor noobaa-mgmt-service-monitor --type=json -p='[
  {"op":"replace","path":"/spec/endpoints/0/tlsConfig","value":{"insecureSkipVerify":true,"serverName":"noobaa-mgmt.'"$NS"'.svc"}},
[...]

PR #9623, opened 2026-05-18
Fix replication_target_status metrics reset and stale series cleanup

aayushchouhan09:
### Describe the Problem
Each scan called `reset()` on replication_target_status but only republished one rule per policy, so multi-rule policies lost metric series every cycle (flapping/absent alerts). Stale series also remained after policy delete or rule changes.

### Explain the Changes
1. Removed per-scan `reset()`
2. bg_workers only — reachability is set to 0/1 per (replication_id, source, target) after each scan; web clears do not affect scraped metrics.
3. Reconcile at scan start — drops stale series for deleted policies, missing source, or outdated rules/labels (no full gauge reset).
4. Helpers — `set_replication_target_status` and `clear_replication_target_status_by_replication_id` in noobaa_core_report, used via replication_utils (update, clear_orphan, reconcile).

1. Fixed: https://redhat.atlassian.net/browse/DFBUGS-6398
2. Gap: #9593 

### Testing Instructions:
1.  To run unit tests: `npx jest src/test/unit_tests/internal/test_replication_metrics.test.js`

Manual testing steps:
1. Create 4 destination/target buckets obcs dst-n,dst-s,dst-w and dst-e.
2. Create two policies policy1.json and policy2.json, and put two buckets in each replication policy.

> $ cat policy.json 
{
  "rules": [
    {
      "rule_id": "rule-1",
      "destination_bucket": "<dst-n-name>"
    },
    {
      "rule_id": "rule-2",
      "destination_bucket": "<dst-s-name>"
    }
  ]
}

3. Create two src buckets obcs src-ns and src-we by using above policies.
4. Verify if replication is working fine, and also check the replication_target_status metrics using below commands (It should contain 4 metrics for each replication rule_id having reachable state).                                                                          
`$ JWT_TOKEN=$(kubectl get secret/noobaa-metrics-auth-secret -o jsonpath={.data.metrics_token} | base64 -d)`
`$ kubectl exec -it noobaa-core-0 -- curl -k -H "Authorization: Bearer ${JWT_TOKEN}" http://localhost:8080/metrics/bg_workers | grep replication_target_status`
5. Try deleting 1 destination bucket, and check the metrics (it will take time to update as replication cycle execute one rule in every `5 min~can_be_changed` by default) --> for that particular src and dst bucket flag should be unreachable/0
6. Now, try deleting replication on src bucket or delete the any of the source bucket, and check the metrics --> all the rows should be removed container that replication_id or src bucket
7. You can try different combinations and check if the metrics are reflecting correct values.

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of replication target status for orphaned policies when source buckets are missing.
  * More accurate reachability updates and failure reporting during replication scans.
  * Metric operations now safely no-op when metrics are unavailable.

* **Refactor**
  * Replaced bulk-reset with a targeted reconciliation flow that prunes stale replication metrics and validates against current policies.

* **Tests**
  * Added unit tests covering reconciliation, cleanup, deleted/dangling buckets, and edge cases in metric handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/noobaa/noobaa-core/pull/9623?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/server/analytic_services/prometheus_reports/noobaa_core_report.js
+++ b/src/server/analytic_services/prometheus_reports/noobaa_core_report.js
@@ -705,15 +705,22 @@ class NooBaaCoreReport extends BasePrometheusReport {
         this._metrics.bucket_last_cycle_error_objects_num.set({ bucket_name }, repl_info.bucket_last_cycle_error_objects_num);
     }
 
-    reset_replication_target_status() {
-        if (!this._metrics) return;
-        this._metrics.replication_target_status.reset();
-    }
-
     set_replication_target_status(replication_id, source_bucket, target_bucket, is_reachable) {
         if (!this._metrics) return;
         this._metrics.replication_target_status.set({ replication_id, source_bucket, target_bucket }, Number(is_reachable));
     }
+
+    clear_replication_target_status_by_replication_id(replication_id) {
+        const gauge_hash_map = this._metrics?.replication_target_status?.hashMap;
+        if (!gauge_hash_map) return;
+
+        const id = String(replication_id);
+        for (const key of Object.keys(gauge_hash_map)) {
+            const entry = gauge_hash_map[key];
+            if (entry?.labels?.replication_id === id) delete gauge_hash_map[key];
+        }
+    }
+
 }
 
 exports.NooBaaCoreReport = NooBaaCoreReport;
--- a/src/server/bg_services/replication_scanner.js
+++ b/src/server/bg_services/replication_scanner.js
@@ -60,8 +60,7 @@ class ReplicationScanner {
 
     async scan() {
         if (!this.noobaa_connection) throw new Error('noobaa endpoint connection is not started yet...');
-        // full reset of replication_target_status, then set reachability again for each policy for this scan cycle
-        replication_utils.reset_replication_target_status();
+        await replication_utils.reconcile_replication_target_status();
         // find rule for each replication policy that was not updated for the longest period
         const least_recently_replicated_rules = await replication_store.find_rules_updated_longest_time_ago();
 
@@ -72,12 +71,22 @@ class ReplicationScanner {
             const { src_bucket, dst_bucket } = replication_utils.find_src_and_dst_buckets(rule.destination_bucket, replication_id);
             if (!src_bucket || !dst_bucket) {
                 dbg.error('replication_scanner: can not find src_bucket or dst_bucket object', src_bucket, dst_bucket);
-                // mark target bucket as unreachable (resolve display name from id while bucket row still exists)
-                if (src_bucket && !dst_bucket) {
+                if (!src_bucket) {
+                    replication_utils.clear_replication_target_status_for_orphan_policy(replication_id);
+                    return;
+                }
+                if (!dst_bucket) {
                     const dst_bucket_name = await replication_utils.resolve_destination_bucket_name(rule.destination_bucket);
                     replication_utils.update_replication_target_status(replication_id, src_bucket.name, dst_bucket_name, false);
                     replication_utils.report_failed_replication_cycle(src_bucket.name, replication_id,
                         rule.rule_id, _.get(src_bucket, 'storage_stats.objects_count', 0));
+                    // advance last_cycle_end for rule rotation; keep cont tokens so replication resumes where it left off
+                    await replication_store.update_replication_status_by_id(replication_id, rule.rule_id, {
+                        ...status,
+                        last_cycle_end: Date.now(),
+                        src_cont_token: (rule.rule_status && rule.rule_status.src_cont_token) || '',
+                        dst_cont_token: (rule.rule_status && rule.rule_status.dst_cont_token) || '',
+                    });
                 }
                 return;
             }
@@ -112,10 +121,8 @@ class ReplicationScanner {
                 src_cont_token = buckets_diff_result.first_bucket_cont_token;
                 dst_cont_token = buckets_diff_result.second_bucket_cont_token;
 
-                // target bucket is reachable
                 replication_utils.update_replication_target_status(replication_id, src_bucket.name, dst_bucket.name, true);
             } catch (err) {
-                // target bucket is unreachable
                 dbg.error('replication_scanner: failed to get buckets diff, target may be unreachable:',
                     src_bucket.name, dst_bucket.name, err);
                 replication_utils.update_replication_target_status(replication_id, src_bucket.name, dst_bucket.name, false);
--- a/src/server/utils/replication_utils.js
+++ b/src/server/utils/replication_utils.js
@@ -5,6 +5,7 @@ const _ = require('lodash');
 const dbg = require('../../util/debug_module')(__filename);
 const SensitiveString = require('../../util/sensitive_string');
 const system_store = require('../system_services/system_store').get_instance();
+const replication_store = require('../system_services/replication_store').instance();
 const prom_reporting = require('../analytic_services/prometheus_reporting');
 const auth_server = require('../common_services/auth_server');
 
@@ -106,17 +107,70 @@ function report_failed_replication_cycle(bucket_name, replication_id, rule_id, t
         'replication_id:', replication_id, 'rule_id:', rule_id, 'total:', total);
 }
 
-function reset_replication_target_status() {
-    const core_report = prom_reporting.get_core_report();
-    core_report.reset_replication_target_status();
-}
-
 function update_replication_target_status(replication_id, source_bucket, target_bucket, is_reachable) {
     if (!replication_id) return;
     const core_report = prom_reporting.get_core_report();
     const src_name = source_bucket instanceof SensitiveString ? source_bucket.unwrap() : source_bucket;
     const dst_name = target_bucket instanceof SensitiveString ? target_bucket.unwrap() : target_bucket;
-    core_report.set_replication_target_status(replication_id, src_name, dst_name, is_reachable);
+    core_report.set_replication_target_status(String(replication_id), src_name, dst_name, is_reachable);
+}
+
+// clear target reachability metrics when the policy was deleted or has no source bucket (leftover series)
+// orphan policy: replication_id not in replication DB, or no source bucket references it
+function clear_replication_target_status_for_orphan_policy(replication_id) {
+    if (!replication_id) return;
+    const core_report = prom_reporting.get_core_report();
+    core_report.clear_replication_target_status_by_replication_id(String(replication_id));
+}
+
+// reconcile_replication_target_status reconciles target reachability metrics at replication scan start
+// drop stale replication_target_status series: one DB read, then clear gauge entries that no longer match active policies
+async function reconcile_replication_target_status() {
+    const hash_map = prom_reporting.get_core_report()._metrics?.replication_target_status?.hashMap;
+    if (!hash_map) return;
+
+    const keys = Object.keys(hash_map);
+    if (!keys.some(k => hash_map[k]?.labels?.replication_id)) return;
+
+    const replications = await replication_store.get_all_replication_configs();
+    if (!replications?.length) return;
+
+    const valid_label_keys = new Set();
+    const clear_all_ids = new Set();
+
+    // build allowed (replication_id|source|target) valid keys from DB and mark policies with no source for full clear
+    for (const repl of replications) {
+        const repl_id = String(repl._id);
+        if (repl.deleted) {
+            clear_all_ids.add(repl_id);
+            continue;
+        }
+        const rules = repl.rules;
+        if (!rules?.length) {
+            clear_all_ids.add(repl_id);
+            continue;
+        }
+        const { src_bucket } = find_src_and_dst_buckets(rules[0].destination_bucket, repl_id);
+        if (!src_bucket) {
+            clear_all_ids.add(repl_id);
+            continue;
+        }
+        const src_name = bucket_name_for_metrics(src_bucket);
+        // one valid series per rule destination for this policy
+        for (const rule of rules) {
+            const dst_name = await resolve_destination_bucket_name(rule.destination_bucket);
+            valid_label_keys.add(`${repl_id}|${src_name}|${dst_name}`);
+        }
+    }
+
+    // remove gauge samples that are orphaned, policy-less, or no longer match DB labels
+    for (const key of keys) {
+        const labels = hash_map[key]?.labels;
+        if (!labels?.replication_id) continue;
+        const repl_id = String(labels.replication_id);
+        const label_key = `${repl_id}|${labels.source_bucket}|${labels.target_bucket}`;
+        if (clear_all_ids.has(repl_id) || !valid_label_keys.has(label_key)) delete hash_map[key];
+    }
 }
 
 // remove the `-deleting-<timestamp>` suffix noobaa appends while a bucket is being deleted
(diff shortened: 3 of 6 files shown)


shirady:
@aayushchouhan09 can you update the PR description?
Can you add manual testing instructions? (in a comment or in the PR description)

src/server/analytic_services/prometheus_reports/noobaa_core_report.js
```diff
@@ -705,15 +705,20 @@ class NooBaaCoreReport extends BasePrometheusReport {
...
-        this._metrics.replication_target_status.reset();
+        this._metrics.replication_target_status.set({ replication_id, source_bucket, target_bucket }, Number(is_reachable));
     }
 
-    set_replication_target_status(replication_id, source_bucket, target_bucket, is_reachable) {
+    clear_replication_target_status_by_replication_id(replication_id) {
         if (!this._metrics) return;
-        this._metrics.replication_target_status.set({ replication_id, source_bucket, target_bucket }, Number(is_reachable));
+        const id = String(replication_id);
+        const gauge = this._metrics.replication_target_status;
```
shirady: What ensures that `gauge.hashMap` is not undefined?

src/server/analytic_services/prometheus_reports/noobaa_core_report.js
```diff
@@ -705,15 +705,20 @@ class NooBaaCoreReport extends BasePrometheusReport {
...
+        for (const key of Object.keys(gauge.hashMap)) {
+            if (gauge.hashMap[key]?.labels?.replication_id === id) delete gauge.hashMap[key];
+        }
```
shirady: Could you add a short comment about this loop?
If you think it is too obvious, then - no, just explain here (PR).

src/server/bg_services/replication_scanner.js
```diff
@@ -72,8 +71,11 @@ class ReplicationScanner {
             const { src_bucket, dst_bucket } = replication_utils.find_src_and_dst_buckets(rule.destination_bucket, replication_id);
             if (!src_bucket || !dst_bucket) {
                 dbg.error('replication_scanner: can not find src_bucket or dst_bucket object', src_bucket, dst_bucket);
-                // mark target bucket as unreachable (resolve display name from id while bucket row still exists)
-                if (src_bucket && !dst_bucket) {
+                if (!src_bucket) {
+                    replication_utils.clear_replication_target_status_for_orphan_policy(replication_id);
+                    return;
+                }
+                if (!dst_bucket) {
```
shirady, line 78: If we don't have `dst_bucket` don't we need to call `clear_replication_target_status_for_orphan_policy`? (like the condition `if (!src_bucket)` above).

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
+/* Copyright (C) 2016 NooBaa */
```
shirady: New file, you can update the year.

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
...
+'use strict';
+
+// bg_workers gauge: reconcile fixes stale series after policy delete, src removal,
+// rule change, id-vs-name labels, and -deleting-* names (web clear does not apply here).
```
shirady: Do you need this comment or is it redundant?

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
...
+        mock_store.deleted.clear();
+        mock_store.get_by_id = id => mock_store.buckets.find(doc => String(doc._id) === String(id));
+        mock_get_replication_by_id.mockReset();
+        mock_core.mockReturnValue(report);
+    });
+
+    it.each([
+        [
+            'removes all stale metrics when replication policy was deleted from DB (mark_deleted)',
+            () => mock_get_replication_by_id.mockResolvedValue(null),
```
shirady: Do you get `null` in real replication by ID?

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
...
+        report._metrics = null;
+        report.set_replication_target_status(R1, SRC, DST, 1);
+        report.clear_replication_target_status_by_replication_id(R1);
+        report._metrics = m;
+        expect(Object.keys(g().hashMap)).toHaveLength(0);
+    });
+
+    it.each([
+        [
+            'returns bucket name from soft-deleted row via get_by_id_include_deleted',
```
shirady: "soft-deleted row" I'm not sure I understand it

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
...
+            [DST],
+        ],
+    ])('reconcile at scan start: %s', async (_desc, setup, want) => {
+        probe(R1, SRC, DST);
+        setup();
+        await reconcile();
+        expect(tgts(R1)).toEqual(want);
+    });
+
+    it('reconcile prunes targets removed from policy and drops series where target_bucket is destination object id', async () => {
```
shirady: I'm not sure I understand this title

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
...
+        probe(R1, SRC, 'stale');
+        probe(R1, SRC, 'dst-id-a');
+        g().hashMap.bad = { labels: null, value: 0 };
+        seed(R1, [DST, 'dst-b']);
+        await reconcile();
+        expect(tgts(R1).sort()).toEqual([DST, 'dst-b'].sort());
+        expect(g().hashMap.bad).toBeDefined();
+    });
+
+    it('clear_replication_target_status_for_orphan_policy clears one policy only; null replication_id is ignored', () => {
```
shirady: As already asked, I'm not sure you would have a `null` value, only `undefined`.

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
...
+        ],
+    ])('reconcile early exit: %s', async (_desc, prep) => {
+        prep();
+        await reconcile();
+        expect(mock_get_replication_by_id).not.toHaveBeenCalled();
+    });
+
+    it('NooBaaCoreReport set_replication_target_status and clear do nothing when _metrics is null', () => {
+        const m = report._metrics;
+        report._metrics = null;
```
shirady, line 163: Is `report._metrics` `null` or just `undefined`?

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
...
+
+    it.each([
+        [
+            'returns bucket name from soft-deleted row via get_by_id_include_deleted',
+            () => mock_store.deleted.set('d', b({ _id: 'd', name: DST, deleting: new Date() })),
+            'd',
+            DST,
+        ],
+        [
+            'falls back to bucket id when live and deleted rows are missing',
```
shirady: Could you improve the title?
"falls back to bucket id when live and deleted rows are missing"
What is "when live"?

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
...
+            () => undefined,
+            'gone',
+            'gone',
+        ],
+        [
+            'falls back to bucket id when system_store lookup throws',
```
shirady: Could you improve the title?
```suggestion
            'falls back to bucket id when system_store lookup throws an error',
```
What is the context of the fallback?

src/test/unit_tests/internal/test_replication_metrics.test.js
```diff
@@ -0,0 +1,201 @@
...
+            'd',
+            '',
+        ],
+        [
+            'returns empty string when deleting bucket has empty name before suffix strip',
+            () => { mock_store.buckets = [b({ _id: 'd', name: '', deleting: new Date() })]; },
+            'd',
+            '',
+        ],
+        [
```
shirady: > returns empty string when bucket name unwraps to empty'
> 'returns empty string when deleting bucket has empty name before suffix strip'

Do you have cases where the bucket name is empty?
Is the mock directly on the DB? I'm not sure it can happen

src/server/utils/replication_utils.js
```diff
@@ -106,17 +107,63 @@ function report_failed_replication_cycle(bucket_name, replication_id, rule_id, t
...
 function update_replication_target_status(replication_id, source_bucket, target_bucket, is_reachable) {
     if (!replication_id) return;
     const core_report = prom_reporting.get_core_report();
     const src_name = source_bucket instanceof SensitiveString ? source_bucket.unwrap() : source_bucket;
     const dst_name = target_bucket instanceof SensitiveString ? target_bucket.unwrap() : target_bucket;
-    core_report.set_replication_target_status(replication_id, src_name, dst_name, is_reachable);
+    core_report.set_replication_target_status(String(replication_id), src_name, dst_name, is_reachable);
+}
+
+function clear_replication_target_status_for_orphan_policy(replication_id) {
```
shirady, line 120: Could you add a comment what " orphan_policy " exactly is?

src/server/utils/replication_utils.js
```diff
@@ -106,17 +107,63 @@ function report_failed_replication_cycle(bucket_name, replication_id, rule_id, t
...
+    if (!replication_id) return;
+    const core_report = prom_reporting.get_core_report();
+    core_report.clear_replication_target_status_by_replication_id(String(replication_id));
+}
+
+// bg_workers: prune stale series for replication_ids that already have metrics
```
shirady: Is it defined somewhere what "stale " is?

shirady (commented):
I tried to raise questions and added a few minor comments

src/server/analytic_services/prometheus_reports/noobaa_core_report.js
aayushchouhan09: We are deleting all the metrics entries here on the basis of `replication_id` in the argument.

src/server/bg_services/replication_scanner.js
aayushchouhan09: No, in case of `dst_bucket` is not found we are updating the flag to unreachable, we don't need to delete the entries. And with this entries we are firing the alert for target/dst bucket.

If there is no src bucket present or replication is deleted then the policy are of no use and we are removing there entries only.

src/server/utils/replication_utils.js
aayushchouhan09: Updated definition, Thanks

src/test/unit_tests/internal/test_replication_metrics.test.js
aayushchouhan09: yes it returns `null` or `undefined` if no rules found in DB

src/test/unit_tests/internal/test_replication_metrics.test.js
aayushchouhan09: NooBaaCoreReport sets this._metrics = null when PROMETHEUS_ENABLED is false.

Updated title, Thanks

src/test/unit_tests/internal/test_replication_metrics.test.js
aayushchouhan09: If a bucket is deleted from the cluster, its entry still present in the DB (for upto 3months) until the DB cleaner removes it. And the bucket name is suffixed with `-deleting-<timestamp>` in that case and these rows are being called as soft-deleted rows in DB.
We can still fetch the name of the deleted bucket from that row.

src/test/unit_tests/internal/test_replication_metrics.test.js
aayushchouhan09: When we are not able to fetch the bucket name even from the soft-deleted rows, we fallback and print the bucket_id instead of name in the alert.

src/test/unit_tests/internal/test_replication_metrics.test.js
aayushchouhan09: When we are not able to fetch the bucket name even from the soft-deleted rows, we fallback and print the bucket_id instead of name in the alert.

src/test/unit_tests/internal/test_replication_metrics.test.js
aayushchouhan09: We cannot have such cases, removing the two test cases. Thanks

src/server/analytic_services/prometheus_reports/noobaa_core_report.js
```diff
@@ -705,15 +705,22 @@ class NooBaaCoreReport extends BasePrometheusReport {
...
     set_replication_target_status(replication_id, source_bucket, target_bucket, is_reachable) {
         if (!this._metrics) return;
         this._metrics.replication_target_status.set({ replication_id, source_bucket, target_bucket }, Number(is_reachable));
     }
+
+    clear_replication_target_status_by_replication_id(replication_id) {
+        const gauge_hash_map = this._metrics?.replication_target_status?.hashMap;
+        if (!gauge_hash_map) return;
+
+        const id = String(replication_id);
```
shirady, line 717: Do you guard somewhere that `replication_id` is not `undefined`?

src/server/utils/replication_utils.js
```diff
@@ -106,17 +107,66 @@ function report_failed_replication_cycle(bucket_name, replication_id, rule_id, t
...
+    core_report.clear_replication_target_status_by_replication_id(String(replication_id));
+}
+
+// bg_workers: reconcile target reachability metrics at replication scan start
+// compare exported series to replication DB: clear orphans, remove source/target labels that no longer match
+async function reconcile_replication_target_status() {
+    const gauge = prom_reporting.get_core_report()._metrics?.replication_target_status;
+    if (!gauge?.hashMap) return;
+
+    // collect replication_ids that already have exported series
```
shirady: "exported series"?
I'm not I understand it, if you can define what it is in brackets.

src/test/unit_tests/internal/test_replication_metrics.test.js
shirady: Could you add an example to show it?

shirady (approved):
Approving not to block you - added final comments

src/test/unit_tests/internal/test_replication_metrics.test.js
aayushchouhan09: We have added a fix for it in this [PR](https://github.com/noobaa/noobaa-core/pull/9566/changes) where we resolving the bucket name, you can see the testing steps there.
This is the worst case scenario when soft-deleted rows are also deleted from the DB by the `DB_cleaner`

src/server/utils/replication_utils.js
```diff
@@ -106,17 +107,66 @@ function report_failed_replication_cycle(bucket_name, replication_id, rule_id, t
...
+        const src_name = bucket_name_for_metrics(src_bucket);
+
+        // build label set that should exist per current DB rules
+        const valid_keys = new Set();
+        for (const rule of rules) {
+            const dst_name = await resolve_destination_bucket_name(rule.destination_bucket);
+            valid_keys.add(`${repl_id}|${src_name}|${dst_name}`);
+        }
+        // remove (src, dst) series that no longer match DB rules for this replication_id
+        for (const key of Object.keys(gauge.hashMap)) {
```
naveenpaul1: going through complete metrics and removing one could be resourse consuming, If possible refactor this, But shouldnt be a blocker for PR Merge, LGTM

naveenpaul1 (approved):

src/server/analytic_services/prometheus_reports/noobaa_core_report.js
aayushchouhan09: We are guarding it from replication_utils caller functions for both update and clear.

src/server/utils/replication_utils.js
aayushchouhan09: I did the refactor and calling DB once to fetch all the rules and then clearing the metrics using one loop. Please notice here `find_src_and_dst_buckets()` is not quering the DB.
I think this makes it memory efficient. Thanks!

PR #9628, opened 2026-05-18
add nc_corestest to test_sts

sakshimunjal:
### Describe the Problem
test_sts.js was not part of the nc_coretest suite. NC mode has architectural differences (separate nsfs.js process, file-based config, no system_store) that required several fixes to make the STS tests run.

### Explain the Changes
1. Registered test_sts.js in the nc_coretest suite.
2. Added NC-specific account config (uid, gid, new_buckets_path) and policy adjustments across all test suites.
3. Fixed JWT secret mismatch between test process and nsfs.js by loading .env before any config imports.
4. Fixed CLI error propagation so test assertions can correctly match error codes from manage_nsfs.
5. Added null-guard for system_store in STS request handling to prevent server crashes in NC mode.
6. Added semantic validation for effect and action fields in role config validation.

1. 

### Testing Instructions:
1. For testing locally
`sudo -E NC_CORETEST=true node node_modules/.bin/mocha --timeout 60000 \
  src/test/utils/index/nc_index.js \
  --grep "STS tests|Session token tests|Assume role policy tests|Assume role with web indentity tests" \
  2>&1 | tee sts_test_output.log`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * LDAP reconnection behavior corrected and STS system-owner lookup made null-safe.
  * STS authorization bypass now correctly respects NC environment.

* **New Features**
  * Stricter role configuration validation: only allowed effects and a whitelisted set of STS actions accepted.

* **Tests**
  * Expanded STS tests for NC coretest mode (NC-specific setup, expectations, skips) and improved test CLI argument handling and coretest helpers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/noobaa/noobaa-core/pull/9628?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/endpoint/sts/sts_rest.js
+++ b/src/endpoint/sts/sts_rest.js
@@ -8,6 +8,7 @@ const js_utils = require('../../util/js_utils');
 const http_utils = require('../../util/http_utils');
 const signature_utils = require('../../util/signature_utils');
 const system_store = require('../../server/system_services/system_store').get_instance();
+const { is_nc_environment } = require('../../nc/nc_utils');
 
 const STS_MAX_BODY_LEN = 4 * 1024 * 1024;
 
@@ -137,8 +138,8 @@ async function authorize_request_policy(req) {
     const method = _get_method_from_req(req);
     const cur_account_email = req.sts_sdk.requesting_account && req.sts_sdk.requesting_account.email.unwrap();
     // system owner by design can always assume role policy of any account
-    if ((cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
-
+    // skip for NC environments since system owner is not applicable
+    if (!is_nc_environment() && (cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
     const permission = has_assume_role_permission(assume_role_policy, method, cur_account_email);
     dbg.log0('sts_rest.authorize_request_policy permission is: ', permission);
     if (permission === 'DENY' || permission === 'IMPLICIT_DENY') {
@@ -148,7 +149,8 @@ async function authorize_request_policy(req) {
 }
 
 function _get_system_owner() {
-    const system = system_store.data.systems[0];
+    const system = system_store.data && system_store.data.systems && system_store.data.systems[0];
+    if (!system) return null;
     return system.owner.email;
 }
 
--- a/src/manage_nsfs/manage_nsfs_validations.js
+++ b/src/manage_nsfs/manage_nsfs_validations.js
@@ -582,6 +582,15 @@ function validate_role_config(user_input) {
             throw_cli_error(ManageCLIError.InvalidRoleConfig,
                 'each statement in assume_role_policy must have "effect", "action" (array) and "principal" (array)');
         }
+        if (statement.effect !== 'allow' && statement.effect !== 'deny') {
+            throw_cli_error(ManageCLIError.InvalidRoleConfig, 'effect must be "allow" or "deny"');
+        }
+        const valid_actions = ['*', 'sts:AssumeRole', 'sts:AssumeRoleWithWebIdentity', 'sts:AssumeRoleWithSAML', 'sts:*'];
+        for (const action of statement.action) {
+            if (!valid_actions.includes(action)) {
+                throw_cli_error(ManageCLIError.InvalidRoleConfig, 'Policy has invalid action');
+            }
+        }
     }
 }
 
--- a/src/test/integration_tests/api/sts/test_sts.js
+++ b/src/test/integration_tests/api/sts/test_sts.js
@@ -2,10 +2,13 @@
 'use strict';
 
 // setup coretest first to prepare the env
-const coretest = require('../../../utils/coretest/coretest');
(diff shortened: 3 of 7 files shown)


src/endpoint/sts/sts_rest.js
```diff
@@ -137,7 +137,9 @@ async function authorize_request_policy(req) {
     const method = _get_method_from_req(req);
     const cur_account_email = req.sts_sdk.requesting_account && req.sts_sdk.requesting_account.email.unwrap();
     // system owner by design can always assume role policy of any account
-    if ((cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
+    // _get_system_owner returns null in NC mode (no system_store), so the bypass is skipped
+    const system_owner = _get_system_owner();
```
jackyalbo: just skip this part for nc_environment:

```suggestion
    if (!nc_utils.is_nc_environment() && (cur_account_email === _get_system_owner().unwrap()) &&  req.op_name.endsWith('assume_role')) return;
```
and update the comment above that we will skip this check for NC.

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -582,6 +582,16 @@ function validate_role_config(user_input) {
...
+        if (statement.effect !== 'allow' && statement.effect !== 'deny') {
+            throw_cli_error(ManageCLIError.InvalidRoleConfig, 'effect must be "allow" or "deny"');
+        }
+        const valid_actions = ['sts:AssumeRole', 'sts:AssumeRoleWithWebIdentity', 'sts:AssumeRoleWithSAML',
+            'sts:TagSession', 'sts:SetSourceIdentity', 'sts:*'];
+        for (const action of statement.action) {
+            if (!valid_actions.includes(action)) {
+                throw_cli_error(ManageCLIError.InvalidRoleConfig, 'Policy has invalid action');
+            }
+        }
```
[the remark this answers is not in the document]
jackyalbo: I agree with the rabbit - check if '*' is supported by AWS - if so, add it here - if not, remove it from _is_statement_fit.
Also, I don't think we support these as well: 'sts:TagSession', 'sts:SetSourceIdentity'
Where are they coming from?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -141,7 +158,8 @@ mocha.describe('STS tests', function() {
             Version: '2012-10-17',
             Statement: [{
                 Effect: 'Allow',
-                Principal: { AWS: [`arn:aws:iam::${account_info_a._id.toString()}:root`, `arn:aws:iam::${account_info_b._id.toString()}:root`,
+                Principal: { AWS: is_nc_coretest ? '*' : [`arn:aws:iam::${account_info_a._id.toString()}:root`, // existing gap in config_fs.is_account_exists_by_principal for NC
```
jackyalbo: what is the issue here?

sakshimunjal: config_fs.is_account_exists_by_principal explicitly returns false for ARN-format principals 
`if (principal_as_string.includes(arn_prefix)) {
        return false; // GAP
 }`
hence, added '*' as a workaround until ARN-based principal lookup is supported

jackyalbo: can't we do it with the account name?

sakshimunjal: yeah you're right, account names are more precise than *, updating

src/test/utils/coretest/nc_coretest.js
```diff
@@ -1,6 +1,14 @@
...
 'use strict';
 
+// dotenv should be loaded before config.js is loaded to set JWT_SECRET for nsfs.js
+process.env.DEBUG_MODE = 'true';
+// process.env.ACCOUNTS_CACHE_EXPIRY = '1'; In NC we check if the config file was changed as validation
+process.env.NC_CORETEST = 'true';
+
+require('../../../util/dotenv').load();
+require('../../../util/panic');
+require('../../../util/fips');
```
jackyalbo, line 11: Why was it suppose to move? The code was before loading config.js also before

sakshimunjal: JsonWebTokenError: invalid signature was thrown during session token verification.

**Root cause:** nsfs.js calls dotenv.load() before loading config.js, so config.JWT_SECRET was set from the .env file. However, the test process (nc_coretest.js) was requiring config.js before dotenv.load() ran — so process.env.JWT_SECRET was still unset at that point, causing config.JWT_SECRET to fall back to the default secret:

config.JWT_SECRET = process.env.JWT_SECRET || _get_data_from_file(`/etc/noobaa-server/jwt`);
This meant the server and the test process were signing/verifying tokens with different secrets.

Fix: Moved dotenv.load() to the very top of nc_coretest.js, before any require() that triggers config.js. This ensures process.env.JWT_SECRET is populated from .env when config.js evaluates the line above, so both sides use the same secret.

jackyalbo (approved):

PR #9641, opened 2026-05-21
STS | Azure error code check for connevction

naveenpaul1:
### Describe the Problem

Misleading INVALID_ENDPOINT error when Azure STS managed identity lacks permissions on target storage account

### Explain the Changes
1. When the error code or name is startswith() 'Authentication') or 'Authorization' then return INVALID_CREDENTIALS error.

1. https://redhat.atlassian.net/browse/DFBUGS-6948

### Testing Instructions:
1. Try to create backingstore with invalid client or tenant or
2. remove storage account or blob access for the managed identity and try to create azure sts backingstore

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of Azure connection failures so authentication and authorization errors are more accurately identified and reported, providing clearer feedback when credentials or permissions are the cause.
  * Preserved existing special-case handling for time skew and other unknown failures to keep current diagnostics intact.

[Review Change Stack](https://app.coderabbit.ai/change-stack/noobaa/noobaa-core/pull/9641?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -800,8 +800,14 @@ async function _check_azure_connection_internal(params) {
             throw err_to_status(err, 'TIME_SKEW');
         } else if (err.code === 'ENOTFOUND') {
             throw err_to_status(err, 'UNKNOWN_FAILURE');
+        } else if (err.name?.toLowerCase().includes('authentication') || err.code?.toLowerCase().includes('authorization')) {
+            /* 
+                For AuthenticationRequiredError err.code is undefined, err.name is AuthenticationRequiredError, 
+                In case of AuthorizationPermissionMismatch err.code is AuthorizationPermissionMismatch and err.name is RestError
+            */
+            throw err_to_status(err, 'INVALID_CREDENTIALS');
         } else {
-            throw err_to_status(err, err.code === 'AuthenticationFailed' ? 'INVALID_CREDENTIALS' : 'INVALID_ENDPOINT');
+            throw err_to_status(err, 'INVALID_ENDPOINT');
         }
     }
 


src/server/system_services/account_server.js
```diff
@@ -800,8 +800,10 @@ async function _check_azure_connection_internal(params) {
             throw err_to_status(err, 'TIME_SKEW');
         } else if (err.code === 'ENOTFOUND') {
             throw err_to_status(err, 'UNKNOWN_FAILURE');
+        } else if (err.name.startsWith('Authentication') || err.code.startsWith('Authorization')) {
```
shirady: Are you sure you want only the errors that starts with `Authentication` and `Authorization` prefix as case sensitive?

I looked at the list of errors ([link](https://learn.microsoft.com/en-us/rest/api/storageservices/common-rest-api-error-codes)) and I could see examples like `NoAuthenticationInformation` and `InvalidAuthenticationInfo` that include the string `Authentication` but not starts with and wanted to check it with you.

src/server/system_services/account_server.js
```diff
@@ -800,8 +800,10 @@ async function _check_azure_connection_internal(params) {
             throw err_to_status(err, 'TIME_SKEW');
         } else if (err.code === 'ENOTFOUND') {
             throw err_to_status(err, 'UNKNOWN_FAILURE');
+        } else if (err.name?.toLowerCase().includes('authentication') || err.code?.toLowerCase().includes('authorization')) {
```
shirady, line 803: What is the reason for checking the property of `name` and not `code`?
In the rest of the conditions it checks the `code`.
Additional question - Why `err.code?.`, is there is a case where you have an error `err` but not a property of `code`?

naveenpaul1: @shirady for `AuthenticationRequiredError` err.code is undefined, only err.name is  `AuthenticationRequiredError`, And in case of `AuthorizationPermissionMismatch` error err.code is `AuthorizationPermissionMismatch` and err.name is `RestError`.

shirady: Could you add a comment about it? So it would be clear for other developers

naveenpaul1: Sure, Will add

shirady (approved):
LGTM

PR #9642, opened 2026-05-21
proposal md file for ldap grouping role assumption

sakshimunjal:
### Describe the Problem
LDAP Group-Based Role Assumption — Design Proposal

### Explain the Changes
1. md file added for proposal

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Documentation**
  * Added a design proposal for LDAP-aware authorization of AssumeRoleWithWebIdentity: fetching LDAP attributes (OU, groups), evaluating role policy condition patterns (OU, group, combined AND/OR), supported condition operators, example patterns, a two-stage rollout plan for policy evaluation, and optional config-time validation to reject malformed condition blocks.

### DIFF
--- a/docs/ldap_group_assume_role_proposal.md
+++ b/docs/ldap_group_assume_role_proposal.md
@@ -1,0 +1,280 @@
+# LDAP Group-Based Role Assumption — Design Proposal
+
+## Background
+
+NooBaa supports `AssumeRoleWithWebIdentity` for LDAP users, where a user sends a JWT token containing their LDAP credentials to obtain temporary S3 credentials scoped to a role. Currently, any authenticated LDAP user can assume a role — there is no mechanism to restrict assumption based on LDAP group membership or any specific attribute.
+
+This proposal adds group-based access control to the `AssumeRoleWithWebIdentity` flow, allowing administrators to configure roles that only specific LDAP groups or departments can assume.
+
+---
+
+## Current Flow
+
+```text
+1. Client sends HTTP POST to STS:
+      Action=AssumeRoleWithWebIdentity
+      RoleArn=arn:aws:sts::<access_key>:role/<role_name>
+      WebIdentityToken=<JWT containing username + password>
+
+2. NooBaa extracts username and password from the JWT token
+
+3. NooBaa authenticates the user against LDAP
+      → receives the user's DN (Distinguished Name)
+
+4. NooBaa fetches the account and role_config from the role ARN
+
+5. Temporary credentials are issued to the caller
+```
+
+**Out of scope:** Step 5 always succeeds for any authenticated LDAP user. The `assume_role_policy` inside `role_config` is never evaluated for the `AssumeRoleWithWebIdentity` operation.
+
+---
+
+## Proposed Flow
+
+```text
+1. Client sends HTTP POST to STS:
+      Action=AssumeRoleWithWebIdentity
+      RoleArn=arn:aws:sts::<access_key>:role/<role_name>
+      WebIdentityToken=<JWT containing username + password>
+
+2. NooBaa extracts username and password from the JWT token
+
+3. NooBaa authenticates the user against LDAP
+      → receives the user's DN
+      → additionally fetches: ou, memberOf attributes  [NEW]
+
+4. NooBaa fetches the account + role_config (including assume_role_policy) from the role ARN
+
+5. NooBaa evaluates the assume_role_policy against the user's LDAP attributes  [NEW]
+      → True  : issue temporary credentials
+      → False : return Access Denied
+```
+
+---
+
+## Role Config Format
+
+A role is configured by setting `role_config` on a NooBaa account. The account's `access_key` forms the role ARN:
+
+```text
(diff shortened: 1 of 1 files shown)


docs/ldap_group_assume_role_proposal.md
```diff
@@ -0,0 +1,287 @@
...
+#### 1. `src/endpoint/sts/sts_rest.js`
+
+Add LDAP auth before policy, Web Identity only:
+
+```javascript
+async function authorize_request(req) {
+    await req.sts_sdk.load_requesting_account(req);
+    req.sts_sdk.authorize_request_account(req);
+
+    if (req.op_name === 'post_assume_role_with_web_identity') {
```
jackyalbo, line 235: I think we should keep this authentication as part of the authentication flow - let's not mix the authenticate phase with authorize one - we can call the web_identity_auth flow just right after the regular authenticate for AK/SK

docs/ldap_group_assume_role_proposal.md
```diff
@@ -0,0 +1,287 @@
...
+- Conditions within one `condition` block: **AND**
+- Multiple allow statements with conditions: **OR** (existing statement-matching behaviour)
+
+If a statement has no `condition` block, behaviour is unchanged (action + principal only).
+
+#### 2. `src/sdk/sts_sdk.js`
+
+Extract JWT verify + LDAP bind from `get_assumed_ldap_user` into `authenticate_web_identity(req)`:
+
+- Cache result on the SDK instance: `this.ldap_identity = { dn, ou, memberOf }`
```
jackyalbo: We can have here something similar to load_requesting_account, maybe load_identity_info which can be used outside of LDAP as well

jackyalbo: it will be cached as well - just like in requesting account

docs/ldap_group_assume_role_proposal.md
```diff
@@ -0,0 +1,287 @@
...
+In sts_rest.js, authorize_request(), LDAP user details are required before authorize_request_policy(). Hence, reordering is required ie for Web Identity LDAP auth will be moved to authorize phase.
+
+### Request pipeline (Stage 2)
+
+```
+AssumeRole (unchanged):
+  load_requesting_account → authorize_request_account → authorize_request_policy → handler
+
+AssumeRoleWithWebIdentity:
+  load_requesting_account → authorize_request_account → authenticate_web_identity → authorize_request_policy → handler
```
jackyalbo:
```suggestion
  authenticate_web_identity -> (see where load_identity_info fits in)  -> load_requesting_account(which will be empty in our case - anonymous) → authorize_request_account (also not relevant) → authorize_request_policy → handler
```

docs/ldap_group_assume_role_proposal.md
```diff
@@ -0,0 +1,287 @@
...
+            "ldap:ou": "Delivering Crew"
+          }
+        }
+      },
+      {
+        "effect": "allow",
+        "principal": ["*"],
+        "action": ["sts:AssumeRoleWithWebIdentity"],
+        "condition": {
+          "ForAnyValue:StringEquals": {
```
jackyalbo, line 174: Do we need ForAnyValue when there is only one value in the list?

docs/ldap_group_assume_role_proposal.md
```diff
@@ -0,0 +1,287 @@
...
+
+#### 3. `src/util/ldap_client.js`
+
+Extend `authenticate(user, password)` to fetch and return LDAP attributes needed for policy:
+
+returns `dn`, `ou` and `memberOf` (or any other required fields)
+
+#### 4. Optional — config-time validation
+
+Reject malformed `condition` blocks when a role is created/updated (fail fast at CLI/API time, not at first STS call)
```
jackyalbo: We should - let's change validate_role_config to be more like validate_bucket_policy.
it should take out the assume_role_policy and call _validate_policy(with needed changes of course)

docs/ldap_group_assume_role_proposal.md
jackyalbo: Well, it all depends if we need a second/another call to LDAP to get this info - if we will get all the needed info in the authentication flow, so we don't need to cache, as we will always do authentication first, but if we will need to ask LDAP for this info in another call, then we should split and cache.

docs/ldap_group_assume_role_proposal.md
sakshimunjal: Yes, since on LDAP side, memberOf is a list by design, so ForAnyValue will match if any of the user’s groups is in the allowed list

docs/ldap_group_assume_role_proposal.md
sakshimunjal: makes sense, will move this to authentication phase

docs/ldap_group_assume_role_proposal.md
sakshimunjal: will verify this, and check if cache is required

sakshimunjal:
appended in #9702

PR #9646, opened 2026-05-26
[Backport to 5.16] db_cleaner, md_store - don't use executeSQL (df 6821)

alphaprinz:
### Describe the Problem
Commit https://github.com/noobaa/noobaa-core/commit/ef2b2583b72eebc068aa3a73dea4827a5462f96f uses executeSQL(), but this new function was added in 22 and was not part of the cherry-picked backport.

Symptom is that db_cleaner fails to clean md_store.

### Explain the Changes
1. Switch executeSQL() to single_query().

1. [DFBUGS-7030](https://redhat.atlassian.net/browse/DFBUGS-7030)

### Testing Instructions:
1. Tested by removing db_cleaner limitations to clean md_store and then just running core pod.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -878,7 +878,7 @@ class MDStore {
         FROM ${this._objects.name}
         WHERE (to_ts(data->>'deleted')<to_ts($1) and data ? 'deleted' and data ? 'reclaimed') 
         LIMIT ${query_limit};`;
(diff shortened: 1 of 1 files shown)


dannyzaken (approved):

dannyzaken (changes requested):
just noticed that this is a direct PR to 5.16. I think that we should make the change to 5.21 and backport from there

nimrod-becker (approved):

PR #9656, opened 2026-05-27
vectors - prevent deletion of NSS if it used by a vector bucket (df 6947, take 2)

alphaprinz:
### Describe the Problem
NSS that is used by a vector bucket can be deleted.
Check that NSS is not used by a vector bucket before deletion.
If it used, don't delete.
This is similar behaviour to trying to delete NSS that is used by an object bucket.

This is a second fix for 6947.
(first was https://github.com/noobaa/noobaa-core/pull/9643)

### Explain the Changes
1. Check if NSS is being used by a vector bucket. If so, don't delete.

1. https://redhat.atlassian.net/browse/DFBUGS-6947

### Testing Instructions:
1. Create NSS.
2. Create vector bucket using above NSS.
3. Delete NSS, should fail.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved namespace deletion safety: deletion is now blocked when a namespace is referenced by any bucket type (including associated vector buckets), preventing accidental removal and providing clearer feedback (marked as in-use).

### DIFF
--- a/src/server/system_services/pool_server.js
+++ b/src/server/system_services/pool_server.js
@@ -1296,7 +1296,8 @@ function check_resource_pool_deletion(pool) {
 function check_namespace_resource_deletion(ns) {
     //Verify namespace resource is not used by any namespace bucket
     const buckets = get_associated_buckets_ns(ns);
-    if (buckets.length) {
+    const vector_buckets = get_associated_vector_buckets_ns(ns);
+    if (buckets.length || vector_buckets.length) {
         return 'IN_USE';
     }
 }
@@ -1309,6 +1310,14 @@ function get_associated_buckets_ns(ns) {
     });
 
     return _.map(associated_buckets, 'name');
+}
+
+function get_associated_vector_buckets_ns(ns) {
+    const associated_vector_buckets = _.filter(ns.system.vector_buckets_by_name, vector_bucket =>
+        (String(vector_bucket.namespace_resource?.resource?._id) === String(ns._id))
+    );
+
+    return _.map(associated_vector_buckets, 'name');
 }
 
 function _is_cloud_pool(pool) {


src/server/system_services/pool_server.js
```diff
@@ -1310,6 +1311,14 @@ function get_associated_buckets_ns(ns) {
     return _.map(associated_buckets, 'name');
 }
 
+function get_associated_vector_buckets_ns(ns) {
+    const associated_vector_buckets = _.filter(ns.system.vector_buckets_by_name, vector_bucket =>
```
shirady, line 1316: What about the string conversion (string) that is usually used in ID comparison? (appears in `get_associated_buckets_ns`)?
https://github.com/noobaa/noobaa-core/blob/68ca22d3308df43dd8e951381b62ec391330e2f2/src/server/system_services/pool_server.js#L1303-L1308
more info - [here](https://github.com/noobaa/noobaa-core/pull/9656#pullrequestreview-4375313932).

shirady (approved):
Approving not to block you - 
I added a comment

src/server/system_services/pool_server.js
alphaprinz: I don't see a reason it is needed. Am I missing sometihng?

shirady: In general, when comparing the IDs, which are `ObjectId` (when we had MongoDB), which means those are JS objects.
In JS, when using comparison (`===` in this case) on objects, it compares their references in memory, not their actual contents or values.

Anyway, it is not a must, as Code Rabbit wrote, but I thought it was clearer:

> vector_bucket_schema defines namespace_resource.resource as an objectid, and SystemStoreData runs resolve_object_ids_recursive() to replace those ObjectIds with the linked namespace_resource object. With that linking, vector_bucket.namespace_resource.resource._id should reference the same in-memory namespace resource as ns._id, so the === comparison is likely safe here (unlike comparing raw, unrelated ObjectId instances).

PR #9663, opened 2026-05-29
vectors - iam

alphaprinz:
### Describe the Problem
Implement IAM policy authorization for vectors rest.

### Explain the Changes
1. 

1. https://redhat.atlassian.net/browse/DFBUGS-6968

### Testing Instructions:
1. create iam user
s3i --no-verify-ssl create-user --user-name amit-sts --path /a/b/
s3i --no-verify-ssl create-access-key --user-name amit-sts
2. create policy
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "ListVectorBuckets",
      "Effect": "Deny",
      "Action": [
        "s3vectors:ListVectorBuckets"
      ],
      "Resource": "*"
    }
  ]
}
3.
s3i put-user-policy --user-name amit-sts --policy-name list --policy-document file:/policy.json
4. Execute s3vectors list-vector-buckets with iam user credentials.
5. In order to allow iam user to list vector buckets, change Deny to Allow in policy.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * IAM policy authorization now enforced for vector bucket operations, ensuring consistent access control across S3 and vector services. Enhanced handling of list-bucket permissions for both services.
* **Tests**
  * Added integration test coverage validating that IAM policy enforcement correctly prevents unauthorized vector index creation.

### DIFF
--- a/src/endpoint/iam/iam_utils.js
+++ b/src/endpoint/iam/iam_utils.js
@@ -105,7 +105,7 @@ function create_detailed_message_for_iam_user_access(user_account, method, resou
     const message_end = `because no identity-based policy allows the ${full_action_name} action`;
 
     let message_with_details;
-    if (full_action_name === 's3:ListAllMyBuckets' || full_action_name === 'vectors:ListVectorBuckets') {
+    if (full_action_name === 's3:ListAllMyBuckets' || full_action_name === 's3vectors:ListVectorBuckets') {
         message_with_details = message_start + message_end;
     } else {
         message_with_details = message_start + message_resource + message_end;
@@ -1058,7 +1058,6 @@ function get_owner_account_id(user_account) {
 }
 
 /**
- *
  * @param {object} req - http request
  * @param {String} bucket_name
  * @param {String} service - Either s3 or vectors
@@ -1075,21 +1074,20 @@ function _get_resource_arn_from_req(req, bucket_name, service) {
 }
 
 /**
- *
  * @param {object} req - http request
  * @param {Function} method - s3 method to authorize policy for
  * @param {String} bucket_name
- * @param {String} service - Either s3 or s3vectors
+ * @param {String} service - Either s3 or s3vectors, default is s3.
  * @returns true if request is authorized by policy, an object with account and resource_arn if not
  */
 async function authorize_request_iam_policy_impl(req, method, bucket_name, service = 's3') {
     const auth_token = req.object_sdk.get_auth_token();
     const is_anonymous = !(auth_token && auth_token.access_key);
-    if (is_anonymous) return true;
+    if (is_anonymous) return;
 
     const account = req.object_sdk.requesting_account;
     const is_iam_user = account.owner !== undefined;
-    if (!is_iam_user) return true; // IAM policy is only on IAM users (account root user is authorized here)
+    if (!is_iam_user) return; // IAM policy is only on IAM users (account root user is authorized here)
 
     const resource_arn = _get_resource_arn_from_req(req, bucket_name, service);
     const deny_result = {
@@ -1151,10 +1149,6 @@ exports.validate_tag_user_params = validate_tag_user_params;
 exports.validate_untag_user_params = validate_untag_user_params;
 exports.validate_list_user_tags_params = validate_list_user_tags_params;
 exports.get_owner_account_id = get_owner_account_id;
-<<<<<<< HEAD
 exports.throw_malformed_policy_document_error = throw_malformed_policy_document_error;
-exports.create_detailed_message_for_iam_user_access_in_s3 = create_detailed_message_for_iam_user_access_in_s3;
-=======
 exports.create_detailed_message_for_iam_user_access = create_detailed_message_for_iam_user_access;
->>>>>>> 01b99fdbd (vectors - iam - pr notes)
 exports.authorize_request_iam_policy_impl = authorize_request_iam_policy_impl;
--- a/src/endpoint/iam/iam_utils.js
+++ b/src/endpoint/iam/iam_utils.js
@@ -9,6 +9,8 @@ const { AWS_IAM_PATH_REGEXP, AWS_IAM_LIST_MARKER, AWS_IAM_ACCESS_KEY_INPUT_REGEX
 const iam_constants = require('./iam_constants');
 const { RpcError } = require('../../rpc');
 const validation_utils = require('../../util/validation_utils');
+const dbg = require('../../util/debug_module')(__filename);
+const access_policy_utils = require('../../util/access_policy_utils');
 
 /**
  * format_iam_xml_date return the date without milliseconds
@@ -85,14 +87,14 @@ function check_iam_path_was_set(iam_path) {
 }
 
 /**
- * _create_detailed_message_for_iam_user_access_in_s3 returns a detailed message with details needed for user who
+ * create_detailed_message_for_iam_user_access_in_s3 returns a detailed message with details needed for user who
  * tried to perform S3 operation
  *  - resource_arn is only relevant for operations related to a bucket
  * @param {object} user_account
  * @param {string|string[]} method
  * @param {string} resource_arn
  */
-function _create_detailed_message_for_iam_user_access_in_s3(user_account, method, resource_arn) {
+function create_detailed_message_for_iam_user_access_in_s3(user_account, method, resource_arn) {
     const owner_account_id = get_owner_account_id(user_account);
     const arn_for_requesting_account = create_arn_for_user(owner_account_id,
         user_account.name.unwrap(), user_account.iam_path);
@@ -1055,6 +1057,65 @@ function get_owner_account_id(user_account) {
     return owner_account_id;
 }
 
+function _get_arn_from_req_path(req, bucket_name) {
+    if (!bucket_name) return "*"; // special case for list all buckets in an account
+    const key = req.params.key;
+    let arn_path = `arn:aws:s3:::${bucket_name}`;
+    if (key) {
+        arn_path += `/${key}`;
+    }
+    return arn_path;
+}
+
+async function authorize_request_iam_policy_impl(req, method, bucket_name) {
+    const auth_token = req.object_sdk.get_auth_token();
+    const is_anonymous = !(auth_token && auth_token.access_key);
+    if (is_anonymous) return true;
+
+    const account = req.object_sdk.requesting_account;
+    const is_iam_user = account.owner !== undefined;
+    if (!is_iam_user) return true; // IAM policy is only on IAM users (account root user is authorized here)
+
+    const resource_arn = _get_arn_from_req_path(req, bucket_name);
+    const deny_result = {
+        account,
+        resource_arn
+    };
+    const iam_policies = account.iam_user_policies || [];
+    if (iam_policies.length === 0) {
+        if (req.object_sdk.nsfs_config_root) return true; // We do not have IAM policies in NC yet
+        dbg.error('authorize_request_iam_policy: IAM user has no inline policies configured');
+        //_throw_iam_access_denied_error_for_s3_operation(account, method, resource_arn);
+        return deny_result;
+    }
+
+    // parallel policy check
+    const promises = [];
+    for (const iam_policy of iam_policies) {
+        const promise = access_policy_utils.has_access_policy_permission(
+            iam_policy.policy_document, undefined, method, resource_arn, req,
+            { should_pass_principal: false }
+        );
+        promises.push(promise);
+    }
+    const permission_result = await Promise.all(promises);
+    let has_allow_permission = false;
+    for (const permission of permission_result) {
+        if (permission === "DENY") {
+            dbg.error('authorize_request_iam_policy: user has explicit DENY inline policy');
+            //_throw_iam_access_denied_error_for_s3_operation(account, method, resource_arn);
+            return deny_result;
+        }
+        if (permission === "ALLOW") {
+            has_allow_permission = true;
+        }
+    }
+    if (has_allow_permission) return true;
+    dbg.error('authorize_request_iam_policy: user has inline policies but none of them matched the method');
+    //_throw_iam_access_denied_error_for_s3_operation(account, method, resource_arn);
+    return deny_result;
+}
+
 // EXPORTS
 exports.format_iam_xml_date = format_iam_xml_date;
 exports.create_arn_for_user = create_arn_for_user;
@@ -1077,5 +1138,6 @@ exports.validate_tag_user_params = validate_tag_user_params;
 exports.validate_untag_user_params = validate_untag_user_params;
 exports.validate_list_user_tags_params = validate_list_user_tags_params;
 exports.get_owner_account_id = get_owner_account_id;
-exports._create_detailed_message_for_iam_user_access_in_s3 = _create_detailed_message_for_iam_user_access_in_s3;
 exports.throw_malformed_policy_document_error = throw_malformed_policy_document_error;
+exports.create_detailed_message_for_iam_user_access_in_s3 = create_detailed_message_for_iam_user_access_in_s3;
+exports.authorize_request_iam_policy_impl = authorize_request_iam_policy_impl;
--- a/src/endpoint/iam/iam_utils.js
+++ b/src/endpoint/iam/iam_utils.js
@@ -1059,7 +1059,7 @@ function get_owner_account_id(user_account) {
 
 function _get_arn_from_req_path(req, bucket_name) {
     if (!bucket_name) return "*"; // special case for list all buckets in an account
-    const key = req.params.key;
+    const key = req.params?.key;
     let arn_path = `arn:aws:s3:::${bucket_name}`;
     if (key) {
         arn_path += `/${key}`;
--- a/src/endpoint/iam/iam_utils.js
+++ b/src/endpoint/iam/iam_utils.js
@@ -87,14 +87,14 @@ function check_iam_path_was_set(iam_path) {
 }
 
 /**
- * create_detailed_message_for_iam_user_access_in_s3 returns a detailed message with details needed for user who
+ * create_detailed_message_for_iam_user_access returns a detailed message with details needed for user who
  * tried to perform S3 operation
  *  - resource_arn is only relevant for operations related to a bucket
  * @param {object} user_account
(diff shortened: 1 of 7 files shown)


src/endpoint/iam/iam_utils.js
```diff
@@ -890,6 +892,65 @@ function get_owner_account_id(user_account) {
     return owner_account_id;
 }
 
+function _get_arn_from_req_path(req, bucket_name) {
```
shirady: In this file, we tried to include JSDoc for every function; could you add it to this one as well?

Why did you add the additional argument `bucket_name`, as you had it in `req.params.bucket`?

src/endpoint/iam/iam_utils.js
```diff
@@ -890,6 +892,65 @@ function get_owner_account_id(user_account) {
...
+    if (!bucket_name) return "*"; // special case for list all buckets in an account
+    const key = req.params?.key;
+    let arn_path = `arn:aws:s3:::${bucket_name}`;
```
shirady: I saw it in the comment below - [link](https://github.com/noobaa/noobaa-core/pull/9663/changes#r3364120079).
As the bucket can be S3 (`arn:aws:s3:::`) or vector (`arn:aws:s3vectors:::`), right?

src/endpoint/iam/iam_utils.js
```diff
@@ -890,6 +892,65 @@ function get_owner_account_id(user_account) {
...
+    if (key) {
+        arn_path += `/${key}`;
+    }
+    return arn_path;
+}
+
+async function authorize_request_iam_policy_impl(req, method, bucket_name) {
+    const auth_token = req.object_sdk.get_auth_token();
+    const is_anonymous = !(auth_token && auth_token.access_key);
+    if (is_anonymous) return true;
```
shirady: I see that you check the `authorize_result`, for the cases that it is irrelevant, I think that we need another result (not just `true`) for the human reader.

src/endpoint/iam/iam_utils.js
```diff
@@ -890,6 +892,65 @@ function get_owner_account_id(user_account) {
     return owner_account_id;
 }
 
+function _get_arn_from_req_path(req, bucket_name) {
+    if (!bucket_name) return "*"; // special case for list all buckets in an account
```
shirady, line 1067: If this function is widely used, it might create a bug, so better return and use the `|| *` where you need it.

src/endpoint/iam/iam_utils.js
```diff
@@ -890,6 +892,65 @@ function get_owner_account_id(user_account) {
...
+    if (!is_iam_user) return true; // IAM policy is only on IAM users (account root user is authorized here)
+
+    const resource_arn = _get_arn_from_req_path(req, bucket_name);
+    const deny_result = {
+        account,
+        resource_arn
+    };
+    const iam_policies = account.iam_user_policies || [];
+    if (iam_policies.length === 0) {
+        if (req.object_sdk.nsfs_config_root) return true; // We do not have IAM policies in NC yet
```
shirady, line 1100: You can move it to the area of not relevant cases.

src/endpoint/iam/iam_utils.js
```diff
@@ -890,6 +892,65 @@ function get_owner_account_id(user_account) {
...
+            { should_pass_principal: false }
+        );
+        promises.push(promise);
+    }
+    const permission_result = await Promise.all(promises);
+    let has_allow_permission = false;
+    for (const permission of permission_result) {
+        if (permission === "DENY") {
+            dbg.error('authorize_request_iam_policy: user has explicit DENY inline policy');
+            //_throw_iam_access_denied_error_for_s3_operation(account, method, resource_arn);
```
shirady: Do you need this comment?

src/endpoint/iam/iam_utils.js
```diff
@@ -63,14 +65,14 @@ function check_iam_path_was_set(iam_path) {
 }
 
 /**
- * _create_detailed_message_for_iam_user_access_in_s3 returns a detailed message with details needed for user who
+ * create_detailed_message_for_iam_user_access_in_s3 returns a detailed message with details needed for user who
```
shirady: You're using it in Vectors as well, right?
So you can probably change the name so it would be suitable for this case as well.

src/endpoint/iam/iam_utils.js
shirady: You're using it in Vectors as well, right?
So your `full_action_name` would not be `s3:ListAllMyBuckets` as it starts with the service name S3.

src/endpoint/vector/vector_rest.js
```diff
@@ -8,7 +8,9 @@ const js_utils = require('../../util/js_utils');
 const http_utils = require('../../util/http_utils');
 const signature_utils = require('../../util/signature_utils');
 const access_policy_utils = require('../../util/access_policy_utils');
-const { get_owner_account_id } = require('../iam/iam_utils');
+const { create_detailed_message_for_iam_user_access_in_s3,
```
shirady: As mentioned before I think we need to change the function name - it has the `s3` service name, but it is imported to the `vector_rest`.

src/endpoint/vector/vector_rest.js
```diff
@@ -200,13 +202,35 @@ async function handle_request(req, res) {
...
     await authorize_request_vector_policy(req);
     const reply = await op.handler.handler(req, res);
     dbg.log0("VECTOR reply =", reply);
 
     http_utils.send_reply(req, res, reply, options);
 }
 
+async function authorize_request_iam_policy(req) {
+
+    const method = access_policy_utils.VECTOR_OP_NAME_TO_ACTION[req.op_name];
```
shirady, line 215: If you want, you can use it like in s3_rest `_get_method_from_req`.

src/endpoint/vector/vector_rest.js
```diff
@@ -200,13 +202,35 @@ async function handle_request(req, res) {
...
+    const bucket_name = req.body?.vectorBucketName;
+
+    const authorize_result = await authorize_request_iam_policy_impl(req, method, bucket_name);
+
+    if (authorize_result === true) return;
+    _throw_iam_access_denied_error_for_vector_operation(authorize_result.account, method, authorize_result.resource_arn);
+}
+
+function _throw_iam_access_denied_error_for_vector_operation(requesting_account, method, resource_arn) {
+    const message_with_details = create_detailed_message_for_iam_user_access_in_s3(requesting_account, method, resource_arn);
```
shirady: Should the method include the `s3vectors` prefix in its name?
Could you share what you checked in AWS?

src/sdk/bucketspace_fs.js
```diff
@@ -1500,7 +1500,8 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             const { vector_bucket_name, policy } = params;
             dbg.log0('BucketSpaceFS.put_vector_bucket_policy:', vector_bucket_name, policy);
             const vector_bucket = await this.config_fs.get_vector_bucket_by_name(vector_bucket_name);
-            await access_policy_utils.validate_vector_bucket_policy(policy, vector_bucket_name, []);
+            await access_policy_utils.validate_vector_bucket_policy(policy, vector_bucket_name, async principal =>
```
shirady: This is related to bucket policy, not to IAM policy, right?
Maybe you would want it in a separate PR or a separate commit and backport it?
In case it will be in this PR, could you add it to the PR description?

src/test/integration_tests/api/vectors/test_vectors_ops.js
```diff
@@ -1674,6 +1674,133 @@ mocha.describe('vectors_ops', function() {
             await send(s3_vectors_client_no_header, command);
         });
 
+        mocha.it('should enforce IAM policy to prevent vector index creation', async function() {
+
+            if (is_nc_coretest) { //No IAM for NC
```
shirady: We have IAM in NC, but we do not have IAM user inline policies.

```suggestion
            if (is_nc_coretest) { // // We do not have IAM policies in NC yet
```

src/test/integration_tests/api/vectors/test_vectors_ops.js
```diff
@@ -1674,6 +1674,133 @@ mocha.describe('vectors_ops', function() {
...
+                return;
+            }
+
+            // Import IAM commands
+            const { CreateUserCommand, CreateAccessKeyCommand, DeleteUserCommand, DeleteAccessKeyCommand } = require('@aws-sdk/client-iam');
+            const { generate_iam_client } = require('../../../system_tests/test_utils');
```
shirady: Why here?
Could you move it to the beginning of the file?

src/test/integration_tests/api/vectors/test_vectors_ops.js
```diff
@@ -1674,6 +1674,133 @@ mocha.describe('vectors_ops', function() {
...
+                UserName: iam_username
+            };
+            const create_access_key_command = new CreateAccessKeyCommand(create_access_key_input);
+            const access_key_response = await iam_client.send(create_access_key_command);
+            assert.strictEqual(access_key_response.$metadata.httpStatusCode, 200);
+            const iam_access_key = access_key_response.AccessKey.AccessKeyId;
+            const iam_secret_key = access_key_response.AccessKey.SecretAccessKey;
+
+            // Create S3 vectors client for IAM user
+            const iam_client_params = {
```
shirady: It is for the vector client, not the IAM client, right?

```suggestion
            const vector_client_params = {
```
We already have a function for creating a client if you need something else you can create it here:
https://github.com/noobaa/noobaa-core/blob/dbd08019fbcbd2723213738c86512a8fc45a703d/src/test/system_tests/test_utils.js#L464-L490

src/test/integration_tests/api/vectors/test_vectors_ops.js
```diff
@@ -1674,6 +1674,133 @@ mocha.describe('vectors_ops', function() {
...
+                },
+                region: config.DEFAULT_REGION,
+                requestHandler: new NodeHttpHandler({
+                    httpsAgent: new https.Agent({ rejectUnauthorized: false })
+                }),
+            };
+
+            const iam_s3_vectors_client = new s3vectors.S3VectorsClient(iam_client_params);
+
+            // Add custom namespace header for IAM client
```
shirady: It is not the IAM client anymore; it was only for creating the user.

src/test/integration_tests/api/vectors/test_vectors_ops.js
```diff
@@ -1674,6 +1674,133 @@ mocha.describe('vectors_ops', function() {
...
+                        request.headers[config.VECTORS_NSR_HEADER] = nsr;
+                    }
+                    return await next(args);
+                },
+                {
+                    step: 'build',
+                    name: 'noobaa_vector_headers',
+                    priority: 'high',
+                }
+            );
```
shirady: > Add custom namespace header


What is this? Why do we need it?
I see it was also used in:
https://github.com/noobaa/noobaa-core/blob/dbd08019fbcbd2723213738c86512a8fc45a703d/src/test/integration_tests/api/vectors/test_vector_bucket_policy.js#L85-L98

Could you move it to somewhere we can reuse it instead of writing it in every test file?

src/test/integration_tests/api/vectors/test_vectors_ops.js
```diff
@@ -1674,6 +1674,133 @@ mocha.describe('vectors_ops', function() {
...
+                }],
+            };
+
+            const put_policy_command = new s3vectors.PutVectorBucketPolicyCommand({
+                vectorBucketName: test_bucket_name,
+                policy: JSON.stringify(deny_policy),
+            });
+            await send(s3_vectors_client, put_policy_command);
+
+            // Attempt to create vector index with IAM user - should fail
```
shirady: I think that this test is not enough and we also need allow policy.
Since if we have IAM policy on one action, the rest of the actions are implicit DENY, so if you want to test the code you need also the allow policy case.

I think, as mentioned, there is a bug in the resource - so as part of the test we would need to fix it.
The policy resource prefix is `arn:aws:s3vectors:::` while in the function it is `arn:aws:s3:::`.

shirady:
@alphaprinz could you add to the testing instructions an allow policy for a user? (not just deny)

src/endpoint/iam/iam_utils.js
alphaprinz: Sorry, not sure what you mean. What would that include?
The previous function also returned true.

src/endpoint/iam/iam_utils.js
shirady: JSDoc function name doesn't match: `create_detailed_message_for_iam_user_access` vs `create_detailed_message_for_iam_user_access_in_s3`.

src/endpoint/iam/iam_utils.js
alphaprinz: I think it's better that way. More explicit. The other way has the risk of forgetting to manually add the default.

src/endpoint/iam/iam_utils.js
```diff
@@ -890,6 +892,78 @@ function get_owner_account_id(user_account) {
...
 }
 
+/**
+ *
+ * @param {object} req - http request
+ * @param {String} bucket_name
+ * @param {String} service - Either s3 or vectors
+ * @returns resource arn for requested resource - either * if no bucket, with bucket, or with bucket and key
+ */
+function _get_arn_from_req_path(req, bucket_name, service) {
```
shirady: Since it was moved to the util function, we can improve it a bit: I'm not sure about the `path` in the function... and your name (`resource_arn`) was clear.
- Rename the function: `_get_arn_from_req_path` -> `_get_resource_arn_from_req`
- Rename the variable: `arn_path` -> `resource_arn`

src/endpoint/iam/iam_utils.js
```diff
@@ -890,6 +892,78 @@ function get_owner_account_id(user_account) {
...
+    if (key) {
+        arn_path += `/${key}`;
+    }
+    return arn_path;
+}
+
+/**
+ *
+ * @param {object} req - http request
+ * @param {*} method - s3 method to authorize policy for
```
shirady: Only s3 method? also vectors, right?

src/endpoint/iam/iam_utils.js
shirady: It returned `undefined` before.

src/endpoint/iam/iam_utils.js
```diff
@@ -890,6 +892,78 @@ function get_owner_account_id(user_account) {
...
+ * @param {*} bucket_name
+ * @param {*} service - Either s3 or vectors
```
shirady: You would probably want to guard (from the developer) the case where the service argument was forgotten to be passed somehow.

shirady: Why `@param {*}` in all cases? I think you know the types, for example: `bucket_name` is a string or undefined.

src/test/integration_tests/api/vectors/test_vectors_ops.js
```diff
@@ -1674,6 +1682,129 @@ mocha.describe('vectors_ops', function() {
...
         });
 
+        mocha.it('should enforce IAM policy to prevent vector index creation', async function() {
+
+            if (is_nc_coretest) { // We do not have IAM policies in NC yet
+                return;
+            }
+
+            // Create IAM client using admin credentials
+            const iam_endpoint = coretest.get_https_address_iam();//.replace('/s3', '');
```
shirady: Why do we need the comment?

> //.replace('/s3', '');

src/test/integration_tests/api/vectors/test_vectors_ops.js
```diff
@@ -1674,6 +1682,129 @@ mocha.describe('vectors_ops', function() {
...
+                        request.headers[config.VECTORS_NSR_HEADER] = nsr;
+                    }
+                    return await next(args);
+                },
+                {
+                    step: 'build',
+                    name: 'noobaa_vector_headers',
+                    priority: 'high',
+                }
+            );
```
shirady: I think that you would want to reuse the functions that you already have and that you also created.

src/endpoint/iam/iam_utils.js
alphaprinz: This is the areaof non-relevant cases. What line did you think? Everything above is non-relevant cases handling, and I need to the deny_result object for the error message.

shirady: Never mind, leave it as is for now, thanks.

src/endpoint/iam/iam_utils.js
alphaprinz: Ok.

For deny case, I had to return some extra information for the detailed message. The allow case does not need that. This is the current behavior. I don't understand why this is bad or what is your suggestion.

Maybe adding some comments clarifying the different cases would help?

src/test/integration_tests/api/vectors/test_vectors_ops.js
alphaprinz: I meant it's for the iam user. I will change the name.

src/endpoint/iam/iam_utils.js
alphaprinz: Yes.

src/endpoint/iam/iam_utils.js
shirady: My suggestion:
- return true on allow
- return `deny_result` on deny
(Until now this is what you have in the code)
- return something else (`undefined`?) when it's not relevant, we didn't check the policy.

src/test/integration_tests/api/vectors/test_vectors_ops.js
alphaprinz: Added tests.

src/endpoint/vector/vector_rest.js
alphaprinz: An error occurred (AccessDeniedException) when calling the ListVectorBuckets operation: User: arn:aws:iam::255735097469:user/amit_test is not authorized to perform: s3vectors:ListVectorBuckets on resource: arn:aws:s3vectors:us-east-1:255735097469:* because no identity-based policy allows the s3vectors:ListVectorBuckets action

shirady (approved):
LGTM

PR #9644, opened 2026-05-26
GCP WIF (STS) | Updated Enums and  Consts and Create GCP Client

shirady:
### Describe the Problem
As part of [RHSTOR-8661](https://redhat.atlassian.net/browse/RHSTOR-8661), this PR sets the new endpoint type `GOOGLE_STS` (we had `GOOGLE`) and starts implementing code changes in parts of the flows related to the GCP WIF (STS) backing store.

### Explain the Changes
1. Add `GOOGLE_STS` as const/enum in all the places where we had `GOOGLE`.
2. In `cloud_utils.js` add helper functions for mutual code for creating the GCP client for types of `GOOGLE` and `GOOGLE_STS` (like we have for Azure).
Update: I added validation to the JSON after a comment from CodeRabbit - [link](https://github.com/noobaa/noobaa-core/pull/9644#discussion_r3303699417).
4. In `account_server` add the changes needed for the type `GOOGLE_STS` and add logs in level `warn` with the error (to align with Azure log level).
5. Add unit tests for the new functions in `cloud_utils.js` with the Jest framework (generated by AI).
6. Added a timeout for the list buckets on GCP (`GOOGLE` and `GOOGLE_STS`) after a comment from CodeRabbit - [link1](https://github.com/noobaa/noobaa-core/pull/9644#discussion_r3304836574), [link2](https://github.com/noobaa/noobaa-core/pull/9644#discussion_r3304836585) (reusing constants that were defined in the files). 

Note: in the code comments, I mentioned "AIP-4117 External Account Credentials (Workload Identity Federation)" - attaching the [link](https://google.aip.dev/auth/4117).

### Issues:
List of GAPs:
1. All the files that are related to GCP WIF have not been changed yet:
- `object_sdk.js` - this part was not updated yet
https://github.com/noobaa/noobaa-core/blob/06da0fae5cc9f66c9f30fe2bd2e5c62f918b87fc/src/sdk/object_sdk.js#L563-L574
- `namespace_monitor` did update the new type yet, currently added a comment in this PR
 https://github.com/noobaa/noobaa-core/blob/06da0fae5cc9f66c9f30fe2bd2e5c62f918b87fc/src/server/bg_services/namespace_monitor.js#L70-L71
2. Backingstore changes were not tested yet, and would need additional PR in the operator, and then can be tested, meaning these files: `agent.js`, `block_store_client.js`, `block_store_google.js`.
[...]

PR #9664, opened 2026-05-31
General | Fix `timeout` Function in `src/util/promise.js` 

shirady:
### Describe the Problem
While I was testing a case as part of #9644 I saw in the logs `UnhandledPromiseRejectionWarning` and used AI tools for solving and testing it.

### Explain the Changes
1. Fix a bug in the function `timeout` that caused `UnhandledPromiseRejectionWarning` since it used 2 listeners on the same promise (listener 1 `.then(success_only)`, listener 2 `.catch(failure_only)`) - and the bug appears when the inner promise (in my case, list buckets) fails. The change is to use a single `.then(success, failure)`.

### Issues: 
1. None, I found it during testing.

### Unit tests
1.  Please run: `npx jest src/test/unit_tests/util_functions_tests/test_promise_utils.test.js`

- [ ] Doc added/updated
- [X] Tests added

## Release Notes

* **Tests**
  * Added comprehensive unit test suite for promise timeout functionality, including tests for undefined parameters, missing promises, resolution before timeout, timeout behavior with default and custom errors, and edge cases.

* **Refactor**
  * Restructured internal promise timeout handling implementation to consolidate success and error handling logic.

### DIFF
--- a/src/test/unit_tests/util_functions_tests/test_promise_utils.test.js
+++ b/src/test/unit_tests/util_functions_tests/test_promise_utils.test.js
@@ -1,0 +1,108 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const P = require('../../../util/promise');
+const Defer = require('../../../util/defer');
+
+describe('P.timeout', () => {
+
+    it('returns the promise as-is when millis is undefined', async () => {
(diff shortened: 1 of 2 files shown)


liranmauda (approved):
LGTM

PR #9669, opened 2026-06-02
fix null timing in nsfs op metrics

sakshimunjal:
### Describe the Problem
https://redhat.atlassian.net/browse/DFBUGS-7012
Some of the metrics values were coming as null (in_time_milisec, max_time_milisec, avg_time_milisec)
Reason: In some error batches (no success), the min_time, max_time and sum_time are coming as undefined 
Hence, in update_nsfs_stats Math.min/Math.max with undefined produce NaN 

### Explain the Changes
1. Skip min/max updates when no success ops
2. Handle undefined values
3. Compute average over successful ops only (count - error_count) on first batch init and when deriving old_sum_time from previous stats.

### Issues: Fixed #DFBUGS-7012 / Gap #xxx
1. Fixed null NSFS op timing metrics when error-only stat batches are merged after successful operations.

### Testing Instructions:
1. Start NSFS 
`node --unhandled-rejections=warn src/cmd/nsfs.js --config_root $CONFIG_ROOT --debug=5`
2. Create 2 buckets successfully.
`aws s3 mb s3://mtest-a --endpoint-url 'https://[::1]:6443' --no-verify-ssl`
`aws s3 mb s3://mtest-b --endpoint-url 'https://[::1]:6443' --no-verify-ssl`
3. Wait ≥12s (stats flush interval).
4. Attempt to create one of the same buckets again (expect error).
`aws s3 mb s3://mtest-a --endpoint-url 'https://[::1]:6443' --no-verify-ssl`
5.Wait ≥12s, then scrape metrics:
`curl -sk 'https://<host>:9443/metrics/nsfs_stats' | jq '.op_stats_counters | with_entries(select(.key | startswith("noobaa_nsfs_op_create_bucket")))'
`

- [ ] Doc added/updated
- [x] Tests added
Specific file test run:
`npx jest src/test/unit_tests/util_functions_tests/test_stats_collector_utils.test.js`

## Summary by CodeRabbit

* **Bug Fixes**
  * Average operation times now exclude failed operations when computing aggregates for more accurate performance metrics.
  * Min/max timing updates only apply when incoming data includes at least one successful operation, preserving prior values otherwise.
  * Initial metric population computes averages using successful-operation counts to avoid skew from failures.

* **Tests**
  * Added unit tests covering initialization, merging, error-only batches, weighted averages, and handling of missing timing sums.

### DIFF
--- a/src/test/unit_tests/util_functions_tests/test_stats_collector_utils.test.js
+++ b/src/test/unit_tests/util_functions_tests/test_stats_collector_utils.test.js
@@ -1,0 +1,192 @@
+/* Copyright (C) 2026 NooBaa */
+'use strict';
+
+const { update_nsfs_stats, op_names } = require('../../../util/stats_collector_utils');
+
+/** @param {Record<string, unknown>} entry */
+function expect_finite_timing(entry) {
+    for (const field of ['min_time_milisec', 'max_time_milisec', 'avg_time_milisec']) {
+        expect(entry[field]).not.toBeNull();
+        expect(entry[field]).not.toBeUndefined();
+        expect(Number.isFinite(entry[field])).toBe(true);
+    }
+}
+
+describe('stats_collector_utils.update_nsfs_stats', () => {
+
+    it('does not create stats for an error-only initial batch', () => {
+        const op_name = 'upload_object';
+        expect(op_names).toContain(op_name);
+        const stats = {};
+        update_nsfs_stats(op_name, stats, {
+            count: 1,
+            error_count: 1,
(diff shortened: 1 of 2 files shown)


jackyalbo:
I will let @nadavMiz review this PR, but @sakshimunjal, I think as part of this PR or maybe the next one regarding metrics, maybe we should have tests for these areas - making sure we are getting what we expect.

nadavMiz (approved):
LGTM. I agree with @jackyalbo regarding tests

PR #9673, opened 2026-06-03
Swapped https.Agent.getName with http.Agent.getName

dannyzaken:
### Describe the Problem
`https.Agent.getName` concatenates ca/cert/ciphers/servername/… per request, which costs ~8% CPU on warp runs. 
These are identical for every request through a given agent instance (the agent owns the options), so it is sufficient to disambiguate by TCP identity alone (host:port:localAddress:family) as done in `http.Agent`.

### Explain the Changes
1. Override `getName` in https agents with `http.getName`

1. https://redhat.atlassian.net/browse/DFBUGS-7328

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Refactor**
  * Improved HTTPS connection handling to reduce per-request overhead, yielding faster and more efficient network requests.
  * Transparent performance optimization with no changes to public interfaces or caller-visible behavior.

* **Documentation**
  * Clarified inline comments and JSDoc annotations to improve maintainability and developer understanding.

### DIFF
--- a/src/util/http_utils.js
+++ b/src/util/http_utils.js
@@ -76,6 +76,18 @@ const unsecured_https_proxy_agent = proxy_env_https ?
         proxyEnv: proxy_env_https,
     }) : null;
 
+// https.Agent.getName concatenates ca/cert/ciphers/servername/… per request,
+// which costs ~8% CPU on warp runs. These are identical for every request
+// through a given agent instance (the agent owns the options), so it is sufficient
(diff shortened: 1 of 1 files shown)


nimrod-becker (approved):

liranmauda (commented):
Should we allow `update_https_agents` to change the `getName` or should we throw an error similar to the way we do for `rejectUnauthorized`?

PR #9674, opened 2026-06-04
[Glacier] Fix multi node concurrent runs issue

tangledbytes:
### Describe the Problem
When multiple nodes try to run `manage_nsfs glacier migrate` concurrently, right now they all proceed to process the logs concurrently. This creates the problem that the log files are too small and not enough batching is being done by NooBaa.

### Explain the Changes
This PR adds the time and log size checks within the cluster lock to ensure that one node does the check at one time.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Refactor**
  * Improved staging and execution logic for migration and restore workflows to better coordinate when operations run.
  * Centralized operation orchestration now accepts optional gating and post-staging callbacks to control processing flow.

* **Bug Fixes**
  * Fixed a resource-management issue to ensure file descriptors are always released.
  * Added immediate low-space checks to avoid unnecessary work when storage is constrained.

### DIFF
--- a/src/manage_nsfs/manage_nsfs_glacier.js
+++ b/src/manage_nsfs/manage_nsfs_glacier.js
@@ -12,30 +12,35 @@ const { is_desired_time, record_current_time } = require('./manage_nsfs_cli_util
 async function process_migrations() {
     const fs_context = native_fs_utils.get_process_fs_context();
     const backend = Glacier.getBackend();
+    const timestamp_file_path = path.join(config.NSFS_GLACIER_LOGS_DIR, Glacier.MIGRATE_TIMESTAMP_FILE);
 
-    if (
-        await backend.low_free_space() ||
-        await time_exceeded(fs_context, config.NSFS_GLACIER_MIGRATE_INTERVAL, Glacier.MIGRATE_TIMESTAMP_FILE) ||
-        await migrate_log_exceeds_threshold()
-    ) {
+    if (await backend.low_free_space()) {
         await backend.perform(prepare_galcier_fs_context(fs_context), "MIGRATION");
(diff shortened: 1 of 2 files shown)


src/manage_nsfs/manage_nsfs_glacier.js
```diff
@@ -12,30 +12,35 @@ const { is_desired_time, record_current_time } = require('./manage_nsfs_cli_util
...
     const fs_context = native_fs_utils.get_process_fs_context();
     const backend = Glacier.getBackend();
+    const timestamp_file_path = path.join(config.NSFS_GLACIER_LOGS_DIR, Glacier.MIGRATE_TIMESTAMP_FILE);
 
-    if (
-        await backend.low_free_space() ||
-        await time_exceeded(fs_context, config.NSFS_GLACIER_MIGRATE_INTERVAL, Glacier.MIGRATE_TIMESTAMP_FILE) ||
-        await migrate_log_exceeds_threshold()
-    ) {
+    if (await backend.low_free_space()) {
```
naveenpaul1, line 17: When space is low, we don't need `time_exceeded` check, is that right?

src/sdk/glacier.js
```diff
@@ -222,10 +222,14 @@ class Glacier {
     }
 
     /**
```
naveenpaul1, line 224: Can you add a small comment about what this method is doing?

src/manage_nsfs/manage_nsfs_glacier.js
tangledbytes: Yes, that's right

naveenpaul1 (approved):

PR #9686, opened 2026-06-08
Object Lock: support PutObjectLockConfiguration on existing buckets

kajalpareek-lab:
### Describe the Problem

`PutObjectLockConfiguration` on this branch initially rejected enabling Object Lock on existing buckets, and behavior/validation around Object Lock had some gaps between intended behavior and what CI actually enforced.

### Explain the Changes

1. Enabled Object Lock configuration on existing buckets (AWS-aligned flow), with explicit requirement that bucket versioning is `ENABLED` before lock configuration is applied.
2. Enforced that Object Lock cannot be disabled once enabled, and blocked versioning changes to non-`ENABLED` states on lock-enabled buckets.
3. Improved error mapping/propagation for bucket state and validation failures in the Object Lock S3 path.
4. Updated schema handling for internal object lock metadata states used by bucket metadata.
5. Added and enabled Object Lock integration coverage in CI (`test_s3_worm`), including existing-bucket lock-enable scenarios.
6. Added NSFS validation to reject bucket creation when Object Lock is requested while NSFS versioning is disabled.
7. Included NC stability follow-up adjustments required for this test scope.

- Fixed #RHSTOR-841
- Gap #N/A (no separate gap tracker added in this PR)

### Testing Instructions:

1. Run `node --allow-natives-syntax ./node_modules/.bin/_mocha src/test/integration_tests/api/s3/test_s3_worm.js --timeout 120000`
2. Verify existing-bucket lock-enable flow:
   - fails when versioning is not enabled
   - succeeds after enabling versioning
3. Verify versioning-state protections on lock-enabled buckets.
4. Verify NSFS rejects Object Lock bucket creation when `NSFS_VERSIONING_ENABLED=false`.

- [ ] Doc added/updated
- [x] Tests added

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -1707,7 +1707,9 @@ module.exports = {
             properties: {
                 object_lock_enabled: {
                     type: 'string',
-                    enum: ['Enabled']
+                    // 'Disabled' is an internal bucket metadata state for non-locked buckets.
+                    // S3 PutObjectLockConfiguration still validates request payload as 'Enabled'.
+                    enum: ['Enabled', 'Disabled']
                 },
                 rule: {
                     type: 'object',
--- a/src/endpoint/s3/ops/s3_put_bucket_object_lock.js
+++ b/src/endpoint/s3/ops/s3_put_bucket_object_lock.js
@@ -11,10 +11,10 @@ async function put_bucket_object_lock(req, res) {
     if (!config.WORM_ENABLED) {
         throw new S3Error(S3Error.NotImplemented);
     }
-    // TODO: may require at the future Content-MD5 & bucket-object-lock-token support
     if (req.body.ObjectLockConfiguration.ObjectLockEnabled[0] !== 'Enabled') {
         throw new S3Error(S3Error.MalformedXML);
     }
+
     const lock_configuration = s3_utils.parse_body_object_lock_conf_xml(req);
 
     try {
@@ -23,12 +23,17 @@ async function put_bucket_object_lock(req, res) {
             object_lock_configuration: lock_configuration,
         });
     } catch (err) {
-        let error = err;
         if (err.rpc_code === 'INVALID_SCHEMA' || err.rpc_code === 'INVALID_SCHEMA_PARAMS') {
             console.error('put_bucket_object_lock: Invalid schema provided', err);
-            error = new S3Error(S3Error.MalformedXML);
+            throw new S3Error(S3Error.MalformedXML);
         }
-        throw error;
+        if (err.rpc_code === 'INVALID_BUCKET_STATE') {
+            throw new S3Error(
+                S3Error.InvalidBucketState,
+                err.message
+            );
+        }
+        throw err;
     }
 }
 
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -856,7 +856,8 @@ async function read_bucket_sdk_info(req) {
 async function update_bucket(req) {
     const bucket = find_bucket(req, req.name);
     const conf = bucket.object_lock_configuration;
-    if (config.WORM_ENABLED && conf && conf.object_lock_enabled === 'Enabled' && req.rpc_params.versioning === 'SUSPENDED') {
+    if (config.WORM_ENABLED && conf && conf.object_lock_enabled === 'Enabled' &&
+        req.rpc_params.versioning && req.rpc_params.versioning !== 'ENABLED') {
         throw new RpcError('INVALID_BUCKET_STATE', 'An Object Lock configuration is present on this bucket, so the versioning state cannot be changed.');
     }
 
@@ -2001,9 +2002,21 @@ async function put_object_lock_configuration(req) {
     dbg.log0('add object lock configuration to bucket', req.rpc_params);
     const bucket = find_bucket(req);
 
-    if (bucket.object_lock_configuration.object_lock_enabled !== 'Enabled') {
-        throw new RpcError('INVALID_BUCKET_STATE');
+    const enabling_lock_request = req.rpc_params.object_lock_configuration.object_lock_enabled === 'Enabled';
+
+    if (!enabling_lock_request) {
+        dbg.error('put_object_lock_configuration: ObjectLockEnabled must be Enabled');
+        throw new RpcError('INVALID_SCHEMA_PARAMS');
     }
+
+    if (bucket.versioning !== 'ENABLED') {
+        dbg.error('put_object_lock_configuration: versioning must be ENABLED before enabling Object Lock');
+        throw new RpcError(
+            'INVALID_BUCKET_STATE',
+            "Versioning must be 'Enabled' on the bucket to apply a Object Lock configuration"
+        );
+    }
+
     await system_store.make_changes({
         update: {
             buckets: [{
--- a/src/test/integration_tests/api/s3/test_s3_worm.js
+++ b/src/test/integration_tests/api/s3/test_s3_worm.js
@@ -12,12 +12,14 @@ const mocha = require('mocha');
 const assert = require('assert');
 const path = require('path');
 const fs_utils = require('../../../../util/fs_utils');
+const config = require('../../../../../config');
 const today = new Date();
 const tomorrow = new Date(today);
 tomorrow.setDate(tomorrow.getDate() + 1);
 const todayPlus15 = new Date(today);
 todayPlus15.setDate(todayPlus15.getDate() + 15);
 const malformedXML_message = "The XML you provided was not well-formed or did not validate against our published schema.";
+const original_worm_enabled = config.WORM_ENABLED;
 async function assert_throws_async(promise, expected_code = 'AccessDenied', expected_message = 'Access Denied') {
     try {
         await promise;
@@ -48,6 +50,7 @@ mocha.describe('s3 worm', function() {
     let s3_owner;
     //let s3_a;
     mocha.before(async function() {
+        config.WORM_ENABLED = true;
         const self = this; // eslint-disable-line no-invalid-this
         self.timeout(60000);
         const s3_creds = {
@@ -88,6 +91,9 @@ mocha.describe('s3 worm', function() {
             s3_creds.credentials.secretAccessKey = admin_keys[0].secret_key.unwrap();
             s3_owner = new S3(s3_creds);
         }
+    });
+    mocha.after(function() {
+        config.WORM_ENABLED = original_worm_enabled;
     });
     mocha.describe('buckets creation', function() {
         mocha.it('create bucket BKT & enable lock', async function() {
@@ -980,6 +986,77 @@ mocha.describe('s3 worm', function() {
                 Bucket: BKT,
                 Key: OBJ1,
             }), 'NoSuchObjectLockConfiguration', 'The specified object does not have a ObjectLock configuration');
+        });
+    });
+
+    mocha.describe('enable Object Lock on existing bucket', function() {
+        const EXISTING_BKT = 'existing-bucket-lock-test';
+        const EXISTING_OBJ = 'existing-obj';
+
+        mocha.it('should create bucket without Object Lock', async function() {
+            await s3_owner.createBucket({ Bucket: EXISTING_BKT });
+        });
+
+        mocha.it('should put object before enabling lock', async function() {
+            await s3_owner.putObject({
+                Bucket: EXISTING_BKT,
+                Key: EXISTING_OBJ,
+                Body: file_body,
+                ContentType: 'text/plain'
+            });
+        });
+
+        mocha.it('should fail to enable lock without versioning', async function() {
+            await assert_throws_async(s3_owner.putObjectLockConfiguration({
+                Bucket: EXISTING_BKT,
+                ObjectLockConfiguration: {
+                    ObjectLockEnabled: 'Enabled',
+                    Rule: {
+                        DefaultRetention: { Mode: 'GOVERNANCE', Days: 1 }
+                    }
+                }
+            }), 'InvalidBucketState',
+                "Versioning must be 'Enabled' on the bucket to apply a Object Lock configuration");
+        });
+
+        mocha.it('should enable versioning on bucket', async function() {
+            await s3_owner.putBucketVersioning({
+                Bucket: EXISTING_BKT,
+                VersioningConfiguration: { Status: 'Enabled' }
+            });
+        });
+
+        mocha.it('should enable lock on existing bucket', async function() {
+            await s3_owner.putObjectLockConfiguration({
+                Bucket: EXISTING_BKT,
+                ObjectLockConfiguration: {
+                    ObjectLockEnabled: 'Enabled',
+                    Rule: {
+                        DefaultRetention: { Mode: 'GOVERNANCE', Days: 1 }
+                    }
+                }
+            });
+        });
+
+        mocha.it('should verify lock configuration is set', async function() {
+            const conf = await s3_owner.getObjectLockConfiguration({
+                Bucket: EXISTING_BKT
+            });
+            assert.equal(
+                conf.ObjectLockConfiguration.ObjectLockEnabled, 'Enabled');
+            assert.equal(
+                conf.ObjectLockConfiguration.Rule.DefaultRetention.Mode,
+                'GOVERNANCE');
+            assert.equal(
+                conf.ObjectLockConfiguration.Rule.DefaultRetention.Days, 1);
+        });
+
+        mocha.it('existing object should not have retention', async function() {
+            await assert_throws_async(s3_owner.getObjectRetention({
+                Bucket: EXISTING_BKT,
+                Key: EXISTING_OBJ,
+            }), 'NoSuchObjectLockConfiguration',
+            'The specified object does not have a ObjectLock configuration');
         });
     });
 
--- a/src/endpoint/s3/ops/s3_put_object_legal_hold.js
+++ b/src/endpoint/s3/ops/s3_put_object_legal_hold.js
@@ -12,7 +12,6 @@ async function put_object_legal_hold(req) {
     if (!config.WORM_ENABLED) {
         throw new S3Error(S3Error.NotImplemented);
     }
-    // TODO: may require at the future Content-MD5 support
     const legal_hold_status = req.body.LegalHold.Status[0];
     if (legal_hold_status !== 'ON' && legal_hold_status !== 'OFF') {
         throw new S3Error(S3Error.MalformedXML);
--- a/src/endpoint/s3/ops/s3_put_object_retention.js
+++ b/src/endpoint/s3/ops/s3_put_object_retention.js
@@ -12,7 +12,6 @@ async function put_object_retention(req) {
     if (!config.WORM_ENABLED) {
         throw new S3Error(S3Error.NotImplemented);
     }
-    // TODO: may require at the future Content-MD5 support
     if (!req.body.Retention) throw new S3Error(S3Error.MalformedXML);
     const mode = req.body.Retention.Mode[0];
     let retain_until_date = req.body.Retention.RetainUntilDate[0];
--- a/src/test/utils/index/index.js
+++ b/src/test/utils/index/index.js
@@ -101,7 +101,7 @@ require('../../unit_tests/util_functions_tests/test_cloud_utils');
 require('../../integration_tests/internal/test_upgrade_scripts.js');
 require('../../integration_tests/internal/test_tiering_ttl_worker');
 // require('./test_tiering_upload');
-//require('../../integration_tests/api/s3/test_s3_worm.js');
+require('../../integration_tests/api/s3/test_s3_worm.js');
 require('../../integration_tests/api/s3/test_bucket_logging');
 require('../../integration_tests/api/s3/test_notifications');
 require('../../integration_tests/api/s3/test_chunked_upload');
--- a/src/util/native_fs_utils.js
+++ b/src/util/native_fs_utils.js
@@ -681,6 +681,12 @@ function validate_bucket_creation(params) {
         !VALID_BUCKET_NAME_REGEXP.test(params.name)) {
         throw new RpcError('INVALID_BUCKET_NAME');
     }
+    if (params.lock_enabled && !config.NSFS_VERSIONING_ENABLED) {
+        throw new RpcError(
+            'INVALID_BUCKET_STATE',
+            'Object Lock cannot be enabled when NSFS versioning is disabled'
+        );
+    }
 }
 
 /**


src/api/common_api.js
```diff
@@ -1596,7 +1596,7 @@ module.exports = {
             properties: {
                 object_lock_enabled: {
                     type: 'string',
-                    enum: ['Enabled']
+                    enum: ['Enabled', 'Disabled']
```
nadavMiz, line 1712: note that from aws API. only valid value is Enabled.(https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectLockConfiguration.html#API_PutObjectLockConfiguration_RequestBody) why do we need Disabled value?

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
```diff
@@ -28,6 +31,9 @@ async function put_bucket_object_lock(req, res) {
             console.error('put_bucket_object_lock: Invalid schema provided', err);
             error = new S3Error(S3Error.MalformedXML);
         }
+        if (err.rpc_code === 'INVALID_BUCKET_STATE') {
+            error = new S3Error(S3Error.InvalidBucketState);
+        }
```
shirady, line 35: I'm not sure you need this.
As part of the flow in `s3_rest` it is mapped:

the flow: in `s3_rest` the function `handle_error` -> `_prepare_error` -> there is the mapping :
https://github.com/noobaa/noobaa-core/blob/d107738420e9b185b8368d16134d19506035b2aa/src/endpoint/s3/s3_errors.js#L652

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
```diff
@@ -11,10 +11,13 @@ async function put_bucket_object_lock(req, res) {
     if (!config.WORM_ENABLED) {
         throw new S3Error(S3Error.NotImplemented);
     }
-    // TODO: may require at the future Content-MD5 & bucket-object-lock-token support
+    if (!req.content_md5) {
+        throw new S3Error(S3Error.InvalidDigest);
+    }
```
shirady: I'm not sure about this change (whether it is needed and the scope of validation for encryption).
Could you share more information on it?

What I saw we have in the code when I searched:
https://github.com/noobaa/noobaa-core/blob/d107738420e9b185b8368d16134d19506035b2aa/src/endpoint/s3/s3_utils.js#L182-L188

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
```diff
@@ -28,6 +31,9 @@ async function put_bucket_object_lock(req, res) {
             console.error('put_bucket_object_lock: Invalid schema provided', err);
             error = new S3Error(S3Error.MalformedXML);
         }
+        if (err.rpc_code === 'INVALID_BUCKET_STATE') {
```
nadavMiz, line 30: note that we should already have rpc to s3 error convertion in s3_rest.js `handle_error function`. also, not sure if its importent you would lose the error message you replace with S3 error here.

src/server/system_services/bucket_server.js
```diff
@@ -2001,15 +2001,32 @@ async function put_object_lock_configuration(req) {
     dbg.log0('add object lock configuration to bucket', req.rpc_params);
     const bucket = find_bucket(req);
 
-    if (bucket.object_lock_configuration.object_lock_enabled !== 'Enabled') {
-        throw new RpcError('INVALID_BUCKET_STATE');
+    const is_lock_enabled = bucket.object_lock_configuration &&
+        bucket.object_lock_configuration.object_lock_enabled === 'Enabled';
+    const enabling_lock = req.rpc_params.object_lock_configuration.object_lock_enabled === 'Enabled';
```
shirady: Could you explain the difference between `is_lock_enabled` and `enabling_lock`?

src/server/system_services/bucket_server.js
```diff
@@ -2001,15 +2001,32 @@ async function put_object_lock_configuration(req) {
...
+
+    if (!is_lock_enabled && !enabling_lock) {
```
nadavMiz: you should verify me but I think the error for not having `object_lock_enabled === 'Enabled'` is MalformedXML and not InvalidBucketState

src/server/system_services/bucket_server.js
```diff
@@ -2001,15 +2001,32 @@ async function put_object_lock_configuration(req) {
...
+            'Object Lock cannot be disabled once it is enabled');
+    }
+
+    const updates = {
+        _id: bucket._id,
+        object_lock_configuration: req.rpc_params.object_lock_configuration
+    };
+
+    if (enabling_lock && bucket.versioning !== 'ENABLED') {
+        updates.versioning = 'ENABLED';
```
shirady: Did you test it in AWS? Would that change for both cases? (bucket versioning suspended and bucket versioning disabled).

src/api/common_api.js
shirady, line 1712: I'm not sure about this change.
From what I found in the AWS CLI docs (looked there because it is quick to find) [link](https://docs.aws.amazon.com/cli/latest/reference/s3api/put-object-lock-configuration.html):

> --object-lock-configuration (structure)
> 
> The Object Lock configuration that you want to apply to the specified bucket.
> 
> ObjectLockEnabled -> (string)
> 
> Indicates whether this bucket has an Object Lock configuration enabled. Enable ObjectLockEnabled when you apply ObjectLockConfiguration to a bucket.
> 
> Possible values:
> 
> Enabled

The possible value is only "Enabled".

src/server/system_services/bucket_server.js
nadavMiz: since the error message is ignored anyway. I think we should either leave the schemas as it was before. where it only allowed Enabled. or restore the condition to what it was before where it only validated Enabled:
```
if (bucket.object_lock_configuration.object_lock_enabled !== 'Enabled') {
        throw new RpcError('INVALID_BUCKET_STATE');
```

if you feel like the messages are informative, maybe log them instead?
somehting like:
```
log.error('Object Lock configuration cannot be set when Object Lock is not enabled')
throw new RpcError('INVALID_BUCKET_STATE')
```
but in that case you should verify you don't break NC, as I think it relies on the schema validating this value

src/server/system_services/bucket_server.js
```diff
@@ -2001,15 +2001,32 @@ async function put_object_lock_configuration(req) {
...
+        throw new RpcError('INVALID_BUCKET_STATE',
+            'Object Lock cannot be disabled once it is enabled');
+    }
+
+    const updates = {
+        _id: bucket._id,
+        object_lock_configuration: req.rpc_params.object_lock_configuration
+    };
+
+    if (enabling_lock && bucket.versioning !== 'ENABLED') {
```
nadavMiz: can you validate this against aws. I remmember that it returns an error in this case. something like:
"Versioning must be 'Enabled' on the bucket to apply a Object Lock configuration"

shirady (commented):
I added a few comments and questions

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -983,6 +983,70 @@ mocha.describe('s3 worm', function() {
         });
     });
 
+    mocha.describe('enable Object Lock on existing bucket', function() {
+        const EXISTING_BKT = 'existing-bucket-lock-test';
+        const EXISTING_OBJ = 'existing-obj';
+
+        mocha.before(async function() {
+            if (is_nc_coretest) {
```
nadavMiz: its not a problem but object locking should be fully implemented in NC. in case some functionality is missing / not alligned in NC, please let me know, and I'll fix it

shirady:
@kajalpareek-lab I saw this in AWS docs ([link](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-configure.html)):

> After you enable Object Lock on a bucket, you can't disable Object Lock or suspend versioning for that bucket.

I'm not sure that NooBaa blocks setting versioning `suspend` on a bucket enabled lock bucket (@nadavMiz maybe you know?)
Is it in the scope of the epic? (If not, I would open an issue).

src/api/common_api.js
kajalpareek-lab: The 'Disabled' value is used internally when storing bucket metadata for buckets created without Object Lock (bucket_server.js:72). It's not exposed in the S3 API response — GetObjectLockConfiguration throws ObjectLockConfigurationNotFoundError for non-locked buckets. The schema covers internal RPC validation, not just the S3 API surface.

kajalpareek-lab: The 'Disabled' value is used internally when storing bucket metadata for buckets created without Object Lock (bucket_server.js:72). It's not exposed in the S3 API response — GetObjectLockConfiguration throws ObjectLockConfigurationNotFoundError for non-locked buckets. The schema covers internal RPC validation, not just the S3 API surface.

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
kajalpareek-lab: yes you are right my explicit catch is redundant. Let me remove it.

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
kajalpareek-lab: added changes.

src/server/system_services/bucket_server.js
kajalpareek-lab: is_lock_enabled — checks the current state of the bucket in the database (is Object Lock already enabled on this bucket right now?)
enabling_lock — checks the incoming request (is the user trying to enable Object Lock with this API call?) We need both because:
If bucket is already locked (is_lock_enabled=true) and request sends Enabled → just updating the default retention rule
If bucket is NOT locked (is_lock_enabled=false) and request sends Enabled → enabling lock for the first time (also auto-enables versioning)
If bucket is already locked and request sends Disabled → reject (lock is immutable)
If bucket is NOT locked and request sends Disabled → reject (nothing to configure)

src/server/system_services/bucket_server.js
kajalpareek-lab: Can't revert schema to only ['Enabled'] — bucket_server.js:72 stores 'Disabled' during bucket creation without lock, and NC uses the same path. Reverting would break bucket creation.

For the condition — the old check (!== 'Enabled' → throw) was the bug this PR fixes. But I agree the messages are not surfaced. I'll simplify to:

```

if (!is_lock_enabled && !enabling_lock) {
    dbg.error('put_object_lock_configuration: cannot set lock config when lock is not enabled');
    throw new RpcError('INVALID_BUCKET_STATE');
}
if (is_lock_enabled && !enabling_lock) {
    dbg.error('put_object_lock_configuration: cannot disable Object Lock once enabled');
    throw new RpcError('INVALID_BUCKET_STATE');
}
```

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: Acknowledged. Currently skipping NC in this test block because the flow relies on read_bucket + RPC path which differs in NC. Will follow up if any NC-specific gaps are found during further testing.

src/server/system_services/bucket_server.js
kajalpareek-lab: Did not test directly against AWS — validated from the AWS docs: https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock-configure.html

"Object Lock works only with versioned buckets."

The check bucket.versioning !== 'ENABLED' covers both cases:

Versioning disabled (never enabled) → rejected
Versioning suspended (was enabled, then suspended) → rejected
Only 'ENABLED' passes. This matches AWS behavior — both disabled and suspended are invalid states for enabling Object Lock.

src/server/system_services/bucket_server.js
kajalpareek-lab: Updated the code to throw INVALID_BUCKET_STATE with message "Versioning must be Enabled on the bucket to apply a Object Lock configuration" instead of auto-enabling versioning. Also added a test case that verifies this behavior.

nadavMiz:
@shirady @kajalpareek-lab should probably verify me. but it looks like it is implemented:

https://github.com/noobaa/noobaa-core/blob/b65c8e7a8a9ee89f346325aeb5ca28e77435909c/src/server/system_services/bucket_server.js#L856-L861

there is also a test for this: `should fail suspend versioning on locked bucket`. but I am not sure if it actually passes

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
kajalpareek-lab: he req.content_md5 is parsed from the Content-MD5 HTTP header in http_utils.js:716-722 — it's a base64-encoded MD5 of the request body, used for integrity verification. Not related to SSE/encryption.
The same requirement applies to PutObjectRetention and PutObjectLegalHold — AWS mandates Content-MD5 for all Object Lock write APIs.

src/server/system_services/bucket_server.js
nadavMiz: why was the old check a bug? isn't the new check logically equivalent to the old one?
also isn't bucket_server.js:72 a bug? wouldn't that mean that get_object_lock_configuration will return object_lock_enabled = Disabled? which is not allowed by the API. 
maybe its better to have it set as undefined instead of Disabled

src/test/integration_tests/api/s3/test_s3_worm.js
nadavMiz: what do you mean by the flow relies on read_bucket + RPC path? don't they use S3 client? I think its better if we have the same tests for both paths. unless one of them doesn't work properly (might be intentional), or if there is a specific reason for paths to differ

src/server/system_services/bucket_server.js
shirady: Maybe you can change the variable name `enabling_lock` to have a suffix of `_req` or _`request` so it would be clearer when reading the code?

src/api/common_api.js
shirady: Could you write it as a comment near the change? so it would be clear for everyone

src/server/system_services/bucket_server.js
```diff
@@ -2001,9 +2001,25 @@ async function put_object_lock_configuration(req) {
...
+
+    if (is_lock_enabled && !enabling_lock) {
+        throw new RpcError('INVALID_BUCKET_STATE',
+            'Object Lock cannot be disabled once it is enabled');
     }
+
+    if (enabling_lock && bucket.versioning !== 'ENABLED') {
+        throw new RpcError('INVALID_BUCKET_STATE',
+            'Versioning must be Enabled on the bucket to apply a Object Lock configuration');
+    }
```
shirady: I think you need to make sure that the final error (using AWS CLI for example) you see this message.
With the current code, I'm not sure about it - 
https://github.com/noobaa/noobaa-core/blob/fbdb7f8b8fe3ad845d40e703a1c073ace9295a35/src/endpoint/s3/s3_rest.js#L533-L540

src/server/system_services/bucket_server.js
kajalpareek-lab: Updated and old check (current != Enabled -> reject) blocks Disabled -> Enabled, which is the exact transition needed to enable Object Lock on existing buckets.
Current logic compares current state (is_lock_enabled) and requested state (enabling_lock) separately, so:
Disabled -> Enabled is allowed (with versioning check)
Enabled -> Disabled is blocked (immutability)
'Disabled' at bucket_server.js:72 is internal metadata only; API never returns it because get_object_lock_configuration() throws OBJECT_LOCK_CONFIGURATION_NOT_FOUND_ERROR when state is not Enabled.

src/server/system_services/bucket_server.js
kajalpareek-lab: Validated on live cluster (ci-ln-dywyq62) with deployed dev-20260615 images.
Without versioning, PutObjectLockConfiguration returns InvalidBucketState; after enabling versioning, the same request succeeds.
So I updated the code to reject when bucket.versioning !== 'ENABLED' with: "Versioning must be Enabled on the bucket to apply a Object Lock configuration" and added an integration test covering this flow.

src/server/system_services/bucket_server.js
kajalpareek-lab: Thanks, agreed. For put_object_lock_configuration, I now log detailed context with dbg.error(...) and throw generic RpcError('INVALID_BUCKET_STATE'), same pattern as other paths.

So behavior is now consistent with existing error handling conventions.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: Good point — agreed we should have the same S3-level coverage for both paths. The S3 calls are the same, but NC currently has different backend implementation paths and some Object Lock behavior is not fully aligned yet, so I skipped NC to avoid mixing unrelated NC gaps into this PR.
I’ll split this cleanly: Keep these tests enabled for containerized path in this PR (to validate the behavior changed here).
Open follow-up to enable the same block for NC once we align/confirm NC behavior for existing-bucket lock enable flow.

src/server/system_services/bucket_server.js
kajalpareek-lab: Done — renamed for clarity:
enabling_lock -> enabling_lock_request

src/server/system_services/bucket_server.js
kajalpareek-lab: You're right for the S3 path — ObjectLockEnabled != "Enabled" should return MalformedXML, and that validation already happens in s3_put_bucket_object_lock.js before RPC is called.

I updated bucket_server.js to simplify this:

removed the previous is_lock_enabled/enabling_lock_request mixed branch logic
kept only a defensive generic check for non-S3/internal RPC callers (INVALID_BUCKET_STATE + log)
So S3 behavior remains MalformedXML, and backend stays protected.

kajalpareek-lab:
> @shirady @kajalpareek-lab should probably verify me. but it looks like it is implemented:
> 
> https://github.com/noobaa/noobaa-core/blob/b65c8e7a8a9ee89f346325aeb5ca28e77435909c/src/server/system_services/bucket_server.js#L856-L861
> 
> there is also a test for this: `should fail suspend versioning on locked bucket`. but I am not sure if it actually passes

Verified locally in test_s3_worm.js: should fail suspend versioning on locked bucket passes and returns InvalidBucketState via the update_bucket guard when versioning is set to SUSPENDED on a locked bucket.

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
```diff
@@ -11,10 +11,10 @@ async function put_bucket_object_lock(req, res) {
     if (!config.WORM_ENABLED) {
         throw new S3Error(S3Error.NotImplemented);
     }
-    // TODO: may require at the future Content-MD5 & bucket-object-lock-token support
```
shirady, line 14: I see that you removed the TODO "may require in the future Content-MD5 & bucket-object-lock-token support" in several places - do you address those in this PR? Could you point where?
If not, I think it would be better to leave the comment as is in the meantime.

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -983,6 +983,82 @@ mocha.describe('s3 worm', function() {
...
+        mocha.it('should fail to enable lock without versioning', async function() {
+            await assert_throws_async(s3_owner.putObjectLockConfiguration({
+                Bucket: EXISTING_BKT,
+                ObjectLockConfiguration: {
+                    ObjectLockEnabled: 'Enabled',
+                    Rule: {
+                        DefaultRetention: { Mode: 'GOVERNANCE', Days: 1 }
+                    }
+                }
+            }), 'InvalidBucketState', 'Versioning must be Enabled on the bucket to apply a Object Lock configuration');
```
shirady: Minor - the message should be "Versioning must be 'Enabled' on the bucket to apply a Object Lock configuration" (single quote on the 'Enabled').

Where is the flow do you add the message? I didn't find it in the flow

src/test/integration_tests/api/s3/test_s3_worm.js
shirady, line 1: I think this test is not running in the CI.
You need to enable it by removing the comment in this line:

https://github.com/noobaa/noobaa-core/blob/37d250cd2bc0b05e78dc90fa573b8fa6fe512b58/src/test/utils/index/index.js#L104

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -983,6 +983,82 @@ mocha.describe('s3 worm', function() {
         });
     });
 
+    mocha.describe('enable Object Lock on existing bucket', function() {
+        const EXISTING_BKT = 'existing-bucket-lock-test';
+        const EXISTING_OBJ = 'existing-obj';
+
+        mocha.before(async function() {
```
shirady: Did it work for you to skip in the `mocha.before`?
I think you would need to skip each test or skip at the level of the `describe` (I'm not sure). How does the report/summary look when it is like that?

src/server/system_services/bucket_server.js
nadavMiz: did you validate against aws this is the expected error and message?

src/test/integration_tests/api/s3/test_s3_worm.js
nadavMiz: NC object lock should be fully implemented, there is no reason the paths should not be align except for the error message maybe if its not fully propagated. also if there is something missing in either paths we should know about it, not hide it. in order to know if NC work you only need to remove the skip and let it run on CI. if it doesn't work you can let me know and I can look at it. if its really something missing you can return the skip, and I'll remove it once fixed.

src/server/system_services/bucket_server.js
nadavMiz: but wouldn't throwing OBJECT_LOCK_CONFIGURATION_NOT_FOUND_ERROR when state is not enabled is a bug by itself? this scenario should never be possible. at best it should modify the response to return an empty object. but I don't understand why we are doing so much work to accommodate a value that should not be in the first place. is there any value to setting object_lock_enabled to false in the first place? why not just have it undefined instead and avoid the whole handling?

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
kajalpareek-lab: yes, this PR addresses that TODO by design.

x-amz-bucket-object-lock-token: we removed special handling because AWS deprecated this requirement for enabling Object Lock on existing buckets.
Content-MD5: we removed strict req.content_md5 enforcement for Object Lock PUT APIs (put_bucket_object_lock, put_object_retention, put_object_legal_hold) because it breaks AWS CLI v2/flexible-checksum requests and is not required for current AWS-compatible behavior.
So the old TODO is no longer accurate as-is. If you prefer, I can add an updated comment stating this rationale (instead of keeping the old TODO text).

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: I did run src/test/integration_tests/api/s3/test_s3_worm.js manually while developing/validating these changes, but you’re right that it is currently not part of the default CI mocha index because src/test/utils/index/index.js still has it commented.
I’ll enable it in the index in this PR so it runs in CI as well.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: I rechecked this. this.skip() in the before of that describe works as expected for this case — the two NC-only tests are reported as pending (- ...) and the summary includes them as pending (not passed).
If you prefer clearer intent, I can switch to conditional describe-level skipping (is_nc_coretest ? describe(...) : describe.skip(...)).

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: Agreed, and thanks for pushing on this.
I removed the NC skip from this block and enabled test_s3_worm.js in the default mocha index, so this now runs in CI instead of being hidden.
I also reran the suite in NC mode locally after unskipping; the flow executes and passes.
I did not reintroduce a skip.

The only path difference I saw is error-message text (InvalidBucketState message string), which is expected today due to NC vs containerized error propagation. Behavior-wise the test coverage is now aligned and active for both paths.

If CI shows an NC-specific failure, I’ll share the exact failing case/logs here and tag you immediately.

src/server/system_services/bucket_server.js
kajalpareek-lab: This scenario is possible by design now that we support enabling Object Lock on existing buckets.
A bucket can legitimately exist without lock, so GetObjectLockConfiguration should return not-found until lock is enabled.

In our current model, that “no lock yet” state is stored internally as object_lock_enabled: 'Disabled' (internal metadata only).
We do not expose Disabled via S3 API:

get_object_lock_configuration() returns OBJECT_LOCK_CONFIGURATION_NOT_FOUND_ERROR unless state is Enabled, which matches S3 behavior for buckets without lock config.
We could refactor storage to use undefined instead of Disabled, but that is a broader cleanup (schema + multiple code paths that currently assume object_lock_configuration exists) and is outside this PR scope.
This PR keeps behavior correct for existing-bucket enablement and API compatibility.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: Good catch on the quote — agreed.
I updated the expectation to use "Versioning must be 'Enabled' on the bucket to apply a Object Lock configuration" for the NC path.

About where this message comes from:

In NC flow, it is set explicitly in BucketSpaceFS.put_object_lock_configuration() in src/sdk/bucketspace_fs.js (err.message = "Versioning must be 'Enabled' ...").
In the containerized flow, bucket_server throws RpcError('INVALID_BUCKET_STATE'), and s3_rest maps RPC code to InvalidBucketState with the generic message ("The request is not valid with the current state of the bucket."), since custom RPC message passthrough was removed.
So the test now checks message per path (NC vs containerized) while still asserting the same error code (InvalidBucketState).

shirady: The error message should be the same as in AWS in all environments (NC, containerized).

src/test/utils/coretest/nc_coretest.js
```diff
@@ -333,11 +333,13 @@ const get_name_by_email = email => {
  * get_admin_mock_account_details returns the account details that we use in NC core tests
  */
 function get_admin_mock_account_details() {
+    const uid = process.getuid ? process.getuid() : 200;
```
nadavMiz: why is this change needed? I don't think we even use this function in test_s3_worm.js

kajalpareek-lab: It is used indirectly by test_s3_worm in NC mode via require_coretest() -> nc_coretest -> nc-coretest-before -> admin_account_creation() -> get_admin_mock_account_details().
That said, I reverted this uid/gid change now because it affected existing NC namespace-fs permission expectations in CI (run-nc-unit-tests). So this concern is addressed and the branch no longer contains that override.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: updated.

src/server/system_services/bucket_server.js
```diff
@@ -2001,9 +2001,18 @@ async function put_object_lock_configuration(req) {
     dbg.log0('add object lock configuration to bucket', req.rpc_params);
     const bucket = find_bucket(req);
 
-    if (bucket.object_lock_configuration.object_lock_enabled !== 'Enabled') {
+    const enabling_lock_request = req.rpc_params.object_lock_configuration.object_lock_enabled === 'Enabled';
+
+    if (!enabling_lock_request) {
+        dbg.error('put_object_lock_configuration: ObjectLockEnabled must be Enabled');
         throw new RpcError('INVALID_BUCKET_STATE');
```
nadavMiz: even though its defensive. we might as well throw the correct error. I think its InvalidXML

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
```diff
@@ -23,12 +23,17 @@ async function put_bucket_object_lock(req, res) {
...
-        let error = err;
         if (err.rpc_code === 'INVALID_SCHEMA' || err.rpc_code === 'INVALID_SCHEMA_PARAMS') {
             console.error('put_bucket_object_lock: Invalid schema provided', err);
-            error = new S3Error(S3Error.MalformedXML);
+            throw new S3Error(S3Error.MalformedXML);
+        }
+        if (err.rpc_code === 'INVALID_BUCKET_STATE') {
+            throw new S3Error(
+                S3Error.InvalidBucketState,
+                "Versioning must be 'Enabled' on the bucket to apply a Object Lock configuration"
```
nadavMiz: I think it would be better to save the error message in the rpc error and transfer it to the s3error here.
this will enable us to have different INVALID_BUCKET_STATE errors coming from the server in the future. also it where the error is thrown so its more clear why we have this message.

kajalpareek-lab: Updated.

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
```diff
@@ -23,12 +23,17 @@ async function put_bucket_object_lock(req, res) {
...
-        let error = err;
         if (err.rpc_code === 'INVALID_SCHEMA' || err.rpc_code === 'INVALID_SCHEMA_PARAMS') {
             console.error('put_bucket_object_lock: Invalid schema provided', err);
-            error = new S3Error(S3Error.MalformedXML);
+            throw new S3Error(S3Error.MalformedXML);
+        }
+        if (err.rpc_code === 'INVALID_BUCKET_STATE') {
+            throw new S3Error(
+                S3Error.InvalidBucketState,
+                err.message || "Versioning must be 'Enabled' on the bucket to apply a Object Lock configuration"
```
nadavMiz: why `|| "Versioning must be 'Enabled' on the bucket to apply a Object Lock configuration"`
isn't it already passed in err.message? I think the default case should be the default s3err.
you can just leave the message undefined in this case, the default message should be passed by S3Error.InvalidBucketState,
something like:
```
throw new S3Error(
                S3Error.InvalidBucketState,
                err.message
```

kajalpareek-lab: I added it initially as a defensive fallback while moving the INVALID_BUCKET_STATE message source from S3 layer to backend RPC error, to avoid returning an empty/undefined message if any path raised that code without a message.
Now that the backend always sets the message for this flow, the fallback is unnecessary, so I removed it.

nadavMiz (approved):

src/api/common_api.js
```diff
@@ -1707,7 +1707,9 @@ module.exports = {
             properties: {
                 object_lock_enabled: {
                     type: 'string',
-                    enum: ['Enabled']
+                    // 'Disabled' is an internal bucket metadata state for non-locked buckets.
+                    // S3 PutObjectLockConfiguration still validates request payload as 'Enabled'.
+                    enum: ['Enabled', 'Disabled']
```
shirady, line 1712: @kajalpareek-lab can you add a test where the client tries to `put-object-lock-configuration` with `ObjectLockEnabled` as `Disabled`? (to ensure that it is not valid)

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
```diff
@@ -23,12 +23,17 @@ async function put_bucket_object_lock(req, res) {
...
         if (err.rpc_code === 'INVALID_SCHEMA' || err.rpc_code === 'INVALID_SCHEMA_PARAMS') {
             console.error('put_bucket_object_lock: Invalid schema provided', err);
-            error = new S3Error(S3Error.MalformedXML);
+            throw new S3Error(S3Error.MalformedXML);
+        }
+        if (err.rpc_code === 'INVALID_BUCKET_STATE') {
+            throw new S3Error(
+                S3Error.InvalidBucketState,
+                err.message
+            );
```
shirady, line 34: I can suggest adding another `rpc_code` and another S3 error (and mapping it) for this case if you want to avoid this additional code here (if you want, it can be also in a separate PR).
I did something like that here:
https://github.com/noobaa/noobaa-core/blame/0322f2257ef6c30e1dcf1d7b878d241587b9a148/src/endpoint/s3/s3_errors.js#L552-L561

src/api/common_api.js
kajalpareek-lab: Added already — this case is covered in test_s3_worm by should fail put lock configuration with disabled lock, which calls putObjectLockConfiguration with ObjectLockEnabled: 'Disabled' and asserts MalformedXML.

src/endpoint/s3/ops/s3_put_bucket_object_lock.js
kajalpareek-lab: Good suggestion, thanks.
For this PR I kept the explicit mapping in s3_put_bucket_object_lock.js to preserve current behavior with minimal change.
Adding a dedicated rpc_code + S3 error mapping is a larger refactor of error contracts, so I prefer to keep this PR scoped and handle that as a separate cleanup PR.

shirady (approved):
LGTM

PR #9693, opened 2026-06-10
Move to use HeadBucket instead of list buckets for verifications

nimrod-becker:
### Describe the Problem
1. Scoped down IAM policies that prevent users from using list buckets would fail to set up cloud resources.

### Explain the Changes
1. In verifications of existence (pool server), use HeadBucket instead of list to verify a specific bucket exists. Head operation failing with AccessDenied means no real access to the bucket, and is ok to fail on.
2. In verifications of credentials (external connection), use list operation, but allow for AccessDenied (creds are ok, just operation is scoped)

3. Due to failures of PSQL to be ready in 20sec during tests, increase max t/o to 60 sec but poll every 2sec

1. Example https://redhat.atlassian.net/browse/DFBUGS-6783

### Testing Instructions:
1. Allow the user to access a bucket, but put in an IAM policies reventing him from listing buckets. See that the external connection and resource can be created using these credentials.

- [X] Doc added/updated - No need
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added a cloud bucket existence verification endpoint that returns `{ exists: boolean }`.

* **Bug Fixes**
  * Improved AWS connection validation: `AccessDenied` during bucket listing is now treated as successful.

* **Refactor**
  * Updated cloud pool creation and namespace provisioning to verify the target bucket directly (without listing buckets).

* **Chores**
  * Improved Postgres startup in tests by polling `pg_isready` (up to 60s) instead of using a fixed delay.

### DIFF
--- a/src/api/bucket_api.js
+++ b/src/api/bucket_api.js
@@ -622,35 +622,25 @@ module.exports = {
             }
         },
 
-        get_cloud_buckets: {
+        verify_cloud_bucket_exists: {
             method: 'GET',
             params: {
                 type: 'object',
-                required: ['connection'],
+                required: ['connection', 'target_bucket'],
                 properties: {
                     connection: {
+                        type: 'string'
+                    },
+                    target_bucket: {
                         type: 'string'
                     }
                 }
             },
             reply: {
-                type: 'array',
-                items: {
-                    type: 'object',
-                    required: ['name'],
-                    properties: {
-                        name: { $ref: 'common_api#/definitions/bucket_name' },
-                        used_by: {
-                            type: 'object',
-                            required: ['name', 'usage_type'],
-                            properties: {
-                                name: { $ref: 'common_api#/definitions/bucket_name' },
-                                usage_type: {
-                                    type: 'string',
-                                    enum: ['CLOUD_RESOURCE', 'NAMESPACE_RESOURCE']
-                                },
-                            }
-                        }
+                type: 'object',
+                properties: {
+                    exists: {
+                        type: 'boolean'
                     }
                 }
             },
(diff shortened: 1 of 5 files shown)


src/api/bucket_api.js
```diff
@@ -622,35 +622,25 @@ module.exports = {
...
             params: {
                 type: 'object',
                 required: ['connection'],
                 properties: {
                     connection: {
                         type: 'string'
+                    },
+                    target_bucket: {
+                        type: 'string'
                     }
```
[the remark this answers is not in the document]
nimrod-becker: target_bucket moved to required, and as such would fail on the client if not sent

src/server/system_services/bucket_server.js
```diff
@@ -1407,68 +1407,32 @@ async function get_cloud_buckets(req) {
...
-            const res = await P.timeout(EXTERNAL_BUCKET_LIST_TO, s3.listBuckets({}));
-            const buckets = res.Buckets ?? [];
-            return buckets.map(bucket => _inject_usage_to_cloud_bucket(bucket.Name, connection.endpoint, used_cloud_buckets));
+            await P.timeout(EXTERNAL_BUCKET_LIST_TO, Promise.resolve(s3.headBucket({ Bucket: target_bucket })));
         }
+        return { exists: true };
     } catch (err) {
         noobaa_s3_client.fix_error_object(err); // only relevant when using AWS SDK v3
+        const code = err.Code || err.code || (err.details && err.details.code);
+        if (code === 'ContainerNotFound' || code === 'NoSuchBucket' || code === 'NotFound') {
```
tangledbytes, line 1473: Do we not want to catch `AccessDenied` here?

nimrod-becker: Good catch

PR #9695, opened 2026-06-10
core HA | change init script for core pod to work with node

nadavMiz:
### Describe the Problem
for core HA work,  currently noobaa core init script is the bash script this limit the ability to run lease renewal in the background causing problems with implementing it on a different process

### Explain the Changes
1. replace the current supervisord.orig script with a node module core-init.js that does the same thing
2. remove old upervisord.orig script
4. change NooBaa.docker CMD to the new core_init.js and remove supervisord.orig related code 
5. remove noobaa core path from noobaa_init.sh as it is now implemented inside core_init.sh

### Issues
part of https://redhat.atlassian.net/browse/RHSTOR-7491

### Testing Instructions:
1. `node ./node_modules/.bin/_mocha src/test/unit_tests/internal/test_core_init.js`

- [x] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Refactor**
  * Updated container and init startup to execute the Node core init entrypoint directly (removing the legacy wrapper flow) and removing the server init/upgrade fallback behavior from the container init script.
  * Streamlined supervisor setup by removing the old `supervisord` wrapper, moving the binary to `/usr/bin/supervisord`, ensuring `/var/log/supervisor` exists, and routing supervisor logs under `/var/log/supervisor`.
  * Updated Linux diagnostics to collect supervisor logs from `/var/log/supervisor/*`.
* **Tests**
  * Added unit tests for core init: version validation, upgrade/subprocess error propagation, and startup sequencing; also asserts container/Docker startup expectations (no reference to the removed `supervisord.orig`).


src/deploy/NVA_build/setup_platform.sh
```diff
@@ -40,8 +40,7 @@ function install_supervisor {
     # Autostart supervisor
     deploy_log "setup_supervisors autostart"
     bin_supervisord=$(find / -name supervisord | grep bin)
-    mv ${bin_supervisord} /usr/bin/supervisord_orig
-    mv /tmp/supervisord /usr/bin/supervisord
+    mv ${bin_supervisord} /usr/bin/supervisord
```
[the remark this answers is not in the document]
nadavMiz: this was already there and worked for years before this pr. I feel like this is out of scope for this pr. not sure if its even possible in current flow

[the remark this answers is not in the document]
nadavMiz: @coderabbitai correct me if I am wrong on this. but this case is not possible. setup_platform.sh only runs as part of the image build process. if container restart or re-run it will start from the run command (i.e. core_init.sh). am I missing something?

src/cmd/core_init.js
```diff
@@ -0,0 +1,157 @@
...
+ * @returns {boolean}
+ */
+function passwd_has_user_entry(passwd_content, username) {
+    return passwd_content.split('\n').some(line => line.startsWith(`${username}:`));
+}
+
+/**
+ * Ensures the non-root runtime user has an /etc/passwd entry on OpenShift.
+ */
+function fix_non_root_user() {
```
dannyzaken: I think that this code is not necessary anymore. We can also remove this flow in noobaa_init

src/cmd/core_init.js
```diff
@@ -0,0 +1,157 @@
...
+    }
+    fs.appendFileSync('/etc/passwd',
+        `${NOOBAA_USER}:x:${process.getuid()}:${process.getgid()}:,,,:/home/${NOOBAA_USER}:/bin/bash\n`);
+}
+
+/**
+ * Creates PV-backed directories required before supervisord starts.
+ */
+function prepare_server_pvs() {
+    fs.mkdirSync(SUPERVISOR_LOG_DIR, { recursive: true });
```
dannyzaken: I'm pretty sure we can get rid of this code. the only reason we do it is because we set the supervisord log file as `/log/supervisor/supervisord.log` in `setup_platform.sh`, and then we mount a the logs volume at `/log`. We can chage the supervisord log to a different path or set it to /dev/stdout.

src/cmd/core_init.js
```diff
@@ -0,0 +1,157 @@
...
+}
+
+/**
+ * Runs upgrade_manager.js for the core server init path.
+ * TODO: leaving this function the same as it used to be innoobaa_init.sh for now, but it should be moved out of core in the future.
+ */
+function run_server_upgrade() {
```
dannyzaken: Not for this PR, but as we discussed offline, this will eventually also be moved out to a separate job

src/cmd/core_init.js
```diff
@@ -0,0 +1,157 @@
...
+    return pkg.version;
+}
+
+/**
+ * Logs to stdout and to syslog with the Superd tag.
+ * @param {string} message
+ */
+function superd_log(message) {
+    console.log(message);
+    spawnSync('logger', ['-p', 'local0.info', '-t', 'Superd', message], {
```
dannyzaken: Why do we need this?

dannyzaken (commented):
A lot of what we have today in noobaa_init.sh is legacy leftovers. No need to bring that to this new file, and this is a good opportunity to get rid of it. 
I think most of the comments require very few changes, but if it's more complicated, we can do it i future PRs

src/cmd/core_init.js
nadavMiz: wanted to do this the same as it was done in supervisord.orig. but console.log should do the same, so I replaced it with console.log

src/cmd/core_init.js
```diff
@@ -0,0 +1,119 @@
...
+const DEFAULT_UPGRADE_SCRIPTS_DIR = `${NOOBAA_ROOT}/src/upgrade/upgrade_scripts`;
+
+/**
+ * Runs a child process and resolves with its exit code.
+ * @param {string} command
+ * @param {string[]} args
+ * @param {import('child_process').SpawnOptions} [options]
+ * @returns {Promise<number>}
+ */
+function spawn_async(command, args, options) {
```
dannyzaken: we already have `os_utils.spawn`, see if you can use it

dannyzaken (approved):

PR #9699, opened 2026-06-11
partial changes for role config condition attribute - unblock keycloak

sakshimunjal:
### Describe the Problem
Note: Partial PR to unblock Keycloak development

Allow condition as part of role_config 
Details: https://github.com/noobaa/noobaa-core/pull/9642/changes

### Explain the Changes
1. Added condition schema support in common_api.js
2. Extended STS policy evaluation in sts_rest.js to match conditional attributes

Gaps:
1. identity_info not populated
2. Schema changes are pending for trust policy
3. condition validation is pending

### Testing Instructions:

- [ ] Doc added/updated
- [ ] Tests added

## Release Notes

* **New Features**
  * Expanded policy condition evaluation to support additional operators (including `StringEquals` and `ForAnyValue:StringEquals`) and LDAP-based identity attribute matching.
  * Improved assume-role authorization by integrating web identity context into permission checks.
  * Enhanced condition matching to evaluate policy conditions against the provided identity context and treat statements as fitting only when conditions fully match.

### DIFF
--- a/src/endpoint/sts/sts_rest.js
+++ b/src/endpoint/sts/sts_rest.js
@@ -40,6 +40,27 @@ const STS_OPS = js_utils.deep_freeze({
     post_assume_role_with_web_identity: require('./ops/sts_post_assume_role_with_web_identity'),
 });
 
+function string_equals_predicate(user_value, policy_value) {
+    if (Array.isArray(user_value)) return user_value.includes(policy_value);
+    return user_value === policy_value;
+}
+
+function for_any_value_string_equals_predicate(user_values, policy_values) {
+    let user_arr = [];
+    if (Array.isArray(user_values)) {
+        user_arr = user_values;
+    } else if (user_values) {
+        user_arr = [user_values];
+    }
+    const policy_arr = Array.isArray(policy_values) ? policy_values : [policy_values];
+    return user_arr.some(user_value => policy_arr.includes(user_value));
+}
+
+const predicate_map = {
+    'StringEquals': string_equals_predicate,
+    'ForAnyValue:StringEquals': for_any_value_string_equals_predicate,
+};
+
 async function sts_rest(req, res) {
     try {
         await handle_request(req, res);
@@ -130,6 +151,11 @@ async function authorize_request(req) {
     await authorize_request_policy(req);
 }
 
+function fetch_web_identity_info(req) {
+    // add conditional logic to fetch web identity info from LDAP or Keycloak or default
+    return req.sts_sdk.identity_info;
+}
+
 async function authorize_request_policy(req) {
     if (req.op_name !== 'post_assume_role') return;
     const account_info = await req.sts_sdk.get_assumed_role(req);
@@ -140,7 +166,8 @@ async function authorize_request_policy(req) {
     // system owner by design can always assume role policy of any account
     // skip for NC environments since system owner is not applicable
     if (!is_nc_environment() && (cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
-    const permission = has_assume_role_permission(assume_role_policy, method, cur_account_email);
+    const web_identity_info = fetch_web_identity_info(req);
+    const permission = has_assume_role_permission(assume_role_policy, method, cur_account_email, web_identity_info);
     dbg.log0('sts_rest.authorize_request_policy permission is: ', permission);
     if (permission === 'DENY' || permission === 'IMPLICIT_DENY') {
         throw new StsError(StsError.AccessDeniedException);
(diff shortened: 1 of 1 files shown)


src/api/common_api.js
```diff
@@ -117,6 +117,37 @@ module.exports = {
             enum: ['DISABLED', 'SUSPENDED', 'ENABLED']
         },
 
+        assume_role_policy_condition_value: {
```
naveenpaul1: Aayush is working on the same please check with him

src/endpoint/sts/sts_rest.js
```diff
@@ -225,12 +224,69 @@ function _is_statements_fit(statements, method, cur_account_email) {
                 principal_fit = true;
             }
         }
-        dbg.log0('assume_role_policy: is_statements_fit', action_fit, principal_fit);
-        if (action_fit && principal_fit) return true;
+
+        // look for conditions
+        let condition_fit = true;
```
naveenpaul1, line 256: move LDAP related code to LDAP util, we need different condition check for keycloak

src/endpoint/sts/sts_rest.js
```diff
@@ -225,12 +224,69 @@ function _is_statements_fit(statements, method, cur_account_email) {
...
+            }
+        }
+
+        dbg.log0('assume_role_policy: is_statements_fit', action_fit, principal_fit, condition_fit);
+        if (action_fit && principal_fit && condition_fit) return true;
     }
     return false;
 }
 
+function _is_ldap_identity_fit(condition_key, policy_value, identity_info, predicate) {
```
naveenpaul1, line 278: Please unit test case for the possible combination for condition

src/endpoint/sts/sts_rest.js
```diff
@@ -192,20 +192,19 @@ function handle_error(req, res, err) {
...
     // look for explicit allows
-    if (_is_statements_fit(allow_statements, method, cur_account_email)) return 'ALLOW';
+    if (_is_statements_fit(allow_statements, method, cur_account_email, identity_info)) return 'ALLOW';
 
     // implicit deny
     return 'IMPLICIT_DENY';
 }
 
-function _is_statements_fit(statements, method, cur_account_email) {
+function _is_statements_fit(statements, method, cur_account_email, identity_info) {
```
naveenpaul1: here the identity_info is tag from token right?

naveenpaul1:
@sakshimunjal this wont unlock us, we are not using the `account.role_config` for keycloak, we are uning role schema for this

src/endpoint/sts/sts_rest.js
sakshimunjal: only _is_ldap_identity_fit is related to ldap, other code is generic. Please check if this works for keycloak as well.

src/endpoint/sts/sts_rest.js
sakshimunjal: will add that with ldap code, this was just generic flow so that keycloak is unblcked

src/endpoint/sts/sts_rest.js
sakshimunjal: For LDAP, identity_info is the authenticated user’s LDAP attributes, not tags from the JWT — the token only carries credentials, and attributes come from the LDAP lookup after bind.
The details will be fetched before this function call is made in case of ldap

src/endpoint/sts/sts_rest.js
naveenpaul1, line 256: it should be false

src/endpoint/sts/sts_rest.js
```diff
@@ -140,7 +140,7 @@ async function authorize_request_policy(req) {
     // system owner by design can always assume role policy of any account
     // skip for NC environments since system owner is not applicable
     if (!is_nc_environment() && (cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
-    const permission = has_assume_role_permission(assume_role_policy, method, cur_account_email);
+    const permission = has_assume_role_permission(assume_role_policy, method, cur_account_email, req.sts_sdk.identity_info);
```
naveenpaul1: what is `req.sts_sdk.identity_info`? we need `role_config` here? Can you check what we will get for `req.sts_sdk.identity_info`?

sakshimunjal: req.sts_sdk.identity_info would contain user info. Example, for LDAP the data we receive from LDAP like department/ou/memberOf in json

src/endpoint/sts/sts_rest.js
sakshimunjal: my thought here was to consider it as true and if any condition is not satisfied then mark it as false and break out of the loop at line 236, which will return false overall

naveenpaul1 (approved):

PR #9701, opened 2026-06-12
Add ability to update endpoint on backingstores

karthikps97:
### Describe the Problem
The endpoint field in the backingstore cannot be changed after its creation. This causes a problem if the user needs to change the endpoint while data is present in the backingstore.

### Explain the Changes
1. Add `endpoint` and `endpoint_type` to `update_external_connection` API definition and make `secret` as not mandatory.
2. In `update_external_connection` implementation, reject the request if backingstore is not of types `S3 compatible` or `IBM COS`.
3. Update the accounts and pools table with the new endpoint details.
4. Update the hosted agents by calling RPC `update_hosted_agents` with the new endpoint.
5. Add a new API `update_hosted_agents` which accepts credentials and endpoint details.
6. In agents class implementation rename `update_credentials ` to `update_hosted_agents` so that we have a common method to support both secret and endpoint changes.

1. 

### Testing Instructions:
1. Create a backingstore of type S3 compatible with any target endpoint.
2. Once it goes to ready phase, create a bucketclass for the backingstore.
3. Create an OBC with the newly created bucketclass.
4. Edit the backingstore using `kubectl edit` and update the endpoint. 
5. Test that the new endpoint is now accessible from the OBC.

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added an admin API to update hosted agent settings, including credentials plus optional endpoint and endpoint type.
  * Hosted agent updates now support applying S3-backed endpoint configuration.
* **Changes**
  * External connection updates now only require a name; endpoint, endpoint type, and bucket are captured as strings.
  * Endpoint changes are gated to specific connection types and validated during update.
  * Hosted-agent propagation now uses the new hosted-agent update path and conditionally includes credentials and endpoint details.
* **Bug Fixes**
  * Improved AWS/S3 endpoint validation by verifying objects exist under the expected prefix.

### DIFF
--- a/src/api/account_api.js
+++ b/src/api/account_api.js
@@ -491,7 +491,7 @@ module.exports = {
             method: 'PUT',
             params: {
                 type: 'object',
-                required: ['name', 'secret'],
+                required: ['name'],
                 properties: {
                     name: {
                         type: 'string'
@@ -501,6 +501,17 @@ module.exports = {
                     azure_log_access_keys: { $ref: 'common_api#/definitions/azure_log_access_keys' },
                     region: {
                         type: 'string'
+                    },
+                    endpoint_info: {
+                        type: 'object',
+                        properties: {
+                            endpoint: {
+                                type: 'string'
+                            },
+                            endpoint_type: {
+                                type: 'string'
+                            },
+                        }
                     }
                 }
             },
@@ -542,6 +553,9 @@ module.exports = {
                     },
                     ignore_name_already_exist: {
                         type: 'boolean'
+                    },
+                    bucket: {
+                        type: 'string'
                     }
                 }
             },
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -1,4 +1,5 @@
 /* Copyright (C) 2016 NooBaa */
+/*eslint max-lines: ["error", 2100]*/
 'use strict';
 
 const P = require('../../util/promise');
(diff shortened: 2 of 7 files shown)


src/server/system_services/account_server.js
```diff
@@ -583,22 +584,37 @@ async function update_external_connection(req) {
     const secret = req.rpc_params.secret;
     const azure_log_access_keys = req.rpc_params.azure_log_access_keys;
     const region = req.rpc_params.region || connection.region;
-
+    const bucket = req.rpc_params.bucket;
+    const new_endpoint = req.rpc_params.endpoint || connection.endpoint;
+    const new_endpoint_type = req.rpc_params.endpoint_type || connection.endpoint_type;
     const encrypted_secret = system_store.master_key_manager.encrypt_sensitive_string_with_master_key_id(
```
dannyzaken: secret is now not required. this will probably fail

src/api/account_api.js
```diff
@@ -482,7 +482,16 @@ module.exports = {
...
+                    },
+                    endpoint: {
+                        type: 'string'
+                    },
+                    endpoint_type: {
+                        type: 'string'
+                    },
+                    bucket: {
+                        type: 'string'
+                    },
```
dannyzaken: since all of these are expected to be sent together, we should maybe define it as `endpoint_info`, and then make all of them required in the object's definition

src/server/system_services/account_server.js
```diff
@@ -922,11 +958,26 @@ async function check_aws_connection(params) {
...
+        if (is_endpoint_update) {
+            const resp = await s3.listObjectsV2({
+                Bucket: params.bucket,
+                Prefix: 'noobaa_blocks/',
+                MaxKeys: 1
+            });
+
+            if (!resp?.KeyCount) {
+                throw Object.assign(new Error(`new endpoint should be a valid location used by noobaa`),
+                    { code: 'INVALID_ENDPOINT' });
```
dannyzaken: I think that this `INVALID_ENDPOINT` has no effect. the error will be translated to `UNKNOWN_FAILURE` status below and evnetually the update operation will return `INVALID_CREDENTIALS`

src/server/system_services/account_server.js
karthikps97: I added the secret back.

src/api/account_api.js
karthikps97: Done

src/server/system_services/account_server.js
karthikps97: Modified the error handling segments so that it will be passed on as is.

dannyzaken (approved):

karthikps97:
The only major difference here with the new linked operator [PR](https://github.com/noobaa/noobaa-operator/pull/2076) is, the secret is now optional with additional checks added to support this change.

PR #9711, opened 2026-06-16
Support PutRolePolicy and DeleteRolePolicy for IAM Roles

aayushchouhan09:
### Describe the Problem
NooBaa IAM had role creation/deletion, but missing inline role policy APIs (`PutRolePolicy`, `DeleteRolePolicy`).

### Explain the Changes
1. Added IAM API routing and handlers for `PutRolePolicy` and `DeleteRolePolicy` (`iam_rest` + new ops files), including malformed policy translation to `MalformedPolicyDocument`.
2. Added action constants and validation flow in `iam_constants/iam_utils` (role name, policy name, policy document checks) reusing existing policy validators.
3. Added RPC/account API methods and SDK wiring (`account_api`, `account_sdk`, `accountspace_nb`, `accountspace_fs`, `nb.d.ts`) to expose both APIs end-to-end.

// TODO
4. Trust and Role policy Validations - can be moved to separate PR.

### Testing Instructions:
1. Try putRolePolicy and deleteRolePolicy apis using below commands:

`$ AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url <endpoint-url> iam create-role --role-name test-role  --assume-role-policy-document file://trust-policy.json --path / --max-session-duration 3600` (create role)

`$ AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url <endpoint-url> iam put-role-policy --role-name test-role --policy-name allow-list-buckets --policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"s3:ListAllMyBuckets","Resource":"*"}]}'` (put role policy)

`$ AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url <endpoint-url> iam delete-role-policy --role-name test-role --policy-name allow-list-buckets` (delete role policy)

`$ AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url <endpoint-url> iam delete-role --role-name test-role` (delete role)

> $ cat trust-policy.json
{
"Version": "2012-10-17",
"Statement": [
{
"Effect": "Allow",
"Principal": {
"AWS": "arn:aws:iam::6a1ff4c334ef1400227bb371:root"
},
"Action": "sts:AssumeRole"
}
]
}

- [ ] Doc added/updated
- [ ] Tests added

## Release Notes

**New Features**
- Added `PutRolePolicy` to create or update inline policies attached to IAM roles.
- Added `DeleteRolePolicy` to remove inline policies from IAM roles.
- Both operations are restricted to administrator access, validate required parameters (including policy document content where applicable), and return structured XML responses.

**Bug Fixes**
- Improved inline-policy size-limit enforcement and clarified “no such entity” / “limit exceeded” error messaging for better troubleshooting.

### DIFF
--- a/src/endpoint/iam/iam_utils.js
+++ b/src/endpoint/iam/iam_utils.js
@@ -230,7 +230,7 @@ function validate_tagging_params(action, params) {
 }
 
 /**
- * validate_policy_params will call the aquivalent function for each action in user policy API
+ * validate_policy_params will call the equivalent function for each action in user and role policy APIs
  * @param {string} action
  * @param {object} params
  */
@@ -247,6 +247,12 @@ function validate_policy_params(action, params) {
             break;
         case iam_constants.IAM_ACTIONS.LIST_USER_POLICIES:
             validate_list_user_policies(params);
+            break;
+        case iam_constants.IAM_ACTIONS.PUT_ROLE_POLICY:
+            validate_put_role_policy(params);
+            break;
+        case iam_constants.IAM_ACTIONS.DELETE_ROLE_POLICY:
+            validate_delete_role_policy(params);
             break;
         default:
             throw new RpcError('INTERNAL_ERROR', `${action} is not supported`);
@@ -618,6 +624,38 @@ function validate_list_roles(params) {
         validate_marker(params.marker);
         validate_max_items(params.max_items);
         validate_iam_path(params.iam_path_prefix, iam_constants.IAM_ROLE_PARAMETER_NAME.IAM_PATH_PREFIX);
+    } catch (err) {
+        translate_rpc_error(err);
+    }
+}
+
+ /**
+ * validate_put_role_policy checks the params for put_role_policy action
+ * @param {object} params
+ */
+function validate_put_role_policy(params) {
+    try {
+        check_required_role_name(params);
+        validate_role_name(params.role_name, iam_constants.IAM_ROLE_PARAMETER_NAME.ROLE_NAME);
+        check_required_policy_name(params);
+        validate_policy_name(params.policy_name, iam_constants.IAM_ROLE_PARAMETER_NAME.POLICY_NAME);
+        check_required_policy_document(params);
+        validate_policy_document(params.policy_document, iam_constants.IAM_ROLE_PARAMETER_NAME.POLICY_DOCUMENT);
+    } catch (err) {
+        translate_rpc_error(err);
+    }
+}
+
+/**
+ * validate_delete_role_policy checks the params for delete_role_policy action
+ * @param {object} params
+ */
+function validate_delete_role_policy(params) {
+    try {
+        check_required_role_name(params);
+        validate_role_name(params.role_name, iam_constants.IAM_ROLE_PARAMETER_NAME.ROLE_NAME);
+        check_required_policy_name(params);
+        validate_policy_name(params.policy_name, iam_constants.IAM_ROLE_PARAMETER_NAME.POLICY_NAME);
     } catch (err) {
         translate_rpc_error(err);
     }
--- a/src/sdk/accountspace_fs.js
+++ b/src/sdk/accountspace_fs.js
@@ -623,6 +623,20 @@ class AccountSpaceFS {
         return { members, is_truncated };
     }
 
+    async put_role_policy(params, account_sdk) {
+        const action = IAM_ACTIONS.PUT_ROLE_POLICY;
+        dbg.log1(`AccountSpaceFS.${action}`, params);
+        const { code, http_code, type } = IamError.NotImplemented;
+        throw new IamError({ code, message: 'NotImplemented', http_code, type });
+    }
+
+    async delete_role_policy(params, account_sdk) {
+        const action = IAM_ACTIONS.DELETE_ROLE_POLICY;
+        dbg.log1(`AccountSpaceFS.${action}`, params);
+        const { code, http_code, type } = IamError.NotImplemented;
+        throw new IamError({ code, message: 'NotImplemented', http_code, type });
+    }
+
     ////////////////////////
     // INTERNAL FUNCTIONS //
     ////////////////////////
--- a/src/server/system_services/account_server.js
+++ b/src/server/system_services/account_server.js
@@ -1215,12 +1215,23 @@ function _verify_can_delete_account(req, account_to_delete) {
     }
 }
 
+/**
+ * returns all non-deleted IAM roles for the given owner account
+ * @param {string|nb.ID} account_id
+ * @returns {nb.IamRole[]}
+ */
 function _list_active_iam_roles_for_account(account_id) {
+    const account_id_str = String(account_id);
     const iam_roles_by_owner = system_store.data.iam_roles_by_owner || {};
-    const indexed_roles = iam_roles_by_owner[account_id] || [];
+    const indexed_roles = iam_roles_by_owner[account_id_str] || [];
     return _.filter(indexed_roles, role => !role.deleted);
 }
 
+/**
+ * validates role creation constraints for the target account
+ * @param {string} role_name
+ * @param {nb.IamRole[]} account_roles
+ */
 function _validate_create_role_preconditions(role_name, account_roles) {
     if (account_roles.length >= MAX_NUMBER_OF_IAM_ROLES) {
         const message_with_details = `Cannot exceed quota for RolesPerAccount: ${MAX_NUMBER_OF_IAM_ROLES}.`;
@@ -1234,6 +1245,12 @@ function _validate_create_role_preconditions(role_name, account_roles) {
     }
 }
 
+/**
+ * returns an IAM role by name from a role list
+ * @param {nb.IamRole[]} account_roles
+ * @param {string} role_name
+ * @returns {nb.IamRole|undefined}
+ */
 function _find_iam_role_by_name(account_roles, role_name) {
     return _.find(account_roles, role => role.name === role_name);
 }
@@ -1601,7 +1618,7 @@ async function put_user_policy(req) {
     const requesting_account = req.account;
     dbg.log1(`AccountSpaceNB.${action}`, req.rpc_params);
     const requested_account = account_util.validate_and_return_requested_account(req.rpc_params, action, requesting_account);
-    const iam_user_policies = requested_account.iam_user_policies || [];
+    const iam_user_policies = [...(requested_account.iam_user_policies || [])];
     const index_of_iam_user_policy = account_util._get_iam_user_policy_index(iam_user_policies, req.rpc_params.policy_name);
     const iam_user_policy_to_add = {
         policy_name: req.rpc_params.policy_name,
@@ -1643,7 +1660,7 @@ async function delete_user_policy(req) {
     dbg.log1(`AccountSpaceNB.${action}`, req.rpc_params);
     const requesting_account = req.account;
     const requested_account = account_util.validate_and_return_requested_account(req.rpc_params, action, requesting_account);
-    const iam_user_policies = requested_account.iam_user_policies || [];
+    const iam_user_policies = [...(requested_account.iam_user_policies || [])];
     const iam_user_policy_index = account_util._check_user_policy_exists(action, iam_user_policies, req.rpc_params.policy_name);
     iam_user_policies.splice(iam_user_policy_index, 1);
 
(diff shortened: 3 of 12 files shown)


shirady:
@aayushchouhan09 I have a couple of questions:
1. Could you share if AWS allows adding an inline policy as on a non-existing role name?

From you example - 
`$ AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url <endpoint-url> iam put-role-policy --role-name test-role --policy-name allow-list-buckets --policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"s3:ListAllMyBuckets","Resource":"*"}]}'`

What happens when `test-role` does not exist on AWS?

2. [API_GetRolePolicy](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetRolePolicy.html) is expected to be in a separate PR?

src/endpoint/iam/iam_utils.js
```diff
@@ -246,6 +246,12 @@ function validate_policy_params(action, params) {
         case iam_constants.IAM_ACTIONS.LIST_USER_POLICIES:
             validate_list_user_policies(params);
             break;
+        case iam_constants.IAM_ACTIONS.PUT_ROLE_POLICY:
+            validate_put_role_policy(params);
+            break;
+        case iam_constants.IAM_ACTIONS.DELETE_ROLE_POLICY:
+            validate_delete_role_policy(params);
+            break;
```
shirady, line 256: If you're adding this function, please update its JSDoc. Now it contains only the user:

validate_policy_params will call the aquivalent function for each action in user policy API

(and you can fix the typo).

src/endpoint/iam/iam_utils.js
```diff
@@ -569,6 +575,38 @@ function validate_delete_role(params) {
     }
 }
 
+/**
+ * validate_put_role_policy checks the params for put_role_policy action
+ * @param {object} params
+ */
+function validate_put_role_policy(params) {
```
shirady, line 636: Did you validate the order against AWS?
For example, you call the API to role policy with an invalid name without the required policy document - what error do you get?

src/sdk/accountspace_fs.js
```diff
@@ -583,6 +583,34 @@ class AccountSpaceFS {
         return { tags, is_truncated };
     }
```
shirady, line 585: Could you add a title of role before the new functions?

```
    //////////////
    // ROLE //
    //////////////
```

src/util/account_util.js
```diff
@@ -609,10 +609,10 @@ function _get_iam_user_policy_index(iam_user_policies, policy_name) {
     return iam_user_policy_index;
 }
 
-function _check_total_policy_size(iam_user_policies, username) {
+function _check_total_policy_size(iam_user_policies, username, entity = 'user') {
     const total_chars_size = _get_total_size_of_policies(iam_user_policies);
     if (total_chars_size > AWS_LIMIT_CHARS_USER_INlINE_POLICY) {
```
shirady: Could you rename the variable to better match that it is for user and role policy - `AWS_LIMIT_CHARS_USER_INlINE_POLICY`?

src/server/system_services/account_server.js
```diff
@@ -1709,6 +1709,69 @@ async function delete_role(req) {
     });
 }
 
+async function put_role_policy(req) {
+    const action = IAM_ACTIONS.PUT_ROLE_POLICY;
+    const requesting_account = req.account;
+    account_util._check_if_requesting_account_is_root_account(action, requesting_account,
+        { role_name: req.rpc_params.role_name, path: IAM_DEFAULT_PATH }, 'ROLE');
+
+    const account_roles = _list_active_iam_roles_for_account(String(requesting_account._id));
```
shirady: Why do you need to convent the `ID` to `String`?
To avoid confusion, could you add a JSDoc to the function `_list_active_iam_roles_for_account` (you wrote).

src/server/system_services/account_server.js
```diff
@@ -1709,6 +1709,69 @@ async function delete_role(req) {
...
+    const role_to_update = _find_iam_role_by_name(account_roles, req.rpc_params.role_name);
+    if (!role_to_update) {
+        const message_with_details = `The role with name ${req.rpc_params.role_name} cannot be found.`;
+        throw new RpcError('NO_SUCH_ENTITY', message_with_details);
+    }
+
+    const iam_role_policies = [...(role_to_update.iam_role_policies || [])];
```
shirady, line 1821: Could you explain this line? why do you need the spread (`...`) operator?

src/server/system_services/account_server.js
```diff
@@ -1709,6 +1709,69 @@ async function delete_role(req) {
...
+    });
+}
+
+async function delete_role_policy(req) {
+    const action = IAM_ACTIONS.DELETE_ROLE_POLICY;
+    const requesting_account = req.account;
+    account_util._check_if_requesting_account_is_root_account(action, requesting_account,
+        { role_name: req.rpc_params.role_name, path: IAM_DEFAULT_PATH }, 'ROLE');
+
+    const account_roles = _list_active_iam_roles_for_account(String(requesting_account._id));
```
shirady: Same question - Why do you need to convent the ID to String?

aayushchouhan09:
> @aayushchouhan09 I have a couple of questions:
> 
> 1. Could you share if AWS allows adding an inline policy as on a non-existing role name?
> 
> From you example - `$ AWS_ACCESS_KEY_ID=<access-key> AWS_SECRET_ACCESS_KEY=<secret-key> aws --no-verify-ssl --endpoint-url <endpoint-url> iam put-role-policy --role-name test-role --policy-name allow-list-buckets --policy-document '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"s3:ListAllMyBuckets","Resource":"*"}]}'`
> 
> What happens when `test-role` does not exist on AWS?
>
AWS rejects it: iam put-role-policy on a non-existent role returns `NoSuchEntity` (the role with that name cannot be found). So inline policy attachment is allowed only for an existing role.
> 2. [API_GetRolePolicy](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetRolePolicy.html) is expected to be in a separate PR?

Yes we will do other remaining apis in separate PRs. Firstly, we are trying to do the required apis to test keycloak and ldap.

src/endpoint/iam/iam_utils.js
aayushchouhan09: Yes, I have verified this and all the errors are aws complaint. We are also planning to create a separate PR to cover this validations in unit tests (may be after completing the apis we can work on it).

src/endpoint/iam/iam_utils.js
aayushchouhan09: done, thanks

src/sdk/accountspace_fs.js
aayushchouhan09: done, thanks

src/util/account_util.js
aayushchouhan09: done, thanks

src/server/system_services/account_server.js
aayushchouhan09: The spread operator ([...]) clones `iam_role_policies` before mutation so we don’t modify the in-memory role object directly prior to `make_changes`, preventing pre-commit state leakage on failures.

src/server/system_services/account_server.js
aayushchouhan09: We normalize account_id with `String(account_id)` because `iam_roles_by_owner` is a plain object map keyed by stringified IDs, so this keeps lookups consistent and avoids ObjectId-vs-string type ambiguity.

src/server/system_services/account_server.js
aayushchouhan09: updated thanks.

src/server/system_services/account_server.js
shirady: I didn't understand. Why do we need this? If you can give a simple case.
I don't understand what is wrong in modifying the `iam_role_policies` in memory.

aayushchouhan09: `iam_role_policies` is the same live object used by system_store, so changing it directly updates memory before the request is fully validated/saved.
Simple case: you add a policy, then `_check_total_policy_size()` fails — API returns error, DB is not updated, but memory may still show that new policy (temporary wrong state).

shirady: `account_roles` is returned from `_list_active_iam_roles_for_account`:
https://github.com/noobaa/noobaa-core/blob/2edf8f2d22cc1b0bb9a134078a701b3cb20da61e/src/server/system_services/account_server.js#L1218-L1222
 
The implementation uses lodash `filter` which returns (Lodash doc [link](https://lodash.com/docs/4.18.1#filter))

> "Returns the new filtered array"

So why do you need to copy this again?

aayushchouhan09: `_.filter()` only creates a new outer array; each role inside is still the same live object from system_store, so mutating `role.iam_role_policies` mutates shared in-memory state immediately.
If validation (like total policy size) fails after that mutation, DB update is rejected but memory may already contain the change, causing temporary inconsistency.

shirady: I think it was missed in `put_user_policy` and `delete_user_policy` the matching operations for IAM inline policy in users (in this PR, it is in roles). If those can be updated to align with the explanation, it would be great.

aayushchouhan09: Updated, Thanks

shirady (approved):
LGTM

src/server/system_services/account_server.js
```diff
@@ -1709,6 +1726,69 @@ async function delete_role(req) {
     });
 }
 
+async function put_role_policy(req) {
+    const action = IAM_ACTIONS.PUT_ROLE_POLICY;
+    const requesting_account = req.account;
+    account_util._check_if_requesting_account_is_root_account(action, requesting_account,
+        { role_name: req.rpc_params.role_name, path: IAM_DEFAULT_PATH }, 'ROLE');
+
+    const account_roles = _list_active_iam_roles_for_account(requesting_account._id);
```
naveenpaul1, line 1814: who can put_role_policy in root and role owner , right?

src/server/system_services/account_server.js
```diff
@@ -1709,6 +1726,69 @@ async function delete_role(req) {
...
+}
+
+async function delete_role_policy(req) {
+    const action = IAM_ACTIONS.DELETE_ROLE_POLICY;
+    const requesting_account = req.account;
+    account_util._check_if_requesting_account_is_root_account(action, requesting_account,
+        { role_name: req.rpc_params.role_name, path: IAM_DEFAULT_PATH }, 'ROLE');
+
+    const account_roles = _list_active_iam_roles_for_account(requesting_account._id);
+    const role_to_update = _find_iam_role_by_name(account_roles, req.rpc_params.role_name);
```
naveenpaul1: role_to_delete make sense

src/endpoint/iam/iam_utils.js
```diff
@@ -571,6 +577,38 @@ function validate_delete_role(params) {
...
+    try {
+        check_required_role_name(params);
```
naveenpaul1, line 638: are we checking access for action here? only root and owner can put_role_policy, right?

src/server/system_services/account_server.js
aayushchouhan09: For this basic IAM implementation, put-role-policy is restricted to the root/owner account (account created via cli) only (checked in account_server), not a separate role-owner principal.

src/endpoint/iam/iam_utils.js
aayushchouhan09: Access is intentionally not checked in `iam_utils` validation following noobaa pattern; authorization is enforced in `account_server` via `_check_if_requesting_account_is_root_account` (root-only in this PR)

naveenpaul1 (approved):
LGTM

PR #9713, opened 2026-06-17
support condition attribute for ldap

sakshimunjal:
### Describe the Problem
LDAP AssumeRoleWithWebIdentity allowed any authenticated user to assume a role when assume_role_policy used principal: ["*"], with no way to restrict by LDAP attributes (e.g. OU or group membership).

assume_role_policy also had no schema support for condition, and STS did not evaluate conditions during authorization for Web Identity requests.

### Explain the Changes
1. Added optional condition on assume_role_policy.statement
2. Evaluate assume_role_policy conditions for AssumeRoleWithWebIdentity
3. Make identity info available before authorize
4. Return identity attributes from authenticate() instead of DN only, for condition matching.

1. LDAP AssumeRoleWithWebIdentity lacks attribute-based access control on assume_role_policy

### Testing Instructions:
1. Configure LDAP and create a role account with conditions
`sudo node src/cmd/manage_nsfs.js account update --name ldap_role \
  --role_config '{"role_name":"ldap_user","assume_role_policy":{"statement":[{"effect":"allow","principal":["*"],"action":["sts:AssumeRoleWithWebIdentity"],"condition":{"StringEquals":{"ldap:ou":"Delivering Crew"}}}]}}'`
2. Build a Web Identity JWT with LDAP user/password claims and call:
`aws sts assume-role-with-web-identity \
  --endpoint https://127.0.0.1:7443 \
  --role-arn "arn:aws:sts::<role-access-key>:role/ldap_user" \
  --role-session-name test1 \
  --web-identity-token "$TOKEN" \
  --no-verify-ssl`
3. Verify different conditions as per user info
Refer: docs/ldap_group_assume_role_proposal.md

- [ ] Doc added/updated
- [ ] Tests added

## Release Notes

* **New Features**
  * Added support for IAM role trust policy `Condition` blocks in policy schema validation.

* **Improvements**
  * Updated STS request authentication to run asynchronously and strengthened web-identity assume-role handling for the web-identity flow.
  * Broadened assume-role authorization checks to include the web-identity operation.
  * Improved web-identity error handling (invalid/expired tokens) and clarified access-denied mapping.

* **Improvements (Identity Data)**
  * Enhanced LDAP web-identity resolution to return richer identity attributes (including group membership, user/org details, and email).

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -143,6 +143,9 @@ module.exports = {
                                 items: {
                                     $ref: '#/definitions/email',
                                 }
+                            },
+                            condition: {
+                                $ref: '#/definitions/trust_policy_condition'
                             }
                         }
                     }
@@ -628,6 +631,11 @@ module.exports = {
                 }
             }]
         },
+        trust_policy_condition: {
+            type: 'object',
+            additionalProperties: true,
+            properties: {}
+        },
         iam_trust_policy_statement: {
             type: 'object',
             additionalProperties: false,
@@ -647,9 +655,7 @@ module.exports = {
                     $ref: '#/definitions/iam_trust_policy_principal'
                 },
                 Condition: {
-                    type: 'object',
-                    additionalProperties: true,
-                    properties: {}
+                    $ref: '#/definitions/trust_policy_condition'
                 },
                 // unsupported in role trust policy statements
                 NotAction: false,
--- a/src/endpoint/sts/sts_rest.js
+++ b/src/endpoint/sts/sts_rest.js
@@ -22,7 +22,8 @@ const RPC_ERRORS_TO_STS = Object.freeze({
     INVALID_ACCESS_KEY_ID: StsError.AccessDeniedException,
     DEACTIVATED_ACCESS_KEY_ID: StsError.AccessDeniedException,
     NO_SUCH_ACCOUNT: StsError.AccessDeniedException,
-    NO_SUCH_ROLE: StsError.AccessDeniedException
+    NO_SUCH_ROLE: StsError.AccessDeniedException,
+    ACCESS_DENIED: StsError.AccessDeniedException,
 });
 
 const ACTIONS = Object.freeze({
@@ -116,7 +117,7 @@ async function handle_request(req, res) {
     req.op_name = op_name;
 
     http_utils.authorize_session_token(req, headers_options);
-    authenticate_request(req);
+    await authenticate_request(req);
     await authorize_request(req);
 
     dbg.log1('STS REQUEST', req.method, req.originalUrl, 'op', op_name, 'request_id', req.request_id, req.headers);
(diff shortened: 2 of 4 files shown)


src/endpoint/sts/sts_rest.js
```diff
@@ -157,7 +163,6 @@ function fetch_web_identity_info(req) {
 }
 
 async function authorize_request_policy(req) {
-    if (req.op_name !== 'post_assume_role') return;
```
naveenpaul1, line 160: may be we need to allow only post_assume_role and post_assume_role_with_web_identity

src/sdk/sts_sdk.js
```diff
@@ -135,20 +131,31 @@ class StsSDK {
         if (!(await ldap_client.is_ldap_configured()) || !ldap_client.instance().is_connected()) {
             throw new RpcError('ACCESS_DENIED', 'LDAP is not configured or not connected');
         }
-        let dn;
+        let ldap_auth_result = {};
         try {
-            dn = await ldap_client.instance().authenticate(ldap_user, ldap_password);
+            ldap_auth_result = await ldap_client.instance().authenticate(ldap_user, ldap_password);
```
naveenpaul1, line 136: What will happen authenticate() return with error response? Should we check and show appropriate error message?

src/util/ldap_client.js
```diff
@@ -117,11 +117,12 @@ class LdapClient extends EventEmitter {
             attribute: this.ldap_params.dn_attribute,
             value: user,
         });
+        const attrs = ['dn', 'ou', 'memberOf', 'uid', 'cn', 'mail'];
         /** @type {ldap.SearchOptions} */
         const search_options = {
             filter: eqFilter,
             scope: this.ldap_params.search_scope,
-            attributes: ['dn']
+            attributes: attrs
```
naveenpaul1, line 125: Why are we setting these attributes also to our request? `'ou', 'memberOf', 'uid', 'cn', 'mail'`

src/sdk/sts_sdk.js
```diff
@@ -135,20 +131,31 @@ class StsSDK {
...
+        const ldap_auth_result = await this.authenticate_web_identity(req);
         const account = await this._assume_role(req.body.role_arn);
         dbg.log0('sts_sdk.get_assumed_role_with_web_identity res', account,
             'account.role_config: ', account.role_config);
         return {
             access_key: req.body.role_arn.split(':')[4],
             role_config: account.role_config,
-            dn,
+            dn: ldap_auth_result.dn,
+            ldap_auth_result,
```
naveenpaul1: I dont think we need to get and return all the LDAP user values, WDYT?

src/endpoint/sts/sts_rest.js
sakshimunjal: As per the current conditions both mean the same, but I'll add the explicit guard for future-proofing.

src/sdk/sts_sdk.js
sakshimunjal: Tested locally, both wrong username and wrong password return InternalFailure instead of AccessDeniedException. Will add ACCESS_DENIED: StsError.AccessDeniedException to RPC_ERRORS_TO_STS so it returns the correct error

src/sdk/sts_sdk.js
sakshimunjal: thanks, missed this after extracting out authenticate_web_identity

src/util/ldap_client.js
sakshimunjal: Added common attributes that can be used as condition key. Without fetching them during authentication, condition evaluation wouldn't be possible.

naveenpaul1 (approved):
LGTM,
Aayush will check the Schema side

src/api/common_api.js
```diff
@@ -628,6 +631,11 @@ module.exports = {
                 }
             }]
         },
+        iam_trust_policy_condition: {
```
aayushchouhan09: you can use some common name here `trust_policy_condition` may be

aayushchouhan09 (commented):
one small comment

PR #9735, opened 2026-06-19
KeyCloak | Assume role after validating session tags

naveenpaul1:
### Explain the Changes
1. Trust policy condition evaluation (`StringEquals`, `ForAnyValue:StringEquals`)  added for assume-role with web identity
2. principal.Federated values is validated against issuer in token inside pronciple check
3. Code refactoring for LDAP condition check methods
4. `_assume_role` method updated to accept account_id and and access_key and role policy can be fetch from account.role_config and role schema
5. Token verification with signin key is removed, we will be using only keycloak introspect method to validated the JWT token
6. ARN added in STS principle validation
7. Remove test inside `Assume role policy tests`, all the test inside this block was testing the account creation with `role_config` with different set of inputs(valida and invalid), And this test cases do not make sense after we implement IAM role APIs to create roles

**Improvements:** 
1. STS authenticate_request() updated with authenticate_web_identity() only for LDAP

**GAPs**

1. Code duplication in sts rest `keycloak_predicate_map` with what we already have in access_policy_utils, Need check already existing code can be used or not
2. Unit and integration test for `assume-role` and `assume-role-with-web-identity`.

1. https://redhat.atlassian.net/browse/RHSTOR-8930

### Testing Instructions:
1.  npx jest /src/test/unit_tests/util_functions_tests/test_iam_policy_utils.test.js

- [ ] Doc added/updated
- [X] Tests added

## Release Notes

* **New Features**
  * Enhanced STS assume-role with web identity using AWS-style trust policy condition evaluation (e.g., `StringEquals` and `ForAnyValue:StringEquals`).

* **Improvements**
  * Improved web-identity handling to distinguish LDAP vs OIDC-style requests and evaluate assume-role permissions using AWS-like principals/actions/conditions.
  * Improved role resolution behavior and clearer errors when roles can’t be found.

* **Bug Fixes**
  * STS web-identity responses no longer include session tags.

* **Tests**
  * Added unit coverage for trust-policy condition matching and tag-key extraction.

### DIFF
--- a/src/endpoint/sts/ops/sts_post_assume_role.js
+++ b/src/endpoint/sts/ops/sts_post_assume_role.js
@@ -33,6 +33,7 @@ async function assume_role(req) {
         AssumeRoleResponse: {
             AssumeRoleResult: {
                 AssumedRoleUser: {
+                    // TODO: change the role arn with account id
                     Arn: `arn:aws:sts::${assumed_role.access_key}:assumed-role/${assumed_role.role_config.role_name}/${req.body.role_session_name}`,
                     AssumedRoleId: `${assumed_role.access_key}:${req.body.role_session_name}`
                 },
--- a/src/endpoint/sts/sts_rest.js
+++ b/src/endpoint/sts/sts_rest.js
@@ -2,6 +2,8 @@
 'use strict';
 
 const _ = require('lodash');
+const jwt = require('jsonwebtoken');
+
 const dbg = require('../../util/debug_module')(__filename);
 const StsError = require('./sts_errors').StsError;
 const js_utils = require('../../util/js_utils');
@@ -9,6 +11,7 @@ const http_utils = require('../../util/http_utils');
 const signature_utils = require('../../util/signature_utils');
 const system_store = require('../../server/system_services/system_store').get_instance();
 const { is_nc_environment } = require('../../nc/nc_utils');
+const access_policy_utils = require('../../util/access_policy_utils');
 
 const STS_MAX_BODY_LEN = 4 * 1024 * 1024;
 
@@ -41,26 +44,6 @@ const STS_OPS = js_utils.deep_freeze({
     post_assume_role_with_web_identity: require('./ops/sts_post_assume_role_with_web_identity'),
 });
 
-function string_equals_predicate(user_value, policy_value) {
-    if (Array.isArray(user_value)) return user_value.includes(policy_value);
-    return user_value === policy_value;
-}
-
-function for_any_value_string_equals_predicate(user_values, policy_values) {
-    let user_arr = [];
-    if (Array.isArray(user_values)) {
-        user_arr = user_values;
-    } else if (user_values) {
-        user_arr = [user_values];
-    }
-    const policy_arr = Array.isArray(policy_values) ? policy_values : [policy_values];
-    return user_arr.some(user_value => policy_arr.includes(user_value));
-}
-
-const predicate_map = {
-    'StringEquals': string_equals_predicate,
-    'ForAnyValue:StringEquals': for_any_value_string_equals_predicate,
-};
 
 async function sts_rest(req, res) {
     try {
@@ -134,8 +117,12 @@ async function authenticate_request(req) {
     try {
         signature_utils.authenticate_request_by_service(req, req.sts_sdk);
         if (req.op_name === 'post_assume_role_with_web_identity') {
-            // fetch LDAP identity info
-            req.sts_sdk.identity_info = await req.sts_sdk.authenticate_web_identity(req);
+            const web_identity_info = fetch_web_identity_info(req);
+            const is_ldap_request = web_identity_info.username;
+            if (is_ldap_request) {
+                // fetch LDAP identity info
+                req.sts_sdk.identity_info = await req.sts_sdk.authenticate_web_identity(req);
+            }
         }
     } catch (err) {
         dbg.error('authenticate_request: ERROR', err.stack || err);
@@ -158,15 +145,10 @@ async function authorize_request(req) {
     await authorize_request_policy(req);
 }
 
-function fetch_web_identity_info(req) {
-    // add conditional logic to fetch web identity info from LDAP or Keycloak or default
-    return req.sts_sdk.identity_info;
-}
-
 async function authorize_request_policy(req) {
     if (req.op_name !== 'post_assume_role' && req.op_name !== 'post_assume_role_with_web_identity') return;
-    const account_info = await req.sts_sdk.get_assumed_role(req);
-    const assume_role_policy = account_info.role_config && account_info.role_config.assume_role_policy;
+
+    const assume_role_policy = await get_assume_role_policy(req);
     if (!assume_role_policy) throw new StsError(StsError.AccessDeniedException);
     const method = _get_method_from_req(req);
     const cur_account_email = req.sts_sdk.requesting_account && req.sts_sdk.requesting_account.email.unwrap();
@@ -174,7 +156,7 @@ async function authorize_request_policy(req) {
     // skip for NC environments since system owner is not applicable
     if (!is_nc_environment() && (cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
     const web_identity_info = fetch_web_identity_info(req);
-    const permission = has_assume_role_permission(assume_role_policy, method, cur_account_email, web_identity_info);
+    const permission = has_assume_role_permission(assume_role_policy, method, req.sts_sdk.requesting_account, web_identity_info);
     dbg.log0('sts_rest.authorize_request_policy permission is: ', permission);
     if (permission === 'DENY' || permission === 'IMPLICIT_DENY') {
         throw new StsError(StsError.AccessDeniedException);
@@ -233,55 +215,33 @@ function handle_error(req, res, err) {
     res.end(reply);
 }
 
-function has_assume_role_permission(policy, method, cur_account_email, web_identity_info) {
-    const [allow_statements, deny_statements] = _.partition(policy.statement, statement => statement.effect === 'allow');
-
+function has_assume_role_permission(policy, method, account, web_identity_info) {
+    const [allow_statements, deny_statements] = _.partition(policy.Statement, statement => statement.Effect === 'Allow');
     // look for explicit denies
-    if (_is_statements_fit(deny_statements, method, cur_account_email, web_identity_info)) return 'DENY';
+    if (_is_statements_fit(deny_statements, method, account, web_identity_info)) return 'DENY';
     // look for explicit allows
-    if (_is_statements_fit(allow_statements, method, cur_account_email, web_identity_info)) return 'ALLOW';
+    if (_is_statements_fit(allow_statements, method, account, web_identity_info)) return 'ALLOW';
 
     // implicit deny
     return 'IMPLICIT_DENY';
 }
 
-function _is_statements_fit(statements, method, cur_account_email, web_identity_info) {
+function _is_statements_fit(statements, method, account, web_identity_info) {
     for (const statement of statements) {
         let action_fit = false;
         let principal_fit = false;
         dbg.log0('assume_role_policy: statement', statement);
 
         // what action can be done
-        for (const action of statement.action) {
-            dbg.log0('assume_role_policy: action fit?', action, method);
-            if ((action === '*') || (action === 'sts:*') || (action === method)) {
-                action_fit = true;
-            }
-        }
-        // who can do that action
-        for (const principal of statement.principal) {
-            dbg.log0('assume_role_policy: principal fit?', principal.unwrap().toString(), cur_account_email);
-            if ((principal.unwrap() === cur_account_email) || (principal.unwrap() === '*')) {
-                principal_fit = true;
-            }
-        }
 
-        // look for conditions
-        let condition_fit = true;
-        if (statement.condition) {
-            let statement_conditions = statement.condition;
-            // if the condition is an object, wrap it in an array
-            if (!Array.isArray(statement.condition)) {
-                statement_conditions = [statement.condition];
-            }
-            for (const condition of statement_conditions) {
-                condition_fit = _is_identity_condition_fit(condition, web_identity_info);
-                if (!condition_fit) {
-                    condition_fit = false;
-                    break;
-                }
-            }
-        }
+        action_fit = _is_action_fit(method, statement);
+        dbg.log0('assume_role_policy: action fit?', action_fit);
+
+        principal_fit = _is_principal_fit(statement, account, web_identity_info);
+        dbg.log0('assume_role_policy: principal fit?', principal_fit);
+
+        const condition_fit = _is_condition_fit(statement.Condition, web_identity_info);
+        dbg.log0('assume_role_policy: condition fit?', condition_fit);
 
         dbg.log0('assume_role_policy: is_statements_fit', action_fit, principal_fit, condition_fit);
         if (action_fit && principal_fit && condition_fit) return true;
@@ -289,27 +249,117 @@ function _is_statements_fit(statements, method, cur_account_email, web_identity_
     return false;
 }
 
-function _is_ldap_identity_fit(condition_key, policy_value, identity_info, predicate) {
-    const ldap_attr = condition_key.slice('ldap:'.length);
-    const user_value = identity_info && identity_info[ldap_attr];
-    return predicate(user_value, policy_value);
+/**
+ * _is_principal_fit checks if the statement can be applied to the account
+ * @param {Object | Array} statement - The condition(s) from the policy statement
+ * @param {Object} account - The account to validate against
+ * @param {Object} web_identity_info - The web identity information to validate against
+ * @returns {boolean} - true if all principle are satisfied, false otherwise
+ */
+function _is_principal_fit(statement, account, web_identity_info) {
+    const statement_principal = statement.Principal || statement.NotPrincipal;
+
+        let principal_fit = false;
+        if (statement_principal.Federated) {
+            for (const federated of _.flatten([statement_principal.Federated])) {
+                const federated_url = typeof federated === 'string' ? federated : federated.unwrap();
+                dbg.log0('assume_role_policy: principal federated fit?', federated_url, web_identity_info.iss);
+                if (federated_url.split("oidc-provider/")[1] === web_identity_info.iss?.split('//')[1]) {
+                    principal_fit = true;
+                }
+            }
+        }
+
+        if (statement_principal.AWS) {
+            for (const principal_aws of _.flatten([statement_principal.AWS])) {
+                const principal = typeof principal_aws === 'string' ? principal_aws : principal_aws.unwrap();
+                const cur_account_email = account?.email.unwrap();
+                // Added ARN to principle validation
+                const account_identifier_arn = account ? access_policy_utils.get_bucket_policy_principal_arn(account) : "";
+                dbg.log0('assume_role_policy: principal fit?', principal, cur_account_email, account_identifier_arn);
+                if ((principal === cur_account_email) || (principal === '*') || (principal === account_identifier_arn)) {
+                    principal_fit = true;
+                }
+            }
+        }
+
+    return statement.Principal ? principal_fit : !principal_fit;
 }
 
-function _is_identity_condition_fit(condition, web_identity_info) {
-    for (const [operator, condition_statements] of Object.entries(condition || {})) {
-        const predicate = predicate_map[operator];
-        if (!predicate) return false; // unsupported operator
 
-        for (const [condition_key, policy_value] of Object.entries(condition_statements)) {
-            if (condition_key.startsWith('ldap:')) { // LDAP identity condition
-                if (!_is_ldap_identity_fit(condition_key, policy_value, web_identity_info, predicate)) return false;
-            } else {
-                return false;
-            }
+/**
+ * _is_action_fit checks if the method is allowed by the statement
+ * @param {String} method - The account to validate against
+ * @param {Object|Array} statement - The condition(s) from the policy statement
+ * @returns {boolean} - true if all method are satisfied, false otherwise
+ */
+function _is_action_fit(method, statement) {
+    const statement_action = statement.Action || statement.NotAction;
+    let action_fit = false;
+    for (const action of _.flatten([statement_action])) {
+        dbg.log1('access_policy: ', statement.Action ? 'Action' : 'NotAction', ' fit?', action, method);
+        if (action === method || access_policy_utils._is_wildcard_match(action, method)) {
+            action_fit = true;
+            break;
+        }
+    }
+    return statement.Action ? action_fit : !action_fit;
+}
+
+
+/**
+ * _is_condition_fit checks if the statement conditions match the web identity info
+ * @param {Object|Array} statement_condition - The condition(s) from the policy statement
+ * @param {Object} web_identity_info - The web identity information to validate against
+ * @returns {boolean} - true if all conditions are satisfied, false otherwise
+ */
+function _is_condition_fit(statement_condition, web_identity_info) {
+    if (!statement_condition) return true;
+
+    let statement_conditions = statement_condition;
+    // if the condition is an object, wrap it in an array
+    if (!Array.isArray(statement_condition)) {
+        statement_conditions = [statement_condition];
+    }
+    const is_keycloak_request = web_identity_info.iss;
+    for (const condition of statement_conditions) {
+        const condition_fit = access_policy_utils._is_identity_condition_fit(is_keycloak_request, condition, web_identity_info);
+        if (!condition_fit) {
+            return false;
         }
     }
     return true;
 }
+
+/**
+ * fetch web identity object from request web_identity_token param
+ * @param {Object} req - Request object
+ * @returns {Object} - web_identity_info
+ */
+function fetch_web_identity_info(req) {
+    let web_identity_info;
+    if (req.body.web_identity_token) {
+        web_identity_info = jwt.decode(req.body.web_identity_token, { json: true });
+    }
+    return web_identity_info || {};
+}
+
+/**
+ * get_assume_role_policy retrieves the assume role policy document
+ * @param {Object} req - Request object
+ * @returns {Promise<Object>} - Assume role policy document
+ */
+async function get_assume_role_policy(req) {
+    const role_arn = req.body.role_arn;
+    const role_name = role_arn.slice(role_arn.lastIndexOf('/') + 1);
+    // TODO: Get the iam_role from cache
+    const iam_role = _.find(system_store.data.iam_roles || [], role => {
+        if (role.deleted) return false;
+        return role.name === role_name;
+    });
+    return iam_role?.assume_role_policy_document;
+}
+
 
 // EXPORTS
 module.exports = sts_rest;
--- a/src/sdk/sts_sdk.js
+++ b/src/sdk/sts_sdk.js
@@ -1,6 +1,8 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
+const _ = require('lodash');
+
 const cloud_utils = require('../util/cloud_utils');
 const dbg = require('../util/debug_module')(__filename);
 const { RpcError } = require('../rpc');
@@ -8,9 +10,9 @@ const signature_utils = require('../util/signature_utils');
 const { account_cache, dn_cache } = require('./object_sdk');
 const BucketSpaceNB = require('./bucketspace_nb');
 const jwt = require('jsonwebtoken');
+const system_store = require('../server/system_services/system_store').get_instance();
 const ldap_client = require('../util/ldap_client');
 const keycloak_client = require('../util/keycloak_client');
-const { extract_session_tags } = require('../util/keycloak_utils');
 
 class StsSDK {
 
@@ -70,32 +72,34 @@ class StsSDK {
     async _assume_role(role_arn) {
         const role_name_idx = role_arn.lastIndexOf('/') + 1;
         const role_name = role_arn.slice(role_name_idx);
-        const access_key = role_arn.split(':')[4];
-
-        const account = await account_cache.get_with_cache({
-            bucketspace: this._get_bucketspace(),
-            access_key,
+        const account_id = role_arn.split(':')[4];
+        // TODO: Use cache instead of system_store
+        const iam_role = _.find(system_store.data?.iam_roles || [], role => {
+            if (role.deleted) return false;
+            return role.name === role_name && role.owner._id.toString() === account_id;
         });
-        if (!account) {
-            throw new RpcError('NO_SUCH_ACCOUNT', 'No such account with access_key: ' + access_key);
+        if (!iam_role) {
+            throw new RpcError('NO_SUCH_ROLE', `No such Role found with name: ${role_name} and account id : ${account_id}`);
         }
-        if (!account.role_config || account.role_config.role_name !== role_name) {
-            throw new RpcError('NO_SUCH_ROLE', `Role not found`);
-        }
-        dbg.log0('sts_sdk.get_assumed_role res', account,
-            'account.role_config: ', account.role_config);
-
-        return account;
+        const account = iam_role.owner;
+        dbg.log1('sts_sdk.get_assumed_role res', 'iam_role: ', iam_role.name);
+        return {
+            ...iam_role,
+            role_name: iam_role.name,
+            account_id: account._id.toString(),
+            access_key: account.access_keys[0].access_key.unwrap(),
+            assume_role_policy: iam_role.assume_role_policy,
+        };
     }
 
     async get_assumed_role(req) {
         dbg.log1('sts_sdk.get_assumed_role body', req.body);
-        // arn:aws:sts::access_key:role/role_name
-        const account = await this._assume_role(req.body.role_arn);
+        const role_config = await this._assume_role(req.body.role_arn);
 
         return {
-            access_key: req.body.role_arn.split(':')[4],
-            role_config: account.role_config
+            account_id: role_config.account_id,
+            access_key: role_config.access_key,
+            role_config,
         };
     }
     async authenticate_web_identity(req) {
@@ -148,12 +152,11 @@ class StsSDK {
      */
     async get_assumed_ldap_user(req) {
         const ldap_auth_result = await this.authenticate_web_identity(req);
-        const account = await this._assume_role(req.body.role_arn);
-        dbg.log0('sts_sdk.get_assumed_role_with_web_identity res', account,
-            'account.role_config: ', account.role_config);
+        const role_config = await this._assume_role(req.body.role_arn);
+        dbg.log0('sts_sdk.get_assumed_role_with_web_identity res', 'account.role_config: ', role_config);
         return {
             access_key: req.body.role_arn.split(':')[4],
-            role_config: account.role_config,
+            role_config,
             dn: ldap_auth_result.dn,
         };
     }
@@ -173,10 +176,8 @@ class StsSDK {
             if (!keycloak_instance.initialized) {
                 await keycloak_instance.initialize();
             }
-            // Also verify the JWT signature for additional security
-            const verified_token = await keycloak_instance.verify_token(
-                req.body.web_identity_token
-            );
+            // JWT token decoded and check token issuer is in provider list
+            await keycloak_instance.verify_token(req.body.web_identity_token);
 
             // Introspect token with Keycloak using client_id, client_secret, and access_token
             // This is the key implementation for Keycloak - validates token is active and not revoked
@@ -184,26 +185,23 @@ class StsSDK {
                 req.body.web_identity_token
             );
 
-            // Extract session tags from verified token (not from introspection)
-            const session_tags = extract_session_tags(verified_token);
             // Assume role
-            const account = await this._assume_role(req.body.role_arn);
-            dbg.log1('sts_sdk.get_assumed_oidc_user _assume_role res', account,
-                'account.role_config:', account.role_config,
-                'session_tags:', session_tags);
+            const role_config = await this._assume_role(req.body.role_arn);
+            dbg.log1('sts_sdk.get_assumed_oidc_user _assume_role res',
+                'account.role_config:', role_config);
 
             return {
-                access_key: req.body.role_arn.split(':')[4],
-                role_config: account.role_config,
+                access_key: role_config.access_key,
+                role_config,
                 sub: introspection_resp.sub,
                 aud: introspection_resp.client_id || introspection_resp.aud,
                 iss: introspection_resp.iss,
-                session_tags,
                 // Store additional claims for audit
-                email: introspection_resp.email || verified_token.email,
-                name: introspection_resp.name || verified_token.name || verified_token.preferred_username,
+                email: introspection_resp.email,
+                name: introspection_resp.name,
             };
         } catch (err) {
+            // TODO: Create error map mapKeycloakError
             dbg.error('get_assumed_oidc_user error:', err);
             if (err.name === 'TokenExpiredError') {
                 throw new RpcError('EXPIRED_WEB_IDENTITY_TOKEN', err.message);
--- a/src/test/integration_tests/api/sts/test_sts.js
+++ b/src/test/integration_tests/api/sts/test_sts.js
@@ -2,9 +2,10 @@
 'use strict';
 
 // setup coretest first to prepare the env
-const { require_coretest, is_nc_coretest } = require('../../../system_tests/test_utils');
+const { require_coretest, is_nc_coretest, generate_iam_client } = require('../../../system_tests/test_utils');
 const coretest = require_coretest();
 coretest.setup();
+// TODO: Migrate the AWS package to lates on, avoid importing multiple places line #24   
 const AWS = require('aws-sdk');
 const https = require('https');
 const path = require('path');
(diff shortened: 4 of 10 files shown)


src/util/iam_role_utils.js
```diff
@@ -0,0 +1,76 @@
...
+        }
+        for (const [condition_key, expected_value] of Object.entries(condition_statements)) {
+            if (is_keycloak_request && !predicate({ [condition_key]: expected_value }, evaluation_context)) {
+                dbg.log0('_is_identity_condition_fit: Condition validation failed for operator:', operator,
+                    'condition_key:', condition_key);
+                return false;
+            }
+            if (condition_key.startsWith('ldap:')) { // LDAP identity condition
+                if (!_is_ldap_identity_fit(condition_key, expected_value, web_identity_info, predicate)) return false;
+            }
```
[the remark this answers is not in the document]
naveenpaul1: will add in different PR

src/endpoint/sts/ops/sts_post_assume_role_with_web_identity.js
```diff
@@ -15,6 +15,7 @@ async function assume_role_with_web_identity(req) {
     dbg.log0('sts_post_assume_role_with_web_identity body: ', _.omit(req.body, 'web_identity_token'));
     const duration_ms = sts_utils.parse_sts_duration(req.body.duration_seconds);
     const duration_sec = Math.ceil(duration_ms / 1000);
+    // TODO: default values is max_session_duration in role, if duration_seconds is set pick it only if it is less than max_session_duration
     const expiration_time = Date.now() + duration_ms;
```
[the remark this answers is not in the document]
naveenpaul1: will add it in different PR

src/endpoint/sts/sts_rest.js
```diff
@@ -133,7 +118,9 @@ async function handle_request(req, res) {
 async function authenticate_request(req) {
     try {
         signature_utils.authenticate_request_by_service(req, req.sts_sdk);
-        if (req.op_name === 'post_assume_role_with_web_identity') {
+        const web_identity_info = fetch_web_identity_info(req);
+        const is_ldap_request = web_identity_info.password;
```
sakshimunjal: we might have to update this when we move to federated principal. There is a field 'type': 'ldap' as well, i'm not sure if it comes with each request though

src/endpoint/sts/sts_rest.js
```diff
@@ -159,22 +146,31 @@ async function authorize_request(req) {
 }
 
 function fetch_web_identity_info(req) {
-    // add conditional logic to fetch web identity info from LDAP or Keycloak or default
-    return req.sts_sdk.identity_info;
+    let web_identity_info;
+    if (req.body.web_identity_token) {
+        web_identity_info = jwt.decode(req.body.web_identity_token, { json: true });
+    }
+    return web_identity_info || {};
```
sakshimunjal: we should return req.sts_sdk.identity_info by default

src/endpoint/sts/sts_rest.js
```diff
@@ -159,22 +146,31 @@ async function authorize_request(req) {
...
+    const assume_role_policy = await get_assume_role_policy(req, web_identity_info);
     if (!assume_role_policy) throw new StsError(StsError.AccessDeniedException);
     const method = _get_method_from_req(req);
     const cur_account_email = req.sts_sdk.requesting_account && req.sts_sdk.requesting_account.email.unwrap();
     // system owner by design can always assume role policy of any account
     // skip for NC environments since system owner is not applicable
     if (!is_nc_environment() && (cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
-    const web_identity_info = fetch_web_identity_info(req);
-    const permission = has_assume_role_permission(assume_role_policy, method, cur_account_email, web_identity_info);
+    const permission = has_assume_role_permission(assume_role_policy, method, req.sts_sdk.requesting_account, web_identity_info);
```
sakshimunjal, line 159: any specific reason 'cur_account_email' was replaced?

src/endpoint/sts/sts_rest.js
```diff
@@ -159,22 +146,31 @@ async function authorize_request(req) {
...
+    let web_identity_info;
+    if (req.body.web_identity_token) {
+        web_identity_info = jwt.decode(req.body.web_identity_token, { json: true });
+    }
+    return web_identity_info || {};
 }
 
 async function authorize_request_policy(req) {
+
+    const web_identity_info = fetch_web_identity_info(req);
```
sakshimunjal: can we move this after checking if op name is either of 'post_assume_role' or 'post_assume_role_with_web_identity'?

src/endpoint/sts/sts_rest.js
```diff
@@ -159,22 +146,31 @@ async function authorize_request(req) {
...
     if (req.op_name !== 'post_assume_role' && req.op_name !== 'post_assume_role_with_web_identity') return;
-    const account_info = await req.sts_sdk.get_assumed_role(req);
-    const assume_role_policy = account_info.role_config && account_info.role_config.assume_role_policy;
+    // GAP: If assume_role_with_web_identity api call with LDAP token skip the policy check,
+    // account.role_config schema missing Principal.Federated. For LDAP authorize_request_policy was skipping before
+    if (!req.sts_sdk.requesting_account && web_identity_info.password) {
```
sakshimunjal: why are we skipping policy check for LDAP? has_assume_role_permission needs to be called for LDAP as well in order to check principal, effect, conditions etc

src/endpoint/sts/sts_rest.js
naveenpaul1: @sakshimunjal you are saying LDAP JWT token have `type: ldap` in it? If yes I will use it identify the LDAP and KeyCloak request

src/endpoint/sts/sts_rest.js
naveenpaul1: I didnt get you, dose `req.sts_sdk.identity_info` have any value? JWT token we pass for `assume-role-with-web-identity` is `--web-identity-token`, that in code is `req.body.web_identity_token`

src/endpoint/sts/sts_rest.js
naveenpaul1: yes, we need to support ARN also in principle validation, we check only email and `*`, for creating ARN for validation we need account id and name, that account have

src/endpoint/sts/sts_rest.js
sakshimunjal: yes, at line 125, the response of authenticate_web_identity(req) is stored in req.sts_sdk.identity_info ie the user details from ldap

src/endpoint/sts/sts_rest.js
naveenpaul1: like I mentioned in comment, account.role_config schema missing Principal.Federated, without it principle validation will faill all the time

src/endpoint/sts/ops/sts_post_assume_role_with_web_identity.js
```diff
@@ -15,6 +15,7 @@ async function assume_role_with_web_identity(req) {
     dbg.log0('sts_post_assume_role_with_web_identity body: ', _.omit(req.body, 'web_identity_token'));
     const duration_ms = sts_utils.parse_sts_duration(req.body.duration_seconds);
     const duration_sec = Math.ceil(duration_ms / 1000);
+    // TODO: default values is max_session_duration in role, if duration_seconds is set pick it only if it is less than max_session_duration
```
shirady: Is this TODO relevant?
I think it is covered in the function `parse_sts_duration`.

src/endpoint/sts/sts_rest.js
```diff
@@ -133,7 +118,9 @@ async function handle_request(req, res) {
 async function authenticate_request(req) {
     try {
         signature_utils.authenticate_request_by_service(req, req.sts_sdk);
-        if (req.op_name === 'post_assume_role_with_web_identity') {
+        const web_identity_info = fetch_web_identity_info(req);
```
shirady: Why is it done before the condition of the method ('req.op_name === 'post_assume_role_with_web_identity'')?
`authenticate_request` is general for STS service, and I'm not sure that the use of `web_identity_info` is relevant for all the methods in the service.

src/endpoint/sts/sts_rest.js
```diff
@@ -133,7 +118,9 @@ async function handle_request(req, res) {
...
+        const is_keycloack_request = web_identity_info.iss;
```
shirady: I'm not sure if it is correct to have a value if the issuer doesn't mean that it is a `keycloak_request`. Either something is missing in this statement, rename the variable, or add a comment.

src/endpoint/sts/sts_rest.js
```diff
@@ -159,22 +146,31 @@ async function authorize_request(req) {
 }
 
 function fetch_web_identity_info(req) {
```
shirady: I think this function should bot be in the sts_rest, it is very specific to one method and not to all the methods in the service.
Moving it would also avoid the additional import that were added (for example `const jwt = require('jsonwebtoken');`).

src/endpoint/sts/sts_rest.js
shirady: I agree with this comment.

src/endpoint/sts/sts_rest.js
```diff
@@ -159,22 +146,31 @@ async function authorize_request(req) {
...
+    return web_identity_info || {};
 }
 
 async function authorize_request_policy(req) {
+
+    const web_identity_info = fetch_web_identity_info(req);
     if (req.op_name !== 'post_assume_role' && req.op_name !== 'post_assume_role_with_web_identity') return;
-    const account_info = await req.sts_sdk.get_assumed_role(req);
-    const assume_role_policy = account_info.role_config && account_info.role_config.assume_role_policy;
+    // GAP: If assume_role_with_web_identity api call with LDAP token skip the policy check,
```
shirady: Will this GAP be handled as part of the KeyCloak or LDAP epics?

src/endpoint/sts/sts_rest.js
```diff
@@ -159,22 +146,31 @@ async function authorize_request(req) {
...
+    // account.role_config schema missing Principal.Federated. For LDAP authorize_request_policy was skipping before
+    if (!req.sts_sdk.requesting_account && web_identity_info.password) {
+        return;
+    }
+
+    const assume_role_policy = await get_assume_role_policy(req, web_identity_info);
```
shirady: What are you trying to verify here?
That the `requesting account` (account/user) has the IAM permission to call the op `AssumeRolePolicy`?
It should be in a separate function - you can see in `s3_rest`, and later it was added in `vector_rest` so it can be shared between services.

src/endpoint/sts/sts_rest.js
```diff
@@ -233,84 +229,153 @@ function handle_error(req, res, err) {
...
+    const statement_principal = statement.Principal || statement.NotPrincipal;
+
+    let principal_fit = false;
+    for (const principal of _.flatten([statement_principal])) {
+        if (principal.Federated) {
+                dbg.log0('assume_role_policy: principal federated fit?', principal.Federated, web_identity_info.iss);
+                if (principal.Federated.split("oidc-provider/")[1] === web_identity_info.iss?.split('//')[1]) {
+                    principal_fit = true;
+                }
+        } else {
```
shirady: Better add a comment here why it is needed (if I understand correctly, it is part of the GAP).

src/endpoint/sts/sts_rest.js
```diff
@@ -233,84 +229,153 @@ function handle_error(req, res, err) {
...
+            const cur_account_email = account?.email.unwrap();
+            // Added ARN to principle validation
+            // TODO - Add arraay of ARN to support multiple ARN in assume role policy
+            const account_identifier_arn = account ? access_policy_utils.get_bucket_policy_principal_arn(account) : "";
+            dbg.log0('assume_role_policy: principal fit?', principal.unwrap(), cur_account_email);
+            if ((principal.unwrap() === cur_account_email) || (principal.unwrap() === '*') || (principal.unwrap() === account_identifier_arn)) {
```
shirady: Something is missing in the check - the first condition is `principal.Federated`, what are we expecting the `principal` to be? Can it be `principal.AWS`? if so, then you can't call `unwarp()` on the `principal`.

src/endpoint/sts/sts_rest.js
```diff
@@ -233,84 +229,153 @@ function handle_error(req, res, err) {
...
-        const predicate = predicate_map[operator];
-        if (!predicate) return false; // unsupported operator
 
-        for (const [condition_key, policy_value] of Object.entries(condition_statements)) {
-            if (condition_key.startsWith('ldap:')) { // LDAP identity condition
-                if (!_is_ldap_identity_fit(condition_key, policy_value, web_identity_info, predicate)) return false;
-            } else {
-                return false;
-            }
+function _is_wildcard_match(action, method) {
```
shirady: What should this function do?
It returns the `method.startsWith(service_prefix)` in some conditions, but I'm not sure if it is used.

src/endpoint/sts/sts_rest.js
```diff
@@ -233,84 +229,153 @@ function handle_error(req, res, err) {
...
+ * @returns {Promise<Object>} - Assume role policy document
+ */
+async function get_assume_role_policy(req, web_identity_info) {
+    const is_keycloack_req = web_identity_info.iss;
+    let assume_role_policy;
+    if (is_keycloack_req) {
+        const role_arn = req.body.role_arn;
+        const role_name = role_arn.slice(role_arn.lastIndexOf('/') + 1);
+        // TODO: Get the iam_role from cache
+        const iam_role = _.find(system_store.data.iam_roles || [], role => {
```
shirady: Having the endpoint to call the system store does not follow the current architecture if I understand correctly.

src/endpoint/sts/sts_utils.js
```diff
@@ -51,5 +51,37 @@ function _sts_duration_validation_error(duration_input, constraint, constraint_v
     };
 }
 
+/**
+ * convert_to_aws_policy_format converts a policy object with lowercase keys
+ * to the standard AWS policy format with capitalized keys (Version, Statement, etc.)
+ * @param {Object} policy - Policy object with lowercase keys
+ * @returns {Object} Policy object with AWS standard capitalized keys
+ */
+function convert_to_aws_policy_format(policy) {
```
shirady: How it work in current policies? Do we have something like that in the bucket policy?

src/sdk/sts_sdk.js
```diff
@@ -1,16 +1,18 @@
...
+const _ = require('lodash');
+
 const cloud_utils = require('../util/cloud_utils');
 const dbg = require('../util/debug_module')(__filename);
 const { RpcError } = require('../rpc');
 const signature_utils = require('../util/signature_utils');
 const { account_cache, dn_cache } = require('./object_sdk');
 const BucketSpaceNB = require('./bucketspace_nb');
 const jwt = require('jsonwebtoken');
+const system_store = require('../server/system_services/system_store').get_instance();
```
shirady, line 13: (copying the comment) - Having the endpoint to call the system store does not follow the current architecture, if I understand correctly.

src/sdk/sts_sdk.js
```diff
@@ -173,37 +194,32 @@ class StsSDK {
             if (!keycloak_instance.initialized) {
                 await keycloak_instance.initialize();
             }
-            // Also verify the JWT signature for additional security
-            const verified_token = await keycloak_instance.verify_token(
-                req.body.web_identity_token
-            );
+            // Decode the JWT token for and verify token issuer in provider list
```
shirady: I think the comment is partial:

> Decode the JWT token for and...

(for what?)

src/sdk/sts_sdk.js
```diff
@@ -173,37 +194,32 @@ class StsSDK {
...
+                role_config,
                 sub: introspection_resp.sub,
                 aud: introspection_resp.client_id || introspection_resp.aud,
                 iss: introspection_resp.iss,
-                session_tags,
                 // Store additional claims for audit
-                email: introspection_resp.email || verified_token.email,
-                name: introspection_resp.name || verified_token.name || verified_token.preferred_username,
+                email: introspection_resp.email,
+                name: introspection_resp.name,
```
shirady, line 201: Why did you change here?

src/sdk/sts_sdk.js
```diff
@@ -173,37 +194,32 @@ class StsSDK {
...
+            await keycloak_instance.verify_token(req.body.web_identity_token);
 
             // Introspect token with Keycloak using client_id, client_secret, and access_token
             // This is the key implementation for Keycloak - validates token is active and not revoked
             const introspection_resp = await keycloak_instance.introspect_token(
                 req.body.web_identity_token
             );
 
-            // Extract session tags from verified token (not from introspection)
-            const session_tags = extract_session_tags(verified_token);
```
shirady, line 188: Why did you remove this?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -998,6 +998,8 @@ mocha.describe('Assume role with web indentity tests', function() {
     });
 
     mocha.it('anonymous user a with bad jwt - should be rejected', async function() {
+        // TODO: Need to update when authorize flow check for user and password based authentication
+        // now checking issuer there
```
shirady: I'm not sure this TODO is clear

src/util/iam_role_utils.js
```diff
@@ -0,0 +1,78 @@
+/* Copyright (C) 2026 NooBaa */
+'use strict';
+
+const dbg = require('./debug_module')(__filename);
+
+const keycloak_utils = require('../util/keycloak_utils');
+
+//TODO: Move the code iam_util.js file.
```
shirady: I'm not sure it should be `iam_utils`.
I think it is related to `access_policy_utils`.

src/util/iam_role_utils.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+function _is_identity_condition_fit(is_keycloak_request, condition, web_identity_info) {
+    const predicate_map = is_keycloak_request ? keycloak_utils.keycloak_predicate_map : ldap_predicate_map;
+    const evaluation_context = {
+        // TODO: TOKEN_CLAIMS_NAME missing error response in console
+        tags_claim: web_identity_info ? web_identity_info[TOKEN_CLAIMS_NAME]?.principal_tags : {},
+        token_claims: web_identity_info || {},
+    };
+
+    for (const [operator, condition_statements] of Object.entries(condition || {})) {
```
shirady: I think it would be more clear `condition_key, value` instead of `operator, condition_statements` like it is in:
https://github.com/noobaa/noobaa-core/blob/79225a8c90b4d6e042ba14f401f98cff03a5f5a3/src/util/access_policy_utils.js#L286
Best would be the reuse of existing functions.

src/util/iam_role_utils.js
shirady, line 1: It is a new file, try to add JSDoc to the new functions for easier use in the future.

src/util/keycloak_client.js
```diff
@@ -74,9 +74,8 @@ class KeyCloakClientManager {
 
     /**
      * Verify token and return provider + verified token
-     * This method uses JWT signature verification first, then optionally introspects
+     * This method decode the token, introspects will do actual verification
```
shirady, line 77: Where the "introspects will do actual verification"?

src/util/keycloak_utils.js
```diff
@@ -157,19 +107,157 @@ class KeyCloakProvider {
...
+ * Example:
+ * Condition: { "StringEquals": { "aws:RequestTag/Department": "Engineering" } }
+ * tags_claim: { "Department": "Engineering" }
+ * Result: true
+ * 
+ * @param {Object} condition_values - Condition key-value pairs
+ * @param {Object} evaluation_context - Tags and token claims from JWT token
+ * @returns {boolean}
+ */
+function validate_string_equals(condition_values, evaluation_context) {
```
shirady: Seems like part of the functions are not specific to Keycloak, and can be written in more relevant files.
I think @nadavMiz wrote parts of the conditions that are related to S3, for example - 
https://github.com/noobaa/noobaa-core/blob/79225a8c90b4d6e042ba14f401f98cff03a5f5a3/src/util/access_policy_utils.js#L115-L119

Locating the functionality in a more general place can help us when we try to reuse.

src/util/keycloak_utils.js
```diff
@@ -157,19 +107,157 @@ class KeyCloakProvider {
...
+
+/**
+ * Get the actual value from evaluation context based on condition key type
+ * Determines whether to retrieve from tags_claim or token_claims based on the condition key format
+ *
+ * Tag conditions (e.g., "aws:RequestTag/Department") retrieve from evaluation_context.tags_claim
+ * Token claim conditions (e.g., "keycloak.example.com:aud") retrieve from evaluation_context.token_claims
+ *
+ * @param {string} condition_key - The condition key from the policy (e.g., "aws:RequestTag/Team" or "keycloak.example.com:aud")
+ * @param {Object} evaluation_context - Context containing token claims and tags
```
shirady: Can you give an example of what is `evaluation_context`?

src/util/keycloak_utils.js
```diff
@@ -157,19 +107,157 @@ class KeyCloakProvider {
...
+ */
+function compare_string_equals(actual, expected) {
+    if (actual === undefined || actual === null) {
+        return false;
+    }
+    const actual_array = Array.isArray(actual) ? actual : [actual];
+    const expected_array = Array.isArray(expected) ? expected : [expected];
+
+    // For StringEquals, we need exact match
+    // If expected is a single value, actual must contain that exact value
```
shirady: I'm not sure this comment is relevant, as you converted all the values to an array.

src/test/unit_tests/util_functions_tests/test_assume_iam_role_trust_policy.test.js
```diff
@@ -0,0 +1,657 @@
...
+                },
+            };
+            const result = iam_role_utils._is_identity_condition_fit(true, conditions, web_identity_info);
+            expect(result).toBe(true);
+        });
+    });
+
+    describe('Edge Cases', () => {
+
+        it('should return true when no conditions provided', () => {
```
shirady, line 324: I don't think this is an edge case

src/test/unit_tests/util_functions_tests/test_assume_iam_role_trust_policy.test.js
```diff
@@ -0,0 +1,657 @@
...
+            const conditions = {
+                StringEquals: {
+                    'aws:RequestTag/Department': 'Engineering'
+                }
+            };
+            const result = iam_role_utils._is_identity_condition_fit(true, conditions, web_identity_info);
+            expect(result).toBe(false);
+        });
+
+        it('should handle custom condition key formats', () => {
```
shirady, line 374: What's the case here? lowercase value in `aws:request_tag`?
Is it supported in AWS?

src/test/unit_tests/util_functions_tests/test_assume_iam_role_trust_policy.test.js
shirady, line 1: I'm not sure about the name of the file, "iam_policy", it seems more like "trust policy".

src/test/unit_tests/util_functions_tests/test_iam_policy_utils.test.js
```diff
@@ -0,0 +1,657 @@
...
+                        'aws:RequestTag/Project': ['NooBaa', 'S3Compatible']
+                    }
+                }
+            };
+
+            const result = iam_role_utils._is_identity_condition_fit(true, trust_policy_statement.Condition, web_identity_info);
+            expect(result).toBe(true);
+        });
+
+        it('should validate production environment access control', () => {
```
shirady: I'm not sure about this title

src/endpoint/sts/sts_rest.js
naveenpaul1: @shirady We need `signature_utils.authenticate_request_by_service(req, req.sts_sdk);` for KeyCloak, `await req.sts_sdk.authenticate_web_identity(req)` is pecific check for the LDAP where its gets the condition tages using username and password.

src/endpoint/sts/sts_rest.js
naveenpaul1: I dont think this method can be share services, but still we can move this to access_policy_utils

src/endpoint/sts/sts_rest.js
naveenpaul1: Like I mentioned in GAP, We will replce it with cache

src/sdk/sts_sdk.js
naveenpaul1: @shirady Need to find a option to get account by account_id, May be will create RPC to get account by ID, WDYT?

src/sdk/sts_sdk.js
naveenpaul1: @shirady I realized that the token introspection_endpoint call already verifies the code and checks the client status. Previously, we were decoding the token with a signing key obtained from Keycloak in a separate call, but since that would be a duplicate check, I have removed the signing key verification. We are now relying solely on the introspection_endpoint to validate the token.

src/sdk/sts_sdk.js
naveenpaul1: We dont need it for now

src/sdk/sts_sdk.js
shirady: I'm not aware of the options, but it sounds like you need something that we didn't have before.
Since it is something general, better create it's own PR so it would get reviewed in a wider scope.

src/util/keycloak_client.js
naveenpaul1: yes,

src/util/keycloak_utils.js
naveenpaul1: You can find the property inside the `evaluation_context` right below this line, 

```
* @param {Object} evaluation_context.tags_claim - Tag values from the JWT token
* @param {Object} evaluation_context.token_claims - Standard JWT claims (aud, sub, azp, etc.)
```

src/test/unit_tests/util_functions_tests/test_assume_iam_role_trust_policy.test.js
naveenpaul1: not realy, but we keep it(request_tag) just to be in safer side.

src/sdk/sts_sdk.js
naveenpaul1: @shirady thats the plan, create a PR specifc to this part, just like role cache changes

src/endpoint/sts/sts_rest.js
```diff
@@ -233,84 +213,123 @@ function handle_error(req, res, err) {
     res.end(reply);
 }
 
-function has_assume_role_permission(policy, method, cur_account_email, web_identity_info) {
-    const [allow_statements, deny_statements] = _.partition(policy.statement, statement => statement.effect === 'allow');
-
+function has_assume_role_permission(policy, method, account, web_identity_info) {
```
shirady, line 218: (I added the comment on this function, but it is on the flow that it creates: `has_assume_role_permission` -> `_is_statements_fit` -> `_is_action_fit` -> `_is_principal_fit` -> `_is_condition_fit`.

I think we'd better avoid this code duplication and edit the `access_policy_utils` to match what you need.
@naveenpaul1 @nadavMiz and @jackyalbo WDYT?

src/endpoint/sts/sts_utils.js
```diff
@@ -51,5 +51,6 @@ function _sts_duration_validation_error(duration_input, constraint, constraint_v
     };
 }
 
+
```
shirady: Redundant new line?

src/sdk/sts_sdk.js
```diff
@@ -70,32 +72,39 @@ class StsSDK {
     async _assume_role(role_arn) {
         const role_name_idx = role_arn.lastIndexOf('/') + 1;
         const role_name = role_arn.slice(role_name_idx);
-        const access_key = role_arn.split(':')[4];
-
-        const account = await account_cache.get_with_cache({
-            bucketspace: this._get_bucketspace(),
-            access_key,
+        const account_key_or_id = role_arn.split(':')[4];
+        // TODO: Use cache instead of system_store
```
shirady, line 76: Could you add the comment "// TODO: Use cache instead of system_store" in all places where you imported the system store in endpoint files, so it's clear this is a temporary change.

src/sdk/sts_sdk.js
```diff
@@ -70,32 +72,39 @@ class StsSDK {
...
-            'account.role_config: ', account.role_config);
-
-        return account;
+        dbg.log1('sts_sdk.get_assumed_role res', 'iam_role: ', iam_role.name);
+        return {
+            ...iam_role,
+            role_name: iam_role.name,
+            account_id: account._id.toString(),
+            access_key: account.access_keys[0].access_key.unwrap(),
+            assume_role_policy: iam_role.assume_role_policy || iam_role.assume_role_policy_document,
```
shirady: Please explain in a comment why both `assume_role_policy` `assume_role_policy_document` (it is confusing), If it is temporary, you can mention it as well.

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -20,6 +20,7 @@ const jwt_utils = require('../../../../util/jwt_utils');
 const config = require('../../../../../config');
 const ldap_client = require('../../../../util/ldap_client');
 const { S3Error } = require('../../../../endpoint/s3/s3_errors');
+const { CreateRoleCommand, DeleteRoleCommand, UpdateAssumeRolePolicyCommand } = require('@aws-sdk/client-iam');
```
shirady, line 24: I think having 2 SDK versions in 1 test file is confusing.
I guess you don't have the time to migrate it, so could you create a new file? One would be for assume role (what we already had in STS), and one would be for assume role with web identity, WDYT?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -132,28 +139,39 @@ mocha.describe('STS tests', function() {
...
         const policy = {
-            version: '2012-10-17',
-            statement: [{
-                effect: 'allow',
-                principal: is_nc_coretest ? [user_c, EMAIL] : [user_c],
-                action: ['sts:AssumeRole'],
+            Version: '2012-10-17',
+            Statement: [{
+                Effect: 'Allow',
+                Principal: { AWS: [user_c, EMAIL]},
```
shirady: Removing the condition `is_nc_coretest ?` might be harder when they want to use the test again.
@naveenpaul1  @jackyalbo @sakshimunjal WDYT?
What is the plan about this GAP in NC with the ARN VS access keys?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -837,126 +866,6 @@ mocha.describe('Session token tests', function() {
     }
 });
 
-
-mocha.describe('Assume role policy tests', function() {
```
shirady, line 841: Why was this deleted?

src/endpoint/sts/sts_rest.js
```diff
@@ -133,7 +114,9 @@ async function handle_request(req, res) {
 async function authenticate_request(req) {
     try {
         signature_utils.authenticate_request_by_service(req, req.sts_sdk);
-        if (req.op_name === 'post_assume_role_with_web_identity') {
+        const web_identity_info = access_policy_utils.fetch_web_identity_info(req);
+        const is_ldap_request = web_identity_info.username;
+        if (req.op_name === 'post_assume_role_with_web_identity' && is_ldap_request) {
```
shirady: I think `web_identity_info` should have moved to the condition, for example:

```suggestion
        if (req.op_name === 'post_assume_role_with_web_identity' ) {
                const web_identity_info = access_policy_utils.fetch_web_identity_info(req);
        const is_ldap_request = web_identity_info.username;
        if (is_ldap_request) {...
```

src/util/access_policy_utils.js
```diff
@@ -528,6 +538,255 @@ const VECTOR_OP_NAME_TO_ACTION = Object.freeze({
...
+ * @param {Object} req - Request object
+ * @returns {Object} - web_identity_info
+ */
+function fetch_web_identity_info(req) {
+    let web_identity_info;
+    if (req.body.web_identity_token) {
+        web_identity_info = jwt.decode(req.body.web_identity_token, { json: true });
+    }
+    return web_identity_info || {};
+}
```
shirady: Why did you choose to put the function `fetch_web_identity_info` in the file `src/util/access_policy_utils.js`? 
Is it related to bucket policy, user policy, role policy, etc.?

src/endpoint/sts/sts_rest.js
shirady: Why?

src/util/keycloak_client.js
shirady: Where in the flow/code?

src/sdk/sts_sdk.js
naveenpaul1: Its already added

src/test/integration_tests/api/sts/test_sts.js
naveenpaul1: This test dont not make sense with latest code changes in STS flow, where we are checking the STS request flow with trust policy validations
Creating account with `role_config` success or not, we do not support it after this release.

src/test/integration_tests/api/sts/test_sts.js
naveenpaul1: its planned for next PR

src/util/access_policy_utils.js
naveenpaul1: its better to keep it in sts_rest, will move

src/util/keycloak_client.js
naveenpaul1: its not in our code, its in keycloak code flow

src/test/integration_tests/api/sts/test_sts.js
naveenpaul1: We can not use this test as it is in NC, we need more changes when implement roles in NC,

src/util/access_policy_utils.js
```diff
@@ -528,6 +537,242 @@ const VECTOR_OP_NAME_TO_ACTION = Object.freeze({
...
+ * @returns {Promise<Object>} - Assume role policy document
+ */
+async function get_assume_role_policy(req) {
+    const role_arn = req.body.role_arn;
+    const role_name = role_arn.slice(role_arn.lastIndexOf('/') + 1);
+    // TODO: Get the iam_role from cache
+    const iam_role = _.find(system_store.data.iam_roles || [], role => {
+        if (role.deleted) return false;
+        return role.name === role_name;
+    });
```
aayushchouhan09: same role name can be assigned to different root users (aws complaint), I think we should filter here with `owner` as well.
If two accounts have same role name, wrong role policy can be evaluated, should resolve role by `ARN account identifier + owner-scoped index`.

src/endpoint/sts/sts_rest.js
```diff
@@ -158,23 +145,19 @@ async function authorize_request(req) {
...
-    const account_info = await req.sts_sdk.get_assumed_role(req);
-    const assume_role_policy = account_info.role_config && account_info.role_config.assume_role_policy;
+
+    const assume_role_policy = await access_policy_utils.get_assume_role_policy(req);
     if (!assume_role_policy) throw new StsError(StsError.AccessDeniedException);
     const method = _get_method_from_req(req);
     const cur_account_email = req.sts_sdk.requesting_account && req.sts_sdk.requesting_account.email.unwrap();
     // system owner by design can always assume role policy of any account
     // skip for NC environments since system owner is not applicable
     if (!is_nc_environment() && (cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
```
aayushchouhan09, line 157: That is too broad for IAM roles (AWS semantics), and should only be allowed for legacy `role_config` compatibility paths, otherwise system owner can assume roles it should not, causing policy bypass/privilege escalation risk

src/sdk/sts_sdk.js
```diff
@@ -70,32 +72,39 @@ class StsSDK {
...
-
-        const account = await account_cache.get_with_cache({
-            bucketspace: this._get_bucketspace(),
-            access_key,
+        const account_key_or_id = role_arn.split(':')[4];
+        // TODO: Use cache instead of system_store
+        const iam_role = _.find(system_store.data?.iam_roles || [], role => {
+            if (role.deleted) return false;
+            return role.name === role_name;
         });
```
aayushchouhan09, line 80: same here
Scanning global roles by name + and doing checks later is not good, should resolve account first from ARN, then search only that owner’s roles

src/sdk/sts_sdk.js
```diff
@@ -70,32 +72,39 @@ class StsSDK {
...
-        if (!account) {
-            throw new RpcError('NO_SUCH_ACCOUNT', 'No such account with access_key: ' + access_key);
+        if (!iam_role) {
+            throw new RpcError('NO_SUCH_ROLE', `No such Role found with name: ${role_name}`);
         }
-        if (!account.role_config || account.role_config.role_name !== role_name) {
-            throw new RpcError('NO_SUCH_ROLE', `Role not found`);
+        const account = iam_role.owner;
+        const access_key_obj = _.find(account.access_keys, access_key => access_key.access_key.unwrap() === account_key_or_id);
```
aayushchouhan09: Access key selection should avoid deactivated keys

src/util/access_policy_utils.js
```diff
@@ -528,6 +537,242 @@ const VECTOR_OP_NAME_TO_ACTION = Object.freeze({
...
+            if (is_keycloak_request) {
+                if (!predicate({ [expected_key]: expected_value }, evaluation_context)) {
+                    dbg.log0('_is_identity_condition_fit: Condition validation failed for operator:', expected_key,
+                        'condition_key:', expected_key);
+                    return false;
+                }
+            }
+            if (condition_key.startsWith('ldap:')) { // LDAP identity condition
+                if (!_is_ldap_identity_fit(condition_key, expected_value, web_identity_info, predicate)) return false;
+            }
```
aayushchouhan09, line 582:
```suggestion
            if (expected_key.startsWith('ldap:')) {
                if (!_is_ldap_identity_fit(expected_key, expected_value, web_identity_info, predicate)) return false;
            } else if (is_keycloak_request) {
                if (!predicate({ [expected_key]: expected_value }, evaluation_context)) {
                    dbg.log0('_is_identity_condition_fit: Condition validation failed for operator:', expected_key,
                        'condition_key:', expected_key);
                    return false;
                }
            }
            } else {
                return false;
            }
```

Bug: LDAP matching is checked against the operator variable (`condition_key = StringEquals`) instead of the actual claim key, so LDAP conditions can be skipped.
Fix: check `expected_key.startsWith('ldap:')` and evaluate LDAP predicate on that key.

aayushchouhan09 (commented):
I was also working on the trust policy validation, but as you did the PR for the same part (partial), I will review it and cover the remaining part later.

Some comments. Thanks!

src/sdk/sts_sdk.js
naveenpaul1: Dont think so, Here we are not checking the access key active or not, We are checking role is active or not, Deleted access key should not deactivate the role in my opinion.

src/endpoint/sts/sts_rest.js
naveenpaul1: didnt get you, if condition was there before

src/endpoint/sts/sts_rest.js
naveenpaul1: I will prefer to keep it like this for this PR, refactoring should be for another PR

src/endpoint/sts/ops/sts_post_assume_role.js
```diff
@@ -33,6 +33,7 @@ async function assume_role(req) {
         AssumeRoleResponse: {
             AssumeRoleResult: {
                 AssumedRoleUser: {
+                    // TODO: cnahge the role arn with account id
```
shirady: typo
```suggestion
                    // TODO: change the role ARN with account id
```

src/endpoint/sts/sts_rest.js
```diff
@@ -158,23 +145,19 @@ async function authorize_request(req) {
     await authorize_request_policy(req);
 }
 
-function fetch_web_identity_info(req) {
-    // add conditional logic to fetch web identity info from LDAP or Keycloak or default
-    return req.sts_sdk.identity_info;
-}
-
 async function authorize_request_policy(req) {
+
```
shirady: redundant newline?

src/util/access_policy_utils.js
```diff
@@ -2,6 +2,7 @@
 'use strict';
 
 const _ = require('lodash');
+
```
shirady: Redundant newline?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -197,18 +216,22 @@ mocha.describe('STS tests', function() {
         for (const email of accounts) {
             await rpc_client.account.delete_account({ email });
         }
+
+        try {
+            await iam_client_b.send(new DeleteRoleCommand({ RoleName: iam_role_name }));
+        } catch (_err) { /* ignore */ }
```
shirady, line 222: Do we have an error on `DeleteRoleCommand`?
I see:
`catch (_err) { /* ignore */ }`

src/test/integration_tests/api/sts/test_sts.js
shirady, line 1: If you didn't change the ARN in this PR, do you know why I see the account ID and not the access key?

src/test/integration_tests/api/sts/test_sts.js
naveenpaul1: no, its added for safty

src/test/integration_tests/api/sts/test_sts.js
naveenpaul1: `${account_info_b._id.toString()` if for ARN as input, not as out put, output ARN is `arn:aws:sts::${user_a_key}:assumed-role/${role_alice}/${params.RoleSessionName}`
this is expected

shirady (approved):
Approving not to block you

Expecting the following PRs with the open comments (access policy code reuse, test using one AWS SDK, ARN structure with account ID, avoid DB calls from the endpoint, etc.).

PR #9765, opened 2026-06-22
Add Leader Election module to noobaa core

nadavMiz:
### Describe the Problem
Today the NooBaa core pod runs as a single active instance. If it fails or is rescheduled, there is no built-in way for a standby core pod to take over cleanly. We need active-passive high availability so only one core pod serves at a time, with automatic failover when the leader is lost.

### Explain the Changes
This PR adds a new Kubernetes Lease-based leader election module to the core library.

The new module is responsible for acquiring and renewing leadership, releasing it on shutdown, and notifying callers when leadership is gained or lost. It provides a simple event-driven interface that can later be used by core_init and other HA-related flows.

This PR only adds the reusable leader election module and its unit tests.
The actual wiring into core_init is done separately in a follow-up PR.

New pieces:
1. Added a new Lease client in `core_lease_utils.js`. this client is provides functions to acquire, renew, and release the lease
2. added config variables CORE_LEASE_ACQUIRE_RETRY_MS and CORE_LEASE_RENEW_DEADLINE_MS to configure lease timings
3. Unit tests for the lease client 

### Issues:
1. part of https://redhat.atlassian.net/browse/RHSTOR-7491

operator side change: https://github.com/noobaa/noobaa-operator/pull/1964

### Testing Instructions:
1. `node --allow-natives-syntax ./node_modules/.bin/_mocha \
  src/test/unit_tests/util_functions_tests/test_core_lease.js`

- [ ] Doc added/updated
- [x] Tests added

## Summary

- **New Features**
  - Added Kubernetes Lease–based active/passive leader election with `leadership_acquired` and `leadership_lost` events.
  - Introduced core lease timing and Kubernetes API request timeout configuration for the leader-election behavior.
  - Expanded HTTPS trust roots by optionally loading the Kubernetes CA certificate via a new `KUBE_CA_CERTS` setting.

- **Tests**
  - Added a comprehensive leader-election unit test suite covering lease parsing/formatting, expiry/takeability logic, and acquire/renew/release edge cases.
  - Wired the new leader-election tests into the existing unit test runner.

### DIFF
--- a/config.js
+++ b/config.js
@@ -672,6 +672,22 @@ config.KUBE_API_ENDPOINTS = {
     TOKEN_REVIEW: '/apis/authentication.k8s.io/v1/tokenreviews'
 };
 
+// Core active-passive HA (Kubernetes Lease). Disabled when NOOBAA_CORE_LEASE_NAME is unset.
+// Time to wait between lease acquisition retries and between successful renew iterations.
+config.CORE_LEASE_ACQUIRE_RETRY_MS = 3000;
+// Short sleep between renew retries after a transient error (non-200, non-409 response).
+// Keeping this short (1s) lets us retry many times within the renew deadline.
+config.CORE_LEASE_RENEW_ERROR_SLEEP_MS = 1000;
+// Safety margin (ms) subtracted from the live leaseDurationSeconds when computing
+// how long the renew loop may fail before the leader steps down.
+// deadline = leaseDuration - CORE_LEASE_RENEW_ERROR_SLEEP_MS - CORE_LEASE_REQUEST_TIMEOUT_MS - LEASE_RENEW_DEADLINE_MARGIN_MS
+// e.g. 20000 - 1000 - 5000 - 4000 = 10000ms with a 20s lease.
+config.LEASE_RENEW_DEADLINE_MARGIN_MS = 4000;
+// Timeout for each individual Kubernetes API request in the lease client.
+// Keeps a hung API server from blocking the renew loop past the deadline.
+// note on go leader election module it is calculateed as max(renewDeadline / 2, time.Second). use a slightly shorter timeout to account for the extra delay.
+config.CORE_LEASE_REQUEST_TIMEOUT_MS = 5000;
+
 //////////////////////////////
 // ACCOUNT PREFERENCES      //
 //////////////////////////////
--- a/config.js
+++ b/config.js
@@ -673,19 +673,16 @@ config.KUBE_API_ENDPOINTS = {
 };
 
 // Core active-passive HA (Kubernetes Lease). Disabled when NOOBAA_CORE_LEASE_NAME is unset.
+// Desired lease duration..
+config.CORE_LEASE_DURATION_MS = 20000;
 // Time to wait between lease acquisition retries and between successful renew iterations.
 config.CORE_LEASE_ACQUIRE_RETRY_MS = 3000;
-// Short sleep between renew retries after a transient error (non-200, non-409 response).
-// Keeping this short (1s) lets us retry many times within the renew deadline.
+// Short sleep between renew retries after an error
 config.CORE_LEASE_RENEW_ERROR_SLEEP_MS = 1000;
-// Safety margin (ms) subtracted from the live leaseDurationSeconds when computing
-// how long the renew loop may fail before the leader steps down.
-// deadline = leaseDuration - CORE_LEASE_RENEW_ERROR_SLEEP_MS - CORE_LEASE_REQUEST_TIMEOUT_MS - LEASE_RENEW_DEADLINE_MARGIN_MS
-// e.g. 20000 - 1000 - 5000 - 4000 = 10000ms with a 20s lease.
-config.LEASE_RENEW_DEADLINE_MARGIN_MS = 4000;
+// The deadline for the lease renewal.
+config.LEASE_RENEW_DEADLINE_MS = 13000;
 // Timeout for each individual Kubernetes API request in the lease client.
-// Keeps a hung API server from blocking the renew loop past the deadline.
-// note on go leader election module it is calculateed as max(renewDeadline / 2, time.Second). use a slightly shorter timeout to account for the extra delay.
+// note on go leader election module it is calculated as max(renewDeadline / 2, time.Second). use a slightly shorter timeout to account for the extra delay.
 config.CORE_LEASE_REQUEST_TIMEOUT_MS = 5000;
 
 //////////////////////////////
--- a/src/util/leader_election.js
+++ b/src/util/leader_election.js
@@ -1,0 +1,495 @@
+/* Copyright (C) 2016 NooBaa */
+'use strict';
+
+const fs = require('fs');
+const { EventEmitter } = require('events');
+const dbg = require('./debug_module')(__filename);
+const config = require('../../config');
+const { make_https_request } = require('./http_utils');
+const { read_stream_join } = require('./buffer_utils');
+const { randomUUID } = require('crypto');
+
+const LEASE_API = 'coordination.k8s.io/v1';
+
+const {
+    KUBERNETES_SERVICE_HOST,
+    KUBERNETES_SERVICE_PORT,
+} = process.env;
+
+/**
+ * Named events emitted by LeaderElector.
+ */
+const EVENTS = {
+    LEADERSHIP_ACQUIRED: 'leadership_acquired',
+    LEADERSHIP_LOST: 'leadership_lost',
+};
+
+/**
+ * @param {string|undefined} iso_string
+ * @returns {number}
+ */
+function parse_lease_time(iso_string) {
+    if (!iso_string) return 0;
+    const ms = Date.parse(iso_string);
+    return Number.isFinite(ms) ? ms : 0;
+}
+
+/**
+ * Formats a Date for Kubernetes Lease spec acquireTime/renewTime (metav1.MicroTime).
+ * JavaScript toISOString() uses millisecond precision, the API expects microseconds.
+ * @param {Date} [date]
+ * @returns {string}
+ */
+function format_lease_time(date = new Date()) {
+    return date.toISOString().replace(/\.(\d{3})Z$/, (_, ms) => `.${ms}000Z`);
+}
+
+/**
+ * Returns true if the lease record has no valid holder (expired or empty renew time).
+ * @param {{ spec?: { holderIdentity?: string, renewTime?: string, leaseDurationSeconds?: number } }} lease
+ * @param {number} [observed_change_ms] local clock time when we last saw the lease record change.
+ *   Falls back to spec.renewTime when not provided.
+ * @returns {boolean}
+ */
+function is_lease_expired(lease, observed_change_ms = 0) {
+    const spec = lease?.spec || {};
+    if (!spec.leaseDurationSeconds) return true;
+    const reference_ms = observed_change_ms || parse_lease_time(spec.renewTime);
+    if (!reference_ms) return true;
+    return Date.now() >= reference_ms + (spec.leaseDurationSeconds * 1000);
+}
+
+/**
+ * Returns true if the given holder may acquire or renew this lease according to spec.
+ * @param {{ spec?: { holderIdentity?: string, renewTime?: string, leaseDurationSeconds?: number } }} lease
+ * @param {string} holder
+ * @param {number} [observed_change_ms] local clock time when we last saw the lease record change (clock-skew-safe)
+ * @returns {boolean}
+ */
+function is_lease_takeable(lease, holder, observed_change_ms = 0) {
+    const spec = lease?.spec || {};
+    const current_holder = spec.holderIdentity;
+    if (!current_holder) return true;
+    if (current_holder === holder) return true;
+    return is_lease_expired(lease, observed_change_ms);
+}
+
+/**
+ * Returns true if the lease is currently held by the given holder identity.
+ * @param {{ spec?: { holderIdentity?: string } }} lease
+ * @param {string} holder
+ * @returns {boolean}
+ */
+function is_lease_held_by(lease, holder) {
+    return lease?.spec?.holderIdentity === holder;
+}
+
+/**
+ * Leader elector backed by a Kubernetes coordination.k8s.io/v1 Lease.
+ * Emits EVENTS.LEADERSHIP_ACQUIRED when the lease is acquired and
+ * EVENTS.LEADERSHIP_LOST when the lease is lost or the renew deadline is exceeded.
+ * LeaderElector instances configured with the same lease name form a coordination group
+ * in which one, and only one, instance can be considered the leader at a time.
+ */
+class LeaderElector extends EventEmitter {
+    /**
+     * @param {{ lease_name: string, holder: string }} options
+     */
+    constructor({ lease_name, holder }) {
+        super();
+        if (!KUBERNETES_SERVICE_HOST || !KUBERNETES_SERVICE_PORT) {
+            throw new Error('LeaderElector requires in-cluster Kubernetes environment variables');
+        }
+        this._lease_name = lease_name;
+        this._holder = holder;
+        this._service_host = KUBERNETES_SERVICE_HOST;
+        this._service_port = KUBERNETES_SERVICE_PORT;
+        this._initialized = false;
+        this._stop_requested = false;
+        /** @type {number} lease duration in ms, set from the live lease spec at acquire time */
+        this._lease_duration_ms = 0;
+        /** @type {string|null} last resourceVersion we observed from the API */
+        this._observed_resource_version = null;
+        /** @type {number} local clock ms when we last saw the lease record change (clock-skew protection) */
+        this._observed_change_ms = 0;
+        /** @type {Promise<void>|null} */
+        this._running_promise = null;
+        /** @type {NodeJS.Timeout|null} */
+        this._sleep_timer = null;
+        /** @type {(() => void)|null} */
+        this._wake_sleep = null;
+    }
+
+    async _init() {
+        if (this._initialized) return;
+        const token_buf = await fs.promises.readFile(config.KUBE_SA_TOKEN_FILE);
+        this._sa_token = token_buf.toString('utf8').trim();
+        const ns_buf = await fs.promises.readFile(config.KUBE_NAMESPACE_FILE);
+        this._namespace = ns_buf.toString('utf8').trim();
+        this._initialized = true;
+    }
+
+    _lease_path() {
+        return `/apis/${LEASE_API}/namespaces/${this._namespace}/leases/${this._lease_name}`;
+    }
+
+    /**
+     * @param {string} method
+     * @param {object} [body]
+     * @returns {Promise<{ status_code: number, body: object }>}
+     */
+    async _request(method, body) {
+        await this._init();
+        const response = await make_https_request({
+            method,
+            hostname: this._service_host,
+            port: this._service_port,
+            path: this._lease_path(),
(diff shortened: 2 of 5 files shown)


src/util/core_lease_utils.js
```diff
@@ -0,0 +1,410 @@
...
+                        return;
+                    }
+                    continue;
+                } else {
+                    dbg.warn('lease PUT failed, retrying', this._lease_name, status_code);
+                }
+                if (this._stop_renew) {
+                    break;
+                }
+                await this._interruptible_sleep(config.CORE_LEASE_ACQUIRE_RETRY_MS);
```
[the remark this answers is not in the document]
nadavMiz: this is intentional. this is a rare thing that should not even happen. in fact sleep will only increase the chance of another collision. should retry immediately to reduce release time

src/util/leader_election.js
```diff
@@ -0,0 +1,410 @@
...
+            headers: {
+                'Content-Type': 'application/json',
+                Accept: 'application/json',
+                Authorization: `Bearer ${this._sa_token}`,
+            },
+        }, body && JSON.stringify(body), 'utf8');
+        const buffer = await read_stream_join(response);
+        const res_body = buffer.length ? JSON.parse(buffer.toString('utf8')) : {};
+        return { status_code: response.statusCode, body: res_body };
+    }
```
nadavMiz: this is an existing issue. exists in kubestore as well. there shouldn't be much risk as communication is internal only. out of scope for now. should be addressed in a seperate PR

src/util/leader_election.js
```diff
@@ -0,0 +1,409 @@
...
+            hostname: this._service_host,
+            port: this._service_port,
+            path: this._lease_path(),
+            rejectUnauthorized: false,
+            headers: {
+                'Content-Type': 'application/json',
+                Accept: 'application/json',
+                Authorization: `Bearer ${this._sa_token}`,
+            },
+        }, body && JSON.stringify(body), 'utf8');
```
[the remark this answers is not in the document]
nadavMiz: currently make_https_request does not support timeout and no timeout is used throughout the code.  I agree this might be a problem but its out of scope for this PR. no new issue was introduced

src/util/core_lease_utils.js
```diff
@@ -0,0 +1,409 @@
...
+     */
+    async acquire_lease() {
+        dbg.log0('acquiring core lease', this._lease_name, 'holder', this._holder);
+        for (;;) {
+            const { lease, can_take } = await this.read_lease_state();
+            if (!lease) {
+                throw new Error(`lease ${this._lease_name} not found in namespace ${this._namespace}`);
+            }
+            if (can_take) {
+                const status_code = await this.update_lease(lease, true);
```
dannyzaken: handle errors. e.g.: timeout, connection refused, etc. 
right now it will just throw and eventually exit the container

src/util/leader_election.js
```diff
@@ -0,0 +1,409 @@
...
+    }
+
+    /**
+     * @param {string} method
+     * @param {object} [body]
+     * @returns {Promise<{ status_code: number, body: object }>}
+     */
+    async _request(method, body) {
+        await this._init();
+        const response = await make_https_request({
```
dannyzaken: we should have a reasonable timeout on the request.

src/cmd/core_init.js
```diff
@@ -44,29 +59,174 @@ async function run_server_upgrade() {
...
+
+async function start() {
+    const version = read_package_version();
+    console.log(`Version is: ${version}`);
+    console.log('running core server init');
+
+    if (is_core_lease_enabled()) {
+        _lease_client = create_client_from_env();
+        await _lease_client.acquire_lease();
+        start_core_lease_renewal(_lease_client);
```
dannyzaken: as we discusses, maybe the flow will be clearer if you make the lease_client an event emitter. after acquire_lease add a listener for `lost_leadership` event. once got the event go into TERM sequence.

src/util/leader_election.js
```diff
@@ -0,0 +1,409 @@
...
+                    }
+                } else {
+                    dbg.warn('lease PUT failed, retrying', this._lease_name, status_code);
+                }
+                if (this._stop_renew) {
+                    break;
+                }
+                await this._interruptible_sleep(config.CORE_LEASE_ACQUIRE_RETRY_MS);
+            }
+        } finally {
```
dannyzaken, line 522: add catch block for better error handling. Also retry with a shorter sleep if failed to access the lease.

config.js
```diff
@@ -671,6 +671,14 @@ config.KUBE_API_ENDPOINTS = {
     TOKEN_REVIEW: '/apis/authentication.k8s.io/v1/tokenreviews'
 };
 
+// Core active-passive HA (Kubernetes Lease). Disabled when NOOBAA_CORE_LEASE_NAME is unset.
+// Time to wait between lease acquisition retries.
+config.CORE_LEASE_ACQUIRE_RETRY_MS = 3000;
+// Max time without a successful renew PUT before the leader steps down. 
+// should be less than lease duration to avoid split brain (equivalent to client-go RenewDeadline).
+//lease duration defined in the Lease yaml on the noobaa operator.
+config.CORE_LEASE_RENEW_DEADLINE_MS = 18000;
```
dannyzaken: deadline should be shorter. 18 seconds + 3 seconds for CORE_LEASE_ACQUIRE_RETRY_MS can bring us over the 20 seconds lease timeout. 
if renew errors are retried with a shorter delay (1 sec maybe), we can make the deadline much sorter.

src/util/leader_election.js
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    async _init() {
+        if (this._initialized) return;
+        const token_buf = await fs.promises.readFile(config.KUBE_SA_TOKEN_FILE);
+        this._sa_token = token_buf.toString('utf8').trim();
+        const ns_buf = await fs.promises.readFile(config.KUBE_NAMESPACE_FILE);
+        this._namespace = ns_buf.toString('utf8').trim();
+        this._initialized = true;
+    }
```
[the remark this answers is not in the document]
nadavMiz: the should not be more then one leader_election instance per pod. also it will run once sequentialy from core_init

src/util/leader_election.js
```diff
@@ -0,0 +1,464 @@
...
+
+    /**
+     * Returns true when too long has passed since the last successful renew.
+     * Step down in this case to avoid split brain if we can no longer refresh the lease.
+     * @param {number} last_successful_renew_ms
+     * @returns {boolean}
+     */
+    _is_renew_deadline_exceeded(last_successful_renew_ms) {
+        return Date.now() - last_successful_renew_ms > config.CORE_LEASE_RENEW_DEADLINE_MS;
+    }
```
dannyzaken: there is no margin for losing leadership - `CORE_LEASE_RENEW_DEADLINE_MS` (14) + `CORE_LEASE_REQUEST_TIMEOUT_MS` (5) + 'CORE_LEASE_RENEW_ERROR_SLEEP_MS' (1) = 20 seconds. We should either reduce `CORE_LEASE_RENEW_DEADLINE_MS` (maybe to 10), or account for the possible `CORE_LEASE_REQUEST_TIMEOUT_MS` when checking the deadline

src/util/leader_election.js
```diff
@@ -0,0 +1,464 @@
...
+    /**
+     * GET the lease object from the API.
+     * @returns {Promise<object|null>}
+     */
+    async read_lease() {
+        const { status_code, body } = await this._request('GET');
+        if (status_code === 200) return body;
+        if (status_code === 404) return null;
+        throw new Error(`read lease failed status=${status_code} body=${JSON.stringify(body)}`);
+    }
```
dannyzaken: All our timing constants assume that `leaseDurationSeconds` is 20 seconds. I think we should verify this  and either log or throw an error. 
Alternatively, we can change the deadline based on the lease duration, so instead of having `CORE_LEASE_RENEW_DEADLINE_MS`, we can define `LEASE_RENEW_DEADLINE_MARGIN` to be 4-5 seconds, and then `deadline = leaseDurationSeconds - CORE_LEASE_RENEW_ERROR_SLEEP_MS - CORE_LEASE_REQUEST_TIMEOUT_MS - LEASE_RENEW_DEADLINE_MARGIN`.

src/util/leader_election.js
```diff
@@ -0,0 +1,464 @@
...
+}
+
+/**
+ * Returns true if the lease record has no valid holder (expired or empty renew time).
+ * @param {{ spec?: { holderIdentity?: string, renewTime?: string, leaseDurationSeconds?: number } }} lease
+ * @returns {boolean}
+ */
+function is_lease_expired(lease) {
+    const spec = lease?.spec || {};
+    const renew_ms = parse_lease_time(spec.renewTime);
```
dannyzaken: We should consider the possibility of clock skew between the pods. It should be rare in production clusters with NTP, but in some environments it is possible. We can deal with it by not checking the expiration based on `spec.renewTime` (which is set by the other pod), but instead track the time since the lease changed its state (e.g., renewed by the holder).

src/util/leader_election.js
```diff
@@ -0,0 +1,464 @@
...
+
+/**
+ * @returns {LeaderElector}
+ */
+function create_elector_from_env() {
+    const lease_name = process.env.NOOBAA_CORE_LEASE_NAME;
+    if (!lease_name) {
+        throw new Error('NOOBAA_CORE_LEASE_NAME is not set');
+    }
+    const holder = process.env.HOSTNAME;
```
dannyzaken: We should have a unique holder identity for each instance to avoid split-brain when we have an orphaned container. We have had similar cases with the old DB statefulset. For example, when a pod is force-deleted, its container is detached from the pod and may continue running in the background for some time before being killed.

src/util/leader_election.js
```diff
@@ -0,0 +1,464 @@
...
+     * @returns {Promise<{ status_code: number, body: object }>}
+     */
+    async _request(method, body) {
+        await this._init();
+        const response = await make_https_request({
+            method,
+            hostname: this._service_host,
+            port: this._service_port,
+            path: this._lease_path(),
+            rejectUnauthorized: false,
```
dannyzaken: We should not introduce new code with `rejectUnauthorized: false`.
you can remove it and add `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` to the `https_agent` in http_utils

src/util/leader_election.js
```diff
@@ -0,0 +1,464 @@
...
+            const lease = await this.read_lease();
+            if (!lease) {
+                throw new Error(`lease ${this._lease_name} not found in namespace ${this._namespace}`);
+            }
+            if (!is_lease_held_by(lease, this._holder)) {
+                dbg.log0('core lease not held by us, skipping release',
+                    'holder', lease.spec?.holderIdentity, 'we are', this._holder);
+                return;
+            }
+            const status_code = await this._put_release_lease(lease);
```
dannyzaken, line 372: add error handling for the release flow

src/util/leader_election.js
```diff
@@ -0,0 +1,464 @@
...
+            loop_done = resolve;
+        });
+
+        try {
+            dbg.log0('starting core lease renew loop', this._lease_name);
+            let last_successful_renew_ms = Date.now();
+            while (!this._stop_renew) {
+                // lease is about to expire, step down
+                if (this._is_renew_deadline_exceeded(last_successful_renew_ms)) {
+                    dbg.error('renew deadline exceeded, shutting down pod', this._lease_name);
```
dannyzaken: message should be `leadership lost` instead of `shutting down pod` which is the decision of the calling code

src/util/leader_election.js
nadavMiz: chose the second method as it seemed clearer and safer in case one of the times changes

src/util/leader_election.js
nadavMiz: right, changed it to use the observed time instead. this is also aligned with the go module. we can have a delay of a few seconds between observing the change and when it actually happened but its fine since its only used for acquire_lease, or if someone takes our lease. so we are more conservative in this way. and for our own time we still use last_successful_renew_ms, which is set immediately after we renew the lease, so we shouldn't need to change _get_renew_deadline_ms

src/util/leader_election.js
nadavMiz: right, this was intentional. the idea was that if the pod crash, it would not need to wait for the lease to expire before taking it. but I didn't think about the orphaned container case. changed it to have a uuid in the name

src/util/leader_election.js
nadavMiz: done, also removed the 409 handler as its not possible in our case because renew loop has already stopped at this point and it just complicates the function. worst case we will wait for the lease to expire

src/util/leader_election.js
```diff
@@ -0,0 +1,464 @@
...
+            }
+            // Renew loop or another pod may have updated the lease between read and PUT,
+            // re-read and retry.
+            if (status_code === 409) {
+                dbg.log1('release lease 409 conflict, retrying', this._lease_name);
+                continue;
+            }
+            throw new Error(`release lease failed status=${status_code}`);
+        }
+    }
```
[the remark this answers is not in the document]
nadavMiz: added try catch. is this resolved to your satisfaction?

src/util/leader_election.js
```diff
@@ -0,0 +1,495 @@
...
+            headers: {
+                'Content-Type': 'application/json',
+                Accept: 'application/json',
+                Authorization: `Bearer ${this._sa_token}`,
+            },
+        }, body && JSON.stringify(body), 'utf8');
+        const buffer = await read_stream_join(response);
+        const res_body = buffer.length ? JSON.parse(buffer.toString('utf8')) : {};
+        return { status_code: response.statusCode, body: res_body };
+    }
```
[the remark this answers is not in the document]
nadavMiz: should be prevented by new abort controller. is this acceptable?

src/util/leader_election.js
```diff
@@ -0,0 +1,535 @@
...
+                        // another pod acquired the lease or lease was deleted, step down
+                        dbg.error('lost core lease to another holder, leadership lost', this._lease_name, 'holder', lease?.spec?.holderIdentity, 'we are', this._holder);
+                        this.emit(EVENTS.LEADERSHIP_LOST);
+                        return;
+                    }
+                    if (this._cycle_abort.signal.aborted) continue;
+                    const status_code = await this.update_lease(lease, false, this._cycle_abort.signal);
+                    if (status_code === 200) {
+                        this._reset_renew_deadline();
+                        sleep_ms = config.CORE_LEASE_ACQUIRE_RETRY_MS;
```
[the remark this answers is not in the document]
nadavMiz: at this point we are holding the lease. we always set this to the same value. this is redundant

src/util/leader_election.js
```diff
@@ -0,0 +1,528 @@
...
+        if (!KUBERNETES_SERVICE_HOST || !KUBERNETES_SERVICE_PORT) {
+            throw new Error('LeaderElector requires in-cluster Kubernetes environment variables');
+        }
+        this._lease_name = lease_name;
+        this._holder = holder;
+        this._service_host = KUBERNETES_SERVICE_HOST;
+        this._service_port = KUBERNETES_SERVICE_PORT;
+        this._initialized = false;
+        this._stop_requested = false;
+        this._lease_duration_ms = config.CORE_LEASE_DURATION_MS;
```
dannyzaken: A small inconsistency: `this._lease_duration_ms` is only used for _get_renew_deadline_ms, but for the `leaseDurationSeconds` config.CORE_LEASE_DURATION_MS is used directly.

src/util/leader_election.js
```diff
@@ -0,0 +1,528 @@
...
+    /**
+     * Computes the renew deadline from the configured lease duration and timing constants.
+     * deadline = CORE_LEASE_DURATION_MS - CORE_LEASE_RENEW_ERROR_SLEEP_MS - CORE_LEASE_REQUEST_TIMEOUT_MS - LEASE_RENEW_DEADLINE_MARGIN_MS
+     * @returns {number} deadline in ms
+     */
+    _get_renew_deadline_ms() {
+        return this._lease_duration_ms -
+            config.CORE_LEASE_RENEW_ERROR_SLEEP_MS -
+            config.CORE_LEASE_REQUEST_TIMEOUT_MS -
+            config.LEASE_RENEW_DEADLINE_MARGIN_MS;
```
dannyzaken: Worth adding a validation that this is > 0

src/util/leader_election.js
```diff
@@ -0,0 +1,528 @@
...
+
+/**
+ * @returns {LeaderElector}
+ */
+function create_elector_from_env() {
+    const lease_name = process.env.NOOBAA_CORE_LEASE_NAME;
+    if (!lease_name) {
+        throw new Error('NOOBAA_CORE_LEASE_NAME is not set');
+    }
+    const holder = `${process.env.HOSTNAME}_${randomUUID()}`;
```
dannyzaken: Just for completeness, validate that `HOSTNAME` is defined (even though this should be injected by Kubernetes automatically)

src/util/leader_election.js
```diff
@@ -0,0 +1,528 @@
...
+        /** @type {number} local clock ms when we last saw the lease record change (clock-skew protection) */
+        this._observed_change_ms = 0;
+        /** @type {Promise<void>|null} */
+        this._running_promise = null;
+        /** @type {NodeJS.Timeout|null} */
+        this._sleep_timer = null;
+        /** @type {(() => void)|null} */
+        this._wake_sleep = null;
+        /** @type {AbortController} single controller reused across renew cycles; clearTimeout prevents abort from firing on success */
+        this._cycle_abort = new AbortController();
```
dannyzaken: Why do we need`this._cycle_abort`?

src/util/leader_election.js
nadavMiz: changed it to always use config.CORE_LEASE_DURATION_MS. seemed more consistent with other config variables

src/util/leader_election.js
nadavMiz: changed it to use "noobaa-core" if undefined. this is not needed for correctness. I think randomUUID() is enough. its so it would be easier for us to understand who hold the lease. and for debugging. so that the holder would be something like:
`noobaa-core-0_8f387f47-653e-44da-b01c-fceffc973c4c`

src/util/leader_election.js
nadavMiz: The reason is that each renewal cycle performs three HTTP requests. In our case, each request has a 5 second timeout, so a single iteration could take up to 1 second of sleep + 3 × 5 seconds for the HTTP requests = 16 seconds before we get a chance to check whether the lease deadline has been exceeded.

The AbortController acts as a timeout for the entire renewal cycle. If the deadline is reached while a request or the sleep is still in progress, the controller aborts it immediately instead of waiting for the current iteration to finish.

I chose AbortController because it can be passed to both HTTP requests and interruptible sleep, so they can be canceled while they are running. It also works well with the 5 second timeout for HTTP requests. whichever timeout fire first aborts the request.

nadavMiz: however I see that the name is not clear. I will change it to this._renew_deadline_abort

src/util/leader_election.js
nadavMiz: right I feel like now that we have the abort controller we don't need this calculation anymore. its also incorrect as the CORE_LEASE_REQUEST_TIMEOUT_MS is per request. I will return it to be a config variable. I will still validate this just in case

nadavMiz:
was replaced by leader election module in the operator https://github.com/noobaa/noobaa-operator/pull/2036
can reopen if needed

PR #9776, opened 2026-06-24
Validate comma-separated CORS preflight headers

kajalpareek-lab:
## Summary

- Fixes bug: [DFBUGS-7973](https://redhat.atlassian.net/browse/DFBUGS-7973)
- Parse `Access-Control-Request-Headers` as a comma-separated list in S3 CORS preflight handling.
- Require every requested header in preflight to match an allowed CORS header rule.
- Add integration tests for allowed and invalid comma-separated preflight header sets.

## Test plan
- [x] Validate CORS preflight `OPTIONS` for comma-separated allowed headers returns 200.
- [x] Validate CORS preflight with one invalid header returns 403.
- [x] Verified backend behavior on live cluster with patched `noobaa-core` image and `curl` preflight checks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved S3 CORS handling for `OPTIONS` requests with comma-separated `Access-Control-Request-Headers`.
  * Requests are now accepted only when every requested header is allowed, which prevents invalid header combinations from being incorrectly permitted.
  * Added coverage for both successful and rejected CORS preflight requests in this scenario.

### DIFF
--- a/src/test/integration_tests/api/s3/test_s3_ops.js
+++ b/src/test/integration_tests/api/s3/test_s3_ops.js
@@ -692,6 +692,79 @@ mocha.describe('s3_ops', function() {
             assert.deepEqual(response && response.statusCode, 200);
         });
 
+        mocha.it('should match OPTIONS request with comma-separated allowed headers', async function() {
+            const request_method = 'PUT';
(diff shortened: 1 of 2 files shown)


shirady (approved):
LGTM

PR #9796, opened 2026-06-30
KeyCloak | Migrate roles from accounts to iam_role schema

naveenpaul1:
### Describe the Problem

Migrate roles from accounts to iam_role schema, 

### Explain the Changes
1. Migrate data from accounts.role_config to iam_roles schema, 
2. Remove object from accounts.role_config after the migration

1. https://redhat.atlassian.net/browse/RHSTOR-9443

Note: If the role with same already exists in iam_role schema, Script will skip the migration for that entry in account.role_config

### Testing Instructions:
1. node ./node_modules/.bin/mocha ./src/test/integration_tests/internal/test_upgrade_scripts.js

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **New Features**
  * During upgrade, legacy IAM role settings are migrated into the centralized role management system, including assume-role policy details and default session duration.

* **Bug Fixes**
  * Assume-role policy `Effect` values are normalized for consistent IAM semantics.
  * Migration is idempotent: already-migrated accounts remain unchanged, and accounts with an existing role of the same name are skipped while preserving the pre-existing role.
  * Legacy role settings are cleared after successful migration.

* **Tests**
  * Added integration tests validating multiple assume-role scenarios, conditional propagation of policy `Version`, and safe re-running of the upgrade.

### DIFF
--- a/src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
+++ b/src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
@@ -1,0 +1,100 @@
+/* Copyright (C) 2023 NooBaa */
+"use strict";
+
+const util = require('util');
+const _ = require('lodash');
+
+const SensitiveString = require('../../../util/sensitive_string');
+const { DEFAULT_MAX_SESSION_DURATION_SECS } = require('../../../endpoint/iam/iam_constants');
+
+// Note: If the role with same already exists in iam_role schema, 
+//      Script will skip the migration for that entry in account.role_config
+
+
+/**
+ * unwarp principle for Principal property
+ *
+ * @param {String[] | SensitiveString[]} principals 
+ * @returns {String[]}
+ */
+function unwrap_principal(principals) {
+    return principals.map(principal =>
+        (principal instanceof SensitiveString ? principal.unwrap() : principal)
+    );
+}
+
+async function run({ dbg, system_store, system_server }) {
+
+    try {
+        dbg.log0('Starting migration of roles from account schema to role schema...');
+        const new_roles = [];
+        const migrated_account_ids = [];
+
+        for (const account of system_store.data.accounts) {
+            //Do not update if there are no role_config.
+            if (!account.role_config) continue;
+
+            const role_config = account.role_config;
+            const new_policy = {};
+
+            const iam_role = system_store.data.iam_roles.find(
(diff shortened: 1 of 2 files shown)


src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
```diff
@@ -0,0 +1,107 @@
...
+            if (role_config.version) new_policy.Version = role_config.assume_role_policy.version;
+            new_policy.Statement = role_config.assume_role_policy.statement.map(statement => ({
+                Effect: statement.effect === 'allow' ? 'Allow' : 'Deny',
+                Action: statement.action,
+                Principal: { AWS: unwrap_principal(statement.principal) },
+                Sid: 'RoleMigration0'
+            }));
+
+            const iam_role = system_store.data.iam_roles.find(
+                a => a.name === role_config.role_name
```
sakshimunjal, line 41: should we add owner check also?
    a => a.name === role_config.role_name && a.owner.equals(account._id)

sakshimunjal (approved):

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
naveenpaul1: I added that condition to make sure we are not adding role with same name, if we add two roles with same name, it return primery key validation error in DB
a.owner.equals(account._id) adding this conditinn allow us to add role with same name when roles are from different accounts, and this will break the migration.

jackyalbo: @naveenpaul1 Do we suppose to have roles already in the system? Don't they suppose to be added in this version as well?

jackyalbo: if we are planning on keeping this check - move it to before you are creating new_policy - as we won't use it in that case - also better readability.

naveenpaul1: @jackyalbo We are creating the iam role in this release only, 
This is to avoide primery key validation error when two or more accounts.role_config with same role_name.
I will move this check before creating new_policy. Thank you

jackyalbo (approved):

PR #9807, opened 2026-07-01
NC | dynamic supplemental groups - dont throw when user not found

nadavMiz:
### Describe the Problem
When dynamic supplemental groups are enabled, get_supplemental_groups_by_uid throws on a UID with no passwd entry (NO_SUCH_USER). supplemental_groups_cache does not cache failures, so every request for that account retries the native lookup and fails again. causing repeated errors and unnecessary syscall load.

Additionally, the S3 path had no graceful handling for supplemental-group lookup failures, so any error could fail the whole request.

### Explain the Changes
1. Treat NO_SUCH_USER as a normal outcome: return [] instead of throwing, so the cache stores the result and stops retrying on every request.
2. Transient/syscall errors are not cached. other errors are logged and rethrown from the helpers, so supplemental_groups_cache does not store a misleading empty result.
3.  unit tests in test_nsfs_access.js covering the empty-array return and that the cache serves the cached [] on subsequent lookups.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Treats “no such user” supplemental-group lookups as non-errors and returns an empty list.
  * When dynamic supplemental groups are enabled, request loading continues if supplemental-group retrieval fails, falling back to an empty set.
  * Supplemental-group caching now avoids caching transient lookup failures.
* **Tests**
  * Added unit tests covering missing-user, lookup failure vs unknown UID, cache hit/invalidation behavior, and dynamic supplemental-group loading scenarios.

### DIFF
--- a/src/util/native_fs_utils.js
+++ b/src/util/native_fs_utils.js
@@ -568,6 +568,7 @@ async function get_user_by_distinguished_name({ distinguished_name }) {
 /**
  * get_supplemental_groups_by_uid - gets supplemental groups by uid via getpwuid_r + getgrouplist.
  * Used by supplemental_groups_cache load when user data is not already available.
+ * Returns [] for NO_SUCH_USER so the cache can store a permanent empty result.
  */
 async function get_supplemental_groups_by_uid({ uid }) {
     try {
@@ -575,6 +576,10 @@ async function get_supplemental_groups_by_uid({ uid }) {
         const groups = await nb_native().fs.getSupplementalGroupsByUid(context, uid);
         return groups;
     } catch (err) {
+        if (err.message === 'NO_SUCH_USER') {
+            dbg.log0('native_fs_utils.get_supplemental_groups_by_uid: uid not found in passwd, uid', uid);
+            return [];
+        }
         dbg.error('native_fs_utils.get_supplemental_groups_by_uid: failed with error', err, err.code, uid);
         throw err;
     }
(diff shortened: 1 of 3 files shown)


src/util/native_fs_utils.js
```diff
@@ -568,13 +568,18 @@ async function get_user_by_distinguished_name({ distinguished_name }) {
...
+ * Returns [] for NO_SUCH_USER so the cache can store a permanent empty result.
  */
 async function get_supplemental_groups_by_uid({ uid }) {
     try {
         const context = get_process_fs_context();
         const groups = await nb_native().fs.getSupplementalGroupsByUid(context, uid);
         return groups;
     } catch (err) {
+        if (err.message === 'NO_SUCH_USER') {
+            dbg.log2('native_fs_utils.get_supplemental_groups_by_uid: uid not found in passwd, uid', uid);
```
jackyalbo: @nadavMiz, the question is whether we want it to be more visible. I guess not having the user in the system shouldn't happen, no?
I'm ok with not throwing excessive errors - but warn, maybe?

jackyalbo (approved):

src/util/native_fs_utils.js
nadavMiz: Not having a passwd entry for the account uid is technically allowed. I changed it to log0 so it will be visable. can move to warn if you think its better

PR #9810, opened 2026-07-06
remove cluster master

nadavMiz:
### Describe the Problem
cluster_master added extra coordination between core processes, even though each process can start on its own once SystemStore is loaded. This made the flow more complex and added an unnecessary startup delay.

### Explain the Changes
1. Remove cluster_master and the old code paths that were only used to update master state between processes.
2. Add system_store.wait_for_load() so each process can wait for `SystemStore` to finish loading in a simple and safe way.
3. Start `BGWorkers`, `HostedAgents`, and `WebServer` directly after `SystemStore` is ready, instead of waiting for cluster_master.
4. Clean up dead code left from the old master coordination flow.

1. https://redhat.atlassian.net/browse/RHSTOR-9434

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved startup reliability by waiting for core system state to finish loading before starting node monitoring and master background workers.
  * Streamlined cluster join/update flow so heartbeats and member IP updates proceed more predictably, including on non-master nodes.

* **Chores**
  * Removed obsolete master-state controls and related inter-process/RPC endpoints, simplifying cluster/master update handling.
  * Cleaned up cluster/master background-worker wiring and trimmed related configuration.


liranmauda (commented):
for [server_inter_process.js](https://github.com/noobaa/noobaa-core/pull/9810/changes#diff-078dc6010871520e868a610b5d51480926ec6f7f26cf76ab0ebe44b94f249888) We can probably move the only left function into `system_server.js` and move the same api into the system_server api.
WDYT?

nadavMiz:
@liranmauda  this file is used for common functions used by all processes. moving this function to `system_server.js` require registering system_api (at least partially) as a common service too (`register_common_services`). I think it would be best to keep this file for the common functions. even if there is only one of those currently

liranmauda (approved):

PR #9821, opened 2026-07-13
KeyCloack | Use existing has_access_policy_permission method in access_policy_utils

naveenpaul1:
### Describe the Problem

This code use existing has_access_policy_permission method in access_policy_utils and remove duplicate code sts_rest

### Explain the Changes
1. Make changes in `sts_rest.authorize_request_policy()` method to use access_policy_utils.has_access_policy_permission()
2. Update role ARN with account id

1. 

### Testing Instructions:
1. 

manual testing
1. Create role for assume role
`iam create-role --role-name test-role-iam3 --assume-role-policy-document file://../../iam_trust_policy_01.json --max-session-duration 3600`
with trust policy
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "Federated": "arn:aws:iam:::oidc-provider/keycloak.noobaa.svc.cluster.local:8080/realms/noobaa"
      },
      "Action": ["sts:AssumeRoleWithWebIdentity", "sts:TagSession"],
      "Condition": {
        "ForAnyValue:StringEquals": {
          "keycloak.noobaa.svc.cluster.local:8080/realms/noobaa:aud": "noobaa-client",
          "aws:RequestTag/Department":"Engineering"
        }
      }
    }
  ]
}
```
2. Call assume-role-with-web-identity API with web-identity-toke that satify the `Principal`, `Action` and `Condition`.
```
 aws sts assume-role-with-web-identity --endpoint https://127.0.0.1:62561 --role-arn arn:aws:iam::6a5f08ac3b67a2001e4ba029:role/test-role-iam3 --role-session-name arn_root --no-verify-ssl --web-identity-toke eyJhbGciOiJSUzI1NiIsInR5cCIgOiAiSldUIiwia2lkIiA6ICJ3Wjl3UmV0MWpia2h6TlVqRGhpdW9
```

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bucket, vector, and permission authorization by using consistent account principal ARN resolution across policy checks.
  * Strengthened assume-role web-identity authorization by decoding web-identity info, enabling Federated/OIDC principal matching, and applying identity-based condition evaluation when applicable.
  * Updated STS web-identity error handling to return the correct general STS error types, and streamlined JWT decoding in token verification/introspection.

### DIFF
--- a/src/endpoint/sts/ops/sts_post_assume_role.js
+++ b/src/endpoint/sts/ops/sts_post_assume_role.js
@@ -33,9 +33,8 @@ async function assume_role(req) {
         AssumeRoleResponse: {
             AssumeRoleResult: {
                 AssumedRoleUser: {
-                    // TODO: change the role arn with account id
-                    Arn: `arn:aws:sts::${assumed_role.access_key}:assumed-role/${assumed_role.role_config.role_name}/${req.body.role_session_name}`,
-                    AssumedRoleId: `${assumed_role.access_key}:${req.body.role_session_name}`
+                    Arn: `arn:aws:sts::${assumed_role.account_id}:assumed-role/${assumed_role.role_config.role_name}/${req.body.role_session_name}`,
+                    AssumedRoleId: `${assumed_role.account_id}:${req.body.role_session_name}`
                 },
                 Credentials: {
                     AccessKeyId: access_keys.access_key.unwrap(),
--- a/src/endpoint/sts/sts_rest.js
+++ b/src/endpoint/sts/sts_rest.js
@@ -1,8 +1,6 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
-const _ = require('lodash');
-const jwt = require('jsonwebtoken');
 
 const dbg = require('../../util/debug_module')(__filename);
 const StsError = require('./sts_errors').StsError;
@@ -118,7 +116,7 @@ async function authenticate_request(req) {
     try {
         signature_utils.authenticate_request_by_service(req, req.sts_sdk);
         if (req.op_name === 'post_assume_role_with_web_identity') {
-            const web_identity_info = fetch_web_identity_info(req);
+            const web_identity_info = access_policy_utils.fetch_web_identity_info(req);
             const is_ldap_request = web_identity_info.username;
             if (is_ldap_request) {
                 // fetch LDAP identity info
@@ -156,8 +154,26 @@ async function authorize_request_policy(req) {
     // system owner by design can always assume role policy of any account
     // skip for NC environments since system owner is not applicable
     if (!is_nc_environment() && (cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
-    const web_identity_info = fetch_web_identity_info(req);
-    const permission = has_assume_role_permission(assume_role_policy, method, req.sts_sdk.requesting_account, web_identity_info);
+
+    // Build the account identifier array (email + ARN + account id) for Principal.AWS matching.
+    const account = req.sts_sdk.requesting_account;
+    const account_arr = [];
+    if (account) {
+        account_arr.push(account.email.unwrap());
+        account_arr.push(access_policy_utils.get_policy_principal_arn(account));
+        account_arr.push(account._id.toString());
+    }
+
+    const permission = await access_policy_utils.has_access_policy_permission(
+        assume_role_policy,
+        account_arr,
+        method,
+        undefined,
+        req,
+        {
+            is_trust_policy: true,
+        }
+    );
     dbg.log0('sts_rest.authorize_request_policy permission is: ', permission);
     if (permission === 'DENY' || permission === 'IMPLICIT_DENY') {
         throw new StsError(StsError.AccessDeniedException);
@@ -214,135 +230,6 @@ function handle_error(req, res, err) {
         res.setHeader('Content-Length', Buffer.byteLength(reply));
     }
     res.end(reply);
-}
-
-function has_assume_role_permission(policy, method, account, web_identity_info) {
-    const [allow_statements, deny_statements] = _.partition(policy.Statement, statement => statement.Effect === 'Allow');
-    // look for explicit denies
-    if (_is_statements_fit(deny_statements, method, account, web_identity_info)) return 'DENY';
-    // look for explicit allows
-    if (_is_statements_fit(allow_statements, method, account, web_identity_info)) return 'ALLOW';
-
-    // implicit deny
-    return 'IMPLICIT_DENY';
-}
-
-function _is_statements_fit(statements, method, account, web_identity_info) {
-    for (const statement of statements) {
-        let action_fit = false;
-        let principal_fit = false;
-        dbg.log0('assume_role_policy: statement', statement);
-
-        // what action can be done
-
-        action_fit = _is_action_fit(method, statement);
-        dbg.log0('assume_role_policy: action fit?', action_fit);
-
-        principal_fit = _is_principal_fit(statement, account, web_identity_info);
-        dbg.log0('assume_role_policy: principal fit?', principal_fit);
-
-        const condition_fit = _is_condition_fit(statement.Condition, web_identity_info);
-        dbg.log0('assume_role_policy: condition fit?', condition_fit);
-
-        dbg.log0('assume_role_policy: is_statements_fit', action_fit, principal_fit, condition_fit);
-        if (action_fit && principal_fit && condition_fit) return true;
-    }
-    return false;
-}
-
-/**
- * _is_principal_fit checks if the statement can be applied to the account
- * @param {Object | Array} statement - The condition(s) from the policy statement
- * @param {Object} account - The account to validate against
- * @param {Object} web_identity_info - The web identity information to validate against
- * @returns {boolean} - true if all principle are satisfied, false otherwise
- */
-function _is_principal_fit(statement, account, web_identity_info) {
-    const statement_principal = statement.Principal || statement.NotPrincipal;
-
-        let principal_fit = false;
-        if (statement_principal.Federated) {
-            for (const federated of _.flatten([statement_principal.Federated])) {
-                const federated_url = typeof federated === 'string' ? federated : federated.unwrap();
-                dbg.log0('assume_role_policy: principal federated fit?', federated_url, web_identity_info.iss);
-                if (federated_url.split("oidc-provider/")[1] === web_identity_info.iss?.split('//')[1]) {
-                    principal_fit = true;
-                }
-            }
-        }
-
-        if (statement_principal.AWS) {
-            for (const principal_aws of _.flatten([statement_principal.AWS])) {
-                const principal = typeof principal_aws === 'string' ? principal_aws : principal_aws.unwrap();
-                const cur_account_email = account?.email.unwrap();
-                // Added ARN to principle validation
-                const account_identifier_arn = account ? access_policy_utils.get_bucket_policy_principal_arn(account) : "";
-                dbg.log0('assume_role_policy: principal fit?', principal, cur_account_email, account_identifier_arn);
-                if ((principal === cur_account_email) || (principal === '*') || (principal === account_identifier_arn)) {
-                    principal_fit = true;
-                }
-            }
-        }
-
-    return statement.Principal ? principal_fit : !principal_fit;
-}
-
-
-/**
- * _is_action_fit checks if the method is allowed by the statement
- * @param {String} method - The account to validate against
- * @param {Object|Array} statement - The condition(s) from the policy statement
- * @returns {boolean} - true if all method are satisfied, false otherwise
- */
-function _is_action_fit(method, statement) {
-    const statement_action = statement.Action || statement.NotAction;
-    let action_fit = false;
-    for (const action of _.flatten([statement_action])) {
-        dbg.log1('access_policy: ', statement.Action ? 'Action' : 'NotAction', ' fit?', action, method);
-        if (action === method || access_policy_utils._is_wildcard_match(action, method)) {
-            action_fit = true;
-            break;
-        }
-    }
-    return statement.Action ? action_fit : !action_fit;
-}
-
-
-/**
- * _is_condition_fit checks if the statement conditions match the web identity info
- * @param {Object|Array} statement_condition - The condition(s) from the policy statement
- * @param {Object} web_identity_info - The web identity information to validate against
- * @returns {boolean} - true if all conditions are satisfied, false otherwise
- */
-function _is_condition_fit(statement_condition, web_identity_info) {
-    if (!statement_condition) return true;
-
-    let statement_conditions = statement_condition;
-    // if the condition is an object, wrap it in an array
-    if (!Array.isArray(statement_condition)) {
-        statement_conditions = [statement_condition];
-    }
-    const is_keycloak_request = web_identity_info.iss;
-    for (const condition of statement_conditions) {
-        const condition_fit = access_policy_utils._is_identity_condition_fit(is_keycloak_request, condition, web_identity_info);
-        if (!condition_fit) {
-            return false;
-        }
-    }
-    return true;
-}
-
-/**
- * fetch web identity object from request web_identity_token param
- * @param {Object} req - Request object
- * @returns {Object} - web_identity_info
- */
-function fetch_web_identity_info(req) {
-    let web_identity_info;
-    if (req.body.web_identity_token) {
-        web_identity_info = jwt.decode(req.body.web_identity_token, { json: true });
-    }
-    return web_identity_info || {};
 }
 
 /**
--- a/src/manage_nsfs/manage_nsfs_validations.js
+++ b/src/manage_nsfs/manage_nsfs_validations.js
@@ -585,7 +585,7 @@ function validate_role_config(user_input) {
         if (statement.effect !== 'allow' && statement.effect !== 'deny') {
             throw_cli_error(ManageCLIError.InvalidRoleConfig, 'effect must be "allow" or "deny"');
         }
-        const valid_actions = ['*', 'sts:AssumeRole', 'sts:AssumeRoleWithWebIdentity', 'sts:AssumeRoleWithSAML', 'sts:*'];
+        const valid_actions = ['*', 'sts:AssumeRole', 'sts:AssumeRoleWithWebIdentity', 'sts:AssumeRoleWithSAML', 'sts:TagSession', 'sts:*'];
         for (const action of statement.action) {
             if (!valid_actions.includes(action)) {
                 throw_cli_error(ManageCLIError.InvalidRoleConfig, 'Policy has invalid action');
--- a/src/test/integration_tests/api/sts/test_sts.js
+++ b/src/test/integration_tests/api/sts/test_sts.js
@@ -103,6 +103,8 @@ mocha.describe('STS tests', function() {
     let sts_creds;
     let accounts = [];
 
+    let user_b_id = '';
+
     /** @type {import('@aws-sdk/client-iam').IAMClient} */
     let iam_client_b;
     const iam_role_name = 'IamCreatedRole1';
@@ -159,6 +161,7 @@ mocha.describe('STS tests', function() {
 
         user_b_key = user_b_keys[0].access_key.unwrap();
         account_info_b = await rpc_client.account.read_account({ email: user_b });
+        user_b_id = account_info_b._id.toString();
 
         // Build an IAM client authenticated as the role-owner account
         iam_client_b = generate_iam_client(
@@ -236,24 +239,24 @@ mocha.describe('STS tests', function() {
 
     mocha.it('user a assume role of admin - should be rejected', async function() {
         await assert_throws_async(sts.assumeRole({
-            RoleArn: `arn:aws:sts::${account_info_b._id.toString()}:role/${'dummy_role'}`,
+            RoleArn: `arn:aws:sts::${user_b_id}:role/${'dummy_role'}`,
             RoleSessionName: 'just_a_dummy_session_name'
         }).promise(), errors.access_denied.code, errors.access_denied.message);
     });
 
     mocha.it('admin assume role of user b - should be allowed', async function() {
         const params = {
-            RoleArn: `arn:aws:sts::${account_info_b._id.toString()}:assumed-role/${role_b}`,
+            RoleArn: `arn:aws:sts::${user_b_id}:assumed-role/${role_b}`,
             RoleSessionName: 'just_a_dummy_session_name'
         };
         const json = await assume_role_and_parse_xml(sts_admin, params);
-        validate_assume_role_response(json, `arn:aws:sts::${user_b_key}:assumed-role/${role_b}/${params.RoleSessionName}`,
-            `${user_b_key}:${params.RoleSessionName}`, user_b_key, defualt_expiry_seconds);
+        validate_assume_role_response(json, `arn:aws:sts::${user_b_id}:assumed-role/${role_b}/${params.RoleSessionName}`,
+            `${user_b_id}:${params.RoleSessionName}`, user_b_key, defualt_expiry_seconds);
     });
 
     mocha.it('admin assume non existing role of user b - should be rejected', async function() {
         await assert_throws_async(sts_admin.assumeRole({
-            RoleArn: `arn:aws:sts::${account_info_b._id.toString()}:role/${'dummy_role2'}`,
+            RoleArn: `arn:aws:sts::${user_b_id}:role/${'dummy_role2'}`,
             RoleSessionName: 'just_a_dummy_session_name1'
         }).promise(), errors.access_denied.code, errors.access_denied.message);
     });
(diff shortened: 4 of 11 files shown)


src/util/keycloak_client.js
```diff
@@ -95,11 +91,7 @@ class KeyCloakClientManager {
...
     async introspect_token(token) {
-        const decoded = require('jsonwebtoken').decode(token);
-        if (!decoded || !decoded.iss) {
-            throw new Error('Invalid token: missing issuer');
-        }
-
+        const decoded = jwt.decode(token);
         const provider = this.get_provider(decoded.iss);
         if (!provider) {
             throw new Error(`No KeyCloak provider configured for issuer: ${decoded.iss}`);
```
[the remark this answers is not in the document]
naveenpaul1: we are already doing the null check in verify_token() method just before the introspect_token() method, no need to duplicate same code here

src/util/access_policy_utils.js
```diff
@@ -162,11 +163,11 @@ async function _is_object_version_fit(req, predicate, value) {
...
  * @param {string[] | string} account
  * @param {string[] | string} method
- * @param {string} resource_arn
- * @param {object} req
+ * @param {string} [resource_arn]
+ * @param {object} [req]
  */
 async function has_access_policy_permission(policy, account, method, resource_arn, req,
-    { disallow_public_access = false, should_pass_principal = true } = {}) {
+    { disallow_public_access = false, should_pass_principal = true, is_assume_role_policy = false } = {}) {
```
aayushchouhan09: Instead of using this role specific flag, we can consider options like `match_resource / condition_mode` (or an explicit request-context map). That keeps `has_access_policy_permission` generic as we add more trust features later.

src/util/access_policy_utils.js
```diff
@@ -261,22 +328,33 @@ function _is_resource_fit(arn_path, statement) {
...
         // When evaluating IAM user inline policies, should_pass_principal is false since these policies
         // don't have a Principal field (the principal is implicitly the user)
-        const principal_fit = should_pass_principal ? _is_principal_fit(account_arr, statement, disallow_public_access) : true;
-        const resource_fit = _is_resource_fit(arn_path, statement);
-        const condition_fit = await _is_condition_fit(statement, req, method);
-
-        dbg.log1('access_policy - is_statements_fit: action_fit, principal_fit, resource_fit, condition_fit', action_fit, principal_fit, resource_fit, condition_fit);
+        let web_identity_info;
+        if (is_assume_role_policy) {
+            web_identity_info = fetch_web_identity_info(req);
```
aayushchouhan09: Decode the web-identity token once in sts_rest, build account_arr + condition request context there, and pass them into `has_access_policy_permission`. Avoid re-decoding from req inside _is_statements_fit.

src/endpoint/sts/sts_rest.js
```diff
@@ -156,8 +154,25 @@ async function authorize_request_policy(req) {
...
-    const web_identity_info = fetch_web_identity_info(req);
-    const permission = has_assume_role_permission(assume_role_policy, method, req.sts_sdk.requesting_account, web_identity_info);
+
+    // Build the account identifier array (email + ARN) for Principal.AWS matching.
+    const account = req.sts_sdk.requesting_account;
+    const account_arr = [];
+    if (account) {
+        account_arr.push(account.email.unwrap());
+        account_arr.push(access_policy_utils.get_policy_principal_arn(account));
+    }
```
aayushchouhan09, line 165: expand `account_arr` for AWS-aligned Principal matching

Today STS passes email + principal ARN. Please also include account id and root ARN (arn:aws:iam::<account>:root) so trust policies using those Principal forms match AWS behavior.

src/util/access_policy_utils.js
```diff
@@ -211,26 +240,64 @@ function _is_action_fit(method, statement) {
...
+ *
+ * @param {string[]} account_arr - Array of account identifiers for the requester
+ * @param {Object} statement - Policy statement containing Principal or NotPrincipal
+ * @param {boolean} [ignore_public_principal=false] - When true, wildcard ('*') principals are
+ *   skipped (used to disallow public access)
+ * @param {Object} [web_identity_info={}] - Web-identity JWT claims (must contain `iss` for
+ *   Federated principal matching)
+ * @returns {boolean} - true if the principal matches, false otherwise
+ */
+function _is_principal_fit(account_arr, statement, ignore_public_principal = false, web_identity_info = {}) {
```
aayushchouhan09: For assume-role policies, statements with `NotPrincipal` should not fit (deny/implicit-deny path), rather than inverting like bucket policies.

src/endpoint/sts/ops/sts_post_assume_role.js
aayushchouhan09, line 17: In sts_post_assume_role.js / sts_post_assume_role_with_web_identity.js, enforce duration after role resolve via sts_utils using min(config.STS_MAX_DURATION_SECONDS, role.max_session_duration) (default 3600).

aayushchouhan09 (commented):
We can try to make the functions more optimized and generic.

Trying the similar changes in #9828 , I will update this PR later with minimal changes. Thanks

src/util/keycloak_client.js
sakshimunjal: please update the function description

src/util/access_policy_utils.js
```diff
@@ -778,16 +872,31 @@ function compare_for_any_value_string_equals(actual, expected) {
...
 exports.has_access_policy_permission = has_access_policy_permission;
 exports.validate_bucket_policy = validate_bucket_policy;
 exports.validate_vector_bucket_policy = validate_vector_bucket_policy;
 exports.allows_public_access = allows_public_access;
-exports.get_bucket_policy_principal_arn = get_bucket_policy_principal_arn;
+exports.get_policy_principal_arn = get_policy_principal_arn;
 exports.create_arn_for_root = create_arn_for_root;
 exports.get_account_identifier_id = get_account_identifier_id;
 exports._is_wildcard_match = _is_wildcard_match;
+exports._is_action_fit = _is_action_fit;
```
sakshimunjal: can we avoid exporting `_is_action_fit` just for unit tests and test via `has_access_policy_permission` with a trust-policy fixture if possible?

src/manage_nsfs/manage_nsfs_validations.js
```diff
@@ -585,7 +585,7 @@ function validate_role_config(user_input) {
         if (statement.effect !== 'allow' && statement.effect !== 'deny') {
             throw_cli_error(ManageCLIError.InvalidRoleConfig, 'effect must be "allow" or "deny"');
         }
-        const valid_actions = ['*', 'sts:AssumeRole', 'sts:AssumeRoleWithWebIdentity', 'sts:AssumeRoleWithSAML', 'sts:*'];
+        const valid_actions = ['*', 'sts:AssumeRole', 'sts:AssumeRoleWithWebIdentity', 'sts:AssumeRoleWithSAML', 'sts:*', 'sts:TagSession'];
```
shirady: Could you add it before the `'sts:*'`?

src/util/keycloak_client.js
```diff
@@ -77,15 +78,10 @@ class KeyCloakClientManager {
...
-        const decoded = require('jsonwebtoken').decode(token);
+        const decoded = jwt.decode(token);
         if (!decoded || !decoded.iss) {
             throw new Error('Invalid token: missing issuer');
         }
-
-        const provider = this.get_provider(decoded.iss);
-        if (!provider) {
-            throw new Error(`No KeyCloak provider configured for issuer: ${decoded.iss}`);
-        }
```
shirady, line 88: Why did you remove this?

src/util/keycloak_client.js
```diff
@@ -95,11 +91,7 @@ class KeyCloakClientManager {
      * @returns {Promise<Object>} - introspection result
      */
     async introspect_token(token) {
-        const decoded = require('jsonwebtoken').decode(token);
-        if (!decoded || !decoded.iss) {
-            throw new Error('Invalid token: missing issuer');
-        }
-
```
shirady, line 102: Why did you remove this?

src/util/access_policy_utils.js
```diff
@@ -162,11 +163,11 @@ async function _is_object_version_fit(req, predicate, value) {
  * @param {object} policy
  * @param {string[] | string} account
  * @param {string[] | string} method
- * @param {string} resource_arn
- * @param {object} req
+ * @param {string} [resource_arn]
+ * @param {object} [req]
```
shirady: Could you add a comment on why those became optional?

src/util/access_policy_utils.js
shirady: Why do we need the `is_assume_role_policy`?

src/util/access_policy_utils.js
```diff
@@ -261,22 +328,33 @@ function _is_resource_fit(arn_path, statement) {
...
+        }
+        const action_fit = _is_action_fit(method, statement, web_identity_info);
+        const principal_fit = should_pass_principal ?
+                        _is_principal_fit(account_arr, statement, disallow_public_access, web_identity_info) : true;
+        const resource_fit = is_assume_role_policy ? true : _is_resource_fit(arn_path, statement);
+        const condition_fit = is_assume_role_policy ?
+            _is_identity_condition_fit(Boolean(web_identity_info.iss), statement.Condition, web_identity_info) :
+            await _is_condition_fit(statement, req, method);
+
+        dbg.log0('access_policy - is_statements_fit: action_fit, principal_fit, resource_fit, condition_fit', action_fit, principal_fit, resource_fit, condition_fit);
```
shirady: Suggesting a minor improvement: (for human debugging)

```suggestion
        dbg.log0('access_policy - is_statements_fit:', 'action_fit' _is_action_fit, 'principal_fit', principal_fit, 'resource_fit' , resource_fit, 'condition_fit', condition_fit);
```

src/util/access_policy_utils.js
```diff
@@ -261,22 +328,33 @@ function _is_resource_fit(arn_path, statement) {
...
+        if (is_assume_role_policy) {
+            web_identity_info = fetch_web_identity_info(req);
+        }
+        const action_fit = _is_action_fit(method, statement, web_identity_info);
+        const principal_fit = should_pass_principal ?
+                        _is_principal_fit(account_arr, statement, disallow_public_access, web_identity_info) : true;
+        const resource_fit = is_assume_role_policy ? true : _is_resource_fit(arn_path, statement);
+        const condition_fit = is_assume_role_policy ?
+            _is_identity_condition_fit(Boolean(web_identity_info.iss), statement.Condition, web_identity_info) :
+            await _is_condition_fit(statement, req, method);
```
shirady: Can it be done inside the `_is_condition_fit` instead of the general function?

src/util/access_policy_utils.js
```diff
@@ -316,6 +394,22 @@ function _parse_condition_keys(condition_statement) {
...
 }
 
+/**
+ * _get_aws_principals resolves the AWS principals from a statement principal object.
+ * When the principal value is a plain string / array (no sub-keys), treat as AWS principal.
+ * This preserves the original behaviour for root-level wildcard ('*') entries.
+ * @param {Object} statement_principal - The Principal or NotPrincipal value from the statement
+ * @returns {string|string[]} - The AWS principal(s)
+ */
+function _get_aws_principals(statement_principal) {
```
shirady: What if I create a bucket policy with `Federated` instead of `AWS`?
It used to fail before, and it should fail after the code changes.

src/test/integration_tests/api/sts/test_sts.js
shirady, line 1: Could you use the same pattern?
For example, in some places it is `account_info_b._id.toString()` directly without `user_b_id` and others with `user_b_id`.

src/util/access_policy_utils.js
naveenpaul1: @aayushchouhan09 `is_assume_role_policy`  isnt required param, If you dont wnated to skip it, its possible and  `is_assume_role_policy` remains generic. checking resource /condition_mode will make it complicated.

src/util/access_policy_utils.js
naveenpaul1: we are not calling fetch_web_identity_info() in STS for authorization flow. This will avoid passing property to this method, and keep the method more generic, align with  your previous comment.

src/endpoint/sts/sts_rest.js
naveenpaul1: get_policy_principal_arn() will create ARN for user or account, Before these changes we where not supporting account id, I dont think we should support account id in future also

src/endpoint/sts/ops/sts_post_assume_role.js
naveenpaul1: not part of this PR,

src/util/access_policy_utils.js
naveenpaul1: I didnt get you, could you please add more context?

src/util/keycloak_client.js
naveenpaul1: There were few duplicate checks in `verify_token()` and `introspect_token()`, With these changes `verify_token()` only verify the token and `introspect_token()` check for its already in provider list

src/util/access_policy_utils.js
naveenpaul1: That was a mistake, will revert that

src/util/access_policy_utils.js
naveenpaul1: When its a trust policy, we dont ned to check the resource_fit and  condition fit check will differ when its from STS.

src/util/access_policy_utils.js
naveenpaul1: yes, your Keycloak/OIDC setup we bucket policy can have `Federated` principal, I am not sure we support it  in NooBaa, 
I will revert back to previous code

src/util/keycloak_client.js
naveenpaul1: There were few duplicate checks in verify_token() and introspect_token(), With these changes verify_token() only verify the token and introspect_token() check for its already in provider list

src/test/unit_tests/util_functions_tests/test_assume_iam_role_trust_policy.test.js
```diff
@@ -724,4 +724,5 @@ describe('IAM Policy Utils - Trust Policy Condition Validation', () => {
             expect(result).toBe(false);
         });
     });
+
```
shirady: I think it is a redundant newline

src/util/access_policy_utils.js
shirady: Why do you use the name `is_assume_role_policy` and not `is_trust_policy`? (like in the explanation here)

src/util/access_policy_utils.js
```diff
@@ -201,7 +204,33 @@ function _is_wildcard_match(action, method) {
...
+ * `sts:AssumeRoleWithWebIdentity`, an additional check is performed: if the JWT
+ * carries session tags the statement must also allow `sts:TagSession` (or `sts:*`);
+ * otherwise the statement does not fit, mirroring AWS behaviour.
+ *
+ * @param {string} method - The action being requested (e.g. 'sts:AssumeRoleWithWebIdentity')
+ * @param {Object} statement - Policy statement containing Action or NotAction
+ * @param {Object} [web_identity_info={}] - Decoded JWT payload from the web identity token
+ * @returns {boolean}
+ */
+function _is_action_fit(method, statement, web_identity_info = {}) {
```
shirady: Can it be in 2 statements?
One for `sts:AssumeRoleWithWebIdentity`
Second one for `sts:TagSession` to be allowed?

naveenpaul1: like this 
```
if (action_fit && method === 'sts:AssumeRoleWithWebIdentity')
        if (_has_session_tags(web_identity_info)) {
```

shirady: In this function, you're working on the same statement.
I was referring to 2 separate statements.

src/util/access_policy_utils.js
```diff
@@ -201,7 +204,33 @@ function _is_wildcard_match(action, method) {
...
+    return Boolean(tags && Object.keys(tags).length > 0);
+}
+
+/**
+ * _is_action_fit checks whether a policy statement's Action (or NotAction) covers
+ * the requested method.
+ *
+ * When `web_identity_info` is supplied and the method is
+ * `sts:AssumeRoleWithWebIdentity`, an additional check is performed: if the JWT
+ * carries session tags the statement must also allow `sts:TagSession` (or `sts:*`);
```
shirady: As I asked, I think it doesn't cover 2 statements: one with (`sts:AssumeRoleWithWebIdentity` and the second with `sts:TagSession`).
In the implementation, you're reading the same statement and checking, what if it is in 2 statements?

src/endpoint/sts/sts_rest.js
aayushchouhan09: For `Principal.AWS` matching user can directly pass the account_id in trust policy and not the ARN. So, we should keep the logic like AWS working. Also, check for other types.

Please check this [aws doc for Principal](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html#principal-roles) , we can check here what user could pass in `Principal.AWS`

src/util/access_policy_utils.js
```diff
@@ -211,13 +240,40 @@ function _is_action_fit(method, statement) {
             break;
         }
     }
+    // AWS requires sts:TagSession to be allowed in the trust policy when the JWT
+    // carries session tags and the requested action is AssumeRoleWithWebIdentity.
+    if (action_fit && method === 'sts:AssumeRoleWithWebIdentity') {
+        if (_has_session_tags(web_identity_info)) {
+            const tag_session_fit = _is_action_fit('sts:TagSession', statement);
```
aayushchouhan09: Can you explain in short why we doing a recursive call here for `_is_action_fit()` and how it works?

src/util/access_policy_utils.js
aayushchouhan09: Make sense, we are calling it once/here only.

src/util/access_policy_utils.js
aayushchouhan09: We should have a condition like:

```suggestion
function _is_principal_fit(account_arr, statement, ignore_public_principal = false, web_identity_info = {}) {
    // Trust policies must not invert NotPrincipal like bucket policies
    // Invert would turn Allow+NotPrincipal into "everyone except X can assume" — treat as non-fit/implicit deny instead
    if (is_trust_policy && statement.NotPrincipal) return false;
```

src/util/access_policy_utils.js
```diff
@@ -214,26 +237,77 @@ function _is_action_fit(method, statement) {
...
+ *   skipped (used to disallow public access)
+ * @param {boolean} [is_trust_policy=false] - Identify IAM trust policy or not
+ * @param {Object} [web_identity_info={}] - Web-identity JWT claims (must contain `iss` for
+ *   Federated principal matching)
+ * @returns {boolean} - true if the principal matches, false otherwise
+ */
+function _is_principal_fit(account_arr, statement, ignore_public_principal = false, is_trust_policy = false, web_identity_info = {}) {
+
+    // Trust policies must not invert NotPrincipal like bucket policies
+    if (is_trust_policy && statement.NotPrincipal) return false;
```
shirady, line 265: As part of the epic (not this PR) - verify AWS behavior with the trust policy with `NotPrincipal` and ensure we behave the same.

src/util/access_policy_utils.js
```diff
@@ -214,26 +237,77 @@ function _is_action_fit(method, statement) {
...
 
+    const statement_principal = statement.Principal || statement.NotPrincipal;
     let principal_fit = false;
-    statement_principal = statement_principal.AWS ? statement_principal.AWS : statement_principal;
-    for (const principal of _.flatten([statement_principal])) {
+
+    // --- AWS principal ---
+    // When the principal value is a plain string / array (no sub-keys), treat as AWS principal.
+    // This preserves the original behaviour for root-level wildcard ('*') entries.
+    const principals = _get_principals(statement_principal);
```
shirady, line 273: I thought it was removed from the code ([link](https://github.com/noobaa/noobaa-core/pull/9821#discussion_r3613659847)) - so I'm asking again. Would it allow a bucket policy with `tatement_principal.Federated` which was not allowed before?

src/util/access_policy_utils.js
```diff
@@ -214,26 +237,77 @@ function _is_action_fit(method, statement) {
...
-
             principal_fit = true;
             break;
         }
     }
+    // --- Federated (OIDC) principal ---
+    if (!principal_fit && statement_principal.Federated && web_identity_info.iss) {
+        for (const federated of _.flatten([statement_principal.Federated])) {
+            const federated_url = typeof federated === 'string' ? federated : federated.unwrap();
+            dbg.log1('access_policy: principal federated fit?', federated_url, web_identity_info.iss);
```
shirady: I think you need to edit the log, the printing doesn't answer if it is fit or not, for example:

```suggestion
            dbg.log1('access_policy: federated url details', federated_url, web_identity_info.iss);
```

src/util/access_policy_utils.js
```diff
@@ -261,40 +335,72 @@ function _is_resource_fit(arn_path, statement) {
...
+ * @param {boolean} needs_tag_session_check - needs tag session check
+ * @param {boolean} tag_session_allowed 
+ * @returns {Promise<boolean>}
+ */
+async function _is_tag_session_fit(needs_tag_session_check, tag_session_allowed) {
+    if (needs_tag_session_check && !tag_session_allowed) return false;
+    return true;
+}
+
+async function _is_condition_fit(policy_statement, req, method, web_identity_info, is_trust_policy) {
```
shirady: Suggesting to add the new arguments in an object with defaults (like in `has_access_policy_permission`).

src/util/access_policy_utils.js
naveenpaul1: yes, It will allow now, because I update the `_get_principals()` 
```
else if (statement_principal.Federated) {
        return statement_principal.Federated;
    }
```

    before it was
    
```
relse if (statement_principal.Federated) {
        return [];
    }
```

src/util/access_policy_utils.js
naveenpaul1: Sure, will check

src/util/access_policy_utils.js
shirady: The question is: should it be allowed for a bucket policy? This should be checked in AWS.

src/util/access_policy_utils.js
```diff
@@ -214,26 +237,72 @@ function _is_action_fit(method, statement) {
...
-    let statement_principal = statement.Principal || statement.NotPrincipal;
+/**
+ * _is_principal_fit checks if the statement principal matches the given account or web identity.
+ *
+ * Handles both bucket policies (Principal.AWS) and assume-role policies (Principal.AWS +
+ * Principal.Federated).  The `account_arr` should contain all identifiers for the requesting
+ * account (email, ARN, account-id, etc.) so that any of them can match a policy principal.
+ *
+ * @param {string[]} account_arr - Array of account identifiers for the requester
+ * @param {Object} statement - Policy statement containing Principal or NotPrincipal
```
shirady: I think you need Yto update the JSDoc here with the new param

src/util/access_policy_utils.js
```diff
@@ -261,40 +330,93 @@ function _is_resource_fit(arn_path, statement) {
...
 
-async function _is_condition_fit(policy_statement, req, method) {
+/**
+ * _is_tag_session_fit check tag session have Action "sts:TagSession" and 
+ * _has_session_tags returns true when the JWT payload contains a non-empty
+ * @param {boolean} needs_tag_session_check - needs tag session check
+ * @param {boolean} tag_session_allowed 
+ * @returns {Promise<boolean>}
+ */
+async function _is_tag_session_fit(needs_tag_session_check, tag_session_allowed) {
```
shirady: Why is the function `async`?

src/util/access_policy_utils.js
```diff
@@ -261,40 +330,93 @@ function _is_resource_fit(arn_path, statement) {
...
     return false;
 }
 
-async function _is_condition_fit(policy_statement, req, method) {
+/**
+ * _is_tag_session_fit check tag session have Action "sts:TagSession" and 
+ * _has_session_tags returns true when the JWT payload contains a non-empty
+ * @param {boolean} needs_tag_session_check - needs tag session check
+ * @param {boolean} tag_session_allowed 
+ * @returns {Promise<boolean>}
```
shirady: Why `Promise`?

src/util/access_policy_utils.js
```diff
@@ -261,40 +330,93 @@ function _is_resource_fit(arn_path, statement) {
...
+ *      Iterates over every operator (e.g. StringEquals, StringLike) and condition key
+ *      (e.g. s3:ExistingObjectTag, s3:x-amz-server-side-encryption, s3:VersionId) in
+ *      the Condition block.
+ *
+ * @param {Object} policy_statement    - A single Statement object from the policy document.
+ * @param {Object} req                 - The incoming HTTP request object.
+ * @param {string} method              - The normalised S3 action being evaluated (e.g. 's3:GetObject').
+ * @returns {Promise<boolean>} - Resolves to true if all conditions are satisfied, false otherwise.
+ */
+async function _is_condition_fit(policy_statement, req, method, { web_identity_info = {}, is_trust_policy = false }) {
```
shirady: I think it missing empty object
```suggestion
async function _is_condition_fit(policy_statement, req, method, { web_identity_info = {}, is_trust_policy = false } = {}) {
```

src/util/access_policy_utils.js
naveenpaul1: Buckey policy alos can have Federated
answer  from claude.
`Federated is a valid principal type in an AWS S3 bucket policy's Principal element — but there's an important nuance for your Keycloak/OIDC setup.`

shirady: Since we don't support it, I think as part of the epic (not this PR) test the behavior when using put-bucket policy, so it works as expected in NooBaa.

src/endpoint/sts/sts_rest.js
```diff
@@ -156,8 +154,26 @@ async function authorize_request_policy(req) {
     // system owner by design can always assume role policy of any account
     // skip for NC environments since system owner is not applicable
     if (!is_nc_environment() && (cur_account_email === _get_system_owner().unwrap()) && req.op_name.endsWith('assume_role')) return;
-    const web_identity_info = fetch_web_identity_info(req);
-    const permission = has_assume_role_permission(assume_role_policy, method, req.sts_sdk.requesting_account, web_identity_info);
+
+    // Build the account identifier array (email + ARN) for Principal.AWS matching.
```
aayushchouhan09: please update the comment here, as per your changes

src/util/access_policy_utils.js
naveenpaul1: That why I skip it first time for normal principal validation, We do not support it now.

shirady: I'm not sure I understand. Where do I see the skip exactly?

naveenpaul1: we where not considering `Principal.Federated` for normal buckt principal, Even though we are adding it should not throw any error, it just consider if anyone provide Federated  in Principal, Thats it, not sure its a issue

shirady: If someone can put the bucket policy with `Federated` and it would be treated like a normal bucket policy, it is a bug. @naveenpaul1 I suggest opening a gap and testing it later, so it would not block this PR.

shirady (approved):
LGTM

I think verification against AWS and tests should be expanded in the future, for example:
- As mentioned in the discussion, the changes in the mutual functions to trust policy and bucket policy need to be retested with `Federated` and make sure that it behaves as it should (we don't support it, so we should not be able to put such a policy).
- `NotPrincipal` in trust policy behavior.
- Test cases for assume role with web identity with tags in multiple statements vs. a single one, etc.

aayushchouhan09 (approved):
LGTM, Thanks!

PR #9823, opened 2026-07-13
block COMPLIANCE to GOVERNANCE retention mode downgrade

kajalpareek-lab:
## Summary
- Reject PutObjectRetention when an object has active COMPLIANCE retention and the request tries to change mode away from COMPLIANCE (e.g. to GOVERNANCE).
- Keep allowing COMPLIANCE date extension, GOVERNANCE → COMPLIANCE upgrade, and new retention after the retain-until date has passed.
- Fix the existing S3 Object Lock test so it uses a non-shorter retain date (otherwise it only exercised "cannot shorten", not mode downgrade), and add unit tests for the retention mode rules.

## Test plan
- [x] Unit tests: `put_object_retention` cases in `test_object_server.test.js` (downgrade blocked, upgrade allowed, extend allowed, expired allowed, shorten blocked)
- [ ] Integration: `test_s3_worm.js` — `should fail to change compliance mode to governance`
- [ ] Manual: PutObjectRetention COMPLIANCE → GOVERNANCE with same/longer date returns AccessDenied; GOVERNANCE → COMPLIANCE still succeeds

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented active COMPLIANCE retention modes from being changed.
  * Preserved protections against shortening or clearing locked retention periods.
  * Improved handling of missing retention requests to avoid errors.
  * Added clearer validation for retention-mode changes.

* **Tests**
  * Added coverage for retention-mode transitions, expiration, date changes, and authorization failures.
  * Verified that governance-locked objects can be upgraded to COMPLIANCE mode successfully.

### DIFF
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -332,50 +332,66 @@ async function get_object_retention(req) {
     };
 }
 /**
- *
  * put_object_retention
- *
  */
 async function put_object_retention(req) {
     dbg.log1('put_object_retention:', req.rpc_params);
-
     throw_if_maintenance(req);
     load_bucket(req);
     const obj = await find_object_md(req);
     const info = get_object_info(obj, { role: req.role });
-
     if (req.role !== 'admin') {
         throw new RpcError('UNAUTHORIZED');
     }
     if (!req.bucket.object_lock_configuration || req.bucket.object_lock_configuration.object_lock_enabled !== 'Enabled') {
         throw new RpcError('INVALID_REQUEST');
     }
-    if (info.lock_settings && info.lock_settings.retention &&
-        (new Date(req.rpc_params.retention.retain_until_date) < new Date(info.lock_settings.retention.retain_until_date) ||
-            !req.rpc_params.retention)) {
-
-        if ((info.lock_settings.retention.mode === 'GOVERNANCE' && (!req.rpc_params.bypass_governance || req.role !== 'admin')) ||
-            info.lock_settings.retention.mode === 'COMPLIANCE') {
-            dbg.error('put object retention failed due object retention mode', obj);
-            throw new RpcError('OBJECT_LOCKED',
-                'Access Denied because object protected by object lock.');
-        }
-    }
-    let legal_hold;
-    if (info.lock_settings && info.lock_settings.legal_hold) {
-        legal_hold = { status: info.lock_settings.legal_hold.status };
-    }
+    const current_retention = info.lock_settings?.retention;
+    const new_retention = req.rpc_params.retention;
+    _throw_if_retention_update_forbidden({
+        key: obj.key,
+        obj_id: info.obj_id,
+        current_retention,
+        new_retention,
+        bypass_governance: Boolean(req.rpc_params.bypass_governance),
+    });
+    const legal_hold_status = info.lock_settings?.legal_hold?.status;
+    const legal_hold = legal_hold_status ? { status: legal_hold_status } : undefined;
     await MDStore.instance().update_object_by_id(
         obj._id, {
             lock_settings: {
                 retention: {
-                    mode: req.rpc_params.retention.mode,
-                    retain_until_date: req.rpc_params.retention.retain_until_date,
+                    mode: new_retention.mode,
+                    retain_until_date: new_retention.retain_until_date,
                 },
                 legal_hold,
             }
         }, undefined, undefined
     );
+}
+
+function _throw_if_retention_update_forbidden({ key, obj_id, current_retention, new_retention, bypass_governance } = {}) {
+    if (!current_retention) return;
+    const current_retain_until = new Date(current_retention.retain_until_date);
+    const log_ctx = { key, obj_id, current_retention, new_retention };
+    // Active COMPLIANCE cannot change mode, even when retain-until stays the same or extends.
+    if (new_retention &&
+        current_retention.mode === 'COMPLIANCE' &&
+        current_retain_until > new Date() &&
+        new_retention.mode !== 'COMPLIANCE') {
+        dbg.error('put object retention failed: cannot change active COMPLIANCE mode', log_ctx);
+        throw new RpcError('OBJECT_LOCKED',
+            'Access Denied because object protected by object lock.');
+    }
+    const date_shortened = !new_retention ||
+        new Date(new_retention.retain_until_date) < current_retain_until;
+    if (!date_shortened) return;
+    if ((current_retention.mode === 'GOVERNANCE' && !bypass_governance) ||
+        current_retention.mode === 'COMPLIANCE') {
+        dbg.error('put object retention failed due object retention mode', log_ctx);
+        throw new RpcError('OBJECT_LOCKED',
+            'Access Denied because object protected by object lock.');
+    }
 }
 
 const ZERO_SIZE_ETAG = crypto.createHash('md5').digest('hex');
--- a/src/test/integration_tests/api/s3/test_s3_worm.js
+++ b/src/test/integration_tests/api/s3/test_s3_worm.js
@@ -745,8 +745,11 @@ mocha.describe('s3 worm', function() {
         });
 
         mocha.it('should fail to change compliance mode to governance', async function() {
-            const futureDate = new Date();
-            futureDate.setDate(futureDate.getDate() + 30);
+            // Previous test extended COMPLIANCE retain-until to +60 days.
+            // Use +90 here (after that date) so this assert fails on mode change
+            // (COMPLIANCE -> GOVERNANCE), not on "cannot shorten COMPLIANCE".
+            const longer_retain_until_date = new Date();
+            longer_retain_until_date.setDate(longer_retain_until_date.getDate() + 90);
 
             await assert_throws_async(s3_owner.putObjectRetention({
                 Bucket: BKT1,
@@ -754,10 +757,45 @@ mocha.describe('s3 worm', function() {
                 VersionId: compliance_version_id,
                 Retention: {
                     Mode: 'GOVERNANCE',
-                    RetainUntilDate: futureDate
+                    RetainUntilDate: longer_retain_until_date
                 },
                 BypassGovernanceRetention: true
             }), 'AccessDenied', 'Access Denied because object protected by object lock.');
+        });
+
+        mocha.it('should allow changing governance mode to compliance', async function() {
+            const GOVERNANCE_KEY = 'governance-to-compliance-test';
+            const retain_until = new Date();
+            retain_until.setDate(retain_until.getDate() + 30);
+
+            const put_res = await s3_owner.putObject({
+                Bucket: BKT1,
+                Key: GOVERNANCE_KEY,
+                Body: file_body,
+                ContentType: 'text/plain',
+                ObjectLockMode: 'GOVERNANCE',
+                ObjectLockRetainUntilDate: retain_until,
+            });
+
+            const res = await s3_owner.putObjectRetention({
+                Bucket: BKT1,
+                Key: GOVERNANCE_KEY,
+                VersionId: put_res.VersionId,
+                Retention: {
+                    Mode: 'COMPLIANCE',
+                    RetainUntilDate: retain_until,
+                },
+                BypassGovernanceRetention: true,
+            });
+            delete res.$metadata;
+            assert.deepEqual(res, {});
+
+            const conf = await s3_owner.getObjectRetention({
+                Bucket: BKT1,
+                Key: GOVERNANCE_KEY,
+                VersionId: put_res.VersionId,
+            });
+            assert.equal(conf.Retention.Mode, 'COMPLIANCE');
         });
     });
 
--- a/src/test/unit_tests/internal/test_object_lock_retention.test.js
+++ b/src/test/unit_tests/internal/test_object_lock_retention.test.js
@@ -1,0 +1,181 @@
+/* Copyright (C) 2026 NooBaa */
+
+'use strict';
+
+const { VERSIONING } = require('../../../common/constants').S3;
+const SensitiveString = require('../../../util/sensitive_string');
+const system_utils = require('../../../server/utils/system_utils');
+const { MDStore } = require('../../../server/object_services/md_store');
+const object_server = require('../../../server/object_services/object_server');
+
+describe('object_lock - put_object_retention', () => {
+    const system_store = require('../../../server/system_services/system_store').get_instance();
+
+    let mdstore_instance_stub;
+    let update_object_by_id_stub;
+    const bucket_id = 'bucket_id_retention';
+    const system_id = 'system_id_retention';
+
+    function make_obj({ mode, retain_until_date }) {
+        const retain = retain_until_date || new Date(Date.now() + 7 * 24 * 3600 * 1000);
+        return {
+            _id: {
+                toHexString: () => 'objidhex',
+                getTimestamp: () => new Date(),
+            },
+            system: system_id,
+            bucket: bucket_id,
+            key: 'locked-key',
+            size: 1,
+            lock_settings: {
+                retention: {
+                    mode,
+                    retain_until_date: retain,
+                },
+            },
+        };
+    }
+
+    function make_req({ mode, retain_until_date, bypass_governance }) {
+        return {
+            role: 'admin',
+            system: {
+                _id: system_id,
+                buckets_by_name: {
+                    'test-bucket': {
+                        _id: bucket_id,
+                        name: 'test-bucket',
+                        versioning: VERSIONING.ENABLED,
+                        object_lock_configuration: { object_lock_enabled: 'Enabled' },
+                    },
+                },
+            },
+            rpc_params: {
+                bucket: new SensitiveString('test-bucket'),
+                key: 'locked-key',
+                bypass_governance,
+                retention: {
+                    mode,
+                    retain_until_date,
+                },
+            },
+        };
+    }
+
+    beforeEach(() => {
+        mdstore_instance_stub = {
+            find_object_latest: jest.fn(),
+            update_object_by_id: jest.fn().mockResolvedValue(undefined),
+            get_object_version_id: jest.fn().mockReturnValue('v1'),
+            make_md_id: jest.fn(),
+        };
+        update_object_by_id_stub = mdstore_instance_stub.update_object_by_id;
+        jest.spyOn(MDStore, 'instance').mockReturnValue(mdstore_instance_stub);
+        jest.spyOn(system_utils, 'system_in_maintenance').mockReturnValue(false);
+        if (!system_store.data) {
+            system_store.data = {};
+        }
+        system_store.data.get_by_id = jest.fn().mockReturnValue({
+            name: new SensitiveString('test-bucket'),
+            versioning: VERSIONING.ENABLED,
+        });
+    });
+
+    afterEach(() => {
+        jest.restoreAllMocks();
+    });
+
+    test('rejects COMPLIANCE to GOVERNANCE downgrade with same retain date', async () => {
+        const retain_until = new Date(Date.now() + 15 * 24 * 3600 * 1000);
+        mdstore_instance_stub.find_object_latest.mockResolvedValue(
+            make_obj({ mode: 'COMPLIANCE', retain_until_date: retain_until })
+        );
+        const req = make_req({
+            mode: 'GOVERNANCE',
+            retain_until_date: retain_until,
+            bypass_governance: true,
+        });
+
+        await expect(object_server.put_object_retention(req))
+            .rejects.toMatchObject({ rpc_code: 'OBJECT_LOCKED' });
+        expect(update_object_by_id_stub).not.toHaveBeenCalled();
+    });
+
+    test('rejects COMPLIANCE to GOVERNANCE downgrade with longer retain date', async () => {
+        const current_until = new Date(Date.now() + 15 * 24 * 3600 * 1000);
+        const longer_until = new Date(Date.now() + 90 * 24 * 3600 * 1000);
+        mdstore_instance_stub.find_object_latest.mockResolvedValue(
+            make_obj({ mode: 'COMPLIANCE', retain_until_date: current_until })
+        );
+        const req = make_req({
+            mode: 'GOVERNANCE',
+            retain_until_date: longer_until,
+            bypass_governance: true,
+        });
+
+        await expect(object_server.put_object_retention(req))
+            .rejects.toMatchObject({ rpc_code: 'OBJECT_LOCKED' });
+        expect(update_object_by_id_stub).not.toHaveBeenCalled();
+    });
+
+    test('allows GOVERNANCE to COMPLIANCE upgrade with same retain date', async () => {
+        const retain_until = new Date(Date.now() + 15 * 24 * 3600 * 1000);
+        mdstore_instance_stub.find_object_latest.mockResolvedValue(
+            make_obj({ mode: 'GOVERNANCE', retain_until_date: retain_until })
+        );
+        const req = make_req({
+            mode: 'COMPLIANCE',
+            retain_until_date: retain_until,
+        });
+
+        await expect(object_server.put_object_retention(req)).resolves.toBeUndefined();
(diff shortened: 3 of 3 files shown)


src/server/object_services/object_server.js
```diff
@@ -348,15 +348,29 @@ async function put_object_retention(req) {
...
-            info.lock_settings.retention.mode === 'COMPLIANCE') {
-            dbg.error('put object retention failed due object retention mode', obj);
+    const current_retention = info.lock_settings && info.lock_settings.retention;
+    if (current_retention) {
+        const now = new Date();
+        const current_retain_until = new Date(current_retention.retain_until_date);
+        const new_retention = req.rpc_params.retention;
+
+        // AWS: while COMPLIANCE retention is active, mode cannot change (e.g. to GOVERNANCE).
+        // Must run outside the shorten-date check — same/longer dates previously bypassed it.
```
nadavMiz: may be personal, but I feel like this comment is not very understandable. what does it mean: "Must run outside the shorten-date check — same/longer dates previously bypassed it."?
I feel like it should be more intuitive to someone that doesn't necessarily understand object lock semantics. something like:
"Cannot change the COMPLIANCE mode when replacing an existing policy with a new one"

src/server/object_services/object_server.js
nadavMiz, line 335: I think its worth adding a small comment explaining the function. and generally what it allows and disallows

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -745,16 +745,18 @@ mocha.describe('s3 worm', function() {
         });
 
         mocha.it('should fail to change compliance mode to governance', async function() {
-            const futureDate = new Date();
-            futureDate.setDate(futureDate.getDate() + 30);
+            // Use a date >= current retain-until (extended to +60 above).
+            // A shorter date would fail for the wrong reason (cannot shorten COMPLIANCE).
+            const longerOrEqualDate = new Date();
+            longerOrEqualDate.setDate(longerOrEqualDate.getDate() + 90);
```
nadavMiz: why is this change needed? why does it matters if we are 30 or 90 days in the future?
also its never an equal date

src/test/unit_tests/internal/test_object_server.test.js
```diff
@@ -543,3 +543,175 @@ describe('object_server - update_bulk_delete_results', () => {
         expect(results[3]).toHaveProperty('seq', 103);
     });
 });
+
+describe('object_server - put_object_retention', () => {
```
nadavMiz: I don't dislike the idea of adding unit tests. but we are already using integration tests for object lock. I think it would be good to have this coverage in test_s3_worm.js. the full path doesn't add much complexity, and I feel like its worth testing. we would also get the added benefit of testing NC as well.

having said that if you still feel like unit tests are needed we should probably move them to their own file for object_lock. no one would expect object lock tests to be here, and I fear they would be missed and potentially recreated elsewhere

src/server/object_services/object_server.js
```diff
@@ -348,15 +348,29 @@ async function put_object_retention(req) {
...
+        if (new_retention &&
+            current_retention.mode === 'COMPLIANCE' &&
+            current_retain_until > now &&
+            new_retention.mode !== 'COMPLIANCE') {
+            dbg.error('put object retention failed: cannot change active COMPLIANCE mode', obj);
             throw new RpcError('UNAUTHORIZED');
         }
+
+        if (!new_retention || new Date(new_retention.retain_until_date) < current_retain_until) {
+            if ((current_retention.mode === 'GOVERNANCE' && (!req.rpc_params.bypass_governance || req.role !== 'admin')) ||
```
nadavMiz: worth adding a TODO that bypass_governance should be allowed to users with IAM permission and not specificly to admin

src/test/integration_tests/api/s3/test_s3_worm.js
nadavMiz: ahh its because of the previous test. then its  a good fix. you should just change the name of the variable

src/server/object_services/object_server.js
shirady: @nadavMiz I think the `req.role` admin is not related to the IAM users.
@kajalpareek-lab I don't think you need to check the `req.role` again, as it is already checked in the function before:
```
   if (req.role !== 'admin') {
        throw new RpcError('UNAUTHORIZED');
    }
```

and anyway I think that you can test the feature of object lock with IAM users vs root users to see you have the expected results according to AWS and if not fix accordingly.

src/server/object_services/object_server.js
```diff
@@ -348,15 +348,29 @@ async function put_object_retention(req) {
...
+                current_retention.mode === 'COMPLIANCE') {
+                dbg.error('put object retention failed due object retention mode', obj);
```
shirady: Could you add the `new_retention` in the logs? something like:

```suggestion
                dbg.error('put object retention failed due object retention mode', obj, 'new_retention', new_retention);
```

Could you share what we see in this log?
I want to see that printing the whole `obj` is not overwhelming (since the interesting details are the object name and the retention mode).

src/test/integration_tests/api/s3/test_s3_worm.js
shirady: @kajalpareek-lab could you add a comment about the days so it would be clear for others? (the dependency between the tests).

shirady (commented):
I agree with @nadavMiz comments and added a few comments

src/server/object_services/object_server.js
kajalpareek-lab: Updated. Changed it to:
"Cannot change COMPLIANCE mode when replacing an existing retention policy (e.g. COMPLIANCE -> GOVERNANCE) while retention is still active."
Also kept a short note that this check is separate from date shortening so a longer retain-until cannot bypass the mode restriction.

src/server/object_services/object_server.js
kajalpareek-lab: @nadavMiz Added a TODO that BypassGovernanceRetention should be allowed based on IAM / bucket-policy permission, not specifically the admin role.

src/test/unit_tests/internal/test_object_server.test.js
kajalpareek-lab: Agreed. Moved the unit tests out of `test_object_server.test.js`. I’ll also add the COMPLIANCE -> GOVERNANCE / allowed transition coverage to `test_s3_worm.js` so we exercise the full S3 path (including NC).

src/server/object_services/object_server.js
kajalpareek-lab: Added a short JSDoc on put_object_retention describing what it allows/disallows for COMPLIANCE vs GOVERNANCE while retention is active.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: @nadavMiz Yes — needed because the previous test extends retain-until to +60 days, so this one must use a later date (+90) so the failure is on COMPLIANCE -> GOVERNANCE mode change, not on shortening.
Renamed the variable to `longerRetainUntilDate` and added a comment explaining the dependency between the two tests.

src/server/object_services/object_server.js
kajalpareek-lab: @shirady Removed the redundant admin check in the shorten/bypass path. Admin is already enforced earlier in the function. Agree that IAM-user vs root-user Object Lock coverage is useful; I’d prefer to handle that as a follow-up outside this COMPLIANCE downgrade fix.

src/server/object_services/object_server.js
kajalpareek-lab: Updated the failure logs to include `new_retention`, for example:
`dbg.error('put object retention failed: cannot change active COMPLIANCE mode', obj, 'new_retention', new_retention);`
If logging the full `obj` is too noisy, I can reduce it to key fields (object name / id and current retention). Let me know if you want that in this PR.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: @shirady Added a comment explaining the dependency between the tests: previous test sets +60, so this one uses +90 to assert mode-change denial rather than date shortening.

src/server/object_services/object_server.js
shirady: Could you attach what you see in the logs? so we can see if it's too noisy

kajalpareek-lab: Full obj was noisy, so I changed the log to only key / obj_id / current_retention / new_retention. Example below — that should be enough without dumping the whole object_md. 

`Aug-3 14:45:19.056 [/8928] [ERROR] core.server.object_services.object_server:: put object retention failed: cannot change active COMPLIANCE mode { key: 'locked-key', obj_id: 'objidhex', current_retention: { mode: 'COMPLIANCE', retain_until_date: 2026-08-18T09:15:19.055Z }, new_retention: { mode: 'GOVERNANCE', retain_until_date: 2026-08-18T09:15:19.055Z } }`

src/server/object_services/object_server.js
```diff
@@ -331,9 +331,13 @@ async function get_object_retention(req) {
     };
 }
 /**
- *
  * put_object_retention
  *
+ * Sets or updates object retention (mode + retain-until date) when Object Lock is enabled.
+ * While retention is active:
+ * - COMPLIANCE: cannot shorten retain-until, and cannot change mode (e.g. to GOVERNANCE).
+ * - GOVERNANCE: cannot shorten retain-until unless bypass_governance is set.
```
shirady: Can we change the mode from GOVERNANCE -> GOVERNANCE?
It is implicit from the JSDoc
If so, could you write this explicitly?

src/server/object_services/object_server.js
```diff
@@ -343,21 +347,50 @@ async function put_object_retention(req) {
     const obj = await find_object_md(req);
     const info = get_object_info(obj, { role: req.role });
 
+    // TODO: BypassGovernanceRetention should be allowed for callers with IAM /
+    // bucket-policy permission, not only the admin role.
```
shirady: This TODO is not related to the check `if (req.role !== 'admin')`; this is the `auth` of the `RPC`.

src/server/object_services/object_server.js
```diff
@@ -343,21 +347,50 @@ async function put_object_retention(req) {
...
+        const new_retention = req.rpc_params.retention;
+
+        // Cannot change COMPLIANCE mode when replacing an existing retention policy
+        // (e.g. COMPLIANCE -> GOVERNANCE) while retention is still active.
+        // Checked separately from date shortening so same/longer retain-until dates cannot bypass it.
+        if (new_retention &&
+            current_retention.mode === 'COMPLIANCE' &&
+            current_retain_until > now &&
+            new_retention.mode !== 'COMPLIANCE') {
+            // Log only retention-relevant fields; full object_md is too noisy.
```
shirady: I don't think we need this comment

src/server/object_services/object_server.js
```diff
@@ -343,21 +347,50 @@ async function put_object_retention(req) {
...
+            dbg.error('put object retention failed: cannot change active COMPLIANCE mode', {
+                key: obj.key,
+                obj_id: obj._id && (obj._id.toHexString ? obj._id.toHexString() : String(obj._id)),
```
shirady: Why do you need the object_id in the logs?
Why use the `obj` when you have the `info` (from `get_object_info`) and you can easily use: `info.obj_id`?

src/server/object_services/object_server.js
```diff
@@ -343,21 +347,50 @@ async function put_object_retention(req) {
...
+                current_retention,
+                new_retention,
+            });
             throw new RpcError('UNAUTHORIZED');
         }
+
+        if (!new_retention || new Date(new_retention.retain_until_date) < current_retain_until) {
+            // req.role === 'admin' was already enforced above.
```
shirady: I don't think we need that comment

src/server/object_services/object_server.js
```diff
@@ -343,21 +347,50 @@ async function put_object_retention(req) {
...
+            if ((current_retention.mode === 'GOVERNANCE' && !req.rpc_params.bypass_governance) ||
+                current_retention.mode === 'COMPLIANCE') {
+                dbg.error('put object retention failed due object retention mode', {
+                    key: obj.key,
+                    obj_id: obj._id && (obj._id.toHexString ? obj._id.toHexString() : String(obj._id)),
+                    current_retention,
+                    new_retention,
+                });
+                throw new RpcError('UNAUTHORIZED');
```
shirady: Could you check the error message in AWS for those cases to see if we have the same error message?

src/test/unit_tests/internal/test_object_lock_retention.test.js
shirady, line 1: What are this new tests add that we do not cover in `src/test/unit_tests/internal/test_object_lock_retention.test.js`?

src/server/object_services/object_server.js
kajalpareek-lab: Yes — same-mode updates are allowed when retain-until is the same or later. Updated the JSDoc to call out COMPLIANCE → COMPLIANCE and GOVERNANCE → GOVERNANCE explicitly.

src/server/object_services/object_server.js
kajalpareek-lab: Agree — that TODO is about BypassGovernanceRetention permission, not RPC admin auth. Moved it next to the GOVERNANCE shorten / bypass check.

src/server/object_services/object_server.js
kajalpareek-lab: Removed.

src/server/object_services/object_server.js
kajalpareek-lab: Keeping obj_id so we can correlate retention denials with a specific object version in support logs (key alone isn’t enough when there are multiple versions). Switched to info.obj_id from get_object_info.

src/server/object_services/object_server.js
kajalpareek-lab: removed.

src/test/unit_tests/internal/test_object_lock_retention.test.js
kajalpareek-lab: The unit file keeps the edge cases that are awkward or slow over S3: same-date downgrade (proves mode check isn’t only hit via shorten), longer-date downgrade, expired COMPLIANCE then GOVERNANCE, and isolated put_object_retention without a cluster. Per Nadav’s earlier note we moved them out of test_object_server.test.js into this dedicated file.

src/server/object_services/object_server.js
kajalpareek-lab: Checked on a live hosted cluster with this change: PutObjectRetention denials (COMPLIANCE → GOVERNANCE and shortening COMPLIANCE) return AccessDenied / Access Denied, same as other Object Lock denials and what our worm tests already expect for hosted. That matches AWS’s AccessDenied behavior for these cases. I’d leave the message as-is in this PR.

src/server/object_services/object_server.js
```diff
@@ -349,15 +354,42 @@ async function put_object_retention(req) {
...
+                obj_id: info.obj_id,
+                current_retention,
+                new_retention,
+            });
             throw new RpcError('UNAUTHORIZED');
         }
+
+        if (!new_retention || new Date(new_retention.retain_until_date) < current_retain_until) {
+            // TODO: BypassGovernanceRetention should follow IAM / bucket-policy permission
+            // (not only callers that already passed RPC admin auth above).
```
shirady: It is not related to this code block, right?
The bucket policy and IAM user inline policy were authorized in `rest_s3.
Could you remove the comment?

src/test/unit_tests/internal/test_object_lock_retention.test.js
```diff
@@ -0,0 +1,181 @@
+/* Copyright (C) 2026 NooBaa */
+
+'use strict';
+
+const CONSTANTS = require('../../../common/constants');
```
shirady: Can you destruct and only have the VERSIONING so you can use it?

src/test/unit_tests/internal/test_object_lock_retention.test.js
```diff
@@ -0,0 +1,181 @@
...
+        const req = make_req({
+            mode: 'GOVERNANCE',
+            retain_until_date: new_until,
+        });
+
+        await expect(object_server.put_object_retention(req)).resolves.toBeUndefined();
+        expect(update_object_by_id_stub).toHaveBeenCalled();
+    });
+
+    test('still rejects shortening active COMPLIANCE retention', async () => {
```
shirady: Why is the name with "still"?
```suggestion
    test('rejects shortening active COMPLIANCE retention', async () => {
```

src/server/object_services/object_server.js
```diff
@@ -331,9 +331,14 @@ async function get_object_retention(req) {
...
  * put_object_retention
  *
+ * Sets or updates object retention (mode + retain-until date) when Object Lock is enabled.
+ * While retention is active:
+ * - COMPLIANCE: cannot shorten retain-until, and cannot change mode (e.g. to GOVERNANCE).
+ * - GOVERNANCE: cannot shorten retain-until unless bypass_governance is set.
+ * - Same mode is allowed: COMPLIANCE -> COMPLIANCE and GOVERNANCE -> GOVERNANCE
+ *   when retain-until is the same or later (extend / no-op date).
  */
 async function put_object_retention(req) {
```
shirady, line 337: BTW, with all the changes, the function became quite long, if you can reduce the number of lines.

src/server/object_services/object_server.js
shirady: Why do you need to move it?
When you add the fix, would it be in the `object_server`?
I think you can remove it from the code, and if you want, you can document it in the PR description.

src/server/object_services/object_server.js
shirady: I meant the error message, not just the error code.

src/server/object_services/object_server.js
kajalpareek-lab: Removed.

src/server/object_services/object_server.js
kajalpareek-lab: Extracted _throw_if_retention_update_forbidden so put_object_retention is the auth/load/update path only. Behavior unchanged validation lives in one helper.

src/test/unit_tests/internal/test_object_lock_retention.test.js
kajalpareek-lab: Done

src/test/unit_tests/internal/test_object_lock_retention.test.js
kajalpareek-lab: Renamed.

src/server/object_services/object_server.js
kajalpareek-lab: Removed. BypassGovernanceRetention IAM/bucket-policy auth belongs in s3_rest / #9881, not in this retention validation path.

src/server/object_services/object_server.js
kajalpareek-lab: You’re right — I was looking at the error code only.
Hosted today still returns "Access Denied" for these PutObjectRetention denials (UNAUTHORIZED). Aligning the message to "Access Denied because object protected by object lock." (and mapping via OBJECT_LOCKED / AccessDeniedObjectLocked) is already covered in #9868, including the worm test updates for both hosted and NC.
I’ll leave the message as-is in this PR and rebase onto #9868 after it merges.

nadavMiz (approved):

nadavMiz:
looks like error message is still wrong. after that is fixed. its fine by me

shirady:
> looks like error message is still wrong. after that is fixed. its fine by me

@nadavMiz @kajalpareek-lab plans to fix it in another PR - see [comment](https://github.com/noobaa/noobaa-core/pull/9823#discussion_r3763169012).

src/server/object_services/object_server.js
```diff
@@ -349,33 +354,71 @@ async function put_object_retention(req) {
...
+        key: obj.key,
+        obj_id: info.obj_id,
+        current_retention,
+        new_retention,
+        bypass_governance: Boolean(req.rpc_params.bypass_governance),
+    });
+
+    const legal_hold = info.lock_settings?.legal_hold ?
+        { status: info.lock_settings.legal_hold.status } :
+        undefined;
```
shirady: Why do you need a ternary conditional operator?
I think with the optional chining that you're using, if we don't have the value, the `legal_hold` value will be `undefined`, so you can continue the optional chaining for `status`, unless I'm missing something...

```suggestion
    const legal_hold = info.lock_settings?.legal_hold?.status
```

src/server/object_services/object_server.js
```diff
@@ -349,33 +354,71 @@ async function put_object_retention(req) {
...
+/**
+ * Enforce Object Lock retention update rules for an existing active retention.
+ */
+function _throw_if_retention_update_forbidden({
+    key,
+    obj_id,
+    current_retention,
+    new_retention,
+    bypass_governance,
+}) {
```
shirady: I think that without defaulting the object when you destruct, calling the function without an object is prone to the error `TypeError`.
I think you can add it:

```suggestion
function _throw_if_retention_update_forbidden({
    key,
    obj_id,
    current_retention,
    new_retention,
    bypass_governance,
} = {}) {
```

src/server/object_services/object_server.js
```diff
@@ -2688,6 +2731,6 @@ exports.update_transition_status = update_transition_status;
 
 if (process.env.NODE_ENV === 'test') {
     exports.__testing = {
-        update_bulk_delete_results
+        update_bulk_delete_results,
```
shirady, line 2790: No change here, right?
So you can remove this addition (fewer changed lines).

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -745,20 +745,58 @@ mocha.describe('s3 worm', function() {
         });
 
         mocha.it('should fail to change compliance mode to governance', async function() {
-            const futureDate = new Date();
-            futureDate.setDate(futureDate.getDate() + 30);
+            // Previous test extended COMPLIANCE retain-until to +60 days.
+            // Use +90 here (after that date) so this assert fails on mode change
+            // (COMPLIANCE -> GOVERNANCE), not on "cannot shorten COMPLIANCE".
+            const longerRetainUntilDate = new Date();
```
shirady: Our code style is snake case:
```suggestion
            const longer_retain_until_date = new Date();
```

src/test/unit_tests/internal/test_object_lock_retention.test.js
shirady: @nadavMiz by using this file it would not be tested in NC, could you explain why not adding the test cases in `src/test/integration_tests/api/s3/test_s3_worm.js`?

I'm trying to look at the test cases, and I think it is good coverage for both environments:
- `rejects COMPLIANCE to GOVERNANCE downgrade with same retain date`
- `rejects COMPLIANCE to GOVERNANCE downgrade with longer retain date`
- `'allows GOVERNANCE to COMPLIANCE upgrade with same retain date`
- `allows extending COMPLIANCE retention while keeping COMPLIANCE mode`
- `allows setting GOVERNANCE after COMPLIANCE retention has expired`
- `rejects shortening active COMPLIANCE retention`

src/server/object_services/object_server.js
kajalpareek-lab: removed the old ternary form and now use optional chaining for the status read, then build the legal_hold object from that value.

src/server/object_services/object_server.js
kajalpareek-lab: Agreed. Added a default object in the helper signature (= {}) so accidental direct calls won’t throw on destructuring.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: Agreed. Renamed to snake_case: + "longer_retain_until_date" + .

PR #9824, opened 2026-07-14
Keycloak | IAM Roles caching

VershaAgrawal:
### Describe the Problem
Implement IAM roles caching 

### Explain the Changes
1. Added an in-memory cache for IAM roles on the endpoint, keyed by role name and owner id .
2. Added a new internal RPC read_role_by_name so the endpoint can fetch a single role from core.
3. Added ObjectSDK.get_iam_role_by_name(role_name, owner_account_id) so S3/STS callers can read roles through the cache.

1. https://redhat.atlassian.net/browse/RHSTOR-9422

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added admin-protected role lookup by name in the account API (`read_role_by_name`), exposed in the SDK as `get_iam_role_by_name`.
  * Introduced an in-SDK IAM role cache to speed up repeated lookups, with configurable expiry (default: 1 minute).
* **Bug Fixes**
  * Role changes (update/delete/policy updates) now automatically refresh cached role data.
* **Tests**
  * Added unit tests covering cache miss/hit behavior, cache invalidation reloads, and missing-`owner_account_id` authorization handling.

### DIFF
--- a/config.js
+++ b/config.js
@@ -712,6 +712,8 @@ config.OBJECT_SDK_BUCKET_CACHE_EXPIRY_MS = 60000;
 config.OBJECT_SDK_ACCOUNT_CACHE_EXPIRY_MS = Number(process.env.ACCOUNTS_CACHE_EXPIRY) || 10 * 60 * 1000; // TODO: Decide on a time that we want to invalidate
 // Accountspace_fs account id cache expiration time
 config.ACCOUNTS_ID_CACHE_EXPIRY = 3 * 60 * 1000; // TODO: Decide on a time that we want to invalidate
+// IAM ListRoles cache expiration time (containerized endpoint)
+config.IAM_ROLES_CACHE_EXPIRY_MS = Number(process.env.IAM_ROLES_CACHE_EXPIRY) || 10 * 60 * 1000;
 
 // Nodes identity cache (list_nodes_by_identity per-node entries)
 config.NODES_IDENTITY_CACHE_MAX = 100;
--- a/src/sdk/accountspace_nb.js
+++ b/src/sdk/accountspace_nb.js
@@ -3,6 +3,7 @@
 
 const account_util = require('../util/account_util');
 const { IAM_DEFAULT_PATH} = require('../endpoint/iam/iam_constants');
+const { iam_roles_cache } = require('./object_sdk');
 
 /**
  * @implements {nb.AccountSpace}
@@ -146,11 +147,21 @@ class AccountSpaceNB {
     }
 
     async update_role(params, account_sdk) {
-        return account_sdk.rpc_client.account.update_role(params);
+        const result = await account_sdk.rpc_client.account.update_role(params);
+        iam_roles_cache.invalidate({
+            role_name: params.role_name,
+            owner_account_id: String(account_sdk.requesting_account._id),
+        });
+        return result;
     }
 
     async delete_role(params, account_sdk) {
-        return account_sdk.rpc_client.account.delete_role(params);
+        const result = await account_sdk.rpc_client.account.delete_role(params);
+        iam_roles_cache.invalidate({
+            role_name: params.role_name,
+            owner_account_id: String(account_sdk.requesting_account._id),
+        });
+        return result;
     }
 
     async list_roles(params, account_sdk) {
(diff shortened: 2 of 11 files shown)


aayushchouhan09:
In case if we need to redesign the schema for roles (containerized) to make it align with NC and if we put iam_roles inside account schema as a property then this PR is not making sense, we need to make a lot of changes later. 

@jackyalbo @sakshimunjal  Could you please confirm if we can proceed with this.

cc @naveenpaul1 @shirady

Please feel free to correct me. Thanks!

src/sdk/accountspace_nb.js
```diff
@@ -174,7 +179,9 @@ class AccountSpaceNB {
     }
 
     async update_assume_role_policy(params, account_sdk) {
-        return account_sdk.rpc_client.account.update_assume_role_policy(params);
+        const result = await account_sdk.rpc_client.account.update_assume_role_policy(params);
+        iam_roles_cache.invalidate_key(params.role_name);
```
naveenpaul1: Do we need invalidate_key for update_assume_role_policy()?

src/sdk/object_sdk.js
```diff
@@ -340,6 +357,19 @@ class ObjectSDK {
         }
     }
 
+    /**
```
naveenpaul1, line 362: Please add comments

src/sdk/object_sdk.js
```diff
@@ -340,6 +357,19 @@ class ObjectSDK {
...
+     * @param {string} role_name
+     * @returns {Promise<object>}
+     */
+    async get_iam_role_by_name(role_name) {
+        const owner_account_id = String(this.requesting_account._id);
```
naveenpaul1: For AssumeRoleWithWebIdentity, there is no this.requesting_account, we need to pass the owner_account_id, otherwise it will fail.

src/sdk/accountspace_nb.js
VershaAgrawal: The cache stores the full role object (role_info). That object includes the trust policy (assume_role_policy_document).

UpdateAssumeRolePolicy changes that trust policy in the database. If we leave the old entry in cache, the endpoint can keep using the previous trust policy until the TTL expires. That can wrongly allow or deny assume-role requests.

src/sdk/object_sdk.js
VershaAgrawal: added

naveenpaul1 (approved):
LGTM

config.js
```diff
@@ -712,6 +712,8 @@ config.OBJECT_SDK_BUCKET_CACHE_EXPIRY_MS = 60000;
 config.OBJECT_SDK_ACCOUNT_CACHE_EXPIRY_MS = Number(process.env.ACCOUNTS_CACHE_EXPIRY) || 10 * 60 * 1000; // TODO: Decide on a time that we want to invalidate
 // Accountspace_fs account id cache expiration time
 config.ACCOUNTS_ID_CACHE_EXPIRY = 3 * 60 * 1000; // TODO: Decide on a time that we want to invalidate
+// IAM ListRoles cache expiration time (containerized endpoint)
+config.IAM_ROLES_CACHE_EXPIRY_MS = Number(process.env.IAM_ROLES_CACHE_EXPIRY) || 1 * 60 * 1000;
```
aayushchouhan09: Why default is 1minute here? Should we increase it?

For accounts cache we have default time of 10minutes (for containerized) and 3minutes for NC.

src/sdk/object_sdk.js
```diff
@@ -65,6 +65,23 @@ const account_cache = new LRUCache({
...
+    expiry_ms: config.IAM_ROLES_CACHE_EXPIRY_MS,
+    /**
+     * Set type for the generic template
+     * @param {{
+     *      role_name: string;
+     *      owner_account_id: string;
+     *      bucketspace: nb.BucketSpace;
+     * }} params
+     */
+    make_key: ({ role_name }) => role_name,
```
aayushchouhan09: Why `role_name` only and not `owner_id` here?

As per our discussion,  roles are unique per account, not globally. We should use both I think as a composite key.

aayushchouhan09 (commented):
Some comments

aayushchouhan09 (changes requested):

aayushchouhan09 (approved):
LGTM
Two comments/questions:
1. Can you please update the PR description as per the current changes.
2. Are we covering the use of iam_roles cache in IAM roles functions in the same PR?

VershaAgrawal:
@aayushchouhan09 
> 1. Can you please update the PR description as per the current changes. --> Updated
> 2. Are we covering the use of iam_roles cache in IAM roles functions in the same PR? --> We will add it in next PR

PR #9827, opened 2026-07-16
Remove server.js which is not used and `express ` mock from `test_cli_diagnose.test.js`

liranmauda:
### Explain the Changes

Remove server.js which is not used and `express ` mock from `test_cli_diagnose.test.js`

## Summary by CodeRabbit

- **Tests**
  - Updated CLI diagnostics test infrastructure to use a lightweight built-in HTTP server.
  - Improved mock metrics endpoint behavior to serve only the expected metrics path and return `404` for other URLs.
  - Removed the unused example server setup.


src/test/integration_tests/nc/cli/test_cli_diagnose.test.js
```diff
@@ -83,25 +83,31 @@ describe('noobaa cli - diagnose flow', () => {
...
+            res.writeHead(200, { 'Content-Type': 'application/json' });
+            res.end(JSON.stringify(metrics_obj_mock));
+            return;
+        }
+        res.writeHead(404).end();
+    });
+    return new Promise((resolve, reject) => {
+        server.once('error', reject);
+        server.listen(config.EP_METRICS_SERVER_PORT, () => {
+            server.off('error', reject);
```
romayalon: what is this off() function? I don't see it in node.js documentation

liranmauda: `off()` is a Node.js `EventEmitter` alias for `removeListener()` (see [emitter.off(eventName, listener)](https://nodejs.org/api/events.html#emitteroffeventname-listener)).

We use it after a successful `listen` so the temporary `error` handler (`reject`) does not stay attached for the rest of the server lifetime.

Switched to the more familiar `removeListener('error', reject)` for clarity.

romayalon (approved):

PR #9829, opened 2026-07-16
NamespaceMultiStorageClass base

romayalon:
### Explain the Changes
* Add NamespaceMultiStorageClass: MD always in NB (STANDARD); data for DEEP_ARCHIVE/GLACIER routed to archive NamespaceS3. List/head/delete/tagging/ACL/lock pass through MD; put/get/MPU/restore are stubs.
* Wire it from object_sdk when the bucket has an archive_policy.
* Expose archive_policy (extended NSR info) on read_bucket_sdk_info / bucket_sdk_info.
* Add archive_upload_id on object MD/API so archive MPU UploadIds can be stored while the client keeps using the NB obj_id.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **New Features**
  - Added support for bucket archive policies with deep-archive resource and path details.
  - Added multi-storage-class bucket routing for standard, deep archive, and glacier storage.
  - Added archive upload identifiers to object uploads and object metadata.
  - Added support for retrieving archive policy details through bucket information APIs.
  - Added consistent deep-archive object key generation.

- **Documentation**
  - Updated deep-archive listing and restore behavior guidance, including restored-object status tracking.

### DIFF
--- a/src/sdk/namespace_multi_storage_class.js
+++ b/src/sdk/namespace_multi_storage_class.js
@@ -1,0 +1,411 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+
+const s3_utils = require('../endpoint/s3/s3_utils');
+const S3Error = require('../endpoint/s3/s3_errors').S3Error;
+
+
+/**
+ * NamespaceMultiStorageClass routes operations to different namespaces based on
+ * the object's storage class — analogous to NamespaceMerge, but keyed by
+ * storage_class instead of read/write resources.
+ *
+ * Metadata (MD) is always stored in the default (STANDARD) namespace and is
+ * the single source of truth for all storage classes. Non-default namespaces
+ * (e.g. NamespaceS3 for DEEP_ARCHIVE) store only object data.
+ *
+ * @implements {nb.Namespace}
+ */
+class NamespaceMultiStorageClass {
+
+    /**
+     * @param {{
+     *   namespace_by_storage_class: { [storage_class: string]: nb.Namespace },
+     *   default_storage_class?: string,
+     * }} args
+     */
+    constructor({ namespace_by_storage_class, default_storage_class }) {
+        if (!namespace_by_storage_class || !Object.keys(namespace_by_storage_class).length) {
+            throw new Error('NamespaceMultiStorageClass requires a non-empty namespaces map');
+        }
+        this.namespace_by_storage_class = namespace_by_storage_class;
+        this.default_storage_class = default_storage_class || s3_utils.STORAGE_CLASS_STANDARD;
+        this._metadata_ns = this.namespace_by_storage_class[this.default_storage_class];
+        if (!this._metadata_ns) {
+            throw new Error(`NamespaceMultiStorageClass: missing namespace for default storage class '${this.default_storage_class}'`);
+        }
+    }
+
+    /**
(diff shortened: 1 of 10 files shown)


src/util/deep_archive_utils.js
```diff
@@ -0,0 +1,20 @@
+/* Copyright (C) 2024 NooBaa */
+'use strict';
+
+/** Prefix for temporary restored copies of deep-archive objects (internal to NooBaa). */
+const RESTORED_OBJECTS_DIR = 'restored_objects/';
```
dannyzaken: How is `RESTORED_OBJECTS_DIR` used in restore? Is the data restored under a different key? why not use the original key?

src/sdk/deep_archive_utils.js
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Returns the key used to store object data in the deep-archive backend.
+ * Format: `{bucket._id}/{object_md.obj_id}`
+ *
+ * @param {nb.Bucket} bucket
+ * @param {nb.ObjectInfo} object_md
+ * @returns {string}
+ */
+function get_archive_key(bucket, object_md) {
+    return `${String(bucket._id)}/${object_md.obj_id}`;
```
karthikps97: The object will be stored with a prefix`noobaa_storage/`. So the key name should include this also.

src/util/deep_archive_utils.js
```diff
@@ -0,0 +1,20 @@
+/* Copyright (C) 2024 NooBaa */
```
karthikps97, line 1: Can we add utils under `src/util`? I have a similar requirement and have added a new utils module there.

src/util/deep_archive_utils.js
romayalon: I planned to store it under restored_objects/{original_object_id}
I can use the original key (restored_objects/key) but if the user restores several versions of the same key it might be a bit more complex

src/sdk/namespace_multi_storage_class.js
```diff
@@ -0,0 +1,448 @@
...
+     * Metadata for all storage classes is stored there.
+     * excludes that internal prefix such as `restored_objects/`.
+     * @param {object} params
+     * @param {nb.ObjectSDK} object_sdk
+     * @returns {Promise<object>}
+     */
+    async list_objects(params, object_sdk) {
+        this._assert_no_internal_prefix_access(params);
+        return this._metadata_ns.list_objects(params, object_sdk);
+    }
```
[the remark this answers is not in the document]
romayalon: I can't filter it after because it might cause list results with less than 1000 keys, we must filter them on the db query, added this change, please check

karthikps97 (approved):

src/util/deep_archive_utils.js
dannyzaken: restore can be done to the same objectmd, so it is version agnostic. 
In any case, there is a tradeoff - either the reads become more complex or the restore.

romayalon: yes we can, but actually this directory is for the data, not the metadata
so the path of the restored object will be in the standard backingstore on path s3://{bucket_name}/restored_objects/{original_object_id}

PR #9830, opened 2026-07-16
KeyCloack | Support aws:PrincipalTag in S3 request flow

naveenpaul1:
### Describe the Problem

aws:PrincipalTag was not suopported before

### Explain the Changes
1. Session tags are pulled from the Keycloak-issued JWT (get_tags_claim)
2. Threaded through sts_sdk.js → session token → signature_utils.js (auth_token.session_tags)
3. Evaluated in access_policy_utils.js against bucket/role policy Condition blocks
4. Matching schema addition in common_api.js for Condition on trust policies

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added support for `Condition` rules in IAM policy statements, shared with bucket policy condition handling.
  * Added extraction of session tags from OIDC/Keycloak identity tokens and included them in assumed-role responses.
  * Added support for evaluating `aws:PrincipalTag` conditions in S3 bucket policies.
* **Bug Fixes**
  * Fixed token verification to return decoded identity claims for use during authentication.
  * Preserved session tags when authenticating service requests.

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -407,30 +407,7 @@ module.exports = {
                                         type: 'string'
                                     },
                                     Condition: {
-                                        type: 'object',
-                                        properties: {
-                                            StringEquals: {
-                                                $ref: '#/definitions/bucket_policy_string_condition'
-                                            },
-                                            StringNotEquals: {
-                                                $ref: '#/definitions/bucket_policy_string_condition'
-                                            },
-                                            StringEqualsIgnoreCase: {
-                                                $ref: '#/definitions/bucket_policy_string_condition'
-                                            },
-                                            StringNotEqualsIgnoreCase: {
-                                                $ref: '#/definitions/bucket_policy_string_condition'
-                                            },
-                                            StringLike: {
-                                                $ref: '#/definitions/bucket_policy_string_condition'
-                                            },
-                                            StringNotLike: {
-                                                $ref: '#/definitions/bucket_policy_string_condition'
-                                            },
-                                            Null: {
-                                                $ref: '#/definitions/bucket_policy_null_condition'
-                                            }
-                                        }
+                                        $ref: '#/definitions/policy_condition'
                                     }
                                 }
                             },
@@ -491,7 +468,7 @@ module.exports = {
         },
 
         // based on bucket policy without Principal and NotPrincipal since are not used in inline policies
-        // removed the Condition as we don't support it yet
+        // This schema is used for IAM user and roles.
         iam_user_policy_document: {
             type: 'object',
             required: ['Statement'],
@@ -522,6 +499,9 @@ module.exports = {
                                     Effect: {
                                         enum: ['Allow', 'Deny'],
                                         type: 'string'
+                                    },
+                                    Condition: {
+                                        $ref: '#/definitions/policy_condition'
                                     },
                                 }
                             },
@@ -560,6 +540,33 @@ module.exports = {
                         ]
                     }
                 },
+            }
+        },
+        policy_condition: {
+            type: 'object',
+            additionalProperties: false,
+            properties: {
+                StringEquals: {
+                    $ref: '#/definitions/bucket_policy_string_condition'
+                },
+                StringNotEquals: {
+                    $ref: '#/definitions/bucket_policy_string_condition'
+                },
+                StringEqualsIgnoreCase: {
+                    $ref: '#/definitions/bucket_policy_string_condition'
+                },
+                StringNotEqualsIgnoreCase: {
+                    $ref: '#/definitions/bucket_policy_string_condition'
+                },
+                StringLike: {
+                    $ref: '#/definitions/bucket_policy_string_condition'
+                },
+                StringNotLike: {
+                    $ref: '#/definitions/bucket_policy_string_condition'
+                },
+                Null: {
+                    $ref: '#/definitions/bucket_policy_null_condition'
+                }
             }
         },
 
--- a/src/sdk/sts_sdk.js
+++ b/src/sdk/sts_sdk.js
@@ -11,6 +11,7 @@ const jwt = require('jsonwebtoken');
 const { resolve_iam_role_by_arn } = require('../endpoint/iam/iam_utils');
 const ldap_client = require('../util/ldap_client');
 const keycloak_client = require('../util/keycloak_client');
+const { get_tags_claim } = require('../util/access_policy_utils');
 
 class StsSDK {
 
@@ -170,13 +171,16 @@ class StsSDK {
                 await keycloak_instance.initialize();
             }
             // JWT token decoded and check token issuer is in provider list
-            await keycloak_instance.verify_token(req.body.web_identity_token);
+            const decoded_token = await keycloak_instance.verify_token(req.body.web_identity_token);
 
             // Introspect token with Keycloak using client_id, client_secret, and access_token
             // This is the key implementation for Keycloak - validates token is active and not revoked
             const introspection_resp = await keycloak_instance.introspect_token(
                 req.body.web_identity_token
             );
+
+            // Extract session tags from decoded token.
+            const session_tags = get_tags_claim(decoded_token);
 
             // Assume role
             const role_config = await this._assume_role(req.body.role_arn);
@@ -189,6 +193,7 @@ class StsSDK {
                 sub: introspection_resp.sub,
                 aud: introspection_resp.client_id || introspection_resp.aud,
                 iss: introspection_resp.iss,
+                session_tags,
                 // Store additional claims for audit
                 email: introspection_resp.email,
                 name: introspection_resp.name,
--- a/src/util/access_policy_utils.js
+++ b/src/util/access_policy_utils.js
@@ -116,7 +116,8 @@ const predicate_map = {
 const condition_fit_functions = {
     's3:ExistingObjectTag': _is_object_tag_fit,
     's3:x-amz-server-side-encryption': _is_server_side_encryption_fit,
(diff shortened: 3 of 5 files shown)


src/api/common_api.js
```diff
@@ -523,6 +523,9 @@ module.exports = {
                                         enum: ['Allow', 'Deny'],
                                         type: 'string'
                                     },
+                                    Condition: {
+                                        $ref: '#/definitions/trust_policy_condition'
+                                    },
```
shirady, line 505: It is for the IAM user policy. Why was it added here?

If it changes in this version, then the docs should be updated as well; currently:

> (2) We do not support the Condition field (as we decided at this point)


([link](https://github.com/noobaa/noobaa-core/blob/master/docs/design/IamUserInlinePolicy.md)).

src/sdk/sts_sdk.js
```diff
@@ -170,14 +171,17 @@ class StsSDK {
...
-            await keycloak_instance.verify_token(req.body.web_identity_token);
+            const decoded_token = await keycloak_instance.verify_token(req.body.web_identity_token);
 
             // Introspect token with Keycloak using client_id, client_secret, and access_token
             // This is the key implementation for Keycloak - validates token is active and not revoked
             const introspection_resp = await keycloak_instance.introspect_token(
                 req.body.web_identity_token
             );
 
+            // Extract session tags from verified token (not from introspection)
```
shirady: Can you explain this comment?

src/util/access_policy_utils.js
```diff
@@ -129,9 +130,14 @@ const keycloak_predicate_map = {
 const supported_actions = {
     's3:ExistingObjectTag': ['s3:DeleteObjectTagging', 's3:DeleteObjectVersionTagging', 's3:GetObject', 's3:GetObjectAcl', 's3:GetObjectTagging', 's3:GetObjectVersion', 's3:GetObjectVersionTagging', 's3:PutObjectAcl', 's3:PutObjectTagging', 's3:PutObjectVersionTagging'],
     's3:x-amz-server-side-encryption': ['s3:PutObject'],
-    's3:VersionId': ['s3:GetObjectVersion', 's3:DeleteObjectVersion', 's3:GetObjectVersionAttributes', 's3:GetObjectVersionTagging', 's3:PutObjectVersionTagging', 's3:DeleteObjectVersionTagging']
+    's3:VersionId': ['s3:GetObjectVersion', 's3:DeleteObjectVersion', 's3:GetObjectVersionAttributes', 's3:GetObjectVersionTagging', 's3:PutObjectVersionTagging', 's3:DeleteObjectVersionTagging'],
```
shirady: I think it is a redundant change, right?

src/util/access_policy_utils.js
```diff
@@ -129,9 +130,14 @@ const keycloak_predicate_map = {
...
 };
 
+// Condition keys that are principal/session-scoped rather than action-scoped —
+// they apply regardless of which S3 method is being called.
+const SKIPPED_CONDITIONS_ACTION = new Set(['aws:PrincipalTag']);
+
+
 const SUPPORTED_BUCKET_POLICY_CONDITIONS = Object.keys(supported_actions);
```
shirady, line 141: Should the session tags be supported in bucket policy?

src/util/access_policy_utils.js
```diff
@@ -157,6 +163,19 @@ async function _is_object_version_fit(req, predicate, value) {
...
 }
 
+async function _is_aws_principal_tag_fit(req, predicate, value) {
+    const auth_token = req.object_sdk.get_auth_token();
+    const session_tags = auth_token.session_tags;
+    // If S3 request is not with temp session tags, should not check the aws:PrincipalTag
+    if (!session_tags) {
+        return true;
+    }
+    const tag_value = session_tags?.[value.key] ?? null;
```
shirady: Why default this to `null` and not to stay with `undefined`?

src/api/common_api.js
naveenpaul1: This was added for IAM Role not for IAM users

src/sdk/sts_sdk.js
naveenpaul1: I think the comment itself is not needed, here we are getting the custom tagges from decoded token thats all.

src/util/access_policy_utils.js
naveenpaul1: May be but we need more discussion on it, and not in this PR

src/api/common_api.js
shirady: In lines 493-495I see this:

>        // based on bucket policy without Principal and NotPrincipal since are not used in inline policies
>         // removed the Condition as we don't support it yet
>         iam_user_policy_document:

src/sdk/sts_sdk.js
shirady: What made this token "verified"?
What did you mean by "not from introspection"?

src/util/access_policy_utils.js
shirady: But in this PR, it is added in `SUPPORTED_BUCKET_POLICY_CONDITIONS`, so it's better not to add it there if we don't want this in the bucket policy.

src/sdk/sts_sdk.js
naveenpaul1: will update the comment

src/api/common_api.js
naveenpaul1: But for IAM role we need condition to support `aws:PrincipalTag`, or else we need to create a different schema just for IAM role policy,

src/util/access_policy_utils.js
naveenpaul1: @shirady But we are checking only if there is session_tags in request, that is present only in STS temp session S3 request  [link](https://github.com/noobaa/noobaa-core/pull/9830/changes#diff-973695a10fce9a35745f2b2766f193b83d67ed78132056a0ca0798c84d9d8558R170)

src/api/common_api.js
shirady: Now I see that:
https://github.com/noobaa/noobaa-core/blob/b3e65d46cef10de2f8fff59a518347d495ea0186/src/server/system_services/schemas/iam_role_schema.js#L39-L46

@aayushchouhan09 do you plan to rename it?
It is confusing that it is with "user" in the name.

@naveenpaul1 I think you can either block it in the code or update the documentation (at least remove the mentioned line),

aayushchouhan09: @shirady 
It is pending on our side, we were fixing it in the very first PR for schema changes but removed it due to some complications. Both role and user policy are similar in many ways so we were trying to unify them but removed those changes later.

@naveenpaul1 I think we can fix it in your PR by creating a separate `iam_role_policy` for roles schema and also the unification part for the policies. wdyt?

or if you want me to create a separate PR for those changes, please let me know whatever be the case. Thanks!

src/api/common_api.js
```diff
@@ -501,28 +478,63 @@ module.exports = {
                     type: 'array',
                     items: {
                         allOf: [{
```
shirady, line 480: Why did you change in `iam_user_policy_document`?

src/api/common_api.js
```diff
@@ -563,6 +575,60 @@ module.exports = {
             }
         },
 
+        iam_policy_statement: {
```
shirady: Why did you add `iam_user_policy_document`?

src/api/common_api.js
```diff
@@ -573,6 +639,16 @@ module.exports = {
                 }
             }
         },
+        iam_role_policy: {
```
shirady: @aayushchouhan09 if you plan to save the roles in the same place as users, I think you would need only one `iam_policy` property (and not both `iam_user_policy` and `iam_role_policy`).
Do plan to do this change soon?

src/api/common_api.js
shirady: @aayushchouhan09 don't you prefer to have a single `iam_policy` for both users and roles?
Unless there is a difference that you think we can have them both.
@naveenpaul1 @aayushchouhan09 I guess it is on your side to decide in which PR to have the change. I think the fastest way would be in a separate one since the schema of the policy is a bit complicated.

src/api/common_api.js
```diff
@@ -491,7 +468,7 @@ module.exports = {
         },
 
         // based on bucket policy without Principal and NotPrincipal since are not used in inline policies
-        // removed the Condition as we don't support it yet
```
shirady, line 494: I think we still don't support the `Condition` in user policy, and worth blocking it in the code somewhere.

src/api/common_api.js
aayushchouhan09: Make sense, we can have a separate PR and reviews for schema changes.

shirady (approved):
LGTM

Since in the IAM user policy we didn't have a `Condition` and with the schema changes in this PR it would be allowed, it would be worth opening a gap so we can block it in the code.

src/api/common_api.js
aayushchouhan09: Instead of blocking it temporary.

I think we can have the changes in the separate PR for common `iam_policy` (having required fields) which is used by all bucket, user and role policies and their specific fields can be added respectively to each of them. Which we trying to achieve in the same PR and now we have a separate PR for it.

@naveenpaul1 @shirady Do you want me to create a PR with those changes?

naveenpaul1: @aayushchouhan09 Shira was talking about handle it in different PR.
I have created an issue for that https://github.com/noobaa/noobaa-core/issues/9860 
You can pick it

PR #9839, opened 2026-07-20
CI | Reduce Jest unit tests run time

liranmauda:
### Describe the Problem
`run-jest-unit-tests` on master PRs averages ~18–20 minutes, largely gated by `test_file_writer.test.js` (~10+ minutes). That suite reused CLI stress defaults (`1000` parts × `20MiB`) plus an artificial `P.delay(100)` on every mock `writev`, which inflated wall clock without adding assertions.

### Explain the Changes
1. Shrink FileWriter Jest workloads: default cases use `parts=50` / `part_size=2MiB`; keep `iov_max=1/2` at `parts=20`; buffer-flush case still forces `num_chunks > 1024` with smaller parts/size.
2. Make `part_size` a parameter of `hash_target` / `file_target` (CLI defaults unchanged for manual stress runs).
3. Gate the artificial `TargetHash.writev` delay behind `--writev_delay` (default 0 so Jest does not pay it).
4. Lower suite `RUN_TIMEOUT` from 10m to 3m.

1. N/A — CI performance improvement.

### Testing Instructions:
1. Confirm Jest Unit Tests on this branch lands near ~8–10m (validated ~9.9m on fork: https://github.com/liranmauda/noobaa-core/actions/runs/29744468325).
2. Same 7 FileWriter cases still pass (hash/file targets, iov_max=1/2, num_chunks>1024 flush path).

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added configurable write pacing for file-writing via a new load-shaping delay option.
  * Added support for customizing per-part sizes in the hashing and file generation utilities.
* **Tests**
  * Improved file-writer test reliability and speed by using explicit concurrency parameters and reducing per-test timeouts.
  * Refined large-part scenarios to be more deterministic in CI.

### DIFF
--- a/src/tools/file_writer_hashing.js
+++ b/src/tools/file_writer_hashing.js
@@ -20,6 +20,8 @@ const CHUNK = Number(argv.chunk) || 16 * 1024;
 const PART_SIZE = Number(argv.part_size) || 20 * 1024 * 1024;
 const F_PREFIX = argv.dst_folder || '/tmp/file_writer_hashing/';
 const IOV_MAX = argv.iov_max || config.NSFS_DEFAULT_IOV_MAX;
+// Artificial writev delay for CLI load-shaping only. Jest must not pay this cost.
+const WRITEV_DELAY_MS = Number(argv.writev_delay) || 0;
 
 const DEFAULT_FS_CONFIG = {
     uid: Number(argv.uid) || process.getuid(),
@@ -40,7 +42,7 @@ class TargetHash {
         return this.hash.digest('hex');
     }
     async writev(_config, buffers) {
-        await P.delay(100);
+        if (WRITEV_DELAY_MS) await P.delay(WRITEV_DELAY_MS);
         for (const buf of buffers) this.hash.update(buf);
     }
 }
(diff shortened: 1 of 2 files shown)


src/tools/file_writer_hashing.js
```diff
@@ -40,7 +42,7 @@ class TargetHash {
         return this.hash.digest('hex');
     }
     async writev(_config, buffers) {
-        await P.delay(100);
```
romayalon, line 43: per my understanding, it was intentionally added...

romayalon (approved):

src/tools/file_writer_hashing.js
liranmauda: I think the timeout is just a mock.
It is mocking the hash time, but the Target hash is actually doing it in memory and not flashing to the disk.

The inattention was on the manual run of it. We can add this timeout via arg:
```
// Artificial writev delay for CLI load-shaping only. Jest must not pay this cost.
const WRITEV_DELAY_MS = Number(argv.writev_delay) || 0;
```

PR #9833, opened 2026-07-19
[Deep Archive] - Add archive API under bg worker process

karthikps97:
### Explain the Changes
1. Add new RPC API to support transitioning objects to deep_archive and register it under bg services.
2. Added few constants related to S3 deep archive.

### Issues: Fixed #xxx / Gap #xxx
1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added archive API support to the SDK/RPC interface, including an admin-only operation to archive an object.
  * Introduced new archive-related storage classes and transition status values (e.g., Deep Archive/Glacier; in-progress/done).

* **Limitations**
  * The new `archive_object` operation is currently registered but not yet functional; requests return a “not yet implemented” error.

PR #9840, opened 2026-07-21
[Deep Archive] - Transition lifecycle rule implementation in lifecycle bg worker

karthikps97:
### Description
Implement S3-compatible lifecycle `Transition` and `NoncurrentVersionTransition` rules in the lifecycle background worker, enabling objects to be moved from standard storage to an archive backend (e.g., DEEP_ARCHIVE).

- Each rule is processed in batches with a configurable size.
- The eligible objects from each batch are processed async by `archive_worker`, which would be reading from the standard storage class and writing the object to archive storage class configured in the `deep_archive_resource` of the namespace store.
- Objects with errors are retried on the next cycle of lifecycle.

### Explain the Changes
1. Added md store functions to find objects to transition.
2. lifecycle bg worker processes `Transition` and `NoncurrentVersionTransition` rules.

1. Additional S3 filters are not yet handled.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary

* **New Features**
  * Lifecycle transition rules can now archive eligible objects to configured archival storage classes, including noncurrent/versioned transitions.
  * Added admin endpoints to discover objects pending transition and to update an individual object’s transition status.
  * Added optional deep-archive resource support and batched/paginated transition processing.

* **Bug Fixes**
  * Objects that fail during transition have their transition state cleared for safe retry in later runs.

### DIFF
--- a/config.js
+++ b/config.js
@@ -520,6 +520,7 @@ config.LIFECYCLE_INTERVAL = 8 * 60 * 60 * 1000; // 8h
 config.LIFECYCLE_BATCH_SIZE = 1000;
 config.LIFECYCLE_SCHEDULE_MIN = 5 * 1000 * 60; // run every 5 minutes
 config.LIFECYCLE_ENABLED = true;
+config.LIFECYCLE_TRANSITION_BATCH_SIZE = 50;
 
 /////////////////////
 // ARCHIVE CHECK   //
--- a/src/api/object_api.js
+++ b/src/api/object_api.js
@@ -1540,6 +1540,93 @@ module.exports = {
             auth: { system: ['admin', 'user'] }
         },
 
+        find_objects_to_transition: {
+            method: 'GET',
+            params: {
+                type: 'object',
+                required: ['bucket', 'transition_ts'],
+                properties: {
+                    bucket: { $ref: 'common_api#/definitions/bucket_name' },
+                    transition_ts: { type: 'number' },
+                    batch_size: { type: 'integer' },
+                    key_marker: { type: 'string' },
+                },
+            },
+            reply: {
+                type: 'object',
+                properties: {
+                    objects: {
+                        type: 'array',
+                        items: {
+                            $ref: '#/definitions/object_info'
+                        },
+                    },
+                    is_truncated: { type: 'boolean' },
+                    next_marker: { type: 'string' },
+                },
+            },
+            auth: {
+                system: 'admin',
+            },
+        },
+
+        find_versioned_objects_to_transition: {
+            method: 'GET',
+            params: {
+                type: 'object',
+                required: ['bucket', 'transition_ts'],
+                properties: {
+                    bucket: { $ref: 'common_api#/definitions/bucket_name' },
+                    is_latest: { type: 'boolean' },
+                    transition_ts: { type: 'number' },
+                    batch_size: { type: 'integer' },
+                    key_marker: { type: 'string' },
+                    version_seq_marker: { type: 'integer' },
+                    noncurrent_days: { type: 'integer' },
+                    newer_noncurrent_versions: { type: 'integer' },
+                },
+            },
+            reply: {
+                type: 'object',
+                properties: {
+                    objects: {
+                        type: 'array',
+                        items: {
+                            $ref: '#/definitions/object_info'
+                        },
+                    },
+                    is_truncated: { type: 'boolean' },
+                    next_marker: { type: 'string' },
+                    next_version_seq_marker: { type: 'integer' },
+                },
+            },
+            auth: {
+                system: 'admin',
+            },
+        },
+
+        update_transition_status: {
+            method: 'PUT',
+            params: {
+                type: 'object',
+                required: ['obj_id'],
+                properties: {
+                    obj_id: { objectid: true },
+                    update_transition_status: { $ref: 'common_api#/definitions/transition_status_enum' },
+                    deleted: { idate: true },
+                    transition_status: { $ref: 'common_api#/definitions/transition_status_enum' },
+                    unset_transition_status: { type: 'boolean' },
+                    storage_class: { $ref: 'common_api#/definitions/storage_class_enum' },
+                },
+            },
+            reply: {
+                type: 'boolean'
+            },
+            auth: {
+                system: 'admin',
+            },
+        },
+
         get_object_restore_info: {
             method: 'GET',
             params: {
--- a/src/server/bg_services/lifecycle.js
+++ b/src/server/bg_services/lifecycle.js
@@ -12,6 +12,12 @@ const auth_server = require('../common_services/auth_server');
 const config = require('../../../config');
 const { get_notification_logger, check_notif_relevant,
     OP_TO_EVENT, compose_notification_lifecycle } = require('../../util/notifications_util');
+const COMMON_CONSTANTS = require('../../common/constants');
+
+/*************************/
+/******* CONSTANTS *******/
+/*************************/
+const ARCHIVE = COMMON_CONSTANTS.ARCHIVE;
 
 function get_expiration_timestamp(expiration) {
     if (!expiration) {
@@ -20,6 +26,124 @@ function get_expiration_timestamp(expiration) {
         return Math.floor(new Date(expiration.date).getTime() / 1000);
     } else if (expiration.days) {
         return moment().subtract(expiration.days, 'days').unix();
+    }
+}
+
+/**
+ * Gets the Unix timestamp associated with a transition.
+ * Transition for objects can be set to 0 unlike expiration which should be a positive integer.
+ *
+ * @param {Object} transition - The transition configuration.
+ * @param {string|Date} [transition.date] - An absolute date to convert to a
+ * Unix timestamp.
+ * @param {number} [transition.days] - The number of days to subtract from
+ * the current time.
+ * @returns {number|undefined} The Unix timestamp in seconds, or `undefined`
+ * if no valid transition value is provided.
+ */
+function get_transition_timestamp(transition) {
+    if (!transition) {
+        return undefined; // undefined
+    } else if (transition.date !== undefined) {
+        return Math.floor(new Date(transition.date).getTime() / 1000);
+    } else if (transition.days !== undefined) {
+        return moment().subtract(transition.days, 'days').unix();
+    }
+}
+
+/**
+ * Transition eligible objects from standard storage to the archive backend.
+ * Transition is done based on the set transition lifecycle rule.
+ *
+ * @param {Object} system - the NooBaa system object from system_store
+ * @param {nb.Bucket} bucket_info - the bucket object (includes name, _id, versioning, archive_policy)
+ * @param {Object} rule - the lifecycle rule containing transition or noncurrent_version_transition
+ */
+async function process_transition(system, bucket_info, rule) {
+    if (!rule) {
+        return;
+    }
+
+    try {
+        const batch_size = config.LIFECYCLE_BATCH_SIZE;
+        const object_server = server_rpc.client.object;
+        const target_storage_class = rule.storage_class;
+
+        const transition = {
+            days: rule.days ?? rule.noncurrent_days,
+            date: rule.date
+        };
+        const transition_ts = get_transition_timestamp(transition);
+        if (!transition_ts) {
+            dbg.error("found transition rule with invalid transition day/date", bucket_info.name, rule);
+            return;
+        } else if (Object.keys(bucket_info.archive_policy?.deep_archive_resource || {}).length <= 0) {
+            dbg.error("found bucket with invalid archive resource", bucket_info.name, bucket_info.archive_policy);
+            return;
+        } else if (!Object.keys(ARCHIVE.STORAGE_CLASS).includes(target_storage_class)) {
+            dbg.error(`target storage class should be one of: ${Object.keys(ARCHIVE.STORAGE_CLASS)}`);
+            return;
+        }
+
+        // Behavior for versioning suspended bucket is same as an enabled bucket
+        const versioning_disabled = bucket_info.versioning === COMMON_CONSTANTS.S3.VERSIONING.DISABLED;
+        let is_truncated = true;
+        let key_marker = '';
+        let version_seq_marker;
+        // List objects that are not deleted, reclaimed and not having transition_status as in_progress
+        while (is_truncated) {
+            let result;
+            if (versioning_disabled) {
+                result = await object_server.find_objects_to_transition({
+                    bucket: bucket_info.name,
+                    batch_size,
+                    key_marker,
+                    transition_ts,
+                }, {
+                    auth_token: auth_server.make_auth_token({
+                        system_id: system._id,
+                        account_id: system.owner._id,
+                        role: 'admin'
+                    })
+                });
+            } else {
+                /* 
+                For versioned buckets, the objects depends on the following rules: 
+                    1. Transition - All the latest versions of the objects
+                    2. NoncurrentVersionTransition - AND operation between NoncurrentDays and NewerNoncurrentVersions
+                */
+                const is_latest = !(rule.noncurrent_days || rule.newer_noncurrent_versions);
+                result = await object_server.find_versioned_objects_to_transition({
+                    bucket: bucket_info.name,
+                    batch_size,
+                    key_marker,
+                    version_seq_marker,
+                    transition_ts,
+                    is_latest,
+                    noncurrent_days: rule.noncurrent_days,
+                    newer_noncurrent_versions: rule.newer_noncurrent_versions,
+                }, {
+                    auth_token: auth_server.make_auth_token({
+                        system_id: system._id,
+                        account_id: system.owner._id,
+                        role: 'admin'
+                    })
+                });
+            }
+
+            try {
+                const obj_ids = await transition_objects(system, bucket_info, result.objects, target_storage_class);
+                dbg.log1("successfully transitioned batch with object id's", obj_ids);
+            } catch (err) {
+                dbg.error("error occurred while transitioning objects batch", err);
+            }
+
+            key_marker = result.next_marker;
+            version_seq_marker = result.next_version_seq_marker;
+            is_truncated = result.is_truncated;
+        }
+    } catch (e) {
+        dbg.error("error occurred while executing batch transition", e);
     }
 }
 
@@ -126,7 +250,8 @@ async function handle_bucket_rule(system, rule, j, bucket) {
     if (
         rule.expiration === undefined &&
         rule.abort_incomplete_multipart_upload === undefined &&
-        rule.noncurrent_version_expiration === undefined
+        rule.noncurrent_version_expiration === undefined &&
+        !rule.transition && !rule.noncurrent_version_transition
     ) {
         dbg.log0('LIFECYCLE SKIP bucket:', bucket.name, '(bucket id:', bucket._id, ') rule', util.inspect(rule), 'now', now, 'last_sync', rule.last_sync, 'rule contains no expiration parameters');
         return;
@@ -214,6 +339,19 @@ async function handle_bucket_rule(system, rule, j, bucket) {
         );
     }
 
+    if (rule.transition || rule.noncurrent_version_transition) {
+        // NoncurrentVersionTransition has no effect on versioning disabled buckets
+        if (rule.noncurrent_version_transition &&
+            bucket.versioning === COMMON_CONSTANTS.S3.VERSIONING.DISABLED) {
+            dbg.log1("skipping noncurrent_version_transition rule as bucket versioning is disabled", bucket.name);
+        } else {
+            await process_transition(system, bucket, rule.transition);
+            await process_transition(system, bucket, rule.noncurrent_version_transition);
+
+            dbg.log1("bucket", bucket.name, "transition batch completed");
+        }
+    }
+
     if (num_objects_deleted >= config.LIFECYCLE_BATCH_SIZE) should_rerun = true;
 
     bucket.lifecycle_configuration_rules[j].last_sync = Date.now();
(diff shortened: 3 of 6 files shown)


karthikps97:
Depends on #9833

src/api/object_api.js
```diff
@@ -1540,6 +1540,89 @@ module.exports = {
...
+                system: 'admin',
+            },
+        },
+        transition_object: {
+            method: 'PUT',
+            params: {
+                type: 'object',
+                required: ['_id'],
+                properties: {
+                    _id: { objectid: true },
```
romayalon: you probably want to use obj_id

src/server/bg_services/lifecycle.js
```diff
@@ -23,6 +24,135 @@ function get_expiration_timestamp(expiration) {
     }
 }
 
+/**
+ * Transition eligible objects from standard storage to the archive backend.
+ * Transition is done based on the set transition lifecycle rule.
+ *
+ * @param {Object} system - the NooBaa system object from system_store
+ * @param {Object} bucket - the bucket object (includes name, _id, versioning, archive_policy)
```
romayalon: bucket should probably be renamed to bucket_info and use nb.Bucket type if this is what you are referring to

src/server/object_services/md_store.js
```diff
@@ -2402,6 +2476,166 @@ class MDStore {
...
+        const query_limit = params.batch_size || 100;
+        const bucket_id = String(params.bucket._id);
+        const create_time_cutoff = moment.unix(params.transition_ts).toISOString();
+        const values = [bucket_id, create_time_cutoff];
+
+        let key_marker_condition = '';
+        let idx = 3;
+        if (params.key_marker) {
+            values.push(params.key_marker);
+            key_marker_condition = `AND data->>'key' > $${idx += 1}`;
```
romayalon: why adding 1 to 3? shouldn't it be $3?

src/server/object_services/md_store.js
```diff
@@ -2402,6 +2476,166 @@ class MDStore {
...
+            ) < $2::timestamptz
+        AND (data->'deleted' IS NULL OR data->'deleted' = 'null'::jsonb)
+        AND (data->'reclaimed' IS NULL OR data->'reclaimed' = 'null'::jsonb)
+        AND (data->'upload_started' IS NULL OR data->'upload_started' = 'null'::jsonb)
+        AND (data->'version_past' IS NULL OR data->'version_past' = 'null'::jsonb)
+        AND (data->'delete_marker' IS NULL OR data->'delete_marker' = 'null'::jsonb)
+        AND (data->'transition_status' IS NULL OR data->'transition_status' = 'null'::jsonb)
+        ${key_marker_condition}
+        ORDER BY data->>'key' ASC
+        LIMIT $${idx};`;
```
romayalon, line 2485: why the limit is idx and not the query limit?

src/server/object_services/object_server.js
```diff
@@ -1077,7 +1077,98 @@ async function update_object_md(req) {
...
+            version_seq_marker,
+            noncurrent_days,
+            newer_noncurrent_versions,
+        });
+    }
 
+    const last_obj = objects.length ? objects[objects.length - 1] : undefined;
+    return {
+        objects: objects.map(get_object_info),
+        is_truncated: objects.length >= batch_size,
```
romayalon, line 1163: batch_size is the limit of the query no? how would object.length will be larger than batch size?

src/server/bg_services/lifecycle.js
```diff
@@ -23,6 +24,135 @@ function get_expiration_timestamp(expiration) {
...
+ * @param {Object} rule - the lifecycle rule containing transition or noncurrent_version_transition
+ * @param {number} transition_ts - unix timestamp cutoff; objects created before this are eligible
+ */
+async function transition_objects(system, bucket, rule, transition_ts) {
```
romayalon: this function is very lengthy, I would try to split to helper functions for readability

src/server/bg_services/lifecycle.js
```diff
@@ -23,6 +24,135 @@ function get_expiration_timestamp(expiration) {
...
+                        account_id: system.owner._id,
+                        role: 'admin'
+                    })
+                });
+            }
+
+            const obj_ids = await P.map_with_concurrency(batch_size, result.objects, (async obj => {
+                const obj_id = obj.obj_id;
+                try {
+                    const res = await server_rpc.client.object.transition_object({
```
romayalon: I would rename transition_object() since it's only sets the transition status on the object md and not really transitioning the data

src/server/bg_services/lifecycle.js
```diff
@@ -23,6 +24,135 @@ function get_expiration_timestamp(expiration) {
...
+                }, {
+                    auth_token: auth_server.make_auth_token({
+                        system_id: system._id,
+                        account_id: system.owner._id,
+                        role: 'admin'
+                    })
+                });
+            }
+
+            const obj_ids = await P.map_with_concurrency(batch_size, result.objects, (async obj => {
```
romayalon: concurrenc=batch size might cause lots of stress no?

src/server/bg_services/lifecycle.js
```diff
@@ -23,6 +24,135 @@ function get_expiration_timestamp(expiration) {
...
+                        _id: obj_id,
+                    }, {
+                        auth_token: auth_server.make_auth_token({
+                            system_id: system._id,
+                            account_id: system.owner._id,
+                            role: 'admin'
+                        })
+                    });
+
+                    dbg.log1("object:", obj_id, "transition status updated:", res);
```
romayalon: this is very important log, please use our usual logs pattern so it'll be more understood

src/server/bg_services/lifecycle.js
```diff
@@ -23,6 +24,135 @@ function get_expiration_timestamp(expiration) {
...
+
+                    const rpc_client = server_rpc.rpc.new_client({
```
romayalon: Try passing rpc_client when initiating the lifecycle bg_worker - on`bg_workers.js` instead
we pass `client: server_rpc.client,` I don't see a reason why it won't work on this case as well

src/api/object_api.js
```diff
@@ -1540,6 +1540,89 @@ module.exports = {
...
         },
+        find_objects_to_transition: {
+            method: 'GET',
+            params: {
+                type: 'object',
+                required: ['bucket', 'transition_ts'],
+                properties: {
+                    bucket: { $ref: 'common_api#/definitions/bucket_name' },
+                    transition_ts: { type: 'number' },
+                    batch_size: { type: 'integer' },
```
romayalon, line 1551: why do we need to pass batch_size between servers? asking because it's a global config

src/server/object_services/object_server.js
```diff
@@ -1077,7 +1077,98 @@ async function update_object_md(req) {
...
+        _id: rpc_params._id,
+        deleted: rpc_params.deleted || null,
+        transition_status: rpc_params.transition_status || null,
+    };
+
+    try {
+        await MDStore.instance().find_and_update_object(filter, set_update, unset_updates);
+    } catch (e) {
+        dbg.warn("object to transition does not match criteria", req.rpc_params);
+        return false;
```
romayalon: what does "object to transition does not match criteria", IMO we should throw the error and catch it outside

src/api/object_api.js
```diff
@@ -1540,6 +1540,89 @@ module.exports = {
...
+                    _id: { objectid: true },
+                    update_transition_status: { $ref: 'common_api#/definitions/transition_status_enum' },
+                    deleted: { idate: true },
+                    transition_status: { $ref: 'common_api#/definitions/transition_status_enum' },
+                    unset_transition_status: { type: 'boolean' },
+                    storage_class: { $ref: 'common_api#/definitions/storage_class_enum' },
+                },
+            },
+            reply: {
+                type: 'boolean'
```
romayalon, line 1623: boolean replies usually are not very informative

src/server/object_services/object_server.js
```diff
@@ -1077,7 +1077,98 @@ async function update_object_md(req) {
     await MDStore.instance().update_object_by_id(obj._id, set_updates);
 }
 
+async function find_objects_to_transition(req) {
```
romayalon, line 1096: please add jsdocs for all new functions

src/server/object_services/object_server.js
```diff
@@ -1077,7 +1077,98 @@ async function update_object_md(req) {
...
+    const objects = await MDStore.instance().find_objects_to_transition({
+        bucket: req.bucket,
+        batch_size,
+        key_marker,
+        transition_ts,
+    });
+
+    return {
+        objects: objects.map(get_object_info),
+        is_truncated: objects.length >= batch_size,
```
romayalon, line 1111: same as we the other is_trauncated comment - batch_size is the limit of the query no? how would object.length will be larger than batch size?

src/server/bg_services/lifecycle.js
```diff
@@ -23,6 +24,135 @@ function get_expiration_timestamp(expiration) {
...
+                        }, {
+                            auth_token: auth_server.make_auth_token({
+                                system_id: system._id,
+                                account_id: system.owner._id,
+                                role: 'admin'
+                            })
+                        });
+                    } else {
+                        dbg.warn("failed to archive object", obj_id, "retry in next batch");
+                        await unset_transition_status(server_rpc.client, system, obj_id);
```
romayalon: I would consider to keep the error instead of unsetting the tansition status

src/server/object_services/object_server.js
```diff
@@ -1077,7 +1077,98 @@ async function update_object_md(req) {
...
+    throw_if_maintenance(req);
+    load_bucket(req, { include_deleting: true });
+
+    const { key_marker, version_seq_marker, transition_ts, is_latest,
+        newer_noncurrent_versions, noncurrent_days } = req.rpc_params;
+    const batch_size = req.rpc_params.batch_size || 1000;
+    let objects = [];
+
+    if (is_latest) {
+        objects = await MDStore.instance().find_objects_latest({
```
romayalon: maybe I misunderstand something but find_objects_to_transition() and find_objects_latest() seem to do the same thing, is this a leftover or something?

src/server/bg_services/lifecycle.js
```diff
@@ -23,6 +24,135 @@ function get_expiration_timestamp(expiration) {
...
+                        role: 'admin'
+                    })
+                });
+            } else {
+                /* 
+                For versioned buckets, the objects depends on the following rules: 
+                    1. Transition - All the latest versions of the objects
+                    2. NoncurrentVersionTransition - AND operation between NoncurrentDays and NewerNoncurrentVersions
+                */
+                const is_latest = Boolean(rule.transition);
```
romayalon: per my understanding this line assumes that Transition and NoncurrentVersionTransition shouldn't be set on the same rule but they can IIRC

src/server/bg_services/lifecycle.js
```diff
@@ -214,6 +345,32 @@ async function handle_bucket_rule(system, rule, j, bucket) {
...
+        const current_rule = rule.transition || rule.noncurrent_version_transition;
+        const transition = {
+            days: current_rule.days || current_rule.noncurrent_days,
+            date: current_rule.date
+        };
+        const transition_ts = get_expiration_timestamp(transition);
+        if (!transition_ts) {
+            dbg.error("found transition rule with invalid transition day/date", bucket.name, rule);
+            return;
+        } else if (Object.keys(bucket.archive_policy?.deep_archive_resource || {}).length <= 0) {
```
romayalon: it depends on the storage class of the rule, we should make sure it's set to deep archive/GLACIER && this condition

romayalon (changes requested):

src/server/object_services/md_store.js
karthikps97: It probably got changed from increment operator when i ran a lint fix.

src/server/object_services/md_store.js
karthikps97: Its also passed as a parameter to query. Since `key_marker_condition` can be omitted, we need this.

src/server/object_services/object_server.js
karthikps97: Yes, at max it can be equal. But kept it as a defensive check.

src/server/bg_services/lifecycle.js
karthikps97: Yes, it could. I probably missed to reduce this number

src/api/object_api.js
karthikps97: IMO, we should give ability for RPC caller to control batch size as it gives more flexibility.

src/server/bg_services/lifecycle.js
karthikps97: We already have an import for rpc in this module. So reused this since we have existing code which uses this. WDYT?

src/server/bg_services/lifecycle.js
karthikps97: We need to unset the transition status so that the object would be picked in the next batch.

src/server/object_services/object_server.js
karthikps97: removed

src/server/object_services/object_server.js
karthikps97: between finding the object and picking it up for transition, it could already be deleted. I would have more control here to check return a reply indicating this scenario. Modified this to throw an RPC error if its anything other

src/api/object_api.js
karthikps97: We actually dont need anything else here other than the status.

src/server/object_services/object_server.js
romayalon: good but also maybe worth rephrasing the warning because it's unclear

src/api/object_api.js
romayalon: yeah I mean that we usually don't return anything or throw an error on failure, because when success is false it's not really saying what happened

src/server/bg_services/lifecycle.js
romayalon: I understand, it'll be harder to debug if we unset the transition status that implies that we started transition but it failed, we can add it in the future if needed if you feel the logging are enough for now and the query can be changed as well

src/server/bg_services/lifecycle.js
romayalon: what is the benefit of creating a new one if there is already one existing that the other BG workers are using?

src/server/object_services/object_server.js
karthikps97: Rephrased it now

src/server/bg_services/lifecycle.js
karthikps97: We'll refactor this if needed, as per offline discussion.

src/server/bg_services/lifecycle.js
```diff
@@ -265,4 +403,109 @@ function update_lifecycle_rules_last_sync(bucket, rules) {
...
+                        system_id: system._id,
+                        account_id: system.owner._id,
+                        role: 'admin'
+                    })
+                });
+                dbg.log1('object successfully transitioned', obj_id);
+                return obj_id;
+            } else {
+                dbg.log1("got success false status from archive_server for object", obj_id);
+                throw new Error("archive_server failed to transition object");
```
romayalon: add here the object name

src/server/bg_services/lifecycle.js
```diff
@@ -265,4 +403,109 @@ function update_lifecycle_rules_last_sync(bucket, rules) {
...
+            });
+
+            if (!res) {
+                dbg.warn("LIFECYCLE_TRANSITION: object no longer valid for transition", obj_id);
+                return;
+            }
+
+            dbg.log1("LIFECYCLE_TRANSITION: transition status updated to 'in_progress' for object:", obj_id);
+
+            const rpc_client = server_rpc.rpc.new_client({
```
romayalon: why do we create an rpc_client inside the loop?

src/server/bg_services/lifecycle.js
```diff
@@ -265,4 +403,109 @@ function update_lifecycle_rules_last_sync(bucket, rules) {
...
+ * @param {string} target_storage_class - The target storage class to transition objects to (e.g. 'GLACIER').
+ * @returns {Promise<string[]>} Array of successfully transitioned object IDs.
+ */
+async function transition_objects(system, bucket_info, objects, target_storage_class) {
+    const batch_size = config.LIFECYCLE_TRANSITION_BATCH_SIZE;
+    const failed_objects = [];
+    const obj_ids = await P.map_with_concurrency(batch_size, objects, (async obj => {
+        const obj_id = obj.obj_id;
+        try {
+            const res = await server_rpc.client.object.update_transition_status({
```
romayalon, line 454: we should consider updating transition statuses of all the objects once and call archive_objects([objects]) instead of do it one by one
we are stressing the db and the RPC with concurrent 1000 calls or at least minimize the concurrency as a start on this PR

src/server/object_services/md_store.js
```diff
@@ -2402,6 +2419,193 @@ class MDStore {
...
+        /* 
+           Amazon S3 calculates the time by adding the number of days specified in the rule to the 
+           object creation time and rounding up the resulting time to the next day at midnight UTC 
+        */
+        const query = `SELECT *
+        FROM ${this._objects.name}
+        WHERE data->>'bucket' = $1
+        AND (
+                date_trunc('day', (data->>'create_time')::timestamptz) + interval '1 day'
+            ) < $2::timestamptz
```
romayalon, line 2476: the filter properties will be added in the next PR?

src/server/object_services/md_store.js
```diff
@@ -2402,6 +2419,193 @@ class MDStore {
...
+
+        /* 
+           Amazon S3 calculates the time by adding the number of days specified in the rule to the 
+           object creation time and rounding up the resulting time to the next day at midnight UTC 
+        */
+        const query = `SELECT *
+        FROM ${this._objects.name}
+        WHERE data->>'bucket' = $1
+        AND (
+                date_trunc('day', (data->>'create_time')::timestamptz) + interval '1 day'
```
romayalon, line 2475: on expiration they do
```
  if (max_create_time) {
      params.push(new Date(moment.unix(max_create_time).toISOString()).toISOString());
      sql_conditions.push(`data->>'create_time' < $${params.length}`);
  }
```
is there a reason why not do the same?

src/server/object_services/md_store.js
```diff
@@ -2402,6 +2419,193 @@ class MDStore {
...
+            ) < $2::timestamptz
+        AND (data->'deleted' IS NULL OR data->'deleted' = 'null'::jsonb)
+        AND (data->'reclaimed' IS NULL OR data->'reclaimed' = 'null'::jsonb)
+        AND (data->'upload_started' IS NULL OR data->'upload_started' = 'null'::jsonb)
```
romayalon, line 2479: in general, why not taking delete_objects_by_query() and moving the part of the query that checks what needs to be updated before deleting and re-using it for transition? later you can add the transition status is null query and for the delete

src/server/object_services/md_store.js
```diff
@@ -2402,6 +2419,193 @@ class MDStore {
...
+        // we resume from versions with a lower version_seq than the marker.
+        if (key_marker && version_seq_marker) {
+            ranked_conditions.push(
+                `((key = '${key_marker}' AND version_seq < ${version_seq_marker}) OR key > '${key_marker}')`
+            );
+        } else if (key_marker) {
+            ranked_conditions.push(`key > '${key_marker}'`);
+        }
+
+        const query = `
```
romayalon, line 2574: why not re-use remove_noncurrent_versions() - move out to a helper function the common logic and add the delete/update transition status in the original functions

src/server/object_services/object_server.js
```diff
@@ -1077,7 +1077,145 @@ async function update_object_md(req) {
...
+    let objects = [];
+
+    if (is_latest) {
+        objects = await MDStore.instance().find_objects_to_transition({
+            bucket: req.bucket,
+            batch_size,
+            key_marker,
+            transition_ts,
+        });
+    } else {
```
romayalon, line 1149: on versioning enabled buckets, if a lifecycle rule has transition and non_current_version_transition actions, will we handle both actions?

config.js
```diff
@@ -520,6 +520,7 @@ config.LIFECYCLE_INTERVAL = 8 * 60 * 60 * 1000; // 8h
 config.LIFECYCLE_BATCH_SIZE = 1000;
 config.LIFECYCLE_SCHEDULE_MIN = 5 * 1000 * 60; // run every 5 minutes
 config.LIFECYCLE_ENABLED = true;
+config.LIFECYCLE_TRANSITION_BATCH_SIZE = 50;
```
romayalon, line 523: why 50?

src/server/bg_services/lifecycle.js
karthikps97: Yes, I'll add this. Currently I have kept a low number on the concurrency (50)

src/server/bg_services/lifecycle.js
karthikps97: Moved this outside.

src/server/object_services/md_store.js
karthikps97: Yes, I have those in the next one.

src/server/object_services/md_store.js
karthikps97: S3 calculates time based on midnight UTC. Following this would mean an object can be transitioned earlier than midnight

src/server/object_services/md_store.js
karthikps97: I have more conditions coming up in the next PR. I think combining all would make the function complex. I can try to reuse some parts, but I'll check along with the next PR changes

src/server/object_services/object_server.js
karthikps97: Yes, we will now handle both. This is handled at the function call itself.

```js
await process_transition(system, bucket, rule.transition);
await process_transition(system, bucket, rule.noncurrent_version_transition);
```

https://github.com/noobaa/noobaa-core/pull/9840/changes#diff-c6480de3ae600ad0fc3ed51f3b8bfaa5572864c09a080be057fedda180a48b4fR348

src/server/object_services/md_store.js
karthikps97: Same as the other function, I have more conditional checks in the next PR, so it would make it more complex.

romayalon (approved):

shirady:
Hi,
@romayalon @karthikps97 I got a comment from CodeRabbit that is out of scope for my PR, but I think it is related to this one: [link](https://github.com/noobaa/noobaa-core/pull/9894#pullrequestreview-4847023944).

PR #9853, opened 2026-07-22
Test coverage for IAM Roles

aayushchouhan09:
### Describe the Problem
Currently, no tests implemented for IAM Roles.

### Explain the Changes
1. Jest: Role input-validation cases in existing ops validation file (no separate roles unit file / no mocked success-path ops). Utils coverage for create_arn_for_role, parse_role_arn, and role action message titles.
2. Mocha IAM: Role + role-policy happy path and negatives (duplicate, AccessDenied, NoSuchEntity, DeleteConflict, PathPrefix), aligned with user-API style; NC skip.
3. Mocha STS: Assumed-role suite for restrictive role policy → listBuckets AccessDenied, and ListAllMyBuckets policy → success (NC skip). Shared generate_sts_client (SDK v3) + generate_s3_client session-token/HTTPS support for reuse.

### Testing Instructions:
1.  From `noobaa-core` dir (run the below commands):
`$ make run-single-test testpath=integration_tests/api/iam/test_iam_basic_integration.js`
`$ make run-single-test testpath=integration_tests/api/sts/test_sts.js`

(jest tests)
`$ npx jest   src/test/unit_tests/api/iam/test_iam_utils.test.js \ src/test/unit_tests/api/iam/test_iam_ops_input_validation.test.js   --no-coverage`

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Tests**
  * Expanded IAM integration coverage to include role and inline role policy flows (create/list/get/update/delete), including path filtering, conflict handling, and missing-resource/authorization failures.
  * Expanded IAM unit tests for role ARN helpers and IAM role handler input validation (names, path format, pagination, and session duration constraints).
  * Updated STS integration tests to use SDK v3 command-based flows, improved response/error assertions, and added assumed-role authorization scenarios for S3 access.
  * Enhanced test utilities to support optional session tokens for both S3 and STS clients.

### DIFF
--- a/src/test/integration_tests/api/iam/test_iam_basic_integration.js
+++ b/src/test/integration_tests/api/iam/test_iam_basic_integration.js
@@ -27,7 +27,10 @@ const { ListUsersCommand, CreateUserCommand, GetUserCommand, UpdateUserCommand, 
     ListRoleTagsCommand, ListSAMLProvidersCommand, ListServerCertificatesCommand,
     ListServerCertificateTagsCommand, ListServiceSpecificCredentialsCommand,
     ListSigningCertificatesCommand, ListSSHPublicKeysCommand,
-    ListVirtualMFADevicesCommand } = require('@aws-sdk/client-iam');
+    ListVirtualMFADevicesCommand,
+    CreateRoleCommand, GetRoleCommand, UpdateRoleCommand, DeleteRoleCommand,
+    PutRolePolicyCommand, GetRolePolicyCommand, DeleteRolePolicyCommand, ListRolePoliciesCommand,
+    UpdateAssumeRolePolicyCommand } = require('@aws-sdk/client-iam');
 const { ACCESS_KEY_STATUS_ENUM } = require('../../../../endpoint/iam/iam_constants');
 const IamError = require('../../../../endpoint/iam/iam_errors').IamError;
 
@@ -369,6 +372,149 @@ mocha.describe('IAM integration tests', async function() {
                 };
                 const command = new DeleteUserPolicyCommand(input);
                 const response = await iam_account.send(command);
+                _check_status_code_ok(response);
+            });
+        });
+
+        mocha.describe('IAM Role API', async function() {
+            // containerized-only (NotImplemented on NC)
+            if (is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+            const role_name = 'TestRole';
+            const updated_role_description = 'updated-role-description';
+            const assume_role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRole"}]}';
+            const assume_role_with_web_identity_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRoleWithWebIdentity"}]}';
+
+            mocha.it('list roles - should be empty', async function() {
+                const response = await iam_account.send(new ListRolesCommand({}));
+                _check_status_code_ok(response);
+                assert.equal(response.Roles.length, 0);
+            });
+
+            mocha.it('create a role', async function() {
+                const response = await iam_account.send(new CreateRoleCommand({
+                    RoleName: role_name,
+                    AssumeRolePolicyDocument: assume_role_policy,
+                    Description: 'integration-test-role',
+                    MaxSessionDuration: 7200,
+                }));
+                _check_status_code_ok(response);
+                assert.equal(response.Role.RoleName, role_name);
+                assert.equal(response.Role.MaxSessionDuration, 7200);
+
+                const list_response = await iam_account.send(new ListRolesCommand({}));
+                _check_status_code_ok(list_response);
+                assert.equal(list_response.Roles.length, 1);
+                assert.equal(list_response.Roles[0].RoleName, role_name);
+            });
+
+            mocha.it('get a role', async function() {
+                const response = await iam_account.send(new GetRoleCommand({ RoleName: role_name }));
+                _check_status_code_ok(response);
+                assert.equal(response.Role.RoleName, role_name);
+                assert(response.Role.AssumeRolePolicyDocument !== undefined);
+            });
+
+            mocha.it('update a role', async function() {
+                const response = await iam_account.send(new UpdateRoleCommand({
+                    RoleName: role_name,
+                    Description: updated_role_description,
+                    MaxSessionDuration: 3600,
+                }));
+                _check_status_code_ok(response);
+
+                const get_response = await iam_account.send(new GetRoleCommand({ RoleName: role_name }));
+                _check_status_code_ok(get_response);
+                assert.equal(get_response.Role.Description, updated_role_description);
+                assert.equal(get_response.Role.MaxSessionDuration, 3600);
+            });
+
+            mocha.it('update assume role policy', async function() {
+                const response = await iam_account.send(new UpdateAssumeRolePolicyCommand({
+                    RoleName: role_name,
+                    PolicyDocument: assume_role_with_web_identity_policy,
+                }));
+                _check_status_code_ok(response);
+
+                const get_response = await iam_account.send(new GetRoleCommand({ RoleName: role_name }));
+                _check_status_code_ok(get_response);
+                const assume_role_policy_document = JSON.parse(get_response.Role.AssumeRolePolicyDocument);
+                assert.equal(assume_role_policy_document.Statement[0].Action, 'sts:AssumeRoleWithWebIdentity');
+            });
+
+            mocha.it('delete a role', async function() {
+                const response = await iam_account.send(new DeleteRoleCommand({ RoleName: role_name }));
+                _check_status_code_ok(response);
+
+                const list_response = await iam_account.send(new ListRolesCommand({}));
+                _check_status_code_ok(list_response);
+                assert.equal(list_response.Roles.length, 0);
+            });
+        });
+
+        mocha.describe('IAM Role Policy API', async function() {
+            if (is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+            const role_name = 'PolicyTestRole';
+            const policy_name = 'RoleS3Access';
+            const assume_role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRole"}]}';
+            const role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"s3:*","Resource":"*"}]}';
+
+            mocha.before(async function() {
+                const response = await iam_account.send(new CreateRoleCommand({
+                    RoleName: role_name,
+                    AssumeRolePolicyDocument: assume_role_policy,
+                }));
+                _check_status_code_ok(response);
+            });
+
+            mocha.after(async function() {
+                const list_response = await iam_account.send(new ListRolePoliciesCommand({ RoleName: role_name }));
+                for (const name of list_response.PolicyNames) {
+                    await iam_account.send(new DeleteRolePolicyCommand({
+                        RoleName: role_name,
+                        PolicyName: name,
+                    }));
+                }
+                const response = await iam_account.send(new DeleteRoleCommand({ RoleName: role_name }));
+                _check_status_code_ok(response);
+            });
+
+            mocha.it('list role policies - should be empty', async function() {
+                const response = await iam_account.send(new ListRolePoliciesCommand({ RoleName: role_name }));
+                _check_status_code_ok(response);
+                assert.equal(response.PolicyNames.length, 0);
+            });
+
+            mocha.it('put role policy', async function() {
+                const response = await iam_account.send(new PutRolePolicyCommand({
+                    RoleName: role_name,
+                    PolicyName: policy_name,
+                    PolicyDocument: role_policy,
+                }));
+                _check_status_code_ok(response);
+
+                const list_response = await iam_account.send(new ListRolePoliciesCommand({ RoleName: role_name }));
+                _check_status_code_ok(list_response);
+                assert.equal(list_response.PolicyNames.length, 1);
+                assert.equal(list_response.PolicyNames[0], policy_name);
+            });
+
+            mocha.it('get role policy', async function() {
+                const response = await iam_account.send(new GetRolePolicyCommand({
+                    RoleName: role_name,
+                    PolicyName: policy_name,
+                }));
+                _check_status_code_ok(response);
+                assert.equal(response.RoleName, role_name);
+                assert.equal(response.PolicyName, policy_name);
+                const policy_document = JSON.parse(response.PolicyDocument);
+                assert.deepEqual(policy_document.Statement[0], { Effect: 'Allow', Action: 's3:*', Resource: '*' });
+            });
+
+            mocha.it('delete role policy', async function() {
+                const response = await iam_account.send(new DeleteRolePolicyCommand({
+                    RoleName: role_name,
+                    PolicyName: policy_name,
+                }));
                 _check_status_code_ok(response);
             });
         });
@@ -2129,6 +2275,435 @@ mocha.describe('IAM integration tests', async function() {
                 });
             });
         });
+
+        mocha.describe('IAM Role API', async function() {
+            // containerized-only (NotImplemented on NC)
+            if (is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+
+            const role_name = 'AdvRole';
+            const role_path = '/div_a/sub_div_b/';
+            const assume_role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRole"}]}';
+            const role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"s3:*","Resource":"*"}]}';
+            const policy_name = 'AdvRolePolicy';
+            const iam_username = 'RoleTestUser';
+            let access_key_id;
+            let iam_user_client;
+
+            mocha.before(async function() {
+                await create_iam_role(iam_account, role_name, assume_role_policy);
+                await create_iam_user(iam_account, iam_username);
+                const res = await create_access_key_iam_user(iam_account, iam_username);
+                access_key_id = res.access_key_id;
+                iam_user_client = generate_iam_client(res.access_key_id, res.secret_access_key, coretest_endpoint_iam);
+            });
+
+            mocha.after(async function() {
+                await delete_all_inline_role_policies(iam_account, role_name);
+                await delete_iam_role(iam_account, role_name);
+                if (access_key_id) await delete_access_key_iam_user(iam_account, access_key_id, iam_username);
+                await delete_iam_user(iam_account, iam_username);
+            });
+
+            mocha.describe('IAM CreateRole API', async function() {
+                mocha.it('create a role with role name that already exists should fail', async function() {
+                    try {
+                        await iam_account.send(new CreateRoleCommand({
+                            RoleName: role_name, AssumeRolePolicyDocument: assume_role_policy,
+                        }));
+                        assert.fail('create role with existing role name - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.EntityAlreadyExists.code);
+                        assert.strictEqual(err_message, `Role with name ${role_name} already exists.`);
+                    }
+                });
+
+                mocha.it('create a role - requester is IAM user - should fail', async function() {
+                    try {
+                        await iam_user_client.send(new CreateRoleCommand({
+                            RoleName: 'RoleByIamUser', AssumeRolePolicyDocument: assume_role_policy,
+                        }));
+                        assert.fail('create role - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:CreateRole'));
+                    }
+                });
+            });
+
+            mocha.describe('IAM GetRole API', async function() {
+                mocha.it('get a role - non-existing role - should fail', async function() {
+                    const role_name_non_existing = 'non-existing-role';
+                    try {
+                        await iam_account.send(new GetRoleCommand({ RoleName: role_name_non_existing }));
+                        assert.fail('get role - non-existing role - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role with name ${role_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('get a role - requester is IAM user - should fail', async function() {
+                    try {
+                        await iam_user_client.send(new GetRoleCommand({ RoleName: role_name }));
+                        assert.fail('get role - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:GetRole'));
+                    }
+                });
+            });
+
+            mocha.describe('IAM UpdateRole API', async function() {
+                mocha.it('update a role - non-existing role - should fail', async function() {
+                    const role_name_non_existing = 'non-existing-role';
+                    try {
+                        await iam_account.send(new UpdateRoleCommand({
+                            RoleName: role_name_non_existing, Description: 'updated description',
+                        }));
+                        assert.fail('update role - non-existing role - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role with name ${role_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('update a role - requester is IAM user - should fail', async function() {
+                    try {
+                        await iam_user_client.send(new UpdateRoleCommand({
+                            RoleName: role_name, Description: 'updated by iam user',
+                        }));
+                        assert.fail('update role - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:UpdateRole'));
+                    }
+                });
+            });
+
+            mocha.describe('IAM UpdateAssumeRolePolicy API', async function() {
+                mocha.it('update assume role policy - non-existing role - should fail', async function() {
+                    const role_name_non_existing = 'non-existing-role';
+                    try {
+                        await iam_account.send(new UpdateAssumeRolePolicyCommand({
+                            RoleName: role_name_non_existing, PolicyDocument: assume_role_policy,
+                        }));
+                        assert.fail('update assume role policy - non-existing role - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role with name ${role_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('update assume role policy - requester is IAM user - should fail', async function() {
+                    try {
+                        await iam_user_client.send(new UpdateAssumeRolePolicyCommand({
+                            RoleName: role_name, PolicyDocument: assume_role_policy,
+                        }));
+                        assert.fail('update assume role policy - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:UpdateAssumeRolePolicy'));
+                    }
+                });
+            });
+
+            mocha.describe('IAM DeleteRole API', async function() {
+                mocha.it('delete a role - non-existing role - should fail', async function() {
+                    const role_name_non_existing = 'non-existing-role';
+                    try {
+                        await iam_account.send(new DeleteRoleCommand({ RoleName: role_name_non_existing }));
+                        assert.fail('delete role - non-existing role - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role with name ${role_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('delete a role - requester is IAM user - should fail', async function() {
+                    try {
+                        await iam_user_client.send(new DeleteRoleCommand({ RoleName: role_name }));
+                        assert.fail('delete role - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:DeleteRole'));
+                    }
+                });
+
+                mocha.it('delete a role - role has inline IAM policy - should fail', async function() {
+                    await iam_account.send(new PutRolePolicyCommand({
+                        RoleName: role_name, PolicyName: policy_name, PolicyDocument: role_policy,
+                    }));
+                    try {
+                        await iam_account.send(new DeleteRoleCommand({ RoleName: role_name }));
+                        assert.fail('delete role - role has inline IAM policy - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.DeleteConflict.code);
+                        assert.strictEqual(err_message, 'Cannot delete entity, must delete policies first.');
+                    }
+                    await iam_account.send(new DeleteRolePolicyCommand({
+                        RoleName: role_name, PolicyName: policy_name,
+                    }));
+                });
+            });
+
+            mocha.describe('IAM ListRoles API', async function() {
+                mocha.it('list roles - requester is IAM user - should fail', async function() {
+                    try {
+                        await iam_user_client.send(new ListRolesCommand({}));
+                        assert.fail('list roles - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:ListRoles'));
+                    }
+                });
+
+                mocha.it('list roles - by path prefix', async function() {
+                    const role_under_path = 'test-role-1';
+                    await create_iam_role(iam_account, role_under_path, assume_role_policy, role_path);
+                    const response = await iam_account.send(new ListRolesCommand({ PathPrefix: role_path }));
+                    _check_status_code_ok(response);
+                    assert.equal(response.Roles.length, 1);
+                    assert.equal(response.Roles[0].RoleName, role_under_path);
+                    assert.equal(response.Roles[0].Path, role_path);
+                    await delete_iam_role(iam_account, role_under_path);
+                });
+            });
+        });
+
+        mocha.describe('IAM Role Policy API', async function() {
+            // containerized-only (NotImplemented on NC)
+            if (is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+
+            const role_name = 'AdvPolicyRole';
+            const assume_role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRole"}]}';
+            const iam_username = 'RolePolicyTestUser';
+            const policy_name = 'AllAccessPolicy';
+            const role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"*","Resource":"*"}]}';
+            const policy_name_to_replace = 'my_policy';
+            const role_policy_replaced = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"s3:CreateBucket","Resource":"*"}]}';
+            const policy_name_multiple_statements = 'MultipleStatementsPolicy';
+            const role_policy_multiple_statements = '{"Version":"2012-10-17","Statement":[{"Sid":"id1","Effect":"Allow","Action":"s3:GetBucketLocation","Resource":"*"},{"Sid":"id2","Effect":"Allow","Action":"s3:CreateBucket","Resource":"*"},{"Sid":"","Effect":"Allow","Action":"s3:HeadBucket","Resource":"*"}]}';
+            let access_key_id;
+            let iam_user_client;
+
+            mocha.before(async function() {
+                await create_iam_role(iam_account, role_name, assume_role_policy);
+                await create_iam_user(iam_account, iam_username);
+                const res = await create_access_key_iam_user(iam_account, iam_username);
+                access_key_id = res.access_key_id;
+                iam_user_client = generate_iam_client(res.access_key_id, res.secret_access_key, coretest_endpoint_iam);
+            });
+
+            mocha.after(async function() {
+                await delete_all_inline_role_policies(iam_account, role_name);
+                await delete_iam_role(iam_account, role_name);
+                await delete_access_key_iam_user(iam_account, access_key_id, iam_username);
+                await delete_iam_user(iam_account, iam_username);
+            });
+
+            mocha.describe('IAM PutRolePolicy API', async function() {
+                mocha.it('put role policy - non-existing role - should throw an error', async function() {
+                    const role_name_non_existing = 'non-existing-role';
+                    try {
+                        await iam_account.send(new PutRolePolicyCommand({
+                            RoleName: role_name_non_existing, PolicyName: policy_name, PolicyDocument: role_policy,
+                        }));
+                        assert.fail('put role policy - non-existing role - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role with name ${role_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('put role policy - requester is IAM user - should throw an error', async function() {
+                    try {
+                        await iam_user_client.send(new PutRolePolicyCommand({
+                            RoleName: role_name, PolicyName: policy_name, PolicyDocument: role_policy,
+                        }));
+                        assert.fail('put role policy - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:PutRolePolicy'));
+                    }
+                });
+
+                mocha.it('put role policy - with large number of statements - should throw an error', async function() {
+                    const large_policy = JSON.stringify({
+                        Version: '2012-10-17',
+                        Statement: Array.from({ length: 50 }, () => ({ Effect: 'Allow', Action: '*', Resource: '*' })),
+                    });
+                    try {
+                        await iam_account.send(new PutRolePolicyCommand({
+                            RoleName: role_name, PolicyName: policy_name, PolicyDocument: large_policy,
+                        }));
+                        assert.fail('put role policy - with large number of statements - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.LimitExceeded.code);
+                        assert.strictEqual(err_message, `Maximum policy size of 2048 bytes exceeded for role ${role_name}`);
+                    }
+                });
+
+                mocha.it('put role policy with multiple statements', async function() {
+                    const response = await iam_account.send(new PutRolePolicyCommand({
+                        RoleName: role_name,
+                        PolicyName: policy_name_multiple_statements,
+                        PolicyDocument: role_policy_multiple_statements,
+                    }));
+                    _check_status_code_ok(response);
+                });
+
+                mocha.it('put role policy with the same policy name - should replace the policy', async function() {
+                    await iam_account.send(new PutRolePolicyCommand({
+                        RoleName: role_name, PolicyName: policy_name_to_replace, PolicyDocument: role_policy,
+                    }));
+                    await iam_account.send(new PutRolePolicyCommand({
+                        RoleName: role_name, PolicyName: policy_name_to_replace, PolicyDocument: role_policy_replaced,
+                    }));
+                    const response = await iam_account.send(new GetRolePolicyCommand({
+                        RoleName: role_name, PolicyName: policy_name_to_replace,
+                    }));
+                    _check_status_code_ok(response);
+                    assert.deepEqual(JSON.parse(response.PolicyDocument).Statement[0],
+                        { Effect: 'Allow', Action: 's3:CreateBucket', Resource: '*' });
+                });
+            });
+
+            mocha.describe('IAM GetRolePolicy API', async function() {
+                mocha.it('get role policy - non-existing role - should throw an error', async function() {
+                    const role_name_non_existing = 'non-existing-role';
+                    try {
+                        await iam_account.send(new GetRolePolicyCommand({
+                            RoleName: role_name_non_existing, PolicyName: policy_name,
+                        }));
+                        assert.fail('get role policy - non-existing role - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role with name ${role_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('get role policy - non-existing role policy - should throw an error', async function() {
+                    const policy_name_non_existing = 'non-existing-policy';
+                    try {
+                        await iam_account.send(new GetRolePolicyCommand({
+                            RoleName: role_name, PolicyName: policy_name_non_existing,
+                        }));
+                        assert.fail('get role policy - non-existing role policy - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role policy with name ${policy_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('get role policy - requester is IAM user - should throw an error', async function() {
+                    try {
+                        await iam_user_client.send(new GetRolePolicyCommand({
+                            RoleName: role_name, PolicyName: policy_name,
+                        }));
+                        assert.fail('get role policy - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:GetRolePolicy'));
+                    }
+                });
+
+                mocha.it('get role policy with multiple statements', async function() {
+                    const response = await iam_account.send(new GetRolePolicyCommand({
+                        RoleName: role_name, PolicyName: policy_name_multiple_statements,
+                    }));
+                    _check_status_code_ok(response);
+                    const statements = JSON.parse(response.PolicyDocument).Statement;
+                    assert.strictEqual(statements.length, 3);
+                    assert.deepStrictEqual(statements.map(s => s.Sid), ['id1', 'id2', '']);
+                });
+            });
+
+            mocha.describe('IAM DeleteRolePolicy API', async function() {
+                mocha.it('delete role policy - non-existing role - should throw an error', async function() {
+                    const role_name_non_existing = 'non-existing-role';
+                    try {
+                        await iam_account.send(new DeleteRolePolicyCommand({
+                            RoleName: role_name_non_existing, PolicyName: policy_name,
+                        }));
+                        assert.fail('delete role policy - non-existing role - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role with name ${role_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('delete role policy - non-existing role policy - should throw an error', async function() {
+                    const policy_name_non_existing = 'non-existing-policy';
+                    try {
+                        await iam_account.send(new DeleteRolePolicyCommand({
+                            RoleName: role_name, PolicyName: policy_name_non_existing,
+                        }));
+                        assert.fail('delete role policy - non-existing role policy - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role policy with name ${policy_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('delete role policy - requester is IAM user - should throw an error', async function() {
+                    try {
+                        await iam_user_client.send(new DeleteRolePolicyCommand({
+                            RoleName: role_name, PolicyName: policy_name,
+                        }));
+                        assert.fail('delete role policy - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:DeleteRolePolicy'));
+                    }
+                });
+            });
+
+            mocha.describe('IAM ListRolePolicies API', async function() {
+                mocha.it('list role policies for non-existing role - should throw an error', async function() {
+                    const role_name_non_existing = 'non-existing-role';
+                    try {
+                        await iam_account.send(new ListRolePoliciesCommand({ RoleName: role_name_non_existing }));
+                        assert.fail('list role policies for non-existing role - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.NoSuchEntity.code);
+                        assert.strictEqual(err_message, `The role with name ${role_name_non_existing} cannot be found.`);
+                    }
+                });
+
+                mocha.it('list role policies - requester is IAM user - should throw an error', async function() {
+                    try {
+                        await iam_user_client.send(new ListRolePoliciesCommand({ RoleName: role_name }));
+                        assert.fail('list role policies - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:ListRolePolicies'));
+                    }
+                });
+            });
+        });
     });
 });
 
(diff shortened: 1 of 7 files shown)


src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -373,6 +376,147 @@ mocha.describe('IAM integration tests', async function() {
             });
         });
 
+        // IAM Role APIs are containerized-only (AccountSpaceFS returns NotImplemented on NC).
+        mocha.describe('IAM Role API', async function() {
+            if (is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+            const role_name = 'TestRole';
+            const trust_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRole"}]}';
+            const updated_trust_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRoleWithWebIdentity"}]}';
```
shirady: Are you sure that on AWS we can have a trust policy with `*` in the Principal?
You can take example from AWS, for example ([link](https://aws.amazon.com/blogs//security/how-to-use-trust-policies-with-iam-roles/)):
```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::111122223333:root"
      },
      "Action": "sts:AssumeRole"
    }
  ]
}
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -373,6 +376,147 @@ mocha.describe('IAM integration tests', async function() {
...
+                const response = await iam_account.send(new GetRoleCommand({ RoleName: role_name }));
+                _check_status_code_ok(response);
+                assert.equal(response.Role.RoleName, role_name);
+                assert(response.Role.AssumeRolePolicyDocument !== undefined);
+            });
+
+            mocha.it('update a role', async function() {
+                const response = await iam_account.send(new UpdateRoleCommand({
+                    RoleName: role_name,
+                    Description: 'updated-role-description',
```
shirady: Could you move it to a variable? (I saw it was used twice 'updated-role-description')

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -373,6 +376,147 @@ mocha.describe('IAM integration tests', async function() {
...
+                const get_response = await iam_account.send(new GetRoleCommand({ RoleName: role_name }));
+                _check_status_code_ok(get_response);
+                assert.equal(get_response.Role.Description, 'updated-role-description');
+                assert.equal(get_response.Role.MaxSessionDuration, 3600);
+            });
+
+            mocha.it('update assume role policy', async function() {
+                const response = await iam_account.send(new UpdateAssumeRolePolicyCommand({
+                    RoleName: role_name,
+                    PolicyDocument: updated_trust_policy,
```
shirady: It is `UpdateAssumeRolePolicy` so it should be for assume role policy, not trust policy, right?

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -373,6 +376,147 @@ mocha.describe('IAM integration tests', async function() {
...
+                assert.equal(list_response.Roles.length, 0);
+            });
+        });
+
+        mocha.describe('IAM Role Policy API', async function() {
+            if (is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+            const role_name = 'PolicyTestRole';
+            const policy_name = 'RoleS3Access';
+            const trust_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRole"}]}';
+            const role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"s3:*","Resource":"*"}]}';
```
shirady: Can we have in AWS trust policy with `*` in `Principal` and role policy with `*` in `Resource`?

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -373,6 +376,147 @@ mocha.describe('IAM integration tests', async function() {
...
+                _check_status_code_ok(response);
+            });
+
+            mocha.after(async function() {
+                try {
+                    await iam_account.send(new DeleteRolePolicyCommand({
+                        RoleName: role_name,
+                        PolicyName: policy_name,
+                    }));
+                } catch (_err) { /* ignore if already deleted */ }
```
shirady: I think we'd better avoid catching and ignoring; it might hide errors.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2273,92 @@ mocha.describe('IAM integration tests', async function() {
...
         });
+
+        // IAM Role APIs are containerized-only (AccountSpaceFS returns NotImplemented on NC).
+        mocha.describe('IAM Role API', async function() {
+            if (is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+
+            const role_name = 'AdvRole';
+            const role_path = '/div_a/sub_div_b/';
+            const trust_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRole"}]}';
+            const role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Action":"s3:*","Resource":"*"}]}';
```
shirady: Raising the questions about the use of `*` again.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2273,92 @@ mocha.describe('IAM integration tests', async function() {
...
+            mocha.before(async function() {
+                await create_iam_role(iam_account, role_name, trust_policy);
+                await create_iam_user(iam_account, iam_username);
+                const res = await create_access_key_iam_user(iam_account, iam_username);
+                access_key_id = res.access_key_id;
+                iam_user_client = generate_iam_client(res.access_key_id, res.secret_access_key, coretest_endpoint_iam);
+            });
+
+            mocha.after(async function() {
+                const ignore = () => undefined;
```
shirady: What is this line? Why is it needed? `const ignore = () => undefined;`

I don't think that we need to ignore errors.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2273,92 @@ mocha.describe('IAM integration tests', async function() {
...
+                await iam_account.send(new DeleteRolePolicyCommand({
+                    RoleName: role_name, PolicyName: policy_name,
+                })).catch(ignore);
+                await delete_iam_role(iam_account, role_name).catch(ignore);
+                if (access_key_id) await delete_access_key_iam_user(iam_account, access_key_id, iam_username).catch(ignore);
+                await delete_iam_user(iam_account, iam_username).catch(ignore);
+            });
+
+            mocha.it('create role - already exists / IAM user denied', async function() {
```
shirady: Could you separate it into 2 cases?
mocha.it('create role - already exists`)...
mocha.it('IAM user denied`)...

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2273,92 @@ mocha.describe('IAM integration tests', async function() {
...
+                }
+                try {
+                    await iam_user_client.send(new CreateRoleCommand({
+                        RoleName: 'RoleByIamUser', AssumeRolePolicyDocument: trust_policy,
+                    }));
+                    assert.fail('expected AccessDeniedException');
+                } catch (err) {
+                    const { err_code, err_message } = _get_err_code_and_message(err);
+                    assert.equal(err_code, IamError.AccessDeniedException.code);
+                    assert.ok(err_message.includes('iam:CreateRole'));
```
shirady: Maybe you can check also in the message that this permission is missing.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2273,92 @@ mocha.describe('IAM integration tests', async function() {
...
+                }
+            });
+
+            mocha.it('get missing role / delete role with policy / list by path prefix', async function() {
```
shirady: Again - could you separate the cases?

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2273,92 @@ mocha.describe('IAM integration tests', async function() {
...
+                }
+
+                await iam_account.send(new PutRolePolicyCommand({
+                    RoleName: role_name, PolicyName: policy_name, PolicyDocument: role_policy,
+                }));
+                try {
+                    await iam_account.send(new DeleteRoleCommand({ RoleName: role_name }));
+                    assert.fail('expected DeleteConflict');
+                } catch (err) {
+                    assert.equal(_get_err_code_and_message(err).err_code, IamError.DeleteConflict.code);
```
shirady: Could you check something about the message?
In AWS docs ([link](https://docs.aws.amazon.com/IAM/latest/APIReference/API_DeleteRole.html)), it should include information about having a policy:

> DeleteConflict
> The request was rejected because it attempted to delete a resource that has attached subordinate entities. The error message describes these entities.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2273,92 @@ mocha.describe('IAM integration tests', async function() {
...
+                }
+                await iam_account.send(new DeleteRolePolicyCommand({
+                    RoleName: role_name, PolicyName: policy_name,
+                }));
+
+                const path_role = 'PathPrefixedRole';
+                await create_iam_role(iam_account, path_role, trust_policy, role_path);
+                try {
```
shirady: Why do you need the "try...finally"?
You delete the roles in the `after`.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2159,6 +2389,19 @@ async function create_iam_user(iam_client, username_to_create, iam_path) {
...
+        AssumeRolePolicyDocument: trust_policy_document,
+    };
+    if (iam_path) input.Path = iam_path;
+    _check_status_code_ok(await iam_client.send(new CreateRoleCommand(input)));
+}
+
+async function delete_iam_role(iam_client, role_name_to_delete) {
+    _check_status_code_ok(await iam_client.send(new DeleteRoleCommand({ RoleName: role_name_to_delete })));
+}
+
```
shirady: The new functions were added between `create_iam_user` and `delete_iam_user`, could you move them to the end of the file?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
         }).promise(), stsErr.AccessDeniedException.code, stsErr.AccessDeniedException.message);
     });
 });
+
+// Restrictive IAM role policy authz — own suite so Session token tests stay unchanged.
```
shirady: What is authz? Could you write the explicit words?
What does the comment mean in simple words?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+mocha.describe('STS assumed-role restrictive policy tests', function() {
+    if (is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+
+    const { rpc_client } = coretest;
+    const owner_email = 'role-authz-owner';
+    const assumer_email = 'role-authz-assumer';
```
shirady, line 934: Why do you need emails?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+    const role_name = 'role_authz_restrictive';
+    const policy_name = 'Role_Authz_S3Access';
+    const accounts = [{ email: owner_email }, { email: assumer_email }];
+    let sts_creds;
+    let owner_account_info;
+
+    mocha.before(async function() {
+        const self = this; // eslint-disable-line no-invalid-this
+        self.timeout(60000);
+        // Session token suite sets expiry to 0 and does not restore it; fix before AssumeRole.
```
shirady: I'm not sure I understand this comment

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+        sts_creds = {
+            endpoint: coretest.get_https_address_sts(),
+            region: 'us-east-1',
+            sslEnabled: true,
+            computeChecksums: true,
+            httpOptions: { agent: new https.Agent({ keepAlive: false, rejectUnauthorized: false }) },
+            s3ForcePathStyle: true,
+            signatureVersion: 'v4',
+            s3DisableBodySigning: false,
+        };
```
shirady: Could you create `generate_sts_client` in `src/test/system_tests/test_utils.js`?
It would be easier to reuse

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+            account.access_keys = (await rpc_client.account.create_account({
+                has_login: false,
+                s3_access: true,
+                name: account.email,
+                email: account.email,
+            })).access_keys;
+            const access_key = account.access_keys[0].access_key.unwrap();
+            const secret_key = account.access_keys[0].secret_key.unwrap();
+            account.sts = new AWS.STS({ ...sts_creds, accessKeyId: access_key, secretAccessKey: secret_key });
+            account.iam = generate_iam_client(access_key, secret_key, coretest.get_https_address_iam());
```
shirady: Suggesting to add suffix of `_client` hoping it would be cleaer.

```suggestion
            account.sts_client = new AWS.STS({ ...sts_creds, accessKeyId: access_key, secretAccessKey: secret_key });
            account.iam_client = generate_iam_client(access_key, secret_key, coretest.get_https_address_iam());
```

What about S3? Don't you need it?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+        }
+
+        owner_account_info = await rpc_client.account.read_account({ email: owner_email });
+        await accounts[0].iam.send(new CreateRoleCommand({
+            RoleName: role_name,
+            AssumeRolePolicyDocument: JSON.stringify({
+                Version: '2012-10-17',
+                Statement: [{
+                    Effect: 'Allow',
+                    Principal: { AWS: [assumer_email] },
```
shirady: Why with emails? I think you can use the account ID
cc: @naveenpaul1

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+            PolicyName: policy_name,
+            PolicyDocument: JSON.stringify({
+                Version: '2012-10-17',
+                Statement: [{
+                    Effect: 'Allow',
+                    Action: ['s3:GetObject'],
+                    Resource: ['arn:aws:s3:::no-such-bucket/*'],
+                }],
+            }),
+        }));
```
shirady, line 988: I thought it should be continued...
Could you write what you planned the test to include?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+    });
+
+    mocha.after(async function() {
```
shirady, line 991: As I have already written, I don't think we should ignore the errors; we might have real errors that we would not notice.

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+    const policy_name = 'Role_Authz_S3Access';
+    const accounts = [{ email: owner_email }, { email: assumer_email }];
+    let sts_creds;
+    let owner_account_info;
+
+    mocha.before(async function() {
+        const self = this; // eslint-disable-line no-invalid-this
+        self.timeout(60000);
+        // Session token suite sets expiry to 0 and does not restore it; fix before AssumeRole.
+        config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS = defualt_expiry_seconds * 1000;
```
shirady: I think it was raised by codeRabbot - you changed something in the `config` make sure you set it back as it was.

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+        const session = 'restrictive_policy_session';
+        const json = await assume_role_and_parse_xml(accounts[1].sts, {
+            RoleArn: `arn:aws:sts::${owner_id}:role/${role_name}`,
+            RoleSessionName: session,
+        });
+        const creds = validate_assume_role_response(json,
+            `arn:aws:sts::${owner_id}:assumed-role/${role_name}/${session}`,
+            `${owner_id}:${session}`, owner_key, defualt_expiry_seconds);
+        const temp_s3 = new AWS.S3({
+            ...sts_creds,
```
shirady: confusing that `temp_s3` is using `...sts_creds`.

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+        const owner_key = accounts[0].access_keys[0].access_key.unwrap();
+        const session = 'restrictive_policy_session';
+        const json = await assume_role_and_parse_xml(accounts[1].sts, {
+            RoleArn: `arn:aws:sts::${owner_id}:role/${role_name}`,
+            RoleSessionName: session,
+        });
+        const creds = validate_assume_role_response(json,
+            `arn:aws:sts::${owner_id}:assumed-role/${role_name}/${session}`,
+            `${owner_id}:${session}`, owner_key, defualt_expiry_seconds);
+        const temp_s3 = new AWS.S3({
```
shirady: I think you can use `generate_s3_client`

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +972,115 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+            ...sts_creds,
+            endpoint: coretest.get_https_address(),
+            accessKeyId: creds.access_key,
+            secretAccessKey: creds.secret_key,
+            sessionToken: creds.session_token,
+        });
+        try {
+            await temp_s3.listBuckets().promise();
```
shirady: I think you should AWS SDK V3 client for your s3 client

src/test/unit_tests/api/iam/test_iam_ops_input_validation.test.js
```diff
@@ -542,4 +552,315 @@ describe('input validation flow in IAM ops - IAM ACCESS KEY API', () => {
     });
 });
 
+describe('input validation flow in IAM ops - IAM ROLES API', () => {
+    const trust_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRole"}]}';
```
shirady: You can copy AWS example ([link](https://aws.amazon.com/blogs//security/how-to-use-trust-policies-with-iam-roles/)):

> {
>   "Version": "2012-10-17",
>   "Statement": [
>     {
>       "Effect": "Allow",
>       "Principal": {
>         "AWS": "arn:aws:iam::111122223333:root"
>       },
>       "Action": "sts:AssumeRole"
>     }
>   ]
> }

src/test/unit_tests/api/iam/test_iam_ops_input_validation.test.js
```diff
@@ -542,4 +552,315 @@ describe('input validation flow in IAM ops - IAM ACCESS KEY API', () => {
...
+    const role_name = 'TestRole';
+    const policy_name = 's3-access';
+
+    describe('iam_create_role', () => {
+        let res;
+
+        it('iam_create_role without required parameters (role_name) should throw error', async () => {
```
shirady, line 558: Could you add, as a comment near each variable, what made it invalid? (easier to understand)

src/test/unit_tests/api/iam/test_iam_ops_input_validation.test.js
```diff
@@ -542,4 +552,315 @@ describe('input validation flow in IAM ops - IAM ACCESS KEY API', () => {
...
+                throw new NoErrorThrownError();
+            } catch (err) {
+                expect(err).toBeInstanceOf(IamError);
+                expect(err).toHaveProperty('code', IamError.MalformedPolicyDocument.code);
+            }
+        });
+    });
+});
+
+describe('success path in IAM ops - IAM ROLES API', () => {
```
shirady: I don't understand how is this test is related to this file - `src/test/unit_tests/api/iam/test_iam_ops_input_validation.test.js` I think it should move.

src/test/unit_tests/api/iam/test_iam_utils.test.js
```diff
@@ -1,6 +1,12 @@
 /* Copyright (C) 2024 NooBaa */
 /* eslint-disable max-lines-per-function */
 'use strict';
+
+jest.mock('../../../../server/system_services/system_store', () => ({
+    get_instance: jest.fn(() => ({ data: {} })),
+}));
+
+const system_store = require('../../../../server/system_services/system_store');
```
shirady: Why was this added here?
It is a file of testing utility functions (should be simple unit tests)

src/test/unit_tests/api/iam/test_iam_utils.test.js
```diff
@@ -116,6 +122,82 @@ describe('get_action_message_title', () => {
         const res = iam_utils.get_action_message_title(action);
         expect(res).toBe(`iam:ListAccessKeys`);
     });
+
+    it('create_role', () => {
+        expect(iam_utils.get_action_message_title('create_role')).toBe('iam:CreateRole');
```
shirady: Why the newline? I think it can be removed in all the following tests

src/test/unit_tests/api/iam/test_iam_utils.test.js
```diff
@@ -116,6 +122,82 @@ describe('get_action_message_title', () => {
...
+        const res = iam_utils.create_arn_for_role(dummy_account_id, undefined, undefined);
+        expect(res).toBe(`${arn_prefix}${dummy_account_id}:role/`);
+    });
+
+    it('create_arn_for_role with role_name and no iam_path should return only role_name in arn', () => {
+        const res = iam_utils.create_arn_for_role(dummy_account_id, dummy_role_name, undefined);
+        expect(res).toBe(`${arn_prefix}${dummy_account_id}:role/${dummy_role_name}`);
+    });
+
+    it('create_arn_for_role with role_name and AWS DEFAULT PATH should return only role_name in arn', () => {
```
shirady:
```suggestion
    it('create_arn_for_role with role_name and IAM_DEFAULT_PATH should return only role_name in arn', () => {
```

src/test/unit_tests/api/iam/test_iam_utils.test.js
```diff
@@ -116,6 +122,82 @@ describe('get_action_message_title', () => {
...
+    it('should return MISSING_ROLE_ARN when role_arn is empty', () => {
+        expect(iam_utils.parse_role_arn('')).toEqual({ error: 'MISSING_ROLE_ARN' });
+    });
+
+    it('should return INVALID_ROLE_ARN when role_arn is malformed', () => {
+        expect(iam_utils.parse_role_arn('not-an-arn')).toEqual({ error: 'INVALID_ROLE_ARN' });
+    });
+});
+
+describe('resolve_iam_role_by_arn', () => {
```
shirady: Why do you add those tests?
I thought that the `system_store` in this file is temporary

shirady (commented):
I added a couple of questions and comments

src/test/integration_tests/api/iam/test_iam_basic_integration.js
aayushchouhan09: Yes we can put `{"AWS": "*"}` or `"*"` as wildcard to allow all the principals. Please check the [aws doc](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html#principal-anonymous)

src/test/integration_tests/api/iam/test_iam_basic_integration.js
aayushchouhan09: yes it is possible for both.

[trust principal](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html#principal-anonymous)

[role resource](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_resource.html#reference_policies_elements_resource_wildcards)

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: Emails are only for create_account / read_account (RPC requires them). Trust Principal now uses `arn:aws:iam::<assumer_account_id>:root`.

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: Emails are only for create_account / read_account (RPC requires them). Trust Principal now uses `arn:aws:iam::<assumer_account_id>:root`.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -373,6 +376,149 @@ mocha.describe('IAM integration tests', async function() {
             });
         });
 
+        // IAM Role APIs are containerized-only (AccountSpaceFS returns NotImplemented on NC)
+        mocha.describe('IAM Role API', async function() {
+            if (is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
+            const role_name = 'TestRole';
+            const updated_role_description = 'updated-role-description';
+            const assume_role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRole"}]}';
+            const updated_assume_role_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRoleWithWebIdentity"}]}';
```
shirady: Suggest using the full name so it would not be confused with `AssumeRoleWithWebIdentity` and `AssumeRole`
```suggestion
            const assume_role_with_web_identity_policy = '{"Version":"2012-10-17","Statement":[{"Effect":"Allow","Principal":{"AWS":"*"},"Action":"sts:AssumeRoleWithWebIdentity"}]}';
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
...
+                await iam_account.send(new PutRolePolicyCommand({
+                    RoleName: role_name, PolicyName: policy_name, PolicyDocument: role_policy,
+                }));
+                try {
+                    await iam_account.send(new DeleteRoleCommand({ RoleName: role_name }));
+                    assert.fail('expected DeleteConflict');
+                } catch (err) {
+                    const { err_code, err_message } = _get_err_code_and_message(err);
+                    assert.equal(err_code, IamError.DeleteConflict.code);
+                    assert.strictEqual(err_message, 'Cannot delete entity, must delete policies first.');
```
shirady: Is it the exact message that we have in AWS?
Could you share AWS's output?

src/test/integration_tests/api/sts/test_sts.js
shirady: @aayushchouhan09 can you add a comment about it?
@naveenpaul1 do we have a way to read the account using the account ID?

src/test/integration_tests/api/sts/test_sts.js
shirady: @aayushchouhan09 I see it is still open.

src/test/system_tests/test_utils.js
```diff
@@ -490,6 +493,21 @@ function generate_iam_client(access_key, secret_key, endpoint) {
...
+        sslEnabled: true,
+        computeChecksums: true,
+        httpOptions: { agent: new https.Agent({ keepAlive: false, rejectUnauthorized: false }) },
+        s3ForcePathStyle: true,
+        signatureVersion: 'v4',
+        s3DisableBodySigning: false,
+        accessKeyId: access_key,
+        secretAccessKey: secret_key,
+    });
+}
```
shirady: Why do you use the AWS SDK V2 and not AWS SDK V3?
If it is intended, then I would stay using it in the code and not add it to a helper function.

src/test/integration_tests/api/sts/test_sts.js
shirady: @aayushchouhan09 I see this is still open.

src/test/integration_tests/api/sts/test_sts.js
shirady: @aayushchouhan09 I see this is still open

src/test/unit_tests/api/iam/test_iam_ops_input_validation.test.js
shirady: @aayushchouhan09 I see this is still open.

src/test/unit_tests/api/iam/test_iam_utils.test.js
```diff
@@ -1,6 +1,7 @@
 /* Copyright (C) 2024 NooBaa */
 /* eslint-disable max-lines-per-function */
 'use strict';
+
```
shirady: I think this is a redundant newline.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
...
+                    await iam_account.send(new CreateRoleCommand({
+                        RoleName: role_name, AssumeRolePolicyDocument: assume_role_policy,
+                    }));
+                    assert.fail('expected EntityAlreadyExists');
+                } catch (err) {
+                    assert.equal(_get_err_code_and_message(err).err_code, IamError.EntityAlreadyExists.code);
+                }
+            });
+
+            mocha.it('create role - IAM user denied', async function() {
```
shirady: Suggesting having more details in the test title
```suggestion
            mocha.it('create a role - requester is IAM user - should fail', async function() {
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
...
+                    await iam_account.send(new DeleteRolePolicyCommand({
+                        RoleName: role_name, PolicyName: name,
+                    }));
+                }
+                await delete_iam_role(iam_account, role_name);
+                if (access_key_id) await delete_access_key_iam_user(iam_account, access_key_id, iam_username);
+                await delete_iam_user(iam_account, iam_username);
+            });
+
+            mocha.it('create role - already exists', async function() {
```
shirady: Suggesting having more details in the test title
```suggestion
            mocha.it('create a role with role name that already exists should fail', async function() {
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
...
+                try {
+                    await iam_account.send(new CreateRoleCommand({
+                        RoleName: role_name, AssumeRolePolicyDocument: assume_role_policy,
+                    }));
+                    assert.fail('expected EntityAlreadyExists');
```
shirady:
```suggestion
                    assert.fail('create user with existing role name - should throw an error');
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
...
+                    assert.equal(_get_err_code_and_message(err).err_code, IamError.EntityAlreadyExists.code);
+                }
+            });
+
+            mocha.it('create role - IAM user denied', async function() {
+                try {
+                    await iam_user_client.send(new CreateRoleCommand({
+                        RoleName: 'RoleByIamUser', AssumeRolePolicyDocument: assume_role_policy,
+                    }));
+                    assert.fail('expected AccessDeniedException');
```
shirady:
```suggestion
                    assert.fail('create role - requester is IAM user - should throw an error');
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
...
+                } catch (err) {
+                    const { err_code, err_message } = _get_err_code_and_message(err);
+                    assert.equal(err_code, IamError.AccessDeniedException.code);
+                    assert.ok(err_message.includes('not authorized to perform: iam:CreateRole'));
+                }
+            });
+
+            mocha.it('get missing role - should fail', async function() {
```
shirady:
```suggestion
            mocha.it('get a role - non-existing role - should fail', async function() {
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
...
+                try {
+                    await iam_account.send(new GetRoleCommand({ RoleName: 'non-existing-role' }));
+                    assert.fail('expected NoSuchEntity');
```
shirady:
```suggestion
                    assert.fail('get role - non-existing role - should throw an error');
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
                 });
             });
         });
+
+        // IAM Role APIs are containerized-only (AccountSpaceFS returns NotImplemented on NC)
+        mocha.describe('IAM Role API', async function() {
```
shirady: Could you align with the style that we have for IAM users?
For example, have a separate describe for each API, make sure that we covered all the test cases, test case titles, and messages in `assert.fail`.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
...
+            mocha.it('get missing role - should fail', async function() {
+                try {
+                    await iam_account.send(new GetRoleCommand({ RoleName: 'non-existing-role' }));
+                    assert.fail('expected NoSuchEntity');
+                } catch (err) {
+                    assert.equal(_get_err_code_and_message(err).err_code, IamError.NoSuchEntity.code);
+                }
+            });
+
+            mocha.it('delete role with policy - should fail', async function() {
```
shirady: There is a mismatch between the title and the actions:
title - delete role with policy
action - PutRolePolicyCommand

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,100 @@ mocha.describe('IAM integration tests', async function() {
...
+                    const { err_code, err_message } = _get_err_code_and_message(err);
+                    assert.equal(err_code, IamError.DeleteConflict.code);
+                    assert.strictEqual(err_message, 'Cannot delete entity, must delete policies first.');
+                }
+                await iam_account.send(new DeleteRolePolicyCommand({
+                    RoleName: role_name, PolicyName: policy_name,
+                }));
+            });
+
+            mocha.it('list roles by path prefix', async function() {
```
shirady: Can you explain this test? (What we had, what we do, and what is expected, etc.).

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2256,3 +2496,16 @@ function _get_err_code_and_message(err) {
...
+        RoleName: role_name_to_create,
+        AssumeRolePolicyDocument: assume_role_policy_document,
+    };
+    if (iam_path) input.Path = iam_path;
+    _check_status_code_ok(await iam_client.send(new CreateRoleCommand(input)));
+}
+
+async function delete_iam_role(iam_client, role_name_to_delete) {
+    _check_status_code_ok(await iam_client.send(new DeleteRoleCommand({ RoleName: role_name_to_delete })));
+}
```
shirady, line 2860: Could you add JSDoc to the new functions?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -517,6 +519,7 @@ mocha.describe('Session token tests', function() {
         const self = this; // eslint-disable-line no-invalid-this
         self.timeout(60000);
 
+        config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS = defualt_expiry_seconds * 1000;
```
shirady: Save the config exactly as it was so you can restore the config when you finish the test.

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +974,99 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+        await accounts[0].iam_client.send(new DeleteRolePolicyCommand({
+            RoleName: role_name, PolicyName: policy_name,
+        }));
+        await accounts[0].iam_client.send(new DeleteRoleCommand({ RoleName: role_name }));
+        for (const account of accounts) {
+            await rpc_client.account.delete_account({ email: account.email });
+        }
+    });
+
+    mocha.it('assumer listBuckets with GetObject-only role policy - should be denied', async function() {
```
shirady: What about the success cases? (For example: assumer listBuckets with list buckets-only role policy)

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +974,99 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+            `arn:aws:sts::${owner_id}:assumed-role/${role_name}/${session}`,
+            `${owner_id}:${session}`, owner_key, defualt_expiry_seconds);
+        const s3_client = generate_s3_client(
+            creds.access_key, creds.secret_key, coretest.get_https_address(), creds.session_token);
+        try {
+            await s3_client.listBuckets({});
+            assert.fail('expected AccessDenied');
+        } catch (err) {
+            assert.equal(err.Code || err.code, errors.s3_access_denied.code);
+        }
```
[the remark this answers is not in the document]
shirady: @aayushchouhan09 I think @alphaprinz started using it in the IAM user policy in the vector service test code, so you can use it as well.

src/test/system_tests/test_utils.js
```diff
@@ -8,6 +8,7 @@ const http = require('http');
 const https = require('https');
 const P = require('../../util/promise');
 const config = require('../../../config');
+const AWS = require('aws-sdk');
```
shirady: We want to avoid importing the former SDK version in files we already migrated.

src/test/unit_tests/api/iam/test_iam_ops_input_validation.test.js
shirady, line 1: Seems like redundant lines removals.

src/test/unit_tests/api/iam/test_iam_roles.test.js
```diff
@@ -0,0 +1,418 @@
...
+        const res = iam_utils.get_action_message_title(action);
+        expect(res).toBe(`iam:CreateRole`);
+    });
+
+    it('put_role_policy', () => {
+        const action = 'put_role_policy';
+        const res = iam_utils.get_action_message_title(action);
+        expect(res).toBe(`iam:PutRolePolicy`);
+    });
+});
```
shirady: This should be moved to `test_iam_utils.test.js`.
It is a basic unit test, and it doesn't use any mocks.

src/test/unit_tests/api/iam/test_iam_roles.test.js
```diff
@@ -0,0 +1,418 @@
...
+    it('create_arn_for_role with role_name and IAM_DEFAULT_PATH should return only role_name in arn', () => {
+        const res = iam_utils.create_arn_for_role(dummy_account_id, dummy_role_name, iam_constants.IAM_DEFAULT_PATH);
+        expect(res).toBe(`${arn_prefix}${dummy_account_id}:role/${dummy_role_name}`);
+    });
+
+    it('create_arn_for_role with role_name and iam_path should return them in arn', () => {
+        const res = iam_utils.create_arn_for_role(dummy_account_id, dummy_role_name, dummy_iam_path);
+        expect(res).toBe(`${arn_prefix}${dummy_account_id}:role${dummy_iam_path}${dummy_role_name}`);
+    });
+});
```
shirady: Same here - 
This should be moved to `test_iam_utils.test.js`.
It is a basic unit test, and it doesn't use any mocks.

src/test/unit_tests/api/iam/test_iam_roles.test.js
```diff
@@ -0,0 +1,418 @@
...
+
+describe('parse_role_arn', () => {
+    it('should parse account_id and role_name from a valid role ARN', () => {
+        const res = iam_utils.parse_role_arn('arn:aws:iam::123456789012:role/TestRole');
+        expect(res).toEqual({ account_id: '123456789012', role_name: 'TestRole' });
+    });
+
+    it('should return MISSING_ROLE_ARN when role_arn is empty', () => {
```
shirady:
```suggestion
    it('should throw MISSING_ROLE_ARN error when role_arn is empty', () => {
```

src/test/unit_tests/api/iam/test_iam_roles.test.js
```diff
@@ -0,0 +1,418 @@
...
+        expect(iam_utils.parse_role_arn('')).toEqual({ error: 'MISSING_ROLE_ARN' });
+    });
+
+    it('should return INVALID_ROLE_ARN when role_arn is malformed', () => {
```
shirady:
```suggestion
    it('should throw INVALID_ROLE_ARN error when role_arn is malformed', () => {
```

src/test/unit_tests/api/iam/test_iam_roles.test.js
```diff
@@ -0,0 +1,418 @@
...
+        expect(res).toBe(`${arn_prefix}${dummy_account_id}:role/${dummy_role_name}`);
+    });
+
+    it('create_arn_for_role with role_name and iam_path should return them in arn', () => {
+        const res = iam_utils.create_arn_for_role(dummy_account_id, dummy_role_name, dummy_iam_path);
+        expect(res).toBe(`${arn_prefix}${dummy_account_id}:role${dummy_iam_path}${dummy_role_name}`);
+    });
+});
+
+describe('parse_role_arn', () => {
```
shirady: Same here - 
This should be moved to `test_iam_utils.test.js`.
It is a basic unit test, and it doesn't use any mocks.

src/test/unit_tests/api/iam/test_iam_roles.test.js
```diff
@@ -0,0 +1,418 @@
...
+    const account_id = '123456789012';
+    const role_name = 'TestRole';
+    const role_arn = `arn:aws:iam::${account_id}:role/${role_name}`;
+
+    afterEach(() => {
+        system_store.get_instance.mockReset();
+        system_store.get_instance.mockReturnValue({ data: {} });
+    });
+
+    it('should return NO_SUCH_ROLE when role is not found', async () => {
```
shirady:
```suggestion
    it('should throw NO_SUCH_ROLE error when role is not found', async () => {
```

src/test/unit_tests/api/iam/test_iam_roles.test.js
```diff
@@ -0,0 +1,418 @@
...
+        system_store.get_instance.mockReturnValue({ data: { iam_roles_by_owner: {} } });
+        const res = await iam_utils.resolve_iam_role_by_arn(role_arn);
+        expect(res).toEqual({ error: 'NO_SUCH_ROLE', account_id, role_name });
+    });
+
+    it('should return iam_role when role exists', async () => {
+        const iam_role = { name: role_name };
```
shirady: A basic role is just a name? I think it must have a policy, right?

src/test/unit_tests/api/iam/test_iam_roles.test.js
```diff
@@ -0,0 +1,418 @@
...
+        system_store.get_instance.mockReturnValue({
+            data: { iam_roles_by_owner: { [account_id]: [iam_role] } },
+        });
+        const res = await iam_utils.resolve_iam_role_by_arn(role_arn);
+        expect(res).toEqual({ iam_role, account_id, role_name });
+    });
+});
+
+describe('input validation flow in IAM ops - IAM ROLES API', () => {
```
shirady: I think it can be simplified and more general when it is in `test_validation_utils.test.js` without a specific op.

src/test/unit_tests/api/iam/test_iam_roles.test.js
```diff
@@ -0,0 +1,418 @@
...
+        await iam_put_role_policy_op.handler(req({ role_name, policy_name, policy_document: identity_policy }), {});
+        await iam_get_role_policy_op.handler(req({ role_name, policy_name }), {});
+        await iam_list_role_policies_op.handler(req({ role_name }), {});
+        await iam_update_assume_role_policy_op.handler(req({ role_name, policy_document: assume_role_policy }), {});
+        await iam_delete_role_policy_op.handler(req({ role_name, policy_name }), {});
+        await iam_delete_role_op.handler(req({ role_name }), {});
+
+        for (const fn of Object.values(account_sdk)) {
+            expect(fn).toHaveBeenCalled();
+        }
```
shirady: What does this mean?
What will happen when we expand the `account_sdk` functions?

src/test/unit_tests/api/iam/test_iam_roles.test.js
shirady, line 1: Please reconsider the need for this new file; only create it if you think that the cases are not covered in the simple unit test cases and integration files that we already have.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
aayushchouhan09:
```
aayush:noobaa-core$ aws iam delete-role --role-name "$ROLE_NAME" 

aws: [ERROR]: An error occurred (DeleteConflict) when calling the DeleteRole operation: Cannot delete entity, must delete policies first.
```

src/test/integration_tests/api/iam/test_iam_basic_integration.js
aayushchouhan09: Creates a role under path /div_a/sub_div_b/, then lists roles with that PathPrefix. Expects only that role (not the default-path AdvRole from setup). Same idea as list users - by path prefix.

src/test/integration_tests/api/iam/test_iam_basic_integration.js
aayushchouhan09: Sure, Thanks!

src/test/integration_tests/api/iam/test_iam_basic_integration.js
aayushchouhan09: Done, Thanks

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: added comment for it.

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: added comment and tests.

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: Cleanup no longer swallows errors - DeleteRolePolicy / DeleteRole / delete_account are awaited directly.
If any of those fail, the suite fails and we notice it.

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: added. Thanks

src/test/integration_tests/api/sts/test_sts.js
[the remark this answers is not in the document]
aayushchouhan09: done, Thanks

src/test/system_tests/test_utils.js
aayushchouhan09: removed

src/test/system_tests/test_utils.js
aayushchouhan09: removed and followed the same pattern. Thanks

src/test/unit_tests/api/iam/test_iam_utils.test.js
aayushchouhan09: removed, Thanks.

src/test/unit_tests/api/iam/test_iam_ops_input_validation.test.js
aayushchouhan09: I think we can keep these changes. Thanks

src/test/unit_tests/api/iam/test_iam_roles.test.js
aayushchouhan09: We don't need this new file. Thanks

src/test/unit_tests/api/iam/test_iam_roles.test.js
aayushchouhan09: removed.

src/test/unit_tests/api/iam/test_iam_roles.test.js
aayushchouhan09: Not required.

src/test/system_tests/test_utils.js
```diff
@@ -462,16 +462,18 @@ function generate_anon_s3_client(endpoint) {
...
 
-function generate_s3_client(access_key, secret_key, endpoint) {
+function generate_s3_client(access_key, secret_key, endpoint, session_token) {
     return new S3({
         forcePathStyle: true,
         region: config.DEFAULT_REGION,
         requestHandler: new NodeHttpHandler({
-            httpAgent: new http.Agent({ keepAlive: false })
+            httpAgent: new http.Agent({ keepAlive: false }),
+            httpsAgent: new https.Agent({ keepAlive: false, rejectUnauthorized: false }),
```
shirady: Why do we need to have them both?
Do you know what does it mean to have them both?

src/test/unit_tests/api/iam/test_iam_roles.test.js
shirady: I still see this in the code.

src/test/unit_tests/api/iam/test_iam_roles.test.js
shirady: I still see this in the code.

src/test/unit_tests/api/iam/test_iam_roles.test.js
shirady: I think it is required in AWS docs when you create a role - [link](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateRole.html):

> AssumeRolePolicyDocument
> The trust relationship policy document that grants an entity permission to assume the role.
> ...
> Required: Yes

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -512,6 +514,11 @@ mocha.describe('Session token tests', function() {
     let sts_creds;
     const role_alice = 'role_alice';
     let account_info_alice;
+    const original_sts_expiry_ms = config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS;
+
+    mocha.afterEach(function() {
+        config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS = original_sts_expiry_ms;
+    });
```
shirady, line 484: Where exactly do you change the value of `config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS`?

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,434 @@ mocha.describe('IAM integration tests', async function() {
...
+                        assert.fail('list roles - requester is IAM user - should throw an error');
+                    } catch (err) {
+                        const { err_code, err_message } = _get_err_code_and_message(err);
+                        assert.equal(err_code, IamError.AccessDeniedException.code);
+                        assert.ok(err_message.includes('not authorized to perform: iam:ListRoles'));
+                    }
+                });
+
+                mocha.it('list roles - by path prefix', async function() {
+                    const path_role = 'PathPrefixedRole';
```
shirady: This is (`PathPrefixedRole`) the path? 
It is not a valid value; it should start and end with a slash.

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +978,147 @@ mocha.describe('Assume role with web indentity tests', function() {
         }).promise(), stsErr.AccessDeniedException.code, stsErr.AccessDeniedException.message);
     });
 });
+
+// Assumed-role IAM policy authorization:
+// - GetObject-only role policy -> listBuckets is denied
+// - ListAllMyBuckets-only role policy -> listBuckets succeeds
```
shirady: I don't think those comments are needed, as someone might expand the tests here.

src/test/integration_tests/api/sts/test_sts.js
shirady: Where is the comments?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -971,3 +978,147 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+                accessKeyId: access_key,
+                secretAccessKey: secret_key,
+            });
+            account.iam_client = generate_iam_client(access_key, secret_key, coretest.get_https_address_iam());
+        }
+
+        owner_account_info = await rpc_client.account.read_account({ email: owner_email });
+        assumer_account_info = await rpc_client.account.read_account({ email: assumer_email });
+        const assumer_account_id = assumer_account_info._id.toString();
+        await accounts[0].iam_client.send(new CreateRoleCommand({
```
shirady: I think it is hard to follow that you work on index 0 in the array of accounts.

src/test/system_tests/test_utils.js
aayushchouhan09: `NodeHttpHandler` uses httpAgent for `http://` and httpsAgent for `https://`. This helper is shared — most callers use HTTP; the new assumed-role checks use HTTPS (self-signed), so both are set. Older STS tests already did HTTPS via SDK v2 `sts_creds.httpOptions.agent`; here we reuse the v3 helper, which previously only had httpAgent.

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: Replaced `accounts[0]/accounts[1]` with named `owner/assumer` refs, `accounts` is only used for create/delete loops.

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: Done, Thanks

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: already there, expanded and moved to correct location now for proper explaination.

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: removed. Thanks

src/test/integration_tests/api/iam/test_iam_basic_integration.js
aayushchouhan09: It is the role name under the `role_path`, updated it to some good name. thanks

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: We set `config.STS_DEFAULT_SESSION_TOKEN_EXPIRY_MS = 0` in these two tests (non-NC only) so AssumeRole issues an already-expired session token:

1. user b assume role of user a - expiry 0 - list s3 - should be rejected
2. user b assume role of user a - expiry 0 - assume role sts - should be rejected

The `afterEach` restores `original_sts_expiry_ms` so later tests keep the normal default expiry.

src/test/unit_tests/api/iam/test_iam_roles.test.js
aayushchouhan09: Agreed, for CreateRole - `AssumeRolePolicyDocument` is required. That fixture was only a mock for `resolve_iam_role_by_arn` lookup by name (not CreateRole). Those `tests/test_iam_roles.test.js` were removed; `CreateRole coverage now uses a full trust policy in the integration/ops validation tests.

src/test/system_tests/test_utils.js
shirady: But when we use the client, do we know which agent we use?

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -373,6 +376,149 @@ mocha.describe('IAM integration tests', async function() {
             });
         });
 
+        // IAM Role APIs are containerized-only (AccountSpaceFS returns NotImplemented on NC)
```
shirady: Can you remove this comment or put it near `if (is_nc_coretest) this.skip();`?

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,435 @@ mocha.describe('IAM integration tests', async function() {
                 });
             });
         });
+
+        // IAM Role APIs are containerized-only (AccountSpaceFS returns NotImplemented on NC)
```
shirady: Can you remove this comment or put it near `if (is_nc_coretest) this.skip();`?

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -2129,6 +2275,435 @@ mocha.describe('IAM integration tests', async function() {
...
+                    _check_status_code_ok(response);
+                    assert.equal(response.Roles.length, 1);
+                    assert.equal(response.Roles[0].RoleName, role_under_path);
+                    assert.equal(response.Roles[0].Path, role_path);
+                    await delete_iam_role(iam_account, role_under_path);
+                });
+            });
+        });
+
+        // IAM Role Policy APIs are containerized-only (AccountSpaceFS returns NotImplemented on NC)
```
shirady: Can you remove this comment or put it near `if (is_nc_coretest) this.skip();`?

src/test/system_tests/test_utils.js
```diff
@@ -490,6 +493,20 @@ function generate_iam_client(access_key, secret_key, endpoint) {
...
 }
 
+function generate_sts_client(access_key, secret_key, endpoint, session_token) {
+    const httpsAgent = new https.Agent({ keepAlive: false, rejectUnauthorized: false });
+    return new STSClient({
+        region: config.DEFAULT_REGION,
+        endpoint,
+        credentials: {
+            accessKeyId: access_key || 'none',
+            secretAccessKey: secret_key || 'none',
```
shirady: What is this string `none`?

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -424,57 +407,43 @@ mocha.describe('STS tests', function() {
...
-async function assume_role_and_parse_xml(sts, params) {
-    const req = sts.assumeRole(params);
-    let json;
-    req.on('complete', async function(resp) {
-        json = await http_utils.parse_xml_to_js(resp.httpResponse.body);
-    });
-    await req.promise();
-    return json;
+async function assume_role_and_parse_xml(sts_client, params) {
+    return sts_client.send(new AssumeRoleCommand(params));
```
shirady: Where is the parse?
There was:

```
    let json;
    req.on('complete', async function(resp) {
        json = await http_utils.parse_xml_to_js(resp.httpResponse.body);
    });
```

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -424,57 +407,43 @@ mocha.describe('STS tests', function() {
...
-function validate_assume_role_response(json, expected_arn, expected_role_id, assumed_access_key, duration_seconds) {
-    dbg.log0('test.sts.validate_assume_role_response: ', json);
-    assert.ok(json && json.AssumeRoleResponse && json.AssumeRoleResponse.AssumeRoleResult);
-    const result = json.AssumeRoleResponse.AssumeRoleResult[0];
-    assert.ok(result);
-
-    // validate credentials
-    const credentials = result.Credentials[0];
-    assert.ok(credentials && credentials.AccessKeyId[0] && credentials.SecretAccessKey[0]);
+function validate_assume_role_response(response, expected_arn, expected_role_id, assumed_access_key, duration_seconds) {
```
shirady, line 424: Why did you remove access to index 0?
If you can explain the changes

shirady (approved):
Approving not to block you
Added a few questions

src/test/system_tests/test_utils.js
aayushchouhan09: NodeHttpHandler picks the agent from request.protocol: https: -> httpsAgent, otherwise httpAgent. So with coretest.get_https_address() we use httpsAgent; HTTP callers use httpAgent.

[see this](https://github.com/smithy-lang/smithy-typescript/blob/main/packages/node-http-handler/src/node-http-handler.ts#L184-L226)

src/test/system_tests/test_utils.js
aayushchouhan09: Updated to empty string.

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: after the v3 migrate this helper no longer parses XML (`send(AssumeRoleCommand)` returns a deserialized object). Renamed it to `assume_role`

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: With the old XML path, parse_xml_to_js put leaf values in arrays (`Credentials[0].AccessKeyId[0]`, etc.). AWS SDK v3 returns plain objects/strings (`response.Credentials.AccessKeyId`), so the `[0]` indexing is no longer needed.

src/test/system_tests/test_utils.js
shirady: But we mostly used the https endponit in our test, so why did you have to add the agent? Did you face any errors?

src/test/system_tests/test_utils.js
shirady: Why an empty string?

src/test/integration_tests/api/sts/test_sts.js
shirady: Do you need this function?
I think you can directly use the `AssumeRole`.

src/test/system_tests/test_utils.js
shirady: Since it is a helper function and you added a flag to the agent, I don't want it to change behavior in tests.

src/test/system_tests/test_utils.js
shirady: I mean, if you pass `undefined.`, you can leave it as-is?
The other S3 client didn't have this added.

aayushchouhan09: updated.

src/test/integration_tests/api/sts/test_sts.js
aayushchouhan09: removed and updated the calls.

src/test/system_tests/test_utils.js
aayushchouhan09: removed.

naveenpaul1 (approved):

src/test/integration_tests/api/sts/test_sts.js
```diff
@@ -936,38 +890,166 @@ mocha.describe('Assume role with web indentity tests', function() {
...
+        const creds = validate_assume_role_response(json,
+            `arn:aws:sts::${owner_id}:assumed-role/${role_name}/${session}`,
+            `${owner_id}:${session}`, owner_key, defualt_expiry_seconds);
+        const s3_client = generate_s3_client(
+            creds.access_key, creds.secret_key, coretest.get_http_address(), creds.session_token);
+        try {
+            await s3_client.listBuckets({});
+            assert.fail('assumer listBuckets with GetObject-only role policy - should throw an error');
+        } catch (err) {
+            assert.equal(err.Code || err.code || err.name, errors.s3_access_denied.code);
```
shirady: You can move this function to utils and use it here (and other places):
https://github.com/noobaa/noobaa-core/blob/ff1c75d4fe08427af2b0a8d892b6585a42e2dcd1/src/test/integration_tests/api/s3/test_deep_archive_via_s3.js#L40-L46

aayushchouhan09: Sure, Thanks

PR #9854, opened 2026-07-22
NC | Fixing md conditions for delete object/s

jackyalbo:
### Describe the Problem
We were missing support for md conditions in NC for delete operation to be better aligned with AWS: https://docs.aws.amazon.com/AmazonS3/latest/userguide/conditional-requests.html 

### Explain the Changes
1. Adding support for single/multiple deletes, both versioned and none-versioned

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `If-Match` behavior for single-object and bulk deletes, including correct handling for missing objects and `If-Match: *`.
  * Added support for per-object `ETag` to strengthen conditional matching in bulk deletes.
  * Disabled bulk-delete optimizations when conditional requirements are present, and improved per-object error reporting to preserve upstream RPC error codes.
  * Enhanced conditional evaluation for directory objects to behave consistently across empty vs non-empty cases.
* **Tests**
  * Expanded unit tests to cover conditional delete scenarios for single and bulk operations, including directory and versioning edge cases.

### DIFF
--- a/src/endpoint/s3/ops/s3_post_bucket_delete.js
+++ b/src/endpoint/s3/ops/s3_post_bucket_delete.js
@@ -30,8 +30,12 @@ async function post_bucket_delete(req) {
     for (const item of delete_list) {
         const key = item.Key?.[0];
         const version_id = item.VersionId?.[0];
-        const key_version = (key || '') + '\0' + (version_id || ''); // using null char (\x00) as separator
-        uniq_map.set(key_version, { key, version_id });
+        const etag = item.ETag?.[0];
+        const md_conditions = etag ? { if_match_etag: etag } : undefined;
+        const key_version = (key || '') + '\0' + (version_id || '') + '\0' + (etag || ''); // using null char (\x00) as separator
+        const obj = { key, version_id };
+        if (md_conditions) obj.md_conditions = md_conditions;
+        uniq_map.set(key_version, obj);
     }
     const objects = Array.from(uniq_map.values());
     dbg.log3('post_bucket_delete: objects without duplications', objects);
@@ -47,11 +51,12 @@ async function post_bucket_delete(req) {
         const req_obj = objects[i];
         const res_obj = reply[i];
         if (res_obj.err_code && !quiet) {
+            const mapped = S3Error.RPC_ERRORS_TO_S3[res_obj.err_code];
             results[i] = {
                 Error: {
                     Key: req_obj.key,
                     VersionId: req_obj.version_id,
-                    Code: res_obj.err_code,
+                    Code: mapped ? mapped.code : res_obj.err_code,
                     Message: res_obj.err_message,
                 }
             };
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -1005,15 +1005,7 @@ class NamespaceFS {
                     stat = await nb_native().fs.stat(fs_context, file_path);
                     isDir = native_fs_utils.isDirectory(stat);
                     if (isDir) {
-                        if (!stat.xattr?.[XATTR_DIR_CONTENT] || !params.key.endsWith('/')) {
-                            throw error_utils.new_error_code('ENOENT', 'NoSuchKey');
-                        } else if (stat.xattr?.[XATTR_DIR_CONTENT] !== '0') {
-                            // find dir object content file path  and return its stat + xattr of its parent directory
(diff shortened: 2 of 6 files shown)


src/endpoint/s3/ops/s3_post_bucket_delete.js
```diff
@@ -30,8 +30,10 @@ async function post_bucket_delete(req) {
     for (const item of delete_list) {
         const key = item.Key?.[0];
         const version_id = item.VersionId?.[0];
-        const key_version = (key || '') + '\0' + (version_id || ''); // using null char (\x00) as separator
-        uniq_map.set(key_version, { key, version_id });
+        const etag = item.ETag?.[0];
+        const md_conditions = etag ? { if_match_etag: etag } : undefined;
```
nadavMiz, line 34: we already have a function http_utils.get_md_conditions. why not use it here as well?

src/sdk/namespace_fs.js
```diff
@@ -2122,27 +2115,29 @@ class NamespaceFS {
             await this._load_bucket(params, fs_context);
             let res = [];
             if (this._is_versioning_disabled()) {
-                for (const { key, version } of params.objects) {
+                for (const { key, version, version_id, md_conditions } of params.objects) {
```
nadavMiz: first I am not sure why we need version_id here for md_conditions. this case is for disabled versioning.
second I can't find what `version` is. it looks like it was intended to skip delete with versions in cases where version id was supplied in disabled case. but we forgot to change it when changing it to version_id

src/server/object_services/object_server.js
```diff
@@ -1197,14 +1199,14 @@ async function delete_multiple_objects(req) {
                             bucket: req.bucket.name,
                             key: obj.key,
                             version_id: obj.version_id,
+                            md_conditions: obj.md_conditions,
```
nadavMiz, line 1202: is it needed? isn't it passed from req? we should probably add tests for this. its weird we don't have ones already. we would have recognized the pass of md_conditions got broken

nadavMiz: sorry this is the path for multiple delete. this wasn't implemented. just need tests

src/sdk/namespace_fs.js
```diff
@@ -3280,6 +3295,49 @@ class NamespaceFS {
...
+     * Stat the object targeted by delete and evaluate md_conditions.
+     * Stat/version resolution mirrors {@link read_object_md}.
+     * Missing objects (ENOENT, or latest delete marker) yield undefined metadata so
+     * If-Match fails while unconditional deletes stay idempotent.
+     * @param {import('../util/http_utils').MDConditions} md_conditions
+     * @param {import('./nb').NativeFSContext} fs_context
+     * @param {{ bucket?: string, key: string, version_id?: string }} params
+     * @param {string} [file_path] pre-resolved path from the caller, when already known
+     */
+    async _check_md_conditions_delete(md_conditions, fs_context, params, file_path) {
```
nadavMiz: I feel like a lot of the work here duplicates what we already do inside the individual delete functions. I think we can drop this helper and call check_md_conditions farther down in `_delete_single_object`, `_delete_single_object_versioned`, and `_delete_latest_version`, same place we already have the object md.

src/endpoint/s3/ops/s3_post_bucket_delete.js
jackyalbo: not the same structure - in single delete it can work, but in a delete-list like in this case the condition coming from the xml and not from the headers

src/sdk/namespace_fs.js
jackyalbo: saved the check for null_version like we do in object_server - removed version

src/server/object_services/object_server.js
jackyalbo: As we don't have md_conditions test for containerized - I will skip

src/sdk/namespace_fs.js
jackyalbo: Not completely - the issue with this function is its ability to get to the most relevant info for the file we want to delete, so we need to stat the file in cases we don't - like for normal deletes. And we need special handling in case it's a directory, and we need to get the info from somewhere else. Before this change we didn't really care about the etags while deleting

nadavMiz (approved):

PR #9855, opened 2026-07-23
map PutObjectLegalHold to the correct bucket policy action

kajalpareek-lab:
## Summary
- Map `put_object_legal_hold` to `s3:PutObjectLegalHold` (was incorrectly mapped to `s3:GetObjectLegalHold`).
- This allows PutBucketPolicy to accept `s3:PutObjectLegalHold` and makes runtime PutObjectLegalHold authorization check the correct permission.

`s3:BypassGovernanceRetention` PutBucketPolicy acceptance and runtime Bypass authorization will be handled together in a follow-up PR.

Fixes: [DFBUGS-8950](https://redhat.atlassian.net/browse/DFBUGS-8950)

## Test plan
- [x] PutBucketPolicy with `s3:PutObjectLegalHold` succeeds (no MalformedPolicy)
- [x] Unit: `test_access_policy_utils.test.js` accepts `s3:PutObjectLegalHold`
- [x] Integration: `test_s3_bucket_policy.js` — principal with only `s3:PutObjectLegalHold` can put legal hold; other principal is denied

### DIFF
--- a/src/test/integration_tests/api/s3/test_s3_bucket_policy.js
+++ b/src/test/integration_tests/api/s3/test_s3_bucket_policy.js
@@ -253,34 +253,57 @@ mocha.describe('s3_bucket_policy', function() {
     });
 
     // put_object_legal_hold was incorrectly mapped to GetObjectLegalHold, so
-    // PutBucketPolicy rejected s3:PutObjectLegalHold as an invalid action.
-    mocha.it('should accept PutObjectLegalHold in bucket policy', async function() {
+    // PutBucketPolicy rejected s3:PutObjectLegalHold and runtime auth checked the
+    // wrong permission. Grant only PutObjectLegalHold and exercise the op.
+    mocha.it('should allow PutObjectLegalHold when granted by bucket policy', async function() {
+        const LOCK_BKT = 'test2-bucket-policy-legal-hold';
+        const LOCK_KEY = 'legal-hold-obj';
+        await s3_owner.createBucket({
+            Bucket: LOCK_BKT,
+            ObjectLockEnabledForBucket: true,
+        });
+        await s3_owner.putObject({
+            Bucket: LOCK_BKT,
+            Key: LOCK_KEY,
+            Body: BODY,
+        });
+
         const policy = {
             Version: '2012-10-17',
             Statement: [{
                 Sid: 'id-put-object-legal-hold',
                 Effect: 'Allow',
                 Principal: { AWS: a_principal },
-                Action: [
-                    's3:PutObjectLegalHold',
-                    's3:GetObjectLegalHold',
-                ],
+                Action: ['s3:PutObjectLegalHold'],
                 Resource: [
-                    `arn:aws:s3:::${BKT}`,
-                    `arn:aws:s3:::${BKT}/*`,
+                    `arn:aws:s3:::${LOCK_BKT}`,
+                    `arn:aws:s3:::${LOCK_BKT}/*`,
                 ]
             }]
         };
         await s3_owner.putBucketPolicy({
-            Bucket: BKT,
+            Bucket: LOCK_BKT,
             Policy: JSON.stringify(policy)
         });
-        const res = await s3_owner.getBucketPolicy({ Bucket: BKT });
-        const stored = JSON.parse(res.Policy);
-        assert.deepStrictEqual(
-            stored.Statement[0].Action,
-            policy.Statement[0].Action
-        );
+
+        // Succeeds only if put_object_legal_hold maps to s3:PutObjectLegalHold.
+        // Wrong mapping to GetObjectLegalHold would AccessDeny with this policy.
+        await s3_a.putObjectLegalHold({
+            Bucket: LOCK_BKT,
+            Key: LOCK_KEY,
+            LegalHold: { Status: 'ON' },
+        });
+        const hold = await s3_owner.getObjectLegalHold({
+            Bucket: LOCK_BKT,
+            Key: LOCK_KEY,
+        });
+        assert.equal(hold.LegalHold.Status, 'ON');
+
+        await assert_throws_async(s3_b.putObjectLegalHold({
+            Bucket: LOCK_BKT,
+            Key: LOCK_KEY,
+            LegalHold: { Status: 'OFF' },
+        }));
     });
 
     mocha.it('should only read bucket policy when have permission to', async function() {
--- a/src/test/integration_tests/api/s3/test_s3_bucket_policy.js
+++ b/src/test/integration_tests/api/s3/test_s3_bucket_policy.js
@@ -252,6 +252,37 @@ mocha.describe('s3_bucket_policy', function() {
         }), 'Policy has invalid action');
     });
 
+    // put_object_legal_hold was incorrectly mapped to GetObjectLegalHold, so
+    // PutBucketPolicy rejected s3:PutObjectLegalHold as an invalid action.
+    mocha.it('should accept PutObjectLegalHold in bucket policy', async function() {
+        const policy = {
+            Version: '2012-10-17',
+            Statement: [{
+                Sid: 'id-put-object-legal-hold',
+                Effect: 'Allow',
+                Principal: { AWS: a_principal },
+                Action: [
+                    's3:PutObjectLegalHold',
+                    's3:GetObjectLegalHold',
+                ],
+                Resource: [
+                    `arn:aws:s3:::${BKT}`,
+                    `arn:aws:s3:::${BKT}/*`,
+                ]
+            }]
+        };
+        await s3_owner.putBucketPolicy({
+            Bucket: BKT,
+            Policy: JSON.stringify(policy)
+        });
+        const res = await s3_owner.getBucketPolicy({ Bucket: BKT });
+        const stored = JSON.parse(res.Policy);
+        assert.deepStrictEqual(
+            stored.Statement[0].Action,
+            policy.Statement[0].Action
+        );
+    });
+
     mocha.it('should only read bucket policy when have permission to', async function() {
         const policy = {
             Statement: [{
--- a/src/util/access_policy_utils.js
+++ b/src/util/access_policy_utils.js
@@ -94,16 +94,6 @@ const OP_NAME_TO_ACTION = Object.freeze({
     put_object: { regular: "s3:PutObject" },
 });
 
-/**
- * Extra s3 actions accepted in bucket policies that are not mapped 1:1 to an S3 op.
- * BypassGovernanceRetention is an additional permission required when using
- * x-amz-bypass-governance-retention (see AWS S3 Object Lock docs).
- */
-const BYPASS_GOVERNANCE_RETENTION_ACTION = 's3:BypassGovernanceRetention';
-const ADDITIONAL_BUCKET_POLICY_ACTIONS = Object.freeze([
-    BYPASS_GOVERNANCE_RETENTION_ACTION,
-]);
-
 const qm_regex = /\?/g;
 const ar_regex = /\*/g;
 const IAM_DEFAULT_PATH = '/';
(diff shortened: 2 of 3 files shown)


src/util/access_policy_utils.js
```diff
@@ -89,10 +89,20 @@ const OP_NAME_TO_ACTION = Object.freeze({
...
+    put_object_legal_hold: { regular: "s3:PutObjectLegalHold" },
     put_object: { regular: "s3:PutObject" },
 });
 
+/**
+ * Extra s3 actions accepted in bucket policies that are not mapped 1:1 to an S3 op.
+ * BypassGovernanceRetention is an additional permission required when using
+ * x-amz-bypass-governance-retention (see AWS S3 Object Lock docs).
+ */
+const BYPASS_GOVERNANCE_RETENTION_ACTION = 's3:BypassGovernanceRetention';
```
shirady: Why do you need this as const?
The rest of the actions were written in a string, and you use this action once.

src/util/access_policy_utils.js
```diff
@@ -546,7 +556,7 @@ async function validate_bucket_policy(policy, bucket_name, get_account_handler)
     return _validate_policy(policy, bucket_name, get_account_handler, {
         resource_arn_prefix: 'arn:aws:s3:::',
         action_wildcard: 's3:*',
-        valid_actions: all_op_names,
+        valid_actions: all_op_names.concat(ADDITIONAL_BUCKET_POLICY_ACTIONS),
```
shirady: I think the name of the variable is confusing `valid_actions` as now `s3:BypassGovernanceRetention` is not an action but a permission.

kajalpareek-lab: Fair point — it's an additional policy Action, not a mapped S3 op. I'll rename the helper constant/comment to make that clearer; valid_actions itself is the existing validator field for allowed Action strings.

src/util/access_policy_utils.js
kajalpareek-lab: Good catch — it's only used once here, so the const isn't needed. I'll inline 's3:BypassGovernanceRetention' (same style as the other actions) and drop the extra export.

src/util/access_policy_utils.js
```diff
@@ -543,10 +543,14 @@ async function _validate_policy(policy, bucket_name, get_account_handler, option
 
 async function validate_bucket_policy(policy, bucket_name, get_account_handler) {
     const all_op_names = _.flatten(_.compact(_.flatMap(OP_NAME_TO_ACTION, action => [action.regular, action.versioned])));
+    // Include Object Lock permissions that are not mapped 1:1 from an S3 op
+    // (BypassGovernanceRetention is required with the bypass header).
     return _validate_policy(policy, bucket_name, get_account_handler, {
         resource_arn_prefix: 'arn:aws:s3:::',
         action_wildcard: 's3:*',
-        valid_actions: all_op_names,
+        allowed_policy_actions: all_op_names.concat([
```
nadavMiz: I understand the reasoning of contacting this to this array.but I feel like we are missing nuance here by wiring it inside validate_bucket_policy. policy actions checks is something we check on multiple paths (e.g. IAM, vector bucket policy). we need to check that they don't share this table. I am mostly worried about IAM policy, as I think it should mostly be the same. its also hard to understand what exactly are the supported actions for someone who doesn't know where to look @shirady WDYT?

nadavMiz:
another thing that worries me is that `s3:BypassGovernanceRetention` is an action part of the policy. but it checks a flag, we have some paths for something like that with versioning but in that case we use a completely different permission. in our case we have two permissions. one for the action and one for the flag as an action. did you verify both are validated? 

can you add integration tests (`src/test/integration_tests/api/s3/test_s3_bucket_policy.js`) that it works end to end. I remember unit tests don't test some parts

src/util/access_policy_utils.js
shirady: From what I understand, the array is a list of permissions, not actions.

src/util/access_policy_utils.js
shirady: @nadavMiz in the IAM policy, there is no step of validation.
This PR is only about having the ability to put a bucket policy with this permission.
From the PR description - 

> Runtime Bypass authorization (IAM/bucket policy on NC and hosted, DeleteObjects wiring) will be handled in a follow-up PR.

And there we need to see that everything is well covered - bucket policy and IAM policy.

src/util/access_policy_utils.js
shirady, line 11: I think that the confusing name starts here - 
The key is the op name
The value is the permission(s).

I was not confusing as long as the action was the same as the permission.

cc: @nadavMiz

shirady: I think I was confused, marking as resolved.
No need for any change.

src/util/access_policy_utils.js
shirady: I dropped that idea, marking it as resolved.

shirady:
@kajalpareek-lab I'm adding a link that @nadavMiz sent:
https://docs.aws.amazon.com/service-authorization/latest/reference/list_s3.html

It is useful and will help with the next PR.

shirady:
@kajalpareek-lab I thought about it again, and I tend to think that in this PR fix ` put_object_legal_hold: { regular: "s3:PutObjectLegalHold" },` and in the later PR, handle the `s3:BypassGovernanceRetention` together with `Runtime Bypass authorization` so you can merge this one soon.

src/util/access_policy_utils.js
kajalpareek-lab: @nadavMiz Confirmed — this allowlist is only used by validate_bucket_policy (S3 PutBucketPolicy).

- Vector policies use a separate VECTOR_BUCKET_POLICY_ACTIONS list.
- IAM PutUserPolicy / PutRolePolicy validates document syntax/structure only; it does not reuse this S3 action allowlist.

Also, following Shira’s suggestion: I’m dropping s3:BypassGovernanceRetention from this PR and handling it with runtime Bypass authorization in #9881. This PR now only fixes the PutObjectLegalHold mapping.

kajalpareek-lab:
> another thing that worries me is that `s3:BypassGovernanceRetention` is an action part of the policy. but it checks a flag, we have some paths for something like that with versioning but in that case we use a completely different permission. in our case we have two permissions. one for the action and one for the flag as an action. did you verify both are validated?
> 
> can you add integration tests (`src/test/integration_tests/api/s3/test_s3_bucket_policy.js`) that it works end to end. I remember unit tests don't test some parts

You’re right that AWS requires both the operation permission and s3:BypassGovernanceRetention when the bypass header is used.
That dual-permission check is runtime behavior, so it belongs with the Bypass authorization work in #9881 — not in this PR.
Per Shira’s suggestion, this PR is scoped to PutObjectLegalHold only. I added an integration test in test_s3_bucket_policy.js that PutBucketPolicy accepts s3:PutObjectLegalHold. Bypass PutBucketPolicy acceptance + e2e auth coverage will be in #9881.

kajalpareek-lab:
> @kajalpareek-lab I thought about it again, and I tend to think that in this PR fix ` put_object_legal_hold: { regular: "s3:PutObjectLegalHold" },` and in the later PR, handle the `s3:BypassGovernanceRetention` together with `Runtime Bypass authorization` so you can merge this one soon.
Agreed — for completeness, not only merge speed.
put_object_legal_hold → s3:PutObjectLegalHold is a self-contained fix: PutBucketPolicy accepts that Action, and runtime PutObjectLegalHold checks the correct permission.
s3:BypassGovernanceRetention is different. Accepting it in PutBucketPolicy without runtime authorization would let us store a Bypass grant that still has no effect (hosted still uses admin role; NC still uses allow_bypass_governance). I’ll keep only the LegalHold mapping in this PR, and handle Bypass allowlist + runtime authorization together in #9881.

src/test/integration_tests/api/s3/test_s3_bucket_policy.js
```diff
@@ -252,6 +252,37 @@ mocha.describe('s3_bucket_policy', function() {
         }), 'Policy has invalid action');
     });
 
+    // put_object_legal_hold was incorrectly mapped to GetObjectLegalHold, so
+    // PutBucketPolicy rejected s3:PutObjectLegalHold as an invalid action.
+    mocha.it('should accept PutObjectLegalHold in bucket policy', async function() {
```
shirady: I think that the success expected test cases in this file usually test that the operation can be done (not just that the policy was stored as expected).

src/util/access_policy_utils.js
```diff
@@ -481,12 +481,12 @@ function _parse_condition_keys(condition_statement) {
  * @param {Object} options
  * @param {string} options.resource_arn_prefix - ARN prefix for resource validation (e.g. 'arn:aws:s3:::' or 'arn:aws:s3vectors:::')
  * @param {string} options.action_wildcard - wildcard action string (e.g. 's3:*' or 's3vectors:*')
- * @param {readonly string[]} options.valid_actions - list of valid action strings
+ * @param {readonly string[]} options.allowed_policy_actions - Action / permission strings allowed in the policy
```
shirady: You can drop the naming change `valid_actions` -> `allowed_policy_actions` as it was raised in the discussion regarding the bypass (and after I looked at the table in the link, I'm not sure it is needed).

src/test/integration_tests/api/s3/test_s3_bucket_policy.js
kajalpareek-lab: Updated. The test now creates an Object Lock bucket, grants only s3:PutObjectLegalHold to user A, and asserts A can putObjectLegalHold while B is denied. That covers PutBucketPolicy acceptance and the OP_NAME_TO_ACTION mapping (a wrong GetObjectLegalHold mapping would AccessDeny A with this policy).

src/util/access_policy_utils.js
kajalpareek-lab: Done — reverted to valid_actions. The rename was only from the earlier Bypass naming discussion and isn’t needed for the LegalHold fix.

src/test/integration_tests/api/s3/test_s3_bucket_policy.js
```diff
@@ -252,6 +252,60 @@ mocha.describe('s3_bucket_policy', function() {
         }), 'Policy has invalid action');
     });
 
+    // put_object_legal_hold was incorrectly mapped to GetObjectLegalHold, so
+    // PutBucketPolicy rejected s3:PutObjectLegalHold and runtime auth checked the
+    // wrong permission. Grant only PutObjectLegalHold and exercise the op.
```
shirady, line 257: I'm not sure you need this comment

src/test/integration_tests/api/s3/test_s3_bucket_policy.js
```diff
@@ -252,6 +252,60 @@ mocha.describe('s3_bucket_policy', function() {
...
+                ]
+            }]
+        };
+        await s3_owner.putBucketPolicy({
+            Bucket: LOCK_BKT,
+            Policy: JSON.stringify(policy)
+        });
+
+        // Succeeds only if put_object_legal_hold maps to s3:PutObjectLegalHold.
+        // Wrong mapping to GetObjectLegalHold would AccessDeny with this policy.
```
shirady, line 290: I think you can remove this comment as well.
It might in the future be more confusing than helpful for the developers to read in the future.

shirady (approved):
LGTM
Suggesting to remove the comments from the test to avoid confusions for the future readers.

nadavMiz:
@kajalpareek-lab @shirady I noticed in https://docs.aws.amazon.com/service-authorization/latest/reference/list_s3.html
that `s3:PutObjectLegalHold` is valid to other commands. I guess its the same idea as bypass governance permission. as you can add legal hold through other commands like put. I guess its out of scope for this PR, and maybe even out of scope for the opened bug. but its a gap we have

nadavMiz (approved):
LGTM, added comment out of scoped for this PR

PR #9868, opened 2026-07-26
skip Object Lock protected versions in lifecycle noncurrent expiration

kajalpareek-lab:
## Summary
- Lifecycle NoncurrentVersionExpiration no longer permanently deletes object versions under active retention or legal hold.
- Current-version Expiration still only creates delete markers (unchanged; Object Lock allows this).
- Shared lock check for version deletes; lifecycle never bypasses Governance retention.
## Test plan
- [ ] Unit: `test_object_server.test.js` (`_is_object_locked`)
- [ ] MDStore: `remove_noncurrent_versions` skips retention + legal hold
- [ ] Lifecycle integration: locked noncurrent versions remain after NoncurrentVersionExpiration
- [ ] Unlocked noncurrent versions still expire as before

## Summary by CodeRabbit

- **Bug Fixes**
  - Lifecycle expiration now preserves noncurrent versions protected by Object Lock legal holds or active retention periods.
  - Object deletions consistently honor COMPLIANCE and GOVERNANCE retention settings, including authorized governance bypasses.
  - Eligible, unlocked noncurrent versions continue to be removed as expected.
  - Retained versions are no longer soft-deleted prematurely during cleanup.
  - Object Lock failures now return a clear access-denied message with HTTP 403 status.

- **Documentation**
  - Clarified lifecycle behavior for delete markers, locked versions, and permanent noncurrent-version expiration.

### DIFF
--- a/src/server/bg_services/lifecycle.js
+++ b/src/server/bg_services/lifecycle.js
@@ -173,12 +173,12 @@ async function process_transition(system, bucket_info, rule, lifecycle_filter) {
 }
 
 async function delete_expired_objects(system, bucket, rule, reply_objects) {
-    /* 
-    Note that expired delete-markers are also deleted by this query
-    since the lack of filter_delete_markers means the function's internal
-    delete_marker variable remains as undefined, which is then dropped by compact()
-    effectively matching all object_md entries regardless of their delete_marker status
-    */
+    /*
+     * Versioned buckets: expire current versions by creating a delete marker
+     * (delete_version: false). Object Lock does not block delete-marker creation;
+     * locked object versions remain and are protected from permanent delete.
+     * Permanent NoncurrentVersionExpiration skips locked versions in MDStore.
+     */
     return await server_rpc.client.object.delete_multiple_objects_by_filter({
         bucket: bucket.name,
         create_time: get_expiration_timestamp(rule.expiration),
@@ -356,6 +356,8 @@ async function handle_bucket_rule(system, rule, j, bucket) {
         // According to https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configure-notification.html
         // it doesn't seem like deletion of noncurrent version should generate
         // any events.
+        // Object Lock: MDStore.remove_noncurrent_versions skips versions with
+        // active retention or legal hold (lifecycle never bypasses Governance).
         await delete_noncurrent_versions(
             system,
             bucket.name,
--- a/src/server/bg_services/lifecycle.js
+++ b/src/server/bg_services/lifecycle.js
@@ -24,12 +24,12 @@ function get_expiration_timestamp(expiration) {
 }
 
 async function delete_expired_objects(system, bucket, rule, reply_objects) {
-    /* 
-    Note that expired delete-markers are also deleted by this query
-    since the lack of filter_delete_markers means the function's internal
-    delete_marker variable remains as undefined, which is then dropped by compact()
-    effectively matching all object_md entries regardless of their delete_marker status
-    */
+    /*
+     * Versioned buckets: expire current versions by creating a delete marker
+     * (delete_version: false). Object Lock does not block delete-marker creation;
+     * locked object versions remain and are protected from permanent delete.
+     * Permanent NoncurrentVersionExpiration skips locked versions in MDStore.
+     */
     return await server_rpc.client.object.delete_multiple_objects_by_filter({
         bucket: bucket.name,
         create_time: get_expiration_timestamp(rule.expiration),
@@ -206,6 +206,8 @@ async function handle_bucket_rule(system, rule, j, bucket) {
         // According to https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configure-notification.html
         // it doesn't seem like deletion of noncurrent version should generate
         // any events.
+        // Object Lock: MDStore.remove_noncurrent_versions skips versions with
+        // active retention or legal hold (lifecycle never bypasses Governance).
         await delete_noncurrent_versions(
             system,
             bucket.name,
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -417,8 +417,10 @@ class MDStore {
         const sql_condition3 = size_less === undefined ? "" : `data->>'size' < ${size_less}`;
         const sql_condition4 = size_greater === undefined ? "" : `data->>'size' > ${size_greater}`;
         const sql_condition5 = tags && tags.length ? `ranked.tags @> '${JSON.stringify(tags)}'::jsonb` : "";
-        // Object Lock: lifecycle must not permanently delete versions under legal hold
-        // or active retention (Governance or Compliance). No governance bypass for lifecycle.
+        // Object Lock filter for this bulk lifecycle UPDATE (same SQL path that already
+        // filters by age/prefix/size/tags). Soft-delete only unlocked versions: skip
+        // legal hold ON and any retention whose retain_until_date is still in the future.
+        // Mode is not checked — any active retention date blocks delete. No governance bypass.
         const sql_condition_unlocked = `(
             (ranked.lock_settings IS NULL OR ranked.lock_settings = 'null'::jsonb)
             OR (
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -417,6 +417,19 @@ class MDStore {
         const sql_condition3 = size_less === undefined ? "" : `data->>'size' < ${size_less}`;
         const sql_condition4 = size_greater === undefined ? "" : `data->>'size' > ${size_greater}`;
         const sql_condition5 = tags && tags.length ? `ranked.tags @> '${JSON.stringify(tags)}'::jsonb` : "";
+        // Object Lock: lifecycle must not permanently delete versions under legal hold
+        // or active retention (Governance or Compliance). No governance bypass for lifecycle.
+        const sql_condition_unlocked = `(
+            (ranked.lock_settings IS NULL OR ranked.lock_settings = 'null'::jsonb)
+            OR (
+                (ranked.lock_settings->'legal_hold'->>'status' IS DISTINCT FROM 'ON')
+                AND (
+                    ranked.lock_settings->'retention' IS NULL
+                    OR ranked.lock_settings->'retention' = 'null'::jsonb
+                    OR (ranked.lock_settings->'retention'->>'retain_until_date')::timestamptz <= CURRENT_TIMESTAMP
+                )
+            )
+        )`;
 
         const sql_limit = limit === undefined ? "" : `LIMIT ${limit}`;
 
@@ -426,6 +439,7 @@ class MDStore {
                     _id,
                     (data->>'size')::BIGINT AS size,
                     data->'tagging' AS tags,
+                    data->'lock_settings' AS lock_settings,
                     ROW_NUMBER() OVER (
                         PARTITION BY data->>'key'
                         ORDER BY (data->>'version_seq')::BIGINT DESC
@@ -450,6 +464,7 @@ class MDStore {
                     ${sql_and_conditions(
                         sql_condition1, sql_condition2,
                         sql_condition3, sql_condition4, sql_condition5,
+                        sql_condition_unlocked,
                     )}
                 ${sql_limit}
             );`;
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -448,7 +448,7 @@ class MDStore {
                         PARTITION BY data->>'key'
                         ORDER BY (data->>'version_seq')::BIGINT
                     ) AS successor_time
-                FROM objectmds
+                FROM ${table_name}
                 WHERE
                     ${sql_and_conditions(
                         `data->>'bucket' = '${bucket_id}'`,
--- a/src/server/object_services/md_store.js
+++ b/src/server/object_services/md_store.js
@@ -471,6 +471,21 @@ class MDStore {
         const sql_condition3 = size_less === undefined ? "" : `data->>'size' < ${size_less}`;
         const sql_condition4 = size_greater === undefined ? "" : `data->>'size' > ${size_greater}`;
         const sql_condition5 = tags && tags.length ? `ranked.tags @> '${JSON.stringify(tags)}'::jsonb` : "";
+        // Object Lock filter for this bulk lifecycle UPDATE (same SQL path that already
+        // filters by age/prefix/size/tags). Soft-delete only unlocked versions: skip
+        // legal hold ON and any retention whose retain_until_date is still in the future.
+        // Mode is not checked — any active retention date blocks delete. No governance bypass.
+        const sql_condition_unlocked = `(
+            (ranked.lock_settings IS NULL OR ranked.lock_settings = 'null'::jsonb)
+            OR (
+                (ranked.lock_settings->'legal_hold'->>'status' IS DISTINCT FROM 'ON')
+                AND (
+                    ranked.lock_settings->'retention' IS NULL
+                    OR ranked.lock_settings->'retention' = 'null'::jsonb
+                    OR (ranked.lock_settings->'retention'->>'retain_until_date')::timestamptz <= CURRENT_TIMESTAMP
+                )
+            )
+        )`;
 
         const sql_limit = limit === undefined ? "" : `LIMIT ${limit}`;
 
@@ -480,6 +495,7 @@ class MDStore {
                     _id,
                     (data->>'size')::BIGINT AS size,
                     data->'tagging' AS tags,
+                    data->'lock_settings' AS lock_settings,
                     ROW_NUMBER() OVER (
                         PARTITION BY data->>'key'
                         ORDER BY (data->>'version_seq')::BIGINT DESC
@@ -488,7 +504,7 @@ class MDStore {
                         PARTITION BY data->>'key'
                         ORDER BY (data->>'version_seq')::BIGINT
                     ) AS successor_time
-                FROM objectmds
+                FROM ${table_name}
                 WHERE
                     ${sql_and_conditions(
                         `data->>'bucket' = '${bucket_id}'`,
@@ -504,6 +520,7 @@ class MDStore {
                     ${sql_and_conditions(
                         sql_condition1, sql_condition2,
                         sql_condition3, sql_condition4, sql_condition5,
+                        sql_condition_unlocked,
                     )}
                 ${sql_limit}
             );`;
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -2163,6 +2163,31 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
     }
 }
 
+/**
+ * True when Object Lock still protects the version from permanent delete.
+ * Lifecycle never bypasses Governance retention.
+ * @param {object} obj
+ * @param {{ bypass_governance?: boolean, now?: Date }} [options]
+ */
+function _is_object_locked(obj, options = {}) {
+    const lock = obj?.lock_settings;
+    if (!lock) return false;
+    if (lock.legal_hold?.status === 'ON') return true;
+    if (!lock.retention) return false;
+    const retain_until_date = new Date(lock.retention.retain_until_date);
+    if (!(retain_until_date > (options.now || new Date()))) return false;
+    if (lock.retention.mode === 'COMPLIANCE') return true;
+    if (lock.retention.mode === 'GOVERNANCE') return !options.bypass_governance;
+    return false;
+}
+
+function _throw_if_object_locked(obj, req) {
+    const bypass_governance = Boolean(req.rpc_params?.bypass_governance && req.role === 'admin');
+    if (!_is_object_locked(obj, { bypass_governance })) return;
+    dbg.error('object is locked, can not delete object', obj);
+    throw new RpcError('UNAUTHORIZED', 'can not delete locked object.');
+}
+
 async function _delete_object_version(req) {
     const { version_id } = req.rpc_params;
     const bucket_versioning = req.bucket.versioning;
@@ -2190,26 +2215,7 @@ async function _delete_object_version(req) {
         http_utils.check_md_conditions(req.rpc_params.md_conditions, obj);
         if (!obj) return { reply: {} };
 
-        if (obj.lock_settings) {
-            if (obj.lock_settings.legal_hold && obj.lock_settings.legal_hold.status === 'ON') {
-                dbg.error('object is locked, can not delete object', obj);
-                throw new RpcError('UNAUTHORIZED', 'can not delete locked object.');
-            }
-            if (obj.lock_settings.retention) {
-                const now = new Date();
-                const retain_until_date = new Date(obj.lock_settings.retention.retain_until_date);
-
-                if (obj.lock_settings.retention.mode === 'COMPLIANCE' && retain_until_date > now) {
-                    dbg.error('object is locked, can not delete object', obj);
-                    throw new RpcError('UNAUTHORIZED', 'can not delete locked object.');
-                }
-                if (obj.lock_settings.retention.mode === 'GOVERNANCE' &&
-                    (!req.rpc_params.bypass_governance || req.role !== 'admin') && retain_until_date > now) {
-                    dbg.error('object is locked, can not delete object', obj);
-                    throw new RpcError('UNAUTHORIZED', 'can not delete locked object.');
-                }
-            }
-        }
+        _throw_if_object_locked(obj, req);
         if (obj.version_past) {
             // 2, 8
             await MDStore.instance().delete_object_by_id(obj._id);
@@ -2512,6 +2518,8 @@ exports.calc_retention = calc_retention;
 
 if (process.env.NODE_ENV === 'test') {
     exports.__testing = {
-        update_bulk_delete_results
+        update_bulk_delete_results,
+        _is_object_locked,
+        _throw_if_object_locked,
     };
 }
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -357,7 +357,8 @@ async function put_object_retention(req) {
         if ((info.lock_settings.retention.mode === 'GOVERNANCE' && (!req.rpc_params.bypass_governance || req.role !== 'admin')) ||
             info.lock_settings.retention.mode === 'COMPLIANCE') {
             dbg.error('put object retention failed due object retention mode', obj);
-            throw new RpcError('UNAUTHORIZED');
+            throw new RpcError('OBJECT_LOCKED',
+                'Access Denied because object protected by object lock.');
         }
     }
     let legal_hold;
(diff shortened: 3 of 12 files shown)


src/server/bg_services/lifecycle.js
shirady, line 1: I see only comments in this file, is that intended?

src/server/object_services/md_store.js
```diff
@@ -434,7 +448,7 @@ class MDStore {
                         PARTITION BY data->>'key'
                         ORDER BY (data->>'version_seq')::BIGINT
                     ) AS successor_time
-                FROM objectmds
+                FROM ${table_name}
```
shirady, line 507: I think this change looks fine: `objectmds` -> `${table_name}`
@tangledbytes, I want to check with you

src/server/object_services/object_server.js
```diff
@@ -2139,6 +2139,31 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+ */
+function _is_object_locked(obj, options = {}) {
+    const lock = obj?.lock_settings;
+    if (!lock) return false;
+    if (lock.legal_hold?.status === 'ON') return true;
+    if (!lock.retention) return false;
+    const retain_until_date = new Date(lock.retention.retain_until_date);
+    if (!(retain_until_date > (options.now || new Date()))) return false;
+    if (lock.retention.mode === 'COMPLIANCE') return true;
+    if (lock.retention.mode === 'GOVERNANCE') return !options.bypass_governance;
```
shirady, line 2424: In the comments, it is written: `Lifecycle never bypasses Governance retention,` but here I see a case it uses the `options.bypass_governance` so I don't understand.

src/server/object_services/object_server.js
```diff
@@ -2139,6 +2139,31 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+    return false;
+}
+
+function _throw_if_object_locked(obj, req) {
+    const bypass_governance = Boolean(req.rpc_params?.bypass_governance && req.role === 'admin');
```
shirady, line 2430: Why do you need the `req.role === 'admin'`?

src/server/object_services/object_server.js
```diff
@@ -2139,6 +2139,31 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+    if (!_is_object_locked(obj, { bypass_governance })) return;
+    dbg.error('object is locked, can not delete object', obj);
+    throw new RpcError('UNAUTHORIZED', 'can not delete locked object.');
```
shirady: Will this error message get to the client?
Could you attach the steps that you took to check this

src/test/unit_tests/internal/test_object_server.test.js
```diff
@@ -543,3 +543,63 @@ describe('object_server - update_bulk_delete_results', () => {
...
+        expect(_is_object_locked({ lock_settings: {} })).toBe(false);
+    });
+
+    test('returns true for legal hold ON', () => {
+        expect(_is_object_locked({
+            lock_settings: { legal_hold: { status: 'ON' } },
+        })).toBe(true);
+    });
+
+    test('returns false for legal hold OFF', () => {
```
shirady, line 563: Can the legal hold be `undefined`?

src/server/object_services/object_server.js
kajalpareek-lab: As confirmed in [S3 open questions ](https://ibm.box.com/s/hnrhgg963bz3e2mwhmrrwpfmcmbjyhwf)admin will have  bypass permission and user who have iam policy and header passed in request.(Bypass IAM/bucket-policy work is a separate PR)

src/server/object_services/object_server.js
kajalpareek-lab: _is_object_locked is a shared helper. The bypass_governance option is only for normal S3 deletes (when the client sends the Bypass header).
Lifecycle does not use this helper at all. For NoncurrentVersionExpiration it goes through MDStore SQL, which always skips locked versions and has no Bypass path.
So lifecycle never bypasses Governance is still correct. The Bypass option in that helper is only for user S3 deletes, not for lifecycle.

src/server/bg_services/lifecycle.js
kajalpareek-lab: Yes, intended. The real skip is in MDStore.remove_noncurrent_versions (SQL). lifecycle.js only documents that current-version expiration creates a delete marker (Object Lock does not block that), while NoncurrentVersionExpiration permanently deletes and must skip locked versions in MDStore.

src/server/object_services/md_store.js
kajalpareek-lab: Yes. The UPDATE must use the same table as the rest of MDStore (this._objects.name), not a hard-coded name, so it stays correct if the table name is configured/overridden.

src/test/unit_tests/internal/test_object_server.test.js
kajalpareek-lab: Yes. Legal hold can be missing if it was never set — that case is already covered by the test with no / empty lock settings (we treat that as not locked). OFF is also a real state, after someone turns legal hold off with PutObjectLegalHold. So that separate test is still useful.

src/server/object_services/object_server.js
kajalpareek-lab: Yes for the error itself deleting a locked version fails with AccessDenied.
I have not run a live client check in this PR. From the code path, UNAUTHORIZED is mapped to AccessDenied. The detailed server message (can not delete locked object.) stays in logs; the S3 client should see the normal Access Denied message.

src/server/object_services/object_server.js
shirady: Could you add a comment about it?

src/server/object_services/object_server.js
shirady: I'm not sure that by checking `req.role === 'admin'` you will get the storage admin.
Why do you authorize in the `object_server` and not in the `s3_rest` layer?

src/server/object_services/md_store.js
```diff
@@ -417,6 +417,19 @@ class MDStore {
         const sql_condition3 = size_less === undefined ? "" : `data->>'size' < ${size_less}`;
         const sql_condition4 = size_greater === undefined ? "" : `data->>'size' > ${size_greater}`;
         const sql_condition5 = tags && tags.length ? `ranked.tags @> '${JSON.stringify(tags)}'::jsonb` : "";
+        // Object Lock: lifecycle must not permanently delete versions under legal hold
+        // or active retention (Governance or Compliance). No governance bypass for lifecycle.
+        const sql_condition_unlocked = `(
```
nadavMiz, line 478: can you explain this code? I don't think we should interact with db directly for object look at this point

src/server/object_services/object_server.js
```diff
@@ -2139,6 +2139,31 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+    const retain_until_date = new Date(lock.retention.retain_until_date);
+    if (!(retain_until_date > (options.now || new Date()))) return false;
+    if (lock.retention.mode === 'COMPLIANCE') return true;
+    if (lock.retention.mode === 'GOVERNANCE') return !options.bypass_governance;
+    return false;
+}
+
+function _throw_if_object_locked(obj, req) {
+    const bypass_governance = Boolean(req.rpc_params?.bypass_governance && req.role === 'admin');
+    if (!_is_object_locked(obj, { bypass_governance })) return;
```
nadavMiz: I feel like it would be clearer to do if (_is_object_locked(obj, { bypass_governance })) => throw
instead of  if (!_is_object_locked(obj, { bypass_governance })) => return, then throw for all other cases

src/server/object_services/object_server.js
```diff
@@ -2139,6 +2139,31 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+ * @param {object} obj
+ * @param {{ bypass_governance?: boolean, now?: Date }} [options]
+ */
+function _is_object_locked(obj, options = {}) {
+    const lock = obj?.lock_settings;
+    if (!lock) return false;
+    if (lock.legal_hold?.status === 'ON') return true;
+    if (!lock.retention) return false;
+    const retain_until_date = new Date(lock.retention.retain_until_date);
+    if (!(retain_until_date > (options.now || new Date()))) return false;
```
nadavMiz: again it would be clearer to do:
if (((options.now || new Date()) > retain_until_date))

src/server/object_services/object_server.js
```diff
@@ -2139,6 +2139,31 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+    if (lock.retention.mode === 'COMPLIANCE') return true;
+    if (lock.retention.mode === 'GOVERNANCE') return !options.bypass_governance;
+    return false;
```
nadavMiz: is it possible to reach this point?

nadavMiz:
please add tests to `test_s3_worm.js`

src/server/object_services/object_server.js
kajalpareek-lab: Today Bypass is allowed only if the header is set and req.role === 'admin', same as PutObjectRetention. s3_rest only forwards the header; allow/deny stays in object_server with the other Object Lock checks. Full IAM/bucket-policy Bypass is a separate PR (#9881). This PR’s main work is lifecycle skip of locked versions, not redesigning Bypass auth.

src/server/object_services/md_store.js
kajalpareek-lab: Noncurrent lifecycle delete was already a batch SQL update. Object Lock has to be in that same WHERE, or locked versions get deleted before JS ever sees them. Loading every candidate into object_server and calling _is_object_locked would change the lifecycle design and be much slower. S3 DeleteObject still uses the JS helpers lifecycle uses this SQL filter only.

kajalpareek-lab:
> please add tests to `test_s3_worm.js`

We already cover this in lifecycle/MDStore tests. I’ll also add a focused case in test_s3_worm.js so WORM suite shows lifecycle doesn’t permanently delete locked noncurrent versions.

src/server/object_services/object_server.js
kajalpareek-lab: With valid API data, no — mode is always GOVERNANCE or COMPLIANCE. It can only happen with bad/legacy metadata. Lifecycle SQL already treats any active retention date as locked. I’ll make the JS helper fail-closed the same way (return true after a future date), and note that in a comment.

src/server/object_services/object_server.js
kajalpareek-lab: I’ll rewrite for readability, but keep equal = expired/unlocked so we stay aligned with the SQL filter. Use >=, not >.

src/server/object_services/object_server.js
kajalpareek-lab: Agreed — positive check is clearer. I’ll flip it.

src/server/object_services/object_server.js
```diff
@@ -2163,6 +2163,42 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+ * options.bypass_governance is only for S3 DeleteObject / PutObjectRetention
+ * when the client sends x-amz-bypass-governance-retention. Lifecycle never
+ * calls this helper; NoncurrentVersionExpiration skips locked versions in
+ * MDStore SQL and has no governance bypass path.
+ *
+ * @param {object} obj
+ * @param {{ bypass_governance?: boolean, now?: Date }} [options]
+ */
+function _is_object_locked(obj, options = {}) {
+    const lock = obj?.lock_settings;
```
shirady: Why optional chaining?
`obj?.lock_settings`
According to the JSdoc of this function it is required and it is an object.

src/server/object_services/object_server.js
```diff
@@ -2163,6 +2163,42 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+    if (!lock) return false;
+    if (lock.legal_hold?.status === 'ON') return true;
+    if (!lock.retention) return false;
+    const retain_until_date = new Date(lock.retention.retain_until_date);
+    const now = options.now || new Date();
+    // Match MDStore SQL: retain_until_date <= now means retention expired (unlocked).
+    if (now >= retain_until_date) return false;
+    if (lock.retention.mode === 'COMPLIANCE') return true;
+    if (lock.retention.mode === 'GOVERNANCE') return !options.bypass_governance;
+    // Unrecognized/missing mode with a future retain-until: fail closed (same as SQL).
```
shirady: Are you sure the comment is in the right place?

src/server/object_services/object_server.js
shirady: I think there is a confusion between the `auth` where the `req.role === 'admin'` and the storage admin permissions you mentioned related to bucket policy.

src/test/integration_tests/db/test_md_store.js
```diff
@@ -801,6 +801,66 @@ mocha.describe('md_store', function() {
             return { chunks: [chunk], parts: [part], blocks: [block] };
         }
 
+        mocha.it('remove_noncurrent_versions skips Object Lock retention and legal hold', async function() {
+            if (config.DB_TYPE !== 'postgres') this.skip(); // eslint-disable-line no-invalid-this
+            const key = `ncv_lock_${Date.now().toString(36)}`;
```
shirady: Can you add a comment with what you meant?
```suggestion
            const key = `ncv_lock_${Date.now().toString(36)}`; // NCV = noncurrent versions
```

src/test/unit_tests/internal/test_object_server.test.js
```diff
@@ -543,3 +543,80 @@ describe('object_server - update_bulk_delete_results', () => {
...
+            },
+        })).toBe(true);
+    });
+
+    test('returns false for active GOVERNANCE retention with bypass', () => {
+        expect(_is_object_locked({
+            lock_settings: {
+                retention: { mode: 'GOVERNANCE', retain_until_date: future },
+            },
+        }, { bypass_governance: true })).toBe(false);
```
shirady, line 592: Do you know why the `bypass_governance` is saved as a property on the object?
I thought it was a permission at the account (IAM user policy) or bucket policy level to bypass and delete an object.

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -11,7 +11,12 @@ const http = require('http');
 const mocha = require('mocha');
 const assert = require('assert');
 const path = require('path');
+const moment = require('moment');
+const mongodb = require('mongodb');
 const fs_utils = require('../../../../util/fs_utils');
+const config = require('../../../../../config');
+const MDStore = require('../../../../server/object_services/md_store').MDStore;
```
shirady: Why did you have to import `MDStore`?

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -11,7 +11,12 @@ const http = require('http');
...
+const lifecycle = require('../../../../server/bg_services/lifecycle');
```
shirady: I think it would be more relevant to add it to a file related to the lifecycle.

src/server/object_services/object_server.js
shirady: I think you should test this in AWS.

From what I tested, I saw:
```
(AccessDenied) when calling the DeleteObject operation: Access Denied because object protected by object lock
```
It is not a message in the error object that we have:
https://github.com/noobaa/noobaa-core/blob/0ab16d92056ae34cd2e08030a0298ee8c6bae5a7/src/endpoint/s3/s3_errors.js#L52

shirady:
> please add tests to `test_s3_worm.js`

@nadavMiz I thought it would be more suitable to add them in the lifecycle test (not sure what exactly the files are).
@kajalpareek-lab added them in `test_s3_worm.js` as you asked.

nadavMiz:
> > please add tests to `test_s3_worm.js`
> 
> @nadavMiz I thought it would be more suitable to add them in the lifecycle test (not sure what exactly the files are). @kajalpareek-lab added them in `test_s3_worm.js` as you asked.

@shirady makes sense. I missed that this change was strictly for lifecycle when reviewing. its fine by me putting them in a lifecycle tests

src/server/object_services/md_store.js
```diff
@@ -417,6 +417,21 @@ class MDStore {
         const sql_condition3 = size_less === undefined ? "" : `data->>'size' < ${size_less}`;
         const sql_condition4 = size_greater === undefined ? "" : `data->>'size' > ${size_greater}`;
         const sql_condition5 = tags && tags.length ? `ranked.tags @> '${JSON.stringify(tags)}'::jsonb` : "";
+        // Object Lock filter for this bulk lifecycle UPDATE (same SQL path that already
+        // filters by age/prefix/size/tags). Soft-delete only unlocked versions: skip
+        // legal hold ON and any retention whose retain_until_date is still in the future.
+        // Mode is not checked — any active retention date blocks delete. No governance bypass.
+        const sql_condition_unlocked = `(
```
nadavMiz, line 478: I am not sure if its a good name. wouldn't it be better to use a condition like sql_condition_object_lock.

src/server/object_services/object_server.js
kajalpareek-lab: Yes  those are different. req.role === 'admin' is the NooBaa RPC/auth-token role, not IAM/bucket-policy storage admin. We’re not claiming it is. This PR keeps the same temporary Bypass gate as PutObjectRetention. Moving Bypass auth to s3_rest + s3:BypassGovernanceRetention is #9881, not this lifecycle change.

src/server/object_services/md_store.js
kajalpareek-lab: Prefer keeping sql_condition_unlocked. The predicate is true when the version may be deleted (not actively locked). sql_condition_object_lock reads like “has lock,” which is the opposite of what the WHERE selects.

src/server/object_services/object_server.js
kajalpareek-lab: Fair — obj is required. Dropped ?.

src/test/integration_tests/db/test_md_store.js
kajalpareek-lab: added.

src/server/object_services/object_server.js
kajalpareek-lab: Moved the Bypass note onto the options param so it isn’t mixed with the GOVERNANCE branch.

src/test/unit_tests/internal/test_object_server.test.js
kajalpareek-lab: It isn’t on the object. It’s the second argument to _is_object_locked — per-call options meaning “Bypass already allowed for this request.” Real Bypass auth (IAM/bucket policy + header) is not this helper’s job; that’s endpoint work (#9881). This unit test only checks the helper applies the flag.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: Only needed for the lifecycle case we added earlier (to age create_time). Removed that case and the MDStore import — lifecycle coverage stays in test_lifecycle.js.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: Agreed. Removed from here. Coverage is in test_lifecycle.js

src/test/integration_tests/api/s3/test_lifecycle.js
```diff
@@ -600,12 +600,66 @@ mocha.describe('lifecycle', function() {
...
+                        mode: 'COMPLIANCE',
+                        retain_until_date: moment().add(7, 'days').toDate(),
+                    },
+                },
+            });
+            await MDStore.instance().update_object_by_id(version_ids[1], {
+                lock_settings: {
+                    legal_hold: { status: 'ON' },
+                },
+            });
```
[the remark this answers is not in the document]
kajalpareek-lab: Lifecycle NoncurrentVersionExpiration filters in MDStore SQL and does not check retention mode — any future retain_until_date blocks delete, and there is no Bypass path. COMPLIANCE and GOVERNANCE take the same branch, so a separate GOVERNANCE case here wouldn’t cover new code. GOVERNANCE is already asserted against that SQL filter in test_md_store.js.

src/server/object_services/object_server.js
```diff
@@ -2334,6 +2334,41 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
 
+/**
+ * True when Object Lock still protects the version from permanent delete.
+ *
+ * @param {object} obj
+ * @param {{ bypass_governance?: boolean, now?: Date }} [options]
+ *   bypass_governance: caller already decided Bypass is allowed for this request
+ *   (S3 DeleteObject / PutObjectRetention with the Bypass header). Lifecycle never
+ *   calls this helper; NoncurrentVersionExpiration skips locked versions in MDStore
+ *   SQL and has no governance bypass path.
```
shirady, line 2410: Why is the comment here and before the parameters comments?

src/server/object_services/object_server.js
```diff
@@ -2334,6 +2334,41 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+ */
+function _is_object_locked(obj, options = {}) {
+    const lock = obj.lock_settings;
+    if (!lock) return false;
+    if (lock.legal_hold?.status === 'ON') return true;
+    if (!lock.retention) return false;
+    const retain_until_date = new Date(lock.retention.retain_until_date);
+    const now = options.now || new Date();
+    // Match MDStore SQL: retain_until_date <= now means retention expired (unlocked).
+    if (now >= retain_until_date) return false;
```
shirady: The condition and the comment are written in opposite directions; I think you can remove the comment.
It seems that the "=" overlaps, so better to verify what it should be - is it locked or not?

src/server/object_services/object_server.js
```diff
@@ -2334,6 +2334,41 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+    if (lock.retention.mode === 'COMPLIANCE') return true;
+    if (lock.retention.mode === 'GOVERNANCE') return !options.bypass_governance;
+    // Unrecognized/missing mode with a future retain-until: fail closed (same as SQL).
```
shirady: I see this comment:

> // Unrecognized/missing mode with a future retain-until: fail closed (same as SQL).

What is it related to?

nadavMiz (approved):

src/server/object_services/object_server.js
```diff
@@ -2334,6 +2334,41 @@ async function _put_object_handle_latest({ req, put_obj, set_updates, unset_upda
...
+    return true;
+}
+
+function _throw_if_object_locked(obj, req) {
+    // req.role === 'admin' is the NooBaa auth-token role (RPC), not IAM/bucket-policy
+    // "storage admin". Same temporary Bypass gate as PutObjectRetention until #9881.
```
shirady: I don't think those comments are related to the function `_throw_if_object_locked`, could you remove them?

src/server/object_services/object_server.js
shirady: So why do you still have the `&& req.role === 'admin'` in the condition; I didn't understand.

src/test/integration_tests/api/s3/test_lifecycle.js
```diff
@@ -600,12 +600,66 @@ mocha.describe('lifecycle', function() {
             assert.strictEqual(versions_list.objects.length, 2);
         });
 
+        // Object Lock + NoncurrentVersionExpiration: locked noncurrent versions must not be
+        // permanently deleted (coverage lives here, not in the Object Lock S3 suite).
```
shirady: I don't think you need this part: 
> (coverage lives here, not in the Object Lock S3 suite).

src/test/integration_tests/api/s3/test_lifecycle.js
```diff
@@ -600,12 +600,66 @@ mocha.describe('lifecycle', function() {
...
+                    legal_hold: { status: 'ON' },
+                },
+            });
+
+            await lifecycle.background_worker();
+
+            const versions_list = await rpc_client.object.list_object_versions({ bucket, prefix: key });
+            // Locked noncurrent versions kept; unlocked noncurrent removed; latest kept.
+            assert.strictEqual(versions_list.objects.length, 3);
+            const remaining_ids = new Set(versions_list.objects.map(o => String(o.obj_id)));
```
shirady: What is the `remaining_ids` I think the variable name is confusing.

src/server/object_services/object_server.js
kajalpareek-lab: We should keep this condition in this PR. Today Bypass is intentionally gated by RPC admin role + header, matching current PutObjectRetention behavior. Reworking it to full IAM/bucket-policy authorization is tracked separately in #9881. This PR is scoped to lifecycle lock-skip behavior, not auth-model changes.

src/test/integration_tests/api/s3/test_lifecycle.js
kajalpareek-lab: Agreed. Renamed to + "remaining_version_ids" + for clarity.

src/server/object_services/object_server.js
kajalpareek-lab: Agreed. I removed those _throw_if_object_locked comments.

src/test/integration_tests/api/s3/test_lifecycle.js
kajalpareek-lab: removed.

src/server/object_services/object_server.js
kajalpareek-lab: Unknown retention mode with active retention is treated as locked.

src/server/object_services/object_server.js
kajalpareek-lab: Agreed. I simplified the check to + "retain_until_date <= now" + for the unlocked path and removed the confusing directional comment.

src/server/object_services/object_server.js
kajalpareek-lab: removed the extra comment from the JSDoc parameter block to keep the function header clean.

src/server/object_services/object_server.js
shirady: Thanks, so can you change it to:
```suggestion
    // unknown/missing retention mode with a future retain-until (active retention)
```
Why do you need the part of "fail closed (same as SQL)"?

src/server/object_services/object_server.js
shirady: @kajalpareek-lab, can you document the gap or add the link for the following PR to fix the message?

src/server/object_services/object_server.js
kajalpareek-lab: updated. fail closed is design intent wording, not behavior description.

src/server/object_services/object_server.js
kajalpareek-lab: Thanks — you were right. Hosted was mapping OBJECT delete lock failures through UNAUTHORIZED → generic AccessDenied ("Access Denied"), so the AWS-specific message never reached the client.

shirady: @nadavMiz see the new error `AccessDeniedObjectLocked`; we probably need to use it in NC.

shirady (approved):
LGTM

PR #9869, opened 2026-07-27
Restrict bucket access based on client IP address (aws:SourceIP).

janfrode:
### Describe the Problem

Bucket policies should support adding restrictions based on client source IP (aws:SourceIP). F.ex. only allow S3 GET or S3 PUT from given IPs or IP ranges.

### Explain the Changes
1.  Implement support for aws:SourceIP in bucket policy conditions.
2.  Add unit test to make sure not to trust x-forwarded-for in this condition, since that can be spoofed.

### Testing Instructions:
1. Add bucket policy like the below, and test if it allows only the given client IPs:

```
{
  "Id": "SourceIP",
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "AllowTwoIPs",
      "Effect": "Allow",
      "Principal": "*",
      "Action": "s3:*",
      "Resource": "*"
    },
    {
      "Sid": "DenyOthers",
      "Effect": "Deny",
      "Principal": "*",
      "Action": "s3:*",
      "Resource": "*",
      "Condition": {
        "NotIpAddress": {
          "aws:SourceIp": [
            "11.11.11.11/32",
            "22.22.22.22/32"
          ]
        }
      }
    }
  ]
}
```

and/or run unit tests:

```
npx jest src/test/unit_tests/util_functions_tests/test_access_policy_utils.test.js --no-coverage
```

- [ ] Doc added/updated

none

- [ ] Tests added

```
  access_policy_utils
    validate_bucket_policy
      valid aws:SourceIp conditions
        ✓ should accept aws:SourceIp with IpAddress operator (1 ms)
        ✓ should accept aws:SourceIp with NotIpAddress operator
        ✓ should accept aws:SourceIp with an array of CIDR ranges
        ✓ should accept aws:SourceIp with an exact IPv4 address
        ✓ should accept aws:SourceIp with an IPv6 CIDR (2 ms)

    aws:SourceIp condition evaluation
      IpAddress operator — CIDR matching
        ✓ should ALLOW when client IP is in the allowed CIDR (7 ms)
        ✓ should IMPLICIT_DENY when client IP is outside the allowed CIDR
        ✓ should ALLOW when client IP matches one of multiple CIDRs
        ✓ should IMPLICIT_DENY when client IP matches none of the CIDRs (1 ms)
        ✓ should ALLOW when client IP matches an exact IP entry
      NotIpAddress operator
        ✓ should ALLOW when client IP is outside the excluded CIDR
        ✓ should IMPLICIT_DENY when client IP is inside the excluded CIDR (2 ms)
      Deny + IpAddress (block unless in range)
        ✓ should DENY when client IP is outside the allowed range (1 ms)
        ✓ should ALLOW when client IP is inside the allowed range
      IPv6-mapped IPv4 normalisation
        ✓ should ALLOW when client uses IPv6-mapped IPv4 address inside the CIDR (1 ms)
        ✓ should IMPLICIT_DENY when client uses IPv6-mapped IPv4 address outside the CIDR
      native IPv6
        ✓ should ALLOW when client IPv6 address is inside an IPv6 CIDR
        ✓ should IMPLICIT_DENY when client IPv6 address is outside the IPv6 CIDR
        ✓ should ALLOW exact IPv6 address match
        ✓ should IMPLICIT_DENY when IPv6 client is tested against an IPv4 CIDR
        ✓ should IMPLICIT_DENY when IPv4 client is tested against an IPv6 CIDR
        ✓ should ALLOW for NotIpAddress when IPv6 client is outside the excluded range
        ✓ should IMPLICIT_DENY for NotIpAddress when IPv6 client is inside the excluded range
      X-Forwarded-For spoofing resistance
        ✓ should IMPLICIT_DENY when client forges X-Forwarded-For to appear inside the CIDR but real TCP IP is outside
        ✓ should ALLOW when real TCP IP is inside the CIDR regardless of X-Forwarded-For
      missing client IP
        ✓ should IMPLICIT_DENY for Allow+IpAddress when socket has no remoteAddress
        ✓ should ALLOW for Allow+NotIpAddress when socket has no remoteAddress (1 ms)
        ✓ should DENY for Deny+NotIpAddress when socket has no remoteAddress
```

This intentionally does *not* trust x-forwarded-for header, since that's a client controlled header, and hence will not work when there is a HTTP/HTTPS proxy between client and noobaa. If we're to support this via proxy, we will need to implement a "trusted proxy" configuration before we can change the code to consider x-forwarded-for header.

## Summary by CodeRabbit

* **New Features**
  * Added support for `aws:SourceIp` conditions via `IpAddress` and `NotIpAddress`, supporting single values, CIDR ranges, and IPv4/IPv6 matching (including IPv4-mapped IPv6).
  * Bucket policy evaluation now applies source-IP checks using the server-observed client IP.

* **Bug Fixes**
  * Improved allow/deny behavior for matching and non-matching client IPs.
  * Ensures policies behave safely when the client IP is missing (implicit deny for `IpAddress`).

* **Tests**
  * Added unit and S3 integration coverage, including schema validation for these new operators.

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -361,6 +361,13 @@ module.exports = {
             }
         },
 
+        bucket_policy_ip_condition: {
+            type: 'object',
+            additionalProperties: {
+                $ref: '#/definitions/string_or_string_array'
+            }
+        },
+
         bucket_policy_null_condition: {
             type: 'object',
             additionalProperties: {
@@ -566,6 +573,15 @@ module.exports = {
                 },
                 Null: {
                     $ref: '#/definitions/bucket_policy_null_condition'
+                },
+                // IpAddress / NotIpAddress values are objects mapping a condition key (aws:SourceIp)
+                // to a single IP/CIDR string or an array of them. Per-value IP/CIDR format validation
+                // is performed in _validate_policy inside access_policy_utils.js.
+                IpAddress: {
+                    $ref: '#/definitions/bucket_policy_ip_condition'
+                },
+                NotIpAddress: {
+                    $ref: '#/definitions/bucket_policy_ip_condition'
                 }
             }
         },
--- a/src/util/access_policy_utils.js
+++ b/src/util/access_policy_utils.js
@@ -6,6 +6,7 @@ const dbg = require('./debug_module')(__filename);
 const s3_utils = require('../endpoint/s3/s3_utils');
 const RpcError = require('../rpc/rpc_error');
 const jwt = require('jsonwebtoken');
+const net_utils = require('./net_utils');
 
 const OP_NAME_TO_ACTION = Object.freeze({
     delete_bucket_analytics: { regular: "s3:PutAnalyticsConfiguration" },
@@ -111,13 +112,18 @@ const predicate_map = {
         return !value_regex.test(request_value);
     },
     'Null': function(request_value, policy_value) { return policy_value === 'true' ? request_value === null : request_value !== null; },
+    // IpAddress: true when the request IP matches any of the given CIDR ranges (or exact IPs).
+    'IpAddress': _ip_address_predicate,
+    // NotIpAddress: true when the request IP does NOT match any of the given CIDR ranges (or exact IPs).
+    'NotIpAddress': (request_value, policy_value) => !_ip_address_predicate(request_value, policy_value),
 };
 
 const condition_fit_functions = {
     's3:ExistingObjectTag': _is_object_tag_fit,
     's3:x-amz-server-side-encryption': _is_server_side_encryption_fit,
     's3:VersionId': _is_object_version_fit,
-    'aws:PrincipalTag': _is_aws_principal_tag_fit
+    'aws:PrincipalTag': _is_aws_principal_tag_fit,
+    'aws:SourceIp': _is_source_ip_fit,
 };
 
 const keycloak_predicate_map = {
@@ -135,10 +141,66 @@ const supported_actions = {
 
 // Condition keys that are principal/session-scoped rather than action-scoped —
 // they apply regardless of which S3 method is being called.
-const SKIPPED_CONDITIONS_ACTION = new Set(['aws:PrincipalTag']);
+const SKIPPED_CONDITIONS_ACTION = new Set(['aws:PrincipalTag', 'aws:SourceIp']);
 
 
-const SUPPORTED_BUCKET_POLICY_CONDITIONS = Object.keys(supported_actions);
+const SUPPORTED_BUCKET_POLICY_CONDITIONS = [...Object.keys(supported_actions), 'aws:SourceIp'];
+
+/**
+ * _ip_address_predicate returns true when the given IP matches the policy value.
+ * The policy value may be a single CIDR/IP string or an array of CIDR/IP strings.
+ * Matching logic mirrors AWS: CIDR ranges use subnet containment; bare IPs use exact equality
+ * after normalising IPv6-mapped IPv4 addresses (e.g. ::ffff:1.2.3.4 → 1.2.3.4).
+ *
+ * An absent or empty request_ip means the source address was not available on the
+ * request (e.g. internal or test path with no socket). An unknown IP cannot
+ * meaningfully match any range, so we always return false — callers that invert
+ * this result (NotIpAddress) will receive true, which is the agreed semantics:
+ * "unknown IP is treated as not being in any range".
+ *
+ * @param {string} request_ip - The client IP address, or '' if unavailable.
+ * @param {string|string[]} policy_value - One or more CIDR ranges / exact IPs from the policy.
+ * @returns {boolean}
+ */
+function _ip_address_predicate(request_ip, policy_value) {
+    if (!request_ip) return false;
+    const ip = net_utils.unwrap_ipv6(request_ip);
+    for (const entry of _.flatten([policy_value])) {
+        if (net_utils.is_cidr(entry)) {
+            if (net_utils.cidr_subnet_contains(entry, ip)) return true;
+        } else if (ip === net_utils.unwrap_ipv6(entry)) {
+            return true;
+        }
+    }
+    return false;
+}
+
+/**
+ * _is_source_ip_fit evaluates the aws:SourceIp condition key.
+ * Used by both IpAddress (allow when matches) and NotIpAddress (allow when does not match).
+ *
+ * X-Forwarded-For is intentionally ignored: it is fully client-controlled when
+ * there is no trusted proxy, and NooBaa has no trusted-proxy configuration.
+ * req.socket.remoteAddress is the only tamper-proof source of the client IP —
+ * it is set by the OS from the TCP handshake and cannot be forged.
+ *
+ * An absent remoteAddress is passed as '' to the predicate. _ip_address_predicate
+ * has an explicit null/empty check that returns false for a missing IP, making
+ * the "no IP" semantics clear at the predicate level:
+ *   - IpAddress    + no IP → predicate false → Allow skipped,  Deny skipped
+ *   - NotIpAddress + no IP → predicate true  → Allow granted,  Deny fires
+ *
+ * @param {Object} req - Incoming HTTP request.
+ * @param {Function} predicate - The operator predicate from predicate_map.
+ * @param {string|string[]} value - CIDR ranges or exact IPs from the policy condition.
+ * @returns {Promise<boolean>}
+ */
+async function _is_source_ip_fit(req, predicate, value) {
+    const client_ip = net_utils.unwrap_ipv6(req?.socket?.remoteAddress || '');
+    const res = predicate(client_ip, value);
+    dbg.log1('access_policy: source-ip fit?', value, client_ip, res);
+    return res;
+}
 
 async function _is_server_side_encryption_fit(req, predicate, value) {
     const encryption = s3_utils.parse_encryption(req);
@@ -474,6 +536,19 @@ function _parse_condition_keys(condition_statement) {
 }
 
 /**
+ * _validate_ip_condition_values checks that every entry in an IpAddress / NotIpAddress
+ * condition value is a valid IP address or CIDR range (IPv4 or IPv6).
+ * @param {string|string[]} condition_value
+ */
+function _validate_ip_condition_values(condition_value) {
+    for (const entry of _.flatten([condition_value])) {
+        if (!net_utils.is_cidr(entry) && !net_utils.is_ip(entry)) {
+            throw new RpcError('MALFORMED_POLICY', 'Policy has invalid IP or CIDR in condition', { detail: entry });
+        }
+    }
+}
+
+/**
  * _validate_policy is the shared validation logic for both S3 bucket policies and vector bucket policies.
  * @param {Object} policy - the policy document to validate
  * @param {string} bucket_name - the bucket name to validate resources against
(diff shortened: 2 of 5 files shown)


jackyalbo:
@janfrode is this suppose to support ipv6 as well? if so, can we add ipv6 test cases as well?
Adding @liranmauda and @nadavMiz as reviewers.

janfrode:
> @janfrode is this suppose to support ipv6 as well? if so, can we add ipv6 test cases as well? Adding @liranmauda and @nadavMiz as reviewers.

Yes, IPv6 should already be covered. 

Added IPv6 unit tests now. Will update list of tests in top comment ASAP.

src/util/access_policy_utils.js
```diff
@@ -135,10 +141,61 @@ const supported_actions = {
...
+ * @param {Function} predicate - The operator predicate from predicate_map.
+ * @param {string|string[]} value - CIDR ranges or exact IPs from the policy condition.
+ * @returns {Promise<boolean>}
+ */
+async function _is_source_ip_fit(req, predicate, value) {
+    const raw = req?.socket?.remoteAddress;
+    if (!raw) {
+        dbg.log1('access_policy: source-ip fit? no client IP available — denying');
+        return false;
+    }
```
liranmauda: @janfrode Doesn't the means that it flip the logic of the policy in NotIpAddress ?
what about returning it as an empty string?


```suggestion
    const raw = req?.socket?.remoteAddress || '';
```

In this case I think:
Allow + IpAddress + !raw will deny
Deny + IpAddress + !raw will skip
Allow + NotIpAddress + !raw will Allow
Deny + NotIpAddress + !raw will deny

liranmauda:
@janfrode Thanks for the PR.
I had 1 comment could you review it?
Regarding the commits, please squash them before merging. 
Also do not forget to rebase from master

src/util/access_policy_utils.js
janfrode: Coderabbit complained about:

>             req.connection is a deprecated alias for req.socket and is not
>             guaranteed on every request object reaching policy evaluation (HTTP/2
>             requests, destroyed/torn-down sockets, and internal callers that
>             construct a synthetic req). A missing connection throws a TypeError
>             inside authorization instead of producing a deny. Note the related
>             semantics: with an empty client_ip, NotIpAddress evaluates to true, so
>             an Allow statement would grant access when the source IP is unknown.


The intention was that if we don't have a client IP we should always fail -- or do you think it make sense to not fail in this case?

janfrode:
```
raw = ''   (falsy remoteAddress, fallback to '')

_ip_address_predicate('', '192.0.2.0/24')
  → is_cidr? yes → cidr_subnet_contains('192.0.2.0/24', '') → false
  → returns false

NotIpAddress predicate: !false  → true  → access granted
```

liranmauda: As Far as I can understand from https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html 
Th Important part in the top of the page states that if a key is empty or not present, then for a ky that includes Not (NotIpAddress) the condition should be true. 

@janfrode @jackyalbo @nadavMiz WDYT?

janfrode: The empty string matches nothing in _ip_address_predicate, so IpAddress would correctly deny, but NotIpAddress inverts that false to true. A request with no verifiable IP gets through. The explicit return false guard exists precisely to short-circuit before the predicate is ever called, making the outcome the same regardless of which operator is in use.

janfrode:
> @janfrode Thanks for the PR. I had 1 comment could you review it? Regarding the commits, please squash them before merging. Also do not forget to rebase from master

I _think_ your suggestion about just returning empty string is wrong. **not** **not** quickly gets confusing. So I left the code as is. Please have another think about it, and let me know if you still think it's wrong.

I've rebased, squashed and pushed again.

src/util/access_policy_utils.js
jackyalbo: It feels to me that we shouldn't skip calling the predicate - but will be happy hearing @nadavMiz on that.

liranmauda: We do not want IpAddress and NotIpAddress to behave the same.
We want:

In this case I think:
Allow + IpAddress + !raw will deny
Deny + IpAddress + !raw will skip
Allow + NotIpAddress + !raw will Allow
Deny + NotIpAddress + !raw will deny

liranmauda:
> > @janfrode Thanks for the PR. I had 1 comment could you review it? Regarding the commits, please squash them before merging. Also do not forget to rebase from master
> 
> I _think_ your suggestion about just returning empty string is wrong. **not** **not** quickly gets confusing. So I left the code as is. Please have another think about it, and let me know if you still think it's wrong.
> 
> I've rebased, squashed and pushed again.

https://github.com/noobaa/noobaa-core/pull/9869#discussion_r3658214127

@nadavMiz have a look at it make sure that AWS is not blocking in NotIpAdress case of missing ip.

Regarding `aws:SourceIp` it should not containg any empty strings and probably should be rejected buy PutBucketPolicy.
As far as I can see, the only valid entries for `aws:SourceIp` should be IP or CIDR and not empty strings.

src/util/access_policy_utils.js
janfrode: > We do not want IpAddress and NotIpAddress to behave the same. We want:
> 
> Allow + IpAddress + !raw will deny
>  Deny + IpAddress + !raw will skip 
> Allow + NotIpAddress + !raw will Allow 
> Deny + NotIpAddress + !raw will deny

Oh, I see -- fixing... Done.

nadavMiz:
@liranmauda @janfrode @jackyalbo 
new code looks better
hopefully I understand the problem
the way bucket policy works, false is always skip
so currently, after the change, the logic should be:
```
Allow + IpAddress + !raw will skip (eventually implicit deny)
Deny + IpAddress + !raw will skip (eventually implicit deny)
Allow + NotIpAddress + !raw will Allow
Deny + NotIpAddress + !raw will Deny
```

compared to before where everything skipped

so moving the logic to the predicate is correct behavior, its also cleaner if we would ever want to reuse this

nadavMiz:
another thing that bothers me:
we don't have schema validation for the new condition predicates. like:
https://github.com/noobaa/noobaa-core/blob/4ba7e9cbcf8acea53bebd652f5db606ef4e70179/src/api/common_api.js#L563-L570
this should not work. I think its missed because we don't have integration tests. can you add simple integration tests. just to verify end to end flow works?
see `src/test/integration_tests/api/s3/test_s3_bucket_policy.js`

janfrode:
> another thing that bothers me: we don't have schema validation for the new condition. like:
> 
> https://github.com/noobaa/noobaa-core/blob/4ba7e9cbcf8acea53bebd652f5db606ef4e70179/src/api/common_api.js#L563-L570
> 
> 
> this should not work. is this tested?

No, was not tested (except for unit tests). I don't currently have a lab environment for this..

Added schema validation.

janfrode:
> can you add simple integration tests. just to verify end to end flow works?

Done -- but still not tested with running noobaa...


```
$ node --check src/test/integration_tests/api/s3/test_s3_bucket_policy_source_ip.js && echo "syntax ok"
syntax ok
```

nadavMiz:
@jackyalbo @janfrode @liranmauda regarding aws handling, looking at aws documentation:

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html
```
If the key that you specify in a policy condition is not present in the request context, the values do not match and the condition is false. If the policy condition requires that the key is not matched, such as StringNotLike or ArnNotLike, and the right key is not present, the condition is true. This logic applies to all condition operators except [...IfExists](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html#Conditions_IfExists) and [Null check](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_condition_operators.html#Conditions_Null). These operators test whether the key is present (exists) in the request context.
``` 

so this looks correct, but I would prefer an explicit ip null / "" check inside the predicate so this is clear

janfrode:
> so this looks correct, but I would prefer an explicit ip null / "" check inside the predicate so this is clear

Ok, added an `if (!request_ip) return false;` to top of _ip_address_predicate.

janfrode:
@coderabbitai review

src/api/common_api.js
```diff
@@ -361,6 +361,16 @@ module.exports = {
             }
         },
 
+        bucket_policy_ip_condition: {
+            type: 'object',
+            additionalProperties: {
+                oneOf: [
+                    { type: 'string' },
+                    { type: 'array', items: { type: 'string' } }
```
nadavMiz: we should add a validation that we get the correct value, we need to check that its a valid ip / cid and not just a string. its best be done here. if its too problematic, we can add a validation helper function inside _validate_policy. but in that case we should use string_or_string_array as its the same as this object, and should add a comment that the validation is done in _validate_policy.

janfrode: Ok, added validation to _validate_policy.

nadavMiz:
@janfrode notice tests are failing because of lint issues:
```
> noobaa-core@5.23.0-874143c lint
> eslint src --quiet

/root/node_modules/noobaa-core/src/test/integration_tests/api/s3/test_s3_bucket_policy_source_ip.js
  29:5  error  'owner_principal' is assigned a value but never used  no-unused-vars

/root/node_modules/noobaa-core/src/util/access_policy_utils.js
  602:29  error  Blocks are nested too deeply (6). Maximum allowed is 5  max-depth

✖ 2 problems (2 errors, 0 warnings)
```

once those are resolved. tests will run on CI and we can verify pass

janfrode:
> @janfrode notice tests are failing because of lint issues:

Oh, thanks! I just saw some Postgres was failing, and thought it was unrelated. Will fix.

jackyalbo:
@janfrode it seems you still have lint errors here:
```
/root/node_modules/noobaa-core/src/test/integration_tests/api/s3/test_s3_bucket_policy_source_ip.js
  6:37  error  'is_nc_coretest' is assigned a value but never used  no-unused-vars
```

janfrode:
> @janfrode it seems you still have lint errors here:
> 
> ```
> /root/node_modules/noobaa-core/src/test/integration_tests/api/s3/test_s3_bucket_policy_source_ip.js
>   6:37  error  'is_nc_coretest' is assigned a value but never used  no-unused-vars
> ```

yes, saw it - -fixed.

jackyalbo:
Now tests are failing on this: ```
MalformedPolicy: Policy was not well formed or did not validate against the published schema```
for all the tests you added

src/test/utils/index/index.js
```diff
@@ -93,6 +92,7 @@ require('../../integration_tests/api/s3/test_s3_ops');
 require('../../integration_tests/api/s3/test_deep_archive_via_s3');
 require('../../integration_tests/api/s3/test_s3_encryption');
 require('../../integration_tests/api/s3/test_s3_bucket_policy');
+require('../../integration_tests/api/s3/test_s3_bucket_policy_source_ip');
```
jackyalbo: @janfrode can you move this test to be running only for NC:
src/test/utils/index/nc_index.js

janfrode: ok, done

jackyalbo: It seems like some extra changes are needed to adapt the test to NC - Will try to have a deeper dive on this tomorrow

src/test/utils/index/index.js
```diff
@@ -17,7 +17,6 @@ require('../../unit_tests/util_functions_tests/test_linked_list');
 require('../../unit_tests/util_functions_tests/test_keys_lock');
 require('../../unit_tests/util_functions_tests/test_lru');
 require('../../unit_tests/util_functions_tests/test_lru_cache');
-require('../../unit_tests/util_functions_tests/test_iam_role_cache');
```
jackyalbo, line 20: @janfrode was this removed by mistake?

janfrode: Wow, sorry, good catch! Guess my workflow is a bit weird. I copied src/test/utils/index/index.js from another tree that didn't have that line. Double checked my diff now, and think that must be the only thing I broke.

jackyalbo: ok, once we return this test and everything pass we can probably merge.

janfrode: Fixed.

jackyalbo (approved):

jackyalbo:
Thanks for the contribution, @janfrode  Appreciate you taking the time to add this. Merging now.

Issue #9866, opened 2026-07-26
IAM User Inline Policy With Bucket Policy


shirady:
### Environment info
- NooBaa Version: 4.23 (current master)
- Platform: not relevant - containerized (in NC, we don't have IAM user inline policy)

### Actual behavior
Currently, in S3 rest we ask for IAM policy AND bucket policy for all IAM users:
https://github.com/noobaa/noobaa-core/blob/ae08790e65602e7ce89ef09c352edb57ce13fb0a/src/endpoint/s3/s3_rest.js#L235-L238

### Expected behavior
- **Cross accounts** (IAM users in another account) - the current behavior is as expected.
- **Same account** (IAM users under the account) - having the permission only in one layer is enough.

From AWS docs ([link](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_principal.html#principal-anonymous)):

> For IAM users and role principals within your account, no other permissions are required. For principals in other accounts, they must also have identity-based permissions in their account that allow them to access your resource. This is called [cross-account access](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic-cross-account.html).

### Steps to reproduce
**Cross account case**
(should work as expected) - need bucket policy AND IAM user inline policy

**Same account case**
1. Create an IAM User and access key (no IAM user line permission).
2. Create a bucket in the account and add an object
3. Add a bucket policy on the principal of the user using ARN and a specific action
4. Try to use the IAM user with the specific action using AWS CLI: in AWS should succeed; in our current implementation, you would need to add an IAM inline policy for this.

3. ### More information - Screenshots / Logs / Other output
- Probably also impacts role inline policy
- Probably also impacts other services in NooBaa (Vectors, STS).

For example: in vector rest:
https://github.com/noobaa/noobaa-core/blob/ae08790e65602e7ce89ef09c352edb57ce13fb0a/src/endpoint/vector/vector_rest.js#L205-L206

PR #9879, opened 2026-07-27
nc - iam - add inline policy to account

alphaprinz:
### Describe the Problem
Implement iam policy CRUD for NC.
Note native_fs_utils.entity_enum.USER was used for native_fs_utils.translate_error_codes.

### Explain the Changes
1. Add new iam_user_policies field to account schema.
2. Implement put, get, list, update, delete user policy in accountspace_fs

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added full support for IAM user inline policies, including create/replace, retrieve, list, and delete operations.
  * Enforced aggregate inline policy size limits and improved policy document formatting.
  * Policy names are now listed in sorted order, with clear errors for policies that cannot be found.
* **Bug Fixes**
  * Blocked user deletion when inline policies remain, returning a conflict error.
* **Tests**
  * Expanded IAM integration coverage for inline policy scenarios.

### DIFF
--- a/src/sdk/accountspace_fs.js
+++ b/src/sdk/accountspace_fs.js
@@ -10,7 +10,7 @@ const { ConfigFS, CONFIG_TYPES } = require('./config_fs');
 const native_fs_utils = require('../util/native_fs_utils');
 const { create_arn_for_user, create_arn_for_root, get_action_message_title, check_iam_path_was_set } = require('../endpoint/iam/iam_utils');
 const { IAM_ACTIONS, MAX_NUMBER_OF_ACCESS_KEYS, IAM_DEFAULT_PATH,
-    ACCESS_KEY_STATUS_ENUM, IDENTITY_ENUM } = require('../endpoint/iam/iam_constants');
+    ACCESS_KEY_STATUS_ENUM, IDENTITY_ENUM, AWS_LIMIT_CHARS_INLINE_POLICY } = require('../endpoint/iam/iam_constants');
 const IamError = require('../endpoint/iam/iam_errors').IamError;
 const cloud_utils = require('../util/cloud_utils');
 const SensitiveString = require('../util/sensitive_string');
@@ -323,7 +323,7 @@ class AccountSpaceFS {
         dbg.log1(`AccountSpaceFS.${action}`, params);
         try {
             const requesting_account = account_sdk.requesting_account;
-            const requester = this._check_if_requesting_account_is_root_account_or_user_om_himself(action,
+            const requester = this._check_if_requesting_account_is_root_account_or_user_on_himself(action,
                 requesting_account, params.username);
             const username = params.username ?? requester.name;
             const on_itself = !params.username;
@@ -413,7 +413,7 @@ class AccountSpaceFS {
         try {
             const requesting_account = account_sdk.requesting_account;
             const access_key_id = params.access_key;
-            const requester = this._check_if_requesting_account_is_root_account_or_user_om_himself(action,
+            const requester = this._check_if_requesting_account_is_root_account_or_user_on_himself(action,
                 requesting_account, params.username);
             const username = params.username ?? requester.name; // username is not required
             await this._check_if_account_exists_by_access_key_symlink(action, access_key_id);
@@ -460,7 +460,7 @@ class AccountSpaceFS {
         try {
             const requesting_account = account_sdk.requesting_account;
             const access_key_id = params.access_key;
-            const requester = this._check_if_requesting_account_is_root_account_or_user_om_himself(action,
+            const requester = this._check_if_requesting_account_is_root_account_or_user_on_himself(action,
                 requesting_account, params.username);
             const username = params.username ?? requester.name; // username is not required
             await this._check_if_account_exists_by_access_key_symlink(action, access_key_id);
@@ -498,7 +498,7 @@ class AccountSpaceFS {
         dbg.log1(`AccountSpaceFS.${action}`, params);
         try {
             const requesting_account = account_sdk.requesting_account;
-            const requester = this._check_if_requesting_account_is_root_account_or_user_om_himself(action,
+            const requester = this._check_if_requesting_account_is_root_account_or_user_on_himself(action,
                 requesting_account, params.username);
             const username = params.username ?? requester.name;
             const on_itself = !params.username;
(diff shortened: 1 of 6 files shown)


src/sdk/accountspace_fs.js
```diff
@@ -91,6 +91,16 @@ function _clean_account_id_cache(requested_account) {
...
 
+/**
+ * Returns the index of the policy with policy_name in the iam_policies array, or -1 if not found.
+ * @param {Array} iam_policies
+ * @param {string} policy_name
+ * @returns {number}
+ */
+function _get_iam_policy_index(iam_policies, policy_name) {
+    return iam_policies.findIndex(current_iam_policy => current_iam_policy.policy_name === policy_name);
+}
```
shirady: Could you move new helper functions to the end of the file?

src/sdk/accountspace_fs.js
```diff
@@ -528,36 +538,132 @@ class AccountSpaceFS {
     // The function here are implemented in AccountSpaceNB, but not in AccountSpaceFS
     // and will throw NotImplemented error, except for the function with list which will return an empty list
 
+    // 1 - check that the requesting account is a root user account
```
shirady: Can you add a title for the implemented part?
like we have for:
```
    ////////////////
    // ACCESS KEY //
    ////////////////
```

src/sdk/accountspace_fs.js
```diff
@@ -528,36 +538,132 @@ class AccountSpaceFS {
...
+            } else {
+                iam_user_policies[index_of_iam_user_policy] = iam_user_policy_to_add;
+            }
+            this._check_total_policy_size(action, iam_user_policies, params.username);
+            requested_account.iam_user_policies = iam_user_policies;
+            await this.config_fs.update_account_config_file(requested_account);
+            _clean_account_id_cache(requested_account);
+        } catch (err) {
+            dbg.error(`AccountSpaceFS.${action} error`, err);
+            throw native_fs_utils.translate_error_codes(err, native_fs_utils.entity_enum.USER);
```
shirady, line 571: I'm not sure it should be `USER` (not only here), could you recheck this?

src/sdk/accountspace_fs.js
```diff
@@ -528,36 +538,132 @@ class AccountSpaceFS {
...
+        }
     }
 
+    // 1 - check that the requesting account is a root user account
```
shirady, line 637: I think we allowed the user to list its own resources in other APIs.
For example in `list_access_keys`:

>     // 1 - check that the requesting account is a root user account or that the username is same as the requester

src/sdk/accountspace_fs.js
alphaprinz: I have no preference, but note all other helper functions are before this new function. All of them are before the AccountspaceFS class.

src/sdk/accountspace_fs.js
alphaprinz: Since we're dealing with inline policy which is part of the user it makes sense to me. We do not a new fs entity. As long as it's just for native_fs_utils.translate_error_codes log, I don't mind adding new USER_POLICY 'native_entity'.

src/sdk/accountspace_fs.js
shirady: The function that are before are related to the cache and related operation to it.
The rest of the functions that are helpers are in the lower part of the class.

src/sdk/accountspace_fs.js
shirady: It is for the `translate_error_codes`, you need to test and check so you can decide if you need to add it.
Anyway, if you change, you need to check the mapping and the error message to the client in the end.

src/sdk/accountspace_fs.js
shirady: I checked that again and you raised about the test `list user policies - requester is IAM user - should throw an error` so keep as you did before.
Anyway, all those assumptions were before we had use policy - and in containerized the use inline policy was only for S3 and later for vectors.

src/sdk/accountspace_fs.js
alphaprinz: I went over the relevant code. Where there is policy-related error, an IamError with correct message is thrown, and iam_rest.handle_error() uses that error (ie no translation). For all the get/put/delete commands I've tested the cli result was the same.

shirady: I think worth adding a comment or a note in the PR description so the reader would understand that it was written as `USER` intentionally.

src/sdk/accountspace_fs.js
```diff
@@ -522,44 +522,147 @@ class AccountSpaceFS {
...
+            const owner_account_id = this._get_owner_account_argument(requesting_account);
+            const requested_account = await this.config_fs.get_account_or_user_by_name(params.username, owner_account_id);
+            this._check_if_requested_account_is_root_account_or_IAM_user(action, requesting_account, requested_account);
+            this._check_if_requested_is_owned_by_root_account(action, requesting_account, requested_account);
+            iam_user_policies = requested_account.iam_user_policies || [];
+        } catch (err) {
+            dbg.error(`AccountSpaceFS.${action} error`, err);
+            throw native_fs_utils.translate_error_codes(err, native_fs_utils.entity_enum.USER);
+        }
+        //_check_iam_policy_exists will throw IamError.NoSuchEntity if policy doesn't exist
```
shirady: Why do we need this comment? I think you can remove it
It is mentioned in the comments above
If you think the name can be improved so the comment will be removed, you can rename it.

src/sdk/accountspace_fs.js
```diff
@@ -522,44 +522,147 @@ class AccountSpaceFS {
         }
     }
 
-    /////////////////////
-    // OTHER FUNCTIONS //
-    /////////////////////
-    // The function here are implemented in AccountSpaceNB, but not in AccountSpaceFS
-    // and will throw NotImplemented error, except for the function with list which will return an empty list
+    /////////////////
+    // USER POLICY //
```
shirady, line 526: We had the `iam_operate_on_root_account` flag that was created as a solution for creating accounts (not users) instead of using the CLI.
Where do you block the APIs in this case? I didn't see any mention.

shirady:
@alphaprinz we have the doc related to IAM in NC - do you plan to update it in a separate PR?
https://github.com/noobaa/noobaa-core/blob/master/docs/design/iam_nc.md

What is the plan regarding the issue of users inside the account related to adding the user inline policy (#9866)?

src/sdk/accountspace_fs.js
alphaprinz: Sorry, not sure I follow. You mean blocking the user policy apis in iam_operate_on_root_account===true? If so, why? What would be the issue in that scenario?

src/sdk/accountspace_fs.js
alphaprinz: Updated the PR description.

src/sdk/accountspace_fs.js
shirady: I will refer you to our docs ([link](https://github.com/noobaa/noobaa-core/blob/master/docs/design/iam_nc.md)) - as background:

> In AWS root accounts are only created in the console.
> While in NooBaa, accounts can be created by -
> NooBaa CLI account add command.
> IAM API CreateUser operation. The requesting account must have the iam_operate_on_root_account property set to true. An account that has iam_operate_on_root_account property set to true, will operate on accounts instead of users when calling the IAM API, although it does not own them.

You asked:
> You mean blocking the user policy apis in iam_operate_on_root_account===true? If so, why? What would be the issue in that scenario?

just to make sure (add a test for this case in `src/test/integration_tests/api/iam/test_iam_basic_integration.js`):
The one account cannot add and IAM user inline policies to another **account** (yes, account not user).
We added the APIs for Accounts and Users to allow another way for creating accounts without the CLI (the first one must have this). once we have the new account, the other account does not control it.
I hope it's clear.

alphaprinz: Went by Policies and the root user in https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html and disalllowed user policy for root. This should include the iam_operate_on_root_account account. WDYT?

shirady: You can add a test for this.
I think it will not include it because the account which has the `iam_operate_on_root_account` is running operations on another account.
which mean that it would not fall in the condition you wrote `if (requesting_account.name.unwrap() === params.username) {`

shirady (approved):
LGTM
I added a comment about a GAP that we need in a follow-up PR (should be small).

src/test/integration_tests/api/iam/test_iam_basic_integration.js
```diff
@@ -1910,6 +1911,24 @@ mocha.describe('IAM integration tests', async function() {
                     }
                 });
 
+                mocha.it('put user policy - requested is root user - should throw an error', async function() {
+                    if (!is_nc_coretest) this.skip(); // eslint-disable-line no-invalid-this
```
shirady, line 1915: @coderabbitai can you open an issue for the skipped case?
I think in this PR it was added:

>             //root cannot put user policies on itself
>             //see "Policies and the root user" in https://docs.aws.amazon.com/IAM/latest/UserGuide/access_policies.html
>             if (requesting_account.name.unwrap() === params.username) {
>                 this._throw_access_denied_error(action, requesting_account, user_details, native_fs_utils.entity_enum.USER);
>             }

which is not covered in containerized.

PR #9881, opened 2026-07-28
authorize extra S3 actions from Object Lock headers

kajalpareek-lab:
## Summary
- Extra S3 actions are authorized at the endpoint from a shared header → permission map. The extra action is checked only when that header is present. No header means no extra check.
- Map:
  - `x-amz-bypass-governance-retention: true` → `s3:BypassGovernanceRetention`
  - `x-amz-object-lock-legal-hold` on PutObject / CreateMultipartUpload → `s3:PutObjectLegalHold`
  - `x-amz-object-lock-mode` / `x-amz-object-lock-retain-until-date` on PutObject / CreateMultipartUpload → `s3:PutObjectRetention`
- Same rules on hosted and NC: IAM (user or assumed-role policy) or bucket policy; explicit Deny wins; system admin and bucket owner do not need an extra grant.
- Dedicated PutObjectLegalHold / PutObjectRetention APIs skip the extra check because that action is already primary.
- Remove `allow_bypass_governance`. NamespaceFS trusts `bypass_governance` from the endpoint.
- Wire `bypass_governance` through DeleteObjects. Accept `s3:BypassGovernanceRetention` in PutBucketPolicy validation.

## Test plan
- [x] Unit: `test_s3_bypass_governance_permission.test.js` (Deny wins; extra check only when header is present; NC IAM Allow, implicit deny, bucket Allow, IAM Deny, owner without IAM, assumed-role IAM)
- [x] Unit: access policy utils accept `s3:BypassGovernanceRetention`
- [x] Integration: PutBucketPolicy accepts `s3:BypassGovernanceRetention` (`test_s3_bucket_policy.js`)
- [x] Integration: `test_s3_worm.js` NC bucket-owner Bypass without policy
- [x] Live OpenShift CI (`quay.io/kpareek1/noobaa-core:dev-20260826`): owner/admin extra actions are implicit Allow. IAM user with only `s3:PutObject`: no lock/Bypass header → allowed; header without grant → AccessDenied; bucket-policy Allow → succeeds; explicit Deny wins.
- [ ] Ceph NSFS / hosted: DeleteObjects cleanup with Bypass header succeeds for bucket owner


shirady:
@kajalpareek-lab, what does it mean in simple words (copied from the PR description):

> NC: no IAM user policies yet. Non-owners get Bypass via bucket policy only. Remove allow_bypass_governance so auth is not split across s3_rest and namespace_fs.

docs/NooBaaNonContainerized/AccountsAndBuckets.md
```diff
@@ -40,7 +40,7 @@ See all available account properties - [NC Account Schema](../../src/server/syst
 
   - `custom_bucket_path_allowed_list` - When an account creates a bucket using the S3 protocol, He can override the default bucket path location (under new_buckets_path) using `x-noobaa-custom-bucket-path` HTTP header. This directory will be created only if this path will be under one of the provided allowed list paths in custom_bucket_path_allowed_list. Must be a list of absolute paths (divided by colons). 
 
-  - `allow_bypass_governance` - Give permission to the user to bypass governance-mode retention object lock, allowing them to set the `x-amz-bypass-governance-retention` header for deleting the object and modifying retention lock. See https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lock.html#object-lock-retention-modes 
+  - Object Lock governance Bypass (`x-amz-bypass-governance-retention`) is granted with a bucket policy action `s3:BypassGovernanceRetention` (not an account flag). When IAM policies are available on NC, Bypass can also be controlled from IAM.
```
shirady: You can remove this only, you're inside a doc file explaining properties.

src/api/object_api.js
shirady, line 716: What about the API of `delete_object`? I see `delete_multiple_objects`

src/endpoint/iam/iam_utils.js
```diff
@@ -1306,14 +1306,37 @@ async function _get_identity_policies(account, is_iam_user, assumed_role_arn) {
...
+        account,
+        resource_arn,
+        // Assumed-role sessions: put the role ARN in AccessDenied via 'principal_arn' field
+        // without this, the message builds a user ARN from requesting_account
+        // (the role owner, loaded via assumed_role_access_key)
+        principal_arn: assumed_role_arn,
+        explicit_deny,
+    };
+}
+
```
shirady, line 1390: It is `iam_utils` file, this function is for bucket policy as well, could you check for another place that would be a better match?

src/endpoint/iam/iam_utils.js
```diff
@@ -1306,14 +1306,37 @@ async function _get_identity_policies(account, is_iam_user, assumed_role_arn) {
...
 
+/**
+ * Build an IAM deny result. Callers that combine IAM with bucket policy must treat
+ * explicit_deny as final (AWS: explicit Deny overrides any Allow).
+ * @param {object} account
+ * @param {string} resource_arn
+ * @param {string} [assumed_role_arn]
+ * @param {boolean} explicit_deny
+ */
+function _make_iam_deny_result(account, resource_arn, assumed_role_arn, explicit_deny) {
```
shirady, line 1379: Why do you need this function?
I don't see a reason for this one-line function.

src/endpoint/iam/iam_utils.js
```diff
@@ -1306,14 +1306,37 @@ async function _get_identity_policies(account, is_iam_user, assumed_role_arn) {
...
+    };
+}
+
 /**
  * authorize_request_iam_policy_impl evaluates IAM inline policies for IAM users and assumed-role sessions on the requested action/resource
- * returns true on allow, undefined when IAM policy auth is not applicable, or a deny context object
+ * returns:
+ * - true on allow
+ * - undefined when IAM policy auth is not applicable
+ * - deny context with explicit_deny true|false (Effect Deny vs no matching Allow)
```
shirady: It is not clear - what is "deny context with explicit_deny"?

src/endpoint/s3/s3_rest.js
```diff
@@ -14,9 +14,10 @@ const http_utils = require('../../util/http_utils');
 const signature_utils = require('../../util/signature_utils');
 const config = require('../../../config');
 const s3_utils = require('./s3_utils');
+const iam_utils = require('../iam/iam_utils'); // for IAM policy
 const { create_detailed_message_for_iam_user_access,
     get_owner_account_id,
-    authorize_request_iam_policy_impl } = require('../iam/iam_utils'); // for IAM policy
+    authorize_request_iam_policy_impl } = iam_utils;
```
shirady: Either import the file or destruct the file to what you use.

src/endpoint/s3/s3_rest.js
```diff
@@ -236,20 +237,25 @@ async function authorize_request(req) {
...
+    await authorize_bypass_governance_if_requested(req);
 }
 
 async function authorize_request_policy(req) {
     if (!req.params.bucket) return;
     if (req.op_name === 'put_bucket') return;
     // owner_account is { id: bucket.owner_account, email: bucket.bucket_owner };
+    const policy_info = await req.object_sdk.read_bucket_sdk_policy_info(req.params.bucket);
+    // Reused by bypass authorization for this request (avoid a second read).
+    req._bucket_sdk_policy_info = policy_info;
```
shirady, line 247: Why do you set a property in the request in the `s3_rest`?
What is planned for the other services?
I'm not sure it is a needed change.

src/endpoint/s3/s3_rest.js
```diff
@@ -361,6 +350,116 @@ async function authorize_request_iam_policy(req) {
...
+    dbg.error('authorize_bypass_governance_if_requested: AccessDenied for',
+        req.op_name, req.params.bucket, req.params.key);
+    throw new S3Error(S3Error.AccessDenied);
+}
+
+/**
+ * @param {nb.S3Request} req
+ * @returns {Promise<boolean>}
+ */
+async function _has_bypass_governance_permission(req) {
```
shirady: It seems like duplicate code. Can we avoid it?

src/endpoint/s3/s3_rest.js
```diff
@@ -361,6 +350,106 @@ async function authorize_request_iam_policy(req) {
...
+/**
+ * Containerized: bucket_claim_owner / owner id.
+ * NC: owner id, then name for backward compatibility.
+ */
+function _is_bucket_owner(account, bucket_name, { owner_account, bucket_owner, account_identifier_name }) {
+    if (account.bucket_claim_owner && account.bucket_claim_owner.unwrap() === bucket_name) return true;
+    if (owner_account && owner_account.id === account._id) return true;
+    if (account.owner === undefined && account_identifier_name === bucket_owner.unwrap()) return true;
+    return false;
+}
```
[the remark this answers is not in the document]
shirady: @kajalpareek-lab could you look at this comment

src/endpoint/s3/s3_rest.js
shirady, line 1: You added functions that are mutual to other services as well (vector, sts, etc).
Changing the main function of bucket policy should probably be done there as well.

src/endpoint/s3/s3_utils.js
```diff
@@ -444,6 +444,17 @@ function _is_valid_retention(mode, retain_until_date) {
     return true;
 }
 
+/**
+ * Client intent to bypass Governance retention (header only).
+ * Authorization is enforced earlier in s3_rest (IAM or bucket policy / owner overrides).
```
shirady: It is a util function; I don't see a reason to explain the general flow in the JSDoc.

src/sdk/namespace_fs.js
```diff
@@ -2372,22 +2373,18 @@ class NamespaceFS {
...
-    _has_bypass_governance_permission(fs_context) {
-        return fs_context?.allow_bypass_governance;
-    }
-
     /**
      * check if the object deletion should be blocked by retention lock. if the object is blocked will throw AccessDenied error
      * @param {Object} retention - object retention lock settings
-     * @param {boolean} bypass_governance - if true, and user has permission to use this flag, will allow to bypass governance mode retention lock. compliance mode retention lock cannot be bypassed.
+     * @param {boolean} bypass_governance - authorized Bypass intent from S3 endpoint
+     *   (hosted: IAM/bucket policy; NC: bucket policy). Namespace trusts that decision.
```
shirady: I don't think you need this comment as a part of the JSDoc

src/server/object_services/object_server.js
```diff
@@ -341,27 +345,25 @@ async function put_object_retention(req) {
     throw_if_maintenance(req);
     load_bucket(req);
     const obj = await find_object_md(req);
-    const info = get_object_info(obj, { role: req.role });
+    // Read lock settings from MD so non-admin callers (IAM/bucket-policy granted) can evaluate retention.
```
shirady: What is this comment?

src/server/object_services/object_server.js
```diff
@@ -341,27 +345,25 @@ async function put_object_retention(req) {
...
+    const current_lock = obj.lock_settings;
 
-    if (req.role !== 'admin') {
-        throw new RpcError('UNAUTHORIZED');
-    }
```
shirady, line 344: Why did you remove this?
```
   if (req.role !== 'admin') {
        throw new RpcError('UNAUTHORIZED');
    }
```

src/server/object_services/object_server.js
```diff
@@ -341,27 +345,25 @@ async function put_object_retention(req) {
...
     if (!req.bucket.object_lock_configuration || req.bucket.object_lock_configuration.object_lock_enabled !== 'Enabled') {
         throw new RpcError('INVALID_REQUEST');
     }
-    if (info.lock_settings && info.lock_settings.retention &&
-        (new Date(req.rpc_params.retention.retain_until_date) < new Date(info.lock_settings.retention.retain_until_date) ||
+    if (current_lock && current_lock.retention &&
+        (new Date(req.rpc_params.retention.retain_until_date) < new Date(current_lock.retention.retain_until_date) ||
             !req.rpc_params.retention)) {
 
-        if ((info.lock_settings.retention.mode === 'GOVERNANCE' && (!req.rpc_params.bypass_governance || req.role !== 'admin')) ||
```
shirady: Why did you remove this?
```
req.role !== 'admin')
```

src/server/object_services/object_server.js
```diff
@@ -1200,13 +1202,14 @@ async function delete_multiple_objects(req) {
...
                             md_conditions: obj.md_conditions,
+                            bypass_governance: req.rpc_params.bypass_governance,
                         }
                     }, req)
                 );
             } catch (err) {
                 dbg.error('Multiple delete for obj', obj, 'failed with reason', err);
                 res = {
-                    err_code: err.rpc_code || 'InternalError',
+                    err_code: err.rpc_code === 'UNAUTHORIZED' ? 'AccessDenied' : (err.rpc_code || 'InternalError'),
```
shirady: Seems like a redundant change - 
`UNAUTHORIZED` RPC error is mapped to `AccessDenied`.
https://github.com/noobaa/noobaa-core/blob/c386bc4dce4ceb7489617f29ba2e1041be42ef71/src/endpoint/s3/s3_errors.js#L631-L632

src/server/object_services/object_server.js
```diff
@@ -2182,7 +2185,7 @@ async function _delete_object_version(req) {
                     throw new RpcError('UNAUTHORIZED', 'can not delete locked object.');
                 }
                 if (obj.lock_settings.retention.mode === 'GOVERNANCE' &&
-                    (!req.rpc_params.bypass_governance || req.role !== 'admin') && retain_until_date > now) {
```
shirady: Why did you remove ` req.role !== 'admin'`?

src/server/object_services/object_server.js
```diff
@@ -2489,6 +2492,7 @@ exports.calc_retention = calc_retention;
 
 if (process.env.NODE_ENV === 'test') {
     exports.__testing = {
-        update_bulk_delete_results
+        update_bulk_delete_results,
+        _can_bypass_governance,
```
shirady: Where do you use this?

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -1080,33 +1081,35 @@ mocha.describe('s3 worm', function() {
...
+        mocha.it('should allow put retention with Bypass header as bucket owner without policy', async function() {
+            const res = await s3_owner.putObjectRetention({
                 Bucket: BKT,
                 Key: OBJ1,
                 Retention: { Mode: 'COMPLIANCE', RetainUntilDate: tomorrow },
                 VersionId: version_id,
                 BypassGovernanceRetention: true
-            }), 'AccessDenied', is_nc_coretest ? 'Access Denied because object protected by object lock.' : 'Access Denied');
+            });
+            delete res.$metadata;
```
shirady, line 1198: Why this was done? `delete res.$metadata`
We can have the generated metadata.

src/test/unit_tests/api/iam/test_iam_utils.test.js
```diff
@@ -4,6 +4,7 @@
 const iam_utils = require('../../../../endpoint/iam/iam_utils');
 const iam_constants = require('../../../../endpoint/iam/iam_constants');
 const { IamError } = require('../../../../endpoint/iam/iam_errors');
+const access_policy_utils = require('../../../../util/access_policy_utils');
```
shirady: Why do you import `const access_policy_utils`?

src/test/unit_tests/api/iam/test_iam_utils.test.js
```diff
@@ -873,3 +874,47 @@ describe('validate_user_input_iam', () => {
         });
     });
 });
+
+describe('authorize_request_iam_policy_impl deny kinds', () => {
```
shirady: The file should be with simple unit tests for the helper functions

src/test/unit_tests/internal/test_object_server.test.js
```diff
@@ -543,3 +543,13 @@ describe('object_server - update_bulk_delete_results', () => {
         expect(results[3]).toHaveProperty('seq', 103);
     });
 });
+
+describe('object_server._can_bypass_governance', () => {
```
shirady: Do you see a value in this test?

src/test/unit_tests/util_functions_tests/test_access_policy_utils.test.js
shirady, line 1: I think a test in `src/test/integration_tests/api/s3/test_s3_bucket_policy.js` is enough.

src/test/unit_tests/endpoint/test_s3_bypass_governance_auth.test.js
shirady, line 1: Why do we need this new file?
Are the existing cases not already covered in the other test files?

shirady:
@nadavMiz there are no systems that run this code, right?
Removing the property of `allow_bypass_governance` from the `src/server/system_services/schemas/nsfs_account_schema.js` is fine, or would we need an upgrade? To avoid creating the existing objects as an invalid schema with an additional property of `allow_bypass_governance`.

nadavMiz:
@shirady I don't think there are any systems using it today, scale are only now testing this, I think its fine. that said, @jackyalbo you'll know better then I would, WDYT?

jackyalbo:
I am not aware of any system using it; it's just now being tested by the QE. I will validate with Scale anyway and will update in case we need one.

src/endpoint/s3/s3_rest.js
```diff
@@ -384,6 +483,11 @@ async function authorize_anonymous_access(s3_policy, method, arn_path, req, publ
     throw new S3Error(S3Error.AccessDenied);
 }
 
+function _is_bypass_governance_requested(req) {
```
nadavMiz: you have the same function in both s3_rest and s3_utils. why not just use the s3_utils one?

src/endpoint/s3/s3_rest.js
```diff
@@ -265,10 +271,6 @@ async function authorize_request_policy(req) {
     const account = req.object_sdk.requesting_account;
     const is_nc_deployment = Boolean(req.object_sdk.nsfs_config_root);
     const account_identifier_name = is_nc_deployment ? account.name.unwrap() : account.email.unwrap();
-    // Both NSFS NC and containerized will validate bucket policy against account id
```
nadavMiz, line 268: why did you remove this

src/endpoint/s3/s3_rest.js
```diff
@@ -361,6 +350,116 @@ async function authorize_request_iam_policy(req) {
     );
 }
 
+/**
+ * Require s3:BypassGovernanceRetention when the bypass header is set.
```
nadavMiz: might be personal but this comment doesn't make any sense to me. what are you trying to say here?

nadavMiz:
@kajalpareek-lab @shirady I feel like this design is short sighted. we need a general mechanism to check for permission requiring flags. and to have a map between flags and permissions. so it would be easy to add such permissions to bucket policy / IAM. and validate that it behaves logically correct with the other permissions. I think we should have a small design discussion about this. WDYT?

kajalpareek-lab:
> @kajalpareek-lab, what does it mean in simple words (copied from the PR description):
> 
> > NC: no IAM user policies yet. Non-owners get Bypass via bucket policy only. Remove allow_bypass_governance so auth is not split across s3_rest and namespace_fs.

@shirady On NC today, IAM users cannot get a real user policy for Bypass (PutUserPolicy is NotImplemented). So a non-owner only gets Bypass if the bucket policy allows s3:BypassGovernanceRetention. We removed the old account flag allow_bypass_governance so we don’t check Bypass in two places (s3_rest and namespace_fs). System admin and bucket owner still get Bypass without a policy.

kajalpareek-lab:
> @kajalpareek-lab @shirady I feel like this design is short sighted. we need a general mechanism to check for permission requiring flags. and to have a map between flags and permissions. so it would be easy to add such permissions to bucket policy / IAM. and validate that it behaves logically correct with the other permissions. I think we should have a small design discussion about this. WDYT?

@nadavMiz Agree this Bypass path is a one-off on top of the existing IAM + bucket-policy checks. Happy to do a short design discussion for a reusable “header/flag → extra action” map. This PR still closes the Object Lock Bypass auth gap; we can refactor onto a shared mechanism after that.

docs/NooBaaNonContainerized/AccountsAndBuckets.md
kajalpareek-lab: You’re right — that section lists account properties. I’ll delete the bullet instead of rewriting it. Bypass belongs in CLI/policy docs, not in the account property list.

src/api/object_api.js
kajalpareek-lab: delete_object already has bypass_governance in the API. This PR only adds it to delete_multiple_objects so DeleteObjects can pass the Bypass flag into each delete.

src/endpoint/iam/iam_utils.js
kajalpareek-lab: It’s small on purpose. Bypass needs explicit_deny so an IAM Deny is not overridden by a bucket-policy Allow. Without this shape, we’d only get a boolean and lose that distinction.

src/endpoint/iam/iam_utils.js
kajalpareek-lab: authorize_request_iam_policy_impl only evaluates IAM. Bucket-policy Bypass stays in s3_rest. Putting the combine step in iam_utils would mix layers. Best home for shared pieces later is something like access_policy_utils / a shared auth helper so keeping bucket policy into iam_utils now.

src/endpoint/iam/iam_utils.js
kajalpareek-lab: When IAM does not Allow, we return an object (not just false) with explicit_deny:
true = policy had Effect Deny → stop; bucket policy cannot Allow Bypass
false = no matching Allow → bucket policy may still Allow Bypass

src/endpoint/s3/s3_rest.js
kajalpareek-lab: Bypass is S3 Object Lock only for now. Helpers like _is_bucket_owner were extracted inside s3_rest for reuse by Bypass + bucket policy. Vector/STS don’t need Bypass yet. If we generalize, we should lift shared owner/identifier helpers — not block this PR on that.

src/endpoint/s3/s3_rest.js
kajalpareek-lab: Agreed — redundant. I’ll keep iam_utils (needed for authorize_request_iam_policy_impl in Bypass) and drop the unused destructured names, or only destructure what we use.

src/endpoint/s3/s3_rest.js
kajalpareek-lab: Only a request-scoped cache to avoid a second bucket-policy read when Bypass runs after normal policy auth. Not a new public API. Other services don’t need it until they grow a similar second check.

src/endpoint/s3/s3_rest.js
kajalpareek-lab: It looks similar because Bypass reuses the same decision order (owner → IAM → bucket policy → root). The action and Deny/Allow combine rules differ, so it isn’t a straight copy-paste of the op permission. We can extract shared “evaluate IAM + bucket policy for action X” later; for this PR I’d rather keep Bypass logic readable in one place.

src/endpoint/s3/s3_rest.js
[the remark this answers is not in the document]
kajalpareek-lab: Valid. I’ll guard like system owner: only unwrap when bucket_owner is present.

src/endpoint/s3/s3_rest.js
kajalpareek-lab: Agreed. I’ll delete the local helper and call s3_utils.is_bypass_governance_requested.

src/endpoint/s3/s3_rest.js
kajalpareek-lab: Those lines weren’t dropped — they moved into _get_account_identifiers because Bypass also needs the same principal list when checking bucket policy for s3:BypassGovernanceRetention. Same logic as before (NC: id+name, hosted: id+arn), just shared by both call sites. I can restore the old explanatory comment on that helper.Only removed the comment when extracting _get_account_identifiers.

src/endpoint/s3/s3_rest.js
kajalpareek-lab: Fair — it tries to say too much. I’ll replace it with a short note: when the Bypass header is set, require s3:BypassGovernanceRetention (or owner/admin override); NC uses bucket policy only for non-owners.

src/endpoint/s3/s3_utils.js
kajalpareek-lab: Agreed. I’ll keep a one-liner: returns whether the Bypass header is true.

src/sdk/namespace_fs.js
kajalpareek-lab: OK — I’ll drop that from the JSDoc (or leave a single short @param description).

src/server/object_services/object_server.js
kajalpareek-lab: It explains why we read obj.lock_settings instead of get_object_info(..., { role }) after removing the admin-only gate — so IAM/bucket-policy users can still evaluate retention. I’ll shorten it.

src/server/object_services/object_server.js
kajalpareek-lab: That check made PutObjectRetention admin-only on the server. Bypass (and normal retention updates) must work for IAM/bucket-policy users after endpoint auth. Admin-only would undo that. Authorization for the Bypass header is in s3_rest; here we only honor the already-authorized bypass_governance flag.

src/server/object_services/object_server.js
kajalpareek-lab: Same as above — role check was the old “only admin may Bypass” rule. Now the endpoint decides; object_server only applies the flag.

src/server/object_services/object_server.js
kajalpareek-lab: You’re right for DeleteObjects — post_bucket_delete already maps UNAUTHORIZED via RPC_ERRORS_TO_S3. I’ll remove this redundant mapping.

src/server/object_services/object_server.js
kajalpareek-lab: Delete Bypass used to require admin role. Now s3_rest authorizes Bypass; delete only checks the flag.

src/server/object_services/object_server.js
kajalpareek-lab: Only in unit tests under NODE_ENV=test, same pattern as other __testing exports.

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: Same pattern as the rest of this file — $metadata is SDK noise so deepEqual against {} works. Can switch to asserting no error / status only if you prefer.

src/test/unit_tests/api/iam/test_iam_utils.test.js
kajalpareek-lab: Needed only for the new explicit_deny tests. If you want this file to stay pure helper tests, I can move those cases into test_s3_bypass_governance_auth.test.js.

src/test/unit_tests/util_functions_tests/test_access_policy_utils.test.js
kajalpareek-lab: Agreed — integration coverage is enough. I’ll remove the unit-test addition.

src/test/unit_tests/endpoint/test_s3_bypass_governance_auth.test.js
kajalpareek-lab: Dropped the large dedicated suite. The S3 integration tests already cover the main Bypass paths (header on delete/retention, owner free Bypass). What’s left is a small unit file (test_s3_bypass_governance_permission.test.js) with a few edge cases that are hard to assert end-to-end: Deny wins over Allow (IAM ↔ bucket policy), NC ignoring the IAM stub Allow for non-owners, and DeleteObjects evaluating both bucket and bucket/* ARNs.

src/test/unit_tests/internal/test_object_server.test.js
kajalpareek-lab: Low value — it’s a one-line Boolean wrapper. I can remove the test (and even the helper, inlining Boolean(req.rpc_params?.bypass_governance)).

src/test/unit_tests/api/iam/test_iam_utils.test.js
kajalpareek-lab: Needed only for the new explicit_deny tests. If you want this file to stay pure helper tests, I can move those cases into test_s3_bypass_governance_auth.test.js.

nadavMiz:
> > @kajalpareek-lab @shirady I feel like this design is short sighted. we need a general mechanism to check for permission requiring flags. and to have a map between flags and permissions. so it would be easy to add such permissions to bucket policy / IAM. and validate that it behaves logically correct with the other permissions. I think we should have a small design discussion about this. WDYT?
> 
> @nadavMiz Agree this Bypass path is a one-off on top of the existing IAM + bucket-policy checks. Happy to do a short design discussion for a reusable “header/flag → extra action” map. This PR still closes the Object Lock Bypass auth gap; we can refactor onto a shared mechanism after that.

@kajalpareek-lab any reason why we shouldn't do it correctly in the first place? I don't think it will add much complexity. and those temporary solutions has a tendency to be forgotten and turn permanent with time. you would also need to use a similer flag with`s3:PutObjectLegalHold`. it seems to me it will just make the general solution PR more complicated

PR #9882, opened 2026-07-28
[Deep Archive] - Archive worker implementation for transitioning objects

karthikps97:
### Desciption
Archive worker transitions the object to archive storage class. 
Each object's parts are fetched, and ranges are generated based on configured parameters. Multipart upload is created and each range is uploaded concurrently and status returned back to caller.

Each parameter is configurable:
PART_SIZE - currently set to the max part size for object.
MULTIPART_UPLOAD_SIZE - Max size of the object stream that would be uploaded in a single S3 multipart upload. Controlled by `PART_SIZE` * `MULTIPART_PARTS_COUNT`
MAX_ATTEMPTS - controls the number of times a part is tried to be uploaded to the archive resource.
MULTIPART_CONCURRENCY - controls the number of parts uploaded concurrently.
Retries use exponential backoff strategy.

LIFECYCLE_TRANSITION_BATCH_SIZE - number of objects sent to archive worker concurrently.
LIFECYCLE_TRANSITION_MAX_BATCHES_PER_RUN - The `object_server` functions are paginated and this controls how many batches are run where each batch is of size `LIFECYCLE_BATCH_SIZE`.

Additional improvements:
1. Perform batch updates for transitioned objects. Objects are grouped by original storage class.
2. S3 allows date based transition. Add support for it.
3. Some transition eligibility calculation improvements to conform with S3 behavior.
4. Add support for lifecycle rule filters in transition.

### Explain the Changes
1. archive_object - reads the object's chunk mapping via read_object_mapping, splits into multiple ranges aligned to chunk boundaries, then uploads each range to the archive backend concurrently (with retries). Supports multipart initiate/upload/complete/abort lifecycle.
2. Add new object_server function to read object metadata using object_id. Current function has a mandatory check for object key.
3. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## New Features
- Added lifecycle transitions to archive storage for current, versioned, and noncurrent objects.
- Added administrative object-metadata retrieval and transition-management APIs.
- Added filtering by prefix, object size, tags, and date-based eligibility.
- Added structured transition status and expiration metadata.
- Added configurable multipart upload concurrency and lifecycle batching.

## Reliability
- Added upload retries, bounded concurrency, and cleanup after failures.
- Added lifecycle batch limits and automatic retry processing for remaining work.

### DIFF
--- a/src/api/common_api.js
+++ b/src/api/common_api.js
@@ -1699,6 +1699,15 @@ module.exports = {
             type: 'string',
             enum: ['IN_PROGRESS', 'DONE']
         },
+        transition_status: {
+            type: 'object',
+            required: ['status'],
+            properties: {
+                status: { $ref: '#/definitions/transition_status_enum' },
+                expired_data_ts: { idate: true },
+                expired_data_storage_class: { type: 'string' },
+            },
+        },
         bucket_logging: {
             type: 'object',
             required: ['log_bucket', 'log_prefix'],
--- a/src/server/bg_services/archive_server.js
+++ b/src/server/bg_services/archive_server.js
@@ -1,18 +1,32 @@
 /* Copyright (C) 2026 NooBaa */
 'use strict';
 
+const { RpcError } = require('../../rpc');
 const config = require('../../../config');
-const dbg = require('../../util/debug_module')(__filename);
-const system_store = require('../system_services/system_store').get_instance();
-const pool_server = require('../system_services/pool_server');
-const NamespaceS3 = require('../../sdk/namespace_s3');
-const noobaa_s3_client = require('../../sdk/noobaa_s3_client/noobaa_s3_client');
+const server_rpc = require('../server_rpc');
+const ObjectIO = require('../../sdk/object_io');
 const LRUCache = require('../../util/lru_cache');
+const ObjectSDK = require('../../sdk/object_sdk');
+const NamespaceS3 = require('../../sdk/namespace_s3');
+const { ARCHIVE } = require('../../common/constants');
+const utils = require('../../util/deep_archive_utils');
+const dbg = require('../../util/debug_module')(__filename);
+const auth_server = require('../common_services/auth_server');
+const pool_server = require('../system_services/pool_server');
 const { get_archive_key } = require('../../util/deep_archive_utils');
+const { map_with_concurrency_and_attempts } = require('../../util/promise');
+const system_store = require('../system_services/system_store').get_instance();
+const noobaa_s3_client = require('../../sdk/noobaa_s3_client/noobaa_s3_client');
+
+/* S3 multipart upload size is limited by 5GB. Each multipart upload size 
+is currently configured by max part size in IO pipeline */
+const PART_SIZE = config.MAX_OBJECT_PART_SIZE;
+const MULTIPART_UPLOAD_SIZE = config.MULTIPART_PARTS_COUNT * PART_SIZE;
+const MAX_ATTEMPTS = 3;
+const RETRY_DELAY_MS = 500;
 
 // Max batch size for S3 DeleteObjects — AWS accepts at most 1000 keys per request.
 const S3_DELETE_OBJECTS_BATCH_SIZE = 1000;
-
 
 const archive_ns_cache = new LRUCache({
     name: 'ArchiveNamespaceCache',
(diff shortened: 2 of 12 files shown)


karthikps97:
@coderabbitai ignore 

Waiting for #9840 to merge

karthikps97:
@coderabbitai resume

src/api/common_api.js
```diff
@@ -1696,6 +1660,15 @@ module.exports = {
...
             enum: ['IN_PROGRESS', 'DONE']
         },
+        transition_status: {
+            type: 'object',
+            required: ['status'],
+            properties: {
+                status: { $ref: '#/definitions/transition_status_enum' },
+                expired_data_ts: { idate: true },
+                expired_data_storage_class: { type: 'string' },
+            },
```
romayalon, line 1709: I did some changes there as well, https://github.com/noobaa/noobaa-core/pull/9921/changes#diff-eee06a508f9049bbd7be6bdd78bf1a6f4b32dc6d7805371560d376502305ac42R1948-R1953
i'll flatten it on my PR if possible

romayalon: maybe we need to rename 
expired_data_ts -> transition_timestamp
expired_data_storage_class -> source_storage_class
and we probably should also add target_storage_class

src/server/bg_services/lifecycle.js
```diff
@@ -90,15 +110,21 @@ async function process_transition(system, bucket_info, rule) {
         let is_truncated = true;
         let key_marker = '';
         let version_seq_marker;
-        // List objects that are not deleted, reclaimed and not having transition_status as in_progress
-        while (is_truncated) {
+        let batches = 0;
+        const max_batches = config.LIFECYCLE_TRANSITION_MAX_BATCHES_PER_RUN;
```
romayalon: why did you add this?

src/server/bg_services/lifecycle.js
```diff
@@ -504,7 +527,26 @@ async function transition_objects(system, bucket_info, objects, target_storage_c
     if (failed_objects.length) {
         dbg.warn("failed to archive objects", failed_objects, "retry in next cycle");
     }
-    return obj_ids.filter(Boolean);
+    const transitioned = results.filter(Boolean);
+    if (transitioned.length) {
+        const groups = _.groupBy(transitioned, 'original_storage_class');
```
romayalon: why do we need this grouping? it's probably won't be needed if reverting to update one by one

src/server/bg_services/lifecycle.js
```diff
@@ -504,7 +527,26 @@ async function transition_objects(system, bucket_info, objects, target_storage_c
...
+        await P.map_with_concurrency(10, Object.entries(groups), async ([orig_class, group]) => {
+            await server_rpc.client.object.update_transition_status({
```
romayalon: we should handle errors inside the loop so other calls won't fail

src/server/bg_services/lifecycle.js
```diff
@@ -476,23 +511,11 @@ async function transition_objects(system, bucket_info, objects, target_storage_c
             });
 
             if (archive_status.success) {
-                await server_rpc.client.object.update_transition_status({
```
romayalon, line 479: we shouldn't wait with updating the transition_status after the loop, we should do it immidately

src/server/object_services/object_server.js
```diff
@@ -1085,6 +1085,24 @@ function _check_encryption_permissions(src_enc, req_enc) {
 }
 
 
+/**
+ *
+ * READ_OBJECT_MD_BY_ID
+ *
+ */
+async function read_object_md_by_id(req) {
```
romayalon, line 1093: why do you need a new function? I think find_cached_object_upload() does what you want and also goes to cache before going to the db

src/server/object_services/object_server.js
```diff
@@ -1213,17 +1246,27 @@ async function update_transition_status(req) {
     throw_if_maintenance(req);
     const { rpc_params } = req;
 
+    if (!rpc_params.obj_ids && !rpc_params.obj_id) {
```
romayalon: why it's here and not in the api?

src/server/object_services/md_store.js
```diff
@@ -31,6 +31,43 @@ const config = require('../../../config');
...
+        idx += 1;
+    }
+    if (filters.size_less !== undefined && filters.size_less !== null) {
+        conditions.push(`(data->>'size')::BIGINT < $${idx}`);
+        values.push(filters.size_less);
+        idx += 1;
+    }
+    if (filters.tags && filters.tags.length) {
+        conditions.push(`data->'tagging' @> $${idx}::jsonb`);
+        values.push(JSON.stringify(filters.tags.map(t => ({ key: t.key, value: t.value }))));
```
romayalon: why we need this map?

src/server/object_services/md_store.js
```diff
@@ -31,6 +31,43 @@ const config = require('../../../config');
...
+        values.push(filters.size_greater);
+        idx += 1;
+    }
+    if (filters.size_less !== undefined && filters.size_less !== null) {
+        conditions.push(`(data->>'size')::BIGINT < $${idx}`);
+        values.push(filters.size_less);
+        idx += 1;
+    }
+    if (filters.tags && filters.tags.length) {
+        conditions.push(`data->'tagging' @> $${idx}::jsonb`);
```
romayalon, line 64: why here we don't use ranked.tags like we have on lifecycle for expiry

src/server/object_services/md_store.js
```diff
@@ -2450,41 +2487,69 @@ class MDStore {
...
+        WHERE 
+            data->>'bucket' = $1
+            AND (data->'deleted' IS NULL OR data->'deleted' = 'null'::jsonb)
+            AND (data->'reclaimed' IS NULL OR data->'reclaimed' = 'null'::jsonb)
+            AND (data->'upload_started' IS NULL OR data->'upload_started' = 'null'::jsonb)
+            AND (data->'version_past' IS NULL OR data->'version_past' = 'null'::jsonb)
+            AND (data->'delete_marker' IS NULL OR data->'delete_marker' = 'null'::jsonb)
+            AND (data->'transition_status' IS NULL OR data->'transition_status' = 'null'::jsonb)`;
+
+        let idx = 2;
```
romayalon: why using this idx mechanism instead of pushing to the values array and then check the size

src/server/object_services/md_store.js
```diff
@@ -2450,41 +2487,69 @@ class MDStore {
...
+            query +=
+                `
+                AND (
+                    date_trunc('day', (data->>'create_time')::timestamptz AT TIME ZONE 'UTC') + interval '1 day'
+                ) AT TIME ZONE 'UTC' <= $${idx}::timestamptz`;
+
+            const create_time_cutoff = moment.unix(params.transition_ts).toISOString();
+            values.push(create_time_cutoff);
+            idx += 1;
+        }
```
romayalon, line 2530: don't we need else?

src/server/object_services/md_store.js
```diff
@@ -2550,14 +2615,43 @@ class MDStore {
             `(data->'delete_marker' IS NULL OR data->'delete_marker' = 'null'::jsonb)`,
         ];
 
+        // Prefix filter goes in base_conditions — all versions of a key share the
+        // same key, so filtering early is efficient.
+        if (prefix) {
+            const escaped = prefix.replace(/'/g, "''").replace(/%/g, '\\%').replace(/_/g, '\\_');
```
romayalon, line 2616: we do this escaping of the prefix in multiple places, can't we do it in a function? 
why do we need this escaping, is it missing on the expiry lifecycle? because i don't see it there

src/server/bg_services/archive_server.js
```diff
@@ -1,62 +1,227 @@
...
 */
 async function archive_object(req) {
-    dbg.log1('archive_object', req.rpc_params);
-    throw new Error('archive_object not yet implemented');
+    try {
+        const { obj_id, bucket_id, target_storage_class } = req.rpc_params;
+        dbg.log1('archive_object: starting', obj_id, bucket_id, target_storage_class);
+
+        if (!Object.keys(ARCHIVE.STORAGE_CLASS).includes(target_storage_class)) {
+            throw new RpcError('INVALID_ARGUMENTS',
```
romayalon:
```suggestion
            throw new RpcError('INVALID_TRANSITION_TARGET_STORAGE_CLASS',
```

src/server/bg_services/archive_server.js
```diff
@@ -1,62 +1,227 @@
...
+                    start,
+                    end,
+                    size,
+                    part_num,
+                });
+                part_num += 1;
+            }
+            size = part_size;
+            start = part.start;
+            end = part.end;
```
romayalon: can you explain these?

src/server/bg_services/archive_server.js
```diff
@@ -1,62 +1,227 @@
...
+            const part = chunk.parts[0];
+            const part_size = part.end - part.start;
+            if (size + part_size <= MULTIPART_UPLOAD_SIZE) {
+                end = part.end;
+                size += part_size;
+                continue;
+            }
+
+            if (size > 0) {
+                multipart_ranges.push({
```
romayalon: can you try call functions that receive an object inline? it just makes the function huge for no reason

src/server/bg_services/archive_server.js
```diff
@@ -1,62 +1,227 @@
...
+                    part_num,
+                });
+                part_num += 1;
+            }
+            size = part_size;
+            start = part.start;
+            end = part.end;
+        }
+
+        if (size > 0) {
```
romayalon: why do we have this out of the loop? what this is handling?

src/server/bg_services/archive_server.js
```diff
@@ -1,62 +1,227 @@
...
+                dbg.error("error aborting multipart upload for key", object_key,
+                    "bucket", bucket_id, "error", err);
+            });
+            throw e;
+        }
+    } catch (e) {
+        if (e instanceof RpcError) {
+            throw e;
+        }
+        throw new Error('archive_object: internal_error', { cause: e });
```
romayalon, line 202: why this is needed?

src/server/bg_services/archive_server.js
```diff
@@ -1,62 +1,227 @@
...
-});
-
 /* 
     This function will transition an object from Standard storage class to Deep Archive storage class.
     It will use the endpoint defined by the deep_archive_resource of the bucket.
 */
 async function archive_object(req) {
-    dbg.log1('archive_object', req.rpc_params);
-    throw new Error('archive_object not yet implemented');
+    try {
```
romayalon, line 69: this is a complex function, we must have above the function explaining what is does
also splitting to functions and giving meaningful names can help with that

src/util/promise.js
```diff
@@ -54,6 +54,32 @@ async function map_with_concurrency(concurrency, arr, func) {
     return Promise.all(arr.map(async (key, index) => sem.surround(async () => func(key, index))));
 }
 
+async function map_with_concurrency_and_attempts(concurrency, max_attempts, delay_ms, arr, func) {
+    const sem = new semaphore.Semaphore(concurrency);
```
romayalon, line 58: why this function is needed?

src/server/bg_services/archive_server.js
```diff
@@ -1,62 +1,227 @@
...
+            });
+        }
+
+        const multiparts = [];
+        try {
+            const results = await map_with_concurrency_and_attempts(config.MULTIPART_CONCURRENCY,
+                MAX_ATTEMPTS, RETRY_DELAY_MS, multipart_ranges, async r => {
+                    try {
+                        const source_stream = object_io.read_object_stream({
+                            client: rpc_client,
```
romayalon, line 149: per my understanding you are utilizing the object io buffers semaphore, which is good, I just wonder if the configs are correct for noobaa-core as well
WDYT?  @karthikps97  @dannyzaken

src/api/common_api.js
karthikps97: We wouldn't need `target_storage_class` since the object MD's `storage_class`field  is itself updated.

romayalon: it's good to keep it so we will know the source and the target, regardless of the current storageclass mentioned on the object itself

romayalon: we can add it in a following PR

karthikps97: `source_storage_class` maybe confusing as it may imply the current source where data is stored? `expired_data_storage_class` may be better?

src/server/bg_services/archive_server.js
karthikps97: This is for the scenario where we exit the loop but still have a part left since it was less than MULTIPART_UPLOAD_SIZE

src/server/bg_services/lifecycle.js
karthikps97: Yes, if we do go with that approach we don't need all this.

src/util/promise.js
karthikps97: I added this so that we can have both concurrency and retry in a single helper function.

src/server/bg_services/lifecycle.js
```diff
@@ -475,25 +510,30 @@ async function transition_objects(system, bucket_info, objects, target_storage_c
...
                     })
-                });
-                dbg.log1('object successfully transitioned', obj_id);
-                return obj_id;
-            } else {
-                dbg.log1("got success false status from archive_server for object", obj_id);
-                throw new Error("archive_server failed to transition object", obj.key);
-            }
+                }),
+            });
```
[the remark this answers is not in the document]
karthikps97: It's ok for a later run to rearchive the same object.

src/server/object_services/object_server.js
karthikps97: I would still have to add a object server API since `find_cached_object_upload` is not exposed as an API. IMO we shouldn't rely on cache for this since it has an expiry of 1sec. Non critical operations can still use this

src/server/object_services/md_store.js
karthikps97: No,  we already check whether the date has elapsed early on. If elapsed, all objects satisying the filters would we eligible for transition.

src/server/bg_services/archive_server.js
karthikps97: yes, added a function to calculate the range

src/server/bg_services/archive_server.js
karthikps97: we need to take into account the current part after we push the existing range. So current part gets added to next range.

src/server/bg_services/archive_server.js
karthikps97: I added a log statement. So, i would need this catch block. Any error apart from rpc errors would be marked as internal error.

src/server/bg_services/archive_server.js
karthikps97: I added a separate function for range calculation.

src/server/object_services/schemas/object_md_schema.js
```diff
@@ -146,9 +146,14 @@ module.exports = {
...
-
-        data_expired: { date: true },
+        transition_status: {
+            type: 'object',
+            properties: {
+                status: { $ref: '#/definitions/transition_status_enum' },
+                expired_data_ts: { date: true },
+                expired_data_storage_class: { type: 'string' },
+            },
+        },
```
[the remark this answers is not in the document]
karthikps97: This code is not yet deployed on any production systems. So we needn't  any backward compatibility. Its safe to directly modify the code.

src/server/object_services/md_store.js
karthikps97: We can use that also. This is just another approach i follow.

src/server/object_services/md_store.js
karthikps97: this was to sanitize the input, but it's not needed. We have other methods not following this approach. So removed

src/server/object_services/md_store.js
karthikps97: This function is used for query where we dont have any rank. So its referrring to `tagging` directly

src/server/object_services/object_server.js
karthikps97: These are not needed anymore as per latest changes. Removed them

romayalon (approved):
approving so we can continue
we should check that transitions of large object work and also still some refactoring is needed

PR #9887, opened 2026-07-29
S3 API compatibility | Multipart Upload | GetObject | Bug fixes

romayalon:
### Explain the Changes
1. **NamespaceNB.js** - Validate ObjectId format in NamespaceNB for upload-part, list-parts, complete, and abort. Malformed IDs now return S3 NoSuchUpload instead of RPC INVALID_SCHEMA_PARAMS (we expected objectid but when providing non mongo object id we got this invalid schema issue)

2. **http_utils.js** -  In normalize_http_ranges, treat first-byte-pos >= object size as unsatisfiable (RFC 7233 / AWS), including empty objects, so GET returns 416 InvalidRange instead of 206 with a bad Content-Range.
see RFC - 
https://datatracker.ietf.org/doc/html/rfc7233#section-2.1
```
   If a valid byte-range-set includes at least one byte-range-spec with
   a first-byte-pos that is less than the current length of the
   representation, or at least one suffix-byte-range-spec with a
   non-zero suffix-length, then the byte-range-set is satisfiable.
   Otherwise, the byte-range-set is unsatisfiable.
  ```

4. **Tests** - Coverage in test_s3_ops (invalid upload id) and test_http_utils (empty / past-EOF ranges).

1. Fixes https://redhat.atlassian.net/browse/DFBUGS-8009 and https://redhat.atlassian.net/browse/DFBUGS-8008

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Multipart upload operations now consistently fail fast with **NoSuchUpload (404)** when the upload ID is missing or malformed.
  * Unsatisfiable HTTP byte-range requests, including requests for empty objects or ranges starting at/after the end, now return **416**.

* **Tests**
  * Added integration and unit coverage for invalid upload IDs and unsatisfiable byte ranges.

* **Documentation**
  * Updated range-validation guidance to reflect S3/RFC 7233 rules.

### DIFF
--- a/src/test/integration_tests/api/s3/test_s3_ops.js
+++ b/src/test/integration_tests/api/s3/test_s3_ops.js
@@ -1,6 +1,8 @@
 /* Copyright (C) 2016 NooBaa */
 /* eslint-disable max-lines-per-function */
 /* eslint-disable no-invalid-this */
+/* eslint-disable max-lines */
+
 
 'use strict';
 const _ = require('lodash');
@@ -1491,6 +1493,40 @@ mocha.describe('s3_ops', function() {
             });
         });
 
+        mocha.it('should return NoSuchUpload for invalid multipart upload id', async function() {
+            // Data buckets only — validates NamespaceNB invalid UploadId handling.
+            // TODO - fix on other Namespace such as NamespaceS3
+            if (bucket_type !== 'regular') this.skip();
+            const key = 'test-nonexistent-upload';
+            const invalid_upload_id = '00000000-does-not-exist-00000000';
+
+            await assert_no_such_upload(() => s3.uploadPart({
+                Bucket: bucket_name,
+                Key: key,
+                UploadId: invalid_upload_id,
+                PartNumber: 1,
+                Body: 'part',
+            }));
+            await assert_no_such_upload(() => s3.listParts({
+                Bucket: bucket_name,
+                Key: key,
+                UploadId: invalid_upload_id,
+            }));
+            await assert_no_such_upload(() => s3.completeMultipartUpload({
+                Bucket: bucket_name,
+                Key: key,
+                UploadId: invalid_upload_id,
+                MultipartUpload: {
+                    Parts: [{ ETag: 'jkdsnfkndfsjknfdsk', PartNumber: 1 }],
+                },
+            }));
+            await assert_no_such_upload(() => s3.abortMultipartUpload({
+                Bucket: bucket_name,
+                Key: key,
+                UploadId: invalid_upload_id,
+            }));
+        });
+
         mocha.it('should list objects with text-file', async function() {
             this.timeout(60000);
             const ORIG_INLINE_MAX_SIZE = config.INLINE_MAX_SIZE;
@@ -1962,4 +1998,14 @@ function is_namespace_blob_bucket(bucket_type, remote_endpoint_type) {
 // so we would skip copy tests and deletions of objects created by copy
 function is_namespace_blob_mock(bucket_type, remote_endpoint_type) {
     return remote_endpoint_type === 'AZURE' && bucket_type === 'namespace' && !process.env.NEWAZUREPROJKEY;
+}
+
+async function assert_no_such_upload(fn) {
+    try {
+        await fn();
+        assert.fail('expected NoSuchUpload');
+    } catch (err) {
+        assert.strictEqual(test_utils.err_code(err), 'NoSuchUpload');
+        assert.strictEqual(err.$metadata.httpStatusCode, 404);
+    }
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/test/integration_tests/api/s3/test_s3_ops.js
Comment (shirady) [inline src/test/integration_tests/api/s3/test_s3_ops.js:1]: Could you fix the lint error so we can see it running in the CI? 

> 2001:1  error  File has too many lines (2007). Maximum allowed is 2000  max-lines

File: src/util/http_utils.js [hunk 1]
```diff
@@ -378,6 +378,9 @@ function _format_one_http_range({ start, end }) {
 }
 
 /**
+ * Normalize parsed HTTP ranges against entity size.
+ * Per RFC 7233 §2.1 / AWS S3, a range is satisfiable only if first-byte-pos < size;
```
shirady, line 382: Not sure, but I think instead of `§` `section` will be more readable.

src/util/http_utils.js
```diff
@@ -394,7 +397,7 @@ function normalize_http_ranges(ranges, size, throw_error_ranges = false) {
             }
             r.end = size;
         }
-        if (r.start < 0 || r.start > r.end) throw_ranges_error(416);
+        if (r.start < 0 || r.start > r.end || r.start >= size) throw_ranges_error(416);
```
shirady, line 400: In this condition, we are in `r.end <= size`.
So it seems like `r.start >= size` includes the check `r.start > r.end`.
Am I missing something?
```suggestion
        if (r.start < 0 || r.start >= r.end) throw_ranges_error(416);
```

src/test/unit_tests/util_functions_tests/test_http_utils.js
```diff
@@ -60,4 +60,42 @@ mocha.describe('http_utils', function() {
...
+        function assert_416(range_header, size) {
+            try {
+                normalize(range_header, size);
+                assert.fail('expected 416 InvalidRange');
+            } catch (err) {
+                assert.strictEqual(err.ranges_code, 416);
+            }
+        }
+
+        mocha.it('should return 416 for range on empty object (AWS S3)', function() {
```
shirady, line 81: Could you edit the titles so that it is clear that it is an error?
Also, it is a test for the general helper, I don't see a reason to add terms like AWS S# Object here
```suggestion
        mocha.it('should throw error code 416 for range on empty range'', function() {
```

src/test/unit_tests/util_functions_tests/test_http_utils.js
```diff
@@ -60,4 +60,42 @@ mocha.describe('http_utils', function() {
...
+            assert_416('bytes=0-99', 0);
+            assert_416('bytes=0-0', 0);
+            assert_416('bytes=0-', 0);
+        });
+
+        mocha.it('should return 416 when first-byte-pos is past the end', function() {
+            assert_416('bytes=100-200', 100);
+            assert_416('bytes=5-10', 5);
+        });
```
shirady, line 90: If I understand correctly, this is what used to pass, and now it's fixed, right?
Worth editing the title and adding a comment so it would be clearer for others.

```suggestion
        mocha.it('should throw 416 error on empty range', function() {
            assert_416('bytes=100-200', 100); // start=5, end=5, empty range
            assert_416('bytes=5-10', 5); // start=5, end=5, empty range
```

src/test/integration_tests/api/s3/test_s3_ops.js
```diff
@@ -1963,3 +1995,13 @@ function is_namespace_blob_bucket(bucket_type, remote_endpoint_type) {
 function is_namespace_blob_mock(bucket_type, remote_endpoint_type) {
     return remote_endpoint_type === 'AZURE' && bucket_type === 'namespace' && !process.env.NEWAZUREPROJKEY;
 }
+
+async function assert_no_such_upload(fn) {
+    try {
+        await fn();
+        assert.fail('expected NoSuchUpload');
+    } catch (err) {
+        assert.strictEqual(err.Code || err.name, 'NoSuchUpload');
```
shirady: You can reuse the function that you created in the past:
https://github.com/noobaa/noobaa-core/blob/d9213d35a24f918ee24384921900be86566d2523/src/test/system_tests/test_utils.js#L969-L975

shirady (approved):
LGTM
Suggested a few minor changes

src/test/unit_tests/util_functions_tests/test_http_utils.js
romayalon: no, the original bug when the size is 0

src/util/http_utils.js
romayalon: basically yes but it find it more understood that start larger than size, also, changing to r.start >= r.end actually changes the behavior of another use case but it's already incorrect - I opened an issue, seems like ranges are incompatible to AWS on more cases, regardless of this specific issue - https://github.com/noobaa/noobaa-core/issues/9889

PR #9890, opened 2026-08-02
NC | lifecycle | Fixing delete filter for noncurrent versions

jackyalbo:
### Describe the Problem
process_rule built one filter that always included Expiration.Days (against create_time) and applied it to all delete candidates, including noncurrent ones.
noncurrent version could qualify under NoncurrentDays, then fail the delete filter because its create_time was still younger than Expiration.Days — and never get deleted.

### Explain the Changes
1. The fix keeps the two candidate lists separate and deletes noncurrent objects with only the prefix/tags/size filter, not Expiration.Days.

1. Fixed https://redhat.atlassian.net/browse/DFBUGS-9068 

### Testing Instructions:
1. Create a bucket with versioning and this lifecycle rules, for example:
```
"Rules": [
    {
      "ID": "expiration-must-not-block-noncurrent",
      "Filter": { "Prefix": "" },
      "Status": "Enabled",
      "Expiration": { "Days": 30 },
      "NoncurrentVersionExpiration": { "NoncurrentDays": 1 }
    }
]
```
2. Create two versions of the same key (latest stays current; first becomes noncurrent)
3. Wait one day
4. Run the lifecycle worker:
```noobaa-cli lifecycle --disable_runtime_validation --disable_service_validation```
5. Make sure only one version left(the latest):
aws s3api list-object-versions \
  --bucket <bucket> --prefix repro-key \
  --endpoint-url <noobaa-endpoint>
6. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved lifecycle expiration handling for current and noncurrent object versions.
  * Ensured noncurrent versions are evaluated using the correct expiration rules.
  * Fixed versioned-object cleanup so eligible older versions are removed while the latest version remains.
  * Improved reporting of multipart-upload candidates during lifecycle processing.

* **Tests**
  * Added regression coverage for versioned lifecycle expiration scenarios.

* **Style**
  * Cleaned up documentation formatting.

### DIFF
--- a/src/manage_nsfs/nc_lifecycle.js
+++ b/src/manage_nsfs/nc_lifecycle.js
@@ -300,27 +300,41 @@ class NCLifecycle {
                 op_func: async () => this.get_candidates(bucket_json, lifecycle_rule, object_sdk)
             });
 
-            if (candidates.delete_candidates?.length > 0) {
-                const expiration = lifecycle_rule.expiration ? this._get_expiration_time(lifecycle_rule.expiration) : 0;
-                const filter_func = lifecycle_utils.build_lifecycle_filter({filter: lifecycle_rule.filter, expiration});
-                dbg.log0('process_rule: calling delete_multiple_objects, num of objects to be deleted', candidates.delete_candidates.length);
-                const delete_res = await this._call_op_and_update_status({
-                    bucket_name,
-                    rule_id,
-                    op_name: TIMED_OPS.DELETE_MULTIPLE_OBJECTS,
-                    op_func: async () => object_sdk.delete_multiple_objects({
-                        bucket: bucket_json.name,
-                        objects: candidates.delete_candidates,
-                        filter_func
-                    })
-                });
-                if (should_notify) {
-                    await this.send_lifecycle_notifications(delete_res, candidates.delete_candidates, bucket_json, object_sdk);
-                }
-            }
+            // Expiration and noncurrent candidates must use different delete-time filters:
+            // Expiration re-checks Days/Date against create_time; noncurrent must not, or
+            // NoncurrentVersionExpiration would be blocked by Expiration.Days.
+            // Marker-only rules (expired_object_delete_marker) use expiration: 0 — same as candidate
+            // selection — because _get_expiration_time() returns -1 when neither days nor date is set.
+            // NoncurrentDays / newer_noncurrent_versions are applied only at candidate selection time.
+            const expiration_for_delete = (lifecycle_rule.expiration?.days || lifecycle_rule.expiration?.date) ?
+                this._get_expiration_time(lifecycle_rule.expiration) : 0;
+            await this._delete_candidates_with_filter({
+                bucket_name,
+                rule_id,
+                bucket_json,
+                object_sdk,
+                objects: candidates.delete_candidates,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/manage_nsfs/nc_lifecycle.js [hunk 1]
```diff
@@ -300,27 +300,35 @@ class NCLifecycle {
...
+            });
+            await this._delete_candidates_with_filter({
+                bucket_name,
+                rule_id,
+                bucket_json,
+                object_sdk,
+                objects: candidates.noncurrent_delete_candidates,
+                filter_func: lifecycle_utils.build_lifecycle_filter({
+                    filter: lifecycle_rule.filter,
+                    expiration: 0,
```
romayalon, line 331: @jackyalbo @nadavMiz why don't we pass the filter the non current version expiration properties, noncurrent_days and newer_noncurrent_versions, we will need to re-validate it there
I see we have filter_noncurrent_days() and filter_newer_versions() but per my understanding they should be a part of the filter_func

romayalon: we should move them inside filter_func if possible for re-validation before deletion which also missing right now

romayalon: another benefit would be that we won't need to split the delete calls to current and non current, we will only need to know if we should validate expiration / non expiration

jackyalbo: @romayalon Added nc_age (in a new commit for now) to filter_func but still used two lists as we can't use both expiration and nc_expiration at the same time - for some it's relevant and for NC it doesn't - anyway I think splitting the lists is not that bad.

src/manage_nsfs/nc_lifecycle.js
```diff
@@ -300,27 +300,37 @@ class NCLifecycle {
...
-                        filter_func
-                    })
-                });
-                if (should_notify) {
-                    await this.send_lifecycle_notifications(delete_res, candidates.delete_candidates, bucket_json, object_sdk);
-                }
-            }
+            // Expiration candidates re-check Expiration.Days against create_time at delete time.
+            // Noncurrent candidates re-check NoncurrentDays against nc_noncurrent_time (not create_time).
+            // newer_noncurrent_versions is list-scoped and remains validated only at candidate selection.
```
romayalon: // newer_noncurrent_versions is list-scoped and remains validated only at candidate selection.
why? do we plan to add it?

romayalon (commented):
@jackyalbo I left one comment but check code's rabbit comments, he claims we have a similar issue when having expiration + delete marker actions

src/manage_nsfs/nc_lifecycle.js
jackyalbo: @romayalon We don't, and we never did - I think it may be a bit confusing. Should I remove it? or you think the new comment is less confusing?

romayalon: I know but i'm saying it's a gap so we probably want to fix it

jackyalbo: We can't really fix it in the delete filter function as we need info about how many other versions we have, and it's a per object filter function

romayalon (approved):

PR #9895, opened 2026-08-04
Block account deletion when owned roles remain

aayushchouhan09:
### Describe the Problem
Accounts that still own IAM roles can be deleted, leaving orphaned roles.

### Explain the Changes
1. Block account deletion when the account still owns active IAM roles.

### Testing Instructions:
1.  Try deleting the account owned the iam role.

- [ ] Doc added/updated
- [ ] Tests added

## Bug Fixes

* Account deletion is now blocked when the account owns active IAM roles.
* A clear forbidden response is returned when deletion cannot proceed.
* IAM role policies are cleaned up correctly during account and role deletion workflows, improving cleanup reliability.


shirady (approved):
LGTM

PR #9896, opened 2026-08-04
block bucket delete when Object Lock protected objects remain

kajalpareek-lab:
## Summary
- Fail closed on OBC/bucket wipe when any object still has Object Lock protection (legal hold or active retention).
- Fence writes by marking the bucket deleting before the lock check; roll back that fence if protected objects are found.
- Independently refuse unordered reclaim deletion of locked objects so Object Lock cannot be bypassed.
## Test plan
- [ ] Unit tests: `npx jest src/test/unit_tests/internal/test_delete_bucket_object_lock.test.js --forceExit`
- [ ] Delete OBC/bucket with no locked objects → succeeds
- [ ] Delete OBC/bucket with retention or legal hold → fails with clear Object Lock error; bucket not left Terminating
- [ ] After unlock / retention expiry, retry delete → succeeds

## Summary by CodeRabbit

* **Bug Fixes**
  * Bucket and object deletion now respects Object Lock protections.
  * Deletion is rejected when objects have active legal holds or unexpired retention periods, preventing protected data from being removed.
  * Failed bucket deletions correctly restore the bucket’s prior state, including when lock validation encounters an error.
  * Unlocked objects and buckets continue to be deleted successfully, with accurate bucket availability and activity records.

### DIFF
--- a/src/server/bg_services/buckets_reclaimer.js
+++ b/src/server/bg_services/buckets_reclaimer.js
@@ -49,7 +49,15 @@ class BucketsReclaimer {
                     dbg.log0(`bucket ${bucket.name} is not empty yet`);
                 }
             } catch (err) {
-                dbg.error(`got error when trying to empty and delete bucket ${bucket.name} :`, err);
+                // Object Lock safety invariant: reclaim/unordered deletion independently
+                // refuses to remove protected objects. Object Lock cannot be bypassed;
+                // bucket deletion may fail here and require a retry after unlock/expiry.
+                if (err && err.rpc_code === 'UNAUTHORIZED') {
+                    dbg.error(`bucket_reclaimer: bucket ${bucket.name} has Object Lock protected objects; ` +
+                        `refusing to empty/delete. Remove legal hold / wait for retention, then retry delete.`, err);
+                } else {
+                    dbg.error(`got error when trying to empty and delete bucket ${bucket.name} :`, err);
+                }
                 has_errors = true;
             }
         }));
--- a/src/server/bg_services/buckets_reclaimer.js
+++ b/src/server/bg_services/buckets_reclaimer.js
@@ -49,9 +49,8 @@ class BucketsReclaimer {
                     dbg.log0(`bucket ${bucket.name} is not empty yet`);
                 }
             } catch (err) {
-                // Object Lock safety invariant: reclaim/unordered deletion independently
-                // refuses to remove protected objects. Object Lock cannot be bypassed;
-                // bucket deletion may fail here and require a retry after unlock/expiry.
+                // Do not delete Object Lock protected objects. If any are still locked,
+                // stop here; unlock or wait, then retry delete.
                 if (err && err.rpc_code === 'UNAUTHORIZED') {
                     dbg.error(`bucket_reclaimer: bucket ${bucket.name} has Object Lock protected objects; ` +
                         `refusing to empty/delete. Remove legal hold / wait for retention, then retry delete.`, err);
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -1287,6 +1287,14 @@ async function delete_multiple_objects_by_filter(req) {
  * This function is inteded to use in the case of a bucket deletion
  * where we want to delete all the objects in the bucket but we don't
  * care about the order in which they are deleted or the versioning, etc.
+ *
+ * Object Lock safety invariant:
+ * Every path capable of deleting objects independently enforces Object Lock.
+ * Therefore, even if a protected object is committed after the pre-delete
+ * validation due to an in-flight request, subsequent reclaim or unordered
+ * deletion will refuse to remove it. This guarantees that Object Lock
+ * protection cannot be bypassed, though bucket deletion may fail and
+ * require a retry.
  * @param {*} req
  */
 async function delete_multiple_objects_unordered(req) {
@@ -1294,6 +1302,17 @@ async function delete_multiple_objects_unordered(req) {
     const bucket_id = req.bucket._id;
     const limit = req.rpc_params.limit;
     dbg.log0(`delete_multiple_objects_unordered: bucket=${req.bucket.name} limit=${limit}`);
+
+    // Independently enforce Object Lock on this wipe path (see safety invariant).
+    const has_locked_objects = await MDStore.instance().has_any_locked_objects_in_bucket(bucket_id);
+    if (has_locked_objects) {
+        dbg.error('delete_multiple_objects_unordered: refusing to wipe bucket with Object Lock protected objects',
+            req.bucket.name);
+        throw new RpcError(
+            'UNAUTHORIZED',
+            'Cannot delete bucket objects: one or more objects are protected by Object Lock (retention or legal hold)'
+        );
+    }
 
     // find the objects - no need to paginate explicitly because
     // find_objects will ensure that it does not return any object
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -1288,13 +1288,9 @@ async function delete_multiple_objects_by_filter(req) {
  * where we want to delete all the objects in the bucket but we don't
  * care about the order in which they are deleted or the versioning, etc.
  *
- * Object Lock safety invariant:
- * Every path capable of deleting objects independently enforces Object Lock.
- * Therefore, even if a protected object is committed after the pre-delete
- * validation due to an in-flight request, subsequent reclaim or unordered
- * deletion will refuse to remove it. This guarantees that Object Lock
- * protection cannot be bypassed, though bucket deletion may fail and
- * require a retry.
+ * Also checks Object Lock: do not wipe locked objects. Bucket delete
+ * marks the bucket deleting first so new uploads are blocked; this path
+ * still refuses if any locked object is present.
  * @param {*} req
  */
 async function delete_multiple_objects_unordered(req) {
@@ -1303,7 +1299,7 @@ async function delete_multiple_objects_unordered(req) {
     const limit = req.rpc_params.limit;
     dbg.log0(`delete_multiple_objects_unordered: bucket=${req.bucket.name} limit=${limit}`);
 
-    // Independently enforce Object Lock on this wipe path (see safety invariant).
+    // Stop if any object in the bucket is still locked.
     const has_locked_objects = await MDStore.instance().has_any_locked_objects_in_bucket(bucket_id);
     if (has_locked_objects) {
         dbg.error('delete_multiple_objects_unordered: refusing to wipe bucket with Object Lock protected objects',
@@ -1324,11 +1320,35 @@ async function delete_multiple_objects_unordered(req) {
         key: undefined,
     });
 
+    // Re-check the batch before soft-delete (narrow race with an in-flight
+    // upload that passed load_bucket before the deleting fence).
+    if (objects.some(obj => _is_object_lock_active(obj))) {
+        dbg.error('delete_multiple_objects_unordered: refusing to wipe locked objects in batch',
+            req.bucket.name);
+        throw new RpcError(
+            'UNAUTHORIZED',
+            'Cannot delete bucket objects: one or more objects are protected by Object Lock (retention or legal hold)'
+        );
+    }
+
     // delete the objects
     await MDStore.instance().remove_objects_and_unset_latest(objects);
 
     const bucket_has_objects = await MDStore.instance().has_any_objects_for_bucket(bucket_id);
     return { is_empty: !bucket_has_objects };
+}
+
+/**
+ * True when Object Lock still protects this object version (no governance bypass).
+ * @param {object} obj
+ * @param {Date} [now]
+ */
+function _is_object_lock_active(obj, now = new Date()) {
+    const lock = obj && obj.lock_settings;
+    if (!lock) return false;
+    if (lock.legal_hold && lock.legal_hold.status === 'ON') return true;
+    if (!lock.retention || !lock.retention.retain_until_date) return false;
+    return new Date(lock.retention.retain_until_date) > now;
 }
 
 // Sets the "deleted" field for all Object MDs with `upload_started: { $exists: true } and create_time < expiration threshold
@@ -2499,6 +2519,7 @@ exports.calc_retention = calc_retention;
 
 if (process.env.NODE_ENV === 'test') {
     exports.__testing = {
-        update_bulk_delete_results
+        update_bulk_delete_results,
+        _is_object_lock_active,
     };
 }
--- a/src/server/system_services/bucket_server.js
+++ b/src/server/system_services/bucket_server.js
@@ -1061,17 +1061,9 @@ async function delete_bucket_and_objects(req) {
     const original_name = bucket.name.unwrap();
     const now = new Date();
 
-    // Object Lock safety invariant:
-    // Every path capable of deleting objects independently enforces Object Lock.
-    // Therefore, even if a protected object is committed after the pre-delete
-    // validation due to an in-flight request, subsequent reclaim or unordered
-    // deletion will refuse to remove it. This guarantees that Object Lock
-    // protection cannot be bypassed, though bucket deletion may fail and
-    // require a retry.
-    //
-    // Fence writes before the Object Lock check. Once deleting is set (and the
-    // bucket is renamed), load_bucket() rejects new PutObject / complete_upload
-    // against the original name.
+    // Mark deleting first so new uploads are blocked, then check Object Lock.
+    // If locked objects exist, undo only *this* delete attempt (match deleting=now)
+    // so a concurrent newer delete is not cleared by our rollback.
     await system_store.make_changes({
         update: {
             buckets: [{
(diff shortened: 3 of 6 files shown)


src/server/system_services/bucket_server.js
```diff
@@ -1129,22 +1058,59 @@ async function update_buckets(req) {
...
+    // Fence writes before the Object Lock check. Once deleting is set (and the
+    // bucket is renamed), load_bucket() rejects new PutObject / complete_upload
+    // against the original name.
     await system_store.make_changes({
         update: {
             buckets: [{
                 _id: bucket._id,
                 $set: {
-                    name: `${bucket.name.unwrap()}-deleting-${now.getTime()}`,
+                    name: `${original_name}-deleting-${now.getTime()}`,
```
naveenpaul1, line 1072: has_any_locked_objects_in_bucket() should be done before deleting the bucket? Why are we reverting it, I couldnt get the comments

src/test/unit_tests/internal/test_delete_bucket_object_lock.test.js
```diff
@@ -0,0 +1,163 @@
+/* Copyright (C) 2026 NooBaa */
+
+'use strict';
+
+const SensitiveString = require('../../../util/sensitive_string');
+const { MDStore } = require('../../../server/object_services/md_store');
+const object_server = require('../../../server/object_services/object_server');
+const bucket_server = require('../../../server/system_services/bucket_server');
+const system_store = require('../../../server/system_services/system_store').get_instance();
+const Dispatcher = require('../../../server/notifications/dispatcher');
```
naveenpaul1, line 10: Can you add few test in integration tets suite also?

src/server/object_services/object_server.js
```diff
@@ -1295,6 +1303,17 @@ async function delete_multiple_objects_unordered(req) {
...
+    // Independently enforce Object Lock on this wipe path (see safety invariant).
+    const has_locked_objects = await MDStore.instance().has_any_locked_objects_in_bucket(bucket_id);
+    if (has_locked_objects) {
+        dbg.error('delete_multiple_objects_unordered: refusing to wipe bucket with Object Lock protected objects',
+            req.bucket.name);
+        throw new RpcError(
+            'UNAUTHORIZED',
+            'Cannot delete bucket objects: one or more objects are protected by Object Lock (retention or legal hold)'
+        );
+    }
```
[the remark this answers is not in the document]
naveenpaul1: @kajalpareek-lab can you check this comment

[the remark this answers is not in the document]
kajalpareek-lab: Thanks. Agreed there is a narrow window after has_any_locked_objects_in_bucket where an in-flight upload (that already passed load_bucket before the deleting fence) could show up in find_objects.
We tightened that by re-checking lock on the batch before remove_objects_and_unset_latest and failing closed if any object is locked.
A fully atomic “SELECT FOR UPDATE + lock predicate on the destructive update” would be a larger redesign of this reclaim path and isn’t needed with the deleting fence + dual checks. New uploads after the fence already fail in load_bucket.

src/server/system_services/bucket_server.js
```diff
@@ -1058,22 +1058,59 @@ async function update_buckets(req) {
...
+                        _id: bucket._id,
+                        $set: {
+                            name: original_name,
+                        },
+                        $unset: {
+                            deleting: 1,
+                        }
+                    }]
+                }
+            });
```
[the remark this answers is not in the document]
kajalpareek-lab: Good catch. Rollback now matches _id and the same deleting timestamp we set for this attempt, so an older failed delete cannot clear a newer deletion fence.

src/server/system_services/bucket_server.js
kajalpareek-lab: We mark the bucket deleting first so new uploads are blocked (load_bucket fails).
Then we check for locked objects.
If any are still locked, we undo the deleting mark and restore the name so the bucket is not left Terminating — user unlocks/waits and retries.
Checking only before marking would leave a race where a locked object can be uploaded between the check and the wipe.

src/test/unit_tests/internal/test_delete_bucket_object_lock.test.js
kajalpareek-lab: I’ll add integration coverage

src/server/bg_services/buckets_reclaimer.js
```diff
@@ -49,7 +49,14 @@ class BucketsReclaimer {
                     dbg.log0(`bucket ${bucket.name} is not empty yet`);
                 }
             } catch (err) {
-                dbg.error(`got error when trying to empty and delete bucket ${bucket.name} :`, err);
+                // Do not delete Object Lock protected objects. If any are still locked,
```
nadavMiz, line 52: this is just logging. is this really helpful? wouldn't we have the same information from the existing error?

src/server/object_services/object_server.js
```diff
@@ -1442,6 +1299,17 @@ async function delete_multiple_objects_unordered(req) {
     const limit = req.rpc_params.limit;
     dbg.log0(`delete_multiple_objects_unordered: bucket=${req.bucket.name} limit=${limit}`);
 
+    // Stop if any object in the bucket is still locked.
+    const has_locked_objects = await MDStore.instance().has_any_locked_objects_in_bucket(bucket_id);
```
nadavMiz, line 1303: is delete_multiple_objects_unordered used specificly for deleting buckets? I am concerned we are blocking deletion of unlocked objects just because one object is locked. if that is what we do. did you verify this behavior on AWS

src/server/object_services/object_server.js
```diff
@@ -1452,13 +1320,37 @@ async function delete_multiple_objects_unordered(req) {
         key: undefined,
     });
 
+    // Re-check the batch before soft-delete (narrow race with an in-flight
+    // upload that passed load_bucket before the deleting fence).
+    if (objects.some(obj => _is_object_lock_active(obj))) {
```
nadavMiz, line 1325: do we need both this code and the above code. isn't this enough to validate there aren't locked objects?

src/server/object_services/object_server.js
```diff
@@ -1452,13 +1320,37 @@ async function delete_multiple_objects_unordered(req) {
...
     const bucket_has_objects = await MDStore.instance().has_any_objects_for_bucket(bucket_id);
     return { is_empty: !bucket_has_objects };
 }
 
+/**
+ * True when Object Lock still protects this object version (no governance bypass).
+ * @param {object} obj
+ * @param {Date} [now]
+ */
+function _is_object_lock_active(obj, now = new Date()) {
```
nadavMiz, line 1346: can we maybe make this a util function to use in different areas of the code? maybe we already have function that we can reuse?

src/server/object_services/object_server.js
```diff
@@ -1442,6 +1299,17 @@ async function delete_multiple_objects_unordered(req) {
     const limit = req.rpc_params.limit;
     dbg.log0(`delete_multiple_objects_unordered: bucket=${req.bucket.name} limit=${limit}`);
 
+    // Stop if any object in the bucket is still locked.
+    const has_locked_objects = await MDStore.instance().has_any_locked_objects_in_bucket(bucket_id);
+    if (has_locked_objects) {
+        dbg.error('delete_multiple_objects_unordered: refusing to wipe bucket with Object Lock protected objects',
+            req.bucket.name);
+        throw new RpcError(
+            'UNAUTHORIZED',
```
nadavMiz, line 1308: did you verify on AWS this is the correct error code?

src/server/system_services/bucket_server.js
```diff
@@ -1129,22 +1058,52 @@ async function update_buckets(req) {
...
             }]
         }
     });
 
+    if (!bucket.namespace) {
+        const has_locked_objects = await MDStore.instance().has_any_locked_objects_in_bucket(bucket._id);
+        if (has_locked_objects) {
+            dbg.error('delete_bucket_and_objects: bucket has Object Lock protected objects',
+                original_name);
+            await system_store.make_changes({
```
nadavMiz, line 1084: I am a bit concern about this code. from what I understand this undo the delete of the bucket? since we already marked the bucket deleting before this check, I think we need to be careful about exception safety in this abort path. if has_any_locked_objects_in_bucket throws, we never reach this rollback, and the bucket stays renamed / deleting even though we did not accept the delete. same if this make_changes itself throws.

src/server/object_services/object_server.js
```diff
@@ -1305,13 +1320,37 @@ async function delete_multiple_objects_unordered(req) {
...
+        dbg.error('delete_multiple_objects_unordered: refusing to wipe locked objects in batch',
+            req.bucket.name);
+        throw new RpcError(
+            'UNAUTHORIZED',
+            'Cannot delete bucket objects: one or more objects are protected by Object Lock (retention or legal hold)'
+        );
+    }
+
     // delete the objects
     await MDStore.instance().remove_objects_and_unset_latest(objects);
```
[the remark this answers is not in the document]
kajalpareek-lab: We are not making the soft-delete fully atomic in this PR. After the deleting fence, new uploads fail in load_bucket. We still fail closed with a bucket-level lock check and a batch re-check before remove. A transactional wipe with a lock predicate on the update would be a larger reclaim redesign and is not required for this OBC safety fix

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -1038,4 +1038,124 @@ mocha.describe('s3 worm', function() {
...
+            } catch (err) {
+                assert.strictEqual(err.rpc_code, 'UNAUTHORIZED');
+                assert.match(err.message, /Object Lock/);
+            }
+
+            const bucket_info = await rpc_client.bucket.read_bucket({ name: BKT_LEGAL });
+            const legal_bucket_name = bucket_info.name.unwrap ?
+                bucket_info.name.unwrap() : bucket_info.name;
+            assert.strictEqual(legal_bucket_name, BKT_LEGAL);
+        });
```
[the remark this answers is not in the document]
kajalpareek-lab: Agreed that would strengthen coverage. Current tests already cover reject + rollback; retry-after-unlock can be a follow-up.

src/server/object_services/object_server.js
kajalpareek-lab: I checked callers: `delete_multiple_objects_unordered` is only used by `buckets_reclaimer` for bucket wipe after OBC/`delete_bucket_and_objects`, not for normal S3 deletes. We intentionally fail the whole wipe if any object is still locked, rather than deleting unlocked objects and leaving the bucket Terminating with locked data. AWS has no equivalent wipe RPC; locked versions still prevent finishing bucket delete until they are gone. Happy to discuss skipping locked objects on reclaim if we want to change that product decision.

src/server/object_services/object_server.js
kajalpareek-lab: I left both checks. The bucket-level `EXISTS` check fails fast if any locked object exists anywhere in the bucket (including outside the current batch/`limit`). The batch check only covers the narrow race after that returns false. Keeping only the batch check could wipe unlocked objects while a locked object remains elsewhere, which weakens fail-closed reclaim.

src/server/object_services/object_server.js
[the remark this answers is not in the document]
kajalpareek-lab: Addressed the underlying concern. I did not make soft-delete fully atomic (no SELECT FOR UPDATE / lock predicate on the destructive update). After the deleting fence, new uploads fail in `load_bucket`. We still fail closed with a bucket-level lock check and a batch re-check of found objects before `remove_objects_and_unset_latest`. A transactional wipe would be a larger reclaim redesign and is not required for this OBC safety path.

src/server/object_services/object_server.js
kajalpareek-lab: I checked existing lock handling: S3 delete still uses inline `lock_settings` checks on this branch, and there is no shared helper here yet. I left `_is_object_lock_active` local to avoid expanding scope. Extracting/reusing with the S3 delete path is reasonable cleanup for a follow-up.

src/server/object_services/object_server.js
kajalpareek-lab: I left `UNAUTHORIZED` unchanged. This is the internal wipe RPC (OBC reclaim), not S3 `DeleteBucket`. It matches existing Object Lock object-delete handling in NooBaa (`UNAUTHORIZED`). AWS `DeleteBucket` when non-empty is typically `BucketNotEmpty`; deleting a locked object is `AccessDenied`. If you want a dedicated rpc_code for the operator, we can change it — but this is not an AWS S3 API surface as-is.

src/server/system_services/bucket_server.js
kajalpareek-lab: Addressed. The concern is valid: a failed lock check after fencing could leave the bucket stuck deleting. I wrapped the abort path so we always attempt to roll back this fence when we do not accept the delete (including when the lock check throws), and we still throw the original error. Unit coverage added.

src/test/unit_tests/internal/test_delete_bucket_object_lock.test.js
kajalpareek-lab: Addressed. Integration coverage is in `src/test/integration_tests/api/s3/test_s3_worm.js` under `delete_bucket_and_objects with Object Lock`.

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -1038,4 +1038,124 @@ mocha.describe('s3 worm', function() {
...
         });
     });
+
+    mocha.describe('delete_bucket_and_objects with Object Lock', function() {
+        // OBC wipe uses containerized delete_bucket_and_objects + MD store.
+        // NC coretest has no equivalent RPC path.
+        const BKT_RETENTION = 'worm-obc-delete-retention';
+        const BKT_LEGAL = 'worm-obc-delete-legalhold';
+        const BKT_PLAIN = 'worm-obc-delete-plain';
+        const KEY = 'obc-lock-obj';
```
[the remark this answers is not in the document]
kajalpareek-lab: I left this unchanged for this PR. It is valid test hygiene (fixed names can collide if teardown leaves locked buckets), but it is not a product bug in the Object Lock delete path. I will treat unique suffixes and clearer teardown as a follow-up.

src/server/bg_services/buckets_reclaimer.js
kajalpareek-lab: Addressed. You’re right  err already carries `rpc_code` and the Object Lock message. I removed the special `UNAUTHORIZED` log branch and kept the single existing error log for all reclaim failures.

src/server/system_services/bucket_server.js
```diff
@@ -1130,22 +1130,75 @@ async function update_buckets(req) {
...
+
+    if (!bucket.namespace) {
+        let has_locked_objects;
+        try {
+            has_locked_objects = await MDStore.instance().has_any_locked_objects_in_bucket(bucket._id);
+        } catch (err) {
+            // Lock check failed after fencing — roll back so the bucket is not left deleting.
+            dbg.error('delete_bucket_and_objects: lock check failed, rolling back deleting fence',
+                original_name, err);
+            try {
```
naveenpaul1, line 1178: could you please move this repeating try catch block inside a method

kajalpareek-lab: Addressed.

src/server/system_services/bucket_server.js
```diff
@@ -1058,22 +1058,75 @@ async function update_buckets(req) {
...
         }
     });
 
+    const rollback_delete_fence = async () => {
+        await system_store.make_changes({
+            update: {
+                buckets: [{
+                    $find: {
+                        _id: bucket._id,
+                        deleting: now,
```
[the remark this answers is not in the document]
kajalpareek-lab: Matching rollback only on `_id` + the same millisecond `Date` was not unique enough if two deletes raced.
I still set `deleting` as a `Date` for existing code, but each attempt now gets a unique rename (`bucket-deleting-<objectId>`). Rollback matches `_id` + that name, so one failed attempt can’t clear another attempt’s fence. Also updated the metrics helper that strips the `-deleting-…` suffix.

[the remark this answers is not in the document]
kajalpareek-lab: check latest commit.

src/server/system_services/bucket_server.js
nadavMiz: something still feels wrong about the overall approach. we mark the bucket for deletion, then undo this deletion if there is a locked object. why not do this check before marking the bucket for deletion?

for example I see there is the function can_delete_bucket that checks if the bucket is empty. can't we add a similer check before deleting the bucket in this case as well. you are changing delete_bucket_and_objects then maybe add a similar function like can_delete_bucket. that gate it before we mark the bucket for deletion? WDYT?

nadavMiz: I see that those are different functions. you are changing `delete_bucket_and_objects` then maybe add a similar function like can_delete_bucket. that gate it before we mark the bucket for deletion? WDYT?

PR #9912, opened 2026-08-06
NC | CLI | Phone home addition

jackyalbo:
### Describe the Problem
We want to start collecting noobaa usage statistics as part of scale phone home. We will do that by adding a new CLI command that can be run by phone home and also by the user if he wants to see what begin collected

### Explain the Changes
1. Add a new diagnose action `usage-stats` (`noobaa-cli diagnose usage-stats`) that scans ConfigFS and prints aggregate usage statistics
2. Report account counts (total, anonymous, root account managers, allow bucket creation, default connection, accounts with users)
3. Report IAM user total count
4. Report bucket total and feature adoption counts (versioning, lifecycle, notifications, logging, bucket policy, encryption, website, CORS, object lock, public access block, tags, force_md5_etag)
5. Report connection counts by protocol, plus NooBaa/config-dir version and hosts count
6. Add CLI help, response/error codes, docs, and an NC CLI integration test

1. 

### Testing Instructions:
1. Create accounts, IAM users, buckets with various features enabled, and a connection
2. Run:
```noobaa-cli diagnose usage-stats```
Verify the JSON reply contains only aggregate counts (no account/bucket names or secrets)

- [x] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added the `diagnose usage-stats` command for aggregate metrics covering accounts, users, buckets, bucket features, rules, and connections.
  * Added support for the optional `--config_root` option.
  * Added help text and documentation describing the command and output.

* **Bug Fixes**
  * Improved error handling when usage statistics cannot be collected.

* **Tests**
  * Verified aggregate results exclude resource names and credentials.

### DIFF
--- a/src/manage_nsfs/manage_nsfs_constants.js
+++ b/src/manage_nsfs/manage_nsfs_constants.js
@@ -32,7 +32,8 @@ const GLACIER_ACTIONS = Object.freeze({
 const DIAGNOSE_ACTIONS = Object.freeze({
     HEALTH: 'health',
     GATHER_LOGS: 'gather-logs',
-    METRICS: 'metrics'
+    METRICS: 'metrics',
+    USAGE_STATS: 'usage-stats'
 });
 
 const UPGRADE_ACTIONS = Object.freeze({
@@ -79,7 +80,8 @@ const VALID_OPTIONS_GLACIER = {
 const VALID_OPTIONS_DIAGNOSE = {
     'health': new Set([ 'https_port', 'deployment_type', 'all_account_details', 'all_bucket_details', 'all_connection_details', 'notif_storage_threshold', 'lifecycle', ...CLI_MUTUAL_OPTIONS]),
     'gather-logs': new Set([ CONFIG_ROOT_FLAG]),
-    'metrics': new Set([CONFIG_ROOT_FLAG])
+    'metrics': new Set([CONFIG_ROOT_FLAG]),
+    'usage-stats': new Set([CONFIG_ROOT_FLAG])
 };
 
 const VALID_OPTIONS_UPGRADE = {
--- a/src/manage_nsfs/manage_nsfs_constants.js
+++ b/src/manage_nsfs/manage_nsfs_constants.js
@@ -32,7 +32,8 @@ const GLACIER_ACTIONS = Object.freeze({
 const DIAGNOSE_ACTIONS = Object.freeze({
     HEALTH: 'health',
     GATHER_LOGS: 'gather-logs',
-    METRICS: 'metrics'
+    METRICS: 'metrics',
+    USAGE_STATS: 'usage-stats'
 });
 
 const UPGRADE_ACTIONS = Object.freeze({
@@ -79,7 +80,8 @@ const VALID_OPTIONS_GLACIER = {
 const VALID_OPTIONS_DIAGNOSE = {
     'health': new Set([ 'https_port', 'deployment_type', 'all_account_details', 'all_bucket_details', 'all_connection_details', 'notif_storage_threshold', 'lifecycle', ...CLI_MUTUAL_OPTIONS]),
     'gather-logs': new Set([ CONFIG_ROOT_FLAG]),
-    'metrics': new Set([CONFIG_ROOT_FLAG])
+    'metrics': new Set([CONFIG_ROOT_FLAG]),
+    'usage-stats': new Set([...CLI_MUTUAL_OPTIONS])
 };
 
 const VALID_OPTIONS_UPGRADE = {
--- a/src/manage_nsfs/usage_stats.js
+++ b/src/manage_nsfs/usage_stats.js
@@ -1,0 +1,191 @@
+/* Copyright (C) 2026 NooBaa */
+'use strict';
+
+const dbg = require('../util/debug_module')(__filename);
+const config = require('../../config');
+const pkg = require('../../package.json');
+const ManageCLIError = require('./manage_nsfs_cli_errors').ManageCLIError;
+const { ManageCLIResponse } = require('./manage_nsfs_cli_responses');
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by nimrod-becker** [comment]
@jackyalbo Question, speficially on lifecycle and notifications, which can contain many rules per single bucket
maybe save an array denoting # of such rules?
So for example, if I have 4 buckets:
Buck1 with 2 lifecycle rules, Buck2 no rules, Buck3 5 rules, Buck4 no rules
I would get
```
[2,0,5,0]
```
Same for notifications, later on transition rules, anything which is not a binary (on/off)
just need to make sure all such returned arrays are ordered in the same way (so arr[0] always refer to the same bucket)

Wdyt?

**Comment by jackyalbo** [comment]
> @jackyalbo Question, speficially on lifecycle and notifications, which can contain many rules per single bucket maybe save an array denoting # of such rules? So for example, if I have 4 buckets: Buck1 with 2 lifecycle rules, Buck2 no rules, Buck3 5 rules, Buck4 no rules I would get
> 
> ```
> [2,0,5,0]
> ```
> 
> Same for notifications, later on transition rules, anything which is not a binary (on/off) just need to make sure all such returned arrays are ordered in the same way (so arr[0] always refer to the same bucket)
> 
> Wdyt?

@nimrod-becker Added top-x for some features - take a look, say what you think

File: src/manage_nsfs/usage_stats.js [hunk 1]
```diff
@@ -0,0 +1,241 @@
...
+
+/**
+ * top_counts returns the highest counts in descending order, capped at limit.
+ * @param {number[]} counts
+ * @param {number} limit
+ * @returns {number[]}
+ */
+function top_counts(counts, limit) {
+    return counts
+        .slice()
```
alphaprinz: Why is this needed?

alphaprinz (approved):
Wasn't sure why the first slice() is needed in the top10 fund. Maybe I'm missing sth. Either way no a serious issue. Otherwise LGTM.

src/manage_nsfs/manage_nsfs_constants.js
```diff
@@ -79,7 +80,8 @@ const VALID_OPTIONS_GLACIER = {
 const VALID_OPTIONS_DIAGNOSE = {
     'health': new Set([ 'https_port', 'deployment_type', 'all_account_details', 'all_bucket_details', 'all_connection_details', 'notif_storage_threshold', 'lifecycle', ...CLI_MUTUAL_OPTIONS]),
     'gather-logs': new Set([ CONFIG_ROOT_FLAG]),
-    'metrics': new Set([CONFIG_ROOT_FLAG])
+    'metrics': new Set([CONFIG_ROOT_FLAG]),
+    'usage-stats': new Set([...CLI_MUTUAL_OPTIONS])
```
shirady: I'm not sure you need `CLI_MUTUAL_OPTIONS`, Do you use those flags?

shirady (commented):
LGTM

shirady:
@jackyalbo I think with the code changes of @alphaprinz and @sakshimunjal you would want to collect also IAN user inline policy and roles-related information (in a separate PR in the future).

src/manage_nsfs/usage_stats.js
```diff
@@ -0,0 +1,241 @@
...
+            accounts.total += 1;
+            if (account.name === config.ANONYMOUS_ACCOUNT_NAME) {
+                accounts.anonymous += 1;
+                continue;
+            }
+            if (account.iam_operate_on_root_account) accounts.root_account_managers += 1;
+            if (account.allow_bucket_creation) accounts.with_allow_bucket_creation += 1;
+            if (account.default_connection) accounts.with_default_connection += 1;
+
+            if (!account._id) continue;
```
romayalon, line 84: doesn't this mean we have an issue with the account? 
don't you want to have this recorded as well?

romayalon (commented):
Why do we need a new phone home CLI if we already have the health CLI? it's already going through accounts/buckets/connections and might be missing the properties count but very similar and actually I would expect to see in phone home some health info about buckets/account/other/endpoint exactly like we have in health CLI
also why IO stats usage are not collected?

jackyalbo:
> Why do we need a new phone home CLI if we already have the health CLI? it's already going through accounts/buckets/connections and might be missing the properties count but very similar and actually I would expect to see in phone home some health info about buckets/account/other/endpoint exactly like we have in health CLI also why IO stats usage are not collected?

I don't think we should be mixing the 2 - health will keep giving us data about the health of the system - more for debugging. phone-home will let us know how Nobbaa is being used by the customers, and the data can be aggregated between customers to see what features are used and how much - more of statistics collection.

src/manage_nsfs/usage_stats.js
jackyalbo: I don't think we want to find issues using this tool - we want to gather statistics.

src/manage_nsfs/usage_stats.js
jackyalbo: It was a protection on the array before sorting - not needed really- removing

src/manage_nsfs/manage_nsfs_constants.js
jackyalbo: I agree CONFIG_ROOT_FLAG should be enough - removing

romayalon:
@jackyalbo Not sure where this request was coming from, I checked what we have today in our existing stats_aggregator and came to this conclusion.
also, I know that on health script Scale were asking us to have a flag for not running buckets and accounts list in case there are lots of buckets/accounts and they didn't want us to do the whole list + read check
not sure how often this phone home is going to be called but put this thought in your mind

src/manage_nsfs/usage_stats.js
romayalon: we collect num_unhealthy_<pools/buckets/namespace_resources> on stats_aggregator

PR #9916, opened 2026-08-07
IAM Roles | Store roles in accounts with type=role instead of a separate iam_roles collection

aayushchouhan09:
### Describe the Problem
IAM roles lived in a separate `iam_roles` collection/schema, unlike NC’s single identity model, which made `role/account/user` handling inconsistent and harder to maintain.

### Explain the Changes
1. Store roles in accounts with type: 'role' (and fake email role/<name>:<owner>); remove `iam_role_schema` / `iam_roles` collection.
2. Point role CRUD/cache at accounts, exclude roles from account/user list APIs.
3. Migrate script changes.
4. Renamed `iam_inline_policies` (unified name) for containerized only.

// TODO (Gap - we will fix it in separate PR if required)
5. Enforce per-type required fields on account_schema via thin oneOf (account / user / role).

1.  EPIC: https://redhat.atlassian.net/browse/MCGI-330

### Testing Instructions:
1. Run the upgrade-script integration test:
`make run-single-test testpath=integration_tests/internal/test_upgrade_scripts.js`

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

- **New Features**
  - IAM roles now support descriptions, paths, creation dates, policies, and configurable session durations.
  - Role management and policy operations are integrated with account records.
  - Role identities are distinguished from accounts and users.
  - Role filtering by path is supported.
- **Bug Fixes**
  - Role names are handled consistently regardless of capitalization.
  - Deleted roles and roles are excluded from inappropriate listings.
  - Duplicate active role names are prevented within an owner account.
  - IAM role migrations now avoid creating duplicates when rerun.


src/server/system_services/schemas/account_schema.js
```diff
@@ -22,11 +22,17 @@ module.exports = {
         email: { wrapper: SensitiveString },
         is_support: { type: 'boolean' },
         is_external: { type: 'boolean' },
+        // optional identity kind: account | user | role
+        type: {
+            type: 'string',
+            enum: ['account', 'user', 'role'],
```
shirady: Raising it as a question:
What do you find more clear - 'account' or 'account root user'?
@aayushchouhan09 @naveenpaul1 @alphaprinz 
WDYT?

src/server/system_services/schemas/account_schema.js
shirady, line 1: @aayushchouhan09 @sakshimunjal i think you need to sync about the schema (`src/server/system_services/schemas/account_schema.js` and `src/server/system_services/schemas/nsfs_account_schema.js`).

src/server/system_services/schemas/account_schema.js
```diff
@@ -38,7 +44,27 @@ module.exports = {
...
+        },
+        max_session_duration: {
+            type: 'integer',
+            minimum: 3600,
+            maximum: 43200,
+        },
+        assume_role_policy_document: {
+            $ref: 'common_api#/definitions/iam_trust_policy_document',
+        },
+        iam_role_policies: {
```
shirady: Do you prefer to have `iam_role_policies` over reusing the `iam_user_policies` and renaming it?

It is confusing since it uses `iam_user_policy`.

src/server/system_services/schemas/account_schema.js
```diff
@@ -38,7 +44,27 @@ module.exports = {
...
+            type: 'array',
+            items: {
+                $ref: 'common_api#/definitions/iam_user_policy',
+            }
+        },
         creation_date: { idate: true },
+
```
shirady: redundant newline

src/sdk/nb.d.ts
```diff
@@ -92,26 +92,36 @@ interface Account extends Base {
     email: SensitiveString;
     next_password_change: Date;
     is_support?: boolean;
+    // optional identity kind: account | user | role
```
shirady: I don't think we need this comment, we can see that it is optional from `?`.
You can rename the `type` property to `identity_type` so it would be clearer.

src/sdk/nb.d.ts
```diff
@@ -92,26 +92,36 @@ interface Account extends Base {
...
+    type?: 'account' | 'user' | 'role';
```
shirady: Why lowercase for those constants?

I think you can add something like this:
`type IdentityType = 'ACCOUNT' | 'USER' | 'ROLE';`
and then use it:

```suggestion
    type?: IdentityType;
```

src/sdk/nb.d.ts
```diff
@@ -92,26 +92,36 @@ interface Account extends Base {
...
-    iam_path: string;
+    iam_path?: string;
+    iam_user_policies?: object[];
     description?: string;
     max_session_duration?: number;
-    assume_role_policy_document: object;
+    assume_role_policy_document?: object;
     iam_role_policies?: object[];
-    creation_date: Date;
+    creation_date?: Date;
```
shirady: You can this attribute `creation_date` to roles so we will have the information when the role was created.

src/sdk/nb.d.ts
```diff
@@ -92,26 +92,36 @@ interface Account extends Base {
...
+    owner: ID;
+};
+
+/** IAM role identity stored in accounts with type === 'role' */
+type IamRole = Account & {
+    type: 'role';
+    owner: ID;
+    assume_role_policy_document: object;
+};
+
```
shirady, line 124: Where do you use those `type IamUser` and `type IamRole`?

src/sdk/object_sdk.js
```diff
@@ -69,6 +69,8 @@ const account_cache = new LRUCache({
 const iam_roles_cache = new LRUCache({
     name: 'IamRolesCache',
     expiry_ms: config.IAM_ROLES_CACHE_EXPIRY_MS,
+    // Bound cache growth: keep enough entries for the per-account IAM role quota
```
shirady: I don't think there is a real connection between the `max_usage` and the quota of roles per account. so you can remove this comment.

The cache key seems like combination between `role_name` and `owner_account_id`, if we have 2 accounts with max roles each and they use all the roles, the cache would hold partial of the quota.

src/server/system_services/schemas/account_indexes.js
```diff
@@ -11,4 +11,17 @@ module.exports = [{
             deleted: null,
         }
     }
+}, {
+    // unique role name per owner when type === 'role' (roles stored as accounts)
```
shirady: I think you better not use code and just explanation in case of refactor, something like:
```suggestion
    // unique role name per account owner ID when the identity type is for IAM role (roles stored in the accounts table)
```

src/util/account_util.js
```diff
@@ -325,6 +325,18 @@ function get_account_email_from_username(username, requesting_account_id) {
...
 }
 
+function _get_role_name(role_name) {
+    return role_name instanceof SensitiveString ? role_name.unwrap() : role_name;
+}
+
+// To make the role name unique across system:
+// - first part is role name in lower case with role/ prefix
+// - second part is owner account id
+function get_account_email_from_role_name(role_name, owner_account_id) {
```
shirady, line 348: in AWS we can have in the account role and user with the same name?
This is way you had to create this function?

src/server/system_services/account_server.js
```diff
@@ -472,6 +465,8 @@ function list_accounts(req) {
     }
 
     const accounts = system_store.data.accounts
+        // IAM roles are stored in accounts with type=role - exclude from account list
+        .filter(account => account.type !== 'role')
```
shirady: I'm not sure about this addition, and if you need it - better add your filter last.
@naveenpaul1 worked on Users API in containrized.

src/server/system_services/account_server.js
```diff
@@ -495,7 +490,7 @@ function list_accounts(req) {
  */
 function accounts_status(req) {
     const any_non_support_account = _.find(system_store.data.accounts, function(account) {
-        return !account.is_support;
+        return !account.is_support && account.type !== 'role';
```
shirady: Same here, I'm not sure whether you need to add here.
@naveenpaul1

src/server/system_services/account_server.js
```diff
@@ -1262,6 +1257,7 @@ function _verify_can_delete_account(req, account_to_delete) {
     }
     if (account_to_delete.owner === undefined) {
         const has_iam_users = _.some(system_store.data.accounts, function(account) {
+            if (account.type === 'role' || account.deleted) return false;
```
shirady: It seems not related.
`account_to_delete.owner === undefined` is account root user, so it should not be role, I think you can remove this change.

src/server/system_services/account_server.js
```diff
@@ -1452,6 +1371,7 @@ async function list_users(req) {
     // TODO: Pagination not supported - currently returns all users, ignoring marker and max_items params
     const is_truncated = false;
     const requesting_account_iam_users = _.filter(system_store.data.accounts, function(account) {
+        if (account.type === 'role' || account.deleted) return false;
```
shirady: I'm not sure it is better for maintenance the code, what if someone would add more types?
then we need to remember to edit this line?
It is prone to bugs.

shirady (commented):
I did partial review -
I didn't review all the changes `account_server` (Suddenly I had the message "Diff is too big to render. To view, check out this pull request locally.")

src/sdk/nb.d.ts
aayushchouhan09: Updated, thanks

src/sdk/nb.d.ts
aayushchouhan09: Done, Thanks

src/sdk/nb.d.ts
aayushchouhan09: We have used `type IamRole` in some function jsdoc.
Removed `IamUser` for now.

src/sdk/object_sdk.js
aayushchouhan09: removed, thanks

src/server/system_services/schemas/account_schema.js
aayushchouhan09: here `ACCOUNT` means root account identity, while `USER` and `ROLE` are IAM identities. We kept `ACCOUNT` for consistency and we can add a short clarifying comment I think.

src/server/system_services/schemas/account_schema.js
aayushchouhan09: We should keep `iam_role_policies` as a separate field so user and role data stay explicit and easy to read.

If needed, we can rename `iam_user_policy` to a neutral shared name like `iam_inline_policy`. wdyt?

src/util/account_util.js
aayushchouhan09: Yes, AWS allows a user and a role with the same name in the same account, since they are different identity types.
That helper exists to avoid key collisions in our unified accounts collection by prefixing role keys (`role/...`).

src/server/system_services/account_server.js
aayushchouhan09: updated.

src/server/system_services/account_server.js
aayushchouhan09: updated.

src/server/system_services/account_server.js
aayushchouhan09: This check applies to each iterated account in system_store.data.accounts, not to account_to_delete itself.
We still need to exclude role identities here, otherwise roles may be incorrectly counted as IAM users.

src/server/system_services/account_server.js
aayushchouhan09: I moved this to shared identity helpers instead of inline type checks.
If new types are added, we update one place only.

src/api/common_api.js
```diff
@@ -117,6 +117,11 @@ module.exports = {
             enum: ['DISABLED', 'SUSPENDED', 'ENABLED']
         },
 
+        identity_type: {
```
sakshimunjal, line 120: will make this field name consistent in NC. I had used 'type', will rename to 'identity_type'

src/api/common_api.js
```diff
@@ -117,6 +117,11 @@ module.exports = {
...
+            type: 'string',
+            enum: ['ACCOUNT', 'USER', 'ROLE']
```
naveenpaul1, line 122: are we planning to add type for users too?

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
```diff
@@ -57,8 +56,12 @@ async function run({ dbg, system_store, system_server }) {
             const max_session_duration = DEFAULT_MAX_SESSION_DURATION_SECS;
             const new_role = _.omitBy({
                 _id: system_store.new_system_store_id(),
+                identity_type: 'ROLE',
                 owner: account._id,
-                name: role_config.role_name,
+                name: new SensitiveString(role_config.role_name),
+                email: role_email,
```
naveenpaul1, line 59: are we adding email only becuse of its required field, right?

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
```diff
@@ -37,10 +38,8 @@ async function run({ dbg, system_store, system_server }) {
             const role_config = account.role_config;
             const new_policy = {};
 
-            const iam_role = system_store.data.iam_roles.find(
-                role => role.name === role_config.role_name &&
-                role.owner?._id?.toString() === account._id.toString()
-            );
+            const role_email = account_util.get_account_email_from_role_name(role_config.role_name, account._id.toString());
+            const iam_role = system_store.get_account_by_email(role_email);
```
naveenpaul1: can we use role name here for searching? that make more. 
Will it also return account/user with same email? In that case get role_emil only if iam_role is empty

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
```diff
@@ -57,8 +56,12 @@ async function run({ dbg, system_store, system_server }) {
             const max_session_duration = DEFAULT_MAX_SESSION_DURATION_SECS;
             const new_role = _.omitBy({
                 _id: system_store.new_system_store_id(),
+                identity_type: 'ROLE',
                 owner: account._id,
-                name: role_config.role_name,
+                name: new SensitiveString(role_config.role_name),
```
naveenpaul1, line 58: why role name is SensitiveString? It should be normal name, right?

src/util/account_util.js
```diff
@@ -325,6 +326,51 @@ function get_account_email_from_username(username, requesting_account_id) {
     return new SensitiveString(`${username.toLowerCase()}:${requesting_account_id}`);
 }
 
+function _get_role_name(role_name) {
```
naveenpaul1, line 341: Like I mention above role_name should be simple string

src/util/account_util.js
naveenpaul1: In that case we save role ARN, this tring looks similer to ARN to me

src/util/account_util.js
```diff
@@ -325,6 +326,51 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+    return new SensitiveString(`role/${role_name_str.toLowerCase()}:${owner_account_id}`);
+}
+
+function _get_identity_type(account) {
+    const identity_type = account.identity_type || account.type;
+    if (identity_type) return String(identity_type).toUpperCase();
+    if (!_.isUndefined(account.assume_role_policy_document) || !_.isUndefined(account.iam_role_policies)) {
+        return 'ROLE';
+    }
+    return account.owner === undefined ? 'ACCOUNT' : 'USER';
```
naveenpaul1: Its better to assign a const for ACCOUNT', 'USER' and 'ROLE'

src/util/account_util.js
```diff
@@ -325,6 +326,51 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+// - second part is owner account id
+function get_account_email_from_role_name(role_name, owner_account_id) {
+    const role_name_str = _get_role_name(role_name);
+    return new SensitiveString(`role/${role_name_str.toLowerCase()}:${owner_account_id}`);
+}
+
+function _get_identity_type(account) {
+    const identity_type = account.identity_type || account.type;
+    if (identity_type) return String(identity_type).toUpperCase();
+    if (!_.isUndefined(account.assume_role_policy_document) || !_.isUndefined(account.iam_role_policies)) {
```
naveenpaul1: I think we can have user without assume_role_policy_document and iam_role_policies?

src/util/account_util.js
```diff
@@ -325,6 +326,51 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+
+// To make the role name unique across system:
+// - first part is role name in lower case with role/ prefix
+// - second part is owner account id
+function get_account_email_from_role_name(role_name, owner_account_id) {
+    const role_name_str = _get_role_name(role_name);
+    return new SensitiveString(`role/${role_name_str.toLowerCase()}:${owner_account_id}`);
+}
+
+function _get_identity_type(account) {
```
naveenpaul1: its better to rename `account` variable to identity or something similar

src/server/system_services/account_server.js
```diff
@@ -1757,37 +1678,117 @@ async function list_user_policies(req) {
...
+    }
+    return iam_role;
+}
+
+/**
+ * validates role creation constraints for the target account
+ * @param {string} role_name
+ * @param {string|nb.ID} owner_account_id
+ */
+function _check_create_role_preconditions(role_name, owner_account_id) {
```
naveenpaul1: can you check the account deleted or not?

src/server/system_services/account_server.js
```diff
@@ -1757,37 +1678,117 @@ async function list_user_policies(req) {
...
+    if (_list_active_iam_roles_for_account(owner_account_id).length >= MAX_NUMBER_OF_IAM_ROLES) {
+        throw new RpcError('LIMIT_EXCEEDED',
+            `Cannot exceed quota for RolesPerAccount: ${MAX_NUMBER_OF_IAM_ROLES}.`);
+    }
+    if (_get_iam_role_account_by_name(role_name, owner_account_id)) {
```
naveenpaul1: Method name is kind of confusing, if ok can we have something like
_get_iam_role_by_name_and_owner_id()

src/server/system_services/account_server.js
```diff
@@ -1757,37 +1678,117 @@ async function list_user_policies(req) {
...
+/**
+ * build account_api role_info response object
+ * @param {nb.IamRole} iam_role
+ * @param {string} account_id
+ * @returns {object}
+ */
+function _return_iam_role_info(iam_role, account_id) {
+    const role_name = account_util._get_role_name(iam_role.name);
+    const owner_account = iam_role.owner?.access_keys ? iam_role.owner :
+        system_store.data.get_by_id(iam_role.owner?._id || iam_role.owner);
```
naveenpaul1: why we need to fetch the account again? we already have access_keys from owner, If owner id `undefined` or owner got deleted from account schema there is no way you can get a account in owner_account variable, they we do DB query again

src/server/system_services/schemas/nsfs_account_schema.js
```diff
@@ -24,6 +24,9 @@ module.exports = {
         creation_date: {
             type: 'string',
         },
+        identity_type: {
```
sakshimunjal: had to add same change in my PR as well since there were some related changes. Either you can remove or if any conflict comes, please pay attention in case my PR is merged first, else I will take care

src/util/account_util.js
```diff
@@ -325,6 +326,51 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+    return _get_identity_type(account) === 'USER';
+}
+
+function _list_iam_roles_by_owner(owner_id) {
+    const owner_id_str = owner_id.toString();
+    return (system_store.data.accounts || []).filter(account =>
+        _is_role_identity(account) &&
+        account.owner &&
+        get_owner_account_id(account) === owner_id_str
+    );
```
sakshimunjal, line 385: do we need to check account deleted similar to _list_active_iam_roles_for_account?

src/util/account_util.js
```diff
@@ -325,6 +326,51 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+}
+
+function _list_iam_roles_by_name(role_name) {
+    return (system_store.data.accounts || []).filter(account =>
+        _is_role_identity(account) &&
+        _get_role_name(account.name) === role_name
+    );
```
sakshimunjal, line 396: same, is account deleted check required?

src/api/common_api.js
aayushchouhan09: we set it for all three: `create_account` writes `ACCOUNT/USER` based on owner, and `create_role` (plus the upgrade script) writes `ROLE`.

src/server/system_services/schemas/account_schema.js
aayushchouhan09: renaming it to `iam_inline_policies` as per [this comment](https://github.com/noobaa/noobaa-core/pull/9947#discussion_r3802082541)

src/server/system_services/schemas/nsfs_account_schema.js
aayushchouhan09: Sure, I will check and update accordingly.

src/server/system_services/account_server.js
aayushchouhan09: `owner_account_id` here is the authenticated root account, and deleted-role entries are already filtered by the helpers, so an extra deleted-owner check in this precondition is not required. Thanks

src/server/system_services/account_server.js
aayushchouhan09: done, thanks

src/server/system_services/account_server.js
aayushchouhan09: We re-resolve owner because create/update flows can have owner as an ID (without access_keys), `get_by_id` here is an in-memory system_store lookup (not DB), and safely returns undefined if owner is missing/deleted.

src/util/account_util.js
aayushchouhan09: I think we should keep this custom string for internal lookup and uniqueness, because ARN is an external display format (includes path/format details) and using it as DB key would be complex from storage to API format.

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
aayushchouhan09: We should keep get_account_by_email(role_email) here because it gives deterministic O(1) lookup with owner-scoped, type-safe key (`role/<name>:<owner>`), whereas role-name search needs scan + owner filtering and yes it is ambiguity-prone in the unified accounts collection.

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
aayushchouhan09: Even in upgrade, we keep name as `SensitiveString` to match how accounts data is stored everywhere else.
Using plain string only here can create inconsistent role records.

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
aayushchouhan09: Yes, `email` is required by the unified accounts schema, and for roles we use a derived internal email key (owner + normalized role name) for uniqueness and fast lookup.

src/util/account_util.js
aayushchouhan09: this helper only keeps backward-safe handling for existing wrapped (`SensitiveString`) values in current data/call paths.

src/util/account_util.js
aayushchouhan09: done, thanks

src/util/account_util.js
aayushchouhan09: fixed, Thanks

src/util/account_util.js
aayushchouhan09: done, thanks

src/util/account_util.js
aayushchouhan09: No, `_list_iam_roles_by_owner()` is intentionally a generic owner-role helper, while deleted filtering is applied only in `_list_active_iam_roles_for_account()` where active-role behavior is needed.

src/util/account_util.js
aayushchouhan09: No, it was intentional. thanks

src/api/account_api.js
shirady, line 1616: Now that you have the `account_info` and it is in the same schema, why do you need the `role_info`?

src/api/common_api.js
shirady: Current accounts saved in the DB don't have this enum (`['ACCOUNT', 'USER', 'ROLE']`) and the existing code also assumes that if we have `owner` defined it is a user - what is tha plan for what we had?
Are you sure you need this? I think @sakshimunjal mentioned that didn't use it.

src/sdk/nb.d.ts
```diff
@@ -92,25 +93,28 @@ interface Account extends Base {
...
     access_keys: Array<{
         access_key: SensitiveString;
         secret_key: SensitiveString;
     }>;
     master_key_id: ID;
+    iam_path?: string;
+    iam_inline_policies?: object[];
+    description?: string;
+    max_session_duration?: number;
+    assume_role_policy_document?: object;
```
shirady: Can you add a comment so it would be clear that those fields are only relevant for roles?

src/sdk/nb.d.ts
```diff
@@ -92,25 +93,28 @@ interface Account extends Base {
...
+    creation_date?: Date;
+    deleted?: Date;
 }
 
-interface IamRole extends Base {
-    _id: ID;
+/** IAM role identity stored in accounts with identity_type === 'ROLE' */
+type IamRole = Account & {
```
shirady, line 118: Do we use it?
If so, it means that we miss `IamUser`?

src/sdk/nb.d.ts
shirady: I'm not sure you need to remove, we just need to understand where it is used.

src/server/system_services/schemas/account_indexes.js
```diff
@@ -11,4 +11,17 @@ module.exports = [{
             deleted: null,
         }
     }
+}, {
+    // ensure unique role name per owner for role identities stored in accounts
```
shirady, line 15: Why do we need this new index?
Does it mean that we have a GAP for index in users?
cc: @naveenpaul1

src/server/system_services/schemas/account_schema.js
```diff
@@ -22,22 +22,38 @@ module.exports = {
         email: { wrapper: SensitiveString },
         is_support: { type: 'boolean' },
         is_external: { type: 'boolean' },
+        // optional identity kind: ACCOUNT | USER | ROLE
```
shirady: I think you can remove this comment, we can read this from the code.

src/server/system_services/schemas/account_schema.js
```diff
@@ -22,22 +22,38 @@ module.exports = {
...
-        iam_user_policies: {
+        iam_inline_policies: {
             type: 'array',
             items: {
-                $ref: 'common_api#/definitions/iam_user_policy',
+                $ref: 'common_api#/definitions/iam_inline_policy',
             }
         },
+
+        // IAM role fields (identity_type === 'ROLE')
```
shirady: I think this comment is hard to maintain (it is not clear where is the scope of it).
Maybe you can mention it inline somewhere else.

src/server/system_services/account_server.js
```diff
@@ -472,6 +465,8 @@ function list_accounts(req) {
     }
 
     const accounts = system_store.data.accounts
+        // IAM roles are stored in accounts, exclude them from account list
+        .filter(account => !account_util._is_role_identity(account))
```
shirady, line 469: Do we use this for users as well?
I didn't understand if we have a bug or that we used the same function
I'm not sure that you need to exclude it - as I think it tries to list the documents that we have in the accounts table.

src/server/system_services/account_server.js
```diff
@@ -495,7 +490,7 @@ function list_accounts(req) {
  */
 function accounts_status(req) {
     const any_non_support_account = _.find(system_store.data.accounts, function(account) {
-        return !account.is_support;
+        return !account.is_support && !account_util._is_role_identity(account);
```
shirady, line 493: Again, I'm not sure you need to exclude a role in the `accounts_status` if you can explain.

src/server/system_services/account_server.js
```diff
@@ -1749,45 +1670,122 @@ async function list_user_policies(req) {
...
+    }
+    return iam_role;
+}
+
+/**
+ * validates role creation constraints for the target account
+ * @param {string} role_name
+ * @param {string|nb.ID} owner_account_id
+ */
+function _check_create_role_preconditions(role_name, owner_account_id) {
```
shirady: I see that you use this function once, are you sure you need it?

src/server/system_services/account_server.js
```diff
@@ -1749,45 +1670,122 @@ async function list_user_policies(req) {
...
+    }
+}
+
+/**
+ * build account_api role_info response object
+ * @param {nb.IamRole} iam_role
+ * @param {string} account_id
+ * @returns {object}
+ */
+function _return_iam_role_info(iam_role, account_id) {
```
shirady, line 1733: Where do you need this function?

src/server/system_services/account_server.js
shirady: I also find myself confused from this.
@aayushchouhan09 if it is possible to make it easier to understand.

src/server/system_services/account_server.js
```diff
@@ -1749,45 +1670,122 @@ async function list_user_policies(req) {
...
-        name: role_name,
-        iam_path: iam_path,
+        name: new SensitiveString(role_name),
+        email: account_util.get_account_email_from_role_name(role_name, account_id),
+        has_login: false,
+        access_keys: [],
+        iam_path: req.rpc_params.iam_path || IAM_DEFAULT_PATH,
         description: req.rpc_params.description,
-        max_session_duration: max_session_duration,
+        max_session_duration: req.rpc_params.max_session_duration ?? DEFAULT_MAX_SESSION_DURATION_SECS,
```
shirady, line 1781: I think you need to fix the places where you default the DEFAULT_MAX_SESSION_DURATION_SECS in the reply - for example:
https://github.com/noobaa/noobaa-core/blob/fcb149fca74527264dd1d601a1de857260ed9311/src/endpoint/iam/ops/iam_get_role.js#L32

src/server/system_services/account_server.js
```diff
@@ -1749,45 +1670,122 @@ async function list_user_policies(req) {
...
+        iam_inline_policies: [],
         creation_date: Date.now(),
     }, _.isUndefined);
-
-    await system_store.make_changes({
-        insert: {
-            iam_roles: [new_role],
-        }
-    });
+    await system_store.make_changes({ insert: { accounts: [new_role] } });
```
shirady: Why did you move to one line? I think in code style of `system_store.make_changes` we call it with multi lines.

src/server/system_services/account_server.js
```diff
@@ -1796,38 +1794,25 @@ async function get_role(req) {
     const requesting_account = req.account;
     account_util._check_if_requesting_account_is_root_account(action, requesting_account,
         { role_name: req.rpc_params.role_name, path: IAM_DEFAULT_PATH }, 'ROLE');
-
-    const role_name = req.rpc_params.role_name;
     const account_id = String(requesting_account._id);
-    const account_roles = _list_active_iam_roles_for_account(account_id);
-    const iam_role = _get_iam_role_by_name_or_throw(account_roles, role_name);
-    return _return_iam_role_info(iam_role, account_id);
+    return _return_iam_role_info(
```
shirady, line 1800: I think you can reuse the `get_account_info` instead of new function.

src/server/system_services/account_server.js
```diff
@@ -1796,38 +1794,25 @@ async function get_role(req) {
...
-
     if (!_.isEmpty(updates)) {
         await system_store.make_changes({
-            update: {
-                iam_roles: [{
-                    _id: role_to_update._id,
-                    $set: updates,
-                }]
-            }
+            update: { accounts: [{ _id: role_to_update._id, $set: updates }] }
```
shirady: Same here - 
Why did you move to one line? I think in code style of `system_store.make_changes` we call it with multi lines.

src/server/system_services/account_server.js
```diff
@@ -1837,75 +1822,49 @@ async function delete_role(req) {
...
+    if ((role_to_delete.iam_inline_policies || []).length > 0) {
         account_util._throw_error_delete_conflict(action, role_to_delete, 'policies');
     }
-
-    await system_store.make_changes({
-        remove: {
-            iam_roles: [role_to_delete._id],
-        }
-    });
+    await system_store.make_changes({ remove: { accounts: [role_to_delete._id] } });
```
shirady: Same here - 
Why did you move to one line? I think in code style of `system_store.make_changes` we call it with multi lines.

src/server/system_services/account_server.js
```diff
@@ -1837,75 +1822,49 @@ async function delete_role(req) {
...
-    const role_name = req.rpc_params.role_name;
-    const account_id = String(requesting_account._id);
-    const account_roles = _list_active_iam_roles_for_account(account_id);
-    const role_to_delete = _get_iam_role_by_name_or_throw(account_roles, role_name);
-
-    const iam_role_policies = role_to_delete.iam_role_policies || [];
-    if (iam_role_policies.length > 0) {
+    const role_to_delete = _get_iam_role_by_name_or_throw(
+        req.rpc_params.role_name, requesting_account._id);
+    if ((role_to_delete.iam_inline_policies || []).length > 0) {
```
shirady: I'm not sure if it was done on purpose, I think 2 lines were fine -
```suggestion
    const iam_inline_policies = role_to_delete.iam_inline_policies || [];
    if (iam_inline_policies.length > 0) {
```

src/server/system_services/account_server.js
```diff
@@ -1837,75 +1822,49 @@ async function delete_role(req) {
...
+        members = _.filter(members, role =>
+            (role.iam_path || IAM_DEFAULT_PATH).startsWith(req.rpc_params.iam_path_prefix));
     }
-    members = members.sort((a, b) => a.name.localeCompare(b.name));
-    members = members.map(role => _return_iam_role_info(role, account_id));
-    return { members, is_truncated };
+    members = members.sort((a, b) =>
+        account_util._get_role_name(a.name).localeCompare(account_util._get_role_name(b.name)))
+        .map(role => _return_iam_role_info(role, account_id));
+    return { members, is_truncated: false };
```
shirady: I don't see a reason for changing this, it seems fine as it was.

src/server/system_services/account_server.js
```diff
@@ -1837,75 +1822,49 @@ async function delete_role(req) {
...
-    }
-
-    const iam_role_policies = [...(role_to_update.iam_role_policies || [])];
-    const policy_index = account_util._get_iam_policy_index(iam_role_policies, req.rpc_params.policy_name);
-    const iam_role_policy_to_add = {
+    const role_to_update = _get_iam_role_by_name_or_throw(
+        req.rpc_params.role_name, requesting_account._id);
+    const iam_inline_policies = [...(role_to_update.iam_inline_policies || [])];
+    const policy_index = account_util._get_iam_policy_index(iam_inline_policies, req.rpc_params.policy_name);
+    const policy = {
```
shirady: I think the name `policy` is short and might be confusing for others that don't have the knowledge that you have and vaguely know about role policy and assume role policy, etc.

src/server/system_services/account_server.js
```diff
@@ -1837,75 +1822,49 @@ async function delete_role(req) {
...
     };
-    if (policy_index === -1) {
-        iam_role_policies.push(iam_role_policy_to_add);
-    } else {
-        iam_role_policies[policy_index] = iam_role_policy_to_add;
-    }
-
-    account_util._check_total_policy_size(iam_role_policies, req.rpc_params.role_name, 'role');
+    if (policy_index === -1) iam_inline_policies.push(policy);
+    else iam_inline_policies[policy_index] = policy;
```
shirady: it is less readable, can you not change it and have it with brackets?

```suggestion
    if (policy_index === -1) {
        iam_inline_policies.push(policy);
    } else {
        iam_role_policies[policy_index] = policy;
    }
```

src/server/system_services/account_server.js
```diff
@@ -1837,75 +1822,49 @@ async function delete_role(req) {
...
+    account_util._check_total_policy_size(iam_inline_policies, req.rpc_params.role_name, 'role');
     await system_store.make_changes({
-        update: {
-            iam_roles: [{
-                _id: role_to_update._id,
-                $set: { iam_role_policies },
-            }]
-        }
+        update: { accounts: [{ _id: role_to_update._id, $set: { iam_inline_policies } }] }
```
shirady: Same here - 
Why did you move to one line? I think in code style of `system_store.make_changes` we call it with multi lines.

src/server/system_services/account_server.js
```diff
@@ -1932,25 +1890,14 @@ async function delete_role_policy(req) {
...
+        action, iam_inline_policies, req.rpc_params.policy_name, 'role');
+    iam_inline_policies.splice(policy_index, 1);
     await system_store.make_changes({
-        update: {
-            iam_roles: [{
-                _id: role_to_delete._id,
-                $set: { iam_role_policies },
-            }]
-        }
+        update: { accounts: [{ _id: role_to_delete._id, $set: { iam_inline_policies } }] }
```
shirady: Same here - 
Why did you move to one line? I think in code style of `system_store.make_changes` we call it with multi lines.

src/server/system_services/account_server.js
```diff
@@ -1959,33 +1906,26 @@ async function list_role_policies(req) {
...
     account_util._check_if_requesting_account_is_root_account(action, requesting_account,
         { role_name: req.rpc_params.role_name, path: IAM_DEFAULT_PATH }, 'ROLE');
-
     // TODO: Pagination not supported - currently returns all role policies, ignoring marker and max_items params
-    const account_roles = _list_active_iam_roles_for_account(requesting_account._id);
-    const requested_role = _get_iam_role_by_name_or_throw(account_roles, req.rpc_params.role_name);
-    const members = _.map(requested_role.iam_role_policies || [], iam_role_policy => iam_role_policy.policy_name)
+    const requested_role = _get_iam_role_by_name_or_throw(
+        req.rpc_params.role_name, requesting_account._id);
+    const members = _.map(requested_role.iam_inline_policies || [], p => p.policy_name)
```
shirady: Can you avoid using `p` as the variable name?

src/server/system_services/account_server.js
```diff
@@ -1959,33 +1906,26 @@ async function list_role_policies(req) {
...
         .sort((a, b) => a.localeCompare(b));
-    return {
-        is_truncated: false,
-        members,
-    };
+    return { is_truncated: false, members };
```
shirady: Suggesting to avoid uneeded refactoring in this PR, we want to be focus on the major changes and this PR has already many code line changes.

src/server/system_services/account_server.js
```diff
@@ -1959,33 +1906,26 @@ async function list_role_policies(req) {
...
+        req.rpc_params.role_name, requesting_account._id);
     await system_store.make_changes({
         update: {
-            iam_roles: [{
+            accounts: [{
                 _id: role_to_update._id,
-                $set: {
-                    assume_role_policy_document: req.rpc_params.policy_document,
-                },
+                $set: { assume_role_policy_document: req.rpc_params.policy_document },
```
shirady: Same here - 
Why did you move to one line? I think in code style of `system_store.make_changes` we call it with multi lines.

src/util/account_util.js
```diff
@@ -33,6 +39,7 @@ async function create_account(req) {
             system_store.parse_system_store_id(req.rpc_params.new_system_parameters.account_id) :
             system_store.new_system_store_id()
         ),
+        identity_type: req.rpc_params.owner ? IDENTITY_TYPES.USER : IDENTITY_TYPES.ACCOUNT,
```
shirady: You can reuse it for roles as well.
I guess you would to change this condition.

src/util/account_util.js
shirady: But you can add the type to the JSdoc so it would be clear that you expect this function to be used only with strings.
Do you have a case where you pass `SensitiveString` type?

src/util/account_util.js
```diff
@@ -325,6 +332,61 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+// To make the role name unique across system:
+// - first part is role name in lower case with role/ prefix
+// - second part is owner account id
+function get_account_email_from_role_name(role_name, owner_account_id) {
+    const role_name_str = _get_role_name(role_name);
+    return new SensitiveString(`role/${role_name_str.toLowerCase()}:${owner_account_id}`);
+}
+
+function _get_identity_type(identity) {
+    const identity_type = identity.identity_type || identity.type;
```
shirady: Why do you need `identity.type`?

src/util/account_util.js
```diff
@@ -325,6 +332,61 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+    if (identity_type) return String(identity_type).toUpperCase();
+    const identity_email = _get_identity_email(identity);
+    if ((!_.isUndefined(identity.assume_role_policy_document) || !_.isUndefined(identity.iam_inline_policies)) &&
+        (_.isUndefined(identity_email) || _is_role_identity_email(identity_email))) {
```
shirady: We try to avoid using `_.isUndefined`.

src/util/account_util.js
shirady: @aayushchouhan09 
In which case do you need to have the a deleted role in the list?

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
shirady, line 1: @aayushchouhan09 I didn't have the chance to write upgrade script, suggesting to have a reviewer with experience on this part.
Can you also add testing instruction for this?

shirady, line 1: @aayushchouhan09 where is the migration script or part that is related to IAM users?
You change the property name of `iam_user_policies`.

shirady (commented):
I added questions and comments -
I think missing in the upgrade script the change that you did in the account schema `iam_user_policies` -> `iam_inline_policies`.
What will be the case for existing users saved in the accounts table?

src/sdk/nb.d.ts
aayushchouhan09: added, thanks

src/sdk/nb.d.ts
aayushchouhan09: Yes, we use IamRole in typed/JSDoc IAM role flows, and I added IamUser too for symmetry in the unified identity model.

src/sdk/nb.d.ts
aayushchouhan09: Added and used, thanks

src/server/system_services/schemas/account_indexes.js
aayushchouhan09: We need this index for roles to enforce per-owner role-name uniqueness after roles moved into the accounts collection.

src/server/system_services/account_server.js
aayushchouhan09: After the refactor, roles are physically stored in that same collection, so without filtering, ListAccounts would start returning IAM roles too, which changes API behavior and breaks separation between `ListAccounts` and `ListRoles`.

src/server/system_services/account_server.js
aayushchouhan09: If we don’t exclude roles in accounts_status, the status can become misleading: roles alone can make `has_accounts` look true even when no real account/root user exists.

src/server/system_services/account_server.js
aayushchouhan09: we can remove it. Thanks

src/server/system_services/account_server.js
aayushchouhan09: `get_account_info()` is shared and account-oriented (used by read_account, list_accounts, and system_server) and returns a different shape than IAM role APIs.

`role_info` has a separate contract (role_id, role_name, assume_role_policy_document, owner_access_key, iam_role_policies), so reusing get_account_info() would still require an extra role-mapping layer and broader refactor; keeping get_account_info + _return_iam_role_info is the minimal and safer path.

src/server/system_services/account_server.js
aayushchouhan09: We use _return_iam_role_info as the single formatter for role API responses (read_role_by_name, create_role, get_role, and list_roles) so all role endpoints return the same role_info shape.

src/server/system_services/account_server.js
aayushchouhan09: `owner` can arrive either as a populated account object or just an owner ID, so we normalize both cases to one owner_account before reading owner_access_key.
`get_by_id` here is only an in-memory `system_store` lookup (not a DB query), and it safely returns undefined if the owner is missing.

src/server/system_services/account_server.js
aayushchouhan09: done, thanks

src/server/system_services/account_server.js
aayushchouhan09: update, thanks

src/server/system_services/account_server.js
aayushchouhan09: updated, thanks

src/server/system_services/account_server.js
aayushchouhan09: updated, Thanks

src/server/system_services/account_server.js
aayushchouhan09: updated.

src/server/system_services/account_server.js
aayushchouhan09: updated, thanks

src/server/system_services/account_server.js
aayushchouhan09: updated, Thanks

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
aayushchouhan09: updated and added tests. Thanks

src/util/account_util.js
aayushchouhan09: done, Thanks

src/util/account_util.js
aayushchouhan09: Yes, there are real `SensitiveString` call paths (`iam_role.name / account.name` from stored role records).   I’ll clarify it with JSDoc: `@param {string|SensitiveString} role_name`.

src/util/account_util.js
aayushchouhan09: removed.

src/util/account_util.js
aayushchouhan09: We now filter deleted entries in `_list_iam_roles_by_owner()`, so it returns only active roles.

sakshimunjal (approved):
LGTM

src/sdk/nb.d.ts
shirady: I think better have it inline comment as some are mutual to users.

```suggestion
    description?: string; // role-only fields for identity_type ROLE
    max_session_duration?: number; // role-only fields for identity_type ROLE
```

src/server/system_services/schemas/account_schema.js
```diff
@@ -38,6 +40,18 @@ module.exports = {
                 $ref: 'common_api#/definitions/iam_inline_policy',
             }
         },
+
+        description: { // role-only
+            type: 'string',
+        },
+        max_session_duration: {
```
shirady:
```suggestion
        max_session_duration: { //role-only
```

src/util/account_util.js
```diff
@@ -20,19 +20,29 @@ const demo_access_keys = Object.freeze({
     access_key: new SensitiveString('123'),
     secret_key: new SensitiveString('abc')
 });
+
+const IDENTITY_TYPES = Object.freeze({
```
shirady, line 24: Just making sure that you are aware that there is also:
https://github.com/noobaa/noobaa-core/blob/62356ae384400dde5689ea01f5850705260c8494/src/util/native_fs_utils.js#L24-L31

src/util/account_util.js
```diff
@@ -20,19 +20,29 @@ const demo_access_keys = Object.freeze({
...
+});
 /**
  *
  * CREATE_ACCOUNT
  *
  */
 async function create_account(req) {
+    let identity_type = IDENTITY_TYPES.ACCOUNT;
+    if (req.rpc_params.owner) identity_type = IDENTITY_TYPES.USER;
+    if (req.rpc_params.role_config) identity_type = IDENTITY_TYPES.ROLE;
```
shirady: I think it would be more redable to have the condition nested:
if (req.rpc_params.owner)
... identify by role and set IDENTITY_TYPES.USER
(else or without) set to IDENTITY_TYPES.ROLE

Do you still use `req.rpc_params.role_config`?

src/util/account_util.js
```diff
@@ -325,6 +335,60 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+// To make the role name unique across system:
+// - first part is role name in lower case with role/ prefix
+// - second part is owner account id
+function get_account_email_from_role_name(role_name, owner_account_id) {
+    const role_name_str = _get_role_name(role_name);
+    return new SensitiveString(`role/${role_name_str.toLowerCase()}:${owner_account_id}`);
+}
+
+function _get_identity_type(identity) {
+    const identity_type = identity.identity_type;
```
shirady: I think you want to check if it included in `IDENTITY_TYPES`.

src/util/account_util.js
```diff
@@ -325,6 +335,60 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+    if (identity_type) return String(identity_type).toUpperCase();
+    if (!_.isUndefined(identity.assume_role_policy_document)) {
```
shirady: I think we should avoid using `_.isUndefined`.
I has to fix the code in the past - https://github.com/noobaa/noobaa-core/pull/8267

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
```diff
@@ -26,26 +26,36 @@ function unwrap_principal(principals) {
...
+            if (!_.isUndefined(account.iam_user_policies)) {
+                const update = {
+                    _id: account._id,
+                    $unset: { iam_user_policies: 1 },
+                };
+                if (_.isUndefined(account.iam_inline_policies)) {
+                    update.$set = { iam_inline_policies: account.iam_user_policies };
+                }
+                policy_migration_updates.push(update);
+            }
```
shirady: I'm not sure, but can we use $rename? (instead of $unset and $set)

src/test/integration_tests/internal/test_upgrade_scripts.js
```diff
@@ -678,27 +665,22 @@ mocha.describe('test upgrade_iam_role script 5.23.0', async function() {
             'role_config must be unset from the account after a successful migration');
     });
 
-    mocha.it('re-running the script inserts exactly one more iam_role per qualifying account', async function() {
-        // The upgrade script does not guard against duplicate inserts.
-        // Running it twice produces exactly one additional role per qualifying account.
+    mocha.it('re-running the script does not duplicate role accounts for qualifying account', async function() {
+        // Verify idempotency for already migrated accounts.
```
shirady: What is this comment?

src/test/integration_tests/internal/test_upgrade_scripts.js
```diff
@@ -896,4 +873,50 @@ mocha.describe('test upgrade_iam_role script 5.23.0', async function() {
...
+
+    mocha.it('legacy iam_user_policies are migrated to iam_inline_policies for existing IAM users', async function() {
+        const owner_account = await _insert_account_with_role_config(role_account_legacy_user_owner, undefined);
+        const iam_user_params = {
+            name: role_account_legacy_user_policies,
+            email: role_account_legacy_user_policies,
+            owner: owner_account._id.toString(),
+            has_login: false,
+            s3_access: true,
+            default_resource: process.env.NC_CORETEST ? 's3_bucket_policy_nsr' : POOL_LIST[1].name,
```
shirady, line 885: why do you need the `process.env.NC_CORETEST`?

src/api/common_api.js
shirady: @aayushchouhan09 I see this still open

src/server/system_services/account_server.js
shirady: I think that In containerized the owner should be the account object.
Where did you see it was just ID?

src/util/account_util.js
```diff
@@ -325,6 +335,60 @@ function get_account_email_from_username(username, requesting_account_id) {
...
+
+// To make the role name unique across system:
+// - first part is role name in lower case with role/ prefix
+// - second part is owner account id
+function get_account_email_from_role_name(role_name, owner_account_id) {
+    const role_name_str = _get_role_name(role_name);
+    return new SensitiveString(`role/${role_name_str.toLowerCase()}:${owner_account_id}`);
+}
+
+function _get_identity_type(identity) {
```
shirady, line 358: You can add JSDoc to the new function

src/util/account_util.js
shirady: @aayushchouhan09 I see it is still open

src/test/integration_tests/internal/test_upgrade_scripts.js
aayushchouhan09: Not required and clear, removed.

src/test/integration_tests/internal/test_upgrade_scripts.js
aayushchouhan09: We follow the existing suite pattern here, so I kept the inline NC switch for `default_resource`.

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
aayushchouhan09: We intentionally use conditional `$set + unconditional $unset` (instead of `$rename`) because it safely handles by preserving existing `iam_inline_policies`, migrating only when missing, and always removing legacy `iam_user_policies` so no stale field remains.

src/util/account_util.js
aayushchouhan09: `native_fs_utils` uses an error-entity enum, while this `IDENTITY_TYPES` constant is for account identity typing, so they serve different purposes and shouldn’t be merged.

src/util/account_util.js
aayushchouhan09: We can remove the roles part here.

src/util/account_util.js
aayushchouhan09: Thanks

src/upgrade/upgrade_scripts/5.23.0/upgrade_iam_roles.js
```diff
@@ -26,26 +26,36 @@ function unwrap_principal(principals) {
...
 
     try {
-        dbg.log0('Starting migration of roles from account schema to role schema...');
+        dbg.log0('Starting IAM role migration from account role_config entries...');
         const new_roles = [];
         const migrated_account_ids = [];
+        const policy_migration_updates = [];
 
         for (const account of system_store.data.accounts) {
+            // migrate legacy iam_user_policies to iam_inline_policies for identities in accounts
```
naveenpaul1: I think this section nothing to do with role upgrade, it is more like migrate user policy `iam_user_policies` to a common name  `iam_inline_policies`, Should be a different upgrade script.

aayushchouhan09: Created new upgrade script file for users. Thanks

naveenpaul1 (approved):
LGTM

src/server/system_services/account_server.js
aayushchouhan09: I will create a new helper to resolve the owner for better code readability. It just act as a safer fallback.

PR #9918, opened 2026-08-08
allow removing GOVERNANCE retention with empty Retention body

kajalpareek-lab:
## Summary
- [DFBUGS-8949](https://redhat.atlassian.net/browse/DFBUGS-8949)
- PutObjectRetention with `--retention '{}'` (empty `<Retention/>` XML) crashed with `TypeError: Cannot read properties of undefined (reading '0')` which surfaced as `InternalError` to the client.
- Root cause: the handler unconditionally accessed `.Mode[0]` and `.RetainUntilDate[0]` without null checks, and had no codepath for clearing retention.
- Fix adds safe property access, an "empty Retention = clear retention" path at the S3 endpoint, namespace_fs, and object_server layers, and makes the `retention` RPC param optional.

## Tested on live cluster

Verified on a remote IBM Cloud OpenShift cluster (`shirshfe-5ibm05.ibmcloud2.qe.rh-ocs.com`, amd64, RHCOS 10.2) by hot-patching JS files into the running endpoint pod:

| # | Test | Expected | Result |
|---|------|----------|--------|
| 1 | Clear GOVERNANCE retention with `--bypass-governance-retention --retention '{}'` | Success (was InternalError before fix) | **PASS** |
| 2 | `get-object-retention` after clearing | `NoSuchObjectLockConfiguration` | **PASS** |
| 3 | Re-apply GOVERNANCE retention (normal set path) | Success | **PASS** |
| 4 | Verify retention was set correctly | Shows GOVERNANCE mode + date | **PASS** |
| 5 | Clear retention WITHOUT bypass (GOVERNANCE active) | `AccessDenied` | **PASS** |
| 6 | Clear retention WITH bypass | Success | **PASS** |
| 7 | Confirm retention removed | `NoSuchObjectLockConfiguration` | **PASS** |

## Test plan

- [x] `aws s3api put-object-retention --retention '{}' --bypass-governance-retention` on a GOVERNANCE-locked object succeeds
- [x] `get-object-retention` after clearing returns `NoSuchObjectLockConfiguration`
- [x] Clearing without `--bypass-governance-retention` on active GOVERNANCE retention returns `AccessDenied`
- [x] Normal set-retention flow (Mode + RetainUntilDate) still works
- [ ] Clearing COMPLIANCE retention (even with bypass) returns `AccessDenied`
- [ ] Legal hold is preserved when retention is cleared
- [ ] Unit/integration tests in `test_s3_worm.js` pass

## Summary by CodeRabbit

* **New Features**
  * Added support for clearing object retention using an empty retention request.
  * Added authorization checks when clearing, shortening, or changing retention settings.
  * Preserved legal holds when clearing retention and improved retention mode transitions.
* **Bug Fixes**
  * Improved handling of missing or empty retention fields.
  * Prevented invalid partial retention requests while maintaining existing date validation.
* **Tests**
  * Added coverage for clearing governance retention, bypass authorization, and switching retention modes.

### DIFF
--- a/src/endpoint/s3/ops/s3_put_object_retention.js
+++ b/src/endpoint/s3/ops/s3_put_object_retention.js
@@ -10,26 +10,31 @@ const s3_utils = require('../s3_utils');
 async function put_object_retention(req) {
     // TODO: may require at the future Content-MD5 support
     if (!req.body.Retention) throw new S3Error(S3Error.MalformedXML);
-    const mode = req.body.Retention.Mode[0];
-    let retain_until_date = req.body.Retention.RetainUntilDate[0];
-    if (!mode && !retain_until_date) throw new S3Error(S3Error.AccessDenied);
-    if (!mode || !retain_until_date) throw new S3Error(S3Error.MalformedXML);
-    retain_until_date = new Date(req.body.Retention.RetainUntilDate[0]);
 
-    const bypass_governance = req.headers['x-amz-bypass-governance-retention'] && req.headers['x-amz-bypass-governance-retention'].toUpperCase() === 'TRUE';
+    const bypass_governance = req.headers['x-amz-bypass-governance-retention'] &&
+        req.headers['x-amz-bypass-governance-retention'].toUpperCase() === 'TRUE';
 
-    if (s3_utils._is_valid_retention(mode, retain_until_date)) {
-        await req.object_sdk.put_object_retention({
-            bucket: req.params.bucket,
-            key: req.params.key,
-            version_id: s3_utils.parse_version_id(req.query.versionId),
-            bypass_governance,
-            retention: {
-                mode,
-                retain_until_date,
-            }
-        });
+    const mode = req.body.Retention.Mode && req.body.Retention.Mode[0];
+    const retain_until_date_str = req.body.Retention.RetainUntilDate && req.body.Retention.RetainUntilDate[0];
+
+    let retention;
+    if (!mode && !retain_until_date_str) {
+        retention = {};
+    } else if (!mode || !retain_until_date_str) {
+        throw new S3Error(S3Error.MalformedXML);
+    } else {
+        const retain_until_date = new Date(retain_until_date_str);
+        if (!s3_utils._is_valid_retention(mode, retain_until_date)) return;
+        retention = { mode, retain_until_date };
     }
+
+    await req.object_sdk.put_object_retention({
+        bucket: req.params.bucket,
+        key: req.params.key,
+        version_id: s3_utils.parse_version_id(req.query.versionId),
+        bypass_governance,
+        retention,
+    });
 }
 
 module.exports = {
--- a/src/sdk/namespace_fs.js
+++ b/src/sdk/namespace_fs.js
@@ -84,8 +84,10 @@ const XATTR_DIR_CONTENT = XATTR_NOOBAA_INTERNAL_PREFIX + 'dir_content';
 const XATTR_NON_CURRENT_TIMESTASMP = XATTR_NOOBAA_INTERNAL_PREFIX + 'non_current_timestamp';
 const XATTR_TAG = XATTR_NOOBAA_INTERNAL_PREFIX + 'tag.';
 const XATTR_LEGAL_HOLD = XATTR_NOOBAA_INTERNAL_PREFIX + 'legal_hold';
-const XATTR_RETENTION_MODE = XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_mode';
-const XATTR_RETENTION_DATE = XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_date';
+// prefix used by set_fs_xattr_op to clear all retention xattrs by prefix
+const XATTR_RETENTION_PREFIX = XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_';
+const XATTR_RETENTION_MODE = XATTR_RETENTION_PREFIX + 'mode';
+const XATTR_RETENTION_DATE = XATTR_RETENTION_PREFIX + 'date';
 const HIDDEN_VERSIONS_PATH = '.versions';
 const NULL_VERSION_ID = 'null';
 const NULL_VERSION_SUFFIX = '_' + NULL_VERSION_ID;
@@ -2399,24 +2401,29 @@ class NamespaceFS {
     /**
      * check if the object retention lock settings can be updated to the new retention settings. if not, will throw AccessDenied error
      * the rules are:
-     * 1. if the new retention is longer than the current retention, it can be updated (can increase retention time)
-     * 2. if the new retention is shorter than the current retention, it cannot be updated and will throw error, unless the user has bypass_governance permission and the current retention mode is GOVERNANCE
+     * 1. if new_retention is omitted/empty, retention is being cleared — same bypass rules as delete protection
+     * 2. if the new retention is longer than the current retention, it can be updated (can increase retention time)
+     * 3. if the new retention is shorter than the current retention, it cannot be updated and will throw error, unless the user has bypass_governance permission and the current retention mode is GOVERNANCE
      * @param {Object} current_retention - current object retention lock settings
-     * @param {Object} new_retention - new object retention lock settings
+     * @param {Object} [new_retention] - new object retention lock settings; omit/empty to clear retention
      * @param {boolean} bypass_governance - if true, and user has permission to use this flag, will allow to bypass governance mode retention lock. compliance mode retention lock cannot be bypassed.
      * @throws {S3Error.AccessDenied} if the object is protected by object lock and the user does not have permission to bypass the lock
      */
     _compare_object_retention(fs_context, current_retention, new_retention, bypass_governance) {
-        if (current_retention) {
-            const retain_until_date = new Date(current_retention.retain_until_date);
-            const new_date = new Date(new_retention.retain_until_date);
-            //can always increase retention time when mode is unchanged
-            if (new_date >= retain_until_date && new_retention.mode === current_retention.mode) return;
-            bypass_governance = bypass_governance && this._has_bypass_governance_permission(fs_context);
-            if (current_retention.mode === 'COMPLIANCE' ||
-                (current_retention.mode === 'GOVERNANCE' && !bypass_governance)) {
-                throw new S3Error(S3Error.AccessDeniedObjectLocked);
-            }
+        if (!current_retention) return;
+        // empty retention object means clear — enforce bypass rules before removing
+        if (!new_retention.mode && !new_retention.retain_until_date) {
+            this._check_object_retention(fs_context, current_retention, bypass_governance);
+            return;
+        }
+        const retain_until_date = new Date(current_retention.retain_until_date);
+        const new_date = new Date(new_retention.retain_until_date);
+        //can always increase retention time when mode is unchanged
+        if (new_date >= retain_until_date && new_retention.mode === current_retention.mode) return;
+        bypass_governance = bypass_governance && this._has_bypass_governance_permission(fs_context);
+        if (current_retention.mode === 'COMPLIANCE' ||
+            (current_retention.mode === 'GOVERNANCE' && !bypass_governance)) {
+            throw new S3Error(S3Error.AccessDeniedObjectLocked);
         }
     }
 
@@ -2518,14 +2525,19 @@ class NamespaceFS {
         const fs_context = this.prepare_fs_context(object_sdk);
         const file_path = await this._find_version_path(fs_context, params, true);
         await this._check_path_in_bucket_boundaries(fs_context, file_path);
+        const is_clear = !params.retention.mode && !params.retention.retain_until_date;
         try {
             const stat = await nb_native().fs.stat(fs_context, file_path);
             const current_retention = this._get_retention_mode_from_xattr(stat.xattr);
             this._compare_object_retention(fs_context, current_retention, params.retention, params.bypass_governance);
-            const fs_xattr = {};
-            fs_xattr[XATTR_RETENTION_MODE] = params.retention.mode;
-            fs_xattr[XATTR_RETENTION_DATE] = params.retention.retain_until_date.toISOString();
-            await this.set_fs_xattr_op(fs_context, file_path, fs_xattr, undefined);
+            if (is_clear) {
+                await this.set_fs_xattr_op(fs_context, file_path, undefined, XATTR_RETENTION_PREFIX);
+            } else {
+                const fs_xattr = {};
+                fs_xattr[XATTR_RETENTION_MODE] = params.retention.mode;
+                fs_xattr[XATTR_RETENTION_DATE] = params.retention.retain_until_date.toISOString();
+                await this.set_fs_xattr_op(fs_context, file_path, fs_xattr, undefined);
+            }
         } catch (err) {
             dbg.error(`NamespaceFS.put_object_retention: failed for file ${file_path} with error: `, err);
             throw native_fs_utils.translate_error_codes(err, native_fs_utils.entity_enum.OBJECT);
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -193,7 +193,6 @@ async function get_object_tagging(req) {
     };
 }
 
-
 /**
  *
  * delete_object_tagging
@@ -346,17 +345,30 @@ async function put_object_retention(req) {
     if (!req.bucket.object_lock_configuration || req.bucket.object_lock_configuration.object_lock_enabled !== 'Enabled') {
         throw new RpcError('INVALID_REQUEST');
     }
-    const current_retention = info.lock_settings?.retention;
     const new_retention = req.rpc_params.retention;
+    const is_clear = !new_retention.mode && !new_retention.retain_until_date;
+    const current_retention = info.lock_settings?.retention;
     _throw_if_retention_update_forbidden({
         key: obj.key,
         obj_id: info.obj_id,
         current_retention,
-        new_retention,
+        new_retention: is_clear ? undefined : new_retention,
         bypass_governance: Boolean(req.rpc_params.bypass_governance),
     });
     const legal_hold_status = info.lock_settings?.legal_hold?.status;
     const legal_hold = legal_hold_status ? { status: legal_hold_status } : undefined;
+    if (is_clear) {
+        if (legal_hold) {
+            await MDStore.instance().update_object_by_id(
+                obj._id, { lock_settings: { legal_hold } }, undefined, undefined
+            );
+        } else {
+            await MDStore.instance().update_object_by_id(
+                obj._id, undefined, { lock_settings: 1 }, undefined
+            );
+        }
+        return;
+    }
     await MDStore.instance().update_object_by_id(
         obj._id, {
             lock_settings: {
@@ -495,7 +507,6 @@ async function _complete_multipart_upload(req) {
     if (req.rpc_params.last_modified_time) {
         set_updates.last_modified_time = new Date(req.rpc_params.last_modified_time);
     }
-
 
     await _put_object_handle_latest_with_retries({ req, put_obj: obj, set_updates, unset_updates });
 
@@ -646,7 +657,6 @@ async function _complete_simple_upload(req) {
     };
 }
 
-
 async function update_bucket_counters({ system, bucket_name, content_type, read_count, write_count }) {
     const bucket = system.buckets_by_name[bucket_name.unwrap()];
     if (!bucket || bucket.deleting) return;
@@ -658,8 +668,6 @@ async function update_bucket_counters({ system, bucket_name, content_type, read_
         write_count,
     });
 }
-
-
 
 /**
  *
@@ -839,7 +847,6 @@ async function get_mapping(req) {
     return { chunks: res_chunks.map(chunk => chunk.to_api()) };
 }
 
-
 /**
  *
  * PUT_MAPPING
@@ -928,7 +935,6 @@ async function read_object_mapping(req) {
         chunks: chunks.map(chunk => chunk.to_api()),
     };
 }
-
 
 /**
  *
--- a/src/test/integration_tests/api/s3/test_s3_worm.js
+++ b/src/test/integration_tests/api/s3/test_s3_worm.js
@@ -566,6 +566,68 @@ mocha.describe('s3 worm', function() {
         });
     });
 
+    mocha.describe('clear GOVERNANCE retention with empty Retention', function() {
+        const CLEAR_RET_KEY = 'clear-retention-test';
+        let clear_ret_version_id;
+
+        mocha.it('should create object with GOVERNANCE retention', async function() {
+            const shortDate = new Date();
+            shortDate.setSeconds(shortDate.getSeconds() + 60);
+            const res = await s3_owner.putObject({
+                Bucket: BKT1,
+                Key: CLEAR_RET_KEY,
+                Body: file_body,
+                ContentType: 'text/plain',
+                ObjectLockMode: 'GOVERNANCE',
+                ObjectLockRetainUntilDate: shortDate
+            });
+            clear_ret_version_id = res.VersionId;
+            assert.ok(res.VersionId);
+        });
+
+        mocha.it('should fail to clear retention without bypass flag', async function() {
+            await assert_throws_async(s3_owner.putObjectRetention({
+                Bucket: BKT1,
+                Key: CLEAR_RET_KEY,
+                VersionId: clear_ret_version_id,
+                Retention: {},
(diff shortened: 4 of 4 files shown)


src/endpoint/s3/ops/s3_put_object_retention.js
```diff
@@ -6,17 +6,34 @@ const s3_utils = require('../s3_utils');
...
-    const bypass_governance = req.headers['x-amz-bypass-governance-retention'] && req.headers['x-amz-bypass-governance-retention'].toUpperCase() === 'TRUE';
+    const bypass_governance = req.headers['x-amz-bypass-governance-retention'] &&
+        req.headers['x-amz-bypass-governance-retention'].toUpperCase() === 'TRUE';
+
+    // Safe access: Mode/RetainUntilDate may be absent when clearing retention
+    const mode = req.body.Retention.Mode && req.body.Retention.Mode[0];
+    const retain_until_date_str = req.body.Retention.RetainUntilDate && req.body.Retention.RetainUntilDate[0];
+
+    // Empty Retention clears object retention (AWS / MinIO compatible).
+    if (!mode && !retain_until_date_str) {
```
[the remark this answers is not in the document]
kajalpareek-lab: Valid point — the S3 endpoint already validates partial payloads (throws MalformedXML at line 35 of s3_put_object_retention.js), so this path cannot be reached through normal S3 requests. However, namespace_fs could be called via RPC without that validation layer. Added a defensive MalformedXML check before accessing .toISOString() to prevent a potential TypeError from internal callers.

src/server/object_services/object_server.js
```diff
@@ -349,26 +349,55 @@ async function put_object_retention(req) {
...
+            // Shortening always requires bypass (existing behavior).
+            (new Date(new_retention.retain_until_date) < new Date(current_retention.retain_until_date));
+        if (needs_bypass_check) {
+            if ((current_retention.mode === 'GOVERNANCE' && (!req.rpc_params.bypass_governance || req.role !== 'admin')) ||
+                current_retention.mode === 'COMPLIANCE') {
+                dbg.error('put object retention failed due object retention mode', obj);
+                throw new RpcError('UNAUTHORIZED');
+            }
         }
     }
```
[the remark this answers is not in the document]
kajalpareek-lab: Correct — this was a real gap. The NSFS path (line 2422 of namespace_fs.js) allows skipping bypass only when both date is not shortened AND mode is unchanged. The object_server path was only comparing dates, meaning a COMPLIANCE→GOVERNANCE downgrade with an equal/later date would bypass the guard entirely. Fixed by adding the mode-change condition to needs_bypass_check, aligning both backends.

src/sdk/namespace_fs.js
```diff
@@ -2524,14 +2529,20 @@ class NamespaceFS {
...
-            fs_xattr[XATTR_RETENTION_DATE] = params.retention.retain_until_date.toISOString();
-            await this.set_fs_xattr_op(fs_context, file_path, fs_xattr, undefined);
+            if (is_clear) {
+                // Remove retention xattrs by prefix
+                await this.set_fs_xattr_op(fs_context, file_path, undefined, XATTR_RETENTION_PREFIX);
+            } else {
+                const fs_xattr = {};
+                fs_xattr[XATTR_RETENTION_MODE] = params.retention.mode;
+                fs_xattr[XATTR_RETENTION_DATE] = params.retention.retain_until_date.toISOString();
+                await this.set_fs_xattr_op(fs_context, file_path, fs_xattr, undefined);
```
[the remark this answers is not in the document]
kajalpareek-lab: The S3 endpoint already blocks partial payloads (MalformedXML at line 35), so this can't be reached via normal S3 requests. But namespace_fs could be called internally via RPC without that guard. Added a MalformedXML throw before .toISOString() to prevent a potential TypeError.

src/server/object_services/object_server.js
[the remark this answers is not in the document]
kajalpareek-lab: Acknowledged as a correct finding with the fix applied.

src/endpoint/s3/ops/s3_put_object_retention.js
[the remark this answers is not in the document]
kajalpareek-lab: Good catch — XATTR_RETENTION_PREFIX was used but never declared. Defined it as XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_' (matching both retention_mode and retention_date), and refactored XATTR_RETENTION_MODE/DATE to derive from it. This keeps the prefix clear path working without touching legal_hold.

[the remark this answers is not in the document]
kajalpareek-lab: Added now.

src/api/object_api.js
```diff
@@ -1298,14 +1298,15 @@ module.exports = {
                 required: [
                     'key',
                     'bucket',
-                    'retention'
```
nadavMiz, line 1318: even if the retention is an empty object. wouldn't you still need to have a retention?

src/api/object_api.js
```diff
@@ -1298,14 +1298,15 @@ module.exports = {
...
                 ],
                 properties: {
                     bucket: { $ref: 'common_api#/definitions/bucket_name' },
                     key: {
                         type: 'string',
                     },
                     version_id: { type: 'string' },
+                    // Omit retention (or pass empty) to clear object retention.
```
nadavMiz: I don't think this comment is necessary.

src/endpoint/s3/ops/s3_put_object_retention.js
```diff
@@ -6,17 +6,34 @@ const s3_utils = require('../s3_utils');
 
 /**
  * https://docs.aws.amazon.com/AmazonS3/latest/API/API_PutObjectRetention.html
+ *
+ * Empty Retention (no Mode / RetainUntilDate) removes object retention when
+ * allowed (GOVERNANCE + x-amz-bypass-governance-retention:true).
  */
 async function put_object_retention(req) {
-    // TODO: may require at the future Content-MD5 support
```
nadavMiz, line 11: is removing this comment on purpose? why?

src/endpoint/s3/ops/s3_put_object_retention.js
```diff
@@ -6,17 +6,34 @@ const s3_utils = require('../s3_utils');
...
-    let retain_until_date = req.body.Retention.RetainUntilDate[0];
-    if (!mode && !retain_until_date) throw new S3Error(S3Error.AccessDenied);
-    if (!mode || !retain_until_date) throw new S3Error(S3Error.MalformedXML);
-    retain_until_date = new Date(req.body.Retention.RetainUntilDate[0]);
 
-    const bypass_governance = req.headers['x-amz-bypass-governance-retention'] && req.headers['x-amz-bypass-governance-retention'].toUpperCase() === 'TRUE';
+    const bypass_governance = req.headers['x-amz-bypass-governance-retention'] &&
+        req.headers['x-amz-bypass-governance-retention'].toUpperCase() === 'TRUE';
+
+    // Safe access: Mode/RetainUntilDate may be absent when clearing retention
```
nadavMiz: again I think this comment is redundent

src/endpoint/s3/ops/s3_put_object_retention.js
```diff
@@ -6,17 +6,34 @@ const s3_utils = require('../s3_utils');
...
+    const mode = req.body.Retention.Mode && req.body.Retention.Mode[0];
+    const retain_until_date_str = req.body.Retention.RetainUntilDate && req.body.Retention.RetainUntilDate[0];
+
+    // Empty Retention clears object retention (AWS / MinIO compatible).
+    if (!mode && !retain_until_date_str) {
+        await req.object_sdk.put_object_retention({
```
nadavMiz: I don't think a seconds call for put_object_retention is a good idea. why not set the parameters differently for different cases? you can just assign retention to different values, is missing empty object, if not empty the actual object

src/sdk/namespace_fs.js
```diff
@@ -84,8 +84,9 @@ const XATTR_DIR_CONTENT = XATTR_NOOBAA_INTERNAL_PREFIX + 'dir_content';
 const XATTR_NON_CURRENT_TIMESTASMP = XATTR_NOOBAA_INTERNAL_PREFIX + 'non_current_timestamp';
 const XATTR_TAG = XATTR_NOOBAA_INTERNAL_PREFIX + 'tag.';
 const XATTR_LEGAL_HOLD = XATTR_NOOBAA_INTERNAL_PREFIX + 'legal_hold';
-const XATTR_RETENTION_MODE = XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_mode';
-const XATTR_RETENTION_DATE = XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_date';
+const XATTR_RETENTION_PREFIX = XATTR_NOOBAA_INTERNAL_PREFIX + 'retention_';
```
nadavMiz, line 88: why change this

src/sdk/namespace_fs.js
```diff
@@ -2524,14 +2530,23 @@ class NamespaceFS {
...
             this._compare_object_retention(fs_context, current_retention, params.retention, params.bypass_governance);
-            const fs_xattr = {};
-            fs_xattr[XATTR_RETENTION_MODE] = params.retention.mode;
-            fs_xattr[XATTR_RETENTION_DATE] = params.retention.retain_until_date.toISOString();
-            await this.set_fs_xattr_op(fs_context, file_path, fs_xattr, undefined);
+            if (is_clear) {
+                // Remove retention xattrs by prefix
+                await this.set_fs_xattr_op(fs_context, file_path, undefined, XATTR_RETENTION_PREFIX);
+            } else {
+                if (!params.retention.mode || !params.retention.retain_until_date) {
```
nadavMiz: is this possible? wouldn't it already be blocked by schema / s3 layer

nadavMiz:
can you add tests to test_s3_worm.js?

src/api/object_api.js
kajalpareek-lab: You're right — keeping retention required is cleaner. Changed to always pass retention: {} for the clear case instead of omitting it.

src/api/object_api.js
kajalpareek-lab: Agreed, removed.

src/endpoint/s3/ops/s3_put_object_retention.js
kajalpareek-lab: Not intentional — restored the TODO. It's unrelated to this fix.

src/endpoint/s3/ops/s3_put_object_retention.js
kajalpareek-lab: Fair, removed.

src/endpoint/s3/ops/s3_put_object_retention.js
kajalpareek-lab: Agreed — unified to a single call. Retention is set to {} for clear, or {mode, retain_until_date} for set.

src/sdk/namespace_fs.js
kajalpareek-lab: XATTR_RETENTION_PREFIX is needed by the clear path (line 2539) — set_fs_xattr_op uses it to strip both retention_mode and retention_date xattrs by prefix. Without it declared, that path crashes with ReferenceError. Deriving MODE/DATE from PREFIX ensures consistency and avoids a separate string that could drift. The final xattr values are unchanged (user.noobaa.retention_mode, user.noobaa.retention_date).

src/sdk/namespace_fs.js
kajalpareek-lab: You're right — the S3 layer rejects partial payloads before reaching here, and the schema enforces retention is present. Removed the guard.

src/api/object_api.js
kajalpareek-lab: Restored 'retention' to required array

src/api/object_api.js
kajalpareek-lab: Removed

src/endpoint/s3/ops/s3_put_object_retention.js
kajalpareek-lab: Removed.

src/endpoint/s3/ops/s3_put_object_retention.js
kajalpareek-lab: Unified to single call with retention = {} or {mode, retain_until_date}

src/sdk/namespace_fs.js
kajalpareek-lab: XATTR_RETENTION_PREFIX is needed by clear path, would crash without it

src/sdk/namespace_fs.js
kajalpareek-lab: Removed the defensive MalformedXML guard

src/endpoint/s3/ops/s3_put_object_retention.js
kajalpareek-lab: Removed it — Content-MD5 is not required for PutObjectRetention per the S3 spec. It's an optional integrity header that SDKs calculate automatically, but AWS does not reject requests without it. The TODO was misleading since it implied it might become required, when it's actually just an optional body checksum validation.

kajalpareek-lab:
> can you add tests to test_s3_worm.js?

Added.

src/endpoint/s3/ops/s3_put_object_retention.js
nadavMiz: the comment is still removed. I though we already talked about keeping it. has something changed?

src/endpoint/s3/ops/s3_put_object_retention.js
```diff
@@ -8,28 +8,31 @@ const s3_utils = require('../s3_utils');
...
-            retention: {
-                mode,
-                retain_until_date,
-            }
-        });
+    const mode = req.body.Retention.Mode && req.body.Retention.Mode[0];
+    const retain_until_date_str = req.body.Retention.RetainUntilDate && req.body.Retention.RetainUntilDate[0];
+
+    let retention;
+    if (!mode && !retain_until_date_str) {
```
nadavMiz, line 21: just for readability. I think it would be better to do it like:
    if (!mode && !retain_until_date_str) {
        ...
    else if (!mode || !retain_until_date_str) {
          throw new S3Error(S3Error.MalformedXML);
    } else {
         ....
    }

WDYT?

src/sdk/namespace_fs.js
nadavMiz: I still don't understand why the prefix is needed. they are still called by XATTR_RETENTION_MODE and XATTR_RETENTION_DATE. the prefix is never used. why would separate string drift. and even if they do. why does it matters

src/sdk/namespace_fs.js
```diff
@@ -2399,24 +2400,28 @@ class NamespaceFS {
...
-            const new_date = new Date(new_retention.retain_until_date);
-            //can always increase retention time when mode is unchanged
-            if (new_date >= retain_until_date && new_retention.mode === current_retention.mode) return;
-            bypass_governance = bypass_governance && this._has_bypass_governance_permission(fs_context);
-            if (current_retention.mode === 'COMPLIANCE' ||
-                (current_retention.mode === 'GOVERNANCE' && !bypass_governance)) {
-                throw new S3Error(S3Error.AccessDeniedObjectLocked);
-            }
+        if (!current_retention) return;
+        if (!new_retention.mode && !new_retention.retain_until_date) {
```
nadavMiz, line 2415: can you add a comment. that this is the empty object scenario

src/sdk/namespace_fs.js
nadavMiz: ahh its so you can clear them by prefix. can you add a comment explaining this

src/api/object_api.js
```diff
@@ -1315,7 +1315,7 @@ module.exports = {
                 required: [
                     'key',
                     'bucket',
-                    'retention'
+                    'retention',
```
shirady: Can you remove the extra `.`?
(to have less files in "Files changed" list).

src/endpoint/s3/ops/s3_put_object_retention.js
```diff
@@ -8,28 +8,31 @@ const s3_utils = require('../s3_utils');
...
+        retention = {};
+    } else {
+        if (!mode || !retain_until_date_str) throw new S3Error(S3Error.MalformedXML);
+        const retain_until_date = new Date(retain_until_date_str);
+        if (!s3_utils._is_valid_retention(mode, retain_until_date)) return;
```
shirady, line 27: Is that on purpose? To return if the retention is not valid?

src/sdk/namespace_fs.js
```diff
@@ -2399,24 +2400,28 @@ class NamespaceFS {
...
+            this._check_object_retention(fs_context, current_retention, bypass_governance);
```
shirady, line 2416: Why did you add this call to `_check_object_retention`?

src/sdk/namespace_fs.js
```diff
@@ -2518,14 +2523,19 @@ class NamespaceFS {
...
             const stat = await nb_native().fs.stat(fs_context, file_path);
             const current_retention = this._get_retention_mode_from_xattr(stat.xattr);
             this._compare_object_retention(fs_context, current_retention, params.retention, params.bypass_governance);
-            const fs_xattr = {};
-            fs_xattr[XATTR_RETENTION_MODE] = params.retention.mode;
-            fs_xattr[XATTR_RETENTION_DATE] = params.retention.retain_until_date.toISOString();
-            await this.set_fs_xattr_op(fs_context, file_path, fs_xattr, undefined);
+            if (is_clear) {
+                await this.set_fs_xattr_op(fs_context, file_path, undefined, XATTR_RETENTION_PREFIX);
+            } else {
```
shirady, line 2535: If one of them empty it is fine? did you intentionally set the `undefined` value?

src/server/object_services/object_server.js
```diff
@@ -350,27 +350,47 @@ async function put_object_retention(req) {
...
+            await MDStore.instance().update_object_by_id(
+                obj._id, {
+                    lock_settings: {
+                        legal_hold,
+                    }
+                }, undefined, undefined
+            );
+        } else {
+            await MDStore.instance().update_object_by_id(
+                obj._id, undefined, { lock_settings: 1 }, undefined
```
shirady, line 367: What about the last argument? Do you want to explicitly pass `undefined`?

src/endpoint/s3/ops/s3_put_object_retention.js
kajalpareek-lab: Good suggestion, restructured to if / else if / else.

src/sdk/namespace_fs.js
kajalpareek-lab: Added

src/sdk/namespace_fs.js
kajalpareek-lab: Added.

src/api/object_api.js
kajalpareek-lab: Done — removed, file drops out of the diff.

src/endpoint/s3/ops/s3_put_object_retention.js
kajalpareek-lab: This is pre-existing behavior (not from this PR). _is_valid_retention returns false if mode is invalid or retain_until_date is in the past — the early return silently no-ops, matching the original code.

src/sdk/namespace_fs.js
kajalpareek-lab: _check_object_retention enforces bypass-governance rules — it checks whether retention is still active (now < retain_until_date) and whether GOVERNANCE can be bypassed or COMPLIANCE which can never be bypassed. When clearing retention (empty object), we still need to verify the caller has permission to override the existing lock before removing it.

src/sdk/namespace_fs.js
kajalpareek-lab: Yes, intentional. set_fs_xattr_op(ctx, path, undefined, XATTR_RETENTION_PREFIX) — first arg undefined means "nothing to set", second arg (prefix) means "remove all xattrs matching this prefix". This removes both retention_mode and retention_date without touching legal_hold.

src/server/object_services/object_server.js
kajalpareek-lab: Yes, intentional. update_object_by_id takes 4 positional args (id, set, unset, inc). The trailing undefineds are needed to pass the right value to the right position. Same pattern used for tagging operations above in the same file.

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -566,6 +566,68 @@ mocha.describe('s3 worm', function() {
...
+            const res = await s3_owner.putObjectRetention({
+                Bucket: BKT1,
+                Key: CLEAR_RET_KEY,
+                VersionId: clear_ret_version_id,
+                Retention: { Mode: 'GOVERNANCE', RetainUntilDate: newDate },
+            });
+            delete res.$metadata;
+            assert.deepEqual(res, {});
+        });
+    });
```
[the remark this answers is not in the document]
kajalpareek-lab: Good suggestions. Will add COMPLIANCE rejection and legal_hold preservation coverage in a follow-up.

src/server/object_services/object_server.js
```diff
@@ -346,17 +345,30 @@ async function put_object_retention(req) {
     if (!req.bucket.object_lock_configuration || req.bucket.object_lock_configuration.object_lock_enabled !== 'Enabled') {
         throw new RpcError('INVALID_REQUEST');
     }
-    const current_retention = info.lock_settings?.retention;
     const new_retention = req.rpc_params.retention;
+    const is_clear = !new_retention.mode && !new_retention.retain_until_date;
+    const current_retention = info.lock_settings?.retention;
```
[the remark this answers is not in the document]
kajalpareek-lab: retention is required in the RPC schema (object_api.js line 1315-1318), so it's guaranteed to be present. The RPC layer rejects calls without it before reaching this function.

src/server/object_services/object_server.js
shirady: What does it mean unset of "{ lock_settings: 1 }"?

src/test/integration_tests/api/s3/test_s3_worm.js
```diff
@@ -566,6 +566,68 @@ mocha.describe('s3 worm', function() {
         });
     });
 
+    mocha.describe('clear GOVERNANCE retention with empty Retention', function() {
+        const CLEAR_RET_KEY = 'clear-retention-test';
+        let clear_ret_version_id;
+
+        mocha.it('should create object with GOVERNANCE retention', async function() {
```
shirady, line 573: Is that a test case, or should it be in the `before`?

shirady (commented):
I added 2 small questions

src/test/integration_tests/api/s3/test_s3_worm.js
kajalpareek-lab: Follows the existing pattern in this file all other describe blocks use mocha.it('should create ...') for object setup rather than before. Happy to refactor to before if you'd prefer, but it would be inconsistent with the rest of the test suite.

src/server/object_services/object_server.js
kajalpareek-lab: { lock_settings: 1 } in the unset position tells MongoDB to remove the lock_settings field from the document. The 1 just means "yes, remove this field"  same pattern used at line 210 for removing tagging.

src/test/integration_tests/api/s3/test_s3_worm.js
shirady: No need to, thanks.

shirady (approved):
LGTM

nadavMiz (approved):

src/sdk/namespace_fs.js
```diff
@@ -2518,14 +2525,19 @@ class NamespaceFS {
         const fs_context = this.prepare_fs_context(object_sdk);
         const file_path = await this._find_version_path(fs_context, params, true);
         await this._check_path_in_bucket_boundaries(fs_context, file_path);
+        const is_clear = !params.retention.mode && !params.retention.retain_until_date;
         try {
             const stat = await nb_native().fs.stat(fs_context, file_path);
             const current_retention = this._get_retention_mode_from_xattr(stat.xattr);
             this._compare_object_retention(fs_context, current_retention, params.retention, params.bypass_governance);
```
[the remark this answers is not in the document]
kajalpareek-lab: Invalid — retention is required in the RPC schema and the S3 handler always passes it (either {} for clear or {mode, retain_until_date} for set). The RPC layer rejects calls without it before reaching this code, so params.retention is guaranteed to be an object here.

PR #9921, opened 2026-08-09
DEEP_ARCHIVE | Reclaim expired restored object copies and transitioned objects on ObjectReclaimer BG worker + refactoring

romayalon:
### Describe the Problem
Temporary STANDARD restore copies and unreclaimed source storage class copies after lifecycle transition to Deep Archive were not cleaned up. After restore expiry (or transition DONE), local mappings could remain indefinitely, wasting capacity and leaving object MD out of sync.

### Explain the Changes
1. Extended ObjectsReclaimer with two expire paths alongside deleted-object reclaim: reclaim_expired_restores and reclaim_transition_expired_data, orchestrated each cycle with { had_work, had_errors } → _next_delay.
2. Added MDStore finds + partial indexes: find_expired_restore_objects (restore_status_index), find_objects_with_transition_done_expired_data (transition_expired_data_index).
3. Added map_deleter.delete_object_expired_data_mappings — deletes parts/chunks and only multiparts referenced by those parts (keeps MD-only archive multiparts).
4. Modeled nested transition_info / expired_data_info (incl. optional reclaimed); restore expiry unsets restore_status; transition reclaim sets expired_data_info.reclaimed and skips objects with restore_status.
5. Config batch sizes: OBJECT_RECLAIMER_EXPIRE_RESTORE_BATCH_SIZE, OBJECT_RECLAIMER_EXPIRE_TRANSITION_BATCH_SIZE.
6. Tests: functional coverage in test_md_store.js + query-plan coverage in test_query_plans.js.

1. Gap - we should consider changing the singular transitionInfo to transitionInfo[] for keeping the transitions history in case that in the future we will support more transitions like aws does.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added structured lifecycle transition metadata with status, source storage details, timestamps, and reclamation information.
  * Added automatic cleanup for expired restore copies and transitioned data.
  * Added separate batch-size settings for restore and transition cleanup.

* **Improvements**
  * Updated object metadata, APIs, and SDK types for consistent lifecycle status reporting.
  * Improved cleanup scheduling, filtering, and expired-data mapping removal.
  * Prevented conflicting metadata updates while expired data awaits cleanup.


src/server/bg_services/lifecycle.js
```diff
@@ -453,6 +453,7 @@ async function transition_objects(system, bucket_info, objects, target_storage_c
         try {
             const res = await server_rpc.client.object.update_transition_status({
                 update_transition_status: ARCHIVE.TRANSITION_STATUS.IN_PROGRESS,
+                expired_data_info: { storage_class: obj.storage_class || 'STANDARD' },
```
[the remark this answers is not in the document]
romayalon: added

src/server/bg_services/objects_reclaimer.js
```diff
@@ -64,22 +78,112 @@ class ObjectsReclaimer {
...
+        }
+
+        let had_errors = false;
+        dbg.log0('object_reclaimer: reclaiming expired restores:',
+            expired_restores.map(o => o.key).join(', '));
+
+        await P.all(expired_restores.map(async obj => {
+            try {
+                await map_deleter.delete_object_expired_data_mappings(obj);
+                await MDStore.instance().update_object_by_id(obj._id, undefined, { restore_status: 1 });
```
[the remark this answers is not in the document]
romayalon: added blocking restoreObject API if object was restored and expired but still not cleaned and transitioning started but not reclaimed yet

src/util/deep_archive_utils.js
```diff
@@ -65,6 +65,33 @@ function is_restore_active(restore_status, now = new Date()) {
...
+
+/**
+ * True when transition source-class data is marked for purge but not yet reclaimed —
+ * RestoreObject must wait until ObjectsReclaimer finishes (client can retry later).
+ * @param {{ transition_info?: { expired_data_info?: { reclaimed?: Date|number } } }} obj
+ * @returns {boolean}
+ */
+function is_transition_source_pending_purge(obj) {
+    const expired = obj?.transition_info?.expired_data_info;
+    return Boolean(expired && !expired.reclaimed);
```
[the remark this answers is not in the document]
romayalon: expiration_timestamp is always in the past, it's set to now when added

src/server/bg_services/objects_reclaimer.js
```diff
@@ -18,6 +19,20 @@ class ObjectsReclaimer {
         this.client = client;
     }
 
+    /**
+     * Orchestrates reclaim paths and returns the delay until the next run.
+     */
+    async run_batch() {
+        if (!this._can_run()) return;
```
karthikps97, line 25: Don't we need any default delay here or is it handled by bg_worker scheduler itself?

src/server/bg_services/objects_reclaimer.js
```diff
@@ -64,22 +78,118 @@ class ObjectsReclaimer {
...
+                    dbg.log0('object_reclaimer: skip transition reclaim while restore_status set:',
+                        obj.key);
+                    return;
+                }
+                await map_deleter.delete_object_expired_data_mappings(obj);
+                await MDStore.instance().update_object_by_id(
+                    obj._id,
+                    {
+                        transition_info: {
+                            status: CONSTANTS.ARCHIVE.TRANSITION_STATUS.DONE,
```
karthikps97: Do we need to pass all these again? We can just update the `reclaimed` ts field isn't it.

src/server/object_services/map_deleter.js
```diff
@@ -39,6 +39,24 @@ async function delete_object_multiparts(obj) {
...
+ * @param {nb.ObjectMD} obj
+ */
+async function delete_object_expired_data_mappings(obj) {
+    if (!obj || obj.delete_marker) return;
+    const parts = await MDStore.instance().find_all_parts_of_object(obj);
+    const chunk_ids = db_client.instance().uniq_ids(parts, 'chunk');
+    const multipart_ids = db_client.instance().uniq_ids(parts.filter(part => part.multipart), 'multipart');
+    await Promise.all([
+        MDStore.instance().delete_parts_of_object(obj),
+        multipart_ids.length && MDStore.instance().update_multiparts_by_ids(multipart_ids, { deleted: new Date() })
```
karthikps97: Maybe we can add a MD store method for deleting multiparts? I can see we already have `delete_multiparts`.

src/api/object_api.js
```diff
@@ -1944,5 +1944,24 @@ module.exports = {
...
+                status: { $ref: 'common_api#/definitions/transition_status_enum' },
+                expired_data_info: { $ref: '#/definitions/expired_data_info' },
+            }
+        },
+
+        expired_data_info: {
+            type: 'object',
+            required: ['storage_class', 'expiration_timestamp'],
+            properties: {
+                storage_class: { $ref: 'common_api#/definitions/storage_class_enum' },
```
shirady, line 1994: I understood from the docs that it is the source (before the transition) storage class, WDYT about reflecting this in the variable name?

src/server/bg_services/lifecycle.js
```diff
@@ -450,9 +450,14 @@ async function transition_objects(system, bucket_info, objects, target_storage_c
     });
     const obj_ids = await P.map_with_concurrency(batch_size, objects, (async obj => {
         const obj_id = obj.obj_id;
+        const expired_data_info = {
+            storage_class: obj.storage_class || 'STANDARD',
+            expiration_timestamp: Date.now(),
+        };
```
shirady: You have `STORAGE_CLASS_STANDARD`
```suggestion
        const expired_data_info = {
            storage_class: obj.storage_class || STORAGE_CLASS_STANDARD,
            expiration_timestamp: Date.now(),
        };
```

src/server/object_services/object_server.js
```diff
@@ -1107,6 +1109,11 @@ async function update_object_md(req) {
         set_updates.restore_status.expiry_time = new Date(set_updates.restore_status.expiry_time);
     }
     const obj = await find_object_md(req);
+
+    if (set_updates.restore_status && (is_expired_restore_pending_purge(obj) || is_transition_source_pending_purge(obj))) {
```
karthikps97, line 1133: If the user want to do a restore, why cant we just extend the restore expiry timestamp and prevent reclaimer from cleaning?

src/server/bg_services/objects_reclaimer.js
```diff
@@ -64,22 +78,118 @@ class ObjectsReclaimer {
...
+     *
+     * Delete mappings first so a failed cleanup stays retryable (restore_status
+     * still expired). RestoreObject is blocked while expired restore_status
+     * remains (see update_object_md), so unset after delete is safe.
+     * @returns {Promise<{ had_work: boolean, had_errors: boolean }>}
+     */
+    async reclaim_expired_restores() {
+        const batch_size = config.OBJECT_RECLAIMER_EXPIRE_RESTORE_BATCH_SIZE;
+        const expired_restores = await MDStore.instance().find_expired_restore_objects(batch_size);
+        if (!expired_restores || !expired_restores.length) {
```
shirady, line 107: I think it is easier to read when compared to zero.
```suggestion
        if (!expired_restores || expired_restores.length === 0) {
```

src/server/bg_services/objects_reclaimer.js
```diff
@@ -64,22 +78,118 @@ class ObjectsReclaimer {
...
+            return { had_work: false, had_errors: false };
+        }
+
+        let had_errors = false;
+        dbg.log0('object_reclaimer: reclaiming transition expired data for:',
+            transition_expired.map(o => o.key).join(', '));
+
+        await P.all(transition_expired.map(async obj => {
+            try {
+                // Defense in depth: find already excludes restore_status.
```
shirady: I'm not sure the text in this comment is clear.

src/server/object_services/schemas/object_md_indexes.js
shirady, line 1: I'm not sure you need the additional comments; they are in the code (for example, "Partial index" is `partialFilterExpression` (mongo docs [link](https://www.mongodb.com/docs/manual/core/index-partial/)).

src/server/object_services/map_deleter.js
```diff
@@ -39,6 +39,24 @@ async function delete_object_multiparts(obj) {
...
+/**
+ * Delete mappings for expired (non-target) data only — parts/chunks and multiparts
+ * referenced by those parts. Leaves MD-only archive multiparts untouched so
+ * deep-archive data remains addressable after restore expiry / lifecycle transition.
+ * @param {nb.ObjectMD} obj
+ */
+async function delete_object_expired_data_mappings(obj) {
+    if (!obj || obj.delete_marker) return;
+    const parts = await MDStore.instance().find_all_parts_of_object(obj);
+    const chunk_ids = db_client.instance().uniq_ids(parts, 'chunk');
```
shirady, line 67: Why didn't you use the function `find_parts_chunk_ids`?

src/server/object_services/map_deleter.js
shirady: I don't think you need this condition `multipart_ids.length && ` because the function `update_multiparts_by_ids` checks `if (!multipart_ids || !multipart_ids.length) return;`.
https://github.com/noobaa/noobaa-core/blob/c06d26bc8381749a7411fe5baa350e7e1dd95c4b/src/server/object_services/md_store.js#L1296-L1301

src/server/object_services/md_store.js
```diff
@@ -901,6 +901,47 @@ class MDStore {
...
+     * @returns {Promise<nb.ObjectMD[]>}
+     */
+    async find_expired_restore_objects(limit, now = new Date()) {
+        const results = await this._objects.find({
+            deleted: null,
+            upload_started: null,
+            'restore_status.ongoing': false,
+            'restore_status.expiry_time': { $lte: now },
+        }, {
+            limit: limit || 1000,
```
shirady: Letting you know that `limit || 1000` is not the same as defaulting to 1000 (`limit ?? 1000`) for the zero case.

src/server/object_services/object_server.js
```diff
@@ -1088,7 +1089,8 @@ function _check_encryption_permissions(src_enc, req_enc) {
 /**
  *
  * UPDATE_OBJECT_MD
- *
+ * object_md should not be updated for a new restore status while expired restore copy or unreclaimed transition
+ * source copy still needs db mapping purge — client can retry after reclaimer runs.
```
shirady: I don't think that the comment is in the right place.
It is the general `update_object_md` in `object_server`.

You can move it near the relevant code changes.

src/server/object_services/object_server.js
```diff
@@ -1107,6 +1109,11 @@ async function update_object_md(req) {
         set_updates.restore_status.expiry_time = new Date(set_updates.restore_status.expiry_time);
     }
     const obj = await find_object_md(req);
+
+    if (set_updates.restore_status && (is_expired_restore_pending_purge(obj) || is_transition_source_pending_purge(obj))) {
+        throw new RpcError('INVALID_OBJECT_STATE',
```
shirady: Are you sure about the RPC error?
I thought that `INVALID_OBJECT_STATE` is when we try to GetObject while it is not restored yet.

src/test/integration_tests/db/test_md_store.js
```diff
@@ -310,6 +310,238 @@ mocha.describe('md_store', function() {
...
+                key: `expired_restore_${suffix}`,
+                restore_status: {
+                    ongoing: false,
+                    expiry_time: new Date(Date.now() - 60_000),
+                },
+            };
+            const future = {
+                ...base(),
+                _id: md_store.make_md_id(),
+                key: `future_restore_${suffix}`,
```
shirady: I think the name is confusing (the restore is not future; it was restored and has a future expiry).

```suggestion
                key: `future_expiry_restore_${suffix}`,
```

src/test/integration_tests/db/test_md_store.js
```diff
@@ -310,6 +310,238 @@ mocha.describe('md_store', function() {
...
+                restore_status: {
+                    ongoing: false,
+                    expiry_time: new Date(Date.now() + 60 * 60_000),
+                },
+            };
+            const ongoing = {
```
shirady, line 340: What about adding a case where the ongoing is true, and the object was deleted during the time of the restore?

src/util/deep_archive_utils.js
```diff
@@ -65,6 +65,33 @@ function is_restore_active(restore_status, now = new Date()) {
...
 
+/**
+ * True when a completed restore copy is past expiry and still on MD — ObjectsReclaimer
+ * must purge mappings before a new RestoreObject may run (client can retry later).
+ * @param {{ restore_status?: { ongoing?: boolean, expiry_time?: Date|number } }} obj
+ * @param {Date} [now]
+ * @returns {boolean}
+ */
+function is_expired_restore_pending_purge(obj, now = new Date()) {
+    const rs = obj?.restore_status;
```
shirady: `obj` is required according to the JSDoc; why use optional chaining?
I prefer to use the full name of the variable for readability:

```suggestion
    const restore_status = obj.restore_status;
```

src/util/deep_archive_utils.js
```diff
@@ -65,6 +65,33 @@ function is_restore_active(restore_status, now = new Date()) {
...
+}
+
+/**
+ * True when transition source-class data is marked for purge but not yet reclaimed —
+ * RestoreObject must wait until ObjectsReclaimer finishes (client can retry later).
+ * @param {{ transition_info?: { expired_data_info?: { reclaimed?: Date|number } } }} obj
+ * @returns {boolean}
+ */
+function is_transition_source_pending_purge(obj) {
+    const expired = obj?.transition_info?.expired_data_info;
```
shirady: Same here - 
`obj` is required according to the JSDoc; why use optional chaining?
```suggestion
    const expired = obj.transition_info?.expired_data_info;
```

src/test/integration_tests/db/test_query_plans.js
```diff
@@ -132,6 +132,39 @@ mocha.describe('md_store query plan verification', function() {
...
+            bucket: bucket_id,
+            key: 'plan-expired-data-' + Date.now().toString(36),
+            content_type: 'text/plain',
+            size: 1,
+            create_time: new Date(),
+            transition_info: {
+                status: 'DONE',
+                expired_data_info: {
+                    storage_class: 'STANDARD',
+                    expiration_timestamp: new Date(),
```
shirady: If it is expired, don't you prefer setting it in the past so it would be clear?

src/test/integration_tests/db/test_query_plans.js
```diff
@@ -549,6 +584,15 @@ mocha.describe('md_store query plan verification', function() {
         await check('find_unreclaimed_objects', () => md_store.find_unreclaimed_objects(10), [OBJ]);
```
shirady, line 584: For readability, I think it would be clearer using a constant (instead of adding the argument of 10).

src/test/integration_tests/internal/test_objects_reclaimer.js
```diff
@@ -0,0 +1,443 @@
...
+        mocha.it('leaves non-expired restore_status untouched', async function() {
+            const obj = await upload_and_patch({
+                storage_class: CONSTANTS.ARCHIVE.STORAGE_CLASS.DEEP_ARCHIVE,
+                restore_status: {
+                    ongoing: false,
+                    expiry_time: new Date(Date.now() + 24 * 3600_000),
+                },
+            });
+
+            await reclaimer.reclaim_expired_restores();
```
shirady: Could you work on the result and assert on it, like in the following tests?
```suggestion
            const res = await reclaimer.reclaim_expired_restores();
            ...
```

src/test/integration_tests/internal/test_objects_reclaimer.js
```diff
@@ -0,0 +1,443 @@
...
+
+            const after = await assert_object_not_deleted(obj._id);
+            assert.ok(!after.restore_status, 'restore_status should be unset after expiry reclaim');
+            assert.strictEqual(after.storage_class, CONSTANTS.ARCHIVE.STORAGE_CLASS.DEEP_ARCHIVE,
+                'object remains DEEP_ARCHIVE after reclaiming the STANDARD restore copy');
+            const parts = await MDStore.instance().find_all_parts_of_object(after);
+            assert.strictEqual(parts.length, 0, 'local restore parts should be deleted');
+        });
+
+        mocha.it('on MPU deletes STANDARD multiparts only and keeps deep-archive multiparts', async function() {
```
shirady, line 124: I find this test case a bit complex. Could you add a few comments to help other developers better understand it?

src/test/integration_tests/internal/test_objects_reclaimer.js
```diff
@@ -0,0 +1,443 @@
...
+        mocha.it('returns a configured delay', async function() {
+            const delay = await reclaimer.run_batch();
+            assert.ok([
+                config.OBJECT_RECLAIMER_EMPTY_DELAY,
+                config.OBJECT_RECLAIMER_BATCH_DELAY,
+                config.OBJECT_RECLAIMER_ERROR_DELAY,
+            ].includes(delay), `unexpected delay ${delay}`);
+        });
+
+        mocha.it('_next_delay prefers errors over work over empty', function() {
```
shirady, line 272: Since it is the same value, I'm not sure that it actually tests what it should, and I guess you can change the values so they would be different.

src/test/integration_tests/internal/test_objects_reclaimer.js
```diff
@@ -0,0 +1,443 @@
...
+            const parts = await MDStore.instance().find_all_parts_of_object(after);
+            assert.strictEqual(parts.length, 0);
+        });
+
+        mocha.it('is a no-op for objects with no parts', async function() {
+            const obj = await upload_and_patch({
+                storage_class: CONSTANTS.ARCHIVE.STORAGE_CLASS.DEEP_ARCHIVE,
+            });
+            await map_deleter.delete_object_expired_data_mappings(obj);
+            await map_deleter.delete_object_expired_data_mappings(obj);
```
shirady: Maybe you can add a comment that you called `delete_object_expired_data_mappings` twice intentionally.

src/server/bg_services/objects_reclaimer.js
romayalon: yeah it's handled by the bg_worker scheduler, I didn't change it

src/server/object_services/object_server.js
romayalon: I did this for blocking these race conditions that might happen in the middle of this non atomic mapping deletion so we won't be left with corrupted restore copy, will add a TODO to try to resolve it

src/api/object_api.js
romayalon: yeah, Karthik is changing it on his PR

src/server/object_services/map_deleter.js
romayalon: we need the parts for the multiparts as well, if i'll use find_parts_chunk_ids() i'll need to go to the db 3 times instead of 2 because find_parts_chunk_ids() goes to the db twice anyway

src/server/object_services/map_deleter.js
romayalon: still needed because I changed to delete_multiparts() per your comment above

src/test/integration_tests/db/test_query_plans.js
romayalon: let's keep the current code as was used before adding this test

src/server/bg_services/objects_reclaimer.js
romayalon: let's keep the current code to fit the other functions in this file

src/server/object_services/schemas/object_md_indexes.js
romayalon: let's keep them for documentation

src/server/bg_services/objects_reclaimer.js
romayalon: right

karthikps97 (approved):

PR #9945, opened 2026-08-13
DEEP_ARCHIVE | Fix - `read_object_md` To Omit `restore_status` Expiry Invalid or in the Past

shirady:
### Describe the Problem
Part of [RHSTOR-8823](https://redhat.atlassian.net/browse/RHSTOR-8823)
In case the object was already restored and its expiry date is in the past, and the BG worker that handles removing the STANDARD copy did not finish removing it, we would like to avoid head-object calls that would include an expiry date in the past (property and header).

### Explain the Changes
1. Edit the function `read_object_md` instead of being just a passthrough to the metadata namespace to omit the `restore_status` in case the expiry is invalid or in the past.
2. Added AI-generated tests.

### Issues: 
List of GAPs:
1. We handled the call for HeadObject, but when listing objects with, for example: `s3 s3api list-objects-v2 --bucket <bucket name> --optional-object-attributes RestoreStatus`, it is not handled here.

#### Unit Tests
Please run:
1. `npx jest src/test/unit_tests/internal/test_namespace_multi_storage_class_restore.test.js`

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Object metadata now omits restore status when no restore is active or when restore information has expired or is invalid.
  * Restore status remains available for active and ongoing restores.

* **Tests**
  * Added coverage for active, ongoing, expired, missing-expiry, and invalid-expiry restore scenarios.

### DIFF
--- a/src/sdk/namespace_multi_storage_class.js
+++ b/src/sdk/namespace_multi_storage_class.js
@@ -123,13 +123,18 @@ class NamespaceMultiStorageClass {
     /////////////////
 
     /**
-     * Reads object metadata from the metadata namespace
+     * Reads object metadata from the metadata namespace.
+     * Omits restore_status when restore is expired or incomplete
      * @param {object} params
      * @param {nb.ObjectSDK} object_sdk
      * @returns {Promise<nb.ObjectInfo>}
      */
     async read_object_md(params, object_sdk) {
-        return this._metadata_ns.read_object_md(params, object_sdk);
+        const object_md = await this._metadata_ns.read_object_md(params, object_sdk);
+        const { restore_status } = object_md;
+
+        return (!restore_status || restore_status.ongoing || is_restore_active(restore_status)) ?
+            object_md : _.omit(object_md, 'restore_status');
     }
 
     /**
--- a/src/test/unit_tests/internal/test_namespace_multi_storage_class_restore.test.js
+++ b/src/test/unit_tests/internal/test_namespace_multi_storage_class_restore.test.js
@@ -61,6 +61,21 @@ function make_msc_fixture({ object_restore_info: object_restore_info_override, a
     };
 }
 
+function make_read_object_md_msc_fixture(object_md) {
+    const read_object_md = jest.fn().mockResolvedValue(object_md);
+    const metadata_ns = { read_object_md };
+    const object_sdk = {};
+
+    const ns_msc = new NamespaceMultiStorageClass({
+        namespace_by_storage_class: {
+            [s3_utils.STORAGE_CLASS_STANDARD]: metadata_ns,
+            [s3_utils.STORAGE_CLASS_DEEP_ARCHIVE]: {},
+        },
+    });
+
+    return { ns_msc, metadata_ns, read_object_md, object_sdk };
+}
+
 describe('NamespaceMultiStorageClass.restore_object', () => {
     const params = { bucket: BUCKET, key: KEY, days: 7 };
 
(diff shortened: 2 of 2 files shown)


src/sdk/namespace_multi_storage_class.js
```diff
@@ -123,13 +123,20 @@ class NamespaceMultiStorageClass {
...
      */
     async read_object_md(params, object_sdk) {
-        return this._metadata_ns.read_object_md(params, object_sdk);
+        const object_md = await this._metadata_ns.read_object_md(params, object_sdk);
+        const { restore_status } = object_md;
+
+        if (!restore_status) return object_md;
+        if (restore_status.ongoing || is_restore_active(restore_status)) return object_md;
+
+        return _.omit(object_md, 'restore_status');
```
romayalon: I would avoid having 3 different return branches
```suggestion
        const res = !restore_status || restore_status.ongoing || is_restore_active(restore_status)) object_md ? _.omit(object_md, 'restore_status');
        return res;
```

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
```diff
@@ -0,0 +1,154 @@
...
+
+    it('returns restore_status when restore is ongoing', async () => {
+        const restore_status = { ongoing: true, days: 7 };
+        const metadata_object_md = {
+            key: KEY,
+            storage_class: s3_utils.STORAGE_CLASS_DEEP_ARCHIVE,
+            restore_status,
+        };
+        const { ns_msc, object_sdk } = make_msc_fixture(metadata_object_md);
+
```
romayalon: you have 2 empty lines in the middle of the test in all below tests

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
```diff
@@ -0,0 +1,154 @@
+/* Copyright (C) 2026 NooBaa */
+'use strict';
+
+const NamespaceMultiStorageClass = require('../../../sdk/namespace_multi_storage_class');
```
romayalon: if possible use existing test files

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
```diff
@@ -0,0 +1,154 @@
...
+        };
+        const { ns_msc, object_sdk } = make_msc_fixture(metadata_object_md);
+
+        const object_md = await ns_msc.read_object_md(params, object_sdk);
+
+        expect(object_md.restore_status).toEqual(restore_status);
+    });
+
+    it('omits restore_status when restore failed without expiry_time', async () => {
+        const metadata_object_md = {
```
romayalon:
```suggestion
        const object_md = {
```

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
```diff
@@ -0,0 +1,154 @@
...
+            key: KEY,
+            storage_class: s3_utils.STORAGE_CLASS_GLACIER,
+            restore_status: { ongoing: false },
+        };
+        const { ns_msc, object_sdk } = make_msc_fixture(metadata_object_md);
+
+        const object_md = await ns_msc.read_object_md(params, object_sdk);
```
romayalon:
```suggestion
        const object_md_res = await ns_msc.read_object_md(params, object_sdk);
```

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
```diff
@@ -0,0 +1,154 @@
...
+            restore_status: {
+                ongoing: false,
+                expiry_time: new Date('2000-01-01T00:00:00Z').getTime(),
+            },
+        };
+        const { ns_msc, read_object_md, object_sdk } = make_msc_fixture(metadata_object_md);
+
+        const object_md = await ns_msc.read_object_md(params, object_sdk);
+
+        expect(read_object_md).toHaveBeenCalledWith(params, object_sdk);
```
romayalon: why do we need this?

src/sdk/namespace_multi_storage_class.js
shirady: I will change it, thanks.

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
shirady: Thanks, I will remove the empty lines.

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
shirady: Thanks, I will change it across the file.

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
shirady: Thanks, I will change it across the file.

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
shirady: I don't need this; we can see the exact arguments when we use the call. Thanks

src/test/unit_tests/internal/test_namespace_multi_storage_class_read_object_md.test.js
shirady: It is possible - I dropped header cases for `x-amz-restore` since I saw it was covered by `test_s3_utils.test.js` and I expanded the file `test_namespace_multi_storage_class_restore.test` which was used only for restore object. Thanks

romayalon (approved):

PR #9947, opened 2026-08-13
nc - iam - role policy crud

alphaprinz:
### Describe the Problem
Implement role inline policy crud api for NC.

Also enable tests for NC.

### Explain the Changes
1. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added support for creating, replacing, retrieving, deleting, and listing inline policies for roles.
  * Added role-specific authorization details for role operations.
  * Policy documents now support validation, serialization, and size limits.

* **Bug Fixes**
  * Improved policy-related error handling and existence checks.
  * Role changes now persist reliably and refresh cached data.
  * Corrected role policy retrieval.
  * Prevented deletion of roles that still contain inline policies.

* **Tests**
  * Enabled IAM role and role-policy integration tests in additional environments.

### DIFF
--- a/src/sdk/accountspace_fs.js
+++ b/src/sdk/accountspace_fs.js
@@ -552,19 +552,19 @@ class AccountSpaceFS {
             const requested_account = await this.config_fs.get_account_or_user_by_name(
                 params.username, owner_account_id, { show_secrets: true, decrypt_secret_key: true });
             this._check_if_requested_is_owned_by_root_account(action, requesting_account, requested_account);
-            const iam_user_policies = [...(requested_account.iam_user_policies || [])];
-            const index_of_iam_user_policy = _get_iam_policy_index(iam_user_policies, params.policy_name);
-            const iam_user_policy_to_add = {
+            const iam_inline_policies = [...(requested_account.iam_inline_policies || [])];
+            const index_of_iam_inline_policy = _get_iam_policy_index(iam_inline_policies, params.policy_name);
+            const iam_inline_policy_to_add = {
                 policy_name: params.policy_name,
                 policy_document: params.policy_document,
             };
-            if (index_of_iam_user_policy === -1) {
-                iam_user_policies.push(iam_user_policy_to_add);
+            if (index_of_iam_inline_policy === -1) {
+                iam_inline_policies.push(iam_inline_policy_to_add);
             } else {
-                iam_user_policies[index_of_iam_user_policy] = iam_user_policy_to_add;
+                iam_inline_policies[index_of_iam_inline_policy] = iam_inline_policy_to_add;
             }
-            this._check_total_policy_size(action, iam_user_policies, params.username);
-            requested_account.iam_user_policies = iam_user_policies;
+            this._check_total_policy_size(action, iam_inline_policies, params.username);
+            requested_account.iam_inline_policies = iam_inline_policies;
             await this.config_fs.update_account_config_file(requested_account);
             _clean_account_id_cache(requested_account);
         } catch (err) {
@@ -582,7 +582,7 @@ class AccountSpaceFS {
     async get_user_policy(params, account_sdk) {
         const action = IAM_ACTIONS.GET_USER_POLICY;
         dbg.log1(`AccountSpaceFS.${action}`, params);
-        let iam_user_policies;
+        let iam_inline_policies;
         try {
             const requesting_account = account_sdk.requesting_account;
             this._check_if_requesting_account_is_root_account(action, requesting_account, { username: params.username });
@@ -591,16 +591,16 @@ class AccountSpaceFS {
             const requested_account = await this.config_fs.get_account_or_user_by_name(params.username, owner_account_id);
             this._check_if_requested_account_is_root_account_or_IAM_user(action, requesting_account, requested_account);
             this._check_if_requested_is_owned_by_root_account(action, requesting_account, requested_account);
-            iam_user_policies = requested_account.iam_user_policies || [];
+            iam_inline_policies = requested_account.iam_inline_policies || [];
         } catch (err) {
             dbg.error(`AccountSpaceFS.${action} error`, err);
             throw native_fs_utils.translate_error_codes(err, native_fs_utils.entity_enum.USER);
         }
-        const iam_user_policy_index = this._check_iam_policy_exists(action, iam_user_policies, params.policy_name);
+        const iam_inline_policy_index = this._check_iam_policy_exists(action, iam_inline_policies, params.policy_name);
         return {
             username: params.username,
             policy_name: params.policy_name,
-            policy_document: JSON.stringify(iam_user_policies[iam_user_policy_index].policy_document),
+            policy_document: JSON.stringify(iam_inline_policies[iam_inline_policy_index].policy_document),
         };
     }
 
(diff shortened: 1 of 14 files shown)


src/sdk/accountspace_fs.js
```diff
@@ -840,33 +840,114 @@ class AccountSpaceFS {
...
     async put_role_policy(params, account_sdk) {
         const action = IAM_ACTIONS.PUT_ROLE_POLICY;
         dbg.log1(`AccountSpaceFS.${action}`, params);
-        const { code, http_code, type } = IamError.NotImplemented;
-        throw new IamError({ code, message: 'NotImplemented', http_code, type });
+        try {
+            const requesting_account = account_sdk.requesting_account;
+            this._check_if_requesting_account_is_root_account(action, requesting_account, {});
+            const { owner_account_id, role_data } = await this._check_if_role_exists(params, requesting_account);
+            const iam_user_policies = [...(role_data.iam_user_policies || [])];
```
sakshimunjal: for roles, we should update iam_role_policies and not iam_user_policies. Please add a new field in nsfs_account_schema, you can refer [PR](https://github.com/noobaa/noobaa-core/pull/9916/changes)

src/sdk/accountspace_fs.js
sakshimunjal, line 810: @jackyalbo @alphaprinz As per [AWS doc](https://docs.aws.amazon.com/boto3/latest/reference/services/iam/client/delete_role.html#), delete role is blocked if there is an inline policy attached. Should we follow the same?

src/sdk/accountspace_fs.js
```diff
@@ -840,33 +840,114 @@ class AccountSpaceFS {
...
+        try {
+            const requesting_account = account_sdk.requesting_account;
+            this._check_if_requesting_account_is_root_account(action, requesting_account, {});
+            const { role_data } = await this._check_if_role_exists(params, requesting_account);
+            iam_role_policies = role_data.iam_user_policies || [];
+        } catch (err) {
+            dbg.error(`AccountSpaceFS.${action} error`, err);
+            throw native_fs_utils.translate_error_codes(err, native_fs_utils.entity_enum.ROLE);
+        }
+        const policy_index = this._check_iam_policy_exists(action, iam_role_policies, params.policy_name);
```
sakshimunjal: currently _check_iam_policy_exists has user policy specific message, can we customize it?

src/sdk/accountspace_fs.js
sakshimunjal: also, add a default empty value in _new_role_defaults() if required

src/sdk/accountspace_fs.js
alphaprinz: added

src/sdk/accountspace_fs.js
alphaprinz: added

src/sdk/accountspace_fs.js
alphaprinz: I don't think we need to add a new field. If only one of the fields will be used, we can use just one.

aayushchouhan09: I think even if the schema is the same, roles should use `iam_role_policies` and users should use `iam_user_policies` so behavior stays clear and consistent with the APIs.
We should keep both fields and set `iam_role_policies: []` in role defaults if needed.
@naveenpaul1 @shirady wdyt?

sakshimunjal: I understand only one will be used but both the fields have a different purpose. Also, since containerized design is also using 2 different fields, i think it's better to be aligned since codebase might be shared at places like iam_utils.js -> _get_identity_policies(), we might have to add special handling in such cases

aayushchouhan09: We still having #9916 open for containerized schema changes. 
In the PR we have two separate `iam_user_policies` and `iam_role_policies`, we can decide here and make changes accordingly in both.
cc @jackyalbo If you have anything to add

alphaprinz: @jackyalbo @shirady Can you please weigh in. It's not a big change, just a matter of preference. Another option is to use one field with the name "inline_policies".

jackyalbo: Any reason we will need both? I agree with @alphaprinz it sound like we should use the one iam_policy for role/user/group whatever

aayushchouhan09: Fine with `iam_inline_policy`, will also update the same in my PR #9916 

We do not need multiple fields, one is fine if we have proper/generic name for all.

shirady: I'm fine with the suggestion of having a single array in the schema that would serve user and role (actually, I raised this as question [here](https://github.com/noobaa/noobaa-core/pull/9916#discussion_r3756462026) as well).

Anyway, please remember that schema changes in containerized must be with upgrade script since it was already GA'd (for NC it is new in this version so it not a change yet).
cc: @aayushchouhan09 @naveenpaul1

src/sdk/accountspace_fs.js
```diff
@@ -840,33 +843,114 @@ class AccountSpaceFS {
...
+    // 4 - return role_name, policy_name and policy_document as string
     async get_role_policy(params, account_sdk) {
         const action = IAM_ACTIONS.GET_ROLE_POLICY;
         dbg.log1(`AccountSpaceFS.${action}`, params);
-        const { code, http_code, type } = IamError.NotImplemented;
-        throw new IamError({ code, message: 'NotImplemented', http_code, type });
+        let iam_role_policies;
+        try {
+            const requesting_account = account_sdk.requesting_account;
+            this._check_if_requesting_account_is_root_account(action, requesting_account, {});
```
shirady: Seems like passing `{}` as the argument of `user_details` would result in incorrect error message, did you verify that?

src/sdk/accountspace_fs.js
```diff
@@ -840,33 +843,114 @@ class AccountSpaceFS {
...
+                iam_user_policies.push(iam_role_policy_to_add);
+            } else {
+                iam_user_policies[policy_index] = iam_role_policy_to_add;
+            }
+            this._check_total_policy_size(action, iam_user_policies, params.role_name);
+            role_data.iam_user_policies = iam_user_policies;
+            await this.config_fs.update_role_config_file(role_data);
+            iam_roles_cache.invalidate({
+                role_name: role_data.name,
+                owner_account_id: String(owner_account_id),
```
shirady: Why do you need to convert to `String`?
In NC the IDs are strings.
https://github.com/noobaa/noobaa-core/blob/fcb149fca74527264dd1d601a1de857260ed9311/src/server/system_services/schemas/nsfs_account_schema.js#L15-L17

https://github.com/noobaa/noobaa-core/blob/fcb149fca74527264dd1d601a1de857260ed9311/src/server/system_services/schemas/nsfs_account_schema.js#L28-L30

src/sdk/accountspace_fs.js
```diff
@@ -1275,11 +1359,11 @@ class AccountSpaceFS {
         }
     }
 
-    _check_iam_policy_exists(action, iam_policies, policy_name) {
+    _check_iam_policy_exists(action, iam_policies, policy_name, container = 'user') {
```
shirady: Why did you choose `container` as the name of the variable?
I think @aayushchouhan09 is using `entity_type`.

shirady:
@alphaprinz I think you can add test cases in the file `src/test/unit_tests/nsfs/test_accountspace_fs.test.js` I saw new tests of role CRUD API that were added there.

src/sdk/accountspace_fs.js
alphaprinz: Thanks, fixed.

alphaprinz:
> @alphaprinz I think you can add test cases in the file `src/test/unit_tests/nsfs/test_accountspace_fs.test.js` I saw new tests of role CRUD API that were added there.

Added.

sakshimunjal (commented):
LGTM

sakshimunjal (approved):
LGTM

aayushchouhan09 (approved):
LGTM

upgrade script changes are not in this PR, I will handle it in my PR #9916

src/sdk/accountspace_fs.js
shirady: @alphaprinz I still see the convert to string:
`owner_account_id: String(owner_account_id)`

PR #9957, opened 2026-08-18
Replace openssl with base64 in nobaa_init.sh

naveenpaul1:
### Describe the Problem

`openssl enc -base64 -d -A` failed to decode the base64-encoded `agent_conf` and exited without throwing an error.
This issue likely stems from the unreliable behavior of OpenSSL's `-A` flag, as documented in the [Base64 Encoding Strings](https://wiki.openssl.org/index.php/Command_Line_Utilities) section of the OpenSSL man page.

### Explain the Changes
1. prefer `base64` over `openssl` for encoding the `AGENT_CONFIG` in `nobaa_init.sh`
2. Remove `jq`

1. https://redhat.atlassian.net/browse/DFBUGS-9553

### Testing Instructions:
1. create backingstore and check for message 
`AGENT_CONFIG successfully moved to /noobaa_storage/agent_conf.json`

- [ ] Doc added/updated
- [ ] Tests added

## Bug Fixes

* Improved configuration handling by validating existing and newly supplied configuration files.
* Added support for plain JSON and base64-encoded configuration input.
* Invalid, empty, or malformed configuration data is now rejected with an error.
* Configuration updates are applied safely to help prevent incomplete or corrupted files.
* Improved compatibility when decoding base64-encoded configuration data.

### DIFF
--- a/src/deploy/NVA_build/noobaa_init.sh
+++ b/src/deploy/NVA_build/noobaa_init.sh
@@ -93,10 +93,18 @@ run_internal_process() {
   done
 }
 
+is_valid_json_file() {
+  [ -s "$1" ] && node -e "JSON.parse(require('fs').readFileSync(process.argv[1], 'utf8'))" "$1" >/dev/null 2>&1
+}
+
 prepare_agent_conf() {
   AGENT_CONF_FILE="/noobaa_storage/agent_conf.json"
 
-  [ -f "$AGENT_CONF_FILE" ] && return 0
+  # A previous run may have left behind an empty/corrupt file.
+  # Only trust an existing file if it is actually valid JSON, otherwise regenerate it.
+  if is_valid_json_file "$AGENT_CONF_FILE"; then
+    return 0
+  fi
 
   # get AGENT_CONFIG from env var or file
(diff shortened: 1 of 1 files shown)


dannyzaken (commented):
@naveenpaul1 did you test this code on a downstream image? I think that we don't have `jq` installed on the DS image and this might even be the root cause for the bug.

did you consider decoding the base64 json in the operator side? This should eliminate this entire code

naveenpaul1:
@dannyzaken affected version is 4.18.8, and `jq` was added in PR [9243](https://github.com/noobaa/noobaa-core/pull/9243) which I think is not part of 4.18.
I couldnt test this DS image, I will try to test it there
And also try to move the logic to operator

dannyzaken (approved):

src/deploy/NVA_build/noobaa_init.sh
```diff
@@ -158,4 +186,4 @@ then
 else
   echo "noobaa_init.sh: unknown or missing RUN_INIT arg '${RUN_INIT}'"
   exit 1
-fi
+fi
```
aayushchouhan09: missing new line at the end of file
```suggestion
fi

```

aayushchouhan09 (approved):
LGTM

PR #9961, opened 2026-08-19
s3 block store - adapt buffer to sdk V3

nadavMiz:
### Describe the Problem
AWS SDK v3 getObject returns a stream, not a Buffer (see https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/migrate-s3.html#amazon-s3-stream-vs-buffer). The block store still expected a Buffer, so reads and mirror copies failed with `Block data must be a buffer`.

### Explain the Changes
1.  s3_body_to_buffer: if Body is already a Buffer (SDK v2), return it; if v3, call transformToByteArray() and wrap in a Buffer.
2. Use it on block read and usage-file read.
3. Add unit tests

1. https://redhat.atlassian.net/browse/DFBUGS-9429

### Testing Instructions:
1. `node ./node_modules/.bin/_mocha src/test/unit_tests/internal/test_block_store_s3.js`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of S3 object data across supported AWS SDK versions.
  * Block initialization and reads now consistently process response bodies and report unsupported formats clearly.
  * Improved reliability when retrieving data returned as either buffers or streams.

* **Tests**
  * Added coverage for buffer responses, stream-based responses, and unsupported body types.
  * Verified consistent conversion and error handling across supported response formats.

### DIFF
--- a/src/agent/block_store_services/block_store_s3.js
+++ b/src/agent/block_store_services/block_store_s3.js
@@ -87,8 +87,9 @@ class BlockStoreS3 extends BlockStoreBase {
                 Key: this.usage_path,
             });
 
+            const body = await noobaa_s3_client.s3_body_to_buffer(res.Body);
             const usage_data = this.disable_metadata ?
-                res.Body.toString() :
+                body.toString() :
                 res.Metadata[this.usage_md_key];
             if (usage_data && usage_data.length) {
                 this._usage = this._decode_block_md(usage_data);
@@ -174,7 +175,7 @@ class BlockStoreS3 extends BlockStoreBase {
                 Key: this._block_key(block_md.id),
             });
             return {
-                data: res.Body,
+                data: await noobaa_s3_client.s3_body_to_buffer(res.Body),
                 block_md: this._get_store_block_md(block_md, res),
             };
         } catch (err) {
--- a/src/sdk/noobaa_s3_client/noobaa_s3_client.js
+++ b/src/sdk/noobaa_s3_client/noobaa_s3_client.js
@@ -148,6 +148,22 @@ function add_response_header_capture(s3) {
     return () => response_headers;
 }
 
+/**
+ * Converts AWS SDK getObject Body to a Buffer.
+ * SDK v2 returns a Buffer. SDK v3 returns a stream with transformToByteArray.
+ * @param {*} body - getObject Body from AWS SDK v2 or v3
+ * @returns {Promise<Buffer>}
+ */
+async function s3_body_to_buffer(body) {
+    // SDK v2 returns a Buffer
+    if (Buffer.isBuffer(body)) return body;
+    // SDK v3 returns a stream with transformToByteArray
+    if (body && typeof body.transformToByteArray === 'function') {
+        return Buffer.from(await body.transformToByteArray());
+    }
+    throw new Error(`Unexpected Body type: ${body && body.constructor && body.constructor.name}`);
+}
+
 // EXPORTS
 exports.get_s3_client_v3_params = get_s3_client_v3_params;
 exports.change_s3_client_params_to_v2_structure = change_s3_client_params_to_v2_structure;
@@ -155,3 +171,4 @@ exports.get_sdk_class_str = get_sdk_class_str;
 exports.fix_error_object = fix_error_object;
 exports.get_requestHandler_with_suitable_agent = get_requestHandler_with_suitable_agent;
 exports.add_response_header_capture = add_response_header_capture;
+exports.s3_body_to_buffer = s3_body_to_buffer;
--- a/src/test/unit_tests/internal/test_noobaa_s3_client.test.js
+++ b/src/test/unit_tests/internal/test_noobaa_s3_client.test.js
@@ -109,3 +109,30 @@ describe('noobaa_s3_client change_s3_client_params_to_v2_structure', () => {
     });
 
 });
+
+describe('noobaa_s3_client s3_body_to_buffer', () => {
+
+    const data = Buffer.from('s3-body-payload');
+
+    it('returns a Buffer Body unchanged (SDK v2 shape)', async () => {
+        const buf = await noobaa_s3_client.s3_body_to_buffer(data);
+        expect(Buffer.isBuffer(buf)).toBe(true);
+        expect(buf.equals(data)).toBe(true);
+    });
+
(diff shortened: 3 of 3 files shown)


nadavMiz:
@coderabbitai review

src/test/unit_tests/internal/test_block_store_s3.js
```diff
@@ -0,0 +1,111 @@
...
+        assert.ok(Buffer.isBuffer(block.data));
+        assert.ok(data.equals(block.data));
+    });
+
+    mocha.it('converts an SDK v3 stream Body to a Buffer', async function() {
+        const store = make_store({ getObjectBody: () => make_sdk_v3_body(data) });
+        const block = await store._read_block_and_verify(block_md);
+        assert.ok(Buffer.isBuffer(block.data), `expected Buffer, got ${block.data && block.data.constructor && block.data.constructor.name}`);
+        assert.ok(data.equals(block.data));
+    });
```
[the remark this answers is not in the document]
nadavMiz: The reject branch is only a fail-fast for Body shapes getObject never returns; Buffer and SDK v3 stream coverage already pin the two real types

src/agent/block_store_services/block_store_s3.js
```diff
@@ -452,5 +452,21 @@ class BlockStoreS3 extends BlockStoreBase {
 
 }
 
+/**
+ * Converts AWS SDK getObject Body to a Buffer.
+ * SDK v2 returns a Buffer. SDK v3 returns a stream with transformToByteArray.
+ * @param {*} body - getObject Body from AWS SDK v2 or v3
+ * @returns {Promise<Buffer>}
+ */
+async function s3_body_to_buffer(body) {
```
shirady: This function is a helper function for the SDK, right?
So I'm suggeting to move it to `noobaa_s3_client.js` so it would be clear that others can reuse it if needed (and not only for `block_store_s3`).

src/agent/block_store_services/block_store_s3.js
```diff
@@ -452,5 +452,21 @@ class BlockStoreS3 extends BlockStoreBase {
...
+    // SDK v2 returns a Buffer
+    if (Buffer.isBuffer(body)) return body;
+    // SDK v3 returns a stream with transformToByteArray
+    if (body && typeof body.transformToByteArray === 'function') {
+        return Buffer.from(await body.transformToByteArray());
+    }
+    throw new Error(`Unexpected S3 getObject Body type: ${body && body.constructor && body.constructor.name}`);
```
shirady: Can you share how this error message looks like?

src/test/unit_tests/internal/test_block_store_s3.js
shirady, line 1: I'm not sure you need this test.
I think you can test only the helper function - in `test_noobaa_s3_client.test.js` (new describe)

src/agent/block_store_services/block_store_s3.js
nadavMiz: Error: Unexpected S3 getObject Body type: Object
    at s3_body_to_buffer (/home/nmizrahi/noobaa-core/src/sdk/noobaa_s3_client/noobaa_s3_client.js:164:11)
    at [eval]:3:1
    at runScriptInThisContext (node:internal/vm:218:10)
    at node:internal/process/execution:451:12
    at [eval]-wrapper:6:24
    at runScriptInContext (node:internal/process/execution:449:60)
    at evalFunction (node:internal/process/execution:283:30)
    at evalTypeScript (node:internal/process/execution:295:3)
    at node:internal/main/eval_string:71:3

src/agent/block_store_services/block_store_s3.js
```diff
@@ -87,8 +87,10 @@ class BlockStoreS3 extends BlockStoreBase {
                 Key: this.usage_path,
             });
 
+            // consumes getObject Body so SDK v3 does not hold the S3 connection.
```
shirady: Are you sure about this comment?
Anyway, I'm not sure it is needed.
As I understand it, you define the variable `body` since it is handled differently between the 2 SDK versions and use it more than once.

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -148,10 +148,27 @@ function add_response_header_capture(s3) {
...
+ * @returns {Promise<Buffer>}
+ */
+async function s3_body_to_buffer(body) {
+    // SDK v2 returns a Buffer
+    if (Buffer.isBuffer(body)) return body;
+    // SDK v3 returns a stream with transformToByteArray
+    if (body && typeof body.transformToByteArray === 'function') {
+        return Buffer.from(await body.transformToByteArray());
+    }
+    throw new Error(`Unexpected S3 getObject Body type: ${body && body.constructor && body.constructor.name}`);
```
shirady: This is a general function, and we might use it in other places in the future, suggesting not couple it to `getObject` in the logs.
```suggestion
    throw new Error(`Unexpected Body type: ${body && body.constructor && body.constructor.name}`);
```

BTW, will the extra arguments help in debugging? `${body && body.constructor && body.constructor.name}`
Update: I saw that you attached the logs in [comment](https://github.com/noobaa/noobaa-core/pull/9961#discussion_r3841664159).

shirady (approved):
LGTM
Added a few comments and questions

naveenpaul1 (approved):
LGTM

src/sdk/noobaa_s3_client/noobaa_s3_client.js
nadavMiz: they don't help by much. just to understand what we did get. we should never reach this point so though it would be good to know. just in case. I can remove it

src/agent/block_store_services/block_store_s3.js
nadavMiz: its because of code rabbits comment:

```
Drain res.Body in metadata mode.

When this.disable_metadata is false, init() reads res.Metadata without consuming res.Body. Consume or destroy the body stream to release the S3 connection.
```
the only deference is that we always consume the body even in the metadata option. but it looks like the comment is confusing. also I am not sure  if its even that much of a problem, because the body is emty anyway. I'll remove the comment

PR #9966, opened 2026-08-20
Pass content_type to read_object_mapping

karthikps97:
### Describe the Problem
The object's `content_type` was not being passed to `read_object_stream` which was resulting in errors in case of objects larger than 1MB.

### Explain the Changes
1. We already have `read_object_md_by_id` object API which returns object of type `ObjectInfo`. So remove the additional declaration.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Archiving now preserves complete source object metadata, improving consistency and retaining file information in archived objects.

### DIFF
--- a/src/server/bg_services/archive_server.js
+++ b/src/server/bg_services/archive_server.js
@@ -96,14 +96,6 @@ async function archive_object(req) {
             throw new Error('archive_object: object not found ' + obj_id);
         }
 
-        const object_md = {
-            obj_id: obj_md.obj_id,
-            bucket: obj_md.bucket,
-            key: obj_md.key,
-            size: obj_md.size,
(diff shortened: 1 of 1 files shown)


src/server/bg_services/archive_server.js
shirady: Why do you need to create the object?
You have `obj_md`.

karthikps97: Earlier the object was of type `ObjectMD` but I added another function which returns `ObjectInfo`. So its not needed anymore. Removed.

shirady (approved):
LGTM

PR #9967, opened 2026-08-21
Deep Archive | cleanup failed objects on crash

karthikps97:
### Describe the Problem
On process/pod crash, objects which were being processed by archive_server would not be picked up for retry since we select objects with no `transition_info`. 

### Explain the Changes
1. Add a function which unsets `transition_info` for all `in_progress` objects which have exceeded the configured cutoff time. This is called each time lifecycle worker runs and also during startup of the worker.
2. Add a timestamp field `in_progress_timestamp` to `transition_info` when setting the object to `in_progress` state. This field gets removed once the object is marked as `done`.
3. Add a new env variable which controls the cutoff time.
4. Add mdstore method and object_server API.
5. Remove unused env's.
6. Minor fixes.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Object transition details now show when processing began.
  * Administrators can clear transitions that have exceeded the eight-hour timeout.
* **Bug Fixes**
  * Stale in-progress lifecycle transitions are automatically cleaned up.
  * Lifecycle processing starts immediately with concurrency maintained at three.
  * Improved transition storage-class validation.
* **Tests**
  * Added coverage for stale, recent, completed, deleted, and mixed transition states.


src/api/object_api.js
```diff
@@ -1690,6 +1691,20 @@ module.exports = {
             },
         },
 
+        unset_transition_in_progress: {
```
romayalon, line 1694: I think that setting the transition_info status to FAILED makes more sense than unsetting it, when searching objects for transition we need to check if transition_info is null or transition_info.status = FAILED

src/server/object_services/schemas/object_md_schema.js
```diff
@@ -151,6 +151,7 @@ module.exports = {
             required: ['status'],
             properties: {
                 status: { $ref: 'common_api#/definitions/transition_status_enum' },
+                in_progress_timestamp: { date: true },
```
romayalon:
```suggestion
                transition_start_ts: { date: true },
```

src/api/object_api.js
karthikps97: Will do these changes in a separate PR to avoid too limit this PR's scope.

src/server/object_services/schemas/object_md_schema.js
karthikps97: Changed.

src/server/bg_services/lifecycle.js
```diff
@@ -20,6 +20,30 @@ const { STORAGE_CLASS_STANDARD } = require('../../endpoint/s3/s3_utils');
 /*************************/
 const ARCHIVE = COMMON_CONSTANTS.ARCHIVE;
 
+/**
+ * Cleans up lifecycle transitions that have remained in progress
+ * beyond the configured transition timeout.
+ *
+ * @returns {Promise<void>} Resolves when the cleanup operation completes.
+ */
+async function cleanup() {
```
romayalon, line 29: cleanup() is very vague, I would choose a more informative name
```suggestion
async function cleanup_idle_in_progress_transitions() {
```

src/sdk/nb.d.ts
```diff
@@ -450,6 +450,7 @@ interface ObjectMD {
 
 interface TransitionInfo {
     status: TransitionStatus;
+    in_progress_timestamp?: Date;
```
romayalon: transition_start_ts

src/server/bg_workers.js
```diff
@@ -197,6 +197,7 @@ function run_master_workers() {
         register_bg_worker({
             name: 'lifecycle',
             delay: config.LIFECYCLE_INTERVAL,
+            run_immediate: true,
```
romayalon, line 200: why?

src/server/bg_services/lifecycle.js
karthikps97: I kept this name since its a generic function in the framework. We can extend the same function for other similar use cases also in future. WDYT.

src/server/bg_workers.js
karthikps97: We would prefer it running the worker immediately rather than waiting for the default lifecycle interval. In a scenario where a pod crashed and there were too many objects in progress, we can reduce the load on DB by running this early on rather than just before the worker runs.

romayalon (approved):

src/server/bg_services/lifecycle.js
romayalon: yes, just add a todo that we need to add expiration flow in it

src/server/bg_services/lifecycle.js
```diff
@@ -419,7 +419,6 @@ async function background_worker() {
     try {
         dbg.log0('LIFECYCLE READ BUCKETS configuration: BEGIN');
         await system_store.refresh();
-        await cleanup();
```
romayalon: notice that when moving cleanup() from here you are skipping the system_store.refresh()

src/server/bg_services/lifecycle.js
```diff
@@ -20,6 +20,30 @@ const { STORAGE_CLASS_STANDARD } = require('../../endpoint/s3/s3_utils');
...
+ *
+ * @returns {Promise<void>} Resolves when the cleanup operation completes.
+ */
+async function cleanup() {
+    const now = Date.now();
+    const cutoff = now - config.LIFECYCLE_TRANSITION_TIMEOUT;
+    const system = system_store.data.systems[0];
+    if (!system) return;
+
+    const modified = await server_rpc.client.object.unset_transition_in_progress({
```
romayalon, line 47: running this query on every cycle without checking for buckets with transition rules is blocking the lifecycle expiration for no good reason

src/util/background_scheduler.js
```diff
@@ -66,11 +66,17 @@ class Background_Scheduler {
...
         dbg.log0('Registering', worker.name, 'bg worker');
         if (run_batch_function) {
             worker.run_batch = run_batch_function;
         }
+        if (pre_batch_fn) {
+            if (typeof pre_batch_fn !== 'function') {
+                throw new Error('passed argument pre_batch_fn is not a function');
+            }
+            worker.pre_batch_fn = pre_batch_fn;
+        }
```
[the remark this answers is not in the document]
karthikps97: If a falsy value is provided, then pre-batch function is never registered. This is acceptable.

src/server/bg_services/lifecycle.js
karthikps97: Added this check now. Thanks.

src/server/bg_services/lifecycle.js
karthikps97: Yes, I added a refresh within cleanup itself.

PR #9969, opened 2026-08-23
Deep Archive | Support multiple transitions

romayalon:
### Describe the Problem
AWS S3 deprecates singular Transition / NoncurrentVersionTransition actions per lifecycle rule and now allows Plural  Transitions / NoncurrentVersionTransitions actions per rule.
NooBaa missing validation for transition/NoncurrentVersionTransition actions - for example - archive policy missing on the bucket, storage class, Days vs Date, ordering vs expiration, duplicates, or related fields the way AWS does.

### Explain the Changes
1. common_api.js — store transitions and noncurrent_version_transitions as arrays (no migration; not used in production).
2. s3_put_bucket_lifecycle.js — parse every <Transition> / <NoncurrentVersionTransition> element and validate them against AWS (archive policy required, GLACIER/DEEP_ARCHIVE storage_class only, Days xor Date, mixed Days/Date, Expiration/NoncurrentVersionExpiration ordering, duplicate StorageClass, midnight UTC Date, non-negative Days, filter text in errors).
3. s3_get_bucket_lifecycle.js — return all stored transitions as repeated XML elements.
4. lifecycle.js — apply each transition / NoncurrentVersionTransition entry on the rule (still only DEEP_ARCHIVE and GLACIER).
5. Tests and design doc — multi-transition PUT/GET round-trip, PUT validation coverage, archive-policy rejects in test_s3_ops.js, and a note that a rule may include multiple Transitions/NoncurrentVersionTransitions.
6. 2 bug fixes -  
a. lifecycle didn't run when both transition and noncurrentversionTransition actions were set if the bucket's versioning state was disabled, the fix was to run only transition action.
b. transition date parsing called parse_lifecycle_field() instead of date: parse_date_field(tran.Date, 'Date').

1. 

### Testing Instructions:
auto -

npm run mocha -- --grep 'lifecycle-transitions'
npm run mocha -- --grep 'bucket-lifecycle'

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **New Features**
  - Added support for multiple current-version and noncurrent-version lifecycle transitions.
  - Lifecycle configurations now preserve and return all transition actions.
  - Lifecycle processing executes each configured transition independently.

- **Bug Fixes**
  - Improved validation for storage classes, dates, ordering, duplicates, filters, and archive-policy requirements.
  - Correctly supports valid zero-day transitions and rejects unsupported configurations with clear errors.

- **Documentation**
  - Updated lifecycle API documentation, CLI examples, and verification output for multiple transitions.

### DIFF
--- a/src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
+++ b/src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
@@ -3,6 +3,7 @@
 
 const _ = require('lodash');
 const s3_const = require('../s3_constants');
+const s3_utils = require('../s3_utils');
 const crypto = require('crypto');
 const dbg = require('../../../util/debug_module')(__filename);
 const S3Error = require('../s3_errors').S3Error;
@@ -10,6 +11,17 @@ const S3Error = require('../s3_errors').S3Error;
 const true_regex = /true/i;
 const MAX_LIFECYCLE_RULES = 1000; // AWS limit
 const MAX_TAGS_IN_AND_FILTER = 10;
+
+/**
+ * Lifecycle Transition / NoncurrentVersionTransition waterfall, earlier (warmer) to later (colder).
+ * A later class must use a strictly greater Days / Date / NoncurrentDays than an earlier class.
+ * Insert new transition targets here; do not special-case class names in the combination validators.
+ * @type {readonly nb.StorageClass[]}
+ */
+const LIFECYCLE_TRANSITION_STORAGE_CLASS_ORDER = Object.freeze([
+    s3_utils.STORAGE_CLASS_GLACIER,
+    s3_utils.STORAGE_CLASS_DEEP_ARCHIVE,
+]);
 
 /**
  * @param {*} field 
@@ -39,18 +51,10 @@ function validate_lifecycle_expiration_rule(rule, bucket_versioning) {
             throw new S3Error(S3Error.MalformedXML);
         }
         if (expiration_content.Date) {
-            const date = new Date(expiration_content.Date[0]);
-            if (isNaN(date.getTime()) || date.getTime() !== Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate())) {
-                dbg.error('Expiration Date value must conform to the ISO 8601 format and be at midnight UTC. Provided:', expiration_content.Date[0]);
-                throw new S3Error({ ...S3Error.InvalidArgument, message: "'Date' must be at midnight GMT" });
-            }
+            reject_if_not_midnight_utc(expiration_content.Date[0], 'Date');
         }
         if (expiration_content.Days) {
-            const days = parseInt(expiration_content.Days[0], 10);
-            if (isNaN(days) || days < 1) {
-                dbg.error('Minimum value for Expiration Days is 1, received:', expiration_content.Days[0]);
-                throw new S3Error({ ...S3Error.InvalidArgument, message: 'Expiration Days must be a positive integer' });
-            }
+            parse_positive_int(expiration_content.Days[0], 'Expiration Days');
         }
         if (expiration_content.ExpiredObjectDeleteMarker) {
             if (expiration_content.ExpiredObjectDeleteMarker[0].toLowerCase() !== 'true') {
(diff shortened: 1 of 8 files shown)


src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -450,6 +412,341 @@ async function put_bucket_lifecycle(req) {
...
+            throw new S3Error({ ...S3Error.InvalidArgument, message: "'Transition' action must specify either 'Days' or 'Date'" });
+        }
+        if (has_days_field) {
+            parse_non_negative_int(transition.Days[0], "'Days' in Transition action must be nonnegative");
+        } else {
+            reject_if_not_midnight_utc(transition.Date[0], 'Date');
+        }
+    }
+
+    if (rule.Transition.some(t => t.Days) && rule.Transition.some(t => t.Date?.[0])) {
```
karthikps97: Could you link the source where this scenario is mentioned?

romayalon: I tried it on AWS - 
Input - 
```
{
  "Rules": [
    {
      "ID": "duplicate-storage-class",
      "Status": "Enabled",
      "Filter": {
        "And": {
          "Prefix": "test/",
          "ObjectSizeLessThan": 120120
        }
      },
      "Transitions": [{ "Days": 1, "StorageClass": "DEEP_ARCHIVE" }, {"Date": "2027-01-01T00:00:00.000Z", "StorageClass": "GLACIER" }]
   }
  ]
}
```

Output - 
```
aws: [ERROR]: An error occurred (InvalidRequest) when calling the PutBucketLifecycleConfiguration operation (reached max retries: 1): Found mixed 'Date' and 'Days' based Transition actions in lifecycle rule for filter '(prefix=test/ and objectsizelessthan=120120)'
```

and also from docs - `However, you can't use both Date and Days in the same rule.`
https://docs.aws.amazon.com/AmazonS3/latest/userguide/lifecycle-configuration-examples.html

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -384,13 +362,7 @@ async function put_bucket_lifecycle(req) {
 
         // Parse Transition
         if (rule.Transition?.length > 0) {
```
karthikps97, line 375: shouldn't this be transitions?

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -450,6 +412,341 @@ async function put_bucket_lifecycle(req) {
...
+ * @throws {S3Error} MalformedXML when an element is empty; InvalidRequest on duplicate StorageClass
+ */
+function parse_transition_actions(transitions, filter) {
+    const parsed_items = transitions.map(tran => {
+        const parsed = _.omitBy({
+            storage_class: parse_lifecycle_field(tran.StorageClass, String),
+            date: parse_date_field(tran.Date, 'Date'),
+            days: parse_lifecycle_field(tran.Days),
+        }, _.isUndefined);
+        reject_empty_action_field(parsed, 'Transition');
```
karthikps97: Since we have all validations in `validate_lifecycle_rule` we would never have a scenario where the parsed object would be empty. Why do we need this?

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -450,6 +412,341 @@ async function put_bucket_lifecycle(req) {
...
+ *     are missing; InvalidRequest on duplicate StorageClass
+ */
+function parse_noncurrent_version_transition_actions(noncurrent_version_transitions, filter) {
+    const parsed_items = noncurrent_version_transitions.map(nvt => {
+        const parsed = _.omitBy({
+            storage_class: parse_lifecycle_field(nvt.StorageClass, String),
+            noncurrent_days: parse_lifecycle_field(nvt.NoncurrentDays),
+            newer_noncurrent_versions: parse_lifecycle_field(nvt.NewerNoncurrentVersions),
+        }, _.isUndefined);
+        reject_empty_action_field(parsed, 'NoncurrentVersionTransition');
```
karthikps97: Same here. Do we need this?

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
```diff
@@ -450,6 +412,341 @@ async function put_bucket_lifecycle(req) {
...
+        if (parsed.noncurrent_days === undefined || parsed.storage_class === undefined) {
```
karthikps97: Are we expecting storage class to be undefined even after validation? Also, we don't have a similar check in `parse_transition_actions`.

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
romayalon: The XML element name is still `<Transition>` (repeatable). xml2js keeps that as `rule.Transition` — an array when there are multiple. The AWS SDK JSON shape uses `Transitions`, but this endpoint parses the REST XML body, not the SDK shape.

The stored RPC/DB field is already plural: `current_rule.transitions`. Same pattern below for `rule.NoncurrentVersionTransition` → `noncurrent_version_transitions`.

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
romayalon: You're right — `validate_lifecycle_transition_rule` already rejects empty/incomplete Transition actions, so this could never fire. Removed it.

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
romayalon: Same as Transition — validation already covers this. Removed.

src/endpoint/s3/ops/s3_put_bucket_lifecycle.js
romayalon: No — after `validate_lifecycle_noncurrent_transition_rule`, StorageClass and NoncurrentDays are always present. This was leftover from before that validator existed (and yes, `parse_transition_actions` didn't have the same check). Removed it.

karthikps97 (approved):

PR #9981, opened 2026-08-25
AgentBlocksReclaimer should handle invalid blocks

karthikps97:
### Describe the Problem
In an earlier issue with db_cleaner we had scenario where chunks which were referenced by valid blocks were removed. This scenario causes a crash of AgentBlocksReclaimer when it tries to load the frags for the chunk.

### Explain the Changes
1. For reclaimer, we don't need the chunks and frags to be present. So conditionally include blocks which dont have these but prevent the chunks and frags from being initialized for these blocks.
2. For AgentBlocksVerifier, log an error for this scenario and skip those blocks since we dont want further processing for these blocks.

Note: Fixed an unrelated issue in agent_blocks_verifier that was causing DCO check to fail: `check condition '!blocks_with_alive_nodes' is always false. The value of variable 'blocks_with_alive_nodes' is originated from the return value of 'Array.prototype.filter()'`

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of blocks whose associated chunk or fragment data is unavailable.
  * Prevented errors when processing blocks with missing chunk references.
  * Empty blocks can now be retained when explicitly requested; otherwise, unmatched blocks continue to be skipped.

* **Tests**
  * Added unit and integration coverage for missing chunks, deleted chunks, and empty-block handling.

### DIFF
--- a/src/server/bg_services/agent_blocks_reclaimer.js
+++ b/src/server/bg_services/agent_blocks_reclaimer.js
@@ -113,7 +113,8 @@ class AgentBlocksReclaimer {
      * @override in unit tests for decoupling dependencies
      */
     populate_nodes_for_blocks(blocks) {
-        return map_server.prepare_blocks_from_db(blocks);
+        return map_server.prepare_blocks_from_db(blocks,
+            /*include_empty_blocks=*/true);
     }
 
 }
--- a/src/server/object_services/map_db_types.js
+++ b/src/server/object_services/map_db_types.js
@@ -318,8 +318,8 @@ class BlockDB {
             pool: optional_id_str(this.block_db.pool),
             address: this.node && this.node.rpc_address,
             size: this.block_db.size,
-            digest_type: this.digest_type || this.chunk.chunk_coder_config.frag_digest_type,
-            digest_b64: this.frag.digest_b64,
+            digest_type: this.digest_type || this.chunk?.chunk_coder_config?.frag_digest_type,
+            digest_b64: this.frag?.digest_b64,
             node_type: this.node && this.node.node_type,
             is_preallocated: this.is_preallocated,
         };
--- a/src/server/object_services/map_server.js
+++ b/src/server/object_services/map_server.js
@@ -629,19 +629,34 @@ async function prepare_blocks(blocks) {
 /**
  *
  * @param {nb.BlockSchemaDB[]} blocks
+ * @param {boolean} include_empty_blocks - include blocks with no valid chunks in return
  * @return {Promise<nb.Block[]>}
  */
-async function prepare_blocks_from_db(blocks) {
-    const chunk_ids = blocks.map(block => block.chunk);
+async function prepare_blocks_from_db(blocks, include_empty_blocks) {
+    const chunk_ids = blocks.map(block => block.chunk).filter(Boolean);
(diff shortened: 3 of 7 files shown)


src/server/object_services/map_db_types.js
```diff
@@ -318,8 +318,8 @@ class BlockDB {
             pool: optional_id_str(this.block_db.pool),
             address: this.node && this.node.rpc_address,
             size: this.block_db.size,
-            digest_type: this.digest_type || this.chunk.chunk_coder_config.frag_digest_type,
-            digest_b64: this.frag.digest_b64,
+            digest_type: this.digest_type || this.chunk?.chunk_coder_config.frag_digest_type,
```
dannyzaken: For completeness, handle `chunk_coder_config`

```suggestion
            digest_type: this.digest_type || this.chunk?.chunk_coder_config?.frag_digest_type,
```

src/server/object_services/map_server.js
```diff
@@ -629,19 +629,33 @@ async function prepare_blocks(blocks) {
...
-        const frag_db = _.find(chunk_db.frags, frag =>
-            frag._id.toHexString() === block.frag.toHexString());
+    const db_blocks = _.compact(blocks.map(block => {
+        let valid_block = true;
+        if (!block.chunk || !chunks_by_id[block.chunk.toHexString()]) {
+            valid_block = false;
+        }
+
+        if (!include_empty_blocks && !valid_block) {
+            dbg.error("skipping invalid block with no matching chunk", block);
```
dannyzaken: Use single quotes for strings


```suggestion
            dbg.error('skipping invalid block with no matching chunk', block);
```

src/server/bg_services/agent_blocks_reclaimer.js
```diff
@@ -113,7 +113,7 @@ class AgentBlocksReclaimer {
      * @override in unit tests for decoupling dependencies
      */
     populate_nodes_for_blocks(blocks) {
-        return map_server.prepare_blocks_from_db(blocks);
+        return map_server.prepare_blocks_from_db(blocks, true);
```
dannyzaken: comment the `true` to make the code clearer

```suggestion
        return map_server.prepare_blocks_from_db(blocks, /*include_empty_blocks=*/true);
```

src/server/object_services/map_server.js
```diff
@@ -629,19 +629,33 @@ async function prepare_blocks(blocks) {
 /**
  *
  * @param {nb.BlockSchemaDB[]} blocks
+ * @param {boolean} include_empty_blocks - include blocks with no valid chunks in return
  * @return {Promise<nb.Block[]>}
  */
-async function prepare_blocks_from_db(blocks) {
+async function prepare_blocks_from_db(blocks, include_empty_blocks) {
     const chunk_ids = blocks.map(block => block.chunk);
```
dannyzaken: filter undefined chunks 


```suggestion
async function prepare_blocks_from_db(blocks, include_empty_blocks) {
    const chunk_ids = blocks.map(block => block.chunk).filter(Boolean);
```

dannyzaken (approved):

PR #9983, opened 2026-08-26
DEEP_ARCHIVE | Fix - `list_objects_versions` To Omit `restore_status` Expiry Invalid or in the Past

shirady:
### Describe the Problem
Part of [RHSTOR-8823](https://redhat.atlassian.net/browse/RHSTOR-8823)
In case the object was already restored and its expiry date is in the past, and the BG worker that handles removing the STANDARD copy did not finish removing it, we would like to avoid list object versions calls that would include an expiry date in the past (property and header).

Note: This PR handles the GAP that was mentioned in #9962.

### Explain the Changes
1. Edit the function `list_object_versions ` instead of being just a passthrough to the metadata namespace to omit the `restore_status` in case the expiry is invalid or in the past, and use the helper `_omit_inactive_restore_status`.
2. Move the functions for parsing the header and adding the restore status from list objects to `s3_utils` and use them in list object versions op.
3. Added AI-generated tests.

### Issues:
List of GAPs:
1. We might want to move the decision to omit restore_status expiry from MSC to the op itself - it is mutual to head object, list object, and list object versions (see [comment](https://github.com/noobaa/noobaa-core/pull/9983#discussion_r3869830646)).

#### Unit Tests
Please run:
1. `npx jest src/test/unit_tests/internal/test_namespace_multi_storage_class_restore.test.js`
2. `npx jest src/test/unit_tests/util_functions_tests/test_s3_utils.test.js`

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added support for requesting object restore status when listing bucket versions.
  * Version responses now include restore status details, including completion state and optional expiration dates.
* **Bug Fixes**
  * Expired or incomplete restore information is no longer shown in object-version listings.
  * Invalid optional object-attribute requests are rejected with appropriate validation.
* **Tests**
  * Added coverage for restore-status parsing, response generation, and filtering across multiple restore scenarios.

### DIFF
--- a/src/endpoint/s3/s3_utils.js
+++ b/src/endpoint/s3/s3_utils.js
@@ -884,6 +884,46 @@ function parse_s3_restore_field(restore_field) {
 }
 
 
+/**
+ * Parses x-amz-optional-object-attributes and returns whether RestoreStatus was requested
+ * @param {import('http').IncomingHttpHeaders} headers
+ * @returns {boolean}
+ */
+function parse_optional_object_attributes_header(headers) {
+    const optional_object_attributes_header = headers['x-amz-optional-object-attributes'];
+    const optional_object_attributes = Array.isArray(optional_object_attributes_header) ?
+        optional_object_attributes_header[0] : optional_object_attributes_header;
+    const restore_status_requested = optional_object_attributes === 'RestoreStatus';
+
+    // Only RestoreStatus is a valid attribute for now
+    if (optional_object_attributes && !restore_status_requested) {
+        throw new S3Error({ ...S3Error.InvalidArgument, message: 'Invalid attribute name specified' });
+    }
+    return restore_status_requested;
+}
+
+/**
+ * Returns RestoreStatus XML fields when requested and object has restore_status
+ * @param {nb.ObjectInfo} obj
+ * @param {boolean} restore_status_requested
+ * @returns {{ IsRestoreInProgress: boolean | undefined, RestoreExpiryDate?: string } | undefined}
+ */
+function get_object_restore_status(obj, restore_status_requested) {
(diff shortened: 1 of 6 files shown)


src/endpoint/s3/s3_utils.js
```diff
@@ -884,6 +884,46 @@ function parse_s3_restore_field(restore_field) {
...
+    if (!restore_status_requested || !obj.restore_status) {
+        return;
+    }
+
+    /** @type {{ IsRestoreInProgress: boolean | undefined, RestoreExpiryDate?: string }} */
+    const restore_status = {
+        IsRestoreInProgress: obj.restore_status.ongoing,
+    };
+    if (!obj.restore_status.ongoing && obj.restore_status.expiry_time) {
+        restore_status.RestoreExpiryDate = new Date(obj.restore_status.expiry_time).toUTCString();
```
romayalon, line 921: why not omitting here if the restore is inactive?

romayalon (approved):

src/endpoint/s3/s3_utils.js
shirady: This function was moved from `src/endpoint/s3/ops/s3_get_bucket.js`, which we decided in the MSC implementation, so I added the implementation in that layer.
If we decide to change it, it would be in a follow-up PR.

@tangledbytes WDYT?

shirady: This function was moved from `src/endpoint/s3/ops/s3_get_bucket.js` (already created for tape originally).
We decided on the omiting in the MSC implementation, so I added the implementation in that layer.
If we decide to change it, it would be in a follow-up PR.

PR #9978, opened 2026-08-25
DEEP_ARCHIVE | Large Objects - `getObject` by Range

shirady:
### Describe the Problem
Implement the BG Worker Restore with `GetObject` to handle large objects that were added using MPU (multipart upload) - as part of [RHSTOR-8823](https://redhat.atlassian.net/browse/RHSTOR-8823)

### Explain the Changes
1. Edit `upload_object_range` in `src/sdk/object_io.js` so we can pass the seq (still with the default of 0 in case we didn't pass it).
2. Edit `read_archive_object_stream` and change the parameter from `size` to `start` and `end`. Please node that in namespace_s3 we didn't use the `size` `read_object_stream` and it was just a prepare for the read by range, when passing the `end` it uses the bytes range.
3. Add the config of `config.RESTORE_WORKER_RANGE_SIZE` to read the object stream by range using this size.
4. In the restore worker have a single function `_write_standard_restore_copy` (to avoid code duplication) for both cases, and separate the handling in the inner functions: small objects with `full_archive_read` and large objects to read in ranges, with the handling in case of failure during the write (overlap, hole). remove the previous implementation of `_has_complete_restore_copy` to avoid many return statements and to have one in `_get_restore_copy_coverage`. 
5.  Added a query `find_max_part_seq_for_object` that will be used in case we have issues and ne need to know what the last seq was that was written (to avoid loading all the parts for a single object).

### Issues: 
List of GAPs:
1. We used the `GetObject` with byte ranges, but we didn't enforce the deletion of the object.
When reading the object using `getObject` there is an option in the archive namespace to evict the `STANDARD` automatically; references:
https://github.com/noobaa/noobaa-core/blob/844c734133bb93ac21f4483558a0d90ccaa3b5ac/config.js#L1048-L1051
https://github.com/noobaa/noobaa-core/blob/844c734133bb93ac21f4483558a0d90ccaa3b5ac/src/sdk/namespace_fs.js#L1185-L1189
2. When I accidentally used the wrong etag in the file on complete MPU, I got `InterError` instead of `InvalidPart`. Updated - fix in PR #9994.
3. Again, I have the `TOCTOU race`; see [comment](https://github.com/noobaa/noobaa-core/pull/9978#discussion_r3870650606); as already mentioned in PR #9880  (see [comment](https://github.com/noobaa/noobaa-core/pull/9880#discussion_r3663661142)).
[...]

PR #9992, opened 2026-08-26
DEEP_ARCHIVE | Add force evict custom header

romayalon:
### Explain the Changes
1. namespace_fs - Added force evict custom header in namespace_fs that will call eviction on DBS3.
2. archive_server - Added force_evict_archive_object() that call getObject of 0 (for empty objects)/1 byte with force evict header so DBs3 will evict.
 

1. Gap - we should call force_evict_archive_object() on https://github.com/noobaa/noobaa-core/pull/9978 and add tests

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added support for a configurable request header that forces eviction of restored Glacier and deep-archive objects.
  * Force-eviction now works with full and partial object reads, including archive-backed reads.
  * Added support for custom HTTP headers on object read requests.

* **Bug Fixes**
  * Improved expiration handling after Glacier restore reads.
  * Improved error reporting for mismatched upload-part identifiers.

* **Tests**
  * Added coverage for force-eviction during one-byte ranged reads.

### DIFF
--- a/src/endpoint/s3/ops/s3_get_object.js
+++ b/src/endpoint/s3/ops/s3_get_object.js
@@ -7,6 +7,7 @@ const s3_utils = require('../s3_utils');
 const http_utils = require('../../../util/http_utils');
 const rdma_utils = require('../../../util/rdma_utils');
 const { throw_if_restore_incomplete } = require('../../../util/deep_archive_utils');
+const config = require('../../../../config');
 
 /* eslint-disable max-statements */
 
@@ -73,6 +74,9 @@ async function get_object(req, res) {
     }
     if (part_number) {
         params.part_number = part_number;
+    }
+    if (req.headers[config.NSFS_GLACIER_FORCE_EVICT_HTTP_HEADER] === 'true') {
+        params.glacier_force_evict = true;
     }
     try {
         const ranges = http_utils.normalize_http_ranges(
--- a/src/sdk/namespace_s3.js
+++ b/src/sdk/namespace_s3.js
@@ -261,7 +261,13 @@ class NamespaceS3 {
         this._set_md_conditions(params, request);
         this._assign_encryption_to_request(params, request);
         try {
-            const obj_out = await this.s3.getObject(request);
+            let obj_out;
+            if (params.glacier_force_evict) {
+                const extra_headers = {[config.NSFS_GLACIER_FORCE_EVICT_HTTP_HEADER]: 'true' };
+                obj_out = await noobaa_s3_client.get_object_with_headers(this.s3, request, extra_headers);
+            } else {
+                obj_out = await this.s3.getObject(request);
+            }
             dbg.log0('NamespaceS3.read_object_stream:',
                         this.bucket,
                         inspect(_.omit(params, 'object_md.ns')),
--- a/src/sdk/noobaa_s3_client/noobaa_s3_client.js
+++ b/src/sdk/noobaa_s3_client/noobaa_s3_client.js
@@ -4,6 +4,7 @@
 const _ = require('lodash');
 const http = require('http');
 const https = require('https');
+const { GetObjectCommand } = require('@aws-sdk/client-s3');
 const { S3ClientSDKV2 } = require('./noobaa_s3_client_sdkv2');
 const { S3ClientAutoRegion } = require('./noobaa_s3_client_sdkv3');
 const { NodeHttpHandler } = require("@smithy/node-http-handler");
@@ -149,6 +150,43 @@ function add_response_header_capture(s3) {
 }
 
 /**
+ * Adds custom HTTP headers to an AWS SDK v3 command via build-step middleware.
+ * Use for headers that are not part of the S3 command input shape.
+ * @param {{ middlewareStack: { add: Function } }} cmd
+ * @param {Record<string, string>} [headers]
+ */
+function add_command_headers(cmd, headers) {
+    if (!headers || !Object.keys(headers).length) return;
+    cmd.middlewareStack.add(
+        next => async args => {
+            Object.assign(args.request.headers, headers);
+            return next(args);
+        },
+        { step: 'build', name: 'addCustomHeaders', priority: 'low' }
+    );
+}
+
+/**
+ * getObject that can attach extra HTTP headers on both SDK v2 and v3.
+ * v2: AWS.Request 'build' event. v3: GetObjectCommand middleware + send().
+ * note that getObjectWithHeaders is a function that we created in S3ClientSDKV2 (not derived directly from the SDK.
+ * @param {object} s3 - client returned by get_s3_client_v3_params
+ * @param {object} request - GetObject params
+ * @param {Record<string, string>} [headers]
+ */
+async function get_object_with_headers(s3, request, headers) {
+    if (s3 instanceof S3ClientSDKV2) {
+        return s3.getObjectWithHeaders(request, headers);
+    }
+    if (headers && Object.keys(headers).length) {
+        const cmd = new GetObjectCommand(request);
+        add_command_headers(cmd, headers);
+        return s3.send(cmd);
+    }
+    return s3.getObject(request);
+}
+
+/**
  * Converts AWS SDK getObject Body to a Buffer.
  * SDK v2 returns a Buffer. SDK v3 returns a stream with transformToByteArray.
  * @param {*} body - getObject Body from AWS SDK v2 or v3
@@ -171,4 +209,5 @@ exports.get_sdk_class_str = get_sdk_class_str;
 exports.fix_error_object = fix_error_object;
 exports.get_requestHandler_with_suitable_agent = get_requestHandler_with_suitable_agent;
 exports.add_response_header_capture = add_response_header_capture;
+exports.get_object_with_headers = get_object_with_headers;
 exports.s3_body_to_buffer = s3_body_to_buffer;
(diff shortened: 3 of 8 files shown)


src/sdk/namespace_s3.js
```diff
@@ -261,7 +261,10 @@ class NamespaceS3 {
         this._set_md_conditions(params, request);
         this._assign_encryption_to_request(params, request);
         try {
-            const obj_out = await this.s3.getObject(request);
+            const cmd = new GetObjectCommand(request);
+            const extra_headers = params.glacier_force_evict ? { [config.NSFS_GLACIER_FORCE_EVICT_HTTP_HEADER]: 'true' } : undefined;
+            noobaa_s3_client.add_command_headers(cmd, extra_headers);
+            const obj_out = await this.s3.send(cmd);
```
shirady: I'm not sure it is correct, and I saw that there was code implemented in this file that is using AWS SDK V3 only in:
https://github.com/noobaa/noobaa-core/blob/62356ae384400dde5689ea01f5850705260c8494/src/sdk/namespace_s3.js#L359-L371

The use of `noobaa_s3_client` is to allow the use of the 2 SDK versions.
and this code is only for AWS SDK V3, right?

src/server/bg_services/archive_server.js
```diff
@@ -453,8 +457,28 @@ async function abort_archive_multipart_upload(req) {
     }
 }
 
+/**
+ * Force-evict the restored archive object via a 1-byte GetObject with glacier_force_evict.
+ * Side-effect only; drains and discards the body.
+ * @param {{ bucket_id: string|nb.ID, obj_id: string|nb.ID, size?: number }} params
+ */
+async function force_evict_archive_object({ bucket_id, obj_id, size = 0 }) {
+    // Empty objects cannot use a byte range (S3 416); `end: 0` is falsy so no Range is sent.
```
shirady: What do you mean `(S3 416)`?

src/server/bg_services/archive_server.js
```diff
@@ -453,8 +457,28 @@ async function abort_archive_multipart_upload(req) {
     }
 }
 
+/**
+ * Force-evict the restored archive object via a 1-byte GetObject with glacier_force_evict.
+ * Side-effect only; drains and discards the body.
+ * @param {{ bucket_id: string|nb.ID, obj_id: string|nb.ID, size?: number }} params
+ */
+async function force_evict_archive_object({ bucket_id, obj_id, size = 0 }) {
```
shirady: Currently is it not called, right?
The call will be done once we have the get by range.

src/server/bg_services/archive_server.js
```diff
@@ -453,8 +457,28 @@ async function abort_archive_multipart_upload(req) {
...
+    // Empty objects cannot use a byte range (S3 416); `end: 0` is falsy so no Range is sent.
+    const bytes_to_read = size > 0 ? 1 : 0;
+    const params = { bucket_id, obj_id, size: bytes_to_read, start: 0, end: bytes_to_read, glacier_force_evict: true };
+    const source_stream = await read_archive_object_stream(params);
+    try {
+        await buffer_utils.read_stream_join(source_stream);
+    } catch (err) {
+        dbg.error('force_evict_archive_object failed', { bucket_id, obj_id }, err);
+        source_stream.destroy();
```
shirady: Does it wrap what you wanted in the try-catch?
You want to call `source_stream.destroy` only on failures in `buffer_utils.read_stream_join`?

src/sdk/namespace_s3.js
romayalon: yes, Initially I thought of doing it for both v2 and v3, but then I saw that a year ago we changed upload_object to also work only with v3, so I had a chat with Nimrod on that and said that if this is the case for upload we can do the same for getObject, WDYT? I don't mind adding it for v2 as well but seems like it's not really needed

src/server/bg_services/archive_server.js
romayalon: correct

src/server/bg_services/archive_server.js
romayalon: it means S3 will throw 416 error code - InvalidRange

src/sdk/namespace_s3.js
shirady: If you can have them both, I think it would be better in the scope of this PR (so the `noobaa_s3_client` continues to support both versions).
If you can't, I would recommend adding a comment for others to understand.

If there is a decision not to use the previous SDK (AWS SDK V2) anymore, then we can remove the `noobaa_s3_client` and use the AWS SDK V3 directly, but that is out of scope for this PR.

src/server/bg_services/archive_server.js
shirady: Thanks, so suggesting to write it:

> Empty objects cannot use a byte range (S3 will throw 416 error code - InvalidRange)

src/server/bg_services/archive_server.js
romayalon: added thanks

src/server/bg_services/archive_server.js
romayalon: np, added

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -171,4 +208,6 @@ exports.get_sdk_class_str = get_sdk_class_str;
 exports.fix_error_object = fix_error_object;
 exports.get_requestHandler_with_suitable_agent = get_requestHandler_with_suitable_agent;
 exports.add_response_header_capture = add_response_header_capture;
+exports.add_command_headers = add_command_headers;
```
shirady: Since we don't use it outside anymore, I think it can be removed from the exported function.

src/sdk/noobaa_s3_client/noobaa_s3_client.js
```diff
@@ -148,6 +149,42 @@ function add_response_header_capture(s3) {
...
+/**
+ * getObject that can attach extra HTTP headers on both SDK v2 and v3.
+ * v2: AWS.Request 'build' event. v3: GetObjectCommand middleware + send().
+ * @param {object} s3 - client returned by get_s3_client_v3_params
+ * @param {object} request - GetObject params
+ * @param {Record<string, string>} [headers]
+ */
+async function get_object_with_headers(s3, request, headers) {
+    if (s3 instanceof S3ClientSDKV2) {
+        return s3.getObjectWithHeaders(request, headers);
```
shirady, line 179: I would add a comment that is a function that we created in S3ClientSDKV2 (not derived directly from the SDK.

src/server/bg_services/archive_server.js
```diff
@@ -453,8 +457,34 @@ async function abort_archive_multipart_upload(req) {
...
+    if (typeof size !== 'number' || size < 0) {
+        throw new Error('force_evict_archive_object: size is required');
+    }
+    const range = size > 0 ? { start: 0, end: 1 } : {};
+    const params = { bucket_id, obj_id, size, glacier_force_evict: true, ...range };
+    let source_stream;
+    try {
+        source_stream = await read_archive_object_stream(params);
+        const drained = stream.promises.finished(source_stream);
+        source_stream.resume();
```
shirady: Can you explain this part?

src/endpoint/s3/ops/s3_get_object.js
```diff
@@ -74,6 +75,9 @@ async function get_object(req, res) {
     if (part_number) {
         params.part_number = part_number;
     }
+    if (req.headers[config.NSFS_GLACIER_FORCE_EVICT_HTTP_HEADER]) {
+        params.glacier_force_evict = true;
```
shirady, line 79: I think we need to check that `=== 'true'` before setting this to true.

shirady (approved):
LGTM

src/server/bg_services/archive_server.js
```diff
@@ -453,8 +457,34 @@ async function abort_archive_multipart_upload(req) {
     }
 }
 
+/**
+ * Force-evict the restored archive object via a 1-byte GetObject with glacier_force_evict.
+ * Side-effect only; drains and discards the body.
+ * Empty objects cannot use a byte range (S3 416 InvalidRange); start/end are omitted so no Range is sent.
+ * @param {{ bucket_id: string|nb.ID, obj_id: string|nb.ID, size: number }} params
+ */
+async function force_evict_archive_object({ bucket_id, obj_id, size }) {
```
tangledbytes: Where is this being used?

romayalon: it'll be used in restore_worker after read_archive_object_stream, or inside it
but not in this PR

src/server/bg_services/archive_server.js
romayalon: Those two lines drain a readable stream you do not care about, then wait until that drain is done.
 The GET is only there to trigger force-evict (glacier_force_evict). The body (1/0 bytes) is discarded.

1. stream.promises.finished(source_stream)

This attaches end / finish / error listeners before any data flows, and returns a promise:

resolves when the stream is fully consumed (or closed cleanly)
rejects if the stream errors
You must subscribe first. This GET is often 1 byte. If you resume() first, the stream can already have ended before you start waiting, and you can miss the event or hang.

2. source_stream.resume()

A readable stream starts paused. With no 'data' listener, resume() puts it in flowing mode so Node pulls chunks and drops them. That is the usual “discard the body” pattern.

You still have to drain. Leaving the HTTP body unread can stall the connection (backpressure / keep-alive) even though you never look at the bytes.

3. await drained

Wait until discard finishes (or fails). Only then is the GET complete.

Together: “start watching for completion, throw away the body, wait until that’s done.” The catch then destroy()s the stream if something goes wrong so the socket is not left hanging.

PR #9995, opened 2026-08-27
block put bucket logging on nc

sakshimunjal:
### Describe the Problem
https://redhat.atlassian.net/browse/DFBUGS-9794
NooBaa NC was accepting PutBucketLogging requests even though bucket logging is not supported end to end in that environment. This misled clients into thinking logging was enabled, and later S3 operations could hit the missing rpc_client logging path and emit internal errors.

### Explain the Changes
1. Added an NC-specific guard in PutBucketLogging so enable requests now return NotImplemented instead of persisting a misleading logging configuration.
2. Skip syslog delivery when rpc_client or its syslog router is unavailable, preventing the null-RPC crash seen on later S3 operations.

1. 

### Testing Instructions:
1. Start an NC nsfs endpoint and create a source bucket and a log bucket with the same account.
2. Run aws --endpoint-url https://<nc-endpoint> --no-verify-ssl s3api put-bucket-logging --bucket <source-bucket> --bucket-logging-status '{"LoggingEnabled":{"TargetBucket":"<log-bucket>","TargetPrefix":"logs/"}}' and confirm the request returns NotImplemented.
3. Run the same request against a nonexistent source bucket and confirm it still returns the bucket-not-found error instead of NotImplemented.
4. If a bucket has stale logging configuration from before this change, run a follow-up S3 operation such as get-object or get-bucket-logging and confirm the request no longer produces a Cannot read properties of null (reading 'rpc') error in the endpoint logs.
5. Run the same PutBucketLogging enable request against a hosted/ODF deployment and confirm the operation still succeeds there.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when syslog logging is unavailable by safely skipping the operation and recording a warning.
  * Bucket logging enablement now clearly reports that the feature is not implemented instead of accepting and storing configuration.

### DIFF
--- a/src/sdk/bucketspace_fs.js
+++ b/src/sdk/bucketspace_fs.js
@@ -675,22 +675,10 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
         try {
             const { name, logging } = params;
             dbg.log0('BucketSpaceFS.put_bucket_logging: Bucket name, logging', name, logging);
-            const bucket = await this.config_fs.get_bucket_by_name(name);
-            bucket.logging = logging;
 
-            let target_bucket;
-            try {
-                target_bucket = await this.config_fs.get_bucket_by_name(logging.log_bucket);
-            } catch (err) {
-                dbg.error('ERROR with reading TARGET BUCKET data', logging.log_bucket, err);
-                if (err.code === 'ENOENT') throw new RpcError('INVALID_TARGET_BUCKET', 'The target bucket for logging does not exist');
-                throw err;
-            }
-            if (target_bucket.owner_account !== bucket.owner_account) {
-                dbg.error('TARGET BUCKET NOT OWNED BY USER', target_bucket, bucket);
-                throw new RpcError('INVALID_TARGET_BUCKET', 'The owner for the bucket to be logged and the target bucket must be the same');
-            }
-            await this.config_fs.update_bucket_config_file(bucket);
+            // NC does not support bucket logging delivery yet, so reject enable requests
+            // instead of persisting a misleading logging configuration.
+            throw new RpcError('NOT_IMPLEMENTED', 'Bucket logging is not supported in NC');
         } catch (err) {
             throw translate_error_codes(err, entity_enum.BUCKET);
         }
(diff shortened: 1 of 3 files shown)


src/endpoint/s3/ops/s3_put_bucket_logging.js
```diff
@@ -1,14 +1,22 @@
...
 
 /**
  * http://docs.aws.amazon.com/AmazonS3/latest/API/RESTBucketPUTlogging.html
  */
 async function put_bucket_logging(req) {
     const logging = s3_utils.parse_body_logging_xml(req);
+    const is_nc = is_nc_environment();
+
     if (logging.log_bucket) {
+        if (is_nc) {
```
jackyalbo: Can we move the NC block into BucketSpaceFS.put_bucket_logging instead of s3_put_bucket_logging.js? That file is already NC-only, so we would not need is_nc_environment()
I think we can add a comment there and say that we currently decided not to support it in nc.

src/endpoint/s3/ops/s3_put_bucket_logging.js
```diff
@@ -8,6 +8,7 @@ const s3_utils = require('../s3_utils');
  */
 async function put_bucket_logging(req) {
     const logging = s3_utils.parse_body_logging_xml(req);
+
```
jackyalbo: remove this

src/sdk/bucketspace_fs.js
```diff
@@ -676,6 +676,14 @@ class BucketSpaceFS extends BucketSpaceSimpleFS {
             const { name, logging } = params;
             dbg.log0('BucketSpaceFS.put_bucket_logging: Bucket name, logging', name, logging);
             const bucket = await this.config_fs.get_bucket_by_name(name);
+
+            // NC does not support bucket logging delivery yet, so reject enable requests
+            // instead of persisting a misleading logging configuration.
+            if (logging?.log_bucket) {
```
jackyalbo: We can probably remove this check no? if we got this far we will just throw

jackyalbo (commented):
two small comments - also we probably need to update the test file src/test/unit_tests/nsfs/test_bucketspace_fs.js

src/sdk/bucketspace_fs.js
sakshimunjal: yes, just added it to be on safe side in case we need to run the further code in some cases
If you suggest, will remove this and code after line 686 as well since that would become unreachable

jackyalbo (approved):

Issue #9998, opened 2026-08-31
PutObject with an invalid storage class (e.g. STD1) does not return InvalidStorageClass


romayalon:
## Description

`PutObject` with an unrecognized `x-amz-storage-class` (for example `STD1`) on a regular placement bucket (`first.bucket`) does not return S3 `InvalidStorageClass`.

AWS S3 rejects unknown storage classes with:

- HTTP `400`
- Code `InvalidStorageClass`
- Message `The storage class you specified is not valid.`

NooBaa should do the same.

## Repro

```bash
echo hello > obj1.txt
aws s3api put-object --bucket first.bucket --key obj1 --body obj1.txt --storage-class STD1
```

**Expected:** `InvalidStorageClass` (400).

**Actual:** not `InvalidStorageClass` (generic `Error` from `parse_storage_class` is mapped by `s3_rest` to `InternalError` 500).

## Notes

`src/endpoint/s3/s3_utils.js` `parse_storage_class()` throws:

```js
throw new Error(`No such s3 storage class ${storage_class}`);
```

That is not an `S3Error`. `_prepare_error` in `s3_rest.js` only maps `S3Error` or `err.rpc_code`; anything else becomes `InternalError`.

NSFS already throws `S3Error.InvalidStorageClass` from `_throw_if_storage_class_not_supported`. The S3 header parser should do the same for unknown classes on PutObject / CreateMultipartUpload / CopyObject.

PR #9999, opened 2026-08-31
Deep Archive | Throw error on invalid storage class when bucket has no archive policy

romayalon:
### Describe the Problem
NooBaa didn't validate for creating archive objects on buckets without an archive policy. The original implementation of DBs3 was supposed to utilize data buckets but it's not in use per moving to using nsfs.

### Explain the Changes
1. object_server.js - added a check when creating an object on deep archive/glacier storageclass the bucket must have archive policy and added a config for disabling this check.
2. Added missing INVALID_STORAGE_CLASS S3error which was missing and opened an issue for other cases we should throw invalidStorageClass - https://github.com/noobaa/noobaa-core/issues/9998.

1. Fixed https://redhat.atlassian.net/browse/DFBUGS-9710

### Testing Instructions:
1. Added tests.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added validation for Glacier and Deep Archive storage classes during S3 uploads, copies, and multipart uploads.
  * Buckets without an active archive policy reject these requests with an `InvalidStorageClass` error.
  * Added an option to disable archive storage-class validation when needed.
* **Bug Fixes**
  * Improved S3 error reporting for invalid storage-class requests.

### DIFF
--- a/config.js
+++ b/config.js
@@ -536,7 +536,15 @@ config.MULTIPART_CONCURRENCY = 10;
 /////////////////////
 // ARCHIVE CHECK   //
 /////////////////////
+
+// When true, create_namespace_resource with archive=true HeadBuckets the target
+// and requires x-noobaa-available-storage-classes to include GLACIER.
 config.ARCHIVE_TARGET_BUCKET_CHECK_ENABLED = true;
+
+// When true, create_object_upload rejects GLACIER/DEEP_ARCHIVE unless the bucket
+// has archive_policy.deep_archive_resource. Set false to skip that check
+// (tests, or systems that use GLACIER on data buckets — never released).
+config.ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED = true;
 
 config.ARCHIVE_NS_CACHE_MAX_USAGE = 100;
 config.ARCHIVE_NS_CACHE_EXPIRY_MS = 10 * 60 * 1000;
--- a/src/server/object_services/object_server.js
+++ b/src/server/object_services/object_server.js
@@ -36,7 +36,7 @@ const { ChunkAPI } = require('../../sdk/map_api_types');
 const config = require('../../../config');
 const CONSTANTS = require('../../common/constants');
 const Quota = require('../system_services/objects/quota');
-const { STORAGE_CLASS_STANDARD } = require('../../endpoint/s3/s3_utils');
+const { STORAGE_CLASS_STANDARD, GLACIER_STORAGE_CLASSES } = require('../../endpoint/s3/s3_utils');
 const { is_expired_restore_pending_purge, is_transition_source_pending_purge } = require('../../util/deep_archive_utils');
 
 // short living cache for objects
@@ -61,6 +61,12 @@ async function create_object_upload(req) {
     throw_if_maintenance(req);
     load_bucket(req);
     check_quota(req.bucket);
+    // Glacier/Deep Archive objects are unreadable without an archive target and restore path.
+    if (config.ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED && GLACIER_STORAGE_CLASSES.includes(req.rpc_params.storage_class) &&
+            !req.bucket.archive_policy?.deep_archive_resource) {
+        throw new RpcError('INVALID_STORAGE_CLASS',
+            `The storage class you specified is not valid. ${req.rpc_params.storage_class} requires the bucket to have an archive policy.`);
+    }
 
     const encryption = _get_encryption_for_object(req);
     const obj_id = MDStore.instance().make_md_id();
--- a/src/test/integration_tests/api/s3/test_deep_archive_via_s3.js
+++ b/src/test/integration_tests/api/s3/test_deep_archive_via_s3.js
@@ -382,6 +382,8 @@ async function backdate_object(obj_id, age_days) {
 }
 
 mocha.describe('deep_archive_via_s3', function() {
+    let original_archive_policy_check_enabled;
+
     mocha.before(async function() {
         const account_info = await rpc_client.account.read_account({ email: EMAIL });
         const credentials = {
@@ -399,6 +401,8 @@ mocha.describe('deep_archive_via_s3', function() {
         });
 
         config.ARCHIVE_TARGET_BUCKET_CHECK_ENABLED = false;
+        original_archive_policy_check_enabled = config.ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED;
+        config.ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED = false;
         await s3.createBucket({ Bucket: ARCHIVE_TARGET_BUCKET });
         await rpc_client.account.add_external_connection({
             name: ARCHIVE_CONNECTION,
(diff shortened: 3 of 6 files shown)


config.js
```diff
@@ -539,6 +539,10 @@ config.MULTIPART_CONCURRENCY = 10;
 /////////////////////
 config.ARCHIVE_TARGET_BUCKET_CHECK_ENABLED = true;
 
+// When true, create_object_upload rejects GLACIER/DEEP_ARCHIVE unless the bucket
+// has archive_policy.deep_archive_resource. Set false to skip that check.
+config.ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED = true;
```
shirady, line 547: What is the case where you want this to be configurable?

src/server/object_services/object_server.js
```diff
@@ -61,6 +61,11 @@ async function create_object_upload(req) {
     throw_if_maintenance(req);
     load_bucket(req);
     check_quota(req.bucket);
+    // Glacier/Deep Archive objects are unreadable without an archive target and restore path.
+    if (config.ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED && GLACIER_STORAGE_CLASSES.includes(req.rpc_params.storage_class) &&
+            !req.bucket.archive_policy?.deep_archive_resource) {
```
shirady, line 66: For checking that the bucket has an archive policy you must use this `req.bucket.archive_policy?.deep_archive_resource`? `req.bucket.archive_policy` is not enough?

src/server/object_services/object_server.js
```diff
@@ -61,6 +61,11 @@ async function create_object_upload(req) {
...
+        throw new RpcError('INVALID_STORAGE_CLASS', 'The storage class you specified is not valid.');
```
shirady: I think you can add more details in the message, since in the end it would be the s3error message (after it would be mapped).

config.js
```diff
@@ -539,6 +539,10 @@ config.MULTIPART_CONCURRENCY = 10;
 /////////////////////
 config.ARCHIVE_TARGET_BUCKET_CHECK_ENABLED = true;
```
shirady, line 542: Could you add a comment about this config as well?
I'm not sure from the name I would know where to use this `ARCHIVE_TARGET_BUCKET_CHECK_ENABLED`.

What if `ARCHIVE_TARGET_BUCKET_CHECK_ENABLED` is `false` and `ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED` is `true`?

src/test/integration_tests/api/s3/test_deep_archive_via_s3.js
```diff
@@ -399,6 +399,7 @@ mocha.describe('deep_archive_via_s3', function() {
         });
 
         config.ARCHIVE_TARGET_BUCKET_CHECK_ENABLED = false;
+        config.ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED = false;
```
shirady, line 405: Why `false` here?
Maybe set in the relevant tests only?

src/test/integration_tests/api/s3/test_deep_archive_via_s3.js
```diff
@@ -505,6 +507,101 @@ mocha.describe('deep_archive_via_s3', function() {
...
+                        Bucket: bucket,
+                        Key: 'no-archive/mpu',
+                        ContentType: 'application/octet-stream',
+                        StorageClass: s3_utils.STORAGE_CLASS_DEEP_ARCHIVE,
+                    }),
+                    err => err_code(err) === 'InvalidStorageClass'
+                );
+            });
+
+            mocha.it(`allows STANDARD PutObject on ${bucket}`, async function() {
```
shirady: Are you sure you need this test case in this file? Is it something that should have been tested in the general S3 ops? Or am I missing something?

src/test/integration_tests/api/s3/test_deep_archive_via_s3.js
shirady, line 1: Why do we need both cases in `src/test/integration_tests/api/s3/test_deep_archive_via_s3.js` and `src/test/unit_tests/internal/test_deep_archive_s3.js`?

src/test/unit_tests/internal/test_deep_archive_s3.js
```diff
@@ -216,6 +216,72 @@ mocha.describe('archive_policy', function() {
         info = await rpc_client.bucket.read_bucket({ name: UPDATE_ARCHIVE_BUCKET });
         assert.ok(!info.archive_policy);
     });
+
+    mocha.it('should reject create_object_upload with DEEP_ARCHIVE when bucket never had archive_policy', async function() {
+        await assert.rejects(
+            rpc_client.object.create_object_upload({
+                bucket: 'first.bucket',
```
shirady, line 306: I think it would be easier to understand that by creating the bucket and then `create_object_upload` (like you already had in the previous file).

src/test/unit_tests/internal/test_deep_archive_s3.js
```diff
@@ -216,6 +216,72 @@ mocha.describe('archive_policy', function() {
         info = await rpc_client.bucket.read_bucket({ name: UPDATE_ARCHIVE_BUCKET });
         assert.ok(!info.archive_policy);
     });
+
+    mocha.it('should reject create_object_upload with DEEP_ARCHIVE when bucket never had archive_policy', async function() {
```
shirady, line 303: To align with the next test case, suggest adding the last part and also setting the config explicitly in this test case.
```suggestion
    mocha.it('should reject create_object_upload with DEEP_ARCHIVE when bucket never had archive_policy when check is disabled', async function() {
```

src/test/unit_tests/internal/test_deep_archive_s3.js
```diff
@@ -216,6 +216,72 @@ mocha.describe('archive_policy', function() {
...
+                bucket: 'first.bucket',
+                key: 'archive-test.txt',
+                storage_class: s3_utils.STORAGE_CLASS_DEEP_ARCHIVE,
+            }),
+            err => err.rpc_code === 'INVALID_STORAGE_CLASS'
+        );
+    });
+
+    mocha.it('should allow create_object_upload with DEEP_ARCHIVE without archive_policy when check is disabled', async function() {
+        const prev = config.ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED;
```
shirady: I think that in other cases you set it directly in the `mocha.after`, I think saving the previous one is preferable. Suggesting improving the variable name  (`prev`).

shirady (approved):
LGTM

src/server/object_services/object_server.js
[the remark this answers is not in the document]
romayalon: the check is intentionally on the requested storage class (Deep Archive contract). select_tier_for_write stamping a non-STANDARD class is the TMFS/DBs3 glacier write-tier path, which is not gated on archive_policy and should stay that way.

romayalon: in the future, archive policy might contain other archive storageclasses which are not deep archive

src/test/integration_tests/api/s3/test_deep_archive_via_s3.js
romayalon: all those tests use a noobaa bucket as an archive bucket so it's needed, we would like to allow writing to this bucket and not fail for invalid storage class

src/test/integration_tests/api/s3/test_deep_archive_via_s3.js
romayalon: test_deep_archive_via_s3 - integration tests that use s3 client
test_deep_archive_s3 - tests functions related to deep archive s3

src/test/integration_tests/api/s3/test_deep_archive_via_s3.js
romayalon: removed

config.js
romayalon: I added some description in the comment

karthikps97 (approved):

Issue #7448, opened 2023-08-17
decode_json / encode_json functions should work with oneOf / anyOf schema options 


nadavMiz:
### Environment info
- NooBaa Version: master

### Actual behavior
1. currently encode_json decode_json functions in postgres_client.js don't deal with oneOf / anyOf case of schemas. this may lead to objects using those schemas not being properly stored/retrieved from the Postgres database 

### Expected behavior
1. encode_json/decode_json should parse schemas with oneOf/anyOf cases.

PR #10002, opened 2026-08-31
Unify required params by identity_type in account schema

aayushchouhan09:
### Describe the Problem
Required params were not consistently enforced per identity type (ACCOUNT, USER, ROLE) in the shared account schema.

### Explain the Changes
1. Added identity-based required rules in `account_schema` so each identity type enforces only its relevant required fields.

// TODO
2. NC related changes - (will cover in another PR)

1.  Gap (IAM Roles schema refactor PR): #9916 

### Testing Instructions:
1.  To run test: ` npx jest src/test/unit_tests/internal/test_account_schema_validation.test.js`

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved account record validation by distinguishing `ACCOUNT`, `USER`, and `ROLE` identities.
  - Enforced identity-specific required fields and values for more accurate validation.
  - Preserved support for legacy records that do not specify an identity type.
  - Added support for validating schemas with multiple alternative formats and explicit exclusions.


src/server/system_services/schemas/account_schema.js
```diff
@@ -139,5 +130,46 @@ module.exports = {
...
+        required: ['_id', 'name', 'owner', 'assume_role_policy_document', 'identity_type'],
+        properties: {
+            ...account_properties,
+            identity_type: {
+                type: 'string',
+                enum: ['ROLE'],
+            },
+        },
+    }, {
+        type: 'object',
```
naveenpaul1, line 170: why do we need this? Is it for root account?

src/server/system_services/schemas/account_schema.js
```diff
@@ -139,5 +130,46 @@ module.exports = {
...
             $ref: 'common_api#/definitions/role_config'
         },
-    }
+};
+
+module.exports = {
+    $id: 'account_schema',
+    type: 'object',
+    properties: account_properties,
+    oneOf: [{
```
naveenpaul1, line 139: can you add a comment for each object, what is actually used for, Will help in future

src/test/unit_tests/internal/test_account_schema_validation.test.js
```diff
@@ -0,0 +1,117 @@
...
+function assert_validation(account_to_validate, reason) {
+    try {
+        validate_account_schema(account_to_validate);
+        fail(reason);
+    } catch (err) {
+        expect(err.message).toBe('INVALID_SCHEMA_DB accounts');
+    }
+}
+
+function fail(reason) {
```
naveenpaul1, line 115: do we need a method just to throw?

src/test/unit_tests/internal/test_account_schema_validation.test.js
```diff
@@ -0,0 +1,117 @@
...
+const { KEYWORDS } = require('../../../util/schema_keywords');
+const common_api = require('../../../api/common_api');
+const schema_utils = require('../../../util/schema_utils');
+const account_schema = require('../../../server/system_services/schemas/account_schema');
+
+describe('account_schema validation', () => {
+
+    const OWNER = '6a9971a71ad1d20028db2249';
+
+    describe('account with all needed properties', () => {
```
naveenpaul1, line 15: can you add root account validation test also.

src/server/system_services/schemas/account_schema.js
```diff
@@ -139,5 +130,46 @@ module.exports = {
         role_config: {
             $ref: 'common_api#/definitions/role_config'
```
shirady, line 131: Is this supposed to be deleted?

src/server/system_services/schemas/account_schema.js
```diff
@@ -139,5 +130,46 @@ module.exports = {
...
+            ...account_properties,
+            identity_type: {
+                type: 'string',
+                enum: ['ROLE'],
+            },
+        },
+    }, {
+        type: 'object',
+        properties: account_properties,
+        not: { required: ['identity_type'] },
```
shirady, line 172: Can you explain this part?
I'm not sure about adding new keywords to our JSON schema.
Do you think you can avoid this?

src/util/schema_utils.js
```diff
@@ -17,8 +17,16 @@ function strictify(schema, options, base) {
             illegal_json_schema(schema, base, 'missing properties for object type');
         }
         check_schema_extra_keywords(schema, base, [
-            'type', 'properties', 'additionalProperties', 'patternProperties', 'required', 'wrapper'
+            'type', 'properties', 'additionalProperties', 'patternProperties', 'required', 'wrapper', 'oneOf', 'not'
```
shirady, line 20: Why did you had to add the `oneOf`?
I think we already used this before in the JSON schemas that we have.

See:
https://github.com/noobaa/noobaa-core/blob/ff8b91ae2671bbf075686cd53bea27463dd6568c/src/util/schema_utils.js#L54-L58

src/util/schema_utils.js
```diff
@@ -73,6 +81,8 @@ function strictify(schema, options, base) {
         check_schema_extra_keywords(schema, base, '$ref');
     } else if (schema.type === 'null') {
         check_schema_extra_keywords(schema, base, 'type');
+    } else if (_.isArray(schema.required) && _.keys(schema).length === 1) {
+        check_schema_extra_keywords(schema, base, 'required');
```
shirady, line 85: Why do you need this change?

src/test/unit_tests/internal/test_account_schema_validation.test.js
```diff
@@ -0,0 +1,117 @@
...
+        identity_type: 'ROLE',
+        owner: '6a9971a71ad1d20028db2249',
+        assume_role_policy_document: {
+            Version: '2012-10-17',
+            Statement: [{ Effect: 'Allow', Principal: { Service: 's3.amazonaws.com' }, Action: 'sts:AssumeRole' }],
+        },
+    };
+}
+
+const validate_account = (() => {
```
shirady, line 84: Can you `ajv.compile` once and use it? (instead in every test)

src/test/unit_tests/internal/test_account_schema_validation.test.js
shirady, line 1: Is it possible to avoid the new test file and reuse existing tests?

src/util/schema_utils.js
shirady, line 1: Im not sure, but maybe you want to add tests in `test_schema_keywords.js` since you added new keywords.

src/server/system_services/schemas/account_schema.js
aayushchouhan09: No, it is not related to this PR.

cc @naveenpaul1 if you want to add something here.

src/server/system_services/schemas/account_schema.js
aayushchouhan09: Yes, this branch is for default/root accounts that don’t have `identity_type` yet.

src/server/system_services/schemas/account_schema.js
aayushchouhan09: We can remove it, not required. Thanks

src/server/system_services/schemas/account_schema.js
aayushchouhan09: Updated, Thanks

src/util/schema_utils.js
aayushchouhan09: Removed, Thanks

src/util/schema_utils.js
aayushchouhan09: The existing oneOf handling is for root-level `oneOf` schemas. Here `oneOf` is next to properties on `type: 'object'`, we need root properties for postgres encode/decode, and oneOf for identity validation. That’s why it was added to the object branch in strictify too.

src/test/unit_tests/internal/test_account_schema_validation.test.js
aayushchouhan09: Removed, Thanks

src/util/schema_utils.js
aayushchouhan09: I have put the test in that file before, moving them back with proper structure. Thanks!

src/server/system_services/schemas/account_schema.js
```diff
@@ -139,5 +130,44 @@ module.exports = {
...
+            ...account_properties,
+            identity_type: {
+                type: 'string',
+                enum: ['USER'],
+            },
+        },
+    }, {
+        // ROLE identity
+        type: 'object',
+        required: ['_id', 'name', 'owner', 'assume_role_policy_document', 'identity_type'],
```
shirady: Are you sure you don't need the `email`?
I know that you created a special string for this.

src/util/schema_utils.js
shirady: @aayushchouhan09 I'mm not sure I understand, can you give an example or illustrate it somehow?

@nadavMiz since you used the `oneOf` you probably had experience with it.
I know that opened #7448 it is somehow related to the suggested changes here?

src/test/unit_tests/db/test_schema_keywords.js
```diff
@@ -13,6 +17,30 @@ const assert = require('assert');
...
+        Version: '2012-10-17',
+        Statement: [{ Effect: 'Allow', Principal: { Service: 's3.amazonaws.com' }, Action: 'sts:AssumeRole' }],
+    },
+};
+let validate_account;
+
+const oneOf_properties = {
+    kind: { type: 'string' },
+    name: { type: 'string' },
+};
```
shirady: I'm not sure separating it here contributes to the readability of the object `test_schema_keywords`.

src/test/unit_tests/db/test_schema_keywords.js
```diff
@@ -41,20 +69,48 @@ const test_schema_keywords = {
                 },
             },
         },
+        oneOf: {
+            type: 'object',
+            properties: oneOf_properties,
+            oneOf: [{
```
shirady, line 46: Double `oneOf`?
Is this the case you added or is it just for tests?

src/test/unit_tests/db/test_schema_keywords.js
```diff
@@ -1,9 +1,13 @@
 /* Copyright (C) 2016 NooBaa */
 'use strict';
 
+const _ = require('lodash');
 const mocha = require('mocha');
 const { default: Ajv } = require('ajv');
+const schema_utils = require('../../../util/schema_utils');
 const schema_keywords = require('../../../util/schema_keywords');
+const common_api = require('../../../api/common_api');
+const account_schema = require('../../../server/system_services/schemas/account_schema');
```
shirady: I think they wanted this test to be general and test utils related to schemas, and specific schema.

src/server/system_services/schemas/account_schema.js
aayushchouhan09: Added, Thanks

src/util/schema_utils.js
aayushchouhan09: The existing `oneOf` handling is for schemas where `oneOf` is the root.

`account_schema` now keeps one shared properties list for all account fields, and uses `oneOf` to define different required fields per identity (ACCOUNT/USER/ROLE).
Since `oneOf` is inside type: 'object' and not at the schema root, strictify must allow it in the object branch, otherwise it fails with "extra keywords - oneOf".

Example:
```
// root oneOf (already supported):  { oneOf: [{ type: 'integer' }, { type: 'object', ... }] }
// account_schema (needs object branch):  { type: 'object', properties: {...}, oneOf: [ACCOUNT, USER, ROLE] }
```

@nadavMiz please correct me if I am wrong  and can we proceed with it in this PR scope.

src/test/unit_tests/db/test_schema_keywords.js
aayushchouhan09: It is just for tests.

src/test/unit_tests/db/test_schema_keywords.js
aayushchouhan09: We need a separate file for other schema validations.

src/sdk/accountspace_nb.js
```diff
@@ -139,7 +139,18 @@ class AccountSpaceNB {
...
+            params.role_name, requesting_account._id);
+        const req = {
+            role_name: params.role_name,
+            email: role_email_wrapped,
+            iam_path: params.iam_path,
+            assume_role_policy_document: params.assume_role_policy_document,
+            description: params.description,
+            max_session_duration: params.max_session_duration,
+        };
+        return account_sdk.rpc_client.account.create_role(req);
```
shirady, line 153: Can we avoid this by passing the params as we did and also add the email (destruct is and then the email).

src/server/system_services/schemas/account_schema.js
shirady: I think it is part of schema changes, I understand it's not in this PR, but just to make sure you don't forget this.

src/test/unit_tests/db/test_schema_keywords.js
shirady: I think It's confusing.

src/test/unit_tests/db/test_schema_keywords.js
shirady: Now I understand the need, thanks.

src/sdk/accountspace_nb.js
aayushchouhan09: Updated, Thanks

src/test/unit_tests/db/test_schema_keywords.js
aayushchouhan09: we can rename it, changed it to `params_one_of`. Thanks

naveenpaul1 (approved):

src/server/system_services/schemas/account_schema.js
```diff
@@ -139,5 +130,44 @@ module.exports = {
...
+module.exports = {
+    $id: 'account_schema',
+    type: 'object',
+    properties: account_properties,
+    oneOf: [{
+        // ACCOUNT identity (identity_type defaults to ACCOUNT when omitted)
+        type: 'object',
+        required: ['_id', 'name', 'email', 'has_login'],
+        properties: {
+            ...account_properties,
```
nadavMiz: why are you referencing  account_properties like this? why not use $ref like everywhere else?

src/server/system_services/schemas/account_schema.js
```diff
@@ -139,5 +130,44 @@ module.exports = {
...
+            identity_type: {
```
nadavMiz, line 144: if all combination use identity_type. why not put it inside account_properties. is there a combination without it? I can't see it in this PR

aayushchouhan09: we added it to account_properties in this PR for the shared type/enum; the per-branch `identity_type` in `oneOf` is only to tell the branches apart (all three require it, none omit it).

src/server/system_services/schemas/account_schema.js
aayushchouhan09: Removed it - root `properties: account_properties` covers field types; `oneOf` branches only narrow `identity_type` per branch.

nadavMiz (approved):
LGTM

PR #10003, opened 2026-09-01
NC | Memory allocator change to jemalloc + redesign to memory monitor debug logs

jackyalbo:
### Describe the Problem
We had an issue where we used more and more RSS memory on Scale clusters. The issue seems to be with the default malloc memory allocator; jemalloc does a better job of returning freed memory to the OS.

### Explain the Changes
1. Adding jeamalloc to noobaa-core under lib and using it as default allocator using LD_PRELOAD
2. Setting default free memory configuration to 2sec for dirty/muzzy pages before they return to the OS
3. Adding more informative logs in memory monitor - now we can see how much buffers are being used from which pool and more elaborate info on the different memory types being used

1. Fixed https://redhat.atlassian.net/browse/DFBUGS-6917 

### Testing Instructions:
1. Run noobaa under heavy load - I used wrap:
```
warp mixed --host=H --access-key=AK --secret-key=SK --obj.size=100m --duration=5m --disable-multipart --concurrent=200 --bucket=B --noprefix --insecure --tls
```
3. Stop the load - see that the memory used is back to BUFFERS_MEM_LIMIT for each fork

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **New Features**
  - RPM deployments now bundle jemalloc and include configurable memory-management settings.
  - Administrators can override NooBaa memory settings through the installed service configuration.
  - RPM deployments now include the service configuration file automatically.
  - Buffer-pool statistics now include pooled, active, waiting, live, and allocated memory.

- **Bug Fixes**
  - Glacier objects now honor forced eviction requests during stream reads.

- **Diagnostics**
  - Panic and memory-monitoring reports now include buffer-pool usage, RSS, and unaccounted memory metrics.


src/deploy/standalone/noobaa.sysconfig
```diff
@@ -0,0 +1,10 @@
+# NooBaa endpoint process environment (loaded by noobaa.service).
+# Override locally as needed; RPM installs this as %config(noreplace).
+
+# jemalloc is bundled in the RPM (see /usr/local/noobaa-core/lib/).
+# Clear LD_PRELOAD= to fall back to glibc malloc.
+LD_PRELOAD=/usr/local/noobaa-core/lib/libjemalloc.so.2
+
+# Decay dirty/muzzy pages over ~2s on a background thread (jemalloc only).
+# Tune dirty_decay_ms / muzzy_decay_ms if needed (e.g. 5000 for gentler purge).
+MALLOC_CONF=dirty_decay_ms:2000,muzzy_decay_ms:2000,background_thread:true
```
dannyzaken: Did you test how it affects latency (specifically p99)? I am worried that 2 seconds may trigger more allocations and deallocations unnecessarily. Any reason not to use the 10 seconds default? This is also the setting we use for containerized, so it is already tested in the field. Maybe it's better to be aligned for both.

As far as I understand, the `background_thread` option is the more important one here, since it will perform the decay periodically and not lazily, so after ~20 seconds of being idle, the RSS should drop. I think this is reasonable. We can consider the same background_thread setting also for containerized. 

```suggestion
MALLOC_CONF=dirty_decay_ms:10000,muzzy_decay_ms:10000,background_thread:true
```

src/util/panic.js
```diff
@@ -40,21 +46,94 @@ function enable_heapdump(name, next_mb, step_mb) {
...
+                pools_part = format_buffers_pool_stats(c.buffers_pool_stats());
+            } catch (err) {
+                pools_part = ` | pools_stats_error ${err.message}`;
+            }
+        }
+        // Also trigger on rss: heapUsed alone misses native/buffer growth
+        // (ArrayBuffers, jemalloc retention) that dominates NSFS RSS.
+        const should_log = c.logging_threshold &&
+            (c.logging_threshold <= current || c.logging_threshold <= m.rss);
+        if (should_log) {
```
dannyzaken, line 111: why not skippin everything if `should_log == true`?

jackyalbo: done

src/deploy/RPM_build/RPM.Dockerfile
```diff
@@ -14,6 +14,10 @@ ARG BUILD_S3SELECT_PARQUET=0
 # Install GCC11 toolchain on Centos8 to match the default toolchain of Centos9
 RUN if [ "$CENTOS_VER" == "8" ]; then dnf install -y -q gcc-toolset-11; fi
 
+# jemalloc is bundled into the RPM (not a runtime Requires). Needed at build time
+# to copy libjemalloc.so* into /usr/local/noobaa-core/lib/.
+RUN dnf install -y -q epel-release && dnf install -y -q jemalloc && dnf clean all
```
liranmauda, line 19: Are we using this image, or is it just a means to get the RPM?
If this image is just an intermediate one, divide the directive into 2 directives for easier debugging if it fails, and avoid the clean 

Having it in a single directive only helps if we are using the image and not squashing the layers, and want the image size to be minimal 

```suggestion
RUN dnf install -y -q epel-release 
RUN dnf install -y -q jemalloc 
```

liranmauda (approved):

PR #10022, opened 2026-09-07
DEEP_ARCHIVE | Integration Tests for Restore Object (Using NC)

shirady:
### Describe the Problem
Added integration tests for testing `RestoreObject` when the namespace store is an NC address as endpoint - as part of [RHSTOR-8823](https://redhat.atlassian.net/browse/RHSTOR-8823).
Until now, we used `test_deep_archive_via_s3.js` which used noobaa bucket as the target bucket, but since `RestoreObject` API is not implemented there we could not test it (we had mocks and used a unit test like `npx jest src/test/unit_tests/internal/test_namespace_multi_storage_class_restore.test.js`).

### Explain the Changes
1. Added AI-generated and dedicated tests for RestoreObject, to test the API and to allow testing bucket policy between 2 accounts (in `src/test/integration_tests/api/s3/test_restore_object.js`). Also added running NC for the target bucket on the container (in `src/test/utils/nc_archive_target.js`).
2. Moved mutual functions from `nc_coretest.js` so they can be reused (in `nc_nsfs_test_helpers.js`).
3. Added the new test in `src/test/utils/index/index.js` so it will run in the CI.

### Issues: 
List of GAPs:
1. We still do not have tests for the `RestoreWorker` BG worker. 

#### Automatic Tests:
1. Please run: `make run-single-test testpath=integration_tests/api/s3/test_restore_object.js`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Tests**
  * Added integration coverage for restoring objects from deep-archive storage.
  * Verified successful restores, duplicate restore rejection, cross-account access denial, and bucket-policy authorization.
  * Added coverage for standalone archive targets and archive configuration.
  * Improved test environment startup readiness checks, shared setup, and cleanup reliability for storage-related scenarios.

### DIFF
--- a/src/test/integration_tests/api/s3/test_restore_object.js
+++ b/src/test/integration_tests/api/s3/test_restore_object.js
@@ -1,0 +1,261 @@
+/* Copyright (C) 2026 NooBaa */
+'use strict';
+
+/**
+ * RestoreObject integration tests (owner, repeat restore, cross-account, bucket policy).
+ *
+ * Containerized coretest hosts the user bucket (archive_policy + MSC). A standalone
+ * NC nsfs archive target (see nc_archive_target.js, same pattern as nc_coretest) provides
+ * the deep-archive S3 target — not loopback to coretest, which returns NotImplemented
+ * on archive payload RestoreObject.
+ */
+
+const coretest = require('../../../utils/coretest/coretest');
+coretest.setup({ pools_to_create: [coretest.POOL_LIST[1]] });
+
+const mocha = require('mocha');
+const assert = require('assert');
+
+const config = require('../../../../../config');
+const s3_utils = require('../../../../endpoint/s3/s3_utils');
+const test_utils = require('../../../system_tests/test_utils');
+const { err_code, generate_s3_policy, generate_s3_client } = test_utils;
+const { start_nc_archive_target, stop_nc_archive_target } = require('../../../utils/nc_archive_target');
+
+const { rpc_client } = coretest;
+
+const NC_ARCHIVE_TARGET_BUCKET = 'test-restore-archive-target';
+const ARCHIVE_CONNECTION = 'restore_nc_archive_connection';
+const ARCHIVE_NSR = 'restore_nc_archive_nsr';
+
+const RESTORE_BKT = 'test-restore-object';
+const RESTORE_OWNER = 'restore-owner';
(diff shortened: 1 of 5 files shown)


src/test/integration_tests/api/s3/test_restore_object.js
```diff
@@ -0,0 +1,256 @@
...
+                    });
+                }
+                await s3_nc_archive.deleteBucket({ Bucket: NC_ARCHIVE_TARGET_BUCKET });
+            }
+        } catch (err) {
+            console.warn('restore_object: cleanup:', err);
+        } finally {
+            await stop_nc_archive_target();
+            config.ARCHIVE_TARGET_BUCKET_CHECK_ENABLED = original_archive_target_check_enabled;
+            config.ARCHIVE_POLICY_STORAGE_CLASS_CHECK_ENABLED = original_archive_policy_check_enabled;
```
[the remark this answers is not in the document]
shirady: Why is this still open?

src/test/utils/coretest/nc_coretest.js
```diff
@@ -217,14 +218,7 @@ async function config_dir_setup() {
     await fs.promises.mkdir(config.NSFS_NC_DEFAULT_CONF_DIR, { recursive: true });
     await fs.promises.mkdir(NC_CORETEST_CONFIG_DIR_PATH, { recursive: true });
```
romayalon, line 219: why not use the helpers you created?

shirady: I'm not sure I understand. Which function did you recommend I reuse?

romayalon: per my understanding some of the new code in nc_nsfs_test_helpers are similar if not equal to nc_coretest

romayalon (approved):

src/test/utils/coretest/nc_coretest.js
shirady: You understand correctly. I tried to avoid duplicated code between `nc_coretest.js`  and the new `nc_archive_target.js` but the specific function you pointed out was not shared.

romayalon: I just randomly chose a line in the code... I wanted to say a general note that we should re-use

